PR #1237, opened 2015-11-09
refactor traits::GetIonizer

psychocoderHPC:
- remove fallback to `particles::ionization::None<...>`
- `MySimulation.hpp` - filter species with ionizer before call of `CallIonization<>`

PR #1238, opened 2015-11-09
remove ionizer `None`

psychocoderHPC:
followup of #1237

ionizer `None` is useless since we have a clean way to call only
algorithm which are supported by an specie

- [x] rebase against #1237


psychocoderHPC:
ready for review

src/picongpu/include/simulationControl/MySimulation.hpp
```diff
@@ -63,6 +63,7 @@
 #include "particles/InitFunctors.hpp"
 #include "particles/memory/buffers/MallocMCBuffer.hpp"
 #include "particles/traits/FilterByFlag.hpp"
+#include "particles/traits/FilterByFlag.hpp"
```
erikzenker: Why this double include ?

psychocoderHPC: Because it's wrong^^, thx for spotting

psychocoderHPC:
please review

PR #1239, opened 2015-11-09
change text size and gui layout

psychocoderHPC:
- change text size from 11 to 14 points
- fix layous to support new text size


ax3l:
Sweet, HD ;)

PR #1240, opened 2015-11-09
disable that fps panel hide function

psychocoderHPC:
the "presentation mode" / disable-all key `F11` shall not hide the timestep/fps pane


ax3l:
(can still be hidden/shown individually via F6)

PR #1243, opened 2015-11-10
refactor particle box interface

### LINKED ISSUE #1191: Make getEmptyFrame return a pointer
The function `getEmptyFrame` of the particle box currently returns a reference. However it is possible, that no memory was available (mallocMC out of memory) making the function returning a reference to the dereferenced null pointer.

It might even be a good idea to pass around `FramePtr`s instead if `FRAME&`. This would make e.g. `getNextFrame` easier, as the pointer can be checked for validy directly instead of introducing an additional variable that is passed by reference.

psychocoderHPC:
- refactor interface of the `ParticlesBox`
- use `FramePointer<...>` instead of plain C pointer
- use coding guide line for code formation

Sry that this pull request contains code style changes, I pressed button for auto formation in my IDE.

This pull close #1191

**Tests:**
- [x] KHI electrons/positrons heating
- [x] KHI macro particles count

### DIFF
--- a/src/picongpu/include/fields/FieldTmp.kernel
+++ b/src/picongpu/include/fields/FieldTmp.kernel
@@ -47,7 +47,7 @@ namespace picongpu
     template<class BlockDescription_, uint32_t AREA, class TmpBox, class ParBox, class FrameSolver, class Mapping>
     __global__ void kernelComputeSupercells( TmpBox fieldTmp, ParBox boxPar, FrameSolver frameSolver, Mapping mapper )
     {
-
+        typedef typename ParBox::FramePtr FramePtr;
         typedef typename BlockDescription_::SuperCellSize SuperCellSize;
         const DataSpace<simDim> block( mapper.getSuperCellIndex( DataSpace<simDim > ( blockIdx ) ) );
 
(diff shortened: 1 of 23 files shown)


psychocoderHPC:
This pull request reduce the register count in many kernel. The speed improvement is very small or in the range of the measurement error.

psychocoderHPC:
- added the new line to improve the human readability
- size type of the offset is [not changed ](https://github.com/psychocoderHPC/picongpu/commit/b5609ecec142d1e36b940e2ae044b4a2d51ca123#commitcomment-14351033)

src/picongpu/include/fields/FieldTmp.kernel
```diff
@@ -55,19 +55,19 @@ namespace picongpu
...
         if( linearThreadIdx == 0 )
         {
-            frame = &( boxPar.getLastFrame( block, isValid ) );
+            frame = boxPar.getLastFrame( block );
             particlesInSuperCell = boxPar.getSuperCell( block ).getSizeLastFrame( );
         }
         __syncthreads( );
 
-        if( !isValid )
+        if( frame.isValid() )
```
ax3l: regression spotted in #1271: fixed in #1272

Issue #1194, opened 2015-10-19
[cuSTL] Assignment of Buffers ignores sizes/pitches...


Flamefire:
The code for the copy-operator is basically:

    template<typename DBuffer>
    HostBuffer<Type, dim>& operator=(const DBuffer& rhs)
    {
        cudaWrapper::Memcopy<dim>()(this->dataPointer, this->pitch, rhs.getDataPointer(), rhs.getPitch(),
                                this->_size, cudaWrapper::flags::Memcopy::deviceToHost);

        return *this;
    }

Similar for the CartBuffer itself: `Copier::copy(this->dataPointer, this->pitch, rhs.dataPointer, rhs.pitch, rhs._size);`

If the source and destination sizes are not equal then the result is at least unexpected as you expect to have the same values after an assignment

ax3l:
@Heikman pls take a look

ax3l:
closed with  #1248

PR #1248, opened 2015-11-12
Add asserts to cuSTL containers

Heikman:
Add asserts to cuSTL containers checking sizes at assignment.

@Flamefire 

solves #1194

### DIFF
--- a/src/libPMacc/include/cuSTL/container/DeviceBuffer.hpp
+++ b/src/libPMacc/include/cuSTL/container/DeviceBuffer.hpp
@@ -31,6 +31,9 @@
 
 #include <memory/buffers/DeviceBuffer.hpp>
 #include <memory/buffers/HostBuffer.hpp>
+#include <exception>
+#include <sstream>
+#include <boost/assert.hpp>
 
 namespace PMacc
 {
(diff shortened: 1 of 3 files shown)


Flamefire:
`assert()` is only evaluated in debug builds. I'd at least make this a runtime error (e.g. throw std::invalid_argument) and mention the size restriction in the documentation.
Besides: You use the boost macro in 2 places and plain `assert` in 2 other in a single PR. Pretty inconsistent.

Heikman:
Now, the assert is suppressed on device because first it is very expensive yet and second the assign operator is even just called hypothetically on device so far.

Flamefire:
I'd still use runtime errors instead of asserts.
You can still use an `assert` (plain version) for the device side of `HDINLINE` functions (only `CartBuffer`)

Heikman:
done.

Flamefire:
Small issue: Boost headers should be included before STL headers
Big issue: Still no runtime checks. `BOOST_ASSERT` is equal to `assert` but can be ***disabled*** separately.

Heikman:
done.

psychocoderHPC:
@Flamefire: Did you accept the last changes that `assert` is only used on device?

src/libPMacc/include/cuSTL/container/DeviceBuffer.hpp
```diff
@@ -23,6 +23,7 @@
 
 #pragma once
 
+#include <boost/assert.hpp>
```
Flamefire: Please move this down before the std includes

src/libPMacc/include/cuSTL/container/HostBuffer.hpp
```diff
@@ -22,6 +22,7 @@
 
 #pragma once
 
+#include <boost/assert.hpp>
```
Flamefire: Move down

Flamefire:
Yes it is ok since there are no exceptions on the device AFAIK and this is better overhead-wise.
However I'd add a temporary for the stringstream which makes the code cleaner and removes the typecast.
Also 2 small issues with include order

psychocoderHPC:
@Heikman  please move the include marked by @Flamefire down, after that I can merge this pull request

Heikman:
done

PR #1155, opened 2015-09-30
Closes #1154: Introduce CONSTEXPR macro

Flamefire:
This introduces a new define `CONSTEXPR` which should be used instead of `const` for all in-class `static` variables. It *must* be used for declaring `static const float foo = 42.1337` as this is an error for C++11 and was probably never allowed. (See #1154)

It does not hurt to use this for all `static const`s used and may even increase performance in C++11 mode.

However it does not try to emulate `constexpr` in any way for C++98. So functions cannot be decorated with this (e.g. `CONSTEXPR inc(int x){ return x+1; }` which would be possible in C++11 but is **NOT** allowed for C++98.

PR #1250, opened 2015-11-12
fix missing replacement of CONSTEXPR to BOOST_CONSTEXPR_OR_CONST

PrometheusPi:
This is a **follow up** pull request to #1155 where some `CONSTEXPR` in `laserConfig.param` were not replaced by `BOOST_CONSTEXPR_OR_CONST`.

@Flamefire please have a look if I used `BOOST_CONSTEXPR_OR_CONST` correctly.

Using `CONSTEXPR` does not compile.


Flamefire:
Thanks, this is correct :sparkles:
The `CONSTEXPR` macro was used before but later replaced by the boost stuff. Seems I missed some of them.

PR #995, opened 2015-07-31
use boost:::threads for png creation

psychocoderHPC:
- `CMakeLists.txt`: add `thread` to boost requirements (#1251)
- add method `join()` to
  - `PngCreator`
  - `LiveViewClient`
- `PngCreator.hpp`: use `boos::threads` for asynchronous png creation

- [x] Tests with the current dev
- [x] rebase against #1251

PR #1251, opened 2015-11-13
Boost: 1.56.0+ Requirement

ax3l:
This PR increases the boost requirement to be at least 1.56.0+ (first required by #995).

This allows us to drop a list of work-arounds as a result.

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -51,13 +51,13 @@ Requirements
   - *Debian/Ubuntu:* `sudo apt-get install zlib1g-dev`
   - *Arch Linux:* `sudo pacman --sync zlib`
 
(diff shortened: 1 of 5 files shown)


Flamefire:
I have not pulled for a while so I've just seen this now. Was it really necessary to upgrade to boost 1.56 as the minimum? Most linux distros provide only 1.55 as the highest version via the package managers and self-compiling boost is at least an additional difficulty (e.g. what about "regular" users on a large HPC system? would that even work correctly with submit queues etc?)

ax3l:
yes that is unfortunate but solves a bug in `boost::thread` in previous versions and allows up to skip a good amount of work-arounds.

Our `INSTALL.md` shows a two-liner as in [this gist](https://gist.github.com/ax3l/7450482#example-build) to build and install boost.

PR #1252, opened 2015-11-13
Add a policy to GetKeyFromAlias

Flamefire:
This allows e.g. `ResolveAndRemoveFromSeq` to not fail if a value did not exist in the first place.

Current policies are:

1. Fail if alias is not in the sequence
2. Return the (supposed to be) alias
3. Return a given type

Default policy is to return bmpl::void_ so it does not change current behaviour.

Partly based on offline discussion with @psychocoderHPC

### DIFF
--- a/src/libPMacc/include/compileTime/GetKeyFromAlias.hpp
+++ b/src/libPMacc/include/compileTime/GetKeyFromAlias.hpp
@@ -45,7 +45,7 @@ namespace PMacc
  */
 template<typename T_MPLSeq,
          typename T_Key,
-	     typename T_KeyNotFoundPolicy = errorHandlerPolicies::ReturnType<>
+         typename T_KeyNotFoundPolicy = errorHandlerPolicies::ReturnType<>
 >
 struct GetKeyFromAlias
 {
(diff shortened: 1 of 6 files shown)


src/libPMacc/include/compileTime/GetKeyFromAlias.hpp
```diff
@@ -35,16 +35,26 @@
...
+ *
+ * \tparam T_MPLSeq Sequence of keys to search
+ * \tparam T_Key Key or alias of a key in the sequence
+ * \tparam T_KeyNotFoundPolicy Binary meta-function that is called like (T_MPLSeq, T_Key)
+ *         when T_Key is not found in the sequence. Default is to return bmpl::void_
+ */
 template<typename T_MPLSeq,
-         typename T_Key
+         typename T_Key,
+	     typename T_KeyNotFoundPolicy = errorHandlerPolicies::ReturnType<>
```
psychocoderHPC: wrong alignment

Flamefire: A tab slipped through... Fixed

Issue #1256, opened 2015-11-16
`SliceFieldPrinter` hard coded block size for kernel call


psychocoderHPC:
SliceFieldPrinter assume that the local grid size is divideable by 4 in a 3D and 16 in a 2D simulation.

see: https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/plugins/SliceFieldPrinter.tpp#L151
and 
https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/plugins/SliceFieldPrinter.tpp#L161

psychocoderHPC:
closed with #1309

PR #1257, opened 2015-11-17
Fix SliceFieldPrinter

Heikman:
Set the blockDim parameter to (1,1,1) in SliceFieldPrinter. Before this it was set to (4,4,1) (3D) respectively (16,1,1) (2D) and the container sizes were restricted to be multiples of these values.

Solves #1256

### DIFF
--- a/src/picongpu/include/plugins/SliceFieldPrinter.tpp
+++ b/src/picongpu/include/plugins/SliceFieldPrinter.tpp
@@ -148,7 +148,7 @@ void SliceFieldPrinter<Field>::printSlice(const TField& field, int nAxis, float 
 
     /* convert data to higher precision and to SI units */
     SliceFieldPrinterHelper::ConversionFunctor<Field> cf;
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
Please run a test!

ax3l:
isn't it possible to be a tiny bit more sophisticated in block size selection?
(e.g., use super cell size if it fits the data extends?)

psychocoderHPC:
Please use the super cells size  to define to blocksize that we not lose to much performance

ax3l:
that requires an "over-lengths" handling for the residuum in case the
size of the container mod super cell is != 0

psychocoderHPC:
The better way is to use the supercell size. twisted accessor should solve the problem but I am but sure if runtime transformation is supported.

ax3l:
yes, for field slicing the super-cell size should work fine. just needs to take care of the according axis in case of twisting and 2D zones.

Heikman:
I close this PR because #1309 is a better solution.

PR #1265, opened 2015-11-23
Gas Gaussian Cloud: Add Missing Documentation

ax3l:
This adds the missing documentation in `gasConfig.param` for the GaussianCloud profile.

Was removed during refactoring since even the [original documentation](https://github.com/ComputationalRadiationPhysics/picongpu/blob/f31b31ab5afe8951ba58754fc6266a4ac07ffd20/src/picongpu/include/simulation_defines/param/gasConfig.param#L167-L173) was copy-pasted from an other profile (and therefore wrong).

### DIFF
--- a/examples/Bunch/include/simulation_defines/param/gasConfig.param
+++ b/examples/Bunch/include/simulation_defines/param/gasConfig.param
@@ -51,7 +51,7 @@ PMACC_STRUCT(GaussianCloudParam,
      *     exponent = |globalCellPos - center| / sigma
      *     density = e^[ gasFactor * exponent^gasPower ]
      */
-    (PMACC_C_VALUE(float_X, gasFactor, -1.0))
+    (PMACC_C_VALUE(float_X, gasFactor, -0.5))
     (PMACC_C_VALUE(float_X, gasPower, 2.0))
 
     /** height of vacuum area on top border
--- a/examples/Bunch/include/simulation_defines/param/gasConfig.param
+++ b/examples/Bunch/include/simulation_defines/param/gasConfig.param
@@ -1,5 +1,5 @@
 /**
- * Copyright 2013-2014 Axel Huebl, Heiko Burau, Rene Widera, Felix Schmitt,
+ * Copyright 2013-2015 Axel Huebl, Heiko Burau, Rene Widera, Felix Schmitt,
  *                     Richard Pausch
  *
  * This file is part of PIConGPU.
@@ -47,6 +47,13 @@ namespace gasProfiles
 {
 
 PMACC_STRUCT(GaussianCloudParam,
+    /** Gas Formula:
+     *     exponent = |globalCellPos - center| / sigma
+     *     density = e^[ gasFactor * exponent^gasPower ]
+     */
+    (PMACC_C_VALUE(float_X, gasFactor, -1.0))
+    (PMACC_C_VALUE(float_X, gasPower, 2.0))
+
     /** height of vacuum area on top border
      *
      * this vacuum is important because of the laser initialization,
@@ -54,9 +61,6 @@ PMACC_STRUCT(GaussianCloudParam,
      * unit: cells
      */
     (PMACC_C_VALUE(uint32_t, vacuumCellsY, 50))
-
-    (PMACC_C_VALUE(float_X, gasFactor, -1.0))
-    (PMACC_C_VALUE(float_X, gasPower, 2.0))
 
     /** The central position of the gas distribution
      *  unit: meter
--- a/src/picongpu/include/simulation_defines/param/gasConfig.param
+++ b/src/picongpu/include/simulation_defines/param/gasConfig.param
@@ -138,6 +138,13 @@ typedef LinearExponentialImpl<LinearExponentialParam> LinearExponential;
 
 
 PMACC_STRUCT(GaussianCloudParam,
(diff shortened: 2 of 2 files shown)


src/picongpu/include/simulation_defines/param/gasConfig.param
```diff
@@ -163,6 +167,7 @@ PMACC_STRUCT(GaussianCloudParam,
 /* definition of gas cloud profile*/
 typedef GaussianCloudImpl<GaussianCloudParam> GaussianCloud;
 
+
```
PrometheusPi: why is here a newline added?

ax3l: because all profiles are added by two newlines, this one was not.

ax3l: added more below now.

examples/Bunch/include/simulation_defines/param/gasConfig.param
```diff
@@ -47,6 +47,13 @@ namespace gasProfiles
 {
 
 PMACC_STRUCT(GaussianCloudParam,
+    /** Gas Formula:
+     *     exponent = |globalCellPos - center| / sigma
+     *     density = e^[ gasFactor * exponent^gasPower ]
+     */
+    (PMACC_C_VALUE(float_X, gasFactor, -1.0))
```
PrometheusPi: Could we use -1/2 as default value here as the standard [Gauss distribution](https://en.wikipedia.org/wiki/Gaussian_function)?

src/picongpu/include/simulation_defines/param/gasConfig.param
```diff
@@ -138,6 +138,13 @@ typedef LinearExponentialImpl<LinearExponentialParam> LinearExponential;
 
 
 PMACC_STRUCT(GaussianCloudParam,
+    /** Gas Formula:
+     *     exponent = |globalCellPos - center| / sigma
+     *     density = e^[ gasFactor * exponent^gasPower ]
+     */
+    (PMACC_C_VALUE(float_X, gasFactor, 1.0))
+    (PMACC_C_VALUE(float_X, gasPower, 4.0))
```
PrometheusPi: same question here: why different than standard Gauss?

src/picongpu/include/simulation_defines/param/gasConfig.param
PrometheusPi: :+1:

PrometheusPi:
looks godd to me - just need a answer or  a fix to my two questions. 
@ax3l great job

examples/Bunch/include/simulation_defines/param/gasConfig.param
ax3l: I did not want to change your definition of your `Bunch` example (this line).
I would definitely change the default, good point.

Do you want to change it in the Bunch example as well?

src/picongpu/include/simulation_defines/param/gasConfig.param
ax3l: oh yes, it's even wrong this way...

examples/Bunch/include/simulation_defines/param/gasConfig.param
PrometheusPi: Yes - change the Bunch example as well.

ax3l: let's use `-0.5` and `2.0` for a real gauss

src/picongpu/include/simulation_defines/param/gasConfig.param
ax3l: let's use `-0.5` and `2.0` for a real gauss

examples/Bunch/include/simulation_defines/param/gasConfig.param
ax3l: just the factor or also the power? (sry for dbl post)

ax3l:
updated, thanks! :)

examples/Bunch/include/simulation_defines/param/gasConfig.param
```diff
@@ -47,6 +47,13 @@ namespace gasProfiles
 {
 
 PMACC_STRUCT(GaussianCloudParam,
+    /** Gas Formula:
+     *     exponent = |globalCellPos - center| / sigma
+     *     density = e^[ gasFactor * exponent^gasPower ]
+     */
+    (PMACC_C_VALUE(float_X, gasFactor, -0.5))
```
ax3l: changed :)

src/picongpu/include/simulation_defines/param/gasConfig.param
```diff
@@ -138,6 +138,13 @@ typedef LinearExponentialImpl<LinearExponentialParam> LinearExponential;
 
 
 PMACC_STRUCT(GaussianCloudParam,
+    /** Gas Formula:
+     *     exponent = |globalCellPos - center| / sigma
+     *     density = e^[ gasFactor * exponent^gasPower ]
+     */
+    (PMACC_C_VALUE(float_X, gasFactor, -0.5))
```
ax3l: fixed :)

ax3l:
@PrometheusPi can be merged :)

PR #1268, opened 2015-11-26
Test checks setValue of HostBufferIntern

erikzenker:
This PR test the method `setValue` of HostBufferIntern by calling `setValue(255)` and checking afterwards wether the values have been set correctly by iterating over all elements in the HostBufferIntern.

### DIFF
--- a/src/libPMacc/CMakeLists.txt
+++ b/src/libPMacc/CMakeLists.txt
@@ -55,6 +55,5 @@ target_link_libraries(check ${LIBS})
 # CTest
 enable_testing()
 add_test(PMacc_test_build "${CMAKE_COMMAND}" --build ${CMAKE_BINARY_DIR} --target check)
-# --alow-run-as-root because some ci run test as root
 add_test(PMacc_test_run  mpiexec -n 1 ./check --log_level=test_suite)
-set_tests_properties(PMacc_test_run PROPERTIES DEPENDS graybat_check_build)
+set_tests_properties(PMacc_test_run PROPERTIES DEPENDS PMacc_test_build)
--- a/src/libPMacc/CMakeLists.txt
+++ b/src/libPMacc/CMakeLists.txt
@@ -56,5 +56,5 @@ target_link_libraries(check ${LIBS})
 enable_testing()
 add_test(PMacc_test_build "${CMAKE_COMMAND}" --build ${CMAKE_BINARY_DIR} --target check)
 # --alow-run-as-root because some ci run test as root
-add_test(PMacc_test_run  mpiexec --allow-run-as-root -n 2 check )
+add_test(PMacc_test_run  mpiexec -n 1 ./check --log_level=test_suite)
 set_tests_properties(PMacc_test_run PROPERTIES DEPENDS graybat_check_build)
--- a/src/libPMacc/test/memory/HostBufferIntern/setValue.hpp
+++ b/src/libPMacc/test/memory/HostBufferIntern/setValue.hpp
@@ -47,9 +47,10 @@ struct setValueTest
             const Data value = 255;
             hostBufferIntern.setValue(value);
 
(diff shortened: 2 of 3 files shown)


src/libPMacc/CMakeLists.txt
```diff
@@ -56,5 +56,5 @@ target_link_libraries(check ${LIBS})
 enable_testing()
 add_test(PMacc_test_build "${CMAKE_COMMAND}" --build ${CMAKE_BINARY_DIR} --target check)
 # --alow-run-as-root because some ci run test as root
-add_test(PMacc_test_run  mpiexec --allow-run-as-root -n 2 check )
+add_test(PMacc_test_run  mpiexec -n 1 ./check --log_level=test_suite)
```
psychocoderHPC: why you are removing `--allow-run-as-root` and the usage of two mpi instances?

src/libPMacc/test/memory/HostBufferIntern/setValue.hpp
```diff
@@ -0,0 +1,65 @@
...
+        {
+            ::PMacc::DataSpace<T_Dim::value> const dataSpace = ::PMacc::DataSpace<T_Dim::value>::create(nElementsPerDim[i]);
+            ::PMacc::HostBufferIntern<Data, T_Dim::value> hostBufferIntern(dataSpace);
+            
+            const Data value = 255;
+            hostBufferIntern.setValue(value);
+
+            for(size_t j = 0; j < static_cast<size_t>(dataSpace.productOfComponents()); ++j)
+            {
+                BOOST_CHECK_EQUAL( hostBufferIntern.getPointer()[j], value );
```
psychocoderHPC: Could you please move `getPointer()` out of the loop, this is a expensive call which should not used in a loop.

```C++
PMACC_AUTO(ptr,hostBufferIntern.getPointer());
for(size_t j = 0; j < static_cast<size_t>(dataSpace.productOfComponents()); ++j)
{
    BOOST_CHECK_EQUAL( ptr[j], value );
}
```

src/libPMacc/CMakeLists.txt
```diff
@@ -56,5 +56,5 @@ target_link_libraries(check ${LIBS})
...
 set_tests_properties(PMacc_test_run PROPERTIES DEPENDS graybat_check_build)
```
psychocoderHPC: Uhh this is currently in the dev branch but I think we don't use `graybat` ^^

psychocoderHPC:
Thanks for this great tests!

src/libPMacc/CMakeLists.txt
erikzenker: Copy and paste ....a good source of failures ^^

src/libPMacc/test/memory/HostBufferIntern/setValue.hpp
erikzenker: Since, this is a test case expensive calls should not matter. But you are right, we are not testing the event system ...

src/libPMacc/CMakeLists.txt
erikzenker: Cmake had some problems to call mpiexec with these arguments on the hypnos cluster. I will remove the comment. At the point in time when we start testing in a ci fashion we need to think about how to start the application.

src/libPMacc/CMakeLists.txt
erikzenker: fixed

src/libPMacc/test/memory/HostBufferIntern/setValue.hpp
erikzenker: fixed

erikzenker:
Are there some border cases we might need to check ?

psychocoderHPC:
There is no need to check border cases. The method `setValues()` fill the full buffer with the given value.

PR #1269, opened 2015-11-26
Fixed failure in getBasePointer of HostBufferIntern

erikzenker:
- Uses this->getPointer as much as possible
- Fixes getBasePointer
- Renames ownPointer to ownsPointer
- [ ] merge after #1283

### DIFF
--- a/src/libPMacc/include/memory/buffers/HostBufferIntern.hpp
+++ b/src/libPMacc/include/memory/buffers/HostBufferIntern.hpp
@@ -41,25 +41,31 @@ class HostBufferIntern : public HostBuffer<TYPE, DIM>
 {
 public:
 
-    typedef typename DeviceBuffer<TYPE, DIM>::DataBoxType DataBoxType;
+    typedef typename HostBuffer<TYPE, DIM>::DataBoxType DataBoxType;
 
     /** constructor
      *
      * @param size extent for each dimension (in elements)
      */
     HostBufferIntern(DataSpace<DIM> size) :
-    HostBuffer<TYPE, DIM>(size, size),
-    pointer(NULL),ownPointer(true)
+        HostBuffer<TYPE, DIM>(size, size),
+        pointer(NULL),
+        basePointer(NULL),
+        ownsPointer(true)
     {
         CUDA_CHECK(cudaMallocHost(&pointer, size.productOfComponents() * sizeof (TYPE)));
+        basePointer = pointer;
         reset(false);
     }
 
     HostBufferIntern(HostBufferIntern& source, DataSpace<DIM> size, DataSpace<DIM> offset=DataSpace<DIM>()) :
-    HostBuffer<TYPE, DIM>(size, source.getPhysicalMemorySize()),
-    pointer(NULL),ownPointer(false)
+        HostBuffer<TYPE, DIM>(size, source.getPhysicalMemorySize()),
+        pointer(NULL),
+        basePointer(NULL),
+        ownsPointer(false)
     {
         pointer=&(source.getDataBox()(offset));/*fix me, this is a bad way*/
+        basePointer = source.getPointer();
         reset(true);
     }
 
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/memory/buffers/HostBufferIntern.hpp
```diff
@@ -70,9 +76,9 @@ class HostBufferIntern : public HostBuffer<TYPE, DIM>
     {
         __startOperation(ITask::TASK_HOST);
 
-        if (pointer && ownPointer)
+        if (this->getPointer() && ownsPointer)
```
psychocoderHPC: This should be `this->getBasePointer()`

src/libPMacc/include/memory/buffers/HostBufferIntern.hpp
```diff
@@ -70,9 +76,9 @@ class HostBufferIntern : public HostBuffer<TYPE, DIM>
...
         {
-            CUDA_CHECK(cudaFreeHost(pointer));
+            CUDA_CHECK(cudaFreeHost(this->getPointer()));
```
psychocoderHPC: Should be `CUDA_CHECK(cudaFreeHost(this->getBasePointer()));`

src/libPMacc/include/memory/buffers/HostBufferIntern.hpp
```diff
@@ -70,9 +76,9 @@ class HostBufferIntern : public HostBuffer<TYPE, DIM>
...
         }
```
psychocoderHPC: Please set both pointer `pointer` and `basePointer` to `NULL` after the if condition, this helps to detect if already freed memory is used.

psychocoderHPC:
I add the flag **feature** and not **bug** to this pull request. The reason is that this was no error only a not optimal implementation.

erikzenker:
If the PR gets merged, then there might be problems with methods calling getBasePointer from HostBufferIntern. Because it might be possible that they wanted to call getPointer. 

I found one example for such a behaviour in HostBuffer when [cartBuffer()](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/libPMacc/include/memory/buffers/HostBuffer.hpp#L72-L86) is called. I think the implementer does not want to get the base pointer there.

src/libPMacc/include/memory/buffers/HostBufferIntern.hpp
erikzenker: fixed

src/libPMacc/include/memory/buffers/HostBufferIntern.hpp
erikzenker: fixed

src/libPMacc/include/memory/buffers/HostBufferIntern.hpp
erikzenker: fixed

psychocoderHPC:
We must fix the CuSTL parts within this pull request, else we creates a not working version of PMacc.

psychocoderHPC:
@Heikman please take a look which pointer is needed [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/libPMacc/include/memory/buffers/HostBuffer.hpp#L72-L86)

psychocoderHPC:
I opened an new issue for the wrong usage of `getPointer()` and the corresponding `getPitched()` for device buffer.
We should first solve #1283 and than merge this one.

Heikman:
cuSTL/HostBuffer needs the BasePointer because it owns it, therefore it releases it at destruction.

psychocoderHPC:
we should switch to #1283 for the `getPointer()` discussion

sbastrakov:
I believe this can be closed.

psychocoderHPC:
I think this issue is still not solved. But would also not start to fix it because @michaelsippel is refactoring this part anyway.

PR #1278, opened 2015-12-01
fix racecondition in `ConcatListOfFrames`

### LINKED ISSUE #1274: ADIOS plugin with moving window broken
The adios plugin crashes if moving window is enabled.
**Adios restart is not effected!**

Regression for #907 (not yet in HDF5)

reasons: 
- [assert](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/plugins/adios/WriteSpecies.hpp#L146) is triggert
- `globalParticleOffset` (bad naming ^^) is wrong (it looks like a OpenMP racecondition)

psychocoderHPC:
- close #1274
- explicit copy deep copy of the filter to each OpenMP thread

Tests:

- [x] LWFA with moving window and adios output

### DIFF
--- a/src/picongpu/include/plugins/kernel/CopySpecies.kernel
+++ b/src/picongpu/include/plugins/kernel/CopySpecies.kernel
@@ -55,7 +55,7 @@ struct ConcatListOfFrames
 
     /** concatenate list of frames to single frame
      *
-     * @param counter scalar offset in `destFrame`
+     * @param counter[in,out] scalar offset in `destFrame`
      * @param destFrame single frame were all particles are copied in
      * @param srcBox particle box were particles are read from
      * @param filter filter to select particles
@@ -63,7 +63,7 @@ struct ConcatListOfFrames
      * @param mapper mapper which describe the area were particles are copied from
      */
     template<class T_DestFrame, class T_SrcBox, class T_Filter, class T_Space, class T_Mapping>
-    void operator()(int& counter, T_DestFrame destFrame, T_SrcBox srcBox, T_Filter filter, T_Space particleOffset, T_Mapping mapper)
+    void operator()(int& counter, T_DestFrame destFrame, T_SrcBox srcBox, T_Filter filterWindowDomain, T_Space particleOffset, T_Mapping mapper)
     {
         #pragma omp parallel for
         for (int linearBlockIdx = 0;
@@ -71,6 +71,8 @@ struct ConcatListOfFrames
              ++linearBlockIdx
              )
         {
+            // local copy for each omp thread
+            T_Filter filter = filterWindowDomain;
             DataSpace<simDim> blockIdx(DataSpaceOperations<simDim>::map(m_gridDim, linearBlockIdx));
 
             using namespace PMacc::particles::operations;


psychocoderHPC:
ready for review

src/picongpu/include/plugins/kernel/CopySpecies.kernel
```diff
@@ -71,6 +71,7 @@ struct ConcatListOfFrames
              ++linearBlockIdx
              )
         {
+            T_Filter filter = filterWindowDomain;
```
erikzenker: You could add a comment e.g.: `// Local copy for each omp thread`

ax3l: yes, please.

and you are sure you can't/don't need to add more instructions in the `pragma omp parallel for` block?

ax3l: you should also declare the `counter` as `shared(counter)`

src/picongpu/include/plugins/kernel/CopySpecies.kernel
```diff
@@ -63,7 +63,7 @@ struct ConcatListOfFrames
      * @param mapper mapper which describe the area were particles are copied from
      */
     template<class T_DestFrame, class T_SrcBox, class T_Filter, class T_Space, class T_Mapping>
-    void operator()(int& counter, T_DestFrame destFrame, T_SrcBox srcBox, T_Filter filter, T_Space particleOffset, T_Mapping mapper)
+    void operator()(int& counter, T_DestFrame destFrame, T_SrcBox srcBox, T_Filter filterWindowDomain, T_Space particleOffset, T_Mapping mapper)
```
ax3l: why is `counter` given by ref, do you return it? (not documented in doxygen as output parameter)

src/picongpu/include/plugins/kernel/CopySpecies.kernel
psychocoderHPC: everything is by default shared in OpenMP if not `default(none)` is used.

@erikzenker yes I will add the comment

psychocoderHPC: @ax3l  
- openmp `private(var)` means that the variable is not initialized per thread.
- openmp  `firstprivate(var)` can used but it has some [restrictions](https://www.ualberta.ca/CNS/RESEARCH/NumStatsServers/Intel/Documentation/ssadiag_docs/index.htm#pt_reference/references/sc_omp_private_uninit.htm)

src/picongpu/include/plugins/kernel/CopySpecies.kernel
psychocoderHPC: yes it is an [inout] value

src/picongpu/include/plugins/kernel/CopySpecies.kernel
ax3l: I see, my OpenPM seems rusty.

psychocoderHPC:
updated

ax3l:
thank you, well done! :sparkles:

Issue #1279, opened 2015-12-01
Creation of very big PNG Images: Memory


ax3l:
Follow-up to #693.

The main problem currently is, that even if a user activates `scale_image` in `visualization.param` the following is done:

- original size image is created per rank
- reduce of full image
- slice of moving window from full image (copy)
- creation of full image in pngwriter (copy)
- scaling to `scale_image` in pngwriter (reduce)

which easily reaches the **available memory** of the *master node* in **2D3V** simulations (and large, many-GPU 3D3V simulations).

An alternative work-flow with low memory footprint would be:

- calculate local image (full size)
- communicate neighbors (2D in slice)
- interpolate (reduce)
- cut to moving window (can also be before interpolate, but here is fine too)
- gather reduced image to master

PR #1281, opened 2015-12-01
reduce visualization gpu memory footprint

psychocoderHPC:
allocate image memory only if gpu participate on the visualization

related to #1279

**Tests:**

- [x] 3D LWFA with sliding window (checked png images)

### DIFF
--- a/src/picongpu/include/plugins/output/images/Visualisation.hpp
+++ b/src/picongpu/include/plugins/output/images/Visualisation.hpp
@@ -451,7 +451,8 @@ public:
     m_slicePoint(slicePoint),
     isMaster(false),
     header(NULL),
-    reduce(1024)
+    reduce(1024),
+    img(NULL)
     {
         sliceDim = 0;
         if (m_transpose.x() == 0 || m_transpose.y() == 0)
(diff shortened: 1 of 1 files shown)


erikzenker:
The initialization of img has been moved to the [init()](https://github.com/psychocoderHPC/picongpu/blob/topic-reducePngPluginGpuMemoryFootprint/src/picongpu/include/plugins/output/images/Visualisation.hpp#L609-L637). What happends if the Visualisation class is used without calling init() ? I guess segmentation faults, since img was not initialized correctly. Should we add some bool isInitialized ?

ax3l:
it's not directly a plugin, but since it implements [ILightweightPlugin](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/Develop-Plugins) instead of `init` the methods `pluginLoad`/`pluginUnload` would be a way to implement that more transparent.

(Be aware that some kind of convention needs to be done since constructors and destructors are often too early/late when dealing with GPU memory. Nevertheless, in this case `Visualisation` is directly used in `LiveViewPlugin` and `PngPlugin` which might be in control of it's construction in their `plugin(Un)load` methods.)

erikzenker:
> Nevertheless, in this case Visualisation is directly used in LiveViewPlugin and PngPlugin which might be in control of it's construction in their plugin(Un)load methods.)

If it can be used wrong it will be used.

ax3l:
just have a short talk to @psychocoderHPC, maybe you find a saver way

Issue #532, opened 2014-08-11
Clean-up deletes, Ptrs to Null


elfring:
[An extra null pointer check is not needed in functions](http://dietmar-kuehl.de/mirror/c++-faq/freestore-mgmt.html#faq-16.8) like the following.
- [DataConnector](https://github.com/ComputationalRadiationPhysics/picongpu/blob/34abc9586421b1947ce4556c933642bbe059f55f/src/libPMacc/include/dataManagement/DataConnector.hpp#L201)
- [GatherSlice::finalize](https://github.com/ComputationalRadiationPhysics/picongpu/blob/34abc9586421b1947ce4556c933642bbe059f55f/src/libPMacc/examples/gameOfLife2D/include/GatherSlice.hpp#L76)
- [GatherSlice::reset](https://github.com/ComputationalRadiationPhysics/picongpu/blob/34abc9586421b1947ce4556c933642bbe059f55f/src/picongpu/include/plugins/output/GatherSlice.hpp#L197)
- [LineSliceFields::pluginUnload](https://github.com/ComputationalRadiationPhysics/picongpu/blob/34abc9586421b1947ce4556c933642bbe059f55f/src/picongpu/include/plugins/LineSliceFields.hpp#L230)
- [Radiation::pluginUnload](https://github.com/ComputationalRadiationPhysics/picongpu/blob/34abc9586421b1947ce4556c933642bbe059f55f/src/picongpu/include/plugins/radiation/Radiation.hpp#L272)
- [SumCurrents::pluginUnload](https://github.com/ComputationalRadiationPhysics/picongpu/blob/34abc9586421b1947ce4556c933642bbe059f55f/src/picongpu/include/plugins/SumCurrents.hpp#L178)

PR #1282, opened 2015-12-01
reduce `GatherSlice` host memory footprint

psychocoderHPC:
free mpi gatherv memory after usage

related to #1279

**Tests:**

- [x] 3D LWFA with sliding window (checked png images)
- 2D LWFA with sliding window 
  - [x] checked png images
  - [x] check memory usage with `top -d 0.5`

### DIFF
--- a/src/picongpu/include/plugins/output/GatherSlice.hpp
+++ b/src/picongpu/include/plugins/output/GatherSlice.hpp
@@ -188,6 +188,7 @@ struct GatherSlice
                 insertData(dstBox, srcBox, head->node.offset, head->node.maxSize);
             }
 
+            __deleteArray(fullData);
         }
 
         delete[] recvHeader;


src/picongpu/include/plugins/output/GatherSlice.hpp
```diff
@@ -188,6 +188,7 @@ struct GatherSlice
                 insertData(dstBox, srcBox, head->node.offset, head->node.maxSize);
             }
 
+            __deleteArray(fullData);
```
erikzenker: Why using __deleteArray when the rest of the code uses delete[] ?

psychocoderHPC: `__deleteArray()` is a PMacc macro which sets the pointer after it is deleted to `NULL`. This macro is newer than tthis class, therefore it is currently not used in the destructor or at other parts of this class.

erikzenker:
fullData does not need to be a class member if it is just used in the [operator()](https://github.com/psychocoderHPC/picongpu/blob/topic-reduceMemoryFootprintInGatherSlice/src/picongpu/include/plugins/output/GatherSlice.hpp#L116) methods. Or do you want to reuse it somewhere else, then you should ne delete it at the end. Furthermore, why not use the RAII principle of C++ and declare fullData as std::vector and leave the cleaup to the destructor ...just IHO as a c++ programmer.

src/picongpu/include/plugins/output/GatherSlice.hpp
erikzenker: I there some policy about replacing delete[] by __deleteArray when a file is touched ?

ax3l: not yet, but if you want you can create a PR.
also, the macros need a little update, see #532

PR #1287, opened 2015-12-03
Close #1262 Fix Larmor Power in FieldTmp

### LINKED ISSUE #1262: Is the Larmor Energy correctly calculated?
*This issue was requested by @psychocoderHPC in #1259.*

The Larmor energy calculation in [`LarmorEnergy.hpp` or similar](https://github.com/ComputationalRadiationPhysics/picongpu/blob/f31b31ab5afe8951ba58754fc6266a4ac07ffd20/src/picongpu/include/particles/particleToGrid/ComputeGridValuePerFrame.hpp#L122-L129) seems not correct.

Based on Jackson (14.26):

which is in cgs-units, we get the following SI-unit equation:

which is not the implemented equation.

Additionally, @Heikman and I are just discussing whether `m` is `m_0 * gamma` or just the rest mass `m_0` which regard to the radiation reaction in RLL (#1216) and photon generation.

PrometheusPi:
This pull request closes issue #1262. It fixes the calculation of the [Larmor energy](https://en.wikipedia.org/wiki/Larmor_formula). 

 - [x] check if the Larmor  energy of ions in KHI  is still negative

### DIFF
--- a/src/picongpu/include/particles/particleToGrid/derivedAttributes/LarmorEnergy.hpp
+++ b/src/picongpu/include/particles/particleToGrid/derivedAttributes/LarmorEnergy.hpp
@@ -55,13 +55,16 @@ namespace derivedAttributes
         const float_X gamma2 = gamma * gamma;
         const float_X c2 = SPEED_OF_LIGHT * SPEED_OF_LIGHT;
 
-        const float3_X mom_dt = mom - mom_mt1;
+        const float3_X mom_dt = (mom - mom_mt1) / float_X(DELTA_T);
         const float_X el_factor = charge * charge
             / (float_X(6.0) * PI * EPS0 *
-               c2 * c2 * SPEED_OF_LIGHT * mass * mass);
-        const float_X energyLarmor = el_factor * gamma2 * gamma2
-            * (math::abs2(mom_dt) -
-               math::abs2(math::cross(mom, mom_dt)));
+               c2 * SPEED_OF_LIGHT * mass * mass) * gamma2 * gamma2;
+        const float_X momentumToBetaConvert = float_X(1.0)/ (mass * SPEED_OF_LIGHT * gamma);
+        const float_X energyLarmor = el_factor
+                                     * ( math::abs2(mom_dt)
+                                         - momentumToBetaConvert * momentumToBetaConvert
+                                           * math::abs2(math::cross(mom, mom_dt))
+                                       );
 
         /* return attribute */
         return energyLarmor;
--- a/src/picongpu/include/particles/particleToGrid/derivedAttributes/LarmorPower.hpp
+++ b/src/picongpu/include/particles/particleToGrid/derivedAttributes/LarmorPower.hpp
@@ -20,7 +20,7 @@
 
 #pragma once
 
-#include "particles/particleToGrid/derivedAttributes/LarmorEnergy.def"
+#include "particles/particleToGrid/derivedAttributes/LarmorPower.def"
 
 #include "simulation_defines.hpp"
 
@@ -33,14 +33,14 @@ namespace derivedAttributes
 {
 
     HDINLINE float1_64
-    LarmorEnergy::getUnit() const
+    LarmorPower::getUnit() const
     {
         return UNIT_ENERGY;
     }
 
     template< class T_Particle >
     DINLINE float_X
-    LarmorEnergy::operator()( T_Particle& particle ) const
+    LarmorPower::operator()( T_Particle& particle ) const
     {
         /* read existing attributes */
         const float3_X mom = particle[momentum_];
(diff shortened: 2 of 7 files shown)


PrometheusPi:
Due to previously wrong second term (missing `1/(mass*c*gamma)²`) the second term was overestimated by ions in the KHI example. Now all energies are positive.

PrometheusPi:
This is a physics pull request. @ax3l or @Heikman please assign yourself.

PrometheusPi:
@ax3l should it not be called "Larmor Power" not "Larmor Energy" since it is a energy lost per time?

ax3l:
It *is* called "larmor energy"... [code](https://github.com/ComputationalRadiationPhysics/picongpu/blob/2d7a6d6adb6a6af2dbc6e4a8738ec195c74806eb/src/picongpu/include/particles/particleToGrid/derivedAttributes/LarmorEnergy.def#L41)

ax3l:
@PrometheusPi yes, in your current implementation it's now a *power*. you then need to adjust the unit and the unitDimension, too.

src/picongpu/include/particles/particleToGrid/derivedAttributes/LarmorEnergy.hpp
```diff
@@ -55,13 +55,16 @@ namespace derivedAttributes
...
             / (float_X(6.0) * PI * EPS0 *
-               c2 * c2 * SPEED_OF_LIGHT * mass * mass);
-        const float_X energyLarmor = el_factor * gamma2 * gamma2
-            * (math::abs2(mom_dt) -
-               math::abs2(math::cross(mom, mom_dt)));
+               c2 * SPEED_OF_LIGHT * mass * mass) * gamma2 * gamma2;
+        const float_X momentumToBetaConvert = float_X(1.0)/ (mass * SPEED_OF_LIGHT * gamma);
+        const float_X energyLarmor = el_factor
+                                     * ( math::abs2(mom_dt)
+                                         - momentumToBetaConvert * momentumToBetaConvert
```
ax3l: that will be very inprecise for relativistic particles.
can't your formulate it in terms of beta*gamma?

PrometheusPi:
@ax3l Yes, you were right, there was a "not" too much in the sentence. I fixed it. It is now a power and I will fix the naming too.

PrometheusPi:
@ax3l 
Somehow this comment got lost (or better speaking: I am not able to find it anymore):
>that will be very inprecise for relativistic particles.
>can't your formulate it in terms of beta*gamma?
The momentum used is `p = m*c*gamma*beta` thus the equation is already formulated in terms of `beta*gamma`.

PrometheusPi:
I updated the naming from `energy` to `power`.

ax3l:
Yes, idk out of hands if that was true.

src/picongpu/include/particles/particleToGrid/derivedAttributes/LarmorPower.hpp
```diff
@@ -55,16 +55,19 @@ namespace derivedAttributes
...
 
-        const float3_X mom_dt = mom - mom_mt1;
+        const float3_X mom_dt = (mom - mom_mt1) / float_X(DELTA_T);
         const float_X el_factor = charge * charge
             / (float_X(6.0) * PI * EPS0 *
-               c2 * c2 * SPEED_OF_LIGHT * mass * mass);
-        const float_X energyLarmor = el_factor * gamma2 * gamma2
-            * (math::abs2(mom_dt) -
-               math::abs2(math::cross(mom, mom_dt)));
+               c2 * SPEED_OF_LIGHT * mass * mass) * gamma2 * gamma2;
```
Heikman: I get: `charge * charge * charge * charge / (float_X(6.0) * PI * EPS0 * c2 * SPEED_OF_LIGHT * mass * mass) * gamma2`

src/picongpu/include/particles/particleToGrid/derivedAttributes/LarmorPower.hpp
```diff
@@ -55,16 +55,19 @@ namespace derivedAttributes
...
+        const float_X momentumToBetaConvert = float_X(1.0)/ (mass * SPEED_OF_LIGHT * gamma);
+        const float_X larmorPower = el_factor
+                                    * ( math::abs2(mom_dt)
+                                        - momentumToBetaConvert * momentumToBetaConvert
+                                          * math::abs2(math::cross(mom, mom_dt))
```
Heikman: This is not the part of `mom_dt` perpendicular to `mom`.

src/picongpu/include/particles/particleToGrid/derivedAttributes/LarmorPower.hpp
ax3l: in #1262 we thought it should be gamma^4 so right now `gamma2 * gamma2` is that implementation. is the description in #1262 wrong?

PrometheusPi: If you have a look at the [non-coveriant form of the Larmor power](https://en.wikipedia.org/wiki/Larmor_formula#Non-covariant_form) (Jackson 3rd Ed. Eq. 14.26) it goes with gamma^6 for beta^2. Here we use momentum^2 which is proportional to gamma^2, thus only gamma^4 remains in the prefactor.

@Heikman How do you derive an equation with gamma^2 ?

Heikman: It's because I am using gamma^2 for bremsstrahlung, but I am also assuming orthogonal acceleration and velocity and according to the wikipedia article, this yields gamma^2 and not gamma^4 in my case. So it should be correct here.

Issue #1035, opened 2015-08-11
PMath::MapTuple (and PMacc::Frame) violate boost::result_of protocol


Flamefire:
This issues is caused by the following:
The class PMacc::math::MapTuple defines a compiletime map with access via []-operator. To get the return value type boost::result_of is used. To make that work a member struct "result" is defined which returns the return type.

However this is in violation to the boost::result_of protocol that defines, that the result member struct has to return the return value of the ()-operator. See: http://www.boost.org/doc/libs/1_58_0/libs/utility/utility.htm#result_of 
This causes compilation to fail once the compiler supports "decltype", which is also included in C++11.

Possible solutions:

1. Globally define BOOST_RESULT_OF_USE_TR1 which always uses the member structs
2. Correctly use boost::result_of and therefore the ()-operator
3. Directly access the result member struct instead of the boost::result_of detour (and possibly rename it accordingly)
4. Use decltype (or boost equivalent) to get the return type of the []-operator directly

Note that without the BOOST_RESULT_OF_USE_TR1 (workaround-)define this is a showstopper for PMacc and therefor picongpu when compiling with C++11 enabled or where decltype is supported.

ax3l:
Closed with #1151

ax3l:
@Flamefire should we open a follow-up: refactoring PMacc to use result_of as a `C++11 feature` to use the c++11 `std::result_of`?

Flamefire:
Sounds reasonable. However it cannot be replaced 1:1 due to the use of boost::result_of for the []-operators etc. So this can only be replaced by C++11 features (decltype and std::result_of) once we switch to that.

ax3l:
ok, sounds good. can you please open an issue as `C++11 feature` for the refactoring (which can only be done when we drop support for C++98) and I will put it on future.

Flamefire:
--> #1163

PR #1289, opened 2015-12-04
CMake 3.1.0, Hypnos GCC 4.8.2 Chain

psychocoderHPC:
- #1035 added `CMAKE_CXX_STANDARD` that is only available in CMake >=3.1.0
- updated the minimum CMake version to 3.1.0
- update cmake version in `picongpu.profile.example` to available version >= 3.1.0
- **ATTENTION:** on the HZDR cluster `hypnos` the gcc version is changed (cmake is incompatible with gcc <4.8.2

- [x] new hypnos module combination needs a run/compile time test

**CMake release notes:** https://cmake.org/cmake/help/v3.1/release/3.1.0.html


ax3l:
if updates, also need update in the docs such as `doc/INSTALL.md`

ax3l:
compile suite updated to load `cmake/3.1.0`

psychocoderHPC:
updated:
- base CMakeLists.txt
- update documentation
- `picongpu.profile.example`s

Issue #1290, opened 2015-12-04
cuda 7.5 compile broken


psychocoderHPC:
Currently PIConGPU does not compile with `cuda 7.5`

**Error:**
```
.../fields/MaxwellSolver/None/NoSolver.hpp:51:76: error: ‘picongpu::SPEED_OF_LIGHT’ cannot appear in a constant-expression
.../fields/MaxwellSolver/None/NoSolver.hpp:51:93: error: ‘picongpu::SPEED_OF_LIGHT’ cannot appear in a constant-expression
.../fields/MaxwellSolver/None/NoSolver.hpp:51:111: error: ‘picongpu::DELTA_T’ cannot appear in a constant-expression
.../fields/MaxwellSolver/None/NoSolver.hpp:51:122: error: ‘picongpu::DELTA_T’ cannot appear in a constant-expression
.../fields/MaxwellSolver/None/NoSolver.hpp:51:133: error: ‘picongpu::INV_CELL2_SUM’ cannot appear in a constant-expression
.../fields/MaxwellSolver/Yee/YeeSolver.hpp: In member function ‘void picongpu::yeeSolver::YeeSolver<CurlE, CurlB>::updateE()’:
.../fields/MaxwellSolver/Yee/YeeSolver.hpp:68:76: error: ‘picongpu::SPEED_OF_LIGHT’ cannot appear in a constant-expression
.../fields/MaxwellSolver/Yee/YeeSolver.hpp:68:93: error: ‘picongpu::SPEED_OF_LIGHT’ cannot appear in a constant-expression
.../fields/MaxwellSolver/Yee/YeeSolver.hpp:68:111: error: ‘picongpu::DELTA_T’ cannot appear in a constant-expression
.../fields/MaxwellSolver/Yee/YeeSolver.hpp:68:122: error: ‘picongpu::DELTA_T’ cannot appear in a constant-expression
.../fields/MaxwellSolver/Yee/YeeSolver.hpp:68:133: error: ‘picongpu::INV_CELL2_SUM’ cannot appear in a constant-expression
.../fields/MaxwellSolver/DirSplitting/DirSplitting.hpp: At global scope:
.../fields/MaxwellSolver/DirSplitting/DirSplitting.hpp:62:77: error: ‘picongpu::SI::SPEED_OF_LIGHT_SI’ cannot appear in a constant-expression
[...]

PR #1291, opened 2015-12-06
Compile Suite: New CUDA 7.5 Chain

ax3l:
Updates the compile suite modules to load a `gcc/4.8.5` and `cuda/7.5.18` compile chain.

Requirement for C++98/C++11 compile tests.


psychocoderHPC:
We can not merge this before #1290 is solved else no branch compiles anymore.

ax3l:
The compile suite is updated manually, this PR is just reflecting the change that will come.

(That's why I am trying it on a Sunday to see what fails without affecting others ;) )

ax3l:
hm, boost/1.59.0 throws so many warnings ^^

but shows the same errors as posted.

ax3l:
I will close this for now, we have it for documentation.

PR #1294, opened 2015-12-09
ArgsParser: Help on Missing Params

ax3l:
Found by @erikzenker: if no parameters are set while calling `./picongpu` the *required* parameter `-g` for the grid size will also be missing. This should call `--help` as one would expect for zero arguments.

New behaviour:
```bash
./picongpu --help
# calls --help
# zero return code (correct usage)

./picongpu
# no cmd line arguments at all:
# calls --help
# non-zero return code (wrong usage)

./picongpu -d 1 1 1
# missing or malformed required cmd line arguments (-g)
#   Invalid or missing grid size.
#   use -g width height [depth=1]
# non-zero return code (wrong usage)
# (currently segfaults)
```

### DIFF
--- a/src/picongpu/ArgsParser.cpp
+++ b/src/picongpu/ArgsParser.cpp
@@ -32,8 +32,6 @@
 
 namespace picongpu
 {
-    namespace po = boost::program_options;
-
     ArgsParser::ArgsParser( )
     {
 
@@ -64,6 +62,8 @@ namespace picongpu
 
     ArgsParser::ArgsErrorCode ArgsParser::parse( int argc, char** argv )
     {
+        namespace po = boost::program_options;
+
         try
         {
             // add help message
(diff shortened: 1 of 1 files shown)


PrometheusPi:
I tested your pull request:

old code behaves **the same** as new code:
`./picongpu --help` and `./picongpu -h` -> prints help **and** simulation time :+1:
`./picongpu -d 1 1 1` -> crashes with a seg fault after printing:
```
Invalid or missing grid size.
use -g width height [depth=1]
```

**fixed:**
`./picongpu` now prints help :+1:

PrometheusPi:
@ax3l Is the seg fault the intended behavior in this case?

ax3l:
This specific case is also still old behaviour and caused by failing `pluginLoad()` calls (in this case [MySimulation::pluginLoad()](https://github.com/ComputationalRadiationPhysics/picongpu/blob/499fc5195e927ec53259f9a1ae661b67f842d682/src/picongpu/include/simulationControl/MySimulation.hpp#L150-L154)).

For this segfault to catch we need to implement a general handling in `libPMacc` that allows `pluginLoad()` to return non-success states that can then be used to stop further initialization, `unload()` already loaded plugins and then clean shut-down-ing the whole simulation.

PrometheusPi:
@ax3l You have a conflict in the [ArgsParser.cpp](https://github.com/ComputationalRadiationPhysics/picongpu/blob/d12a95c51094fc2a24c21dc1de278cbb3145a89a/src/picongpu/ArgsParser.cpp#L123) please rebase.

ax3l:
you are so fast.

PR #1295, opened 2015-12-09
Catch By Reference

ax3l:
C++ catch parameters should be passed by const reference.

Sorry for the cross-over change.

Did not see relevant implications right now, but should be fixed.

References:
  - http://en.cppreference.com/w/cpp/language/try_catch

### DIFF
--- a/src/tools/splash2txt/splash2txt.cpp
+++ b/src/tools/splash2txt/splash2txt.cpp
@@ -159,8 +159,9 @@ bool parseOptions( int argc, char** argv, ProgramOptions &options )
         }
 
     }
-    catch ( boost::program_options::error e )
+    catch ( const boost::program_options::error& )
     {
+        errorStream << desc << "\n";
         throw std::runtime_error( "Error parsing command line options!" );
     }
 
@@ -205,7 +206,7 @@ int main( int argc, char** argv )
     {
         parseSuccessfull = parseOptions( argc, argv, options );
     }
-    catch ( std::runtime_error e )
+    catch ( const std::runtime_error& e )
     {
         errorStream << "Error: " << e.what( ) << std::endl;
         mpi_finalize();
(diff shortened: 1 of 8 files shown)


PrometheusPi:
@ax3l You are author of the commit but did not update the license header.
Did you want to use `tool` as author or did you miss updating the header?

src/tools/splash2txt/splash2txt.cpp
```diff
@@ -159,8 +159,9 @@ bool parseOptions( int argc, char** argv, ProgramOptions &options )
         }
 
     }
-    catch ( boost::program_options::error e )
+    catch ( const boost::program_options::error& )
     {
+        errorStream << desc << "\n";
```
PrometheusPi: why did you add a error stream output here?
(not mentioned in the pull request description)

PrometheusPi:
@ax3l great job

src/tools/splash2txt/splash2txt.cpp
ax3l: oh I forgot that I added that. it's just printing "--help" before throwing the error that something in the input was parsed wrong.

ax3l:
I consider this change too trivial (1 line per file in avg) to be bent on being mentioned in the header (if not already) with an other line change per file.

I updated them manually to avoid missing or breaking some.

PrometheusPi:
There is a some cases of `catch` **without** `const` references remaining:
 - [`ArgsParser.cpp`](https://github.com/ComputationalRadiationPhysics/picongpu/blob/499fc5195e927ec53259f9a1ae661b67f842d682/src/picongpu/ArgsParser.cpp#L123) 
 - [`MySimulation.hpp`](https://github.com/ComputationalRadiationPhysics/picongpu/blob/499fc5195e927ec53259f9a1ae661b67f842d682/src/picongpu/include/simulationControl/MySimulation.hpp#L586) (is `const`, but syntax confuses)

ax3l:
yes I kept them since they are known to work. `const` would probably still be more excellent.

src/tools/splash2txt/splash2txt.cpp
PrometheusPi: :+1:

PR #1296, opened 2015-12-09
SimulationDescription: New Environment Singleton

ax3l:
The singleton `SimulationDescription` collects a general description of the simulation, such as maximum and current time step and the author that configures and controls it.

The later idea is, that general simulation controls such as the ones described in #19 can manipulate these (e.g., extend or shorten the maximum simulation step or "pause" after next iteration.) Additionally, it might allow to reduce some interfaces that do not depend on "a timestep" but on "the current time step".

Related to #1260

PR #1298, opened 2015-12-09
PNG Plugin: Write Simulation Author

ax3l:
Uses the new simulation author from ~~#1296~~ as meta information for the created png files.

### DIFF
--- a/src/picongpu/include/plugins/output/images/PngCreator.hpp
+++ b/src/picongpu/include/plugins/output/images/PngCreator.hpp
@@ -196,10 +196,10 @@ namespace picongpu
         header.writeToConsole( description );
 
         char title[] = "PIConGPU preview image";
-        char author[] = "The awesome PIConGPU-Team";
+        std::string author = Environment<>::get().SimulationDescription().getAuthor();
         char software[] = "PIConGPU with PNGwriter";
 
-        png.settext( title, author, description.str().c_str(), software);
+        png.settext( title, author.c_str(), description.str().c_str(), software);
 
         // write to disk and close object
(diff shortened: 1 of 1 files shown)


ax3l:
```bash
# ImageMagick
identify -verbose pngElectronsYX/e_png_yx_0.5_000000.png | grep "Author:"
```

Output:
- `--author` in PIConGPU not set: `Author: ` (empty string, the "Author: " prefix is from imagemagick)
- set: `Author: Your Name <yourMail>`

erikzenker:
Very useful feature, thank you very much !

ax3l:
uh, I just realized this change might destroy the "Creation Time" entry... need to check if that is our parameter or in PNGwriter
  https://github.com/pngwriter/pngwriter/issues/76

ax3l:
ok found it, problem in pngwriter

erikzenker:
Is there a pngwriter fix necessary which might raise the pngwriter version ?

ax3l:
yes, that is a tiny bug in `pngwriter 0.5.5` causing the malformed "Creation Time" attribute in some cases.

It effectively *reduces* the maximum length of the string to a random number between zero-size and actual-size.
Also, there is the slight possibility that [some random byte](https://github.com/pngwriter/pngwriter/blob/0.5.5/src/pngwriter.cc#L1084) behind a certain position (where [this](https://github.com/pngwriter/pngwriter/blob/0.5.5/src/pngwriter.cc#L1078) is stored) is set to `\0`, depending on what `sizeof()` from an undefined variable gives back...

PR #1301, opened 2015-12-10
Environment: Remove "Env" Prefix from `MemoryInfo`

ax3l:
Removes the odd "Env" prefix for the `MemoryInfo` singleton in our `Environment`.


ax3l:
@erikzenker pls have a look

erikzenker:
I can not spot any other occurrence of `EnvMemoryInfo`. Thank you, well done!

PR #1302, opened 2015-12-10
generic particle creation

Heikman:
This pull request adds a directory `particles/creation` for the creation of particles during runtime.
The created particle species is called `target species`, which is created out of `source species`. If needed, the E and B-Field can be interpolated at the source particle's position.
Most of the code is taken from the ionization module.

@ax3l @psychocoderHPC @erikzenker @PrometheusPi  who wants? :)

PR #1299, opened 2015-12-10
synchrotron functions

Heikman:
Implements synchroton functions via lookup table. 
The file is located in `particles/bremsstrahlung` because it is used by a follow-up pull request for the bremsstrahlung effect.
See [here](https://en.wikipedia.org/wiki/Synchrotron_function) for the definition.

PR #1303, opened 2015-12-10
Bremsstrahlung

Heikman:
Implement the bremstrahlung effect for electrons. It takes the numerical model from:
 > Gonoskov, A., et al. "Extended particle-in-cell schemes for physics in ultrastrong laser fields: Review and developments." Physical Review E 92.2 (2015): 023305.

**Dependencies:**
 - [ ] #1302 
 - [x] #1299

#### Synchrotron radiation in the classical regime

In a constant B-field the energy loss due to bremsstrahlung is: 
gamma_loss_dgl
Analytical solution:
gamma_loss_analy png
Simulation:
gamma_loss
Zoom:
gamma_loss_zoom
Photon spectrum:
photon_spec_t_0
Zoom:
photon_spec_t_0_zoom

**legend:**
green: analytical result
blue: simulation result, error bars: statistical error

### Known Issues
 - photons are created, but they don't move despite of the photon pusher.

### DIFF
--- a/src/picongpu/include/simulation_defines/param/speciesDefinition.param
+++ b/src/picongpu/include/simulation_defines/param/speciesDefinition.param
@@ -71,6 +71,29 @@ AttributRadiationFlag
 
 /*########################### define species #################################*/
 
+/*--------------------------- photons -------------------------------------------*/
+
+value_identifier(float_X, DummyMass, 1.0);
+value_identifier(float_X, DummyCharge, 0.0);
+
+typedef bmpl::vector<
+    particlePusher<particles::pusher::Photon>,
+    shape<UsedParticleShape>,
+    interpolation<UsedField2Particle>,
+    current<UsedParticleCurrentSolver>,
+    massRatio<DummyMass>,
+    chargeRatio<DummyCharge>
+> ParticleFlagsPhotons;
+
+/*define specie ions*/
+typedef Particles<
+    ParticleDescription<
+        bmpl::string<'p'>,
+        SuperCellSize,
+        DefaultParticleAttributes,
+        ParticleFlagsPhotons
+    >
+> PIC_Photons;
 
 /*--------------------------- electrons --------------------------------------*/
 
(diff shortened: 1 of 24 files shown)


ax3l:
sweet! :sparkles:
what do the error bars in your plots stand for and which is what (green/blue)?

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -71,6 +71,29 @@ AttributRadiationFlag
 
 /*########################### define species #################################*/
 
+/*--------------------------- photons -------------------------------------------*/
+
+value_identifier(float_X, DummyMass, 1.0);
```
ax3l: what is the technical problem again with setting mass to zero or not defining it at all?

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -71,6 +71,29 @@ AttributRadiationFlag
...
+value_identifier(float_X, DummyCharge, 0.0);
```
ax3l: it's actually the charge that is zero, not a dummy mass :)

src/picongpu/include/simulation_defines/param/speciesDefinition.param
ax3l: pls call them as usual, they are actually zero and no dummy values.

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -71,6 +71,29 @@ AttributRadiationFlag
...
+    interpolation<UsedField2Particle>,
+    current<UsedParticleCurrentSolver>,
+    massRatio<DummyMass>,
+    chargeRatio<DummyCharge>
+> ParticleFlagsPhotons;
+
+/*define specie ions*/
+typedef Particles<
+    ParticleDescription<
+        bmpl::string<'p'>,
```
ax3l: Protons would disagree ;)

Heikman:
This PR is aborted because of having too much trouble with micromanagement. Currently the code from this PR is under reconstruction and will be published soon.

PR #1305, opened 2015-12-11
SimulationHelper `softRestarts`: split `init` and add `resetAll`

ax3l:
This is a port of the `--softRestarts` option added to libPMacc `SimulationHelper`. It can be used to loop whole simulation for presentations, etc. by splitting `init()`:

- `init()`: allocate memory
- `fillSimulation()`: fill allocated memory with data (from checkpoint or t=0)
- `resetAll()`: reset to state after `init()` so `fillSimulation()`
                can be called again (does *not* deallocate memory)

We used it first in our 3D Live visualization and this PR ports it to the mainline.

Unfurtunately, we have to update PIConGPU examples ~~and GoL~~ at the same time since the `SimulationHelper` interface change requires it. The individual changes are in separate commits in this PR.

### To Do
- ~~update~~ GoL example: ~~add `resetAll` and `fillSimulation`~~ does not use `SimulationHelper` and does therefore not support `--softRestarts`
- [x] add `--softRestarts <N>` to `doc/TBG_macros.cfg`
- [x] SingleParticle tests and `ThermalTest` ~~might [not compile](https://github.swim-results.de/?status=$1$PblFPFnQ$QVjQDqvRgn3.4ujB0qOaP0) atm~~ updated
- [x] runtime test again (PIConGPU)

### DIFF
--- a/examples/SingleParticleCurrent/include/OneParticleSimulation.hpp
+++ b/examples/SingleParticleCurrent/include/OneParticleSimulation.hpp
@@ -64,7 +64,7 @@ public:
     {
     }
 
-    virtual uint32_t init()
+    virtual void init()
     {
 
         MySimulation::init();
@@ -80,11 +80,15 @@ public:
 #endif
 
         }
+    }
+
+    virtual uint32_t fillSimulation()
+    {
+        MySimulation::fillSimulation();
 
         const SubGrid<simDim>& subGrid = Environment<simDim>::get().SubGrid();
 
         const DataSpace<simDim> halfSimSize(subGrid.getGlobalDomain().size / 2);
-
 
         GridLayout<simDim> layout(subGrid.getLocalDomain().size, MappingDesc::SuperCellSize::toRT());
         MappingDesc cellDescription = MappingDesc(layout.getDataSpace(), GUARD_SIZE, GUARD_SIZE);
@@ -112,7 +116,6 @@ public:
 
 
         return 0;
-
     }
 
     /**
--- a/src/libPMacc/include/simulationControl/SimulationHelper.hpp
+++ b/src/libPMacc/include/simulationControl/SimulationHelper.hpp
@@ -287,15 +287,16 @@ public:
                    (int) (tSimCalculation.getInterval() / 1000.) << " sec" << std::endl;
             }
 
-        } // softRestart loop
+        } // softRestarts loop
     }
 
     virtual void pluginRegisterHelp(po::options_description& desc)
     {
         desc.add_options()
             ("steps,s", po::value<uint32_t > (&runSteps), "Simulation steps")
-            ("softRestart", po::value<uint32_t > (&softRestarts)->default_value(0),
-             "Number of times to restart the simulation after simulation has finished (for presentations)")
+            ("softRestarts", po::value<uint32_t > (&softRestarts)->default_value(0),
+             "Number of times to restart the simulation after simulation has finished (for presentations). "
+             "Note: does not yet work with all plugins, see issue #1305")
             ("percent,p", po::value<uint16_t > (&progress)->default_value(5),
              "Print time statistics after p percent to stdout")
             ("restart", po::value<bool>(&restartRequested)->zero_tokens(), "Restart simulation")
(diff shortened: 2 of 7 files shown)


examples/SingleParticleCurrent/include/OneParticleSimulation.hpp
```diff
@@ -64,7 +64,7 @@ class OneParticleSimulation : public MySimulation
     {
     }
 
-    virtual uint32_t init()
+    virtual void init()
```
erikzenker: I`am not sure if you can change the return type to void here (See [SO](http://stackoverflow.com/questions/4665117/c-virtual-function-return-type) post.

What about `return MySimulation::init()` ?

src/libPMacc/include/simulationControl/SimulationHelper.hpp
```diff
@@ -95,12 +95,27 @@ class SimulationHelper : public IPlugin
     virtual void runOneStep(uint32_t currentStep) = 0;
 
     /**
-     * Initializes simulation state.
+     * Initialize simulation
+     *
+     * Does hardware selections/reservations, memory allocations and
+     * initializes data structures as empty.
+     */
+    virtual void init() = 0;
```
erikzenker: Ahh well you change the signature of init here too ...

src/libPMacc/include/simulationControl/SimulationHelper.hpp
```diff
@@ -198,80 +213,89 @@ class SimulationHelper : public IPlugin
      */
     void startSimulation()
     {
-        uint32_t currentStep = init();
-        Environment<>::get().SimulationDescription().setCurrentStep( currentStep );
+        init();
 
-        tInit.toggleEnd();
-        if (output)
+        for (uint32_t softRestartLoop = 0; softRestartLoop <= softRestarts; ++softRestartLoop)
```
erikzenker: Could you pls provide some more meaningful name for the loop index e.g. nthSoftRestart or softRestart_i

examples/SingleParticleCurrent/include/OneParticleSimulation.hpp
ax3l: written below: signature changed in the base class.

src/libPMacc/include/simulationControl/SimulationHelper.hpp
ax3l: sure, np

theZiz:
I did pull the commits for my isaac branch, but it does only work with moving window __disabled__. With moving window the E, B and J field are just empty and the Electron density field stays between 0 and -0,67.

psychocoderHPC:
The moving window reset is missing. Sry thats my fault that it is no clean pull inside the insitu repo.

Please add this to you pull request https://github.com/ComputationalRadiationPhysics/insituvolvis/pull/23/files#diff-27ab651b1e352e1783fb08c3aed891c9R325

ax3l:
@theZiz I rebased the branch against `dev` and added the support for moving window from our nsituvolvis repo.

ax3l:
@erikzenker when the with/without moving window RT tests of @theZiz are successful you can perform your review -> merge powers.

erikzenker:
:carousel_horse:

theZiz:
Läuft.

ax3l:
excellent, thx.

psychocoderHPC:
Please keep in mind that all plugins which accumulate data in an own buffer over time can not used with `--softRestart`.
Our plugins has no `reset()` or `initialFill()` method.

theZiz:
Maybe this should be mentioned in the parameter help of --softRestart? I doubt anyone looks in github pull request comments before using a feature ;)

ax3l:
-> affects only `Radiation` so far, documented in #1331

ax3l:
note: renamed in user interface to `--softRestarts` for consistency and clarity that this is not a true/false but a counter flag.

ax3l:
@erikzenker @psychocoderHPC is you want we can merge this now.
I will post a follow-up that will do some renamings:

 -  `init()` -> `allocate()`: allocate memory
 - `fill*` -> `initialize()`: fill allocated memory with data, e.g., add initial particles to a frame list
 - `reset*` -> `clear()`: reset to state after `allocate()` so `initialize()` or `restart()` can be called again (does not deallocate memory but might for example drop frames with particles)

erikzenker:
Thx to @ax3l and @theZiz to make this PR happen ...greetings from the space office :rocket:

erikzenker:
Review finished --> merge !

PR #1309, opened 2015-12-15
let cuSTL/kernel/runtime/Foreach compute best BlockDim

Heikman:
If no BlockDim is given `Foreach` computes it by itself.

Solves #1256

Tested with SliceFieldPrinter plugin.

### DIFF
--- a/src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
+++ b/src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
@@ -22,8 +22,9 @@
 
 #pragma once
 
-
 #include "types.h"
+#include "cudaSpecs.hpp"
+#include "static_assert.hpp"
 #include "math/vector/Size_t.hpp"
 #include "math/vector/Int.hpp"
 #include "lambda/make_Functor.hpp"
@@ -37,6 +38,7 @@
 #include <boost/preprocessor/arithmetic/inc.hpp>
 #include <boost/preprocessor/repetition/repeat.hpp>
 #include <boost/preprocessor/repetition/repeat_from_to.hpp>
+#include <boost/math/common_factor.hpp>
 
 #include "eventSystem/tasks/TaskKernel.hpp"
 #include "eventSystem/events/kernelEvents.hpp"
@@ -51,6 +53,64 @@ namespace kernel
 {
 namespace RT
 {
+
+/** Heuristic maximum threads per block and per axis
+ * in agreement to sm_2.x - sm_5.3
+ *
+ * These values don't fully exploit the limits from the cuda specification
+ * but they give reasonable speed.
+ */
+template<int dim>
+struct MaxCudaBlockDim;
+
+template<>
+struct MaxCudaBlockDim<DIM1>
+{
+    typedef math::CT::Size_t<1024, 1, 1> type;
+};
+
+template<>
+struct MaxCudaBlockDim<DIM2>
+{
+    typedef math::CT::Size_t<32, 32, 1> type;
+};
+
+template<>
+struct MaxCudaBlockDim<DIM3>
+{
+    typedef math::CT::Size_t<8, 8, 8> type;
+};
+
+/* Check if MaxCudaBlockDim holds the cuda specification limits */
+PMACC_CASSERT_MSG(_cuda_blockDim_exceeds_maximum_number_of_threads_per_block,
+    math::CT::volume<typename MaxCudaBlockDim<DIM1>::type >::type::value <= cudaSpecs::maxNumThreadsPerBlock);
+PMACC_CASSERT_MSG(_cuda_blockDim_exceeds_maximum_number_of_threads_per_block,
+    math::CT::volume<typename MaxCudaBlockDim<DIM2>::type >::type::value <= cudaSpecs::maxNumThreadsPerBlock);
+PMACC_CASSERT_MSG(_cuda_blockDim_exceeds_maximum_number_of_threads_per_block,
+    math::CT::volume<typename MaxCudaBlockDim<DIM3>::type >::type::value <= cudaSpecs::maxNumThreadsPerBlock);
+
+/** Return a suitable cuda blockDim for a given gridDim.
+ *
+ * @param gridDim 1D, 2D or 3D grid size
+ * @return cuda blockDim
+ */
+template<int dim>
+math::Size_t<DIM3> getBestCudaBlockDim(const math::Size_t<dim> gridDim)
+{
+    math::Size_t<DIM3> result = math::Size_t<DIM3>::create(1);
+
+    /* The greatest common divisor of each component of the volume size
+     * and a certain power of two value yield the best suitable block size */
+    const boost::math::gcd_evaluator<size_t> gcd; /* greatest common divisor */
+    const math::Size_t<DIM3> maxThreads =
+        MaxCudaBlockDim<dim>::type::toRT(); /* max threads per axis */
+    for(int i = 0; i < dim; i++)
+    {
+        result[i] = gcd(gridDim[i], maxThreads[i]);
+    }
+
+    return result;
+}
 
 #ifndef FOREACH_KERNEL_MAX_PARAMS
 #define FOREACH_KERNEL_MAX_PARAMS 4
@@ -68,11 +128,14 @@ namespace RT
         /* C0 c0_shifted = c0(p_zone.offset); ...; CN cN_shifted = cN(p_zone.offset); */                              \
         BOOST_PP_REPEAT(N, SHIFT_CURSOR_ZONE, _)                                                                    \
                                                                                                                     \
-        /* the maximum number of threads per block for devices with                                                 \
-         * compute capability > 2.0 is 1024 */                                                                      \
-        assert(this->_blockDim.productOfComponents()<=1024);                                                        \
-        /* the maximum block size in z direction is 64 for all compute capabilities */                              \
-        assert(this->_blockDim.z()<=64);                                                                            \
+        if(this->_blockDim == math::Size_t<DIM3>::create(0))                                                        \
+            this->_blockDim = getBestCudaBlockDim(p_zone.size);                                                     \
+                                                                                                                    \
+        assert(this->_blockDim.productOfComponents() <= cudaSpecs::maxNumThreadsPerBlock);                          \
+        assert(this->_blockDim.x() <= cudaSpecs::MaxNumThreadsPerBlockDim::x::value);                               \
+        assert(this->_blockDim.y() <= cudaSpecs::MaxNumThreadsPerBlockDim::y::value);                               \
+        assert(this->_blockDim.z() <= cudaSpecs::MaxNumThreadsPerBlockDim::z::value);                               \
+                                                                                                                    \
         dim3 blockDim(this->_blockDim.x(), this->_blockDim.y(), this->_blockDim.z());                               \
         kernel::detail::SphericMapper<Zone::dim> mapper;                                                            \
         using namespace PMacc;                                                                                      \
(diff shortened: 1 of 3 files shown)


src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
```diff
@@ -68,6 +69,18 @@ namespace RT
         /* C0 c0_shifted = c0(p_zone.offset); ...; CN cN_shifted = cN(p_zone.offset); */                              \
         BOOST_PP_REPEAT(N, SHIFT_CURSOR_ZONE, _)                                                                    \
                                                                                                                     \
+        if(this->_blockDim == math::Size_t<3>::create(0))                                                           \
```
psychocoderHPC: offline discussed: thus should be moved in a helper function, there is no need to write it in a macro.

psychocoderHPC:
Please keep in mind that it is not easy to calculate the best block size for a generic functor if you not know how many static shared memory is used intern.
NVIDIA created for this an API.

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
```diff
@@ -68,6 +69,18 @@ namespace RT
...
+        {                                                                                                           \
+            this->_blockDim = math::Size_t<3>::create(1);                                                           \
```
ax3l: pls use `DIM3` here and above

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
Heikman: see one line below.

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
```diff
@@ -68,6 +69,18 @@ namespace RT
...
+            /* The greatest common divisor of each component of the volume size                                         \
+             * and a certain power of two value gives the best suitable block size */                                   \
+            boost::math::gcd_evaluator<size_t> gcd; /* greatest common divisor */                                       \
+            size_t maxValues[] = {16, 16, 4}; /* maximum values for each axis */                                      \
```
ax3l: can we define this constant somewhere more central?

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
```diff
@@ -52,6 +53,28 @@ namespace kernel
...
 
+/** Return a suitable cuda blockDim for a given gridDim.
+ *
+ * @param gridDim 1D, 2D or 3D grid size
+ * @return cuda blockDim
+ */
+template<int dim>
+math::Size_t<3> getBestCudaBlockDim(math::Size_t<dim> gridDim)
+{
+    math::Size_t<3> result = math::Size_t<3>::create(1);
```
ax3l: pls use `DIM3`

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
```diff
@@ -52,6 +53,28 @@ namespace kernel
 namespace RT
 {
 
+/** Return a suitable cuda blockDim for a given gridDim.
+ *
+ * @param gridDim 1D, 2D or 3D grid size
+ * @return cuda blockDim
+ */
+template<int dim>
+math::Size_t<3> getBestCudaBlockDim(math::Size_t<dim> gridDim)
```
ax3l: pls use `DIM3` instead of `3` here

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
```diff
@@ -52,6 +53,28 @@ namespace kernel
...
+{
+    math::Size_t<3> result = math::Size_t<3>::create(1);
+
+    /* The greatest common divisor of each component of the volume size
+     * and a certain power of two value gives the best suitable block size */
+    boost::math::gcd_evaluator<size_t> gcd; /* greatest common divisor */
+    size_t maxValues[] = {16, 16, 4}; /* maximum values for each axis */
```
ax3l: can we define this a bit more central and add a reasoning why we don't want to use more than 1024 threads? (e.g., what does sm_20+ allow)

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
```diff
@@ -37,6 +37,7 @@
 #include <boost/preprocessor/arithmetic/inc.hpp>
 #include <boost/preprocessor/repetition/repeat.hpp>
 #include <boost/preprocessor/repetition/repeat_from_to.hpp>
+#include <boost/math/common_factor.hpp>
```
ax3l: ok, header only https://gist.github.com/ax3l/7450482

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
ax3l: `DIM3`

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
```diff
@@ -97,7 +122,7 @@ struct Foreach
      * E.g. (8,8,4) fits into (256, 256, 256)
      *
      */
-    Foreach(math::Size_t<3> _blockDim) : _blockDim(_blockDim) {}
+    Foreach(math::Size_t<3> _blockDim = math::Size_t<3>::create(0)) : _blockDim(_blockDim) {}
```
ax3l: `DIM3` pls

Heikman:
@psychocoderHPC shared memory cannot be used per thread anyway if the blockDim is not know at compile-time.

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
Heikman: I wouldn't worry about this here so much because firstly it's a very rarely used function and secondly tuning `maxValues` is just an optimization.

ax3l: > it's a very rarely used function

the first point is not an argument that I can accept if I use cuSTL :)

> tuning maxValues is just an optimization

the second  one is not only an optimization: a new integrated SoC might for example not support so many threads or a port to alpaka would also like to limit the max threads to fit on a multi-core CPU.

the maximum number of threads is a global device attribute and should be adjustable centrally in either cuSTL or libPMacc.

ax3l: (I am not saying we need immediately a user-tunable, compile-time API for that. but for the start, let us at least collect such constants in a common, in-code, libPMacc "config" file.)

Heikman:
I cannot add an include for `cudaSpecs.hpp` because of header dependencies. So one has to include it just when needed.

src/libPMacc/include/cudaSpecs.hpp
```diff
@@ -0,0 +1,57 @@
...
+{
+
+/** Suitable maximum threads per block and per axis
+ * according to sm_2.x - sm_5.3
+ */
+template<int dim>
+struct MaxCudaBlockDim;
+
+template<>
+struct MaxCudaBlockDim<1>
```
psychocoderHPC: Please use the maximum values out of the programming guide. (Cuda Programming Guide Apendix G.)

```
BOOST_CONSTEXPR_OR_CONST maxNumThreadsPerBlock = 1024;
math::CT::Size_t<1024, 124, 64> MaxNumThreadsPerBlockDim;
```

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
```diff
@@ -68,6 +92,8 @@ namespace RT
         /* C0 c0_shifted = c0(p_zone.offset); ...; CN cN_shifted = cN(p_zone.offset); */                              \
         BOOST_PP_REPEAT(N, SHIFT_CURSOR_ZONE, _)                                                                    \
                                                                                                                     \
+        if(this->_blockDim == math::Size_t<DIM3>::create(0))                                                           \
```
psychocoderHPC: the `\` at the end is not alligned

psychocoderHPC: Is the `getBestCudaBlockDim` method only called if the user defines the block size as `(0,0,0)`? 
If so please add somewhere a documentation that the user knows about this feature.

src/libPMacc/include/cudaSpecs.hpp
```diff
@@ -0,0 +1,37 @@
...
+#pragma once
+
+#include "math/vector/compile-time/Size_t.hpp"
+#include <stdint.h>
+
+namespace PMacc
+{
+namespace cudaSpecs
+{
+
```
psychocoderHPC: Please add a small documentation:
- what is the meaning of the value(s)
- add a reference where the magic numbers are from

psychocoderHPC:
please rebase and add a small documentation to the annotated part.

Heikman:
done.

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
```diff
@@ -68,6 +128,8 @@ namespace RT
         /* C0 c0_shifted = c0(p_zone.offset); ...; CN cN_shifted = cN(p_zone.offset); */                              \
         BOOST_PP_REPEAT(N, SHIFT_CURSOR_ZONE, _)                                                                    \
                                                                                                                     \
+        if(this->_blockDim == math::Size_t<DIM3>::create(0))                                                        \
+            this->_blockDim = getBestCudaBlockDim(p_zone.size);                                                     \
         /* the maximum number of threads per block for devices with                                                 \
          * compute capability > 2.0 is 1024 */                                                                      \
         assert(this->_blockDim.productOfComponents()<=1024);                                                        \
```
ax3l: use the new `cudaSpec` limits here :)

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
```diff
@@ -89,15 +151,17 @@ namespace RT
...
     /* \param _blockDim size of the cuda blockDim.
      *
      * blockDim has to fit into the computing volume.
      * E.g. (8,8,4) fits into (256, 256, 256)
      *
+     * If no argument is given, the blockDim will be computed heuristically.
+     *
      */
-    Foreach(math::Size_t<3> _blockDim) : _blockDim(_blockDim) {}
+    Foreach(math::Size_t<DIM3> _blockDim = math::Size_t<DIM3>::create(0)) : _blockDim(_blockDim) {}
```
ax3l: can't we write `= getBestCudaBlockDim(p_zone.size);` here as default?

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
ax3l: two lines below, too! (`64`)

src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
Heikman: we don't know the zone before `operator()`.

ax3l: true! :D

src/picongpu/include/plugins/SliceFieldPrinter.tpp
```diff
@@ -149,7 +149,7 @@ void SliceFieldPrinter<Field>::printSlice(const TField& field, int nAxis, float
 
     /* convert data to higher precision and to SI units */
     SliceFieldPrinterHelper::ConversionFunctor<Field> cf;
```
psychocoderHPC: Could we remove the `ifdef`'s? Maybe in a new pull request.

```C++
    vec::UInt32<simDim> twistedAxesVec;
   for( uint32_t d = 0; d < simDim; ++d )
   {
        twistedAxesVec[d] = (nAxis + d)%simDim;
   }
   /* convert data to higher precision and to SI units */
   SliceFieldPrinterHelper::ConversionFunctor<Field> cf;
   PMACC_AUTO( lastPlainShift, vec::UInt32<simDim>::create(0) );
   lastPlainShift[simDim-1] = localPlane;
  algorithm::kernel::RT::Foreach()(
        dBuffer_SI->zone(), dBuffer_SI->origin(),
        cursor::tools::slice(field.originCustomAxes(twistedAxesVec)(lastPlainShift)),
        cf );
```

Heikman: We should keep this for a follow-up PR.

psychocoderHPC:
Thank for this fix

PR #1232, opened 2015-11-09
Doc Laser Waist: GaussianBeam, PulseFrontTilt

ax3l:
This ~~changes~~ adds additional information to the the doc-string for the [beam waist](https://github.com/ComputationalRadiationPhysics/picongpu/blob/71af8e8bf07d9fb36c33df561cc1f796133fc16f/src/picongpu/include/fields/laserProfiles/laserGaussianBeam.hpp#L138) W0 of the Gaussian Beam and the tilted version of it for laser implementations.
    
~~Before that, the in-focus intensity was still correct but the waist would have been to small if one was following the documentation. This is an old documentation bug that someone changed in pre-review (svn) times that slipped through.~~ The very short documentation before caused some confusion when trying to compare to experiments (for me :) ).

~~Users: if you are using the GaussianBeam profile in your simulation, please cross-check if the beam waist was set to what you intended it to be. It might be too small (resulting in a still correct in-focus intensity but smaller spot size and steeper divergence).~~

laserwaist4

(image: [v1](https://cloud.githubusercontent.com/assets/1353258/11030380/f1ad804a-86cd-11e5-94c2-334cf050fb66.png), ~~[v2](https://cloud.githubusercontent.com/assets/1353258/11211733/c1aca8f2-8d31-11e5-8453-eea6bcfa5df9.png)~~, [v3](https://cloud.githubusercontent.com/assets/1353258/11213215/0b601c64-8d3b-11e5-88b6-a1ba032568c5.png), [v4](https://cloud.githubusercontent.com/assets/1353258/11837505/af2eecfa-a3e1-11e5-8b03-d81be6d91203.png))

### Tasks

- [x] update examples (values kept, default LWFA slight increase for consistency)
- [x] check [wavepacket definition](https://github.com/ComputationalRadiationPhysics/picongpu/blob/71af8e8bf07d9fb36c33df561cc1f796133fc16f/src/picongpu/include/fields/laserProfiles/laserWavepacket.hpp#L113-L116) -> wrong -> separate PR
- ~~[LaserPolynom](https://github.com/ComputationalRadiationPhysics/picongpu/blob/71af8e8bf07d9fb36c33df561cc1f796133fc16f/src/picongpu/include/fields/laserProfiles/laserPolynom.hpp#L77)~~
- ~~write announcement on mailing list (after review)~~

PR #1312, opened 2015-12-17
Laser Wavepacket: Implementation != Doc

ax3l:
Follow-Up to #1232: the implementation of the laser wavepacket regarding the *beam waist* does not fit its [documentation in `laserConfig.param`](https://github.com/ComputationalRadiationPhysics/picongpu/blob/7bf113456bf64d82b6571171fc039c1f3fba929c/src/picongpu/include/simulation_defines/param/laserConfig.param#L227-L234).

This commit updates the implementation to match the documentation which is then identical to the other laser pulses. Currently, the parameter that was specified is a factor `sqrt(2)` off making the pulse waist effectively *tighter* by this factor than intended.

### To Do
- [ ] announcement of mailing list (after review)
- [x] add some documentation as in #1232

### DIFF
--- a/src/picongpu/include/fields/laserProfiles/laserWavepacket.hpp
+++ b/src/picongpu/include/fields/laserProfiles/laserWavepacket.hpp
@@ -113,7 +113,7 @@ HDINLINE float3_X laserTransversal(float3_X elong, const float_X, const float_X 
     const float_X exp_x = posX * posX / (W0_X * W0_X);
     const float_X exp_z = posZ * posZ / (W0_Z * W0_Z);
 
-    return elong * math::exp(float_X(-0.5) * (exp_x + exp_z));
+    return elong * math::exp(exp_x + exp_z);
 
 }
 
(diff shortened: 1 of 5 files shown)


PrometheusPi:
looks good to me :+1:
The documentation is correct and I think adding FWHM Intensity conversion factors is not needed for this "test" laser. 
Could be merged this way.

ax3l:
ah, same doc for all lasers to be explicit :)
added in 2nd commit. can be merged if it looks ok to you.

PrometheusPi:
:+1: will merge

PrometheusPi:
@ax3l thx

ax3l:
np, thx

ax3l:
@PrometheusPi can we merge? :) (I already have the mail open for hours for the user list ;) )

PrometheusPi:
there was a minus sign removed in `laserWavepacket` see pull request #1722 for the fix.

ax3l:
awww. good catch!

src/picongpu/include/fields/laserProfiles/laserWavepacket.hpp
```diff
@@ -113,7 +113,7 @@ HDINLINE float3_X laserTransversal(float3_X elong, const float_X, const float_X
     const float_X exp_x = posX * posX / (W0_X * W0_X);
     const float_X exp_z = posZ * posZ / (W0_Z * W0_Z);
 
-    return elong * math::exp(float_X(-0.5) * (exp_x + exp_z));
+    return elong * math::exp(exp_x + exp_z);
```
ax3l: Gosh! Arrrrr! :see_no_evil:

PR #1316, opened 2016-01-04
Taurus-TUD: K20X and K80 Queues

PrometheusPi:
This pull requests avoids mixing of two gpu types (k20x and k80) on taurus/ZIH/TU Dresden.
The old k20x are back and are also available in the `gpu` queue. To only acces k80 one needs to specify `gpu2` as queue. Additionly the k20x are available via queue `gpu1`.

This pull requetst:
 - add k20x queue (gpu1)
 - specify k80 queue to only run on gpu2

see:
https://doc.zih.tu-dresden.de/hpc-wiki/bin/view/Compendium/SystemTaurus#Partitions
for more details on the current setup of the taurus cluster

### DIFF
--- a/src/picongpu/submit/taurus-tud/k20x_profile.tpl
+++ b/src/picongpu/submit/taurus-tud/k20x_profile.tpl
@@ -19,19 +19,19 @@
 #
 
 ## calculations will be performed by tbg ##
-TBG_queue="gpu2"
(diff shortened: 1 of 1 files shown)


PrometheusPi:
@ax3l thanks for updating/fixing the title.

ax3l:
thx, looks correct to me!

PR #1323, opened 2016-01-08
Plugins: Common String Helpers (Splash)

ax3l:
Adds ArrayOfString helpers for libSplash write calls.
See
  https://github.com/ComputationalRadiationPhysics/libSplash/issues/220
for the description of the problem.

If a user wants to write a list of strings into a single HDF5 attribute or date they need to be padded to the longest one, `\0` separated and concatenated. This helper does that.

Use it like this:
```C++
#include "plugins/common/stringHelpers.hpp"
// [...]

// create some strings
std::list<std::string> myListOfStr;
myListOfStr.push_back("short");
myListOfStr.push_back("middle");
myListOfStr.push_back("loooong");

// convert to splash format
helper::GetSplashArrayOfString getSplashArrayOfString;
helper::GetSplashArrayOfString::Result myArrOfStr;
myArrOfStr = getSplashArrayOfString( myListOfStr );
ColTypeString ctSomeListOfStr( myArrOfStr.maxLen );

// splash call
dc->writeGlobalAttribute(
    threadParams->currentStep,
    ctSomeListOfStr,
    "someListOfStr",
    1u, /* ndims: 1D array */
    Dimensions( myListOfStr.size(), 0, 0 ), /* size of 1D array */
    &( myArrOfStr.buffers.at(0) )
);
```
(in C++11 this creation + conversation will be a two-liner.)

### Misc

- Also adds a `helper::` namespace to avoid polluting the `PIConGPU` namespace.
- Declutters the helpers in a `hpp` + `cpp` file.

@PrometheusPi if you like have a look. Still needs runtime testing (hypnos is in maintenance), I just coded it out of the blue in parallel to waiting for some scripts to run ;)

P.S.: It might be possible we need to wait for this PR to be useful until we switched to `libSplash 1.3.0`

### DIFF
--- a/src/picongpu/include/plugins/common/stringHelpers.hpp
+++ b/src/picongpu/include/plugins/common/stringHelpers.hpp
@@ -23,31 +23,59 @@
 #include <string>
 #include <sstream>
 #include <ctime>
-
+#include <vector>
+#include <list>
+#include <algorithm>
+#include <iostream>
 
 namespace picongpu
+{
+namespace helper
 {
     /** Return the current date as string
      *
      * \param format, \see http://www.cplusplus.com/reference/ctime/strftime/
      * \return std::string with formatted date
      */
-    std::string getDateString( std::string format )
+    std::string getDateString( std::string format );
+
+    /** Create array of c-strings suitable for libSplash
+     *
+     * Convert a std::list of strings to a format that is suitable to
+     * be written into libSplash (concated and padded array of constant
+     * c-strings). Strings will be padded to longest string.
+     *
+     * Independent of the padding you chose, the strings will be '\0'
+     * separated & terminated. \0 padding is default and recommended.
+     */
+    class GetSplashArrayOfString
     {
-        time_t rawtime;
-        struct tm* timeinfo;
-        const size_t maxLen = 30;
-        char buffer [maxLen];
+    private:
+        // compare two std::string by their size
+        struct CompStrBySize
+        {
+            bool operator()( std::string i, std::string j )
+            {
+                return i.size() < j.size();
+            }
+        };
 
-        time( &rawtime );
-        timeinfo = localtime( &rawtime );
+    public:
+        // resulting type containing all attributes for a libSplash write call
+        struct Result
+        {
+            size_t maxLen;                // size of the longest string
+            std::vector<char> buffers;    // all of same length lenMax
 
-        strftime( buffer, maxLen, format.c_str(), timeinfo );
+            Result() : maxLen(0)
+            {}
+        };
 
-        std::stringstream dateString;
-        dateString << buffer;
+        Result operator()(
+            std::list<std::string> listOfStrings,
+            char padding = '\0'
+        );
+    };
 
-        return dateString.str();
-    }
-
+} // namespace helper
 } // namespace picongpu
--- a/src/picongpu/stringHelpers.cpp
+++ b/src/picongpu/stringHelpers.cpp
@@ -1,0 +1,101 @@
+/**
+ * Copyright 2015-2016 Axel Huebl
+ *
+ * This file is part of PIConGPU.
+ *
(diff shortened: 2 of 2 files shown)


ax3l:
I got a RT crash when testing right now, lets see how far we get
```
[1,2]<stderr>:[SPLASH_LOG:2] [1,3]<stderr>:[SPLASH_LOG:3] Exception: Exception for DCAttribute [someListOfStr] Attribute could not be opened or created
[1,2]<stderr>:Exception: Exception for DCAttribute [someListOfStr] Attribute could not be opened or created[1,2]<stderr>:terminate called after throwing an instance of 'splash::DCException'
[1,2]<stderr>:  what():  Exception for ParallelDataCollector::writeGlobalAttribute: failed to write attribute (someListOfStr)
[1,2]<stderr>:[kepler004:08974] *** Process received signal ***
```

PrometheusPi:
Interesting run time crash.
Was `someListOfStr` a global attribute or part of data/record?

ax3l:
ah I used the wrong `writeAttribute` signature in Splash, re-testing...

ax3l:
Ok I found the problem, the result needs to be contiguous and `\0` separated and *not* with an index table as in `main()`.

```C++
#include <typeinfo>    // for 'typeid'
#include <iostream>

    char *string_att[2] = {"This is array attribute",
                                 "It has two strings"};
    //printf("%s", string_att[0]);
    printf("%s\n", typeid(string_att).name());

    std::cout << std::endl;
    int i;
    for( i = 0; i < 50; ++i )
       std::cout << *(&string_att[0][0] + i)
                 << " ("
                 << int( *(&string_att[0][0] + i) )
                 << ")" << std::endl;
```

ax3l:
fixed, usage updated and successfully tested.

src/picongpu/include/plugins/common/stringHelpers.hpp
```diff
@@ -50,4 +58,86 @@ namespace picongpu
...
+            std::string longestString = *std::max_element(
+                    listOfStrings.begin(),
+                    listOfStrings.end(),
+                    compStrBySize
+            );
+            size_t lenMax = longestString.size();
+
+            // create libSplash data type suitable for longest string
+            Result result;
+            result.colType = splash::ColTypeString(lenMax);
```
PrometheusPi: @ax3l should that not be `splash::ColTypeString(lenMax+1)` since your string is `N+1` bytes long?

PrometheusPi:
Currently, I get a run time crash with out any exception calls from:
```
.../picongpu/bin/picongpu(_ZN6splash19SerialDataCollector14writeAttributeEiRKNS_14CollectionTypeEPKcS5_jNS_10DimensionsEPKv+0xa04)[0x721d24]
```

src/picongpu/include/plugins/common/stringHelpers.hpp
ax3l: nope, that is the [correct usage](https://github.com/ComputationalRadiationPhysics/libSplash/pull/167#discussion_r28645785).

ax3l:
I'll check it again, maybe I pushed an outdated commit.

PrometheusPi:
I tested the code for the **serial data collector** for both `.writeGlobalAttribute` and `.writeAttribute`. Both crash during these libSplash calls.

ax3l:
Ok, I found it. Accidently copied the CollectionType which leads to undefined results. Will push you a new even more shiny version tomorrow (caution, syntax will change a bit).

bussmann:
You wrote this 3:30 am?

ax3l:
Data analysis at daytime, hacking at nighttime Bat

ax3l:
@PrometheusPi updated and tested, worked for me :)

PrometheusPi:
I will have a look night-coder.

PrometheusPi:
run time test worked fine

PR #1325, opened 2016-01-08
Fix #1324 CUDA<7.5 + Boost 1.60.0 + C++11

### LINKED ISSUE #1324: Compile failure for CUDA 7.0 with boost 1.60
There is a compiler bug in CUDA 7.0 that surfaces with boost 1.60 and C++11 (which must be used in 1.60 due to a boost bug). That prevents compiling. This is fixed in CUDA 7.5

https://svn.boost.org/trac/boost/ticket/11897

ax3l:
Fix #1324 Work around as found by @Flamefire to mitigate a CUDA <=7.0 bug that prevents Boost 1.60.0 from building.

Previous boost versions are not affected.

fantastic work @Flamefire ! Can you confirm the work-around works for you?

### DIFF
--- a/src/picongpu/CMakeLists.txt
+++ b/src/picongpu/CMakeLists.txt
@@ -282,6 +282,16 @@ if( (Boost_VERSION EQUAL 106000) AND
                         "`-std=c++11` when compiling with Boost 1.60.0")
 endif()
 
+# Boost 1.60.0 and CUDA releases prior to 7.5 fail on variadic templates
+# when used with C++11
+if( (Boost_VERSION EQUAL 106000) AND
+    (CUDA_VERSION VERSION_LESS 7.5) AND
+    (NOT CMAKE_CXX_STANDARD EQUAL 98) )
+    # Boost Bug https://svn.boost.org/trac/boost/ticket/11897
+    message(STATUS "Boost: Disable variadic templates")
+    add_definitions(-DBOOST_NO_CXX11_VARIADIC_TEMPLATES)
+endif()
+
 
 ################################################################################
 # Find OpenMP


src/picongpu/CMakeLists.txt
```diff
@@ -282,6 +282,16 @@ if( (Boost_VERSION EQUAL 106000) AND
                         "`-std=c++11` when compiling with Boost 1.60.0")
 endif()
 
+# Boost 1.60.0 and CUDA releases prior to 7.5 fail on variadic templates
+# when used with C++11
+if( (Boost_VERSION EQUAL 106000) AND
+    (CUDA_VERSION VERSION_LESS 7.5) AND
+    (NOT CMAKE_CXX_STANDARD EQUAL 98) )
```
ax3l: there is only `CMAKE_CXX_STANDARD` `98` and `11` right now defined but a newer number might show up some time, too (and will be affected).

if you don't like that line we can write more clearly `(CMAKE_CXX_STANDARD EQUAL 11)`

src/picongpu/CMakeLists.txt
```diff
@@ -282,6 +282,16 @@ if( (Boost_VERSION EQUAL 106000) AND
...
+    # Boost Bug https://svn.boost.org/trac/boost/ticket/11897
+    message(STATUS "Boost: Disable variadic templates")
+    set(CUDA_NVCC_FLAGS
```
Flamefire: Rather use add_definitions which would be the correct way to add definitions

Flamefire:
Tested on hypnos with CUDA 7.0 and the use of  `add_definition`: WORKS :+1:

src/picongpu/CMakeLists.txt
ax3l:
```cmake
add_definitions(-DBOOST_NO_CXX11_VARIADIC_TEMPLATES)
```

ax3l:
updated. @erikzenker interested in this PR? :)

erikzenker:
I will have a look !

erikzenker:
Very good work! Thx to @ax3l  and @Flamefire. Could you please squash the commits to a single one? If this is done I will merge this PR :sun_with_face:

ax3l:
@erikzenker done :)

src/picongpu/CMakeLists.txt
ax3l: later finding in #1603: `-DBOOST_NO_CXX11_TEMPLATE_ALIASES` is enough as  a work-around

PR #1326, opened 2016-01-11
fix deadlock in asyncSend

### LINKED ISSUE #1322: Freeze of picongpu if used MPI_Allgather in plugin
Hi,

I noticed a bug in picongpu while writing at isaac. I managed to write a minimal example plugin, which shows the error. The plugin called "MPIStressTest" is avalaible at my picongpu fork in the branch "mpistresstest": https://github.com/theZiz/picongpu/tree/mpistresstest

The plugin uses this code in the `notify()` call:

``` C++
testClass.update( cellDescription );
//Some sample MPI magic
int numProc;
MPI_Comm_size(mpi_world, &numProc);
float buffer[numProc];
float my_value = float(rand());
if (mode & 1)
    MPI_Allgather( &my_value, 1, MPI_FLOAT, buffer, 1, MPI_FLOAT, mpi_world);
if (mode & 2)
    MPI_Bcast( buffer, numProc, MPI_FLOAT, 0, mpi_world);
```

The `mode` in line 7 and 9 is a parameter set in the submit cfg. The applications freezes if the mode is 1 or 3, so if "MPI_Allgather" is called. However it may be, that MPI_Bcast makes picongpu freeze, too, from time to time. I was just not able to repdroduce this.

However this MPI call is the reason for the freeze, but the application does not freeze at this point. It freezes in the `update` method of the dummy test class `testClass` with this implementation:

```
void update( MappingDesc *cellDescription )
{
    const SubGrid<simDim>& subGrid = Environment< simDim >::get().SubGrid();
    DataConnector &dc = Environment< simDim >::get().DataConnector(); 
    FieldE * pField = &(dc.getData< FieldE > (FieldE::getName(), true));
    DataSpace< simDim > guarding = SuperCellSize::toRT() * cellDescription->getGuardingSuperCells();
    //printf("get device data box\n");
    FieldE::DataBoxType dataBox = pField->getDeviceDataBox();
    //printf("done\n");
    FieldE::DataBoxType shifted = dataBox.shift( guarding );
}
```

The function `FieldE::DataBoxType dataBox = pField->getDeviceDataBox();` does not return (Checked with the surrounded printfs) for one node. However every else node will block at any later MPI call, too, because of the frozen rank.

I defined DEBUG_EVENTS in src/libPMacc/include/eventSystem/Manager.tpp to have some debug informations at this point. If the appliaction freezes this is printed to stdout:

```
[1,13]<stdout>:TaskReceive 1 2
[1,13]<stdout>:TaskReceiveMPI 1 2
```

Most of the time it happens right at the start, but sometimes after more than 100 steps, sometimes is runs the full simulation without problems. Most of the time the error is exactly as posted, but sometimes the 2 is switched for 3 or the message appears twice, but in this case the ranks in front of <stdout> are different.

I used the Laserwakefield example with default parameters. In my submit cfg I have only the test plugin activated. I am using 16 gpus with x=2, y=4, z=2 and a grid of 128 x 512 x 128. If I set x=1, y=16 and z=1 the bug appears more rarely, but still occurs.

It doesn't matter, how I start the job (direct qsub or qsub with bash).

If you need more informations about my test settings, don't hesitate to ask!

Greetings, Ziz

psychocoderHPC:
- `GridBuffer.hpp` and `ParticlesBuffer.hpp`: set gpuFree event to communication event
- close #1322 **Freeze of picongpu if used MPI_Allgather in plugin**

A deadlock in the async communication is triggered when process one opened a `MPI_Receive` while the destination process two not opened the `MPI_Send` and starts a MPI collective operation. In this case the event system stops and not start the `MPI_Send` from process two which wait in a side dependency chain. Process one wait for finish of the `MPI_Receive` before it can start the collective operation. Now we had a classical deadlock.

This patch only solve the deadlock a cleanup of the interfaces will done in a separate pull request.

**Tests:**
- [x] example plugin from #1322 ( command: `mpiexec -n  4 ./picongpu -d 2 2 1 -g  64 128 32 -s 10 --mpitest.period 1 --mpitest.mode 3`)

### DIFF
--- a/src/libPMacc/include/memory/buffers/GridBuffer.hpp
+++ b/src/libPMacc/include/memory/buffers/GridBuffer.hpp
@@ -519,6 +519,7 @@ public:
         {
             __startTransaction(serialEvent + sendEvents[sendEx]);
             sendEvents[sendEx] = sendExchanges[sendEx]->startSend(gpuFree);
+            gpuFree = sendEvents[sendEx];
             __endTransaction();
             return sendEvents[sendEx];
         }
--- a/src/libPMacc/include/particles/memory/buffers/ParticlesBuffer.hpp
+++ b/src/libPMacc/include/particles/memory/buffers/ParticlesBuffer.hpp
@@ -282,7 +282,7 @@ public:
         EventTask exchangeMemoryIndexerGPUEvent;
         EventTask returnEvent = framesExchanges->asyncSend(serialEvent, ex, framesExchangesGPUEvent) +
             exchangeMemoryIndexer->asyncSend(serialEvent, ex, exchangeMemoryIndexerGPUEvent);
-        gpuFree = framesExchangesGPUEvent + exchangeMemoryIndexerGPUEvent;
+        gpuFree = returnEvent;
         return returnEvent;
     }
 


ax3l:
thank you a lot! :sparkles:

PR #1327, opened 2016-01-11
clean up send interfaces

psychocoderHPC:
- remove copyEvent from all send methods in buffer and tasks
- `TaskSend`: remove member `EventTask copyEvent`
- `tasks/Factory`:
  - remove copy event from interface
  - update documentation

Follow up of ~~#1326~~

Tests:

- [x] LWFA with 16 K20


ax3l:
@psychocoderHPC thanks for the split! :)

Issue #1333, opened 2016-01-15
Allocate/Initialization Naming


ax3l:
Name all allocation and initialization members correctly.

- `allocate()`:
   - allocate ressources (such as GPUs) and memory
   - register in `DataConnector`
- `initialize()`:
   - fill allocated memory with data
   - add exchanges
- `clear()`:
   - reset to state after `allocate()` so `initialize()` can be called again (does not deallocate memory but might for example drop frames with particles)
- `deallocate()`:
   - deallocate memory
   - unsign from `DataConnector` (not implemented)

For functions that are currently not splitted or not useful to be splitted between `allocate()` and `initialize()` call them
- `allocateAndInitialize()`

ax3l:
ccing @psychocoderHPC

erikzenker:
I would like to have an offgithub meeting about this issue.

ax3l:
Absolutely, just recorded our offline discussion results so far for continuation (-> [WIP])

bussmann:
Would like to join.

PR #1335, opened 2016-01-15
FieldJ: `clear()` -> `assign(value)`

ax3l:
`Clear()` in STL language might resize members. What this member does is `assign` a value.

Related to #1333 (naming collision).

@psychocoderHPC @erikzenker


erikzenker:
Using `assign(val)` instead of `clear()` makes the source code more clear to me ...more explicit -> I like it !

PR #1334, opened 2016-01-15
Add fmod-function to libPMacc

BeyondEspresso:
Recently I used the `fmod()` function at several occasions (TWTS pulse, upcoming fix for the moving Window). So we have to include this math-function in the templates of libPMacc.

Note: The `fmod` function is entirely distinct from the `modf` function, which is already part of libPMacc.

PR #1336, opened 2016-01-15
Close #1330 Bugfix for slow moving window

### LINKED ISSUE #1330: Moving Window: Too Slow
@BeyondEspresso discovered that the moving window implementation is broken in terms of correctness of speed. Currently, it moves a few percent too slow due to rounding errors.

To fix this reasonably, the implementation in [`simulationControl/MovingWindow.hpp`](https://github.com/ComputationalRadiationPhysics/picongpu/blob/c17cbf98292c2b0e29035dfd05186d811e4748e7/src/picongpu/include/simulationControl/MovingWindow.hpp) needs a full refactoring, is highly obfuscated, very old and needs doxygen documentation (we knew that at least). (In contrast, [the wiki](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-domain-definitions#moving-window) documents all its non-code aspects very well.)

### Code Analysis

It's very old, pre-git code and we are not surprised that there will be dragons :)

**PIConGPU Code**
Affected `MovingWindow` members:
 - private (central and to-be-redone API): `getCurrentSlideInfo`
 - public: `getSlideCounter`, `slideInCurrentStep`, `getWindow` (the last one is the only one that will actually update the offset)

`getWindow` is the *only one* that will actually update the offset members in `MovingWindow`, but is not triggered when actually doing the slide in `simulationControl/MySimulation.hpp` leaving the offsets partly outdated for some time.

**PMacc code**
`MySimulation` calls then `GridController::slide(...)` which then implements the offset calculation *again on it's own* in PMacc `mappings/simulation/GridController::updateLocalDomainOffset` that finally also updates the `SubGrid`. Also, it triggers `communication/CommunicatorMPI::slide`... . Finally, as a special feature, the `GridController` and `CommunicatorMPI` each implement a `setStateAfterSlides` that can be used to restart a simulation and "emulate" the number of slides it did before (for `SubGrid` and `CommunicatorMPI`) and return true if a slide has happened *for this MPI-rank*.

Again, `slide/setStateAfterSlides` calls will perform a slide as soon as called and are collective but return only true if *one was one of the GPUs that was moved from back to front*.

Last word on the `CommunicatorMPI`: even if documented called in-code precisely, it updates only the position in the 3D grid of MPI ranks (no manipulation of cell domain information or offsets). At time of slide the exchange partners are updated. An interesting and fishy member is `yoffset` which is together with `coords` defining a virtual layer for enumerating the MPI-rank positions in the domain decomposition since the `MPI_Cart_create` operation is expensive and only done once at the beginning of the sim. One wants a continuously, monotonically increasing numbered space of GPU positions after slides. Its member `updateCoordinates` might still need a good review but is unaffected as long as slides continue to happen in `y`.

### Proposed Refactoring

[WIP!]

"Slide" describes the moving of a whole plane of GPUs from the back of the simulation box to the front.

**Algorithm** (PIConGPU Code)
`MovingWindow` should first separate the *raw algorithm on moving window offset calculation* and the *manipulation of members*.

Raw members for algorithm:
 - `setStartAfterStep(uint32_t)`: start in steps; step after that the moving window starts to move (time offset for the moving starts, precision: one time step)
 - `setSpeed(float64_t)`: currently `c`
 - `int64_t getCurrentMovedCells( float32_t cellSizeInDir, uint32_t currentStep)`: current number of cells moved which is a scalar, not a 3D position, which makes the direction irrelevant

The `MovingWindow`'s member `window` should be a separate singleton as `SubGrid`, manipulated with the results from the previous class.

**Slide Detection** (PIConGPU Code)

To detect if a slide occurs after the current step (which is the connection to PMacc that controls MPI & reinit after sliding) one can simply [calculate](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-domain-definitions):
```C++
    // allows static load balancing in y later on
    bool isSlideAfterStep = (globalDomain.size - globalWindow.offset - globalWindow.size) == 0;
```

**Offset and Window Size Updates**

`SubGrid` is already there for *local domain* size and offset remembering and can be manipulated by all members. `Window` is regularly updated every time step and `SubGrid` only on slides.

PIConGPU now decides on a `slideDirection`.

PIConGPU (MySimulation) code logic at the beginning of a step:
 - `isSlideAfterStep(t-1)`
 - `isSliding = `GPU position in Communicator...(slideDirection) is back of box in that slideDirection
 - if `isSlideAfterStep`: slide GC/Communicator
   - offset updates `SubGrid` (assume cell in y equal for all GPUs):
     - `SubGrid.globalDomain.offset(slideDirection) += SubGrid.localDomain.size`
     - `SubGrid.localDomain.offset(slideDirection) = globalWindow.offset(slideDirection) + globalWindow.size(slideDirection) - SubGrid.localDomain.size`
 - `moveCells = MovingWindow::getCurrentMovedCells`
 - update `Window`:
   - (can be implemented in a `Window` member)
   - `globalWindow.offset = MovingWindow::getCurrentMovedCells - SubGrid.globalDomain.offset`
   - `localWindow.offset = SubGrid.localDomain.offset - globalWindow.offset`
     - `< 0 ?` (means back of simulation box): `localWindow.size -= moveCells`
   - `[local&global]Window.offset += moveCells`
   - if `localWindow.size <= 0`: (or `isSliding == true`):
     - `localWindow.size += SubGrid.localDomain.size`
     - `localWindow.offset = 0`

BeyondEspresso:
Close #1330: As @ax3l described in #1330 we have a problem with the speed of the moving window in the current implementation. On a scale of several 100000 timesteps this can easily make an unaccounted window-offset of approx. 100 cells.

This pull request provides a minimal invasive fix for the physics side of the problem. I tested the code for the correct speed. The offset difference to the ideal is now smaller than the cell size and does not become larger with time.

- [x] depends on and needs rebase after #1334 was reviewed & merged

### DIFF
--- a/src/libPMacc/include/algorithms/math/defines/fmod.hpp
+++ b/src/libPMacc/include/algorithms/math/defines/fmod.hpp
@@ -1,0 +1,51 @@
+/**
+ * Copyright 2016 Alexander Debus
+ *
+ * This file is part of libPMacc.
+ *
+ * libPMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * libPMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with libPMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+namespace PMacc
+{
+namespace algorithms
+{
+namespace math
+{
+
+template<typename Type>
+struct Fmod;
+
+/**
+ * Equivalent of the modulus-operator for float types.
+ * Returns the floating-point remainder of x / y (rounded towards zero):
+ * fmod = x - tquot * y
+ * Where tquot is the truncated (i.e., rounded towards zero) result of: x / y.
+ * @return float value
+ */
+template<typename T1>
+HDINLINE typename Fmod< T1>::result fmod(T1 x,T1 y)
+{
+    return Fmod< T1 > ()(x, y);
+}
+
+} //namespace math
+} //namespace algorithms
+} //namespace PMacc
+
--- a/src/picongpu/include/simulationControl/MovingWindow.hpp
+++ b/src/picongpu/include/simulationControl/MovingWindow.hpp
@@ -1,5 +1,5 @@
 /**
- * Copyright 2013-2016 Axel Huebl, Heiko Burau, Rene Widera, Felix Schmitt
+ * Copyright 2013-2016 Axel Huebl, Heiko Burau, Rene Widera, Felix Schmitt, Alexander Debus
  *
  * This file is part of PIConGPU.
  *
@@ -63,20 +63,22 @@ private:
         /* later used to calculate smoother offsets */
         float_64 stepsInFutureAfterComma = stepsInFuture_tmp - (float_64) stepsInFuture;
 
-        /* round to nearest step so we get smaller sliding dfference
+        /* round to nearest step so we get smaller sliding difference
          * this is valid if we activate sliding window because y direction has
          * the same size for all gpus
          */
-        const uint32_t stepsPerGPU = (uint32_t) math::floor(
-                                                            (float_64) (subGrid.getLocalDomain().size.y() * cell_height) / light_way_per_step + 0.5);
-        const uint32_t firstSlideStep = stepsPerGPU * devices_y - stepsInFuture;
-        const uint32_t firstMoveStep = stepsPerGPU * (devices_y - 1) - stepsInFuture;
+        const float_64 stepsPerGPU = (float_64) (subGrid.getLocalDomain().size.y() * cell_height) / light_way_per_step;
+	const uint32_t firstSlideStep = (uint32_t) math::floor(   stepsPerGPU * (float_64) devices_y 
+                                                                - (float_64) stepsInFuture + 0.5           );
+        const uint32_t firstMoveStep =  (uint32_t) math::floor(   stepsPerGPU * (float_64) (devices_y - 1)
+                                                                - (float_64) stepsInFuture + 0.5           );
 
         if (slidingWindowActive==true && firstMoveStep <= currentStep)
         {
-            const uint32_t stepsInLastGPU = (currentStep + stepsInFuture) % stepsPerGPU;
+            const float_64 stepsInLastGPU = math::fmod( (float_64) currentStep + (float_64) stepsInFuture ,
+			                                           stepsPerGPU );
             /* moving window start */
-            if (firstSlideStep <= currentStep && stepsInLastGPU == 0)
+            if (firstSlideStep <= currentStep && stepsInLastGPU < float_64(1.0) )
             {
                 incrementSlideCounter(currentStep);
                 if (doSlide)
@@ -86,8 +88,8 @@ private:
             /* round to nearest cell to have smoother offset jumps */
             if (offsetFirstGPU)
             {
-                *offsetFirstGPU = math::floor(((float_64) stepsInLastGPU + stepsInFutureAfterComma) *
-                                              light_way_per_step / cell_height + 0.5);
+                *offsetFirstGPU = math::floor(  ( stepsInLastGPU + stepsInFutureAfterComma ) *
+                                                  light_way_per_step / cell_height + 0.5       );
             }
         }
     }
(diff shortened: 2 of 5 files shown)


ax3l:
thank you!
@psychocoderHPC let us have a look together.

src/libPMacc/include/algorithms/math/defines/fmod.hpp
```diff
@@ -0,0 +1,51 @@
...
+{
+
+template<typename Type>
+struct Fmod;
+
+/**
+ * Equivalent of the modulus-operator for float types.
+ * Returns the floating-point remainder of x / y (rounded towards zero):
+ * fmod = x - tquot * y
+ * Where tquot is the truncated (i.e., rounded towards zero) result of: x / y.
```
ax3l: That is a great documentation but pls try to avoid copy-pasting word-by-word the internet (copyright!)
  http://www.cplusplus.com/reference/cmath/fmod/

Can you rephrase it or set a reference?

src/picongpu/include/simulationControl/MovingWindow.hpp
```diff
@@ -63,20 +63,22 @@ class MovingWindow
...
          * this is valid if we activate sliding window because y direction has
          * the same size for all gpus
          */
-        const uint32_t stepsPerGPU = (uint32_t) math::floor(
-                                                            (float_64) (subGrid.getLocalDomain().size.y() * cell_height) / light_way_per_step + 0.5);
-        const uint32_t firstSlideStep = stepsPerGPU * devices_y - stepsInFuture;
-        const uint32_t firstMoveStep = stepsPerGPU * (devices_y - 1) - stepsInFuture;
+        const float_64 stepsPerGPU = (float_64) (subGrid.getLocalDomain().size.y() * cell_height) / light_way_per_step;
+	const uint32_t firstSlideStep = (uint32_t) math::floor(   stepsPerGPU * (float_64) devices_y 
+                                                                - (float_64) stepsInFuture + 0.5           );
```
ax3l: offline discussion: we will remove the `+ 0.5` to avoid troubles caused by "early slides"

src/picongpu/include/simulationControl/MovingWindow.hpp
```diff
@@ -63,20 +63,22 @@ class MovingWindow
...
+        const uint32_t firstMoveStep =  (uint32_t) math::floor(   stepsPerGPU * (float_64) (devices_y - 1)
+                                                                - (float_64) stepsInFuture + 0.5           );
```
ax3l: 0.5

src/picongpu/include/simulationControl/MovingWindow.hpp
```diff
@@ -86,8 +88,8 @@ class MovingWindow
             /* round to nearest cell to have smoother offset jumps */
             if (offsetFirstGPU)
             {
-                *offsetFirstGPU = math::floor(((float_64) stepsInLastGPU + stepsInFutureAfterComma) *
-                                              light_way_per_step / cell_height + 0.5);
+                *offsetFirstGPU = math::floor(  ( stepsInLastGPU + stepsInFutureAfterComma ) *
+                                                  light_way_per_step / cell_height + 0.5       );
```
ax3l: 0.5

BeyondEspresso:
I changed the rounding behaviour and removed a couple of decorating spaces according to @psychocoderHPC 's offline suggestions.

BeyondEspresso:
Note: With this commit, the corresponding specification in the wiki which is shown in `images/movingWindowOffsetEvolution.png` needs an update.

psychocoderHPC:
I checked your implementation and saw that it looks like the slide point detection for cells with `SPEED_OF_LIGHT * DELTA_T * 2 <= CELL_HEIGHT` was and is wrong.
IMO a better implementation is to calculate offset not in the step regime but in cells. 
I am working on it will show my results as fast as possible.

BeyondEspresso:
As already discussed offline: The slide-point detection in above implementation is correct. The second condition in
``if (firstSlideStep <= currentStep && stepsInLastGPU < float_64(1.0) )``
guarantees that it also works for cells with `SPEED_OF_LIGHT * DELTA_T * 2 <= CELL_HEIGHT`.

BeyondEspresso:
Since I included all suggestions two days ago, can we now remove the wait-4-author/WIP tag? :smiley:

ax3l:
sure, will review together with you guys Friday afternoon or next week.

src/picongpu/include/simulationControl/MovingWindow.hpp
```diff
@@ -86,8 +88,8 @@ class MovingWindow
             /* round to nearest cell to have smoother offset jumps */
             if (offsetFirstGPU)
             {
-                *offsetFirstGPU = math::floor(((float_64) stepsInLastGPU + stepsInFutureAfterComma) *
-                                              light_way_per_step / cell_height + 0.5);
+                *offsetFirstGPU = math::floor( ( stepsInLastGPU + stepsInFutureAfterComma ) *
+                                               light_way_per_step / cell_height + 0.5 );
```
ax3l: I think that might need an update, too (not with 0.5?)

ax3l: -> alex: correct, avoid twice rounding down

BeyondEspresso:
Since René has submitted and merged #1337 , which achieves the same goal, we can probably close this pull request.

psychocoderHPC:
solved with #1337

PR #1259, opened 2015-11-19
Refactor FieldTmp Particle Solvers

ax3l:
Refactors the particle solvers that populate `FieldTmp`, e.g., the HDF5 and ADIOS plugins.

The primarily goal of this refactoring is to make the current structure cleaner with separate classes and to allow later on to add solvers that populate `FieldTmp`, e.g., from pure field sources (e.g., div of E, |S|, etc.).

Additionally, namings and results of solvers have been clearified and standardized.

### Changes
 - all namings (according to openPMD ED-PIC extension, [version 1.0.0](https://github.com/openPMD/openPMD-standard/blob/1.0.0/EXT_ED-PIC.md#naming-conventions-for-mesh-records-field-records))
 - `<species>_chargeDensity` (previously called wrong "density", related to #909)
 - `<species>_energyDensity` (now without charge, previously created an
                          energy-charge-density)

### Adds
 - `<species>_density` (an actual density)

### To Do
- [x] run (looked ok)
- [x] run time comparison of results (dev/PR) (help wanted)

PR #1340, opened 2016-01-19
Tools: Fix Naming of HDF5 Density

PrometheusPi:
Follow up to #1259: Due to the update of the naming of some fields in hdf5, the charge conservation tools were outdated. This pull request changes the naming to the new convention. 
```
Density_e -> e_chargeDensity
```

### DIFF
--- a/src/tools/bin/plot_chargeConservation.py
+++ b/src/tools/bin/plot_chargeConservation.py
@@ -89,7 +89,7 @@ def plotError(h5file, slice_pos=[0.5, 0.5, 0.5]):
     charge = np.zeros_like(Ex)
     norm = 0.0
     for field_name in f["/data/{}/fields/".format(timestep)].keys():
-        if field_name[0:8] == "Density_":
+        if field_name[-14:] == "_chargeDensity":
             # load species density
             species_Density = np.array(f["/data/{}/fields/".format(timestep) + field_name])
             # choose norm to be the maximal charge density of all species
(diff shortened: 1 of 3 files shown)


src/tools/bin/plot_chargeConservation.py
```diff
@@ -89,7 +89,7 @@ def plotError(h5file, slice_pos=[0.5, 0.5, 0.5]):
     charge = np.zeros_like(Ex)
     norm = 0.0
     for field_name in f["/data/{}/fields/".format(timestep)].keys():
-        if field_name[0:8] == "Density_":
+        if field_name[1:15] == "_chargeDensity":
```
PrometheusPi: This will not work for species names longer than one character. Let me fix this tomorrow.

src/tools/bin/plot_chargeConservation_overTime.py
```diff
@@ -106,7 +106,7 @@ def deviation_charge_conservation(h5file):
     charge = np.zeros_like(Ex)
     norm = 0.0
     for field_name in f["/data/{}/fields/".format(timestep)].keys():
-        if field_name[0:8] == "Density_":
+        if field_name[1:15] == "_chargeDensity":
```
PrometheusPi: Same here.

src/tools/bin/plot_chargeConservation.py
ax3l: sure, just regex it to match all before the *first* `_`

ax3l:
great, thank you for updating those, too! Did you `grep` if we catched all now?

src/tools/bin/plot_chargeConservation.py
PrometheusPi: instead of using regex, python allows checking the last characters. 
Fixed.

ax3l: even better solution!

PrometheusPi:
Yes, I `grep`ed for the hdf5 string.

ax3l:
excellent! :sparkles:

PR #1342, opened 2016-01-22
cmath: Use Correct C++ Include

ax3l:
Fix all the `<cmath>` includes to be [C++](https://en.wikipedia.org/wiki/C%2B%2B_Standard_Library#C_standard_library).


ax3l:
will fix more in a follow-up

erikzenker:
Very good work :-)

ax3l:
just a little cleanup :cake:

Issue #1344, opened 2016-01-22
Clarification on PhysicalMemorySize


Flamefire:
I just run into an issue about the understanding of `getPhysicalMemorySize` from the buffers (in this case `DeviceBuffer`)
I assumed, that `getPhysicalMemorySize()[0] * sizeof(Type)` would be the same as the pitch. However it is not. The pitch is stored in the cudaPitchedPtr. For bigger types this would mostly be the same, but e.g. for uneven byte-size types, this is not the case. CUDA will pad the line so that e.g. `getPhysicalMemorySize()[0]`will be 3 but the pitch is 4.
This especially means, that `getPhysicalMemorySize().productOfComponents() * sizeof(Type)` will not be the amount of allocated memory, although it suggest that it would. This can lead to subtile bugs like setting all elements but the last few to a value.

This should either be clarified or changed to return the size in bytes.

@psychocoderHPC

psychocoderHPC:
This is the size of the memory in elements. If you need the ptch please use `getPitch()`. Pitch times physicleMemorySize.y is the amount of bytes needed to move from a y plane to the next.

Flamefire:
But `getPhysicalMemorySize().productOfComponents()` is not the size of the whole buffers memory (in elements)

psychocoderHPC:
Its the size of the buffer in elements. But is the size of the allocated buffer. If you create a buffer out of an other one than you get the size of the oldest parent buffer.

Flamefire:
Did you forgot a word in your first 2 sentences? I don't get what you mean.
I'd expect that something like the following would work:

    Type* ptr = getStartPtr();
    for(unsigned i=0; i<getPhysicalMemorySize().productOfComponents(); i++)
        ptr[i] = Type(0);

Or the equivalent using BytePtrs.   
This would however ignore some elements at the end of the (oldest parent) buffer which is at least counter-intuitive.

psychocoderHPC:
Use `getDataSpace()` as size and `getDataBox` to access the memory to loop over elements.

Flamefire:
Concerning your edit: Then the name and documentation needs to be updated. Currently it is:

         /** get size of the physical memory (in elements)
         */
        DataSpace<DIM> getPhysicalMemorySize()

If it is the size of the oldest parent buffer in elements then it should state exactly that. Otherwise `getPhysicalMemorySize().productOfComponents() * sizeof(Type)` should be the allocation size (=physical memory size in bytes)

PR #1094, opened 2015-09-11
[WIP] cuSTL Host/DeviceBuffers: Contructors (Pointers)

Flamefire:
This implements the cartBuffer functions correctly by calling constructors that actually take parameters. Allowing the default ctor is dangerous.

This also adds the missing move operators required for effective returning.

This also supersedes #1093 

- [x] Do not merge until rebased and changed according to PR #1096 (see: https://github.com/ComputationalRadiationPhysics/picongpu/pull/1094#issuecomment-139565612)
- [x] Test plugins using this

PR #1345, opened 2016-01-22
Use correct pitch for cartBuffer conversion

Flamefire:
`getPhysicalMemorySize()` returns the size in elements, not in bytes.
Also `getPhysicalMemorySize()[0] * sizeof(Type)` may not be the pitch for device pointers (See #1344)

@Heikman: Please merge urgently

introduced by #1094.


Heikman:
introduced by #1094. Thanks for fixing.

PR #1346, opened 2016-01-22
C Includes Examples: Fix Correct C++ Style

ax3l:
Fix all `cassert` (RT assert) includes in PIConGPU examples to use correct [C++ include style](https://en.wikipedia.org/wiki/C%2B%2B_Standard_Library#C_standard_library) instead of C-style.


PrometheusPi:
Could you add your name to the license header (of the last file).

PrometheusPi:
The rest looks great.

ax3l:
updated

PrometheusPi:
@ax3l great :+1:

PR #1347, opened 2016-01-22
C Includes Tools: Fix Correct C++ Style

ax3l:
Fix all includes in PIConGPU tools to use correct [C++ include style](https://en.wikipedia.org/wiki/C%2B%2B_Standard_Library#C_standard_library) instead of C-style.


PrometheusPi:
@ax3l Please add your name to **all** license headers.
Do you want to keep the commit *Revert `<stdint.h>` for now* and remove it later or could you squash them to a single commit?
Rest looks fine :+1:

ax3l:
will do so if that trivial change of +/- 6 lines justifies that in your eyes.

PrometheusPi:
@ax3l thx

PR #1348, opened 2016-01-22
C Includes: Fix Correct C++ Style

ax3l:
Fix all includes in PIConGPU to use correct [C++ include style](https://en.wikipedia.org/wiki/C%2B%2B_Standard_Library#C_standard_library) instead of C-style.

### DIFF
--- a/src/picongpu/include/plugins/output/header/MessageHeader.hpp
+++ b/src/picongpu/include/plugins/output/header/MessageHeader.hpp
@@ -24,7 +24,7 @@
 #include "simulation_defines.hpp"
 #include "dimensions/DataSpace.hpp"
 #include "iostream"
-#include "stdlib.h"
+#include "cstdlib"
 
 #include "plugins/output/header/DataHeader.hpp"
 #include "plugins/output/header/NodeHeader.hpp"
--- a/src/picongpu/include/plugins/output/header/NodeHeader.hpp
+++ b/src/picongpu/include/plugins/output/header/NodeHeader.hpp
@@ -23,7 +23,7 @@
 #include "types.h"
 #include "dimensions/DataSpace.hpp"
 #include "iostream"
-#include "stdlib.h"
+#include "cstdlib"
 
 struct NodeHeader
 {
--- a/src/picongpu/include/plugins/output/header/SimHeader.hpp
+++ b/src/picongpu/include/plugins/output/header/SimHeader.hpp
@@ -23,7 +23,7 @@
 #include "types.h"
 #include "dimensions/DataSpace.hpp"
 #include "iostream"
-#include "stdlib.h"
+#include "cstdlib"
 
 struct SimHeader
 {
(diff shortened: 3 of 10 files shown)


ax3l:
@ComputationalRadiationPhysics/picongpu-maintainers I opened a couple of PRs of this style fixing the whole project.

Please read/understand and make sure in reviews that people do not include c-style includes any more. Thanks!

References:
  - [C vs C++ include Wikipedia](https://en.wikipedia.org/wiki/C%2B%2B_Standard_Library#C_standard_library)
  - [cstdint is C++11](http://www.cplusplus.com/reference/cstdint/) and `stdint.h` is a plain `C` header (see also [this bug](https://github.com/ComputationalRadiationPhysics/mallocMC/pull/105)) -> we can update this to `<cstdint>` within the C++11 merge window #1350

PrometheusPi:
Could you add your name to the license header:
 - `radiation_list_freq.hpp`
 - `SocketConnector.hpp`

Same question as in #1347 - could you squash the two commits?

ax3l:
done

PrometheusPi:
@ax3l thx

src/picongpu/include/plugins/output/header/MessageHeader.hpp
```diff
@@ -24,7 +24,7 @@
 #include "simulation_defines.hpp"
 #include "dimensions/DataSpace.hpp"
 #include "iostream"
-#include "stdlib.h"
+#include "cstdlib"
```
Flamefire: Ouch! How could this slip through? Please always include system headers with <> style

ax3l: eieiei!

ax3l: will fix

Flamefire: Same for iostream

src/picongpu/include/plugins/output/header/NodeHeader.hpp
```diff
@@ -23,7 +23,7 @@
 #include "types.h"
 #include "dimensions/DataSpace.hpp"
 #include "iostream"
-#include "stdlib.h"
+#include "cstdlib"
```
Flamefire: <>

src/picongpu/include/plugins/output/header/SimHeader.hpp
```diff
@@ -23,7 +23,7 @@
 #include "types.h"
 #include "dimensions/DataSpace.hpp"
 #include "iostream"
-#include "stdlib.h"
+#include "cstdlib"
```
Flamefire: <>

src/picongpu/include/plugins/output/header/WindowHeader.hpp
```diff
@@ -23,7 +23,7 @@
 #include "types.h"
 #include "dimensions/DataSpace.hpp"
 #include "iostream"
-#include "stdlib.h"
+#include "cstdlib"
 
 struct WindowHeader
```
Flamefire: <>

src/picongpu/include/plugins/output/header/MessageHeader.hpp
ax3l: thx, saw that and only affects this specific plugin

PR #1349, opened 2016-01-22
C Includes PMacc: Fix Correct C++ Style

ax3l:
Fix all includes in PMacc to use correct [C++ include style](https://en.wikipedia.org/wiki/C%2B%2B_Standard_Library#C_standard_library) instead of C-style.


PrometheusPi:
@ax3l why did you close this pull request?

ax3l:
wrong button.

PrometheusPi:
My fault - the revert in the second commit resulted in no files changed.

ax3l:
ah no, indeed that was the reason I closed it.

PR #1352, opened 2016-01-26
Add LazySharedMemAllocator

Heikman:
Add a new compile-time allocator which enables a shared memory buffer object to be constructed at host site and to allocate its memory later on the device.

Tested for the use case that allocation and access are in the same function.
If this is not the case, the program freezes. Therefore we need further investigation on this.

### DIFF
--- a/src/libPMacc/include/cuSTL/container/compile-time/CartBuffer.hpp
+++ b/src/libPMacc/include/cuSTL/container/compile-time/CartBuffer.hpp
@@ -40,27 +40,24 @@ namespace CT
  * \tparam _Size compile-time vector specifying the size of the container
  */
 template<typename Type, typename _Size, typename Allocator, typename Copier, typename Assigner>
-class CartBuffer
+class CartBuffer : public Allocator
 {
 public:
     typedef Type type;
     typedef _Size Size;
     typedef typename Allocator::Pitch Pitch;
-    typedef cursor::CT::BufferCursor<Type, Pitch> Cursor;
+    typedef typename Allocator::Cursor Cursor;
     BOOST_STATIC_CONSTEXPR int dim = Size::dim;
     typedef zone::CT::SphericZone<_Size, typename math::CT::make_Int<dim, 0>::type> Zone;
-private:
-    Type* dataPointer;
-    //HDINLINE void init();
+
 public:
-    DINLINE CartBuffer();
-    DINLINE CartBuffer(const CT::CartBuffer<Type, Size, Allocator, Copier, Assigner>& other);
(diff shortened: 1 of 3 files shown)


src/libPMacc/include/cuSTL/container/allocator/compile-time/SharedMemAllocator.hpp
```diff
@@ -42,11 +42,23 @@ struct SharedMemAllocator<Type, Size, 1, uid>
     BOOST_STATIC_CONSTEXPR int dim = 1;
     typedef cursor::CT::BufferCursor<type, math::CT::UInt32<> > Cursor;
 
-    __device__ static Cursor allocate()
+protected:
```
psychocoderHPC: changes in this file are independent of the new allocator `LazySharedMemAllocator` please open a new refactoring pull request

src/libPMacc/include/cuSTL/container/allocator/compile-time/SharedMemAllocator.hpp
```diff
@@ -73,11 +97,23 @@ struct SharedMemAllocator<Type, Size, 3, uid>
...
+
+    Cursor cursor;
+
+    HDINLINE SharedMemAllocator() : cursor(NULL) {}
+
+    /* has to be static, unless a crash is acceptable */
+    DINLINE static Cursor _allocate()
```
psychocoderHPC: Please [avoid prefix `_` ](http://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier)

src/libPMacc/include/cuSTL/container/compile-time/CartBuffer.hpp
```diff
@@ -40,27 +40,24 @@ namespace CT
  * \tparam _Size compile-time vector specifying the size of the container
  */
 template<typename Type, typename _Size, typename Allocator, typename Copier, typename Assigner>
-class CartBuffer
+class CartBuffer : public Allocator
```
psychocoderHPC: This should be part of the shared memory refactoring pull request

psychocoderHPC:
I added the flag `WIP`

We should wait with this feature until #1358 and the freezes are solved

psychocoderHPC:
Can we close this pull request? Since this type of lazy allocator make no sense if the allocator can only used in the same stack level than it is allocated and only in the same function (#1358)
If this problem is solved in the future we can create a new one.

Heikman:
I'm fine with this.

PR #1353, opened 2016-01-27
Add general particle creation module

Heikman:
Add a general module for particle creation. A `particleCreator` functor may generate one or multiple new target particles out of a single source particle.

PR #1339, opened 2016-01-19
add memory shift option to particle to field interpolation

PrometheusPi:
This pull request allows to shift memory during the particle to field interpolation, that  allows to perform interpolation during sub-stepping **if a particle leaves its cell**.
It was suggested by @psychocoderHPC in [this comment](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1201/files#r44545194).

The following test have to be performed  before merging:
 - [x] thermal test `NoShift`
 - [x] charge conservation `NoShift`
 - [x] thermal test `ShiftToValidRange`
 - [x] charge conservation`ShiftToValidRange`

Issue #1338 requires using `PMacc::math`.

PR #1236, opened 2015-11-09
Implement random buffer

Flamefire:
This implements a buffer for storing the random states `RNGProvider` to avoid reinitialising the RNG on every step.
All functionality currently provided by the RNG implementation of PMacc is also provided by this, so it can be used instead of the current implementation.

Typical usage: 
***1) Setup***

1. Create a typedef for the RNGProvider(s) you want to use: `using RNGProvider = PMacc::random::RNGProvider<simDim>;`
2. Create a (per rank) instance of the provider with the required size: `new RNGProvider(Environment::get().SubGrid().getLocalDomain().size)`   
You can optionally specify an id for this instance. The default id is sufficient if there is only one instance per `RNGProvider` type.
3. Initialize it with a (per rank) seed: `rngProvider_->init(seed);`

***2a) Usage with reusable handle (one handle, many distributions)***

1. Get the Handle type out of the RNGProvider:   
`using RNGHandle = typename RNGProvider::Handle;`
2. Choose a distribution:   
`using Distribution = PMacc::random::distributions::Uniform_float<>;`
3. Get the Random (Functor) type by combining those:   
`using Random = typename RNGHandle::GetRandomType<Distribution>::type;`
4. Construct the handle:   
`RNGHandle randHandle(RNGProvider::createHandle())`   
Here you specify the id of the RNGProvider instance to use. Again the default id is used.
5. (Usually on the device) initialize the handle:   
`randHandle.init(localCellIdx);`
6. Get the random functor:   
`Random rand = randHandle.applyDistribution<Distribution>();`
7. Use it by calling it: `rand()`

It is possible, to store the state into shared memory by calling `randHandle.getState()` after 5. and use that for multiple `Random` functors by passing a pointer to the state to `Random`s constructor.

***2b) Shortcut without reusable handle (one handle, one distribution)***

1. Choose a distribution:   
`using Distribution = PMacc::random::distributions::Uniform_float<>;`
2. Get the Random (Functor) type out of the Provider:   
`using Random = typename RNGProvider::GetRandomType<Distribution>::type;`
3. Construct the functor:   
`Random random(RNGProvider::createRandom())`   
Here you specify the id of the RNGProvider instance to use. Again the default id is used.
4. (Usually on the device) initialize the functor:   
`random.init(localCellIdx);`
5. Use it by calling it: `rand()`

PR #1215, opened 2015-10-27
add option to extend margin for pusher interpolation

PrometheusPi:
This pull request adds an option in ~~`pusher.unitless`~~ *the particle pusher* to extend the margin for field to particle interpolation for the in-pusher interpolation.

This will be necessary for sub-sampling if a particle moves out of the currently stored field data (by one cell) and then wants to interpolate again. 

**Update:**
I use the pusher margins for calculating the neighbors in fields too.

PR #1354, opened 2016-01-27
Synchrotron photon emission

Heikman:
Implement synchrotron radiation of photons for electrons. It takes the numerical model from:
 > Gonoskov, A., et al. "Extended particle-in-cell schemes for physics in ultrastrong laser fields: Review and developments." Physical Review E 92.2 (2015): 023305.

**Dependencies:**
 - [x] #1353 
 - [x] #1459 
 - [x] ~~add support for optional memory usage for the RNG~~ -> done via check of radiation electrons
 - [x] check register memory usage for critical log output
 - [x] remove pre-compiler flag as switch from param
 - [x] use pre-compiler flag in cmake compile test

#### Synchrotron radiation in the classical regime

Photon spectrum:
photon_spec_t_0
Zoom:
photon_spec_t_0_zoom

**legend:**
green: analytical result
blue: simulation result, error bars: statistical error

### Known Issues
 - ~~photons are created, but they don't move despite of the photon pusher.~~
 - ~~ugly compiler warnings.~~

### DIFF
--- a/src/picongpu/include/particles/ParticlesFunctors.hpp
+++ b/src/picongpu/include/particles/ParticlesFunctors.hpp
@@ -1,5 +1,6 @@
 /**
- * Copyright 2014-2016 Rene Widera, Marco Garten, Alexander Grund
+ * Copyright 2014-2016 Rene Widera, Marco Garten, Alexander Grund,
+ *                     Heiko Burau
  *
  * This file is part of PIConGPU.
  *
@@ -32,6 +33,9 @@
 #include "communication/AsyncCommunication.hpp"
 #include "particles/traits/GetIonizer.hpp"
 #include "particles/traits/FilterByFlag.hpp"
+#include "particles/traits/GetPhotonCreator.hpp"
+#include "particles/synchrotronPhotons/SynchrotronFunctions.hpp"
+#include "particles/creation/creation.hpp"
 
 namespace picongpu
 {
@@ -313,6 +317,63 @@ struct CallIonization
 
 }; // struct CallIonization
 
-} //namespace particles
+/** Handles the synchrotronPhotons effect for electrons.
+ *
+ * \tparam T_SpeciesName name of electron species
+ */
+template<typename T_SpeciesName>
+struct CallSynchrotronPhotons
+{
+    typedef T_SpeciesName SpeciesName;
+    typedef typename SpeciesName::type SpeciesType;
+    typedef typename SpeciesType::FrameType FrameType;
+    /* SelectedPhotonCreator will be either PhotonCreator or fallback: CreatorBase */
+    typedef typename GetPhotonCreator<SpeciesType>::type SelectedPhotonCreator;
 
-} //namespace picongpu
+    /** Functor implementation
+     *
+     * \tparam T_StorageStuple contains info about the particle species
+     * \tparam T_CellDescription contains the number of blocks and blocksize
+     *                           that is later passed to the kernel
+     * \param tuple An n-tuple containing the type-info of multiple particle species
+     * \param cellDesc points to logical block information like dimension and cell sizes
+     * \param currentStep The current time step
+     */
+    template<typename T_StorageTuple, typename T_CellDescription>
+    HINLINE void operator()(
+                            T_StorageTuple& tuple,
+                            T_CellDescription* cellDesc,
+                            const uint32_t currentStep,
+                            const synchrotronPhotons::SynchrotronFunctions& synchrotronFunctions) const
+    {
+        typedef typename SelectedPhotonCreator::PhotonSpecies PhotonSpecies;
+
+        /* get E- and B-field */
+        DataConnector &dc = Environment<>::get().DataConnector();
+        BOOST_AUTO(fieldE,
+                 dc.getData<FieldE>(FieldE::getName(), true).getGridBuffer().
+                 getDeviceBuffer().cartBuffer());
+        BOOST_AUTO(fieldB,
+                 dc.getData<FieldB>(FieldB::getName(), true).getGridBuffer().
+                 getDeviceBuffer().cartBuffer());
+
+        /* alias for pointer on source species */
+        PMACC_AUTO(electronSpeciesPtr, tuple[SpeciesName()]);
+        /* alias for pointer on destination species */
+        PMACC_AUTO(photonSpeciesPtr,  tuple[typename MakeIdentifier<PhotonSpecies>::type()]);
+
+        using namespace synchrotronPhotons;
+
+        SelectedPhotonCreator photonCreator(
+            synchrotronFunctions.getCursor(SynchrotronFunctions::first),
+            synchrotronFunctions.getCursor(SynchrotronFunctions::second),
+            currentStep);
+
+        creation::createParticles(*electronSpeciesPtr, *photonSpeciesPtr, photonCreator, cellDesc);
+    }
+
+}; // struct CallSynchrotronPhotons
+
+} // namespace particles
+
+} // namespace picongpu
--- a/src/picongpu/include/particles/ParticlesFunctors.hpp
+++ b/src/picongpu/include/particles/ParticlesFunctors.hpp
@@ -339,10 +339,10 @@ struct CallSynchrotronPhotons
      * \param tuple An n-tuple containing the type-info of multiple particle species
      * \param cellDesc points to logical block information like dimension and cell sizes
      * \param currentStep The current time step
+     * \param synchrotronFunctions synchrotron functions wrapper object
      */
     template<typename T_StorageTuple, typename T_CellDescription>
-    HINLINE void operator()(
-                            T_StorageTuple& tuple,
+    HINLINE void operator()(T_StorageTuple& tuple,
                             T_CellDescription* cellDesc,
                             const uint32_t currentStep,
                             const synchrotronPhotons::SynchrotronFunctions& synchrotronFunctions) const
--- a/src/picongpu/include/particles/ParticlesFunctors.hpp
+++ b/src/picongpu/include/particles/ParticlesFunctors.hpp
@@ -319,7 +319,7 @@ struct CallIonization
 
 }; // struct CallIonization
 
-/** Handles the synchrotronPhotons effect for electrons.
+/** Handles the synchrotron radiation emission of photons from electrons
  *
  * \tparam T_SpeciesName name of electron species
  */
@@ -330,7 +330,7 @@ struct CallSynchrotronPhotons
     typedef typename SpeciesName::type SpeciesType;
     typedef typename SpeciesType::FrameType FrameType;
     /* SelectedPhotonCreator will be either PhotonCreator or fallback: CreatorBase */
-    typedef typename GetPhotonCreator<SpeciesType>::type SelectedPhotonCreator;
+    typedef typename traits::GetPhotonCreator<SpeciesType>::type SelectedPhotonCreator;
 
     /** Functor implementation
      *
@@ -350,15 +350,6 @@ struct CallSynchrotronPhotons
     {
         typedef typename SelectedPhotonCreator::PhotonSpecies PhotonSpecies;
 
-        /* get E- and B-field */
-        DataConnector &dc = Environment<>::get().DataConnector();
-        BOOST_AUTO(fieldE,
-                 dc.getData<FieldE>(FieldE::getName(), true).getGridBuffer().
-                 getDeviceBuffer().cartBuffer());
-        BOOST_AUTO(fieldB,
-                 dc.getData<FieldB>(FieldB::getName(), true).getGridBuffer().
-                 getDeviceBuffer().cartBuffer());
-
         /* alias for pointer on source species */
         PMACC_AUTO(electronSpeciesPtr, tuple[SpeciesName()]);
         /* alias for pointer on destination species */
@@ -369,7 +360,7 @@ struct CallSynchrotronPhotons
             synchrotronFunctions.getCursor(SynchrotronFunctions::first),
             synchrotronFunctions.getCursor(SynchrotronFunctions::second));
 
-        creation::createParticles(*electronSpeciesPtr, *photonSpeciesPtr, photonCreator, cellDesc);
+        creation::createParticlesFromSpecies(*electronSpeciesPtr, *photonSpeciesPtr, photonCreator, cellDesc);
     }
 
 }; // struct CallSynchrotronPhotons
--- a/src/picongpu/include/particles/ParticlesFunctors.hpp
+++ b/src/picongpu/include/particles/ParticlesFunctors.hpp
@@ -30,11 +30,13 @@
 #include <boost/mpl/plus.hpp>
 #include <boost/mpl/accumulate.hpp>
 
+
 #include "communication/AsyncCommunication.hpp"
 #include "particles/traits/GetIonizer.hpp"
 #include "particles/traits/FilterByFlag.hpp"
 #include "particles/traits/GetPhotonCreator.hpp"
 #include "particles/synchrotronPhotons/SynchrotronFunctions.hpp"
+#include "particles/synchrotronPhotons/PhotonCreator.hpp"
 #include "particles/creation/creation.hpp"
 
 namespace picongpu
@@ -360,14 +362,12 @@ struct CallSynchrotronPhotons
         /* alias for pointer on source species */
         PMACC_AUTO(electronSpeciesPtr, tuple[SpeciesName()]);
         /* alias for pointer on destination species */
-        PMACC_AUTO(photonSpeciesPtr,  tuple[typename MakeIdentifier<PhotonSpecies>::type()]);
+        PMACC_AUTO(photonSpeciesPtr, tuple[typename MakeIdentifier<PhotonSpecies>::type()]);
 
         using namespace synchrotronPhotons;
-
         SelectedPhotonCreator photonCreator(
             synchrotronFunctions.getCursor(SynchrotronFunctions::first),
-            synchrotronFunctions.getCursor(SynchrotronFunctions::second),
-            currentStep);
+            synchrotronFunctions.getCursor(SynchrotronFunctions::second));
 
         creation::createParticles(*electronSpeciesPtr, *photonSpeciesPtr, photonCreator, cellDesc);
     }
--- a/src/picongpu/include/particles/ParticlesFunctors.hpp
+++ b/src/picongpu/include/particles/ParticlesFunctors.hpp
@@ -36,7 +36,6 @@
 #include "particles/traits/FilterByFlag.hpp"
 #include "particles/traits/GetPhotonCreator.hpp"
 #include "particles/synchrotronPhotons/SynchrotronFunctions.hpp"
-#include "particles/synchrotronPhotons/PhotonCreator.hpp"
 #include "particles/creation/creation.hpp"
 
 namespace picongpu
--- a/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.def
+++ b/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.def
@@ -24,17 +24,17 @@ namespace picongpu
 {
 namespace particles
 {
-namespace bremsstrahlung
+namespace synchrotronPhotons
 {
 
-/** Sample point stepping */
-BOOST_CONSTEXPR_OR_CONST float_64 SYNC_FUNCS_STEP_WIDTH =
-    SYNC_FUNCS_CUTOFF / static_cast<float_64>(SYNC_FUNCS_NUM_SAMPLES - 1u);
+/**
+ *
+ * \tparam T_ElectronSpecies
+ * \tparam T_PhotonSpecies
+ */
+template<typename T_ElectronSpecies, typename T_PhotonSpecies>
+struct PhotonCreator;
 
-/** In the definition of the first synchrotron function the bessel function is integrated
- * up to infinity but in fact it is sufficient to integrate up to this constant. */
-BOOST_CONSTEXPR_OR_CONST float_64 SYNC_FUNCS_F1_INTEGRAL_BOUND = 50.0;
-
-} // namespace bremsstrahlung
+} // namespace synchrotronPhotons
 } // namespace particles
 } // namespace picongpu
--- a/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.def
+++ b/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.def
@@ -27,7 +27,15 @@ namespace particles
 namespace synchrotronPhotons
 {
 
-/**
+/** Functor creating photons from electrons according to synchrotron radiation.
+ *
+ * The numerical model is taken from:
+ *
+ * Gonoskov, A., et al. "Extended particle-in-cell schemes for physics
+ * in ultrastrong laser fields: Review and developments."
+ * Physical Review E 92.2 (2015): 023305.
+ *
+ * This functor is called by the general particle creation module.
  *
  * \tparam T_ElectronSpecies
  * \tparam T_PhotonSpecies
--- a/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
+++ b/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
@@ -29,6 +29,8 @@
 #include "algorithms/math/defines/cross.hpp"
 #include "traits/frame/GetMass.hpp"
 #include "traits/frame/GetCharge.hpp"
+#include "particles/operations/Assign.hpp"
+#include "particles/operations/Deselect.hpp"
 
 #include "random/methods/XorMin.hpp"
 #include "random/distributions/Uniform.hpp"
@@ -320,11 +322,17 @@ public:
     DINLINE void operator()(Electron& electron, Photon& photon) const
     {
         photon[multiMask_] = 1;
-        photon[localCellIdx_] = electron[localCellIdx_];
-        photon[position_] = electron[position_];
         photon[momentum_] = this->photon_mom;
         electron[momentum_] -= this->photon_mom;
-        photon[weighting_] = electron[weighting_];
+        PMACC_AUTO(destPhoton,
+            deselect<
+                boost::mpl::vector<
+                    multiMask,
+                    momentum
+                >
+            >(photon)
+        );
+        particles::operations::assign( destPhoton, electron );
     }
 };
 
--- a/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
+++ b/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
@@ -284,8 +284,12 @@ public:
 
         if(this->randomGen() < x)
         {
-            this->photon_mom = mom_norm * z*z*z * mass*c * gamma * particle[weighting_];
-            return 1;
+            const float_X photonMom_abs = z*z*z * mass*c * gamma;
+            if(photonMom_abs > SOFT_PHOTONS_CUTOFF_MOM)
+            {
+                this->photon_mom = mom_norm * photonMom_abs * particle[weighting_];
+                return 1;
+            }
         }
 
         return 0;
--- a/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
+++ b/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
@@ -30,6 +30,10 @@
 #include "traits/frame/GetMass.hpp"
 #include "traits/frame/GetCharge.hpp"
 
+#include "random/methods/XorMin.hpp"
+#include "random/distributions/Uniform.hpp"
+#include "random/RNGProvider.hpp"
+
 #include "traits/Resolve.hpp"
 #include "mappings/kernel/AreaMapping.hpp"
 
@@ -38,9 +42,6 @@
 
 #include "compileTime/conversion/TypeToPointerPair.hpp"
 #include "memory/boxes/DataBox.hpp"
-
-// Random number generator
-#include "particles/ionization/ionizationMethods.hpp"
 
 namespace picongpu
 {
@@ -98,21 +99,21 @@ private:
 
     PMACC_ALIGN(photon_mom, float3_X);
 
-    /* random number generator for Monte Carlo */
-    typedef ionization::RandomNrForMonteCarlo<T_ElectronSpecies> RandomGen;
-    /* \todo fix: cannot PMACC_ALIGN() because it seems to be too large */
+    /* random number generator */
+    typedef PMacc::random::RNGProvider<simDim, PMacc::random::methods::XorMin> RNGFactory;
+    typedef PMacc::random::distributions::Uniform<float> Distribution;
+    typedef typename RNGFactory::GetRandomType<Distribution>::type RandomGen;
     RandomGen randomGen;
 
 public:
     /* host constructor initializing member : random number generator */
     PhotonCreator(
         const SynchrotronFunctions::SyncFuncCursor& curF_1,
-        const SynchrotronFunctions::SyncFuncCursor& curF_2,
-        const uint32_t currentStep)
+        const SynchrotronFunctions::SyncFuncCursor& curF_2)
             : curF_1(curF_1),
               curF_2(curF_2),
-              randomGen(currentStep),
-              photon_mom(float3_X::create(0))
+              photon_mom(float3_X::create(0)),
+              randomGen(RNGFactory::createRandom<Distribution>())
     {
         DataConnector &dc = Environment<>::get().DataConnector();
         /* initialize pointers on host-side E-(B-)field databoxes */
@@ -159,7 +160,7 @@ public:
         /* wait for shared memory to be initialized */
         __syncthreads();
 
-        /* initialize random number generator with the local cell index in the simulation*/
+        /* initialize random number generator with the local cell index in the simulation */
         this->randomGen.init(localCellOffset);
     }
 
--- a/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
+++ b/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
@@ -190,7 +190,7 @@ public:
         const float_X mass = frame::getMass<FrameType>();
         const float_X charge = frame::getCharge<FrameType>();
 
-        const float_X sqrtOf3 = 1.732050807;
+        const float_X sqrtOf3 = 1.7320508075688772;
         const float_X factor = DELTA_T * charge*charge * mass * SPEED_OF_LIGHT / (float_X(4.0) * PI * EPS0 * HBAR*HBAR) *
                                sqrtOf3 / (float_X(2.0) * PI) * chi / gamma;
 
@@ -215,16 +215,17 @@ public:
      *
      * The scaling avoids an infrared divergence.
      *
-     * @param z scaled and normalized (to the electron energy) photon energy
+     * @param deltaScaled scaled and normalized (to the electron energy) photon energy
      * @param chi quantum-nonlinearity parameter
      * @param gamma electron gamma
      */
     DINLINE float_X emission_prob_scaled(
-        const float_X z,
+        const float_X deltaScaled,
         const float_X chi,
         const float_X gamma) const
     {
-        return float_X(3.0) * z*z * emission_prob(z*z*z, chi, gamma);
+        const float_X delta = deltaScaled*deltaScaled*deltaScaled;
+        return float_X(3.0) * deltaScaled*deltaScaled * emission_prob(delta, chi, gamma);
     }
 
     /** Return the number of target particles to create from each source particle.
@@ -242,7 +243,7 @@ public:
         PMACC_AUTO(particle, sourceFrame[localIdx]);
 
         /* particle position, used for field-to-particle interpolation */
-        floatD_X pos = particle[position_];
+        const floatD_X pos = particle[position_];
         const int particleCellIdx = particle[localCellIdx_];
         /* multi-dim coordinate of the local cell inside the super cell */
         DataSpace<TVec::dim> localCell(DataSpaceOperations<TVec::dim>::template map<TVec > (particleCellIdx));
@@ -255,6 +256,8 @@ public:
         ValueType_B fieldB = Field2ParticleInterpolation()
             (cachedB.shift(localCell).toCursor(), pos, fieldPosB());
 
+        /* All computation below is in the single "real" particle picture.
+         * The macroparticle weighting factor is reintroduced at the end of this code block. */
         const float3_X mom = particle[momentum_] / particle[weighting_];
         const float_X mom2 = math::dot(mom, mom);
         const float3_X mom_norm = mom * math::rsqrt(mom2);
@@ -263,28 +266,39 @@ public:
         const float_X gamma = Gamma<>()(mom, mass);
         const float3_X vel = mom / (gamma * mass); // low accuracy?
 
-        const float3_X lorentz = fieldE + math::cross(vel, fieldB);
-        const float_X lorentz2 = math::dot(lorentz, lorentz);
+        const float3_X lorentzForceOverCharge = fieldE + math::cross(vel, fieldB);
+        const float_X lorentzForceOverCharge2 = math::dot(lorentzForceOverCharge, lorentzForceOverCharge);
         const float_X fieldE_long = math::dot(mom_norm, fieldE);
 
-        // effective magnetic strength
-        const float_X H_eff = math::sqrt(lorentz2 - fieldE_long*fieldE_long);
+        // effective magnetic strength (in cgs)
+        const float_X H_eff = math::sqrt(lorentzForceOverCharge2 - fieldE_long*fieldE_long);
 
         const float_X charge = math::abs(frame::getCharge<FrameType>());
 
         const float_X c = SPEED_OF_LIGHT;
-        // Schwinger limit, unit: V/m
+        // Schwinger limit, unit: V/m (in cgs)
         const float_X E_S = mass*mass * c*c*c / (charge * HBAR);
         // quantum-nonlinearity parameter
         const float_X chi = gamma * H_eff / E_S;
 
-        const float_X z = this->randomGen();
+        const float_X deltaScaled = this->randomGen();
 
-        const float_X x = emission_prob_scaled(z, chi, gamma);
+        const float_X x = emission_prob_scaled(deltaScaled, chi, gamma);
+
+        // raise a warning if the emission probability is too high.
+        if(picLog::log_level & picLog::CRITICAL::lvl)
+        {
+            if(x > float_X(SINGLE_EMISSION_PROB_LIMIT))
+            {
+                printf("[SynchrotronPhotons] warning: emission probability is too high: p = %g, at delta = %g, chi = %g, gamma = %g\n",
+                    x, delta, chi, gamma);
+            }
+        }
 
         if(this->randomGen() < x)
         {
-            const float_X photonMom_abs = z*z*z * mass*c * gamma;
+            const float_X delta = deltaScaled*deltaScaled*deltaScaled;
+            const float_X photonMom_abs = delta * mass*c * gamma;
             if(photonMom_abs > SOFT_PHOTONS_CUTOFF_MOM)
             {
                 this->photon_mom = mom_norm * photonMom_abs * particle[weighting_];
--- a/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
+++ b/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
@@ -31,6 +31,7 @@
 #include "traits/frame/GetCharge.hpp"
 #include "particles/operations/Assign.hpp"
 #include "particles/operations/Deselect.hpp"
+#include "particles/traits/ResolveAliasFromSpecies.hpp"
 
 #include "random/methods/XorMin.hpp"
 #include "random/distributions/Uniform.hpp"
@@ -74,9 +75,10 @@ struct PhotonCreator
     typedef typename ElectronSpecies::FrameType FrameType;
 
     /* specify field to particle interpolation scheme */
-    typedef typename PMacc::traits::Resolve<
-        typename GetFlagType<FrameType,interpolation<> >::type
-        >::type Field2ParticleInterpolation;
+    typedef typename PMacc::particles::traits::ResolveAliasFromSpecies<
+        ElectronSpecies,
+        interpolation<>
+    >::type Field2ParticleInterpolation;
 
     /* margins around the supercell for the interpolation of the field on the cells */
     typedef typename GetMargin<Field2ParticleInterpolation>::LowerMargin LowerMargin;
@@ -292,6 +294,7 @@ public:
         {
             if(x > float_X(SINGLE_EMISSION_PROB_LIMIT))
             {
+                const float_X delta = deltaScaled*deltaScaled*deltaScaled;
                 printf("[SynchrotronPhotons] warning: emission probability is too high: p = %g, at delta = %g, chi = %g, gamma = %g\n",
                     x, delta, chi, gamma);
             }
@@ -321,18 +324,19 @@ public:
     template<typename Electron, typename Photon>
     DINLINE void operator()(Electron& electron, Photon& photon) const
     {
-        photon[multiMask_] = 1;
-        photon[momentum_] = this->photon_mom;
-        electron[momentum_] -= this->photon_mom;
         PMACC_AUTO(destPhoton,
-            deselect<
+            PMacc::particles::operations::deselect<
                 boost::mpl::vector<
                     multiMask,
                     momentum
                 >
             >(photon)
         );
-        particles::operations::assign( destPhoton, electron );
+        PMacc::particles::operations::assign( destPhoton, electron );
+
+        photon[multiMask_] = 1;
+        photon[momentum_] = this->photon_mom;
+        electron[momentum_] -= this->photon_mom;
     }
 };
 
--- a/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
+++ b/src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
@@ -1,0 +1,290 @@
+/**
+ * Copyright 2015-2016 Heiko Burau
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "simulation_defines.hpp"
+#include "SynchrotronFunctions.hpp"
+#include "pmacc_types.hpp"
+#include "algorithms/Gamma.hpp"
+#include "algorithms/math/defines/sqrt.hpp"
+#include "algorithms/math/defines/dot.hpp"
+#include "algorithms/math/defines/cross.hpp"
+#include "traits/frame/GetMass.hpp"
+#include "traits/frame/GetCharge.hpp"
+
+#include "traits/Resolve.hpp"
+#include "mappings/kernel/AreaMapping.hpp"
+
+#include "fields/FieldB.hpp"
+#include "fields/FieldE.hpp"
+
+#include "compileTime/conversion/TypeToPointerPair.hpp"
+#include "memory/boxes/DataBox.hpp"
+
+// Random number generator
+#include "particles/ionization/ionizationMethods.hpp"
+
+namespace picongpu
+{
+namespace particles
+{
+namespace synchrotronPhotons
+{
+
+/**
+ *
+ * \tparam T_ElectronSpecies
+ * \tparam T_PhotonSpecies
+ */
+template<typename T_ElectronSpecies, typename T_PhotonSpecies>
+struct PhotonCreator
+{
+    typedef T_ElectronSpecies ElectronSpecies;
+    typedef T_PhotonSpecies PhotonSpecies;
+
+    typedef typename ElectronSpecies::FrameType FrameType;
+
+    /* specify field to particle interpolation scheme */
+    typedef typename PMacc::traits::Resolve<
+        typename GetFlagType<FrameType,interpolation<> >::type
+        >::type Field2ParticleInterpolation;
+
+    /* margins around the supercell for the interpolation of the field on the cells */
+    typedef typename GetMargin<Field2ParticleInterpolation>::LowerMargin LowerMargin;
+    typedef typename GetMargin<Field2ParticleInterpolation>::UpperMargin UpperMargin;
+
+    /* relevant area of a block */
+    typedef SuperCellDescription<
+        typename MappingDesc::SuperCellSize,
+        LowerMargin,
+        UpperMargin
+        > BlockArea;
+
+    BlockArea BlockDescription;
+
+    typedef MappingDesc::SuperCellSize TVec;
+
+    typedef FieldE::ValueType ValueType_E;
+    typedef FieldB::ValueType ValueType_B;
+
+private:
+    /* global memory EM-field device databoxes */
+    PMACC_ALIGN(eBox, FieldE::DataBoxType);
+    PMACC_ALIGN(bBox, FieldB::DataBoxType);
+    /* shared memory EM-field device databoxes */
+    PMACC_ALIGN(cachedE, DataBox<SharedBox<ValueType_E, typename BlockArea::FullSuperCellSize,1> >);
+    PMACC_ALIGN(cachedB, DataBox<SharedBox<ValueType_B, typename BlockArea::FullSuperCellSize,0> >);
+
+    PMACC_ALIGN(curF_1, SynchrotronFunctions::SyncFuncCursor);
+    PMACC_ALIGN(curF_2, SynchrotronFunctions::SyncFuncCursor);
+
+    PMACC_ALIGN(photon_mom, float3_X);
+
+    /* random number generator for Monte Carlo */
+    typedef ionization::RandomNrForMonteCarlo<T_ElectronSpecies> RandomGen;
+    /* \todo fix: cannot PMACC_ALIGN() because it seems to be too large */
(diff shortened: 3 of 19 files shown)


PrometheusPi:
Looks great :+1: 

Could you check for two magnetic field strength, two electron energies and two time step duration?

PrometheusPi:
@Heikman Could you add your name to the license header of the following files:
 - `src/picongpu/include/particles/ParticlesFunctors.hpp`
 - `src/picongpu/include/simulation_defines/param/speciesAttributes.param`
 - `src/picongpu/include/simulation_defines/param/speciesDefinition.param `

bussmann:
Wonderful!

psychocoderHPC:
I readded the flag `WIP`
- please check the compiler issues
- please remove all compiler **warning** before this pull request can be accepted

src/picongpu/include/simulationControl/MySimulation.hpp
```diff
@@ -332,6 +333,9 @@ class MySimulation : public SimulationHelper<simDim>
 
         /* add CUDA streams to the StreamController for concurrent execution */
         Environment<>::get().StreamController().addStreams(6);
+
+        // Initialize synchrotron functions
+        this->synchrotronFunctions.init();
```
psychocoderHPC: Please avoid that synchrotron function creates memory if we not calculate/use them.

src/picongpu/include/particles/ParticlesFunctors.hpp
```diff
@@ -317,6 +321,63 @@ struct CallIonization
 
 }; // struct CallIonization
 
-} //namespace particles
+/** Handles the synchrotronPhotons effect for electrons.
```
PrometheusPi: `synchrotronPhotons effect` could you specify what you mean with that?

src/picongpu/include/particles/ParticlesFunctors.hpp
```diff
@@ -317,6 +321,63 @@ struct CallIonization
...
+                            T_StorageTuple& tuple,
+                            T_CellDescription* cellDesc,
+                            const uint32_t currentStep,
+                            const synchrotronPhotons::SynchrotronFunctions& synchrotronFunctions) const
+    {
+        typedef typename SelectedPhotonCreator::PhotonSpecies PhotonSpecies;
+
+        /* get E- and B-field */
+        DataConnector &dc = Environment<>::get().DataConnector();
+        BOOST_AUTO(fieldE,
```
PrometheusPi: where do you use `fieldE` in this method?

src/picongpu/include/particles/ParticlesFunctors.hpp
```diff
@@ -317,6 +321,63 @@ struct CallIonization
...
+                 dc.getData<FieldE>(FieldE::getName(), true).getGridBuffer().
+                 getDeviceBuffer().cartBuffer());
+        BOOST_AUTO(fieldB,
```
PrometheusPi: where do you use `fieldB` in this method?

PrometheusPi:
@ax3l  and @psychocoderHPC 
Would it make sense to put ` src/picongpu/include/particles/ParticlesFunctors.hpp` into a sub-directory and write every class/function into its own file **and** namespace?

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.def
```diff
@@ -0,0 +1,40 @@
...
+#pragma once
+
+namespace picongpu
+{
+namespace particles
+{
+namespace synchrotronPhotons
+{
+
+/**
```
PrometheusPi: Please write a short (and long) doxygen description to this class - not only give its template arguments.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+#include "particles/ionization/ionizationMethods.hpp"
+
+namespace picongpu
+{
+namespace particles
+{
+namespace synchrotronPhotons
+{
+
+/**
```
PrometheusPi: same here:
Please write a short (and long) doxygen description to this class - not only give its template arguments.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+    typedef typename ElectronSpecies::FrameType FrameType;
+
+    /* specify field to particle interpolation scheme */
+    typedef typename PMacc::traits::Resolve<
+        typename GetFlagType<FrameType,interpolation<> >::type
+        >::type Field2ParticleInterpolation;
+
+    /* margins around the supercell for the interpolation of the field on the cells */
+    typedef typename GetMargin<Field2ParticleInterpolation>::LowerMargin LowerMargin;
+    typedef typename GetMargin<Field2ParticleInterpolation>::UpperMargin UpperMargin;
```
PrometheusPi: Is this working with the recent changes on interpolation margins introduces in #1215 and #1339 ?

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+                  assign,
+                  cachedE,
+                  fieldEBlock
+                  );
+
+        /* wait for shared memory to be initialized */
+        __syncthreads();
+
+        /* initialize random number generator with the local cell index in the simulation*/
+        this->randomGen.init(localCellOffset);
```
PrometheusPi: Is this the original random number generator used in PIConGPU?
Would it make sense to use #1236 (by @Flamefire) instead?
The `seed` in `init(seed)` is for all runs the same. Could you add a dependency on [this seed](https://github.com/ComputationalRadiationPhysics/picongpu/blob/5f82fc19bb78ef0bb6ebe42db44a628454be5c94/src/picongpu/include/simulation_defines/param/seed.param#L40) or [this](https://github.com/ComputationalRadiationPhysics/picongpu/blob/5f82fc19bb78ef0bb6ebe42db44a628454be5c94/src/picongpu/include/simulation_defines/param/seed.param#L47).

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    DINLINE float_X emission_prob(
+        const float_X delta,
```
PrometheusPi: What `delta` is this (time, frequency, ...)? please add short description.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
Flamefire: The random number generator used is this one: https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/particles/ionization/ionizationMethods.hpp
It uses the original RNG from picongpu with with the ionisation seed (and a few others) set up here: https://github.com/ComputationalRadiationPhysics/picongpu/pull/1354/files#diff-c51358fcd763da76cbd1cdd86809d342R110 and https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/particles/ionization/ionizationMethods.hpp#L102

Short: The usage is correct. It would make sense to change the RNG in the class itself, but that change would be transparent and is for another PR

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+            return DELTA_T * charge*charge * mass*SPEED_OF_LIGHT / (float_X(4.0)*PI*EPS0 * HBAR*HBAR) *
+                float_X(1.732050807) / (float_X(2.0) * PI) * chi/gamma * (float_X(1.0) - delta) / delta *
+                (this->curF_1[z] + float_X(1.5) * delta * chi * z * this->curF_2[z]);
+        }
+        else
+        {
+            // classical
+            const float_X z = float_X(2.0/3.0) * delta / chi;
+
+            return DELTA_T * charge*charge * mass*SPEED_OF_LIGHT / (float_X(4.0)*PI*EPS0 * HBAR*HBAR) *
```
PrometheusPi: This looks very similar to the first part of the quantum part. 
Could you write something like:
```c++
float_X result = DELTA_T * charge*charge * mass*SPEED_OF_LIGHT / (float_X(4.0)*PI*EPS0 * HBAR*HBAR) *
             float_X(1.732050807) / (float_X(2.0) * PI) * chi/gamma / delta ;
if(enableQEDTerm) 
{
    \\ quantum
    const float_X z = float_X(2.0/3.0) * delta / ((float_X(1.0) - delta) * chi);

    result *= (float_X(1.0) - delta) * (this->curF_1[z] + float_X(1.5) * delta * chi * z * this->curF_2[z]);
}
else
{
    \\ classical 
    result *= this->curF_1[z];
}
return result;
```

to avoid hard to maintain double implementations.

PrometheusPi: and please use a const variable for the magic numbers `1.732050807`.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+                float_X(1.732050807) / (float_X(2.0) * PI) * chi/gamma / delta * this->curF_1[z];
+        }
+    }
+
+    DINLINE float_X emission_prob_modified(
```
PrometheusPi: please add doxygen documentation.

psychocoderHPC:
@PrometheusPi  IMO yes, could you please open an issue for that?

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+
+        ValueType_E fieldE;
+        ValueType_B fieldB;
+
+        /* particle position, used for field-to-particle interpolation */
+        floatD_X pos = particle[position_];
+        const int particleCellIdx = particle[localCellIdx_];
+        /* multi-dim coordinate of the local cell inside the super cell */
+        DataSpace<TVec::dim> localCell(DataSpaceOperations<TVec::dim>::template map<TVec > (particleCellIdx));
+        /* interpolation of E- */
```
PrometheusPi: what do you mean by `E-`?

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
psychocoderHPC: Yes, as I know he used only the interpolation therefore it is correct.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+        const fieldSolver::numericalCellType::traits::FieldPosition<FieldE> fieldPosE;
+        fieldE = Field2ParticleInterpolation()
+            (cachedE.shift(localCell).toCursor(), pos, fieldPosE());
+        /*                     and B-field on the particle position */
+        const fieldSolver::numericalCellType::traits::FieldPosition<FieldB> fieldPosB;
+        fieldB = Field2ParticleInterpolation()
+            (cachedB.shift(localCell).toCursor(), pos, fieldPosB());
+
+        const float3_X mom = particle[momentum_] / particle[weighting_];
```
PrometheusPi: Marija Vranic mentioned some dependencies on the weighting due to different powers in momentum, mass and charge. Do you get the same physics results for different weighting?
Why do you have to normalize to a single electron momentum?

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+        const float_X mom2 = math::dot(mom, mom);
+        const float3_X mom_norm = mom * math::rsqrt(mom2);
+        const float_X mass = frame::getMass<FrameType>();
+        const float_X mc = mass * SPEED_OF_LIGHT;
+
+        const float_X gamma = math::sqrt(float_X(1.0) + mom2 / (mc*mc));
+        const float3_X vel = mom / (gamma * mass); // low accuracy?
```
PrometheusPi: This assumes `SPEED_OF_LIGHT = 1`. Please avoid this assumption.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+
+        const float3_X lorentz = fieldE + math::cross(vel, fieldB);
+        const float_X lorentz2 = math::dot(lorentz, lorentz);
+        const float_X fieldE_long = math::dot(mom_norm, fieldE);
+
+        const float_X H_eff = math::sqrt(lorentz2 - fieldE_long*fieldE_long);
+
+        const float_X charge = math::abs(frame::getCharge<FrameType>());
+        const float_X c = SPEED_OF_LIGHT;
```
PrometheusPi: In the previous code lines you used `SPEED_OF_LIGHT`. If you move this definition up, you could avoid using `SPEED_OF_LIGHT` in line 235, ...

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+        // Schwinger limit, unit: V/m
+        const float_X E_S = mass*mass * c*c*c / (charge * HBAR);
+
+        const float_X chi = gamma * H_eff / E_S;
+
+        const float_X z = this->randomGen();
+        if(z == float_X(0.0) || z == float_X(1.0))
```
PrometheusPi: why do you need these two special cases?

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+        if(this->randomGen() < x)
+        {
+            this->photon_mom = mom_norm * z*z*z * mass*c * gamma * particle[weighting_];
+            return 1;
+        }
+
+        return 0;
+    }
+
+    /** Functor implementation.
```
PrometheusPi: please give a more detailed description

src/picongpu/include/particles/traits/GetPhotonCreator.hpp
```diff
@@ -0,0 +1,57 @@
+/**
+ * Copyright 2015 Heiko Burau
```
PrometheusPi: please update the year

src/picongpu/include/particles/traits/GetPhotonCreator.hpp
```diff
@@ -0,0 +1,57 @@
...
+#include "simulation_defines/unitless/speciesDefinition.unitless"
+
+#include "particles/synchrotronPhotons/PhotonCreator.def"
+#include "particles/synchrotronPhotons/PhotonCreator.hpp"
+
+namespace picongpu
+{
+
+template<typename T_SpeciesType>
+struct GetPhotonCreator
```
PrometheusPi: @psychocoderHPC Is placing this  trait in the `picongpu` namespace OK or should it be in a sub-namespace?

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
psychocoderHPC: please use https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/algorithms/Velocity.hpp

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+        fieldB = Field2ParticleInterpolation()
+            (cachedB.shift(localCell).toCursor(), pos, fieldPosB());
+
+        const float3_X mom = particle[momentum_] / particle[weighting_];
+        const float_X mom2 = math::dot(mom, mom);
+        const float3_X mom_norm = mom * math::rsqrt(mom2);
+        const float_X mass = frame::getMass<FrameType>();
+        const float_X mc = mass * SPEED_OF_LIGHT;
+
+        const float_X gamma = math::sqrt(float_X(1.0) + mom2 / (mc*mc));
```
psychocoderHPC: please use this [functor](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/algorithms/Gamma.hpp)

src/picongpu/include/simulationControl/MySimulation.hpp
```diff
@@ -428,6 +432,16 @@ class MySimulation : public SimulationHelper<simDim>
         ForEach<VectorSpeciesWithIonizer, particles::CallIonization<bmpl::_1>, MakeIdentifier<bmpl::_1> > particleIonization;
         particleIonization(forward(particleStorage), cellDescription, currentStep);
 
+        /* SynchrotronPhotons */
+        typedef typename PMacc::particles::traits::FilterByFlag<VectorAllSpecies,
+                                                                synchrotronPhotons<> >::type AllSynchrotronPhotonsSpecies;
```
PrometheusPi: This are all radiating particles (electrons) but **no photons**. Please rename this list to avoid confusion.

src/picongpu/include/simulationControl/MySimulation.hpp
```diff
@@ -428,6 +432,16 @@ class MySimulation : public SimulationHelper<simDim>
...
+
+        ForEach<AllSynchrotronPhotonsSpecies,
+                particles::CallSynchrotronPhotons<bmpl::_1>,
+                MakeIdentifier<bmpl::_1> > electronSynchrotronPhotons;
+        electronSynchrotronPhotons(forward(particleStorage), cellDescription, currentStep, this->synchrotronFunctions);
```
PrometheusPi: If you make `this->synchrotronFunctions` only available if `electronSynchrotronPhotons` is not the empty list, is it still required to resolve `this->synchrotronFunctions`? If so, you need to deal with this here via a `#ifdefine` or similar.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+     *
+     * Called once for each single particle creation.
+     *
+     * \tparam Electron type of electron which creates the photon
+     * \tparam Photon type of photon that is created
+     */
+    template<typename Electron, typename Photon>
+    DINLINE void operator()(Electron& electron, Photon& photon) const
+    {
+        photon[multiMask_] = 1;
```
psychocoderHPC: please use something like this:

```C++
#include "particles/operations/Assign.hpp"
#include "particles/operations/Deselect.hpp"
...

photon[multiMask_] = 1;
photon[momentum_] = this->photon_mom;
electron[momentum_] -= this->photon_mom;
PMACC_AUTO( destPhoton, 
    deselect<
        boost::mpl::vector<
            multiMask,
            momentum        
        >(photon) );
particles::operations::assign( destPhoton, electron );
```

Than
- all attributes from electron were copied to phothon
- it is save whether electron has more attributes than phothon
- if the phothon has a attribute which is defined in electron this one is set to the user defined default value

PrometheusPi:
@psychocoderHPC opened an issue #1359

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
Heikman: Because I consider the macroelectron as an ensemble of electrons for which the synchrotron emission is calculated for a single electron and then the result is applied to the others.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
Heikman: 1. There is no speed of light in this formula
2. I really dont need precision here because the whole algorithm assumes highly relativistic electrons.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
Heikman: zero yields to infinite probability for omega-zero photons and one corresponds to the cutoff frequency.

Flamefire: Note that `z==1` will always be false as the range for the RNGs is typically [0,1)

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
PrometheusPi: Sounds good. Do you see any difference due to different  weighting? (Could be parts of your test block next week.)

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
PrometheusPi: @Heikman You are right. The speed is:
```
p = v * gamma * m
thus 
v = p / (gamma * m) 
```
:+1:

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
PrometheusPi: @Heikman Thank you for the  explanation. Could you add a comment?

@Flamefire I would avoid magic numbers as `1e-9`.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,289 @@
...
+        const float_X charge) const
+    {
+
+        if(enableQEDTerm)
+        {
+            // quantum
+            const float_X z = float_X(2.0/3.0) * delta / ((float_X(1.0) - delta) * chi);
+
+            return DELTA_T * charge*charge * mass*SPEED_OF_LIGHT / (float_X(4.0)*PI*EPS0 * HBAR*HBAR) *
+                float_X(1.732050807) / (float_X(2.0) * PI) * chi/gamma * (float_X(1.0) - delta) / delta *
```
PrometheusPi: offline discussion with @Heikman and @Flamefire:
if `delta_t` is too small `1/delta_t` will be `NaN`.

If the smalest number the random number generator produces, next to zero, is `EPSILON ~ 1e-7`, this should be save, since `1 / EPSILON ~ 1e7` can be represented in float. 

Please check the output of the random number generator.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
Flamefire: Problem I have seen is that further down in the function you have `xxx/z^3` if `z` would be a very small positive number, then it is possible, that `z^3` gets rounded to 0 resulting in infinity. However in this case(!) the smallest number from the (CUDA uniform) rng is ~`2^-32` which is rougly 10^-10 and the smallest normalized floating point number is ~10^-38 which means, that it fits.

PrometheusPi:
@Heikman Did this pull request depend on #1398?

Heikman:
Basically yes, but somehow the problem didn't arise when this PR was created.

PrometheusPi:
@Heikman what is the status on this pull request?

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
Heikman: There is no physical reason to take photon properties from the incident electron. So if there was a new attribute in the future which both have, there is no reason that this attribute should be copied. Additionally if there is an electron attribute not available for photons what happens then?

Heikman:
there have been accumulated quite a lot of commits, so please tell me if you need a rebase ;)
I definitely have to rebase before merging.

PrometheusPi:
@Heikman How is the rebase going?

Heikman:
I will move the last commit into a new pull request.

PrometheusPi:
@Heikman Is this pull request now "final" and can I review it again or do you need to add/remove some more code?

PrometheusPi:
*offline discussion:* @Heikman wants  to perform some more tests before the code review starts.

Heikman:
good news: the test went well.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
psychocoderHPC: If the user add `IDs` for the photon but not for the electron than yout current version of the code is not running. 
The suggested code snipped takes care that `unknown` attributes can be added to `photon` and are initialized correct

psychocoderHPC:
Please implement [this](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1354#discussion_r51141881) before merging

Heikman:
I've just found a NaN-photon in the last test run with 2M photons.

PrometheusPi:
@Heikman Do you already have an idea what might caused the `NaN`?

Heikman:
not yet

src/picongpu/include/simulationControl/MySimulation.hpp
```diff
@@ -335,6 +341,17 @@ class MySimulation : public SimulationHelper<simDim>
 
         /* add CUDA streams to the StreamController for concurrent execution */
         Environment<>::get().StreamController().addStreams(6);
+
+        // create factory for the random number generator
+        this->rngFactory = new RNGFactory(Environment<simDim>::get().SubGrid().getLocalDomain().size);
```
psychocoderHPC: This should contain also the guards

src/picongpu/include/simulationControl/MySimulation.hpp
```diff
@@ -335,6 +341,17 @@ class MySimulation : public SimulationHelper<simDim>
...
+        // init factory
+        PMacc::GridController<simDim>& gridCon = PMacc::Environment<simDim>::get().GridController();
+        this->rngFactory->init(gridCon.getScalarPosition());
+
+#if(ENABLE_SYNCHROTRON_PHOTONS == 1)
```
psychocoderHPC: Could you please use the trait `hasFlag()` and scan all species if the feature is enabled and depending on this result initialize the `synchrotronFunctions()`.

This avoid code flow control with macros.

src/picongpu/include/simulation_defines/param/synchrotronPhotons.param
```diff
@@ -24,9 +24,13 @@ namespace picongpu
 {
 namespace particles
 {
-namespace bremsstrahlung
+namespace synchrotronPhotons
 {
 
+#define ENABLE_SYNCHROTRON_PHOTONS 0
```
psychocoderHPC: documentation is missing

psychocoderHPC:
Please do not merge before the RNG (rando number generator) has the option that allow to avoid that the RNG state is always stored in the global memory. This would reduce the GPU memory in all cases, also if RNG is not used.

I am working on the new feature.

Heikman:
The NaN problem is solved now.

psychocoderHPC:
Please let the world know how you solved the `NaN` problem.

Heikman:
As the energy of a photon decreases its momentum is eventually clipped to zero by numerical rounding. This causes a division by zero in the photon pusher.

bussmann:
Did it involve slaying a princess and kissing a dragon?

Heikman:
1sec too late Micha^^

bussmann:
No dragons, sigh...

PrometheusPi:
@Heikman Please add your name to the license headers of all `physicalConstants.unitless`.

src/picongpu/include/particles/ParticlesFunctors.hpp
```diff
@@ -313,6 +318,52 @@ struct CallIonization
...
+     * \param tuple An n-tuple containing the type-info of multiple particle species
+     * \param cellDesc points to logical block information like dimension and cell sizes
+     * \param currentStep The current time step
+     */
+    template<typename T_StorageTuple, typename T_CellDescription>
+    HINLINE void operator()(
+                            T_StorageTuple& tuple,
+                            T_CellDescription* cellDesc,
+                            const uint32_t currentStep,
+                            const synchrotronPhotons::SynchrotronFunctions& synchrotronFunctions) const
```
PrometheusPi: The last parameter of this method is not mentioned/described in the doxygen description. Please add a description.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,319 @@
...
+
+    /* random number generator */
+    typedef PMacc::random::RNGProvider<simDim, PMacc::random::methods::XorMin> RNGFactory;
+    typedef PMacc::random::distributions::Uniform<float> Distribution;
+    typedef typename RNGFactory::GetRandomType<Distribution>::type RandomGen;
+    RandomGen randomGen;
+
+public:
+    /* host constructor initializing member : random number generator */
+    PhotonCreator(
```
PrometheusPi: The name of this methods does not suggest that it is just a initialization. Please think of a more suitable name.

PrometheusPi: Sorry my fault. :+1:

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,319 @@
...
+        const float_X gamma) const
+    {
+        // catch these special values because otherwise a NaN is returned whereas it should be a zero.
+        if(chi == float_X(0.0) || delta == float_X(0.0) || (float_X(1.0) - delta) == float_X(0.0))
+            return float_X(0.0);
+
+        const float_X mass = frame::getMass<FrameType>();
+        const float_X charge = frame::getCharge<FrameType>();
+
+        const float_X sqrtOf3 = 1.732050807;
```
PrometheusPi: please add more digits to be precise for `float64` as well.

PrometheusPi:
Please add a warning if the PIC time step is too large **or** add an adaptive event generator 😈

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,319 @@
...
+     * @return number of particle to be created from each source particle
+     */
+    DINLINE unsigned int numNewParticles(FrameType& sourceFrame, int localIdx)
+    {
+        using namespace PMacc::algorithms;
+
+        PMACC_AUTO(particle, sourceFrame[localIdx]);
+
+        /* particle position, used for field-to-particle interpolation */
+        floatD_X pos = particle[position_];
```
PrometheusPi: why is `pos` not `const`?

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,319 @@
...
+            (cachedE.shift(localCell).toCursor(), pos, fieldPosE());
+        /* interpolation of B-field on the particle position */
+        const fieldSolver::numericalCellType::traits::FieldPosition<FieldB> fieldPosB;
+        ValueType_B fieldB = Field2ParticleInterpolation()
+            (cachedB.shift(localCell).toCursor(), pos, fieldPosB());
+
+        const float3_X mom = particle[momentum_] / particle[weighting_];
+        const float_X mom2 = math::dot(mom, mom);
+        const float3_X mom_norm = mom * math::rsqrt(mom2);
+        const float_X mass = frame::getMass<FrameType>();
```
PrometheusPi: Please add a comment to this code block to emphasize that your values are considering a singe "real" electron and not a macro electron. (The calculation is correct.)

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,319 @@
...
+
+        const float_X gamma = Gamma<>()(mom, mass);
+        const float3_X vel = mom / (gamma * mass); // low accuracy?
+
+        const float3_X lorentz = fieldE + math::cross(vel, fieldB);
```
PrometheusPi: please rename: this is not the Lorentz force but the Lorentz force per charge.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,319 @@
...
+        const float_X lorentz2 = math::dot(lorentz, lorentz);
+        const float_X fieldE_long = math::dot(mom_norm, fieldE);
+
+        // effective magnetic strength
+        const float_X H_eff = math::sqrt(lorentz2 - fieldE_long*fieldE_long);
```
PrometheusPi: Since you are in SI units, could this be an `B_eff`? Otherwise there is a conversion factor of `H = 1/mu_0 B` in SI units.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,319 @@
...
+
+        const float_X charge = math::abs(frame::getCharge<FrameType>());
+
+        const float_X c = SPEED_OF_LIGHT;
+        // Schwinger limit, unit: V/m
+        const float_X E_S = mass*mass * c*c*c / (charge * HBAR);
```
PrometheusPi: This is cgs units es well. Thus your chi calculation is fine, since it is done entirely in cgs units (in contrast to the rest of the code, which uses SI units.)

PrometheusPi: Please add a comment that you are in cgs units or convert to SI units (more multiplications).

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,319 @@
...
+    /** Get the *scaled* photon emission probability
+     *
+     * The scaling avoids an infrared divergence.
+     *
+     * @param z scaled and normalized (to the electron energy) photon energy
+     * @param chi quantum-nonlinearity parameter
+     * @param gamma electron gamma
+     */
+    DINLINE float_X emission_prob_scaled(
+        const float_X z,
```
PrometheusPi: Please use a different variable name to avoid confusion with the z argument of the synchrotron functions.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,319 @@
...
+
+        const float_X charge = math::abs(frame::getCharge<FrameType>());
+
+        const float_X c = SPEED_OF_LIGHT;
+        // Schwinger limit, unit: V/m
+        const float_X E_S = mass*mass * c*c*c / (charge * HBAR);
+        // quantum-nonlinearity parameter
+        const float_X chi = gamma * H_eff / E_S;
+
+        const float_X z = this->randomGen();
```
PrometheusPi: Please use a different variable name to avoid confusion with the z argument of the synchrotron functions.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,319 @@
...
+
+        const float_X x = emission_prob_scaled(z, chi, gamma);
+
+        if(this->randomGen() < x)
+        {
+            const float_X photonMom_abs = z*z*z * mass*c * gamma;
```
PrometheusPi: Could you add a line
```c++
const float_X delta = z*z*z;
```
for clarity?

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,319 @@
...
+     */
+    DINLINE float_X emission_prob_scaled(
+        const float_X z,
+        const float_X chi,
+        const float_X gamma) const
+    {
+        return float_X(3.0) * z*z * emission_prob(z*z*z, chi, gamma);
+    }
+
+    /** Return the number of target particles to create from each source particle.
```
PrometheusPi: Please add to the description that this method **not only** counts the photons emitted (0 or 1) **but also sets the photon momentum**.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -0,0 +1,319 @@
...
+
+    /** Functor implementation: setting photon and electron properties
+     *
+     * Called once for each single particle creation.
+     *
+     * \tparam Electron type of electron which creates the photon
+     * \tparam Photon type of photon that is created
+     */
+    template<typename Electron, typename Photon>
+    DINLINE void operator()(Electron& electron, Photon& photon) const
```
PrometheusPi: This method could be called without calling `::numNewParticles` thus one might end up with a photon momentum of zero.

@psychocoderHPC @Heikman @ax3l Is it OK in PIConGPU to require these dependencies without checking them?

src/picongpu/include/particles/traits/GetPhotonCreator.hpp
```diff
@@ -0,0 +1,54 @@
...
+    typedef T_SpeciesType SpeciesType;
+    typedef typename SpeciesType::FrameType FrameType;
+
+    /* The following line only fetches the alias */
+    typedef typename GetFlagType<FrameType, synchrotronPhotons<> >::type FoundSynchrotronPhotonsAlias;
+
+    /* This now resolves the alias into the actual object type */
+    typedef typename PMacc::traits::Resolve<FoundSynchrotronPhotonsAlias>::type FoundPhotonSpecies;
+
+    /* This specifies the source species as the second template parameter of the photon creator */
```
PrometheusPi: Isn't the `source species` = `source species` and thus the **first** template parameter?

src/picongpu/include/particles/traits/GetPhotonCreator.hpp
```diff
@@ -0,0 +1,54 @@
...
+    typedef particles::synchrotronPhotons::PhotonCreator<T_SpeciesType, FoundPhotonSpecies> type;
```
PrometheusPi: Please avoid using the template argument `T_SpeciesType` directly in the code by using a `typedef` before. (not my idea but part of the (not yet documented) coding guideline).

src/picongpu/include/simulation_defines/param/physicalConstants.param
```diff
@@ -38,6 +38,11 @@ namespace picongpu
         BOOST_CONSTEXPR_OR_CONST float_64 EPS0_SI = 1.0 / MUE0_SI / SPEED_OF_LIGHT_SI
             / SPEED_OF_LIGHT_SI;
 
+        /** reduced Planck constant
+         * unit: J * s
+         */
+        BOOST_CONSTEXPR_OR_CONST float_64 HBAR_SI = 1.054571800e-34;
```
PrometheusPi: please use more digits (if possible)

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -71,6 +71,33 @@ AttributRadiationFlag
 
 /*########################### define species #################################*/
 
+/*--------------------------- photons -------------------------------------------*/
+
+#if(ENABLE_SYNCHROTRON_PHOTONS == 1)
```
PrometheusPi: see @psychocoderHPC comment before

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -71,6 +71,33 @@ AttributRadiationFlag
...
+
+value_identifier(float_X, DummyMass, 1.0);
+value_identifier(float_X, DummyCharge, 0.0);
```
PrometheusPi: why do you need these dummy values?

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -71,6 +71,33 @@ AttributRadiationFlag
...
+
+typedef bmpl::vector<
+    particlePusher<particles::pusher::Photon>,
+    shape<UsedParticleShape>,
+    interpolation<UsedField2Particle>,
+    current<UsedParticleCurrentSolver>,
```
PrometheusPi: Is the current solver called for photons as well (and just no current is returned due to  charge `=0`)?
Could you use a "None" current solver instead?

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -71,6 +71,33 @@ AttributRadiationFlag
...
+
+#if(ENABLE_SYNCHROTRON_PHOTONS == 1)
+
+value_identifier(float_X, DummyMass, 1.0);
+value_identifier(float_X, DummyCharge, 0.0);
+
+typedef bmpl::vector<
+    particlePusher<particles::pusher::Photon>,
+    shape<UsedParticleShape>,
+    interpolation<UsedField2Particle>,
```
PrometheusPi: Your pusher does not need a interpolation. Would this be necessary later for interactions with the density field?

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -85,6 +112,9 @@ typedef bmpl::vector<
     current<UsedParticleCurrentSolver>,
     massRatio<MassRatioElectrons>,
     chargeRatio<ChargeRatioElectrons>
+#if(ENABLE_SYNCHROTRON_PHOTONS == 1)
```
PrometheusPi: see @psychocoderHPC's  comment above

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -143,6 +173,9 @@ PIC_Ions
 typedef MakeSeq<
 Species1,
 Species2
+#if(ENABLE_SYNCHROTRON_PHOTONS == 1)
```
PrometheusPi: see @psychocoderHPC's comment above

src/picongpu/include/simulation_defines/param/synchrotronPhotons.param
```diff
@@ -36,6 +40,9 @@ BOOST_CONSTEXPR_OR_CONST float_64 SYNC_FUNCS_BESSEL_INTEGRAL_STEPWIDTH = 1.0e-3;
 /** Number of sampling points of the lookup table */
 BOOST_CONSTEXPR_OR_CONST uint32_t SYNC_FUNCS_NUM_SAMPLES = 8192;
 
-} // namespace bremsstrahlung
+/** Photons softer than this ratio: (2pi / omega) / DELTA_T, are neglected */
+BOOST_CONSTEXPR_OR_CONST float_64 SOFT_PHOTONS_CUTOFF_RATIO = 1.0;
```
PrometheusPi: Why do you use this cutoff frequency if you are using the *modified event generator*?

src/picongpu/include/simulation_defines/param/synchrotronPhotons.param
```diff
@@ -36,6 +40,9 @@ BOOST_CONSTEXPR_OR_CONST float_64 SYNC_FUNCS_BESSEL_INTEGRAL_STEPWIDTH = 1.0e-3;
 /** Number of sampling points of the lookup table */
 BOOST_CONSTEXPR_OR_CONST uint32_t SYNC_FUNCS_NUM_SAMPLES = 8192;
 
-} // namespace bremsstrahlung
+/** Photons softer than this ratio: (2pi / omega) / DELTA_T, are neglected */
```
PrometheusPi: This comment is not clear to me. What do you mean? `2 pi /  (omega DELTA_T)` has no unit. So you need to answer to what do I compare this unitless number 

Answer: if the frequency omega of your photon (`omega_phot`) is larger than `omega_phot > 2 pi * 1/(SOFT_PHOTONS_CUTOFF_RATIO * DELTA_T )` you create a photon.

src/picongpu/include/simulation_defines/unitless/synchrotronPhotons.unitless
```diff
@@ -35,6 +35,11 @@ BOOST_CONSTEXPR_OR_CONST float_64 SYNC_FUNCS_STEP_WIDTH =
  * up to infinity but in fact it is sufficient to integrate up to this constant. */
 BOOST_CONSTEXPR_OR_CONST float_64 SYNC_FUNCS_F1_INTEGRAL_BOUND = 50.0;
 
-} // namespace bremsstrahlung
+BOOST_CONSTEXPR_OR_CONST float_X SOFT_PHOTONS_CUTOFF_MOM = static_cast<float_X>(
```
PrometheusPi: please the discussion above for "complicated" implementation.

PrometheusPi:
The pull request looks good to me. Please add the minor remarks on comments and naming. And please resolve the conflicts and implement @psychocoderHPC remarks on particle attribute settings.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
Heikman: In my opinion this is an internal detail and not interesting for the user.
Here it's important to have a hint, that this method has to be called before `operator()`.

src/picongpu/include/simulation_defines/param/physicalConstants.param
Heikman: this is the best known value.

src/picongpu/include/simulation_defines/param/speciesDefinition.param
Heikman: because of #1208, #1209.

src/picongpu/include/simulation_defines/param/speciesDefinition.param
Heikman: Here, the precompiler statement is necessary because it's the only occurrence in the code. :)

src/picongpu/include/simulation_defines/param/speciesDefinition.param
Heikman: I assume so.

src/picongpu/include/simulation_defines/param/speciesDefinition.param
Heikman: Here, the precompiler statement is necessary because the `HasFlag` condition is based on a complete species list.

src/picongpu/include/simulation_defines/param/speciesDefinition.param
Heikman: Here, the precompiler statement is necessary because the `HasFlag` condition is based on a complete species list.

src/picongpu/include/simulation_defines/param/synchrotronPhotons.param
Heikman: This cutoff is to avoid bad numerical rounding and double counting. It has nothing to do with the event generator.

src/picongpu/include/simulation_defines/param/physicalConstants.param
PrometheusPi: OK :+1:

src/picongpu/include/simulation_defines/param/speciesDefinition.param
PrometheusPi: OK, then we have to postpone removing these dummy values. They are fine for now. :+1:

src/picongpu/include/simulation_defines/param/speciesDefinition.param
PrometheusPi: OK :+1:

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
Heikman: I implemented the code snippet.

src/picongpu/include/simulation_defines/param/speciesDefinition.param
PrometheusPi: *offline discussion with @Heikman and @Flamefire:*
Please check if the list
```c++
typedef typename PMacc::particles::traits::FilterByFlag<VectorAllSpecies,
+                                                                synchrotronPhotons<> >::type AllSynchrotronPhotonsSpecies;
```
is empty or not. If it is not empty, add photons.

src/picongpu/include/simulation_defines/param/speciesDefinition.param
PrometheusPi: *offline discussion with @Heikman and @Flamefire:*
Please check if the list
```c++
typedef typename PMacc::particles::traits::FilterByFlag<VectorAllSpecies,
+                                                                synchrotronPhotons<> >::type AllSynchrotronPhotonsSpecies;
```
is empty or not. If it is not empty, add photons.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -263,28 +266,39 @@ struct PhotonCreator
...
 
-        const float_X x = emission_prob_scaled(z, chi, gamma);
+        const float_X x = emission_prob_scaled(deltaScaled, chi, gamma);
+
+        // raise a warning if the emission probability is too high.
+        if(picLog::log_level & picLog::CRITICAL::lvl)
+        {
+            if(x > float_X(SINGLE_EMISSION_PROB_LIMIT))
+            {
+                printf("[SynchrotronPhotons] warning: emission probability is too high: p = %g, at delta = %g, chi = %g, gamma = %g\n",
```
PrometheusPi: @ComputationalRadiationPhysics/picongpu-maintainers 
Is the use of `printf` ok with you or should we use `log` here instead?

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
```diff
@@ -320,11 +322,17 @@ struct PhotonCreator
...
-        photon[weighting_] = electron[weighting_];
+        PMACC_AUTO(destPhoton,
+            deselect<
+                boost::mpl::vector<
+                    multiMask,
+                    momentum
+                >
+            >(photon)
+        );
+        particles::operations::assign( destPhoton, electron );
```
PrometheusPi: I think it would be saver to first copy the electron attributes to the photons (with `deselect` if you want to), and after that write the new attributes just in case someone alters your code and forgets to include the attribute set in the `deselect` list.

src/picongpu/include/simulation_defines/param/speciesDefinition.param
PrometheusPi: *offline discussion with @Heikman and @Flamefire:*
Please check if the list
```c++
typedef typename PMacc::particles::traits::FilterByFlag<VectorAllSpecies,
                                                                synchrotronPhotons<> >::type AllSynchrotronPhotonsSpecies;
```
is empty or not. If it is not empty, add photons.

PrometheusPi:
@Heikman Have you found the error that causes your code to not compile anymore so far?

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -143,7 +172,21 @@ PIC_Ions
 typedef MakeSeq<
 Species1,
 Species2
+>::type VectorSpeciesTmp;
+
+// If there are species containing the `synchrotronPhotons` alias, add photon species to `VectorAllSpecies`
+typedef typename PMacc::particles::traits::FilterByFlag<
```
psychocoderHPC: This reason why this not compile is that `FilterByFlag` is unknown at this point.

output of the compile suite:
```
>  /home/buildbot/build/params/KelvinHelmholtz/cmakePreset_4/include/simulation_defines/param/speciesDefinition.param(178): error: name followed by "::" must be a class or namespace name
>  
>  /home/buildbot/build/params/KelvinHelmholtz/cmakePreset_4/include/simulation_defines/param/speciesDefinition.param(179): error: expected an identifier
>  
>  /home/buildbot/build/params/KelvinHelmholtz/cmakePreset_4/include/simulation_defines/param/speciesDefinition.param(180): error: an explicit template argument list is not allowed on this declaration
>  
>  /home/buildbot/build/params/KelvinHelmholtz/cmakePreset_4/include/simulation_defines/param/speciesDefinition.param(181): error: expected a ";"
>  
>  /home/buildbot/build/params/KelvinHelmholtz/cmakePreset_4/include/simulation_defines/param/speciesDefinition.param(184): error: identifier "AllSynchrotronPhotonsSpecies" is undefined
>  
>  /home/buildbot/build/params/KelvinHelmholtz/cmakePreset_4/include/simulation_defines/param/speciesDefinition.param(188): error: identifier "PIC_Photons" is undefined
>  
```

PrometheusPi:
@psychocoderHPC Thanks for finding the bug. :+1:

PrometheusPi:
@Heikman:
*offline discussionc with @psychocoderHPC *:
Since refactoring the random number generator will need some more time, please check in `MySimulation` if any species has a `photonCreator` flag, and if this is the case initialize the random number generator (which allocates memory). Thus we avoid allocating a lot of memory all the time and a double structure from pre-compiler flags.

Heikman:
looks good to me, don't hesitate to merge :)

Heikman:
great pull request! :D

src/picongpu/include/simulation_defines/param/speciesDefinition.param
Heikman: Omitting this line causes a compile time error. I reopened #1207.

PrometheusPi: OK - with #1207 this is no longer in the scope of this pull request :+1:

PrometheusPi:
@Heikman There seems to be a self-praise filter for notifications 😄 I did not see your update in notifications. I will have a look and merge a.s.a.p.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
PrometheusPi: @Heikman I think you missed this.

Heikman:
Oh this is too bad, I just wanted to initiate the final phase of this pull request by quoting the common phrases... :smile:

PrometheusPi:
Please have a look at the comment [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1354/commits/431d0b618ece2073113de50ee3c4fd6f06f16cc3).

Heikman:
put it in the last commit (wrong name)

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
PrometheusPi: Please use 
```c++
log<picLog::CRITICAL > ("[SynchrotronPhotons] warning: emission probability is too high: p = %g, at delta = %g, chi = %g, gamma = %g\n",) % x % delta % chi % gamma;
```
instead

PrometheusPi:
Please have a second lock at the warnings [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1354/commits/23ca27fbc2ace6c5fb6c20d8c11680e6e7102d11) (see new comment with code example on how to use the log output).

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
PrometheusPi: Thank you for fixing this.

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
Heikman: By the use of `printf` we save the if-condition.

PrometheusPi: but you perform this if-condition in line 289! So please skip this remove this if-condition and use the log level directly instead.

Heikman: Please note that the if-condition in line 289 is optimized out (checked this).

src/picongpu/include/simulation_defines/param/speciesDefinition.param
PrometheusPi: here is still a pre-compiler flag - please use your boost empty list check instead

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
PrometheusPi: and the same if statement in `log` is not?

PrometheusPi:
There is still one pre-compiler if statement. Please remove that if possible.

PrometheusPi:
The rest looks good to me.

src/picongpu/include/simulation_defines/param/speciesDefinition.param
Heikman: But this would only move the inevitable precompiler flag.

PrometheusPi: that is true - accepted :+1:

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
PrometheusPi: or do you mean line 289 is optimized ou in such a way that line 291 is only part of the code if 289 is true?

PrometheusPi:
@Heikman Have you checked whether you memory consumption at run time if you have set `ENABLE_SYNCHROTRON_PHOTONS` to zero?  

(I just want to be sure that we really avoid initializing the memory of the random number generator.)

PrometheusPi:
@psychocoderHPC @ax3l any comments on using `printf` over `log<...>` (as discussed [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1354/commits/23ca27fbc2ace6c5fb6c20d8c11680e6e7102d11)).

src/picongpu/include/particles/synchrotronPhotons/PhotonCreator.hpp
PrometheusPi: as discussed with @psychocoderHPC *offline*:
since this runs on the device, `log` will not work (it uses `std::cout`) - thus using `printf` is the only option :+1:

PrometheusPi: suggestion by @psychocoderHPC :
Please compare the register memory for critical log level on and off.

src/picongpu/include/simulation_defines/param/speciesDefinition.param
PrometheusPi: *Offline* discussion with @psychocoderHPC:
However this pre-compiler flag is used as switch here. 
A user can add the attribute on its own to add photon-emitting-electrons. 
A pre-compiler flag is only used for cmake compile tests, not as switch from your param file.

Heikman:
memory consumption looks as expected.

Heikman:
I've restored the precompiler if-condition in accordance to @psychocoderHPC

PrometheusPi:
PIConGPU now supports QED recoil and photons.

PR #1367, opened 2016-02-10
rename `types.h` to "pmacc_types.hpp`

Flamefire:
Does what is suggested in #1355.
This is the PMacc part. For the picongpu part see the next PR

Issue #1355, opened 2016-01-27
rename `types.h` to "pmacc_types.hpp`


psychocoderHPC:
Currently the global  PMacc definitions are places in a `C` header files.
we should:
- rename it to a `C++` header (file ending `.hpp`)
- change the name to `pmacc_types` to avoid conflicts with system headers

BenjaminW3:
:+1:

ax3l:
yeees pls! :)

bussmann:
Noooo! Introduce more C header files and give them names of existing standard C header files!!!

ax3l:
Next step: try to get PRs merged upstream in glibc renaming those standard headers to "resolve conflicts" :guitar:

ax3l:
@psychocoderHPC just a side note: we will *not* install libPMacc without a prefix directory, since this is messy and will alwasy collide *somewhere*.

We can still rename the file, but users of PMacc should include it in a `pmacc/some_header.hpp` and `pmacc/pmacc_types.hpp` way.

ax3l:
closed with #1367

PR #1369, opened 2016-02-10
rename `types.h` to "pmacc_types.hpp`

Flamefire:
Does what is suggested in #1355.
This is the picongpu part. For the PMacc part see #1367


Flamefire:
Everything is in #1367 now

PR #1371, opened 2016-02-10
Doxygen: Update File Patterns

ax3l:
Update the file patterns for doxygen.

Run via:
```bash
cd doc
doxygen Doxyfile
```


doc/Doxyfile
```diff
@@ -95,47 +95,22 @@ WARN_LOGFILE           =
...
+                         *.h        \
+                         *.hpp      \
+                         *.def      \
+                         *.param    \
+                         *.unitless \
+                         *.loader   \
+                         *.md
 RECURSIVE              = YES
-EXCLUDE                = ../src/cuda_memtest \
+EXCLUDE                = ../thirdParty  \
```
psychocoderHPC: `INPUT` is set to `../src` therefore we can skip remove `../buildsystem`, `../thirdParty` and `../nbproject` from the exclude list.

doc/Doxyfile
```diff
@@ -95,47 +95,22 @@ WARN_LOGFILE           =
...
-                         *.m \
-                         *.mm \
-                         *.dox \
-                         *.py \
-                         *.f90 \
-                         *.f \
-                         *.vhd \
-                         *.vhdl \
-                         *.cu
+FILE_PATTERNS          = *.cu       \
```
psychocoderHPC: extension `*.tpp` is missing

psychocoderHPC: Should we add `*.cfg?`

ax3l: good catch, added!

doc/Doxyfile
ax3l: updated, ty!

doc/Doxyfile
ax3l: `*.cfg` and `*.tpl` and and all other bash/python scripts are currently not added. one could do so, but they are not complete right now either and become independent projects in many cases, too.

(also: python uses its own param syntax)

PR #1370, opened 2016-02-10
Split off HostDeviceBuffer from GridBuffer

Flamefire:
The GridBuffer had actually 2 functions: Combine a Host- with a DeviceBuffer and provide exchange methods. This is a design flaw.

Also: Sometimes you know, you never need exchange methods, but only want a simple Buffer, that is on host and device. The GridBuffer is overkill for that usecase.

This splits off the DoubleBuffering to a new class `HostDeviceBuffer`

- [x] Charge conservation test density

- [x] Heating test heating

Issue #194, opened 2014-02-11
Particle Tracking Support


ax3l:
We should implement a general work-flow for efficient particle tracking.

ax3l:
required features:

- new particle attribute (e.g. `idx`)
- `idx` initialization*
- a general "filter" plugin**

* I recommend using the "there will not be more than N particles per GPUs" estimation from `MySimulation`. By that, we can multiply at t=0 with the MPI_Rank (+ moving window "offset" of the rank) to get unique (but not contiguous) IDs *in parallel* during init time. having "holes" in the id's should not be a problem.

** that (multi) plugin should be simply called for a given period and filter via functors for "species", "attribute in range of ..." and similar.

ax3l:
The bit-wise ID-creation per GPU could look like this, following @psychocoderHPC (similar to @Flamefire 's seed-per-GPU implementation):

- take a `uint64_t` as global unique ID per particle (also unique over all species!)
- `uint64_t prefixID` shall be initial, global GPU-id where the particle is initially created
- `uint64_t nextID`: the next available (or last used) ID for each GPU for to-be-created particles (ionization e-, photons/e-/e+, ...) must be stored in global GPU RAM and in checkpoints

### t=0

Initially, `nextID` is zeroed + the **reversed bits** of `prefixID`, which fills the upper bits. This is similar to doing `maxUint64_t` mod `numGPUs` to get the start of the GPU-lokal ID counter.

### Example

- `numGPUs = 1.e5` (~6x Titan/ORNL)
- `maxParticlesPerGPU = 1.e9` (~100GByte RAM / GPU)

-> `max(prefixId) = 1.e5 = 17 bits`
-> free bits per GPU for each id = `47` -> 1.4e14 IDs per GPU
-> 1.4e14 IDs / `maxParticlesPerGPU` = 140e3

Before an ID is re-assigned to a new particle:
- 140'000 complete re-initializations of each GPU with fresh particles are possible (e.g., in moving window simulations) or generally
- 1.4e14 particles can be created per GPU

### Notes

- particles moving between GPUs keep their ID
- particles moved from one species to an other keep their ID (`e` -> `e_fast`; `C_neutral` -> `C_ion`; ...)
- that number can be increased efficiently with `atomicAllInc` or `adomicAdd + 1` (some `atomicInc` operations allow to define an upper bound for overflow handling, which would be great to avoid writing in the GPU's `prefix` mask)

[...]

PR #1373, opened 2016-02-11
Implement IdProvider for unique ids

Flamefire:
This can be used to create unique particle IDs as requested in #194 

As requested by @psychocoderHPC everything is done in a static context to allow setting the particle id to the next id as a default, e.g. when assigning particles from a species without an Id to one with an Id.

Getter and setter is implemented for checkpointing.

- [x] Requires #1370 as the HostDeviceBuffer is used instead of the (overkill) GridBuffer

### DIFF
--- a/src/libPMacc/include/particles/IdProvider.hpp
+++ b/src/libPMacc/include/particles/IdProvider.hpp
@@ -1,0 +1,142 @@
+/**
+ * Copyright 2016 Alexander Grund
+ *
+ * This file is part of libPMacc.
+ *
+ * libPMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * libPMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with libPMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "types.h"
+#include "Environment.hpp"
+#include "eventSystem/EventSystem.hpp"
+#include "algorithms/reverseBits.hpp"
+#include "nvidia/atomic.hpp"
+#include "memory/buffers/HostDeviceBuffer.hpp"
+
+namespace PMacc {
+
+    namespace IdDetail {
+
+        __device__ uint64_cu nextId;
+
+        __global__ void setNextId(uint64_cu id)
+        {
+            nextId = id;
+        }
+
+        template<class T_Box>
+        __global__ void getNextId(T_Box box)
+        {
+            box(0) = nextId;
+        }
+
+    }  // namespace IdDetail
+
+    /**
+     * Provider for globally unique (even across ranks) ids
+     * Implemented for use in static contexts
+     */
+    class IdProvider
+    {
+    public:
+        /** Initializes the state so it is read for use
+         */
+        static void init()
+        {
+            const uint64_t globalUniqueStartId = getStartId();
+            setNextId(globalUniqueStartId);
+        }
+
+        /** Sets the next id to a given value (e.g. after a restart)
+         */
+        static void setNextId(const uint64_t nextId)
+        {
+            __cudaKernel(IdDetail::setNextId)(1, 1)(nextId);
+        }
+
+        /** Returns the next id without changing it (e.g. for saving)
+         */
+        static uint64_t getNextId()
+        {
+            HostDeviceBuffer<uint64_cu, 1> nextIdBuf(DataSpace<1>(1));
+            __cudaKernel(IdDetail::getNextId)(1, 1)(nextIdBuf.getDeviceBuffer().getDataBox());
+            nextIdBuf.deviceToHost();
+            return nextIdBuf.getHostBuffer().getDataBox()(0);
+        }
+
+        /** Functor that returns a new id */
+        struct GetNewId
+        {
+            uint64_cu operator()() const
+            {
+                return getNewId();
+            }
+        };
+
+        /** Function that returns a new id */
+        HDINLINE static uint64_cu getNewId()
+        {
+#ifdef __CUDA_ARCH__
+            return nvidia::atomicAllInc(&IdDetail::nextId);
+#else
+            throw std::runtime_error("getNewId not implemented for host");
+#endif
+        }
+
+        /**
+         * Return true, if an overflow of the counter is detected and hence there might be duplicate ids
+         */
+        static bool isOverflown()
+        {
+            // Get current value
+            uint64_t nextId = getNextId();
+            // Get start value
+            uint64_t globalUniqueStartId = getStartId();
+
+            /* Start value contains globally unique bits in the high order bits
+             * Hence compare all high order bits till the last set one and check,
+             * if any of them differs from the start value --> overflow
+             */
+            // How far do we need to shift to get the highest bit
+            BOOST_STATIC_CONSTEXPR int bitsToHighest = sizeof(uint64_t) * CHAR_BIT - 1;
+            while(globalUniqueStartId)
+            {
+                // Compare highest bit
+                if(globalUniqueStartId >> bitsToHighest != nextId >> bitsToHighest)
+                    return true;
+                // Proceed to next bit
+                globalUniqueStartId <<= 1;
+            }
+            return false;
+        }
+
+    private:
+        static uint64_t getStartId()
+        {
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
Could you please test the id creation with LWFA or KHI?

ax3l:
@Flamefire thank you so much, great and very needed work! :rocket: 

We can make a test run next week or we wait for you to return from your well-deserved vacations!

src/libPMacc/include/particles/IdProvider.hpp
```diff
@@ -0,0 +1,161 @@
...
+        template<class T_Box, class T_GetNewId>
+        __global__ void getNewId(T_Box boxOut, T_GetNewId getNewId)
+        {
+            boxOut(0) = getNewId();
+        }
+
+    }  // namespace IdDetail
+
+    /**
+     * Provider for globally unique (even across ranks) ids
```
ax3l: maybe we can write: `globally unique idx (even across MPI ranks)`

src/libPMacc/include/particles/IdProvider.hpp
```diff
@@ -0,0 +1,161 @@
...
+     * Implemented for use in static contexts
+     */
+    class IdProvider
+    {
+    public:
+        /** Initializes the state so it is read for use
```
ax3l: typo: ready

src/libPMacc/include/particles/IdProvider.hpp
```diff
@@ -0,0 +1,161 @@
...
+            getNewIdHost();
+            // Reset to start value
+            setNextId(globalUniqueStartId);
+        }
+
+        /** Sets the next id to a given value (e.g. after a restart)
+         */
+        static void setNextId(const uint64_t nextId)
+        {
+            __cudaKernel(IdDetail::setNextId)(1, 1)(nextId);
```
ax3l: @psychocoderHPC do we need event handling here?

src/libPMacc/include/particles/IdProvider.hpp
```diff
@@ -0,0 +1,161 @@
...
+        /** Functor that returns a new id */
+        struct GetNewId
+        {
+            DINLINE uint64_cu operator()() const
+            {
+                return getNewId();
+            }
+        };
+
+        /** Function that returns a new id */
```
ax3l: pls add a bit more documentation, here the value is changed so in contrast to `getNextId()` this changes the state.

(Yes, in a non static design the difference would be obvious with `const` and non-const member functions.)

src/libPMacc/include/particles/IdProvider.hpp
```diff
@@ -0,0 +1,161 @@
...
+        {
+            __cudaKernel(IdDetail::setNextId)(1, 1)(nextId);
+        }
+
+        /** Returns the next id without changing it (e.g. for saving)
+         */
+        static uint64_t getNextId()
+        {
+            HostDeviceBuffer<uint64_cu, 1> nextIdBuf(DataSpace<1>(1));
+            __cudaKernel(IdDetail::getNextId)(1, 1)(nextIdBuf.getDeviceBuffer().getDataBox());
```
ax3l: @psychocoderHPC event handling needed?

src/libPMacc/include/particles/IdProvider.hpp
```diff
@@ -0,0 +1,161 @@
...
+            return false;
+        }
+
+    private:
+        static uint64_t getStartId()
+        {
+            uint64_t rank = Environment<>::get().GridController().getGlobalRank();
+
+            /* We put the rank into the upper bits to have the lower bits for counting up and still
+             * getting unique numbers. Reversing the bits instead of shifting gives some more room
```
ax3l: *are still getting* or *still get*

src/libPMacc/include/particles/IdProvider.hpp
```diff
@@ -0,0 +1,161 @@
...
+             * as the upper bits of the rank or often also zero
```
ax3l: typo: ~~or~~ are

src/libPMacc/include/particles/IdProvider.hpp
```diff
@@ -0,0 +1,161 @@
...
+             */
+            return reverseBits(rank);
+        }
+
+        static uint64_t getNewIdHost()
+        {
+            HostDeviceBuffer<uint64_cu, 1> newIdBuf(DataSpace<1>(1));
+            __cudaKernel(IdDetail::getNewId)(1, 1)(newIdBuf.getDeviceBuffer().getDataBox(), GetNewId());
```
ax3l: @psychocoderHPC event handling needed?

src/libPMacc/include/particles/IdProvider.hpp
psychocoderHPC: Event handling is  not needed here.

src/libPMacc/include/particles/IdProvider.hpp
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        /** Returns the next id without changing it (e.g. for saving)
+         */
+        static uint64_t getNextId()
+        {
+            HostDeviceBuffer<uint64_cu, 1> nextIdBuf(DataSpace<1>(1));
+            __cudaKernel(IdDetail::getNextId)(1, 1)(nextIdBuf.getDeviceBuffer().getDataBox());
+            nextIdBuf.deviceToHost();
+            return nextIdBuf.getHostBuffer().getDataBox()(0);
```
psychocoderHPC: `getDataBox()(0)` should result in a compile time error. `operator()` is only allowed with `DataSpace` like `getDataBox()(DataSpace<DIM1>(0))`. 
The `operator[]` should allow an int.
A better way is to write `return *nextIdBuf.getHostBuffer().getDataBox();` what means give me this first value.

src/libPMacc/include/particles/IdProvider.hpp
```diff
@@ -0,0 +1,161 @@
...
+#include "types.h"
+#include "Environment.hpp"
+#include "eventSystem/EventSystem.hpp"
+#include "algorithms/reverseBits.hpp"
+#include "nvidia/atomic.hpp"
+#include "memory/buffers/HostDeviceBuffer.hpp"
+
+namespace PMacc {
+
+    namespace IdDetail {
```
psychocoderHPC: namespaces must begin with lower case letters.

src/libPMacc/include/particles/IdProvider.hpp
```diff
@@ -0,0 +1,161 @@
...
+             */
+            return reverseBits(rank);
+        }
+
+        static uint64_t getNewIdHost()
+        {
+            HostDeviceBuffer<uint64_cu, 1> newIdBuf(DataSpace<1>(1));
+            __cudaKernel(IdDetail::getNewId)(1, 1)(newIdBuf.getDeviceBuffer().getDataBox(), GetNewId());
+            newIdBuf.deviceToHost();
+            return newIdBuf.getHostBuffer().getDataBox()(0);
```
psychocoderHPC: Some like in line 90. This should be a compile time error

src/libPMacc/include/particles/IdProvider.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+        __global__ void setNextId(uint64_cu id)
+        {
+            nextId = id;
+        }
+
+        template<class T_Box>
+        __global__ void getNextId(T_Box boxOut)
+        {
+            boxOut(0) = nextId;
```
psychocoderHPC: For a DataBox this should result in a compile time error because there should be no cast operator from `int` to `DataBox<DIM1>`

psychocoderHPC: I checked `PMacc::DataSpace<>` it looks like a implizit cast from `int -> DataSpace<DIM1>` is possible. I this should marked as bug.

src/libPMacc/include/particles/IdProvider.hpp
Flamefire: I think this is correct: _for doing this and (for) doing that_

ax3l: no prob with me, should be understandable.

src/libPMacc/include/particles/IdProvider.hpp
psychocoderHPC: No event handling is not needed

src/libPMacc/include/particles/IdProvider.hpp
psychocoderHPC: OK this is **no bug**, maybe we avoid that in the future.

src/libPMacc/include/particles/IdProvider.hpp
psychocoderHPC: No not needed

src/libPMacc/include/particles/IdProvider.hpp
Flamefire: Really? For 1D it should be totally ok. What is the advantage of `return nextIdBuf.getHostBuffer().getDataBox()(DataSpace<DIM1>(0));` over this?
The way it is right now, scalars are equivalent to 1D dataspaces. Sounds ok.

Flamefire:
Rebased and fixed issues mentioned here so far.

Also did a quick test on 1 rank that ids are working in XRT.

PR #1374, opened 2016-02-15
add asin to algorithms::math.

Heikman:
add asin to algorithms::math.

PR #1363, opened 2016-02-05
add log10 to algorithms::math

Heikman:
add log10 to algorithms::math

Issue #1209, opened 2015-10-23
compile time error if mass is missing


Heikman:
If a species doesn't specify a mass, it will cause a compile-time error in:
InitialiserController.hpp(129):
```
ForEach<VectorAllSpecies, LogOmegaP<> > logOmegaP;
logOmegaP();
```

PrometheusPi:
If fixed please rove line 78 in ` src/picongpu/include/simulation_defines/param/speciesDefinition.param`.

PR #1376, opened 2016-02-15
Feature: Calorimeter Plugin

Heikman:
Adds the particle ~~detector~~ calorimeter plugin.
See the [wiki entry](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/Plugin%3A-ParticleCalorimeter) for a detailed description.

Dependencies:
- [x] #1375 
- [x] #1374 
- [x] #1363 
- [x] #1393 
- [x] #1394 
- [x] #1435 

Tested with #1354. The spectrum has also been tested against this python script:
```python
import numpy as np
import h5py as h5
%matplotlib notebook

timestep = 1
species = "ph"

f = h5.File("<path/to/hdf5-calorimeter-file>")
calorimeter = np.array(f["/data/{}/calorimeter".format(timestep)])
unitSI = f["/data/{}/calorimeter".format(timestep)].attrs["unitSI"]
minEnergy = f["/data/{}/calorimeter".format(timestep)].attrs["minEnergy[keV]"]
maxEnergy = f["/data/{}/calorimeter".format(timestep)].attrs["maxEnergy[keV]"]
UNITCONV_Joule_to_keV = 6.241509e15
calorimeter *= unitSI * UNITCONV_Joule_to_keV

calorimeter = np.sum(calorimeter, axis=(1,2))
plt.plot(calorimeter[1:])

f = h5.File("<path/to/hdf5-output-file>")
p_unit = f["/data/{}/particles/{}/momentum/x".format(timestep, species)].attrs["sim_unit"]
px = np.array(f["/data/{}/particles/{}/momentum/x".format(timestep, species)])
py = np.array(f["/data/{}/particles/{}/momentum/y".format(timestep, species)])
weights = np.array(f["/data/{}/particles/{}/weighting".format(timestep, species)])
energies = np.sqrt(px**2 + py**2) * p_unit * 3e8 / weights * UNITCONV_Joule_to_keV

hist, bins = np.histogram(energies, bins=np.linspace(minEnergy, maxEnergy, len(calorimeter) - 2), weights=energies*weights)
plt.plot(hist)
```

### DIFF
--- a/doc/TBG_macros.cfg
+++ b/doc/TBG_macros.cfg
@@ -198,6 +198,11 @@ TBG_liveViewYZ="--<species>_liveView.period 1 --<species>_liveView.slicePoint 0.
 # Print the maximum charge deviation between particles and div E to textfile 'chargeConservation.dat':
 TBG_chargeConservation="--chargeConservation.period 100"
 
+# Particle calorimeter: (virtually) propagates and collects particles to infinite distance
+TBG_<species>_calorimeter="--<species>_calorimeter.period 100 --<species>_calorimeter.openingYaw 90 --<species>_calorimeter.openingPitch 30
+                        --<species>_calorimeter.numBinsEnergy 32 --<species>_calorimeter.minEnergy 10 --<species>_calorimeter.maxEnergy 1000
+                        --<species>_calorimeter.logScale"
+
 ################################################################################
 ## Section: Program Parameters
 ## This section contains TBG internal variables, often composed from required
--- a/src/picongpu/include/plugins/PluginController.hpp
+++ b/src/picongpu/include/plugins/PluginController.hpp
@@ -36,6 +36,7 @@
 #include "plugins/BinEnergyParticles.hpp"
 #include "plugins/ChargeConservation.hpp"
 #if(ENABLE_HDF5 == 1)
+#include "plugins/particleCalorimeter/ParticleCalorimeter.hpp"
 #include "plugins/PhaseSpace/PhaseSpaceMulti.hpp"
 #endif
 
@@ -173,6 +174,7 @@ private:
      , PngPlugin< Visualisation<bmpl::_1, PngCreator> >
 #endif
 #if(ENABLE_HDF5 == 1)
+      , ParticleCalorimeter<bmpl::_1>
       , PerSuperCell<bmpl::_1>
       , PhaseSpaceMulti<particles::shapes::Counter::ChargeAssignment, bmpl::_1>
 #endif
--- a/src/picongpu/include/plugins/particleCalorimeter/ParticleCalorimeter.hpp
+++ b/src/picongpu/include/plugins/particleCalorimeter/ParticleCalorimeter.hpp
@@ -1,0 +1,530 @@
+/**
+ * Copyright 2016 Heiko Burau
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "ParticleCalorimeterFunctors.hpp"
+#include "ParticleCalorimeter.kernel"
+
+#include "traits/PICToSplash.hpp"
+#include "plugins/ISimulationPlugin.hpp"
+#include "cuSTL/container/DeviceBuffer.hpp"
+#include "cuSTL/container/HostBuffer.hpp"
+#include "cuSTL/algorithm/kernel/Foreach.hpp"
+#include "cuSTL/cursor/MultiIndexCursor.hpp"
+#include "cuSTL/algorithm/mpi/Reduce.hpp"
+#include "cuSTL/algorithm/host/Foreach.hpp"
+#include "particles/policies/ExchangeParticles.hpp"
+#include "math/Vector.hpp"
+#include "algorithms/math.hpp"
+#include <boost/shared_ptr.hpp>
+
+/* libSplash data output */
+#include <splash/splash.h>
+#include <boost/filesystem.hpp>
+#include <string>
+#include <iostream>
+#include <fstream>
+#include <stdlib.h>
+
+namespace picongpu
+{
+using namespace PMacc;
+
+namespace po = boost::program_options;
+
+
+/** Virtual particle calorimeter plugin.
+ *
+ * (virtually) propagates and collects particles to infinite distance.
+ *
+ */
+template<class ParticlesType>
+class ParticleCalorimeter : public ISimulationPlugin
+{
+private:
+    std::string name;
+    std::string prefix;
+    std::string foldername;
+    uint32_t notifyPeriod;
+    MappingDesc* cellDescription;
+    std::ofstream outFile;
+    const std::string leftParticlesDatasetName;
+
+    uint32_t numBinsYaw;
+    uint32_t numBinsPitch;
+    uint32_t numBinsEnergy;
+    float_X minEnergy;
+    float_X maxEnergy;
+    bool logScale;
+    float_X openingYaw_deg;
+    float_X openingPitch_deg;
+    float_X maxYaw_deg;
+    float_X maxPitch_deg;
+
+    float_64 posYaw_deg;
+    float_64 posPitch_deg;
+
+    /* Rotated calorimeter frame */
+    float3_X calorimeterFrameVecX;
+    float3_X calorimeterFrameVecY;
+    float3_X calorimeterFrameVecZ;
+
+    typedef PMacc::container::DeviceBuffer<float_X, DIM3> DBufCalorimeter;
+    typedef PMacc::container::HostBuffer<float_X, DIM3> HBufCalorimeter;
+
+    /* device calorimeter buffer for a single gpu */
+    DBufCalorimeter* dBufCalorimeter;
+    /* device calorimeter buffer for all particles which have left the simulation volume  */
+    DBufCalorimeter* dBufLeftParsCalorimeter;
+    /* host calorimeter buffer for a single mpi rank */
+    HBufCalorimeter* hBufCalorimeter;
+    /* host calorimeter buffer for summation of all mpi ranks */
+    HBufCalorimeter* hBufTotalCalorimeter;
+
+public:
+    typedef CalorimeterFunctor<typename DBufCalorimeter::Cursor> MyCalorimeterFunctor;
+private:
+    typedef boost::shared_ptr<MyCalorimeterFunctor> MyCalorimeterFunctorPtr;
+    MyCalorimeterFunctorPtr calorimeterFunctor;
+
+    typedef boost::shared_ptr<PMacc::algorithm::mpi::Reduce<simDim> > AllGPU_reduce;
+    AllGPU_reduce allGPU_reduce;
+
+    void restart(uint32_t restartStep, const std::string restartDirectory)
+    {
+        if(this->notifyPeriod == 0)
+            return;
+
+        HBufCalorimeter hBufLeftParsCalorimeter(this->dBufLeftParsCalorimeter->size());
+
+        PMacc::GridController<simDim>& gridCon = PMacc::Environment<simDim>::get().GridController();
+        pmacc::CommunicatorMPI<simDim>& comm = gridCon.getCommunicator();
+        uint32_t rank = comm.getRank();
+
+        if(rank == 0)
+        {
+            splash::SerialDataCollector hdf5DataFile(1);
+            splash::DataCollector::FileCreationAttr fAttr;
+
+            splash::DataCollector::initFileCreationAttr(fAttr);
+            fAttr.fileAccType = splash::DataCollector::FAT_READ;
+
+            std::stringstream filename;
+            filename << restartDirectory << "/" << prefix << "_" << restartStep;
+
+            hdf5DataFile.open(filename.str().c_str(), fAttr);
+
+            splash::Dimensions dimensions;
+
+            hdf5DataFile.read(restartStep,
+                              this->leftParticlesDatasetName.c_str(),
+                              dimensions,
+                              &(*hBufLeftParsCalorimeter.origin()));
+
+            hdf5DataFile.close();
+
+            /* rank 0 divides and distributes the calorimeter to all ranks in equal parts */
+            uint32_t numRanks = gridCon.getGlobalSize();
+            using namespace lambda;
+            PMacc::algorithm::host::Foreach()(hBufLeftParsCalorimeter.zone(),
+                                              hBufLeftParsCalorimeter.origin(),
+                                              _1 = _1 / float_X(numRanks));
+        }
+
+        MPI_Bcast(&(*hBufLeftParsCalorimeter.origin()),
+                  hBufLeftParsCalorimeter.size().productOfComponents() * sizeof(float_X),
+                  MPI_CHAR,
+                  0, /* rank 0 */
+                  comm.getMPIComm());
+
+        *this->dBufLeftParsCalorimeter = hBufLeftParsCalorimeter;
+    }
+
+
+    void checkpoint(uint32_t currentStep, const std::string checkpointDirectory)
+    {
+        if(this->notifyPeriod == 0)
+            return;
+
+        HBufCalorimeter hBufLeftParsCalorimeter(this->dBufLeftParsCalorimeter->size());
+        HBufCalorimeter hBufTotal(hBufLeftParsCalorimeter.size());
+
+        hBufLeftParsCalorimeter = *this->dBufLeftParsCalorimeter;
+
+        /* mpi reduce */
+        using namespace lambda;
+        (*this->allGPU_reduce)(hBufTotal, hBufLeftParsCalorimeter, _1 + _2);
+        if(!this->allGPU_reduce->root())
+            return;
+
+        splash::SerialDataCollector hdf5DataFile(1);
+        splash::DataCollector::FileCreationAttr fAttr;
+
+        splash::DataCollector::initFileCreationAttr(fAttr);
+
+        std::stringstream filename;
+        filename << checkpointDirectory << "/" << prefix << "_" << currentStep;
+
+        hdf5DataFile.open(filename.str().c_str(), fAttr);
+
+        typename PICToSplash<float_X>::type SplashTypeX;
+
+        splash::Dimensions bufferSize(hBufTotal.size().x(),
+                                      hBufTotal.size().y(),
+                                      hBufTotal.size().z());
+
+        /* if there is only one energy bin, omit the energy axis */
+        uint32_t dimension = this->numBinsEnergy == 1 ? DIM2 : DIM3;
+        hdf5DataFile.write(currentStep,
+                           SplashTypeX,
+                           dimension,
+                           splash::Selection(bufferSize),
+                           this->leftParticlesDatasetName.c_str(),
+                           &(*hBufTotal.origin()));
+
+        hdf5DataFile.close();
+    }
+
+
+    void pluginLoad()
+    {
+        namespace pm = PMacc::math;
+        namespace pam = PMacc::algorithms::math;
+
+        if(this->notifyPeriod == 0)
+            return;
+
+        if(!(this->openingYaw_deg > float_X(0.0) && this->openingYaw_deg <= float_X(360.0)))
+        {
+            std::stringstream msg;
+            msg << "[Plugin] [" << this->prefix
+                << "] openingYaw has to be within (0, 360]."
+                << std::endl;
+            throw std::runtime_error(msg.str());
+        }
+        if(!(this->openingPitch_deg > float_X(0.0) && this->openingPitch_deg <= float_X(180.0)))
+        {
+            std::stringstream msg;
+            msg << "[Plugin] [" << this->prefix
+                << "] openingPitch has to be within (0, 180]."
+                << std::endl;
+            throw std::runtime_error(msg.str());
+        }
+        if(this->numBinsEnergy > 1 && this->maxEnergy <= this->minEnergy)
+        {
+            std::stringstream msg;
+            msg << "[Plugin] [" << this->prefix
+                << "] minEnergy has to be less than maxEnergy."
+                << std::endl;
+            throw std::runtime_error(msg.str());
+        }
+
+        this->maxYaw_deg = float_X(0.5) * this->openingYaw_deg;
+        this->maxPitch_deg = float_X(0.5) * this->openingPitch_deg;
+        /* convert units */
+        const float_64 minEnergy_SI = this->minEnergy * UNITCONV_keV_to_Joule;
+        const float_64 maxEnergy_SI = this->maxEnergy * UNITCONV_keV_to_Joule;
+        this->minEnergy = minEnergy_SI / UNIT_ENERGY;
+        this->maxEnergy = maxEnergy_SI / UNIT_ENERGY;
+
+        /* allocate memory buffers */
+        this->dBufCalorimeter = new DBufCalorimeter(this->numBinsYaw, this->numBinsPitch, this->numBinsEnergy);
+        this->dBufLeftParsCalorimeter = new DBufCalorimeter(this->dBufCalorimeter->size());
+        this->hBufCalorimeter = new HBufCalorimeter(this->dBufCalorimeter->size());
+        this->hBufTotalCalorimeter = new HBufCalorimeter(this->dBufCalorimeter->size());
+
+        /* fill calorimeter for left particles with zero */
+        this->dBufLeftParsCalorimeter->assign(float_X(0.0));
+
+        /* create mpi reduce algorithm */
+        PMacc::GridController<simDim>& con = PMacc::Environment<simDim>::get().GridController();
+        pm::Size_t<simDim> gpuDim = (pm::Size_t<simDim>)con.getGpuNodes();
+        zone::SphericZone<simDim> zone_allGPUs(gpuDim);
+        this->allGPU_reduce = AllGPU_reduce(new PMacc::algorithm::mpi::Reduce<simDim>(zone_allGPUs));
+
+        /* calculate rotated calorimeter frame from posYaw_deg and posPitch_deg */
+        const float_64 posYaw_rad = this->posYaw_deg * float_64(M_PI / 180.0);
+        const float_64 posPitch_rad = this->posPitch_deg * float_64(M_PI / 180.0);
+        this->calorimeterFrameVecY = float3_X(pam::sin(posYaw_rad) * pam::cos(posPitch_rad),
+                                              pam::cos(posYaw_rad) * pam::cos(posPitch_rad),
+                                              pam::sin(posPitch_rad));
+        /* If the y-axis is pointing exactly up- or downwards we need to define the x-axis manually */
+        if(pam::abs(this->calorimeterFrameVecY.z()) == float_X(1.0))
+        {
+            this->calorimeterFrameVecX = float3_X(1.0, 0.0, 0.0);
+        }
+        else
+        {
+            /* choose `calorimeterFrameVecX` so that the roll is zero. */
+            const float3_X vecUp(0.0, 0.0, -1.0);
+            this->calorimeterFrameVecX = pam::cross(vecUp, this->calorimeterFrameVecY);
+            /* normalize vector */
+            this->calorimeterFrameVecX /= pam::abs(this->calorimeterFrameVecX);
+        }
+        this->calorimeterFrameVecZ = pam::cross(this->calorimeterFrameVecX, this->calorimeterFrameVecY);
+
+        /* create calorimeter functor instance */
+        this->calorimeterFunctor = MyCalorimeterFunctorPtr(new MyCalorimeterFunctor(
+            this->maxYaw_deg * float_X(M_PI / 180.0),
+            this->maxPitch_deg * float_X(M_PI / 180.0),
+            this->numBinsYaw,
+            this->numBinsPitch,
+            this->numBinsEnergy,
+            this->logScale ? pam::log10(this->minEnergy) : this->minEnergy,
+            this->logScale ? pam::log10(this->maxEnergy) : this->maxEnergy,
+            this->logScale,
+            this->calorimeterFrameVecX,
+            this->calorimeterFrameVecY,
+            this->calorimeterFrameVecZ));
+
+        /* create folder for hdf5 files*/
+        Environment<simDim>::get().Filesystem().createDirectoryWithPermissions(this->foldername);
+
+        Environment<>::get().PluginConnector().setNotificationPeriod(this, this->notifyPeriod);
+    }
+
+
+    void pluginUnload()
+    {
+        if(this->notifyPeriod == 0)
+            return;
+
+        __delete(this->dBufCalorimeter);
+        __delete(this->dBufLeftParsCalorimeter);
+        __delete(this->hBufCalorimeter);
+        __delete(this->hBufTotalCalorimeter);
+    }
+
+
+    void writeToHDF5File(uint32_t currentStep)
+    {
+        splash::SerialDataCollector hdf5DataFile(1);
+        splash::DataCollector::FileCreationAttr fAttr;
+
+        splash::DataCollector::initFileCreationAttr(fAttr);
+
+        std::stringstream filename;
+        filename << this->foldername << "/" << this->prefix << "_" << currentStep;
+
+        hdf5DataFile.open(filename.str().c_str(), fAttr);
+
+        typename PICToSplash<float_X>::type SplashTypeX;
+        typename PICToSplash<float_64>::type SplashType64;
+        typename PICToSplash<bool>::type SplashTypeBool;
+
+        splash::Dimensions bufferSize(this->hBufTotalCalorimeter->size().x(),
+                                      this->hBufTotalCalorimeter->size().y(),
+                                      this->hBufTotalCalorimeter->size().z());
+
+        hdf5DataFile.write(currentStep,
+                           SplashTypeX,
+                           this->numBinsEnergy == 1 ? DIM2 : DIM3,
+                           splash::Selection(bufferSize),
+                           "calorimeter",
+                           &(*this->hBufTotalCalorimeter->origin()));
+
+        const float_64 unitSI = particles::TYPICAL_NUM_PARTICLES_PER_MACROPARTICLE * UNIT_ENERGY;
+
+        hdf5DataFile.writeAttribute(currentStep,
+                                    SplashType64,
+                                    "calorimeter",
+                                    "unitSI",
+                                    &unitSI);
+
+        hdf5DataFile.writeAttribute(currentStep,
+                                    SplashType64,
+                                    "calorimeter",
+                                    "posYaw[deg]",
+                                    &posYaw_deg);
+
+        hdf5DataFile.writeAttribute(currentStep,
+                                    SplashType64,
+                                    "calorimeter",
+                                    "posPitch[deg]",
+                                    &posPitch_deg);
+
+        hdf5DataFile.writeAttribute(currentStep,
+                                    SplashTypeX,
+                                    "calorimeter",
+                                    "maxYaw[deg]",
+                                    &this->maxYaw_deg);
+
+        hdf5DataFile.writeAttribute(currentStep,
+                                    SplashTypeX,
+                                    "calorimeter",
+                                    "maxPitch[deg]",
+                                    &this->maxPitch_deg);
+
+        if(this->numBinsEnergy > 1)
+        {
+            const float_64 minEnergy_SI = this->minEnergy * UNIT_ENERGY;
+            const float_64 maxEnergy_SI = this->maxEnergy * UNIT_ENERGY;
+            const float_64 minEnergy_keV = minEnergy_SI * UNITCONV_Joule_to_keV;
+            const float_64 maxEnergy_keV = maxEnergy_SI * UNITCONV_Joule_to_keV;
+
+            hdf5DataFile.writeAttribute(currentStep,
+                                        SplashType64,
+                                        "calorimeter",
+                                        "minEnergy[keV]",
+                                        &minEnergy_keV);
+
+            hdf5DataFile.writeAttribute(currentStep,
+                                        SplashType64,
+                                        "calorimeter",
+                                        "maxEnergy[keV]",
+                                        &maxEnergy_keV);
+
+            hdf5DataFile.writeAttribute(currentStep,
+                                        SplashTypeBool,
+                                        "calorimeter",
+                                        "logScale",
+                                        &this->logScale);
+        }
+
+        hdf5DataFile.close();
+    }
+
+public:
+    ParticleCalorimeter() :
+        name("ParticleCalorimeter: (virtually) propagates and collects particles to infinite distance"),
+        prefix(ParticlesType::FrameType::getName() + std::string("_calorimeter")),
+        foldername(prefix),
+        notifyPeriod(0),
+        cellDescription(NULL),
+        leftParticlesDatasetName("calorimeterLeftParticles"),
+        dBufCalorimeter(NULL),
+        dBufLeftParsCalorimeter(NULL),
+        hBufCalorimeter(NULL),
+        hBufTotalCalorimeter(NULL)
+    {
+        Environment<>::get().PluginConnector().registerPlugin(this);
+    }
+
+
+    void notify(uint32_t currentStep)
+    {
+        /* initialize calorimeter with already detected particles */
+        *this->dBufCalorimeter = *this->dBufLeftParsCalorimeter;
+
+        /* data is written to dBufCalorimeter */
+        this->calorimeterFunctor->setCalorimeterCursor(this->dBufCalorimeter->origin());
+
+        /* create kernel functor instance */
+        DataConnector &dc = Environment<>::get().DataConnector();
+        ParticlesType* particles = &(dc.getData<ParticlesType > (ParticlesType::FrameType::getName(), true));
+
+        ParticleCalorimeterKernel<typename ParticlesType::ParticlesBoxType,
+                                  MyCalorimeterFunctor>
+                                  particleCalorimeterKernel(particles->getDeviceParticlesBox(),
+                                                            *this->calorimeterFunctor);
+
+        /* create zone */
+        typedef typename MappingDesc::SuperCellSize SuperCellSize;
+
+        const PMacc::math::Int<simDim> coreBorderGuardSuperCells = this->cellDescription->getGridSuperCells();
+        const uint32_t guardSuperCells = this->cellDescription->getGuardingSuperCells();
+        const PMacc::math::Int<simDim> coreBorderSuperCells = coreBorderGuardSuperCells - 2*guardSuperCells;
+
+        /* this zone represents the core+border area with guard offset in unit of cells */
+        const zone::SphericZone<simDim> zone(
+            static_cast<PMacc::math::Size_t<simDim> >(coreBorderSuperCells * SuperCellSize::toRT()),
+            guardSuperCells * SuperCellSize::toRT());
+
+        /* kernel call */
+        algorithm::kernel::Foreach<SuperCellSize> foreach;
+        foreach(zone, cursor::make_MultiIndexCursor<simDim>(), particleCalorimeterKernel);
+
+        /* copy to host */
+        *this->hBufCalorimeter = *this->dBufCalorimeter;
+
+        /* mpi reduce */
+        using namespace lambda;
+        (*this->allGPU_reduce)(*this->hBufTotalCalorimeter, *this->hBufCalorimeter, _1 + _2);
+        if(!this->allGPU_reduce->root())
+            return;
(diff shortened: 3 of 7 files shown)


src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,537 @@
...
+            "number of bins for angle theta.")
+        ((this->prefix + ".numBinsPhi").c_str(), po::value<uint32_t > (&this->numBinsPhi)->default_value(64),
+            "number of bins for angle phi.")
+        ((this->prefix + ".numBinsEnergy").c_str(), po::value<uint32_t > (&this->numBinsEnergy)->default_value(1),
+            "number of bins for the energy spectrum. Disabled by default.")
+        ((this->prefix + ".minEnergy").c_str(), po::value<float_X > (&this->minEnergy)->default_value(float_X(0.0)),
+            "minimal detectable energy in keV.")
+        ((this->prefix + ".maxEnergy").c_str(), po::value<float_X > (&this->maxEnergy)->default_value(float_X(1.0e3)),
+            "maximal detectable energy in keV.")
+        ((this->prefix + ".logScale").c_str(), po::value<uint32_t > (&this->logScale)->default_value(0),
```
PrometheusPi: here you emulate a bool flag with an integer. Please use a bool flag directly. An example of such a flag is [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/da64aace46f634cb4bd96a6ecdde020cd2046e16/src/picongpu/include/plugins/radiation/Radiation.hpp#L198)

PrometheusPi:
@Heikman Is your hdf5/libSplash output compatible with openPMD?

PrometheusPi:
Please add your name to the license header of `PluginConnector.hpp`.

ax3l:
`The term histogram may be a bit misleading here, because it's not counting particles but collecting energy, so a bin holds a certain amount of energy`

What you are building is a [calorimeter](https://en.wikipedia.org/wiki/Calorimeter_%28particle_physics%29).

ax3l:
`The spectrum has also been tested against a python script.` pls add the script to the PR description, e.g., as `yourScript.py.txt`

doc/TBG_macros.cfg
```diff
@@ -198,6 +198,11 @@ TBG_liveViewYZ="--<species>_liveView.period 1 --<species>_liveView.slicePoint 0.
 # Print the maximum charge deviation between particles and div E to textfile 'chargeConservation.dat':
 TBG_chargeConservation="--chargeConservation.period 100"
 
+# ParticleDetector: detects particles at infinite distance
+TBG_<species>_detector="--<species>_detector.period 100 --<species>_detector.openingTheta 90 --<species>_detector.openingPhi 30
+                        --<species>_detector.numBinsEnergy 32 --<species>_detector.minEnergy 10 --<species>_detector.maxEnergy 1000
+                        --<species>_detector.logScale 1"
```
ax3l: use a boolean flag here. e.g., not set == lin, set == log

doc/TBG_macros.cfg
```diff
@@ -198,6 +198,11 @@ TBG_liveViewYZ="--<species>_liveView.period 1 --<species>_liveView.slicePoint 0.
 # Print the maximum charge deviation between particles and div E to textfile 'chargeConservation.dat':
 TBG_chargeConservation="--chargeConservation.period 100"
 
+# ParticleDetector: detects particles at infinite distance
```
ax3l: should read: `(virtually) propagates and collects particles to infinite distance`

the question is: does this only work for massless particles? write photons in that case to not obfuscate the use case.

src/libPMacc/include/pluginSystem/PluginConnector.hpp
```diff
@@ -180,6 +180,19 @@ namespace PMacc
             }
         }
 
+        template<typename Plugin>
+        Plugin* getPluginFromType()
```
ax3l: this chance should not be part of this PR but proposed/reviewed and merged beforehand

ccing @psychocoderHPC

src/picongpu/include/plugins/PluginController.hpp
```diff
@@ -35,6 +35,7 @@
 #include "plugins/PositionsParticles.hpp"
 #include "plugins/BinEnergyParticles.hpp"
 #include "plugins/ChargeConservation.hpp"
+#include "plugins/particleDetector/ParticleDetector.hpp"
```
ax3l: this is wrong.

as with radiation, this include should *only* be made if `HDF5` is available. otherwise you will crash all users that do not use HDF5, e.g., all people benchmarking the core code.

src/picongpu/include/plugins/particleDetector/DetectGuardParticles.def
```diff
@@ -0,0 +1,28 @@
...
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+namespace picongpu
+{
+
+struct DetectGuardParticles;
```
ax3l: to general problem to be sneaked into this PR.

pls propose independently so we can find a general, re-usable and well-documented solution.

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,537 @@
...
+#include <stdlib.h>
+
+namespace picongpu
+{
+using namespace PMacc;
+
+namespace po = boost::program_options;
+
+/**
+ *
```
ax3l: missing doxygen string

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
ax3l: +1, annotated the same above

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,537 @@
...
+
+        typename PICToSplash<float_X>::type SplashTypeX;
+
+        splash::Dimensions bufferSize(hBufTotal.size().x(),
+                                      hBufTotal.size().y(),
+                                      hBufTotal.size().z());
+
+        hdf5DataFile.write(currentStep,
+                           SplashTypeX,
+                           this->numBinsEnergy == 1 ? DIM2 : DIM3,
```
ax3l: I understand what you are doing here but it is in-place and obfuscated.

write this in a const variable above, add a comment string and explain that numBinsEnergy with a size larger then 1 means you will *actually do an energy binning* on not just a spatially/radially resolved energy-sum

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,537 @@
...
+        using namespace PMacc::math;
+        Size_t<simDim> gpuDim = (Size_t<simDim>)con.getGpuNodes();
+        zone::SphericZone<simDim> zone_allGPUs(gpuDim);
+        this->allGPU_reduce = AllGPU_reduce(new PMacc::algorithm::mpi::Reduce<simDim>(zone_allGPUs));
+
+        /* calculate rotated detector frame from posTheta and posPhi */
+        const float_64 posTheta_rad = this->posTheta * float_64(M_PI / 180.0);
+        const float_64 posPhi_rad = this->posPhi * float_64(M_PI / 180.0);
+        using namespace PMacc::algorithms::math;
+        this->detectorFrameVecY = float3_X(sin(posTheta_rad) * cos(posPhi_rad),
```
ax3l: should we use our `math::` wrappers at this point? I think so.

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,537 @@
...
+                                           cos(posTheta_rad) * cos(posPhi_rad),
+                                           sin(posPhi_rad));
+        /* If the y-axis is pointing exactly up- or downwards we need to define the x-axis manually */
+        if(abs(this->detectorFrameVecY) == float_X(1.0))
```
ax3l: `math::abs`

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,537 @@
...
+        this->dBufLeftParsHistogram = new DBufHistogram(this->dBufHistogram->size());
+        this->hBufHistogram = new HBufHistogram(this->dBufHistogram->size());
+        this->hBufTotalHistogram = new HBufHistogram(this->dBufHistogram->size());
+
+        /* fill histogram for left particles with zero */
+        this->dBufLeftParsHistogram->assign(float_X(0.0));
+
+        /* create mpi reduce algorithm */
+        PMacc::GridController<simDim>& con = PMacc::Environment<simDim>::get().GridController();
+        using namespace PMacc::math;
```
ax3l: I am not a big fan of `using namespace` in the middle of a very long member function.

either define a new namespace and use that (`namespace pm = PMacc::math` ?) or do not do so at all. additionally, limit the scope of this namespace usage, e.g., use `{}` blocks or put the first suggestion on top of the method.

src/libPMacc/include/pluginSystem/PluginConnector.hpp
psychocoderHPC: Yes please move this to an own pull feature request and add a description.

This method only return the first plugin of the type. Is it possible to have more than one plugin with the given type in list  named `plugins`?

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,537 @@
...
+        Size_t<simDim> gpuDim = (Size_t<simDim>)con.getGpuNodes();
+        zone::SphericZone<simDim> zone_allGPUs(gpuDim);
+        this->allGPU_reduce = AllGPU_reduce(new PMacc::algorithm::mpi::Reduce<simDim>(zone_allGPUs));
+
+        /* calculate rotated detector frame from posTheta and posPhi */
+        const float_64 posTheta_rad = this->posTheta * float_64(M_PI / 180.0);
+        const float_64 posPhi_rad = this->posPhi * float_64(M_PI / 180.0);
+        using namespace PMacc::algorithms::math;
```
ax3l: double usage, better define a
```cpp
namespace pm = PMacc::math;
namespace pma = PMacc::algorithms::math;
```
at the beginning of this method

ax3l:
nice plugin and well written, please work through the review comments carefully! :sparkles:

src/libPMacc/include/pluginSystem/PluginConnector.hpp
Heikman: I opened a new PR: #1393

Heikman:
a test script can be found in the wiki.

src/libPMacc/include/pluginSystem/PluginConnector.hpp
```diff
@@ -180,6 +181,19 @@ namespace PMacc
...
+        {
+            for (std::list<IPlugin*>::iterator iter = plugins.begin();
+                    iter != plugins.end(); ++iter)
```
PrometheusPi: please fix indentation

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,542 @@
...
+    uint32_t numBinsTheta;
+    uint32_t numBinsPhi;
+    uint32_t numBinsEnergy;
+    float_X minEnergy;
+    float_X maxEnergy;
+    bool logScale;
+    float_X openingTheta;
+    float_X openingPhi;
+    float_X maxTheta;
+    float_X maxPhi;
```
PrometheusPi: You are using a fixed relation between the two quantities here `maxPhi = 0.5 * openingPhi`.
Is it thus necessary to store both as class member variables?

Would it be better to just store one of these variables as member and recalculate the other on a member function scope?
e.g.
```
template<class AngleType> 
AngleType getMaxAngle(AngleType openingAngle)
{
    return openingAngle * AngleType(0.5);
}

void someMemeberfunction()
{
    ...
    const float_X maxPhi = getMaxAngle(openingPhi);
    const float_X maxTheta = getMaxAngle(openingTheta);
    ...
}


@psychocoderHPC what do you think as a future maintainer of this code?

File: src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp [hunk 15]
```diff
@@ -0,0 +1,542 @@
...
+                  0,
+                  MPI_COMM_WORLD);
+
+        *this->dBufLeftParsCalorimeter = hBufLeftParsCalorimeter;
+    }
+
+
+    void checkpoint(uint32_t currentStep, const std::string checkpointDirectory)
+    {
+        if(this->notifyPeriod == 0)
```
PrometheusPi: Is it essential that your checkpoint period is equal to your data output period? If not, please only write backups at checkpoint intervals.

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,542 @@
...
+
+        if(this->notifyPeriod == 0)
+            return;
+
+        if(!(this->openingTheta > float_X(0.0) && this->openingTheta <= float_X(360.0)))
+        {
+            std::cerr << "[Plugin] [" << this->prefix
+                      << "] openingTheta has to be within (0, 360]."
+                      << std::endl;
+            return;
```
PrometheusPi: Do you relay want to continue the simulation with just a warning or do you want to end the simulation?

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,542 @@
...
+        }
+        if(!(this->openingPhi > float_X(0.0) && this->openingPhi <= float_X(360.0)))
+        {
+            std::cerr << "[Plugin] [" << this->prefix
+                      << "] openingPhi has to be within (0, 360]."
+                      << std::endl;
+            return;
```
PrometheusPi: same here

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,542 @@
...
+
+        if(!(this->openingTheta > float_X(0.0) && this->openingTheta <= float_X(360.0)))
+        {
+            std::cerr << "[Plugin] [" << this->prefix
+                      << "] openingTheta has to be within (0, 360]."
+                      << std::endl;
+            return;
+        }
+        if(!(this->openingPhi > float_X(0.0) && this->openingPhi <= float_X(360.0)))
```
PrometheusPi: with both `theat` and `phi` at `360°` you cover the 4 pi sphere twice

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,542 @@
...
+        {
+            std::cerr << "[Plugin] [" << this->prefix
+                      << "] openingPhi has to be within (0, 360]."
+                      << std::endl;
+            return;
+        }
+
+        this->maxTheta = float_X(0.5) * this->openingTheta;
+        this->maxPhi = float_X(0.5) * this->openingPhi;
```
PrometheusPi: see comment on member variables above

PrometheusPi: Would a direct conversion to rad make sense here? All other quantities are converted to "simulation units" as well.

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
PrometheusPi: Sorry my fault. This is fine. :+1:

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
Heikman: I want to end the simulation. How can I do this?

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,542 @@
...
+                                      hBufTotal.size().y(),
+                                      hBufTotal.size().z());
+
+        /* if there is only one energy bin, omit the energy axis */
+        uint32_t dimension = this->numBinsEnergy == 1 ? DIM2 : DIM3;
+        hdf5DataFile.write(currentStep,
+                           SplashTypeX,
+                           dimension,
+                           splash::Selection(bufferSize),
+                           "detectorLeftPars",
```
PrometheusPi: Reuse of this string. Please define it as const string as a member variable and use the new variable both for writing and reading. 
What do you think of using `"detectorLeftParticles"` instead (self explaining and more intuitive)?

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
Heikman: In the coordinate system [used here](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/Plugin%3A-ParticleDetector) it covers the 2 pi sphere's surface.

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
Heikman: I've left it this way, because we need the degree value also for the hdf5 metadata.

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,542 @@
...
+        /* calculate rotated detector frame from posTheta and posPhi */
+        const float_64 posTheta_rad = this->posTheta * float_64(M_PI / 180.0);
+        const float_64 posPhi_rad = this->posPhi * float_64(M_PI / 180.0);
+        this->detectorFrameVecY = float3_X(pam::sin(posTheta_rad) * pam::cos(posPhi_rad),
+                                           pam::cos(posTheta_rad) * pam::cos(posPhi_rad),
+                                           pam::sin(posPhi_rad));
+        /* If the y-axis is pointing exactly up- or downwards we need to define the x-axis manually */
+        if(pam::abs(this->detectorFrameVecY) == float_X(1.0))
+        {
+            this->detectorFrameVecX = float3_X(1.0, 0.0, 0.0);
```
PrometheusPi: In case of `posTheta_rad = 0` and `posPhi_rad = 0` (`detectorFrameVecX = float3_X(0.0, 1.0, 0.0)`) the above condition is fulfilled as well. This leads to an error with this special case treatment.

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,542 @@
...
+        }
+        else
+        {
+            const float3_X vecUp(0.0, 0.0, -1.0);
+            this->detectorFrameVecX = pam::cross(vecUp, this->detectorFrameVecY);
+            /* normalize vector */
+            this->detectorFrameVecX /= pam::abs(this->detectorFrameVecX);
```
PrometheusPi: your vector is already normalized

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,542 @@
...
+        this->detectorFrameVecZ = pam::cross(this->detectorFrameVecX, this->detectorFrameVecY);
+
+        /* create calorimeter functor instance */
+        this->calorimeterFunctor = MyCalorimeterFunctorPtr(new MyCalorimeterFunctor(
+            this->maxTheta * float_X(M_PI / 180.0),
+            this->maxPhi * float_X(M_PI / 180.0),
+            this->numBinsTheta,
+            this->numBinsPhi,
+            this->numBinsEnergy,
+            this->logScale ? pam::log10(this->minEnergy) : this->minEnergy,
```
PrometheusPi: Please add a comment, that in case  of a logarithmic scale your `minEnergy` and `maxEnergy` is already converted to log values.

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
PrometheusPi: sorry my fault - this is `detectorFrameVecX` and not `detectorFrameVecY` (which is normalized)

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
Heikman: done in `ParticleDetectorFunctors.hpp`

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,542 @@
...
+        foreach(zone, cursor::make_MultiIndexCursor<simDim>(), particleDetectorKernel);
+
+        /* copy to host */
+        *this->hBufCalorimeter = *this->dBufCalorimeter;
+
+        /* mpi reduce */
+        using namespace lambda;
+        (*this->allGPU_reduce)(*this->hBufTotalCalorimeter, *this->hBufCalorimeter, _1 + _2);
+        if(!this->allGPU_reduce->root())
+            return;
```
PrometheusPi: This is the second mpi reduce of this kind in this file. Could you abstract it?

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,542 @@
...
+        ((this->prefix + ".numBinsPhi").c_str(), po::value<uint32_t > (&this->numBinsPhi)->default_value(64),
+            "number of bins for angle phi.")
+        ((this->prefix + ".numBinsEnergy").c_str(), po::value<uint32_t > (&this->numBinsEnergy)->default_value(1),
+            "number of bins for the energy spectrum. Disabled by default.")
+        ((this->prefix + ".minEnergy").c_str(), po::value<float_X > (&this->minEnergy)->default_value(float_X(0.0)),
+            "minimal detectable energy in keV.")
+        ((this->prefix + ".maxEnergy").c_str(), po::value<float_X > (&this->maxEnergy)->default_value(float_X(1.0e3)),
+            "maximal detectable energy in keV.")
+        ((this->prefix + ".logScale").c_str(), po::bool_switch(&this->logScale),
+            "enable logarithmic energy scale to the base of 10.")
```
PrometheusPi: Since you are not entering the energy as exponent of `10`, it is not necessary to give a base:
test it for yourself:
```python
N = 10

E_min = 2
E_max = 2000

logScale10 = np.zeros(N)
logScaleEuler = np.zeros(N)

E_min_logE = np.log(E_min)
E_max_logE = np.log(E_max)
scale_logE = np.exp(np.linspace(E_min_logE, E_max_logE, N))

E_min_log10 = np.log10(E_min)
E_max_log10 = np.log10(E_max)
scale_log10 = np.power(10., np.linspace(E_min_log10, E_max_log10, N))

print(scale_logE)
print(scale_log10)
```

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
```diff
@@ -0,0 +1,542 @@
...
+                                           pam::sin(posPhi_rad));
+        /* If the y-axis is pointing exactly up- or downwards we need to define the x-axis manually */
+        if(pam::abs(this->detectorFrameVecY) == float_X(1.0))
+        {
+            this->detectorFrameVecX = float3_X(1.0, 0.0, 0.0);
+        }
+        else
+        {
+            const float3_X vecUp(0.0, 0.0, -1.0);
+            this->detectorFrameVecX = pam::cross(vecUp, this->detectorFrameVecY);
```
PrometheusPi: please document this arbitrary selection for the (general) direction of `detectorFrameVecX`

src/picongpu/include/plugins/particleDetector/ParticleDetectorFunctors.hpp
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    template<typename ParticlesFrame>
+    DINLINE void operator()(ParticlesFrame& particlesFrame, const uint32_t linearThreadIdx)
+    {
+        const float3_X mom = particlesFrame[linearThreadIdx][momentum_];
+        const float_X mom2 = math::dot(mom, mom);
+        float3_X dirVec = mom * math::rsqrt(mom2);
+
+        /* rotate dirVec into the detector frame */
```
PrometheusPi: is rotate here the correct word. You are doing a coordinate transformation of `dirVec` from the Cartesian coordinate system of PIConGPU to the  Cartesian coordinate system of your detector.

src/picongpu/include/plugins/particleDetector/ParticleDetectorFunctors.hpp
```diff
@@ -0,0 +1,208 @@
...
+            int32_t phiBin = calorimeterPos.y() * static_cast<float_X>(numBinsPhi);
+            // catch out-of-range values
+            phiBin = phiBin >= numBinsPhi ? numBinsPhi - 1 : phiBin;
+            phiBin = phiBin < 0 ? 0 : phiBin;
+
+            // energy
+            const float_X weighting = particlesFrame[linearThreadIdx][weighting_];
+            const float_X normedWeighting = weighting /
+                                            static_cast<float_X>(particles::TYPICAL_NUM_PARTICLES_PER_MACROPARTICLE);
+            const float_X energy = SPEED_OF_LIGHT * math::sqrt(mom2) / weighting;
```
PrometheusPi: this is only valid for photons
However, your detector/plugin works for any species.

src/picongpu/include/plugins/particleDetector/ParticleDetectorFunctors.hpp
```diff
@@ -0,0 +1,208 @@
...
+
+            int32_t energyBin = 0;
+            if(this->numBinsEnergy > 1u)
+            {
+                energyBin = math::floor(((logScale ? log10(energy) : energy) - minEnergy) /
+                              (maxEnergy - minEnergy) * static_cast<float_X>(this->numBinsEnergy - 2)) + 1;
```
PrometheusPi: please fix the magic number two 
e.g. `const unsigned int N_outOfRange_bin = 2;`

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
Heikman: Amazing! Is this a math bug or a feature?

src/picongpu/include/simulation_defines/param/particleDetector.param
```diff
@@ -0,0 +1,50 @@
...
+ * Useful for tuning the spacial detector resolution.
+ *
+ * \param theta -maxTheta...maxTheta
+ * \param phi -maxPhi...maxPhi
+ * \param maxTheta maximum value of angle theta
+ * \param maxPhi maximum value of angle phi
+ * \return Two values within [-1,1]
+ */
+HDINLINE float2_X mapThetaPhiToNormedRange(const float_X theta,
+                                           const float_X phi,
```
PrometheusPi: Why  do you put this quite general method into a param file?

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
Heikman: can you give an example for your point?

src/picongpu/include/plugins/particleDetector/ParticleDetectorFunctors.hpp
PrometheusPi: Do you check whether  `maxEnergy > minEnery`?

src/picongpu/include/plugins/particleDetector/ParticleDetectorFunctors.hpp
Heikman: Well the coordinate transformation is a rotation.

src/picongpu/include/plugins/particleDetector/ParticleDetectorFunctors.hpp
Heikman: to be precise, this plugin only works for *massless species* as well as this code block.

src/picongpu/include/simulation_defines/param/particleDetector.param
Heikman: Thus the user can tune the spatial resolution for his special purposes.

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
PrometheusPi: e.g. by a `std::runtime_error`, see [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/73befbda38bfec86561aeca666a816dd5d1a8e9a/src/picongpu/include/simulationControl/MySimulation.hpp#L294)

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
PrometheusPi: your link is not leading to a valid page - please update

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
PrometheusPi: that is fin with me, perhaps you could think of an explicit renaming to `maxThetaDeg`

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
PrometheusPi: that's fine :+1:

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
PrometheusPi: you are using the same lines of code [here](https://github.com/Heikman/picongpu/blob/bba18337b644a5e453e176605f175e3a26b00fec/src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp#L171-L175) as well.

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
PrometheusPi: @Heikman Are you kidding me?

src/picongpu/include/plugins/particleDetector/ParticleDetectorFunctors.hpp
PrometheusPi: You are right, because your coordinate transformation will never consist of a translation. Could you still add `rotate/coordinate transformation` just to ease understanding.

src/picongpu/include/plugins/particleDetector/ParticleDetectorFunctors.hpp
PrometheusPi: How do you avoid that a user calls the plugin for massive particles.

src/picongpu/include/simulation_defines/param/particleDetector.param
PrometheusPi: How would a user tune resolution here? 
Why does he need to tune resolution in the param file if he can already tune resolution with the command line flags of the plugin (by `--ph_calorimeter.numBinsTheta``etc.)?

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
PrometheusPi: new link is [here](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/Plugin:-ParticleCalorimeter)

src/picongpu/include/plugins/particleDetector/ParticleDetectorFunctors.hpp
```diff
@@ -0,0 +1,210 @@
...
+        float3_X dirVec = mom * math::rsqrt(mom2);
+
+        /* rotate dirVec into the detector frame */
+        using namespace PMacc::algorithms::math;
+        dirVec = float3_X(dot(this->detectorFrameVecX, dirVec),
+                          dot(this->detectorFrameVecY, dirVec),
+                          dot(this->detectorFrameVecZ, dirVec));
+
+        /* convert dirVec to theta and phi */
+        const float_X theta = atan2(dirVec.x(), dirVec.y());
```
PrometheusPi: reading your wiki entry - should `asin(dirVec.x())` not be enough here?

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
PrometheusPi: following the description of the wiki page, I agree and your setup is fine. 
But does this not disagree with  this [this line of code](https://github.com/Heikman/picongpu/blob/bba18337b644a5e453e176605f175e3a26b00fec/src/picongpu/include/plugins/particleDetector/ParticleDetectorFunctors.hpp#L95)? `atan` has only a co-domain of [-90°, +90°].

src/picongpu/include/plugins/particleDetector/DetectGuardParticles.def
Heikman: Offline discussion: the general solution will be implied in a follow-up PR.

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
Heikman: This is why I use [`atan2`](https://de.wikipedia.org/wiki/Arkustangens_und_Arkuskotangens#atan2).

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
Heikman: but the arguments are different.

src/picongpu/include/plugins/particleDetector/ParticleDetectorFunctors.hpp
Heikman: The restriction to massless species is just a short-term solution, since for now a species always has to have a mass attribute, see #1209. So we cannot distinguish the massive and the massless case right now, i.e. the arbitrary choice of the massless energy formula is just of pragmatic reason.

src/picongpu/include/plugins/particleDetector/ParticleDetectorFunctors.hpp
Heikman: I've added an exception call.

src/picongpu/include/simulation_defines/param/particleDetector.param
Heikman: the number of bins is a coarse tool for tuning the resolution. If the user needs for example a finer resolution in the middle of the calorimeter, he can change the mapping in the param file here.

Heikman:
The out is not yet compatible to openPMD

src/picongpu/include/plugins/particleDetector/ParticleDetector.hpp
PrometheusPi: OK :+1:

src/picongpu/include/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -0,0 +1,562 @@
...
+    std::ofstream outFile;
+    const std::string leftParticlesDatasetName;
+
+    uint32_t numBinsYaw;
+    uint32_t numBinsPitch;
+    uint32_t numBinsEnergy;
+    float_X minEnergy;
+    float_X maxEnergy;
+    bool logScale;
+    float_X openingYaw;
```
PrometheusPi: please add `...Deg` to show that these are angles in degree not in radiant (as anywhere else)

PrometheusPi:
@Heikman great job!

PrometheusPi:
@Heikman There seems to exist confilcts with the current `dev`. Please resolve.

Heikman:
ready to merge?

PrometheusPi:
yes - ready to merge from my side
@ax3l @psychocoderHPC Any more comments from your side? Please tell me so till 12:00 UTC today.

ax3l:
Why UTC, I think that's a trap.

doc/TBG_macros.cfg
```diff
@@ -198,6 +198,11 @@ TBG_liveViewYZ="--<species>_liveView.period 1 --<species>_liveView.slicePoint 0.
 # Print the maximum charge deviation between particles and div E to textfile 'chargeConservation.dat':
 TBG_chargeConservation="--chargeConservation.period 100"
 
+# ParticleDetector: (virtually) propagates and collects particles to infinite distance
+TBG_<species>_detector="--<species>_detector.period 100 --<species>_detector.openingTheta 90 --<species>_detector.openingPhi 30
```
ax3l: outdated naming

src/picongpu/include/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -0,0 +1,530 @@
...
+
+        splash::SerialDataCollector hdf5DataFile(1);
+        splash::DataCollector::FileCreationAttr fAttr;
+
+        splash::DataCollector::initFileCreationAttr(fAttr);
+
+        std::stringstream filename;
+        filename << checkpointDirectory << "/" << prefix << currentStep;
+
+        hdf5DataFile.open(filename.str().c_str(), fAttr);
```
ax3l: why is this the same name as the regular output?

Checkpoints should *not* be exposed to the user and should be written inside `--checkpoint-directory`. See [this wiki entry](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/Restarting-a-simulation).

ax3l: also, give our checkpoints a suffix with `.<timeStep>`

src/picongpu/include/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -0,0 +1,530 @@
...
+
+    void writeToHDF5File(uint32_t currentStep)
+    {
+        splash::SerialDataCollector hdf5DataFile(1);
+        splash::DataCollector::FileCreationAttr fAttr;
+
+        splash::DataCollector::initFileCreationAttr(fAttr);
+
+        std::stringstream filename;
+        filename << this->foldername << "/" << this->prefix << currentStep << "-";
```
ax3l: suggested change:
```C++
filename << this->foldername << "/" << this->prefix << "_" << currentStep;
```

for `<species>_calorimeter_<step>_0_0_0.h5`

src/picongpu/include/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -0,0 +1,530 @@
...
+        if(!this->allGPU_reduce->root())
+            return;
+
+        splash::SerialDataCollector hdf5DataFile(1);
+        splash::DataCollector::FileCreationAttr fAttr;
+
+        splash::DataCollector::initFileCreationAttr(fAttr);
+
+        std::stringstream filename;
+        filename << checkpointDirectory << "/" << prefix << currentStep;
```
ax3l: pls give our checkpoints a suffix with `.<timeStep>` (example for [txt files](https://github.com/ComputationalRadiationPhysics/picongpu/blob/f5fe0487f8ed070894feb11a40da11703a9f2391/src/picongpu/include/plugins/common/txtFileHandling.hpp#L88-L112))

suggested change:
```C++
filename << checkpointDirectory << "/" << prefix << "." << currentStep;
```

for `checkpoints/<species>_calorimeter_0_0_0.h5.<step>`

ax3l: no I am wrong, will update

src/picongpu/include/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -0,0 +1,530 @@
...
+        if(rank == 0)
+        {
+            splash::SerialDataCollector hdf5DataFile(1);
+            splash::DataCollector::FileCreationAttr fAttr;
+
+            splash::DataCollector::initFileCreationAttr(fAttr);
+            fAttr.fileAccType = splash::DataCollector::FAT_READ;
+
+            std::stringstream filename;
+            filename << restartDirectory << "/" << prefix << restartStep;
```
ax3l: your rebase removed my comment... (this is wasting time, please be more careful)


suggested change:
```C++
filename << restartDirectory << "/" << prefix <<< "_" << restartStep;
```

src/picongpu/include/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -0,0 +1,530 @@
...
+        if(!this->allGPU_reduce->root())
+            return;
+
+        splash::SerialDataCollector hdf5DataFile(1);
+        splash::DataCollector::FileCreationAttr fAttr;
+
+        splash::DataCollector::initFileCreationAttr(fAttr);
+
+        std::stringstream filename;
+        filename << checkpointDirectory << "/" << prefix << "." << currentStep;
```
ax3l: suggested change:

```C++
filename << checkpointDirectory << "/" << prefix << "_" << currentStep;
```

(I was not finished with the review)

Heikman: I have not rebased, just pushed.

src/picongpu/include/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -0,0 +1,530 @@
...
+
+    std::string pluginGetName() const
+    {
+        return this->name;
+    }
+
+
+    void onParticleLeave(const std::string& speciesName, int32_t direction)
+    {
+        if(this->notifyPeriod == 0)
```
ax3l: that is an interesting question. should we ignore steps or should we store those particles beforehand? ;)
maybe someone wants to restart and enable the plugin later on :D (can be solved later)

src/picongpu/include/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -0,0 +1,530 @@
...
+            hdf5DataFile.read(restartStep,
+                              this->leftParticlesDatasetName.c_str(),
+                              dimensions,
+                              &(*hBufLeftParsCalorimeter.origin()));
+
+            hdf5DataFile.close();
+
+            /* rank 0 divides and distributes the calorimeter to all ranks in equal parts */
+            int numRanks;
+            MPI_Comm_size(MPI_COMM_WORLD, &numRanks);
```
psychocoderHPC: Please use the `GridController` to get the number of gpus.

src/picongpu/include/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -0,0 +1,530 @@
...
+
+    void restart(uint32_t restartStep, const std::string restartDirectory)
+    {
+        if(this->notifyPeriod == 0)
+            return;
+
+        HBufCalorimeter hBufLeftParsCalorimeter(this->dBufLeftParsCalorimeter->size());
+
+        int rank;
+        MPI_Comm_rank(MPI_COMM_WORLD, &rank);
```
psychocoderHPC: Please use the `GridController` to get the number of gpus.

src/picongpu/include/plugins/particleCalorimeter/ParticleCalorimeter.hpp
ax3l: the PR only has one commit, so it was amended which is a rebase :)

Heikman: ah, right.

src/picongpu/include/plugins/particleCalorimeter/ParticleCalorimeter.hpp
Heikman: but I want the rank, that's what I want

src/picongpu/include/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -0,0 +1,530 @@
...
+            using namespace lambda;
+            PMacc::algorithm::host::Foreach()(hBufLeftParsCalorimeter.zone(),
+                                              hBufLeftParsCalorimeter.origin(),
+                                              _1 = _1 / float_X(numRanks));
+        }
+
+        MPI_Bcast(&(*hBufLeftParsCalorimeter.origin()),
+                  hBufLeftParsCalorimeter.size().productOfComponents() * sizeof(float_X),
+                  MPI_CHAR,
+                  0,
```
psychocoderHPC: Ahh ok if you use this you need to use the real MPI_Rank above. (sry my fault)

PrometheusPi:
@ax3l and @psychocoderHPC any more remarks?
If not, @Heikman please give your commits meaningful names and squash if suitable.

psychocoderHPC:
Please rebase your branch because #1427 changed the `libSplash` version to `1.4.0` and start a runtime test to verify your results.

Heikman:
waiting for #1432. Multiple runtime tests have done already, see e.g. the test script in the PR description. If you want I can also add another test for the angular distribution.

Flamefire:
Please rebase to current dev. This branch seems to be based on https://github.com/ComputationalRadiationPhysics/picongpu/commit/5f82fc19bb78ef0bb6ebe42db44a628454be5c94 according to the network graph.

What @psychocoderHPC wants to say: libSpalsh 1.4.0 has some breaking changes. Please test that HDF5 handling is still correct after the rebase and usage of the new libSplash

Flamefire:
This compiles with #1435 which is the prober usage of Boost.Move.

@Heikman: Again, please rebase to latest dev and test with Splash 1.4.0

Heikman:
works fine with Splash 1.4.0

PR #1386, opened 2016-02-22
Swap members and adapt comment

### LINKED ISSUE #1372: Wrong member order in particle
There is a comment in Particle.hpp that seems to be wrong: https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/libPMacc/include/particles/memory/dataTypes/Particle.hpp#L73-L83

It states that "sizeof(uint32_t)>sizeof(reference)" where `reference` means `pointer` in this case. But pointers are either 32 bit or 64 bit types and hence `sizeof(uint32_t) <= sizeof(pointer)`

So the order should be swapped and the comment corrected.

(File by @psychocoderHPC )

Flamefire:
This fixes #1372 by swapping the pointer with the uint32_t member

### DIFF
--- a/src/libPMacc/include/particles/memory/dataTypes/Particle.hpp
+++ b/src/libPMacc/include/particles/memory/dataTypes/Particle.hpp
@@ -53,7 +53,7 @@ namespace pmacc = PMacc;
 
 /** A single particle of a @see Frame
  *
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
well spotted

PR #1387, opened 2016-02-22
FieldTmp Central Solver: Update Documentation

ax3l:
Updates the documentation and variable naming after several rebases of the central kernel in `FieldTmp`'s `particleToGrid` methods.

Does not change the implementation in any way.

- [x] needs RT testing, played around with that and should avoid pulling in unwanted changes


ax3l:
fieldTmp results look identical up to the 3rd/4th place in density

PR #1388, opened 2016-02-22
Doxygen `@code` Snippet Examples

ax3l:
Corrects the wrong usage of `@example` in doxygen blocks.
This generated warnings during doxygen generation.

- [`@example`](https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdexample)
  is used to cross link example *files*
- [`@code/@endcode`](https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdcode)
  can be used to add code snippets


psychocoderHPC:
thy well done

PR #1390, opened 2016-02-22
Fix wrong order in CMAKE_PREFIX_PATH

Flamefire:
This allows using user defined libraries from the paths in the environment variables (like CMAKE_PREFIX_PATH or BOOST_ROOT) even if a library of the same name is in `/usr/lib/x86_64-linux-gnu/`

@ax3l

### DIFF
--- a/src/picongpu/CMakeLists.txt
+++ b/src/picongpu/CMakeLists.txt
@@ -36,8 +36,16 @@ if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
 endif(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
 
 # set helper pathes to find libraries and packages
-set(CMAKE_PREFIX_PATH "/usr/lib/x86_64-linux-gnu/" "$ENV{MPI_ROOT}"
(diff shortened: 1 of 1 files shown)


src/picongpu/CMakeLists.txt
```diff
@@ -258,6 +266,7 @@ find_package(Boost 1.56.0 REQUIRED COMPONENTS program_options regex filesystem
                                               system thread math_tr1)
 include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
 set(LIBS ${LIBS} ${Boost_LIBRARIES})
+message(STATUS "Boost libs = ${Boost_LIBRARIES}")
```
ax3l: not sure if we want this because usually CMake outputs this on its own. but if we want  to add it, use the commonly used cmake output syntax
```cmake
message(STATUS "Found Boost: ${Boost_LIBRARIES}")
```

Flamefire: Should not be there. Was debug stuff

ax3l: ok :)

Flamefire:
Follow-up in #1391 for sub-projects

ax3l:
excellent, thank you!

PR #1393, opened 2016-02-22
add 'getPluginFromType()' method in PluginConnector

Heikman:
add 'getPluginFromType()' method in PluginConnector.

See the discussion in #1376.

### DIFF
--- a/src/libPMacc/include/pluginSystem/PluginConnector.hpp
+++ b/src/libPMacc/include/pluginSystem/PluginConnector.hpp
@@ -1,5 +1,6 @@
 /**
- * Copyright 2013-2016 Rene Widera, Felix Schmitt, Axel Huebl, Benjamin Worpitz
+ * Copyright 2013-2016 Rene Widera, Felix Schmitt, Axel Huebl, Benjamin Worpitz,
+ *                     Heiko Burau
  *
  * This file is part of libPMacc.
  *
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/pluginSystem/PluginConnector.hpp
```diff
@@ -180,6 +181,19 @@ namespace PMacc
             }
         }
 
+        template<typename Plugin>
+        Plugin* getPluginFromType()
```
psychocoderHPC: Please add a description to this function.
What is if there are more than one plugin with the same type inside the list named `plugins`?

psychocoderHPC: I checked if it is possible that more than one plugin can have the same type. It is possible e.g., in PIConGPU the plugin `Visualisation<ParticleType,OuputMethod>` is registered multiple times with the same type signature.
This means that the name must be `getFirstPluginFromType()` or it should be return a vector with all instances. The second version is IMO the cleanest solution.

Heikman:
tested with #1376

src/libPMacc/include/pluginSystem/PluginConnector.hpp
```diff
@@ -180,6 +182,27 @@ namespace PMacc
             }
         }
 
+        /**
+         * Get a vector of pointers of all plugin instances from a given type.
```
ax3l: of all *registered* plugin instances

Heikman:
ready for merging?

Issue #517, opened 2014-07-29
Output: Particles That Leave the Box


ax3l:
For non-moving window simulations, particle histograms with particles that leave the box can look quite awkward if some particles leave the box over time.

An idea could be to keep book of particles that leave the box (thx for the idea Peter Hilz).

Alternatively, one could add, for specific plugins like the histogram, a separate histogram that stores the influence of "lost" particles.
## 
### Status
- [x] Plugin interfaces updated to support handling of leaving particles each time step: ~~#1394~~
- [ ] default still handling missing for species #1396
- [ ] most plugins need update to support leaving particles (if wanted) [wiki](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-Plugins)

Flamefire:
There is a policy that can be passed to the `ParticleDescription` struct. This is e.g. `PMacc::HandleGuardRegion` where leaving particles can be handled in a functor. This could then store the particles somewhere else or convert them to a "leaving" species that can then periodically be collected by a plugin.

ax3l:
thx!

my idea is: we move the leaving particles in each time step to an
additional species (per species), called "species_leaving". It has the
additional scalar record ("attribute") "tleft" that is filled with the
time step the particle was moved to that (not interacting,
"leaving"/"dead") species.

The output (hdf5/adios) plugins can now do the following: also dump this
species (since the last dump), e.g., vectorAllLeavingSpecies.

This does not yet solve if online plugins want to access them over all
time steps (because that might become memory heavy in the outer GPUs)
but already is a major step forward in post-processing.

hm.... dead would be an appropriate naming, too. could also be used for
absorbed/annihilated particles inside the box.

ax3l:
@Heikman found a solution in his new calorimeter plugin #1376 that can be a very useful alternative for many other plugins (such as energy histograms, etc.):

Use in PIConGPU an new `HandleGuardRegion` in `ParticleDescription` as described above that does:
- iterate all plugins
- search if the plugin has a new `IPlugin` method, maybe called `onParticleLeave`
- call those methods (for now: read only operations) to update the plugins internal state
- then finally call the PIConGPU handling that generally depends on the particle boundary condition and is currently simply delete

For futher plugins to support leaving particles, they simply have to add this `onParticleLeave` method hook.

ax3l:
ccing @psychocoderHPC

PR #1394, opened 2016-02-25
Notify plugins on leaving particles

Heikman:
add a new guard region policy class that calls the new method `IPlugin::onParticleLeave()` for each registered plugin whenever particles are leaving the global simulation volume.

Solves #517.
Tested with #1376.

### DIFF
--- a/src/libPMacc/include/pluginSystem/IPlugin.hpp
+++ b/src/libPMacc/include/pluginSystem/IPlugin.hpp
@@ -116,6 +116,21 @@ namespace PMacc
          */
         virtual std::string pluginGetName() const = 0;
 
+        /**
+         * Called each timestep if particles are leaving the global simulation volume.
+         *
+         * This method is only called for species which are marked with the
+         * `GuardHandlerCallPlugins` policy in their descpription.
+         *
+         * The order in which the plugins are called is undefined, so this means
+         * read-only access to the particles.
+         *
+         * \param speciesName name of the particle species
+         * \param direction the direction the particles are leaving the simulation
+         */
+        virtual void onParticleLeave(const std::string& /*speciesName*/, const int32_t /*direction*/)
+        {}
+
         /** When was the plugin checkpointed last?
          *
          * @return last checkpoint's time step
--- a/src/picongpu/include/simulationControl/GuardHandlerCallPlugins.hpp
+++ b/src/picongpu/include/simulationControl/GuardHandlerCallPlugins.hpp
@@ -1,0 +1,64 @@
+/**
+ * Copyright 2016 Heiko Burau
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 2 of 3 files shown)


ax3l:
pls rebase after #1395 was merged :)

ax3l:
Can you please update [this *how to write your plugin* documentation](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/Develop-Plugins) in our wiki in the meantime? :)

src/picongpu/include/simulationControl/GuardHandlerCallPlugins.hpp
```diff
@@ -0,0 +1,64 @@
...
+        {
+            (*iter)->onParticleLeave(T_Particles::FrameType::getName(), direction);
+        }
+
+        particles.deleteGuardParticles(direction);
+    }
+
+    template< class T_Particles >
+    void
+    handleIncoming(T_Particles& par, int32_t direction) const
```
ax3l: pls comment out the unused `par` and `direction` (or remove the args) :) otherwise some compilers throw warnings

ax3l:
Can you pls add the `GuardHandlerCallPlugins` as a default to the general `simulation_defines/param/speciesDefinition.param`?

src/picongpu/include/simulationControl/GuardHandlerCallPlugins.hpp
```diff
@@ -0,0 +1,64 @@
...
+ * Guard handler policy calling all registered plugins when particles
+ * leave the global simulation volume. This class serves as policy for
+ * the `ParticleDescription` template class.
+ *
+ * For each plugin the method `IPlugin::onParticleLeave()` is called.
+ * After that the guard particles are deleted.
+ */
+struct GuardHandlerCallPlugins
+{
+    typedef PMacc::particles::policies::ExchangeParticles HandleExchanged;
```
ax3l: `typedef typename`  ?

src/libPMacc/include/pluginSystem/IPlugin.hpp
```diff
@@ -116,6 +116,21 @@ namespace PMacc
...
+         * This method is only called for species which are marked with the
+         * `GuardHandlerCallPlugins` policy in their descpription.
+         *
+         * The order in which the plugins are called is undefined, so this means
+         * read-only access to the particles.
+         *
+         * \param speciesName name of the particle species
+         * \param direction the direction the particles are leaving the simulation
+         */
+        virtual void onParticleLeave(const std::string& /*speciesName*/, int32_t /*direction*/)
```
ax3l: why not `const int32_t` ? :)

ax3l:
looks very good, thank you! :sparkles: just a very few minor remarks and a small addition in the defaults

src/picongpu/include/simulationControl/GuardHandlerCallPlugins.hpp
Heikman: There is no template argument, so it's not necessary.

Heikman:
`ParticleDescription` is from libPMacc. We could inherit from that class in picongpu to set the default template parameter but that looks to me like another task.

ax3l:
Oh that is an even better idea, shall we put that in a follow-up you say?

ax3l:
great, code looks good, like it :sparkles:

PR #1399, opened 2016-02-29
Speed up sliceFieldReader.py

BeyondEspresso:
The sample script for reading the output files of the SliceFieldPrinter was very slow. This change speeds up execution by approximately a factor of 60. In tests, the readout of a single file is now performed in 30 seconds.


Heikman:
that's the pythonic way, very nice :=)

PrometheusPi:
Great work.

PR #1400, opened 2016-03-02
minor fix: float_32 -> float_X

Heikman:
minor fix: `float_32` -> `float_X`

`float_32` causes a compiler error in the Bunch example with 64 bit precision.


psychocoderHPC:
well spotted!

PR #1401, opened 2016-03-09
Add include and fix wrong assignment operator

Flamefire:
The include is required for the boost move semantics.

The function fixed caused problems in C++11 mode (compile failure) as there the class does not exist. The correct move assignment is here: https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/libPMacc/include/cuSTL/container/DeviceBuffer.hpp#L103
The bug was introduced by @Heikman in #1375.


ax3l:
@Heikman pls have a look

Flamefire:
@ax3l This is already merged :)

ax3l:
upsi :) thx!

PR #1403, opened 2016-03-15
Resource monitor class and resource log plugin

erikzenker:
The resource monitor is located in libPMacc and provides information about allocated resources of the library. Currently, particle count and cell count per rank are provided.

The resource log is a user adaptable resource information log plugin which collects resource information and writes these information to a stream.

The user can choose which resource properties to write e.g.:
`--resourceLog.properties rank position currentStep particleCount cellCount`

which format to use e.g.:
`--resourceLog.format json` other options are `jsonpp`, `xml` and `xmlpp`

which stream the information should be written e.g.:
`--resourceLog.stream file` other options are `stdout` and `stderr`

and which prefix to use for filestreams:
`--resourceLog.prefix myprefix_`

Using the options ` --resourceLog.period 1 --resourceLog.stream stdout --resourceLog.properties rank position currentStep particleCount cellCount --resourceLog.format jsonpp` will write resource objects to stdout such as:

    [1,1]<stdout>:    "resourceLog": {
    [1,1]<stdout>:        "rank": "1",
    [1,1]<stdout>:        "position": {
    [1,1]<stdout>:            "x": "0",
    [1,1]<stdout>:            "y": "1",
    [1,1]<stdout>:            "z": "0"
    [1,1]<stdout>:        },
    [1,1]<stdout>:        "currentStep": "357",
    [1,1]<stdout>:        "cellCount": "1048576",
    [1,1]<stdout>:        "particleCount": "2180978"
    [1,1]<stdout>:    }
    [1,1]<stdout>:}

For each format there exists always a non pretty print version to simplify further processing:

    [1,3]<stdout>:{"resourceLog":{"rank":"3","position":{"x":"1","y":"1","z":"0"},"currentStep":"415","cellCount":"1048576","particleCount":"2322324"}}

### DIFF
--- a/src/libPMacc/include/mappings/simulation/ResourceMonitor.hpp
+++ b/src/libPMacc/include/mappings/simulation/ResourceMonitor.hpp
@@ -1,0 +1,67 @@
+/**
+ * Copyright 2016 Erik Zenker
+ *
+ * This file is part of libPMacc.
+ *
+ * libPMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * libPMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with libPMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+#include <vector>  /* std::vector */
+#include <cstdlib> /* std::size_t */
+
+namespace PMacc
+{
+
+    /**
+     * Provides ressource information of the current subgrid
+     *
+     * @tparam DIM number of dimensions of the simulation
+     */
+    template <unsigned T_DIM>
+    class ResourceMonitor
+    {
+    public:
+
+        /**
+         * Constructor
+         */
+        ResourceMonitor();
+
+    private:
+
+        /**
+         * Constructor
+         */
+        ResourceMonitor(const ResourceMonitor& fs);
+
+    public:
+        /**
+         *  Returns the number of cells on the device
+         */
+        std::size_t getCellCount();
+
+        /**
+         * Returns the number of particles per species on the device
+         */
+        template <typename T_Species, typename T_MappingDesc>
+        std::vector<std::size_t> getParticleCounts(T_MappingDesc &cellDescription);
+
+    };
+
+} //namespace PMacc
+
--- a/src/picongpu/include/plugins/ResourceLog.hpp
+++ b/src/picongpu/include/plugins/ResourceLog.hpp
@@ -1,0 +1,244 @@
+/**
+ * Copyright 2016 Erik Zenker
+ *
+ * This file is part of libPMacc.
+ *
+ * libPMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * libPMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with libPMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+// PMacc
+#include "Environment.hpp"
+#include "mappings/simulation/ResourceMonitor.hpp"
+
+// PIConGPU
+#include "plugins/ILightweightPlugin.hpp"
+#include "ILightweightPlugin.hpp"
+#include "simulation_defines.hpp"
+
+// Boost
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/xml_parser.hpp>
+#include <boost/filesystem.hpp>
+
+// STL
+#include <iostream>  /* std::cout, std::ostream */
+#include <numeric>   /* std::accumulate */
+#include <string>    /* std::string */
+#include <sstream>   /* std::stringstream */
+#include <fstream>   /* std::filebuf */
+#include <map>       /* std::map */
+
+// C LIB
+#include <stdlib.h> /* itoa */
+#include <stdint.h> /* uint32_t */
+
+namespace picongpu
+{
+    using namespace PMacc;
+
+
+    class ResourceLog : public ILightweightPlugin
+    {
+    private:
+        MappingDesc *cellDescription;
+        ResourceMonitor<simDim> resourceMonitor;
+
+        // programm options
+        std::string outputFilePrefix;
+        std::string streamType;
+        std::string outputFormat;
+        std::vector<std::string> properties;
+
+        std::filebuf fileBuf;
+        std::map<std::string, bool> propertyMap;
+
+    public:
+
+        ResourceLog() :
+                cellDescription(NULL)
+        {
+            Environment<>::get().PluginConnector().registerPlugin(this);
+        }
+
+        std::string pluginGetName() const
+        {
+            return "ResourceLog";
+        }
+
+        void notify(uint32_t currentStep)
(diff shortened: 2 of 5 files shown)


PrometheusPi:
Cool! Would it be possible to add this kind of information to the **hdf5 output as well** (or to write it in a separate hdf5 file? Would it make sense to have this information available with the hdf5 output?

How much resources does this output need?
I personally believe it would be great to have this output available (by default) in the hdf5 output. We often encounter user/physicist problems where a simulation crashed. One usually has to calculate or estimate the  resource usage afterwards. With this plugin (as default output) looking at the last hdf5 dump would tremendously help to find these problems.

erikzenker:
The resource usage depends on the properties the user embeds into the stream. Properties which are excluded from the stream will not be retrieved.

By "in the hdf5 output" you mean the output of the hdf5 plugin? Well the hdf5 plugin could retrieve resource information from the resource monitor class and embed this information into the hdf5 files, but I would not prefer this approach. Better, the resourceLog plugin gather the hdf5 file descriptor and writes resource information directly into the hdf5 file. I am not sure if this works, I have to check it first.

Maybe this plugin could also prevent the system to crash by notifying other components when e.g.: memory is low etc.

src/libPMacc/include/mappings/simulation/ResourceMonitor.hpp
```diff
@@ -0,0 +1,67 @@
...
+        /**
+         *  Returns the number of cells on the device
+         */
+        std::size_t getCellCount();
+
+        /**
+         * Returns the number of particles per species on the device
+         */
+        template <typename T_Species, typename T_MappingDesc>
+        std::vector<std::size_t> getParticleCounts(T_MappingDesc &cellDescription);
```
ax3l: maybe you want to return a pair here with species name and count

ax3l: or a vector of struct

ax3l: because each species' particles might have different attributes and might therefore be more or less "memory heavy" in terms of load-balancing metrics

src/picongpu/include/plugins/ResourceLog.hpp
```diff
@@ -0,0 +1,244 @@
...
+            {
+                throw std::runtime_error(std::string("resourcelog.stream ") + streamType + std::string(" is not known, use stdout, stderr or file instead."));
+            }
+        }
+
+        void pluginRegisterHelp(po::options_description& desc)
+        {
+            /* register command line parameters for your plugin */
+            desc.add_options()
+                    ("resourceLog.period", po::value<uint32_t>(&notifyPeriod)->default_value(0),
```
ax3l: please add an example to `doc/TBG_macros.cfg` and document the plugin in our [wiki](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-Plugins) :)

ax3l:
great to see you already have the ressource monitor going on! :)

You might want to split the new libPMacc RessourceMonitor and the PIConGPU plugin in two PRs.
One idea for the plugin: one could also open a high port on each rank (or master) to query it in a style [munin nodes (clients)](http://munin-monitoring.org/wiki/MuninTroubleshooting) can be accessed (via simple telnet commands or alternatively [(web)sockets](https://github.com/clarkduvall/jsterm), etc.)

ax3l:
regarding HDF5: just write an *independent* HDF5 file if necessary. but for now the implemented formats are quite enough, I would rely on auto-serialization of the final (json) object for other outputs to keep the maintainance low.

src/picongpu/include/plugins/ResourceLog.hpp
```diff
@@ -0,0 +1,244 @@
...
+        void setMappingDescription(MappingDesc *cellDescription)
+        {
+            this->cellDescription = cellDescription;
+        }
+
+    private:
+        uint32_t notifyPeriod;
+
+        void pluginLoad() {
+            Environment<>::get().PluginConnector().setNotificationPeriod(this, notifyPeriod);
```
psychocoderHPC: If the notifyPeriod is zero than this plugin should not registered and initialized.

src/picongpu/include/plugins/ResourceLog.hpp
```diff
@@ -0,0 +1,244 @@
...
+            }
+            else
+            {
+                for(size_t i = 0; i < properties.size(); ++i)
+                {
+                    propertyMap[properties[i]] = true;
+                }
+            }
+
+            //
```
psychocoderHPC: empty comment

src/picongpu/include/plugins/ResourceLog.hpp
```diff
@@ -0,0 +1,244 @@
...
+            // Prepare file for output stream
+            if (streamType == "file")
+            {
+                size_t rank = static_cast<size_t>(Environment<simDim>::get().GridController().getGlobalRank());
+                std::stringstream ss;
+                ss << outputFilePrefix << rank;
+                boost::filesystem::path resourceLogPath(ss.str());
+                fileBuf.open(resourceLogPath.string().c_str(), std::ios::out);
```
psychocoderHPC: Where is this file closed?

src/picongpu/include/plugins/ResourceLog.hpp
```diff
@@ -0,0 +1,244 @@
...
+                    ("resourceLog.period", po::value<uint32_t>(&notifyPeriod)->default_value(0),
+                     "Enable ResourceLog plugin [for each n-th step]")
+                    ("resourceLog.prefix", po::value<std::string>(&outputFilePrefix)->default_value("resourcelog_"),
+                     "Set the filename prefix for output file if a filestream was selected")
+                    ("resourceLog.stream", po::value<std::string>(&streamType)->default_value("file"),
+                     "Output stream [stdout, stderr, file]")
+                    ("resourceLog.properties", po::value<std::vector<std::string> >(&properties)->multitoken(),
+                     "List of properties to log [rank, position, currentStep, cellCount, particleCount]")
+                    ("resourceLog.format", po::value<std::string>(&outputFormat)->default_value("json"),
+                     "Output format of log (pp for pretty print) [json, jsonpp xml, xmlpp]");
```
psychocoderHPC: Comma is missing `json, jsonpp, xml, xmlpp`

src/picongpu/include/plugins/ResourceLog.hpp
```diff
@@ -0,0 +1,244 @@
...
+            }
+        }
+
+        void pluginRegisterHelp(po::options_description& desc)
+        {
+            /* register command line parameters for your plugin */
+            desc.add_options()
+                    ("resourceLog.period", po::value<uint32_t>(&notifyPeriod)->default_value(0),
+                     "Enable ResourceLog plugin [for each n-th step]")
+                    ("resourceLog.prefix", po::value<std::string>(&outputFilePrefix)->default_value("resourcelog_"),
```
psychocoderHPC: `resourcelog_` should be camel case.

src/picongpu/include/plugins/ResourceLog.hpp
```diff
@@ -0,0 +1,244 @@
...
+#include "mappings/simulation/ResourceMonitor.hpp"
+
+// PIConGPU
+#include "plugins/ILightweightPlugin.hpp"
+#include "ILightweightPlugin.hpp"
+#include "simulation_defines.hpp"
+
+// Boost
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/json_parser.hpp>
```
psychocoderHPC: Are there any changes in our `CMakeLists.txt` necessary?

ax3l: good question.
I think no, since I can not find [it](http://www.boost.org/doc/libs/1_60_0/doc/html/property_tree.html) in the list of [to-be-compiled boost libraries](https://gist.github.com/ax3l/7450482).

psychocoderHPC:
please check the annotations

src/picongpu/include/plugins/ResourceLog.hpp
psychocoderHPC: ping

src/picongpu/include/plugins/ResourceLog.hpp
erikzenker: I also do not think, that any changes are necessary. The compiler does not complain using the current one.

src/picongpu/include/plugins/ResourceLog.hpp
erikzenker: I will close the fileBuf within the pluginUnload method (in the case it was opened before).

src/picongpu/include/plugins/ResourceLog.hpp
erikzenker: You are right, thx !

src/libPMacc/include/mappings/simulation/ResourceMonitor.hpp
erikzenker: I am not sure if this is really necessary, since the values in the vector have the same order as the species sequence. Accordingly, v[0] corresponds to species 0 and so on. Providing, the species name is string would be redundant information. I would adapt the method documentation.

erikzenker:
This PR will be split into a ResourceMonitor and ResourceLog PR

PR #1404, opened 2016-03-15
Find PMacc after #1367

Flamefire:
Followup so that the CMake for the GoL example finds PMacc as the `types.h` is now `pmacc_types.hpp`


psychocoderHPC:
well spotted, thx

PR #1405, opened 2016-03-16
fix `StrideMapping` usage for 2D

psychocoderHPC:
- fix `FieldTmp::computeValue`
- fix `FieldJ::computeCurrent`

### Extended Description

added by @ax3l:
The result of the wrong usage are race conditions in 2D during *current deposition*: the stride of super cells was accidentally set to `2` super cells instead of `3`. The result of that is, that current writes in the strided ("skipped") super-cell (with the reminiscence of particles close to the super-cells border) are not protected by atomic adds.
For this to have a negative effect in form of a race condition (resulting in *lowered* current deposited in this area), the support of the particle shape must be at least the size of the shortest super-cell-edge, *this is usually not the case*.

3D simulations are not effected.

### DIFF
--- a/src/picongpu/include/fields/FieldJ.tpp
+++ b/src/picongpu/include/fields/FieldJ.tpp
@@ -275,7 +275,7 @@ void FieldJ::computeCurrent( ParticlesClass &parClass, uint32_t )
         typename GetMargin<ParticleCurrentSolver>::UpperMargin
         > BlockArea;
 
-    StrideMapping<AREA, simDim, MappingDesc> mapper( cellDescription );
+    StrideMapping<AREA, 3, MappingDesc> mapper( cellDescription );
     typename ParticlesClass::ParticlesBoxType pBox = parClass.getDeviceParticlesBox( );
     FieldJ::DataBoxType jBox = this->fieldJ.getDeviceBuffer( ).getDataBox( );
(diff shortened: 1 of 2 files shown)


ax3l:
Plot: [weibel_2d_race_J.py.txt](https://github.com/ComputationalRadiationPhysics/picongpu/files/177704/weibel_2d_race_J.py.txt)
Diff for testing: [WeibelWithoutAdd2EMF.diff.txt](https://github.com/ComputationalRadiationPhysics/picongpu/files/177713/WeibelWithoutAdd2EMF.diff.txt)

### Shape: P4S (Support=5)

top to bottom super cells: 8x8 (`dev`), 4x4 (`dev`), 4x4 (`fix`)

weibel_2d_race_p4s_j

ax3l:
Great catch! I think the fix is working as expected.

Since particle shapes are very limited in their extent and a race condition can only occur when the shapes of two next but one super-cells overlap (which means less then a 5x5 super cell in our case for 2D) no user should have triggered that yet.

Affected setups

| super cell | shape |
| ---- | ----- |
| 8x8 (default) | / |
| < 5x5 | P4S (5) |
| < 4x4 | PCS (4) |
| < 3x3 | TSC (default, 3) |
| < 2x2 | CIC (2) |

PR #1410, opened 2016-03-21
Particle ids

Flamefire:
Requires:
- [x] #1409 
- [x] #1415 
- [x] Runtime test with Splash 1.4.0

This adds a speciesAttribute `particleId`, the traits for units and initializes the `IdProvider` in `MySimulation`.
It also implements saving and loading from HDF5 and ADIOS just like the moving window slider.

### DIFF
--- a/src/picongpu/include/plugins/adios/ADIOSWriter.def
+++ b/src/picongpu/include/plugins/adios/ADIOSWriter.def
@@ -74,16 +74,15 @@ namespace po = boost::program_options;
     }                                                                         \
 }
 
-#define ADIOS_CMD_EXPECT_NONZERO(_cmd)                                        \
+#define ADIOS_CMD_EXPECT_NONNULL(_cmd)                                        \
 {                                                                             \
-    int _err_code = _cmd;                                                     \
-    if (_err_code == 0)                                                       \
+    if (!(_cmd))                                                              \
     {                                                                         \
         std::string errMsg( adios_errmsg() );                                 \
         if( errMsg.empty() ) errMsg = '\n';                                   \
         std::stringstream s;                                                  \
         s << "ADIOS: error at cmd '" << #_cmd                                 \
-          << "' (" << _err_code << ", " << adios_errno << ") in "             \
+          << "' (" << adios_errno << ") in "                                  \
           << __FILE__ << ":" << __LINE__ << " " << errMsg;                    \
         throw std::runtime_error(s.str());                                    \
     }                                                                         \
--- a/src/picongpu/include/plugins/adios/ADIOSWriter.hpp
+++ b/src/picongpu/include/plugins/adios/ADIOSWriter.hpp
@@ -32,6 +32,7 @@
 #include "plugins/adios/ADIOSWriter.def"
 
 #include "particles/frame_types.hpp"
+#include "particles/IdProvider.def"
 
 #include <adios.h>
 #include <adios_read.h>
@@ -574,6 +575,21 @@ public:
         mThreadParams.window = MovingWindow::getInstance().getDomainAsWindow(restartStep);
         mThreadParams.localWindowToDomainOffset = DataSpace<simDim>::create(0);
 
+        uint64_t* idProviderStatePtr = NULL;
+        int idProviderStateSize;
+        enum ADIOS_DATATYPES idProviderStateType;
+        ADIOS_CMD(adios_get_attr( mThreadParams.fp,
+                                  (mThreadParams.adiosBasePath + std::string("sim_IdProvider_State")).c_str(),
+                                  &idProviderStateType,
+                                  &idProviderStateSize,
+                                  (void**) &idProviderStatePtr ));
+
+        log<picLog::INPUT_OUTPUT > ("ADIOS: value of sim_IdProvider_State = %1%") % *idProviderStatePtr;
+
+        assert(idProviderStateType == PICToAdios<uint64_t>().type);
+        assert(idProviderStateSize == sizeof(*idProviderStatePtr));
+        IdProvider<simDim>::setNextId(*idProviderStatePtr);
+
         /* load all fields */
         ForEach<FileCheckpointFields, LoadFields<bmpl::_1> > forEachLoadFields;
         forEachLoadFields(&mThreadParams);
@@ -585,6 +601,7 @@ public:
         /* free memory allocated in ADIOS calls */
         free(slidesPtr);
         free(lastStepPtr);
+        free(idProviderStatePtr);
 
         /* clean shut down: close file and finalize */
         adios_release_step( mThreadParams.fp );
@@ -832,6 +849,12 @@ private:
         ADIOS_CMD(adios_define_attribute_byvalue(threadParams->adiosGroupHandle,
                   "sim_slides", threadParams->adiosBasePath.c_str(),
                   adiosUInt32Type.type, 1, (void*)&slides ));
+
+        uint64_t idProviderState = IdProvider<simDim>::getNextId();
+        log<picLog::INPUT_OUTPUT > ("ADIOS: meta: sim_IdProvider_State");
+        ADIOS_CMD(adios_define_attribute_byvalue(threadParams->adiosGroupHandle,
+                  "sim_IdProvider_State", threadParams->adiosBasePath.c_str(),
+                  PICToAdios<uint64_t>().type, 1, static_cast<void*>(&idProviderState) ));
 
         /* write normed grid parameters */
         log<picLog::INPUT_OUTPUT > ("ADIOS: meta: grid");
--- a/src/picongpu/include/plugins/adios/ADIOSWriter.hpp
+++ b/src/picongpu/include/plugins/adios/ADIOSWriter.hpp
@@ -73,6 +73,7 @@
 #include "plugins/adios/ADIOSCountParticles.hpp"
 #include "plugins/adios/restart/LoadSpecies.hpp"
 #include "plugins/adios/restart/RestartFieldLoader.hpp"
+#include "plugins/adios/NDScalars.hpp"
 
 
 namespace picongpu
@@ -575,21 +576,6 @@ public:
         mThreadParams.window = MovingWindow::getInstance().getDomainAsWindow(restartStep);
         mThreadParams.localWindowToDomainOffset = DataSpace<simDim>::create(0);
 
-        uint64_t* idProviderStatePtr = NULL;
-        int idProviderStateSize;
-        enum ADIOS_DATATYPES idProviderStateType;
-        ADIOS_CMD(adios_get_attr( mThreadParams.fp,
-                                  (mThreadParams.adiosBasePath + std::string("sim_IdProvider_State")).c_str(),
-                                  &idProviderStateType,
-                                  &idProviderStateSize,
-                                  (void**) &idProviderStatePtr ));
-
-        log<picLog::INPUT_OUTPUT > ("ADIOS: value of sim_IdProvider_State = %1%") % *idProviderStatePtr;
-
-        assert(idProviderStateType == PICToAdios<uint64_t>().type);
-        assert(idProviderStateSize == sizeof(*idProviderStatePtr));
-        IdProvider<simDim>::setNextId(*idProviderStatePtr);
-
         /* load all fields */
         ForEach<FileCheckpointFields, LoadFields<bmpl::_1> > forEachLoadFields;
         forEachLoadFields(&mThreadParams);
@@ -598,10 +584,16 @@ public:
         ForEach<FileCheckpointParticles, LoadSpecies<bmpl::_1> > forEachLoadSpecies;
         forEachLoadSpecies(&mThreadParams, restartChunkSize);
 
+        uint64_t idProviderNextId, idProviderMaxNumProc;
+        ReadNDScalars<uint64_t, uint64_t>()(mThreadParams,
+                "picongpu/idProviderState", &idProviderNextId,
+                "maxNumProc", &idProviderMaxNumProc);
+        log<picLog::INPUT_OUTPUT > ("Setting id on current rank: %1%") % idProviderNextId;
+        IdProvider<simDim>::setState(idProviderNextId, idProviderMaxNumProc);
+
         /* free memory allocated in ADIOS calls */
         free(slidesPtr);
         free(lastStepPtr);
-        free(idProviderStatePtr);
 
         /* clean shut down: close file and finalize */
         adios_release_step( mThreadParams.fp );
@@ -850,12 +842,6 @@ private:
                   "sim_slides", threadParams->adiosBasePath.c_str(),
                   adiosUInt32Type.type, 1, (void*)&slides ));
 
-        uint64_t idProviderState = IdProvider<simDim>::getNextId();
-        log<picLog::INPUT_OUTPUT > ("ADIOS: meta: sim_IdProvider_State");
-        ADIOS_CMD(adios_define_attribute_byvalue(threadParams->adiosGroupHandle,
-                  "sim_IdProvider_State", threadParams->adiosBasePath.c_str(),
-                  PICToAdios<uint64_t>().type, 1, static_cast<void*>(&idProviderState) ));
-
         /* write normed grid parameters */
         log<picLog::INPUT_OUTPUT > ("ADIOS: meta: grid");
         ADIOS_CMD(adios_define_attribute_byvalue(threadParams->adiosGroupHandle,
@@ -987,6 +973,10 @@ private:
         }
         log<picLog::INPUT_OUTPUT > ("ADIOS: ( end ) counting particles.");
 
+        PMACC_AUTO(idProviderState, IdProvider<simDim>::getState());
+        WriteNDScalars<uint64_t, uint64_t> writeIdProviderState("picongpu/idProviderState", "maxNumProc");
+        writeIdProviderState.prepare(*threadParams, idProviderState.get<1>());
+
         /* allocate buffer in MB according to our current group size */
         /* `1 + mem` minimum 1 MiB that we can write attributes on empty GPUs */
         size_t writeBuffer_in_MiB=1+threadParams->adiosGroupSize / 1024 / 1024;
@@ -1042,7 +1032,9 @@ private:
         }
         log<picLog::INPUT_OUTPUT > ("ADIOS: ( end ) writing particle species.");
 
-        /* close adios file, most liekly the actual write point */
+        writeIdProviderState(*threadParams, idProviderState.get<0>());
+
+        /* close adios file, most likely the actual write point */
         log<picLog::INPUT_OUTPUT > ("ADIOS: closing file: %1%") % threadParams->fullFilename;
         ADIOS_CMD(adios_close(threadParams->adiosFileHandle));
 
--- a/src/picongpu/include/plugins/adios/ADIOSWriter.hpp
+++ b/src/picongpu/include/plugins/adios/ADIOSWriter.hpp
@@ -1032,6 +1032,8 @@ private:
         }
         log<picLog::INPUT_OUTPUT > ("ADIOS: ( end ) writing particle species.");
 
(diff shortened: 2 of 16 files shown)


bussmann:
THIS would be AWESOME!

src/picongpu/include/plugins/adios/ADIOSWriter.hpp
```diff
@@ -543,7 +544,7 @@ class ADIOSWriter : public ILightweightPlugin
         log<picLog::INPUT_OUTPUT > ("ADIOS: value of sim_slides = %1%") %
             slides;
 
-        assert(slidesType == adiosUInt32Type.type);
+        assert( slidesType == adiosUInt32Type.type);
```
psychocoderHPC: one space to much

Flamefire:
Tested and verified with ADIOS and HDF5 restart. Can be merged once the PMacc patch is merged and this is refactored.

@ax3l Please review

ax3l:
@Flamefire pls rebase against `dev` so the PR gets shorter. can't promise I will finish the review before GTC but I will try.

Flamefire:
Rebased

ax3l:
ty!

src/picongpu/include/plugins/adios/ADIOSWriter.def
```diff
@@ -74,16 +74,15 @@ namespace po = boost::program_options;
     }                                                                         \
 }
 
-#define ADIOS_CMD_EXPECT_NONZERO(_cmd)                                        \
+#define ADIOS_CMD_EXPECT_NONNULL(_cmd)                                        \
 {                                                                             \
-    int _err_code = _cmd;                                                     \
-    if (_err_code == 0)                                                       \
+    if (!(_cmd))                                                              \
```
ax3l: why did you change this part? I think we explicitly wrote it that way to access the `_cmd` var only once.

Flamefire: This define was never used. But I found, that one needs a define to check for Non-NULL-pointers e.g. when defining attributes. So now I can use `ADIOS_CMD_EXPECT_NONNULL( varInfo = adios_inq_var(params.fp, datasetName.c_str()) );`

ax3l: great!

src/picongpu/include/plugins/hdf5/HDF5Writer.hpp
```diff
@@ -221,6 +221,15 @@ class HDF5Writer : public ISimulationPlugin
         ForEach<FileCheckpointParticles, LoadSpecies<bmpl::_1> > forEachLoadSpecies;
         forEachLoadSpecies(params, restartChunkSize);
 
+        IdProvider<simDim>::State idProvState;
+        ReadNDScalars<uint64_t, uint64_t>()(mThreadParams,
+                "picongpu/idProvider/startId", &idProvState.startId,
+                "maxNumProc", &idProvState.maxNumProc);
+        ReadNDScalars<uint64_t>()(mThreadParams,
+                "picongpu/idProvider/nextId", &idProvState.nextId);
+        log<picLog::INPUT_OUTPUT > ("Setting id on current rank: %1%") % idProvState.nextId;
```
ax3l: be more explicit, let us write "Setting next id on ..."

src/picongpu/include/plugins/adios/NDScalars.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+namespace picongpu {
+namespace adios {
+
+/* Functors for reading and writing ND scalar fields with N=simDim
+ * In the current implementation each process (of the ND grid) reads/writes 1 scalar value
+ * Optionally the processes can also read/write an attribute for this dataset by using a non-empty attrName
+ */
+
+template<typename T_Skalar, typename T_Attribute = uint64_t>
```
ax3l: typo: `T_Scalar`

src/picongpu/include/plugins/adios/NDScalars.hpp
```diff
@@ -0,0 +1,155 @@
...
+struct WriteNDScalars
+{
+    WriteNDScalars(const std::string& name, const std::string& attrName = ""):name(name), attrName(attrName){}
```
ax3l: pls put the initialization on the next line

src/picongpu/include/plugins/adios/NDScalars.hpp
```diff
@@ -0,0 +1,155 @@
...
+        log<picLog::INPUT_OUTPUT> ("ADIOS: write %1%D scalars: %2%") % simDim % name;
+
+        ADIOS_CMD( adios_write_byid(params.adiosFileHandle, varId, &value) );
+    }
+private:
+    const std::string name, attrName;
+    int64_t varId;
+};
+
+template<typename T_Skalar, typename T_Attribute = uint64_t>
```
ax3l: `T_Scalar` typo

src/picongpu/include/plugins/adios/restart/ReadAttribute.hpp
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Read an attribute from an open ADIOS file, check that its type is correct and return it
+     *
+     * @param fp       Open ADIOS file handle
+     * @param basePath Path where the attribute is located in the file (with or w/o trailing slash)
+     * @param attrName Name of the attribute. Used for status output and concatenated with basePath
+     * @retval Attribute value
+     */
+    template<typename T_Attribute>
+    T_Attribute readAttribute(ADIOS_FILE* fp, const std::string& basePath, const std::string& attrName)
```
ax3l: don't we just the the same thing while reading `sim_slides` but way less verbose?

ax3l: otherwise: let's read `sim_slides` the same way so we have the same check.

btw: can you move sim_slide to the new `picongpu/...` dir, too?

src/picongpu/include/plugins/hdf5/NDScalars.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+namespace picongpu {
+namespace hdf5 {
+
+/* Functors for reading and writing ND scalar fields with N=simDim
+ * In the current implementation each process (of the ND grid) reads/writes 1 scalar value
+ * Optionally the processes can also read/write an attribute for this dataset
+ */
+
+template<typename T_Skalar, typename T_Attribute = uint64_t>
```
ax3l: Typo `T_Scalar`

src/picongpu/include/plugins/hdf5/NDScalars.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+#include "pmacc_types.hpp"
+#include "plugins/hdf5/HDF5Writer.def"
+#include "traits/PICToSplash.hpp"
+#include "Environment.hpp"
+
+namespace picongpu {
+namespace hdf5 {
+
+/* Functors for reading and writing ND scalar fields with N=simDim
```
ax3l: just a side note: even if `simDim` is 3 it does not mean one needs to do the domain decomposition in 3 dimensions. In the context of ID providers this is a bit irritating, but ... ook.

src/picongpu/include/plugins/hdf5/NDScalars.hpp
```diff
@@ -0,0 +1,122 @@
...
+ * In the current implementation each process (of the ND grid) reads/writes 1 scalar value
```
ax3l: I would write ND domain decomposition, grid is reserved in our speech for actual cells

src/picongpu/include/simulation_defines/unitless/speciesAttributes.unitless
```diff
@@ -194,6 +195,25 @@ struct UnitDimension<weighting >
     }
 };
 
+template<>
+struct Unit<particleId>
+{
+    /* zero-sized vector indicating unitless flag for hdf5 and adios output */
```
ax3l: that is right now changing in #1428 .

please make it directly
```C++
    /* unitless and not scaled by a factor: by convention 1.0 */
    static std::vector<double> get()
    {
        std::vector<double> unit( 1, 1.0 );
        return unit;
    }
```

src/picongpu/include/traits/AdiosToPIC.tpp
```diff
@@ -44,6 +44,12 @@ namespace traits
         typedef unsigned int type;
     };
 
+    template<>
+    struct AdiosToPIC<adios_unsigned_long>
+    {
+        typedef uint64_t type;
```
ax3l: I think `unsigned long` would be better (see above)

ax3l: pls add `adios_long` for completeness

src/picongpu/include/traits/PICToAdios.tpp
```diff
@@ -60,6 +60,15 @@ namespace traits
         type(adios_unsigned_long) {}
     };
 
+    template<>
+    struct PICToAdios<uint64_cu>
```
ax3l: hm... necessary? can't you cast to `unit64_t` before?

ax3l: above: I think our current implementation is not ideal, we should do
```C++
/** Trait for unsigned long */
     template<>
     struct PICToAdios<unsigned long>
     {
         ADIOS_DATATYPES type;
         
         PICToAdios() :
         type(adios_unsigned_long) {}
     };
```

since adios does not identify 64 and 32bit directly for ints.

src/picongpu/include/traits/PICToSplash.tpp
```diff
@@ -57,6 +57,18 @@ namespace traits
...
     };
 
+    template<>
+    struct PICToSplash<uint64_t>
+    {
+        typedef splash::ColTypeUInt64 type;
+    };
+
+    template<>
+    struct PICToSplash<uint64_cu>
```
ax3l: same here, better use `uint64_t` for writing if possible.

src/picongpu/include/plugins/adios/NDScalars.hpp
```diff
@@ -0,0 +1,155 @@
...
+#include <stdexcept>
+
+namespace picongpu {
+namespace adios {
+
+/* Functors for reading and writing ND scalar fields with N=simDim
+ * In the current implementation each process (of the ND grid) reads/writes 1 scalar value
+ * Optionally the processes can also read/write an attribute for this dataset by using a non-empty attrName
+ */
+
```
ax3l: no space between doxygen und class impl pls :)

ax3l: add `/**` to make it doxygen-ish

ax3l: pls add `@tparam descriptors`

ax3l:
comments are inline, looks good to me if the runtime tests were all stable.

@psychocoderHPC please have a look about the new `traits`, too (adios exposes no `uint64_t` just `int` and `long` which I think we should specialize for).

psychocoderHPC:
Please rebase your branch because #1427 changed the `libSplash` version to `1.4.0` and start a runtime test to verify your results.
Maybe this not effects your branch.

src/picongpu/include/traits/PICToAdios.tpp
psychocoderHPC: We can't cast our data befor we write because it is a generic method.
But as wrote in our chat session we can test if it is possible to define `struct PICToAdios<unsigned long int>` and `struct PICToAdios<unsigned long long int>`.

src/picongpu/include/plugins/adios/restart/ReadAttribute.hpp
Flamefire: Yeah I have seen quite some duplicated code and before adding another instance of the code I generalized it into a functor.
Moving is possible, but please open an issue and let's do this in another PR. They are unaffected by this and can be done separately AFTER this one.

src/picongpu/include/plugins/hdf5/NDScalars.hpp
Flamefire: If simDim is 3 than the processes are organized in a 3D grid (or however you name them). And per definition (of this functor) every process writes 1 value.

src/picongpu/include/plugins/hdf5/NDScalars.hpp
Flamefire: This is less related to the domain but the the process organization. So I'd rather write `ND grid of processes`

src/picongpu/include/traits/AdiosToPIC.tpp
Flamefire: We would not be the only ones doing that: https://github.com/Kitware/VTK/blob/master/IO/ADIOS/ADIOSScalar.cxx#L64, http://fossies.org/linux/VTK/IO/ADIOS/ADIOSUtilities.cxx

src/picongpu/include/traits/PICToSplash.tpp
Flamefire: See above, it is required unfortunately

src/picongpu/include/traits/AdiosToPIC.tpp
psychocoderHPC: Offline discussed with many emotions :-)
Adios defines `adios_unsigned_long` as `unsigned long long` see [here](https://github.com/ornladios/ADIOS/blob/master/src/core/common_read.c#L1362-L1363)

We should do the same even if `unsigned long long` is not defined in `C++03` but in `C++11`.
`unsigned long long` is defined in `C99` and Adios is C code.

`unsigned long long` in C++11 is defined as `at least 64bit` see [here](http://de.cppreference.com/w/cpp/language/types)

Flamefire:
Code is rebased and all suggestions so far introduced except the ADIOS types thing where there was some discussion. I added the missing signed versions but used the standard fixed-width typedefs and a conditional for the `uint64_cu` so it is only defined when its not equal to `uint64_cu`. Reasoning:
- ADIOS may **use** `unsigned long long` but **defines** `adios_unsigned_long` to refer to an 8-byte type: https://github.com/ornladios/ADIOS/blob/37d0b2b44899e1c2bb1e538f2b949e39bcde68a/src/public/adios_types.h#L16-L26. This file belongs to the public interface and should be the reference. Not some (probably faulty) internal function
- ADIOS also enforces the width to be 8 bytes: https://github.com/ornladios/ADIOS/blob/37d0b2b44899e1c2bb1e538f2b949e39bcde68a1/src/core/adios_bp_v1.c#L2387-L2389
- So passing anything that is not an **exactly** 8 byte type will cause an error: https://github.com/ornladios/ADIOS/blob/37d0b2b44899e1c2bb1e538f2b949e39bcde68a1/src/core/common_adios.c#L768

I adapted the other trait specializations too, to have a consistent and correct implementation.

psychocoderHPC:
@Flamefire Thanks for the links to the Adios implementation.

src/picongpu/include/traits/PICToAdios.tpp
```diff
@@ -32,7 +34,7 @@ namespace traits
 {
     /** Trait for int */
     template<>
-    struct PICToAdios<int>
+    struct PICToAdios<int32_t>
```
psychocoderHPC: if you change this you should also change the definition in `PICToSplash`

src/picongpu/include/traits/PICToAdios.tpp
```diff
@@ -32,7 +34,7 @@ namespace traits
 {
     /** Trait for int */
     template<>
-    struct PICToAdios<int>
```
psychocoderHPC: If we remove this with fixed size types than we should add an description at the begin of the files that it is not allowed to define native C/C++ includes types e.g., `unsigned long`, `int` ,... because they can conflict with the types `int32_t`, `uint64_t`, ...

src/picongpu/include/plugins/adios/restart/ReadAttribute.hpp
ax3l: ok for me!

ax3l:
ok for me. The only thing is: I would avoid `uint64_cu` alltogether in the code but we can do it with the enable_if for now.

ax3l:
@psychocoderHPC ok from my side, pls merge if you are ok, too.

src/picongpu/include/traits/PICToAdios.tpp
Flamefire: Using them would simply not compile if someone was to forget this. So I see no need for that description

src/picongpu/include/traits/PICToAdios.tpp
Flamefire: Changed that anyway so it is now consistent

src/picongpu/include/traits/PICToAdios.tpp
psychocoderHPC: I mean not a comment for the user but a comment for developer that we not add types like int or unsigned long in a half year because we forget that it can conflict with uint32_t.
This double definition can compile but if a user switch to a a tegra or similar maybe it crash.

Flamefire: Well it has a) to be merged and b) pass the tests on the CI suite. So I don't think it would compile. (How could it, if it conflicts?) And if it even crashes then it is a bug introduced by the developer and undetected by the reviewer.
I just think that a comment like this is nowhere else and if we start adding comments with coding guidelines to the code there will be quite some code bloat with something that should be obvious or (if at all) documented in the general coding guidelines.

Flamefire:
I don't see that we can avoid `uint64_cu` as the cuda API is defined with that type. Especially the atomics. So with the conditional here we are at least flexible and if cuda was to use `uint64_t` in the future, than this code does not need to change.

src/picongpu/include/traits/PICToAdios.tpp
ax3l: yes the problem is more if you go a system like tk1 which has a 32bit host, the compile suite will not cross-compile to it for testing.

nevertheless, I think it's fine as it is @psychocoderHPC

ax3l:
Why do you use it in `IDProvider` again? You could also use a `HostDeviceBuffer<uint64_t, 1>` buffer and be set, no?

Flamefire:
The `atomicAdd` is defined on a `uint64_cu`. I could probably make the idProvider itself return an uint64_t using an implicit or explicit conversion. Might work as well. Just wanted to avoid that conversion as we do have that type

Flamefire:
@psychocoderHPC I removed the uint64_cu from the interface. So I consider this ready for merge

psychocoderHPC:
Thanks for this nice new feature.

**Note:** Next time we should split it up into two or three pull requests. But nobody is still using this feature therefore it can be merged to `dev`.

ax3l:
@Flamefire did you do a runtime test after your last changes? :)

When rebasing against the new changes that got into dev (first view: this branch) I get a `boost::traits` runtime error
```
boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::io::too_many_args> >
boost::too_many_args: format-string referred to less arguments than were passed
```

I am just guessing it might be this PR since it occurs on enabled HDF5 output since the last rebase.

src/picongpu/include/plugins/hdf5/HDF5Writer.hpp
```diff
@@ -452,6 +462,14 @@ class HDF5Writer : public ISimulationPlugin
         }
         log<picLog::INPUT_OUTPUT > ("HDF5: ( end ) writing particle species.");
 
+        PMACC_AUTO(idProviderState, IdProvider<simDim>::getState());
+        log<picLog::INPUT_OUTPUT>("HDF5: Writing IdProvider state (StartId: %1%, NextId: %1%, maxNumProc: %2%)")
```
ax3l: found the RT error :)

ax3l:
pushed the fix to #1440 ;)

Flamefire:
I actually did test it, but I guess this bug just shows when the log option is enabled. Thanks for fixing :+1:

ax3l:
Yes that's what probably happened, no prob :)

PR #1411, opened 2016-03-21
Actually exit when pression no

Flamefire:
@ax3l 
This fixes the bug, that "NO" is ignored by the compilesuite in non-parallel mode


ax3l:
hehe, thanks! :)
probably I was always curious (or quick on Ctrl+C fires :D)

PR #1412, opened 2016-03-21
Make CommunicatorMPI dim-safe

Flamefire:
The current implementation makes the compiler throw some warnings about out-of-bounds array accesses which are especially correct for an instantiation with `DIM=1`

This implementation fixes this by adding appropriate checks.

It also closes the open TODO in the code.

### DIFF
--- a/src/libPMacc/include/communication/CommunicatorMPI.hpp
+++ b/src/libPMacc/include/communication/CommunicatorMPI.hpp
@@ -1,6 +1,6 @@
 /**
  * Copyright 2013-2016 Axel Huebl, Felix Schmitt, Heiko Burau, Rene Widera,
- *                     Wolfgang Hoenig, Benjamin Worpitz
+ *                     Wolfgang Hoenig, Benjamin Worpitz, Alexander Grund
  *
  * This file is part of libPMacc.
  *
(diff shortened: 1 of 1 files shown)


ax3l:
well spotted, will review! :)

src/libPMacc/include/communication/CommunicatorMPI.hpp
```diff
@@ -38,6 +38,35 @@
 namespace PMacc
 {
 
+namespace detail
+{
+    template <unsigned T_DIM>
```
ax3l: it's a bit notorious but we could also do:

leave the general case unimplemented and implement DIM1 explicitly.

src/libPMacc/include/communication/CommunicatorMPI.hpp
```diff
@@ -192,23 +220,24 @@ class CommunicatorMPI : public ICommunicator
 
     bool slide()
     {
+        if(DIM < DIM2)
```
ax3l: add comment:
```C++
// we can only slide in y direction right now
```

src/libPMacc/include/communication/CommunicatorMPI.hpp
```diff
@@ -192,23 +220,24 @@ class CommunicatorMPI : public ICommunicator
...
-                return true;
-        }
 
-        return false;
+        return coordinates[1] == dims[1] - 1;
     }
 
     bool setStateAfterSlides(size_t numSlides)
     {
+        if(DIM < DIM2)
```
ax3l: same comment here

ax3l:
ty! :sparkles:

PR #1413, opened 2016-03-21
Score-P 2.0 instrumentation

winklerf-zih:
All cmake files were cleaned up from scorep. We are now using scorep wrapper scripts to instrument the code, see https://github.com/ComputationalRadiationPhysics/picongpu/wiki/Performance-Analysis.

For longer runs manual instrumentation is useful. Since manual instrumentation is guarded by SCOREP_USER_ENABLE we can put this into the source code.

cudaDeviceSynchornize() at program exit is required for scorep to flush the cuda buffers. This prevents the use of flushatexit.

### DIFF
--- a/src/picongpu/submit/titan-ornl/picongpu.profile.example
+++ b/src/picongpu/submit/titan-ornl/picongpu.profile.example
@@ -36,8 +36,21 @@ export MPI_ROOT=$MPICH_DIR
 #   configure with -c "-DVAMPIR_ENABLE=ON"
 #   e.g.:
 #     $PICSRC/configure -c "-DVAMPIR_ENABLE=ON" ~/paramSets/case001
-module load vampirtrace/5.14.4
-export VT_ROOT=$VAMPIRTRACE_DIR
+#module load vampirtrace/5.14.4
+#export VT_ROOT=$VAMPIRTRACE_DIR
+ 
+# scorep (optional) ############################################
+#   configure with -c "-DCMAKE_CXX_COMPILER=`which scorep-CC` \
+#                      -DCUDA_NVCC_EXECUTABLE=`which scorep-nvcc`"
+#   e.g.:
+#     SCOREP_WRAPPER=OFF $PICSRC/configure -a sm_35 \
+#         -c "-DCMAKE_CXX_COMPILER=`which scorep-CC` \
+#         -DCUDA_NVCC_EXECUTABLE=`which scorep-nvcc`" \
+#         ~/paramSets/case001
+#     export SCOREP_WRAPPER_INSTRUMENTER_FLAGS="--cuda --mpp=mpi"
+#     make -j
+#     make install
+module load scorep/2.0
 
 # plugins (optional) ################################################
 module load cray-hdf5-parallel/1.8.14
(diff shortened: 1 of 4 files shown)


ax3l:
@winklerf-zih thank you for the PR!
I added you to the developer team so the compile suite tests your pull requests. We will have a look on the changes and might comment inline to move things a bit around.

ax3l:
@winklerf-zih do you also want to add a new template file for score-p in `submit/titan-ornl/` ? :) Jaguar is a bit outdated ;)

ax3l:
The changes in
- `src/libPMacc/include/simulationControl/SimulationHelper.hpp` are a good information but we should not merge this generally since this is very instrumenter-specific
- the removal of CMakeList.txt sections due to the new external wrapper usage documented in the [wiki](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/Performance-Analysis/f357fbbc97c9e4d54cff8c8d13247a157c331926) is great, we should merge that
- the `cudaDeviceSynchronize()` is a bit suprising, because we already to that before `cudaDeviceReset()` in our [StreamController](https://github.com/ComputationalRadiationPhysics/picongpu/blob/6bc280eaf6bed7bb0b65a978d653fcee5e8180a5/src/libPMacc/include/eventSystem/streams/StreamController.hpp#L76-L79) - maybe `StreamController` is destructed too late
 -> @psychocoderHPC proposal: move the environment explicitly (init/finalize) into main

winklerf-zih:
@ax3l I added a new scorep template in titan-ornl. Also, I think it is better to have picongpu.profile in $PROJWORK/$projid/$USER, since each user should use his own configuration, see last [commit](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1413/commits/6779381667037d62781d87785fb451fc5dae4fe2)

ax3l:
`$USER` is not so bad, we just shared it during the last INCITE time across the team. but this is actually a bit out of scope for this PR ;)

Can you accept the github team invitation pls so the compile suite can test your commits :)

src/libPMacc/include/simulationControl/SimulationHelper.hpp
```diff
@@ -40,6 +40,13 @@
 #include <iomanip>
 #include <fstream>
 
+#ifdef SCOREP_USER_ENABLE
```
ax3l: please revert the changes in this single file, we can not merge those to the mainline since they are too specific.

src/picongpu/submit/titan-ornl/picongpu.profile.example
```diff
@@ -36,8 +36,22 @@ export MPI_ROOT=$MPICH_DIR
 #   configure with -c "-DVAMPIR_ENABLE=ON"
 #   e.g.:
 #     $PICSRC/configure -c "-DVAMPIR_ENABLE=ON" ~/paramSets/case001
-module load vampirtrace/5.14.4
-export VT_ROOT=$VAMPIRTRACE_DIR
+#module load vampirtrace/5.14.4
+#export VT_ROOT=$VAMPIRTRACE_DIR
+
+
+# scorep (optional) ############################################
```
ax3l: fantastic, thank you! :sparkles:

ax3l:
@winklerf-zih I think the `.tpl` changes are gone now :) pls add those (and pls leave `flushatexit` for now, we can remove it safely when #1414 is implemented)

winklerf-zih:
@ax3l I did the .tpl changes in another [PR](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1417) just for practice :-)

ax3l:
fantastic, thank you!

this PR looks good now, too.

PR #1415, opened 2016-03-22
Improve overflow detection

Flamefire:
This stores an ever-increasing number maxNumProc in addition to the current next id. This is the maximum number of processes ever used to generate ids. With this, an overflow can be detected on all ranks in cases where it was not possible before.

Due to this change the get/setNextId was changed to get/setState which is more appropriate.

ccing @psychocoderHPC

### DIFF
--- a/src/libPMacc/include/particles/IdProvider.hpp
+++ b/src/libPMacc/include/particles/IdProvider.hpp
@@ -64,7 +64,7 @@ namespace PMacc {
         maxNumProc_ = Environment<T_dim>::get().GridController().getGpuNodes().productOfComponents();
         setState(globalUniqueStartId, maxNumProc_);
         // Instantiate kernel
-        getNewId();
+        getNewIdHost();
         // Reset to start value
         setState(globalUniqueStartId, maxNumProc_);
     }
--- a/src/libPMacc/include/particles/IdProvider.hpp
+++ b/src/libPMacc/include/particles/IdProvider.hpp
@@ -56,40 +56,50 @@ namespace PMacc {
     }  // namespace idDetail
 
     template<unsigned T_dim>
-    uint64_t IdProvider<T_dim>::maxNumProc_;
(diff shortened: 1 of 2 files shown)


Flamefire:
@psychocoderHPC This now works as it should and can be merged

src/libPMacc/include/particles/IdProvider.hpp
```diff
@@ -55,29 +56,40 @@ namespace PMacc {
...
         const uint64_t globalUniqueStartId = getStartId();
-        setNextId(globalUniqueStartId);
+        maxNumProc_ = Environment<T_dim>::get().GridController().getGpuNodes().productOfComponents();
+        // Init value to avoid uninitialized read warnings
+        setState(globalUniqueStartId, 0);
         // Instantiate kernel
         getNewIdHost();
         // Reset to start value
-        setNextId(globalUniqueStartId);
+        setState(globalUniqueStartId, maxNumProc_);
```
psychocoderHPC: please use a name without leading `_`

src/libPMacc/include/particles/IdProvider.hpp
```diff
@@ -94,26 +106,33 @@ namespace PMacc {
...
+        // Number of bits in the ids
+        BOOST_STATIC_CONSTEXPR int32_t numBitsOfType = sizeof(maxNumProc_) * CHAR_BIT;
+
+        // Get current value
+        uint64_t nextId = getState().get<0>();
+        /* Example: maxNumProc_ has 3 set bits (<8 ranks), 64bit value used
+         * --> Shift by 61 bits
+         * => 3 upper bits are left untouched (besides moving), rest is zero
+         */
+        nextId >>= numBitsOfType - bitsToCheck;
```
psychocoderHPC: offline discussed: 
- the current implementation is wrong
- we need the start id

ax3l:
@psychocoderHPC updated :)

psychocoderHPC:
Thanks for this great pull request.

PR #1416, opened 2016-03-24
fix int division bug if python 3 is used

PrometheusPi:
This pull request fixes a bug causing array size mismatch in the `smooth.py` routine (used by the `plotRadiation` tool) if run under python 3. 

No other python code in `src/tools` seems to rely on int division.

### DIFF
--- a/src/tools/bin/smooth.py
+++ b/src/tools/bin/smooth.py
@@ -20,6 +20,7 @@
 
 import numpy
 import sys
+from __future__ import division
 
 __doc__ = "This is the 'smooth' module which provides several functions that\n\
 provide methods to smooth data from simulation or experiments.\n\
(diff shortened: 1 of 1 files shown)


PrometheusPi:
@Heikman Are you interested in reviewing this pr?

src/tools/bin/smooth.py
```diff
@@ -20,6 +20,7 @@
 
 import numpy
 import sys
+from __future__import division
```
ax3l: all right, but please add a space between `__future__` and `import`

PrometheusPi: Minimal change and already one error - sorry

PrometheusPi:
@ax3l fixed

ax3l:
ty!

PR #1417, opened 2016-03-24
Changed path for picongpu.profile on lustre.

winklerf-zih:
I think it is better to have picongpu.profile in $PROJWORK/$projid/$USER, since each user should use his own configuration.

### DIFF
--- a/src/picongpu/submit/titan-ornl/batch_profile.tpl
+++ b/src/picongpu/submit/titan-ornl/batch_profile.tpl
@@ -56,7 +56,7 @@ echo -n "present working directory:"
 pwd
 
 
-source $PROJWORK/!TBG_nameProject/picongpu.profile 2>/dev/null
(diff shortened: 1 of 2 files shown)


ax3l:
yes, excellent. thank you! :sparkles:

ax3l:
actually the "right" user-place would be `$MEMBERWORK/!TBG_nameProject/picongpu.profile` with the implication of a much shorter purge time.

ax3l:
@winklerf-zih any comments on this topic?

winklerf-zih:
Well, if you want to share an environment, $PROJWORK/!TBG_nameProject/pigongpu.profile is the best place. But if a user wants to do specific things I prefer $PROJWORK/!TBG_nameProject/$USER rather than $MEMBERWORK/!TBG_nameProject due to a very short purge time.  It is up to you! But it is annoying if you loose your settings after 2 weeks...

ax3l:
makes sense, thanks!

Issue #780, opened 2015-03-26
Radiation Plugin: `RAD_MARK_PARTICLE` flag is obsolete


PrometheusPi:
In the `particleInit.kernel` the `rng()` function [is used](https://github.com/ComputationalRadiationPhysics/picongpu/blob/2d1cd819bbc90f51c23d86413cf34cb0332f3e18/src/picongpu/include/particles/ParticlesInit.kernel#L162-L166) to select a fraction of particles for radiation calculation randomly.
This call is obsolete with the introduction of the new particle initialization pipeline. 

This could be fixed by introducing _radiation_ particles as subset of e.g. electrons in the init pipeline.

ax3l:
@PrometheusPi @psychocoderHPC this is solved by now, right?

psychocoderHPC:
solved with #1866, it can now implemented as manipulator within the species initializer list.

PR #1421, opened 2016-03-30
fix gamma filter for radiation

### LINKED ISSUE #1420: rng not defined in ParticlesInit.kernel
There is a compile time error in the `particles/ParticlesInit.kernel` if the radiation flag is on.

``` shell
ParticlesInit.kernel(165): error: identifier "rng" is undefined
```

This correlates to #780.

PrometheusPi:
This pull request re-enables the gamma filter flag for the radiation plugin. 
It was not usable before. 

This **does not** fix #1420.

### DIFF
--- a/src/picongpu/include/particles/ParticlesInit.kernel
+++ b/src/picongpu/include/particles/ParticlesInit.kernel
@@ -161,10 +161,14 @@ __global__ void kernelFillGridWithParticles(T_GasProfile gasFunctor,
             particle[weighting_] = macroWeighting;
 
 #if(ENABLE_RADIATION == 1)
-#if(RAD_MARK_PARTICLE>1) || (RAD_ACTIVATE_GAMMA_FILTER!=0)
+#    if(RAD_MARK_PARTICLE>1) && (RAD_ACTIVATE_GAMMA_FILTER==0)
             particle[radiationFlag_] = (bool)(rng() < (1.0 / (float_32) RAD_MARK_PARTICLE));
+#    endif
+#    if(RAD_ACTIVATE_GAMMA_FILTER!=0)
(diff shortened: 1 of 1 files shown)


src/picongpu/include/particles/ParticlesInit.kernel
```diff
@@ -161,9 +161,13 @@ __global__ void kernelFillGridWithParticles(T_GasProfile gasFunctor,
             particle[weighting_] = macroWeighting;
 
 #if(ENABLE_RADIATION == 1)
```
ax3l: you could make this nasty `define` cascade a bit more readable by writing:
```C++
#if(ENABLE_RADIATION == 1)
#    if(RAD_MARK_PARTICLE>1) && (RAD_ACTIVATE_GAMMA_FILTER==0)
            particle[radiationFlag_] = (bool)(rng() < (1.0 / (float_32) RAD_MARK_PARTICLE));
#    endif
#    if(RAD_ACTIVATE_GAMMA_FILTER!=0)
            particle[radiationFlag_] = (bool)(false);
#    endif
 #endif
```

ax3l:
thanks for the fix, accidently removed I guess.

Are there plans to remove the `define` crazyness? :) (such as #780)

src/picongpu/include/particles/ParticlesInit.kernel
PrometheusPi: @ax3l I adjusted the spacing.

PrometheusPi:
@ax3l Yes, using species instead of a radiation flag should avoid this `define` madness in the near future.

ax3l:
thanks! :)

PR #1422, opened 2016-04-01
fix documentation (and values) in radiationObserver.param

PrometheusPi:
This pull request fixes wrong documentation in the `radiationObserver.param` 
The laser propagation was wrongly stated as towards `+x direction.
Thanks @Heikman for pointing me to these errors.

### DIFF
--- a/examples/Bunch/include/simulation_defines/param/radiationObserver.param
+++ b/examples/Bunch/include/simulation_defines/param/radiationObserver.param
@@ -65,7 +65,10 @@ namespace picongpu
       /* + picongpu::PI -> turn observation direction 180 degrees towards -y */
 
       /* compute observation unit vector */
-      return vector_64(sinf(theta), cosf(theta), 0.0);
+      picongpu::float_32 sinTheta;
+      picongpu::float_32 cosTheta;
+      math::sincos(precisionCast<picongpu::float_32>(theta), sinTheta, cosTheta);
+      return vector_64(sinTheta, cosTheta, 0.0);
       
     }
     
--- a/examples/Bunch/include/simulation_defines/param/radiationObserver.param
+++ b/examples/Bunch/include/simulation_defines/param/radiationObserver.param
@@ -42,8 +42,8 @@ namespace picongpu
     DINLINE vector_64 observation_direction(const int observation_id_extern)
     {
       /** Computes observation angles along the x-y plane.
-       *  Assuming electron(s) fly in -x direction and the laser
-       *  propages in +x direction, the observation angles are centered
+       *  Assuming electron(s) fly in -y direction and the laser
+       *  propages in +y direction, the observation angles are centered
        *  around the -y-axis (0,-1,0) .
        *  By setting gamma, the angle range can be adjusted to the
        *  energy of the electrons.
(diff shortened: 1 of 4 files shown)


src/picongpu/include/simulation_defines/param/radiationObserver.param
```diff
@@ -96,7 +96,7 @@ namespace picongpu
       const picongpu::float_64 phi(    my_index_phi   * delta_angle_phi   - angle_phi_start   );
 
       /* compute observation unit vector */
-      return vector_64( sinf(theta)*cosf(phi) , sinf(theta)*sinf(phi) , cosf(theta) ) ;
```
psychocoderHPC: please use `mat::sin()`, `math::cos()` or maybe `math::sincos(phi)`
If you need 32bit please cast your values with `precisionCast<float_X>()` before.

ax3l:
let me add the bug flag since its an documentation bug so users can spot it.

src/picongpu/include/simulation_defines/param/radiationObserver.param
PrometheusPi: @psychocoderHPC I fixed the use of `sinf` and `cosf`  - I am not sure why the now removed line does not show up as "outdated diff".

examples/Bunch/include/simulation_defines/param/radiationObserver.param
```diff
@@ -65,7 +65,9 @@ namespace picongpu
       /* + picongpu::PI -> turn observation direction 180 degrees towards -y */
 
       /* compute observation unit vector */
-      return vector_64(sinf(theta), cosf(theta), 0.0);
+      return vector_64(math::sin(precisionCast<picongpu::float_32>(theta)),
```
psychocoderHPC: could you please use `math::sincos()`

examples/SingleParticleRadiationWithLaser/include/simulation_defines/param/radiationObserver.param
```diff
@@ -66,7 +66,9 @@ namespace picongpu
       /* + picongpu::PI -> turn observation direction 180 degrees towards -y */
 
       /* compute observation unit vector */
-      return vector_64(sinf(theta), cosf(theta), 0.0);
+      return vector_64(math::sin(precisionCast<picongpu::float_32>(theta)),
```
psychocoderHPC: could you please use `math::sincos()`

PrometheusPi:
@psychocoderHPC I added `math::sincos` to the two simple cases as well.

examples/Bunch/include/simulation_defines/param/radiationObserver.param
```diff
@@ -65,7 +65,10 @@ namespace picongpu
       /* + picongpu::PI -> turn observation direction 180 degrees towards -y */
 
       /* compute observation unit vector */
-      return vector_64(sinf(theta), cosf(theta), 0.0);
+      picongpu::float_32 sinTheta;
+      picongpu::float_32 cosTheta;
+      math::sincos(precisionCast<picongpu::float_32>(theta), sinTheta, cosTheta);
```
PrometheusPi: Just a **question**:
@psychocoderHPC what would happen if I avoid the `precisionCast` here. Would it run in `double` or `single` precision or would it not run at all?

psychocoderHPC: If depends on `theta`, if it is a 32bit type than float is used else double.

psychocoderHPC:
I removed the example flag because it is more a part of the plugin radiation.

PrometheusPi:
@psychocoderHPC  That is fine with me, I just thought @ax3l likes the `examples` flag, if he builds a new release.

PR #1425, opened 2016-04-12
fix warning in TWTS

psychocoderHPC:
use `DIM2` instead of `simDim` in methods which are specialized to a specific simulation dimension

The compiler evaluates all methods which are specialized for a simulation dimension. In the three dimensional case `pos[i+1]` in the 2d specialization inside a loop from `[0;simDim)` creates a out of array access even this specialization is not used in the three dimensional case.

CC-ing: @BeyondEspresso

### DIFF
--- a/src/picongpu/include/fields/background/templates/TWTS/BField.tpp
+++ b/src/picongpu/include/fields/background/templates/TWTS/BField.tpp
@@ -159,7 +159,7 @@ namespace twts
 
         for (uint32_t k = 0; k<detail::numComponents;++k) {
             /* 2D (y,z) vectors are mapped on 3D (x,y,z) vectors. */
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
I found out that this warning is only triggered if the method are not defined as `force inline`.
This is currently not the case, but we should fix that to avoid future side effects.

ax3l:
looks good to me, default values are `0` and copies what is needed I guess.

@BeyondEspresso pls have a look to confirm.

BeyondEspresso:
Looks OK to me! :+1: At some point we should also refactor the `DIM3` case accordingly in order to make the code consistent with the fix provided here.

ax3l:
great, in it goes!

Issue #732, opened 2015-02-20
ICC: libPMacc support


ax3l:
With the new taurus config (#622), it was a bit hard to find a corresponding [cuda/intel/boost combination](https://gist.github.com/ax3l/9489132) but I found a very recent one.
- `intel/2013-sp1` (aka intel 14.0)
- `boost/1.57.0-intel2013-sp1`
- `cuda/6.5.14`

This pull already fixes a problem in `src/libPMacc/include/math/Tuple.hpp` where `mpl::integral_c` from `<boost/mpl/integral_c.hpp>` was missing an include #627. This pull is a follow-up to re-provide ICC support.

It seems to be in `src/libPMacc/include/math/MapTuple.hpp`:

```
[ 16%] Building NVCC (Device) object build_picongpu/CMakeFiles/picongpu.dir//./picongpu
_generated_main.cu.o
$PICSRC/src/libPMacc/include/math/MapTuple.hpp(107): error: incomplete type is not allowed
          detected during:
            instantiation of class "PMacc::math::MapTuple<Map_, PODType, false> [with M
ap_=boost::mpl::m_mask<picongpu::placeholder_definition15::momentum, boost::mpl::m_mask
<picongpu::placeholder_definition12::position<picongpu::placeholder_definition14::posit
ion_pic, PMacc::placeholder_definition11::pmacc_isAlias>, boost::mpl::m_item<5L, PMacc:
:placeholder_definition8::localCellIdx, PMacc::StaticArray<PMacc::placeholder_definitio
n8::localCellIdx::type, boost::mpl::integral_c<uint32_t, 1U>>, boost::mpl::m_item<4L, p
icongpu::placeholder_definition16::momentumPrev1, PMacc::StaticArray<picongpu::placehol
der_definition14::position_pic::type, boost::mpl::integral_c<uint32_t, 1U>>, boost::mpl
::m_item<3L, picongpu::placeholder_definition17::weighting, PMacc::StaticArray<picongpu
::placeholder_definition17::weighting::type, boost::mpl::integral_c<uint32_t, 1U>>, boo
st::mpl::m_item<2L, picongpu::placeholder_definition15::momentum, PMacc::StaticArray<pi
congpu::placeholder_definition14::position_pic::type, boost::mpl::integral_c<uint32_t, 
1U>>, boost::mpl::m_item<1L, picongpu::placeholder_definition12::position<picongpu::pla
[...]

PR #1426, opened 2016-04-13
Lawrencium (LBNL): Module Updates

ax3l:
- [x] fix git module
- [x] Boost 1.56.0 is not yet available but a support ticket is opened -> they will not update the old gcc chain and do not (yet) have a newer one; we need to build boost ourselves or support icc (#732)


ax3l:
you can merge this for now, we won't get further as "build own boost" or "support icc" right now since they only have an old gcc compile chain and newer icc chains so far.

PR #1427, opened 2016-04-13
libSplash: Update to 1.4.0

ax3l:
The changes in 1.3 (PDC) and 1.4 (SDC) are needed to write the [openPMD standard](https://github.com/openPMD/openPMD-standard/tree/1.0.0).

This changes the API of libSplash and also the files created by it (bool representations, some internal vars, pathes of global attributes).
For more information, please read the libSplash [change log](https://github.com/ComputationalRadiationPhysics/libSplash/blob/v1.4.0/CHANGELOG.md).

### PR's that will need to rebase *after* this PR was merged

- #1053 (enables correct global attributes for this PR)
- #1376 (enables correct global attributes for this PR)
- #1410 (might be unaffected)
- the upcoming `topic-openPMD` PR of mine

PR #1428, opened 2016-04-13
openPMD 1.0.0: Base Standard in HDF5

ax3l:
### Description

@psychocoderHPC 

This PR implements:

- the openPMD base standard, [v1.0.0](https://github.com/openPMD/openPMD-standard/tree/1.0.0), for our HDF5 output
- much of the [`ED-PIC` extension](https://github.com/openPMD/openPMD-standard/blob/1.0.0/EXT_ED-PIC.md) (just missing a few string-attributes)
- constant particle records: `mass`, `charge`
- [`particlePatches`](https://github.com/openPMD/openPMD-standard/blob/1.0.0/STANDARD.md#sub-group-for-each-particle-species) instead of the old `particle_info` object
- restarts from same GPU configuration (tested with `-m` in LWFA example)
- checkpoints and dumps tested with openPMD validator scripts (base standard is passing)
- change: unitless attributes now have the `unitSI` of `1.0` since we can identify those already via an all-zero `unitDimension`

### Requirements

- [x] requires libSplash `1.4.0` #1427 
- [x] fields: `getUnitDimension()` trait/interface #1429 (will remove the e6e396fdc3561214cf4ef93640297fe0ba39d890 commit)
- [x] particles: `PatchReader` helper #1430 (will remove the ~260 new lines in the helper class of the bebaaf39056ac5cc696277a7e31bed58339b7efc commit and speeds up compile time due to separate object files)
- [x] particle weighting traits: #1445
- [x] rebase
- [x] after review, before merge: redo 3D3V rt test
- [x] after review, before merge: perform missing 2D3V rt test
- [x] read with [openPMD-viewer](https://github.com/openPMD/openPMD-viewer)

### Possible Collisions

- [x] check against #1410 (rebase works smoothly, so I will review & merge #1410 while you review this PR)
- [x] apply RT fix from #1440
- [x] rebase against #1453

### Follow-Ups

- fields: further, general ED-PIC attributes:
  - `fieldSolver`, `fieldBoundary`, `particleBoundary`, `currentSmoothing`, `chargeCorrection`
- fields: per-field smoothing attribute: `fieldSmoothing`
- particles: further, general ED-PIC per-species attributes:
  - `particleShape`, `currentDeposition`, `particlePush`, `particleInterpolation`, `particleSmoothing`
- ~~particles: per-species-record attribute: `weightingPower`, `macroWeighted`~~ -> #1445
- particles: non-`uint64_t` types in `extent` and `offset` of `particlePatches`

### DIFF
--- a/src/picongpu/include/fields/FieldB.tpp
+++ b/src/picongpu/include/fields/FieldB.tpp
@@ -229,7 +229,7 @@ FieldB::getUnitDimension( )
 std::string
 FieldB::getName( )
 {
-    return "FieldB";
+    return "B";
 }
 
 uint32_t
--- a/src/picongpu/include/plugins/hdf5/WriteSpecies.hpp
+++ b/src/picongpu/include/plugins/hdf5/WriteSpecies.hpp
@@ -178,6 +178,9 @@ public:
         ColTypeDouble ctDouble;
         GridController<simDim>& gc = Environment<simDim>::get().GridController();
 
+        const uint64_t numRanks( gc.getGlobalSize() );
+        const uint64_t myRank( gc.getGlobalRank() );
+
         /* For collective write calls we need the information:
          *   - how many particles will be written globally
          *   - what is my particle offset within this global data set
@@ -188,10 +191,10 @@ public:
          * the mpi rank is an arbitrary quantity and might change after a
          * restart, but we only use it to order our patches and offsets
          */
-        std::vector<uint64_t> particleCounts( 2 * gc.getGlobalSize(), 0u );
+        std::vector<uint64_t> particleCounts( 2 * numRanks, 0u );
         uint64_t myParticlePatch[ 2 ];
         myParticlePatch[ 0 ] = numParticles;
-        myParticlePatch[ 1 ] = uint64_t(gc.getGlobalRank());
+        myParticlePatch[ 1 ] = myRank;
 
         /* we do the scan over MPI ranks since it does not matter how the
          * global rank or scalar position (which are not idential) are
@@ -206,7 +209,7 @@ public:
             gc.getCommunicator().getMPIComm()
         ));
 
-        for( uint64_t r = 0; r < gc.getGlobalSize(); ++r )
+        for( uint64_t r = 0; r < numRanks; ++r )
         {
             numParticlesGlobal += particleCounts.at(2 * r);
             if( particleCounts.at(2 * r + 1) < myParticlePatch[ 1 ] )
@@ -239,23 +242,32 @@ public:
             FrameType::getName() + std::string("/") + subGroup +
             std::string("/particlePatches") );
 
+        /* offset and size of our particle patches
+         *   - numPatches: we write as many patches as MPI ranks
+         *   - myPatchOffset: we write in the order of the MPI ranks
+         *   - myPatchEntries: every MPI rank writes exactly one patch
+         */
+        const Dimensions numPatches( numRanks, 1, 1 );
+        const Dimensions myPatchOffset( myRank, 0, 0 );
+        const Dimensions myPatchEntries( 1, 1, 1 );
+
         /* numParticles: number of particles in this patch */
         params->dataCollector->write(
             params->currentStep,
-            Dimensions(gc.getGlobalSize(), 1, 1),
-            Dimensions(gc.getGlobalRank(), 0, 0),
+            numPatches,
+            myPatchOffset,
             ctUInt64, 1,
-            Dimensions(1, 1, 1),
+            myPatchEntries,
             (particlePatchesPath + std::string("/numParticles")).c_str(),
             &numParticles);
 
         /* numParticlesOffset: number of particles before this patch */
         params->dataCollector->write(
             params->currentStep,
-            Dimensions(gc.getGlobalSize(), 1, 1),
-            Dimensions(gc.getGlobalRank(), 0, 0),
+            numPatches,
+            myPatchOffset,
             ctUInt64, 1,
-            Dimensions(1, 1, 1),
+            myPatchEntries,
             (particlePatchesPath + std::string("/numParticlesOffset")).c_str(),
             &numParticlesOffset);
 
@@ -275,19 +287,19 @@ public:
 
             params->dataCollector->write(
                 params->currentStep,
-                Dimensions(gc.getGlobalSize(), 1, 1),
-                Dimensions(gc.getGlobalRank(), 0, 0),
+                numPatches,
+                myPatchOffset,
                 ctUInt64, 1,
-                Dimensions(1, 1, 1),
+                myPatchEntries,
                 (particlePatchesPath + std::string("/offset/") +
                  name_lookup[d]).c_str(),
                 &patchOffset);
             params->dataCollector->write(
                 params->currentStep,
-                Dimensions(gc.getGlobalSize(), 1, 1),
-                Dimensions(gc.getGlobalRank(), 0, 0),
+                numPatches,
+                myPatchOffset,
                 ctUInt64, 1,
-                Dimensions(1, 1, 1),
+                myPatchEntries,
                 (particlePatchesPath + std::string("/extent/") +
                  name_lookup[d]).c_str(),
                 &patchExtent);
--- a/src/picongpu/include/plugins/hdf5/WriteSpecies.hpp
+++ b/src/picongpu/include/plugins/hdf5/WriteSpecies.hpp
@@ -26,6 +26,7 @@
 #include "pmacc_types.hpp"
 #include "simulation_types.hpp"
 #include "plugins/hdf5/HDF5Writer.def"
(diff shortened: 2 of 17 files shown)


src/picongpu/include/fields/FieldB.tpp
```diff
@@ -229,7 +229,7 @@ FieldB::getUnitDimension( )
 std::string
 FieldB::getName( )
 {
-    return "FieldB";
+    return "B";
```
psychocoderHPC: If we change this name we need to change the hdf5 tools in `src/tools`, too.

src/picongpu/include/traits/PICToOpenPMD.hpp
```diff
@@ -0,0 +1,110 @@
...
+    struct OpenPMDName
+    {
+        std::string operator()() const
+        {
+            return T_Identifier::getName();
+        }
+    };
+
+    template<typename T_Type>
+    struct OpenPMDName<globalCellIdx<T_Type> >
```
psychocoderHPC: please move the specializations to a `*.unitless`
  - it separates the traits definition from specializations
  - if new attributes in `speciesAttributes.param` are created the place for the trait specialization is more clear (not inside the PIConGPU` code

Maybe our intention is that `globalCellIdx` is no user attributes, if so please create a `*.tpp` file in the same directory and add the specializations there.

src/picongpu/include/traits/PICToOpenPMD.hpp
```diff
@@ -0,0 +1,110 @@
...
+    struct OpenPMDUnit
+    {
+        std::vector<double> operator()() const
+        {
+            return Unit<T_Identifier>::get();
+        }
+    };
+
+    template<typename T_Type>
+    struct OpenPMDUnit<globalCellIdx<T_Type> >
```
psychocoderHPC: Why we not add this definition to the trait `getUnit<>`?

src/picongpu/include/traits/PICToOpenPMD.hpp
```diff
@@ -0,0 +1,110 @@
...
+    {
+        std::vector<float_64> operator()() const
+        {
+            /* L, M, T, I, theta, N, J
+             *
+             * positionOffset is in meter: m
+             *   -> L
+             */
+            std::vector<float_64> unitDimension( NUnitDimension, 0.0 );
+            unitDimension.at(SIBaseUnits::length) = 1.0;
```
psychocoderHPC: This is already defined in the trait `UnitDimensions`

src/picongpu/include/traits/PICToOpenPMD.hpp
psychocoderHPC: Maybe the specialization for `globalCellIdx` in `getUnit<>` is wrong?

src/picongpu/include/fields/FieldB.tpp
ax3l: good catch, thank you!

src/picongpu/include/traits/PICToOpenPMD.hpp
ax3l: > please move the specializations to a *.unitless

this is an other abstraction between **PIConGPU** and **openPMD** and is therefore not in a `unitless` file (which converts between **user input** and **PIConGPU**).

We should set up a short offline meeting to make it more clear.

> Maybe your intention is that globalCellIdx is no user attributes, if so please create a *.tpp file in the same directory and add the specializations there. 

this might be better, let us talk offline to make sure we are talking about the same :)

src/picongpu/include/traits/PICToOpenPMD.hpp
ax3l: nope it is the same as above: I convert a PIConGPU quantity into a quantity known in openPMD. `openPMD` does intentionally expect a `positionOffset` and not a cell id to make it more general in the `base standard`. We therefore convert our `cellIdx` into an `offset` which is an actual *length* :)

src/picongpu/include/traits/PICToOpenPMD.hpp
ax3l: uh, maybe I want to have it without a `length` dimension there, will cross-check!

ax3l: yes, `globalCellIdx` is an actual index and I will change `speciesAttributes.unitless` to make it unitless.

src/picongpu/include/traits/PICToOpenPMD.hpp
ax3l: I think you actually understood already what I am trying to do: splitted the specialization into a `.tpp` file, good point!

Can still meet if you want to talk.

src/picongpu/include/plugins/hdf5/WriteSpecies.hpp
```diff
@@ -157,48 +164,176 @@ struct WriteSpecies
...
+        log<picLog::INPUT_OUTPUT > ("HDF5:  (begin) writing particlePatches for %1%") % Hdf5FrameType::getName();
+
+        std::string particlePatchesPath( std::string("particles/") +
+            FrameType::getName() + std::string("/") + subGroup +
+            std::string("/particlePatches") );
+
+        /* numParticles: number of particles in this patch */
+        params->dataCollector->write(
+            params->currentStep,
+            Dimensions(gc.getGlobalSize(), 1, 1),
```
psychocoderHPC: only a note: 
Because of the multiple usages of `gc.getGlobalSize()` and `gc.getGlobalRank()` a helper variable would help to read the code.
`gc.getGlobalRank()` == the offset inside the array which is written
`gc.getGlobalRank()` == the size of the array

src/picongpu/include/plugins/hdf5/writer/Field.hpp
```diff
@@ -135,14 +155,80 @@ struct Field
...
+                                              ctDataOrder, recordName.c_str(),
+                                              "dataOrder", dataOrder.c_str());
+
+        typedef char MyChar2[2];
+        MyChar2 *axisLabels = new MyChar2[simDim];
+        ColTypeString ctAxisLabels(1);
+        for( uint32_t d = 0; d < simDim; ++d )
+        {
+            /* \todo is the order correct? */
+            axisLabels[d][0] = char(120 + d); // x, y, z
```
psychocoderHPC: Must this description match to the plain data in HDF5 or to the attribute (written with `writeAttribute()`) order?

Note: 
- HDF5 store its data in Fortran style 
- an array written as attributes in `libSplash` is not changed in the representation "x y z" stays "x y z"

src/tools/bin/splash2vtk.sh
```diff
@@ -21,8 +21,8 @@
 
 filename="$1"
 step=1000
-dataNames="fields_FieldE_y"
-#dataNames="fields_FieldE_y"
+dataNames="fields_E_y"
+#dataNames="fields_E_y"
```
psychocoderHPC: This can be removed but because it was in before you could keep it.

psychocoderHPC:
Thanks for this pull request.

I added only two notes but all in all it is ready to merge if your tests are finished.

src/picongpu/include/plugins/hdf5/WriteSpecies.hpp
ax3l: great idea, I just added some const vars with self-describing names and a 3 line doc string to describe the libSplash API intrinsically with it

src/picongpu/include/plugins/hdf5/writer/Field.hpp
ax3l: - must match the data
- seems to fit from my first tests with `openPMD-viewer`

I leave the \todo in it still for later checks when the full `ED-PIC` extension is implemented

src/tools/bin/splash2vtk.sh
ax3l: removed

src/picongpu/include/plugins/hdf5/writer/Field.hpp
psychocoderHPC: please use `char('x' + d)` instead of the magic ascii number. Maybe a cast `uint8_t('x')` is needed.

ax3l: ok, `char( 'x' + d )` is fully valid and compiles with `-Wall -Wextra -pedantic` in `g++ 4.9.2`:

```c++
#include <iostream>

int main()
{
  char c = 'a';
  unsigned int d = 1;

  c = char('x' + d);
  std::cout << c << std::endl;
  return 0;
}
```

ax3l:
I add an other 17 lines when you merge #1445 first. This implements an other important attribute for `ED-PIC` so one can directly get started with [openPMD-viewer](https://github.com/openPMD/openPMD-viewer).

PR #1429, opened 2016-04-13
Fields: Add getUnitDimension() Interface

ax3l:
Added alongside `getUnit()` the new interface to get the openPMD dimensionality of a data set for all basic fields and `FieldTmp` solvers.

### DIFF
--- a/src/picongpu/include/fields/FieldB.tpp
+++ b/src/picongpu/include/fields/FieldB.tpp
@@ -53,6 +53,7 @@
 #include "particles/traits/FilterByFlag.hpp"
 
 #include "traits/GetMargin.hpp"
+#include "traits/SIBaseUnits.hpp"
 #include "particles/traits/GetMarginPusher.hpp"
 
 namespace picongpu
@@ -218,9 +219,9 @@ FieldB::getUnitDimension( )
      *   -> M * T^-2 * I^-1
      */
     std::vector<float_64> unitDimension( 7, 0.0 );
-    unitDimension.at(1) =  1.0; // M^1
-    unitDimension.at(2) = -2.0; // T^-2
-    unitDimension.at(3) = -1.0; // I^-1
+    unitDimension.at(SIBaseUnits::mass) =  1.0;
+    unitDimension.at(SIBaseUnits::time) = -2.0;
+    unitDimension.at(SIBaseUnits::electricCurrent) = -1.0;
 
     return unitDimension;
 }
--- a/src/picongpu/include/fields/FieldB.tpp
+++ b/src/picongpu/include/fields/FieldB.tpp
@@ -208,6 +208,23 @@ FieldB::getUnit( )
     return UnitValueType( UNIT_BFIELD, UNIT_BFIELD, UNIT_BFIELD );
 }
 
+HDINLINE
+std::vector<float_64>
+FieldB::getUnitDimension( )
+{
+    /* L, M, T, I, theta, N, J
+     *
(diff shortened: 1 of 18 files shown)


ax3l:
@PrometheusPi feel free to have a look checking my units, too! :)

src/picongpu/include/fields/FieldB.tpp
```diff
@@ -208,6 +208,23 @@ FieldB::getUnit( )
...
+std::vector<float_64>
+FieldB::getUnitDimension( )
+{
+    /* L, M, T, I, theta, N, J
+     *
+     * B is in Tesla : kg / (A * s^2)
+     *   -> M * T^-2 * I^-1
+     */
+    std::vector<float_64> unitDimension( 7, 0.0 );
+    unitDimension.at(1) =  1.0; // M^1
```
psychocoderHPC: only a note: I opened #1431 to allow a more human readable definition that avoids errors with wrong index usage.

psychocoderHPC: Thus `SIBaseUnits` are defined in `unitless` we can use it everywhere.
Could you please use `SIBaseUnits::mass`, ... to avoid wrong index usage

psychocoderHPC:
There is already a trait `UnitDimension` with the same functionality like this new interface.
IMO it is better to defines this trait for `FieldE`, `FieldB`, ... to stay compatible with the way how we get the unit dimensions out of species attributes.

Never the less we can merge this interface extension first and refactor it in a new pull request to a uniform way.

src/picongpu/include/particles/particleToGrid/derivedAttributes/ChargeDensity.def
```diff
@@ -35,6 +36,22 @@ namespace derivedAttributes
...
+        HDINLINE std::vector<float_64>
+        getUnitDimension() const
+        {
+           /* L, M, T, I, theta, N, J
+            *
+            * ChargeDensity is in Coulomb / cubic meter: Q / m^3 = I * t / m^3
+            *   -> L^-3 * T * I
+            */
+           std::vector<float_64> unitDimension( 7, 0.0 );
+           unitDimension.at(0) = -3.0; // L^-3
```
psychocoderHPC: Is it for a 2D simulation also `L^-3`?

psychocoderHPC:
@PrometheusPi  is currently not at work, if you need this pull request merged today please write me an email.
I checked your all your units again Wikipedia^^.

src/picongpu/include/fields/FieldB.tpp
ax3l: ah great idea!
yes, forgot we improved that in the previous PR!

ax3l:
adding your comments.

regarding traits: in the case as here, where we are doing *endpoint* implementations which are not intended to be *derived* again (but just placed side-by-side) I personally find interface implementations way more intuitive and less error-prune (same file, on-the-spot, self-describing the class) then traits that are scattered. traits also bring no added value when one is not planning to derive again (which would obfuscate in such a scenario more then it helps).

src/picongpu/include/particles/particleToGrid/derivedAttributes/ChargeDensity.def
ax3l: ;) good quesiton and actually the same for `FieldJ`.

Currently, as in `getUnit()`, we just implement the 3D unit.

PR #1430, opened 2016-04-13
openPMD: Particle Patch Reader

ax3l:
@psychocoderHPC @erikzenker 

Implements a helper class `PatchReader` to read libSplash HDF5 files with particle patches and to build up a class containing the full list of particle patches for each rank.

For storing the read patches in RAM on each node, a common `ParticlePatches` storage class was implemented (that can also be used with the ADIOS plugin later on).

Compilation of the HDF5 class only occures when libSplash was found (guarded by `ENABLE_HDF5` macros).

Usage, see `plugins/hdf5/restart/LoadSpecies.hpp` in #1428

```C++
#include "plugins/common/particlePatches.hpp"
#include "plugins/hdf5/openPMD/patchReader.hpp"

// [...]
const std::string speciesSubGroup(
    std::string("particles/") + FrameType::getName() + std::string("/")
);
// load particle patches offsets to find own patch
const std::string particlePatchesPath(
    speciesSubGroup + std::string("particlePatches/")
);

// read particle patches
openPMD::PatchReader patchReader;

openPMD::ParticlePatches particlePatches(
    patchReader(
        params->dataCollector,
        gc.getGlobalSize(),
        simDim,
        params->currentStep,
        particlePatchesPath
    )
);

// do something with the particlePatches object to find your particles
```

If you are wondering about the `.cpp` files in `src/picongpu`: this is a bit of CMake globber messy-ness and we need to clean that up an other time for the few object files we generate.

### DIFF
--- a/src/picongpu/include/plugins/common/particlePatches.hpp
+++ b/src/picongpu/include/plugins/common/particlePatches.hpp
@@ -1,0 +1,102 @@
+/**
+ * Copyright 2016 Axel Huebl
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "pmacc_types.hpp"
+
+#include <vector>
+#include <list>
+#include <iostream>
+
+namespace picongpu
+{
+namespace openPMD
+{
+
+    /** Struct for a list of particle patches
+     *
+     * Object for all particle patches.
+     * @see https://github.com/openPMD/openPMD-standard/blob/1.0.0/STANDARD.md#sub-group-for-each-particle-species
+     */
+    class ParticlePatches
+    {
+    private:
+        /** Disallow (empty) default contructor
+         */
+        ParticlePatches ();
+
+    public:
+        std::vector<uint64_t> numParticles;
+        std::vector<uint64_t> numParticlesOffset;
+
+        std::vector<uint64_t> offsetX;
+        std::vector<uint64_t> offsetY;
+        std::vector<uint64_t> offsetZ;
+
+        std::vector<uint64_t> extentX;
+        std::vector<uint64_t> extentY;
+        std::vector<uint64_t> extentZ;
+
+        /** Fill-Constructor with n empty-sized patches
+         *
+         * @param n number of patches to store
+         */
+        ParticlePatches( const size_t n );
+
+        /** Return the beginning of one of the components of the
+         *  offset as pointer
+         *
+         * Be aware that the pointer is pointing to the beginning
+         * of a C-array of size `size()` and is only allocated as long
+         * as the `ParticlePatches` object is alive.
+         *
+         * @param comp component (0=x, 1=y, 2=z) of offset array
+         *             for the list of patches
+         * @return uint64_t* pointing to the beginning of a c-array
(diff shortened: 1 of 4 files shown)


src/picongpu/include/plugins/hdf5/openPMD/patchReader.hpp
```diff
@@ -0,0 +1,115 @@
...
+         *             "particles/e/particlePatches/offset/x"
+         * @param[out] dest beginning of c-array of length size()
+         *             to write the patch record component to
+         */
+        void readPatchAttribute(
+            splash::DataCollector* const dc,
+            const uint32_t availableRanks,
+            const int32_t id,
+            const std::string particlePatchPathComponent,
+            uint64_t* const dest
```
psychocoderHPC: The destination array for yout attributes is always defined as `uint64_t` but the [openPMD standard](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md#sub-group-for-each-particle-species) defines that `offset` and `extent` were of the type `float`,`int` or `uint`.

IMO this is currently not a problem if PIConGPU writes all attributes as 64bit values but than it is not standard conform.

ax3l: absolutely correct, but for now I want to read our files.

for that reason line 55 implements a check and an assert before the data is even read into RAM.

ax3l: the idea is: if someone needs other offsets already -> tiny pre-process script in python.

later: clean conversions in c++ on our side (too much boilerplate for now).

src/picongpu/include/plugins/common/particlePatches.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+        /** Return the beginning of one of the components of the
+         *  offset as pointer
+         *
+         * @param comp component (0=x, 1=y, 2=z) of offset array
+         *             for the list of patches
+         * @return uint64_t* pointing to the beginning of a c-array
+         *                   with length as given in size()
+         */
+        uint64_t* getOffsetComp( const uint32_t comp );
```
psychocoderHPC: Could you please add a note to the description that the returned pointer is only valid as long the instance of `ParticlePatches` is alive.

src/picongpu/include/plugins/common/particlePatches.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+        /** Return the beginning of one of the components of the
+         *  extent as pointer
+         * @param comp component (0=x, 1=y, 2=z) of extent array
+         *             for the list of patches
+         * @return uint64_t* pointing to the beginning of a c-array
+         *                   with length as given in size()
+         */
+        uint64_t* getExtentComp( const uint32_t comp );
```
psychocoderHPC: Please add a note to the live time of the returned pointer.

src/picongpu/include/plugins/common/particlePatches.hpp
ax3l: yes sure!

src/picongpu/include/plugins/hdf5/openPMD/patchReader.hpp
```diff
@@ -0,0 +1,115 @@
...
+         * one was used and how many patches exist.
+         *
+         * @note currently we force the type to be `uint64_t`,
+         *       we can implement type conversions later on
+         * @note currently we force the number of patches
+         *       to stay constant during restarts
+         *
+         * @see operator() for parameters
+         */
+        void checkSpatialTypeSize(
```
psychocoderHPC: please also describe all parameters.

src/picongpu/include/plugins/common/particlePatches.hpp
ax3l: will do

src/picongpu/include/plugins/hdf5/openPMD/patchReader.hpp
ax3l: `@see` ;)

psychocoderHPC:
Thx wait for the compile suite.

src/picongpu/include/plugins/hdf5/openPMD/patchReader.hpp
ax3l: the latest libSplash updates such as https://github.com/ComputationalRadiationPhysics/libSplash/pull/248 will reduce that boilerplate in the future

PR #1432, opened 2016-04-14
fix cuSTL-DeviceBuffer to enable device-to-device copies.

Heikman:
This is supposed to be the final solution to the well known (#1375, #1401) DeviceBuffer-to-DeviceBuffer copy problem. The problem seemed to be fixed with #1401 but it reappeared in the ParticleCalorimeter plugin branch at a line similar to this one: `*dBufA_ptr = *dBufB_ptr;`.

Solves #1434 

@Flamefire

### DIFF
--- a/src/libPMacc/include/cuSTL/container/DeviceBuffer.hpp
+++ b/src/libPMacc/include/cuSTL/container/DeviceBuffer.hpp
@@ -23,9 +23,16 @@
 
 #pragma once
 
+/* device buffer policies */
 #include <cuSTL/container/allocator/DeviceMemAllocator.hpp>
 #include <cuSTL/container/copier/D2DCopier.hpp>
 #include <cuSTL/container/assigner/DeviceMemAssigner.hpp>
+
+/* host buffer policies */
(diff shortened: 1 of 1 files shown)


Flamefire:
Strange... Can you point me to the line where the bug without this PR is shown? Especially what are the types? The error message would also be very helpful for that, as I'd think this PR might not be required. But I just want to know, what exactly is going on

Heikman:
This solution is not better or worse than the dev version. I think it's even more explicit now. I need this PR for #1376 now.

Heikman:
I'm sorry there is of course the question why this error appeared, it should work without this PR, it's just that I don't want to care about internal compiler selection rules so much, as soon as I have an acceptable solution.

psychocoderHPC:
Please post the error message, currently that is not clear what goes wrong.

Heikman:
Here is the issue #1434, just for you guys.

Flamefire:
The linked issue does not make clear why this must be fixed with a code change rather than just adding the missing includes

Flamefire:
Superseded by #1435 which does not require the additional includes and fixes similar issue with `HostBuffer`

PR #1445, opened 2016-04-22
Particles: Weighting Traits

ax3l:
This commit adds a new trait to the particle attributes describing each attribute's correlation in PIConGPU regarding macro particles.

### MacroWeighted

Depending on the implementation of an attribute, it might be sometimes useful to return a quantity regarding one of the underlying real particles (false: "this attribute is not weighted accordingly for the whole ensemble of particles in its macro particle) or just handle the whole macro particle at once (true: "this attribute is already weighted").

This trait defines for each attribute if it needs to be scaled with the weighting. *How* the scaling with weighting is applied can be queried in the trait `WeightingPower`.

### WeightingPower

This trait defines how each attribute needs to be scaled with the weighting (linear, quadratic, ...) to convert between "real" and "macro" particle attributes.

### Further references

Besides the fact that these new traits describe particle attributes *inside* the code better and consistenly, they will also be used in the `ED-PIC` extension of `openPMD`.

For futher information please refer to the `openPMD 1.0.0` documents:
 - https://git.io/vwlWa
 - http://www.openPMD.org
 - http://dx.doi.org/10.5281/zenodo.33624

cc'ing @psychocoderHPC @PrometheusPi

### DIFF
--- a/src/picongpu/include/simulation_defines/unitless/speciesAttributes.unitless
+++ b/src/picongpu/include/simulation_defines/unitless/speciesAttributes.unitless
@@ -27,6 +27,8 @@
 #include "traits/Unit.hpp"
 #include "traits/UnitDimension.hpp"
 #include "traits/SIBaseUnits.hpp"
+#include "particles/traits/MacroWeighted.hpp"
+#include "particles/traits/WeightingPower.hpp"
 
 namespace picongpu
 {
@@ -41,7 +43,8 @@ struct Unit<position<T_Type> >
     {
         std::vector<double> unit(simDim);
         /* in-cell position needs two transformations to get to SI:
-           in-cell [0;1) -> dimensionless scaling to grid -> SI */
+           in-cell [0;1) -> dimensionless scaling to grid -> SI
+        */
         for(uint32_t i=0;i<simDim;++i)
             unit[i]=cellSize[i]*UNIT_LENGTH;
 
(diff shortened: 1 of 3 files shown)


src/picongpu/include/simulation_defines/unitless/speciesAttributes.unitless
```diff
@@ -64,6 +66,26 @@ struct UnitDimension<position<T_Type> >
...
+     */
+    static bool get()
+    {
+        return false;
+    }
+};
+template<typename T_Type>
+struct WeightingPower<position<T_Type> >
+{
+    /* x * weighting^0 == x: same for real and macro particle
```
psychocoderHPC: Could you change one line comments to one line?

`/* comment */` or `// comment`

ax3l: no

ax3l: ;P

PrometheusPi:
@ax3l Is this required with the current openPMD standard or will this be part of the standard in the future?

ax3l:
it's only part of the [`ED-PIC` extension](https://github.com/openPMD/openPMD-standard/blob/1.0.0/EXT_ED-PIC.md#additional-attributes-for-each-particle-record), that means your radiation output does not need it since you are using only the `base standard` :)

PR #1449, opened 2016-04-29
superfluous and missing include

mxmlnkn:


ax3l:
looks all right, thank you!

src/libPMacc/include/nvidia/functors/Assign.hpp
```diff
@@ -22,8 +22,6 @@
 
 #pragma once
 
-#include "pmacc_types.hpp"
```
BenjaminW3: The `HDINLINE` used in this file is defined in `pmacc_types.hpp` so I think this is not superfluous.

ax3l: oh dang you are right!

ax3l: arrr, I must have been sleeping.... :S

mxmlnkn: Sorry for that, I didn't know that.
At least the review process works in the end.

ax3l: no prob! :)

PR #1319, opened 2016-01-05
Trait: FieldPosition

ax3l:
Refactors the getter of field positions into a trait. Read commit-wise (commit 1: refactor, commit 2: update usage in code).

Adds the previously just commented positions of FieldTmp (currently used in ParticleToGrid always at zero corner.)

Used in the upcoming openPMD branch.

PR #1453, opened 2016-05-02
FieldPosition Trait: ReturnType of FieldTmp Consistent

Heikman:
Fix `FieldPosition` trait for `FieldTmp`. ~~The vector dimensions were mixed up.~~

The `FieldTmp` trait was not wrappend in it's one-dimensional vector-vector wrapper.

### DIFF
--- a/src/picongpu/include/fields/numericalCellTypes/EMFCenteredCell.hpp
+++ b/src/picongpu/include/fields/numericalCellTypes/EMFCenteredCell.hpp
@@ -123,8 +123,9 @@ namespace traits
         }
     };
 
-    /** position (floatD_X in case of T_simDim == simDim) in cell for the
-     *  scalar field FieldTmp
+    /** position (floatD_X in case of T_simDim == simDim) in cell, wrapped in
+     * one-component vector since it's a scalar field with only one component, for the
+     * scalar field FieldTmp
      */
     template<uint32_t T_simDim>
     struct FieldPosition<FieldTmp, T_simDim>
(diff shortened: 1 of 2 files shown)


ax3l:
careful: needs update in HDF5 output, too: `src/picongpu/include/plugins/hdf5/WriteFields.hpp`

ax3l:
ah that's in #1428 ....

ax3l:
refactoring of #1319, not used yet besides in `openPMD` PR.

src/picongpu/include/fields/numericalCellTypes/EMFCenteredCell.hpp
```diff
@@ -129,7 +129,8 @@ namespace traits
     template<uint32_t T_simDim>
     struct FieldPosition<FieldTmp, T_simDim>
```
ax3l: pls update doxygen string:
```C++
/** position (floatD_X in case of T_simDim == simDim) in cell, wrapped in
 * one-component vector since it's a scalar field with only one component, for the
 * scalar field FieldTmp
 */
```

src/picongpu/include/fields/numericalCellTypes/YeeCell.hpp
```diff
@@ -150,7 +150,8 @@ namespace traits
     template<uint32_t T_simDim>
     struct FieldPosition<FieldTmp, T_simDim>
```
ax3l: pls update doxygen string as above

ax3l:
thanks! :sparkles:

PR #1454, opened 2016-05-03
Close #1424: add species name to file name of hdf5 rad files

### LINKED ISSUE #1424: Radiation restart does not support multiple species
The current implementation of the radiation restart does not support multiple species.
Due to the hard coded file names [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/ff856bc7fa3d36f43488012dd6803747e5b53719/src/picongpu/include/plugins/radiation/Radiation.hpp#L254) all species write into the same hdf5 file, thus overwriting files from other species.

PrometheusPi:
This pull request enables radiation restart and any multi species hdf5 output with more than one radiating species. This closes issue #1424.

 - [x] test with LWFA ions and electrons

### DIFF
--- a/src/picongpu/include/plugins/radiation/Radiation.hpp
+++ b/src/picongpu/include/plugins/radiation/Radiation.hpp
@@ -179,14 +179,14 @@ public:
             // end
             if (currentStep <= radEnd || radEnd == 0)
             {
-                log<radLog::SIMULATION_STATE > ("Radiation: calculate time step %1% ") % currentStep;
+                log<radLog::SIMULATION_STATE > ("Radiation (%1%): calculate time step %2% ") % speciesName % currentStep;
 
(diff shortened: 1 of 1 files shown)


ax3l:
does this change anything in the usage and if yes, is it documented in the according wiki article? :)

PrometheusPi:
No, from the user side this does not change anything except that writing checkpoints for two and more radiating species does not crash/hang anymore. 

I will update the wiki, stating that we support multi species restart.

PrometheusPi:
@ax3l I updated the wiki.

PrometheusPi:
@BeyondEspresso This pull request allows to separately calculate the radiation of electrons from different ion species while using checkpoints/restarts.

ax3l:
great, looks good. the `doc/TBG_macros.cfg` are already updated since the last PR I guess :)

Issue #755, opened 2015-03-13
Document the "Generic Change" User


ax3l:
Add to `CONTRIBUTION.md`:

```bash
GIT_AUTHOR_NAME="Third Party" GIT_AUTHOR_EMAIL="picongpu@hzdr.de" \
  git [...]
```

PR #1455, opened 2016-05-03
Reintroduce PMacc Include

ax3l:
Follow-up to #1449, thanks to @BenjaminW3 for spotting my incorrect merge.

### DIFF
--- a/src/libPMacc/include/nvidia/functors/Assign.hpp
+++ b/src/libPMacc/include/nvidia/functors/Assign.hpp
@@ -22,6 +22,8 @@
 
 #pragma once
 
+#include "pmacc_types.hpp"
+
 namespace PMacc
 {
     namespace nvidia


erikzenker:
Well, sometimes it is hard to see were some type is coming from. We can just hope that our IDEs will warn us in the future. Otherwise, exhausting automatic header checks can help.

ax3l:
in principle yes, but in this case it was obvious and the only include ^^

BenjaminW3:
alpaka has such self-contained header checks in the CI. However, they are slow as each header has to be built by a single `.cpp` file (especially when using the nvcc compiler).

src/libPMacc/include/nvidia/functors/Assign.hpp
```diff
@@ -22,6 +22,8 @@
 
 #pragma once
 
+#include "pmacc_types.hpp"
+
 namespace PMacc
 {
     namespace nvidia
```
PrometheusPi: Following the [coding guideline](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/Coding-Guide-Lines#1-spaces-and-indentation) there should be no indentation here (and all following namespaces).

PrometheusPi:
@ax3l Your name is missing in the license header.

ax3l:
no it is not, I just reintroduced a fix I merged accidentally.

also, it's do not consider such changes to be above the required "threshold of originality" for future reference.

PrometheusPi:
@ax3l That's fine withe me as long as `git blame` still works :sunglasses:

src/libPMacc/include/nvidia/functors/Assign.hpp
ax3l: yes, but this PR is, as stated in the description, a bug fix and tries to fix exactly *one* thing.

all files inside `src/libPMacc/include/nvidia/functors` have wrong formatting, we are not picking things we come along in our PRs but we change focused and topic-based.

PrometheusPi: I agree - we should keep the keep the wrong indentation in mind (I will fix them, as soon as this pull request is merged).

ax3l: that's not really necessary but if you do so, pls keep #755 in mind

PR #1456, opened 2016-05-04
Resource monitor class

erikzenker:
This PR was split from PR #1403.

The resource monitor is located in libPMacc and provides information about allocated resources of the library. Currently, particle count and cell count per rank are provided.

PR #1457, opened 2016-05-04
Resource log plugin

erikzenker:
This PR is a split of #1403 and needs to be merged after #1456 
- [x] Merge after #1456 

The resource log is a user adaptable resource information log plugin which collects resource information and writes these information to a stream.

The user can choose which resource properties to write e.g.:
`--resourceLog.properties rank position currentStep particleCount cellCount`

which format to use e.g.:
`--resourceLog.format json` other options are `jsonpp`, `xml` and `xmlpp`

which stream the information should be written e.g.:
`--resourceLog.stream file` other options are `stdout` and `stderr`

and which prefix to use for filestreams:
`--resourceLog.prefix myprefix_`

Using the options `--resourceLog.period 1 --resourceLog.stream stdout --resourceLog.properties rank position currentStep particleCount cellCount --resourceLog.format jsonpp` will write resource objects to stdout such as:

```
[1,1]<stdout>:    "resourceLog": {
[1,1]<stdout>:        "rank": "1",
[1,1]<stdout>:        "position": {
[1,1]<stdout>:            "x": "0",
[1,1]<stdout>:            "y": "1",
[1,1]<stdout>:            "z": "0"
[1,1]<stdout>:        },
[1,1]<stdout>:        "currentStep": "357",
[1,1]<stdout>:        "cellCount": "1048576",
[1,1]<stdout>:        "particleCount": "2180978"
[1,1]<stdout>:    }
[1,1]<stdout>:}
```

For each format there exists always a non pretty print version to simplify further processing:

```
[1,3]<stdout>:{"resourceLog":{"rank":"3","position":{"x":"1","y":"1","z":"0"},"currentStep":"415","cellCount":"1048576","particleCount":"2322324"}}
```

### DIFF
--- a/src/picongpu/include/plugins/ResourceLog.hpp
+++ b/src/picongpu/include/plugins/ResourceLog.hpp
@@ -1,0 +1,243 @@
+/**
+ * Copyright 2016 Erik Zenker
+ *
+ * This file is part of libPMacc.
+ *
+ * libPMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * libPMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with libPMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+// PMacc
+#include "Environment.hpp"
+#include "mappings/simulation/ResourceMonitor.hpp"
+
+// PIConGPU
+#include "plugins/ILightweightPlugin.hpp"
+#include "ILightweightPlugin.hpp"
+#include "simulation_defines.hpp"
+
+// Boost
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/xml_parser.hpp>
+#include <boost/filesystem.hpp>
+
+// STL
+#include <iostream>  /* std::cout, std::ostream */
+#include <numeric>   /* std::accumulate */
+#include <string>    /* std::string */
+#include <sstream>   /* std::stringstream */
+#include <fstream>   /* std::filebuf */
+#include <map>       /* std::map */
+
+// C LIB
+#include <stdlib.h> /* itoa */
+#include <stdint.h> /* uint32_t */
+
+namespace picongpu
+{
+    using namespace PMacc;
+
+
+    class ResourceLog : public ILightweightPlugin
+    {
+    private:
+        MappingDesc *cellDescription;
+        ResourceMonitor<simDim> resourceMonitor;
+
+        // programm options
+        std::string outputFilePrefix;
+        std::string streamType;
+        std::string outputFormat;
+        std::vector<std::string> properties;
+
+        std::filebuf fileBuf;
+        std::map<std::string, bool> propertyMap;
+
+    public:
+
+        ResourceLog() :
+                cellDescription(NULL)
+        {
+            Environment<>::get().PluginConnector().registerPlugin(this);
+        }
+
+        std::string pluginGetName() const
+        {
+            return "ResourceLog";
+        }
+
+        void notify(uint32_t currentStep)
+        {
+            //
+            // Create property tree which contains the resource information
+            using boost::property_tree::ptree;
+            ptree pt;
+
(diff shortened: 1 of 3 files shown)


erikzenker:
**PrometheusPi:**Cool! Would it be possible to add this kind of information to the hdf5 output as well (or to write it in a separate hdf5 file? Would it make sense to have this information available with the hdf5 output?

How much resources does this output need?
I personally believe it would be great to have this output available (by default) in the hdf5 output. We often encounter user/physicist problems where a simulation crashed. One usually has to calculate or estimate the resource usage afterwards. With this plugin (as default output) looking at the last hdf5 dump would tremendously help to find these problems.

**erikzenker:**The resource usage depends on the properties the user embeds into the stream. Properties which are excluded from the stream will not be retrieved.

By "in the hdf5 output" you mean the output of the hdf5 plugin? Well the hdf5 plugin could retrieve resource information from the resource monitor class and embed this information into the hdf5 files, but I would not prefer this approach. Better, the resourceLog plugin gather the hdf5 file descriptor and writes resource information directly into the hdf5 file. I am not sure if this works, I have to check it first.

Maybe this plugin could also prevent the system to crash by notifying other components when e.g.: memory is low etc.

**ax3l:**regarding HDF5: just write an independent HDF5 file if necessary. but for now the implemented formats are quite enough, I would rely on auto-serialization of the final (ressource) object to other outputs to keep the maintenance costs low (or do it in post-processing).

erikzenker:
**ax3l:**great to see you already have the ressource monitor going on! :)

~~You might want to split the new libPMacc RessourceMonitor and the PIConGPU plugin in two PRs.~~
One idea for the plugin: one could also open a high port on each rank (or master) to query it in a style munin nodes (clients) can be accessed (via simple telnet commands or ~~alternatively (web)sockets~~ (argh, won't work), etc.)

erikzenker:
**ax3l:**please add an example to `doc/TBG_macros.cfg` and document the plugin in our [wiki](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-Plugins) :)

PrometheusPi:
@erikzenker Your name is missing in the license header of `PluginController.hpp`.

erikzenker:
@ax3l I added a TBC_macro description and an entry to the [wiki](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-Plugins)

erikzenker:
@PrometheusPi I added my name to the `PluginController.hpp` authors

psychocoderHPC:
You can rebase this pull request

src/picongpu/include/plugins/ResourceLog.hpp
```diff
@@ -0,0 +1,243 @@
...
+                    ss << outputFilePrefix << rank;
+                    boost::filesystem::path resourceLogPath(ss.str());
+                    fileBuf.open(resourceLogPath.string().c_str(), std::ios::out);
+                }
+            }
+        }
+
+        void pluginUnload()
+        {
+            if(fileBuf.is_open()){
```
psychocoderHPC: Is it possible to close the file after each write and reopen and append data? 
This would allow that we can add restart capability to this plugin.

psychocoderHPC: ping

psychocoderHPC:
Could you please rebase this branch and check the last open question

ax3l:
@erikzenker can you pls rebase? :)

src/picongpu/include/plugins/ResourceLog.hpp
ax3l: actually a flush should be sufficient.

ax3l: no, actually `std::ostream os` is desctructed (& closed) after write which should be sufficiently trigger a flush (anyway, that's never super-guaranteed on a HPC file system).

erikzenker:
@ax3l I force pushed your rebased rebase

ax3l:
regression: interesting, boost property tree fails for me on hypnos with boost 1.62
```
  1) gcc/4.9.2                     6) pngwriter/0.5.6
  2) cmake/3.7.2                   7) hdf5-parallel/1.8.15
  3) boost/1.62.0                  8) libsplash/1.6.0
  4) cuda/8.0                      9) adios/1.10.0
  5) openmpi/1.8.6.kepler.cuda80
```

via
```
/opt/pkg/devel/boost/1.62.0/gnu/4.9.2/64/opt/include/boost/property_tree/json_parser/detail/parser.hpp(217): error: identifier "_1" is undefined

/opt/pkg/devel/boost/1.62.0/gnu/4.9.2/64/opt/include/boost/property_tree/json_parser/detail/parser.hpp(520): error: identifier "_1" is undefined
```

I think I fixed a similar problem in boost before...

ax3l:
Update: yes, yes [I did](https://svn.boost.org/trac/boost/ticket/10932)

ax3l:
related to #1848 and https://github.com/boostorg/property_tree/pull/26

ax3l:
linked the new plugin to your wiki page in the [plugin overview page](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-Plugins). Awesome, thanks again for creating the page! :heart:

PR #1461, opened 2016-05-11
Use correct sqrt function for abs(Vector)

Flamefire:
Currently `sqrtf` was used, which is the floating point function. This is crucially wrong for double vectors, especially as taking the square root already looses significant digits.

This uses the type generic version from `PMacc::algorithms::math`


PrometheusPi:
@Flamefire great catch

PrometheusPi:
looks good to me - I will merge as soon as all checks are finished

ax3l:
thanks! :)
update the flags accordingly.

PR #1462, opened 2016-05-11
replace sqrtf with algorithms::math::sqrt

PrometheusPi:
This is a pull request replaces `sqrtf` by `algorithms::math::sqrt` in the picongpu main code.
This is similar to #1461 where @Flamefire replaced `sqrtf` in libPMacc.

### DIFF
--- a/src/picongpu/include/fields/laserProfiles/laserGaussianBeam.hpp
+++ b/src/picongpu/include/fields/laserProfiles/laserGaussianBeam.hpp
@@ -129,7 +129,7 @@ namespace picongpu
             return elong;
 #else
             //beam waist in the near field: w_y(y=0) == W0
-            const float_X w_y = W0 * sqrtf( float_X(1.0) + ( FOCUS_POS / y_R )*( FOCUS_POS / y_R ) );
+            const float_X w_y = W0 * algorithms::math::sqrt( float_X(1.0) + ( FOCUS_POS / y_R )*( FOCUS_POS / y_R ) );
             //! the Gouy phase shift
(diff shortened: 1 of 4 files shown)


Flamefire:
I'd suggest to instantiate the template with `sqrt_x` which was introduced to allow double precision square roots taken from single precision values

src/picongpu/include/fields/laserProfiles/laserGaussianBeam.hpp
```diff
@@ -129,7 +129,7 @@ namespace picongpu
             return elong;
 #else
             //beam waist in the near field: w_y(y=0) == W0
-            const float_X w_y = W0 * sqrtf( float_X(1.0) + ( FOCUS_POS / y_R )*( FOCUS_POS / y_R ) );
+            const float_X w_y = W0 * algorithms::math::sqrt( float_X(1.0) + ( FOCUS_POS / y_R )*( FOCUS_POS / y_R ) );
             //! the Gouy phase shift
             const float_X xi_y = atanf( -FOCUS_POS / y_R );
```
psychocoderHPC: Could you please use `math::atan` here

PrometheusPi: @psychocoderHPC Should I know add this in another pull request (as discussed *offline*  3 days ago) or should I add this to this pull request?

psychocoderHPC: Please add a new pull request

PR #1465, opened 2016-05-12
PNG: avoid double scaling

### LINKED ISSUE #1464: PNG: Avoid Double-Scaling
Avoid scaling *twice* with non-cubic cells.

By scaling the smaller direction up first we increase the already large png memory. We could instead scale directly with the final scaling factor per direction.

ccing @psychocoderHPC

psychocoderHPC:
close #1464

- avoid that PNGs are scaled two times if `scale_image` is not one and the cell size in PIConGPU is not quadratic
- remove member method `resizeAndScaleImage()`

### DIFF
--- a/src/picongpu/include/plugins/output/images/PngCreator.hpp
+++ b/src/picongpu/include/plugins/output/images/PngCreator.hpp
@@ -124,12 +124,6 @@ namespace picongpu
                         const Size2D size,
                         const MessageHeader header);
 
-        void resizeAndScaleImage(pngwriter* png, double scaleFactor)
-        {
-            if (scaleFactor != 1.)
-                png->scale_k(scaleFactor);
-        }
-
         std::string m_name;
         std::string m_folder;
         bool m_createFolder;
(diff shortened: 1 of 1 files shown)


src/picongpu/include/plugins/output/images/PngCreator.hpp
```diff
@@ -157,8 +151,8 @@ namespace picongpu
 
         std::stringstream step;
         step << std::setw(6) << std::setfill('0') << header.sim.step;
-        float_X scale_x = header.sim.scale[0];
-        float_X scale_y = header.sim.scale[1];
+        float_X scale_x = header.sim.scale[0] * scale_image;
```
ax3l: better write:
```C++
float_X scale_x = scale_image;
float_X scale_y = scale_image;
if( scale_to_cellsize )
{
    scale_x *= header.sim.scale[0];
    scale_y *= header.sim.scale[1];
}
```

src/picongpu/include/plugins/output/images/PngCreator.hpp
```diff
@@ -186,9 +180,6 @@ namespace picongpu
             if ((scale_x != float_X(1.0)) || (scale_y != float_X(1.0)))
```
ax3l: careful: the if in the line above should now be removed.

ax3l:
@psychocoderHPC small bug, comments inline.

src/picongpu/include/plugins/output/images/PngCreator.hpp
psychocoderHPC: You are right I need to handle the case with the scaling to the cell size and without

psychocoderHPC:
updated

ax3l:
now it's wonderful :)

PR #1466, opened 2016-05-13
Fix 2 includes and a duplicate define

Flamefire:
Titel says it all. Just discovered that.
Did not cause any problems because files were not used.

### DIFF
--- a/src/libPMacc/include/pmacc_types.hpp
+++ b/src/libPMacc/include/pmacc_types.hpp
@@ -58,8 +58,6 @@ typedef uint64_t id_t;
 typedef unsigned long long int uint64_cu;
 typedef long long int int64_cu;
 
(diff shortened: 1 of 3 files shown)


src/libPMacc/include/pmacc_types.hpp
```diff
@@ -58,8 +58,6 @@ typedef uint64_t id_t;
 typedef unsigned long long int uint64_cu;
 typedef long long int int64_cu;
 
-#define BOOST_MPL_LIMIT_VECTOR_SIZE 20
```
psychocoderHPC: are you sure that this can be removed?

Flamefire: Yes: https://github.com/Flamefire/picongpu/blob/4b4896ab2b19f97985dddfc84d48e92853eec6ea/src/libPMacc/include/pmacc_types.hpp#L29

PR #1467, opened 2016-05-13
Fix C&P mistake: Use Host DataBodx

Flamefire:
Nothing to add


psychocoderHPC:
This bug effects the latest master but the DataBox for `HostBuffer` and `DeviceBuffer` are equal.

ax3l:
how did this work in the first place? `DataBoxType` is actually used...

Flamefire:
Both are (currently) the same type I think. Or at least convertible.

PR #1470, opened 2016-05-20
fix usage of wrong `abs()` function

psychocoderHPC:
- on device use `::fabs()` for double values
- on host use `std:abs()` for double values
- use `::fabsf()` for float values

[Documenation `fabsf`](http://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__SINGLE.html)
[Documenation `fabs`](http://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__DOUBLE.html)

I can't find a documentation for `::abs()` in the cuda header files. The wrong usage for float values should only result in an implicit cast and a slow abs call (maybe optimized by the compiler beauce we use fast_math)

This bug effects the latest release but we saw no issues in the past.

- [x] compare runtime dev/pull - runtime changes
- [x] test charge conservation

### DIFF
--- a/src/libPMacc/include/algorithms/math/doubleMath/abs.tpp
+++ b/src/libPMacc/include/algorithms/math/doubleMath/abs.tpp
@@ -41,7 +41,14 @@ struct Abs<double>
 
     HDINLINE double operator( )(double value)
     {
-        return ::abs( value );
+#ifdef __CUDA_ARCH__
+      return ::fabs( value );
+#else
+      /* \bug on cpu `::abs(double)` always return zero -> maybe this is the
+	* integer version of `abs()`
+	*/
+      return std::abs( value );
+#endif
     }
 };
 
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
heat_abs

Compare charge conservation: 
 - 1. Pull reqest
 - 2. Dev branch

ax3l:
huh, how dangerous for the `double` math!

Did you do the heating test for `double` for `float`? I guess the influence for `double` sims would be higher if it is used there.

psychocoderHPC:
Th test above is single precision. I did heating tests in the past which all look good. 
I am don't know which function is used in cude if we call `::abs()` with a double.

psychocoderHPC:
heat_abs_double

This is the charge conservation for **double** with electrons/positrons.

PR #1471, opened 2016-05-20
add floor specialization for PMacc vectors

psychocoderHPC:
add floor specialization for `PMacc::math::Vector<>`


ax3l:
@psychocoderHPC can you somehow rebase? The "Checking for ability to merge automatically…" seems to hang on GitHub.

PR #1472, opened 2016-05-20
Topic remove vector math implementations

### LINKED ISSUE #1338: libPMacc using algorithm::math instead of ::math
While rewriting the particle to field interpolation for the pusher I found the following strange behavior.
The code is using

``` c++
{
using namespace PMacc;
...
}
```

thus `math::floor(...)` should reference to `PMacc::math::floor`. But the compiler searches for `floor` in `PMacc::algorithms::math`, where there is no `floor`.

Using  `PMacc::math::floor(...)` **solves** the problem. But why is the compiler looking in the namespace `algorithms`?

There is no `using namespace *algorithms` anywhere in libPMacc or PIConGPU. 
This might relate to #846.

psychocoderHPC:
- remove math function implementations in Vector.hpp ￼…
  - remove `dot()`
  - remove `abs()`
  - remove `abs2()`
  - remove `floor()`
-  fix usage of deleted math functions

This pull requests close #1338

Todo:
- [x] need rebase against #1471


psychocoderHPC:
rebased

ax3l:
ty!

PR #1475, opened 2016-05-24
fix rng uniform float results

psychocoderHPC:
- add range definition policies
- specialize the generation of uniform floats based on the policy
- update documentation

Tests:
- [x] run tests (use #1494)

CC-ing: @Heikman, @Flamefire

### DIFF
--- a/src/libPMacc/include/random/distributions/uniform/Uniform_float.hpp
+++ b/src/libPMacc/include/random/distributions/uniform/Uniform_float.hpp
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2016 Alexander Grund
+ * Copyright 2015-2016 Alexander Grund, Rene Widera
  *
  * This file is part of libPMacc.
  *
@@ -24,6 +24,7 @@
 
 #include "pmacc_types.hpp"
 #include "random/distributions/Uniform.hpp"
+#include "random/distributions/uniform/Range.hpp"
 
 namespace PMacc
 {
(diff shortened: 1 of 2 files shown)


Flamefire:
Good catch. But what about generating the rng value (`raw` here) directly in [0,1)? The function called does basically: `uintValue * 2^-32 + 2^-33`
As the `Get32Bits` is already defined to return a random 32 bit value we can skip the nvidia call here completely and calculate simply `uintValue * 2^-32` which will get us the desired range. As random needs to be as fast as possible I'd rather favor this solution instead of that adjustment after calculating a wrong value

psychocoderHPC:
@Flamefire Please see #1476. Not all rng methods use `uintValue * 2^-32 + 2^-33` but looks like a valid way for Xor implementation of curand.

Flamefire:
I added a comment there. It is still possible, to get an `uint32_t` making the mentioned calculation here possible with all methods. Hence I'd favor this one.

Flamefire:
I just tried `RNGMethod().get32Bits(state) * CURAND_2POW32_INV` and still got `1.0` as a result. This is probably due to the fact, that the highest possible value is `1 - 2^-32` ~= `1 - 2.3e-10` but the next lowest value than 1 is ~ `1 - 1.2e-7`
Simply swapping 1->0 might produce to many 0s

psychocoderHPC:
But this default range is `(0,1]` and than we swap one to zero. This should not produce to much zeros.

Flamefire:
Depending on your definition of "to many". For a uniform distribution in [0,1) one should expect probabilities of `p(0<=x<0.5) == p(0.5<=x<1) == 50%`. I'm pretty sure it is possible to show, that the swapping will not ensure this, as "to many" values are moved. According to the above numbers ~1000 values that result in 1 will be 0. Not completely sure about that.

Flamefire:
Just discussed this with @PrometheusPi: It will change the statistic.
The solution is simply: For values outside the range, re-run the RNG. Thus:

    do{
        float result = rng() / 2^32;
    }while(result == 1);

ax3l:
First of all it is written "too" many.

If zero-es are guaranteed to be excluded in one, then swapping 1 which is already uniform to it should not change the statistics.

re-running the rng is messy and jumps depending on runtime parameters.

PrometheusPi:
@ax3l I agree, **but** the jumps occur once in 2^24 ~ 1.7e7 cases. I would consider this acceptable.

Flamefire:
Note also that this is not about subnormal numbers. Subnormals for float start around 2^-126 but the smallest value generated is around 2^-32. The problem is that the spacing between adjacent floats is not linear. But you try a linear mapping from [0, 2^23) to (0,1] (CUDA).
I don't see why the 3-liner posted is more messy than the 2-liner in the original "fix". And that one also contained a jump (although it is cheaper)

psychocoderHPC:
I removed the swap from one to zero by calculation the floating point number out of the `uint32_t` value with a equation from the paper [Conversion of high-period random numbers to floating point; DOI: 10.1145/1189756.1189759](http://dl.acm.org/citation.cfm?doid=1189756.1189759)

Flamefire:
I just made some tests. I simply generate random numbers and check if they are `<0.5` or not.
Results:

>>              <0.5          >=0.5           Time
1->0  2621478885 2621401115 43662.5ms
while 2621478803 2621401197 44146.4ms

Other seed;
               <0.5          >=0.5           Time
1->0  2621386295 2621493705 43676.6ms
while 2621386222 2621493778 44148.8ms

Seems it doesn't really matter...

Flamefire:
About the update: This does not work! It is just a variation from `x/2^32` and can still produce 1 for the same reason. I don't even think the 2 algorithms are different (`x/2^32` <-> `int(x)/2^32+0.5`) as the calculation is a FMA and the result correctly rounded to 1 ULP. Hence we get the same value range. Complicated however...

psychocoderHPC:
@Flamefire: yes you are right there are still `1.0` in :-(

ax3l:
@psychocoderHPC @Flamefire status of this PR?

psychocoderHPC:
Need some more time. I will add a policy were the user can choose the behavior. 
 - rerun the rng method if result is 1.0 (the only way for uniform numbers :-(
 - swap 1.0 to 0.0 (results in non uniform random numbers)

psychocoderHPC:
I updated this branche. 
I added the possibility to define in which range the random numbers should be.
The interface changes are backwards compatible. 
The default case uses a reduce the random numbers to 24 bit and create numbers in the same range like the original PMacc implementation [0,1).
The reason to set the the 24bit rng as default are the following:
  - `uniform::ExcludeOne<float>::SwapOneToZero` create small errors in the uniform distribution
  - `uniform::ExcludeZero<float>` is not equal to the old PMacc random number generator
  - `uniform::ExcludeOne<float>::Repeat` has a non constant runtime

Flamefire:
Note: This is backwards **compatible** but not backwards **equivalent** as the default (using 24Bit) results in less possible values in the range of `[0,0.5)` than there were before.

src/libPMacc/include/random/distributions/uniform/Uniform_float.hpp
```diff
@@ -48,10 +54,125 @@ namespace detail
         DINLINE result_type
         operator()(StateType& state)
         {
-            return _curand_uniform(RNGMethod().get32Bits(state));
+            const result_type 2pow32Inv = 2.3283064e-10f;
```
Flamefire: I think we should use `float` here instead of result_type as the constants are `float`

src/libPMacc/include/random/distributions/uniform/Uniform_float.hpp
```diff
@@ -34,11 +35,16 @@ namespace distributions
 namespace detail
 {
 
-    /**
-     * Returns a random float value uniformly distributed in [0,1)
+    /** Returns a random float value uniformly distributed in (0,1]
+     *
+     * The smallest created value is ~ 1.164*10^-10
```
Flamefire: Be more specific: `2^-33`

src/libPMacc/include/random/distributions/uniform/Uniform_float.hpp
```diff
@@ -48,10 +54,125 @@ namespace detail
...
+                    uniform::ExcludeZero<result_type>
+            >()(state);
+            return randomValue == 1.0f ? 0.0f : randomValue;
+        }
+    };
+
+    /** Returns a random float value uniformly distributed in [0,1)
+     *
+     * Reduce the random range to 2^24.
+     * Creates intervals with the width of epsilon/2.
```
Flamefire: `2^-24` to be exact.

src/libPMacc/include/random/distributions/uniform/Uniform_float.hpp
```diff
@@ -48,10 +54,125 @@ namespace detail
...
+                if( randomValue != 1.0f )
+                    return randomValue;
+            }
+            while(true);
+        }
+    };
+
+    /** Returns a random float value uniformly distributed in [0,1)
+     *
+     * The default case use the method
```
Flamefire: Just write here: `Equivalent to uniform::ExcludeOne<float>::Use24Bit` or better (IMO): Unspecified method is used for corner cases because this is the default for people only interested in getting a uniform float in `[0,1)` not caring how. So leave us the freedom to choose an implementation here.

src/libPMacc/include/random/distributions/uniform/Uniform_float.hpp
```diff
@@ -48,10 +54,125 @@ namespace detail
...
+    class Uniform<float, T_RNGMethod, void> :
+        public PMacc::random::distributions::Uniform<
+            uniform::ExcludeOne<float>::Use24Bit
+        >
+    {
+    };
+
+    /** Returns a random float value uniformly distributed in [0,1)
+     *
+     * The default case  for `ExcludeOne` use the method
```
Flamefire: Here should be only `Equivalent to uniform::ExcludeOne<float>::Use24Bit` or similar to above (freedom to choose implementation).`Default case` is meaningless here. This should go to the `Uniform` template

src/libPMacc/include/random/distributions/uniform/Uniform_float.hpp
PrometheusPi: no that's fine - this reduces the random range of the uniform integer random number to a random value in [0, 2^24-1].

PrometheusPi: Either the binning is `2^-24` **or** there are `2^24` random numbers possible. The **range** is `[0,1)`.

Flamefire: Maybe better write: Uses a uniform distance of 2^-24 between each possible random number

psychocoderHPC:
I updated this version
- fix broken compile 
- run test

src/libPMacc/include/random/distributions/uniform/Uniform_float.hpp
```diff
@@ -34,24 +35,148 @@ namespace distributions
...
+                const float randomValue =
+                    PMacc::random::distributions::Uniform<
+                        uniform::ExcludeZero<result_type>,
+                        RNGMethod
+                    >()(state);
+
+                if( randomValue != 1.0f )
+                    return randomValue;
+            }
+            while(true);
```
Flamefire: I'd favor `while(randomValue == 1.0f); return randomValue;` as it makes the intent clear. This looks a bit hackish to me.
Generated PTX code is the same however (CUDA 7.5)

psychocoderHPC: IMO the current version is better because we can use a const variable (this should be give the compiler more freedom for optimization)

ax3l: like the scoped const float, too :)

ax3l:
thank you two! :sparkles:

PR #1478, opened 2016-05-24
openPMD (HDF5): `axisLabels` Order Corrected

ax3l:
Corrected the order of `axisLabels` to be correctly `z;y;x` as in `F[z][y][x]`.
(Or in 2D: `y;x` as in `F[y][x]`)

Also moves the allocation of the static size c-array of strings from heap to the stack.


PrometheusPi:
@ax3l looks good to me.

PR #1485, opened 2016-05-26
Cuda event fix to avoid deadlock

Flamefire:
- [x] Rebase on #1484 

There is a potential deadlock, when the eventsystem waits for the completion of an event (e.g. after a kernel) but the kernel errors (e.g. invalid memory access, timeout, ...) As in this case `cudaEventQuery` will always return an errorcode the eventsystem will never receive a `cudaSuccess`/`CudaEvent::isFinished` and loops forever.


psychocoderHPC:
ready for rebase

Flamefire:
Rebased

psychocoderHPC:
well spotted!

PR #1486, opened 2016-05-27
fix particle out of mem error/warning

psychocoderHPC:
fix inverted warning and error message

This bug effects the latest release **but there is no need for a back port**. Only the error message was wrong.


PrometheusPi:
Looks good to me. Will be merges as soon as the compile tests are done.

PR #1487, opened 2016-05-27
Add MRG32k3a RNG generator

Flamefire:
This adds another random number generator that can be used to compare those against each other.

Note that this are the only 2 implemented in CUDA that can be used easily. The others require multiple threads calling them in sync with the same state which is very hard in the current design of picongpu/PMacc.

Also added the corresponding Min state generator that cannot directly be used for normal distributions but uses less memory and is much faster (~20% in my local test)
Update: Just tested this on the K20: I see speedups of about 300% when using the Min state. Talking numbers:

- Gridsize=256x256
- Samples per Cell=100.000
- Time CUDA state: 24.4s
- Time Min state: 8.1s

### DIFF
--- a/src/libPMacc/include/random/methods/MRG32k3a.hpp
+++ b/src/libPMacc/include/random/methods/MRG32k3a.hpp
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2016 Alexander Grund
+ * Copyright 2016 Alexander Grund
  *
  * This file is part of libPMacc.
  *
--- a/src/libPMacc/include/random/methods/MRG32k3a.hpp
+++ b/src/libPMacc/include/random/methods/MRG32k3a.hpp
@@ -1,0 +1,62 @@
+/**
+ * Copyright 2015-2016 Alexander Grund
+ *
+ * This file is part of libPMacc.
+ *
+ * libPMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * libPMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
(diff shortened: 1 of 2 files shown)


src/libPMacc/include/random/methods/MRG32k3aMin.hpp
```diff
@@ -0,0 +1,94 @@
+/**
+ * Copyright 2015-2016 Alexander Grund
```
PrometheusPi: please adjust to `2016` only

src/libPMacc/include/random/methods/MRG32k3a.hpp
```diff
@@ -0,0 +1,62 @@
+/**
+ * Copyright 2015-2016 Alexander Grund
```
PrometheusPi: please adjust to `2016` only

PrometheusPi:
@Flamefire Looks good to me! (can be merged)

Please state why one can skip the box muller variables in the `state`.

Flamefire:
Already done: 

>Also added the corresponding Min state generator that cannot directly be used for normal distributions but uses less memory and is much faster (~20% in my local test)

Those variables are solely used for normal distributions and not required for others (uniform,...)

PrometheusPi:
@Flamefire thank you.

src/libPMacc/include/random/methods/MRG32k3aMin.hpp
```diff
@@ -0,0 +1,94 @@
...
+            PMACC_STATIC_ASSERT_MSG(
+                    offsetof(StateType, s1) == offsetof(curandStateMRG32k3a, s1) &&
+                    offsetof(StateType, s2) == offsetof(curandStateMRG32k3a, s2),
+                    Incompatible_structs);
+        }
+        static HDINLINE void
+        AssignState(StateType& dest, const curandStateMRG32k3a& src)
+        {
+            // Check if we can do this cast
+            CheckStateCompatibility();
```
ax3l: due to the static assert comparisons inside those, which are compile time, it is a bit odd to "call" this instead of putting the static asserts simply in the `private` section of the class.

Flamefire: Correct. There were some argument dependetn checks before, but not anymore. Anything else?

ax3l: I think that's it.
I did not dig deep into the MRG32k3a interfaces and reasoning behind this for a double-check but trust your judgement from your explanations.

Issue #1419, opened 2016-03-30
charge conservation plugin creates files if disabled


psychocoderHPC:
The charge conservation plugin creates files and may be consume memory even if the plugin is not enabled.

https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/plugins/ChargeConservation.tpp#L66-L76

PR #1488, opened 2016-06-01
fix charge conservation plugin

Heikman:
Creates the output file only if the plugin is enabled.

Solves #1419.

### DIFF
--- a/src/picongpu/include/plugins/ChargeConservation.tpp
+++ b/src/picongpu/include/plugins/ChargeConservation.tpp
@@ -61,6 +61,9 @@ std::string ChargeConservation::pluginGetName() const {return this->name;}
 
 void ChargeConservation::pluginLoad()
 {
+    if(this->notifyPeriod == 0u)
+        return;
+
     Environment<>::get().PluginConnector().setNotificationPeriod(this, this->notifyPeriod);
 
     PMacc::GridController<simDim>& con = PMacc::Environment<simDim>::get().GridController();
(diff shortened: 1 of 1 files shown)


src/picongpu/include/plugins/ChargeConservation.tpp
```diff
@@ -69,7 +69,7 @@ void ChargeConservation::pluginLoad()
     zone::SphericZone<simDim> zone_allGPUs(gpuDim);
     this->allGPU_reduce = AllGPU_reduce(new PMacc::algorithm::mpi::Reduce<simDim>(zone_allGPUs));
 
-    if(this->allGPU_reduce->root())
+    if(this->allGPU_reduce->root() && this->notifyPeriod > 0u)
```
psychocoderHPC: please move this to the top of the method to avoid any action of this plugin if it is not needed. (e.g. set `setNotificationPeriod()`)

ax3l:
looks good, thank you!

src/picongpu/include/plugins/ChargeConservation.tpp
```diff
@@ -61,6 +61,9 @@ std::string ChargeConservation::pluginGetName() const {return this->name;}
 
 void ChargeConservation::pluginLoad()
 {
+    if(this->notifyPeriod == 0u)
+        return;
+
     Environment<>::get().PluginConnector().setNotificationPeriod(this, this->notifyPeriod);
```
ax3l: @psychocoderHPC not sure if we should still call this with `zero` period.

psychocoderHPC: It is not called because the method exit if period is zero.

ax3l: ah sure, could have checked that. all right.

PR #1492, opened 2016-06-08
Add missing const for getter in GridLayout

Flamefire:
Allows usage of  `const GridLayout&` parameters


PrometheusPi:
@Flamefire please add your name to the license header

Flamefire:
I did not do this on purpose, as this fix is so minor, just like fixing a typo or similar. So I wouldn't claim any (Co)authorship here except the committer.

PrometheusPi:
OK, accepted.

PR #1494, opened 2016-06-10
Add tests for all RNG generators

Flamefire:
This adds simple tests for the RNG generators with which compile tests and manual verification can be done.

Runtimetests can be done with `make run`. Manual execution with `mpiexec -n 1 TestRng <numNumbersPerCell>`

Currently only uniform float distribution is used, but others can be used as well. Good fit for tests by @psychocoderHPC 

Later also automatic verification via Boost.Test shall be added. 

Requires MPI execution with 1 GPU and should run reasonably fast (<10 min)

### DIFF
--- a/src/libPMacc/test/random/2DDistribution.cu
+++ b/src/libPMacc/test/random/2DDistribution.cu
@@ -32,9 +32,10 @@
 #include <stdint.h>
 #include <iostream>
 #include <fstream>
+#include <limits>
 
-typedef PMacc::DataSpace<2> Space2D;
-typedef PMacc::DataSpace<3> Space3D;
+typedef PMacc::DataSpace<DIM2> Space2D;
+typedef PMacc::DataSpace<DIM3> Space3D;
 
 template<class T_DataBox, class T_Random>
 __global__ void RandomFiller(T_DataBox box, Space2D boxSize, T_Random rand, uint32_t numSamples)
@@ -73,6 +74,7 @@ private:
     PMACC_ALIGN8(rand, Random);
 };
 
+/** Write in PGM grayscale file format (easy to read/interpret) */
 template<class T_Buffer>
 void writePGM(const std::string& filePath, T_Buffer& buffer)
 {
@@ -87,20 +89,30 @@ void writePGM(const std::string& filePath, T_Buffer& buffer)
                 maxVal = val;
         }
     }
-    if(maxVal > 0xFFFF)
-        maxVal = 0xFFFF;
+    
+    // Standard format is single byte per value which limits the range to 0-255
+    // An extension allows 2 bytes so 0-65536)
+    if(maxVal > std::numeric_limits<uint16_t>::max())
+        maxVal = std::numeric_limits<uint16_t>::max();
+    const bool isTwoByteFormat = maxVal > std::numeric_limits<uint8_t>::max();
+    
     std::ofstream outFile(filePath.c_str());
+    // TAG
     outFile << "P5\n";
+    // Size and maximum value (at most 65536 which is 2 bytes per value)
     outFile << size.x() << " " << size.y() << " " << maxVal << "\n";
     for(int y=0; y<size.y(); y++)
     {
         for(int x=0; x<size.x(); x++)
         {
             uint32_t val = buffer.getDataBox()(Space2D(x, y));
+            // Clip value
             if(val > maxVal)
                 val = maxVal;
-            if(maxVal > 0xFF)
+            // Write first byte (higher order bits) if file is in 2 byte format
+            if(isTwoByteFormat)
                 outFile << uint8_t(val >> 8);
+            // Write remaining bytze
             outFile << uint8_t(val);
         }
     }
--- a/src/libPMacc/test/random/2DDistribution.cu
+++ b/src/libPMacc/test/random/2DDistribution.cu
@@ -1,0 +1,223 @@
+/**
+ * Copyright 2016 Alexander Grund
+ *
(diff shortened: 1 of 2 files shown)


src/libPMacc/test/random/2DDistribution.cu
```diff
@@ -0,0 +1,223 @@
...
+    init(Space2D globalCellIdx)
+    {
+        rand.init(globalCellIdx);
+    }
+
+    DINLINE Space2D
+    operator()(Space2D size)
+    {
+        using PMacc::algorithms::math::float2int_rd;
+        return Space2D(float2int_rd(rand() * size.x()), float2int_rd(rand() * size.y()));
```
psychocoderHPC: this needs a modulo ` return Space2D(float2int_rd(rand() * size.x()) % size.x(), float2int_rd(rand() * size.y()) % size.y() );` else the float value `1.0` creates  out of memory access and later on the host assert fails. This is not the best solution but save.

Flamefire: This assumes that rand() is already fixed returning only values in `[0, 1)`
Try to put in the `if(val == 1.f) return 0.f` fix to test this.

src/libPMacc/PMaccConfig.cmake
```diff
@@ -29,7 +29,7 @@
 ###############################################################################
 # PMacc
 ###############################################################################
-cmake_minimum_required(VERSION 3.3)
+cmake_minimum_required(VERSION 3.0)
```
PrometheusPi: Why could you downgrade this requirement?

Please add your name to the author list.

Flamefire: The version used was pretty arbitrary. No features were used that require such a high version and PIConGPU runs also with less. Reason for changing this was that the most common cmakes are not that new (Cluster installations, linux package managers)
No need to add my name because this is a (too) minor change to claim authorship

psychocoderHPC: Please remove this change from the pull request. There is no need to mix the cmake change with the new feature (TNG test)

Flamefire: I wasn't able to test this w/o this on standards config on hypnos...

ax3l:
@Flamefire can you please add the runtime and hardware requirements in the PR description for the test?

Also, pls rebase :)

src/libPMacc/PMaccConfig.cmake
ax3l: we just updated the requirements to `3.3` yesterday, since a necessary [C++11 fix in FindCUDA.cmake](https://github.com/Kitware/CMake/blob/v3.3.0/Modules/FindCUDA.cmake#L1271-L1282) was first added in that version.

Therefore, now keep `3.3` when you rebase.

src/libPMacc/test/random/2DDistribution.cu
```diff
@@ -0,0 +1,223 @@
...
+#include "random/methods/Xor.hpp"
+#include "random/methods/XorMin.hpp"
+#include "random/methods/MRG32k3a.hpp"
+#include "random/methods/MRG32k3aMin.hpp"
+#include "dimensions/DataSpace.hpp"
+#include <stdint.h>
+#include <iostream>
+#include <fstream>
+
+typedef PMacc::DataSpace<2> Space2D;
```
ax3l: can you pls use our `DIM2` and `DIM3` defines here, since you are talking about *spatial* dimensions?

src/libPMacc/test/random/2DDistribution.cu
```diff
@@ -0,0 +1,223 @@
...
+    for(int y=0; y<size.y(); y++)
+    {
+        for(int x=0; x<size.x(); x++)
+        {
+            uint32_t val = buffer.getDataBox()(Space2D(x, y));
+            if(val > maxVal)
+                maxVal = val;
+        }
+    }
+    if(maxVal > 0xFFFF)
```
ax3l: why obfuscate the `65535` which us the maximum range of a `unsigned short`?

src/libPMacc/test/random/2DDistribution.cu
```diff
@@ -0,0 +1,223 @@
...
+    outFile << "P5\n";
+    outFile << size.x() << " " << size.y() << " " << maxVal << "\n";
+    for(int y=0; y<size.y(); y++)
+    {
+        for(int x=0; x<size.x(); x++)
+        {
+            uint32_t val = buffer.getDataBox()(Space2D(x, y));
+            if(val > maxVal)
+                val = maxVal;
+            if(maxVal > 0xFF)
```
ax3l: what's happening here?

src/libPMacc/test/random/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License and the GNU Lesser General Public License
+# for more details.
+#
+# You should have received a copy of the GNU General Public License
+# and the GNU Lesser General Public License along with libPMacc.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
+cmake_minimum_required(VERSION 3.0)
```
ax3l: go for `3.3` directly, please so it's consistent with the rest.

Flamefire:
Everything added. I'm afraid I can't do more, but should work now.

ax3l:
thank you, that's great!

PR #1495, opened 2016-06-13
Make static_asserts class members

Flamefire:
As suggested by @ax3l in #1487


ax3l:
thank you! :sparkles:

PR #1498, opened 2016-06-15
Move readCheckpointMasterFile to PMacc

Flamefire:
The creation of that file is handled by PMacc so should be the reading.

This also changes the function, so that it returns a vector of checkpoints instead only the last one. This is required, because actually one should only restart from one of those checkpoints even when `--restartStep` is passed. At least one needs to load the last checkpoint before `restartStep`

However the implementation of that change is left to another issue.


psychocoderHPC:
thx

PR #1504, opened 2016-06-21
Add Bremsstrahlung

Heikman:
Implementation of the Bremsstrahlung effect.

Here the screened Bethe-Heitler cross section is used. See e.g.:

> Salvat, F., et al. "Monte Carlo simulation of bremsstrahlung emission by electrons."
>    Radiation Physics and Chemistry 75.10 (2006): 1201-1219.

The numerics separates the energy spectrum into two parts. In the low-energy part
photon emission is neglected and a drag force is applied to the electrons. In the high-energy part
photons are created in addition to the drag force.

Electron deflection is treated as screened Rutherford scattering, see e.g. Jackson, chap. 13.5

The photon emission angle is taken from the Lorentz-boosted dipole radiation formula,
see e.g. Jackson, chap. 15.2

**Tests**

Energy loss without photons. 100MeV electrons should loose 10% of their energy during transmission through a solid copper target.

energy_loss_drag_force

With photons:
- photon threshold is 5 MeV.

energy_loss_photon

Photon spectrum:

photon_spectrum

Photon emission angle with deactivated electron deflection and energy loss:

photon_emission_angle

Electron deflection on single-scattering:
- scattering probability per target transmission is 10 %.
- minimum angle is 0.01 rad. Smaller values can be attributed to multi-scattering.
- energy loss is deactivated

rutherford_single

Electron deflection on multi-scattering:
- electrons are scattered 20 times on average.
- energy loss is deactivated.
- mean squared theta deviation equals 20 times mean squared theta deviation from single-scattering as expected.

rutherford_multi

### DIFF
--- a/src/picongpu/include/particles/ParticlesFunctors.hpp
+++ b/src/picongpu/include/particles/ParticlesFunctors.hpp
@@ -30,12 +30,13 @@
 #include <boost/mpl/plus.hpp>
 #include <boost/mpl/accumulate.hpp>
 
-
 #include "communication/AsyncCommunication.hpp"
 #include "particles/traits/GetIonizer.hpp"
 #include "particles/traits/FilterByFlag.hpp"
 #include "particles/traits/GetPhotonCreator.hpp"
+#include "particles/traits/ResolveAliasFromSpecies.hpp"
 #include "particles/synchrotronPhotons/SynchrotronFunctions.hpp"
+#include "particles/bremsstrahlung/Bremsstrahlung.hpp"
 #include "particles/creation/creation.hpp"
 
 namespace picongpu
@@ -317,6 +318,62 @@ struct CallIonization
     }
 
 }; // struct CallIonization
+
+/** Handles the bremsstrahlung effect for electrons on ions.
+ *
+ * \tparam T_ElectronSpeciesName name of electron species
+ */
+template<typename T_ElectronSpeciesName>
+struct CallBremsstrahlung
+{
+    typedef T_ElectronSpeciesName ElectronSpeciesName;
+    typedef typename ElectronSpeciesName::type ElectronSpecies;
+
+    typedef typename PMacc::particles::traits::ResolveAliasFromSpecies<
+        ElectronSpecies,
+        bremsstrahlungIons<>
+    >::type IonSpecies;
+    typedef typename PMacc::particles::traits::ResolveAliasFromSpecies<
+        ElectronSpecies,
+        bremsstrahlungPhotons<>
+    >::type PhotonSpecies;
+    typedef bremsstrahlung::Bremsstrahlung<IonSpecies, ElectronSpecies, PhotonSpecies> BremsstrahlungFunctor;
+
+    /** Functor implementation
+     *
+     * \tparam T_StorageStuple contains info about the particle species
+     * \tparam T_CellDescription contains the number of blocks and blocksize
+     *                           that is later passed to the kernel
+     * \param tuple an n-tuple containing the type-info of multiple particle species
+     * \param cellDesc points to logical block information like dimension and cell sizes
+     * \param currentStep the current time step
+     */
+    template<typename T_StorageTuple, typename T_CellDescription, typename ScaledSpectrumMap>
+    HINLINE void operator()(
+                            T_StorageTuple& tuple,
+                            T_CellDescription* cellDesc,
+                            const uint32_t currentStep,
+                            const ScaledSpectrumMap& scaledSpectrumMap,
+                            const bremsstrahlung::GetPhotonAngle& photonAngle) const
+    {
+        /* alias for pointer on source species */
+        auto electronSpeciesPtr = tuple[ElectronSpeciesName()];
+        /* alias for pointer on destination species */
+        auto photonSpeciesPtr = tuple[typename MakeIdentifier<PhotonSpecies>::type()];
+
+        const float_X targetZ = GetAtomicNumbers<IonSpecies>::type::numberOfProtons;
+
+        using namespace bremsstrahlung;
+        BremsstrahlungFunctor bremsstrahlungFunctor(
+            scaledSpectrumMap.at(targetZ).getScaledSpectrumFunctor(),
+            scaledSpectrumMap.at(targetZ).getStoppingPowerFunctor(),
+            photonAngle.getPhotonAngleFunctor(),
+            currentStep);
+
+        creation::createParticlesFromSpecies(*electronSpeciesPtr, *photonSpeciesPtr, bremsstrahlungFunctor, cellDesc);
+    }
+
+}; // struct CallBremsstrahlung
 
 /** Handles the synchrotron radiation emission of photons from electrons
  *
--- a/src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.hpp
+++ b/src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.hpp
@@ -1,0 +1,159 @@
+/**
+ * Copyright 2016-2017 Heiko Burau
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "simulation_defines.hpp"
+#include "ScaledSpectrum.hpp"
+#include "PhotonEmissionAngle.hpp"
+#include "fields/FieldTmp.hpp"
+
+#include "random/methods/XorMin.hpp"
+#include "random/distributions/Uniform.hpp"
+#include "random/RNGProvider.hpp"
+
+#include "traits/Resolve.hpp"
+
+
+namespace picongpu
+{
+namespace particles
+{
+namespace bremsstrahlung
+{
+
+/** Handling of the Bremsstrahlung effect.
+ *
+ * Here the screened Bethe-Heitler cross section is used. See e.g.:
+ * Salvat, F., et al. "Monte Carlo simulation of bremsstrahlung emission by electrons."
+ * Radiation Physics and Chemistry 75.10 (2006): 1201-1219.
+ *
+ * The numerics separates the energy spectrum into two parts. In the low-energy part
+ * photon emission is neglected and a drag force is applied to the electrons. In the high-energy part
+ * photons are created in addition to the drag force.
+ *
+ * Electron deflection is treated as screened Rutherford scattering, see e.g. Jackson, chap. 13.5
+ *
+ * The photon emission angle is taken from the Lorentz-boosted dipole radiation formula,
+ * see e.g. Jackson, chap. 15.2
+ *
+ * \tparam T_ElectronSpecies
+ * \tparam T_PhotonSpecies
+ */
+template<typename T_IonSpecies, typename T_ElectronSpecies, typename T_PhotonSpecies>
+struct Bremsstrahlung
+{
+    typedef T_IonSpecies IonSpecies;
+    typedef T_ElectronSpecies ElectronSpecies;
+    typedef T_PhotonSpecies PhotonSpecies;
+
+    typedef typename ElectronSpecies::FrameType FrameType;
+
+    /* specify field to particle interpolation scheme */
+    typedef typename PMacc::traits::Resolve<
+        typename GetFlagType<FrameType,interpolation<> >::type
+        >::type Field2ParticleInterpolation;
+
+    /* margins around the supercell for the interpolation of the field on the cells */
+    typedef typename GetMargin<Field2ParticleInterpolation>::LowerMargin LowerMargin;
+    typedef typename GetMargin<Field2ParticleInterpolation>::UpperMargin UpperMargin;
+
+    /* relevant area of a block */
+    typedef SuperCellDescription<
+        typename MappingDesc::SuperCellSize,
+        LowerMargin,
+        UpperMargin
+        > BlockArea;
+
+    BlockArea BlockDescription;
+
+    typedef MappingDesc::SuperCellSize TVec;
+
+    typedef FieldTmp::ValueType ValueTypeIonDensity;
+
+private:
+    /* global memory ion density field device databoxes */
+    PMACC_ALIGN(ionDensityBox, FieldTmp::DataBoxType);
+    /* shared memory ion density device databoxes */
+    PMACC_ALIGN(cachedIonDensity, DataBox<SharedBox<ValueTypeIonDensity, typename BlockArea::FullSuperCellSize, 0> >);
+
+    PMACC_ALIGN(scaledSpectrumFunctor, ScaledSpectrum::LookupTableFunctor);
+    PMACC_ALIGN(stoppingPowerFunctor, ScaledSpectrum::LookupTableFunctor);
+    PMACC_ALIGN(getPhotonAngleFunctor, GetPhotonAngle::GetPhotonAngleFunctor);
+
+    PMACC_ALIGN(photonMom, float3_X);
+
+    /* random number generator */
+    typedef PMacc::random::RNGProvider<simDim, PMacc::random::methods::XorMin> RNGFactory;
+    typedef PMacc::random::distributions::Uniform<float> Distribution;
+    typedef typename RNGFactory::GetRandomType<Distribution>::type RandomGen;
+    RandomGen randomGen;
+
+public:
+    /* host constructor initializing member */
+    Bremsstrahlung(
+        const ScaledSpectrum::LookupTableFunctor& scaledSpectrumFunctor,
+        const ScaledSpectrum::LookupTableFunctor& stoppingPowerFunctor,
+        const GetPhotonAngle::GetPhotonAngleFunctor& getPhotonAngleFunctor,
+        const uint32_t currentStep);
+
+    /** Initialization function on device
+     *
+     * \brief Cache ion density field on device
+     *         and initialize possible prerequisites, like e.g. random number generator.
+     *
+     * This function will be called inline on the device which must happen BEFORE threads diverge
+     * during loop execution. The reason for this is the `__syncthreads()` call which is necessary after
+     * initializing the ion density field in shared memory.
+     */
+    DINLINE void init(const DataSpace<simDim>& blockCell, const int& linearThreadIdx, const DataSpace<simDim>& localCellOffset);
+
+    /** Rotates a vector to a given polar angle and a random azimuthal angle.
+     *
+     * @param vec vector to be rotated
+     * @param theta polar angle
+     * @return rotated vector
+     */
+    DINLINE float3_X scatterByTheta(const float3_X vec, const float_X theta);
+
+    /** Return the number of target particles to be created from each source particle.
+     *
+     * Called for each frame of the source species.
+     *
+     * @param sourceFrame Frame of the source species
+     * @param localIdx Index of the source particle within frame
+     * @return number of particle to be created from each source particle
+     */
+    DINLINE unsigned int numNewParticles(FrameType& sourceFrame, int localIdx);
+
+    /** Functor implementation.
+     *
+     * Called once for each single particle creation.
+     *
+     * \tparam Electron type of electron which creates the photon
+     * \tparam Photon type of photon that is created
+     */
+    template<typename Electron, typename Photon>
+    DINLINE void operator()(Electron& electron, Photon& photon);
+};
+
+} // namespace bremsstrahlung
+} // namespace particles
+} // namespace picongpu
--- a/src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
+++ b/src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
@@ -1,0 +1,261 @@
+/**
+ * Copyright 2016-2017 Heiko Burau
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "particles/particleToGrid/derivedAttributes/Density.hpp"
+#include "algorithms/Gamma.hpp"
+#include "algorithms/math/defines/sqrt.hpp"
+#include "algorithms/math/defines/dot.hpp"
+#include "algorithms/math/defines/cross.hpp"
+#include "traits/frame/GetMass.hpp"
+#include "traits/frame/GetCharge.hpp"
+#include "particles/operations/Assign.hpp"
+#include "particles/operations/Deselect.hpp"
+#include "particles/traits/GetAtomicNumbers.hpp"
+
+namespace picongpu
+{
+namespace particles
+{
+namespace bremsstrahlung
+{
+
+template<typename T_IonSpecies, typename T_ElectronSpecies, typename T_PhotonSpecies>
+Bremsstrahlung<T_IonSpecies, T_ElectronSpecies, T_PhotonSpecies>::Bremsstrahlung(
+    const ScaledSpectrum::LookupTableFunctor& scaledSpectrumFunctor,
+    const ScaledSpectrum::LookupTableFunctor& stoppingPowerFunctor,
+    const GetPhotonAngle::GetPhotonAngleFunctor& getPhotonAngleFunctor,
+    const uint32_t currentStep)
+        : scaledSpectrumFunctor(scaledSpectrumFunctor),
+          stoppingPowerFunctor(stoppingPowerFunctor),
+          getPhotonAngleFunctor(getPhotonAngleFunctor),
+          photonMom(float3_X::create(0)),
+          randomGen(RNGFactory::createRandom<Distribution>())
+{
+    DataConnector &dc = Environment<>::get().DataConnector();
+
+    /* initialize pointers on host-side tmp-field databoxes */
+    FieldTmp* fieldIonDensity = &(dc.getData<FieldTmp > (FieldTmp::getUniqueId( 0 ), true));
+    /* reset values to zero */
+    fieldIonDensity->getGridBuffer().getDeviceBuffer().setValue(FieldTmp::ValueType(0.0));
+
+    /* load species without copying the particle data to the host */
+    T_IonSpecies* ionSpecies = &(dc.getData<T_IonSpecies >(T_IonSpecies::FrameType::getName(), true));
+
+    /* compute ion density */
+    typedef typename CreateEnergyDensityOperation<T_IonSpecies>::type::Solver DensitySolver;
+    fieldIonDensity->computeValue < CORE + BORDER, DensitySolver > (*ionSpecies, currentStep);
+    dc.releaseData(T_IonSpecies::FrameType::getName());
+
+    /* initialize device-side tmp-field databoxes */
+    this->ionDensityBox = fieldIonDensity->getDeviceDataBox();
+}
+
+template<typename T_IonSpecies, typename T_ElectronSpecies, typename T_PhotonSpecies>
+DINLINE
+void Bremsstrahlung<T_IonSpecies, T_ElectronSpecies, T_PhotonSpecies>::init
+    (const DataSpace<simDim>& blockCell, const int& linearThreadIdx, const DataSpace<simDim>& localCellOffset)
+{
+    /* caching of ion density field */
+    cachedIonDensity = CachedBox::create < 0, ValueTypeIonDensity > (BlockArea());
+
+    /* instance of nvidia assignment operator */
+    nvidia::functors::Assign assign;
+    /* copy fields from global to shared */
+    const auto fieldIonDensityBlock = ionDensityBox.shift(blockCell);
+    ThreadCollective<BlockArea> collective(linearThreadIdx);
+    collective(
+              assign,
+              cachedIonDensity,
+              fieldIonDensityBlock
+              );
+
+    /* wait for shared memory to be initialized */
+    __syncthreads();
+
+    /* initialize random number generator with the local cell index in the simulation */
+    this->randomGen.init(localCellOffset);
+}
+
+
+template<typename T_IonSpecies, typename T_ElectronSpecies, typename T_PhotonSpecies>
+DINLINE
+float3_X Bremsstrahlung<T_IonSpecies, T_ElectronSpecies, T_PhotonSpecies>::scatterByTheta
+    (const float3_X vec, const float_X theta)
+{
+    using namespace PMacc::algorithms;
+
+    float_X sinTheta, cosTheta;
+    math::sincos(theta, sinTheta, cosTheta);
+
+    const float_X phi = -float_X(M_PI) + float_X(2.0) * float_X(M_PI) * this->randomGen();
+    float_X sinPhi, cosPhi;
+    math::sincos(phi, sinPhi, cosPhi);
+
+    const float3_X vecUp(0.0, 0.0, 1.0);
+    float3_X vecOrtho1 = math::cross(vecUp, vec);
+    const float_X vecOrtho1Abs = math::abs(vecOrtho1);
+
+    float3_X vecOrtho1_norm;
+    if(vecOrtho1Abs == float_X(0.0))
+        vecOrtho1_norm = float3_X(1.0, 0.0, 0.0);
+    else
+        vecOrtho1_norm = vecOrtho1 / vecOrtho1Abs;
+    const float3_X vecOrtho2 = math::cross(vecOrtho1_norm, vec);
+    vecOrtho1 = vecOrtho1_norm * math::abs(vec);
+
+    return vec * cosTheta +
+           vecOrtho1 * (sinTheta * cosPhi) +
+           vecOrtho2 * (sinTheta * sinPhi);
+}
+
+template<typename T_IonSpecies, typename T_ElectronSpecies, typename T_PhotonSpecies>
+DINLINE
+unsigned int Bremsstrahlung<T_IonSpecies, T_ElectronSpecies, T_PhotonSpecies>::numNewParticles(FrameType& sourceFrame, int localIdx)
+{
+    using namespace PMacc::algorithms;
+
+    auto particle = sourceFrame[localIdx];
+
+    /* particle position, used for field-to-particle interpolation */
+    const floatD_X pos = particle[position_];
+    const int particleCellIdx = particle[localCellIdx_];
+    /* multi-dim coordinate of the local cell inside the super cell */
+    const DataSpace<TVec::dim> localCell(DataSpaceOperations<TVec::dim>::template map<TVec > (particleCellIdx));
+    /* interpolation of fieldTmp */
+    const fieldSolver::numericalCellType::traits::FieldPosition<FieldTmp, simDim> fieldTmpPos;
+    const ValueTypeIonDensity ionDensity_norm = Field2ParticleInterpolation()
+        (cachedIonDensity.shift(localCell).toCursor(), pos, fieldTmpPos());
+
+    /* TODO: obtain the ion density from the molare ion density in order to avoid the rescaling.
+     * So this should be: ionDensity = ionMolDensity / UNIT_AMOUNT_SUBSTANCE */
+    const float_X ionDensity = ionDensity_norm.x() * particles::TYPICAL_NUM_PARTICLES_PER_MACROPARTICLE;
+
+    const float_X weighting = particle[weighting_];
+
+    const float_X c = SPEED_OF_LIGHT;
+    float3_X mom = particle[momentum_] / weighting;
+    const float_X momAbs = math::abs(mom);
+    float3_X mom_norm = mom / momAbs;
+
+    const float_X mass = frame::getMass<FrameType>();
+    const float_X Ekin = (Gamma<>()(mom, mass) - float_X(1.0)) * mass * c*c;
+    if(Ekin < electron::MIN_ENERGY)
+        return 0;
+
+    /* electron deflection due to Rutherford scattering without modifying the electron
+       energy based on radiation emission */
+    const float_X zMin = float_X(1.0) / (float_X(M_PI)*float_X(M_PI));
+    const float_X zMax = float_X(1.0) / (electron::MIN_THETA*electron::MIN_THETA);
+    const float_X z = zMin + this->randomGen() * (zMax - zMin);
+    const float_X theta = math::rsqrt(z);
+    const float_X targetZ = GetAtomicNumbers<T_IonSpecies>::type::numberOfProtons;
+    const float_X rutherfordCoeff = float_X(2.0) * ELECTRON_CHARGE*ELECTRON_CHARGE / (float_X(4.0) * float_X(M_PI) * EPS0) * targetZ / Ekin;
+    const float_X scaledDeflectionDCS = float_X(M_PI) * (zMax - zMin) * rutherfordCoeff*rutherfordCoeff;
+    const float_X deflectionProb = ionDensity * c * DELTA_T * scaledDeflectionDCS;
+
+    if(this->randomGen() < deflectionProb)
+    {
+        mom = this->scatterByTheta(mom, theta);
+        mom_norm = mom / momAbs;
+    }
+
+    /* non-radiative Bremsstrahlung */
+    const float_X kappaCutoff = math::min(photon::SOFT_PHOTONS_CUTOFF / Ekin, float_X(1.0));
+    const float_X stoppingPower = ionDensity * c * this->stoppingPowerFunctor(Ekin, kappaCutoff);
+    const float_X newEkin = math::max(Ekin - stoppingPower * DELTA_T, float_X(0.0));
+    const float_X newEkin_norm = newEkin / (mass * c*c);
+    /* This is based on: (p / mc)^2 = (E_kin / mc^2)^2 + 2 * (E_kin / mc^2) */
+    const float_X newMomAbs = mass * c * math::sqrt(newEkin_norm*newEkin_norm + float_X(2.0) * newEkin_norm);
+    const float_X deltaMom = newMomAbs - momAbs;
+    particle[momentum_] = (mom + deltaMom * mom_norm) * weighting;
+
+    /* photon emission */
+    const float_X delta = this->randomGen();
+    const float_X kappa = math::pow(kappaCutoff, delta);
+    const float_X scalingFactor = -math::log(kappaCutoff);
+    const float_X emissionProb = photon::WEIGHTING_RATIO * scalingFactor * ionDensity * c * DELTA_T * this->scaledSpectrumFunctor(Ekin, kappa);
+
+    // raise a warning if the emission probability is too high.
+    if(picLog::log_level & picLog::CRITICAL::lvl)
+    {
+        if(emissionProb > float_X(photon::SINGLE_EMISSION_PROB_LIMIT))
+        {
+            const float_X Ekin_SI = Ekin * UNIT_ENERGY;
+            printf("[Bremsstrahlung] warning: emission probability is too high: \
+                    p = %g, at Ekin = %g keV, kappa = %g, ion density = %g m^-3\n",
+                    emissionProb,
+                    Ekin_SI * UNITCONV_Joule_to_keV,
+                    kappa,
+                    ionDensity / (UNIT_LENGTH*UNIT_LENGTH*UNIT_LENGTH));
+        }
+    }
+
+    if(this->randomGen() < emissionProb)
+    {
+        const float_X photonEnergy = kappa * Ekin;
+        this->photonMom = mom_norm * weighting / photon::WEIGHTING_RATIO * photonEnergy / c;
+        return 1;
+    }
+
+    return 0;
+}
+
+
+template<typename T_IonSpecies, typename T_ElectronSpecies, typename T_PhotonSpecies>
+template<typename Electron, typename Photon>
+DINLINE
+void Bremsstrahlung<T_IonSpecies, T_ElectronSpecies, T_PhotonSpecies>::operator()
+    (Electron& electron, Photon& photon)
+{
+    auto destPhoton =
+        PMacc::particles::operations::deselect<
+            boost::mpl::vector<
+                multiMask,
+                momentum,
+                weighting
+            >
+        >(photon);
+
+    namespace parOp = PMacc::particles::operations;
+    parOp::assign( destPhoton, parOp::deselect<particleId>(electron) );
+
+    const float3_X elMom = electron[momentum_];
+    const float_X weighting = electron[weighting_] / photon::WEIGHTING_RATIO;
+    electron[momentum_] = elMom - this->photonMom; // ultra relativistic limit in terms of energy
+
+    /* photon emission angle */
+    const float_X mass = frame::getMass<FrameType>();
+    const float_X gamma = Gamma<>()(elMom / weighting, mass);
+
+    const float_X theta = this->getPhotonAngleFunctor(this->randomGen(), gamma);
+
+    const float3_X scatteredPhotonMom = this->scatterByTheta(this->photonMom, theta);
+
+    photon[multiMask_] = 1;
+    photon[momentum_] = scatteredPhotonMom;
+    photon[weighting_] = weighting;
+}
+
+
+} // namespace bremsstrahlung
+} // namespace particles
+} // namespace picongpu
--- a/src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
+++ b/src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
@@ -1,0 +1,263 @@
+/**
+ * Copyright 2016-2017 Heiko Burau
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "cuSTL/container/HostBuffer.hpp"
+#include "cuSTL/cursor/Cursor.hpp"
+#include "cuSTL/cursor/navigator/PlusNavigator.hpp"
+#include "cuSTL/cursor/tools/LinearInterp.hpp"
+#include "cuSTL/cursor/BufferCursor.hpp"
+#include "algorithms/math.hpp"
+#include <boost/array.hpp>
+#include <boost/shared_ptr.hpp>
+#include <boost/numeric/odeint/integrate/integrate.hpp>
+#include <boost/math/tools/minima.hpp>
+#include <limits>
+#include <utility>
+
+namespace picongpu
+{
+namespace particles
+{
+namespace bremsstrahlung
+{
+
+namespace detail
+{
+
+/** Functor mapping `delta` to the photon emission polar angle `theta`,
+ * where delta is a uniformly distributed random number between zero and one.
+ */
+struct GetPhotonAngleFunctor
+{
+    typedef typename ::PMacc::result_of::Functor<
+        ::PMacc::cursor::tools::LinearInterp<float_X>,
+        ::PMacc::cursor::BufferCursor<float_X, DIM2> >::type LinInterpCursor;
+
+    typedef float_X type;
+
+    LinInterpCursor linInterpCursor;
+    float_X lnMinGamma;
+    float_X lnMaxGamma;
+
+    /** constructor
+     *
+     * @param linInterpCursor lookup table for the photon emission angle.
+     */
+    HDINLINE GetPhotonAngleFunctor(LinInterpCursor linInterpCursor)
+        : linInterpCursor(linInterpCursor)
+    {
+        this->lnMinGamma = math::log(photon::MIN_GAMMA);
+        this->lnMaxGamma = math::log(photon::MAX_GAMMA);
+    }
+
+    /** Return the polar emission angle of the photon.
+     *
+     * @param delta uniformly distributed random number between zero and one.
+     * @param gamma relativistic factor of the incident electron.
+     */
+    HDINLINE float_X operator()(const float_X delta, const float_X gamma) const
+    {
+        const float_X deltaLookupPos = delta * static_cast<float_64>(photon::NUM_SAMPLES_DELTA - 1);
+
+        const float_X lnGamma = algorithms::math::log(gamma);
+        const float_X gammaLookupPos =
+            (lnGamma - this->lnMinGamma) /
+            (this->lnMaxGamma - this->lnMinGamma) *
+            static_cast<float_X>(photon::NUM_SAMPLES_GAMMA - 1);
+
+        if (picLog::log_level & picLog::CRITICAL::lvl)
+        {
+            if(gamma > photon::MAX_GAMMA)
+            {
+                printf("[Bremsstrahlung] error lookup table: gamma = %g is out of range.\n",
+                       gamma);
+            }
+        }
+
+        return this->linInterpCursor[float2_X(deltaLookupPos, gammaLookupPos)];
+    }
+};
+
+} // namespace detail
+
+/** Creates and holds the lookup table for the photon emission angle.
+ */
+struct GetPhotonAngle
+{
+    typedef detail::GetPhotonAngleFunctor GetPhotonAngleFunctor;
+
+private:
+
+    typedef boost::shared_ptr<PMacc::container::DeviceBuffer<float_X, DIM2> > MyBuf;
+    MyBuf dBufTheta;
+
+    /** probability density at polar angle theta.
+     * It's the ultrarelativistic limit of the dipole radiation formula, see e.g. Jackson, chap. 15.2
+     */
+    struct Probability
+    {
+        const float_64 gamma2;
+        Probability(const float_64 gamma) : gamma2(gamma*gamma) {}
+
+        template<typename T_State>
+        void operator()(const T_State &p, T_State &dpdtheta, const float_64 theta) const
+        {
+            const float_64 theta2 = theta*theta;
+            const float_64 denom = float_64(1.0) + gamma2 * theta2;
+
+            dpdtheta[0] = float_64(3.0) * theta * gamma2 * (float_64(1.0) + gamma2*gamma2 * theta2*theta2) /
+                          (denom*denom*denom*denom);
+        }
+    };
+
+    /** Return the absolute deviation of a delta, computed from a given theta, and a reference delta.
+     *
+     * Delta is the angular emission probability (normalized to one) integrated from zero to theta,
+     * where theta is the angle between the photon momentum and the final electron momentum.
+     */
+    struct AimForDelta
+    {
+        const float_64 targetDelta;
+        const float_64 gamma;
+
+        /** constructor
+         *
+         * @param targetDelta reference delta
+         * @param gamma relativistic factor
+         */
+        AimForDelta(const float_64 targetDelta, const float_64 gamma) :
+            targetDelta(targetDelta), gamma(gamma) {}
+
+        float_64 delta(const float_64 theta, const float_64 gamma) const
+        {
+            namespace odeint = boost::numeric::odeint;
+
+            typedef boost::array<float_64, 1> state_type;
+
+            state_type integral_result = {0.0};
+            const float_64 lowerLimit = 0.0;
+            const float_64 upperLimit = theta;
+            const float_64 stepwidth = (upperLimit - lowerLimit) / float_64(1000.0);
+            Probability integrand(gamma);
+            odeint::integrate(integrand, integral_result, lowerLimit, upperLimit, stepwidth);
+
+            return integral_result[0];
+        }
+
+        float_64 operator()(const float_64 theta) const
+        {
+            return math::abs(this->delta(theta, this->gamma) - this->targetDelta);
+        }
+    };
+
+    /** Return the maximal theta which corresponds to the maximal delta and a given gamma
+     *
+     * @param gamma relativistic factor
+     */
+    float_64 maxTheta(const float_64 gamma) const
+    {
+        AimForDelta aimForDelta(photon::MAX_DELTA, gamma);
+
+        std::pair<float_64, float_64> minimum;
+
+        minimum = boost::math::tools::brent_find_minima(
+            aimForDelta,
+            0.0,
+            M_PI,
+            std::numeric_limits<float_64>::digits);
+
+        return minimum.first;
+    }
+
+    /** computes the polar emission angle theta.
+     *
+     * @param delta uniformly distributed random number within [0, 1] or (0, 1)
+     * @param gamma relativistic factor
+     * @param maxTheta maximal theta
+     */
+    float_64 theta(const float_64 delta, const float_64 gamma, const float_64 maxTheta) const
+    {
+        AimForDelta aimForDelta(delta, gamma);
+        const float_64 minTheta = 0.0;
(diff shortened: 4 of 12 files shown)


PrometheusPi:
@Heikman Great job. I will have a look at your pull request as soon as possible.

PrometheusPi:
@Heikman Could you please check the spectrum/energy of the emitted photons.

PrometheusPi:
*offline discussion with @Heikman and @HighIander:*
Please check the influence of your Z_eff. Is charge screening negligible or not?

src/picongpu/include/particles/ParticlesFunctors.hpp
```diff
@@ -318,6 +320,60 @@ struct CallIonization
...
+    typedef bremsstrahlung::Bremsstrahlung<IonSpecies, ElectronSpecies, PhotonSpecies> BremsstrahlungFunctor;
+
+    /** Functor implementation
+     *
+     * \tparam T_StorageStuple contains info about the particle species
+     * \tparam T_CellDescription contains the number of blocks and blocksize
+     *                           that is later passed to the kernel
+     * \param tuple An n-tuple containing the type-info of multiple particle species
+     * \param cellDesc points to logical block information like dimension and cell sizes
+     * \param currentStep The current time step
```
PrometheusPi: please check capitalization of your words

Heikman:
The differential cross section of a fully ionized copper ion differs only about 10% at most (in the low energy limit) from the neutral atom.

PrometheusPi:
@Heikman great jobs with these tests.  Thank you!

I still have a question regarding your last plot "Electron deflection on multi-scattering":
What is the theoretical curve? A single scattering prediction or a multiple scattering prediction? If the later, does it take into account the averaged number of interactions (20) ore is the number of assumed interactions also random. 
(or in other words)
How do you explain the higher number of electrons at very low and very high theta compared to your theoretical prediction?

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -96,58 +96,62 @@ typedef Particles<
...
-    massRatio<MassRatioIons>,
-    chargeRatio<ChargeRatioIons>
-> ParticleFlagsIons;
+    massRatio<MassRatioElectrons>,
+    chargeRatio<ChargeRatioElectrons>
+#if(ENABLE_SYNCHROTRON_PHOTONS == 1)
+    ,synchrotronPhotons<PIC_Photons>
+#endif
+#if(ENABLE_BREMSSTRAHLUNG == 1)
+    ,bremsstrahlungIons<PIC_Ions>,
```
ax3l: chicken-egg problem. in other processes, such as ionization, the ion species need to know the electron species they shall create free electrons into.

please use aliases, ccing @psychocoderHPC

psychocoderHPC: Maybe it is a little bit more complicated and need some small changes in PIConGPU:

```C++
typedef  bmpl::string<'i'> Ion;
typedef  bmpl::string<'e'> Electron;

typdef Particles<
  ParticleDescription<
      Ion,
      SuperCellSize,
     ATTRIBUTELIST,
     bmpl::vector<Ionisation<Electron> >
   >
> PIC_Ion;

typdef Particles<
  ParticleDescription<
      Electron,
      SuperCellSize,
     ATTRIBUTELIST,
     bmpl::vector<Bremsstrahlung<Ion> >
   >
> PIC_Electron;

//...

typedef bmpl::vector<PIC_Electrons,PIC_Ions> VectorAllSpecies;
```

```C++
//SOLVER
template<typename T_Species>
struct Bremsstrahlung
{
  typedef GetSpecies<T_Species, VectorAllSpecies>::type MySpeciesType;
};
```

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
```diff
@@ -0,0 +1,252 @@
...
+
+    PMACC_AUTO(destPhoton,
+        PMacc::particles::operations::deselect<
+            boost::mpl::vector<
+                multiMask,
+                momentum
+            >
+        >(photon)
+    );
+    PMacc::particles::operations::assign( photon, electron );
```
psychocoderHPC: Please add tho folowing lines to support that particle ids can used.
```C++
namespace parOp = PMacc::particles::operations;
parOp::assign( photon, parOp::deselect<particleId>(electron) );
```

src/picongpu/include/particles/ParticlesFunctors.hpp
```diff
@@ -318,6 +319,60 @@ struct CallIonization
...
+ */
+template<typename T_ElectronSpeciesName>
+struct CallBremsstrahlung
+{
+    typedef T_ElectronSpeciesName ElectronSpeciesName;
+    typedef typename ElectronSpeciesName::type ElectronSpecies;
+
+    typedef typename PMacc::particles::traits::ResolveAliasFromSpecies<
+        ElectronSpecies,
+        bremsstrahlungIons<>
```
PrometheusPi: @Heikman How do you treat multiple ion species?

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.hpp
```diff
@@ -0,0 +1,173 @@
...
+        const uint32_t currentStep);
+
+    /** Initialization function on device
+     *
+     * \brief Cache ion density field on device
+     *         and initialize possible prerequisites, like e.g. random number generator.
+     *
+     * This function will be called inline on the device which must happen BEFORE threads diverge
+     * during loop execution. The reason for this is the `__syncthreads()` call which is necessary after
+     * initializing the E-/B-field shared boxes in shared memory.
```
PrometheusPi: @Heikman Why do you need E-/B-fields? (Is this an outdated comment?)

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+    /** Return the number of target particles to be created from each source particle.
+     *
+     * Called for each frame of the source species.
+     *
+     * @param sourceFrame Frame of the source species
+     * @param localIdx Index of the source particle within frame
+     * @return number of particle to be created from each source particle
+     */
+    DINLINE unsigned int numNewParticles(FrameType& sourceFrame, int localIdx);
```
PrometheusPi: @Heikman Could your target particles be something other than photons? If not - why not call this more clearly `numNewPhotons`?

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
```diff
@@ -0,0 +1,252 @@
...
+    /* compute ion density */
+    typedef typename CreateCounterOperation<T_IonSpecies>::type::Solver CounterSolver;
+    fieldIonCounter->computeValue < CORE + BORDER, CounterSolver > (*ionSpecies, currentStep);
+    dc.releaseData(T_IonSpecies::FrameType::getName());
+
+    /* initialize device-side tmp-field databoxes */
+    this->ionCounterBox = fieldIonCounter->getDeviceDataBox();
+
+    const float_64 softPhotonsCutoff_SI = photon::SOFT_PHOTONS_CUTOFF_keV * UNITCONV_keV_to_Joule;
+    this->softPhotonsCutoff = softPhotonsCutoff_SI / UNIT_ENERGY;
```
PrometheusPi: This is in SI units! Please move these two lines in `param`/`unitless` and use the unitless value here only.

PrometheusPi: second - `softPhotonsCutoff_SI / UNIT_ENERGY` is a `float_64` - `softPhotonsCutoff` is `float_X` - please cast "verbosely".

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
```diff
@@ -0,0 +1,252 @@
...
+    const float_X weighting = particle[weighting_];
+
+    const float_X c = SPEED_OF_LIGHT;
+    float3_X mom = particle[momentum_] / weighting;
+    const float_X momAbs = math::abs(mom);
+    float3_X mom_norm = mom / momAbs;
+
+    const float_X mass = frame::getMass<FrameType>();
+    const float_X momAbs_norm = momAbs / (mass * c);
+    const float_X Ekin = (math::sqrt(float_X(1.0) + momAbs_norm*momAbs_norm) - float_X(1.0)) * mass * c*c;
```
PrometheusPi: This equation is correct but might cause some numerical errors for small `momAbs_norm`. 

Additionally it is code duplication.
Could you use [`Gamma`](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/algorithms/Gamma.hpp) instead. It does currently the same but could be updated (in one place) to improve numerical stability.

Just calculate
```
E_kin = (gamma -1) m*c*c;
```

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
```diff
@@ -0,0 +1,252 @@
...
+    if(Ekin < electron::MIN_ENERGY)
+        return 0;
+
+    /* electron deflection */
+    const float_X zMin = float_X(1.0) / (float_X(M_PI)*float_X(M_PI));
+    const float_X zMax = float_X(1.0) / (electron::MIN_THETA*electron::MIN_THETA);
```
PrometheusPi: @Heikman Where is the minimal theta coming from? (I could not find it in the paper.)

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
```diff
@@ -0,0 +1,252 @@
...
+    const float_X theta2 = theta*theta;
+    const float_X rutherfordCoeff = ELECTRON_CHARGE*ELECTRON_CHARGE / (float_X(4.0) * float_X(M_PI) * EPS0) * ion::TARGET_Z / Ekin;
+    const float_X denominator = (electron::MIN_THETA*electron::MIN_THETA + theta2);
+    const float_X scaledDeflectionDCS = float_X(1.0) / float_X(2.0) * (zMax - zMin) * rutherfordCoeff*rutherfordCoeff *
+        float_X(2.0) * float_X(M_PI) * theta2*theta2 / (denominator*denominator);
+    const float_X deflectionProb = ionDensity * c * DELTA_T * scaledDeflectionDCS;
+
+    if(this->randomGen() < deflectionProb)
+    {
+        mom = this->scatterByTheta(mom, theta);
```
PrometheusPi: @Heikman Is here the meaning of `mom` changed. It has been the electron momentum so far. Is it now the photon momentum?

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
```diff
@@ -0,0 +1,252 @@
...
+        return 0;
+
+    /* electron deflection */
+    const float_X zMin = float_X(1.0) / (float_X(M_PI)*float_X(M_PI));
+    const float_X zMax = float_X(1.0) / (electron::MIN_THETA*electron::MIN_THETA);
+    const float_X z = zMin + this->randomGen() * (zMax - zMin);
+    const float_X theta = math::rsqrt(z);
+    const float_X theta2 = theta*theta;
+    const float_X rutherfordCoeff = ELECTRON_CHARGE*ELECTRON_CHARGE / (float_X(4.0) * float_X(M_PI) * EPS0) * ion::TARGET_Z / Ekin;
+    const float_X denominator = (electron::MIN_THETA*electron::MIN_THETA + theta2);
```
PrometheusPi: @Heikman: `theta`ranges from `electron::MIN_THETA` to *pi*. Why do you add an additional (arbitrary value) of `electron::MIN_THETA^2` here?

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
```diff
@@ -0,0 +1,252 @@
...
+    const float_X scaledDeflectionDCS = float_X(1.0) / float_X(2.0) * (zMax - zMin) * rutherfordCoeff*rutherfordCoeff *
+        float_X(2.0) * float_X(M_PI) * theta2*theta2 / (denominator*denominator);
```
PrometheusPi: could you combine the code to reduce the number of multiplications and avoid the double float_X(2.0)
```c++
const float_X factorScaledDeflectionDCS = rutherfordCoeff * theta2 / denominator;
const float_X scaledDeflectionDCS = (zMax - zMin) * float_X(M_PI) *  factorScaledDeflectionDCS * factorScaledDeflectionDCS;
```
This requires 1 division and 5 multiplications less.

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
```diff
@@ -0,0 +1,252 @@
...
+        mom = this->scatterByTheta(mom, theta);
+        mom_norm = mom / momAbs;
+    }
+
+    /* non-radiative Bremsstrahlung */
+    const float_X kappaCutoff = math::min(this->softPhotonsCutoff / Ekin, float_X(1.0));
+    const float_X stoppingPower = ionDensity * c * this->stoppingPowerFunctor(Ekin, kappaCutoff);
+    const float_X newEkin = math::max(Ekin - stoppingPower * DELTA_T, float_X(0.0));
+    const float_X newEkin_norm = newEkin / (mass * c*c);
+    const float_X newMomAbs = mass * c * math::sqrt(newEkin_norm*newEkin_norm + float_X(2.0) * newEkin_norm);
```
PrometheusPi: @Heikman Could you add a comment above this line stating that you derived this equation starting from
```
(E_kin + E_0)^2 = (pc)^2 + E_0^2
```

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
```diff
@@ -0,0 +1,252 @@
...
+    const float_X deltaMom = newMomAbs - momAbs;
+    particle[momentum_] = (mom + deltaMom * mom_norm) * weighting;
+
+    /* photon emission */
+    const float_X delta = this->randomGen();
+    const float_X kappa = math::pow(kappaCutoff, float_X(1.0) - delta);
```
PrometheusPi: @Heikman Why do you "invert" the range of your random number `alpha = float_X(1.0) - delta`?
If it is because `delta` is in `[0, 1)` and `alpha` is in `(0, 1]` please state this in a comment.

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
```diff
@@ -0,0 +1,252 @@
...
+
+    // raise a warning if the emission probability is too high.
+    if(picLog::log_level & picLog::CRITICAL::lvl)
+    {
+        if(emissionProb > float_X(photon::SINGLE_EMISSION_PROB_LIMIT))
+        {
+            const float_X Ekin_SI = Ekin * UNIT_ENERGY;
+            printf("[Bremsstrahlung] warning: emission probability is too high: \
+                    p = %g, at Ekin = %g keV, kappa = %g, ion density = %g m^-3\n",
+                emissionProb,
```
PrometheusPi: please fix alignment

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
```diff
@@ -0,0 +1,252 @@
...
+    const float_X scalingFactor = -math::log(kappaCutoff);
+    const float_X emissionProb = scalingFactor * ionDensity * c * DELTA_T * this->scaledSpectrumFunctor(Ekin, kappa);
+
+    // raise a warning if the emission probability is too high.
+    if(picLog::log_level & picLog::CRITICAL::lvl)
+    {
+        if(emissionProb > float_X(photon::SINGLE_EMISSION_PROB_LIMIT))
+        {
+            const float_X Ekin_SI = Ekin * UNIT_ENERGY;
+            printf("[Bremsstrahlung] warning: emission probability is too high: \
```
PrometheusPi: This should go to stderr not to stdout

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
```diff
@@ -0,0 +1,252 @@
...
+    );
+    PMacc::particles::operations::assign( photon, electron );
+
+    const float3_X elMom = electron[momentum_];
+    const float_X weighting = electron[weighting_];
+    electron[momentum_] = elMom - this->photonMom; // ultra relativistic limit in terms of energy
+
+    /* photon emission angle */
+    const float_X mass = frame::getMass<FrameType>();
+    const float_X gamma = Gamma<>()(elMom / weighting, mass);
```
PrometheusPi: `Gamma` is a struct that can be given to the implementation via template parameter as done for example in the pushers.

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
```diff
@@ -0,0 +1,264 @@
...
+
+        const float_X lnGamma = algorithms::math::log(gamma);
+        const float_X binGamma =
+            (lnGamma - this->lnMinGamma) /
+            (this->lnMaxGamma - this->lnMinGamma) *
+            static_cast<float_X>(photon::NUM_SAMPLES_GAMMA - 1);
+
+        if (picLog::log_level & picLog::CRITICAL::lvl)
+        {
+            if(binGamma > float_X(photon::NUM_SAMPLES_GAMMA - 1))
```
PrometheusPi: This is the same check as `gamma > photon::MAX_GAMMA` which is (in my opinion) easier to read. @Heikman Could you change this check?

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
```diff
@@ -0,0 +1,264 @@
...
+            {
+                printf("[Bremsstrahlung] error lookup table: gamma = %g is out of range.\n",
```
PrometheusPi: This log message should go to stderr.

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
psychocoderHPC: Please check if this part is kernel code. On device yhere is no stderr or stdout avail.

Am 23. August 2016 09:09:57 MESZ, schrieb Richard Pausch <notifications@github.com>:
>> +    const float_X deltaMom = newMomAbs - momAbs;
>> +    particle[momentum_] = (mom + deltaMom * mom_norm) * weighting;
>> +
>> +    /* photon emission */
>> +    const float_X delta = this->randomGen();
>> +    const float_X kappa = math::pow(kappaCutoff, float_X(1.0) -
>delta);
>> +    const float_X scalingFactor = -math::log(kappaCutoff);
>> +    const float_X emissionProb = scalingFactor * ionDensity * c *
>DELTA_T * this->scaledSpectrumFunctor(Ekin, kappa);
>> +
>> +    // raise a warning if the emission probability is too high.
>> +    if(picLog::log_level & picLog::CRITICAL::lvl)
>> +    {
>> +        if(emissionProb >
>float_X(photon::SINGLE_EMISSION_PROB_LIMIT))
>> +        {
>> +            const float_X Ekin_SI = Ekin * UNIT_ENERGY;
>> +            printf("[Bremsstrahlung] warning: emission probability
>is too high: \
>
>This should go to stderr not to stdout
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/ComputationalRadiationPhysics/picongpu/pull/1504/files/a6cadca2fe833a2b91fd025e38cf2e60e8faeca4#r75812425

-- 

PrometheusPi: @psychocoderHPC You are right - this is device code only. So this is fine!
Or does `fprintf(stderr, "...")` work?

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
PrometheusPi: same here - see answer of @psychocoderHPC above

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
```diff
@@ -0,0 +1,264 @@
...
+        Probability(const float_64 gamma) : gamma2(gamma*gamma) {}
+
+        template<typename T_State>
+        void operator()(const T_State &p, T_State &dpdtheta, const float_64 theta) const
+        {
+            const float_64 theta2 = theta*theta;
+            const float_64 denom = 1.0 + gamma2 * theta2;
+
+            dpdtheta[0] = 3.0 * theta * gamma2 * (1.0 + gamma2*gamma2 * theta2*theta2) /
+                          (denom*denom*denom*denom);
```
PrometheusPi: @Heikman: I assume this is the angular probability of the scattered photon.
If compared to [F. Salvat et al.](http://www.sciencedirect.com/science/article/pii/S0969806X05001325) (Eq. 26) you are missing a normalization factor of `1/(2*pi)`. The same is true for [Jackson](https://books.google.de/books?id=tYEAkAEACAAJ&dq=john+david+jackson+classical+electrodynamics&hl=de&sa=X&ved=0ahUKEwjsr4vdkNfOAhVI7xQKHaJHD3wQ6AEIHzAA) (Eq. 13.36). Here you are having additionally a factor `theta` too much. How did you derive this probability?

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
```diff
@@ -0,0 +1,264 @@
...
+        }
+    };
+
+    /** Gives the absolute deviation of a delta, computed from a given theta, and a reference delta.
```
PrometheusPi: Could you please state what the physical meaning of `delta` is in this context. Previously is has been a random number.

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
```diff
@@ -0,0 +1,264 @@
...
+            0.0,
+            M_PI,
+            std::numeric_limits<float_64>::digits);
+
+        return minimum.first;
+    }
+
+    /** computes the polar emission angle theta.
+     *
+     * @param delta uniformly distributed random number within [0, 1]
```
PrometheusPi: Are you sure about the inclusive limits of random number?

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
```diff
@@ -0,0 +1,264 @@
...
+    {
+        AimForDelta aimForDelta(delta, gamma);
+        const float_64 minTheta = 0.0;
+        std::pair<float_64, float_64> minimum;
+
+        minimum = boost::math::tools::brent_find_minima(
+            aimForDelta,
+            minTheta,
+            maxTheta,
+            std::numeric_limits<float_64>::digits);
```
PrometheusPi: This is a quite strict limit - is this really required - or could or reduce this number and thus save computational time (or is computational time not an issue with this algorithm)?

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
```diff
@@ -0,0 +1,264 @@
...
+        {
+            const float_64 lnGamma_norm = static_cast<float_64>(gammaIdx) /
+                                          static_cast<float_64>(photon::NUM_SAMPLES_GAMMA - 1);
+            const float_64 gamma = pam::exp(lnMinGamma + (lnMaxGamma - lnMinGamma) * lnGamma_norm);
+            const float_64 maxTheta = this->maxTheta(gamma);
+
+            for(uint32_t deltaIdx = 0; deltaIdx < photon::NUM_SAMPLES_DELTA; deltaIdx++)
+            {
+                const float_64 delta = photon::MAX_DELTA * static_cast<float_64>(deltaIdx) /
+                                       static_cast<float_64>(photon::NUM_SAMPLES_DELTA - 1);
```
PrometheusPi: You could calculate 
```c++
const float_64 deltaDelta = photon::MAX_DELTA / static_cast<float_64>(photon::NUM_SAMPLES_DELTA - 1);
```
in advance and than just compute:
```c++
for(uint32_t deltaIdx = 0; deltaIdx < photon::NUM_SAMPLES_DELTA; deltaIdx++)
{
   const float_64 delta = deltaDelta * static_cast<float_64>(deltaIdx);
    ...
}
```
thus avoiding `photon::NUM_SAMPLES_DELTA -1` additional subtractions and divisions.

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
```diff
@@ -0,0 +1,264 @@
...
+
+                *curTheta(deltaIdx, gammaIdx) = static_cast<float_X>(this->theta(delta, gamma, maxTheta));
+            }
+        }
+
+        *this->dBufTheta = hBufTheta;
+    }
+
+    /** Return a functor mapping `delta` to the photon emission polar angle `theta`,
+     * where delta is a uniformly distributed random number between zero and one.
```
PrometheusPi: again - please specify the range of the possible random numbers more precisely (if needed)

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+
+    /** constructor
+     *
+     * @param linInterpCursor
+     */
+    HDINLINE LookupTableFunctor(LinInterpCursor linInterpCursor)
+        : linInterpCursor(linInterpCursor)
+    {
+        const float_64 EMin_SI = electron::MIN_ENERGY_MeV * 1.0e3 * UNITCONV_keV_to_Joule;
+        const float_64 EMax_SI = electron::MAX_ENERGY_MeV * 1.0e3 * UNITCONV_keV_to_Joule;
```
PrometheusPi: Please use only PIConGPU units outside the `param`/`unitless` files.

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+
+        const float_X binE = (lnE - this->lnEMin) / (this->lnEMax - this->lnEMin) * static_cast<float_X>(electron::NUM_SAMPLES_EKIN - 1);
+        // in the low-energy limit Bremsstrahlung is not taken into account
+        if(binE < float_X(0.0))
+            return float_X(0.0);
+        const float_X binKappa = kappa * static_cast<float_X>(electron::NUM_SAMPLES_KAPPA - 1);
+
+        if (picLog::log_level & picLog::CRITICAL::lvl)
+        {
+            if(binE > float_X(electron::NUM_SAMPLES_EKIN - 1))
```
PrometheusPi: This is fine. In my opinion a check for `Ekin > Emax` would be more intuitive.

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+            {
+                const float_64 Ekin_SI = Ekin * UNIT_ENERGY;
+                printf("[Bremsstrahlung] error lookup table: Ekin=%g MeV is out of range.\n",
+                       float_X(Ekin_SI * UNITCONV_Joule_to_keV * 1.0e-3));
```
PrometheusPi: Here, SI units are fine with me since this is a error message.

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+            }
+            if(binKappa < float_X(0.0) || binKappa > float_X(electron::NUM_SAMPLES_KAPPA - 1))
```
PrometheusPi: here a check for `kappa < 0.0 || kappa > 1.0` would be more intuitive **and** you could avoid a cast.

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+        const float_64 r_e = classicalElRadius;
+        const float_64 alpha = fineStructureConstant;
+
+        const float_64 W = kappa * Ekin;
+        const float_64 eps = W / (Ekin + m_e * c2);
+        const float_64 R = pam::pow(static_cast<float_64>(ion::TARGET_Z), -1.0/3.0) * bohrRadius;
+        const float_64 gamma = Ekin / (m_e * c2) + 1.0;
+        const float_64 b = R * m_e * c / HBAR / (2.0 * gamma) * eps / (1.0 - eps);
+
+        const float_64 phi_1 = 4.0 * pam::log(R * m_e * c / HBAR) + 2.0 - 2.0 * pam::log(1.0 + b*b) - 4.0 * b * pam::atan2(1.0 / b, 1.0);
```
PrometheusPi: @Heikman Why do you prefer `atan2` here over `atan` if you are not using the sign dependency of the second argument?

PrometheusPi: especially here `pam::atan2(1.0, b)` would be the correct usage of `atan2`.

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+        const float_64 phi_2 = 4.0 * pam::log(R * m_e * c / HBAR) + 7.0/3.0 - 2.0 * pam::log(1.0 + b*b) - 6.0 * b * pam::atan2(1.0 / b, 1.0) \
+                             - b*b * (4.0 - 4.0 * b * pam::atan2(1.0 / b, 1.0) - 3.0 * pam::log(1.0 + 1.0 / (b*b)));
```
PrometheusPi: same here and the line above

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+        const float_64 c2 = c*c;
+        const float_64 m_e = ELECTRON_MASS;
+        const float_64 r_e = classicalElRadius;
+        const float_64 alpha = fineStructureConstant;
+
+        const float_64 W = kappa * Ekin;
+        const float_64 eps = W / (Ekin + m_e * c2);
+        const float_64 R = pam::pow(static_cast<float_64>(ion::TARGET_Z), -1.0/3.0) * bohrRadius;
+        const float_64 gamma = Ekin / (m_e * c2) + 1.0;
+        const float_64 b = R * m_e * c / HBAR / (2.0 * gamma) * eps / (1.0 - eps);
```
PrometheusPi: could you summarize the dividend and divisor in order to reduce the number of divisions?

src/picongpu/include/particles/ParticlesFunctors.hpp
Heikman: I assume a single ion species.

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.hpp
Heikman: This is a policy function used by `ParticleCreator`.

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+
+        const float_64 c = SPEED_OF_LIGHT;
+        const float_64 c2 = c*c;
+        const float_64 m_e = ELECTRON_MASS;
+        const float_64 r_e = classicalElRadius;
+        const float_64 alpha = fineStructureConstant;
+
+        const float_64 W = kappa * Ekin;
+        const float_64 eps = W / (Ekin + m_e * c2);
+        const float_64 R = pam::pow(static_cast<float_64>(ion::TARGET_Z), -1.0/3.0) * bohrRadius;
```
PrometheusPi: @Heikman Are you sure about this equation? In Salvat et al. below Eq. 6, this is just a scaling relation not an equation.

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
Heikman: It's still the electron momentum.

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+    {
+        namespace pam = PMacc::algorithms::math;
+        /*
+        const float_64 c2 = SPEED_OF_LIGHT*SPEED_OF_LIGHT;
+        const float_64 m_e = ELECTRON_MASS;
+
+        const float_64 beta = pam::sqrt(Ekin * (Ekin + 2.0 * m_e * c2)) / (Ekin + m_e * c2);*/
+        const float_64 W = kappa * Ekin;
+
+        return /*beta*beta / (ion::TARGET_Z*ion::TARGET_Z) **/ W * this->dcs(Ekin, kappa);
```
PrometheusPi: please remove the commented part of your (old) version

PrometheusPi: but than this is not chi anymore - please find a better/other name or go back to the original version

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
PrometheusPi: But it is rotated by theta!

The initial electron momentum has been `mom` (before this line).
The final electron momentum is later `mom + deltaMom * mom_norm`.
The scattered photon momentum is `mom_norm * weighting * photonEnergy / c`.
With `mom_norm = mom / momAbs`. 

So here `mom` (or `mom_norm`) becomes parallel to both the photon momentum and the new electron momentum. I thought the bremsstrahlung as described by F. Salvat et al. just reduces the magnitude of the electron momentum not its direction. 

Is this counter propagation (of photon and electron) here not violating momentum conservation?
```
p_e(before interaction) = mom (before line 170)
p_e(after interaction) = scatterByTheta(mom, theta) + scatterByTheta(mom, theta)/abs(scatterByTheta(mom, theta)) * factor1 = scatterByTheta(mom, theta)/abs(scatterByTheta(mom, theta)) * (1 + factor1) = vec * (1+factor1)
p_ph(after interaction) = scatterByTheta(mom, theta)/abs(scatterByTheta(mom, theta)) * factor2 = vec * factor2
```
so if `scatterByTheta(mom, theta)=vec` is not parallel to `mom (before line 170)`, there is **no**  conservation of momentum. 

What am I assuming wrong here? 
Could you give it a meaningful (`const`) name without reusing old momentum vectors

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+        hBufScaledSpectrum.assign(float_X(0.0));
+        hBufStoppingPower.assign(float_X(0.0));
+
+        PMACC_AUTO(curScaledSpectrum, hBufScaledSpectrum.origin());
+        PMACC_AUTO(curStoppingPower, hBufStoppingPower.origin());
+
+        const float_64 EMin_SI = electron::MIN_ENERGY_MeV * 1.0e3 * UNITCONV_keV_to_Joule;
+        const float_64 EMax_SI = electron::MAX_ENERGY_MeV * 1.0e3 * UNITCONV_keV_to_Joule;
+        const float_64 lnEMin = pam::log(EMin_SI / UNIT_ENERGY);
+        const float_64 lnEMax = pam::log(EMax_SI / UNIT_ENERGY);
```
PrometheusPi: In these lines you are using SI units - please switch to PIConGPU units

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+                float_64 kappa = static_cast<float_64>(kappaIdx) /
+                                 static_cast<float_64>(electron::NUM_SAMPLES_KAPPA - 1);
+                if(kappa == 0.0)
+                    kappa = electron::MIN_KAPPA;
+
+                const float_64 lnE_norm = static_cast<float_64>(EkinIdx) /
+                                          static_cast<float_64>(electron::NUM_SAMPLES_EKIN - 1);
+                const float_64 Ekin = pam::exp(lnEMin + (lnEMax - lnEMin) * lnE_norm);
+
+                *curScaledSpectrum(EkinIdx, kappaIdx) = static_cast<float_X>(this->chi(Ekin, kappa));
```
PrometheusPi: The index range is (in both dimensions) by one smaller then the buffer size. Thus the upper boarder of the buffer array is still zero. Is this intentional?

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+
+                state_type integral_result = {0.0};
+                const float_64 lowerLimit = electron::MIN_KAPPA * Ekin;
+                const float_64 upperLimit = kappa * Ekin;
+                const float_64 stepwidth = upperLimit / 1.0e3;
```
PrometheusPi: Where is this magic number coming from? - Please avoid any magic numbers!

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+                StoppingPowerIntegrand integrand(Ekin, *this);
+                odeint::integrate(integrand, integral_result, lowerLimit, upperLimit, stepwidth);
+                *curStoppingPower(EkinIdx, kappaIdx) = static_cast<float_X>(integral_result[0]);
+
+                // check for nans
+                if(*curScaledSpectrum(EkinIdx, kappaIdx) != *curScaledSpectrum(EkinIdx, kappaIdx))
+                {
+                    const float_64 Ekin_SI = Ekin * UNIT_ENERGY;
+                    const float_64 Ekin_MeV = Ekin_SI * UNITCONV_Joule_to_keV / 1.0e3;
```
PrometheusPi: SI units -> PIConGPU units

PrometheusPi: Sorry here it is OK.

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+                    std::stringstream errMsg;
+                    errMsg << "[Bremsstrahlung] lookup table (scaled spectrum) has NaN-entry at Ekin = "
```
PrometheusPi: Just for me (@PrometheusPi) Here, SI units are fine since it is a error message.

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,312 @@
...
+            new PMacc::container::DeviceBuffer<float_X, DIM2>(
+                electron::NUM_SAMPLES_EKIN + 1,
+                electron::NUM_SAMPLES_KAPPA + 1));
+        this->dBufStoppingPower = MyBuf(
+            new PMacc::container::DeviceBuffer<float_X, DIM2>(
+                electron::NUM_SAMPLES_EKIN + 1,
+                electron::NUM_SAMPLES_KAPPA + 1));
+
+        PMacc::container::HostBuffer<float_X, DIM2> hBufScaledSpectrum(this->dBufScaledSpectrum->size());
+        PMacc::container::HostBuffer<float_X, DIM2> hBufStoppingPower(this->dBufStoppingPower->size());
```
PrometheusPi: This allocates memory. (?) Do you free it anywhere?

src/picongpu/include/simulationControl/MySimulation.hpp
```diff
@@ -355,6 +357,15 @@ class MySimulation : public SimulationHelper<simDim>
...
             this->synchrotronFunctions.init();
         }
+
+        // Initialize bremsstrahlung lookup tables, if there are species containing bremsstrahlung photons
+        typedef typename PMacc::particles::traits::FilterByFlag<VectorAllSpecies,
+                                                                bremsstrahlungPhotons<> >::type AllBremsstrahlungPhotonsSpecies;
+        if(!bmpl::empty<AllBremsstrahlungPhotonsSpecies>::value)
+        {
+            this->scaledBremsstrahlungSpectrum.init();
+            this->bremsstrahlungPhotonAngle.init();
```
PrometheusPi: Please move this part (memory allocation) before the particle memory allocation to avoid bug #1547

src/picongpu/include/simulation_defines/param/bremsstrahlung.param
```diff
@@ -0,0 +1,132 @@
...
+#ifndef ENABLE_BREMSSTRAHLUNG
+#define ENABLE_BREMSSTRAHLUNG 0
+#endif
+
+/** ion related params
+ */
+namespace ion
+{
+    /** atomic number of the target ion species */
+    BOOST_CONSTEXPR_OR_CONST uint32_t TARGET_Z = 29;
```
PrometheusPi: This is actually an attribute of the ion species - see [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/particles/traits/GetAtomicNumbers.hpp).

PrometheusPi:
great use of encapsulated namespaces in your `param` file :+1:
Please have a look at the comments - especially on the open questions regarding physics.

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
Heikman: Your expression is essentially the same as mine:
```
gamma = sqrt(1 + (p / mc)^2)
=> E_kin = (gamma -1) m*c*c = (sqrt(1 + (p / mc)**2) - 1) m*c**2
```

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
Heikman: The minimum theta is coming from Jackson chap. 13.5

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
PrometheusPi: I know, I just want to avoid code duplication and es mentioned before automatically include improvements done in the general gamma calculation.

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
psychocoderHPC: IMO it should by `if(binGamma >= photon::NUM_SAMPLES_GAMMA )`
`binGamma` should be an integer type because there are no half bins. This also avoid aut of memory access.

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
Heikman: Here the angular probability is depending on the polar angle `theta` instead of the solid angle `Omega`, so:
```
dp / dtheta = dp / dOmega * 2pi * theta
```

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
Heikman: The time to generate the lookup table is neglectable.

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
Heikman: This part is not time-critical.

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
Heikman: this s not time critical

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
Heikman: The buffers have a margin of one cell with undefined content to ensure access within [0, 1].

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
Heikman: It's freed when the buffer is destroyed at the end of the scope (RAII).

src/picongpu/include/particles/ParticlesFunctors.hpp
n01r: doesn't it feel nice? :cat:

PrometheusPi: :+1:

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -177,19 +174,26 @@ typedef bmpl::vector<
...
     current<UsedParticleCurrentSolver>,
-    massRatio<MassRatioIons>,
-    chargeRatio<ChargeRatioIons>
-> ParticleFlagsIons;
+    massRatio<MassRatioElectrons>,
+    chargeRatio<ChargeRatioElectrons>
+#if(ENABLE_SYNCHROTRON_PHOTONS == 1)
+    ,synchrotronPhotons<PIC_Photons>
+#endif
+#if(ENABLE_BREMSSTRAHLUNG == 1)
```
psychocoderHPC: @ComputationalRadiationPhysics/picongpu-maintainers 

Should we keep the `ENABLE_BREMSSTRAHLUNG` in the gernic picongpu part or should we only show this as  within the `LaserWakefield` example.
The reason why I am asking is that the user need to define `ENABLE_BREMSSTRAHLUNG` within the cmake to enable this code path.hHow to enable is only visiable in the example `LaserWakefield` in the file [cmakeFlags](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1504/files#diff-17e35932abac366d12163ba8d613a304L42)

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.tpp
```diff
@@ -0,0 +1,221 @@
...
+
+    const float_64 c = SPEED_OF_LIGHT;
+    const float_64 c2 = c*c;
+    const float_64 m_e = ELECTRON_MASS;
+    const float_64 r_e = classicalElRadius;
+    const float_64 alpha = fineStructureConstant;
+
+    const float_64 W = kappa * Ekin;
+    const float_64 eps = W / (Ekin + m_e * c2);
+    const float_64 R = pam::pow(targetZ, -1.0/3.0) * bohrRadius;
```
psychocoderHPC: note: since #1472 it is possible to use `math::pow()`

src/picongpu/include/simulation_defines/param/speciesDefinition.param
ax3l: I would prefer if @Heikman just creates a very small, new example, lets say a laser on a 20n_c foil, that uses a minimal set of input `.param` files to test Bremsstrahlung and synchrotron photons. That would avoid overloading our LWFA example even more and removes unnecessary `#define` macro magic.

You can also add a little notebook that then visualizes the expected output if you want with [openPMD-viewer](https://github.com/openPMD/openPMD-viewer).

Is that ok with you? :)

src/picongpu/include/simulation_defines/unitless/bremsstrahlung.unitless
```diff
@@ -0,0 +1,53 @@
...
+BOOST_CONSTEXPR_OR_CONST float_X MAX_ENERGY = MAX_ENERGY_SI / UNIT_ENERGY;
+
+BOOST_CONSTEXPR_OR_CONST float_X NUM_STEPS_STOPPING_POWER_INTERGRAL = 1.0e3;
+
+} // namespace electron
+
+namespace photon
+{
+
+BOOST_CONSTEXPR_OR_CONST float_64 SOFT_PHOTONS_CUTOFF_SI = SOFT_PHOTONS_CUTOFF_keV * UNITCONV_keV_to_Joule;
```
ax3l: please use `constexpr` directly now

src/picongpu/include/simulation_defines/param/bremsstrahlung.param
```diff
@@ -0,0 +1,124 @@
...
+    /** Maximal value of delta for the lookup table.
+     * Delta is the angular emission probability (normalized to one) integrated from zero to theta,
+     * where theta is the angle between the photon momentum and the final electron momentum.
+     *
+     * A value close to one is reasonable. Though exactly one was actually correct,
+     * because it would map to theta = pi (maximum polar angle), the sampling then would be bad
+     * in the ultrarelativistic case. In this regime the emission primarily takes place at small thetas.
+     * So a maximum delta close to one maps to a reasonable maximum theta.
+     */
+    BOOST_CONSTEXPR_OR_CONST float_64 MAX_DELTA = 0.95;
```
ax3l: please use `constexpr` directly now

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
```diff
@@ -0,0 +1,267 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "cuSTL/container/HostBuffer.hpp"
```
ax3l: unused includes?

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,181 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "particles/traits/GetAtomicNumbers.hpp"
+
+#include "cuSTL/container/HostBuffer.hpp"
```
ax3l: unused includes?

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
Heikman: It's used in line 227.

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
psychocoderHPC: Yes please include header always where they are needed.

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
```diff
@@ -0,0 +1,267 @@
...
+    struct Probability
+    {
+        const float_64 gamma2;
+        Probability(const float_64 gamma) : gamma2(gamma*gamma) {}
+
+        template<typename T_State>
+        void operator()(const T_State &p, T_State &dpdtheta, const float_64 theta) const
+        {
+            const float_64 theta2 = theta*theta;
+            const float_64 denom = 1.0 + gamma2 * theta2;
```
psychocoderHPC: please cast `1.0` with `float_64(1.0)`

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
```diff
@@ -0,0 +1,267 @@
...
+
+            dpdtheta[0] = 3.0 * theta * gamma2 * (1.0 + gamma2*gamma2 * theta2*theta2) /
```
psychocoderHPC: please use `float_X(3.0)` and `float_X(1.0)`

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
```diff
@@ -0,0 +1,267 @@
...
+        float_64 delta(const float_64 theta, const float_64 gamma) const
+        {
+            namespace odeint = boost::numeric::odeint;
+
+            typedef boost::array<float_64, 1> state_type;
+
+            state_type integral_result = {0.0};
+            const float_64 lowerLimit = 0.0;
+            const float_64 upperLimit = theta;
+            const float_64 stepwidth = (upperLimit - lowerLimit) / 1000.0;
```
psychocoderHPC: cast

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,180 @@
...
+{
+
+/** Functor for the scaled differential cross section (dcs) which
+ * equals to the electron energy loss times the cross section per unit energy.
+ */
+struct LookupTableFunctor
+{
+    typedef typename ::PMacc::result_of::Functor<
+        ::PMacc::cursor::tools::LinearInterp<float_X>,
+        ::PMacc::cursor::BufferCursor<float_X, DIM2> >::type LinInterpCursor;
```
psychocoderHPC: Is this only working in a 2d simulation.

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.tpp
```diff
@@ -0,0 +1,221 @@
...
+
+
+/** constructor
+ *
+ * @param linInterpCursor
+ */
+HDINLINE LookupTableFunctor::LookupTableFunctor(LinInterpCursor linInterpCursor)
+    : linInterpCursor(linInterpCursor)
+{
+    this->lnEMin = algorithms::math::log(electron::MIN_ENERGY);
```
psychocoderHPC: note: it is possile to skip `algorithms::`, also in the following methods

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.tpp
```diff
@@ -0,0 +1,221 @@
...
+        return float_X(0.0);
+    const float_X binKappa = kappa * static_cast<float_X>(electron::NUM_SAMPLES_KAPPA - 1);
+
+    if (picLog::log_level & picLog::CRITICAL::lvl)
+    {
+        if(Ekin < electron::MIN_ENERGY || Ekin > electron::MAX_ENERGY)
+        {
+            const float_64 Ekin_SI = Ekin * UNIT_ENERGY;
+            printf("[Bremsstrahlung] error lookup table: Ekin=%g MeV is out of range.\n",
+                   float_X(Ekin_SI * UNITCONV_Joule_to_keV * 1.0e-3));
```
psychocoderHPC: please write `float_X(1.0e-3)`

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.tpp
```diff
@@ -0,0 +1,221 @@
...
+    const float_64 r_e = classicalElRadius;
+    const float_64 alpha = fineStructureConstant;
+
+    const float_64 W = kappa * Ekin;
+    const float_64 eps = W / (Ekin + m_e * c2);
+    const float_64 R = math::pow(targetZ, -1.0/3.0) * bohrRadius;
+    const float_64 gamma = Ekin / (m_e * c2) + 1.0;
+    const float_64 b = R * m_e * c / HBAR / (2.0 * gamma) * eps / (1.0 - eps);
+
+    const float_64 phi_1 = 4.0 * math::log(R * m_e * c / HBAR) + 2.0 - 2.0 * math::log(1.0 + b*b) - 4.0 * b * math::atan(1.0 / b);
```
psychocoderHPC: please cast all constant values

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -163,6 +138,31 @@ using PIC_Ions = Particles<
...
+value_identifier( float_X, ChargeRatioElectrons, 1.0 );
+
+using ParticleFlagsElectrons = bmpl::vector<
+    particlePusher< UsedParticlePusher >,
+    shape< UsedParticleShape >,
+    interpolation< UsedField2Particle >,
+    current< UsedParticleCurrentSolver >,
+    massRatio< MassRatioElectrons >,
+    chargeRatio< ChargeRatioElectrons >
+#if( ENABLE_SYNCHROTRON_PHOTONS == 1 )
```
psychocoderHPC: please remove `ENABLE_SYNCHROTRON_PHOTONS` and please it in the example without the guard.

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -163,6 +138,31 @@ using PIC_Ions = Particles<
     ParticleFlagsIons
 >;
 
+/*--------------------------- electrons --------------------------------------*/
+
+/* ratio relative to BASE_CHARGE and BASE_MASS */
+value_identifier( float_X, MassRatioElectrons, 1.0 );
```
psychocoderHPC: please revert all changes in this file

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
```diff
@@ -0,0 +1,267 @@
...
+{
+
+/** Functor mapping `delta` to the photon emission polar angle `theta`,
+ * where delta is a uniformly distributed random number between zero and one.
+ */
+struct GetPhotonAngleFunctor
+{
+    typedef typename ::PMacc::result_of::Functor<
+        ::PMacc::cursor::tools::LinearInterp<float_X>,
+        ::PMacc::cursor::BufferCursor<float_X, DIM2> >::type LinInterpCursor;
```
Heikman: no, `DIM2` is related to the energy and angle axis.

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
Heikman: is this really neccesary? (it's not `float_X`)

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
Heikman: no, `DIM2` is related to the energy and angle axis.

psychocoderHPC:
You mist to change some of my commented code parts.

Heikman:
you tell me when it's time to rebase.

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
PrometheusPi: :+1:

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
PrometheusPi: :+1:

src/picongpu/include/particles/bremsstrahlung/PhotonEmissionAngle.hpp
PrometheusPi: But it would make the code easier to read. Please use this const value.

examples/Bremsstrahlung/include/simulation_defines/param/starter.param
```diff
@@ -0,0 +1,35 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+
+#pragma once
```
psychocoderHPC: there is no need to copy this file, please copy only param file must be changed for your example

src/picongpu/include/particles/bremsstrahlung/Bremsstrahlung.tpp
```diff
@@ -0,0 +1,260 @@
...
+void Bremsstrahlung<T_IonSpecies, T_ElectronSpecies, T_PhotonSpecies>::init
+    (const DataSpace<simDim>& blockCell, const int& linearThreadIdx, const DataSpace<simDim>& localCellOffset)
+{
+    /* caching of ion density field */
+    cachedIonCounter = CachedBox::create < 0, ValueTypeIonCounter > (BlockArea());
+
+    /* instance of nvidia assignment operator */
+    nvidia::functors::Assign assign;
+    /* copy fields from global to shared */
+    PMACC_AUTO(fieldIonCounterBlock, ionCounterBox.shift(blockCell));
```
psychocoderHPC: please use c++11 auto else it wont compile see #1743 (remove `PMACC_AUTO`)

src/picongpu/include/simulationControl/MySimulation.hpp
```diff
@@ -480,6 +494,19 @@ class MySimulation : public SimulationHelper<simDim>
                 MakeIdentifier<bmpl::_1> > synchrotronRadiation;
         synchrotronRadiation(forward(particleStorage), cellDescription, currentStep, this->synchrotronFunctions);
 
+        /* Bremsstrahlung */
+        typedef typename PMacc::particles::traits::FilterByFlag<
+            VectorAllSpecies,
+            bremsstrahlungIons<>
+        >::type VectorSpeciesWithBremsstrahlung;
+        ForEach<VectorSpeciesWithBremsstrahlung,
+                particles::CallBremsstrahlung<bmpl::_1>,
```
psychocoderHPC: Could you please use one indention instead of align.

src/picongpu/include/simulationControl/MySimulation.hpp
```diff
@@ -480,6 +494,19 @@ class MySimulation : public SimulationHelper<simDim>
...
+                MakeIdentifier<bmpl::_1> > particleBremsstrahlung;
+        particleBremsstrahlung(forward(particleStorage),
+                               cellDescription,
```
psychocoderHPC: please use on indention lvl.

ax3l:
@Heikman can you please rebase on `dev` and force push again? The travis tests were accidentally disabled in the last days (always positive).

Heikman:
I want to fine tune the physical parameters first before merging.

bussmann:
Sounds like cheating ;-)

ax3l:
@Heikman do you mind putting the new example in a follow up so we can merge this PR?

You will otherwise be constantly rebasing and I already see things I would annotate due to wrong usage in the new example.

Heikman:
ready to be merged?

src/picongpu/include/simulationControl/MySimulation.hpp
```diff
@@ -321,6 +325,16 @@ class MySimulation : public SimulationHelper<simDim>
...
         }
 
+        // Initialize bremsstrahlung lookup tables, if there are species containing bremsstrahlung photons
+        if(!bmpl::empty<AllBremsstrahlungPhotonsSpecies>::value)
+        {
+            ForEach<AllBremsstrahlungPhotonsSpecies,
+                particles::bremsstrahlung::FillScaledSpectrumMap<bmpl::_1> > fillScaledSpectrumMap;
+            fillScaledSpectrumMap(forward(this->scaledBremsstrahlungSpectrumMap));
+
+            this->bremsstrahlungPhotonAngle.init();
```
psychocoderHPC: @Heikman  Is there a problem with this init method that the particle memory is not allocated at this point?

Note: memory for particles is allocated at [line 338](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1504/files#diff-27ab651b1e352e1783fb08c3aed891c9R338).

ax3l:
@PrometheusPi please add your review :)

src/picongpu/include/simulationControl/MySimulation.hpp
Heikman: no, that's fine.

PrometheusPi: QUESTION:
This is still  open, please answer.

src/picongpu/include/simulation_defines/unitless/bremsstrahlung.unitless
```diff
@@ -0,0 +1,53 @@
...
+namespace electron
+{
+
+constexpr float_64 MIN_ENERGY_SI = MIN_ENERGY_MeV * 1.0e3 * UNITCONV_keV_to_Joule;
+constexpr float_X MIN_ENERGY = MIN_ENERGY_SI / UNIT_ENERGY;
+
+constexpr float_64 MAX_ENERGY_SI = MAX_ENERGY_MeV * 1.0e3 * UNITCONV_keV_to_Joule;
+constexpr float_X MAX_ENERGY = MAX_ENERGY_SI / UNIT_ENERGY;
+
+constexpr float_X NUM_STEPS_STOPPING_POWER_INTERGRAL = 1.0e3;
```
PrometheusPi: QUESTION:
Why is this part of the unitless and not part of the param?

src/picongpu/include/particles/bremsstrahlung/ScaledSpectrum.hpp
```diff
@@ -0,0 +1,180 @@
...
+    template<typename T_Map>
+    void operator()(T_Map& map) const
+    {
+        const float_X targetZ = GetAtomicNumbers<IonSpecies>::type::numberOfProtons;
+
+        if(map.count(targetZ) == 0)
+        {
+            ScaledSpectrum scaledSpectrum;
+            scaledSpectrum.init(static_cast<float_64>(targetZ));
+            map[targetZ] = scaledSpectrum;
```
Heikman: `map` is a STL-map.

src/picongpu/include/simulation_defines/unitless/bremsstrahlung.unitless
Heikman: it is not a real user parameter.

PrometheusPi:
can be merges as soon as the compile test is finished

PR #1508, opened 2016-06-30
Dependency: Boost 1.57.0+

ax3l:
- ~~Does not need a merge before 0.2.0 release branch is opened!~~

For compatibility between C++98/11 build versions of boost and our build, we increase the dependency to boost 1.57.0+

Refs:
- https://svn.boost.org/trac/boost/ticket/6124
- https://svn.boost.org/trac/boost/ticket/6779
- https://svn.boost.org/trac/boost/ticket/10038

> # 6124, #6779, and #10038 are all the same problem.
> 
> The private library interface has been changed to use a plain-old C++03 enum.
> This is the fix suggested by Andy in 6779.
> The fix was too late for 1.57 beta 1, but should be in the final 1.57 release.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -56,7 +56,7 @@ Requirements
   - *Debian/Ubuntu:* `sudo apt-get install zlib1g-dev`
   - *Arch Linux:* `sudo pacman --sync zlib`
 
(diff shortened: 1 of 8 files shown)


psychocoderHPC:
please rebase

PR #1511, opened 2016-07-06
PMacc Types: Name Exchanges

ax3l:
Name the exchanges in PMacc.


src/libPMacc/include/pmacc_types.hpp
```diff
@@ -98,11 +98,38 @@ typedef long long int int64_cu;
...
+ * Axis index relation:
+ *   right & left are in X
+ *   bottom & top are in Y
+ *   back & front are in Z
  */
 enum ExchangeType
 {
     RIGHT = 1u, LEFT = 2u, BOTTOM = 3u, TOP = 6u, BACK = 9u, FRONT = 18u // 3er-System
 };
+struct ExchangeTypeNames
```
psychocoderHPC: maybe we should add an empty line?

PR #1513, opened 2016-07-06
Plugin: Fix ChargeConservation Checkpointing

ax3l:
Fix a segfault in the charge conservation (checker) plugin which is triggered when the plugin is disabled (default) and checkpoints are written.

ccing @Heikman


Heikman:
Thanks for fixing!

ax3l:
no prob!

PR #1507, opened 2016-06-28
add trait `GetStringProperties`

psychocoderHPC:
- add trait `GetStringProperties<>`
- add function `getStringProperties()`
- add object `StringProperty`

PR #1514, opened 2016-07-06
PIC Methods: Add (openPMD) String Properties

ax3l:
Add ([openPMD](https://github.com/openPMD/openPMD-standard) compatible) string properties to all essential PIC solvers so we can query them easily in plugins, bindings and I/O.

Ideally, all new parts of the PIC algorithm should now try to implement names properties (see #1507) the same way as we implement margins, etc. to make the code base self-describing.

### DIFF
--- a/src/picongpu/include/fields/FieldManipulator.hpp
+++ b/src/picongpu/include/fields/FieldManipulator.hpp
@@ -20,12 +20,15 @@
 
 #pragma once
 
-#include "pmacc_types.hpp"
 #include "simulation_defines.hpp"
 #include "simulation_classTypes.hpp"
 #include "mappings/simulation/GridController.hpp"
+#include "memory/dataTypes/Mask.hpp"
 #include "FieldManipulator.kernel"
 #include "simulationControl/MovingWindow.hpp"
+
+#include <string>
+#include <sstream>
 
 namespace picongpu
 {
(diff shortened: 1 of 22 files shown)


src/picongpu/include/fields/FieldManipulator.hpp
```diff
@@ -82,8 +85,57 @@ class FieldManipulator
             }
         }
     }
-};
 
+    static PMacc::traits::StringProperty getStringProperties()
+    {
+        PMacc::traits::StringProperty propList;
+        const DataSpace<DIM3> periodic =
+            Environment<simDim>::get().EnvironmentController().getCommunicator().getPeriodic();
```
ax3l: - [ ] depends on #1510

src/picongpu/include/fields/FieldManipulator.hpp
```diff
@@ -82,8 +85,57 @@ class FieldManipulator
...
+                if( isPeriodic )
+                    boundaryName = "periodic";
 
+                if( boundaryName == "open" )
+                {
+                    std::ostringstream ss;
+                    ss << ABSORBER_CELLS[axis][axisDir] << " cells";
+                    const std::string numAbsorb( ss.str() );
+                    const std::string boundaryParam =
+                        std::string("exponential damping over ") + numAbsorb;
```
psychocoderHPC: Exponential dumping over X. But what is X cells,meter,...

ax3l: finally reads: `exponential damping over 32 cells` (see `numAbsorb`)

ax3l:
@psychocoderHPC ping :)

psychocoderHPC:
Sry I gave an introduction in PIConGPU/CPU and forgot the pull request

PR #1520, opened 2016-07-14
openPMD: ADIOS

ax3l:
Implements the openPMD standard including the ED-PIC extension for our ADIOS plugin.

After first visual tests, this works with the latest ADIOS 1.9.0+ release but some string endings are broken if the latest `master` (~ https://github.com/ornladios/ADIOS/commit/b1038dea0145e46474b610163a184c000c371566) of ADIOS is not used (fixed bug).

What is not implemented is the refactoring of `particlePatches` (former `particles_info`) for openPMD (optional and recommended). This can be done in a follow-up PR and also after the next release. For now this does still work and one can generate the particle patch information in post-processing (I have done this successfully on Titan during the last year).

Also, I could not test the [openPMD-validator](https://github.com/openPMD/openPMD-validator) scripts since `bp2h5` eats some attributes during convert and the ADIOS `numpy` wrapper is not yet fully object oriented/h5py like https://github.com/ornladios/ADIOS/issues/56 which makes maintaining [such checker scripts](https://github.com/openPMD/openPMD-validator/blob/1.0.0/checkOpenPMD_h5.py) still too cumbersome.

Depends on:
- [x] #1518
- [x] #1519

### DIFF
--- a/src/picongpu/include/plugins/adios/ADIOSCountParticles.hpp
+++ b/src/picongpu/include/plugins/adios/ADIOSCountParticles.hpp
@@ -47,6 +47,10 @@
 #include "plugins/adios/writer/ParticleAttributeSize.hpp"
 #include "compileTime/conversion/RemoveFromSeq.hpp"
 
+#include "particles/traits/GetSpeciesFlagName.hpp"
+
+#include <string>
+
 namespace picongpu
 {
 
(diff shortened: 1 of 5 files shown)


psychocoderHPC:
ready for rebase

src/picongpu/include/plugins/adios/ADIOSCountParticles.hpp
```diff
@@ -87,14 +91,17 @@ struct ADIOSCountParticles
...
-    HINLINE void operator()(ThreadParams* params,
-                            const std::string subGroup)
+    HINLINE void operator()(ThreadParams* params)
     {
         DataConnector &dc = Environment<>::get().DataConnector();
         GridController<simDim>& gc = Environment<simDim>::get().GridController();
         uint64_t mpiSize = gc.getGlobalSize();
         uint64_t mpiRank = gc.getGlobalRank();
 
+        const std::string speciesGroup( FrameType::getName() + std::string("/") );
```
psychocoderHPC: only a note: it is allowed to write ` const std::string speciesGroup( FrameType::getName() + "/" );`

ax3l: [true story!](http://www.cplusplus.com/reference/string/string/operator+/)

PR #1523, opened 2016-07-15
IO: Density & Charge Density

ax3l:
User interface, `fileOutput.param`: Refactor the previous "sloppy" `density` which always was a `charge density` and add a real `density` option in `FieldTmp` generated, particle-derived fields.

### Changes in Meaning
- `DensityOperation`: from charge density to real density
- `ChargeDensityOperation`: new, previously called `DensityOperation`

(namings in the output files have already been fixed in previous PRs such as #1259 in the current dev)


ax3l:
@Heikman this is the fix to the I/O interface you spotted

PrometheusPi:
looks good to me

PR #1526, opened 2016-07-15
fix illegal memory access

psychocoderHPC:
remove shared memory `bool` with `int` variable to allow save atomic operations

The issue with the old implementation is that there is no atomicExch operation for `bool` and if we use a reinterpret cast to `int` we overwrite more than 1byte and create an memory overflow.


ax3l:
this bug only affects non-CUDA architectures (CPU alpaka version).

PR #1527, opened 2016-07-15
avoid division by zero

psychocoderHPC:
If no zero particles per cell should be created the implementation of `Quiet` started creates a division by zero.
This pull request solves the problem and calculate the position only if we need to create a particle.


ax3l:
great fix.
just for the record: this is a non-critical bug as it does not influence the result of the init process (but div by zeroes are not cool in any case).

PR #1541, opened 2016-08-02
Specialize ADK for linear or circular laser polarization

n01r:
The ADK model has been updated so that now the user is given the choice between the linear and circular polarization case, according to [Delone, Krainov 1998 Phys.-Usp. 41 469](http://iopscience.iop.org/article/10.1070/PU1998v041n05ABEH000393).

### DIFF
--- a/src/picongpu/include/particles/ionization/byField/ADK/ADK.def
+++ b/src/picongpu/include/particles/ionization/byField/ADK/ADK.def
@@ -29,9 +29,7 @@ namespace particles
 namespace ionization
 {
 
-    /** \struct ADK_Impl
-     *
-     * \brief Ammosov-Delone-Krainov tunneling model
+    /** Ammosov-Delone-Krainov tunneling model
      *
      * \tparam T_DestSpecies electron species to be created
      * \tparam T_SrcSpecies ion species to be ionized
@@ -39,12 +37,10 @@ namespace ionization
      *         cannot be known in list of particle species' flags
      *         \see speciesDefinition.param
      */
-    template<typename T_DestSpecies, typename T_SrcSpecies = bmpl::_1>
+    template<typename T_IonizationAlgorithm, typename T_DestSpecies, typename T_SrcSpecies = bmpl::_1>
     struct ADK_Impl;
 
-    /** \struct ADK
-     *
-     * \brief Ammosov-Delone-Krainov tunneling model
+    /** Ammosov-Delone-Krainov tunneling model - linear laser polarization
      *
      * - takes the ionization energies of the various charge states of ions
      * - calculates the ionization rates and then the ionization probabilities from them
@@ -61,9 +57,39 @@ namespace ionization
      * \see speciesDefinition.param
      */
     template<typename T_DestSpecies>
-    struct ADK
+    struct ADKLinPol
     {
-        typedef ADK_Impl<T_DestSpecies> type;
+        /* Boolean value that results in an additional polarization factor in
+         * the ionization rate for linear polarization */
+        BOOST_STATIC_CONSTEXPR bool linPol = true;
+        typedef particles::ionization::AlgorithmADK<linPol> IonizationAlgorithm;
+        typedef ADK_Impl<IonizationAlgorithm, T_DestSpecies> type;
+    };
+
+    /** Ammosov-Delone-Krainov tunneling model - circular laser polarization
+     *
+     * - takes the ionization energies of the various charge states of ions
+     * - calculates the ionization rates and then the ionization probabilities from them
+     * - ATTENTION: this approach is not very applicable for rapidly changing high intensity laser fields
+     * - this is a Monte Carlo method: if a random number is smaller
+     *   or equal than the ionization probability -> increase the charge state
+     * - see for example: Delone, N. B.; Krainov, V. P. (1998).
+     *   "Tunneling and barrier-suppression ionization of atoms and ions in a laser radiation field"
+     *   doi:10.1070/PU1998v041n05ABEH000393
+     *
+     * wrapper class,
+     * needed because the SrcSpecies cannot be known during the
+     * first specialization of the ionization model in the particle definition
+     * \see speciesDefinition.param
+     */
+    template<typename T_DestSpecies>
+    struct ADKCircPol
+    {
+        /* Boolean value that results in an additional polarization factor in
+         * the ionization rate for linear polarization */
+        BOOST_STATIC_CONSTEXPR bool linPol = false;
+        typedef particles::ionization::AlgorithmADK<linPol> IonizationAlgorithm;
+        typedef ADK_Impl<IonizationAlgorithm, T_DestSpecies> type;
     };
 
 } // namespace ionization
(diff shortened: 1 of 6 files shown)


n01r:
Compiles and runs.
Checking against experiments when I find the time. ^^;

bussmann:
+1

PrometheusPi:
Could you check `laserProfile::Polarisation` to decide for whether one should use 
 - linear `laserProfile::Polarisation<=2u` or 
 - circular `laserProfile::Polarisation==4u`?

n01r:
Sadly not all laser profiles are using this enum type, yet. Until then I thought it's better to have two independent realizations of ADK.

PrometheusPi:
you mix British and American english in `ionization`/`ionisation` and `polarization`/`polarisation`.

n01r:
my apologies :bow: ... it's an internal conflict
will fix that

src/picongpu/include/particles/ionization/byField/ADK/AlgorithmADKLinPol.hpp
```diff
@@ -0,0 +1,120 @@
...
+                float_X eInAU = math::abs(eField) / ATOMIC_UNIT_EFIELD;
+
+                /* effective principal quantum number (unitless) */
+                float_X nEff = protonNumber / math::sqrt(float_X(2.0) * iEnergy );
+                /* nameless variable for convenience dFromADK*/
+                float_X dBase = float_X(4.0) * util::cube(protonNumber) / (eInAU * util::quad(nEff)) ;
+                float_X dFromADK = math::pow(dBase,nEff);
+
+                /* factor from averaging over one laser cycle with LINEAR polarization */
+                float_X polarizationFactor = math::sqrt(float_X(3.0) * util::cube(nEff) * eInAU / (pi * util::cube(protonNumber)));
```
PrometheusPi: since this line is the **only** difference between `AlgorithmADKCircPol`and `AlgorithmADKLinPol`, but the  code for `AlgorithmADKLinPol` takes about 120 additional lines, could you create a `AlgorithmADKPol<...>`that takes a template argument to switch between both cases? That would reduce the code duplication dramatically.

n01r:
I now pass a boolean value as a template parameter that activates the calculation of the additional polarization factor for linear polarization or leaves it out for circular polarization. As we discussed offline before: we cannot just check for the polarization so easily because the laser could be initialized via an HDF5 input file.

src/picongpu/include/particles/ionization/byField/ADK/AlgorithmADK.hpp
```diff
@@ -82,8 +85,21 @@ namespace ionization
                 float_X dBase = float_X(4.0) * util::cube(protonNumber) / (eInAU * util::quad(nEff)) ;
                 float_X dFromADK = math::pow(dBase,nEff);
 
+                float_X polarizationFactor;
+                /* set up the polarization factor */
+                if(T_linPol)
```
PrometheusPi: This is still a run-time decision thus is not as efficient as possible. 
Just use:
```c++
float_X polarizationFactor = TEMPLATE_PARAM::polarizationFactor;
```
with a struct containing the value **not just a bool**.

src/picongpu/include/particles/ionization/byField/ADK/ADK.def
```diff
@@ -61,9 +83,13 @@ namespace ionization
...
      */
     template<typename T_DestSpecies>
-    struct ADK
+    struct ADKCircPol
     {
-        typedef ADK_Impl<T_DestSpecies> type;
+        /* Boolean value that results in an additional polarization factor in
+         * the ionization rate for linear polarization */
+        BOOST_STATIC_CONSTEXPR bool linPol = false;
+        typedef particles::ionization::AlgorithmADK<linPol> IonizationAlgorithm;
```
PrometheusPi: Please give a struct with the float value of the polarization factor instead of a `bool`.

src/picongpu/include/particles/ionization/byField/ADK/AlgorithmADK.hpp
n01r: The value is not `const` so I cannot just calculate it during compile time (it depends on the ionization potential). Nevertheless, it is known if the bool is `true` or `false` during compile time. Therefore the compiler should optimize the code in such a way that the unnecessary code vanishes.

src/picongpu/include/particles/ionization/byField/ADK/ADK.def
n01r: the value has to be calculated during runtime -- see above

src/picongpu/include/particles/ionization/byField/ADK/AlgorithmADK.hpp
n01r: @psychocoderHPC 
your verdict?

src/picongpu/include/particles/ionization/byField/ADK/ADK.def
PrometheusPi: OK, thanks for clarifying. :+1:

src/picongpu/include/particles/ionization/byField/ADK/AlgorithmADK.hpp
psychocoderHPC: @PrometheusPi  Your suggestion is not possible because like @n01r said the factor is only known at runtime.

A possible solution would be to add a functor for the factor calculation but it is a little bit complicated because of the needed input values.

My suggestion: (small design change)

```C++
/* ionization rate */
float_X rateADK = polarizationFactor \
                * eInAU * util::square(dFromADK) / (float_X(8.0) * pi * protonNumber) \
                * math::exp(float_X(-2.0) * util::cube(protonNumber) / (float_X(3.0) * util::cube(nEff) * eInAU));

if(T_linPol)
{
    /* factor from averaging over one laser cycle with LINEAR polarization */
    const float_X polarizationFactor = math::sqrt(float_X(3.0) * util::cube(nEff) * eInAU / (pi * util::cube(protonNumber)));
    rateADK *= polarizationFactor;
}
``` 
Then the code is better readable.

n01r: :+1: Yep, looks good to me. I think it's okay to stay away from overcomplicating things and not introduce a new functor for this.

src/picongpu/include/particles/ionization/byField/ADK/AlgorithmADK.hpp
```diff
@@ -82,11 +85,23 @@ namespace ionization
                 float_X dBase = float_X(4.0) * util::cube(protonNumber) / (eInAU * util::quad(nEff)) ;
                 float_X dFromADK = math::pow(dBase,nEff);
 
+                /* factor for CIRCULAR polarization */
+                float_X polarizationFactor = float_X(1.);
```
PrometheusPi: please remove this line as discussed offline

PrometheusPi:
physics looks OK to me :+1:

PR #1542, opened 2016-08-02
Make use of faster RNG implementation for ionization

n01r:
**Important refactoring**

Greatly overdue, the random number generator for the ADK model (and all future rate models for ionization) has been exchanged. 
Before we used an RNG that was creating new states during each time step of the simulation which were unique for each step, particle species, cell and MPI rank.
Now the state is created only once and then saved as well as changed for each cell after each use. This implies the uniqueness of the seed for the aforementioned conditions and reduces workload at the same time.

This now allows for faster testing of all the ionization models and more importantly: **highly increased simulation speed** for more sophisticated ionization models than BSI.

Big kudos to @Flamefire ! :tada: 

This pull request depends on the changes in #1541

### DIFF
--- a/src/picongpu/include/particles/ionization/byField/ADK/ADK_Impl.hpp
+++ b/src/picongpu/include/particles/ionization/byField/ADK/ADK_Impl.hpp
@@ -36,6 +36,10 @@
 
 #include "particles/ionization/ionizationMethods.hpp"
 
+#include "random/methods/XorMin.hpp"
+#include "random/distributions/Uniform.hpp"
+#include "random/RNGProvider.hpp"
+
 namespace picongpu
 {
 namespace particles
@@ -83,9 +87,10 @@ namespace ionization
             /* define ionization ALGORITHM (calculation) for ionization MODEL */
             typedef T_IonizationAlgorithm IonizationAlgorithm;
 
-            /* random number generator for Monte Carlo */
-            typedef particles::ionization::RandomNrForMonteCarlo<SrcSpecies> RandomGen;
-            /* \todo fix: cannot PMACC_ALIGN() because it seems to be too large */
+            /* random number generator */
+            typedef PMacc::random::RNGProvider<simDim, PMacc::random::methods::XorMin> RNGFactory;
+            typedef PMacc::random::distributions::Uniform<float> Distribution;
+            typedef typename RNGFactory::GetRandomType<Distribution>::type RandomGen;
             RandomGen randomGen;
 
             typedef MappingDesc::SuperCellSize TVec;
@@ -101,7 +106,7 @@ namespace ionization
 
         public:
             /* host constructor initializing member : random number generator */
-            ADK_Impl(const uint32_t currentStep) : randomGen(currentStep)
+            ADK_Impl(const uint32_t currentStep) : randomGen(RNGFactory::createRandom<Distribution>())
             {
                 DataConnector &dc = Environment<>::get().DataConnector();
                 /* initialize pointers on host-side E-(B-)field databoxes */
@@ -147,11 +152,11 @@ namespace ionization
                           fieldEBlock
                           );
 
-                /* initialize random number generator with the local cell index in the simulation*/
-                randomGen.init(localCellOffset);
-
                 /* wait for shared memory to be initialized */
                 __syncthreads();
+
+                /* initialize random number generator with the local cell index in the simulation */
+                this->randomGen.init(localCellOffset);
             }
 
             /** Functor implementation
@@ -186,7 +191,7 @@ namespace ionization
                 IonizationAlgorithm ionizeAlgo;
                 ionizeAlgo(
                      bField, eField,
-                     particle, randomGen()
+                     particle, this->randomGen()
                      );
 
                 /* determine number of new macro electrons to be created */
--- a/src/picongpu/include/particles/ionization/byField/ADK/ADK_Impl.hpp
+++ b/src/picongpu/include/particles/ionization/byField/ADK/ADK_Impl.hpp
@@ -20,8 +20,11 @@
 
 #pragma once
 
+#include <boost/type_traits/integral_constant.hpp>
+
 #include "simulation_defines.hpp"
 #include "traits/Resolve.hpp"
+#include "traits/UsesRNG.hpp"
 #include "mappings/kernel/AreaMapping.hpp"
 
 #include "fields/FieldB.hpp"
@@ -42,6 +45,18 @@
 
 namespace picongpu
 {
+namespace traits
+{
+    /** specialization of the UsesRNG trait
+     * --> ionization module uses random number generation
+     */
+    template<typename T_IonizationAlgorithm, typename T_DestSpecies, typename T_SrcSpecies>
+    struct UsesRNG<particles::ionization::ADK_Impl<T_IonizationAlgorithm, T_DestSpecies, T_SrcSpecies> > :
+    public boost::true_type
+    {
+    };
+} // namespace traits
+
 namespace particles
 {
 namespace ionization
--- a/src/picongpu/include/traits/UsesRNG.hpp
+++ b/src/picongpu/include/traits/UsesRNG.hpp
@@ -1,0 +1,43 @@
+/**
+ * Copyright 2016 Marco Garten, Rene Widera
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <boost/type_traits/integral_constant.hpp>
+
+namespace picongpu
+{
+namespace traits
+{
+
+/** Checks if an object requires the RNG
+ *
+ * @tparam T_Object any object (class or typename)
+ *
+ * This struct must inherit from (boost::true_type/false_type)
+ */
+template<typename T_Object>
+struct UsesRNG : public boost::false_type
+{
+};
+
+}// namespace traits
+
+}// namespace picongpu
--- a/src/picongpu/include/particles/ionization/ionizationMethods.hpp
+++ b/src/picongpu/include/particles/ionization/ionizationMethods.hpp
@@ -89,68 +89,6 @@ namespace ionization
         }
     };
 
-    /* Random number generation */
-    namespace nvrng = nvidia::rng;
-    namespace rngMethods = nvidia::rng::methods;
-    namespace rngDistributions = nvidia::rng::distributions;
-
-    template<typename T_SpeciesType>
-    struct RandomNrForMonteCarlo
-    {
-        typedef T_SpeciesType SpeciesType;
-        typedef typename MakeIdentifier<SpeciesType>::type SpeciesName;
-
-        HINLINE RandomNrForMonteCarlo(uint32_t currentStep) : isInitialized(false)
-        {
-            typedef typename SpeciesType::FrameType FrameType;
-
-            GlobalSeed globalSeed;
-            mpi::SeedPerRank<simDim> seedPerRank;
-            /* creates global seed that is unique for
-             * the particle species and ionization as a method in PIC
-             */
-            seed = globalSeed() ^
-                   PMacc::traits::GetUniqueTypeId<FrameType, uint32_t>::uid() ^
-                   IONIZATION_SEED;
-            /* makes the seed unique for each MPI rank (GPU)
-             * and each time step
-             */
-            seed = seedPerRank(seed) ^ currentStep;
-
-            const SubGrid<simDim>& subGrid = Environment<simDim>::get().SubGrid();
-            /* size of the local domain on the designated GPU in units of cells */
-            localCells = subGrid.getLocalDomain().size;
-        }
-
-        DINLINE void init(const DataSpace<simDim>& localCellIdx)
-        {
-            if (!isInitialized)
-            {
-                /* mapping the multi-dim cell index in the local domain on this GPU
-                 * to a linear index
-                 */
-                const uint32_t linearLocalCellIdx = DataSpaceOperations<simDim>::map(
-                                                                          localCells,
-                                                                          localCellIdx);
-                rng = nvrng::create(rngMethods::Xor(seed, linearLocalCellIdx), rngDistributions::Uniform_float());
-                isInitialized = true;
-            }
-        }
-
-        DINLINE float_X operator()()
-        {
-            return rng();
-        }
-
-        private:
-            typedef PMacc::nvidia::rng::RNG<rngMethods::Xor, rngDistributions::Uniform_float> RngType;
-
-            PMACC_ALIGN(rng, RngType);
-            PMACC_ALIGN(isInitialized, bool);
-            PMACC_ALIGN(seed, uint32_t);
-            PMACC_ALIGN(localCells, DataSpace<simDim>);
-    };
-
 } // namespace ionization
 
 } // namespace particles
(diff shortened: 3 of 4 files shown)


n01r:
*Very (!)* naive runtime testing of the branch: 

Ions activated, Ionization with `ADKLinPol`, 1 node, 4 NVIDIA K20 GPUs
`mpiexec -n 4 ./picongpu -d 2 2 1 -g 128 128 64 -s 500 --e_png.period 32 --e_png.axis yz --e_png.slicePoint 0.5 --e_png.folder pngElectronsYZ`

**before**
```
initialization time: 58sec 521msec = 58 sec
  0 % =        0 | time elapsed:                    0msec | avg time per step:   0msec
  5 % =       25 | time elapsed:             6sec 187msec | avg time per step: 247msec
 10 % =       50 | time elapsed:            12sec 371msec | avg time per step: 247msec
 15 % =       75 | time elapsed:            18sec 557msec | avg time per step: 247msec
 20 % =      100 | time elapsed:            24sec 742msec | avg time per step: 247msec
 25 % =      125 | time elapsed:            30sec 928msec | avg time per step: 247msec
 30 % =      150 | time elapsed:            37sec 111msec | avg time per step: 247msec
 35 % =      175 | time elapsed:            43sec 340msec | avg time per step: 249msec
 40 % =      200 | time elapsed:            49sec 798msec | avg time per step: 258msec
 45 % =      225 | time elapsed:            56sec 501msec | avg time per step: 268msec
 50 % =      250 | time elapsed:       1min  3sec 376msec | avg time per step: 274msec
 55 % =      275 | time elapsed:       1min 10sec 374msec | avg time per step: 279msec
 60 % =      300 | time elapsed:       1min 17sec 515msec | avg time per step: 285msec
 65 % =      325 | time elapsed:       1min 24sec 797msec | avg time per step: 291msec
 70 % =      350 | time elapsed:       1min 32sec 265msec | avg time per step: 298msec
 75 % =      375 | time elapsed:       1min 39sec 761msec | avg time per step: 299msec
 80 % =      400 | time elapsed:       1min 47sec 160msec | avg time per step: 295msec
 85 % =      425 | time elapsed:       1min 54sec 450msec | avg time per step: 291msec
 90 % =      450 | time elapsed:       2min  1sec 644msec | avg time per step: 287msec
 95 % =      475 | time elapsed:       2min  8sec 771msec | avg time per step: 285msec
100 % =      500 | time elapsed:       2min 15sec 834msec | avg time per step: 282msec
calculation  simulation time:  2min 15sec 834msec = 135 sec
full simulation time:  3min 14sec 759msec = 194 sec
```

**after**
```
initialization time: 58sec 208msec = 58 sec
  0 % =        0 | time elapsed:                    0msec | avg time per step:   0msec
  5 % =       25 | time elapsed:                  441msec | avg time per step:  17msec
 10 % =       50 | time elapsed:                  881msec | avg time per step:  17msec
 15 % =       75 | time elapsed:             1sec 319msec | avg time per step:  17msec
 20 % =      100 | time elapsed:             1sec 756msec | avg time per step:  17msec
 25 % =      125 | time elapsed:             2sec 192msec | avg time per step:  17msec
 30 % =      150 | time elapsed:             2sec 629msec | avg time per step:  17msec
 35 % =      175 | time elapsed:             3sec 113msec | avg time per step:  19msec
 40 % =      200 | time elapsed:             3sec 840msec | avg time per step:  29msec
 45 % =      225 | time elapsed:             4sec 816msec | avg time per step:  39msec
 50 % =      250 | time elapsed:             5sec 959msec | avg time per step:  45msec
 55 % =      275 | time elapsed:             7sec 230msec | avg time per step:  50msec
 60 % =      300 | time elapsed:             8sec 646msec | avg time per step:  56msec
 65 % =      325 | time elapsed:            10sec 194msec | avg time per step:  61msec
 70 % =      350 | time elapsed:            11sec 928msec | avg time per step:  69msec
 75 % =      375 | time elapsed:            13sec 696msec | avg time per step:  70msec
 80 % =      400 | time elapsed:            15sec 364msec | avg time per step:  66msec
 85 % =      425 | time elapsed:            16sec 919msec | avg time per step:  62msec
 90 % =      450 | time elapsed:            18sec 371msec | avg time per step:  58msec
 95 % =      475 | time elapsed:            19sec 755msec | avg time per step:  55msec
100 % =      500 | time elapsed:            21sec  81msec | avg time per step:  53msec
calculation  simulation time: 21sec  81msec = 21 sec
full simulation time:  1min 19sec 679msec = 79 sec
```

**speedup: 6.43x** (ratio of the calculation simulation times)

bussmann:
Nice! Is this the speedup you expected?

n01r:
Yes, very much so! It will change with different domain sizes and with the number of GPUs used, of course but I expected a speedup of about 5-10x. It's finally feasible to run simulations with ADK and whatever else I'll implement from now on.

PrometheusPi:
I added a **WIP** becazse it still depends on #1541.

n01r:
Waiting for #1547

PrometheusPi:
#1547 has been merged

PrometheusPi:
@n01r please:
 -  rebase against the current `dev`
 - safeguard your RNG init with a Boolean OR in `MySimulation.hpp

src/picongpu/include/traits/UsesRNG.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+/** Checks if an object requires the RNG
+ *
+ * @tparam T_Object any object (class or typename)
+ *
+ * This struct must define
+ * ::type (boost::mpl::bool_<>)
+ */
+template<typename T_Object>
+struct UsesRNG
```
n01r: This trait looks fairly generic to me. Should it rather be in PMacc?

src/libPMacc/include/particles/traits/FilterByRNG.hpp
```diff
@@ -0,0 +1,66 @@
...
+ * @tparam T_MPLSeq sequence of particle species
+ *
+ * \TODO rather pass a sequence of objects, like ionization modules
+ */
+template<typename T_MPLSeq>
+struct FilterByRNG
+{
+    typedef T_MPLSeq MPLSeq;
+
+    template<typename T_Species>
```
n01r: @psychocoderHPC 
As discussed offline:
You'd rather have a generic object passed here, right?
So that the class doesn't know anything about ionization and just checks for the requirement of an RNG. That would mean that we need a trait "ExtractFlag" which returns a list of flags and not a list of species with certain flags. Such a trait sounds fairly useful to me, though.

n01r:
@PrometheusPi there you go :wink:

PrometheusPi:
@n01r thank you

src/libPMacc/include/particles/traits/FilterByRNG.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+/** Return a new sequence of particle species that require random number
+ *  generation for their ionization.
+ *
+ * @tparam T_MPLSeq sequence of particle species
+ *
+ * \TODO rather pass a sequence of objects, like ionization modules
+ */
+template<typename T_MPLSeq>
+struct FilterByRNG
```
psychocoderHPC: The name must reflect that only species with ionizer can be handled

src/picongpu/include/traits/UsesRNG.hpp
Flamefire: Note: It is more common to have a value result defined as a member `value` rather than a subtype. This allows e.g. deriving from `bmpl::true_type which is much more compact.
In general: Type results as `type` value results as `value` and a bool certainly is a value

n01r: But isn't `boost::mpl::bool_<C>` a type? Just so that `boost::mpl::bool_<C>::value == C`.

Flamefire: This is correct. But with this trait you want a value as its return value, not a type. Or is there any benefit in returning `type<true>` instead of `true`?
Examples: `boost::make_unsigned<...>` obviously returns a type, but `boost::is_unsigned<...>` also obviously returns a bool value(!).

n01r: Is this true as well for the `HasFlag` trait, then? #1548

Flamefire: Actually yes. I think we should stick to the convention defined by boost and even the C++ standard for such traits and not define a boolean subtype but a boolean value (or inherit from a boolean type as also often done and results in the same)
It also makes checks easier. Compare: `if(typename PMacc::HasFlag<Type, Flag>::type::value)` to `if(PMacc::HasFlag<Type, Flag>::value)` and also boost meta functions work correctly with values instead of types

psychocoderHPC: The advance to use a bool type instead of `bool` value is that it can directly used as operator in any other boot mpl algorithm like `copy_if`, `transform`, ...
Therefore the more generic case is `bool type` instead of `bool value`. 

The best solution would be to use [Tag Dispatched Metafunction](http://www.boost.org/doc/libs/1_53_0/libs/mpl/doc/refmanual/tag-dispatched-metafunction.html)

Than this is allowed:
```C++
if(PMacc::HasFlag<Type, Flag>::value)
```
and also use it as unary functor
```C++
typedef typename boost::mpl::copy_if<LIST,PMacc::HasFlag<boost::mpl::_1, Flag>>::type X;
```>

Comment (Flamefire) [inline src/picongpu/include/traits/UsesRNG.hpp]: I don't see the requirement you call an advantage. Boost happily works with meta-functions returning a value instead of a type. The `meta-function class` is formed by using a placeholder, so the apply-construct has a `::type` But I was not able to find any requirement to have the function itself have a type member instead of a class. The following works fine:

    #include <boost/mpl/vector.hpp>
    #include <boost/mpl/copy_if.hpp>
    #include <boost/type_traits.hpp>

    namespace bmpl = boost::mpl;

    template<class T>
    struct Foo{ enum{value=true};};

    template<>
    struct Foo<int>{ enum{value=false};};

    typedef bmpl::vector<int, unsigned> MyVec;

    typedef bmpl::copy_if<MyVec, boost::is_unsigned<bmpl::_1> > Result;
    typedef bmpl::copy_if<MyVec, Foo<bmpl::_1> > Result2;
    typedef bmpl::copy_if<MyVec, bmpl::and_<Foo<bmpl::_1>, boost::is_unsigned<bmpl::_1> > > Result3;

    int main(){}

So IMO using a type for boolean traits just complicates things and distracts from the very nature of the trait.

Comment (ax3l) [inline src/picongpu/include/traits/UsesRNG.hpp]: @Flamefire you are not using integral `bool` types in your example but enum types which are not the same.

@psychocoderHPC is absolutely correct and if you want a `::value` to shorten the usage there is exactly the `usesRNG` function for it.

**Comment by ax3l** [comment]
transferred review powers by @PrometheusPi - thank you for the great work guys!

File: src/libPMacc/include/particles/traits/FilterByRNG.hpp (hunk 3 above)
Comment (ax3l) [inline src/libPMacc/include/particles/traits/FilterByRNG.hpp]: @n01r rename: `FilterIonsByRNG`

add to description: `@tparam T_MPLSeq sequence of particle species (ions) with ionization enabled`

Comment (ax3l) [inline src/libPMacc/include/particles/traits/FilterByRNG.hpp]: actually dependency logic is messed up in this place already.
RNG *might* be required by a species independent of ionization, too.

File: src/picongpu/include/traits/UsesRNG.hpp (hunk 1 above)
Comment (Flamefire) [inline src/picongpu/include/traits/UsesRNG.hpp]: Example updated to use a bool type AND/OR an enum type

Comment (ax3l) [inline src/picongpu/include/traits/UsesRNG.hpp]: what @Flamefire used are static const bool integrals and in the enum types which are by definition "nullary" constants (but not [nullary meta functions](http://www.boost.org/doc/libs/1_61_0/libs/mpl/doc/refmanual/metafunction.html)).

In the strict mpl context, [`boost integral constants`](http://www.boost.org/doc/libs/1_61_0/libs/type_traits/doc/html/boost_typetraits/reference/integral_constant.html) as in [`is_unsigned`](http://www.boost.org/doc/libs/1_61_0/libs/type_traits/doc/html/boost_typetraits/reference/is_unsigned.html) are not identical to [`boost::mpl integral types`](http://www.boost.org/doc/libs/1_61_0/libs/mpl/doc/refmanual/integral-constant.html) (e.g. [`mpl::bool_`](http://www.boost.org/doc/libs/1_61_0/libs/mpl/doc/refmanual/bool.html)).

For the predicate `Pred`  of [`mpl _if Transformations`](http://www.boost.org/doc/libs/1_61_0/libs/mpl/doc/refmanual/transformation-algorithms.html) a *Unary [Lambda Expression](http://www.boost.org/doc/libs/1_61_0/libs/mpl/doc/refmanual/lambda-expression.html)* is required, which is either a metafunction (not an integral constant!) or a lambda.

Nevertheless, as soon as a strict meta function is expected, e.g., in the context of [`mpl::apply`](http://www.boost.org/doc/libs/1_61_0/libs/mpl/doc/refmanual/apply.html) the missing requirement of a nullary, `::type`-able meta-function will cause trouble. The thing that saves his example here again is, that the integral constant is wrapped in a lambda which is `mpl::apply`-able.

Comment (Flamefire) [inline src/picongpu/include/traits/UsesRNG.hpp]: I stand corrected: I forgot the `...::type` in the above example which turned it into unevaluated expressions not triggering the error.
As per offline discussion with @ax3l the solution is to inherit from `bmpl::bool_<true/false>` like in the following, corrected example:

    #include <boost/mpl/vector.hpp>
    #include <boost/mpl/copy_if.hpp>
    #include <boost/mpl/if.hpp>
    #include <boost/type_traits.hpp>

    namespace bmpl = boost::mpl;

    template<class T>
    struct Foo: public bmpl::bool_<true>{};

    template<>
    struct Foo<int>: public bmpl::bool_<false>{};

    typedef bmpl::vector<int, unsigned> MyVec;

    typedef typename bmpl::copy_if<MyVec, boost::is_unsigned<bmpl::_1> >::type Result;
    typedef typename bmpl::copy_if<MyVec, Foo<bmpl::_1> >::type Result2;
    typedef typename bmpl::copy_if<MyVec, bmpl::and_<Foo<bmpl::_1>, boost::is_unsigned<bmpl::_1> > >::type Result3;
    typedef typename bmpl::if_<bmpl::and_<Foo<int>, Foo<unsigned> >, int, bool>::type Result4;

    int main(){
        if(Foo<int>::value) return 1;
        else return 0;
    }

Comment (ax3l) [inline src/picongpu/include/traits/UsesRNG.hpp]: yep that is wonderful, removes the `useRNG` run-time only function need and makes the world a pretty place.

Comment (Flamefire) [inline src/picongpu/include/traits/UsesRNG.hpp]: Note that boost `integral_constant`s **are** in fact nullary metafunctions. See their definition:

    [...]struct integral_constant : public mpl::integral_c<T, val>[...]

This is why the above use of `boost::is_unsigned` works. This is defined [here](http://www.boost.org/doc/libs/1_53_0/libs/type_traits/doc/html/boost_typetraits/reference/integral_constant.html) and also in [C++11](http://www.cplusplus.com/reference/type_traits/true_type/)

So as a matter of fact you can (and should) inherit from `boost::true_type`/`boost::false_type` which was my original statement :)

Comment (ax3l) [inline src/picongpu/include/traits/UsesRNG.hpp]: also, they are mpl compatible: [boost note](http://www.boost.org/doc/libs/1_61_0/libs/type_traits/doc/html/boost_typetraits/mpl.html)

File: src/libPMacc/include/particles/traits/FilterByRNG.hpp (hunk 3 above)
Comment (ax3l) [inline src/libPMacc/include/particles/traits/FilterByRNG.hpp]: offline result: @n01r will fix that in a follow-up PR

File: src/libPMacc/include/particles/traits/FilterByRNG.hpp (hunk 2 above)
Comment (ax3l) [inline src/libPMacc/include/particles/traits/FilterByRNG.hpp]: as comment below: shall be done in follow-up.

File: src/picongpu/include/traits/UsesRNG.hpp (hunk 1 above)
Comment (ax3l) [inline src/picongpu/include/traits/UsesRNG.hpp]: @n01r pls implement to finalise PR

File: src/picongpu/include/particles/ionization/byField/ADK/ADK_Impl.hpp [hunk 4]
```diff
@@ -83,9 +102,10 @@ namespace ionization
             /* define ionization ALGORITHM (calculation) for ionization MODEL */
             typedef T_IonizationAlgorithm IonizationAlgorithm;
 
-            /* random number generator for Monte Carlo */
-            typedef particles::ionization::RandomNrForMonteCarlo<SrcSpecies> RandomGen;
-            /* \todo fix: cannot PMACC_ALIGN() because it seems to be too large */
+            /* random number generator */
+            typedef PMacc::random::RNGProvider<simDim, PMacc::random::methods::XorMin> RNGFactory;
+            typedef PMacc::random::distributions::Uniform<float> Distribution;
```
psychocoderHPC: This should be a `float_X`!

PR #1546, opened 2016-08-09
fix RNG initialization location

Heikman:
The random number generator is now initialized only if there exist any photon species.


PrometheusPi:
@ax3l Please add this pull request to the current release.

PrometheusPi:
I checked the memory consumption - this pull request fixes it.

Flamefire:
@n01r This affects your code

PR #1547, opened 2016-08-09
move RNG memory init before particle memory init

PrometheusPi:
When allocating the RNG memory only 300 MB were free. Thus if the RNG needed a lot of memory (e.g. more than 250MB) no kernel could start. This is avoided by placing the RNG memory allocation before the particle memory allocation.

cc: @Heikman @Flamefire @ax3l 

@ax3l This affects the latest release. **Please back-port this bug fix.**

### DIFF
--- a/src/picongpu/include/simulationControl/MySimulation.hpp
+++ b/src/picongpu/include/simulationControl/MySimulation.hpp
@@ -285,6 +285,20 @@ public:
 
         laser = new LaserPhysics(cellDescription->getGridLayout());
 
+        // Initialize random number generator and synchrotron functions, if there are synchrotron photon species
+        typedef typename PMacc::particles::traits::FilterByFlag<VectorAllSpecies,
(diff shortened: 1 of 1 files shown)


PrometheusPi:
@psychocoderHPC I checked this fix without the fix of #1546 - the 250MB of memory are allocated but the simulation does not crash anymore.

PrometheusPi:
@n01r Then this affects your code as well.

psychocoderHPC:
**Short description without magic numbers:**
  - one **rule** in PIConGPU is that it is *not allowed* to allocate memory after the particles memory pool is created (memory pool allocates all free memory)
  - in the current `dev`branch the `RNGProvider`  is created after the memory pool (**not allowed**)
   -> this can results in an out of memory error

**Solution of this pull request:** move the creation of the `RNGProvider` before the particle memory pool creation

PR #1548, opened 2016-08-12
Fix documentation in HasFlag.hpp

n01r:
The specialization of HasFlag has to define a boost::mpl::bool_<>.
(Not a boost::bool_<>)


PrometheusPi:
*as offline discussed with @n01r* he thinks its a minimal change and thus will not add his name to the license header.  
I agree and thus will merge thus pull request.

PR #1549, opened 2016-08-12
Fix Sliding Window: Offset for dt * c = dy != Slide

psychocoderHPC:
- fix that window begin to move one time step too late
- fix sliding for the case delta time is equal delta cell in y

`firstSlideStep` and `firstMoveStep` (in the old implementation) were the steps where the virtual particle has already passed the border and not the time step before it will pass the border. This pull request fix it.

@axel This bug effects the realease candidate. I only see wrong slides with directional splitting. Never the less we should add this fix to the current release branch

[update:] This bug can also happen if the result of the division to calculate `firstSlideStep` and `firstMoveStep` has no fraction.

Test:

- [x] Directional splitting with moving window
- [ ] Yee Solver with moving window

### Offline result

@ax3l: this pull request solves a inconsistency with exact bounds of `t == floor(t)` between the first[Move/Slide]Step calculation and the `virtualParticlePassesGPUBorder` check 20 lines below.

### DIFF
--- a/src/picongpu/include/simulationControl/MovingWindow.hpp
+++ b/src/picongpu/include/simulationControl/MovingWindow.hpp
@@ -91,12 +91,12 @@ private:
             );
 
             /* Is the time step when the virtual particle **passed** the GPU next to the last
-             * in the current step
+             * in the current to the next step
              */
-            const uint32_t firstSlideStep = math::floor(
+            const uint32_t firstSlideStep = math::ceil(
                 float_64(subGrid.getGlobalDomain().size[moveDirection] - virtualParticleInitialStartCell) *
                 cellSizeInMoveDirection / deltaWayPerStep
-            );
+            ) - 1;
 
             /* way which the virtual particle must move before the window begins
              * to move the first time [in pic length] */
@@ -104,11 +104,11 @@ private:
                 float_64(globalWindowSizeInMoveDirection - virtualParticleInitialStartCell) *
                 cellSizeInMoveDirection;
             /* Is the time step when the virtual particle **passed** the moving window
-             * in the current step
+             * in the current to the next step
              */
-            const uint32_t firstMoveStep = math::floor(
+            const uint32_t firstMoveStep = math::ceil(
                 wayToFirstMove / deltaWayPerStep
-            );
+            ) - 1;
 
             if (firstMoveStep <= currentStep)
             {


psychocoderHPC:
The current implementation of this branch is not correct for the case that we slide much slower than speed of light.
The problem why I created this branch can also happen if ` delta time * speed of light  != cell size y`. It is triggert in any case where e.g.  `wayToFirstMove / deltaWayPerStep` has no fractional part. We need to reduce the resulting time step only in this special case. 
We can use `math::modf` to check for the fraction. I will create I correct fix next week.

psychocoderHPC:
I have implemented a version without `math::modf`.

src/picongpu/include/simulationControl/MovingWindow.hpp
```diff
@@ -91,24 +91,24 @@ class MovingWindow
...
             /* Is the time step when the virtual particle **passed** the GPU next to the last
-             * in the current step
+             * in the current to the next step
              */
-            const uint32_t firstSlideStep = math::floor(
+            const uint32_t firstSlideStep = math::ceil(
                 float_64(subGrid.getGlobalDomain().size[moveDirection] - virtualParticleInitialStartCell) *
                 cellSizeInMoveDirection / deltaWayPerStep
-            );
+            ) - 1;
```
ax3l: `- 1u` in both cases  ;)

ax3l:
the question really is how ciritcal that is: do you rather like to loose a particle at the exact box-TOP (`dev`) or a particle at the absolute-BOTTOM of your box (this PR). in both cases, one should not place important particles there.

nevertheless, changing this to prioritize the TOP is fine with me.

ax3l:
images of runs help...

here is the example run that did slide exactly a *full GPU* too early:

step 850
e_png_yx_0 5_000850

step 900
e_png_yx_0 5_000900

step 950
e_png_yx_0 5_000950

ax3l:
I don't see how this fix shall fix that problem. you are correcting exactly starting to move one step later, not a full GPU of steps later.

src/picongpu/include/simulationControl/MovingWindow.hpp
psychocoderHPC: both are wrong because `ceil` returns a double. I need a clean static_cast before

ax3l: buh, that's true!

but still, it should read `- 1u` after the cast ;)

ax3l:
comment after offline discussion added to description:
this pull request solves a inconsistency with exact bounds of t == floor(t) between the first[Move/Slide]Step calculation and the virtualParticlePassesGPUBorder check 20 lines below.

ax3l:
wonderful, thank you!

PR #1557, opened 2016-08-23
ADIOS: 1.10.0 Update

ax3l:
- we used ADIOS post-1.9.0 `master` already and can now use the latest stable release 1.10.0
- pre ADIOS 1.9.0 we had to avoid zero-writes in GPUs without
particles since ADIOS transforms did otherwise crash.
This has been fixed since ADIOS 1.9.0 on our report and
additionally a test has been added upstream in
[`tests/suite/programs/zerolength.c`](https://github.com/ornladios/ADIOS/blob/v1.9.0/tests/suite/programs/zerolength.c)

### DIFF
--- a/src/picongpu/submit/titan-ornl/picongpu.profile.example
+++ b/src/picongpu/submit/titan-ornl/picongpu.profile.example
@@ -54,10 +54,9 @@ module load scorep/2.0
 
 # plugins (optional) ################################################
 module load cray-hdf5-parallel/1.8.14
-#module load adios/1.9.0 mxml/2.9 dataspaces/1.4.0
+#module load adios/1.10.0 dataspaces/1.4.0
 export HDF5_ROOT=$HDF5_DIR
 #export ADIOS_ROOT=$ADIOS_DIR
-#export MXML_ROOT=$MXML_DIR
 #export DATASPACES_ROOT=$DATASPACES_DIR
 
 # download libSplash and compile it yourself from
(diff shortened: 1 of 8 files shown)


PrometheusPi:
looks good to me :+1:

src/picongpu/submit/titan-ornl/picongpu.profile.example
```diff
@@ -54,7 +54,7 @@ module load scorep/2.0
 
 # plugins (optional) ################################################
 module load cray-hdf5-parallel/1.8.14
-#module load adios/1.9.0 mxml/2.9 dataspaces/1.4.0
+#module load adios/1.10.0 mxml/2.9 dataspaces/1.4.0
```
PrometheusPi: is `mxml/2.9` still needed here - I thought it now ships with adios 1.10.

src/picongpu/submit/titan-ornl/pythonOnRhea.profile.example
```diff
@@ -16,7 +16,7 @@ module load python_scipy
 export IPYTHONDIR=$PROJWORK/$proj/ipython
 
 # ADIOS
-export ADIOS_ROOT=$PROJWORK/$proj/lib/adios-1.9.0-rhea
+export ADIOS_ROOT=$PROJWORK/$proj/lib/adios-1.10.0-rhea
 export LD_LIBRARY_PATH=$ADIOS_ROOT/lib:$LD_LIBRARY_PATH
 export PATH=$ADIOS_ROOT/bin:$PATH
 module load mxml
```
PrometheusPi: Same question here - is loading `mxml` still needed after aditos 1.10.?

ax3l: nope, can be removed

src/picongpu/submit/titan-ornl/picongpu.profile.example
ax3l: likely not, but I don't have access right now

PR #1556, opened 2016-08-23
INSTALL.md: Update Install hints

ax3l:
- update an outdated boost install line
- update gcc recommendation to 4.9
- C++11 hints
- adios 1.10.0 was released (ships mxml)

PR #1559, opened 2016-08-24
remove libmxml from library dependency graph

PrometheusPi:
This pull request is a followup to #1556 and #1557. It removes the `libmxml` dependency of adios from the library dependency graph.


ax3l:
thanks!
but now the png it looks ugly/chaotic.

PrometheusPi:
Yes, but that's what the hypnos version of `Graphviz` produces. The other options looks even more uglier.

ax3l:
last time I added the optional dependency hdf5->adios by hand to clean it a bit up.

ax3l:
P.S.: adios also depends on zlib

PrometheusPi:
Is this not covered by the dependency on mpi (which depends on zlib)?

ax3l:
> mpi (which depends on zlib)?

it does not in our file

PrometheusPi:
@ax3l Sorry my fault, I meant hdf5-parallel.

ax3l:
Ok, but it's still not included in the adios transports. Pls add the link :)

PrometheusPi:
@ax3l I added the zlib dependency.

ax3l:
pls use the attached image instead.
librarydependencies

PrometheusPi:
@ax3l done

PR #1566, opened 2016-09-05
Data Alignment for Kernel Parameters: 16 Byte

### LINKED ISSUE #1553: Warning: ABI change for 32-byte alignment
When compiling picongpu with CUDA 7.0 there is a warning _src/picongpu/src/picongpu/include/particles/ionization/ionization.hpp:63:1: note: The ABI for passing parameters with 32-byte alignment has changed in_
 _GCC 4.6_

related to #36 

Possible way of fixing this: passing non-native types by const reference instead of by copy.

**update:**
The warning only occurs when an ionization model is used which requires a random number generator.
It started occuring when I replaced the old version with the one written by @Flamefire. Maybe sth is not correctly aligned in the struct that contains the implementation of the ionization model. Also the order of `typedef`s and member definitions plays a role in that.

### LINKED ISSUE #1563: Align > 32 bytes
reported by @Flamefire [here](https://github.com/ComputationalRadiationPhysics/picongpu/issues/1553#issuecomment-242809705):

Just found an interesting issue while testing. Executing the following code yields in semi-random results:

``` C++
#include <curand_kernel.h>

#define __optimal_align__(byte)   \
        __align__(                \
        ((byte)==1?1:             \
        ((byte)<=2?2:             \
        ((byte)<=4?4:             \
        ((byte)<=8?8:             \
        ((byte)<=16?16:           \
        ((byte)<=32?32:           \
        ((byte)<=64?64:128        \
        ))))))))

#define PMACC_ALIGN(var,...) __optimal_align__(sizeof(__VA_ARGS__)) __VA_ARGS__ var

struct Foo{
PMACC_ALIGN(rng, curandStateXORWOW_t);
};

    template<class T_RNG, class T_Mapper>
    __global__ void
    testRNG(T_RNG rng, T_Mapper mapper)
    {
        if(!threadIdx.x)
            printf("value %i\n", mapper);
    }

int main(){
    int value=1;
    Foo foo;
    testRNG<<<1,1>>>(foo, value);
    cudaDeviceSynchronize();
}
```

This is basically a proof-of concept of a bug caused by alignment. I get the same ABI-change warning when compiling this with nvcc (7.0, 7.5) and g++4-8. In my more complex case I don't get that warning although the behaviour is the same (semi-random results, big struct somewhere)

This does not happen, if the maximum alignment is set to 32 bytes. It also does not happen, if there is another 32-byte aligned param before value. If that param is more or less aligned than 32 bytes then the bug does happen again.

psychocoderHPC:
close #1563 and close #1553

~~- add type to get a architecture depending useful alignment~~
- change `__optimal_align__` based on discussion #1563
- add pre processor macro `PMACC_ROUND_UP_NEXT_POW2`

Tests:
- [x] remove warning #1553 
- [x] KHI heating test
- [x] KHI speedtest (result: no speed decrease) 

CC-ing: @Flamefire

### DIFF
--- a/src/libPMacc/include/pmacc_types.hpp
+++ b/src/libPMacc/include/pmacc_types.hpp
@@ -25,6 +25,7 @@
 #pragma once
 
 #include "debug/PMaccVerbose.hpp"
+#include "ppFunctions.hpp"
 
 #define BOOST_MPL_LIMIT_VECTOR_SIZE 20
 #define BOOST_MPL_LIMIT_MAP_SIZE 20
@@ -172,20 +173,18 @@ enum EventType
 
 #define CUDA_CHECK_NO_EXCEP(cmd) {cudaError_t error = cmd; if(error!=cudaSuccess){ PMACC_PRINT_CUDA_ERROR(""); }}
 
-/* calculate and set the optimal alignment for data
- * you must align all array and structs which can used on device
- * @param byte byte of data which must aligned
- */
-#define __optimal_align__(byte)   \
-        __align__(                \
-        ((byte)==1?1:             \
-        ((byte)<=2?2:             \
-        ((byte)<=4?4:             \
-        ((byte)<=8?8:             \
-        ((byte)<=16?16:           \
-        ((byte)<=32?32:           \
-        ((byte)<=64?64:128        \
-        ))))))))
+/** calculate and set the optimal alignment for data
+  *
+  * you must align all arrays and structs that are used on the device
+  * @param byte size of data in bytes
+  */
+#define __optimal_align__(byte)                                                \
+    __align__(                                                                 \
+        /** \bug avoid bug if alignment is >16 byte                            \
+         * https://github.com/ComputationalRadiationPhysics/picongpu/issues/1563 \
+         */                                                                    \
+        PMACC_MIN(PMACC_ROUND_UP_NEXT_POW2(byte),16)                           \
+    )
 
 #define PMACC_ALIGN(var,...) __optimal_align__(sizeof(__VA_ARGS__)) __VA_ARGS__ var
 #define PMACC_ALIGN8(var,...) __align__(8) __VA_ARGS__ var
--- a/src/libPMacc/include/ppFunctions.hpp
+++ b/src/libPMacc/include/ppFunctions.hpp
@@ -52,3 +52,22 @@
  * @return false if no arguments are given, else true
  */
 #define PMACC_HAS_ARGS(...)  ((sizeof((int[]){0, ##__VA_ARGS__}))==sizeof(int)?false:true)
+
+/** round up to next higher pow 2 value
+ *
+ * - if value is pow2, value is returned
+ * - maximal pow 2 value is 128
+ * - negative values are not supported
+ *
+ * @param value integral number between [1,Inf]
+ * @return next higher pow 2 value
+ */
+#define PMACC_ROUND_UP_NEXT_POW2(value) \
+        ((value)==1?1:                  \
+        ((value)<=2?2:                  \
+        ((value)<=4?4:                  \
+        ((value)<=8?8:                  \
+        ((value)<=16?16:                \
+        ((value)<=32?32:                \
+        ((value)<=64?64:128             \
+        )))))))


src/libPMacc/include/ppFunctions.hpp
```diff
@@ -52,3 +52,21 @@
...
+
+/** round up to next higher pow 2 value
+ *
+ * - if value is pow2, value is returned
+ * - maximal pow 2 value is 128
+ *
+ * @param value integral number between [1,Inf]
+ * @return next higher pow 2 value
+ */
+#define PMACC_ROUND_UP_NEXT_POW2(value) \
```
Flamefire: You need to mention, that this is for UNSIGNED numbers only.

src/libPMacc/include/pmacc_types.hpp
```diff
@@ -176,16 +178,14 @@ enum EventType
...
-        ((byte)<=8?8:             \
-        ((byte)<=16?16:           \
-        ((byte)<=32?32:           \
-        ((byte)<=64?64:128        \
-        ))))))))
+#define __optimal_align__(byte)                                                \
+    __align__(                                                                 \
+        PMACC_MIN(                                                             \
+            /* 32 byte is the L2 cache line size of NVIDIA GPUs */             \
+            PMACC_MIN(PMACC_ROUND_UP_NEXT_POW2(byte),32),                      \
```
Flamefire: This will not work. 32 byte is the maximum alignment possible that is compatible between nvcc and gcc 4.3+.
So there is no change to the current, buggy behaviour for structs >32byte.
Also it probably would be better, to state, that this is the maximum vector load size (which happens to be the cache line size)

src/libPMacc/include/pmacc_types.hpp
```diff
@@ -176,16 +178,14 @@ enum EventType
  * you must align all array and structs which can used on device
  * @param byte byte of data which must aligned
```
Flamefire: You could also update this. "can be used on device" and "size of data in bytes" for better understanding

src/libPMacc/include/archAlignment.hpp
```diff
@@ -0,0 +1,42 @@
...
+namespace PMacc
+{
+    /** object to test for a useful alignment
+     *
+     * The compiler auto align the member array to a useful architecture
+     * depending value.
+     */
+    struct UsefulAlignTestObject
+    {
+        char x[512] __attribute__ ((aligned));
```
Flamefire: Does this work for other compilers? (Clang, icc)

Flamefire:
Note: I still don't see why we should limit the alignment to the maximum useful alignment on the host if the purpose of the macro is to align structs for the device.

src/libPMacc/include/ppFunctions.hpp
psychocoderHPC: thx I will extent the description

src/libPMacc/include/archAlignment.hpp
psychocoderHPC: I have not tested other compile. PMacc currently only support GCC.

- ICC is definitely not compiling
- Clang is not tested

psychocoderHPC:
The limitation to the useful alignment is added to solve #1553. 
Than we are on the save side. 32byte alignment prints the note: `The ABI for passing parameters with 32-byte alignment has changed in GCC 4.6`

Flamefire:
Ah ok, I though it was 64 bytes and up.
However I think this is still not the correct approach. We know that 32 byte alignment causes problems at gcc 4.6+. But this solution relies on the fact, that the "optimal" align chosen by the compiler is less then 32 byte or the ABI happens to still be the same. So I'd rather have 16 Bytes as the hard maximum and not introduce another dependency on GCC. Especially as icc is mentioned in https://github.com/ComputationalRadiationPhysics/picongpu/blob/release-0.2.0/INSTALL.md

src/libPMacc/include/archAlignment.hpp
ax3l: can you please still do a minimal, "external" test with those, too?

ax3l:
Why should this introduce a dependency on GCC? It's just ABI incompatibility within binaries created from various versions of GCC itself.

Flamefire:
@ax3l Because `__attribute__ ((aligned))` is GCC only(?)

ax3l:
but that has nothing to do with the wish for 16byte hard limit.

the other point (syntax) is addressed above, he will check.

src/libPMacc/include/archAlignment.hpp
ax3l: probably using the `BOOST_ALIGNMENT` macro is a portable way to do this.

btw, also a [Boost.Align](http://www.boost.org/doc/libs/1_61_0/doc/html/align.html) exists (1.56+?).

Flamefire:
I meant: This fix uses a GCC-only feature to work around an issue. Moreover it relies on a specific evaluation of that feature (namely that the optimal align is less than what the ABI was changed for).
The same workaround can be done by using a 16byte hard-limit which a) avoids both of this problems: No GCC syntax and no guessing on how that feature is evaluated.
And as mentioned: That "optimal" align is meaningless as we align for the GPU but that align (might) be optimal for CPU

psychocoderHPC:
Tested this current version with icc/15.2, gcc 4.8 and clang 3.4

```C++
#include <iostream>
#include <cstdio>
#include <boost/align/alignment_of.hpp>

#define PMACC_MIN(x,y) (((x)<=(y))?x:y)

namespace PMacc
{
    /** object to test for a useful alignment
     *
     * The compiler auto align the member array to a useful architecture
     * depending value.
     */
    struct UsefulAlignTestObject
    {
        char x[512] __attribute__ ((aligned));
    };

    /** type which defines a useful alignment for the architecture */
    typedef boost::alignment::alignment_of<UsefulAlignTestObject> useful_align_t;
}

#define PMACC_ROUND_UP_NEXT_POW2(value) \
        ((value)==1?1:                  \
        ((value)<=2?2:                  \
        ((value)<=4?4:                  \
        ((value)<=8?8:                  \
        ((value)<=16?16:                \
        ((value)<=32?32:                \
        ((value)<=64?64:128             \
        )))))))

#define __optimal_align__(byte)                                                \
    __attribute__((aligned(                                                     \
        PMACC_MIN(                                                             \
            /* 32 byte is the L2 cache line size of NVIDIA GPUs */             \
            PMACC_MIN(PMACC_ROUND_UP_NEXT_POW2(byte),32),                      \
            PMacc::useful_align_t::value                                       \
        )                                                                      \
    )))

#define PMACC_ALIGN(var,...) __optimal_align__(sizeof(__VA_ARGS__)) __VA_ARGS__ var
template<size_t N>
struct array{ char v[N];

array()
{}

array(const array& a)
{
  for(int i=0;i<N;++i)
    v[i]=a.v[i];
}
};

template<size_t T_N>
struct Foo{
static const size_t N = T_N;
PMACC_ALIGN(dummy,array<N> );
Foo()
{
for(int i=0;i<N;++i)
   dummy.v[i]=i;
}

};
template<size_t N>
struct CallKernel{
    void operator()(int value){
        CallKernel<N-1>()(value);
        Foo<N> foo;
        int size = sizeof(Foo<N>);
        printf("host %i %i\n", N,size);
    }
};
template<>
struct CallKernel<0>
{
    void operator()(int){}
};

int main(){
    int value=1;
    CallKernel<128>()(value);
}
```

Flamefire:
Ok so this is not GCC-only syntax? Then only remaining question is the maximum alignment safely supported for CUDA params. (see #1553) (besides that I still see no use for host-optimal alignment for GPUs)

psychocoderHPC:
Normally we need two alignment options one for fitting a object on stack (as parameter) and one for dynamic objects (e.g. Frames).
If we know that a object is never passed copy by value to a function than we can go to higher alignment up to 128 byte. 

Side note: Currently I am testing if 16 byte influenced the speed.

Flamefire:
I don't think it is that easy. First: Alignment introduces overhead. The positions e.g. are not stored as SoA but as AoS, so with the current alignment a bunch of threads will load 4 values/thread but will only use 3 of them. With 4/8 byte alignment the L2 cache should be used much better (using the model: all threads load x -> cacheline is loaded with x,y,z,unused for each element, all load y (from cache), all load z (from cache), but quite some cache is wasted as all those unused values had to be loaded, of course a 16/32 byte vector load might be possible with aligned data)
Even worse for the 44 byte random structs which get aligned to 64 bytes. This wastes bandwith and memory especially with increasing alignment.
Second: Vector loads are 32 byte max. So why align more?

psychocoderHPC:
Why align to more than 32byte: Because 128bytes is read out of the global RAM of a GPU with each memory access.

psychocoderHPC:
I tested the 16byte alignment vs the old and I cant see any speedup or slowdown. It looks like we are save with 16byte.

psychocoderHPC:
Heating Electron-Positron:
heat_posel

psychocoderHPC:
**offline discussion with @axel:**
- currently we cut our alignment to 16byte ( this avoid's bug #1563)
- this pull request is only a bugfix
  - we can tune our alignment for each type depending of the usage in a separate pull request, if needed
  - with C++11 we can check each kernel parameter to full fill the alignment at compile time

psychocoderHPC:
I removed the calculation of the useful alignment from this pull request.

src/libPMacc/include/pmacc_types.hpp
ax3l:
```
/* calculate and set the optimal alignment for data
  *
  * you must align all arrays and structs that are used on the device
  * @param byte size of data in bytes
*/
```

ax3l: actually we should check if the `must align` is now actually a `should align for performance`
( @psychocoderHPC  remembers that in sm_13 non-aligned data did cause bugs)

Flamefire:
@ax3l Please comment to latest changes. Can't see you updates ;)
Re "must vs should align" see my comment: https://github.com/ComputationalRadiationPhysics/picongpu/pull/1566#issuecomment-244768114
If our arrays are uniform (only floats or doubles) at least in size AND mallocMC returns aligned base offsets then it is a "should". For non-uniform it could be a "must" and for non-aligned base it IS a "must"

ax3l:
I just helped him integrate a comment of yours he forgot.
See here: https://github.com/ComputationalRadiationPhysics/picongpu/pull/1566#discussion-diff-77500487

ax3l:
thank you both for looking into the problem!

PR #893, opened 2015-05-27
Rename TwistVectorAxes -> TwistComponents

Heikman:
Rename RT version of TwistVectorAxes to TwistComponents

add Functor: math::tools::CT::TwistVectorAxes

PR #1568, opened 2016-09-07
extent function `twistVectorFieldAxes<>()`

psychocoderHPC:
Allow different permutation for navigation in space and the accessor to the vector field component.
- add two new function signatures
  - permutate navigator and accessor with the same permutation vector
  - permute navigator and accessor different
- update documentation
- change template parameter nameing

This pull request is needed for the directional splitting changes
- [x] please not merge before the release 0.2.0 is finsihed

### DIFF
--- a/src/libPMacc/include/cuSTL/cursor/tools/twistVectorFieldAxes.hpp
+++ b/src/libPMacc/include/cuSTL/cursor/tools/twistVectorFieldAxes.hpp
@@ -36,12 +36,18 @@ namespace tools
 namespace result_of
 {
 
-template<typename Axes, typename TCursor>
+/** result for TwistVectorFieldAxes
+ *
+ * \tparam T_NavigatorPerm permutation vector for navigator
+ * \tparam T_AccessorPerm permutation vector for the accessor
+ * \tparam T_Cursor cursor to permute
+ */
+template<typename T_NavigatorPerm, typename T_AccessorPerm, typename T_Cursor>
 struct TwistVectorFieldAxes
 {
-    typedef Cursor<TwistAxesAccessor<TCursor, Axes>,
-                   PMacc::cursor::CT::TwistAxesNavigator<Axes>,
-                   TCursor> type;
+    typedef Cursor<TwistAxesAccessor<T_Cursor, T_AccessorPerm>,
+                   PMacc::cursor::CT::TwistAxesNavigator<T_NavigatorPerm>,
+                   T_Cursor> type;
 };
 
 } // result_of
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/cuSTL/cursor/tools/twistVectorFieldAxes.hpp
```diff
@@ -55,18 +61,59 @@ struct TwistVectorFieldAxes
...
  * e.g.: new_cur = twistVectorFieldAxes<math::CT::Int<1,2,0> >(cur); // x -> y, y -> z, z -> x
  *
- * \tparam Axes compile-time vector (PMacc::math::CT::Int) that describes the mapping.
- * x-axis -> Axes::at<0>, y-axis -> Axes::at<1>, ...
+ * \tparam T_Permutation compile-time vector (PMacc::math::CT::Int) that describes the mapping.
+ * x-axis -> T_Permutation::at<0>, y-axis -> T_Permutation::at<1>, ...
+ *
+ */
+template<typename T_Permutation, typename T_Cursor>
+HDINLINE
```
Heikman: please keep the order of template arguments (see below).

Heikman:
code looks fine but if the permutation of the navigator and the accessor can possibly be different, then the name `twistVectorFieldAxes` is wrong. `vector field` is a physical term which implies that the permutation is equal. Please find a more technical name, see e.g. #893.

src/libPMacc/include/cuSTL/cursor/tools/twistVectorFieldAxes.hpp
psychocoderHPC: The order is changed that it is not allowed to call this method like `twistVectorFieldAxes<CT::Int<1,1> >(cursor, permutationWithWrongType)`
The order for the case that you need to give the permutation as compile time signature is still the same as the old.

src/libPMacc/include/cuSTL/cursor/tools/twistVectorFieldAxes.hpp
```diff
@@ -55,18 +61,59 @@ struct TwistVectorFieldAxes
...
+ * \param accessorPermutation compile time permutation vector for the accessor
  */
-template<typename Axes, typename TCursor>
+template<typename T_NavigatorPerm, typename T_Cursor, typename T_AccessorPerm>
 HDINLINE
-typename result_of::TwistVectorFieldAxes<Axes, TCursor>::type
-twistVectorFieldAxes(const TCursor& cursor)
+typename result_of::TwistVectorFieldAxes<T_NavigatorPerm, T_AccessorPerm, T_Cursor>::type
+twistVectorFieldAxes(const T_Cursor& cursor,
+                       const T_NavigatorPerm& /*navigatorPermutation*/,
```
psychocoderHPC: Do you mean that I should change this order to `twistVectorFieldAxes( cursor, accessorPermutation, navigatorPermutation )` instead of the current order `twistVectorFieldAxes( cursor, navigatorPermutation, accessorPermutation )`

Heikman: No, it's just that the order in line 68 is the opposite to the one in line 88.

psychocoderHPC: Yes this is to avoid user mistakes [ descriped in comment]( https://github.com/ComputationalRadiationPhysics/picongpu/pull/1568/files#r78394231)

psychocoderHPC: I see that line [107](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1568/files#diff-f4b543b557290758e842bcdcc7c9d21bR107) needs to be changed to a void the same problem.

psychocoderHPC:
I moved `T_Cursor` as first parameter `twistVectorFieldAxes` to avoid wrong usage.

We should change the naming in a separate  pull request.
My suggestion:
  - rename `twistAxes` -> `permuteNavigator`
  - rename `twistVectorFieldAxes` -> `permuteNavigatorAndAccessor`

Both methods are tools for the cursor but not represent any explicit physics meaning. I think that's the good way because a cursor is only a vector field if space and component dimension is equal. But there is no need for this restriction.

psychocoderHPC:
ping

Heikman:
code looks fine to me, proposed renaming too. I'll merge now.

PR #948, opened 2015-07-21
define species mass/charge relative to a base

psychocoderHPC:
- change param and unitless files
  - add new file speciesConstants.param/unitless
  - move `GAMMA_*THRESH` to `speciesConstants.param`
  - add SI::BASE_MASS_SI/SI::BASE_CHARGE_SI
  - add BASE_MASS/BASE_CHARGE
  - `speciesDefinition.unitless` add default method to calculate mass and charge
  - `speciesAttributes.param`:
    - add alias `massRatio`
    - add alias `chargeRatio`
- remove usage of Q_EL with BASE_CHARGE
- remove usage of M_EL with BASE_MASS

PR #1571, opened 2016-09-09
fix unit mass/charge calculation

psychocoderHPC:
derive unit mass and charge from `BASE_MASS_SI` and `BASE_CHARGE_SI`

This change was missed in #948. The result is that if the user change the base mass/charge the units are might be over/under flowing.

**This fix should be part of the release 0.2.0**


src/picongpu/include/simulation_defines/unitless/physicalConstants.unitless
```diff
@@ -65,9 +65,9 @@ namespace picongpu
 
 
     /** Unit of mass */
-    BOOST_CONSTEXPR_OR_CONST float_64 UNIT_MASS = SI::ELECTRON_MASS_SI * double(particles::TYPICAL_NUM_PARTICLES_PER_MACROPARTICLE);
+    BOOST_CONSTEXPR_OR_CONST float_64 UNIT_MASS = SI::BASE_MASS_SI * double(particles::TYPICAL_NUM_PARTICLES_PER_MACROPARTICLE);
     /** Unit of charge */
-    BOOST_CONSTEXPR_OR_CONST float_64 UNIT_CHARGE = -1.0 * SI::ELECTRON_CHARGE_SI * double(particles::TYPICAL_NUM_PARTICLES_PER_MACROPARTICLE);
```
psychocoderHPC: @axel: I think if we derive from a base value than we need to skip the `-1.0` in front. This is only in because of the fact that we typically used electrons as base.

ax3l: typo in meaning

> we need

no need, one *can* skip.


Our `BASE_CHARGE_SI` (speciesConstants.param) *is* the `ELECTRON_CHARGE_SI` (physicalConstants.param) which is currently defined negative (so both are) and this does not change with this PR.

So in principle you are right, but if we change it we will cause "raw charge densities" of ions to be negative. That's counter-intuitive. Please keep the `-1.0` therefore, electrons will so far stay our species one.

ax3l:
That's rather a convenience update and will not deviate much in reality. I will mark this as a refactoring, there is not real risk over over/underflow.

PR #1575, opened 2016-09-12
fix wrong 2D `FieldJ` unit

### LINKED ISSUE #1572: FieldJ `unitSI` in 2D is wrong
The (unitSi)[https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/fields/FieldJ.tpp#L242] inside of our HDF5 files for FieldJ x and y component is wrong by a factor of `UNIT_LENGTH`

In 2D we calculate inside of Esirkepov not the surface current (for J_x and J_y) but a line current but we not take care of it while the unit calculation.

@abukva and me currently checking the current

psychocoderHPC:
fix #1572 

calculate unit for the field `FieldJ` depending on the dimension of
the simulation
- [ ] this fix needs to be checked again if #1569 is merged

### DIFF
--- a/src/picongpu/include/fields/FieldJ.tpp
+++ b/src/picongpu/include/fields/FieldJ.tpp
@@ -239,7 +239,14 @@ HDINLINE
 FieldJ::UnitValueType
 FieldJ::getUnit( )
 {
-    const float_64 UNIT_CURRENT = UNIT_CHARGE / UNIT_TIME / ( UNIT_LENGTH * UNIT_LENGTH );
+    float_64 lengthFactor( 1.0 );
+    /** the current is a
+     * - surface density in 3D
+     * - line density in 2D
+     */
+    for( uint32_t d = 0; d < simDim - 1u;  ++d)
+        lengthFactor *= UNIT_LENGTH;
+    const float_64 UNIT_CURRENT = UNIT_CHARGE / UNIT_TIME / ( lengthFactor );
     return UnitValueType( UNIT_CURRENT, UNIT_CURRENT, UNIT_CURRENT );
 }
 


src/picongpu/include/fields/FieldJ.tpp
```diff
@@ -239,7 +239,14 @@ HDINLINE
 FieldJ::UnitValueType
 FieldJ::getUnit( )
 {
-    const float_64 UNIT_CURRENT = UNIT_CHARGE / UNIT_TIME / ( UNIT_LENGTH * UNIT_LENGTH );
+    float_64 lengthFactor( 1.0 );
+    /** the current is a
+     * - surface density in 3D
+     * - line density in 2D
+     */
+    for( uint32_t d = 0; d < simDim - 1u;  ++d)
```
ax3l: I see your fix but it is very obfuscated.

current in z in 2D3V is still a "per plane". the fix is actually about the definition of the charge.

src/picongpu/include/fields/FieldJ.tpp
```diff
@@ -239,7 +239,14 @@ HDINLINE
 FieldJ::UnitValueType
 FieldJ::getUnit( )
 {
-    const float_64 UNIT_CURRENT = UNIT_CHARGE / UNIT_TIME / ( UNIT_LENGTH * UNIT_LENGTH );
+    float_64 lengthFactor( 1.0 );
+    /** the current is a
```
ax3l: do be updated: a current is not a density.

psychocoderHPC:
I checked  the unit for the `FieldJ` in the original `dev` branch for 2D with our `SingleParticleTest` and it look correct.
I used one electron in the single particle test, I am not sure if we ignore the defined density in this case and there the unit is current.
Maybe the `unitSI` of `FieldJ` is correct and this is an issue with the `UNIT_CHARGE`

2D test, electron is moving with 0.5 beta in x, y and z direction.
2d_oneparcurrent

src/picongpu/include/fields/FieldJ.tpp
PrometheusPi: but it is not a current (I) either - it is a current density (J)

psychocoderHPC:
fixed with #1601

Issue #654, opened 2015-01-29
collect information after crashes


psychocoderHPC:
After a cash we need more information to  find the reason.

In the most cases it is very hard or not possible to reproduce crashes. In such case it would be useful to have more information about the simulation state before the crash.

possible information sources:
- [create core dumps](https://wiki.archlinux.org/index.php/Core_dump)
- [create a backtrace](http://stackoverflow.com/questions/77005/how-to-generate-a-stacktrace-when-my-gcc-c-app-crashes)
- create a crashdump with usefull information
   - all registered transaction in PMacc
   - current waiting `Tasks` 
   - the state of all singletons in `PMacc` und `PIConGPU`

psychocoderHPC:
A possible solution to create a crash dump is to use [boost serialization](http://www.boost.org/doc/libs/1_57_0/libs/serialization/doc/index.html) (IMO: not really what we need )

I am working on the smart `crash dump` system which can collect the state of all importend classes in `PMacc` and `PIConGPU` while the crash.

PR #1578, opened 2016-09-14
expression validation/assertion

psychocoderHPC:
Many assertion checks inside PMacc and PIConGPU were performed with `assert`. 
If PIConGPU/PMACC is compiled in release mode these checks were disabled.
In the most cases this checks are essential checks to validate the user input parameters and should always be evaluated.
This pull request adds PMacc specific asserts and verification macros to choose if a expression needs to be evaluated only in debug mode or also inside the release binary.
- add function `abortWithError()` to create an exception with useful information
- add macro function
  - PMACC_ASSERT
  - PMACC_ASSERT_MSG
  - PMACC_VERIFY
  - PMACC_VERIFY_MSG

**Why we are not using`BOOST_ASSERT, BOOST_VERIFY?`**

The boost assert and verify functions ended up with `std::abort()`. This needs signal handling to implement something like #654.
After we added signal handling in PMacc we can easily switch to the boost macros without changing our asserts in the code.

TODO:
- [x] test new macros

### DIFF
--- a/src/libPMacc/include/assert.hpp
+++ b/src/libPMacc/include/assert.hpp
@@ -1,0 +1,61 @@
+/**
+ * Copyright 2016 Rene Widera
+ *
+ * This file is part of libPMacc.
+ *
+ * libPMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * libPMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with libPMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#pragma once
+
+#include "debug/abortWithError.hpp"
+
+#ifdef NDEBUG
+    // debug mode is disabled
+
+    /* `(void)0` force a semicolon after the macro function */
+#   define PMACC_ASSERT( expr ) ( (void) 0 )
+
+    /* `(void)0` force a semicolon after the macro function */
+#   define PMACC_ASSERT_MSG( expr, msg ) ( (void) 0 )
+
+#else
+
+    // debug mode is enabled
+
+    /** assert check
+     *
+     * if `NDEBUG` is not defined: macro expands to (void)0
+     *
+     * @param expr expression to be evaluated
+     */
+#   define PMACC_ASSERT( expr )                                                \
+    ( !!(expr) ) ? ( (void) 0 ) : PMacc::abortWithError( #expr, __FILE__, __LINE__ )
+
+    /** assert check with message
+     *
+     * if `NDEBUG` is not defined: macro expands to (void)0
+     *
+     * @param expr expression to be evaluated
+     * @param msg output message (of type `std::string`) which is printed if the
+     *            expression is evaluated to false
+     */
+#   define PMACC_ASSERT_MSG( expr, msg )                                       \
+    ( !!(expr) ) ? ( (void) 0 ) : PMacc::abortWithError( #expr, __FILE__, __LINE__, msg )
+
+#endif
--- a/src/libPMacc/include/debug/abortWithError.hpp
+++ b/src/libPMacc/include/debug/abortWithError.hpp
@@ -1,0 +1,59 @@
+/**
+ * Copyright 2016 Rene Widera
+ *
+ * This file is part of libPMacc.
+ *
+ * libPMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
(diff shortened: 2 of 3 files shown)


src/libPMacc/include/assert.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#pragma once
+
+#include "debug/abortWithError.hpp"
+
+#ifdef NDEBUG
+// debug mode is enabled
```
ax3l: exactly the other way around: http://en.cppreference.com/w/c/error/assert

"If `NDEBUG` is defined as a macro name at the point in the source code where `<assert.h>` is included, then *assert does nothing*."

src/libPMacc/include/assert.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+/** assert check
```
ax3l: pls align accordingly above `define`

src/libPMacc/include/verify.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include "debug/abortWithError.hpp"
+
+/** verify expression
+ *
+ * Same behavior as PMACC_ASSERT but the expression is always evaluated.
+ *
+ * @param expr expression to be evaluated
+ */
+#   define PMACC_VERIFY( expr )                                                \
```
ax3l: pls do not align (it's not in an #if level)

src/libPMacc/include/assert.hpp
psychocoderHPC: Well spotted , never trust my code

src/libPMacc/include/debug/abortWithError.hpp
```diff
@@ -0,0 +1,59 @@
...
+ * and the GNU Lesser General Public License along with libPMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <iostream>
+#include <string>
+#include <stdexcept>
+#include <stringstream>
```
psychocoderHPC: must be `#include <sstream>`

src/libPMacc/include/debug/abortWithError.hpp
```diff
@@ -0,0 +1,59 @@
...
+    )
+    {
+        std::stringstream line;
+        line<<lineNumber;
+
+        throw std::runtime_error(
+            "expression (" +
+            exp +
+            ") fails in file (" +
+            filename + ":" + line + "): " +
```
psychocoderHPC: `line` needs to be a string `line.str()`

src/libPMacc/include/debug/abortWithError.hpp
```diff
@@ -0,0 +1,59 @@
...
+     */
+    void abortWithError(
+        const std::string exp,
+        const std::string filename,
+        const uint32_t lineNumber,
+        const std::string msg = std::string()
+    )
+    {
+        std::stringstream line;
+        line<<lineNumber;
```
psychocoderHPC: missing spaces around `<<`

src/libPMacc/include/verify.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include "debug/abortWithError.hpp"
+
+/** verify expression
+ *
+ * Same behavior as PMACC_ASSERT but the expression is always evaluated.
+ *
+ * @param expr expression to be evaluated
+ */
+#define PMACC_VERIFY( expr )                                                   \
+    (!(expr)) ? ((void)0) : PMacc::abortWithError( #expr, __FILE__, __LINE__)
```
psychocoderHPC: missing space before the last closing bracket

src/libPMacc/include/assert.hpp
```diff
@@ -0,0 +1,61 @@
...
+    // debug mode is enabled
+
+    /** assert check
+     *
+     * if `NDEBUG` is not defined: macro expands to (void)0
+     *
+     * @param expr expression to be evaluated
+     */
+#   define PMACC_ASSERT( expr )                                                \
+    (!(expr)) ? ((void)0) : PMacc::abortWithError( #expr, __FILE__, __LINE__)
```
psychocoderHPC: missing space before closing bracket

psychocoderHPC:
Assert looks like this:
```
  what():  expression (initController != NULL) fails in file (/bigdata/hplsim/scratch/widera/dev/src/picongpu/include/simulationControl/MySimulation.hpp:599):
```

src/libPMacc/include/assert.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+    /** assert check with message
+     *
+     * if `NDEBUG` is not defined: macro expands to (void)0
+     *
+     * @param expr expression to be evaluated
+     * @param msg output message (of type `std::string`) which is printed if the
```
psychocoderHPC: We can keep C++ strings and if we need to match BOOST_ASSERT_MSG we can convert the c++ strings.

psychocoderHPC:
@ComputationalRadiationPhysics/picongpu-maintainers please review me

src/libPMacc/include/debug/abortWithError.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <iostream>
+#include <string>
+#include <stdexcept>
+#include <sstream>
+
+namespace PMacc
+{
+    /** abort program with an exception
+     *
+     * This function always throw a `runtime_error`.
```
PrometheusPi: throw**s**

psychocoderHPC:
rebased and updated

PR #1581, opened 2016-09-26
particles exchange: avoid message spamming

psychocoderHPC:
- remove retry message in `TaskReceiveParticlesExchange`
- aggregate retry message in `TaskSendParticlesExchange`

offline request from: @Flamefire

### DIFF
--- a/src/libPMacc/include/particles/tasks/TaskSendParticlesExchange.hpp
+++ b/src/libPMacc/include/particles/tasks/TaskSendParticlesExchange.hpp
@@ -44,7 +44,7 @@ namespace PMacc
         state(Constructor),
         maxSize(parBase.getParticlesBuffer().getSendExchangeStack(exchange).getMaxParticlesCount()),
         initDependency(__getTransactionEvent()),
-        lastSize(0),lastSendEvent(EventTask()){ }
+        lastSize(0),lastSendEvent(EventTask()),retryCounter(0){ }
 
         virtual void init()
         {
@@ -86,7 +86,7 @@ namespace PMacc
                         //check for next bash round
                         if (lastSize == maxSize)
                         {
-                            std::cerr<<"send max size "<<maxSize<<" particles"<<std::endl;
+                            ++retryCounter;
                             init(); //call init and run a full send cycle
 
                         }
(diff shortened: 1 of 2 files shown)


src/libPMacc/include/particles/tasks/TaskSendParticlesExchange.hpp
```diff
@@ -113,6 +113,15 @@ namespace PMacc
         virtual ~TaskSendParticlesExchange()
         {
             notify(this->myId, RECVFINISHED, NULL);
+            if(retryCounter != 0)
+            {
+                std::cerr << "Send buffer for particle species " <<
+                    ParBase::FrameType::getName() <<
+                    " is to small (max particles per send operation:" << maxSize <<
+                    ", direction:" << exchange <<
+                    ", retried:" << retryCounter <<
```
PrometheusPi: Could you add spaces after the colons so that the values do not collide with the associated names.

Flamefire: I'd also shorten the message as much as possible while you are on it. So repeated output doesn't produce that much file size (Requested that due to 13GB text file sizes...)
E.g.: "Sendbuffer for species X to small (Max. particles: Y, Dir: Z, Retries: Q)"

psychocoderHPC: I added the spaces but not shortened all words. IMO there is no reason the cut all words.

Flamefire: Well my reasoning was, that if you get this, you'll get this a lot. So for 300k timesteps you may get this e.g. 3600k times. So saving 10 bytes in the string will save 36MB output and it is still enough to understand the error. But no strong opinion here as this already helps a lot

src/libPMacc/include/particles/tasks/TaskSendParticlesExchange.hpp
```diff
@@ -113,6 +113,15 @@ namespace PMacc
         virtual ~TaskSendParticlesExchange()
         {
             notify(this->myId, RECVFINISHED, NULL);
+            if(retryCounter != 0)
+            {
+                std::cerr << "Send/receive buffer for species " <<
```
ax3l: well, we would need std:cerr writing LogLevels and a default lvl for PMacc, so not for now...

PR #1587, opened 2016-10-06
Fix warpBroadcast

### LINKED ISSUE #1583: broken IdProvider
@Heikman found out that the generation of id's for particles is not working.

I checked the HDF5 files and can reproduce the bug.
I checked `ParticlesInit.kernel` and found that attribute `particlesId` is never initialized with the call `IdProvider<simDim>::getNewId()`

Flamefire:
This fixes #1583.
The warpBroadcast for 64 bit types was broken. The unsigned call resulted in an infinite loop (for some reason this never resulted in any errors) and the signed call did not use the return values. Also using unions the way they were used before is undefined behaviour. So this was replaced by pointer operations.


Heikman:
Particle IDs are all different now. Tested with bunch setup.

PR #1589, opened 2016-10-07
Improve PMacc testsystem

Flamefire:
This fixes some issues with the PMacc test system, so it can be used for more tests.

Core:

- Fix cuda warnings (lots of output)
- One exectuable per testcase and dividing into 2D and 3D tests (singletons make it impossible to test this in one program)
- Move some common code into extra files (MPI-init and PMacc init)

Note: This triggers a CMake bug, that may prevent parallel `make` builds due to wrong dependencies between the test object files. See http://stackoverflow.com/questions/39913202/adding-multiple-cuda-executables-leaks-dependencies

### DIFF
--- a/src/libPMacc/CMakeLists.txt
+++ b/src/libPMacc/CMakeLists.txt
@@ -1,5 +1,5 @@
 #
-# Copyright 2015-2016 Erik Zenker
+# Copyright 2015-2016 Erik Zenker, Alexander Grund
 #
 # This file is part of libPMacc.
 #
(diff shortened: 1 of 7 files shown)


src/libPMacc/CMakeLists.txt
```diff
@@ -23,11 +23,26 @@
 ################################################################################
 # PMacc tests
 ################################################################################
-cmake_minimum_required(VERSION 3.3)
-project ("PMacc-tests")
+cmake_minimum_required(VERSION 3.1)
```
psychocoderHPC: Why you are changing this requirement? 

I think `CMAKE_CXX_STANDARD` was introduced in  cmake 3.1.3

src/libPMacc/PMaccConfig.cmake
```diff
@@ -29,8 +29,7 @@
 ###############################################################################
 # PMacc
 ###############################################################################
-cmake_minimum_required(VERSION 3.3)
-project("PMacc")
+cmake_minimum_required(VERSION 3.1)
```
psychocoderHPC: please revert this change

src/libPMacc/CMakeLists.txt
Flamefire: This is now the same as PIConGPU. I don't see a reason, why we should use a higher requirement here. `CMAKE_CXX_STANDARD` is not an issue, as the correct value is ensured (the same way as in PIConGPU).
The higher versions just makes it harder to use this with a default-configured system.

src/libPMacc/PMaccConfig.cmake
Flamefire: See above. Same version as PIConGPU and the `project` is just plain wrong here. This is a `*Config.cmake` file should not contain that but simply provide paths for `FindModule`

src/libPMacc/CMakeLists.txt
psychocoderHPC: Changing requirements hidden inside a pull request which solves an other issue should avoided.

Flamefire: Then please merge this first: https://github.com/ComputationalRadiationPhysics/picongpu/pull/1591

psychocoderHPC:
please rebase

Flamefire:
Rebased

PR #1593, opened 2016-10-12
ForEach always use const references

### LINKED ISSUE #848: Warning host/device function call
On taurus/zih I get the following warning:

```
Warning: calling a __host__ function("std::_List_base<picongpu::ISimulationPlugin *, std::allocator<picongpu::ISimulationPlugin *> > ::~_List_base")
  from a __host__ __device__ function("std::_List_base<picongpu::ISimulationPlugin *, std::allocator<picongpu::ISimulationPlugin *> > ::~_List_base [subobject]")
  is not allowed
```

I am using:
- `gcc/4.6.2`
- `boost/1.54.0`
- `cuda/5.0.35`

psychocoderHPC:
- use function parameters always as const reference
- fix wrong documenation
- fix warning `host function called from host/device`

close #848

This pull request should be back ported to the current release candidate.

### DIFF
--- a/src/libPMacc/include/algorithms/ForEach.hpp
+++ b/src/libPMacc/include/algorithms/ForEach.hpp
@@ -70,11 +70,14 @@ namespace PMacc
 namespace algorithms
 {
 
+/** macro creates  typeNameN & */
+#define PMACC_UNUSED_REF(_, n, typeName) typeName##n &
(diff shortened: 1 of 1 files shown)


ax3l:
great catch, thank you! :sparkles:

PR #1595, opened 2016-10-14
Add typedefs for Host/DeviceBuffer

Flamefire:
Often getting the individual buffers is required. For that one might need to know the type which are now defined.
No behaviour changes

### DIFF
--- a/src/libPMacc/include/memory/buffers/HostDeviceBuffer.hpp
+++ b/src/libPMacc/include/memory/buffers/HostDeviceBuffer.hpp
@@ -36,6 +36,8 @@ namespace PMacc {
         typedef HostBufferIntern<T_Type, T_dim> HostBufferType;
         typedef DeviceBufferIntern<T_Type, T_dim> DeviceBufferType;
     public:
+        typedef HostBuffer<T_Type, T_dim> HostBuffer;
+        typedef DeviceBuffer<T_Type, T_dim> DeviceBuffer;
         typedef typename HostBufferType::DataBoxType DataBoxType;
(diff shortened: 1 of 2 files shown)


src/libPMacc/include/memory/buffers/HostDeviceBuffer.hpp
```diff
@@ -36,6 +36,8 @@ namespace PMacc {
         typedef HostBufferIntern<T_Type, T_dim> HostBufferType;
         typedef DeviceBufferIntern<T_Type, T_dim> DeviceBufferType;
     public:
+        typedef HostBuffer<T_Type, T_dim> HostBuffer;
```
psychocoderHPC: Why not add `DeviceBufferType` to the public space?

Flamefire: Because `HostBufferIntern` is the implementation and `HostBuffer` is the "interface". Only the latter should be exposed to the user.

Flamefire:
@psychocoderHPC Can this be merged?

psychocoderHPC:
There is still an open question. If the new definition hides the pmacc definition.

Flamefire:
Didn't see the question...?
Yes it hides the PMacc definition, but only inside the scope of the class which should be ok. If you'd want to use the PMacc class (template) you'd need to use `PMacc::HostBuffer<foo, bar>`. But I think this is unlikely and the current solution is better in terms of usability then renaming the typedef to e.g. `HostBufferType` or so. It is not possible to accidentally use the wrong class as one of them is a template and the other is not. So it would result in a compile failure (see 2nd commit)

PR #1596, opened 2016-10-14
Fix constructor for HDBuffer

Flamefire:
A HostBuffer can never be casted to a HostDeviceBuffer so that CTor would fail to compile when it is used. Removed the cast.


ax3l:
applied in d5f14b6

PR #1599, opened 2016-10-24
fix boost 1.62  boost greatest common divisor usage 

### LINKED ISSUE #1598: Boost 1.62.0 bug with `gcd_evaluator`
The call `boost::math::gcd_evaluator<size_t> gcd;` creates an error with boost 1.62.0
```
picongpu/src/picongpu/../libPMacc/include/cuSTL/algorithm/kernel/run-time/F
oreach.hpp(104): error: qualified name is not allowed
          detected during:
```

psychocoderHPC:
close  #1598

- fix boost 1.62.0 compile error
  -> reason:  `gcd_evaluator` is moved out of the namespace `boost::math`
- use `boost::math::gcd()` function

The definition of `gcd_evaluator` is not longer located in the namespace `boost::math`.
There is a definition in `boost::integer` but IMO the more clean solution is to use the function `boost::math::gcd()`.
`boost::math::gcd()` select at compile time the best method to calculate the greatest common divisor.

### DIFF
--- a/src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
+++ b/src/libPMacc/include/cuSTL/algorithm/kernel/run-time/Foreach.hpp
@@ -38,7 +38,7 @@
 #include <boost/preprocessor/arithmetic/inc.hpp>
 #include <boost/preprocessor/repetition/repeat.hpp>
 #include <boost/preprocessor/repetition/repeat_from_to.hpp>
-#include <boost/math/common_factor.hpp>
+#include <boost/math/common_factor_rt.hpp>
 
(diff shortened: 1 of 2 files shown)


ax3l:
line 41 `#include <boost/math/common_factor.hpp>`

we are now using boost::math (RT) already anyway, so let us switch to the explicit header using the RT here, too.

[Ref.:](http://www.boost.org/doc/libs/1_62_0/libs/math/doc/html/math_toolkit/gcd_header.html):
> **Note**  This is a legacy header: it used to contain the actual implementation, but the compile-time and run-time facilities were moved to separate headers (since they were independent of each other).

Use ` <boost/math/common_factor_rt.hpp>` instead, we are including the additional runtime already.

PR #1600, opened 2016-10-25
Hypnos & Bash Tpl: Return Code Cleanup

ax3l:
The return code of hypnos-hzdr and bash templates for `tbg` are determined by an additional `killall` cleanup line.

This line creates noise on *successful* runs in case no process was found to be killed. This noise is now suppressed.


ax3l:
applied in 57c6e5a

Issue #240, opened 2014-02-28
Energy Scaling for 2D and 3D looks different


psychocoderHPC:
The total energy for KHI 2D and 3D is not equal.
(if normalized to the respective number of cells again)

This bug could come from `physicalConstants.unitless:33`. 
For 2D NUM_EL_PER_PARTICLE is scaled with `SI::SPEED_OF_LIGHT_SI*SI::DELTA_T_SI` so `omega_pe` is equal in 2D and 3D.

psychocoderHPC:
closed with #1601  -> energy in 2D and 3D is equal with respect to the number of cells in Z direction

PR #1601, opened 2016-10-25
2D3V Unit Normalization: Use dZ

ax3l:
This changes the way we normalize 2D3V simulations by the following, common way: evaluate dZ (choose something in the same order such as dX and dY).

A 2D3V simulation in a cartesian PIC simulation such as ours only changes the degrees of freedom in motion for (macro) particles and all (field) information in z travels instantaneous, making the 2D3V simulation behave like the interaction of infinite "wire particles" in fields with perfect symmetry in Z.

Quantities such as `CELL_VOLUME` are now _always 3D_.

This replaces #1569, should fix #1575 and #240.
- [x] needs _intense_ RT testing

### DIFF
--- a/src/picongpu/include/simulation_defines/param/gridConfig.param
+++ b/src/picongpu/include/simulation_defines/param/gridConfig.param
@@ -30,15 +30,30 @@ namespace picongpu
         /** Duration of one timestep
          *  unit: seconds */
         BOOST_CONSTEXPR_OR_CONST float_64 DELTA_T_SI = 0.8e-16;
+
         /** equals X
          *  unit: meter */
         BOOST_CONSTEXPR_OR_CONST float_64 CELL_WIDTH_SI = 0.1772e-6;
-        /** equals Y - the propagation direction
+        /** equals Y - the laser & moving window propagation direction
          *  unit: meter */
         BOOST_CONSTEXPR_OR_CONST float_64 CELL_HEIGHT_SI = 0.4430e-7;
         /** equals Z
          *  unit: meter */
         BOOST_CONSTEXPR_OR_CONST float_64 CELL_DEPTH_SI = CELL_WIDTH_SI;
+
+        /** Note on units in reduced dimensions
+         *
+         * In 2D3V simulations, the CELL_DEPTH_SI (Z) cell length
+         * is still used for normalization of densities, etc.
+         *
+         * A 2D3V simulation in a cartesian PIC simulation such as
+         * ours only changes the degrees of freedom in motion for
+         * (macro) particles and all (field) information in z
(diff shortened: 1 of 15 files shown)


src/picongpu/include/simulation_defines/param/gridConfig.param
```diff
@@ -30,15 +30,30 @@ namespace picongpu
         /** Duration of one timestep
          *  unit: seconds */
         BOOST_CONSTEXPR_OR_CONST float_64 DELTA_T_SI = 0.8e-16;
+
         /** equals X
          *  unit: meter */
         BOOST_CONSTEXPR_OR_CONST float_64 CELL_WIDTH_SI = 0.1772e-6;
-        /** equals Y - the propagation direction
+        /** equals Y - the laser & moving window propagation direction
```
psychocoderHPC: There is no laser by default, could we remove `laser` from the description?

psychocoderHPC:
We need remove the 2D part in  [Visualisation.hpp](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/plugins/output/images/Visualisation.hpp#L215-L227)

src/picongpu/include/simulation_defines/unitless/physicalConstants.unitless
```diff
@@ -30,38 +30,14 @@ namespace picongpu
     /** Unit of length */
     BOOST_CONSTEXPR_OR_CONST float_64 UNIT_LENGTH = UNIT_TIME*UNIT_SPEED;
 
-#if(SIMDIM==DIM3)
-    namespace SI
-    {
-        /** density normed to dimension of the simulation*/
-        BOOST_CONSTEXPR_OR_CONST float_64  GAS_DENSITY_NORMED= SI::GAS_DENSITY_SI;
```
psychocoderHPC: `GAS_DENSITY_NORMED` is needed by [gasConfig.unitless](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1601/files#diff-236b4e3b436ad7ebb3cf9a6d32d7fefeR30)

psychocoderHPC:
I compared 2d KHI electrons/positrons with particles moving in Z direction. 
The global energy in the system was 12x higher in 3D than in 2D. The 3D simulation used 12 cells in Z direction.
This means this pull request passed the test.

src/picongpu/include/simulation_defines/param/gridConfig.param
ax3l: put in this general file to point users to what the currently chosen propagation axis for settings in `laserConfig.param` is.

PR #1602, opened 2016-10-25
fix cmake c++11 enable section

psychocoderHPC:
`CUDA_NVCC_FLAG` is defined as CACHE, therefore all assignments before were ignored and not given to `nvcc`.
`CUDA_NVCC_FLAG` needs a list of semicolon separated flags.

- move c++11 enabler after `find_package(CUDA ..)`
- add `-std=c++11` separated by semicolon instead of space


PrometheusPi:
:+1: great job - I will check whether this fixes the c++11 issues seen on taurus #1603

PR #1604, opened 2016-10-25
Specialize HasFlag and GetFlagType for Particle

Flamefire:
Currently you can call the `HasIdentifier` trait also for the Particle type which is very useful in some occasions. This adds the counterparts `HasFlag` and `GetFlagType` so they can be used there as well.

Minor fixes not worth mentioning:
- Move boost includes to other boost includes
- Make the `HasIdentifier` look the same as the new traits (much shorter)

### DIFF
--- a/src/libPMacc/include/particles/memory/dataTypes/Particle.hpp
+++ b/src/libPMacc/include/particles/memory/dataTypes/Particle.hpp
@@ -24,10 +24,9 @@
 
 #include "pmacc_types.hpp"
 #include "particles/boostExtension/InheritLinearly.hpp"
-#include <boost/utility/result_of.hpp>
-#include <boost/type_traits.hpp>
-#include <boost/mpl/if.hpp>
 #include "traits/HasIdentifier.hpp"
+#include "traits/HasFlag.hpp"
+#include "traits/GetFlagType.hpp"
 #include "compileTime/GetKeyFromAlias.hpp"
 #include "compileTime/conversion/ResolveAliases.hpp"
 #include "compileTime/conversion/RemoveFromSeq.hpp"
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/particles/memory/dataTypes/Particle.hpp
```diff
@@ -139,20 +141,21 @@ struct Particle : public InheritLinearly<typename T_FrameType::MethodsList>
...
->
-struct HasIdentifier<
-PMacc::Particle<T_FrameType>,
-T_Key
->
-{
-private:
-    typedef T_FrameType FrameType;
-public:
-    typedef typename HasIdentifier<FrameType, T_Key>::type type;
```
psychocoderHPC: It look like the old implementation of HasIdentifier for a particle is wrong implemented. If HasIdentifier<deselect<Particle,X>,X> is called it returns true but should return false if the identifier is already deselected.
I will check this tomorrow again.

Flamefire: Offline discussion: `HasIdentifier<deselect<Particle,X>,X>` would not compile in old code because 2nd template param was missing. This PR would introduce this bug.
Semantic should be: `HasIdentifier<Particle, X> == true` `<=>` `particle[X_]` is valid

psychocoderHPC:
Please use #1607 as base for your changes.

src/libPMacc/include/particles/memory/dataTypes/Particle.hpp
psychocoderHPC: fixed in #1607

Flamefire:
Rebased

PR #1606, opened 2016-10-26
Add generic atomicAdd

Flamefire:
This is intended for a generic use of the atomicAdd function. There is also a `atomicAddWrapper` in a private namespace but that does only work with float/double and does not return the old value.
This is also ready for the new atomicAdd for doubles in CUDA 8 / CC 6.x


psychocoderHPC:
Thanks for updating !

PR #1609, opened 2016-10-26
fix tbg variable substitution

### LINKED ISSUE #695: TBG: removes variable definition
`tbg` removes variable definitions in the `*.tpl` file.

e.g. example.tpl:
```bash
myVariable="value"

runMyProgram $myVariable
```
is changed to:
```bash
runMyProgram $myVariable
```

psychocoderHPC:
close #695

- delete command that removed variable definitions `someThing=...` from the template file
- `*.tpl` files: move batch system commands above (the now still existing) bash commands

**Reason for `*.tpl` change:**
Batch system commands must be defined before any bash command. Before this pull request `tbg` deleted all variable definitions in the template file and therefore fulfilled the batch system restriction. 

Removing variable assignments (and due to that defined variables in `.start`) with `tbg` results in side effects and limited functionality for the template file (since those runtime variables will be missing in the `.start` script).

**Old `*.tpl` files are not compatible to this pull request**

### DIFF
--- a/src/picongpu/submit/davinci-rice/picongpu.tpl
+++ b/src/picongpu/submit/davinci-rice/picongpu.tpl
@@ -1,32 +1,25 @@
 #!/usr/bin/env bash
 # Copyright 2013-2016 Axel Huebl, Rene Widera, Richard Pausch
-# 
-# This file is part of PIConGPU. 
-# 
-# PIConGPU is free software: you can redistribute it and/or modify 
-# it under the terms of the GNU General Public License as published by 
-# the Free Software Foundation, either version 3 of the License, or 
-# (at your option) any later version. 
-# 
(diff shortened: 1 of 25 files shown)


psychocoderHPC:
This bug fix need that we change all `*.tpl` files before the first batch system command e.g. `#PBS -q queueName` it is not allowed to place any bash commands.

psychocoderHPC:
updated this pull request

psychocoderHPC:
UPDATE: I have not seen in my tests that this pull request now creates error's during the job start. This is not critical but should be solved.

psychocoderHPC:
@ComputationalRadiationPhysics/picongpu-developers This pull request breaks the old template files. Please use the new `*.tpl` files!

PR #1611, opened 2016-10-26
fix tbg evaluation error

psychocoderHPC:
ignore errors during the cfg and tpl file evaluation

example:
```
# tpl file

x=$(basename !TBG_VARIABLE)
```
This example crashed before this pull request do to the fact that *all* variable definitions (`something=...`) were evaluated and then replaced.

The pull request ignored errors during the evaluation. (IMO not critical)

CC-ing: @PrometheusPi

### DIFF
--- a/src/tools/bin/tbg
+++ b/src/tools/bin/tbg
@@ -69,7 +69,7 @@ function run_cfg_and_get_solved_variables
     #append template file variable definitions and solve them
     while read -r data_set
     do
-        eval "$data_set"
+        eval "$data_set" &> /dev/null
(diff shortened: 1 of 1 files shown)


ax3l:
looks a little bit brutal, can't we change it to `only substitute !tbgvar=` assignments or do we rely on general evaluation?

ax3l:
related to #695

ax3l:
just for the record: that is just ... crazy... :tada: 

we are now able to define a variable `x` which can be
 - undefined at tpl evaluation time
 - defined at start time

or
 - defined to 1 at evaluation time
 - and changed to 2 at star time

or
 - ...

@psychocoderHPC : "they are quantum variables!"

PR #1617, opened 2016-11-01
CUDA 7.5+: C++11 Autoswitch

ax3l:
In CUDA 7.5+ we automatically switch to C++11. We allow no fall-back to C++98 via `CXX_STANDARD_REQUIRED`.

Via `CMAKE_CXX_EXTENSIONS` we enforce `-std` as `c++11` and no gnu extensions, simply because the CMake `FindCUDA` scripts are not flexible enough.

In case of the combination with boost 1.60.0 the user is requested to update boost or just use a newer CUDA version which will auto-enable C++11.

Tested with cmake 3.3.0 and CUDA 6.5, 7.0, 8.0. `make VERBOSE=True` revealed that `--std=c++11` is automatically passed to `nvcc` itself.

**Drop official support for:**
- CUDA 5.0 
- GCC <4.8
- CMake <3.3 (needed for [CUDA C++11](https://github.com/Kitware/CMake/blob/v3.3.0/Modules/FindCUDA.cmake#L1271-L1282))

Update compile suite to:
- CUDA  7.5.18 -> C++11
- GCC 4.8.5
- Boost 1.57
- [CMake 3.3+](https://github.com/Kitware/CMake/blob/v3.3.0/Modules/FindCUDA.cmake#L1271-L1282)

**References:**

```bash
$ nvcc --help
# ...
--std {c++11}                              (-std)
        Select a particular C++ dialect.  Note that this flag also turns on the corresponding
        dialect flag for the host compiler.
        Allowed values for this option:  'c++11'.
# ...
```

https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD_REQUIRED.html

https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html

https://cmake.org/cmake/help/v3.3/variable/CMAKE_CXX_EXTENSIONS.html#variable:CMAKE_CXX_EXTENSIONS


INSTALL.md
```diff
@@ -19,31 +19,26 @@ Requirements
...
   - *Arch Linux:* `sudo pacman --sync cuda`
 - C++11 build: [CUDA 7.5+](https://developer.nvidia.com/cuda-downloads)
-  - `export PIC_COMPILE_SUITE_CMAKE="-DCMAKE_CXX_STANDARD=11"`
-  - `export CPPFLAGS="-std=c++11"`
-  - if you run `$PICSRC/configure` below, add `-c"-DCMAKE_CXX_STANDARD=11"` to it
+  - note: we recommend using **gcc 4.9**
   - *Arch Linux:* `sudo pacman --sync cuda`
 
 - at least one **CUDA** capable **GPU**
   - *Compute capability* **sm\_20** or higher
```
psychocoderHPC: Could we add a note that we mark sm_20 as deprecated? In CUDA 8 sm_20 is also deprecated.

psychocoderHPC: offline discussed: we mark this deprecated with the next release

psychocoderHPC:
ready for cherry-pick

PR #1618, opened 2016-11-01
fix tbg runtime errors and param `-o`

psychocoderHPC:
- fix runtime errors introduced with #1609
- revert dirty evaluation from #1611
- fix tbg `-o` parameter
- add new syntax to define variables in the tpl files `.varname=value`
- update all tpl files with new syntax: `.var=value` 

Documentation
============

We define two life times for variables.
  -  **tbg-time**: this means a variable is visible during the time `tbg` creates the start file `submit.start` for the batch system. it contains several steps:
      1. substitute variables defined with `tbg -o` in the `*.cfg` file (overwrite assignments `varName=value`)
      2. evaluate the `*.cfg` file
      3. substitute variables defined with `tbg -o` in the `*.tpl` file (overwrite **only** assignments `.varName=value`)
      4. catch all variables from the current environment and substitute all `!varName` usages
      5. remove all `.varname=value` from the `*.tpl` file
      6. substitude all `!varName` in tpl file with the results of the step v.
  - **run-time**: is the time where the job is started from the batch system and the file `submit.start` is executed

variables
------------
  - each placeholder in the form `!varName` is replaced by the corresponding variable value
  - assignments of tbg variables must begin with `.` e.g. `.myTplVar=5` (syntax is only allowed within a template file) -> named **tplVar**

`*.cfg` file
--------
 - is a bash script and is executed at **tbg-time**
 - the syntax follows the restrictions of the bash
 - variables with the accessor `!varName` are not replaced
 - all assigned variables can be used for substitutions in the tpl file
 - each variable assignment in the form `varName=value` is overwritten by variables given by the `tbg` parameter `-o` before the cfg file is executed

`*.tpl` file
--------
 - is used to create `submit.start`
 - variable assignments with **tplVar** are executed/evaluated at **tbg-time**
 - **tplVar**, environment variables and variables from the cfg file can be used with the accessor `!varName`
 - **tplVar** are overwritten by variables given with `-o` to `tbg`
 - **tplVar** are removed from the `submit.start`

`submit.start`
----------------
 - is created by `tbg` out of the `*.tpl` file and the substituted variables
 - it is not allowed to have any bash code before the batch script commands, e.g. `#PBS ...`
 - all variable definitions and assignments are evaluated at **run-time** (note: **tplVar** are removed at **tbg-time**)

### DIFF
--- a/src/picongpu/submit/davinci-rice/picongpu.tpl
+++ b/src/picongpu/submit/davinci-rice/picongpu.tpl
@@ -43,9 +43,9 @@
 ## calculation are done by tbg ##
 
 # settings that can be controlled by environment variables before submit
-TBG_mailSettings=${MY_MAILNOTIFY:-"n"}
-TBG_mailAddress=${MY_MAIL:-"someone@example.com"}
-TBG_author=${MY_NAME:+--author \"${MY_NAME}\"}"
+.TBG_mailSettings=${MY_MAILNOTIFY:-"n"}
+.TBG_mailAddress=${MY_MAIL:-"someone@example.com"}
+.TBG_author=${MY_NAME:+--author \"${MY_NAME}\"}"
 
 ## end tbg calculation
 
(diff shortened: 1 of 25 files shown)


ax3l:
all right, thanks! :sparkles:

Issue #1128, opened 2015-09-18
ZigZag Charge Conservation


ax3l:
The status of our implementation of ZigZag for current deposition is, that it does not yet conserve charge to the high amount one is used to from Esirkepov which is our default (down to floating point precision; ZigZag: just in the 0.01 to 0.1% range for KHI tests).

We are evaluating the consequences and if this is due to limitations in the scheme itself or due to an implementation bug.

Related to #1102 and #1084.
### Scope

Affects all previous releases, `dev` and both 3D3V and 2D3V simulations.
### Expected consequences
- laser penetrates plasmas above normal depths (observed)
- slight heating (not yet observed in KHI test)
- divergences and emittances could be overestimated
- stronger noise
- scattered fake fields of "scattered resting charges" that influence dynamics
- triggering of small wavelength plasma instabilties (artificial seeds)
- possible wrong growth rates of plasma instabilities
- domain specific problems need evaluation: e.g., could influence LWFA injection dynamics
- ...

ax3l:
Comparison of `dev` as of 96dd7326d46bfc9886231242269e314f74aa0dd1 with our default KHI example (p+/e-) on 2x8x1 GPUs on 192x512x12 cells with sm_37 using TSC (cmake case `0` and `5`) at t = 1000 steps:

```bash
# default slices in middle
plot_chargeConservation.py khi-esirkepov/simOutput/h5_data_1000.h5
```

### Esirkepov (Default)

![khi_esirkepov_1000](https://cloud.githubusercontent.com/assets/1353258/9962169/b1fb0160-5e24-11e5-9101-196b2f9de340.png)


### ZigZag

![khi_zigzag_1000](https://cloud.githubusercontent.com/assets/1353258/9962176/b818cc80-5e24-11e5-8229-bdd0cf13d939.png)

### Comparison

Take a look at the range of the colorbars. As we can clearly see in both the scale and the image of Esirkepov, we can follow the clean float noise. One can even see the float super-cell structure from the local reduces (additions).

On the other hand, ZigZag has four orders of magnitude less precision in the results. The data itself is still good for this very specific case but overly noisy and structured.

psychocoderHPC:
I added the native paper version of CIC in my branch https://github.com/psychocoderHPC/picongpu/commit/ecc51642fe74a9fc61d7d1c4a7f9ce6b0ca25ce7
[...]

PR #1622, opened 2016-11-02
ZigZag: Update input Docs (Do not Use)

ax3l:
ZigZag in our implementation, and probably even in the original publications, is not charge conserving #1128.

Move it in the input files down to "development" solvers so users do not use it.

### DIFF
--- a/examples/Bunch/include/simulation_defines/param/species.param
+++ b/examples/Bunch/include/simulation_defines/param/species.param
@@ -57,11 +57,11 @@ typedef FieldToParticleInterpolation<UsedParticleShape, AssignedTrilinearInterpo
 /*! select current solver method -----------------------------------------------
  * - currentSolver::Esirkepov<SHAPE>  : particle shapes - CIC, TSC, PCS, P4S (1st to 4th order)
  * - currentSolver::VillaBune<>       : particle shapes - CIC (1st order) only
- * - currentSolver::ZigZag<SHAPE>     : particle shapes - CIC, TSC, PCS , P4S (1st to 4th order)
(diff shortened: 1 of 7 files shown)


examples/Bunch/include/simulation_defines/param/species.param
```diff
@@ -57,11 +57,11 @@ typedef FieldToParticleInterpolation<UsedParticleShape, AssignedTrilinearInterpo
 /*! select current solver method -----------------------------------------------
  * - currentSolver::Esirkepov<SHAPE>  : particle shapes - CIC, TSC, PCS, P4S (1st to 4th order)
  * - currentSolver::VillaBune<>       : particle shapes - CIC (1st order) only
- * - currentSolver::ZigZag<SHAPE>     : particle shapes - CIC, TSC, PCS , P4S (1st to 4th order)
  *
  * For development purposes: ---------------------------------------------------
  * - currentSolver::EsirkepovNative<SHAPE> : generic version of currentSolverEsirkepov
  *   without optimization (~4x slower and needs more shared memory)
+ * - currentSolver::ZigZag<SHAPE>     : particle shapes - CIC, TSC, PCS , P4S (1st to 4th order)
```
PrometheusPi: do you want to align the colon with `currentSolver::EsirkepovNative<SHAPE> :`?

ax3l: bit too random I think

PrometheusPi: OK - looks good to me :+1:

PR #1627, opened 2016-11-04
Fix #1625: HDF5 openPMD positionOffset

### LINKED ISSUE #1625: HDF5 Particle Patches: Offset broken (openPMD)
The openPMD particles need an adjustment to address the following violation of the openPMD standard.

- particlePatch: offset wrong
- positionOffset: adjust to total position, not global position (moving window)

ax3l:
This fixes #1625 by creating the correct `positionOffset` in HDF5 writes+restarts which is a *total cellIdx*.

The methods to create and convert the localCellIdx to a cellIdx respectively to certain domain have been abstracted. The PMacc user can now choose to create an arbitrary new cell idx, e.g., a `globalCellIdx` with respect to the global domain as before or a `totalCellIdx` with respect to an absolute start. For PIConGPU, we nowadays always speak in positions with respect to the total domain for all inputs and outputs which is more useful then a globalCellIdx which is rather arbitrarily choosen to be t*c but might not be that physically useful, especially with respect to tracking. This also reduces a lot of code offset calc complexity.

The ADIOS plugin has not yet been updates, simply because it lacks particle patches still anyway and temporarily redefining the `particle_info` object is not a great idea. For ADIOS, just be aware we are still dumping a `globalCellIdx` in the openPMD `positionOffset` when doing moving window simulations #1628.

RT tested with LWFA example on 1x4x1 GPUs including restarts after 5 different checkpoints along several slides. Reading and projecting particles above fields now also works with our new yt project reader.

### DIFF
--- a/src/libPMacc/include/particles/operations/ConcatListOfFrames.hpp
+++ b/src/libPMacc/include/particles/operations/ConcatListOfFrames.hpp
@@ -39,8 +39,7 @@ namespace operations
  *   super-cell on the GPU to a single frame on the CPU RAM
  * - the deep on-GPU hierarchy must be copied to the CPU beforehand
  * - remove species attributes `multiMask` and `localCellIdx`
- * - add new attribute `globalCellIdx` (particle offset to begin of global
- *   moving window)
+ * - add new cellIdx attribute relative to a user-defined domain
  */
 template<unsigned T_dim>
 struct ConcatListOfFrames
@@ -59,12 +58,24 @@ struct ConcatListOfFrames
      * @param destFrame single frame were all particles are copied in
      * @param srcBox particle box were particles are read from
      * @param particleFilter filter to select particles
-     * @param localOffset Offset of the local domain, used to calculate globalCellIdx,
-     *                    can be negative for the first GPU: localDomain.offset - globalWindow.offset
+     * @param domainOffset offset to a user-defined domain. Can, e.g. be used to
+     *                     calculate a totalCellIdx, adding
+     *                     globalDomain.offset + localDomain.offset
+     * @param domainCellIdxIdentifier the identifier for the particle cellIdx
+     *                                that is calculated with respect to
+     *                                domainOffset
      * @param mapper mapper which describes the area where particles are copied from
      */
-    template<class T_DestFrame, class T_SrcBox, class T_Filter, class T_Space, class T_Mapping>
-    void operator()(int& counter, T_DestFrame destFrame, T_SrcBox srcBox, T_Filter particleFilter, T_Space localOffset, T_Mapping mapper)
+    template<class T_DestFrame, class T_SrcBox, class T_Filter, class T_Space, class T_Identifier, class T_Mapping>
+    void operator()(
+        int& counter,
+        T_DestFrame destFrame,
+        T_SrcBox srcBox,
+        const T_Filter particleFilter,
+        const T_Space domainOffset,
+        const T_Identifier domainCellIdxIdentifier,
+        const T_Mapping mapper
+    )
     {
         #pragma omp parallel for
         for (int linearBlockIdx = 0;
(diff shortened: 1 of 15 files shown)


src/libPMacc/include/particles/operations/ConcatListOfFrames.hpp
```diff
@@ -59,12 +58,24 @@ struct ConcatListOfFrames
      * @param destFrame single frame were all particles are copied in
      * @param srcBox particle box were particles are read from
      * @param particleFilter filter to select particles
-     * @param localOffset Offset of the local domain, used to calculate globalCellIdx,
-     *                    can be negative for the first GPU: localDomain.offset - globalWindow.offset
+     * @param domainOffset offset to a user-defined domain. Can, e.g. be used to
+     *                     calculate a totalCellIdx relative to
```
psychocoderHPC: maybe better /*... calculate a totalCellIdx out of globalDomain.offset + localDomain.offset*/

src/picongpu/include/fields/background/cellwiseOperation.hpp
```diff
@@ -38,7 +38,7 @@ namespace cellwiseOperation
 
     /** Kernel that calls T_OpFunctor and T_ValFunctor on each cell of a field
      *
-     *  Pseudo code: opFunctor( cell, valFunctor( globalCellIdx, currentStep ) );
+     *  Pseudo code: opFunctor( cell, valFunctor( totalCellIdx, currentStep ) );
```
psychocoderHPC: Please revert or give a better name. This offset must be the offset to the origin of the global domain.

src/picongpu/include/plugins/hdf5/HDF5Writer.hpp
```diff
@@ -391,10 +391,9 @@ class HDF5Writer : public ISimulationPlugin
     {
         ThreadParams *threadParams = (ThreadParams*) (p_args);
         const PMacc::Selection<simDim>& localDomain = Environment<simDim>::get().SubGrid().getLocalDomain();
+        const PMacc::Selection<simDim>& globalDomain = Environment<simDim>::get().SubGrid().getGlobalDomain();
```
psychocoderHPC: Please join this and the line above
```
const SubGrid& subGrid = Environment<simDim>::get().SubGrid();
DataSpace<simDim> domainOffset(
    subGrid.getGlobalDomain().offset + 
    subGrid.getLocallDomain().offset
);
```

src/picongpu/include/plugins/hdf5/WriteSpecies.hpp
```diff
@@ -346,13 +350,14 @@ struct WriteSpecies
          *         global domain offsets (slides), etc.
          * extent: size of this particle patch, upper bound is excluded
          */
+        const PMacc::Selection<simDim>& globalDomain = Environment<simDim>::get().SubGrid().getGlobalDomain();
```
psychocoderHPC: Note: If we use domain offsets those are not coming from the `param`  object than we can introduce side effects.
Currently we get the all domain definitions from `params`

src/picongpu/include/fields/background/cellwiseOperation.hpp
ax3l: I don't think so...? I implemented it so it can be used with a total position for background fields.

src/libPMacc/include/particles/operations/ConcatListOfFrames.hpp
ax3l: or *adding* instead of *out of* ...?

src/picongpu/include/plugins/hdf5/HDF5Writer.hpp
ax3l: :+1:

src/picongpu/include/fields/background/cellwiseOperation.hpp
psychocoderHPC: Yes you are right

src/picongpu/include/plugins/hdf5/HDF5Writer.hpp
```diff
@@ -391,10 +391,13 @@ class HDF5Writer : public ISimulationPlugin
...
```
psychocoderHPC: This and the next line needs to be removed.

PR #1634, opened 2016-11-10
fix hdf5 gas profile

psychocoderHPC:
local variable definition shadows class member variable


ax3l:
thank you!

weird, we both think we fixed this before... can't find a PR...

PR #1640, opened 2016-11-11
remove namespace workaround

psychocoderHPC:
- delete explicit call to `PMacc::algorithms::math`
- use short cut `math::*`

The explicit definition of the namespace is a workaround for #1338 and solved with #1472.
This pull request is a cleanup.

### DIFF
--- a/src/picongpu/include/plugins/radiation/amplitude.hpp
+++ b/src/picongpu/include/plugins/radiation/amplitude.hpp
@@ -44,9 +44,9 @@ public:
       picongpu::float_X cosValue;
       picongpu::float_X sinValue;
       picongpu::math::sincos(phase, sinValue, cosValue);
-      amp_x=PMacc::algorithms::math::euler(vec.x(), picongpu::precisionCast<picongpu::float_64>(sinValue), picongpu::precisionCast<picongpu::float_64>(cosValue) );
-      amp_y=PMacc::algorithms::math::euler(vec.y(), picongpu::precisionCast<picongpu::float_64>(sinValue), picongpu::precisionCast<picongpu::float_64>(cosValue) );
-      amp_z=PMacc::algorithms::math::euler(vec.z(), picongpu::precisionCast<picongpu::float_64>(sinValue), picongpu::precisionCast<picongpu::float_64>(cosValue) );
+      amp_x=picongpu::math::euler(vec.x(), picongpu::precisionCast<picongpu::float_64>(sinValue), picongpu::precisionCast<picongpu::float_64>(cosValue) );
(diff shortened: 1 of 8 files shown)


src/picongpu/include/plugins/radiation/amplitude.hpp
```diff
@@ -44,9 +44,9 @@ class Amplitude
       picongpu::float_X cosValue;
       picongpu::float_X sinValue;
       picongpu::math::sincos(phase, sinValue, cosValue);
-      amp_x=PMacc::algorithms::math::euler(vec.x(), picongpu::precisionCast<picongpu::float_64>(sinValue), picongpu::precisionCast<picongpu::float_64>(cosValue) );
-      amp_y=PMacc::algorithms::math::euler(vec.y(), picongpu::precisionCast<picongpu::float_64>(sinValue), picongpu::precisionCast<picongpu::float_64>(cosValue) );
-      amp_z=PMacc::algorithms::math::euler(vec.z(), picongpu::precisionCast<picongpu::float_64>(sinValue), picongpu::precisionCast<picongpu::float_64>(cosValue) );
+      amp_x=math::euler(vec.x(), picongpu::precisionCast<picongpu::float_64>(sinValue), picongpu::precisionCast<picongpu::float_64>(cosValue) );
+      amp_y=math::euler(vec.y(), picongpu::precisionCast<picongpu::float_64>(sinValue), picongpu::precisionCast<picongpu::float_64>(cosValue) );
+      amp_z=math::euler(vec.z(), picongpu::precisionCast<picongpu::float_64>(sinValue), picongpu::precisionCast<picongpu::float_64>(cosValue) );
```
PrometheusPi: please add `picongpu::math::euler` here

psychocoderHPC: Ahh thx this class is not in the picongpu namespace.

PrometheusPi:
You are not part of the author list in 5/8 of the files - @psychocoderHPC please think about adding your name (only if you want).

ax3l:
@PrometheusPi please read the wikipedia article on threshold of originality (German copyright: [Schöpfungshöhe](https://de.wikipedia.org/wiki/Sch%C3%B6pfungsh%C3%B6he)). A copy&replace is usually not that case, and if it is we have git. He is doing everything correctly.

ax3l:
Why do we name `PMacc::algorithms::math` short `pmMath`? It is not `PMacc::math`... aww :(

ax3l:
this PR reintroduced the bug in #1552, I will add a compile test for it now

PR #1645, opened 2016-11-21
fix 2 dimensional `cellSize` definition

psychocoderHPC:
`cellSize` does not use the cell depth in 2d.

follow-up to: #1601
note: no code-correctness implications so far, we always use `CELL_VOLUME` in-code and no `cellSize.productOfComponents()`, etc.


psychocoderHPC:
@BeyondEspresso please take a look to the TWTS changes. Keep in mind that PIConGPU in 2d had now also a cell depth. My solution is to remove the cell depth for TWTS in 2D.

BeyondEspresso:
The changes are fine!

psychocoderHPC:
we need to solve #1650 to pass the compile suite

src/picongpu/include/simulation_defines/param/gasConfig.param
```diff
@@ -249,7 +249,7 @@ struct FreeFormulaFunctor
      *
      * @return float_X density [normalized to 1.0]
      */
-    HDINLINE float_X operator()(const floatD_64& position_SI, const floatD_64& cellSize_SI)
+    HDINLINE float_X operator()(const floatD_64& position_SI, const float3_64& cellSize_SI)
```
ax3l: just a note: `cellSize_SI` is a debug parameter and now in different dimensionality so that even `.shrink<simDim>()` would be needed for usage, so we might remove or refactor that in the future since I don't like confusing the user with it.

PR #1651, opened 2016-11-22
fix adios 1.11.0 compile

### LINKED ISSUE #1650: adios 1.11.0 broken
PIConGPU with  with adios 1.11.0 enabled does not compile.

Error:
```
/home/buildbot/picongpu-src/src/picongpu/include/plugins/adios/ADIOSWriter.hpp(965): error: identifier "adios_stat_no" is undefined
```

The interface of
```C++
int adios_declare_group ( int64_t * id ,
const char * name ,
const char * time_index ,
enum ADIOS_FLAG stats )
```
is changed to 

```C++
int adios_declare_group ( int64_t * id ,
const char * name ,
const char * time_index ,
enum ADIOS_STATISTICS_FLAG stats )
```

~~This bug is triggered the first time by our compile suite inside the issue #1645.~~ [update:] compile suite used adios 1.10.0 the error in #1645 is triggered by a wrong change

psychocoderHPC:
remove `adios_flag_no` with integer value `2`

The interface of `adios_declare_group` is changed in adios version 1.11.0, see #1650

close #1650

### DIFF
--- a/src/picongpu/include/plugins/adios/ADIOSWriter.hpp
+++ b/src/picongpu/include/plugins/adios/ADIOSWriter.hpp
@@ -961,11 +961,21 @@ private:
         DataSpace<simDim> particleOffset(localDomain.offset);
         particleOffset.y() -= threadParams->window.globalDimensions.offset.y();
 
+        /* ADIOS API change:
+         * <= 1.10.0: `enum ADIOS_FLAG`
+         * >= 1.11.0: `enum ADIOS_STATISTICS_FLAG`
+         */
+#if ( ( ADIOS_VERSION_MAJOR * 100 + ADIOS_VERSION_MINOR ) >= 111 )
+        ADIOS_STATISTICS_FLAG noStatistics = adios_stat_no;
+#else
+        ADIOS_FLAG noStatistics = adios_flag_no;
+#endif
+
         /* create adios group for fields without statistics */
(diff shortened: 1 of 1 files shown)


src/picongpu/include/plugins/adios/ADIOSWriter.hpp
```diff
@@ -961,11 +961,21 @@ class ADIOSWriter : public ILightweightPlugin
         DataSpace<simDim> particleOffset(localDomain.offset);
         particleOffset.y() -= threadParams->window.globalDimensions.offset.y();
 
+        /* The interface of `adios_declare_group()` is changed from adios 1.10.0 to 1.11.0.
+         *
+         * Before 1.11.0 we need to use the enum `ADIOS_FLAG` else `ADIOS_STATISTICS_FLAG`
```
ax3l: shorter and more precise:
```C++
/* ADIOS API change:
 * <= 1.10.0: `enum ADIOS_FLAG`
 * >= 1.11.0: `enum ADIOS_STATISTICS_FLAG`
 */
```

ax3l:
thx for the quick fix! I would just condense the inline comment a bit more to the point

ax3l:
note (out of scope of this PR): generally, we should expose that option to the user similar to the transform

ax3l:
thx for the fix! :)

PR #1644, opened 2016-11-18
add compile time power function

PrometheusPi:
This pull request adds a compile time power function for unsigned integer exponents.

Thus, for methods like the radiation plugin where a power function needs to be called N_omega * N_particles * N_directions per time step the more expensive `pow()` is avoided.

Usage:
```c++
/* 12^3 */
Pow<float_x, 3> pow3;
float_X results = pow3(12.0);
```

PR #1653, opened 2016-11-23
Add pow compile time using c++11

PrometheusPi:
This is an alternative to #1644. Instead of template meta programming C++11 `constexpr` is used.

 - [x] verify that this pow is not slower on GPUs than version #1644 
 - [x] retest more efficient method (verify against old results)

### DIFF
--- a/src/picongpu/include/plugins/radiation/utilities.hpp
+++ b/src/picongpu/include/plugins/radiation/utilities.hpp
@@ -70,9 +70,40 @@ namespace util
         }
     };
 
-}
 
 
+namespace details
+{
+  /** power function - with extra const parameter for efficient code
+    *
+    * T_type requires cast from int and multiplication
+    * @tparam T_Type - base type
+    * @param x - base value
+    * @param exp - exponent
+    * @param results (=1) - do not change - workaround to produce efficient code
+    * @return std::pow(x, exp)
+    */
+  template< typename T_Type >
+  HDINLINE constexpr T_Type pow( T_Type const x , uint32_t const exp, const T_Type result = T_Type( 1 ) )
+  {
+    return exp == 0 ? result : (
+        exp == 1 ? x * result : util::details::pow( x, exp - 1, result * x )
+    );
+  }
+} // namespace details
 
+  /** power function
+    *
+    * T_type requires cast from int and multiplication
+    * @tparam T_Type - base type
+    * @param x - base value
+    * @param exp - exponent
+    * @return std::pow(x, exp)
+    */
+  template< typename T_Type >
+  HDINLINE constexpr T_Type pow( T_Type const x , uint32_t const exp )
+  {
+    return util::details::pow( x, exp );
+  }
 
-
+} // namespace util


src/picongpu/include/plugins/radiation/utilities.hpp
```diff
@@ -70,9 +70,17 @@ namespace util
...
-
+  /** power function
+    *
+    * @tparam T - base type
+    * @param x - base value
+    * @param exp - exponent
+    * @return std::pow(x, exp)
+    */
+  template< typename T >
+  constexpr HDINLINE T pow( T x , unsigned int exp )
```
psychocoderHPC: please use `uint32_t` instead of `unsigned int`

src/picongpu/include/plugins/radiation/utilities.hpp
```diff
@@ -70,9 +70,17 @@ namespace util
...
-
-
+  /** power function
+    *
+    * @tparam T - base type
+    * @param x - base value
+    * @param exp - exponent
+    * @return std::pow(x, exp)
+    */
+  template< typename T >
```
psychocoderHPC: please call the template parameter `T_Type`

src/picongpu/include/plugins/radiation/utilities.hpp
```diff
@@ -70,9 +70,17 @@ namespace util
...
+    *
+    * @tparam T_Type - base type
+    * @param x - base value
+    * @param exp - exponent
+    * @return std::pow(x, exp)
+    */
+  template< typename T_Type >
+  constexpr HDINLINE T_Type pow( T_Type x , uint32_t exp )
+  {
+    return exp == 0 ? 1 : util::pow( x, exp - 1 );
```
psychocoderHPC: please update to:

```C++
return exp == 0 ? 1 : (
    exp == 1 ? x : util::pow( x, exp - 1 )
);
```

This save one multiplication. An effect of this change is only visible in compute intensive use cases.

psychocoderHPC: Please do not forget the cast of the first return value

PrometheusPi:
The influence of the new implementation on the calculation time is negligible.

src/picongpu/include/plugins/radiation/utilities.hpp
```diff
@@ -70,9 +70,19 @@ namespace util
...
-
+  /** power function
+    *
+    * @tparam T_Type - base type
+    * @param x - base value
+    * @param exp - exponent
+    * @return std::pow(x, exp)
+    */
+  template< typename T_Type >
+  constexpr HDINLINE T_Type pow( T_Type x , uint32_t exp )
```
PrometheusPi: `T_Type const x` and `uint32_t const exp`

src/picongpu/include/plugins/radiation/utilities.hpp
```diff
@@ -70,9 +70,19 @@ namespace util
...
+  {
+    return exp == 0 ? 1 : (
+        exp == 1 ? T_Type( x ) : util::pow( x, exp - 1 )
```
PrometheusPi: test this new version

src/picongpu/include/plugins/radiation/utilities.hpp
```diff
@@ -70,9 +70,19 @@ namespace util
...
+    *
+    * @tparam T_Type - base type
+    * @param x - base value
+    * @param exp - exponent
+    * @return std::pow(x, exp)
+    */
+  template< typename T_Type >
+  constexpr HDINLINE T_Type pow( T_Type x , uint32_t exp )
+  {
+    return exp == 0 ? 1 : (
```
PrometheusPi: here cast

src/picongpu/include/plugins/radiation/utilities.hpp
```diff
@@ -70,9 +70,19 @@ namespace util
...
 
-}
-
-
-
-
-
+  /** power function
+    *
+    * @tparam T_Type - base type
```
PrometheusPi: add comment on requirements on T_Type

PrometheusPi:
@psychocoderHPC The C++11 version of `pow` (this pull request) is **3.3 times slower** than the meta-template version (#1644). Do you have any idea why?
I think the current version is executed (with jumps) at run time and not decomposed to simple multiplications. If we can not avoid this slowdown, I would call for using #1644 instead.

PrometheusPi:
@psychocoderHPC I updated the code  so that the compile produces an efficient code now. There is no difference in register use to the meta-template version.

src/picongpu/include/plugins/radiation/utilities.hpp
```diff
@@ -70,9 +70,40 @@ namespace util
...
+    * @param x - base value
+    * @param exp - exponent
+    * @param results (=1) - do not change - workaround to produce efficient code
+    * @return std::pow(x, exp)
+    */
+  template< typename T_Type >
+  HDINLINE constexpr T_Type pow( T_Type const x , uint32_t const exp, const T_Type result = T_Type( 1 ) )
+  {
+    return exp == 0 ? result : (
+        exp == 1 ? x : util::details::pow( x, exp - 1, result * x )
```
PrometheusPi: here a `exp == 1 ? x*result : util::details::pow( x, exp - 1, result * x )`  is missing

PrometheusPi: yes it is - otherwise it is always `x^1`

PrometheusPi:
@psychocoderHPC It works: it gives the correct result and is fast now!

psychocoderHPC:
Do you checked the number of registers used?

PrometheusPi:
@psychocoderHPC Yes the register use differs (wen can discuss this in detail *offline* today) but it does not effect the performance.

psychocoderHPC:
Offline checked: register for radiation kernel not differ from the template class implementation.

PR #1655, opened 2016-11-25
C++11 constexpr: Remove Boost Macros

ax3l:
Remove all boost macros to use C++11 `constexpr` directly.

Changed via
```bash
# examples src
find examples -type f -print0 | xargs -0 sed -i 's/BOOST_STATIC_CONSTEXPR/static constexpr/g'
find examples -type f -print0 | xargs -0 sed -i 's/BOOST_CONSTEXPR_OR_CONST/constexpr/g'
```

 & minor manual typos and style issues fixed on the way.

Committed via
```bash
  git [...]
```

### DIFF
--- a/src/picongpu/include/plugins/radiation/Radiation.hpp
+++ b/src/picongpu/include/plugins/radiation/Radiation.hpp
@@ -550,7 +550,7 @@ private:
    *  Return:
    *  std::string - name
    *
-   * This method avoids initializing BOOST_STATIC_CONSTEXPR string arrays.
+   * This method avoids initializing static constexpr string arrays.
    */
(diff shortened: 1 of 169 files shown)


src/picongpu/include/plugins/radiation/Radiation.hpp
```diff
@@ -550,7 +550,7 @@ class Radiation : public ISimulationPlugin
    *  Return:
    *  std::string - name
    *
-   * This method avoids initializing BOOST_STATIC_CONSTEXPR string arrays.
+   * This method avoids initializing static constexpr string arrays.
```
psychocoderHPC: Should we remove this

PR #1656, opened 2016-11-25
openPMD base units: scoped enum

ax3l:
Use a *scoped enum* for openPMD `SIBaseUnits`. This is a C++11 feature making its declaration less verbose.


ax3l:
ah, that's ugly. the members need a `static_cast<int>` now to be used as `int`s if the enum is scoped. ugly.

ax3l:
ah, even with explicit type via `: unit32_t {...}`... no, that's ugly, we keep it was it is.

PR #1659, opened 2016-11-25
Fix Event System Deadlock on MPI_Barrier

psychocoderHPC:
Deadlock can be triggered if a mpi rank with unfinished PMacc tasks enters a MPI_Barrier.

- wait for all events before enter `MPI_Barrier`


ax3l:
note for the interested reader: the event system is a non-threaded *local* polling look while `MPI_Barrier` is a global block of the main thread controlling a GPU.

This part uses barriers to ensure that checkpoints are written successfully by all ranks before notifying the user about it. (That means: if you see a checkpoint, it will be valid and not thread crashed writing it.)

PR #1665, opened 2016-11-30
Device Selection: Guard Valid Range

psychocoderHPC:
device number outside valid range can be selected

The bug can be triggered if `CUDA_VISIBLE_DEVICES` is set to the number of GPUs in the node or if more MPI ranks as gpus per node are started.
The result is a run time crash. This bug is **not** critical and need no back port.

### DIFF
--- a/src/libPMacc/include/Environment.hpp
+++ b/src/libPMacc/include/Environment.hpp
@@ -180,7 +180,7 @@ private:
         {
             throw std::runtime_error("no CUDA capable devices detected");
         }
-        else if (num_gpus < deviceNumber) //check if device can be selected by deviceNumber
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/Environment.hpp
```diff
@@ -180,7 +180,7 @@ class Environment
         {
             throw std::runtime_error("no CUDA capable devices detected");
         }
-        else if (num_gpus < deviceNumber) //check if device can be selected by deviceNumber
+        else if (num_gpus <= deviceNumber) //check if device can be selected by deviceNumber
```
ax3l: offline discussion: yes, writing `num_gpus >= deviceNumber` is easier to read

PR #1668, opened 2016-12-01
Restart: Allow Disable of Moving Window

ax3l:
### Feature Description

Allows running a simulation with `-m`, writing a checkpoint and restarting that checkpoint *without* `-m`.

### Notes

This does not break restarts that keep the moving window "-m" on.

Warning: *enabling* the moving window from a checkpoint that had no moving window does still not work nor from a sim was "paused" with that method will not work. This would require remembering "how many iterations did someone stop the moving window".

First build for @Heikman during his Master's thesis and requested again by [Bifeng](https://cg.hzdr.de/Lists/picongpu-users/Message/151.html) on the user mailing list. The patch was so small and quick that I forgot to write a PR for it.

### DIFF
--- a/src/libPMacc/include/communication/CommunicatorMPI.hpp
+++ b/src/libPMacc/include/communication/CommunicatorMPI.hpp
@@ -244,6 +244,10 @@ public:
 
     bool setStateAfterSlides(size_t numSlides)
     {
+        // nothing happens
+        if(numSlides == 0)
+            return false;
+
         // we can only slide in y direction right now
         if(DIM < DIM2)
             return false;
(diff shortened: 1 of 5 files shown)


psychocoderHPC:
To avoid side effect that something is not valid initialized we should also remove [the moving window check in MySimulation](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/simulationControl/MySimulation.hpp#L391)

ax3l:
for non-moving window with static load balancing in y direction calling `setStateAfterSlides` kills the static load balancing

PR #1670, opened 2016-12-02
improve error message

psychocoderHPC:
- improve error message for `GetNComponents` if not specialized for a given type

If the trait is not defined for a type the compiler fire a error message `incomplete type is not allow`. There is no information what does this mean. With this pull request the developer get useful information why the code is not compiling.

### DIFF
--- a/src/libPMacc/include/traits/GetNComponents.hpp
+++ b/src/libPMacc/include/traits/GetNComponents.hpp
@@ -35,7 +35,11 @@ namespace traits
  * \return \p ::value as public with number of components (uint32_t)
  */
 template<typename T_Type, bool T_IsFundamental = boost::is_fundamental<T_Type>::value>
-struct GetNComponents;
+struct GetNComponents
+{
+    PMACC_CASSERT_MSG_TYPE( __GetNComponents_is_not_defined_for_this_type, T_Type, false ); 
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/traits/GetNComponents.hpp
```diff
@@ -35,7 +35,11 @@ namespace traits
  * \return \p ::value as public with number of components (uint32_t)
  */
 template<typename T_Type, bool T_IsFundamental = boost::is_fundamental<T_Type>::value>
-struct GetNComponents;
+struct GetNComponents
+{
+    PMACC_CASSERT_MSG_TYPE( __GetNComponents_is_not_defined_for_this_type, T_Type, false ); 
+    static constexpr uint32_t dim = 0;
```
PrometheusPi: @psychocoderHPC Why do you need this integer variable here and why was it not needed before?

psychocoderHPC: This trait was unspecified by default. Now it is always specialized but if it is used on a unspecified  type the assert is triggered. 
The variable should only show the signature of this trait and avoid error message in form of `class has no member called `value`)

btw: `dim` should be called `value`

psychocoderHPC:
updated

PR #1669, opened 2016-12-02
Kernels: Refactor Functions to Functors

psychocoderHPC:
- remove macro `__cudaKernel`
- remove  macro `__picKernelArea`
- add new kernel creation methods
  - PMacc kernel wrapper: `Kernel<>`
  - helper function: `kernel()`
  - macro: `PMACC_KERNEL()`
- use new kernel starter
- indent kernel body within the functor
- remove the usage of the type `dim3` within the user code
- `Reduce.hpp` and `IdProvider.hpp`: remove early returns within a kernel

I can't avoid the white space changes, else the code become not readable and maintainable.

Why is `__picKernelArea` removed?
- this macro add an extra parameter to the kernel (this is confusing)
- the macro is hard to maintain and to read

Tests:

- [x] compiled with `compie -l examples ...`
- [x] compile PMacc tests
- [x] compile PMacc GameOfLife
- [x] runtime test KHI 2/3D positrons/electrons shape CIC/TSC/PCS/P4S

~~This is only a preview. I will split do separate commits for PMacc and PIConGPU~~

**How to refactor a CUDA kernel to a functor:**

- remove `__cudaKernel(...)` with `PMACC_KERNEL(...)`
- remove `__picKernelArea`
```C++
__picKernelArea( kernelName, cellDescription, AREA )( blockExtent )( ... );
```
to
```C++
AreaMapping< AREA, MappingDesc > mapper( cellDescription );
PMACC_KERNEL( kernelName{} )( mapper.getGridDim(), blockExtent )( ..., mapper );
```
- transform kernel function to a functor:
```C++
template< typename T_Type >
__global__ void kernelName( T_Type a ){ }
```
to
```C++
struct KernelName
{
  template< typename T_Type >
  DINLINE void operator( )( T_Type a ) const
  { };
}
```

PR #1675, opened 2016-12-07
Travis: Simple Style Checks

ax3l:
We add travis to do tests on source files which are light-weight. This currently only includes #1672 but could also look for things like "careful, found a new external include", "found a new usage of a boost library", etc.

Let the scripts come!
- no `__global__` keyword
- no EOL white spaces
- non non-ASCII chars in source and script files

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+language: cpp
+sudo: false
+
+script:
+  #############################################################################
+  # Conformance with Alpaka: Do not write __global__ CUDA kernels directly    #
+  #############################################################################
+  - test/hasCudaGlobalKeyword src/libPMacc
+  - test/hasCudaGlobalKeyword src/picongpu
+  - test/hasCudaGlobalKeyword examples
+
+  #############################################################################
+  # Disallow end-of-line (EOL) white spaces                                   #
+  #############################################################################
+  - test/hasEOLwhiteSpace
+
+  #############################################################################
+  # Disallow non-ASCII in source files and scripts                            #
+  #############################################################################
+  - test/hasNonASCII
--- a/test/hasEOLwhiteSpace
+++ b/test/hasEOLwhiteSpace
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+#
+# Copyright 2016 Axel Huebl, Rene Widera
+#
+# This file is part of PIConGPU.
+#
+# PIConGPU is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# PIConGPU is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PIConGPU.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
+# search recursive inside a folder if a file contains end-of-line
+#   (EOL) white spaces
+#
+# @result 0 if no files are found, else 1
+#
+# note: .md files can contain EOL white spaces for syntax reasons
+#       (newline without starting a new paragraph)
+
+ok=0
+
+for i in $(find . \
+                -not -path "./.git/*" \
+                -not -path "./thirdParty/*" \
+                -type f | \
+           egrep "\.def$|\.h$|\.cpp$|\.cu$|\.hpp$|\.tpp$|\.kernel$|\.loader$|\
+                  \.param$|\.unitless$|\.sh$|\.bash$|\.cfg$|\.tpl$|\.conf$|\
+                  \.awk$|\.gnuplot$|\.cmake$|\.profile$|\.example$|\.py$")
+do
+  grep -q "[[:blank:]]\+$" $i
+  if [ $? -eq 0 ]
+  then
+    echo "$i contains EOL white spaces!"
+    ok=1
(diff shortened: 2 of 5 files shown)


ax3l:
great, will rebase after #1669 to let the check pass.

PrometheusPi:
@ax3l I merged #1680 - please recommit in order to run compile test again

test/hasEOLwhiteSpace
```diff
@@ -0,0 +1,51 @@
...
+# @param $1 path to the folder
+# @result 0 if no files are found, else 1
+#
+# note: .md files can contain EOL white spaces for syntax reasons
+#       (newline without starting a new paragraph)
+folderName=$1
+ok=0
+for i in $(find . -type f | egrep "*.def$|*.h$|*.cpp$|*.cu$|*.hpp$|*.tpp$|*.kernel$|*.loader$|*.param$|*.unitless$")
+do
+  grep ^.*[[:space:]]$ $i >/dev/null
```
psychocoderHPC: please use `"grep "[[:blank:]]\+$" $i"`

- [more information about `\+`](https://www.gnu.org/software/sed/manual/html_node/Regular-Expressions.html) 
- `[[:blank:]]` substitute spaces and tabs

ax3l:
depends also on #1682 now :)

test/hasEOLwhiteSpace
```diff
@@ -0,0 +1,51 @@
...
+done
+
+if [ $ok -ne 0 ]
+then
+  echo ""
+  echo "SUMMARY"
+  echo "-------"
+  echo "Run the following command on the above files to remove your"
+  echo "end-of-line (EOL) white spaces:"
+  echo "  sed -i 's/\(^.*[^[:space:]]\)[[:space:]]*$/\1/g' putYourFileNameHere"
```
psychocoderHPC: as you found in #1679 the sed is wrong and should be `sed -i 's/[[:blank:]]\+$//'`

test/hasEOLwhiteSpace
```diff
@@ -0,0 +1,51 @@
...
+# search recursive inside a folder if a file contains end-of-line
+#   (EOL) white spaces
+# @param $1 path to the folder
+# @result 0 if no files are found, else 1
+#
+# note: .md files can contain EOL white spaces for syntax reasons
+#       (newline without starting a new paragraph)
+folderName=$1
+ok=0
+for i in $(find . -type f | egrep "*.def$|*.h$|*.cpp$|*.cu$|*.hpp$|*.tpp$|*.kernel$|*.loader$|*.param$|*.unitless$")
```
psychocoderHPC: shoudl we extend the list with the following types?
```
awk
cfg
cmake
conf
example
gnuplot
py
sh
tpl
```

.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: cpp
+sudo: false
+
+script:
+  #############################################################################
+  # Conformance with Alpaka: Do not write __global__ CUDA kernels directly    #
+  #############################################################################
+  - test/hasCudaGlobalKeyword src/libPMacc
+  - test/hasCudaGlobalKeyword src/picongpu
```
psychocoderHPC: we should also check `src/tools`

test/hasEOLwhiteSpace
ax3l: added!

test/hasEOLwhiteSpace
ax3l: done

test/hasEOLwhiteSpace
ax3l: done

test/hasEOLwhiteSpace
```diff
@@ -0,0 +1,56 @@
...
+
+folderName=$1
+ok=0
+
+for i in $(find $folderName -type f | \
+           egrep "\.def$|\.h$|\.cpp$|\.cu$|\.hpp$|\.tpp$|\.kernel$|\.loader$|\
+                  \.param$|\.unitless$|\.sh$|\.bash$|\.cfg$|\.tpl$|\.conf$|\
+                  \.awk$|\.gnuplot$|\.cmake$|\.profile$|\.example$|\.py$")
+do
+  grep "[[:blank:]]\+$" $i >/dev/null
```
psychocoderHPC: only a note: grep has an option `-q` or `--silent`

ax3l: changed, will push after all fixing PRs are merged

test/hasNonASCII
```diff
@@ -0,0 +1,50 @@
...
+
+for i in $(find . \
+                -not -path "./.git/*" \
+                -not -path "./thirdParty/*" \
+                -type f | \
+           egrep "\.def$|\.h$|\.cpp$|\.cu$|\.hpp$|\.tpp$|\.kernel$|\.loader$|\
+                  \.param$|\.unitless$|\.sh$|\.bash$|\.cfg$|\.tpl$|\.conf$|\
+                  \.awk$|\.gnuplot$|\.cmake$|\.profile$|\.example$\.py$")
+do
+  result=$(grep --color='always' -P -n "[\x80-\xFF]" $i)
```
ax3l: reference: http://stackoverflow.com/questions/3001177/how-do-i-grep-for-all-non-ascii-characters-in-unix/9395552#9395552

psychocoderHPC:
Note:  to detect non ASCII files we can also use the command `file`

```
for i in `find src/ -type f | egrep "\.def$|\.h$|\.cpp$|\.cu$| \
    \.hpp$|\.tpp$|*.kernel$|*.loader$|*.param$| \
    \.unitless$" ` ;  do file --mime-encoding $i | grep  -q "us-ascii" ; if [ $? -ne 0 ] ; then file --mime-encoding $i; fi; done 
```

PR #1159, opened 2015-10-01
Third Party: cuda_memtest update

ax3l:
This PR updates `cuda_memtest` with the latest updates from our [upstream repository](https://github.com/ComputationalRadiationPhysics/cuda_memtest):
  https://github.com/ComputationalRadiationPhysics/cuda_memtest/pull/2
  https://github.com/ComputationalRadiationPhysics/cuda_memtest/pull/3
  https://github.com/ComputationalRadiationPhysics/cuda_memtest/pull/4
  https://github.com/ComputationalRadiationPhysics/cuda_memtest/pull/6 (follow up of #1156)

Updated via:
```bash
GIT_AUTHOR_NAME="Third Party" GIT_AUTHOR_EMAIL="picongpu@hzdr.de" \
  git subtree pull --prefix thirdParty/cuda_memtest/ \
  git@github.com:ComputationalRadiationPhysics/cuda_memtest.git dev --squash
```

PR #1684, opened 2016-12-08
Source Files: Remove non-ASCII Chars

ax3l:
We should not enforce UTF-8 locales on user systems.

CUDA memtest updated as in #1159, also contains an other minor fix from upstream: https://github.com/ComputationalRadiationPhysics/cuda_memtest/pull/7

For comparison, travis does not even handle UTF-8 grep-ing gracefully:
  https://travis-ci.org/ComputationalRadiationPhysics/picongpu/builds/182302127#L209

### DIFF
--- a/src/picongpu/include/plugins/IsaacPlugin.hpp
+++ b/src/picongpu/include/plugins/IsaacPlugin.hpp
@@ -398,8 +398,8 @@ private:
             if (rank == 0)
             {
                 json_object_set_new( visualization->getJsonMetaRoot(), "time step", json_string( "Time step" ) );
-                json_object_set_new( visualization->getJsonMetaRoot(), "drawing time", json_string( "Drawing time in µs" ) );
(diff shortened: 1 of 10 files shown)


src/picongpu/include/plugins/IsaacPlugin.hpp
```diff
@@ -398,8 +398,8 @@ class IsaacPlugin : public ILightweightPlugin
             if (rank == 0)
             {
                 json_object_set_new( visualization->getJsonMetaRoot(), "time step", json_string( "Time step" ) );
-                json_object_set_new( visualization->getJsonMetaRoot(), "drawing time", json_string( "Drawing time in µs" ) );
```
psychocoderHPC: is `µs` not `us`?

theZiz: 1999 called - they want their errors back! :phone: :smiling_imp:

Issue #1683, opened 2016-12-08
Radiation: vector trows an compile error if [ ] operator is used on const 


PrometheusPi:
Currently, the [vector class used in the radiation plugin](https://github.com/ComputationalRadiationPhysics/picongpu/blob/321298cccb69ba86db611c683db8772ea17b4f6d/src/picongpu/include/plugins/radiation/vector.hpp) throws an compile error if the `[ ]` operator is used on `const vector`.
This should be resolved.

PrometheusPi:
closed with  #1686

PR #1686, opened 2016-12-08
fix vector.operator[] const access in radiation plugin

PrometheusPi:
This pull request fixes a compile time bug in the radiation plugin's vector class that prevented using the `[ ]` operator on constant vectors.

This solves issue #1683


PrometheusPi:
@psychocoderHPC - rebased after white space travis check inclusion

PR #1691, opened 2016-12-13
begin kernel names with upper case letter

psychocoderHPC:
- rename kernel which begin with lower case letter
- update kernel call with new name

### DIFF
--- a/src/libPMacc/examples/gameOfLife2D/include/Evolution.hpp
+++ b/src/libPMacc/examples/gameOfLife2D/include/Evolution.hpp
@@ -38,7 +38,7 @@ namespace gol
     {
         using namespace PMacc;
 
-        struct evolution
+        struct Evolution
         {
             template<class BoxReadOnly, class BoxWriteOnly, class Mapping>
             DINLINE void operator()(BoxReadOnly buffRead,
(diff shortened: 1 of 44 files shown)


PrometheusPi:
@psychocoderHPC In `src/libPMacc/include/particles/IdProvider.hpp` there are still some instances of `setNextId` - line 81 and 92

psychocoderHPC:
`setNextId` in line 81 and 92 are not kernels

PrometheusPi:
In `./src/libPMacc/include/particles/operations/splitIntoListOfFrames.kernel` line 178 you have a function defined as `splitIntoListOfFrames` which you use in `./src/picongpu/include/plugins/hdf5/restart/LoadSpecies.hpp:193` and `./src/picongpu/include/plugins/adios/restart/LoadSpecies.hpp:169`. Do you realy want to use the function there or the kernel? Wky the name conflict?

psychocoderHPC:
The kernel is in it's own namespace, therefore it not conflicts. IN both cases the function should be called. A kernel is called with `PMACC_KERNEL(...)`

@PrometheusPi  Please use links to code parts, this helps to follow the issu.

PrometheusPi:
@psychocoderHPC thanks for checking

Issue #829, opened 2015-04-18
Docker: Compile Suite and Examples


ax3l:
[Docker](https://en.wikipedia.org/wiki/Docker_%28software%29) is an easy, open-source way to share complete images of operating systems including software.

The extremely [easy accessebility](https://www.docker.com/tryit) might not only be very useful for our compile suite (multiple os, multiple images with "oldest"/"newest" deps, very easy command line control and access, chroot) but might also be useful (needs evaluation) for a first try without needing to install all dependencies.

Docker containers can also be stacked/depend on each other, e.g., offical docker images for common OS exist.

bussmann:
I would LOVE to see that!

ax3l:
Since docker is using LXC on linux, CUDA should be no problem at all.

Would be interesting if running a linux instance with a CUDA application can be run [on a mac](https://docs.docker.com/installation/mac/), too. And yes, there might be [windows people](https://docs.docker.com/installation/windows/) too - but the VM-ization there might be a problem for CUDA.

Site note: one can spin up docker instances on Amazone AWS easily, so that issue is related to #764.

ax3l:
[rkt app container](https://github.com/coreos/rkt/) (de: [heise](http://www.heise.de/open/meldung/Anwendungs-Container-mit-Rocket-Docker-Konkurrent-stoesst-auf-breite-Zustimmung-2633392.html))

ax3l:
[new blog article](https://devblogs.nvidia.com/parallelforall/nvidia-docker-gpu-server-application-deployment-made-easy) about containerization of cuda applications :) (I want us to be listed [here](https://github.com/NVIDIA/nvidia-docker/wiki/Third-party) :D )

ax3l:
Here is a first `Dockerfile` based on [`nvidia/cuda:7.0-devel`](https://hub.docker.com/r/nvidia/cuda/) building one/all PIConGPU examples in the `release-0.2.0` branch with C++98:
```bash
FROM       nvidia/cuda:7.0-devel
MAINTAINER Axel Huebl <a.huebl[ät]hzdr.de>

ENV        HOME /root
ENV        DEBIAN_FRONTEND noninteractive

# install dependencies provided by packages
RUN        apt-get update \
           && apt-get install -y --no-install-recommends \
              autoconf \
              git \
              pkg-config \
              make \
              libhdf5-openmpi-dev \
              libopenmpi-dev \
              libpng-dev \
              wget \
[...]

PR #1692, opened 2016-12-13
CMake: handle PMacc as separate library

psychocoderHPC:
- PMaccConfig.cmake: search for all dependencies
  - CUDA
  - Boost
  - mallocMC
  - MPI
- refactor PMacc test CMake files
- refactor PIConGPU CMake file
- refactor gol CMake file

**Tests**

- [x] compile PMacc tests
- [x] compile PMacc example gol
- [x] compile PIConGPU with ISAAC

### DIFF
--- a/src/libPMacc/CMakeLists.txt
+++ b/src/libPMacc/CMakeLists.txt
@@ -23,24 +23,25 @@
 ################################################################################
 # PMacc tests
 ################################################################################
+
 cmake_minimum_required(VERSION 3.3.0)
 project(PMaccTest)
 
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_EXTENSIONS OFF)
-set(CMAKE_CXX_STANDARD 11)
 
 ################################################################################
 # PMacc
 ################################################################################
+
 find_package(PMacc REQUIRED CONFIG PATHS ${CMAKE_CURRENT_SOURCE_DIR})
 include_directories(SYSTEM ${PMacc_INCLUDE_DIRS})
 set(LIBS ${LIBS} ${PMacc_LIBRARIES})
 add_definitions(${PMacc_DEFINITIONS})
 
+
 ###############################################################################
 # Boost.Test
 ###############################################################################
+
 find_package(Boost 1.57.0 COMPONENTS unit_test_framework  REQUIRED)
 include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
 set(LIBS ${LIBS} ${Boost_LIBRARIES})
--- a/src/libPMacc/CMakeLists.txt
+++ b/src/libPMacc/CMakeLists.txt
@@ -27,6 +27,24 @@
 cmake_minimum_required(VERSION 3.3.0)
 project(PMaccTest)
 
+# set helper pathes to find libraries and packages
+# Add specific hints
+list(APPEND CMAKE_PREFIX_PATH "$ENV{MPI_ROOT}")
+list(APPEND CMAKE_PREFIX_PATH "$ENV{CUDA_ROOT}")
+list(APPEND CMAKE_PREFIX_PATH "$ENV{BOOST_ROOT}")
+list(APPEND CMAKE_PREFIX_PATH "$ENV{VT_ROOT}")
+# Add from environment after specific env vars
+list(APPEND CMAKE_PREFIX_PATH "$ENV{CMAKE_PREFIX_PATH}")
+
+
+###############################################################################
+# Language Flags
+###############################################################################
+
+# enforce C++11
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_CXX_STANDARD 11)
 
 ################################################################################
 # PMacc
@@ -38,13 +56,29 @@ set(LIBS ${LIBS} ${PMacc_LIBRARIES})
 add_definitions(${PMacc_DEFINITIONS})
 
 
+################################################################################
+# Find CUDA
+################################################################################
+
+find_package(CUDA 7.5 REQUIRED)
+
+
 ###############################################################################
 # Boost.Test
 ###############################################################################
 
-find_package(Boost 1.57.0 COMPONENTS unit_test_framework  REQUIRED)
+find_package(Boost 1.57.0 COMPONENTS unit_test_framework REQUIRED)
 include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
 set(LIBS ${LIBS} ${Boost_LIBRARIES})
+
+
+################################################################################
+# Find MPI
+################################################################################
+
+find_package(MPI REQUIRED)
+include_directories(SYSTEM ${MPI_C_INCLUDE_PATH})
+set(LIBS ${LIBS} ${MPI_C_LIBRARIES})
 
 
 ###############################################################################
--- a/src/libPMacc/PMaccConfig.cmake
+++ b/src/libPMacc/PMaccConfig.cmake
@@ -31,8 +31,6 @@
 ###############################################################################
 cmake_minimum_required(VERSION 3.3.0)
 
-set(PMacc_INCLUDE_DIRS ${PMacc_INCLUDE_DIRS} "${PMacc_DIR}/include")
-
 # set helper pathes to find libraries and packages
 # Add specific hints
 list(APPEND CMAKE_PREFIX_PATH "$ENV{MPI_ROOT}")
@@ -41,8 +39,6 @@ list(APPEND CMAKE_PREFIX_PATH "$ENV{BOOST_ROOT}")
 list(APPEND CMAKE_PREFIX_PATH "$ENV{VT_ROOT}")
 # Add from environment after specific env vars
 list(APPEND CMAKE_PREFIX_PATH "$ENV{CMAKE_PREFIX_PATH}")
-# Last add generic system path to the end (as last fallback)
-list(APPEND "/usr/lib/x86_64-linux-gnu/")
 
 # own modules for find_packages e.g. FindmallocMC
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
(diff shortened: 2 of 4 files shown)


src/libPMacc/CMakeLists.txt
```diff
@@ -23,24 +23,25 @@
 ################################################################################
 # PMacc tests
 ################################################################################
+
 cmake_minimum_required(VERSION 3.3.0)
 project(PMaccTest)
 
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
ax3l: hm, if you actually *have* a `PMaccTest` project, that one *should* be C++11 but ok-ish...
did you get problems leaving that in?

src/libPMacc/PMaccConfig.cmake
```diff
@@ -33,50 +33,41 @@ cmake_minimum_required(VERSION 3.3.0)
...
+# set helper pathes to find libraries and packages
+# Add specific hints
+list(APPEND CMAKE_PREFIX_PATH "$ENV{MPI_ROOT}")
+list(APPEND CMAKE_PREFIX_PATH "$ENV{CUDA_ROOT}")
+list(APPEND CMAKE_PREFIX_PATH "$ENV{BOOST_ROOT}")
+list(APPEND CMAKE_PREFIX_PATH "$ENV{VT_ROOT}")
+# Add from environment after specific env vars
+list(APPEND CMAKE_PREFIX_PATH "$ENV{CMAKE_PREFIX_PATH}")
+# Last add generic system path to the end (as last fallback)
+list(APPEND "/usr/lib/x86_64-linux-gnu/")
```
ax3l: bug: lacking the list variable to append to

ax3l: side note: we should review if those are actually still necessary and useful

src/libPMacc/PMaccConfig.cmake
```diff
@@ -167,41 +155,93 @@ if(VAMPIR_ENABLE)
...
+# Find MPI
 ################################################################################
-find_package(MPI MODULE REQUIRED)
+
+find_package(MPI REQUIRED)
 set(PMacc_INCLUDE_DIRS ${PMacc_INCLUDE_DIRS} ${MPI_C_INCLUDE_PATH})
 set(PMacc_LIBRARIES ${PMacc_LIBRARIES} ${MPI_C_LIBRARIES})
-set(PMacc_LIBRARIES ${PMacc_LIBRARIES} ${MPI_CXX_LIBRARIES})
+
+# bullxmpi fails if it can not find its c++ counter part
```
ax3l: good one putting that in a guard, but I am not sure if `MPI_CXX_FOUND` is set to true if one is not specifically requesting that `COMPONENT` in `find_package`, but might work. (Side check in taurus would be nice.)

src/libPMacc/PMaccConfig.cmake
```diff
@@ -167,41 +155,93 @@ if(VAMPIR_ENABLE)
...
 endif(VAMPIR_ENABLE)
 
+
 ################################################################################
-# MPI LIB
+# Find MPI
 ################################################################################
-find_package(MPI MODULE REQUIRED)
+
+find_package(MPI REQUIRED)
```
ax3l: why not as a module any more?

src/libPMacc/PMaccConfig.cmake
```diff
@@ -167,41 +155,93 @@ if(VAMPIR_ENABLE)
...
-  list(APPEND PNGwriter_DEFINITIONS "-DGOL_ENABLE_PNG=1")
-  set(PMacc_DEFINITIONS ${PMacc_DEFINITIONS} ${PNGwriter_DEFINITIONS})
-  set(PMacc_LIBRARIES ${PMacc_LIBRARIES} ${PNGwriter_LIBRARIES})
-endif(PNGwriter_FOUND)
 
-###############################################################################
-# Boost LIB
-###############################################################################
-find_package(Boost 1.57.0 MODULE REQUIRED COMPONENTS program_options regex system)
+find_package(Boost 1.57.0 REQUIRED COMPONENTS filesystem system thread math_tr1)
```
ax3l: did you check if we need all of those in PMacc itself?

ax3l: careful, you should request system alongside filesystem

src/libPMacc/examples/gameOfLife2D/CMakeLists.txt
```diff
@@ -131,84 +57,22 @@ endif(VAMPIR_ENABLE)
...
-find_package(Threads REQUIRED)
-set(LIBS ${LIBS} ${CMAKE_THREAD_LIBS_INIT})
-
-
 ###############################################################################
 # Find Boost
 ###############################################################################
 
-find_package(Boost 1.49.0 REQUIRED COMPONENTS program_options regex system)
+find_package(Boost 1.49.0 REQUIRED COMPONENTS program_options )
```
ax3l: aren't we using things like `regex` or `system` used in GoL any more?
Can you update to boost 1.57.0+ for consistency?
Weird space before ` )`

src/libPMacc/examples/gameOfLife2D/CMakeLists.txt
```diff
@@ -131,84 +57,22 @@ endif(VAMPIR_ENABLE)
...
 include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
 set(LIBS ${LIBS} ${Boost_LIBRARIES})
 
-# work-arounds
```
ax3l: requireing that PMacc will take care of all boost work-arounds is a bit odd, because we might need more but ok-ish... thise one is outdated with 1.57+ anyway

src/libPMacc/PMaccConfig.cmake
ax3l: this line was not useful/active in the last version where it was copied from, so remove it right away

src/picongpu/CMakeLists.txt
```diff
@@ -167,33 +59,15 @@ endif(VAMPIR_ENABLE)
...
 else(PIC_RELEASE)
     message(STATUS "Debug version")
-
     set(CMAKE_BUILD_TYPE Debug)
-    set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS}" -g "-Xcompiler=-g,-pthread")
 endif(PIC_RELEASE)
 
 
 ################################################################################
-# Find MPI
```
ax3l: We **are** using MPI directly in PIConGPU, so PIConGPU should  do a `find_package` on its own (again).

src/picongpu/CMakeLists.txt
```diff
@@ -219,73 +93,6 @@ find_package(Boost 1.57.0 REQUIRED COMPONENTS program_options regex filesystem
 include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
 set(LIBS ${LIBS} ${Boost_LIBRARIES})
 
-# Boost 1.55 added support for a define that makes result_of look for
```
ax3l: Not the biggest fan of putting this in PMacc only if we use it in PIConGPU `result_of` definitions, too.

src/picongpu/CMakeLists.txt
```diff
@@ -219,73 +93,6 @@ find_package(Boost 1.57.0 REQUIRED COMPONENTS program_options regex filesystem
...
-if( (Boost_VERSION GREATER 105999) AND
-    (NOT CUDA_VERSION VERSION_EQUAL 7.5) )
-    # Boost Bug https://svn.boost.org/trac/boost/ticket/11897
-    message(STATUS "Boost: Disable template aliases")
-    add_definitions(-DBOOST_NO_CXX11_TEMPLATE_ALIASES)
-endif()
-
-
-################################################################################
-# Find OpenMP
```
ax3l: We **use** `OpenMP` in PIConGPU directly, so PIConGPU **should** run a `find_package` for `OpenMP` itself (again)!

src/picongpu/CMakeLists.txt
```diff
@@ -219,73 +93,6 @@ find_package(Boost 1.57.0 REQUIRED COMPONENTS program_options regex filesystem
...
-################################################################################
-
-find_package(OpenMP)
-if(OPENMP_FOUND)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
-endif()
-
-
-################################################################################
-# Find mallocMC
```
ax3l: Same with mallocMC: used directly in PIConGPU!

src/libPMacc/PMaccConfig.cmake
psychocoderHPC: MPI find_package auto detect all available languages. [documentation](https://cmake.org/cmake/help/v3.3/module/FindMPI.html)

src/libPMacc/examples/gameOfLife2D/CMakeLists.txt
psychocoderHPC: Yes the workaround are handled by PMacc.

src/picongpu/CMakeLists.txt
psychocoderHPC: PMacc is the root for the non common usage of `result_of`

src/libPMacc/PMaccConfig.cmake
psychocoderHPC: [good short documentation `CONFIG` vs. `MODULE` ](https://github.com/ruslo/hunter/wiki/example.find_package.config)

If we switch to modern cmake target we should use ĆONFIG`

psychocoderHPC:
Currently I can't test the cmake files, because our GPU system is down.

ax3l:
no prob, use [Docker](https://github.com/ComputationalRadiationPhysics/picongpu/issues/829#issuecomment-241768839) :)

src/picongpu/CMakeLists.txt
```diff
@@ -184,14 +91,9 @@ endif(PIC_RELEASE)
 ################################################################################
 
 find_package(MPI REQUIRED)
-include_directories(SYSTEM ${MPI_C_INCLUDE_PATH})
```
psychocoderHPC: dropped SYSTEM

note to me, check this

ax3l:
Oh noooo. System for -isystem is our friend! Please keep for all dependencies.

psychocoderHPC:
I updated the cmake files and tested all on our system.

src/libPMacc/PMaccConfig.cmake
ax3l: all right! also: compiles on taurus

PR #1693, opened 2016-12-13
Close #1687 Fix Pre-Compiler Spaces

### LINKED ISSUE #1687: white space before precompiler macros
With pull request #1669 several pre-compiler macros where moved 4 spaces to the right.
This conflicts the the coding style. 

As example: [here](https://github.com/psychocoderHPC/picongpu/blob/c66fedf6949f0a7a8120a04597a5bd722866ca40/examples/SingleParticleCurrent/include/particles/ParticlesInitOneParticle.hpp#L92-L96)

ax3l:
Close #1687: Updated with sed script (see separate PR) and comitted via generic "tools" user.

~~Looks a bit odd for `pragma omp ...` and `pragma unroll ...` instructions. Not sure if we really want that for `#pragma`, too because in such a case it is not a `pre-compiler` instruction but a *non-standard, extended type description* (attributes, no-inline, etc.) or *non-standard, compiler-specific compile-time hint*.~~ script and style rule updated

### DIFF
--- a/src/libPMacc/include/lambda/make_Functor.hpp
+++ b/src/libPMacc/include/lambda/make_Functor.hpp
@@ -30,7 +30,7 @@
 #include "CT/FillTerminalList.hpp"
 #include <math/Tuple.hpp>
 #include "RefWrapper.hpp"
- #include <boost/type_traits/is_reference.hpp>
+#include <boost/type_traits/is_reference.hpp>
 
 #include <boost/mpl/if.hpp>
 #include <boost/mpl/front.hpp>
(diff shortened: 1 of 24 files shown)


src/libPMacc/include/lambda/make_Functor.hpp
```diff
@@ -30,7 +30,7 @@
 #include "CT/FillTerminalList.hpp"
 #include <math/Tuple.hpp>
 #include "RefWrapper.hpp"
- #include <boost/type_traits/is_reference.hpp>
+# include <boost/type_traits/is_reference.hpp>
```
PrometheusPi: @ax3l Is this space  OK for you?

PR #1694, opened 2016-12-13
Test: Check & Remove Pre-Compiler Spaces

ax3l:
Tests and removes spaces before using pre-compiler commands. Was used to fix #1693.

~~Needs decision on `#pragma` lines https://github.com/ComputationalRadiationPhysics/contributing/issues/16 before merging.~~ ok

### DIFF
--- a/test/hasSpaceBeforePrecompiler
+++ b/test/hasSpaceBeforePrecompiler
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+#
+# Copyright 2016 Axel Huebl
+#
+# This file is part of PIConGPU.
+#
+# PIConGPU is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# PIConGPU is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PIConGPU.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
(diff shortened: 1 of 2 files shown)


test/hasSpaceBeforePrecompiler
```diff
@@ -0,0 +1,72 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PIConGPU.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
+# Search recursive inside our C++ source files if they violate our coding
+# style by using the wrong pre-compiler spacing style.
```
PrometheusPi: **a** wrong pre-compiler spacing style
(or is there only one?)

test/hasSpaceBeforePrecompiler
```diff
@@ -0,0 +1,72 @@
...
+# See:
+#   https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md
```
PrometheusPi: Could you use the more explicit:
`https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md#7-preprocessor` here?

test/hasSpaceBeforePrecompiler
```diff
@@ -0,0 +1,72 @@
...
+#
+# @result 0 if no files are found, else 1
```
PrometheusPi: if no files **with tabs** are found

test/hasSpaceBeforePrecompiler
ax3l: I think the number of the chapter will change too quickly but can do

test/hasSpaceBeforePrecompiler
ax3l: sure!

ax3l:
all updated, will push after #1693 was merged

PR #1701, opened 2016-12-14
[WIP] make absorber reach a specific damping value

PrometheusPi:
This pull request is a minimal change in the behavior of the absorber.

Currently `ABSORBER_STRENGTH` and `ABSORBER_CELLS` are independent and one might reach no damping if the first is too small compared to the second.

This pull request redefines the meaning of `ABSORBER_STRENGTH` to be the damping factor the be reached (exponentially). 

 - [ ] perform tests

### DIFF
--- a/src/picongpu/include/fields/FieldManipulator.hpp
+++ b/src/picongpu/include/fields/FieldManipulator.hpp
@@ -62,7 +62,7 @@ public:
 
                 uint32_t thickness = ABSORBER_CELLS[direction][pos_or_neg];
 
-                float_X absorber_strength = -1.0 * math::log( ABSORBER_STRENGTH[direction][pos_or_neg] ) * thickness;
+                float_X absorber_strength = -1.0 * math::log( ABSORBER_MAXFACTOR[direction][pos_or_neg] ) * thickness;
 
                 if (thickness == 0) continue; /*if the absorber has no thickness we check the next side*/
 
--- a/src/picongpu/include/fields/FieldManipulator.hpp
+++ b/src/picongpu/include/fields/FieldManipulator.hpp
@@ -1,5 +1,5 @@
 /**
- * Copyright 2013-2016 Axel Huebl, Rene Widera
+ * Copyright 2013-2016 Axel Huebl, Rene Widera, Richard Pausch
  *
  * This file is part of PIConGPU.
  *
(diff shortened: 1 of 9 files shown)


src/picongpu/include/fields/FieldManipulator.hpp
```diff
@@ -61,7 +61,8 @@ class FieldManipulator
                 uint32_t pos_or_neg = i % 2;
 
                 uint32_t thickness = ABSORBER_CELLS[direction][pos_or_neg];
-                float_X absorber_strength = ABSORBER_STRENGTH[direction][pos_or_neg];
+		PMacc::algorthm::math::Log log;
+                float_X absorber_strength = -1.0 * log( ABSORBER_STRENGTH[direction][pos_or_neg] ) thickness;
```
psychocoderHPC: A `*` before thickness ist missing

src/picongpu/include/fields/FieldManipulator.hpp
```diff
@@ -61,7 +61,8 @@ class FieldManipulator
                 uint32_t pos_or_neg = i % 2;
 
                 uint32_t thickness = ABSORBER_CELLS[direction][pos_or_neg];
-                float_X absorber_strength = ABSORBER_STRENGTH[direction][pos_or_neg];
+
+                float_X absorber_strength = -1.0 * math::log( ABSORBER_MAXFACTOR[direction][pos_or_neg] ) * thickness;
```
psychocoderHPC: This should be `-1.0 * math::log( ABSORBER_MAXFACTOR[direction][pos_or_neg] ) / thickness;` ??

PrometheusPi: @psychocoderHPC You are off course complete right with that - I made an error in deriving it in my lab book :sad:

PrometheusPi:
*offline discussion with* @psychocoderHPC:
the old and this implementation have the strongest damping first - we should check if an inital weaker damping is better.

PrometheusPi:
The exponential damping towards a total damping in the end seems to be counter productive.
It appears to be best to slowly reduce the field thus that it is still reflected on the boundary of the simulation box. In the current version, too low damping factors cause a close to total reflection at the outer boundary. Too high values (which would be favored  by this pull request) result in a reflection at the damping zone. The best damping factor in `dev` lies around `1e-2`. Which corresponds to a maximum damping of `0.7614` in this pull request. Since this appears to be non intuitive, I would favour the current `dev` implementation. 

index
 electromagnetic field energy evolution with different damping factors **p** as defined in `dev`.
simulated laser, each step (in E) is a "reflection/damping" at either y=0 or y=1024 boundary.

PrometheusPi:
The best damping is counter-intuitive with this new pull request - I will close this proposed code change.

PrometheusPi:
as discussed with @psychocoderHPC a slow initial damping which grows stronger towards the outer walls might be an improvement.

Issue #1700, opened 2016-12-14
extend help for disabled plugins


psychocoderHPC:
If a dependency for a plugin is not full filled than we disable the plugin.
We should keep all plugins avail and register a help that a plugin is disabled.

e.g.
```C++
./picongpu --help
....
--hdf5  plugin is disabled (NO SLPASH avail)
```

PrometheusPi:
Does this mean we enable all possible plugins (as e.g. the radiation plugin - which requires more memory per particle) by default in case all dependencies are fulfilled?

ax3l:
Good question.
Our current policy is: *compile* all plugins that have their dependencies fulfilled.

In your case which also will cost more resources: ideally you would compile radiation as soon as your libSplash dependency is covered but only require more memory if enabled at RT. Since we are not flexible enough for that yet... Use a `constexpr` switch in your param to explicitly enable :(

PR #1702, opened 2016-12-14
PngPlugin: help message if dependency is not compiled

psychocoderHPC:
If `pngwriter` is not available at compile time, the plugin `PngPlugin` register a help with information why the plugin can't use at runtime.

- PngCreator: add "tpp` file for `operator()`
- PngPlugin: add help message if `pngwriter` is not avail

This is the first plugin which full fill #1700

```bash
./picongpu --help
# ...
PositionsParticles: write position of one particle of a species to std::cout:
  --i_position.period arg               enable analyser [for each n-th step]

PngPlugin: create pngs of a species and fields:
  --i_png                               plugin disabled [compiled without 
                                        dependency pngwriter]
# ...
PositionsParticles: write position of one particle of a species to std::cout:
  --e_position.period arg               enable analyser [for each n-th step]

PngPlugin: create pngs of a species and fields:
  --e_png                               plugin disabled [compiled without 
                                        dependency pngwriter]
# ...
```

### DIFF
--- a/src/picongpu/include/plugins/PngPlugin.hpp
+++ b/src/picongpu/include/plugins/PngPlugin.hpp
@@ -71,11 +71,16 @@ namespace picongpu
 
         void pluginRegisterHelp(po::options_description& desc)
         {
+#if( PIC_ENABLE_PNG == 1 )
             desc.add_options()
                     ((analyzerPrefix + ".period").c_str(), po::value<std::vector<uint32_t> > (&notifyFrequencys)->multitoken(), "enable data output [for each n-th step]")
                     ((analyzerPrefix + ".axis").c_str(), po::value<std::vector<std::string > > (&axis)->multitoken(), "axis which are shown [valid values x,y,z] example: yz")
                     ((analyzerPrefix + ".slicePoint").c_str(), po::value<std::vector<float_32> > (&slicePoints)->multitoken(), "value range: 0 <= x <= 1 , point of the slice")
                     ((analyzerPrefix + ".folder").c_str(), po::value<std::vector<std::string> > (&folders)->multitoken(), "folder for output files");
+#else
+            desc.add_options()
+                    ((analyzerPrefix).c_str(), "plugin disabled [compiled without dependency PNGwriter]");
+#endif
         }
 
         void setMappingDescription(MappingDesc *cellDescription)
--- a/src/picongpu/include/plugins/output/images/PngCreator.hpp
+++ b/src/picongpu/include/plugins/output/images/PngCreator.hpp
@@ -22,26 +22,17 @@
 
 #include "pmacc_types.hpp"
 #include "simulation_defines.hpp"
-#include "simulation_types.hpp"
 
 #include <string>
-#include "mappings/simulation/GridController.hpp"
-
-#include <pngwriter.h>
-
 #include <iostream>
 #include <sstream>
-
 #include <iomanip>
 
-#include "memory/boxes/PitchedBox.hpp"
 #include "memory/boxes/DataBox.hpp"
 #include "plugins/output/header/MessageHeader.hpp"
 
-
 #include <boost/thread.hpp>
-//c includes
-#include <sys/stat.h>
+
 
 namespace picongpu
 {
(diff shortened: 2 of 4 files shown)


psychocoderHPC:
The only issu is that some plugin were generated for each species type and therefore write the message with missing dependency multiple time (see description example).
We can avoid this if the png plugin itself generate an instance for each species.

**We can solve each issue with a new abstraction :-)**

ax3l:
that's ok for me, it's as verbose as it is when enabled...

src/picongpu/include/plugins/output/images/PngCreator.hpp
```diff
@@ -136,75 +128,25 @@ namespace picongpu
 
     };
 
+} /* namespace picongpu */
+
+#if(PIC_ENABLE_PNG==1)
+#include "plugins/output/images/PngCreator.tpp"
```
ax3l: ah ah
```C++
#if( PIC_ENABLE_PNG == 1 )
#    include "plugins/output/images/PngCreator.tpp"
#else
```

;)

src/picongpu/include/plugins/PngPlugin.hpp
```diff
@@ -71,11 +71,16 @@ namespace picongpu
 
         void pluginRegisterHelp(po::options_description& desc)
         {
+#if(PIC_ENABLE_PNG==1)
```
ax3l: nobody does spaces :( `#if( PIC_ENABLE_PNG == 1 )`

src/picongpu/include/plugins/output/images/PngCreator.hpp
```diff
@@ -136,75 +128,25 @@ namespace picongpu
...
+#else
+namespace picongpu
+{
+    /** empty implementation if PIConGPU is compiled without PNGwriter
+     *
+     * If called this method will always fails with an exception at runtime.
+     */
     template<class Box>
     inline void PngCreator::createImage(
                                         const Box data,
```
ax3l: Y u still align so badly? ;)

src/picongpu/include/plugins/output/images/PngCreator.hpp
```diff
@@ -136,75 +128,25 @@ namespace picongpu
...
+} /* namespace picongpu */
+
+#if(PIC_ENABLE_PNG==1)
+#include "plugins/output/images/PngCreator.tpp"
+#else
+namespace picongpu
+{
+    /** empty implementation if PIConGPU is compiled without PNGwriter
+     *
+     * If called this method will always fails with an exception at runtime.
```
ax3l: typo: If called this method always fail with ...

src/picongpu/include/plugins/output/images/PngCreator.tpp
```diff
@@ -0,0 +1,110 @@
...
+
+#include "memory/boxes/DataBox.hpp"
+#include "plugins/output/header/MessageHeader.hpp"
+
+
+namespace picongpu
+{
+    template<class Box>
+    inline void PngCreator::createImage(
+                                        const Box data,
```
ax3l: Y u still align so badly? ;)

psychocoderHPC:
I updated this style issues.
I moved the assert for the case that `PNGwriter` is not loaded but the png creation method is called from `PngCreator.hpp` to `PngCreator.tpp`. This avoid copy past of the signature of `createImage(...)`

src/picongpu/include/plugins/output/images/PngCreator.tpp
```diff
@@ -0,0 +1,120 @@
...
+
+#if( PIC_ENABLE_PNG == 1 )
+#   include <pngwriter.h>
+#endif
+
+namespace picongpu
+{
+    template< class Box >
+    inline void PngCreator::createImage(
+        const Box data,
```
ax3l: this will cause unused variable warnings when PNGwriter is not available, *even if the user does not plan to use it*.

ax3l: either use
```C++
inline void PngCreator::createImage(
    const Box /*data*/,
    const Size2D /*size*/,
    const MessageHeader /*header/*
)
```
there or avoid calling that interface at RT.

ax3l: offline discussion: self-implement boost `unused`

psychocoderHPC: I realized that `createImage` is a host only function without any prefix, therefore we should not use an self implemented ignore function. For this case we should use `boost::ignore_unused`.
Never the less we need a equivalent function with the prefix `__host__ __device__` for other code parts.

psychocoderHPC:
updated

src/picongpu/include/plugins/output/images/PngCreator.hpp
```diff
@@ -136,75 +127,6 @@ namespace picongpu
...
-
-        png.settext( title, author.c_str(), description.str().c_str(), software);
-
-        // write to disk and close object
-        png.close();
-    }
-
 } /* namespace picongpu */
+
+#include "plugins/output/images/PngCreator.tpp"
```
ax3l: Can't we make this a regular `.cpp` file? `PngCreator` is template-free :)

src/picongpu/include/plugins/output/images/PngCreator.tpp
```diff
@@ -0,0 +1,122 @@
...
+         * scale_y is != 1.0
+         */
+        if( ( scale_x != float_X( 1.0 ) ) ||
+            ( scale_y != float_X( 1.0 ) )
+        )
+            //process the cell size and by factor scaling within one step
+            png.scale_kxky( scale_x, scale_y );
+
+        // add some meta information
+        //header.writeToConsole( std::cout );
```
ax3l: got ya! dead code :)

src/picongpu/include/plugins/output/images/PngCreator.hpp
ax3l: Can't we make this a regular `.cpp` file? `PngCreator` is template-free :)

src/picongpu/include/plugins/output/images/PngCreator.tpp
```diff
@@ -0,0 +1,122 @@
...
+        std::string filename( m_name + "_" + step.str( ) + ".png" );
+
+        pngwriter png( size.x( ), size.y( ), 0, filename.c_str( ) );
+
+        /* default compression: 6
+         * zlib level 1 is ~12% bigger but ~2.3x faster in write_png( )
+         */
+        png.setcompressionlevel( 1 );
+
+        //PngWriter coordinate system begin with 1,1
```
ax3l: ok, in that case lets add a space after `//` to be consistent

src/picongpu/include/plugins/output/images/PngCreator.tpp
BenjaminW3: You could try to add `BOOST_GPU_ENABLED` to [`boost::ignore_unused`](https://github.com/boostorg/core/blob/develop/include/boost/core/ignore_unused.hpp) like:
```
template <typename... Ts>
BOOST_FORCEINLINE BOOST_GPU_ENABLED BOOST_CXX14_CONSTEXPR void ignore_unused(Ts const& ...)
{}
```
If this would work and could be integrated into boost, alpaka would be happy too 😄

src/picongpu/include/plugins/output/images/PngCreator.hpp
psychocoderHPC: The method `createImage` used a template `Box`, therefore it is not possible.

src/picongpu/include/plugins/output/images/PngCreator.tpp
BenjaminW3: However, I do not currently have access to a Linux Box with an easily customizable Boost and CUDA, so I can not verify this beforehand. Could someone do this for me?

src/picongpu/include/plugins/output/images/PngCreator.hpp
ax3l: :'(

PR #1703, opened 2016-12-14
FieldTmp: Multiple Slots

ax3l:
Allows to use multiple `GridBuffers` ("slots") in `FieldTmp` for all operations done with it so far.

Required for, e.g., T_e and n_e in Thomas-Fermi Ionization.

Current control: the number of scalar temporary fields can be set in `memory.param` via:
```C++
constexpr uint32_t fieldTmpNumSlots = 1;
```

~~Still throws dozens of `__host__ __device__` warnings but might compile already.~~

Thanks to @psychocoderHPC for pair programming support.

### Usage

```C++
/* I want to access FieldTmp! */

/* And while we are defining this code, just saying: I will use two (2) slots of it,
 *   <<So, do you copy?! - ... Brzzzz >> */
PMACC_CASSERT_MSG(
    _please_allocate_at_least_two_FieldTmp_slots_in_memory_param,
    fieldTmpNumSlots >= 2
);
 
/* load each (scalar field) slot without copy data to host */
FieldTmp& density = dc.getData< FieldTmp >( FieldTmp::getUniqueId( 0 ), true );
FieldTmp& eneKin = dc.getData< FieldTmp >( FieldTmp::getUniqueId( 1 ), true );

/* reset density and kinetic energy values to zero */
density.getGridBuffer().getDeviceBuffer().setValue( FieldTmp::ValueType( 0.0 ) );
eneKin.getGridBuffer().getDeviceBuffer().setValue( FieldTmp::ValueType( 0.0 ) );
 
/* calculate and add the density values from all species */
ForEach<
    VectorAllSpecies, /* you want to use only e- species here */
    picongpu::detail::ComputeDensity< bmpl::_1, bmpl::int_< CORE + BORDER > >,
    MakeIdentifier< bmpl::_1 >
> computeDensity;
computeDensity( forward(&density), currentStep );

/* calculate and add the kinetic energy values from all species */
ForEach<
    VectorAllSpecies, /* you want to use only e- species here */
    picongpu::detail::ComputeEnergy< bmpl::_1, bmpl::int_< CORE + BORDER > >,
    MakeIdentifier< bmpl::_1 >
> computeEnergy;
computeEnergy( forward(&eneKin), currentStep );
 
/* add results of all species that are still in GUARD to next GPUs BORDER */
EventTask densityEvent = density.asyncCommunication( __getTransactionEvent() );
EventTask energyEvent = eneKin.asyncCommunication( __getTransactionEvent() );
__setTransactionEvent( fieldTmpEvent + energyEvent );

/* do work on density & eneKin */
// ...

/* release data */
dc.releaseData( FieldTmp::getUniqueId( 0 ) );
dc.releaseData( FieldTmp::getUniqueId( 1 ) );
```

### To Do

- [x] remove all new warnings
- [x] RT test in default slot 0
- [x] RT test with a slot number > 0

### DIFF
--- a/examples/Bunch/include/simulation_defines/param/memory.param
+++ b/examples/Bunch/include/simulation_defines/param/memory.param
@@ -53,4 +53,8 @@ constexpr uint32_t BYTES_EXCHANGE_Y = 6 * 512 * 1024; //6 MiB
 constexpr uint32_t BYTES_EXCHANGE_Z = 4 * 256 * 1024; //4 MiB
 constexpr uint32_t BYTES_CORNER = 8 * 1024; //8 kiB;
 constexpr uint32_t BYTES_EDGES = 32 * 1024; //32 kiB;
+
+/** number of scalar fields that are reserved as temporary fields */
+constexpr uint32_t fieldTmpNumSlots = 1;
+
 }//namespace picongpu
--- a/src/picongpu/include/particles/gasProfiles/FromHDF5Impl.hpp
+++ b/src/picongpu/include/particles/gasProfiles/FromHDF5Impl.hpp
@@ -21,6 +21,7 @@
 
 #pragma once
 
+#include "static_assert.hpp"
 #include "simulation_defines.hpp"
 #include "memory/buffers/GridBuffer.hpp"
 #include "memory/boxes/DataBoxDim1Access.hpp"
@@ -74,7 +75,12 @@ private:
     {
         using namespace splash;
         DataConnector &dc = Environment<>::get().DataConnector();
-        FieldTmp& fieldTmp = dc.getData<FieldTmp > (FieldTmp::getName(), true);
+
+        PMACC_CASSERT_MSG(
+            _please_allocate_at_least_one_FieldTmp_in_memory_param,
+            fieldTmpNumSlots > 0
+        );
+        FieldTmp& fieldTmp = dc.getData<FieldTmp >( FieldTmp::getUniqueId( 0 ), true );
         PMACC_AUTO(&fieldBuffer, fieldTmp.getGridBuffer());
 
         deviceDataBox = fieldBuffer.getDeviceBuffer().getDataBox();
--- a/src/picongpu/include/plugins/adios/ADIOSWriter.hpp
+++ b/src/picongpu/include/plugins/adios/ADIOSWriter.hpp
@@ -28,6 +28,7 @@
 #include <vector>
 
 #include "pmacc_types.hpp"
+#include "static_assert.hpp"
 #include "simulation_types.hpp"
 #include "plugins/adios/ADIOSWriter.def"
 
@@ -232,7 +233,11 @@ private:
             /*## update field ##*/
 
             /*load FieldTmp without copy data to host*/
-            FieldTmp* fieldTmp = &(dc.getData<FieldTmp > (FieldTmp::getName(), true));
+            PMACC_CASSERT_MSG(
+                _please_allocate_at_least_one_FieldTmp_in_memory_param,
+                fieldTmpNumSlots > 0
+            );
+            FieldTmp* fieldTmp = &(dc.getData<FieldTmp > (FieldTmp::getUniqueId( 0 ), true));
             /*load particle without copy particle data to host*/
             Species* speciesTmp = &(dc.getData<Species >(Species::FrameType::getName(), true));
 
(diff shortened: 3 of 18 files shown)


examples/Bunch/include/simulation_defines/param/memory.param
```diff
@@ -53,4 +53,8 @@ constexpr uint32_t BYTES_EXCHANGE_Y = 6 * 512 * 1024; //6 MiB
 constexpr uint32_t BYTES_EXCHANGE_Z = 4 * 256 * 1024; //4 MiB
```
n01r: A little off-topic but just because I'm seeing it right now: in many `memory.param` files of the different examples there are different sizes of memory that need to be reserved for exchange but the comments often read the same numbers.
Just for consistency we could remove the comments there and state in the beginning that the unit used there is *Byte*.
Here at this point it would be 1 rather than 4 MiBs, right?

n01r:
Great! :smile:  :sparkles:

src/picongpu/include/particles/gasProfiles/FromHDF5Impl.hpp
```diff
@@ -74,7 +75,12 @@ struct FromHDF5Impl : public T_ParamClass
     {
         using namespace splash;
         DataConnector &dc = Environment<>::get().DataConnector();
-        FieldTmp& fieldTmp = dc.getData<FieldTmp > (FieldTmp::getName(), true);
+
+        PMACC_CASSERT_MSG(
```
n01r: I like how with this implementation some overhead from the `Particles` class is removed.
Do I understand correctly, though, that every time that I would instatiate a `FieldTmp` object I would have to add this `cassert`, right? I guess this could not just be moved into the constructor?

examples/Bunch/include/simulation_defines/param/memory.param
n01r: I can do that if you want, after yours is through - if you don't wanna rebase. But since this is such a minor change which only affects the comments I'd also be fine if it happened in the same PR.

There were some more typos in parts of the code that you didn't specifically touch in the files. [Here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1703/files#diff-685d4f0ffd9e2a0bfe7283c093e47544R156) to be precise, and in the comment below where it says *negativ* instead of *negative*. But a separate commit would be enough for this.

src/picongpu/include/particles/gasProfiles/FromHDF5Impl.hpp
ax3l: every time you want to run `dc.getData` you assume the slot you access is really allocated in the current simulation. the assert checks that the user did not fool you on that and that `getData()` will not throw at RT.

examples/Bunch/include/simulation_defines/param/memory.param
ax3l: yep, not touched, saw those too.

n01r:
thanks for the lovely usage example :+1:

src/picongpu/include/plugins/adios/ADIOSWriter.hpp
```diff
@@ -259,7 +264,7 @@ class ADIOSWriter : public ILightweightPlugin
                        getName(),
                        fieldTmp->getHostDataBox().getPointer());
 
-            dc.releaseData(FieldTmp::getName());
+            dc.releaseData( FieldTmp::getName() + "0" );
```
psychocoderHPC: please use `FieldTmp::getUniqueId( 0 )`

src/picongpu/include/plugins/hdf5/WriteFields.hpp
```diff
@@ -210,7 +215,7 @@ class WriteFields<FieldTmpOperation<Solver, Species> >
                           fieldTmp->getHostDataBox(),
                           ValueType());
 
-        dc.releaseData(FieldTmp::getName());
+        dc.releaseData( FieldTmp::getName() + "0" );
```
psychocoderHPC: please use `FieldTmp::getUniqueId( 0 )`

ax3l:
@psychocoderHPC review comments updated! :)

@n01r if you want, you can write a kernel using two FieldTmp quantities now for testing the infrastructure is ready for TF.

ax3l:
RT test: oh oh, segfaults with 2 slots enabled and standard I/O of FieldTmp quantities....

n01r:
Thx for the update! I will rebase on that and continue implementing.

PR #1707, opened 2016-12-16
use new host device warning suppressor

psychocoderHPC:
- use `#pragma nv_exec_check_disable` instead of `#pragma hd_warning_disable`
- `nv_exec_check_disable` is available since CUDA 7.5

PMacc only supports CUDA 7.5+ therefore we can fully switch to the new flag.

There is no documentation about the old and the new flag available. [thrust](https://github.com/thrust/thrust/blob/8e35dce3d2003f9e9f7eed8dac31be7f050c8b2d/thrust/detail/config/exec_check_disable.h) and [VTK-m](http://m.vtk.org/documentation/ExportMacros_8h_source.html) changed the flag from `hd_warning_disable` to `nv_exec_check_disable`. 
From the [thrust pull request](https://github.com/thrust/thrust/pull/680) it looks like the new prgma can also suppress error messages.

### DIFF
--- a/src/libPMacc/include/pmacc_types.hpp
+++ b/src/libPMacc/include/pmacc_types.hpp
@@ -77,7 +77,7 @@ typedef long long int int64_cu;
 #define PMACC_GLOBAL_KEYWORD __location__(global)
 
 /*
- * Disable nvcc warning:
+ * Disable nvcc warnings and errors:
  * calling a __host__ function from __host__ __device__ function.
  *
  * Usage:
@@ -90,8 +90,8 @@ typedef long long int int64_cu;
  * WARNING: only use this method if there is no other way to create runable code.
  * Most cases can solved by #ifdef __CUDA_ARCH__ or #ifdef __CUDACC__.
  */
-#if defined(__CUDACC__)
-#define PMACC_NO_NVCC_HDWARNING _Pragma("hd_warning_disable")
+#if defined(__CUDACC__) && !(defined(__CUDA__) && defined(__clang__))
+#define PMACC_NO_NVCC_HDWARNING _Pragma("nv_exec_check_disable")
 #else
 #define PMACC_NO_NVCC_HDWARNING
 #endif


ax3l:
Behavior:
```
warning:
  calling a __host__ function("h") from a __host__ __device__ function("d") is not allowed

error:
calling a __device__ function("d2") from a __host__ __device__ function("d") is not allowed
```

Question: will it silently not create ptx code for those?

psychocoderHPC:
Test Code:

```C++
#include <cuda_runtime.h>
#include <cstdio>

#if( CASE == 1 )
#   define X_PREFIX __host__ __device__
#   define Y_PREFIX __host__
#   define SUSPRESSION _Pragma("hd_warning_disable")
#elif( CASE == 2 )
#   define X_PREFIX __host__ __device__
#   define Y_PREFIX __device__
#   define SUSPRESSION _Pragma("nv_exec_check_disable")
#elif( CASE == 3 )
#   define X_PREFIX __device__
#   define Y_PREFIX __host__
#   define SUSPRESSION _Pragma("nv_exec_check_disable")
#elif( CASE == 4 )
#   define X_PREFIX __host__
#   define Y_PREFIX __device__
#   define SUSPRESSION _Pragma("nv_exec_check_disable")
#endif

Y_PREFIX void y(int* a)
{
    *a=2;
    printf("y=%i ",1);
}

SUSPRESSION
X_PREFIX void x(int* a)
{
        printf("x=%i ",1);
        y(a+1);
        *a=1;
}
__global__ void kernel(int * a)
{
#if( CASE != 4 )
        x(a);
#endif
}

int main ()
{
        int a[2];
        printf("GPU< ");
        kernel<<<1,1>>>(a);
        cudaDeviceSynchronize();
        printf(">\n");
        printf("CPU< ");
        x(a);
        printf(">\n");
        return 0;
}
```

Run Tests:
```bash
for i in `seq 1 4` ;  do echo "compile case $i"; nvcc -Wno-deprecated-gpu-targets main.cu -DCASE="$i"; ./a.out;  done
```

Result:
```bash
compile case 1
GPU< x=1 >  # removed y() call 
CPU< x=1 y=1 >

compile case 2
GPU< x=1 y=1 >
CPU< x=1  # crashed on CPU

compile case 3 # this case does not compile
main.cu(50): error: calling a __device__ function("x") from a __host__ function("main") is not allowed

1 error detected in the compilation of "/tmp/tmpxft_00002d3e_00000000-5_main.cpp4.ii".
GPU< x=1 y=1 >
CPU< x=1 # crashed on CPU

compile case 4
GPU< > # GPU disabled in kernel call of __host__ from __global__ not allowed
CPU< x=1   # crashed on cpu
```

psychocoderHPC:
I close this pull request because  `nv_exec_check_disable` is a new feature with a new behavior. Currently we need only to suppress the `CASE 1` (possible with `hd_warning_disable`)

psychocoderHPC:
Maybe it is not possible to use the suppression commands together with `clang` in this case we need to define `PMACC_NO_NVCC_HDWARNING` to empty

ax3l:
... and can't use code that would rely on the *error suppression* at all!

psychocoderHPC:
I checked the example with **clang** and find out that non of the test cases compile.

Clang also includes a a [semantic checking file](https://github.com/llvm-mirror/clang/blob/master/test/SemaCUDA/function-target.cu) were all allowed and forbidding combination are visible.

PR #1709, opened 2016-12-16
Param: Type List Syntax

ax3l:
Use a more natural `using Sequence = MakeSeq_t< 1, 2, 3 >;` syntax for user input of type lists via `MakeSeq_t` and C++11 `using`.

Avoids boiler plate code without `=` and additional `::type` access:
`typedef Seq< 1, 2, 3 >::type List;`

Cleaned up syntax of `speciesDefinition` on the way.

### DIFF
--- a/examples/SingleParticleTest/include/simulation_defines/param/speciesDefinition.param
+++ b/examples/SingleParticleTest/include/simulation_defines/param/speciesDefinition.param
@@ -30,40 +30,23 @@
 #include "particles/Particles.hpp"
 #include <boost/mpl/string.hpp>
 
-#include "particles/ionization/byField/ionizers.def"
-
 namespace picongpu
 {
 
 /*########################### define particle attributes #####################*/
 
 /** describe attributes of a particle*/
-typedef MakeSeq<position<position_pic>, momentum, weighting>::type DefaultParticleAttributes;
-
-
-/** \todo: not nice, we change this later with nice interfaces
- * Plugins should add required attributes
- */
-
-/*add old momentum for radiation plugin*/
-typedef MakeSeq<
+using DefaultParticleAttributes = MakeSeq_t<
+    position<position_pic>,
+    momentum,
+    weighting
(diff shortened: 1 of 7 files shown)


src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -169,52 +154,52 @@ struct Hydrogen
...
+>;
 
 /*define specie ions*/
-typedef Particles<
+using PIC_Ions = Particles<
     ParticleDescription<
-        bmpl::string<'i'>,
+        bmpl::string< 'i' >,
         SuperCellSize,
-        DefaultAttributesSeq,
```
ax3l: ha, nice bug in `PluginController.hpp`: radiation assumes *all species* have momentumPrev1.

psychocoderHPC: Right the reason is that the pusher extention can not differ between species with and without `momentumPrev1`.
To change this wee need to cleanup the pusher extension first.

ax3l: yep, will keep the "oh, ions also need all radiation flags" setting for now

examples/SingleParticleTest/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -90,76 +73,24 @@ typedef bmpl::vector<
...
 
 /*########################### end species ####################################*/
 
-
-/*! we delete this ugly definition of VectorAllSpecies after all picongpu components
- * support multi species */
-/** \todo: not nice, but this should be changed in the future*/
-typedef MakeSeq<
+using VectorAllSpecies = MakeSeq_t<
+    MakeSeq_t<
```
psychocoderHPC: Please remove one `MakeSeq_t`

ax3l: hups, done!

PR #1711, opened 2016-12-16
simplify species definition

psychocoderHPC:
This removes the complicated interface to define a species away from the PIConGPU user. Objects like `ParticleDesction` and `SuperCellSize` are hard to handle for the user.

- refactor `Particle<>` interface
- remove `SuperCellSize` and `ParticleDesciption<>` from PIConGPU user interface
- remove unused includes

### DIFF
--- a/examples/LaserWakefield/include/simulation_defines/param/speciesDefinition.param
+++ b/examples/LaserWakefield/include/simulation_defines/param/speciesDefinition.param
@@ -26,13 +26,9 @@
 #include "simulation_defines.hpp"
 #include "particles/Identifier.hpp"
 #include "compileTime/conversion/MakeSeq.hpp"
-#include "dimensions/DataSpace.hpp"
-#include "identifier/identifier.hpp"
-#include "identifier/alias.hpp"
 #include "identifier/value_identifier.hpp"
 
 #include "particles/Particles.hpp"
-#include "particles/ParticleDescription.hpp"
 #include <boost/mpl/string.hpp>
 
 #include "particles/ionization/byField/ionizers.def"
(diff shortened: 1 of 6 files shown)


ax3l:
that's ok, I will rebase against your PR!

psychocoderHPC:
OK I removed the checkpoint flag in my desctiption

ax3l:
@psychocoderHPC oh nooo, compile failure

ax3l:
I'll push an update to your PR

psychocoderHPC:
@ax3l thx for the fix

ax3l:
@psychocoderHPC I would generally extent the operations that `Particles` does before continuing to derive from `ParticlesBase`.

Especially, we have access to `T_Attributes` and `T_Flags` at this point and can ask all `T_Flags` if their `T_Attribute` requirements are already fulfilled and *if not*, we add those to the list.

psychocoderHPC:
Yes god idea :+1: , we also extend the attribute sequence in `ParticlesBase`. But this should be a follow up pull request.
We can collect all attributes and afterward we can call unique on the sequence and pass it to `ParticlesDescription`.

psychocoderHPC:
Maybe we can get all attributes out of the flags. If so the user needs only to define the flags (solver, const attributes, ...) and all dynamic attributes are created from the needs of the flags.

ax3l:
yep, that's the idea. I would still allow adding additional attributes for runtime purposes (e.g., no solver or plugin will *require* the particleID but it is still useful). So flags should define the *absolute minimum* of what will be used.

n01r:
What you did here @psychocoderHPC was to make use of the C++11 feature [type alias](http://en.cppreference.com/w/cpp/language/type_alias), correct?

psychocoderHPC:
@n01r No I am not using any type alias. I removed the complicate interface (ParticleDescription) from the end user.

n01r:
Ohhh, right, that happened in a different pull request (#1709). There you replaced many
```c++
typedef <Something> <Identifier> ...
```
by
```c++
using <Identifier> = <Something>
```

My mistake ^^

PR #1714, opened 2016-12-17
Maxwell solver, use curl instance

psychocoderHPC:
The interface change is transparent for the PIConGPU user.

- add curl instance to kernel parameter
- add `T_` to template names
- remove template keyword  `class` with `typename`

This change allow  to move constant curl calculations like `sin()` in the Lehe solver to the host side.

### DIFF
--- a/src/picongpu/include/fields/MaxwellSolver/Yee/YeeSolver.hpp
+++ b/src/picongpu/include/fields/MaxwellSolver/Yee/YeeSolver.hpp
@@ -75,9 +75,13 @@ private:
                 > BlockArea;
 
         AreaMapping<AREA, MappingDesc> mapper(m_cellDescription);
-        PMACC_KERNEL(KernelUpdateE<BlockArea, CurlB>{})
-                (mapper.getGridDim(), SuperCellSize::toRT())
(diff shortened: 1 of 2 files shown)


src/picongpu/include/fields/MaxwellSolver/Lehe/LeheCurl.hpp
```diff
@@ -42,53 +42,62 @@ namespace picongpu
...
             template<class Memory >
             HDINLINE typename Memory::ValueType operator( )(const Memory & mem ) const
             {
                 /* Distinguished direction where the numerical Cherenkov Radiation
                  * of moving particles is suppressed.
                  */
-                const float_X isDir_x = float_X( 1.0 );
-                const float_X isDir_y = float_X( 0.0 );
-                const float_X isDir_z = float_X( 0.0 );
+                constexpr float_X isDir_x = float_X( 1.0 );
```
ax3l: did you see the registers decrease in this kernel? :)

ax3l:
I think both of your PRs accidentally contain both commits ;)

psychocoderHPC:
Yes pushed to much commits to this branch. This branch should only contain the interface change.

PR #1715, opened 2016-12-17
Lehe solver: optimize performance

psychocoderHPC:
- move sinus calculation to host side
- use 64bit instead of `float_X` to calculate sinus 
- add `constexpr` to constant calculations

Dependencies:
--------------------
- [x] rebase against ~~#1714~~
- [x] runtime tests

### DIFF
--- a/src/picongpu/include/fields/MaxwellSolver/Lehe/LeheCurl.hpp
+++ b/src/picongpu/include/fields/MaxwellSolver/Lehe/LeheCurl.hpp
@@ -42,53 +42,62 @@ namespace picongpu
             typedef PMacc::math::CT::Int< 1, 1, 1 > LowerMargin;
             typedef PMacc::math::CT::Int< 2, 2, 2 > UpperMargin;
 
+            float_X mySin;
+
+            HINLINE CurlELehe( )
+            {
+                mySin = float_X(
+                    math::sin(
+                        float_64( 0.5 ) *
+                        float_64( M_PI ) *  float_64( SPEED_OF_LIGHT ) *
+                        float_64( DELTA_T ) / float_64( CELL_WIDTH )
+                    )
+                );
+            }
+
             template<class Memory >
             HDINLINE typename Memory::ValueType operator( )(const Memory & mem ) const
             {
                 /* Distinguished direction where the numerical Cherenkov Radiation
                  * of moving particles is suppressed.
                  */
-                const float_X isDir_x = float_X( 1.0 );
-                const float_X isDir_y = float_X( 0.0 );
-                const float_X isDir_z = float_X( 0.0 );
+                constexpr float_X isDir_x = float_X( 1.0 );
+                constexpr float_X isDir_y = float_X( 0.0 );
+                constexpr float_X isDir_z = float_X( 0.0 );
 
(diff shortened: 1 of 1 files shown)


src/picongpu/include/fields/MaxwellSolver/Lehe/LeheCurl.hpp
```diff
@@ -176,59 +185,66 @@ namespace picongpu
...
             template<class Memory >
             HDINLINE typename Memory::ValueType operator( )(const Memory & mem ) const
             {
                 /* Distinguished direction where the numerical Cherenkov Radiation
                  * of moving particles is suppressed.
                  */
-                const float_X isDir_x = float_X( 0.0 );
-                const float_X isDir_y = float_X( 1.0 );
-                const float_X isDir_z = float_X( 0.0 );
+                constexpr float_X isDir_x = float_X( 0.0 );
```
ax3l: imo, we should move this out of the device code all together and pass it as a parameter to the kernel

ax3l: ah wait, the sine is outside now and those constexpr should work. lets check the registers if they pop :)

psychocoderHPC:
run-time test:
- example LaserWakefield
- disable all species
- global simulation size: 256x512x256 cells
- gpus 1x4x1 (moving window enabled)
- propagate laser 3500 steps (23 slides)

I compered each component of field E and B of the current dev branch against this pull request.
**All values are equal .**

ax3l:
Excellent, thanks!

psychocoderHPC:
Register count: There are no difference in used register on GPU. (checked with CUDA 8)

I think `nvcc` pre-computed all constant variable at compile time (also in the old version)

src/picongpu/include/fields/MaxwellSolver/Lehe/LeheCurl.hpp
```diff
@@ -42,53 +42,62 @@ namespace picongpu
             typedef PMacc::math::CT::Int< 1, 1, 1 > LowerMargin;
             typedef PMacc::math::CT::Int< 2, 2, 2 > UpperMargin;
 
+            float_X mySin;
+
+            HINLINE CurlELehe( )
```
PrometheusPi: why are the two implementations of `CurlELehe( )`?
here and line 190 of **this file**

src/picongpu/include/fields/MaxwellSolver/Lehe/LeheCurl.hpp
```diff
@@ -176,59 +185,66 @@ namespace picongpu
...
```
PrometheusPi: why are the two implementations of `CurlELehe( )`?
here and line 47 of **this file**

src/picongpu/include/fields/MaxwellSolver/Lehe/LeheCurl.hpp
PrometheusPi: OK - it is due to the hidden template specialization

psychocoderHPC:
rebased

PR #1716, opened 2016-12-18
Single Particle Tests: Use Standard MySimulation

ax3l:
Due to limited support of initializing hand-picked (single) particles into a simulation and background fields in our early days (open alpha release), PIConGPU tests that initialize and propagate single particles for tests such as pusher, current deposition, heads-on Thomson scattering etc. did use their own "starter" which implements `MySimulation` again for those needs.

Nowadays (tm) all that functionality is standard in PIConGPU and we can benefit from the whole function pack in those tests, too. This PR will therefore clean up the unnecessary code and improve maintainability and extensibility.

### To Do

- [x] if RT test is successful, move the new `StartPostion` "OnePosition" to it's own PR ~~#1753~~
- [x] refactor, unite or remove `SingleParticleCurrent` and `SingleParticleRadiationWithLaser`, too (take care compile-time feature test-coverage stays the same)
 - [x] Current: [CheckCurrent](https://github.com/ComputationalRadiationPhysics/picongpu/blob/4d37b8904a5ab150cb98b235e35080f63a25a6f2/examples/SingleParticleCurrent/include/CheckCurrent.hpp) is in history and can be written as a few liner with python from HDF5 output if necessary (but [we](https://github.com/ComputationalRadiationPhysics/picongpu/blob/4d37b8904a5ab150cb98b235e35080f63a25a6f2/src/tools/bin/plot_chargeConservation.py) have [better](https://github.com/ComputationalRadiationPhysics/picongpu/blob/4d37b8904a5ab150cb98b235e35080f63a25a6f2/src/tools/bin/plot_chargeConservation_overTime.py) tests now)
- [x] RadiationWithLaser: Implement a single particle as alternative `cmakeFlags` density in `Bunch` #1755

### SingleParticle Test: Little Gyro Spin

Modifies the single particle test to make a little spin in 2 periods during 100 steps with relativistic momentum. At the same time, the particle is accelerated non-relativistically by about 0.005 beta in orthogonal direction with an E-Field.

Can be used to validate pushers, current, charge, radiation or photon creation. Does only need a single GPU and creates just around ~1GB output which can be compressed to about 37 MB.

Currently no tools added alongside, since I am waiting for a small bug in yt to be fixed (and also for our new CI node).

So long, [here](https://gist.github.com/ax3l/9b49780a1bbedbe054f7d98654644e71/128a383a55ee62e342028b0744d72b3e40b60c8e) is an example script that does not look nice yet:
```python
import opmd_viewer
import numpy as np
# %matplotlib notebook

ts3 = opmd_viewer.OpenPMDTimeSeries("singleParticle-004/simOutput/h5/")

pos_x = []
pos_y = []
pos_z = []
mom_z = []
t = []

for t_it in ts3.iterations:
    tmp = ts3.get_particle(["x", "y", "z", "uz"], "e", iteration=t_it)
    pos_x.append(tmp[0][0])
    pos_y.append(tmp[1][0])
    pos_z.append(tmp[2][0])
    mom_z.append(tmp[3][0])
    t.append(ts3.current_t)

plt.figure()
plt.plot(pos_x, pos_y, "x", label="pos_xy")
plt.legend()
plt.show()

plt.figure()
plt.plot(pos_z, label="z")
plt.legend()
plt.show()

plt.figure()
plt.plot(mom_z, label="mom_z")
plt.show()

plt.figure()
plt.plot(t, label="t")
plt.show()
```

There are other tests that I like more when we add a RT test analysis, e.g. a drift in an inhomogeneous B field.


ax3l:
@PrometheusPi @Heikman hey reviewers, can you please take a look and merge me? :)

ax3l:
offline discussion results:

> How would you deactivate the background fields easily?

yes, just set the member `bool` flags in the classes to `false`.

> is this not another pull request ?

no, those examples are fully removed now and can be tested in Bunch (which was indeed a separate PR)

PR #1719, opened 2016-12-19
ForEach: remove boost preprocessor

psychocoderHPC:
This change is backwards compatible (no interface change)

- use variadic templates to remove boost preprocessor
- reformat with new coding style
- cleanup includes

The documentation of the class `ForEach` is not touched.

### DIFF
--- a/src/libPMacc/include/algorithms/ForEach.hpp
+++ b/src/libPMacc/include/algorithms/ForEach.hpp
@@ -26,224 +26,169 @@
 #include "compileTime/accessors/Identity.hpp"
 #include "forward.hpp"
 
-#include <boost/preprocessor/repetition/enum.hpp>
-#include <boost/preprocessor/repetition/enum_params.hpp>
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/algorithms/ForEach.hpp
```diff
@@ -26,224 +26,169 @@
...
+            end,
+            end
+        > Functor;
+
+        PMACC_NO_NVCC_HDWARNING
+        template< typename ... T_Types >
+        HDINLINE void
+        operator( )( const T_Types& ... ts ) const
+        {
+            Functor()( getForwardedValue( ts )... );
```
ax3l: I only see here are missing spaces ^^

theZiz:
I remember a PIConGPU developers meeting, in which we decided not to change the indentation and the code itself in the same commit / pull request. ;)

ax3l:
Yep, but basically he refactored the whole code. I did just read the full new file ;)

Issue #1720, opened 2016-12-20
wavepacket laser not working


PrometheusPi:
Currently  the wavepacket laser is not working. It either produces NaNs or some non-physical propagation.

ax3l:
weird, can you post your `gridConfig`, `componentsConfig` and `laserConfig` + your grid size and steps?

ax3l:
also: you title says wavepacket is broken, your description says plane wave is broken :)

psychocoderHPC:
Is this issu fixed with#1722?

ax3l:
yes, it's even backported to the latest stable release :)

PR #1722, opened 2016-12-20
add missing minus sign wavepacket laser transversal

PrometheusPi:
This fixes issue #1720.

Since [December 2015](https://github.com/ComputationalRadiationPhysics/picongpu/commit/0acbd9050fdb4af40aa14434d3cdc11e8a862222) we fixed the implementation of the  transversely asymmetric envelope of the `laserWavepacket`. Due to this change a minus disappeared, which causes different (see issue) numerical artifacts. 

 - [x] test new version


ax3l:
Omg, how did we do that. :speak_no_evil: A lot of people seem to use wavepacket lasers - nad ;)

PR #1725, opened 2016-12-20
add `Array<>` with static size

psychocoderHPC:
add C++11 STL like static array `Array<>`

Currently it is not possible to use `std::array<>` on the device side. We need the functionality for the up coming shared memory changes.

### DIFF
--- a/src/libPMacc/include/memory/Array.hpp
+++ b/src/libPMacc/include/memory/Array.hpp
@@ -1,0 +1,128 @@
+/**
+ * Copyright 2016 Rene Widera
+ *
+ * This file is part of libPMacc.
+ *
+ * libPMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * libPMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with libPMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/memory/Array.hpp
```diff
@@ -0,0 +1,128 @@
...
+         *
+         * @param value element assigned to each member
+         */
+        HDINLINE Array( T_Type const & value )
+        {
+            for( size_type i = 0; i < size(); ++i )
+                m_data[ i ] = value;
+        }
+
+        /** get N-st value
```
ax3l: I think it's `N-th` value

src/libPMacc/include/memory/Array.hpp
```diff
@@ -0,0 +1,128 @@
...
+        {
+            return m_data;
+        }
+        /** @} */
+
+        /** default constructor
+         *
+         * the default constructor of each member is called
+         */
+        HDINLINE Array() = default;
```
ax3l: that's a thing now? :-o

src/libPMacc/include/memory/Array.hpp
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        /** get the direct access to the data
+         *
+         * @{
+         */
+        HDINLINE
+        pointer data( )
+        {
+            return data;
```
ax3l: What is happening here? What `data` is meant? I am confused.

psychocoderHPC: bug: this should be m_data

src/libPMacc/include/memory/Array.hpp
psychocoderHPC: This is C++11 and equal to 
```C++
 HDINLINE Array()
{
}
```

psychocoderHPC:
@ax3l updated

src/libPMacc/include/memory/Array.hpp
```diff
@@ -0,0 +1,128 @@
...
+         *
+         * @{
+         */
+        template< typename T_Idx >
+        HDINLINE
+        const_reference
+        operator[]( T_Idx const idx ) const
+        {
+            return m_data[ idx ];
+        }
```
psychocoderHPC: The current implementation give the compiler the freedom to optimize.
e.g. maybe the array can addressed with `int` and the user called this operator with an integer, than no cast is needed.

Issue #652, opened 2015-01-29
Collection of CUDA Bugs and Work Arounds


ax3l:
@psychocoderHPC let us please collect all bugs that we noticed in PIConGPU with CUDA.

_Currently and CUDA 6.5 specific:_
In case they did all not show up before (CUDA 5.0-6.0) and are fixed in CUDA 7.0(rc) we might have to consider to either force CUDA 7.0+ or to _exclude_ the 6.5 support at all.
### Currently applied work-arounds in PIConGPU for CUDA Bugs

**CUDA 5.5+**
- #475 `-(-static_cast<>(...))` in loops, [forum](https://devtalk.nvidia.com/default/topic/752200/cuda-programming-and-performance/nvcc-loop-bug-since-cuda-5-5/) -> **upsteam bug report number? fixed in 6.0+?** -> work-arounds removed (refactored to `short-if`'s) in #680
- #538 #539 #2831 ? zero if not `volatile` bug -> **needs upstream bug report!**
- #401 unique id for (external?) `__shared__` memory required -> **needs upstream bug report!**
- #199 (external?) `__shared__` memory can not be created by template -> **needs upstream RFE!**

**CUDA 6.5**
- #570 `numBins *= int( numBins > 0);` bug, we found `short ifs ?:;` are efficient, too (no workarounds applied but refactored the code) -> **nvidia bug report number 1604407** (@psychocoderHPC)

**CUDA 6.5-7.0rc**
- #655 loop optimization bug -> not (yet) reproduce able with a minimal example ->~~needs upstream bug report~~ (solved in 7.0)

**CUDA 6.5 & 7.0**
- spotted alignment issues with pointer members in structs (sort them last to fix) #762 #1579

**CUDA 6.5+**
- constexpr function calls with constexpr is not working (@flamefire upstream bug report #1659894)

**CUDA 7.0**
- C++11 usage of `decltype` in [specific templates](https://stackoverflow.com/questions/31969644/compilation-error-with-nvcc-and-c11-need-minimal-failing-example) is broken, e.g., visible in `boost::result_of` in decltype mode (no define BOOST_RESULT_OF_USE_TR1) -> boost [work-around](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1151/files#diff-baaba04b0773a6bcf8a6d1c04ee53f5f) applied #1151
- `a`, `b` and `c` are reserved https://github.com/ComputationalRadiationPhysics/mallocMC/pull/108

[...]

PR #1726, opened 2016-12-20
add shared memory allocator

psychocoderHPC:
The user can allocate static sized shared memory without using the cuda prefix `__shared__`.

- add namespace `PMacc::memory::shared`
- add class `Allocate<>`
- add helper functions `allocate()`
- add macro to allocate shared memroy `PMACC_SMEM()`

This pull request allow an easier switch to `cupla` and minimize the up coming changes.

### DIFF
--- a/src/libPMacc/include/memory/shared/Allocate.hpp
+++ b/src/libPMacc/include/memory/shared/Allocate.hpp
@@ -1,0 +1,111 @@
+/**
+ * Copyright 2016 Rene Widera
+ *
+ * This file is part of libPMacc.
+ *
+ * libPMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * libPMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with libPMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+
+#include "pmacc_types.hpp"
+
+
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/memory/shared/Allocate.hpp
```diff
@@ -0,0 +1,111 @@
...
+} // namespace shared
+} // namespace memory
+} // namespace PMacc
+
+/** allocate shared memory
+ *
+ * @param varName name of the variable
+ * @param ... type of the variable
+ */
+#define PMACC_SMEM( varName, ... ) auto & varName = PMacc::memory::shared::allocate< __COUNTER__, __VA_ARGS__ >( )
```
ax3l: with this class, can we finally create a minimal example demonstrating that CUDA bug?

psychocoderHPC: which bug do you mean?

ax3l:
The bug that we even need a counter in multiple shared mem usages.

psychocoderHPC:
thats not a bug that is a feature :-)
Shared  memory allocations within a member function are more than less equal to a static member within a class.
Thats the reason why we need a unique id to create different shared memory variables from the some factory `Allocate<>`.
As I know this is not documented in CUDA but possible to see in the ptx code.

src/libPMacc/include/memory/shared/Allocate.hpp
ax3l: offline discussion: we need the macro since otherwise the user would have to write `__COUNTER__` which is a pre-compiler macro giving a "unique" number, in user code. we use that to prevent that N>1 shared memory usages in the same function are identified as one (since CUDA has a long-standing bug/"feature" not to differentiate those by name but by type... https://github.com/ComputationalRadiationPhysics/picongpu/issues/652)

ax3l:
Questions answered offline two days ago, @psychocoderHPC can you just write the left-over answer to his inline comments again so we document it?

src/libPMacc/include/memory/shared/Allocate.hpp
```diff
@@ -0,0 +1,111 @@
...
+        typename T_Type
+    >
+    DINLINE T_Type&
+    allocate( )
+    {
+        return Allocate<
+            T_uniqueId,
+            T_Type
+        >::get( );
+    }
```
psychocoderHPC: This give use the possibility to partially specialize the allocator if needed.

PR #1045, opened 2015-08-14
add trait  GetDefaultConstructibleType

psychocoderHPC:
- define new trait to get shared memory compatible types

It is not allowed that a object in cuda shared memory initialize any member in it's default constructor.
This trait returned a shared memory compatible type.

This pull request shipped only the definition of the trait, at the moment there is no specialized type that can used with the new trait.

PR #1727, opened 2016-12-21
Topic use remove shared keyword

psychocoderHPC:
- use PMacc macro `PMACC_SMEM` instead of CUDA native keyword `__shared__`
- remove usage of `GetEmptyDefaultConstructibleType<>` to workaround nvcc warnings

Commits from ~~#1726~~ and ~~#1725~~ are included in this pull request and would be deleted after the rebase.

The trait "GetEmptyDefaultConstructibleType<>" was introduced with #1045 to suppress nvcc warning if   a class with a non empty constructor is stored in shared memory. This is not longer the case if ~~#1726~~ is used to create variables in the shared memory. I removed the use of the trait because I need to touch all lines with the usage to use the new `PMACC_SMEM()` macro.
I will remove the trait `GetEmptyDefaultConstructibleType<>` in an separate pull request.

Not changed
-----------------
- The keyword `__shared__` in `SharedMemoryBox` and `SharedMemAllocator` are not removed. I will refactor this part to a more clean implementation in a follow up pull request.
- extern shared memory is not touched

dependencies
------------------

- [x] depends on ~~#1726~~
- [x] depends on ~~#1725~~
- [x] runtime KHI test

### DIFF
--- a/src/libPMacc/include/particles/ParticlesBase.kernel
+++ b/src/libPMacc/include/particles/ParticlesBase.kernel
@@ -33,6 +33,8 @@
 #include "particles/operations/Deselect.hpp"
 #include "traits/NumberOfExchanges.hpp"
 #include "nvidia/atomic.hpp"
+#include "memory/shared/Allocate.hpp"
+#include "memory/Array.hpp"
 
 namespace PMacc
 {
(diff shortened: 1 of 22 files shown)


ax3l:
@psychocoderHPC please rebase :)

psychocoderHPC:
rebased

PR #1729, opened 2017-01-02
CMake: allow definition of multiple architectures

psychocoderHPC:
Allow that multiple device architecture code is generated and embedded into the binary.
This reduce the start time of the binary if `isaac` live visualization is enabled.
The CMake variable `CUDA_ARCH` **does not** allow the prefix `sm_` anymore.

- define `CUDA_ARCH` as string list
- generate embedded SAS code for each user defined architecture
- `configure`: update documentation
- update all `cmakeFlags` and `picongpu.profile.example

- [x] rebase against #1735


src/picongpu/CMakeLists.txt
```diff
@@ -200,7 +200,7 @@ add_definitions(-DPIC_VERBOSE_LVL=${PIC_VERBOSE})
 option(PIC_ENABLE_INSITU_VOLVIS "Enable In Situ Volume Visualization" OFF)
 if(PIC_ENABLE_INSITU_VOLVIS)
     add_definitions(-DENABLE_INSITU_VOLVIS=1)
-    set(CUDA_ARCH sm_35 CACHE STRING "set GPU architecture")
+    set(CUDA_ARCH 35 CACHE STRING "Set GPU architecture (semicolon or comma separated list, e.g. '-DCUDA_ARCH=20,35,60')")
```
psychocoderHPC: Note to @ax3l: this line should not part of PIConGPU or should trigger an error.
We should we remove  `PIC_ENABLE_INSITU_VOLVIS` because the old insitu code base is not longer compatible with the dev branch of PIConGPU.

ax3l:
interesting and useful.

does this break *forward compatibility* of generated binaries? e.g., will a `sm_35` project still run on K80 (`sm_37`)?

psychocoderHPC:
Yes it will be forward compatible. The ptx code is embedded and compiled at runtime if the device SAS code is not included. Than the behavior is like the old implementation.

src/libPMacc/PMaccConfig.cmake
```diff
@@ -66,11 +66,19 @@ if(NOT "${CMAKE_CXX_FLAGS}" MATCHES "-std=c\\+\\+11")
...
 
-set(CUDA_ARCH sm_20 CACHE STRING "Set GPU architecture")
-string(COMPARE EQUAL ${CUDA_ARCH} "sm_10" IS_CUDA_ARCH_UNSUPPORTED)
-string(COMPARE EQUAL ${CUDA_ARCH} "sm_11" IS_CUDA_ARCH_UNSUPPORTED)
-string(COMPARE EQUAL ${CUDA_ARCH} "sm_12" IS_CUDA_ARCH_UNSUPPORTED)
-string(COMPARE EQUAL ${CUDA_ARCH} "sm_13" IS_CUDA_ARCH_UNSUPPORTED)
+set(CUDA_ARCH 20 CACHE STRING "Set GPU architecture (semicolon or comma separated list, e.g. '-DCUDA_ARCH=20,35,60')")
+
+# convert all commas to semicolons to support the following foreach call
+string(REPLACE "," ";" PMACC_CUDA_ARCH ${CUDA_ARCH})
```
ax3l: I would make it a semicolon separeted list by requirement and spare all the inconveniences and corner cases that come in when allowing both. There are many scripts setting and using this and allowing two syntax forms just causes trouble because all of them have to care.

src/libPMacc/PMaccConfig.cmake
```diff
@@ -66,11 +66,19 @@ if(NOT "${CMAKE_CXX_FLAGS}" MATCHES "-std=c\\+\\+11")
...
+foreach(PMACC_CUDA_ARCH_ELEM ${PMACC_CUDA_ARCH})
+    # set flags to create device code for the given architecture
+    set(CUDA_NVCC_FLAGS ${CUDA_NVCC_FLAGS}
+        "--generate-code arch=compute_${PMACC_CUDA_ARCH_ELEM},code=sm_${PMACC_CUDA_ARCH_ELEM}")
+    string(COMPARE EQUAL ${PMACC_CUDA_ARCH_ELEM} "10" IS_CUDA_ARCH_UNSUPPORTED)
```
ax3l: make two checks:
- is the element an integer? no-> unsupported (without output of element string)
- is the integer >20? no-> unsupported (without output of element string)

this will help users to do the transition

src/picongpu/CMakeLists.txt
ax3l: yes, @theZiz was actually eager to provide a PR removing the old insituvolvis

psychocoderHPC:
updated

psychocoderHPC:
rebased

src/libPMacc/PMaccConfig.cmake
```diff
@@ -66,16 +66,24 @@ if(NOT "${CMAKE_CXX_FLAGS}" MATCHES "-std=c\\+\\+11")
...
+    string(REGEX MATCH "^[0-9]+$" PMACC_IS_NUMBER ${CUDA_ARCH})
+    if(NOT PMACC_IS_NUMBER)
+        message(FATAL_ERROR "Defined compute architecture '${PMACC_CUDA_ARCH_ELEM}' in "
+                            "'${CUDA_ARCH}' is not an integral number, use e.g. '20' (for SM 2.0).")
+    endif()
+    unset(PMACC_IS_NUMBER)
+
+    # set flags to create device code for the given architecture
+    set(CUDA_NVCC_FLAGS ${CUDA_NVCC_FLAGS}
+        "--generate-code arch=compute_${PMACC_CUDA_ARCH_ELEM},code=sm_${PMACC_CUDA_ARCH_ELEM}")
```
ax3l: ok, I understood the final changes here now via http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#application-compatibility and http://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-steering-gpu-code-generation.

```
The nvcc user manual lists various shorthand for the -arch, -code, and -gencode compiler options. For example, -arch=sm_35 is a shorthand for -arch=compute_35 -code=compute_35,sm_35 (which is the same as -gencodearch=compute_35,code=\'compute_35,sm_35\'). 
```

Nevertheless, I think you got a bug in here. If you check the [*virtual* vs. *real* compute architectures](http://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#virtual-architecture-feature-list), there is not *every* version that we know from SM's also in the *virtual* compute capabilities that you put in `arch=compute_XY`. (*The above table lists the currently defined virtual architectures.*)

ax3l: hm, maybe this list is just not complete...

ax3l: actually I am not sure if you are still JIT forward compatible. By skipping the `compute_XY` part in `code` you do not embed a *PTX* binary load image and only a *assembled SASS*  binary load image for the specific SM (via `sm_XY`), which means there is no PTX code to generate missing SASS for other (newer) SMs.
```

This is because your `arch=compute_XY` does *not* trigger PTX generation!
```
Normally, this option alone does not trigger assembly of the generated PTX for a real architecture (that is the role of nvcc option --gpu-code, see below); rather, its purpose is to control preprocessing and compilation of the input to PTX. 
```
 	

Specify the name of the NVIDIA GPU to assemble and optimize PTX for.

nvcc embeds a compiled code image in the resulting executable for each specified code architecture, which is a true binary load image for each real architecture (such as sm_50), and PTX code for the virtual architecture (such as compute_50).

During runtime, such embedded PTX code is dynamically compiled by the CUDA runtime system if no binary load image is found for the current GPU. 
```
http://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-steering-gpu-code-generation

ax3l: hm, I am probably wrong about that...
... true binary load image for each real architecture (such as sm_50), **and** PTX code for the virtual architecture (such as compute_50).

ax3l:
Can you elaborate a bit more **why**
> This reduce[s] the start time of the binary if isaac live visualization is enabled.

Didn't we create the correct SASS code because we forgot to set `--gpu-code`, too and so we always did depend on the JIT?

psychocoderHPC:
Let use check this tomorrow, I am sure that ptx code is generated because if you enable show registers the option only creates an output if ptx code is generated. But maybe I missed `compute_XY` in the second option.

ax3l:
Yes, I think I got it finally but we should really cross-check what a virtual architecture (`compute_`) does on the `-code` part and why it is added there by default usually.

ax3l:
Interesting are also the pre-defined short-hands.
```
 For example, nvcc --gpu-architecture=sm_50 is equivalent to nvcc --gpu-architecture=compute_50 --gpu-code=sm_50,compute_50. 
```

So what you change right now is: you remove the `compute_XY` from `-code`. And even *before* this PR, `-code` was already set by default...

psychocoderHPC:
I tested all again and found that the ptx code was not added to the binary (it was not possible to execute a binary for `sm_35` on `sm_60`.
I added a second option to include the ptx code.

ax3l:
thx for the update and fix! :)

psychocoderHPC:
CMake command summery:

generate forward compatible ptx code and embed it in the binary
```
--generate-code arch=compute_XX,code=compute_XX
```

generate SAS code and embed it in the binary
```
--generate-code arch=compute_XX,code=sm_XX
```

ax3l:
@psychocoderHPC for some reason, this PR breaks separable compilation.

in `src/picongpu/CMakeLists.txt`:
```CMake
set(CUDA_SEPARABLE_COMPILATION ON)
```

-> `nvcc fatal   : Unknown option '-generate-code arch'`

Issue #120, opened 2013-12-10
Rerunable / Restartable


ax3l:
Many batch systems support a flag if a job can be aborted and/or restarted. We may want to use that sometimes.
### List of flags and their meaning

qsub/torque: `Rerunable`, `fault_tolerant`
...

BrianMarre:
@psychocoderHPC is this issue still relevant?

PrometheusPi:
Good point.
With automatic checkpoint via [signals](https://picongpu.readthedocs.io/en/latest/expert/signals.html) enabled and [`--checkpoint.tryRestart`](https://picongpu.readthedocs.io/en/latest/usage/plugins/checkpoint.html#cfg-file) as default, this would already be possible. But I think no (default) `*.cfg` uses this currently 
Should we set this as default? Or should we close this issue since users can do it on their own?

psychocoderHPC:
I will close the issue, we have everything to support rerun and basic fault tolerance.
For crusher I also shipped a not easy to understand example of how you could allocate more nodes as you need for your job to handle broken nodes: https://github.com/ComputationalRadiationPhysics/picongpu/blob/1e50e0c69e1d371763ebcf56903496d8942ebee1/etc/picongpu/crusher-ornl/batch.tpl#L120-L155

> Should we set this as default?

I do not think that this is a good default because it is only working with your configured checkpointing and it could lead to unexpected large IO dumps and could fill up the disk above the disk quota.

PR #1731, opened 2017-01-02
[WIP] allow full compile with clang

psychocoderHPC:
This is only a preview how many changes are needed to support clang for host and device side.

- use clang for host and device compile
- fix missing typedefs
- handle blockDim and `threadDim` for clang
- remove `BOOST_AUTO` with `PMACC_AUTO`
- use native c++11 `auto` for `PMACC_AUTO`
- CMake: add path to compile with clang
- disable static assertion for clang compile (error on compile)
- cuSTL: add missing prefix `HDINLINE`

Changes in the third party library `mallocMC` need to be ported to mallocMC.

- [ ] split into independent parts 

@ax3l this could be interesting for you

### DIFF
--- a/src/picongpu/include/plugins/PhaseSpace/PhaseSpaceFunctors.hpp
+++ b/src/picongpu/include/plugins/PhaseSpace/PhaseSpaceFunctors.hpp
@@ -104,8 +104,8 @@ namespace picongpu
             int p_bin = int( rel_bin * float_X(num_pbins) );
 
             /* out-of-range bins back to min/max */
-            p_bin >= 0 ? /* do not change p_bin */ : p_bin=0;
-            p_bin < num_pbins ? /* do not change p_bin */ : p_bin=num_pbins-1;
+            p_bin >= 0 ? p_bin /* do not change p_bin */ : p_bin=0;
+            p_bin < num_pbins ? p_bin/* do not change p_bin */ : p_bin=num_pbins-1;
 
             /** \todo take particle shape into account */
             atomicAddWrapper( &(*curDBufferOriginInBlock( p_bin, r_bin )),
--- a/src/libPMacc/PMaccConfig.cmake
+++ b/src/libPMacc/PMaccConfig.cmake
@@ -54,51 +54,61 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_CXX_STANDARD 11)
 
+option(CLANG_DEVICE_COMPILE "enable clang compile for device code" OFF)
 
 ################################################################################
 # Find CUDA
 ################################################################################
 find_package(CUDA 7.5 REQUIRED)
 
-# work-around since the above flag does not necessarily put -std=c++11 in
-# the CMAKE_CXX_FLAGS, which is unfurtunately hiding the switch from FindCUDA
-if(NOT "${CMAKE_CXX_FLAGS}" MATCHES "-std=c\\+\\+11")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+set(PMacc_LIBRARIES ${PMacc_LIBRARIES} ${CUDA_LIBRARIES})
+
+if(CLANG_DEVICE_COMPILE)
+    #set(LIBS ${LIBS} cudart_static)
+    set(CLANG_BUILD_FLAGS "-ftemplate-backtrace-limit=0 -O3 -x cuda --cuda-path=${CUDA_TOOLKIT_ROOT_DIR}")
+    set(CLANG_BUILD_FLAGS "${CLANG_BUILD_FLAGS} --cuda-gpu-arch=${CUDA_ARCH}")
+    add_definitions(-DBOOST_NO_CXX11_SMART_PTR)
+    add_definitions(-DCLANG_DEVICE_COMPILE=1)
+else()
+    # work-around since the above flag does not necessarily put -std=c++11 in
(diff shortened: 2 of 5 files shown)


src/libPMacc/include/cuSTL/container/CartBuffer.tpp
```diff
@@ -298,22 +298,22 @@ CartBuffer<Type, T_dim, Allocator, Copier, Assigner>::zone() const
...
-bool
+HDINLINE bool
 CartBuffer<Type, T_dim, Allocator, Copier, Assigner>::isContigousMemory() const
 {
     return this->pitch == detail::PitchHelper<dim>()(this->_size);
 }
 
 template<typename Type, typename Allocator, typename Copier, typename Assigner>
-std::ostream& operator<<(std::ostream& s, const CartBuffer<Type, 1, Allocator, Copier, Assigner>& con)
+HDINLINE std::ostream& operator<<(std::ostream& s, const CartBuffer<Type, 1, Allocator, Copier, Assigner>& con)
```
psychocoderHPC: can be removed this is a host only function

psychocoderHPC:
Results after a short test:
  - IMAGES: looks good and correct (no HDF5 tests done)
  - speed test KHI 2000 steps electrons/ions: clang is a factor of `1.12` slower :-(   
    - clang: 385 sec
    - nvcc: 344 sec

psychocoderHPC:
Maybe clang is slower because I didn't set fast math for clang: http://llvm.org/docs/CompileCudaWithLLVM.html#id5

ax3l:
omg, how awesome is that! happy new year! :sparkles:

ax3l:
@slizzered mallocMC patch: related to #120

src/picongpu/include/plugins/PhaseSpace/PhaseSpaceFunctors.hpp
```diff
@@ -104,8 +104,8 @@ namespace picongpu
             int p_bin = int( rel_bin * float_X(num_pbins) );
 
             /* out-of-range bins back to min/max */
-            p_bin >= 0 ? /* do not change p_bin */ : p_bin=0;
-            p_bin < num_pbins ? /* do not change p_bin */ : p_bin=num_pbins-1;
+            p_bin >= 0 ? p_bin /* do not change p_bin */ : p_bin=0;
+            p_bin < num_pbins ? p_bin/* do not change p_bin */ : p_bin=num_pbins-1;
```
theZiz: Is there are good reason, why you use this method instead of an classicl "if"? Furthermore if clang doesn't like emptyness between `?` and `:`, `p_bin >= 0 ? true /* do not change p_bin */ : p_bin=0;` should also work? This way it looks like, something happends to p_bin, but that is not the case....

psychocoderHPC: This version can optimized by the compiler with special asm command without branching. Maybe the compile can also optimize simple if conditions, we should check it.

theZiz: Okay, fair enough. But with this assumption every if-statement should be avoided. :sweat_smile:

ax3l: > Maybe the compile can also optimize simple if conditions, we should check it.

We checked this in the past (~CUDA 6) and nvcc is indeed creating way more efficient SIMD ptx for those kinds of short hand assign-ifs.

Generally speaking, `if`s are way more general and harder to optimize then short-hand assigns  (*pattern matching* is *never* your friend. *Looking at you, #pragma based acceleration language extensions!*). But maybe the compiler became more "clever" (has better pattern matching) since then.

ax3l: fun fact: [gcc](http://stackoverflow.com/questions/12260777/ternary-operator-without-else-in-c/22394622#22394622) doesn't allow to skip the 2nd element of the tenary operator, and clang does not allow to skip the 1st one.

psychocoderHPC: I check this explicit code an changed it to 
```C++
 if(p_bin < 0)
    p_bin=0;
 if(p_bin >= num_pbins) p_bin=num_pbins-1;
```

For both cases the PTX code is equal , this means we can change the tenary operator.

theZiz: With Cuda 8? Anyway: Awesome!

ax3l: Wuhu, same ptx with
```C++
#include <cuda.h>
#include <cstdio>
#include <cstdlib>

__global__ void a( int p_bin )
{
    const int num_pbins = 1024;

//    p_bin >= 0 ? p_bin /* do not change p_bin */ : p_bin=0;
//    p_bin < num_pbins ? p_bin/* do not change p_bin */ : p_bin=num_pbins-1;

    if( p_bin < 0 )
      p_bin = 0;
    if( p_bin >= num_pbins )
      p_bin = num_pbins - 1;

  printf("%i\n", p_bin);
}

int main( int argc, const char* argv[] )
{
    unsigned int p_bin = std::atoi(argv[1]);

    a<<<1,1>>>( p_bin );

    cudaDeviceSynchronize();
}
```

ax3l: busted

psychocoderHPC:
final pull equest for this feature is #1933

PR #1733, opened 2017-01-03
License Header: Update 2017

ax3l:
Updating years in header files is cumbersome, so we update all files to include 2017 at once.

Updated via

``` bash
grep -iR "Copyright 20" . | awk -F: '{print $1}' | \
  xargs -n1 -P1 -I{} sed -i 's/\(Copyright 20..\) /\1-2017 /g' {}
grep -iR "Copyright 20" . | awk -F: '{print $1}' | \
  xargs -n1 -P1 -I{} sed -i 's/\(Copyright 20..-201\). /\17 /g' {}
```

and commited via

``` bash
git checkout thirdParty/

  git commit -a -S
```

Happy new year! :) :star2:

### DIFF
--- a/src/libPMacc/include/cudaSpecs.hpp
+++ b/src/libPMacc/include/cudaSpecs.hpp
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015 Heiko Burau
+ * Copyright 2015-2017 Heiko Burau
  *
  * This file is part of libPMacc.
  *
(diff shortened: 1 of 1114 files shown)


src/libPMacc/include/cudaSpecs.hpp
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015 Heiko Burau
+ * Copyright 2015-2017 Heiko Burau
```
PrometheusPi: good catch (by your code) - why did we miss this last year?

ax3l: ;) probably an open PR during the update or something pasted in later from an outdated local branch of him.

PrometheusPi: you are right - pr from Dec 2015 which was merged Jan 2016

PR #1750, opened 2017-01-05
Fix energy density

Heikman:
Fix energy density calculation. Prior to this the weighting factor was wrongly included in the energy density expression resulting in a quadratic dependency.

**Work-Around:** Use `CreateEnergyOperation` instead and divide by cell volume.


ax3l:
@psychocoderHPC that one is for you to finish.

PR #1751, opened 2017-01-09
openPMD: iterationFormat only Basename

ax3l:
The `iterationFormat` in `openPMD`'s base attributes in `/` should only contain the *basename*, not the path to the file in "fileBased" `iterationEncoding`.

This is not a minor bug since most post-processing tools I am aware of are currently not reading that attribute, but they might. Nevertheless, it violates the standard (I did not check it carefully when moving `.h5` and `.bp` files to their own dir in `simOutput`).

### DIFF
--- a/src/picongpu/include/plugins/adios/WriteMeta.hpp
+++ b/src/picongpu/include/plugins/adios/WriteMeta.hpp
@@ -144,7 +144,10 @@ namespace writeMeta
             ADIOS_CMD(adios_define_attribute_byvalue(threadParams->adiosGroupHandle,
                 "iterationEncoding", "/", adios_string, 1, (void*)iterationEncoding.c_str()));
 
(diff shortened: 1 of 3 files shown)


src/picongpu/include/plugins/adios/WriteMeta.hpp
```diff
@@ -144,7 +144,10 @@ namespace writeMeta
             ADIOS_CMD(adios_define_attribute_byvalue(threadParams->adiosGroupHandle,
                 "iterationEncoding", "/", adios_string, 1, (void*)iterationEncoding.c_str()));
 
-            const std::string iterationFormat( threadParams->adiosFilename + std::string("_%T.h5") );
```
ax3l: ha, an other great copy&paste error!

psychocoderHPC: solved by @ax3l

PR #1754, opened 2017-01-09
Thomas-Fermi impact ionization model

n01r:
#### This pull request adds the Thomas-Fermi impact/collisional ionization model to PIConGPU.

The model is implemented using the field ionization framework introducing grid-mapped and particle-position-interpolated *density* and *"temperature"* fields as input parameters. For more information see this [issue](https://github.com/n01r/picongpu/issues/12).

The implementation also makes use of multiple `FieldTmp` slots (#1703).   

Preliminary compile- and runtime tests have been successful and show good agreement of HDF5 output and analytical expectation of charge states.   

I propose to do a follow-up PR with a new `ThinFoil` example to free up the `LaserWakefield` example.

##### ToDo

- [x] Clean up the commit history and regroup into meaningful units

-----------------------------------------

~~Please refrain from commenting on small inconsistencies until the above box has been checked.~~

- [x] Replace `round2Nearest(avgChargeState)` by RNG generated distribution between `floor` and `ceil` (Update 2017-01-10)


n01r:
@ax3l Does it make sense to add this PR to the *Collisional Ionization Methods* project?

ax3l:
yes sure, done!

> I propose to do a follow-up PR with a new ThinFoil example to free up the LaserWakefield example.

yes, sounds good!

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,150 @@
...
+            /* ionization condition */
+            if (chargeState < protonNumber)
+            {
+                /* atomic mass number (usually A) A = N + Z */
+                float_64 massNumber = neutronNumber + protonNumber;
+
+                /** @TODO replace the static_cast<float_64> by casts to float_X
+                 * or leave out entirely and compute everything in PIConGPU scaled units
+                 */
+                float_64 const densityUnit = static_cast<float_64>(particleToGrid::derivedAttributes::Density().getUnit()[0]);
```
n01r: Do calculations in numerically stable scaled units if possible.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,150 @@
...
+                float_64   R = massDensity/(protonNumber * massNumber);
+
+                float_64 Q_1 = A * math::pow(R,B);
+
+                float_64   Q = math::pow(math::pow(R,C) + math::pow(Q_1,C), float_64(1.)/C);
+
+                float_64  x = TFAlpha * math::pow(Q,TFBeta_temp);
+
+                /* Thomas-Fermi average ionization state */
+                float_X ZStar = static_cast<float_X>(protonNumber * x / (float_64(1.) + x + math::sqrt(float_64(1.) + float_64(2.)*x)));
```
n01r: thomas-fermi_to_not_infringe_copyright

Let's see if we can go from `float_64` to `float_X`.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+                /** determine the new number of bound electrons from the TF ionization state by rounding to nearest charge state
+                 * @TODO introduce partial macroparticle ionization / ionization distribution at some point
+                 */
+                float_X newBoundElectrons = protonNumber - float_X(math::float2int_rn(ZStar));
```
n01r: Follow this up with a commit that calculates
```c++
/* works only for positive charge states */
float_X chargeState  = math::floor(ZStar) + static_cast<float_X>(randNr < math::modf(ZStar))
float_X newBoundElectrons = protonNumber - chargeState;
```
where `randNr` is a random number.   
If for example `ZStar = 3.6` this would yield a charge state of 4+ in 60% and a charge state of 3+ in 40% of the cases.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,150 @@
...
+                 * or leave out entirely and compute everything in PIConGPU scaled units
+                 */
+                float_64 const densityUnit = static_cast<float_64>(particleToGrid::derivedAttributes::Density().getUnit()[0]);
+                float_64 const kinEnergyDensityUnit = static_cast<float_64>(particleToGrid::derivedAttributes::EnergyDensity().getUnit()[0]);
+                /* convert from kinetic energy density to average kinetic energy per particle */
+                float_64 kinEnergy = (kinEnergyDensity / density) * (kinEnergyDensityUnit / densityUnit);
+                /** convert kinetic energy in J to "temperature" in eV by assuming an ideal electron gas
+                 * E_kin = 3/2 k*T
+                 */
+                float_64 temperature = kinEnergy * UNITCONV_Joule_to_keV * float_64(1.e3) * float_64(2./3.);
```
n01r: What do we actually want to call an *"electron temperature"*?

@ax3l , @HighIander : ideas?

**ToDo:** Implement cut-off at 50 keV (?)  / Make the cut-off a user-definable parameter.

HighIander: Well, the cut-off is an important problem and there is a very recent scientific discussion about exactly this topic here: Phys. Rev. Lett. 113, 255001 (2014) and in the corresponding comment and reply. 
Summarizing, I think a 30 keV cutoff as default is OK for now, but one should allow for a user-defined value and (later) even better would allow for a real fit with a Maxwell to the low-energy part or getting the thermal velocity by any other means "on-the-fly".

ax3l: Yes, we should keep the assumption that one can derive a "temperature" in the simple model from the mean with user-defined cut-off and should then rather check against one of the many extensions of TF, if one is better suited non-thermal distributions for our use cases (see the references from last week in *David Salzmann, 1998, Ch. 2.4*).

ax3l: update: ah no, of course the width of the MW would be suitable to get a better T_e even in the default TF.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
n01r: @ax3l could you suggest a way to calculate this in more numerically stable units?
I don't really see it right now. Normally this is fine because you need to arrive at the correct units in the end and can just multiply after all calculations back into SI. But since this is a fitted formula what do I do here?

E.g. with a terms like `math::pow(R,B)` and `math::pow(R,C)`

n01r:
Go ahead @ax3l , @psychocoderHPC  you can review from here. :+1: .   
No need to introduce new feature parts all in this PR.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
ax3l: yep, it's nasty but possible.
first start by checking  the dimensions of the constants/fit parameters (and add them). We can then normalize those quantities in formula accordingly. If the formula is SI already, then it's just simple scaling of the constants before doing the calc.

n01r: So we do agree on keeping the calculation in double precision for now since there is no easy way of retracing how and with which units this was fitted?

ax3l:
@n01r compile error, e.g.
```
src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp(136):
  error: no instance of function template "PMacc::algorithms::math::fmod" matches the argument list
    argument types are: (picongpu::float_X)
```

psychocoderHPC:
`fmod` takes two arguments.

Am 27. Januar 2017 18:57:31 MEZ, schrieb Axel Huebl <notifications@github.com>:
>@n01r compile error, e.g.
>```
>src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp(136):
>error: no instance of function template "PMacc::algorithms::math::fmod"
>matches the argument list
>    argument types are: (picongpu::float_X)
>```
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/ComputationalRadiationPhysics/picongpu/pull/1754#issuecomment-275730057

-- 

n01r:
Yeah, corrected that offline already and had it in a `git stash` which I then forgot :man_facepalming:

n01r:
@ax3l checks are passing, could you review this week, please?

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,158 @@
...
+#include "simulation_defines.hpp"
+#include "particles/traits/GetAtomicNumbers.hpp"
+#include "traits/attribute/GetChargeState.hpp"
+#include "algorithms/math/floatMath/floatingPoint.tpp"
+#include "particles/ionization/byCollision/ThomasFermi/TFFittingParameters.def"
+
+/** \file AlgorithmThomasFermi.hpp
+ *
+ * IONIZATION ALGORITHM for the Thomas-Fermi model
+ *
```
ax3l: can you add the original paper and maybe other materials you found relevant for your implementation at this point?

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,158 @@
...
+{
+namespace ionization
+{
+
+    /** \struct AlgorithmThomasFermi
+     *
+     * \brief calculation for the Thomas-Fermi pressure ionization model
+     *
+     * This model uses local density and "temperature" values as input
+     * parameters. To be able to speak of a "temperature" an equilibrium state
```
ax3l: Phrasing suggestion: ~~To be able to speak of~~ -> A physical temperature requires a defined equilibrium state ...

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,158 @@
...
+     * would be required. Typical high power laser-plasma interaction is highly
+     * non-equilibrated, though. The name "temperature" is kept to illustrate
+     * the origination from the Thomas-Fermi model. It is nevertheless
+     * more accurate to think of it as an averaged kinetic energy.
```
ax3l: ... to think of it as an averaged kinetic energy, *which is not backed by the Model and should therefore be only used with a certain suspicion in such Non-LTE scenarios.*

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,158 @@
...
+         * \param parentIon particle instance to be ionized
+         * \param randNr random number
+         */
+        template<typename KinEnergyDensityType, typename DensityType, typename ParticleType >
+        HDINLINE void
+        operator()( const KinEnergyDensityType kinEnergyDensity, const DensityType density, ParticleType& parentIon, float_X randNr )
+        {
+
+            /* @TODO replace the float_64 with float_X and make sure the values are scaled to PIConGPU units */
+            const float_64 protonNumber = GetAtomicNumbers<ParticleType>::type::numberOfProtons;
```
ax3l: I think both lines can be `constexpr`

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,158 @@
...
+            const float_64 neutronNumber = GetAtomicNumbers<ParticleType>::type::numberOfNeutrons;
+            float_64 chargeState = attribute::getChargeState(parentIon);
+
+            /* ionization condition */
+            if (chargeState < protonNumber)
+            {
+                /* atomic mass number (usually A) A = N + Z */
+                float_64 massNumber = neutronNumber + protonNumber;
```
ax3l: should also be a `constexpr`

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+                /** @TODO replace the static_cast<float_64> by casts to float_X
+                 * or leave out entirely and compute everything in PIConGPU scaled units
+                 */
+                float_64 const densityUnit = static_cast<float_64>(particleToGrid::derivedAttributes::Density().getUnit()[0]);
+                float_64 const kinEnergyDensityUnit = static_cast<float_64>(particleToGrid::derivedAttributes::EnergyDensity().getUnit()[0]);
+                /* convert from kinetic energy density to average kinetic energy per particle */
+                float_64 kinEnergy = (kinEnergyDensity / density) * (kinEnergyDensityUnit / densityUnit);
```
ax3l: `const float_64 kinEnergy`

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,158 @@
...
+                /** convert kinetic energy in J to "temperature" in eV by assuming an ideal electron gas
+                 * E_kin = 3/2 k*T
+                 */
```
ax3l: line below: `const float_64 temperature`

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,158 @@
...
+                float_64 temperature = kinEnergy * UNITCONV_Joule_to_keV * float_64(1.e3) * float_64(2./3.);
+
+                float_64 T_0 = temperature/math::pow(protonNumber,float_64(4./3.));
```
ax3l: `const float_64 T_0`

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+                float_64 T_F = T_0 / (float_64(1.) + T_0);
```
ax3l: `const float_64 T_F`

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+                /* for all the fitting parameters @see TFFittingParameters.def */
+
+                /** this is weird - I have to define temporary variables because
+                 * otherwise the math::pow function won't recognize those at the
+                 * exponent position */
+                float_64 TFA2_temp = TFA2;
```
ax3l: all once written vars: `const float_64`

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,158 @@
...
+                float_64 TFBeta_temp = TFBeta;
+
+                float_64 A = TFA1 * math::pow(T_0,TFA2_temp) + TFA3 * math::pow(T_0,TFA4_temp);
+
+                float_64 B = -math::exp(TFB0 + TFB1*T_F + TFB2*math::pow(T_F,float_64(7.)));
+
+                float_64 C = TFC1 * T_F + TFC2;
+
+                /** requires mass density in g/cm^3
+                 * @TODO relocate constants to param file or leave out and calculate unitless
```
ax3l: please move those out: `physicalConstants.param` :)

it's ok to access them here, put them in `SI::`

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,158 @@
...
+                 */
+                float_64 const nAvogadro = 6.022e23;
+                float_64 const convM3ToCM3 = 1.e6;
+
+                float_64 massDensity = density * densityUnit * massNumber / nAvogadro / convM3ToCM3;
```
ax3l: besides `density * densityUnit` you can use an other `constexpr` here to reduce RT calculations

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,158 @@
...
+                float_64 R = massDensity/(protonNumber * massNumber);
```
ax3l: calculate inverse `protonNumber * massNumber` as `constexpr` and multiply with `massDensity` at RT

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
ax3l: -> done so in the `.def` file, very good!

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi.def
```diff
@@ -0,0 +1,78 @@
...
+     *         default is boost::mpl placeholder because specialization
+     *         cannot be known in list of particle species' flags
+     *         \see speciesDefinition.param
+     */
+    template<typename T_IonizationAlgorithm, typename T_DestSpecies, typename T_SrcSpecies = bmpl::_1>
+    struct ThomasFermi_Impl;
+
+    /** Thomas-Fermi impact ionization model
+     *
+     *
```
ax3l: only one newline between title & brief of a docstring

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi.def
```diff
@@ -0,0 +1,78 @@
...
+     * This ionization model describes the atom inside the Thomas-Fermi framework
+     * in a self-consistent way. There the electrons are modeled as a density
+     * with respect to the distance from the core while the atomic core is often
```
ax3l: *often*: this specific implementation of the model does one assumption about the core, not a set of varying ones

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
```diff
@@ -0,0 +1,274 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <boost/type_traits/integral_constant.hpp>
```
ax3l: please move this to the bottom, just before stdlib includes

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
```diff
@@ -0,0 +1,274 @@
...
+             * \param localIdx local (linear) index in super cell / frame
+             * \param newMacroElectrons reference to variable for each thread that stores the number
+             *        of macro electrons to be created during the current time step
+             */
+            DINLINE void operator()(FrameType& ionFrame, int localIdx, unsigned int& newMacroElectrons)
+            {
+                /* alias for the single macro-particle */
+                auto particle = ionFrame[localIdx];
+                /* particle position, used for field-to-particle interpolation */
+                floatD_X pos = particle[position_];
```
ax3l: `const`

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
```diff
@@ -0,0 +1,274 @@
...
+            /* define ionization ALGORITHM (calculation) for ionization MODEL */
+            typedef T_IonizationAlgorithm IonizationAlgorithm;
+
+            /* random number generator */
+            typedef PMacc::random::RNGProvider<simDim, PMacc::random::methods::XorMin> RNGFactory;
+            typedef PMacc::random::distributions::Uniform<float> Distribution;
+            typedef typename RNGFactory::GetRandomType<Distribution>::type RandomGen;
+            RandomGen randomGen;
+
+            typedef MappingDesc::SuperCellSize TVec;
```
ax3l: that is a very nasty renaming, because you use it below as actual mapping to get a 3D index inside the supercell.
not sure if the rename is needed at all, otherwise just make it shorter, e.g., `using SuperCellSize = MappingDesc::SuperCellSize;`

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
```diff
@@ -0,0 +1,274 @@
...
+     * \brief Thomas-Fermi pressure ionization - Implementation
+     *
+     * \tparam T_DestSpecies electron species to be created
+     * \tparam T_SrcSpecies particle species that is ionized
+     */
+    template<typename T_IonizationAlgorithm, typename T_DestSpecies, typename T_SrcSpecies>
+    struct ThomasFermi_Impl
+    {
+
+        typedef T_DestSpecies DestSpecies;
```
ax3l: if you don't mind, please use the C++11 `using NewType = OtherType;` since it's more readable for new code.

src/picongpu/include/particles/ionization/byCollision/collisionalIonizationCalc.def
```diff
@@ -0,0 +1,46 @@
...
+namespace picongpu
+{
+
+namespace particles
+{
+
+namespace ionization
+{
+
+    struct AlgorithmNone;
```
ax3l: do you really need that?

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
n01r: Do they have to be in SI? They are partly unitless and partly with weird units. I'd rather put them into sth like `ionizerConfig.param`. (That was actually what I intended the file for.)

src/picongpu/include/particles/ionization/byCollision/collisionalIonizationCalc.def
n01r: That is the fallback like in `fieldIonizationCalc.def`. There should be other ways to do a fallback, though.

ax3l: yep, both should be removed. if no algo is used, just don't add the flag to the `Particle` class in the species.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
n01r: Seems not really possible in general right now. Maybe at a later point.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi.def
n01r: I think the documentation here is sufficient. If we later go to more detailed versions of the Thomas-Fermi model we can then document each variant on its own. For now it's safe to say that most Thomas-Fermi models assume that the atomic core has no spatial dimension and cannot interact except via the potential well it creates.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
n01r: there are no stdlib includes here, right? ^^; so this will be placed at the last place in the include list

n01r:
Merge conflicts in `ionizerConfig.param` - also still recent fix to `ionizationMethods.hpp` in #1824.
Need to rebase this against `dev` before finally uploading some test results and the eventual merge.

n01r:
I force-pushed the rebased and amended commit history. The branch was compile- and runtested for a couple of simulation steps. Before I will test the more complicated setup of a spatial temperature gradient I created a little steady-state scenario.

It is in an unphysical regime but that doesn't matter for the model functionality. 

We have: 
- **neutral nitrogen gas** at a **homogenous density of 1e25 m^-3**
- **electrons** at the **same density** with an **initial temperature of 17eV** 
- **no pushers**
- **no current**
- **ThomasFermi** ionization model (of course)
- **3 time steps** (already 1 is enough)

We expect an average charge state of roughly 4.5.
nitrogen_z_depending_on_t_thomasfermi_theoretical

That leaves us with an expected average number of `boundElectrons` equal to 2.5.   
This is the statistics overview of the `boundElectrons` attribute of the 4,194,291 nitrogen ions after the first time step. :arrow_down: 
steady_state_nitrogen_ionization_boundelectrons

Looks successful to me.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+                float_64 const B = -math::exp(TFB0 + TFB1*T_F + TFB2*math::pow(T_F,float_64(7.)));
+
+                float_64 const C = TFC1 * T_F + TFC2;
+
+                /* requires mass density in g/cm^3 */
+                float_64 constexpr nAvogadro = SI::N_AVOGADRO;
+                float_64 constexpr convM3ToCM3 = 1.e6;
+
+                float_64 const convToMassDensity = densityUnit * massNumber / nAvogadro / convM3ToCM3;
```
ax3l: `massNumber / nAvogadro / convM3ToCM3` can also be expressed as `constexpr` :-)

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,161 @@
...
+                float_64 const massDensity = density * convToMassDensity;
+
+                float_64 constexpr atomicTimesMassNumber = protonNumber * massNumber;
```
ax3l: review comment not applied fully. Please rite
```C++
float_64 constexpr invAtomicTimesMassNumber = float_64(1.) / (protonNumber * massNumber);
float_64 const R = massDensity * invAtomicTimesMassNumber;
```
instead.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,161 @@
...
+                float_64 constexpr TFBeta_temp = TFBeta;
+
+                float_64 const A = TFA1 * math::pow(T_0,TFA2_temp) + TFA3 * math::pow(T_0,TFA4_temp);
+
+                float_64 const B = -math::exp(TFB0 + TFB1*T_F + TFB2*math::pow(T_F,float_64(7.)));
+
+                float_64 const C = TFC1 * T_F + TFC2;
+
+                /* requires mass density in g/cm^3 */
+                float_64 constexpr nAvogadro = SI::N_AVOGADRO;
```
ax3l: please keep `constexpr` as the *first* argument. `constexpr` is *not* a type specifier which can be read right to left but is a specifier for the expression you are writing currently. discussed again today with @psychocoderHPC 

example difference to const: not all arguments on the right hand side need to be const to define a const.
but: to define a constexpr variable, the expression itself needs to be a constant expression.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
ax3l:
```C++
constexpr float_64 cleverNameYouChoose = massNumber / nAvogadro / convM3ToCM3;
float_64 const convToMassDensity = densityUnit * cleverNameYouChoose;
```

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi.def
```diff
@@ -0,0 +1,77 @@
+/**
```
ax3l: `/**` -> `/*`

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -0,0 +1,161 @@
+/**
```
ax3l: `/**` -> `/*`

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
```diff
@@ -0,0 +1,278 @@
+/**
```
ax3l: `/**` -> `/*`

src/picongpu/include/particles/ionization/byCollision/collisionalIonizationCalc.def
```diff
@@ -0,0 +1,44 @@
+/**
```
ax3l: `/**` -> `/*`

src/picongpu/include/particles/ionization/byCollision/collisionalIonizationCalc.hpp
```diff
@@ -0,0 +1,23 @@
+/**
```
ax3l: `/**` -> `/*`

src/picongpu/include/particles/ionization/byCollision/ionizers.def
```diff
@@ -0,0 +1,28 @@
+/**
```
ax3l: `/**` -> `/*`

src/picongpu/include/particles/ionization/byCollision/ionizers.hpp
```diff
@@ -0,0 +1,30 @@
+/**
```
ax3l: `/**` -> `/*`

src/picongpu/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -138,6 +139,10 @@ struct Hydrogen
  *                -> circularly polarized lasers
  * - Keldysh : Keldysh ionization model
  *
+ * - ThomasFermi : statistical impact ionization based on Thomas-Fermi
```
ax3l: after rebase, please add this block in `ionizerConfig.param`

src/picongpu/include/particles/ionization/byCollision/ionizers.hpp
```diff
@@ -0,0 +1,30 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+/** \file ionizers.hpp
```
ax3l: above `#pragma once` pls

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
ax3l: please remove the empty line and keep one block. had problems with adding newlines in the `\file` description in the last days.

Write like this please:

```C++
/** \file AlgorithmThomasFermi.hpp
 *
 * IONIZATION ALGORITHM for the Thomas-Fermi model
 * - implements the calculation of average ionization degree and changes charge states
 *   by decreasing the number of bound electrons
 * - is called with the IONIZATION MODEL, specifically by setting the flag in @see speciesDefinition.param
 */
```

src/picongpu/include/particles/ionization/byCollision/ionizers.def
```diff
@@ -0,0 +1,28 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+/* \file ionizers.def
```
ax3l: please above `#pragma once`

src/picongpu/include/simulation_defines/param/ionizerConfig.param
```diff
@@ -164,4 +165,35 @@ namespace picongpu
         28.339,
         28.339
     );
-}
+
+namespace particles
+{
+namespace ionization
+{
+
```
ax3l: plase add a `thomasFermi` namespace

src/picongpu/include/simulation_defines/param/ionizerConfig.param
```diff
@@ -164,4 +165,35 @@ namespace picongpu
...
+    /* Fitting parameters to average ionization degree Z* = 4/3*pi*R_0^3 * n(R_0)
+     * as an extension towards arbitrary atoms and temperatures
+     *
+     * See table IV of
+     * \url http://www.sciencedirect.com/science/article/pii/S0065219908601451
+     * doi:10.1016/S0065-2199(08)60145-1
+     */
+    float_X constexpr TFAlpha = 14.3139;
```
ax3l: keep constexpr left pls (see above)

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
n01r: done

examples/LaserWakefield/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -121,6 +122,10 @@ struct Hydrogen
  *                -> circularly polarized lasers
  * - Keldysh : Keldysh ionization model
  *
+ * - ThomasFermi : statistical impact ionization based on Thomas-Fermi
+ *                 atomic model
```
n01r: done

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
n01r: done

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
n01r: Thanks for the clarification!
Done!

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
n01r: Sorry, you're right. Done.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi.def
n01r: done

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
n01r: done

src/picongpu/include/particles/ionization/byCollision/collisionalIonizationCalc.def
n01r: done

src/picongpu/include/particles/ionization/byCollision/collisionalIonizationCalc.hpp
n01r: done

src/picongpu/include/particles/ionization/byCollision/ionizers.def
n01r: done

src/picongpu/include/particles/ionization/byCollision/ionizers.def
n01r: done

src/picongpu/include/particles/ionization/byCollision/ionizers.hpp
n01r: done

src/picongpu/include/particles/ionization/byCollision/ionizers.hpp
n01r: done

src/picongpu/include/simulation_defines/param/speciesDefinition.param
n01r: done

src/picongpu/include/simulation_defines/param/ionizerConfig.param
n01r: I added a `particles` namespace as well - but just for the TF fitting parameters for now to not break the methods which use the atomic and effective atomic numbers.

I still had to add a lot of `thomasFermi::` though. (Unless every model gets its own namespace.)

n01r:
offline talk: refactor all `typedef`s to `using`

n01r:
offline talk: `dc.releaseData` in `ThomasFermi_Impl.hpp` still produces a seg fault after the simulation has finished but this will be changed later in `DataConnector.hpp`

Apart from that the runtime checks seem okay after rebase and changes.

n01r:
*Merrrrrr-ge me .... harrrrrr-der!*

PR #1755, opened 2017-01-09
Bunch Example: Add Single e- Setup

ax3l:
Follow-up to #1716: Adds a single e- setup with the same center as the bunch's Gaussian
cloud. Re-adds a single-particle radiation test which is less hard to maintain as the previous one.

Also adds a `particleId` to all particles in the `Bunch` example, since it is
a) useful in that test and 
b) `particleId`s have not yet been covered by the compile suite and/or documented in examples

### To Do

- [x] rebase after depending PRs are merged
- [x] RT test if it works as expected


examples/Bunch/include/simulation_defines/param/gasConfig.param
```diff
@@ -46,35 +52,84 @@ namespace SI
...
+                    isStartCell = false;
+
+            if( isStartCell )
+                return 1.0;
 
-    /** the distance from gasCenter_SI until the gas density decreases to its 1/e-th part
-      *  unit: meter */
-    (PMACC_C_VECTOR_DIM(float_64, simDim, sigma_SI, 6.0e-6, 6.0e-6, 6.0e-6))
-); /* struct GaussianCloudParam */
+            return 0.0;
```
ax3l: imho, that is a question of personal preference.
since they is no code obfuscating what is happening, I will keep it that way if we do not find something more to change along.

ax3l:
@PrometheusPi did it create the expected radiation signature for you in both cases (bunch vs. single e-)?

PrometheusPi:
@ax3l My run time test failed for both bunch and single particle when picongpu tried to write radiation data (hdf5). I will check - might be just my setup.

PrometheusPi:
@ax3l Please adjust the walltime in the `*.cfg` to 5h

PrometheusPi:
@ax3l Due to the `gas -> density` change, there are some conflicts with this pull request. Please resolve.

PrometheusPi:
Run time test works as expected.

PR #1756, opened 2017-01-09
Fix ADIOS Trait for `bool`

ax3l:
For default I/O and checkpointing via ADIOS, simulations using our in situ radiation plugin with selected/subsets of particle species (by stride or gamma) did not compile since ADIOS' C API does not define a `bool` type for `radiationFlag`.

This adds a trait assuming it can be casted to one byte. It's nasty but I have no idea what to use instead since ADIOS has nothing on `bool`s in its manual.

### To Do

- [x] RT test
- [x] we should add a radiation test where `RAD_MARK_PARTICLE` is *larger* 1 or `RAD_ACTIVATE_GAMMA_FILTER` is non-zero to cover that case in the compile suite


src/picongpu/include/traits/PICToAdios.tpp
```diff
@@ -33,6 +33,20 @@ namespace picongpu
 namespace traits
 {
     template<>
+    struct PICToAdios<bool>
+    {
+        ADIOS_DATATYPES type;
+
+        PICToAdios() :
+        type(adios_unsigned_byte) {}
```
PrometheusPi: I assume `ADIOS_DATATYPES` is  a type for data-types - is this correct?
If so could you not write something like 
```c++
constexpr ADIOS_DATATYPES type = adios_unsigned_byte;

PICToAdios() { }
```

ax3l: Thanks for the question. It's an enum, or in-code an int.
There is no need to modify that, since this is in a non-sensitive part of the code. The only sensitive part on the host side is data movement and data re-processing, besides that we needs such tunings on the device. For all other places, like here, usability and maintainability first since it does not matter for performance.

It would also unnecessarily refactor the already existing interface we have right now which is sufficient.

ax3l:
@PrometheusPi in case that is interesting for you, you can check with [HDFCompass](https://github.com/HDFGroup/hdf-compass), `pip install adios` or `bpls`/`bpdump` if the data of the flag makes some kind of sense to you?

PrometheusPi:
@ax3l What is the status of the ToDo list?

ax3l:
I did hope a little that we talk how to test it xD

ax3l:
Tested in `hypnos:/bigdata/hplsim/development/huebl/bunch-adios-radflag-002/simOutput/bp`

At least from
```bash
bpls simData_0.bp -d "/data/0/particles/e/radiationFlag" | less
bpls simData_2500.bp -d "/data/2500/particles/e/radiationFlag" | less
```

there seem to be zeroes (at t=0) and ones (for all later stes) in it for `radiationFlag` as expected.

Same picture with python
```python
import adios as ad

f = ad.File("simData_2500.bp")
print( f["/data/2500/particles/e/radiationFlag"][42] )
#1

f = ad.File("simData_0.bp")
print( f["/data/0/particles/e/radiationFlag"][42] )
#0
```

so imo we are good to go.

Issue #1757, opened 2017-01-11
Missing include in GetDensityRatio.hpp


PrometheusPi:
I am using [`GetDensityRatio.hpp`](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/particles/traits/GetDensityRatio.hpp) on a taurus simulation (bases on 0.2.2) and by including this file in `particleConfig.param` I get the following compile error:
```
.../libraries/boost/1.60.0-gnu5.3/include/boost/mpl/if.hpp(63): error: name followed by "::" must be a class or namespace name
          detected during:
            instantiation of class "boost::mpl::if_<T1, T2, T3> [with T1=<error-type>, T2=picongpu::placeholder_definition58::DensityRatioElectrons, T3=picongpu::traits::detail::placeholder_definition47::DefaultDensityRatio]" 
.../picongpu/src/picongpu/include/particles/traits/GetDensityRatio.hpp(61): here
```

How can I solve this issue and use `GetDensityRatio.hpp`?

EDIT: 
caused by missing include in GetDensityRatio.hpp. The above error was not the first error but just a followup error.

n01r:
That was not the full extent of the error message. The part we see here shows that `T1=<error-type>` meaning that it does not understand the first template argument. A few lines above there it says `HasFlag` *is not a template* and that is probably where the error lies.

PrometheusPi:
This was due to a missing include in `GetDensityRatio.hpp` - fixed by #1759.

PR #1018, opened 2015-08-06
Particle Manipulators: Init Multi Ion Species

ax3l:
This commit adds new manipulators that can be used during the species init-pipeline.

- `RatioWeightingImpl`: Useful to clone one ion species from an other ion species by still satisfying the `densityRatio` of the new created species.

- `ProtonTimesWeightingImpl`: Useful to clone electron species from an ion species for pre-ionized plasmas. The weighting of the electrons is simply multiplied by the proton number of the initial ion species.

### Example Usage

in `particleConfig.param`:
```c++
namespace picongpu
{
namespace particles
{
// [...]
namespace manipulators
{
    // [...]
    typedef ProtonTimesWeightingImpl ProtonTimesWeighting;
    typedef RatioWeightingImpl RatioWeighting;
} /* namespace manipulators */
// [...]
} /* namespace particles */
} /* namespace picongpu */
```

in `speciesInitialization.param`:
```c++
namespace picongpu
{
namespace particles
{
    // [...]
    // Init-Pipeline for PMMA with Carbon, Oxygen, Proton and a general Electron species
    // (4 species in total, need to be defined in `speciesDefinition.param`)
    typedef mpl::vector<
        /* at this point, the reference density * densityRatio is taken to
         * calculate weightings */
        CreateGas<gasProfiles::SphereFlanks, startPosition::Random, PIC_Proton>,
        /* clone and weight additional ions according to their `densityRatio` */
        ManipulateCloneSpecies<manipulators::RatioWeighting, PIC_Proton, PIC_Carbon>,
        ManipulateCloneSpecies<manipulators::RatioWeighting, PIC_Proton, PIC_Oxygen>,
        /* re-distribute new ion species in their cell */
        Manipulate<manipulators::RandomPosition, PIC_Carbon>,
        Manipulate<manipulators::RandomPosition, PIC_Oxygen>,
        /* create electrons with numberOfProtons-higher weighting to emulate pre-ionization,
         * additionally, the initial plasma distribution is now field-neutral */
        ManipulateCloneSpecies<manipulators::ProtonTimesWeighting, PIC_Carbon, PIC_Electrons>,
        ManipulateCloneSpecies<manipulators::ProtonTimesWeighting, PIC_Oxygen, PIC_Electrons>,
        /* cloning from Hydrogen (Protons) is always easy: numberOfProtons == 1 */
        CloneSpecies<PIC_Proton, PIC_Electrons>
    > InitPipeline;
} /* namespace particles */
} /* namespace picongpu */
[...]

PR #956, opened 2015-07-24
multiply `GAS_DENSITY` by `densityRatio`

psychocoderHPC:
- add type of the species to template signature of
  - the kernel `kernelFillGridWithParticles`
  - the function `calcRealDensity `
- multiply `GAS_DENSITY` by `densityRatio` of the species

~~Do not merge before #955~~

PR #1759, opened 2017-01-11
Fix RatioWeighting (DensityWeighting) manipulator

### LINKED ISSUE #1758: RatioWeightingImpl.hpp deprecated
`picongpu/src/picongpu/include/particles/manipulators/RatioWeightingImpl.hpp` seems to be deprecated.

```C++
const float_X densityRatioDes =
    traits::GetDensityRatio<T_DesParticle>::type::getDefaultValue();
const float_X densityRatioSrc =
    traits::GetDensityRatio<T_SrcParticle>::type::getDefaultValue();
```

A scan of all the code returned only those two results for the method `getDefaultValue()` which is also not defined anywhere anymore.

PrometheusPi:
This pull request fixes compile errors when using `RatioWeightingImpl` during species initialization as manipulator. 

- [x] It is **still required to add** `#include "particles/manipulators/RatioWeightingImpl.hpp"` in `speciesInitialization.param`. @ax3l and @psychocoderHPC is this intended? -> fixed

Thanks @n01r for finding this bug and helping to fix it. 

This closes #1758 and also (the wrongly defined) #1757. 

***

@ax3l's notes:
- only the first two commits are relevant for a backport

### DIFF
--- a/src/picongpu/include/particles/manipulators/RatioWeightingImpl.def
+++ b/src/picongpu/include/particles/manipulators/RatioWeightingImpl.def
@@ -40,7 +40,7 @@ namespace manipulators
  * note: needs the densityRatio flag on both species,
  *       used by the GetDensityRatio trait.
  */
-struct RatioWeighting;
+struct RatioWeightingImpl;
 
 } //namespace manipulators
 } //namespace particles
--- a/src/picongpu/include/particles/manipulators/RatioWeightingImpl.hpp
+++ b/src/picongpu/include/particles/manipulators/RatioWeightingImpl.hpp
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2017 Axel Huebl
+ * Copyright 2015-2017 Axel Huebl, Richard Pausch
  *
  * This file is part of PIConGPU.
  *
@@ -73,9 +73,9 @@ struct RatioWeightingImpl
         if (isDesParticle && isSrcParticle)
         {
             const float_X densityRatioDes =
-                traits::GetDensityRatio<T_DesParticle>::type::getDefaultValue();
+                traits::GetDensityRatio<T_DesParticle>::type::getValue();
             const float_X densityRatioSrc =
-                traits::GetDensityRatio<T_SrcParticle>::type::getDefaultValue();
+                traits::GetDensityRatio<T_SrcParticle>::type::getValue();
 
             particleDes[weighting_] *= densityRatioDes / densityRatioSrc;
         }
--- a/src/picongpu/include/particles/traits/GetDensityRatio.hpp
+++ b/src/picongpu/include/particles/traits/GetDensityRatio.hpp
@@ -1,5 +1,5 @@
 /**
- * Copyright 2015-2017 Rene Widera
+ * Copyright 2015-2017 Rene Widera, Richard Pausch
  *
  * This file is part of PIConGPU.
  *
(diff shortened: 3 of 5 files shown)


ax3l:
Dang, we refactored the trait and the implementation at the same time when bringing that feature to mainline in #1018 (and should have added a test).
(Sorry, the example I gave you yesterday was in my local branch side-hacking that feature before mainline where it still worked ^^)

The actual bug then came in via #956 (ratio was always 1.0, #955) and survived/became active during #1018.

In #1018, after several refactorings, I think I just RT checked the proton-weighting manipulator.

src/picongpu/include/particles/manipulators/RatioWeightingImpl.hpp
```diff
@@ -73,9 +73,9 @@ struct RatioWeightingImpl
         if (isDesParticle && isSrcParticle)
         {
             const float_X densityRatioDes =
-                traits::GetDensityRatio<T_DesParticle>::type::getDefaultValue();
+	      traits::GetDensityRatio<T_DesParticle>::type::getValue();
             const float_X densityRatioSrc =
-                traits::GetDensityRatio<T_SrcParticle>::type::getDefaultValue();
+	      traits::GetDensityRatio<T_SrcParticle>::type::getValue();
```
ax3l: both lines: introduced in an untested combination of #956 and #1018 ...

ax3l:
> It is still required to add #include "particles/manipulators/RatioWeightingImpl.hpp" in speciesInitialization.param. [...] is this intended?

weird, `speciesInitialization` is included after `particleConfig` in `_defaultParam.loader` and there it is. No, you should not need to include the manipulator in `speciesInitialization`...

src/picongpu/include/particles/traits/GetDensityRatio.hpp
```diff
@@ -25,6 +25,8 @@
 #include "traits/Resolve.hpp"
 
 #include <boost/mpl/if.hpp>
+#include "particles/memory/frames/Frame.hpp"
+
 
 namespace picongpu
```
ax3l: @PrometheusPi can you please also correct the missed alignment in line 60?

src/picongpu/include/particles/manipulators/RatioWeightingImpl.hpp
```diff
@@ -73,9 +73,9 @@ struct RatioWeightingImpl
         if (isDesParticle && isSrcParticle)
         {
             const float_X densityRatioDes =
-                traits::GetDensityRatio<T_DesParticle>::type::getDefaultValue();
+                traits::GetDensityRatio<T_DesParticle>::type::getValue();
```
ax3l: ah no, your change is invalid. we use `value_identifier` in `speciesDefinition` which is a "default" value for an identifier that *could* be an l value, but is not used as such in the species.

ax3l:
I'll push the following changes:
- `RatioWeighting`: had inconsistent implementation in `.hpp` (with `Impl`) -> no parameters -> no `Impl` interface needed
- `ProtonTimesWeightingImpl` (works): rename without `Impl` (since no parameters, too)

ax3l:
@psychocoderHPC wuhu, approved. feel free to merge then :)

src/picongpu/include/particles/manipulators/RatioWeightingImpl.def
```diff
@@ -40,7 +40,7 @@ namespace manipulators
  * note: needs the densityRatio flag on both species,
  *       used by the GetDensityRatio trait.
  */
-struct RatioWeighting;
+struct RatioWeightingImpl;
```
PrometheusPi: @ax3l Was this the reason why the extra include was needed?

PR #1769, opened 2017-01-16
Add particle attribute `inCellOffset`

n01r:
Adds a particle attribute which holds the offset from one particle to the next
in the same cell in units of frame position. This is a meta attribute used for
future particle-particle methods like collision.

### DIFF
--- a/src/picongpu/include/simulation_defines/param/speciesAttributes.param
+++ b/src/picongpu/include/simulation_defines/param/speciesAttributes.param
@@ -63,6 +63,11 @@ value_identifier(float3_X,momentumPrev1,float3_X::create(0.));
 value_identifier(float_X, weighting, 0.0);
 /** use this particle for radiation diagnostics */
 value_identifier(bool, radiationFlag, false);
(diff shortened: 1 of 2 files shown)


src/picongpu/include/simulation_defines/param/speciesAttributes.param
```diff
@@ -63,6 +63,11 @@ value_identifier(float3_X,momentumPrev1,float3_X::create(0.));
 value_identifier(float_X, weighting, 0.0);
 /** use this particle for radiation diagnostics */
 value_identifier(bool, radiationFlag, false);
+/** offset from current particle to next particle in the same cell in units of
+ *  of frame position
+ *  - used for in-cell sorting and particle-particle (e.g. collisional) methods
+ */
+value_identifier(uint32_t, inCellOffset, 0);
```
n01r: Since this is not an attribute that should be messed with by the user the proper place for the definition is probably `picongpu/src/libPMacc/include/particles/Identifier.hpp`.

PR #1771, opened 2017-01-17
Add inCellOffset to list of default attributes

n01r:
Makes the `inCellOffset` a default particle attribute.


examples/LaserWakefield/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -42,7 +42,8 @@ namespace picongpu
 using DefaultParticleAttributes = MakeSeq_t<
     position<position_pic>,
     momentum,
-    weighting
+    weighting,
+    inCellOffset
```
theZiz: Too complicated!

PR #1777, opened 2017-01-20
Fix #1775 tbg config newline check

### LINKED ISSUE #1775: TBG: Windows (&Mac) EOL Rejected
I got a user bug report yesterday, that files edited under windows - which naturally get converted to `CR` `LF` line endings, are catched by our tbg tool via the ["contains spaces after line continuation"](https://github.com/ComputationalRadiationPhysics/picongpu/blob/0.2.2/src/tools/bin/tbg#L379) check.

We should evaluate if that is really critical, if we output a more sophisticated error message or if we do a conversion.

Also, we should write out `sed` commands as in our style checks that can be used by the user to copy-&-paste-fix the file in concern.

### Note on EOL
- Windows: `CR` `LF`
- Unix: `LF`
- Mac: `CR`

theZiz:
This PR fixes the check for end lines allowing windows new lines (Issue https://github.com/ComputationalRadiationPhysics/picongpu/issues/1775)
Furthermore as tbg does only support unix new lines, the config files are check for this trait.

### DIFF
--- a/src/tools/bin/tbg
+++ b/src/tools/bin/tbg
@@ -374,7 +374,7 @@ if [ ! -f "$cfg_file" ] ; then
 fi
 
 # cfg file sanity check - space after \ at EOL ?
-cfg_err=`egrep "\\\\\[[:space:]]+$" $cfg_file | wc -l`
+cfg_err=`egrep "\\\\\[[:blank:]]+$" $cfg_file | wc -l`
 if [ $cfg_err != 0 ] ; then
     echo "ERROR: file \"$cfg_file\" contains spaces after line continuation \\"
     echo "Check the following lines for end-of-line spaces:"
--- a/src/tools/bin/tbg
+++ b/src/tools/bin/tbg
@@ -379,7 +379,7 @@ if [ $cfg_err != 0 ] ; then
     echo "ERROR: file \"$cfg_file\" contains spaces after line continuation \\"
     echo "Check the following lines for end-of-line spaces:"
     echo ""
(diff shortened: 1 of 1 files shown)


src/tools/bin/tbg
```diff
@@ -374,12 +374,29 @@ if [ ! -f "$cfg_file" ] ; then
 fi
 
 # cfg file sanity check - space after \ at EOL ?
-cfg_err=`egrep "\\\\\[[:space:]]+$" $cfg_file | wc -l`
+cfg_err=`egrep "\\\\\[[:blank:]]+$" $cfg_file | wc -l`
 if [ $cfg_err != 0 ] ; then
     echo "ERROR: file \"$cfg_file\" contains spaces after line continuation \\"
     echo "Check the following lines for end-of-line spaces:"
     echo ""
     egrep -n "\\\[[:space:]]+$" $cfg_file
```
psychocoderHPC: `:space: ` should be `:blank:`

ax3l: is there a reason it is once 5x and an other time 3x `\`? Afaik, `-n` does not change the syntax of `egrep`, does it?

theZiz: Fixed it: https://github.com/ComputationalRadiationPhysics/picongpu/pull/1777/commits/4043dce648af730aa21e177c9ec0bb4dd51fdd80

theZiz: I have no idea, I am just the pattern matching boy. I see stuff and replace it with other stuff. I am not paid to think about the stuff! :stuck_out_tongue_winking_eye:

PR #1779, opened 2017-01-23
mallocMC: Organize Setting

ax3l:
Move the mallocMC settings to a clean, common header that can be used on top of `.cu` files.


src/picongpu/main.cu
```diff
@@ -28,52 +28,13 @@
...
-
-//use ScatterAllocator to replace malloc/free
-MALLOCMC_SET_ALLOCATOR_TYPE( ScatterAllocator );
-
+#include "mallocMCconfig.hpp"
 #include "ArgsParser.hpp"
 #include "communication/manager_common.h"
-#include "ArgsParser.hpp"
+
+#include "plugins/PluginController.hpp"
```
psychocoderHPC: Please remove all includes.

PR #1781, opened 2017-01-23
Source Files: Move to Headers

ax3l:
- not the cleanest way to keep them in `include/` too, but better then accumulating all in the root path
- CMakeLists: Compile host code separately without nvcc "code splitting" and link it statically

### DIFF
--- a/src/picongpu/CMakeLists.txt
+++ b/src/picongpu/CMakeLists.txt
@@ -324,15 +324,19 @@ include_directories(BEFORE ${PIC_EXTENSION_PATH}/include)
 # Compile & Link PIConGPU
 ################################################################################
 
-file(GLOB CUDASRCFILES "*.cu")
-file(GLOB SRCFILES "*.cpp")
+file(GLOB_RECURSE CUDASRCFILES "*.cu")
+file(GLOB_RECURSE SRCFILES "*.cpp")
 
-cuda_add_executable(picongpu
-    ${CUDASRCFILES}
+add_library(picongpu-hostonly
+    STATIC
     ${SRCFILES}
 )
 
-target_link_libraries(picongpu ${LIBS} m)
+cuda_add_executable(picongpu
+    ${CUDASRCFILES}
+)
+
+target_link_libraries(picongpu ${LIBS} picongpu-hostonly m)
 
 ################################################################################
 # Install PIConGPU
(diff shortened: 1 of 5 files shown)


src/picongpu/include/plugins/common/particlePatches.cpp
```diff
@@ -18,7 +18,7 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "include/plugins/common/particlePatches.hpp"
+#include "particlePatches.hpp"
```
psychocoderHPC: please use relative path from the root `include`. In this case `plugins/common/particlePatches.hpp`

src/picongpu/CMakeLists.txt
```diff
@@ -324,17 +324,19 @@ include_directories(BEFORE ${PIC_EXTENSION_PATH}/include)
...
+    ${SRCFILES}
+)
 
 cuda_add_executable(picongpu
     ${CUDASRCFILES}
-    ${SRCFILES}
 )
 
-target_link_libraries(picongpu ${LIBS} m)
+target_link_libraries(picongpu ${LIBS} picongpu-hostonly m)
```
psychocoderHPC: We should check if the static library is installed with `make install`

src/picongpu/include/plugins/common/stringHelpers.cpp
```diff
@@ -18,7 +18,7 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "include/plugins/common/stringHelpers.hpp"
+#include "stringHelpers.hpp"
```
psychocoderHPC: please use relative path from the root `inlcude`

src/picongpu/include/plugins/hdf5/openPMD/patchReader.cpp
```diff
@@ -20,7 +20,7 @@
 
 #if( ENABLE_HDF5 == 1 )
 
-#  include "include/plugins/hdf5/openPMD/patchReader.hpp"
+#  include "patchReader.hpp"
```
psychocoderHPC: please use relative path from the root `inlcude`

src/picongpu/CMakeLists.txt
ax3l: tested and it's not (since it is not part of our install targets)

Issue #1780, opened 2017-01-23
Moving windows does not work for slide_point set to zero


BeyondEspresso:
For use cases, such as loading external electron distributions into PIConGPU, it is desirable that the moving window starts directly with the first time step.

However, this is not supported, by the current implementation. Setting `slide_point = 0.0` leads to a non-moving window, because unsigned variables cannot become negative and thus `firstMoveStep` wraps around and becomes something very large instead. As a result the simulation does not get to the step where it should move.

``
            const uint32_t firstMoveStep = math::ceil(
                wayToFirstMove / deltaWayPerStep
            ) - 1;
``

A workaround is not trivial in the sense of just choosing a small non-zero number for `slide_point`.

``
            const uint32_t virtualParticleInitialStartCell = math::ceil(
                float_64(globalWindowSizeInMoveDirection) * (float_64(1.0) - slide_point)
            );
``

Due to the `ceil`-operation, the `slide_point` must be a small number corresponding to a distance slightly larger than one cell height. Then the `firstMoveStep` becomes `1`.

Hence, I suggest to provide a pull request, which makes `firstMoveStep` a signed variable, so it can assume negative values. What do you think?

psychocoderHPC:
solved with #1783

PR #1783, opened 2017-01-23
Bugfix in MovingWindow.hpp, slide_point now can be set to zero.

BeyondEspresso:
This pull request is based on #1780 and fixes the moving window for the edge case `slide_point = 0.0`. This makes it possible to activate the moving window directly at simulation initialization.

Happy merging :grinning:


src/picongpu/include/simulationControl/MovingWindow.hpp
```diff
@@ -105,12 +105,14 @@ class MovingWindow
                 cellSizeInMoveDirection;
             /* Is the time step when the virtual particle **passed** the moving window
              * in the current to the next step
+         * Signed type of firstMoveStep to allow for edge case slide_point = 0.0
```
psychocoderHPC: wrong indention

BeyondEspresso: Ok, the `sed`-command suggested by `continuous-integration/travis-ci/pr` did not work as expected. I fixed it manually.

PR #1788, opened 2017-01-24
fileOutput: MacroParticleCounter

ax3l:
Add a macro particle counter to the `FieldTmp` per cell particle to grid operations. Assumes all (macro) particles are point like in space and can be used to validate particle memory structures.

Removes Counter as a default used FieldTmp output since it is actually a debug quantity and wastes spaces.

### DIFF
--- a/src/picongpu/include/particles/particleToGrid/derivedAttributes/Counter.hpp
+++ b/src/picongpu/include/particles/particleToGrid/derivedAttributes/Counter.hpp
@@ -20,7 +20,7 @@
 
 #pragma once
 
-#include "particles/particleToGrid/derivedAttributes/ChargeDensity.def"
+#include "particles/particleToGrid/derivedAttributes/Counter.def"
 
 #include "simulation_defines.hpp"
 
--- a/src/picongpu/include/particles/particleToGrid/derivedAttributes/MacroCounter.def
+++ b/src/picongpu/include/particles/particleToGrid/derivedAttributes/MacroCounter.def
@@ -1,0 +1,75 @@
+/**
+ * Copyright 2017 Axel Huebl
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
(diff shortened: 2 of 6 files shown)


src/picongpu/include/particles/particleToGrid/derivedAttributes/Counter.hpp
```diff
@@ -20,7 +20,7 @@
 
 #pragma once
 
-#include "particles/particleToGrid/derivedAttributes/ChargeDensity.def"
+#include "particles/particleToGrid/derivedAttributes/Counter.def"
```
ax3l: hupsi. I was just correcting that on the way :)

ax3l:
@n01r does that help you? :)

src/picongpu/include/particles/particleToGrid/derivedAttributes/Counter.hpp
```diff
@@ -52,6 +52,6 @@ namespace derivedAttributes
         /* return attribute */
         return particleCounter;
     }
-} /* namespace derivedAttributes */
-} /* namespace particleToGrid */
-} /* namespace picongpu */
+}
```
n01r: Sorry :pray: for having sparked this discussion again  @ax3l , @psychocoderHPC   
I understand both points - more than once I found having these end-of-namespace comments convenient for reading - but it's more work and can bear the danger of being outdated

ax3l: doesn't matter, I just add it and follow the herd until you all realize Benjamin was right :P

n01r:
Thanks, and all only in **2:59 min** hacking time!

src/picongpu/include/particles/particleToGrid/derivedAttributes/MacroCounter.def
```diff
@@ -0,0 +1,75 @@
...
+         * This method is called on a per-thread basis (each thread of a block
+         * handles a particle of a frame).
+         *
+         * \tparam T_Particle particle in the frame
+         * \param particle particle in the frame
+         *
+         * \return new attribute for the particle (type \see T_AttributeType)
+         */
+        template< class T_Particle >
+        DINLINE float_X
```
n01r: Does it have advantages to keep this a `float` value?

ax3l: `FieldTmp` only supports `float_X`

PR #1767, opened 2017-01-13
fix shared memory alignment

psychocoderHPC:
fix #1766 I did not investigate the effect of this bug on other kernels than radiation.

- add type alignment to shared memory allocator

Tests:

- [x] broken bunch example from #1766

This bug was introduced with #1726. The new method to create shard memory in #1726 removed the alignment information from the shared memory, this could result in unaligned memory access errors.

PR #1795, opened 2017-01-31
fix clang compile

psychocoderHPC:
fix alignment definition of shared memory

This bug was introduced with #1767.
You can find more information about this issue in our similar [alpaka shared memory fix](https://github.com/ComputationalRadiationPhysics/alpaka/pull/318).

### DIFF
--- a/src/libPMacc/include/memory/shared/Allocate.hpp
+++ b/src/libPMacc/include/memory/shared/Allocate.hpp
@@ -56,7 +56,7 @@ namespace shared
         DINLINE T_Type &
         get()
         {
-            __shared__ alignas( alignof( T_Type ) ) uint8_t smem[ sizeof( T_Type ) ];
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
I tested the `alignas` placement for array's also with a small example for icc/15.2.

```
icc main.cpp -std=c++11
```
```C++
#include <typeinfo>
#include <cstdint>

template< typename T>
void test(T* foo)
{
        uint8_t shMem  alignas(alignof(T)) [ sizeof(T)];
        auto& sRef = *(reinterpret_cast<T*>(shMem));
        sRef.bar[0] = (*foo).bar[0];
        sRef.bar[0]++;
        (*foo).bar[0] = sRef.bar[0];
}

struct Foo
{
        int bar alignas(16) [3];
};

int main ()
{
    Foo foo;
    test(&foo);
    return 0;
}
```

PR #1796, opened 2017-02-02
allow to define the laser initialization plane

psychocoderHPC:
- `FieldE.kernel`: add laser scaling factor if `laser::initPlaneY != 0`
- `FieldE.tpp`:
  - add asserts to check conditions for the laser initialization plane
  - model: a change of initialization plane does not make a change in focus position or time-to-target necessary
- `FieldManipulation.hpp`: enable top absorber if `laser::initPlaneY != 0`
- `laserConfig.param`
  - add namespace laser
  - add parameter `initPlaneY`
- laser profiles (keep first top cell as origin for all laser parameters)
  - manipulate focus depending of the `initPlaneY`
  - shift the internal runtime based of `initPlaneY`

Todo:
- [x] fix typo at all places: `plain` -> `plane`
- [x] add constexpr `gridConfig.param`  change as independent pull request
- [x] discuss the changes with the first cell as origin for laser parameter
- [x] maybe that the scaling factor need to respect cell size X/Z for laser not moving in Y direction (#1893)
- [x] rebase against #1909

scaling factor
--------------
The current implementation calculated a scaling factor for the amplitude to respect that we initialize only the electrical field and not magnetic field. The result of the missing magnetic field is a wave which is moving in positive and negative Y direction. If we use no scaling factor the amplitude of the positive moving wave is too low, depending on the cell size in Y direction.
**Open question**: Do we need care of the laser phase to scale the amplitude for laser those not moving only in Y direction?

origin for laser parameter:
------------------------------------
What should be the origin for e.g. focus? Is it the laser initialization plane or the begin of the simulation.
If we define everything relative to the init plane than it is possible that the laser collide with the bottom outer border of the simulation in the case we enable sliding window.
The physicist need use the non intuitive init plane definition in cells (instead of SI meter) if we define laser parameter relative to the plane.

Test:
------
Gaussian beam focused to cell 300 (linear polarized in X)
==============
lasergaussianbeam_line_65
green = dev
blue = pull request (initialized at cell 65)

The amplitude for this pull request is ~1.5% over the defined value. The amplitude for the dev is ~1 under the defined value.

Plane wave test (up ramp)
============
plainwave_line_256_upramp
green = dev
blue = pull request (initialized at cell 65

Small shift in space for the positions of the amplitude.

Plane wave test (down ramp)
==============
plainwave_line_256_downramp

The plot shows the down ramp of the laser after a reflection on the right border. The pulse was long  enough that the reflected wave overlapped the init phase for a longer time. This should show how the laser initialization is influenced by the reflected wave.

CC-ing: @HighIander

### DIFF
--- a/examples/Bunch/include/simulation_defines/param/laserConfig.param
+++ b/examples/Bunch/include/simulation_defines/param/laserConfig.param
@@ -28,15 +28,15 @@ namespace laser
 {
 /** cell from top where the laser is initialized
  *
- * if `laserInitPlaneY == 0` than the absorber are disabled.
- * if `laserInitPlaneY > absorbercells negative Y` the negative absorber in y
+ * if `initPlaneY == 0` than the absorber are disabled.
+ * if `initPlaneY > absorbercells negative Y` the negative absorber in y
  * direction is enabled
  *
  * valid ranges:
- *   - laserInitPlaneY == 0
- *   - absorber cells negative Y < laserInitPlaneY < cells in y direction of the top gpu
+ *   - initPlaneY == 0
+ *   - absorber cells negative Y < initPlaneY < cells in y direction of the top gpu
 */
-constexpr uint32_t laserInitPlaneY = 0;
+constexpr uint32_t initPlaneY = 0;
 }
 
 namespace laserGaussianBeam
--- a/examples/Bunch/include/simulation_defines/param/laserConfig.param
+++ b/examples/Bunch/include/simulation_defines/param/laserConfig.param
@@ -24,6 +24,21 @@
 
 namespace picongpu
 {
+namespace laser
+{
+/** cell from top where the laser is initialized
+ *
+ * if `laserInitPlaneY == 0` than the absorber are disabled.
+ * if `laserInitPlaneY > absorbercells negative Y` the negative absorber in y
+ * direction is enabled
+ *
+ * valid ranges:
+ *   - laserInitPlaneY == 0
+ *   - absorber cells negative Y < laserInitPlaneY < cells in y direction of the top gpu
+*/
+constexpr uint32_t laserInitPlaneY = 0;
+}
+
 namespace laserGaussianBeam
 {
 // Asymetric sinus used: starts with phase=0 at center --> E-field=0 at center
--- a/src/picongpu/include/fields/FieldE.kernel
+++ b/src/picongpu/include/fields/FieldE.kernel
@@ -69,6 +69,8 @@ struct KernelLaserE
         //const int max_y_neighbors = Get<fieldSolver::FieldSolver::OffsetOrigin_E, 1 >::value;
         const int max_y_neighbors = 1;
 
+        cellOffset.y() += laser::laserInitPlaneY;
+        EFieldOffset.y() += laser::laserInitPlaneY;
         for (int totalOffsetY = 0; totalOffsetY < max_y_neighbors; ++totalOffsetY)
         {
             /** \todo Right now, the phase could be wrong ( == is cloned)
@@ -76,8 +78,25 @@ struct KernelLaserE
              *
              *  \todo What about the B-Field in the second plane?
              */
-            cellOffset.y() = totalOffsetY;
-            fieldE(EFieldOffset) = lMan.getManipulation(cellOffset);
+            if (laser::laserInitPlaneY != 0) // compile time if
+            {
+                /* If the laser is not initialized in the first cell we create a negative
+                 * and positive moving wave. Therefore we need to multiply the
+                 * amplitude with a correction factor depending of the cell size in
+                 * propagation direction.
+                 * The negative moving wave is filtered by the absorber.
+                 *
+                 * The `correctionFactor` assume that the wave is moving in y direction.
+                 */
+                auto correctionFactor = (SPEED_OF_LIGHT * DELTA_T) / CELL_HEIGHT * float_X(2.);
+                fieldE(EFieldOffset) +=  correctionFactor * lMan.getManipulation(cellOffset);
+            }
+            else
+            {
+                fieldE(EFieldOffset) = lMan.getManipulation(cellOffset);
+            }
+            cellOffset.y() += 1;
+            EFieldOffset.y() += 1;
         }
     }
 };
--- a/src/picongpu/include/fields/FieldE.kernel
+++ b/src/picongpu/include/fields/FieldE.kernel
@@ -46,6 +46,9 @@ struct KernelLaserE
         cellOffset.z() = (blockIdx.y * blockDim.y) + threadIdx.y;
         EFieldOffset.z() = cellOffset.z() + (MappingDesc::SuperCellSize::z::value * GUARD_SIZE);
 #endif
+        // shift offsets to the laser initialization plane
+        cellOffset.y() += laser::initPlaneY;
+        EFieldOffset.y() += laser::initPlaneY;
 
         /** Calculate how many neighbors to the left we have
          * to initialize the laser in the E-Field
@@ -65,12 +68,13 @@ struct KernelLaserE
          *
          * Problem: that's still not our case. For example our Yee does a
          *          dE = curlLeft(B) - therefor, we should init B, too.
+         *
+         *
+         *  @todo: might also lack temporal offset since our formulas are E(x,z,t) instead of E(x,y,z,t)
+         *  `const int max_y_neighbors = Get<fieldSolver::FieldSolver::OffsetOrigin_E, 1 >::value;`
          */
-        //const int max_y_neighbors = Get<fieldSolver::FieldSolver::OffsetOrigin_E, 1 >::value;
         const int max_y_neighbors = 1;
 
-        cellOffset.y() += laser::laserInitPlaneY;
-        EFieldOffset.y() += laser::laserInitPlaneY;
         for (int totalOffsetY = 0; totalOffsetY < max_y_neighbors; ++totalOffsetY)
         {
             /** \todo Right now, the phase could be wrong ( == is cloned)
@@ -78,13 +82,13 @@ struct KernelLaserE
              *
              *  \todo What about the B-Field in the second plane?
              */
-            if (laser::laserInitPlaneY != 0) // compile time if
+            if (laser::initPlaneY != 0) // compile time if
             {
-                /* If the laser is not initialized in the first cell we create a negative
-                 * and positive moving wave. Therefore we need to multiply the
+                /* If the laser is not initialized in the first cell we emit a
+                 * negatively and positively propagating wave. Therefore we need to multiply the
                  * amplitude with a correction factor depending of the cell size in
                  * propagation direction.
-                 * The negative moving wave is filtered by the absorber.
+                 * The negatively propagating wave is damped by the absorber.
                  *
                  * The `correctionFactor` assume that the wave is moving in y direction.
                  */
@@ -95,6 +99,7 @@ struct KernelLaserE
             {
                 fieldE(EFieldOffset) = lMan.getManipulation(cellOffset);
             }
+            /* move to next cell @see max_y_neighbors */
             cellOffset.y() += 1;
             EFieldOffset.y() += 1;
         }
--- a/src/picongpu/include/fields/FieldE.tpp
+++ b/src/picongpu/include/fields/FieldE.tpp
@@ -184,7 +184,7 @@ void FieldE::laserManipulation( uint32_t currentStep )
     /* initialize the laser not in the first cell is equal to a negative shift
      * in time
      */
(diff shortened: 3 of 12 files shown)


examples/Bunch/include/simulation_defines/param/laserConfig.param
```diff
@@ -25,6 +25,21 @@
...
+ *
+ * if `laserInitPlaneY == 0` than the absorber are disabled.
+ * if `laserInitPlaneY > absorbercells negative Y` the negative absorber in y
+ * direction is enabled
+ *
+ * valid ranges:
+ *   - laserInitPlaneY == 0
+ *   - absorber cells negative Y < laserInitPlaneY < cells in y direction of the top gpu
+*/
+constexpr uint32_t laserInitPlaneY = 0;
```
psychocoderHPC: Maybe this variable should only called `initPlaneY` because we are in the namespace `laser`

src/picongpu/include/fields/laserProfiles/laserGaussianBeam.hpp
```diff
@@ -38,9 +38,16 @@ namespace picongpu
...
-            const float_64 runTime = DELTA_T*currentStep;
+            /* initialize the laser not in the first cell is equal to a negative shift
+             * in time
+             */
+            constexpr float_X laserTimeShift = laser::laserInitPlaneY * CELL_HEIGHT / SPEED_OF_LIGHT;
+            const float_64 runTime = DELTA_T * currentStep - laserTimeShift;
             const float_64 f = SPEED_OF_LIGHT / WAVE_LENGTH;
 
+            /* calculate focus position relative to the laser initialization plane */
+            const float_X focusPos = FOCUS_POS - laser::laserInitPlaneY * CELL_HEIGHT / SPEED_OF_LIGHT;
```
PrometheusPi: ERROR:
`focusPos` is a position in space but `laser::laserInitPlaneY * CELL_HEIGHT / SPEED_OF_LIGHT` is a time.
Please fix it to 
```c++
const float_X focusPos = FOCUS_POS - laser::laserInitPlaneY * CELL_HEIGHT; 
```

src/picongpu/include/fields/laserProfiles/laserGaussianBeam.hpp
```diff
@@ -123,13 +130,15 @@ namespace picongpu
          */
         HDINLINE float3_X laserTransversal( float3_X elong, float_X phase, const float_X posX, const float_X posZ )
         {
+            /* calculate focus position relative to the laser initialization plane */
+            const float_X focusPos = FOCUS_POS - laser::laserInitPlaneY * CELL_HEIGHT / SPEED_OF_LIGHT;
```
PrometheusPi: ERROR:
same here - (location **not** time)
```c++
const float_X focusPos = FOCUS_POS - laser::laserInitPlaneY * CELL_HEIGHT; 
```

src/picongpu/include/fields/laserProfiles/laserPulseFrontTilt.hpp
```diff
@@ -38,9 +38,16 @@ namespace picongpu
...
-            const float_64 runTime = DELTA_T*currentStep;
+            /* initialize the laser not in the first cell is equal to a negative shift
+             * in time
+             */
+            constexpr float_X laserTimeShift = laser::laserInitPlaneY * CELL_HEIGHT / SPEED_OF_LIGHT;
+            const float_64 runTime = DELTA_T * currentStep - laserTimeShift;
             const float_64 f = SPEED_OF_LIGHT / WAVE_LENGTH;
 
+            /* calculate focus position relative to the laser initialization plane */
+            constexpr float_X focusPos = FOCUS_POS - laser::laserInitPlaneY * CELL_HEIGHT / SPEED_OF_LIGHT;
```
PrometheusPi: ERROR:
same here
remove `.../ SPEED_OF_LIGHT`

src/picongpu/include/fields/laserProfiles/laserPulseFrontTilt.hpp
```diff
@@ -92,41 +99,45 @@ namespace picongpu
         {
             //const float_X modMue = float_X(PI) * float_X(SPEED_OF_LIGHT / WAVE_LENGTH) * INIT_TIME;
             const float_X f = SPEED_OF_LIGHT / WAVE_LENGTH;
-            const float_X timeShift = phase / (float_X(2.0) * float_X(PI) * float_X(f)) + FOCUS_POS / SPEED_OF_LIGHT;
+            /* calculate focus position relative to the laser initialization plane */
+            constexpr float_X focusPos = FOCUS_POS - laser::laserInitPlaneY * CELL_HEIGHT / SPEED_OF_LIGHT;
```
PrometheusPi: ERROR:
same here
remove `.../ SPEED_OF_LIGHT`

src/picongpu/include/fields/FieldE.kernel
```diff
@@ -70,15 +70,34 @@ struct KernelLaserE
...
+            {
+                /* If the laser is not initialized in the first cell we create a negative
+                 * and positive moving wave. Therefore we need to multiply the
+                 * amplitude with a correction factor depending of the cell size in
+                 * propagation direction.
+                 * The negative moving wave is filtered by the absorber.
+                 *
+                 * The `correctionFactor` assume that the wave is moving in y direction.
+                 */
+                auto correctionFactor = (SPEED_OF_LIGHT * DELTA_T) / CELL_HEIGHT * float_X(2.);
```
psychocoderHPC: The factor two is only correct if `(SPEED_OF_LIGHT * DELTA_T) / CELL_HEIGHT == 1`. This means that a wave is moving within one time step with the half of its energy in negative and the other half in positive direction. 
In a yee solver the cell is `(SPEED_OF_LIGHT * DELTA_T) / CELL_HEIGHT < 1`, this means that not the full wave leafe the cell within a timestep.

src/picongpu/include/fields/FieldE.kernel
```diff
@@ -70,15 +70,34 @@ struct KernelLaserE
         //const int max_y_neighbors = Get<fieldSolver::FieldSolver::OffsetOrigin_E, 1 >::value;
         const int max_y_neighbors = 1;
 
+        cellOffset.y() += laser::laserInitPlaneY;
+        EFieldOffset.y() += laser::laserInitPlaneY;
         for (int totalOffsetY = 0; totalOffsetY < max_y_neighbors; ++totalOffsetY)
```
psychocoderHPC: please see the comment above.

PrometheusPi: As discussed *offline* today - this loop is fine (for other algorithms). :+1:

src/picongpu/include/fields/FieldE.kernel
PrometheusPi: Thanks for the detailed *offline* discussion today. What was the result of the tests?

psychocoderHPC:
@PrometheusPi I teste the plane wave with the half time step.

green = halfe time step
blue = test data from the runs above

# upramp
plainwave_line_256_upramp_halfe

# downramp
plainwave_line_256_downramp_halfe

psychocoderHPC:
ping

src/picongpu/include/fields/FieldE.kernel
psychocoderHPC: [click here to see the results](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1796#issuecomment-280610017)

ax3l:
> FieldE.kernel: add laser scaling factor if laser::initPlaneY != 0

note: also useful to add the scaling factor when periodic in Y is enabled

src/picongpu/include/fields/FieldE.kernel
PrometheusPi: great :+1: - looks perfectly okay

ax3l:
> What should be the origin for e.g. focus? Is it the laser initialization plane or the begin of the simulation.

I like your model to keep the origin, focus position and time-to-target constant, independent of initialization plane. That means: everything is relative to our origin, great :+1:

ax3l:
> maybe that the scaling factor need to respect cell size X/Z for laser not moving in Y direction

good point, I think it might actually be ok since the excitation is spatially limited/orginiated by the plane normal to Y but we should test and verify that, ideally with an off-axis lineout or 2D difference

src/picongpu/include/fields/FieldE.kernel
```diff
@@ -76,8 +79,25 @@ struct KernelLaserE
...
              */
-            cellOffset.y() = totalOffsetY;
-            fieldE(EFieldOffset) = lMan.getManipulation(cellOffset);
+            if (laser::initPlaneY != 0) // compile time if
+            {
+                /* If the laser is not initialized in the first cell we create a negative
+                 * and positive moving wave. Therefore we need to multiply the
+                 * amplitude with a correction factor depending of the cell size in
+                 * propagation direction.
+                 * The negative moving wave is filtered by the absorber.
```
ax3l: The negative*ly* *propagating* wave is *damped* by the absorber.

src/picongpu/include/fields/FieldE.kernel
```diff
@@ -76,8 +79,25 @@ struct KernelLaserE
              *
              *  \todo What about the B-Field in the second plane?
              */
-            cellOffset.y() = totalOffsetY;
-            fieldE(EFieldOffset) = lMan.getManipulation(cellOffset);
+            if (laser::initPlaneY != 0) // compile time if
+            {
+                /* If the laser is not initialized in the first cell we create a negative
```
ax3l: we *emit* a *negatively* and *positively* *propagating* wave

src/picongpu/include/fields/FieldE.kernel
```diff
@@ -76,8 +79,25 @@ struct KernelLaserE
...
+                 * and positive moving wave. Therefore we need to multiply the
```
ax3l: *Since this effectively splits the emitted energy, we have* to ...

src/picongpu/include/fields/FieldE.tpp
```diff
@@ -187,13 +187,37 @@ void FieldE::laserManipulation( uint32_t currentStep )
...
+        Environment<simDim>::get().GridController().getCommunicationMask( ).isSet( TOP ) || numSlides != 0
+    )
+    {
+        return;
+    }
+    else
+    {
+        PMACC_VERIFY_MSG(
+            laser::initPlaneY < static_cast<uint32_t>( Environment<simDim>::get().SubGrid().getLocalDomain().size.y() ),
+            "initPlaneY must be located in the top GPU"
```
ax3l: I think we can relax that constrain

src/picongpu/include/fields/FieldE.tpp
```diff
@@ -187,13 +187,37 @@ void FieldE::laserManipulation( uint32_t currentStep )
...
+        );
+    }
+
+    PMACC_CASSERT_MSG(
+        __initPlaneY_need_to_be_greate_than_the_top_absorber_cells_or_zero,
```
ax3l: need*s*

src/picongpu/include/fields/FieldManipulator.hpp
```diff
@@ -64,13 +64,19 @@ class FieldManipulator
 
                 if (thickness == 0) continue; /*if the absorber has no thickness we check the next side*/
 
-                /* disable the absorber on top side if
-                 *      no slide was performed and
-                 *      laser init time is not over
+                /* allow to disable the absorber on the top side if the laser
+                 * initialization plane in y direction is *not* in cell zero
```
ax3l: I think it's vice versa: you disable the absorber *if*  the init plane is == 0

src/picongpu/include/fields/FieldManipulator.hpp
```diff
@@ -64,13 +64,19 @@ class FieldManipulator
...
+                if (laser::initPlaneY == 0)
                 {
-                    if (i == TOP) continue; /*disable laser on top side*/
+                    /* disable the absorber on top side if
+                     *      no slide was performed and
+                     *      laser init time is not over
+                     */
+                    if (numSlides == 0 && ((currentStep * DELTA_T) <= laserProfile::INIT_TIME))
+                    {
+                        if (i == TOP) continue; /*disable laser on top side*/
```
ax3l: this comment was and is still wrong. we disable the absorber here, not the laser

psychocoderHPC:
updated

src/picongpu/include/fields/FieldE.tpp
psychocoderHPC: added a new issue #1891

psychocoderHPC:
>> FieldE.kernel: add laser scaling factor if laser::initPlaneY != 0

>note: also useful to add the scaling factor when periodic in Y is enabled

opened a new issue #1892

psychocoderHPC:
updated

PR #1800, opened 2017-02-07
DataConnector: Fix Includes

ax3l:
Fix missing `DataConnector` includes and clean up:

- include order:
  - header/forward declaration of same file
  - PIConGPU includes
  - PMacc includes
  - other external library includes (if not required to be included first: exceptions are mpi/boost.fusion/cuda/...)
  - boost includes (because its a very common, external C++ lib)
  - stdlib includes
- correct style via `"`
- correct spaces (one before pragma once, 2 after all includes)
- remove includes in PIConGPU core which are already included via `simulation_defines.hpp`

### DIFF
--- a/src/libPMacc/include/mappings/simulation/ResourceMonitor.tpp
+++ b/src/libPMacc/include/mappings/simulation/ResourceMonitor.tpp
@@ -22,13 +22,14 @@
 
 // PMacc
 #include "Environment.hpp"
-#include <particles/operations/CountParticles.hpp>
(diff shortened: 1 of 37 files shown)


psychocoderHPC:
please rebase

ax3l:
cool, I wanted to change your include order in that file anyway but did let it through in your PR ^^

updated! :)

ax3l:
@psychocoderHPC reviewable :)

PR #1801, opened 2017-02-07
DataConnector: Shared Pointer Storage

ax3l:
Rebuilds the PMacc `DataConnector` to serve as a storage to share `shared_ptr`s with.

## Motivation

Previously, resources in simulations that are only created once and used from several places needed to be held by a "master" class which keeps them "alive" in form of members. This is cumbersome and leads in most cases to a member in `MySimulation` (which is a class we need to reduce in complexity anyway to allow better APIs to it and more user-centric flexibility).

Additionally, this design made it hard to call a functor which calls a factory, e.g. for each particle species, and then allocating new fields *without returning those objects* as done in `CallXY` functors called by `ForEach` loops for N species within `particles/ParticlesFunctors.hpp`. I currently need that functionality since I want to add additional helper fields for a user-selected subset of species without adding more book-keeping in `MySimulation` of the `Particle`s themselves.

The new `DataConnector`, which is as before a Singleton, now serves as an easily accessible storage that can share (e.g. as the last owner) access to a resource which avoids creating an other singleton for each data set with singleton-characteristics per simulation that we have already.

## Why Not References / Pointers?

In contrast to exchanging references as before (or plain pointers) one can now keep track of the number of shared accesses to a resource. The syntax of exchanging a smart-pointer is also slicker as reference passing, which is why we did unwind the references to (plain) pointers in most-places already.

## Other Changes

One can now deregister selected resources and the member namings have been cleaned up to avoid word-doubling, etc.

The sorting abstraction was removed since we only need as before a strictly sorted FILO storage. A `vector` is preferred over a `list` for slight performance benefits (re-allocations during sharing a data resource initially is only done once, have less performance penalty then one would think, otherwise we can still `reserve()` and following frequent find-and-access is the main usage).

Updates the usage of shared resources via the `DataConnector` inside PIConGPU and removes "owned" pointers within classes where those are actually only accessed at one specific and local part of an algorithm, plugin or similar.

## To Do

- [x] figure out and fix RT crash (push: `getDeviceParticleBox`)
- [x] test `softRestart`
- [x] heating test
- [x] speed test: khi without plugins
- [x] update GOL? -> no update needed (`DataConnector` not used)


psychocoderHPC:
The error is triggered by the invalidate call in [SimulationHelper](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/libPMacc/include/simulationControl/SimulationHelper.hpp#L134)

ax3l:
> The error is triggered by the invalidate call in SimulationHelper

good catch!

offline discussion results for documentation purposes: initially, we used this call to avoid syncing data multiple times to host. this has several drawbacks, including the assumption that at that magic place in a step the data does not change any more and is only changed on device in general. In critical situations, such as the [mallocMC buffer for I/O](https://github.com/ComputationalRadiationPhysics/picongpu/blob/2603b97c8f9fc6a71610394455a747f685fbff85/src/picongpu/include/plugins/adios/ADIOSWriter.hpp#L815) we are already taking care of avoiding double-syncs independently.

re-implementing such functionality with less assumptions is then a job for our new and upcoming resource manager.

src/picongpu/include/simulationControl/MySimulation.hpp
```diff
@@ -246,21 +246,23 @@ class MySimulation : public SimulationHelper<simDim>
...
 
-        __delete(fieldE);
+        __delete(myFieldSolver);
+
+        dc.unshare( fieldB->getUniqueId() );
+
+        dc.unshare( fieldE->getUniqueId() );
 
-        __delete(fieldJ);
+        dc.unshare( fieldJ->getUniqueId() );
```
psychocoderHPC: Note: currently we need to unshare all buffers else we can get problems with out PMacc singletons (buffer can life longer than we hold the GPU). If we add a clean interface to start/finalize MPI and release the GPU after all buffers, events and streams are destroyed we can skip this unshares.

ax3l:
## Speed Test :white_check_mark: 

### dev *without plugins*

```
[1,0]<stdout>:100 % =     2000 | time elapsed:       4min  2sec 476msec | avg time per step: 153msec
[1,0]<stdout>:calculation  simulation time:  4min  2sec 476msec = 242 sec
[1,0]<stdout>:full simulation time:  4min  7sec 644msec = 247 sec
used walltime=00:04:35
```


### PR *without plugins*

```
[1,0]<stdout>:100 % =     2000 | time elapsed:       4min  2sec 464msec | avg time per step: 153msec
[1,0]<stdout>:calculation  simulation time:  4min  2sec 464msec = 242 sec
[1,0]<stdout>:full simulation time:  4min  6sec 964msec = 246 sec
used walltime=00:04:34
```

ax3l:
## Heating & Charge Conservation Tests :white_check_mark: 

`python $PICSRC/src/tools/bin/plot_chargeConservation_overTime.py khi-dev/ khi-dc/`:
chargeconservation

`python $PICSRC/src/tools/bin/plotNumericalHeating khi-dev/ khi-dc/`:
heating


### dev

```
[1,0]<stdout>:100 % =     2000 | time elapsed:       7min 27sec 243msec | avg time per step: 153msec
[1,0]<stdout>:calculation  simulation time:  7min 52sec 227msec = 472 sec
[1,0]<stdout>:full simulation time:  7min 58sec 974msec = 478 sec
used walltime=00:08:42
```

### PR

```
[1,0]<stdout>:100 % =     2000 | time elapsed:       7min 53sec 779msec | avg time per step: 153msec
[1,0]<stdout>:calculation  simulation time:  8min 22sec 762msec = 502 sec
[1,0]<stdout>:full simulation time:  8min 29sec 180msec = 509 sec
used walltime=00:09:12
```

src/libPMacc/include/dataManagement/DataConnector.hpp
```diff
@@ -154,68 +154,86 @@ namespace PMacc
...
+        ~DataConnector()
         {
-            std::map<SimulationDataId, Dataset*>::const_iterator iter;
-            for (iter = datasets.mapping.begin(); iter != datasets.mapping.end(); iter++)
-                delete iter->second;
-
-            if (datasets.sorter != nullptr)
+            log< ggLog::MEMORY >( "DataConnector: being destroyed (%1% datasets left to destroy)" ) %
+                                datasets.size();
+            //datasets.clear();
```
ax3l: @n01r little tear-down bug here that I discovered yesterday.
use this line and remove the for loop or change the for loop for now (RNG seems to forget to unshare itself)

Issue #1802, opened 2017-02-08
Ionization Params: Verify Vector Lengths


ax3l:
@HighIander and me created Carbon ionization energies and effective Z with wrong input which was not covered by input validation and resulted in tricky NaN's at runtime. We can do a bit better here validating the input.

Used release: 0.2.2

By copying from nitrogen, we did this:
```C++
    /* Example: carbon */
    PMACC_CONST_VECTOR(float_X,
      7, // ooops
      Z_EFFECTIVE_CARBON,
        /* 2p */
        3.136, // oops
        3.136,
        3.136,
        /* 2s */
        3.217,
        3.217,
        /* 1s */
        5.673,
        5.673
        );
```

and that:
```C++
    /* ionization energy for carbon atomic units */
    PMACC_CONST_VECTOR(float_X,
      7, // oops
      IONIZATION_ENERGY_CARBON,
        11.2603 * UNITCONV_eV_to_AU,
        24.3845 * UNITCONV_eV_to_AU,
        47.88778 * UNITCONV_eV_to_AU,
        64.49351 * UNITCONV_eV_to_AU,
        392.0905 * UNITCONV_eV_to_AU,
        489.993177 * UNITCONV_eV_to_AU,
    );
```

Two things we should not do:
- ionization: static assert that the vector length of `ionizationEnergies` and `effectiveAtomicNumbers` are exactly the proton number
- PMACC_CONST_VECTOR: allows to specify *more* values then length (which are ignored) but should not allow *less* values then length (because they are then uninitialized

ax3l:
@n01r regarding the static assert: we could also allow to use a vector that is *smaller* then the proton number, since for field ionization methods and high-Z targets we might want to spare us from adding ionization energies that we won't reach anyway. can you add that functionality?

PR #1803, opened 2017-02-08
ConstVector: Check Arguments Init Full Length

psychocoderHPC:
fix the first part of #1802 

`ConstVector` allows that the number of given arguments is lesser than the dimentsion
of the vector it self. This results in unitialized values.

solution: add static assert to avoid wrong usage

### DIFF
--- a/src/libPMacc/include/math/ConstVector.hpp
+++ b/src/libPMacc/include/math/ConstVector.hpp
@@ -65,6 +65,9 @@ namespace PMACC_JOIN(pmacc_static_const_storage,id)                            \
     template<typename T_Type, int T_Dim>                                       \
     struct ConstArrayStorage                                                   \
     {                                                                          \
+        PMACC_CASSERT_MSG(                                                     \
+            __PMACC_CONST_VECTOR_dimension_needs_to_be_less_than_or_equal_to_the_number_of_arguments__, \
+            Dim <= count );                                                    \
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/math/ConstVector.hpp
```diff
@@ -65,6 +65,9 @@ namespace PMACC_JOIN(pmacc_static_const_storage,id)                            \
     template<typename T_Type, int T_Dim>                                       \
     struct ConstArrayStorage                                                   \
     {                                                                          \
+        PMACC_CASSERT_MSG(                                                     \
+            __PMACC_CONST_VECTOR_dimension_needs_to_be_lesser_or_equal_than_the_given_number_of_arguments__, \
```
n01r: `...needs_to_be_less_or...`

n01r: Sorry! It has to be `...less_than_or_equal_to_the_number...`.

theZiz: Dafuq? This is the ugliest macro I ever saw!

ax3l: [BOOST_MPL_ASSERT_MSG](http://www.boost.org/doc/libs/1_63_0/libs/mpl/doc/refmanual/assert-msg.html) ;)

PR #1805, opened 2017-02-08
DataSpaceOperations: Simplify Formula

ax3l:
This is a left-over from my 4D experiments last week: simplify and align the used formula for ND mapping.

Makes it straight forward to extent mapping to arbitrary dimensions whenever needed.

Yes, alignments are intentional, "because it's math" :)

@BenjaminW3 might be useful for alpaka, too.

### DIFF
--- a/src/libPMacc/include/dimensions/DataSpaceOperations.hpp
+++ b/src/libPMacc/include/dimensions/DataSpaceOperations.hpp
@@ -150,23 +150,33 @@ namespace PMacc
         template<class TVEC>
         static HDINLINE DataSpace<DIM2> map(uint32_t pos)
         {
-            return DataSpace<DIM2 > (pos % TVEC::x::value, pos / TVEC::x::value);
+            return DataSpace< DIM2 >(
+                pos % TVEC::x::value,
+                pos / TVEC::x::value
+            );
         }
 
         template<class TVEC>
         static HDINLINE uint32_t map(const DataSpace<DIM2>& pos)
         {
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
Could you please run a test!

ax3l:
Yep! :)
numericalheating
chargeovertime

BenjaminW3:
Yes, the 1D to 3D optimization would be really usefull as it is needed for each thread when using OpenMP.
How to prove `(pos % (size.x() * size.y())) / size.x() == (pos / size.x()) % size.y()`?
How to generalize this for 1D to ND mapping?

BenjaminW3:
The current alpaka calculation precalculates `size.x() * size.y()` because it is also required to compute the x-index. So this does not reduce the number of operations in this optimized version. I just reduces the size of the modulant.
I do not know how `%` is  implemented, but in the most naive way, a small modulant is even contraproductive.

ax3l:
I'll post the ND form of it tomorrow a bit clearer, for map it's just:

current dim id = id / (volume of dimensions before this one) % current dim size

Shall I sent you a PR to alpaka?

BenjaminW3:
Please send a PR if you want 😃

ax3l:
Sure, can you just point me to the files to patch?

BenjaminW3:
You can find the equivalent in alpaka under `include\alpaka\idx\MapIdx.hpp. Please not the inverted (mathemaical) index order.

PR #1806, opened 2017-02-08
TBG: Outdated Header

ax3l:
Updates the first doc lines in tbg to be a bit more up-to-date

### DIFF
--- a/src/tools/bin/tbg
+++ b/src/tools/bin/tbg
@@ -19,15 +19,11 @@
 # If not, see <http://www.gnu.org/licenses/>.
 #
 
-# This script parse a tpl file and combine data from cfg file to start a batch job
-# in a PBS based batch system
-# example: ./tpg qsub testRun hypnos/hypnos.tpl hypnos/picongpu.cfg
-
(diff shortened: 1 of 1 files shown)


src/tools/bin/tbg
```diff
@@ -19,15 +19,11 @@
...
-# This script parse a tpl file and combine data from cfg file to start a batch job
-# in a PBS based batch system
-# example: ./tpg qsub testRun hypnos/hypnos.tpl hypnos/picongpu.cfg
-
-#$1        - submit command e.g. qsub
-#$2        - job name (must uniqu to older jobs)
-#$3        - tpl file
-#$4        - cfg file
-
+# This script parses template files for a batch systems and fills them
```
n01r: > for a batch systems

*for batch systems* or *for a batch system*
why was the usage example removed?

ax3l: dang.

the example was outdated, those options are not positional arguments anymore. `--help` provides the correct syntax.

n01r: a'ight thx

PR #1811, opened 2017-02-08
fix wrong sign of delta_angle in radiation observer direction

PrometheusPi:
This pull request fixes a wrongly set `delta_angle_theta` and `delta_angle_phi` calculation in the default `radiationObserver.param`. To avoid a miss-configuration if the default param file is used in a setup, this should be fixed.


BeyondEspresso:
This line is in conflict with the description in the comments, since the center of the observer field points in the x-direction instead into the y-direction. Please open an additional pull request for this.
`` return vector_64( sinTheta*sinPhi , sinTheta*cosPhi , cosTheta ) ;``

PR #1817, opened 2017-02-10
fix possible NAN momenta in ionization

### LINKED ISSUE #1810: Small Weightings: Ionizer Causing NaNs
Using two species (C, N) both cloned from another species (protons), each using ionizer BSIEffectiveZ, produces NaN's in the inCellPosition when the laser reaches the plasma.
With the same paramSet, disabling one of the ionizers works fine (at least seemingly).

Release: 0.2.2
System: Hypnos, K20 with
```
        module load gcc/4.9.2
        module load cmake/3.3.0
        module load boost/1.62.0
        module load cuda/8.0
        module load openmpi/1.8.6.kepler.cuda80

        # Plugins (optional)
        module load pngwriter/0.5.6
        module load hdf5-parallel/1.8.15 libsplash/1.6.0
```

Executable example: [flat foil](https://github.com/ComputationalRadiationPhysics/projects-ions/tree/topic-20170206-LCDUltrathinFoil) (internal, see [issue for a list of patches](https://github.com/ComputationalRadiationPhysics/projects-ions/issues/14)).

psychocoderHPC:
fix #1810

`WriteElectronIntoFrame` creates NAN's for the electron momenta if the weighting of the ion
is less than one.
If the ion weighting is greater than one the electron momentum is not exact.

- [x] tested with foil example and ions species weighting less one.

### DIFF
--- a/src/picongpu/include/particles/ionization/ionizationMethods.hpp
+++ b/src/picongpu/include/particles/ionization/ionizationMethods.hpp
@@ -63,7 +63,7 @@ namespace ionization
 
             /* each thread sets the multiMask hard on "particle" (=1) */
             childElectron[multiMask_] = 1;
-            const uint32_t weighting = parentIon[weighting_];
+            const float_X weighting = parentIon[weighting_];
 
             /* each thread initializes a clone of the parent ion but leaving out
              * some attributes:


src/picongpu/include/particles/ionization/ionizationMethods.hpp
```diff
@@ -63,7 +63,7 @@ namespace ionization
 
             /* each thread sets the multiMask hard on "particle" (=1) */
             childElectron[multiMask_] = 1;
-            const uint32_t weighting = parentIon[weighting_];
```
ax3l: ouch, big cast fail.
shouldn't that have caused a warning?

psychocoderHPC: No because we allow implicit cast, I am searching for a flag to disallow implicit floating point to int cast.
Disallow all implicit cast is currently to much :-(

ax3l: fun fact: in this routine, one could skip reading `weighting` all together since only a `getMass` ratio is calculated.

psychocoderHPC: @ax3l  but if mass is an per particle attribute this is correct.

ax3l: but since all masses are scaled equally, and both are weighted with the same `weighting` this could still be skipped if one wants to spare the read. just for future notice.

n01r:
Dang, where is my hat for that occasion :disappointed_relieved: .
Nicely spotted! Thanks! :+1:

PR #1821, opened 2017-02-10
WarmCopper Example: FLYlite Benchmark

ax3l:
This setup initializes a homogenous, non-moving, copper block irradiated by a laser with 10^18 W/cm^3 as a benchmark for [SCFLY](https://dx.doi.org/10.1016/j.hedp.2007.02.001) (FLYlite) atomic population dynamics.

We follow the setup from [FLYCHK page 10, figure 4](https://dx.doi.org/10.1016/j.hedp.2005.07.001) assuming a quasi 0D setup with homogenous density of a 1+ ionized copper target. The laser (not modeled) already generated a thermal electron density at 10, 100 or 1000 eV and a delta-distribution like "hot" electron distribution with 200 keV (directed stream).

The observable of interest is <Z> over time of the copper ions. For low thermal energies, collisional excitation, de-excitation and recombinations should be sufficient to reach the LTE state after about 0.1-1 ps. For higher initial temperatures, radiative rates get more relevant and the Non-LTE steady-state solution can only be reached correctly when also adding radiative rates.

### DIFF
--- a/examples/WarmCopper/include/simulation_defines/param/gridConfig.param
+++ b/examples/WarmCopper/include/simulation_defines/param/gridConfig.param
@@ -1,0 +1,96 @@
+/* Copyright 2013-2017 Axel Huebl, Rene Widera, Benjamin Worpitz
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+
+namespace picongpu
+{
+
+    namespace SI
+    {
+        /** Duration of one timestep
+         *
+         *  Non-LTE steady state or LTE respectively should be reached
+         *  within 1 ps. We discretize the time in 1'000 equal steps.
+         *
+         *  unit: seconds */
+        constexpr float_64 DELTA_T_SI = 1.0e-12 / 1000.;
+
+        /** epsilon to be >= 1 per mill close to CFL */
+        constexpr float_64 EPS_CFL = 1.001;
+        constexpr float_64 SQRT_OF_3 = 1.73205 * EPS_CFL;
+
+        /** equals X
+         *
+         *  note: formulated to automatically fullfill the 3D CFL criteria for Yee
+         *
+         *  unit: meter */
+        constexpr float_64 CELL_WIDTH_SI = DELTA_T_SI * SPEED_OF_LIGHT_SI * SQRT_OF_3;
+        /** equals Y - the laser & moving window propagation direction
+         *  unit: meter */
+        constexpr float_64 CELL_HEIGHT_SI = CELL_WIDTH_SI;
+        /** equals Z
+         *  unit: meter */
+        constexpr float_64 CELL_DEPTH_SI = CELL_WIDTH_SI;
+
+        /** Note on units in reduced dimensions
+         *
+         * In 2D3V simulations, the CELL_DEPTH_SI (Z) cell length
+         * is still used for normalization of densities, etc.
+         *
+         * A 2D3V simulation in a cartesian PIC simulation such as
+         * ours only changes the degrees of freedom in motion for
+         * (macro) particles and all (field) information in z
+         * travels instantaneous, making the 2D3V simulation
+         * behave like the interaction of infinite "wire particles"
+         * in fields with perfect symmetry in Z.
+         */
+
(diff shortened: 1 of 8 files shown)


examples/WarmCopper/include/simulation_defines/param/gridConfig.param
```diff
@@ -0,0 +1,86 @@
...
+
+    namespace SI
+    {
+        /** Duration of one timestep
+         *  unit: seconds */
+        constexpr float_64 DELTA_T_SI = 1.0e-12 / 1000.;
+
+        /** equals X
+         *  unit: meter */
+        constexpr float_64 CELL_WIDTH_SI = DELTA_T_SI * 2.998e8 * 1.8;
```
psychocoderHPC: please explain the magic numbers `2.998e8 * 1.8`

examples/WarmCopper/include/simulation_defines/param/gridConfig.param
```diff
@@ -0,0 +1,86 @@
...
+
+
+namespace picongpu
+{
+
+    namespace SI
+    {
+        /** Duration of one timestep
+         *  unit: seconds */
+        constexpr float_64 DELTA_T_SI = 1.0e-12 / 1000.;
```
psychocoderHPC: Why not `1.e-9`?

ax3l:
interesting, only building with `module load gcc/4.9.4 boost/1.57.0 cmake/3.3.0 cuda/7.5.18 openmpi/1.10.4` at the CI works...

examples/WarmCopper/include/simulation_defines/param/gridConfig.param
PrometheusPi: Please write it as 
```c++
constexpr float_64 SPEED_OF_LIGHT_SI = 2.9979e8;
constexpr float_64 SQRT_OF_3 = 1.732;
constexpr float_64 CELL_WIDTH_SI = DELTA_T_SI * SPEED_OF_LIGHT_SI * SQRT_OF_3;
```
This makes clear that you just fulfill the CFL condition for cubic cells.

examples/WarmCopper/include/simulation_defines/param/particleConfig.param
```diff
@@ -0,0 +1,112 @@
...
+
+    constexpr uint32_t TYPICAL_PARTICLES_PER_CELL = 2;
+
+namespace manipulators
+{
+
+    CONST_VECTOR(float_X, 3, DriftParam_direction, 1.0, 0.0, 0.0);
+    struct DriftParam
+    {
+        static constexpr float_64 gamma = 1.39139;
```
PrometheusPi: Please add a comment that this drift is applied to the 200keV electrons.

examples/WarmCopper/include/simulation_defines/param/speciesDefinition.param
```diff
@@ -0,0 +1,217 @@
...
+ * It is wrapped into a struct because of C++ restricting floats from being
+ * template arguments.
+ *
+ * Do not forget to set the correct mass of the atom in
+ * @see physicalConstants.param !
+ */
+struct Copper
+{
+    static constexpr float_X numberOfProtons  = 29.0;
+    static constexpr float_X numberOfNeutrons = 34.0;
```
PrometheusPi: QUESTION:
What is with the 30% Cu^65?

examples/WarmCopper/include/simulation_defines/param/gridConfig.param
ax3l: I am going for 1 pico second splitted into 1000 steps :D

examples/WarmCopper/include/simulation_defines/param/speciesDefinition.param
ax3l: I just decided to stick with the most common isotope for this test.

PrometheusPi: Okay 👍

examples/WarmCopper/include/simulation_defines/param/particleConfig.param
ax3l: this is directly documented in the line below where this struct is used via `Assign200keVDrift`, but I can add it...

```C++
// ...
        static constexpr float_64 gamma = 1.39139;
        const DriftParam_direction_t direction;
    };
    /* definition of SetDrift start*/
    using Assign200keVDrift = DriftImpl< DriftParam, nvidia::functors::Assign >;
```

ax3l: I will just rename the struct to `Drift200keVParam`

ax3l:
@PrometheusPi @psychocoderHPC comments addressed!

I am still expecting the compile suite to fail with a segfault for being out of resources, but review wise everything is addressed.

ax3l:
it succeeded! :)
I disabled ISAAC for this example to reduce the compile complexity. Some resources or the marvelous antivir on the compile node probably did run out before.

theZiz:
You can also reduce compile time complexity of ISAAC by deactivating functors or reducing the maximum amount of functors per functor chain. :wink:

ax3l:
good point, it's just for now for this specific case until we have a new compile suite. ISAAC not really useful for this example anyway, since it's static.

theZiz:
Thought so, just for the future if you want ISAAC being enabled for some tests. :full_moon_with_face:

ax3l:
Oh it is, in all other tests :)

PR #1828, opened 2017-02-14
Fix comments and definition in default radiationObserver.param

PrometheusPi:
This fixes a miss-definition in the default `radiationObserver.param` discovered by @BeyondEspresso in pull request #1811 .


PrometheusPi:
sorry - should not go to master

PR #1841, opened 2017-02-15
Copyright Header is no Doxygen

ax3l:
The copyright header is no doxygen and if put as such, spams the doc of what ever comes after it (most of the time it's appended to the `namespace picongpu` documentation).

```bash
grep -iR "\* Copyright" examples | awk -F: '{print $1}' | sort -u | xargs -n1 -P1 -I{} sed -i '1d' {}
grep -iR "\* Copyright" examples | awk -F: '{print $1}' | sort -u | xargs -n1 -P1 -I{} sed -i '1 s/ \*/\/\*/' {}

GIT_AUTHOR_NAME="Tools" GIT_AUTHOR_EMAIL="picongpu@hzdr.de" git commit -a -S
```

### DIFF
--- a/src/picongpu/include/plugins/PhaseSpace/PhaseSpaceMulti.tpp
+++ b/src/picongpu/include/plugins/PhaseSpace/PhaseSpaceMulti.tpp
@@ -1,4 +1,4 @@
-/** Copyright 2013-2017 Axel Huebl
+/* Copyright 2013-2017 Axel Huebl
  *
  * This file is part of PIConGPU.
  *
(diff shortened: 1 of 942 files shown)


src/picongpu/include/plugins/PhaseSpace/PhaseSpaceMulti.tpp
```diff
@@ -1,4 +1,3 @@
-/** Copyright 2013-2017 Axel Huebl
```
psychocoderHPC: Missing copyright and comment begin `/*`

PR #1844, opened 2017-02-15
Ionization: Fix Charge of Ionized Ions

ax3l:
The charge of all but hydrogen ion species, as seen by the PIC algorithm, was wrong after ionizing an atom/ion.

Accidentially, the charge state inside our `GetCharge` trait was calculated as
```
    proton number * ( proton number - bound electrons)
```
instead of
```
    proton number - bound electrons
```

in case `chargeRatio` of the species in `speciesDefinition.param` was different from `1.0` (we usually set it to the fully ionized species' charge density).

**Affected Setups**

If all of the following criteria are true for at least one of your ion species, your simulation is affected:

- your species in `speciesDefinition.param` used an `ionizer`
- your species is not hydrogen
- your species moves and/or creates currents (pusher, current solver)
- you defined `chargeRatio` for that species different from `1.0`

Pre-ionized setups without ionization physics are unaffected.

**Implications**

The following quantities were falsely calculated:

- plasma dynamics (pusher used wrong charge to mass ratio,
  current solver used wrong charge)
- outputs to files (`fields/<species>_chargeDensity`)

Note that the wrong plasma dynamics resulting from that highly overcharged ion species is a severe bug and will likely render the complete simulation unphysical if the above mentioned criteria are met.

**Kudos**

Thanks to @HighIander for reporting the issue.

### DIFF
--- a/src/picongpu/include/traits/attribute/GetCharge.hpp
+++ b/src/picongpu/include/traits/attribute/GetCharge.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2014-2017 Rene Widera
+/* Copyright 2014-2017 Rene Widera, Axel Huebl
  *
  * This file is part of PIConGPU.
  *
@@ -47,25 +47,26 @@ struct LoadBoundElectrons
     /** Functor implementation
      *
      * \tparam T_Particle particle type
-     * \param singlyChargedResult charge resulting from multiplying a single
-     * electron charge (positive OR negative) by the macro particle weighting
+     * \param weighting the particle's weighting
      * \param particle particle reference
      */
     template<typename T_Particle>
-    HDINLINE float_X operator()(const float_X singlyChargedResult, const T_Particle& particle)
+    HDINLINE float_X operator()(const float_X weighting, const T_Particle& particle)
     {
         const float_X protonNumber = GetAtomicNumbers<T_Particle>::type::numberOfProtons;
 
-        return singlyChargedResult * (protonNumber - particle[boundElectrons_]);
+        /* note: ELECTRON_CHARGE is negative and the second term is also negative
+         */
+        return
+            ELECTRON_CHARGE *
+            ( particle[boundElectrons_] - protonNumber ) *
+            weighting;
     }
 };
 
 /**  Calculate the real charge of a particle
  *
  * This is the fallback implementation if no `boundElectrons` are available for a particle
- *
- * \tparam T_HasBoundElectrons boolean that describes if species allows multiple charge states
- * due to bound electrons
  */
 template<>
 struct LoadBoundElectrons<false>
@@ -73,14 +74,13 @@ struct LoadBoundElectrons<false>
     /** Functor implementation
      *
      * \tparam T_Particle particle type
-     * \param singlyChargedResult charge resulting from multiplying a single
-     * electron charge (positive OR negative) by the macro particle weighting
+     * \param weighting the particle's weighting
      * \param particle particle reference
      */
     template<typename T_Particle>
-    HDINLINE float_X operator()(const float_X singlyChargedResult, const T_Particle& particle)
+    HDINLINE float_X operator()(const float_X weighting, const T_Particle&)
     {
-        return singlyChargedResult;
+        return frame::getCharge< typename T_Particle::FrameType >() * weighting;
     }
 };
 } // namespace detail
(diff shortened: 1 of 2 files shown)


src/picongpu/include/traits/attribute/GetCharge.hpp
```diff
@@ -48,40 +48,40 @@ struct LoadBoundElectrons
...
     template<typename T_Particle>
-    HDINLINE float_X operator()(const float_X singlyChargedResult, const T_Particle& particle)
+    HDINLINE float_X operator()(const float_X weighting, const T_Particle& particle)
     {
         const float_X protonNumber = GetAtomicNumbers<T_Particle>::type::numberOfProtons;
 
-        return singlyChargedResult * (protonNumber - particle[boundElectrons_]);
+        return
+            ELECTRON_CHARGE *
+            ( particle[boundElectrons_] - protonNumber ) *
```
n01r: I would add a comment 
`/* electron charge * -1.0 * charge state */` or something of the sort

PrometheusPi: OPTIONAL:
Here a comment on `( particle[boundElectrons_] - protonNumber )`being a negative number and thus `ELECTRON_CHARGE * ( particle[boundElectrons_] - protonNumber )` being a positive value would be helpful.

src/picongpu/include/traits/attribute/GetCharge.hpp
```diff
@@ -48,40 +48,40 @@ struct LoadBoundElectrons
...
  */
 template<>
 struct LoadBoundElectrons<false>
 {
     /** Functor implementation
      *
      * \tparam T_Particle particle type
-     * \param singlyChargedResult charge resulting from multiplying a single
+     * \param weighting the particle's weighting
      * electron charge (positive OR negative) by the macro particle weighting
```
n01r: I think this comment line is outdated.

PrometheusPi:
@ax3l Have you performed any run time tests?

src/picongpu/include/traits/attribute/GetCharge.hpp
```diff
@@ -100,8 +100,9 @@ HDINLINE float_X getCharge(const float_X weighting, const T_Particle& particle)
     typedef T_Particle ParticleType;
     typedef typename PMacc::traits::HasIdentifier<ParticleType, boundElectrons>::type hasBoundElectrons;
```
n01r: In `speciesDefinition.param` we might add another comment for the `chargeRatio`. It should probably state that charge calculation is not affected by this value if `boundElectrons` is set as a species attribute.

src/picongpu/include/traits/attribute/GetCharge.hpp
ax3l: I thought about that for a sec but guessed it's understandable. no strong opinion on that.

ax3l:
> @ax3l Have you performed any run time tests?

yes, we are currently launching one with @HighIander 's setup and he will report back

src/picongpu/include/traits/attribute/GetCharge.hpp
n01r: It is understandable but a comment wouldn't hurt either

ax3l: applied!

ax3l: see above

src/picongpu/include/traits/attribute/GetCharge.hpp
ax3l: true! fixed

src/picongpu/include/traits/attribute/GetCharge.hpp
ax3l: we add our comment, the handling of "atomic numbers XOR chargeRatio" and asserts in a follow-up PR

ax3l:
offline discussion with @psychocoderHPC:
- [ ] skip the constant record `charge` for ions with `boundElectrons` in HDF5/ADIOS

PrometheusPi:
The `hasTrait` fails for electrons - please fix

ax3l:
saw that already, but boost and the new resource plugin broke an other part so I did not get to push the fix yet. stay tuned.

PR #1845, opened 2017-02-15
Fix plot radiation matplotlib2.x

PrometheusPi:
This pull request fixes the matplotlib2.x incompatibility which occurs when `plt.imshow()` is used with `plt.xscale('log')`. See the following [matplotlib issue](https://github.com/matplotlib/matplotlib/issues/7661) for details. 

Furthermore 
 - the `plt.contour()` option is removed because there is no equivalent for logarithmic axis
 - The final `print`is now python3 compatible
 - the `file`type is replaced 
 - use `plt.tight_layout()`instead of `plt.subplot_adjust()`

The following pictures are the inital radiation of an LWFA - the spectral signatures should be at multiples of `omega_0`. 

**Prior to this fix:**
`matplotlib 1.3.1`
plot_radiation_lwfa_old

`matplotlib 2.0.0rc1`
plot_radiation_lwfa_new

**With the fix of this pull request:**
`matplotlib 1.3.1`
plot_radiation_lwfa_oldfixed

`matplotlib 2.0.0rc1`
plot_radiation_lwfa_newfixed

### DIFF
--- a/src/tools/bin/plotRadiation
+++ b/src/tools/bin/plotRadiation
@@ -391,13 +391,7 @@ for myfile in args.path2Data:
     CB.ax.yaxis.get_offset_text().set_fontsize(exponentsize)
 
     # set appearance of plot
-    plt.subplots_adjust(left=0.13,
-                        right=0.85,
-                        bottom=0.15,
-                        top=0.95,
-                        wspace=0.0,
-                        hspace=0.0)
-
+    plt.tight_layout()
     # get name of file
     filename = os.path.basename(myfile.name)
 
--- a/src/tools/bin/plotRadiation
+++ b/src/tools/bin/plotRadiation
@@ -70,7 +70,7 @@ parser = argparse.ArgumentParser(
             )
 
 # Path to input files (several are possible) - necesairy argument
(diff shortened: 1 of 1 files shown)


src/tools/bin/plotRadiation
```diff
@@ -70,7 +70,7 @@ parser = argparse.ArgumentParser(
             )
 
 # Path to input files (several are possible) - necesairy argument
```
ax3l: typo: *necessary* :)

src/tools/bin/plotRadiation
```diff
@@ -346,44 +343,33 @@ for myfile in args.path2Data:
...
+    shape = shape(dataZoomed)
     if(args.logOmega):
         SP.set_xscale('log')
+        x = logspace(log10(omega_min_draw), log10(omega_max_draw), shape[1])
+    else:
+        x = linspace(omega_min_draw, omega_max_draw, shape[1])
+
+    y = linspace(theta_min_draw, theta_max_draw, shape[0])
+
+    # plot data: (only call!!!)
```
ax3l: but: what does it *mean*? ;)

ax3l:
yes, I was not too concerned about them :) thanks for the follow-up :)

PR #1851, opened 2017-02-16
Examples: Prepare for Sphinx

ax3l:
Refactor the old documentation of examples to be sphinx compatible.
Requires to use restFUL files instead of plain markdown, nevertheless both are rendered in GitHub.

The old `.yaml` file was removed since we never finished the auto-testing inside PIConGPU itself. Nevertheless, we can later on add those files again for plain test setups (without example doc) as seen in [parataxis](https://github.com/ComputationalRadiationPhysics/parataxis/blob/ee10cd91cb54ad314bacdff4d01f5f5909ec6b6d/examples/DoubleSlit/documentation.yml).

@Heikman @PrometheusPi @n01r please take the same format and add an equally formatted `README.rst` to your new examples.

### Useful Links

- [A primer on writing restFUL files for sphinx](http://www.sphinx-doc.org/en/stable/rest.html)


examples/Bunch/README.rst
```diff
@@ -0,0 +1,35 @@
+Bunch: Thomson scattering from laser electron-bunch interaction
+===============================================================
+
+* author:      Richard Pausch <r.pausch (at) hzdr.de>, Rene Widera <r.widera (at) hzdr.de>
+* maintainer:  Richard Pausch <r.pausch (at) hzdr.de>
+
+This is a simulation of an electron bunch that collides head-on with a laser pulse.
+Depending on the number of electrons in the bunch, their momentum and their distribution and depending on the laser wavelength and intensity, the emitted radiation differs.
```
ax3l: yep, kept it from the original. but your suggestion does not make it shorter ;)

n01r: I know :wink: ... I counted on your self-aware freedom to put a full stop there somewhere and suggested an alternative way of modifying the long sentence (in case you wanted to keep it long) that makes you not read twice to get it.

PR #1852, opened 2017-02-16
Particle manipulators: Fix Position Offset

psychocoderHPC:
The first parameter for a particle manipulator is the offset to the origin of the local domain. IN the current dev it was not defined if it is the offset of the superCell or of a cell inside of a superCell. Some manipulators need the superCell offset (in cell) some a cell offset e.g. to create a global unique id.
This fix defined the offset as offset from the superCell to the origin of the local domain.

There will be a future full refactoring of the interfaces to support alpaka, therefore please ignore the new TODO within the code.

- rename `localCellIdx` to `localSuperCellOffset`
- add interface documentation
- Particles.kernel: give superCell offset to particle manipulators

tests:
- [x] KHI positrons/electrons (TSC shape) heating and charge conservation

### DIFF
--- a/src/picongpu/include/particles/manipulators/IfRelativeGlobalPositionImpl.hpp
+++ b/src/picongpu/include/particles/manipulators/IfRelativeGlobalPositionImpl.hpp
@@ -45,30 +45,54 @@ struct IfRelativeGlobalPositionImpl : private T_Functor
     }
 
     template<typename T_Particle1, typename T_Particle2>
-    DINLINE void operator()(const DataSpace<simDim>& localCellIdx,
+    DINLINE void operator()(const DataSpace<simDim>& localSuperCellOffset,
                             T_Particle1& particle1, T_Particle2& particle2,
                             const bool isParticle1, const bool isParticle2)
     {
         typedef typename SpeciesType::FrameType FrameType;
 
+        /* offset of the superCell (in cells, without any guards) to the origin of the global domain */
+        const DataSpace<simDim> globalSuperCellOffset = localSuperCellOffset + localDomainOffset;
+        bool particleInRange1 = isParticle1;
+        bool particleInRange2 = isParticle2;
 
-        DataSpace<simDim> myCellPosition = localCellIdx + localDomainOffset;
+        if( isParticle1 )
+        {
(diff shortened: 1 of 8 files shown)


ax3l:
can you add to the PR description what that bug implicated for implementations  like `SetDrift` or even `SetDrift` with a moving window?

src/picongpu/include/particles/manipulators/IfRelativeGlobalPositionImpl.hpp
```diff
@@ -45,31 +45,55 @@ struct IfRelativeGlobalPositionImpl : private T_Functor
...
+        /* offset of the superCell (in cells, without any guards) to the origin of the global domain */
+        const DataSpace<simDim> globalSuperCellOffset = localSuperCellOffset + localDomainOffset;
+        bool particleInRange1 = isParticle1;
+        bool particleInRange2 = isParticle2;
+
+        if( isParticle1 )
+        {
+            particleInRange1 = isParticleInsideRange( particle1, globalSuperCellOffset);
+        }
+        if( isParticle1 )
```
psychocoderHPC: BUG: This must be `isPartilce2` ([spotted in the backport pull request](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1906/files/9b569022b21b6daa234b8a6ebc93ec00a80394c6#r103636517))

psychocoderHPC: As I know this bug has no effect on any currently used example. The binary version of this functor is not used anywhere.

ax3l:
@psychocoderHPC good, well catched.

I have an other open question in this PR that you did not answer regarding the impact:

> can you add to the PR description what that bug implicated for implementations  like `SetDrift` or even `SetDrift` with a moving window?

psychocoderHPC:
@ax3l  sry forget to announce the change of the description: it is already in the description

ax3l:
> As I know SetDrift with the relative position filter is never used with moving window.

thanks! yes, not in our examples, but a user could generally do that for a setup of we don't know of, that's why I tried to understand the scope of affected manipulators :)

PR #1853, opened 2017-02-16
Fix typos plot radiation tool

PrometheusPi:
This pull request fixes several typos in the `plotRadition` tool and adjust the code alignment. 
These issues were discovered by @ax3l in #1845 but this pull request was merged anyway.


src/tools/bin/plotRadiation
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #
-# Copyright 2013-2017 Richard Pausch
```
ax3l: @PrometheusPi uh oh :)

src/tools/bin/plotRadiation
```diff
@@ -420,3 +477,5 @@ for myfile in args.path2Data:
     myfile.close()
 
     print(filename + "\t Done")
+
+#  LocalWords:  PIConGPU
```
ax3l: what is that?

src/tools/bin/plotRadiation
```diff
@@ -354,20 +405,26 @@ for myfile in args.path2Data:
...
-    # plot data: (only call!!!)
+    # plot data: (this is the only command that plots data)
     cax =  SP.pcolormesh(x, y, dataZoomed,
                          shading=my_interpolation,
                          cmap=my_cmap,
                          vmax=colorbarMax,
                          norm=colorNorm)
 
-    # set x and y range
+    # set x and y range (required after pcolormesh(?))
```
ax3l: sure, you can always set your limits independent of your plot area.
in your case, the default might do (show all).

PrometheusPi: pcolormesh seems to overwrite xlim in the case of logarithmic axis for matplotlib 1.3.x.
That's why I moved setting the axis limits after the pcolormesh call.

src/tools/bin/plotRadiation
PrometheusPi: This is for automated spell checks - ignore `PIConGPU` as typo.

src/tools/bin/plotRadiation
ax3l: they should be always behind the `plot()`, `imshow()` and `pcolormesh()` call afaik

src/tools/bin/plotRadiation
ax3l: in which editor? only emacs?

src/tools/bin/plotRadiation
PrometheusPi: Beforehand it worked even when `xlim`was set with the subplot. So I am not sure about that.

src/tools/bin/plotRadiation
PrometheusPi: actually not emacs - but anything that uses `aspell`

ax3l: interesting but a bit random to put it in one file
one could think about a per-project or per-repository file but even there it's a bit... out of scope.

is it ok for you not to add it?

PrometheusPi: Of course - I will remove it. (I can not use it with emacs anyway 😄 )

PR #1857, opened 2017-02-17
refactor gamma calculation

psychocoderHPC:
- add a function `gamma<>()` to calculate the gamma of a particle
- add a forward declaration for the class `Gamma<>`

This change is needed to add filter functors for radiation in a follow up.

### DIFF
--- a/src/picongpu/include/algorithms/Gamma.def
+++ b/src/picongpu/include/algorithms/Gamma.def
@@ -1,0 +1,84 @@
+/* Copyright 2013-2017 Axel Huebl, Heiko Burau, Rene Widera
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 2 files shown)


src/picongpu/include/algorithms/Gamma.hpp
```diff
@@ -19,29 +19,36 @@
...
+        T_MassType const mass
+    ) const
     {
-        const valueType fMom2 = math::abs2( precisionCast<valueType >( mom ) );
-        const valueType c2 = SPEED_OF_LIGHT*SPEED_OF_LIGHT;
+        using namespace PMacc;
 
-        const valueType m2_c2_reci = valueType(1.0) / precisionCast<valueType >( mass * mass * c2 );
+        valueType const fMom2 = math::abs2( precisionCast< valueType >( mom ) );
+        valueType const c2 = SPEED_OF_LIGHT * SPEED_OF_LIGHT;
```
ax3l: constexpr

src/picongpu/include/algorithms/Gamma.def
```diff
@@ -0,0 +1,84 @@
...
+
+#include "simulation_defines.hpp"
+
+
+namespace picongpu
+{
+
+    /** calculate the gamma of a particle
+     *
+     * @tparam T_PrecisionType type in which the calculation is performed
```
ax3l: type -> precision

src/picongpu/include/algorithms/Gamma.def
```diff
@@ -0,0 +1,84 @@
...
+        operator()(
+            T_MomType const & mom,
+            T_MassType const mass
+        ) const;
+
+    };
+
+    /** calculate the gamma of a particle
+     *
+     * @tparam T_PrecisionType type in which the calculation is performed
```
ax3l: type -> precision

psychocoderHPC:
changed

PR #1860, opened 2017-02-17
Gold: Z_eff, Energies, Isotope

ax3l:
Adds the most common isotope for gold, includings its effective nuclear charge (calculated from the [noted reference](https://dx.doi.org/10.1063/1.1712084) *Table IV*) and the NIST ionization energies.


src/picongpu/include/simulation_defines/param/ionizerConfig.param
```diff
@@ -270,6 +277,104 @@ namespace effectiveNuclearCharge
...
+        56.703,
+        56.703,
+        56.703,
+        56.703,
+        56.703,
+        /* 3s */
+        55.763,
+        55.763,
+        /* 2p */
+        74.513,
```
n01r: I really wonder what the implications are that the effective Z is not strictly ascending. Right now my own personal experimental BSI model is the only thing that uses them but maybe one should treat the whole thing differently. Haven't looked into that.

ax3l: yep, as discussed offline today: me too.

PR #1868, opened 2017-02-17
ReadTheDocs: Fix the Build

ax3l:
This fixes the build on readthedocs:

### Fixes

- clean Up RTD Setup
  - doxygen is provided
  - include for markdown
  - github extensions is not available on rtd
- doxygen: skip parsing three static asserts that are fasely interpreted as declarations
- sphinx: the full doxygen index currently takes ~2GB ram to build which is above the 1 GB limit on RTD, so we skip it for now

### Misc

- update badges
- finish param file structure
- after merging, the build will be triggered and you can find the first doc (stubs) under http://picongpu.readthedocs.io

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,4 @@
+requirements_file: docs/requirements.txt
+
+formats:
+    - htmlzip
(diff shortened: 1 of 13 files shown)


n01r:
One can apply for more RAM, I read. Will it be necessary or is it already foreseeable that we could require less RAM during the build process?

ax3l:
Yes, but this becomes relevant when we want to build the full index which is not highly important right now.

There are several ways how to get it build if we need it:
- ask for it
- pay for it (should be possible)
- build locally via cron jobs and push to a gh repo / `gh-pages`

PR #1879, opened 2017-02-21
update documentation `struct.hpp`

psychocoderHPC:
- fix `static const` replacements from older commit
- add example to `PMACC_STRUCT`
- update comments within the examples

### DIFF
--- a/src/libPMacc/include/preprocessor/struct.hpp
+++ b/src/libPMacc/include/preprocessor/struct.hpp
@@ -19,6 +19,16 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
+/** @file
+ *
+ * collection of `TypeMemberPair`s
+ *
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/preprocessor/struct.hpp
```diff
@@ -87,7 +89,7 @@
  *
  *   @code{.cpp}
  *     PMACC_C_VALUE(float_64, power_SI, 2.0);
- *     // is the compile time equivalent of
+ *     // is syntactically equivalent to
  *     static constexpr float_64 power_SI = float_64(2.0);
```
ax3l: `constexpr` or `const`? :)

psychocoderHPC: constexpr is correct for this macro

PR #1886, opened 2017-02-22
DataConnector: Add Clean() Method

ax3l:
Add a `clean()` method to `DataConnector` to trigger unsharing (and potentially removing/destroying) all still shared `ISimulationData` objects.

@n01r also fixes the segfault on simulation shutdown if a RNG is used. RNG forgets to `unshare` itself from the DataConnector.

### DIFF
--- a/src/libPMacc/include/dataManagement/DataConnector.hpp
+++ b/src/libPMacc/include/dataManagement/DataConnector.hpp
@@ -121,6 +121,10 @@ namespace PMacc
             datasets.push_back( data );
         }
 
+        /** End sharing a dataset with identifier id
+         *
+         * @param id id of the dataset to remove
+         */
         void
         unshare( SimulationDataId id )
         {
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/dataManagement/DataConnector.hpp
```diff
@@ -216,13 +238,7 @@ namespace PMacc
         {
             log< ggLog::MEMORY >( "DataConnector: being destroyed (%1% datasets left to destroy)" ) %
                                 datasets.size();
-            //datasets.clear();
-            for( auto it = datasets.begin(); it != datasets.end(); ++it )
-            {
-                log< ggLog::MEMORY >( "DataConnector: destroying %1%" ) %
-                                    (*it)->getUniqueId();
-                datasets.erase( it );
```
ax3l: that construct destroyed the data structure it looped over.
segfault if some objects forgot to `unshare` themselves

PR #1887, opened 2017-02-22
Freedom for Fields (& Species)

ax3l:
Remove fields and species from `MySimulation` member lists.

> *"We [open data activists] do not want to emancipate the [data], we want the [data] to emancipate themselves.”*
> -- Errico Malatesta

This simplifies the `MySimulation` interface to allow easier overloading and re-implementing of parts of it (like runOneStep) without the need to manage too many members. This becomes especially helpful if one tries to reimplement the `MySimulation` class in Python to *hack the PIC cycle*.

> *"[DataConnector] could never get a man to the moon, but it may the only mode that can allow us to survive on earth."*
> -- Sheldon B. Kopp

Global singleton classes like E, B, J, FieldTmp slots, species are only created via functors once and then shared with `DataConnector`. All algorithms accessing those can then query (`dc.get( ... )`) for reads & manipulation.

> *"The [authors of DataConnector] never have claimed that liberty [of data] will bring perfection; they simply say that its results are vastly preferable to those that follow [a one-class has to own all gobal data structures] authority."*
> -- Benjamin Tucker

- [x] rebase: depends on #1886

### DIFF
--- a/src/picongpu/include/simulationControl/MySimulation.hpp
+++ b/src/picongpu/include/simulationControl/MySimulation.hpp
@@ -76,6 +76,7 @@
 #include <boost/mpl/int.hpp>
 #include <memory>
 
+
 namespace picongpu
 {
 using namespace PMacc;
@@ -97,9 +98,6 @@ public:
      */
     MySimulation() :
     laser(nullptr),
-    fieldB(nullptr),
-    fieldE(nullptr),
-    fieldJ(nullptr),
     mallocMCBuffer(nullptr),
     myFieldSolver(nullptr),
     myCurrentInterpolation(nullptr),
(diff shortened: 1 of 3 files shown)


src/picongpu/include/simulationControl/MySimulation.hpp
```diff
@@ -281,12 +273,24 @@ class MySimulation : public SimulationHelper<simDim>
...
     {
         namespace nvmem = PMacc::nvidia::memory;
+
+        DataConnector &dc = Environment<>::get().DataConnector();
+
         // create simulation data such as fields and particles
-        fieldB = new FieldB(*cellDescription);
-        fieldE = new FieldE(*cellDescription);
-        fieldJ = new FieldJ(*cellDescription);
+        auto fieldB = new FieldB( *cellDescription );
```
psychocoderHPC: only a note: `std::shared_ptr<FieldB> fieldB(new FieldB( *cellDescription ));` prevent that someone can use the plain pointer later in this method.

ax3l: could do, but `share()` requires a `std::shared_ptr< ISimulationData >` not a `std::shared_ptr<FieldB>`.

`ISimulationData` as no `init()` interface which still need sto be called and I would still need to cast via `std::static_pointer_cast< ISimulationData>` before passing to `share()`.

PR #1890, opened 2017-02-22
PMacc environment refactoring

psychocoderHPC:
This change allows to execute `picongpu --help` also if MPI is not available (e.g. on the hypnos head nodes).

- create `detail::Environment` for all singletons without dimension
- add `detial::EnvironmentContext` to hold context information variables
  for `Environment<>` independent of number of simulation dimensions
- add documentation
- change friend from `Environment<>` to `detail::Environment` (for all singletons without dependency to the simulation dimension)
- use clean environment finalize of PMacc
- remove gloable usage of `PMacc` and `picongpu` namespace
- update `gameOfLife`
- update PMacc test cases

- [x] rebase against #1888
- [x] runtime test gameOfLife
- [x] runtime test PMacc test cases

### DIFF
--- a/src/libPMacc/include/Environment.def
+++ b/src/libPMacc/include/Environment.def
@@ -21,9 +21,16 @@
 
 #pragma once
 
+#include "pmacc_types.hpp"
+
 namespace PMacc {
 
-    template <unsigned DIM = DIM1>
+    template< uint32_t T_dim = DIM1 >
     class Environment;
 
-}  // namespace PMacc
+namespace detail
+{
+    struct Environment;
+
+} // namespace detail
+} // namespace PMacc
--- a/src/libPMacc/include/Environment.hpp
+++ b/src/libPMacc/include/Environment.hpp
@@ -36,129 +36,369 @@
 #include "mappings/simulation/Filesystem.hpp"
 #include "eventSystem/events/EventPool.hpp"
 #include "Environment.def"
+#include "communication/manager_common.hpp"
+#include "assert.hpp"
 
 #include <cuda_runtime.h>
+#include <mpi.h>
 
 namespace PMacc
 {
 
-/**
- * Global Environment singleton for Picongpu
- */
+namespace detail
+{
+    /** collect state variables of the environment context
+     *
+     * This class handle the initialization and finalize of the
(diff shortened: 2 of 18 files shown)


src/picongpu/main.cu
```diff
@@ -65,8 +61,8 @@ int main(int argc, char **argv)
             break;
```
ax3l: offline discussion: line 58: the comment looks wrong (should be 0?) and outdated. Maybe we just add a note that the missing break is voluntarily to set the error code to 0 if everything went well.

src/libPMacc/include/Environment.def
```diff
@@ -21,9 +21,102 @@
...
+namespace detail
+{
+    struct Environment;
+
+    /** collect state variables of the environment context
+     *
+     * This class handle the initialization and finalize of the
+     * MPI context and the selection of the GPU.
+     */
+    class EnvironmentContext
```
ax3l: is it really necessary to put this in the .def file?

src/libPMacc/include/Environment.hpp
```diff
@@ -36,141 +36,326 @@
...
+    }
+
+    void EnvironmentContext::finalize()
+    {
+        if( m_isMpiInitilized )
+        {
+            PMacc::Environment<>::get().Manager().waitForAllTasks();
+            // Required by scorep for flushing the buffers
+            cudaDeviceSynchronize();
+            m_isMpiInitilized = false;
```
ax3l: typo: `m_isMpiInitialized` :)

src/picongpu/main.cu
ax3l: offline discussion: line 58: the comment looks wrong (should be 0?) and outdated. Maybe we just add a note that the missing break is voluntarily to set the error code to 0 if everything went well.

psychocoderHPC: changed

psychocoderHPC:
rebased

psychocoderHPC:
I updated the missing Bremsstrahlungs example, game of life and the PMacc tests.

psychocoderHPC:
can be merged, all tests passed

PR #1896, opened 2017-02-23
Doc: Slurm

ax3l:
Add and update our Slurm example for Sphinx docs.


src/picongpu/submit/taurus-tud/Slurm_Tutorial.rst
```diff
@@ -0,0 +1,39 @@
+Slurm examples
+==============
+
+Job Submission
+""""""""""""""
+
+PIConGPU job submission on the *Taurus* cluster at *TU Dresden*:
+
+* `tbg -s sbatch -c submit/0008gpus.cfg -t submit/taurus-tud/k80_profile.tpl $SCRATCH/runs/test123`
```
PrometheusPi: QUESTION:
Do you think explaining what `$SCRATCH` is necessary?
Would `[your simulation path]/runs/test123` be better?

src/picongpu/submit/taurus-tud/Slurm_Tutorial.rst
```diff
@@ -0,0 +1,39 @@
...
+
+
+Job Control
+"""""""""""
+
+* interactive job:
+
+  * `salloc --time=1:00:00 --nodes=1 --ntasks-per-node=2 --cpus-per-task=8 --partition gpu-interactive`
+  * e.g. `srun "hostname"`
+  * e.g. some clusters need additional flags, e.g. taurus `--gres=gpu:2`
```
PrometheusPi: QUESTION:
This is for taurus only. Would it be better to write:
```
* for GPU allocations taurus requires an additional flag: `--gres=gpu:2` 
```

src/picongpu/submit/taurus-tud/Slurm_Tutorial.rst
ax3l: Imo, since it's a specific example to see how to work on a HPC system on the example of Taurus, using the fast and large storage such as `$SCRATCH` is not bad.

src/picongpu/submit/taurus-tud/Slurm_Tutorial.rst
ax3l: sure!

PR #1897, opened 2017-02-23
check if binary `cuda_memtest` exists

psychocoderHPC:
- skip GPU memory test if the binary `cuda_memtest` does not exists
- add `cuda_memtest` to  `bash` and `davinci` submit templates

We need this optional memory test for the case that it is not possible to build cuda_memtest ~~(e.g. on titan ORNL)~~ and if PIConGPU is compiled with clang as device compiler #1731.

[update] On titan ORN it is not possible to call a bash script with aprun, therefore it is not possible to activate `cuda_memtest`

### DIFF
--- a/src/picongpu/submit/bash/bash_mpiexec.tpl
+++ b/src/picongpu/submit/bash/bash_mpiexec.tpl
@@ -43,7 +43,12 @@ umask 0027
 mkdir simOutput 2> /dev/null
 cd simOutput
 
-mpiexec --prefix $MPI_ROOT -tag-output --display-map -x LIBRARY_PATH -x LD_LIBRARY_PATH -am !TBG_dstPath/tbg/openib.conf --mca mpi_leave_pinned 0 -npernode !TBG_gpusPerNode -n !TBG_tasks !TBG_dstPath/picongpu/bin/cuda_memtest.sh
+# test if cuda_memtest binary is available
+if [ -f !TBG_dstPath/picongpu/bin/cuda_memtest ] ; then
+  mpiexec --prefix $MPI_ROOT -tag-output --display-map -x LIBRARY_PATH -x LD_LIBRARY_PATH -am !TBG_dstPath/tbg/openib.conf --mca mpi_leave_pinned 0 -npernode !TBG_gpusPerNode -n !TBG_tasks !TBG_dstPath/picongpu/bin/cuda_memtest.sh
+else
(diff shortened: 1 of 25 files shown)


src/picongpu/submit/bash/bash_mpiexec.tpl
```diff
@@ -43,9 +43,17 @@ umask 0027
...
+haveCudaMemtest=$?
+
+if [ $haveCudaMemtest -eq 0 ] ; then
+  mpiexec --prefix $MPI_ROOT -tag-output --display-map -x LIBRARY_PATH -x LD_LIBRARY_PATH -am !TBG_dstPath/tbg/openib.conf --mca mpi_leave_pinned 0 -npernode !TBG_gpusPerNode -n !TBG_tasks !TBG_dstPath/picongpu/bin/cuda_memtest.sh
+else
+  echo "no binary 'cuda_memtest' available, skip GPU memory test" >&2
+fi
 
-if [ $? -eq 0 ] ; then
+if [ $? -eq 0 ] || [ $haveCudaMemtest -eq 0 ] ; then
```
ax3l: actuallly, `$?` will be 0 anyway, because the commands before it all succeeded when cuda_memtest was not found

src/picongpu/submit/bash/bash_mpiexec.tpl
```diff
@@ -43,9 +43,17 @@ umask 0027
 mkdir simOutput 2> /dev/null
 cd simOutput
 
-mpiexec --prefix $MPI_ROOT -tag-output --display-map -x LIBRARY_PATH -x LD_LIBRARY_PATH -am !TBG_dstPath/tbg/openib.conf --mca mpi_leave_pinned 0 -npernode !TBG_gpusPerNode -n !TBG_tasks !TBG_dstPath/picongpu/bin/cuda_memtest.sh
+# test if cuda_memtest binary is available
+test -f !TBG_dstPath/picongpu/bin/cuda_memtest
+haveCudaMemtest=$?
+
+if [ $haveCudaMemtest -eq 0 ] ; then
```
ax3l: since the line below is not necessary, you could do `if [ -f  ...]` here directly

ax3l:
pls add to the description if you also need this for clang, otherwise this should only be added in the cray (titan) template for now

src/picongpu/submit/bash/bash_mpiexec.tpl
PrometheusPi: true :+1:

PR #1900, opened 2017-02-24
openPMD: Fix Grid Spacing, GlobalOffset

### LINKED ISSUE #1899: openPMD: gridSpacing & gridGlobalOffset
The order of `gridSpacing` and `gridGlobalOffset` is likely not the same as `axisLabels` and needs to be adjusted accordingly.

Reported in https://github.com/openPMD/openPMD-viewer/issues/146

This is only affecting your output if at least one of the following is true
- moving window is used
- non-quadratic cells are used

this problem is independent if your simulation box cell sizes (`-g`), quadratic or not is not a problem.

ax3l:
Fix #1899: our PIConGPU arrays describing cell sizes, offsets and sizes are in a rather inconsequent `{x, y, z}` order while the datasets they describe are *labeled* in `F[z][y][x]` order.

(note: this has nothing to do with slow or fast varying index aka C vs F order. This is just a inconsequent *labeling* of our `F[i][j][k]` indizes in PIConGPU, which arbitrarily did choose that `x` will always be the *fastest* varying "index"...)

This fixes the same issue as previously seen with `axisLabels` for openPMD in ADIOS and HDF5 output #1478.

Only affects if at least one of the following is true:
- moving window is used
- non-quadratic cells are used

the size / layout of the simulation box (`-g`) is independent of this issue and works quadratic and non-quadratic.

### DIFF
--- a/src/picongpu/include/plugins/adios/ADIOSWriter.hpp
+++ b/src/picongpu/include/plugins/adios/ADIOSWriter.hpp
@@ -355,9 +355,10 @@ private:
                 adios_string_array, simDim, axisLabels ));
         }
 
+        // cellSize is {x, y, z} but fields are F[z][y][x]
         std::vector<float_X> gridSpacing(simDim, 0.0);
         for( uint32_t d = 0; d < simDim; ++d )
-            gridSpacing.at(d) = cellSize[d];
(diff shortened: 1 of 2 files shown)


PrometheusPi:
Run time test displays the correct ranges now.
openpmd_viewer_correctslides

PR #1910, opened 2017-03-01
fix manipulator `IfRelativeGlobalPositionImpl`

psychocoderHPC:
the position for the second particle is only checked if the first particle is valid

This bug is only triggered if `IfRelativeGlobalPositionImpl` is used as binary operator. The example `KelvinHelmholtz` is only using the unary version of this manipulator.

### DIFF
--- a/src/picongpu/include/particles/manipulators/IfRelativeGlobalPositionImpl.hpp
+++ b/src/picongpu/include/particles/manipulators/IfRelativeGlobalPositionImpl.hpp
@@ -60,7 +60,7 @@ struct IfRelativeGlobalPositionImpl : private T_Functor
         {
             particleInRange1 = isParticleInsideRange( particle1, globalSuperCellOffset);
         }
-        if( isParticle1 )
+        if( isParticle2 )
         {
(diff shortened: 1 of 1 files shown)


src/picongpu/include/particles/manipulators/IfRelativeGlobalPositionImpl.hpp
```diff
@@ -60,7 +60,7 @@ struct IfRelativeGlobalPositionImpl : private T_Functor
         {
             particleInRange1 = isParticleInsideRange( particle1, globalSuperCellOffset);
         }
-        if( isParticle1 )
+        if( isParticle2 )
         {
             particleInRange2 = isParticleInsideRange( particle2, globalSuperCellOffset);
```
PrometheusPi: QUESTION:
if `particle1` and `particle2` are at the same position, both `particleInRange1` and `particleInRange2`have the same boolean value. Is this intended?

PrometheusPi: *offline discussion with @psychocoderHPC:*
Yes, this is the intended behavior.

PR #1911, opened 2017-03-01
clang compatible static assert usage

psychocoderHPC:
An expression within a static assert those not depend on a template can be evaluated from the compiler even if the code is never instantiated.
This means that a context check inside the body of a class is not possible with some compiler e.g. clang.
A workaround is to extent the condition for the static assert with a template dependent expression those is always evaluated to `true`.
The expression `sizeof(AnyType)!=0` is always `true` and can be used as workaround.

This pull request is needed by the pull request to compile with clang #1731.

Tested with
- [x] clang
- [x] gcc/nvcc

### DIFF
--- a/src/libPMacc/include/compileTime/errorHandlerPolicies/ThrowValueNotFound.hpp
+++ b/src/libPMacc/include/compileTime/errorHandlerPolicies/ThrowValueNotFound.hpp
@@ -37,7 +37,15 @@ struct ThrowValueNotFound
     template<typename T_MPLSeq, typename T_Value>
     struct apply
     {
-        PMACC_CASSERT_MSG_TYPE(value_not_found_in_seq, T_Value, false);
+        /* The compiler is allowed to evaluate an expression that does not depend on a template parameter
+         * even if the class is never instantiated. In that case static assert is always
+         * evaluated (e.g. with clang), this results in an error if the condition is false.
+         * http://www.boost.org/doc/libs/1_60_0/doc/html/boost_staticassert.html
+         *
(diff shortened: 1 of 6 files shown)


src/libPMacc/include/compileTime/errorHandlerPolicies/ThrowValueNotFound.hpp
```diff
@@ -37,7 +37,15 @@ struct ThrowValueNotFound
     template<typename T_MPLSeq, typename T_Value>
     struct apply
     {
-        PMACC_CASSERT_MSG_TYPE(value_not_found_in_seq, T_Value, false);
+        /* The compiler is allowed to evaluate an expression those not depends on a template parameter
```
ax3l: that does not depend on a

src/libPMacc/include/compileTime/errorHandlerPolicies/ThrowValueNotFound.hpp
```diff
@@ -37,7 +37,15 @@ struct ThrowValueNotFound
...
+         * even if the class is never instantiated. In that case static assert is always
+         * evaluated (e.g. with clang), this results in an error if the condition is false.
+         * http://www.boost.org/doc/libs/1_60_0/doc/html/boost_staticassert.html
+         *
+         * A workaround is to add a template dependency to the expression.
+         * `sizeof(ANY_TYPE) != 0` is always true and defer the evaluation.
```
ax3l: defer*s*

psychocoderHPC:
updated

PR #1580, opened 2016-09-21
Add transverse Laguerre-modes to standard Gauss laser.

BeyondEspresso:
Hello y'all,

This pull request implements basic support of rotationally symmetric Gauss-Laguerre laser modes within the standard Gauss laser.

Literature on this can be found at numerous places, but you can start here:
[F. Pampaloni et al. (2004), Gaussian, Hermite-Gaussian, and Laguerre-Gaussian beams: A primer](https://arxiv.org/pdf/physics/0410021)
[Wikipedia on Gaussian laser beams](https://en.wikipedia.org/wiki/Gaussian_beam)

Although this implementation has passed basic testing, it is still work in progress. Especially, the implementation of evaluating Laguerre-Polynomials is very simple, but apparently good enough in performance for this purpose. Nevertheless a better library implementation for GPUs would be welcome. So far, I did not stumble across anything related to CUDA. On the host side we can definitely use the appropriate boost-library.

Anyway, please don't hold your fire, but start the bombardment immediately! :bomb: :scream: What needs be done? :smiley: 

Cheers! :cyclone: :zap: :coffee:

PR #1912, opened 2017-03-01
remove useless `PMACC_VERIFY`

psychocoderHPC:
It is not allowed to use `PMACC_VERIFY` in a device or host/device function ( because std::string is not available on device).
  - remove `PMACC_VERIFY_MSG` from `laserTransversal`

Never the less the check is useless because it is already checked at compile time.

I added the compile time range check also to the method `laserLongitudinal` to be sure that
the value is valid before it is used inside the followed loop.

This bug was introduced with #1580 (currently not in the master branch). It is a non critical bug for `nvcc` but avoid to compile with `clang`.

### DIFF
--- a/src/picongpu/include/fields/laserProfiles/laserGaussianBeam.hpp
+++ b/src/picongpu/include/fields/laserProfiles/laserGaussianBeam.hpp
@@ -44,6 +44,8 @@ namespace picongpu
 
             // This check is done here on HOST, since std::numeric_limits<float_X>::epsilon() does not compile on laserTransversal(), which is on DEVICE.
             float_X etrans_norm = float_X(0.0);
+
+            PMACC_CASSERT_MSG(MODENUMBER_must_be_smaller_than_number_of_entries_in_LAGUERREMODES_vector, MODENUMBER<LAGUERREMODES_t::dim);
             for ( uint32_t m = 0 ; m <= MODENUMBER ; ++m )
                 etrans_norm += LAGUERREMODES[m];
             PMACC_VERIFY_MSG( algorithms::math::abs(etrans_norm) > std::numeric_limits<float_X>::epsilon(), "Sum of LAGUERREMODES can not be 0." );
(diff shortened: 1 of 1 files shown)


src/picongpu/include/fields/laserProfiles/laserGaussianBeam.hpp
```diff
@@ -134,7 +136,6 @@ namespace picongpu
             float_X etrans = float_X(0.0);
             float_X etrans_norm = float_X(0.0);
             PMACC_CASSERT_MSG(MODENUMBER_must_be_smaller_than_number_of_entries_in_LAGUERREMODES_vector, MODENUMBER<LAGUERREMODES_t::dim);
```
PrometheusPi: This is the same code you added above

src/picongpu/include/fields/laserProfiles/laserGaussianBeam.hpp
```diff
@@ -44,6 +44,8 @@ namespace picongpu
 
             // This check is done here on HOST, since std::numeric_limits<float_X>::epsilon() does not compile on laserTransversal(), which is on DEVICE.
             float_X etrans_norm = float_X(0.0);
+            
+            PMACC_CASSERT_MSG(MODENUMBER_must_be_smaller_than_number_of_entries_in_LAGUERREMODES_vector, MODENUMBER<LAGUERREMODES_t::dim);
```
PrometheusPi: See comment below (in code) this is (still) a code repetition.

src/picongpu/include/fields/laserProfiles/laserGaussianBeam.hpp
psychocoderHPC: Yes, I also added at in the other method that everything is also valid if someone refactor only one of this methods.

PrometheusPi: *offline discussion with @psychocoderHPC:*
One might call just one of these methods thus checking it twice is okay.

Issue #1913, opened 2017-03-02
reflections from boundary after slide 


PrometheusPi:
~~There is an error when the `slide_point = 0.0` and restarts are used because the restart calculates the slide steps wrongly.~~

@psychocoderHPC and me are investigating it currently. More details will follow. 

UPDATE:
The error is do a sudden activation of the absorbing (and thus reflecting) boundary conditions after a slide. The variable `slide_point` **does not define a slide_point but the move point** and thus does not cause this effect. To avoid further confusion:
 - the name conflict should be fixed #1915 
 - the issue with boundary conditions after slides #1916 should be solved

Thus this issue will be closed.

PrometheusPi:
The detailed study of the hdf5 data (with @psychocoderHPC) showed that we do not slide on the first time step - but this is intended.
So the only misconception (for me) was that `slide_point` is actually a `moving_point` in PIConGPU and libPMacc speak. 
 - move - translate the co-moving window
 - slide - empty last GPU(s) in y and place them to the front (y) of the simulation

PrometheusPi:
I will still check if the h5 output is correct and if this really solves the restart issue I encountered.

ax3l:
Thanks for the clarification, that all seems normal.

Can you please specify what the *restart issue* is specifically that you encountered? Does it crash or does it screw the offset or did you just misinterpret it in post-processing (checkpoint or output data)?

PrometheusPi:
In the hdf5 checkpoint and data files the fields looked like wrongly copied:
![picongpu_bunch_mismatch](https://cloud.githubusercontent.com/assets/5121158/23554213/6ca6b69c-0024-11e7-9a19-07f173923a68.png)

However this is an effect that comes **with every slide** (relocation of GPUs) because our absorber is defined on `y_entire=[0:32]` on the **entire** simulation volume not in the **moving volume**. Thus we have non-zero fields in the moving box   `y_moving = [0:32]` which regularly are reduced and thus "reflected" by an (unphysical) absorber if we slide/relocate GPUs. Thus, this is **not an effect of slide_point==0** but of **slides in general**. However, since I have seen this so far only with setups using `slide_point = 0` and because we "missed sliding after the first iteration" I suspected the origin of these field jumps to be the `slide_point = 0`.

ax3l:
[...]

PR #1914, opened 2017-03-03
add python module for particle initialization

PrometheusPi:
This pull request adds a **python module** that allows to set up an arbitrary particle distribution to simulate e.g. an electron bunch. It also comes with a method to copy particle + field data after an initialization run with a free pusher to a new setup with e.g. a Boris pusher to take into account bunch evolution due to external fields. 

Still **to do**:

 - [x] add jupyter notebook as an example of a full initialization
 - [x] load data from h5 dump instead of checkpoint (see discussion with @psychocoderHPC in #1913)
 - [ ] add as modules inside `lib/python/picongpu/init/`
 - [ ] missing license header in new files
 - [ ] make sure all new files pass `pep8 fileName.py`

### DIFF
--- a/src/tools/share/python/initialzeBunch.ipynb
+++ b/src/tools/share/python/initialzeBunch.ipynb
@@ -1,0 +1,354 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import matplotlib.pyplot as plt\n",
+    "from scipy import constants\n",
+    "%matplotlib inline\n",
+    "\n",
+    "import sys"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "sys.path.append(\"./modules4picongpu/\")\n",
+    "\n",
+    "from bunchInit import vec3D\n",
+    "from bunchInit import addParticles2Checkpoint"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Set up the parameters for the electron bunch"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "# define values\n",
+    "\n",
+    "# grof values\n",
+    "delta_t = 1.6674254e-17\n",
+    "delta_x = 8.120995783129748e-8 * 4.0\n",
+    "delta_y = 5.0e-9\n",
+    "\n",
+    "# number of GPUs\n",
+    "GPUs = (512, 3840, 512)\n",
+    "\n",
+    "# iterations till PIConGPU has inialized the fields\n",
+    "N_init_timesteps = 200000\n",
+    "\n",
+    "# characteristic parameters of the electron bunch\n",
+    "mean_kinetic_energy = 15.0e6 # [eV]\n",
+    "norm_trans_emmitt_notSI = 0.5 # [mm mrad]\n",
+    "rel_energy_spread = 0.008\n",
+    "interaction_distance = 5.66e-3 + N_init_timesteps * delta_t * constants.speed_of_light # m\n",
+    "bunch_cross_section_radius = 7.0e-6 # m (rms)\n",
+    "bunch_duration = 17.5e-15 # s (rms)\n",
+    "bunch_charge = 35.0e-12 # C\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "# compute position in simulation box\n",
+    "center_pos_x = delta_x * GPUs[0]/2.\n",
+    "center_pos_z = delta_x * GPUs[2]/2.\n",
+    "y_center = 5.5e-6 + interaction_distance\n",
+    "\n",
+    "# number of macro particles in simulation\n",
+    "N_particles = np.int(bunch_charge / (100.*constants.elementary_charge) )\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "# define helper functions\n",
+    "\n",
+    "def gammaFromEkin_eV(Ekin_eV):\n",
+    "    Ekin_J = Ekin_eV * constants.elementary_charge\n",
+    "    return Ekin_J/(constants.electron_mass * constants.speed_of_light**2) +1\n",
+    "\n",
+    "def betaFromGamma(gamma):\n",
+    "    return np.sqrt(1.0 - 1.0/gamma**2)\n",
+    "\n",
+    "def momentumFromGammaBeta(gamma, beta):\n",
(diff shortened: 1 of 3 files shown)


ax3l:
pretty cool, shall we maybe move it to the scope of [`deppp`](https://github.com/ComputationalRadiationPhysics/deppp)?

PrometheusPi:
That would be a good idea - I think `picongpu-tools` would be the right place because it is not just a density profile tool but allows defining the entire phase space. 

We should move it (with all other tools and modules) to such a repo if it exists.

src/tools/share/python/initialzeBunch.ipynb
```diff
@@ -0,0 +1,354 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "# define values\n",
+    "\n",
+    "# grof values\n",
```
PrometheusPi: typo: grid

src/tools/share/python/initialzeBunch.ipynb
```diff
@@ -0,0 +1,354 @@
...
+    "delta_t = 1.6674254e-17\n",
+    "delta_x = 8.120995783129748e-8 * 4.0\n",
+    "delta_y = 5.0e-9\n",
+    "\n",
+    "# number of GPUs\n",
+    "GPUs = (512, 3840, 512)\n",
```
PrometheusPi: wrong: number of cells

src/tools/share/python/initialzeBunch.ipynb
```diff
@@ -0,0 +1,354 @@
...
+    "\n",
+    "# iterations till PIConGPU has inialized the fields\n",
+    "N_init_timesteps = 200000\n",
+    "\n",
+    "# characteristic parameters of the electron bunch\n",
+    "mean_kinetic_energy = 15.0e6 # [eV]\n",
+    "norm_trans_emmitt_notSI = 0.5 # [mm mrad]\n",
+    "rel_energy_spread = 0.008\n",
+    "interaction_distance = 5.66e-3 + N_init_timesteps * delta_t * constants.speed_of_light # m\n",
```
PrometheusPi: better focus position

ax3l:
while we are on it, let us rename the dir `modules4picongpu/` to `modules/` in one swipe (even though it existed already)

ax3l:
> That would be a good idea - I think picongpu-tools would be the right place because it is not just a density profile tool but allows defining the entire phase space.

as discussed offline: it always starts in our case with a density profile, generates particles in 3D and then the 3V part is just modulated on top. We are doing a very similar thing for the scattering profiles (Malte, Melanie) so one could put it on a common ground with `deppp` which already has all the meta data in place before modulating a `T_e` or general `3V` on it.

ax3l:
does it make sense to quash this PR in one or two (module, notebook) commits?

PrometheusPi:
Yes, that would make sense - I will squash it into two commits after some (typo) fixes.

ax3l:
admittedly, I am not the biggest fan of adding full-flavored notebooks to this repo since they are rather non-source/"compiled" and hard to review/maintain along the actual code. modules, md/reStrcuturedText code examples or a link to a gist with a ipynb with a usage example might be better...

PrometheusPi:
@ax3l I agree - is there any better option for adding `*.ipynp` **and** having a version control on it (which gist does not support as far as I know)

ax3l:
yep as I wrote: separate repo *or* a gist, both are versioned :)

PrometheusPi:
@ax3l How would you generate a markdown text (with code examples) from this `*.ipynb` (as mentioned in your previous post)?

ax3l:
If you installed [pandoc](http://pandoc.org/), you can export it directly from jupyter. This requires some non-python modules and might be hard to do on the cluster.

In order to do the same from your CLI locally where installing pandoc is no problem, just save your `*.ipynb` locally, install pandoc locally and use [`jupyter nbconvert`](https://github.com/jupyter/nbconvert). You can try if `.rst` or `.md` works best for you, both are fine for example check-in.

ax3l:
I added you outstanding todo's in your PR description, so I don't have to look through my comments while you are still working on this PR

ax3l:
I still think that should go in a `picongpu-utils` or `picongpu-tools` repo instead of here since it is basically openPMD file interaction, but we can talk about this when the other todo's are finished next week

ax3l:
as in #1979, I think a good place to put those scripts is as a python module for an example.
The best example might be in `examples/Bunch/pymodules/`. Don't forget to add a `__init__.py` in this (new) dir and we might want to add it to `submitAction.sh` to be copied when found.

PrometheusPi:
@ax3l What do you mean with using `.rst` files instead of `.ipynb`? Should I save the `.ipynb` as `.rst` and commit that file?

ax3l:
no, instead of `.ipynb` files please only add `.py` modules. you are free to show *very short* code snippets in `README.rst` files, but you can do the same just in the module in a `__main__` section of in classes' doc strings.

src/tools/share/python/doc_bunchInit.md
```diff
@@ -0,0 +1,219 @@
+```python
```
ax3l: this file (as `.rst` file) can go in the `docs/source/usage/workflows/` section.

src/tools/share/python/initialzeBunch.ipynb
```diff
@@ -0,0 +1,354 @@
+{
```
ax3l: this file should not be added

src/tools/share/python/modules4picongpu/__init__.py
```diff
@@ -18,3 +18,5 @@
 #
 
 from .hdf5Radiation import radiationHDF5  # NOQA
```
ax3l: this file does not exist anymore and is not needed anymore

src/tools/share/python/modules4picongpu/bunchInit.py
```diff
@@ -0,0 +1,430 @@
+import numpy as np
```
ax3l: this module should go into `src/tools/lib/python/`

ax3l:
@PrometheusPi what is the status of this PR? Can you please add it to the modules as commented?

ax3l:
@PrometheusPi can you please update the PR as indicated? :)

PrometheusPi:
@ax3l Hopefully I will find time this week to finish this.

BeyondEspresso:
I tried out several versions of this code (serial as submitted here and parallel). I extended it with MPI communication and parallel HDF5-calls very similar to [PR: Added MPI collective I/O support](https://github.com/h5py/h5py/pull/648/files). All reading is realively fast on 16 Sandy cores on Taurus and finished 30GB within minutes. However, writing happens with a bandwidth of staggering 50kB/s per core according to `pidstat -dl 20` 😭 . So it appears at least, that there is still work left on this issue.

My question to everybody in this thread: Did anyone realize *parallel* writing with a h5py interface *and* a good bandwidth?

ax3l:
@BeyondEspresso thanks for the observation, that's valuable feedback!

50kB/s looks less like a problem that needs to be solved immediately via parallel I/O but more like a problem in the read-write patterns used in the script.

Can you quickly profile the script and check where the actual hotspots sit? Use the `line_profiler` described in this blog post: https://www.huyng.com/posts/python-performance-analysis (an alternative is [cProfiler](https://julien.danjou.info/blog/2015/guide-to-python-profiling-cprofile-concrete-case-carbonara))

ax3l:
@PrometheusPi can you please move the module in `lib/python/picongpu`?

PrometheusPi:
@ax3l There are further updates to this pull request
I will close this pull request and resubmit soon with a corrected path.

PR #1921, opened 2017-03-08
Doc: Models

ax3l:
Start a section on models for the default and especially the extended PIC algorithm in PIConGPU.

This section should document the assumptions, applicable ranges and implementation limits of our models, so users can follow those in order to do meaningful science in validated & verified regimes.

@n01r @Heikman please extend your sections after merge :)

### DIFF
--- a/docs/source/models/photons.rst
+++ b/docs/source/models/photons.rst
@@ -1,0 +1,50 @@
+Photons
+=======
+
+Radiation reaction and (hard) photons: why and when are they needed.
+Models we implemented and verified:
+
+* Landau-Lifschitz Model (semi-classical)
+* QED Models (Synchrotron & Bremsstrahlung)
+
+Would be great to add your Diploma Thesis talk with pictures and comments here.
+
+Please add notes and warnings on the models' assumptions for an easy guiding on their usage :)
+
+.. note::
+   Assumptions in Furry-picture and Volkov-States: classical em wave part and QED "pertubation".
+   EM fields on grid (Synchrotron) and density modulations (Bremsstrahlung) need to be locally constant compared to radiated coherence interval ("constant-crossed-field approximation").
+
+.. attention::
+   Bremsstrahlung: The individual electron direction and gamma emission are not correlated.
+   (momentum is microscopically / per e- not conserved, only collectively.)
+
+.. attention::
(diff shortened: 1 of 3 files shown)


ax3l:
@PrometheusPi also interesting for you for the semi-classical RR (LL) part :)

ax3l:
I should also add something on the general pic part in this chapter, have some material in my diploma thesis on that we can use

docs/source/models/photons.rst
```diff
@@ -0,0 +1,50 @@
+Photons
+=======
+
+Radiation reaction and (hard) photons: why and when are they needed.
+Models we implemented and verified:
+
+* Landau-Lifschitz Model (semi-classical)
```
PrometheusPi: The Reduced Landau Lifschitz Model (RLL) is actually a particle pusher and does not support photons, thus I would not place it under *photon models*.

docs/source/models/photons.rst
```diff
@@ -0,0 +1,50 @@
...
+* QED Models (Synchrotron & Bremsstrahlung)
+
+Would be great to add your Diploma Thesis talk with pictures and comments here.
```
PrometheusPi: @Heikman needs replacement

docs/source/models/photons.rst
```diff
@@ -0,0 +1,50 @@
...
+.. note::
+   Assumptions in Furry-picture and Volkov-States: classical em wave part and QED "pertubation".
+   EM fields on grid (Synchrotron) and density modulations (Bremsstrahlung) need to be locally constant compared to radiated coherence interval ("constant-crossed-field approximation").
+
+.. attention::
+   Bremsstrahlung: The individual electron direction and gamma emission are not correlated.
+   (momentum is microscopically / per e- not conserved, only collectively.)
+
+.. attention::
+   "Soft" photons from low energy electrons will get underestimated in intensity below a threshold of ... .
```
PrometheusPi: @Heikman value missing

ax3l:
They all need updates, he can add a PR to dev after merge :)

docs/source/models/photons.rst
ax3l: Yes, we can make a separate section for it but should still link it *from* here since nearly all comparisons will need it :)

PR #1924, opened 2017-03-10
Plugins: Replace Old Analyzer Naming

ax3l:
Replace left-over "analy[s|z]er" speach we used before calling them *plugins*.

Done via one-liner script and commited as generic author:
```bash
grep -iR "analyser" src/ | awk -F: '{print $1}' | \
    sort -u | xargs -n1 -P1 -I{} sed -i 's/analyser/plugin/g' {}
grep -iR "analyzer" src/ | awk -F: '{print $1}' | \
    sort -u | xargs -n1 -P1 -I{} sed -i 's/analyzer/plugin/g' {}

    git commit -S
```

ignored:
 - src/tools/bin/radiationSyntheticDetector


src/tools/bin/radiationSyntheticDetector
```diff
@@ -52,7 +52,7 @@ parser.add_argument('--dataExtend', '-d',
                             [default=0.5 3.5 0.0 360.0]''')
 
 # extend of the data to analyze
-parser.add_argument('--analyzerExtend', '-a',
+parser.add_argument('--pluginExtend', '-a',
```
PrometheusPi: REQUIRED:
please revert this - analyzer is the data analyzer in this tool - not a plugin

src/tools/bin/radiationSyntheticDetector
```diff
@@ -83,11 +83,11 @@ for myfile in args.path2Data:
...
-    # set extent analyzer:
-    omega_min_analy = args.analyzerExtend[0]
-    omega_max_analy = args.analyzerExtend[1]
-    theta_min_analy = args.analyzerExtend[2]
-    theta_max_analy = args.analyzerExtend[3]
+    # set extent plugin:
+    omega_min_analy = args.pluginExtend[0]
+    omega_max_analy = args.pluginExtend[1]
+    theta_min_analy = args.pluginExtend[2]
+    theta_max_analy = args.pluginExtend[3]
```
PrometheusPi: REQUIRED:
same here

ax3l: sure!

but is "analyzer" a great name here either? maybe you want to express something like "selection" or "area of interest" or "extent" here?

PrometheusPi: @ax3l I agree - selection or ROI is a better name for this variable
 I will fix it in another pull request.

PrometheusPi: Sorry - I was in the wrong code - here it is a **statistical analysis** this the name is correct (ROI, etc. would be confusing). 
Just leave it as it is.

ax3l: ok

ax3l: > Just leave it as it is.

oh, you mean *revert those lines*?

PrometheusPi: exactly

PR #1928, opened 2017-03-13
shared memory allocator documenation

psychocoderHPC:
add more documentation to avaoid wrong usage of the macro `PMACC_SMEM()`

### DIFF
--- a/src/libPMacc/include/memory/shared/Allocate.hpp
+++ b/src/libPMacc/include/memory/shared/Allocate.hpp
@@ -105,6 +105,38 @@ namespace shared
 
 /** allocate shared memory
  *
+ * @warning Do not use this macro within a factory method. This macro contains
+ * a pre compiler counter which is only increased once per source code line.
+ * You must not use this macro to create an aliased pointer to shared memory.
+ * Please use `PMacc::memory::shared::allocate< id, type >()` to create shared memory
+ * which is used outside of the method where it is created.
+ *
+ * @code
+ * // THIS IS A EXAMPLE HOW `PMACC_SMEM` SHOULD NOT BE USED
+ * struct Factory
+ * {
+ *     int& getSharedMem( )
+ *     {
+ *         // this macro points always to the same memory address
+ *         // even if this method is called twice
+ *         PMACC_SMEM( sharedMem, int );
+ *         return sharedMem;
+ *     }
+ * };
+ *
+ * // RIGHT USAGE
+ * template< uint32_t T_id >
+ * struct Factory
+ * {
+ *     int& getSharedMem( )
+ *     {
+ *         // create new shared memory for each `T_id`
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/memory/shared/Allocate.hpp
```diff
@@ -105,6 +105,27 @@ namespace shared
 
 /** allocate shared memory
  *
+ * Do not use this macro within a factory method. This macro contains a pre compiler
```
ax3l: worth a `@warning` I think :)

src/libPMacc/include/memory/shared/Allocate.hpp
```diff
@@ -105,6 +105,27 @@ namespace shared
...
+ * counter which is only increased once per source code line.
+ * Please do also not assume that this macro can be used to create aliased pointer
```
ax3l: You can even write: `You must not use this macro to create an aliased pointer to shared memory.`

src/libPMacc/include/memory/shared/Allocate.hpp
```diff
@@ -105,6 +105,27 @@ namespace shared
...
+ * to shared memory!
+ *
+ * @code
+ * // THIS IS A EXAMPLE HOW `PMACC_SMEM` SHOULD NOT BE USED
```
ax3l: an example

really a wrong-usage but not a a right-usage example?

src/libPMacc/include/memory/shared/Allocate.hpp
```diff
@@ -105,6 +105,27 @@ namespace shared
...
+ *     {
+ *         // this macro points always to the same memory address
+ *         // even if this method is called twice
+ *         PMACC_SMEM( sharedMem, int );
+ *         return sharedMem;
+ *     }
+ * };
+ * @endcode
+ *
+ * Use `PMacc::memory::shared::allocate<id,type>()` instead.
```
ax3l: how? code example would be great :)

psychocoderHPC:
updated

src/libPMacc/include/memory/shared/Allocate.hpp
```diff
@@ -105,6 +105,38 @@ namespace shared
...
+ *     }
+ * };
+ *
+ * // RIGHT USAGE
+ * template< uint32_t T_id >
+ * struct Factory
+ * {
+ *     int& getSharedMem( )
+ *     {
+ *         // create new shared memory for each `T_id`
```
PrometheusPi: QUESTION:
What do you mean with *for each `T_id`* - there is just **one** `T_id`?

ax3l: for each specialization of the class with always an other T_id. I find the wording ok

PrometheusPi: Okay - that makes sense

PR #1929, opened 2017-03-14
remove CUDA native static shared memory

### LINKED ISSUE #1927: CUSTL shared memory allocation is wrong
I am on the way to remove the last usage of `__shared__` in PMacc.
I found that CUSTL is using the wrong index order to allocate N-dimensional shared memory.
CUSTL is using `int mem[x][y][z];` but it must be `int mem[z][y][x];` [link to the code](https://github.com/ComputationalRadiationPhysics/picongpu/blob/master/src/libPMacc/include/cuSTL/container/allocator/compile-time/SharedMemAllocator.hpp#L78)

Maybe it breaks nothing because later on the array is handled as 1-dimensional array and the plain data size is in both cases equal.

Currently it is not clear if this can have any side effects, never the less I marked this as BUG and that this effects the latest release.

psychocoderHPC:
replace `__shared__ type varName` with `PMacc::memory::shared::allocate<...>()`

This is a follow up of #1727 and fix #1927 by flat all N dimensional arrays to 1D. 

There is a very small performance increase due to more 32Bit integer calculations (instead of 64Bit) during the address calculation for each shared memory access. 

Tests:
- [x] KHI with electrons/positrons (heating and charge conservation)

### DIFF
--- a/src/libPMacc/include/cuSTL/container/allocator/compile-time/SharedMemAllocator.hpp
+++ b/src/libPMacc/include/cuSTL/container/allocator/compile-time/SharedMemAllocator.hpp
@@ -21,8 +21,12 @@
 
 #pragma once
 
-#include "math/Vector.hpp"
 #include "cuSTL/cursor/compile-time/BufferCursor.hpp"
+#include "memory/shared/Allocate.hpp"
+#include "memory/Array.hpp"
+#include "math/Vector.hpp"
+#include "pmacc_types.hpp"
+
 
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
PTX for a shared memory access within the Esirkepov current deposition (only for documentation)

sharedmemoryaccessptx

ax3l:
I have the slight feeling, the syntax was way more readable before ;)

Issue #667, opened 2015-02-03
Radiation: Remove Defines


ax3l:
With the introduction of multi species, we should refactor the radiation plugin:
- [x] compile the kernels as soon as `HDF5_ENABLE` is available (remove `#define ENABLE_RADIATION`)
- [ ] add the `momentum_mt1` attribute as soon as a user adds the radiation flag to a species
- [ ] throw a compile error, if a user sets the radiation flag in a species but HDF5 is not available

PrometheusPi:
I agree :+1:
@ComputationalRadiationPhysics/picongpu-maintainers 
Should I self assign to this issue?

ax3l:
as you prefer. you can also assign it when you actually start working on it.

PrometheusPi:
I will self assign **now** otherwise I will forget about it :cold_sweat:

ax3l:
sure :)

PrometheusPi:
Discussed *offline* with @psychocoderHPC 
 - radiation flags could be used outside the context of the radiation plugin (and `momentum_mt1` can also used outside) thus we should avoid such a tight coupling
  - currently the radiation plugin **only** compiles if openPMD-api **is** available - thus any attempt to run the radiation output with output accidentally disabled is no longer an issue. 
  
  We thus decided to close this issue as no longer relevant.

PR #1930, opened 2017-03-14
Remove nyquist limit switch

PrometheusPi:
This pull request removes the pre-compiler switch to exclude the Nyquist limit calculation and thus partially solves issue #667. 

Not calculating the Nyquist limit makes only sense if computing low frequency radiation - below `omega < pi / (2 * delta_t) ` (factor 2 due to propagation away from the detector at the speed of light). This is already resolved by the grid. In most cases, the Nyquist limit should be taken into account. Thus removing the option avoid misconfiguration by accidentally not activating this limit.

### DIFF
--- a/src/picongpu/include/plugins/radiation/Radiation.kernel
+++ b/src/picongpu/include/plugins/radiation/Radiation.kernel
@@ -51,9 +51,7 @@
 #include "memory/Array.hpp"
 
 
-#if( __NYQUISTCHECK__ == 1 )
-#   include "plugins/radiation/nyquist_low_pass.hpp"
-#endif
+#include "plugins/radiation/nyquist_low_pass.hpp"
 
 #include "plugins/radiation/radFormFactor.hpp"
 #include "sys/stat.h"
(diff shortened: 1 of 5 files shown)


src/picongpu/include/plugins/radiation/Radiation.kernel
```diff
@@ -51,9 +51,7 @@
 #include "memory/Array.hpp"
 
 
-#if( __NYQUISTCHECK__ == 1 )
 #   include "plugins/radiation/nyquist_low_pass.hpp"
```
psychocoderHPC: please remove the indention

src/picongpu/include/simulation_defines/unitless/radiationConfig.unitless
```diff
@@ -28,6 +28,7 @@
 
 PMACC_CASSERT_MSG(ENABLE_ELECTRONS_must_set_to_1__change_ENABLE_ELECTRONS_in_file_componentsConfig_param,ENABLE_ELECTRONS==1);
 
+PMACC_CASSERT_MSG( The_Nyquist_limit_needs_to_be_below_one, ( picongpu::radiationNyquist::NyquistFactor < 1.0 ) );
```
psychocoderHPC: Is a negative value allowed? If not please check.

src/picongpu/include/plugins/radiation/Radiation.kernel
PrometheusPi: Good point - will fix it

src/picongpu/include/simulation_defines/unitless/radiationConfig.unitless
PrometheusPi: Good idea. The value should be larger than zero. Thus another check would be great.

PrometheusPi:
@psychocoderHPC I updated the pull request.

PR #1936, opened 2017-03-16
optimize `AssignedTrilinearInterpolation`

psychocoderHPC:
This pull request remove the usage of floats for loop indices. The advantage is that the loop length is known at compile time and the compiler can optimize much more.

- use integer indices for the loops
- remove pragma unroll

Tests KHI positrons/electrons:
- clang [version 4.0.0 (trunk 290769)]
  - [X] speed: no speedup
  - [X] heating test
- nvcc [release 8.0, V8.0.44] 
  - [X] speed: 2-4% speedup
  - [X] heating test

### DIFF
--- a/src/picongpu/include/algorithms/AssignedTrilinearInterpolation.hpp
+++ b/src/picongpu/include/algorithms/AssignedTrilinearInterpolation.hpp
@@ -74,24 +74,21 @@ struct AssignedTrilinearInterpolation
         typedef typename ::PMacc::result_of::Functor<AssignedTrilinearInterpolation, Cursor>::type type;
 
         type result_z = type(0.0);
-#pragma unroll 4
-        for (float_X z = Begin; z <= End; z += float_X(1.0))
+        for (int z = Begin; z <= End; ++z)
         {
             type result_y = type(0.0);
-#pragma unroll 4
-            for (float_X y = Begin; y <= End; y += float_X(1.0))
+            for (int y = Begin; y <= End; ++y)
             {
                 type result_x = type(0.0);
-#pragma unroll 4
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
The register usage for `moveAndMarkParticles` increase but this has no negative effect on the performance.

interpolationopti

ax3l:
LGTM, any comment on the removed `#pragma unroll 4`?

src/picongpu/include/algorithms/AssignedTrilinearInterpolation.hpp
```diff
@@ -74,24 +74,21 @@ struct AssignedTrilinearInterpolation
         typedef typename ::PMacc::result_of::Functor<AssignedTrilinearInterpolation, Cursor>::type type;
 
         type result_z = type(0.0);
-#pragma unroll 4
-        for (float_X z = Begin; z <= End; z += float_X(1.0))
+        for (int z = Begin; z <= End; ++z)
```
ax3l: since you are touching the for loops anyway, feel free to make `their( spacings, pretty )`, there is not much other code in the file :)

psychocoderHPC: OK I do it for you^^

psychocoderHPC:
@ax3l I updated the code style of the full file. All changes are in an extra commit and only changes the style.

ax3l:
huh that was more then expected. anyway, ty!

psychocoderHPC:
I can only think binary, all or nothing^^

ax3l:
ok ok, I won't ask again :D

PR #1798, opened 2017-02-06
update mallocMC

psychocoderHPC:
Update `mallocMC`. The interface changes in `mallocMC` make it necessary  that `PMacc` and `PIConGPU` also need some interface changes. All in all the change is that we need to pass the mallocMC heap to the constructor of the particle classes.

- update third party mallocMC
- PIConGPU
  - delete `mallocMCconfig.hpp`
  - add param file `mallocMC.param`
  - `main.cu` - remove mallocMC include
  - ADIOS: change `MallocMCBuffer` to `MallocMCBuffer<DeviceHeap>`
  - `MySimulation.hpp`
    - add member `deviceHeap` (shared pointer)
    - change `MallocMCBuffer` to `MallocMCBuffer<DeviceHeap>`
  - `Particles<>` - extend constructor parameters
  - `ParticlesFunctors` - consider `Particles<>` interface changes
- PMacc
  - `ParticlesBase`:
    - add template parameter `T_DeviceHeap`
    - add device heap to constructor parameter
  - `MallocMCBuffer` - add template parameter `T_DeviceHeap`
  - `ParticlesBox` - add member for mallocMC device heap

Tests:
- [x] KHI heating electrons/positrons

PR #1938, opened 2017-03-16
fix `PMacc::MallocMCBuffer`

### LINKED ISSUE #1934: adios IO broken
The Adios IO plugin is broken. It is not possible to dump data with adios.
It end up with the error message in `MallocMcBuffer`

error message:
```
[CUDA] Error: </bigdata/hplsim/scratch/widera/dev/src/libPMacc/include/particles/memory/buffers/MallocMCBuffer.tpp>:82 
  what():  [CUDA] Error: invalid argument
```
The reason for this bug is that the member `deviceHeapInfo` of the `MallocMCBuffer` contains wrong data. The data about the mallocMC provided from mallocMC are correct but something went wrong during the copy.

**The bug is not in the latest release**

psychocoderHPC:
close  #1934

The bug is that `decltype` returned a reference pointing to temporary memory and was introduced with #1798.
Use explicit `mallocMC::HeapInfo` instead of catching the type with `std::declval` and `decltype`.

This bug is not effecting the latest release!

Tests: 

- [x] KHI restart

### DIFF
--- a/src/libPMacc/include/particles/memory/buffers/MallocMCBuffer.hpp
+++ b/src/libPMacc/include/particles/memory/buffers/MallocMCBuffer.hpp
@@ -25,9 +25,9 @@
 #include "dataManagement/ISimulationData.hpp"
 
 #include "mallocMC/mallocMC.hpp"
+
 #include <string>
 #include <memory>
-#include <utility>
 
 namespace PMacc
 {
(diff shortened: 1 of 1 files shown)


ax3l:
thanks for catching that, great!

that's the reason why explicit types are superior in readability and safety to `auto`/`decltype` indirections ;)

psychocoderHPC:
I finished the restart test

PR #1949, opened 2017-03-27
Docs: Typos, Modules

ax3l:
Fix a few typos in install docs, add classic modules as reference.

### DIFF
--- a/docs/source/install/path.rst
+++ b/docs/source/install/path.rst
@@ -26,7 +26,7 @@ Spack
 ^^^^^
 
 [Spack]_ is a flexible package manager for HPC systems that can organize versions and dependencies for you.
-It can be configured once for your hardware architecture to create optimally tuned binaries and provides module file support (e.g. [Lmod]_).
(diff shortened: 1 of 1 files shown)


docs/source/install/path.rst
```diff
@@ -114,9 +114,16 @@ References
 .. [Spack]
         T. Gamblin and contributors.
         *A flexible package manager that supports multiple versions, configurations, platforms, and compilers*,
+        SC '15 Proceedings of the International Conference for High Performance Computing, Networking, Storage and Analysis (2015),
```
PrometheusPi: OPTIONAL:
On the [web site](https://www.ieee.org/conferences_events/conferences/conferencedetails/index.html?Conf_ID=32761) it says: `SC15: International Conference for High Performance Computing, Networking, Storage and Analysis` so no `SC '15` but a `SC15` instead.

ax3l: They don't do this quite consequently, I used the spelling from [here](https://dx.doi.org/10.1145/2807591.2807623)

PrometheusPi: Okay - if ACM varies in its spelling - we can do that too 😄

PR #1955, opened 2017-04-07
Doc: Start a User Workflows Section

ax3l:
Starts a user workflows section in our docs with workflows to:
- set the number of cells of a simulation
- change the resolution of a simulation

### DIFF
--- a/docs/source/usage/workflows/numberOfCells.rst
+++ b/docs/source/usage/workflows/numberOfCells.rst
@@ -1,0 +1,15 @@
+.. _usage-workflows-numberOfCells:
+
+Setting the Number of Cells
+---------------------------
+
+Together with the grid resolution in :ref:`gridConfig.param <usage-params-core>`, the number of cells in our :ref:`.cfg files <usage-tbg>` determine the overall size of a simulation (box).
+The following rules need to be applied when setting the number of cells:
+
+Each GPU needs to:
+
+#. contain an integer *multiple* of supercells
+#. at least *three* supercells
+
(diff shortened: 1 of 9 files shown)


docs/source/usage/workflows/numberOfCells.rst
```diff
@@ -0,0 +1,15 @@
+.. _usage-workflows-numberOfCells:
+
+Setting the Number of Cells
+---------------------------
+
+Together with the grid size in :ref:`gridConfig.param <usage-params-core>`, the number of cells in our :ref:`.cfg files <usage-tbg>` determine the overall size of a simulation (box).
```
PrometheusPi: OPTIONAL:
better use `resolution` than `size`

docs/source/usage/workflows/numberOfCells.rst
```diff
@@ -0,0 +1,15 @@
...
+The following rules need to be applied when setting the number of cells:
+
+Each GPU needs to:
+
+#. contain an integer *multiple* of supercells
+#. at least *three* supercells
+
+Supercell sizes in terms of number of cells are set in :ref:`memory.param <usage-params-memory>` and are by default ``8x8x4`` for 3D3V simulations on GPUs.
+For 2D3V simulations, ``16x16`` is usually a good supercell size.
```
PrometheusPi: OPTIONAL:
One could mention that however the default is `8x8`.

ax3l: ... hm, we might better change the default, too.

PR #1958, opened 2017-04-10
Feature: index pool

Heikman:
Add a new container to libpmacc, that represents an index pool.

important, self-explaining methods:
 - `get()` get a new index
 - `release()` release an index

Tested with the particle merging plugin (future PR).

### DIFF
--- a/src/libPMacc/include/memory/IndexPool.hpp
+++ b/src/libPMacc/include/memory/IndexPool.hpp
@@ -1,0 +1,157 @@
+/* Copyright 2017 Heiko Burau
+ *
+ * This file is part of libPMacc.
+ *
+ * libPMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * libPMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with libPMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "pmacc_types.hpp"
+#include "memory/Array.hpp"
+#include <type_traits>
+
+namespace PMacc
+{
+namespace memory
+{
+
+    /* A memory pool of dynamic size containing indices
+     *
+     * \tparam T_Index type of index
+     * \tparam T_maxSize maximum number of indices
+     *
+     * Warning: This class is not thread-safe!
+     */
+    template<
+        typename T_Index,
+        size_t T_maxSize
+    >
+    struct IndexPool
+    {
+    private:
+
+        /* Reverse-iterator of the memory pool. The pool is iterated reversely
+         * to ensure removal of the current element while iterating.
+         */
+        struct ReverseIterator
+        {
+            T_Index* pointer;
+
+            HDINLINE
+            ReverseIterator( T_Index* const pointer ) : pointer( pointer )
+            {}
+
+            HDINLINE
+            void operator++()
+            {
+                this->pointer--;
+            }
+
+            HDINLINE
+            T_Index& operator*()
+            {
+                return *(this->pointer);
+            }
+
+            HDINLINE
+            bool operator!=( const ReverseIterator& other ) const
+            {
+                return this->pointer != other.pointer;
+            }
+        };
+
+        size_t m_size;
+        Array<
+            T_Index,
+            T_maxSize
+        > listIds;
+
+    public:
+
+        using Index = T_Index;
+
+        /* init pool with consecutive indices
+         *
+         * \param size initial number of indices
+         */
+        HDINLINE
+        IndexPool( const Index size = 0 ) : m_size( size )
+        {
+            /* TODO: parallelize */
+            for( size_t i = 0; i < T_maxSize; i++ )
+                this->listIds[i] = static_cast< Index >( i );
+        }
+
+        /* get a new index */
+        HDINLINE
+        Index get()
+        {
+            if( this->m_size == T_maxSize - 1 )
+                return Index(-1);
+
+            return this->listIds[this->m_size++];
+        }
+
+        /* release an index */
+        HDINLINE
+        void release( const Index idx )
+        {
+            /* find position of `idx` */
+            size_t pos;
+            for( size_t i = 0; i < this->m_size; i++ )
+            {
+                if( this->listIds[i] == idx )
+                {
+                    pos = i;
+                    break;
+                }
+            }
+
+            this->listIds[pos] = this->listIds[--this->m_size];
+            this->listIds[this->m_size] = idx;
+        }
+
+        /** get number of indices within pool */
+        HDINLINE
+        size_t size( ) const
+        {
+            return this->m_size;
+        }
+
+        /** get maximum number of indices within pool */
+        HDINLINE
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/memory/IndexPool.hpp
```diff
@@ -0,0 +1,157 @@
...
+#include <type_traits>
+
+namespace PMacc
+{
+namespace memory
+{
+
+    /* A memory pool of dynamic size containing indices
+     *
+     * \tparam T_Index type of index
```
psychocoderHPC: only a note: We use `@` for doxygen comments

src/libPMacc/include/memory/IndexPool.hpp
```diff
@@ -0,0 +1,157 @@
...
+     * \tparam T_maxSize maximum number of indices
+     *
+     * Warning: This class is not thread-safe!
```
psychocoderHPC: Please move this to the sort brief part

```
/** brief
 *
 * short brief
 *
 * @param ....
 */

File: src/libPMacc/include/memory/IndexPool.hpp [hunk 3]
```diff
@@ -0,0 +1,157 @@
...
+     */
+    template<
+        typename T_Index,
+        size_t T_maxSize
+    >
+    struct IndexPool
+    {
+    private:
+
+        /* Reverse-iterator of the memory pool. The pool is iterated reversely
```
psychocoderHPC: need `/**` to be passed to doxygen

src/libPMacc/include/memory/IndexPool.hpp
```diff
@@ -0,0 +1,157 @@
...
+        Array<
+            T_Index,
+            T_maxSize
+        > listIds;
+
+    public:
+
+        using Index = T_Index;
+
+        /* init pool with consecutive indices
```
psychocoderHPC: `/**`

src/libPMacc/include/memory/IndexPool.hpp
```diff
@@ -0,0 +1,157 @@
...
+         */
+        HDINLINE
+        IndexPool( const Index size = 0 ) : m_size( size )
+        {
+            /* TODO: parallelize */
+            for( size_t i = 0; i < T_maxSize; i++ )
+                this->listIds[i] = static_cast< Index >( i );
+        }
+
+        /* get a new index */
```
psychocoderHPC: `/**`

src/libPMacc/include/memory/IndexPool.hpp
```diff
@@ -0,0 +1,157 @@
...
+        HDINLINE
+        Index get()
+        {
+            if( this->m_size == T_maxSize - 1 )
+                return Index(-1);
+
+            return this->listIds[this->m_size++];
+        }
+
+        /* release an index */
```
psychocoderHPC: `/**`

src/libPMacc/include/memory/IndexPool.hpp
```diff
@@ -0,0 +1,157 @@
...
+            }
+
+            HDINLINE
+            T_Index& operator*()
+            {
+                return *(this->pointer);
+            }
+
+            HDINLINE
+            bool operator!=( const ReverseIterator& other ) const
```
psychocoderHPC: note: new coding style is `VarType const & name`

src/libPMacc/include/memory/IndexPool.hpp
```diff
@@ -0,0 +1,157 @@
...
+            /* TODO: parallelize */
+            for( size_t i = 0; i < T_maxSize; i++ )
+                this->listIds[i] = static_cast< Index >( i );
+        }
+
+        /* get a new index */
+        HDINLINE
+        Index get()
+        {
+            if( this->m_size == T_maxSize - 1 )
```
psychocoderHPC: `T_maxSize` is of the type `size_t` it should be `if( this->m_size == T_maxSize - 1u)`

**Warning** this line can produce a unsigned integer underflow! Is this allowed?

src/libPMacc/include/memory/IndexPool.hpp
```diff
@@ -0,0 +1,157 @@
...
+                return Index(-1);
```
psychocoderHPC: **Warning** integer underflow! If this behavior is volitional than please add a note to the doxygen string that the returned  index can be extreme large and if used to access an array it can produce out of memory access.

src/libPMacc/include/memory/IndexPool.hpp
```diff
@@ -0,0 +1,157 @@
...
+        HDINLINE
+        constexpr size_t max_size( ) const
+        {
+            return T_maxSize;
+        }
+
+        HDINLINE
+        ReverseIterator begin()
+        {
+            return ReverseIterator( this->listIds.data() + this->m_size - 1 );
```
psychocoderHPC: should be `... - 1u` to avoid implicit casts.

src/libPMacc/include/memory/IndexPool.hpp
ax3l: you can also use Doxygen's `@warning` in there :)

src/libPMacc/include/memory/IndexPool.hpp
ax3l: `Index(-1)` is not a good idea at all for arbitrary template arguments and can break static code analysis tools. Please use a well-defined const instead from `<limits>`, e.g. `std::numeric_limits< Index >::max()`

and also see my comment above to add a static assert if `Index` actually is an integer-like type.

src/libPMacc/include/memory/IndexPool.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+        size_t m_size;
+        Array<
+            T_Index,
+            T_maxSize
+        > listIds;
+
+    public:
+
+        using Index = T_Index;
```
ax3l: the question is, if one actually needs that flexibility or just uses a meaningful type directly and writes a `.cpp` file to reduce compile time.

If it makes sense to keep a `T_Index`, please add at least a static assert to verify it is integer-like, e.g. via the trait [`std::is_integer`](http://en.cppreference.com/w/cpp/types/numeric_limits/is_integer) ( I don't even thing integral types would be useful)

src/libPMacc/include/memory/IndexPool.hpp
```diff
@@ -0,0 +1,157 @@
...
+#include "pmacc_types.hpp"
+#include "memory/Array.hpp"
+#include <type_traits>
+
+namespace PMacc
+{
+namespace memory
+{
+
+    /* A memory pool of dynamic size containing indices
```
ax3l: can you please add a brief about its behavior, performance expectations / what it is useful for?

ax3l: behavior example: are indexes given consecutively, is adding/removing and index expensive or lightweight (list vs array vs map vs ...) etc.

src/libPMacc/include/memory/IndexPool.hpp
Heikman: I agree adding a static assert for an signed(!) integer-like type.
But then as a user I would prefer to check the result against `Index(-1)` because it is intuitively clear that the index in invalid.

src/libPMacc/include/memory/IndexPool.hpp
Heikman: I have added a static assert checking that `T_maxSize` is greater than zero. But honestly I find we should be too strict in this case and persume also common sense to the user.

src/libPMacc/include/memory/IndexPool.hpp
Heikman: Remark: this makes no sense for references.

Heikman:
Having a `T_Index` makes sense in size-critical environments like the shared memory in cuda.

ax3l:
@psychocoderHPC if you are ok with this I would merge it

ax3l:
@psychocoderHPC is it ok if I merge this PR? it's updated :)

psychocoderHPC:
it is yours now

Issue #1652, opened 2016-11-23
move radiation  utililities to libPMacc


PrometheusPi:
As suggested by @ax3l in #1644 the utility functions used in the radiation plugin should go into a more general scope e.g. into libPMacc.

PrometheusPi:
@psychocoderHPC and @ax3l Where in linPMacc would you best put the utilities?

ax3l:
`libPMacc/include/algorithms/math`, no?

PrometheusPi:
The main problem with that is that these utilities are templates methods and those work for both scalar and vectors (if some operations are defined). 
@ax3l and @psychocoderHPC Any suggestions?

n01r:
changed the title from
> move radiation utililities to libPMacc

to   
**move radiation utilities to libPMacc**

ax3l:
@PrometheusPi did this happen already? otherwise please move it to milestone "Future" until someone starts working on it.

PR #1959, opened 2017-04-10
Feature: particle merging

Heikman:
add a particle merging algorithm based on:

Luu, P. T., Tückmantel, T., & Pukhov, A. (2016).
Voronoi particle merging algorithm for PIC codes.
Computer Physics Communications, 202, 165-174.

## To Do
 - [x] rebase against `dev` (#1958  is merged)
 - [x] add wiki page: https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-Plugins
 - [x] add section in `docs/TBG_macros.cfg`
 - [x] add test case / tests

## Dependencies
 - [x] #1958

### KHI phase space, standart example, timestep: 1400, without merging
download1
### with merging
download2
parameter: 
```json
"e_merger": {
  "period": 100,
  "minParticlesToMerge": 8,
  "posSpreadThreshold": 0.2,
  "absMomSpreadThreshold": 0.01,
  "minMeanEnergy": 0
}
```
download3

### DIFF
--- a/src/picongpu/include/plugins/particleMerging/ParticleMerger.hpp
+++ b/src/picongpu/include/plugins/particleMerging/ParticleMerger.hpp
@@ -25,6 +25,8 @@
 #include "simulation_classTypes.hpp"
 #include "plugins/ISimulationPlugin.hpp"
 
+#include "traits/HasFlag.hpp"
+
 #include <string>
 #include <iostream>
 #include <fstream>
@@ -38,6 +40,7 @@ namespace particleMerging
 {
 
     using namespace PMacc;
+    namespace bmpl = boost::mpl;
 
     /** Implements a particle merging algorithm based on
     *
@@ -47,8 +50,19 @@ namespace particleMerging
     *
     * \tparam T_ParticlesType particle species
     */
+    template<
+        class T_ParticlesType,
+        bool hasVoronoiCellId =
+            PMacc::traits::HasFlag<
+                typename T_ParticlesType::FrameType,
+                voronoiCellId
+            >::type::value
+    >
+    struct ParticleMergerWrapped;
+
+
     template< class T_ParticlesType >
-    class ParticleMerger : public ISimulationPlugin
+    struct ParticleMergerWrapped< T_ParticlesType, true > : ISimulationPlugin
     {
     private:
         std::string name;
@@ -67,7 +81,7 @@ namespace particleMerging
     public:
         using ParticlesType = T_ParticlesType;
 
-        ParticleMerger() :
+        ParticleMergerWrapped() :
             name(
                 "ParticleMerger: merges several macroparticles with"
                 " similar position and momentum into a single one"
@@ -193,7 +207,7 @@ namespace particleMerging
             return this->name;
         }
 
-    private:
+    protected:
 
         void pluginLoad()
         {
@@ -247,6 +261,63 @@ namespace particleMerging
         void checkpoint( uint32_t, const std::string )
         {}
     };
+
+
+    template< class T_ParticlesType >
+    struct ParticleMergerWrapped< T_ParticlesType, false > : ISimulationPlugin
+    {
+    private:
+        std::string name;
+        std::string prefix;
+    uint32_t notifyPeriod;
+        MappingDesc* cellDescription;
+
+    public:
+        using ParticlesType = T_ParticlesType;
+
+        ParticleMergerWrapped() :
+            name(
+                "ParticleMerger: merges several macroparticles with"
+                " similar position and momentum into a single one.\n"
+                "plugin disabled. Enable plugin by adding the `voronoiCellId`"
+                " attribute to the particle attribute list."
+            ),
+            prefix( ParticlesType::FrameType::getName() + std::string("_merger") ),
+            notifyPeriod( 0 ),
+            cellDescription( nullptr )
+        {
+            Environment<>::get().PluginConnector().registerPlugin( this );
+        }
+
+        std::string pluginGetName() const
+        {
+            return this->name;
+        }
+
+    protected:
+        void setMappingDescription(MappingDesc*)
+        {}
+
+        void pluginRegisterHelp(po::options_description&)
+        {}
+
+        void pluginUnload()
+        {}
+
+        void restart( uint32_t, const std::string )
+        {}
+
+        void checkpoint( uint32_t, const std::string )
+        {}
+
+        void notify(uint32_t)
+        {}
+    };
+
+
+    template< typename T_ParticlesType >
+    struct ParticleMerger : ParticleMergerWrapped< T_ParticlesType >
+    {};
 
 } // namespace particleMerging
 } // namespace plugins
--- a/src/picongpu/include/plugins/particleMerging/ParticleMerger.hpp
+++ b/src/picongpu/include/plugins/particleMerging/ParticleMerger.hpp
@@ -25,7 +25,7 @@
 #include "simulation_classTypes.hpp"
 #include "plugins/ISimulationPlugin.hpp"
 
-#include "traits/HasFlag.hpp"
+#include "traits/HasIdentifier.hpp"
 
 #include <string>
 #include <iostream>
@@ -44,7 +44,7 @@ namespace particleMerging
 
     /** Implements a particle merging algorithm based on
     *
-    * Luu, P. T., Tückmantel, T., & Pukhov, A. (2016).
+    * Luu, P. T., Tueckmantel, T., & Pukhov, A. (2016).
     * Voronoi particle merging algorithm for PIC codes.
     * Computer Physics Communications, 202, 165-174.
     *
@@ -53,7 +53,7 @@ namespace particleMerging
     template<
         class T_ParticlesType,
         bool hasVoronoiCellId =
-            PMacc::traits::HasFlag<
+            PMacc::traits::HasIdentifier<
                 typename T_ParticlesType::FrameType,
                 voronoiCellId
             >::type::value
@@ -95,7 +95,8 @@ namespace particleMerging
 
         void notify(uint32_t currentStep)
         {
-            typedef typename MappingDesc::SuperCellSize SuperCellSize;
+            using SuperCellSize = MappingDesc::SuperCellSize;
+
             const PMacc::math::Int<simDim> coreBorderGuardSuperCells =
                 this->cellDescription->getGridSuperCells();
             const uint32_t guardSuperCells =
@@ -220,22 +221,22 @@ namespace particleMerging
             );
 
             // clean user parameters
-            PMACC_ASSERT_MSG(
+            PMACC_VERIFY_MSG(
                 this->minParticlesToMerge > 1,
                 std::string("[Plugin: ") + this->prefix + "] minParticlesToMerge"
                 " has to be greater than one."
             );
-            PMACC_ASSERT_MSG(
+            PMACC_VERIFY_MSG(
                 this->posSpreadThreshold >= float_X(0.0),
                 std::string("[Plugin: ") + this->prefix + "] posSpreadThreshold"
                 " has to be non-negative."
             );
-            PMACC_ASSERT_MSG(
-                this->absMomSpreadThreshold * this->relMomSpreadThreshold < float( 0.0 ),
+            PMACC_VERIFY_MSG(
+                this->absMomSpreadThreshold_mc * this->relMomSpreadThreshold < float( 0.0 ),
                 std::string("[Plugin: ") + this->prefix + "] either"
                 " absMomSpreadThreshold or relMomSpreadThreshold has to be given"
             );
-            PMACC_ASSERT_MSG(
+            PMACC_VERIFY_MSG(
                 this->minMeanEnergy >= float_X(0.0),
                 std::string("[Plugin: ") + this->prefix + "] minMeanEnergy"
                 " has to be non-negative."
@@ -295,10 +296,10 @@ namespace particleMerging
         }
 
     protected:
-        void setMappingDescription(MappingDesc*)
+        void setMappingDescription( MappingDesc* )
         {}
 
-        void pluginRegisterHelp(po::options_description&)
+        void pluginRegisterHelp( po::options_description& )
         {}
 
         void pluginUnload()
@@ -310,7 +311,7 @@ namespace particleMerging
         void checkpoint( uint32_t, const std::string )
         {}
 
-        void notify(uint32_t)
+        void notify( uint32_t )
         {}
     };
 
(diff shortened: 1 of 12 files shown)


ax3l:
@Heikman please fix the travis-ci issues and rebase against `dev` (index pool is merged now) :)

ax3l:
@Heikman can you please add your test results to the PR description?

src/picongpu/include/plugins/particleMerging/VoronoiCell.hpp
```diff
@@ -0,0 +1,261 @@
...
+namespace picongpu
+{
+namespace plugins
+{
+namespace particleMerging
+{
+
+    /* Status of a Voronoi cell
+     *
+     * collecting:
```
psychocoderHPC: Please check [this](http://stackoverflow.com/a/13756348) short HOWTO document enums with doxygen

src/picongpu/include/plugins/particleMerging/VoronoiCell.hpp
```diff
@@ -0,0 +1,261 @@
...
+        }
+
+        HDINLINE
+        float_X getMeanMomentum2() const
+        {
+            return math::abs2( this->meanValue );
+        }
+
+
+        /* determine in which of the two sub-Voronoi cells a particle falls */
```
psychocoderHPC: `/**`

src/picongpu/include/plugins/particleMerging/VoronoiCell.hpp
```diff
@@ -0,0 +1,261 @@
...
+                valParticle < meanVoronoi ?
+                this->lowerCellId :
+                this->higherCellId
+            ;
+        }
+
+
+        /** calculate the maxmimum squared spread in position
+         *
+         * @param component index of position component of maxmimum spread
```
psychocoderHPC: `@return` is missing

src/picongpu/include/plugins/particleMerging/VoronoiCell.hpp
```diff
@@ -0,0 +1,261 @@
...
+                }
+            }
+
+            return maxPosSpread2;
+        }
+
+
+        /** calculate the maxmimum squared spread in momentum
+         *
+         * @param component index of momentum component of maxmimum spread
```
psychocoderHPC: `@return` is missing

src/picongpu/include/simulation_defines/param/particleMerging.param
```diff
@@ -0,0 +1,34 @@
...
+#pragma once
+
+namespace picongpu
+{
+namespace plugins
+{
+namespace particleMerging
+{
+
+	/* maximum number of active Voronoi cells per supercell */
```
psychocoderHPC: - need `/**`
- Can I choose any number or is the number of cells restricted by the size of the shared memory?

src/picongpu/include/plugins/particleMerging/ParticleMerger.hpp
```diff
@@ -0,0 +1,253 @@
...
+            if( notifyPeriod == 0 )
+                return;
+
+            Environment<>::get().PluginConnector().setNotificationPeriod(
+                this,
+                notifyPeriod
+            );
+
+            // clean user parameters
+            PMACC_ASSERT_MSG(
```
psychocoderHPC: please use `PMACC_VERIFY_MSG` for all checks, to avoid that this checks are disabled when compiling in release mode.

src/picongpu/include/plugins/particleMerging/ParticleMerger.kernel
```diff
@@ -0,0 +1,422 @@
...
+        MAX_VORONOI_CELLS
+    >;
+    using VoronoiIndexPool = memory::IndexPool<
+        int32_t,
+        MAX_VORONOI_CELLS
+    >;
+
+
+    ParticlesBox particlesBox;
+    uint32_t minParticlesToMerge;
```
psychocoderHPC: please add documentation

src/picongpu/include/plugins/particleMerging/ParticleMerger.kernel
```diff
@@ -0,0 +1,422 @@
...
+
+    ParticleMergerKernel(
+        ParticlesBox particlesBox,
+        uint32_t minParticlesToMerge,
+        float_X posSpreadThreshold,
+        float_X absMomSpreadThreshold,
+        float_X relMomSpreadThreshold,
+        float_X minMeanEnergy
+    ) :
+        particlesBox( particlesBox ),
```
psychocoderHPC: Note: It would be more save to to use `m_VARNAME` for the local member to avoid that variables are shadowed (can be checked with the compiler flag `-Wshadow`). This was a source of many bugs in the past.

src/picongpu/include/plugins/particleMerging/ParticleMerger.hpp
ax3l: note: release note is the new default

src/picongpu/include/simulation_defines/param/particleMerging.param
ax3l: no tabs please (see travis check: failed)

src/picongpu/include/plugins/particleMerging/ParticleMerger.hpp
```diff
@@ -0,0 +1,253 @@
...
+            prefix( ParticlesType::FrameType::getName() + std::string("_merger") ),
+            notifyPeriod( 0 ),
+            cellDescription( nullptr )
+        {
+            Environment<>::get().PluginConnector().registerPlugin( this );
+        }
+
+        void notify(uint32_t currentStep)
+        {
+            typedef typename MappingDesc::SuperCellSize SuperCellSize;
```
ax3l: pls use C++11 `using` syntax

src/picongpu/include/plugins/particleMerging/ParticleMerger.kernel
```diff
@@ -0,0 +1,422 @@
...
+/** Implements a particle merging algorithm based on
+ *
+ * Luu, P. T., Tückmantel, T., & Pukhov, A. (2016).
+ * Voronoi particle merging algorithm for PIC codes.
+ * Computer Physics Communications, 202, 165-174.
+ *
+ * \tparam T_ParticlesBox container of the particle species
+ */
+template< class T_ParticlesBox >
+struct ParticleMergerKernel
```
ax3l: pls align this struct to level 1 (4 spaces)

src/picongpu/include/plugins/particleMerging/ParticleMerger.kernel
```diff
@@ -0,0 +1,422 @@
...
+#include "memory/IndexPool.hpp"
+
+namespace picongpu
+{
+namespace plugins
+{
+namespace particleMerging
+{
+
+using namespace PMacc;
```
ax3l: pls align this line to level 1 (4 spaces)

src/picongpu/include/simulation_defines/unitless/speciesAttributes.unitless
```diff
@@ -296,6 +296,49 @@ struct WeightingPower<weighting>
...
+    // the voronoiCellId attribute is not a physical parameter
+    static bool get()
+    {
+        return false;
+    }
+};
+template<>
+struct WeightingPower<voronoiCellId>
+{
+    /* 1 * weighting^1 == weighting: real particles contibute linearily
```
ax3l: wrong comment. add: `// voronoi cell ids ids do not scale with weighting`

src/picongpu/include/plugins/particleMerging/ParticleMerger.kernel
```diff
@@ -0,0 +1,422 @@
...
+
+        /* conversion from a multi-dim cell coordinate to a linear coordinate
+         * of the cell in its super cell */
+        const int linearThreadIdx = PMacc::math::linearize(
+            PMacc::math::CT::shrinkTo<SuperCellSize, simDim-1>::type::toRT(),
+            threadIndex
+        );
+
+        /* number of initial Voronoi cells */
+        PMACC_SMEM( numInitialVoronoiCells, uint16_t );
```
Heikman: Well, it's technically a runtime variable due to the `pow( 2, simDim )` function call in its initialization. We could surely replace it with `simDim == DIM3 ? 8u : 4u`, but how pretty is it?

psychocoderHPC: Pow can also be calculated at compile time, we hat already a compile time pow implemented in [radioation utilities](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/picongpu/include/plugins/radiation/utilities.hpp#L86) but this function is not exposed to PMacc (#1652).
For this special case you can use `1<<simDim` and add a comment that this is equal to `pow(2,simDim)` than it removes a shared memory variable which speedups the code if we go to cpus.

src/picongpu/include/plugins/particleMerging/VoronoiCell.hpp
```diff
@@ -0,0 +1,261 @@
...
+                }
+            }
+        }
+
+
+        /** finalize mean value calculation */
+        HDINLINE
+        void finalizeMeanValues()
+        {
+            this->meanValue /= this->numRealParticles;
```
Heikman: this is a collective method but how do you see a problem here anyway?

psychocoderHPC: In [this line](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1959/files#diff-00f205db21c4990b088ff306527f100fR143) `meanValue` is changes by all threads with a atomic function. If a warp is still changing `meanValue` during a other thread is performing `finalizeMeanValues()` than we have a classical race condition.

src/picongpu/include/plugins/particleMerging/VoronoiCell.hpp
```diff
@@ -0,0 +1,261 @@
...
+            numMacroParticles( 0 ),
+            numRealParticles( float_X( 0.0 ) ),
+            meanValue( float3_X::create( 0.0 ) ),
+            meanSquaredValue( float3_X::create( 0.0 ) ),
+            firstParticleFlag( 0 )
+        {}
+
+
+        HDINLINE
+        void setToAbort()
```
psychocoderHPC: Please do not forget the documentation!

src/picongpu/include/plugins/particleMerging/VoronoiCell.hpp
Heikman: it is safe in this case because `finalizeMeanValues` is serially executed.

psychocoderHPC:
But all threads accumulate into the same variable if there is no synchronize it end in a racecondition.
What do you mean with serially accecuted? 

Am 29. Mai 2017 11:34:45 MESZ, schrieb Heiko Burau <notifications@github.com>:
>Heikman commented on this pull request.
>
>
>
>> +                const float3_X momentum2 = momentum * momentum;
>+
>+                for( int i = 0; i < DIM3; i++ )
>+                {
>+                    nvidia::atomicAdd( &this->meanValue[i], weighting
>* momentum[i] );
>+                    nvidia::atomicAdd( &this->meanSquaredValue[i],
>weighting * momentum2[i] );
>+                }
>+            }
>+        }
>+
>+
>+        /** finalize mean value calculation */
>+        HDINLINE
>+        void finalizeMeanValues()
>+        {
>+            this->meanValue /= this->numRealParticles;
>
>it is safe in this case because `finalizeMeanValues` is serially
>executed.
>
>-- 
>You are receiving this because your review was requested.
>https://github.com/ComputationalRadiationPhysics/picongpu/pull/1959#discussion_r118901091

-- 

src/picongpu/include/plugins/particleMerging/ParticleMerger.kernel
ax3l: update: please don't do this, the namespace `PMacc` us used by default in PIConGPU

Heikman:
it is executed by just one thread and furthermore the calls are separated by a `__synchronize()`.

ax3l:
@Heikman can you please address the two to-do's in the pull request description?

src/picongpu/include/simulation_defines/param/particleMerging.param
```diff
@@ -0,0 +1,34 @@
...
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
```
ax3l: add
```C++
/* [...]
 * You should have received a copy of the GNU General Public License
 * along with PIConGPU.
 */

/* @file
 *
 * [What is this file used for in a few sentences for a user.]
 */
#pragma once
```

src/picongpu/include/plugins/particleMerging/VoronoiCell.hpp
Heikman: done

src/picongpu/include/plugins/particleMerging/VoronoiCell.hpp
Heikman: done

src/picongpu/include/simulation_defines/param/particleMerging.param
Heikman: done

src/picongpu/include/plugins/particleMerging/VoronoiCell.hpp
```diff
@@ -0,0 +1,255 @@
...
+{
+namespace plugins
+{
+namespace particleMerging
+{
+
+    /** Status of a Voronoi cell */
+    enum struct VoronoiStatus : uint8_t
+    {
+        collecting,     /* !< a Voronoi cell is collecting particles (first state) */
```
psychocoderHPC: please move the documentation above the key's

Heikman:
this PR ist not perfect, but even if I implement these last remarks there will always be something to make better. So I strongly recommend to finish this one now today, because it is already documentated in a good (but not perfect) amount.

sbastrakov:
Not sure if anyone remembers this, but asking just in case.

The component of position or momentum in which to perform a subdivision is selected slightly differently from the paper: in the paper it is one of the six with the largest variance (using normalizations from the paper). But the plugin always checks the three positions first and selects one with max variance among it in case it is above the threshold; only in the latter case, the momentums are checked in a similar way. While I do not think it leads to wrong behavior, this is still somewhat confusing and does not look documented.

ax3l:
Can you add this behaviour to the plugin documentation, please?

sbastrakov:
@ax3l yes, good point, as well as the current implementation is not fully parallel and so is (or at least potentially is) rather time-consuming.

PR #1963, opened 2017-04-12
Fix #1953 RNG Shutdown via DataConnector

### LINKED ISSUE #1953: Double Free for Ionization (RNG-using) Simulations
While investigating a PIConGPU crash on taurus using ADK as ionization method I stumbled upon a  segmentation fault at the end of the **default** Laser Wakefield example using the `dev` version. 
In the default setup **with ions and ADK** it finishes the simulation but runs into a segmentation fault right at the end. With the *kernel blocking* option, the segmentation fault happens right at start.

I am currently investigating the cause for this crash with `cuda-gdb`.

Could any one of you (@ax3l, @psychocoderHPC) verify this bug please? 
(Not that is is just a bad module combination I use on hypnos and taurus)

**Update:** (2017-04-07)
It turns out there are two issues:
 - If using `sm_20` instead of `sm_35`, both `cuda_memtest` and `picongpu` cause error. Switching to sm_35 solves this issue (this is now moved to issue #1954)
 - If using atomic Hydrogen, we still see an **out of memory** error even for very small simulation volumes. 

Thus I will rename the topic of this issue to only cover ionization. Please see #1954 for the `sm_20` vs `sm_35` issue. 

**Update:** (2017-04-12)
Modules used: 
```
  1) gcc/4.9.2
  2) cmake/3.3.0  
  3) boost/1.60.0   
  4) cuda/8.0  
  5) openmpi/1.8.6.kepler.cuda80   
  6) pngwriter/0.5.6
  7) hdf5-parallel/1.8.15
```
and own `libSplash` (the current master) at `4aa0c039f98295aa75a490ed4fc4df93ae3c9dac`.

ax3l:
Our RNGFactory should be shared with the `DataConnector` within `MySimulation` and should not share itself in its constructor.

This fixes #1953.

I will also change this behavior for `MallocMCBuffer` in a follow-up, although we have not issues from it.

### DIFF
--- a/src/libPMacc/include/random/RNGProvider.tpp
+++ b/src/libPMacc/include/random/RNGProvider.tpp
@@ -23,7 +23,7 @@
 
 #include "random/RNGProvider.hpp"
 #include "dimensions/DataSpaceOperations.hpp"
-#include "dataManagement/DataConnector.hpp"
+#include "Environment.hpp"
 
 #include <memory>
 
@@ -60,8 +60,6 @@ namespace random
     {
         if(m_size.productOfComponents() == 0)
             throw std::invalid_argument("Cannot create RNGProvider with zero size");
-
-        Environment<T_dim>::get().DataConnector().share( std::shared_ptr< ISimulationData >( this ) );
     }
 
     template<uint32_t T_dim, class T_RNGMethod>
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
Do you performed a runtime test?

ax3l:
yep, fixed the seen issue.

ax3l:
cc @n01r

PR #1966, opened 2017-04-13
mention current solver `EmZ` and compile tests

psychocoderHPC:
- `species.param`: add short docstring to mention the current solver `EmZ`
- LaserWakefield: compile test for 2D/3D EmZ current solver

I choose the laser wake field example for compile test of the current solver `EmZ` because:
  - we still test VillaBune and Esirkepov there
  - there is only one defined species (reduce compile time)


examples/LaserWakefield/cmakeFlags
```diff
@@ -40,6 +40,8 @@ flags[7]="-DCUDA_ARCH=20 -DPARAM_OVERWRITES:LIST=-DPARAM_CURRENTSOLVER=Esirkepov
 flags[8]="-DCUDA_ARCH=20 -DPARAM_OVERWRITES:LIST=-DPARAM_PRECISION=precision64Bit;-DPARAM_DIMENSION=DIM2"
 flags[9]="-DCUDA_ARCH=35 -DPARAM_OVERWRITES:LIST=-DENABLE_CURRENT=0;-DPARAM_DIMENSION=DIM2"
 flags[10]="-DCUDA_ARCH=35 -DPARAM_OVERWRITES:LIST=-DPARAM_IONS=1;-DPARAM_IONIZATION=1"
+flags[11]="-DCUDA_ARCH=20 -DPARAM_OVERWRITES:LIST=-DPARAM_CURRENTSOLVER=EmZ"
```
ax3l: is it ok if we add those new test to the KHI example instead?

ax3l: hm, ok (description comment)

PR #1971, opened 2017-04-18
Workflow: Rename tools to set up a sim

### LINKED ISSUE #1926: Default Tools: Rename, Move & PATH
@psychocoderHPC In order to streamline our install and usage process, I would like to move `configure`, `compile` and `createParameterSet` alongside `tools/bin`, keep a soft-link in `/` with the old namings if you like and make them similar to `tbg` available by default in a user's `PATH` in the `.profile`.

Maybe also name them
```
configureParameterSet
compileParameterSet
```

because otherwise they will collide (specifically `configure`), which caused trouble last time I checked. With that, we can also prepare bash completion scripts for those three commands and get rid of the bash madness with `$PICSRC/...` and TAB.

Is that ok for you?

ax3l:
Close #1926

Rename:
- compile -> pic-compile
- configure -> pic-configure
- createParameterSet -> pic-create

remove root-path `CMakeLists.txt` wrapper and motivate to extent `PATH` with `PICSRC` to have those tools available.

Also adjusts manual/docs and compile suite.


src/picongpu/submit/pizdaint-cscs/picongpu.profile.example
```diff
@@ -47,7 +47,7 @@ export CC=`which cc`
 export CXX=`which CC`
 
 export PICSRC=$HOME/src/picongpu
-# needed to call `tbg`
+export PATH=$PATH:$PICSRC
 export PATH=$PATH:$PICSRC/src/tools/bin
```
psychocoderHPC: This path is now invalid: Should be `export PATH=$PATH:$PICSRC/../tools/bin`

psychocoderHPC: ok not changed, my fault

PR #1818, opened 2017-02-10
add Bremsstrahlung example

Heikman:
This is a simulation of a flat solid density target hit head-on by a high-intensity laser pulse.  

In this process Bremsstrahlung is emitted via creation of macrophotons.

Macrophotons are travelling in every direction, yet they have a higher energy in laser direction.

PR #1979, opened 2017-04-19
Reduce # photons in Bremsstrahlung example

### LINKED ISSUE #1961: Out-of-memory in Bremsstrahlung example
I recently tried out the vanilla `Bremsstrahlung` example and noticed that `mallocMC` reports that it is out of memory.

The default in `cmakeFlags` in the example is 
```cmake
flags[0]="-DCUDA_ARCH=20"
```

I first used 8 k20 GPUs and the simulation was out of memory after 3000 time steps. By the same time the issues #1953 and #1954 came up. I double checked by compiling with 
```cmake
flags[0]="-DCUDA_ARCH=35"
```
and running on 8 k80 GPUs.

The simulation ran out of memory again and `stderr` is showing messages like:
```
[1,1]<stderr>:vsetenv LD_LIBRARY_PATH failed
...
[1,4]<stderr>:Send/receive buffer for species e is to small (max: 13797, direction: 2, retries: 1)
...
[1,3]<stderr>:  what():  [CUDA] Error: an illegal memory access was encountered[1,3]<stderr>:
[1,3]<stderr>:[kepler020:31195] *** Process received signal ***
[1,3]<stderr>:[kepler020:31195] Signal: Aborted (6)
[1,3]<stderr>:[kepler020:31195] Signal code:  (-6)
```
and long error messages pointing to memory exchanges.

Hopefully this is connected and can all be solved with #1960 .

**update** I forgot to post the module environment so here it is:
```
gcc/4.9.2
cmake/3.7.2
cuda/8.0
openmpi/1.8.6.kepler.cuda80
boost/1.60.0
pngwriter/0.5.6
hdf5-parallel/1.8.15
libsplash/1.6.0
numactl/2.0.7
paraview/3.98.laser
apr/1.5.1
apr-util/1.5.3
serf/1.3.6
svn/1.8.8
gnuplot/4.2.6
```

n01r:
In the `Bremsstrahlung` example we have typically 100 macro electrons per cell.
Initially the photon weighting was 100 times less than the electron weighting.
This would lead to good photon statistics but created a memory overflow.
Therefore the photon weighting was increased 20 times.
Now the default example runs on k20 in 21 minutes and does not use too much
memory.

The influence of the parameter `WEIGHTING_RATIO` from `bremsstrahlung.param` is two-fold:   
From `Bremsstrahlung.tpp`:   
directly proportional to the **photon emission probability**
```
const float_X emissionProb = photon::WEIGHTING_RATIO * scalingFactor * ionDensity * c * DELTA_T * this->scaledSpectrumFunctor(Ekin, kappa);
```   
inversely proportional to the spawned **photons' weighting**
```
if(this->randomGen() < emissionProb)
{
    const float_X photonEnergy = kappa * Ekin;
    this->photonMom = mom_norm * weighting / photon::WEIGHTING_RATIO * photonEnergy / c;
    return 1;
}
```

**update 2017-04-20**
Merging this closes #1961.

- [x] add pymodule(s)
- [ ] depends on #2060

### DIFF
--- a/examples/Bremsstrahlung/bin/plot_energy_histogram.py
+++ b/examples/Bremsstrahlung/bin/plot_energy_histogram.py
@@ -1,0 +1,67 @@
+"""
+This file is part of PIConGPU.
+
+PIConGPU is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+PIConGPU is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with PIConGPU.
+If not, see <http://www.gnu.org/licenses/>.
+
+Example script for analyzing data from the *energyHistogram* plugin.
+If the file is just executed in place it will plot the photon energy
+histogram of the default Bremsstrahlung example.
+There will be 5 datasets for the 5 different output iterations. The plots
+will also not contain the outliers.
+
+Copyright 2017 Marco Garten, Axel Huebl
+Authors: Axel Huebl
+License: GPLv3+
+"""
+import matplotlib as mpl
+from energy_histogram import EnergyHistogram
+
+
+# custom matplotlib settings
+params = {
+    'font.size': 20,
+    'xtick.labelsize': 20,
+    'ytick.labelsize': 20,
+    'figure.figsize': [12, 8],
+    'legend.fontsize': 20,
+    'legend.frameon': "False",
+    'legend.markerscale': 2,
+    'legend.numpoints': 1,
+    'lines.linestyle': '',
+    'lines.marker': "+",
+    'lines.markeredgewidth': 2,
+    'lines.markersize': 6
+}
+
+# overwrite matplotlib defaults
+mpl.rcParams.update(params)
+
+
+# simulation time steps
+times = np.linspace(1e3, 5e3, 5)
+hist = EnergyHistogram(
+    species_name="ph",
+    sim="../",
+    period=1000,
+    bin_count=1024,
+    min_energy=10,
+    max_energy=10000,
+    distance_to_detector=0,
+    slit_detector_x=None,
+    slit_detector_z=None
+)
+
+for i in times:
+    hist.plot(time=int(i))
--- a/examples/Bremsstrahlung/bin/plot_particle_calorimeter.py
+++ b/examples/Bremsstrahlung/bin/plot_particle_calorimeter.py
@@ -1,0 +1,73 @@
+"""
+This file is part of PIConGPU.
+
+PIConGPU is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+PIConGPU is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with PIConGPU.
+If not, see <http://www.gnu.org/licenses/>.
+
+Example script for analyzing data from the *particleCalorimeter* plugin.
+If the file is just executed in place it will plot the photon energy
+histogram of the default Bremsstrahlung example.
+There will be 5 datasets for the 5 different output iterations. The plots
+will also not contain the outliers.
+
+Copyright 2017 Marco Garten, Axel Huebl
+Authors: Axel Huebl
+License: GPLv3+
+"""
+import matplotlib as mpl
+from particle_calorimeter import ParticleCalorimeter
+
+
+# custom matplotlib settings
+params = {
+    'font.size': 20,
+    'xtick.labelsize': 20,
+    'ytick.labelsize': 20,
+    'figure.figsize': [12, 8],
+    'legend.fontsize': 20,
+    'legend.frameon': "False",
+    'legend.markerscale': 2,
+    'legend.numpoints': 1,
+    'lines.linestyle': '',
+    'lines.marker': "+",
+    'lines.markeredgewidth': 2,
+    'lines.markersize': 6
+}
+
+# overwrite matplotlib defaults
+mpl.rcParams.update(params)
+
+times = np.linspace(1e3, 5e3, 5)
+PC = ParticleCalorimeter(
+    species_name="ph",
+    sim="../",
+    period=1000,
+    num_bins_yaw=64,
+    num_bins_pitch=64,
+    num_bins_energy=1024,
+    min_energy=10,
+    max_energy=10000,
+    logscale=False,
+    opening_yaw=360,
+    opening_pitch=180,
+    pos_yaw=0,
+    pos_pitch=0
+)
+
+for i in times:
+    PC.load_step(step=i)
+    PC.energy_histogram()
+    PC.plot_histogram(norm=False)
+
+PC.plot_calorimeter()
--- a/examples/Bremsstrahlung/include/simulation_defines/param/bremsstrahlung.param
+++ b/examples/Bremsstrahlung/include/simulation_defines/param/bremsstrahlung.param
@@ -124,7 +124,7 @@ namespace photon
      *
      * The emission probability is proportional to this parameter.
      */
-    constexpr float_64 WEIGHTING_RATIO = 100;
+    constexpr float_64 WEIGHTING_RATIO = 5;
 } // namespace photon
 
 } // namespace bremsstrahlung
--- a/examples/Bremsstrahlung/submit/0008gpus.cfg
+++ b/examples/Bremsstrahlung/submit/0008gpus.cfg
@@ -46,11 +46,17 @@ TBG_periodic="--periodic 0 0 0"
 ## Section: Optional Variables ##
 #################################
 
+TBG_ph_calorimeter="--ph_calorimeter.period 1000 --ph_calorimeter.openingYaw 360 --ph_calorimeter.openingPitch 180 \
+                    --ph_calorimeter.numBinsEnergy 1024 --ph_calorimeter.minEnergy 10 --ph_calorimeter.maxEnergy 10000"
+
+TBG_ph_energyHistogram="--ph_energyHistogram.period 1000  --ph_energyHistogram.minEnergy 10 --ph_energyHistogram.maxEnergy 10000"
 
 TBG_plugins="--hdf5.period 1000  \
              --e_macroParticlesCount.period 1000 \
              --i_macroParticlesCount.period 1000 \
-             --ph_macroParticlesCount.period 1000"
+             --ph_macroParticlesCount.period 1000 \
+             !TBG_ph_calorimeter \
+             !TBG_ph_energyHistogram"
 
 
 #################################


n01r:
@Heikman can you verify that reducing the sampling quality of the photon statistics by a factor of 20 one can still see all the features that you intended to show in your example?

n01r:
I just saw that the `particleCalorimeter` plugin is not enabled by default. But that is what we also want to see from this example, right? The angular emission spectrum!

ax3l:
can you add it and compare to his original PR?

n01r:
@ax3l I agree that the plugins for seeing that output should be activated by default.
It was left off a bit abruptly, [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1818#discussion_r102182929). 

To compare it directly is a bit difficult since the #1504 PR had a 100MeV electron bunch impinging on a target foil (not an `a0 = 40` - laser) while the #1818 PR did not test for `particleCalorimeter` and `energyHistogram`. 
I would therefore just test before and after this PR with both plugins activated. 
@Heikman, your comments on this are appreciated, concerning useful parameters for those two plugins in your example.

ax3l:
makes sense, yes please feel free to figure out useful diagnostics and add them at your own discretion.

n01r:
Okay, so I checked the plugins `particleCalorimeter` and `energyHistogram`. I set up the latter with mostly default parameters and compared if there are any major changes before and after the change of the `WEIGHTING_RATIO` in this PR.
```
TBG_ph_energyHistogram="--ph_energyHistogram.period 1000  \
                        --ph_energyHistogram.minEnergy 10 \
                        --ph_energyHistogram.maxEnergy 10000"
```
**plot for \#timestep = 5000**
energyhistogram_old_new


* limited detected energies to a region between **10 ... 10000 keV**
* used the default of **1024 linearly spaced energy bins**
* used the default of detection in **full solid angle**
Since the energy intervals are linearly spaced, plotting them in logarithmic scaling will compress the data points for larger energies. One can see that for higher energies of which there are less photons the statistics seem to become worse but only by a negligible degree since the overall shape of the curve is still the same.   

Then I had a look at the `particleCalorimeter` output:
```
TBG_ph_calorimeter="--ph_calorimeter.period 1000 \
                    --ph_calorimeter.openingYaw 360 \
                    --ph_calorimeter.openingPitch 180 \
                    --ph_calorimeter.numBinsEnergy 1024 \
                    --ph_calorimeter.minEnergy 10 \
                    --ph_calorimeter.maxEnergy 10000"
```
particlecalorimeter_new
* number of **bins in pitch angle = 64** (default)
* number of **bins in yaw angle = 64** (default)
Since the angular spectrum is integrated over the energy, there should be no visible change between the old and the new spectrum (before/after PR). I checked that and only plotted the new spectrum here. The features we see (angles in ° w.r.t laser propagation direction) are:
* much energy deposited under 0°   
*Bremsstrahlung by electrons directly accelerated by the laser*
* considerable amounts of energy deposited under 180°   
*Bremsstrahlung by electrons returning at the rear of the foil*
* energy also deposited under +/- 90°   
*motion of electrons in polarization direction of laser (linearly in x) in matter*

The opening pitch and yaw detection angle options are the default values, so they could be neglected. I used the same number of energy bins so that instead of summing over the energies I could sum over the full solid angle to get an energy histogram as well.
energyhistogram_particlecalorimeter_new
Accounting to the Wiki the first bin accounts for all particles with energies less than the lower detection boundary.   

**In general the features seem to be reproduced correctly!** 
I also tried using the 2nd and 2nd last bin of `energyHistogram` as well and the relative heights are the same. For the energy histogram from the `particleCalorimeter` I also divided the energy deposited inside the bin by the center of each bin interval, respectively in order to get the actual *counts*. There is a missing factor in orders of magnitude because the data of `particleCalorimeter` is still in units of `simUnit` and not in `keV`.

n01r:
One other thing, @Heikman , the HDF5 output contains the datasets for `particles/<species>_momentumPrev1`. That should only be required if you activate the radiation plugin, right? Can this output be removed to use less memory?

n01r:
**Offline**: pymodules should be copied by `submitAction.sh` to the `<simDir>/picongpu/` directory because this is where it lies inside the example as well.

ax3l:
I'll only port the reduced number of created macro-photons to `0.3.0` to make the new example more useful, the python modules can go in the following `0.4.0/1.0.0` release.

n01r:
Sorry for the delay - had the copying of the pymodules already working.
I just wanted to make sure that the energy histograms of the `energyHistogram` class and the `particleCalorimeter` class are really comparable. Now you can choose to normalize them to total bin count and you get the choice of including the outlier bins as well.

examples/Bremsstrahlung/pymodules/energyHistogram.py
```diff
@@ -0,0 +1,227 @@
...
+        # upper range of each bin in keV
+        # note: only reads first row and selects the 1024 bins (default, see
+        # binCount) in columns 2:1026
+        self.bins = np.loadtxt(
+            self.histFile,
+            comments='',
+            usecols=range(
+                2,
+                2 + self.bin_count)
+                )[0, :]
```
Heikman: fix indentation

examples/Bremsstrahlung/pymodules/energyHistogram.py
```diff
@@ -0,0 +1,227 @@
...
+            sim="./",
+            period=1000,
+            bin_count=1024,
+            min_energy=10,
+            max_energy=10000,
+            distance_to_detector=0,
+            slit_detector_x=None,
+            slit_detector_z=None):
+        """
+        species_name
```
Heikman: please use `.rst` format for the docstring, i.e.:
```
:param species_name: name of the species for which the energy histogram should be plotted, e.g. "ph"
```

examples/Bremsstrahlung/pymodules/energyHistogram.py
```diff
@@ -0,0 +1,227 @@
...
+        self.period = period
+        self.bin_count = bin_count
+        self.min_energy = min_energy
+        self.max_energy = max_energy
+        self.distance_to_detector = distance_to_detector
+        self.slit_detector_x = slit_detector_x
+        self.slit_detector_z = slit_detector_z
+
+        self.histFile = "{}/simOutput/{}_energyHistogram.dat".format(
+            sim, species_name)
```
Heikman: indentation

examples/Bremsstrahlung/pymodules/energyHistogram.py
```diff
@@ -0,0 +1,227 @@
...
+        # cumulates all counts of photons that have energies above
+        # ``max_energy``
+        self.overflow_bins = np.loadtxt(self.histFile)[:, -2]
+        # simulation time steps for output
+        self.time_steps = np.loadtxt(self.histFile)[:, 0]
+        # cumulative sum of all counts in all bins for each output time step
+        self.total_counts = np.loadtxt(self.histFile)[:, -1]
+
+    def plot(self,
+             time=5000, with_outliers=False, norm=False):
```
Heikman: indentation

examples/Bremsstrahlung/pymodules/energyHistogram.py
```diff
@@ -0,0 +1,227 @@
...
+                self.bins,
+                self.histData[iteration] / self.total_counts[iteration],
+                label=time
+            )
+        else:
+            # plot the histogram without normalization
+            EHist = subEHist.plot(
+                self.bins,
+                self.histData[iteration],
+                label=time)
```
Heikman: indentation

examples/Bremsstrahlung/pymodules/energyHistogram.py
```diff
@@ -0,0 +1,227 @@
...
+
+        subEHist.set_xscale("log")
+        subEHist.set_yscale("log")
+        subEHist.set_xlabel("photon energy [keV]")
+        subEHist.set_ylabel("photon count")
+        subEHist.set_title("`{}` energy histogram".format(self.species_name))
+
+        if (with_outliers):
```
Heikman: remove brackets

examples/Bremsstrahlung/pymodules/energyHistogram.py
```diff
@@ -0,0 +1,227 @@
...
+
+            if (norm):
```
Heikman: remove brackets

examples/Bremsstrahlung/pymodules/particleCalorimeter.py
```diff
@@ -0,0 +1,283 @@
...
+        """
+        Load an HDF5 File of a certain time step
+
+        time
+            PIConGPU output time step
+        """
+        # convert the time step into a string
+        self.stepStr = str(int(step))
+
+        self.calo_file = self.sim + "/simOutput/" \
```
Heikman: please put these lines into round brackets in order to avoid the backslashes, i.e.:
```
self.calo_file = (
  self.sim + "/simOutput/" +
  self.species_name + "_" +
  self.attribute_name + "/"
)
```

examples/Bremsstrahlung/pymodules/energyHistogram.py
```diff
@@ -0,0 +1,227 @@
...
+            )[0], ".."
+        )
+    )
+)
+if cmd_subfolder not in sys.path:
+    sys.path.insert(0, cmd_subfolder)
+
+
+# custom matplotlib settings
+params = {
```
ax3l: this should not be done in a module.
if someone changes mpl rc params, it's user/app code and not a module

examples/Bremsstrahlung/pymodules/__init__.py
```diff
@@ -0,0 +1,3 @@
+__doc__ = None
+
+__all__ = ["energyHistogram","particleCalorimeter"]
```
ax3l: pep8: space after `,`

examples/Bremsstrahlung/pymodules/energyHistogram.py
```diff
@@ -0,0 +1,227 @@
...
+import os
+import inspect
+
+import numpy as np
+import matplotlib as mpl
+from matplotlib import pyplot as plt
+import h5py
+
+# prepare imports
+cmd_subfolder = os.path.realpath(
```
ax3l: yoo don't need this

examples/Bremsstrahlung/pymodules/energyHistogram.py
```diff
@@ -0,0 +1,227 @@
+import sys
```
ax3l: you don't need those

ax3l: - [ ] missing copyright info and completely taken from my wiki example
- [ ] generic module and not specific to this example, put it in `src/tools/share/python/` instead

examples/Bremsstrahlung/pymodules/particleCalorimeter.py
```diff
@@ -0,0 +1,283 @@
+import sys
```
ax3l: absolutely the same as in the other module

examples/Bremsstrahlung/pymodules/energyHistogram.py
ax3l: I am ok with both classical PEP 257 and PEP 287 for rst.

But you are right, this one ain’t either one nor the other :D

examples/Bremsstrahlung/pymodules/energyHistogram.py
ax3l: that's ok, valid pep8

examples/Bremsstrahlung/pymodules/energyHistogram.py
ax3l: valid pep8

ax3l:
the new scripts won't work without the modules in #2060 but can be merged into `dev` for now

PR #1980, opened 2017-04-19
add peridoic boundary conditions to the bremsstrahlungs example

PrometheusPi:
The `Bremsstrahlung` example was missing periodic boundary conditions. 

This leads to false results on the borders. 

@ax3l and @Heikman Do you think this is necessary?

### DIFF
--- a/examples/Bremsstrahlung/submit/0008gpus.cfg
+++ b/examples/Bremsstrahlung/submit/0008gpus.cfg
@@ -40,7 +40,7 @@ TBG_gpu_z=1
 TBG_gridSize="-g 2048 2048 1"
 TBG_steps="-s 5000"
 
-TBG_periodic="--periodic 0 0 0"
+TBG_periodic="--periodic 1 0 1"
 
 #################################
 ## Section: Optional Variables ##


n01r:
Right, forgot that ^^;

But yes, I can see that the simulation loses particles on the boundaries.

ax3l:
Why periodic boundary conditions, it's simulating a Gaussian beam on a flat foil... o.0

PrometheusPi:
@ax3l I included these boundary conditions because without periodoc boundaries one loses particles at the edges - which is nonphysical. This has **nothing to do with the laser**. You those limit the transversal extent of your target to 4.5 FWHM spot sizes.

PrometheusPi:
@ax3l Thanks for editing your reply - I agree thermal boundary conditions would be better. Do you think **absorbing** are better than **periodic** in this case?

ax3l:
yes, absorbing (as it is in `dev`) is the right choice.

ax3l:
periodic makes no sense, we don't shoot an array of lasers on a foil ;)

PrometheusPi:
I trust @ax3l expertise on foils and thus will close this pull request.

PrometheusPi:
@ax3l Yes I agree that we do not shoot an array of lasers, but we neither have a `T=0 K` boundary either were no particles are coming from but only particles diffuse to.

ax3l:
yes that's a different story and the motivation for thermal boundary conditions to compensate the flux going outward. With absorbing you just try to damp all effects from the boundaries and try to get as less influence from it as possible, which of course neglects return currents.
Nevertheless, feeding the same e- again to the laser with periodic boundaries would dramatically overestimate the e- temperature and can not be motivated with symmetry arguments either.

Issue #1982, opened 2017-04-20
Param Files: Contribute Conceptiual Docs


ax3l:
Dear @ComputationalRadiationPhysics/picls-developer 's

please contribute some documentation strings describing the task of individual `*.param` files.
The idea of this doxygen string is to both:
- explain the user what the current open file is used for **conceptually** (not individual options: for those in-line comments are used)
- automatically fills the **user documentation** with the same information in our **manual** (note: this output will look better as soon as https://github.com/michaeljones/breathe/issues/318 is implemented)

For a good and on-the-point example, see `grid.param`
- [in code](https://github.com/ComputationalRadiationPhysics/picongpu/blob/94acb88ff1e5e5fb6eb1d753daedbf3ccb9cc43a/src/picongpu/include/simulation_defines/param/grid.param#L19-L46) (note the order of copyright, file docu, `pragma once`/includes and number of empty newlines!)
- [in manual](https://picongpu.readthedocs.io/en/dev/usage/param/core.html#grid-param)

Here is a small suggestion, feel free to contribute PRs:

- [ ] bremsstrahlung.param @Heikman 
- [ ] components.param @ax3l
- [x] density.param @ax3l
- [x] dimension.param @ax3l
- [ ] fieldBackground.param @ax3l
- [ ] fieldSolver.param @ax3l
- [ ] fileOutput.param @ax3l
- [x] grid.param @ax3l
- [ ] ionizationEnergies.param @n01r 
- [ ] ionizer.param @n01r 
- [ ] isaac.param @theZiz 
- [ ] laser.param @ax3l
- [ ] mallocMC.param @psychocoderHPC 
- [ ] memory.param @psychocoderHPC 
- [ ] particleCalorimeter.param @Heikman 
- [x] particle.param @ax3l
- [ ] physicalConstants.param @ax3l
- [ ] precision.param @psychocoderHPC 
- [x] pusher.param @ax3l
- [ ] radiationObserver.param @PrometheusPi 
- [ ] radiation.param @PrometheusPi 
- [ ] seed.param @ax3l
- [x] speciesAttributes.param @ax3l
- [x] speciesConstants.param @ax3l
- [x] speciesDefinition.param @ax3l
- [ ] speciesInitialization.param @ax3l
- [x] species.param @ax3l
- [ ] starter.param @psychocoderHPC 
- [ ] synchrotronPhotons.param @Heikman 
- [ ] visColorScales.param @ax3l
[...]

PR #1983, opened 2017-04-20
Document ionization param files for issue 1982

n01r:
In order to contribute to the closing of issue #1982 this pull request adds general file documentation for the `.param` files connected to ionization.


ax3l:
really great and quick, thank you @n01r !

n01r:
I think that for the models that are implemented, we are actually fine. Maybe I'll remove that `todo` section.

ax3l:
sounds about right to me, sure!

PR #1662, opened 2016-11-29
use PMACC_ASSERT and PMACC_VERIFY

psychocoderHPC:
use PMacc internal assert and verification macros

Both calls are available since #1578

PR #1988, opened 2017-04-21
option for CMake build type

psychocoderHPC:
The issue with the old implementation is that the variable `CMAKE_BUILD_TYPE` was always empty and therefore we build everything as `debug`. This should be avoided and is since #1662 not necessary.

- PMacc
  - change default behavior from build in debug mode to build as release
  - allow user to switch `CMAKE_BUILD_TYPE` between `Release` and `Debug`
- PIConGPU
  - remove CMake option `PIC_RELEASE`

This changes the build type default to `Release`.

## Tests

- [x] compile debug
- [x] compile release

@ax3l Should we port this the the `0.3.0-rc`?

### DIFF
--- a/src/libPMacc/PMaccConfig.cmake
+++ b/src/libPMacc/PMaccConfig.cmake
@@ -46,6 +46,18 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
 
 
 ###############################################################################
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
I pushed a second release to remove the variable `PIC_RELEASE`.

PR #2007, opened 2017-04-28
ThomasFermi: Fix Domain Decomposition

ax3l:
Adds communication to the `FieldTmp` slots to cure an unholy domain-decomposition bug in Thomas-Fermi ionization. (Density and Energy in the a local GPU's border has contributions from the next GPU.) The full example shown below is given in #2008 but applies for all cases where Thomas-Fermi ionization is used.

**BSI+ADK+TF_dev**
tf_unholy

**BSI+ADK**
bsi_adk

Unfortunately, there seems an other bug (maybe wrongly shifted or some caching issue?). But at least it's already less severe:

**BSI+ADK+TF_PR**
tf_still_unholy

### To Do

- [x] what's the 2nd issue?
- [x] depends on #2010 
- [x] ~~and does it affect the field ionization methods, too?~~

### DIFF
--- a/src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
+++ b/src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
@@ -75,7 +75,11 @@ namespace ionization
 
         using FrameType =  typename SrcSpecies::FrameType;
 
-        /* specify field to particle interpolation scheme */
+        /** specify field to particle interpolation scheme
+         *
+         * @todo this needs to be done independently/twice if ion species (rho) and electron
+         *       species (ene) are of different shape
+         */
         using Field2ParticleInterpolation = typename PMacc::traits::Resolve<
             typename GetFlagType<FrameType,interpolation<> >::type
         >::type;
@@ -166,6 +170,8 @@ namespace ionization
                 /* kernel call for weighted ion density calculation */
                 density->template computeValue< CORE + BORDER, DensitySolver >(*srcSpecies, currentStep);
                 dc.releaseData( SrcSpecies::FrameType::getName() );
+                EventTask densityEvent = density->asyncCommunication( __getTransactionEvent() );
+                densityEvent += density->asyncCommunicationGather( densityEvent );
 
                 /* load species without copying the particle data to the host */
                 auto destSpecies = dc.get< DestSpecies >( DestSpecies::FrameType::getName(), true );
@@ -173,6 +179,11 @@ namespace ionization
                 /* kernel call for weighted electron energy density calculation */
                 eneKinDens->template computeValue< CORE + BORDER, EnergyDensitySolver >(*destSpecies, currentStep);
                 dc.releaseData( DestSpecies::FrameType::getName() );
+                EventTask eneKinEvent = eneKinDens->asyncCommunication( __getTransactionEvent() );
+                eneKinEvent += eneKinDens->asyncCommunicationGather( eneKinEvent );
+
+                /* contributions from neighboring GPUs to our border area */
+                __setTransactionEvent( densityEvent + eneKinEvent );
 
                 /* initialize device-side density- and energy density field databox pointers */
                 rhoBox = density->getDeviceDataBox();


psychocoderHPC:
OK I get it. This structure is coming from the effect that fieldTmp can not receive fields to the guard to build a local volume without borders. It is only possible to send fields from the local guard to the border of the next GPU.

ax3l:
Ah sure, yes. totally forgot about that. The border area is correct now but the guard is still not updated correctly with the next GPUs border...

ax3l:
better but not yet perfect. with #2010 (first draft) we are now away from the [unholy target](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2007#issue-225126134) and are straight on to the highway to hell:

highway

ax3l:
@psychocoderHPC and I just found we messed up a manual event dependency in our local changes creating a race condition during energy density creation for the last image. Pushing fixed update to this PR now.

tf_fixed

ax3l:
@psychocoderHPC @n01r done (pls add your review)

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
```diff
@@ -75,7 +75,11 @@ namespace ionization
 
         using FrameType =  typename SrcSpecies::FrameType;
 
-        /* specify field to particle interpolation scheme */
+        /** specify field to particle interpolation scheme
+         *
+         * @todo this needs to be done independently/twice if ion species (rho) and electron
+         *       species (ene) are of different shape
```
n01r: In principle one could add another **todo** here
If one were to use two different atomic species they would currently each do their own density and energy calculations. Yet they would have no knowledge of each other. But of course IPD effects do not care about the species and so it remains a technical detail to be solved on how to include that. :smiley:

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
```diff
@@ -166,13 +170,20 @@ namespace ionization
                 /* kernel call for weighted ion density calculation */
                 density->template computeValue< CORE + BORDER, DensitySolver >(*srcSpecies, currentStep);
                 dc.releaseData( SrcSpecies::FrameType::getName() );
+                EventTask densityEvent = density->asyncCommunication( __getTransactionEvent() );
+                densityEvent += density->asyncCommunicationGather( densityEvent );
```
n01r: Just for me again: the first of the two lines above gets the event of the `FieldTmp` operation and the second line says that in addition to this event the results from the previous event in the `BORDER` region of the neigboring GPU have ton be communicated to the `GUARD` of this GPU. Only afterwards the density values for particles whose shape extends outside the `BORDER` region will be correct, right?

ax3l: yes, this code had two bugs with regards to domain decomposition:
- the border was not filled correct (with additions from neighbor-GPU particles that have shape contributions) and
- since you access the density and energy density fields with respect to shape you also access the guard which was not created at all.

The first event, `asyncCommunication` builds a valid `BORDER`, the second event, `asyncCommunicationGather` builds a valid `GUARD`. Both APIs are also doxygen documented now since #2010

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
ax3l: yep, that's a todo I have in my collisional method, too xD

n01r: Sure sure, that's a follow-up feature and I have a hunch that it requires some careful thinking. :laughing:

ax3l:
@psychocoderHPC pls merge me :)

ax3l:
@n01r more of a coding-priest, spreading the word of bug fixing

PR #2009, opened 2017-05-03
`FieldTmp` gather support

psychocoderHPC:
Add an explicit method to trigger that border data from the neighbor is gathered into the local guard of the GPU.

- add method `asyncCommunicationGather()` (gather neighbor's border data)
- add buffer for gather communication
- define `fieldTmpSupportGatherCommunication` variable in `memory.param` to activate gather support for `FieldTmp`
- remove method `getCommTag()`

## Note
Currently the new method `asyncCommunicationGather()` is only available in `FieldTmp` but should be later also be ported to all other buffers like `FieldE` and `FieldB`. The best way to do is is to extent the PMacc `GridBuffer` to support `asyncCommunicationScatter()` and `asyncCommunicationGather()`.
This is a bigger refactoring and should not be part of this pull request to allow a backport to the current release candidate `0.3.0`.

### DIFF
--- a/examples/Bunch/include/simulation_defines/param/memory.param
+++ b/examples/Bunch/include/simulation_defines/param/memory.param
@@ -56,4 +56,14 @@ constexpr uint32_t BYTES_EDGES = 32 * 1024; //32 kiB;
 /** number of scalar fields that are reserved as temporary fields */
 constexpr uint32_t fieldTmpNumSlots = 1;
 
+/** can `FieldTmp` gather neighbor information
+ *
+ * If `true` it is possible to call the method `asyncCommunicationGather()`
+ * to copy data from the border of neighboring GPU into the local guard.
+ * This is also known as building up a "ghost" or "halo" region in domain
+ * decomposition and only necessary for specific algorithms that extend
+ * the basic PIC cycle, e.g. with dependence on derived density or energy fields.
+ */
+constexpr bool fieldTmpSupportGatherCommunication = true;
+
 }//namespace picongpu
--- a/src/picongpu/include/fields/FieldTmp.hpp
+++ b/src/picongpu/include/fields/FieldTmp.hpp
@@ -26,9 +26,7 @@
 #include <vector>
 
 /*pic default*/
-#include "pmacc_types.hpp"
 #include "simulation_defines.hpp"
-#include "simulation_classTypes.hpp"
 
 #include "fields/Fields.def"
 #include "fields/SimulationFieldHelper.hpp"
@@ -88,9 +86,21 @@ namespace picongpu
 
         static std::string getName();
 
-        uint32_t getCommTag();
-
+        /** scatter data to neighboring GPUs
+         *
+         * Add data from the local guard of the GPU to the border of the neighboring GPUs.
+         * This method can be called before or after asyncCommunicationGather without
+         * explicit handling to avoid race conditions between both methods.
+         */
         virtual EventTask asyncCommunication( EventTask serialEvent );
+
+        /** gather data from neighboring GPUs
+         *
+         * Copy data from the border of neighboring GPUs into the local guard.
+         * This method can be called before or after asyncCommunication without
+         * explicit handling to avoid race conditions between both methods.
+         */
+        EventTask asyncCommunicationGather( EventTask serialEvent );
 
         void init( );
 
@@ -125,10 +135,14 @@ namespace picongpu
     private:
 
         GridBuffer<ValueType, simDim> *fieldTmp;
+        GridBuffer<ValueType, simDim>* fieldTmpRecv;
 
         uint32_t m_slotId;
 
-        uint32_t m_commTag;
+        EventTask m_scatterEv;
+        uint32_t m_commTagScatter;
+        EventTask m_gatherEv;
+        uint32_t m_commTagGather;
     };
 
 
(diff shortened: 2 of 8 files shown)


src/picongpu/include/fields/FieldTmp.hpp
```diff
@@ -88,10 +86,14 @@ namespace picongpu
 
         static std::string getName();
 
-        uint32_t getCommTag();
-
         virtual EventTask asyncCommunication( EventTask serialEvent );
 
+        /** gather data from neighbor GPU
+         *
+         * copy data from the border of neighboring GPU into the local guard.
```
psychocoderHPC: Add missing information that there is no race condition if `asyncCommunicationGather` is called before or after `asyncCommunication`

examples/Bunch/include/simulation_defines/param/memory.param
```diff
@@ -56,4 +56,11 @@ constexpr uint32_t BYTES_EDGES = 32 * 1024; //32 kiB;
 /** number of scalar fields that are reserved as temporary fields */
 constexpr uint32_t fieldTmpNumSlots = 1;
 
+/** can `FieldTmp` gather neighbor information
+ *
+ * If `true` it is possible to call the method `asyncCommunicationGather()`
+ * to copy data from the border of neighboring GPU into the local guard.
+ */
```
ax3l: Add a bit more specific info: "This is also known as building up a "ghost" or "halo" region in domain decomposition and only necessary for specific algorithms that extend the basic PIC cyle, e.g. with dependence on derived density or energy fields."

src/picongpu/include/fields/FieldTmp.hpp
```diff
@@ -88,10 +86,14 @@ namespace picongpu
 
         static std::string getName();
 
-        uint32_t getCommTag();
-
         virtual EventTask asyncCommunication( EventTask serialEvent );
 
+        /** gather data from neighbor GPU
```
ax3l: neighboring GPUs

src/picongpu/include/fields/FieldTmp.tpp
```diff
@@ -165,9 +171,21 @@ namespace picongpu
...
-            // std::cout << "ex " << i << " x=" << guardingCells[0] << " y=" << guardingCells[1] << " z=" << guardingCells[2] << std::endl;
-            fieldTmp->addExchangeBuffer( i, guardingCells, m_commTag );
+
+            fieldTmp->addExchangeBuffer( i, guardingCells, m_commTagScatter );
+
+            if( fieldTmpRecv )
+            {
+                /* guarding cells depend on direction
+                 * for negative direction use originGuard else endGuard (relative direction ZERO is ignored)
+                 * don't switch end and origin because this is a read buffer and no send buffer
```
ax3l: and not a send buffer

src/picongpu/include/fields/FieldTmp.tpp
```diff
@@ -224,16 +242,26 @@ namespace picongpu
...
+        __startTransaction( serialEvent + m_gatherEv + m_scatterEv);
         FieldFactory::getInstance( ).createTaskFieldSend( *this );
         ret += __endTransaction( );
+        m_scatterEv = ret;
         return ret;
     }
 
+    EventTask FieldTmp::asyncCommunicationGather( EventTask serialEvent )
+    {
+        PMACC_VERIFY_MSG( fieldTmpSupportGatherCommunication == true, "fieldTmpSupportGatherCommunication in memory.param must be set to true" );
```
ax3l: make me a two liner

src/picongpu/include/fields/FieldTmp.tpp
```diff
@@ -224,16 +242,26 @@ namespace picongpu
...
+
+        if( fieldTmpRecv != nullptr )
+            m_gatherEv = fieldTmpRecv->asyncCommunication( m_scatterEv + m_gatherEv );
```
ax3l: self-found by @psychocoderHPC : `serialEvent` needs to be handled here

examples/Bunch/include/simulation_defines/param/memory.param
psychocoderHPC: Ahh I see you tried if I copy&past your text without reading it^^ `cyle` must be `cycle`

src/picongpu/include/simulation_defines/param/memory.param
```diff
@@ -56,4 +56,11 @@ namespace picongpu
     /** number of scalar fields that are reserved as temporary fields */
     constexpr uint32_t fieldTmpNumSlots = 1;
 
+    /** can `FieldTmp` gather neighbor information
+     *
+     * If `true` it is possible to call the method `asyncCommunicationGather()`
+     * to copy data from the border of neighboring GPU into the local guard.
+     */
```
ax3l: forgot: 
* This is also known as building up a "ghost" or "halo" region in domain
* decomposition and only necessary for specific algorithms that extend
* the basic PIC cycle, e.g. with dependence on derived density or energy fields.

src/picongpu/include/fields/FieldTmp.hpp
```diff
@@ -88,10 +86,16 @@ namespace picongpu
 
         static std::string getName();
 
-        uint32_t getCommTag();
-
         virtual EventTask asyncCommunication( EventTask serialEvent );
```
ax3l: it would make sense to add a doxygen string here now, too

PR #2013, opened 2017-05-03
Rm BSIHydrogenLike, Add BSI with charge Z

n01r:
The `BSIHydrogenLike` model was much too narrow since it was only applicable
to hydrogenlike states, i.e. states that have charge (Z-1)+.
The following replacement was done in

    F_BSI = E_i^2 / (4*Z)

    Z (nuclear charge) --> Z_i (ion residual charge)

That means that in a naive way the inner electrons shield the nuclear charge
from the electron that is to be ionized.

### DIFF
--- a/src/picongpu/include/particles/ionization/byField/BSI/AlgorithmBSI.hpp
+++ b/src/picongpu/include/particles/ionization/byField/BSI/AlgorithmBSI.hpp
@@ -63,21 +63,21 @@ namespace ionization
         operator()( const EType eField, ParticleType& parentIon )
         {
 
-            const float_X protonNumber = GetAtomicNumbers<ParticleType>::type::numberOfProtons;
-            float_X chargeState = attribute::getChargeState(parentIon);
+            float_X const protonNumber = GetAtomicNumbers<ParticleType>::type::numberOfProtons;
+            float_X const chargeState = attribute::getChargeState(parentIon);
 
             /* verify that ion is not completely ionized */
             if (chargeState < protonNumber)
             {
-                uint32_t cs = math::float2int_rd(chargeState);
+                uint32_t const cs = math::float2int_rd(chargeState);
                 /* ionization potential in atomic units */
-                const float_X iEnergy = GetIonizationEnergies<ParticleType>::type()[cs];
+                float_X const iEnergy = GetIonizationEnergies<ParticleType>::type()[cs];
                 /* the charge that attracts the electron that is to be ionized:
- +               * equals `protonNumber - #allInnerElectrons`
+                 * equals `protonNumber - no. allInnerElectrons`
                  */
- +              float_X effectiveCharge = chargeState + float_X(1.0);
+                float_X const effectiveCharge = chargeState + float_X(1.0);
                 /* critical field strength in atomic units */
-                float_X critField = iEnergy*iEnergy / (float_X(4.0) * effectiveCharge);
+                float_X const critField = iEnergy*iEnergy / (float_X(4.0) * effectiveCharge);
                 /* ionization condition */
                 if (math::abs(eField) / ATOMIC_UNIT_EFIELD >= critField)
                 {
--- a/src/picongpu/include/particles/ionization/byField/BSI/AlgorithmBSI.hpp
+++ b/src/picongpu/include/particles/ionization/byField/BSI/AlgorithmBSI.hpp
@@ -24,7 +24,7 @@
 #include "particles/traits/GetAtomicNumbers.hpp"
 #include "traits/attribute/GetChargeState.hpp"
 
-/** @file AlgorithmBSIHydrogenLike.hpp
+/** @file AlgorithmBSI.hpp
  *
  * IONIZATION ALGORITHM for the BSI model
  *
@@ -42,7 +42,7 @@ namespace ionization
 
     /** Calculation for the Barrier Suppression Ionization model
      */
-    struct AlgorithmBSIHydrogenLike
+    struct AlgorithmBSI
     {
 
         /** Functor implementation
@@ -72,8 +72,12 @@ namespace ionization
                 uint32_t cs = math::float2int_rd(chargeState);
                 /* ionization potential in atomic units */
                 const float_X iEnergy = GetIonizationEnergies<ParticleType>::type()[cs];
+                /* the charge that attracts the electron that is to be ionized:
+ +               * equals `protonNumber - #allInnerElectrons`
+                 */
+ +              float_X effectiveCharge = chargeState + float_X(1.0);
                 /* critical field strength in atomic units */
-                float_X critField = iEnergy*iEnergy / (float_X(4.0) * protonNumber);
+                float_X critField = iEnergy*iEnergy / (float_X(4.0) * effectiveCharge);
                 /* ionization condition */
                 if (math::abs(eField) / ATOMIC_UNIT_EFIELD >= critField)
                 {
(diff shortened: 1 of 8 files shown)


src/picongpu/include/particles/ionization/byField/BSI/AlgorithmBSI.hpp
```diff
@@ -72,8 +72,12 @@ namespace ionization
                 uint32_t cs = math::float2int_rd(chargeState);
                 /* ionization potential in atomic units */
                 const float_X iEnergy = GetIonizationEnergies<ParticleType>::type()[cs];
+                /* the charge that attracts the electron that is to be ionized:
+ +               * equals `protonNumber - #allInnerElectrons`
```
PrometheusPi: OPTIONAL: 
perhaps just write number instead of #

src/picongpu/include/particles/ionization/byField/BSI/BSI.def
```diff
@@ -48,9 +48,10 @@ namespace ionization
      *   "Tunneling and barrier-suppression ionization of atoms and ions in a laser radiation field"
      *   doi:10.1070/PU1998v041n05ABEH000393
      *
-     * - This model assumes a hydrogenlike ion configuration,
-     *   neglecting core shielding by other electrons and Stark upshifting
-     *   of ionization energies according to strong external electric fields
+     * - This model accounts for naive ion charge shielding by inner electrons
+     *   as it assumes that the charge the electron 'feels' is equal to
+     *   `Z - #NumberOfBoundInnerShellElectrons`.
```
PrometheusPi: REQUIRED:
you are using both # and number - please just use number

PrometheusPi:
@n01r What happens now to `BSIEffectiveZ`? Where is the difference between the new `BSI` and `BSIEffectiveZ`?

n01r:
@PrometheusPi The difference lies in how the residual ion charge is determined.   
In the simple BSI case one would naively say: *The charge the electron sees comes from the proton number reduced by the number of other bound electrons.* This reduces the residual ion to a single point and neglects all geometric effects. So when I introduced `BSIEffectiveZ` I tried to account for that by using effective charges that were calculated from shielding and considering orbital structure. 

Just for illustration (this will also find its way into the documentation).
2017_05_04_effective_potentials
You see here nuclear potentials influenced by a homogenous external electric field. The potentials are created just by varying the charge `Z`. We compare Hydrogen with Carbon-1+ (often denoted as C-II). The external electric field is the same in all three cases and is just exactly the BSI field strength that you would calculate with `BSIEffectiveZ`. It becomes apparent that even though the outermost electron in C-II lies energetically lower than the one in Hydrogen both will become **classically** ionized in the `BSIEffectiveZ` model. The simple `BSI` model sets the barrier to a larger value and thus does not allow ionization at this point. That does not make it better, though. Other effects are neglected as well in both models, e.g. Stark upshift of ionization energies which would deepen the potential well and increase the necessary field strength for ionization. Both of these models should just be used as a simple continuation from the tunneling regime to the barrier suppression regime when one increases external field strength. There are other empirical models for this purpose as well, though.

PrometheusPi:
@n01r Could you please add the changes.

src/picongpu/include/particles/ionization/byField/BSI/AlgorithmBSI.hpp
ax3l: or `no. of`

PrometheusPi: that would be fine too

n01r:
Sorry for the delay.

Changes have been implemented. Please check if you're okay with them, @ax3l and @PrometheusPi .

PrometheusPi:
@n01r thank you for the fast reply, I will have a look at the code today

ax3l:
@PrometheusPi looks god to me, if you are ok I can review and merge it

PrometheusPi:
I am fine with that. Go ahead.

src/picongpu/include/particles/ionization/byField/BSI/AlgorithmBSI.hpp
```diff
@@ -72,8 +72,12 @@ namespace ionization
                 uint32_t cs = math::float2int_rd(chargeState);
                 /* ionization potential in atomic units */
                 const float_X iEnergy = GetIonizationEnergies<ParticleType>::type()[cs];
+                /* the charge that attracts the electron that is to be ionized:
+ +               * equals `protonNumber - no. allInnerElectrons`
+                 */
+ +              float_X effectiveCharge = chargeState + float_X(1.0);
                 /* critical field strength in atomic units */
-                float_X critField = iEnergy*iEnergy / (float_X(4.0) * protonNumber);
+                float_X critField = iEnergy*iEnergy / (float_X(4.0) * effectiveCharge);
```
ax3l: this should be a `float_X const`

src/picongpu/include/particles/ionization/byField/BSI/AlgorithmBSI.hpp
```diff
@@ -72,8 +72,12 @@ namespace ionization
                 uint32_t cs = math::float2int_rd(chargeState);
                 /* ionization potential in atomic units */
                 const float_X iEnergy = GetIonizationEnergies<ParticleType>::type()[cs];
+                /* the charge that attracts the electron that is to be ionized:
+ +               * equals `protonNumber - no. allInnerElectrons`
+                 */
+ +              float_X effectiveCharge = chargeState + float_X(1.0);
```
ax3l: - this should be a `float_X const`
- there is a `+` in front of this line

src/picongpu/include/particles/ionization/byField/BSI/AlgorithmBSI.hpp
```diff
@@ -72,8 +72,12 @@ namespace ionization
                 uint32_t cs = math::float2int_rd(chargeState);
                 /* ionization potential in atomic units */
                 const float_X iEnergy = GetIonizationEnergies<ParticleType>::type()[cs];
+                /* the charge that attracts the electron that is to be ionized:
+ +               * equals `protonNumber - no. allInnerElectrons`
```
ax3l: there is a `+` in front if this line

src/picongpu/include/particles/ionization/byField/BSI/AlgorithmBSI.hpp
ax3l: there is a + in front of this line

src/picongpu/include/particles/ionization/byField/BSI/AlgorithmBSI.hpp
```diff
@@ -72,8 +72,12 @@ namespace ionization
                 uint32_t cs = math::float2int_rd(chargeState);
```
ax3l: should be `uint32_t const cs = ...`

src/picongpu/include/particles/ionization/byField/BSI/AlgorithmBSI.hpp
n01r: oO where the hell did that come from?

ax3l: copied from a diff?

n01r:
fixed the 4 lines and amended the commit

src/picongpu/include/particles/ionization/byField/BSI/AlgorithmBSI.hpp
n01r: mmh, must be it :face_with_head_bandage:

PR #2014, opened 2017-05-04
lock step kernel `KernelShiftParticles`

psychocoderHPC:
- refactor `KernelShiftParticles` (use lockstep description)
- add documentation `lockstep.rst`

# Notes for the reviewer

- each variable which was previously the state of one thread is now a `CtxArray` and got the postfix `Ctx` for *the virtual worker's context* (this avoid side effects during the refactoring and helps to read the code)
- the number of worker per kernel is chosen at compile time and is currently fixed to the size of the supercell but will depend in the future on the alpaka accelerator restrictions.
- I tried not to change the logical flow of the algorithm or add optimizations.
- the changed code is formatted with the new code style

# Lockstep programming model

The *lockstep programming model* separates code those is evaluated by worker (threads) collectively and independently.
A problem is described by index domains. A index domain is **independent** from data but **can** be mapped to a data domain one to one but also with more complex mappings.
Code which is implemented by the *lockstep programming model* is free of any dependencies between the number of worker and processed data elements.
To simplify the implementation each index within a domain can be seen as a *virtual worker* which is processing one data element (like the common workflow to programming CUDA). Each *worker i* can be executed as N_i *virtual workers* (1:N_i).

## PMacc helper

- `mappings::threads::IdxConfig<>`
  - describe the size of the index domain and the number of workers to operate on the domain 
- `memory::CtxArray<>`
  - hold states in a context for each worker
  - the number of states depends on the size of a index domain and the number of worker
- `mappings::threads::ForEachIdx<>`
  - execute a user functor for each index within the index domain

## Common Pattern

- collective loop (each worker needs to pass a loop N times)
  - in this example, there are more dates then workers that process them

```C++
    // `frame` is a list which must be traversed collectively
    while ( frame.isValid() )
    {
        const uint32_t workerIdx = threadIdx.x;
        using ParticleDomCfg = IdxConfig<
            frameSize,
            numWorker
        >;
        ForEachIdx< ParticleDomCfg > forEachParticle( workerIdx );
       forEachParticle(
           [&]( uint32_t const linearIdx, uint32_t const idx )
           {
               // independent work
           }
       );
    }
````
- non-collective loop (each *virtual worker* increments a private variable)
    ```C++
    const uint32_t workerIdx = threadIdx.x;
    using ParticleDomCfg = IdxConfig<
        frameSize,
        numWorker
    >;
    ForEachIdx< ParticleDomCfg > forEachParticle( workerIdx );
    memory::CtxArray< int, ParticleDomCfg > vWorkerIdx( 0 );
    forEachParticle(
        [&]( uint32_t const linearIdx, uint32_t const idx )
        {
            vWorkerIdx[ idx ] = linearIdx;
            for( int i = 0; i < 100; i++ )
                vWorkerIdx[ idx ]++;
        }
    );
    ```
- create a context variable and initialize with the index of the virtual worker
    ```C++
    const uint32_t workerIdx = threadIdx.x;
    using ParticleDomCfg = IdxConfig<
        frameSize,
        numWorker
    >;
    memory::CtxArray< int, ParticleDomCfg > vIdx(
        workerIdx,
        [&]( uint32_t const linearIdx, uint32_t const ) -> DataSpace<dim>
        {
            return linearIdx;
        }
    );

    // is equal to

    memory::CtxArray< int, ParticleDomCfg > vIdx;
    ForEachIdx< ParticleDomCfg > forEachParticle{ workerIdx }(
        [&]( uint32_t const linearIdx, uint32_t const idx )
        {
            vIdx[ idx ] = linearIdx;
        }
    );
    ```

### DIFF
--- a/src/libPMacc/include/particles/ParticlesBase.hpp
+++ b/src/libPMacc/include/particles/ParticlesBase.hpp
@@ -31,6 +31,7 @@
 
 #include "mappings/kernel/StrideMapping.hpp"
 #include "traits/NumberOfExchanges.hpp"
+#include "traits/GetNumWorkers.hpp"
 #include "assert.hpp"
 
 #include <memory>
@@ -110,11 +111,14 @@ protected:
         StrideMapping<AREA, 3, MappingDesc> mapper(this->cellDescription);
         ParticlesBoxType pBox = particlesBuffer->getDeviceParticleBox();
 
+        constexpr uint32_t numWorkers = traits::GetNumWorkers<
+            math::CT::volume<typename FrameType::SuperCellSize>::type::value
+        >::value;
         __startTransaction(__getTransactionEvent());
         do
         {
-            PMACC_KERNEL(KernelShiftParticles{})
-                (mapper.getGridDim(), (int)TileSize)
+            PMACC_KERNEL(KernelShiftParticles< numWorkers >{})
+                (mapper.getGridDim(), numWorkers)
                 (pBox, mapper);
             PMACC_KERNEL(KernelFillGaps{})
                 (mapper.getGridDim(), (int)TileSize)
--- a/src/libPMacc/include/particles/ParticlesBase.kernel
+++ b/src/libPMacc/include/particles/ParticlesBase.kernel
@@ -27,6 +27,9 @@
 #include "mappings/kernel/ExchangeMapping.hpp"
 #include "particles/memory/boxes/ExchangePushDataBox.hpp"
 #include "particles/memory/boxes/ExchangePopDataBox.hpp"
+#include "memory/CtxArray.hpp"
+#include "mappings/threads/ForEachIdx.hpp"
+#include "mappings/threads/IdxConfig.hpp"
 
 #include "particles/operations/Assign.hpp"
 #include "particles/operations/Deselect.hpp"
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by theZiz** [comment]
I really like the idea and implementation!
I see benefits in the GPU programming, but we afaik in the end you do this to enable autovectorization when we later™ switch the Alpaka.
Did you test yet how autovectorizeable your implementation is?
If not, are you interesting in me testing your implementation on different architectures and compilers?

**Comment by psychocoderHPC** [comment]
@theZiz I haven't tested it with alpaka because of the missing alpaka functionality in PIConGPU/dev.

I we would test the vectorization we need to move out the helper methods to a smaller test case (like our matmul) and test the model there.

All in all yes I am interested that someone studies the auto vectorization capability.

**Comment by theZiz** [comment]
@psychocoderHPC:

> I we would test the vectorization we need to move out the helper methods to a smaller test case (like our matmul) and test the model there.
That was the idea, sorry, I could have made it more obvious, what I meant. I will __not__ test vectorization in PIConGPU (yet). :scream: 

> All in all yes I am interested that someone studies the auto vectorization capability.
Awesome, I will do it. I am really agog. :smiley:

File: src/libPMacc/include/particles/ParticlesBase.kernel [hunk 1]
```diff
@@ -49,181 +52,301 @@ getPreviousFrameAndRemoveLastFrame( const typename T_ParticleBox::FramePtr& fram
...
+    >
+    DINLINE void operator()(
+        T_ParBox pb,
+        Mapping mapper
+    ) const
     {
-        typedef T_ParBox ParBox;
-        typedef typename ParBox::FrameType FrameType;
-        typedef typename ParBox::FramePtr FramePtr;
+        using ParBox = T_ParBox;
```
theZiz: Didn't we have a rule, that we make two different pull requestens for code formating and new features?

src/libPMacc/include/particles/ParticlesBase.kernel
```diff
@@ -49,181 +52,301 @@ getPreviousFrameAndRemoveLastFrame( const typename T_ParticleBox::FramePtr& fram
...
+        using ExchangeDomCfg = IdxConfig<
+            numExchanges,
+            numWorker
+        >;
+
+        memory::CtxArray<
+            uint32_t,
+            ExchangeDomCfg
+        > lastFrameSizeCtx( 0 );
+
```
theZiz: Why is `lastFrameSize`(`Ctx`) not a index local variable anymore? Do you need this this after the `forEach`? Do you assume any optimizations benefits here, I can see? If yes, it would be helpful to comment them. :wink:

src/libPMacc/include/particles/ParticlesBase.kernel
psychocoderHPC: Yes, but this is a full rewrite the new helper are hard to use with the old syntax thus I rewrite everything with the new code style.
All other kernel in this file will also be refactored but it is easier to give the changes step by step.

theZiz: I understand - it still made the review harder. :stuck_out_tongue_winking_eye:

src/libPMacc/include/particles/ParticlesBase.kernel
psychocoderHPC: ## Long

The reason is that this variable was available in all threads before this PR but was only valid initialized for all threads with `threadIdx.x < numExchanges`. 
Now I have a index domain with  numExchanges indices but I don't know which worker handles and store `lastFrameSizeCtx[N]`. You could see each index within the index domain `ExchangeDomCfg` as a virtual worker, it is not important to know which worker is modifying `lastFrameSizeCtx[N]` as long as I only use this context variable always with the some `forEachExchange()` method.

## Short

`lastFrameSizeCtx`is a worker local variable, each worker local variable must be stored in an `CtxArray` to fulfill the *lockstep programming model*

theZiz: Sorry, I overlooked the final write  in line 347. The rest was clear too me. :wink:

src/libPMacc/include/particles/ParticlesBase.kernel
```diff
@@ -49,181 +52,301 @@ getPreviousFrameAndRemoveLastFrame( const typename T_ParticleBox::FramePtr& fram
     return result;
 }
 
+template< uint32_t T_worker >
```
ax3l: `T_numWorkers`

src/libPMacc/include/particles/ParticlesBase.kernel
```diff
@@ -49,181 +52,301 @@ getPreviousFrameAndRemoveLastFrame( const typename T_ParticleBox::FramePtr& fram
...
+        using FramePtr = typedef typename ParBox::FramePtr;
 
-        const uint32_t dim = Mapping::Dim;
-        const uint32_t frameSize = math::CT::volume<typename FrameType::SuperCellSize>::type::value;
+        constexpr uint32_t dim = Mapping::Dim;
+        constexpr uint32_t frameSize = math::CT::volume< typename FrameType::SuperCellSize >::type::value;
         /* number exchanges in 2D=9 and in 3D=27 */
-        const uint32_t numExchanges = traits::NumberOfExchanges<dim>::value;
+        constexpr uint32_t numExchanges = traits::NumberOfExchanges< dim >::value;
+        constexpr uint32_t numWorker = T_worker;
```
ax3l: numWorkers

src/libPMacc/include/particles/ParticlesBase.hpp
```diff
@@ -110,11 +111,14 @@ class ParticlesBase : public SimulationFieldHelper<T_MappingDesc>
         StrideMapping<AREA, 3, MappingDesc> mapper(this->cellDescription);
         ParticlesBoxType pBox = particlesBuffer->getDeviceParticleBox();
 
+        constexpr uint32_t worker = traits::GetNumWorkers<
```
psychocoderHPC: it is to precise :-) and give the user to much information

theZiz:
Good news everyone!

I did some first tests with the most recent Intel compiler and the Knights Landing. Autovectorization is working with `ForEachIdx` if `#pragma ivdep` is used (which is ignored for any other compiler) for the outer loop. Right now I am not changing the simd template argument.

Next task is to have a deeper look, whether `CtxArray` works for autovectorization.

psychocoderHPC:
@theZiz good job, thx for evaluating the new implementation.

Issue #1787, opened 2017-01-24
Test number of particles per cell


n01r:
- [ ] test with `CounterSeq` in particleConfig.param i.e. `<species>_particleCounter` in the HDF5 output
- [ ] test with `CountPerSuperCell` plugin

 and check against our array from the implementation

Issue #1786, opened 2017-01-24
Make `inCellOffset` an optional attribute and include it in the future example ThinFoil.


n01r:
**Update 2017-01-24 : ** 
More generally: remove the in-cell indexing device function from FillGaps kernel but leave it usable in the FillGaps kernel.

theZiz:
You see, we can play this game in two directions. ;)

n01r:
Damn! My plan is foiled!

theZiz:
Guess this is only fair...

theZiz:
Okay, I am sitting on the issue now, but I wonder, why I should move the function out of the struct? Even if it is needed somewhere else, it can be reference with `KernelFillGaps::updateCellsForOneFrame`.

But even if I move this function out of the struct. is there anything else to do what I oversee?

n01r:
Well, yes - this is a more general method. We will likely need it for certain radiation transport methods where we would propagate photons through our simulation volume and have them interact with a randomly chosen ion (i.e. their bound electrons) in the cell. 
Therefore it does not exactly belong to `KernelFillGaps` but to a namespace of collisional-radiative methods.

PR #2016, opened 2017-05-04
Hackathon in cell list PR issue1786

theZiz:
Finally handled issue https://github.com/ComputationalRadiationPhysics/picongpu/issues/1786.

This can now be used to check the results of the changes (see issue https://github.com/ComputationalRadiationPhysics/picongpu/issues/1787) in and outside the FillGaps kernel.

### DIFF
--- a/src/libPMacc/include/particles/ParticlesBase.kernel
+++ b/src/libPMacc/include/particles/ParticlesBase.kernel
@@ -332,8 +332,9 @@ struct KernelFillGaps
 
         if (updateCellList)
         {
-            // if updateCellList == false this code is not compiled at all and no
-            // shared memory reserved from the cuda compiler.
+            /* if updateCellList == false this code is not compiled at all and no
+             * shared memory reserved from the cuda compiler.
+             */
             frameNr = &PMacc::memory::shared::allocate< __COUNTER__, int >();
             cellEntryPoint_sh = &PMacc::memory::shared::allocate< __COUNTER__, memory::Array< uint32_t, TileSize> >();
             cellCount_sh = &PMacc::memory::shared::allocate< __COUNTER__, memory::Array< uint32_t, TileSize> >();
--- a/src/libPMacc/include/particles/ParticlesBase.kernel
+++ b/src/libPMacc/include/particles/ParticlesBase.kernel
@@ -20,6 +20,7 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "particles/operations/updateCellListOfFrame.kernel"
 #include "pmacc_types.hpp"
 #include "particles/memory/boxes/ParticlesBox.hpp"
 #include "particles/memory/boxes/PushDataBox.hpp"
@@ -302,30 +303,7 @@ struct KernelFillGapsLastFrame
 
 struct KernelFillGaps
 {
-    template
-    <
-        size_t TileSize,
-        bool has_no_gaps = true,
-        typename T_CellEntryPoint,
-        typename T_CellCount,
-        typename T_FramePtr,
-        typename T_FrameNr
-    >
-    DINLINE static void updateCellsForOneFrame(T_CellEntryPoint& cellEntryPoint, T_CellCount& cellCount, T_FramePtr& frame, T_FrameNr& frameNr)
-    {
-        auto myParticle = frame[threadIdx.x];
-        if (has_no_gaps || myParticle[multiMask_])
-        {
-            //absolute position in frame list
-            auto myID = threadIdx.x + frameNr * TileSize;
-            auto & myCell = myParticle[localCellIdx_];
-            myParticle[picongpu::inCellOffset_] = ::atomicExch(&(cellEntryPoint[myCell]),myID);
-            ::atomicAdd(&(cellCount[myCell]),1);
-        }
-        __syncthreads( );
-    }
-
-    template<class ParBox, class Mapping>
+    template<class ParBox, class Mapping, bool updateCellList = false>
     DINLINE void operator()( ParBox pb, Mapping mapper ) const
     {
         using namespace particles::operations;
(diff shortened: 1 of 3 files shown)


src/picongpu/include/particles/ParticlesFunctors.hpp
```diff
@@ -313,7 +313,14 @@ struct CallIonization
...
             electronsPtr->fillAllGaps();
-
+            /*constexpr auto TileSize =
+                PMacc::math::CT::volume<
+                    typename decltype(electronsPtr)::MappingDesc::SuperCellSize
+                >::type::value;
+            auto pBox = electronsPtr->getParticlesBuffer->getDeviceParticleBox();
+            PMACC_KERNEL(PMacc::particles::operations::kernel::UpdateCellListOfFramesKernel{})
+                (mapper.getGridDim(), (int)TileSize)
+                (pBox, mapper);*/
```
n01r: Did you add this comment here in order to stress how one would proceed using the kernel?
Maybe add a line of description that it is an example.

theZiz:
Yes, this is why it took so long as I wanted to make the PR on monday, but we had some cluster problems...

src/libPMacc/include/particles/ParticlesBase.kernel
```diff
@@ -347,23 +325,33 @@ struct KernelFillGaps
...
-        for (int i = 0; i < blockDim.x / TileSize; ++i)
-            cellCount_sh[i * blockDim.x + threadIdx.x] = 0;
+
+        int* frameNr = NULL;
+        memory::Array< uint32_t, TileSize>* cellEntryPoint_sh = NULL;
+        memory::Array< uint32_t, TileSize>* cellCount_sh = NULL;
+
+        if (updateCellList)
+        {
+            // if updateCellList == false this code is not compiled at all and no
```
theZiz: Ok

src/libPMacc/include/particles/operations/updateCellListOfFrame.kernel
```diff
@@ -0,0 +1,133 @@
...
+        {
+            TileSize = math::CT::volume<typename Mapping::SuperCellSize>::type::value,
+            Dim = Mapping::Dim
+        };
+
+        typedef typename ParBox::FramePtr FramePtr;
+
+        DataSpace<Dim> superCellIdx( mapper.getSuperCellIndex( DataSpace<Dim > (blockIdx) ) );
+
+        //data copied from right (last) to left (first)
```
theZiz: This is not my comment! I just left it there from the past™.

src/libPMacc/include/particles/ParticlesBase.kernel
```diff
@@ -389,11 +377,13 @@ struct KernelFillGaps
 
             if ( counterGaps == 0 )
             {
-                updateCellsForOneFrame<TileSize>(cellEntryPoint_sh, cellCount_sh, firstFrame, frameNr);
+                if (updateCellList)
+                    kernel::UpdateCellListOfFrame{}.template operator()<TileSize>(*cellEntryPoint_sh, *cellCount_sh, firstFrame, *frameNr);
```
theZiz: Did you ever try to call `operator()` with template arguments?
Hell! https://stackoverflow.com/questions/1762049/templated-operator-overload-c

src/picongpu/include/particles/ParticlesFunctors.hpp
theZiz: Yes, I tested here a bit around and thought it would be pity to loose the code as we could reuse parts later for testing.

I will add a comment. :wink:

n01r:
You did yours already, at least. I still have to but there are so many other important things to do as well :smile: .

src/libPMacc/include/particles/operations/updateCellListOfFrame.kernel
n01r: Your chance to set right the wrongdoings of the past! :3

theZiz: Well, I have now idea, what the comment meant, so I am afraid to touch it. Probably some secret language between René and Axel :smiling_imp:

n01r:
Interesting, the compile suite reports that *something* fails. :smile:

theZiz:
In that case, we should fix _something_. :stuck_out_tongue_winking_eye:

src/libPMacc/include/particles/ParticlesBase.kernel
ax3l: yes, that's just wonderful C++ syntax in member functions with templates :)

you might know this style from things like our `map<...>` functions in kernels :)

ax3l: Here is an other example I documented years ago: https://github.com/ax3l/cpp_snippets/blob/master/template_error/template.cpp

src/libPMacc/include/particles/operations/updateCellListOfFrame.kernel
ax3l: haha, just kick it it's a really useless comment ^^

ax3l:
just compile locally on an interactive node, sometimes the compile suite is lazy if problems are *too obvious* ;)

src/libPMacc/include/particles/operations/updateCellListOfFrame.kernel
theZiz: Will do so.

theZiz:
Okay, it compiles and runs on an interactive node, so I guess we are fine.

theZiz:
So, can I merge so that @n01r can start with the funny part of checking the results?

psychocoderHPC:
I looked over this pull request and I think we must move the cell linked list generation out of `KernelFillGaps`.

The reasons are:
  - the generated linked list is invalid after `KernelFillGapsLstFrame` is called (this kernel is called after KernelFillGaps)
  - the code of `KernelFillGaps` gets to hard to maintain with the optional elements for the linked list
  - an independent kernel will be much easier to maintain (there is no need for optional shared memory if a template parameter is true)

ax3l:
but wouldn't be impossible to optionally also call this functor inside the `kernelFillGaps` & `KernelFillGapsLstFrame`, no?

psychocoderHPC:
@ax3l Yes it is possible, but if you look to the implementation you would see it is hard to maintain.
Also the of the next steps would be to create other linked list structures like a linked list of all particles N cells around of a cell. Maybe we need to call multiple time the linked list generation within one PIC cycle.

Issue #2012, opened 2017-05-03
err: "TILT_X" is undefined in device code


BifengLei:
Building err occurs in the current master version 0.2.4 when the laser profile _laserPulseFrontTilt_ is used:
`picongpu/src/picongpu/include/fields/laserProfiles/laserPulseFrontTilt.hpp(96): error: identifier "picongpu::laserPulseFrontTilt::TILT_X" is undefined in device code.`

It is sure that _TILT_X_ is correctly defined in [laserConfig.unitless](https://github.com/ComputationalRadiationPhysics/picongpu/blob/master/src/picongpu/include/simulation_defines/unitless/laserConfig.unitless#L38)

PrometheusPi:
I can reproduce this **bug** with the current `dev` using the default LaserWakefield example and changing the laser to *laserPulseFrontTilt*.

PrometheusPi:
The commit in pull request #2017 should solve your problem.

PrometheusPi:
@BifengLei Thank you for reporting this bug. I hope that the fix mentioned above helps you. 

If you encounter another issue, please do not hesitate to contact us again.

BifengLei:
@PrometheusPi . Great. It works. #2017

PR #2017, opened 2017-05-04
Fix undefined device variable in tilted laser pulse

PrometheusPi:
This fixes issue #2012. 
For some reasons the `constexp` variable `TITLT_X` is not known on the device side in the `tan()` call. If defined before as `const`, the compile issue is solved. 
@psychocoderHPC Is there a better solution than the work around of this pull request?

### DIFF
--- a/src/picongpu/include/fields/laserProfiles/laserPulseFrontTilt.hpp
+++ b/src/picongpu/include/fields/laserProfiles/laserPulseFrontTilt.hpp
@@ -102,7 +102,8 @@ namespace picongpu
             constexpr float_X focusPos = FOCUS_POS - laser::initPlaneY * CELL_HEIGHT;
 
             const float_X timeShift = phase / (float_X(2.0) * float_X(PI) * float_X(f)) + focusPos / SPEED_OF_LIGHT;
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
The issue is that it is not allowed to take a reference to a `constexpr` variable.

```C++
// it is not possible to call this function with a constexpr variable
int foo( int const &  bar)
{
    return bar;
};
constrexpr int x = 1;
foo( x ); // this should fail because nvcc not allowed to take a reference from a constexpr variable
```

PrometheusPi:
Okay - thank you for the explanation. So is adding a `const` variable the only way to avoid this issue or is there a better solution to avoid running into this "feature"?

psychocoderHPC:
`tan(T const x)` is a partial solution but only for the tan() method. But the interface of `tan` is a template so `T` can be anything, therefore it should not used by value.

PrometheusPi:
I agree - so adding the extra line ```const ... ; ```is the only option?

psychocoderHPC:
> I agree - so adding the extra line `const ... ;` is the only option?

IMO yes

PR #2028, opened 2017-05-09
Python: Fix Scripts PEP8

ax3l:
Fix PEP8 of python scripts, but a 4 files are left before we can introduce a Travis test to enforce it for new PRs.


ax3l:
For this files I checked all changes with pep8 until there were no further warnings.
But I did not RT test your scripts in case you want to do that?

PrometheusPi:
@ax3l No, I was just interested in the PEP8 test - I did not see any changes that might break the code, thus I am fine with not doing run time tests on all scripts

PR #2030, opened 2017-05-09
update pizdaint data

psychocoderHPC:
- update profile exmaple
- update tbg templates
- rename `total.tpl` to `large.tpl` (the queue names has changed with the last pizdaint upgrade)

## Tests

- [x] pizdaint 05/2017
- [x] X30 -> X50: only 12 CPU cores / node (.tpl files, too)

### DIFF
--- a/src/picongpu/submit/pizdaint-cscs/picongpu.profile.example
+++ b/src/picongpu/submit/pizdaint-cscs/picongpu.profile.example
@@ -7,31 +7,31 @@
 # if the wrong environment is loaded we switch to the gnu environment
 module li 2>&1 | grep "PrgEnv-cray" > /dev/null
 if [ $? -eq 0 ] ; then
-    module swap PrgEnv-cray PrgEnv-gnu/5.2.82
+    module swap PrgEnv-cray PrgEnv-gnu/6.0.3
 else
-    module load PrgEnv-gnu/5.2.82
+    module load PrgEnv-gnu/6.0.3
 fi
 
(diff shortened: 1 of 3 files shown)


src/picongpu/submit/pizdaint-cscs/picongpu.profile.example
```diff
@@ -46,7 +46,7 @@ export CRAY_CPU_TARGET=x86-64
 export CC=`which cc`
 export CXX=`which CC`
 
-export PICSRC=$HOME/src/picongpu
+export PICSRC=$HOME/picongpu
```
ax3l: please keep this in `$HOME/src` as in our examples :)

ax3l:
@psychocoderHPC yes, only 12 cores per node: http://www.cscs.ch/computers/piz_daint/index.html

ax3l:
@psychocoderHPC with 16 GB global, do we need to evaluate if we might better take smaller particle heaps or e.g. 4 MPI host ranks (with individual domains) per P100 or is the 2/4GB problem with slow access to global no more?

psychocoderHPC:
@ax3l Currently we can't support GPU sharing over multiple MPI ranks. 

The issue with the slow random access to global memory is solved with pascal. IMO Pascal support 16 GB RAM random memory access without performance impact.
We have random access memory access pattern in PIConGPU but we are cluster our species data in frame therefore the pressure to the TLB is not critical.

PR #2031, opened 2017-05-09
Fix #2029 MPL Embedded Plots Sphinx

### LINKED ISSUE #2029: Docs SVG: Ionization
As with the PIConGPU logo, we need to create `.pdf` alternatives of the `.svg` images for `.. only:: latex` for `make latexpdf`

also, html build on RTD seems to forget copying the live created new images (`make html`)

fixing via http://matplotlib.org/sampledoc/extensions.html

ax3l:
Fix #2029 by using the official matplotlib (MPL) sphinx extension to embed plots on-the-fly.

Also fixed broken `make latexpdf` compile in usage of own align (use [nowrap](http://stackoverflow.com/questions/24867685/aligned-equations-with-mathjax-and-with-equation-numbers)).

Note:
- this might need regressions after merge, e.g. to set `agg` as backend by hand

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,3 @@ docs/xml/
 docs/doxygen_sqlite3.db
 # sphinx & breathe output
 docs/build/
-
-# generated docs output
-docs/source/models/field_ionization_comparison_c_ii_ionization.svg
-docs/source/models/field_ionization_effective_potentials.svg
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -37,7 +37,10 @@ show_authors = True
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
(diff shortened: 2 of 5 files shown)


ax3l:
@PrometheusPi wuhu, tests are done :)

n01r:
I tried adding a caption to the `.. plot::` directive but somehow that didn't work. Even though I looked it up and it seemed possible. ^^;

ax3l:
@n01r nah it's not. maybe a hidden option but I did not find it in the docs. did you find a list of available options?

n01r:
Nope - had a look before leaving today and just found an issue.
https://github.com/matplotlib/matplotlib/issues/2850

PrometheusPi:
@ax3l Do we want to add the fix suggested by @aa956 or do we just live without caption?
Why is the test suite pending while the pull request is already merged and you already deleted the branch?

ax3l:
@PrometheusPi there seems to be no caption with this plugin, so we moved the caption back to the title for now

PR #2033, opened 2017-05-11
fix unit ionization rate plot

PrometheusPi:
The unit of a (probability) rate is `1/s` not `Hz` (see [wikipedia article](https://en.wikipedia.org/wiki/Hertz#Definition)).
This pull request fixes the unit in the plot.

### DIFF
--- a/docs/source/models/field_ionization_comparison_c_ii_ionization.py
+++ b/docs/source/models/field_ionization_comparison_c_ii_ionization.py
@@ -97,7 +97,7 @@ if __name__ == "__main__":
     plt.xlim([1e-2, 1e1])
     plt.yscale("log")
     plt.xscale("log")
-    plt.ylabel("ionization rate $\Gamma$ [Hz]")
+    plt.ylabel("ionization rate $\Gamma$ [\mathrm{s}^{-1}]")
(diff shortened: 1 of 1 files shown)


docs/source/models/field_ionization_comparison_c_ii_ionization.py
```diff
@@ -97,7 +97,7 @@ def ADK_rate_simple(Z, E_i, F):
     plt.xlim([1e-2, 1e1])
     plt.yscale("log")
     plt.xscale("log")
-    plt.ylabel("ionization rate $\Gamma$ [Hz]")
+    plt.ylabel("ionization rate $\Gamma$ [\mathrm{s}^{-1}]")
```
ax3l: nooooo, this is NOT in math `$$` and I just merged it. pls fix in a follow-up to `[s$^{-1}$]`

PrometheusPi: Sorry - I will fix it now.

steindev:
This issue is also explained by NIST. See footnote on unit Hz and Bq in Table 3 of the following NIST page: (https://www.nist.gov/pml/nist-guide-si-chapter-4-two-classes-si-units-and-si-prefixes)

PR #2037, opened 2017-05-11
Examples: Collisional Ionization Include

ax3l:
Adds the include in `speciesDefiniton.param` to examples where it was still missing. Otherwise Thomas-Fermi ionization could not be used in the `ionizers< ... >` sequence.

### DIFF
--- a/examples/Bremsstrahlung/include/simulation_defines/param/speciesDefinition.param
+++ b/examples/Bremsstrahlung/include/simulation_defines/param/speciesDefinition.param
@@ -28,9 +28,6 @@
 #include "particles/Particles.hpp"
 #include <boost/mpl/string.hpp>
 
-#include "particles/ionization/byField/ionizers.def"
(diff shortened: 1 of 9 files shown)


psychocoderHPC:
#2040 fix is solving a missing include which could be related to this PR.

ax3l:
imo, #2040 is not related to this PR. This PR is purely about having a class `::ThomasFermi` as forward define available in `speciesDefinition.param`

ax3l:
ionizers.def only contains data tables and no algorithms, but as long as it is included before `speciesDefinition` we could in principle also put the includes there, although they are not used there.

a cleaner way would be to hide those from the user completely and add such forward defines in the loader

psychocoderHPC:
@ax3l  I created an pull request to your PR to move the includes out of speciesDefinition.param

ax3l:
@psychocoderHPC thanks, applied your PR!

PR #2042, opened 2017-05-12
fix typo causing no absorber in 3D

PrometheusPi:
This fixes a bug introduces **3 years ago** with #212.

The typo `SDIMDIM` in the precompiler if-clause evaluates as false, thus causing **no absorber in z**. 

This effects all 3D simulations with absorbers. 

The field along z-are not absorbed, see here:
absorber_issue

- [x] check if this pull request fixes the missing absorber


PrometheusPi:
This pul request fixes the missing absorber in z-direction:
absorber_issue_solved

ax3l:
Great catch! Pls also link it to project 0.2.5 and 0.3.0 so I can backport it in a week when I work through the latest updates!

PR #2043, opened 2017-05-14
Copy pymodules from examples to run directory

n01r:
This pull request modifies `pic-create` in order to copy analysis tools from
the examples in source to the parameter set.

Additionally `submitAction.sh` was modified to copy these tools to the run
directory.

This is a follow-up to #1979 .

### DIFF
--- a/pic-create
+++ b/pic-create
@@ -23,8 +23,13 @@ this_dir=`dirname $0`
 
 #only copy submit and params if we clone from default pic folder
 default_folder_to_copy="submit include/simulation_defines/param include/simulation_defines/unitless"
-#if we clone a project we copy full include 
-folder_to_clone="submit include"
+#if we clone a project we copy full include
+if [ -d "$1/pymodules" ] ; then
+    folder_to_clone="submit include pymodules"
+else
+    folder_to_clone="submit include"
+fi
+
 files_to_copy="cmakeFlags executeOnClone"
 
 help()
--- a/src/picongpu/submit/submitAction.sh
+++ b/src/picongpu/submit/submitAction.sh
@@ -24,6 +24,10 @@ mkdir picongpu
 cp -r $TBG_projectPath/bin picongpu
 cp -r $TBG_projectPath/include picongpu
 cp -r $TBG_projectPath/submit picongpu
+if [ -d $TBG_projectPath/pymodules ]
+then
+  cp -r $TBG_projectPath/pymodules picongpu
+fi
 if [ -f $TBG_projectPath/cmakeFlags ]
 then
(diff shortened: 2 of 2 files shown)


pic-create
```diff
@@ -23,8 +23,13 @@ this_dir=`dirname $0`
 
 #only copy submit and params if we clone from default pic folder
 default_folder_to_copy="submit include/simulation_defines/param include/simulation_defines/unitless"
-#if we clone a project we copy full include 
-folder_to_clone="submit include"
+#if we clone a project we copy full include
+if [ -d "$1/pymodules" ] ; then
```
PrometheusPi: REQUIRED:
The code did not check yet, if `$1` is set, this is done in line 53. Please move this code after line 53 and perhaps add `pymodules` there **or* check the positional arguments here.

src/picongpu/submit/submitAction.sh
```diff
@@ -24,6 +24,10 @@ mkdir picongpu
 cp -r $TBG_projectPath/bin picongpu
 cp -r $TBG_projectPath/include picongpu
 cp -r $TBG_projectPath/submit picongpu
+if [ -d $TBG_projectPath/pymodules ]
+then
+  cp -r $TBG_projectPath/pymodules picongpu
```
PrometheusPi: QUESTION:
@ComputationalRadiationPhysics/picongpu-developers is the `picongpu` directory the right place for this python code?

n01r: Had an offline talk with @ax3l about this. The directories `include` and `submit` will be in the `picongpu` directory as well. Before, they are in their respective `examples/<MyExample>/` directory - as is the `pymodules` directory. Therefore it makes sense to treat all of those directories the same. Especially since the analysis tools in `pymodules` will most-likely be example-specific.

PrometheusPi: Okay - that's fine with me 👍

n01r:
**offline** @psychocoderHPC suggests to rename `pymodules` to `tools` in order to avoid binding this too much to python. He also asks whether to have those copied into the `bin` directory of the run.

I agree with the renaming. I would like the tools to remain next to `include` and `submit`, though. :)

ax3l:
Please not, this dir should contain modules and python init files and not generic tools.
If your module does, by coincidence, contain a main routine that is fine. Pls keep the naming I suggested, can elaborate next week if necessary.

pic-create
ax3l: I will supersede this change with #2058

ax3l:
superseded by #2058

PR #2044, opened 2017-05-15
remove extra typename in documentation of manipulators

PrometheusPi:
This pull request removes a wrong `typename` in the documentation of the particle manipulators. 

Adding this `typename` will lead to a compile time error.


PrometheusPi:
@psychocoderHPC Is the `bug` flag correct here? The error is just in the example of the documentation, not in the active code?

PrometheusPi:
*discussed offline* with @psychocoderHPC: Yes, since a simple copy and past would cause a compile error - we consider it a bug.

ax3l:
sure, documentation bug :)

Issue #1850, opened 2017-02-16
Ionization: tie up loose ends


n01r:
There is a list of little things that should be refactored about ionization

* comments that @psychocoderHPC made on already merged PRs concerning ionization should be gathered and processed
* comments describing the `Algorithm<Model>.hpp` files (like recently touched in #1842) are not correct anymore after multiple refactorings  
* using @Heikman's particle creation (but that is a separate issue #1357)

Issue #1357, opened 2016-01-28
Refactor Ionization: Particle Creation


ax3l:
With the refactoring of the technical parts that do the low level particle creation in #1353 the ionization module needs a refactoring to use that new API. This overall refactoring of ionization should make it less technical and more physics-focused.

Please do this timely after #1353 was merged.

CCing @Heikman @n01r @psychocoderHPC

ax3l:
@n01r has this been implemented?

n01r:
Actually no, not yet. I remember having offline discussions with @Heikman he treats something slightly different than I do. I have to get back onto this to remember where the difference was.

n01r:
In offline discussions with @psychocoderHPC we agreed on moving the change of the parent ion's attributes to one place in the code to keep the state of the ion always valid between algorithms. 
Currently when a new electron is created the functor `WriteElectronIntoFrame` clones the electron from the ion and changes the ion's momentum. Yet the number of `boundElectrons` is changed in the `<Model>_Impl`. That means that even though nothing touches the ion in between, its state is invalid which can lead to unwanted side-effects.

It is foreseeable that we might want to give additional properties to the new free macro electrons as well as change the residual ion in a model-specific way. E.g.:
- give initial momentum to the electron (*single/ multi-photon ionization*)
- create one electron with higher weighting instead of multiple electrons in one step (maybe *Thomas-Fermi*)
- leave the ion in an excited state (*direct impact ionization*)
...

The naming `WriteElectronIntoFrame` is too narrow then. It should also take a model-specific creation/ attribute assignment functor, then, for instance.

n01r:
I'm closing this issue due to it being resolved for a while now.

PR #2046, opened 2017-05-15
Fix charge conservation in ThomasFermi

### LINKED ISSUE #2024: TF: Not Charge Conserving
The Thomas-Fermi ionization creates either electrons with too high weighting or too many of them.

See the comparison from the new LCT foil example in #2008 

```python
#!/usr/bin/env python
#

import h5py as h5
import numpy as np

starts = {
    "random": "004",
    "quiet": "005"
}
start = starts["random"]
sims = {
    "BSI": "lctFoil-" + start + "-BSI",
    "ADK": "lctFoil-" + start + "-ADK",
    "TF": "lctFoil-" + start + "-TF",
    "combined": "lctFoil-" + start + "-BSI-ADK-TF"
}
dt = 4.91356e-18 # s
dx = 800.e-9 / 384. # m
step = 750

fig = plt.figure()
cax = fig.add_axes([0.87, 0.1, 0.01, 0.75])
ax1 = fig.add_axes([0.1, 0.5, 0.35, 0.35])
ax2 = fig.add_axes([0.5, 0.5, 0.35, 0.35])
ax3 = fig.add_axes([0.1, 0.1, 0.35, 0.35])
ax4 = fig.add_axes([0.5, 0.1, 0.35, 0.35])
axes = [ax1, ax2, ax3, ax4]

def get_nZ(flds, species):
    r = flds[species + "_chargeDensity"]
    d = r[()] * r.attrs["unitSI"] / 1.602e-19 * 1.e-6 # elements / cm^3
    return d

def plot_sim(ax, sim):

    f = h5.File("../" + sims[sim] + "/simOutput/h5/simData_" + str(step) + ".h5" , "r")
    ne = get_nZ(f["/data/" + str(step) + "/fields/"], "e")
    nH = get_nZ(f["/data/" + str(step) + "/fields/"], "H")
    nC = get_nZ(f["/data/" + str(step) + "/fields/"], "C")
    nN = get_nZ(f["/data/" + str(step) + "/fields/"], "N")

    d = ne + nH + nC + nN
    ax.set_title(sim)
    return ax.imshow(
      #np.abs(d),
      d,
      #cmap='CMRmap_r',
      cmap='RdBu',
      origin="lower",
      aspect="auto",
      interpolation="nearest",
      #vmin=0.,
      #vmax=2.e24,
      vmin=-3.e22, vmax=3.e22,
      extent=[0., dx * 1.e6 * d.shape[0], 0., dx * 1.e6 * d.shape[1]]
    )
    f.close()
    

i = 0
for sim in sims:
    print(sim)
    im = plot_sim(axes[i], sim)
    i += 1

#fig.colorbar(im, cax=cax, label=r'$n_e$ [$q_e \cdot$ cm$^{-3}$]')
#fig.colorbar(im, cax=cax, label=r'$n_{Z,C}$ [$q_e \cdot$ cm$^{-3}$]')
fig.colorbar(im, cax=cax, label=r'$n_{Z,\Sigma{e,H,C,N}}$ [$q_e \cdot$ cm$^{-3}$]')
fig.suptitle("time = {:5.3f} fs".format(dt * step * 1.e15))
ax3.set_xlabel(r'$x$ [$\mu$m]')
ax4.set_xlabel(r'$x$ [$\mu$m]')
ax1.set_ylabel(r'$y$ [$\mu$m]')
ax3.set_ylabel(r'$y$ [$\mu$m]')
plt.show()
```

## Comparisons

(laser from bottom plane)

### Quiet Start

**n_e**
quietstart

**n_Z,C**
quietstart_c

**sum(n_Z)**
quietstart_sum

### Random Start

**n_e**
randomstart

**n_Z,C**
randomstart_c

**sum(n_Z)**
randomstart_sum

n01r:
Closes #2024.

This pull request fixes Thomas-Fermi ionization to ignore predicted re-combinations which did lead to ionizing the same ion state several times.

### Outdated

This pull request moves

    particle[boundElectrons_] = newBoundElectrons;

after the safety check. Now the number of bound electrons correctly changes,
i.e. **only** if the new number is lower than the number before and ionization
happens.

This PR also increases the detail of documentation and fixes a code line
which overwrote a variable. The latter issue had no functional relevance
and did not produce any erroneous behaviour.

Thanks @ax3l, @psychocoderHPC for spotting this one!

### DIFF
--- a/src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
+++ b/src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2016-2017 Marco Garten
+/* Copyright 2016-2017 Marco Garten, Axel Huebl
  *
  * This file is part of PIConGPU.
  *
@@ -17,11 +17,11 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
-/** \file AlgorithmThomasFermi.hpp
+/** @file AlgorithmThomasFermi.hpp
  *
  * IONIZATION ALGORITHM for the Thomas-Fermi model
- * - implements the calculation of average ionization degree and returns the
- *   new number of bound electrons
+ * - implements the calculation of the new number of free macro electrons
+ *   from the Thomas-Fermi average charge state
  * - is called with the IONIZATION MODEL, specifically by setting the flag in
  *   @see speciesDefinition.param
  */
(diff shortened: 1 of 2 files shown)


PrometheusPi:
as discussed *offline* with @n01r:
He will add the simulation results tomorrow

ax3l:
cool, the line I showed you last week ;)

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
```diff
@@ -272,19 +272,33 @@ namespace ionization
...
                 IonizationAlgorithm ionizeAlgo;
                 float_X newBoundElectrons = ionizeAlgo(
                     kinEnergyDensity, density,
                     particle, this->randomGen()
                     );
 
-                particle[boundElectrons_] = newBoundElectrons;
-
-                /* safety check to avoid double counting since recombination is not yet implemented */
+                /* Safety Check: Would the number of bound electrons after the
```
ax3l: this line does not explain why it is save.
as noted in the second paragraph (only changed when ionization occurs), one would expect, that your algorithm above should keep the the variable `newBoundElectrons` to its original value and that this if only guards an necessary write of the same value.

if that is the case, then this comment line makes no sense.
if that is not the case, the `ionizeAlgo` call behaves weird or in an undocumented/unexpected way and needs to be fixed.

n01r: Yeah, I get your point.

My line of thinking was:
Our model does not know anything about implementation. We ask from it what the average charge state depending on local parameters at that very specific point in time would be and have it transfer that answer into number of bound electrons. Then at the place where we decide what we actually do with the ion we find that we only want to change the value `boundElectrons` for the ions in case of ionization. That requires a check, comparing bound electron number before and after. Only if the model predicts that it is lower afterwards we *actually* change it.

Thus I wanted to separate the model and its result from the decision what to do with the result.

I will check this again for consistency. Thanks for the feedback!

@psychocoderHPC suggested to include the option to create electrons with higher weighting instead of multiple macroparticles in case the number of electrons created in one time step is larger than 1.

n01r:
```python
#!/usr/bin/env python
#

import matplotlib as mpl
import h5py as h5
import numpy as np

# I left out the update of the matplotlib params here

# path to simulation
sim = "./"

# output of macro particle counter every 50 steps
e_mpc_file = sim + "/simOutput/e_macroParticlesCount.dat"
C_mpc_file = sim + "/simOutput/C_macroParticlesCount.dat"
# read the macro particle counts
e_mpc = np.loadtxt(e_mpc_file)
C_mpc = np.loadtxt(C_mpc_file)

mPartRatio = e_mpc[:, 1] / C_mpc[:, 1]

# steps for HDF5 output
steps = np.arange(9) * 250
# create array for the mean ion charge of carbon
mean_charges = np.zeros([9], dtype=float)
# carbon proton number
Z_C = 6

for i in range(9):
    # output file handle
    f = h5.File(sim + "/simOutput/h5/simData_" + str(steps[i]) + ".h5", "r")
    # bound electrons of carbon
    be_C = np.array(f["/data/" + str(steps[i]) +
                      "/particles/C/boundElectrons"])
    # average number of bound electrons
    mean_be_C = np.mean(be_C)
    # average ion charge
    mean_Z_i = Z_C - mean_be_C
    # fill the array
    mean_charges[i] = mean_Z_i

# visualization of results
plt.plot(e_mpc[:, 0], e_mpc[:, 1] / C_mpc[:, 1], label=r"$N_e/N_C$")
plt.ylabel(r"$N_e / N_C$")
plt.xlabel("Step")

plt.plot(
    steps,
    mean_charges,
    marker="+",
    mew=3,
    ls="",
    label=r"average $Z_\mathrm{ion}$")
plt.legend(loc="best")

plt.draw()
plt.show()
```
2017_05_16_fix_charge_conservation

I used only carbon ions in this simulation and compared the average charge state of the ions against the ratio of macroparticles (electrons / ions). The output of macroparticles happens each 50 steps while the average charge state is only calculated for every 250 steps. Before the fix we had too many electrons by a factor of 25. After the fix the matching is very good.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
```diff
@@ -272,19 +272,33 @@ namespace ionization
...
                 {
-                    /* determine number of new macro electrons to be created */
+                    /* change the ion attribute: 
+                     * set number of bound electrons after ionization
+                     */
+                    particle[boundElectrons_] = newBoundElectrons;
+                    /* determine number of new free macro electrons 
+                     * to be created in the ionization routine
+                     */
                     newMacroElectrons = static_cast<unsigned int>(prevBoundElectrons - newBoundElectrons);
```
psychocoderHPC: why not calling it `numFreeMacroElectrons` or `numFreeMacroElectrons`?

n01r:
I hope the documentation is okay for now, until we refactor the particle creation for consistent separation of physics and informatics in #1357 . Do you agree for now @psychocoderHPC ?

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
```diff
@@ -272,20 +276,39 @@ namespace ionization
...
 
-                /* this is the point where actual ionization takes place */
+                /* Thomas-Fermi calculation step:
+                 * Determines the new average charge state for each ion under
+                 * LTE conditions. Returns the new number of bound electrons
+                 * according to this average charge state and a Monte-Carlo
+                 * step which decides for an integer number of macro electrons.
+                 */
                 IonizationAlgorithm ionizeAlgo;
                 float_X newBoundElectrons = ionizeAlgo(
```
ax3l: this is then *not* an "ionizeAlgo" but a detailed balance algorithm

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
```diff
@@ -272,20 +276,39 @@ namespace ionization
...
                     kinEnergyDensity, density,
                     particle, this->randomGen()
                     );
 
-                particle[boundElectrons_] = newBoundElectrons;
-
-                /* safety check to avoid double counting since recombination is not yet implemented */
+                /* Safety Check: ionize only if the number of bound electrons
```
ax3l: Safety check is generic and misleading.

You are doing a detailed balance and now voluntarily ignore re-combinations.
```C++
/* Only account for ionization: we only increase the charge state of an ion
  * if necessary, but ignore recombination of electrons as prediced by the implemented
  * detailed balance algorithm.
  */
```

This is also the reason why I would change the result of the TF implementation *inside* the functor with `if (prevBoundElectrons > newBoundElectrons)` instead of throwing this back all the way to here and making a dozen of comments that the implementation is not what one expects.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -58,7 +58,14 @@ namespace ionization
     {
         /** Functor implementation
          *
-         * \tparam DensityType type of number
+         * Calling this functor gives a prediction on the ion charge state
```
ax3l: I would make the `operator()` a `DINLINE` member function and add the `if (prevBoundElectrons > newBoundElectrons)` check in the operator to make this a real ionizer algorithm and not a detailed balance functor

n01r:
Refactoring done & runtest successful, @ax3l

ax3l:
... let me fix that for you. implemented *exactly* in the wrong file :)

ax3l:
@n01r I fixed that for you, now in a clean state. can you pls do a RT test again?

n01r:
RT test look good :+1:

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -41,118 +41,178 @@ namespace particles
...
+         * free macro electrons to create. This prediction is based on the
+         * average charge state in the Thomas-Fermi model.
+         * The functor calculates the integer number of bound electrons from
+         * this state by a Monte-Carlo step.
+         *
+         * \tparam ParticleType type of particle to be ionized
+         *
+         * \param ZStar average charge state in the Thomas-Fermi model
+         * \param parentIon particle instance to be ionized
+         * \param randNr random number
```
psychocoderHPC: return value description is missing, please do not forget the the the output range

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -41,118 +41,178 @@ namespace particles
...
+                )
+            );
 
-                /* integral part of the charge state */
-                float_64 intZStar;
-                /* fractional part of the charge state */
-                float_X fracZStar = static_cast<float_X>(math::modf(ZStar,&intZStar));
+            return ZStar;
+        }
+        /** Functor implementation
```
ax3l: missing empty newline

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
```diff
@@ -264,29 +267,27 @@ namespace ionization
...
+                    this->randomGen()
+                );
+
+                /* ionization of the ion by reducing the number of bound electrons
+                 *
+                 * optimization: only accesses global memory if the charge
+                 *               state did really change
+                 */
+                if( numNewFreeMacroElectrons > 0u )
+                    particle[ boundElectrons_ ] -= float_X( numNewFreeMacroElectrons );
```
psychocoderHPC: possible BUG: This creates a negative number of bound electrons if `numNewFreeMacroElectrons > particle[ boundElectrons_ ]` because the output range of `ionizeAlgo` is not defined.

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
n01r:
```c++
/* 
 * \return numNewFreeMacroElectrons number of new macro electrons to create
 *    range: [0 , boundElectrons]
 */
```

src/picongpu/include/particles/ionization/byCollision/ThomasFermi/ThomasFermi_Impl.hpp
n01r: **offline:** Since we always check beforehand if `chargeState < protonNumber` we prevent that from happening and we can leave the redesign open until we do it for all the ionization models.
We could put a `\todo` here until it is fixed by design with issue #1850 
```c++
/*
 * \todo refactor for consistency with the other ionization methods
 *    and to preserve a valid ion state (issue #1850)
 */

PR #2051, opened 2017-05-17
fix shadowed member variable

psychocoderHPC:
Remove usage of local instead of the member variable.

The result of the bug is that more buffer than needed have stored the buffer size on the device.
This results in more memory copies from/to the device.
This could be negative for the strong scaling.

### DIFF
--- a/src/libPMacc/include/memory/buffers/DeviceBufferIntern.hpp
+++ b/src/libPMacc/include/memory/buffers/DeviceBufferIntern.hpp
@@ -56,14 +56,14 @@ public:
         //create size on device before any use of setCurrentSize
         if (useVectorAsBase)
         {
-            sizeOnDevice = false;
-            createSizeOnDevice(sizeOnDevice);
+            this->sizeOnDevice = false;
+            createSizeOnDevice(this->sizeOnDevice);
             createFakeData();
             this->data1D = true;
         }
         else
         {
-            createSizeOnDevice(sizeOnDevice);
+            createSizeOnDevice(this->sizeOnDevice);
             createData();
             this->data1D = false;
         }


src/libPMacc/include/memory/buffers/DeviceBufferIntern.hpp
```diff
@@ -56,14 +56,14 @@ class DeviceBufferIntern : public DeviceBuffer<TYPE, DIM>
...
-            sizeOnDevice = false;
-            createSizeOnDevice(sizeOnDevice);
+            this->sizeOnDevice = false;
+            createSizeOnDevice(this->sizeOnDevice);
             createFakeData();
             this->data1D = true;
         }
         else
         {
-            createSizeOnDevice(sizeOnDevice);
```
PrometheusPi: QUESTION:
This means that the constructor argument `sizeOnDevice` is always ignored?
Is this the wanted behavior?

Why is then in the following constructor the constructor argument used but not the member variable  `sizeOnDevice`.

psychocoderHPC: The parameter `sizeOnDevice` is used in [line 52](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2051/files/cb08b91920612dd15dc5695d0b407eae832608aa#diff-b55f61e3b7f4d0536bd72930ba24ad25L52).

psychocoderHPC:
@PrometheusPi  You are right it would be good to refactor this part of code. The second constructor is using the parameter instead of the member variable because the value is not changed. In the first constructor the value of the local variable is changed but we forget to propagate this change to the class member.

PrometheusPi:
As a bug fix this is fine.

Issue #2048, opened 2017-05-16
CT Error: Boost 1.64.0 & CUDA 8.0


psychocoderHPC:
It is not possible to compile PIConGPU with boost 1.64.0.

Error:
```
src/libPMacc/include/memory/buffers/DeviceBufferIntern.hpp(50): internal error: assertion failed at:
"/dvs/p4/build/sw/rel/gpu_drv/r361/r361_00/drivers/compiler/edg/EDG_4.10/src/types.c", line 7537

```

**update**: 
- this is a **CUDA 8** bug. 
- CUDA 7.5 and boost 1.64.0 is solved with #2053

**This issue is solved in CUDA9** (information from NVIDIA support)

psychocoderHPC:
[Someone else](https://github.com/dmlc/mxnet/issues/2672) become an error in the some line of the cuda driver.
But I can't see how they solved it and why was the error triggered.

ax3l:
can you pls add host compiler and nvcc version to your PR description?

psychocoderHPC:
## Used System Modules:

  1) gcc/5.3.0                     4) openmpi/1.8.6.kepler.cuda80   7) numactl/2.0.7
  2) cmake/3.3.0                   5) boost/1.62.0                  8) valgrind/3.8.1
  3) cuda/8.0                      6) hdf5-parallel/1.8.15

and own compiled boost 1.64.0

psychocoderHPC:
The issu is comming from the synchrotron or bremstrahlungs module. If I remove both everything is compiling.
Because of the fact that the compile crashs if we switch to boost 1.64.0 I think that the error is coming from the boost runge kutta solver.

I get the following error if I only remove Bremstrahlung: 
```
boost/numeric/odeint/stepper/controlled_runge_kutta.hpp(66): internal error: assertion failed at: "/dvs/p4/build/sw/rel/gpu_drv/r361/r361_00/drivers/compiler/edg/EDG_4.10/src/types.c", line 7543

```

ax3l:
Error in boost numeric, so we need to report it to boost on their Trac. 
Maybe you can spot it in the source, too

psychocoderHPC:
I opened first a CUDA bug report id 1928813 to get more information if it is a internal bug in cudafe or a boost bug. If it is a boost bug more information would be helpful.

Mini example (needs boost 1.64.0): 

Install boost:
```
[...]

PR #2053, opened 2017-05-18
fix boost 1.64.0 odeint integrate bug

psychocoderHPC:
Fix error 
```boost/numeric/ublas/matrix.hpp(5977): error: namespace "boost::serialization" has no member "make_array"```
This bug is introduced with boost 1.64.0.

boost issue [12516](https://svn.boost.org/trac/boost/ticket/12516).

This fix allows to compile PIConGPU with boost 1.64.0 and cuda 7.5.
This fix is **not solving** #2048 for CUDA 8.0.

- [x] 1.64.0 include guard: [`BOOST_VERSION` as `106400`](https://github.com/boostorg/config/blob/boost-1.64.0/include/boost/version.hpp#L22) (not a string!)

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -92,14 +92,14 @@ zlib
 
 boost
 """""
-- 1.57.0+ (``program options``, ``regex`` , ``filesystem``, ``system``, ``thread``, ``math`` and nearly all header-only libs)
+- 1.57.0+ (``program options``, ``regex`` , ``filesystem``, ``system``, ``thread``, ``math``, ``serialization`` and nearly all header-only libs)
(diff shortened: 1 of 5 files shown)


ax3l:
thanks for the fix, but requires new [pre-compiled boost lib `serialization`](https://gist.github.com/ax3l/7450482). I will update the compile suite and push a dependency update to your PR

ax3l:
The include file `boost/serialization/array_wrapper.hpp` was introduced with [boost 1.64.0](https://github.com/boostorg/serialization/tree/boost-1.64.0/include/boost/serialization).

We currently support 1.57.0+, so you can add an include guard if you want.

also, this really needs a boost trac but report.

psychocoderHPC:
I added the inlcude guards

PR #2054, opened 2017-05-18
`FieldJ` bash/insert lockstep refactoring

psychocoderHPC:
rewrite kernel with the lockstep programming model

[Lockstep Programming Model ](http://picongpu.readthedocs.io/en/dev/prgpatterns/lockstep.html) documentation.

### DIFF
--- a/src/picongpu/include/fields/FieldJ.kernel
+++ b/src/picongpu/include/fields/FieldJ.kernel
@@ -361,83 +361,153 @@ struct KernelAddCurrentToEMF
     }
 };
 
+/** copy the current from the local to the intermediate buffer
+ *
+ * @tparam T_numWorkers number of workers
+ */
+template< uint32_t T_numWorkers >
 struct KernelBashCurrent
 {
-    template<class Mapping>
-    DINLINE void operator()(J_DataBox fieldJ,
-                                      J_DataBox targetJ,
-                                      DataSpace<simDim> exchangeSize,
-                                      DataSpace<simDim> direction,
-                                      Mapping mapper) const
+    template< typename Mapping >
+    DINLINE void operator()(
+        J_DataBox const & fieldJ,
+        J_DataBox & targetJ,
+        DataSpace<simDim> const & exchangeSize,
+        DataSpace<simDim> const & direction,
+        Mapping mapper
+    ) const
     {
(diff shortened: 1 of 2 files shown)


src/picongpu/include/fields/FieldJ.kernel
```diff
@@ -361,83 +361,145 @@ struct KernelAddCurrentToEMF
     }
 };
 
+template< uint32_t T_workers >
```
psychocoderHPC: should be `T_numWorkers`

src/picongpu/include/fields/FieldJ.kernel
```diff
@@ -361,83 +361,145 @@ struct KernelAddCurrentToEMF
...
 {
-    template<class Mapping>
-    DINLINE void operator()(J_DataBox fieldJ,
-                                      J_DataBox targetJ,
-                                      DataSpace<simDim> exchangeSize,
-                                      DataSpace<simDim> direction,
-                                      Mapping mapper) const
+    template< typename Mapping >
+    DINLINE void operator()(
+        J_DataBox const & fieldJ,
```
psychocoderHPC: You are right it is illegal in a CUDA `__global__` function.
In this case it fine to use the reference because the functor is already within a kernel.

theZiz: I hoped something like this but as I had no idea about the implementation of `PMACC_KERNEL`, I thought it is better to ask ;)

psychocoderHPC:
> And could you please stop to change to the lockstep programming and doing changes in the indentation at the same time? 

This is not possible because the most parts of the previous kernel are moved to a lamda function, this is always in a deeper depth and therefore the indention needs to be changed.

All changes needs to be checked because the *lockstep programming model* changed the way how threads operate on the data, therefore it is good that the hole kernel gets a review.

src/picongpu/include/fields/FieldJ.kernel
psychocoderHPC: that's absolute the right way! :heart:

theZiz:
Hm, makes sense.

ax3l:
@theZiz if everything is addressed, pls approve the PR

Issue #1968, opened 2017-04-18
Restructure Source Repo


ax3l:
Follow-up to #1926

Restructure PIConGPU repo to mirror a possible install structure:

- `/`
  - `include/` (set `-I` here)
    - `picongpu/` (headers and source files)
    - `libPMacc/` (headers and source files)
    - `mpiInfo/` (if not removed by then)
    - `splash2txt/` (if not removed by then)
    - `png2gas/` (if not moved to an openPMD tool by then)
    - ...
  - `lib/`
    - `python/` (modules, e.g. for RT interfaces, pre- & post-processing; set `PYTHONPATH` here)
  - `etc/`
    - `picongpu` (previously: `picongpu/submit/`; we need to abstract the mpi call ⌛ and the profile✓; copied on `pic-create`)
      - tpl dirs
      - `openib.conf`
      - `cuda.filter`, `scorep.filter` (VT/Score-P filter)
  - `share/`
    - `picongpu/`
      - `man/`: `Slurm_Tutorial.rst`, `PBS_Tutorial.rst`
      - `examples/` (root with same structure as `/`)
        - e.g. `LaserWakefield/`
          - `bin/` (e.g. example-specific plot scripts with ranges, *not* modules)
          - `include/picongpu/simulation_defines/...` (param files)
          - `etc/` (cfg files; previously `submit/`)
      - things from `src/tools/share/`
  - `bin/` (set `PATH` here)
    - `pic-create`
    - `pic-configure`
    - `pic-compile`
    - `cuda_memtest.sh` (copied on `pic-create` to example's `bin/` & executed in `tbg` via absolute path)
    - things from `src/tools/bin`
  - `docs/` (as before; or move to `lib/picongpu/docs/` and builds to `share/picongpu/docs/`)

ax3l:
@ComputationalRadiationPhysics/picongpu-developers @ComputationalRadiationPhysics/picongpu-maintainers as mentioned offline, we are now starting to implement this new structure.

n01r:
Awesome work!

bussmann:
Fantastic!

BrianMarre:
@PrometheusPi is this issue still relevant?

PrometheusPi:
This has been implemented 2017

PR #2058, opened 2017-05-22
Prepare for Python Modules

ax3l:
Prepare repo structure for more sorted inclusion of python modules. This is still half-way to #1968 but works for now.

Important to know for now:

- `src/tools/`: pre & post-processing for PIConGPU; could become separate repo one day
  - `lib/python/`: path for python *modules* (set `PYTHONPATH` here)
- `examples/SomeName/`
  - `bin/`: example-specific (!) plot scripts that are executable (e.g. using a module and setting ranges)
  - `lib/python/`: example-specific (!) python module (seldom needed)

Also fixes the wrong-named `scripts/` dir in `picongpu` to be named as in install, `bin/`


psychocoderHPC:
@ax3l Feel free to add a `README.txt` to the new folders with the short description from this PR

psychocoderHPC:
We need to update [this line](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2058/files#diff-baaba04b0773a6bcf8a6d1c04ee53f5fR334) and add the folder `lib`

ax3l:
can do, but that's old `make install` logic when working with `pic-create` which we are not using anymore and should be removed before adding an actual pre-install

pic-create
```diff
@@ -24,7 +24,7 @@ this_dir=`dirname $0`
 #only copy submit and params if we clone from default pic folder
 default_folder_to_copy="submit include/simulation_defines/param include/simulation_defines/unitless"
 #if we clone a project we copy full include 
-folder_to_clone="submit include"
+folder_to_clone="submit include lib/python"
```
psychocoderHPC: `lib/python` must be `lib` only. Rsync is going recursive throw all folder in lib

ax3l: uh, and also `bin` for scripts of an example!

src/picongpu/submit/hypnos-hzdr/picongpu.profile.example
```diff
@@ -45,6 +45,8 @@ export PATH=$PATH:$PICSRC
 export PATH=$PATH:$PICSRC/src/splash2txt/build
 export PATH=$PATH:$PICSRC/src/tools/bin
 
+export PYTHONPATH=$PICSRC/src/tools/lib/python:$PYTHONPATH
```
ax3l: @ComputationalRadiationPhysics/picongpu-developers please add this line to your `picongpu.profile` if you are working on `dev`

PR #2061, opened 2017-05-22
lockstep refactoring of `KernelCountParticles`

psychocoderHPC:
- add documentation

## Tests

- [x] KHI with resource plugin (is using the changed kernel)

### DIFF
--- a/src/libPMacc/include/particles/operations/CountParticles.hpp
+++ b/src/libPMacc/include/particles/operations/CountParticles.hpp
@@ -67,7 +67,7 @@ struct KernelCountParticles
         T_PBox & pb,
         uint64_cu* gCounter,
         T_Filter & filter,
-        T_Mapping & mapper
+        T_Mapping const & mapper
     ) const
     {
         using namespace mappings::threads;
@@ -145,12 +145,11 @@ struct KernelCountParticles
                 {
                     if( linearIdx < particlesInSuperCell )
                     {
-                        if(
-                            filter(
-                                *frame,
-                                linearIdx
-                            )
-                        )
+                        bool const useParticle = filter(
+                            *frame,
+                            linearIdx
+                        );
+                        if( useParticle )
                             nvidia::atomicAllInc( &counter );
                     }
                 }
@@ -171,8 +170,6 @@ struct KernelCountParticles
 
             __syncthreads( );
         }
-
-        __syncthreads( );
 
         onlyMaster(
             [&](
--- a/src/libPMacc/include/particles/operations/CountParticles.hpp
+++ b/src/libPMacc/include/particles/operations/CountParticles.hpp
@@ -21,7 +21,6 @@
 
 #pragma once
 
-
 #include "pmacc_types.hpp"
 #include "memory/buffers/GridBuffer.hpp"
 #include "mappings/kernel/AreaMapping.hpp"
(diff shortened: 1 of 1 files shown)


src/libPMacc/include/particles/operations/CountParticles.hpp
```diff
@@ -31,72 +30,163 @@
...
+        using Frame = typename T_PBox::FrameType;
+        using FramePtr = typename T_PBox::FramePtr;
+        constexpr uint32_t dim = T_Mapping::Dim;
+        constexpr uint32_t frameSize = math::CT::volume< typename Frame::SuperCellSize >::type::value;
+        constexpr uint32_t numWorkers = T_numWorkers;
 
-        PMACC_SMEM( frame, FramePtr );
-        PMACC_SMEM( counter, int );
-        PMACC_SMEM( particlesInSuperCell, lcellId_t );
+        PMACC_SMEM(
```
ax3l: its quite weird to write this `__shared__ frame = FramePtr;` assignment in such a way...
can't we make it
```C++
auto frame = PMACC_SMEM( FramePtr );
```
? :D

src/libPMacc/include/particles/operations/CountParticles.hpp
```diff
@@ -31,72 +30,163 @@
...
+     * @param filter functor to filter particles those should be counted
+     * @param mapper functor to map a block to a supercell
+     */
+    template<
+        typename T_PBox,
+        typename T_Filter,
+        typename T_Mapping
+    >
+    DINLINE void operator( )(
+        T_PBox & pb,
```
ax3l: const ref?

src/libPMacc/include/particles/operations/CountParticles.hpp
```diff
@@ -31,72 +30,163 @@
...
         uint64_cu* gCounter,
-        Filter filter,
-        Mapping mapper
+        T_Filter & filter,
```
ax3l: const ref?

src/libPMacc/include/particles/operations/CountParticles.hpp
```diff
@@ -31,72 +30,163 @@
...
+        T_Mapping & mapper
```
ax3l: const ref?

src/libPMacc/include/particles/operations/CountParticles.hpp
```diff
@@ -31,72 +30,163 @@
...
+            forEachParticle(
+                [&](
+                    uint32_t const linearIdx,
+                    uint32_t const idx
+                )
+                {
+                    if( linearIdx < particlesInSuperCell )
+                    {
+                        if(
+                            filter(
```
ax3l: storing the result in a const first and then `if`-ing that const might increase readability :)

src/libPMacc/include/particles/operations/CountParticles.hpp
```diff
@@ -31,72 +30,163 @@
...
+                {
+                    frame = pb.getPreviousFrame( frame );
+                    particlesInSuperCell = frameSize;
+                }
+            );
+
+            __syncthreads( );
         }
+
+        __syncthreads( );
```
ax3l: not sure if both are really needed (returns above)

src/libPMacc/include/particles/operations/CountParticles.hpp
```diff
@@ -31,72 +30,163 @@
...
+     * @param mapper functor to map a block to a supercell
+     */
+    template<
+        typename T_PBox,
+        typename T_Filter,
+        typename T_Mapping
+    >
+    DINLINE void operator( )(
+        T_PBox & pb,
         uint64_cu* gCounter,
```
ax3l: not sure why we don't design this kernel as a pretty `uint64_t &`, that's messy downstream on all places

src/libPMacc/include/particles/operations/CountParticles.hpp
```diff
@@ -113,23 +203,33 @@ struct CountParticles
      * @return number of particles in defined area
      */
     template<uint32_t AREA, class PBuffer, class Filter, class CellDesc>
-    static uint64_cu countOnDevice(PBuffer& buffer, CellDesc cellDescription, Filter filter)
+    static uint64_cu countOnDevice( PBuffer& buffer, CellDesc cellDescription, Filter filter )
```
ax3l: - `PBuffer& buffer`: const ref?
- `cellDescription`: const?
- `Filter`: const?
- signature: return as `uint64_t`?

src/libPMacc/include/particles/operations/CountParticles.hpp
psychocoderHPC: - atomic operation are not defined for `uint64_t`
- `uint64_cu* gCounter` is the return value of the kernel, therefore a reference is not allowed

src/libPMacc/include/particles/operations/CountParticles.hpp
psychocoderHPC: `T_Filter` can not be const else the internal state can not be changed.

src/libPMacc/include/particles/operations/CountParticles.hpp
psychocoderHPC: Offline discussed: This example is one reason why we can't use this syntax :-(
The example is wrong and must be
```C++
auto & frame = PMACC_SMEM( FramePtr );
```

This is a common mistake and is very hard to debug. 
The factor `PMACC_SMEM` returns a reference to the object in the shared memory.

src/libPMacc/include/particles/operations/CountParticles.hpp
psychocoderHPC: true, can be removed

src/libPMacc/include/particles/operations/CountParticles.hpp
ax3l: signature: yes we can (and should; offline result) change this via a static cast before of the var we return.

note: please not in this PR, this would change to much downstream code, refactor in a follow-up

src/libPMacc/include/particles/operations/CountParticles.hpp
theZiz: I would suggest keeping the old formating:
```C++
if ( filter( *frame, linearThreadIdx ) )
```
I know, this breaks some new rules, but just looks the best...

src/libPMacc/include/particles/operations/CountParticles.hpp
psychocoderHPC: Can't add const here because `ParticlesBox` is not using const for it's member functions.

psychocoderHPC: I added a issue for this task #2065

src/libPMacc/include/particles/operations/CountParticles.hpp
psychocoderHPC: issue #2066

PR #2063, opened 2017-05-23
Warning in Socket Write

### LINKED ISSUE #2062: warning: SocketConnector return not used
While compiling the current `dev` on taurus I got the following warning:
```
.../picongpu/src/picongpu/include/plugins/output/sockets/SocketConnector.hpp(112): warning: result of call is not used
```

This code has not changed since the alpha release. Do we want to fix this warning? 

System details:
 - cmake/3.3.1
 - cuda/8.0.44
 - gcc/5.3.0
 - bullxmpi/1.2.8.4
 - boost/1.60.0-gnu5.3

ax3l:
Handle the return code of the [write to a socket](http://kirste.userpage.fu-berlin.de/chemnet/use/info/libc/libc_11.html).
Close #2062

### DIFF
--- a/src/picongpu/include/plugins/output/sockets/SocketConnector.hpp
+++ b/src/picongpu/include/plugins/output/sockets/SocketConnector.hpp
@@ -109,7 +109,9 @@ public:
             size_t zipedSize = zip.compress(tmp + MessageHeader::bytes, ((char*) array) + MessageHeader::bytes, size - MessageHeader::bytes, 6);
             MessageHeader* header = (MessageHeader*) tmp;
             header->data.byte = (uint32_t) zipedSize;
-            write(SocketFD, tmp, zipedSize + MessageHeader::bytes);
+            int nbytes = write(SocketFD, tmp, zipedSize + MessageHeader::bytes);
+            if (nbytes < 0)
+                perror("a socket write error occured");
             __deleteArray(tmp);
         }
     }


PrometheusPi:
I compiled this pull request on taurus and the warning disappeared as expected.

src/picongpu/include/plugins/output/sockets/SocketConnector.hpp
```diff
@@ -109,7 +109,9 @@ class SocketConnector
             size_t zipedSize = zip.compress(tmp + MessageHeader::bytes, ((char*) array) + MessageHeader::bytes, size - MessageHeader::bytes, 6);
             MessageHeader* header = (MessageHeader*) tmp;
             header->data.byte = (uint32_t) zipedSize;
-            write(SocketFD, tmp, zipedSize + MessageHeader::bytes);
+            int nbytes = write(SocketFD, tmp, zipedSize + MessageHeader::bytes);
+            if (nbytes < 0)
+                perror("a socket write error occured");
```
psychocoderHPC: I would prefer the C++ method `std::err` over the `C` function call `perror()` .

Please extent the error message with more information: `a socket write error occured in SocketConnector::send`

psychocoderHPC: offline discussed: `perror` is also used in the rest of the class.

Never the less this code is not used anymore therefore we will remove this class in the near future.

PR #2069, opened 2017-05-24
Abort on CUDA8 + Boost 1.64.0

ax3l:
Until we find a fix for Boost 1.64.0 and CUDA 8.0, we throw a clean abort at configure time to inform the user.

See #2048

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -92,7 +92,7 @@ zlib
 
 boost
 """""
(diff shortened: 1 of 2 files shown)


ax3l:
> but this also prevents use of boost 1.64.0 with CUDA 7 etc. - is this Okay ?

@PrometheusPi no, it does not. only on 1.64.0 *and* CUDA 8.0

PrometheusPi:
@ax3l Thanks for the clarification. Yes you are completely right - only the `*.rst` does not state it. The `CMake` file only prevents it for CUDA 8.

PR #2080, opened 2017-05-31
Tool: New Version

ax3l:
Adds a tool for maintainer ([see wiki](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/Maintainer-Documentation)) to update the version number of PIConGPU on all places during releases.

### DIFF
--- a/src/tools/bin/newVersion.sh
+++ b/src/tools/bin/newVersion.sh
@@ -1,0 +1,129 @@
+#!/usr/bin/env bash
+#
+# Copyright 2017 Axel Huebl
+#
+# This file is part of PIConGPU.
+#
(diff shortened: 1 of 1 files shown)


src/tools/bin/newVersion.sh
```diff
@@ -0,0 +1,127 @@
...
+echo
+echo "Your current source directory is: $REPO_DIR"
+echo
+
+read -p "Are you sure you want to continue? [y/N] " -r
+echo
+
+if [[ ! $REPLY =~ ^[Yy]$ ]]
+then
+    exit 1
```
psychocoderHPC: an error message would be usefull

Issue #2092, opened 2017-06-06
Docs Basics-section


wmles:
- what is $SCRATCH, no example given and no comment, that I can safely put it in /home for the tutorial
 - section about cloning source code is apparently missing (therefore no pic-create)

PrometheusPi:
Welcome to PIConGPU @wmles and thank you for the report on the documentation.
I will fix it as soon as possible.

PR #2093, opened 2017-06-06
Docs: Extended $SCRATCH Info

PrometheusPi:
This pull request addresses issues brought up by @wmles in #2092 

 - It adds a section describing how to download the PIConGPU source code (otherwise all the `pic-config` etc. used later commands won't work)
 - It adds extra information on $SCRATCH for users on their laptop.

### DIFF
--- a/USAGE.rst
+++ b/USAGE.rst
@@ -15,6 +15,8 @@ Preparation
 First, decide where to store input files, a good place might be ``$HOME`` (``~``) because it is usually backed up.
 Second, decide where to store your output of simulations which needs to be placed on a high-bandwidth, large-storage file system which we will refer to as ``$SCRATCH``.
 
+For a first test you can also use your home directory ``$HOME`` as simulation directory ``$SCRATCH``.
+
 As in our :ref:`compiling from source <install-source>` section, we need a few directories to structure our workflow:
 
 .. code-block:: bash


ax3l:
Thanks!

the picongpu download alone is not enough and the page you change starts with
  see also: "install dependencies and set environment" which does that properly. 

Please remove that new block with git clone completely, all other changes are OK. Please squash the rest.

ax3l:
On the user question in general: I find

> Second, decide where to store your output of simulations which needs to be placed on a high-bandwidth, large-storage file system which we will refer to as ``$SCRATCH``.

Is very well defined and mentioned in this place... This is far from undefined.

ax3l:
In case you are introducing that user: new users should not jump into chapters they seem to fit but should read the manual from the front chapter by chapter. 

(this is also mentioned in the first page of the manual) 

Not having an environment set up which comes way before the basic usage seems to me like someone is jumping through the manual and complaining he missed points that way, tbh. All sections have a structure, intentional minimal (!)  repetition and back links - those were ignored when posting "git clone is missing".

PrometheusPi:
@ax3l I agree with you that strictly everything required is mentioned. However, the `git clone` part is only given once in [dependencies](https://picongpu.readthedocs.io/en/dev/install/dependencies.html) under sub-point 9. If someone uses hypnos (or any other prepared cluster), he/she will skip that long decency list and thus miss the point of downloading the PIConGPU code (as happened here). 

Thus I would vote for either a recapitulation of what should/has been done to get to code - or a [direct reference to that part in dependencies](https://picongpu.readthedocs.io/en/dev/install/dependencies.html#picongpu-source-code). 

Furthermore I think we should move PIConGPU source code out of the PIConGPU dependency list. That it depends on its own code should be clear 😀.

ax3l:
> If someone uses hypnos (or any other prepared cluster), he/she will skip that long decency list and thus miss the point of downloading the PIConGPU code (as happened here).

This is a special case and if someone shows you that *undocumented* install path, then this person has the responsibility.

Indeed, we already mention we have part for HPC systems prepared, and we do so in the right way, the [*Dependencies chapter*](https://picongpu.readthedocs.io/en/dev/install/dependencies.html). It explicitly says: *See the following chapter if you can skip this step fully __or in part__ by loading existing modules on those systems.*

> or a direct reference to that part in dependencies.

that *is* already there and it is highlighted on top.

Until fully is given, which depends on several issues which are documented here, we can really really not take the responsibility *to read a full page from the top* from users.

> Furthermore I think we should move PIConGPU source code out of the PIConGPU dependency list. That it depends on its own code should be clear :grinning:.

It's not, it's a download like every other and it needs a huge bunch of environment vars like any other. It is purely technical, adds not understanding and should be prepared the same way as any other header-only dependency is prepared once.

In the future we will have modules and images for that in HPC environments, which just stresses the point that it is no difference then for instance *boost*.

USAGE.rst
```diff
@@ -14,6 +14,7 @@ Preparation
 
 First, decide where to store input files, a good place might be ``$HOME`` (``~``) because it is usually backed up.
 Second, decide where to store your output of simulations which needs to be placed on a high-bandwidth, large-storage file system which we will refer to as ``$SCRATCH``.
+(For a first test you can also use your home directory as simulation directory. Just make sure the available disk space is large enough.)
```
ax3l: I am not a big fan of making a manual verbose when it is clear. Nevertheless, if you want to be overexplicit then let us:

- add a newline to make a new paragraph
- No "()"
- Add keywords: For a first test you can also use your home directory `$HOME` as simulation directory `$SCATCH`. ~~Just make sure the available disk space is large enough.~~ -> Remove since it does not help if you are not saying *how large*.

USAGE.rst
```diff
@@ -29,6 +30,16 @@ As in our :ref:`compiling from source <install-source>` section, we need a few d
     # PIConGPU simulation output
     mkdir $SCRATCH/runs
 
+Download the PIConGPU source code either from the main repository or from your fork.
```
ax3l: Please remove, this is *not* enough to get PIConGPU source, tools, environment, etc. set up. See the above linked and previously mentioned dependencies and picongpu.profile sections.

ax3l:
I will prepare a 2nd PR to update the instructions before the individual machines [here](https://picongpu.readthedocs.io/en/dev/install/profile.html#install-profile).

ax3l:
@PrometheusPi ping: please this one, too :)

PrometheusPi:
@ax3l Changes are included (now it's a minimal pull request)

Issue #1969, opened 2017-04-18
Install PIConGPU (Sources)


ax3l:
Follow up to #1968 

Create a new `/` CMakeLists.txt that can install the headers of PIConGPU in two modes:
- `install` (copy as usual)
- `develop` (symlinks of dirs)

as in [Python](http://stackoverflow.com/questions/19048732/python-setup-py-develop-vs-install#19048754).

- [ ] find a way to configure the path and name of the to-be-used `picongpu.profile` (ideally, each version and picongpu-dev should have their own `picongpu-<version>.profile`)

ax3l:
I am currently working on this.

ax3l:
I am currently working on this.

The CMake logic is as follows.

**the base dir** of the repo will contain a "super" project *PIConGPU* `CMakeLists.txt` that select components one wants to install. This will allow to install all sources (as before), or just part, e.g. `PMacc` or `tbg` or `tools`  or ... only.

The `include/pmacc/` dir will contain a regular `CMakeLists.txt` that can either be `add_subdirectory`-ed for usage or installed on which it creates a proper `PMaccConfig.cmake`. For that I will move logic from the config to the first file, where it belongs (it's auto-included on installed PMacc).

Sub-projects such as `mpiInfo`, `png2gas`, `splash2txt` and `cuda_memtest` can be build on install. If not, they are as today build on PIConGPU application build time, via `add_subdirectory`.

The PIConGPU application itself, CMake project name `PIConGPUapp` is in `include/picongpu/` as usual. It contains the nifty logic we use so far to create a binary from an input directory. It's "source install logic" ("target") would go in the base `CMakeLists.txt` instead of the `include/picongpu/CMakelists.txt` since the latter is used for the described "user-workflow" as usual.

All in all, this will not change anything for the user but will allow installing PIConGPU sources, sub-projects of the PIConGPU repo and pre-compiling sub-projects if wanted.

PR #2095, opened 2017-06-06
Profile Preparations

ax3l:
For the `picongpu.profile` files already listed, note which preparations are necessary to be installed. (See comment in #2093, report in #2092)

In the future #1969, PIConGPU's source will be a module on HPC systems, too. (Alternatively, a `spack install` command or similar.)

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -119,8 +119,11 @@ rsync
 - *Arch Linux:* ``sudo pacman --sync rsync``
 - *Spack:* ``spack install rsync``
 
-PIConGPU source code
-""""""""""""""""""""
+.. _install-dependencies-picongpu:
+
+PIConGPU Source Code
+^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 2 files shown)


PrometheusPi:
I think that might help with issue #2092

PrometheusPi:
Just a question:
Taking taurus as an example: The libraries `pngwriter` and `libsplash` need to be installed manually (but are not required) - should we mention this in this section too?

E.g.: 
The `picongpu.profile` loads all required dependencies but if you 
need hdf5 output or png snapshots please install  `pngwriter` and
 `libsplash` manually as described [here](https://picongpu.readthedocs.io/en/dev/install/dependencies.html) 

?

ax3l:
As I committed it now it should be fine: the *profile for Taurus* that we propose *needs* to have libSplash and PNGwriter installed beforehand.

We don't need to put *ifs* and more verbosity around it. If one wants to use *that profile* then it will provide something that we found useful and has certain requirements it mentions (which might be optional as a dependency in general, but that is not the point of this page which shows *production environments*).

ax3l:
note: pushed again with same content since the GitHub API seems not to have received the successful test state

PR #2096, opened 2017-06-08
Fix typo in ADK documentation

n01r:
When I wanted to write the Euler number (*exp(1)*) I wrote `\exp` instead of `mathrm{e}`.
This PR fixes that.


n01r:
merge me :3

PR #2104, opened 2017-06-16
lockstep refactoring of `KernelMoveAndMarkParticles`

psychocoderHPC:
- rename parameters
- add documentation

## Tests

- [x] 3D default KHI

### DIFF
--- a/src/picongpu/include/particles/Particles.kernel
+++ b/src/picongpu/include/particles/Particles.kernel
@@ -255,93 +255,186 @@ struct KernelManipulateAllParticles
     }
 };
 
-template< class BlockDescription_ >
+/** move over all particles
+ *
+ * Move frame-wise over all  and call a functor for each particle.
+ * This kernel is optimized for the particle push step and handles the
+ * special flag `mustShift` of the supercell to optimize the kernel shift particles
+ * in PMacc.
+ *
+ * @tparam T_numWorkers number of workers
+ * @tparam T_DataDomain PMacc::SuperCellDescription, compile time data domain
+ *                      description with a CORE and GUARD
+ */
+template<
+    uint32_t T_numWorkers,
+    typename T_DataDomain
+>
 struct KernelMoveAndMarkParticles
 {
-    template<class ParBox, class BBox, class EBox, class Mapping, class FrameSolver>
+    /** update all particles
+     *
+     * @tparam T_ParBox PMacc::ParticlesBox, particle box type
+     * @tparam T_EBox PMacc::DataBox, electric field box type
+     * @tparam T_BBox PMacc::DataBox, magnetic field box type
+     * @tparam T_ParticleFunctor particle functor type
+     * @tparam T_Mapping mapper functor type
+     *
+     * @param pb particle memory
+     * @param fieldE electric field data
+     * @param fieldB magnetic field data
+     * @param particleFunctor functor to manipulate (update) a particle
+     * @param mapper functor to map a block to a supercell
+     */
+    template<
+        typename T_ParBox,
+        typename T_EBox,
+        typename T_BBox,
+        typename T_ParticleFunctor,
+        typename T_Mapping
+    >
     DINLINE void operator()(
-       ParBox pb,
-       EBox fieldE,
-       BBox fieldB,
-       FrameSolver frameSolver,
-       Mapping mapper) const
-   {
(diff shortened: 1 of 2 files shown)


src/picongpu/include/particles/Particles.kernel
```diff
@@ -255,93 +255,186 @@ struct KernelManipulateAllParticles
     }
 };
 
-template< class BlockDescription_ >
+/** move over all particles
+ *
+ * Move frame-wise over all  and call a functor for each particle.
```
ax3l: Move frame-wise over a species and call a functor for each particle.

src/picongpu/include/particles/Particles.kernel
```diff
@@ -255,93 +255,186 @@ struct KernelManipulateAllParticles
...
+        typedef typename T_ParBox::FramePtr FramePtr;
+
+        DataSpace< simDim > const block(
+            mapper.getSuperCellIndex( DataSpace< simDim >( blockIdx ) )
+        );
+
+        // relative offset (in cells) to the supercell (including the guard)
+        DataSpace< simDim > const superCellOffset = block * SuperCellSize::toRT();
+
+         using ParticleDomCfg = IdxConfig<
```
ax3l: space

src/picongpu/include/particles/Particles.kernel
```diff
@@ -255,93 +255,186 @@ struct KernelManipulateAllParticles
...
+            typename T_BBox::ValueType
+        >( T_DataDomain() );
+        auto cachedE = CachedBox::create <
+            1,
+            typename T_EBox::ValueType
+        >( T_DataDomain( ) );
+
+        __syncthreads();
+
+        //end kernel if we have no frames
```
ax3l: space

src/picongpu/include/particles/Particles.kernel
```diff
@@ -255,93 +255,186 @@ struct KernelManipulateAllParticles
...
+            frameSize,
+            numWorkers
+        >;
+
+        PMACC_SMEM(
+            mustShift,
+            int
+        );
+
+        /* unique for each worker */
```
ax3l: this comment harms more then it helps, the declarations says that it is a pointer copy to worker-local mem (register)

src/picongpu/include/particles/Particles.kernel
```diff
@@ -255,93 +255,186 @@ struct KernelManipulateAllParticles
...
+            }
+        );
+
+        frame = pb.getLastFrame( block );
+        particlesInSuperCell = pb.getSuperCell( block ).getSizeLastFrame( );
+
+        auto cachedB = CachedBox::create <
+            0,
+            typename T_BBox::ValueType
+        >( T_DataDomain() );
```
ax3l: inconsistent `()` with 4 lines below ;)

src/picongpu/include/particles/Particles.kernel
```diff
@@ -255,93 +255,186 @@ struct KernelManipulateAllParticles
...
+
+        __syncthreads();
+
+        onlyMaster(
+            [&](
+                uint32_t const,
+                uint32_t const
+            )
+            {
+                /* set in SuperCell the mustShift flag which is a optimization
```
ax3l: a**n** optimization

src/picongpu/include/particles/Particles.kernel
```diff
@@ -255,93 +255,186 @@ struct KernelManipulateAllParticles
...
+                 * for shift particles and fillGaps
```
ax3l: maybe write the class names of the kernels/functors here?

ax3l: maybe add what this optimizes at the place where `mustShift` is defined in the supercell

src/picongpu/include/particles/Particles.kernel
theZiz: Well, imo it helps as in I think every kernel I saw until all workers worked on the same frame and now every worker has it's own framePtr -- which I don't understand btw...

ax3l: we will add an other comment on how it's processed below - but this comment here is confusing for readers since:
- the pointer is stored in a local var is obvious from the definition
- if that means the frame will be uniquely processed or not is not meant by this comment

PR #2111, opened 2017-06-30
fix wrong time step in reduction of field background after restart

PrometheusPi:
This pull request fixes a wrong background field reduction during restart. The checkpoints contains the background field of `currentStep -1` but we subtract the background field of `currentStep` and than add the background field of `currentStep -1`. This leads to a wrong field if the background field is not constant over time. 
**Thanks @psychocoderHPC for spotting this.**

I will check the results on TWTS as soon as possible.

Please wait with merging till the test are finished. 

CC-ing: @steindev, @BeyondEspresso

### DIFF
--- a/src/picongpu/include/simulationControl/MySimulation.hpp
+++ b/src/picongpu/include/simulationControl/MySimulation.hpp
@@ -466,10 +466,13 @@ public:
         {
             namespace nvfct = PMacc::nvidia::functors;
 
+            /* subtract the background field as it has been stored in the
+             * checkpoint: at its iteration step = currentStep - 1
+             */
             (*pushBGField)( fieldE, nvfct::Sub(), FieldBackgroundE(fieldE->getUnit()),
-                            step, FieldBackgroundE::InfluenceParticlePusher);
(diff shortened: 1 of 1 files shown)


src/picongpu/include/simulationControl/MySimulation.hpp
```diff
@@ -466,10 +466,13 @@ class MySimulation : public SimulationHelper<simDim>
         {
             namespace nvfct = PMacc::nvidia::functors;
 
+            /* subtract the background field as it has been stored in the
+             * checkpoint: at its iteration step = currentStep - 1
```
psychocoderHPC: white space at line end

ax3l:
@psychocoderHPC is still checking if this only happens on restarts at t=0 and breaks restarts at t>0 now

psychocoderHPC:
I checked this fix again and found that this fix fixes the wrong field background for the restart from time step 0 and add wrong restart handling for later time steps.
I opened #2113 to fully solve the issue and will close this PR.

PR #2112, opened 2017-06-30
fix plugin energy field reduce

PrometheusPi:
This pull request fixes a wrong MPI reduce in the field energy plugin.
It caused wrong field values (f.e. in TWTS runs)  because the mpi reduce reduces over two `float3_X[2]` instead of over just one `float3_X[2]` (which is equal two 2 `float3_X` - which was probably the intention of the developer).

**Thanks @psychocoderHPC for spotting this!**

Run time test show taht this solves the wrong output we see during TWTS runs. 
(This effects **all** heating tests)

**Effects so far:**
So far this caused mostly some massively increased `Ez` values. 
This **does not** effect the physics - just the output. 

## Test

- [x] test field energy with TWTS enabled

CC-ing: @steindev, @BeyondEspresso

### DIFF
--- a/src/picongpu/include/plugins/EnergyFields.hpp
+++ b/src/picongpu/include/plugins/EnergyFields.hpp
@@ -217,8 +217,8 @@ private:
         localReducedFieldEnergy[1] = reduceField(fieldE);
 
         mpiReduce(nvidia::functors::Add(),
-                  &globalFieldEnergy,
-                  &localReducedFieldEnergy,
+                  globalFieldEnergy,
+                  localReducedFieldEnergy,
                   2,
                   mpi::reduceMethods::Reduce());
 


psychocoderHPC:
All in all this is a out of memory access within the stack and could have unknown side effects.

src/picongpu/include/plugins/EnergyFields.hpp
```diff
@@ -219,7 +219,7 @@ class EnergyFields : public ISimulationPlugin
         mpiReduce(nvidia::functors::Add(),
                   &globalFieldEnergy,
                   &localReducedFieldEnergy,
-                  2,
+                  1, /* only 1 reduce because type is float3_x[2] */
```
psychocoderHPC: after thinking about the fix I think we should fix it the other way around, this would be more intuitive

```
globalFieldEnergy,
localReducedFieldEnergy,
2,
mpi::reduceMethods::Reduce());
```

PrometheusPi:
I tested this with the TWTS/plane wave setup and got the same result for both 1 GPU and 2 GPUs.

PR #2113, opened 2017-07-03
fix restart with background fields enabled

psychocoderHPC:
This PR replace #2111 and solves the issue by removing dirty detection of a restart by checking the simulation time step.

- PIConGPU: remove magic time step increase to detect restarts
- PMacc (SimulationHelper): use class member `restartRequested` to  detect a restart

## Tests
- [x] field background restart from step 0 and X, `background = currentStep + 1`

Cc-ing: @PrometheusPi, @steindev, @BeyondEspresso

### DIFF
--- a/src/libPMacc/include/simulationControl/SimulationHelper.hpp
+++ b/src/libPMacc/include/simulationControl/SimulationHelper.hpp
@@ -231,21 +231,18 @@ public:
             TimeIntervall tSimCalculation;
             TimeIntervall tRound;
             double roundAvg = 0.0;
+            Environment<>::get().SimulationDescription().setCurrentStep( currentStep );
+
+            /* Since in the main loop movingWindow is called always before the dump, we also call it here for consistency.
+             * This becomes only important, if movingWindowCheck does more than merely checking for a slide.
+             * TO DO in a new feature: Turn this into a general hook for pre-checks (window slides are just one possible action).
+             */
+            movingWindowCheck(currentStep);
 
             /* dump initial step if simulation starts without restart */
-            if (currentStep == 0)
+            if (!restartRequested)
             {
-                /* Since in the main loop movingWindow is called always before the dump, we also call it here for consistency.
-                This becomes only important, if movingWindowCheck does more than merely checking for a slide.
-                TO DO in a new feature: Turn this into a general hook for pre-checks (window slides are just one possible action). */
-                movingWindowCheck(currentStep);
                 dumpOneStep(currentStep);
-            }
-            else
-            {
-                currentStep--; //We dump before calculation, thus we must go one step back when doing a restart.
-                Environment<>::get().SimulationDescription().setCurrentStep( currentStep );
-                movingWindowCheck(currentStep); //If we restart at any step check if we must slide.
             }
 
             /* dump 0% output */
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
I verified this fix. I used a simulation without field solver and without particles. Added a backgrodun field per simulation step `return float3_X(float3_X::create(currentStep + 1));`
I restarted from the initial checkpoint `0` and `5` of `10`.

PrometheusPi:
I applied this fix to my code for running TWTS (actually only plane wave) and my particles still disappear. 
If I however avoid restart, they do not vanish!

*smocking gun* plots:
particles disappear:
n_macroparticles

field energy shows jumps:
fieldenergy

src/picongpu/include/simulationControl/MySimulation.hpp
```diff
@@ -462,7 +462,7 @@ class MySimulation : public SimulationHelper<simDim>
         auto fieldE = dc.get< FieldE >( FieldE::getName(), true );
         auto fieldB = dc.get< FieldB >( FieldB::getName(), true );
 
-        if( step != 0 )
+        if( this->restartRequested )
         {
             namespace nvfct = PMacc::nvidia::functors;
```
ax3l: if the intention was to keep the logic identical, I think be two background field below need to be called with `step + 1u` now

ax3l: on the other hand, our logic now runs `runOneStep(currentStep);` again for the *restart step*! so the above `step = this->restartStep;` should actually keep the `+1u`!

psychocoderHPC: The logic before was wrong. 
`step + 1u` is wrong and is equal to the old implementation.
The current version is tested and correct.

PrometheusPi: @psychocoderHPC Please see the tests in #2137 - especially the one without the fix - it shows an equivalent behavior as with this commit. 

I will try to reproduce your background field now, to test restarts.

ax3l: my first comment was just about that the PR description does not state that it fixes the time offset (only the title).

my second comment is, that the `+1u` needs to be added to line 484 to fix `dev`

src/libPMacc/include/simulationControl/SimulationHelper.hpp
```diff
@@ -231,22 +231,19 @@ class SimulationHelper : public IPlugin
             TimeIntervall tSimCalculation;
             TimeIntervall tRound;
             double roundAvg = 0.0;
+            Environment<>::get().SimulationDescription().setCurrentStep( currentStep );
```
ax3l: doubled with line 221

src/libPMacc/include/simulationControl/SimulationHelper.hpp
```diff
@@ -231,22 +231,19 @@ class SimulationHelper : public IPlugin
...
+
+            /* Since in the main loop movingWindow is called always before the dump, we also call it here for consistency.
+             * This becomes only important, if movingWindowCheck does more than merely checking for a slide.
+             * TO DO in a new feature: Turn this into a general hook for pre-checks (window slides are just one possible action).
+             */
+            movingWindowCheck(currentStep);
```
ax3l: note: his also adds the background fields again

PR #2117, opened 2017-07-05
add interfaces for functor and filter

psychocoderHPC:
Introduce a generic implementation for a functor. The functor is instantiated on the host side then passed to the kernel and on the device/accelerator a device instance is created.

This workflow allows to pass a generic functor to the device without defining a fixed interface for the `operator()` but keeps the possibility that the user can embed container or other members into the functor even if there is a strong dependency to the accelerator. A example is the random number generator in alpaka, the generator depends on the accelerator but the accelerator is not known before the kernel is executed on the device.

At runtime, the following logic happens, for a binary filtered functor (`PMacc::functor::Filtered`):
- call filter for the two arguments (`PMacc::filter::Interface`)
- combine both results via the filter operator (`PMacc::filter::operators`)
- call the functor if this combined filter result is true (`PMacc::functor::Interface`)

For unary functors, the typical filter operator is just a pass-through of the one argument (as long as one does not negate its result).

# Features

- add generic functor interface
- add define a filter interface
- add a filtered functor interface (confitional functor)
- add operators to combine filter results

# Examples

```C++
    // definition for a unary filtered functor
    template<
        typename T_UnaryFunctor,
        typename T_UnaryFilter = filter::IsValid
    >
    using IUnary = PMacc::functor::Filtered<
        PMacc::filter::operators::And,
        PMacc::filter::Interface<
            T_UnaryFilter,
            1u
        >,
        PMacc::functor::Interface<
            T_UnaryFunctor,
            1u,
            void
        >
    >;

    // definition for a binary filtered functor
    template<
        typename T_BinaryFunctor,
        typename T_UnaryFilter = filter::IsValid
    >
    using IBinary = PMacc::functor::Filtered<
        PMacc::filter::operators::And,
        PMacc::filter::Interface<
            T_UnaryFilter,
            1u
        >,
        PMacc::functor::Interface<
            T_BinaryFunctor,
            2u,
            void
        >
    >;
```

### DIFF
--- a/src/libPMacc/include/filter/Interface.hpp
+++ b/src/libPMacc/include/filter/Interface.hpp
@@ -1,0 +1,49 @@
+/* Copyright 2017 Rene Widera
+ *
+ * This file is part of libPMacc.
+ *
+ * libPMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * libPMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with libPMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "pmacc_types.hpp"
+#include "functor/Interface.hpp"
+
+
+namespace PMacc
+{
+namespace filter
+{
+
+    /** Interface for a filter
+     *
+     * @tparam T_UserFunctor PMacc::functor::Interface, type of the functor
+     * @tparam T_numArguments number which must be supported by T_UserFunctor
+     */
+    template<
+        typename T_UserFunctor,
+        uint32_t T_numArguments
+    >
+    using Interface = PMacc::functor::Interface<
+        T_UserFunctor,
+        T_numArguments,
+        bool
+    >;
+
+} // namespace filter
+} // namespace PMacc
--- a/src/libPMacc/include/filter/Interface.hpp
+++ b/src/libPMacc/include/filter/Interface.hpp
@@ -32,8 +32,13 @@ namespace filter
 
     /** Interface for a filter
      *
-     * @tparam T_UserFunctor PMacc::functor::Interface, type of the functor
-     * @tparam T_numArguments number which must be supported by T_UserFunctor
+     * A filter is a functor which is evaluated to true or false depending
+     * on the input parameters.
+     * A filter can be used to decide e.g. if a particle is located in a user
+     * defined area or if an attribute is above a threshold.
+     *
+     * @tparam T_UserFunctor PMacc::functor::Interface, type of the functor (filter rule)
+     * @tparam T_numArguments number of arguments which must be supported by T_UserFunctor
      */
     template<
         typename T_UserFunctor,
--- a/src/libPMacc/include/filter/operators/And.hpp
+++ b/src/libPMacc/include/filter/operators/And.hpp
@@ -47,7 +47,7 @@ namespace operators
         /** get AND combined result
          *
          * @param args arguments to combine
-         * @return the input argument
+         * @return AND combination of all arguments
          */
         template<
             typename T_Arg1,
--- a/src/libPMacc/include/filter/operators/And.hpp
+++ b/src/libPMacc/include/filter/operators/And.hpp
@@ -1,0 +1,65 @@
+/* Copyright 2017 Rene Widera
+ *
+ * This file is part of libPMacc.
+ *
+ * libPMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * libPMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with libPMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+
+namespace PMacc
+{
+namespace filter
+{
+namespace operators
+{
+
+    //! combine all arguments by AND `&&`
+    struct And
+    {
+        /** return a
+         *
+         * @param a a boolean value
+         * @return the input argument
+         */
+        template< typename T_Arg >
+        DINLINE bool
+        operator()( T_Arg const a ) const
+        {
+            return a;
+        }
+
+        /** get AND combined result
+         *
+         * @param args arguments to combine
+         * @return the input argument
+         */
+        template<
+            typename T_Arg1,
+            typename ... T_Args
+        >
+        DINLINE bool
+        operator()( T_Arg1 const a, T_Args const ... args ) const
+        {
+            return a && And{}( args ... );
+        }
+    };
+
+} // namespace operators
+} // namespace filter
+} // namespace PMacc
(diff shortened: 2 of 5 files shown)


src/libPMacc/include/filter/Interface.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+namespace PMacc
+{
+namespace filter
+{
+
+    /** Interface for a filter
+     *
+     * @tparam T_UserFunctor PMacc::functor::Interface, type of the functor
+     * @tparam T_numArguments number which must be supported by T_UserFunctor
```
theZiz: maybe replace *"number"* with *"number of arguments"*, sounds better for me

src/libPMacc/include/filter/operators/And.hpp
```diff
@@ -0,0 +1,65 @@
...
+        DINLINE bool
+        operator()( T_Arg const a ) const
+        {
+            return a;
+        }
+
+        /** get AND combined result
+         *
+         * @param args arguments to combine
+         * @return the input argument
```
theZiz: Not really? It returns the AND combined input arguments. ;)

src/libPMacc/include/filter/operators/Or.hpp
```diff
@@ -0,0 +1,65 @@
...
+        DINLINE bool
+        operator()( T_Arg const a ) const
+        {
+            return a;
+        }
+
+        /** get OR combined result
+         *
+         * @param args arguments to combine
+         * @return the input argument
```
theZiz: Not really? It returns the OR combined input arguments. ;)

src/libPMacc/include/functor/Filtered.hpp
```diff
@@ -0,0 +1,255 @@
...
+        typename T_FilterOperator,
+        typename T_Filter,
+        typename T_Functor
+    >
+    struct Filtered;
+
+    /** specialization of Filtered (with unary filter)
+     *
+     * This specialization can only used if T_Filter is of the type PMacc::filter::Interface
+     * and T_Functor is of the type PMacc::filter::Interface
```
theZiz: Is the functor really of the filter interface, not a functor interface?

src/libPMacc/include/functor/Filtered.hpp
```diff
@@ -0,0 +1,255 @@
...
+                    workerCfg
+                ),
+                ( *reinterpret_cast< Functor * >( this ) )(
+                    domainOffset,
+                    workerCfg
+                )
+            );
+        }
+
+        /** get the of the filtered functor
```
theZiz: get the what?

src/libPMacc/include/functor/Interface.hpp
```diff
@@ -0,0 +1,239 @@
...
+            return ( *reinterpret_cast< UserFunctor * >( this ) )( args ... );
+        }
+    };
+
+} // namespace acc
+
+    /** Interface for a user functor
+     *
+     * @tparam T_UserFunctor user functor type
+     * @tparam T_numArguments number which must be supported by T_UserFunctor
```
theZiz: number of arguments I guess?

src/libPMacc/include/functor/Interface.hpp
```diff
@@ -0,0 +1,239 @@
...
+        /** create a functor which can be used on the accelerator
+         *
+         * @tparam T_OffsetType type to describe the size of a domain
+         * @tparam T_numWorkers number of workers
+         *
+         * @param domainOffset offset to the origin of the local domain
+         *                     This can be e.g a supercell or cell offset and depends
+         *                     of the context where the interface is specialized.
+         * @param workerCfg configuration of the worker
+         * @return a instance of the user functor wrapped by the accelerator
```
theZiz: a**n**

src/libPMacc/include/functor/Filtered.hpp
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        /** execute the functor depending of the filter result
+         *
+         * Call the filter for each argument and if the combined result is true
+         * the user functor is called.
+         *
+         * @param args arguments passed to the functor
+         *             - the filter is executed for each argument
+         *             - the functor get passed all arguments
```
ax3l: arguments passed to the functor if the filter results of each argument evaluate to true when combined

src/libPMacc/include/functor/Filtered.hpp
```diff
@@ -0,0 +1,255 @@
...
+        -> void
+        {
+            if(
+                T_FilterOperator{ }(
+                    ( *reinterpret_cast< Filter * >( this ) )( args ) ...
+                )
+            )
+                ( *reinterpret_cast< Functor * >( this ) )( args ... );
+        }
+    };
```
ax3l: Keep what you have now, but for future reference to myself, this is what this operator() implements:

```C++
// needed since filter is not implemented as a member var to save 1 byte
auto filter = *reinterpret_cast< Filter * >( this );
T_FilterOperator filterOperator{};

// call the filter on each argument and combine the results
bool const combinedResult = filterOperator( filter( args ) ... );

if( combinedResult )
{
    // needed since functor is not implemented as a member var to save 1 byte
    auto functor = *reinterpret_cast< Functor * >( this );
    functor( args ... );
}
```

src/libPMacc/include/functor/Filtered.hpp
```diff
@@ -0,0 +1,255 @@
...
+    template<
+        typename T_FilterOperator,
+        typename T_Filter,
+        typename T_Functor
+    >
+    struct Filtered;
+
+    /** specialization of Filtered (with unary filter)
+     *
+     * This specialization can only used if T_Filter is of the type PMacc::filter::Interface
```
ax3l: *be* used

src/libPMacc/include/functor/Filtered.hpp
ax3l: Add: `. An unary filters means that each argument can only pass the same filter check before its results are combined.`

psychocoderHPC: copy past failure ^^

src/libPMacc/include/filter/Interface.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include "functor/Interface.hpp"
+
+
+namespace PMacc
+{
+namespace filter
+{
+
+    /** Interface for a filter
+     *
```
psychocoderHPC: done

psychocoderHPC:
updated

src/libPMacc/include/filter/Interface.hpp
```diff
@@ -32,8 +32,13 @@ namespace filter
 
     /** Interface for a filter
      *
-     * @tparam T_UserFunctor PMacc::functor::Interface, type of the functor
-     * @tparam T_numArguments number which must be supported by T_UserFunctor
+     * A filter is a functor which is evaluated to true or false depending
+     * on the input parameters.
+     * A filter can be used to decide e.g. if a particle is located in a user
+     * defined area or if a attribute is above a threshold.
```
theZiz: a**n** attribute

src/libPMacc/include/functor/Filtered.hpp
```diff
@@ -110,8 +109,10 @@ namespace acc
 
     /** specialization of Filtered (with unary filter)
      *
-     * This specialization can only used if T_Filter is of the type PMacc::filter::Interface
-     * and T_Functor is of the type PMacc::filter::Interface
+     * This specialization can only be used if T_Filter is of the type PMacc::filter::Interface
+     * and T_Functor is of the type PMacc::functor::Interface.
+     * An unary filters means that each argument can only pass the same filter
```
theZiz: A~~n~~ unary filter~~s~~

Furthermore I don't understand what you want to tell me with the sentence.

ax3l: nope, an because u is a vowel.

Filters are functors with boolean return value. An unary filter is an unary functor. That means if you want to run a functor on several arguments, an unary filter can only pass one argument after the other and chain the results to an other bool which then decides, if the (then filtered) functor can be executed.

Admittedly, also took me a while...

psychocoderHPC: @theZiz Is everything fine or should I change something?

src/libPMacc/include/functor/Interface.hpp
```diff
@@ -0,0 +1,239 @@
...
+namespace functor
+{
+namespace acc
+{
+namespace detail
+{
+    /** Helper class to compare void with void
+     *
+     * std::is_same not allows use void as type, by wrapping the type before
+     * compare we can workaround this limitation.
```
theZiz: This sound a bit rough. How about
> std::is_same does not allow to use void as type. By wrapping the type before
> comparing we can workaround this limitation.

psychocoderHPC: updated

PR #2121, opened 2017-07-06
Esirkepov: reorder code order

psychocoderHPC:
This PR adds intermediate results to the core routines to reduce the computation. The optimizations are copied from the EmZ Esirkepov solver. Currently it is not possible to use a shared version of the Esirkepov core deposition algorithm to both solvers to reduce the code duplication's.

- use the EmZ Esirkepov implementation as base to reorder the Esirkepov code
- use helper variables to avoid multiple calculations of the some value
- add `S1()` helper method

This pull request increase the speed of PIConGPU in the KHI example with positron/ electrons 3D by a factor of 1.05.

# Tests
I cherry picked this PR for the tests to the branch for the emz paper ( picongpu_0.3.0+#2112), because there I have already all results to validate the solver changes.

- [x] KHI 3D/2D electrons/positrons X-direction flow (all shapes)
- [x] KHI 3D/2D electrons/positrons Z-direction flow (all shapes)

### DIFF
--- a/src/picongpu/include/fields/currentDeposition/Esirkepov/Esirkepov.hpp
+++ b/src/picongpu/include/fields/currentDeposition/Esirkepov/Esirkepov.hpp
@@ -147,6 +147,11 @@ struct Esirkepov<T_ParticleShape, DIM3>
         constexpr int begin = -currentLowerMargin + 1;
         constexpr int end = begin + supp;
 
+        /* We multiply with `cellEdgeLength` due to the fact that the attribute for the
+         * in-cell particle `position` (and it's change in DELTA_T) is normalize to [0,1)
+         */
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
# TSC

## heating 3D

esi_reorder_heat

## charge conservation

esi_reorder_charge

psychocoderHPC:
# PCS X-direction flow

## heating 3D

esi_reorder_heat_pcs

## charge conservation

esi_reorder_charge_pcs

ax3l:
memo to ourselves: the KHI test could cover more aspects if we set up non-cubic cells.

psychocoderHPC:
> memo to ourselves: the KHI test could cover more aspects if we set up non-cubic cells.

Would it be in this case also useful to propagate the particles over the diagonals?

ax3l:
yes, as long as they are a lot of sub-relativistic particles as in this test case and enough turbulence, enough will leave and enter cells in any way possible.

PR #2122, opened 2017-07-06
Energy PostProcessing: Python Modules

n01r:
Adds Python modules for
- energy histogram and
- particle calorimeter plugin.

Currently used in Bremsstrahlung example #1979 
- [ ] retest the plot scripts in `Bremsstrahlung` example when this PR is ready

### DIFF
--- a/src/tools/lib/python/energy_histogram.py
+++ b/src/tools/lib/python/energy_histogram.py
@@ -27,55 +27,53 @@ import h5py
 
 class EnergyHistogram:
     """
-    Class to help plotting data fromt the energyHistogram plugin.
+    Class to help plotting data from the energyHistogram plugin.
     """
     def __init__(
             self,
-            species_name="ph",
+            species_name="e",
+            species_filter="",
             sim="./",
             period=1000,
             bin_count=1024,
             min_energy=10,
-            max_energy=10000,
-            distance_to_detector=0,
-            slit_detector_x=None,
-            slit_detector_z=None):
+            max_energy=10000
+        ):
         """
         Parameters
         ----------
         species_name : string
             name of the species for which the energy histogram should be
-            plotted, e.g. "ph"
+            plotted (default: "e")
+        species_filter : string
+            file suffix for plugin output created by a run over a subselection
+            of the species according to an optional filter argument
+            (default: "")
         sim : string
-            path to simOutput directory
+            path to simOutput directory (default: "./")
         period : unsigned integer
-            output periodicity of the histogram
+            output periodicity of the histogram (default: 1000)
         bin_count : unsigned integer
-            energy resolution of the histogram (number of bins)
+            energy resolution of the histogram (number of bins) (default: 1024)
         min_energy : float
-            minimum detectable energy in keV
+            minimum detectable energy in keV (default: 10)
         max_energy : float
-            maximum detectable energy in keV
-        distance_to_detector : float
-            distance to detector in y-direction in meters (if == 0: all
-            particles are considered)
-        slit_detector_x : float
-            slit opening in x in meters
-        slit_detector_z : float
-            slit opening in z in meters
+            maximum detectable energy in keV (default: 10000)
         """
         self.species_name = species_name
+        self.species_filter = species_filter
         self.sim = sim
         self.period = period
         self.bin_count = bin_count
         self.min_energy = min_energy
         self.max_energy = max_energy
-        self.distance_to_detector = distance_to_detector
-        self.slit_detector_x = slit_detector_x
-        self.slit_detector_z = slit_detector_z
 
-        self.histFile = "{}/simOutput/{}_energyHistogram.dat".format(
-            sim, species_name)
+        filter_fname = ""
+        if self.species_filter != "":
+            filter_fname = "_"+ species_filter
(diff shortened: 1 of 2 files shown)


src/tools/lib/python/energy_histogram.py
```diff
@@ -0,0 +1,197 @@
...
+        species_name : string
+            name of the species for which the energy histogram should be
+            plotted (default: "e")
+        species_filter : string
+            file suffix for plugin output created by a run over a subselection
+            of the species according to an optional filter argument
+            (default: "")
+        sim : string
+            path to simOutput directory (default: "./")
+        period : unsigned integer
```
ax3l: the defaults of those four variables should be `=None` and auto-detected.

Only if the user chooses to change the limits it should be required (the user can't change the resolution or periodicity anyway) after the sim

src/tools/lib/python/particle_calorimeter.py
```diff
@@ -0,0 +1,300 @@
...
+        species_name : string
+            name of the species for which the particle calorimeter
+            should be plotted (default: "e")
+        species_filter : string
+            file suffix for plugin output created by a run over a subselection
+            of the species according to an optional filter argument
+            (default: "")
+        sim : string
+            path to simOutput directory (default: "./")
+        period : unsigned integer
```
ax3l: should be auto-detected and only necessary if the user wants just a subset of the outputs that are available

src/tools/lib/python/particle_calorimeter.py
```diff
@@ -0,0 +1,300 @@
...
+            output periodicity of the histogram (default: 1000)
+        num_bins_yaw : unsigned integer
+            number of bins resolving the yaw angle (default: 64)
+        num_bins_pitch : unsigned integer
+            number of bins resolving the pitch angle (default: 64)
+        num_bins_energy : unsigned integer
+            number of bins resolving the detected energies (default: 1024)
+        min_energy : float
```
ax3l: same as above

src/tools/lib/python/particle_calorimeter.py
```diff
@@ -0,0 +1,300 @@
...
+        # error message
+        if np.sum(self.hist_data) <= 0:
+            print(
+                "There is no valid data for the histogram. "+
+                "You may have forgotten to execute `energy_histogram` first."
+            )
+            return
+
+        # unit conversion: Joules to eV
+        J2EV = sc.electron_volt
```
ax3l: `J2eV`

src/tools/lib/python/particle_calorimeter.py
```diff
@@ -0,0 +1,300 @@
...
+        # unit conversion: simUnit to EV
+        simUnitEV = self.sim_unit / J2EV
```
ax3l: `simUnit_eV` (or even `sim_unit_eV`): python does not camelCase in vars

src/tools/lib/python/particle_calorimeter.py
```diff
@@ -0,0 +1,300 @@
...
+        # unit conversion factor: 1 eV in keV
+        EV2KEV = 1e-3
```
ax3l: should read: `eV2keV = 1.e-3`

src/tools/lib/python/particle_calorimeter.py
```diff
@@ -0,0 +1,300 @@
...
+
+        figHist = plt.figure(0)
```
ax3l: pls remove the `0`

src/tools/lib/python/particle_calorimeter.py
```diff
@@ -0,0 +1,300 @@
...
+        axHist = figHist.add_subplot(111)
```
ax3l: use `figHist.gca()`

src/tools/lib/python/particle_calorimeter.py
```diff
@@ -0,0 +1,300 @@
...
+        half_yaw = self.opening_yaw / 2
+        half_pitch = self.opening_pitch / 2
+        # solid angle "view range" of the calorimeter detector
+        solid_angle_range = (-half_yaw + self.pos_yaw,
+                             half_yaw + self.pos_yaw,
+                             -half_pitch + self.pos_pitch,
+                             half_pitch + self.pos_pitch)
+
+        # unit conversion: Joules to eV
+        J2EV = sc.electron_volt
```
ax3l: as above

src/tools/lib/python/particle_calorimeter.py
```diff
@@ -0,0 +1,300 @@
...
+        # unit conversion: simUnit to EV
+        simUnitEV = self.sim_unit / J2EV
+        # unit conversion factor: 1 eV in keV
+        EV2KEV = 1e-3
```
ax3l: `eV2keV = 1.e-3`

ax3l:
@PrometheusPi I migrated the PR here for future reference since @n01r won't have time to fix it this month and it started to confuse me with it's wrong author

ax3l:
@n01r since the API is so very much off from what I would like to suggest (close to the openPMD-viewer read API), I will close this PR and write a new module.

After that, the `share/picongpu/examples/Bremsstrahlung/bin/plot_energy_histogram.py` will need an adjustment to work with my API.

Issue #412, opened 2014-05-14
Detect changed include files after compilation


PrometheusPi:
Currently, the user has its set of include files for each set of simulation. (Those usually created using `createParameterSet`.) He  can apply changes to it, compile it and then use `tbg` to start a simulation and save all `include/...` files as well.

However, currently there is only one set of `include/...` files. If the user  applies changes but forgets to  compile, a wrong set up is stored in the simulation directory.

This could be avoided by generating a second directory with `include`  files that is updated after running `make install` and which is then used by `tbg`.
This would avoid guessing if one forgot to  compile :smile: .

wmles:
Seems to be a forgotten issue, which would still be useful. I did forget to recompile today and was confused; I guess such a check would be easy to implement (not for me though :)), and would have no negative side-effects.

ax3l:
Yes it would be quite useful :)

Most of our issues are not forgotten but just nobody had the time yet to work on them. The file we are talking about that might be able to add this feature is `src/picongpu/submit/submitAction.sh`.

ax3l:
@PrometheusPi currently it will be hard to implement anything interactive. Nevertheless, if the implementation in #2123 is already sufficient, feel free to close the issue.

BrianMarre:
@PrometheusPi is this issue still relevant?

PrometheusPi:
We do not log it but we warn the user. I am fin e with this solution.

PR #2123, opened 2017-07-06
TBG: Check Modified Input

ax3l:
Raise a warning in submitAction during `tbg` cfg file evaluation if compilable input files have been modified and seem to be newer then the picongpu binary.

Does not abort or do anything interactive but at least creates awareness. Related to #412

```bash
$ tbg -s qsub -c submit/0008gpus.cfg \
      -t submit/hypnos-hzdr/k80_profile.tpl \
      /bigdata/hplsim/scratch/huebl/lwfa-dev-017
WARNING: 1 input file(s) in include/
         have been modified since the last compile!
         Did you forget to recompile?
List of modified files:
/home/huebl/paramSets/lwfa-dev/include/simulation_defines/param/laser.param
1422697.hypnos3
```

### DIFF
--- a/src/picongpu/submit/submitAction.sh
+++ b/src/picongpu/submit/submitAction.sh
@@ -18,21 +18,34 @@
 # If not, see <http://www.gnu.org/licenses/>.
 #
 
+## check if input files are newer than the picongpu binary
+newerFiles=$(find $TBG_projectPath/include -type f \
+  -newer $TBG_projectPath/bin/picongpu)
+numNewerFiles=$(echo -e "$newerFiles" | wc -l)
+if [ $numNewerFiles -gt 0 ]
+then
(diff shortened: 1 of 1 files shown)


src/picongpu/submit/submitAction.sh
```diff
@@ -18,21 +18,34 @@
 # If not, see <http://www.gnu.org/licenses/>.
 #
 
+## check if input files are newer than the picongpu binary
+numNewerFiles=$(find $TBG_projectPath/include -type f \
```
psychocoderHPC: Could we pass the results of find to a variable and than using `wc`. This would allow to use the intermediate result in case of the error and reduce the file system.

ax3l: yes should try that to reduce FS load, it's just always a mess to work with multiline bash strings :D

PR #2136, opened 2017-07-13
splash2txt: C++98

ax3l:
Keep splash2txt in C++98 so we do not need to artificially push up it's CMakeLists.txt requirements to C++11.

**Or:** also make it require C++11 (the main project that ships it does so anyway). I might prefer the second option (patch below).

Fixes an issue reported in https://github.com/ComputationalRadiationPhysics/picongpu/issues/1984#issuecomment-313293471 (compile fail with, e.g. gcc 4.9 that still defaults to C++98 but supports C++11)

### DIFF
--- a/src/tools/splash2txt/splash2txt.cpp
+++ b/src/tools/splash2txt/splash2txt.cpp
@@ -218,7 +218,7 @@ int main( int argc, char** argv )
         return 1;
     }
 
-    ITools *tools = nullptr;
+    ITools *tools = NULL;
     switch ( options.fileMode)
     {
         case FM_SPLASH: tools = new ToolsSplashParallel( options, mpi_topology, *outStream );
--- a/src/tools/splash2txt/tools_adios_parallel.cpp
+++ b/src/tools/splash2txt/tools_adios_parallel.cpp
@@ -61,7 +61,7 @@ void ToolsAdiosParallel::convertToText()
 
         pVarInfo = adios_inq_var(pFile, nodeName.c_str());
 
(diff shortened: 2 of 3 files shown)


ax3l:
Alternative patch:
```diff
diff --git a/src/tools/splash2txt/CMakeLists.txt b/src/tools/splash2txt/CMakeLists.txt
index 7936f10..b4d8218 100644
--- a/src/tools/splash2txt/CMakeLists.txt
+++ b/src/tools/splash2txt/CMakeLists.txt
@@ -48,6 +48,16 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
 include_directories(include)
 
 
+###############################################################################
+# Language Flags
+###############################################################################
+
+# enforce C++11
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_CXX_STANDARD 11)
+
+
 ################################################################################
 # Build type (debug, release, relwithdebinfo)
 ################################################################################
```

if we want to push to require C++11 (as the main project) instead.

psychocoderHPC:
We should keep it as C++98 project, there is no need to enforce C+11 only because `nullptr` is used.

But it would be good explicit `set(CMAKE_CXX_STANDARD 98)` within CMake.

psychocoderHPC:
offline discussed: we stay for this tool with C++98

ax3l:
Ah, we decided exactly the wrong way around: #2188

PR #2139, opened 2017-07-17
Background Fields: Fix Restart GUARD

### LINKED ISSUE #2137: Restart issue with background fields continues to exist
This is a follow up to #2113:

Test with the TWTS/plane wave background field setup (based on `dev`) still showed disappearing particles and wrong field values when used with restarts. Thus a restart issue still exists.

------
**UPDATE/important to mention** (thanks @ax3l for pointing this out)
**This is (most likely) an issue with the field restart!** The disappearing particles is the most obvious result, but is probably only caused by the wrong field values after restart. 

**UPDATE No. 2:**
I forgot to cherry-pick #2112 `aee7784672da6c3ded6fe00f47be9fdcd4b97948` - I know updated the plots and **the error is about a factor 2** in the simple setup presented below.

-----

The following difference could be seen for the same simulation setup - fist with restart every 10kth iteration, the second without any restart.

n_macroparticles

fieldenergy

As mentioned by @psychocoderHPC in #2113 I will try to find an easy setup to reproduce this bug based on 0.3.0.

mentioning:
@BeyondEspresso 
@steindev

ax3l:
(Trying) to close #2137:
Restart of background fields could cause:
- artificial noise after restart from the outer regions with non-periodic boundary conditions
- if background fields are used (e.g. TWTS pulse)

Simplify the logic of restarting fields:
- do not add the already contained fields in `movingWindowCheck` again
- spares us the substraction we had before
- GUARDS are not part of the checkpoint data, which is cumbersome for the "outer" GUARDS of our global domain
- we need to add the background field ONLY in those to generate a valid restart field

### DIFF
--- a/include/picongpu/simulationControl/MySimulation.hpp
+++ b/include/picongpu/simulationControl/MySimulation.hpp
@@ -421,6 +421,10 @@ public:
         GridController<simDim> &gc = Environment<simDim>::get().GridController();
         gc.setStateAfterSlides(0);
 
+        DataConnector &dc = Environment<>::get().DataConnector();
+        auto fieldE = dc.get< FieldE >( FieldE::getName(), true );
+        auto fieldB = dc.get< FieldB >( FieldB::getName(), true );
+
         /* fill all objects registed in DataConnector */
         if (initialiserController)
         {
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
I verified this fix with the field background example from #2137. I used only the field background for the E field.

PrometheusPi:
@ax3l l see above for the backport to 0.3.0 - it is the last commit

PrometheusPi:
run time test results:
fixed
(for the 0.3.0 backport)

ax3l:
@BeyondEspresso please take note of this bugfix fo restarts with background fields if you have not seen it yet. also applies to (all) previous releases!

BeyondEspresso:
Guys, this is awesome! :smiley: Great work, all of you! :1st_place_medal:

PR #2140, opened 2017-07-18
lockstep refactoring `KernelManipulateAllParticles`

psychocoderHPC:
- add documentation

# Tests

- [x] 3D KHI electrons/positrons

### DIFF
--- a/include/picongpu/particles/Particles.kernel
+++ b/include/picongpu/particles/Particles.kernel
@@ -204,7 +204,7 @@ struct KernelDeriveParticles
     }
 };
 
-/** manipulate particle of a species
+/** manipulate particles of a species
  *
  * This functor prepares a particle box to call
  * a user defined unary functor which allows to manipulate the particles.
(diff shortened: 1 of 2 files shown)


include/picongpu/particles/Particles.kernel
```diff
@@ -204,65 +204,146 @@ struct KernelDeriveParticles
...
-            if (linearThreadIdx == 0)
-            {
-                frame = pb.getPreviousFrame(frame);
-            }
-            isParticle = true;
-            __syncthreads();
+                    if( !isPar )
+                        particle.setHandleInvalid( );
+
+                    // call manipulator even if the particle is not valid
```
ax3l: ok, but why not a simple `else` to save the following check?

include/picongpu/particles/Particles.kernel
```diff
@@ -204,65 +204,146 @@ struct KernelDeriveParticles
...
+                    accManipulator( particle );
+
+                    // assume that all other frames has no gaps
```
ax3l:
```C++
/* only the last frame is allowed to be non-full: all following
 * frames' particles will be valid, since we iterate the list of
 * frames from back to front
 */
```

include/picongpu/particles/Particles.kernel
psychocoderHPC: It is not forbidden that the FilteredFunctor has a inner state and creates statistics . Therefore this will keep the freedom to do it.

psychocoderHPC:
updated

psychocoderHPC:
merge merge merge ^^

PR #2162, opened 2017-07-31
png2gas: C++98

### LINKED ISSUE #2159: PNG2Gas: cannot be built
The png2gas tool cannot be built like the [tutorial](http://picongpu.readthedocs.io/en/latest/install/dependencies.html?highlight=png2gas#png2gas) says.
```
[ 50%] Building CXX object CMakeFiles/png2gas.dir/png2gas.cpp.o
/HOME/src/picongpu/src/tools/png2gas/png2gas.cpp:128:5: warning: identifier ‘nullptr’ is a keyword in C++11 [-Wc++0x-compat]
     MPI_Init(nullptr, nullptr);
     ^
/HOME/src/picongpu/src/tools/png2gas/png2gas.cpp: In function ‘int main(int, char**)’:
/HOME/src/picongpu/src/tools/png2gas/png2gas.cpp:128:14: error: ‘nullptr’ was not declared in this scope
     MPI_Init(nullptr, nullptr);
              ^
make[2]: *** [CMakeFiles/png2gas.dir/png2gas.cpp.o] Error 1
make[1]: *** [CMakeFiles/png2gas.dir/all] Error 2
```

Exchanging `nullptr`back to `NULL` fixed it but that is a backport ^^;

ax3l:
Close #2159: Keep `png2gas` in C++98 so we do not need to artificially push up it's `CMakeLists.txt` requirements to C++11.

### DIFF
--- a/src/tools/png2gas/png2gas.cpp
+++ b/src/tools/png2gas/png2gas.cpp
@@ -125,7 +125,7 @@ int main(int argc, char **argv)
     if (!parseCmdLine(argc, argv, options))
         return -1;
 
-    MPI_Init(nullptr, nullptr);
+    MPI_Init(NULL, NULL);
 
     Dimensions data_size(options.dataSize);
     std::cout << "Creating density data with size " << data_size.toString() << std::endl;


ax3l:
Ah, we decided exactly the wrong way around: #2188

PR #2166, opened 2017-08-01
Hypnos: PNGwriter 0.6.0

ax3l:
PNGwriter 0.6.0 is now available on Hypnos (HZDR). Using it by default adds some performance improvements for PIConGPU and fulfills the requirement of the png2gas tool.


ax3l:
@psychocoderHPC ping ;)

PR #2167, opened 2017-08-02
functor interface refactoring

psychocoderHPC:
This pull request refactors the new functor interface to allow the usage for the startPosition functors as well. 

- change the visibility of inherited classes
- move the random number initialization to a separate class


ax3l:
@psychocoderHPC I just realized the C++11 `final` keyword could have potentially saved you some headaches here (to protect some of the member functions).

psychocoderHPC:
Final is for virtual methods, I can't see how this could help here. We can discuss it tomorrow offline.

ax3l:
ah true, I overlooked that! forget what I just wrote...

PR #2172, opened 2017-08-04
Sphinx: Migrate Plugins

ax3l:
Migrates the plugins pages from the wiki to the manual.

Most committed as generic user.

@ComputationalRadiationPhysics/picongpu-developers Please feel free to fix images and style errors in follow-ups!


ax3l:
offline discussion: we decided to do a first merge to `dev` even if not everything is 100% perfect yet (but I tried).
This allows us to shut down the section in the wiki so nobody contributes to the wrong pages.

Please improve further via PRs to dev everyone :)

PrometheusPi:
@ax3l as agreed, I will merge as soon as possible after the compile suite has finished. 
There should be no side effects with the `*.rst` files. 
Any errors will be fixed afterwards.

PR #2173, opened 2017-08-07
PMacc:: generic `min()` and `max()` implementation

psychocoderHPC:
- add generic `min()` and `max()` implementation (also supports integer types)
- use PMacc min/max within PIConGPU

### DIFF
--- a/include/pmacc/algorithms/math/defines/comparison.hpp
+++ b/include/pmacc/algorithms/math/defines/comparison.hpp
@@ -35,6 +35,27 @@ struct Max;
 template<typename T1,typename T2>
 struct Min;
 
+template<typename T>
+struct Max<T,T>
+{
+    typedef T result;
+
+    HDINLINE T operator()(T value1, T value2)
+    {
+        return value1 > value2 ? value1 : value2;
+    }
+};
+
+template<typename T>
+struct Min<T,T>
+{
+    typedef T result;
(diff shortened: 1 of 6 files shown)


include/pmacc/algorithms/math/defines/comparison.hpp
```diff
@@ -35,6 +35,27 @@ struct Max;
 template<typename T1,typename T2>
 struct Min;
 
+template<typename T>
+struct Max<T,T>
+{
+    typedef T result;
```
ax3l: pls use `using` for new code

include/pmacc/algorithms/math/defines/comparison.hpp
```diff
@@ -35,6 +35,27 @@ struct Max;
...
+    HDINLINE T operator()(T value1, T value2)
+    {
+        return value1 > value2 ? value1 : value2;
+    }
+};
+
+template<typename T>
+struct Min<T,T>
+{
+    typedef T result;
```
ax3l: pls use `using` for new code

include/pmacc/algorithms/math/defines/comparison.hpp
psychocoderHPC: I will reformat this code after cupla support is enable.

PR #2184, opened 2017-08-09
Sphinx: PDF Radiation 

ax3l:
Remove latex unexpected character so sphinx pdf builds.

Also fix a typo in the nonACII test:
- fix typo in selection of .py files


ax3l:
@PrometheusPi tested :)

PR #2187, opened 2017-08-09
Bash Subshells: `cmd` to $(cmd)

ax3l:
Generally, prefer this style in call bash scripts.

It can be:
- nested,
- is a tiny bit more portable and
- is not prune to be [mismatched with '](https://github.com/ComputationalRadiationPhysics/picongpu/issues/1984#issuecomment-321235872)

and has no difference in meaning.

For more information, see e.g.
  https://github.com/azet/community_bash_style_guide#common-mistakes-and-useful-tricks


psychocoderHPC:
please rebase there is a merge conflict

PR #2193, opened 2017-08-11
CMake: Use Imported Targets Zlib, Boost

ax3l:
Use imported targets for Zlib and Boost. This makes linking with existing libraries in system paths more robust, e.g. on centos, and spares the need to add include paths.

Do not add math on windows machines.

With our CMake requirement (version) we can rely on targets for all dependencies.
All dependencies, but Boost, where we need a fallback:

> Since the boost transient dependencies are updated via a manual list currently, we explicitly request all transient libraries for now (instead of letting the cmake script find and add them). This helps users with a newer boost version but an older CMake that does not know the latest boost releases (yet).

Since boost targets are not available in the same situation either, we add the `include_directories` + `${Boost_LIBRARIES}` fallbacks until boost has a forward-compatible CMake build chain.

### To Do

- [x] rebase against #2195 
- [x] rebase against #2198 (last commit migrated there)
- [x] Boost: evaluate and react on `if(TARGET Boost::boost)`
- [x] Boost: also include `Boost::boost` for *explicit* request of header-only dependencies?

### DIFF
--- a/include/picongpu/CMakeLists.txt
+++ b/include/picongpu/CMakeLists.txt
@@ -129,13 +129,23 @@ if(ALPAKA_ACC_GPU_CUDA_ENABLE)
     set(LIBS ${LIBS} ${mallocMC_LIBRARIES})
 endif()
 
+
 ################################################################################
 # Find zlib
 ################################################################################
 
 find_package(ZLIB REQUIRED)
-include_directories(SYSTEM ${ZLIB_INCLUDE_DIRS})
-set(LIBS ${LIBS} ${ZLIB_LIBRARIES})
+set(LIBS ${LIBS} ZLIB::ZLIB)
+
+
+################################################################################
+# Find math from stdlib
+################################################################################
+
+if(NOT WIN32)
+    # automatically added on windows
+    set(LIBS ${LIBS} m)
+endif()
 
 
 ################################################################################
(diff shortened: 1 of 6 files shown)


include/pmacc/PMaccConfig.cmake
```diff
@@ -141,6 +141,9 @@ if((NOT ALPAKA_ACC_GPU_CUDA_ENABLE) AND ALPAKA_ACC_GPU_CUDA_ONLY_MODE)
     message(WARNING "ALPAKA_ACC_GPU_CUDA_ONLY_MODE is set to OFF because cuda backend is not activated")
 endif()
 
+set(PMacc_LIBRARIES ${PMacc_LIBRARIES} ${cupla_LIBRARIES})
```
ax3l: @BenjaminW3 I could probably also just add the target `cupla` here..., right?

BenjaminW3: I think so but I do not know the full context. When you do this you also have to remove all explicit setters for cupla_IncludeDir etc. If they exist.

ax3l: I am thinking if we just want to do this directly in the `cupla/alpaka_add_executable` wrappers (where also the includes are currently set)

ax3l: double setting the includes does not harm, but yes with the target they can (should) be removed since the target already covers them which is also reducing a bit of code

ax3l: I think this line should be here but I leave it out since it is out-of-scope of this PR.

ax3l:
@PrometheusPi review-able :)

include/picongpu/CMakeLists.txt
```diff
@@ -365,14 +382,14 @@ add_library(picongpu-hostonly
     STATIC
     ${SRCFILES}
 )
-target_link_libraries(picongpu-hostonly ${LIBS} m)
```
PrometheusPi: QUESTION: 
what is the meaning of `m` here?

ax3l: it refers to the std math libraries.
On Windows, those are added by default (and adding them again leads to an error), but on all other OS one has to add that lib when accessing std math functions.

PR #2195, opened 2017-08-14
Boost Transient Deps: date_time, chrono, atomic

ax3l:
The Boost.threads library pulls in `date_time` and `chrono` as additional dependencies that need a compile.

```bash
$ grep -iR "_Boost_THREAD_DEPENDENCIES" /usr/share/cmake-3.7/Modules/FindBoost.cmake
    set(_Boost_THREAD_DEPENDENCIES date_time)
[...]
    set(_Boost_THREAD_DEPENDENCIES chrono system date_time)
    set(_Boost_THREAD_DEPENDENCIES chrono system date_time atomic)
[...]
```

Since the boost transient dependencies are updated via a manual list currently, we explicitly request all transient libraries for now (instead of letting the cmake script find and add them). This helps users with a newer boost version but an older CMake that does not know the latest boost releases (yet).

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -92,14 +92,14 @@ zlib
 
 boost
 """""
-- 1.62.0-1.64.0 (``program options``, ``regex`` , ``filesystem``, ``system``, ``thread``, ``math``, ``serialization`` and header-only libs, optional: ``fiber``, ``context``)
+- 1.62.0-1.64.0 (``program options``, ``regex`` , ``filesystem``, ``system``, ``thread``, ``chrono``, ``atomic``, ``date_time``, ``math``, ``serialization`` and header-only libs, optional: ``fiber``, ``context``)
 - download from `http://www.boost.org <http://sourceforge.net/projects/boost/files/boost/1.62.0/boost_1_62_0.tar.gz/download>`_
-- *Debian/Ubuntu:* ``sudo apt-get install libboost-program-options-dev libboost-regex-dev libboost-filesystem-dev libboost-system-dev libboost-thread-dev libboost-math-dev libboost-serialization-dev # libboost-fiber-dev libboost-context-dev``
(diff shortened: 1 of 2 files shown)


ax3l:
We actually use the `boost/thread` header only in the PNG plugin and only a very limited feature set of it. We should replace it with `std::thread` in C++11 now...

ax3l:
@BenjaminW3 fyi, that's how I handle transient deps in PIConGPU.

> Since the boost transient dependencies are updated via a manual list currently, we explicitly request all transient libraries for now (instead of letting the cmake script find and add them). This helps users with a newer boost version but an older CMake that does not know the latest boost releases (yet).

ax3l:
@PrometheusPi pls merge me :)

INSTALL.rst
```diff
@@ -92,14 +92,14 @@ zlib
 
 boost
 """""
-- 1.62.0-1.64.0 (``program options``, ``regex`` , ``filesystem``, ``system``, ``thread``, ``math``, ``serialization`` and header-only libs, optional: ``fiber``, ``context``)
+- 1.62.0-1.64.0 (``program options``, ``regex`` , ``filesystem``, ``system``, ``thread``, ``chrono``, ``atomic``, ``date_time``, ``math``, ``serialization`` and header-only libs, optional: ``fiber``, ``context``)
 - download from `http://www.boost.org <http://sourceforge.net/projects/boost/files/boost/1.62.0/boost_1_62_0.tar.gz/download>`_
-- *Debian/Ubuntu:* ``sudo apt-get install libboost-program-options-dev libboost-regex-dev libboost-filesystem-dev libboost-system-dev libboost-thread-dev libboost-math-dev libboost-serialization-dev # libboost-fiber-dev libboost-context-dev``
```
PrometheusPi: QUESTION:
Why was there a `#` here?

ax3l: just to separate the optional dependencies that are not of interest yet since the alpaka fiber backend is not yet exposed to users

PR #2198, opened 2017-08-14
PMacc: Forward cupla/alpaka Libs

ax3l:
Forward additional third-party libraries. Alpaka might pull further libraries from `Boost.fiber` or any other backend, that are not used directly in the final (PIConGPU) target(s).

### DIFF
--- a/include/pmacc/PMaccConfig.cmake
+++ b/include/pmacc/PMaccConfig.cmake
@@ -141,6 +141,12 @@ if((NOT ALPAKA_ACC_GPU_CUDA_ENABLE) AND ALPAKA_ACC_GPU_CUDA_ONLY_MODE)
     message(WARNING "ALPAKA_ACC_GPU_CUDA_ONLY_MODE is set to OFF because cuda backend is not activated")
 endif()
 
+# add possible indirect/transient library dependencies from alpaka backends
+# note: includes and definitions are already added in the cupla_add_executable
+#       wrapper
+set(PMacc_LIBRARIES ${PMacc_LIBRARIES} ${cupla_LIBRARIES})
+
+
 ################################################################################
 # VampirTrace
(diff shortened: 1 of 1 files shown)


include/pmacc/PMaccConfig.cmake
```diff
@@ -141,6 +141,11 @@ if((NOT ALPAKA_ACC_GPU_CUDA_ENABLE) AND ALPAKA_ACC_GPU_CUDA_ONLY_MODE)
     message(WARNING "ALPAKA_ACC_GPU_CUDA_ONLY_MODE is set to OFF because cuda backend is not activated")
 endif()
 
+# includes and definitions are already added in cupla's & alpaka's
```
BenjaminW3: `alpaka_add_executable` does not add any definitions or includes. This is all done by cupla.
I do not intend to add those within alpaka because `alpaka_add_executable` should be identical to the CMake `add_executable` in which it only creates a target with the given files.

ax3l: you are right, and alpaka does it correctly!

ideally we would not need a wrapper at all but only a property set.
Shouldn't this be possible with CUDA language support in CMake 3.8+. Or should we still hide the language property from the alpaka user and only talk to them in form of alpaka backends?

BenjaminW3: I would hope that the CMake native CUDA language support would allow this. However, it does not yet support clang as native CUDA compiler at all. Furthermore, I do not know how it handles `.cpp` files containing CUDA code. Currently `alpaka_add_executable` enforces CUDA compilation of all extensions.

ax3l:
@PrometheusPi I think this change is fine for now and @psychocoderHPC can cross-check it after his vacations even if we already merge it to `dev`

PR #2204, opened 2017-08-15
New Tool: pic-build

ax3l:
Just run `pic-build` inside a input directory.

This updates the binaries for a `tbg` call with the latest options in an input's param sets.

Just short-hand for

```bash
mkdir .build
cd .build
pic-configure ..
make -j install
cd -
```

Avoids confusion for users that are not used to out-of-source builds.

cc'ing @ComputationalRadiationPhysics/picls-developer @ComputationalRadiationPhysics/picls-maintainer please provide feedback and give this new tool in `dev` a try! You will never have to switch directories during input design again 💃

### DIFF
--- a/USAGE.rst
+++ b/USAGE.rst
@@ -15,17 +15,19 @@ Preparation
 First, decide where to store input files, a good place might be ``$HOME`` (``~``) because it is usually backed up.
 Second, decide where to store your output of simulations which needs to be placed on a high-bandwidth, large-storage file system which we will refer to as ``$SCRATCH``.
 
-For a first test you can also use your home directory ``$HOME`` as simulation directory ``$SCRATCH``.
-
-As in our :ref:`compiling from source <install-source>` section, we need a few directories to structure our workflow:
+For a first test you can also use your home directory:
 
 .. code-block:: bash
 
-    # temporary build directory
-    mkdir $HOME/build
+   export SCRATCH=$HOME
+
+We need a few directories to structure our workflow:
+
+.. code-block:: bash
 
     # PIConGPU input files
     mkdir $HOME/picInputs
+
     # PIConGPU simulation output
     mkdir $SCRATCH/runs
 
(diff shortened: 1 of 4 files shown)


USAGE.rst
```diff
@@ -111,6 +116,41 @@ A run simulation can also be reused to create derived input sets via ``pic-creat
...
+
+   # example for running efficiently on the CPU you are currently compiling on
+   pic-build -a "omp2b"
+
+Its full documentation from ``pic-build --help`` reads:
+
+.. program-output:: ../../pic-build --help
+
 pic-configure
 """""""""""""
```
ax3l: shall I add the sentence below:
```
You will likely not use this tool directly if you are using `pic-build` from above.
```
?

PrometheusPi: Yes, that's a great idea.

ax3l: all right, added.

pic-build
```diff
@@ -0,0 +1,112 @@
...
+
+this_dir=`dirname $0`
+build_dir=".build"
+
+help()
+{
+    echo "Build new binaries for a PIConGPU input set"
+    echo ""
+    echo "Creates or updates the binaries in an input set. This step needs to"
+    echo "be performed every time a .param file is changed."
```
PrometheusPi: OPTIONAL:
or a `*.unitless` file

pic-build
```diff
@@ -0,0 +1,112 @@
...
+    echo "This is just a short-hand tool for switching to a temporary build"
+    echo "directory and running 'pic-configure ..' and 'make install'"
+    echo "manually."
+    echo ""
+    echo "You must run this command inside an input directory."
+    echo ""
+    echo "usage: pic-build [OPTIONS]"
+    echo ""
+    # lend the rest of the options from pic-configure
+    echo "$($this_dir/pic-configure --help | tail -n +12)"
```
PrometheusPi: QUESTION:
skipping the first 12 lines is quite explicit. What happens when we shorten/lengthen the first lines in the `pic-configure` help?

pic-build
```diff
@@ -0,0 +1,112 @@
...
+    echo "ERROR: Could not find directory 'include/picongpu'!" >&2
+    echo "       Are you in a PIConGPU input directory?" >&2
+    exit 1
+fi
+
+# diagnostics
+echo -e "\033[32mbuild directory:\033[0m $build_dir"
+
+# create or re-use an existing build directory
+mkdir -p $build_dir
```
PrometheusPi: OPTIONAL:
the `-p` option can be removed here, since it is not necessary

USAGE.rst
```diff
@@ -41,51 +43,46 @@ Step-by-Step
...
    pic-create $PICSRC/examples/LaserWakefield/ $HOME/picInputs/myLWFA
 
-Now edit ``$HOME/picInputs/myLWFA/include/picongpu/simulation_defines/param/*`` to change the :ref:`physical configuration of this input set <usage-params>`.
+   # switch to your input directory
+   cd $HOME/picInputs/myLWFA
 
-Now edit ``$HOME/picInputs/myLWFA/etc/picongpu/*.cfg`` to adjust :ref:`runtime parameters (simulation size, number of GPUs, plugins, ...) <usage-cfg>`.
+PIConGPU is controlled via two kinds of input sets: compile-time options and runtime options.
+
+Edit the :ref:`.param files<usage-params>` inside ``include/picongpu/simulation_defines/param/``.
```
PrometheusPi: QUESTION:
what about `*.unitless` files?

pic-build
ax3l: we don't mention them here to not overcomplicate things.
users should not interact with those.

in principle *anything* inside `include/` triggers a re-compile need (our check in `tbg` will remind them before submission as a safety net).

pic-build
ax3l: if we do so, which we don't do often, the output of `pic-build --help` will just start a bit in-the-middle.

nevertheless, it's better than re-typing the exact same options. also one would see it quickly if forgotten, since it's output ins included in the basic usage in the manual.

pic-build
ax3l: nope, `-p` does not only guarantee that parents are created.
it also guarantees that it will not complain if `.build` already exists

USAGE.rst
ax3l: not relevant here, see above. users shall only interface with `.param` files, everything else is a hack for power-users.

pic-build
PrometheusPi: Thanks. I didn't know that.

PR #2209, opened 2017-08-17
Python: Energy Histogram Reader

ax3l:
Re-implementation of #2122 

Reader class for the energy histogram plugin.

Use as follows:
```python
from picongpu.plugins.energy_histogram import EnergyHistogram

energy_histogram = EnergyHistogram('/home/axel/runs/lwfa_001')
energies, bins = energy_histogram.get(
    'e',
    iteration=2000
)
```

Also adds a helper utility to match requested time in seconds to actually available iterations:

```python
from picongpu.utils.find_time import FindTime
import numpy as np

ft = FindTime('/home/axel/runs/lwfa_001/')
dt = ft.get_dt()

# basic logic
ft.get_iteration(100 * dt, np.array([0, 100,200,300]))
# 100, 1.79e-14
ft.get_iteration(155 * dt, np.array([0, 100,200,300]), method='closest')
# 200, 3.58e-14
ft.get_iteration(255 * dt, np.array([0, 100,200,300]))
# 200, 3.58e-14
```

Or used with the energy histogram plugin:
```python
from picongpu.plugins.energy_histogram import EnergyHistogram
from picongpu.utils.find_time import FindTime

run_directory = '/home/axel/runs/lwfa_001'
energy_histogram = EnergyHistogram(run_directory)
ft = FindTime(run_directory)

request_time = 50.e-12 # 50 ps

iteration, time = ft.get_iteration(request_time, energy_histogram.get_iterations('e'))
# time is potentially NOT the same as request_time anymore

energies, bins = energy_histogram.get(
    'e',
    iteration=iteration
)

plt.plot(bins, energies, label="t={} ps".format(time * 1.e12))
plt.show()
```

Needed as the basis for a more efficient post-processing chain and the upcoming PIConGPU notebook GUI.

Also moves & renames the radiation plugin reader appropriately.

@ComputationalRadiationPhysics/picongpu-developers  please try to orientate on this class when writing further python helper classes.

### Follow-Up Work

Outside of the scope of this PR, we can do the following follow-up tasks:
- [x] implement over-/underflow bin returns
- implement meta class for bin arguments and other interesting information (particle filters, etc.). maybe even include all return values there
- [x] implement a `picongpu.utils.find_time` class that reads the `dt` of the simulation and converts iterations (steps) and actual times back & forth, including finding the "closest timestep for a requested time" (round-down as in openPMD-viewer)


ax3l:
ccing @codingS3b here is a first reader for the energy histogram plugin (currently under review. will be going to `dev` and the upcoming 0.4.0 release in a few months)

src/tools/lib/python/picongpu/plugins/energy_histogram.py
```diff
@@ -0,0 +1,160 @@
...
+    def __init__(self, run_directory):
+        """
+        Parameters
+        ----------
+        simulation_directory : string
+            path to the run directory of PIConGPU
+            (the path before ``simOutput/``)
+        """
+        if run_directory is None:
+            raise ValueError('The run_directory parameter can not be None!')
```
PrometheusPi: OPTIONAL:
I would suggest adding a check if run_directory is a directory instead.
That would make the warning below clearer see line 50.

src/tools/lib/python/picongpu/plugins/energy_histogram.py
```diff
@@ -0,0 +1,160 @@
...
+        Returns
+        -------
+        A string with a path.
+        """
+        if species is None:
+            raise ValueError('The species parameter can not be None!')
+
+        sim_ouput_dir = os.path.join(self.run_directory, "simOutput")
+        if not os.path.isdir(sim_ouput_dir):
+            raise IOError('The simOutput/ directory does not exist inside '
```
PrometheusPi: OPTIONAL:
This error message could be clearer (just stating `simOutput` was not found) if the existence of the base directory was checked before.

src/tools/lib/python/picongpu/plugins/energy_histogram.py
```diff
@@ -0,0 +1,160 @@
...
+        if not os.path.isfile(data_file_path):
+            raise IOError('The file {} does not exist.\n'
+                          'Did the simulation already run?'
+                          .format(data_file_path))
+
+        return data_file_path
+
+    def get_iterations(self, species):
+        """
+        Return the an array of iterations with available data.
```
PrometheusPi: REQUIRED:
remove either `the` or `an`

src/tools/lib/python/picongpu/plugins/energy_histogram.py
```diff
@@ -0,0 +1,160 @@
...
+        Get a histogram.
+
+        Parameters
+        ----------
+        species : string
+            short name of the particle species, e.g. 'e' for electrons
+            (defined in ``speciesDefinition.param``)
+        iteration : (unsigned) int [unitless], optional
+            The iteration at which to read the data.
+            Either `t` or `iteration` must be given.
```
PrometheusPi: OPTIONAL:
~~please add a `@TODO` here too~~
Is fine - check is implemented, only `t` does not work.

src/tools/lib/python/picongpu/plugins/energy_histogram.py
```diff
@@ -0,0 +1,160 @@
...
+            iteration of the histogram
+        t : float [seconds]
+            time of the histogram
+            @TODO Not implemented. Always zero.
+        energies : np.array of dtype float [unitless]
+            count of particles in each bin
+        bins : np.array of dtype float [keV]
+            upper ranges of each energy bin
+
+        A list of 1darray corresponding to the data requested in `var_list`
```
PrometheusPi: REQUIRED:
plural `1darrays`

docs/source/usage/plugins/energyHistogram.rst
```diff
@@ -58,25 +58,17 @@ Each row describes another simulation time step.
 Analysis Tools
 ^^^^^^^^^^^^^^
 
-The easiest way is to load the data in Python:
+You can quickly plot the data in Python with:
 
 .. code:: python
 
-   import numpy as np
+   from picongpu.plugins.energy_histogram import *
```
PrometheusPi: OPTIONAL:
make it clearer: 
```python
from picongpu.plugins.energy_histogram import EnergyHistogram
```

docs/source/usage/plugins/energyHistogram.rst
```diff
@@ -58,25 +58,17 @@ Each row describes another simulation time step.
...
-   # upper range of each bin in keV
-   #    note: only reads first row and selects the 1024 bins (default, see binCount) in columns 2:1026
-   e_bins = np.loadtxt("simOutput/e_energyHistogram.dat", comments='', usecols=range(2,1026) )[0,:]
-   # time steps on which histogram was written
-   e_steps = np.loadtxt("simOutput/e_energyHistogram.dat", usecols=range(0,1))
-
-   # example: plot the 10th dump (10 * .period)
-   plt.plot( e_bins, e_ene[10] )
+   energy_histogram = EnergyHistogram('/home/axel/runs/lwfa_001')
+   iteration, t, energies, bins = energy_histogram.get('e', iteration=2000)
```
as discussed *offline* today - where should a method to get **all iterations** go?

ax3l: I just changed this to make it easier.
all-iterations can be implemented as an update later on

src/tools/lib/python/picongpu/plugins/energy_histogram.py
ax3l: The check is done on first data access.
Due to the species' influence, it also makes no sense to check it earlier (dir + file)

src/tools/lib/python/picongpu/plugins/energy_histogram.py
ax3l: see next line. that's already in there

src/tools/lib/python/picongpu/plugins/energy_histogram.py
ax3l: removed

src/tools/lib/python/picongpu/plugins/energy_histogram.py
ax3l: removed, outdated

src/tools/lib/python/picongpu/utils/find_time.py
```diff
@@ -0,0 +1,166 @@
...
+        ----------
+        simulation_directory : string
+            path to the run directory of PIConGPU
+            (the path before ``simOutput/``)
+        """
+        if run_directory is None:
+            raise ValueError('The run_directory parameter can not be None!')
+
+        self.run_directory = run_directory
+        self.data_file = "output"
```
ax3l: note: relying on this file is still a bit messy, especially during restarts, but we can generate a cleaner log & lookup of basic parameters later on, too

src/tools/lib/python/picongpu/utils/find_time.py
```diff
@@ -0,0 +1,166 @@
...
+        if method == 'previous':
+            iteration = iterations_sorted[prev_i]
+            if times_sorted[prev_i] > t:
+                raise IndexError('Time t not found in valid range!')
+        if method == 'closest':
+            if np.abs(times_sorted[prev_i] - t) <= \
+               np.abs(times_sorted[next_i] - t):
+                iteration = iterations_sorted[prev_i]
+            else:
+                iteration = iterations_sorted[next_i]
```
PrometheusPi: REQUIRED:
This can be shortened:
```python
iteration = (np.abs(times_sorted - t)).argmin()
```

docs/source/usage/plugins/energyHistogram.rst
PrometheusPi: Okay

src/tools/lib/python/picongpu/plugins/energy_histogram.py
PrometheusPi: Okay

src/tools/lib/python/picongpu/plugins/energy_histogram.py
PrometheusPi: Okay

src/tools/lib/python/picongpu/utils/find_time.py
```diff
@@ -0,0 +1,166 @@
...
+                          'Did the simulation already run?'
+                          .format(self.run_directory))
+
+        data_file_path = os.path.join(sim_ouput_dir, self.data_file)
+        if not os.path.isfile(data_file_path):
+            raise IOError('The file {} does not exist.\n'
+                          'Did the simulation already run?'
+                          .format(data_file_path))
+
+        return data_file_path
```
ax3l: partly, I thought about this but since some might need an extra prefix for species and other might actually get a directory instead of a file I did not dig deeper yet.

But in principle yes, one can abstract that in the future into a common helper class.

src/tools/lib/python/picongpu/utils/find_time.py
ax3l: Good point! But I think it must read
```python
closest_i = (np.abs(times_sorted - t)).argmin()
iteration = iterations_sorted[closest_i]
```

no?

src/tools/lib/python/picongpu/utils/find_time.py
```diff
@@ -0,0 +1,163 @@
...
+            prev_i = next_i - 1
+        else:
+            prev_i = 0
+
+        if method == 'previous':
+            iteration = iterations_sorted[prev_i]
+            if times_sorted[prev_i] > t:
+                raise IndexError('Time t not found in valid range!')
+        if method == 'closest':
+            closest_i = np.abs(times_sorted - t).argmin()
```
ax3l: @PrometheusPi ah now, now I am accessing and sorting the whole array again instead of just using the closest result... your proposal is more costly

PrometheusPi: @ax3l Yes, this is looks more costly on the CPU side, but be aware you are using an interpreter here thus the if clauses above are (more) costly too. Thus in the end you trade multi line python code (**slow**) against probably complicated but fast C++/C/Fortran code which is just a single line in python.

User / code review happiness increases :+1:
compute time decreases :+1:

Okay writing your code in pure C/C++ would be faster 😉

PrometheusPi: *offline discussion* 
For large arrays @ax3l initial design is better.

ax3l:
oh, we just found there is a python3 issue with the energy histogram reader. will fix

ax3l:
@codingS3b all right, fixed in the last commit. thanks for noticing!

PR #2214, opened 2017-08-22
Density: Param Gaussian Density

ax3l:
Fix wrong parameter usage in `Gaussian` density profile:
Instead of `gasSigmaRight` the `gasCenterRight` parameter was taken for the scale-length of the downramp.

Bug since the 0.2.x release series: 7b451c01304a228895d00fbe27507859c1e93569


PrometheusPi:
This looks like a serious bug for LWFA down ramp simulations. 
(Does not effect the Jurjen Simulation.)

ax3l:
yes it's quite horrible if this has been used (and it's used in the default examples).

BeyondEspresso:
It seems I was almost bitten by it in my propagation simulation last week, where I used this feature. The `else`-statement in `else if (globalCellPos.y() >= gas_center_right)` and the fact that I did not need to simulate beyond the gas jet peak has saved me. An `if`-statement alone would have been somewhat a disaster. :sweat_smile:  :sweat_drops:

ax3l:
if @PrometheusPi would already merge it it could already be backported by me to `0.3.1-rc3` :)

PR #2217, opened 2017-08-22
Python Tools: Move lib to /

ax3l:
Related to #1968 

- Create a proper `lib/` directory in our repo
- `pic-create`: copy `lib/`: 
  - Allows to overwrite default files in lib/ the same way as input params.
  - Makes sure that python tools for post-processing stay with the version of the source that created them.


pic-create
```diff
@@ -118,7 +118,7 @@ do
     if [ ! -d "$dir_dst" ] ; then
         mkdir -p "$dir_dst"
     fi
-    rsync --inplace -q -avc --exclude=".*" $dir_src/ $dir_dst
+    rsync --inplace -r -q -avc --exclude=".*" $dir_src/ $dir_dst
```
PrometheusPi: QUESTION:
Why was there no `-r` flag before?

PrometheusPi:
@ax3l What about executable python scripts in `picongpu/src/tools/bin/` - should they be stored alongside a simulation as well to ensure compatibility?

ax3l:
`src/tools/bin/` is quite of a mess right now.

- plot tools might need to go to a separate repo, since they are high-level and user dependent; or to the specific example's `bin/` dir that uses them
- data readers (without plot) should go in `lib/python/picongpu/plugins/
- all others might need to go to `bin/` directly as long as they are core functionality (`egetopt`, ...)

pic-create
ax3l: because we never needed to copy a dir recursively but added them explicitly

PR #2219, opened 2017-08-23
New Tool: pic-edit

ax3l:
For easier forward-compatibility, we should not fill-up all default `.param` and `.unitless` files during pic-create. This also makes version control of input easier.

Our defaults are sane (basic PIC cycle without any density or lasers) for a PIC simulation and will only cause unnecessary conflicts when the defaults change.

Instead, we can clone only the files that are changed in input on `pic-create` and add the others automatically when needed.

The workflow possible with `pic-edit` editing `.param` files is:

```bash
  pic-create $PICSRC/share/picongpu/LaserWakefield lwfa
  cd lwfa

  # edit compile time options
  pic-edit grid laser isaac
  # will copy in a default isaac.param
  # will open the already existing grid.param and laser.param plus
  #   the added isaac.param inside the simulation_defines/param/
  #   directory

  pic-build

  # edit run time options in .cfg files
  nano etc/picongpu/0001gpus.cfg
  # submit
  tbg ...
```

This also solves the issue of telling users "where to get additional files from" when they change inputs, because they do not need to know anymore.

`.unitless` files are not able to be opened with `pic-edit` since they are for power-users only anyway.

See also the help of this tool:
```
$ pic-edit -h
Edit compile-time options for a PIConGPU input set

Opens .param files in an input set with the default "$EDITOR".
If a .param file is not yet part of the input set but exists in the
defaults, it will be transparently added to the input set.

You must run this command inside an input directory.

The currently selected editor is: /bin/nano
You can change it via the "$EDITOR" environment variable.

usage: pic-edit <input>

Available <input>s:
components density dimension fieldBackground fieldSolver fileOutput flylite grid ionizationEnergies ionizer isaac laser mallocMC memory particle particleCalorimeter particleMerger physicalConstants precision pusher radiation radiationObserver seed species speciesAttributes speciesConstants speciesDefinition speciesInitialization starter synchrotronPhotons visColorScales visualization visualization bremsstrahlung
```

### To Do

- feedback? @ComputationalRadiationPhysics/picongpu-developers 
- [x] add `.rst` docs
- [x] change `pic-create`
- [x] `$EDITOR` in `picongpu.profile.example`s

### Future Ideas

- [ ] do not copy-in an input file if it is unchanged to defaults (try to *avoid* adding a new third-party dependency such as `diff`/`cmp`/...?) -> `grep -Fxvf file1 file2`

### DIFF
--- a/pic-create
+++ b/pic-create
@@ -21,9 +21,9 @@
 
 this_dir=$(cd $(dirname $0) && pwd)
 
-# only copy submit and params if we clone from default pic folder
-default_folder_to_copy="etc/picongpu include/picongpu/simulation_defines/param include/picongpu/simulation_defines/unitless lib"
-# if we clone a project we copy full include
+# files always cloned from defaults, can be overwritten by input set
+default_folder_to_copy="etc/picongpu lib"
+# if we clone an input set we copy from it
 folder_to_clone="etc/picongpu bin include/picongpu lib"
 files_to_copy="cmakeFlags executeOnClone"
 
--- a/pic-edit
+++ b/pic-edit
@@ -67,7 +67,7 @@ help()
 {
     echo "Edit compile-time options for a PIConGPU input set"
     echo ""
-    echo 'Opens .param files in an input set with the default "$EDITOR".'
+    echo 'Opens .param files in an input set with the default "EDITOR".'
     echo "If a .param file is not yet part of the input set but exists in the"
     echo "defaults, it will be transparently added to the input set."
     echo ""
@@ -79,7 +79,7 @@ help()
         test_editor="NOT FOUND"
     fi
     echo "The currently selected editor is: $test_editor"
-    echo 'You can change it via the "$EDITOR" environment variable.'
+    echo 'You can change it via the "EDITOR" environment variable.'
     echo ""
     echo "usage: pic-edit <input>"
     echo ""
@@ -148,9 +148,15 @@ use_editor="$(find_editor)"
 if [ $? -ne 0 ]
 then
     echo 'ERROR: Could not find a working text editor!' >&2
-    echo '       You might want to set the "$EDITOR" environment variable.' >&2
+    echo '       You might want to set the "EDITOR" environment variable.' >&2
     exit 3
 fi
 
 # open the files for editing
 $use_editor $file_paths
+editor_exit=$?
+
+# remove new files that are unchanged
+# TODO: grep -Fxvf file1 file2
+
+exit $editor_exit
--- a/pic-edit
+++ b/pic-edit
@@ -1,0 +1,156 @@
+#!/usr/bin/env bash
+#
+# Copyright 2017 Axel Huebl
+#
+# This file is part of PIConGPU.
+#
+# PIConGPU is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# PIConGPU is distributed in the hope that it will be useful,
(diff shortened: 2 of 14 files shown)


PrometheusPi:
@ax3l can we review this pull request or should we wait till the rest is added?

ax3l:
Tool can already be reviewed

pic-create
```diff
@@ -21,9 +21,9 @@
 
 this_dir=`dirname $0`
```
ax3l: note: use `this_dir=$(cd $(dirname $0) && pwd)`

pic-edit
```diff
@@ -0,0 +1,156 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PIConGPU.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
+this_dir=`dirname $0`
```
ax3l: note: use `this_dir=$(cd $(dirname $0) && pwd)`

pic-edit
```diff
@@ -0,0 +1,156 @@
...
+    echo ""
+    echo "You must run this command inside an input directory."
+    echo ""
+    test_editor="$(find_editor)"
+    if [ -z "$test_editor" ]
+    then
+        test_editor="NOT FOUND"
+    fi
+    echo "The currently selected editor is: $test_editor"
+    echo 'You can change it via the "$EDITOR" environment variable.'
```
psychocoderHPC: should be only `EDITOR` without `$`

pic-edit
```diff
@@ -0,0 +1,156 @@
...
+    done
+    # none found
+    return 3
+}
+
+help()
+{
+    echo "Edit compile-time options for a PIConGPU input set"
+    echo ""
+    echo 'Opens .param files in an input set with the default "$EDITOR".'
```
psychocoderHPC: should be only `EDITOR` without `$`

pic-edit
```diff
@@ -0,0 +1,156 @@
...
+        fi
+    fi
+done
+
+# get editor
+use_editor="$(find_editor)"
+if [ $? -ne 0 ]
+then
+    echo 'ERROR: Could not find a working text editor!' >&2
+    echo '       You might want to set the "$EDITOR" environment variable.' >&2
```
psychocoderHPC: should be only `EDITOR` without `$`

pic-edit
```diff
@@ -0,0 +1,156 @@
...
+    exit 3
+fi
+
+# open the files for editing
+$use_editor $file_paths
```
psychocoderHPC: should be `$(use_editor $file_paths)`

psychocoderHPC: After this line we can check if the file differ from the default file, if not we can remove the file from the parameter set:

diff a file without the diff dependency [stackoverflow link](https://stackoverflow.com/a/4079109)
```
grep -Fxvf file1 file2
```

psychocoderHPC: maybe we should store the result value and exit with the result of the editor at the end of this script.

ax3l: > $(...)

not sure if it should! are you sure?

> return value

good point!

ax3l:
@psychocoderHPC I added your review comments for all but the "future" point with the `grep` since I found no elegant way yet to store the "new-but-to-be-removed-files" without much code duplication. I guess just need to restructure the common part into a function again

ax3l:
@psychocoderHPC @PrometheusPi ready for review (& merge) :)

Issue #2221, opened 2017-08-27
pic-build error during execution


PrometheusPi:
While testing the new `pic-build` (see #2204) I encountered the following error:
Running `../../picongpu/pic-build` from within my project directory lead to:
```bash
build directory: .build
../../picongpu/pic-build: line 94: ../../picongpu/pic-configure: No such file or directory
```

However I can run `../../picongpu/pic-configure -h` from within the same directory. 

The reason for the error is that the script is already in the `.build` directory and thus can no access `pic-configure` with its relative path.

PR #2222, opened 2017-08-27
Tools: Allow Call With Relative Path

PrometheusPi:
This fixed the bug reported in issue #2221.
 
Allows to call the scripts `pic-create`, `pic-build`, `pic-configure` even with relative paths to the scripts instead of relying on the `$PATH` environment var.

### DIFF
--- a/pic-build
+++ b/pic-build
@@ -19,7 +19,7 @@
 # If not, see <http://www.gnu.org/licenses/>.
 #
 
-this_dir=`dirname $0`
+this_dir=$(cd $(dirname $0) && pwd)
 build_dir=".build"
 
 help()
--- a/pic-configure
+++ b/pic-configure
@@ -19,7 +19,7 @@
 # If not, see <http://www.gnu.org/licenses/>.
 #
 
-this_dir=`dirname $0`
+this_dir=$(cd $(dirname $0) && pwd)
 
 
 help()
--- a/pic-create
+++ b/pic-create
@@ -19,7 +19,7 @@
 # If not, see <http://www.gnu.org/licenses/>.
 #
 
-this_dir=`dirname $0`
+this_dir=$(cd $(dirname $0) && pwd)
 
 # only copy submit and params if we clone from default pic folder
 default_folder_to_copy="etc/picongpu include/picongpu/simulation_defines/param include/picongpu/simulation_defines/unitless lib"


ax3l:
This should not be necessary if you set PATH and execute the script from therein just as "pic-build"

ax3l:
In short: the reporting issue calls the scripts wrongly, a user should not know where any of the "pic-..." scripts are living since PATH abstracts that away

PrometheusPi:
Yes, this is true, but not all users want a fixed path to the python source in `$PATH`, since they might use various picongpu directories in parallel (due to parallel compilation, etc.).

Thus this adds the advantage of allowing accessing `pic-build` directly without breaking the usage for people accessing `pic-build` directly (via `$PATH`)

PrometheusPi:
Sorry - my fault - I think this **will break** the usage for uses accessing `pic-build` via `$PATH`. Will fix this.

ax3l:
Yes, but a parallel picongpu install requires more than selecting an other script, it needs a full environment. 

Most importantly, the scripts must not depend on $(pwd) of the caller. 

I am aware of the limitation of the current implementation of all those scripts, but right now you are really just using it wrongly (sorry) :)

PrometheusPi:
I think this is not an urgent problem ☺️ . We can discuss the script design when you are back.

ax3l:
Absolutely, but I am also annoyed that not both call works, we should fix this in a reliable manner.

pic-build
```diff
@@ -21,6 +21,7 @@
 
 this_dir=`dirname $0`
 build_dir=".build"
+this_dir=$(pwd)/$this_dir
```
psychocoderHPC: offline discussed: this is the way how to get the absolute path `this_dir=$(cd `dirname $0` && pwd)`

PrometheusPi:
@ax3l I followed @psychocoderHPC recommendation - This should solve the issue with absolute vs. relative path

ax3l:
Sure, will review it next week

PrometheusPi:
@ax3l No reason to hurry - have a look at it next week

pic-build
```diff
@@ -21,6 +21,7 @@
 
 this_dir=`dirname $0`
 build_dir=".build"
+this_dir=$(cd $this_dir && pwd)
```
ax3l: this is too verbose with double-assign and invalid-in-between and also affects more tools. I will force-push updates to this PR.

PR #2168, opened 2017-08-02
refactor picongpu::particles::startPosition

psychocoderHPC:
- use PMacc functor interface
  - change `operator()` interface
  - `QuietImpl` use an internal counter to place the particle
  - `RandomImpl` use `manipulators::generic::Rng` for the random number generator
- remove class `MacroParticleCfg`
- refactore `ParticlesInit.kernel` (use new startPosition functors)

# Tests

- [x] 3D KHI electrons positrons

PR #2125, opened 2017-07-10
refactor particle manipulators

psychocoderHPC:
- add newly implemented manipulators
  - group manipulators in `unary`, `binary` and `generic` functors
  - rewrite interfaces of all manipulators
  - add a interface description `IBinary` and `IUnary
- add filter
  - `IHandleValid`
  - `RelativeGlobalDomainPosition`
- add a interface xample for each functor/filter which can be configured
- update usage of the new functors/filter
- update example and default parameter files 
  - `particles.param`
  - `radiation.param`
- update documentation

related to #1288

# Notes for the reviewer

Please review all commit wise. I rewrote so much code thus the best way for a review would be to ignore the diff and read everything as if it is absolutely new code.

# Test

- [x] KHI runtime test

PR #2223, opened 2017-08-28
fix class down cast

### LINKED ISSUE #2215: RandomPosition Broken
The **random** in-cell position initialization `picongpu::particles::startPosition` in `dev` seems to be broken (CUDA backend) or defaulting to quiet or something else. Probably caused by #2168 or a refactoring the in the rng.

See the default, 1 GPU LWFA example at step 576

**0.3.0**

`/hplsim/scratch/huebl/lwfa-dev-031-0.3.0/simOutput/pngElectronsYX`
e_png_yx_0 5_000576

**dev** (4992bc0fdc07f6121e620e3f516cba3bb6208362)

`/hplsim/scratch/huebl/lwfa-dev-031/simOutput/pngElectronsYX`
e_png_yx_0 5_000576

Looks a bit as if this would have been a quiet start instead (but it should not be one).

psychocoderHPC:
fix #2215

`reinterpret_cast` is used to down cast a class. This is wrong and leads to wrong (not illegal) memory access within the class to the member's.

`static_cast` is used to solve the issue.

The bug was introduced with #2117, #2125 and #2168. 
The result of the bug is that the values within the member's of a class can be wrong, this can result in illegal memory access if the member is an offset within an array or in bad random number as in #2215.

# Tests

- [x] laser wake field default example on GPU

### DIFF
--- a/include/picongpu/particles/manipulators/generic/Free.hpp
+++ b/include/picongpu/particles/manipulators/generic/Free.hpp
@@ -146,7 +146,7 @@ namespace acc
             T_WorkerCfg const &
         )
         {
-            return acc::Free< Functor >( *reinterpret_cast< Functor * >( this ) );
+            return acc::Free< Functor >( *static_cast< Functor * >( this ) );
         }
     };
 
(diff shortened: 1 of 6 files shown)


psychocoderHPC:
@PrometheusPi I checked all `reinterpret_cast`'s withing PMacc and PIConGPU, all not changed casts must be reinterpret_cast's.

PR #2224, opened 2017-08-29
fix random start for CPU back-end

psychocoderHPC:
The position functor in `KernelFillGridWithParticles` is not correctly initialized therefore the random numbers between different cells within a supercell are correlated.
The bug is only triggered if the CPU backend `omp2b` is used. 
The bug was introduced with #2168.

- use virtual thread linear idx to initialize the position functor

### DIFF
--- a/include/picongpu/particles/ParticlesInit.kernel
+++ b/include/picongpu/particles/ParticlesInit.kernel
@@ -118,6 +118,7 @@ namespace picongpu
             using namespace mappings::threads;
 
             constexpr uint32_t frameSize = pmacc::math::CT::volume< SuperCellSize >::type::value;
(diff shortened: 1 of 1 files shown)


include/picongpu/particles/ParticlesInit.kernel
```diff
@@ -165,10 +171,13 @@ namespace picongpu
                     positionFunctor(
                         acc,
                         std::declval< DataSpace< simDim > const >( ),
-                        std::declval< WorkerCfg< numWorkers > const >( )
+                        /* cellsPerSupercell is used because each virtual worker
+                         * is creating **exact one** functor
```
PrometheusPi: TYPO exact**ly**

PR #2230, opened 2017-09-05
use fast `atomicAllExch` in `KernelFillGridWithParticles`

psychocoderHPC:
use fast `atomicAllExch` instead of slow `atomicExch`

All threads within the kernel reducing it's value to the same pointer, tehrefore the faster version of `atomicExch` can be used.

### DIFF
--- a/include/picongpu/particles/ParticlesInit.kernel
+++ b/include/picongpu/particles/ParticlesInit.kernel
@@ -331,10 +331,12 @@ namespace picongpu
 
                             numParsPerCellCtx[ idx ]--;
                             if( numParsPerCellCtx[ idx ] > 0 )
-                                atomicExch(
+                                nvidia::atomicAllExch(
+                                    acc,
                                     &finished,
-                                    0
-                                ); //one or more cell has particles to create
+                                    0,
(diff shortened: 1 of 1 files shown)


include/picongpu/particles/ParticlesInit.kernel
```diff
@@ -331,10 +331,12 @@ namespace picongpu
 
                             numParsPerCellCtx[ idx ]--;
                             if( numParsPerCellCtx[ idx ] > 0 )
-                                atomicExch(
+                                nvidia::atomicAllExch(
```
PrometheusPi: QUESTION: 
Is using the nvidia namespace for an general (alpaka) routine not counter-intuitive? 
This routine seems to be defined for various `acc` not only for `gpus`.

psychocoderHPC: The nvidia namespace is within PMacc but PMacc used alpaka under the hut even if the namespace is called nvidia.

PrometheusPi: But that is confusing 😉 - will this be changed in the future?

PR #2178, opened 2017-08-08
Marriage of PIConGPU and cupla/alpaka

psychocoderHPC:
This pull request contains all changes to get the PIConGPU core algorithm running on the CPU and GPU.
The feature set on the GPU is not changed compared to the current dev.
Some features and plugins are disabled if the CPU accelerator is used #2180.

Cfg files to start PIConGPU will not be updated to handle the plugins which are not ported. Please use `picongpu --help` to check all available plugins. 

# HowTo Compile

The usage of `pic-configure` changed, please see #2185

# Review

Please review everything commit wise.

# Todo
- [x] CMake Update to 3.7.0+ #2181
- [x] Boost Update to 1.62.0+ #2182
- [x] add *.rst HowTo compile for CPU & update docs on `pic-configure` #2185
- [x] compile/run PMacc test cases

CC-ing: @ComputationalRadiationPhysics/picongpu-developers

PR #2229, opened 2017-09-05
fix `ParticlesBox::setAsLastFrame()`

psychocoderHPC:
fix #2228

The function `setAsLAstFrame()` must be atomic between different grids.
With #2178 [this line](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2178/files#diff-82ce98f3d12c8577e81d5cc7c1a6245cR285) the function was wrongly changed to be atomic only between threads within a block.

- use `::alpaka::hierarchy::Threads{}` to `::alpaka::hierarchy::Grids{}`

# Tests

- [x] LWFA with `omp2b`

PR #2231, opened 2017-09-05
fix wrong used hierarchy for atomic operation

psychocoderHPC:
The shared memory histogram is reduced with the hierarchy `Threads`
to the global memory. This will produce raceconditions between blocks of a grid.

- use hierarchy `Blocks` instead of `Threads`

This bug was introduced with #2178.


psychocoderHPC:
I try to keep independent parts separate even if the type of the bug is equal to #2229

PR #2232, opened 2017-09-06
refactor `PushDataBox` and `ExchangePushDataBox`

psychocoderHPC:
- allow the user to control the hierarchy of the atomics for
  - `PushDataBix`
  - `ExchangePushDataBox`
- add documentation
- `KernelCopyGuardToExchange`: pass hierarchy to `pushN()` to controll the level for the atomic function.

This commit fix also the wrong used hierarchy in the method `push()` which was intruduced with #2178. This method is currently not used within PMacc therefore it not effects PIConGPU.

### DIFF
--- a/include/pmacc/particles/memory/boxes/ExchangePushDataBox.hpp
+++ b/include/pmacc/particles/memory/boxes/ExchangePushDataBox.hpp
@@ -54,18 +54,34 @@ public:
     {
     }
 
-    template< typename T_Acc >
+    /** give access to push N elements into the memory
+     *
+     * The method is threadsave within the e given alpaka hierarchy.
+     *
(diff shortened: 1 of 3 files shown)


include/pmacc/particles/memory/boxes/ExchangePushDataBox.hpp
```diff
@@ -54,18 +54,34 @@ class ExchangePushDataBox : public DataBox<PitchedBox<VALUE, DIM1> >
     {
     }
 
-    template< typename T_Acc >
+    /** give access to push N elements into the memory
+     *
+     * The method is threadsave within the e given alpaka hierarchy.
```
PrometheusPi: REQUIRED:
typo there is an extra "e"

PR #2233, opened 2017-09-06
cuSTL `Foreach` with lockstep support

psychocoderHPC:
- add functor `ForeachLockstep` to start a cuSTL functor which is lockstep ready

This functor is only a workaround up to all cuSTL functors are refactored and are lockstep ready.

- [x] check internal used mapper needs to be updated (tested with #2234)

### DIFF
--- a/include/pmacc/cuSTL/algorithm/kernel/Foreach.hpp
+++ b/include/pmacc/cuSTL/algorithm/kernel/Foreach.hpp
@@ -26,6 +26,7 @@
 #include "pmacc/math/vector/Int.hpp"
 #include "detail/SphericMapper.hpp"
 #include "detail/ForeachKernel.hpp"
+#include "pmacc/cuSTL/zone/SphericZone.hpp"
 #include "pmacc/forward.hpp"
 
 #include <boost/preprocessor/repetition/enum.hpp>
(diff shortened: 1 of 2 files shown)


include/pmacc/cuSTL/algorithm/kernel/Foreach.hpp
```diff
@@ -92,11 +92,55 @@ struct Foreach
...
+     * \param zone Accepts currently only a zone::SphericZone object (e.g. containerObj.zone())
+     * \param functor or lambdaFun either a functor with N arguments or a N-ary lambda function (e.g. _1 = _2)
+     * \param args cursor for the N-th data source (e.g. containerObj.origin())
+     *
+     * The functor is called for each worker within the zone.
+     * It is called like functor(*cursor0(cellBlockOffset), ..., *cursorN(cellBlockOffset))
+     *
+     */
+    template<
+        typename T_Zone,
```
ax3l: instead of `T_Zone` let us use `T_SphericZone` with it's `uint` signature as a free template to enforce that only spheric zones (see docstring) can be used.

include/pmacc/cuSTL/algorithm/kernel/Foreach.hpp
```diff
@@ -92,11 +92,55 @@ struct Foreach
...
+    uint32_t T_numWorkers,
+    typename BlockDim
+>
+struct ForeachLockstep
+{
+
+    /* operator()(zone, functor, cursor0, cursor1, ..., cursorN-1)
+     *
+     * \param zone Accepts currently only a zone::SphericZone object (e.g. containerObj.zone())
+     * \param functor or lambdaFun either a functor with N arguments or a N-ary lambda function (e.g. _1 = _2)
```
ax3l: `lambdaFun`ctor?

include/pmacc/cuSTL/algorithm/kernel/Foreach.hpp
```diff
@@ -92,11 +92,55 @@ struct Foreach
...
+     * \param args cursor for the N-th data source (e.g. containerObj.origin())
+     *
+     * The functor is called for each worker within the zone.
+     * It is called like functor(*cursor0(cellBlockOffset), ..., *cursorN(cellBlockOffset))
```
ax3l: C++ doxygen code highlight?

PR #2226, opened 2017-08-30
PMacc: cupla compatible RNG

psychocoderHPC:
- add forward declaration
- remove random methods:
  - Xor
  - XorMin
  - MRG32k3a
  - MRG32k3aMin
- add random method `AlpakaRand` which use the random method from the accelerator
- lockstep rewrite for kernel `InitRNGProvider`
- add `acc` to all accelerator dependent methods
- PIConGPU activate fast pmacc random number generator with stored state
- PMacc test: fix 2DDistribution ￼…
  - lockstep rewrite of kernel `RandomFiller`
  - fix that CUDA event is not recorded in the stream of the last kernel
  - reduce samples per worker from `1000` to `100`

alpaka supports only a predefined random number generator per accelerator. The generator can not be selected by the user. Therefore the generator those are heavenly depends on CUDA RNG states are removed. 
This will result in a larger state stored in the GPU memory for random numbers and maybe a performance decrease. 
I opened an [issue #390](https://github.com/ComputationalRadiationPhysics/alpaka/issues/390) in alpaka to provide different RNG generator methods to bring the full performance back.

The old implementation of `XorMin` was not able to create normal distributed random numbers, therefore I have not guarded the generator to be activated in the case we are using the CUDA back-end.

# Tests
- [x] 2DDistribution test CPU/GPU

PR #2234, opened 2017-09-06
enable ionization for CPU backend

psychocoderHPC:
- remove define which disabled the ionization code on CPU
- add documentation
- use new functor interface with `collectiveInit`
- use a source frame for each worker (instead of all worker in the block)
- extend interface of all functors used together with `CreateParticlesKernel` (add method `collectiveInit()`)

# Review

Please check the changes commit wise!

# Tests

- [x] LWFA test case 10 (CPU/GPU vs GPU 0.3.X)
- [x] rebase against #2233

### DIFF
--- a/include/picongpu/particles/bremsstrahlung/Bremsstrahlung.tpp
+++ b/include/picongpu/particles/bremsstrahlung/Bremsstrahlung.tpp
@@ -31,6 +31,7 @@
 #include <pmacc/algorithms/math/defines/sqrt.hpp>
 #include <pmacc/algorithms/math/defines/dot.hpp>
 #include <pmacc/algorithms/math/defines/cross.hpp>
+#include <pmacc/mappings/threads/WorkerCfg.hpp>
 
 
 namespace picongpu
@@ -76,13 +77,14 @@ template<
     typename T_ElectronSpecies,
     typename T_PhotonSpecies
 >
-template< typename T_Acc >
-DINLINE
-void Bremsstrahlung<T_IonSpecies, T_ElectronSpecies, T_PhotonSpecies>::init(
-    T_Acc const & acc,
+template<
+    typename T_Acc ,
+    typename T_WorkerCfg
+>
+DINLINE void Bremsstrahlung<T_IonSpecies, T_ElectronSpecies, T_PhotonSpecies>::collectiveInit(
+    const T_Acc & acc,
     const DataSpace<simDim>& blockCell,
-    const int& linearThreadIdx,
-    const DataSpace<simDim>& localCellOffset
+    const T_WorkerCfg & workerCfg
 )
 {
     /* caching of ion density field */
@@ -101,8 +103,8 @@ void Bremsstrahlung<T_IonSpecies, T_ElectronSpecies, T_PhotonSpecies>::init(
 
     ThreadCollective<
         BlockArea,
-        pmacc::math::CT::volume< typename BlockArea::SuperCellSize >::type::value
-    > collective( linearThreadIdx );
+        T_WorkerCfg::numWorkers
+    > collective( workerCfg.getWorkerIdx( ) );
     collective(
               acc,
               assign,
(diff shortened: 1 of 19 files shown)


psychocoderHPC:
ionizationcompare

The number of particles spawned on the cpu in peak ~5000 particles higher than on the gpu (release 0.3.0).
The new alpaka cuda version is in peak by a factor 1.13 slower than the release 0.3.0, this is coming from #2226 where the minimal state of the random number generator is removed.

ax3l:
code looks good!
the test would be good to be repeated with the check for free charge in the `FoilLCT` since the Hydrogen ionization physics in the LWFA example (with a strong `a_0`) is rather insensitive to potential bugs.

include/picongpu/particles/creation/creation.kernel
```diff
@@ -30,282 +30,458 @@
...
+                        )
+                        {
+                            if( 0 <= targetParIdCtx[ idx ] &&  targetParIdCtx[ idx ] < maxParticlesInFrame )
+                            {
+                                // each thread makes the attributes of its source particle accessible
+                                auto sourceParticle = sourceFrame[ linearIdx ];
+                                // each thread initializes an target particle if one should be created
+                                auto targetParticle = targetFrame[ targetParIdCtx[ idx ] ];
+
+                                // create an target particle in the new target particle frame:
```
n01r: ... `a target` ...

include/picongpu/particles/creation/creation.kernel
```diff
@@ -30,282 +30,458 @@
...
+                            uint32_t const idx
+                        )
+                        {
+                            if( targetParIdCtx[ idx ] >= maxParticlesInFrame )
+                            {
+                                targetParIdCtx[ idx ] -= maxParticlesInFrame;
+
+                                // each thread makes the attributes of its source particle accessible
+                                auto sourceParticle = sourceFrame[ linearIdx ];
+                                // each thread initializes an target particle if one should be created
```
n01r: ... `a target` ...

include/picongpu/particles/creation/creation.kernel
```diff
@@ -30,282 +30,458 @@
...
+                                auto targetParticle = targetFrame[ targetParIdCtx[ idx ] ];
+
+                                // create an target particle in the new target particle frame:
```
n01r: ... `a target` ...

include/picongpu/particles/bremsstrahlung/Bremsstrahlung.tpp
```diff
@@ -76,13 +77,14 @@ template<
     typename T_ElectronSpecies,
     typename T_PhotonSpecies
 >
-template< typename T_Acc >
-DINLINE
-void Bremsstrahlung<T_IonSpecies, T_ElectronSpecies, T_PhotonSpecies>::init(
-    T_Acc const & acc,
+template<
+    typename T_Acc ,
```
n01r: The docstrings are missing here as well but the `T_Acc acc` is probably in every kernel, right? So the question arises if it makes sense to document its every occurence. I'd prefer it personally, though.

n01r:
> ... the test would be good to be repeated with the check for free charge in the FoilLCT ...

I agree with @ax3l on the testing. A hydrogen-only setup will not be prone to many potential mistakes as there is just one ionization process per ion.

include/picongpu/particles/bremsstrahlung/Bremsstrahlung.tpp
psychocoderHPC: the documentation is in the `hpp` file

n01r: ah, alright, sorry :)

psychocoderHPC:
# Test

script to create the plots will be provided by @ax3l

ionization

include/picongpu/particles/creation/creation.kernel
```diff
@@ -30,282 +30,458 @@
...
+            sourceBox( sourceBox ),
+            targetBox( targetBox ),
+            particleCreator( particleCreator ),
+            guardSuperCells( guardSuperCells )
+        { }
+
+        /** Goes over all frames and calls `ParticleCreator`
+         *
+         * @param blockCell n-dim. block offset (in cells) relative to the origin
+         *                  of the local domain plus guarding cells
```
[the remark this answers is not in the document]
ax3l: added

PR #2237, opened 2017-09-07
remove `LiveViewPlugin` 

psychocoderHPC:
- remove `LiveViewPlugin.hpp`
- remove all dependencies only used by the plugin
- remove all documentations for the plugin

This plugin was not used since years and the server is not available there it is removed.


ax3l:
wuhu, this removes our *direct* zlib dependency :)
(CMake, Install notes, ...)

ax3l:
ah, PNGwriter is currently exposing `zlib.h` in it's header and has no proper CMake targets yet, so we let it in for now.

PR #2240, opened 2017-09-11
lockstep kernel refactoring `KernelRadiationParticles`

psychocoderHPC:
- activate Radiation accelerator `omp2b`

# Tests

- [x] default bunch test case 3 with accelerator `omp2b` (plot: `plotRadiation totalRad/e_radiation_3000.dat`)
- [x] default bunch test case 3 with accelerator `cuda:35` (plot: `plotRadiation totalRad/e_radiation_3000.dat`)

### DIFF
--- a/include/picongpu/plugins/PluginController.hpp
+++ b/include/picongpu/plugins/PluginController.hpp
@@ -47,8 +47,6 @@
 #   include "picongpu/plugins/ChargeConservation.hpp"
 #   include "picongpu/plugins/particleMerging/ParticleMerger.hpp"
 #   if(ENABLE_HDF5 == 1)
-#       include "picongpu/plugins/radiation/parameters.hpp"
-#       include "picongpu/plugins/radiation/Radiation.hpp"
 #       include "picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp"
 #       include "picongpu/plugins/PhaseSpace/PhaseSpaceMulti.hpp"
 #       include "picongpu/plugins/makroParticleCounter/PerSuperCell.hpp"
@@ -66,7 +64,9 @@
 #endif
 
 #if (ENABLE_HDF5 == 1)
-#include "picongpu/plugins/hdf5/HDF5Writer.hpp"
+#   include "picongpu/plugins/radiation/parameters.hpp"
+#   include "picongpu/plugins/radiation/Radiation.hpp"
+#   include "picongpu/plugins/hdf5/HDF5Writer.hpp"
 #endif
 
 #include "picongpu/plugins/ResourceLog.hpp"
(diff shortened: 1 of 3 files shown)


psychocoderHPC:
8 x GPU:
total_radiation_3000_gpu

ax3l:
ping @PrometheusPi

PrometheusPi:
@psychocoderHPC Thanks for the plots they look really great. 
Do you still see the factor 10x slow down compared to theoretical peak performance on CPUs? 
Do you want to investigate this? (I assume you tested this on the laser nodes with AMD 16-Core Opteron 2,3 GHz - if you run the same simulation on the  	Intel 16-Core Xeon 2,3 GHz on hydra, you might already see a big difference (L2 AMD 16MB, Intel 4 MB, L3 AMD 16 MB, Intel 40 MB) that would point out that this slow down is related to the cache utilization).

include/picongpu/plugins/PluginController.hpp
```diff
@@ -66,7 +64,9 @@
 #endif
 
 #if (ENABLE_HDF5 == 1)
```
PrometheusPi: REQUIRED:
please combine both `#if (ENABLE_HDF5 == 1) ...  #endif` parts

psychocoderHPC:
@PrometheusPi Normalized to the peak performance there is a factor 10 gab between CPU and GPU.
I will not investigate time to look into it.

include/picongpu/plugins/radiation/Radiation.kernel
```diff
@@ -56,430 +58,482 @@
...
 
-        const int blockSize=pmacc::math::CT::volume<Block>::type::value;
-        // vectorial part of the integrand in the Jackson formula
-        PMACC_SMEM( acc, real_amplitude_s, memory::Array< vector_64, blockSize > );
+            constexpr uint32_t frameSize = pmacc::math::CT::volume< SuperCellSize >::type::value;
+            constexpr uint32_t numWorker = T_numWorkers;
 
-        // retarded time
-        PMACC_SMEM( acc, t_ret_s, memory::Array< picongpu::float_64, blockSize > );
+            using FrameType = typename ParBox::FrameType ;
```
PrometheusPi: QUESTION:
Does `FrameType` replaces `particlesInFrame` from the old code?
Could you  port the comment as well?

PrometheusPi: same one line below

include/picongpu/plugins/radiation/Radiation.kernel
```diff
@@ -56,430 +58,482 @@
...
+             * through all particles and calculates the radiation for one direction
+             * for all frequencies
+             */
+            constexpr int blockSize = pmacc::math::CT::volume<SuperCellSize>::type::value;
 
+            // vectorial part of the integrand in the Jackson formula
+            PMACC_SMEM( acc, real_amplitude_s, memory::Array< vector_64, blockSize > );
 
-        const int theta_idx = blockIdx.x; //blockIdx.x is used to determine theta
-        const uint32_t linearThreadIdx = threadIdx.x; // used for determine omega and particle id
```
PrometheusPi: Question:
The second palatalization range was removed (`linearThreadIdx`) - how does this work?

include/picongpu/plugins/radiation/Radiation.kernel
```diff
@@ -56,430 +58,482 @@
...
-
-            /*
-             * The Master process (thread 0) in every thread block is in
-             * charge of loading a frame from
-             * the current super cell and evaluate the total number of
-             * particles in this frame.
+            /* number of super cells on GPU per dimension (still including guard cells)
+             * remove both guards from count [later one sided guard needs to be added again]
              */
-            if (linearThreadIdx == 0)
```
PrometheusPi: QUESTION:
How do we get the frame know?

include/picongpu/plugins/radiation/Radiation.kernel
PrometheusPi: It now called: `linearIdx` and is defined below.

PrometheusPi:
The above spectra or both from the current pull request (but computed on GPU and CPU), the results for approximately 0.3.0 branch are given below:
spectraoutput_pr2240
The agreement is great 👍 

(please be aware that both **axis** ranges and labels are **wrong**)

PR #2242, opened 2017-09-11
add HZDR cluster `hydra`

psychocoderHPC:
- add `tpl` file for the default queue
- add example profile for the default queue
- add link to documentation `*.rst`

The default queue of the hydra cluster has mixed nodes with different main memory settings (64GB - 256GB). To avoid side effect only  nodes with 256GB main memory are used.


etc/picongpu/hydra-hzdr/default_picongpu.profile.example
```diff
@@ -0,0 +1,65 @@
...
+export MY_MAIL="someone@example.com"
+export MY_NAME="$(whoami) <$MY_MAIL>"
+
+export PATH=$PATH:$PICSRC
+export PATH=$PATH:$PICSRC/src/splash2txt/build
+export PATH=$PATH:$PICSRC/src/tools/bin
+
+export PYTHONPATH=$PICSRC/src/tools/lib/python:$PYTHONPATH
+
+# Development #################################################################
```
ax3l: please remove this. users do not need this and it complicates installs

etc/picongpu/hydra-hzdr/default_picongpu.profile.example
```diff
@@ -0,0 +1,65 @@
...
+
+# Architecture flags ##########################################################
+#
+export CXXFLAGS="$CXXFLAGS -march=ivybridge -mtune=ivybridge"
+
+# Environment #################################################################
+#
+alias getNode='qsub -I -q default -lwalltime=00:30:00 -lnodes=1:ppn=32'
+
+export PICSRC=/home/`whoami`/src/picongpu
```
ax3l: please use `$(whoami)` now

etc/picongpu/hydra-hzdr/default_picongpu.profile.example
```diff
@@ -0,0 +1,65 @@
...
+        # Debug Tools
+        #module load gdb
+        #module load valgrind/3.8.1
+
+#       unset MODULES_NO_OUTPUT
+fi
+
+# Architecture flags ##########################################################
+#
+export CXXFLAGS="$CXXFLAGS -march=ivybridge -mtune=ivybridge"
```
ax3l: Future: My general idea is more to add a configuration env option such as `PIC_ARCH` (for `pic-configure`) for queues_clusters in this file.

ax3l:
this will be useful in `dev` as soon as #2243 is in, but we can already merge it
@PrometheusPi do you mind taking over? :)

etc/picongpu/hydra-hzdr/default.tpl
```diff
@@ -0,0 +1,83 @@
...
+
+
+# PIConGPU batch script for hydra PBS batch system
+
+#PBS -q !TBG_queue
+#PBS -l walltime=!TBG_wallTime
+# Sets batch job's name
+#PBS -N !TBG_jobName
+#PBS -l nodes=!TBG_nodes:ppn=!TBG_coresPerNode
+#PBS -l mem=256GB
```
PrometheusPi: QUESTION:
Do you gibe memory per node (or per core)? From slum I know, that per core is used.

ax3l: in PBS it should be per node, otherwise this should not have been able to be scheduled on the available resources (the ones he mentions in the PR description)

ax3l: in [this ref](https://www.osc.edu/supercomputing/batch-processing-at-osc/pbs-directives-summary) it's *for all nodes* ^^

PrometheusPi: Thanks for the reference - there it states "all memory across all nodes" thus this should be multiplied with the number of nodes requested. This only ensures 256GB/node nodes if the node number is below or equal to 4. With higher node numbers, we could land on a 64GB/node machine. 😱

PrometheusPi: @psychocoderHPC Please multiply this number with the number of nodes requested.

PR #2243, opened 2017-09-11
pic-configure: backend

ax3l:
The `-a|--arch` flag should now read `-b|--backend` since we select whole (alpaka) *compute backends* (with optional architectures) now.

When this flag is not set during `pic-configure` (which is also called from `pic-build` and `pic-compile`) then the new environment variable `PIC_BACKEND` allows to set a convenient default. This is useful for `picongpu.profile` settings per cluster queue.

Documentation, examples and profiles have been updated accordingly. If the `cuda` backend has been selected, our CMake scripts already default to the oldest supported CUDA architecture (currently sm 2.0 with CUDA 8).


psychocoderHPC:
please remove [this](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2243/files#diff-c60d7e81ef8e4c665cd91e8d7c26bf48R33) line if we switch to the new variable.

share/picongpu/examples/WarmCopper/cmakeFlags
```diff
@@ -29,8 +29,8 @@
 #   - start with zero index
 #   - increase by 1, no gaps
 
-flags[0]="-DALPAKA_CUDA_ARCH=20 -DCMAKE_DISABLE_FIND_PACKAGE_ISAAC=True"
-flags[1]="-DALPAKA_CUDA_ARCH=20 -DCMAKE_DISABLE_FIND_PACKAGE_ISAAC=True -DPARAM_OVERWRITES:LIST='-DPARAM_ENABLE_PUSHER=1;-DPARAM_ENABLE_CURRENT=1'"
+flags[0]="-DCMAKE_DISABLE_FIND_PACKAGE_ISAAC=True"
```
psychocoderHPC: Only a question which is not related to this PR: Why do you disable ISAAC explicit?

ax3l:
probably needs an update of the compile suite...

share/picongpu/examples/WarmCopper/cmakeFlags
ax3l: I can't remember, probably to cut compile time for this example :)

pic-configure
```diff
@@ -73,20 +74,29 @@ get_backend_flags()
...
     exit 1
 fi
 
 # configure a specific preset in the cmakeFlags file
+# note: can be overwritten with a command line flag
 cmakeFlagsNr=0
 
+# set a default backend (and architecture) if supplied via environment var
+# note: can be overwritten with a command line flag
+pic_backend=${PIC_BACKEND:-""}
```
ax3l: @psychocoderHPC I just realized that we could, if we want, default to `"cuda"` here.

ax3l:
@ComputationalRadiationPhysics/picongpu-developers we are really sorry to tell you that `-a` is `-b` in the next release 🤦‍

`pic-configure` & `pic-build` can select an alpaka *backend* now where the *architecture* is just a sub-option. follow-up to #2185

pic-configure
psychocoderHPC: offline discussed: will be added in a follow up PR

PR #2249, opened 2017-09-13
Hypnos & Taurus: Profile Examples Per Queue

ax3l:
Add a profile example per queue for Hypnos (HZDR) and Taurus (TU Dresden).

Allows to optimize the compile for the exact right architecture by default and cleans up confusion, e.g. for upcoming KNL profiles, and requires less user interaction.


etc/picongpu/lawrencium-lbnl/picongpu.profile.example
```diff
@@ -35,7 +35,7 @@ alias allocK20='salloc --time=0:30:00 --nodes=1 --ntasks-per-node=1 --cpus-per-t
 alias allocFermi='salloc --time=0:30:00 --nodes=1 --ntasks-per-node=2 --cpus-per-task=6 --partition mako_manycore'
 
 export PICSRC=$HOME/src/picongpu
-export PIC_BACKEND="cuda:35"
+export PIC_BACKEND="cuda:20"
```
ax3l: that was a typo in #2243, see last commit message

PR #2254, opened 2017-09-14
Inefficient CMake Compile

psychocoderHPC:
The PIConGPU CmakeLists.txt is separating host only and accelerator source files.
In the current implementation the host only and accelarator code is compiled twice (with host and device compiler).
The result of this bug is that compiling takes up to a factor of two longer than needed and all symboles are defined twice which can lead in linker issues.
This bug was introduced with #2178.

Fix: compile `main.cpp` with the device compiler and all other files with the host compiler

### DIFF
--- a/include/picongpu/CMakeLists.txt
+++ b/include/picongpu/CMakeLists.txt
@@ -378,7 +378,6 @@ include_directories(BEFORE ${PIC_EXTENSION_PATH}/include)
 list(APPEND ACCSRCFILES
     "${CMAKE_CURRENT_SOURCE_DIR}/main.cpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/versionFormat.cpp"
-    #"${CMAKE_CURRENT_SOURCE_DIR}/plugins/common/particlePatches.cpp"
 )
 
 # host-only sources
--- a/include/picongpu/CMakeLists.txt
+++ b/include/picongpu/CMakeLists.txt
@@ -374,7 +374,16 @@ include_directories(BEFORE ${PIC_EXTENSION_PATH}/include)
 # Compile & Link PIConGPU
 ################################################################################
 
+# files with accelerator code (kernel and API calls)
+list(APPEND ACCSRCFILES
+    "${CMAKE_CURRENT_SOURCE_DIR}/main.cpp"
+    "${CMAKE_CURRENT_SOURCE_DIR}/versionFormat.cpp"
+    #"${CMAKE_CURRENT_SOURCE_DIR}/plugins/common/particlePatches.cpp"
+)
+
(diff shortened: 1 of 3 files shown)


include/picongpu/CMakeLists.txt
```diff
@@ -384,7 +386,7 @@ target_link_libraries(picongpu-hostonly PUBLIC ${LIBS})
 
 
 cupla_add_executable(picongpu
-    ${SRCFILES}
+    "main.cpp"
```
ax3l: `${CMAKE_CURRENT_SOURCE_DIR}/main.cpp` ?

include/picongpu/CMakeLists.txt
```diff
@@ -375,6 +375,8 @@ include_directories(BEFORE ${PIC_EXTENSION_PATH}/include)
 ################################################################################
 
 file(GLOB_RECURSE SRCFILES "*.cpp")
```
ax3l: maybe we just set a "DEVICESRCFILES" variable here so we have it all together (and not scattered)

include/picongpu/CMakeLists.txt
psychocoderHPC: Yes must be the absolute pass, I missed to add my last change^^

include/picongpu/CMakeLists.txt
ax3l:
```cmake
# host-only sources
file(GLOB_RECURSE SRCFILES "*.cpp")
# device sources containing kernels
set(DEVICESRCFILES "${CMAKE_CURRENT_SOURCE_DIR}/main.cpp")

# separate host only and accelerator code
list(REMOVE_ITEM SRCFILES ${DEVICESRCFILES})
```

psychocoderHPC: note `list(REMOVE_ITEM SRCFILES ${ACCSRCFILES})` is an undocumented feature in cmake

include/picongpu/CMakeLists.txt
```diff
@@ -374,17 +374,25 @@ include_directories(BEFORE ${PIC_EXTENSION_PATH}/include)
 # Compile & Link PIConGPU
 ################################################################################
 
+# files with accelerator code (kernel and API calls)
+list(APPEND ACCSRCFILES
+    "${CMAKE_CURRENT_SOURCE_DIR}/main.cpp"
+    "${CMAKE_CURRENT_SOURCE_DIR}/versionFormat.cpp"
+    #"${CMAKE_CURRENT_SOURCE_DIR}/plugins/common/particlePatches.cpp"
```
ax3l: ;-) outdated

PR #2256, opened 2017-09-14
Docs: Streamline Install

ax3l:
Streamlines the install instructions.

Organizes the install instructions to make them more clear.

Also adresses commonly seen user problems:
- only select *one* install path
- remove confusing recommendations for systems
- prefer [spack](https://github.com/LLNL/spack) over self-build in order of appearance

### DIFF
--- a/docs/source/install/instructions/spack.rst
+++ b/docs/source/install/instructions/spack.rst
@@ -1,0 +1,65 @@
+.. _install-spack:
+
+.. seealso::
+
+   You will need to understand how to use `the terminal <http://www.ks.uiuc.edu/Training/Tutorials/Reference/unixprimer.html>`_.
+
+Spack
+-----
+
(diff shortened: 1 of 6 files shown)


docs/source/install/instructions/spack.rst
```diff
@@ -0,0 +1,65 @@
...
+   spack install picongpu %gcc@5.4.0
+
+Use PIConGPU
+^^^^^^^^^^^^
+
+PIConGPU can now be loaded with
+
+.. code-block:: bash
+
+   spack load picongpu %gcc@5.4.0
```
ax3l: maybe repeat that one needs `spack load gcc@5.4.0` first

ax3l: but I can also address that in the PIConGPU recipe for now

PR #2257, opened 2017-09-15
check cells needed for stencils

psychocoderHPC:
- PIConGPU: add static assert to check if a stencil fits into a supercell
- PMacc: add specialization to get the minimum component of a compile time vector


include/picongpu/fields/currentDeposition/VillaBune/CurrentVillaBune.hpp
```diff
@@ -271,6 +271,12 @@ struct GetMargin<picongpu::currentSolver::VillaBune<T_ParticleShape> >
 {
     typedef ::pmacc::math::CT::Int < 1, 1, 1 > LowerMargin;
     typedef ::pmacc::math::CT::Int < 2, 2, 2 > UpperMargin;
+
+    PMACC_CASSERT_MSG(
+        __VillaBune_supercell_is_to_small_for_stencil,
+        pmacc::math::CT::min< SuperCellSize >::type::value >= 1 &&
```
ax3l: this is weird, logically this expression will only check the 2nd line `>= 2`

psychocoderHPC: will remove the first line

PR #2258, opened 2017-09-15
ionization: speedup particle creation

psychocoderHPC:
Avoid that electrons are created within two independent steps.
This will increase the concurrency and increase the bandwidth utilization.

# Tests

- [x] default foil example

### DIFF
--- a/include/picongpu/particles/creation/creation.kernel
+++ b/include/picongpu/particles/creation/creation.kernel
@@ -30,6 +30,7 @@
 #include <pmacc/math/vector/Int.hpp>
 #include <pmacc/nvidia/atomic.hpp>
 #include <pmacc/memory/shared/Allocate.hpp>
+#include <pmacc/memory/Array.hpp>
 #include <pmacc/mappings/threads/ForEachIdx.hpp>
 #include <pmacc/mappings/threads/IdxConfig.hpp>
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
# Foil example (cpu vs gpu vs dev)
ionization

include/picongpu/particles/creation/creation.kernel
```diff
@@ -382,58 +399,24 @@ namespace creation
...
-                                    acc,
-                                    targetFrame,
-                                    block
-                                );
-                                newFrameFillLvl -= maxParticlesInFrame;
+                                 newFrameFillLvl -= maxParticlesInFrame;
+                                 // copy the not filled frame pointer to the beginning
+                                 targetFrames[ 0 ] = targetFrames[ 1 ];
+                                 // reset second frame
+                                 targetFrames[ 1 ] = nullptr;
```
n01r: By removing the second creation step the documentation `< CREATE 1 >` can be just changed to `< CREATE >` and `< CREATE 2 >` [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2258/files#diff-adb193a5d6509402921e4732bc77dd78R360) can be removed

psychocoderHPC: changes to `< create>`

PR #2259, opened 2017-09-15
FoilLCT: Plot Densities

ax3l:
Add a plot script to benchmark the ionization physics in the FoilLCT example.

Useful to compare various runs against each other, e.g.
- with different ionizers
- with different backends

We used a prototype of this script already in the last pull requests for verification.

Example usage:
```bash
plot_electron_density.py /bigdata/hplsim/scratch/widera/buildPIC/foil/runs/{012_k80_doubleFlush,013_laser_doubleFlush}
```
ne_2000

```bash
plot_electron_density.py /bigdata/hplsim/scratch/widera/buildPIC/foil/runs/{012_k80_doubleFlush,013_laser_doubleFlush} --step 1000 --sum
```
sum_1000

Tested with Python 2.7 and 3.5

### DIFF
--- a/share/picongpu/examples/FoilLCT/bin/plot_charge_density.py
+++ b/share/picongpu/examples/FoilLCT/bin/plot_charge_density.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python
+#
+"""
+This file is part of the PIConGPU.
+
+Copyright 2017 PIConGPU contributors
+Authors: Axel Huebl
+License: GPLv3+
+"""
+import matplotlib.pyplot as plt
+import h5py as h5
+import numpy as np
+import argparse
+import os
+
+
+parser = argparse.ArgumentParser(
+    description='Compare the electron charge density of FoilLCT simulations'
+)
+parser.add_argument(
+    'run_directory',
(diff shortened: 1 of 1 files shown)


share/picongpu/examples/FoilLCT/bin/plot_charge_density.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env python
+#
+"""
+This file is part of the PIConGPU.
+
+Copyright 2017 PIConGPU contributors
+Authors: Axel Huebl
+License: GPLv3+
```
psychocoderHPC: Should we add the full license header like in all other tools?

ax3l:
In the last check ins of python tools we just kept it as simple as that one. I think that is ok.

share/picongpu/examples/FoilLCT/bin/plot_charge_density.py
psychocoderHPC: https://github.com/ComputationalRadiationPhysics/picongpu/pull/2259#issuecomment-330198911

PR #2260, opened 2017-09-15
Added doxygen documentation for isaac.param file

theZiz:
Tried to add some useful information for the `isaac.param` as demanded in https://github.com/ComputationalRadiationPhysics/picongpu/issues/1982 .

### DIFF
--- a/include/picongpu/simulation_defines/param/isaac.param
+++ b/include/picongpu/simulation_defines/param/isaac.param
@@ -24,7 +24,7 @@
  * the PIC simulation can be observed while it is running avoiding the time
  * consuming writing and reading of simulation data for the classical post
  * processing of data.
- * 
+ *
  * ISAAC can directly visualize natives fields like the E or B field, but
  * density fields of particles need to be calculated from PIConGPU on the fly
  * which slightly increases the runtime and the memory consumption. Every
--- a/include/picongpu/simulation_defines/param/isaac.param
+++ b/include/picongpu/simulation_defines/param/isaac.param
@@ -19,8 +19,15 @@
 
 /** @file
  *
- * Definition of native fields and density fields of particles which shall be
- * visualized with ISAAC.
+ * Definition which native fields and density fields of particles will be
+ * visualizable with ISAAC. ISAAC is an in-situ visualization library with which
+ * the PIC simulation can be observed while it is running avoiding the time
+ * consuming writing and reading of simulation data for the classical post
+ * processing of data.
+ * 
+ * ISAAC can directly visualize natives fields like the E or B field, but
+ * density fields of particles need to be calculated from PIConGPU beforehand
+ * which slightly increases the runtime.
  *
  */
 
@@ -31,7 +38,7 @@ namespace picongpu
 namespace isaacP
 {
 
-    /** Compile time sequence of native fields of PIConGPU which shall be
+    /** Intermediate list of native fields of PIConGPU which shall be
      *  visualized. */
     using Native_Seq = MakeSeq_t<
         FieldE,
@@ -39,7 +46,7 @@ namespace isaacP
         FieldJ
     >;
 
-    /** Compile time sequence of particle species, from which density fields
+    /** Intermediate list of particle species, from which density fields
      *  shall be created at runtime to visualize them. */
     using Density_Seq = bmpl::transform<
         VectorAllSpecies,
--- a/include/picongpu/simulation_defines/param/isaac.param
+++ b/include/picongpu/simulation_defines/param/isaac.param
@@ -17,6 +17,13 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
+/** @file
+ *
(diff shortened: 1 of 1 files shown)


include/picongpu/simulation_defines/param/isaac.param
```diff
@@ -17,27 +17,37 @@
...
+
 #pragma once
 
 namespace picongpu
 {
 namespace isaacP
 {
 
-    // Native fields to visualize
+    /** Compile time sequence of native fields of PIConGPU which shall be
```
psychocoderHPC: Maybe: `Intermediate list of native ...`

I would skip the information that this is a compile time list.

include/picongpu/simulation_defines/param/isaac.param
```diff
@@ -17,27 +17,37 @@
...
+     *  visualized. */
     using Native_Seq = MakeSeq_t<
         FieldE,
         FieldB,
         FieldJ
     >;
 
-    // Creating a density field for every species to visualize
+    /** Compile time sequence of particle species, from which density fields
```
psychocoderHPC: Maybe: `Intermediate list of native ...`

include/picongpu/simulation_defines/param/isaac.param
```diff
@@ -17,27 +17,37 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
+/** @file
+ *
+ * Definition of native fields and density fields of particles which shall be
```
ax3l: It would be great if you describe here in a sentence or two what the `ISAAC` plugin *itself* is doing for a user.

at the end, the description of `@file` is what will show up in the manual (the in-code docs below will later not be rendered anymore in the manual).

Explain the user that ISAAC does *in situ* visualization of fields and what needs to be known simulation-setup wise, such as that ISAAC operates in-memory for native fields, etc.

Concepts and detailed techniques are explained in the manual in the plugins section, but here prepare the user to do the right decisions for the options below, so they do not blow up their memory or do stupid things.

theZiz: Hm, I thought the in-code doc is also be rendered: http://picongpu.readthedocs.io/en/latest/usage/param/core.html#grid-param.

I will add some details ;)

theZiz:
Please have a look at my changes. :smiley:

include/picongpu/simulation_defines/param/isaac.param
```diff
@@ -17,27 +17,44 @@
...
+/** @file
+ *
+ * Definition which native fields and density fields of particles will be
+ * visualizable with ISAAC. ISAAC is an in-situ visualization library with which
+ * the PIC simulation can be observed while it is running avoiding the time
+ * consuming writing and reading of simulation data for the classical post
+ * processing of data.
+ * 
+ * ISAAC can directly visualize natives fields like the E or B field, but
+ * density fields of particles need to be calculated from PIConGPU beforehand
```
ax3l: beforehand -> on the fly

include/picongpu/simulation_defines/param/isaac.param
```diff
@@ -17,27 +17,44 @@
...
+ * which slightly increases the runtime.
```
ax3l: the runtime -> the memory consumption

theZiz: Really? The calculation is for free? And I thought the memory is hold back for such fields anyway?

ax3l: Of course the calcs are also not for free, but the more severe point a user has to be aware of is that the memory is less, because it influences load balancing/GPU partitioning, global simulation size, etc.

But when I think about it: you use the regular FieldTmp slots for your derived fields and only one at a time, so that does not change the memory. so forget what I said.

theZiz: Well... Technically as I need to access all fields **at the same time** I copy them nevertheless. So yes, more particles need more memory. But this memory is not stolen from PIConGPU, as PIConGPU reserves afair 400 MB for such cases. I don't know where FieldTmp resists, but ISAAC's copies are part of this 400 MB and if too many particles are activated ISAAC will crash the simulation while allocating memory for the copies... I should mention this, also where to increase the amount of not-libPMacc-handled-memory. *cough*

theZiz: I remember something like all the GPU memory is allocated __after__ all plugins are initialized. Is this true? In that case ISAAC will indeed steal memory from PIConGPU - but not crash.

psychocoderHPC: Yes plugins allocated its memory first and the rest is for the particles. Therefore each *non persistent* field reduce the available memory for particles.

theZiz: Okay, will document it this way

theZiz:
`./include/picongpu/simulation_defines/param/isaac.param contains EOL white spaces!` Ups?

Issue #2264, opened 2017-09-18
python string formatting broken


PrometheusPi:
The pull request #2028 broke the radiation amplitude loader for python. 
The hdf5 data handler does not replace times correctly anymore.

PrometheusPi:
There are several more errors regarding string formatting.

PrometheusPi:
closed with #2265

PR #2265, opened 2017-09-18
Radiation: fix wrong python formatting

PrometheusPi:
This pull request  solves issue #2264.

With #2028 there were several strings separated to make their length agree with PEP8. Due to this several formatting processes were broken. 
This pull request fixes these where it is necessary.

### DIFF
--- a/src/tools/bin/smooth.py
+++ b/src/tools/bin/smooth.py
@@ -74,8 +74,8 @@ def makeOddNumber(number, larger=True):
         else:
             return number - 1
     else:
-        error_msg = "ERROR: number (= {}) neather odd " + \
-                    "nor even".format(number)
+        error_msg = ("ERROR: number (= {}) neither odd " +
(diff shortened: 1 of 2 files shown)


src/tools/bin/smooth.py
```diff
@@ -74,8 +74,8 @@ def makeOddNumber(number, larger=True):
         else:
             return number - 1
     else:
-        error_msg = "ERROR: number (= {}) neather odd " + \
-                    "nor even".format(number)
+        error_msg = ("ERROR: number (= {}) neather odd " +
```
n01r: neather --> *neither*

PrometheusPi: @n01r Thanks for spotting the typo - it is fixed now.

PrometheusPi:
@ax3l Can it be that our continuous integration stopped working 3 hours ago?

PR #2266, opened 2017-09-18
tbg: Add -f | --force

jkelling:
This change provides the option to force tbg to accept an already existing directory for a new job. This is useful when tbg is used in other workflows, where tbg is only used to submit the job, but the working directory is prepared by the workflow prior to the invocation of tbg.

The changed code also produces a warning when the '-f' option causes tbg to actually use an existing folder.

### DIFF
--- a/src/tools/bin/tbg
+++ b/src/tools/bin/tbg
@@ -236,7 +236,7 @@ help()
     echo "create a new folder for a batch job and copy in all important files"
     echo ""
     echo "usage: tbg -c [cfgFile] [-s [submitsystem]] [-t [templateFile]]"
-    echo "          [-o \"VARNAME1=10 VARNAME2=5\"] [-h]"
+    echo "          [-o \"VARNAME1=10 VARNAME2=5\"] [-f] [-h]"
     echo "          [projectPath] destinationPath"
     echo ""
     echo "-c | --cfg      [file]         - Configuration file to set up batch file."
@@ -250,7 +250,7 @@ help()
     echo "                                 spaces within the right side of assign are not allowed"
     echo "                                 e.g. -o \"VARNAME1=10 VARNAME2=5\""
     echo "                                 Overwriting is done after cfg file was executed"
-    echo "-f | --force                   - Start new job even if TBG_dstPath exists."
+    echo "-f | --force                   - Override if TBG_dstPath exists."
     echo "-h | --help                    - Shows help (this output)."
     echo ""
     echo "[projectPath]                  - Project directory containing source code and"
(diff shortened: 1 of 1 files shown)


ax3l:
@jkelling I added you to the developers team so your CI can process your PR.
please fix the tabs in your PR (see travis test) to pass the style checks :)

src/tools/bin/tbg
```diff
@@ -250,6 +250,7 @@ help()
     echo "                                 spaces within the right side of assign are not allowed"
     echo "                                 e.g. -o \"VARNAME1=10 VARNAME2=5\""
     echo "                                 Overwriting is done after cfg file was executed"
+    echo "-f | --force                   - Start new job even if TBG_dstPath exists."
```
ax3l: I would replace `Start new job` with `Overwrite` since this is independent of the `-s` option which does submission/job starts

src/tools/bin/tbg
```diff
@@ -423,8 +427,12 @@ fi
 start_dir=`dirname $0`
 
 if [ -d "$job_outDir" ] ; then
-    echo "job name already in use, can't create new folder"
-    exit 1
+	if [ "$force" = "true" ] ; then
+		echo "Warning: using existing folder for job on user-request ( -f )"
+	else
+		echo "job name already in use, can't create new folder"
```
ax3l: the old error message was not well put, maybe write `destinationPath already in use, ...`

src/tools/bin/tbg
```diff
@@ -423,8 +427,12 @@ fi
 start_dir=`dirname $0`
 
 if [ -d "$job_outDir" ] ; then
-    echo "job name already in use, can't create new folder"
-    exit 1
+	if [ "$force" = "true" ] ; then
+		echo "Warning: using existing folder for job on user-request ( -f )"
```
ax3l: maybe remove *for job*

ax3l: maybe no spaces around the `-f`, also no `-` in *user request*:
`... user request (-f)`

src/tools/bin/tbg
ax3l: in line 239 the `[-f` needs to be added before the `[-h]`

src/tools/bin/tbg
jkelling: I removed the spaces, but left the '-'. In exchange I swapped the parenthesis with square brackets to achieve some consistency with the help message.

jkelling:
Hallo Axel,

kann man die Check nochmal anstoßen? Es sagt immernoch, dass ich nicht pullen 
darf. Ist das Überhaupti wichtig? Pullt nicht am Ende ohnehin der der den 
Mergerequest am Ende annimmt?

Viele Grüße

Jeffrey

psychocoderHPC:
@jkelling Please rebase against the latest dev and than force push to your banch, this will trigger the CI again.
We merge PR only if all CI tests passed even if no core code is touched.

src/tools/bin/tbg
```diff
@@ -423,8 +427,12 @@ fi
 start_dir=`dirname $0`
 
 if [ -d "$job_outDir" ] ; then
-    echo "job name already in use, can't create new folder"
-    exit 1
+    if [ "$force" = "true" ] ; then
```
ax3l: note: this is valid, but I usually prefer `==` to be more explicit

http://tldp.org/LDP/abs/html/comparison-ops.html

jkelling: Fixed. Amended.

Issue #2212, opened 2017-08-21
ISAAC Broken in dev


ax3l:
Besides CPU compatibility of ISAAC in #2180 , the CUDA backend is broken in dev.

@theZiz @psychocoderHPC ISAAC fails on errors such as
```
picongpu/plugins/IsaacPlugin.hpp(399): here

isaac.hpp(525) error:
  argument of type "isaac::isaac_float4 **" is incompatible with parameter of type "void **"

isaac.hpp(824) [...]
isaac.hpp(830) [...]
isaac.hpp(1462) [...]
isaac.hpp(1463): error:
  a value of type "cudaError_t" cannot be used to initialize an entity of type "const cuplaError_t"
```

theZiz:
I guess, the cuda back end should just not be used for the Cupla version of PIConGPU.

ax3l:
Sure, we can just channel it through alpaka.

Can you make the changes to the plugin so it works with both current enabled backends?

psychocoderHPC:
There is a explicit
void** cast missing. The reason is that cupla is a c++ interface and a pointer to void is not implicit casted. Native Cuda is C where everything can be casted to a void pointer.

ax3l:
@theZiz please close this issue when your tests with upcoming ISAAC 1.4.0 are finished

theZiz:
Okay, tested with recent PIConGPU: It's working. :+1:

Issue #2180, opened 2017-08-08
cupla/alpaka compatibility list


psychocoderHPC:
This issue contains a list of features and plugins that are not ported to lockstep and are not yet supported by the cpu backend of cupla.

# Plugins [not supported on CPU]
- [ ] PositionsParticles
- [ ] ChargeConservation
- [ ] ParticleMerging
- [ ] Radiation
- [ ] PhaseSpace
- [x] Count Particles (already working) #2192
- [ ] count per supercell (macro particles)
- [ ] InSituVolumeRenderer
- [ ] LiveViewPlugin
- [x] PngPlugin #2225
- [ ] IntensityPlugin
- [ ] ADIOSWriter #2236
- [ ] IsaacPlugin

# PMacc Features [ not supported on CPU]
- [ ] MallocMCBuffer
- [ ] cuSTL Buffer
- [ ] RNGProvider (not tested)
- [ ] muli stream support for CPU (maybe a BUG in the event system)

# Core Features [ not supported on CPU]
- [ ] Bremsstrahlung
- [ ] SynchrotronPhotons
- [ ] Ionization  #2234

# PMacc Workarounds
- [ ] shared Memory detection `MemoryInfo.hpp`
- [ ] memory allocation in `ParticlesBox` (on CPU plain `new` instead of `mallocMC::malloc` is used)
- [ ] `types.hpp` remove cupla renaming if CUDA backend is used

ax3l:
@theZiz ISAAC fails on errors such as
```
picongpu/plugins/IsaacPlugin.hpp(399): here

isaac.hpp(1463): error:
  a value of type "cudaError_t" cannot be used to initialize an entity of type "const cuplaError_t"
```

psychocoderHPC:
@ComputationalRadiationPhysics/picongpu-maintainers 

Can

PR #2268, opened 2017-09-19
Reenabled support for ISAAC using the alpaka version

theZiz:
I reenabled ISAAC in PIConGPU using the Alpaka branch of ISAAC. For this I needed to extend ISAAC a bit, so keep sure to use the most recent ISAAC dev version for testing.
This is related to https://github.com/ComputationalRadiationPhysics/picongpu/issues/2180 and should also fix https://github.com/ComputationalRadiationPhysics/picongpu/issues/2212


include/picongpu/plugins/IsaacPlugin.hpp
```diff
@@ -373,30 +378,16 @@ class IsaacPlugin : public ILightweightPlugin
...
-                SourceList,
-                DataSpace< simDim >,
-                textureDim, float3_X,
-#if( ISAAC_STEREO == 0 )
-                    isaac::DefaultController,
-                    isaac::DefaultCompositor
-#else
-                    isaac::StereoController,
-#   if( ISAAC_STEREO == 1 )
-                        isaac::StereoCompositorSideBySide<isaac::StereoController>
```
psychocoderHPC: Do you removed the stereo rendering with this PR?

include/picongpu/plugins/IsaacPlugin.hpp
```diff
@@ -27,6 +27,7 @@
 #include <pmacc/dataManagement/DataConnector.hpp>
 #include <pmacc/static_assert.hpp>
 
+#define ISAAC_IDX_TYPE cupla::IdxType
```
psychocoderHPC: Is this a workaround or the common way to set the index type for ISAAC?
Is it possible to extract the index type from one of the classes passed to `IsaacVisualization<>`?

include/picongpu/plugins/IsaacPlugin.hpp
theZiz: No. The diff looks bigger than it is. I set two variables for findISAAC and removed the "only on GPUs" if.

include/picongpu/plugins/IsaacPlugin.hpp
theZiz: This is the new common way, as ISAAC did not have this feature before. The problem is, that some internal types are using the index type, too. So without this solution (yes, macros are ugly) I would break backwards compatibility. You can't have two different visualization objects with different index types atm but before you were not able to change the index type at all.

PR #2278, opened 2017-09-22
Removed libm dependency for Intel compiler...

theZiz:
... as the intel compiler has it's own, faster math functions
automatically included.


theZiz:
I checked whether this worked with `ldd -v ./picongpu` ;)

psychocoderHPC:
@theZiz And the result is ....

theZiz:
Obviously yes?

include/picongpu/CMakeLists.txt
```diff
@@ -142,8 +142,9 @@ set(LIBS ${LIBS} ZLIB::ZLIB)
 # Find math from stdlib
 ################################################################################
 
-if(NOT WIN32)
+if((NOT WIN32) AND NOT ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel") )
```
PrometheusPi: QUESTION:
Why did you add perentesis arounf `NOT WIN32` but not around `(NOT ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel"))` ?

theZiz: the cmake documentation was not very clear, whether `NOT` is stronger than `AND`. However in the second case it doesn't matter. ;)

psychocoderHPC: Please add `()` also around the second term.

include/picongpu/CMakeLists.txt
```diff
@@ -142,8 +142,9 @@ set(LIBS ${LIBS} ZLIB::ZLIB)
 # Find math from stdlib
 ################################################################################
 
-if(NOT WIN32)
+if((NOT WIN32) AND (NOT ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel")))
```
ax3l: this is unnecessary. please write
```cmake
if(NOT WIN32 AND NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel")
    # ...
endif()
```
Reference:
- https://cmake.org/cmake/help/v3.0/command/if.html


```
[...] there is a traditional order of precedence. Parenthetical expressions
are evaluated first followed by unary tests such as EXISTS, COMMAND, and
DEFINED. Then any binary tests such as EQUAL, LESS, GREATER, STRLESS,
STRGREATER, STREQUAL, and MATCHES will be evaluated. Then boolean NOT operators
and finally boolean AND and then OR operators will be evaluated.
```

We could also decide to make it easier readable with an empty path:
```cmake
if(WIN32 OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel")
    # automatically added on windows
    # should not be added for Intel compiler as it has an optimized lib included
else()
     set(LIBS ${LIBS} m)
endif()
```

it's also actually likely that `WIN32` is not fully true and is actually MSVC... but we would need to do a MinGW/gc test on windows for that to confirm, so that is OT.

theZiz: The if-else work around makes the condition shorter, but the code __not__ easier readable!

ax3l: I don't mind which one you choose, but please remove the superfluous brackets which are already defined well by operator precedence.

theZiz:
Done

Issue #2283, opened 2017-10-04
verification synchrotron solver


psychocoderHPC:
We have no examples which creates useful data to verify the synchrotron solver. 

# Todo

- create a example preset which can be used to verify the solver
- add tools if needed to create the needed plots for verification

CC-ing: @PrometheusPi

PR #2284, opened 2017-10-04
activate synchrotron for CPU back-end

psychocoderHPC:
allow using synchrotron solver with the CPU back-end

# Tests

- [x] CPU runtime test (bunch example case 4)
- [x] GPU runtime test (bunch example case 4)

I can't verify the solver because we have no useful preset to verify the correctness #2283


psychocoderHPC:
The lockstep programming for the kernel was already done in #2234 for the ionization.

PR #2288, opened 2017-10-05
tbg: Fix existing-folder error message to stderr

jkelling:
EDITED by @PrometheusPi:
Stream error message to stderr and fix minor typos.

### DIFF
--- a/src/tools/bin/tbg
+++ b/src/tools/bin/tbg
@@ -283,7 +283,7 @@ fi
 
 egetoptTool=`which $pathToegetopt/egetopt`
 if [ $? -ne 0 ] ; then
-    echo "Can't find program egetopt" >&2
+    echo "Cannot find program egetopt" >&2
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


src/tools/bin/tbg
```diff
@@ -430,7 +430,7 @@ if [ -d "$job_outDir" ] ; then
     if [ "$force" == "true" ] ; then
         echo "Warning: using existing folder on user-request [-f]"
     else
-        echo "destination path already in use, can't create new folder"
+        echo "Destination path already in use, can't create new folder" >&2
```
PrometheusPi: OPTIONAL: 
If you want a capital letter to start a sentence please ad a dot in the end too. 
Good point with streaming this to stderr.

jkelling:
Regarding capitalization: I wanted to be consistent withe other messages in tbg wich start with capital letter and do not end with a period. They would all need to be changed.

src/tools/bin/tbg
PrometheusPi: regarding @jkelling reply below: I can accept that - I will merge as soon as the compile test is possitive

ax3l: please write `cannot` while we are on it.

PR #2289, opened 2017-10-05
Python: Parameter Ranges for Param Files (LWFA)

codingS3b:
Under lib/python/picongpu we added a python file that defines the parameters that can be adjusted by the user in the notebook. 
But in order to do so we have to include some Parameter classes as external dependencies. Not sure how/if this Parameter code should also be part of picongpu.


share/picongpu/examples/LaserWakefield/lib/python/picongpu/ui_params.py
```diff
@@ -0,0 +1,20 @@
+from eupraxia_picongpu.parameters import *
```
PrometheusPi: REQUIRED:
Please add a license header to this file.

share/picongpu/examples/LaserWakefield/lib/python/picongpu/ui_params.py
```diff
@@ -0,0 +1,20 @@
...
+
+# if more params are added, make sure to add their names and sensible default
+# values to uiParamDefines.param within picongpu simulation_defines and adapt
+# the *.param files that define those values
```
PrometheusPi: REQUIRED:
grammar define**s**

share/picongpu/examples/LaserWakefield/lib/python/picongpu/ui_params.py
```diff
@@ -0,0 +1,20 @@
...
+    LogScaledParameter(name="Density", ptype="compile", unit="1/m^3",
+                       default=25, slider_min=20, slider_max=26, slider_step=1, base=10),
+
+    UiParameter(name="Laser_A0", ptype="compile",
+                unit="", default=1.5, slider_min=0.1, slider_max=50.01, slider_step=0.1),
+
+    LinearScaledParameter(name="Laser_Wavelength", ptype="compile", unit="m", default=0.8,
+                          slider_min=0.4, slider_max=1400.1, slider_step=0.1, scale_factor=1.e-6),
+
+    LinearScaledParameter(name="LASER_PULSE_LENGTH", ptype="compile", unit="s", default=5,
```
PrometheusPi: QUESTION:
How do you ensure that here a FWHM intensity value is set (in contrast tu the name which is a sigma electric field amplitude value in PIConGPU)

share/picongpu/examples/LaserWakefield/lib/python/picongpu/ui_params.py
jkelling: The reference to uiParamDefines.param is outdated. The default defines are to be placed at the beginnings of the respective *.param file.

codingS3b:
Thanks for your hints, I am in the process of fixing them. Those changes are only what we needed to change inside the picongpu code. The functions and classes for rendering the sliders, submitting and displaying results are contained in a different repo (the same as for the parameter classes) that probably has no place within picongpu. Atm it can be found on the HZDR gitlab server.

jkelling:
@PrometheusPi 
> there are still no interface yet, only the class describing the sliders(?)

We should discuss here to what extent the interface is to be integrated into PIConGPU upstream. There are two obvious options, but the may be details to be discussed.

1. Integrate the whole jupyter interface code. This would draw in some stuff which is quite specific for the widget-based jupyter notebook interface, which may not be desirable.
The logic to start PIC jobs is quite system dependent even though it uses tbg, so there would need to be some plugin architecture / configs to take care of this.

2. Cherry-pick the minimal set of components for integration into PIConGPU. For this pull request this would the set of classes defining the parameters (the external import in ui_param.py). These classes do not depend on the interface but they are still quite specific due to the interface they provide for interaction with the interface.

share/picongpu/examples/LaserWakefield/lib/python/picongpu/ui_params.py
codingS3b: since "define" refers to many files, it is correct. The comment still is outdated as mentioned by @jkelling and will be updated

ax3l:
@jkelling as discussed previously, it's option 2. we go for. We can integrate general tools/APIs and general post-processing widgets would be ok as well, but especially not the special-setup input widgets and notebooks.

share/picongpu/examples/LaserWakefield/include/picongpu/simulation_defines/param/density.param
```diff
@@ -24,6 +24,10 @@
 /* preprocessor struct generator */
 #include <pmacc/preprocessor/struct.hpp>
 
+// given in 1/m^3
+#ifndef UI_MACRO_DENSITY
```
ax3l: @codingS3b @jkelling contrary to what we discussed, let us please not use a `UI_MACRO_` prefix but a `PARAM_` prefix for these defines.
Please call the defines *exactly* like their variable counterparts, just with a `PARAM_` prefix (and also all caps).

The reason is, that also other defines use the same prefix already and it can be controlled / forwarded via a CMake switch in `-DPARAM_OVERWRITES:LIST`. So this would just be consistent.

Sorry for first passing the wrong message on that.

share/picongpu/examples/LaserWakefield/include/picongpu/simulation_defines/param/laser.param
```diff
@@ -45,7 +59,7 @@ namespace picongpu
         namespace SI
         {
             /** unit: meter */
-            constexpr float_64 WAVE_LENGTH_SI = 0.8e-6;
+            constexpr float_64 WAVE_LENGTH_SI = UI_MACRO_LASER_WAVELENGTH;
```
ax3l: Can you please add the macros to the defined of the other lasers below as well? This would allow us to switch the laser profile later on in `components.param` as well.

share/picongpu/examples/LaserWakefield/lib/python/picongpu/ui_params.py
ax3l: we should add a list of self-contained classes here, so the GUI repo will depend on PIConGPU but not vice versa.

share/picongpu/examples/LaserWakefield/include/picongpu/simulation_defines/param/laser.param
ax3l: > you adjusted all lasers in laser.param but only Gaussian is used. So please revert your changes for the other lasers or add a switch between the laser profile used.

sorry for that first review comment. you did it right in the first place, please keep it for all lasers.

share/picongpu/examples/LaserWakefield/lib/python/picongpu/ui_params.py
ax3l: location: `lib/python/picongpu/input/`

share/picongpu/examples/LaserWakefield/include/picongpu/simulation_defines/param/laser.param
```diff
@@ -21,6 +21,20 @@
 
 #pragma once
 
+// unitless
+#ifndef UI_MACRO_LASER_A0
+#define UI_MACRO_LASER_A0 1.5
```
ax3l: minor: we nowadays try to align `#if`s [like this](https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md#7-preprocessor):
```C++
#ifndef UI_MACRO_LASER_A0
#   define UI_MACRO_LASER_A0 1.5
#endif
```

share/picongpu/examples/LaserWakefield/lib/python/picongpu/ui_params.py
```diff
@@ -0,0 +1,36 @@
...
+from eupraxia_picongpu.parameters import *
+
+# If more params are added, make sure to adapt the corresponding
+# *.param files that define those values within picongpu.
+# In order to visualize parameters they need to be at least of
+# class UiParameter or inherited
+
+PARAMETER_LIST = [
+    LogScaledParameter(
+        name="Density", ptype="compile", unit="1/m^3",
```
ax3l: one could think about adding a `name` and a `label` here

share/picongpu/examples/LaserWakefield/lib/python/picongpu/ui_params.py
```diff
@@ -0,0 +1,36 @@
+"""
```
ax3l: can we call this file just `params.py` or with the also included `.cfg` options `input.py`?

Because we could also set all other options exposed in its `share/picongpu/examples/LaserWakefield/cmakeFlags` file (and the .cfg file).

share/picongpu/examples/LaserWakefield/lib/python/picongpu/ui_params.py
codingS3b: from which directory on would the include then be needed?
Like this
```python
from picongpu.input.parameters import *
```
or some directory above/below?

lib/python/picongpu/input/parameters.py
```diff
@@ -0,0 +1,103 @@
...
+Copyright 2017 PIConGPU contributors
+Authors: Sebastian Starke, Jeffrey Kelling
+License: GPLv3+
+"""
+
+from numpy import log10
+
+
+class Parameter(object):
+    def __init__(self, name, ptype, unit, default, value=None):
```
ax3l: please provide doc strings for the class and methods.

See for example `lib/python/picongpu/plugins/energy_histogram.py`

share/picongpu/examples/LaserWakefield/include/picongpu/simulation_defines/param/density.param
```diff
@@ -37,7 +36,12 @@ namespace SI
      *
      * unit: ELEMENTS/m^3
      */
-    constexpr float_64 BASE_DENSITY_SI = 1.e25;
+
+// given in 1/m^3
```
ax3l: this can be removed, including the empty line above

share/picongpu/examples/LaserWakefield/lib/python/picongpu/ui_params.py
ax3l: @codingS3b please still rename the file

share/picongpu/examples/LaserWakefield/lib/python/picongpu/ui_params.py
```diff
@@ -0,0 +1,36 @@
...
+This file is part of PIConGPU.
+
+Copyright 2017 PIConGPU contributors
+Authors: Sebastian Starke, Jeffrey Kelling
+License: GPLv3+
+"""
+
+from picongpu.input.parameters import *
+
+# If more params are added, make sure to adapt the corresponding
```
ax3l: please add this to the file documentation in the first lines of this file.

share/picongpu/examples/LaserWakefield/lib/python/picongpu/ui_params.py
```diff
@@ -0,0 +1,36 @@
...
+# In order to visualize parameters they need to be at least of
+# class UiParameter or inherited
+
+PARAMETER_LIST = [
+    LogScaledParameter(
+        name="Base_Density_SI", ptype="compile", unit="1/m^3",
+        default=25, slider_min=20, slider_max=26,
+        slider_step=1, base=10),
+
+    UiParameter(
```
ax3l: `a0` is also a linearly scaled parameter

share/picongpu/examples/LaserWakefield/lib/python/picongpu/ui_params.py
```diff
@@ -0,0 +1,36 @@
+"""
+This file is part of PIConGPU.
+
+Copyright 2017 PIConGPU contributors
+Authors: Sebastian Starke, Jeffrey Kelling
+License: GPLv3+
+"""
+
+from picongpu.input.parameters import *
```
ax3l: you can just do explicitly: `from picongpu.input.parameters import LogScaledParameter, LinearScaledParameter`

lib/python/picongpu/input/parameters.py
```diff
@@ -0,0 +1,103 @@
...
+        return self.value
+
+    def get_value(self):
+        return self.value
+
+
+class LinearScaledParameter(UiParameter):
+
+    def __init__(self, name, ptype, unit, default, slider_min, slider_max,
+                 slider_step, scale_factor, value=None):
```
ax3l: if you set `scale_factor = 1.0, ` as the default, you will only need two imports in user code

ax3l:
sorry for the delay in merging, I was on two working travels during the last 7 days.

share/picongpu/examples/LaserWakefield/include/picongpu/simulation_defines/param/laser.param
```diff
@@ -21,6 +21,17 @@
 
 #pragma once
 
+#ifndef PARAM_A0
+#   define PARAM_A0 1.5
```
ax3l: ah, the default should have stayed with 8.0. will fix!

ccing @theZiz

share/picongpu/examples/LaserWakefield/lib/python/picongpu/params.py
```diff
@@ -0,0 +1,41 @@
...
+        default=25, slider_min=20, slider_max=26,
+        slider_step=1, base=10),
+
+    LinearScaledParameter(
+        name="_A0", ptype="compile", unit="",
+        default=1.5, slider_min=0.1, slider_max=50.01,
+        slider_step=0.1),
+
+    LinearScaledParameter(
+        name="Wave_Length_SI", ptype="compile", unit="m",
```
ax3l: @codingS3b can we add a "combo-box" parameter that for the central laser wavelength only allows:
- 800nm (Ti:Sa lasers)
- 1054nm (Nd:YAG lasers)

that's all we will need for now here (later we might add selected frequency-doubled ones).

PR #2296, opened 2017-10-09
PMacc: Set CPU Architecture

ax3l:
Re-implementation of #2279 in pure CMake.
Easier to maintain a [list of compilers](https://cmake.org/cmake/help/v3.9/variable/CMAKE_LANG_COMPILER_ID.html) in one place and CXX+CXXFLAGS will always match.

## Description

Set the CPU architecture via `PMACC_CPU_ARCH`. Defaults to an environment variable of the same name. Only set if the (host) compiler is known, otherwise use CXXFLAGS if you experiment with more compilers.

Only set as a convinient setter for the user instead of going directly to CXXFLAGS as the user. Helpful in commands such as `pic-configure` to avoid building a compiler lookup list therein.

## Additional Usage Options

For a CUDA backend user, that still wants to set the CPU arch properly, this can be done with `-DPMACC_CPU_ARCH` (`pic-configure -c` or `ccmake`) or `export PMACC_CPU_ARCH`. For CPU backends, it is directly taken from the `[:architecture]`.

## Examples with `pic-build`

(which is actually `pic-configure`)

### OpenMP Backend
```bash
pic-build -b omp2b:native
# [...]
cmake -L .build/ | grep "CPU_ARCH"
# PMACC_CPU_ARCH:STRING=native
```
```bash
pic-build -b omp2b
# [...]
cmake -L .build/ | grep "CPU_ARCH"
# PMACC_CPU_ARCH:STRING=
```

### CUDA Backend
```bash
pic-build -b cuda:35
# [...]
cmake -L .build/ | grep "CPU_ARCH"
# PMACC_CPU_ARCH:STRING=
```

### CUDA Backend + Extra-Fine Tuning of Host Code
```bash
pic-build -b cuda:35 -c"-DPMACC_CPU_ARCH=native"
# [...]
cmake -L .build/ | grep "CPU_ARCH"
# PMACC_CPU_ARCH:STRING=native
```

### DIFF
--- a/include/pmacc/PMaccConfig.cmake
+++ b/include/pmacc/PMaccConfig.cmake
@@ -147,6 +147,56 @@ endif()
 set(PMacc_LIBRARIES ${PMacc_LIBRARIES} ${cupla_LIBRARIES})
 
 
+###############################################################################
+# CPU Architecture: available instruction sets for e.g. SIMD extensions
+#
+# Conveniently set the architecture for the CPU compiler via this option.
+# For unsupported compilers, ignore this option and set CXXFLAGS.
+###############################################################################
+
+set(PMACC_CPU_ARCH $ENV{PMACC_CPU_ARCH} CACHE STRING
+    "compiler dependent CPU architecture string"
+)
+
+# list of known compiler flags to set the CPU architecture
+# GNU
+if(CMAKE_COMPILER_IS_GNUCXX)
+    if("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "ppc64le")
+        set(PMACC_CPU_ARCH_TEMPLATE "-mcpu={} -mtune={}")
+    else()
+        set(PMACC_CPU_ARCH_TEMPLATE "-march={} -mtune={}")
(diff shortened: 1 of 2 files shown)


include/pmacc/PMaccConfig.cmake
```diff
@@ -147,6 +147,57 @@ endif()
 set(PMacc_LIBRARIES ${PMacc_LIBRARIES} ${cupla_LIBRARIES})
 
 
+###############################################################################
+# CPU Architecture: available instruction sets for e.g. SIMD extensions
+#
+# Conviniently set the architecture for the CPU compiler via this option.
```
ax3l: typo: Conveniently

include/pmacc/PMaccConfig.cmake
```diff
@@ -147,6 +147,57 @@ endif()
...
+# For unsupported compilers, setting this to none and setting CXXFLAGS is
```
ax3l: "setting this to none" -> "ignoring this option"

PR #2300, opened 2017-10-10
Const POD Default Constructor

ax3l:
Fixes clang(-tidy) errors of the kind [clang-diagnostic-error]:
"default initialization of an object of const type '...' without a user-provided default constructor":

First approach before discussion:
```C++
const typename Difference::template GetDifference<0> Dx;
// to
const typename Difference::template GetDifference<0> Dx{};
```

Final approach: add explicit constructors to POD classes that are used as `const`:
```C++
    struct GetDifference
    {
        static constexpr uint32_t direction = T_direction;

        HDINLINE GetDifference()
        {
        }
        // ...
    };
```

### DIFF
--- a/include/picongpu/fields/numericalCellTypes/EMFCenteredCell.hpp
+++ b/include/picongpu/fields/numericalCellTypes/EMFCenteredCell.hpp
@@ -17,13 +17,12 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
-
-
 #pragma once
 
 #include "picongpu/simulation_defines.hpp"
 #include "picongpu/fields/Fields.def"
 #include <pmacc/math/Vector.hpp>
+
 
 namespace picongpu
 {
@@ -57,6 +56,10 @@ namespace traits
             typedef ReturnType type;
         };
 
+        HDINLINE FieldPosition()
+        {
+        }
+
         HDINLINE ReturnType operator()() const
         {
             const auto center = PosType::create( 0.5 );
@@ -89,6 +92,10 @@ namespace traits
             typedef VectorVector2D3V type;
         };
 
+        HDINLINE FieldPosition()
+        {
+        }
+
         HDINLINE VectorVector2D3V operator()() const
         {
             const float2_X posJ_x( 0.5, 0.0 );
@@ -111,6 +118,10 @@ namespace traits
         struct result<F()> {
             typedef VectorVector3D3V type;
         };
+
+        HDINLINE FieldPosition()
+        {
+        }
 
         HDINLINE VectorVector3D3V operator()() const
         {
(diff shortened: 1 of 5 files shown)


psychocoderHPC:
Based on [this](https://stackoverflow.com/a/7411708) stakoverflow post we should add a explicit default constructor to `GetDifference`, the reult of `traits::FieldPosition`and `ShiftMeIfYouCan`.
The reason is that a type without a constructor is a POD data type and can be uninitialized. An uninitialized const variable make no sense. With the current PR we defining from the outside that we a fine with that uninitialized type but this requires internal knowledge about the implementation.

We should add a default constructor e.g. `GetDifference() = default;`

ax3l:
thanks!

I thinks it's actually exactly the opposite implementation wise, instead of explicitly not adding a constructor in the current class with [`= default;`](https://stackoverflow.com/questions/28337951/clang-compile-error-with-default-initialization/28338123#28338123) we just add one with `GetDifference(){};` instead.

include/picongpu/fields/numericalCellTypes/EMFCenteredCell.hpp
```diff
@@ -112,6 +119,10 @@ namespace traits
             typedef VectorVector3D3V type;
         };
 
+        HDINLINE FieldPosition()
```
psychocoderHPC: In c++11 it should be `FieldPosition = default;` Please to not add a prefix qualifier else we will get nvcc or clang warnings.

psychocoderHPC: Ohh missread your comment. `= default` is equal to the old c++98 way.
Whe can check it in a mini ap with std::is_pod.

ax3l:
I tried but = default did not work :(
See link above

include/picongpu/fields/numericalCellTypes/EMFCenteredCell.hpp
psychocoderHPC: offline discussed: `= default` is not equal to a hand written  empty constructor :-(

ax3l:
Observed today: this is a clang 3.8 weirdness and gone in 3.9+:

Has problems to compile with the error message above:
```C++
#include <string>
#include <iostream>

constexpr float f = 4.5;

struct A
{
    static constexpr float value = f;

    float
    operator()() const
    {
        return f;
    }
};

int main()
{
    A const a;
    std::cout << a() << std::endl;

    return 0;
}
```

We should just require clang 3.9+

sbastrakov:
Wow, that look horrifying. Even a [trivial example](http://rextester.com/TWXT53317) with const object + automatic default constructor does not work in clang 3.8.

ax3l:
Yep, also tried it online: https://godbolt.org/g/sfhPXz

sbastrakov:
Omg, it was actually a standard-conforming behaviour, which then was marked as a defect: http://open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#253

ax3l:
omg -.-

PR #2301, opened 2017-10-10
Fix Dead Code

ax3l:
Fix dead code and unused variables in scenarios such as:
- no laser or
- CPU only

found with clang-tidy, fixed manually.


include/picongpu/fields/laserProfiles/laserGaussianBeam.hpp
```diff
@@ -187,7 +187,7 @@ namespace picongpu
                               / SPEED_OF_LIGHT / SPEED_OF_LIGHT / ( float_X(2.0) * PULSE_LENGTH ) / ( float_X(2.0) * PULSE_LENGTH ) );
                 }
                 elong.z() *= etrans / etrans_norm;
-                phase -= float_X( PI / 2.0 );
+                // phase -= float_X( PI / 2.0 );
```
ax3l: I kept this here as a reminder that the + pi/2 above is quite voluntarily

psychocoderHPC: Please add a short description else the next who see it will remove it because he think it is dead code.

PR #2306, opened 2017-10-11
PizDaint (CSCS): Update Profiles

ax3l:
Update to the latest modules and add libSplash for HDF5 support.

I drafted a very specific, quickly outdated, hands on install description in
  https://gist.github.com/ax3l/68cb4caa597df3def9b01640959ea56b
which just summarizes what is already [in the manual](https://picongpu.readthedocs.io/en/latest/install/profile.html#piz-daint-cscs).

ccing @HighIander @BeyondEspresso

### DIFF
--- a/etc/picongpu/pizdaint-cscs/picongpu.profile.example
+++ b/etc/picongpu/pizdaint-cscs/picongpu.profile.example
@@ -1,23 +1,39 @@
+# User Information ############################################################
+#   - edit those lines!
+#   - automatically add your name and contact to output file meta data
+#   - send me a mail on batch system jobs: BEGIN, END, FAIL, REQUEUE, ALL,
+#     TIME_LIMIT, TIME_LIMIT_90, TIME_LIMIT_80 and/or TIME_LIMIT_50
+export MY_MAILNOTIFY="ALL"
+export MY_MAIL="someone@example.com"
+export MY_NAME="$(whoami) <$MY_MAIL>"
+
+# Name and Path of this Script ################################################
+#   - do NOT change this line!
+export PIC_PROFILE=$(cd $(dirname $BASH_SOURCE) && pwd)"/"$(basename $BASH_SOURCE)
+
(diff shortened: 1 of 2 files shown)


etc/picongpu/pizdaint-cscs/picongpu.profile.example
```diff
@@ -70,8 +81,8 @@ export TBG_TPLFILE="etc/picongpu/pizdaint-cscs/normal.tpl"
 # helper tools ################################################################
 
 # allocate an interactive shell for one hour
-#   getInteractive 2  # allocates to interactive nodes (default: 1)
-getInteractive() {
+#   getNode 2  # allocates to interactive nodes (default: 1)
+getNode() {
```
ax3l: @psychocoderHPC still works for me.

as usual:
```bash
getNode
# wait

# srun commandOnNode
srun hostname
```

etc/picongpu/pizdaint-cscs/picongpu.profile.example
```diff
@@ -26,12 +42,14 @@ module load cray-hdf5-parallel/1.10.0
 #
 # needs to be compiled by the user
 export BOOST_ROOT=$SCRATCH/lib/boost-1.62.0
-export PNGWRITER_ROOT=$SCRATCH/lib/pngwriter
+export PNGWRITER_ROOT=$SCRATCH/lib/pngwriter-0.6.0
```
PrometheusPi: QUESTION:
If this dependency is compiled by the user anyway why bother to add a version number?

ax3l: just a best practice one can do, so one knows at a glance what version is residing in there

etc/picongpu/pizdaint-cscs/picongpu.profile.example
psychocoderHPC: yes it is running I not realized that I am using an old profile.

PR #2308, opened 2017-10-12
Profile: Draco (MPCDF)

ax3l:
Adds a `picongpu.profile.example` for the Draco cluster at MPCDF.

### To Do

- [x] add in manual (`.rst`)
- [x] add `.tpl` for submission via `tbg` @psychocoderHPC

### DIFF
--- a/etc/picongpu/draco-mpcdf/general.tpl
+++ b/etc/picongpu/draco-mpcdf/general.tpl
@@ -1,0 +1,91 @@
+#!/usr/bin/env bash
+# Copyright 2013-2017 Axel Huebl, Richard Pausch, Rene Widera
+#
+# This file is part of PIConGPU.
+#
+# PIConGPU is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# PIConGPU is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PIConGPU.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
+
+# PIConGPU batch script for taurus' SLURM batch system
+
+#SBATCH --partition=!TBG_queue
+#SBATCH --time=!TBG_wallTime
+# Sets batch job's name
+#SBATCH --job-name=!TBG_jobName
+#SBATCH --nodes=!TBG_nodes
+#SBATCH --ntasks=!TBG_tasks
+#SBATCH --mincpus=!TBG_mpiTasksPerNode
+#SBATCH --cpus-per-task=!TBG_coresPerGPU
+#SBATCH --exclusive
+# send me mails on BEGIN, END, FAIL, REQUEUE, ALL,
+# TIME_LIMIT, TIME_LIMIT_90, TIME_LIMIT_80 and/or TIME_LIMIT_50
+#SBATCH --mail-type=!TBG_mailSettings
+#SBATCH --mail-user=!TBG_mailAddress
+#SBATCH --workdir=!TBG_dstPath
+
+#SBATCH -o stdout
+#SBATCH -e stderr
+
+
+## calculations will be performed by tbg ##
+.TBG_queue="general"
+
+# settings that can be controlled by environment variables before submit
+.TBG_mailSettings=${MY_MAILNOTIFY:-"ALL"}
+.TBG_mailAddress=${MY_MAIL:-"someone@example.com"}
+.TBG_author=${MY_NAME:+--author \"${MY_NAME}\"}
+.TBG_profile=${PIC_PROFILE:-"~/picongpu.profile"}
+
+# 1 gpus per node
+.TBG_gpusPerNode=2
+
+# number of cores to block per GPU - we got 6 cpus per gpu
+#   and we will be accounted 6 CPUs per GPU anyway
+.TBG_coresPerGPU=32
+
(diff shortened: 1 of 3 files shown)


etc/picongpu/draco-mpcdf/picongpu.profile.example
```diff
@@ -0,0 +1,63 @@
...
+module load gcc/6.3
+module load cmake/3.7
+module load boost/gcc/1.64
+module load impi/2017.3
+module load hdf5-mpi/gcc/1.8.18
+
+# Other Software ##############################################################
+#
+# needs to be compiled by the user
+export PNG_ROOT=$HOME/lib/libpng-1.6.32
```
ax3l: is this the correct hint? system zlib and system libpng might be taken currently

etc/picongpu/draco-mpcdf/picongpu.profile.example
```diff
@@ -0,0 +1,63 @@
...
+export PNGWRITER_ROOT=$HOME/lib/pngwriter-0.6.0
+export SPLASH_ROOT=$HOME/lib/splash-1.6.0
+
+export LD_LIBRARY_PATH=$PNG_ROOT/lib:$LD_LIBRARY_PATH
+export LD_LIBRARY_PATH=$PNGWRITER_ROOT/lib:$LD_LIBRARY_PATH
+export LD_LIBRARY_PATH=$SPLASH_ROOT/lib:$LD_LIBRARY_PATH
+
+# export MPI_ROOT=$MPICH_DIR
```
ax3l: removable?

etc/picongpu/draco-mpcdf/picongpu.profile.example
```diff
@@ -0,0 +1,63 @@
...
+#   - edit those lines!
+#   - examples: "nano", "vim", "emacs -nw", "vi" or without terminal: "gedit"
+#export EDITOR="nano"
+
+# General Modules #############################################################
+#
+module purge
+
+module load git/2.14
+module load gcc/6.3
```
ax3l: to do: add
```cmake
export CXX=$(which g++)
export CC=$(which gcc)
```

etc/picongpu/draco-mpcdf/general.tpl
```diff
@@ -0,0 +1,91 @@
...
+.TBG_nodes="$((( TBG_tasks + TBG_gpusPerNode -1 ) / TBG_gpusPerNode))"
+
+## end calculations ##
+
+echo 'Running program...'
+
+cd !TBG_dstPath
+
+export MODULES_NO_OUTPUT=1
+source !TBG_profile
```
ax3l: does not need to be loaded on this cluster?

psychocoderHPC:
export CXX and CC must be defined to find the correct c++ and c compiler

etc/picongpu/draco-mpcdf/picongpu.profile.example
```diff
@@ -0,0 +1,68 @@
...
+module load hdf5-mpi/gcc/1.8.18
+
+# Other Software ##############################################################
+#
+# needs to be compiled by the user
+export PNG_ROOT=$HOME/lib/libpng-1.6.32
+export PNGWRITER_ROOT=$HOME/lib/pngwriter-0.6.0
+export SPLASH_ROOT=$HOME/lib/splash-1.6.0
+
+export LD_LIBRARY_PATH=$PNG_ROOT/lib:$LD_LIBRARY_PATH
```
ax3l: can be removed if zlib and libpng are taken from system

etc/picongpu/draco-mpcdf/picongpu.profile.example
```diff
@@ -0,0 +1,68 @@
...
+
+export HDF5_ROOT=$HDF5_HOME
+
+export CXX=$(which g++)
+export CC=$(which gcc)
+
+# PIConGPU Helper Variables ###################################################
+#
+export PICSRC=$HOME/src/picongpu
+export PIC_BACKEND="omp2b"
```
ax3l: add architecture again? `:haswell` did not work in our first tests, but this could have been due to an too old gcc in the beginning.

ax3l:
this setup worked for our colleague at IPP.
I have some inline comments that I would improve but since I can not test it personally we can also keep it as it is currently.

psychocoderHPC:
We need to load `numctl`, I will check how the module is named on the cluster

ax3l:
since @psychocoderHPC and I tested on the cluster we will make the final merge when it is ready, but a general review from a third person such as you @PrometheusPi is very welcome here! :)

etc/picongpu/draco-mpcdf/picongpu.profile.example
psychocoderHPC: Yes please add it.

etc/picongpu/draco-mpcdf/picongpu.profile.example
psychocoderHPC: This can be removed. I got feedback from Katharina that it is not needed.

psychocoderHPC:
numactl is fixed installed on all compute nodes.

psychocoderHPC:
@ax3l I added an PR https://github.com/ax3l/picongpu/pull/7 to your PR to change the last points (remove own compiled PNG and add architecture)

etc/picongpu/draco-mpcdf/picongpu.profile.example
ax3l: did it work now (with the right version of gcc after setting `$CXX`?)

ax3l:
@psychocoderHPC just push the new commit directly on my branch please :) that's a new feature, you have push rights

ax3l:
we still need to test this please, the reason I did not add those comments yet was the lack of testing on the system

psychocoderHPC:
I know know that I can push directly but I love to create PR for a PR^^. Could you please press merge, than there is no need for me to switch the current branch :-)

PR #2145, opened 2017-07-24
Travis: Enforce PEP8

ax3l:
Add Python style checks to travis via flake8 tests.

PR #2309, opened 2017-10-16
XDMF Scripts: Fix Replacements & Offset

ax3l:
Fix the particle offset naming since the introduction of openPMD and fix missing `.format()` replacements with multi-line strings.

The latter issue was introduces with #2145, the former with release `0.2.0`.

### DIFF
--- a/src/tools/bin/pic2xdmf.py
+++ b/src/tools/bin/pic2xdmf.py
@@ -35,7 +35,7 @@ VECTOR_IDENTS = ["x", "y", "z", "w"]
 NAME_DELIM = "/"
 
 # PIC-specific names
-NAME_GLOBALCELLIDX = "globalCellIdx"
+NAME_GLOBALCELLIDX = "positionOffset"
 NAME_POSITION = "position"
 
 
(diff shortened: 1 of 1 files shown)


src/tools/bin/pic2xdmf.py
```diff
@@ -151,8 +151,8 @@ def join_references_from_components(node_list, original_nodes_map, prefix,
             extra_grid = "Grid[@Name='{}']/".format(
                 parent_node.getAttribute("Name"))
 
-        node_text = "/Xdmf/Domain/Grid/{}Attribute[@Name='{}']/" + \
-                    "DataItem[1]".format(extra_grid, orig_name)
+        node_text = ("/Xdmf/Domain/Grid/{}Attribute[@Name='{}']/"
+                     "DataItem[1]").format(extra_grid, orig_name)
```
PrometheusPi: OPTIONAL:
A plus sign `+` is optional here but would improve readability of the code to emphasize that here the second string is attached.

PrometheusPi: same below

ax3l: I would not add it since it is superfluous like additional brackets.
the way it is now is the common way to do it properly.

PrometheusPi: Okay

PR #2311, opened 2017-10-18
Fixes for python parameters 

codingS3b:
Just minor fixes of typos in documentation of parameter classes and the parameter range of Wave_Length_SI parameter of the LaserWakefield example.


ax3l:
looks good.

can we also make a parameter that has options instead of a range? e.g. wavelengths for high power lasers are quite limited: 800nm, frequency doubled and 1054 are the typical choices.

PR #926, opened 2015-06-22
remove invalid typename from .param and .unitless

BenjaminW3:
#917 has been divided for better overview.

PR #2314, opened 2017-10-20
Examples: C++11 Using for Typedef

ax3l:
replace `typedef` in examples with modern C++11 `using`. Commited as tools user.

Also added missing `typenames` where `::types` are taken from templated classes.

### DIFF
--- a/share/picongpu/examples/Bremsstrahlung/include/picongpu/simulation_defines/param/particle.param
+++ b/share/picongpu/examples/Bremsstrahlung/include/picongpu/simulation_defines/param/particle.param
@@ -24,6 +24,7 @@
 #include <pmacc/nvidia/functors/Add.hpp>
 #include <pmacc/nvidia/functors/Assign.hpp>
 
+
 namespace picongpu
 {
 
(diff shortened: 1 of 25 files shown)


psychocoderHPC:
Note: the `typename` was removed in #926 to fulfill C++98, since we are now using C++11 this change should be correct.
Never the less I will test it with clang.

ax3l:
oh wait, do we even *need* the `typename` or is it superfluous?

psychocoderHPC:
I found [this](https://stackoverflow.com/a/613132) while searching to rules to use `typename`

The usage of `typename` can help to have a little bit of concept checking see [here](http://pages.cs.wisc.edu/~driscoll/typename.html). By using `typename` it say's the compiler that the component used from a class is a `type` and not a variable.

psychocoderHPC:
I tested this PR with clang, clang is accepting the typename, like @BenjaminW3 wrote  in #926 . 
IMO we should **keep** the `typename`.

ax3l:
ok, thanks!
can we merge then? :)

PR #2315, opened 2017-10-20
Default Inputs: C++11 Using for Typedef

ax3l:
replace `typedef` in examples with modern C++11 `using`. Commited as tools user.

Also added missing `typenames` where `::types` are taken from templated classes.


include/picongpu/simulation_defines/unitless/precision.unitless
```diff
@@ -21,53 +21,54 @@
...
     {
-        typedef precisionType float_X;
-        /*32 Bit defines*/
-        typedef ::pmacc::math::Vector<float_X, 1> float1_X;
-        typedef ::pmacc::math::Vector<float_X, 2> float2_X;
-        typedef ::pmacc::math::Vector<float_X, 3> float3_X;
-        typedef ::pmacc::math::Vector<float_X, simDim> floatD_X;
+        using float_X = precisionType;
+        /* 32 Bit defines */
+        using float1_X = ::pmacc::math::Vector< float_X, 1 >;
```
ax3l: @psychocoderHPC I think we should move everything that is not `_X` and not `D_` to a general place in PMacc so we have it broader available.

psychocoderHPC: I am not sure, `float_32`, `...` is never used in *PMacc* I would keep it in the user (PIConGPU) code expect there is a really strong argument to move it to `PMacc`

ax3l: it would be as useful as the other fundamental types and we could specialize more functions for pre-compile when needed.

PR #2317, opened 2017-10-20
pic-compile: make cmakeFlags optional

psychocoderHPC:
Since #2243 the cmakeFlags file for examples without user params is removed.
`pic-compile` used the wrong default value and assumed that `cmakeFlags` is always setting the correct number of test cases within an example.

- fix pic-compile
- `WeibelTransverse` test: add `DefaultExchangeMemCfg` to `memory.param`


share/picongpu/examples/WeibelTransverse/include/picongpu/simulation_defines/param/memory.param
```diff
@@ -51,6 +51,22 @@ constexpr uint32_t BYTES_EXCHANGE_Y = 12 * 512 * 1024; //12 MiB
 constexpr uint32_t BYTES_EXCHANGE_Z = 256 * 256 * 1024; //256 MiB
 constexpr uint32_t BYTES_CORNER = 2 * 256 * 1024; //2 MiB
 constexpr uint32_t BYTES_EDGES = 8 * 256 * 1024; //8 MiB
+/** bytes reserved for species exchange buffer
```
ax3l: tzzz ;)

PR #2324, opened 2017-10-25
LWFA Example: Restore a0=8.0

ax3l:
restore the default laser strength of 8.0 in the LWFA example. Follow-up to the unintentional change introduced in #2289.

ccing @codingS3b @theZiz


theZiz:
I hope this explains my new visualization differences. Testing right now!

Issue #2325, opened 2017-10-25
FieldJ is not part of the checkpointing


psychocoderHPC:
Currently `FieldJ` is not stored in the checkpoint files. This means that after a restart the `FieldJ` will be zero instead of the last computed current.
This results in different behavior if the current is somewhere in the particle in cell cycle used before it is update for the time step where we restart the simulation.

Note: As I know we are currently not using `FieldJ` somewhere in a partictle in cell cycle extention.

CC-ing: @ax3l

ax3l:
that's ok for now, since we currently have no PIC cycle extensions that need the current at a point where it was not freshly calculated.

psychocoderHPC:
note: FieldJ can be added by the user to the file `checkpoints.unitless`

ax3l:
as an alternative approach (when it should be needed in the future), it would be possible to derive the current during restart again from the particles, background fields and antennas instead of loading a full vector field.

sbastrakov:
I think this is not an issue. If it needs to be stored for some reason, could be added manually to the checkpointed fields. Or if we really need it in most simulations, recalculated after loading, as described above.

PR #2326, opened 2017-10-25
Wrong border with current background field

### LINKED ISSUE #943: BG Field: Current Added to Guard
The BG fields for at least `FieldJ` is added to the `GUARD` region (but should only apply for the `BORDER` + `CORE` region since the current is communicated afterward).

Needs also to be checked for `FieldE` and `FieldB`.

psychocoderHPC:
fix #943

If field background is used than the border regions will be wrong.
The field background current is operationg on the local domain regions
`CORE + BORDER + GUARD`. The current is than scattered to the neighboring GPUs
and will create wrong border regions.

- operate background field current only on the region `CORE + BORDER`
- fix `cellwiseOperation` - index passed to the user functor was only correct if `CORE+GUARD + BORDER` was manipulated

With this fix current smoothing is still possible. **Attention** FieldJ is not part of the checkpointing #2325

~~do not merge, there is a wrong index calculation in `cellwiseOperation.hpp`~~

CC-ing: @steindev

### DIFF
--- a/include/picongpu/fields/background/cellwiseOperation.hpp
+++ b/include/picongpu/fields/background/cellwiseOperation.hpp
@@ -60,6 +60,7 @@ namespace cellwiseOperation
          * @param opFunctor binary operator used with the old and functor value
          *                  (collective functors are not supported)
          * @param valFunctor functor to execute (collective functors are not supported)
+         * @param totalDomainOffset offset to the local domain relative to the origin of the global domain
          * @param currentStep simulation time step
          * @param mapper functor to map a block to a supercell
          */
@@ -76,7 +77,7 @@ namespace cellwiseOperation
             T_FieldBox field,
             T_OpFunctor opFunctor,
             T_ValFunctor valFunctor,
-            DataSpace< simDim > const totalCellOffset,
+            DataSpace< simDim > const totalDomainOffset,
             uint32_t const currentStep,
             T_Mapping mapper
         ) const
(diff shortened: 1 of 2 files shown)


ax3l:
looks good, we just need a RT validation and maybe a check with current smoothing on

psychocoderHPC:
@steindev Could you please post the plot from the broken background field J to this PR.

steindev:
The picture shows the hdf5 output of `fields/J` where the current density in the simulation is defined via `FieldBackgroundJ` which is set to `-1` everywhere. The simulation used 8 GPUs, with 2 per direction.

The `FieldBackgroundJ` is added in the region between two GPUs resulting in a wrong current density.

background-current_error

steindev:
Repeated test with applied fix: works!

background-current_error_fix

steindev:
Update: Works as expected, ALMOST. Now I ran a test with a straight conductor of one cell width. The conductor is (supposed to be) located at positions x=z=96 for all y. But it is not.

In the following picture, the current should be located in the top right corner of the picture. Btw. this is just a part of my simulation volume. In total it is 192x192x192.

background-current_error_fix_error

PrometheusPi:
@steindev Did @psychocoderHPC fix solves your issue?

steindev:
he current is now at the correct position. Furthermore, it fixes a bug which caused the background current to be added with every time step in the left border of the simulation volume. This caused an electric field pile up within the absorber.

background-current_error_fix_error_fix

PrometheusPi:
@steindev I am confused by the left plot. If the current is equal along y, why does it reduce towards the boarder?

steindev:
Cells 0-32 are the absorber and the left plot shows the electric field generated by the current. Therefore, a reduction is expected.

ax3l:
thank you both, great catch! :)

PR #2286, opened 2017-10-04
Docker: Simpler & CUDA 9.0

ax3l:
Simplify our docker and singularity recipes.

- remove extra user (easier for "overlay" philosophy in Singularity)
- use new [PIConGPU spack recipe](https://github.com/ComputationalRadiationPhysics/spack-repo) (integrates the `picongpu.profile` for module load)
- use new, [minimal CUDA 9.0 docker images](https://hub.docker.com/r/nvidia/cuda/)
- remove emacs to save 90MB
- only download the HEAD of spack as a tarball
- use simple tcl modules in spack (and `spack load ...`)
- use new `pic-build` mechanism

### To Do

- [x] wait for merge of some upstream package fixes & sha collisions in current `develop` of spack
- [x] alpaka (& PIConGPU) CUDA 9.0 support
- [x] need ISAAC 1.3.3 release for CUDA 9.0 support and boost 1.65.0+ support https://github.com/ComputationalRadiationPhysics/isaac/issues/72
- [x] update spack github link to new org: `spack/spack`
- [x] OpenMPI: `--allow-run-as-root` https://github.com/open-mpi/ompi/issues/4451
- [x] RT testing

PR #2327, opened 2017-10-25
Doc: $PIC_EXAMPLES

ax3l:
Add a default installed `PIC_EXAMPLES` environment variable that eases to create new input from examples without struggling with long paths.

Also clean up header of profile examples on the way to be more intuitive:
- profile path first (sensitive to some system's module loads)
- name and mail notifications
- optional: computing project on some clusters
- optional: editor
- then all things that the user does not change in normal operation:
  - modules
  - required environment vars
  - defaults in batch system of this file

Also added to [spack package](https://github.com/ComputationalRadiationPhysics/spack-repo) (https://github.com/ComputationalRadiationPhysics/spack-repo/commit/9c05bba8158a31f3f7f233be634345c4cb37fd54) and will be available to users also in simplified docker file #2286.

### DIFF
--- a/USAGE.rst
+++ b/USAGE.rst
@@ -41,7 +41,7 @@ Step-by-Step
    :emphasize-lines: 2
 
    # clone the LWFA example to $HOME/picInputs/myLWFA
-   pic-create $PICSRC/share/picongpu/examples/LaserWakefield $HOME/picInputs/myLWFA
+   pic-create $PIC_EXAMPLES/LaserWakefield $HOME/picInputs/myLWFA
 
    # switch to your input directory
    cd $HOME/picInputs/myLWFA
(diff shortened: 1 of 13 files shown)


USAGE.rst
```diff
@@ -41,7 +41,7 @@ Step-by-Step
    :emphasize-lines: 2
 
    # clone the LWFA example to $HOME/picInputs/myLWFA
-   pic-create $PICSRC/share/picongpu/examples/LaserWakefield $HOME/picInputs/myLWFA
+   pic-create $PIC_EXAMPLES/LaserWakefield $HOME/picInputs/myLWFA
```
ax3l: this is the central line of this PR

PR #2331, opened 2017-10-25
(Re)Move simulation_defines/

ax3l:
Move input directories from `simulation_defines/` directly into `include/picongpu/`.

This flattens the hierachy users see in their input sets and make overall interaction less verbose and long-ish. The additional directory is just grouping the directories `param/` and `unitless/` of which a user does only manipulate `param/` anyway.

One could also think about putting the loaders into the `param/` and `unitless/` dirs.

As a follow-up, I would also propose to rename `simulation_defines.hpp` to camelCase `simulationInput.hpp`/`paramInput.hpp` or similar. On the long run, our goal is to get rid of `unitless/` alltogether with a stronger, compile-time (constexpr) type system.

**Note for review:** due to many moved files, it is easiest to review commit-wise.

@ComputationalRadiationPhysics/picongpu-maintainers this is a proposal, please tell me what you think about it.

- [x] `pic-configure` add check that `include/picongpu/simulation_defines` does not exist (in the case folder)
- [x] `pic-create` add check that `include/picongpu/simulation_defines` does not exist (in the source and destination folder)
- [x] `pic-edit` check if still working


psychocoderHPC:
Should we extent `pic-create` with a check which checks that `include/picongpu/simulation_defines` is not existing. The last time where we changed the folder structure we are running in problems where the user tried to update an existing case and  the user wondering why the updated parameter files not effects the compiled binary.

ax3l:
@jkelling (and after your vacations CC @codingS3b) we need to check if this impacts your new tools when merged

ax3l:
@ComputationalRadiationPhysics/picls-maintainer @ComputationalRadiationPhysics/picongpu-developers we would introduce this rather breaking change to `dev` now. If you are using cutting-edge `dev` and want to continue using an input set, move all content of inside your `simulation_defines/` directory just one directory up after this change.

PR #2334, opened 2017-10-26
Python: Phase Space Reader

ax3l:
Python reader class for phase space plugin.

Use as follows:
```python
from picongpu.plugins.phase_space import PhaseSpace

phase_space = PhaseSpace('/home/axel/runs/foil_001')
e_ps, e_ps_meta = phase_space.get('e', ps='ypy', iteration=1000)
```

All returned values are in SI.

A full example for plotting phase space data with matplotlib, including unit conversion has been added to the manual.

ccing @codingS3b

### DIFF
--- a/docs/source/usage/plugins/phaseSpace.rst
+++ b/docs/source/usage/plugins/phaseSpace.rst
@@ -37,47 +37,60 @@ Option                                 Usage                                    
 Output
 ^^^^^^
 
+The 2D histograms are stored in ``.hdf5`` files in the ``simOutput/phaseSpace/`` directory.
+A file is created per species, phasespace selection and time step.
+
 Values are given as *charge density* per phase space bin.
-In order to scale to a simpler *charge of particles* per :math:`\mathrm{d}r_i` and :math:`\mathrm{d}p_i` -bin multiply by the cell volume:
+In order to scale to a simpler *charge of particles* per :math:`\mathrm{d}r_i` and :math:`\mathrm{d}p_i` -bin multiply by the cell volume ``dV``.
+
+Analysis
+^^^^^^^^
+
+The easiest way is to load the data in Python:
 
 .. code:: python
 
-   ps = ... # get data set from h5py
+   from picongpu.plugins.phase_space import PhaseSpace
+   import matplotlib.pyplot as plt
+   from matplotlib.colors import LogNorm
+   import numpy as np
 
-   # 3D3V example
-   dV = ps.attrs['dV'] * ps.attrs['dr_unit']**3
-   charge_per_bin = ps[:,:] * ps.attrs['sim_unit'] * dV
+
+   # load data
+   phase_space = PhaseSpace('/home/axel/runs/foil_001')
+   e_ps, e_ps_meta = phase_space.get('e', ps='ypy', iteration=1000)
+
+   # unit conversion from SI
+   mu = 1.e6  # meters to microns
+   e_mc_r = 1. / (9.1e-31 * 2.9979e8)  # electrons: kg * m / s to beta * gamma
+
+   # plotting
+   plt.imshow(
+       np.abs(e_ps) * e_ps_meta.dV,
+       extent = e_ps_meta.extent * [mu, mu, e_mc_r, e_mc_r],
+       interpolation = 'nearest',
+       aspect = 'auto',
+       origin='lower',
+       norm = LogNorm()
+   )
+
+   # annotations
+   cbar = plt.colorbar()
+   cbar.set_label(r'$Q / \mathrm{d}r \mathrm{d}p$ [$\mathrm{C s kg^{-1} m^{-2}}$]')
+
+   ax = plt.gca()
+   ax.set_xlabel(r'${0}$ [$\mathrm{\mu m}$]'.format(e_ps_meta.r))
+   ax.set_ylabel(r'$p_{0}$ [$\beta\gamma$]'.format(e_ps_meta.p))
+
+   plt.show()
+
+Note that the spatial extent of the output over time might change when running a moving window simulation.
 
 Out-of-Range Behavior
 ^^^^^^^^^^^^^^^^^^^^^
 
 Particles that are *not* in the range of ``<ValL>``/``<ValR>`` get automatically mapped to the lowest/highest bin respectively.
 Take care about that when setting your range and during analysis of the results.
-
-Spatial Offset
-^^^^^^^^^^^^^^
-
-The spatial extent of the output might change due to a moving window.
-Additional attributes are provided to retrieve that spatial information:
-
-See this python example:
-
-.. code:: python
-
-   ps = ... # get data set from h5py
-
-   mv_start = ps.attrs['movingWindowOffset']
-   mv_end = mv_start + ps.attrs['movingWindowSize']
-   spatial_offset = ps.attrs['_global_start'][1] # 2D data set: 0 (r_i); 1 (p_i)
-
-   dr = ps.attrs['dr'] * ps.attrs['dr_unit']
-
-   spatial_extend_cells = np.array([mv_start, mv_end]) + spatial_offset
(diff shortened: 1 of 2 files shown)


ax3l:
@PrometheusPi ready for review :)

docs/source/usage/plugins/phaseSpace.rst
```diff
@@ -37,47 +37,61 @@ Option                                 Usage
...
-Values are given as *charge density* per phase space bin.
-In order to scale to a simpler *charge of particles* per :math:`\mathrm{d}r_i` and :math:`\mathrm{d}p_i` -bin multiply by the cell volume:
+The histograms are stored in ``.hdf5`` files in the ``simOutput/...`` directory.
 
-.. code:: python
+...
 
-   ps = ... # get data set from h5py
+Values are given as *charge density* per phase space bin.
+In order to scale to a simpler *charge of particles* per :math:`\mathrm{d}r_i` and :math:`\mathrm{d}p_i` -bin multiply by the cell volume ``dV``.
```
PrometheusPi: OPTIONAL:
space between p_i and -bin could be removed

docs/source/usage/plugins/phaseSpace.rst
```diff
@@ -37,47 +37,61 @@ Option                                 Usage
...
 
-See this python example:
+   # load data
+   phase_space = PhaseSpace('/home/axel/runs/foil_001')
+   e_ps, e_ps_meta = phase_space.get('e', ps='ypy', iteration=1000)
 
-.. code:: python
+   # unit conversion from SI
+   mu = 1.e6  # meters to microns
+   e_mc_r = 3.6618e21  # electrons: kg * m / s to beta * gamma
```
PrometheusPi: OPTIONAL:
I would prefere loadung scipy constanst and then:
```python
import scipy.constants as const
e_mc_r = 1./(const.electron_mass * const.speed_of_light) # electrons: kg * m / s to beta * gamma
```

docs/source/usage/plugins/phaseSpace.rst
```diff
@@ -37,47 +37,61 @@ Option                                 Usage
...
-   mv_start = ps.attrs['movingWindowOffset']
-   mv_end = mv_start + ps.attrs['movingWindowSize']
-   spatial_offset = ps.attrs['_global_start'][1] # 2D data set: 0 (r_i); 1 (p_i)
+   # annotations
+   cbar = plt.colorbar()
+   cbar.set_label(r'$Q / \mathrm{d}r \mathrm{d}p$ [$C s kg^{-1} m^{-2}$]')
 
-   dr = ps.attrs['dr'] * ps.attrs['dr_unit']
+   ax = plt.gca()
+   ax.set_xlabel(r'${0}$ [$\mu m$]'.format(e_ps_meta.r))
```
PrometheusPi: QUESTION:
why do you write `{0}` instead of `{}`?

docs/source/usage/plugins/phaseSpace.rst
```diff
@@ -37,47 +37,61 @@ Option                                 Usage
...
 
-.. code:: python
+   # unit conversion from SI
+   mu = 1.e6  # meters to microns
+   e_mc_r = 3.6618e21  # electrons: kg * m / s to beta * gamma
 
-   ps = ... # get data set from h5py
+   # plotting
+   plt.imshow(
+       np.abs(e_ps) * e_ps_meta.dV,
```
PrometheusPi: QUESTION:
Doesn't `e_ps` contain under- and over-flow bins as well?
Shoudn't they be removed from plotting, it the range is used as below?

docs/source/usage/plugins/phaseSpace.rst
ax3l: that's true, but also pulls in a rather huge dependency

docs/source/usage/plugins/phaseSpace.rst
ax3l: I had this in a draft but did not add it to the functionality yet.
A section below it is mentioned, that the highest and lowest p bin will also contain overflows.

docs/source/usage/plugins/phaseSpace.rst
ax3l: this is just making it python 2.6 compatible. not really important

docs/source/usage/plugins/phaseSpace.rst
PrometheusPi: I know - iy you do not want to add a scipy depency how about 
```python
e_mc_r = 1./(9.1e-31 * 2.9979e8)
```
instead: at least these magic numbers are more familiar for physicists

docs/source/usage/plugins/phaseSpace.rst
```diff
@@ -37,47 +37,61 @@ Option                                 Usage
 Output
 ^^^^^^
 
-Values are given as *charge density* per phase space bin.
-In order to scale to a simpler *charge of particles* per :math:`\mathrm{d}r_i` and :math:`\mathrm{d}p_i` -bin multiply by the cell volume:
+The histograms are stored in ``.hdf5`` files in the ``simOutput/...`` directory.
 
-.. code:: python
+...
```
ax3l: oh, that is a TODO :D forgot to finish this ^^

docs/source/usage/plugins/phaseSpace.rst
PrometheusPi: how about `np.abs(e_ps[:, 1:-1])`and a brief comment - otherwise it gets confusing with the detailed (and good) description of the under- and over-flow bins below.

docs/source/usage/plugins/phaseSpace.rst
PrometheusPi: Ahh - thanks I didn't know that

docs/source/usage/plugins/phaseSpace.rst
ax3l: I had this since the bin is spanned by drdp and I did not want to raise the impresion its a dp-bin

docs/source/usage/plugins/phaseSpace.rst
ax3l: that's true. I am also thinking about adding `m_species * c` to the output of PS just to make it easier for integrated analysis not to have a table of each species' mass by hand.

docs/source/usage/plugins/phaseSpace.rst
PrometheusPi: Okay

docs/source/usage/plugins/phaseSpace.rst
ax3l: it's a bit more than that, removing two bins also changes the extent, etc. I just wanted to avoid overcomplicating it for the first round :)

docs/source/usage/plugins/phaseSpace.rst
ax3l: me neither, learned it while contributing to spack :)

ax3l:
@PrometheusPi changed the magic number in the example to more well-known magic numbers and finished the `...` in the docs

docs/source/usage/plugins/phaseSpace.rst
```diff
@@ -37,47 +37,60 @@ Option                                 Usage
...
+       origin='lower',
+       norm = LogNorm()
+   )
 
-.. code:: python
-
-   ps = ... # get data set from h5py
+   # annotations
+   cbar = plt.colorbar()
+   cbar.set_label(r'$Q / \mathrm{d}r \mathrm{d}p$ [$C s kg^{-1} m^{-2}$]')
```
PrometheusPi: REQUIRED/OPTIONAL:
The units are italic here - please use $\mathrm{C s kg ...}$ or something similar

(sorry this was for some reasons not posted yesterday and popped up as not committed today)

ax3l: thx, done

PR #2336, opened 2017-10-26
LWFA: GUI .cfg & Additional Parameters

ax3l:
#### Add to python input `Parameter` class

- dtype: data type of the PIConGPU value
- label: UI label instead of name
- formatter: UI value label instead of name

### New cfg for GUI: `0004gpus_gui.cfg`

 - make more easily modifiable with `tbg -o`:
   - steps
   - grid size
   - devices
- expose TBG_steps properly as integers.
- all plugins dump in same interval: `TBG_plugin_period`
- add histogram, second cut in PNG and phase space

### DIFF
--- a/lib/python/picongpu/input/parameters.py
+++ b/lib/python/picongpu/input/parameters.py
@@ -7,6 +7,7 @@ License: GPLv3+
 """
 
 import numpy as np
+import inspect
 
 
 class Parameter(object):
@@ -68,8 +69,16 @@ class Parameter(object):
         --------
         A dictionary with the member variables as (key, value) pairs.
         """
+        d = dict(
+            cls=type(self).__name__,
+            name=self.name,
+            type=self.type,
+            unit=self.unit,
+            default=self.default,
+            value=self.value,
+            dtype=self.dtype.__name__)
 
-        return self.__dict__
+        return d
 
     def macro_name(self):
         """
@@ -175,8 +184,10 @@ class UiParameter(Parameter):
         """
 
         members = super(UiParameter, self).as_dict()
-        del members["formatter"]
-        del members["dtype"]
+        members["label"] = self.label
+        members["formatter"] = str(inspect.getsourcelines(self.formatter)[
+                                   0]).strip("['\\n']").split(" = ")[1]
+
         return members
 
 
--- a/lib/python/picongpu/input/parameters.py
+++ b/lib/python/picongpu/input/parameters.py
@@ -6,7 +6,7 @@ Authors: Sebastian Starke, Jeffrey Kelling
 License: GPLv3+
 """
 
-from numpy import log10
+import numpy as np
 
 
 class Parameter(object):
@@ -14,7 +14,7 @@ class Parameter(object):
     Base class for exposable parameters.
     """
 
-    def __init__(self, name, ptype, unit, default, value=None):
+    def __init__(self, name, ptype, unit, default, value=None, dtype=None):
         """
         Parameters
         ----------
@@ -31,6 +31,9 @@ class Parameter(object):
         value: float (optional)
             The value of the parameter object. If it is not provided, the value
             of the 'default' argument is used.
+        dtype: type (optional)
+            The data type of the parameter value. If it is not provided, the
+            type of the 'default' argument is used.
         """
         self.name = name
         self.type = ptype  # decides if compile time or runtime parameter
@@ -38,6 +41,7 @@ class Parameter(object):
         self.default = default
 
         self.value = value if value is not None else default
+        self.dtype = dtype if dtype is not None else type(default)
 
     def __str__(self):
         """
@@ -101,7 +105,8 @@ class UiParameter(Parameter):
     """
 
     def __init__(self, name, ptype, unit, default, slider_min, slider_max,
-                 slider_step, value=None):
+                 slider_step, label=None, formatter=None,
+                 value=None, dtype=None):
         """
         Parameters
         ----------
@@ -113,14 +118,27 @@ class UiParameter(Parameter):
             The maximal value of the slider widget
         slider_step: float
             The stepsize when adjusting the slider
+        label: string [optional]
+            Overwrite the name for UI
+        formatter: function or lambda of form f(x) -> string [optional]
+            representation of the current value for UI
 
         """
-        Parameter.__init__(self, name, ptype, unit, default, value)
+        Parameter.__init__(self, name, ptype, unit, default, value, dtype)
 
         # for slider widget creation
         self.min = slider_min
         self.max = slider_max
         self.step = slider_step
+
+        if label is None:
+            self.label = name
+        else:
+            self.label = label
+        if formatter is None:
+            self.formatter = lambda x: str(x)
+        else:
+            self.formatter = formatter
 
     def set_value(self, x):
         """
@@ -136,8 +154,8 @@ class UiParameter(Parameter):
         A float containing the updated 'value' attribute. Is printed
         below the slider-widget for user-feedback.
         """
-        self.value = x
-        return self.value
+        self.value = self.dtype(x)
+        return self.formatter(self.value)
 
     def get_value(self):
         """
@@ -147,6 +165,20 @@ class UiParameter(Parameter):
         """
         return self.value
 
+    def as_dict(self):
+        """
+        Convert Parameter object into a plain dictionary
+
+        Returns
+        --------
+        A dictionary with the member variables as (key, value) pairs.
+        """
+
+        members = super(UiParameter, self).as_dict()
+        del members["formatter"]
+        del members["dtype"]
+        return members
+
 
 class LinearScaledParameter(UiParameter):
     """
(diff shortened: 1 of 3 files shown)


lib/python/picongpu/input/parameters.py
```diff
@@ -147,6 +165,19 @@ def get_value(self):
...
+        """
+        Convert Parameter object into a plain dictionary
+
+        Returns
+        --------
+        A dictionary with the member variables as (key, value) pairs.
+        """
+
+        members = super(UiParameter, self).as_dict()
+        del members["formatter"]
```
codingS3b: Seems like we need to exclude the "dtype" member as well since it can't be serialzed either

lib/python/picongpu/input/parameters.py
```diff
@@ -208,7 +241,8 @@ class LogScaledParameter(Parameter):
     """
 
     def __init__(self, name, ptype, unit, default, slider_min, slider_max,
```
codingS3b: LogScaledParameter should actually inherit from UiParameter. I commited that wrong and it needs to be fixed.

share/picongpu/examples/LaserWakefield/lib/python/picongpu/params.py
```diff
@@ -36,6 +39,12 @@
 
     LinearScaledParameter(
         name="Pulse_Length_SI", ptype="compile", unit="s",
-        default=5, slider_min=1, slider_max=150,
+        default=5.0, slider_min=1.0, slider_max=150.0,
         slider_step=1, scale_factor=1.e-15),
+
+    LinearScaledParameter(
+        name="TBG_steps", ptype="run", unit="s",
```
codingS3b: It would probably be less confusing to use the units the user is accustomed to like pico-seconds, nano-meters etc as units. The formatted internal picongpu value of each parameter is, as discussed yesterday, indeed not relevant to the user. Sorry for switching back and forth about this issue.

lib/python/picongpu/input/parameters.py
```diff
@@ -240,8 +275,8 @@ def set_value(self, x):
         A float containing the internal 'value' attribute after adjustment.
         """
 
-        self.value = self.base ** x
-        return self.value
+        self.value = self.dtype(self.base ** x)
+        return self.formatter(self.value)
 
     def get_value(self):
```
codingS3b: Before calling ```log10(self.value)``` we need to cast the value back to float, because in case we use ```self.dtype = int```, I got the error `AttributeError: 'int' object has no attribute 'log10'`

share/picongpu/examples/LaserWakefield/lib/python/picongpu/params.py
codingS3b: For the `Wave_Length_SI` parameter the default should be a float or the dtype should be specified as float, since right now the dtype is taken as `int` causing much trouble in the computations that follow.

ax3l:
@codingS3b thanks for the review, I addressed all inline comments.
can you pls take a look again?

lib/python/picongpu/input/parameters.py
codingS3b: It just occured to me that we could store the the dtype as a string representation using
`members["dtype"] = members["dtype"].__name__`. Maybe it would be cleaner to do this for later restoring the parameter objects from the json representation correctly.

lib/python/picongpu/input/parameters.py
```diff
@@ -166,7 +199,8 @@ def __init__(self, name, ptype, unit, default, slider_min, slider_max,
             internal value.
         """
         UiParameter.__init__(self, name, ptype, unit, default,
-                             slider_min, slider_max, slider_step, value)
+                             slider_min, slider_max, slider_step,
+                             label, formatter, value, dtype)
```
codingS3b: for some reason when executing
```
p = LinearScaledParameter(
         name="Wave_Length_SI", ptype="compile", unit="nm",
         default=800.0, slider_min=400.0, slider_max=1400.0,
         slider_step=1, scale_factor=1.e-9)
p.formatter
```        
I am getting an `AttributeError: 'LinearScaledParameter' object has no attribute 'formatter'` error. Strange since I thought this should be inherited from 'UiParameter' or 'Parameter' class ...

lib/python/picongpu/input/parameters.py
ax3l: ok, but careful that a lambda does not mean it is just *one* source line ;)

ax3l: shall we add the string representation of both in a follow-up? do we need it now?
otherwise feel free to approve the PR :)

lib/python/picongpu/input/parameters.py
```diff
@@ -113,15 +118,28 @@ def __init__(self, name, ptype, unit, default, slider_min, slider_max,
...
         slider_step: float
             The stepsize when adjusting the slider
+        label: string [optional]
+            Overwrite the name for UI
+        formatter: function or lambda of form f(x) -> string [optional]
+            representation of the current value for UI
 
         """
-        Parameter.__init__(self, name, ptype, unit, default, value)
+        Parameter.__init__(self, name, ptype, unit, default, value, dtype)
```
PrometheusPi: QUESTION:
`dtype`is (as far as I can tell) not used below, why is is then used as input parameter?
Add a todo if you want to add functionality using `dtype` in the future.

share/picongpu/examples/LaserWakefield/lib/python/picongpu/params.py
```diff
@@ -18,10 +18,13 @@
 
 from picongpu.input.parameters import LogScaledParameter, LinearScaledParameter
 
+pico = 1.e-12
+dt_r = 1. / 1.39e-16 * pico
```
PrometheusPi: REQUIRED:
Please explain this magic number (`1.39e-16`) in a comment or give it a name that makes it clear (as e.g. pico).

share/picongpu/examples/LaserWakefield/etc/picongpu/0004gpus_gui.cfg
```diff
@@ -0,0 +1,86 @@
...
+                 --e_energyHistogram.minEnergy 0 --e_energyHistogram.maxEnergy 100000"
+
+# Calculate a 2D phase space
+# - requires parallel libSplash for HDF5 output
+# - momentum range in m_<species> c
+TBG_e_PSxpx="--e_phaseSpace.period !TBG_plugin_period --e_phaseSpace.space x --e_phaseSpace.momentum px --e_phaseSpace.min -8.0 --e_phaseSpace.max 8.0"
+TBG_e_PSypy="--e_phaseSpace.period !TBG_plugin_period --e_phaseSpace.space y --e_phaseSpace.momentum py --e_phaseSpace.min  0.0 --e_phaseSpace.max 1.0"
+TBG_e_PSzpz="--e_phaseSpace.period !TBG_plugin_period --e_phaseSpace.space z --e_phaseSpace.momentum pz --e_phaseSpace.min -2.0 --e_phaseSpace.max 2.0"
+
+TBG_plugins="!TBG_pngYX                    \
```
PrometheusPi: QUESTION:
Why is this so much different to the other `*.cfg` files that it needs to be a separate file?
Couldn't we just use a combined `*.cfg`file to avoid duplicated code/pile up.

ax3l: We want to control steps directly from the outside, but our tools do not understand `-s number` but only tbg variables such as `!TBG_steps`, controlled with `tbg -o`.

We will also add probably more changes here, so it is best to make it a fully separate setup.

PrometheusPi: Okay - that's a good idea :+1:

share/picongpu/examples/LaserWakefield/lib/python/picongpu/params.py
ax3l: this is the inverse (historically suffixed `_r`) of `dt`.

We might be able to just add is as a `Parameter` (but not `UiParameter`) in the list below to make it more accessible.

lib/python/picongpu/input/parameters.py
ax3l: dtype is used to init a member of the parent class.

it is then used in all `set_value`

PrometheusPi: Sorry for the confusion - my fault - it is used right there

lib/python/picongpu/input/parameters.py
codingS3b: I think for the moment we can leave them out, but maybe the class of each parameter object should be stored as a string representation.
This is because, as of right now, we still have issues in our ui-code when reading back the parameter variables from json: there, we implicitely have to cast to the base class "Parameter" (since we don't have any other information) losing all info concerning the dtype and formatter which the program then tries to access, of course crashing.
So I still need to figure out a proper way to restore an "exact" copy of the parameter object on our side of the code. But the dtype and the formatter are handled with defaults when not given, so that should be not a problem.

codingS3b: Hold on, sorry...maybe our bug is somewhere else and we don't need the class representation. Don't change it just yet, I will get back to you :)

codingS3b: I found the critical section: calling `del members["formatter"]` seems to remove the entry not only from the members dictionary, but since it is a reference also from the true object.

ax3l:
@PrometheusPi @codingS3b if you are ok with the intermediate state of this feature we can already merge it so @codingS3b can work on with the `dev`

PR #2076, opened 2017-05-30
fix nvcc compile with boost 1.64.0

psychocoderHPC:
fix #2048 follow-up to #2069

disable boost noexept if nvcc with (CUDA 8.0) is the device compiler and boost 1.64.0 is used

# Tests

- [x] nvcc compile with boost 1.64.0
- [x] clang compile with boost 1.64.0

PR #2342, opened 2017-10-27
CMake: Re-Enable Boost 1.64.0 + NVCC 8.0 Work-Around

ax3l:
Replace missing `PMACC_CUDA_COMPILER` with `ALPAKA_CUDA_COMPILER`.

The work-around was added in #2076 but the controlling var vanished during the marriage with alpaka in #2178 (https://github.com/ComputationalRadiationPhysics/picongpu/commit/0cc58f6443d15ed36aeb1a2fd9d9399ae0aed647#diff-d1fb8e3eeab4d2001da0932360713480).

Boost report: https://svn.boost.org/trac10/ticket/13049

Add missing work-around status' message to `PMaccConfig.cmake` when enabled.

### DIFF
--- a/include/pmacc/PMaccConfig.cmake
+++ b/include/pmacc/PMaccConfig.cmake
@@ -285,6 +285,7 @@ endif()
 # Boost 1.55 added support for a define that makes result_of look for
 # the result<> template and falls back to decltype if none is found. This is
 # great for the transition from the "wrong" usage to the "correct" one as
+message(STATUS "Boost: result_of with TR1 style and decltype fallback")
 set(PMacc_DEFINITIONS ${PMacc_DEFINITIONS} -DBOOST_RESULT_OF_USE_TR1_WITH_DECLTYPE_FALLBACK)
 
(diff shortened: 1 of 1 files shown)


include/pmacc/PMaccConfig.cmake
```diff
@@ -314,6 +318,7 @@ endif()
 if( ("${PMACC_CUDA_COMPILER}" STREQUAL "nvcc") AND
```
ax3l: @BeyondEspresso here is your bug on Taurus (TUD): `PMACC_CUDA_COMPILER` is an empty string which makes the work-around not active as it should.

will investigate why.

ax3l: The work-around was added in #2076 but the controlling var vanished during the marriage with alpaka in #2178

ax3l: We can just use `ALPAKA_CUDA_COMPILER` now.

PR #2348, opened 2017-11-01
add support for CUDA9 `__shfl_snyc, __ballot_sync`

psychocoderHPC:
This is a follow up of #2333

- use `__shfl_sync` with CUDA9
- use `__ballot_sync` with CUDA9

### DIFF
--- a/include/pmacc/nvidia/atomic.hpp
+++ b/include/pmacc/nvidia/atomic.hpp
@@ -89,7 +89,11 @@ namespace nvidia
             operator()(const T_Acc& acc,T_Type* ptr, const T_Hierarchy& hierarchy)
             {
                 /* Get a bitmask with 1 for each thread in the warp, that executes this */
+#if(__CUDACC_VER_MAJOR__ >= 9)
+                const int mask = __ballot_sync(0xFFFFFFFF, 1);
+#else
                 const int mask = __ballot(1);
+#endif
                 /* select the leader */
                 const int leader = __ffs(mask) - 1;
                 T_Type result;
@@ -180,7 +184,11 @@ DINLINE void
 atomicAllExch(const T_Acc& acc, T_Type* ptr, const T_Type value, const T_Hierarchy& hierarchy)
 {
 #if (__CUDA_ARCH__ >= 200)
+#   if(__CUDACC_VER_MAJOR__ >= 9)
+    const int mask = __ballot_sync(0xFFFFFFFF, 1);
+#   else
     const int mask = __ballot(1);
+#   endif
     // select the leader
     const int leader = __ffs(mask) - 1;
     // leader does the update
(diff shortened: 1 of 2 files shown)


Flamefire:
I'm currently "relearning" CUDA warps now and remembered we had the elaborate `atomicAllInc` function in PMacc which uses warp intrinsics and might got broken by CUDA 9.

IMO this PR will result in undefined or wrong behaviour for prior Volta. Example: `if(data[threadIdx] == foo) atomicAllInc(...)`

This boils down to `if(data[threadIdx] == foo) const int mask = __ballot_sync(0xFFFFFFFF, 1);`

2 problems: First `__ballot_sync` returns an unsigned not an int, second the mask is wrong:

> All active threads named in mask must execute the same intrinsic with the same mask, or the result is undefined.
http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#warp-vote-functions

And

>  Note, however, that for Pascal and earlier architectures, all threads in mask must execute the same warp intrinsic instruction in convergence, and the union of all values in mask must be equal to the warp's active mask.
http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#independent-thread-scheduling-7-x

Although this is quite hard to understand (for me) I'd assume from the first quote that the implementation is ok. But I don't get the `must execute the same intrinsic` part because how could they execute different intrinsics when they are "active" (executing the current instruction)? `the same mask` is given as the mask is a constant.

But the 2nd quote strikes. I assume the `union of all values in mask` (again, what is this? Didn't they say before the mask must be the same?) is something like an `OR` of all active threads mask and hence in this case `0xFFFFFFFF`. But this is not equal to the `warp's active mask` which is what `__ballot(1)` used to be and is now `__activemask()`.

So my requests:

-  Could anyone explain the questions above to me regarding mask and "same intrinsic"?
- **At least** change the `__ballot_sync` to `__activemask()` and fix the types
- Note that the linked nvidia blog entry has been [updated for CUDA 9](http://devblogs.nvidia.com/parallelforall/cuda-pro-tip-optimized-filtering-warp-aggregated-atomics/) and that example is now also in the [programming guide](http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#discovery-pattern-cg) which shows a better alternative which also does not rely on inline ptx for getting the lane-id
- The blog post mentions, that this optimization is done by the compiler for post Kepler in CUDA 7.5 and Kepler+ in CUDA 9 with additionally even higher performance. As the warp-level programming got much harder with CUDA 9/Volta I'd suggest refraining from it as much as possible. In this case the whole `atomicAllInc` can be replaced by a simple atomicAdd for CUDA 9 as it worked on Kepler+ only anyway.

ax3l:
cc @psychocoderHPC

PR #2358, opened 2017-11-05
Spack: Own GitHub Org

ax3l:
Spack has its own GitHub organization now 🚀 ✨ 

This changes all links from LLNL to it.


PrometheusPi:
The old URL is redirected to the new one - so it would still work (for now).
@ax3l Good job in preventing any failure if they stop the redirect.

PR #2330, opened 2017-10-25
Add upper energy cut-off to ThomasFermi model

n01r:
Separate bulk from prompt electrons by defining an energy cut-off.
Above this threshold value electrons are classified as "prompt" electrons
and are therefore neglected in the calculation of electron temperature.

The actual cut-off value is highly debated, see [Kemp, Sentoku](http://link.aps.org/doi/10.1103/PhysRevLett.116.159501).

This is one of 4 planned improvements of the `ThomasFermi` implementation.

A test with a directed stream of electrons has shown that if the electrons have energies higher than the cut-off they will not be considered for the "temperature" calculation.
The test features a 2-dimensional 256 x 256 grid with period boundaries in every direction and about 6 copper ions distributed in every cell.
The ions are preionized once and electrons are derived from them.
There are three test setups:
1. electron temperature of 0 eV
2. electron temperature of 100 eV
3. electron temperature of 0 eV and drift of 200 keV
An energy cut-off of 50 keV has been configured.

The results show that in case 1 and 2 the predicted charge state matches the one reached in the simulation.
The TF model predicts an initial charge state for copper at density 8.92 g/cm³ and T_e = 0 eV of 4+.
Case 3 performs exactly as case 1 until the fast electrons distribute some of their energy to the cold electrons and TF ionization continues.

~~- [ ] Perform a more advanced test with disabled current deposition and a velocity gradient that depends on the cell index perpendicular to the stream direction   
Thereby redistribution of electron energies is avoided and we should see the cut-off spatially in a jump of ion charge state / electron density~~
**update 2017-11-07**   
A simple test has been done and it shows that the implementation works *BUT* a more detailed test will be relocated to the issue #1967.

Expectation:
 
04_high_energy_cutoff_gradient_test_expectation

- [x] Add documentation to the `/docs` in the TF section

cc'ing @HighIander 

This PR closes #1967 .

PR #2361, opened 2017-11-07
Add lower ion density cut-off to ThomasFermi model

n01r:
In tests that I did a while back, I found an unphysical prediction of ionization states that the Thomas-Fermi fitted formula by More, 1985 delivered. 
tf_prediction_argon_low_density_artifacts
This applies to typical ion densities for gases but also pre-plasmas and can, depending on electron temperature, predict the ionization of a charge state that is much more deeply bound than the average energy of a free electron that would be responsible for the collision process.
A simple man's approach to this issue is to set a cut-off value for low densities.
It is advisable to strongly motivate the exact value of this cut-off parameter with independent calculations done for instance with FLYCHK.

**update 2017-11-08** This is how the average charge state at liquid argon density looks like for electron temperatures from 0 to 450 electron volts.
auswahl_057

These calculations should be done for each material separately, at the moment.

- [x] This pull request depends on #2330 and should be merged afterwards

The value for the lower density cutoff is currently set to the critical electron density for an 800nm laser.
A test has been performed where I placed randomly distributed copper ions of into a 2D grid with a linear density gradient from 0 to roughly 2 critical densities along the y-direction. These ions have been ionized once and a temperature of 100 eV has been applied to the electrons that are derived from the ions.

simple_test_low_density_cutoff

The cutoff works as expected and the `ThomasFermi` model is inactive for particles which "see" an interpolated ion density smaller than the cutoff.
In the picture above, the red contours mark the density threshold.

### DIFF
--- a/include/picongpu/simulation_defines/param/ionizer.param
+++ b/include/picongpu/simulation_defines/param/ionizer.param
@@ -437,6 +437,19 @@ namespace thomasFermi
     /** cutoff energy for electron "temperature" calculation in SI units*/
     constexpr float_X CUTOFF_MAX_ENERGY = CUTOFF_MAX_ENERGY_KEV * UNITCONV_keV_to_Joule;
 
+    /** lower density cutoff
+     *
+     * The Thomas-Fermi model yields unphysical artifacts for low ion densities.
+     * Low ion densities imply lower collision frequency and thus less collisional ionization.
+     * The Thomas-Fermi model yields an increasing charge state for decreasing densities and electron temperatures of 10eV and above.
+     * This cutoff will be used to set the lower application threshold for charge state calculation.
+     *
+     * @note This cutoff value should be set in accordance to FLYCHK calculations,
+     *       for instance! It is not a universal value and requires some preliminary
+     *       approximations!
+     */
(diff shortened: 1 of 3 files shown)


n01r:
Oh! I forgot to add this contribution to the `/docs` section.

include/picongpu/simulation_defines/param/ionizer.param
```diff
@@ -437,6 +437,19 @@ namespace thomasFermi
...
+     * The Thomas-Fermi model yields unphysical artifacts for low ion densities.
+     * Low ion densities imply lower collision frequency and thus less collisional ionization.
+     * The Thomas-Fermi model yields an increasing charge state for decreasing densities and electron temperatures of 10eV and above.
+     * This cutoff will be used to set the lower application threshold for charge state calculation.
+     *
+     * @note This cutoff value should be set in accordance to FLYCHK calculations,
+     *       for instance! It is not a universal value and requires some preliminary
+     *       approximations!
+     */
+    constexpr float_X CUTOFF_LOW_DENSITY = 1.7422e27; // m^-3 - critical density 800nm laser
```
ax3l: note: what you mean here is the critical (hypothetically free) electron density. the "critical ion density" would be much different.

PR #2364, opened 2017-11-07
New Trait: SpeciesEligibleForSolver

ax3l:
Allows to check if a defined `picongpu::Particles` class is eligible for computations with a specific solver.

Algorithms, Plugins and core methods can implement this trait to signal which identifiers (particle attributes), flags (particle frame attributes) or other requirements they have regarding particle species.

Use as follows:

```C++
using isEligible = traits::SpeciesEligibleForSolver<
    PIC_Electrons,      // any picongpu::Particles species
    ChargeConservation  // any solver class here
>::type; // this is a boost::mpl::bool_

// or if you need a bool
bool const okToUse = isEligible::value;
```

Of course you can also filter for the species from `VectorAllSpecies` for which a solver can be compiled:

```C++
using SeqEligibleSpecies = typename bmpl::copy_if<
    VectorAllSpecies,
    traits::SpeciesEligibleForSolver<
        bmpl::_1,
        ChargeConservation
    >
>::type; // this is a standard boost::mpl forward sequence
```

### DIFF
--- a/include/picongpu/particles/traits/SpeciesEligibleForSolver.hpp
+++ b/include/picongpu/particles/traits/SpeciesEligibleForSolver.hpp
@@ -1,0 +1,50 @@
+/* Copyright 2017 Axel Huebl
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


include/picongpu/traits/SpeciesEligibleForSolver.hpp
```diff
@@ -0,0 +1,46 @@
...
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <boost/mpl/bool.hpp>
+
+namespace picongpu
+{
+namespace traits
```
psychocoderHPC: please move this traid to namespace `picongpu::particles::traits`

ax3l: moved! :)

PR #2366, opened 2017-11-07
HDF5: Allow Unphysical Particle Dump

ax3l:
Allow to dump particles that do not have a charge or mass frame attribute in HDF5.

Note: we allow to define particles without such fundamental types in PIConGPU (instead of falling back to zero) since this allows us to differentiate between physical particles (such as photons or neutrals) and unphysical particles (test particles, probes, etc.).

If such a particle species in encountered, we simply do not write the constant openPMD records for charge and mass (instead of writing implicitly a zero value, which would suggest these particles are physical - which they are not).

### DIFF
--- a/include/picongpu/plugins/hdf5/WriteSpecies.hpp
+++ b/include/picongpu/plugins/hdf5/WriteSpecies.hpp
@@ -48,6 +48,7 @@
 #include <boost/type_traits.hpp>
 
 #include <string>
+#include <type_traits>
 
 
 namespace picongpu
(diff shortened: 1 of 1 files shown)


include/picongpu/plugins/hdf5/WriteSpecies.hpp
```diff
@@ -62,6 +62,58 @@ TYPE_ARRAY(UInt64_5, H5T_INTEL_U64, uint64_t, 5);
 
 using namespace splash;
 
+namespace detail
+{
+    template<
```
psychocoderHPC: Note this can be also written as:

```C++
template<
    typename T_FrameType
>
struct GetChargeOrZero
{
    static constexpr bool hasChargeRatio = pmacc::traits::HasFlag<
        T_FrameType,
         chargeRatio<>
     >::type::value;

     auto operator()()
     -> typename std::enable_if< hasChargeRatio, float_X >::type
     {
         return frame::getCharge< T_FrameType >();
     }

     auto operator()()
     -> typename std::enable_if< !hasChargeRatio, float_X >::type
     {
         return float_X( 0. );
     }
};
```

psychocoderHPC: A little bit docu: https://stackoverflow.com/questions/14600201/why-should-i-avoid-stdenable-if-in-function-signatures

PR #2371, opened 2017-11-08
Particle Pusher: Probe

ax3l:
Add a new particle pusher (probe).

Interpolates the electro-magnetic fields as a usual pusher but instead of pushing the particle (changing the momentum) it saves the result on particle attributes.

As an additional feature, this allows to add an additional pusher attribute. This allows to *probe* the fields and at the same time move the particle (free streaming with a moving window) or even to create a tracer particle (which fully interacts with the simulation).

Also adds two species attributes that can be used to store a particle's E and B field as seen after fieldToParticle interpolation with respect to shape.

More details on the usage will follow in the main PR.

### DIFF
--- a/include/picongpu/particles/pusher/particlePusherProbe.hpp
+++ b/include/picongpu/particles/pusher/particlePusherProbe.hpp
@@ -1,0 +1,157 @@
+/* Copyright 2017 Axel Huebl
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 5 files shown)


n01r:
**Bam!** Awesome work! :sparkles:

include/picongpu/particles/pusher/particlePusherProbe.hpp
```diff
@@ -0,0 +1,164 @@
...
+    template<
+        typename T_ValueFunctor = pmacc::nvidia::functors::Assign,
+        typename T_ActualPush = void
+    >
+    struct Push
+    {
+        /* this is an optional extension for sub-sampling pushes that enables grid to particle interpolation
+         * for particle positions outside the super cell in one push
+         */
+        using LowerMargin = typename pmacc::math::CT::make_Int<
```
psychocoderHPC: Please use the margins of the child pusher.

PR #2378, opened 2017-11-10
Start Pos: OnePosition w/o Weighting

ax3l:
Allows to transparently use the `OnePosition` particle initialization `startPosition` with particles without weighting (such as probe particles).

This will straight initialize the number of particles per cell (ppc). It is recommended to only use one ppc with this start position.

### DIFF
--- a/include/picongpu/particles/startPosition/OnePositionImpl.hpp
+++ b/include/picongpu/particles/startPosition/OnePositionImpl.hpp
@@ -23,6 +23,8 @@
 #include "picongpu/particles/startPosition/OnePositionImpl.def"
 #include "picongpu/particles/startPosition/detail/WeightMacroParticles.hpp"
 
+#include <pmacc/traits/HasIdentifier.hpp>
+
 #include <boost/mpl/integral_c.hpp>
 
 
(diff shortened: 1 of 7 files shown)


include/picongpu/particles/startPosition/OnePositionImpl.hpp
```diff
@@ -56,19 +90,37 @@ namespace acc
...
-                realParticlesPerCell,
-                T_ParamClass::numParticlesPerCell,
-                m_weighting
-            );
+            bool const hasWeighting = pmacc::traits::HasIdentifier<
+                typename T_Particle::FrameType,
+                weighting
+            >::type::value;
+
+            if( hasWeighting )
```
psychocoderHPC: Could you please transform this if that we have only one exit point within the method. This is a better pattern for compiler to optimize the code.

```C++

uint32_t result = T_ParamClass::numParticlesPerCell;
if( hasWeighting )
    result = startPosition::detail::WeightMacroParticles{}(
       realParticlesPerCell,
       T_ParamClass::numParticlesPerCell,
       m_weighting
    );
return result;
```

include/picongpu/particles/startPosition/OnePositionImpl.hpp
```diff
@@ -56,19 +90,37 @@ namespace acc
...
             particle[ position_ ] = T_ParamClass{}.inCellOffset.template shrink< simDim >( );
-            particle[ weighting_ ] = m_weighting;
+
+            // set the weighting attribute if the particle species has it
+            bool const hasWeighting = pmacc::traits::HasIdentifier<
+                typename T_Particle::FrameType,
+                weighting
+            >::type::value;
+            detail::SetWeighting< hasWeighting > setWeighting;
+            setWeighting( particle, m_weighting );
```
psychocoderHPC: please write

```C++
setWeighting( 
    particle, 
    m_weighting 
);
```

Issue #1288, opened 2015-12-03
Species Plugins: Multi & --<name>.filter


ax3l:
One thing I realised during the last years of large scale runs is the following:
### Objective

Make every plugin that run over `species` should be **multi** plugins (that can be called multiple times) and should define an _optional argument_ `<name>.filter=<filterName>` (default=`all`).
### Advantage

The same plugin can be run multiple times over a species but only selects a sub-selection of that species by user defined filters. These filters can be something like position, direction of propagation, energy, ... or what ever might be interesting for a user.
### Implementation

We should provide a `speciesFilter.param` where the user can define very simple functors of the form `bool isRelevant( T_Particle& particle )`. Additionally, we need a `getName` method so we can use it on the cmd line.

Of course we pre-define and include an `all` filter that is the default fallback in case no filter was used.

As with other "multi" plugins, the plugin is simply run multiple times (even if the calculation could be done in one swipe over the particle list) since this is the easiest and most generic implementation.

Functors should be designed to easily allow filters based on the simple access to the particle's attributes (and its frame's attributes) but should be also not too strict so we can also forward field attributes later on.
### Examples

select only particles:
- in a certain **direction and solid angle**
- use _more than one_ **energy binning detector** in a simulation, e.g., individually forward & transversal
- in a certain **energy range**
- **inside the moving window** #89
- store **N videos** zooming on various places of the simulation (see [long shot vs. close-up vs. detail etc.](https://de.wikipedia.org/wiki/Einstellungsgr%C3%B6%C3%9Fe)) for offline (or live) analysis as in the [bullet time paper](http://arxiv.org/pdf/1301.4546v1.pdf)
- ...

n01r:
:+1: Need that one, too.
To be able to compare PIConGPU to PICLS we would need to have cut-off on the electron energy to be able to compute the "temperature" of the thermalized electrons for the Thomas-Fermi pressure ionization model.

n01r:
I would say that we need filters for more than just plugins but also to be able to implement certain models in the main code.

ax3l:
yes, the "filter" kernels should of course be general `PMacc` methods.

PrometheusPi:
Sounds like a great idea :+1:.

steindev:
[...]

PR #2381, opened 2017-11-13
rename filter: `IsHandleValid` -> `All`

psychocoderHPC:
This pull request changed the name for the particle filter which is necessary to allow the usage later in our plugin e.g. `--e_energy.filter All`

- rename particle filter
- add static `getName()` method to the filter
- update usage
- update documentation

related to #1288

### DIFF
--- a/include/picongpu/particles/filter/All.hpp
+++ b/include/picongpu/particles/filter/All.hpp
@@ -33,7 +33,7 @@ namespace acc
 {
 
     //! check the particle handle
-    struct IsHandleValid
+    struct All
     {
 
         /** check particle handle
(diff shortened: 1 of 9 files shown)


include/picongpu/particles/filter/All.hpp
```diff
@@ -79,16 +79,23 @@ namespace acc
...
-            return acc::IsHandleValid{ };
+            return acc::All{ };
 
         }
+
+        static
+        HINLINE std::string
+        getName( )
+        {
+            return std::string("All");
```
ax3l: I would write them lowerCase

ax3l: we can do this in the end

psychocoderHPC:
I will rename all `All` to lower case letters in  a follow up.

PR #2377, opened 2017-11-09
FieldTmp: SpeciesEligibleForSolver Traits

ax3l:
Define the `SpeciesEligibleForSolver` trait for FieldTmp derived
particle attributes.

Refactor the generators to be more generic and user friendly.

## Changed Input
- `fileOutput.param`:
- `isaac.param`

before:
```C++
// will fail on particles without a defined `chargeRatio<>`
using ChargeDensity_Seq = bmpl::transform<
    VectorAllSpecies,
    CreateChargeDensityOperation< bmpl::_1 >
>::type;
```
after
```C++
// does not fail and skips code generation for particles with undefined charge density
using ChargeDensity_Seq = deriveField::CreateEligible_t<
    VectorAllSpecies,
    deriveField::derivedAttributes::ChargeDensity
>;
```

## Side-Effects

Fix leftover bugs and changed interfaces due to
- properly introduced namespaces
- removed ill-set using namespace in input file
.

Please read the PR commit-wise.

## Depends On

- [x] #2380

PR #2368, opened 2017-11-07
PluginController: Eligible Species

ax3l:
Only specialize and compile particle plugins for particle species which are elibible for it, based on the `SpeciesEligibleForSolver` trait.

Silently disables components which are not fulfilled, e.g. for nonphysical particles such as probes. Moves user (input) errors from compile time (which are in this case always very long and unreadable template meta programming errors) to runtime where it is (ideally) easier to spot.

Possible follow-up: For further usability, we can add log messages on high-level debug levels to output which species are ignored on `pluginLoad()`.

Depends on

- [x] #2364
- [x] #2367
- [x]  #2380 
- [x] #2394
- [x] CT: no significant impact after 2nd commit

## Compile Times LWFA

**CUDA 7.5**
`dev`: 6:03 (real) | 7:38 (user)
`PR`: 6:08m (real) | 7:40m (user)

**CUDA 8.0.44** + ISAAC
`dev`: 6:35m (real) | 7:36m (user)
`PR`: 6:46m (real) | 7:43m (user)

PR #2384, opened 2017-11-13
Probe Particle Usage

ax3l:
Documentation on how to use probe particles ("probes") in PIConGPU. Adds probe particles to the `FoilLCT` ion acceleration example.

Probe particles (“probes”) can be used to record field quantities at selected positions over time.

As a geometric data-reduction technique, analyzing the discrete, regular field of a particle-in-cell simulation only at selected points over time can greatly reduce the need for I/O. Such particles are often arranged at isolated points, regularly as along lines, in planes or in any other user-defined manner.

Probe particles are usually neutral, non-interacting test particles that are statically placed in the simulation or co-moving with along pre-defined path. Self-consistently interacting particles are usually called tracer particles.

0563c0b16c6d2ba590c28c012529675f6a9d669c also adds a new density profile putting particles "every nth cell":  Useful to initialize probe particles or material dopings. The result is either 0 (no particle) or the full density. The result of this particular functor can be larger 1.0 with `T_SkipCells::toRT().productOfComponents()` in order to properly fulfill the density of a species via increased weighting.

### Example Plots

FoilLCT step 10'000: Ex raw fields vs. probe (every 4x4x4 with varying shapes) image
[plotFields.py](https://github.com/ComputationalRadiationPhysics/picongpu/files/1491639/plotFields.py.txt)

probes_ex

### Depends on

- [x] RT tests
- [x] #2368
- [x] #2377
- [x] #2394
- [x] #2400

### DIFF
--- a/docs/source/usage/workflows/probeParticles.rst
+++ b/docs/source/usage/workflows/probeParticles.rst
@@ -1,0 +1,116 @@
+.. _usage-workflows-probeParticles:
+
+Probe Particles
+---------------
+
+.. sectionauthor:: Axel Huebl
+
+Probe particles ("probes") can be used to record field quantities at selected positions over time.
+
+As a geometric data-reduction technique, analyzing the discrete, regular field of a particle-in-cell simulation only at selected points over time can greatly reduce the need for I/O.
+Such particles are often arranged at isolated points, regularly as along lines, in planes or in any other user-defined manner.
+
+Probe particles are usually neutral, non-interacting test particles that are statically placed in the simulation or co-moving with along pre-defined path.
+Self-consistently interacting particles are usually called :ref:`tracer particles <usage-workflows-tracerParticles>`.
+
+Workflow
+""""""""
+
+* ``speciesDefinition.param``: create a species specifically for probes and add ``fieldE`` and ``fieldB`` attributes to it for storing interpolated fields
+
+.. code-block:: cpp
+
+    using ParticleFlagsProbes = bmpl::vector<
+        particlePusher< particles::pusher::Probe >,
+        shape< UsedParticleShape >,
+        interpolation< UsedField2Particle >
+    >;
+
+    using Probes = Particles<
+        bmpl::string< 'p', 'r', 'o', 'b', 'e' >,
+        ParticleFlagsProbes,
+        MakeSeq_t<
+            position< position_pic >,
+            probeB,
+            probeE
+        >
+    >;
+
+and add it to ``VectorAllSpecies``:
+
+.. code-block:: cpp
+
+   using VectorAllSpecies = MakeSeq_t<
+       Probes,
+       // ...
+   >;
+
+* ``density.param``: select in which cell a probe particle shall be placed, e.g. in each 4th cell per direction:
+
+.. code-block:: cpp
+
+   // put probe particles every 4th cell in X, Y(, Z)
+   using ProbeEveryFourthCell = EveryNthCellImpl<
+       mCT::UInt32<
+           4,
+           4,
+           4
+       >
+   >;
+
+* ``particle.param``: initialize the individual probe particles in-cell, e.g. always in the left-lower corner and only one per selected cell
+
+.. code-block:: cpp
+
+   CONST_VECTOR(
+       float_X,
+       3,
+       InCellOffset,
+       /* each x, y, z in-cell position component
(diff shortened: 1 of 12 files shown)


share/picongpu/examples/FoilLCT/include/picongpu/simulation_defines/param/density.param
```diff
@@ -102,5 +102,51 @@ namespace densityProfiles
     /* definition of free formula profile */
     using FlatFoilWithRamp = FreeFormulaImpl< FlatFoilWithRampFunctor >;
 
+    struct ProbeEveryFourthCellFunctor
```
psychocoderHPC: Can we implement it as `ProbeEveryNthCellFunctor`.

```C++
template< uint32_t T_N >
struct ProbeEveryNthCellFunctor;
```

ax3l: sure, I can also make it
```C++
template<
    uint32_t T_Nx, 
    uint32_t T_Ny,
    uint32_t T_Nz
>
struct ProbeEveryNthCellFunctor;
```

But in both cases, it might be too complicated for user code and I should maybe move it to a general particle density "profile" and just use it. What do you think?

psychocoderHPC: Yes a general profile would be a good idea. Please use `CT::vector<...>` instead of three numbers to allow the usage for 2 and 3 dimensional densities.

ax3l:
@n01r @BeyondEspresso @PrometheusPi just wanted to ping you on the new density profile for dopings and the doc strings that come with it

n01r:
Great!

Uh, and initializing particles every *n*th cell! Noice!

include/picongpu/particles/densityProfiles/EveryNthCellImpl.hpp
```diff
@@ -0,0 +1,90 @@
...
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/particles/densityProfiles/EveryNthCellImpl.def"
+#include "picongpu/simulation_defines.hpp"
+
+#include "pmacc/math/Vector.hpp"
```
psychocoderHPC: please use `<>` for the include

include/picongpu/particles/densityProfiles/EveryNthCellImpl.def
```diff
@@ -0,0 +1,55 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "pmacc/static_assert.hpp"
```
psychocoderHPC: please use `<>` for the include

include/picongpu/particles/densityProfiles/EveryNthCellImpl.hpp
```diff
@@ -0,0 +1,90 @@
...
+        {
+            // modulo!
+            auto const isThisCellWithProbe( totalCellOffset % SkipCells::toRT() );
+
+            // is this cell populated with a probe particle?
+            bool const isPopulated(
+                isThisCellWithProbe.productOfComponents() == 0
+            );
+
+            // every how many (volumentric) cells do we set a particle:
```
psychocoderHPC: please use `**` for multi line comments

docs/source/usage/workflows/probeParticles.rst
```diff
@@ -0,0 +1,107 @@
...
+.. code-block:: cpp
+
+   // either all via VectorAllSpecies or just select
+   using FileOutputParticles = MakeSeq_t< Probes >;
+
+Known Limitations
+"""""""""""""""""
+
+* currently, only the electric field :math:`\vec E` and the magnetic field :math:`\vec B` can be recorded
+* we currently do not support time averaging
```
ax3l: Add:

* if the probe particles are dumped in the file output, the instantaneous fields they recorded will be one time step behind the last field update (since our `runOneStep` pushed the particles first and then calls the field solver)

ax3l:
@ComputationalRadiationPhysics/picongpu-maintainers ping, mergable :)

PR #2397, opened 2017-11-20
add support for particle filters

psychocoderHPC:
- PMacc: use particle filters in `ResourceMonitor`, `ConcatListOfFrames` and `CountParticles`
- PIConGPU:
  - add helper class `SpeciesFilter`
  - Prepare all plugins which uses PMacc particle methods to be particle filter ready.

Note: `ConcatListOfFrames` is currently hacky. All usages of alpaka accelerators are filled with fake values. This means that is it currently not possible to use accelerator dependent functions (RNG, ...) with the filter. We need to rewrite `ConcatListOfFrames` as proper alpaka kernel. This is **not** possible currently until we fully switch to alpaka (currently we are running in a mixed mode with native CUDA).

**WARNING** the species names in the output files from HDF5 and ADIOS will be change after this PR. New name are `<speciesName>_<filternName>`. The current used default filter is `all` this means electrons are named `e_all`.

Related to #1288 

# Tests

- [x] write HDF5 and HDF5
- [x] restart ADIOS
- [x] restart HDF5

### DIFF
--- a/include/picongpu/plugins/CountParticles.hpp
+++ b/include/picongpu/plugins/CountParticles.hpp
@@ -24,6 +24,7 @@
 #include <pmacc/mappings/kernel/AreaMapping.hpp>
 
 #include "picongpu/plugins/ISimulationPlugin.hpp"
+#include "picongpu/particles/filter/filter.hpp"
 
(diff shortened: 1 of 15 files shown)


psychocoderHPC:
ready for a review

PR #2399, opened 2017-11-21
bracket test for external libraries includes

psychocoderHPC:
- add test check if external libraries are included with `<>` brackets
- add `hasExtLibIncludeBrackets` to travis code style checks
- fix wrong include brackets in `LocalDensity.kernel` and `MallocMCBuffer.hpp`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,6 +104,23 @@ jobs:
         # Disallow spaces before pre-compiler macros                                #
         #############################################################################
         - test/hasSpaceBeforePrecompiler
+
+        #############################################################################
+        # Enforce angle brackets <...> for includes of external library files       #
+        #############################################################################
+        - test/hasExtLibIncludeBrackets include boost
+        - test/hasExtLibIncludeBrackets include alpaka
+        - test/hasExtLibIncludeBrackets include cupla
+        - test/hasExtLibIncludeBrackets include splash
+        - test/hasExtLibIncludeBrackets include mallocMC
+        - test/hasExtLibIncludeBrackets include/picongpu pmacc
+        - test/hasExtLibIncludeBrackets share/picongpu/examples pmacc
+        - test/hasExtLibIncludeBrackets share/picongpu/examples boost
+        - test/hasExtLibIncludeBrackets share/picongpu/examples alpaka
+        - test/hasExtLibIncludeBrackets share/picongpu/examples cupla
+        - test/hasExtLibIncludeBrackets share/picongpu/examples splash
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -104,6 +104,23 @@ jobs:
         # Disallow spaces before pre-compiler macros                                #
         #############################################################################
         - test/hasSpaceBeforePrecompiler
+
+        #############################################################################
+        # Disallow include brackets `"` to include external library files           #
```
ax3l: Suggestion: since `"` is hard to understand, maybe we could write:
```bash
# Enforce angle brackets `<...>` for includes of external library files
```

test/hasExtLibIncludeBrackets
```diff
@@ -0,0 +1,66 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PIConGPU.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
+# search recursive inside the `$1` folder if a file is using the
+#   wrong include brackets `"` to include extern library files
```
ax3l: maybe add:

```
wrong include brackets `"..."` instead of `<...>` to include ...
```

test/hasExtLibIncludeBrackets
```diff
@@ -0,0 +1,66 @@
...
+    fi
+done
+
+if [ $ok -ne 0 ]
+then
+    echo ""
+    echo "SUMMARY"
+    echo "-------"
+    echo "Run the following command(s) on the above files to fix"
+    echo "the wrong include brackets to include '$libPrefix' files:"
```
ax3l:
```bash
echo "Run the following command(s) on the above files to fix"
echo "the include brackets which are including '$libPrefix' files:"
```

test/hasExtLibIncludeBrackets
```diff
@@ -0,0 +1,66 @@
...
+for i in $(find $folderName \
+    -not -path "./.git/*" \
+    -type f | \
+    grep -P "${pattern}")
+do
+    grep -q "include.*\"$libPrefix\/.*\"" $i
+    if [ $? -eq 0 ]
+    then
+        files+=($i)
+        echo "$i contains wrong brackets '\"' for '$libPrefix' includes!"
```
ax3l:
```bash
echo "$i uses \"...\" instead of <...> for '$libPrefix' includes!"
```

PR #2410, opened 2017-11-28
RNG with reduced state

psychocoderHPC:
- re-add `RNG` XorMin and `MRG32k3aMin` (was emoved with #2226)
- on CPU: the RNG has a fallback to the default alpaka RNG
- PIConGPU: use XorMin by default

- [x] add XorMin (6xint32!)
- [x] Runtime test

CC-ing: @n01r

### DIFF
--- a/include/pmacc/random/methods/MRG32k3aMin.hpp
+++ b/include/pmacc/random/methods/MRG32k3aMin.hpp
@@ -1,0 +1,141 @@
+/* Copyright 2016-2017 Alexander Grund, Rene Widera
+ *
+ * This file is part of PMacc.
+ *
+ * PMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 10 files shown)


include/pmacc/random/methods/MRG32k3aMin.hpp
```diff
@@ -0,0 +1,141 @@
...
+    using MRG32k3aMin = AlpakaRand< T_Acc >;
+#else
+    //! Mersenne-Twister random number generator with a reduced state
+    template< typename T_Acc >
+    class MRG32k3aMin
+    {
+    public:
+        struct StateType
+        {
+            double s1[3];
```
ax3l: memory cost per cell: 6x double (48 byte)

BenjaminW3:
We could add exactly the same directly to alpaka. Only the method name should be something like `MRG32k3aMinIfAvailable`.

psychocoderHPC:
@BenjaminW3 before we add the tine versions to alpaka we need to add the functionality to create uniform distributed numbers.

I will add first all in PMacc and than we can adapt it to alpaka and remove it later in PMacc.

PR #2416, opened 2017-12-04
Piz Daint: ADIOS 1.13.0

ax3l:
This updates the profiles for Piz Daint (CSCS):

- ADIOS 1.11.1 -> 1.13.0
  - enables transforms:
    - zlib (serial) (`--adios.compression zlib`)
    - [blosc](https://github.com/ornladios/ADIOS/pull/135) (with zlib and more; parallel) (e.g. `--adios.compression blosc:threshold=2048,shuffle=bit,lvl=1,threads=10,compressor=zstd`)
- own zlib
- own libpng

The install scripts under https://gist.github.com/ax3l/68cb4caa597df3def9b01640959ea56b are updated accordingly.

### Tests on Piz Daint

- [x] compile foil example
- [x] run foil example (HDF5, ADIOS+ZLIB, ADIOS+BLOSC(zlib))

Note: blosc with 10 theads and zstd was the [fastest & best-compressing option](https://arxiv.org/abs/1706.00522) of the three tested above

ccing @n01r @PrometheusPi @BeyondEspresso @HighIander

### DIFF
--- a/etc/picongpu/pizdaint-cscs/picongpu.profile.example
+++ b/etc/picongpu/pizdaint-cscs/picongpu.profile.example
@@ -18,8 +18,8 @@ export MY_NAME="$(whoami) <$MY_MAIL>"
 #
 # if the wrong environment is loaded we switch to the gnu environment
 # note: this loads gcc/5.3.0 (6.0.4 is the version of the programming env!)
-module li 2>&1 | grep "PrgEnv-cray" > /dev/null
-if [ $? -eq 0 ] ; then
+CRAYENV_FOUND=$(module li 2>&1 | grep "PrgEnv-cray" > /dev/null && { echo 0; } || { echo 1; })
+if [ $CRAYENV_FOUND -eq 0 ]; then
     module swap PrgEnv-cray PrgEnv-gnu/6.0.4
 else
     module load PrgEnv-gnu/6.0.4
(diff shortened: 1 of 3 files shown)


etc/picongpu/pizdaint-cscs/picongpu.profile.example
```diff
@@ -19,6 +19,7 @@ export MY_NAME="$(whoami) <$MY_MAIL>"
 # if the wrong environment is loaded we switch to the gnu environment
 # note: this loads gcc/5.3.0 (6.0.4 is the version of the programming env!)
 module li 2>&1 | grep "PrgEnv-cray" > /dev/null
```
ax3l: maybe we can rewrite this to something like
```bash
CRAYENV_FOUND=$(module li 2>&1 | grep "PrgEnv-cray" > /dev/null && { echo 0; } || { echo 1; })
if [ $CRAYENV_FOUND -eq 0 ]; then
    module swap PrgEnv-cray PrgEnv-gnu/6.0.4
else
```

etc/picongpu/pizdaint-cscs/picongpu.profile.example
```diff
@@ -18,8 +18,8 @@ export MY_NAME="$(whoami) <$MY_MAIL>"
 #
 # if the wrong environment is loaded we switch to the gnu environment
 # note: this loads gcc/5.3.0 (6.0.4 is the version of the programming env!)
-module li 2>&1 | grep "PrgEnv-cray" > /dev/null
-if [ $? -eq 0 ] ; then
+CRAYENV_FOUND=$(module li 2>&1 | grep "PrgEnv-cray" > /dev/null && { echo 0; } || { echo 1; })
```
PrometheusPi: OPTIONAL:
Would it not be more readable if we just execute?:
```bash
module li 2>&1 | grep "PrgEnv-cray" > /dev/null
CRAYENV_FOUND=$?
```

ax3l: the idea here is to avoid depending on `$?` since that means a command needs to fail.

In clean bash scripts, one would set `set -e` at the first executed line, which will prevent that *intermediate failing commands* go unnoticed by a user. Since this script is used in the mighty [*install-all-the-things* script](https://gist.github.com/ax3l/68cb4caa597df3def9b01640959ea56b) you want to see when things fail, e.g. because a URL changed or something did not work as expected.

As an example reference, see: https://github.com/azet/community_bash_style_guide#style-conventions

PrometheusPi: Thanks for the detailed information.

PR #2427, opened 2017-12-10
Docs Plugins: Update PS & Energy Histo

ax3l:
Update the python examples in the manual for the phase space and energy histogram plugin.

### DIFF
--- a/docs/source/usage/plugins/energyHistogram.rst
+++ b/docs/source/usage/plugins/energyHistogram.rst
@@ -65,10 +65,26 @@ You can quickly plot the data in Python with:
    from picongpu.plugins.energy_histogram import EnergyHistogram
    import matplotlib.pyplot as plt
 
+   # load data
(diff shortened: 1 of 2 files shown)


docs/source/usage/plugins/energyHistogram.rst
```diff
@@ -65,10 +65,26 @@ You can quickly plot the data in Python with:
...
+   MeV = 1.e-3  # keV to MeV
+
+   # plotting
+   plt.plot(bins * MeV, counts)
+
+   # range
+   ax = plt.gca()
+   # log scale example
+   # ax.set_yscale('log')
+   # ax.set_ylim([1.e7, 1.e12])
```
@ax3l Do you want to keep this dead code?

ax3l: yes, people always look this part up but it is not very simple to auto-express it when I tried via np.min/max, e.g. for single-bin distributions it will always look odd.

PR #2425, opened 2017-12-08
add filter support to phase space plugin

psychocoderHPC:
- remove `PhaseSpaceMulti`
- update `Cell2Particle` to support filter
  - this also requires an update in ParticleMerger
- update documentation

# Todo

- [x] tests
- [x] update documentation
- [x] merge bug fix #2428 by hand (rebase after it is merged)

PR #2428, opened 2017-12-11
Phase Space Plugin: Fix Weighted Particles

ax3l:
### Phase Space Plugin

Fix for weighted particles with a weighting off the "typical" (nominal) weighting. The plugin calculated, since its introduction in #347 (releases: `beta-rc5`-`0.3.1`), the momentum bin of particles with non-typical weighting wrong: particles with lower weighting were put in proportionally lower momentum bins and with higher weighting than "typical weighting" were put in proportionally higher bins.

## Affected Setups

Entries to phase space bins originating from
- up/downramp particles
- material dopings with non-"typical" weighting
- below-1.0 areas of the normalized density profile

are ill-placed in the phase space image. Momenta of lower-weighting (macro) particles were underestimated, momenta of higher-weighting (macro) particles were overestimated.

## Unaffected setups

This *only* affects a plugin, deriving a 2D phase space on the fly. The general PIC cycle, the physics it solves and any other output of particle data, e.g. ADIOS or HDF5, energy histograms, etc. are NOT affected. If you did e.g. filter the particle data by hand and created a phase space image from yourself the result was correct.

If you did, e.g. a self-injection LWFA setup, an ion acceleration setup where the main phase space signature is coming from particles in the "bulk", a plasma instability setup (KHI, Weibel, etc.) where the density is always nominal "1.0", etc. the error does not affect you.

Also, integrals of the phase-space matrix over the momentum direction will still lead to the correct particle count (just wrong bins were selected, but the particle numbers are correct).

## Demonstrating Setup

The following setup shows a typical TNSA foil setup: an exponential density slope (due to finite laser contrast) before a plateau phase starting at 0.5 microns. The plateau density was set with a free density profile to `0.5` (instead of 1.0) times the `BASE_DENSITY_SI` (all in `density.param`). An initial particle drift of `1 MeV` for electrons (gamma = 2.957; momentum in beta gamma = 2.783) in y direction. In phase space y-py one would expect a straight line at beta gamma.

**Before:**

(instead of a straight line, the momentum-bins are weighted with the density resulting from the lower macro-particle weighting)

ps_broken

**After:**

ps_fix

### To Do

- [ ] after initial review, announce on user-list
- [x] add image demonstrating the issue
- [x] add to 0.3.2 backport candidates
- [ ] please merge this PR *before* merging #2425

### DIFF
--- a/include/picongpu/plugins/PhaseSpace/PhaseSpaceFunctors.hpp
+++ b/include/picongpu/plugins/PhaseSpace/PhaseSpaceFunctors.hpp
@@ -103,7 +103,7 @@ namespace picongpu
             const float_PS particleChargeDensity =
               precisionCast<float_PS>( charge / CELL_VOLUME );
 
-            const float_X rel_bin = (mom_i - axis_p_range.first)
+            const float_X rel_bin = (mom_i / weighting - axis_p_range.first)
                                   / (axis_p_range.second - axis_p_range.first);
             int p_bin = int( rel_bin * float_X(num_pbins) );
 
--- a/include/picongpu/plugins/PhaseSpace/PhaseSpace.tpp
+++ b/include/picongpu/plugins/PhaseSpace/PhaseSpace.tpp
@@ -239,13 +239,16 @@ namespace picongpu
         const float_64 unit = UNIT_CHARGE / UNIT_VOLUME;
 
         /* (momentum) p range: unit is m_species * c
-         *   During the kernels we calculate with a typical MAKRO momentum range
-         *   to avoid over- / underflows. Now for the dump the meta information
+         *   During the kernels we calculate with a typical single/real
+         *   momentum range. Now for the dump the meta information of units
          *   on the p-axis should be scaled to represent single/real particles.
-         *   \see PhaseSpaceMulti::pluginLoad( ) */
-        float_64 pRange_unit = float_64( frame::getMass<typename Species::FrameType>() ) *
-                               float_64( SPEED_OF_LIGHT ) *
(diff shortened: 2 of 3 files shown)


include/picongpu/plugins/PhaseSpace/PhaseSpaceFunctors.hpp
```diff
@@ -99,11 +99,11 @@ namespace picongpu
 
             const uint32_t r_bin    = cellIdx[r_dir];
             const float_X weighting = particle[weighting_];
-            const float_X charge    = attribute::getCharge( weighting,particle );
+            const float_X charge    = attribute::getCharge( 1.0, particle );
             const float_PS particleChargeDensity =
               precisionCast<float_PS>( charge / CELL_VOLUME );
```
ax3l: to do: add `weighting` here!

include/picongpu/plugins/PhaseSpace/PhaseSpaceFunctors.hpp
```diff
@@ -99,11 +99,11 @@ namespace picongpu
 
             const uint32_t r_bin    = cellIdx[r_dir];
             const float_X weighting = particle[weighting_];
-            const float_X charge    = attribute::getCharge( weighting,particle );
+            const float_X charge    = attribute::getCharge( 1.0, particle );
```
ax3l: todo: keep `weighting` here!

PrometheusPi:
@ax3l great catch :+1:

ax3l:
thanks! :)

PR #2403, opened 2017-11-21
IO with filtered particles

psychocoderHPC:
- refactor HDF, ADIOS to allow usage as multi plugin
- refactor interface class and helper for multi plugins
  - `IHelp`: is the new factory to create a `ISlave`
  - `ISlave`: remove factory method
  - `Master`: use `IHelp` fectory method to create a instance of `ISlave`
- `IIOBackend`:
  - inherit from `ISlave`
  - `expandHelp()` is moved into `IHelp`
- refactor `EnergyParticle` to use new interafces
- add documentation for the new option
- add function to split a string ￼…
  - add `splitString` function to `plugins/misc`
  - add function `containsObject()` to check if a container holds an object
  - add function `removeSpaces()` to remove all spaces within a string
  - add collection include file `misc.hpp`

# Review

Please review the PR commit wise!

# Todo
- [x] add Doxygen documentation
- [x] update rst files
- [x] restart tests
- [x] io tests

PR #2430, opened 2017-12-11
add `hdf5.file` to cfg's with IO

psychocoderHPC:
follow up of #2403 (thx @ax3l for the hint that I missed it in the original PR)

add missing required option `hdf5.file` to example cfg's


psychocoderHPC:
I forced pushed some more changes because I forget to save all files in my IDE

PR #2424, opened 2017-12-08
multi plugin energy histogram with filter

psychocoderHPC:
- refactor energy histogram that the usage as multi plugin is possible
- add `--<species>_energyHistogram.filter` (allow to filter particles)
- update documentaion

PR #2431, opened 2017-12-14
Python Tools (PS, Histo): Filter Aware

ax3l:
Add filter awareness to python post-processing tools for phase space and energy histogram.

Follow-up to #2425 and #2424

@codingS3b can you check if these updates work for you? :)

### DIFF
--- a/lib/python/picongpu/plugins/phase_space.py
+++ b/lib/python/picongpu/plugins/phase_space.py
@@ -18,13 +18,16 @@ class PhaseSpaceMeta(object):
     Meta information for data of a phase space iteration.
     """
 
-    def __init__(self, species, ps, shape, extent, dV):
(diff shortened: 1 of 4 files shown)


lib/python/picongpu/plugins/phase_space.py
```diff
@@ -175,7 +188,7 @@ def get_iterations(self, species, ps):
 
         return iterations
 
-    def get(self, species, ps, iteration):
+    def get(self, species, species_filter='all', ps=None, iteration):
```
ax3l: `iteration=None` missing

ax3l:
@codingS3b I think about moving the `iteration` argument to the first argument in `get()` since it is always needed. Would that be ok with you?

codingS3b:
Ok, no problem!

PR #2432, opened 2017-12-15
Clearer conversion functions for Parameter values between UI scale and internal scale

codingS3b:
This pull request adds functions to the `UiParameter` and derived classes for proper (and clearer) conversion of values adjusted by the user in the UI via e.g. sliders and their possibly transformed (linear scaled, or log scaled) 'internal' values which will be passed to PIConGPU.

In the docs, we will from now on refer to the `self.value` of each Parameter object as being on the 'internal scale' and possible slider widgets values (that might be used to change the `self.value` attribute)
that the user is allowed to adjust as being on the 'ui scale'.

Each class deriving from `UiParameter` now needs to override the following functions:
- convert_value_to_internal_scale
- convert_value_from_internal_scale

Besides, the `get_value()` function was renamed into the hopefully clearer `get_value_on_ui_scale()`
because it always did an inverse transformation to produce a value on the ui scale which one might not expect from a regular get() function.

### DIFF
--- a/lib/python/picongpu/input/parameters.py
+++ b/lib/python/picongpu/input/parameters.py
@@ -246,8 +246,8 @@ class UiParameter(Parameter):
         value to internal value.
         Since for UiParameter there is no conversion, it returns
         the self.value.
-        In derived classes however, there will be a reverse trans-
-        formation of the self.value to the scale of the ui range
+        In derived classes however, there will be a reverse
+        transformation of the self.value to the scale of the ui range
         (which e.g. the slider will use for displaying purposes)
 
         Returns
--- a/lib/python/picongpu/input/parameters.py
+++ b/lib/python/picongpu/input/parameters.py
@@ -210,13 +210,51 @@ class UiParameter(Parameter):
         if change["type"] == "change":
             return self.set_value(x=change["new"])
 
-    def get_value(self):
+    def convert_value_from_internal_scale(self, x):
         """
+        Used for value conversion from the parameters internal scale
+        to the ui scale
+        It computes the inverse of the transformation
+        that is used in convert_value_to_internal_scale().
+
+        Since those scales are the same for UiParameter
+        objects, this computes the identity function.
+
+        In derived classes however, there might exist a
+        transformation function to convert between ui scale
+        values and internal parameter scale values
+        (e.g. slider shows values between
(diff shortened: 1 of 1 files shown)


ax3l:
@codingS3b can you please rebase the PR against the latest `dev`? The travis images got updated and we need to re-build our caches (now updated and cleared already).

lib/python/picongpu/input/parameters.py
```diff
@@ -185,13 +210,51 @@ def on_value_change(self, change):
...
+
+        return x
+
+    def get_value_on_ui_scale(self):
         """
+        Function that reverts a possible transformation of ui
+        value to internal value.
+        Since for UiParameter there is no conversion, it returns
+        the self.value.
+        In derived classes however, there will be a reverse trans-
```
ax3l: please avoid manual line-breaks since documentation generators will then "gen-erate" the same hyphen.
Just put the word on the next line :)

ax3l:
Aren't we already trying to get rid of classes such as `LogScaledParameter`?

codingS3b:
In this case, I take it you would like to keep only the `UiParameter` class? 
We would then still have to have a general enough mechanism to allow for such conversions as the `LogScaledParameter` and the `LinearScaledParameter` provide and I'm not sure how we can achieve this in a nice way without having to pass two lambda functions (one for the conversion and one for the inverse conversion).
Do you have an idea in mind there?

lib/python/picongpu/input/parameters.py
codingS3b: Ok, I fixed that.

ax3l:
I might be a bit off since it's a few weeks ago, but I think we decided that scale conversions for ranges and steps besides simple transforms (e.g. normalization to PIConGPU units) should be rolled over to the widget side. In that case we really just need one `UiParameter` class, right? Do you still have the meeting nodes, we can look it up.

codingS3b:
Ok, I did not write that down explicitely, but the file you sent with a sketch of the config file
[2017_11_03_GUI_InputSettings.txt](https://github.com/ComputationalRadiationPhysics/picongpu/files/1571637/2017_11_03_GUI_InputSettings.txt) kind of seemed to imply that.
But could we adress this in another PR since I think it will require some more refactoring...

PR #2439, opened 2017-12-19
particle boundary conditions

### LINKED ISSUE #2433: GuardHandlerCallPlugins usage deprecated
The [picongpu documentation](http://picongpu.readthedocs.io/en/latest/usage/plugins/particleCalorimeter.html?highlight=GuardHandler#detection-of-outgoing-particles) is outdated regarding the use of `GuardHandlerCallPlugins`
```c++
// example for photons
using PIC_Photons = Particles<
    ParticleDescription<
        bmpl::string<'p', 'h'>,
        SuperCellSize,
        DefaultAttributesSeq,
        ParticleFlagsPhotons,
        GuardHandlerCallPlugins
    >
> ;
```

Adding this option activates the plugin member function 
`virtual void onParticleLeave(const std::string&, const int32_t)`
which handles the call of the plugin (currently only `ParticleCalorimeter`) whenever a particles leaves the `CORE + BORDER` region of a simulation.

**offline with @psychocoderHPC**   
Since `ParticleDescription` is not exposed to the user, anymore, this option cannot be used but the code still allocates memory for this feature (see [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/ac7ae616c87f6287a8869c9296621d686929e409/include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp#L105)).
```c++
/* device calorimeter buffer for all particles which have left the simulation volume  */
DBufCalorimeter* dBufLeftParsCalorimeter;
```
By adding an alias and looking for this alias, this feature could be made switchable and we could thus save a lot of device memory.

psychocoderHPC:
- add species attribute `boundaryCondition<>`
- extent particle description with the in flag defined boundary condition
- call method `onParticleLeave` for all plugins by default if `boundaryCondition` is not defined
- remove section that the user needs to activate monitoring of leaving particles via extenting of the species flags

**attention** This PR change the default behavior, now all plugins will be notified if particles leave the global simulation box.

Fixes #2433

# Tests

I run the LWFA example 3000 steps once with calorimeter which monitors the leaving particles and one without. I compared a few samples in the hdf5 (plain) and saw that the energies in the file where the leaving particles are monitored are higher than within the other run. I also checked that the function to monitors the leaving particles is only called if the alias `boundaryCondition` was set.
Any further independent tests are welcome.

CC-ing: @n01r

### DIFF
--- a/docs/source/usage/plugins/particleCalorimeter.rst
+++ b/docs/source/usage/plugins/particleCalorimeter.rst
@@ -72,27 +72,6 @@ Tuning the spatial resolution
 By default, the spatial bin size is chosen by dividing the opening angle by the number of bins for yaw and pitch respectively.
 The bin size can be tuned by customizing the mapping function in ``particleCalorimeter.param``.
 
-Detection of outgoing particles
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-In order to detect particles which are leaving the global simulation volume, the calorimeter plugin has to be triggered whenever this happens.
-This is established by adding the ``GuardHandlerCallPlugins`` template argument to ``ParticleDescription`` (see :ref:`speciesDefinition.param <usage-params-core>`):
-
-.. code:: cpp
-
-   // example for photons
-   using PIC_Photons = Particles<
-       ParticleDescription<
-           bmpl::string<'p', 'h'>,
-           SuperCellSize,
-           DefaultAttributesSeq,
-           ParticleFlagsPhotons,
-           GuardHandlerCallPlugins
-       >
-   > ;
-
-
-Please make sure to have ``picongpu/GuardHandlerCallPlugins.hpp`` included.
 
 Memory Complexity
 ^^^^^^^^^^^^^^^^^
@@ -106,7 +85,6 @@ Host
 """"
 
 as on accelerator.
-
 Output
 ^^^^^^
 
--- a/include/picongpu/particles/Particles.hpp
+++ b/include/picongpu/particles/Particles.hpp
@@ -24,14 +24,21 @@
 
 #include "picongpu/fields/Fields.def"
 #include "picongpu/fields/Fields.hpp"
-#include <pmacc/particles/ParticlesBase.hpp>
-#include <pmacc/particles/memory/buffers/ParticlesBuffer.hpp>
+#include "picongpu/simulationControl/GuardHandlerCallPlugins.hpp"
 #include "picongpu/particles/manipulators/manipulators.def"
-#include <pmacc/particles/ParticleDescription.hpp>
 
 #include <pmacc/memory/dataTypes/Mask.hpp>
 #include <pmacc/mappings/simulation/GridController.hpp>
 #include <pmacc/dataManagement/ISimulationData.hpp>
+#include <pmacc/particles/ParticleDescription.hpp>
+#include <pmacc/particles/ParticlesBase.hpp>
+#include <pmacc/particles/memory/buffers/ParticlesBuffer.hpp>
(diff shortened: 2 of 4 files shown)


n01r:
WOOHOO :sparkles: :tada:

docs/source/usage/plugins/particleCalorimeter.rst
```diff
@@ -75,19 +75,18 @@ Detection of outgoing particles
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 In order to detect particles which are leaving the global simulation volume, the calorimeter plugin has to be triggered whenever this happens.
-This is established by adding the ``GuardHandlerCallPlugins`` template argument to ``ParticleDescription`` (see :ref:`speciesDefinition.param <usage-params-core>`):
+This is established by adding the alias ``boundaryCondition< GuardHandlerCallPlugins >`` to the particle species flags (see :ref:`speciesDefinition.param <usage-params-core>`):
```
n01r: I would maybe call it `onParticleLeave` / `whenLeavingSimVolume` or sth of the sort - just like the optional plugin member function it activates.
The name `boundaryCondition` may be a bit misleading.

include/picongpu/simulation_defines/param/speciesAttributes.param
```diff
@@ -254,4 +254,15 @@ namespace picongpu
...
+    /** alias to specify the boundary condition for particles
+     *
+     * The default behavior if this alias is not given to a species is that the
+     * particles which leave the global simulation box were deleted.
+     *
+     * other options:
+     *   GuardHandlerCallPlugins - notify all plugins to handle leaving particles
+     *   and delete the particles after wards.
+     */
+    alias( boundaryCondition );
```
n01r: Same here: I would suggest to rename this to something more pictoral like:
- `onParticleLeave`
- `onLeavingSimVolume`
- `whenLeavingSimVolume`

include/picongpu/simulation_defines/param/speciesAttributes.param
```diff
@@ -254,4 +254,15 @@ namespace picongpu
      */
     alias( exchangeMemCfg );
 
+    /** alias to specify the boundary condition for particles
+     *
+     * The default behavior if this alias is not given to a species is that the
+     * particles which leave the global simulation box were deleted.
```
ax3l: where

ax3l: Add sentence: `This also notifies all plugins that can handle leaving particles.`

docs/source/usage/plugins/particleCalorimeter.rst
ax3l: nope, we can make this the beginning of the initial boundary condition handling beyond absorbing/periodic, but the interface needs to be refactored.

By default, we will make absorbing boundary conditions call the plugin guard handler.

include/picongpu/particles/Particles.hpp
```diff
@@ -72,7 +79,29 @@ class Particles : public ParticlesBase<
...
+            typename pmacc::traits::Resolve<
+                typename GetKeyFromAlias<
+                    T_Flags,
+                    boundaryCondition< >
+                >::type
+            >::type,
+            // fallback if the species has not defined the alias boundaryCondition
+            pmacc::HandleGuardRegion<
+                pmacc::particles::policies::ExchangeParticles,
+                GuardHandlerCallPlugins
```
ax3l: please let us rename this class in a follow-up to `CallPluginsAndDeleteParticles` or so

PR #2447, opened 2018-01-02
PIConGPU use tiny RNG

psychocoderHPC:
- add class `RngWrapper`
- refactor template signature for particle functors/manipulator/filter with RNG
- update documentation
- MySimulation: enable always the RNG which holds the states over the full simulation
- update default param files

# Tests

- [x] default LWFA
  - LWFA is running two times faster due to faster initialization during the slide
- [x] default KHI
- [x] rebase against #2451 to solve the compile issues

### DIFF
--- a/include/picongpu/particles/filter/generic/FreeRng.def
+++ b/include/picongpu/particles/filter/generic/FreeRng.def
@@ -38,8 +38,7 @@ namespace generic
      *
      *
      * @tparam T_Functor user defined unary functor
-     * @tparam T_Distribution random number distribution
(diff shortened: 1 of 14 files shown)


ax3l:
@psychocoderHPC the double-precision compile tests are not happy with your change set yet ;)

psychocoderHPC:
:-(  The generator is currently not implemented for double. I will implement it for double too.

psychocoderHPC:
rebase against #2451 is needed to solve the compile issues

ax3l:
@psychocoderHPC pls rebase :)

PR #2448, opened 2018-01-03
License Header: Update 2018

ax3l:
Updating years in header files is cumbersome, so we update all files to include 2018 at once.

Updated via

``` bash
grep -iR "Copyright 20" . | awk -F: '{print $1}' | \
  xargs -n1 -P1 -I{} sed -i 's/\(Copyright 20..\) /\1-2018 /g' {}
grep -iR "Copyright 20" . | awk -F: '{print $1}' | \
  xargs -n1 -P1 -I{} sed -i 's/\(Copyright 20..-201\). /\18 /g' {}
```

and commited via

``` bash
git checkout thirdParty/

  git commit -a -S
```

Happy new year! :) :star2:

### DIFF
--- a/bin/cuda_memtest.sh
+++ b/bin/cuda_memtest.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Copyright 2013-2017  Rene Widera
+# Copyright 2013-2018  Rene Widera
 #
 # This file is part of PIConGPU.
 #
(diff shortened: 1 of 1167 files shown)


PrometheusPi:
@ax3l The Travis CI test failed, but it looks like an issue with the test system (apt-get did not work properly) and not with your code change.

ax3l:
@PrometheusPi seems to have re-queued and passed

PrometheusPi:
@ax3l Yes, @psychocoderHPC required the failed test.

PR #2449, opened 2018-01-03
fix RNG cpu compile

psychocoderHPC:
The current dev is not compiling with the accelerator `omp2b`.

- fix `GetRandomType` trait (add missing `template` keyword)


PrometheusPi:
@psychocoderHPC Please comment on *adding your name the the license header*. This is a minimal change, so no name is fine with me.  I will merge as soon as all compile tests pass.

psychocoderHPC:
@PrometheusPi It is not necessary to add my name because the change is to small.

PR #2451, opened 2018-01-04
PMacc RNG 64bit support

psychocoderHPC:
- fix minor documentation bugs
- add support for 64Bit floating point
  - normal distributed numbers
  - uniform distributed numbers
- add 64bit integer support to generate uniform distributed random numbers
- fix usage of `float` in the generic `MullerBox`
- RNG methods: extent interface with to generate a 64bit integer random number
- rename the range `ExcludeOne<...>::Use24Bit` to `ExcludeOne<...>::Reduced`

I fixed a documentation to the normal distributes number ranges and some hard coded usages of `float`. I have not added the label **BUG** because both are very tiny and has no real effect to the users.

# Tests

- [x] run PMacc RNG unit test

### DIFF
--- a/include/pmacc/random/distributions/uniform/Range.hpp
+++ b/include/pmacc/random/distributions/uniform/Range.hpp
@@ -41,7 +41,7 @@ namespace uniform
     struct ExcludeZero
     {};
 
-    /**  floating point number in the range (0,1]
+    /**  floating point number in the range [0,1)
      *
      * @tparam T_Type type of the result
      */
@@ -49,11 +49,14 @@ namespace uniform
     struct ExcludeOne
     {
 
-        /** Reduce the random range to 2^24.
+        /** Reduce the random range.
+         *
+         * range for `float` is `2^24`
+         * range for double is `2^53`
          *
          * Creates intervals with the width of epsilon/2.
          */
-        struct Use24Bit
+        struct Reduced
         {};
 
         /** Loops until a random value inside the defined range is created.
--- a/include/pmacc/random/distributions/uniform/Range.hpp
+++ b/include/pmacc/random/distributions/uniform/Range.hpp
@@ -49,19 +49,21 @@ namespace uniform
     struct ExcludeOne
     {
 
-        /** Reduce the random range.
+        /** Reduce the random range
          *
-         * range for `float` is `2^24`
-         * range for double is `2^53`
+         * number of unique random numbers for
+         *   - `float` is `2^24`
+         *   - `double` is `2^53`
          *
          * Creates intervals with the width of epsilon/2.
          */
         struct Reduced
         {};
 
-        /** Loops until a random value inside the defined range is created.
+        /** Loops until a random value inside the defined range is created
          *
          * The runtime of this method is not deterministic.
+         * @warning zero is excluded which results in a range (0,1)
          */
         struct Repeat
         {};
@@ -74,7 +76,6 @@ namespace uniform
         {};
 
     };
-
 
 }  // namespace uniform
 }  // namespace distributions
--- a/include/pmacc/random/distributions/uniform/Uniform_Integral64Bit.hpp
+++ b/include/pmacc/random/distributions/uniform/Uniform_Integral64Bit.hpp
@@ -22,14 +22,9 @@
 #pragma once
 
 #include "pmacc/types.hpp"
-#include "pmacc/random/distributions/Normal.hpp"
-#include "pmacc/random/distributions/misc/MullerBox.hpp"
-#include "pmacc/random/methods/XorMin.hpp"
-#include "pmacc/random/methods/MRG32k3aMin.hpp"
 #include "pmacc/random/distributions/Uniform.hpp"
-#include "pmacc/algorithms/math.hpp"
 
-#include <type_traits>
+#include <boost/type_traits.hpp>
 
 
 namespace pmacc
@@ -40,39 +35,43 @@ namespace distributions
 {
 namespace detail
 {
-    //! specialization for XorMin
-    template<
-        typename T_Acc
-    >
-    struct Normal<
-        float,
-        methods::XorMin< T_Acc >,
-        void
-    > :
-        public MullerBox<
-            float,
-            methods::XorMin< T_Acc >
-        >
-    {
 
+    /**
+     * Returns a random, uniformly distributed (up to) 64 bit integral value
+     */
+    template<
+        typename T_Type,
+        class T_RNGMethod
+    >
+    class Uniform<
+        T_Type,
+        T_RNGMethod,
+        typename bmpl::if_c<
+            boost::is_integral< T_Type >::value && sizeof( T_Type ) == 8,
+            void,
+            T_Type
+        >::type
+    >
+    {
+        typedef T_RNGMethod RNGMethod;
+        typedef typename RNGMethod::StateType StateType;
+    public:
+        typedef T_Type result_type;
+
+        template< typename T_Acc >
+        DINLINE result_type
+        operator()(
+            T_Acc const & acc,
+            StateType& state
+        )
+        {
+            return static_cast< result_type >( RNGMethod().get64Bits(
+                acc,
+                state
+            ) );
+        }
     };
 
-    //! specialization for MRG32k3aMin
-    template<
-        typename T_Acc
-    >
-    struct Normal<
-        float,
-        methods::MRG32k3aMin< T_Acc >,
-        void
-    > :
-        public MullerBox<
-            float,
-            methods::MRG32k3aMin< T_Acc >
-        >
-    {
-
-    };
 }  // namespace detail
 }  // namespace distributions
 }  // namespace random
--- a/include/pmacc/random/distributions/uniform/Uniform_double.hpp
+++ b/include/pmacc/random/distributions/uniform/Uniform_double.hpp
@@ -34,13 +34,13 @@ namespace distributions
 namespace detail
 {
 
-    /** Returns a random float value uniformly distributed in (0,1]
+    /** Returns a random double value uniformly distributed in (0,1]
      *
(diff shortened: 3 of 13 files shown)


include/pmacc/random/methods/XorMin.hpp
```diff
@@ -113,6 +113,19 @@ namespace methods
...
 
+        DINLINE uint64_t
+        get64Bits(
+            T_Acc const & acc,
+            StateType& state
+        ) const
+        {
+            // two 32bit values are packed into a 64bit value
+            uint64_t result = get32Bits( acc, state);
+            result <<= 31;
```
PrometheusPi: First of all, shouldn't this be a 32 for unsigned integers?

I think this only works for 32 bits - used on a 64 bit unsigned int it fails (for me) resulting in loss of all bits beyond the 32 bits.

Example code:
```c
#include<stdio.h>
#include<stdlib.h>
#include <string.h>

int main()
{
  unsigned int init = 0;
  unsigned int init2 = 1;
  printf("binaer: %016x\n",  init);

  init = ~init;
  printf("binaer: %016x\n",  init);

  long unsigned int result = init;
  printf("long unsigned int:  %2d bit\n", sizeof(result)*8);
  printf("binaer: %016x\n",  result);

  unsigned int i = 0;
  for(i=0; i<34; i++)
    printf("%d: %016x\n", i, result<<i);

  result <<= 31;
  printf("binaer: %016x\n",  result);

  result ^= init2;
  printf("binaer: %016x\n",  result);
  return 0;
}
```

results in:
```
binaer: 0000000000000000
binaer: 00000000ffffffff
long unsigned int:  64 bit
binaer: 00000000ffffffff
0: 00000000ffffffff
1: 00000000fffffffe
2: 00000000fffffffc
3: 00000000fffffff8
4: 00000000fffffff0
5: 00000000ffffffe0
6: 00000000ffffffc0
7: 00000000ffffff80
8: 00000000ffffff00
9: 00000000fffffe00
10: 00000000fffffc00
11: 00000000fffff800
12: 00000000fffff000
13: 00000000ffffe000
14: 00000000ffffc000
15: 00000000ffff8000
16: 00000000ffff0000
17: 00000000fffe0000
18: 00000000fffc0000
19: 00000000fff80000
20: 00000000fff00000
21: 00000000ffe00000
22: 00000000ffc00000
23: 00000000ff800000
24: 00000000ff000000
25: 00000000fe000000
26: 00000000fc000000
27: 00000000f8000000
28: 00000000f0000000
29: 00000000e0000000
30: 00000000c0000000
31: 0000000080000000
32: 0000000000000000
33: 0000000000000000
binaer: 0000000080000000
binaer: 0000000080000001
```

psychocoderHPC: yes my shift is wrong. I will fix it and validate the rang again.

PrometheusPi: *offline* discussion with @psychocoderHPC: it should be shifted by 32 bits (occurs in several places)

psychocoderHPC:
@PrometheusPi I fixed the shift and validated the RNG again

PrometheusPi:
@psychocoderHPC Great - I will habe a look at the code as soon as possible.

PrometheusPi:
@psychocoderHPC Would it be possible to squash your fix to the respective commit, not just to the last commit? This avoids seeing the wrong shifts when going through the commits individually. (If this request is too late, that is fine too.)

psychocoderHPC:
I cleaned the commits and force pushed the branch again

include/pmacc/random/distributions/uniform/Range.hpp
```diff
@@ -41,19 +41,22 @@ namespace uniform
...
+    /**  floating point number in the range [0,1)
      *
      * @tparam T_Type type of the result
      */
     template<typename T_Type>
     struct ExcludeOne
     {
 
-        /** Reduce the random range to 2^24.
+        /** Reduce the random range.
```
PrometheusPi: OPTIONAL:
replace point by colon

include/pmacc/random/distributions/uniform/Uniform_double.hpp
```diff
@@ -0,0 +1,177 @@
...
+namespace random
+{
+namespace distributions
+{
+namespace detail
+{
+
+    /** Returns a random double value uniformly distributed in (0,1]
+     *
+     * The smallest created value is `2^-65` (~ `2.710505431213761*10^-20`)
```
PrometheusPi: QUESTION:
is this distribution uniform in the possible numbers of float values or quasi-uniform the domain of real numbers?

include/pmacc/random/distributions/uniform/Uniform_double.hpp
```diff
@@ -0,0 +1,177 @@
...
+            T_Acc const & acc,
+            StateType& state
+        ) const
+        {
+            double const value2pow64Inv = 5.421010862427522e-20;
+            uint64_t const random = RNGMethod().get64Bits(
+                acc,
+                state
+            );
+            return static_cast< double >( random ) * value2pow64Inv +
```
PrometheusPi: ANSWER to previous questions and NEW QUESTION:
It is quasi-uniform in the real numbers **but** what is the result of `0.0 + ( value2pow64Inv / 2.0 )` - `0` or `value2pow64Inv`? It should be `0.0`.

PrometheusPi: Okay - all fine - this is **Exclude Zero**.

include/pmacc/random/distributions/uniform/Uniform_double.hpp
```diff
@@ -0,0 +1,177 @@
...
+
+        template< typename T_Acc >
+        DINLINE double
+        operator()(
+            T_Acc const & acc,
+            StateType& state
+        ) const
+        {
+            double const value2pow53Inv = 1.1102230246251565e-16;
+            double const randomValue53Bit = RNGMethod().get64Bits( acc, state ) >> 11;
```
PrometheusPi: COMMENT:
Thanks for the fix.

include/pmacc/random/distributions/uniform/Uniform_double.hpp
```diff
@@ -0,0 +1,177 @@
...
+            return randomValue53Bit * value2pow53Inv;
+        }
+    };
+
+    /** Returns a random double value uniformly distributed in [0,1)
```
PrometheusPi: QUESTION:
For what do we need this function with a **uniform** distribution. This brute force approach works great for arbitrary distributions.

include/pmacc/random/distributions/uniform/Range.hpp
```diff
@@ -41,19 +41,22 @@ namespace uniform
     struct ExcludeZero
     {};
 
-    /**  floating point number in the range (0,1]
+    /**  floating point number in the range [0,1)
```
psychocoderHPC: range is wrong for Repeat

include/pmacc/random/distributions/uniform/Uniform_double.hpp
psychocoderHPC: It makes only sense for uniform distribution because the ranges for the normal distribution differ.

include/pmacc/random/distributions/uniform/Uniform_double.hpp
psychocoderHPC: This creates random numbers without equal sized bins. I think this is what you mean with real numbers.

include/pmacc/random/distributions/uniform/Range.hpp
```diff
@@ -41,24 +41,28 @@ namespace uniform
...
      * @tparam T_Type type of the result
      */
     template<typename T_Type>
     struct ExcludeOne
     {
 
-        /** Reduce the random range to 2^24.
+        /** Reduce the random range
+         *
+         * range for `float` is `2^24`
```
ax3l: - remove `` from float or add to double?
- 2^24 is the no. of uniquely different floats? (This is not really the "range".)

include/pmacc/random/distributions/uniform/Uniform_Integral64Bit.hpp
```diff
@@ -0,0 +1,78 @@
...
+     */
+    template<
+        typename T_Type,
+        class T_RNGMethod
+    >
+    class Uniform<
+        T_Type,
+        T_RNGMethod,
+        typename bmpl::if_c<
+            boost::is_integral< T_Type >::value && sizeof( T_Type ) == 8,
```
ax3l: why not `sizeof( T_Type ) == sizeof( double )`?

include/pmacc/random/distributions/uniform/Uniform_double.hpp
PrometheusPi: Here, "normal" means not normal in the sense of Gaussian probability distribution, but in the sense of standard [0,1) or (nvidia) (0,1] implementation - and thus means (0,1). 
Thanks for fixing the comment.

include/pmacc/random/distributions/uniform/Uniform_double.hpp
PrometheusPi: Yes, the bins are not equal and thus their probability is not equal too - however both combined lead to a uniform distribution and that s fine - see comments below

include/pmacc/random/distributions/uniform/Uniform_Integral64Bit.hpp
psychocoderHPC: because 64bit are 8 byte

ax3l: yes but you can write instead of `8` `sizeof( uint64_t )` ...?

PR #2452, opened 2018-01-08
extended period syntax

psychocoderHPC:
- redistribute the function `splitString()` from PIConGPU within PMacc
-  add helper function `cointainsStep()`
- add parser function `toTimeSlice()`
- add class TimeSlice
- refactore `PluginConnector`
  - change internal stored type
  - change interface of `setNotificationPeriod()`, change type of period
- use new syntax for checkpoint period definition
- add time slice feature to all PIConGPU plugins
- add documentation

# Tests

- [x] create checkpoints `--checkpoint.period 50,10:12:1`
- [x] phase space plugin with `--e_phaseSpace.period 0:50:10,::`

PR #2455, opened 2018-01-10
ISAAC: unify the usage of period

psychocoderHPC:
follow up of #2452

Unify the meaning of `isaac.period` with other plugins.

- define `notifyPeriod` as `std::string`
- map the command line option `.period` to `notifyPeriod` instead of `render_interval`

### DIFF
--- a/include/picongpu/plugins/IsaacPlugin.hpp
+++ b/include/picongpu/plugins/IsaacPlugin.hpp
@@ -236,7 +236,7 @@ public:
         visualization(nullptr),
         cellDescription(nullptr),
         movingWindow(false),
-        render_interval(0),
+        render_interval(1),
         step(0),
         drawing_time(0),
         cell_count(0),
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
@theZiz It is still possible to change the render interval during the runtime via the GUI. Period is an string to allow the user to define multiple intervals where the plugin is active (notified) e.g. .period 0:50:10,:: see #2452.
To run the simulation equal to the current dev implementation you need to set `isaac.period 1`.

theZiz:
So, the interface is not unified... In fact the interface is ignored now for isaac. 🤣 Why not setting `render_interval = atoi(notifyPeriod.c_str())` (or similar)? This does not support the new `begin:end:period` syntax and of course not a comma separated list, but at least the old behavior is accomplished. 😉

psychocoderHPC:
No period is not ignored. The new behavior is that `render_interval` is only handled within the user configured `period`

PR #2459, opened 2018-01-15
fix clang device compile

psychocoderHPC:
- remove `std::declval` with `alpaka::core::declval`
- use C++ static assert for clang
- defer assert if the compiler is allowed to evaluate the expression even if it is never used

# Tests

- [x] compile with clang 7.0.0

### DIFF
--- a/include/picongpu/particles/flylite/IFlyLite.hpp
+++ b/include/picongpu/particles/flylite/IFlyLite.hpp
@@ -67,9 +67,16 @@ namespace flylite
         )
         {
             boost::ignore_unused( ionSpeciesName, currentStep );
-
+            /* The compiler is allowed to evaluate an expression those not depends on a template parameter
+             * even if the class is never instantiated. In that case static assert is always
+             * evaluated (e.g. with clang), this results in an error if the condition is false.
+             * http://www.boost.org/doc/libs/1_60_0/doc/html/boost_staticassert.html
+             *
+             * A workaround is to add a template dependency to the expression.
+             * `sizeof(ANY_TYPE) != 0` is always true and defers the evaluation.
+             */
             PMACC_STATIC_ASSERT_MSG(
-                false,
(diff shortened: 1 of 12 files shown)


include/picongpu/particles/flylite/IFlyLite.hpp
```diff
@@ -67,9 +67,16 @@ namespace flylite
...
+             * even if the class is never instantiated. In that case static assert is always
+             * evaluated (e.g. with clang), this results in an error if the condition is false.
+             * http://www.boost.org/doc/libs/1_60_0/doc/html/boost_staticassert.html
+             *
+             * A workaround is to add a template dependency to the expression.
+             * `sizeof(ANY_TYPE) != 0` is always true and defers the evaluation.
+             */
             PMACC_STATIC_ASSERT_MSG(
-                false,
+                false && sizeof(T_IonSpecies) != 0,
```
BenjaminW3: You could also use `alpaka::meta::DependentFalseType<T_IonSpecies>`

psychocoderHPC: good to know, I will use in in a separate PR because we have used a few of this code constructs in PMacc and PIConGPU.

include/pmacc/static_assert.hpp
```diff
@@ -45,7 +45,7 @@ namespace pmacc
  * @param pmacc_unique_id pre compiler unique id
  * @param pmacc_typeInfo a type that is shown in error message
  */
-#if ( PMACC_CUDA_COMPILER_CLANG == 1 )
+#if BOOST_LANG_CUDA && BOOST_COMP_CLANG_CUDA
```
ax3l: @BenjaminW3 I think alpaka should either set this define with an `ALPAKA_` prefix or we move it upstream into boost for real

PR #2465, opened 2018-01-22
EnergyHistogram: Remove Detector Filter

ax3l:
The filter for energy histograms can now be implemented with the more generic particle filter concept.

This removes the old "distance to detector" and opening angle implementation and adds an example how to limit the detector acceptance.

### DIFF
--- a/docs/source/usage/plugins/energyHistogram.rst
+++ b/docs/source/usage/plugins/energyHistogram.rst
@@ -31,27 +31,28 @@ For example, to limit to particles within a cone with an opening angle of five d
                T_Particle const & particle
            )
            {
-               float3_X const mom = particle[ momentum_ ];
-
-               /* place detector in y direction, "infinite distance" to target,
-                * and five degree opening angle
-                */
-               float_X const openingAngle = 5.0 * PI / 180.;
-               float_X const dotP = mom.y() / math::abs(mom);
-               float_X const degForw = math::acos( dotP );
-
                bool result = false;
-               if( degForw <= openingAngle / 2.0 )
-                   if( degForw >= openingAngle / -2.0 )
+               float3_X const mom = particle[ momentum_ ];
+               float_X const absMom = math::abs( mom );
+
+               if( absMom > float_X( 0. ) )
+               {
+                   /* place detector in y direction, "infinite distance" to target,
+                    * and five degree opening angle
+                    */
+                   constexpr float_X openingAngle = 5.0 * PI / 180.;
+                   float_X const dotP = mom.y() / absMom;
+                   float_X const degForw = math::acos( dotP );
+
+                   if( math::abs( degForw ) <= openingAngle * float_X( 0.5 ) )
                        result = true;
+               }
                return result;
            }
        };
        using ParticlesForwardPinhole = generic::Free<
           FunctorParticlesForwardPinhole
        >;
-
-       // ...
    }
    }
    }
--- a/docs/source/usage/plugins/energyHistogram.rst
+++ b/docs/source/usage/plugins/energyHistogram.rst
@@ -3,10 +3,67 @@
 Energy Histogram
 ----------------
 
-This plugin computes the energy histogram particle species simulated with PIConGPU.
-The energy binning can be set up using command line parameters.
-A *virtual detector* can be located in the y direction of the simulation to count only particles that would reach such a detector in an experimental setup.
-If the detector setup is not specified, all particles of a species are considered in the histogram. 
+This plugin computes the energy histogram (spectrum) of a selected particle species and stores it to plain text files.
+The acceptance of particles for counting in the energy histogram can be adjusted, e.g. to model the limited acceptance of a realistic spectrometer.
+
+.param file
+^^^^^^^^^^^
+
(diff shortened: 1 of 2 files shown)


docs/source/usage/plugins/energyHistogram.rst
```diff
@@ -3,10 +3,67 @@
...
+               T_Particle const & particle
+           )
+           {
+               float3_X const mom = particle[ momentum_ ];
+
+               /* place detector in y direction, "infinite distance" to target,
+                * and five degree opening angle
+                */
+               float_X const openingAngle = 5.0 * PI / 180.;
+               float_X const dotP = mom.y() / math::abs(mom);
```
PrometheusPi: REQUIRED / QUESTION:
How do you prevent NaNs for zero momentum values?

docs/source/usage/plugins/energyHistogram.rst
```diff
@@ -3,10 +3,67 @@
...
+               float_X const degForw = math::acos( dotP );
+
+               bool result = false;
+               if( degForw <= openingAngle / 2.0 )
+                   if( degForw >= openingAngle / -2.0 )
```
PrometheusPi: OPTIONAL:
Could be simplified to 
```
if( math::abs(degForw) <= 0.5 * openingAngle  )
    result = true;
```

docs/source/usage/plugins/energyHistogram.rst
```diff
@@ -3,10 +3,67 @@
...
+                       result = true;
+               return result;
+           }
+       };
+       using ParticlesForwardPinhole = generic::Free<
+          FunctorParticlesForwardPinhole
+       >;
+
+       // ...
```
Why do you need this suggestion for more code?

include/picongpu/plugins/BinEnergyParticles.hpp
```diff
@@ -218,63 +211,49 @@ struct KernelBinEnergyParticles
...
-                                 */
-                                float_X const normedWeighting = weighting /
-                                    float_X( particles::TYPICAL_NUM_PARTICLES_PER_MACROPARTICLE );
-                                atomicAdd(
-                                    &( shBin[ binNumber ] ),
-                                    normedWeighting,
-                                    ::alpaka::hierarchy::Threads{}
-                                );
-                            }
+                            /*!\todo: we can't use 64bit type on this place (NVIDIA BUG?)
```
PrometheusPi: QUESTION:
is this still valid?

include/picongpu/plugins/BinEnergyParticles.hpp
```diff
@@ -218,63 +211,49 @@ struct KernelBinEnergyParticles
...
+                             * COMPILER ERROR: ptxas /tmp/tmpxft_00005da6_00000000-2_main.ptx, line 4246; error   : Global state space expected for instruction 'atom'
+                             * I think this is a problem with extern shared mem and atmic (only on TESLA)
+                             * NEXT BUG: don't do uint32_t w=__float2uint_rn(weighting); and use w for atomic, this create wrong results
+                             *
+                             * uses a normed float weighting to avoid a overflow of the floating point result
```
PrometheusPi: REQUIRED:
typo: `an overflow`

docs/source/usage/plugins/energyHistogram.rst
ax3l: right, that's what we usually add. I will add an if clause and ignore resting particles!

docs/source/usage/plugins/energyHistogram.rst
ax3l: just because there might be more and is more, e.g. at least the list below. but I can also skip that.

psychocoderHPC: I think we should skip `/ ...` because we will only highlight the example here and not that there can be code after it.

docs/source/usage/plugins/energyHistogram.rst
```diff
@@ -3,10 +3,68 @@
...
+               if( absMom > float_X( 0. ) )
+               {
+                   /* place detector in y direction, "infinite distance" to target,
+                    * and five degree opening angle
+                    */
+                   float_X const openingAngle = 5.0 * PI / 180.;
+                   float_X const dotP = mom.y() / absMom;
+                   float_X const degForw = math::acos( dotP );
+
+                   if( math::abs( degForw ) <= openingAngle * 0.5 )
```
psychocoderHPC: please write `float_X( 0.5 )`

docs/source/usage/plugins/energyHistogram.rst
```diff
@@ -3,10 +3,68 @@
...
+               bool result = false;
+               float3_X const mom = particle[ momentum_ ];
+               float_X const absMom = math::abs( mom );
+
+               if( absMom > float_X( 0. ) )
+               {
+                   /* place detector in y direction, "infinite distance" to target,
+                    * and five degree opening angle
+                    */
+                   float_X const openingAngle = 5.0 * PI / 180.;
```
psychocoderHPC: is double precision needed for this calculation? If not please cast the values to `float_X`

ax3l: that calc happens at CT anyway

PR #2466, opened 2018-01-22
Param Docs: Title Only

ax3l:
In the section that introduces the usage of .param files, only the titles of the files shall be shown. This is currently not possible, see https://github.com/michaeljones/breathe/issues/318

This PR applies a work-around for HTML to hide additional doxygen information (does not fix the underlying issue and is issue still exists in pdf & epub).

### DIFF
--- a/docs/source/_static/hide_chapter_titles.css
+++ b/docs/source/_static/hide_chapter_titles.css
@@ -1,3 +1,25 @@
+/* front page: hide chapter titles
+ * needed for consistent HTML-PDF-EPUB chapters
+ */
 #how-to-read-this-document.section div.section {
     display:none;
 }
+
+/* parameter files: only show short intro of title
+ * and description in HTML.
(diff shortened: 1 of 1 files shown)


docs/source/_static/hide_chapter_titles.css
```diff
@@ -1,3 +1,25 @@
+/* front page: hide chapter titles
+ * needed for consistent HTML-PDF-EPUB chapters
+ */
 #how-to-read-this-document.section div.section {
     display:none;
 }
+
+/* parameter files: only show short intro of title
+ * and description in HTML.
+ * Work-Around for https://github.com/michaeljones/breathe/issues/318
```
PrometheusPi: QUESTION:
Is this a workaround by @michaeljones then use `by` instead of `for`.

PrometheusPi: read the issue - `for` is perfectly fine

PR #2475, opened 2018-01-25
pic-build: CMake multi-args

ax3l:
CMake multi-args to `pic-build -c "... ..."` were not forwarded properly to `pic-configure` and `cmake`.

The example `pic-build -c "-DPIC_VERBOSE=21 -DCMAKE_BUILD_TYPE=Debug"` did not work, the same way passing a non-standard generator with `-c "-G Ninja -D..."` failed.

@sbastrakov you were right, this is actually a problem in the script forwarding multi-args for cmake with spaces in general. This also affects linux systems and should fix your generator issue on Visual Studio as well. Thanks for the report!

### DIFF
--- a/pic-build
+++ b/pic-build
@@ -45,7 +45,7 @@ help()
 }
 
 # save cmd line args
-cmd_line_args="$@"
+cmd_line_args=$@
 
 # show help
 while [[ $# -gt 0 ]] ; do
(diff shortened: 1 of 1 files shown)


sbastrakov:
Sorry, did not notice before. Now -c parameter of pic-build must be last, otherwise later parameters are also mistakenly passed to cmake, so e.g. `./pic-build -b omp2b -c "..."` works fine, while `./pic-build -c "..." -b omp2b` will pass this -b to cmake (and not interpret it as backend), which is wrong.

PR #2480, opened 2018-01-26
fix #2477: adjust LWFA setup for 8GPUs

### LINKED ISSUE #2477: Problem with moving window
I used the current dev branch 5e781594a02322f855eb181fbd573b623caa8279. The modules are 
- gcc/4.9.2
- cuda/8.0
- cmake/3.7.2 
- boost/1.62.0
- openmpi/1.8.6.kepler.cuda80
- pngwriter/0.7.0
- hdf5-parallel/1.8.15
- libsplash/1.6.0

I follow the [basics](http://picongpu.readthedocs.io/en/latest/usage/basics.html) guideline. The only difference was the usage of [/etc/picongpu/0008gpus.cfg](https://github.com/ComputationalRadiationPhysics/picongpu/blob/master/examples/LaserWakefield/submit/0008gpus.cfg) instead of [/etc/picongpu/0016gpus.cfg](https://github.com/ComputationalRadiationPhysics/picongpu/blob/master/examples/LaserWakefield/submit/0016gpus.cfg). 
The problem with this configuration is, the outputted pngs doesn't look right. The matter disappear to the left of the image. If I remove the option -m in **TBG_movingWindow** the output looks like it should. I think there is a bug in the moving window.

PrometheusPi:
This pull request resolves issue #2477 by extending the number of cells in y direction by 50% (+256 cells), pushing the slide point a bit farther, adjusting the GPU distribution and shortening the laser init without changing the laser duration. 

Now, moving window simulations with 8 GPUs produce meaningful result.

### DIFF
--- a/share/picongpu/examples/LaserWakefield/etc/picongpu/0008gpus.cfg
+++ b/share/picongpu/examples/LaserWakefield/etc/picongpu/0008gpus.cfg
@@ -1,4 +1,5 @@
-# Copyright 2013-2018 Axel Huebl, Rene Widera, Felix Schmitt
+# Copyright 2013-2018 Axel Huebl, Rene Widera, Felix Schmitt,
+#                     Richard Pausch
 #
 # This file is part of PIConGPU.
 #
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
Please keep the physical parameters and only change the number of cells into the y direction and use 4 gpus in y like in the config for 16 gpu s

PrometheusPi:
@psychocoderHPC done

PrometheusPi:
@Fettpet Did you tested the latest version (with the changes suggested by @psychocoderHPC)?

Fettpet:
Yeah I tested a1639a29b8a19a547569d3e5cbac5af3d16bfca6

PR #2481, opened 2018-01-26
pic-build: CMake multi-args with white spaces

psychocoderHPC:
follow up of #2475 and fix the [reported issue](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2475#issuecomment-360741057)

Allow to use white spaces in arguments. e.g. `./pic-build -c "..." -b omp2b`

This issue with the fix from #2475 was that all arguments of `pic-build` are forwarded as one single argument to `pic-configure`

CC-ing: @sbastrakov

### DIFF
--- a/pic-build
+++ b/pic-build
@@ -44,8 +44,7 @@ help()
     echo "$($this_dir/pic-configure --help | tail -n +12)"
 }
 
-# save cmd line args
-cmd_line_args=$@
+cmd_line_args=("$@")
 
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
Here is a good [link](https://www.thegeekstuff.com/2010/05/bash-shell-special-parameters/) where the differences between `$@` and `"$@"` are highlighted

PR #2479, opened 2018-01-26
Python visualization modules

codingS3b:
This adds a new directory called `visualization` within the `lib/python/picongpu/` directory. 
This contains classes for displaying plots for energy histograms, phase space diagrams and pngs.
All visualization modules can also be accessed from the command line.

PR #2484, opened 2018-01-29
Python MPL Visualization modules

codingS3b:
Supersedes #2479


lib/python/picongpu/visualization/energy_histogram_visualizer.py
```diff
@@ -0,0 +1,120 @@
+from picongpu.plugins.energy_histogram import EnergyHistogram
```
ax3l: uh, forgot to mention: please add the copyright line to each file (see e.g. the data readers)

codingS3b:
Could you please have a look at the axis labels and plot titles for the single visualizers?

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
```diff
@@ -0,0 +1,142 @@
...
+            (the path before ``simOutput/``)
+        """
+        if run_directory is None:
+            raise ValueError('The run_directory parameter can not be None!')
+
+        self.energy_histogram = EnergyHistogram(run_directory)
+        self.plt_obj = None
+
+    def visualize(self, iteration, ax, species='e', species_filter="all",
+                  include_overflow=False, **kwargs):
```
ax3l: I would remove `include_overflow` as an option since it will not be well visualized if it is included

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
```diff
@@ -0,0 +1,142 @@
+"""
+This file is part of the PIConGPU.
+
+Copyright 2017-2018 PIConGPU contributors
+Authors: Sebastian Starke
+License: GPLv3+
+"""
+
+
```
ax3l: one newline less would be ok here :)

lib/python/picongpu/plugins/plot_mpl/phase_space_visualizer.py
```diff
@@ -0,0 +1,158 @@
...
+        if run_directory is None:
+            raise ValueError('The run_directory parameter can not be None!')
+
+        # initialize the data-reader
+        self.phase_space = PhaseSpace(run_directory)
+        self.plt_obj = None
+
+        # for unit-conversion from SI (taken from picongpu readthedocs)
+        self.mu = 1.e6
+        self.e_mc_r = 1. / (9.1e-31 * 2.9979e8)
```
ax3l: if we want beta_gamma here, I have to check if we already specify the species' mass.
also, beta gamma per nucleon for ions would be useful

ax3l: or we just show in units of MeV / c for now...

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
codingS3b: Would you completely remove it or be able to pass it down to the reader as kwargs? So would you want
```
counts, energy_bins = self.energy_histogram.get(
            species=species,
            species_filter=species_filter,
            iteration=iteration,
            **kwargs)
```
instead of
```
counts, energy_bins = self.energy_histogram.get(
            species=species,
            species_filter=species_filter,
            iteration=iteration,
            include_overflow=include_overflow)
```

ax3l: For this specific visualization, which is not great to investigate overflows, I would say we just completely remove it.
If it's easy to pass in kwargs and that's the right place I won't stop you from allowing it there :)

lib/python/picongpu/plugins/plot_mpl/phase_space_visualizer.py
```diff
@@ -0,0 +1,158 @@
...
+        import sys
+        import getopt
+
+        def usage():
+            print("usage:")
+            print(
+                "python", sys.argv[0],
+                "-p <path to run_directory> -i <iteration>"
+                " -s <particle species> -f <species_filter>"
+                " -m <momentum term>")
```
ax3l: `momentum term` is actually both the `spatial component` and the `momentum component`.

`-m` is ok with me, but add `-m <phase space selection>` as in the data api

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
```diff
@@ -0,0 +1,142 @@
...
+            include_overflow=include_overflow)
+
+        # if it is the first time for plotting then object is
+        # created, otherwise only data is updated
+        if self.plt_obj is None:
+            self.plt_obj = ax.semilogy(energy_bins, counts, nonposy='clip')[0]
+        else:
+            self.plt_obj.set_data(energy_bins, counts)
+
+        ax.set_xlabel('Energy')
```
ax3l: 'Energy  [keV]'

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
```diff
@@ -0,0 +1,142 @@
...
+        ax.set_ylabel('Count')
+        ax.set_title('Energy Histogram for species ' +
+                     species + ', iteration ' + str(iteration))
```
ax3l: maybe add info on `species_filter` after species

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
```diff
@@ -0,0 +1,142 @@
...
+        import sys
+        import getopt
+
+        def usage():
+            print("usage:")
+            print(
+                "python", sys.argv[0],
+                "-p <path to run_directory> -i <iteration>"
+                " -s <particle species> -f <species_filter>"
+                " -o <include_overflow>")
```
ax3l: remove overflow option for viz

lib/python/picongpu/plugins/plot_mpl/phase_space_visualizer.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # prevent multiple rendering of colorbar
+        if not self.plt_obj.colorbar:
+            cbar = plt.colorbar(self.plt_obj, ax=ax)
+            cbar.set_label(
+                r'$Q / \mathrm{d}r \mathrm{d}p$ [$\mathrm{C s kg^{-1} m^{-2}}'
+                '$')
+
+        ax.set_xlabel(r'${0}$ [${1}$]'.format(meta.r, "\mathrm{\mu m}"))
+        ax.set_ylabel(r'$p_{0}$ [$\beta\gamma$]'.format(meta.p))
```
ax3l: unit should be `[$\beta\gamma m_\mathrm{species} / m_\mathrm{e}$]` right now...

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
```diff
@@ -0,0 +1,131 @@
...
+            path to the run directory of PIConGPU
+            (the path before ``simOutput/``)
+        """
+        if run_directory is None:
+            raise ValueError('The run_directory parameter can not be None!')
+
+        self.energy_histogram = EnergyHistogram(run_directory)
+        self.plt_obj = None
+
+    def visualize(self, iteration, ax, species='e', species_filter="all",
```
ax3l: would be cool if the argument order would be as in the data api: species, species_filter, iteration, ax.

of course we can also invert additional options and iterations in both, but that should happen at once to be consistent

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
```diff
@@ -0,0 +1,131 @@
...
+                  **kwargs):
+        """
+        Creates a semilogy plot on the provided axes object for
+        the data of the given iteration using matpotlib.
+
+        Parameters
+        ----------
+        iteration: int
+            the iteration number for which data will be plotted.
+        ax: matplotlib axes object
```
ax3l: a default for `ax` could be `plt.gca()`

lib/python/picongpu/plugins/plot_mpl/phase_space_visualizer.py
```diff
@@ -0,0 +1,160 @@
...
+
+        # initialize the data-reader
+        self.phase_space = PhaseSpace(run_directory)
+        self.plt_obj = None
+
+        # for unit-conversion from SI (taken from picongpu readthedocs)
+        self.mu = 1.e6
+        self.e_mc_r = 1. / (9.1e-31 * 2.9979e8)
+
+    def visualize(self, iteration, ax, species='e', species_filter='all',
```
ax3l: order: see comment above

lib/python/picongpu/plugins/plot_mpl/phase_space_visualizer.py
```diff
@@ -0,0 +1,160 @@
...
+                  ps='ypy', **kwargs):
+        """
+        Creates a phase space plot on the provided axes object for
+        the data of the given iteration using matpotlib.
+
+        Parameters
+        ----------
+        iteration: int
+            the iteration number for which data will be plotted.
+        ax: matplotlib axes object
```
ax3l: default: see comment above

lib/python/picongpu/plugins/plot_mpl/png_visualizer.py
```diff
@@ -0,0 +1,128 @@
...
+    def visualize(self, iteration, ax, **kwargs):
+        """
+        Creates a plot on the provided axes object for
+        the PNG file of the given iteration using matpotlib.
+
+        Parameters
+        ----------
+        iteration: int
+            the iteration number for which data will be plotted.
+        ax: matplotlib axes object
```
ax3l: default: see comment above

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
codingS3b: Hm, this might be a problem for us, since as of right now we are using the visualizers in the following way (as a callback function of a jupyter-slider widget):

```{python}
iteration = change["new"]
idx = 0
for plugin in self.plugin_list:
    # each plugin gets an axes object for plotting
    # the sim data resides in a subdir of the experiment called \
    # 'run'
    plugin.visualize(iteration,
        self.ax_lst[idx],
        **self.layout_config)
        idx += 1
```
and the `layout_config` contains entries the user specifies (e.g. the species, filter and momentum type). Only this order allowed us to use this uniform way of calling the visualizers since they expect different optional parameters besides the iteration and the ax.
And the PNG Visualizer has (so far, at least) no options for filter or species (since I don't know whether you have PNG files for something other than electrons). 
All these points call for a common visualizer base class, what do you think? Can we create a common interface for all kinds of visualzers?

ax3l: Sounds reasonable, yes.

codingS3b: By the way, the PNG visualizer does all its reading on its own and has no reader class like the others...maybe that should be split as well...

ax3l:
@codingS3b when you next time push to this PR, please rebase against `dev` beforehand. We updated the travis scripts.

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
codingS3b: Do you have png visualization also for other species than electrons? Does species_filter make sense for png visualization as well?
Only then we could use a base class for visualizers with the signature
```python
class BaseVisualizer
    def visualize(self, species, species_filter, iteration, **kwargs)
```

ax3l: yes, [the PNG plugin](https://picongpu.readthedocs.io/en/latest/usage/plugins/png.html#usage-plugins-png) can be used for any species and potentially in the future also `species_filters` (currently always `all`).

ax3l: Please be aware that not all visualizers are species visualizers.

Our sources can be species, [(electro-magnetic) fields](https://picongpu.readthedocs.io/en/latest/usage/plugins/sliceFieldPrinter.html#usage-plugins-slicefieldprinter) or something totally different, e.g. a [scalar of total energy](https://picongpu.readthedocs.io/en/latest/usage/plugins/energyFields.html#usage-plugins-energyfields).

codingS3b: I see... so what would you think could be an interface that is 
a) general enough and 
b) also reflects the order of arguments of the data readers?

ax3l: probably the only common argument is `ax`. Although one could also query things globally for a simulation, in our case `iteration` could also be useful (maybe supporting ranges as well).

codingS3b: When you have a png file, like `e_png_yz_0.5_006600.png`, what does the 0.5 stand for and what would be good parameter names for data readers for the coord label (yz) and that parameter behind the 0.5?
I'm asking because I'm trying to build a png reader as well

ax3l: `xy` is the plane that is cut out and `0.5` is the relative offset ("slicePoint") in the third, here `z` "axis" (in the box limit: [0.:1.]):
https://picongpu.readthedocs.io/en/latest/usage/plugins/png.html#usage-plugins-png

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
codingS3b: This default leads to the creation of another empty plot even when an ax argument is given, so I don't think we should do that...

codingS3b:
I now updated the visualization module and added a first idea of a `base_visualizer` which all others inherit from and also moved parts of the `PNGVisualizer` into a `PNG` reader class.

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
ax3l: oh, ok.
I was just thinking you do something like:
```python
def func(ax=None, ...)
    if ax is None:
        ax = plt.gca()
```

codingS3b: yeah, that's what came to my mind as well...I'll add that

ax3l:
can you please rebase against the latest `dev` and force-push your branch?

lib/python/picongpu/plugins/plot_mpl/base_visualizer.py
```diff
@@ -0,0 +1,44 @@
+import matplotlib.pyplot as plt
+
+
+class Visualizer(object):
+    def __init__(self, run_directory):
+
+        if run_directory is None:
+            raise ValueError('The run_directory parameter can not be None!')
+
```
ax3l: maybe add a filesystem check if the dir exists?

lib/python/picongpu/plugins/plot_mpl/base_visualizer.py
```diff
@@ -0,0 +1,44 @@
+import matplotlib.pyplot as plt
+
+
+class Visualizer(object):
+    def __init__(self, run_directory):
+
```
ax3l: `run_directory` parameter: please add [the usual docstring](https://github.com/ComputationalRadiationPhysics/picongpu/blob/2c8a2bdd74ccf56c3db78c62d4d33c58110b3533/lib/python/picongpu/plugins/phase_space.py#L62-L68)

lib/python/picongpu/plugins/plot_mpl/base_visualizer.py
```diff
@@ -0,0 +1,44 @@
...
+            raise ValueError('The run_directory parameter can not be None!')
+
+        self.plt_obj = None
+        self.data_reader = self._create_data_reader(run_directory)
+        self.data = None
+
+    def _create_data_reader(self, run_directory):
+        raise NotImplementedError
+
+    def _create_plt_obj(self, ax):
```
ax3l: what is the matplotlib baseclass this "plot_obj" refers to?
Do you mean an [`matplotlib.artist.Artist`](https://matplotlib.org/api/artist_api.html#matplotlib.artist.Artist)?

* https://matplotlib.org/users/artists.html

(I checked `plt.plot` which returns a list of Line2D which is a child class of `matplotlib.artist.Artist`)

lib/python/picongpu/plugins/plot_mpl/requirements.txt
```diff
@@ -0,0 +1,2 @@
+matplotlib
+numpy
```
ax3l: pls include this file in the `requirements.txt` a level above

lib/python/picongpu/plugins/plot_mpl/base_visualizer.py
codingS3b: honestly I have not looked at whether all those matplotlib plotting objects have a common baseclass...What I meant with `self.plt_obj` was any object coming from matplotlib for which you could update data...

ax3l: `plt.imshow` also returns an `matplotlib.artist.Artist`, so let use be explicit here since we are mpl only.

lib/python/picongpu/plugins/plot_mpl/base_visualizer.py
codingS3b: Isn't it sufficient that the reader performs this check? From what I saw they all check for existence in their constructors.

lib/python/picongpu/plugins/plot_mpl/requirements.txt
codingS3b: Sorry, I'm not sure how to write that correctly, do you mean `plugins/requirements.txt` would then look like this?
```
numpy
pandas>=0.21.0
h5py
matplotlib
-r .plot_mpl/requirements.txt
```

lib/python/picongpu/plugins/plot_mpl/base_visualizer.py
codingS3b: But its always some function of the axes that we call, e.g. `ax.imshow()` or `ax.semilogy()` and they all seem to return different kinds of objects

lib/python/picongpu/plugins/plot_mpl/requirements.txt
ax3l: Close, just without the dot:

```bash
$ cat requirements.txt
-r lib/python/picongpu/requirements.txt

$ cat lib/python/picongpu/requirements.txt
-r input/requirements.txt
-r plugins/requirements.txt
-r utils/requirements.txt

$ cat lib/python/picongpu/plugins/requirements.txt
numpy
pandas>=0.21.0
h5py
matplotlib 
-r plot_mpl/requirements.txt
```

codingS3b: Thanks, added this line now.

lib/python/picongpu/plugins/plot_mpl/base_visualizer.py
```diff
@@ -0,0 +1,96 @@
...
+
+        Parameters
+        ----------
+        ax: None or matplotlib.axes.Axes instance
+        Returns
+        -------
+        A matplotlib.axes.Axes instance
+        """
+
+        return ax or plt.gca()
```
ax3l: I don't understand this. Won't this always just return the parameter `ax`?
Shouldn't it read
```python
if ax is None:
    return plt.gca()
else:
    return ax
````
?

File: lib/python/picongpu/plugins/png.py [hunk 20]
```diff
@@ -0,0 +1,207 @@
+"""
+This file is part of the PIConGPU.
+
+Copyright 2017-2018 PIConGPU contributors
+Authors: Sebastian Starke
+License: GPLv3+
+"""
+
+import os
+import matplotlib.image as mpimg
```
ax3l: I wonder if there is an alternative way to read in the PNG.
Would be in principle cool if the "data readers" would not need `matplotlib` but only the plotters...

anyway, ok for now with me.

lib/python/picongpu/plugins/plot_mpl/base_visualizer.py
codingS3b: From my understanding the code that I wrote is just a shorter version of your suggestion (or at least I wanted it to be). When `ax=None` this will return the second variable, in this case `plt.gca()`. I tested it on ipython and it worked as expected.

lib/python/picongpu/plugins/energy_histogram.py
```diff
@@ -100,8 +100,8 @@ def get_iterations(self, species, species_filter="all"):
                            delimiter=" ",
                            dtype=np.uint64).as_matrix()[:, 0]
 
-    def get(self, species, species_filter="all", iteration=None,
-            include_overflow=False):
+    def get(self, species=None, species_filter="all", iteration=None,
```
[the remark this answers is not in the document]
codingS3b: No its not, we can live without it, I just checked it.

lib/python/picongpu/plugins/phase_space.py
```diff
@@ -188,7 +188,8 @@ def get_iterations(self, species, species_filter='all', ps=None):
 
         return iterations
 
-    def get(self, species, species_filter='all', ps=None, iteration=None):
+    def get(self, species=None, species_filter='all', iteration=None, ps=None,
```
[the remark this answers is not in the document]
codingS3b: See above

lib/python/picongpu/plugins/png.py
codingS3b: Using scipy would work without a lot of changes. Converting to a bytearray seems to loose the shape of the image matrix: when feeding that bytearray into matplotlib it complains about invalid dimensions for image data.
So if scipy as a dependency would be okay for you then I will change that

lib/python/picongpu/plugins/energy_histogram.py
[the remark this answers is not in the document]
ax3l: ok cool, then please leave this two files unchanged :)

lib/python/picongpu/plugins/plot_mpl/base_visualizer.py
ax3l: Wow. Magic.

lib/python/picongpu/plugins/phase_space.py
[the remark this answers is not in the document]
codingS3b: Is it still okay to change the order of the variables here? I recall you wanted to have species, filter and iteration before all other class dependent options so that is why I made the change.

lib/python/picongpu/plugins/png.py
ax3l: I checked the `scipy` library (not the [SciPy ecosystem/software stack](https://www.scipy.org/about.html)) dep list and it looks quite small. yes, let's do this.

lib/python/picongpu/plugins/phase_space.py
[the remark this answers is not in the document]
ax3l: Good question here. Well, ok.

If you change it, please adjust the `.rst` docs in `docs/source/usage/plugins/phaseSpace.rst` as well.

ax3l:
@codingS3b Sorry forgot to ping you, ready for rebase :)

codingS3b:
Final rebase was done!

ax3l:
thank you very much, great work! :rocket: :sparkles:

Issue #2485, opened 2018-01-29
Eliminate the usage of M_PI


sbastrakov:
Although widely used, M_PI and its derivative constants are not standard C++. It is also not ideal for template code, as type casts are normally required. 

Most of the code uses PI defined in physicalConstants.param (which also requires type casts, but this is not the point for this issue). However, some pieces still use M_PI, these should be changed to either PI or, in case it's not possible, with some new standard-conforming code.

PR #2486, opened 2018-01-29
Eliminate M_PI from PMacc

sbastrakov:
This is part of #2485. As discussed with @psychocoderHPC , fist I only made changes to PMacc: added a new file with Pi constants and some derivatives (only those currently used, not for all M_... cases), replace usage of M_PI with the new values.

### DIFF
--- a/include/pmacc/algorithms/math/defines/pi.hpp
+++ b/include/pmacc/algorithms/math/defines/pi.hpp
@@ -37,6 +37,7 @@ namespace math
         static constexpr T_Type value = static_cast< T_Type >(
             3.141592653589793238462643383279502884197169399
         );
+        static constexpr T_Type doubleValue = static_cast< T_Type >( 2.0 ) * value;
         static constexpr T_Type halfValue = value / static_cast< T_Type >( 2.0 );
         static constexpr T_Type quarterValue = value / static_cast< T_Type >( 4.0 );
         static constexpr T_Type doubleReciprocalValue = static_cast< T_Type >( 2.0 ) / value;
--- a/include/pmacc/algorithms/math/defines/pi.hpp
+++ b/include/pmacc/algorithms/math/defines/pi.hpp
@@ -30,13 +30,16 @@ namespace math
 namespace pi
 {
 
-    /** Value of pi as T_Type
+    /** Values of pi and related constants as T_Type
      */
     template< typename T_Type >
     struct Pi
-	{
-		static constexpr T_Type value = static_cast<T_Type>(3.141592653589793238462643383279502884197169399);
-	};
+    {
+        static constexpr T_Type value = static_cast<T_Type>(3.141592653589793238462643383279502884197169399);
+        static constexpr T_Type halfValue = value / static_cast<T_Type>(2.0);
+        static constexpr T_Type quarterValue = value / static_cast<T_Type>(4.0);
+        static constexpr T_Type doubleReciprocalValue = static_cast<T_Type>(2.0) / value;
+    };
 
 } //namespace pi
 } //namespace math
--- a/include/pmacc/algorithms/math/defines/pi.hpp
+++ b/include/pmacc/algorithms/math/defines/pi.hpp
@@ -1,0 +1,44 @@
+/* Copyright 2018 Sergei Bastrakov
+ *
+ * This file is part of PMacc.
+ *
+ * PMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with PMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+namespace pmacc
+{
+namespace algorithms
+{
+namespace math
+{
+namespace pi
+{
+
+    /** Value of pi as T_Type
+     */
+    template< typename T_Type >
+    struct Pi
+	{
+		static constexpr T_Type value = static_cast<T_Type>(3.141592653589793238462643383279502884197169399);
+	};
+
+} //namespace pi
+} //namespace math
+} //namespace algorithms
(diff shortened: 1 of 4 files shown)


psychocoderHPC:
Thanks you for this PR: I am not sure if we should use the predefined values `halfValue, ...` or if it is better to write it explicit as `pi::Pi< float_T >::value / float_T( 2.0 )`

include/pmacc/algorithms/math/defines/pi.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+#pragma once
+
+namespace pmacc
+{
+namespace algorithms
+{
+namespace math
+{
+namespace pi
```
psychocoderHPC: please remove the namespace `pi`

include/pmacc/algorithms/math/defines/pi.hpp
```diff
@@ -0,0 +1,47 @@
...
+{
+
+    /** Values of pi and related constants as T_Type
+     */
+    template< typename T_Type >
+    struct Pi
+    {
+        static constexpr T_Type value = static_cast<T_Type>(3.141592653589793238462643383279502884197169399);
```
psychocoderHPC: Missing spaces, must be `static_cast< T_Type >( 3.141592653589793238462643383279502884197169399 );`

include/pmacc/algorithms/math/defines/pi.hpp
```diff
@@ -0,0 +1,47 @@
...
+        static constexpr T_Type halfValue = value / static_cast<T_Type>(2.0);
+        static constexpr T_Type quarterValue = value / static_cast<T_Type>(4.0);
+        static constexpr T_Type doubleReciprocalValue = static_cast<T_Type>(2.0) / value;
+    };
+
+} //namespace pi
```
psychocoderHPC: missing space `// namespace pi`

include/pmacc/algorithms/math/defines/pi.hpp
```diff
@@ -0,0 +1,47 @@
...
+{
+
+    /** Values of pi and related constants as T_Type
+     */
+    template< typename T_Type >
+    struct Pi
+    {
+        static constexpr T_Type value = static_cast<T_Type>(3.141592653589793238462643383279502884197169399);
+        static constexpr T_Type halfValue = value / static_cast<T_Type>(2.0);
+        static constexpr T_Type quarterValue = value / static_cast<T_Type>(4.0);
```
psychocoderHPC: We should discuss the pro/con of this pre-defines   offline with @ax3l.

include/pmacc/algorithms/math/defines/pi.hpp
```diff
@@ -0,0 +1,47 @@
...
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with PMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
```
psychocoderHPC: please add one empty line more.

sbastrakov: In the guidelines it says one line around #pragma once (and other files seem to follow that).

ax3l: but double new-line after the CPP header is finished (similar to the PEP8 requiremet in `import`s in Python). Maybe that is missing still in the contrib. guidelines.

include/pmacc/algorithms/math/defines/pi.hpp
sbastrakov: Ok, was confused that bessel functions have their own namespace and did not notice that other files there indeed do not.

include/pmacc/algorithms/math/defines/pi.hpp
sbastrakov: Ok, I'm also not 100% sure these should exist, just added as they are part of the "M_PI package" and were used already in the code.

ax3l: Since we have `constexpr` in modern C++ we might not need those good ol' C-defines ;)

Anway, if you want to define a few, I insist we add 2 pi, aka [tau](https://tauday.com/) since it's the only true circle constant, superior to all other fractions of it.

include/pmacc/algorithms/math/defines/pi.hpp
```diff
@@ -0,0 +1,47 @@
...
+     */
+    template< typename T_Type >
+    struct Pi
+    {
+        static constexpr T_Type value = static_cast< T_Type >(
+            3.141592653589793238462643383279502884197169399
+        );
+        static constexpr T_Type halfValue = value / static_cast< T_Type >( 2.0 );
+        static constexpr T_Type quarterValue = value / static_cast< T_Type >( 4.0 );
+        static constexpr T_Type doubleReciprocalValue = static_cast< T_Type >( 2.0 ) / value;
```
PrometheusPi: QUESTION:
Why do you expose a double reciprocal value but not a (single) reciprocal value?

PrometheusPi: I assume from the code above you just implemneted the versions we used.
C provides various other constants. See https://www.gnu.org/software/libc/manual/html_node/Mathematical-Constants.html. 
But from my side, it is fine to implement them, when they are needed.

sbastrakov: Exactly, I was not sure if these are needed at all and so only added those that are used in PMacc.

sbastrakov:
`doubleValue` - Might add it when it is actually used? (And it seems to be elsewhere in the code, but not in PMacc). Or could also do it now, but then probably it would be more logical to also add `halfValue`, `reciprocalValue`, etc.

ax3l:
In PIConGPU core, `synchrotronPhotons` & `Bremsstrahlung` have 2pi factors inside :)

psychocoderHPC:
@sbastrakov  We can merge it as soon as 2pi is added and @ax3l  is happy :-)

include/pmacc/algorithms/math/defines/pi.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+    /** Values of pi and related constants as T_Type
+     */
+    template< typename T_Type >
+    struct Pi
+    {
+        static constexpr T_Type value = static_cast< T_Type >(
+            3.141592653589793238462643383279502884197169399
+        );
+        static constexpr T_Type doubleValue = static_cast< T_Type >( 2.0 ) * value;
```
ax3l: @sbastrakov the CI complains about the end-of-line whitespaces in this line :)

PrometheusPi:
@sbastrakov Could you squash all commits that only fix white spaces, newlines and etc. into the corresponding commits so that the total number of commits is reduced? That would keep the pull request compact.

PR #2489, opened 2018-01-29
Fix warning concerning forward declarations of pmacc::detail::Environment

sbastrakov:
This is a structure, but in several places forward declared as a class, which caused a warning. Replace `class `with `struct `in forward declarations.

Of course, it would be easier to just make `pmacc::detail::Environment` a class. However, it does not store any internal state, so to me structure feels better fitting. Btw, there is a similar type `pmacc::Environment`, which is a class. Maybe it would make sense to make it a structure too (or make both a class for uniformity).

### DIFF
--- a/include/pmacc/dataManagement/DataConnector.hpp
+++ b/include/pmacc/dataManagement/DataConnector.hpp
@@ -216,7 +216,7 @@ namespace pmacc
 
     private:
 
-        friend class detail::Environment;
+        friend struct detail::Environment;
 
         static DataConnector&
         getInstance()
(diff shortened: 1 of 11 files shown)


psychocoderHPC:
This is a BUG which also effects the latest release but IMO a back port is not needed.

sbastrakov:
I think while it's a little better to have class/struct matching in forward declarations and definition, it is not a violation of standard, so not actually a bug. It could lead sometimes to linking issues in Visual Studio due to name mangling, but that's probably due to quirks of VS. (And ofc here there is a header-only library).

Any feedback regarding whether` pmacc::Environment` should also be made a struct?

psychocoderHPC:
@sbastrakov If we not mixed struct and class for `pmacc::Environment` than it can currently stay as it is.

Issue #2476, opened 2018-01-25
Compound literals used inside PMACC_CONST_VECTOR


sbastrakov:
In my Visual Studio there are build problems caused by how PMACC_CONST_VECTOR is implemented. I tried to figure it out and think now it is all caused by [compound literals](http://en.cppreference.com/w/c/language/compound_literal) used to check that vector size is correct. Compound literals are C99 and not standard C++, however GCC supports it for C++ as well.

Here is a greatly simplified example I've come up with and some comments there:
http://coliru.stacked-crooked.com/view?id=69f1fbee3fcb86ac

sbastrakov:
The problem is actually in `PMACC_COUNT_ARGS` macro called inside `PMACC_CONST_VECTOR` to check that the size is correct. René suggested implementing `PMACC_COUNT_ARGS` by simply calling `BOOST_PP_VARIADIC_SIZE`. At first glance it seems to help on my machine, however can't be 100% sure as the build still fails for other reasons. Will take a closer look on this.

PR #2494, opened 2018-02-01
Partially replace compound literals in PMacc

sbastrakov:
As pointed out in #2476 compound literals are not standard C++ and ideally should be fully replaced to provide better portability. However, there seems to be no easy and portable way to fully get rid of them in the current code which counts the number of macro arguments. Boost.Preprocessor provides a macro exactly for that, but it is enabled only for some compilers (comment there says others are not tested), notably, enabled for Visual Studio, but disabled for nvcc. So I've decided to use a macro from Boost if it is enabled and keep the previous implementation as a fallback. Currently it works on my Windows system with Boost implementation and for nvcc the previous implementation is used.

### DIFF
--- a/include/pmacc/ppFunctions.hpp
+++ b/include/pmacc/ppFunctions.hpp
@@ -58,7 +58,7 @@
  * @param ... arguments
  * @return false if no arguments are given, else true
  */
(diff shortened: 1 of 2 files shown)


sbastrakov:
Please don't merge yet, I realized that the proposed change only works for up to 64 parameters, and there is one vector of size 79. Will extend the solution to work for this case.

PR #2495, opened 2018-02-01
Get rid of <sys/time.h>

sbastrakov:
It is not portable and only used for the `TimeIntervall `class. I've replaced it with features of C++11 std::chrono.

### DIFF
--- a/include/pmacc/simulationControl/TimeInterval.hpp
+++ b/include/pmacc/simulationControl/TimeInterval.hpp
@@ -23,11 +23,11 @@
 
 #include "pmacc/types.hpp"
 
-#include <sys/time.h>
-
+#include <chrono>
 #include <string>
 #include <iostream>
 #include <sstream>
+
 
 namespace pmacc
 {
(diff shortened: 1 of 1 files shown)


include/pmacc/simulationControl/TimeInterval.hpp
```diff
@@ -46,9 +46,9 @@ namespace pmacc
          */
         static double getTime()
         {
-            struct timeval act_time;
-            gettimeofday(&act_time, nullptr);
-            return (double)act_time.tv_sec*1000. + (double)act_time.tv_usec / 1000.;
```
PrometheusPi: QUESTION:
@ComputationalRadiationPhysics/picongpu-developers 
Does `usec` mean µs? Is this is the case, we have underestimated the time.

sbastrakov: I think that was fine: uses was nano seconds, so multiply by 1000 to get milliseconds.

ax3l: actually I think the old one was wrong.
tv_sec divided by 1000 would have been ms and usec multiplied by 1000 would also have been ms.

but it did it exactly the other way around, no?

ax3l: ah no wait, I am still tired. all was fine and is fine

PR #2499, opened 2018-02-01
tbg: Add content of -o arg to env

jkelling:
This is to allow for all variables used in the *.tpl file to be set through the -o option, instead of only being able to override variables which have otherwise been defined in either the *.tpl or *.cfg.

### DIFF
--- a/src/tools/bin/tbg
+++ b/src/tools/bin/tbg
@@ -461,6 +461,10 @@ fi
 mkdir -p "$job_outDir/tbg"
 cd "$job_outDir"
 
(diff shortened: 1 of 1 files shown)


src/tools/bin/tbg
```diff
@@ -461,6 +461,9 @@ fi
 mkdir -p "$job_outDir/tbg"
 cd "$job_outDir"
 
+for i in $tooltpl_overwrite; do
```
ax3l: pls add a comment above why you do this, otherwise we will forget

PR #2503, opened 2018-02-05
Visual Studio: ERROR macro defined in wingdi.h

sbastrakov:
There is ERROR macro defined in wingdi.h file. I assume it is somehow included from VS's implementation of the standard library. So there is a collision with ArgsParser::ArgsErrorCode::ERROR. Since the problem is only in main.cpp, I decided just to undef it manually there.

### DIFF
--- a/include/picongpu/main.cpp
+++ b/include/picongpu/main.cpp
@@ -33,6 +33,14 @@
 #include <picongpu/simulation_defines.hpp>
 
 
+/* Workaround for Visual Studio to avoid a collision between ERROR macro
+ * defined in wingdi.h file (included from some standard library headers) and
+ * enumerator ArgsParser::ArgsErrorCode::ERROR.
+ */
+#ifdef _MSC_VER
+#   undef ERROR
+#endif
+
 /*! start of PIConGPU
  *
  * @param argc count of arguments in argv


ax3l:
Huh, weird. Can you post some reference on this?
Where does this collide?

include/picongpu/main.cpp
```diff
@@ -33,6 +33,11 @@
 #include <picongpu/simulation_defines.hpp>
 
 
+// Workaroud for Visual Studio to avoid collisions with ERROR define
+#ifdef _MSC_VER
```
ax3l: add note on boost argsparser please as in the PR description

ax3l: please also add an `ifdef`:
```C++
#ifdef _MSC_VER
#   ifdef ERROR
#       undef ERROR
#   endif
#endif
````

ax3l:
And alternative approach could be to add the `NOGDI` pre-processor define for MSVC either in VisualStudio or CMake.

include/picongpu/main.cpp
sbastrakov: Why? If the macro is not defined, undef [does nothing](http://en.cppreference.com/w/cpp/preprocessor/replace)

ax3l: > If the identifier does not have associated macro, the directive is ignored. 

True, in that case the guard as now is fine. Pls just add the comment that we know it's about the enum in boost argsparse that collides

ax3l: typo: workarou**n**d

sbastrakov:
Then it might be better to do just that and not put this workaround into the code. Sorry, did not look it up myself before making a workaround.

include/picongpu/main.cpp
sbastrakov: But it's not boost, it's in picongpu/ArgParser.hpp.
typo - thanks, will fix.

PR #2505, opened 2018-02-05
Fix type of index in OpenMP-parallelized loop

sbastrakov:
A tiny fix of unsigned index into signed - OpenMP officially allows only signed. (And VS of all compilers complains about it).

### DIFF
--- a/include/pmacc/memory/buffers/HostBufferIntern.hpp
+++ b/include/pmacc/memory/buffers/HostBufferIntern.hpp
@@ -129,7 +129,7 @@ public:
         typedef DataBoxDim1Access<DataBoxType > D1Box;
         D1Box d1Box(memBox, this->getDataSpace());
         #pragma omp parallel for
-        for (size_t i = 0; i < current_size; i++)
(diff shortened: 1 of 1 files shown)


PrometheusPi:
good catch @sbastrakov :+1: 

Just for documentation see here:
https://stackoverflow.com/questions/2820621/why-arent-unsigned-openmp-index-variables-allowed

or the user manual section 2.4.1 (as referenced in the stackoverflow post) 
page 14 line 6 
http://www.openmp.org/wp-content/uploads/cspec20.pdf

ax3l:
Seems to be an OpenMP 2.0 restriction and MSVC seems to be quite behind compared to other compilers with OpenMP support. In later versions, as @PrometheusPi linked above, this restriction is not present anymore.

sbastrakov:
Sorry, did not think of it in the first place. Now with gcc there is a super-long warning concerning comparison of signed and unsigned in that loop. I guess should also fix that?

ax3l:
yes please

PR #2506, opened 2018-02-05
Fix issues with name hiding in Particles

sbastrakov:
Some (relatively) global names were hidden by inaccessible base class members. Explicitly put namespace for such names.

### DIFF
--- a/include/picongpu/particles/Particles.hpp
+++ b/include/picongpu/particles/Particles.hpp
@@ -110,7 +110,7 @@ class Particles : public ParticlesBase<
 {
 public:
 
-    typedef ParticleDescription<
+    typedef pmacc::ParticleDescription<
         T_Name,
         SuperCellSize,
         T_Attributes,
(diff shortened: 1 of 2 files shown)


include/picongpu/particles/Particles.hpp
```diff
@@ -138,13 +138,13 @@ class Particles : public ParticlesBase<
             >
         >::type
     > SpeciesParticleDescription;
-    typedef ParticlesBase<SpeciesParticleDescription, MappingDesc, DeviceHeap> ParticlesBaseType;
```
ax3l: weird, `DeviceHeap` should be such a global as well, no?

ax3l:
So you are saying, this is only a problem if a base class has a member that is also named like a public "global" typedef of ours?

Shouldn't we then rather rename all the typdef-ed members properly (or our global)?

include/picongpu/particles/Particles.hpp
ax3l: ah no, `DeviceHeap` is probably not a member of the class or it's parent class

PR #2508, opened 2018-02-05
Python: Multi-Iteration Energy Histogram

### LINKED ISSUE #2482: Plugin Data Reader: Slow np.loadtxt
Reported first by @codingS3b: some python readers, e.g. for the energy histogram plugin, are slow.

The main reason is their usage of `numpy.loadtxt` which we could replace with another numpy reader or pandas' `read_csv` to speed things up.

ax3l:
Close #2482

Add pandas backend and multi-iteration fronend to python reader for energy histogram plugin.

Also add requirements files for python.

### New Features

- support for iteration lists: request multiple iterations as list and get an (ordered) dictionary with iteration as index

```python
counts, bins = energy_histogram.get('H', 'all', [1000, 99000])

counts.keys()
# odict_keys([1000, 99000])
```

- support for "all iterations available": request an unspecified iteration with the default `None` parameter and get all iterations as ordered dictionary (see above)

```python
counts, bins = energy_histogram.get('H', 'all')

counts.keys()
# odict_keys([100, 200, 300, ..., 100000])
```


PrometheusPi:
Great update!
Why did you favor panda csv over numpy loadtxt ? Is it faster? 
Great idea with the dicts.

ax3l:
Yes, `np.loadtxt` is extremely slow #2482 and also pandas is way more versatile for the kind of manipulations and selections we do on top.

lib/python/picongpu/plugins/energy_histogram.py
```diff
@@ -111,46 +115,67 @@ def get(self, species, species_filter="all", iteration=None,
...
+            A list of iterations is allowed as well.
         include_overflow : boolean, default: False
             Include overflow and underflow bins as the first/last bins.
 
         Returns
         -------
         energies : np.array of dtype float [unitless]
             count of particles in each bin
-            @todo if iteration is a list, return a dict
+            If iteration is a list, returns a dict
```
ax3l: ah, that's not true. I return a matrix with the first index being the iteration in the order as the user requested it.

lib/python/picongpu/plugins/energy_histogram.py
```diff
@@ -111,46 +115,68 @@ def get(self, species, species_filter="all", iteration=None,
...
         -------
         energies : np.array of dtype float [unitless]
             count of particles in each bin
-            @todo if iteration is a list, return a dict
+            If iteration is a list, returns a matrix with first index being
+            the iteration as in the order of the requested iterations.
         bins : np.array of dtype float [keV]
             upper ranges of each energy bin
         """
         if iteration is None:
```
codingS3b: Why is it actually that you give a default value to iteration which will throw an exception? 
Since the parameter is needed anyway, wouldn't it be good to not provide a default then?

lib/python/picongpu/plugins/energy_histogram.py
```diff
@@ -111,46 +115,68 @@ def get(self, species, species_filter="all", iteration=None,
...
+
+        # set DataFrame column names properly
+        data.columns = [
+            'iteration',
+            'underflow'
+        ] + list(bins) + [
+            'overflow',
+            'sum'
+        ]
+        # set interation as index
```
codingS3b: little typo in the comment

codingS3b: little typo in the comment

codingS3b:
Looks good! Have you timed it yet and compared to the current implementation?

When I would like to get the histogram values for all available iterations in the file, we would still have to read the file twice, right? (Once for getting all iterations and then again for getting the histogram)..or am I thinking too complicated?

lib/python/picongpu/plugins/energy_histogram.py
ax3l: Great question, thanks!
unfortunately, there is no good default for an iteration to select since it is very case dependent.
The whole idea about defaults here is, that we have
- a consequent order of arguments
- allow to select only parts of them to users

That means a user will always see the order "species, filter, iteration, specific_options" in the data and viz API. But if they are visualizing the default "all" filter for example, they can just use the python-specific  syntax to set a "later" optional argument `get("e", iteration=1234)`

It's therefor ok to set all later to default parameters and still require them and common practice in Python APIs.

ax3l: Of course if one is clever one can get it also done with a good order of arguments, but orders sometimes contradict each other for different workflows. Here for example, iteration could always be the first argument, but I would have to check again why I put it last (and if we change it all APIs need the change at once)

ax3l:
> When I would like to get the histogram values for all available iterations in the file, we would still have to read the file twice, right? (Once for getting all iterations and then again for getting the histogram)..or am I thinking too complicated?

If your workflow is "give me all iterations, I don't care which" then you are right: you would have to query it twice.

We could change that by
- allowing an "all iterations" value for `iteration=` (e.g. `None`) could be interpreted as "all iterations" to connect to your question above
- we probably would have to return still a dict, so the user actually knows the iterations

lib/python/picongpu/plugins/energy_histogram.py
ax3l: Alternative approach to your question below: get "all iterations" if `None` (default) is specified

codingS3b: Sounds like a plan to hand out all iterations when not specific ones are provided. I like that :)

ax3l:
@PrometheusPi @codingS3b updated with new functionality :)

lib/python/picongpu/plugins/energy_histogram.py
```diff
@@ -111,46 +115,74 @@ def get(self, species, species_filter="all", iteration=None,
...
         include_overflow : boolean, default: False
             Include overflow and underflow bins as the first/last bins.
 
         Returns
         -------
-        energies : np.array of dtype float [unitless]
+        counts : np.array of dtype float [unitless]
             count of particles in each bin
-            @todo if iteration is a list, return a dict
+            If iteration is a list, returns (ordered) dict with iterations as
```
PrometheusPi: QUESTION:
why did you write the adjective `ordered` in parenthesis? 
It is always a dict of type `collections.OrderedDict`.

ax3l: yes, I just felt like saying it's a dict and "btw ordered". Shall I be more explicit? I had the feeling the `OrderedDict` type might confuse people but... dunno.

PrometheusPi: that's fine with me

PR #2509, opened 2018-02-06
Fix signed and unsigned comparison

sbastrakov:
The issue was introduced in my earlier pull request #2505 (sorry, did not check it back then) and results in a very long compile warning on Linux.

### DIFF
--- a/include/pmacc/memory/buffers/HostBufferIntern.hpp
+++ b/include/pmacc/memory/buffers/HostBufferIntern.hpp
@@ -124,12 +124,12 @@ public:
     void setValue(const TYPE& value)
     {
         __startOperation(ITask::TASK_HOST);
-        size_t current_size = this->getCurrentSize();
+        int64_t current_size = static_cast< int64_t >(this->getCurrentSize());
         auto memBox = getDataBox();
         typedef DataBoxDim1Access<DataBoxType > D1Box;
         D1Box d1Box(memBox, this->getDataSpace());
         #pragma omp parallel for
-        for (int i = 0; i < current_size; i++)
+        for (int64_t i = 0; i < current_size; i++)
         {
             d1Box[i] = value;
         }


include/pmacc/memory/buffers/HostBufferIntern.hpp
```diff
@@ -124,12 +124,12 @@ class HostBufferIntern : public HostBuffer<TYPE, DIM>
...
     {
         __startOperation(ITask::TASK_HOST);
-        size_t current_size = this->getCurrentSize();
+        long current_size = static_cast< long >(this->getCurrentSize());
         auto memBox = getDataBox();
         typedef DataBoxDim1Access<DataBoxType > D1Box;
         D1Box d1Box(memBox, this->getDataSpace());
         #pragma omp parallel for
-        for (size_t i = 0; i < current_size; i++)
+        for (long i = 0; i < current_size; i++)
```
PrometheusPi: QUESTION:
I am a bit confused. Before this pull request a `size_t` type number `current_size` was compared to a `size_t` type number `i`. 
How could this cause an error of comparing unsigned and signed integers if both are of the same type `size_t` (unsigned).

Is just the last pull request #2505 not shown here - since the omp parallel should not use size_t, as you pointed put in #2505.  But I see a `size_t` as old code.

include/pmacc/memory/buffers/HostBufferIntern.hpp
```diff
@@ -124,12 +124,12 @@ class HostBufferIntern : public HostBuffer<TYPE, DIM>
     void setValue(const TYPE& value)
     {
         __startOperation(ITask::TASK_HOST);
-        size_t current_size = this->getCurrentSize();
+        long current_size = static_cast< long >(this->getCurrentSize());
```
psychocoderHPC: please use `int64_t` instead of long because the size of  `long` is not defined and depend on the compiler. If it should be 32 bit please use `int32_t` or `int` (even if int is also not defined)

include/pmacc/memory/buffers/HostBufferIntern.hpp
PrometheusPi: *offline discussion with @psychocoderHPC:*
Just update your branch to include your own commit. That prevents this merge conflict error.

sbastrakov: Yeah, sorry, I branched out of the older version of mainline/dev.

include/pmacc/memory/buffers/HostBufferIntern.hpp
sbastrakov: Ok. Whether it should be 32 bit or 64 bit I'm not sure. It was size_t originally, so I guess 64 bit?

PrometheusPi: Yes, since this is executed on the CPU, memory is usually not a constrain. Using 64 bits would be saver.

ax3l: This is not the point here.
The `getCurrentSize()` is the number of elements of a container and can potentially be large even on node.
Ideally, we should always assume `int64_t` for such elements, 32bit can be too small in some cases (e.g. 2 billion floats of a particle attribute is just 7.8 GByte)

include/pmacc/memory/buffers/HostBufferIntern.hpp
ax3l: sure, just rebase against the current `dev`

sbastrakov:
I rebased using mainline/dev. And used int64_t. However, it seems to be [not standard, but optional](http://en.cppreference.com/w/cpp/types/integer). But VS supports it as well.

PR #2512, opened 2018-02-06
Calorimeter: Validate minEnergy

### LINKED ISSUE #2511: Calorimeter: logScale does not check minEnergy
The `--logScale` option of the `ParticleCalorimeter` plugin is not entirely safe as it doesn't prevent `nan`s when you also pass `--minEnergy 0` (in keV).

It is a boolean and passing `--logScale 0` will still result in a logarithmic energy scale.
https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/include/picongpu/plugins/particleCalorimeter/ParticleCalorimeterFunctors.hpp#L49
https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp#L85

Therefore the energy bins will be the following `<underflow> nan nan nan ... <maxEnergy> <overflow>`.

ax3l:
Validate the input of `minEnergy` in the particle calorimeter plugin.

Avoids both negative energies as minimum and zero-energies for log scale energy ranges.

Fix #2511

### DIFF
--- a/include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
+++ b/include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
@@ -252,6 +252,22 @@ private:
                 << std::endl;
             throw std::runtime_error(msg.str());
         }
+        if(this->minEnergy < float_X(0.0))
+        {
+            std::stringstream msg;
+            msg << "[Plugin] [" << this->prefix
+                << "] minEnergy can not be negative."
+                << std::endl;
+            throw std::runtime_error(msg.str());
+        }
+        if(this->logScale && this->minEnergy == float_X(0.0))
+        {
+            std::stringstream msg;
+            msg << "[Plugin] [" << this->prefix
+                << "] minEnergy can not be zero in logarithmic scaling."
+                << std::endl;
+            throw std::runtime_error(msg.str());
+        }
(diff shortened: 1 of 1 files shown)


include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -252,6 +252,22 @@ class ParticleCalorimeter : public ISimulationPlugin
                 << std::endl;
             throw std::runtime_error(msg.str());
         }
+        if(this->minEnergy < float_X(0.0))
+        {
+            std::stringstream msg;
+            msg << "[Plugin] [" << this->prefix
+                << "] minEnergy can not be negative."
```
n01r: cannot

include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -252,6 +252,22 @@ class ParticleCalorimeter : public ISimulationPlugin
...
+                << std::endl;
+            throw std::runtime_error(msg.str());
+        }
+        if(this->logScale && this->minEnergy == float_X(0.0))
+        {
+            std::stringstream msg;
+            msg << "[Plugin] [" << this->prefix
+                << "] minEnergy can not be zero in logarithmic scaling."
```
n01r: cannot

Issue #2446, opened 2017-12-23
Prg Env in Piz Daint Profile


ax3l:
Why requesting a new CMake package, I was told we are currently loading the programming environment on Piz Daint the "wrong" / discouraged way.

We should first load `daint-gpu` (GPU nodes) or `daint-mc` (non-GPU nodes) and then all other modules. We should update the profile to make it more future proof. With such, a `CMake/3.10.1` module is now also available :)

ax3l:
While we are on it, update to:
- PNGwriter 0.7.0
- libSplash 1.7.0

in our [quick start script](https://gist.github.com/ax3l/68cb4caa597df3def9b01640959ea56b)

steindev:
Not relevant anymore. Closing.

PR #2520, opened 2018-02-08
Splash: 1.7.0

ax3l:
Updates our dependency to libSplash to 1.7.0+

Unfortunately, as with the last modernized CMake installs of PNGwriter, an install of libSplash 1.7.0 will not configure successfully in an (old) version of PIConGPU requiring 1.6.0+

This also pulls the requirement to CMake 3.10.0+ and removes the old `FindSplash.cmake` scripts.

### Before Merge

- [x] exchange deprecated `readAttribute`/`readGlobalAttribute` with safer `readAttributeInfo`/`readGlobalAttributeInfo`: see https://github.com/ComputationalRadiationPhysics/libSplash/pull/248
- compile suite:
  - [x] CMake 3.10.0
  - [x] libSplash 1.7.0
- update install/dependency docs, example profiles:
  - [x] needs rebase after #2521
  - [x] CMake 3.10+
  - [x] libSplash 1.7.0+
- [x] travis: CMake 3.10.0
- RT tests:
  - [x] KHI HDF5 dump & restart (16 GPUs)
  - [x] LWFA HDF5 dump & restart (16 GPUs)

### After Merge

- update container/[spack](https://github.com/ComputationalRadiationPhysics/spack-repo) recipe:
  - [ ] CMake 3.10+
  - [ ] libSplash 1.7.0+
- Piz Daint: #2446
  - new module chain with proper prgm environment 
  - [Piz Daint QuickStart update](https://gist.github.com/ax3l/68cb4caa597df3def9b01640959ea56b) (PNGwriter 0.7.0 & libSplash 1.7.0)


ax3l:
@Flamefire did I do this right? (See last commit in PR)

buildsystem/CompileSuite/autoTests/new_commits.sh
```diff
@@ -94,7 +94,7 @@ touch "$thisDir"runGuard
             #export PIC_COMPILE_SUITE_CMAKE="-DPIC_ENABLE_PNG=OFF -DALPAKA_CUDA_ARCH=35"
             export PIC_BACKEND="cuda"
             . /etc/profile
-            module load gcc/4.9.4 boost/1.62.0 cmake/3.7.0 cuda/8.0.44 openmpi/1.10.4
+            module load gcc/4.9.4 boost/1.62.0 cmake/3.10.0 cuda/8.0.44 openmpi/1.10.4
```
PrometheusPi: QUESTION:
Why does the compile suite uses cmake/3.10.0 while all other use cmake 3.10.1?

etc/picongpu/titan-ornl/picongpu.profile.example
```diff
@@ -37,7 +37,7 @@ export FC=$(which ftn)
 
 # required tools and libs
 module load git
-module load cmake3/3.9.0
+module load cmake3/3.10.0  # missing
```
PrometheusPi: QUESTION:
Do you want to commit using a missing module?

src/tools/png2gas/CMakeLists.txt
```diff
@@ -96,16 +96,10 @@ endif()
...
-find_package(Splash 1.6.0 REQUIRED COMPONENTS PARALLEL)
+find_package(Splash 1.7.0 REQUIRED CONFIG COMPONENTS PARALLEL)
 
-if(Splash_FOUND)
-    include_directories(SYSTEM ${Splash_INCLUDE_DIRS})
-    list(APPEND Splash_DEFINITIONS "-DENABLE_HDF5=1")
-    add_definitions(${Splash_DEFINITIONS})
-    set(LIBS ${LIBS} ${Splash_LIBRARIES})
-endif(Splash_FOUND)
+add_definitions(-DENABLE_HDF5=1)
```
PrometheusPi: QUESTION:
Why could you remove the if clause?

ax3l: because it's required in `find_package`

etc/picongpu/titan-ornl/picongpu.profile.example
ax3l: it doesn't matter, both won't work and someone looking at it will see to use this version or any newer

buildsystem/CompileSuite/autoTests/new_commits.sh
ax3l: the compile suite tests our *minimal* software requirements, so we know they still work.

individual systems such as hypnos might use newer versions

Flamefire:
Jep, usage of the transition interface of the new libsplash attributes is correct. Note that this does not protect against the wrong datatypes. You could still read the binary data of an int into a float and stuff like that. But at least you will avoid the buffer overflows of the old interface.

PR #2522, opened 2018-02-09
Add a workaround for MSVC bug with capturing constexpr

sbastrakov:
At long last it seems to be the final modification required to build PIConGPU with MSVC.

Due to a bug, MSVC captures constexpr in lambdas as non-constexpr variables, which breaks compilation e.g., if it is used as a template argument in a lambda.

As discussed with @psychocoderHPC and @ax3l I added a macro to be used for such variables instead of constexpr and replace the (very few) problematic cases with it. Also in some cases the only usage was an unnecessary typecast, which I removed.

After some more consideration I chose the name `PMACC_CONSTEXPR_CAPTURE `instead of what we discussed: did not like having `CONSTEXPR_LAMBDA `as it is potentially confusing with a lambda being a constexpr (in C++17), and did not like having MSVC in the name, since the code where it is used is general. Also, in my taste, the chosen name is more direct: it says that this variable is constexpr and will be captured. But ofc it is subjective, could discuss further.

### DIFF
--- a/include/pmacc/fields/operations/AddExchangeToBorder.hpp
+++ b/include/pmacc/fields/operations/AddExchangeToBorder.hpp
@@ -85,7 +85,7 @@ namespace operations
             // number of cells in a superCell
             constexpr uint32_t numCells = pmacc::math::CT::volume< SuperCellSize >::type::value;
             constexpr uint32_t numWorkers = T_numWorkers;
-            constexpr int dim = T_Mapping::Dim;
+            PMACC_CONSTEXPR_CAPTURE int dim = T_Mapping::Dim;
 
             uint32_t const workerIdx = threadIdx.x;
 
--- a/include/pmacc/fields/operations/CopyGuardToExchange.hpp
+++ b/include/pmacc/fields/operations/CopyGuardToExchange.hpp
@@ -82,7 +82,7 @@ namespace operations
             // number of cells in a superCell
             constexpr uint32_t numCells = pmacc::math::CT::volume< SuperCellSize >::type::value;
             constexpr uint32_t numWorkers = T_numWorkers;
-            constexpr int dim = T_Mapping::Dim;
+            PMACC_CONSTEXPR_CAPTURE int dim = T_Mapping::Dim;
 
             uint32_t const workerIdx = threadIdx.x;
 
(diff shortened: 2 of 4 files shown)


ax3l:
@sbastrakov don't worry that the test fails, we need to merge the breaking change in #2520 first.
You can then rebase this PR to trigger the build again.

ax3l:
@sbastrakov ready for rebase against `dev`

sbastrakov:
I rebased after #2520 was merged, checks still fail (or maybe I did not rebase correctly?)

psychocoderHPC:
For our SPEC benchmark we have an internal compiler issue with gcc9 on the same place.

```
boost::mpl::v_item<picongpu::placeholder_definition26::weighting, boost::mpl::v_item<picongpu::placeholder_definition24::momentum, boost::mpl::v_item<picongpu::placeholder_definition21::position<picongpu::placeholder_definition23::position_pic>, boost::mpl::vector0<mpl_::na>, 0>, 0>, 0> >]’
./include/pmacc/particles/tasks/TaskReceiveParticlesExchange.hpp:59:14:   required from here
./include/pmacc/particles/ParticlesBase.kernel:1334:32: internal compiler error: in lookup_template_class_1, at cp/pt.c:9545
 1334 |                     DataSpace< dim > dstSuperCell = DataSpaceOperations < dim - 1 > ::extend(
      |                                ^~~
Please submit a full bug report,
```

only a hint we can maybe follow to understand the behavior better: Since we capture all variables by reference into the lambda can it be that this is not well defined in the C++ standard. If I remember correct than nvcc is also not supporting that you can pass a constexp variable as function argument if the function requires a reference.

jkelling:
Adding `static` to the declaration of `dim` does eliminate the internal compiler error in gcc 9.1.0 and the code compiles.

Still created https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91377 as even if this not well defined, there should not be an ICE here.

ax3l:
why do we even capture by reference `&` and not by copy `=` (which is likely omitted)?

sbastrakov:
@ax3l agree, lots of our default captures by reference can be changed to by value. These values should be only copied if odr-used inside the lambda. Also I feel like we heavily overdo the capture defaults in general, for a couple of variables explicit capture looks more reasonable.

PR #2521, opened 2018-02-08
Hypnos (HZDR): New Modules

ax3l:
Update MPI & HDF5 Modules after today's maintainance.

PR #2523, opened 2018-02-09
fix module name openmpi hypnos

PrometheusPi:
Followup to #2521 - fixes wrong name of openmpi module on hypnos


psychocoderHPC:
This PR is not changing all places, full fix is provided in #2524

PR #2537, opened 2018-02-22
Move TBG to bin/

ax3l:
Start to disassemble the old bin structure. Move `tbg` and `egetopt` to `bin/`.

Profiles:  new `PATH` to `bin/`

Ideally, we will only set a path on `bin/` in the end. But more tools need to move first.

Preparation for #1969


ax3l:
@ComputationalRadiationPhysics/picongpu-developers I try to keep the changes of this cleanup minimal, but after this change you will have to add
```bash
export PATH=$PICSRC/bin:$PATH
```

to your `picongpu.profile` if you are working with `dev`.

(This path will also stay there from now one, others will vanish during clean-ups but this won't affect you if you keep them in.)

PR #2540, opened 2018-02-23
[WIP] CMake: Install PIConGPU Sources

ax3l:
Super (framework) CMake project to install PIConGPU and/or its subprojects.

Implements #1969

*This will take some other PRs to refactor the CMake logic first. I just open this for reference.*

## Description

The CMake logic is as follows.

**the base dir** of the repo will contain a "super" ("framework") project *PIConGPU* `CMakeLists.txt` that select components one wants to install. This will allow to install all sources (as before), or just part, e.g. `PMacc` or `tbg` or `tools`  or ... only.

The `include/pmacc/` dir will contain a regular `CMakeLists.txt` that can either be `add_subdirectory`-ed for usage or installed on which it creates a proper `PMaccConfig.cmake`. For that I will move logic from the config to the first file, where it belongs (it's auto-included on installed PMacc).

Sub-projects such as `mpiInfo`, `png2gas`, `splash2txt` and `cuda_memtest` can be build on install. If not, they are as today build on PIConGPU application build time, via `add_subdirectory`.

The PIConGPU application itself, CMake project name `PIConGPUapp` is in `include/picongpu/` as usual. It contains the nifty logic we use so far to create a binary from an input directory. It's "source install logic" ("target") would go in the base `CMakeLists.txt` instead of the `include/picongpu/CMakelists.txt` since the latter is used for the described "user-workflow" as usual.

All in all, this will not change anything for the user but will allow installing PIConGPU sources, sub-projects of the PIConGPU repo and pre-compiling sub-projects if wanted.

## Resources

- https://www.youtube.com/watch?v=bsXLMQ6WgIk (`add_subdirectory` for shipped vs. `find_package` for external dependency in minute 53+)

## To Do

- [ ] PMacc install (`include/pmacc/`)
  - move logic from `PMaccConfig.cmake` to `CMakeLists.txt`
  - add targets and installer that generates config
- [ ] PIConGPU app build (`include/picongpu/CMakeLists.txt`)
  - add tons of `PIConGPUapp_USE_...` options, e.g. for internal/external libs
  - use `add_subdirectory` for *internal PMacc*
- [ ] tools `splash2txt`, `png2gas`, `mpiInfo`
  - modernize files and add proper install logic

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,206 @@
+# Copyright 2018 Axel Huebl
+#
+# This file is part of PIConGPU.
+#
+# PIConGPU is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# PIConGPU is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PIConGPU.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
+# Preamble ####################################################################
+#
+cmake_minimum_required(VERSION 3.10.0)
+
+project(PIConGPU VERSION 0.1.0)
+
+
+# Project structure ###########################################################
+#
+# temporary build directories
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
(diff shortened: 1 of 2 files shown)


ax3l:
@j-stephan and @bernhardmgruber if any of you wants to apply your CMake skills to this PRs effort, please do not hesitate to take over :)

bernhardmgruber:
@ax3l thank you for the honoring mentioning ;) As far as I can see however, the CMakeLists.txt you added concerns itself solely with installing, in which I am a novice as well :/.

I have stumbled over a great tutorial recently for header-only libraries. I guess this can give some input: https://dominikberner.ch/cmake-interface-lib/
But in the end, I guess you still need to play around a bit.

ax3l:
Closing this PR since I am not able to continue on it in any reasonable time frame.

Please everyone feel free to use from it what you need :)

PR #2546, opened 2018-03-09
DGX Docker

ax3l:
Contains two bug fixes for running a container smoothly on DGX-1.

- TBG bash template: support up to 8 GPUs per node

Usually, one would do one-node runs with those anyway or would need to adjust it by hand of multi-node is desired. With this setting, using a full 8-GPU DGX-1 node is possible without hacks.

- OpenMPI root execution fix and mpiexec as well

Make the root work-around more robust via an export (`export -f mpi...`) and include mpiexec as well.

Also does not not build cuda_memtest on containers anymore since it wastes time on startup for demos.

### DIFF
--- a/etc/picongpu/bash/mpiexec.tpl
+++ b/etc/picongpu/bash/mpiexec.tpl
@@ -24,8 +24,8 @@
 .TBG_author=${MY_NAME:+--author \"${MY_NAME}\"}
 .TBG_profile=${PIC_PROFILE:-"~/picongpu.profile"}
 
-# 4 gpus per node if we need more than 4 gpus else same count as TBG_tasks
-.TBG_gpusPerNode=$(if [ $TBG_tasks -gt 4 ] ; then echo 4; else echo $TBG_tasks; fi)
+# 8 gpus per node if we need more than 8 gpus else same count as TBG_tasks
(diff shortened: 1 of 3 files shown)


ax3l:
@psychocoderHPC tested and runs. can you please review and merge? :)

etc/picongpu/bash/mpiexec.tpl
```diff
@@ -24,8 +24,8 @@
 .TBG_author=${MY_NAME:+--author \"${MY_NAME}\"}
 .TBG_profile=${PIC_PROFILE:-"~/picongpu.profile"}
 
-# 4 gpus per node if we need more than 4 gpus else same count as TBG_tasks
-.TBG_gpusPerNode=$(if [ $TBG_tasks -gt 4 ] ; then echo 4; else echo $TBG_tasks; fi)
+# 8 gpus per node if we need more than 8 gpus else same count as TBG_tasks
```
psychocoderHPC: Our old number `4'and our new number `8` is a magic number which is randomly chosen. We should think about a more generic way (I have currently no idea).

ax3l: maybe we could just do
`ls -l /dev/nvidia* | wc -l` or so

PR #2547, opened 2018-03-10
cupla: no .gitmodules

ax3l:
Remove the .gitmodules file from cupla.  We ship alpaka ourselves and it can confuse git clone steps to have it in there.

First seen as an issue with ReadTheDocs builds (activated by default [since November last year](https://github.com/rtfd/readthedocs-build/issues/30)).

### DIFF
--- a/thirdParty/cupla/.gitmodules
+++ b/thirdParty/cupla/.gitmodules
@@ -1,4 +1,0 @@
-[submodule "alpaka"]
-	path = alpaka
-	url = git://github.com/ComputationalRadiationPhysics/alpaka.git
-	branch = develop


thirdParty/cupla/.gitmodules
```diff
@@ -1,4 +0,0 @@
-[submodule "alpaka"]
-	path = alpaka
-	url = git://github.com/ComputationalRadiationPhysics/alpaka.git
```
ax3l: @psychocoderHPC you might also want to consider adding this url to a https:// url on cupla itself.
Otherwise, anonymous git clones will fail for cupla.

psychocoderHPC:
I am confused because you change a thridt party module. We never change this directly in our project.

ax3l:
Yes I know, but git subtree is really that simple and allows that :)

ax3l:
If we forget it on an update and RTD did not expose the linked issue as an option by then, the badge in the readme will tell us

psychocoderHPC:
offline discussed: this is a temporary solution. In general we should not update our thrid party dependencies here.

PR #2550, opened 2018-03-13
Simpler Python Parameter class

codingS3b:
This PR gets rid of `UiParameter`, `LinearScaledParameter` and `LogScaledParameter` and instead introduces a single `Parameter` class which has the option to pass in arbitrary conversion functions to handle different scales on the UI and on the internal PIC scale of each parameter.
It also makes use of the `pint` package for easier unit conversion.

### DIFF
--- a/lib/python/picongpu/input/parameters.py
+++ b/lib/python/picongpu/input/parameters.py
@@ -28,7 +28,6 @@ class Parameter(object):
     user adjusts time in nanoseconds but internally the precise number is
     needed (--> linear conversion)
     )
-
     Also the kind of widget representation should depend on the attributes of
     this class. Either as MultiRangeSlider or as MultiSelection or as Checkbox
     depending on the 'values' (discrete) or 'range' (continuous)
@@ -57,9 +56,11 @@ class Parameter(object):
             The default value for this parameter (on the UI scale) used when\
             no 'values' or 'range' parameter is passed.
         values: list
-            list of discrete options for the parameter as shown on UI side
+            list of discrete options for the parameter as shown on UI side.
+            Only one of the attributes range/value can be given.
         range: tuple
-            start and stop value for the selectable range on UI side
+            start and stop value for the selectable range on UI side.
+            Only one of the attributes range/value can be given.
         label: string [optional]
             Overwrite the name for UI
         pic_to_SI: callable, e.g. lambda function
@@ -102,7 +103,17 @@ class Parameter(object):
 
     def _check_input(self, vals):
         """
-        Assumes vals are in UI units, i.e. with unit = self.unit
+        For values that are assumed to be on the UI scale (i.e. with unit =
+        self.unit), checks whether they are in the allowed range or the
+        allowed discrete values.
+        Raises a ValueError if a value value outside the range is detected.
+
+        Parameters
+        ----------
+        vals: float or list of floats
+            values on the parameters UI scale which will be checked.
+
+
         """
         if self.values is not None:
             # check for valid values
@@ -122,6 +133,18 @@ class Parameter(object):
         """
         Takes values in UI units, converts them to SI and after that
         to quantities used within PIConGPU.
+
+        Parameters
+        ----------
+        vals: float or list of floats
+            values that will be converted to the PIC scale
+        check_vals: bool
+            Flag to decide whether given values are within the allowed
+            range for this parameters UI scale.
+
+        Returns
+        -------
+        A list of converted values.
         """
         # todo check if vals are in specified range
         if check_vals:
@@ -134,6 +157,19 @@ class Parameter(object):
         """
         Takes PIC values and returns values on UI scale after converting to SI
         values as intermediate step.
+
+        Parameters
+        ----------
+        vals: float or list of floats
+            values that will be converted to the UI scale.
+
+        check_vals: bool
+            Flag to decide whether converted values are within the allowed
+            range for this parameters UI scale.
+
+        Returns
+        -------
+        A list of converted values.
         """
         # v is given in PIC quantity, so we convert to UI unit
         ui_results = [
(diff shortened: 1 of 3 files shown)


lib/python/picongpu/input/parameters.py
```diff
@@ -5,17 +5,41 @@
...
+import pint
+ureg = pint.UnitRegistry()
 
 
 class Parameter(object):
     """
-    Base class for exposable parameters.
+    Parameter that can have a corresponding representation as a jupyter widget
+    within interactive jupyter notebooks.
+    The widget allows the user to adjust values as he likes but this class
```
ax3l: he/she

lib/python/picongpu/input/parameters.py
```diff
@@ -5,17 +5,41 @@
...
+    precise number is needed (--> log scale conversion)
+    or
+    user adjusts time in nanoseconds but internally the precise number is
+    needed (--> linear conversion)
+    )
+    Also the kind of widget representation should depend on the attributes of
+    this class. Either as MultiRangeSlider or as MultiSelection or as Checkbox
+    depending on the 'values' (discrete) or 'range' (continuous)
+
+    Stepsizes will be handled by the widget representation, not here.q
```
ax3l: ~~q~~

lib/python/picongpu/input/parameters.py
```diff
@@ -25,417 +49,155 @@ def __init__(self, name, ptype, unit, default, value=None, dtype=None):
...
-            The value of the parameter object. If it is not provided, the value
-            of the 'default' argument is used.
-        dtype: type (optional)
-            The data type of the parameter value. If it is not provided, the
-            type of the 'default' argument is used.
+            The measurement unit of this parameter as shown on the UI side.
+            Implicitely we assume the pic unit is SI. If this is not possible,
+            the converter needs to translate those scales.
+        default: float, int or string
+            The default value for this parameter (on the UI scale) used when\
```
ax3l: \ ?

lib/python/picongpu/input/parameters.py
```diff
@@ -25,417 +49,155 @@ def __init__(self, name, ptype, unit, default, value=None, dtype=None):
...
+        values: list
+            list of discrete options for the parameter as shown on UI side.
+            Only one of the attributes range/value can be given.
+        range: tuple
+            start and stop value for the selectable range on UI side.
+            Only one of the attributes range/value can be given.
+        label: string [optional]
+            Overwrite the name for UI
+        pic_to_SI: callable, e.g. lambda function
+            Specifies how to transform a value on the pic scale to an\
```
ax3l: \ ?

lib/python/picongpu/input/parameters.py
```diff
@@ -25,417 +49,155 @@ def __init__(self, name, ptype, unit, default, value=None, dtype=None):
...
+            # raise ValueError("Need either 'values' or 'range' parameter!")
+            self.values = [self.default]
+            print("WARNING: Neither 'values' nor 'range' was given, setting"
+                  " 'values' to ", self.values)
+        if values is not None:
+            if not isinstance(values, collections.Iterable):
+                self.values = [values]
+            if not self.values:
+                self.values = [self.default]
+                print("WARNING: Values attribute can not be an empty\
```
ax3l: I think
```python
print("WARNING: Values attribute can not be an empty"
      "iterable! Settting values to: ", self.values)
```

should work

lib/python/picongpu/input/parameters.py
```diff
@@ -25,417 +49,155 @@ def __init__(self, name, ptype, unit, default, value=None, dtype=None):
...
-        Returns
-        -------
-        Whatever the set_value function returns.
         """
+        if self.values is not None:
+            # check for valid values
+            res = all([v in self.values for v in vals])
+            if not res:
+                raise ValueError(
+                    "Invalid values found! Values should be elements of\
```
ax3l:
```python
"Invalid values found! Values should be elements of "
"self.values!")

File: lib/python/picongpu/input/parameters.py [hunk 7]
```diff
@@ -25,417 +49,155 @@ def __init__(self, name, ptype, unit, default, value=None, dtype=None):
...
+                     self.values!")
+        else:
+            # check for valid range
+            res = all([self.range[0] <= v <= self.range[1] for v in vals])
+            if not res:
+                raise ValueError("Invalid values found! Values should be\
```
ax3l: see above

lib/python/picongpu/input/parameters.py
```diff
@@ -25,417 +49,155 @@ def __init__(self, name, ptype, unit, default, value=None, dtype=None):
...
-        scale_factor: float
-            The factor for the linear transformation of slider ui value to
-            internal value.
+        A list of converted values.
         """
-        self.scale_factor = scale_factor
-        # since base class UiParameter now calls the set_value() to carry out
-        # linear scaling of the value, we need to define the scale_factor
-        # before entering the constructor since it is used in set_value()
+        # todo check if vals are in specified range
```
ax3l: open todo :-)

codingS3b:
Thanks for reviewing, I fixed the linebreaks and incorporated the other changes.
Would it be alright to also include the updated params.py for LWFA example in this PR that already uses the new `Parameter` class?

ax3l:
Good point, go ahead and keep it in consistent state.

lib/python/picongpu/input/parameters.py
```diff
@@ -32,8 +32,7 @@ class Parameter(object):
     this class. Either as MultiRangeSlider or as MultiSelection or as Checkbox
     depending on the 'values' (discrete) or 'range' (continuous)
 
-    Stepsizes will be handled by the widget representation, not here.q
-
+    Stepsizes will be handled by the widget representation, not here.
```
ax3l: ups, step sizes is two words

ax3l: Probably "stepping" is easier to understand?

codingS3b: Agreed.

PR #2555, opened 2018-03-21
Free Density: Fix float_X

ax3l:
Properly stay with float_X for the normalized density profile (usually between 0. and 1.)

### DIFF
--- a/include/picongpu/param/density.param
+++ b/include/picongpu/param/density.param
@@ -266,13 +266,13 @@ namespace densityProfiles
 
             /* triangle function example
              * for a density profile from 0 to 400 microns */
-            float_64 s = 1.0 - 5.0 * math::abs(y - 0.2);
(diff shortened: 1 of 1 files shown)


include/picongpu/param/density.param
```diff
@@ -266,13 +266,13 @@ namespace densityProfiles
 
             /* triangle function example
              * for a density profile from 0 to 400 microns */
-            float_64 s = 1.0 - 5.0 * math::abs(y - 0.2);
+            float_X s = float_X( 1.0 ) - float_X( 5.0 ) * math::abs(y - 0.2);
```
psychocoderHPC: depending of which precision is needed you must cast the result of `abs` or the param of `abs`

ax3l: we really need literals, e.g. a `_X`  and `_3X` :-)
http://en.cppreference.com/w/cpp/language/user_literal

Issue #2556, opened 2018-03-22
Wrong number of the entries of 5p orbital in Gold 


Banjafar:
While I checked my own example I found an issue in 5p orbital of Gold in `ionizer.param`.

```c++
       /* 5p */
        25.170,
        25.170,
        25.170,
        25.170,
        25.170,
        25.170,
        25.170,
```
There should be six elements but they are 7.

ax3l:
@n01r can you take a look?

n01r:
Already had a look, he told me about the issue, we had a look and I asked him to write an issue.
Yeah. There are really too many 5p entries in the gold array. I'll write a fix.

n01r:
This one can be closed, right? - Cannot do it myself, tho :3

PR #2558, opened 2018-03-22
Fixed excess 5p shell entry in gold effective Z

n01r:
`ionizer.param`:
* the effective Z array contained too many 5p orbital entries   
      --> fix: delete 1 value and go to 6
* changed the configuration of Cu and Au to their real-life   
      ground state configuration which slightly differs from the   
      Madelung rule prediction
* updated documentation on shell occupation

### DIFF
--- a/include/picongpu/param/ionizer.param
+++ b/include/picongpu/param/ionizer.param
@@ -173,90 +173,95 @@ namespace effectiveNuclearCharge
 {
     /* For hydrogen Z_eff is obviously equal to Z */
     PMACC_CONST_VECTOR(float_X, 1, Hydrogen,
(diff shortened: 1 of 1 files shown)


n01r:
This pull request closes #2556.

PrometheusPi:
@Banjafar great catch. 
Has anyone used gold before?

n01r:
I don't think so. Apart from the `Bremsstrahlung` example it has not been used before, to my knowledge ... and there the target was pre-ionized. :)

ax3l:
True, still it's in released versions so I marked it properly.
this will ensure it will go into changelogs and backport-releases.

PR #2561, opened 2018-03-27
fix zero size transform writes

psychocoderHPC:
Fix that particles dumped with adios (tranform layer enabled, MPI aggregate enabled) causing reading errors.
Solved by enable compression for all MPI ranks even if the size of data which must be written is zero.

issue: https://github.com/ornladios/ADIOS/issues/170

requires a post-1.13.0 ADIOS for the following fixes:
- reads:
  - https://github.com/ornladios/ADIOS/pull/162
  - https://github.com/ornladios/ADIOS/pull/163
- writes:
  - zlib https://github.com/ornladios/ADIOS/pull/165
  - bzip2, isobar and zfp https://github.com/ornladios/ADIOS/issues/174

Cc-ing: @n01r

### DIFF
--- a/include/picongpu/plugins/adios/ADIOSWriter.hpp
+++ b/include/picongpu/plugins/adios/ADIOSWriter.hpp
@@ -109,16 +109,6 @@ int64_t defineAdiosVar(int64_t group_id,
                        bool compression,
                        std::string compressionMethod)
 {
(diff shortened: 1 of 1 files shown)


ax3l:
added note to use a post-1.13.0 ADIOS to #2562

PR #2575, opened 2018-04-11
Doxygen: Publish As Well 

ax3l:
Let's publish our doxygen builds on [`gh-pages`](http://computationalradiationphysics.github.io/picongpu).

This adds the instructions on how to update it. Doxygen docs are not versioned as sphinx docs are, so we should decide when we update the docs to keep possible confusion small.


ax3l:
Dang, my Doxygen `1.8.13` *segfaults* on our HTML docs o.0

ax3l:
@psychocoderHPC I just leave you in FYI because you wished for that feature :)

PR #2576, opened 2018-04-12
Radiation Plugin: Namespace

ax3l:
Add at least the basic picongpu namespace.

Later on, needs proper namespace per directory.

Seen as [polluted scope in Doxygen docs](https://computationalradiationphysics.github.io/picongpu/annotated.html).

### DIFF
--- a/include/picongpu/plugins/radiation/nyquist_low_pass.hpp
+++ b/include/picongpu/plugins/radiation/nyquist_low_pass.hpp
@@ -17,12 +17,15 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
-
 #pragma once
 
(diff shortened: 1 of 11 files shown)


include/picongpu/plugins/radiation/nyquist_low_pass.hpp
```diff
@@ -17,13 +17,16 @@
...
 
-
 #pragma once
 
 #include "calc_amplitude.hpp"
 #include "parameters.hpp"
 #include "particle.hpp"
 
+
+namespace picongpu
```
PrometheusPi: This leads to a failure with the use of `picongpu` in the code, there `::picongpu` or the namespace should be removed should be used.

PR #2594, opened 2018-05-03
mallocMC.param: file doxygen

ax3l:
Add a file doxygen string to the mallocMC input file. Will show up automatically in the manual.

Related to #1982.

### DIFF
--- a/include/picongpu/param/mallocMC.param
+++ b/include/picongpu/param/mallocMC.param
@@ -20,9 +20,9 @@
 
 /** @file
  *
- * Fine-tuning of the particle heap for GPUs. When running on GPUs, we use a
- * high-performance "new" allocator (mallocMC) for that can be parametrized
- * here.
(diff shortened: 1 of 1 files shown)


include/picongpu/param/mallocMC.param
```diff
@@ -18,36 +18,47 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
+/** @file
+ *
+ * Fine-tuning of the particle heap for GPUs. When running on GPUs, we use a
```
PrometheusPi: First sentence is no sentence, use `:` instead of `.`

include/picongpu/param/mallocMC.param
```diff
@@ -18,36 +18,47 @@
...
+ * high-performance "new" allocator (mallocMC) for that can be parametrized
```
PrometheusPi: Here a `which` appears to be missing or the `for` is too much

PR #2599, opened 2018-05-14
HDF5 plugin: fix memory leak

psychocoderHPC:
Fix missing free for ressources in HDF5 plugin.
The result can be that the simulation is running out of memory.

### DIFF
--- a/include/picongpu/plugins/hdf5/HDF5Writer.hpp
+++ b/include/picongpu/plugins/hdf5/HDF5Writer.hpp
@@ -524,10 +524,12 @@ private:
 
     void closeH5File()
     {
-        if (mThreadParams.dataCollector != nullptr)
+        if (mThreadParams.dataCollector)
         {
             log<picLog::INPUT_OUTPUT > ("HDF5 close DataCollector");
             mThreadParams.dataCollector->close();
+            mThreadParams.dataCollector->finalize();
+            __delete(mThreadParams.dataCollector);
         }
     }
 


PrometheusPi:
@psychocoderHPC Could that explain the freeze of PIConGPU on taurus while writing hdf5 data?

include/picongpu/plugins/hdf5/HDF5Writer.hpp
```diff
@@ -524,10 +524,12 @@ class HDF5Writer :
 
     void closeH5File()
     {
-        if (mThreadParams.dataCollector != nullptr)
+        if (mThreadParams.dataCollector)
         {
             log<picLog::INPUT_OUTPUT > ("HDF5 close DataCollector");
             mThreadParams.dataCollector->close();
+            mThreadParams.dataCollector->finalize();
```
ax3l: Did you RT test this?

We close files after every write. But `finalize` will [free the communicator](https://github.com/ComputationalRadiationPhysics/libSplash/blob/v1.7.0/src/ParallelDataCollector.cpp#L203-L211) that was given to the `ParallelDataCollector` in the constructor. We only generate that communicator once during a run, since it's a potentially expensive MPI operation.

We do call `finalize()` during clean shutdown in `unload()`

psychocoderHPC:
OK I did not find the de-allocation for `mThreadParams.dataCollector = new ParallelDomainCollector`. The memory is freed in the destructor of `HDF5Writer`

PR #2602, opened 2018-05-17
`Vector`: add method `remove<...>()`

psychocoderHPC:
Add method to remove a component from a vector.

# Example
```C++
pmacc::Vector< int, 3 > foo( 0, 1, 2 );
pmacc::Vector< int, 2 > bar = foo.remove< 1 >( );
std::cout << "bar: " << bar.toString() << std::endl;
// bar: {0, 2}
```

### DIFF
--- a/include/pmacc/math/vector/Vector.hpp
+++ b/include/pmacc/math/vector/Vector.hpp
@@ -304,6 +304,28 @@ struct Vector : private T_Storage<T_Type, T_dim>, protected T_Accessor, protecte
         return result;
     }
 
+    /** Removes a component
+     *
+     * It is not allowed to call this method on a vector with the dimensionality of one.
+     *
+     * @tparam dimToRemove index which shall be removed; range: [ 0; dim - 1 ]
+     * @return vector with `dim - 1` elements
+     */
+    template<int dimToRemove>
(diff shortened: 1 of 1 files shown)


include/pmacc/math/vector/Vector.hpp
```diff
@@ -304,6 +304,28 @@ struct Vector : private T_Storage<T_Type, T_dim>, protected T_Accessor, protecte
         return result;
     }
 
+    /** removes a component from the vector
+     *
+     * It is not allowed to call this method on a vector with the dimensionality of one.
+     *
+     * @tparam dimToRemove index of the dimension which should be removed; range: [ 0; dim - 1 ]
```
ax3l: little suggestions:
- let us remove ~~of the dimension~~
- should -> shall

include/pmacc/math/vector/Vector.hpp
```diff
@@ -304,6 +304,28 @@ struct Vector : private T_Storage<T_Type, T_dim>, protected T_Accessor, protecte
         return result;
     }
 
+    /** removes a component from the vector
```
ax3l: Remove a component

is enough for the title

psychocoderHPC:
updated

PR #2605, opened 2018-05-17
expose random number method to the user

psychocoderHPC:
Remove the hard coded `XorMin` method with the user changeable `RandomGenerator` from `random.param`

- rename `seed.param` into `random.param`
- remove unused seeds for particles, temperatures, ...
- add type definition `RandomGenerator` (can be changed by the user)

- [x] merge after #2604 (included in this PR as first commit)

PR #2607, opened 2018-05-18
seed generator for RNG

psychocoderHPC:
In some cases it is useful to initialize the random number generator
always with the same value to increase the reproducibility of different
simulation runs. IN some cases the opposit is required.

- re-enable the possibility to choose the seed
- add different seed generators:
  - seed::Value<...>
  - seed::FromEnvironment
  - seed::FromTime

**The seed generators can be moved into `PMacc` as soon as we remove the requirement that PMacc must be header only.**

# Todo

- [x] rebase against #2605 
- [x] rebase against #2608 to solve compile issues

### DIFF
--- a/include/picongpu/random/seed/Seed.cpp
+++ b/include/picongpu/random/seed/Seed.cpp
@@ -1,0 +1,67 @@
+/* Copyright 2018 Rene Widera
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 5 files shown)


include/picongpu/random/seed/Seed.cpp
```diff
@@ -0,0 +1,67 @@
...
+        uint32_t now_ms = std::chrono::time_point_cast< std::chrono::milliseconds >( now ).
+            time_since_epoch().count();
+
+        // receive time from rank zero
+        MPI_Bcast(
+            &now_ms,
+            1,
+            MPI_UINT32_T,
+            0,
+            MPI_COMM_WORLD
```
ax3l: probably a bit saver / more flexible if we took the MPI communicator and root rank from the `GridController` but ok for now.

PR #2610, opened 2018-05-23
(compile issue) fix random seed

psychocoderHPC:
It is currently not possible to compile PIConGPU due to a compiler error (Tested with gcc 5.3.0):
```
include/picongpu/random/seed/Seed.cpp:60:20: error: ‘stoi’ is not a member of ‘std’
             seed = std::stoi( seedStr );
```

- add missing include `<string>`

This bug was introduced with #2607


PrometheusPi:
How did this end up in the `dev` despite the compile suite? 
The default LWFA example does not compile on hypnos (k80). (with default `picongpu.profile`)

PR #2611, opened 2018-05-24
ISAAC 1.3.3+

ax3l:
Since ISAAC 1.3.0 forgot to bump the version define, CMake will not find it with required minimum versions.

1.3.1 fixed that.

Update: there seem to be more issues with ISAAC... hard to figure out after a year without a changelog and PRs. We will go for the latest release 1.3.3+.

ISSAC `${ISAAC_INCLUDE_DIRS}` is even in 1.3.3 broken and points to CMake dir instead of include dir: https://github.com/ComputationalRadiationPhysics/isaac/issues/68

Update: work-arounded both in spack recipe of PIConGPU and compile suite ISAAC module by explicitly setting `CPATH` to ISAAC include dir...

Going for ISAAC 1.3.3 since we fixed some boost and CUDA 9 issues AFAIR.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -271,7 +271,7 @@ ADIOS
 
 ISAAC
 """""
-- 1.3.1+
+- 1.3.3+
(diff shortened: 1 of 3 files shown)


ax3l:

PR #2612, opened 2018-05-24
PMacc: math::CT::max improvement

psychocoderHPC:
- add specialization to get the element with the maximum value
- add documentation

PR #2615, opened 2018-05-24
optimize stride size for update FieldJ

psychocoderHPC:
The stride size for `StrideMapping` depends on the supercell size and the stencil margins.
In the case where the lower and upper margin fits into one supercell we need only a stride of 2 instead of the current used default stride 3.
If we use stencils with margins larger than one supercell the default 3 can be to low and will produce data race conditions (therefore this PR is also a fix for big stencil margins).

- calculate optimal stride size (skip only as much supercells as needed to avoid race conditions for cell updates)

- [x] depends on #2612

### DIFF
--- a/include/picongpu/fields/FieldJ.tpp
+++ b/include/picongpu/fields/FieldJ.tpp
@@ -278,7 +278,28 @@ void FieldJ::computeCurrent( ParticlesClass &parClass, uint32_t )
         typename GetMargin<ParticleCurrentSolver>::UpperMargin
     > BlockArea;
 
-    StrideMapping<AREA, 3, MappingDesc> mapper( cellDescription );
+    /* The needed stride for the stride mapper depends on the stencil width.
+     * If the upper and lower margin of the stencil fits into one supercell
+     * a double checker board (stride 2) is needed.
+     * The round up sum of margins is the number of supercells to skip.
+     */
+    using MarginPerDim = typename pmacc::math::CT::add<
+        typename GetMargin<ParticleCurrentSolver>::LowerMargin,
+        typename GetMargin<ParticleCurrentSolver>::UpperMargin
+    >::type;
+    using MaxMargin = typename pmacc::math::CT::max< MarginPerDim >::type;
+    using SuperCellMinSize = typename pmacc::math::CT::min< SuperCellSize >::type;
+
+    /* number of supercells which must be skipped to avoid overlapping areas
+     * between different blocks in the kernel
+     */
+    constexpr uint32_t skipSuperCells = ( MaxMargin::value + SuperCellMinSize::value - 1u ) / SuperCellMinSize::value;
+    StrideMapping<
+        AREA,
+        skipSuperCells + 1u, // stride 1u means each supercell is used
+        MappingDesc
+    > mapper( cellDescription );
+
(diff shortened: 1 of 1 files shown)


include/picongpu/fields/FieldJ.tpp
```diff
@@ -278,7 +278,24 @@ void FieldJ::computeCurrent( ParticlesClass &parClass, uint32_t )
...
+     * between different blocks in the kernel
+     */
+    constexpr uint32_t skipSuperCells = (MaxMargin::value + SuperCellMinSize::value - 1) / SuperCellMinSize::value;
+    StrideMapping<
+        AREA,
+        skipSuperCells + 1, // stride 1 means each supercell is used
+        MappingDesc
+    > mapper( cellDescription );
+
+    std::cout<<skipSuperCells<<std::endl;
```
ax3l: leftover debug :)

include/picongpu/fields/FieldJ.tpp
```diff
@@ -278,7 +278,24 @@ void FieldJ::computeCurrent( ParticlesClass &parClass, uint32_t )
...
+    using MaxMargin = typename pmacc::math::CT::max< MarginPerDim >::type;
+    using SuperCellMinSize = typename pmacc::math::CT::min< SuperCellSize >::type;
+
+    /* number of supercells which must be skipped to avoid overlapping areas
+     * between different blocks in the kernel
+     */
+    constexpr uint32_t skipSuperCells = (MaxMargin::value + SuperCellMinSize::value - 1) / SuperCellMinSize::value;
+    StrideMapping<
+        AREA,
+        skipSuperCells + 1, // stride 1 means each supercell is used
```
ax3l: 1u

psychocoderHPC:
ready to merge

include/picongpu/fields/FieldJ.tpp
```diff
@@ -278,7 +278,23 @@ void FieldJ::computeCurrent( ParticlesClass &parClass, uint32_t )
         typename GetMargin<ParticleCurrentSolver>::UpperMargin
     > BlockArea;
 
-    StrideMapping<AREA, 3, MappingDesc> mapper( cellDescription );
+    using MarginPerDim = typename pmacc::math::CT::add<
```
ax3l: you will need to add a comment here to describe the logic (see PR description).

include/picongpu/fields/FieldJ.tpp
```diff
@@ -278,7 +278,23 @@ void FieldJ::computeCurrent( ParticlesClass &parClass, uint32_t )
...
+        typename GetMargin<ParticleCurrentSolver>::LowerMargin,
+        typename GetMargin<ParticleCurrentSolver>::UpperMargin
+    >::type;
+    using MaxMargin = typename pmacc::math::CT::max< MarginPerDim >::type;
+    using SuperCellMinSize = typename pmacc::math::CT::min< SuperCellSize >::type;
+
+    /* number of supercells which must be skipped to avoid overlapping areas
+     * between different blocks in the kernel
+     */
+    constexpr uint32_t skipSuperCells = (MaxMargin::value + SuperCellMinSize::value - 1) / SuperCellMinSize::value;
```
ax3l: `- 1u`

PR #2622, opened 2018-06-05
float_X: C++11 User Literal

ax3l:
Define `_X` (and `_32` and `_64`) for the verbose `float_X( ... )` casts we put everywhere.

Refactors the particle shapes to try the newly defined literals.

PR #2625, opened 2018-06-06
Params: float_X Constants to Literals

ax3l:
Increase readability of input `.param` files by using C++11 literals for constant numbers with suffix `_X` instead of `float_X( ... )`.

- [x] rebase after #2622


include/picongpu/param/density.param
```diff
@@ -266,7 +266,7 @@ namespace densityProfiles
 
             /* triangle function example
              * for a density profile from 0 to 400 microns */
-            float_X s = float_X( 1.0 - 5.0 * math::abs( y - 0.2 ) );
+            float_X s = 1.0_X - 5.0_X * math::abs( y - 0.2_X ) );
```
psychocoderHPC: the last `)` is to much

PR #2628, opened 2018-06-11
Limit Supported GCC with nvcc 8.0-9.1

### LINKED ISSUE #2388: Boost::System 1.65.1, CUDA 8.0.61
Compiling PIConGPU with
- CUDA 8.0(.61)
- Boost 1.65.1
- GCC 5.4.0

in the latest `dev` as of 6146de12610d5c984f59b17c0792509686716b65 (already contains `--expt-relaxed-constexpr`) fails with:

```
include/boost/system/error_code.hpp(247): error: exception specification for virtual function "boost::system::error_category::std_category::equivalent(const std::error_code &, int) const" is incompatible with that of overridden function "std::_V2::error_category::equivalent(const std::error_code &, int) const"

include/boost/system/error_code.hpp(245): error: exception specification for virtual function "boost::system::error_category::std_category::equivalent(int, const std::error_condition &) const" is incompatible with that of overridden function "std::_V2::error_category::equivalent(int, const std::error_condition &) const"

include/boost/system/error_code.hpp(243): error: exception specification for virtual function "boost::system::error_category::std_category::default_error_condition" is incompatible with that of overridden function "std::_V2::error_category::default_error_condition"

include/boost/system/error_code.hpp(233): error: exception specification for virtual function "boost::system::error_category::std_category::name" is incompatible with that of overridden function "std::_V2::error_category::name"
```

ax3l:
Sadly, the truly working GCC range is smaller than the official CUDA support. This is caused by missing support of `std::tuple` which is pulled easily in dependencies such as boost.

The latest NVCC 9.2 is supposedly good as it adds official `std::tuple` support and even support for GCC 7.

Close #2388

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -31,7 +31,11 @@ Mandatory
 
 gcc
 """
-- 4.9 to 5.X (depends on your current `CUDA version <https://gist.github.com/ax3l/9489132>`_)
+- 4.9 - 7 (if you want to build of Nvidia GPUs, supported compilers depend on your current `CUDA version <https://gist.github.com/ax3l/9489132>`_)
+
+  - CUDA 8.0: Use gcc 4.9
+  - CUDA 9.0 - 9.1: Use gcc 4.9 - 5
+  - CUDA 9.2: Use gcc 4.9 - 7
 - *note:* be sure to build all libraries/dependencies with the *same* gcc version
 - *Debian/Ubuntu:*
   
--- a/include/pmacc/PMaccConfig.cmake
+++ b/include/pmacc/PMaccConfig.cmake
@@ -322,6 +322,25 @@ if( ("${ALPAKA_CUDA_COMPILER}" STREQUAL "nvcc") AND
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DBOOST_NO_CXX11_NOEXCEPT")
 endif()
 
+# GCC's C++11 tuples are broken in "supported" NVCC versions
+if("${ALPAKA_CUDA_COMPILER}" STREQUAL "nvcc")
+    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+        if(CUDA_VERSION VERSION_EQUAL 8.0)
+            if(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 5.0)
+                message(FATAL_ERROR "NVCC 8.0 does not support the std::tuple "
+                        "implementation in GCC 5+. Please use GCC 4.9!")
+            endif()
+        elseif(
+            (CUDA_VERSION VERSION_EQUAL 9.0) OR
+            (CUDA_VERSION VERSION_EQUAL 9.1))
+            if(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 6.0)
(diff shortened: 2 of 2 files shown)


ax3l:
Reflected in spack recipe via https://github.com/ComputationalRadiationPhysics/spack-repo/commit/0927bfbfe8821c56474426c62f3504c28b906aa9

BenjaminW3:
We should add those checks to alpaka as well (or only there). And update the alpaka readme.

ax3l:
Absolutely!
Yes, we can remove it if we place it in alpaka directly, but since I am finalizing the release for PIConGPU 0.4.0 right now (hopefully) and we are based on Alpaka 0.3.1 I will merge it in PIConGPU for now as well.

INSTALL.rst
```diff
@@ -31,7 +31,11 @@ Mandatory
 
 gcc
 """
-- 4.9 to 5.X (depends on your current `CUDA version <https://gist.github.com/ax3l/9489132>`_)
+- 4.9 to 7 (if you want to build of Nvidia GPUs, supported compilers depend on your current `CUDA version <https://gist.github.com/ax3l/9489132>`_)
+
+  - CUDA 8.0: Use gcc 4.9
+  - CUDA 9.0-9.1: Use gcc 4.9-5
```
theZiz: I know, this is nit-picking, but I just read your change and thought CUDA 9 needs `gcc 4.9-5` as in "the fifth bug fix release of `gcc 4.9`, which surprised me. :laughing: Probably because I am used to such version numbers from Debian... :roll_eyes:  Could you plese add some spaces to make it more clear? :wink:

ax3l: you are right, that makes a lot of sense :+1: updated.

ax3l:
lovely and sad at the same time :hushed:

include/pmacc/PMaccConfig.cmake
```diff
@@ -322,6 +322,25 @@ if( ("${ALPAKA_CUDA_COMPILER}" STREQUAL "nvcc") AND
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DBOOST_NO_CXX11_NOEXCEPT")
 endif()
 
+# GCC's C++11 tuples are broken in "supported" NVCC versions
+if("${ALPAKA_CUDA_COMPILER}" STREQUAL "nvcc")
+    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+        if(CUDA_VERSION VERSION_EQUAL 8.0)
+            if(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 5.0)
```
psychocoderHPC: This is wrong: it must because the tuple issue was introduced with gcc 5.4

```
if(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 5.4)
    message(FATAL_ERROR "NVCC 8.0 does not support the std::tuple "
    "implementation in GCC 5.4+. Please use GCC 4.9 - 5.3!")
endif()
```

psychocoderHPC: I will create a new PR soon to fix this. I like to keep my gcc5.3 chain ;-)

psychocoderHPC: opened #2636 to fix it

ax3l: ok, good to know! thx!

PR #2634, opened 2018-06-20
add clear_cbar to phase_space_visualizer.py

kossag14:
To change plots on the eupraxia GUI it is also necessary to clear the colorbar. Therefore the function clear_cbar was added.

### DIFF
--- a/lib/python/picongpu/plugins/plot_mpl/phase_space_visualizer.py
+++ b/lib/python/picongpu/plugins/plot_mpl/phase_space_visualizer.py
@@ -32,6 +32,7 @@ class Visualizer(BaseVisualizer):
         # for unit-conversion from SI (taken from picongpu readthedocs)
         self.mu = 1.e6
         self.e_mc_r = 1. / (9.1e-31 * 2.9979e8)
+        self.cbar = None
 
     def _create_data_reader(self, run_directory):
         """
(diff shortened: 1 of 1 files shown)


lib/python/picongpu/plugins/plot_mpl/phase_space_visualizer.py
```diff
@@ -89,15 +90,19 @@ def visualize(self, ax=None, **kwargs):
...
                 '$')
 
         ax.set_xlabel(r'${0}$ [${1}$]'.format(meta.r, "\mathrm{\mu m}"))
         ax.set_ylabel(
             r'$p_{0}$ [$\beta\gamma$]'.format(meta.p))
 
+    def clear_cbar(self):
+        """Clear colorbar if present."""
+        if self.cbar is not None:
+            self.cbar.remove()
```
PrometheusPi: ~~Please add a second empty line in order to keep the code pep8 conform. (That is the error causes by travis-ci.)~~

PrometheusPi: This should be a method and not a independent function.
Please indent the code.

PrometheusPi: *discussed offline:*
Issue was the missing space at the end of the class definition.

PR #2636, opened 2018-07-04
re-enable CUDA 8.0 and gcc 5.0 - 5.3 support

psychocoderHPC:
Allow to use CUDA 8.0 together with gcc 5.0 to 5.3.
The support for this compiler combination was removed with #2628.

Note: I used `gcc/5.3.0` and `cuda/8.0` together since over a year and have no issues.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -33,7 +33,7 @@ gcc
 """
 - 4.9 - 7 (if you want to build of Nvidia GPUs, supported compilers depend on your current `CUDA version <https://gist.github.com/ax3l/9489132>`_)
 
-  - CUDA 8.0: Use gcc 4.9
+  - CUDA 8.0: Use gcc 4.9 - 5.3
   - CUDA 9.0 - 9.1: Use gcc 4.9 - 5
   - CUDA 9.2: Use gcc 4.9 - 7 (not yet supported)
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
@ax3l could you please update your gist and add that for CUDA 8 GCC <=5.3 is supported

ax3l:
excellent observation!

INSTALL.rst
```diff
@@ -33,7 +33,7 @@ gcc
 """
 - 4.9 - 7 (if you want to build of Nvidia GPUs, supported compilers depend on your current `CUDA version <https://gist.github.com/ax3l/9489132>`_)
 
-  - CUDA 8.0: Use gcc 4.9
+  - CUDA 8.0: Use gcc 4.9 - 5.3
   - CUDA 9.0 - 9.1: Use gcc 4.9 - 5
```
ax3l: @psychocoderHPC I think this one should also be 4.9 - 5.3 (CUDA 9.2 fixes GCC 5.4/6 std::tuple issues afaik)

ax3l: std::tuples should not work with CUDA 9.0 and 9.1

> From the CMAKE side all gcc 5.X are allowed with CUDA 9.

I guess it's the CUDA "reference side" you are talking about

PR #2637, opened 2018-07-06
supercell particle counter

psychocoderHPC:
Extent the supercell to be aware about the number current number of particles.

- SuperCell:
  - add method `getNumParticles()`
  - remove method `setSizeLastFrame()`
  - add method `setNumParticles()`
  - re-format SuperCell to the new code style
  - add const to read only methods
- ParticleBase
  - update documentation
  - count number of particles during `KernelShiftParticles` and fill gaps kernel
- add unit test to test the supercell method `getSizeLastFrame()`

The second commit contains only the code style changes and add const to read only methods  of the supercell.

I am not sure if we should put this PR into 0.4.0 or wait until we opened the release branch.

- [x] rebase against #2649

### DIFF
--- a/include/pmacc/particles/ParticlesBase.hpp
+++ b/include/pmacc/particles/ParticlesBase.hpp
@@ -172,6 +172,11 @@ public:
     /** copy guard particles to intermediate exchange buffer
      *
      * Copy all particles from the guard of a direction to the device exchange buffer.
+     *  @warning This method resets the number of particles in the processed supercells even
+     * if there are particles left in the supercell and not fulfill the rule that the
+     * last frame is contiguous filled.
+     * Call KernelFillGaps afterwards if you need a valid number of particles
+     * and a contiguous filled last frame.
      */
     void copyGuardToExchange(uint32_t exchangeType);
 
--- a/include/pmacc/particles/ParticlesBase.hpp
+++ b/include/pmacc/particles/ParticlesBase.hpp
@@ -172,11 +172,11 @@ public:
     /** copy guard particles to intermediate exchange buffer
      *
      * Copy all particles from the guard of a direction to the device exchange buffer.
-     *  @warning This method resets the number of particles in the processed supercells even
-     * if there are particles left in the supercell and not fulfill the rule that the
-     * last frame is contiguous filled.
-     * Call KernelFillGaps afterwards if you need a valid number of particles
-     * and a contiguous filled last frame.
+     * @warning This method resets the number of particles in the processed supercells even
+     * if there are particles left in the supercell and does not guarantee that the last frame is
+     * contiguous filled.
+     * Call fillAllGaps afterwards if you need a valid number of particles
+     * and a contiguously filled last frame.
      */
     void copyGuardToExchange(uint32_t exchangeType);
 
--- a/include/pmacc/particles/ParticlesBase.kernel
+++ b/include/pmacc/particles/ParticlesBase.kernel
@@ -968,10 +968,10 @@ struct KernelDeleteParticles
 /** copy particles from the guard to an exchange buffer
  *
  * @warning This kernel resets the number of particles in the processed supercells even
- * if there are particles left in the supercell and not fulfill the rule that the
- * last frame is contiguous filled.
+ * if there are particles left in the supercell and does not guarantee that the last frame is
+ * contiguous filled.
  * Call KernelFillGaps afterwards if you need a valid number of particles
- * and a contiguous filled last frame.
+ * and a contiguously filled last frame.
  *
  * @tparam T_numWorkers number of workers
  */
--- a/include/pmacc/particles/ParticlesBase.kernel
+++ b/include/pmacc/particles/ParticlesBase.kernel
@@ -229,17 +229,21 @@ struct KernelFillGapsLastFrame
                 uint32_t const
             )
             {
-                if( lastFrame.isValid() && counterParticles == 0 )
+                // there is no need to add a zero to the global memory
+                if( counterParticles != 0 )
                 {
-                    lastFrame = getPreviousFrameAndRemoveLastFrame(lastFrame, pb, superCellIdx);
-                    /* if there is a previous valid frame then this frame is fully filled and
-                     * now the last frame
-                     */
-                    if( lastFrame.isValid() )
-                        counterParticles = frameSize;
+                    auto & superCell = pb.getSuperCell( superCellIdx );
+                    superCell.setNumParticles(
+                        superCell.getNumParticles() + counterParticles
+                    );
                 }
-
-                pb.getSuperCell( superCellIdx ).setSizeLastFrame( counterParticles );
+                else
+                {
+                    /* The last frame is empty therefore it must be removed.
+                     * It is save to call this method even if there is no last frame.
+                     */
+                    pb.removeLastFrame( superCellIdx );
+                }
             }
         );
     }
@@ -320,6 +324,8 @@ struct KernelFillGaps
             counterParticles,
             int
         );
+
+        uint32_t numParticlesPerSuperCell = 0u;
 
         ForEachIdx<
             IdxConfig<
@@ -438,6 +444,7 @@ struct KernelFillGaps
                     {
                         if( counterGaps < counterParticles )
                         {
+                            numParticlesPerSuperCell += frameSize;
                             // any gap in the first frame is filled
                             firstFrame = pb.getNextFrame( firstFrame );
                         }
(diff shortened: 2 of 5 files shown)


include/pmacc/particles/ParticlesBase.kernel
```diff
@@ -229,7 +229,14 @@ struct KernelFillGapsLastFrame
                 uint32_t const
             )
             {
-                pb.getSuperCell( superCellIdx ).setSizeLastFrame( counterParticles );
+                // there is no need to add a zero to the global memory
+                if( counterParticles != 0 )
```
sbastrakov: I'm not sure `counterParticles` is a good name. First, as the word 'counter' is also an adjective, for me it seems unnatural to put it first, since here 'particles' is a dependant word. Also, is there any reason a name like 'numParticlesLastFrame' (or any variation with 'number' instead of 'counter') cannot be used here?

The first consideration also applies for `counterGaps`.

include/pmacc/particles/memory/dataTypes/SuperCell.hpp
```diff
@@ -22,70 +22,93 @@
...
+        {
+            return lastFramePtr;
+        }
+
+        HDINLINE T_FrameType const * FirstFramePtr() const
+        {
+            return firstFramePtr;
+        }
+
+        HDINLINE T_FrameType const*  LastFramePtr() const
```
sbastrakov: A space in the wrong position, should be `const * LastFramePtr`

include/pmacc/particles/memory/dataTypes/SuperCell.hpp
```diff
@@ -22,70 +22,93 @@
...
+            mustShiftVal = value;
+        }
+
+        HDINLINE uint32_t getSizeLastFrame() const
+        {
+            constexpr uint32_t frameSize = math::CT::volume<
+                typename T_FrameType::SuperCellSize
+            >::type::value;
+            uint32_t numParLastFrame = numParticles % frameSize;
+            // numParLastFrame == 0 means the last frame is fully filled
```
sbastrakov: Maybe instead of the comment create an explanatory variable `bool isLastFrameFull = ( numParLastFrame == 0 );` ?

include/pmacc/particles/memory/dataTypes/SuperCell.hpp
```diff
@@ -22,70 +22,93 @@
...
+        {
+            mustShiftVal = value;
+        }
+
+        HDINLINE uint32_t getSizeLastFrame() const
+        {
+            constexpr uint32_t frameSize = math::CT::volume<
+                typename T_FrameType::SuperCellSize
+            >::type::value;
+            uint32_t numParLastFrame = numParticles % frameSize;
```
sbastrakov: Typo: Par instead of Per

include/pmacc/particles/memory/dataTypes/SuperCell.hpp
psychocoderHPC: I can do it but I need to keep the comment because it is not clear why zero means the frame is full. I will try to improve the comment

include/pmacc/particles/memory/dataTypes/SuperCell.hpp
psychocoderHPC: Ahh it take me a few seconds to get you comment, `Par` means particles. I will change it to `numParticlesInLastFrame`

sbastrakov: Ah, sorry, I thought it was meant to be "number per last frame".

include/pmacc/particles/memory/dataTypes/SuperCell.hpp
sbastrakov: As discussed offline, there is a bug when `numParticles == 0` - the function now returns frameSize and should actually return 0. I guess one could use smth like `return numParticles ? (1 + (numParticles - 1) % frameSize) : 0`;

include/pmacc/particles/ParticlesBase.hpp
```diff
@@ -172,6 +172,11 @@ class ParticlesBase : public SimulationFieldHelper<T_MappingDesc>
     /** copy gard particles to intermediate exchange buffer
      *
      * Copy all particles from the guard of a direction to the device exchange buffer.
+     *  @warning This method resets the number of particles in the processed supercells even
```
sbastrakov: Should there be an additional space before `@warning`?

include/pmacc/particles/ParticlesBase.hpp
```diff
@@ -172,6 +172,11 @@ class ParticlesBase : public SimulationFieldHelper<T_MappingDesc>
     /** copy gard particles to intermediate exchange buffer
```
sbastrakov: Typo: `gard ` instead of `guard`

include/pmacc/particles/ParticlesBase.hpp
```diff
@@ -172,6 +172,11 @@ class ParticlesBase : public SimulationFieldHelper<T_MappingDesc>
...
      *
      * Copy all particles from the guard of a direction to the device exchange buffer.
+     *  @warning This method resets the number of particles in the processed supercells even
+     * if there are particles left in the supercell and not fulfill the rule that the
+     * last frame is contiguous filled.
+     * Call KernelFillGaps afterwards if you need a valid number of particles
```
sbastrakov: I'm not sure what style is used in picongpu for such things, but maybe `KernelFillGaps` should be written as `KernelFillGaps()` to emphasize it is a function?

include/pmacc/particles/ParticlesBase.hpp
```diff
@@ -172,6 +172,11 @@ class ParticlesBase : public SimulationFieldHelper<T_MappingDesc>
...
+     * and a contiguous filled last frame.
```
sbastrakov: I think `contiguous filled last frame` should be fixed to either `contiguous last frame` or `contiguously filled last frame`.

include/pmacc/particles/ParticlesBase.hpp
```diff
@@ -172,6 +172,11 @@ class ParticlesBase : public SimulationFieldHelper<T_MappingDesc>
     /** copy gard particles to intermediate exchange buffer
      *
      * Copy all particles from the guard of a direction to the device exchange buffer.
+     *  @warning This method resets the number of particles in the processed supercells even
+     * if there are particles left in the supercell and not fulfill the rule that the
```
sbastrakov: At the very least there should be 'does' before `not`. Maybe the whole second part could be rewritten as `...and does not guarantee the last frame is ...` for clarity.

include/pmacc/particles/ParticlesBase.kernel
```diff
@@ -916,16 +951,20 @@ struct KernelDeleteParticles
...
     }
 };
 
 /** copy particles from the guard to an exchange buffer
+ *
+ * @warning This kernel resets the number of particles in the processed supercells even
+ * if there are particles left in the supercell and not fulfill the rule that the
+ * last frame is contiguous filled.
+ * Call KernelFillGaps afterwards if you need a valid number of particles
+ * and a contiguous filled last frame.
```
sbastrakov: This comment seems to be mostly a copy of a comment at the header file, same fixes are probably needed here.

include/pmacc/particles/memory/dataTypes/SuperCell.hpp
psychocoderHPC: note: it must be `numParticles ? (1 + (numParticles - 1) % frameSize) : 0;` because `%` has less precedence than `+`.

sbastrakov: Does it? Cppreference states [the remainder is the same priority as multiplication](https://en.cppreference.com/w/cpp/language/operator_precedence).

psychocoderHPC: you are right, I looked to [this site](https://en.wikipedia.org/wiki/Operators_in_C_and_C%2B%2B) and missed that the `+` there is a unary add. OK I will fix it

include/pmacc/particles/ParticlesBase.hpp
psychocoderHPC: I made this typo very often, as soon as this PR is merged I will fix is globally

psychocoderHPC: issue #2643

psychocoderHPC:
Note: I forced push a fix to this PR to add  #2649. In the old code  #2649 was not a bug but within the old code but together with the new counting method from this PR the last frame is not allowed to be empty after fill gaps is called.

PR #2639, opened 2018-07-09
NVCC 8.0 GCC <= 5.3 && 9.0/9.1: GCC <= 5.5

ax3l:
Specify maximum supported GCC due to std::tuple issues in nvcc <= 9.1

This just adds GCC 5.5 explicitly to the docs.

Follow-up to #2636

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -34,7 +34,7 @@ gcc
 - 4.9 - 7 (if you want to build of Nvidia GPUs, supported compilers depend on your current `CUDA version <https://gist.github.com/ax3l/9489132>`_)
 
   - CUDA 8.0: Use gcc 4.9 - 5.3
-  - CUDA 9.0 - 9.1: Use gcc 4.9 - 5
+  - CUDA 9.0 - 9.1: Use gcc 4.9 - 5.5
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
we need also update CMake https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/include/pmacc/PMaccConfig.cmake#L336 

```
if(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 5.4)
```

INSTALL.rst
```diff
@@ -34,7 +34,7 @@ gcc
 - 4.9 - 7 (if you want to build of Nvidia GPUs, supported compilers depend on your current `CUDA version <https://gist.github.com/ax3l/9489132>`_)
 
   - CUDA 8.0: Use gcc 4.9 - 5.3
-  - CUDA 9.0 - 9.1: Use gcc 4.9 - 5
+  - CUDA 9.0 - 9.1: Use gcc 4.9 - 5.3
```
ax3l: @psychocoderHPC looks in https://github.com/ComputationalRadiationPhysics/alpaka/pull/566 that indeed GCC 5.5 works with CUDA 9.0

PR #2641, opened 2018-07-10
Doxygen: Less XML

ax3l:
From the Doxygen docs:

> If the XML_PROGRAMLISTING tag is set to YES doxygen will dump the program
> listings (including syntax highlighting and cross-referencing information) to
> the XML output. Note that enabling this will significantly increase the size
> of the XML output.
> The default value is: YES.
> This tag requires that the tag GENERATE_XML is set to YES.

This should reduce our memory footprint when building the Sphinx docs: https://github.com/michaeljones/breathe/issues/315
From tests on my local machine, it did roughly half the memory consumption during Sphinx' `make html`.

At the same time, out Sphinx docs still contain all information we need (we e.g. don't include XML program listings and if we would need to, we could include code directly without breathe).

First seen in https://github.com/jcarrano/RIOT/commit/c3555dfec2003465fb1c675db7299117117584e1


ax3l:
@psychocoderHPC mergable :)

PR #2649, opened 2018-07-24
remove empty last frame

psychocoderHPC:
By definition our last frame is allowed to not be fully filled with particles.
To reduce the memory foot print the last frame, it should never be empty after all gaps are filled.
This is currently not always the case.

Remove the last frame in the a supercell in `KernelFillGapsLastFrame` if it is empty.

# Tests

- [x] runtime tests

@theZiz thx for helping me to find this issue with your graphical debugging tool [ISAAC](https://github.com/ComputationalRadiationPhysics/isaac) :-)

### DIFF
--- a/include/pmacc/particles/ParticlesBase.kernel
+++ b/include/pmacc/particles/ParticlesBase.kernel
@@ -229,6 +229,16 @@ struct KernelFillGapsLastFrame
                 uint32_t const
             )
             {
+                if( lastFrame.isValid() && counterParticles == 0 )
+                {
+                    lastFrame = getPreviousFrameAndRemoveLastFrame(lastFrame, pb, superCellIdx);
+                    /* if there is a previous valid frame then this frame is fully filled and
+                     * now the last frame
+                     */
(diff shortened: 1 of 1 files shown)


include/pmacc/particles/ParticlesBase.kernel
```diff
@@ -229,6 +229,16 @@ struct KernelFillGapsLastFrame
                 uint32_t const
             )
             {
+                if( counterParticles == 0 )
+                {
+                    getPreviousFrameAndRemoveLastFrame(lastFrame, pb, superCellIdx);
```
theZiz: I guess this should be `lastFrame = getPreviousFrameAndRemoveLastFrame(lastFrame, pb, superCellIdx);` ?
Otherwise you have no reference to the `lastFrame` and check for a removed frame in the next line

psychocoderHPC: yes absolutly

ax3l:
offline: works

PR #2658, opened 2018-07-31
Python: EnergyHistogram MPL Plot

ax3l:
Fix the python module plotting the energy histogram plugin on several issues:

- data order: order of x-y arguments mismatched
- update image: rescale properly in log axes
- labels: typo fixed, iteration removed

### DIFF
--- a/lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
+++ b/lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
@@ -37,15 +37,18 @@ class Visualizer(BaseVisualizer):
         Implementation of base class function.
         Turns 'self.plt_obj' into a matplotlib.pyplot.plot object.
         """
-        bins, counts = self.data
+        counts, bins = self.data
         self.plt_obj = ax.semilogy(bins, counts, nonposy='clip')[0]
 
     def _update_plt_obj(self):
         """
         Implementation of base class function.
         """
-        bins, counts = self.data
+        counts, bins = self.data
         self.plt_obj.set_data(bins, counts)
+        ax = self._ax_or_gca(None)
+        ax.relim()
+        ax.autoscale_view(True, True, True)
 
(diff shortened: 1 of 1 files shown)


ax3l:
> Could we store the whole pd.DataFrame on the first read or would that be too large?

For this plugin, that's a good idea. We should still verify if the cache is still valid, e.g. via timestamp of the file. Otherwise we would miss new values if we look at a simulation that is still running.

We could do this in another PR if you like.

> Why is the Travis build failing btw?

It's infrastructure is just flaky recently. Restarted.

ax3l:
> Sophie and me are still investigating whether the ax rescaling should rather be done in the visualize() function...

hm, probably both is ok. The `_update_plt_obj` name suggests the whole object get's updated, so the current location seems fine as well.

ax3l:
@codingS3b any results? I think the PR way you propose to update the log axis is fine.

codingS3b:
I have not had another look at it yet. @kossag14  do you have any obligations regarding the update of the scales? Should the `ax.relim()` be called in a different place instead?
I think the issue was that those commands were not executed in the UI if only a different iteration was selected.

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
```diff
@@ -37,15 +37,18 @@ def _create_plt_obj(self, ax):
         Implementation of base class function.
         Turns 'self.plt_obj' into a matplotlib.pyplot.plot object.
         """
-        bins, counts = self.data
+        counts, bins = self.data
```
PrometheusPi: great fix - just a suggestion, `bins` is actually bin borders

ax3l: yep, that should be documented in the data API, not too important in the matplotlib backend to mention again since it's not visible to users.

PR #2659, opened 2018-08-02
fix `Manipulate*`

psychocoderHPC:
The algorithm `Manipulate` and `ManipulateDerive` allows to set a particle.
This filter was not provided with the species information.

- apply corresponding species to the filter functor


ax3l:
@psychocoderHPC can we add a usage to our examples, e.g. copying a few % of particles to a second tracking species?

PR #2660, opened 2018-08-02
remove unused species template signature

psychocoderHPC:
PIConGPU RNG functors, filters and startPosition functors have the species type in the tempate signature. This type is not needed. 
The species was a left over from the old random number generator.

- remove unused template species signature from RNG functors


ax3l:
@psychocoderHPC there seem to be some compile errors left :)

psychocoderHPC:
It was a missing apply in `startPosition::FreeRng`

PR #2666, opened 2018-08-06
PMacc: Throw in Destructors

ax3l:
Throwing in destructors is not allowed in C++11 and causes compile warnings.

We should probably refactor those methods or at least introduce macros that instead of throwing write to `stderr`.


include/pmacc/communication/manager_common.hpp
```diff
@@ -36,3 +36,5 @@ enum {
 };
 
 #define MPI_CHECK(cmd) {int error = cmd; if(error!=MPI_SUCCESS){std::cerr << "<" << __FILE__ << ">:" << __LINE__; throw std::runtime_error(std::string("[MPI] Error"));}}
+
+#define MPI_CHECK_NO_EXCEP(cmd) {int error = cmd; if(error!=MPI_SUCCESS){std::cerr << "[MPI] Error code " << error << " in <" << __FILE__ << ">:" << __LINE__;}}
```
sbastrakov: To me `NO_EXCEP `does not seem a natural way to shorten. Why not `NO_EXCEPT ` or `NO_THROW` (or same without an underscore), since there are `noexcept `and `nothrow `in C++ ?

Unrelated to this PR, to me it is also more logical to have smth like a non-throwing `MPI_CHECK ` and a throwing `MPI_CHECK_THROW` or `MPI_CHECK_EXCEPT`, but that's a matter of personal opinion I guess.

ax3l: I agree, just used the same naming as we use in `CUDA_CHECK_` macros.
Of course we can refactor both.

psychocoderHPC: please change it to `MPI_CHECK_NO_EXCEPT`

ax3l: Ok, but I will change `CUDA_CHECK_NO_EXCEPT` at the same time :)

include/pmacc/communication/manager_common.hpp
```diff
@@ -36,3 +36,5 @@ enum {
 };
 
 #define MPI_CHECK(cmd) {int error = cmd; if(error!=MPI_SUCCESS){std::cerr << "<" << __FILE__ << ">:" << __LINE__; throw std::runtime_error(std::string("[MPI] Error"));}}
+
+#define MPI_CHECK_NO_EXCEPT(cmd) {int error = cmd; if(error!=MPI_SUCCESS){std::cerr << "[MPI] Error code " << error << " in <" << __FILE__ << ">:" << __LINE__;}}
```
sbastrakov: Just realized that these macros are wrapped in a block, which is [potentially wrong when used in `if` statements](https://stackoverflow.com/questions/1067226/c-multi-line-macro-do-while0-vs-scope-block), a `do-while` wrapper is a strictly better alternative. Will create a separate issue for that.

ax3l: That's true.
Nevertheless, those check macros are designed as statements, not as expressions (which could be used in `if` statements).

sbastrakov: But even in this PR changes they *are* used in if-statements, however there is no problem yet.

sbastrakov: Also, is there any point in having these error code checkers as macros, and not functions?

ax3l: > macros, and not functions

Actually no reason. The only thing that's nice is `__FILE__ ` & `__LINE__` can be printed easily that way.

ax3l: > they are used in if-statements (not in the condition, but in the body)

Ah, I see. You mean situations like this:
```C++
if( something )
    MPI_CHECK(MPI_Comm_rank(MPI_COMM_WORLD, &rank));
```

? I probably don't get the linked example

sbastrakov: So your example is fine only as long as it does not have an `else` branch. The issue with `else` is after the macro is expanded it would become
```C++
if( something )
/* start of macro expansion -> */ {
     ... 
} /* <- end of macro expansion*/ ; // <- this semicolon was after the macro, still here 
else
...
```
and the `};` would trigger the end of the `if` statement and a corresponding compile error `else`.

This problem is specific for the case, when three conditions are met:
* `if` body is not contained in `{}`
* the body has `;` after a macro (but having it is kinda natural)
* there is an `else` block.

ax3l: Ah, close! Now I got it!

sbastrakov: I kinda forgotten about these file and line. I think it is actually helpful and probably outweight (my) general disgust for macro. However, nothing prevents calling a function inside a macro and providing `__FILE__` and `__LINE__` as its arguments, like how boost/assert is implemented. I think generally it is versatile in terms of supporting several modes of output (e.g., via a singleton logger object), or in combination with third-party logging libraries.

ax3l: One could also write a function on which one of it's arguments is a macro containing those two args xD

psychocoderHPC: @sbastrakov Yes calling a function inside the macro and move `__FILE__` into it is the best way to avoid side effects with `if, else`.
I use this way also in cupla https://github.com/ComputationalRadiationPhysics/cupla/blob/f5fedcfdedcf7868142ba4a636378fba383031d4/example/CUDASamples/common/helper_cuda.h#L895-L910

sbastrakov: Somehow that never occured to me, but it might actually be nicer, [smth like that](http://coliru.stacked-crooked.com/a/dd773e7b968b4697).

sbastrakov: @psychocoderHPC your implementation has the same issue with expansion-brackets-if interaction. To protect against it, *any* macro called semantically as a function has to be done in a do-while guard.

sbastrakov: Discussed offline, @psychocoderHPC 's implementation is actually fine and does not have problems I suspected.

PR #2669, opened 2018-08-07
fix memory leaks PMacc/PIConGPU

psychocoderHPC:
Fix a memory leak found with valgrind in:
  - `ExchangeIntern.hpp`
  - `FieldTmp`

Use `std::unique_ptr` instead of native pointers to avoid memory leaks.

### DIFF
--- a/include/picongpu/fields/FieldTmp.tpp
+++ b/include/picongpu/fields/FieldTmp.tpp
@@ -55,8 +55,6 @@ namespace picongpu
         uint32_t slotId
     ) :
         SimulationFieldHelper<MappingDesc>( cellDescription ),
-        fieldTmp( nullptr ),
-        fieldTmpRecv( nullptr ),
         m_slotId( slotId )
     {
         m_commTagScatter =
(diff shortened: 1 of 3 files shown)


include/picongpu/fields/FieldTmp.tpp
```diff
@@ -65,10 +63,14 @@ namespace picongpu
         m_commTagGather = ++pmacc::traits::detail::GetUniqueTypeId< uint8_t >::counter +
             SPECIES_FIRSTTAG;
 
-        fieldTmp = new GridBuffer <ValueType, simDim >( cellDescription.getGridLayout( ) );
+        fieldTmp.reset(
```
sbastrakov: I believe it is worth to create an own C++11 implementation of C++14 `std::make_unique`, e.g. given [here](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique), and use it to initialize unique pointers. It avoids explicit `new ` in the user code, and in some more complicated cases is also safer.

include/pmacc/memory/buffers/ExchangeIntern.hpp
```diff
@@ -119,9 +131,6 @@ namespace pmacc
 
         virtual ~ExchangeIntern()
```
sbastrakov: Maybe make this destructor `= default` now?

include/picongpu/fields/FieldTmp.tpp
ax3l: Agreed. Out plan is actually that after the next release we could go for CUDA 9.0+
This will allow us to go straight ot a C++14 code base.

sbastrakov: Wow, that would be so cool, C++14 ftw!

ax3l: I know, I am really happen most systems got the CUDA 9 upgrade :rocket: 
(Still waiting for Hypnos, though... sad)

psychocoderHPC: Somehow the C++ guys missed to include the factories in C++11 :-(
Let me ship a own `make_unique` in a separate PR. I would like to use it on many more places.

PR #2670, opened 2018-08-07
ADIOS: No adios_set_max_buffer_size

### LINKED ISSUE #2656: ADIOS buffer problem
Hi!
I have a problem with buffering with adios/1.13.1.
Here is part of the error message:
image
and path to the files
/bigdata/hplsim/external/mukhar40/ColloidalMelting_TF_I=3.0_1ps.
I tried just recompiling - resubmitting once and it didn't help.

This simulation was restarting automatically as was discussed in https://github.com/ComputationalRadiationPhysics/picongpu/issues/2618#issuecomment-394273964
and for some time it worked perfectly but then that error appeared.

ax3l:
There is no need to set `adios_set_max_buffer_size`. It sets the maximum ADIOS is allowed to allocate in its buffers (but this call does not pre-allocate things).

But the actual buffer is since ADIOS 1.10.0 already dynamically allocated (#1639) and the default for maximum allocation is a large default. If we later on want to control host-side memory, we can just set a ratio of free host mem, but ADIOS will only allocate what it actually needs until then.

See
  https://github.com/ornladios/ADIOS/commit/9fae0d12fea535a4366e81c0c04d7a3be54c3179

Fix #2656 (Crashes reported in some cases since our buffer estimate was not precise enough.)

### DIFF
--- a/include/picongpu/plugins/adios/ADIOSWriter.hpp
+++ b/include/picongpu/plugins/adios/ADIOSWriter.hpp
@@ -1444,12 +1444,8 @@ private:
         writeIdProviderStartId.prepare(*threadParams, idProviderState.maxNumProc);
         writeIdProviderNextId.prepare(*threadParams);
 
-        /* allocate buffer in MB according to our current group size */
-        /* `1 + mem` minimum 1 MiB that we can write attributes on empty GPUs */
-        size_t writeBuffer_in_MiB=1+threadParams->adiosGroupSize / 1024 / 1024;
-        /* value `1.1` is the secure factor if we miss to count some small buffers*/
-        size_t buffer_mem=static_cast<size_t>(1.1 * static_cast<float_64>(writeBuffer_in_MiB));
-        adios_set_max_buffer_size(buffer_mem);
+        // in the past, we had to explicitly estiamte our buffers.
+        // this is now done automatically by ADIOS on `adios_write()`
         threadParams->adiosBufferInitialized = true;
 
         /* open adios file. all variables need to be defined at this point */


include/picongpu/plugins/adios/ADIOSWriter.hpp
```diff
@@ -1446,10 +1446,10 @@ class ADIOSWriter : public IIOBackend
...
         /* allocate buffer in MB according to our current group size */
         /* `1 + mem` minimum 1 MiB that we can write attributes on empty GPUs */
-        size_t writeBuffer_in_MiB=1+threadParams->adiosGroupSize / 1024 / 1024;
+        // size_t writeBuffer_in_MiB=1+threadParams->adiosGroupSize / 1024 / 1024;
         /* value `1.1` is the secure factor if we miss to count some small buffers*/
-        size_t buffer_mem=static_cast<size_t>(1.1 * static_cast<float_64>(writeBuffer_in_MiB));
-        adios_set_max_buffer_size(buffer_mem);
+        // size_t buffer_mem=static_cast<size_t>(1.1 * static_cast<float_64>(writeBuffer_in_MiB));
+        // adios_set_max_buffer_size(buffer_mem);
         threadParams->adiosBufferInitialized = true;
```
sbastrakov: Two potentially silly questions:
* why is the now obsolete code commented out, and not deleted?
* is `threadParams->adiosBufferInitialized` actually used anywhere (could not find the usage for this member at all)?

ax3l: The whole implementation that goes through the data first, stores IDs, estimates sizes, etc. is unfortunately outdated. That was what we needed to do around 2015.

We are planning to unify the ADIOS and HDF5 backend with a student in the next months via the new https://github.com/openPMD/openPMD-api 

But until then, I just want to apply fixes to it and don't invest time into the legacy part of those plugins.

psychocoderHPC:
Offline discussed: The commented code should be removed to avoid dead code within the code-base.

ax3l:
Yep, I am replacing the code with a comment instead, so that `threadParams->adiosBufferInitialized = true;` looks less odd.

PR #2678, opened 2018-08-08
remove define `ENABLE_CURRENT`

psychocoderHPC:
- remove the legacy define `ENABLE_CURRENT` from `components.param`
- remove LWFA compile combination where `ENABLE_CURRENT=0` is used

The warm copper example is using the modern way to disable the current by not adding a
current solver to the species.


ax3l:
@PrometheusPi see PR description:

> The warm copper example is using the modern way to disable the current by not adding a
current solver to the species.

PR #2683, opened 2018-08-10
fix possible deadlock in blocking MPI ops

psychocoderHPC:
Our event system in PMacc is not running in a separate threads. If we call a blocking operation e.g. MPI_Barrier or MPI collective operations and there are open tasks in the event system queue it could be that the full simulation is freezing. The reason is that it is possible that a MPI process `1` is waiting for a MPI operation which is never opened from the peer rank `2` because the peer is hanging in an other blocking operation those can not be finsihed until the rank `1` starting the corresponding counter part operation.
Parts of the bugs was already fixed in #1659 but we missed that this is a behavior with all blocking operations e.g. MPI blocking operations, MPI blocking collectives and parallel file system operations those mostly also use MPI blocking collectives.

### DIFF
--- a/include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
+++ b/include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
@@ -138,6 +138,8 @@ public:
                                               DivideInPlace<float_X>(float_X(numRanks)));
         }
 
+        // avoid deadlock between not finished pmacc tasks and mpi blocking collectives
+        __getTransactionEvent().waitForFinished();
         MPI_Bcast(&(*hBufLeftParsCalorimeter.origin()),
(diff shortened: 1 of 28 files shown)


include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -138,6 +138,8 @@ class ParticleCalorimeter : public plugins::multi::ISlave
                                               DivideInPlace<float_X>(float_X(numRanks)));
         }
 
+        // avoid deadlock between not finished pmacc tasks and mpi blocking collectives
+        __getTransactionEvent().waitForFinished();
         MPI_Bcast(&(*hBufLeftParsCalorimeter.origin()),
```
ax3l: follow-up PR: unguarded MPI-call (`MPI_CHECK`/`MPI_CHECK_NO_EXCEPT`)

Issue #2688, opened 2018-08-15
LWFA on cpu shows different (wrong) physics


PrometheusPi:
When running the default LWFA example with `32.cfg`there is a major difference between using gpus or cpus.

In the following you will see png pictures of the simulation after 600 iterations/time steps. 

The current `master` brach:
![grafik](https://user-images.githubusercontent.com/5121158/44151879-25acdd80-a0a4-11e8-9b0e-da10f76ede44.png)

The current `dev`branch on `k80`GPUs:
![grafik](https://user-images.githubusercontent.com/5121158/44151888-32607ff0-a0a4-11e8-8ffd-5fb274d82953.png)

The current `dev`branch on `laser`CPUs:
![grafik](https://user-images.githubusercontent.com/5121158/44151904-3e8a0e04-a0a4-11e8-95eb-4cdef28dc2bd.png)

The last is definitely wrong, an `a_0=8.0` would cause a blow out, not these streams.

PrometheusPi:
@kossag14 This is probably not an issue that effects you. I think it is more likely the difference you see arise from #2687.

PrometheusPi:
as recommended by @psychocoderHPC *offline* I checked how PIConGPU performs on CPUs (`laser`nodes) at `60 commits` **before** `tags/0.4.0-rc1`. 

The results looks great! We have "correct" physics! ✨  🎉 

600 iterations of `57bcf5cc592b3bf9ad14902d1cadbd4ec8d76193` on `laser` nodes:
![grafik](https://user-images.githubusercontent.com/5121158/44155992-1653e850-a0af-11e8-9d14-08168c536731.png)


Probably the laser refactoring #2587 (or #2624) caused this bug.

PrometheusPi:
@psychocoderHPC as you expected, calling the laser functor is causing the error.

With the fix:
```diff
diff --git a/include/picongpu/fields/LaserPhysics.hpp b/include/picongpu/fields/LaserPhysics.hpp
index 66f8c84..09bab6f 100644
--- a/include/picongpu/fields/LaserPhysics.hpp
+++ b/include/picongpu/fields/LaserPhysics.hpp
@@ -79,12 +79,6 @@ namespace fields
 
             uint32_t cellOffsetInSuperCellFromInitPlaneY = LaserFunctor::Unitless::initPlaneY % SuperCellSize::y::value;
 
-            auto accLaserFunctor = laserFunctor(
-                acc,
[...]

PR #2689, opened 2018-08-15
fix laser init on cpu

PrometheusPi:
This fixes the wrong laser initialization on CPUs due to a reuse of the laser functor in the `ForEachIdx` loop (issue #2688).

co-authored by @psychocoderHPC.


PrometheusPi:
@ax3l This affects the latest release candidate.

Issue #2504, opened 2018-02-05
Crash of the PIConGPU simulation on k20


NastasiaM:
I tried to start PIConGPU simulation and got these error messages:

[kepler002:02637] 1 more process has sent help message help-orte-odls-base.txt / orte-odls-base:could-not-kill
[kepler002:02637] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages

And one week ago the same simulation worked nicely...
Is it due to some cluster update work or?

ax3l:
Dear @NastasiaM, thank you for the report.
Yes, we are struggling with the updates the cluster received last week and are waiting for a solution and tests by our central IT.

Can you post the full error/output message as an attachement? The message you posted looks ok so far.

NastasiaM:
Hi Axel!

Here it is.
![image](https://user-images.githubusercontent.com/31509166/35803722-17f23f3e-0a75-11e8-8743-dcb3e5953ed3.png)

ax3l:
the `stderror` looks ok so far, what's in `stdout`?

NastasiaM:
![image](https://user-images.githubusercontent.com/31509166/35819464-19e34ad0-0aa3-11e8-8ffb-46b3e994452b.png)
![image](https://user-images.githubusercontent.com/31509166/35819489-32c2c954-0aa3-11e8-9371-a1cf67521cab.png)

ax3l:
All right, let's try to re-compile and re-submit tomorrow.
Have not heard about the update yet, but I hope they will announce it on the HPC user list on which you should be on as well.

NastasiaM:
I tried re-compile and re-submit today, and it still gives the same error. What is strange - the simulation crashed after a couple of hours (100 fs was simulated),  and it is just the time of the first checkpoint. Could it happen that something is wrong with the checkpoint?

ax3l:
Hm, we currently hope for the update that is scheduled for Thursday. What was the error & output this time?

NastasiaM:
Error message was the same as previous time. Ok, let's wait till Thursday.

ax3l:
Hi @NastasiaM ,

the cluster has been updated. Please update your `picongpu.profile` modules as follows:
- `openmpi/1.8.6.kepler.cuda80` -> `openmpi/2.1.2.kepler.cuda80`
- `hdf5-parallel/1.8.15` -> `hdf5-parallel/1.8.20`

[...]

PR #2690, opened 2018-08-16
fix out of host memory during HDF5 dumping

psychocoderHPC:
For HDF5 checkpoints and normal dumps we use CUDA mapped memory to transfere the particles to the host memory. Due to the usage of cupla and the reason that mapped memory is not supported by cupla and alpaka we fall back to native CUDA functions. To free mapped memory the function `cudaFreeHost`. This function is also used to free normal host memory allocated with `cudaMallocHost` which is supported by cupla.
The cupla internal macros rename `cudaFreeHost` to `cuplaFreeHost` and fall back to alpaka functionallity to free the memory. In the case where we allocated the memory without the knowledge of cupla, cupla is trowing an error and `cudaFreeHost` will never be called.
The result in PIConGPU is that with each particle dump the host memory footprint grows until we run out of memory.

- fix broken memory freeing
- `MappedBufferIntern`
  - fix that wrong function is used to allocate mapped memory
  - add workaround to free mapped memory
  - add fallpack if an CPU accelerator is used

Note: `MappedBufferIntern` is not used for long time therefore the wrong allocation call was not seen by any user of PMacc

This bugfix ~~should~~ solve: #2504

# Tests

- [x] KHI with 192x512x64 cells on 1  GPU

CC-ing: @PrometheusPi  @steindev @HighIander  @BeyondEspresso

### DIFF
--- a/include/picongpu/plugins/output/WriteSpeciesCommon.hpp
+++ b/include/picongpu/plugins/output/WriteSpeciesCommon.hpp
@@ -145,16 +145,24 @@ struct FreeMemory
         if (ptr != nullptr)
         {
 #if( PMACC_CUDA_ENABLED == 1 )
-            auto rc = cudaFreeHost(ptr);
-            /* cupla can't handle foreign memory allocated with `cudaHostAlloc`
-             * therefore the cupla error `cuplaErrorMemoryAllocation` is ignored
-             */
-            if(rc != cuplaErrorMemoryAllocation)
-                CUDA_CHECK(rc)
+/* cupla 0.1.0 does not support the function cudaHostAlloc to create mapped memory.
+ * Therefore we need to call the native CUDA function cudaFreeHost to free memory.
+ * Due to the renaming of cuda functions with cupla via macros we need to remove
+ * the renaming to get access to the native cuda function.
+ * @todo this is a workaround please fix me. We need to investigate if
+ * it is possible to have mapped/unified memory in alpaka.
+ *
+ * corresponding alpaka issues:
+ *   https://github.com/ComputationalRadiationPhysics/alpaka/issues/296
+ *   https://github.com/ComputationalRadiationPhysics/alpaka/issues/612
+ */
+#   undef cudaFreeHost
+            CUDA_CHECK((cuplaError_t)cudaFreeHost(ptr));
+// re-introduce the cupla macro
+#   define cudaFreeHost(...) cuplaFreeHost(__VA_ARGS__)
 #else
             __deleteArray(ptr);
 #endif
-            ptr=nullptr;
         }
     }
 };
--- a/include/pmacc/memory/buffers/MappedBufferIntern.hpp
+++ b/include/pmacc/memory/buffers/MappedBufferIntern.hpp
@@ -55,7 +55,11 @@ public:
     DeviceBuffer<TYPE, DIM>(size, size),
     pointer(nullptr), ownPointer(true)
     {
-        CUDA_CHECK(cudaMallocHost(&pointer, size.productOfComponents() * sizeof (TYPE), cudaHostAllocMapped));
+#if( PMACC_CUDA_ENABLED == 1 )
+        CUDA_CHECK((cuplaError_t)cudaHostAlloc(&pointer, size.productOfComponents() * sizeof (TYPE), cudaHostAllocMapped));
+#else
+        pointer = new TYPE[size.productOfComponents()];
+#endif
         reset(false);
     }
 
@@ -69,7 +73,25 @@ public:
 
         if (pointer && ownPointer)
         {
-            CUDA_CHECK(cudaFreeHost(pointer));
+#if( PMACC_CUDA_ENABLED == 1 )
+/* cupla 0.1.0 does not support the function cudaHostAlloc to create mapped memory.
+ * Therefore we need to call the native CUDA function cudaFreeHost to free memory.
+ * Due to the renaming of cuda functions with cupla via macros we need to remove
+ * the renaming to get access to the native cuda function.
+ * @todo this is a workaround please fix me. We need to investigate if
+ * it is possible to have mapped/unified memory in alpaka.
+ *
+ * corresponding alpaka issues:
+ *   https://github.com/ComputationalRadiationPhysics/alpaka/issues/296
+ *   https://github.com/ComputationalRadiationPhysics/alpaka/issues/612
+ */
+#   undef cudaFreeHost
+            CUDA_CHECK((cuplaError_t)cudaFreeHost(pointer));
+// re-introduce the cupla macro
+#   define cudaFreeHost(...) cuplaFreeHost(__VA_ARGS__)
+#else
+            __deleteArray(pointer);
+#endif
         }
     }
 


include/picongpu/plugins/output/WriteSpeciesCommon.hpp
```diff
@@ -145,16 +145,20 @@ struct FreeMemory
         if (ptr != nullptr)
         {
 #if( PMACC_CUDA_ENABLED == 1 )
-            auto rc = cudaFreeHost(ptr);
-            /* cupla can't handle foreign memory allocated with `cudaHostAlloc`
-             * therefore the cupla error `cuplaErrorMemoryAllocation` is ignored
-             */
-            if(rc != cuplaErrorMemoryAllocation)
-                CUDA_CHECK(rc)
+/* cupla is not supporting the function cudaHostAlloc to create mapped memory.
```
ax3l: can you please open a cupla & alpaka issue and link it here in the comment?

Even if it might currently be a `wontfix` it needs to be doc-ed as an issue.

include/pmacc/memory/buffers/MappedBufferIntern.hpp
```diff
@@ -69,7 +73,21 @@ class MappedBufferIntern : public DeviceBuffer<TYPE, DIM>
 
         if (pointer && ownPointer)
         {
-            CUDA_CHECK(cudaFreeHost(pointer));
+#if( PMACC_CUDA_ENABLED == 1 )
+/* cupla is not supporting the function cudaHostAlloc to create mapped memory.
```
ax3l: link same issues here, please

include/picongpu/plugins/output/WriteSpeciesCommon.hpp
psychocoderHPC: I can open a feature request in alpaka. 
This bug is not important for cupla because I as user ignored the error cupla provided.

psychocoderHPC: I opened the alpaka feature request: https://github.com/ComputationalRadiationPhysics/alpaka/issues/612

ax3l: thx, please link it in both comments in-code :)

ax3l: Please link https://github.com/ComputationalRadiationPhysics/alpaka/issues/296 as the original alpaka issue

ax3l:
@psychocoderHPC if this is also fixing #2504, then this also affects the CUDA backend. Can you confirm?

psychocoderHPC:
Maybe I mentioned it in the PR description. I started yesterday a test if it solves the issue but missed to check the result today :-) I will check it later this evening or next week.

psychocoderHPC:
#2504 is solved by this PR.

ax3l:
Ok, so I tag this as CUDA bug now as well.

ax3l:
I merge this now, but I think the alpaka issue should have been linked as annotated. If you like to add this, pls do in a follow-up PR.

include/picongpu/plugins/output/WriteSpeciesCommon.hpp
```diff
@@ -145,16 +145,20 @@ struct FreeMemory
...
+ * Therefore we need to call the native CUDA function cudaFreeHost to free the memory.
+ * Due to the renaming of cuda functions with cupla via macros we need to remove
+ * the renaming to get access to the native cuda function.
+ * @todo this is a workaround plese fix me. We need to investigate if
+ * it is possible to have mapped/unified memory in alpaka.
+ */
+#   undef cudaFreeHost
```
ax3l: I would guard this with a `#ifdef`

include/pmacc/memory/buffers/MappedBufferIntern.hpp
```diff
@@ -69,7 +73,21 @@ class MappedBufferIntern : public DeviceBuffer<TYPE, DIM>
...
+ * Therefore we need to call the native CUDA function cudaFreeHost to free the memory.
+ * Due to the renaming of cuda functions with cupla via macros we need to remove
+ * the renaming to get access to the native cuda function.
+ * @todo this is a workaround plese fix me. We need to investigate if
+ * it is possible to have mapped/unified memory in alpaka.
+ */
+#   undef cudaFreeHost
```
ax3l: I would guard this with an `#ifdef cudaFreeHost`

include/picongpu/plugins/output/WriteSpeciesCommon.hpp
ax3l: typo: `cupla 0.1.0 does not support ...`

include/picongpu/plugins/output/WriteSpeciesCommon.hpp
```diff
@@ -145,16 +145,20 @@ struct FreeMemory
...
         {
 #if( PMACC_CUDA_ENABLED == 1 )
-            auto rc = cudaFreeHost(ptr);
-            /* cupla can't handle foreign memory allocated with `cudaHostAlloc`
-             * therefore the cupla error `cuplaErrorMemoryAllocation` is ignored
-             */
-            if(rc != cuplaErrorMemoryAllocation)
-                CUDA_CHECK(rc)
+/* cupla is not supporting the function cudaHostAlloc to create mapped memory.
+ * Therefore we need to call the native CUDA function cudaFreeHost to free the memory.
```
ax3l: need -> have
free ~~the~~ memory

include/picongpu/plugins/output/WriteSpeciesCommon.hpp
```diff
@@ -145,16 +145,20 @@ struct FreeMemory
...
+ * Due to the renaming of cuda functions with cupla via macros we need to remove
+ * the renaming to get access to the native cuda function.
+ * @todo this is a workaround plese fix me. We need to investigate if
```
ax3l: ple**a**se

include/pmacc/memory/buffers/MappedBufferIntern.hpp
```diff
@@ -69,7 +73,21 @@ class MappedBufferIntern : public DeviceBuffer<TYPE, DIM>
 
         if (pointer && ownPointer)
         {
-            CUDA_CHECK(cudaFreeHost(pointer));
+#if( PMACC_CUDA_ENABLED == 1 )
+/* cupla is not supporting the function cudaHostAlloc to create mapped memory.
+ * Therefore we need to call the native CUDA function cudaFreeHost to free the memory.
+ * Due to the renaming of cuda functions with cupla via macros we need to remove
+ * the renaming to get access to the native cuda function.
+ * @todo this is a workaround plese fix me. We need to investigate if
```
ax3l: same typos as above :)

include/picongpu/plugins/output/WriteSpeciesCommon.hpp
ax3l: The reason why one would guard a `def` or `undef` is to make it more flexible for other hacks/workarounds that might (un)set.

You can keep it as is under the mentioned assumption, but that would make it more flexible.

Issue #2687, opened 2018-08-15
KHI Heating test off?


psychocoderHPC:
I run a heating test for KHI with electrons/positrons and saw confusing results:
used branch: https://github.com/ComputationalRadiationPhysics/picongpu/commit/c4ee99340720f4ccacc25f042799cba8661fbd99

it should be look like this:
![28874812-bb5009ce-7793-11e7-833e-1a1e36771c35](https://user-images.githubusercontent.com/4037125/44148298-105d1910-a098-11e8-838e-491c1a296abe.png)
copied from : https://github.com/ComputationalRadiationPhysics/picongpu/pull/2168#issuecomment-319665861

but is: 
![dev_heat](https://user-images.githubusercontent.com/4037125/44148340-3fc61eb8-a098-11e8-8395-b28a1a5e49f7.png)

The error is by a factor 100 to large.

psychocoderHPC:
I will check now the 0.4.0-rc

PrometheusPi:
@kossag14 This might affect you and could explain why the injection is lost in the bubble.

psychocoderHPC:
I discussed this with @ax3l and he pointed me to the bug fix in #2112. It could be that the results are correct and older plots was wrong due to the bug #2112.

psychocoderHPC:
I found the BUG the file for the field energies is names `field_energy.dat` and since we have particle filter the energy files for particles with the fault filtering (`All`) are called `<SPECIESNAME>_energy_all.dat`. The tool `plotNumericalHeating` is changed to check only files with fits the regex `*_energy_all.dat` therefore field energies are not taken into account.

PrometheusPi:
The initial temperature setting is correct #2698.

psychocoderHPC:
solved with #2697

PR #2697, opened 2018-09-03
fix plotting tool for numerical heating

psychocoderHPC:
fix: #2687

The field energy is ignored/not used in the tool `plotNumericalHeating`.
This results into a wrong heating plot.

- add usage of `fields_energy.dat`
- fix documentation

### DIFF
--- a/src/tools/bin/plotNumericalHeating
+++ b/src/tools/bin/plotNumericalHeating
@@ -25,7 +25,7 @@ __doc__ = '''
 This program compares the energy evolution of two PIConGPU runs.
 Just give the directory of two PIConGPU runs and this program will
 get all values needed if they are available.
-You should activate `--energy_fields` and any `--energy_X` (X any species)
+You should activate `--fields_energy` and any `--<species>_energy` (for any species)
 for both simulation runs with the same dumping period.
 A different number of particle species between both simulations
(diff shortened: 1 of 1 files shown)


src/tools/bin/plotNumericalHeating
```diff
@@ -119,7 +119,7 @@ for sim in directories:
     mydir = sim+simDir
     # get relevant files with energy
     files = [f for f in os.listdir(mydir)
-             if os.path.isfile(os.path.join(mydir, f)) and re.search('^.*_energy_all.dat', f)]
+             if os.path.isfile(os.path.join(mydir, f)) and (re.search('^.*_energy_all.dat', f) or re.search('^.fields_energy.dat', f))]
     # check if file list is empty
     if len(files) == 0:
         sys.exit("There were no energy files in \"{}\".".format(mydir))
```
ax3l: I wonder if we should also at least print a warning of the field energy is missing.

psychocoderHPC: But in the case where I will only plot the lose of energy of an moving particle species (maybe one particle) it is not needed.
A suggestion is to rename this to something like `plotGlobalHeating` and enforce at least one particle species and the field energy.

PR #2700, opened 2018-09-05
fix gameOfLife

psychocoderHPC:
The game of life example has a runtime issue that the mapping description verification
to check the number of supercells is always twrowing an error. The reason it that the
not existing default constructor is `MappingDescription` is used.
Please do not ask me why the code before is compiling. The class `MappingDescription` has no default constructor but is a not initialized member of the class `Evolution<>` which means by definition the default constructor should be called. Due to the fact that there is a user defined constructor there should be no default constructor available. 

Error message:
```
./gameOfLife -d  1 1  -g 128 128 -p 1 1 -r 23/3 -s 100
newborn if=3 stay alive if=23 mask=4108
  what():  expression (( guardingSuperCells[ d ] == 0 && gridSuperCells[ d ] >= 1) || gridSuperCells[ d ] >= 2 * guardingSuperCells[ d ] + 1) failed in file (/include/pmacc/mappings/kernel/MappingDescription.hpp:82) : 
Aborted
```

- create the object for mapping description inside the init method of `Evolution`
- change interface of `Evolution<...>::init()`
- remove the default constructor for `MappingDescription` (it make no sense to allow a mapping with zero cells)

### DIFF
--- a/include/pmacc/mappings/kernel/MappingDescription.hpp
+++ b/include/pmacc/mappings/kernel/MappingDescription.hpp
@@ -57,7 +57,7 @@ public:
      * @param guardingSuperCells number of **supercells** within the guard
      */
     MappingDescription(
-        DataSpace<DIM> localGridCells = DataSpace<DIM> (),
+        DataSpace<DIM> localGridCells,
         DataSpace<DIM> guardingSuperCells = DataSpace<DIM>::create(0)
     ) :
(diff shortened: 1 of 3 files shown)


sbastrakov:
@psychocoderHPC the constructor of `MappingDescription` has default values for both parameters, so it is also a default constructor and therefore previously the code has compiled successfully.

psychocoderHPC:
I removed the default constructor for `MappingDescription`

PR #2706, opened 2018-09-25
Boost Template Aliases: Fix Workaround

ax3l:
The workaround disabling template aliases causes compile errors on Titan (ORNL), both when compiling with CPU backend or CUDA 9.1.85.

Modern versions of boost need template aliases, otherwise e.g `boost/type_traits` fails with:
```
include/boost/type_traits/detail/has_binary_operator.hpp:50:61: error: 'void_t' in namespace 'boost' does not name a template type
       template <class T, class U, class Ret, class = boost::void_t<>>
                                                             ^
include/boost/type_traits/detail/has_binary_operator.hpp:50:67: error: expected '>' before '<' token
       template <class T, class U, class Ret, class = boost::void_t<>>
```

Seen with Boost 1.67.0.

Applies the following logic changes:
- only enable when compiling with NVCC for CUDA backend
- only enable when NVCC < 9.1
- NVCC 8.0 known to be broken
- NVCC 9.0 unknown, maybe fixed this already:
  https://svn.boost.org/trac10/ticket/11897#comment:22

Note: `FindCUDA` only provides CUDA up to two version digits. So we can only compare `<9.1` and not `< 9.1.85`.


ax3l:
I am currently seeing this compile error again with:
- PIConGPU 0.4.3
- GCC 5.4.0
- Boost 1.68.0
- CUDA 9.0.176

in our Docker build.

PR #2705, opened 2018-09-25
fix misspelled words in plugin documentation

PrometheusPi:
This pull request fixes various misspelled words in the read-the-docs plugin documentation.

Some errors pointed to further mistakes in the radiation plugin, which have been fixed too.

Issue #2250, opened 2017-09-13
Vote: New Subtitle


ax3l:
Dear @ComputationalRadiationPhysics/picongpu-maintainers ,

we need a new subtitle for our docs and GitHub that describes PIConGPU **in a short one-liner**.

Currently, we use *PIConGPU - A particle-in-cell code for GPGPUs* but this is now a [tremendous understatement](https://github.com/ComputationalRadiationPhysics/picongpu/projects/5).

The idea is to keep the one-liner **short and striking** as before, think of it as a *useful addition* to the name highlighting *more essential characteristics* of our capabilities. Since *PIConGPU* is a well-known brand by now (and GPUs are still the fastest/sexiest general purpose hardware around), the subtitle helps to show that we are also fit for any other hardware (with the upcoming release(s) in a few months) but still keeps our most-successful flavor.

Please make suggestions and/or vote here.

let's start:

*PIConGPU - ...*
- *An exascale-ready particle-in-cell code*: (both describes platforms, I/O and in-situ capabilities and seems honest)
- *A particle-in-cell code for exascale*: (same as above, but more bold)
- *A particle-in-cell code for GPUs, CPUs and beyond*
- *A particle-in-cell code for the manycore era*
- *The performance-portable particle-in-cell code [for exascale]*
- ...

Now you!

(No, [*PICy McPICface*](https://en.wikipedia.org/wiki/Boaty_McBoatface) is not an option ;) )

ccing @ComputationalRadiationPhysics/picongpu-developers @BenjaminW3

n01r:
How about:  *PIConGPU - Performance-portable exascale-ready plasma simulations* ?

... or use *many-core architectures* in the subtitle?

ax3l:
*plasma simulations for the manycore era* ?

n01r:
Also not bad - I like leaving out an article like "A" or the "The" because one is modest and the other bold but we need neither

ax3l:
I find the last update sounds somewhat doubled/too verbose with both the specific *manycore* and the reference *exascale* (which *could* also be something different in the end)

theZiz:
_... for GPU and many more many-core architectures_

berceanu:
PIConGPU - Inherently parallel plasma simulations.

ajitup73:
PIConGPU - An exascale particle-in-cell code for hybrid computing architectures
PIConGPU - An exascale ready particle-in-cell code for hybrid clusters

[...]

PR #2710, opened 2018-09-28
Further improve wording in docs

sbastrakov:
A little inspired by #2705 , I've read a little portion of the docs as a bedtime literature (no PIC-related dreams, in case someone was wondering) and made some notes.

A short comment on some of the fixes: the installation guide had an absurd amount of `you` and `yours`, reduced somewhat, also 'sub-ideal' is not a word (although would be useful).

In addition to the fixes in the PR, here are things I found strange, but not sure whether or how to fix.

In 'How to Read This Document':
- It says the code is 3D3V, should 2D3V be added?
- `It describes the dynamics of a plasma by computing the motion of electrons and ions in the plasma based on Maxwell’s equations.` is confusing to me. Firstly, one might get an impression the particles move due to Maxwell's equations. Secondly, the grid or fields are not mentioned at all.
- Is it on purpose that bold and italic are both used for highlighting? If so, I don't see an obvious distinction.

In INSTALL.rst: broken or incorrectly formated link in splash2txt section.

I plan to gradually continue, so WIP.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -21,7 +21,7 @@ Overview
    :alt: overview of PIConGPU library dependencies
 
    Overview of inter-library dependencies for parallel execution of PIConGPU on a typical HPC system. Due to common binary incompatibilities between compilers, MPI and boost versions, we recommend to organize software with a version-aware package manager such as `spack <https://github.com/spack/spack>`_ and to deploy a hierarchical module system such as `lmod <https://github.com/TACC/Lmod>`_.
-   A Lmod example setup can be found `here <https://github.com/ComputationalRadiationPhysics/compileNode>`_.
+   An Lmod example setup can be found `here <https://github.com/ComputationalRadiationPhysics/compileNode>`_.
 
 Requirements
 ------------
@@ -217,7 +217,7 @@ libSplash
 HDF5
 """"
 - 1.8.6+
-- standard shared version (no c++, enable parallel)
+- standard shared version (no C++, enable parallel)
 - *Debian/Ubuntu:* ``sudo apt-get install libhdf5-openmpi-dev``
 - *Arch Linux:* ``sudo pacman --sync hdf5-openmpi``
 - *Spack:* ``spack install hdf5~fortran``
@@ -260,7 +260,7 @@ splash2txt
 png2gas
 """""""
 - requires *libSplash*, *pngwriter* and *boost* ``program_options``)
-- converts png files to hdf5 files that can be used as an input for a species initial density profiles
+- converts png files to hdf5 files that can be used as an input for species initial density profiles
 - compile and install exactly as *splash2txt* above
 
 ADIOS
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -9,16 +9,16 @@
 
 *Particle-in-Cell Simulations for the Exascale Era*
 
-PIConGPU is a fully relativistic, manycore, 3D3V particle-in-cell (PIC) code.
+PIConGPU is a fully relativistic, manycore, 3D3V and 2D3V particle-in-cell (PIC) code.
 The PIC algorithm is a central tool in plasma physics.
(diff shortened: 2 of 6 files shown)


INSTALL.rst
```diff
@@ -217,7 +217,7 @@ libSplash
 HDF5
 """"
 - 1.8.6+
-- standard shared version (no c++, enable parallel)
+- standard shared version (no C++, enable parallelism)
```
ax3l: not in this context, since "parallel" is a keyword.
change to `enable parallel IO` please

docs/source/index.rst
```diff
@@ -9,16 +9,16 @@
 
 *A particle-in-cell code for GPGPUs*
 
-PIConGPU is a fully relativistic, many GPGPU, 3D3V particle-in-cell (PIC) code.
+PIConGPU is a fully relativistic, manycore, 3D3V particle-in-cell (PIC) code.
```
ax3l: please keep the title, this is intentional a play of words.
new title is discussed in #2250

ax3l: Ah true, I see. It still changes meaning but I am fine with that.

ax3l:
@sbastrakov can you please rebase and finish this PR?

sbastrakov:
Sure, will do tomorrow.

ax3l:
ping @sbastrakov gentle reminder.

tbh, as procrastination tasks I would be more excited if one would modernize the code base to use modern C++11 constructs, copy elision, not throw in destructors #2666, etc., e.g. following [clang-tidy](https://picongpu.readthedocs.io/en/0.4.0/dev/clangtools.html) hints

sbastrakov:
Ready for review.

I did not touch sections in the second half, as some of them have been modified in 0.4.1, but these changes are currently not in dev.

docs/source/usage/reference.rst
```diff
@@ -5,7 +5,7 @@ Reference
 
 .. sectionauthor:: Axel Huebl
 
-PIConGPU is a year-long, scientific project with many people contributing to it.
+PIConGPU is an almost decade-long scientific project with many people contributing to it.
```
sbastrakov: Note: previously used `year-long` is wrong here, since it means 1 year. In case you don't like my version, at least change to `multiyear`.

INSTALL.rst
```diff
@@ -217,7 +217,7 @@ libSplash
 HDF5
 """"
 - 1.8.6+
-- standard shared version (no c++, enable parallel)
+- standard shared version (no C++, enable parallel)
```
ax3l: I tried to use keywords here, but this is ok.
The user does *not* need the partially implemented C++ APIs (`--enable-cxx`) but parallel APIs (`--enable-parallel`).

docs/source/install/path.rst
```diff
@@ -40,21 +40,21 @@ Build from Source
...
-Performance will be ideal if architecture is chosen correctly (and/or if build directly on your hardware).
+Performance will be ideal if architecture is chosen correctly (and/or if built directly on your hardware).
 You then set environment variables to find those installs.
 
 Conda
 ^^^^^
 
 We currently do not have an official conda install (yet).
-Due to pre-build binaries, performance will be sub-ideal and HPC cluster support (e.g. MPI) might be very limited.
+Due to pre-build binaries, performance could be not ideal and HPC cluster support (e.g. MPI) might be very limited.
```
ax3l: please keep `will` instead of `could`. Pre-build binaries will almost always not be tuned for the sub-architecture which will kill a lot of vectorization potential. I don't want anyone complaining about performance when using pre-build binaries, so let us be over-explicit/conservative in this teaser please.

ax3l: Ah anyway, we don't have them yet and we can complain when we do. I am fine with your version.

docs/source/usage/reference.rst
ax3l: Like it, well done thx.

PR #2711, opened 2018-09-28
add MPI linking info in install documentation

PrometheusPi:
This pull request adds an additional bullet point for building hdf5 from source. Based on the discussion on #2667, it is a quite common issue to run into linking errors with MPI when configuring HDF5. This can be solved by specifying to use the MPI compilers instead of the default ones (set e.g. by the GCC/ICC module). This new bullet point adds this missing information.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -234,7 +234,7 @@ HDF5
   - ``make``
   - *optional:* ``make test``
   - ``make install``
-  - if you encounter errors during ``./configure`` related to linking MPI you might try setting ``export CC=mpicc`` and ``export CXX=mpic++`` before running ``./configure``
+  - If you encounter errors related to linking MPI during ``./configure``, you might try setting ``export CC=mpicc`` and ``export CXX=mpic++`` before running ``./configure``.
 - *environment:* (assumes install from source in ``$HOME/lib/hdf5``)
 
   - ``export HDF5_ROOT=$HOME/lib/hdf5``
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -234,6 +234,7 @@ HDF5
   - ``make``
   - *optional:* ``make test``
   - ``make install``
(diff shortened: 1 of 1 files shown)


PrometheusPi:
@sbastrakov You are right, the comma was missing. I fixed it now.

ax3l:
Users usually run those instructions to install a whole bunch of software, on which an export of CC/CXX will influence subsequent unrelated packages as well.

Can you please check if instead the HDF5 configure command knows a CC and CXX argument? Or if simple CC=... CXX=... ./configure serves the same purpose?

ax3l:
@PrometheusPi ping

ax3l:
@PrometheusPi ping

ax3l:
@PrometheusPi ping, please verify and update the PR :)

PrometheusPi:
@ax3l I am on it right now - using the interface directly is a better option. I adjusted the install description accordingly. Right now, I am still testing the entire compile process which takes about an hour. I will inform you if it does not work.

INSTALL.rst
```diff
@@ -234,6 +234,7 @@ HDF5
   - ``make``
   - *optional:* ``make test``
   - ``make install``
+  - If you encounter errors related to linking MPI during ``./configure``, you might try setting the compiler manually via ``./configure --enable-parallel --enable-shared --prefix $HOME/lib/hdf5/  CC=mpicc CXX=mpic++``.
```
ax3l: there is an unnecessary double whitespace before CC.

note: when testing, double check `CC` is not `export`-ed from a previous session to the same value :)

PrometheusPi: fixed white space

PrometheusPi: Install and most tests went well. 
Both are not exported to the previous session (still `gcc` and `g++`).

ax3l: thanks for testing!

PR #2712, opened 2018-10-04
fix tbg error message

psychocoderHPC:
If the path to `egetopt` is not in the environment variable `PATH` tbg falls back to use the `egetopt` in the same location as `tbg` itself. In that case the command `which` is throwing a long error message that egetopt can not be found.
This PR hide the error message and create an own error message if needed.

### DIFF
--- a/bin/tbg
+++ b/bin/tbg
@@ -275,14 +275,14 @@ help()
 initCall="$0 $*"
 projectPath="."
 
-pathToegetopt=`which egetopt`
+pathToegetopt=$(which egetopt 2>/dev/null)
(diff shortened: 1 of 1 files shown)


bin/tbg
```diff
@@ -275,7 +275,7 @@ help()
 initCall="$0 $*"
 projectPath="."
 
-pathToegetopt=`which egetopt`
+pathToegetopt="$(which egetopt 2>/dev/null)"
```
ax3l: why the `"` quotes?

psychocoderHPC: because it is allowed. But I will remove it.

PR #2718, opened 2018-10-08
Fix sliceFieldReader

steindev:
Use of `map()` changed from Python 2 to 3 which results in an error when using the function now. This PR avoids the map function but uses a (more pythonic) list comprehension which is also version independent.

A grammar error in a comment is corrected as well.

### DIFF
--- a/lib/python/picongpu/plugins/sliceFieldReader.py
+++ b/lib/python/picongpu/plugins/sliceFieldReader.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2018 Richard Pausch
+# Copyright 2014-2018 Richard Pausch, Klaus Steiniger
 #
 # This file is part of PIConGPU.
 #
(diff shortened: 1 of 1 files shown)


lib/python/picongpu/plugins/sliceFieldReader.py
```diff
@@ -77,7 +77,7 @@ def readFieldSlices(File):
         # go through all valid field vectors
         for x in range(N_x):
             fieldValue = line[x]
-            data[y, x, :] = map(float, fieldValue[1:-1].split(','))
+            data[y, x, :] = [ float(x) for x in fieldValue[1:-1].split(',') ]
```
ax3l: looks good!
Please just [fix the style](https://travis-ci.org/ComputationalRadiationPhysics/picongpu/jobs/438549431) by removing the whitespaces after `[` and before `]` to make the CI happy :)

PR #2722, opened 2018-10-11
cuda_memtest usage on hypnos

psychocoderHPC:
`cuda_memtest` can currently only be used if the node is exclusive allocated by the user and no other job is running on it.
All hypnos template files for hypnos supports that the user can allocate only a subset of a node. In this case `cuda_memtest` will crash if an other user/job is already using gpu zero.
This PR will disable the GPU memory test if the job has not alloacted all GPUs.

- remove PBS option `-n` (exclusive node usage) because it is not working on hypnos
- disable the test `cuda_memtest` if node is not exclusive allocated

It would be possible to mark this PR as `bug` because the way how `cuda_memtest` is used is wrong.

# Tests

- [x] job allocates all GPUs
- [x] job allocate only the half node

### DIFF
--- a/etc/picongpu/hypnos-hzdr/k20.tpl
+++ b/etc/picongpu/hypnos-hzdr/k20.tpl
@@ -29,7 +29,6 @@
 # send me mails on job (b)egin, (e)nd, (a)bortion or (n)o mail
 #PBS -m !TBG_mailSettings -M !TBG_mailAddress
 #PBS -d !TBG_dstPath
-#PBS -n
 
 #PBS -o stdout
 #PBS -e stderr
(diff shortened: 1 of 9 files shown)


etc/picongpu/hypnos-hzdr/k20.tpl
```diff
@@ -75,11 +77,11 @@ cd simOutput
...
 sleep 1
 
-# test if cuda_memtest binary is available
-if [ -f !TBG_dstPath/input/bin/cuda_memtest ] ; then
+# test if cuda_memtest binary is available and we have the node exclusive
+if [ -f !TBG_dstPath/input/bin/cuda_memtest ] && [ !TBG_numHostedGPUPerNode -eq !TBG_gpusPerNode ] ; then
   mpiexec --prefix $MPIHOME -tag-output --display-map -x LIBRARY_PATH -am !TBG_dstPath/tbg/openib.conf --mca mpi_leave_pinned 0 -npernode !TBG_gpusPerNode -n !TBG_tasks !TBG_dstPath/input/bin/cuda_memtest.sh
 else
-  echo "no binary 'cuda_memtest' available, skip GPU memory test" >&2
+  echo "no binary 'cuda_memtest' available or compute node is not exclusive allocated, skip GPU memory test" >&2
```
ax3l: in all files a little typo: *exclusively* allocated

psychocoderHPC:
fixed

PR #2725, opened 2018-10-11
RTD: Sphinx 1.7

ax3l:
Fix readthedocs sphinx to version 1.7 since the latest
1.8 release line has a bug rendering overloaded C++ code with breathe.

Upstream issue: https://github.com/sphinx-doc/sphinx/issues/5496


ax3l:
cc @theZiz this also affects you with llama

PR #2727, opened 2018-10-15
Shorter python imports for postprocessing tools

codingS3b:
In order to shorten the import statements when using python visualizers, they are now included within the `plot_mpl/__init__.py` with meaningful aliases (since they are all called Visualizer).

Instead of writing
`from picongpu.plugins.plot_mpl.phase_space_visualizer import Visualizer as PhaseSpaceVisualizer ` one could then shorten this to `from picongpu.plugins.plot_mpl import PhaseSpaceVisualizer`


lib/python/picongpu/plugins/__init__.py
```diff
@@ -0,0 +1,8 @@
+from .energy_histogram import EnergyHistogram
+from .phase_space import PhaseSpace
+from .png import PNG
+
+
+__all__ = ["EnergyHistogram",
+           "PhaseSpace",
+           "PNG"]
```
psychocoderHPC: missing EOF

lib/python/picongpu/plugins/plot_mpl/__init__.py
```diff
@@ -0,0 +1,9 @@
+from .energy_histogram_visualizer import Visualizer as \
+    EnergyHistogramVisualizer
+from .phase_space_visualizer import Visualizer as PhaseSpaceVisualizer
+from .png_visualizer import Visualizer as PNGVisualizer
+
+
+__all__ = ["EnergyHistogramVisualizer",
```
ax3l: I would not suffix them with `Visualizer`

codingS3b: No, not really. I would also have to read up on that. 
Naming collisions were the main reason for suffixing them this way.

ax3l: Ok, but then let us at least Suffix them with `PlotMPL` or just `MPL`

codingS3b: Ok, I changed the suffix to 'MPL'

PR #2730, opened 2018-10-16
Python: base class for data readers

codingS3b:
Adds a base class for the data readers to provide a unified interface.
Maybe some general logic could already be included there but I am not sure about this since the derived classes do not really share much code. I am happy to discuss this.


ax3l:
Looks ok for me so far.
Can we also move them all into a `data/` sub-directory with this PR?

PR #2739, opened 2018-10-18
RTD: Mock Sphinx Imports

ax3l:
Mock imports during sphinx build that are not essential. Allows to parse doc strings without the need of accessing and installing otherwise unneeded third-party Python modules.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -49,6 +49,12 @@ if not on_rtd:
 
 # napoleon autodoc config
 napoleon_include_init_with_doc = True
+autodoc_mock_imports = [
+    'h5py',
+    'pandas',
+    'ipywidgets',
+    'ipympl',
+]
 
 # breathe config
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -49,6 +49,12 @@
 
 # napoleon autodoc config
 napoleon_include_init_with_doc = True
+autodoc_mock_imports = [
```
ax3l: everything inslide `lib/python/` that is `imported` and not already needed in `docs/requirements.txt`

lib/python/picongpu/plugins/data/requirements.txt
```diff
@@ -2,4 +2,3 @@ numpy
 pandas>=0.21.0
 h5py
 scipy
--r plot_mpl/requirements.txt
```
ax3l: outdated before modules were moved

PR #2747, opened 2018-10-19
Add .zenodo.json

ax3l:
Make the release, reference, authorship and grant process more transparent. Please add new contributors and creators (authors) as appropriate in the future.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,0 +1,120 @@
+{
+    "creators": [
+      {
+        "affiliation": "HZDR, TU Dresden",
+        "name": "Huebl, Axel",
+        "orcid": "0000-0003-1943-7141"
+      },
+      {
+        "affiliation": "HZDR",
+        "name": "Widera, René",
+        "orcid": "0000-0003-1642-0459"
+      },
+      {
+        "affiliation": "LogMeIn, Inc.",
+        "name": "Worpitz, Benjamin"
+      },
+      {
+        "affiliation": "HZDR, TU Dresden",
+        "name": "Pausch, Richard",
+        "orcid": "0000-0001-7990-9564"
+      },
+      {
+        "affiliation": "HZDR, TU Dresden",
+        "name": "Burau, Heiko"
+      },
+      {
+        "affiliation": "HZDR, TU Dresden",
+        "name": "Garten, Marco",
+        "orcid": "0000-0001-6994-2475"
+      },
+      {
+        "affiliation": "HZDR",
+        "name": "Starke, Sebastian"
+      },
+      {
+        "affiliation": "HZDR, TU Dresden",
+        "name": "Grund, Alexander",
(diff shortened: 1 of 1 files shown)


.zenodo.json
```diff
@@ -0,0 +1,120 @@
...
+        "affiliation": "LogMeIn, Inc.",
+        "name": "Worpitz, Benjamin"
+      },
+      {
+        "affiliation": "HZDR, TU Dresden",
+        "name": "Pausch, Richard",
+        "orcid": "0000-0001-7990-9564"
+      },
+      {
+        "affiliation": "HZDR, TU Dresden",
```
PrometheusPi: isn't his current affiliation: Max Planck Institute for the Physics of Complex Systems

see: https://www.pks.mpg.de/dynamics-in-correlated-quantum-matter/people/heiko-burau/

.zenodo.json
```diff
@@ -0,0 +1,120 @@
...
+        "name": "Garten, Marco",
+        "orcid": "0000-0001-6994-2475"
+      },
+      {
+        "affiliation": "HZDR",
+        "name": "Starke, Sebastian"
+      },
+      {
+        "affiliation": "HZDR, TU Dresden",
```
PrometheusPi: should HZDR be still there?

.zenodo.json
ax3l: I just exported the current authors for now, thanks for the hint! When he contributed the work for the last release, that was his affiliation. We usually would then just update the affiliation for recurring authors over time.

Even if he does not actively contribute anymore, @Heikman contributed a significant part to the codebase that will obviously conserve his co-authorship. In this specific case, it could therefore be a good compromise to keep old and new affiliations at the time to reflect that corner-case.

.zenodo.json
ax3l: yes, contributed to PIConGPU while under HZDR employment during his MA/Dipl thesis at TU DD.

psychocoderHPC:
I assigned @PrometheusPi because he blocked this PR and to be sure that the questions are solved he need to merge it.

.zenodo.json
PrometheusPi: solved - that's fine with me

.zenodo.json
PrometheusPi: @ax3l So do you want to update @Heikman 's affiliation **now** - or **later**?

ax3l: if he doesn't respond currently, we merge it as it for now.

ax3l:
@PrometheusPi can be merged for now, we can update on author request later

.zenodo.json
ax3l: ping @Flamefire : whatever you prefer (see comments above)

Flamefire: I don't mind. Not sure what the correct interpretion of this is so I'd go with "at the time contributed"

PR #2754, opened 2018-10-22
PhaseSpaceData: allow multiple iterations

codingS3b:
This adresses a TODO in the `PhaseSpaceData` class to allow for queries of a list of iterations.
It also changes the parameter order of the interface calls by moving the `ps` argument to the front and providing a default to `species_filter` as in the other reader classes.

### DIFF
--- a/docs/source/usage/plugins/phaseSpace.rst
+++ b/docs/source/usage/plugins/phaseSpace.rst
@@ -73,14 +73,15 @@ You can quickly load and interact with the data in Python with:
 
    # load data
    ps_data = PhaseSpaceData('/home/axel/runs/lwfa_001')
-   ps, meta = ps_data.get(species='e', species_filter='all', ps='ypy', iteration=2000)
+   # a list of iterations is also possible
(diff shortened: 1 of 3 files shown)


codingS3b:
I just updated the documentation for the reader and noted that one still has to set the `mu` and `e_mc_r` variables manually instead of getting them from the readers `meta` object. Is this intended or can this be handled by the meta object as well? 
As a result of this, so far the `plot_mpl` visualizer also has to set those variables as member variables manually (even though they are probably only correct for `species='e'`, right?). Maybe this should be changed but I am lacking the physical knowledge of how to do it correctly :(

ax3l:
@codingS3b can you please squash similar commits and rebase?

> still has to set the mu and e_mc_r variables manually

That's fine in general, since we return them in one unit system and just show in the example usage how to convert it into something different.

> the plot_mpl visualizer also has to set those variables

ok if the `plot_mpl` visualizer chooses to represent these in another unit system as well.

ax3l:
Thanks! :sparkles:

PR #2756, opened 2018-10-24
fix outdated derive species

PrometheusPi:
This syntax is obsolete with the new version (0.4.0). 
Since it was not captured by the compile suite, this case is probably never checked. 
Should we add it to the compile suite?


PrometheusPi:
@ax3l I am running in reproducible travis-ci issues with this pull request. It throws errors with pythons scripts I did not touch. Could you have a look at this?

ax3l:

PR #2757, opened 2018-10-24
deactivate current in Bunch example

PrometheusPi:
With the new version, the current deposition is activated in the bunch example.
This however breaks the electron propagation due to a wrong (non charge neutral) initialization. 

Thanks @lehnertu for finding this bug!


PrometheusPi:
sorry - in line error - started from master

ax3l:
You can click `Edit` to change the PR base branch without reopening it :)

ax3l:
Ah, I see :+1:

PR #2758, opened 2018-10-24
remove current deposition in bunch example

PrometheusPi:
This pull request fixes a bug in the bunch example.
Thank you @lehnertu for discovering it!

The current deposition should be deactivated for this example since otherwise the bunch gets pulled back by the remaining "virtual" ions.


ax3l:

PR #2762, opened 2018-10-24
Python MPL Visualizer: plot for several simulations

codingS3b:
This PR allows the MPL visualizers base class to accept a list of `run_directories` and can deal with plotting several plots in the same axis. 
Since this would be problematic for PNG and PhaseSpace objects, they still require a run_directory string instead of a list and will complain otherwise.
So the only class for which this makes sense so far is the `EnergyHistogramMPL`.

In addition for code cleanup reasons, the `create_data_reader` function is removed and instead the base class accepts now a mandatory `reader_cls` parameter.


codingS3b:
Please don't merge yet, I might still have to do some changes to the interface

lib/python/picongpu/plugins/plot_mpl/base_visualizer.py
```diff
@@ -6,7 +6,9 @@
 License: GPLv3+
 """
 
+
```
jkelling: Maybe remove additional blank line.

ax3l:
I think the `PhaseSpace` plugin can be done for multiple runs. We just need to take a colorbar that goes to transparent and overplot them in-order. I do this quite often, we just need different colors per run_dir:

```python
import matplotlib as mpl
from matplotlib.colors import colorConverter


myCmaps = []
colors = mpl.pyplot.rcParams['axes.prop_cycle'].by_key()['color']

# generate the colors for colormaps
def generateCmaps():
    for s in colors.keys():
        myColor1 = colorConverter.to_rgba('white', alpha=1.0)
        myColor2 = colorConverter.to_rgba(colors[s], alpha=1.0)
        myCmap = mpl.colors.LinearSegmentedColormap.from_list(
            'my_cmap_' + s, [myColor1,myColor2], 1200)
        myCmap._init() # create the _lut array, with rgba values
        myCmaps.append(myCmap)

plt.imshow(
    # ...
    cmap=myCmaps['0']
)
```

codingS3b:
Yes, `eh_vis_one = EnergyHistogramMPL( run_dir)` also still works. I also added support for plotting multiple phase space data from the code you suggested. However, there are still a few issues for me (since colorbars are somewhat ugly to deal with):
- your example code allowed for a max of 10 simulations (which should be sufficient anyway) since `plt.rcParams['axes.prop_cycle'].by_key()['color']` provides only 10 colors (at least in my case), so I changed that to allow an arbitrary number of simulations each with an individual colormap but with the price that the generated colormaps might not be that well to distinguish...
- when switching from an iteration for which data was available for all inputs (and hence we had multiple colorbars) to an iteration for which data is unavailable for some inputs (and hence we get fewer colorbars) the drawing area of the plot is enlarged and the remaining colorbars are now shown within the figure and I don't know how to control that (see picture)

colorbar_bug

lib/python/picongpu/plugins/plot_mpl/phase_space_visualizer.py
```diff
@@ -21,55 +22,89 @@ class Visualizer(BaseVisualizer):
...
             aspect='auto',
             origin='lower',
-            norm=LogNorm())
-        self.cbar = plt.colorbar(self.plt_obj, ax=self.ax)
-        self.cbar.set_label(
-            r'$Q / \mathrm{d}r \mathrm{d}p$ [$\mathrm{C s kg^{-1} m^{-2}}$] ')
+            norm=LogNorm(),
+            cmap=self.colors[idx])
+        # create the colorbar
+        self.colorbars[idx] = plt.colorbar(self.plt_obj[idx], ax=self.ax)
```
ax3l: if you set an alpha < 1.0, maybe also do [this work-around (see *Notes* section)](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.colorbar.html#matplotlib.pyplot.colorbar) by default:
```python
self.colorbars[idx].solids.set_edgecolor("face")
```

ax3l: I think if you assign an explicit *colorbar axes* with `cax` your iteration layout problem might be addressable. Nevertheless, that would mean you have to manually place axes for colorbars.

codingS3b:
@ax3l  Okay thanks, it should be ready now!

lib/python/picongpu/plugins/data/base_reader.py
```diff
@@ -105,8 +105,9 @@ def get(self, **kwargs):
                 # use all times that are available, i.e. all iterations
                 iteration = self.get_iterations(**kwargs)
             else:
-                iteration = self.find_time.get_iteration(time)
-            print("got 'time'=", time, ", converting to iteration", iteration)
+                iteration = self.find_time.get_iteration(
+                    time, method='closest')
```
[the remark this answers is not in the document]
codingS3b: Yes this was necessary and rather a bug in the previous version.

PR #2764, opened 2018-10-24
adjust example for workshop

PrometheusPi:
**Do NOT merge!**

@ax3l This is the example for session 8.


PrometheusPi:
@ax3l I will close this pull request for now. It is probably not being included as a new or updated example in the mainline.

PR #2766, opened 2018-10-25
fix 2D case of single electron init (via density)

PrometheusPi:
The single electron case should also work in 2D, but one vector reduction (to 2 dimensions) is missing and causes a compile error. This pull request fixes this bug.


ax3l:
Thanks! Applied to the workshop `$PICSRC/` that our users are currently using

PR #2768, opened 2018-10-25
fix EveryNthCellImpl

### LINKED ISSUE #2763: Probe particles setup - too many particles
The example in the [documentation](https://picongpu.readthedocs.io/en/0.4.0/usage/workflows/probeParticles.html) to initialize probe particles does not create probe particles in every 4th cell, **but in every cell**. 

The same for tracer particles following a free formula density profile. 
 
Might  this be related to the `OnePosition` setup (in the example) which does not set weights.
Thus zero weighted particles are not removed.

psychocoderHPC:
fix #2763

The number of skipped cells for the functor is due to a bug always dominated by the dimension with the smallest value. If one dimension is one than each cell is used even if other dimensions has a larger value.

### DIFF
--- a/include/picongpu/particles/densityProfiles/EveryNthCellImpl.hpp
+++ b/include/picongpu/particles/densityProfiles/EveryNthCellImpl.hpp
@@ -67,7 +67,7 @@ namespace densityProfiles
 
             // is this cell populated with a probe particle?
             bool const isPopulated(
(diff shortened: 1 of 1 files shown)


ax3l:

PrometheusPi:
great job fixing this so quickly :+1:

PR #2769, opened 2018-10-25
Params: remove `boost::vector<>` usage

psychocoderHPC:
Remove the usage of `boost::mpl::vector` by `MakeSeq_t`.
This will be much easier for the user and will avoid confusion.

### DIFF
--- a/docs/source/usage/plugins/adios.rst
+++ b/docs/source/usage/plugins/adios.rst
@@ -22,7 +22,7 @@ One can e.g. disable the output of particles by setting:
    /* output all species */
    using FileOutputParticles = VectorAllSpecies;
    /* disable */
-   using FileOutputParticles = bmpl::vector0< >;
+   using FileOutputParticles = MakeSeq_t< >;
 
 .cfg file
 ^^^^^^^^^
--- a/docs/source/usage/plugins/hdf5.rst
+++ b/docs/source/usage/plugins/hdf5.rst
@@ -38,7 +38,7 @@ One can e.g. disable the output of particles by setting:
    /* output all species */
    using FileOutputParticles = VectorAllSpecies;
    /* disable */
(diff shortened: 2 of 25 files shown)


psychocoderHPC:
I updated this PR and remove the usage of `bmpl::vector` also from the documentation

ax3l:
@psychocoderHPC can you please take a look what is causing the compile error?

ax3l:
@psychocoderHPC ping

psychocoderHPC:
I am currently locking into it

psychocoderHPC:
I pushed a new commit where I reintroduce `bmpl::vector` for the init pipeline.
The reason is that `MakeSeq_t` is not working with incomplete types.
To solve this issue multiple refactorings are required.
- we should have a `JoinSeq_t` with has the functionally of the current `MakeSeq_t`
- add `Seq_t` which is nothing else than an alias of `boost::mpl::vector`
- I thing we should refactor `PMacc` and `PIConGPU` fully and provide a clean `hpp` file with the full class implementation and the method implementation in `tpp`. I personally do not like it but since we heavily based on compile time algorithm there is no way around it.

I will create issues out of this post soon.

ax3l:
huh, bad I think for ease of use our `MakeSeq_t` alias should support incomplete types generally.

psychocoderHPC:
> huh, bad I think for ease of use our `MakeSeq_t` alias should support incomplete types generally.

Yes but than it can not anymore support joining of two lists. The reason is as soon as you did complex compile time algorithm you need to have the full type information.

psychocoderHPC:
I opened an issue for the required refactoring #2786 to abstract `boost::mpl`

ax3l:
~~abstract~~ -> "wrap" ;-)

PR #2771, opened 2018-10-25
new tool: pic-submit

psychocoderHPC:
Provide a tool to re submit a job created with tbg.
tbg is abstracting the batch system from the user.
In the case the user is not using `-s` during the tbg usage or needs to be restart a job he/she needs a deep knowledge how to use the batch system.
Due to the reason we have configured all for the user in our shipped profile files we need
only a tool which knows how to restart a job.

The reason why I have not added a new parameter to `tbg` is that for a restart the most parameters of `tbg` are useless because we need not to run the substitute process again. A new parameter where most of the other parameters will be ignored will confuse the user and is not a good practice.

### DIFF
--- a/bin/pic-submit
+++ b/bin/pic-submit
@@ -1,0 +1,114 @@
+#!/usr/bin/env bash
+#
+# Copyright 2013-2018 Axel Huebl, Rene Widera, Richard Pausch
+#
+# This file is part of PIConGPU.
+#
+# PIConGPU is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# PIConGPU is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PIConGPU.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
+# This script parses template files for a batch systems and fills them
+# with a platform-independent configuration, which can then be submitted
+# to a batch system or plainly executed. It can also execute additional
+# commands at that point, e.g. to copy and archive input files alongside
+# a batch job.
+
+function absolute_path()
+{
+    cd $1
+    pwd
+}
+
+help()
+{
+    echo "pic-submit"
+    echo "submit a job created by tbg"
+    echo ""
+    echo "usage: pic-submit [-s submitcommand] [-h] simulationDirectory"
+    echo ""
+    echo "-s | --submit   command        - Submit command (qsub, \"qsub -h\", sbatch, ...)"
+    echo "                                 Default: submitcommand via export TBG_SUBMIT"
+    echo "-h | --help                    - Shows help (this output)."
+    echo ""
+    echo "simulationDirectory            - Directory for simulation output."
+}
+
+# by default the submit command from the environment is used
+# this can be overwritten by the user with the option -s
+submit_command=${submit_command:-$TBG_SUBMIT}
+
+# options may be followed by
+# - one colon to indicate they has a required argument
+# - two colons to indicate they has a optional argument
+OPTS=`getopt -o s:h -l submit:,help -- "$@"`
+if [ $? != 0 ] ; then
+    # something went wrong, getopt will put out an error message for us
+    exit 1
+fi
+
+eval set -- "$OPTS"
+
+# parser
+while true ; do
+    case "$1" in
+        -s|--submit)
+            submit_command=$2
+            shift
+            ;;
+        -h|--help)
+            echo -e "$(help)"
+            shift
+            exit 0
+            ;;
+        --) shift; break;;
+    esac
+    shift
+done
+
+if [ -z "$*" ] ; then
+    echo "No output directory is set (last parameter)." >&2
+    exit 1;
+fi
+
+# only one parameter is allowed
+# an empty parameter list is handled later
+if [ $# -gt 1 ] ; then
+    echo "Too many job directories are given '$*'" >&2
+    exit 1;
+fi
+
+# the last parameter is the `destinationPath`
+outDir="$1"
+
+job_name=`basename "$outDir"`
+job_relative_dir=`dirname "$outDir"`
+
+# TBG_SUBMIT is empty/not and '-s' is not used
+if [ -z "$submit_command" ] ; then
+    echo "No submit command provided, set the environment variable 'TBG_SUBMIT' or use the option `-s`." >&2
+    exit 1;
+fi
+
+job_outDir=`cd "$job_relative_dir"; pwd`"/$job_name"
+submitStart=$job_outDir/tbg/submit.start
+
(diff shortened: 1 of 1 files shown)


ax3l:
Cool, but I would call it `pic-submit` :-)

bin/tbg-submit
```diff
@@ -0,0 +1,114 @@
...
+    cd $1
+    pwd
+}
+
+help()
+{
+    echo "TBG-submit (template batch generator)"
+    echo "submit a job created by tbg"
+    echo ""
+    echo "usage: tbg-submit [-s submitcommand] [-h] jobDirectory"
```
ax3l: I would add `[-t template]` in case one wants to change it

bin/tbg-submit
```diff
@@ -0,0 +1,114 @@
...
+            shift
+            exit 0
+            ;;
+        --) shift; break;;
+    esac
+    shift
+done
+
+if [ -z "$*" ] ; then
+    echo "No output directory is set (last tbg parameter)." >&2
```
ax3l: remove ~~tbg~~ :)

bin/tbg-submit
```diff
@@ -0,0 +1,114 @@
...
+    exit 1;
+fi
+
+# only one parameter is allowed
+# an empty parameter list is handled later
+if [ $# -gt 1 ] ; then
+    echo "To many job directories are given '$*'" >&2
```
ax3l: typo: *too* many

bin/tbg-submit
```diff
@@ -0,0 +1,114 @@
...
+    echo "No submit command provided, set the environment variable 'TBG_SUBMIT' or use the option `-s`." >&2
+    exit 1;
+fi
+
+
+job_outDir=`cd "$job_relative_dir"; pwd`"/$job_name"
+submitStart=$job_outDir/tbg/submit.start
+
+if [ ! -f $submitStart  ] ; then
+    echo "File '$submitStart' not exists. Is '$job_outDir/' created by tbg?" >&2
```
ax3l: not exists -> does not exist. tbg...

-> "File '$submitStart' does not exist. Looks like we could not create '$job_outDir/'?"

(check that the `'` quotes work please)

bin/tbg-submit
psychocoderHPC: Than I need to run tbg again. For that case you should use `tbg` directly.
The tool from the PR is not doing any substitution.

ax3l: That might be a misunderstanding.
I want like you that a user can run
```bash
pic-submit /job/dir
```

You allow to optionally set `-s qsub` and we can also add optionally `-t etc/picongpu/abc/some.tpl`.

bin/tbg-submit
psychocoderHPC: Yes works fine because they are embedded in `""`

sbastrakov:
Wouldn't `pic-submit` be a better alternative for the current `tbg ` as well ? It seems to me, users see `tbg ` as just random 3 letters.

ax3l:
@sbastrakov The idea is to add a new PIConGPU specific command that uses `tbg` under the hood.
That's why we should call it `pic-submit` for consistency with the [basic usage](https://picongpu.readthedocs.io/en/0.4.0/usage/basics.html).

sbastrakov:
@ax3l I got the idea. Just saying, in case a user always does `tbg -s`, it would (to me) make much more sense to call that `pic-submit`, rather than the resubmitting tool (which could then be `pic-resubmit`).

bin/tbg-submit
psychocoderHPC: Yes `-s` is is only the explicit way to set the batch system. This is the pro user option.
`As I wrote `-t` needs to call all substitutions again. If I like to have it I need to run `tbg` again.
This is not a wrapper around `tbg`.
`-t` would also mean I need to update the `tpl` file from the user in the simulation folder. This has to much side effects and is hard to explain.
Do not forget you can use `tbg -s -t YETANOTHER.tpl -c PATH/tbg/submit.cfg -f PATH`

ax3l:
[yes ;-)](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2771#issuecomment-433057629)

psychocoderHPC:
I have currently changed the name to `pic-submit` but also like the suggestion of `pic-submit` and `pic-resubmit` as long as we not change the name `tbg` because tbg is an independent tool those can be used not only in PIConGPU.

bin/pic-submit
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env bash
+#
+# Copyright 2013-2018 Axel Huebl, Rene Widera, Richard Pausch
```
ax3l: @psychocoderHPC just add yourself :)

bin/pic-submit
```diff
@@ -0,0 +1,114 @@
...
+
+help()
+{
+    echo "pic-submit"
+    echo "submit a job created by tbg"
+    echo ""
+    echo "usage: pic-submit [-s submitcommand] [-h] simulationDirectory"
+    echo ""
+    echo "-s | --submit   command        - Submit command (qsub, \"qsub -h\", sbatch, ...)"
+    echo "                                 Default: submitcommand via export TBG_SUBMIT"
```
ax3l: - missing space between `submit` and `command`
- maybe put `"... via \"export TBG_SUBMIT\""` in quotes

psychocoderHPC:
Note: If we like to introduce a tool `pic-submit` (not this one from the PR) than it can not be a simple wrapper around `tbg` the reason is that submitting is one option of `tbg` and must be explicitly queried.

bin/pic-submit
```diff
@@ -0,0 +1,114 @@
...
+
+job_outDir=`cd "$job_relative_dir"; pwd`"/$job_name"
+submitStart=$job_outDir/tbg/submit.start
+
+if [ ! -f $submitStart  ] ; then
+    echo "File '$submitStart' not exists. Is '$job_outDir/' created by tbg?" >&2
+    exit 1
+fi
+
+# some batch systems wrote the job status files to the directory from where the submit command is called
```
ax3l: This comment makes no real sense. What do we want to say here?
Probably:
```
# We now created a submission file for the batch system.
# Let us submit it from the job directory in case the batch system creates status files
# in the submission directory.
```

bin/tbg-submit
ax3l: Oh, too bad. I though we write a tbg wrapper.

bin/pic-submit
psychocoderHPC: @ax3l I am `Rene Widera`^^

bin/pic-submit
psychocoderHPC: `submitcommand` is the submit command from the syntax description above. Maybe I should chnage it to `submit_command`?

bin/pic-submit
ax3l: Yeah, I mean *exclusively* you ;-) but ok if it's copied a lot...

ax3l:
Imho, we should rather add a wrapper, e.g. also called `pic-submit` or `pic-run` that abstracts
```bash
tbg -s [cmd] -t [.tpl] -c <.cfg> /job/dir
```
into
```bash
pic-run [-t <.tpl>] <.cfg> </job/dir>
```

psychocoderHPC:
I opened #2801 to have documented the need of `pic-submit`

Issue #2765, opened 2018-10-24
OpenMPI requires both C and Fortran compilers during default spack install.


berceanu:
Followind the `spack` install instructions on an `Ubuntu 18.04` box, the above message appears while building `openmpi`. It can be fixed by setting
```
   f77: /usr/bin/gfortran-7
   fc: /usr/bin/gfortran-7
```
in `compilers.yaml` or running
```
spack compiler find
```
provided of course, `gfortran` is installed.

ax3l:
Yes, that's currently a limitation of the OpenMPI package in Spack itself. Thank you for documenting this! :sparkles:

berceanu:
I'm not sure how important it is for building `openmpi` that `gcc` and `gfortran` should be the same version. If they should be, then the above fix will often fail, as the system `gfortran` might be a different version than the `spack` `gcc` package. So should we add `gfortran` to the spack dependency list instead?

ax3l:
I thought that the current workflow to `spack install gcc` and then `spack compiler find` does that. Did we miss it? By default, the spack `gcc` package installs the `c`, `c++` and `fortran` languages.

berceanu:
You're absolutely right, `gfortran` is installed along with `gcc`.

ax3l:
Ok, so this was probably just a big mixed up while you debugged the other issues?

berceanu:
Yes, looks like everything is working now :)

ax3l:
Thanks for the note!
I now updated the recipe to explicitly build on OpenMPI between 1.8 and 3.99: https://github.com/ComputationalRadiationPhysics/spack-repo/commit/69327543e1892011dbbb8485a213a7c1433fa9ac

ax3l:
@hightower8083 can you please report the GCC 5.5 build issue that you get when running `spack install openmpi@4.0.0 %gcc@5.5` upstream in https://github.com/open-mpi/ompi ? I think it's not reported yet.

hightower8083:
sure -- i'll check it again and will tell em

PR #2773, opened 2018-10-25
Spack: Improve Bootstrap

### LINKED ISSUE #2770: module command not found on spack install
Following the [spack install guide](https://picongpu.readthedocs.io/en/0.4.0/install/instructions/spack.html), the command 
```
spack load gcc@7.3.0
```
fails with the above error. That means that `spack compiler add` is never executed.

Even though `spack bootstrap` installs `environment-modules`, they are not added to the runtime path.
The fix is to run
```
source $HOME/src/spack/share/spack/setup-env.sh
```
once again after `spack bootstrap`.

ax3l:
If the bootstrap pulls in environment modules, these need to be activated. Improve command order to avoid sourcing twice.

Fix #2770 

- [x] WIP: do not merge until checkbox is checked

### DIFF
--- a/docs/source/install/instructions/spack.rst
+++ b/docs/source/install/instructions/spack.rst
@@ -24,14 +24,14 @@ First `install spack <http://spack.readthedocs.io/en/latest/getting_started.html
    # get spack
    git clone https://github.com/spack/spack.git $HOME/src/spack
 
+   # build spack's dependencies via spack :)
+   $HOME/src/spack/bin/spack bootstrap
+
    # activate the spack environment
    source $HOME/src/spack/share/spack/setup-env.sh
 
-   # build spack's dependencies via spack :)
-   spack bootstrap
-
    # install a supported compiler
-   spack compiler list | grep gcc@7.3.0 | spack install gcc@7.3.0 && spack load gcc@7.3.0 && spack compiler add
+   spack compiler list | grep gcc@7.3.0 1>/dev/null && spack install gcc@7.3.0 && spack load gcc@7.3.0 && spack compiler add
 
    # add the PIConGPU repository
    git clone https://github.com/ComputationalRadiationPhysics/spack-repo.git $HOME/src/spack-repo


docs/source/install/instructions/spack.rst
```diff
@@ -24,12 +24,12 @@ First `install spack <http://spack.readthedocs.io/en/latest/getting_started.html
    # get spack
    git clone https://github.com/spack/spack.git $HOME/src/spack
 
+   # build spack's dependencies via spack :)
+   ./spack/bin/spack bootstrap
```
berceanu: Here I would do:
```
$HOME/src/spack/bin/spack bootstrap
```
because the user might not be in the `~/src` folder.

ax3l:
@berceanu there is also another glitch in the workflow that we just realized. In case you already have a compiler that works on your system, we skip bootstrapping it again due to https://github.com/spack/spack/issues/6902#issuecomment-357025592

In that case, this gcc might not come with a gfortran already from the system... Additionally, even after installing it manually, we will hit https://github.com/spack/spack/issues/5101#issuecomment-357232008

So maybe bootstrapping an (uncommon) version by default is best? ;-)

berceanu:
Yes, that is how one can still run into https://github.com/ComputationalRadiationPhysics/picongpu/issues/2765. Which version are you thinking of bootstrapping?

berceanu:
Perhaps related to this, while running PIConGPU after installing it using the system's gcc 7.3.0, I get:
```
Running program...
==> Error: the constraint '['gcc@7.3.0']' matches no package.
==> Error: In this context exactly **one** match is needed: please specify your constraints better.
```
which is probably caused by a `spack load gcc@7.3.0` somewhere - which afaik only works if you've built gcc via spack?

ax3l:
What does `spack compiler list` output? Probably it identifies as `@7.3` instead of `@7.3.0`?

psychocoderHPC:
Is this now merge able or work in progress?

berceanu:
Well, isn't it WIP until we solve the gcc/gfortran corner case?

Also, another (related) thing.

If I understand correctly, the line
```
spack compiler list | grep gcc@7.3.0 | spack install gcc@7.3.0 && spack load gcc@7.3.0 && spack compiler add
```
is only supposed to install `gcc@7.3.0` via `spack` if it does not already exist in the system?

However, if I try:
```
echo "nothing" | grep "gcc" | echo "hello"
```
it seems that `hello` is printed even though the grep fails.

berceanu:
> What does `spack compiler list` output? Probably it identifies as `@7.3` instead of `@7.3.0`?

No, the version number is correct. I asked on the `spack` mailing list and they confirmed that `spack` doesn't generate modules for the detected system compilers, so one can't use `spack load` in that case.

docs/source/install/instructions/spack.rst
```diff
@@ -24,12 +24,12 @@ First `install spack <http://spack.readthedocs.io/en/latest/getting_started.html
...
+   $HOME/src/spack/bin/spack bootstrap
+
    # activate the spack environment
    source $HOME/src/spack/share/spack/setup-env.sh
 
-   # build spack's dependencies via spack :)
-   spack bootstrap
-
    # install a supported compiler
    spack compiler list | grep gcc@7.3.0 | spack install gcc@7.3.0 && spack load gcc@7.3.0 && spack compiler add
```
ax3l: should be
```suggestion
   spack compiler list | grep gcc@7.3.0 1>/dev/null && spack install gcc@7.3.0 && spack load gcc@7.3.0 && spack compiler add
```

psychocoderHPC:
I added a checkbox, this avoids that I merge it before you thing all is fine.

docs/source/install/instructions/spack.rst
berceanu: Looks good and seems to fix the above `echo` counter-example :)

ax3l:
I updated the grep logic and keep the unsolved corner case for later since I don't see a simple solution right now besides scripting the hack out of it.

ax3l:

PR #2774, opened 2018-10-26
avoid two slides

psychocoderHPC:
If someone is querying the moving window class with a time step which is less than
the current simulation step it can happen that the window is sliding.

- only allow sliding if the time step is larger than the last slide point

### DIFF
--- a/include/picongpu/simulationControl/MovingWindow.hpp
+++ b/include/picongpu/simulationControl/MovingWindow.hpp
@@ -180,7 +180,8 @@ private:
      */
     void incrementSlideCounter(const uint32_t currentStep)
     {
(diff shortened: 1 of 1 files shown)


include/picongpu/simulationControl/MovingWindow.hpp
```diff
@@ -180,7 +180,8 @@ class MovingWindow
      */
     void incrementSlideCounter(const uint32_t currentStep)
     {
-        if (slidingWindowActive==true && lastSlideStep != currentStep)
+        // do not slice twice in one simulation step
```
PrometheusPi: typo:
slice -> slide

Issue #2759, opened 2018-10-24
radiation python module - no getter for omega and n


PrometheusPi:
The python module to read radiation hdf5 data does not support an interface for getting the frequency and n-vector array.

ax3l:
Also, it's not following the base classes and common interface for the data readers and `plot_mpl` visualizers :)

PrometheusPi:
I opened a new issue for the interface update suggested.

PR #2776, opened 2018-10-26
Add getter for omega and n to python module

PrometheusPi:
This pull request closes issue #2759 by adding the missing getters for frequency (omega) and observation direction (n) . It still does not make it compatible to the standard iteration oriented interface.

### DIFF
--- a/lib/python/picongpu/plugins/data/radiation.py
+++ b/lib/python/picongpu/plugins/data/radiation.py
@@ -105,3 +105,16 @@ class RadiationData:
         omega_h = self.h5_file[("/data/{}/DetectorMesh/" +
                                 "DetectorFrequency/omega").format(self.timestep)]
         return omega_h[0, :, 0] * omega_h.attrs['unitSI']
+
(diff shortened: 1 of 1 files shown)


codingS3b:
I really don't know a lot about this plugin, but could some of the getter/setter functionality be implemented into a class/container that gives meta information, similar to what is done in [`plugins/data/phase_space`](https://github.com/ComputationalRadiationPhysics/picongpu/blob/a64c8d8bcdf2ba1299936410486167aa935675c3/lib/python/picongpu/plugins/data/phase_space.py#L17)?

PrometheusPi:
@codingS3b Very good point - @ax3l already mentioned in issue #2759 that this module is not compatible with the others and thus should follow the same interface. 

I totally agree on the suggestion of both of you. However, I would prefer adding the functionality for now and the interface adjustment (including a consistent documentation) in a followup pull request.

If you however prefer adding all in one pull request - then just let me know.

PR #2779, opened 2018-10-30
Python Tools: Incorporate sim_time into readers and visualizers

codingS3b:
This PR makes use of the `FindTime` class and adds it to the data readers. This offers the possibility to query for simulation times directly instead of only query for iterations.

Besides, the readers now return an array of data items when multiple iterations are given in the `get` methods instead of dictionaries. This was done to allow reusing the same function for getting data regardless of whether `time` or `iteration` parameter were provided.

### DIFF
--- a/lib/python/picongpu/plugins/data/base_reader.py
+++ b/lib/python/picongpu/plugins/data/base_reader.py
@@ -75,9 +75,11 @@ class DataReader(object):
         an iteration and data for the iteration matching the time
         is returned.
 
-        time: float or np.array of float
+        time: float or np.array of float or None.
+            If None, data for all available times is returned.
 
-        iteration: int or np.array of int
+        iteration: int or np.array of int or None.
+            If None, data for all available iterations is returned.
 
         Returns
         -------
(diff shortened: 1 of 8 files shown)


lib/python/picongpu/plugins/data/base_reader.py
```diff
@@ -23,10 +26,29 @@ def __init__(self, run_directory):
             raise ValueError('The run_directory parameter can not be None!')
 
         self.run_directory = run_directory
+        self.find_time = FindTime(run_directory)
+
         # need to be set in derived classes
         self.data_file_prefix = None
         self.data_file_suffix = None
 
+    def get_dt(self):
```
ax3l: Do we need this as public API for all data readers?
We can do this for convenience, but it might be a bit doubled.

lib/python/picongpu/plugins/data/base_reader.py
```diff
@@ -46,6 +68,52 @@ def get_iterations(self, **kwargs):
 
     def get(self, **kwargs):
         """
+        Parameters
+        ----------
+        Either 'iteration' or 'time' should be present in the kwargs.
+        If both are given, the 'time' argument is converted to
+        an iteration and data for the iteration matching the time
```
ax3l: document how the matching is done.
Internally, you implement the `find_time.get_iteration` default arguments which are:
```
previous : the closest iteration that is <= t (default)
```

PR #2788, opened 2018-11-06
remove grep file redicrect

psychocoderHPC:
Use option `-q|--quiet` instead of redirection the output of `grep` to `/dev/null`

### DIFF
--- a/docs/source/install/instructions/spack.rst
+++ b/docs/source/install/instructions/spack.rst
@@ -31,7 +31,7 @@ First `install spack <http://spack.readthedocs.io/en/latest/getting_started.html
    source $HOME/src/spack/share/spack/setup-env.sh
 
    # install a supported compiler
-   spack compiler list | grep gcc@7.3.0 1>/dev/null && spack install gcc@7.3.0 && spack load gcc@7.3.0 && spack compiler add
+   spack compiler list | grep -q gcc@7.3.0 && spack install gcc@7.3.0 && spack load gcc@7.3.0 && spack compiler add
 
    # add the PIConGPU repository
    git clone https://github.com/ComputationalRadiationPhysics/spack-repo.git $HOME/src/spack-repo


docs/source/install/instructions/spack.rst
```diff
@@ -31,7 +31,7 @@ First `install spack <http://spack.readthedocs.io/en/latest/getting_started.html
    source $HOME/src/spack/share/spack/setup-env.sh
 
    # install a supported compiler
-   spack compiler list | grep gcc@7.3.0 1>/dev/null && spack install gcc@7.3.0 && spack load gcc@7.3.0 && spack compiler add
+   spack compiler list | grep -q gcc@7.3.0 && spack install gcc@7.3.0 && spack load gcc@7.3.0 && spack compiler add
```
ax3l: @PrometheusPi does `grep -q` exist & work on OSX?

berceanu:
Just tested, works on Ubuntu & CentOS.

PrometheusPi:
@ax3l if `-q` is the `--quiet` or `--silent` mode, it exists on Mac OS.

berceanu:
> @ax3l if `-q` is the `--quiet` or `--silent` mode, it exists on Mac OS.

```
       -q, --quiet, --silent
              Quiet;  do  not  write  anything  to  standard   output.    Exit
              immediately  with  zero status if any match is found, even if an
              error was detected.  Also see the -s or --no-messages option.
```

berceanu:
My two cents on this is that if we cover the `POSIX`-compliant case, we're fine.

ax3l:
Should ideally work on OSX and Linux and there are subtle FreeBSD differences.

The exit status matters here, e.g. `--silent` would not do:
```
--silent  is  used  and  a  line is selected, the exit status is 0 even if an
       error occurred.
```

ax3l:
@PrometheusPi what does
```bash
echo "abc" | grep -q "b"
echo $?

echo "abc" | grep -q "a"
echo $?
```

output?

PrometheusPi:
@ax3l 
```bash
$ echo "abc" | grep -q "b"
$ echo $?
0

$ echo "abc" | grep -q "a"
$ echo $?
0
```

ax3l:
ahhh, yes, I will clean up my mess above. works, so we are good to go.

PR #2792, opened 2018-11-08
stop moving window after N steps

psychocoderHPC:
Sometimes for bunch-bunch or bunch-target collision it could be useful to allow to stop the
moving windows. This option is now available via `stopWindow`.

- add cli option `stopWindow`
- refactor `MovingWindow`: add method `setEndSlideOnStep` and `isEnabled`

# Tests

- [x] LWFA with `-m`
- [x] LWFA `-m --stopWindow 2000`
- [x] LWFA no sliding window

### DIFF
--- a/docs/TBG_macros.cfg
+++ b/docs/TBG_macros.cfg
@@ -90,6 +90,10 @@ TBG_periodic="--periodic 1 0 1"
 # Enables moving window (sliding) in your simulation
 TBG_movingWindow="-m"
 
+
+# stop the moving window after given simulation step
+TBG_stopWindow="--stopWindow 1337"
+
 ################################################################################
 ## Placeholder for multi data plugins:
 ##
--- a/include/picongpu/fields/FieldManipulator.hpp
+++ b/include/picongpu/fields/FieldManipulator.hpp
@@ -83,7 +83,7 @@ public:
                 }
 
                 /* if sliding window is active we disable absorber on bottom side*/
-                if (MovingWindow::getInstance().isSlidingWindowActive() && i == BOTTOM) continue;
+                if (MovingWindow::getInstance().isSlidingWindowActive(currentStep) && i == BOTTOM) continue;
 
                 ExchangeMapping<GUARD, MappingDesc> mapper(cellDescription, i);
                 constexpr uint32_t numWorkers = pmacc::traits::GetNumWorkers<
--- a/include/picongpu/simulationControl/MovingWindow.hpp
+++ b/include/picongpu/simulationControl/MovingWindow.hpp
@@ -36,9 +36,7 @@ class MovingWindow
 {
 private:
 
-    MovingWindow() : slidingWindowActive(false), slideCounter(0), lastSlideStep(0)
-    {
-    }
+    MovingWindow() = default;
 
     MovingWindow(MovingWindow& cc);
 
@@ -50,8 +48,15 @@ private:
         if (offsetFirstGPU)
             *offsetFirstGPU = 0.0;
 
-        if (slidingWindowActive)
+        if (slidingWindowEnabled)
         {
+            /* Sliding stayed enabled but if we reach the end step where we should stop sliding
+             * the moving window is freezed.
+             * All offsets will stay constant until the end of the simulation.
+             */
+            if (currentStep >= endSlidingOnStep)
+                currentStep = endSlidingOnStep;
+
             const SubGrid<simDim>& subGrid = Environment<simDim>::get().SubGrid();
 
             /* speed of the moving window */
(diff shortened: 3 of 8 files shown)


PrometheusPi:
@psychocoderHPC why is the `-w` flag needed when stopping while the `-m` is needed when not stopping- Shouldn't it be the same flag to activate a moving window?

include/picongpu/simulationControl/MovingWindow.hpp
```diff
@@ -181,34 +186,55 @@ class MovingWindow
     void incrementSlideCounter(const uint32_t currentStep)
     {
         // do not slide twice in one simulation step
-        if (slidingWindowActive==true && lastSlideStep < currentStep)
+        if (currentStep < endSlidingOnStep && lastSlideStep != currentStep)
```
ax3l: you just reverted #2774 with "!="

ax3l:
@psychocoderHPC in case you are at work, I reviewed this PR :)

include/picongpu/simulationControl/MovingWindow.hpp
psychocoderHPC: well spotted! thx

psychocoderHPC:
> docs/TBG_macros.cfg needs an update as well

@ax3l  the doc update is included
https://github.com/ComputationalRadiationPhysics/picongpu/pull/2792/files#diff-6ed48ed85428745c2f464a9f3643f879R94

psychocoderHPC:
> @psychocoderHPC why is the -w flag needed when stopping while the -m is needed when not stopping- Shouldn't it be the same flag to activate a moving window?

@PrometheusPi  I fixed the initial description. `-w` is not needed it must be a `-m` I pressed the wrong key on my keyboard (they are so near by each other ;-)

include/picongpu/simulationControl/MovingWindow.hpp
psychocoderHPC: fixed

include/picongpu/fields/FieldManipulator.hpp
```diff
@@ -83,7 +83,7 @@ class FieldManipulator
                 }
 
                 /* if sliding window is active we disable absorber on bottom side*/
-                if (MovingWindow::getInstance().isSlidingWindowActive() && i == BOTTOM) continue;
+                if (MovingWindow::getInstance().isSlidingWindowActive(currentStep) && i == BOTTOM) continue;
```
ax3l: @psychocoderHPC why did you change this as well?

ax3l: forget about it, I saw it below... makes sense

include/picongpu/simulationControl/MovingWindow.hpp
```diff
@@ -181,34 +186,55 @@ class MovingWindow
     void incrementSlideCounter(const uint32_t currentStep)
     {
         // do not slide twice in one simulation step
-        if (slidingWindowActive==true && lastSlideStep < currentStep)
+        if (currentStep < endSlidingOnStep && lastSlideStep < currentStep)
```
ax3l: since `endSlidingOnStep` defaults to `0u` and `currentStep` is `>=0` this is fine. But by the logic that was here before, you should first check if `isEnabled()` is true.

include/picongpu/fields/FieldManipulator.hpp
psychocoderHPC: Because without the time step it is not possible to check if the moving window is stopped.
As soon as the window stops the bottom absorber is enabled else the laser get reflected

include/picongpu/simulationControl/MySimulation.hpp
```diff
@@ -117,6 +117,7 @@ class MySimulation : public SimulationHelper<simDim>
     cellDescription(nullptr),
     initialiserController(nullptr),
     slidingWindow(false),
+    endSlidingOnStep(0),
```
ax3l: this should be `-1` for consistency with the default value. Or you make it `uint32_t` as the usage in `MovingWindow` and put `0` as a default, which I find reasonable a well as a special value (do *not* stop at zero).

include/picongpu/simulationControl/MovingWindow.hpp
psychocoderHPC: Sounds good than it is more cleare

include/picongpu/fields/FieldManipulator.hpp
ax3l: Yes, overlooked this

include/picongpu/simulationControl/MySimulation.hpp
psychocoderHPC: To enable the moving window we have a bool switch. Therefore I set the default to `disbale/0`
`uint32_t` is not useful because than I need to show the user a magic large number as default in the help (https://github.com/ComputationalRadiationPhysics/picongpu/pull/2792/files). It is more common to have `-1` as unlimited.

ax3l: But in `MovingWindow` you set it to `0`. I would be fine to make it all-`0` as default as you got it there.

psychocoderHPC: The object `MovingWindow` is Independent of `MySimulation` there is no need that internal variables of two different objects fits in his default value.
IMO `-1` fits much better to `--stopWindow step` instead `zero`

ax3l: Sure, but following that logic, shouldn't `MovingWindow` also use `-1` for "do not move"? ;-)

ax3l: Also, this has a bug. If a user *would* truly request `--stopWindow 0` then it would still not stop at zero. Therefore, make the cmd line interface with `0 means not stopping` as well for simplicity?

ax3l: update, no wait, it works as intended.

docs/TBG_macros.cfg
```diff
@@ -90,6 +90,8 @@ TBG_periodic="--periodic 1 0 1"
 # Enables moving window (sliding) in your simulation
 TBG_movingWindow="-m"
 
+# stop the moving window after given simulation step
+TBG_stopWindow="--stopWindow 1337"
 ################################################################################
```
ax3l: pls give me an empty line :)

psychocoderHPC: You and your white space fetish^^ ( I will add the empty line)

ax3l: The - symmetry - is - broken! xD

psychocoderHPC: I forgot to push my changes

PR #2793, opened 2018-11-09
Fix species.param example doc and grammar

steindev:
Fix 1 grammar error and update(?)  example text for particle shape

### DIFF
--- a/include/picongpu/param/species.param
+++ b/include/picongpu/param/species.param
@@ -43,7 +43,7 @@ namespace picongpu
  *  - particles::shapes::PCS : 3rd order
  *  - particles::shapes::P4S : 4th order
  *
- *  example:             using CICShape = particles::shapes::CIC;
+ *  example for using CIC shape: `using UsedParticleShape = particles::shapes::CIC;`
  */
 using UsedParticleShape = particles::shapes::TSC;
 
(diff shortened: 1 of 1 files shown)


ax3l:
as always, please update all examples that have the same input file in `share/picongpu/examples/` as well... Please do as a follow-up PR.

include/picongpu/param/species.param
```diff
@@ -43,7 +43,7 @@ namespace picongpu
  *  - particles::shapes::PCS : 3rd order
  *  - particles::shapes::P4S : 4th order
  *
- *  example:             using CICShape = particles::shapes::CIC;
+ *  example for using CIC shape: `using UsedParticleShape = particles::shapes::CIC;`
```
ax3l: I also think this is not necessary. We write literally prosa here, now you say:

"example using CIC shape using CIC shape"... I would have stayed with `example:     using UsedParticleShape = particles::shapes::CIC;`

psychocoderHPC:
@steindev You missed to fix the grammer `Define` -> `Defining` in all examples.

steindev:
> @steindev You missed to fix the grammar `Define` -> `Defining` in all examples.

I fixed this in the follow-up PR #2795 as well.

PR #2803, opened 2018-11-13
Plugin Notify: Debug

ax3l:
Just documenting a quick way to debug whenever the `PluginConnector` `load()` of an `IPlugin` and  `notify()` of an `INotify` object.

Most multi-plugins do, unfurtunately, not cast to `IPlugin` and can therefore not be identified by name.

### DIFF
--- a/include/pmacc/pluginSystem/PluginConnector.hpp
+++ b/include/pmacc/pluginSystem/PluginConnector.hpp
@@ -31,6 +31,7 @@
 #include <vector>
 #include <list>
 #include <string>
+#include <iostream>
 
 
 namespace pmacc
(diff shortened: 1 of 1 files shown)


include/pmacc/pluginSystem/PluginConnector.hpp
```diff
@@ -252,6 +266,7 @@ namespace pmacc
         }
 
         std::list<IPlugin*> plugins;
+        std::list<IPlugin*> notificationPlugins;
```
ax3l: not used in the end

psychocrypt:
We can implement your changes if we use the log system to control the amount of messages. Maybe add a new log level for this kind of output.

psychocoderHPC:
We can use this PR if we use the  log system to control the amount of messages. Maybe create anew log lvl for this kind of messages.

Issue #2794, opened 2018-11-11
Send/receive buffer for species


berceanu:
I am testing an LWFA run with Nitrogen 5+ and get this output:
```
Running program...
PIConGPU: 0.4.0
  Build-Type: Release

Third party:
  OS:         Linux-4.15.0-38-generic
  arch:       x86_64
  CXX:        GNU (7.3.0)
  CMake:      3.12.3
  CUDA:       9.2.88
  mallocMC:   2.3.0
  Boost:      1.65.1
  MPI:
    standard: 3.1
    flavor:   OpenMPI (3.1.2)
  PNGwriter:  0.7.0
  libSplash:  1.7.0 (Format 4.0)
  ADIOS:      1.13.1
PIConGPUVerbose PHYSICS(1) | Sliding Window is ON
PIConGPUVerbose PHYSICS(1) | used Random Number Generator: RNGProvider3XorMin seed: 42
PIConGPUVerbose PHYSICS(1) | Courant c*dt <= 1.00229 ? 1
PIConGPUVerbose PHYSICS(1) | species e: omega_p * dt <= 0.1 ? 0.0209094
PIConGPUVerbose PHYSICS(1) | species e_nitro: omega_p * dt <= 0.1 ? 0.0209094
PIConGPUVerbose PHYSICS(1) | y-cells per wavelength: 18.0587
PIConGPUVerbose PHYSICS(1) | macro particles per device: 268435456
PIConGPUVerbose PHYSICS(1) | typical macro particle weighting: 4945.05
PIConGPUVerbose PHYSICS(1) | UNIT_SPEED 2.99792e+08
PIConGPUVerbose PHYSICS(1) | UNIT_TIME 1.39e-16
PIConGPUVerbose PHYSICS(1) | UNIT_LENGTH 4.16712e-08
PIConGPUVerbose PHYSICS(1) | UNIT_MASS 4.50464e-27
PIConGPUVerbose PHYSICS(1) | UNIT_CHARGE 7.92284e-16
PIConGPUVerbose PHYSICS(1) | UNIT_EFIELD 1.22627e+13
PIConGPUVerbose PHYSICS(1) | UNIT_BFIELD 40903.8
PIConGPUVerbose PHYSICS(1) | UNIT_ENERGY 4.04856e-10
initialization time:  5min 20sec 214msec = 320 sec
  0 % =        0 | time elapsed:       2min  1sec 995msec | avg time per step:   0msec
[...]

PR #2806, opened 2018-11-14
Add PIConGPU memory calculator

n01r:
This is the first tentative version of a memory requirement calculator for PIConGPU.

It includes functions for calculating the memory used by fields, particles, random generator and
chosen plugins (only the calorimeter at the moment) per accelerator.

It worked for me like that in the planning of simulations but I would like a more comfortable structure of this.
I am sure you, @ax3l and @codingS3b, have an opinion on how something like that would look like and thus I put my rough version up for dicussion here. :smiley: 

There is currently a very specific example in the code but I can remove or replace this. The output of this example looks like that:

```
$ python picongpu_memory_calculator.py

Memory requirement per GPU for fields: 581.92 MB
Memory requirement per GPU and species:
e: 9117.77 MB
H: 701.37 MB
C: 759.81 MB
O: 759.81 MB
Memory requirement per GPU for RNG states: 474.26 MB
Memory requirement per GPU for particle calorimeters: 253.12 MB
Sum of required GPU memory: 12648.07 MB
```

Documentation on usage follows after we agree on a final version for this pull.


berceanu:
It would be awesome to have this as a small debug plugin, or perhaps part of the PIConGPU output at runtime?

n01r:
> part of the PIConGPU output at runtime

Well, yes, we had some offline talk about it but to really do this nicely we would need some deeper hacks within the memory management.

There is a PIConGPU log level that already shows how much memory is reserved for the particles that are initialized but the tool that I uploaded here is also for planning of really large-scale simulations where you need to take into account how many particles are still created during runtime due to e.g. ionization processes.

lib/python/picongpu/utils/picongpu_memory_calculator.py
```diff
@@ -0,0 +1,363 @@
+#!/usr/bin/env python
+
+"""
+This file is part of PIConGPU.
+
+It is supposed to give an estimate for the memory requirement of a PIConGPU
+simulation per accelerator.
+
+Copyright 2018 PIConGPU contributors
```
PrometheusPi: @ax3l is this short license header fine with you
some other python modules use the same short version

lib/python/picongpu/utils/picongpu_memory_calculator.py
```diff
@@ -0,0 +1,363 @@
...
+
+        req_mem : int
+            required memory {unit: bytes} per accelerator
+        """
+        if n_x is None:
+            n_x = self.n_x
+        if n_y is None:
+            n_y = self.n_y
+        if n_z is None:
+            n_z = self.n_z
```
PrometheusPi: @codingS3b Is there a better solution to this (using self values as default input arguments)?

lib/python/picongpu/utils/picongpu_memory_calculator.py
```diff
@@ -0,0 +1,363 @@
...
+            target_n_x = self.n_x
+        if target_n_y is None:
+            target_n_y = self.n_y
+        if target_n_z is None:
+            target_n_z = self.n_z
+
+        # memory required by the standard particle attributes
+        standard_attribute_mem = np.array([
+            3 * 4,  # momentum
+            3 * 4,  # position
```
PrometheusPi: is this not depending on whether the simulation is in 2D or 3D?

lib/python/picongpu/utils/picongpu_memory_calculator.py
```diff
@@ -0,0 +1,363 @@
...
+            1 * 8,  # multimask
+            1 * 8,  # cell index in supercell (``lcellId_t``)
+            1 * 8  # weighting
```
PrometheusPi: what about extensions like `prevMomentum1` or particle ids?

lib/python/picongpu/utils/picongpu_memory_calculator.py
```diff
@@ -0,0 +1,363 @@
...
+
+        req_mem : int
+            required memory {unit: bytes} per accelerator
+        """
+        if value_size is None:
+            value_size = self.value_size
+
+        req_mem_per_bin = value_size
+        num_bins = n_energy * n_yaw * n_pitch
+        req_mem = req_mem_per_bin * num_bins * 2
```
PrometheusPi: please add a comment for this `2`- you wrote the description in the doc-string above, but please repeat it here, so that this is directly clear

lib/python/picongpu/utils/picongpu_memory_calculator.py
sbastrakov: It's not obvious to me what is the point of having these `n_x, n_y, n_z` arguments here at all? They are all mandatory for the constructor. Or, if they are permitted to be `None` in the constructor, `n_x`, etc. could be still `None` here after these checks (in case the original value is also `None`).

lib/python/picongpu/utils/picongpu_memory_calculator.py
n01r: this is not a standard attribute - for this you have the option to add the number of additional attributes

lib/python/picongpu/utils/picongpu_memory_calculator.py
n01r: for position: yes - for momentum: no

will fix it

ax3l:
ping @x113f you will like that :)

lib/python/picongpu/utils/picongpu_memory_calculator.py
n01r: the idea was to have them be optional if you constructed the class and used the constructor but if you wanted to just know how much memory your fields use up you can just use the function

anyhow - not too much work went into the structure of this thing so suggestions are welcome =)

ax3l:
@n01r @berceanu also, there are the following plugins: https://github.com/ComputationalRadiationPhysics/picongpu/issues/2794#issuecomment-437892501

Yes, ideally we could integrate them deeper! Also, maybe pre-simulating with heavily reduced resolution (e.g. 0.1x per `dx`) could help for estimates before running high-res sims, as ppl around @x113f just suggested.

lib/python/picongpu/utils/picongpu_memory_calculator.py
PrometheusPi: that's great

lib/python/picongpu/utils/picongpu_memory_calculator.py
codingS3b: After checking [stackoverflow](https://stackoverflow.com/questions/15189245/assigning-class-variable-as-default-value-to-class-method-argument) it seems that there is no better solution than the current implementation.

lib/python/picongpu/utils/picongpu_memory_calculator.py
ax3l: that's good and very clear for Python.

lib/python/picongpu/utils/picongpu_memory_calculator.py
```diff
@@ -0,0 +1,363 @@
...
+
+Copyright 2018 PIConGPU contributors
+Authors: Marco Garten
+License: GPLv3+
+"""
+
+import numpy as np
+
+
+class PicongpuMemoryCalculator:
```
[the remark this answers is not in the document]
n01r: Ah, sure, makes sense - will do!

lib/python/picongpu/utils/picongpu_memory_calculator.py
codingS3b: Maybe you could also move the function outside of the class and within the class only create a wrapper where the member attributes are forwarded to the external function?

n01r:
Still to do:
- [ ] Replace the arbitrary example by one that we have in our list of examples, e.g. `FoilLCT`.
- [ ] Possibly implement the wrapper solution suggested by @codingS3b
- [ ] Add a documentation section (working on it - hard to decide where to put it and with what to link it)

n01r:
@psychocoderHPC could you check that the memory requirement for the random number generator state is still valid / suggest what to add for the other two options that we now have for RNGs?

n01r:
I seem to be unable to build the standard docs.
I updated them according to our [readthedocs section](https://picongpu.readthedocs.io/en/latest/dev/sphinx.html?highlight=sphinx#build-locally). 

```python
# Sphinx version: 1.7.0
# Python version: 3.6.5 (CPython)
# Docutils version: 0.14 
# Jinja2 version: 2.10
# Last messages:
#   reading sources... [  1%] dev/CONTRIBUTING
#   
#   reading sources... [  2%] dev/clangtools
#   
#   reading sources... [  3%] dev/doxygen
#   
#   reading sources... [  4%] dev/doxyindex
#   
#   reading sources... [  6%] dev/picongpu
#   
# Loaded extensions:
#   alabaster (0.7.11) from /home/noir/miniconda3/lib/python3.6/site-packages/alabaster/__init__.py
#   sphinx.ext.mathjax (1.7.0) from /home/noir/miniconda3/lib/python3.6/site-packages/sphinx/ext/mathjax.py
#   sphinx.ext.autodoc (1.7.0) from /home/noir/miniconda3/lib/python3.6/site-packages/sphinx/ext/autodoc/__init__.py
#   sphinx.ext.napoleon (1.7.0) from /home/noir/miniconda3/lib/python3.6/site-packages/sphinx/ext/napoleon/__init__.py
#   breathe (4.11.0) from /home/noir/miniconda3/lib/python3.6/site-packages/breathe/__init__.py
#   sphinxcontrib.programoutput (unknown version) from /home/noir/miniconda3/lib/python3.6/site-packages/sphinxcontrib/programoutput/__init__.py
#   matplotlib.sphinxext.plot_directive (unknown version) from /home/noir/miniconda3/lib/python3.6/site-packages/matplotlib/sphinxext/plot_directive.py
#   sphinx.ext.githubpages (1.7.0) from /home/noir/miniconda3/lib/python3.6/site-packages/sphinx/ext/githubpages.py
  File "/home/noir/miniconda3/lib/python3.6/site-packages/breathe/parser/__init__.py", line 84, in parse
    return self.cache[filename]
KeyError: '/home/noir/Arbeit/Coding/picongpu/docs/source/../xml/indexpage.xml'

During handling of the above exception, another exception occurred:

  File "/home/noir/miniconda3/lib/python3.6/site-packages/breathe/parser/compound.py", line 1067, in parse
    doc = minidom.parse(inFilename)
  File "/home/noir/miniconda3/lib/python3.6/xml/dom/minidom.py", line 1958, in parse
    return expatbuilder.parse(file)
  File "/home/noir/miniconda3/lib/python3.6/xml/dom/expatbuilder.py", line 911, in parse
    result = builder.parseFile(fp)
  File "/home/noir/miniconda3/lib/python3.6/xml/dom/expatbuilder.py", line 207, in parseFile
    parser.Parse(buffer, 0)
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 14, column 835

During handling of the above exception, another exception occurred:
```

lib/python/picongpu/utils/picongpu_memory_calculator.py
ax3l: @n01r did you see that comment? Would that help?

docs/source/usage/plugins.rst
```diff
@@ -61,6 +61,16 @@ Examples
 * ``42,30:50:10``: at steps 30 40 42 50 84 126 168 ...
 * ``5,10``: at steps 0 5 10 15 20 25 ... (only executed once per step in overlapping intervals)
 
+Python Preprocessing
```
ax3l: ah, sorry but this does not belong to plugins :)
Maybe add this intro to `usage-python-utils` where you have the docs for the python utils already?
Maybe the section of workflows is sufficient as well?

docs/source/usage/python/utils.rst
```diff
@@ -0,0 +1,13 @@
+.. _usage-python-utils:
+
+Utils
+--------
+
+memory_calculator.py
+^^^^^^^^^^^^^^^^^^^
+
+.. doxygenfile:: memory_calculator.py
```
ax3l: doxygenfile? You can use regular sphinx docs here, see e.g. https://picongpu.readthedocs.io/en/0.4.2/dev/py_postprocessing.html

docs/source/usage/python/utils.rst
```diff
@@ -0,0 +1,13 @@
+.. _usage-python-utils:
+
+Utils
+--------
```
ax3l: Please align with title in `.rst`.
Utils is short for Utilities.

docs/source/usage/workflows/memoryPerAccelerator.rst
```diff
@@ -0,0 +1,115 @@
+.. _usage-workflows-memoryPerAccelerator:
+
+Calculating the memory requirement per accelerator
```
ax3l: Pls start with uppercase letters for nouns in English titles.

docs/source/usage/workflows/memoryPerAccelerator.rst
```diff
@@ -0,0 +1,115 @@
+.. _usage-workflows-memoryPerAccelerator:
```
ax3l: we (try to) call "accelerators" consequently "device" in PIConGPU 0.4+

lib/python/picongpu/utils/memory_calculator.py
```diff
@@ -0,0 +1,368 @@
...
+import numpy as np
+
+
+class MemoryCalculator:
+    """
+    Memory requirement calculation tool for PIConGPU
+
+    Contains calculation for fields, particles, random number generator
+    and the calorimeter plugin. In-situ methods other than the calorimeter
+    so far use up negligible amounts of memory on the accelerator.
```
ax3l: device

docs/source/usage/plugins.rst
n01r: yeah, I was a little confused where to put it but I guess it works if I just show the example in the workflows and then link to usage-python-utils as you suggest

lib/python/picongpu/utils/picongpu_memory_calculator.py
n01r: Jup, saw and [noted](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2806#issuecomment-439960845) it :)
Could be nice to do it that way - I'll try .

docs/source/usage/workflows/memoryPerAccelerator.rst
sbastrakov: I think in this case calling CPUs "accelerators" would have been even more confusing.

lib/python/picongpu/utils/memory_calculator.py
```diff
@@ -0,0 +1,368 @@
...
+            required memory {unit: bytes} per accelerator
+        """
+        if n_x is None:
+            n_x = self.n_x
+        if n_y is None:
+            n_y = self.n_y
+        if n_z is None:
+            n_z = self.n_z
+
+        req_mem_per_cell = 6 * 8  # bytes
```
psychocoderHPC: The default RNG on GPU is `XorMin` and use `6 * 4 byte` per cell on CORE + BORDER (GUARD has currently no rng state)

psychocoderHPC: `6 * 8 byte` is only correct for `MRG32k3aMin` (not the default)

n01r: Perfect, thanks!

**offline** with @psychocoderHPC:
the default for CPU and `AlpakaRand` is TinyMT with a state size of `7 * 4` bytes

lib/python/picongpu/utils/picongpu_memory_calculator.py
n01r: Hmm ... I played around a bit with some solutions but they don't make me happy.
I will leave that up for later.

I updated the documentation and the example and will leave it up for review after my next push.

docs/source/usage/workflows/memoryPerAccelerator.rst
ax3l: everything is a device now :)
(and we also program them as if all devices would be accelerators, muhahaha)

docs/source/usage/workflows/memoryPerDevice.rst
```diff
@@ -0,0 +1,115 @@
...
+.. sectionauthor:: Marco Garten
+
+The planning of simulations for realistically sized problems requires a careful estimation of memory usage and is often a trade-off between resolution of the plasma, overall box size and the available resources.
+The file :ref:`memory_calculator.py <usage-python-utils>` contains a class for this purpose.
+
+The following paragraph shows the use of the ``MemoryCalculator`` for the ``4.cfg`` setup of the :ref:`FoilLCT example <usage-examples-foilLCT>` example.
+
+.. code:: python
+
+    from picongpu.utils import MemoryCalculator
```
ax3l: you know that you can check in this file as a script, `literalinclude` it here and even *run it at sphinx build time* to generate the *output below*? :)

docs/source/usage/workflows/memoryPerDevice.rst
```diff
@@ -0,0 +1,115 @@
+.. _usage-workflows-memoryPerDevice:
+
+Calculating the memory requirement per device
```
ax3l: ### Nouns in Titels

My new book! :)

n01r: WTH, I thought I already fixed that.
Sorry that you had to point it out again. :/

docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -0,0 +1,114 @@
...
+rng_gpu = pmc.mem_req_by_rng(Nx, Ny, Nz)
+
+print("+ species:")
+print("- e: {:.2f} MB".format(e_gpu * megabyte))
+print("- H: {:.2f} MB".format(H_gpu * megabyte))
+print("- C: {:.2f} MB".format(C_gpu * megabyte))
+print("- N: {:.2f} MB".format(N_gpu * megabyte))
+print("+ RNG states: {:.2f} MB".format(
+    rng_gpu * megabyte))
+print(
```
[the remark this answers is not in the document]
ax3l: you can also use... haaaands :open_hands:

PR #2818, opened 2018-11-21
Doc: Add System Links

ax3l:
Add more information on system user guides, and production directories so people have a quick time starting with PIConGPU.

### DIFF
--- a/docs/source/install/profile.rst
+++ b/docs/source/install/profile.rst
@@ -21,6 +21,12 @@ We listed some example ``picongpu.profile`` files below which can be used to set
 Hemera (HZDR)
 -------------
 
+**System overview:** `link (internal) <https://www.hzdr.de/db/Cms?pOid=29813>`_
+
+**User guide:** *None*
+
+**Production directory:** ``/bigdata/hplsim/`` with ``external/``, ``scratch/``, ``development/`` and ``production/``
+
 For this profile to work, you need to download the :ref:`PIConGPU source code <install-dependencies-picongpu>` manually.
 
 Queue: defq (2x Intel Xeon Gold 6148, 20 Cores + 20 HyperThreads/CPU)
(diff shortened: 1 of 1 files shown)


docs/source/install/profile.rst
```diff
@@ -38,6 +44,12 @@ Queue: gpu (4x NVIDIA P100 16GB)
 Hypnos (HZDR)
 -------------
 
+**System overview:** `link (internal) <https://www.hzdr.de/db/Cms?pOid=29813>`_
+
+**User guide:** *None*
```
ax3l: Isn't there a funny web server mini-wiki running on `http://hypnos4/wiki` or so?

PrometheusPi: I do not know - that is the first time I hear this.

psychocoderHPC: it is `http://hypnos3/wiki` but only available within our intranet.

docs/source/install/profile.rst
```diff
@@ -38,6 +44,12 @@ Queue: gpu (4x NVIDIA P100 16GB)
...
+
+**Production directory:** ``/bigdata/hplsim/`` with ``external/``, ``scratch/``, ``development/`` or ``production/``
```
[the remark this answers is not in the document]
ax3l: had that initially as well. now changed back.

docs/source/install/profile.rst
```diff
@@ -103,6 +134,12 @@ For this profile to work, you need to download the :ref:`PIConGPU source code <i
 Taurus (TU Dresden)
 -------------------
 
+**System overview:** `link <https://tu-dresden.de/zih/hochleistungsrechnen/hpc>`_
+
+**User guide:** `link <https://doc.zih.tu-dresden.de/hpc-wiki/bin/view/Compendium/SystemTaurus>`_
+
+**Production directory:** ``/scratch/$USER/`` and ``/scratch/$proj/``
```
PrometheusPi: `$proj` is not defined by default

PrometheusPi: you could write:
```bash
/scratch/`groups | awk '{print $1}'`/
```

ax3l: nah, we define `$proj` in the Titan profiles already and should add it to the taurus `picongpu.profile` as well.

PrometheusPi: I am fine with that.

ax3l: ok, will do as a follow-up

PR #2822, opened 2018-11-23
MemoryCalculator: fix example output for documentation

n01r:
This commit fixes the documentation output of the `MemoryCalculator` example script.

auswahl_047

~~So far I cannot check this because my own `sys.path` contains `...lib/python/picongpu`.~~

I built the doxygen versions **1.8. .. 6, 13** and **14** and all of them fail building the picongpu docu on hypnos.
```
Generating XML output for class pmacc::DataBoxUnaryTransform
/home/garten70/src/picongpu/include/pmacc/memory/boxes/DataBoxUnaryTransform.hpp:34: warning: explicit link request to 'result' could not be resolved
Generating XML output for class pmacc::DataConnector
Segmentation fault (core dumped)
```

**Edit:**   
After being able to test it I can now say the fix does actually work

### DIFF
--- a/docs/source/usage/python/utils.rst
+++ b/docs/source/usage/python/utils.rst
@@ -6,14 +6,11 @@ Memory Calculator
 To aid you in the planning and setup of your simulation PIConGPU provides python tools for educated guesses on simulation parameters.
 They can be found under ``lib/python/picongpu/utils``.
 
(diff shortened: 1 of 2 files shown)


n01r:
I noticed that for me neither `sys.path.insert` nor `sys.path.append` inside the `docs/source/conf.py` actually worked because when I then printed `sys.path` it didn't contain `$HOME/src/picongpu/lib/python/`.

PR #2825, opened 2018-11-26
Update FreeRng.def

hightower8083:
fix a typo in the example code


psychocoderHPC:
@hightower8083 Well spotted! Tank you for the fix.

ax3l:
Thanks @hightower8083!

Added you now to the developers team (you should have gotten a GitHub notification that needs to be confirmed) so any further PRs can be checked by CI. (Not needed for this one, since it's a comment change that is not covered by CI anyway).

Issue #2827, opened 2018-12-04
How to install the PIConGPU code by sources files in Tianhe2 supercomputer center?


QJohn2017:
Hi, I'm a newer for the PIConGPU code, I want to install the code in my Tianhe2 supercomputer center from the source code. I have tried to install the code by using the `PIConGPU Documentation` guide manual, but it failed. I can't find the `configure` file in the source code, so I can't install it and make it, can anyone can help me? Thank you !

`-bash: ./configure: No such file or directory`

ax3l:
Dear @QJohn2017,

just to add to the information on getting PIConGPU set up on Tianhe2.

We usually do the following: when you log into a supercomputer, `source` a script that loads provided modules and set some environment variables to enable tools and paths for your current workflow (running PIConGPU). Here is how this `picongpu.profile` script looks for other systems already: https://picongpu.readthedocs.io/en/0.4.2/install/profile.html

Does Tianhe2 have some software modules we can use? Can you link us a public user guide for the system?
Here are our requirements you can share with your system user support: https://picongpu.readthedocs.io/en/0.4.2/install/dependencies.html

Please feel free to add more details if you need any help getting started. We are glad to assist you. If it helps, you can also try to apply for user accounts for us so we can interact with system support directly during the setup phase.

QJohn2017:
Hi, @sbastrakov , Yes, I followed the guide that you pointed out to install the code.  But the installation description is not clear to me. The PIC codes I have used, like the EPOCH, just compiled the code to create an executable  file and modify the input  file.   So, i want to installation successful firstly, can you help me? Thank you !

QJohn2017:
Hi @ax3l , sorry I have only the Chinese user guide of the Tianhe2 supercomputer center, but have no English user guide.  The attachment file is the software modules in Tianhe2 supercomputer center. Can it do some help for the code installation?
[module.xlsx](https://github.com/ComputationalRadiationPhysics/picongpu/files/2643879/module.xlsx)

ax3l:
Thanks! The modules are plenty but a bit old, including the compiler sets.
Do you think you can forward your system support [this page](https://picongpu.readthedocs.io/en/0.4.2/install/dependencies.html) so they can install recent modules for you? Ideally based on a GCC 7.3 for good KNC support.

QJohn2017:
@ax3l , Can I install the modules by myself or use the current modules to install the code directly?

ax3l:
[...]

PR #2828, opened 2018-12-04
Docs: Source Intro Details

ax3l:
Make the intro in compiling from source more clear.
Some users (e.g. #2827) mismatch it with the actual install that follows later on.


docs/source/install/instructions/source.rst
```diff
@@ -75,8 +83,16 @@ The syntax is still very similar:
...
-Continue with the following section to build our dependencies.
+One can usually pass further options with ``--with-something=VALUE`` or ``--enable-thing`` to ``configure``.
+See ``configure --help`` when installing an *autotools* project.
+
+That is all on the theory of compiling projects!
+
+Now Start
+^^^^^^^^^
+
+You now know all basics to install from source.
```
sbastrakov: I think should be `all the basics`

ax3l: yeah, was thinking about the article and it's actually not necessary but... ok.

ax3l:
Updated a bit on "build" terminology.
This section is just giving a conceptual user-level intro into builds from source, e.g. linking is not necessary to be explained and is intentionally subsumed in build/compile speech.

PR #2830, opened 2018-12-04
Replace output by stdout

n01r:
We write an `output` file on our PBS system because `stdout` doesn't appear until the simulation is done.
But if we run on SLURM there never is an `output` file, thus breaking the script.

### DIFF
--- a/lib/python/picongpu/utils/find_time.py
+++ b/lib/python/picongpu/utils/find_time.py
@@ -27,7 +27,7 @@ class FindTime(object):
             raise ValueError('The run_directory parameter can not be None!')
 
         self.run_directory = run_directory
-        self.data_file = "output"
+        self.data_file = "stdout"
 
         self.dt = self.get_dt()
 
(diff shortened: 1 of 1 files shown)


ax3l:
We should create an `output` file in all cases, which script is lacking it?

n01r:
My runs from Piz Daint don't have `simOutput/output` files.
So when I tried to use the `phase_space_visualizer` it returned an error, asking if the simulation ever ran.
This I tracked back to the `find_time.py` which looked for the `output` file I don't have.

ax3l:
Since both files are kind of ad-hoc solutions (`stdout` is a convention as well in `tpl` files) we could try first one and then the other.

psychocoderHPC:
What do you think about: creating a symlink to `stdout` named `simOutput/output` in the case we are on a system with where `stdout` is live updated.

ax3l:
Would also work!

ax3l:
ping @n01r maybe check `output` first and if missing try `stdout`

ax3l:
Implemented generally as #2839 now :)

Issue #2702, opened 2018-09-10
negative density leads to mallocMC error


PrometheusPi:
When initializing a negative density in `density.param`, PIConGPU crashes with a mallocMC error. 
This could be avoided by capping densities to zero or larger values. 

@kossag14 Could you verify whether a non-zero density solved your problem?
@ComputationalRadiationPhysics/picongpu-developers would you prefer for capping or crashing?

PrometheusPi:
@ax3l That is a good point. One should never remove that multiplication. 
I totally agree with you from a performance point of view. However, how much does such a check or multiplication cost?

sbastrakov:
I probably don't fully understand the context, so a naive question. Why isn't it a case of incorrect parameters given by a user? I mean, does negative density make any more sense than e.g. negative amount of cells?

PrometheusPi:
@sbastrakov It definitely is a case of *incorrect parameters given by a user*. No question about that. 
The question for me is, whether it would be more user-friendly to avoid this issue by default or throw a meaningful error message.

**Current case:**
In the current case, the density is described by a complex and lengthy equation. A corner case was not correctly covered and thus a negative density could be set. That can happen quite easily because in most free formula density descriptions, the density is described in parts and the equations only produce a valid result in their specific case. 

**As a context:**
For LWFA simulations, running into memory errors is not that uncommon and usually points to an underestimation of density accumulation due to the wakefields and requires a better distribution of the compute units (GPUs). 

**My current opinion:**
I actually liked the fact, that the simulation crashed because it pointed to an (unspecific) error in the setup. However, the fact that it crashed with a memory error is quite confusing.
I personally would prefer catching negative densities and throwing a meaningful error message as it is much more user-friendly.

sbastrakov:
@PrometheusPi thanks for such a detailed answer. I agree with your opinion.

ax3l:
The only thing we can do is to (silently) crop to zero by default and on all user-given (free) functors. By that, the user still needs to figure out why no target is there but at least it does not crash. The RT costs are fine for that, imho, since it does not need anything that's not already in the kernel (data-wise).

PrometheusPi:
@ax3l What about writing out a warning or error. Especially from the GPUs that might not be that easy?

ax3l:
That's not easily possible on the GPU, since you are evaluating / sampling a function at unknown points at runtime.

ax3l:
[...]

PR #2831, opened 2018-12-05
fix particle creation if density zero

### LINKED ISSUE #2823: time-resolved probing of a single location
Hi there,

For some plasma instability studies it would be really useful to have a way to probe the fields (including charge-densities) at a given location, and output it as a resolved time-series. 

I imagine that one simple way to do this might be setting a single immobile probe particle and gather the fields on it. I've see that there is already a way to do the probes, so technically it should not be too complicated to implement such plug-in for a case of a single particle and arbitrary fields.. what you think ?

psychocoderHPC:
fix #2823

In the case where the density functor is returning a density of zero or negative the functor which calculates the number of macro-particles (called `startPosition` functor) should not be called.
This avoids that the functor write e.g. the user who is using a free functor implementation must handle those edge cases.

changes:
 - do not call `startPosition` functor if density is `<= 0.0`

# Tests
- [x] test on hemera (P100)
- [x] test on hypnos (k80)

### DIFF
--- a/include/picongpu/particles/ParticlesInit.kernel
+++ b/include/picongpu/particles/ParticlesInit.kernel
@@ -229,7 +229,10 @@ namespace picongpu
                         totalCellOffset
                     );
 
-                    float_X const realParticlesPerCell = realDensity * CELL_VOLUME;
+                    /** @bug volatile is required for CUDA 9.2 and sm_60 else the compiler will
+                     * optimize out `if(realParticlesPerCell > 0.0_X)` later on.
+                     */
+                    volatile float_X const realParticlesPerCell = realDensity * CELL_VOLUME;
 
                     // create an independent position functor for each cell in the supercell
                     positionFunctorCtx[ idx ] = positionFunctor(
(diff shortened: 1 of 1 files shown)


n01r:
Looks good :)

psychocoderHPC:
I added `volatile` to the calculation of `realParticlesPerCell` to avoid a bug with CUDA 9.2 and `sm_60`. As soon as our cluster has CUDA 10 I will check if it is a CUDA 9.2 only bug or still exists in CUDA 10.

ax3l:
that might unintentionally also add the safeguard for #2702

ax3l:
@psychocoderHPC pls tell me when the P100 tests are successful.

psychocoderHPC:
@ax3l P100 tests are finished. With the `volatile` workaround everything works.

psychocoderHPC:
@ax3l I pushed by accident additional changes to this branch. I reverted the wrong push but this means we need to wait again for the CI and you must approve this PR again.

PR #2737, opened 2018-10-17
DataReader: Emittance & PlotMPL: Emittance, SliceEmittance, EnergyWaterfall

kossag14:
- `emittance_evolution_visualizer`: plots value of transverse emittance for each iteration
- `slice_emittance_visualizer`: plots slice emittance for each slice for one iteration
- `slice_emittance_waterfall_visualizer`: plots slice emittance for each slice for each iteration
- `energy_waterfall_visualizer`: plots energy histogram for each iteration

- [x] depends on #2588

PR #2836, opened 2018-12-07
Compare reader visualizer

kossag14:
compare.py: a reader that allows comparing maximum energy, emittance, bunch energy, bunch charge, energy spread for a set of simulations where input parameter was variated.
compare_1d_visualizer: plots a selected observable (max energy, emittance, charge, ...) over a selected variation (changed input parameter for the set of simulations), other changed input parameters can be seen by hover over the points in the plot.
Here:
maximum energy: defined as the highest energy value for which the respective charge exceeding this energy is at least 1pC.
bunch: values in energy histogram within +/- 1/3 of peak energy for energies larger than 40MeV

### DIFF
--- a/lib/python/picongpu/plugins/data/__init__.py
+++ b/lib/python/picongpu/plugins/data/__init__.py
@@ -4,6 +4,7 @@ from .png import PNGData
 from .radiation import RadiationData
 from .sliceFieldReader import FieldSliceData
 from .emittance import EmittanceData
+from .compare.py import CompareData
 
 __all__ = [
     "EnergyHistogramData",
@@ -12,4 +13,5 @@ __all__ = [
     "RadiationData",
     "FieldSliceData",
     "EmittanceData",
+    "CompareData",
 ]
--- a/lib/python/picongpu/plugins/data/compare.py
+++ b/lib/python/picongpu/plugins/data/compare.py
@@ -1,0 +1,215 @@
+"""
+This file is part of the PIConGPU.
+
+Copyright 2017-2018 PIConGPU contributors
+Authors: Sophie Rudat
+License: GPLv3+
+"""
+from picongpu.plugins.data import EmittanceData
+from picongpu.plugins.data import EnergyHistogramData
+from .base_reader import DataReader
+
+import numpy as np
+import os
+import collections
+import json
+
+
+class CompareData(DataReader):
+    """
+    Data Reader for the Energy Histogram Plugin.
+    """
+
+    def __init__(self, run_directory):
+        """
+        Parameters
+        ----------
+        simulation_directory : string
+            path to the run directory of PIConGPU
+            (the path before ``simOutput/``)
+        """
+        super().__init__(run_directory)
+        self.json_suffix = "scan_ranges.json"
+
+    def get_data_path(self, species, variation, observable,
+                      species_filter="all"):
+        """
+        Return the path to the underlying data file.
+
+        Parameters
+        ----------
+        species : string
+            short name of the particle species, e.g. 'e' for electrons
+            (defined in ``speciesDefinition.param``)
+        species_filter: string
+            name of the particle species filter, default is 'all'
+            (defined in ``particleFilters.param``)
+
+        Returns
+        -------
+        A list of strings with path to all sims in scan and a
+        list with the values for the choosen variation.
+        """
+        self.path_to_scan = os.path.split(os.path.split(
+                                          self.run_directory)[0])[0]
+        self.path_to_json = os.path.join(self.path_to_scan, self.json_suffix)
+        if not os.path.isfile(self.path_to_json):
+            raise IOError('The file {} does not exist.\n'
+                          'Did the simulation already run?'
+                          .format(self.path_to_json))
+        # get species, filter
+        if species is None:
+            raise ValueError('The species parameter can not be None!')
+        if species_filter is None:
+            raise ValueError('The species_filter parameter can not be None!')
+        if variation is None:
+            raise ValueError('The variation parameter can not be None!')
+        if observable is None:
+            raise ValueError('The observable parameter can not be None!')
+
+        # take first value of other variable parameters
+        with open(self.path_to_json) as f:
+            range_dict = json.load(f)
+            filtered_range_dict = dict()
+            for name, attrs in range_dict.items():
+                filtered_range_dict[name] = attrs['values']
+
+        # list relevant sims in scan
+        sims_rundir = []
+        for dir_name in os.listdir(self.path_to_scan):
+            sim_path = os.path.join(self.path_to_scan, dir_name, "run")
+            if os.path.isdir(sim_path):
+                sims_rundir.append(sim_path)
+
+        return sims_rundir, filtered_range_dict[variation]
+
+    def get_iterations(self, species, species_filter="all"):
+        """
+        Return an array of iterations with available data.
+
+        Parameters
+        ----------
+        species : string
+            short name of the particle species, e.g. 'e' for electrons
+            (defined in ``speciesDefinition.param``)
+        species_filter: string
+            name of the particle species filter, default is 'all'
+            (defined in ``particleFilters.param``)
+
+        Returns
+        -------
+        An array with unsigned integers.
+        """
+        iteration = EnergyHistogramData(
+                    self.run_directory).get_iterations(species=species)
+        return iteration
+
+    def _get_for_iteration(self, iteration, species, variation,
+                           observable, species_filter="all", **kwargs):
+        """
+
+        Parameters
+        ----------
+        species : string
+            short name of the particle species, e.g. 'e' for electrons
+            (defined in ``speciesDefinition.param``)
+        variation : string
+            name of the variated input parameter
+        observable: string
+            name of the calculated and compared quantity
+        species_filter: string
+            name of the particle species filter, default is 'all'
+            (defined in ``particleFilters.param``)
+        iteration : (unsigned) int [unitless]
+            The iteration at which to read the data.
+            A list of iterations is allowed as well.
+            ``None`` refers to the list of all available iterations.
+
+        Returns
+        -------params,values,other_params
+        variation : np.array
+            values of the variated input parameter
+        observables : np.array
+            calculated observed values for the respectiv input parameter
+        other_params: dict
+            values of other variated input parameters
+        """
+        # get list of path to files
+        if iteration is not None:
+            if not isinstance(iteration, collections.Iterable):
+                iteration = np.array([iteration])
+        sims_rundir, params = self.get_data_path(species, variation,
+                                                 observable,
+                                                 species_filter)
+        values = []
+        params = []
+        for sim in (sims_rundir):
+            sim_json = os.path.join(os.path.split(sim)[0], 'params.json')
+            if not os.path.isfile(sim_json):
+                raise IOError('The file {} does not exist.\n'
+                              'Did the simulation already run?'
+                              .format(self.path_to_json))
+            with open(sim_json) as f:
+                range_dict = json.load(f)
+                param = range_dict[variation]['values'][0]
+            try:
+                if observable is not "emittance":
+                    counts, bins = EnergyHistogramData(sim).get(
+                                   iteration=iteration, species=species)
+                    start = 80
+                    limitE = 500
+                    data = counts[start:]
+                    for i in np.arange(data.size):
+                        if (max(data) == data[i]):
+                            peakEnergy = i
+                    for j in np.arange(data.size-peakEnergy):
+                        if (max(data)/3 - data[peakEnergy + j] > 0.0):
+                            FWHM_max = ((peakEnergy + j + start) *
+                                        limitE / len(bins))
+                            break
+                    for k in np.arange(peakEnergy):
+                        if (max(data)/3-data[peakEnergy - k] > 0.0):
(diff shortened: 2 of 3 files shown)


sbastrakov:
@kossag14 I believe the PR should be to `dev` branch, not `master`. You could edit it in the current PR (without closing and opening a new one).

lib/python/picongpu/plugins/data/__init__.py
```diff
@@ -3,6 +3,7 @@
 from .png import PNGData
 from .radiation import RadiationData
 from .sliceFieldReader import FieldSliceData
+from .compare import CompareData
```
ax3l: CompareData should be used in `__all__` below :)

lib/python/picongpu/plugins/plot_mpl/compare_1d_visualizer.py
```diff
@@ -0,0 +1,185 @@
...
+                species = arg
+            elif opt in ["-f", "--filter"]:
+                filtr = arg
+            elif opt in ["-o", "--observable"]:
+                observable = arg
+            elif opt in ["-v", "--variation"]:
+                variation = arg
+
+        # check that we got all args that we need
+        if (path is None or iteration is None or
```
ax3l: try removing the `()`, they are not needed

lib/python/picongpu/plugins/data/compare.py
```diff
@@ -0,0 +1,222 @@
...
+                          'Did the simulation already run?'
+                          .format(self.path_to_json))
+        # take first value of other variable parameters
+        with open(self.path_to_json) as f:
+            range_dict = json.load(f)
+            filtered_range_dict = dict()
+            self.fix_param_dict = dict()
+            for name, attrs in range_dict.items():
+                filtered_range_dict[name] = attrs['values']
+                if ((name != variation) &
```
ax3l: the brackets are not needed (the outer *and* the inner ones can go).
the reason for that is [operator precedence](https://docs.python.org/3/reference/expressions.html#operator-precedence)

the `&` should be an `and` (otherwise it's a bitwise operation)

lib/python/picongpu/plugins/data/compare.py
```diff
@@ -0,0 +1,222 @@
...
+                            break
+                    # charge within +/- 1/3 peakEnergy, unit pC
+                    charge = data[cmin:cmax].sum() * 1.6021766209e-19 * 1.e12
+                    value = charge
+                if observable == "emittance":
+                    counts, bins, ts = EmittanceData(sim).get(
+                                       species, species_filter)
+                    emit = counts[iteration[0]][0] * 1.e6
+                    value = emit  # unit: pi mm mrad
+            except:
```
ax3l: That's quite a large try-except block. Where do you expect something to throw? In the data constructors or the `.get()`ters?

We know what we usually throw there, so should do something more precise like:
```python
except (ValueError, IOError, IndexError):
```

or you could also handle them individually to throw a more precise error to the user than "Error in sim" :-)

lib/python/picongpu/plugins/data/compare.py
```diff
@@ -0,0 +1,222 @@
...
+from picongpu.plugins.emittance import EmittanceData
+from picongpu.plugins.energy_histogram import EnergyHistogramData
+
+import numpy as np
+import os
+import collections
+import json
+
+
+class CompareData(object):
```
codingS3b: Please inherit from the base class defined in `base_reader.py`

lib/python/picongpu/plugins/data/compare.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def __init__(self, run_directory):
+        """
+        Parameters
+        ----------
+        simulation_directory : string
+            path to the run directory of PIConGPU
+            (the path before ``simOutput/``)
+        """
+        if run_directory is None:
```
codingS3b: this check is already done in the base class, so you can omit it when you call the base class constructor

lib/python/picongpu/plugins/data/compare.py
```diff
@@ -0,0 +1,222 @@
...
+            raise ValueError('The run_directory parameter can not be None!')
+
+        self.run_directory = run_directory
+        self.json_suffix = "scan_ranges.json"
```
codingS3b: this should probably be the `self.data_file_suffix` parameter to be consistent with the other readers.

lib/python/picongpu/plugins/data/compare.py
```diff
@@ -0,0 +1,222 @@
...
+        self.path_to_json = os.path.join(self.path_to_scan, self.json_suffix)
+        if not os.path.isfile(self.path_to_json):
+            raise IOError('The file {} does not exist.\n'
+                          'Did the simulation already run?'
+                          .format(self.path_to_json))
+        # take first value of other variable parameters
+        with open(self.path_to_json) as f:
+            range_dict = json.load(f)
+            filtered_range_dict = dict()
+            self.fix_param_dict = dict()
```
codingS3b: please declare `self.fix_param_dict` in the constructor (maybe as None) but with a comment on what it will be used for

lib/python/picongpu/plugins/data/compare.py
```diff
@@ -0,0 +1,222 @@
...
+                    (np.size(filtered_range_dict[name]) > 1)):
+                    self.fix_param_dict[name] = filtered_range_dict[name]
+        # list relevant sims in scan
+        sims_rundir = []
+        for dir_name in os.listdir(self.path_to_scan):
+            sim_path = os.path.join(self.path_to_scan, dir_name, "run")
+            if os.path.isdir(sim_path):
+                sims_rundir.append(sim_path)
+        # get species, filter
+        if species is None:
```
codingS3b: Maybe perform all those checks before doing any computations :)

lib/python/picongpu/plugins/data/compare.py
```diff
@@ -0,0 +1,222 @@
...
+            (defined in ``speciesDefinition.param``)
+        species_filter: string
+            name of the particle species filter, default is 'all'
+            (defined in ``particleFilters.param``)
+
+        Returns
+        -------
+        An array with unsigned integers.
+        """
+        return EnergyHistogramData(
```
codingS3b: Is this intended to always use the `EnergyHistogramData` reader here to access iterations? Couldn't this fail if data is available for some plugins but not for others? You also seem not to need the `variation` and `observable` parameters

lib/python/picongpu/plugins/data/compare.py
```diff
@@ -0,0 +1,222 @@
...
+               self.run_directory).get_iterations(species, species_filter)
+
+    def get(self, species, variation, observable, species_filter="all",
```
codingS3b: please change that to `_get_for_iteration(...)`

lib/python/picongpu/plugins/data/compare.py
```diff
@@ -0,0 +1,222 @@
...
+                            break
+                    for k in np.arange(peakEnergy):
+                        if (max(data)/3-data[peakEnergy - k] > 0.0):
+                            FWHM_min = ((peakEnergy - k + start)
+                                        * limitE / len(bins))
+                            break
+                if observable == "bunchEnergy":
+                    maxEnergy = 0.5 * (FWHM_max + FWHM_min)
+                    value = maxEnergy  # unit: MeV
+                if observable == "energySpread":
```
codingS3b: Shouldn't that rather be a big `if ... else if ...else if...else` block instead?

lib/python/picongpu/plugins/plot_mpl/compare_1d_visualizer.py
```diff
@@ -0,0 +1,185 @@
...
+License: GPLv3+
+"""
+
+from picongpu.plugins.compare import CompareData
+from picongpu.plugins.plot_mpl.base_visualizer import Visualizer as\
+    BaseVisualizer, plt
+import numpy as np
+
+
+class Visualizer(BaseVisualizer):
```
codingS3b: Please check the updated interface of the `base_visualizer.py` and refactor the code in a similar way as suggested for your PR #2737

lib/python/picongpu/plugins/data/compare.py
ax3l: @kossag14 please work-in this comment as well

lib/python/picongpu/plugins/data/compare.py
ax3l: @kossag14 please work-in this comment as well

lib/python/picongpu/plugins/data/compare.py
```diff
@@ -0,0 +1,218 @@
...
+        species : string
+            short name of the particle species, e.g. 'e' for electrons
+            (defined in ``speciesDefinition.param``)
+        species_filter: string
+            name of the particle species filter, default is 'all'
+            (defined in ``particleFilters.param``)
+
+        Returns
+        -------
+        A string with a path.
```
codingS3b: Documented return type seems outdated.

lib/python/picongpu/plugins/data/compare.py
```diff
@@ -0,0 +1,218 @@
...
+                                break
+                        for k in np.arange(peakEnergy):
+                            if (max(data) / 3 - data[peakEnergy - k] > 0.0):
+                                cmin = (peakEnergy - k)
+                                break
+                        # charge within +/- 1/3 peakEnergy, unit pC
+                        charge = (data[cmin:cmax].sum() *
+                                  1.6021766209e-19 * 1.e12)
+                        value = charge
+                elif observable == "emittance":
```
codingS3b: this should be an `else` since you first ask `if observable is not "emittance"`. Also make your checks with `==` rather than with `is` (see https://stackoverflow.com/questions/1504717/why-does-comparing-strings-in-python-using-either-or-is-sometimes-produce)

lib/python/picongpu/plugins/data/compare.py
```diff
@@ -0,0 +1,218 @@
...
+                    counts, bins, ts = EmittanceData(sim).get(
+                                       species, species_filter)
+                    emit = counts[iteration[0]][0] * 1.e6
+                    value = emit  # unit: pi mm mrad
+            except (ValueError, IOError, IndexError):
+                print("Error in sim ", sim)
```
codingS3b: wouldn't you want to raise here as well?

lib/python/picongpu/plugins/plot_mpl/compare_1d_visualizer.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    Class for creation of histogram plots on a logscaled y-axis.
+    """
+
+    def __init__(self, run_directories=None, ax=None):
+        """
+        Parameters
+        ----------
+        run_directory : string
```
codingS3b: fix docstring

lib/python/picongpu/plugins/plot_mpl/compare_1d_visualizer.py
```diff
@@ -0,0 +1,163 @@
...
+    def visualize(self, **kwargs):
+        """
+        Creates a semilogy plot on the provided axes object for
+        the data of the given iteration using matpotlib.
+
+        Parameters
+        ----------
+        iteration: int
+            the iteration number for which data will be plotted.
+        ax: matplotlib axes object
```
codingS3b: remove `ax` docstring

lib/python/picongpu/plugins/plot_mpl/compare_1d_visualizer.py
```diff
@@ -0,0 +1,163 @@
...
+            sys.exit(2)
+        if species is None:
+            species = 'e'
+            print("Particle species was not given, will use", species)
+        if filtr is None:
+            filtr = 'all'
+            print("Species filter was not given, will use", filtr)
+
+        fig, ax = plt.subplots(1, 1)
+        Visualizer(path).visualize(ax, iteration=iteration, species=species,
```
codingS3b: `ax` needs to be passed on creation.

lib/python/picongpu/plugins/data/compare.py
codingS3b: You are catching an exception and print a statement once such an exception happens. But wouldn't you also want to re-raise the exception that you caught so the program would still fail instead of continuing to run?

codingS3b: Okay, then it's fine :)

ax3l:
@codingS3b can you please review the latest updates? :)

ax3l:
@kossag14 can you please squash the 10 commits into one and rebase against the latest `dev` to solve the little conflict in `lib/python/picongpu/plugins/data/__init__.py`?

ax3l:
@codingS3b can you please take a look at the updated PR?

PrometheusPi:
What is the status of this pull request. 
Should I take over? Should we just reject it?

codingS3b:
I created the follow up PR #2854 but have not worked on it any more since lots of things are not clear to me and Sophie, who wrote the code, is not at HZDR anymore, so probably closing would be best.

ax3l:
@PrometheusPi maybe you can take over on the follow-up PR?

PrometheusPi:
@ax3l I will do that

PrometheusPi:
Since the follow up pull request #2854  was closed (for now) and this is even older, I would close this pull request (for now?) as well.

PR #2837, opened 2018-12-10
remove ZigZag current deposition 

psychocoderHPC:
The ZigZag current deposition is not conserving the charge.
Since we the have the well working EmZ there is not need to keep the original ZigZag implementation in the active code base.

- update species.param: remove ZigZag solver from the list of the current deposition solver
- example
  - KelvinHelmholtz: remove ZigZag with Emz solver
  - LaserWakefield: remove EmZ tests which are already tested in the KHI example


psychocoderHPC:
> Can we instead limit it to CIC and commit the version that we corrected? (Where I added the missing corner terms?)

The ZigZag implementation was also wrong for CIC. We added the missing terms in the CIC equation only to some playground prototype implementations which differ from this implementation.
I currently do not like to waste time to reintroduce an old different ZigZag implementation and repeat all tests to get the PR expected. 
At the current time, we will have no benefit from that.

psychocoderHPC:
for documentation: https://github.com/psychocoderHPC/picongpu/commits/topic-emzPaperZigZagAsEmZSolver

This is the hacky branch we used in the past.

ax3l:
Works for me, I just pulled it as a backup and can add it later on again for CIC only.

include/picongpu/fields/currentDeposition/EmZ/EmZ.def
```diff
@@ -45,6 +45,20 @@ namespace emz
  * the ZigZag way splitting.
  * EmZ support arbitrary symmetric shapes and 2D/3D cartesian grids.
```
ax3l: supports

psychocoderHPC: not part of this PR but fixed

ax3l: (there is still the compile error in the test suite, that's why I commented since you had to update anyway)

sbastrakov: Not related to this PR, just to ask / potentially raise an issue.

The incomplete (and now fixed) type is `fields::Solver::NummericalCellType`. Besides having a typo (double "m"), this name is imo weird for other reasons. Currently it is defined inside field solvers, so needless to say it is about numerics, "numerical" in the name is just excessive. Ofc, it is just a simple renaming and arguably may be not even worth doing just for the sake of it.

However, I feel this name is used in contexts with no logical connection to field solvers. Like that, previously failing, piece of code in Particles.kernel is just a spurious dependency. In fact, glancing through the uses of `fields::Solver::NummericalCellType`, it seems all of them just need the cell type, not the field solver. 

I understand that it is the field solver of choice that defines the cell type, but think the rest of the code should not know of that. So as a potential improvement I suggest the following:
* rename `fields::Solver::NummericalCellType` to `fields::Solver::CellType`
* rename `include/picongpu/fields/numericalCellTypes/YeeCell.hpp` to `include/picongpu/fields/cellTypes/Yee.hpp` and similar for others
* create a new entity that holds the cell type, which is somehow initialized with the type defined by the used field solver, and make all non solver-related code use that instead of `fields::Solver::NummericalCellType` to get rid of spurious dependencies.

I assume the first two are easy mechanical steps and the third might have some pitfalls I haven't considered.

ax3l: yes, one can skip the `numerical`

ax3l: Yes, that was the original intention and refactoring PRs to reduce that include load are welcome, fully agree.

PR #2840, opened 2018-12-13
auto domain adjustment

psychocoderHPC:
Add helper to calculate/adjust the domain sizes and offset of the simulation.

- add class `AdjustDomainSize`
- remove domain validation from `MySimulation`
- add new command line parameter `--autoAdjustGrid`

If the domain is automatically adjusted the changes will be logged on `stdout`. I know that this is not optimal but since we are lacking a logger this is the best way.

### DIFF
--- a/docs/source/usage/workflows/numberOfCells.rst
+++ b/docs/source/usage/workflows/numberOfCells.rst
@@ -8,10 +8,14 @@ Setting the Number of Cells
 Together with the grid resolution in :ref:`grid.param <usage-params-core>`, the number of cells in our :ref:`.cfg files <usage-tbg>` determine the overall size of a simulation (box).
 The following rules need to be applied when setting the number of cells:
 
-Each GPU needs to:
+Each device needs to:
 
 #. contain an integer *multiple* of supercells
 #. at least *three* supercells
+#. for non periodic boundary conditions, the number of absorbing boundary cells for devices at the simulation boundary (see :ref:`grid.param <usage-params-core>`) must fit into the local volume
+
+The grid size will be automatically adjusted if the conditions above are not fulfilled. 
+This behavior can be disabled by using the command line option ``--autoAdjustGrid off``
 
 Supercell sizes in terms of number of cells are set in :ref:`memory.param <usage-params-memory>` and are by default ``8x8x4`` for 3D3V simulations on GPUs.
 For 2D3V simulations, ``16x16`` is usually a good supercell size, however the default is simply cropped to ``8x8``, so make sure to change it to get more performance.
--- a/include/picongpu/simulationControl/DomainAdjuster.hpp
+++ b/include/picongpu/simulationControl/DomainAdjuster.hpp
@@ -30,6 +30,7 @@
 #include <pmacc/dimensions/DataSpace.hpp>
 #include <pmacc/mpi/GetMPI_StructAsArray.hpp>
 
+#include <stdexcept>
 #include <array>
 
 
@@ -106,6 +107,16 @@ namespace picongpu
             localDomainOffset = m_localDomainOffset;
         }
 
+        /** only validate conditions
+         *
+         * Disable domain sizes auto adjustment.
+         * The domain condition will be still checked.
+         */
+        void validateOnly()
+        {
+            m_validateOnly = true;
+        }
+
     private:
 
         /** update local domain offset
@@ -177,9 +188,12 @@ namespace picongpu
 
             if( validLocalSize != m_localDomainSize[ dim ] )
             {
-                std::cout << "Dimension " << dimNames[ dim ] <<
-                    ": local grid size is not a multiple of supercell size. Adjust from " <<
-                    m_localDomainSize[ dim ] << " to " <<  validLocalSize << std::endl;
+                showMessage(
+                    dim,
+                    "Local grid size is not a multiple of supercell size.",
+                    m_localDomainSize[ dim ],
+                    validLocalSize
+                );
 
                 m_localDomainSize[ dim ] = validLocalSize;
             }
@@ -198,9 +212,12 @@ namespace picongpu
             if( numSuperCells < 3 )
             {
                 int newLocalDomainSize = 3 * SuperCellSize::toRT()[ dim ];
-                std::cout << "Dimension " << dimNames[ dim ] <<
-                    ": local grid size is not containing at least 3 supercells. Adjust from " <<
-                    m_localDomainSize[ dim ] << " to " <<  newLocalDomainSize << std::endl;
+                showMessage(
+                    dim,
+                    "Local grid size is not containing at least 3 supercells.",
+                    m_localDomainSize[ dim ],
+                    newLocalDomainSize
+                );
 
                 m_localDomainSize[ dim ] = newLocalDomainSize;
             }
@@ -248,10 +265,13 @@ namespace picongpu
 
                 if( validLocalSize != m_localDomainSize[ dim ] )
                 {
-                    std::cout << "Dimension " << dimNames[ dim ] <<
-                        ": local grid size must be greater or equal than the largest absorber. Adjust from " <<
-                        m_localDomainSize[ dim ] << " to " <<  validLocalSize << std::endl;
-                    
+                    showMessage(
+                        dim,
+                        "Local grid size must be greater or equal than the largest absorber.",
+                        m_localDomainSize[ dim ],
+                        validLocalSize
+                    );
+
                     m_localDomainSize[ dim ] = validLocalSize;
                 }
             }
@@ -293,9 +313,12 @@ namespace picongpu
                 // local size must be equal for all devices in y direction
                 if( m_isMaster && globalMax != globalMin )
                 {
-                    std::cout << "Dimension " << dimNames[ dim ] <<
-                        ": local grid size must be equal for all devices because moving window is enabled. Adjust from " <<
-                        m_localDomainSize[ dim ] << " to " <<  globalMax << std::endl;
+                    showMessage(
+                        dim,
+                        "Local grid size must be equal for all devices because moving window is enabled.",
+                        m_localDomainSize[ dim ],
+                        globalMax
+                    );
                 }
 
                 m_localDomainSize[ dim ] = globalMax;
@@ -342,12 +365,50 @@ namespace picongpu
 
             if( m_isMaster && validGlobalGridSize != static_cast< uint64_t >( m_globalDomainSize[ dim ] ) )
             {
-                    std::cout << "Dimension " << dimNames[ dim ] <<
-                        ": global grid size adjusted from " <<
-                        m_globalDomainSize[ dim ] << " to " <<  validGlobalGridSize << std::endl;
+                showMessage(
+                    dim,
+                    "Invalid global grid size.",
+                    m_globalDomainSize[ dim ],
+                    validGlobalGridSize
+                );
             }
 
             m_globalDomainSize[ dim ] = static_cast< int >( validGlobalGridSize );
+        }
+
+        /** print a message to the user
+         *
+         * Throw an error with the message if is validateOnly was called.
+         *
+         * @param dim dimension index which was checked
+         * @param msg problem description
+         * @param currentSize current domain size in the given direction
+         * @param updatedSize updated/corrected domain size for the given dimension
+         */
+        void showMessage(
+            size_t const dim,
+            std::string const & msg,
+            int const currentSize,
+            int const updatedSize
+        ) const
+        {
+            /**! lookup table to translate a dimension index into a name
+             *
+             * \warning `= { { ... } }` is not required by the c++11 standard but
+             * is necessary for g++ 4.9
+             */
+            std::array< char, 3 > const dimNames = { { 'x', 'y', 'z' } };
+
+            if( m_validateOnly )
+                throw std::runtime_error(
+                    std::string( "Dimension " ) + dimNames[ dim ] + ": " +
+                    msg + " Suggestion: set " + std::to_string( currentSize ) +
+                    " to " + std::to_string( updatedSize )
+                );
+            else
+                std::cout << "Dimension " << dimNames[ dim ] << ": " <<
+                    msg << " Auto adjust from " <<
+                    currentSize << " to " << updatedSize << std::endl;
         }
 
         DataSpace< simDim > m_globalDomainSize;
@@ -358,12 +419,9 @@ namespace picongpu
         DataSpace< simDim > const m_isPeriodic;
         bool const m_movingWindowEnabled;
         bool const m_isMaster;
-        /**! lookup table to translate a dimension index into a name
-         *
-         * \warning `= { { ... } }` is not required by the c++11 standard but
-         * is necessary for g++ 4.9
-         */
-        std::array< char, 3 > dimNames = { { 'x', 'y', 'z' } };
+
+        //! if true it will only validate the conditions
+        bool m_validateOnly = false;
     };
 
 } // namespace picongpu
--- a/include/picongpu/simulationControl/DomainAdjuster.hpp
+++ b/include/picongpu/simulationControl/DomainAdjuster.hpp
@@ -1,0 +1,369 @@
+/* Copyright 2018 Rene Widera
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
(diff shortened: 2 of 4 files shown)


include/picongpu/simulationControl/AdjustDomainSize.hpp
```diff
@@ -0,0 +1,363 @@
...
+         */
+        AdjustDomainSize(
+            DataSpace< simDim > const & numDevices,
+            DataSpace< simDim > const & mpiPosition,
+            DataSpace< simDim > const & isPeriodic,
+            bool const movingWindowEnabled
+        ) :
+            m_numDevices( numDevices),
+            m_mpiPosition( mpiPosition ),
+            m_isPeriodic( isPeriodic),
```
sbastrakov: `( isPeriodic)` -> `( isPeriodic )`

include/picongpu/simulationControl/AdjustDomainSize.hpp
```diff
@@ -0,0 +1,363 @@
...
+            m_movingWindowEnabled( movingWindowEnabled ),
+            m_isMaster( mpiPosition == DataSpace< simDim >::create( 0 ) )
+        {
+        }
+
+        /** adjust the correct grid size
```
sbastrakov: To adjust already implies to correct, so `correct` here is excess.

include/picongpu/simulationControl/AdjustDomainSize.hpp
```diff
@@ -0,0 +1,363 @@
...
+            for( uint32_t d = 0; d < simDim; ++d )
+            {
+                multipleOfSuperCell( d );
+                minThreeSuperCells( d );
+                greaterEqualThanAbsorber( d );
+                deriveGlobalDomainSize( d );
+                updateLocalDomainOffset( d );
+            }
+
+            // write results back
```
sbastrakov: Question: is it not possible have smth like `sizeChanged = (m_globalDomainSize != m_globalDomainSize) || (m_localDomainSize !=  localDomainSize) || (m_localDomainOffset != localDomainOffset )` here? (Instead of having it in different places throughout this class)

include/picongpu/simulationControl/AdjustDomainSize.hpp
```diff
@@ -0,0 +1,363 @@
...
+            m_localDomainSize[ dim ] = validLocalSize;
+        }
+
+        /** ensure that we have a CORE and BORDER region
+         *
+         * Each region must have the size of at least one supercell.
+         *
+         * @param dim dimension to update
+         */
+        void minThreeSuperCells( size_t dim )
```
sbastrakov: Not important, but in other similar situations `dim` is `const`.

include/picongpu/simulationControl/AdjustDomainSize.hpp
```diff
@@ -0,0 +1,363 @@
...
+        {
+            int superCellCount = m_localDomainSize[ dim ] / SuperCellSize::toRT()[ dim ];
+            if( superCellCount < 3 )
+                superCellCount = 3;
```
sbastrakov: Maybe use `std::max` instead of this if statement?

include/picongpu/simulationControl/AdjustDomainSize.hpp
```diff
@@ -0,0 +1,363 @@
...
+         * given dimension is enabled.
+         *
+         * @param dim dimension to update
+         */
+        void greaterEqualThanAbsorber( size_t const dim )
+        {
+            int validLocalSize = m_localDomainSize[ dim ];
+            if(
+                !m_isPeriodic[ dim ] &&
+                // device has absorber
```
sbastrakov: Maybe a self-explanatory variables would be better than the comment? Like 
```
bool const absorberEnabled = !m_isPeriodic[ dim ];
bool const isBoundaryDevice = ( m_mpiPosition[ dim ] == 0 || m_mpiPosition[ dim ] == m_numDevices[ dim ] - 1 );
bool const hasAbsorber = absorberEnabled && isBoundaryDevice ;
if (hasAbsorber) ...
```

include/picongpu/simulationControl/AdjustDomainSize.hpp
```diff
@@ -0,0 +1,363 @@
...
+                ( m_mpiPosition[ dim ] == 0 || m_mpiPosition[ dim ] == m_numDevices[ dim ] - 1 )
+            )
+            {
+                int const maxAbsorberCells = static_cast< int >(
```
sbastrakov: I might have misunderstood it, but the logic of taking the maximum seems wrong to me. What if we have multiple devices along the current direction and the right-side absorber is longer (more cells) than left-side. Then it is possible that the left-most local domain is large enough to fit the left absorber, but not enough for the right absorber, and the right-most local domain is large enough. As I understand it, this case is perfectly fine, but the check would fail.

include/picongpu/simulationControl/AdjustDomainSize.hpp
```diff
@@ -0,0 +1,363 @@
...
+                }
+
+                m_localDomainSize[ dim ] = validLocalSize;
+            }
+        }
+
+        /** derive the local domain size
+         *
+         * Calculate the local domain size.
+         * This function take into account that the local domain size must be
```
sbastrakov: `take` -> `takes` (or remove `This function` and start with `Take`).

include/picongpu/simulationControl/AdjustDomainSize.hpp
```diff
@@ -0,0 +1,363 @@
...
+
+                pmacc::mpi::MPIReduce mpiReduce;
+
+                int globalMax;
+                mpiReduce(
+                    pmacc::nvidia::functors::Max(),
+                    &globalMax,
+                    &m_localDomainSize[ dim ],
+                    1,
+                    pmacc::mpi::reduceMethods::AllReduce()
```
sbastrakov: Out of curiosity, why not use the version of this method without the last parameter?

include/picongpu/simulationControl/AdjustDomainSize.hpp
```diff
@@ -0,0 +1,363 @@
...
+        void minThreeSuperCells( size_t dim )
+        {
+            int superCellCount = m_localDomainSize[ dim ] / SuperCellSize::toRT()[ dim ];
+            if( superCellCount < 3 )
+                superCellCount = 3;
+
+            int validLocalSize = superCellCount * SuperCellSize::toRT()[ dim ];
+            if( validLocalSize != m_localDomainSize[ dim ] )
+            {
+                std::cout << "Dimension " << dim <<
```
sbastrakov: I guess for a user it is more natural to operate and see log messages in terms of x, y, z, not "dimension 0 / 1 / 2". Also in case, e.g. the output only has "dimension 1", without looking to the code it is not even clear if it means x or y.

include/picongpu/simulationControl/MySimulation.hpp
```diff
@@ -213,16 +214,26 @@ class MySimulation : public SimulationHelper<simDim>
...
             // calculate local grid points & offset
             gridSizeLocal[dim] = parserGD.getLocalSize(myGPUpos[dim]);
-            gridOffset[dim] = parserGD.getOffset(myGPUpos[dim], global_grid_size[dim]);
         }
         // by default: use an equal distributed box for all omitted params
         for (uint32_t dim = gridDistribution.size(); dim < simDim; ++dim)
         {
-            gridSizeLocal[dim] = global_grid_size[dim] / gpus[dim];
-            gridOffset[dim] = gridSizeLocal[dim] * myGPUpos[dim];
+            gridSizeLocal[dim] = globalGridSize[dim] / gpus[dim];
```
sbastrakov: In case it does not lead to too much changes, maybe rename one of `gridSizeLocal` and `globalGridSize` to make the names uniform?

include/picongpu/simulationControl/MySimulation.hpp
```diff
@@ -213,16 +214,26 @@ class MySimulation : public SimulationHelper<simDim>
...
+        AdjustDomainSize adjustDomainSize(
+            gpus,
+            myGPUpos,
+            isPeriodic,
+            slidingWindow
+        );
+
+        DataSpace<simDim> gridOffset;
+
+        // adjust domain offset and sizes
```
sbastrakov: I think the next line very clearly explains itself and so the comment is not needed.

ax3l:
Cool, now we just need a way to express relative positioning for `density.param` so targets do not hop out of the focal stop transversely (or alternatively: absolute positioning for the laser x/z focus position).

include/picongpu/simulationControl/AdjustDomainSize.hpp
psychocoderHPC: That is correct I will update this part and handle the cases correct.

include/picongpu/simulationControl/AdjustDomainSize.hpp
sbastrakov: I'm not sure either, hence asked.

psychocoderHPC:
@sbastrakov Thanks for the feedback I added/refactored all

@ax3l 
>  ... (or alternatively: absolute positioning for the laser x/z focus position) ...

I think a refactoring of the lasers is the best solution. Since we have free functors for the density the user can decide if he prefers to describe the density relative or absolute.

ax3l:
> I think a refactoring of the lasers is the best solution.

Agreed!

include/picongpu/simulationControl/DomainAdjuster.hpp
```diff
@@ -0,0 +1,372 @@
...
+        DataSpace< simDim > const m_mpiPosition;
+        DataSpace< simDim > const m_isPeriodic;
+        bool const m_movingWindowEnabled;
+        bool const m_isMaster;
+        /**! lookup table to translate a dimension index into a name
+         *
+         * \warning `= { { ... } }` is not required by the c++11 standard but
+         * is necessary for g++ 4.9
+         */
+        std::array< char, 3 > dimNames = { { 'x', 'y', 'z' } };
```
psychocoderHPC: NOTE: g++ 4.9 is not standard conform and requires extra brackets around the initializer list.

ax3l: I think that's actually a C++11 thing and was lifted in C++14, as I read the other day on SO.

include/picongpu/simulationControl/DomainAdjuster.hpp
```diff
@@ -0,0 +1,372 @@
...
+         */
+        void greaterEqualThanAbsorber( size_t const dim )
+        {
+            int validLocalSize = m_localDomainSize[ dim ];
+
+            bool const isAbsorberEnabled = !m_isPeriodic[ dim ];
+            bool const isBoundaryDevice = ( m_mpiPosition[ dim ] == 0 || m_mpiPosition[ dim ] == m_numDevices[ dim ] - 1 );
+            if( isAbsorberEnabled && isBoundaryDevice )
+            {
+                size_t boundary = m_mpiPosition[ dim ] == 0u ? 0 : 1u;
```
ax3l: `0u: 1u;`

include/picongpu/simulationControl/DomainAdjuster.hpp
```diff
@@ -0,0 +1,372 @@
...
+                int maxAbsorberCells = ABSORBER_CELLS[ dim ][ boundary ];
+
+                if( m_movingWindowEnabled && dim == 1 )
```
ax3l: 1u

include/picongpu/simulationControl/DomainAdjuster.hpp
```diff
@@ -0,0 +1,372 @@
...
+         *
+         * Calculate the local domain size.
+         * This function takes into account that the local domain size must be
+         * equal for all domains if moving window is activated.
+         *
+         * @param dim dimension to update
+         */
+        void deriveLocalDomainSize( size_t const dim )
+        {
+            if( m_movingWindowEnabled && dim == 1 )
```
ax3l: 1u

include/picongpu/simulationControl/DomainAdjuster.hpp
```diff
@@ -0,0 +1,372 @@
...
+
+        /** derive the global domain size
+         *
+         * Calculate the global domain size.
+         *
+         * @param dim dimension to update
+         */
+        void deriveGlobalDomainSize( size_t const dim )
+        {
+            uint64_t validGlobalGridSize = 0;
```
ax3l: 0u

include/picongpu/simulationControl/DomainAdjuster.hpp
```diff
@@ -0,0 +1,372 @@
...
+
+            deriveLocalDomainSize( dim );
+
+            if( m_movingWindowEnabled && dim == 1 )
```
ax3l: 1u

include/picongpu/simulationControl/DomainAdjuster.hpp
psychocoderHPC: NO as I know the double brackets was only needed if you have not the assign operator but since gcc 4.9 was early stage of C++11 it is maybe a bug but not really important.

include/picongpu/simulationControl/DomainAdjuster.hpp
```diff
@@ -0,0 +1,372 @@
...
+         */
+        void minThreeSuperCells( size_t const  dim )
+        {
+            int superCellCount = std::max(
+                m_localDomainSize[ dim ] / SuperCellSize::toRT()[ dim ],
+                3
+            );
+
+            int validLocalSize = superCellCount * SuperCellSize::toRT()[ dim ];
+            if( validLocalSize != m_localDomainSize[ dim ] )
```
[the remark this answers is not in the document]
psychocoderHPC: I would like it as it is because all rules are written in the same way. 

```
- calculate size based on the condition
- check if conditional size differ to old size
- set correct size
```

include/picongpu/simulationControl/MySimulation.hpp
```diff
@@ -213,16 +214,24 @@ class MySimulation : public SimulationHelper<simDim>
...
         {
-            gridSizeLocal[dim] = global_grid_size[dim] / gpus[dim];
-            gridOffset[dim] = gridSizeLocal[dim] * myGPUpos[dim];
+            gridSizeLocal[dim] = gridSizeGlobal[dim] / gpus[dim];
         }
 
-        Environment<simDim>::get().initGrids(global_grid_size, gridSizeLocal, gridOffset);
+        DataSpace<simDim> gridOffset;
+
+        DomainAdjuster domainAdjuster(
```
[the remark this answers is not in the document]
psychocoderHPC: good idea. But I would say it must be enabled by default.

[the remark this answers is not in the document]
ax3l: ok with me, so a `--no-resize-domain` flag then?

include/picongpu/simulationControl/DomainAdjuster.hpp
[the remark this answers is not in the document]
ax3l: ok, but the check here can be a three-liner and one might not want to outsmart the reader ;-)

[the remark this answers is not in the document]
psychocoderHPC: I rewrote this part.

psychocoderHPC:
@ax3l I added the parameter `autoAdjustGrid` and updated the docs.
The reason why this parameter is called `autoAdjustGrid` and not `autoAdjustDomain` is that the other option for the user to set the simulation volume size is also called `-g,--grid`

docs/source/usage/workflows/numberOfCells.rst
```diff
@@ -8,10 +8,14 @@ Setting the Number of Cells
...
 
-Each GPU needs to:
+Each device needs to:
 
 #. contain an integer *multiple* of supercells
 #. at least *three* supercells
+#. for non periodic boundaries the absorber must fit into the local volume
+
+The grid size will be automatically adjusted if the conditions above are not fulfilled. 
+This behavior can be disabled by using the command line option `--autoAdjustGrid off`
```
ax3l: needs two `` in `r.st` :)

docs/source/usage/workflows/numberOfCells.rst
```diff
@@ -8,10 +8,14 @@ Setting the Number of Cells
 Together with the grid resolution in :ref:`grid.param <usage-params-core>`, the number of cells in our :ref:`.cfg files <usage-tbg>` determine the overall size of a simulation (box).
 The following rules need to be applied when setting the number of cells:
 
-Each GPU needs to:
+Each device needs to:
 
 #. contain an integer *multiple* of supercells
 #. at least *three* supercells
+#. for non periodic boundaries the absorber must fit into the local volume
```
ax3l: I think the term `the absorber` is not something a user knows (but our slang).
maybe better write
```suggestion
#. for non periodic boundary conditions, the number of absorbing boundary cells for devices at the simulation boundary (see :ref:`grid.param <usage-params-core>`) must fit into the local volume
```

docs/source/usage/workflows/numberOfCells.rst
psychocoderHPC: :-( ok will fix it

psychocoderHPC:
@ax3l fixed, thx for the feedback

docs/source/usage/workflows/numberOfCells.rst
ax3l: yeah, it's super confusing that rst and md have this subtle difference xD

ax3l:
@psychocoderHPC Just to make sure: this works well with `--gridDist`?

psychocoderHPC:
yes `--gridDist` is working very well too

PR #2846, opened 2019-01-02
Update Versions Script: Containers

ax3l:
Also update container docs on version update.


ax3l:
@psychocoderHPC CI passed :)

PR #2847, opened 2019-01-04
Docker & Singularity Updates

ax3l:
OpenMPI Vader fix:
- instead of disabling vader alltogether, disable the non-functioning copy-mechanism in Docker and allow OpenMPI 3+ usage

OpenMPI fabrics with spack:
- build all possible fabrics, do not rely on auto-detection which will never have IB (psm, psm2 and mxm do not build without actual HW drivers detected during OpenMPI configure time)

OpenMPI parallel IO work-around:
- crashes and data corruption in releases 2.0-4.0, fallback to ROMIO #2857 

OpenMPI CUDA:
- enable CUDA awareness

IceT:
- buggy CMake install path corrected in `package.py` (spack mainline) https://github.com/spack/spack/pull/5014

Docker/Singularity:
- fix access rights to example dirs for non-root users

A big thank you to @AdamSimpson for provided feedback and improvements! :rocket: :sparkles:

### DIFF
--- a/share/picongpu/dockerfiles/ubuntu-1604/Dockerfile
+++ b/share/picongpu/dockerfiles/ubuntu-1604/Dockerfile
@@ -62,13 +62,19 @@ RUN        /bin/echo -e "source $SPACK_ROOT/share/spack/setup-env.sh\n" \
                         "spack load $PIC_PACKAGE\n" \
                         'if [ $(id -u) -eq 0 ]; then\n' \
                         '   function mpirun { $(which mpirun) --allow-run-as-root $@; }\n' \
+                        '   export -f mpirun\n' \
                         'fi\n' \
-                        'export -f mpirun\n' \
                         'if [ $(id -u) -eq 0 ]; then\n' \
                         '   function mpiexec { $(which mpiexec) --allow-run-as-root $@; }\n' \
+                        '   export -f mpiexec\n' \
                         'fi\n' \
-                        'export -f mpiexec\n' \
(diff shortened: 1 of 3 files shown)


share/picongpu/dockerfiles/ubuntu-1604/Dockerfile
```diff
@@ -89,6 +89,11 @@ RUN        /bin/bash -l -c ' \
                pic-build -b "cuda:30;35;37;50;60;70" -c'-DCUDAMEMTEST_ENABLE=OFF' && \
                rm -rf .build'
 
+# make input directories readable and files executable for all users
+RUN        chmod a+rx /opt/picInputs/*/bin/* && \
```
PrometheusPi: since you are adding `r` in the next line, you could remove it here

share/picongpu/dockerfiles/ubuntu-1604/packages.yaml
```diff
@@ -12,7 +12,8 @@ packages:
       python@2.7.12%gcc@5.4.0 arch=linux-ubuntu16-x86_64: /usr
     buildable: False
   openmpi:
-    version: [2.1.2]
+    version: [3.1.3]
+    variants: fabrics=verbs,ucx,libfabric
```
ax3l: maybe also `schedulers=alps,lsf,tm,slurm,sge,loadleveler`

ax3l: keeping like this for now.

PR #2848, opened 2019-01-04
Calorimeter: Fix unitSI

ax3l:
The unitSI in the particle calorimeter plugin should scale the bins (counts). This is not a scaling to energy but just a normalization during data reduction (and still a count).

Energy and opening angles are on the axes of the recorded mesh.


ax3l:
@n01r can you double-check this. I might have introduced an error here if what you said yesterday is right.

ax3l:
From what I see in the functor, it does indeed a `energy * normedWeighting` in the bins, which is not really great and we should do a count... I will revert this PR and if you like (@n01r) you can update it as breaking change for 0.5.0.

PR #2854, opened 2019-01-09
[WIP] Compare reader visualizer

codingS3b:
This is a continuation of #2836 trying to add a comparison reader and visualizer to the python postprocessing.

Unfortunately there are still many things unclear to me and adapting the code is not as straightforward as I first expected.
There are lots of TODO's in the code and questions we need to clarify.

1. What kind of data exactly should the reader provide?
2. Should the run directory be a path to a scan or to a single simulation (since in the code the json file of the whole scan is used but also of a single simulation)
3. often the `EnergyHistogramData` reader is used for obtaining data. Is this general enough? Should the comparison tool work only for this kind of data?
4. I have no clue about the computation of the `value` variable for the different cases of `observable` in `_get_for_iteration(...)`

### DIFF
--- a/lib/python/picongpu/plugins/data/__init__.py
+++ b/lib/python/picongpu/plugins/data/__init__.py
@@ -4,7 +4,7 @@ from .png import PNGData
 from .radiation import RadiationData
 from .sliceFieldReader import FieldSliceData
 from .emittance import EmittanceData
-from .compare.py import CompareData
+from .compare import CompareData
 
 __all__ = [
     "EnergyHistogramData",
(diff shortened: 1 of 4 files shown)


PrometheusPi:
@codingS3b since this pull request proceeds #2836 - this is the better option compared to the old version. Is this correct? (currently both pull request are still open.)

What are your plans with this pull request?

codingS3b:
I think it would be best to close it due to the problems I stated at the top of this PR

sbastrakov:
Thanks @codingS3b , closing (can always be reopened if necessary).

PR #2855, opened 2019-01-15
Fix Python param parser case

codingS3b:
Runtime parameters names are not converted to uppercase by the `param_parser.py` at the moment which is fixed in this pull request.


ax3l:
I am confused, didn't we write per mail that `tbg` options are bash variables and as such case sensitive? E.g. the ones in [LWFA](https://github.com/ComputationalRadiationPhysics/picongpu/blob/0.4.2/share/picongpu/examples/LaserWakefield/etc/picongpu/16.cfg).

codingS3b:
Oops, then I read your mail wrong sorry :(

ax3l:
No problem :)
Yes, `tbg -o ...` params are regular `bash` variables and should not be converted due to their case-sensitivity.

Issue #2777, opened 2018-10-27
`splash::DCException` in the lwfa example when turning on hdf5 output


berceanu:
Running `PIConGPU` on 8 GPUs with the example LWFA setup and increased resolution (see [8.cfg](https://gist.github.com/berceanu/d0f578e23108d458e08e81e4a3616bc9) ), results in the error (full [output](https://gist.github.com/berceanu/c097814e88a99c3038e5ebf677bf036d) ):
```
terminate called after throwing an instance of 'splash::DCException'
  what():  Exception for DCDataSet [x] write: Failed to write dataset
```
when `HDF5` raw data output is activated. The simulation finishes successfully with `HDF5` disabled.

Looks similar to the crashes that @BeyondEspresso was reporting in https://github.com/ComputationalRadiationPhysics/libSplash/issues/160.

berceanu:
### Steps to reproduce:

After following the [spack install](https://picongpu.readthedocs.io/en/0.4.0/install/instructions/spack.html), I do:
```
export SPLASH_VERBOSE=99
pic-create $PIC_EXAMPLES/LaserWakefield $HOME/picInputs/h5LWFA
pic-build -b "cuda:60"
```
then increase resolution and activate hdf5 output in `8.cfg` (see [8.cfg.patch](https://gist.github.com/berceanu/a4cfc3240c1548e84949b58236908818) ) and submit with

`tbg -s bash -c etc/picongpu/8.cfg -t etc/picongpu/bash/mpiexec.tpl $SCRATCH/runs/lwfa_wh5 &> wh5out.txt`

Here is the full [PIConGPU output](https://gist.githubusercontent.com/berceanu/78b041ca7fe57ff0f84aea4d07be1603/raw/72c46f92802152353733bd6fc6e7a4db4e8aef6e/lshw.out) with verbose libSplash.

The simulation crashes in the middle of writing `$SCRATCH/runs/lwfa_wh5/simOutput/h5/simData_600.h5`. The previous `.h5` files are succesfully written, but this one is only about half the size. 

Turning off `hdf5` output in `8.cfg` makes the simulation complete successfully.


### Additional info:

[...]

PR #2857, opened 2019-01-21
OpenMPI: Use ROMIO for IO

### LINKED ISSUE #2841: HDF5: Field "Kinks" with FreeFormula
I am seeing density-"kinks"/dips at multiples of about 1024 cells in one of my simulations:

density_kink

Looks like either an initialization or HDF5 output bug to me, since I can later on see the same dip not only in density but also in electric fields and they stay at the same y over time. Or something with our field strides...

The same output dip is seen for multiples of this position at later times when particles move into the simulation box.

Cells in y: 14720
GPUs in y: 4
GPUs in x: 4
y-cells per GPU: 3680 (<- not at a GPU boundary)

Simulation: 2D
Supercell: 16x16

Particles per cell: 20
Init in cell: random
Shape: PCS

## Software / System

PIConGPU 0.4.1 on Hypnos (K80) and Hemera (P100).

Files: (per mail)

ax3l:
OpenMPI's default for its IO backend is OMPIO, [starting with 2.x](https://www.open-mpi.org/faq/?category=ompio).

Unfortunately, that backend contains severe bugs leading to data corruption and sporadic crashes (as of the latest releases 3.1.3 and 4.0.0). This is most visible with our parallel HDF5 plugin, but ADIOS is potentially affected as well. Please see https://github.com/open-mpi/ompi/issues/6285 for details

For all system templates that rely on OpenMPI (and its derivatives, such as BullMPI), disable the "OMPIO" default IO backend and fallback to the existing ROMIO backend for MPI-I/O.

Other MPI implementations such as MPICH, and [MPICH-based flavors](https://www.mpich.org/about/collaborators/) such as IntelMPI, use ROMIO by default (they develop ROMIO) and are not affected.

Close #2841 #2777

### DIFF
--- a/etc/picongpu/bash/mpiexec.tpl
+++ b/etc/picongpu/bash/mpiexec.tpl
@@ -44,6 +44,11 @@ umask 0027
 mkdir simOutput 2> /dev/null
 cd simOutput
 
+# The OMPIO backend in OpenMPI up to 3.1.3 and 4.0.0 is broken, use the
+# fallback ROMIO backend instead.
+#   see bug https://github.com/open-mpi/ompi/issues/6285
+export OMPI_MCA_io=^ompio
+
 # test if cuda_memtest binary is available
 if [ -f !TBG_dstPath/input/bin/cuda_memtest ] ; then
   mpiexec -am !TBG_dstPath/tbg/openib.conf --mca mpi_leave_pinned 0 -npernode !TBG_gpusPerNode -n !TBG_tasks !TBG_dstPath/input/bin/cuda_memtest.sh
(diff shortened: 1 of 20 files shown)


PrometheusPi:
@ax3l Thanks you for this quick fix. For the errors so far, do they occur while writing or while loading (or both)?

I will add this option to all future runs on `taurus` right now. @BeyondEspresso and @steindev you should do the same for your simulations.

ax3l:
There are at least two bugs, which can occur independently. I experienced both during writes with HDF5 on Hemera, also @berceanu reported reproducible crashes on his system.

The first bug (the crash) is currently being fixed upstream and could also occur during parallel reads (so in our case, checkpoint restarts) as we found out. The second one (partial data corruption) is still being investigated and at least happens during parallel writes.

PR #2859, opened 2019-01-21
Core Particles: C++11 Using for Typedef

ax3l:
replace `typedef` in examples with modern C++11 `using`.
Commited as tools user.

### DIFF
--- a/include/picongpu/algorithms/FieldToParticleInterpolation.hpp
+++ b/include/picongpu/algorithms/FieldToParticleInterpolation.hpp
@@ -48,8 +48,8 @@ struct FieldToParticleInterpolation
 
     static constexpr int lowerMargin = supp / 2 ;
     static constexpr int upperMargin = (supp + 1) / 2;
-    typedef typename pmacc::math::CT::make_Int<simDim,lowerMargin>::type LowerMargin;
-    typedef typename pmacc::math::CT::make_Int<simDim,upperMargin>::type UpperMargin;
+    using LowerMargin = typename pmacc::math::CT::make_Int<simDim,lowerMargin>::type;
+    using UpperMargin = typename pmacc::math::CT::make_Int<simDim,upperMargin>::type;
 
     PMACC_CASSERT_MSG(
         __FieldToParticleInterpolation_supercell_is_too_small_for_stencil,
@@ -81,7 +81,7 @@ struct FieldToParticleInterpolation
          * The following calls seperate the vector interpolation into
          * independent scalar interpolations.
          */
-        typedef typename pmacc::math::CT::make_Int<simDim,supp>::type Supports;
+        using Supports = typename pmacc::math::CT::make_Int<simDim,supp>::type;
 
         typename Cursor::ValueType result;
         for(uint32_t i = 0; i < Cursor::ValueType::dim; i++)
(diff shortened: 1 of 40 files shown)


sbastrakov:
Just as a possible improvement: some of these type names are not dependent, so `'typename` is redundant.

ax3l:
Pls feel free to inline-annotate where that is the case.

include/picongpu/algorithms/FieldToParticleInterpolation.hpp
```diff
@@ -48,8 +48,8 @@ struct FieldToParticleInterpolation
 
     static constexpr int lowerMargin = supp / 2 ;
     static constexpr int upperMargin = (supp + 1) / 2;
-    typedef typename pmacc::math::CT::make_Int<simDim,lowerMargin>::type LowerMargin;
-    typedef typename pmacc::math::CT::make_Int<simDim,upperMargin>::type UpperMargin;
+    using LowerMargin = typename pmacc::math::CT::make_Int<simDim,lowerMargin>::type;
+    using UpperMargin = typename pmacc::math::CT::make_Int<simDim,upperMargin>::type;
```
sbastrakov: I think these are not dependent, also in other cases with `make_Int<simDim, ...>`  (note: in some places it is `make_Int<T_dim, ...>`, which is ofc dependent).

include/picongpu/particles/bremsstrahlung/Bremsstrahlung.hpp
```diff
@@ -100,9 +100,9 @@ struct Bremsstrahlung
     PMACC_ALIGN(photonMom, float3_X);
 
     /* random number generator */
-    typedef pmacc::random::RNGProvider<simDim, random::Generator> RNGFactory;
-    typedef pmacc::random::distributions::Uniform<float_X> Distribution;
-    typedef typename RNGFactory::GetRandomType<Distribution>::type RandomGen;
+    using RNGFactory = pmacc::random::RNGProvider<simDim, random::Generator>;
+    using Distribution = pmacc::random::distributions::Uniform<float_X>;
+    using RandomGen = typename RNGFactory::GetRandomType<Distribution>::type;
```
sbastrakov: This looks not dependent (same usage also in other files).

sbastrakov:
So upon futher looking, I guess my initial evaluation is not accurate, there are just two cases where `typename` is probably redundant (but they are copied several times each).

include/picongpu/particles/pusher/particlePusherReducedLandauLifshitz.hpp
```diff
@@ -81,7 +81,7 @@ struct Push
     const float_X deltaT = DELTA_T;
     const uint32_t dimMomentum = GetNComponents<TypeMomentum>::value;
     // the transver data type adjust to 3D3V, 2D3V, 2D2V, ...
-    typedef pmacc::math::Vector< picongpu::float_X, simDim + dimMomentum > VariableType;
+    usign VariableType = pmacc::math::Vector< picongpu::float_X, simDim + dimMomentum >;
```
psychocoderHPC: type: must be `using ...`

psychocoderHPC:
@sbastrakov You are right it could be that a few `typename`'s can be removed. The problem with that is that some compiler are relaxed with that and some not. IMO it is better to use `typename` since we currently not cover all compiler within the CI.

sbastrakov:
@psychocoderHPC not sure if I understand your point. Do you mean that some compilers treat non dependent names as if they are dependent and, thus, somehow have more strict requirements for `typename` than the standard? Anyways I would agree this is not a big deal and we might simply leave as is.

ax3l:
Yes, e.g. NVCC sometimes behaves weird. I would also not relax the current code in a purely textual change without further validation.
Therefore, I would leave it as is for now, can be done and validated independently.

psychocoderHPC:
@sbastrakov I mean more it is hard to review such changes because sometimes the compiler allows non-dependant typdef's without using. Therefore a simple rule is use `typename` always in combination with `x::type` where you query an inner type of a class.

PR #2862, opened 2019-01-24
[WIP] Update hemera-hzdr/gpu.tpl

steindev:
During a recent maintenance on the hemera cluster at HZDR Hyperthreading was disabled. Therefore, we need to allocate less cores per gpu in order to comply with resource handling by the SLURM scheduler.

### DIFF
--- a/etc/picongpu/hemera-hzdr/gpu.tpl
+++ b/etc/picongpu/hemera-hzdr/gpu.tpl
@@ -63,7 +63,7 @@
 
 # number of cores to block per GPU - we got 6 cpus per gpu
 #   and we will be accounted 6 CPUs per GPU anyway
-.TBG_coresPerGPU=6
+.TBG_coresPerGPU=3
 
 # We only start 1 MPI task per GPU
 .TBG_mpiTasksPerNode="$(( TBG_gpusPerNode * 1 ))"


etc/picongpu/hemera-hzdr/gpu.tpl
```diff
@@ -63,7 +63,7 @@
 
 # number of cores to block per GPU - we got 6 cpus per gpu
```
sbastrakov: Probably makes sense to update the comment as well. However, I'm not sure to what, as I don't understand why 3 is the correct value (but it really is).

psychocoderHPC: change it to: `number of non hyper thread cores to block per GPU - we got 3 cores per gpu`

ATTENTION: before it was `6 cpus` but it must be `cores`

psychocoderHPC:
Please also update the `defq.tpl` for the cpu partition.

etc/picongpu/hemera-hzdr/gpu.tpl
sbastrakov: That is exactly what I do not understand: each node has 2x 12-core CPUs, so there are 6 physical ("non hyperthreaded") cores per GPU on the machine. Ofc now it is more important to just fix and not make it so I understand the fix.

etc/picongpu/hemera-hzdr/gpu.tpl
```diff
@@ -63,7 +63,7 @@
...
 #   and we will be accounted 6 CPUs per GPU anyway
-.TBG_coresPerGPU=6
```
psychocoderHPC: The system contains `2x Intel Xeon Gold 6136, 12 Cores/CPU` therefore 6 is correct.

sbastrakov: Yes, that was exactly my point. But it just does not work like that somehow.

psychocoderHPC:
@sbastrakov @steindev  and me checked the behavior on the cluster. The cluster is somehow misconfigured. I opened a ticket and we hope that it is solved soon. 
There should be no changes needed if the cluster work correct.

psychocoderHPC:
The admin moved the solution time to 08.02.2019 so I would say we should update the comment in the tpl file and merge it for now until the system is fixed.

psychocoderHPC:
OK looks like the problem is solved. I tested it with a interactive shell and I am now able to query all cores and all gpus.
I will run tomorrow a test to be sure that all is working again with our templates.

psychocoderHPC:
This issue is now solved since the vendor fixed SLURM

PR #2863, opened 2019-01-25
[WIP] Draft for PML with Yee solver

sbastrakov:
Since it has all been stuck for super-long time, I've quickly put my considerations of how to *actually* add it to the code.

A more detailed description of the general idea, types of PML, what performance is expected based on my small test app, to be added here.

### DIFF
--- a/include/picongpu/fields/MaxwellSolver/Yee/YeePML.hpp
+++ b/include/picongpu/fields/MaxwellSolver/Yee/YeePML.hpp
@@ -1,0 +1,192 @@
+/* Copyright 2013-2019 Axel Huebl, Heiko Burau, Rene Widera, Benjamin Worpitz
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/simulation_defines.hpp"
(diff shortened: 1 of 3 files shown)


include/picongpu/fields/MaxwellSolver/Yee/YeePML.hpp
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+    public:
+
+        YeePML(MappingDesc cellDescription) : Yee(cellDescription)
+        {
+            // for now reuse parameters of the old absorber
+            // WARNING: this is only correct for a single process,
+            // TODO: for the general case needs to take into account domain decomposition
+            // so that thickness is 0 for "internal" boundaries between domains
```
psychocoderHPC: note: The thickness must be updated within each time step if the moving window is enabled. The reason is that our devices have no fixed position within the simulation volume (https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-domain-definitions#moving-window)

include/picongpu/fields/MaxwellSolver/Yee/YeePML.kernel
```diff
@@ -0,0 +1,252 @@
...
+            collective(
+                acc,
+                assign,
+                cachedE,
+                fieldEBlock
+            );
+
+            __syncthreads();
+
+            constexpr float_X dt = DELTA_T;
```
psychocoderHPC: Do we need to change this to `constexpr float_X dt = DELTA_T / 2;` since we advance the B filed only for a half time step.

sbastrakov:
Literature:

- [Original paper](https://www.sciencedirect.com/science/article/pii/S0021999184711594), this is now known as split-field PML and is most widely used, though not most efficient
- [Uniaxial PML](https://ieeexplore.ieee.org/document/477075/), a differently formulated and sometimes more efficient scheme
- [Convolutional PML](https://onlinelibrary.wiley.com/doi/abs/10.1002/1098-2760%2820001205%2927%3A5%3C334%3A%3AAID-MOP14%3E3.0.CO%3B2-A), this is a generalized formulation, from which one could derive both split-field and uniaxial versions; I believe this is considered state-of-the-art approach
- I find [this book](https://cds.cern.ch/record/1698084/files/1580538320_TOC.pdf) very helpful generally in terms of finite-different field solvers. The first edition has a short note about split-field PML, the second has all three I believe (and the third I have no access to, unfortunately).

cc @steindev

sbastrakov:
#2950 is up-to-date, will close this one.

PR #2866, opened 2019-01-27
Fix pyflakes str/bytes/int compares

ax3l:
Fix pyflakes:
```
use ==/!= to compare str, bytes, and int literals
```


ax3l:
@PrometheusPi this is currently breaking the CI. Just a little fix.

PR #2868, opened 2019-01-28
Docs: Memory per Device Spelling

ax3l:
Replace "GPU"s with device naming in our memory estimator.

### DIFF
--- a/docs/source/usage/workflows/memoryPerDevice.py
+++ b/docs/source/usage/workflows/memoryPerDevice.py
@@ -16,12 +16,12 @@ from picongpu.utils import MemoryCalculator
 This file contains a usage example of the ``MemoryCalculator`` class
 for our :ref:`FoilLCT example <usage-examples-foilLCT>` and its ``4.cfg``.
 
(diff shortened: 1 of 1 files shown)


docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -16,10 +16,10 @@
 This file contains a usage example of the ``MemoryCalculator`` class
 for our :ref:`FoilLCT example <usage-examples-foilLCT>` and its ``4.cfg``.
 
-It is an estimate for how much memory is used per GPU if the whole
+It is an estimate for how much memory is used per device if the whole
 target would be fully ionized but does not move much. Of course the real
 memory usage depends on the case and the dynamics inside the simulation.
-We calculate the memory just one GPU out of the whole group that simulates
+We calculate the memory just one device out of the whole group that simulates
```
PrometheusPi: I think there is a `of` missing here:
```
We calculate the memory of just one device out of the whole group that simulates ...
```

PR #2869, opened 2019-01-28
[WIP] System: JURECA

sbastrakov:
Add new system documentation, profiles and tbg template for the JURECA cluster at JSC. So far only the batch queue is supported, others to be added soon.

So far managed to build and run without ADIOS.
Will update the PR as other queues are added.

Big thanks to @psychocoderHPC !


sbastrakov:
Some other notes during setting this up, might be useful to change in other profiles (ofc not as this PR):

- Since different systems use different variables for scratch storage path, maybe provide an environment variable `SCRATCH` for that (in case the system does not provide it), so that a user could always use that for simulation output.
- Duplicated `#SBATCH --workdir=!TBG_dstPath`. [Example](https://github.com/ComputationalRadiationPhysics/picongpu/blob/0672932a795b16c1b2ce61e7ec749bcfc4a89495/etc/picongpu/davide-cineca/gpu.tpl#L39), also in other `.tpl` files.
- Typo in comments for getNode and getDevice. [Example (`to` instead of `two`)](https://github.com/ComputationalRadiationPhysics/picongpu/blob/0672932a795b16c1b2ce61e7ec749bcfc4a89495/etc/picongpu/davide-cineca/gpu_picongpu.profile.example#L77), also in other profiles.

ax3l:
Thanks! Please also document the new profiles in `docs/source/install/profile.rst` for the manual :)

sbastrakov:
Added docs, also fixed some issues in my original commit.

docs/source/install/profile.rst
```diff
@@ -207,3 +207,20 @@ Queue: dvd_usr_prod (Nvidia P100 GPUs)
...
+
+**System overview:** `link <http://www.fz-juelich.de/ias/jsc/EN/Expertise/Supercomputers/JURECA/JURECA_node.html>`_
+
+**User guide:** `link <http://www.fz-juelich.de/ias/jsc/EN/Expertise/Supercomputers/JURECA/UserInfo/UserInfo_node.html>`_
+
+**Production directory:** ``$SCRATCH`` (`link <http://www.fz-juelich.de/SharedDocs/FAQs/IAS/JSC/EN/JUST/FAQ_00_File_systems.html?nn=1297148>`_)
+
+For this profile to work, you need to download the :ref:`PIConGPU source code <install-dependencies-picongpu>` manually.
+
+Queue: batch (Intel Xeon E5-2680 v3 CPUs)
```
psychocoderHPC: Please add `2 x Intel ...`

sbastrakov: Sure. It's actually also really inconsistent between different machines in this document. I think the best formulation is [this one]( https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/docs/source/install/profile.rst#queue-defq-2x-intel-xeon-gold-6148-20-cores--20-hyperthreadscpu), also covering SMT. I guess I'll follow that and put `Queue: batch (2 x Intel Xeon E5-2680 v3 CPUs, 12 Cores + 12 HyperThreads per CPU)`.

psychocoderHPC: offline discussed: correct `... per node` is not required.

sbastrakov:
GPU added (ugh, their infrastructure seems not good, but finally it works), KNL should be done soon.

etc/picongpu/jureca-jsc/batch.tpl
```diff
@@ -0,0 +1,95 @@
...
+# required devices per node for the current job
+.TBG_devicesPerNode=`if [ $TBG_tasks -gt $TBG_numHostedDevicesPerNode ] ; then echo $TBG_numHostedDevicesPerNode; else echo $TBG_tasks; fi`
+
+# host memory per device
+.TBG_memPerCPU="$((126000 / $TBG_devicesPerNode))"
+# host memory per node
+.TBG_memPerNode="$((TBG_memPerCPU * TBG_devicesPerNode))"
+
+# number of cores to block per device - we got 24 cores (including hyper-threads) per device
+.TBG_coresPerCPU=24
```
psychocoderHPC: we can use this value for `export OMP_NUM_THREADS`

etc/picongpu/jureca-jsc/batch.tpl
```diff
@@ -0,0 +1,95 @@
...
+
+#set user rights to u=rwx;g=r-x;o=---
+umask 0027
+
+mkdir simOutput 2> /dev/null
+cd simOutput
+ln -s ../stdout output
+
+if [ $? -eq 0 ] ; then
+  # Run PIConGPU
```
psychocoderHPC: offline discussed: `export OMP_NUM_THREADS` is required

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
```diff
@@ -0,0 +1,115 @@
...
+
+export PICSRC=$HOME/src/picongpu
+export PIC_EXAMPLES=$PICSRC/share/picongpu/examples
+export PIC_BACKEND="cuda:37"
+
+export PATH=$PATH:$PICSRC
+export PATH=$PATH:$PICSRC/bin
+export PATH=$PATH:$PICSRC/src/tools/bin
+
+export CC=$(which gcc)
```
psychocoderHPC: Is this required? Since we are loading only the gcc toolchain.

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
```diff
@@ -0,0 +1,115 @@
...
+export PATH=$ADIOS_ROOT/bin:$PATH
+export CMAKE_PREFIX_PATH=$ADIOS_ROOT:$CMAKE_PREFIX_PATH
+
+# Environment #################################################################
+#
+#export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$BOOST_LIB
+
+export PICSRC=$HOME/src/picongpu
+export PIC_EXAMPLES=$PICSRC/share/picongpu/examples
+export PIC_BACKEND="cuda:37"
```
psychocoderHPC: could you please add a comment `# architecture NVIDIA k80` or similar because most devs/user do not have all the architecture in there mind.

etc/picongpu/jureca-jsc/batch.tpl
sbastrakov: Sure, but in this file it is only used once. Maybe it is better to define smth in the profile, since we have to use it [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/4e8a90ee247486e32ea4bc354669cd04391a6f97/etc/picongpu/jureca-jsc/batch_picongpu.profile.example#L89) as well?

psychocoderHPC: I would like to keep it in the tpl to allow also the usage with modules loaded by hand. But we can remove this definition and add only a comment above `export OMP_NUM_THREADS`.

You pointed to a very interesting point: **maybe** it would be much nicer for tbg to have a system description and a tpl which is only describing the batch system handling.

sbastrakov: Guess the latter is worth discussing, separate from this PR. Added to the [list](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2869#issuecomment-458311754).

sbastrakov:
Ready for review. Now code is running on all 3 partitions. On KNL settings are almost surely suboptimal, but tuning would probably be done a bit later (and might be time-consuming) and so would be subject for another PR.

etc/picongpu/jureca-jsc/batch.tpl
sbastrakov: Thanks, fixed.

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
sbastrakov: Thanks, fixed.

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
sbastrakov: Done.

ax3l:
Thanks! Can you please fix the CI reported issues?
```
SUMMARY
-------
Run the following command(s) on the above files to remove your
end-of-line (EOL) white spaces:
sed -i 's/[[:blank:]]\+$//' ./etc/picongpu/jureca-jsc/batch.tpl
The command "test/hasEOLwhiteSpace" exited with 1.
```

docs/source/install/profile.rst
```diff
@@ -207,3 +207,32 @@ Queue: dvd_usr_prod (Nvidia P100 GPUs)
...
+**System overview:** `link <http://www.fz-juelich.de/ias/jsc/EN/Expertise/Supercomputers/JURECA/JURECA_node.html>`_
+
+**User guide:** `link <http://www.fz-juelich.de/ias/jsc/EN/Expertise/Supercomputers/JURECA/UserInfo/UserInfo_node.html>`_
+
+**Production directory:** ``$SCRATCH`` (`link <http://www.fz-juelich.de/SharedDocs/FAQs/IAS/JSC/EN/JUST/FAQ_00_File_systems.html?nn=1297148>`_)
+
+For this profile to work, you need to download the :ref:`PIConGPU source code <install-dependencies-picongpu>` manually.
+
+Queue: batch (2 x Intel Xeon E5-2680 v3 CPUs, 12 Cores + 12 Hyperthreads/CPU)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ax3l: in `.rst`, please align the headline markers with the headline length

docs/source/install/profile.rst
```diff
@@ -207,3 +207,32 @@ Queue: dvd_usr_prod (Nvidia P100 GPUs)
 
 .. literalinclude:: profiles/davide-cineca/gpu_picongpu.profile.example
    :language: bash
+
+JURECA (JSC)
+--------------------
```
ax3l: in `.rst`, please align the headline markers with the headline length

docs/source/install/profile.rst
```diff
@@ -207,3 +207,32 @@ Queue: dvd_usr_prod (Nvidia P100 GPUs)
...
+For this profile to work, you need to download the :ref:`PIConGPU source code <install-dependencies-picongpu>` manually.
+
+Queue: batch (2 x Intel Xeon E5-2680 v3 CPUs, 12 Cores + 12 Hyperthreads/CPU)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. literalinclude:: profiles/jureca-jsc/batch_picongpu.profile.example
+   :language: bash
+
+Queue: gpus (2 x Nvidia Tesla K80 GPUs)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ax3l: in `.rst`, please align the headline markers with the headline length

docs/source/install/profile.rst
```diff
@@ -207,3 +207,32 @@ Queue: dvd_usr_prod (Nvidia P100 GPUs)
...
+
+.. literalinclude:: profiles/jureca-jsc/gpus_picongpu.profile.example
+   :language: bash
+
+  Queue: booster (Intel Xeon Phi 7250-F, 68 cores + Hyperthreads)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ax3l: in `.rst`, please align the headline markers with the headline length

docs/source/install/profile.rst
```diff
@@ -207,3 +207,32 @@ Queue: dvd_usr_prod (Nvidia P100 GPUs)
...
+.. literalinclude:: profiles/jureca-jsc/batch_picongpu.profile.example
+   :language: bash
+
+Queue: gpus (2 x Nvidia Tesla K80 GPUs)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. literalinclude:: profiles/jureca-jsc/gpus_picongpu.profile.example
+   :language: bash
+
+  Queue: booster (Intel Xeon Phi 7250-F, 68 cores + Hyperthreads)
```
ax3l: please remove the spaces at the beginning of the title

docs/source/install/profile.rst
```diff
@@ -207,3 +207,32 @@ Queue: dvd_usr_prod (Nvidia P100 GPUs)
...
+JURECA (JSC)
+--------------------
+
+**System overview:** `link <http://www.fz-juelich.de/ias/jsc/EN/Expertise/Supercomputers/JURECA/JURECA_node.html>`_
+
+**User guide:** `link <http://www.fz-juelich.de/ias/jsc/EN/Expertise/Supercomputers/JURECA/UserInfo/UserInfo_node.html>`_
+
+**Production directory:** ``$SCRATCH`` (`link <http://www.fz-juelich.de/SharedDocs/FAQs/IAS/JSC/EN/JUST/FAQ_00_File_systems.html?nn=1297148>`_)
+
+For this profile to work, you need to download the :ref:`PIConGPU source code <install-dependencies-picongpu>` manually.
```
ax3l: You also need PNGwriter, libSplash and ADIOS below. See other profiles above on how to note this here :)

Also, since it's multiple queues please write "For these profiles to work, ..."

etc/picongpu/jureca-jsc/batch_picongpu.profile.example
```diff
@@ -0,0 +1,109 @@
...
+module purge
+module load Intel/2019.0.117-GCC-7.3.0
+module load CMake/3.13.0
+module load IntelMPI/2018.4.274
+module load Python/3.6.6
+module load Boost/1.68.0-Python-3.6.6
+
+# Other Software ##############################################################
+#
+module load zlib/.1.2.11
```
ax3l: typo in version? starts with `.`? (same in `libpng`)

etc/picongpu/jureca-jsc/batch.tpl
```diff
@@ -0,0 +1,93 @@
...
+.TBG_mailAddress=${MY_MAIL:-"someone@example.com"}
+.TBG_author=${MY_NAME:+--author \"${MY_NAME}\"}
+.TBG_nameProject=${proj:-""}
+.TBG_profile=${PIC_PROFILE:-"~/picongpu.profile"}
+
+# number of available/hosted devices per node in the system
+.TBG_numHostedDevicesPerNode=2
+
+# required devices per node for the current job
+.TBG_devicesPerNode=`if [ $TBG_tasks -gt $TBG_numHostedDevicesPerNode ] ; then echo $TBG_numHostedDevicesPerNode; else echo $TBG_tasks; fi`
```
ax3l: please prefer `$()` over "`" for new code
```suggestion
.TBG_devicesPerNode=$(if [ $TBG_tasks -gt $TBG_numHostedDevicesPerNode ] ; then echo $TBG_numHostedDevicesPerNode; else echo $TBG_tasks; fi)
```

etc/picongpu/jureca-jsc/booster.tpl
```diff
@@ -0,0 +1,93 @@
...
+.TBG_mailAddress=${MY_MAIL:-"someone@example.com"}
+.TBG_author=${MY_NAME:+--author \"${MY_NAME}\"}
+.TBG_nameProject=${proj:-""}
+.TBG_profile=${PIC_PROFILE:-"~/picongpu.profile"}
+
+# KNL is used in quadrant mode, treat each quadrant as a device
+.TBG_numHostedDevicesPerNode=4
+
+# required devices per node for the current job
+.TBG_devicesPerNode=`if [ $TBG_tasks -gt $TBG_numHostedDevicesPerNode ] ; then echo $TBG_numHostedDevicesPerNode; else echo $TBG_tasks; fi`
```
ax3l: Please prefer `$()` for new code
```suggestion
.TBG_devicesPerNode=$(if [ $TBG_tasks -gt $TBG_numHostedDevicesPerNode ] ; then echo $TBG_numHostedDevicesPerNode; else echo $TBG_tasks; fi)
```

etc/picongpu/jureca-jsc/batch_picongpu.profile.example
```diff
@@ -0,0 +1,109 @@
...
+#   - automatically add your name and contact to output file meta data
+#   - send me a mail on batch system jobs: NONE, BEGIN, END, FAIL, REQUEUE, ALL,
+#     TIME_LIMIT, TIME_LIMIT_90, TIME_LIMIT_80 and/or TIME_LIMIT_50
+export MY_MAILNOTIFY="NONE"
+export MY_MAIL="someone@example.com"
+export MY_NAME="$(whoami) <$MY_MAIL>"
+
+# Project Information ######################################## (edit this line)
+#   - project account for computing time
+export proj=$(groups | awk '{print $5}')
```
ax3l: maybe there is a variable on this system already which we can use? Just in case, that would be more robust if they provide one. just had to do this on systems that don't export one. (same in other profiles below)

etc/picongpu/jureca-jsc/booster.tpl
```diff
@@ -0,0 +1,93 @@
...
+
+# host memory per device
+.TBG_memPerCPU="$((94000 / $TBG_devicesPerNode))"
+# host memory per node
+.TBG_memPerNode="$((TBG_memPerCPU * TBG_devicesPerNode))"
+
+# We only start 1 MPI task per device
+.TBG_mpiTasksPerNode="$(( TBG_devicesPerNode * 1 ))"
```
ax3l: let's spare the `* 1`, we have a whole comment in the line above

etc/picongpu/jureca-jsc/gpus.tpl
```diff
@@ -0,0 +1,102 @@
...
+.TBG_mailAddress=${MY_MAIL:-"someone@example.com"}
+.TBG_author=${MY_NAME:+--author \"${MY_NAME}\"}
+.TBG_nameProject=${proj:-""}
+.TBG_profile=${PIC_PROFILE:-"~/picongpu.profile"}
+
+# number of available/hosted devices per node in the system
+.TBG_numHostedDevicesPerNode=4
+
+# required GPUs per node for the current job
+.TBG_devicesPerNode=`if [ $TBG_tasks -gt $TBG_numHostedDevicesPerNode ] ; then echo $TBG_numHostedDevicesPerNode; else echo $TBG_tasks; fi`
```
ax3l: Please prefer `$()` for new code
```suggestion
.TBG_devicesPerNode=$(if [ $TBG_tasks -gt $TBG_numHostedDevicesPerNode ] ; then echo $TBG_numHostedDevicesPerNode; else echo $TBG_tasks; fi)
```

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
```diff
@@ -0,0 +1,112 @@
...
+#   - examples: "nano", "vim", "emacs -nw", "vi" or without terminal: "gedit"
+#export EDITOR="nano"
+
+# General modules #############################################################
+#
+module purge
+module load GCC/7.3.0
+module load CUDA/9.2.88
+module load CMake/3.13.0
+module load MVAPICH2/2.3-GDR
```
ax3l: is this MPI build with CUDA awareness? we have to make sure, otherwise we have to control MPI pinning to avoid steels between network and cuda pinned pages.

sbastrakov:
@ax3l sorry, fixed.

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
```diff
@@ -0,0 +1,112 @@
...
+    if [ -z "$1" ] ; then
+        numNodes=1
+    else
+        numNodes=$1
+    fi
+    if [ $numNodes -gt 8 ] ; then
+        echo "The maximal number of interactive nodes is 8." 1>&2
+        return 1
+    fi
+    echo "Hint: please use 'srun --cpu_bind=sockets <COMMAND>' for launching multiple processes in the interactive mode"
```
ax3l: very good!

etc/picongpu/jureca-jsc/batch_picongpu.profile.example
sbastrakov: No, these are hidden (dunno why) and so accessible only this way.

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
sbastrakov: The docs and how the module is exposed say it is CUDA-aware. I did not do any specific checks.

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
sbastrakov: I like it as well, credit to @psychocoderHPC

docs/source/install/profile.rst
sbastrakov: Good point. Just to clarify a little bit: currently the dependencies installed in a way that for users of my project it is ready out-of-the box. However, ofc generally you are correct and additional dependencies are required. Will fix.

etc/picongpu/jureca-jsc/batch_picongpu.profile.example
sbastrakov: I've now looked again for such a variable, but did not find it. However, I've now found their script that sets some other useful env variables (like `$SCRATCH` that we've set by hand), will include that.

sbastrakov:
@ax3l thanks for review, I implemented all the suggestions. For a couple I clicked the auto-apply on github (a cool feature, btw) and it automatically put you as author (which is completely reasonable, just explaining).

PR #2870, opened 2019-02-01
[WIP] Docs: Add ionization prediction plot

n01r:
This PR adds a python script that plots a charge state population prediction for Neon gas that is irradiated by a laser pulse of 30fs duration.
The maximum a0 is 10 and the pulse is only modeled as a Gaussian envelope.
A documentation section has been added as well.

- [x] relocate the main parts of the script to `lib/python/picongpu` so that users can easily benefit from it
- [x] rearrange the plot so that it is easily understandable
- [x] format the python code according to PEP8

At the moment the plot it creates in the docs looks like this:

- *top panel*: ionization rates for Neon in the ADK model calculated with the intensity shape given in the lower panel
- *middle panel*: relative charge state population of Neon based on the ADK model
- *lower panel*: temporal intensity distribution in black and BSI model predicted charge state population overlay in colored regions

field_ionization_charge_state_prediction

### DIFF
--- a/docs/source/models/field_ionization.rst
+++ b/docs/source/models/field_ionization.rst
@@ -73,7 +73,7 @@ Tunneling Ionization
 Tunneling ionization describes the process during which an initially bound electron quantum-mechanically tunnels through a potential barrier of finite height.
 
 Keldysh
-"""""""
+^^^^^^^
 
 .. math::
 
@@ -92,7 +92,7 @@ The Keldysh ionization rate has been implemented according to the equation (9) i
 
 
 Ammosov-Delone-Krainov (ADK)
-""""""""""""""""""""""""""""
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. math::
    :nowrap:
@@ -120,6 +120,14 @@ When we account for orbital structure in shielding of the ion charge :math:`Z` a
 One would expect much earlier ionization of Hydrogen due to lower ionization energy. The following image shows how this can be explained by the shape of the ion potential that is assumed in this model.
 
 .. plot:: models/field_ionization_effective_potentials.py
+
+Predicting Charge State Distributions
+-------------------------------------
+
+Especially for underdense targets it is possible to already give an estimate for how the laser pulse ionizes a certain target.
+Starting from an initially unionized state, calculating ionization rates for each charge state for a given electric field a Markovian approach of transition matrices yields the charge state population for each time.
+
+.. plot:: models/field_ionization_charge_state_prediction.py
 
 References
 ----------
--- a/docs/source/models/field_ionization.rst
+++ b/docs/source/models/field_ionization.rst
@@ -47,7 +47,7 @@ Overview: Implemented Models
     |                     | * ``ADKCircPol``            | * [DeloneKrainov]_        |
     +---------------------+-----------------------------+---------------------------+
     | Barrier Suppression | * ``BSI``                   | * [MulserBauer2010]_      |
-    |                     | * ``BSIEffectiveZ``         | * [ClementiRaimondi1963]_ |
+    |                     | * ``BSIEffectiveZ`` (R&D)   | * [ClementiRaimondi1963]_ |
     |                     |                             |   [ClementiRaimondi1967]_ |
     |                     | * ``BSIStarkShifted`` (R&D) | * [BauerMulser1999]_      |
     +---------------------+-----------------------------+---------------------------+
@@ -125,9 +125,13 @@ Predicting Charge State Distributions
 -------------------------------------
 
 Especially for underdense targets, it is possible to already give an estimate for how the laser pulse ionizes a specific target.
-Starting from an initially unionized state, calculating ionization rates for each charge state for a given electric field a Markovian_ approach of transition matrices yields the charge state population for each time.
+Starting from an initially unionized state, calculating ionization rates for each charge state for a given electric field via a Markovian_ approach of transition matrices yields the charge state population for each time.
 
 .. _Markovian: https://en.wikipedia.org/wiki/Markov_chain
+
+Here, we show an example of Neon gas ionized by a Gaussian laser pulse with maximum amplitude :math:`a_0 = 10` and pulse duration (FWHM intensity) of :math:`30\,\mathrm{fs}`.
+The figure shows the ionization rates and charge state population produced by the ``ADKLinPol`` model obtained from the pulse shape in the lower panel, as well as the step-like ionization produced by the ``BSI`` model.
+
 .. plot:: models/field_ionization_charge_state_prediction.py
 
 References
--- a/docs/source/models/field_ionization.rst
+++ b/docs/source/models/field_ionization.rst
@@ -134,6 +134,35 @@ The figure shows the ionization rates and charge state population produced by th
 
 .. plot:: models/field_ionization_charge_state_prediction.py
 
+You can test the implemented ionization models yourself with the corresponding module shipped in ``picongpu/lib/python``.
+
+.. code:: python
+
+    import numpy as np
+    import scipy.constants as sc
+    from picongpu.utils import FieldIonization
+
+    # instantiate class object that contains functions for
+    #   - ionization rates
+    #   - critical field strengths (BSI models)
+    #   - laser intensity conversion
+    FI = FieldIonization()
+
+    # dictionary with atomic units
+    AU = FI.atomic_unit
+
+    # residual charge state AFTER ionization
+    Z_H = 1
+    # hydrogen ionization energy (13.6 eV) converted to atomic units
+    E_H_AU = 13.6 * sc.electron_volt / AU['energy']
+    # output: 0.50
+    print("%.2f" % (E_H_AU))
+    # barrier suppression threshold field strength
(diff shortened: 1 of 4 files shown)


n01r:
Please don't add style remarks for now since I haven't *PEP8*'ed the code just yet

docs/source/models/field_ionization_charge_state_prediction.py
```diff
@@ -0,0 +1,260 @@
...
+
+def convert_a0_to_Intensity(E_in_a0,lambda_laser=800.e-9):
+    """
+    :param E_in_a0: electric field [unit: a0]
+    :param lambda_laser: laser wavelength [unit: m]
+
+    :returns: intensity [unit: W/m²]
+    """
+
+    return (E_in_a0 / (0.85 * lambda_laser * 1e6))**2 * 1e22
```
PrometheusPi: Could you explain the magic numbers?

- `0.85`
- `1e6`
-  `1e22`

or give them named variable names.

docs/source/models/field_ionization_charge_state_prediction.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+if __name__ == "__main__":
+    """
+    On execution this file produces the image
```
PrometheusPi: comma missing:
```
On execution, this file ...
```

docs/source/models/field_ionization_charge_state_prediction.py
```diff
@@ -0,0 +1,260 @@
...
+    # atomic units
+    AU_E_eV = atomic_unit['energy']/sc.electron_volt  # eV
+    AU_F = atomic_unit['electric field']  # V/m
+    AU_I = atomic_unit['intensity']  # W/m^2
+    AU_T = atomic_unit['time']  # s
+
+    # proton number: Neon
+    Z_max = 10
+    # array of possible charge states: Neon
+    Z = np.arange(Z_max) + np.int(1)
```
PrometheusPi: Why do you have to convert the `int` `1` to a numpy integer `np.int`?
Is this needed? If yes, the reason for it is not apparent to me - please add a comment.

docs/source/models/field_ionization_charge_state_prediction.py
```diff
@@ -0,0 +1,260 @@
...
+    # array of ionization energies in eV for Neon
```
PrometheusPi: Could you add a link to NIST where one can find these values?

docs/source/models/field_ionization_charge_state_prediction.py
```diff
@@ -0,0 +1,260 @@
...
+    # convert ionization energies to atomic units
+    i_pot_AU = ionization_energies_eV / AU_E_eV
+
+    """
+    Create the electric field distribution for our example.
+    """
+    #
+    lambda_laser = 800.e-9
+    # a0 in SI
+    a0 = np.float64(2*np.pi/lambda_laser * sc.m_e*sc.c**2 / sc.e)
```
PrometheusPi: This is confusing - but I use it the same way 😨 
What you mean is the electric field strength of a_0 = 1.0.
Please add this as a comment above.

docs/source/models/field_ionization_charge_state_prediction.py
```diff
@@ -0,0 +1,260 @@
...
+            # probability to stay bound
+            trans_mat[k,k] = np.exp(-rate_matrix[k,i] * dt)
+            # probability to ionize
+            trans_mat[k+1,k] = 1. - np.exp(-rate_matrix[k,i] * dt)
+
+        # Markov step
+        charge_dist[:,i] = np.dot(charge_dist[:,i-1],trans_mat.T)
+
+    """
+    Barrier suppression field strength calculation.
```
PrometheusPi: No dot needed 😜

PrometheusPi:
@n01r Could you add a title to each subplot stating what you want to show.

1. ADK ionization rates
2. ADK relative population
3. BSI (relative) population

Furthermore a `plt.twinx()`-call would be useful in all plots to plot the intensity envelope alongside.
Thus you could plot the relevant physical quantity on the left side and the laser envelope on the right (in black).
By adding the relative BSI population axis label in plot number three as well, would make the plots consistent - even though the values are either 0.0 or 1.0. 

You could also combine plot two and three by keeping plot two as is and adding:

- BSI levels with dashed lines in the same color (no color filling)
- the laser envelope (in black) mapped to the right y-axis

ax3l:
Maybe you want to move the mid panel to the top so one can compare the BSI appearance intensity with the ADK rate gradient visually?

PrometheusPi:
@n01r what is the current status of this pull request?

n01r:
@PrometheusPi yea, sorry, I haven't taken the time to implement all reviewer comments, yet.

PrometheusPi:
@n01r - ping number 2 - what is the status?

PrometheusPi:
@n01r - ping number 3 - what is the status?

PrometheusPi:
@n01r ping number 4 - will you work on this in the future?

docs/source/models/field_ionization_charge_state_prediction.py
n01r: I wasn't actually using this, anymore.
So I deleted it.

n01r:
Alright @PrometheusPi, sorry for the wait.
I made it so that anyone (me included) can now use the python modules to check the ionization models that are implemented in PIConGPU.
I documented that as well.
The plot was rearranged as @ax3l suggested (I really just shifted the panels). However, I did not draw lines because due to the missing orbital structure information in the normal `BSI` model, it will just always be a little off from the ADK prediction. It fits a little better for s-states.
The `BSIStarkShifted` model ensures the condition, that tunneling always happens before the threshold value, a little better. But I checked that only for the Neon there. This model has other simplifications and issues, though. I should, when I have more time again, try to implement the model that @sbastrakov found and sent to me. Maybe the connection between tunneling and barrier suppression is better there and still manageable on GPU.

I also did a relative import for the script that does the plotting and it uses the modules in `picongpu/lib/python` now, which should work, no matter where the documentation is [built](file:///home/noir/Arbeit/Coding/picongpu/docs/build/html/dev/sphinx.html#build-locally).

n01r:
Checks passed. I did not press *resolve conversation* on your comments after I fixed them, so that you can see everything that was changed when you review after you come back from vacation @PrometheusPi. :)

PrometheusPi:
@n01r great work :+1: - I will review it today (and hopefully merge it)

PrometheusPi:
@n01r can we remove the `[WIP]` tag?

docs/source/models/field_ionization_charge_state_prediction.py
```diff
@@ -0,0 +1,315 @@
...
+        s=note_string,
+        fontsize=12,
+        transform=ax_rate.transAxes
+    )
+
+    # labels
+    ax_rate.set_ylabel(r"ADK: Ionization Rate $\Gamma\,\mathrm{(AU^{-1})}$")
+    ax_pop.set_ylabel(r"ADK: Relative Population (%)")
+    ax_bsi_twin.set_ylabel(r"Laser Intensity $I\,\mathrm{(W/cm^2)}$")
+    ax_bsi.set_xlabel(r"Time $t-t_\mathrm{max}\,\mathrm{(fs)}$")
```
PrometheusPi: The ticks numbers without labels confuse a bit the fact, that you have a shared axis. Just remove the ticks numbers for the upper tow plots x-axis. 

```suggestion
    ax_rate.xticks([])
    ax_pop.xticks([])
    ax_bsi.set_xlabel(r"Time $t-t_\mathrm{max}\,\mathrm{(fs)}$")
```

docs/source/models/field_ionization.rst
```diff
@@ -121,6 +121,48 @@ One would expect much earlier ionization of Hydrogen due to lower ionization ene
```
PrometheusPi: In your last sentence, you say that the plot explains why the BSI levels of Hydrogen and Carbon-II (in contrast to atomic Carbon?). But the plot alone misses an explanation. Could you add a sentence explaining why they are so close?

PrometheusPi: Furthermore, the use of Carbon-II in both ADK cases but atomic-Carbon in one of the BSI plots from `models/field_ionization_comparison_c_ii_ionization.py` confuses me here.

docs/source/models/field_ionization_charge_state_prediction.py
PrometheusPi: @n01r If you like the numbers in between - I am fine with that as well. 
(my suggestions removes not only the ticks labels, but the ticks as well - this is definitely not good. There should be better option by just removing the tick labels and keeping the ticks.)

PrometheusPi: *offline discussion with @n01r:* He preserve the numbers in between.

docs/source/models/field_ionization.rst
PrometheusPi: *offline discussion with @n01r:* 
This is not part of this pull request. This this should not be fixed here. I opened an issue  #3186 to address the problem.

PR #2873, opened 2019-02-01
Correct target thickness

pordyna:
`y1` was not really the target thickness but its rear surface position.
Previously it was estimating number of particles for a 0.5 micron foil, not 1.0 like in the example.

### DIFF
--- a/docs/source/usage/workflows/memoryPerDevice.py
+++ b/docs/source/usage/workflows/memoryPerDevice.py
@@ -29,7 +29,7 @@ cell_size = 0.8e-6 / 384.  # 2.083e-9 m
 
 y0 = 0.5e-6  # position of foil front surface (m)
 delta_y = 1.0e-6  # target thickness (m)
-y1 = y0 + delta_y # position of the foil rear surface (m)
+y1 = y0 + delta_y  # position of foil rear surface (m)
 L = 10.e-9  # pre-plasma scale length (m)
 L_cutoff = 4.0 * L  # pre-plasma length (m)
 
--- a/docs/source/usage/workflows/memoryPerDevice.py
+++ b/docs/source/usage/workflows/memoryPerDevice.py
@@ -27,8 +27,9 @@ maximum memory load due to hosting a large part of the target.
 
 cell_size = 0.8e-6 / 384.  # 2.083e-9 m
 
-y0 = 0.5e-6  # position of foil surface (m)
-y1 = 1.0e-6  # target thickness (m)
+y0 = 0.5e-6  # position of foil front surface (m)
+delta_y = 1.0e-6  # target thickness (m)
+y1 = y0 + delta_y # position of the foil rear surface (m)
 L = 10.e-9  # pre-plasma scale length (m)
 L_cutoff = 4.0 * L  # pre-plasma length (m)
 
--- a/docs/source/usage/workflows/memoryPerDevice.py
+++ b/docs/source/usage/workflows/memoryPerDevice.py
@@ -20,7 +20,7 @@ for our :ref:`FoilLCT example <usage-examples-foilLCT>` and its ``4.cfg``.
 It is an estimate for how much memory is used per device if the whole
 target would be fully ionized but does not move much. Of course the real
 memory usage depends on the case and the dynamics inside the simulation.
-We calculate the memory of just one device per row of GPUs in laser 
+We calculate the memory of just one device per row of GPUs in laser
 propagation direction. We hereby assume that particles are distributed
 equally in transverse direction, like it is set up in the FoilLCT example.
 """
--- a/docs/source/usage/workflows/memoryPerDevice.py
+++ b/docs/source/usage/workflows/memoryPerDevice.py
@@ -9,6 +9,7 @@ License: GPLv3+
 """
 
 from picongpu.utils import MemoryCalculator
+from math import ceil
 
 """
 @file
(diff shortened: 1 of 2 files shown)


PrometheusPi:
@pordyna Great catch - thanks for providing a fix via pull request directly.

@ax3l Could you add @pordyna to the list of authorized developers?

docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -27,8 +27,9 @@
 
 cell_size = 0.8e-6 / 384.  # 2.083e-9 m
 
-y0 = 0.5e-6  # position of foil surface (m)
-y1 = 1.0e-6  # target thickness (m)
+y0 = 0.5e-6  # position of foil front surface (m)
+delta_y = 1.0e-6  # target thickness (m)
+y1 = y0 + delta_y # position of the foil rear surface (m)
```
sbastrakov: Usage of `the` has now become a little inconsistent between the comments on the front and rear surfaces.

PrometheusPi: in this line there needs to be one additional space before the `#` to agree with the pep8 standard

ax3l:
@PrometheusPi added @pordyna - welcome to PIConGPU and thanks for your first PR! :)

docs/source/usage/workflows/memoryPerDevice.py
ax3l: Yep, double space between last character and comment in Python

pordyna:
@n01r So, now the whole target together with the vacuum cells in front of it, doesn't fit in the first row of GPUs. 
Should we write it  for both rows separately?  Maybe just point out that the second will need less memory? We should probably also add a `target_cells + vacuum_cells <= Ny` check.

n01r:
Why did you close the pull request? ^^
It's fine - the memory calculator is supposed to be used for the GPU you expect the most workload for.
It does not cover the whole simulation volume of all GPUs.
But your point was absolutely valid that target thickness was misleading since we need to know the number of cells that have particles and the documentation and naming should be changed to accomodate that.

ax3l:
Maybe just clicked the wrong button?

docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -27,8 +27,9 @@
 
 cell_size = 0.8e-6 / 384.  # 2.083e-9 m
 
-y0 = 0.5e-6  # position of foil surface (m)
-y1 = 1.0e-6  # target thickness (m)
+y0 = 0.5e-6  # position of foil front surface (m)
+delta_y = 1.0e-6  # target thickness (m)
```
PrometheusPi: @pordyna Would you add the suggestions from @n01r ?

ax3l: @pordyna ping :) whenever you have the time, please feel free to update the pull request by pushing the suggested changes to its branch :)

ax3l:
@n01r updated, pls feel free to take a look :)

docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -9,6 +9,7 @@
 """
```
n01r: You can add your name to the contributor list :)

docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -26,90 +27,113 @@
```
n01r: Please change the explanation for what this file does and add that it accounts for rows of GPUs.
e.g.:
*"...We calculate the memory of just one device per row of GPUs in laser propagation direction. We hereby assume that particles are distributed equally in transversal direction, like it is set up in the FoilLCT example."*

docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -26,90 +27,113 @@
...
-        cal_device * megabyte))
-
-mem_sum = (field_device + e_device + H_device + C_device +
-           N_device + rng_device + cal_device)
-print("Required memory per device: {:.2f} MB".format(
-    mem_sum * megabyte))
+target_x = Nx  # full transverse dimension of the GPU
+target_z = Nz
+
+sx = lambda n: {1: "st", 2: "nd", 3: "rd"}.get(n if n < 20 else int(str(n)[-1]), "th")
```
n01r: The Travis CI suite fails because it doesn't meet the PEP8 standard to assign a `lambda` function to a memory object.
Maybe transform it into a function definition with `def`. ^^

It's only valid until 20 now but that's okay for the example.

docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -26,90 +27,113 @@
...
+y1 = 1.5e-6  # position of foil rear surface (m)
+L = 10e-9  # pre-plasma scale length (m)
+L_cutoff = 4 * L  # pre-plasma length (m)
+
+sim_dim = 2
+# number of cells in the simulation
+Nx_all, Ny_all, Nz_all = 256, 1280, 1
+# number of GPU rows in each direction
+x_rows, y_rows, z_rows = 2, 2, 1
+# number of cells pro GPU
```
n01r: *... per GPU*

docs/source/usage/workflows/memoryPerDevice.py
pordyna: >  It's only valid until 20 now but that's okay for the example.

I'm pretty sure that it works fine until 111, that should be enough. IAnyway, it's not the most crucial part  :sweat_smile:

n01r: Oh, ha, I chose to overlook the `else` part. You're right. :P

docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -4,11 +4,12 @@
 This file is part of PIConGPU.
 
 Copyright 2018-2019 PIConGPU contributors
-Authors: Marco Garten
+Authors: Marco Garten, Paweł Ordyna
```
ax3l: Just a note: for files consumed by compilers in PIConGPU (C++ code), please try to use ASCII only since some older compilers and systems are not too flexible :-S
For Python code, this is fine though.

n01r:
ping @PrometheusPi ;)

Issue #2871, opened 2019-02-01
Compile issue with icc + Boost


sbastrakov:
It turned out we have issues with the size of Boost::mpl vector with the Intel compiler (discovered for icc version 19.0.0.117 (gcc version 7.3.0 compatibility)), probably same as [here](https://lists.boost.org/Archives/boost/2002/11/39144.php). The suspect are [these workarounds in our cmake](https://github.com/ComputationalRadiationPhysics/picongpu/blob/e8c4268c1771d80a7017068a021d648571841cbe/include/pmacc/PMaccConfig.cmake#L330-L332) and so we are falling back on the [pmacc defines](https://github.com/ComputationalRadiationPhysics/picongpu/blob/e8c4268c1771d80a7017068a021d648571841cbe/include/pmacc/types.hpp#L62-L63). Changing the latter to 30, 40 or 50 does not help with other issues appearing (and over 50 is not supported by Boost).

@psychocoderHPC and me spent some time investigating. It seems the problem is that sometimes there are too many combinations, e.g. with HDF5 and ADIOS the standard LaserWakefield example builds (and runs) fine, but the KelvinHelmholtz does not build. 

Details on the failing configuration:

- System: batch partitions of JURECA and JUWELS.
- Our standard profiles for these partitions, with HDF5 and ADIOS.
- Example: standard KelvinHelmholtz. Note that the standard LaserWakefield works fine, and KelvinHelmholtz works if we disable HDF5 or ADIOS.
- Compiler: icc 19.0.0.117 (gcc version 7.3.0 compatibility).
- Boost 1.68.0, as available on these systems for the enabled compiler. We suspect the issue is not specific to this version of Boost. Manually compiled Boost has not been tried yet.
- PIConGPU version: latest dev e8c4268 . We suspect the issue is not specific to this or any recent commit.
- [Output of the failing build](https://github.com/ComputationalRadiationPhysics/picongpu/files/2822633/Boost_icc_failing_build_log.txt)

**Current workaround**: reduce the number of combinations by disabling either ADIOS or HDF5. Untested: maybe reducing the number of combinations in `fileOutput.param` also helps.

ax3l:
Which version of boost did you test and how was it compiled?

psychocoderHPC:
It was boost 1.68.0 and some nearly up to date dev branch of PIConGPU.
[...]

PR #2874, opened 2019-02-01
System: JUWELS

sbastrakov:
Add new system documentation, profiles and tbg template for the [JUWELS cluster at JSC](http://www.fz-juelich.de/ias/jsc/EN/Expertise/Supercomputers/JUWELS/Configuration/Configuration_node.html).

The software stack seems the same as on JURECA #2869 .

It has 2 queues:

- batch: 2 x 24-core Intel Skylake CPUs per node, max 512 nodes per job
- gpus: 4 x NVidia V100 per node, max 48 nodes per job

TODO:
* [x] Add profile for GPUs
* [x] Update docs
* [x] Test GPU Queue

### DIFF
--- a/etc/picongpu/juwels-jsc/batch.tpl
+++ b/etc/picongpu/juwels-jsc/batch.tpl
@@ -1,0 +1,93 @@
+#!/usr/bin/env bash
+# Copyright 2013-2019 Axel Huebl, Richard Pausch, Rene Widera, Sergei Bastrakov
+#
+# This file is part of PIConGPU.
+#
+# PIConGPU is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# PIConGPU is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PIConGPU.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
+
+# PIConGPU batch script for JUWELS' SLURM batch system
+
+#SBATCH --account=!TBG_nameProject
+#SBATCH --partition=!TBG_queue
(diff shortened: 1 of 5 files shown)


etc/picongpu/juwels-jsc/batch.tpl
```diff
@@ -0,0 +1,93 @@
...
+#set user rights to u=rwx;g=r-x;o=---
+umask 0027
+
+mkdir simOutput 2> /dev/null
+cd simOutput
+ln -s ../stdout output
+
+if [ $? -eq 0 ] ; then
+  # Run PIConGPU
+  export OMP_NUM_THREADS=48
```
psychocoderHPC: Is hyperthreading activated for the nodes? I ask because if hyperthreading is disabled it must be `export OMP_NUM_THREADS=24`

sbastrakov: Yes, hyperthreading is enabled.

sbastrakov:
Ready for review.

sbastrakov:
The batch (CPU) queue works fine, except #2871 which we can't really fix now. The GPU queue is temporary not active due to some cluster issues and only the building has been tested. So I guess should not have removed the WIP from this PR.

ax3l:
If you like we can merge this for now and in case the GPU queues need updates for running you add a new PR.

ax3l:
(I prepare 0.4.3 now and want to include this PR)

ax3l:
@psychocoderHPC @sbastrakov ping: ok if I merge?

sbastrakov:
@ax3l wait a little please, the GPU queue seems to work now.

ax3l:
@sbastrakov any updates?

sbastrakov:
There is a problem with GPUs, investigating.

ax3l:
What's the issue you experience and do you think you can address it in the next hours?

sbastrakov:
Some dynamic library dependencies failing, and not clear why are they introduced at all. Looking now with @psychocoderHPC , theoretically should not take too long.

sbastrakov:
Fixed and checked, ready to be reviewed and then merged.
The issue was that the GPU partition does not provide Boost for the GPU-aware compiler + MPI setup, so I built it manually, but originally did not realize Boost used the system ICU dependency. (And it took hours in the queue to verify the fix).

ax3l:
Thanks, great work!

PR #2876, opened 2019-02-05
gridDist: verify devices and blocks

### LINKED ISSUE #2875: Mismatch -d and --gridDist
There appears to be no check whether the number of GPUs given via `-d` matches the number of GPUs given in `--gridDist`. 

The simulation still runs "fine" but has some empty cells in the front. 

Could this setup be checked during start up?

ax3l:
verify the number of blocks from the given `--gridDist` matches the number of devices `-d`.

Fix #2875 

(Sorry for the devices renaming parts, I was already on it.)

### DIFF
--- a/include/picongpu/initialization/ParserGridDistribution.hpp
+++ b/include/picongpu/initialization/ParserGridDistribution.hpp
@@ -17,18 +17,16 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
-
-
 #pragma once
 
 #include <pmacc/verify.hpp>
 #include <vector>   // std::vector
 #include <string>   // std::string
-#include <utility>  // std::pair
 #include <iterator> // std::distance
 
 #include <boost/regex.hpp>
 #include <boost/lexical_cast.hpp>
+
 
 namespace picongpu
 {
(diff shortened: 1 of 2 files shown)


include/picongpu/initialization/ParserGridDistribution.hpp
```diff
@@ -77,26 +76,39 @@ class ParserGridDistribution
...
             ++iter;
-            // go to last gpu of this block b{n}
+            // go to last device of this block b{n}
             i += iter->second;
         }
 
         return iter->first;
     }
 
+    /** Verify the number of blocks matches the devices
```
psychocoderHPC: `blocks` should be named  `domains` because we splitting the global domain into subdomains.

include/picongpu/initialization/ParserGridDistribution.hpp
```diff
@@ -77,26 +76,39 @@ class ParserGridDistribution
...
+     *
+     * Check that the number of blocks in a dimension does not exceed the
```
psychocoderHPC: `blocks` -> `domains`

include/picongpu/initialization/ParserGridDistribution.hpp
ax3l: agreed, just took the `block` naming already prevalent in this file :)

include/picongpu/initialization/ParserGridDistribution.hpp
```diff
@@ -77,26 +76,39 @@ class ParserGridDistribution
...
+     * number of devices for it.
+     *
+     * \param[in] numDevices number of devices for this dimension
+     */
+    void
+    verifyDevices( const int numDevices ) const
+    {
+        PMACC_VERIFY( parsedInput.size() == numDevices );
```
sbastrakov: The comment says "does not exceed" (and I think that is correct), but the code checks for equality.

ax3l: comment updated

ax3l:
@PrometheusPi can you RT check this solves #2876 for you?

ax3l:
@psychocoderHPC do you remember what the two `uints` in `value_type` stand for? We did not comment them.

include/picongpu/initialization/ParserGridDistribution.hpp
sbastrakov: I probably get it completely wrong (then just ignore, no need to explain in detail), but why do you check for equality here? [This comment](https://github.com/ComputationalRadiationPhysics/picongpu/blob/7830eafbda291bc7a3d344c0c9dabdd8e27ebd94/include/picongpu/initialization/ParserGridDistribution.hpp#L117-L119) says it fills the `parsedInput` with multipliers. In case there are multipliers greater than 1, would not `parsedInput.size()` be lesser than `numDevices` ?

ax3l:
I think it's subdomain extent and count, have to refactor...

include/picongpu/initialization/ParserGridDistribution.hpp
ax3l: yes, just saw that. `size()` will be too small

sbastrakov:
Think we could include smth like 
```C++
using Value = uint32_t;
using Multplier = uint32_t;
// maybe a text comment as well
using value_type = std::vector<std::pair<Value, Multplier> >;
```

ax3l:
I have it done already, thanks. will push an update for @PrometheusPi in a few minutes

ax3l:
@PrometheusPi now you can test it :)

include/picongpu/initialization/ParserGridDistribution.hpp
```diff
@@ -77,40 +90,57 @@ class ParserGridDistribution
...
+     * counts (1,1,n,m,e,f)
      *
-     *  \param[in] s as const std::string in the form a,b{n}
-     *  \return std::vector<pair> with uint32_t (base, multiplier)
+     * \param[in] s as string in the form a,b{n}
+     * \return std::vector<SubdomainPair> with 2x uint32_t (extent, count)
      */
     void
-    parseString( const std::string s )
+    parseString( std::string const s )
```
sbastrakov: Not important, but easy to implement: 

- returning a `value_type` instead of writing the result to directly to `parsedInput` seems more natural in terms of how it is used in the constructor (however, then the weird `value_type` might also have to be renamed to smth like `SubdomainPairs`)
- `parseString` could be safely shortened to `parse` given the argument type;

PrometheusPi:
If I have a mismatch, I get the following error:
```
[1,8]<stderr>:terminate called after throwing an instance of 'std::runtime_error'
[1,8]<stderr>:  what():  expression (numSubdomains == numDevices) failed in file (/bigdata/hplsim/scratch/rp6038/_test_pr2876/picongpu/include/pmacc/../picongpu/initialization/ParserGridDistribution.hpp:127) : 
```

PrometheusPi:
Surprisingly I could not reproduce the missing density bug ([ComputationalRadiationPhysics/projects-electrons#132](https://github.com/ComputationalRadiationPhysics/projects-electrons/issues/132)) on hypnos with ~32 GPUs - any configuration never showed no density.

include/picongpu/initialization/ParserGridDistribution.hpp
ax3l: good point, refactored that way now. I keep the `value_type` for now, it doesn't harm too much.

PrometheusPi:
@ax3l there seems to be an issue with spack - loading dependencies takes to long

ax3l:
@psychocoderHPC @PrometheusPi ready to be merged?

PR #2881, opened 2019-02-08
ForEachIdx::operator(): Use Universal Reference

### LINKED ISSUE #2877: Redundant separate implementations for const and non-const in ForEachIdx
It seems the attempt to separate implementations for const and non-const functors, that currently are just copies of one another, are actually not needed. A simple way seems to be just using a universal reference and a single implementation, [a small example](https://coliru.stacked-crooked.com/a/08fdccb610283a68).

```C++
#include <iostream>


struct Functor {
   void operator()() const { std::cout << "const version\n"; }
   void operator()() { std::cout << "non-const version\n"; }
};

template< class T_Functor >
void apply( T_Functor && functor ) {
   functor();   
}


int main()
{
    const Functor constFunctor;
    Functor nonconstFunctor;
    apply(constFunctor);
    apply(nonconstFunctor);
    return 0;
}
```

Discussed offline with @psychocoderHPC . Also discussed on this class:

- theoretically we probably do not want a non-const version at all (since functors are generally better stateless), but currently need both since some `const`s are missing in other places;
- for (a rather distant) future we might want to use `std::invoke` there instead of the function call syntax, [to be more compatible with how C++17 treats Callables](http://ericniebler.com/2018/12/05/standard-ranges/); ofc it is for distant future and needs to be checked with CUDA, so really not the priority.

sbastrakov:
Fixes #2877 . Also add tiny improvements to the code in this file.

### DIFF
--- a/include/pmacc/mappings/threads/ForEachIdx.hpp
+++ b/include/pmacc/mappings/threads/ForEachIdx.hpp
@@ -24,6 +24,8 @@
 #include "pmacc/mappings/threads/IdxConfig.hpp"
 #include "pmacc/types.hpp"
 
+#include <utility>
+
 
 namespace pmacc
 {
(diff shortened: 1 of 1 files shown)


include/pmacc/mappings/threads/ForEachIdx.hpp
```diff
@@ -75,7 +77,7 @@ namespace threads
          * The functor must fulfill the following interface:
          * @code
          * template< typename ... T_Args >
-         * void operator()( uint32_t const linearIdx, uint32_t const idx, T_Args && ... );
+         * void operator()( uint32_t linearIdx, uint32_t idx, T_Args && ... );
```
sbastrakov: Note: these removed `const`s are *not* part of the interface, which this comment is about, but implementation.

ax3l: But it's kind of an example people can clone from, so let's suggest `const` to them, no?

ax3l:
The term you are looking for is [universal reference](https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers).

include/pmacc/mappings/threads/ForEachIdx.hpp
sbastrakov: Ah, I did not think of that and approached this more formally (then adding `const` here does not change the function signature and so `const` is not part of the interface). Will revert.

sbastrakov:
@ax3l  sure, and I put it in my commit message (not the title, but the body).

ax3l:
@sbastrakov pls rebase against the current `dev` to trigger the (now) fixed CI :)

ax3l:
@sbastrakov pls rebase against the current `dev` to trigger the (now) fixed CI :)

sbastrakov:
Done.

PR #2887, opened 2019-02-13
Hypnos: CMake 3.13.4

ax3l:
Use a newer CMake on Hypnos.

For recent builds of openPMD-api and Alpaka a CMake 3.11+ is recommended, saving later troubles for users in post-processing.

### To Do

- [x] verify module works (first version: broken)

### DIFF
--- a/etc/picongpu/hypnos-hzdr/k20_picongpu.profile.example
+++ b/etc/picongpu/hypnos-hzdr/k20_picongpu.profile.example
@@ -22,7 +22,7 @@ then
 
         # Core Dependencies
         module load gcc/4.9.2
-        module load cmake/3.10.1
+        module load cmake/3.13.4
         module load boost/1.62.0
         module load cuda/8.0
         module load openmpi/2.1.2.cuda80
--- a/etc/picongpu/hypnos-hzdr/k80_picongpu.profile.example
+++ b/etc/picongpu/hypnos-hzdr/k80_picongpu.profile.example
@@ -22,7 +22,7 @@ then
 
         # Core Dependencies
         module load gcc/4.9.2
-        module load cmake/3.10.1
+        module load cmake/3.13.4
         module load boost/1.62.0
         module load cuda/8.0
         module load openmpi/2.1.2.cuda80
--- a/etc/picongpu/hypnos-hzdr/laser_picongpu.profile.example
+++ b/etc/picongpu/hypnos-hzdr/laser_picongpu.profile.example
@@ -22,7 +22,7 @@ then
 
         # Core Dependencies
         module load gcc/5.3.0
-        module load cmake/3.10.1
+        module load cmake/3.13.4
         module load boost/1.62.0
         module load openmpi/1.8.6
         module load numactl


psychocoderHPC:
cmake  3.13.4 is still broken on hypnos.

```
cmake -h
/opt/pkg/devel/cmake/3.13.4/bin/cmake: /opt/pkg/compiler/gnu/gcc/4.9.2/lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by /opt/pkg/devel/cmake/3.13.4/bin/cmake)
/opt/pkg/devel/cmake/3.13.4/bin/cmake: /opt/pkg/compiler/gnu/gcc/4.9.2/lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by /opt/pkg/devel/cmake/3.13.4/bin/cmake)
/opt/pkg/devel/cmake/3.13.4/bin/cmake: /opt/pkg/compiler/gnu/gcc/4.9.2/lib64/libstdc++.so.6: version `GLIBCXX_3.4.22' not found (required by /opt/pkg/devel/cmake/3.13.4/bin/cmake)
```

I will contact the admins.

ax3l:
I already wrote to them, they have to re-compile.

psychocoderHPC:
No problem there is now also an open ticket for that. I added you and @n01r to the ticket system.

n01r:
cc'ing @berceanu

berceanu:
https://github.com/ComputationalRadiationPhysics/picongpu-workshop/issues/8

psychocoderHPC:
The cmake module is now working but since Cmake 3.12 a behavior with environment variables `<PackageName>_ROOT` has changed an produce now a warning.

```
CMake Warning (dev) at /bigdata/hplsim/scratch/widera/dev/include/pmacc/PMaccConfig.cmake:282 (find_package):
  Policy CMP0074 is not set: find_package uses <PackageName>_ROOT variables.
  Run "cmake --help-policy CMP0074" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  Environment variable MPI_ROOT is set to:

    /opt/pkg/mpi/openmpi/2.1.2.cuda80/gnu/4.9.2/64/opt

  For compatibility, CMake is ignoring the variable.
Call Stack (most recent call first):
  CMakeLists.txt:83 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

# policy description
```
cmake --help-policy CMP0074
CMP0074
-------

``find_package()`` uses ``<PackageName>_ROOT`` variables.

In CMake 3.12 and above the ``find_package(<PackageName>)`` command now
searches prefixes specified by the ``<PackageName>_ROOT`` CMake
variable and the ``<PackageName>_ROOT`` environment variable.
Package roots are maintained as a stack so nested calls to all ``find_*``
commands inside find modules also search the roots as prefixes.  This policy
provides compatibility with projects that have not been updated to avoid using
``<PackageName>_ROOT`` variables for other purposes.

The ``OLD`` behavior for this policy is to ignore ``<PackageName>_ROOT``
variables.  The ``NEW`` behavior for this policy is to use
``<PackageName>_ROOT`` variables.

This policy was introduced in CMake version 3.12.  CMake version
3.13.4 warns when the policy is not set and uses ``OLD`` behavior.
Use the ``cmake_policy()`` command to set it to ``OLD`` or ``NEW``
explicitly.

.. note::
  The ``OLD`` behavior of a policy is
  ``deprecated by definition``
  and may be removed in a future version of CMake.
```

psychocoderHPC:
To have the old behavior it would require add the following snipped to all our cmake files. 
```
if(POLICY CMP0074)
    cmake_policy(SET CMP0074 OLD)
endif()
```

ax3l:
We should actually set it to `OLD`, it's an awesome policy.

sbastrakov:
@ax3l did you mean set it to `NEW` (if this is awesome)?

ax3l:
I addressed the warning in #2891 with extensive description and references.

psychocoderHPC:
@ax3l but we need to set it in each cmake files: alpaka, mallocMC, cupla, ...

I am wondering if there is not a way to disable it from a base project

ax3l:
No, base `CMakeLists.txt` where the `project()` is created should be sufficient.

psychocoderHPC:
@PrometheusPi  yes I tested it with the k80 profile.

PR #2889, opened 2019-02-13
One Position Init: Check Density Probes

### LINKED ISSUE #2888: Probes: OnePosition Profile Init
@x113f just reported that probe init (particles without weighting) might falsely ignore a "mask" given in a density profile when used with `CreateDensity` and `startPosition::OnePosition`. (Result: particles everywhere instead of only at > zero cells of the density functor.)

Update: Maybe the upcoming `0.4.3` release and current `dev` fix the `OnePosition` issue already with PR #2831...?

Investigating if a work-around with `startPosition::Quiet1ppc` work-arounds this. -> no: #2890

ax3l:
Probe particles usually have no weighting. Our `startPosition::OnePosition` should still honor the density functor of the user as a binary mask.

Fix #2888.

cc @x113f can you verify this solves your issue?

Update: For upcoming `0.4.3` and current `dev`,  PR #2831 might already solve this issue.

### DIFF
--- a/include/picongpu/particles/startPosition/OnePositionImpl.hpp
+++ b/include/picongpu/particles/startPosition/OnePositionImpl.hpp
@@ -122,6 +122,8 @@ namespace detail
                     m_weighting
                 );
 
+            result = realParticlesPerCell > 0.0_X ? result : 0.0_X;
+
             return result;
         }
 


ax3l:
For upcoming `0.4.3` and current `dev`,  PR #2831 might already solve this issue.

psychocoderHPC:
@ax3l Could you please explain more detailed why this PR is needed and where I see any binary mask.

include/picongpu/particles/startPosition/OnePositionImpl.hpp
```diff
@@ -122,6 +122,8 @@ namespace detail
                     m_weighting
                 );
 
+            result = realParticlesPerCell > 0.0_X ? result : 0.0_X;
```
psychocoderHPC: OK it takes a while until I got why you add this line.
In the case where we have no real particles you must set the number of particles to zero.

Since this is a special case if `hasWeighting == false`, please add this line into the else branch if the if condition.

psychocoderHPC: IMO #2831 should fully solve the issue and I would say this PR is not needed. @ax3l  I am wrong?

psychocoderHPC: I checked it again and #2831 [this line](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2831/files#diff-e80f8636fbb31c1a3c7bc8a2f1366234R246) avoids that `numberOfMacroParticles` is called for any density `<= 0`

ax3l: Yes, as I wrote I expect to #2831 have fixed this already.
@x113f reports the issue on `0.4.2` which doesn't have that fix already.

x113f:
This work-around fixed my bug. :+1: Thanks again @ax3l .
If #2831 fixes this issue for `0.4.3`, even better. I will still wait to upgrade, until you declare `0.4.3` stable

ax3l:
Ok, then I will close this but you can of course still use this as a quick -fix until 0.4.3 is out :)

PR #2891, opened 2019-02-14
CMake: Honor _ROOT Env Hints

ax3l:
CMake 3.12.0+ honor `<Package>_ROOT` environment hints which are often set on HPC systems. Previously, it was only looking for `<Package>_DIR` paths in `find_package` calls.

This new policy is useful since HPC systems usually set `_DIR`, `_ROOT` or expand the `CMAKE_PREFIX_PATH`. Therefore we want to use it as soon as it is available.

On systems where those env vars are set, e.g. Hypnos, this also throws a warning if the default (OLD) policy is used with CMake 3.12.4 or newer.

### References

- https://cmake.org/cmake/help/v3.12/policy/CMP0074.html
- https://github.com/openPMD/openPMD-api/pull/391
- https://github.com/ComputationalRadiationPhysics/picongpu/pull/2887#issuecomment-463602968

## Side Note

If we update to CMake 3.12+ one day, the policies at this point become `NEW` by default as well:

> The cmake_minimum_required() command does more than report an error if a too-old version of CMake is used to build a project. It also sets all policies introduced in that CMake version or earlier to NEW behavior.

https://cmake.org/cmake/help/v3.12/manual/cmake-policies.7.html

### DIFF
--- a/include/mpiInfo/CMakeLists.txt
+++ b/include/mpiInfo/CMakeLists.txt
@@ -45,6 +45,18 @@ list(APPEND CMAKE_PREFIX_PATH "$ENV{CMAKE_PREFIX_PATH}")
 list(APPEND "/usr/lib/x86_64-linux-gnu/")
 
 
+################################################################################
+# CMake policies
+#
+# Search in <PackageName>_ROOT:
+#   https://cmake.org/cmake/help/v3.12/policy/CMP0074.html
+################################################################################
+
+if(POLICY CMP0074)
+    cmake_policy(SET CMP0074 NEW)
+endif()
+
+
 ###############################################################################
 # Language Flags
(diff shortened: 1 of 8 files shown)


ax3l:
@psychocoderHPC note: I would merge this PR and the CMake update for hypnos, which limits the warnings but does not fix all.

We have to update all projects anyway, but I would not do this now for the 0.4.3 but in the next release.

psychocoderHPC:
> (We can also unset `CUDA_ROOT` on hypnos in our profiles if we want to suppress the warning in the meantime.)

I would not update the profiles since we are working on an update of all projects to avoid the warning.

ax3l:
ready to be merged :)

PR #2895, opened 2019-02-14
Merge 0.4.3 Changelog

ax3l:
Merge back `master` to have the changelog of 0.4.3 included.

```bash
git merge --strategy-option ours mainline/master
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,63 @@
 Changelog
 =========
 
+0.4.3
+-----
+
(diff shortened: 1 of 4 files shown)


PrometheusPi:
@ax3l why are all these older commits (which should already be in the `dev`) listed? Was this accidentally caused by your force push?

ax3l:
This is a merge commit synchronizing `master` to `dev` so the git history is straight again. Therefore, all included (not necessarily visible) commits in the merge are shown.

PR #2899, opened 2019-02-19
split `ParserGridDistribution` into `hpp/cpp` file

### LINKED ISSUE #2898: Boost Error in ResourceLog
In combination with:
* GCC 7.3.1
* CUDA 9.2.88
* Boost 1.68.0 (C++11 build) OR Boost 1.65.1
* PIConGPU 0.4.0
* centos 7

I have checked the issue here:
https://github.com/ComputationalRadiationPhysics/picongpu/issues/2714#issue-367113446
But, there  is still such err in ResourceLog.hpp:
```
/trinity/home/b.lei/pic/src/blei/picongpu/include/pmacc/../picongpu/plugins/ResourceLog.hpp(115): here
/trinity/shared/opt/rh/devtoolset-7/root/usr/include/c++/7/bits/stl_pair.h(331): error: identifier "std::basic_string<char, std::char_traits<char>, std::allocator<char> >::_Rep::_S_empty_rep_storage" is undefined in device code

/trinity/shared/opt/rh/devtoolset-7/root/usr/include/c++/7/bits/stl_pair.h(331): error: identifier "std::basic_string<char, std::char_traits<char>, std::allocator<char> >::_Rep::_S_empty_rep_storage" is undefined in device code

2 errors detected in the compilation of "/tmp/tmpxft_0000926d_00000000-6_main.cpp1.ii".
CMake Error at picongpu_generated_main.cpp.o.Release.cmake:279 (message):
  Error generating file
  /trinity/home/b.lei/pic/build/CMakeFiles/picongpu.dir//./picongpu_generated_main.cpp.o
```

psychocoderHPC:
hopefully, fix #2898

- Split the grid parser into an interface file `hpp` and the implementation `cpp` to avoid that boost regex is parsed by nvcc. Boost regex is very often the root of compile issues because nvcc is not able to split the host and device part correctly.
- Define `abortWithError` static to allow the usage in multiple object files without creating the linker error ` multiple definitions of ...
- remove `boost::regex` by `std::regex`
- remove `boost::lexical_cast`

### DIFF
--- a/include/picongpu/initialization/ParserGridDistribution.cpp
+++ b/include/picongpu/initialization/ParserGridDistribution.cpp
@@ -1,0 +1,142 @@
+/* Copyright 2013-2019 Axel Huebl, Rene Widera, Benjamin Worpitz
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "picongpu/initialization/ParserGridDistribution.hpp"
+
+#include <pmacc/verify.hpp>
+#include <cstdint>
+#include <vector>   // std::vector
+#include <string>   // std::string
+#include <iterator> // std::distance
+
+#include <boost/regex.hpp>
+#include <boost/lexical_cast.hpp>
+
+
+namespace picongpu
+{
+
+    ParserGridDistribution::ParserGridDistribution( std::string const s )
+    {
+        parsedInput = parse( s );
+    }
+
+    uint32_t
+    ParserGridDistribution::getOffset( uint32_t const devicePos, uint32_t const maxCells ) const
+    {
+        value_type::const_iterator iter = parsedInput.begin();
+        // go to last device of these n subdomains extent{n}
+        uint32_t i = iter->count - 1u;
+        uint32_t sum = 0u;
+
+        while( i < devicePos )
+        {
+            // add last subdomain
+            sum += iter->extent * iter->count;
+
+            ++iter;
+            // go to last device of these n subdomains extent{n}
+            i += iter->count;
+        }
+
+        // add part of this subdomain that is before me
+        sum += iter->extent * ( devicePos + iter->count - i - 1u );
+
+        // check total number of cells
+        uint32_t sumTotal = 0u;
+        for( iter = parsedInput.begin(); iter != parsedInput.end(); ++iter )
+            sumTotal += iter->extent * iter->count;
+
+        PMACC_VERIFY( sumTotal == maxCells );
+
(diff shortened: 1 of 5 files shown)


psychocoderHPC:
@ax3l The suggested refactoring to c++ regex you mentioned https://github.com/ComputationalRadiationPhysics/picongpu/issues/2898#issuecomment-465128305 can be added in a follow-up PR.

ax3l:
@psychocoderHPC thx! If you ask me, add the C++11 regex in the same PR. We have to backport both anyway.

psychocoderHPC:
I do not think that this PR fill fix #2899 fully. I will also have a look to the resource manager.

sbastrakov:
@psychocoderHPC is it not possible to make `abortWithError` inline and non-static?

ax3l:
`inline` or `static` for global functions only instruct the linker to either always take the local symbol or assume they are all the same. either way, it will work here:

https://stackoverflow.com/questions/185624/static-variables-in-an-inlined-function

ax3l:
> I do not think that this PR fill fix #2899 fully. I will also have a look to the resource manager.

Yep, I just would like to have the `gridDist` parser changes in one place :)

sbastrakov:
@ax3l sure, for our case both `static` and `inline` would do the job. Actually, from a purist point of view it is even better to put a function in an anonymous namespace (yes, even in a header file). But practically (even for me) seems like too much typing for a single function.

ax3l:
Agreed, an anonymous namespace is actually the best option.

psychocoderHPC:
I updated the code and removed the boost dependencies for the parser.
Please be not confused by the strange escaping e.g. ` R"(\{.*})",`. 
Only `{` must be escaped (this behavior cost me over 1h of my lifetime)

psychocoderHPC:
This PR is still WIP because I am checking if we can remove the dependency to `boost::regex` now.

ax3l:
It seems that `boost::iostreams` and `boost::log` depend on `boost::regex`. `boost::program_options` does surprisingly not.

https://gitlab.kitware.com/cmake/cmake/blob/v3.13.4/Modules/FindBoost.cmake#L845-878

include/picongpu/initialization/ParserGridDistribution.cpp
```diff
@@ -25,8 +25,7 @@
 #include <string>   // std::string
 #include <iterator> // std::distance
 
-#include <boost/regex.hpp>
-#include <boost/lexical_cast.hpp>
+#include <regex>
```
ax3l: bye, byte GCC 4.8 :stars: 
(`INSTALL.md` already lists GCC 4.9+, so that's fine.)

psychocoderHPC:
> It seems that `boost::iostreams` and `boost::log` depend on `boost::regex`. `boost::program_options` does surprisingly not, not even transitively.
> 
> https://gitlab.kitware.com/cmake/cmake/blob/v3.13.4/Modules/FindBoost.cmake#L845-878
> 
> https://github.com/ComputationalRadiationPhysics/picongpu/blob/0.4.3/INSTALL.rst#boost
> https://github.com/ComputationalRadiationPhysics/picongpu/blob/0.4.3/include/picongpu/CMakeLists.txt#L168-L169
> https://github.com/ComputationalRadiationPhysics/picongpu/blob/0.4.3/include/pmacc/PMaccConfig.cmake#L308

@ax3l We do not use boost::log or boost::iostream, I tested now also if we can remove the dependency and it looks good.

psychocoderHPC:
@ax3l I removed in the last commit the requirement to have boost::regex. But since splash2txt still need it I only marked boost::regex in the documentation as optional.

I will put on my todo list to remove boost::regex from splash2txt too.

ax3l:
looks good to me, runtime tests were successful?

psychocoderHPC:
Yes I tested different distributions.

PR #2901, opened 2019-02-20
pic-configure: add `--force/-f` option

psychocoderHPC:
In the case where new parameter files are added to the input parameter into a already compiled parameter set, cmake will ignore the new files.
The option `-f` allows to enforce that cmake flushes the cache and rescans for new files.

This option is required in the following workflow:
- create an input set where e.g `particlesFilter.param` is missing
- run `pic-build`
- run `pic-edit particlesFilter`
- add a new filter
- run `pic-build`
- run `./bin/picongpu --help`

The last command will not show the new filter in the hdf5 source or everywhere else where filters will be used. A way to trigger a full rescan of cmake is to delete the folder `.build` or the new option `-f`.

The option `-f` has the advantage that all changes which were made with `ccmake` will still be in.  This is very often required for a developer.

In our workflow an explicit cleanup of the cmake cache is also required if `pic-edit` is copying a new param file into the input set else the new file will not be used by cmake. Since `pic-edit` is not aware that the build folder is `.build` I did not add this feature into `pic-edit`. Never the less we should do it.

### DIFF
--- a/bin/pic-configure
+++ b/bin/pic-configure
@@ -46,6 +46,7 @@ help()
     echo "-c | --cmake         - overwrite options for cmake"
     echo "                       (e.g.: \"-DPIC_VERBOSE=21 -DCMAKE_BUILD_TYPE=Debug\")"
     echo "-t <presetNumber>    - configure this preset from cmakeFlags"
+    echo "-f | --force         - clear the cmake file cache and scan for new param files"
     echo "-h | --help          - show this help message"
 }
 
@@ -92,7 +93,7 @@ get_backend_flags()
 }
 
 # options may be followed by one colon to indicate they have a required argument
-OPTS=`getopt -o i:b:c:p:t:h -l install:,backend:,cmake:,params:,help -- "$@"`
+OPTS=`getopt -o i:b:c:p:t:hf -l install:,backend:,cmake:,params:,help,force -- "$@"`
 if [ $? != 0 ] ; then
     # something went wrong, getopt will put out an error message for us
     exit 1
@@ -128,6 +129,9 @@ while true ; do
         -h|--help)
             echo -e "$(help)"
             exit 0
+            ;;
+       -f|--force)
+            force="true"
             ;;
         -c|--cmake)
             cmake_options="$2"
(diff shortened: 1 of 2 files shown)


ax3l:
by the way, does out `submit.sh` script detect and warn still?

bin/pic-configure
```diff
@@ -46,6 +46,7 @@ help()
     echo "-c | --cmake         - overwrite options for cmake"
     echo "                       (e.g.: \"-DPIC_VERBOSE=21 -DCMAKE_BUILD_TYPE=Debug\")"
     echo "-t <presetNumber>    - configure this preset from cmakeFlags"
+    echo "-f | --force         - flush the cmake cache and scan for new param files"
```
sbastrakov: Maybe `clear` is more descriptive than `flush` ("cache flush" is actually a different thing)?

ax3l: true, @psychocoderHPC can you please replace `flush` with `clear`?

psychocoderHPC:
> by the way, does at `tbg` time our `submit.sh` script still detect and warn these new uncompiled files?

I tested the behavior again and tbg (triggered by submitAction.sh) is warning. I will extend the warning with the warning and point the user to the new option to guide him how he can solve the issue.

psychocoderHPC:
Could someone please check the sentence `Run 'pic-build -f' to recompile with the modified files.` I am not sure if it should be `Run 'pic-build -f' to recompile and include the modified files.` instead.

bin/pic-configure
```diff
@@ -46,6 +46,7 @@ help()
     echo "-c | --cmake         - overwrite options for cmake"
     echo "                       (e.g.: \"-DPIC_VERBOSE=21 -DCMAKE_BUILD_TYPE=Debug\")"
     echo "-t <presetNumber>    - configure this preset from cmakeFlags"
+    echo "-f | --force         - clear the cmake file cache and scan for new param files"
```
ax3l: new [param] files is a bit of a lie, since it will scan for all new `include/` files, but it's ok with me since it is more clear to the user when they should use it.

psychocoderHPC: Yes I keep it more user-focused.

PR #2904, opened 2019-02-21
pic-edit: clean cmake file cache if new param added

psychocoderHPC:
follow up of #2901

If a param file is copied to the input set of the user the cmake cache file cache fill be cleaned to avoid that the new param file is not taken into account during the compile.

Note: I added no output to the terminal that the cmake cache is cleared. I think it will only confuse the user at this place.  

- [x] merge after #2901

### DIFF
--- a/bin/pic-edit
+++ b/bin/pic-edit
@@ -142,6 +142,9 @@ do
         defaults_path="$picongpu_prefix/$file_prefix/$file_name"
         if [ -f "$defaults_path" ]
         then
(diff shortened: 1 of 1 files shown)


ax3l:
Good call. I would even consider this one a potential candidate for a backport, since the currently recommended workflow has a glitch that it fixes.

psychocoderHPC:
I am fine with backporting. I also thought about it but have sometimes the feeling it is useful to backport nearly all PR^^

PR #2906, opened 2019-02-23
Docker CUDA Image: Hwloc Default

ax3l:
Our currently shipped Docker image build for the CUDA backend with Spack. Due to a concretization issue in OpenMPI's package, we also need to specify the `hwloc` default variant to be with `+cuda`.

Follow-up to #2847.


ax3l:

PR #1603, opened 2016-10-25
Boost 1.60+ & CUDA 7/8 Compile

psychocoderHPC:
- disable boost template aliases
- remove that boost variadic templates are disabled for CUDA < 7.5

This pull request removed #1325 and disabled template aliases instead. This allows that Boost is allowed to use variadic templates instead of precompiler magic.
template aliases only disable things like:
```C++11
template< typename T>
using X = typename Trait< T >::type;
```

#1602 is needed to compile CUDA8, but this pull request can merged independently.

PR #2908, opened 2019-02-23
Boost 1.67.0+ Template Aliases

### LINKED ISSUE #2907: Boost 1.68.0 & CUDA 9.0.176 Fails
I am currently seeing the Boost error from #2706 again with:

- PIConGPU 0.4.3
- GCC 5.4.0
- Boost 1.68.0
- CUDA 9.0.176

in our Docker build.

ax3l:
Enable boost template aliases. Fix #2907

Boost 1.67.0+ uses `boost::void_t` even though it is only conditionally defined (controlled via `BOOST_NO_CXX11_TEMPLATE_ALIASES`): https://github.com/boostorg/type_traits/issues/113

https://www.boost.org/doc/libs/1_67_0/boost/type_traits/detail/has_binary_operator.hpp

This section in our config was originally designed to work-around https://svn.boost.org/trac10/ticket/11897
The NVCC bug appeared in CUDA 7.0, was fixed in 7.5 and re-appeared in 8.0. During transition to CUDA 8, the work-around was switched from disabling variadic templates to disabling template aliases, as it also excludes boost `mp_defer` code of the original issues. Nevertheless, there seems to be no reason why we would need boost variadic templates with CUDA 8.0 (at least no details were given in #1603 where this would fail and only indicates stylistic reasons). As the specific usage of variadic templates in boost `mp_defer` are the actual problem with CUDA 8.0, use the original work-around again.

CI will test CUDA 8.0.44 + Boost 1.62.0 .
Boost [1.63.0+](https://github.com/boostorg/config/blob/boost-1.63.0/include/boost/config/compiler/nvcc.hpp) applies the same work-around in its configs for CUDA 8.0 (`BOOST_NO_CXX11_VARIADIC_TEMPLATES`).

### DIFF
--- a/include/pmacc/PMaccConfig.cmake
+++ b/include/pmacc/PMaccConfig.cmake
@@ -320,19 +320,15 @@ endif()
 message(STATUS "Boost: result_of with TR1 style and decltype fallback")
 set(PMacc_DEFINITIONS ${PMacc_DEFINITIONS} -DBOOST_RESULT_OF_USE_TR1_WITH_DECLTYPE_FALLBACK)
 
-# Boost >= 1.60.0 and CUDA != 7.5 failed when used with C++11
-# seen with Boost 1.60.0 - 1.62.0 (maybe later) and CUDA 7.0 and 8.0
-# CUDA 7.5 works without a workaround
-# CUDA 8.0.44 bug reappeared, but work-around was mainlined in 1.63.0+
-# CUDA 9.0 unknown, likely fixed
-# CUDA 9.1.85-3 and above fixed
+# Boost >= 1.60.0 and CUDA 8.0 fail to compile on mp_defer
+# seen with Boost 1.60.0 - 1.63.0 (latter got config work-around below)
+# CUDA 9.0+ fixed
 if("${ALPAKA_CUDA_COMPILER}" STREQUAL "nvcc")
     if( (Boost_VERSION GREATER 105999) AND
-        (NOT CUDA_VERSION VERSION_EQUAL 7.5) AND
-        (CUDA_VERSION VERSION_LESS 9.1) )
+        (CUDA_VERSION VERSION_EQUAL 8.0) )
         # Boost Bug https://svn.boost.org/trac/boost/ticket/11897
-        message(STATUS "Boost: Disable template aliases")
-        set(PMacc_DEFINITIONS ${PMacc_DEFINITIONS} -DBOOST_NO_CXX11_TEMPLATE_ALIASES)
+        message(STATUS "Boost: Disable variadic templates")
+        set(PMacc_DEFINITIONS ${PMacc_DEFINITIONS} -DBOOST_NO_CXX11_VARIADIC_TEMPLATES)
     endif()
 endif()
 


include/pmacc/PMaccConfig.cmake
```diff
@@ -322,14 +322,15 @@ set(PMacc_DEFINITIONS ${PMacc_DEFINITIONS} -DBOOST_RESULT_OF_USE_TR1_WITH_DECLTY
...
 # Boost >= 1.60.0 and CUDA != 7.5 failed when used with C++11
 # seen with Boost 1.60.0 - 1.62.0 (maybe later) and CUDA 7.0 and 8.0
+#   https://svn.boost.org/trac10/ticket/11897
 # CUDA 7.5 works without a workaround
 # CUDA 8.0.44 bug reappeared, but work-around was mainlined in 1.63.0+
-# CUDA 9.0 unknown, likely fixed
+# CUDA 9.0 likely fixed
 # CUDA 9.1.85-3 and above fixed
 if("${ALPAKA_CUDA_COMPILER}" STREQUAL "nvcc")
     if( (Boost_VERSION GREATER 105999) AND
```
ax3l: maybe add `AND (Boost_VERSION LESS 106300)`

include/pmacc/PMaccConfig.cmake
```diff
@@ -320,19 +320,15 @@ endif()
...
+# CUDA 9.0+ fixed
 if("${ALPAKA_CUDA_COMPILER}" STREQUAL "nvcc")
     if( (Boost_VERSION GREATER 105999) AND
-        (NOT CUDA_VERSION VERSION_EQUAL 7.5) AND
-        (CUDA_VERSION VERSION_LESS 9.1) )
+        (CUDA_VERSION VERSION_EQUAL 8.0) )
         # Boost Bug https://svn.boost.org/trac/boost/ticket/11897
-        message(STATUS "Boost: Disable template aliases")
-        set(PMacc_DEFINITIONS ${PMacc_DEFINITIONS} -DBOOST_NO_CXX11_TEMPLATE_ALIASES)
+        message(STATUS "Boost: Disable variadic templates")
```
ax3l: Actually disabling variadic templates was the original work-around to address the issues in `mp_defer`.
Just by coincidence, `mp_defer` is also no included when setting the totally unrelated `BOOST_NO_CXX11_TEMPLATE_ALIASES` switch, introduced here while trying to add CUDA 8.0 support:
https://github.com/ComputationalRadiationPhysics/picongpu/commit/3c40840d99b2493a9a5f2ae011a2003ba226893a

ax3l:

psychocoderHPC:
> there seems to be no reason why we would need boost variadic templates with CUDA 8.0 (at least no details were given in #1603 where this would fail and only indicates stylistic reasons). As the specific usage of variadic templates in boost `mp_defer` are the actual problem with CUDA 8.0, use the original work-around again.

from #1603: `This allows that Boost is allowed to use variadic templates instead of precompiler magic.`

#1603 enabled to use a real language feature (variadic templates) instead of doing some macro stuff which emulates something which is part of C++ and limiting us like in #2871

ax3l:
> `This allows that Boost is allowed to use variadic templates instead of precompiler magic.`

Yep, but *why* would we as users of boost need this? It's stylistic. I think this just slipped through as other things were changed.

psychocoderHPC:
Is your question why we need variadic templates in boost? If so: we need it to allow a arbitrary number of elements for boost::mpl::vector<>

ax3l:
Yes, I was just wondering why we needed it at the time of #1603, since we did not document that we already reached the limit.

psychocoderHPC:
In that time we do not reached the limit as I know but also for compatibility we should try to base on language features as much as possible. 
But since this pr only effects CUDA 8 which should not be used anymore(expect if you like to use fermi gpus) it is ok to activate the workaround.

PR #2911, opened 2019-02-28
[WIP] Jupyter widgets: automatic update of plot options 

codingS3b:
At the moment the options of each jupyter widget (like 'species' or 'species_filter') are hardcoded to a single option (see screenshot). 
widget

This PR will improve this situation by dynamically adapting those options based on the values common to the selected simulations.

Before implementing something we should discuss the following:
a) is it sufficient to update those parameters only when the selected simulations change? Or do we also update them when the selected simulation time changes or when any of the other options changes?

b) Should the query for available species, filters, and other options be implemented in the data classes or in the widget classes?

### DIFF
--- a/lib/python/picongpu/plugins/data/base_reader.py
+++ b/lib/python/picongpu/plugins/data/base_reader.py
@@ -121,3 +121,38 @@ class DataReader(object):
         dependent format and type.
         """
         raise NotImplementedError
+
+    # NOTE: this can be hard to define abstractly since we should make each
+    # parameter query dependent on all other possible choices of parameters
+    # thus requiring kwargs?!
+
+    # e.g. get all species strings for defined values of filter, iteration
+    # and other parameters
+    def get_species(self, iteration=None, species_filter=None):
+        """
+        Get the available species abbreviations for which data exists.
+        Optionally specific iterations and/or filters can be given.
+
+
+        Returns
+        -------
+        A list of species abbreviation strings for which data exists.
+        In case iteration or species_filter are not None, only species
+        are returned that match iteration and species_filter.
+        """
+        raise NotImplementedError
+
+    def get_species_filters(self, iteration=None, species=None):
+        raise NotImplementedError
+
+    def get_options(self, iteration=None, species=None, species_filter=None):
+        """
+        Returns
+        -------
+        a dictionary mapping keywords that a visualizer might use
+        to the values for which data is present.
+        """
+        return {
+            'species': self.get_species(iteration, species_filter),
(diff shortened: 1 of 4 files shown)


ax3l:
Thank you, that's a wonderful update!

> a) is it sufficient to update those parameters only when the selected simulations change? Or do we also update them when the selected simulation time changes or when any of the other options changes?

Currently yes, since we compile in the general options and parametrize them at runtime. Nevertheless, we would like to change them at runtime potentially, e.g. with cling, and have done quite flexible functor chains with ISAAC already.

> b) Should the query for available species, filters, and other options be implemented in the data classes or in the widget classes?

I think that would be a fantastic data (or even input reader) class. do you have an idea where to parse them from? We could also add such info to our `simOutput/output` text file (available species is already in there, yet not the available species filters per species) if it helps.

PrometheusPi:
@codingS3b What is the status of this pull request? Can I start reviewing?

codingS3b:
Sorry, kind of forgot about that issue. It is definitely not ready for review!
I don't really know how to get the necessary information for dynamic option updates based on the picongpu output files and help would be greatly appreciated here.

So basically I would need some class that tells me (given a simulation output directory) which 'species' and which 'Species_filter' options are available for a chosen plugin (like the energy_histogram or the phase_space plugin).

Once this functionality exists, the rest should be relatively easy (by using the options common to all selected simulations).

ax3l:
Hm, it might help to have the openPMD plugin in place #2966 and dump at `t=0` a `make_empty` version of our simulation in `json`. @franzpoeschel

franzpoeschel:
The `json` backend is currently not available for parallel usage in the openPMD API, so your best bet will be the ADIOS backend for the moment. (HDF5 backend still has a [bug](https://github.com/openPMD/openPMD-api/issues/524))

ax3l:
The json backend is intentionally serial (and will likely stay that way for good). The idea is to make a single, `make_empty` "mocked" dump from a master rank after reducing the meta data to it.

PrometheusPi:
@codingS3b Will you continue working on this pull request or should we close it (for now)?

codingS3b:
@PrometheusPi I am not working on this right now since I hope that, after converting all data readers to work with openPMD inputs, getting the necessary information will become much easier. So, if development on the readers is not close to being finished, this can be closed for now and re-opened later.

PrometheusPi:
@codingS3b Thanks for the input. I agree that this should be supported by the readers. As suggested I will close this pull request.

PR #2912, opened 2019-02-28
fix compile splash2txt

psychocoderHPC:
- fix broken compile due to a wrong interface usage introduced with #2520
- cmake missing include/dependency
  - add splash to the cmake target `splash2txt`
  - add splash dependency `zlib` to the `splash2txt` target

### DIFF
--- a/src/tools/splash2txt/CMakeLists.txt
+++ b/src/tools/splash2txt/CMakeLists.txt
@@ -104,6 +104,11 @@ if(MPI_CXX_FOUND)
     set(LIBS ${LIBS} ${MPI_CXX_LIBRARIES})
 endif(MPI_CXX_FOUND)
 
+################################################################################
+# Find zlib
+################################################################################
+
+find_package(ZLIB REQUIRED)
 
 ################################################################################
 # libSplash (+ hdf5 due to required headers)
@@ -160,6 +165,8 @@ add_executable(splash2txt
      ${SRCFILES}
      )
 
+target_link_libraries(splash2txt Splash::Splash)
+target_link_libraries(splash2txt ZLIB::ZLIB)
 target_link_libraries(splash2txt m ${LIBS})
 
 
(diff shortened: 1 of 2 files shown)


src/tools/splash2txt/CMakeLists.txt
```diff
@@ -160,6 +165,8 @@ add_executable(splash2txt
      ${SRCFILES}
      )
 
+target_link_libraries(splash2txt Splash::Splash)
```
ax3l: does `Splash::Splash` not add `zlib`?

psychocoderHPC: No it is not adding zlib. We do it in PIConGPU also separate.

ax3l: ok, then we should report this as an issue in libSplash

ax3l: I don't even know why we depend on zlib in Splash, we don't have it in our interfaces... Probably it's a missing dependency in HDF5 already that we overlink for sanity.

ax3l: Just did testing on hypnos: we don't need to link zlib here.

psychocoderHPC: I tested it on hemera and the zlib is required. Keep in mind that maybe hypnos modules are older and maybe compiled against the system zlib.

psychocoderHPC: this is the error message without zlib:

```
CMake Error at CMakeLists.txt:164 (add_executable):
  Target "splash2txt" links to target "ZLIB::ZLIB" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
````

ax3l: Jep, this is already overlinked by libSplash: https://github.com/ComputationalRadiationPhysics/libSplash/issues/283

PR #2914, opened 2019-03-01
Splash2txt Build: Update deps

ax3l:
- ~~do not link libz (zlib) since it's not needed directly~~ `ZLIB::ZLIB` without target weirdly injected by Splash 1.7.0
- do link libm only on Unix and not with ICC
- link with proper qualifiers (`PRIVATE` in the case of an executable)

Follow-up to #2912

### DIFF
--- a/src/tools/splash2txt/CMakeLists.txt
+++ b/src/tools/splash2txt/CMakeLists.txt
@@ -104,11 +104,6 @@ if(MPI_CXX_FOUND)
     set(LIBS ${LIBS} ${MPI_CXX_LIBRARIES})
 endif(MPI_CXX_FOUND)
 
-################################################################################
-# Find zlib
-################################################################################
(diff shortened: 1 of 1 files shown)


src/tools/splash2txt/CMakeLists.txt
```diff
@@ -166,8 +172,6 @@ add_executable(splash2txt
      )
 
 target_link_libraries(splash2txt Splash::Splash)
-target_link_libraries(splash2txt ZLIB::ZLIB)
```
ax3l: hm, some dependency injects this in our lib list. so we might need it for now.

psychocoderHPC:
@ax3l  do you tested if splash still compiles?

ax3l:
Yep, hypnos and hemera

PR #2919, opened 2019-03-12
Add k80_picongpu.profile and k80.tpl

n01r:
Created the `k80_picongpu.profile.example` as well as the `k80.tpl` for the k80 nodes at HZDR after migration from Hypnos to Hemera.

I compiled an LWFA example and ran it successfully with the `16.cfg`.

Haven't tried HDF5 or ADIOS output, yet.

@ax3l, @psychocoderHPC should the respective files be removed from `hypnos-hzdr` or will they remain there as an example?

**CC'ing** @hightower8083, @wany12


etc/picongpu/hemera-hzdr/k80.tpl
```diff
@@ -0,0 +1,113 @@
...
+# along with PIConGPU.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
+
+# PIConGPU batch script for hemera' SLURM batch system
+
+#SBATCH --partition=!TBG_queue
+# necessary to set the account also to the queue name because otherwise access is not allowed at the moment
+#SBATCH --account=!TBG_queue
```
n01r: The `--account` option is typically used for project allocations but at the moment one has to specify this as well because all users that have access to the k80s are in this "project".

Maybe this can be rather put to the `-C / --constraints` option.

etc/picongpu/hemera-hzdr/k80.tpl
```diff
@@ -0,0 +1,113 @@
...
+.TBG_profile=${PIC_PROFILE:-"~/picongpu.profile"}
+
+# number of available/hosted GPUs per node in the system
+.TBG_numHostedGPUPerNode=8
+
+# required GPUs per node for the current job
+.TBG_gpusPerNode=`if [ $TBG_tasks -gt $TBG_numHostedGPUPerNode ] ; then echo $TBG_numHostedGPUPerNode; else echo $TBG_tasks; fi`
+
+# host memory per gpu
+.TBG_memPerGPU="$((220000 / $TBG_gpusPerNode))"
```
n01r: So `sinfo -lp k80 -N` shows the following

```Shell
garten70@hemera5:~$ sinfo -lp k80 -N
Tue Mar 12 16:41:38 2019
NODELIST   NODES PARTITION       STATE CPUS    S:C:T MEMORY TMP_DISK WEIGHT AVAIL_FE REASON              
kepler020      1       k80        idle   16    2:8:1 240000        0      1   (null) none                
kepler021      1       k80        idle   16    2:8:1 240000        0      1   (null) none                
kepler022      1       k80        idle   16    2:8:1 240000        0      1   (null) none                
kepler023      1       k80        idle   16    2:8:1 240000        0      1   (null) none                
kepler024      1       k80       down*   16    2:8:1 240000        0      1   (null) Not responding      
kepler025      1       k80        idle   16    2:8:1 240000        0      1   (null) none                
kepler026      1       k80        idle   16    2:8:1 240000        0      1   (null) none                
kepler027      1       k80        idle   16    2:8:1 240000        0      1   (null) none                
kepler028      1       k80        idle   16    2:8:1 240000        0      1   (null) none 
```

When I did that for the `gpu` queue I saw that the host memory is `380000`GB but we leave `20000` free and just use `360000`. So I did the same for the `k80` nodes.
Is that fine or should we do it differently?

etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
```diff
@@ -0,0 +1,85 @@
...
+#
+module load zlib/1.2.11
+module load c-blosc/1.14.4
+
+module load adios/1.13.1-cuda100
+module load hdf5-parallel/1.8.20-cuda100
+module load libsplash/1.7.0-cuda100
+
+export CMAKE_PREFIX_PATH=$CMAKE_PREFIX_PATH:$SPLASH_ROOT:$HDF5_ROOT
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$HDF5_ROOT/lib
```
n01r: The two lines above are probably not necessary, right?
I worked based on @psychocoderHPC's profile who loads self-built modules and the `gpu.profile` doesn't actually have them.

etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
```diff
@@ -0,0 +1,85 @@
...
+
+# allocate an interactive shell for one hour
+#   getNode 2  # allocates to interactive nodes (default: 1)
+function getNode() {
+    if [ -z "$1" ] ; then
+        numNodes=1
+    else
+        numNodes=$1
+    fi
+    srun  --time=2:00:00 --nodes=$numNodes --ntasks-per-node=8 --cpus-per-task=2 --gres=gpu:8 --mem=220000 -p gpu --pty bash
```
n01r: Same thing again about the host memory.
We actually have `240000` MB but I'm leaving some free.

etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
```diff
@@ -0,0 +1,85 @@
...
+        numGPUs=1
+    else
+        if [ "$1" -gt 8 ] ; then
+            echo "The maximal number of devices per node is 8." 1>&2
+            return 1
+        else
+            numGPUs=$1
+        fi
+    fi
+    srun  --time=8:00:00 --ntasks-per-node=$(($numGPUs)) --cpus-per-task=$((2 * $numGPUs)) --gres=gpu:$numGPUs --mem=$((27500 * numGPUs)) -p gpu --pty bash
```
n01r: aaaand here is the host memory again: `27500` = `220000` / 8

etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
```diff
@@ -0,0 +1,85 @@
...
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$HDF5_ROOT/lib
+
+module load libpng/1.6.35
+module load pngwriter/0.7.0
+
+# Environment #################################################################
+#
+#export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$BOOST_LIB
+
+export PICSRC=$HOME/src/picongpu
```
wany12: For external users, need to change the path of PICSRC to `export PICSRC=/bigdata/hplsim/external/software/picongpu/workshop`

etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
```diff
@@ -0,0 +1,85 @@
...
+export PYTHONPATH=$PICSRC/lib/python:$PYTHONPATH
+
+# "tbg" default options #######################################################
+#   - SLURM (sbatch)
+#   - "gpu" queue
+export TBG_SUBMIT="sbatch"
+export TBG_TPLFILE="etc/picongpu/hemera-hzdr/k80.tpl"
+
+# allocate an interactive shell for one hour
+#   getNode 2  # allocates to interactive nodes (default: 1)
```
wany12: `getNode` didn't work. The error message is `srun: error: Unable to allocate resources: Requested node configuration is not available`. I change the allocation from gpu to k80, still didn't work

wany12:
Hi, Marco. I run a 2D simulation on hemera and check the output of HDF5. It works well :)
Thanks a lot.

etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
n01r: Thanks for reporting that :+1: I'll have a look

Try `salloc -p k80 -N1 -J k80_interactive -t 2:00:00 -A k80`.
You will also get your job. Just look at `squeue -u $USER`, find out your node and `ssh` to it.
Then you can also build there.

ax3l:
> should the respective files be removed from hypnos-hzdr or will they remain there as an example?

Please feel free to remove the `hypnos-hzdr/k80*` entries. They are documented in previous releases and their manuals if need be.

etc/picongpu/hemera-hzdr/k80.tpl
ax3l: In other profiles, we use a `$proj` variable from `picongpu.profile` for that, check out Titan or Taurus for examples :)

etc/picongpu/hemera-hzdr/k80.tpl
ax3l: It's smart to use slightly less than the full node, even if the batch system does not enforce it.
Leaving maybe 1-2 GByte instead of 20 free is what I would suggest ;-)

etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
ax3l: libSplash and HDF5 should be a module on Hemera (if not, please request them as such)

etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
ax3l: Yep, but leave 1-2 Gbyte free, not 20 :)

etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
ax3l: Nope, that was just for the workshop. Just do the following to become the master of your own PIConGPU source on Hemera:

```bash
# after SSH login on Hemera

cd $HOME
mkdir src
cd src/

git clone https://github.com/ComputationalRadiationPhysics/picongpu.git
# this clones PIConGPU 0.4.3 at the moment
# of course you can
#   git checkout ...
# any other version
```

etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
ax3l: @psychocoderHPC didn't we recently start to change that function for slurm? If I remember correctly, our plan was to `salloc` in the `getNode` function and `echo` the `srun` instructions to the user.

etc/picongpu/hemera-hzdr/k80.tpl
n01r: Then we can do the same for the P100 nodes and reduce the free space from 20GB to 1-2GB as well, right?

psychocoderHPC:
I will review this PR tomorrow. We have currently an alpaka workshop so I am busy.

etc/picongpu/hemera-hzdr/k80.tpl
ax3l: I would think so, if @psychocoderHPC doesn't object.

psychocoderHPC: yes it should be more like 2GiB instead of 20GiB

etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
psychocoderHPC: Yes we changed it on the jülich systems to use salloc but since `srun` on hemera is still not working we can keep it.
If we change it to `salloc` you must call `srun mpiexec ...` to start your mpi environment which is very ugly and non standard.

etc/picongpu/hemera-hzdr/k80.tpl
psychocoderHPC: I think we can stay here with the direct project selection since the name is always `k80` because we have no real projects on this system.

etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
psychocoderHPC: not sure if the admin is now always setting CMAKE_PREFIX_PATH and the LD_LIBRARY_PATH could be required to not fall back to the sequential HDF5 version which is in the system (dependency from SLURM)

psychocoderHPC: @n01r  not sure where you copied this files from but if I look into the master/dev from PIConGPU I do not see both of this lines. So I think they can be removed.

etc/picongpu/hemera-hzdr/k80.tpl
n01r: I'll still call it `$proj` just for consistency.

etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
n01r: It does seem to work now. I had just forgotten the `-A $proj`

n01r: Can I resolve this conversation since both the `getNode` and the `getDevice` function work in the current configuration?

etc/picongpu/hemera-hzdr/k80.tpl
```diff
@@ -83,13 +101,13 @@ export OMPI_MCA_io=^ompio
 
 # test if cuda_memtest binary is available and we have the node exclusive
 if [ -f !TBG_dstPath/input/bin/cuda_memtest ] && [ !TBG_numHostedGPUPerNode -eq !TBG_gpusPerNode ] ; then
-  mpiexec --prefix $MPIHOME -tag-output --display-map -x LIBRARY_PATH -am !TBG_dstPath/tbg/openib.conf --mca mpi_leave_pinned 0 -npernode !TBG_gpusPerNode -n !TBG_tasks !TBG_dstPath/input/bin/cuda_memtest.sh
+  # Run CUDA memtest to check GPU's health
+  mpiexec !TBG_dstPath/input/bin/cuda_memtest.sh
```
n01r: Hey @psychocoderHPC, can you tell me if all the options to the `mpiexec` call are still necessary or not?
The `gpu_picongpu.profile` doesn't have them but the old profiles all do.

```Shell
mpiexec --prefix $MPIHOME -tag-output --display-map -x LIBRARY_PATH -am !TBG_dstPath/tbg/openib.conf --mca mpi_leave_pinned 0 -npernode !TBG_gpusPerNode -n !TBG_tasks !TBG_dstPath/input/bin/cuda_memtest.sh
```

etc/picongpu/hemera-hzdr/k80.tpl
```diff
@@ -83,13 +101,13 @@ export OMPI_MCA_io=^ompio
...
 else
   echo "no binary 'cuda_memtest' available or compute node is not exclusively allocated, skip GPU memory test" >&2
 fi
 
 if [ $? -eq 0 ] ; then
-  mpiexec --prefix $MPIHOME -x LIBRARY_PATH -tag-output --display-map -am !TBG_dstPath/tbg/openib.conf --mca mpi_leave_pinned 0 -npernode !TBG_gpusPerNode -n !TBG_tasks !TBG_dstPath/input/bin/picongpu !TBG_author !TBG_programParams | tee output
+  # Run PIConGPU
+  mpiexec !TBG_dstPath/input/bin/picongpu !TBG_author !TBG_programParams
```
n01r: same here

etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
psychocoderHPC: I will mark it as resolved since and we can change it as soon as our cluster supports srun correctly.
The admin is aware about this broken srun.

etc/picongpu/hemera-hzdr/k80.tpl
psychocoderHPC: This is not required. The confusion was that @n01r copied some files out of my home. 
Offline discussed: this PR based now on gpu.tpl/profile

etc/picongpu/hemera-hzdr/k80.tpl
psychocoderHPC: Without arguments is correct. MPI on hypnos was sometimes not correct compiled and therefore we workaround the issues.

psychocoderHPC:
thx for the update

Issue #2921, opened 2019-03-13
laser produces NaNs if `FOCUS_POS` is at `initPlaneY`


hightower8083:
Current laser implementation fails if  `FOCUS_POS = initPlaneY`(different units but you've got the idea). This happens in calculation of curvature radii (see `include/picongpu/fields/laserProfiles/GaussianBeam.hpp`)
```
float_X const R_y = -focusPos * ( 1.0_X + ( y_R / focusPos )*( y_R / focusPos ) );
```
since `focusPos` can actually be `=0` 

This is easy to fix noting that in you only need `1/R_y`, which can be introduced in a safer way, for example:
```
float_X  R_y_inv = 0.0_X;
if (focusPos != 0.0_X)
{
  R_y_inv = -1.0_X / focusPos / ( 1.0_X + ( y_R / focusPos )*( y_R / focusPos ) );
}
```
and used instead of  `1/R_y`

hightower8083:
can try patching it, if you agree with such modif

sbastrakov:
@hightower8083 it will be very welcome!
I think your suggesting of fixing is perfectly fine. Probably just moving the `focusPos` multiplier inside the brackets to avoid uncertainty should also work.

hightower8083:
Well, the second  term will stay infinite, which is logical -- at focus `R_curve` should go to infinity. I think using `R_curve_inv` is more healthy, cause it never hits any singularity. Will patch it asap

ax3l:
The fix you suggest will introduce some more register usage but no branching, looks good to me. maybe use `constexpr float_X`. One problem I immediately see is, that one cannot compare floats that way easily for a robust fix. Instead, please compare to the "init plane's cell number" which is an integer.

sbastrakov:
@ax3l to our offline discussion. You are right, we do not want to use my suggestion to just transform the expression and have division by zero remaining.

hightower8083:
@ax3l actually with inverse radius used everywhere there is no need for the conditional at all — its z/(z^2+xR^2) and can’t give NaN.

n01r:
this can be closed now, right?

hightower8083:
yep

PR #2922, opened 2019-03-14
Fix laser bug in case focus position is at the init plane

hightower8083:
this PR addresses #2921


hightower8083:
I'm a bit lame with `git` when in comes beyond `commit`, `push` and `pull`.. 
I thick, here I've managed to squash for the last two commits. Let me know if it's OK with you.

sbastrakov:
@hightower8083 looks like the last two commits have been squashed fine. I am not sure why do you need the first commit. In case you actually do not need it, could do rebase again and skip it.

psychocoderHPC:
@hightower8083  I will squash the PR for you

hightower8083:
@psychocoderHPC Thanx, Rene, busy parenting weekend -- no time even to tune a tiny PR =)
@sbastrakov yep, I noticed where I've made a mistake (should have changed both commits to _squash_ instead of _pick_). Will know it better next time ;-)

PR #2923, opened 2019-03-15
Update modules and memory in gpu_picongpu.profile

n01r:
Updated the hemera cluster gpu_picongpu.profile
- updated the modules to builds based on cuda/10.0
- updated the memory requirement to leave only 2GB free on a node, not 20GB


psychocoderHPC:
please also update the coresponding template file: https://github.com/ComputationalRadiationPhysics/picongpu/blob/fa78ada9e6954a009b12d0f477c2b39a8dee88ba/etc/picongpu/hemera-hzdr/gpu.tpl#L60

n01r:
@psychocoderHPC the `.tpl` is now updated as well.

PR #2924, opened 2019-03-17
typo fix in Send/receive buffer warning

hightower8083:
Another revolutionary typo fix

By the way, in order for a user to react to this warning, how you define the _direction_? Foe example, if i get a warning:
```Send/receive buffer for species e is to small (max: 49152, direction: 3, retries: 1)```
should I increase `BYTES_EXCHANGE_Z` or there could be other possibilities?

### DIFF
--- a/include/pmacc/particles/tasks/TaskSendParticlesExchange.hpp
+++ b/include/pmacc/particles/tasks/TaskSendParticlesExchange.hpp
@@ -117,7 +117,7 @@ namespace pmacc
             {
                 std::cerr << "Send/receive buffer for species " <<
                     ParBase::FrameType::getName() <<
-                    " is to small (max: " << maxSize <<
+                    " is too small (max: " << maxSize <<
                     ", direction: " << exchange <<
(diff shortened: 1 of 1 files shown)


ax3l:
Looks good.
yes, we could translate the enum to a human-readable name!

hightower8083:
well, that may sound naive, but if this particular convention is not going to change, may be it can be just copied somewhere accessible by PMacc? or it is a strategic memory concern?

sbastrakov:
@hightower8083 sure, your suggestion is a reasonable way to accomplish it; there are probably several other ways as well. I don't think memory is of any concern here, so should not be any fundamental reason preventing more human-readable output in this place. It just seems the current design maybe does not allow making this change super-easily with a couple of lines of code, but rather requires a little more involved one. As usual, feel free to create an issue!

psychocoderHPC:
> By the way, in order for a user to react to this warning, how you define the _direction_?

@hightower8083 I opened #2925 to extend the message with the name of the direction.

PR #2925, opened 2019-03-19
Name directions in species buffer warnings

psychocoderHPC:
Add the direction name to the warning for multiple send operations in `TaskSendParticlesExchange`.

### DIFF
--- a/include/pmacc/particles/tasks/TaskSendParticlesExchange.hpp
+++ b/include/pmacc/particles/tasks/TaskSendParticlesExchange.hpp
@@ -24,6 +24,7 @@
 
 #include "pmacc/eventSystem/EventSystem.hpp"
 #include "pmacc/assert.hpp"
+#include "pmacc/type/Exchange.hpp"
 
 namespace pmacc
 {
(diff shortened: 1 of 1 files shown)


include/pmacc/particles/tasks/TaskSendParticlesExchange.hpp
```diff
@@ -118,7 +118,7 @@ namespace pmacc
                 std::cerr << "Send/receive buffer for species " <<
                     ParBase::FrameType::getName() <<
                     " is too small (max: " << maxSize <<
-                    ", direction: " << exchange <<
+                    ", direction: " << exchange << " '" << ExchangeTypeNames{}(exchange) << "'" <<
```
ax3l: I think we should add the proper include for this type as well, since it's not defined in this file.

psychocoderHPC: good point, updated

ax3l: might need a git rebase, but the class seems not to be known yet with older GCC (4.9.4):
```
no match for call to ‘(pmacc::ExchangeTypeNames) (uint32_t&)’
```

psychocoderHPC: no it is my fault I need to use the `operator[]`

Issue #2928, opened 2019-03-22
ADIOS: int16-type particle attributes do not compile


BeyondEspresso:
The `voronoiCellId`-attribute of the particle merger is an `int16`, which is not (yet) implemented in `include/picongpu/traits/AdiosToPIC.tpp`. For this reason, the current `dev`-branch does not compile with both particle merger and ADIOS on. It would be an easy fix to also include `int16`-types there. Is there any deeper reason why this should not be done?

Alternatively, I could also modify the type of the `voronoiCellId`-attribute in `speciesAttributes.param` to `int32`, which I have done as a quick workaround for a simulation.

Before I open a pull request, I would like to hear your opinion on which route you prefer.

ax3l:
I don't think ADIOS supports int16 at all, therefore better change the `voronoiCellId` to `int32` ... :-S

ax3l:
fixed with #2929, thanks to @BeyondEspresso!

PR #2929, opened 2019-03-23
Include int16 and uint16 types as traits for ADIOS

BeyondEspresso:
Based on the discussion #2928 this PR fixes a bug, in which picongpu fails to compile when both ADIOS and particle merging are activated. Hence `int16` and `uint16` are now included as traits for ADIOS.


ax3l:
@BeyondEspresso thanks a lot! Can you please also add the new traits to `AdiosToPIC.tpp`, so we are able to read such data as well? :)

BeyondEspresso:
@ax3l Good idea! Done.

PR #2903, opened 2019-02-21
Update taurus-tud profiles for the `ml` partition

steindev:
New modules where installed for the `ml` partition.
Among other things, the MPI machinefile became unnecessary.
I updated the respective `V100_picongpu.profile.example` and `V100.tpl`.
Furthermore, I compiled a script from @ax3l's older scripts
which installs the libraries on which picongpu depends.
The script can be found at (https://gist.github.com/steindev/cc02eae81f465833afa27fc8880f3473#file-picongpu_0-4-3_taurus-tud-sh).
I included the link in the `V100_picongpu.profile.example` as well.

PR #2930, opened 2019-03-26
Fix grammar in all `.profile.example`

steindev:
Follow up pull request fixing a grammar error in all `.profile.example` files. Except the `taurus-tud/V100_picongpu.profile.example` since it will be fixed by the original pull request anyways.

_Requested by @ax3l in https://github.com/ComputationalRadiationPhysics/picongpu/pull/2903_


etc/picongpu/davide-cineca/gpu_picongpu.profile.example
```diff
@@ -1,7 +1,7 @@
 # Name and Path of this Script ############################### (DO NOT change!)
 export PIC_PROFILE=$(cd $(dirname $BASH_SOURCE) && pwd)"/"$(basename $BASH_SOURCE)
 
-# User Information ######################################### (edit those lines)
+# User Information ######################################### (edit the following lines)
```
sbastrakov: Why is it fixed to `those` and not `these`? Then would also be consistent with the later `edit this line`

PR #2933, opened 2019-03-28
Fix k80.tpl and k80_restart.tpl for Hemera

### LINKED ISSUE #2931: Error with k80 partition on hemera
When I try submit a job on the k80 partition on hemera I get following error: "sbatch: error: Batch job submission failed: Invalid account or account/partition combination specified".

I'm using the `k80_picongpu.profile.example` and `k80.tpl` from #2920.

n01r:
Of course environment variables are not replaced in the sbatch script if they are called with a dollar sign in front. 
The right way with TBG is to use an exclamation mark.
This PR fixes #2931.

Thanks @finnolec for reporting this.
**CC'ing** @wany12


n01r:
**CC'ing** @wmles

PR #2941, opened 2019-04-10
Add const getters to ParticleBox

FelixTUD:
Const added to getters in ParticleBox.hpp for compatibility with new ISAAC Renderer

### DIFF
--- a/include/pmacc/particles/memory/boxes/ParticlesBox.hpp
+++ b/include/pmacc/particles/memory/boxes/ParticlesBox.hpp
@@ -141,7 +141,7 @@ public:
     }
 
     HDINLINE
-    FramePtr mapPtr( const FramePtr& devPtr )
+    FramePtr mapPtr( const FramePtr& devPtr ) const
     {
 #ifndef __CUDA_ARCH__
         int64_t useOffset = hostMemoryOffset * static_cast<int64_t> (devPtr.ptr != 0);
@@ -160,7 +160,7 @@ public:
      * @param frame the active frame
      * @return the next frame in the list
      */
-    HDINLINE FramePtr getNextFrame( const FramePtr& frame )
+    HDINLINE FramePtr getNextFrame( const FramePtr& frame ) const
     {
         return mapPtr( frame->nextFrame.ptr );
     }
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
@FelixTUD  Could you please accept the invitation as PIConGPU developer and than force push this PR again. The CI will only check if your changes if you are a developer.

include/pmacc/particles/memory/boxes/ParticlesBox.hpp
```diff
@@ -330,7 +330,7 @@ class ParticlesBox : protected DataBox<PitchedBox<SuperCell<T_Frame>, DIM> >
         return false;
     }
 
-    HDINLINE SuperCellType& getSuperCell( DataSpace<DIM> idx )
+    HDINLINE SuperCellType& getSuperCell( DataSpace<DIM> idx ) const
```
sbastrakov: This method is technically constant, but seems to be logically not constant, as it returns a reference to some internal stuff. In my mind, maybe a better solution is to leave the old method as is and add a const twin, which returns a const reference and then use that new one?

I am also not sure about some others changed, maybe the similar argument applies.

ax3l: If it compiles it's valid C++ and `const` on a method technically only means the method causes no internal side effects.
I think it's ok for now to progress on what he is implementing.

psychocoderHPC:
@FelixTUD What is the state of this PR.

FelixTUD:
> 
> 
> @FelixTUD What is the state of this PR.

@psychocoderHPC 
I haven't tried the new changes proposed by @sbastrakov yet. But @ax3l seems to be fine with this pull request in the current state. Should I still try to change the consts in the logical sense next week?

sbastrakov:
@FelixTUD I think my suggestions are not critical (although stylistically I'm for it). If you feel it is better to delay it or not implement it at all, please feel free to.

psychocoderHPC:
discussed offline: we will keep this PR for now because there are some problems if we like to transform it so that const methods only return const values.

PR #2946, opened 2019-04-17
ADIOS: Avoid unsafe temporary strings

franzpoeschel:
The ADIOS plugin calles `.c_str()` on temporary strings in several places. Introduce the intermediate string value as a variable to avoid use-after-free errors. The [same issue](https://github.com/openPMD/openPMD-api/pull/421) came up in the openPMD API.

### DIFF
--- a/include/picongpu/plugins/adios/ADIOSWriter.hpp
+++ b/include/picongpu/plugins/adios/ADIOSWriter.hpp
@@ -111,11 +111,17 @@ int64_t defineAdiosVar(int64_t group_id,
 {
     int64_t var_id = 0;
 
+    std::string const revertedDimensions =
+        dimensions.revert().toString(",", "");
+    std::string const revertedGlobalDimensions =
+        globalDimensions.revert().toString(",", "");
+    std::string const revertedOffset =
+        offset.revert().toString(",", "");
     var_id = adios_define_var(
         group_id, name, path, type,
-        dimensions.revert().toString(",", "").c_str(),
-        globalDimensions.revert().toString(",", "").c_str(),
-        offset.revert().toString(",", "").c_str()
+        revertedDimensions.c_str(),
+        revertedGlobalDimensions.c_str(),
+        revertedOffset.c_str()
     );
 
     if(compression)
@@ -534,16 +540,15 @@ private:
 
         for( uint32_t c = 0; c < nComponents; c++ )
         {
-            std::stringstream datasetName;
-            datasetName << recordName;
+            std::string datasetName = recordName;
             if (nComponents > 1)
-                datasetName << "/" << name_lookup_tpl[c];
+                datasetName +=  "/" + name_lookup_tpl[c];
 
             /* define adios var for field, e.g. field_FieldE_y */
             const char* path = nullptr;
             int64_t adiosFieldVarId = defineAdiosVar<simDim>(
                     params->adiosGroupHandle,
-                    datasetName.str().c_str(),
+                    datasetName.c_str(),
                     path,
                     adiosType,
                     params->fieldsSizeDims,
@@ -557,11 +562,11 @@ private:
             /* already add the unitSI and further attribute so `adios_group_size`
              * calculates the reservation for the buffer correctly */
             ADIOS_CMD(adios_define_attribute_byvalue(params->adiosGroupHandle,
-                      "position", datasetName.str().c_str(),
+                      "position", datasetName.c_str(),
                       adiosFloatXType.type, simDim, &(*inCellPosition.at(c).begin()) ));
 
             ADIOS_CMD(adios_define_attribute_byvalue(params->adiosGroupHandle,
-                      "unitSI", datasetName.str().c_str(),
+                      "unitSI", datasetName.c_str(),
                       adiosDoubleType.type, 1, &unit.at(c) ));
         }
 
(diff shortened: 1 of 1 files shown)


include/picongpu/plugins/adios/ADIOSWriter.hpp
```diff
@@ -539,11 +539,13 @@ class ADIOSWriter : public IIOBackend
             if (nComponents > 1)
                 datasetName << "/" << name_lookup_tpl[c];
 
+            std::string dataset = datasetName.str( );
```
sbastrakov: Good catch!

As a minor naming consideration. It is currently weird that the new variable for data set name is called `dataset` and the stream used to get this name called `datasetName`. E.g. with `recordName` in the same function it is more natural.

Maybe do a little refactoring with separating [this logic](https://github.com/ComputationalRadiationPhysics/picongpu/blob/f4672c5d84502fecd9a0f653c7b436ddc0df6e22/include/picongpu/plugins/adios/ADIOSWriter.hpp#L537-L540) into a new function, or at least renaming like `datasetName` -> `datasetNameStream` and `dataset` -> `datasetName`?

sbastrakov: Add-up: maybe a new function is even not needed, and smth like that that would suffice (did not check, but to give an idea):
```cpp
std::string datasetName = recordName;
if (nComponents > 1)
    datasetName += "/" + name_lookup_tpl[c];
```

psychocoderHPC:
This PR is not really required because all objects lives long enough. It is only a problem if you try to store the pointer and use it outside of the expression `.str().c_str()`.

https://blog.sensecodons.com/2013/04/dont-let-stdstringstreamstrcstr-happen.html

Never the less we can change it to avoid confusion.

ax3l:
Hm, I would not be sure about the lifetime but maybe you checked the internals of the affected calls already upstream in ADIOS.

Did someone verify with ASAN as I did in openPMD-api? Nevertheless, it's an anti-pattern and should go.
https://github.com/openPMD/openPMD-api/pull/421#issue-242818137

@franzpoeschel can you please make the temporary strings `const`? :)

franzpoeschel:
I've changed the PR according to your suggestions.

Since @psychocoderHPC​'s link suggests not using such constructs even in those cases where they are correct, I would suggest going through with the change.

> Btw, don't [these lines](https://github.com/ComputationalRadiationPhysics/picongpu/blob/f4672c5d84502fecd9a0f653c7b436ddc0df6e22/include/picongpu/plugins/adios/ADIOSWriter.hpp#L116-L118) have the same issue? (I'm not sure, guess depends on how `adios_define_var` uses the values).
> 
> Edit: also in a couple more places in this file.
> Another edit: okay, it is actually fine there.

Those lines do share the same issue or am I missing something? `std::string::c_str()` is called on the temporary created by `pmacc::math::Vector::toString()`, so it would be consistent to change these occurrences, too?

sbastrakov:
@franzpoeschel yes, we should go with the change.
I think changing the [other linked lines](https://github.com/ComputationalRadiationPhysics/picongpu/blob/f4672c5d84502fecd9a0f653c7b436ddc0df6e22/include/picongpu/plugins/adios/ADIOSWriter.hpp#L116-L118) indeed also makes sense.

include/picongpu/plugins/adios/ADIOSWriter.hpp
```diff
@@ -945,8 +944,10 @@ class ADIOSWriter : public IIOBackend
         std::stringstream strFname;
         strFname << restartFilename << "_" << mThreadParams.currentStep << ".bp";
 
+        std::string filename = strFname.str( );
```
sbastrakov: Can `filename` be made `const`?

franzpoeschel: done

franzpoeschel:
The latest commit should cover all occurrences of this in the file.

sbastrakov:
@ax3l it says "Comitter not in authorized team(s)" - could you add @franzpoeschel there?

franzpoeschel:
I fixed the whitespace issue.

sbastrakov:
@ax3l please add to the list of developers.

ax3l:
@ComputationalRadiationPhysics/picongpu-maintainers all of you can add new users to the CI via the team in https://github.com/orgs/ComputationalRadiationPhysics/teams/picongpu-developers/members

ax3l:
@franzpoeschel please rebase and push against the latest `dev` to trigger the CI :)

include/picongpu/plugins/adios/ADIOSWriter.hpp
```diff
@@ -111,11 +111,17 @@ int64_t defineAdiosVar(int64_t group_id,
 {
     int64_t var_id = 0;
 
+    const std::string revertedDimensions =
```
ax3l: we use east-side `const`s :)
```c++
std::string const ...
```

sbastrakov:
@ax3l ah, sorry, I did not consider that.

include/picongpu/plugins/adios/ADIOSWriter.hpp
```diff
@@ -945,8 +950,10 @@ class ADIOSWriter : public IIOBackend
         std::stringstream strFname;
         strFname << restartFilename << "_" << mThreadParams.currentStep << ".bp";
 
+        const std::string filename = strFname.str( );
+
         // adios_read_open( fname, method, comm, lock_mode, timeout_sec )
-        log<picLog::INPUT_OUTPUT > ("ADIOS: open file: %1%") % strFname.str();
+        log<picLog::INPUT_OUTPUT > ("ADIOS: open file: %1%") % filename;
```
ax3l: please add proper spacing while editing this line :)
```suggestion
        log< picLog::INPUT_OUTPUT > ("ADIOS: open file: %1%") % filename;
```

franzpoeschel: Done

include/picongpu/plugins/adios/ADIOSWriter.hpp
```diff
@@ -111,11 +111,17 @@ int64_t defineAdiosVar(int64_t group_id,
 {
     int64_t var_id = 0;
 
+    std::string revertedDimensions const =
```
ax3l: ah nooo... just outermost right of the type
```
std::string const revertedDimensions = ...
```

franzpoeschel: ohh ahmm .. let's agree this didn't happen

PR #2949, opened 2019-04-17
Add pmacc::memory::makeUnique similar to std::make_unique

sbastrakov:
`std::make_unique` is C++14, so we need own C++11 implementation. It has been [discussed and agreed on long time ago](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2669#discussion_r208204183). Now due to recent offline talks to @psychocoderHPC we will use `std::unique_ptr` more actively and so having a creation routine is useful.

New implementation is taken from cppreference (link in comments). It does not handle arrays in the standard way, but I think this is fine for us.

I've also replaced all current creation of unique pointers with the new function, it was really not a lot.

### DIFF
--- a/include/pmacc/memory/MakeUnique.hpp
+++ b/include/pmacc/memory/MakeUnique.hpp
@@ -1,0 +1,48 @@
+/* Copyright 2019 Sergei Bastrakov
+ *
+ * This file is part of PMacc.
+ *
+ * PMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with PMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <memory>
+#include <utility>
+
+
+namespace pmacc
+{
+namespace memory
+{
+
+    /*
+     * Analogue of std::make_unique for C++11, except not disabled for arrays.
+     * Implementation is taken from
+     * https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique
+     */
+    template<
+        typename T,
+        typename ... T_Args
+    >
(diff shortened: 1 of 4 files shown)


include/pmacc/memory/MakeUnique.hpp
```diff
@@ -0,0 +1,49 @@
...
+     * Implementation is taken from
+     * https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique
+     */
+    template<
+        typename T,
+        typename ... T_Args
+    >
+    inline std::unique_ptr< T > makeUnique( T_Args && ... args )
+    {
+        return std::unique_ptr< T >( new T( std::forward< T_Args >( args )...) );
```
psychocoderHPC: could you please add a space before and after `...`

include/pmacc/memory/MakeUnique.hpp
```diff
@@ -0,0 +1,49 @@
...
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with PMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
```
psychocoderHPC: please remove one empty line

include/pmacc/memory/MakeUnique.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include <memory>
+#include <utility>
+
+
```
psychocoderHPC: here the number of empty lines are fine

include/pmacc/memory/MakeUnique.hpp
```diff
@@ -0,0 +1,49 @@
...
+    /*
+     * Analogue of std::make_unique for C++11, except not disabled for arrays.
+     * Implementation is taken from
+     * https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique
+     */
+    template<
+        typename T,
+        typename ... T_Args
+    >
+    inline std::unique_ptr< T > makeUnique( T_Args && ... args )
```
psychocoderHPC: please use an anonymous namespace to support multiple compile units

sbastrakov: I think `inline` is fine here. [ODR](https://en.cppreference.com/w/cpp/language/definition) requires a single definition for non-inline functions, while for inline it is only required that each translation unit has a definition, and having a single definition overall is not required. Basically, that's what we use already with most methods implemented in .hpp files inside a class and thus automatically `inline`.

With the anonymous namespace, iirc, I suggested that at some point, but think here it is not a solution here. At best, it will work the same as inline with every translation unit having a separate, but same, definition. While if we had some internal state (e.g., static local variable inside), it would lead to errors or unexpected behaviour.

sbastrakov:
@psychocoderHPC thanks, fixed and rebased.

PR #2951, opened 2019-04-18
Add DataConnector::consume(), which shares and consumes the input

sbastrakov:
So now there are two scenarios for registering data at `DataConnector`:

- the old `share()` to register and share;
- the new `consume()` to register and consume (the caller gives away ownership to data connector).

The only difference is in ownership. Now it follows [the usual design style](https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/): taking a shared pointer means "I'll share ownership with caller", taking a unique pointer - "I'll take away ownership from caller".

Previously, all calls to `share()` actually wanted just to `consume()` and had to artificially wrap stuff in shared pointers. Which is no longer needed and I've changed it, which also simplifies code a little bit. It turns out, now there are no external calls to `share()` left at all. However, I still kept it as this is a perfectly viable and useful scenario.

### DIFF
--- a/include/picongpu/simulationControl/MySimulation.hpp
+++ b/include/picongpu/simulationControl/MySimulation.hpp
@@ -300,22 +300,8 @@ public:
         namespace nvmem = pmacc::nvidia::memory;
 
         DataConnector &dc = Environment<>::get().DataConnector();
+        initFields(dc);
 
-        // create simulation data such as fields and particles
-        auto fieldB = new FieldB( *cellDescription );
-        dc.share( std::shared_ptr< ISimulationData >( fieldB ) );
-        auto fieldE = new FieldE( *cellDescription );
-        dc.share( std::shared_ptr< ISimulationData >( fieldE ) );
-        auto fieldJ = new FieldJ( *cellDescription );
-        dc.share( std::shared_ptr< ISimulationData >( fieldJ ) );
-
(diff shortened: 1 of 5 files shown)


include/picongpu/simulationControl/MySimulation.hpp
```diff
@@ -803,6 +791,25 @@ class MySimulation : public SimulationHelper<simDim>
...
     bool showVersionOnce;
     bool autoAdjustGrid = true;
+
+private:
+
+    void initFields( DataConnector& dataConnector )
+    {
+        using pmacc::memory::makeUnique;
+        auto fieldB = makeUnique< FieldB >( *cellDescription );
+        dataConnector.consume( std::move( fieldB ) );
```
sbastrakov: Note: this `move` is required, since `fieldB` is an lvalue. If one puts `makeUnique< FieldB >( *cellDescription )` directly instead of `fieldB `, that would work without the move. However, I feel readability is more important here.

sbastrakov:
Offline discussion with @psychocoderHPC : it is maybe better to rename `DataConnector` 's `share` / `consume` -> `register`, `unshare` -> `unregister`. There will be a small problem with overloading + casts, but probably solvable. I'll try later.

sbastrakov:
Ready for review.

PR #2952, opened 2019-04-18
SAXS plugin

ejcjason:
This SAXS plugin simulates the SAXS scattering pattern from the electron position obtained from PIConGPU. Currently, we assume a simple plane wave x-ray laser. In the future, the window function and temporal and transversal laser envelope will be added.

The test results:
image

Edit:
- The color difference between the upper left and right is due to that the left one is using python default color map and the right one is from the png plugin of PIConGPU. 
- The small spikes in the lower right graph are from the limited sampling size and corresponding implied rectangular window function. Because of the low reciprocal space resolution of the lower left FFT, the small spikes are not seen there.
- So basically the result can meet the expectation. I'll make a better graph in the future. 

cc'ing @HighIander


PrometheusPi:
@ejcjason great work! And thank you again for submitting your code :+1:.

Since I was involved in the development, I would suggest @n01r as responsible for the review. Are you fine with that? If not, just assign me again. 

@ax3l Please add @ejcjason to the list of developers that can access the compile suite. 

@HighIander Please also review the physics part of this plugin. 

@psychocoderHPC and @sbastrakov I would be delighted if you could comment on the alpaka parallelisation used. I think there is still some room for improvement there (which I could also directly include in the radiation plugin 😄 )

PrometheusPi:
Regarding the spikes between the main signals above: they come from the missing window function and can be avoided in an upgrade - do not worry about these now.

PrometheusPi:
As discussed before, the `validate.ipynb` probably should not be included in the code. @ax3l What do you think?

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,481 @@
+/* Copyright 2013-2018 Axel Huebl, Heiko Burau, Rene Widera, Richard Pausch,
```
sbastrakov: 2018 -> 2019

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,481 @@
...
+   
+
+    virtual ~Saxs()
+    {
+    }
+
+    /**
+     * This function represents what is actually calculated if the plugin
+     * is called. Here, one only sets the particles pointer to the data of
+     * the latest time step and calls the 'calculateRadiationParticles'
```
PrometheusPi: I think this comment is outdated: It should read `calculateSAXS`.

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,481 @@
...
+// {
+//     unsigned int n_q = n_qx * n_qy * n_qz;
+//     return n_q;
+// }
+    // template<typename gb>
+    // void createQ(gb& q, float_X q_min, float_X q_step, unsigned int n_q)
+    // {   
+    //     for (int i = 0; i < n_q; i++)
+    //         q[i] = q_min + q_step * i;
+    // }
```
PrometheusPi: Please remove the unneeded code.

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,481 @@
...
+
+
+    /**
+     * The plugin is loaded on every MPI rank, and therefor this function is
+     * executed on every MPI rank.
+     * One host with MPI rank 0 is defined to be the master.
+     * It creates a folder where all the
+     * results are saved and, depending on the type of radiation calculation,
```
PrometheusPi: I think this is an old comment - perhaps it should be slightly reworded for the SAXS plugin.

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,481 @@
...
+            // const unsigned int n_qx = getNq(qx_min,qx_max,qx_step);
+            // const unsigned int n_qy = getNq(qy_min,qy_max,qy_step);
+            // const unsigned int n_qz = getNq(qz_min,qz_max,qz_step);
+            isMaster = reduce.hasResult(mpi::reduceMethods::Reduce());
+            // qx = new gridBuffers<float_X,DIM1>(DataSpace<DIM1> (n_qx));
+            // qy = new gridBuffers<float_X,DIM1>(DataSpace<DIM1> (n_qy));
+            // qz = new gridBuffers<float_X,DIM1>(DataSpace<DIM1> (n_qz));
+            // createQ(qx, qx_min, qx_step);
+            // createQ(qy, qy_min, qy_step);
+            // createQ(qz, qz_min, qz_step);
```
PrometheusPi: Please remove unneeded code

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,481 @@
...
+                i_y = (i / n_qz) % n_qy ;
+                i_x = i / (n_qz * n_qy) ;
+                q[0] = q_min[0] + q_step[0] * i_x;
+                q[1] = q_min[1] + q_step[1] * i_y;
+                q[2] = q_min[2] + q_step[2] * i_z;
+                ofile << q[0] << " " << q[1] << " " << q[2] << " " << intensity[i][0] << "\n";
+            }
+
+            // if (ofile.fail())
+            //     std::cerr << "Error on flushing file [" << name << "]. " << std::endl;
```
PrometheusPi: Please remove commented code.

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,481 @@
...
+            std::cerr << "Can't open file [" << name << "] for output, disable plugin output.\n";
+            isMaster = false; // no Master anymore -> no process is able to write
+        }
+        else
+        {
+            ofile <<  "Number of particles:"  << " " << np_master.x() << "\n";
+            ofile <<  "Number of macro particles:"  << " " << nmp_master << "\n";
+
+            /* if (ofile.fail())
+                std::cerr << "Error on flushing file [" << name << "]. " << std::endl; */
```
PrometheusPi: Please remove this code part.

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,481 @@
...
+    sumfsinkr->deviceToHost();
+    np->deviceToHost();
+    nmp->deviceToHost();
+    __getTransactionEvent().waitForFinished();
+  }
+
+
+
+
+  /** combine radiation data from each CPU and store result on master
```
PrometheusPi: Probably replacing radiation by SAXS would improve this comment.

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,481 @@
...
+             &nmp_master,
+             nmp->getHostBuffer().getBasePointer(),
+             1,
+             mpi::reduceMethods::Reduce()
+             );
+        
+      // Calculate intensity on master
+      if (isMaster)
+      {
+        // printf("np_master = %f\n", np_master.x());
```
PrometheusPi: Please remove this debug `printf`.

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,481 @@
...
+        for ( unsigned int i = 0; i < n_q; i++)
+            intensity_master[i] = (sumfcoskr_master[i]*sumfcoskr_master[i]+sumfsinkr_master[i]*sumfsinkr_master[i])/np_master.x();
```
PrometheusPi: These lines are quite long - could you please add a couple newlines to ease reading your code.

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,481 @@
...
+      copyIntensityDeviceToHost();
+      collectIntensityOnMaster();
+  }
+
+
+ 
+
+
+  /**
+   * This functions calls the radiation kernel. It specifies how the
```
PrometheusPi: Please update this comment slightly to fit to the SAXS plugin:
Mainly: `radiation` -> `SAXS`

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,481 @@
...
+   * |   |o      o    |   |o      o    |                       |
+   * |   |o      o    |   |o      o    |                       |
+   * |   |th1    th2  |   |th1    th2  |                       |
+   * |   --------------   --------------                       |
+   * -----------------------------------------------------------
+   *
+   * !!! The TEMPLATE parameter is not used anymore.
+   * !!! But the calculations it is supposed to do is hard coded in the
+   *     kernel.
+   * !!! THIS NEEDS TO BE CHANGED !!!
```
PrometheusPi: @psychocoderHPC What about this issue?

psychocoderHPC:
@ejcjason could you please add on the right in this pull request the checkbox that maintainer can modify your branch. 
I had a commit with basic white space changes which also removes comments copy-pasted from the radiation kernel. I like to push those changes directly to save some time and concentrate on the algorithm.

psychocoderHPC:
@ejcjason I invited you to the PIConGPU developers, this allows our CI to test your code.

ejcjason:
@psychocoderHPC I have checked the checkbox, does it work now?
image

n01r:
@PrometheusPi I wasn't really free up to now but yea, I'd do the review of this one. :)
@ejcjason Do you want to push another version with the remarks implemented until now before I start?

ejcjason:
@n01r Yes, if it's fine for you. I plan to implement them tomorrow.

psychocoderHPC:
I will push my changes thia evening. sry for the delay

psychocoderHPC:
I pushed my changes https://github.com/ComputationalRadiationPhysics/picongpu/pull/2952/commits/92a8315be13945493f5831e30a5792cd96f66280
@ComputationalRadiationPhysics/picongpu-maintainers please feel free to go on with the review. I am not at work until 6th May

n01r:
> @n01r Yes, if it's fine for you. I plan to implement them tomorrow.

@ejcjason Sure, that's fine by me :)

ejcjason:
@n01r I've refined some comments. Please start review whenever you want. Thanks.

n01r:
Great work @ejcjason, thanks!

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,303 @@
+/* Copyright 2013-2018 Axel Huebl, Heiko Burau, Rene Widera, Richard Pausch,
+ *                     Klaus Steiniger, Felix Schmitt, Benjamin Worpitz
```
ax3l: @ejcjason pls do not forget to add your name :)

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,303 @@
...
+                            // set z component to zero in case of simDim==DIM2
+                            r_s[linearIdx][2] = 0.0;
+                            // run over all components and compute gobal position
+                            for (int i = 0; i < simDim; ++i)
+                                r_s[linearIdx][i] =
+                                    (float_X(globalPos[i]) + pos[i]) *
+                                    cellSize[i];
+
+                            /* get macro-particle weighting
+                                         *
```
ax3l: confusing indentation

include/picongpu/plugins/saxs/jupyter/validate.ipynb
```diff
@@ -0,0 +1,4975 @@
+{
```
ax3l: @ejcjason can you please add this file as python modules in `lib/python/picongpu/plugins/data` (data parsing/reading) and `lib/python/picongpu/plugins/plot_mpl` (plotting)?

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,397 @@
...
+///////////////////////////////////////////////////////////////////////////////////////////////
+///////////////////////////////  SAXS Plugin Class  ////////////////////////////////////
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+template<class ParticlesType>
+class Saxs : public ISimulationPlugin
+{
+private:
+
+    typedef MappingDesc::SuperCellSize SuperCellSize;
```
ax3l: please use `using SuperCellSize = MappingDesc::SuperCellSize;`

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,397 @@
...
+    std::string speciesName;
+    std::string pluginName;
+    std::string pluginPrefix;
+    std::string filename_prefix;
+    float3_64 q_min, q_max, q_step;
+    unsigned int n_qx, n_qy, n_qz, n_q;
+    float1_64 *sumfcoskr_master;
+    float1_64 *sumfsinkr_master;
+    float1_64 *intensity_master;
+    float1_X np_master; // Number of particles
```
ax3l: probably a `float1_64` is more precise for such reduces as well.

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,397 @@
...
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+template<class ParticlesType>
+class Saxs : public ISimulationPlugin
+{
+private:
+
+    typedef MappingDesc::SuperCellSize SuperCellSize;
+
+    GridBuffer<float1_64, DIM1> *sumfcoskr;
```
ax3l: please add Doxygen strings to the members so we know what they mean :)

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,397 @@
...
+    void pluginLoad()
+    {
+        if(!notifyPeriod.empty())
+        {
+            /*only rank 0 create a file*/
+            isMaster = reduce.hasResult(mpi::reduceMethods::Reduce());
+            n_q = n_qx * n_qy * n_qz;
+            sumfcoskr = new GridBuffer<float1_64, DIM1 > (DataSpace<DIM1> (n_q));
+            sumfsinkr = new GridBuffer<float1_64, DIM1 > (DataSpace<DIM1> (n_q));
+            np = new GridBuffer<float1_X, DIM1 > (DataSpace<DIM1> (1)); //create one float on GPU and host
```
ax3l: create -> allocate

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,397 @@
...
+            nmp = new GridBuffer<int64_t, DIM1 > (DataSpace<DIM1> (1)); //create one int on GPU and host
```
ax3l: create -> allocate

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,397 @@
...
+
+            if(isMaster)
+            {
+                fs.createDirectory("saxsOutput");
+                fs.setDirectoryPermissions("saxsOutput");
+            }
+        }
+    }
+
+    void writeQintensity(float1_64* intensity, std::string name)
```
ax3l: please add a Doxygen line for the use of the member function and its parameters

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,397 @@
...
+                q[1] = q_min[1] + q_step[1] * i_y;
+                q[2] = q_min[2] + q_step[2] * i_z;
+                ofile << q[0] << " " << q[1] << " " << q[2] << " " << intensity[i][0] << "\n";
+            }
+
+            ofile.close();
+        }
+    }
+
+    void writeLog(float1_X np_master,int64_t nmp_master, std::string name)
```
ax3l: please add a Doxygen line for the use of the member function and its parameters

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,397 @@
...
+        // collect data GPU -> CPU -> Master
+        copyIntensityDeviceToHost();
+        collectIntensityOnMaster();
+    }
+
+  /**
+   * This functions calls the radiation kernel. It specifies how the
+   * calculation is parallelized.
+   **/
+
```
ax3l: please remove the empty newline here

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,303 @@
...
+                );
+
+                __syncthreads(); // wait till every thread has loaded its particle data
+
+                // run over all q for this thread
+                for (int idx = workerIdx; idx < n_q; idx += T_numWorkers)
+                {
+
+                    // Initialize in register memory
+                    float1_64 sumfcoskr_r = 0;
```
ax3l: 0.

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,303 @@
...
+                    float1_64 sumfsinkr_r = 0;
```
ax3l: 0.

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,303 @@
...
+                    float3_64 q_r;
+                    float_64 dotkr;
+
+                    int i_z = idx % n_qz;
+                    int i_y = (idx / n_qz) % n_qy;
+                    int i_x = idx / (n_qz * n_qy);
+
+                    q_r[0] = (q_min[0] + q_step[0] * i_x);
```
ax3l: unnecessary brackets or wrongly placed?

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,303 @@
...
+                    q_r[1] = (q_min[1] + q_step[1] * i_y);
+                    q_r[2] = (q_min[2] + q_step[2] * i_z);
+
+                    /// Particle loop: thread runs through loaded particle data
+                    for (int j = 0; j < particlesInFrame; ++j)
+                    {
+                        dotkr = (q_r[0] * r_s[j][0] + q_r[1] * r_s[j][1] + q_r[2] * r_s[j][2])*1e10*UNIT_LENGTH; //in PIConGPU unit
```
ax3l: `1e10` is a magic constant and of type `int`.
please store it in a temporary `const` variable to explain it (probably something in Ångström?)

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,303 @@
...
+
+                /* First threads starts loading next frame of the super-cell:
+                 *
+                 * Info:
+                 *   The calculation starts with the last SuperCell (must not be full filled)
+                 *   all previous SuperCells are full with particles
+                 */
+                particlesInFrame = frameSize;
+                frame = pb.getPreviousFrame(frame);
+                if (workerIdx == 0) //TODO: master thread, not the master rank of MPI?
```
ax3l: this is the master thread of a block on the device

include/picongpu/plugins/saxs/Saxs.hpp
n01r: also `writeQIntensity` for correct camel case

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,397 @@
...
+    {
+        sumfcoskr->deviceToHost();
+        sumfsinkr->deviceToHost();
+        np->deviceToHost();
+        nmp->deviceToHost();
+        __getTransactionEvent().waitForFinished();
+    }
+
+    /** combine radiation data from each CPU and store result on master
+     *  copyRadiationDeviceToHost() should be called before */
```
n01r: please update the comments

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,397 @@
...
+            __delete(sumfcoskr);
+            __delete(sumfsinkr);
+            __delete(np);
+            __delete(nmp);
+            __deleteArray(intensity_master);
+            CUDA_CHECK(cudaGetLastError());
+        }
+    }
+
+    /** Method to copy data from GPU to CPU */
```
n01r: please change to single line comment (`// Method ...`) or add more description to this method :)

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,397 @@
...
+        np->getDeviceBuffer( ).setValue( 0.0 );
+        nmp->getDeviceBuffer( ).setValue( 0.0 );
+
+        // calculate q_step
+        q_step[0] = (q_max[0]-q_min[0])/n_qx;
+        q_step[1] = (q_max[1]-q_min[1])/n_qy;
+        q_step[2] = (q_max[2]-q_min[2])/n_qz;
+
+
+        //   PIC-like kernel call of the saxs kernel
```
n01r: please reduce to just one white space

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,303 @@
...
+#include <pmacc/dimensions/DataSpaceOperations.hpp>
+#include <pmacc/mappings/kernel/AreaMapping.hpp>
+
+#include "picongpu/plugins/radiation/parameters.hpp"
+#include "picongpu/plugins/radiation/check_consistency.hpp"
+#include "picongpu/plugins/radiation/particle.hpp"
+#include "picongpu/plugins/radiation/amplitude.hpp"
+#include "picongpu/plugins/radiation/calc_amplitude.hpp"
+#include "picongpu/plugins/radiation/windowFunctions.hpp"
+#include "picongpu/plugins/radiation/GetRadiationMask.hpp"
```
n01r: Are the above includes from `picongpu/plugins/radiation/` still necessary or can some be removed?

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,303 @@
...
+     *  - The number of threads per block is equal to the number of cells per
+     *    super cells which is also equal to the number of particles per frame
+     *
+     * The procedure starts with calculating unique ids for the threads and
+     * initializing the shared memory.
+     * Then a loop over all super cells starts.
+     * Every thread loads a particle from that super cell and calculates its
+     * scattering structure factor.
+     * For every Particle
+     * exists therefor a unique space within the shared memory.
```
n01r: therefor**e**

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,303 @@
...
+
+                } // end q (frequency) loop
+
+                // wait till all  for this super cell are done
+                __syncthreads();
+
+                /* First threads starts loading next frame of the super-cell:
+                 *
+                 * Info:
+                 *   The calculation starts with the last SuperCell (must not be full filled)
```
n01r: must not be full filled --> does not have to be fully filled

sbastrakov:
We had several offline discussions on parallelism in this plugin, with @ejcjason , @psychocoderHPC , @PrometheusPi and me participating in different combinations. Here is a summary:

- Currently the plugin is parallelized inside a block, but not between blocks (and launched with 1 block), which is not ideal.
- Currently computation of different frequencies is done by different threads inside a block in parallel, as for each frequency it is indepent (however, there is a dependency with collectively loading particles into shared memory).
- The most straightforward way to use several blocks is to use the same scheme and distribute frequencies between threads globally, not just inside a block. Then each thread can just compute a single frequency, store intermediate results in its local variable and write back to global memory in the end. The kernel structure stays basically the same, with [this loop](https://github.com/ComputationalRadiationPhysics/picongpu/blob/513c4193c13e28ebe797eb434f61f1aa11145726/include/picongpu/plugins/saxs/Saxs.kernel#L251) disappearing. The number of blocks at kernel launch becomes the upper integer part of `n_q / numWorkers`. Of course, some care needs to be taken of thread indexes not going outside of arrays and writing the `np` result. We decided to try this approach first, as, again, the change seems rather quick to do, and compare performance to the current one.
- There is another idea of parallelizing the loop over supercells with blocks. This has an advantage that each thread does not need to go through all particles any more, but will require a bit more work and synchronization. It is hard to say which should be better. We can maybe try this later, but not before the first one is implemented and evaluated.

n01r:
@sbastrakov More parallelization definitely sounds nice and improving performance is always good.
You also said it wouldn't take too much time. I would, however, opt for just the implementation of reviewer comments, then maybe an afterward squash and the merging of this pull request.
That way the pull request doesn't get too long and we already have a working version in the mainline.
Improving the parallelization can be a separate pull request.
@ejcjason What is your opinion on this?

psychocoderHPC:
@n01r Point one from @sbastrakov  `Currently the plugin is parallelized inside a block, but not between blocks (and launched with 1 block), which is not ideal.` must be solved before we can merge the PR. The easiest solution is to have one thread per frequency to avoid that we only use one block.

@ejcjason, please work in the suggestions from the reviews and then do a small benchmark of the current code (one block) and then make a small improvement to use multiple blocks (if you need help you know where you find us^^)

PrometheusPi:
@psychocoderHPC I agree - otherwise we do not use the accelerators efficiently. However, the SAXS scattering just works at a **single** frequency (= q- or k-value) (I think @sbastrakov meant directions not frequencies in his last post.).  Thus a second parallelization could only be over particles or by splitting the parallelization over directions. I would vote for the first, since this would allow an easy adaption in both the original radiation plugin and the upcoming CTR plugin. But the latter might be easier to implement.

sbastrakov:
@PrometheusPi sorry if I introduced confusion, by "number of frequencies" I meant what is denoted `n_q` in the code. Abstractly I am also for the option of parallelizing over supercells and maybe along the way providing some utilities so that is done easier in the future. However, the other option should be really way easier to do now.

ejcjason:
@n01r @psychocoderHPC , I'll finish the implementation of reviewer comments first.

include/picongpu/plugins/saxs/Saxs.hpp
ejcjason: > also `writeQIntensity` for correct camel case

Replace it with `writeIntensity` by just deleting the useless 'Q'.

include/picongpu/plugins/saxs/Saxs.kernel
ejcjason: unnecessary.

include/picongpu/plugins/saxs/Saxs.kernel
ejcjason: All can be removed. Maybe the `picongpu/plugins/radiation/windowFunctions.hpp` will be added back for future version.

include/picongpu/plugins/saxs/Saxs.kernel
ejcjason: Defined a const variable `float_64 const meter2angstrom = 1e10;`

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -110,165 +100,161 @@ struct KernelSaxs
...
+            mapper.getGridSuperCells() - 2 * guardingSuperCells);
 
         // get absolute number of relevant super cells
         int const numSuperCells = superCellsCount.productOfComponents();
 
         /* go over all super cells on GPU
          * but ignore all guarding supercells
          */
-        for (int super_cell_index = 0; super_cell_index <= numSuperCells; ++super_cell_index)
+        for (int super_cell_index = 0; super_cell_index <= numSuperCells;
```
sbastrakov: Should this really be `<=` ? (I would expect `<` )

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,414 @@
...
+            ofile << n_q << "\n";
+            ofile << "# qx qy qz intensity \n";
+            for (unsigned int i = 0; i < n_q; i++)
+            {
+                i_z = i % n_qz;
+                i_y = (i / n_qz) % n_qy;
+                i_x = i / (n_qz * n_qy);
+                q[0] = q_min[0] + q_step[0] * i_x;
+                q[1] = q_min[1] + q_step[1] * i_y;
+                q[2] = q_min[2] + q_step[2] * i_z;
```
sbastrakov: I think this can be replaced with a vector operation `q = q_min + q_step * float3_64( i_x, i_y, i_z );`

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,414 @@
...
+                ofile << q[0] << " " << q[1] << " " << q[2] << " "
```
sbastrakov: There is an overloaded `operator <<` for vectors. (Not sure if you did not use it intentionally).

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,414 @@
...
+        // initialize variables with zero
+        sumfcoskr->getDeviceBuffer().setValue(0.0);
+        sumfsinkr->getDeviceBuffer().setValue(0.0);
+        np->getDeviceBuffer().setValue(0.0);
+        nmp->getDeviceBuffer().setValue(0.0);
+
+        // calculate q_step
+        q_step[0] = (q_max[0] - q_min[0]) / n_qx;
+        q_step[1] = (q_max[1] - q_min[1]) / n_qy;
+        q_step[2] = (q_max[2] - q_min[2]) / n_qz;
```
sbastrakov: A vector variant would be `q_step = (q_max - q_min) / float3_64( n_qx, n_qy, n_qz );`

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,414 @@
...
+          speciesName(ParticlesType::FrameType::getName()),
+          pluginPrefix(speciesName + std::string("_saxs")),
+          filename_prefix(pluginPrefix), sumfcoskr(nullptr), sumfsinkr(nullptr),
+          np(nullptr), nmp(nullptr), cellDescription(nullptr), isMaster(false),
+          currentStep(0)
+    {
+        Environment<>::get().PluginConnector().registerPlugin(this);
+    }
+
+    virtual ~Saxs() {}
```
sbastrakov: Please use `= default;` instead of `{}`

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,414 @@
...
+     * 4*pi*sin(theta)/lambda, where 2theta is the angle
+     * between scattered and incident beam
+     **/
+    float3_64 q_min, q_max, q_step;
+    //! Number of scattering vectors
+    unsigned int n_qx, n_qy, n_qz, n_q;
+
+    float1_64 *sumfcoskr_master;
+    float1_64 *sumfsinkr_master;
+    float1_64 *intensity_master;
```
sbastrakov: I think arrays `sumfcoskr_master`, `sumfsinkr_master` and `intensity_master` are a little better as `std::vector`s. Then you can get a plain pointer, e.g. for reduction, with calling `.data()` method.
Also, not sure why these are `float1_64` and not just `float_64`; but not against it if there are reasons.

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,414 @@
...
+    using SuperCellSize = MappingDesc::SuperCellSize;
+
+    //! The real part of structure factor
+    GridBuffer<float1_64, DIM1> *sumfcoskr;
+    //! The imaginary part of structure factor
+    GridBuffer<float1_64, DIM1> *sumfsinkr;
+    //! Number of real particles
+    GridBuffer<float1_64, DIM1> *np;
+    //! Number of macro particles
+    GridBuffer<int64_t, DIM1> *nmp;
```
sbastrakov: (not that important)
I think these 4 pointers are a little better as `std::unique_ptr`s. There is `pmacc::memory::makeUnique()` to help create them, but it was added to dev recently, so would need to merge first.

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,414 @@
...
+            globalOffset,
+            currentStep,
+            *cellDescription,
+            subGrid.getGlobalDomain().size,
+            q_min,
+            q_max,
+            q_step,
+            n_qx,
+            n_qy,
+            n_qz,
```
sbastrakov: This is purely subjective, so not necessary to take into account at all. Why are `n_qx, n_qy, n_qz` treated as 3 separate numbers, while `q_min`, `q_max` and `q_step` as vectors? For me it seems logical to also put the `n_...` variables into a vector here and in the kernel. Also, in case `n_q` is always a product of components, it can be computed in the kernel and does not need to be passed.

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,291 @@
...
+#include "sys/stat.h"
+
+namespace picongpu
+{
+
+/** calculate the scattering of a species
+ *
+ * @tparam T_numWorkers number of workers
+ */
+template <uint32_t T_numWorkers> struct KernelSaxs
```
sbastrakov: Please start `struct KernelSaxs` at a new line.

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,291 @@
...
+     *
+     * The procedure starts with calculating unique ids for the threads and
+     * initializing the shared memory.
+     * Then a loop over all super cells starts.
+     * Every thread loads a particle from that super cell and calculates its
+     * scattering structure factor.
+     * For every Particle
+     * exists therefore a unique space within the shared memory.
+     * After that, a thread calculates for a specific scattering vector of all
+     * particles.
```
sbastrakov: This kernel has quite a lot of parameters. Some can be not obvious for a reader. Please extend this doxygen comment with parameter description (like other functions do it with `@param` in the comments).

include/picongpu/plugins/saxs/Saxs.hpp
```diff
@@ -0,0 +1,414 @@
...
+        // calculate q_step
+        q_step[0] = (q_max[0] - q_min[0]) / n_qx;
+        q_step[1] = (q_max[1] - q_min[1]) / n_qy;
+        q_step[2] = (q_max[2] - q_min[2]) / n_qz;
+
+        // PIC-like kernel call of the SAXS kernel
+        PMACC_KERNEL(
+            KernelSaxs< numWorkers >{}
+        )(
+            1,
```
sbastrakov: Suggested change of parallelism (1):
Here compute the number of blocks as `(n_q + numWorkers - 1) / numWorkers`
(has to be done together with changes (2), (3), etc. later in the comments).

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,291 @@
...
+        namespace po = boost::program_options;
+
+        constexpr uint32_t frameSize =
+            pmacc::math::CT::volume<SuperCellSize>::type::value;
+        constexpr uint32_t numWorker = T_numWorkers;
+
+        using FrameType = typename ParBox::FrameType;
+        using FramePtr = typename ParBox::FramePtr;
+
+        uint32_t const workerIdx = threadIdx.x;
```
sbastrakov: Suggested change of parallelism(2):
Replace this line with `uint32_t const workerIdx = blockIdx.x * blockDim.x + threadIdx.x;` to get a thread index among all blocks (instead of among the current block).

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,291 @@
...
+                });
+
+                __syncthreads(); // wait till every thread has loaded its
+                                 // particle data
+
+                // meter -> Angstrom
+                float_64 const meter2angstrom = 1e10;
+
+                // run over all q for this thread
+                for (int idx = workerIdx; idx < n_q; idx += T_numWorkers)
```
sbastrakov: Suggested change of parallelism (3): remove this line, the loop body stays here with `idx` replaced with `workerIdx`. (now we will just do a single iteration of the old loop).

sbastrakov:
@ejcjason thanks for the code update. I made some comments. The only important ones are three comments about parallelism. All others are just stylistic and can be ignored.
Also our CI fails to build the code, not sure if it is CI problem or an actual code issue.

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,291 @@
...
+            // go to next supercell
+            while (frame.isValid())
+            {
+                /* since a race condition can occur if "continue loop" is
+                 * called, all threads must wait for the selection of a new
+                 * frame until all threads have evaluated "isValid"
+                 */
+                __syncthreads();
+
+                ForEachIdx<IdxConfig<1, numWorker>> onlyMaster{workerIdx};
```
sbastrakov: Suggested change of parallelism (4): replace `workerIdx` with `threadIdx.x` (or make a new variable for it and use it here).

include/picongpu/plugins/saxs/Saxs.kernel
```diff
@@ -0,0 +1,291 @@
...
+                    np_s = 0;
+                    nmp[0] += particlesInFrame;
+                });
+
+                __syncthreads();
+
+                using ParticleDomCfg = IdxConfig<frameSize, numWorker>;
+
+                // loop over all particles in the frame
+                ForEachIdx<ParticleDomCfg> forEachParticle{workerIdx};
```
sbastrakov: Suggested change of parallelism (5): replace `workerIdx` with `threadIdx.x` (or make a new variable for it and use it here).

ejcjason:
> Thanks a lot @ejcjason for this pull request!
> 
> Please implement the reviewer changes that have been requested so far.
> Could you also please write a documentation section for this plugin by adding a new file to `docs/source/usage/plugins/`? That would be tremendously helpful for us and future users!

Added!

include/picongpu/plugins/saxs/Saxs.kernel
ejcjason: @sbastrakov I would also expect like you. This part is copied from radiation plugin. Maybe @PrometheusPi can help explain this a little bit?

sbastrakov: I think `<` should be used. Maybe having `super_cell_index == numSuperCells` results in accessing an existing supercell in the guard and so does not lead to a crash.

PrometheusPi:
@sbastrakov @psychocoderHPC 
The test used by @ejcjason initialized a regular density pattern (in blue-yellow or PIConGPU-png-colors red-black). They give a distinct and analytically predictable scattering pattern. (The Fourier Transform of the (static) density.) - given below. @HighIander can probably provide more information on suitable test cases.

ax3l:
@n01r @PrometheusPi please do not forget this PR

sbastrakov:
@ax3l it's actually on me for now.

psychocoderHPC:
this PR will be handled by #2973 now

PR #2961, opened 2019-05-13
CUDA 9.0+, GCC 5.1+, Boost 1.65.1+

ax3l:
Increase the dependencies to CUDA 9.0+, which in turn requires Boost 1.65.1+ to run.

With that prepare, we also update our CI to run on GCC 5.1+ which will make the transition to C++14 possible soon. While GCC 4.9 is technically still okay as long as we stick with C++11, we deprecate support and continuous CI for it to move forward for the aforementioned reasons.

Slightly increased the transitive HDF5 requirement to same version as in openPMD-api.

### DIFF
--- a/include/pmacc/PMaccConfig.cmake
+++ b/include/pmacc/PMaccConfig.cmake
@@ -305,7 +305,7 @@ endif(MPI_CXX_FOUND)
 # Find Boost
 ################################################################################
 
-find_package(Boost 1.62.0 REQUIRED COMPONENTS filesystem system math_tr1)
+find_package(Boost 1.65.1 REQUIRED COMPONENTS filesystem system math_tr1)
 if(TARGET Boost::filesystem)
     set(PMacc_LIBRARIES ${PMacc_LIBRARIES} Boost::boost Boost::filesystem
                                            Boost::system Boost::math_tr1)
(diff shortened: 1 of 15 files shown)


include/pmacc/PMaccConfig.cmake
```diff
@@ -346,25 +334,10 @@ elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DBOOST_NO_CXX11_SMART_PTR")
 endif()
 
-# Boost 1.64.0 is broken with CUDA 8.0 (nvcc) and C++11
-#   https://github.com/ComputationalRadiationPhysics/picongpu/issues/2048
-#   fixed in CUDA 9.0 (ticket 1928813)
-if( ("${ALPAKA_CUDA_COMPILER}" STREQUAL "nvcc") AND
-    (Boost_VERSION EQUAL 106400) AND
-    (CUDA_VERSION VERSION_EQUAL 8.0) )
```
ax3l: TODO: hm, we don't really check the CUDA version here since we take whatever alpaka gives us....

psychocoderHPC: Discussed offline with @ax3l : We will drop the CUDA 8 support fully. IN that case it is fine to remove the workaround.

ax3l: We should still add a `message(ERROR ...)` on the CUDA version during CMake time :)

PR #2972, opened 2019-06-04
Refactor cell types

sbastrakov:
This PR makes several refactoring-related changes to cell types:

- renaming and proper namespaces for the cell types themselves
- renaming of `NummericalCellType` (with the double `m` typo) in field solvers to `CellType`
- a new header and type `picongpu::fields::CellType` to be used outside of the field solvers. This aims to reduce dependencies from field solvers in the remaining code, that just needs the cell type. Of course, now eventually the fields solvers are included anyways, but dependencies are more clear and it would be easier to separate in the future. Discussed offline with @psychocoderHPC to have this type in `fieldSolver.unitless` instead so that it is available without additional includes, did not come to an agreement on that.


psychocoderHPC:
I will wait until Friday with merging so @ax3l who implemented `NummericalCellType` can also have a view.

PR #2976, opened 2019-06-07
Remove unused pmacc::traits::GetEmptyDefaultConstructibleType

sbastrakov:
This is probably some legacy, that is no longer needed. I stumbled on it while figuring out the use of stuff in `pmacc::expressions`, which would also become unused and be removed in a following PR after this one is merged.


psychocoderHPC:
Yes it was a workaround for some CUDA restrictions with empty constructors. THX for removing it

PR #2977, opened 2019-06-07
Remove template parameter for initialization method of Pointer and FramePointer

sbastrakov:
Taking into account #2976, there will be no use for alternatives here. So this PR removes this template parameter and now the pointers are always default-initialized with `nullptr`. This does not change the behavior of the existing code, as it has always been the set to zero policy anyways.

Edit: the first commit is the same as in #2976 (merged already at the time of the edit), please review only the second.

- [x] do not merge before #2976 is merged.

### DIFF
--- a/include/pmacc/particles/memory/dataTypes/Pointer.hpp
+++ b/include/pmacc/particles/memory/dataTypes/Pointer.hpp
@@ -21,50 +21,38 @@
 #pragma once
 
 #include "pmacc/types.hpp"
-#include "pmacc/expressions/DoNothing.hpp"
-#include "pmacc/expressions/SetToNull.hpp"
 
 
 namespace pmacc
 {
 
-/** wrapper for native C pointer
+/** Wrapper for a raw pointer
  *
  * @tparam T_Type type of the pointed object
  */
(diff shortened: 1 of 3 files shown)


include/pmacc/particles/memory/dataTypes/Pointer.hpp
```diff
@@ -48,24 +45,18 @@ class Pointer
      */
     HDINLINE Pointer( )
     {
-        T_InitMethod()( ptr );
+        ptr = nullptr;
```
psychocoderHPC: please move it to `: ptr( nullptr )

psychocoderHPC:
@ax3l Could you have a look why the CI is not processing the PR.

ax3l:
I might have some time tomorrow evening to take a look.

ax3l:
looks like either this PR or #2978 has a tremendously long error message that is too large to be displayed/stored from the client.

ax3l:
Please run `pic-compile -l -j 8 -q $PIC_EXAMPLES .` in a build directory manually.

sbastrakov:
@ax3l thanks, found the error in my changes.

sbastrakov:
Now works, ready to be merged.

PR #2989, opened 2019-06-21
Fix signed int overflow in particle migration between supercells

sbastrakov:
I started looking into this due to a warning in Visual Studio, but turned out this is probably a bug.

This line: `newParticleInFrame[ idx ] = -particlesInFrame;` [just below](https://github.com/sbastrakov/picongpu/blob/fe8a709269c4da5e5de4f03559f74c69ce6f3d98/include/pmacc/particles/ParticlesBase.kernel#L663) used a negation of a `uint32_t` value. Which is not a problem itself (well-defined, works by modulo), although probably a little weird style and caused the warning. But then the result is assigned to an `int` value and in all realistic scenarios, it would overflow, which is UB for signed ints. I guess it worked by bitwise copying which would give the correct result.

- [x] check that this change did not accidentally break something

### DIFF
--- a/include/pmacc/particles/ParticlesBase.kernel
+++ b/include/pmacc/particles/ParticlesBase.kernel
@@ -618,7 +618,7 @@ struct KernelShiftParticles
         >;
 
         memory::CtxArray<
-            int,
+            int32_t,
             ExchangeDomCfg
         > newParticleInFrame( 0 );
 
@@ -656,7 +656,7 @@ struct KernelShiftParticles
 
                 if ( tmpFrame.isValid() )
                 {
-                    uint32_t particlesInFrame = pb.getSuperCell( relativeCtx[ idx ] ).getSizeLastFrame( );
+                    int32_t const particlesInFrame = pb.getSuperCell( relativeCtx[ idx ] ).getSizeLastFrame( );
                     // do not use the neighbor's last frame if it is full
                     if ( particlesInFrame < frameSize )
                     {


sbastrakov:
Checked, ready for review.

include/pmacc/particles/ParticlesBase.kernel
```diff
@@ -656,7 +656,7 @@ struct KernelShiftParticles
 
                 if ( tmpFrame.isValid() )
                 {
-                    uint32_t particlesInFrame = pb.getSuperCell( relativeCtx[ idx ] ).getSizeLastFrame( );
+                    int const particlesInFrame = pb.getSuperCell( relativeCtx[ idx ] ).getSizeLastFrame( );
```
ax3l:
```suggestion
                    int32_t const particlesInFrame = pb.getSuperCell( relativeCtx[ idx ] ).getSizeLastFrame( );
```

sbastrakov: Then also change the type of `newParticleInFrame`, since it is now `int` and not `int32_t`?

sbastrakov:
MSVS *does* implicit cast as it should, just warns it is potentially unsafe since the range of values of `uint32_t` is not a subset of values of (32-bit) `int`. Of course, usually values are small and nothing bad happens in reality. In this piece of code, however, the cast is indeed unsafe with a literal interpretation of the code. Expression `-particlesInFrame` should be evaluated as a `uint32_t` value, which for positive values of `particlesInFrame` equals `2^32 - particlesInFrame`. So when `particlesInFrame <= 2^31` that would be larger than the max value of 32-bit signed int and would cause signed overflow. Of course, a compiler probably optimizes something away or performs a bitwise copy, so realistically this did not lead to an error. But we also do not gain anything from this, therefore I fixed.

include/pmacc/particles/ParticlesBase.kernel
ax3l: I would say so, looks sloppy.

ax3l:
Ok, understood now. Thanks!

sbastrakov:
Ready for review.

sbastrakov:
@psychocoderHPC can you merge this one?

Issue #2861, opened 2019-01-23
ADIOS crash on taurus


PrometheusPi:
I recently switched from HDF5 to ADIOS output (for data dumps and checkpoints) on the taurus system. The code compiles flawlessly and also starts to run. However 45 minutes into the simulation, `picongpu` crashes with the following error:
```bash
# multiple versions of the error below with different byte sizes
ERROR: MPI_AMR method (with brigade strategy): Cannot allocate 2 x 2135610252 bytes for aggregation buffers. An aggregator process needs a buffer to hold one process' output for writing, while it needs another buffer to concurrently receive another process' outpterminate called after throwing an instance of 'std::runtime_error'

# multiple versions of the error below with different byte sizes
what():  ADIOS: error at cmd 'adios_close(threadParams->adiosFileHandle)' (-1, -1) in /scratch/p_electron/richard/picongpu2/include/pmacc/../picongpu/plugins/adios/ADIOSWriter.hpp:1556 MPI_AMR method (with brigade strategy): Cannot allocate 2 x 2135610252 bytes for aggregation buffers. An aggregator process needs a buffer to hold one process' output for writing, while it needs another buffer to concurrently receive another process' outp

# traceback origin of the error:
/picongpu(_ZN8picongpu5adios11ADIOSWriter10writeAdiosEPvNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0xc305)[0x1f1bb25]
```

Due to the error, I switched back to HDF5 and that runs flawlessly. 

I build ADIOS myself following the instructions of the `dev` version of [readthedocs](https://picongpu.readthedocs.io/en/latest/install/dependencies.html#c-blosc). I am running however on the `0.4.2` version. (Might this already cause the error?) I installed c-blosc and adios as described in the manual. 

The version output of `picongpu -v` is:
```
PIConGPU: 0.4.2
  Build-Type: Release

Third party:
  OS:         Linux-3.10.0-693.21.1.el7.x86_64
  arch:       x86_64
  CXX:        GNU (6.4.0)
  CMake:      3.10.2
  CUDA:       9.2.88
  mallocMC:   2.3.0
  Boost:      1.66.0
  MPI:        
    standard: 3.1
    flavor:   OpenMPI (2.1.2)
  PNGwriter:  NOTFOUND
  libSplash:  1.7.0 (Format 4.0)
  ADIOS:      1.13.1
```

`ldd` shows no missing libraries. 

[...]

PR #2990, opened 2019-06-27
Adjust profiles for taurus-tud

steindev:
Add a getNode alias to `k80_picongpu.profile.example` requesting a node from the gpu2-interactive partition equipped with k80.
Add comment to `V100.tpl` for taurus ml partition equipped with V100 gpus
in order to provide a solution for issues due to too few hostmemory
on the taurusml nodes.
See ComputationalRadiationPhysics/picongpu#2861.
Add `--cpus-per-task` to `V100_picongpu.profile.example` and do `module switch` instead of purge and load.

### DIFF
--- a/etc/picongpu/taurus-tud/V100.tpl
+++ b/etc/picongpu/taurus-tud/V100.tpl
@@ -29,7 +29,7 @@
 #SBATCH --ntasks=!TBG_tasks
 #SBATCH --mincpus=!TBG_mpiTasksPerNode
 #SBATCH --cpus-per-task=!TBG_coresPerGPU
-#SBATCH --mem-per-cpu=1511
+#SBATCH --mem=0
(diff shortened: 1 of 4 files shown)


etc/picongpu/taurus-tud/V100.tpl
```diff
@@ -49,6 +49,10 @@
 .TBG_profile=${PIC_PROFILE:-"~/picongpu.profile"}
 
 # 6 gpus per node
+# Taurus does not have enough node memory to hold data off all GPUs in node memory during ADIOS output.
+# If you experience random crashes or get killed by the batch systems resource watch dog,
```
n01r: *...batch system's resource...*

n01r:
Thanks @steindev, just fix the apostrophe and I'll merge.

PR #3000, opened 2019-07-17
[WIP] PMacc ForEach algorithm

sbastrakov:
This PR provides the most basic form of out-of-kernel for each algorithm.
This was subject to multiple changes of ideas and is now also in a transition phase. Putting it here so that it does not lay too long in my local repo.

### DIFF
--- a/include/pmacc/algorithms/ForEach.hpp
+++ b/include/pmacc/algorithms/ForEach.hpp
@@ -1,0 +1,210 @@
+/* Copyright 2019 Sergei Bastrakov
+*
+* This file is part of PMacc.
+*
+* PMacc is free software: you can redistribute it and/or modify
+* it under the terms of either the GNU General Public License or
+* the GNU Lesser General Public License as published by
+* the Free Software Foundation, either version 3 of the License, or
+* (at your option) any later version.
+*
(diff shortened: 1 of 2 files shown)


include/pmacc/algorithms/ForEach.hpp
```diff
@@ -0,0 +1,210 @@
...
+
+            DataSpace< dim > const localSuperCellOffset =
+                superCellIdx - mapper.getGuardingSuperCells( );
+            auto accFunctor = functor(
+                acc,
+                localSuperCellOffset,
+                WorkerCfg< T_numWorkers >{ workerIdx }
+            );
+
+            auto const blockIndex{ blockIdx };
```
psychocoderHPC: please use `DataSpace< dim >( blockIdx )` because `blockIdx` is an CUDA build in type and lags dimension information.

psychocoderHPC: NOTE: `blockIndex` will not give you any real position within the simulation domain. `superCellIdx` is the index to a supercell.

psychocoderHPC:
I set this PR to `draft` because there is no progress.

sbastrakov:
Same as in another PR, if this is worked on further, should be based on the current code base.

PR #2991, opened 2019-06-27
`particles::forEach`

psychocoderHPC:
- provide a foreach method to iterate over all particles on a device

One comment points already to `CallFillGaps` This functor will be provided soon in a sperate PR. To avoid that I point later to t the comment is already pointing to the missing functor.

PR #3001, opened 2019-07-17
[WIP] Fields manipulator and for each algorithm

sbastrakov:
This is somewhat similar to #2991 , but for fields.
During the implementation, I decided to separate part of the related changes as another PR #3000.
Currently WIP, some of the functionality still missing.


psychocoderHPC:
I set this PR to `draft` because there is no progress.

sbastrakov:
I will close this, since even if there is further work on it, it should be based on the current lockstep after your refactoring.

PR #3003, opened 2019-07-23
Transition radiation plugin

finnolec:
The `transition radiation plugin` calculates the coherent and incoherent transition radiation of electrons in a PIConGPU simulation. Transition radiation is created by relativistic, charged particles moving through a dielectric interface, e.g. through a metal foil. Thus the Coulomb fields of the electrons acts on the electrons of the metal foil, who start to move and emit photons. The radiation is emitted in a cone ~1/gamma to the direction of motion of the electron bunch. In the radiation spectrum are characteristic features due to coherence effects caused by the electrons distribution. 

This plugin was created for my bachelor thesis and is based on the `radiation plugin`. The implementation follows [Schroeder, C. B. and Esarey, E. and van Tilborg, J. and Leemans, W. P., Theory of coherent transition radiation generated at a plasma-vacuum interface (2004)](https://link.aps.org/doi/10.1103/PhysRevE.69.016501) and [Downer, M. C. and Zgadzaj, R. and Debus, A. and Schramm, U. and Kaluza, M. C., Diagnostics for plasma-based electron accelerators (2018)](https://link.aps.org/doi/10.1103/RevModPhys.90.035002). Additionally the shape of the macro-particles is supported equivalent to the `radiation plugin` as described in [R Pausch and Debus, A. and Huebl, A. and Schramm, U. and Steiniger, K. and Widera, R. and Bussmann, M., Quantitatively consistent computation of coherent and incoherent radiation in particle-in-cell codes - A general form factor formalism for macro-particles (2018)](http://dx.doi.org/10.1016/j.nima.2018.02.020).

The `transition radiation plugin` assumes an infinitely wide, inifinitely thin foil. It is positioned in the x-z-plane and the radiation is calculated for all electrons at their current position.

spectrum
This is the spectrum created with _lib/python/picongpu/plugins/plot_mpl/transition_radiation_visualizer.py_ for the example _share/picongpu/examples/TransitionRadiation/_. The bunch is `~10 microns` long, which leads to the drop off in coherence in the spectrum at this length. 
Additionally the electron bunch is not moving on the y-axis, but with a 45° angle in the x-y plane, resulting in this angular intensity profile: (2 peaks, because one is in the direction of motion and one is reflected)
heatmap

Additional features, which might come, are:
- [ ] Inclinations of the metal foil. this way one does not have to tilt the whole simulation setup
- [ ] Virtual propagation of the electron (calculating transition radiation not for the current positions, but for a later position, if one adds the current velocity vector to the particle ignoring the maxwell equations). This gives an approximation for the transition radiation measured in experiments, without forcing one to continue the simulation for centimeters. 
- [ ] Jupyter widgets


psychocoderHPC:
You need to accept the invitation and push the PR again to trigger the CI.

include/picongpu/param/transitionRadiation.param
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file
+ *
+ * Definition of frequency space, number of observers, filters, form factors and
+ * window functions of the radiation plugin.
```
PrometheusPi: This is an old comment from the radiation plugin - please update.

include/picongpu/param/transitionRadiation.param
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * All values set here determine what the radiation plugin will compute.
+ * The observation direction is defined in a seperate file `radiationObserver.param`.
+ * On the comand line the plugin still needs to be called for each species the radiation
+ * should be computed for.
```
PrometheusPi: same here

include/picongpu/param/transitionRadiation.param
```diff
@@ -0,0 +1,175 @@
...
+ */
+#pragma once
+
+/**
+ * radiation verbose level:
```
PrometheusPi: ~~please adjust naming here~~
Please add a comment that you use the same verbose level scheme as the radiation plugin.

include/picongpu/param/transitionRadiation.param
```diff
@@ -0,0 +1,175 @@
...
+ * 0=nothing, 1=physics, 2=simulation_state, 4=memory, 8=critical
+ */
+#define PIC_VERBOSE_RADIATION 3
```
PrometheusPi: Do you use the same Verbose level scheme? Than please be aware that with #3004 it went to a sub namespace.

include/picongpu/param/transitionRadiation.param
```diff
@@ -0,0 +1,175 @@
...
+
+#include "picongpu/plugins/radiation/debug/PIConGPUVerboseLogRadiation.hpp"
+#include "picongpu/algorithms/Gamma.def"
+#include "picongpu/traits/attribute/GetMass.hpp"
+#include "picongpu/particles/manipulators/manipulators.def"
+
+namespace picongpu
+{
+namespace transitionRadiation
```
PrometheusPi: Optional:
Please hide it in the `plugins` namespace as well. 

This was suggested by @sbastrakov in #3004 as well.

include/picongpu/param/transitionRadiation.param
```diff
@@ -0,0 +1,175 @@
...
+    //! number of frequency values to compute in the logarithmic frequency [unitless] 
+    constexpr unsigned int nOmega = 256;
+
+} // namespace logFrequencies
+
+
+namespace listFrequencies
+{
+    //! path to text file with frequencies
+    constexpr char listLocation[] = "/path/to/frequency_list";
```
PrometheusPi: Have you tested this?

include/picongpu/param/transitionRadiation.param
```diff
@@ -0,0 +1,175 @@
...
+    namespace macroParticleFormFactorCIC_3D { }
+    namespace macroParticleFormFactorTSC_3D { }
+    namespace macroParticleFormFactorPCS_3D { }
+    namespace macroParticleFormFactorCIC_1Dy { }
+    namespace macroParticleFormFactorGaussSpherical { }
+    namespace macroParticleFormFactorGaussCell { }
+    namespace macroParticleFormFactorIncoherent { }
+    namespace macroParticleFormFactorCoherent { }
+
+    namespace macroParticleFormFactor = macroParticleFormFactorGaussSpherical;
```
PrometheusPi: Do you use all these form factors?

include/picongpu/param/transitionRadiationObserver.param
```diff
@@ -0,0 +1,123 @@
...
+
+#pragma once
+
+namespace picongpu
+{
+namespace transitionRadiation
+{
+    /** Compute observation angles
+     *
+     * This function is used in the Radiation plug-in kernel to compute
```
PrometheusPi: Please write transition radiation plugin instead of Radiation plug-in

include/picongpu/param/transitionRadiationObserver.param
```diff
@@ -0,0 +1,123 @@
...
+        /* angles range for theta */
+        /** minimum theta angle [rad] */
+        const picongpu::float_64 angle_theta_start = parameters::thetaMin;
+        /** maximum theta angle [rad] */
+        const picongpu::float_64 angle_theta_end = parameters::thetaMax;
+        /* angles range for phi */
+        /** minimum phi angle [rad] */
+        constexpr picongpu::float_64 angle_phi_start = parameters::phiMin;
+        /** maximum phi angle [rad] */
+        constexpr picongpu::float_64 angle_phi_end = parameters::phiMax;
```
PrometheusPi: If you moved the definition of the angle ranges to the other param file, you do not need a renaming here.

include/picongpu/param/transitionRadiation.param
```diff
@@ -0,0 +1,175 @@
...
+        constexpr unsigned int nTheta = 256;
+        constexpr unsigned int nObserver = nPhi * nTheta;
+
+        // theta goes from 0 to pi
+        constexpr float_64 thetaMin = 0.0; 
+        constexpr float_64 thetaMax = picongpu::PI; 
+
+        // phi goes from 0 to 2*pi
+        constexpr float_64 phiMin = 0.0;
+        constexpr float_64 phiMax = 2 * picongpu::PI;
```
PrometheusPi: Your definition of the observation angle ranges is fine here, but that makes the other param file a bit obsolete. 
How about moving these definitions entirely to the observer param file?

finnolec: I would probably prefer using just a single param file, since `transitionRadiationObserver.param` is just used for creating unit-vectors in the PIC coordinatesystem from the `transitionRadiation.param` observation angles. 2 param files seems like a lot of files for this plugin with basically just 3x3 parameters (min, max, n) x (phi, theta, omega). But if the vector-creation moves into `transitionRadiation.param`, this file looks a little bit messy.

include/picongpu/param/transitionRadiation.param
finnolec: Nope. I included them because they are included in the radiation plugin and maybe people find a use for it. The differences from the list frequencies and the log/lin frequencies are not very big, so I'd expect this to work, if it works in the _radiation plugin_.

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+
+    /* Arbitrary margin which is necessary to prevent division by 0 error
+        * created by particles moving in the plane of the foil.
+        */
+    float_X const DIV_BY_ZERO_MINIMUM = 1.e-7;
+
+    /** Calculator class for calculation of transition radiation.
+     * 
+     * @param particleSet transitionRadiation::Particle to compute transition radiation for
+     * @param lookDirection vector with observation direction
```
PrometheusPi: Optional: ~~with~~-> of ?

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+     */
+    class Calculator
+    {
+
+    private:
+        transitionRadiation::Particle const & particle;
+        float3_X const lookDirection;
```
PrometheusPi: Using a const variable here is great. @psychocoderHPC Do you think having it additionally a reference provide a possible speedup too? (regarding registers used)

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+
+        float_X parMomSinTheta;
+        float_X parMomCosTheta;
+        float_X const parMomPhi;
+        float_X parMomSinPhi;
+        float_X parMomCosPhi;
+        float_X detectorSinTheta;
+        float_X detectorCosTheta;
+        float_X const detectorPhi;
+        float_X const parSqrtOnePlusUSquared;
```
PrometheusPi: I assume that the switches between `const` and non-const are intentional. Please state whether my guess is true.

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+            float3_X const lookDirection
+        ) : 
+            particle( particleSet ),
+            lookDirection( lookDirection ),
+            parMomPhi( particle.getMomPhi( ) ),
+            detectorPhi( 
+                picongpu::math::atan2(
+                    lookDirection.z( ), 
+                    lookDirection.x( )
+                ) + picongpu::PI
```
PrometheusPi: Please add a comment on why you need to add \pi

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+            ),
+            parSqrtOnePlusUSquared( 
+                picongpu::math::sqrt( 1 + particle.getU( ) * particle.getU( ) ) 
+            )
+        { 
+            // Frequent calculations
+            // Momentum Space for Particle:
```
PrometheusPi: please remove capital letters

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+            picongpu::math::sincos( 
+                parMomPhi - detectorPhi, 
+                parMomSinPhi, 
+                parMomCosPhi
+            );
+            
+            // Detector Position
+            float_X const detectorTheta = picongpu::math::acos( 
+                lookDirection.y( ) / picongpu::math::sqrt( ( 
+                    lookDirection * lookDirection
```
PrometheusPi: Is your `lookDirection` not normalized?

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+            lookDirection( lookDirection ),
+            parMomPhi( particle.getMomPhi( ) ),
+            detectorPhi( 
+                picongpu::math::atan2(
+                    lookDirection.z( ), 
+                    lookDirection.x( )
+                ) + picongpu::PI
+            ),
+            parSqrtOnePlusUSquared( 
+                picongpu::math::sqrt( 1 + particle.getU( ) * particle.getU( ) )
```
PrometheusPi: `particle.getU( ) * particle.getU( )` is reused in line 119 - you could save the result in a (register) variable and thus avoid recomputation (5 ALU operations - possible only 3 due to `*+` optimization) 

@psychocoderHPC what are your thoughts on such optimizations?

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+            );
+        }
+
+        /** Perpendicular part of normalized energy
+         * 
+         * Calculates perpendicular part to movement direction of normalized energy
+         * determined by formula:
+         * @f[E_{perp} = (u^2 \cos{\psi} \sin{\psi} \sin{\phi} \cos{\theta}) / 
+         *          ((\sqrt{1 + u^2} - u \sin{\psi} \cos{\phi} \sin{\theta})^2 - u^2 \cos{\phi}^2 \cos{\theta}^2)@f]
+         * where Psi is the azimuth angle of the particle momentum and theta is
```
PrometheusPi: Please use \psi instead of Psi (there is a capital Psi)

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+            float_X denominator = x * x - y * y;
+
+            // Preventing division by 0
+            if( math::abs( denominator ) < DIV_BY_ZERO_MINIMUM )
+            {
+                if( denominator < 0.0 )
+                    denominator = -DIV_BY_ZERO_MINIMUM;
+                else
+                    denominator = DIV_BY_ZERO_MINIMUM;
+            }
```
PrometheusPi: These `if` clauses might light to branching of the parallel code and thus to a slowdown of the parallel execution.

Please consider rewriting this as:
```c++
const bool prevent_div_by_zero = math::abs( denominator ) < DIV_BY_ZERO_MINIMUM;
    denominator = int(!prevent_div_by_zero) * denominator 
        + int(prevent_div_by_zero) * ( -2 * int( denominator < 0.0 ) + 1) * DIV_BY_ZERO_MINIMUM;
```

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+
+            return a * ( 1.0 / denominator );
```
PrometheusPi: Why not `a / denominator`?

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        /** Parallel part of normalized energy
+         * 
+         * Calculates parallel part to movement direction of normalized energy
+         * determined by formula:
+         * @f[E_{para} = (u \cos{\psi} (u \sin{\psi} \cos{\phi} - \sqrt{1 + u^2} \sin{\theta}) / 
+         *          ((\sqrt{1 + u^2} - u \sin{\psi} \cos{\phi} \sin{\theta})^2 - u^2 \cos{\phi}^2 \cos{\theta}^2)@f]
+         * where Psi is the azimuth angle of the particle momentum and theta is
```
PrometheusPi: Please use \psi instead of Psi

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+         * the azimuth angle of the detector position to the movement direction y
+         * 
+         * @return parallel part of normalized energy
+         */
+        HDINLINE
+        float_X 
+        calcEnergyPara( ) const
+        {
+            float_X const a = particle.getU( ) * parMomCosTheta;
```
PrometheusPi: Please add a comment what `a`, `b`, and `c` mean

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+            float_X denominator = x * x - y * y; 
+
+            // Preventing division by 0
+            if( math::abs( denominator ) < DIV_BY_ZERO_MINIMUM )
+            {
+                if( denominator < 0.0 )
+                    denominator = -DIV_BY_ZERO_MINIMUM;
+                else
+                    denominator = DIV_BY_ZERO_MINIMUM;
+            }
```
PrometheusPi: Please use same code as suggested above.

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+
+            return a * ( b - c ) * ( 1.0 / denominator );
```
PrometheusPi: Why not `a * ( b - c ) / denominator` ?

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+    )
+    {
+        // If case for longitudinal moving particles
+        if ( exponent.get_real() == -1.0 )
+            return complex_X( 0.0, 0.0 );
+        else
+            return complex_X( 
+                math::exp( 
+                    //precisionCast< float_64 >( exponent * omega )
+                    exponent * omega
```
PrometheusPi: Here you cloud avoid branching too:
```c++
const bool long_mov_particle = exponent.get_real() == -1.0;
int(long_mov_particle) * complex_X( 0.0, 0.0 ) + int(!long_mov_particle) * complex_X( ... ;
```

PrometheusPi: @psychocoderHPC Could you please comment whether such optimizations are useful here?

include/picongpu/plugins/transitionRadiation/Calculator.hpp
PrometheusPi: @psychocoderHPC Could you please comment whether such optimizations are useful here?

include/picongpu/plugins/transitionRadiation/Calculator.hpp
```diff
@@ -0,0 +1,238 @@
...
+                ) 
+            );
+    }
+    
+} // namespace transitionRadiation
+} // namespace picongpu
```
PrometheusPi: Please add a newline at the end of the file.

include/picongpu/plugins/transitionRadiation/Calculator.hpp
finnolec: I tried to use `const` as often as possible. But since e.g. `parMomSinPhi` is calculated with call-by-reference in the `math::sincos` function, its value isn't constant. I'm not sure how I can circumvent this problem.

include/picongpu/plugins/transitionRadiation/Calculator.hpp
finnolec: Since the lookdirection is set in the param files, it could theoretically be set by the user to a not normalized value... But I guess it should be clear from the comment, that `transtionradiation::observationDirection(const int observation_id_extern)` has to be normalized.

include/picongpu/plugins/transitionRadiation/Calculator.hpp
finnolec: I tried to use `const` as much as possible. But since `detectorSinTheta` is calculated in the call-by-reference function `picongpu::math::sincos`, it is not constant for c++. I don't know how to circumvent this problem. I would guess that the `sincos` should still be _better_ for the calculations as 2 `const` members, which are calculated with normal `picongpu::math::sin` and `picongpu::math::cos` functions. I would guess it is a question of whether performance or storage is more of a problem.

include/picongpu/param/transitionRadiation.param
PrometheusPi: I am fine with that.

include/picongpu/param/transitionRadiation.param
PrometheusPi: I can understand both of your points. Please decide for yourself whether you prefer having two or just one param file. As long as the variable distribution is consistent (in contrast to the radiation plugin 😉 ) I am fine with it.

include/picongpu/plugins/transitionRadiation/Calculator.hpp
PrometheusPi: Thanks for your explanation. That's what I hoped for. Yes, if you use `sincos` you have to have non-const variables.

include/picongpu/plugins/transitionRadiation/Particle.hpp
```diff
@@ -0,0 +1,224 @@
...
+        {
+            return calcBeta( getMomentum( ) ) * picongpu::SPEED_OF_LIGHT;
+        }
+
+        //! @return polar angle phi of momentum
+        HDINLINE
+        float_X 
+        getMomPhi( ) const
+        {
+            return calcMomPhi( );
```
PrometheusPi: You are hiding the `calc...` methods behind `get...` methods - that's Okay for me, but adds unnecessary code and might add confusion.  
@psychocoderHPC and @sbastrakov what do you think?

include/picongpu/plugins/transitionRadiation/Particle.hpp
```diff
@@ -0,0 +1,224 @@
+/* Copyright 2017-2019 Finn-Ole Carstens
```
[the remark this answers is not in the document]
finnolec: I'm not really sure, but I think I wrote it from scratch. But since a few functions are taken from your particle class, I added your header :)

include/picongpu/plugins/transitionRadiation/Particle.hpp
PrometheusPi: I just had an *offline* discussion with @psychocoderHPC - please avoid any getter that just do renaming of getter you call `calc...`

include/picongpu/plugins/transitionRadiation/Particle.hpp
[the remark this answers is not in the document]
PrometheusPi: That confuses me 😕 - if you copied my header shouldn't there by a long author list?

include/picongpu/plugins/transitionRadiation/Particle.hpp
```diff
@@ -0,0 +1,224 @@
...
+        
+        //! @return radial, perpendicular component of location in cylindrical coordinates
+        HDINLINE
+        float_X 
+        getPosPerp( ) const
+        {
+            return calcPosRho( );
+        }
+
+        //! @return parallel component to z of location in cylindrical coordinates
```
PrometheusPi: Please avoid using a rotated coordinate system. Even if the z is the common axis in cylindrical coordinates, please use y here - since this directly makes a connection with the PIConGPU Cartesian coordinate system.

include/picongpu/plugins/transitionRadiation/Particle.hpp
```diff
@@ -0,0 +1,224 @@
...
+    private:
+        //! @return beta=v/c
+        HDINLINE 
+        float_X 
+        calcBeta(
+            float3_X const & momentum
+        ) const
+        {
+            float_X const gamma = calcGamma(momentum);
+            return picongpu::math::sqrt(1 - 1 / (gamma * gamma) );
```
PrometheusPi: you cloud use `1.` instead of `1` to avoid a `int` -> `float` cast here.

include/picongpu/plugins/transitionRadiation/Particle.hpp
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        //! @return gamma = E/(mc^2)
+        HDINLINE 
+        float_X 
+        calcGamma(
+            float3_X const & momentum
+        ) const
+        {
+            float_X const x = 1.0 / ( mass * picongpu::SPEED_OF_LIGHT ) * ( mass * picongpu::SPEED_OF_LIGHT );
```
PrometheusPi: This is wrong!
```c++
 float_X const x = 1.0 / ( mass * picongpu::SPEED_OF_LIGHT ) * ( mass * picongpu::SPEED_OF_LIGHT );
// is equal to:
 float_X const x = (1.0 / ( mass * picongpu::SPEED_OF_LIGHT )) * ( mass * picongpu::SPEED_OF_LIGHT );
// is equal to:
 float_X const x = 1.0;
```
please write: 
```c++
float_X const mass_times_c = mass * picongpu::SPEED_OF_LIGHT;
float_X const x = 1.0 / (mass_times_c * mass_times_c);
```

include/picongpu/plugins/transitionRadiation/Particle.hpp
```diff
@@ -0,0 +1,224 @@
...
+        //! @return normalized momentum u = gamma * beta
+        HDINLINE 
+        float_X 
+        calcU(
+            float3_X const & momentum
+        ) const
+        {
+            float_X const gamma1 = calcGamma( momentum );
+            float_X const beta1 = calcBeta( momentum );
+            return gamma1 * beta1;
```
PrometheusPi: why do you use the additional `...1` in the variable names?
If there is a deeper meaning, please add a comment - if not, please remove the suffix to avoid confusion.

include/picongpu/plugins/transitionRadiation/Particle.hpp
```diff
@@ -0,0 +1,224 @@
...
+
+        //! @return polar angle phi of momentum
+        HDINLINE
+        float_X 
+        calcMomPhi( ) const
+        {
+            return picongpu::math::atan2(
+                momentum.x( ), 
+                momentum.z( )
+            ) + picongpu::PI;
```
PrometheusPi: please add a comment why you have to add π here

include/picongpu/plugins/transitionRadiation/Particle.hpp
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        //! @return azimuth angle psi of momentum
+        HDINLINE
+        float_X 
+        calcMomTheta( ) const
+        {
+            //because of floating point precision x^2+y^2+z^2<y^2 for x,z<<z
```
PrometheusPi: typo `x,z<<z` -> `x,z<<y` ~~or `x,y<<z`?~~

include/picongpu/plugins/transitionRadiation/Particle.hpp
```diff
@@ -0,0 +1,224 @@
...
+
+        //! @return radial angle phi of location in cylindrical coordinates
+        HDINLINE
+        float_X 
+        calcPosPhi( ) const
+        {
+            return picongpu::math::atan2(
+                location.x( ), 
+                location.z( )
+            ) + picongpu::PI;
```
PrometheusPi: same here - please add comment for why you have to add π

include/picongpu/plugins/transitionRadiation/Particle.hpp
```diff
@@ -0,0 +1,224 @@
...
+        }
+    }; // class Particle
+    
+} // namespace transitionRadiation
+} // namespace plugins
+} // namespace picongpu
```
PrometheusPi: please add newline

include/picongpu/plugins/transitionRadiation/frequencies/LinearFrequencies.hpp
```diff
@@ -0,0 +1,76 @@
...
+#pragma once
+
+#include "picongpu/simulation_defines.hpp"
+
+
+namespace picongpu
+{
+namespace plugins
+{
+namespace transitionRadiation
```
PrometheusPi: If this is the same code as in the radiation plugin, please just use that code directly.

include/picongpu/plugins/transitionRadiation/frequencies/ListFrequencies.hpp
```diff
@@ -0,0 +1,132 @@
...
+#include <fstream>
+#include <cstdio>
+#include <pmacc/memory/buffers/GridBuffer.hpp>
+
+
+namespace picongpu
+{
+namespace plugins
+{
+namespace transitionRadiation
```
PrometheusPi: If this is the same code as in the radiation plugin, please just use that code directly.

include/picongpu/plugins/transitionRadiation/frequencies/LogFrequencies.hpp
```diff
@@ -0,0 +1,85 @@
...
+#pragma once
+
+#include "picongpu/simulation_defines.hpp"
+
+
+namespace picongpu
+{
+namespace plugins
+{
+namespace transitionRadiation
```
PrometheusPi: If this is the same code as in the radiation plugin, please just use that code directly.

include/picongpu/unitless/speciesAttributes.unitless
```diff
@@ -146,6 +146,46 @@ struct WeightingPower<radiationMask>
     }
 };
 
+template<>
```
PrometheusPi: Please add your name the the license header

include/picongpu/plugins/transitionRadiation/Particle.hpp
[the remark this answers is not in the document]
finnolec: I just added your header in the upcoming commit

lib/python/picongpu/plugins/data/transitionradiation.py
```diff
@@ -0,0 +1,274 @@
...
+            short name of the particle species, e.g. "e" for electrons
+            (defined in ``speciesDefinition.param``)
+        iteration: (unsigned) int or list of int
+            The iteration at which to read the data.
+
+        Returns
+        -------
+        A string with a path.
+        """
+        if species is None:
```
PrometheusPi: How can species be `None` - you do not give a default `None` value?

lib/python/picongpu/plugins/data/transitionradiation.py
```diff
@@ -0,0 +1,274 @@
...
+                theta = int(np.floor(maxIndex / len(self.phis)))
+
+            print("Angular intensity distribution is sliced at theta={:.2e} "
+                  "with omega={:.2e}.".format(self.thetas[theta], self.omegas[omega]))
+            return self.phis, self.data[theta * len(self.phis):(theta + 1) * len(self.phis), omega]
+        elif type == "heatmap":
+            # find omega with maximum intensity if it is not given as parameter
+            if omega is None:
+                omega = 0
+                print("theta: ", theta, "phi: ", phi, "omega: ", omega)
```
PrometheusPi: I think here you do not need spaces after the colon - when separated by a comma, print adds a space.

lib/python/picongpu/plugins/plot_mpl/requirements.txt
```diff
@@ -1,3 +1,6 @@
 matplotlib
 numpy
+scipy
+sys
+getopt
```
PrometheusPi: I am surprised these dependencies were not included beforehand - `scipy` gets included by `-r ../data/requirements.txt` - the others were not included beforehand

lib/python/picongpu/plugins/data/transitionradiation.py
finnolec: I'm pretty sure, this was just a debug message. I'll remove it

lib/python/picongpu/plugins/plot_mpl/requirements.txt
finnolec: I was surprised too :D I guess it doesn't hurt, to add them to the `requirements.txt`, right?

PrometheusPi: I agree

include/picongpu/plugins/transitionRadiation/frequencies/LogFrequencies.hpp
finnolec: It is not too trivial to use the same frequency functors for the `radiation plugin` as well as the `transition radiation plugin`. E.g. `radiation_log_freq.hpp` uses `plugins::radiation::frequencies::omega_min` directly, because it is in the same namespace, whereas my `LogFrequencies.hpp` uses `plugins::transitionRadiation::frequencies::omegaMin` from the `transitionRadiation` namespace. I see two possible solutions for this problem:

- One merges the namespaces together to use the same functors. But this would deny the usage of both plugins in the same simulation. 
- On the other hand one could refactor the functors, so that they take omega_min, omega_max and N_omega as parameters, which requires refactoring in both plugins, but it seems like the clean way to me.

include/picongpu/plugins/transitionRadiation/Calculator.hpp
finnolec: I have included
```cpp
const bool preventDivByZero = math::abs( denominator ) < DIV_BY_ZERO_MINIMUM;
denominator = int( !preventDivByZero ) * denominator +  int( preventDivByZero ) * ( -2 * int( denominator < 0.0 ) + 1 ) * DIV_BY_ZERO_MINIMUM;
```
in `calcEnergyPara` and `calcEnergyPerp` and 
```cpp
const bool longMovingParticle = exponent.get_real() == -1.0;
return float_X( longMovingParticle ) * complex_X( 0.0, 0.0 ) +  float_X( !longMovingParticle ) * complex_X( math::exp( exponent * omega ) );
```
in the [formfactor](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3003#discussion_r307182735) for the `Transition Radiation Example` setup with a bit of a higher resolution for a quick benchmark. The `calculation simulation time` for these 5 consecutive transition radiation calculations with my original code was `18 minutes` long, where it was  `19 minutes long` with all the three changes. I will quickly run the simulation again with my code in the formfactor and your code in the energies to see how this effects the simulation time.

finnolec: Now I have 2 more simulations with 1) the formfactor-change and 2) the change in the energies. The simulation with the formfactor change from @PrometheusPi takes `17:44 minutes` and the second simulation takes `19:40 minutes`. These benchmarks are just rough approximations, but I think they should be a good estimate for the amount of time which can be saved with these changes. I run one of the simulations twice and the `calclulation simulation time` was identical. Hence I would use [this](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3003#discussion_r307182735) method, but not the one from the energy functions.

include/picongpu/plugins/transitionRadiation/frequencies/LogFrequencies.hpp
PrometheusPi: Okay - that is a valid reason to keep is separate for now - or I do a quick refactoring today and we will tray to include an static object as template argument to give omega_min and omega_max.

PrometheusPi:
There are still some python style issues that prevent the travis-ci from passing and also a c++ file with and-of-line white-space.

lib/python/picongpu/plugins/plot_mpl/transition_radiation_visualizer.py
```diff
@@ -0,0 +1,275 @@
...
+        elif self.type == "sliceoverphi":
+            phis, spectral_power = self.data[idx]
+            self.plt_obj[idx] = self.ax.plot(phis, spectral_power)
+        elif self.type == "heatmap":
+            theta_mesh, phi_mesh, spectral_power = self.data[idx]
+            im = self.plt_obj[idx] = self.ax.pcolormesh(theta_mesh, phi_mesh,
+                                                        spectral_power)
+            self.ax.get_figure().colorbar(im,
+                                          label=r"Spectral Power $d^2 W / "
+                                                r"d\omega d\Omega$")
```
PrometheusPi: In what units do you plot the result? 
```
[Js]
```
?

lib/python/picongpu/plugins/plot_mpl/transition_radiation_visualizer.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def _update_plt_obj(self, idx):
+        """
+        Implementation of base class function.
+        """
+        # counts, bins, iteration, dt = self.data[idx]
+        # self.plt_obj[idx].set_data(bins, counts)
+        # TODO
```
PrometheusPi: Here seems to be something incomplete. 
@finnolec Could you please comment on what should be added here and whether this will be done with this pull request or a follow up.

lib/python/picongpu/plugins/plot_mpl/transition_radiation_visualizer.py
```diff
@@ -0,0 +1,275 @@
...
+            # calculate positions for minor and major ticks for the beauty
+            # of the plot
+            nmaj = 200
+            nmin = 10
+            minorlocs = np.linspace(0.1, 1, nmin, endpoint=True)
+            majorlocs = np.linspace(99, -100, nmaj, endpoint=True)
+            alllocs = np.zeros(nmaj * nmin)
+            for i in range(nmaj):
+                for j in range(nmin):
+                    alllocs[i * nmin + j] = minorlocs[j] * 10 ** majorlocs[i]
```
PrometheusPi: Just as a remark:
The above code looks geared to your current application. 
I would expect issues with the ticks if you go to much larger or much smaller frequency ranges?

lib/python/picongpu/plugins/plot_mpl/transition_radiation_visualizer.py
```diff
@@ -0,0 +1,275 @@
...
+        elif self.type == "sliceovertheta":
+            self.ax.set_title(
+                "Angular transition radiation distribution for " + species)
+            self.ax.set_xlabel(r"Detector Angle $\theta$")
+            self.ax.set_ylabel(r"Spectral Power $d^2 W / d\omega d\Omega$")
+        elif self.type == "sliceoverphi":
+            self.ax.set_title(
+                "Angular transition radiation distribution for " + species)
+            self.ax.set_xlabel(r"Detector Angle $\phi$")
+            self.ax.set_ylabel(r"Spectral Power $d^2 W / d\omega d\Omega$")
```
PrometheusPi: units [Js] ?

lib/python/picongpu/plugins/plot_mpl/transition_radiation_visualizer.py
```diff
@@ -0,0 +1,275 @@
...
+            elif opt in ["-s", "--species"]:
+                species = arg
+            elif opt in ["-t", "--type"]:
+                type = arg
+            elif opt in ["-P", "--phi"]:
+                phi = arg
+            elif opt in ["-T", "--theta"]:
+                theta = arg
+            elif opt in ["-O", "--omega"]:
+                omega = arg
```
PrometheusPi: In order to avoid this manual parsing of arguments and the manual help page, I think using the module `argparse`, see e.g. [radiation plotting here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/d340812641320d3a61e343a521287fd0720ba55b/src/tools/bin/plotRadiation#L67-L210)

n01r:
> new example for the transition radiation?

@PrometheusPi, not a bad idea, I would say.
A dedicated example where the setup is done right once and the analysis fits the model from the literature.

PrometheusPi:
Okay - regarding the new example:
disadvantages: 
 - more code to maintain
 - longer compile test time
advantages:
 - "plug and play" example for the transition radiation
 - new start point for own setups

@ComputationalRadiationPhysics/picongpu-maintainers please vote :+1: for having a new example or :-1: against the new example (in this comment box).

n01r:
I'd say any major physics feature of the code will likely stay mostly dead unless there is an example to start from.

ax3l:
No prob to add a good example, you can just add one instead of a whole list of compiles if feasible.

PrometheusPi:
@finnolec We will keep your example - (more for me to review :cry:)

n01r:
Nice work :+1: 

Could you just fix the style issues that the Travis CI noticed?

PrometheusPi:
Since I had not time to finish the frequency factory till today - we should use the double frequency definition as it is right now and I will refactor your frequency definition together with my commit. 

@finnolec There are still 3 open bullet points. Are they finished or are they still pending? 
How about a documentation of your plugin?
Should we rebase together again?

finnolec:
I will write the documentation again. I implemented one of the bullet points, but the other two are pending. I can also just add them in future pullrequests though :)

lib/python/picongpu/plugins/plot_mpl/transition_radiation_visualizer.py
finnolec: This is true. But it should work with all frequencies which can are measured in experiments, so I don't think it is a major problem.

lib/python/picongpu/plugins/plot_mpl/transition_radiation_visualizer.py
finnolec: I followed the [Energy histogram visualizer](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py) with the argument parsing for consistency...

lib/python/picongpu/plugins/plot_mpl/transition_radiation_visualizer.py
finnolec: yep, always in [Js]. I'll add it right now

PrometheusPi:
Commit structure looks great - can we now review it?

finnolec:
Yep, I don't want to change anything at the moment

include/picongpu/param/transitionRadiation.param
PrometheusPi: It did not work correctly in the radiation plugin - see #3053

include/picongpu/param/transitionRadiation.param
```diff
@@ -0,0 +1,264 @@
...
+        constexpr float_64 thetaMax = picongpu::PI;
+
+        // phi goes from 0 to 2*pi
+        constexpr float_64 phiMin = 0.0;
+        constexpr float_64 phiMax = 2 * picongpu::PI;
+
+        namespace SI
+        {
+            // z position of the foil to calculate transition radiation at
+            // leave at 0 for no virtual particle propagation
```
PrometheusPi: Could you please (in this pull request) specify where the foil is located in case `foilPosition = 0.0`?

include/picongpu/param/transitionRadiation.param
```diff
@@ -0,0 +1,264 @@
...
+        constexpr float_64 thetaMin = 0.0;
+        constexpr float_64 thetaMax = picongpu::PI;
+
+        // phi goes from 0 to 2*pi
+        constexpr float_64 phiMin = 0.0;
+        constexpr float_64 phiMax = 2 * picongpu::PI;
+
+        namespace SI
+        {
+            // z position of the foil to calculate transition radiation at
```
PrometheusPi: I think you mean `y` not `z`?

PrometheusPi:
```suggestion
            // y position of the foil to calculate transition radiation at
```

include/picongpu/param/transitionRadiation.param
finnolec: Should I note this somewhere? I tried to leave everything frequency related similar to how it is currently implemented in the radiation plugin. This way one can just copy+paste the changes from the reworked radiation plugin into the transition radiation plugin.

include/picongpu/plugins/transitionRadiation/Particle.hpp
```diff
@@ -0,0 +1,184 @@
...
+        {
+            return betaAbs * picongpu::SPEED_OF_LIGHT;
+        }
+
+        //! propagates the current window to the foil position
+        HDINLINE
+        void
+        propagate( const float_X & propagationDistance )
+        {
+            location += propagationDistance * beta;
```
PrometheusPi: This is only valid if velocities ae normalized to c and the `propagationDistance` is a time. 
@finnolec is the later the case?

PrometheusPi:
```suggestion
            location += propagationDistance * beta * picongpu::SPEED_OF_LIGHT;
```

include/picongpu/plugins/transitionRadiation/Particle.hpp
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        //! @return azimuth angle psi of momentum
+        HDINLINE
+        float_X
+        getMomTheta( ) const
+        {
+            //because of floating point precision x^2+y^2+z^2<y^2 for x,z<<y
+            float_X const momAbs = getMomAbs( );
+            if( momAbs <= momentum.y( ) )
```
PrometheusPi: This is fine as long our main propagation direction is in positive y-direction - but will fail e.g. for the bunch example where the default propagation direction is in negative y-direction. 
@finnolec What happens if `x^2+y^2+z^2<y^2` is true  but we go into the `else` branch?

PrometheusPi:
```suggestion
            if( momAbs * momAbs <= momentum.y( ) * momentum.y( ) )
```

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -0,0 +1,661 @@
...
+        pluginRegisterHelp(
+            po::options_description& desc
+        )
+        {
+            desc.add_options( )(
+                ( pluginPrefix + ".period" ).c_str( ),
+                po::value< std::string >( &notifyPeriod ),
+                "enable plugin [for each n-th step]"
+            )(
+                ( pluginPrefix + ".omegaList").c_str(),
```
PrometheusPi: see #3053 - this should not work

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -0,0 +1,661 @@
...
+            // collect data GPU -> CPU -> Master
+            copyRadiationDeviceToHost( );
+            collectRadiationOnMaster( );
+            sumTransitionRadiation( theTransRad, tmpITR, tmpCTRpara, tmpCTRperp, tmpNum );
+        }
+
+        /** Final transition radiation calculation on CPU side
+         *
+         * Calculate transition radiation integrals. This can't happen on the GPU
+         * since the absolute square of a sum can't be moved within a sum.
```
PrometheusPi: I do not understand what you mean by "moved". Could you explain it here in GitHub?

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -0,0 +1,661 @@
...
+                 ******** Here happens the true physical calculation ********
+                ************************************************************/
+                for( unsigned int i = 0; i < elementsTransitionRadiation( ); ++i )
+                {
+                    const float_X ctrPara = math::abs2( ctrParaArray[ i ] );
+                    const float_X ctrPerp = math::abs2( ctrPerpArray[ i ] );
+                    if (numArray[i] != 0.0)
+                    {
+                        targetArray[ i ] = (
+                            itrArray[ i ] + ( numArray[ i ] - 1.0 ) * ( ctrPara + ctrPerp ) / numArray[i]
```
PrometheusPi: How do you ensure that `numArray[i]`≠0?

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -0,0 +1,661 @@
...
+                            ( 1.0 / ( 4 * PI * SI::EPS0_SI * PI * PI * SI::SPEED_OF_LIGHT_SI ) );
+                        outFile <<
+                            values[
+                                index_direction * transitionRadiation::frequencies::nOmega + index_omega
+                            ] * transRadUnit << "\t";
+
+                    } // for loop over all frequencies
+
+                    outFile << std::endl;
+                } // for loop over all frequencies
```
PrometheusPi: This is closing the loop over all directions.

PrometheusPi:
```suggestion
                } // for loop over all directions
```

include/picongpu/plugins/transitionRadiation/TransitionRadiation.kernel
```diff
@@ -0,0 +1,391 @@
...
+#include "picongpu/simulation_defines.hpp"
+#include "picongpu/plugins/transitionRadiation/Particle.hpp"
+#include "picongpu/plugins/transitionRadiation/Calculator.hpp"
+#include "picongpu/plugins/radiation/radFormFactor.hpp"
+#include "picongpu/plugins/transitionRadiation/GammaMask.hpp"
+
+#include <pmacc/dimensions/DataSpaceOperations.hpp>
+#include <pmacc/mappings/kernel/AreaMapping.hpp>
+
+// #include "picongpu/plugins/transitionRadiation/frequencies/LogFrequencies.hpp"
```
PrometheusPi: Why are the commented out?

share/picongpu/examples/TransitionRadiation/include/picongpu/param/transitionRadiation.param
```diff
@@ -0,0 +1,264 @@
...
+        constexpr unsigned int nTheta = 128;
+        constexpr unsigned int nObserver = nPhi * nTheta;
+
+        // theta goes from 0 to pi
+        constexpr float_64 thetaMin = 0.0;
+        constexpr float_64 thetaMax = picongpu::PI;
+
+        // phi goes from 0 to 2*pi
+        constexpr float_64 phiMin = 0.0;
+        constexpr float_64 phiMax = 2 * picongpu::PI;
```
PrometheusPi: This looks like the default `transitionRadiation.param`. If this is the case, you could avoid adding it - which however would be strange since this is the example for OTR.

share/picongpu/examples/TransitionRadiation/include/picongpu/param/unit.param
```diff
@@ -0,0 +1,58 @@
...
+ * but might use the defined constants in other input files.
+ */
+
+#pragma once
+
+
+namespace picongpu
+{
+    /** Unit of time */
+    constexpr float_64 UNIT_TIME = SI::DELTA_T_SI;
```
PrometheusPi: What did you need to add in this file?

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
finnolec: see above

include/picongpu/plugins/transitionRadiation/Particle.hpp
finnolec: c/c = 1

include/picongpu/plugins/transitionRadiation/TransitionRadiation.kernel
```diff
@@ -0,0 +1,391 @@
...
+// #include "picongpu/plugins/transitionRadiation/frequencies/LinearFrequencies.hpp"
+// #include "picongpu/plugins/transitionRadiation/frequencies/ListFrequencies.hpp"
```
finnolec:
```suggestion
```

include/picongpu/plugins/transitionRadiation/TransitionRadiation.kernel
```diff
@@ -0,0 +1,391 @@
...
+#include "picongpu/plugins/transitionRadiation/Particle.hpp"
+#include "picongpu/plugins/transitionRadiation/Calculator.hpp"
+#include "picongpu/plugins/radiation/radFormFactor.hpp"
+#include "picongpu/plugins/transitionRadiation/GammaMask.hpp"
+
+#include <pmacc/dimensions/DataSpaceOperations.hpp>
+#include <pmacc/mappings/kernel/AreaMapping.hpp>
+
+// #include "picongpu/plugins/transitionRadiation/frequencies/LogFrequencies.hpp"
+// #include "picongpu/plugins/transitionRadiation/frequencies/LinearFrequencies.hpp"
```
finnolec:
```suggestion
```

include/picongpu/plugins/transitionRadiation/TransitionRadiation.kernel
finnolec:
```suggestion
```

docs/source/usage/plugins/transitionRadiation.rst
```diff
@@ -0,0 +1,271 @@
...
+
+If one wants to virtually propagate the electron bunch to a foil in a further distance to get a rough estimate of the effect of the divergence on the electron bunch, one can include a foil position.
+A foil position which is unequal to zero, adds the electrons momentum vectors onto the electron until they reach the given y-coordinate.
+To contain the longitudinal information of the bunch, the simulation window is actually virtually moved to the foil position and not each single electron.
+
+.. code:: cpp
+
+    namespace SI
+    {
+        // z position of the foil to calculate transition radiation at
```
PrometheusPi: here it should be y as well

PrometheusPi:
```suggestion
        // y position of the foil to calculate transition radiation at
```

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
PrometheusPi:
```suggestion
```

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -0,0 +1,661 @@
...
+            po::options_description& desc
+        )
+        {
+            desc.add_options( )(
+                ( pluginPrefix + ".period" ).c_str( ),
+                po::value< std::string >( &notifyPeriod ),
+                "enable plugin [for each n-th step]"
+            )(
+                ( pluginPrefix + ".omegaList").c_str(),
+                po::value< std::string >( &pathOmegaList )->default_value( "_noPath_" ),
```
PrometheusPi:
```suggestion
```

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -0,0 +1,661 @@
...
+                "path to file containing all frequencies to calculate"
+            );
```
PrometheusPi:
```suggestion
```

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -0,0 +1,661 @@
...
+        MappingDesc * cellDescription = nullptr;
+        std::string notifyPeriod;
+        uint32_t timeStep;
+
+        std::string speciesName;
+        std::string pluginName;
+        std::string pluginPrefix;
+        std::string filenamePrefix;
+        std::string folderTransRad;
+        std::string pathOmegaList;
```
PrometheusPi:
```suggestion
```

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -0,0 +1,661 @@
...
+                incTransRad = new GridBuffer< float_X, DIM1 >(
+                    DataSpace< DIM1 > ( elementsTransitionRadiation( ) ) );
+                cohTransRadPara = new GridBuffer< complex_X, DIM1 >(
+                    DataSpace< DIM1 > ( elementsTransitionRadiation( ) ) );
+                cohTransRadPerp = new GridBuffer< complex_X, DIM1 >(
+                    DataSpace< DIM1 > ( elementsTransitionRadiation( ) ) );
+                numParticles = new GridBuffer< float_X, DIM1 >(
+                    DataSpace< DIM1 > ( elementsTransitionRadiation( ) ) );
+
+                freqInit.Init( pathOmegaList );
```
PrometheusPi:
```suggestion
                freqInit.Init( listFrequencies::listLocation );
```

PrometheusPi:
@finnolec could you squash the last 4 commits (all code review feedback) into one?

PR #3005, opened 2019-07-25
Refactor and modernize implementation of fields

sbastrakov:
This PR provides several small improvements to the implementations of the fields:

- a new base class for `FieldB` and `FieldE` implementations, since they had a substantially identical part
- add missing `inline` qualifiers
- fix shadowing of the virtual method `SimulationFieldHelper::syncToDevice()`, however, I believe it has never been used polymorphically (via a pointer to the base class); so it was a bug but we did not use this behavior 
- occasionally modernize the code style and other minor things

TODO:
- [x] check that I did not break anything
- [x] add documentation, at least to the new code

### DIFF
--- a/include/picongpu/fields/MaxwellSolver/YeePML/Field.tpp
+++ b/include/picongpu/fields/MaxwellSolver/YeePML/Field.tpp
@@ -22,26 +22,10 @@
 
 #include "picongpu/simulation_defines.hpp"
 #include "picongpu/fields/MaxwellSolver/YeePML/Field.hpp"
-#include "picongpu/fields/MaxwellSolver/Solvers.hpp"
-#include "picongpu/particles/traits/GetInterpolation.hpp"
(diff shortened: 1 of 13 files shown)


sbastrakov:
@psychocoderHPC ready for review!

include/picongpu/fields/MaxwellSolver/YeePML/Field.tpp
```diff
@@ -92,9 +65,10 @@ namespace yeePML
         return data->getDeviceBuffer( ).getDataBox( );
     }
 
-    GridBuffer< Field::ValueType, simDim > & Field::getGridBuffer( )
+    EventTask Field::asyncCommunication( EventTask serialEvent )
     {
-        return *data;
+        EventTask eB = data->asyncCommunication( serialEvent );
```
psychocoderHPC: suggestion: `return data->asyncCommunication( serialEvent );` since the name `eB` has for me no obvious meaning.

PR #3006, opened 2019-07-25
Add missing inline qualifiers potentially causing multiple definitions

sbastrakov:
These missing `inline` qualifiers cause multiple definitions of the functions in question, in case it is present in multiple object files. This is one (but not only) of the issues preventing us from using more object files.

For functions defined in header files (so almost all our functions) the rules are as follows:

- `inline` should be part of declaration (not definition)
- for class/struct methods defined at declaration `inline` is applied automatically

This PR catches the cases that my Visual Studio reports on. I assume there might be some more, but probably not too many.

Files related to fields are not fixed in this PR, but as part of #3005.

### DIFF
--- a/include/picongpu/plugins/ResourceLog.hpp
+++ b/include/picongpu/plugins/ResourceLog.hpp
@@ -60,7 +60,7 @@
 namespace boost_swap_impl
 {
     BOOST_GPU_ENABLED
-    void swap(std::string& s1, std::string& s2)
+    inline void swap(std::string& s1, std::string& s2)
     {
 #ifndef __CUDA_ARCH__
         std::swap(s1, s2);
--- a/include/picongpu/particles/synchrotronPhotons/SynchrotronFunctions.hpp
+++ b/include/picongpu/particles/synchrotronPhotons/SynchrotronFunctions.hpp
@@ -102,10 +102,10 @@ private:
 
     /** First synchrotron function
      */
-    float_64 F_1(const float_64 x) const;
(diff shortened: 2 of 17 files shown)


include/picongpu/plugins/ResourceLog.hpp
```diff
@@ -60,7 +60,7 @@
 namespace boost_swap_impl
 {
     BOOST_GPU_ENABLED
-    void swap(std::string& s1, std::string& s2)
+    inline void swap(std::string& s1, std::string& s2)
```
PrometheusPi: Why not `HINLINE` here?

sbastrakov:
@PrometheusPi good point. I mostly used `HINLINE` in files that already have `HINLINE` or `HDINLINE`, and `inline` in contexts that looked to me more pure C++ in a sense of alpaka/device-free. Of course, this is a rather subjective differentiation.

PrometheusPi:
@sbastrakov then I would favor a common naming. @psychocoderHPC what do you think?

include/picongpu/plugins/ResourceLog.hpp
sbastrakov: As I understood from the comment, `BOOST_GPU_ENABLED` can already be `__host__ __device__` (depending on build options) so just adding pure `inline`.

sbastrakov:
Discussed offline with @psychocoderHPC , he agrees with @PrometheusPi , I will fix.

sbastrakov:
@PrometheusPi fixed and rebased, ready for another review.

PR #2856, opened 2019-01-17
Taurus: Power9 V100 Templates

BeyondEspresso:
PIConGPU now runs on the Power9 nodes on Taurus. On the `ml`-partition there are 22 nodes with each 6 V100 GPUs.

A working and tested configuration of the V100-profile, as well as the suitable templates with and without restarting can be found in this pull request.

Boost, HDF5, PNG, pngwriter, c-blosc, adios and splash were required to be self-built. The hint from ComputationalRadiationPhysics/alpaka#702 was used as a workaround to make the code compile. The current MPI-version for the Power9 machines is currently not integrated with SLURM. For this reason we created and used machinefiles in the V100-template V100.tpl.

Now PIConGPU runs and both ADIOS and HDF5-output work. However, the "real-life" HDF5-output for larger files (~100GB) is abysmally slow. Hence, use ADIOS for the heavy-lifting and HDF5 for smaller output (phase-space) etc. .

As a test case, an LWFA production run was used on 108 GPUs and so far produced nice output. Initial crashs seemed to have arisen by an unstable filesystem and so far could not be reproduced. However, the restarting capabilities do work.

At the moment this `ml`-partition is still in the testing phase and will go live by Februrary 1, 2019. The current maximum walltime is 8h.

Enjoy! 😃

PR #3007, opened 2019-07-29
Summit: Profile & Templates

ax3l:
Add a software profile, TBG job template and LSF tutorial for Summit at ORNL.

### To Do

- [x] wait for RT test to complete

### Note

Also see: https://jsrunvisualizer.olcf.ornl.gov/index.html?s4f1o17n6c7g1r11d1b1l0=

### Outdated

The information below is outdated and was fixed via #3084 

~~Our self-paced `hostRank` assignment in https://github.com/ComputationalRadiationPhysics/picongpu/blob/0.4.3/include/pmacc/communication/CommunicatorMPI.hpp#L285-L333 prevents that we can just use the automatic placement in `jsrun` [resource sets](https://www.olcf.ornl.gov/for-users/system-user-guides/summit/summit-user-guide/#job-launcher-(jsrun)) with 7 cores + 1 GPU each.~~

```
.TBG_coresPerGPU=7

jsrun -n !TBG_tasks -g 1 -a 1 -c !TBG_coresPerGPU -b packed:!TBG_coresPerGPU -d packed -l GPU-CPU --smpiargs="-gpu" !TBG_dstPath/input/bin/picongpu !TBG_author !TBG_programParams | tee output 
```

```diff
diff --git a/include/pmacc/communication/CommunicatorMPI.hpp b/include/pmacc/communication/CommunicatorMPI.hpp
index d61730e48..cad2f5492 100644
--- a/include/pmacc/communication/CommunicatorMPI.hpp
+++ b/include/pmacc/communication/CommunicatorMPI.hpp
@@ -153,7 +153,7 @@ public:
         MPI_CHECK(MPI_Cart_create(computing_comm, DIM, dims, periods, 0, &topology));
 
         // 3. update Host rank
-        updateHostRank();
+        //updateHostRank();
 
         //4. update Coordinates
         updateCoordinates();
```

~~We might want to add a picongpu cmd-line option, e.g. `--isolated` or `--pre-placed`, to control this behavior for already well-placed processes in such splendid batch systems.~~

### DIFF
--- a/docs/source/install/profile.rst
+++ b/docs/source/install/profile.rst
@@ -84,6 +84,24 @@ For this profile to work, you need to download the :ref:`PIConGPU source code <i
 .. literalinclude:: profiles/hydra-hzdr/default_picongpu.profile.example
    :language: bash
 
+Summit (ORNL)
+-------------
+
+**System overview:** `link <https://www.olcf.ornl.gov/olcf-resources/compute-systems/summit/>`_
+
+**User guide:** `link <https://www.olcf.ornl.gov/for-users/system-user-guides/summit/>`_
+
+**Production directory:** usually ``$PROJWORK/$proj/`` (`link <https://www.olcf.ornl.gov/for-users/system-user-guides/summit/summit-user-guide/#file-systems>`_).
+Note that ``$HOME`` is mounted on compute nodes as read-only.
+
+For this profile to work, you need to download the :ref:`PIConGPU source code <install-dependencies-picongpu>` and install :ref:`libSplash and PNGwriter <install-dependencies>` manually.
+
+V100 GPUs (recommended)
+^^^^^^^^^^^^^^^^^^^^^^^
+
+.. literalinclude:: profiles/summit-ornl/gpu_picongpu.profile.example
+   :language: bash
+
 Piz Daint (CSCS)
 ----------------
 
--- a/etc/picongpu/summit-ornl/gpu_batch.tpl
+++ b/etc/picongpu/summit-ornl/gpu_batch.tpl
@@ -1,0 +1,91 @@
+#!/usr/bin/env bash
+# Copyright 2019 Axel Huebl, Rene Widera
+#
+# This file is part of PIConGPU.
+#
+# PIConGPU is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# PIConGPU is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PIConGPU.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
+
+# PIConGPU batch script for Summit's LSF (bsub) batch system
+#   https://www.olcf.ornl.gov/for-users/system-user-guides/summit/summit-user-guide/#running-jobs
+
+#BSUB -q !TBG_queue
+#BSUB -W !TBG_wallTimeNoSeconds
+# Sets batch job's name
+#BSUB -J !TBG_jobName
+#BSUB -nnodes !TBG_nodes
+#BSUB -alloc_flags smt4
+# send me mails on job (-B)egin, Fi(-N)ish
+#BSUB !TBG_mailSettings -u !TBG_mailAddress
+#BSUB -cwd !TBG_dstPath
+#BSUB -P !TBG_nameProject
+
+#BSUB -o stdout.%J
+#BSUB -e stderr.%J
+
+
+## calculations will be performed by tbg ##
+.TBG_queue="batch"
+
+# remove seconds from walltime
+.TBG_wallTimeNoSeconds=${TBG_wallTime::-3}
+
+# settings that can be controlled by environment variables before submit
+.TBG_mailSettings=${MY_MAILNOTIFY:-""}
+.TBG_mailAddress=${MY_MAIL:-"someone@example.com"}
+.TBG_author=${MY_NAME:+--author \"${MY_NAME}\"}
+.TBG_nameProject=${proj:-""}
+.TBG_profile=${PIC_PROFILE:-"${PROJWORK}/!TBG_nameProject/${USER}/picongpu.profile"}
+
+# number of available/hosted GPUs per node in the system
+.TBG_numHostedGPUPerNode=6
+
+# required GPUs per node for the current job
+.TBG_gpusPerNode=`if [ $TBG_tasks -gt $TBG_numHostedGPUPerNode ] ; then echo $TBG_numHostedGPUPerNode; else echo $TBG_tasks; fi`
+
+# number of cores to block per GPU - we got 2x22 HW CPU cores per node
+#   and we will be accounted those anyway
+.TBG_coresPerGPU=7
+
+# use ceil to caculate nodes
+.TBG_nodes="$((( TBG_tasks + TBG_gpusPerNode - 1 ) / TBG_gpusPerNode))"
+
+## end calculations ##
+
+echo 'Running program...'
+echo !TBG_jobName
+
+cd !TBG_dstPath
+echo -n "present working directory:"
+pwd
+
+
+source !TBG_profile 2>/dev/null
+if [ $? -ne 0 ] ; then
+  echo "Error: PIConGPU environment profile under \"!TBG_profile\" not found!"
+  exit 1
+fi
+
+mkdir simOutput 2> /dev/null
+cd simOutput
+
+#jsrun  -N 1 -n !TBG_nodes !TBG_dstPath/input/bin/cuda_memtest.sh
+
+#if [ $? -eq 0 ] ; then
+export OMP_NUM_THREADS=!TBG_coresPerGPU
+jsrun --nrs !TBG_tasks --tasks_per_rs 1 --cpu_per_rs !TBG_coresPerGPU --gpu_per_rs 1 --latency_priority GPU-CPU --bind rs --smpiargs="-gpu" !TBG_dstPath/input/bin/picongpu !TBG_author !TBG_programParams | tee output
+# note: instead of the PIConGPU binary, one can also debug starting "js_task_info | sort"
+#fi
--- a/etc/picongpu/summit-ornl/gpu_picongpu.profile.example
+++ b/etc/picongpu/summit-ornl/gpu_picongpu.profile.example
@@ -1,0 +1,66 @@
+# Name and Path of this Script ############################### (DO NOT change!)
+export PIC_PROFILE=$(cd $(dirname $BASH_SOURCE) && pwd)"/"$(basename $BASH_SOURCE)
+
+# User Information ################################# (edit the following lines)
+#   - automatically add your name and contact to output file meta data
+#   - send me a mail on job (-B)egin, Fi(-N)ish
+export MY_MAILNOTIFY=""
+export MY_MAIL="someone@example.com"
+export MY_NAME="$(whoami) <$MY_MAIL>"
+
+# Project Information ######################################## (edit this line)
+#   - project account for computing time
+export proj=<yourProject>
+
+# Text Editor for Tools ###################################### (edit this line)
+#   - examples: "nano", "vim", "emacs -nw", "vi" or without terminal: "gedit"
+#module load nano
+#export EDITOR="nano"
+
+# basic environment ###########################################################
+module load gcc/6.4.0
+
+export CC=$(which gcc)
+export CXX=$(which g++)
+
+# required tools and libs
+module load git
+module load cmake/3.14.2
+module load cuda/10.1.168
+module load boost/1.66.0
+
+# plugins (optional) ##########################################################
+module load hdf5/1.10.3
+module load adios/1.13.1-py2 c-blosc zfp sz lz4
+
+# download libSplash and compile it yourself from
+#   https://github.com/ComputationalRadiationPhysics/libSplash/
+export Splash_DIR=$PROJWORK/$proj/huebl/sw/splash-1.7.0
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$Splash_DIR/lib
+
+#export T3PIO_ROOT=$PROJWORK/$proj/lib/t3pio
+#export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$T3PIO_ROOT/lib
+
+# install pngwriter yourself:
+#   https://github.com/pngwriter/pngwriter#install
+module load zlib/1.2.11
+module load libpng/1.6.34 freetype/2.9.1
+export PNGwriter_DIR=$PROJWORK/$proj/huebl/sw/pngwriter-0.7.0
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$PNGwriter_DIR/lib
+
+# helper variables and tools ##################################################
+export PICSRC=$HOME/src/picongpu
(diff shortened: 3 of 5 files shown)


etc/picongpu/summit-ornl/gpu_picongpu.profile.example
```diff
@@ -0,0 +1,71 @@
...
+#   - project account for computing time
+export proj=<yourProject>
+
+# Text Editor for Tools ###################################### (edit this line)
+#   - examples: "nano", "vim", "emacs -nw", "vi" or without terminal: "gedit"
+#module load nano
+#export EDITOR="nano"
+
+# basic environment ###########################################################
+# source /opt/modules/3.2.6.7/init/bash
```
ax3l: tod: rm me

etc/picongpu/summit-ornl/gpu_picongpu.profile.example
```diff
@@ -0,0 +1,71 @@
...
+module load cmake/3.14.2
+module load cuda/10.1.168
+module load boost/1.66.0
+#export BOOST_ROOT=$BOOST_DIR
+#export MPI_ROOT=$MPICH_DIR
+
+# plugins (optional) ##########################################################
+module load hdf5/1.10.3
+module load adios/1.13.1-py2
+#export HDF5_ROOT=$HDF5_DIR
```
ax3l: todo: rm me

etc/picongpu/summit-ornl/gpu_picongpu.profile.example
```diff
@@ -0,0 +1,71 @@
...
+#export ADIOS_ROOT=$ADIOS_DIR
```
ax3l: todo: rm me

etc/picongpu/summit-ornl/gpu_batch.tpl
```diff
@@ -0,0 +1,89 @@
...
+# Sets batch job's name
+#BSUB -J !TBG_jobName
+#BSUB -nnodes !TBG_nodes
+#BSUB -alloc_flags smt4
+# send me mails on job (-B)egin, Fi(-N)ish
+#BSUB !TBG_mailSettings -u !TBG_mailAddress
+#BSUB -cwd !TBG_dstPath
+#BSUB -P !TBG_nameProject
+
+#BSUB -o stdout.%J
```
ax3l: are we requiring to name those `stdout`/`stderr` without a suffix these days? (post-processing scripts)

ax3l: no, we rely on `simOutput/output`

psychocoderHPC:
@ax3l `updateHostRank` should not destroy the mapping of the resources provided from `jsrun`
For the case that an external tool like `jsrun` is doing the mapping of a GPU to a MPI process we will see [during the dive selection](https://github.com/ComputationalRadiationPhysics/picongpu/blob/1ee0cad7a9f290071dd4422359491e5145a44b1c/include/pmacc/Environment.hpp#L445) only one device and we will than [select this external mapped device](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/include/pmacc/Environment.hpp#L472).

@ax3l Could you please run `jsrun env` and check if jsrun is using the environment variable `CUDA_VISIBLE_DEVICES` to map the GPU to the MPI process.

It could be that https://github.com/ComputationalRadiationPhysics/picongpu/blob/1ee0cad7a9f290071dd4422359491e5145a44b1c/include/pmacc/Environment.hpp#L452 make some trouble. I will check it on our system.

psychocoderHPC:
From https://www.olcf.ornl.gov/for-users/system-user-guides/summit/summit-user-guide/#job-launcher-(jsrun)
```
Job step cgroups are not currently supported (Resolved: February 19, 2019)
A regression was introduced in JSM 10.02.00.10rtm2 that prevents job step cgroups from being created as a result, JSM, is defaulting to setting CUDA_VISIBLE_DEVICES in order to allocate GPUs to specific resource sets. Because of this issue, even if using --gpu_per_rs 0 or -g 0, every resource set in the step will be able to see all 6 GPUs in a node.
```
IMO if you remove `updateHostRank` all MPI ranks will use the same GPU if the access mode for GPUs is set to `default` and if it is set to `one process only` than the mapping will be random.

Do I overlook something?

psychocoderHPC:
@ax3l I think you must set `--gpu_per_rs 1 --tasks_per_rs 1 --cpu_per_rs 7 --rs_per_host 6` to get the pinning correct. I missed this parameter in your current tbg templates.

ax3l:
Yes, that's what I am doing: `jsrun` is properly putting 6 MPI ranks on the same node and properly exposes one GPU per MPI rank. But then our `updateHostRanks` sees the same hostname six times and tries to interate 6 GPUs and only finds one (per rank).

psychocoderHPC:
> Yep, that's what I mean: here it will try to iterate 6 GPUs since the hostname reduction we do updates `hostRank` from `0` to `1-5`.

We only iterate over N devices (in this case 6) if we see 6 devices. If the batchsystem/jsrun already select a device we will use it. The information from `updateHostRanks()` is only used if we see multiple devices.
If we see multiple devices but the device is not process exclusive than we will use the local rank provided by `updateHostRanks()`.

ax3l:
> We only iterate over N devices (in this case 6) if we see 6 devices.

Yes, but in the `updateHostRanks` function we iterate over 6 devices, because we detect 6 ranks on the same host with our MPI reduce, although I want it to stay with one.

psychocoderHPC:
> One therefore aborts in
> https://github.com/ComputationalRadiationPhysics/picongpu/blob/1dfedbd0f0a5a8fcf522f13022b6cca6c822d6ec/include/pmacc/Environment.hpp#L454

This is a bug if CUDA_VISIABLE_DEVICES is set this check should not be performed or we need to call `setDevice` with the correct gpu index. Removing `updateHostRanks` would solve it but we should not manipulate a MPI function from PMacc to solve an CUDA issue.

The suggested parameter `--isolated` can be used to call `setDevice` correctly but I think we can derive the information without explicit user intervention.

psychocoderHPC:
@ax3l if you remove for your tests the following lines the gpus will be selected correct.https://github.com/ComputationalRadiationPhysics/picongpu/blob/1dfedbd0f0a5a8fcf522f13022b6cca6c822d6ec/include/pmacc/Environment.hpp#L452-L456

ax3l:
Yep, we can probably remove this or at least set a control flag for it. E.g. on Summit, running an Nvidia `mps` server is also supported which this check would not consider as well. Tests submitted, let's see if we also have to set the `deviceNumber` back to zero per MPI rank with explicitly set visible devices.

PrometheusPi:
@ax3l What is the status of the run time tests (required in your initial description)?

ax3l:
Debugging some runtime MPI startup issue that is not visible with a minimal hello world example and the same parameters.

ax3l:
After updating layout with 

and keeping the [mentioned patch](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3007#issuecomment-517330914).
LWFA `1.cfg` aborts with #3034 and `4.cfg` shows in `stderr`:
```
[CUDA] Error: </ccs/home/huebl/src/picongpu/include/pmacc/../pmacc/Environment.hpp>:462 
[CUDA] Error: </ccs/home/huebl/src/picongpu/include/pmacc/../pmacc/Environment.hpp>:462 
[CUDA] Error: </ccs/home/huebl/src/picongpu/include/pmacc/../pmacc/Environment.hpp>:462 
Unhandled exception of type 'St13runtime_error' with message '[CUDA] Error: not defined cuplaError', terminating
Unhandled exception of type 'St13runtime_error' with message '[CUDA] Error: not defined cuplaError', terminating
Unhandled exception of type 'St13runtime_error' with message '[CUDA] Error: not defined cuplaError', terminating
```

etc/picongpu/summit-ornl/gpu_batch.tpl
```diff
@@ -0,0 +1,91 @@
...
+fi
+
+mkdir simOutput 2> /dev/null
+cd simOutput
+
+#jsrun  -N 1 -n !TBG_nodes !TBG_dstPath/input/bin/cuda_memtest.sh
+
+#if [ $? -eq 0 ] ; then
+export OMP_NUM_THREADS=7
+jsrun --nrs !TBG_tasks --tasks_per_rs 1 --cpu_per_rs 7 --gpu_per_rs 1 --rs_per_host !TBG_gpusPerNode --latency_priority GPU-CPU --bind rs --smpiargs="-gpu" !TBG_dstPath/input/bin/picongpu !TBG_author !TBG_programParams | tee output
```
ax3l: cosmetic:
```suggestion
jsrun --nrs !TBG_tasks --tasks_per_rs 1 --cpu_per_rs !TBG_coresPerGPU --gpu_per_rs 1 --rs_per_host !TBG_gpusPerNode --latency_priority GPU-CPU --bind rs --smpiargs="-gpu" !TBG_dstPath/input/bin/picongpu !TBG_author !TBG_programParams | tee output
```

psychocoderHPC:
> After updating layout with https://jsrunvisualizer.olcf.ornl.gov/index.html?s4f1o17n6c7g1r11d1b1l0=
> 
> and keeping the [mentioned patch](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3007#issuecomment-517330914).
> LWFA `1.cfg` aborts with #3034 and `4.cfg` shows in `stderr`:
> 
> ```
> [CUDA] Error: </ccs/home/huebl/src/picongpu/include/pmacc/../pmacc/Environment.hpp>:462 
> [CUDA] Error: </ccs/home/huebl/src/picongpu/include/pmacc/../pmacc/Environment.hpp>:462 
> [CUDA] Error: </ccs/home/huebl/src/picongpu/include/pmacc/../pmacc/Environment.hpp>:462 
> Unhandled exception of type 'St13runtime_error' with message '[CUDA] Error: not defined cuplaError', terminating
> Unhandled exception of type 'St13runtime_error' with message '[CUDA] Error: not defined cuplaError', terminating
> Unhandled exception of type 'St13runtime_error' with message '[CUDA] Error: not defined cuplaError', terminating
> ```

can you please add ` log<ggLog::CUDA_RT > ("try %1% of %2%") % tryDeviceId % num_gpus;` before line 462 in Environment.hpp (before `cudaSetDevice`) and active the PMacc verbose level 16 in cmake

etc/picongpu/summit-ornl/gpu_batch.tpl
ax3l: necessary change: with the current startup script and executing e.g. `16` GPUs the startup fails with:
```
Error: nrs (16) should be evenly divided rs_per_host (6)
```

cc @arghyac007

arghyac007: Correct, if you use the -r feature it needs to be divisible by the -n, uneven distribution is not allowed. Wrt, your last comment, if you want to say run 9 resource sets, and just run -n 9, you’re guaranteed that all of the first node will be used and only 3 from the second node will be used (which 3, on which socket also cannot be controlled). 

The -r is mostly used if for some reason you want to use the whole CPU cores (42) but not all of the GPUs on one node, and maybe just 3 GPUs, one way to move the rest of the computation to the next node will be by using the -r.

ax3l: So you say we can just omit the `-r|--rs_per_host`?

The problem I face: I *wan* to use 16 GPUs for a problem (`--nrs`) but that will keep some nodes *intentionally* underutilized.

arghyac007: Yup. You don’t need the -r, I would just use -n 16 -g 1. 

Since the allocation calculation is done in node hours, you always get exclusive node access, no one else will be able to use those nodes, so even if you’re using 2 GPUs on that node you’re billed for the entire node (hr).

ax3l: Thanks, testing right now.

Yes, I am aware of the full-node allocation policy as on titan and it makes a lot of sense. I am just thinking about large-scale simulations where one might have constrains in number of devices by static load balancing and where using 8000/8004 allocated GPUs is easier than finding a perfect match of 6 as well as decomposition :) (Just to keep things flexible for users, they should still try to use allocated resources efficiently, of course.)

etc/picongpu/summit-ornl/gpu_batch.tpl
```diff
@@ -0,0 +1,91 @@
...
+  exit 1
+fi
+
+mkdir simOutput 2> /dev/null
+cd simOutput
+
+#jsrun  -N 1 -n !TBG_nodes !TBG_dstPath/input/bin/cuda_memtest.sh
+
+#if [ $? -eq 0 ] ; then
+export OMP_NUM_THREADS=!TBG_coresPerGPU
```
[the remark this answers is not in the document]
sbastrakov: Sure. On JUWELS 2 threads per core was a little faster.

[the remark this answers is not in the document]
ax3l: The new Taurus ML nodes at TUD are literally identical to the summit nodes. Maybe adding a new profile will also help debugging the one here.

ax3l:
I currently do not have much time to debug this further, it's on my list but has been overtaken by other priorities for the upcoming month at least.
The new Taurus ML nodes have exactly the same hardware, potentially adding a profile for it will help with this PR as well.

n01r:
@ax3l The profile for the Taurus ML nodes was added in #2856, updated in #2903 and can be found [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/etc/picongpu/taurus-tud/V100_picongpu.profile.example) :)

etc/picongpu/summit-ornl/gpu_batch.tpl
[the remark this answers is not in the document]
ax3l: Thx! We should add a host-side setup after this PR as well and should places this in both.

docs/source/install/profile.rst
```diff
@@ -84,6 +84,25 @@ For this profile to work, you need to download the :ref:`PIConGPU source code <i
...
 
+Summit (ORNL)
+-------------
+
+**System overview:** `link <https://www.olcf.ornl.gov/olcf-resources/compute-systems/summit/>`_
+
+**User guide:** `link <https://www.olcf.ornl.gov/for-users/system-user-guides/summit/>`_
+
+**Production directory:** usually ``$PROJWORK/$proj/`` (`link <https://www.olcf.ornl.gov/for-users/system-user-guides/summit/summit-user-guide/#file-systems>`_).
+Note that ``$HOME`` is not mounted on compute nodes as read-only.
```
PrometheusPi: This sentence is a bit confusing. "not ... read-only" could mean *read-and-write* but you mean it is not mounted at all. Just remove ~`as read-only`~.

ax3l:
```suggestion
Note that ``$HOME`` is mounted on compute nodes as read-only.
```

PrometheusPi: @ax3l While I was confused by the grammar I overlooked that the statement was actually wrong. Great catch. It **is** read-only according the the [documentation](https://www.olcf.ornl.gov/for-users/system-user-guides/summit/summit-user-guide/#storage-overview).

psychocoderHPC:
The problem why this PR is currently not working i a bug in PMacc during the device selection if the device is preselected with e.g the environment variable `CUDA_VISIBLE_DEVICES`.

https://github.com/ComputationalRadiationPhysics/picongpu/blob/7311937b06cc94d8a09fb10423b8f4568507641e/include/pmacc/Environment.hpp#L460-L466

This lines must be moved down into the `for` loop and `tryDeviceId` instead of `deviceNumber` must be used.
I will provide the fix soon (hopefully today)

psychocoderHPC:
After applying  #3084 this PR work fine!

psychocoderHPC:
The patch in the initial description is not needed after #3084 is merged. This PR forks together with #3084 as it is.

ax3l:
PR rebased. Happy computing!

psychocoderHPC:
@ax3l thanks for this PR +1

Issue #2714, opened 2018-10-05
CUDA 9.2: Boost PropTree Error in ResourceLog


ax3l:
In the combination of
- GCC 4.9.4
- CUDA 9.2.88
- Boost 1.65.1 (both C++11 build and C++98 "default" build)
- PIConGPU 0.4.0-rc2
- openSUSE Linux Workstation

@ajitup73 in #2704 sees the following compile error in our `ResourceLog` plugin:

```
/home/tempo/src/spack/opt/spack/linux-opensuse42-x86_64/gcc-4.9.4/boost-1.65.1-dy5wuhsxhwb6imd6maxwaeo2jd7ppfsr/include/boost/core/swap.hpp(36): warning: calling a __host__ function from a __host__ __device__ function is not allowed
          detected during:
            instantiation of "void boost_swap_impl::swap_impl(T &, T &) [with T=std::string]" 
(56): here
            instantiation of "void boost::swap(T1 &, T2 &) [with T1=std::string, T2=std::string]" 
/home/tempo/src/spack/opt/spack/linux-opensuse42-x86_64/gcc-4.9.4/boost-1.65.1-dy5wuhsxhwb6imd6maxwaeo2jd7ppfsr/include/boost/property_tree/detail/ptree_implementation.hpp(214): here
            instantiation of "void boost::property_tree::basic_ptree<Key, Data, KeyCompare>::swap(boost::property_tree::basic_ptree<Key, Data, KeyCompare>::self_type &) [with Key=std::string, Data=std::string, KeyCompare=std::less<std::string>]" 
/home/tempo/src/spack/opt/spack/linux-opensuse42-x86_64/gcc-4.9.4/boost-1.65.1-dy5wuhsxhwb6imd6maxwaeo2jd7ppfsr/include/boost/property_tree/detail/ptree_implementation.hpp(201): here
            instantiation of "boost::property_tree::basic_ptree<Key, Data, KeyCompare>::self_type &boost::property_tree::basic_ptree<Key, Data, KeyCompare>::operator=(const boost::property_tree::basic_ptree<Key, Data, KeyCompare>::self_type &) [with Key=std::string, Data=std::string, KeyCompare=std::less<std::string>]" 
[...]

PR #3013, opened 2019-07-31
Resource Monitor Plugin: Warnings

### LINKED ISSUE #3011: ResourceManager: Host & Device Split
If we manage to split the `ResourceManager` plugin properly into host code path (`.cpp`) and partly device code, we will be able to get rid of many NVCC related warnings from boost property tree and MPL, that currently clutter our compiles.

Refs.:
- https://github.com/boostorg/multi_index/pull/26
- https://github.com/boostorg/mpl/issues/45

ax3l:
Move the boost property tree for writing of the final
files out of the NVCC compile path. This avoids awkward work-arounds (see #2714), fixes NVCC-caused warnings in host-code (fix #3011) and reduces compile-time by providing a little more parallelism.

Also fixes a potential use-after-free (`temporary.str().c_str()`).

### DIFF
--- a/include/picongpu/plugins/ResourceLog.hpp
+++ b/include/picongpu/plugins/ResourceLog.hpp
@@ -32,9 +32,6 @@
 #include "picongpu/particles/filter/filter.hpp"
 
 // Boost
-#include <boost/property_tree/ptree.hpp>
-#include <boost/property_tree/json_parser.hpp>
-#include <boost/property_tree/xml_parser.hpp>
 #include <boost/filesystem.hpp>
 
 // STL
@@ -44,34 +41,25 @@
 #include <sstream>   /* std::stringstream */
 #include <fstream>   /* std::filebuf */
 #include <map>       /* std::map */
+#include <algorithm> /* std::accumulate */
 
 // C LIB
 #include <stdlib.h> /* itoa */
 #include <stdint.h> /* uint32_t */
 
-/* provide a specialization of swap for std::string
- *
- * workaround for: https://github.com/ComputationalRadiationPhysics/picongpu/issues/2714
- * Boost is shipping a swap implementation with support for device
- * if `BOOST_GPU_ENABLED` is `__host__ __device__` but is calling a pure host function `std::swap`
- * within the device code. Even if swap is not called on the device this implementation
- * can pull host only code inside the device compile path of CUDA.
- */
-namespace boost_swap_impl
-{
-    BOOST_GPU_ENABLED
-    inline void swap(std::string& s1, std::string& s2)
-    {
-#ifndef __CUDA_ARCH__
-        std::swap(s1, s2);
-#endif
-    }
-}
 
 namespace picongpu
 {
-    using namespace pmacc;
+    using namespace pmacc; /** @todo do not pull into global (header) scope */
 
+namespace detail
+{
+    std::stringstream
+    writeMapToPropertyTree(
+        std::map<std::string, size_t> valueMap,
+        std::string outputFormat
+    );
+}
 
     class ResourceLog : public ILightweightPlugin
     {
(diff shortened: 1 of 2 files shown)


include/picongpu/plugins/ResourceLog.hpp
```diff
@@ -103,67 +90,39 @@ namespace picongpu
...
+                valueMap["resourceLog.rank"] = static_cast<size_t>(Environment<simDim>::get().GridController().getGlobalRank());
 
             if(contains(propertyMap,"position"))
             {
-                DataSpace<simDim> currentPosition = Environment<simDim>::get().GridController().getPosition();
-                pt.put("resourceLog.position.x", currentPosition[0]);
-                pt.put("resourceLog.position.y", currentPosition[1]);
-                pt.put("resourceLog.position.z", currentPosition[2]);
+                auto const currentPosition = Environment<simDim>::get().GridController().getPosition();
+                char coord = 'x';
```
sbastrakov: (Not important) I find an array of 'x', 'y', 'z' a little cleaner than the character arithmetic inside a loop, smth like that:
```cpp
char const axisName[] = { 'x', 'y', 'z' };
for( size_t d = 0; d < simDim; ++d )
    valueMap[std::string("resourceLog.position.") + axisName[d]] = currentPosition[d];
```

include/picongpu/plugins/ResourceLog.hpp
```diff
@@ -103,67 +90,39 @@ namespace picongpu
...
-            }
+                valueMap["resourceLog.cellCount"] = resourceMonitor.getCellCount();
 
             if(contains(propertyMap,"particleCount"))
             {
                 // enforce that the filter interface is fulfilled
                 particles::filter::IUnary< particles::filter::All > parFilter{ currentStep };
                 std::vector<size_t> particleCounts = resourceMonitor.getParticleCounts<VectorAllSpecies>(*cellDescription, parFilter );
-                pt.put("resourceLog.particleCount", std::accumulate(particleCounts.begin(), particleCounts.end(), 0));
+                valueMap["resourceLog.particleCount"] = std::accumulate(particleCounts.begin(), particleCounts.end(), 0);
```
sbastrakov: Missing `#include <algorithm>` for `std::accumulate`

include/picongpu/plugins/ResourceLog.hpp
```diff
@@ -49,29 +46,19 @@
...
-#ifndef __CUDA_ARCH__
-        std::swap(s1, s2);
-#endif
-    }
-}
 
 namespace picongpu
 {
-    using namespace pmacc;
+    using namespace pmacc; /* fixme: do not pull into global scope */
```
psychocoderHPC: could you please add `@todo`

sbastrakov:
@ax3l thanks! Just some nitpicking comments.

include/picongpu/plugins/ResourceLog.hpp
ax3l: Great idea, I love explicitly named things :)

include/picongpu/plugins/ResourceLog.hpp
```diff
@@ -239,7 +196,8 @@ namespace picongpu
                     std::stringstream ss;
                     ss << outputFilePrefix << rank;
                     boost::filesystem::path resourceLogPath(ss.str());
-                    fileBuf.open(resourceLogPath.string().c_str(), std::ios::out);
```
psychocoderHPC: Note: There was no need to change this because the pointer from `c_str()` life as long as the expression is executed.
The construct is only wrong if you write:

```
auto tmp = resourceLogPath.string().c_str();
fileBuf.open(tmp, std::ios::out);
```

ax3l: That's true.

PR #3014, opened 2019-07-31
Pass mapping description by value instead of pointer from simulation stages

sbastrakov:
It is potentially unsafe to pass by pointer in these cases, as we pass a pointer to a local member and inside it could start a kernel asynchronously, then return and destroy the stage class instance. This does not lead to problems now but just unnecessary unsafe and passing by value is more natural anyways.


ax3l:
Could potentially also be passed as `const` in those places, no?

sbastrakov:
@ax3l yes, to be consistent with the style I should have made those `const`. (This does not affect the interface of the functions in question though).

PR #3016, opened 2019-08-06
hemera-hzdr: remove project definition

psychocoderHPC:
On hemera we have no projects like on other system, we have access groups for a few queues.
Since the slurm option `-A` is queue specific and not user-specific (will be never changed by the user) it makes no sense to have it changeable.
The current implementation also breaks the option `tbg -t anytemplate.tpl` because it is not possible to switch the queue via the explicit selection of the used template.


PrometheusPi:
@psychocoderHPC Where is `TBG_queue` set?

n01r:
@PrometheusPi it is set as `.TBG_queue` in the `.tpl` files.

PR #3017, opened 2019-08-09
Singularity: Workaround Chmod Issue, No UCX

ax3l:
Found by @AdamSimpson (Nvidia) et al., singularity incorrectly sets the permissions of some dirs and files when not `chmod`-ed in the same `RUN` statement.

https://github.com/sylabs/singularity/issues/3880

We just fuse them for now.

Also, disable the UCX backend in OpenMPI due to known issues:
- https://github.com/open-mpi/ompi/issues/4948#issuecomment-512987494
- https://github.com/openucx/ucx/issues/3545

### DIFF
--- a/share/picongpu/dockerfiles/ubuntu-1604/packages.yaml
+++ b/share/picongpu/dockerfiles/ubuntu-1604/packages.yaml
@@ -13,7 +13,7 @@ packages:
     buildable: False
   openmpi:
     version: [3.1.3]
-    variants: +cuda fabrics=verbs,libfabric
+    variants: +cuda fabrics=libfabric
   hwloc:
     variants: +cuda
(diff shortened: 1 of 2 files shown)


share/picongpu/dockerfiles/ubuntu-1604/packages.yaml
```diff
@@ -13,7 +13,7 @@ packages:
     buildable: False
   openmpi:
     version: [3.1.3]
-    variants: +cuda fabrics=verbs,ucx,libfabric
+    variants: +cuda fabrics=verbs,libfabric
```
ax3l:
```suggestion
    variants: +cuda fabrics=libfabric
```

Issue #3020, opened 2019-08-14
radiation plugin: no hdf5 output if totalRadiation flag is not set


PrometheusPi:
The following bug was discovered by @Anton-Le :
The radiation plugin writes out hdf5 data for checkpointing and as an advanced output format that still contains the phase of the emitted radiation. However, if [`--<species>_radiation.totalRadiation`](https://picongpu.readthedocs.io/en/0.4.3/usage/plugins/radiation.html#cfg-file) is is not set, the output is not generated - even the *directory for the regular hdf5 output* is not generated during startup. With `--<species>_radiation.totalRadiation` enabled, everything works fine.  

Thanks @Anton-Le for finding this bug.

PrometheusPi:
Interstingly, for @Anton-Le the simulation just hang and did not crash.
I tested it on the `k20` partion of `hemera` at HZDR and ther it charsed with the following error message:
<details>

```
[1,0]<stderr>:[SPLASH_LOG:0] Exception: Exception for HandleMgr::get: Failed to create file (radiationHDF5/e_radAmplitudes_2800_0_0_0.h5)
[1,0]<stderr>:Unhandled exception of type 'N6splash11DCExceptionE' with message 'Exception for HandleMgr::get: Failed to create file (radiationHDF5/e_radAmplitudes_2800_0_0_0.h5)', terminating
[1,0]<stderr>:[kepler003:28937] *** Process received signal ***
[1,0]<stderr>:[kepler003:28937] Signal: Segmentation fault (11)
[1,0]<stderr>:[kepler003:28937] Signal code:  (128)
[1,0]<stderr>:[kepler003:28937] Failing at address: (nil)
```

</detail>

PrometheusPi:
Update:
I just spoke *offline* with @Anton-Le - the simulation did not crash but just stopped working with `mvapich` as MPI. With `openmpi` it did crash after encountering an error. (According to @Anton-Le  `mvapich` should communicate with the SLURM scheduler when encountering an error - which seems to fail and causes the hanging.)

ax3l:
ping @PrometheusPi: can be closed with #3021 merged?

PrometheusPi:
@ax3l Thanks for the reminder.

Anton-Le:
Since some resources were free on the system I could run a comparison on I tried to reproduce the problem with MVAPICH and running on CPUs only (no GPUs in the machine).
[...]

PR #3021, opened 2019-08-14
Radiation Plugin: fix bool conditions for hdf5 output

PrometheusPi:
This pull request fixes issue #3020. It rearranges the dependencies for creating specific (hdf5 related) output directories and arrays. Thus hdf5 output is now always produced, even if `--<species>_radiation.totalRadiation` is not set. 

The following run time test still need to be performed:

- [x] compare `dev` and `pr` total radiation (update: and hdf5)
- [x] compare `dev` and `pr` lastRadiation and hdf5 output when omitting the `--<species>_radiation.totalRadiation` flag (update: of course only in der `pr` branch)
- [x] compare `dev` and `pr` totalRadiation and hdf5 output after a restart

For testing the default Bunch example will be used. 

@Anton-Le this should solve your issue.

### DIFF
--- a/include/picongpu/plugins/radiation/Radiation.hpp
+++ b/include/picongpu/plugins/radiation/Radiation.hpp
@@ -326,6 +326,8 @@ private:
                 if (isMaster)
                 {
                     timeSumArray = new Amplitude[elements_amplitude()];
+                    for (unsigned int i = 0; i < elements_amplitude(); ++i)
+                        timeSumArray[i] = Amplitude::zero();
 
                     /* save detector position / observation direction */
                     detectorPositions = new vector_64[parameters::N_observer];
@@ -343,7 +345,7 @@ private:
 
                 }
 
-                if (isMaster && totalRad)
+                if (isMaster)
                 {
                     fs.createDirectory("radiationHDF5");
                     fs.setDirectoryPermissions("radiationHDF5");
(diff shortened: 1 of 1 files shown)


PrometheusPi:
Comparing total radiation between `dev` and `pr`:

**logarithmic**
grafik

**linear**
grafik

The noise seen in the linear plot originates from the random particle distribution (which differs between the two simulation runs). 

**linear smoothed**
grafik

The change in code did not effect the result.

Anton-Le:
Great. I'll try it later on. As for the weird behaviour of MVAPICH i'll try it on a different cluster - time permitting - to see if it's really due to the SLURM-weirdness on _hemera_ or just MVAPICH itself.

PrometheusPi:
Same is true for hdf5 output:

**logarithmic**
grafik

**linear**
grafik

**linear smoothed**
grafik

n01r:
I will merge after your remaining runtime checks pass

PrometheusPi:
last looks also similar  - there is no characteristic frequency distribution, since the duration is only 10 iterations while a laser cycle is much longer 

grafik

PrometheusPi:
Same appears to be true for the `lastRadiation` output `hdf5` output:

**logarithmic**
grafik

**linear smoothed**
grafik

PrometheusPi:
After a restart, everything looks fine:

**ASCII output:**
**logarithmic**
grafik

**linear**
grafik

**linear smoothed**
grafik


**hdf5 data:**
**logarithmic**
grafik

**linear**
grafik

**linear smoothed**
grafik

PrometheusPi:
@n01r: All run time tests have past successfully.

n01r:
Great work, thanks!

Issue #2943, opened 2019-04-11
Can a moving window be set at run time only?


sbastrakov:
Currently the moving window setup uses a combination of compile-time and run-time parameters: `movePoint` at `grid.param` and `-m` option in `.cfg` files. Now all simulations, even those never needing a moving window, have to set the `movePoint`. From the way this variable is used in the code, it seems there is no constraint on it being a constexpr. So it can be be put to a `.cfg` file, maybe smth like `-m 0.90` for simulations that need it. And other simulations will have one non intuitive variable less in `.param` files.

Of course, this would require changes in our examples and user input files, so not backwards-compatible.

ax3l:
Yes, absolutely. This and many more variables should become runtime parameters to avoid recompilation.

sbastrakov:
@ax3l so do you think it is better to gradually change this and other similar stuff as a low-priority task, or to collect a bunch of such changes and do it all together?

ax3l:
I would say gradually refactor, that's what we have `dev` for.

psychocoderHPC:
implemented in  #3022

PR #3022, opened 2019-08-14
Add a runtime parameter for window move point

sbastrakov:
Solves #2943.

I spent some time considering how to do such changes without breaking the user sets of input files. The issue is that just adding a new runtime parameter and removing the dependency on the compile-time one would make old files silently produce different results, which is far from desired. So this PR goes for a compromise. Now the old compile-time variable is still required, its value is used as a default for the new runtime parameter. In case the moving window is enabled but the new value is not set (so the setup relies on the old behavior) it would still work as usual for now, but there is an explanatory warning in the log. Then we can remove support for the old way e.g. one release after this change is added.

Also noted a description of how the moving window works seems to be missing in the readthedocs.  However, not subject of this PR to fix it.

Edit: with these changes, some of our standard examples will also trigger the warning. This will be fixed after this PR is merged.

### DIFF
--- a/include/picongpu/ArgsParser.cpp
+++ b/include/picongpu/ArgsParser.cpp
@@ -19,6 +19,7 @@
  */
 
 #include "picongpu/ArgsParser.hpp"
+#include "picongpu/debug/PIConGPUVerbose.hpp"
 #include "picongpu/versionFormat.hpp"
 
 #include <boost/program_options.hpp>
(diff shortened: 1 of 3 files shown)


sbastrakov:
@psychocoderHPC your suggestion is implemented.

ax3l:
> Edit: with these changes, some of our standard examples will also trigger the warning. This will be fixed after this PR is merged.

Ideally, add such a commit to the PR. Now all things I use today are throwing warnings ;)

sbastrakov:
@ax3l I was thinking of that, but then the PR kinda does two things. Also, for those using the custom setups (so basically all use except testing), a user has to do it anyways in their files.

ax3l:
Yep, the idea is to keep a PR as a consistent unit that changes a thing properly.
If one e.g. changes an API and that affects also its usage, just group it into two commits that can be reviewed one by one in the PR and keep the code in a consistent state at merge.

It's always an individual decision when a split in two PRs makes sense, just wanted to communicate the basic idea again.

ax3l:
@ComputationalRadiationPhysics/picongpu-maintainers the last notes are a general reminder that our automation, docs and examples only make sense in our workflows if you change them with the PR that changes code and features. Just place them in a separate commit.

Otherwise, they will be disjoint like writing a wiki documentation, become outdated, will be forgotten and/or leave the development branch in a half-updated state. Thanks! :)

ax3l:
@sbastrakov Is it possible this warning is still thrown?

include/picongpu/ArgsParser.cpp
```diff
@@ -33,6 +34,45 @@
...
+
+namespace
+{
+
+    /** Report deprecated parameters
+     *
+     * This function is meant to handle cases when some parameters are changed
+     * but the old versions temporarily kept for backward compatibility and
+     * deprecated. Notably, this applies to compile-time parameters getting a
+     * run-time version. Hence it deliberately ignores incapsulation and code
```
ax3l: typo: encapsulation

sbastrakov:
@ax3l I agree with your objections, except I really do not see how setting 2 variables in one file which can be reduced to a single variable is inferior to (now) having same 2 variables in the same file and both run-time.

bussmann:
Hi,

It is not about the number of variables but how you ensure backward 
compatibility and workflow. Please adapt the PR accordingly.

Thanks, Michael

Am 02.09.2019 um 13:14 schrieb Sergei Bastrakov:

Issue #3027, opened 2019-08-20
Out of memory on init with PML


n01r:
I wanted to run a live simulation on the P100s (16GB global device memory) at HZDR and encountered a runtime error after I changed the boundary conditions to PML.

<details>
<summary>Simulation Parameters</summary>

- Domain decomposition: 4 x 2 x 2 = 16 GPUs
- Global domain: 384 x 720 x 384 cells
- Particle shape = P4S
- Current deposition: EmZ
- Field solver: YeePML
- maximum of 64 cells in longitudinal direction (per GPU) filled with target upon initialization
- target spans the whole box transversally
- 1 ion per cell and 10 electrons per cell are initialized

</details>

With the `MemoryCalculator`, upon manually accounting for the boundary conditions as well, I arrive at the following memory requirements:

```
Total memory required on initialization: 1561 MB
- RNG: 152 MB
- ions: 39 MB
- electrons: 349 MB
- fields: 340 MB
- boundary conditions (PML): 681 MB
``` 

Even though this looks perfectly manageable for the P100s, my sim fails immediately with the following error message.

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  /home/garten70/src/picongpu/thirdParty/alpaka/include/alpaka/mem/buf/BufCpu.hpp(520) 'cudaHostUnregister( const_cast<void *>(reinterpret_cast<void const *>(bufImpl.m_pMem)))' A previous CUDA call (not this one) set the error  : 'cudaErrorMemoryAllocation': 'out of memory'!
[gp006:184631] *** Process received signal ***
[gp006:184631] Signal: Aborted (6)
[gp006:184631] Signal code:  (-6)
[gp006:184631] [ 0] /usr/lib64/libpthread.so.0(+0xf6d0)[0x2aaaaacde6d0]
[gp006:184631] [ 1] /usr/lib64/libc.so.6(gsignal+0x37)[0x2aaaae3ed277]
[gp006:184631] [ 2] /usr/lib64/libc.so.6(abort+0x148)[0x2aaaae3ee968]
[gp006:184631] [ 3] /trinity/shared/pkg/compiler/gcc/7.3.0/lib64/libstdc++.so.6(_ZN9__gnu_cxx27__verbose_terminate_handlerEv+0x125)[0x2aaaadc80ea5]
[...]

PR #3028, opened 2019-08-21
Fix field solver initialization order

sbastrakov:
When developing PML I did not think of the memory allocation order and continued using the old one with field solver being constructed after the particle memory allocation (and my tests were small enough that it worked fine). But since the `YeePML` solver involves allocation of additional field components that lead to out-of-memory error #3027.

Imo there is also room for improvement in terms of design, as such limitations are not obvious at all. I could try to refactor it in a follow-up PR.


n01r:
I'm running my sim with these changes now and I didn't get the crash in the beginning. :sparkle:

sbastrakov:
The checks passed, ready to be merged.

PR #3029, opened 2019-08-21
Add an option to account for PML in the memory calculator

### LINKED ISSUE #3026: Memory Calculator: include PML memory complexity
With the PML boundary conditions becoming a mainline feature, the `MemoryCalculator` needs an update as well.
Especially since using PMLs currently requires significantly more memory that not using them.

sbastrakov:
Resolves #3026.


sbastrakov:
@n01r good point. I also noticed that the list of available field solvers in the comments of our example setups does not contain `YeePML`, except for the example I made. I will also fix that (in a separate PR).

PR #3030, opened 2019-08-21
Remove unused pmacc::IDataSorter

sbastrakov:
This looks like some old-time design that is not used anymore. While doing this, I also noticed `pmacc::AbstractInitializer` and `pmacc::DataConnector::initialise()` that are technically used, but it probably has little sense now and those can be also removed easily.


sbastrakov:
@n01r do you mind pressing the merge button?

n01r:
@sbastrakov haha, I wouldn't have, however, I didn't want to do it before @psychocoderHPC had reviewed it since I only saw that it was formally correct but I didn't know if there were perhaps more planned uses of this class.

PR #3031, opened 2019-08-22
add reset PML fields in MySimulation::resetAll() 

sbastrakov:
This bug caused incorrect behavior in restart mode for visualization, as observed by @n01r.

After some consideration, I went for a more hard-coded solution rather than trying a general one somewhere in `DataConnector`. The reason is that our class hierarchies do not interact too well in this case with `ISimulationData` not providing `reset()`.

### DIFF
--- a/include/picongpu/simulation/control/MySimulation.hpp
+++ b/include/picongpu/simulation/control/MySimulation.hpp
@@ -24,6 +24,8 @@
 #include <pmacc/verify.hpp>
 #include <pmacc/assert.hpp>
 
+#include <algorithm>
+#include <array>
 #include <string>
 #include <vector>
 #include <boost/lexical_cast.hpp>
@@ -48,6 +50,7 @@
 #include "picongpu/fields/FieldJ.hpp"
 #include "picongpu/fields/FieldTmp.hpp"
 #include "picongpu/fields/MaxwellSolver/Solvers.hpp"
+#include "picongpu/fields/MaxwellSolver/YeePML/Field.hpp"
 #include "picongpu/fields/background/cellwiseOperation.hpp"
 #include "picongpu/initialization/IInitPlugin.hpp"
 #include "picongpu/initialization/ParserGridDistribution.hpp"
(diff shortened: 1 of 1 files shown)


n01r:
Just because it is so pretty :sparkles: ... this is the result of this bug which causes the old field values of the PML boundary conditions to be present right at the beginning of the new simulation.
2019_08_21_ISAAC_foil_PML_loop_restart

include/picongpu/simulation/control/MySimulation.hpp
```diff
@@ -668,6 +665,46 @@ class MySimulation : public SimulationHelper<simDim>
...
+                auto field = std::dynamic_pointer_cast< FieldHelper >(
+                    dc.get< ISimulationData >( name, true )
+                );
+                if( field )
+                    field->reset( currentStep );
+                dc.releaseData( name );
+            }
+        };
+
+        /* For now the list of fields is hardcoded here, a more generic solution
```
n01r: You could add a `@todo` here so that we can keep track of it with Doxygen.   
Otherwise, when the `DataConnector` changes we might miss this, later.

sbastrakov:
@n01r no, so far I just checked that calling a function for a restart artificially from inside the code now does reset the PML fields. You are right that actually running a setup with `--checkpoint.restart.loop` would be a proper test.

include/picongpu/simulation/control/MySimulation.hpp
sbastrakov: Good point, fixed and rebased.

include/picongpu/simulation/control/MySimulation.hpp
```diff
@@ -574,18 +577,12 @@ class MySimulation : public SimulationHelper<simDim>
...
     {
-        DataConnector &dc = Environment<>::get().DataConnector();
-
-        auto fieldE = dc.get< FieldE >( FieldE::getName(), true );
-        auto fieldB = dc.get< FieldB >( FieldB::getName(), true );
-
-        fieldB->reset(currentStep);
-        fieldE->reset(currentStep);
-        meta::ForEach< VectorAllSpecies, particles::CallReset< bmpl::_1 > > callReset;
-        callReset( currentStep );
```
PrometheusPi: I am a bit confused by the naming. In the old version you call `callReset` for all species, but now it is called `resetParticles` but I do not see a renaming of this method/function in the new code.

n01r: The line above defines `resetParticles` and still uses `CallReset` in the template parameters.

PrometheusPi: Ah now I get it. Please excuse my stupidity. @n01r thanks for pointing it out.

sbastrakov: `callReset` was a variable name defined just 1 line before it is called. With the new version I thought `resetParticles` would be a better name for it.

PrometheusPi: I agree, the naming is better now.

sbastrakov: Oh, did not see that @n01r already answered.

n01r:
I'm doing the runtime test with my existing example and when it succeeds, I'll merge :)

n01r:
Okay, even after restart no. 3 the results look good!   
Don't be confused that I have only half the box in propagation direction here.
I forgot to check out the PR #3035 while it was still open (to fix #3034 for in this run) and had to resort to adding the runtime arguments `-m --stopWindow <step>` which effectively reduced my *visible* simulation volume by 1 layer of GPUs. Nevertheless, the were simulated and checking between the different restarts seems to show no crazy behavior anymore.

PR #3035, opened 2019-08-23
Fix a crash in reporting deprecated parameters

### LINKED ISSUE #3034: LWFA Example: Boost Bad Anycast Abort
I am trying to run the LWFA example with latest `dev` 27a5ad4bf1e1eec3e234cd68610974a9205f6cee on Summit, see #3007, and running its `1.cfg` leads to a
```
Unhandled exception of type 'N5boost16exception_detail10clone_implINS0_19error_info_injectorINS_12bad_any_castEEEEE' with message 'boost::bad_any_cast: failed conversion using boost::any_cast', terminating
```

That's all I got in `stderr`.
`stdout` is
```
Running program...
lwfa_dev_017_01
present working directory:/gpfs/alpine/proj-shared/$proj/huebl/lwfa_dev_017_01
```

Does that example still run for you?

Please see used software in #3007 . Potentially could need a self-built boost instead of the module ...

sbastrakov:
Fixes  #3034.
The bug was introduced in #3022.

- [x] do a test run on Hemera (in the queue atm)

### DIFF
--- a/include/picongpu/ArgsParser.cpp
+++ b/include/picongpu/ArgsParser.cpp
@@ -54,7 +54,7 @@ namespace
         /* Moving window: a new run-time parameter 'windowMovePoint' to replace
          * compile-time 'movePoint' variable
          */
-        bool isMovingWindowEnabled = vm[ "moving" ].as< bool >( );
+        bool isMovingWindowEnabled = !vm[ "moving" ].empty();
         if( isMovingWindowEnabled )
         {
             bool isWindowMovePointSet = !vm[ "windowMovePoint" ].defaulted( );


sbastrakov:
Checked, ready for review.

include/picongpu/ArgsParser.cpp
```diff
@@ -54,7 +54,7 @@ namespace
         /* Moving window: a new run-time parameter 'windowMovePoint' to replace
          * compile-time 'movePoint' variable
          */
-        bool isMovingWindowEnabled = vm[ "moving" ].as< bool >( );
+        bool isMovingWindowEnabled = !vm[ "moving" ].empty();
```
n01r: Missing space between parentheses.
However, the rest of the file has a mixed style, anyway, so I would merge in order to get back to a working dev that we can use.

n01r: Wanna follow up with a PR that regexes all `()` to `( )`? :rofl: 
... Maybe rather not. Seems like overkill.

sbastrakov: The problem is that some files use the old formatting style consistently, also with `< >` and other things. What we need is automatic formatting tools that cover everything, not just brackets. Like [alpaka's effort](https://github.com/ComputationalRadiationPhysics/alpaka/pull/781)

ax3l: it really doesn't matter for zero-arguments.
The idea is to group for N>=1 arguments with visual grouping ;)

Issue #2998, opened 2019-07-12
PIConGPU on ARIS cluster at GRNET Greece 


StevE-Ong:
Hi, I have installed PIConGPU on ARIS cluster at GRNET, Greece, with the help from Axel in the conference just now. Thanks Axel. I attached the two files here:picongpu.profile and gpus.tpl.

ARIS system overview can be found at: http://doc.aris.grnet.gr

[ARIS.zip](https://github.com/ComputationalRadiationPhysics/picongpu/files/3386231/ARIS.zip)

sbastrakov:
@StevE-Ong thanks for your interest in PIConGPU and preparing the system-specific files!

Would you mind providing these new files and a short system overview as a pull request? There are similar pull requests for other systems, for example, #2874 and #2821. Please do not forget to add your name to the copyright line in the `gpus.tpl` file.

StevE-Ong:
Sure, I will make it later when I return from the conference.

psychocoderHPC:
@StevE-Ong If you like you can add your self via a pull request to the [PIConGPU Community Map](https://github.com/ComputationalRadiationPhysics/picongpu-communitymap) you can find an example [here](https://github.com/ComputationalRadiationPhysics/picongpu-communitymap/pull/8/files)

sbastrakov:
@StevE-Ong sorry to bother, just as a friendly reminder. Would you like to make a pull request with these machine-specific files? In case you are okay with these files being added to the repository but do not have time for making a pull request, I can make one with your files, then I just need your name to put in the copyright header.

ax3l:
Hi @StevE-Ong, thanks for opening the issue to document the usage. Feel free to open the PR and we will guide you during review. Great to read you again! :)

StevE-Ong:
> @StevE-Ong sorry to bother, just as a friendly reminder. Would you like to make a pull request with these machine-specific files? In case you are okay with these files being added to the repository but do not have time for making a pull request, I can make one with your files, then I just need your name to put in the copyright header.

@sbastrakov @ax3l  Hi, Sorry for late raply. I am not sure what is pull request. I am appreatiate if you can make one with the files. My name can be written as: Jian Fuh Ong

sbastrakov:
@StevE-Ong no problem, I will do it tomorrow.

sbastrakov:
@StevE-Ong sorry for a delay, I've just opened a PR #3039.

sbastrakov:
[...]

PR #3039, opened 2019-08-28
System: ARIS

sbastrakov:
The profile has been created by @StevE-Ong in #2998.

### DIFF
--- a/docs/source/install/profile.rst
+++ b/docs/source/install/profile.rst
@@ -276,3 +276,20 @@ Queue: gpus (4 x Nvidia V100 GPUs)
 
 .. literalinclude:: profiles/juwels-jsc/gpus_picongpu.profile.example
    :language: bash
+
+ARIS (GRNET)
+------------
+
+**System overview:** `link <http://doc.aris.grnet.gr/>`_
+
+**User guide:** `link <http://doc.aris.grnet.gr/environment/>`_
+
+**Production directory:** ``$WORKDIR`` (`link <http://doc.aris.grnet.gr/system/storage/>`_)
+
+For these profiles to work, you need to download the :ref:`PIConGPU source code <install-dependencies-picongpu>`.
+
+Queue: gpu (2 x NVIDIA Tesla k40m GPUs)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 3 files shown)


docs/source/install/profile.rst
```diff
@@ -276,3 +276,20 @@ Queue: gpus (4 x Nvidia V100 GPUs)
...
+ARIS (GRNET)
+------------
+
+**System overview:** `link <http://doc.aris.grnet.gr/>`_
+
+**User guide:** `link <http://doc.aris.grnet.gr/environment/>`_
+
+**Production directory:** ``$WORKDIR`` (`link <http://doc.aris.grnet.gr/system/storage/>`_)
+
+For these profiles to work, you need to download the :ref:`PIConGPU source code <install-dependencies-picongpu>` and install :ref:`PNGwriter, c-blosc, adios, libSplash, Boost and HDF5 <install-dependencies>` manually.
```
sbastrakov: @StevE-Ong please check if this software list is correct for the machine: e.g., it might already have `Boost` or other stuff available right away via modules.

sbastrakov: Okay. It is just that I wrote this paragraph and so have no idea.

StevE-Ong: Hi, sorry for late reply. In ARIS, there is a module called picongpu that contains part of the dependencies. PNGwritter and the slice for E,B didnt work and the system gave some error. Without them is working well. For the moment I am using hdf5 output and visualized by VisIt.

sbastrakov: @StevE-Ong so do you agree with @ax3l 's suggested phrasing (just above your reply) or should it be corrected somehow?

Also in case you want Github to track you as the author of this commit, please provide the email associated with your account (see [the question earlier](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3039#issuecomment-525816879)). I tried to guess your ELI email but looks like did it wrongly. In case you do not want to provide the email, that is fine too, then your full name is still there.

StevE-Ong: @sbastrakov I agree with @ax3l. Just downloading the source code should be all. My email: jianfuh.ong@eli-np.ro

sbastrakov: Thanks, now should be all fixed and ready.

Issue #3037, opened 2019-08-26
Error with hdf5 output on current dev branch


finnolec:
I pulled the current dev branch 0f800d0 and created an example based on the Kelvin Helmholtz example. If I have hdf5 output enabled (the default), the simulation instantly crashes and gives me following error messages:
<details>
<summary> stdout </summary>

```
Running program...
PIConGPU: 0.5.0-dev
  Build-Type: Release

Third party:
  OS:         Linux-3.10.0-693.11.6.el7.x86_64
  arch:       x86_64
  CXX:        GNU (7.3.0)
  CMake:      3.11.3
  CUDA:       9.2.88
  mallocMC:   2.3.1
  Boost:      1.68.0
  MPI:        
    standard: 3.1
    flavor:   OpenMPI (2.1.6)
  PNGwriter:  0.7.0
  libSplash:  1.7.0 (Format 4.0)
  ADIOS:      1.13.1
PIConGPUVerbose PHYSICS(1) | Sliding Window is OFF
PIConGPUVerbose PHYSICS(1) | used Random Number Generator: RNGProvider3XorMin seed: 42
PIConGPUVerbose PHYSICS(1) | Courant c*dt <= 1.00556 ? 1
PIConGPUVerbose PHYSICS(1) | species e: omega_p * dt <= 0.1 ? 0.0319333
PIConGPUVerbose PHYSICS(1) | species i: omega_p * dt <= 0.1 ? 0.000745229
PIConGPUVerbose PHYSICS(1) | macro particles per device: 3686400
PIConGPUVerbose PHYSICS(1) | typical macro particle weighting: 326.577
PIConGPUVerbose PHYSICS(1) | UNIT_SPEED 2.99792e+08
PIConGPUVerbose PHYSICS(1) | UNIT_TIME 1.79e-16
PIConGPUVerbose PHYSICS(1) | UNIT_LENGTH 5.36628e-08
PIConGPUVerbose PHYSICS(1) | UNIT_MASS 2.97492e-28
PIConGPUVerbose PHYSICS(1) | UNIT_CHARGE 5.23234e-17
PIConGPUVerbose PHYSICS(1) | UNIT_EFIELD 9.5224e+12
PIConGPUVerbose PHYSICS(1) | UNIT_BFIELD 31763.3
PIConGPUVerbose PHYSICS(1) | UNIT_ENERGY 2.67372e-11
initialization time: 10sec 721msec = 10 sec
[...]

PR #3040, opened 2019-08-28
Remove unused includes of pthread

sbastrakov:
This has been spotted during the investigation of #3037 (but is unrelated to that issue).

Now the core code of PIConGPU does not include or use `pthread`, only `cuda_memtest` and `alpaka` do.


ax3l:
Thx, that's outdated.
In the past we used a thread for I/O, but actually only MPI implementations today are threadsafe to do so ;)

PR #3042, opened 2019-08-29
Add YeePML to comments for field solver selection

sbastrakov:
This was forgotten when YeePML was added.


sbastrakov:
@PrometheusPi you are right that it is not necessary here. My idea was to mentally connect a user to `FieldTmp` slots and thus the word temporary.

PR #3043, opened 2019-09-01
Docs: Cori's KNL Nodes (NERSC)

ax3l:
Add profiles for Cori's KNL nodes.

As on Taurus' KNL nodes at ZIH, I kept the "one MPI rank per KNL" and "one hyperthread per code" strategy.

I am not sure if we ever tuned this for ideal performance, because e.g. [WarpX](https://warpx.readthedocs.io/en/latest/running_cpp/platforms.html)/PICSAR reported better performance with 8 MPI ranks per device to sub-partition the KNL. If I remember correctly, we just checked some hyperthread performance [here](https://arxiv.org/abs/1706.10086) on KNL and sub-partition all other, multi-socket CPU hardware these days.

Either way, as the Taurus KNL profile, this profile works.

I guess @sbastrakov would have some experience tuning the checked-in KNL profiles further (e.g. on Taurus). Maybe we want to revisit partitioning a KNL into multiple "devices" like we do for multi-socket systems. And maybe we want to use hyperthreading in those profiles with "close" process binding.

### DIFF
--- a/docs/source/install/profile.rst
+++ b/docs/source/install/profile.rst
@@ -154,8 +154,8 @@ Queue: gpu2 (Nvidia K80 GPUs)
 .. literalinclude:: profiles/taurus-tud/k80_picongpu.profile.example
    :language: bash
 
-Queue: knl (Intel  Intel Xeon Phi - Knights Landing)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Queue: knl (Intel Xeon Phi - Knights Landing)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 For this profile, you additionally need to install your own :ref:`boost <install-dependencies>`.
 
@@ -205,8 +205,8 @@ Cori (NERSC)
 
 For these profiles to work, you need to download the :ref:`PIConGPU source code <install-dependencies-picongpu>` and install :ref:`PNGwriter and libSplash <install-dependencies>` manually.
 
-Queue: regular (Intel  Intel Xeon Phi - Knights Landing)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Queue: regular (Intel Xeon Phi - Knights Landing)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. literalinclude:: profiles/cori-nersc/knl_picongpu.profile.example
    :language: bash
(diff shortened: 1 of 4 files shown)


sbastrakov:
@ax3l in my experience with Picador, 1 process per KNL and 1 thread per core is never best. Back then our best configuration was 4 or 8 ranks per KNL with 2 to 4 threads per core, depending on a setup. However, notably having 1/1 was relatively not nearly as bad as for KNC.

docs/source/install/profile.rst
```diff
@@ -194,6 +194,23 @@ Additionally, you need to make the ``rsync`` command available as written below.
...
+
+**System overview:** `link <https://www.nersc.gov/users/computational-systems/cori/configuration/>`_
+
+**User guide:** `link <https://docs.nersc.gov/>`_
+
+**Production directory:** ``$SCRATCH`` (`link <https://www.nersc.gov/users/storage-and-file-systems/>`_).
+
+For these profiles to work, you need to download the :ref:`PIConGPU source code <install-dependencies-picongpu>` and install :ref:`PNGwriter and libSplash <install-dependencies>` manually.
+
+Queue: regular (Intel  Intel Xeon Phi - Knights Landing)
```
sbastrakov: Was `Intel` accidentally put twice?
Also, to be consistent with other descriptions, maybe add processor version `7250`?

docs/source/install/profile.rst
```diff
@@ -194,6 +194,23 @@ Additionally, you need to make the ``rsync`` command available as written below.
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
sbastrakov: Guess that was my misunderstanding: should the `.tpl` and `_profile.example` file names be the same as the queue name? Here it is `regular` vs. `knl` (although `knl` makes more sense definitely)

docs/source/install/profile.rst
ax3l: copy paste error, fixed in both places now.

docs/source/install/profile.rst
ax3l: Yes, in the past these were identical. Somehow we have clusters these days (Taurus, Cori) that use the same queue for various hardware and specify the hardware via constrains (e.g. Slurm `-C`).

I think using the device hardware for the `.tpl` make sense, but we might want to change the sub-titles in this profile at some point. But it's still clear and fully described here, so I think the current way works.

etc/picongpu/cori-nersc/knl_picongpu.profile.example
```diff
@@ -0,0 +1,67 @@
...
+export PYTHONPATH=$PICSRC/lib/python:$PYTHONPATH
+
+# "tbg" default options #######################################################
+#   - SLURM (sbatch)
+#   - "defq" queue
+export TBG_SUBMIT="sbatch"
+export TBG_TPLFILE="etc/picongpu/cori-nersc/knl.tpl"
+
+# allocate an interactive shell for one hour
+#   getNode 2  # allocates to interactive nodes (default: 1)
```
PrometheusPi: typo: `to` --> `two`

PrometheusPi: This is a copy and past error nobody discovered before. Opened #3046 to fix it.

PR #3049, opened 2019-09-04
Refactor internals of pmacc::traits::GetUniqueTypeId implementation

sbastrakov:
Separate the global counter and make an explicit interface for accessing it, since we do it from `FieldTmp` anyways (also document there that we probably should not).
These changes also avoid a potential issue with (non-template) static member definition in a header file.


include/pmacc/traits/GetUniqueTypeId.hpp
```diff
@@ -33,41 +33,58 @@ namespace pmacc
...
+
+/** Unique id for a given type
  *
- * This class is based on
- *  - http://stackoverflow.com/a/7562583
- *  - author: MSN (edited version from Sep 27th 2011 at 3:59)
- *  - license: CC-BY-SA 3.0 (http://creativecommons.org/licenses/by-sa/3.0/)
+ * @tparam T_Type type
  */
 template<typename T_Type>
```
sbastrakov: Note: I have studied that whole thread on SO and do not think even the original implementation was based on that one, so removed the reference.

PrometheusPi: I agree - the referenced code follows a different approach

include/pmacc/traits/GetUniqueTypeId.hpp
```diff
@@ -33,41 +33,57 @@ namespace pmacc
...
+ */
 template<typename T_Type>
-uint64_t GetUniqueTypeId<T_Type>::counter = GetUniqueTypeId<T_Type>::id;
+const uint64_t TypeId<T_Type>::id = getNextId( );
 
 } //namespace detail
 
+/** Get next available type id
+ *
+ * Warning: is not thread-safe.
```
PrometheusPi: Is this an issue?

sbastrakov: It is a limitation. As of now, we do not use it in a concurrent environment, nor was the original code (in `FieldTmp`) thread-safe. So just to document the limitation.

PrometheusPi: Thanks for clarifying this.

sbastrakov:
@PrometheusPi yes, ready to be merged.

sbastrakov:
Since this is not urgent, we will wait a couple of days for @psychocoderHPC to also have a look.

include/picongpu/fields/FieldTmp.tpp
```diff
@@ -57,11 +57,11 @@ namespace picongpu
         SimulationFieldHelper<MappingDesc>( cellDescription ),
         m_slotId( slotId )
     {
-        m_commTagScatter =
-            ++pmacc::traits::detail::GetUniqueTypeId< uint8_t >::counter +
-            SPECIES_FIRSTTAG;
-        m_commTagGather = ++pmacc::traits::detail::GetUniqueTypeId< uint8_t >::counter +
-            SPECIES_FIRSTTAG;
+        /** \todo this is a work-around to get a unique tag,
+         *  potentially unsafe in case there are too many type ids
```
ax3l: maybe be more specific what `too many type ids` means? :)
Larger than `SPECIES_FIRSTTAG`?

sbastrakov: Made me think of that and realize it is actually not a problem. Updated the comment.

PR #3050, opened 2019-09-04
[WIP] Change PMACC_C_STRING to produce a static constexpr member

sbastrakov:
The original implementation used a non-`constexpr` member and an external definition. This was problematic since the definition had to be placed on a header file, potentially causing multiple definitions. The new way is also consistent with how other members are handled.

Since this might break something for some compilers, WIP for now until this is properly checked. The only "use" in our codebase now seems to be in the `FromHDF5` particle density, but this is only in our Empty example.

### DIFF
--- a/include/pmacc/preprocessor/struct.hpp
+++ b/include/pmacc/preprocessor/struct.hpp
@@ -238,13 +238,9 @@
 #define PMACC_PP_X_INIT_VALUE_VARIABLE_WITH_PAREN(elem)                        \
     PMACC_PP_SELECT_TYPEID( 5,PMACC_PP_X1_INIT_VALUE_VARIABLE,elem )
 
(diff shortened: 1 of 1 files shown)


sbastrakov:
Okay, was tested and also turned out we already use this thing in another place (which I did not notice). Ready for review.

sbastrakov:
I guess, you mean the `PMACC_STRUCT` and all the macro ~~hell~~ magic inside. Maybe we actually do not.

ax3l:
I mean: do we need `PMACC_PP_X_CREATE_C_STRING_VARIABLE` at all anymore? (Yes, the less of the other stuff, the better too :) )

sbastrakov:
Ah, with this PR the strings are treated the same with other variables. Hence I assumed you mean to decide whether to keep or remove everything.

sbastrakov:
Okay, I finally got it the third time. Will have a look.

PR #3051, opened 2019-09-04
Add missing inline specifiers for functions defined in header files

sbastrakov:
Most of this `inline` stuff was covered by #3006, but a couple of things were missed.

For the context: with all these PRs aimed at avoiding double definitions, it is now almost possible to have multiple object files in picongpu that include pmacc and all the real stuff. With a couple more changes, not submitted yet, LWFA is building on both my MSVS and on Hemera in that "multiple object files" mode.

### DIFF
--- a/include/picongpu/plugins/radiation/check_consistency.hpp
+++ b/include/picongpu/plugins/radiation/check_consistency.hpp
@@ -30,15 +30,15 @@ namespace plugins
 namespace radiation
 {
 
-void check_consistency(void)
+HINLINE void check_consistency(void)
 {
-  using namespace parameters;
-  std::cout << " checking efficiency of radiation code: " ;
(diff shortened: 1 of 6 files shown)


include/picongpu/plugins/radiation/check_consistency.hpp
```diff
@@ -30,15 +30,15 @@ namespace plugins
 namespace radiation
 {
 
-void check_consistency(void)
+inline void check_consistency(void)
```
PrometheusPi: Why did you not use `HINLINE` here?

sbastrakov: Ah, yes, I was again confused of your rules on that. Will change to `HINLINE`.

Issue #800, opened 2015-04-08
radiation window function: replace namespaces with functors


PrometheusPi:
Currently, the radiation window functions are selected via namespaces. This should be replaced by functors as suggested by @psychocoderHPC.

PrometheusPi:
@psychocoderHPC Could you tell me again, why using functors/classes is better than namespaces?

ax3l:
Because `namespaces` cannot be derived from and only form a scope of definitions, while a `class` groups one aspect together for reusing in a different context. We initially used namespaces and free constant variables in input but found many features, e.g. in dependent variables and normalizations are much easier to express when all options are in classes.

psychocoderHPC:
@PrometheusPi @steindev is very unhappy that we use namespaces here!

PrometheusPi:
Thanks for the reminder – if I forget, please reminder me in 8 years again 😜

PR #3052, opened 2019-09-09
Add output and python module doc for radiation plugin

PrometheusPi:
This pull request closes #3041. It adds a description on how to read the outputted data from the radiation plugin both for text-based output and the HDF5 output. Furthermore it provides python code to read and plot the data. The documentation now also covers how to use the python module to read the HDF5 radiation data. 

This pull requests also fixes an older typo and a table misalignment. 

The separate commits add different content and thus can be reviewed separately. 

@Anton-Le please have a look at this pull request.

PR #3053, opened 2019-09-10
refactor: omega read from file

PrometheusPi:
In the current implementation the radiation plugin provides a method to set a path tp a list of frequency (omega) values to compute the radiation for. However, the path is set only via command line option and the value set via namespace setting is totally ignored. 

Since various parameters need to be set in `radiation.param` anyway, I removed the command line option (which is not doing anything if the list option has not be selected in `radiation.param` before) end set the path to the path defined in the namespace. 

The documentation was adjusted accordingly.

I would like to do a run-time test just to see whether anything else is not working properly with this frequency option before continuing to work on #800 (for the frequencies). 

- [x] run time test  

----------
**Update:**

With the latest version, all run time tests produced correct results. Both linear and logarithmic frequency scales work as before. The list now works again and has been tested with newlines, tabs and spaces as separators (in agreement with the newly updated documentation).

### DIFF
--- a/docs/source/usage/plugins/radiation.rst
+++ b/docs/source/usage/plugins/radiation.rst
@@ -75,7 +75,7 @@ The number of total sample frequencies ``N_omega`` need to be defined as ``const
 In the sub-namespace ``SI``, a minimal frequency ``omega_min`` and a maximum frequency ``omega_max`` need to be defined as ``constexpr float_64``.
 
 For the **file-based frequency** definition,  all definitions need to be in the ``picongpu::plugins::radiation::frequencies_from_list`` namespace.
-The number of total frequencies ``N_omega`` need to be defined as ``constexpr unsigned int``  and the path to the file containing the frequency values un units of :math:`[s^{-1}]` needs to be given as ``constexpr char listLocation[] = "/path/to/frequency_list";``. 
+The number of total frequencies ``N_omega`` need to be defined as ``constexpr unsigned int``  and the path to the file containing the frequency values un units of :math:`[s^{-1}]` needs to be given as ``constexpr const char * listLocation = "/path/to/frequency_list";``.
 
 .. note::
 
--- a/docs/source/usage/plugins/radiation.rst
+++ b/docs/source/usage/plugins/radiation.rst
@@ -62,6 +62,24 @@ namespace                     Description
 ============================= ==============================================================================================
 
 
+
+All three options require variable definitions in the according namespaces as described below:
+
+For the **linear frequency** scale all definitions need to be in the ``picongpu::plugins::radiation::linear_frequencies`` namespace. 
+The number of total sample frequencies ``N_omega`` need to be defined as ``constexpr unsigned int``.
+In the sub-namespace ``SI``, a minimal frequency ``omega_min`` and a maximum frequency ``omega_max`` need to be defined as ``constexpr float_64``.
+
+For the **logarithmic frequency** scale all definitions need to be in the ``picongpu::plugins::radiation::log_frequencies`` namespace. 
+Equivalently to the linear case, three variables need to be defined: 
+The number of total sample frequencies ``N_omega`` need to be defined as ``constexpr unsigned int``.
+In the sub-namespace ``SI``, a minimal frequency ``omega_min`` and a maximum frequency ``omega_max`` need to be defined as ``constexpr float_64``.
+
+For the **file-based frequency** definition,  all definitions need to be in the ``picongpu::plugins::radiation::frequencies_from_list`` namespace.
+The number of total frequencies ``N_omega`` need to be defined as ``constexpr unsigned int``  and the path to the file containing the frequency values un units of :math:`[s^{-1}]` needs to be given as ``constexpr char listLocation[] = "/path/to/frequency_list";``. 
+
+
+
+
 Observation directions
 """"""""""""""""""""""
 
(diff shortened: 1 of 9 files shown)


ax3l:
I did not find more than a mention of `listLocation` in the `.rst`, did I overlook it? I though about adding docs on what's *inside* it (it's format) and where it ideally should reside in our inputs.

PrometheusPi:
@ax3l Okay, I was referring to the single mention of `listLocation` [here](https://github.com/PrometheusPi/picongpu/blame/dc5954dec33e5a4a3921a7ff9b08e446cd53af29/docs/source/usage/plugins/radiation.rst#L61) which currently does not show up in the html version (fixed with #3052).

Since I am currently refactoring the entire frequency definition (with adding new docs) I think adding a (short-lived) longer documentation is not needed. But if you prefer a documentation, I will add another commit.

PrometheusPi:
@ax3l #3052 does not yet add the planed changes. It just fixes a table formatting issue that prevented showing `listLocation` at all (in the html version).

PrometheusPi:
I will add a commit to describe the currently needed setup.

share/picongpu/examples/Bunch/include/picongpu/param/radiation.param
```diff
@@ -65,7 +65,8 @@ constexpr unsigned int N_omega = 2048; // number of frequencies
 
 namespace frequencies_from_list
 {
-//const char listLocation[] = "/scratch/s5960712/list001.freq";
+/** path to text file with frequencies */
+const char listLocation[] = "/path/to/frequency.list";
```
psychocoderHPC: Muste be `constexpr char listLocation[]` to fit the documentation above. If this make compile issues change it to `constexpr char const listLocation* = "path";`

share/picongpu/examples/KelvinHelmholtz/include/picongpu/param/radiation.param
```diff
@@ -68,7 +68,8 @@ constexpr unsigned int N_omega = 1024; // number of frequencies
 
 namespace frequencies_from_list
 {
-//const char listLocation[] = "/scratch/s5960712/list001.freq";
+/** path to text file with frequencies */
+const char listLocation[] = "/path/to/frequency.list";
```
psychocoderHPC: same here please sue `constexpr`

share/picongpu/examples/Bunch/include/picongpu/param/radiation.param
sbastrakov: I think actually `constexpr const char *` is preferable over `constexpr char listLocation[]` due to more operations allowed with such a variable w/o requiring a definition.

PrometheusPi:
Currently, using the omega by list system fails with a segfault. I will thus put this pull request as **WIP**. This seems to be not caused by the frequency reading - the vernose autput confirms that data has been read correctly. I suspect an outdated use of the data box. @psychocoderHPC do you see anything obviously wrong. 
In the next test I will add debug options and post more results of the failing here.

psychocoderHPC:
> Currently, using the omega by list system fails with a segfault. I will thus put this pull request as **WIP**. This seems to be not caused by the frequency reading - the verbose output confirms that data has been read correctly. I suspect an outdated use of the data box. @psychocoderHPC do you see anything obviously wrong.
> In the next test I will add debug options and post more results of the failing here.

We found out @PrometheusPi accesses device memory from the host side and therefore get the segfault

PrometheusPi:
@psychocoderHPC all fixes are included now - can be reviewed

docs/source/usage/plugins/radiation.rst
```diff
@@ -62,6 +62,28 @@ namespace                     Description
...
+The number of total sample frequencies ``N_omega`` need to be defined as ``constexpr unsigned int``.
+In the sub-namespace ``SI``, a minimal frequency ``omega_min`` and a maximum frequency ``omega_max`` need to be defined as ``constexpr float_64``.
+
+For the **logarithmic frequency** scale all definitions need to be in the ``picongpu::plugins::radiation::log_frequencies`` namespace. 
+Equivalently to the linear case, three variables need to be defined: 
+The number of total sample frequencies ``N_omega`` need to be defined as ``constexpr unsigned int``.
+In the sub-namespace ``SI``, a minimal frequency ``omega_min`` and a maximum frequency ``omega_max`` need to be defined as ``constexpr float_64``.
+
+For the **file-based frequency** definition,  all definitions need to be in the ``picongpu::plugins::radiation::frequencies_from_list`` namespace.
+The number of total frequencies ``N_omega`` need to be defined as ``constexpr unsigned int``  and the path to the file containing the frequency values un units of :math:`[s^{-1}]` needs to be given as ``constexpr const char * listLocation = "/path/to/frequency_list";``.
```
psychocoderHPC: `... values un units ...` should be `...  and units ...`

docs/source/usage/plugins/radiation.rst
```diff
@@ -62,6 +62,28 @@ namespace                     Description
...
+The frequency values in the file can be separated by newlines, spaces, tabs, or any other whitespace. The numbers should be given in such a way, that c++ standard ``std::ifstream`` can interpret the number e.g., as ``2.5344e+16``. 
+
+.. note::
+
+   Currently, the variable ``listLocation`` is required to be defined in the ``picongpu::plugins::radiation::frequencies_from_list`` namespace.
```
psychocoderHPC: Please extent with `... even if frequencies_from_list is not used`. This is in the current documentation not fully clear.

docs/source/usage/plugins/radiation.rst
PrometheusPi: great catch 👍 should be `in units`

PrometheusPi:
@psychocoderHPC The plan was not to convert to `struct` right away but to first of all fix the bug in frequencies from list in a separate pull request - then work on the refactoring of the interface. Thus I did not use `struct`s right away, but will add them in a followup pull request.

PrometheusPi:
@psychocoderHPC I added your suggestions (except `struct`- see comment before on that)

psychocoderHPC:
Ok I am fine with fixing the bug and adding documentation only.

docs/source/usage/plugins/radiation.rst
psychocoderHPC: saw that @ax3l  allready mentioned it.

PR #3059, opened 2019-09-13
update profile hemera-hzdr: CMake version

psychocoderHPC:
switch from for hemera from cmake/3.11.3 to cmake/3.15.2

Edit: the new version is required by the future (upcoming) alpaka release.


ax3l:
LGTM, do not forget to update
- `INSTALL.rst`
- `CMakeLists.txt`
- [spack-repo](https://github.com/ComputationalRadiationPhysics/spack-repo)

as soon as you update with the new Alpaka release, so the documented CMake version is bumped.

PR #3061, opened 2019-09-17
SpeciesEligibleForSolver for radiation plugin

finnolec:
I added the SpeciesEligibleForSolver for the @PrometheusPi's radiation plugin, so that the `depenciesFulfilled` variable in  `radiation.hpp`, as well as the empty Kernel in `radiation.kernel` are not required anymore.

### DIFF
--- a/include/picongpu/plugins/radiation/Radiation.hpp
+++ b/include/picongpu/plugins/radiation/Radiation.hpp
@@ -28,7 +28,7 @@
 
 #include "picongpu/traits/SplashToPIC.hpp"
 #include "picongpu/traits/PICToSplash.hpp"
-
+#include "picongpu/particles/traits/SpeciesEligibleForSolver.hpp"
 
 #include "picongpu/plugins/radiation/Radiation.kernel"
 #include "picongpu/plugins/radiation/ExecuteParticleFilter.hpp"
@@ -120,16 +120,6 @@ private:
     std::string folderRadPerGPU;
     DataSpace<simDim> lastGPUpos;
 
-    /** defines if all kernel dependencies are full filled
-     *
-     * dependencies:
-     *   - species contains the attribute `momentumPrev1`
-     */
-    static constexpr bool dependenciesFulfilled = pmacc::traits::HasIdentifier<
-        typename ParticlesType::FrameType,
-        momentumPrev1
-    >::type::value;
-
     /**
      * Data structure for storage and summation of the intermediate values of
      * the calculated Amplitude from every host for every direction and
@@ -193,7 +183,7 @@ public:
      */
     void notify(uint32_t currentStep)
     {
-        if (dependenciesFulfilled && currentStep >= radStart)
+        if (currentStep >= radStart)
         {
             // radEnd = 0 is default, calculates radiation until simulation
             // end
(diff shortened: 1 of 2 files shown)


PrometheusPi:
@finnolec :+1: great job

PrometheusPi:
@finnolec You just encountered a merge conflict since @psychocoderHPC merged #3053 a second before your commit - just rebase and resubmit.

PrometheusPi:
@finnolec I am surprised that you last commit shows code that has been added with #3053 - Did you rebase against the dev or just added the changes?

finnolec:
Yes, I didn't rebase for this commit. I read the message about rebasing a bit too late...

sbastrakov:
@finnolec no problem, btw github also shows you if the changes can be auto-merged (so there is no conflict) when you create a PR.

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -302,81 +280,92 @@ class Radiation : public ISimulationPlugin
      * is created.       */
     void pluginLoad()
     {
-        if(dependenciesFulfilled)
+        if(!notifyPeriod.empty())
         {
             // allocate memory for all amplitudes for temporal data collection
             tmp_result = new Amplitude[elements_amplitude()];
```
sbastrakov: It probably does not matter, but here is a slight change of logic: in the old code this allocation was only done when the species have the right attribute, while now only when the plugin is actually enabled.

PrometheusPi: Okay - that is true - should probably check if **both** is fulfilled.

PrometheusPi: That is actually better since the memory is only needed if the plugin is actually called.

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -1250,6 +1238,53 @@ class Radiation : public ISimulationPlugin
...
+    template<
+            typename T_Species,
+            typename T_UnspecifiedSpecies
+    >
+    struct SpeciesEligibleForSolver<
+            T_Species,
+            plugins::radiation::Radiation< T_UnspecifiedSpecies >
+    >
+    {
+    using FrameType = typename T_Species::FrameType;
```
sbastrakov: The insides of this `struct` should be indented.

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -1250,6 +1238,53 @@ class Radiation : public ISimulationPlugin
...
+
+    // this plugin needs the position, weighting, momentum and momentumPrev1 to run
+    using RequiredIdentifiers = MakeSeq_t<
+            position<>,
```
sbastrakov: It looks like the template arguments are on the right level of indentation, but `using RequiredIdentifiers` is not, same for other types below.

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -1250,6 +1238,53 @@ class Radiation : public ISimulationPlugin
...
+            chargeRatio<>
+    >::type;
+
+    using type = typename bmpl::and_<
+            SpeciesHasIdentifiers,
+            SpeciesHasMass,
+            SpeciesHasCharge
+    >;
+    };
+} // namespace traits
```
sbastrakov: Please add an extra line here before the `} // namespace traits`.

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -1250,6 +1238,53 @@ class Radiation : public ISimulationPlugin
...
+
+    using SpeciesHasCharge = typename pmacc::traits::HasFlag<
+            FrameType,
+            chargeRatio<>
+    >::type;
+
+    using type = typename bmpl::and_<
+            SpeciesHasIdentifiers,
+            SpeciesHasMass,
+            SpeciesHasCharge
```
sbastrakov: The requirements for mass and charge were not there in the original version, however I assume those are actually required.

finnolec: I just included them, because it's physically required and they are used in the kernel. I don't think it hurts to require these identifiers. What do you think @PrometheusPi ?

sbastrakov: Sure, if they are actually used from the kernel, then it's a good fix.

PrometheusPi: I think that this is a great improvement.

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -1250,6 +1222,60 @@ class Radiation : public ISimulationPlugin
...
+            momentumPrev1
+        >;
+
+        using SpeciesHasIdentifiers = typename pmacc::traits::HasIdentifiers<
+            FrameType,
+            RequiredIdentifiers
+        >::type;
+
+        // this plugin needs the radiationMask flag
+        using SpeciesHasMask = typename pmacc::traits::HasFlag<
```
PrometheusPi: I am confused by this definition: 
The radiation mask is not necessarily required. It is an optional (old) workaround to species filters (which, due to a missing history, would cause numeric artifacts with the radiation algorithm). You added the definition `SpeciesHasMask`  here but did not include it in `line 1270` - which is fine, since it is not required. 
But why did you keep the definition here?

finnolec: I just didn't see it in one of the merge conflicts I got from rebasing. It should be fixed now.

PrometheusPi:
@sbastrakov Do you have any further comments or requests? If not - we should merge this pull request.

sbastrakov:
@PrometheusPi @finnolec I do not have objections to the code, good work!
Just some commit messages are maybe not very descriptive, particularly the two concerning rebasing.

PrometheusPi:
@sbastrakov I agree, some squashing would improve the pr.

finnolec:
@sbastrakov @PrometheusPi I squashed everything :)

PR #3065, opened 2019-09-18
fix examples particle filters

PrometheusPi:
This pull request fixes a missing name definition in the particle filter examples. (first commit) 
As you can see in the doxygen generated documentation [here](https://picongpu.readthedocs.io/en/latest/usage/param/particles/init.html#manipulation-filters), the names have been missing before. (this also effects previous documentations. @ax3l Do we need to backport this?) 

The second commit fixes a comparison ion the free filter that compares a 3D vector with a scalar. This should probably not work. Do you agree @psychocoderHPC, @sbastrakov ?

### DIFF
--- a/include/picongpu/particles/filter/generic/Free.def
+++ b/include/picongpu/particles/filter/generic/Free.def
@@ -48,6 +48,7 @@ namespace generic
      *               result = true;
      *           return result;
      *       }
+     *       static constexpr char const * name = "eachParticleAboveMiddleOfTheCell";
      *   };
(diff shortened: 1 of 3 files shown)


sbastrakov:
Great!
About vector comparison: it does not work as of now (and rightfully so, imo), so your addition of `.y()` is definitely on point.

ax3l:
Marked it in the according project but I guess we will not do a 0.4.4 but a 0.5.0 instead.

Issue #1972, opened 2017-04-18
Bash Completion for Most Important Scripts


ax3l:
Follow up to #1971 

Write bash_completion scripts for all workflow-relevant tools:

- [x] `pic-create` #3069
- [x] `pic-configure` #3069
- [x] `pic-compile` #3069
- [x] `pic-build` #3069
- [x] `pic-edit` #3069
- [x] `tbg` #3069

Be aware there do exist tools that can semi or fully-automate completion script generation, e.g. for bash, csh and others. Ideally, we would auto-parse the output of `--help` for that.

Add the completions to:

`share/picongpu/picongpu-completion.bash` and source it in `picongpu.profile` files for *interactive* shells only.

### References

- [Debian tutorial](https://debian-administration.org/article/316/An_introduction_to_bash_completion_part_1)
- [compleat](https://github.com/mbrubeck/compleat): *Generate command-line completions using a simple DSL*

finnolec:
I'm looking into it :)

steindev:
Can be closed?

PrometheusPi:
@steindev great catch - I will check it ...

ax3l:
Awesome, all implemented via #3069. Thanks again, @finnolec! :rocket: :sparkles:

PR #3069, opened 2019-09-24
Bashcompletion

finnolec:
I finished a basic completion for the important PIConGPU scripts as mentioned in #1972. If you want to test it, you can `source $PICSRC/bin/picongpu-completion.bash` after sourcing a PIConGPU profile (Or just source a new profile which does it automatically). 

This is a list of features:
- [x] Autocomplete options of picongpu commands if you start with a dash, e.g. `pic-create -[TAB]` suggests you `-f`, `--force`, `-h` and `--help` and `pic-create --f[TAB]` completes to `pic-create --force`. I kept all the duplicates like `-f` and `--force`, which do the same thing because some people might want to see the short and long versions
- [x] `tbg -c [TAB]` shows all `*.cfg` files in `etc/picongpu/` in the current parameter set and autocompletes correspondingly
- [x] `pic-edit` autocompletes to all possible `.param`-files in `$PICSRC/include/picongpu/` without the trailing `.param`

The following points have been considered, but decided for future implementation, not  in this PR:

 - Show correct CMAKE variables if the previous option was `--cmake`
- Only show `.tpl` files at `tbg -t [TAB]`
- Show available backends at `--backend`. Here I'm not sure if it is generally possible or too system dependend
- Show available submit systems at `tbg -s [TAB]`. The same with the previous point

It should work so far. Feel free to test it and leave suggestions @ComputationalRadiationPhysics/picongpu-developers :)

I probably want to add the unfinished features in a separate pull request.

### DIFF
--- a/bin/pic-create
+++ b/bin/pic-create
@@ -36,6 +36,8 @@ help()
     echo ""
     echo "usage: pic-create [OPTION] [src_dir] dest_dir"
     echo "If no src_dir is set picongpu a default case is cloned"
+    echo "If src_dir is not in the currrent directory, pic-create will"
+    echo 'look for it in $PIC_EXAMPLES'
     echo ""
     echo "-f | --force         - merge data if destination already exists"
     echo "-h | --help          - show this help message"
@@ -79,6 +81,10 @@ if [ $# -eq 2 ] ; then
     src_path=$1
     dst_path=$2
     selected_default_folder_to_copy=$folder_to_clone
+    # Use example directory if no source directory available at current directory
+    if [[ -d $PIC_EXAMPLES/$src_path && ! -d ./$src_path ]]; then
+        src_path=$PIC_EXAMPLES/$src_path
+    fi
 else
     src_path=$this_dir/..
     dst_path=$1
@@ -121,6 +127,8 @@ else
         mkdir -p "$dst_path"
     fi
 fi
+
+echo "Cloning from $src_path to $dst_path"
 
 #first copy default parameter
 for d in $default_folder_to_copy
--- a/bin/picongpu-completion.bash
+++ b/bin/picongpu-completion.bash
@@ -1,0 +1,253 @@
+#/usr/bin/env bash
+
+show_directory()
+{
+    # Autocompletion from the bash "cd" command
+    # Copied from "bash-completion(1:2.8-1ubuntu1)" in /usr/share/bash-completion/bash_completion
+    local cur prev words cword
+    _init_completion || return
+
+    local IFS=$'\n' i j k
+
+    compopt -o filenames
+
+    # Show example directories with autocompletion for pic-create
+    if [[ $1 == "examples" ]]; then
+        for i in ${PIC_EXAMPLES//:/$'\n'}; do
+            k="${#COMPREPLY[@]}"
+            for j in $( compgen -d -- $i/$cur ); do
+                if [[ ( ! -d ${j#$i/} ) ]]; then
+                  j+="/"
+                fi
+                COMPREPLY[k++]=${j#$i/}
+            done
+        done
+    fi
+
+    # Use standard dir completion if no CDPATH or parameter starts with /,
+    # ./ or ../
+    if [[ -z "${CDPATH:-}" || "$cur" == ?(.)?(.)/* ]]; then
+        _filedir -d
+        return
+    fi
+
+    local -r mark_dirs=$(_rl_enabled mark-directories && echo y)
+    local -r mark_symdirs=$(_rl_enabled mark-symlinked-directories && echo y)
+
+    # we have a CDPATH, so loop on its contents
+    for i in ${CDPATH//:/$'\n'}; do
+        # create an array of matched subdirs
+        k="${#COMPREPLY[@]}"
+        for j in $( compgen -d -- $i/$cur ); do
+            if [[ ( $mark_symdirs && -h $j || $mark_dirs && ! -h $j ) && ! -d ${j#$i/} ]]; then
+                j+="/"
+            fi
+            COMPREPLY[k++]=${j#$i/}
+        done
+    done
+
+    _filedir -d
+
+    # if only one possible result then autocomplete to this result and enter the directory
+    if [[ ${#COMPREPLY[@]} -eq 1 ]]; then
+        i=${COMPREPLY[0]}
+        if [[ "$i" == "$cur" && $i != "*/" ]]; then
+            COMPREPLY[0]="${i}/"
+        fi
+    fi
+
+    return 0
+}
+
+show_parameters()
+{
+    local opts=$1
+    COMPREPLY=( $(compgen -W "${opts}" -- ${cur}))
+    return 0
+}
+
+_pic-build()
+{
+    # Load current string
+    local cur prev
+    cur="${COMP_WORDS[COMP_CWORD]}"
+    prev="${COMP_WORDS[COMP_CWORD-1]}"
+
+    if [[ "${prev}" == "-b" || "${prev}" == "--backend" ]]; then
+        # available backends from documentation
+        show_parameters "cuda omp2b serial tbb threads"
+        return 0
+    fi
+
+    if [[ "${prev}" == "-c" || "${prev}" == "--cmake" ]]; then
+        show_parameters "-b --backend -c --cmake -t -f --force -h --help"
+        return 0 # @TODO show available cmake variables
+    fi
+
+    if [[ "${prev}" == "-t" ]]; then
+        show_parameters "-b --backend -c --cmake -t -f --force -h --help"
+        return 0 # @TODO show available preset from cmakeFlags
+    fi
+
+    show_parameters "-b --backend -c --cmake -t -f --force -h --help"
+    return 0
+}
+
+_pic-create()
+{
+    # load current string
+    local cur
+    cur="${COMP_WORDS[COMP_CWORD]}"
+    case "$cur" in
+        # check if it is a parameter
+        -*)
+            show_parameters "-f --force -h --help"
+            return 0
+            ;;
+        # default case list directory content
+        *)
+            show_directory "examples"
+            return 0
+            ;;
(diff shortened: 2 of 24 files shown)


ax3l:
Thank you so much, this is great!

I would suggest to just source the completions in all our `picongpu_profile.example` files, do you mind adding it in those? :)

bin/picongpu-completion.bash
```diff
@@ -0,0 +1,232 @@
...
+    local cur
+    cur="${COMP_WORDS[COMP_CWORD]}"
+    case "$cur" in
+        # check if it is a parameter
+        -*)
+            show_parameters "-f --force -h --help"
+            return 0
+            ;;
+        "$PIC_EXAMPLES") #TODO make navigation in PIC_EXAMPLES possible
+            local examples=$(for x in `ls -1 $PIC_EXAMPLES`; do echo "a"; done)
```
ax3l: consider using `$(... $(...) )` for nested calls
```suggestion
            local examples=$(for x in $(ls -1 $PIC_EXAMPLES); do echo "a"; done)
```

bin/picongpu-completion.bash
```diff
@@ -0,0 +1,232 @@
...
+_tbg()
+{
+    # Load current string
+    local cur prev
+    cur="${COMP_WORDS[COMP_CWORD]}"
+    prev="${COMP_WORDS[COMP_CWORD-1]}"
+
+    # List available config files in default directory (inspired by https://debian-administration.org/article/317/An_introduction_to_bash_completion_part_2)
+    if [[ "${prev}" == "-c" || "${prev}" == "--config" ]]; then
+        local names=$(for x in `ls -1 etc/picongpu/*.cfg`; do echo ${x/\/etc\/picongpu\//} ; done )
```
ax3l:
```suggestion
        local names=$(for x in $(ls -1 etc/picongpu/*.cfg); do echo ${x/\/etc\/picongpu\//} ; done )
```

bin/picongpu-completion.bash
```diff
@@ -0,0 +1,232 @@
...
+            ;;
+    esac
+}
+# Invoke the responding function calls for completion of pic-commands
+complete -F _pic-build pic-build
+complete -o nospace -F _pic-create pic-create
+complete -o nospace -F _pic-compile pic-compile
+complete -o nospace -F _pic-configure pic-configure
+complete -F _pic-edit pic-edit
+complete -o nospace -F _tbg tbg
```
ax3l: do not forget the end of file newline for git :)

finnolec:
@ax3l Yes, I'm gonna add it to the profiles :)

finnolec:
@ax3l I added them in b5df523 :)

etc/picongpu/aris-grnet/gpu_picongpu.profile.example
```diff
@@ -88,3 +88,6 @@ function getDevice() {
     fi
     srun  --time=1:00:00 --ntasks-per-node=$numGPUs --cpus-per-task=$((4 * $numGPUs)) --gres=gpu:$numGPUs --mem=$((63000 * numGPUs)) -A $proj -p dvd_usr_prod --pty bash
 }
+
+# Load autocompletion for picongpu scripts
```
ax3l: Maybe call this `... PIConGPU commands`?

ax3l:
> on `pic-create $PIC_EXAMPLES/<TAB>`. Do you mean the generated backslash in front of it? (pic-create \$PIC_EXAMPLES/$) This is an error caused by the bash version you are using.

No, I do not mean that and my bash version is also not doing that :) What I mean is, that if I type `pic-create $PIC_EXAMPLES/<TAB>` nothing is auto-completed.

finnolec:
> `pic-build <TAB>` autocompletes to `pic-build -` instead of showing options first, maybe I do not want any of those and the `-` then must be removed manually. Not sure that is what it shall be doing.

I don't think that there is anything else one can write after the `pic-build` command. So if one presses `<TAB>` I just wanted to show all of the valid arguments for this command. And since they all start with at least one `-`, it autocompletes the minus. The only alternative I see right now, would be, that I allow a directory completion for this command and the options only appear if you type at least one minus. But since there is no usage for a directory as argument for `pic-build`, I don't see the point. So long story short: If one presses `<TAB>` after `pic-build`, it is his/her own fault :smile: 

> the `tbg` auto-completion after args, e.g. `tbg -s -c etc/picongpu/8.cfg -t /<TAB>`, does not complete the filesystem paths anymore.

This is an error, I think I fixed it :) 

> For `-s` and `-b` you can probably circle a full known list for now, and for `-s` only leave the ones in that exist as commands on the system.

That was kinda my plan, but it is a bit more of bash magic, so it might take a while.

> No, I do not mean that and my bash version is also not doing that :) What I mean is, that if I type `pic-create $PIC_EXAMPLES/<TAB>` nothing is auto-completed.

Oops. This was not supposed to happen.

> Ok, makes sense. Or could one suggest the current directory listing of folders as well as `$PIC_EXAMPLES/` ?

I was trying to make something like this work, but it is not too trivial with how bash works with environment variables. I'm still looking into it :)

bin/picongpu-completion.bash
```diff
@@ -0,0 +1,235 @@
...
+    esac
+}
+
+_pic-edit()
+{
+    local cur names
+    cur="${COMP_WORDS[COMP_CWORD]}"
+
+    # Find all param files
+    local names=$(for x in $(ls -1 $PICSRC/include/picongpu/param/*.param); do echo ${x} ; done )
```
ax3l: Already local above?
```suggestion
    names=$(for x in $(ls -1 $PICSRC/include/picongpu/param/*.param); do echo ${x} ; done )
```

finnolec:
@PrometheusPi I think it could help a bit, but z-shell is generally not supported on HPC systems. So we should just stick to bash for simplicity. I'm taking a small break for this pullrequest at this moment, because I have to make myself a better overview of how I want to implement some features.

ax3l:
@PrometheusPi z-shell would force all users to use z-shell.
Unfortunately, that's not what we do ;-) and each shell has its own completion syntax. Bash for starters is excellent.

PrometheusPi:
@finnolec there are many commits (>100!) in your latest push that have nothing to do with the bash complete. Did something go wrong during rebasing?

finnolec:
I'm trying to fix it :D

finnolec:
To summarize, because I removed the git history:
I think I finished the autocompletion for the most important pic-commands. I should've implemented most of @ax3l's suggestions. The last thing I've done is implementing the completion for pic-create, so that it accepts example projects and projects from the current directory as parameters for the source-project. I had to change the `pic-create` command for that as well, but it should still behave as usual for old parameters.
I don't plan on working on the unchecked tickboxes in this PR. I kept TODOs in the skript in the respective places, so that one can implement it if it becomes important. 
This PR is not WIP anymore :)

sbastrakov:
@finnolec thanks for summarizing!

bin/pic-create
```diff
@@ -79,6 +79,10 @@ if [ $# -eq 2 ] ; then
     src_path=$1
     dst_path=$2
     selected_default_folder_to_copy=$folder_to_clone
+    # Use example directory if no source directory available at current directory
+    if [[ -d $PIC_EXAMPLES/$src_path && ! -d ./$src_path ]]; then
```
psychocoderHPC: Can we please add here a check that if the directory exists as local and example directory and print a warning and say that the local directory is used.

finnolec:
I implemented @psychocoderHPC suggestions.

bin/pic-create
```diff
@@ -79,6 +79,14 @@ if [ $# -eq 2 ] ; then
...
+        src_path=$PIC_EXAMPLES/$src_path
+    fi
+
+    if [[ -d $PIC_EXAMPLES/$src_path && -d ./$src_path ]]; then
+        echo "The local directory will be used instead of the PIConGPU example directory for the source path!" >&2
```
psychocoderHPC: suggestion: `warning: 'src_dir' is set to the local directory ./$src_path`

psychocoderHPC: While looking again to it maybe it is better to add before the copy starts a message: `clone $dst_path from $src_path`

finnolec:
And I implemented the changes from the offline discussion with @psychocoderHPC :)

ax3l:
Sweet, glad to see this :)

Can you potentially pls update the spack recipe, so also spack users get the fully benefit of this new feature? :)
https://github.com/ComputationalRadiationPhysics/spack-repo/tree/gh-pages/packages/picongpu

finnolec:
@ax3l I would like to do that, but I'm not really sure how spack is working. What is the default profile spack is loading in [this line](https://github.com/ComputationalRadiationPhysics/spack-repo/blob/8dc07b7bbb3e53257e555cc58dff6da3fcd701c2/packages/picongpu/package.py#L136)? From my understanding the line `source bin/picongpu-completion.bash` will be be in the `.profile`-file. This file has to be written manually by everyone anyway, doesn't it? And the line is called in all of the example profiles on readthedocs...

ax3l:
Yes, it's probably easiest to source your completion in the little stub [`picongpu.profile`](https://github.com/ComputationalRadiationPhysics/spack-repo/tree/gh-pages/packages/picongpu) that I linked.

If you add it at the end of the file, there will be a `PICSRC` environment variable set (after `spack load` which sets this up in `setup_run_environment`.

finnolec:
Were the Problems fixed with #3133 ? :D 
Otherwise the bash completion should be activated by adding the `bin` folder to the path in [this line](https://github.com/ComputationalRadiationPhysics/spack-repo/blob/8dc07b7bbb3e53257e555cc58dff6da3fcd701c2/packages/picongpu/package.py#L168) shouldn't it?

PrometheusPi:
@ax3l The last question by @finnolec was for you - he just forgot to mention you.

ax3l:
Looks good to me now, as suggested above. Thanks!

Issue #3060, opened 2019-09-17
travis-ci suggestions only valid for GNU sed


PrometheusPi:
The suggestions to automatically fix malformatted printed by  the travis-ci dialog are great to quickly solve formating issues - but they are only valid if une used GNU sed. They cause various issues when using BSD sed, e.g.:
 - `\t` is only a tab in GNU, on BSD a real tab must be used (or alternatively the `tr` program) 
 - `'s/[[:blank:]]\+$//'` does not remove trailing whitespaces
 - the flag `-i` requires a file extension 

@ComputationalRadiationPhysics/picongpu-developers Do you think, we should additionally output equivalent commands for BSD?

sbastrakov:
For the Windows version of GNU sed some of recommentations also do not work, while others do. Unfortunately, I do not remember which ones.

ax3l:
I just fixed the same in WarpX for you OSX users. I will share the updates with you :)

psychocoderHPC:
closed with #3073

PR #3073, opened 2019-09-25
Travis CI: OSX sed Support

ax3l:
Make the code style checks work with both GNU sed as well as OSX/vanilla POSIX sed.
This updates the EOL whitespace as well as the TABs to spaces transformations.

There are a few more tests that you can adjust similarly in a follow-up PR :)

Related to #3060

### DIFF
--- a/test/hasEOLwhiteSpace
+++ b/test/hasEOLwhiteSpace
@@ -58,9 +58,11 @@ then
   echo "Run the following command(s) on the above files to remove your"
   echo "end-of-line (EOL) white spaces:"
   echo ""
+  echo "GNU_SED=\$(sed --help >/dev/null 2>&1 && { echo 1; } || { echo 0; })"
+  echo "[[ \${GNU_SED} -eq 1 ]] && REPLACE=\"sed -i 's/[[:blank:]]\+$//'\" || REPLACE=\"sed -i '' -E 's/[[:blank:]]+$//'\""
   for i in ${files[@]}
   do
-    echo "sed -i 's/[[:blank:]]\+$//' $i"
+    echo "eval \${REPLACE} $i"
   done
 fi
 
--- a/test/hasTabs
+++ b/test/hasTabs
@@ -56,9 +56,11 @@ then
   echo "Run the following command(s) on the above files to replace your TABs"
   echo "with four white spaces:"
   echo ""
+  echo "GNU_SED=\$(sed --help >/dev/null 2>&1 && { echo 1; } || { echo 0; })"
+  echo "[[ \${GNU_SED} -eq 1 ]] && REPLACE=\"sed -i 's/\t/\ \ \ \ /g'\" || REPLACE=\"sed -i '' -E 's/\$(printf '\t')/\ \ \ \ /g'\""
   for i in ${files[@]}
   do
-    echo "sed -i 's/\t/    /g' $i"
+    echo "eval \"\${REPLACE}\" $i"
   done
 fi
 


test/hasTabs
```diff
@@ -56,9 +56,11 @@ then
   echo "Run the following command(s) on the above files to replace your TABs"
   echo "with four white spaces:"
   echo ""
+  echo "GNU_SED=\$(sed --help >/dev/null 2>&1 && { echo 1; } || { echo 0; })"
+  echo "[[ \${GNU_SED} -eq 1 ]] && REPLACE=\"sed -i 's/\t/\ \ \ \ /g'\" || REPLACE=\"sed -i '' -E 's/\$(printf '\t')/\ \ \ \ /g'\""
```
PrometheusPi: This evaluetes to:
```
sed -i '' -E 's/ /\ \ \ \ /g'
```
on MacOS and thus does not replace tabs.

PrometheusPi:
```suggestion
  echo "[[ \${GNU_SED} -eq 1 ]] && REPLACE=\"sed -i 's/\t/\ \ \ \ /g'\" || REPLACE=\"sed -i '' -E \\\"s/\\\$(printf '\t')/\ \ \ \ /g\\\"\""
```

PrometheusPi: With these changes, it works on my Mac.

ax3l: Wait, how did you check this? Please execute  the script and copy past what it prints on the terminal, as this is run in travis-ci.

We tested this in our repo and the suggested changes worked, as far as I remember. Just double checking how you executed this to be sure.

PrometheusPi: That is exactly what I tested - I will do it again and print the output below:

PrometheusPi: original text
```
this is a test text ending in many white spaces      
This is a text with a	tab.
```
in detail `od -t a tmp` leads to
```
0000000    t   h   i   s  sp   i   s  sp   a  sp   t   e   s   t  sp   t
0000020    e   x   t  sp   e   n   d   i   n   g  sp   i   n  sp   m   a
0000040    n   y  sp   w   h   i   t   e  sp   s   p   a   c   e   s  sp
0000060   sp  sp  sp  sp  sp  nl   T   h   i   s  sp   i   s  sp   a  sp
0000100    t   e   x   t  sp   w   i   t   h  sp   a  ht   t   a   b   .
0000120   nl                                                            
0000121
```

Execute: `GNU_SED=$(sed --help >/dev/null 2>&1 && { echo 1; } || { echo 0; })`
(result of `$GNU_SED` is `0`)

Execute
```
[[ ${GNU_SED} -eq 1 ]] && REPLACE="sed -i 's/\t/\ \ \ \ /g'" || REPLACE="sed -i '' -E 's/$(printf '\t')/\ \ \ \ /g'"
```
as copied from above.
`echo $REPLACE` returns `sed -i '' -E 's/ /\ \ \ \ /g'` - which is not the intended output for OS X because the `printf` is evaluated before.

Running `eval "${REPLACE}"  tmp` (`tmp` is my file name) leads to the following text:
```
this is a test text ending in many white spaces      
This is a text with a    tab.
```
and `od -t a` reveals:
```
0000000    t   h   i   s  sp   i   s  sp   a  sp   t   e   s   t  sp   t
0000020    e   x   t  sp   e   n   d   i   n   g  sp   i   n  sp   m   a
0000040    n   y  sp   w   h   i   t   e  sp   s   p   a   c   e   s  sp
0000060   sp  sp  sp  sp  sp  nl   T   h   i   s  sp   i   s  sp   a  sp
0000100    t   e   x   t  sp   w   i   t   h  sp   a  sp  sp  sp  sp   t
0000120    a   b   .  nl                                                
0000124
```
Thus the tab has been changed.

PrometheusPi: However, the issue is caused by the quatation marks: in your post above you added them:
```diff
+eval "${REPLACE}"  <file1>
-eval ${REPLACE}  <file1>
```
but when I execute `echo "eval \${REPLACE} $i"` from [here(https://github.com/ComputationalRadiationPhysics/picongpu/blob/3b10f00ee45fa9e9203214867aafa3ab3f19f54c/test/hasEOLwhiteSpace#L65)] on OS X or any other system, echo removed the quotation marks `"`. 

If I then evaluate only that what is actually printed: 
```
eval ${REPLACE}  <file1>
```
all apaces are replaced by for spaces leading to the following text (again `od -t a` output after the text):
```
this    is    a    test    text    ending    in    many    white    spaces                        
This    is    a    text    with    a	tab.
```
```
0000000    t   h   i   s  sp  sp  sp  sp   i   s  sp  sp  sp  sp   a  sp
0000020   sp  sp  sp   t   e   s   t  sp  sp  sp  sp   t   e   x   t  sp
0000040   sp  sp  sp   e   n   d   i   n   g  sp  sp  sp  sp   i   n  sp
0000060   sp  sp  sp   m   a   n   y  sp  sp  sp  sp   w   h   i   t   e
0000100   sp  sp  sp  sp   s   p   a   c   e   s  sp  sp  sp  sp  sp  sp
0000120   sp  sp  sp  sp  sp  sp  sp  sp  sp  sp  sp  sp  sp  sp  sp  sp
0000140   sp  sp  nl   T   h   i   s  sp  sp  sp  sp   i   s  sp  sp  sp
0000160   sp   a  sp  sp  sp  sp   t   e   x   t  sp  sp  sp  sp   w   i
0000200    t   h  sp  sp  sp  sp   a  ht   t   a   b   .  nl            
0000215
```

PrometheusPi: @ax3l Are you shure that on the machine running travis CI, the quatation markes from [line 65](https://github.com/ComputationalRadiationPhysics/picongpu/blob/3b10f00ee45fa9e9203214867aafa3ab3f19f54c/test/hasEOLwhiteSpace#L65) are printed? 

If not, another solution than the one proposed by me would be to use backslashes before the  quotation marks.

PrometheusPi:
@ax3l If you apply the alternative fix, this could be merged.

PrometheusPi:
@ax3l Do you want to continue working on it - or should I just add the fix and merge it? 

(I kept this pull request open, since the missing quotation mark error will most likely also effect your Warp/WarpX addition.)

test/hasEOLwhiteSpace
```diff
@@ -58,9 +58,11 @@ then
   echo "Run the following command(s) on the above files to remove your"
   echo "end-of-line (EOL) white spaces:"
   echo ""
+  echo "GNU_SED=\$(sed --help >/dev/null 2>&1 && { echo 1; } || { echo 0; })"
+  echo "[[ \${GNU_SED} -eq 1 ]] && REPLACE=\"sed -i 's/[[:blank:]]\+$//'\" || REPLACE=\"sed -i '' -E 's/[[:blank:]]+$//'\""
   for i in ${files[@]}
   do
-    echo "sed -i 's/[[:blank:]]\+$//' $i"
+    echo "eval \${REPLACE} $i"
```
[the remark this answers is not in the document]
PrometheusPi: @ax3l This is what you correctly added in https://github.com/ECP-WarpX/WarpX/pull/393/files but missed in this commit.

PrometheusPi:
@ax3l I will add the change myself. (I assume you are too busy with other work.)

test/hasTabs
ax3l: > but when executing echo "eval \${REPLACE} $i" from here

do not execute from *there*. All the quoting is only added to print copy-pasteable output to the console and travis-ci output. That's what we then copy from.

ax3l: Now you are pointing to the hasEOL script while talking about the hasTabs script. they are different.

test/hasEOLwhiteSpace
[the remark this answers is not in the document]
ax3l: The EOL whitespace check prints on the terminal:
```
$ ./test/hasEOLwhiteSpace include/
./include/picongpu/main.cpp contains EOL white spaces!

SUMMARY
-------
Run the following command(s) on the above files to remove your
end-of-line (EOL) white spaces:

GNU_SED=$(sed --help >/dev/null 2>&1 && { echo 1; } || { echo 0; })
[[ ${GNU_SED} -eq 1 ]] && REPLACE="sed -i 's/[[:blank:]]\+$//'" || REPLACE="sed -i '' -E 's/[[:blank:]]+$//'"
eval ${REPLACE} ./include/picongpu/main.cpp
```
and then you copy-paste the fix which is
```bash
GNU_SED=$(sed --help >/dev/null 2>&1 && { echo 1; } || { echo 0; })
[[ ${GNU_SED} -eq 1 ]] && REPLACE="sed -i 's/[[:blank:]]\+$//'" || REPLACE="sed -i '' -E 's/[[:blank:]]+$//'"
eval ${REPLACE} ./include/picongpu/main.cpp
```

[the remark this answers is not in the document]
ax3l: Just tested again with macOS HighSierra and Bash 3.2.57 and this works.

test/hasTabs
ax3l: Just tested again with macOS HighSierra and Bash 3.2.57 and this works.

ax3l:
Commented above. I think you are using the script in a wrong manner and you are confusing both scripts as well. Please follow the instructions [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3073#discussion_r341410361).

I just tested both scripts again on macOS HighSierra and Bash 3.2.5 and they work as they should (with macOS POSIX sed).

PrometheusPi:
@ax3l You misunderstood me, this is not a version issue. I just think that what the CI prints misses quotation marks. However the easiest way to validate it is to test this 😄. Even if it does not work for macOS, we will not break anything since it currently does not provide POSIX compatible anyway. 

Thus I will merge this now and commit to test pull request (white space and tabs) tomorrow, to test the printed output if you are fine with it.

I hope you are right with the output. If not, a fix can be provided quickly and easily. 

Just for completeness:
 - `sed --version` 
POSIX `sed`does neither support `--version` nor `-v` (there is no version output at all) 
 - `bash --version`
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
 - `which sed`
/usr/bin/sed
 - `which bash`
/bin/bash
 - `echo $TERM`
xterm-256color
 - `echo $SHELL`
/bin/bash

test/hasTabs
```diff
@@ -56,9 +56,11 @@ then
...
   for i in ${files[@]}
   do
-    echo "sed -i 's/\t/    /g' $i"
+    echo "eval \"\${REPLACE}\" $i"
```
PrometheusPi: @ax3l you are right - in the tab script, there are the extra quotation marks 👍

ax3l:
> POSIX sed does neither support --version nor -v (there is no version output at all)

Yes, that's why I test for it as a flag to indicate GNU/non-GNU ;)
Thanks!

PrometheusPi:
Both tests passed successfully on OS X and Ubuntu. @ax3l please excuse that I did not trust your expertise from the start - you were right: your addition worked flawlessly.

ax3l:
Excellent, great this works for you as well :)

PR #3074, opened 2019-09-26
Docs: create conda env for building docs

n01r:
This PR adds a way to create a clean conda environment just for building our docs.
This seems useful since installing the required packages might otherwise break users' previous Python configurations.

- [x] try out creating the environment
- [x] build `dev` docs from it

### DIFF
--- a/docs/picongpu-docs-env.yml
+++ b/docs/picongpu-docs-env.yml
@@ -1,12 +1,13 @@
 # Create a separate conda environment which fulfills the package requirements
 # for building the documentation
-# E.g. run `conda create --file picongpu-docs-env.yml`
+# E.g. run `conda env create --file picongpu-docs-env.yml`
 #
 
 name: picongpu-docs-env
 channels:
     - conda-forge
 dependencies:
+    - doxygen
     - sphinx_rtd_theme>=0.3.1
     - recommonmark
     - sphinx==2.0.1
--- a/docs/picongpu-docs-env.yml
+++ b/docs/picongpu-docs-env.yml
@@ -5,17 +5,11 @@
 
 name: picongpu-docs-env
 channels:
-    - conda-forge
+  - conda-forge
 dependencies:
-    - doxygen
-    - sphinx_rtd_theme>=0.3.1
-    - recommonmark
-    - sphinx==2.0.1
-    - breathe>=4.12.0
-    - sphinxcontrib-programoutput
-    - sphinxcontrib-napoleon>=0.7
-    - pygments
-    - matplotlib
-    - scipy
-    - numpy
+  - python>=3.7
+  - doxygen
+  - pip
+  - pip:
+    - "-r requirements.txt"
 
--- a/docs/picongpu-docs-env.yml
+++ b/docs/picongpu-docs-env.yml
@@ -1,0 +1,20 @@
+# Create a separate conda environment which fulfills the package requirements
+# for building the documentation
+# E.g. run `conda create --file picongpu-docs-env.yml`
+#
+
+name: picongpu-docs-env
+channels:
+    - conda-forge
+dependencies:
+    - sphinx_rtd_theme>=0.3.1
+    - recommonmark
+    - sphinx==2.0.1
+    - breathe>=4.12.0
+    - sphinxcontrib-programoutput
+    - sphinxcontrib-napoleon>=0.7
+    - pygments
+    - matplotlib
+    - scipy
+    - numpy
+
--- a/docs/source/dev/sphinx.rst
+++ b/docs/source/dev/sphinx.rst
@@ -39,6 +39,18 @@ The following requirements need to be installed (once) to build our documentatio
     # python tools & style theme
     pip install -r requirements.txt # --user
 
+However, in order to not break any of your existing Python configurations you can also create a new environment that you only use for building the documentation.
+
+.. code-block:: bash
+
+    cd docs/
+
+    # create a bare conda environment containing just all the requirements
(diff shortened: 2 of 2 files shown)


docs/source/dev/sphinx.rst
```diff
@@ -39,6 +39,18 @@ The following requirements need to be installed (once) to build our documentatio
     # python tools & style theme
     pip install -r requirements.txt # --user
 
+However, in order to not break any of your existing Python configurations you can also create a new environment that you only use for building the documentation.
```
PrometheusPi: Required: there needs to be a comma in `... configurations, you can ...`
Optional: remove `However,`

PrometheusPi:
Great idea to add this! :+1:

docs/picongpu-docs-env.yml
```diff
@@ -0,0 +1,20 @@
...
+    - conda-forge
+dependencies:
+    - sphinx_rtd_theme>=0.3.1
+    - recommonmark
+    - sphinx==2.0.1
+    - breathe>=4.12.0
+    - sphinxcontrib-programoutput
+    - sphinxcontrib-napoleon>=0.7
+    - pygments
+    - matplotlib
```
PrometheusPi: why are those decencies below required?

n01r: Since some plots in our documentation (see https://picongpu.readthedocs.io/en/0.4.3/models/field_ionization.html for instance) are created on-the-fly and are not saved as images.

n01r:
From an offline conversation with @PrometheusPi:
I added doxygen to the packages that conda installs, which is a nice way to do it when you do not have superuser rights on the system you want to build the documentation locally on.

docs/picongpu-docs-env.yml
```diff
@@ -0,0 +1,21 @@
+# Create a separate conda environment which fulfills the package requirements
+# for building the documentation
+# E.g. run `conda env create --file picongpu-docs-env.yml`
+#
+
+name: picongpu-docs-env
+channels:
+    - conda-forge
+dependencies:
+    - doxygen
```
ax3l: this duplicates the dependencies in `docs/requirements.txt` and now you will always have to update *two* positions instead of one.
Why is a pip virtual environment not sufficient?

ax3l: In order to avoid duplicating the entries that are already in `docs/requirements.txt`, you can try the following here:
```yaml
# ...
dependencies:
    - doxygen
    - pip
    - pip:
          - "-r requirements.txt"

File: docs/source/dev/sphinx.rst [hunk 4]
```diff
@@ -39,6 +39,20 @@ The following requirements need to be installed (once) to build our documentatio
     # python tools & style theme
     pip install -r requirements.txt # --user
 
+In order to not break any of your existing Python configurations, you can also create a new environment that you only use for building the documentation.
+Since it is possible to install doxygen with conda, the following method employs this way.
```
ax3l:
```suggestion
Since it is possible to install doxygen with conda, the following demonstrates this.
```

docs/picongpu-docs-env.yml
n01r: Thanks for the suggestion! That's a valid point :)
I'm gonna do it like that, then.

docs/source/dev/sphinx.rst
n01r: Implemented your suggestions in one commit

docs/source/dev/sphinx.rst
```diff
@@ -39,6 +39,20 @@ The following requirements need to be installed (once) to build our documentatio
...
+Since it is possible to install doxygen with conda, the following demonstrates this.
+
+.. code-block:: bash
+
+    cd docs/
+
+    # create a bare conda environment containing just all the requirements
+    # for building the picongpu documentation
+    # note: also installs doxygen inside this environment
+    conda env create picongpu-docs-env.yml
```
PrometheusPi: when I run thi scommand, I get the following error:
```bash
SpecNotFound: Invalid name, try the format: user/package
```

PrometheusPi: Might be darwin (MacOS) related issue.

PrometheusPi: I am using `conda 4.4.10`

PrometheusPi:
```suggestion
    conda env create --file picongpu-docs-env.yml
```

n01r: Haha, sorry, the `--file` was missing which is correct in the file comment itself 
Good find!

docs/source/dev/sphinx.rst
```diff
@@ -39,6 +39,20 @@ The following requirements need to be installed (once) to build our documentatio
...
+.. code-block:: bash
+
+    cd docs/
+
+    # create a bare conda environment containing just all the requirements
+    # for building the picongpu documentation
+    # note: also installs doxygen inside this environment
+    conda env create --file picongpu-docs-env.yml
+
+    # start up the environment
```
PrometheusPi:
```suggestion
    # start up the environment as suggested during its creation e.g.
```

docs/source/dev/sphinx.rst
```diff
@@ -39,6 +39,20 @@ The following requirements need to be installed (once) to build our documentatio
...
+    conda activate picongpu-docs-env
```
PrometheusPi:
```suggestion
    conda activate picongpu-docs-env
    # or
    source activate picongpu-docs-env    
```

PR #3078, opened 2019-09-30
added silicon element to ioniser

hightower8083:
For some experiments, for example harmonics from solids, people use silicon-based targets (e.g. fused silica), so it would be good to have `Si` included to the _commonly used elements_ list. This PR fixes it -- i took ionisation potentials from `mendeleev`  and effective nuclear charge from [wiki](http://en.wikipedia.org/wiki/Effective_nuclear_charge) (as recommended).

### DIFF
--- a/include/picongpu/param/ionizationEnergies.param
+++ b/include/picongpu/param/ionizationEnergies.param
@@ -137,6 +137,24 @@ namespace AU
         2304.14 * UNITCONV_eV_to_AU
     );
 
+    /* ionization energy for aluminium in atomic units */
+    PMACC_CONST_VECTOR(float_X, 14, Silicon,
+        8.151683 * UNITCONV_eV_to_AU,
+        16.345845 * UNITCONV_eV_to_AU,
+        33.493 * UNITCONV_eV_to_AU,
+        45.14179 * UNITCONV_eV_to_AU,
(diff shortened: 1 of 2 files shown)


include/picongpu/param/ionizationEnergies.param
```diff
@@ -137,6 +137,24 @@ namespace AU
         2304.14 * UNITCONV_eV_to_AU
     );
 
+    /* ionization energy for aluminium in atomic units */
```
sbastrakov: `aluminium` is probably a typo?

hightower8083: yep --  copy/paste remnant..

include/picongpu/param/ionizationEnergies.param
```diff
@@ -137,6 +137,24 @@ namespace AU
         2304.14 * UNITCONV_eV_to_AU
     );
 
+    /* ionization energy for silicon in atomic units */
+    PMACC_CONST_VECTOR(float_X, 14, Silicon,
+        8.151683 * UNITCONV_eV_to_AU,
```
sbastrakov: This has nothing to do with your PR, just a side note for future self: these constants are double, should be changed to `..._X` to be `float_X`.

PR #3084, opened 2019-10-10
fix CUDA device selection

psychocoderHPC:
If a CUDA device is pre-selected to a MPI rank by e.g. the batch system
with environment variable `CUDA_VISIBLE_DEVICES` than the current
implementations thows an error.

This issue was first seen in #3007

- query device properties for each device instead only for once
- throw an error if no device can be selected

### DIFF
--- a/include/pmacc/Environment.hpp
+++ b/include/pmacc/Environment.hpp
@@ -449,26 +449,39 @@ namespace detail
         {
             throw std::runtime_error("no CUDA capable devices detected");
         }
-        else if (deviceNumber >= num_gpus) //check if device can be selected by deviceNumber
-        {
-            std::cerr << "no CUDA device " << deviceNumber << ", only " << num_gpus << " devices found" << std::endl;
-            throw std::runtime_error("CUDA capable devices can't be selected");
-        }
 #endif
 
         int maxTries = num_gpus;
-#if (PMACC_CUDA_ENABLED == 1)
-        cudaDeviceProp devProp;
-        CUDA_CHECK((cuplaError_t)cudaGetDeviceProperties(&devProp, deviceNumber));
-        /* if the gpu compute mode is set to default we use the given `deviceNumber` */
-        if (devProp.computeMode == cudaComputeModeDefault)
-            maxTries = 1;
-#endif
+        bool deviceSelectionSuccessful = false;
+
         cudaError rc;
 
+        // search the first selectable device in the compute node
         for (int deviceOffset = 0; deviceOffset < maxTries; ++deviceOffset)
         {
+            /* Modulo 'num_gpus' avoids invalid device indices for systems where the environment variable
+             * `CUDA_VISIBLE_DEVICES` is used to pre-select a device.
+             */
             const int tryDeviceId = (deviceOffset + deviceNumber) % num_gpus;
+
+            log<ggLog::CUDA_RT>("Trying to allocate device %1%.") % tryDeviceId;
+#if (PMACC_CUDA_ENABLED == 1)
+            cudaDeviceProp devProp;
+            CUDA_CHECK((cuplaError_t)cudaGetDeviceProperties(&devProp, tryDeviceId));
+
+            /* If the cuda gpu compute mode is 'default'
+             * (https://docs.nvidia.com/cuda/cuda-c-programming-guide/#compute-modes)
+             * then we try to get a device only once.
+             * The index used to select a device is based on the local MPI rank so
+             * that each rank tries a different device.
+             */
+            if (devProp.computeMode == cudaComputeModeDefault)
+            {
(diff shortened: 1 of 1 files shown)


include/pmacc/Environment.hpp
```diff
@@ -449,26 +449,29 @@ namespace detail
...
-#endif
+        bool deviceSelectionSuccessfull = false;
+
         cudaError rc;
 
         for (int deviceOffset = 0; deviceOffset < maxTries; ++deviceOffset)
         {
             const int tryDeviceId = (deviceOffset + deviceNumber) % num_gpus;
+
+            log<ggLog::CUDA_RT>("Try to allocate device %1%.") % tryDeviceId;
```
sbastrakov: Maybe replace `Try` with `Trying` for more clarity.

include/pmacc/Environment.hpp
```diff
@@ -520,6 +524,11 @@ namespace detail
                 CUDA_CHECK(rc); /*error message*/
             }
         }
+        if(!deviceSelectionSuccessfull)
+        {
+            std::cerr << "Failed to select one of the " << deviceNumber << " devices." << std::endl;
```
sbastrakov: Why `deviceNumber` here? This variable looks to be the index of a device, not the number of devices (which is `num_gpus`).

include/pmacc/Environment.hpp
```diff
@@ -449,26 +449,29 @@ namespace detail
...
 
         for (int deviceOffset = 0; deviceOffset < maxTries; ++deviceOffset)
         {
             const int tryDeviceId = (deviceOffset + deviceNumber) % num_gpus;
+
+            log<ggLog::CUDA_RT>("Try to allocate device %1%.") % tryDeviceId;
+#if (PMACC_CUDA_ENABLED == 1)
+            cudaDeviceProp devProp;
+            CUDA_CHECK((cuplaError_t)cudaGetDeviceProperties(&devProp, tryDeviceId));
+            /* if the gpu compute mode is set to default we try to get a device only once */
```
sbastrakov: Not sure, but maybe this needs additional explanation. Like for now, it is not obvious for me that we do not break smth else with this change.

psychocoderHPC: I will extend the doc string a little bit and link the [cuda documentation](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#compute-modes) for compute modes.

sbastrakov:
Looks like there was smth wrong with our CI, hence it failed. Restarted.

include/pmacc/Environment.hpp
```diff
@@ -449,26 +449,35 @@ namespace detail
...
 
         int maxTries = num_gpus;
-#if (PMACC_CUDA_ENABLED == 1)
-        cudaDeviceProp devProp;
-        CUDA_CHECK((cuplaError_t)cudaGetDeviceProperties(&devProp, deviceNumber));
-        /* if the gpu compute mode is set to default we use the given `deviceNumber` */
-        if (devProp.computeMode == cudaComputeModeDefault)
-            maxTries = 1;
-#endif
+        bool deviceSelectionSuccessfull = false;
```
ax3l: typo: 
```suggestion
        bool deviceSelectionSuccessful = false;
```

include/pmacc/Environment.hpp
```diff
@@ -449,26 +449,35 @@ namespace detail
...
             const int tryDeviceId = (deviceOffset + deviceNumber) % num_gpus;
+
+            log<ggLog::CUDA_RT>("Trying to allocate device %1%.") % tryDeviceId;
+#if (PMACC_CUDA_ENABLED == 1)
+            cudaDeviceProp devProp;
+            CUDA_CHECK((cuplaError_t)cudaGetDeviceProperties(&devProp, tryDeviceId));
+
+            /* if the cuda gpu compute mode is set to 'default'
+             * (https://docs.nvidia.com/cuda/cuda-c-programming-guide/#compute-modes)
+             * than we will use the first device. The device id selected based on
```
ax3l: I do not understand what the last sentence means. Can you please clarify *how* this depends on the local MPI rank? (And we should fix the grammar in the sentence.)

I also wonder if this routine is potentially prone to race conditions during the selection?

PrometheusPi:
@ax3l might the "default" compile test be still pending because we had a network downtime yesterday here at HZDR? 
@psychocoderHPC If this is the case you might to re-trigger the compile-suite via another push.

include/pmacc/Environment.hpp
psychocoderHPC: In the case we are working on a process exclusive GPU system there are no races. On a system where the gpu is running in default mode, we use as gpu index the local MPI rank. This means there is no issue with races.

Issue #3085, opened 2019-10-10
Update ORNL System Guide Links


ax3l:
This week, ORNL moved its documentation to the new pages at https://docs.olcf.ornl.gov/

(Yay, Sphinx! Repo: https://github.com/olcf/olcf-user-docs/ )

We potentially have to update some of the links in our manual/documentation for this. PRs welcome!

PrometheusPi:
I will update the links.

PrometheusPi:
Notes:
This seems to be not rue for `https://www.olcf.ornl.gov/center-projects/adios/` but for various other `http://www.olcf.ornl.gov` links. - will check them manually

PrometheusPi:
Okay - since ADIOS is still on the old web page and the only documentation references we have are for the decommissioned Titan cluster updating any links is not needed. 

But the developer of #3007 (which is you @ax3l) should keep the updated URL in mind 😉 .

I will close the issue now and will push a pull request to remove titan references.

ax3l:
ADIOS also has a new homepage, they prefer to link to https://csmd.ornl.gov/adios even though the old link to https://www.olcf.ornl.gov/center-projects/adios still exists.

PrometheusPi:
@ax3l Thanks for updating the status of the ADIOS webpage - I will have a look at our code and update the URLs accordingly.

PR #3086, opened 2019-10-11
remove titan ornl setup and doc

PrometheusPi:
Since [2019-08-02 the Titan system at ORNL is decommissioned](https://www.olcf.ornl.gov/olcf-resources/compute-systems/titan/). This pull request removes all references to and setups for this cluster in the current `dev` to clean up our documentation and setup directory .

If you want to see the setup of this system, please refer to a previous version of PIConGPU. 

(The outdated documentation showed up as part of the research for issue #3085.)


PrometheusPi:
@sbastrakov Thanks for pointing this out - I didn't know this workaround - perhaps @psychocoderHPC or @ax3l know more - but I agree, if it is not certain it is only effecting Titan, we should keep it.

ax3l:
That workaround was only for tracing. Can probably go.

psychocoderHPC:
RIP :cry:

PrometheusPi:
@ax3l Okay - I will remove it.

PR #3089, opened 2019-10-17
fix doc transition radiation plugin

PrometheusPi:
This pull request resolves some errors and incorrect wordings in the transition radiation plugin documentation and adds a command line flag configuration to the `TBG_macro.cfg`.

The plugin was just recently introduced to the `dev` #3003.  

The fixes are:
 - `transitionRaditionMask` is a flag not an attribute
 - the command line names were not correct

@psychocoderHPC strangely I could add the flag `transitionRadiationMask` to the particle attributes without causing a compile time error (of course the plugin wasn't compiled) - shouldn't adding a flag to the attributes cause an error?

### DIFF
--- a/docs/source/usage/plugins/transitionRadiation.rst
+++ b/docs/source/usage/plugins/transitionRadiation.rst
@@ -189,8 +189,6 @@ Gamma filter
 In order to consider the radiation only of particles with a gamma higher than a specific threshold.
 In order to do that, the radiating particle species needs the flag ``transitionRadiationMask`` (which is initialized as ``false``) which further needs to be manipulated, to set to true for specific (random) particles.
 
-It can also avoid division by 0 errors in the transition radiation output, which are caused by slowly, but transversely moving electrons.
-
 Using a filter functor as:
 
 .. code:: cpp
--- a/docs/source/usage/plugins/transitionRadiation.rst
+++ b/docs/source/usage/plugins/transitionRadiation.rst
@@ -58,7 +58,7 @@ There are no external dependencies.
 .param files
 ^^^^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


docs/source/usage/plugins/transitionRadiation.rst
```diff
@@ -216,7 +216,8 @@ For a specific (charged) species ``<species>`` e.g. ``e``, the radiation can be
 ========================================== ==============================================================================================================================
 Command line option                        Description
 ========================================== ==============================================================================================================================
-``--<species>_transitionRadiation.period`` Gives the number of time steps between which the radiation should be calculated.
+``--<species>_transRad.period``            Gives the number of time steps between which the radiation should be calculated.
+``--<species>_transRad.omegaList```        Currently deprecated
```
finnolec: Why are you still listing this command line option, even though it does not appear in [TransitionRadiation.hpp](https://github.com/ComputationalRadiationPhysics/picongpu/blob/f495273b3a398fbe45ca08feb72a74d15b7b18be/include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp#L176)? Is it the `omegaList` still used somewhere else?

PrometheusPi: @finnolec Great catch - I just copy pasted the wrong documentation 😞

finnolec:
> It can also avoid division by 0 errors in the transition radiation output, which are caused by slowly, but transversely moving electrons.

This should actually not be up to date, since I'm checking for transversely moving particles in [transitionRadiation/Calculator.hpp](https://github.com/ComputationalRadiationPhysics/picongpu/blob/f495273b3a398fbe45ca08feb72a74d15b7b18be/include/picongpu/plugins/transitionRadiation/Calculator.hpp#L230). Do you want to remove the line from the documentation with this pull request as well? 
I tested the if-case just now with electrons moving solely on the x-axis and electrons moving solely on the z-axis and it didn't produce any NaNs. So I'd guess that this should not be a problem anymore :)

PrometheusPi:
@finnolec Yes great idea, I will add the change in the docs with this pr

PrometheusPi:
The pull request is ready for review.

PR #3091, opened 2019-10-21
Fix observation direction of transition radiation plugin

### LINKED ISSUE #3090: transition radiation plugin produces NaNs
Currently the transition radiation plugin introduces with #3003 produces NaNs for example in the LWFA simulations for @OmidZar. This is most likely caused by a division by zero as documented in the [gamma filter section of the read the docs documentation](https://picongpu.readthedocs.io/en/latest/usage/plugins/transitionRadiation.html#gamma-filter).

This (hopefully) can be avoided in the LWFA case by increasing the gamma value of the filter (I will try `5` -> `20`) but this is not a general solution. 

As discussed *offline* with @psychocoderHPC, such divisions by zero should be stopped in the kernel (if case is fine, as long as no extra code is executed and only contribution to summation is stopped).

finnolec:
Allow the case of a single azimuth or polar angle for transition radiation calculation. 

Additionally I added a bit of documentation to make it more clear, what happens if you calculate in a single azimuth / polar direction. This should resolve #3090.

I checked the output of simulations quickly (for the 4 following parameters: `nPhi = 128` and `nTheta = 128`, `nPhi = 1` and `nTheta = 128`, `nPhi = 128` and `nTheta = 1` and `nPhi = 1` and `nTheta = 1` and there are no NaNs anymore and the calculated values look good. But I didn't change the angle discretization, so this is to be expected. 

Edit: There are no test at the moment, but I don't see why it shouldn't work since it's just an addition. 
Edit 2: I tested for `nPhi = 1` and `nTheta = 1` and there are no NaNs.

@PrometheusPi @sbastrakov

### DIFF
--- a/include/picongpu/param/transitionRadiation.param
+++ b/include/picongpu/param/transitionRadiation.param
@@ -132,7 +132,11 @@ namespace listFrequencies
 
     namespace parameters
     {
-        // number of observation directions
+        /** Number of observation directions
+         *
+         * If nPhi or nTheta is equal to 1, the transition radiation will be calculated
+         * for phiMin or thetaMin respectively.
+         */
         constexpr unsigned int nPhi = 128;
         constexpr unsigned int nTheta = 128;
         constexpr unsigned int nObserver = nPhi * nTheta;
@@ -216,6 +220,9 @@ namespace listFrequencies
      * 128X128 detectors ranging from 0 to pi for the azimuth angle
      * theta and from 0 to 2 pi for the polar angle phi.
      *
+     * If the calculation is only supposed to be done for a single azimuth
+     * or polar angle, it will use the respective minimal angle.
+     *
      * @param    observation_id_extern
      *           int index that identifies each block on the GPU
      *           to compute the observation direction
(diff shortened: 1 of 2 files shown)


finnolec:
I put it into a ternary operator as @sbastrakov suggested to make it look more clean :) I also squashed it already.

include/picongpu/param/transitionRadiation.param
```diff
@@ -233,20 +240,23 @@ namespace listFrequencies
...
         /** get index for computing angle phi: */
         const int indexPhi = observation_id_extern % parameters::nPhi;
 
-        /** step width angle theta */
-        const picongpu::float_64 deltaTheta = ( parameters::thetaMax -
-            parameters::thetaMin ) / ( parameters::nTheta - 1.0 );
         /** step width angle phi */
-        const picongpu::float_64 deltaPhi = ( parameters::phiMax -
-            parameters::phiMin ) / ( parameters::nPhi - 1.0 );
+        const picongpu::float_64 deltaPhi = ( parameteters::nPhi > 1 ) ?
```
sbastrakov: Typo in `parameters`. (The tests failed because of it). Otherwise I like your changes, thanks!

PR #3101, opened 2019-11-07
add reference to thesis for emittance plugin doc

PrometheusPi:
As requested by @ax3l, this pull request adds the reference to @kossag14's master thesis in the slice emittance plugin documentation. 

I tested the sphinx build locally and the reference/link works fine. 

Thanks to @n01r for find/asking for the DOI.

### DIFF
--- a/docs/source/usage/plugins/sliceEmittance.rst
+++ b/docs/source/usage/plugins/sliceEmittance.rst
@@ -7,6 +7,7 @@ The plugin computes the total emittance and the slice emittance (for ten combine
 
 Currently, it outputs only the emittance of the transverse momentum space x-px.
 
+More details on the implementation and tests can be found in the master's thesis [Rudat2019]_. 
 
 External Dependencies
 ^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


docs/source/usage/plugins/sliceEmittance.rst
```diff
@@ -7,6 +7,7 @@ The plugin computes the total emittance and the slice emittance (for ten combine
 
 Currently, it outputs only the emittance of the transverse momentum space x-px.
 
+More details on the implementation and tests can be found in the master thesis [Rudat2019]_.
```
sbastrakov: Not sure, but shouldn't it be `Master`, not `master`?

PrometheusPi: I just checked wikipedia: https://en.wikipedia.org/wiki/Thesis#Germany (I did not know it either.)
It should neither be `master` nor `Master`, it should be `master's` - I will fix it. 😕

PrometheusPi:
@sbastrakov I force pushed the fix.

PR #3103, opened 2019-11-07
Docs: Fix Plugin Table

ax3l:
Someone broke the plugin table markdown. Fixed it.


PrometheusPi:
Great fix - the table now shows up - but something strange happened to the footnotes:
grafik

PrometheusPi:
Just es reference - the footnotes before:
grafik

ax3l:
Dunno, seams unrelated to me.

ax3l:
This behaviour can probably be set as an option somewhere, but do not have time to look it up right now.

PR #3105, opened 2019-11-11
fix gamma filter transition radiation

PrometheusPi:
This pull request fixes a bug in the transition radiation plugin that prevented the gamma filter from working.
The `transitionRadiationMask` was set as flag not as attribute, thus the filter was never applied and assumed to be always true. 

The documentation was updated alongside.

### DIFF
--- a/docs/source/usage/plugins/transitionRadiation.rst
+++ b/docs/source/usage/plugins/transitionRadiation.rst
@@ -58,7 +58,7 @@ There are no external dependencies.
 .param files
 ^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


PrometheusPi:
ready for review/merge

sbastrakov:
@psychocoderHPC are you planning to review it? Looks good to me.

ax3l:
@PrometheusPi if this bug shall show up in the changelog of the next release, do not forget to set the "affects latest release" label :)

PrometheusPi:
@ax3l luckily it just effects the current dev  :tada: - so no need to add this in the changelog

Issue #3033, opened 2019-08-22
Negative mass and density ratios


Anton-Le:
Observed problem:
Although providing a negative `BASE_DENSITY` is not possible this can easily be circumvented by defining a negative `densityRatio`. Same for mass.
Both cases result in a running code that will eat up execution time and not return meaningful results.

Avoidable case of _invalid input porvided by user_.

ax3l:
Thanks for the idea.
We usually do not assume malicious intent by the user and validate for errors, not intentionally unphysical input like potential tachyons that are unlikely to be covered by Maxwell's equations. Negative base densities and masses fall into that category.
But if you like to provide a validation for such input and design it in a way that it is runtime neutral to the kernels that execute this, PRs are very welcome :)

PR #3109, opened 2019-11-12
[WIP] add check negative mass ratio

PrometheusPi:
This pull request adds the compile-time check if `massRatio` is negative. 
This was suggested by @Anton-Le in #3033. 

Thanks @sbastrakov for the help with the `constexpr` implementation. 

We both think this is a hack and there might be a better solution that this hack. (Thus, the pull equest as a WIP flag.)
 @ComputationalRadiationPhysics/picongpu-developers I am open for any suggestions to make this more general and perhaps avoids breaking old setups (which do not use `const_value_identifier` for `massRatio`s).

### DIFF
--- a/include/picongpu/param/speciesDefinition.param
+++ b/include/picongpu/param/speciesDefinition.param
@@ -59,8 +59,8 @@ using DefaultParticleAttributes = MakeSeq_t<
 
 /*--------------------------- photons -------------------------------------------*/
 
-value_identifier( float_X, MassRatioPhotons, 0.0 );
-value_identifier( float_X, ChargeRatioPhotons, 0.0 );
+const_value_identifier( float_X, MassRatioPhotons, 0.0 );
+const_value_identifier( float_X, ChargeRatioPhotons, 0.0 );
 
 using ParticleFlagsPhotons = MakeSeq_t<
     particlePusher< particles::pusher::Photon >,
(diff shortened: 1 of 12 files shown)


sbastrakov:
Just as an additional comment. We can't just add `constexpr` to the `getValue()` method inside `value_identifier`. The issue is that as initial values we sometimes use a result of a non-`constexpr` function call. Then a `constexpr getValue()` will not compile (not just become a non-`constexpr` function).

PrometheusPi:
@psychocoderHPC and @sbastrakov Do you have any suggestions on how to implement this more consistently?

sbastrakov:
I don't see a way without making some kind of `const_value_identifier`. I think what we can do is to conditionally enable the check so it is only done for such compile-time constant identifiers, and so the setups using just normal `value_identifier` would still compile, just without this check.

psychocoderHPC:
What is the state of the PR? Can this be finished within the next two days or not? If not I will change the milestone to 0.6.0

sbastrakov:
The status is the this fix is possible, but not compatible with older setups. And it is not clear if there can be a backward-compatible similar change.

PrometheusPi:
I agree with @sbastrakov - a better abstraction and/or a  backward-compatible version wold be much better. But the code works right now.

psychocoderHPC:
@PrometheusPi , @sbastrakov, @psychocoderHPC  discussed offline: We will close this PR to avoid breaking existing setups. Also we like to express more parts those are not required to be compile time as runtime constructs.
It would be possible to move such a check to the host side (runtime).

PR #3114, opened 2019-11-19
fix wrong used method prefix

psychocoderHPC:
Fix places where the method prefix of the declaration and implementation
is not matching. Bugs found with hip-clang.

The latest release 0.4.3 is also effected but I never got an error before using hip-clang.

### DIFF
--- a/include/picongpu/fields/FieldB.tpp
+++ b/include/picongpu/fields/FieldB.tpp
@@ -42,7 +42,7 @@ namespace picongpu
     {
     }
 
-    FieldB::UnitValueType FieldB::getUnit( )
+    HDINLINE FieldB::UnitValueType FieldB::getUnit( )
     {
         return UnitValueType{ UNIT_BFIELD, UNIT_BFIELD, UNIT_BFIELD };
     }
--- a/include/picongpu/fields/FieldE.tpp
+++ b/include/picongpu/fields/FieldE.tpp
@@ -42,7 +42,7 @@ namespace picongpu
     {
     }
 
-    FieldE::UnitValueType FieldE::getUnit( )
+    HDINLINE FieldE::UnitValueType FieldE::getUnit( )
     {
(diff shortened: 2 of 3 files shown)


include/picongpu/fields/FieldJ.tpp
```diff
@@ -194,7 +194,7 @@ SimulationDataId FieldJ::getUniqueId( )
     return getName( );
 }
 
-HDINLINE
+HINLINE
```
sbastrakov: It should not be required here at all. In case a definition is separate from a declaration, the declaration defines that a function is `inline` and at definition it does not matter whether you add it or not.

include/picongpu/fields/FieldE.hpp
```diff
@@ -55,7 +55,7 @@ namespace picongpu
         using UnitValueType = promoteType< float_64, ValueType >::type;
 
         //! Get units of field components
-        HDINLINE static UnitValueType getUnit( );
+        HINLINE static UnitValueType getUnit( );
```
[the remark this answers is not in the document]
sbastrakov: Same for `getUnit()` in other files.

psychocoderHPC:
found that also nvcc shows a warnings about it:

```
/bigdata/hplsim/scratch/widera/dev/include/pmacc/../picongpu/fields/FieldTmp.tpp(319): warning: a __host__ function("picongpu::FieldTmp::getUnit") redeclared with __host__ __device__, hence treated as a __host__ __device__ function
```

sbastrakov:
From the tests compile issues, it looks like `getUnit()` should really remain host-device as it is called from both sides. So the fix is, I believe, to have `HDINLINE` at both declarations and definitions.

include/picongpu/fields/FieldJ.tpp
psychocoderHPC: I kept it because of all other methods are defined as `HINLINE` too

psychocoderHPC:
I updated this PR and set `getUnit` to `HDINLINE`

PR #3118, opened 2019-11-21
CMake: mark cuda 10.2 as tested

psychocoderHPC:
remove warning that CUDA 10.2 is untested

I tested CUDA 10.2 (runtime/compile) on our test system with a V100. :fireworks: NVIDIA not introduced breaking changes.

Modules used for the test:

```
boost@1.68.0
cmake@3.15.4
cuda@10.2.89
openmpi@3.1.2
```


psychocoderHPC:
I updated the docs too

PR #2966, opened 2019-05-20
openPMD Plugin

franzpoeschel:
Implement writing to and reading from openPMD.
**Todo**
- [x] The plugin is based very closely on the existing [ADIOS plugin](https://github.com/ComputationalRadiationPhysics/picongpu/tree/master/include/picongpu/plugins/adios) and carries some baggage. Especially the strict separation of (1) declaration of datasets (ADIOS variables) and (2) writing data to them is unnecessary from openPMD's point of view and ought to be removed.
- [x] openPMD does not allow to create datasets that have zero extent in any dimension. This is what happens for example in the WarmCopper example. Fix that.
- [x] The existing HDF5 and ADIOS plugins use different strategies for data preparation. This plugin currently uses the strategy applied in the ADIOS plugin, but both should be supported in the end.
- [x] Further testing, especially (1) reading (e.g. restarting) and (2) using openPMD backends other than ADIOS(1|2), i.e. HDF5.
- [x] Either wait for [this PR](https://github.com/openPMD/openPMD-api/pull/529) in the openPMD API or revert commit [51d55c9](https://github.com/ComputationalRadiationPhysics/picongpu/commit/51d55c9e5db45c38f9c56602208e3c5000defbc8).
- [x] The plugin needs [this commit](https://github.com/openPMD/openPMD-api/commit/d7b6dbd3acdc2e81097b39041f2351005a2b785f) ([PR #511](https://github.com/openPMD/openPMD-api/pull/511)) to compile without error. This should be considered before pulling.

For testing purposes, please use [this branch of the openPMD API](https://github.com/franzpoeschel/openPMD-api/tree/topic-advanced-backend-configuration).
- [x] Wait for [JSON configuration](https://github.com/openPMD/openPMD-api/pull/569) to land in openPMD API.
- [x] Documentation

As [noted below](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2966#issuecomment-501717152), the HDF5 backend in the openPMD API is currently broken and may not be used in this plugin at the moment.
- [x] Investigate a workaround for that. EDIT: HDF5 backend has been patched and can now be used from within openPMD plugin.

[...]

PR #3122, opened 2019-11-27
Reduce PML memory consumption

sbastrakov:
By internally storing only the required near-boundary values instead of the whole local domain as for normal E and B. Along the way did small style improvements.

TODO:

- [x] fix 2D case
- [x] test checkpointing
- [x] refactor and document
- [x] update the memory calculator

### DIFF
--- a/include/picongpu/fields/MaxwellSolver/YeePML/Field.hpp
+++ b/include/picongpu/fields/MaxwellSolver/YeePML/Field.hpp
@@ -22,16 +22,18 @@
 
 #include "picongpu/simulation_defines.hpp"
 #include "picongpu/fields/Fields.def"
+#include "picongpu/fields/MaxwellSolver/YeePML/Parameters.hpp"
 #include "picongpu/fields/cellType/Yee.hpp"
 #include "picongpu/traits/FieldPosition.hpp"
 
 #include <pmacc/dataManagement/ISimulationData.hpp>
 #include <pmacc/fields/SimulationFieldHelper.hpp>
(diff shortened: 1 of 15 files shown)


.build/CMakeFiles/3.12.3/CMakeSystem.cmake
```diff
@@ -0,0 +1,15 @@
+set(CMAKE_HOST_SYSTEM "Windows-10.0.14393")
```
ax3l: feel free to add the `.build` directory to `.gitignore` if that's a thing that VS usually has in place.

sbastrakov: Oops, thx for pointing out.

sbastrakov: That was actually not VS, just me mimicking the usual directory structure.

sbastrakov:
Ready for review.
Note: the workarounds for the HDF5 and ADIOS output plugins are mostly similar, could have been abstracted better, and implemented with less duplication. After offline discussions with @psychocoderHPC we decided not to do it, and just implement patches so that it all works for now. The reason is that hopefully these 2 plugins will be rather soon replaced with a new openPMD plugin developed in #2966 (which perhaps needs to be extended to support non-domain-bound fileds, to be discussed there).

sbastrakov:
Sorry for accidentally introducing other commits, I did rebase inaccurately, now should be fixed.

PR #3126, opened 2019-11-28
Fix cmake flags for MSVS

sbastrakov:
The `/bigobj` flag is required for almost all PIConGPU setups, so reasonable to set by default. The other flag is needed for Visual Studio versions after 15.8, I put an explanatory link. Just discovered it recently after updating my VS.

I also now have one more weird error in the cmake-generated VS project (easy to fix manually). It seems to link to the correct Boost program_options but somehow also adds a dependency to the same library but without the path to it. But not yet sure where it comes from, maybe some issues on my system. Adding the directory to the search path helps.

### DIFF
--- a/include/picongpu/CMakeLists.txt
+++ b/include/picongpu/CMakeLists.txt
@@ -282,6 +282,14 @@ elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel")
 # PGI
 elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "PGI")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Minform=inform")
+# MSVS
+elseif(MSVC)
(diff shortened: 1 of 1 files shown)


ax3l:
Makes sense, virtually no project builds with MSVC without bigobj and its only not default because of potential breakage.

Is it possible your boost library is too new for the used CMake? CMake maintains an internal list of boost libs and if your boost library is newer than what CMake knew on release, we can only guess on a best-effort basis to add dependencies inside boost properly.

sbastrakov:
@ax3l no, looks like I get some weird mismatch of Boost static library names, with and without `lib` in front and I can't make it work seamlessly. What ends up happening is my Visual Studio project tries to link both versions of the same libraries (and I don't understand why, I did not add this or modify this part of cmake) and one is by full path due to CMake and it's fine, but the other somehow had only the library name. So I need to manually add the Boost directory to the additional library search path.

ax3l:
Ok good to know, thanks for the details.

PR #3133, opened 2019-12-12
fix loading picongpu profile

psychocoderHPC:
It is not possible to source the profiles for the PIConGPU environment
if the file `$PICSRC/bin/picongpu-completion.bash` is not available.
All tpl files are loding the profile inside the jobscript. If the
completion file is not available the simulation will not start.

bug was introduced with #3069

### DIFF
--- a/etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
+++ b/etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
@@ -82,4 +82,9 @@ function getDevice() {
 }
 
 # Load autocompletion for PIConGPU commands
-source $PICSRC/bin/picongpu-completion.bash
+BASH_COMP_FILE=$PICSRC/bin/picongpu-completion.bash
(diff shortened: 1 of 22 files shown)


etc/picongpu/hemera-hzdr/k80_picongpu.profile.example
```diff
@@ -82,4 +82,9 @@ function getDevice() {
 }
 
 # Load autocompletion for PIConGPU commands
-source $PICSRC/bin/picongpu-completion.bash
+BASH_COMP_FILE=$PICSRC/bin/picongpu-completion.bash
+if [ -f $BASH_COMP_FILE ] ; then
+    source $PICSRC/bin/picongpu-completion.bash
+else
+    echo "bash completion file '$PICSRC/bin/picongpu-completion.bash' not found." >&2
+fi
```
PrometheusPi: please add a new line

PR #3143, opened 2020-01-13
EnergyHistogram: allow non-keyword arguments

### LINKED ISSUE #3140: EnergyHistogramData broken?
The example here
https://picongpu.readthedocs.io/en/latest/usage/plugins/energyHistogram.html#data-reader

```py
from picongpu.plugins.data import EnergyHistogramData

eh_data = EnergyHistogramData('/path/to/your/sim/eg/this/one/lwfa_001')
# load data for a given iteration
counts, bins_keV = eh_data.get('e', species_filter='all', iteration=2000)
```
seams to be broken and results in
```
TypeError: get() takes 1 positional argument but 2 were given
```

Changing to
```py
counts, bins_keV = eh_data.get(species='e', species_filter='all', iteration=2000)
```
does not work either and results in
```
ValueError: too many values to unpack (expected 2)
```
because this also returns into the 
```py
d, bins, iteration, dt = ...
```
signature now.

codingS3b:
Fixes #3140 by updating the example for the `EnergyHistogramData` in the documentation.
Now the correct number of return values for the `.get` method are provided and all arguments are provided as keyword arguments.
However, providing arguments should now also work since the base class function signature got updated as well.

### DIFF
--- a/docs/source/usage/plugins/energyHistogram.rst
+++ b/docs/source/usage/plugins/energyHistogram.rst
@@ -107,12 +107,10 @@ You can quickly load and interact with the data in Python with:
    counts, bins_keV, _, _ = eh_data.get(species='e', species_filter='all', iteration=2000)
 
    # get data for multiple iterations
-   d, bins, iteration, dt = eh_data.get(species='e', iteration=[200, 400, 8000])
+   counts, bins_keV, iteration, dt = eh_data.get(species='e', iteration=[200, 400, 8000])
 
    # load data for a given time
    counts, bins_keV, iteration, dt = eh_data.get(species='e', species_filter='all', time=1.3900e-14)
-
-
 
 
 Matplotlib Visualizer
--- a/docs/source/usage/plugins/energyHistogram.rst
+++ b/docs/source/usage/plugins/energyHistogram.rst
@@ -104,13 +104,15 @@ You can quickly load and interact with the data in Python with:
    eh_data.get_times(species='e')
 
    # load data for a given iteration
-   counts, bins_keV = eh_data.get('e', species_filter='all', iteration=2000)
-
-   # load data for a given time
-   counts, bins_keV = eh_data.get('e', species_filter='all', time=1.3900e-14)
+   counts, bins_keV, _, _ = eh_data.get(species='e', species_filter='all', iteration=2000)
 
    # get data for multiple iterations
    d, bins, iteration, dt = eh_data.get(species='e', iteration=[200, 400, 8000])
+
+   # load data for a given time
+   counts, bins_keV, iteration, dt = eh_data.get(species='e', species_filter='all', time=1.3900e-14)
+
+
 
 
 Matplotlib Visualizer
--- a/lib/python/picongpu/plugins/data/base_reader.py
+++ b/lib/python/picongpu/plugins/data/base_reader.py
@@ -74,6 +74,10 @@ class DataReader(object):
         If both are given, the 'time' argument is converted to
         an iteration and data for the iteration matching the time
         is returned.
+        For other valid args and kwargs, please look at the
+        documentation of the '_get_for_iteration' methods
+        of the derived classes since the parameters are passed
+        on to that function.
 
         time: float or np.array of float or None.
             If None, data for all available times is returned.
--- a/lib/python/picongpu/plugins/data/base_reader.py
+++ b/lib/python/picongpu/plugins/data/base_reader.py
@@ -38,7 +38,7 @@ class DataReader(object):
         """
         return self.find_time.get_dt()
 
-    def get_times(self, **kwargs):
+    def get_times(self, *args, **kwargs):
         """
         Returns
         -------
@@ -46,10 +46,10 @@ class DataReader(object):
         data is available
         """
 
-        iterations = np.array(self.get_iterations(**kwargs))
+        iterations = np.array(self.get_iterations(*args, **kwargs))
         return self.find_time.get_time(iterations)
 
-    def get_data_path(self, **kwargs):
+    def get_data_path(self, *args, **kwargs):
         """
         Returns
         -------
@@ -57,7 +57,7 @@ class DataReader(object):
         """
         raise NotImplementedError
 
-    def get_iterations(self, **kwargs):
+    def get_iterations(self, *args, **kwargs):
         """
         Returns
         -------
@@ -66,7 +66,7 @@ class DataReader(object):
         """
         raise NotImplementedError
 
-    def get(self, **kwargs):
+    def get(self, *args, **kwargs):
         """
         Parameters
         ----------
@@ -103,15 +103,15 @@ class DataReader(object):
             time = kwargs.pop('time')
             if time is None:
                 # use all times that are available, i.e. all iterations
-                iteration = self.get_iterations(**kwargs)
+                iteration = self.get_iterations(*args, **kwargs)
             else:
                 iteration = self.find_time.get_iteration(
                     time, method='closest')
             # print("got 'time'=", time, ", converted to iter", iteration)
 
-        return self._get_for_iteration(iteration, **kwargs)
+        return self._get_for_iteration(iteration, *args, **kwargs)
 
-    def _get_for_iteration(self, iteration, **kwargs):
+    def _get_for_iteration(self, iteration, *args, **kwargs):
         """
         Get the data for a given iteration.
 


docs/source/usage/plugins/energyHistogram.rst
```diff
@@ -104,14 +104,16 @@ You can quickly load and interact with the data in Python with:
...
 
    # load data for a given iteration
-   counts, bins_keV = eh_data.get('e', species_filter='all', iteration=2000)
-
-   # load data for a given time
-   counts, bins_keV = eh_data.get('e', species_filter='all', time=1.3900e-14)
+   counts, bins_keV, _, _ = eh_data.get(species='e', species_filter='all', iteration=2000)
 
    # get data for multiple iterations
    d, bins, iteration, dt = eh_data.get(species='e', iteration=[200, 400, 8000])
```
ax3l: For consistency?
```suggestion
   count, bins, iteration, dt = eh_data.get(species='e', iteration=[200, 400, 8000])
```

docs/source/usage/plugins/energyHistogram.rst
```diff
@@ -104,14 +104,16 @@ You can quickly load and interact with the data in Python with:
...
 
+   # load data for a given time
+   counts, bins_keV, iteration, dt = eh_data.get(species='e', species_filter='all', time=1.3900e-14)
+
```
ax3l: maybe remove those extra newlines?

lib/python/picongpu/plugins/data/base_reader.py
```diff
@@ -66,7 +66,7 @@ def get_iterations(self, **kwargs):
         """
         raise NotImplementedError
 
-    def get(self, **kwargs):
+    def get(self, *args, **kwargs):
         """
         Parameters
```
ax3l: We should document `species` and `species_filter` here as well or point to the according method that this forwards to. Otherwise users cannot find their signature when checking `.get?`

codingS3b: This might be tricky since this function passes the `*args` and `**kwargs` on to the `_get_for_iteration` method of the derived classes which do not necessarily have common arguments that could be mentioned here. 
Maybe python has some tricks that allow displaying the docstring of `eh_data._get_for_iteration` when calling `help(eh_data.get)` instead of the docstring for `BaseReader.get` but I am not familiar with any. Do you happen to know?

PrometheusPi: @ax3l Are you fine with that?

ax3l: @codingS3b I think you should be able to "monkeypatch" (actual term) the `.__doc__` method in your derived class for this purpose.

codingS3b: After playing around a little bit, that is indeed possible. I tried it using the following example:
```py
def modify_doc(doc):
    def f_with_doc(f):
        f.__doc__ = doc
        return f
    return f_with_doc


class Base(object):
    def _get(self, *args, **kwargs):
        """
        Actual implementation, not given in base class
        """
        raise NotImplementedError
    
    def get(self, *args, **kwargs):
        """
        Not useful docstring, refer to derived _get
        """
        # NOTE: when calling help on that function, we should display help for _get maybe?
        print("Base: common stuff all derived classes need")
        
        #self.get.__doc__ == self._get.__doc__
        #self.get.__func__.__doc__ = self._get.__doc__
        return self._get(*args, **kwargs)
    
class Derived(Base):
    def __init__(self):
        super().__init__()
    def _get(self, a, b=2):
        """
        Documentation of the derived class _get
        Parameters
        ----------
        a: int
            an integer
        b: int
            a second integer
        """
        
        print("Derived: a=", a, "b=", b)
    
    @modify_doc(Derived._get.__doc__)
    def get(self, *args, **kwargs):
        return super().get(*args, **kwargs)

a = Base()
help(a.get)
# Help on method get in module __main__:

# get(*args, **kwargs) method of __main__.Base instance
#    Not useful docstring, refer to derived _get

d = Derived()
help(d.get)
#Help on method get in module __main__:
#
#get(*args, **kwargs) method of __main__.Derived instance
#    Documentation of the derived class _get
#    Parameters
#    ----------
#    a: int
#        an integer
#    b: int
#        a second integer
```
However, this has the downside that we have to add the decorated `get` function to each derived class, only for the purpose of documentation, without adding functionality and simply calling the base class function. Maybe just adding the remark to have a look at the derived classes function signature is simpler...

ax3l: Cool I think you can avoid implementing a `.get` in each derived class by [monkey-patching](https://filippo.io/instance-monkey-patching-in-python/).
Not sure if that makes your code shorter as it already is, though.

codingS3b: Thanks for sharing the link. I'm not sure I get you right.
Staying with the above example, do you mean something like 
```python
d.get.__doc__ = "new documentation"
# throws AttributeError: attribute '__doc__' of 'method' objects is not writable

# however this seems to work
d.__class__.get.__doc__ = "new documentation
```
? When should this monkey patching take place then? During object construction or already at the time of class definition?

ax3l: About the when I have no good enough experience with, preferably the earlier the better. Anyway, at object construction time sounds still good enough since a user will ask the member function of an already instantiated object for its doc string.

codingS3b: I added an `_update_docstrings` function in the constructor of the base reader class that replaces the docstring of the `get` function with the one from `_get_for_iteration` function of the derived class. The minor problem is now that this replacement hides the option to pass a `time=` argument to the `get` function since the `_get_for_iteration` only accepts `iteration` as argument and the base class method takes care of converting time to iteration.

lib/python/picongpu/plugins/data/base_reader.py
```diff
@@ -32,32 +32,41 @@ def __init__(self, run_directory):
...
         self.data_file_suffix = None
 
+        self._update_docstrings()
+
+    def _update_docstrings(self):
+        # only for the derived classes
+        if self.__class__ == DataReader:
+            return
+
+        self.__class__.get.__doc__ = self._get_for_iteration.__doc__
```
ax3l: Should this append or overwrite?
Otherwise LGTM.

PrometheusPi: @ax3l good question - @codingS3b what do you want: overwrite or append?

codingS3b: The current implementation fully replaces the docstring, however, the function signature is unchanged. 
docstring_replace

The docstring of the base class is this
docstring_replace_baseclass

So, I would rely on your judgement, whether overwriting or appending would be better.

codingS3b: I just discovered that the docstring update has a negative side-effect and probably should not be used and would need modification:
the `self.__class__.get` is a reference to the base classes `get` method and alters the docstring for the base class as well and not only for the derived class for which we want that :(

lib/python/picongpu/plugins/data/base_reader.py
PrometheusPi: @codingS3b Did you solve this issue?

lib/python/picongpu/plugins/data/base_reader.py
PrometheusPi: Okay - this is definitely something one should avoid.  WHy do you have to use `self.__class__` anyway? The examples in [the monkey-patch link](https://filippo.io/instance-monkey-patching-in-python/) provided by @ax3l only apply the changes to an instance of the class, not to the class itself?

codingS3b: Yes, but as I tried before, the doc attribute is not writeable and can't be patched, that is why I used the class reference.

PrometheusPi: Okay - that's problematic. @codingS3b Do you agree with me that overwriting the documentation of the base class is not an option?

codingS3b: Yeah, I would also rather avoid it if possible. However, I don't have a solution available right now :(

PrometheusPi: I will thing about a solution as well. 

cc'ing @BrianMarre and @wmles (they might have some ideas as well)

wmles: the cleanest thing is to inherit the method and give it a new docstring:
```
def get(*args, **kwargs):
    """ new docstring """
    return super().get(*args, **kwargs)

this is probably slightly more expensive than not inheriting, but since you are using python, I assume this is no issue.

And I assume you can also overwrite the method for the instances inside their __init__ - but does not have any advantage...

psychocoderHPC:

PrometheusPi:
@codingS3b and @ax3l any ideas so far on how to resolve this `__doc__` issue?

codingS3b:
@PrometheusPi: as discussed on the phone today, I tried appending the docstring. This however has the same problem as the previous idea since the baseclass docstring is still changed.
So I would just revert the last commit. Is that okay with you?

PrometheusPi:
@codingS3b That is fine for now (the 0.5.0) release. I will open an issue on the matter for foxing the doc string later.

PR #3146, opened 2020-01-15
Unify output text names of vector components

sbastrakov:
The previous implementation had two issues:

- for > 4 components it led to out-of-bounds access when iterating by component index
- the logic was copied in multiple places

Create a utility function for text names of vector components. Change hardcoded names to use the new function. The new way guarantees unification and avoids potential out-of-bounds access by index.

~~TODO: do more checks with PML output and checkpointing.~~ (done)

### DIFF
--- a/include/picongpu/plugins/misc/ComponentNames.cpp
+++ b/include/picongpu/plugins/misc/ComponentNames.cpp
@@ -1,0 +1,64 @@
+/* Copyright 2020 Sergei Bastrakov
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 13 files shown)


sbastrakov:
Finished my tests, ready for review!

include/picongpu/plugins/misc/ComponentNames.cpp
```diff
@@ -0,0 +1,58 @@
...
+{
+
+    std::vector< std::string > getComponentNames(
+        uint32_t const numComponents
+    )
+    {
+        auto result = std::vector< std::string >( numComponents );
+        for( auto i = 0u; i < result.size(); i++ )
+            result[ i ] = "component" + std::to_string( i );
+        if( result.size() <= 4 )
```
psychocoderHPC: please use `if .. elseif else` and add a comment that the case <=4 is for backwards compatibility. Please use early returns to avoid the std::copy.

sbastrakov: Good point, fixed and rebased.

sbastrakov:
All fixed, ready for another review.

PR #3149, opened 2020-01-21
Jupyter widgets output capture

codingS3b:
This PR adds the optional keyword argument `output_widget` to the jupyter widgets to provide an instance of `widgets.Output` for capturing relevant output of the member functions.
This has the advantage, that we now get control over where the output is written within the notebook and can clear messages without getting an ever increasing stack of lines below the cell of the widget.

The default valufe for the argument is `None` in which case a new `widgets.Output` instance is created and made part of the widgets list of children.

If a reference to an existing output widget is provided, it is the callers responsibility to display the output widget somewhere in the notebook and it is not made part of our widgets children.


codingS3b:
@PrometheusPi  are there any comments on this PR or can we merge it?

PrometheusPi:
@codingS3b Sorry - I forgot about this pull request, I will have a look at it as soon as possible.

codingS3b:
Yes, this was tested in a local notebook

PR #3152, opened 2020-01-27
fix taurus profile

psychocoderHPC:
Number of tasks started per node was wrong.

- use node exclusive, else scheduling is wrong
- add function `getNode` and `getDevice`

### DIFF
--- a/etc/picongpu/taurus-tud/V100.tpl
+++ b/etc/picongpu/taurus-tud/V100.tpl
@@ -27,10 +27,16 @@
 #SBATCH --job-name=!TBG_jobName
 #SBATCH --nodes=!TBG_nodes
 #SBATCH --ntasks=!TBG_tasks
+#SBATCH --ntasks-per-node=!TBG_gpusPerNode
 #SBATCH --mincpus=!TBG_mpiTasksPerNode
 #SBATCH --cpus-per-task=!TBG_coresPerGPU
 #SBATCH --mem=0
 #SBATCH --gres=gpu:!TBG_gpusPerNode
+#SBATCH --exclusive
+
+# disable hyperthreading (default on taurus)
+#SBATCH --hint=nomultithread
+
 #SBATCH --mail-type=!TBG_mailSettings
 #SBATCH --mail-user=!TBG_mailAddress
 #SBATCH --workdir=!TBG_dstPath
@@ -58,9 +64,6 @@
 # number of CPU cores to block per GPU
 # we got 7 CPU cores per GPU (44cores/6gpus ~ 7cores)
 .TBG_coresPerGPU=7
-# this does not yet include hyperthreading
-#SBATCH --hint=nomultithread
-# (this is currently also the default setting on taurus)
 
 # We only start 1 MPI task per GPU
 .TBG_mpiTasksPerNode="$(( TBG_gpusPerNode * 1 ))"
--- a/etc/picongpu/taurus-tud/V100_picongpu.profile.example
+++ b/etc/picongpu/taurus-tud/V100_picongpu.profile.example
@@ -76,7 +76,34 @@ export CXXFLAGS="-Dlinux"
 export TBG_SUBMIT="sbatch"
 export TBG_TPLFILE="etc/picongpu/taurus-tud/V100.tpl"
 
-alias getNode='srun -p ml --gres=gpu:6 -n 1 --mem=0 --cpus-per-task=44 --pty -t 2:00:00 bash'
+# allocate an interactive shell for two hours
+#   getNode 2  # allocates 2 interactive nodes (default: 1)
+function getNode() {
+    if [ -z "$1" ] ; then
+        numNodes=1
+    else
+        numNodes=$1
+    fi
+    export OMP_NUM_THREADS=7
+    srun --time=2:00:00 --nodes=$numNodes --ntasks=$((6 * $numNodes)) --ntasks-per-node=6 --cpus-per-task=7 --mem=0 --exclusive --gres=gpu:6 -p ml --pty bash
+}
+
+# allocate an interactive shell for two hours
+#   getDevice 2  # allocates 2 interactive devices on one node (default: 1)
+function getDevice() {
+    if [ -z "$1" ] ; then
+        numDevices=1
+    else
+        if [ "$1" -gt 6 ] ; then
+            echo "The maximal number of devices per node is 6." 1>&2
+            return 1
+        else
+            numDevices=$1
+        fi
+    fi
+    export OMP_NUM_THREADS=7
+    srun --time=2:00:00 --nodes=1 --ntasks=$numDevices --ntasks-per-node=$(($numDevices)) --cpus-per-task=7 --mem=$((254000 / $numDevices)) --gres=gpu:$numDevices -p ml --pty bash
+}
 
 # Load autocompletion for PIConGPU commands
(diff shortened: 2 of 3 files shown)


etc/picongpu/taurus-tud/V100.tpl
```diff
@@ -31,6 +31,8 @@
 #SBATCH --cpus-per-task=!TBG_coresPerGPU
 #SBATCH --mem=0
 #SBATCH --gres=gpu:!TBG_gpusPerNode
+#SBATCH --exclusive
```
steindev: I tried that half a year ago and made a bad experience. The scheduler was not able to distribute tasks to nodes correctly.

From my own tests it looks like `--cpus-per-task` produces the wrong task distribution. Adding the `--ntasks-per-node=6`, removing `--cpus-per-task`, and setting `mem=254000` option to the submission script results in correct distribution and should furthermore ensure we allocate a node exclusively. 

But I admit that there are actually two cores (which can be assigned 8 threads) are left over in the node. However, it is not possible to ask for a job with zero memory. So we should indeed use the node exclusively.

I submitted a job using 16 gpus out of 18 available with a total of three nodes in order to see whether my new configuration still works.

etc/picongpu/taurus-tud/V100_picongpu.profile.example
```diff
@@ -76,7 +76,34 @@ export CXXFLAGS="-Dlinux"
...
+# allocate an interactive shell for two hours
+#   getNode 2  # allocates 2 interactive nodes (default: 1)
+function getNode() {
+    if [ -z "$1" ] ; then
+        numNodes=1
+    else
+        numNodes=$1
+    fi
+    export OMP_NUM_THREADS=7
+    srun --time=2:00:00 --nodes=$numNodes --ntasks=$((6 * $numNodes)) --ntasks-per-node=7 --cpus-per-task=7 --mem=0 --exclusive --gres=gpu:6 -p ml --pty bash
```
steindev: Why do you choose `OMP_NUM_THREADS=7` and `--ntasks-per-node=7`? I think `--ntasks-per-node=6` should be used for 6 gpus / node.

etc/picongpu/taurus-tud/V100_picongpu.profile.example
```diff
@@ -76,7 +76,34 @@ export CXXFLAGS="-Dlinux"
...
+        numDevices=1
+    else
+        if [ "$1" -gt 6 ] ; then
+            echo "The maximal number of devices per node is 6." 1>&2
+            return 1
+        else
+            numDevices=$1
+        fi
+    fi
+    export OMP_NUM_THREADS=7
```
steindev: As above: Why `OMP_NUM_THREADS=7`?

etc/picongpu/taurus-tud/V100_picongpu.profile.example
sbastrakov: I think `OMP_NUM_THREADS=7` is because there are 44 cores per 6 GPUs so 7 is the integer part of 44 / 6. However ntasks should probably 6, not 7, indeed.

steindev:
Ok, the test on 16 gpus ran succesfully. On two nodes six tasks where started and one started four tasks. Although I got the error message
```
srun: Warning: can't honor --ntasks-per-node set to 6 which doesn't match the requested tasks 16 with the number of requested nodes 3. Ignoring --ntasks-per-node.
```

The `SBATCH` commands in the `submit.start` where
```
#SBATCH --partition=ml
#SBATCH --time=00:30:00
# Sets batch job's name
#SBATCH --job-name=001_16gpu
#SBATCH --nodes=3
#SBATCH --ntasks=16
#SBATCH --ntasks-per-node=6
#SBATCH --mem=254000
#SBATCH --gres=gpu:6
# ... (mail settings etc.) ...
```
Note, I did not use the `#SBATCH --hint=nomultithread` which is currently configured in the `V100.tpl`

Should I just ask ZIH support what the correct job configuration should look like?

etc/picongpu/taurus-tud/V100_picongpu.profile.example
PrometheusPi: Yes `--ntasks-per-node` should be `6`. See [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/fad264596fa8410c6bb1a2ac2d5d627f43153859/etc/picongpu/taurus-tud/V100.tpl#L56)

etc/picongpu/taurus-tud/V100_picongpu.profile.example
PrometheusPi: This is on OpenMP specification that sets the number of span-able threads of each MPI rank to 7 (see [here](https://gcc.gnu.org/onlinedocs/libgomp/omp_005fget_005fnum_005fthreads.html) - which agrees with the fact that there are 7 cores per MPI rank available (plus a few extra)).

PrometheusPi:
@steindev `#SBATCH --hint=nomultithread` does not need to be used - it is the default (have a look at #3071). 
The warning is fine. It follows that rule for 2 of the 3 nodes allocated but can not fulfill it for the 3rd, since `16 % 6 = 4`- the last node only used 4 MPI ranks (called tasks) instead of the requested 6.

etc/picongpu/taurus-tud/V100.tpl
PrometheusPi: @steindev Did `--exclusive` work for you? I do not get a warning on submission at least - but the cluster is too full - I can not test it.

etc/picongpu/taurus-tud/V100_picongpu.profile.example
PrometheusPi: here you could add `--exclusive` as well

psychocoderHPC: Yes will fix it

etc/picongpu/taurus-tud/V100.tpl
psychocoderHPC: I tested exclusive. IMO it is also better to use this flag because there are only a few MB (256000MiB) main memory available and when we share the node with other it can result in crashes.

etc/picongpu/taurus-tud/V100_picongpu.profile.example
psychocoderHPC: Yes because we do not like to share the node

psychocoderHPC:
@steindev @PrometheusPi @sbastrakov 
I moved the option `#SBATCH --hint=nomultithread` up, they was hidden before.
I fixed `--ntasks-per-node` and kept `--exclusive` to avoid side effect with other jobs. A few idling gpus are better than non debug able states due to side effect with other jobs.

etc/picongpu/taurus-tud/V100.tpl
steindev: Don't you want to add `--ntasks-per-node=6`?

etc/picongpu/taurus-tud/V100_picongpu.profile.example
PrometheusPi: sorry

PrometheusPi:
@steindev Are you fine with merging or do you remaining open questions?

steindev:
As far as I understand, the update has been tested and is working. Let's merge.

Issue #3160, opened 2020-02-10
omega_p * dt <= 0.1  condition for solid target ion acceleration


IoannisTazes:
Hi all!
I am trying to simulate proton acceleration from the interaction of an intense 25fs FWHM gaussian pulse (a0 = 8.5), 800nm main wavelength, with a 13micron 10 times ionized Al target with long scale length plasma expanding from the front and the rear of the target. the electron density is 350nc, and my grid is: 
```
/
** equals X
         *  unit: meter */
        constexpr float_64 CELL_WIDTH_SI = 22e-6/1792;
        /** equals Y - the laser & moving window propagation direction
         *  unit: meter */
        constexpr float_64 CELL_HEIGHT_SI = 60e-6/6144;
```
in the stdout file i get this:
```
PIConGPUVerbose PHYSICS(1) | Courant c*dt <= 1.51771 ? 1
PIConGPUVerbose PHYSICS(1) | species e: omega_p * dt <= 0.1 ? 1.31958
PIConGPUVerbose PHYSICS(1) | species H: omega_p * dt <= 0.1 ? 0.0307951
PIConGPUVerbose PHYSICS(1) | species Al: omega_p * dt <= 0.1 ? 0.0592854
PIConGPUVerbose PHYSICS(1) | y-cells per wavelength: 51.4286
PIConGPUVerbose PHYSICS(1) | macro particles per device: 276480000
PIConGPUVerbose PHYSICS(1) | typical macro particle weighting: 1.89679e+06
```
All conditions seems to be fine, except 

> e: omega_p * dt <= 0.1 ? 1.31958

which in order to resolve i think i will need at lest 10 time higher resolution like:

> constexpr float_64 CELL_WIDTH_SI = 22e-6/17920;

> constexpr float_64 CELL_HEIGHT_SI = 60e-6/61440;

that i dont think i will be able to do. Is this going to highly affect my simulation results or im i good to go? I also observed in the simulations with the resolution i used, that up to 200fs the protons reveal reasonable energy spectrum but beyond that they seem keep accelerating to very high energies.
Thank you in advance!

sbastrakov:
@PrometheusPi @n01r @steindev could you comment?

n01r:
Hi @IoannisTazes, welcome and thanks for your question!

Could you tell us a few more things about the system you are running on and the maximum memory available to you? 
Are you trying to run in 2D or 3D? (The macro particle weighting looks a bit high, too.)
[...]

PR #3162, opened 2020-02-10
Fix a typo in density ratio value of the KHI example

sbastrakov:
While looking into [another potential bug](https://github.com/ComputationalRadiationPhysics/picongpu/issues/3160#issuecomment-584186175), accidentally found out this typo (or a copy-paste error perhaps). Ideally, we would perhaps want to check non-negativity of the resulting densities automatically, however this is probably problematic, same as #3109.

### DIFF
--- a/share/picongpu/examples/KelvinHelmholtz/include/picongpu/param/speciesDefinition.param
+++ b/share/picongpu/examples/KelvinHelmholtz/include/picongpu/param/speciesDefinition.param
@@ -83,7 +83,7 @@ value_identifier( float_X, MassRatioIons, 1836.152672 );
 value_identifier( float_X, ChargeRatioIons, -1.0 );
 
 /* ratio relative to BASE_DENSITY */
-value_identifier( float_X, DensityRatioIons, -1.0 );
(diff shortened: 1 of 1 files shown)


sbastrakov:
@PrometheusPi strange indeed, could it be that it generated 0 ions due to "negative" density? I'll try to understand how did it work.

sbastrakov:
Ah, perhaps ions were not sampled, but [derived](https://github.com/ComputationalRadiationPhysics/picongpu/blob/8c124dc85030c6a82277be60b1f1ae34913337bc/share/picongpu/examples/KelvinHelmholtz/include/picongpu/param/speciesInitialization.param#L48-L51) from electrons and so ion density was never used, also because the omega_p output mistakenly did not include it?

PrometheusPi:
@sbastrakov Could be - that would have saved us - the KHI does not work if it is not initally  charge neutral.
(The code is to old to blame 😄 )

sbastrakov:
Ready to be merged.

ax3l:
`DensityRatio` is only used by some of the methods in `speciesInitialization`.

PR #3163, opened 2020-02-10
Fix calculation of omega_p for logging

sbastrakov:
The previous version used `BASE_DENSITY` for all species. Replace it with a proper density of the species.

Thanks @IoannisTazes for drawing our attention towards it and @n01r for [spotting the issue](https://github.com/ComputationalRadiationPhysics/picongpu/issues/3160#issuecomment-584186175).

### DIFF
--- a/include/picongpu/initialization/InitialiserController.hpp
+++ b/include/picongpu/initialization/InitialiserController.hpp
@@ -28,6 +28,7 @@
 #include "picongpu/fields/FieldE.hpp"
 #include "picongpu/fields/FieldB.hpp"
 #include "picongpu/fields/laserProfiles/profiles.hpp"
+#include "picongpu/particles/traits/GetDensityRatio.hpp"
 
 #include "picongpu/initialization/SimStartInitialiser.hpp"
 
@@ -98,9 +99,9 @@ public:
         log<picLog::SIMULATION_STATE > ("Loading from persistent data finished");
     }
 
-    /** log omega_p for each species
+    /** Log omega_p for each species
      *
-     * calculate omega_p for each given species and create a `picLog::PHYSICS`
+     * Calculate omega_p for each given species and create a `picLog::PHYSICS`
      * log message
      */
     template<typename T_Species = bmpl::_1>
@@ -108,11 +109,18 @@ public:
     {
         void operator()()
         {
+            /* The omega_p calculation is based on species' densityRatio
+             * relative to the BASE_DENSITY. Thus, it is only accurate
+             * for species with macroparticles sampled by density,
+             * but not necessarily for derived ones.
+             */
             using FrameType = typename T_Species::FrameType;
             const float_32 charge = frame::getCharge<FrameType>();
             const float_32 mass = frame::getMass<FrameType>();
+            const auto densityRatio = traits::GetDensityRatio< T_Species >::type::getValue( );
+            const auto density = BASE_DENSITY * densityRatio;
             log<picLog::PHYSICS >("species %2%: omega_p * dt <= 0.1 ? %1%") %
-                                 (sqrt(BASE_DENSITY *  charge / mass * charge / EPS0) * DELTA_T) %
+                                 (sqrt(density * charge / mass * charge / EPS0) * DELTA_T) %
                                   FrameType::getName();
         }
     };
@@ -137,6 +145,10 @@ public:
                 chargeRatio<>
             >::type;
             meta::ForEach< SpeciesWithMassCharge, LogOmegaP<> > logOmegaP;
+            log<picLog::PHYSICS >("Resolving plasma oscillations?\n"
+                "   Estimates are based on DensityRatio to BASE_DENSITY of each species\n"
+                "   (see: density.param, speciesDefinition.param).\n"
+                "   It and does not cover other forms of initialization");
             logOmegaP();
 
             if (fields::laserProfiles::Selected::INIT_TIME > float_X(0.0))


include/picongpu/initialization/InitialiserController.hpp
```diff
@@ -111,8 +112,10 @@ class InitialiserController : public IInitPlugin
             using FrameType = typename T_Species::FrameType;
             const float_32 charge = frame::getCharge<FrameType>();
             const float_32 mass = frame::getMass<FrameType>();
+            const auto densityRatio = traits::GetDensityRatio< T_Species >::type::getValue( );
+            const auto density = BASE_DENSITY * densityRatio;
```
PrometheusPi: What happens if one derives the ion species from an electron species or a density profile? Is this value always applied?

sbastrakov: Also discussed offline, I will check tomorrow.

ax3l: Just a note: we always kepts this as a conservative base-density approach since one can init particles in way more ways than their densityRatio.
For example by cloning and multiplication by proton number, among others in `speciesInitialization.param` - that's why we kept the logic in here simple as a first estimate.

You can add this update, but it's not necessarily "closer"/"more correct" than estimates via the base density - and can be off in some user-given scenarios, too.

sbastrakov: Good to know. However, as far as I understand, the base density could be chosen completely arbitrarily? Which is not wise, but technically possible.

n01r: We should drop a line above these, explaining how the values were calculated.
@ax3l is right that there are still many ways to initialize the density that will not be represented by these quick calculations.
If we leave this in, then at least writing a disclaimer would be appropriate, in which we explain, that it is most important for electrons (the lightest species) which has the highest demand on resolution and that the calculation is a quick estimate based on density ratios that were given, and not by cloning or checkpoint-reading or anything else.

sbastrakov: Maybe better do it in readthedocs, not here?

n01r: Better both. Sure, we tell people to read the docs but a one-liner there does not hurt but can avoid long searches for mistakes.

sbastrakov: Add inline comments, will do readthedocs in a separate PR, since this part is not covered at all afaik.

n01r: I was actually thinking of something really **inline** in the user output, so that they don't get confused before they even look into the code that is producing these messages.
Something of the sort like this:
```
PIConGPUVerbose PHYSICS(1) | Courant c*dt <= 1.51771 ? 1
PIConGPUVerbose PHYSICS(1) | Resolving plasma oscillations?
PIConGPUVerbose PHYSICS(1) |   (Calculations based on speciesDefinition only,
PIConGPUVerbose PHYSICS(1) |   will not account for other forms of initialization) 
PIConGPUVerbose PHYSICS(1) | species e: omega_p * dt <= 0.1 ? 1.31958
PIConGPUVerbose PHYSICS(1) | species H: omega_p * dt <= 0.1 ? 0.0307951
PIConGPUVerbose PHYSICS(1) | species Al: omega_p * dt <= 0.1 ? 0.0592854
PIConGPUVerbose PHYSICS(1) | y-cells per wavelength: 51.4286
PIConGPUVerbose PHYSICS(1) | macro particles per device: 276480000
PIConGPUVerbose PHYSICS(1) | typical macro particle weighting: 1.89679e+06
```
It's not a one-liner but maybe we can still be creative and find one.
What do you think?

The same actually applies for the macro-particles per device, since this assumes that the device is *completely* filled with particles of the given number per cell.
But that could just be covered in the readthedocs.

psychocoderHPC:

include/picongpu/initialization/InitialiserController.hpp
ax3l: > Calculations based on speciesDefinition only
->
> Estimates based on DensityRatio to BASE_DENSITY of each species (see: density.param, speciesDefinition.param), ...

sbastrakov: Extended the output, now it looks like this:
```
...
PIConGPUVerbose PHYSICS(1) | Resolving plasma oscillations?
   Estimates are based on DensityRatio to BASE_DENSITY of each species
   (see: density.param, speciesDefinition.param).
   It and does not cover other forms of initialization
PIConGPUVerbose PHYSICS(1) | species e: omega_p * dt <= 0.1 ? 0.0247974
...
```

Made line breaks manually so that each line is < 80 characters.

I personally find it unreasonably detailed compared to other output, but probably that does not hurt anyone.

n01r: Yea, it does not hurt.
The only other way to keep it much shorter would be a general disclaimer of the form:

```
Estimates quantifying plasma resolution
=======================================
(for details, see <link/to/documentation)
```

To give a general header for the whole section of estimates in `output` and then linking to our documentation below seems like the shortest way to me, if you really don't want to put any explanation into `output`. But yea, it's a text file so it hurts no one ... most people don't even read it anymore after they've checked it once.

n01r:
Okay, I'm gonna merge - we can then extend the documentation also to `readthedocs` with the outcome of #3167.

PR #3172, opened 2020-02-19
Fix python requirements: remove sys and getopt

codingS3b:
Removes requirements `sys` and `getopt` that are already part of all python installations anyway.
At the moment, executing `pip install -r <pic_path>/lib/python/picongpu/requirements.txt` crashes due to not finding requirement files for the above mentioned packages.


psychocoderHPC:

PR #3180, opened 2020-02-24
add pillow as dependency

PrometheusPi:
This pull request adds [pillow](https://python-pillow.org/) as dependency before scipy because it conflicts otherwise with matplotlib - suggested by @codingS3b.

### DIFF
--- a/lib/python/picongpu/plugins/data/requirements.txt
+++ b/lib/python/picongpu/plugins/data/requirements.txt
@@ -1,4 +1,5 @@
 numpy
 pandas>=0.21.0
 h5py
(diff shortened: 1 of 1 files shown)


codingS3b:
This should work!

codingS3b:
Oops: I hit the approve button too early, sorry. This requirement should be put in the `plugins/data/requirements.txt` since the png reader utilizes `scipy.misc.imread` function which is only available in `scipy` if also `pillow` is installed.

PrometheusPi:
@codingS3b I adjusted the code as requested.

PR #3181, opened 2020-02-24
slurm update taurus: workdir to chdir

PrometheusPi:
This pull request fixes the working directory setting for taurus/zih/TU Dresden jobs, which became outdated end of last week due to a slurm update.

### DIFF
--- a/etc/picongpu/taurus-tud/V100.tpl
+++ b/etc/picongpu/taurus-tud/V100.tpl
@@ -39,7 +39,7 @@
 
 #SBATCH --mail-type=!TBG_mailSettings
 #SBATCH --mail-user=!TBG_mailAddress
-#SBATCH --workdir=!TBG_dstPath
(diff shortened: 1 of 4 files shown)


n01r:
**Note:** after the recent update at HZDR, the running the `0.4.0` master on `hemera` also suffers from this issue.
This was discovered by @paschk31.

PrometheusPi:
@ComputationalRadiationPhysics/picongpu-developers Should we backport these kind of issues/bugs?

steindev:
Uhm, if we do a bugfix release, I would say so. But at the moment there is 0.5.0 just around the corner. It becomes the new version (includes these kind of PRs) and so far we did not maintain old versions (as far as I know).

PrometheusPi:
I agree with @steindev. Since 0.5.0 should be out any soon, a backport should not be necessary.

PR #3184, opened 2020-02-25
Fix mapping of cell indexes in the particle merger plugin

KseniaBastrakova:
This caused wrong worker indexes which led to deadlocks. The issue was caused by `DataSpaceOperations::map` no longer handling out-of-domain indices, which the plugin relied on. So now perform the modulo operations inside the plugin.


PrometheusPi:
@KseniaBastrakova Have you performed run time tests?

KseniaBastrakova:
@PrometheusPi Yes, I checked it on KelvinHelmholtz and FoilLCT.
 Before that fix, these examples with applying  Voronoi plugin freeze, and now it works correctly.

PrometheusPi:
@KseniaBastrakova Thanks for performing run time checks.

Issue #3189, opened 2020-02-27
TBG: Connect experiment directory to source directory


Anton-Le:
Currently the last line of the batch script generated using TBG contains the entire call to the script, e.g., 
`#this script was created with call ./bin/tbg -c <path to .cfg> -t <path to .tpl> <target directory>`
if called from the directory where picongpu has been compiled.

I gather this should facilitate reconfiguration, re-running the scripts or general bookkeeping. But since relative pathes are permitted information about the directory of origin is lost.

I think it's better to ensure that `initCall` contains the absolute path to the script instead of just `$0 $*`.

psychocoderHPC:
I would extend the suggestion and add a `cd` before to go into the folder where the user executed the command from to avoid that the path for target directory, cfg and tpl must be explicit too.

```
tbg_path=$0
case $tbg_path in
  /*) ;;
  *) tbg_path=$(which $(basename $tbg_path)) ;;
esac
echo "cd $(pwd); $tbg_path $*"
```

PR #3190, opened 2020-02-27
tbg pathlist

Anton-Le:
Modification of the TBG script which would implement #3189.

### DIFF
--- a/bin/tbg
+++ b/bin/tbg
@@ -275,11 +275,8 @@ help()
 #Check whether the provided path is relative or absolute and
 #convert it to an abolute path
 script=$0
-case $script in
-        /*) absScriptPath=$script;; #called using abs. path
-        ./*) absScriptPath=$PWD/${script#"./"};; #called using ./, remove it
-        *) absScriptPath=$PWD/$script;; #else called e.g. using bash <scriptName>
-esac
+
+absScriptPath=$(realpath $0)
 initCall="cd $(pwd); $absScriptPath $*"
 projectPath="."
 
--- a/bin/tbg
+++ b/bin/tbg
@@ -272,7 +272,15 @@ help()
 }
 
 #!/usr/bin/env bash
-initCall="$0 $*"
+#Check whether the provided path is relative or absolute and
+#convert it to an abolute path
+script=$0
+case $script in
+        /*) absScriptPath=$script;; #called using abs. path
+        ./*) absScriptPath=$PWD/${script#"./"};; #called using ./, remove it
+        *) absScriptPath=$PWD/$script;; #else called e.g. using bash <scriptName>
+esac
+initCall="cd $(pwd); $absScriptPath $*"
 projectPath="."
 
 pathToegetopt=$(which egetopt 2>/dev/null)


bin/tbg
```diff
@@ -272,7 +272,15 @@ help()
 }
 
 #!/usr/bin/env bash
-initCall="$0 $*"
+#Check whether the provided path is relative or absolute and
+#convert it to an abolute path
+script="$0"
+case $script in
+        /*) absScriptPath=$script;; #called using abs. path
+        ./*) absScriptPath=$PWD/${script#"./"};; #called using ./, remove it
```
psychocoderHPC: There is no need for two three cases see sugestion: https://github.com/ComputationalRadiationPhysics/picongpu/issues/3189#issuecomment-591968591

include/picongpu/particles/ParticlesInit.kernel
```diff
@@ -42,7 +42,7 @@ namespace picongpu
      *
      *  That means:
      *     The REAL number of particles density in units of volume**-3,
-     *       normed to UNIT_LENGHT**3
```
psychocoderHPC: please move changes from this file into a separate PR, even if it is doubling the work. The reason is that bose changes are independent from each other.

psychocoderHPC:
please squash your commits. Please have also a look to your local git config. The last commit is different to the first commits.

include/picongpu/particles/ParticlesInit.kernel
```diff
@@ -64,6 +64,10 @@ namespace picongpu
         );
 
         float_X const densityRatioOfSpecies = traits::GetDensityRatio< T_Species >::type::getValue( );
+        PMACC_CASSERT_MSG(
+            Please_set_the_densityRatio_to_a_value_greater_than_0,
+            densityRatioOfSpecies > float_X( 0.0 )
```
psychocoderHPC: density ratio is not `constexpr` and can therefore not checked at compile time. We also have a PR #3109  where we started to introduce a check for `MassRatioValue` but we discarded it. We like to move more stuff into runtime (if possible).

bin/tbg
Anton-Le: I don't think so. Consider the minimal working example:

    #!/bin/bash
    echo "$0"
    curpath=$0
    #convert to abs path
    case $curpath in
            /*) abspath=$curpath;;
           ./*) abspath=$PWD/${curpath#"./"};;
            *) abspath=$PWD/$curpath;;
    esac
    echo "Absolute path: "$abspath

Save it as a script as `realpathExample.sh` in a directory that is NOT in your `PATH`. Then using the suggestion from #3189 would result in an error and no absolute path. 

Removing the default case `*)` from the above will, for instance, prevent the user from obtaining the absolute path when running using `bash realpathExample.sh`.

psychocoderHPC:
@Anton-Le please squash the commits and remove all not required changes that I can review the PR.

psychocoderHPC:
@Anton-Le ping, do you forgot this PR?

psychocoderHPC:
ping

psychocoderHPC:
@Anton-Le PING PING

include/picongpu/particles/ParticlesInit.kernel
Anton-Le: Removed from the commit

bin/tbg
```diff
@@ -272,7 +272,15 @@ help()
 }
 
 #!/usr/bin/env bash
-initCall="$0 $*"
+#Check whether the provided path is relative or absolute and
+#convert it to an abolute path
+script=$0
+case $script in
+        /*) absScriptPath=$script;; #called using abs. path
+        ./*) absScriptPath=$PWD/${script#"./"};; #called using ./, remove it
```
PrometheusPi, line 280: This could be simplified by using `realpath`. It is installed on most systems (we use).

Anton-Le: Admittedly, this is a much more elegant solution which should work assuming the system is running a linux distribution younger than ~2012!

I have incorporated it into the script.

PR #3191, opened 2020-02-28
Correct jupyter widget example

Anton-Le:
The code snippet used to illustrate the usage of the phase space visualization widget has to be corrected. `run_dir_options` expects a list of tuples of the type `( <description>, <path to data>)`.
Additionally add an example illustrating how to add additional phase space slices and to set different particle species for visualization.

### DIFF
--- a/docs/source/usage/plugins/phaseSpace.rst
+++ b/docs/source/usage/plugins/phaseSpace.rst
@@ -178,13 +178,37 @@ After starting the notebook server write the following
    # together with labels that will be used in the plot legends so you still know
    # which data belongs to which simulation
    w = PhaseSpaceWidget(run_dir_options=[
-           ("scan1/sim4", scan1_sim4),
-           ("scan1/sim5", scan1_sim5)])
+           ("scan1/sim4", "/path/to/scan1/sim4"),
+           ("scan1/sim5", "/path/to/scan1/sim5")])
    display(w)
 
 
 and then interact with the displayed widgets.
 
+Plase note that per default the widget allows selection only of the ``ypy`` phase space slice for particles labelled by ``e``.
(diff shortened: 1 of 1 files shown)


PrometheusPi:
@codingS3b It would be great, if you could also review this.

docs/source/usage/plugins/phaseSpace.rst
```diff
@@ -178,13 +178,37 @@ After starting the notebook server write the following
...
+   from picongpu.plugins.jupyter_widgets import PhaseSpaceWidget
+
+   # provide the paths to the simulations you want to be able to choose from
+   # together with labels that will be used in the plot legends so you still know
+   # which data belongs to which simulation
+   w = PhaseSpaceWidget(run_dir_options=[
+           ("scan1/sim4", "/path/to/scan1/sim4"),
+           ("scan1/sim5", "/path/to/scan1/sim5")])
+   w.ps.set_trait('options', ('ypy', 'xpx', 'ypz'))
+   w.species.set_trait('options', ('e', 'b'))
```
codingS3b: Looks good to me. Just out of curiosity: Is there a reason for using the `set_trait` function for altering the options instead of doing e.g. 
```
w.ps.options = ['ypy', 'xpx', 'ypz']
```

Anton-Le: I actually had it that way originally, but switched to the proposed version. The idea being to set properties of an object (here the phase space slice of the widget) using methods of the class - if possible let the class handle how its data is modified, rather than the user.

Both ways work, though.

PrometheusPi: @codingS3b Thank you for your feedback.

PR #3192, opened 2020-03-02
Change imread import from scipy.misc to imageio

codingS3b:
The python reader for the PNG plugin relies on the `scipy.misc.imread` function which was deprecated and removed for `scipy>=1.2.0` according to the docs (https://docs.scipy.org/doc/scipy-1.2.1/reference/generated/scipy.misc.imread.html).
This PR exchanges the import of the `imread` function as suggested in the scipy docs.

### DIFF
--- a/lib/python/picongpu/plugins/data/png.py
+++ b/lib/python/picongpu/plugins/data/png.py
@@ -9,8 +9,10 @@ from .base_reader import DataReader
 
 import numpy as np
 import os
-from scipy import misc
 import collections
+
+from imageio import imread
+
 
 SPECIES_LONG_NAMES = {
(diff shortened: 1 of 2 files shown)


lib/python/picongpu/plugins/data/png.py
```diff
@@ -9,9 +9,11 @@
 
 import numpy as np
 import os
-from scipy import misc
 import collections
```
PrometheusPi: why did you leave an empty line between `import collections`and `from imageio import imread`?

codingS3b: Simply forgot, is fixed.

PR #3197, opened 2020-03-03
Python: fix set rundir options for widgets, EnergyHistogram all zeros

codingS3b:
There are several issues when trying to change the selectable simulation options for a widget using the `set_run_dir_options` function which are addressed in this PR:

1. After updating the dropdown options, for widgets which do not allow multi-selection of simulations (like the png widget), the dropdowns value was also silently set which we do not want since the user has not made a choice on which simulation should be visualized

2. The plot did not get cleared and still visualized a selection of the old run_directories.

3. The simulation time options were not invalidated, possibly containing simulation times from the old selection

4. the plot_mpl object (and therefore also the data readers) was not updated and might have contained data from now outdated simulations.

Moreover, the energy histogram which uses logarithmic plotting did not check if all counts were 0 in which case the plot failed. This is handled now.

### DIFF
--- a/lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
+++ b/lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
@@ -5,10 +5,12 @@ Copyright 2017-2020 PIConGPU contributors
 Authors: Sebastian Starke
 License: GPLv3+
 """
+import numpy as np
 
 from picongpu.plugins.data import EnergyHistogramData
 from picongpu.plugins.plot_mpl.base_visualizer import Visualizer as\
     BaseVisualizer
+from warnings import warn
 
 
 class Visualizer(BaseVisualizer):
@@ -44,6 +46,12 @@ class Visualizer(BaseVisualizer):
 
         counts, bins, iteration, dt = self.data[idx]
         label = self.sim_labels[idx]
+
+        if np.all(counts == 0.):
+            warn("All counts were 0 for {}. ".format(label) +
+                 "No log-plot can be created!")
+            return
+
         self.plt_obj[idx] = self.ax.semilogy(
             bins, counts, nonposy='clip', label=label,
             color=self.colors[idx])[0]
@@ -53,6 +61,13 @@ class Visualizer(BaseVisualizer):
         Implementation of base class function.
         """
         counts, bins, iteration, dt = self.data[idx]
+        label = self.sim_labels[idx]
+
+        if np.all(counts == 0.):
+            warn("All counts were 0 for {}. ".format(label) +
+                 "Log-plot will not be updated!")
+            return
+
         self.plt_obj[idx].set_data(bins, counts)
 
     def visualize(self, **kwargs):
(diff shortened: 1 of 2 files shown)


PrometheusPi:
@psychocoderHPC Do you want to add this to `0.5.0`otherwise, we have to add this to the next release.

codingS3b:
Yes I checked it and everything works as expected, except for the EnergyHistogram widget, which seems to refuse to plot after encountering a first iteration with all zeros. I'm investigating this right now.

PrometheusPi:
@codingS3b Okay - interesting side effect - thank you for setting this issue to WIP

codingS3b:
Okay, it's fixed now. Turned out that the check for `==0` was not within the `np.all` call.

PrometheusPi:
@codingS3b Great finding - sorry that I did not see this 🙇

PrometheusPi:
@codingS3b Could you squash you commits to a few topic-centric commits?

codingS3b:
Commits are squashed. I am a little clueless on why the travis-ci fails for python 2.7 with an invalid-syntax error.

PrometheusPi:
This `E999` error baffles me as well. First of all, it is in a code section you did not touch. Secondly, the [error description](https://gitlab.com/pycqa/flake8/-/merge_requests/130/diffs) is not applicable to that comment section. Restarting the test did not help.

PrometheusPi:
@ax3l Any idea where this strange error comes from?

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
```diff
@@ -44,6 +46,11 @@ def _create_plt_obj(self, idx):
 
         counts, bins, iteration, dt = self.data[idx]
         label = self.sim_labels[idx]
+
+        if np.all(counts == 0.):
+            warn(f"All counts were 0 for {label}. No log-plot can be created!")
```
PrometheusPi: This is not supported by python2
```suggestion
            warn("All counts were 0 for {}. No log-plot can be created!".format(label))
```

lib/python/picongpu/plugins/plot_mpl/energy_histogram_visualizer.py
```diff
@@ -53,6 +60,12 @@ def _update_plt_obj(self, idx):
         Implementation of base class function.
         """
         counts, bins, iteration, dt = self.data[idx]
+
+        if np.all(counts == 0.):
+            warn(f"All counts were 0 for {self.sim_labels[idx]}. "
+                 "Log-plot will not be updated!")
```
PrometheusPi: same python2 issue here 
```suggestion
            warn("All counts were 0 for {}. ".format(self.sim_labels[idx]) +
                 "Log-plot will not be updated!")
```

PrometheusPi:
@codingS3b Strange that the travis-CI did not throw the error earlier. If I remember correctly, you used f-strings before squashing,didn't you? 

@ax3l Do you have any idea, why the travis-CI did not throw the error/warning earlier?

PrometheusPi:
@psychocoderHPC as discussed *offline* today - this should go to `0.5.0`

PR #3208, opened 2020-03-09
Create a new absorber description interface

### LINKED ISSUE #3204: Hardcoded exponential damping absorber in the meta
I just saw it in commit fa64cbbfd4f0018e1d14c876546cabf39b90892d, but the same problem also exists in the HDF5 and ADIOS plugins. I will provide some unified abstraction on top of exponential damping and PML, to be used in such situations. Perhaps it should extend #3161 which only unified the size.

sbastrakov:
This interface is to be used for getting absorber size and string properties.
Move the previously created `fields/absorber/NumCells.hpp` into this file.
This does not provide the unified computational interface, to be done in following PRs (these are not for the upcoming release, but generally for future unification of boundary conditions).
Fixes #3204.

~~I did not check it yet, so currently WIP, will continue on this tomorrow.~~ Edit: checked. This PR has to be part of the 0.5.0.


include/picongpu/fields/absorber/Absorber.hpp
```diff
@@ -0,0 +1,238 @@
...
+        //! Number of absorber cells along the min z boundary
+        static constexpr uint32_t zNegativeNumCells = pml::NUM_CELLS[ 2 ][ 0 ];
+
+        //! Number of absorber cells along the max z boundary
+        static constexpr uint32_t zPositiveNumCells = pml::NUM_CELLS[ 2 ][ 1 ];
+
+        //! Get string properties of the absorber
+        static pmacc::traits::StringProperty getStringProperties()
+        {
+            return getStringProperties( "convolutional PML" );
```
sbastrakov: Note: I should probably clarify here that we can't call `YeePML::getStringProperties()`, since it provides Maxwell's solver properties, as currently the solver and PML are together in that class template.

include/picongpu/fields/absorber/ExponentialDamping.hpp
```diff
@@ -108,51 +108,7 @@ class ExponentialDamping
...
-                else
-                {
-                    propList[directionName]["param"] = "none";
-                }
-
-                propList[directionName]["name"] = boundaryName;
-            }
-        }
-        return propList;
+        return detail::getStringProperties( "exponential damping" );
```
sbastrakov: Note: since this function is reused for the PML, it looked reasonable to move to the new `Absorber.hpp`, also since the inclusion dependencies suggest it.

sbastrakov:
Ready for review.

Issue #3207, opened 2020-03-09
Ionization and charge density


cbontoiu:
Hello,
I am trying to make sense of laser interaction with a carbon structure, preionized one. I have two cases:

1)  the wavelength is 25 nm and a0 = 1e-03 and i get a kind of standing wave; ionization progresses at the maxima of the laser Ex electric field as shown in the first gif. Is this an error? I expected the laser pulse to ionize all the way at a growing rate as the Gaussian pulse reaches the top.

![1](https://user-images.githubusercontent.com/7108691/76215462-b6895200-6206-11ea-83d1-05e59e6193aa.gif)

2) the wavelength is 20 nm and the a0 = 1e-2. Now I get something more related to what i expect bu at some point all electrons disappear and the same happens with the carbon ions. What am I doing wrong here?

![2](https://user-images.githubusercontent.com/7108691/76215821-6e1e6400-6207-11ea-8ce0-1033501686af.gif)

On the other hand for the second case the default png images still show the electrons as you can see in the gif below.

![3](https://user-images.githubusercontent.com/7108691/76216258-2f3cde00-6208-11ea-8548-87b67fd36614.gif)

Yet another situation of losing the beam is shown in the figure below

![4](https://user-images.githubusercontent.com/7108691/76233508-353eb900-6220-11ea-840e-47f93dbac2f7.gif)

for this last one I used absorber at all 6 faces with the YeePML solver so

![image](https://user-images.githubusercontent.com/7108691/76233645-67501b00-6220-11ea-892c-d3a6613979c9.png)

I don't know if it was a good idea.

I would be grateful for any suggestion.
Thank you.

PrometheusPi:
Dear @cbontoiu,

thank you for reporting this issue. In order to make sense of the simulation output, could you please provide the `laser.param`, `density.param`and `speciesInitalization.param` (for now). 

Could you also upload `fields_energy.dat` and ` e_macroParticlesCount.dat`? 

That would help us tremendously in understanding/guessing what happened.

cbontoiu:
Thank you very much for your availability to help. I attach the whole input folder below

[one-tube-thick.zip](https://github.com/ComputationalRadiationPhysics/picongpu/files/4310903/one-tube-thick.zip)

The model uses a PlaneWave laser as 

`struct PlaneWaveParam
            {
                //##################################################
                static constexpr float_64 PULSE_LENGTH_SI = (10e-15)/2.354820045;
[...]

PR #3209, opened 2020-03-10
Add forgotten CFL check for YeePML

sbastrakov:
The reason that was forgotten is that originally the `YeePML` implementation inherited the `Yee` solver which had the CFL check in place, and so PML did not. The inheritance was later changed, and I forgot to add the check back. Thanks to @cbontoiu  for [reporting](https://github.com/ComputationalRadiationPhysics/picongpu/issues/3207#issuecomment-596948103).


sbastrakov:
@psychocoderHPC ready to be merged.

PR #3211, opened 2020-03-12
use cupla instead of cuda prefix for function calls

psychocoderHPC:
Avoid that PMacc/PIConGPU depends on the preprocessor renaming to use
cupla.

Since PMacc is still using sometimes native cuda it is hard to see where cupla and whee CUDA is used. I also work on making the precompiler stuff in cupla optional to increase the compatibility of cupla.

PR #3215, opened 2020-03-19
Replace remaining cupla renamings with calls to cupla API

sbastrakov:
This is a fixup of #3211 which did these renamings for most of the code. Also, update usage in the docs since this is now the preferred way.

Perhaps these were missed in #3211 because only `__syncthreads();` occurances were checked, but not `__syncthreads( );`

I found out as it failed to build on my machine for the CPU accelerator. Wondering why did the tests pass for that PR?

### DIFF
--- a/docs/source/prgpatterns/lockstep.rst
+++ b/docs/source/prgpatterns/lockstep.rst
@@ -45,7 +45,7 @@ Collective Loop
     // `frame` is a list which must be traversed collectively
     while( frame.isValid() )
     {
-        uint32_t const workerIdx = threadIdx.x;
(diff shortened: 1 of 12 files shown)


sbastrakov:
@psychocoderHPC ready for review. Currently, the `dev` is broken for Windows.

psychocoderHPC:
CPU backends are not tested in the CI, therefore the bug was not visible. Thats a reason why we need  our HZDR CI soon.

sbastrakov:
I think it was more complicated with omp2b working on Hemera. But agree with the conclusion.

psychocoderHPC:
Using omp2b on hemera is quite easy you need to load the profile for the defq (queue) and than it works.

Issue #3217, opened 2020-03-26
Weird assymetrical electron density in the LWFA example ( with dev branch ).


pordyna:
I noticed that the LWFA example (default setting, `4.cfg` configuration file) produces weird outputs compared to what it was doing before (current master branch 0.4.3).

Run on hemera ( tried on both `k20` and `gpu` partitions ).
For people with access, here you can find the full simulation outputs (`2020_03_26_dev` and `2020_03_26_0-4-3`):
`/bigdata/hplsim/scratch/ordyna35/runs/LWFA` 

Bellow,  for comparison,  two exemplary pngs generated with the png plugin (XY plane).
**Dev** (900th simulation step, ):
![e_png_yx_0 5_000900](https://user-images.githubusercontent.com/41141557/77641338-32500200-6f5c-11ea-974b-e94c442f16d7.png)

**0.4.3** (900th simulation step):
![e_png_yx_0 5_000900](https://user-images.githubusercontent.com/41141557/77641428-590e3880-6f5c-11ea-866c-033659f74415.png)

This is not only a problem with the png plugin, electron density dumped by the hdf5 or adios looks the same.  

@PrometheusPi, @sbastrakov, @psychocoderHPC

pordyna:
tT me it looks like some of the electrons would stop moving :sweat_smile:

sbastrakov:
Thanks for reporting @pordyna !

sbastrakov:
I am looking into this, as this might be a bug to fix before the release.

sbastrakov:
Intermediate result: this is definitely a bug, introduced to `dev` some time between 5 February and 3 March. I will narrow it down and fix, think today.

PrometheusPi:
@pordyna Thank you for reporting this issue. It is awesome you found this bug. 
I just checked my LWFA simulations I ran tonight and the same error occurs (k20/hemera - 2D only, 16GPUs):
![grafik](https://user-images.githubusercontent.com/5121158/77666878-cb454400-6f81-11ea-8aad-3d6a875e5370.png)

Seems to affect a fixed number of cells at the **absorbers**.

sbastrakov:
@PrometheusPi I think I figured out of bug. Since Feb. 20 the exponential absorber strength was very wrong. Making a PR to fix in a moment.

PrometheusPi:
@sbastrakov Excellent! I am looking forward to your pull request.

sbastrakov:
[...]

PR #3161, opened 2020-02-10
Unify absorber size for PML and exponential damping

sbastrakov:
Add a new variable for the absorber size. Change all usage outside of `.param` files to the new variable. This unifies the size of the exponential damping and PML absorbers. It avoids potential issues with compile-time checks pointed out
in the [comments of #3148](https://github.com/ComputationalRadiationPhysics/picongpu/issues/3148#issuecomment-582865342).

PR #3218, opened 2020-03-26
Fix wrong exponential absorber strength

sbastrakov:
The bug was introduced in #3161 by my copy-paste mistake, sorry. Was likely the cause of report #3217 (checking right now, but 99% sure). 

This needs to go into the 0.5.0 @psychocoderHPC .

### DIFF
--- a/include/picongpu/fields/absorber/ExponentialDamping.hpp
+++ b/include/picongpu/fields/absorber/ExponentialDamping.hpp
@@ -65,7 +65,7 @@ public:
                 uint32_t pos_or_neg = i % 2;
 
                 uint32_t thickness = absorber::numCells[direction][pos_or_neg];
-                float_X absorber_strength = absorber::numCells[direction][pos_or_neg];
+                float_X absorber_strength = ABSORBER_STRENGTH[direction][pos_or_neg];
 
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
Thanks for the fast fix!

sbastrakov:
Confirm that this indeed caused the bug. (On another note, this is yet another reason we should develop tests, as discussed recently)

PrometheusPi:
Just as a note to users: This bug was in the code since [February 20th 2020](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3161#event-3055933893) - so only very recent `dev` versions were effected.

PrometheusPi:
@psychocoderHPC I changed the milestone to `0.5.0` because this should definitely be "back-ported" the the release candidate. - But I did not add, "affects latest release" - since it is just a release candidate - I hope this was fine.

psychocoderHPC:
> @psychocoderHPC I changed the milestone to `0.5.0` because this should definitely be "back-ported" the the release candidate. - But I did not add, "affects latest release" - since it is just a release candidate - I hope this was fine.

OK, note the project is enough to get back ported because it creates a todo list.

psychocoderHPC:
For some reason I was not able to restart the broken travis test. The test was not important for this change, so I merged it with my admin power ^^

PR #3237, opened 2020-05-06
[WIP] Fixed primature exit from voronoi particle merger

KseniaBastrakova:
Problem was: instead of skipping the current particle thread exited the whole of loop frames.
As a result a lot of particles were not processed and we merged less than algorithm expected.

### DIFF
--- a/include/picongpu/plugins/particleMerging/ParticleMerger.kernel
+++ b/include/picongpu/plugins/particleMerging/ParticleMerger.kernel
@@ -206,74 +206,74 @@ namespace particleMerging
                     const voronoiCellId::type voronoiCellId = particle[voronoiCellId_];
                     if( voronoiCellId != -1 )
                     {
-		            	VoronoiCell& voronoiCell = listVoronoiCells[voronoiCellId];
-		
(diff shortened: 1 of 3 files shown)


include/picongpu/plugins/particleMerging/ParticleMerger.kernel
```diff
@@ -204,76 +204,76 @@ namespace particleMerging
                     auto particle = frame[linearThreadIdx];
 
                     const voronoiCellId::type voronoiCellId = particle[voronoiCellId_];
-                    if( voronoiCellId == -1 )
-                        return;
```
psychocoderHPC, line 208: I do not see a difference between both codes. The return is only exiting the lambda function and not skipping other invocations from `forEachFrame`.
Maybe I overlook something but for me both versions are identically from the code flow view.

sbastrakov: Yes, good point. Then maybe two changes were made at once and we got confused.

KseniaBastrakova:
@psychocoderHPC was right, there ist't difference and error here. Close this PR

PR #3239, opened 2020-05-07
Fix incorrect initialization and data race

KseniaBastrakova:
Replace atomicAllInc, which required the same address for all threads in a block, that сaused incorrect number of macroparticles in Voronoi cells with cupla::atomicAdd.
Add syncthreads after voronoi cells start initialization.

### DIFF
--- a/include/picongpu/plugins/particleMerging/VoronoiCell.hpp
+++ b/include/picongpu/plugins/particleMerging/VoronoiCell.hpp
@@ -134,7 +134,7 @@ namespace particleMerging
             const float_X weighting
         )
         {
-            nvidia::atomicAllInc( acc, &this->numMacroParticles, ::alpaka::hierarchy::Threads{} );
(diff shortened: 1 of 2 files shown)


include/picongpu/plugins/particleMerging/VoronoiCell.hpp
```diff
@@ -134,7 +134,7 @@ namespace particleMerging
             const float_X weighting
         )
         {
-            nvidia::atomicAllInc( acc, &this->numMacroParticles, ::alpaka::hierarchy::Threads{} );
+            cupla::atomicAdd(acc, &this->numMacroParticles, static_cast<uint32_t>(1), ::alpaka::hierarchy::Threads{} );
```
sbastrakov, line 137: Note: here different threads in a block may or may not have the same address. So an atomic operation is needed, but `atomicAdd`, not `atomicAllInc`.

sbastrakov:
@psychocoderHPC since I was involved in making the change, think it's better if you reviewed it.

PR #3240, opened 2020-05-11
Backport: Fix incorrect initialization and data race

psychocoderHPC:
backport #3239

- replace atomicAllInc, which required same adress for all threads in block with cupla::atomicAdd
- add syncthreads after voronoi cells initialization


psychocoderHPC:
I updated the PR to the old function names

PR #3242, opened 2020-05-18
update to c++14

psychocoderHPC:
- update cupla to https://github.com/alpaka-group/cupla/tree/b72833cbf0844664b26bbf553289356cfa03c18a and therefore alpaka
  - C++14 is now required
```
GIT_AUTHOR_NAME="Third Party" GIT_AUTHOR_EMAIL="picongpu@hzdr.de" git subtree pull --prefix thirdParty/cupla git@github.com:alpaka-group/cupla.git dev --squash
```
- updated required CMake to 3.15.0 
  - the CMake version in the bash profiles we are shipping for different HPC systems is updated to 3.15.0, we have no access to many systems so I set it to the minimum required version
  - hypnos and  hydra profile is not updated, I will remove both systems in a separate PR
- updated required gcc to 5.5.0

# TODO
- [x] ~~check which CMake module >=3.15.0 is available on taurus (currently not available due to the ssh hacks)~~ If the selected version is not available I will fix it in a follow up PR

*Note:* our compile suite should be updated to gcc 5.5.0 and cmake 3.15.0 (minimal requirement)


psychocoderHPC:
It could be that we need to get this PR soon into mainline and that all other PR's must be rebased. I emptied the travis cache because of the ubuntu update.

cbontoiu:
Hello, 

I encountered a compilation problem of picongpu@develop related to C++ 14.
I work with Spack and %gcc@7.3.0 . All dependencies are there inside the Spack forder. What do I need to upgrade and what will be the commands in Spack, without risking to ruin eveyrhing else.

Thank you,
Cristian

PR #3255, opened 2020-05-27
Fix pedantic warnings

jkelling:


include/pmacc/math/vector/Vector.hpp
```diff
@@ -22,6 +22,11 @@
 
 #pragma once
 
+#if BOOST_COMP_CLANG
+    #pragma clang diagnostic push
```
psychocoderHPC, line 26: Please also add the `pop` in this file else this flag will also effect other files.

jkelling: Ok, if we keep this commit.

I encountered the problem, that even gcc is issuing this warning:
```
 warning: ISO C++11 requires at least one argument for the "..." in a variadic macro
```
This is when using the language standard `-std=c++14`. It only goes away when using `-std=gnu++14`, which I guess is what CMake is doing by default(?).

Would there be a proper fix for this to be compliant with the ISO standard?

psychocoderHPC: Could you please post the full warning. I do not see any variadic template in this file.

jkelling: Apparently this can be made compliant with the standard by putting a comma after the last named argument to explicitly have zero `__VA_ARGS__` , [see GCC docs](https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html). According to this site, the currenty style will become compliant with C++2a (I did not check if it was accepted in C++20).

Would such a change of style be acceptable to get rid of this warning for good?

jkelling: The problem is not the macro but its invocation. The full warning is:
```
./include/pmacc/math/vector/Vector.hpp:89:52: warning: ISO C++11 requires at least one argument for the "..." in a variadic macro
    89 |                           T_Dest::dim == T_Src::dim);
       |                                                    ^
```
https://github.com/ComputationalRadiationPhysics/picongpu/pull/3255/commits/76cc7a437379e3a9e63fa85e8bc5ab341f8650a0 addresses the root cause of this warning.

include/picongpu/particles/flylite/IFlyLite.hpp
```diff
@@ -77,7 +77,7 @@ namespace flylite
              */
             PMACC_STATIC_ASSERT_MSG(
                 false && sizeof(T_IonSpecies) != 0,
-                FLYlite_the_update_method_for_ion_population_kinetics_is_not_implemented
+                FLYlite_the_update_method_for_ion_population_kinetics_is_not_implemented,
```
sbastrakov, line 80: Why are these commas after the last argument for a macro required?

jkelling: Apparently the standard requires the comma before the__VA_ARGS__ to be always there, even if the variadic arguments are empty. C++14 final draft 16.3.4:
```
If the identifier-list in the macro definition does not end with an ellipsis, the number of arguments (including
those arguments consisting of no preprocessing tokens) in an invocation of a function-like macro shall equal
the number of parameters in the macro definition. Otherwise, there shall be more arguments in the invocation
than there are parameters in the macro definition (excluding the ...)
```
It strictly says "more", not allowing for equal.

sbastrakov: Thx for explaining.

include/pmacc/algorithms/reverseBits.hpp
```diff
@@ -43,7 +43,7 @@ T
 reverseBits(T value)
 {
     PMACC_STATIC_ASSERT_MSG( boost::is_integral<T>::value && boost::is_unsigned<T>::value,
-                             Only_allowed_for_unsigned_integral_types );
+                             Only_allowed_for_unsigned_integral_types, );
```
psychocoderHPC, line 46: Please wait with merge. Let us check if we can workaround it different. The problem I see here is that you need to know that the macro is using `__VA_ARGS__`.

jkelling: I do not think this is a problem: When you use macro you need to know its interface anyway. The distinction is not whether the macro actually uses the `__VA_ARGS__` value, but whether it is a variadic macro, i.e. whether the argument list ends with an ellipsis. There will  only be a trailing comma, if there are no variadic arguments given in the invocation of a variadic macro.

psychocoderHPC: We use sometimes macros to give the user a better look and feel. It is not obvious for someone why there mist be a comma after the last argument.

psychocoderHPC: I know now how to get it running without destroying the user  side.

```
#define identifier_do(name,...) \
  PMACC_identifier(name,__COUNTER__, __VA_ARGS__  ) 
#define identifier(...) identifier_do(__VA_ARGS__)
```

psychocoderHPC: @jkelling Could you please remove the commit with the comma at the end. We will ship a fix in a seperate PR. I will add a intermediate definition everywhere where the user normally must add the comma.

psychocoderHPC: STOP wait my test code is wrong

psychocoderHPC: After few hours I found a solution to move the pain to the developer instead the users.

```
#include <boost/preprocessor/variadic/size.hpp>
#include <boost/preprocessor/punctuation/comma_if.hpp>
#include <boost/preprocessor/arithmetic/dec.hpp>
#include <boost/preprocessor/comparison/equal.hpp>

#define PMACC_COMMA_IF(count, ...) BOOST_PP_COMMA_IF(BOOST_PP_EQUAL(count,BOOST_PP_VARIADIC_SIZE(__VA_ARGS__)))
#define TAIL(head, ...) __VA_ARGS__

#define PMACC_identifier(name,id,...) \
        struct name \
        { \
                __VA_ARGS__ \
        };

#define identifier_do2(name,...) PMACC_identifier(name,__COUNTER__, __VA_ARGS__ ) 
#define  identifier_do(...) identifier_do2( __VA_ARGS__ )
#define identifier(...) identifier_do(__VA_ARGS__ PMACC_COMMA_IF(1,__VA_ARGS__))

int main()
{
        identifier(xx)
        identifier(zz, int x;)
        identifier(idd, int x;)

        return 0;
}
```

So let us ship the fixin a separate PR.

psychocoderHPC: `icc 19`, `clang10` and `gcc 7.4.0` and `gcc 8.2.0` working fine with this solution

jkelling: This does indeed compile without warning.

Are you sure this is worth it? It is a lot of code to add some syntactic sugar on the use side, which if not present only causes a warning.

The PIConGPU core could just lead by example and add the trailing comma, leaving the user to put it there or not. In PRs for new examples, core devolpers could enforce this.

jkelling: To explain: My problem with this being a lot of code is, that this makes is harder for a reader to figure out what these macros are actually doing, putting a higher burden on the documentation, which now needs to explain that
1. only `identifier(...)` is the one to use and what it is for
2. the other are implementation details of `identifier(...)`
3. what the intermediated staps are good for.

sbastrakov: I do not have a particular opinion. I think the trailing comma would be not expected by some users or developers (me included before this conversation), so that could lead to mysterious and probably hard-to-figure-out build errors. However, I also see your point about adding even more complexity to enable this small thing.

jkelling: @psychocoderHPC you may either need to annouce a descision or put this to a vote for all picongpu developers.

jkelling: Btw.: If you have a trailing comma in an invocation of a non-variadic macro,
```c++
#define foo(name) name;

int main()
{
	foo(int a,)
	return 0;
}
```
 gcc does give a pretty straight error message:
```
foo.cpp:5:12: error: macro "foo" passed 2 arguments, but takes just 1
    5 |  foo(int a,)
      |            ^
foo.cpp:1: note: macro "foo" defined here
    1 | #define foo(name) name;
      | 
foo.cpp: In function ‘int main()’:
foo.cpp:5:2: error: ‘foo’ was not declared in this scope
    5 |  foo(int a,)
      |  ^~~
```
( apart form the last part where `foo` is suddenly not defined :/ , probably because the mactro was not replaced. ).

So, if people get confused and start placing trailing commas in all macro invocation, the compiler will tell them quite clearly that they should not be doing this.

The missing trailing comma is also really only a `-Wpedantic` warning, it is not raised with only `-Wall -Wextra`, so users may not encounter this.

psychocoderHPC: @jkelling The parts where we use variadic templates is mostly driven by the fact that we like to have a nice look and feel. `...` is in the macros used to capture only one parameter but if you like to do things like `identifier(foo, Vector<int,5>)` you need to write `identifier(foo, (Vector<int,5>))` if you like to write it correctly and make visible that there are only two parameter for the macro function. To avoid `()` we use `...` in PMacc. Also in some cases we can not use `()` e.g. if create strctures wiht a macro.
The added code will not be the problem but requires a doxygen string to describe the behaviour of `...` for each macro function.

The additional code we need to add is only  two lines
```
#define  identifier_do(...) identifier_do2( __VA_ARGS__ )
#define identifier(...) identifier_do(__VA_ARGS__ PMACC_COMMA_IF(1,__VA_ARGS__))
```

for `identifier()`, `PMACC_ALIGN()` and our compile time asserts.

psychocoderHPC: @jkelling could you keep this PR as it is. I will push an update to this PR. I will only change parts the user (PIConGPU developer) is interacting with. There are a few commas which can stay because this is only used internally.

psychocoderHPC:
@jkelling How do you tested your changes. I try to work on this PR but alpaka has similar issue. For me it was not possible to separate PMacc warnings from millions of alpaka warnings when adding `--pedantic` with GCC.

jkelling:
* Regarding Alpaka, I use the OpenACC branch ( https://github.com/jkelling/alpaka ), although I am not aware, that there are commits in this branch removing pedantic warnings, so dev should not give warnings either.
* Regarding PIConGPU, I only compiled the [SPECBenchmark](https://github.com/jkelling/picongpu/tree/fixWarnings/share/picongpu/examples/SPECBenchmark) example form [this branch](https://github.com/jkelling/picongpu/tree/fixWarnings) which also contains the fixes from this PR. This could affect which code-paths are compiled.
* I did not use cmake for the build. Some definitions are required to keep boost from tripping lots of warnings:
  ```bash
  g++ -c -o include/picongpu/main.o -DNDEBUG -DBOOST_RESULT_OF_USE_TR1_WITH_DECLTYPE_FALLBACK -DBOOST_NO_AUTO_PTR -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER_ID=UNKNOWN_IN_SPEC -DCMAKE_CXX_COMPILER_VERSION=UNKNOWN_IN_SPEC -DCMAKE_SYSTEM=UNKNOWN_IN_SPEC -DCMAKE_SYSTEM_PROCESSOR=UNKNOWN_IN_SPEC -DENABLE_ADIOS=0 -DENABLE_NVML=0 -DPIC_VERBOSE_LVL=1 -DPMACC_VERBOSE_LVL=0  -I./boost -I./include -I./thirdParty/cupla/include -I./thirdParty/mallocMC/src/include  -O2 -DALPAKA_DEBUG=2 -g -Wno-extra-semi -Wpedantic    -std=c++14      -I/usr/lib/x86_64-linux-gnu/openmpi/include/     include/picongpu/main.cpp
  ```
  However, the cmake scripts would take care of this.
  * Also: I compiled with C++14, not 11.

In any case would see >100 warning about superflous `;` after marcro invocations, you will need to igrnore those, e.g.:
```bash
make | grep "warning:"  | grep -v "warning: extra ';'"
```
GCC ignores any `-Wno*` flags which conflict with `-pedantic`.

psychocoderHPC:
> Regarding Alpaka, I use the OpenACC branch ( https://github.com/jkelling/alpaka ), although I am not aware, that there are commits in this branch removing pedantic warnings, so dev should not give warnings either.

Alpaka dev is full of equal warning when I use `--pedantic` for PIConGPU. 
I need to think about how I can work on it.

> In any case would see >100 warning about superflous ; after marcro invocations, you will need to igrnore those, e.g.:

This is not the problem I need to work on the comma issue but since alpaka throws the same kind of warnings I can not filter.

jkelling:
I also only tried the backends `ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED` and `ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLED`, not cuda, maybe this helps?

Best Regards,

Jeffrey

jkelling:
Ok, I can confirm, that many warnings about the missing comma are raised when I use the cuda backend. Using e.g. the sequential backend should solve the issue for you.

psychocoderHPC:
> I also only tried the backends `ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED` and `ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLED`, not cuda, maybe this helps?

A this is the reason. I used CUDA, so I expect you will see many issues when CUDA is activated. This must be solved for SPEC too.

jkelling:
Regarding fixing this type of warning in the CUDA backend of Alpaka: nvcc (cuda 10.1) does apparently not support the standard.

The CUDA backend has macro invocations like this:
```c++
#define ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_IGNORE(cmd, ...)\  // (1)
        check_things();

#define ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(cmd) \
    ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_IGNORE(cmd)    // (2)

ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(ALPAKA_API_PREFIX(SetDevice)(m_dev.m_iDevice)); // (3)
```
Since (1) is variadic gcc warns at (3) because of the expansion (2). If we put a trailing comma at (2) nvcc complains:
```
./include/alpaka/event/EventUniformCudaHipRt.hpp(66): error: expected an expression
```

So, if these warnings are fixed in the CUDA backend, it will not compile anymore.

psychocoderHPC:
Sry for the long delay.

@jkelling Could you please rebase this PR. We will move it for now into the code base as it is because I have no time to add a clean solution soon.

jkelling:
Rebased.

psychocoderHPC (approved):

PR #3266, opened 2020-06-12
Add debug info in guard check for super cell size

pordyna:
This adds some additional debugging information in a case of a super cell size that's too small to fit the guard. Maybe it could be useful to someone else. 

I've added this lines when I was wondering why my code keeps throwing this. In that particular case the problem was caused by an initialization of an empty mapping description with a SC size (0,0,0). Something that couldn't be addressed by verifying the `.param` files.


PrometheusPi:
@pordyna You pushed you branch accidentally against `master`. I now set it to `dev` but merge conflicts occured. Could you please resolve those. The code itself looks good to me :+1:

PR #3272, opened 2020-06-24
refactor pmacc  identifier and alias

psychocoderHPC:
- use `constexpr` for instances e.g. `momentum_`
- enforce a semicolon after the macro
- fix `speciesAttributes.param` - add missing semicolon

Co-authored-by: Jeffrey Kelling <j.kelling@hzdr.de>

Since the namespaces for the types are removed our kernel names will be 300-400 characters shorter :tada:

### DIFF
--- a/include/pmacc/identifier/alias.hpp
+++ b/include/pmacc/identifier/alias.hpp
@@ -36,44 +36,6 @@ identifier(pmacc_void);
 identifier(pmacc_isAlias);
 } //namespace pmacc
 
-#ifdef __CUDACC__
-#   define PMACC_alias_CUDA(name,id)                                          \
-        namespace PMACC_JOIN(device_placeholder,id){                          \
-            /* This variable exists only for template parameter deduction, its
-             * value is never used. So in this case it is fine to have a
-             * separate version in each translation unit due to static.
-             */                                                               \
-            static __constant__ PMACC_JOIN(placeholder_definition,id)::name<> \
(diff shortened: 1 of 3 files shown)


PrometheusPi:
Shorter kernel names sound great :smile: - But perhaps I will miss error messages during compilation in the future because they are not a kilometer long anymore 😨

include/pmacc/identifier/alias.hpp
```diff
@@ -90,7 +52,19 @@ identifier(pmacc_isAlias);
...
+#define alias(name)                                                            \
+        template<                                                              \
+            typename T_Type = pmacc::pmacc_void,                               \
+            typename T_IsAlias = pmacc::pmacc_isAlias                          \
+        >                                                                      \
+        struct name                                                            \
+        {                                                                      \
+            static std::string getName()                                       \
+            {                                                                  \
+                return std::string(#name);                                     \
```
PrometheusPi, line 64: why `#`?

sbastrakov: It is to get stringified version of the name, in quotes.

PrometheusPi: @sbastrakov Thnaks for the explanation

PrometheusPi: That allowed me to google: is a pre-compiler command: https://gcc.gnu.org/onlinedocs/gcc-4.8.5/cpp/Stringification.html

PR #3276, opened 2020-07-02
Fix checkpointing and output of PML fields for uneven domains

### LINKED ISSUE #3275: PML with uneven domains is not output and check pointed correctly
The logic of writing and reading of PML fields relies on the domains being of equal size. It equally concerns output and checkpointing. 

Here is the code that assumed it:  [1](https://github.com/ComputationalRadiationPhysics/picongpu/blob/fe322db49e69b09258691c4aa781769470fe693d/include/picongpu/plugins/hdf5/writer/Field.hpp#L119-L142), [2](https://github.com/ComputationalRadiationPhysics/picongpu/blob/fe322db49e69b09258691c4aa781769470fe693d/include/picongpu/plugins/hdf5/restart/RestartFieldLoader.hpp#L93-L115), and similar for the other two output plugins. The error is that while the allocated-for PML thickness is indeed the same (even for internal domains), that does not imply the buffer size is the same, as the latter also depends on the local domain size.

This concerns input/output only, not the correctness of the computational part. Note that the PML fields are actually not needed in the normal simulation output and their output is not enabled by default. However, for checkpointing they are required and enabled.

The incorrect behavior discovered by @Anton-Le

sbastrakov:
Fix #3275 for all three output backends.

Since the three output plugins we have are largely similar, the changes are basically copy-pasted in all 3 plugins. I deliberately did not go into abstracting and reusing, since this is a patch (to a patch for PML support) and the HDF5 and ADIOS plugins will (hopefully) be removed soon.

### DIFF
--- a/include/picongpu/plugins/adios/ADIOSWriter.hpp
+++ b/include/picongpu/plugins/adios/ADIOSWriter.hpp
@@ -42,7 +42,9 @@
 #include "picongpu/fields/MaxwellSolver/YeePML/Field.hpp"
 #include <pmacc/particles/operations/CountParticles.hpp>
 
+#include <pmacc/communication/manager_common.hpp>
 #include <pmacc/dataManagement/DataConnector.hpp>
+#include <pmacc/Environment.hpp>
 #include <pmacc/mappings/simulation/GridController.hpp>
 #include <pmacc/mappings/simulation/SubGrid.hpp>
 #include <pmacc/dimensions/GridLayout.hpp>
@@ -556,9 +558,7 @@ private:
         auto fieldsOffsetDims = params->fieldsOffsetDims;
 
         /* Patch for non-domain-bound fields
-         * This is an ugly fix to allow output of reduced 1d PML buffers,
-         * that are the same size on each domain.
-         * This code is to be replaced with the openPMD output plugin soon.
+         * This is an ugly fix to allow output of reduced 1d PML buffers
          */
         if( !traits::IsFieldDomainBound< T_Field >::value )
         {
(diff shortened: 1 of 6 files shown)


sbastrakov:
Current status: the fix seems to work for a small example with uneven domains for the HDF5 backend. Now I will be testing a more complex setup, with moving window and decomposition along all 3 axes. And if that works will make similar changes for the other two output backends.

sbastrakov:
Ready for review!

sbastrakov:
cc @Anton-Le 
I tried on a small example using `gridDist`, and with this fix everything seemed to work.

sbastrakov:
Note that this functionality is broken in the recent release, so we should perhaps make a backport afterwards.

Anton-Le:
@sbastrakov Great! I will test it with the original problem.

Anton-Le:
Great job! This fixes the problem of failed restarts for the setup it has first been observed in!

I also agree with @sbastrakov that this fix should be applied to the recent release.

sbastrakov:
Thanks for checking @Anton-Le .

@psychocoderHPC could you review this PR?

include/picongpu/plugins/adios/ADIOSWriter.hpp
```diff
@@ -556,23 +558,48 @@ class ADIOSWriter : public IIOBackend
...
             DataConnector &dc = Environment<>::get().DataConnector();
             auto field = dc.get< T_Field >( T_Field::getName() );
             fieldsSizeDims = precisionCast< uint64_t >( field->getGridLayout().getDataSpaceWithoutGuarding() );
             dc.releaseData( T_Field::getName() );
-            auto const & gridController = Environment<simDim>::get().GridController();
-            auto const numRanks = gridController.getGlobalSize();
-            auto const rank = gridController.getGlobalRank();
+            auto & gridController = Environment<simDim>::get().GridController();
+            auto const numRanks = uint64_t{ gridController.getGlobalSize() };
+            auto const rank = uint64_t{ gridController.getGlobalRank() };
```
psychocoderHPC, line 571: Only a note: @sbastrakov and me discussed offline that the rank is not the best unique id as marker to load data. The reason is that we can not guarantee that the rank is deterministic.

psychocoderHPC: We need to introduce a unique reproducible ID in pmacc. This should not be part of this PR and will break the compatibility of checkpoints written with this PR and later checkpoints.

sbastrakov: Discussed offline, I will change MPI rank to the rank based on domain position. There is already `gridController.getScalarPosition()` for getting it.

sbastrakov: Done.

sbastrakov:
I've changed to ranks to local domain index, re-ran the tests and rebased. Ready for another review @psychocoderHPC .

PR #3280, opened 2020-07-06
Add Higuera-Cary pusher

### LINKED ISSUE #2589: Pusher: Structure-preserving, E-Field Compensating
Just found this new pusher from Higuera and Cary (2017): structure-preserving (phase-space volume, energy) like Boris' method **and** E-B-drift conserving (like Vay's method) for relativistic beams (in the lab frame).

Properties of the new pusher (all three 2nd order accurate):

  - Boris: energy conserving, NOT E + UxB
  - Vay: NOT energy conserving, but E + UxB
  - new: energy conserving AND E + UxB

Changes to Boris' method?
One line:

https://github.com/ComputationalRadiationPhysics/picongpu/blob/9b996254c4f6a667c91d119aa5ee9715b21d0912/include/picongpu/particles/pusher/particlePusherBoris.hpp#L66-L67

Reference:
  https://arxiv.org/abs/1701.05605 ([DOI:10.1063/1.4979989](https://dx.doi.org/10.1063/1.4979989))

steindev:
Add Higuera-Cary pusher as presented in [doi:10.1063/1.4979989](https://doi.org/10.1063/1.4979989).
The HC method is 2nd order, structure preserving, and relativistically correct.
Compared to Vay's method, the HC method has the additional property of being structure preserving while it correctly calculates the velocity in ExB direction when E·B=0 in contrast to Boris' method.
A correction to the formulas given in the original article is applied as documented by the WarpX team: (https://github.com/ECP-WarpX/WarpX/issues/320).
 
Further references:
* [Higuera's article on arxiv](https://arxiv.org/abs/1701.05605)
* [Riperda's comparison of relativistic particle integrators](https://doi.org/10.3847/1538-4365/aab114)

**Tests performed so far**
* Single particle acceleration in constant electric field
  - HC produces the exact same result as Boris

* Single particle turning circles in a constant magnetic field
  - particle sticks closer to (expected) circular trajectory than Boris' or Vay's method

* Single particle movement with constant velocity in a force-free field composed of constant magnetic and electric fields
  - ~HC performs better than Boris, but the deviation from the expected straight line motion is still large.
    Only Vay produces the expected result, but it was Vay's design goal to reproduce the correct behavior truthfully.
    HC's issue arises mainly due to insufficient precision of the input values (i.e. fields), as it relies on the cancellation of two very large equal but opposite numbers when calculating the new momentum, in this test case.~
  - UPDATE: I experimented a bit with `precisionCast`s and finally figured how to reduce errors due to upcasting while calculating with the necessary high precision, too.
    The HC pusher now performs as good as the Vay pusher and reproduces the trajectory.
    (Therefore, I would like to spare discussions on that)

* Kelvin-Helmholtz instability
  - Virtually no difference in electron distribution between Boris and HC at step 800. See the plot below showing where the distributions differ.
    KHI_diff_800_gimp
    Note, this picture amplifies the real difference. White pixels just mark that there is a change in that pixel, but not how large it is. Drawing the absolute difference in electron distributions results in a black void only.
  - Charge conservation: for HC is about the same as for Boris, i.e. max|d|/ρ_0=3.8e-5 and 4.0e-5, respectively, at the end of the simulation
  - Numerical heating: curves of particle energy over time relative to initial energy are identical for Boris and HC

**To Do**
- [x] figure out if HC pusher gives expected results in force-free test
- [x] perform many particle test (e.g. KHI)
- [x] add option of the new HC pusher in list of available pushers in `species.param` in all examples

Big THANKS to @aroeszler who implemented most of the method.

This will close #2589.

### DIFF
--- a/include/picongpu/param/pusher.param
+++ b/include/picongpu/param/pusher.param
@@ -62,13 +62,13 @@ namespace picongpu
         namespace sqrt_Vay = precision64Bit;
     }
 
-    namespace particlePusherHC
+    namespace particlePusherHigueraCary
     {
         /** Precision of the square roots during the push step
          *  - precision32Bit
          *  - precision64Bit
          */
-        namespace sqrt_HC = precision64Bit;
+        namespace sqrt_HigueraCary = precision64Bit;
     }
 
     namespace particlePusherAxel
@@ -87,7 +87,7 @@ namespace picongpu
     {
     namespace pusher
     {
-        struct HC;
+        struct HigueraCary;
         struct Vay;
         struct Boris;
         struct Photon;
--- a/include/picongpu/param/pusher.param
+++ b/include/picongpu/param/pusher.param
@@ -69,7 +69,7 @@ namespace picongpu
          *  - precision64Bit
          */
         namespace sqrt_HC = precision64Bit;
-    }    
+    }
 
     namespace particlePusherAxel
     {
--- a/include/picongpu/param/pusher.param
+++ b/include/picongpu/param/pusher.param
@@ -62,6 +62,15 @@ namespace picongpu
         namespace sqrt_Vay = precision64Bit;
     }
 
+    namespace particlePusherHC
+    {
+        /** Precision of the square roots during the push step
+         *  - precision32Bit
+         *  - precision64Bit
+         */
+        namespace sqrt_HC = precision64Bit;
+    }    
+
     namespace particlePusherAxel
     {
 
@@ -78,6 +87,7 @@ namespace picongpu
     {
     namespace pusher
     {
+        struct HC;
         struct Vay;
         struct Boris;
         struct Photon;
--- a/include/picongpu/param/species.param
+++ b/include/picongpu/param/species.param
@@ -89,6 +89,6 @@ using UsedParticleCurrentSolver = currentSolver::Esirkepov< UsedParticleShape >;
  * For development purposes: --------------------------------------------------
  * - particles::pusher::Axel : a pusher developed at HZDR during 2011 (testing)
  */
-using UsedParticlePusher = particles::pusher::HigueraCary;
+using UsedParticlePusher = particles::pusher::Boris;
 
 } // namespace picongpu
--- a/include/picongpu/param/species.param
+++ b/include/picongpu/param/species.param
@@ -1,4 +1,4 @@
-/* Copyright 2014-2020 Rene Widera, Richard Pausch
+/* Copyright 2014-2020 Rene Widera, Richard Pausch, Annegret Roessler
  *
  * This file is part of PIConGPU.
  *
@@ -74,6 +74,7 @@ using UsedParticleCurrentSolver = currentSolver::Esirkepov< UsedParticleShape >;
  *
  * Defining a pusher is optional for particles
  *
+ * - particles::pusher::HC : best of Vay and Boris pusher
  * - particles::pusher::Vay : better suited relativistic boris pusher
  * - particles::pusher::Boris : standard boris pusher
  * - particles::pusher::ReducedLandauLifshitz : 4th order RungeKutta pusher
--- a/include/picongpu/param/species.param
+++ b/include/picongpu/param/species.param
@@ -74,9 +74,9 @@ using UsedParticleCurrentSolver = currentSolver::Esirkepov< UsedParticleShape >;
  *
  * Defining a pusher is optional for particles
  *
- * - particles::pusher::HC : best of Vay and Boris pusher
- * - particles::pusher::Vay : better suited relativistic boris pusher
- * - particles::pusher::Boris : standard boris pusher
+ * - particles::pusher::HC : Higuera & Cary's relativistic pusher preserving both volume and ExB velocity
+ * - particles::pusher::Vay : Vay's relativistic pusher preserving ExB velocity
+ * - particles::pusher::Boris : Boris' relativistic pusher preserving volume
  * - particles::pusher::ReducedLandauLifshitz : 4th order RungeKutta pusher
  *                                              with classical radiation reaction
  *
@@ -89,6 +89,6 @@ using UsedParticleCurrentSolver = currentSolver::Esirkepov< UsedParticleShape >;
  * For development purposes: --------------------------------------------------
  * - particles::pusher::Axel : a pusher developed at HZDR during 2011 (testing)
  */
-using UsedParticlePusher = particles::pusher::Boris;
+using UsedParticlePusher = particles::pusher::HC;
 
 } // namespace picongpu
--- a/include/picongpu/param/species.param
+++ b/include/picongpu/param/species.param
@@ -1,4 +1,4 @@
-/* Copyright 2014-2020 Rene Widera, Richard Pausch, Annegret Roessler
+/* Copyright 2014-2020 Rene Widera, Richard Pausch, Annegret Roeszler
  *
  * This file is part of PIConGPU.
  *
@@ -74,7 +74,7 @@ using UsedParticleCurrentSolver = currentSolver::Esirkepov< UsedParticleShape >;
  *
  * Defining a pusher is optional for particles
  *
- * - particles::pusher::HC : Higuera & Cary's relativistic pusher preserving both volume and ExB velocity
+ * - particles::pusher::HigueraCary : Higuera & Cary's relativistic pusher preserving both volume and ExB velocity
  * - particles::pusher::Vay : Vay's relativistic pusher preserving ExB velocity
  * - particles::pusher::Boris : Boris' relativistic pusher preserving volume
  * - particles::pusher::ReducedLandauLifshitz : 4th order RungeKutta pusher
@@ -89,6 +89,6 @@ using UsedParticleCurrentSolver = currentSolver::Esirkepov< UsedParticleShape >;
  * For development purposes: --------------------------------------------------
  * - particles::pusher::Axel : a pusher developed at HZDR during 2011 (testing)
  */
-using UsedParticlePusher = particles::pusher::HC;
+using UsedParticlePusher = particles::pusher::HigueraCary;
 
 } // namespace picongpu
--- a/include/picongpu/particles/pusher/particlePusherHC.hpp
+++ b/include/picongpu/particles/pusher/particlePusherHC.hpp
@@ -32,10 +32,11 @@ namespace particlePusherHC
 /* Implementation of the Higuera-Cary pusher as presented in doi:10.1063/1.4979989.
  * A correction is applied to the given formulas as documented by the WarpX team:
  * (https://github.com/ECP-WarpX/WarpX/issues/320).
- * 
+ *
  * Note, while Higuera and Ripperda present the formulas for the quantity u = gamma * v,
  * PIConGPU uses the real momentum p = gamma * m * v = u * m for calculations.
- * 
+ * Here, all auxiliary quantities are equal to those used in Ripperda's article.
+ *
  * Further references:
  * [Higuera's article on arxiv](https://arxiv.org/abs/1701.05605)
  * [Riperda's comparison of relativistic particle integrators](https://doi.org/10.3847/1538-4365/aab114)
@@ -79,7 +80,7 @@ struct Push
          * Notation is according to Ripperda's paper
          */
         // First half electric field acceleration
-        const MomType mom_minus = mom + float_X(0.5) * charge * eField * deltaT;
+        const sqrt_HC::float3_X mom_minus = precisionCast<sqrt_HC::float_X>( mom + float_X(0.5) * charge * eField * deltaT );
 
         // Auxiliary quantitites
         const sqrt_HC::float_X gamma_minus = gamma( mom_minus , mass );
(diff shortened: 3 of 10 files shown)


include/picongpu/param/pusher.param
```diff
@@ -62,6 +62,15 @@ namespace picongpu
         namespace sqrt_Vay = precision64Bit;
     }
 
+    namespace particlePusherHC
```
PrometheusPi, line 65: optional: in other places, e.g. the current deposition methods, there is no abbreviation used.

include/picongpu/param/species.param
```diff
@@ -88,6 +89,6 @@ using UsedParticleCurrentSolver = currentSolver::Esirkepov< UsedParticleShape >;
  * For development purposes: --------------------------------------------------
  * - particles::pusher::Axel : a pusher developed at HZDR during 2011 (testing)
  */
-using UsedParticlePusher = particles::pusher::Boris;
+using UsedParticlePusher = particles::pusher::HC;
```
PrometheusPi, line 92: Please do not change the default for now.

include/picongpu/particles/pusher/particlePusherHC.hpp
```diff
@@ -0,0 +1,139 @@
+/* Copyright 2013-2020 Heiko Burau, Rene Widera, Richard Pausch, Annegret Roeszler, Klaus Steiniger
```
PrometheusPi, line 1: Inconsistency in @aroeszler name compared to the other files

include/picongpu/particles/pusher/particlePusherHC.hpp
```diff
@@ -0,0 +1,139 @@
...
+/* Implementation of the Higuera-Cary pusher as presented in doi:10.1063/1.4979989.
+ * A correction is applied to the given formulas as documented by the WarpX team:
+ * (https://github.com/ECP-WarpX/WarpX/issues/320).
+ *
+ * Note, while Higuera and Ripperda present the formulas for the quantity u = gamma * v,
+ * PIConGPU uses the real momentum p = gamma * m * v = u * m for calculations.
+ * Here, all auxiliary quantities are equal to those used in Ripperda's article.
+ *
+ * Further references:
+ * [Higuera's article on arxiv](https://arxiv.org/abs/1701.05605)
```
PrometheusPi, line 41: why do you recite the above article but with an arXiv reference now?

include/picongpu/particles/pusher/particlePusherHC.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+        for( uint32_t d=0 ; d<simDim ; ++d )
+        {
+            pos[d] += ( vel[d] * deltaT ) / cellSize[d];
+        }
+    }
+
+    static pmacc::traits::StringProperty getStringProperties()
+    {
+        pmacc::traits::StringProperty propList( "name", "HC" );
```
PrometheusPi, line 134: a verbose name would be more intuitive

include/picongpu/particles/pusher/particlePusherHC.hpp
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * Further references:
+ * [Higuera's article on arxiv](https://arxiv.org/abs/1701.05605)
+ * [Riperda's comparison of relativistic particle integrators](https://doi.org/10.3847/1538-4365/aab114)
+ */
+
+template<class Velocity, class Gamma>
+struct Push
+{
+    /* this is an optional extension for sub-sampling pushes that enables grid to particle interpolation
```
PrometheusPi, line 48: I do not understand this comment. 
Do you mean sub-sampling in time?

include/picongpu/particles/pusher/particlePusherHC.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+        auto bField  = functorBField(pos);
+        auto eField  = functorEField(pos);
+
+        const float_X deltaT = DELTA_T;
+
+
+        Gamma gamma;
+
+        /*
```
PrometheusPi, line 78: please use the first line of the multi line comment according to the [style guide](https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md#4-comments)

include/picongpu/param/pusher.param
```diff
@@ -62,6 +62,15 @@ namespace picongpu
...
+    {
+        /** Precision of the square roots during the push step
```
PrometheusPi, line 67: In your code, no sqrt operation is used directly, but this datatype is used repeatedly - please update this comment to represent the use of the high-precision data type.

include/picongpu/particles/pusher/particlePusherHC.hpp
sbastrakov: Agree. I would actually like to have `HigueraCary` instead of `HC` as the pushed name. That would be consistent with other pushers named by the last name (except Axel's).

include/picongpu/particles/pusher/particlePusherHC.hpp
sbastrakov: Yes, it is probably sub-stepping [in time], not sub-sampling.

include/picongpu/param/pusher.param
sbastrakov: I would also prefer the full name everywhere instead of HC. (internally in the implementation you could make an alias if the names are too wordy).

include/picongpu/particles/pusher/particlePusherHC.hpp
```diff
@@ -0,0 +1,139 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+
```
sbastrakov, line 21: These two new lines are excessive.

include/picongpu/particles/pusher/particlePusherHC.hpp
```diff
@@ -0,0 +1,139 @@
...
+#pragma once
+
+#include "picongpu/simulation_defines.hpp"
+#include "picongpu/traits/attribute/GetMass.hpp"
+#include "picongpu/traits/attribute/GetCharge.hpp"
+
```
sbastrakov, line 27: And here should be one more new line.

include/picongpu/particles/pusher/particlePusherHC.hpp
```diff
@@ -0,0 +1,139 @@
...
+ * PIConGPU uses the real momentum p = gamma * m * v = u * m for calculations.
+ * Here, all auxiliary quantities are equal to those used in Ripperda's article.
+ *
+ * Further references:
+ * [Higuera's article on arxiv](https://arxiv.org/abs/1701.05605)
+ * [Riperda's comparison of relativistic particle integrators](https://doi.org/10.3847/1538-4365/aab114)
+ */
+
+template<class Velocity, class Gamma>
+struct Push
```
sbastrakov, line 46: This struct template would benefit from a Doxygen comment. Either as part of the previous large one, or separately.

include/picongpu/particles/pusher/particlePusherHC.hpp
```diff
@@ -0,0 +1,139 @@
...
+    /* this is an optional extension for sub-sampling pushes that enables grid to particle interpolation
+     * for particle positions outside the super cell in one push
+     */
+    using LowerMargin = typename pmacc::math::CT::make_Int<simDim,0>::type;
+    using UpperMargin = typename pmacc::math::CT::make_Int<simDim,0>::type;
+
+    template< typename T_FunctorFieldE, typename T_FunctorFieldB, typename T_Particle, typename T_Pos >
+    HDINLINE void operator()(
+        const T_FunctorFieldB functorBField,
+        const T_FunctorFieldE functorEField,
```
sbastrakov, line 57: Where `const` should be right from the type. You are using this (current) style in the implementation already.

include/picongpu/particles/pusher/particlePusherHC.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+        /*
+         * Momentum update
+         * Notation is according to Ripperda's paper
+         */
+        // First half electric field acceleration
+        const sqrt_HC::float3_X mom_minus = precisionCast<sqrt_HC::float_X>( mom + float_X(0.5) * charge * eField * deltaT );
+
+        // Auxiliary quantitites
+        const sqrt_HC::float_X gamma_minus = gamma( mom_minus , mass );
```
sbastrakov, line 86: Optional: could make a using-alias for the `sqrt_HC::float_X / float3_X` types inside this function.

include/picongpu/particles/pusher/particlePusherHC.hpp
```diff
@@ -0,0 +1,139 @@
...
+        const MomType mom_diff2 = precisionCast<float_X>( pmacc::math::cross( mom_plus , t_vector ) );
+        const MomType mom_diff = mom_diff1 + mom_diff2;
+
+        const MomType new_mom = precisionCast<float_X>( mom_plus ) + mom_diff;
+
+        particle[ momentum_ ] = new_mom;
+
+        /*
+         * Position update
+         */
```
sbastrakov, line 121: This is weird style, perhaps just `// Position update` ?

include/picongpu/param/species.param
```diff
@@ -74,8 +74,9 @@ using UsedParticleCurrentSolver = currentSolver::Esirkepov< UsedParticleShape >;
  *
  * Defining a pusher is optional for particles
  *
- * - particles::pusher::Vay : better suited relativistic boris pusher
- * - particles::pusher::Boris : standard boris pusher
+ * - particles::pusher::HC : Higuera & Cary's relativistic pusher preserving both volume and ExB velocity
+ * - particles::pusher::Vay : Vay's relativistic pusher preserving ExB velocity
+ * - particles::pusher::Boris : Boris' relativistic pusher preserving volume
```
sbastrakov, line 79: I like this extended explanation!

include/picongpu/particles/pusher/particlePusherHC.hpp
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+
+    static pmacc::traits::StringProperty getStringProperties()
+    {
+        pmacc::traits::StringProperty propList( "name", "HC" );
+        return propList;
+    }
+};
+} // namespace particlePusherHC
```
sbastrakov, line 138: Could add a new line between end of struct and end of namespace

include/picongpu/param/species.param
steindev: Why not? It is supposed to become the new standard as it is the best of Boris' and Vay's method combined. And it is tested and proven to work.

include/picongpu/particles/pusher/particlePusherHC.hpp
steindev: Well, this line is present in every pusher and we didn't touch it. Don't ask me about its meaning. @psychocoderHPC

include/picongpu/param/pusher.param
steindev: In line 95 of `include/picongpu/particles/pusher/particlePusherHigueraCary.hpp` starting with `const sqrt_HC::float_X gamma_plus` a square root is calculated. This is (one place) where we need the high precision. The other is when we calculate `mom_diff2`.

include/picongpu/particles/pusher/particlePusherHC.hpp
steindev: Do you mean that the whole comment above should become a Doxygen comment for the `struct Push`? (I made the change to `typename`, too)

include/picongpu/particles/pusher/particlePusherHC.hpp
steindev: I don't understand what you mean. :sweat_smile:

include/picongpu/particles/pusher/particlePusherHC.hpp
steindev: I would leave it as it is for now, as it is not too long.

include/picongpu/particles/pusher/particlePusherHC.hpp
PrometheusPi: Okay

include/picongpu/param/species.param
steindev: :grinning:

include/picongpu/particles/pusher/particlePusherHC.hpp
sbastrakov: It may. But at least `struct Push` should have a doxygen comment. I do not have a preference of whether the existing comment is extended to have template parameters there, or it stays as is and a new one is added specifically for the struct.

include/picongpu/param/pusher.param
PrometheusPi: Okay - sorry, I overlooked that `math::sqrt`

include/picongpu/particles/pusher/particlePusherHC.hpp
sbastrakov: I meant the east-const style like `T_FunctorFieldB  const functorBField`. The body of the kernel already uses this style.

include/picongpu/param/species.param
PrometheusPi: We can switch this to the new standard soon. But for now I would prefer having it as a optional pusher so that we can play around with it is simulations using the `dev` branch. @ComputationalRadiationPhysics/picongpu-developers what do you thin? 

This slow approcha allows to observe possible performance penalties and glitches in the behavior of the new pusher before letting everyone else use it by default (without their explicit knowledge of this switch). 

The same will be done with the EmZ algorithm - currently [Esirkepov is still the default ](https://github.com/ComputationalRadiationPhysics/picongpu/blob/fe322db49e69b09258691c4aa781769470fe693d/include/picongpu/param/species.param#L71) despite EmZ being better.

sbastrakov: I would agree. Maybe have this PR adding a new pusher. And another PR solely to change the default. This way may also be better for release notes, as these two PRs would come as two different lines.

include/picongpu/particles/pusher/particlePusherHC.hpp
steindev: Uhh actually, only in lines that we copied from other pusher implementations `const` is written on the right hand side of the type. So this is inconsistent within the implementation. After reading a little, probably the *east-const* style is more instructive (yet a bit harder to see I think). Should I refactor to a consistent style of `const` placement and should it be the *east-const* style?

sbastrakov: In new code, I would say yes.

include/picongpu/particles/pusher/particlePusherHC.hpp
steindev: Because the original article is behind a paywall.

PrometheusPi: Okay

include/picongpu/particles/pusher/particlePusherHC.hpp
steindev: I transformed the original comment to a doxygen comment for the `struct Push` and added a description for the template parameters. Similar to the `class Solver` in the `YeePML` implementation.

sbastrakov:
I asked @psychocoderHPC and he trusts our reviews of this PR. So I assume you did not plan to add anything more to this PR @steindev and we can press the merge button now?

steindev:
> @steindev and we can press the merge button now?

*Squash and merge* probably, but YES you can do so. From my point of view everything is fine. Once we change the default pusher in a separate PR I will add a magnetic bottle test as suggested by @BeyondEspresso.

PR #3281, opened 2020-07-10
add verbose output for openPMD-api

PrometheusPi:
This pull request adds verbose output for finding and not finding the openPMD-api in CMake. 

Thanks @psychocoderHPC for your help.


PrometheusPi:
Output when openPMD-api was **not** found:
```bash
...
-- Could NOT find openPMD (missing: openPMD_DIR)
-- Could NOT find openPMD - set openPMD_DIR or check your CMAKE_PREFIX_PATH
...
```

Ouput when openPMD-api was found:
```bash
...
-- Found openPMD: /trinity/shared/pkg/filelib/openpmd/0.11.1/gcc/7.3.0/openmpi/2.1.2/lib64/cmake/openPMD
...
```

Perhaps, we do not need an `else()` branch.

PR #3283, opened 2020-07-13
Abstract PML definitions from YeePML field solver

steindev:
PMLs could be used in other field solvers, too. But to do so, the variables of the PML in `pml.param` need to be separated from the `YeePML` field solver. This PR puts the definitions of the PML variables in their own namespace `PML` and updates the `YeePML` field solver to use the variables from this namespace.

A test with laser propagation in vacuum using PML boundaries compiles without error. (I am not sure if there is one compile test in the CI using PMLs.)

### DIFF
--- a/include/picongpu/param/pml.param
+++ b/include/picongpu/param/pml.param
@@ -1,4 +1,4 @@
-/* Copyright 2019-2020 Sergei Bastrakov
+/* Copyright 2019-2020 Sergei Bastrakov, Klaus Steiniger
  *
  * This file is part of PIConGPU.
  *
@@ -19,7 +19,7 @@
 
 /** @file
  *
- * Configure the perfectly matched layer (PML).
+ * Configure the Perfectly Matched Layer absorbing boundary conditions (PML).
  *
  * To enable PML use YeePML field solver.
  */
@@ -33,11 +33,11 @@ namespace fields
 {
 namespace maxwellSolver
 {
-namespace yeePML
+namespace Pml
 {
 
-    /* The parameters in this file are only used if the field solver selected is
-     * YeePML.
+    /* The parameters in this file are only used if the field solver selected
+     * uses Perfectly Matched Layer Absorbing Boundary Conditions (PML).
      * The original paper on this approach is J.A. Roden, S.D. Gedney.
      * Convolution PML (CPML): An efficient FDTD implementation of the CFS - PML
      * for arbitrary media. Microwave and optical technology letters. 27 (5),
@@ -155,7 +155,7 @@ namespace yeePML
         0.2
     };
 
-} // namespace yeePML
+} // namespace Pml
 } // namespace maxwellSolver
 } // namespace fields
 } // namespace picongpu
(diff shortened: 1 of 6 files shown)


include/picongpu/param/pml.param
```diff
@@ -33,11 +33,11 @@ namespace fields
 {
 namespace maxwellSolver
 {
-namespace yeePML
+namespace PML
 {
 
     /* The parameters in this file are only used if the field solver selected is
-     * YeePML.
+     * uses Perfectly Matched Layer Absorbing Boundary Conditions (PML).
```
sbastrakov, line 40: I guess need to remove "is" on line 39.

steindev:
I rebased my branch and squashed my commits. Now PrometheusPi 's latest commit to the dev branch `add verbose output for openPMD-api` shows up in my history as well, see above. Is this normal, or should I do something about it?

include/picongpu/param/pml.param
steindev: Yes, thanks. Removed it.

sbastrakov:
@steindev I think this is because I used the wrong merge option for that PR of his. (Changed to squash for your other PR, and then changed "back" to the wrong one). To be sure, you drop that commit from your branch and then rebase on the current `dev`. (I guess there are other ways to achieve that as well, I would have just done it like that)

psychocoderHPC:
note: ci/gitlab/gitlab.com can be ignored, the CI is not required for a merge

steindev:
> @steindev I think this may be because I used the wrong merge option for that PR of his. (Changed to squash for your other PR, and then changed "back" to the wrong one). To be sure, you could drop that commit from your branch and then rebase your only real commit on the current `dev`. (I guess there are other ways to achieve that as well, I would have just done it like that)

Thanks that worked. I think, I also did a mistake by first rebasing and then squashing. As far as I have seen now, the other way round seems to be the common workflow.

include/picongpu/param/pml.param
```diff
@@ -33,11 +33,11 @@ namespace fields
 {
 namespace maxwellSolver
 {
-namespace yeePML
+namespace Pml
```
sbastrakov, line 36: Is this spelling what you agreed upon with @psychocoderHPC ? To be personally it's the least logical of all, but I am fine with any, just wanted to clarify

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/pml.param
```diff
@@ -33,7 +33,7 @@ namespace fields
 {
 namespace maxwellSolver
 {
-namespace yeePML
+namespace Pml
 {
 
     /* The parameters in this file are only used if field solver is YeePML.
```
sbastrakov, line 39: Perhaps you want to update this line as well, to be consistent with other changes in the comments?

include/picongpu/param/pml.param
steindev: Yes, because it is supposed to become a type (struct/class) soon anyway and than the naming does not need to be changed again.

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/pml.param
steindev: Actually not for the moment. As `YeePML` ist the only field solver that uses PMLs currently. I am just afraid that I/we will forget to change this line once there is another field solver making use of PMLs.

PR #3285, opened 2020-07-15
fix KHI for non CUDA devices

psychocoderHPC:
If KHI is compiled for non CUDA devices and the field solver `DirSplitting` is enabled the compile will fail.
`DirSplitting` requires CUDA. One case defined in `cmakeFlags` is using directional splitting.
This PR added a precompiler guard section to switch to the field solver Yee if we compile for non CUDA devices.

### DIFF
--- a/share/picongpu/examples/KelvinHelmholtz/include/picongpu/param/grid.param
+++ b/share/picongpu/examples/KelvinHelmholtz/include/picongpu/param/grid.param
@@ -33,7 +33,34 @@ namespace picongpu
 
         /** equals X
          *  unit: meter */
+
+#ifndef PARAM_FIELDSOLVER
+#   define PARAM_FIELDSOLVER Yee
+#endif
+
 #define DirSplitting 1
+
+#if( PMACC_CUDA_ENABLED != 1 )
+/* DirSplitting is only available for the CUDA accelerator.
+ * Automatically switch back to Yee when we compile for non CUDA accelerator.
+ */
+#   if (PARAM_FIELDSOLVER == 1)
+#       warning "DirSplitting is only for CUDA available, solver will be set to Yee."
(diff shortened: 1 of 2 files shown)


share/picongpu/examples/KelvinHelmholtz/include/picongpu/param/grid.param
```diff
@@ -33,7 +33,34 @@ namespace picongpu
...
+#   define PARAM_FIELDSOLVER Yee
+#endif
+
 #define DirSplitting 1
+
+#if( PMACC_CUDA_ENABLED != 1 )
+/* DirSplitting is only available for the CUDA accelerator.
+ * Automatically switch back to Yee when we compile for non CUDA accelerator.
+ */
+#   if (PARAM_FIELDSOLVER == 1)
```
PrometheusPi, line 47: Why do you check `PARAM_FIELDSOLVER` to be equal to `1`?  Shouldn't that be string equal to `Yee` or `DirSplitting` or etc.?

psychocoderHPC: It is not possible to perform string comparism with precompiler. This is a workaround we used already in the past. 
We will work on removing directional splittung from the kHI example. Currently it ia our way to test different components of PIC. 
The mid term goal is to simplify the examples and move such tests to seperate input sets so that the user will not be co fused by such uglyprecompiler hacks.

PrometheusPi:
@psychocoderHPC What is the status with the `ci/gitlab/gitlab.com` test?

sbastrakov:
@PrometheusPi I believe it is in [that comment](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3286#issuecomment-658985376)

PrometheusPi:
@sbastrakov Thanks for the reference

Issue #3277, opened 2020-07-03
Pusher selection at run time / chains of pushers.


Anton-Le:
Similar to #3185 but more general:

It would be nice to have the ability to switch pushers using command-line arguments.

_Simple use case:_
Accelerate a particle bunch using the acceleration pusher,
propagate it for a while in vacuum using e.g. Vay and 
then simulate a bunch-target interaction using yet another
pusher (e.g. Landau-Lifschitz Radiation Reaction).

In an ideal case a user would provide a list/vector of pushers in the `.param`
files that could be used in the simulation. At run time one would then provide
a start/stop times for the pusher
 `--<species>_<pusher>.start <simulationStep> --<species>_<pusher>.stop <simulationStep>`
(with one pusher being the default fallback).

Issue #3185 would be a first step towards this.

sbastrakov:
So now #3185 is implemented. With the new composite pusher, it is possible to combine two pushers and switch between those with a user-provided functor. Now, one or both of the combined pushers can be composite as well, which theoretically allows chaining any fixed amount. However, that would be really awkward to do in practice. One can implement a generalization which takes a (type-)list of pushers and has a functor to select one of those. I could do it at some point, but do not have time for it right now.

Issue #3185, opened 2020-02-25
acceleration pusher - switch to other pusher 


PrometheusPi:
It would be very convenient if we could add a template argument to the acceleration pusher that allows switching from the acceleration pusher to the "normal" selected other pusher. 

Currently, the acceleration pusher has a [end time](yhttps://github.com/ComputationalRadiationPhysics/picongpu/blob/5725049b384fccca384d74f18a31ace066d25e36/include/picongpu/param/pusher.param#L53) but no followup pusher is set. 

cc'ing @anton-Le, @steindev, @BeyondEspresso   

motivated by #3169

sbastrakov:
I think this is rather straightforward to do. 

As a thought that popped up during thinking of it. Probably, a reasonable default for this prospective new template parameter is to do nothing. For probes we express it using  `void` as`particlePusherProbe::ActualPusher`, however this requires [specializing for it separately](https://github.com/ComputationalRadiationPhysics/picongpu/blob/5725049b384fccca384d74f18a31ace066d25e36/include/picongpu/particles/pusher/particlePusherProbe.hpp#L101). Perhaps a more consistent solution for both probes and this one is to introduce a new `None` particle pusher analogous to the `None` Maxwell's solver. Then it follows all the pusher interfaces and thus there would be no need for a special corner case, like `void`.

sbastrakov:
I am working on a concept for pusher chaining. First, a rather simplistic one, will provide a PR with the draft tomorrow.

sbastrakov:
Okay, today turned out too optimistic. But tomorrow for sure :)

sbastrakov:
This is implemented by #3288.

PR #3288, opened 2020-07-15
Add support for composite pushers

sbastrakov:
Composite pushers allow to switch to one of the underlying pushers. So far it is only for two pushers (however, they could be composite as well), but can be reasonably extended to any amount later (maybe even in this PR, will see).

Note that the decision is done by a user-provided functor. So the easy use is just decide based on the current time iteration. And realistically that's about all one could do right now. But theoretically, this can be used to take into account any globally accessible data, or command-line parameters, once there is a way to access that info from the functor.

Example usage in `species.param`:
```cpp
// Old: always Boris
// using UsedParticlePusher = particles::pusher::Boris;
// New: Boris for 100 time iterations and then Vay
using UsedParticlePusher = particles::pusher::Composite<
    particles::pusher::Boris,
    particles::pusher::Vay,
    particles::pusher::CompositeBinarySwitchActivationFunctor< 100 > /* this is a preset, a user could write their own functor and provide here */
>;
```

ToDo:
- [x] figure out proper traits use, now it is not everywhere correct due to the convoluted relations of things
- [x] decide on 0- vs. 1-based indexing. 1-base currently, as this is more logical with `FirstPusher` and `SecondPusher` names
- [x] test on Hemera, so far only made it compile locally

Implements feature request #3185, and makes some progress towards #3277

### DIFF
--- a/include/picongpu/param/pusher.param
+++ b/include/picongpu/param/pusher.param
@@ -98,6 +98,14 @@ namespace picongpu
 #if(SIMDIM==DIM3)
         struct Axel;
 #endif
+        template<
+            typename T_FirstPusher,
+            typename T_SecondPusher,
+            typename T_ActivationFunctor
+        >
+        struct Composite;
+        template< uint32_t T_switchTimeStep >
+        struct CompositeBinarySwitchActivationFunctor;
     } // namespace pusher
     } // namespace particles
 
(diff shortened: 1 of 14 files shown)


sbastrakov:
cc @psychocoderHPC @PrometheusPi @Anton-Le please have a look at the concept when you have time

sbastrakov:
Update: there was a compile issue that did not occur on my local machine. Now fixed, my test on Hemera seems to have worked as expected with the `UsedParticlePusher` defined as in the PR description. (I added additional output to see which pusher was used).

Anton-Le:
After testing with a PWFA acceleration and impact stage the switch appears to work.

Issue #3278, opened 2020-07-03
Warnings in the openPMD plugin


sbastrakov:
While working on #3275 I am trying the openPMD plugin on Hemera, using the recently installed module as in #3270. There are a few warnings produced during compilation for the `gpu` partition (perhaps for others as well, did not try).

It is not super urgent, but we should fix the warnings. @franzpoeschel could you take a look when you have time?

franzpoeschel:
I will look into it, but don't know yet whether I will get to it next week. Feel free to ping me again in case ;)

franzpoeschel:
>using the recently installed module

What version is that module? The openPMD plugin requires the latest `dev` of the openPMD API, so I am a bit surprised building worked at all. I do remember getting these warnings in earlier times, but if I remember correctly they have been fixed in the openPMD API in the meantime. I did a quick compilation using the latest `dev` of openPMD and saw no warnings caused by the openPMD plugin: [pic_build_log.txt](https://github.com/ComputationalRadiationPhysics/picongpu/files/4870483/pic_log.txt)

sbastrakov:
For the openPMD-API module I used settings from #3270. Looks like v 0.11.1.
It did build (with warnings) and work for writing a checkpoint and later loading from it. However, my run had no particles, only fields.

franzpoeschel:
PR #3289 should fix the warnings.

> It did build (with warnings) and work for writing a checkpoint and later loading from it. However, my run had no particle species, only fields. So I guess the templates for particles were never instantiated.

How did you determine that your run had no particles? Were there no particles in the data dumps?
Can you share a way to reproduce the issue? I just ran a quick test with the LaserWakefield example that still had its particles after restarting, so restarting should work in general, but there might be a template issue as you said.

sbastrakov:
Sorry @franzpoeschel , I did not express it correctly. I meant that my PIConGPU setup had no particle species defined (so not just no particles, but also no particles species whatsoever), so I assume the checkpoint/output for particles was never instantiated and so never compiled, only the field part did. And so it might have worked just because of that, even tho in the particles part there may have been some inconsistencies, etc.

franzpoeschel:
So, if I understand you correctly, you did not test the full thing? 
Since I based my builds for PR #3289 on a simulation that has fields and particles, is this issue now resolved or are there further fixes to apply?

sbastrakov:
[...]

PR #3289, opened 2020-07-15
Fix warnings in the openPMD plugin

franzpoeschel:
See https://github.com/ComputationalRadiationPhysics/picongpu/issues/3278

1. A variable was declared and defined equivalently in two scopes
2. Rename `AccessType` (deprecated) → `Access`

### DIFF
--- a/include/picongpu/plugins/openPMD/openPMDWriter.def
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.def
@@ -110,7 +110,7 @@ namespace openPMD
                                           window to local domain */
 
         ::openPMD::Series &
-        openSeries( ::openPMD::AccessType at );
+        openSeries( ::openPMD::Access at );
 
         void
         closeSeries();
--- a/include/picongpu/plugins/openPMD/openPMDWriter.hpp
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.hpp
@@ -132,7 +132,7 @@ namespace openPMD
     }
 
     ::openPMD::Series &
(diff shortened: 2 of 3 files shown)


sbastrakov:
Thanks @franzpoeschel ! Just a silly question, is this newer `Access` available for the openPMD-api minimum required versions. or do we need to update those?

franzpoeschel:
It is available only in the upcoming version 0.12, just like the configuration of backends via JSON which the plugin also uses. So we will need to update that upon release of version 0.12 anyway.
For some reason, the Hemera module was fine with both, could it be that it has been accidentally been based on the `dev` branch of the openPMD API?

sbastrakov:
I think with Hemera admins everything is possible. Is there a way to check it externally, e.g. by asking for version or something like that.

franzpoeschel:
I just looked it up, it's the current dev:
```
> pwd
/trinity/shared/pkg/filelib/openpmd/0.11.1-cuda102/gcc/7.3.0/openmpi/2.1.2-cuda102/include/openPMD
> cat version.hpp 
[…]

/** version of the openPMD-api library (compile-time)
 * @{
 */
#define OPENPMDAPI_VERSION_MAJOR 0
#define OPENPMDAPI_VERSION_MINOR 12
#define OPENPMDAPI_VERSION_PATCH 0
#define OPENPMDAPI_VERSION_LABEL "dev
```

psychocoderHPC:
please rebase against the dev branch to run the gitlab CI.

PR #3298, opened 2020-07-22
fix clang cuda compile

psychocoderHPC:
- fix clang-cuda boost compatibility
  - The include workaround is required for `main.cpp` too. Move the workaround include to `boost_workaround.hpp`
- Fix early evaluation of a static assert.
  ```
  pmacc/math/vector/Vector.hpp:583:9: error: static_assert failed "Vector_dim_0_create_cannot_be_called"
        PMACC_CASSERT_MSG(Vector_dim_0_create_cannot_be_called, false);
  ```
- fix clang-cuda incompatibility with OpenMP
  - If clang is used as CUDA compiler and OpenMP is enabled the compilation will crash. It is a general clang-cuda bug which will hopfully be fixed with clang 11.

### DIFF
--- a/include/picongpu/boost_workaround.hpp
+++ b/include/picongpu/boost_workaround.hpp
@@ -35,3 +35,8 @@
 #   include <boost/config/detail/select_compiler_config.hpp>
 #   define __CUDACC__
 #endif
+/* workaround for compile error with clang-cuda
+ * boost/type_traits/is_base_and_derived.hpp:142:25: error: invalid application of 'sizeof' to an incomplete type 'boost::in_place_factory_base'
+ *   BOOST_STATIC_ASSERT(sizeof(B) != 0);
+ */
+#include <boost/optional/optional.hpp>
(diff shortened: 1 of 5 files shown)


include/picongpu/boost_workaround.hpp
```diff
@@ -35,3 +35,8 @@
 #   include <boost/config/detail/select_compiler_config.hpp>
 #   define __CUDACC__
 #endif
+/* workaround for compile error with clang-cuda
+ * boost/type_traits/is_base_and_derived.hpp:142:25: error: invalid application of 'sizeof' to an incomplete type 'boost::in_place_factory_base'
+ *   BOOST_STATIC_ASSERT(sizeof(B) != 0);
+ */
+#include <boost/optional/optional.hpp>
```
sbastrakov, line 42: I think it's not very clear what *is* the workaround? that we include `boost/optional/optional` here?

Issue #3296, opened 2020-07-21
PML boundary for the Lehe field solver


Anton-Le:
The PML boundary is currently implemented for the Yee solver only (`YeePML`).

An implementation for the Lehe solver would be very welcome, since PML does not fix the numerical Cherenkov radiation of the Yee solver.

Consider for instance the acceleration of a prolongate electron beam with absorbing boundaries (32 on each side) and PML boundaries ( 8 on each side). In the following image the transverse Lorentz force density is shown in the reference frame of the bunch (note the symmetric *logarithmic* scale):
![LorentzForceDensityComparison_T111k_AccelStage_0 5 0_ABSORBING_VS_PML](https://user-images.githubusercontent.com/50339174/88052278-50202300-cb5a-11ea-892a-4d6f0d3a1b62.png)

The bunch is added for reference and its color scale is not shown! The sizes of the domain is identical.

In the case of the absorbing boundary and Lehe solver we see a strong build-up of transverse forces after the bunch and a (very) weak periodic structure in front of the bunch.
In the case of PML boundary and Yee solver the space in front of the bunch is essentially force-free but the Cherenkov cones are extremely pronounced.

Seeing as this is supposed to be the seed for an instability simulation neither case is satisfactory.

Having the best of both worlds - the minimal-to-non-existent reflections of PML and the suppressed numerical Cherenkov radiation from Lehe solver would be great!

sbastrakov:
So theoretically it should be possible to instantiate `YeePML` with the E curl operator from the Lehe solver. I will try how it works in practice and report here.

In case it does, however I would suggest to rename the existing templated `Yee` and `YeePML` into `Fdtd` and `FdtdPML` as they implement the generic finite-difference time-domain scheme. And then have `Yee/YeePML/Lehe/LehePML` as template specializations with proper curl operators.

sbastrakov:
Okay, it is not exactly possible to instantiate as is, but should be not very hard to enable it. (So that at least it runs, we will see how well it works then)

sbastrakov:
Implemented with #3301, closing the issue.

PR #3301, opened 2020-07-23
Add support for Lehe solver with PML absorber

sbastrakov:
Currently the branch has some version of this combination, that may work, but requires more testing. The code changes are also occasionally hacky, I will refactor if the approach works.
To enable the combination, use this branch and set
```cpp
using Solver = maxwellSolver::YeePML<
        CurrentInterpolation,
        maxwellSolver::lehe::CurlE< maxwellSolver::lehe::CherenkovFreeDirection_Y >
    >;
```
in `fieldSolver.param`.

In case this works, it solves #3296.
@Anton-Le , could you try running your test from that issue with my branch?

### DIFF
--- a/include/picongpu/fields/MaxwellSolver/LehePML/LehePML.def
+++ b/include/picongpu/fields/MaxwellSolver/LehePML/LehePML.def
@@ -1,0 +1,60 @@
+/* Copyright 2013-2020 Axel Huebl, Heiko Burau, Rene Widera, Remi Lehe,
+ *                     Sergei Bastrakov
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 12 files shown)


Anton-Le:
Purely by visual inspection I would say that the results are good:

LorentzForceDensityComparison_T111k_AccelStage_0 5 0_PML_LEHE_vs_YEE

sbastrakov:
Okay, so it seems like the combination makes sense. I will work on bringing it in. However, this PR currently achieves it in a somewhat hacky way, and I now made #3309 to do some ground work for it, and will rebase this PR on top. ETA end of this week.

sbastrakov:
Now I rebased using #3317 which did all the heavy lifting.

I made a backup for the old hacky way here, just in case something does not work or we need to go back: https://github.com/sbastrakov/picongpu/tree/backup-topic-fieldSolverLehePML .

sbastrakov:
@Anton-Le the branch should be about ready. Could you re-run your test with the current state of this branch using the following solver in `fieldSolver.param`:

```cpp
using Solver = maxwellSolver::LehePML<
    CurrentInterpolation,
    maxwellSolver::lehe::CherenkovFreeDirection_Y
>;
```

PR #3303, opened 2020-07-23
Fix implementation of Lehe curl for CherenkovFreeDirection_X

### LINKED ISSUE #3302: Lehe curl implementation for CherenkovFreeDirection_X does not work
Note: this is only for `CherenkovFreeDirection_X`, while we realistically only use it for `CherenkovFreeDirection_Y` (the default) and I do not see any issues with the latter.

There are at least two errors in the implementation for `CherenkovFreeDirection_X`:

- The values accessed by the differential operators do not correspond to the `Margin`s declared: it uses index [-2] with the lower margin of 1. I believe the margins are actually correct, and it's the finite difference expressions that are wrong. This now causes crashes with accessing wrong memory.
- The order of components is wrong: the implementation is written as if `mem` access order is `[x][y][z]` while it's the reverse.

Generally, it is a little weird that the implementation is mostly similar to Y, but different in these places. I guess it was modified or fixed for Y at some point, but not for X. However, those changes must have been too old to track via git. The Y implementation has some weirdness with double sign reversals for both indices and terms in differences, also the margins are overly conservative for x, z, but we re-checked today with the paper and it looks correct.

I will provide a fix for X. We also discussed with @psychocoderHPC how it can be unified together, and perhaps even more unified with the standard Yee curl, but that would be further work and not part of the fixing PR.

sbastrakov:
Fixes #3302

### DIFF
--- a/include/picongpu/fields/MaxwellSolver/Lehe/Curl.hpp
+++ b/include/picongpu/fields/MaxwellSolver/Lehe/Curl.hpp
@@ -123,56 +123,70 @@ namespace lehe
                 - float_X( 2.0 ) * beta_zy * isDir_z
                 - float_X( 3.0 ) * delta_dir0 * isDir_z;
 
+            /* Note: the rest of this function is made by copying and
+             * adapting the implementation for Y, including the sign reversals in
+             * indices, and reversals of terms for finite-difference derivatives
+             */
 
+            // Typedef an accessor to access mem[z][y][x]
+            // in (x,y,z) order :)
+            typedef DataSpace<DIM3> Space;
+
+            // actually computes -curl_x(E) = (-dEz/dy) - (-dEy/dz)
             const float_X curl_x
                 = (
-                    alpha_y * ( mem[0][0][0].z( ) - mem[0][-1][0].z( ) )
-                    + beta_yx * ( mem[1][0][0].z( ) - mem[1][-1][0].z( ) )
-                    + beta_yx * ( mem[-1][0][0].z( ) - mem[-1][-1][0].z( ) )
-                    ) * reci_dy
+                    alpha_y * ( mem(Space(0,0,0)*(-1)).z( ) - mem(Space(0,-1,0)*(-1)).z( ) )
+                    + beta_yz * ( mem(Space(0,0,1)*(-1)).z( ) - mem(Space(0,-1,1)*(-1)).z( ) )
+                    + beta_yz * ( mem(Space(0,0,-1)*(-1)).z( ) - mem(Space(0,-1,-1)*(-1)).z( ) )
+                    + beta_yx * ( mem(Space(1,0,0)*(-1)).z( ) - mem(Space(1,-1,0)*(-1)).z( ) )
+                    + beta_yx * ( mem(Space(-1,0,0)*(-1)).z( ) - mem(Space(-1,-1,0)*(-1)).z( ) )
+                ) * reci_dy
                 - (
-                    alpha_z * ( mem[0][0][0].y( ) - mem[0][0][-1].y( ) )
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -65,7 +65,7 @@ install:
   - spack clean -a
   - source /etc/profile &&
     source $SPACK_ROOT/share/spack/setup-env.sh
-  - spack load cmake
+  - spack load cmake@3.15.0%gcc@7.5.0
```
PrometheusPi, line 68: why did you introduce this change with this pull request?

include/picongpu/fields/MaxwellSolver/Lehe/Curl.hpp
```diff
@@ -123,56 +123,70 @@ namespace lehe
...
+            typedef DataSpace<DIM3> Space;
 
+            // actually computes -curl_x(E) = (-dEz/dy) - (-dEy/dz)
             const float_X curl_x
                 = (
-                    alpha_y * ( mem[0][0][0].z( ) - mem[0][-1][0].z( ) )
-                    + beta_yx * ( mem[1][0][0].z( ) - mem[1][-1][0].z( ) )
-                    + beta_yx * ( mem[-1][0][0].z( ) - mem[-1][-1][0].z( ) )
-                    ) * reci_dy
+                    alpha_y * ( mem(Space(0,0,0)*(-1)).z( ) - mem(Space(0,-1,0)*(-1)).z( ) )
```
PrometheusPi, line 138: What does this double minus `- mem...` and `*(-1)` result in?

include/picongpu/fields/MaxwellSolver/Lehe/Curl.hpp
```diff
@@ -123,56 +123,70 @@ namespace lehe
...
+                    + beta_yx * ( mem(Space(1,0,0)*(-1)).x( ) - mem(Space(1,-1,0)*(-1)).x( ) )
+                    + beta_yx * ( mem(Space(-1,0,0)*(-1)).x( ) - mem(Space(-1,-1,0)*(-1)).x( ) )
                     ) * reci_dy;
 
-            return float3_X( curl_x, curl_y, curl_z );
-
-            //return float3_X(diff(mem, 1).z() - diff(mem, 2).y(),
-            //                diff(mem, 2).x() - diff(mem, 0).z(),
-            //                diff(mem, 0).y() - diff(mem, 1).x());
+            // revert the signs to get the actual curl
```
PrometheusPi, line 188: Thanks for adding the comments?

.travis.yml
PrometheusPi: was answered with #3305

sbastrakov: Yes, was trying out the fix yesterday, as this PR encountered the issue #3304. But you are right, I should have marked the PR as WIP then.

include/picongpu/fields/MaxwellSolver/Lehe/Curl.hpp
sbastrakov: I asked @psychocoderHPC and it seems no one remembers, this code is really old. We checked together with the paper and the implementation for Y (where these minuses came from) is correct, albeit weirdly formulated for an unknown reason.

sbastrakov: I think i will provide some refactoring, as this would also be fitting for the Lehe + PML solver. But as a separate PR, this is just a bug fix.

.travis.yml
sbastrakov: Rebased.

PrometheusPi:
Sorry for confusing `?` with `!`.

sbastrakov:
@PrometheusPi could you press the merge button?

PR #3313, opened 2020-07-30
hip test branch

psychocoderHPC:
This is a test if the HIP changes together with mallocMC (alpaka version) will break PIConGPU.


psychocoderHPC:
HIP support was officially merged with #3456

PR #3315, opened 2020-07-31
fix missing override for virtual functions

psychocoderHPC:
Depending of the compiler it can result in compile errors.

- add missing `override`

### DIFF
--- a/include/pmacc/memory/buffers/ExchangeIntern.hpp
+++ b/include/pmacc/memory/buffers/ExchangeIntern.hpp
@@ -222,24 +222,24 @@ namespace pmacc
 
         }
 
-        virtual HostBuffer<TYPE, DIM>& getHostBuffer()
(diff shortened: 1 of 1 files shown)


include/pmacc/memory/buffers/ExchangeIntern.hpp
```diff
@@ -222,24 +222,24 @@ namespace pmacc
 
         }
 
-        virtual HostBuffer<TYPE, DIM>& getHostBuffer()
+        virtual HostBuffer<TYPE, DIM>& getHostBuffer() override
```
sbastrakov, line 225: I think now with `override` there is no need in `virtual`. Technically it was not needed even before, but then it shows the function was virtual, however now override already guarantees it.

psychocoderHPC: I will remove it

PR #3320, opened 2020-08-03
Refactor margin traits for field derivatives

sbastrakov:
Previously, the margins were sometimes overly conservative. E.g. Lehe solver defined upper margin for derivative, and therefore curl, as (2, 2, 2), while it actually used (1, 2, 1) with 2 along the Cherenkov-free direction. Move the margins to be defined for functors, not tags, and put precise values. Make use of recently introduced #3319 compile-time basis vectors to express the margins. Adjust curl margin definition accordingly, to be max of derivative margins.

This PR will allow easier uniform implementation of `Lehe` / `LehePML` solvers for both 3D and 2D. Practically speaking, the current 3D-only implementation will only need a slight adjustment to work for 2D as well. This will be done as a follow-up PR.

### DIFF
--- a/include/picongpu/fields/MaxwellSolver/Lehe/Derivative.hpp
+++ b/include/picongpu/fields/MaxwellSolver/Lehe/Derivative.hpp
@@ -74,6 +74,40 @@ namespace lehe
         T_direction
     >
     {
+    private:
+
+        //! Internally used derivative functor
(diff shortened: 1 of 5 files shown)


sbastrakov:
@psychocoderHPC tests passed, ready for review.

PR #3323, opened 2020-08-07
reduce example complexity

psychocoderHPC:
The PR is changing cmake flags in the examples. The feature to compile different setups by changing `cmakeFlags` in the examples is useful for parameter scans and was from us hijacked to compile different solver in our CI.
This PR reducing the free parameter for a example to a useful minimum, for the corresponding test.

A new folder `tests` is introducing compile tests. We do not aim to test each possible combination for all solvers, with shapes, ...
We need to be realistic what we can test within the CI. 
Later PRs will introduce unit tests to cover more parts of PIConGPU/PMacc

# Incompatibilities

Old files `cmakeFlag` from the examples are not compatible with this PR. The default case (zero) is always running but other cases can be broken. Broken means that e.g. changing the field solver will have no effect.

# Review

Please review the PR commit wise. Each example I changed has its own commit.


psychocoderHPC:
can be merged

PR #3328, opened 2020-08-18
Fix wrong variable name in the LWFA setup

sbastrakov:
A typo in the name led to it not being resolved and therefore the `.cfg` files were broken. Notably, that includes the 1.cfg for LWFA which we use in the manual (and that's how it was discovered). This was introduced after 0.5.0, so does not affect the last release.


n01r:
Also thanks to @jtrojok for finding the bug. :+1:

sbastrakov:
@n01r the tests passed, could you press the merge button?

Issue #3329, opened 2020-08-18
Crash in dumpOneStep() on Hemera defq


sbastrakov:
At least with current `dev` on the LWFA example with `1.cfg` there is a segmentation fault.
Clarifying the details now, but the crash is at least consistent.

On k20 or gpu partition I did not observe any crashes. So the issue seems CPU-specific (be it code or environment)

sbastrakov:
After investigating, the crash happens when the `phaseSpace` plugin is enabled. Removing it from the LWFA `.cfg` file makes it run.

cc @jtrojok

sbastrakov:
I do not know yet what exactly crashes there, will continue working on it tomorrow.

sbastrakov:
Currently I've found out the crash happens when doing an assignment of device buffer to a host one [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/ec8c3ae210028c1b1404ce2da2684c89bfe53e09/include/picongpu/plugins/PhaseSpace/PhaseSpace.tpp#L276). The sizes seem correct, looking into what goes wrong there.

sbastrakov:
Okay, so the issue (or one of issues at least) is that this plugin tries to always allocate 32 KB of shared memory. While for CPU backends by default 30 KB are available.

PR #3332, opened 2020-08-19
Fix unsafe access to a vector in cuSTL MPI reduce

sbastrakov:
Variable `new_ranks` can be empty, and so `new_ranks.front()` is incorrect without a bounds check. This situation occurs, for example, if there is a single process and `setThisAsRoot` is false.
So add a bounds check before the unsafe code.

This was found during investigation of #3329, as it crashed on my machine, and is an actual bug that this PR fixes. But this is not what caused it on Hemera.

### DIFF
--- a/include/pmacc/cuSTL/algorithm/mpi/Reduce.tpp
+++ b/include/pmacc/cuSTL/algorithm/mpi/Reduce.tpp
@@ -75,16 +75,18 @@ Reduce<dim>::Reduce(const zone::SphericZone<dim>& p_zone, bool setThisAsRoot) : 
         if(i == myWorldId) this->m_participate = true;
     }
 
-    MPI_Group world_group = MPI_GROUP_NULL;
-    MPI_Group new_group = MPI_GROUP_NULL;
-
     // avoid deadlock between not finished pmacc tasks and mpi blocking collectives
     __getTransactionEvent().waitForFinished();
-    MPI_CHECK(MPI_Comm_group(MPI_COMM_WORLD, &world_group));
-    MPI_CHECK(MPI_Group_incl(world_group, new_ranks.size(), &(new_ranks.front()), &new_group));
-    MPI_CHECK(MPI_Comm_create(MPI_COMM_WORLD, new_group, &this->comm));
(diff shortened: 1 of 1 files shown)


include/pmacc/cuSTL/algorithm/mpi/Reduce.tpp
```diff
@@ -80,11 +80,13 @@ Reduce<dim>::Reduce(const zone::SphericZone<dim>& p_zone, bool setThisAsRoot) :
...
     // avoid deadlock between not finished pmacc tasks and mpi blocking collectives
     __getTransactionEvent().waitForFinished();
-    MPI_CHECK(MPI_Comm_group(MPI_COMM_WORLD, &world_group));
-    MPI_CHECK(MPI_Group_incl(world_group, new_ranks.size(), &(new_ranks.front()), &new_group));
-    MPI_CHECK(MPI_Comm_create(MPI_COMM_WORLD, new_group, &this->comm));
-    MPI_CHECK(MPI_Group_free(&new_group));
-    MPI_CHECK(MPI_Group_free(&world_group));
+    if(new_ranks.size())
+    {
+        MPI_CHECK(MPI_Group_incl(world_group, new_ranks.size(), &(new_ranks.front()), &new_group));
```
n01r, line 85: Why did you delete `MPI_CHECK(MPI_Comm_group(MPI_COMM_WORLD, &world_group));` in the line before?

sbastrakov: Ah, good catch. My mistake, it should stay in place.

sbastrakov: Fixed and rebased. Also moved declarations of those two variables inside the `if` statement as they are only used there.

n01r: That would have been my next question ^^

n01r (approved):
Alright, looks good :)

sbastrakov:
@n01r could you press the button?

PR #3333, opened 2020-08-20
Fix shared memory size in the PhaseSpace plugin

### LINKED ISSUE #3329: Crash in dumpOneStep() on Hemera defq

sbastrakov:
The previous value was incorrect for CPUs which caused crashes #3329. This is just a quick patch fix that replaces the old hard-coded value with a better one, not a proper check for available memory.

Fixes #3329

### DIFF
--- a/include/picongpu/plugins/PhaseSpace/PhaseSpace.hpp
+++ b/include/picongpu/plugins/PhaseSpace/PhaseSpace.hpp
@@ -262,7 +262,12 @@ namespace picongpu
             bmpl::int_<0>,
             bmpl::max<bmpl::_1, bmpl::_2>
             >::type SuperCellsLongestEdge;
-        static constexpr uint32_t maxShared = 32*1024; /* 32 KB */
+        /* Note: the previously used 32 KB shared memory size is not correct
+         * for CPUs, as discovered in #3329. As a quick patch, slightly reduce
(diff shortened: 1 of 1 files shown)


n01r (approved):
Nice find!

You wrote in the comment up there "quick fix". What would you suggest as the "real fix"? I think nobody will stumble over this for quite a while, again.

sbastrakov:
@n01r good point, I should have clarified.

I think generally this pattern of use is suboptimal. It did not just have the wrong (CUDA-only) number hardcoded, but the way it is all done makes it hard to change to a more generic implementation. Ideally, I would prefer asking the device in use about shared memory size (so it has to be run-time, not compile-time information) and then computing a valid dynamic shared memory size based on it, and allocating it properly. However, I did not see any easy way to change to this pattern, hence simply adjusted the hard-coded constant to at least work on CPUs as well.

n01r:
Alright, so more generic memory-management for plugins should be noted on the agenda but I'm gonna merge this now because it's an improvement to an otherwise broken case.

PR #3338, opened 2020-09-01
Add AOFDTD field solver

steindev:
# Method
The AOFDTD field solver exploits the arbitrary-order finite-difference time-domain method to advance the fields on a staggered grid. In essence, it is a higher-order version of Yee's method using more neighbors to approximate the spatial derivative of fields with finite differences. The order of the solver is equal to twice the number of neighbors used (to one side). The lowest order AOFDTD solver uses one neighbor to each side. This lowest order solver is thus of order two and is equal to Yee's method.

Advantages of the higher order method are found in its dispersion properties. Compared to Yee's method, the phase velocity of light propagating on the grid has smaller dependence on propagation direction and wavelength sampling for higher order schemes.
The following picture shows the deviation of phase velocity from the speed of light in dependence of the angle ϕ enclosed by the propagation direction and the x-axis on a 2D rectangular grid for a wave sampled 3 times per wavelength and a time step chosen close to the Courant condition (`S = c Δt / Δx < 1/sqrt(2)`)
image

Obviously, higher order solvers show inverse dispersion, i.e. light travels faster than the speed of light. But this effect decreases with smaller time steps and denser spatial sampling, as shown in the following plots for ϕ=45° and 0°
image
image

image
image

# Implementation
Basically implements a new `ForwardDerivative` and `BackwardDerivative` algorithm used by the `Curl` implementation to calculate spatial derivatives.

Also updates texts in `fieldSolver.param` to point out differences between the solvers.

# Tests
* Laser propagation with Yee and AOFDTD 2nd order (which is equal to Yee scheme): identical results
* Laser propagation with Yee and AOFDTD 8th order: AOFDTD has 7% higher amplitude and 10% more total laser pulse energy

# References
* M Ghrist, *High-Order Finite Difference Methods for Wave Equations*, PhD thesis (2000), Department of Applied Mathematics, University of Colorado
* H Vincenti et al, *Detailed analysis of the effects of stencil spatial variations with arbitrary high-order finite-difference Maxwell solver*, Appendix A, doi:10.1016/j.cpc.2015.11.009

# ToDo:
- [x] Add Documentation
- [x] Provide test results
- [x] Refactor implementation of weights calculation to perform this at compile time
- [x] Update `struct StringProperties` to comply with openPMD standard
- [x] Delete superfluous file `Derivative.hpp.OLD` :facepalm:


include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.def
```diff
@@ -0,0 +1,96 @@
+/* Copyright 2013-2020 Klaus Steiniger, Sergei Bastrakov
```
psychocoderHPC, line 1: please correct the copyright date, only 2020

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp
```diff
@@ -0,0 +1,312 @@
...
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/simulation_defines.hpp"
+#include "picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.def"
+#include "picongpu/fields/differentiation/Traits.hpp"
+
+#if 0
```
psychocoderHPC, line 26: please remove this in the final code

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp
```diff
@@ -0,0 +1,312 @@
...
+                ( data( upperIndex ) - data( lowerIndex ) );
+
+            // Compute next finite differences according to order
+            for ( uint32_t l = 1; l < T_neighbors; ++l )
+            {
+                lowerIndex -= indexIncrement;
+                upperIndex += indexIncrement;
+
+                finiteDifference += weights[l] *
+                ( data( upperIndex ) - data( lowerIndex ) );
```
psychocoderHPC, line 140: please indent once, see line 131

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp
```diff
@@ -0,0 +1,312 @@
...
+     * of computation.
+     * 
+     * @tparam T_neighbors Number of neighbors used to calculate
+     *                     the derivative from finite differences.
+     *                     Order of derivative approximation is
+     *                     2 * T_neighbors
+     * 
+     * @tparam T_direction direction to take derivative in, 0 = x, 1 = y, 2 = z
+     */
+    template<
```
psychocoderHPC, line 172: I suggest to avoid copy past of the full class by using a helper class.

```C++
namespace detail
{    
    template<
        uint32_t T_lowerMargin,
        uint32_t T_upperMargin,
        uint32_t T_neighbors,
        uint32_t T_direction
    >
    struct DerivativeFunctor
    {
        //! Lower margin
        using LowerMargin = typename pmacc::math::CT::mul<
            typename pmacc::math::CT::make_Int<
                simDim,
                T_neighbors
            >::type,
            typename pmacc::math::CT::make_BasisVector<
                simDim,
                T_direction,
                int
            >::type
        >::type;

        //! Upper margin
        using UpperMargin = typename pmacc::math::CT::mul<
            typename pmacc::math::CT::make_Int<
                simDim,
                T_neighbors - 1
            >::type,
            typename pmacc::math::CT::make_BasisVector<
                simDim,
                T_direction,
                int
            >::type
        >::type;

        /** Return derivative value at the given point
         *
         * @tparam T_DataBox data box type with field data
         * @param data position in the data box to compute derivative at
         */
        template< typename T_DataBox >
        HDINLINE typename T_DataBox::ValueType operator()(
            T_DataBox const & data ) const
        {
            // Define shorthand type to access DataBox
            using Index = pmacc::DataSpace< simDim >;

            // Define indice vectors for data access
            auto lowerIndex = -pmacc::math::basisVector<
                Index,
                T_direction
            >(); // Vector initialized with zeros

            auto upperIndex = Index{};

            auto const indexIncrement = pmacc::math::basisVector<
                Index,
                T_direction
            >();

            AOFDTDWeights< T_neighbors > const weights{};

            // shortest distance finite difference as initial value
            auto finiteDifference = weights[0] *
                ( data( upperIndex ) - data( lowerIndex ) );

            // Compute next finite differences according to order
            for ( uint32_t l = 1; l < T_neighbors; ++l )
            {
                lowerIndex -= indexIncrement;
                upperIndex += indexIncrement;

                finiteDifference += weights[l] *
                ( data( upperIndex ) - data( lowerIndex ) );
            }

            return finiteDifference / cellSize[ T_direction ];
        }
    };
    
} // namespace detail

    template<
        uint32_t T_neighbors,
        uint32_t T_direction
    >
    using ForwardDerivativeFunctor = detail::DerivativeFunctor<T_neighbors -1, T_neighbors, T_neighbors, T_direction>;
    
    template<
        uint32_t T_neighbors,
        uint32_t T_direction
    >
    using BackwardDerivativeFunctor = detail::DerivativeFunctor<T_neighbors, T_neighbors - 1, T_neighbors, T_direction>;
```

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp.OLD
```diff
@@ -0,0 +1,216 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/simulation_defines.hpp"
```
psychocoderHPC, line 22: IMO this file should be deleted and is a left over.

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
```diff
@@ -0,0 +1,85 @@
...
+ */
+template< uint32_t T_neighbors >
+struct AOFDTDWeights
+{
+    HDINLINE constexpr AOFDTDWeights()
+    {
+        namespace powSpace = ::picongpu::plugins::radiation::util;
+        // Set initial value
+        weights[0] = 4. * T_neighbors * 
+        powSpace::pow( (
```
psychocoderHPC, line 50: pow should be available under `math::pow`

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
```diff
@@ -0,0 +1,85 @@
...
+ *                     2 * T_neighbors
+ */
+template< uint32_t T_neighbors >
+struct AOFDTDWeights
+{
+    HDINLINE constexpr AOFDTDWeights()
+    {
+        namespace powSpace = ::picongpu::plugins::radiation::util;
+        // Set initial value
+        weights[0] = 4. * T_neighbors *
```
psychocoderHPC, line 49: please use `4.0_X` to use `float_X` as type

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
```diff
@@ -0,0 +1,85 @@
...
+        powSpace::pow( (
+            factorial(2*T_neighbors) 
+            / float_X( powSpace::pow(2, 2*T_neighbors) * powSpace::pow( factorial(T_neighbors), 2) )
+        ), 2);
+
+        // Compute all other values
+        for ( uint32_t l = 1; l < T_neighbors; ++l )
+        {
+            weights[l] = -1.
+                * powSpace::pow( l - 0.5, 2 )
```
psychocoderHPC, line 59: please use `float_X` and cast `l` explicit. The second parameter of `pow` can stay integer

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
```diff
@@ -0,0 +1,85 @@
...
+    {
+        return n <= 1 ? 1 : (n * factorial(n - 1));
+    }
+
+    float_X weights[T_neighbors];
+};
+} // namespace aoFDTD
+} // namespace maxwellSolver
+} // namespace fields
+} // namespace picongpu
```
psychocoderHPC, line 85: missing EOF

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
```diff
@@ -0,0 +1,85 @@
...
+                / float_X( T_neighbors + l )
+                / float_X( powSpace::pow( l + 0.5, 2 ) )
+                * weights[l-1];
+        }
+    }
+
+    HDINLINE constexpr float_X operator[](uint32_t const l) const
+    {
+        // PMACC_ASSERT_MSG(l < T_neighbors, "NUMBER_OF_COEFFICIENTS_IS_LIMITED_BY_NUMBER_OF_NEIGHBORS");
+        return l < T_neighbors ? weights[l] : ::std::nan("");
```
psychocoderHPC, line 70: We should not do a range check in the release code. The following code will do the range check only in debug mode.

```C++
#ifdef NDEBUG
  return weights[l];
#else
  return l < T_neighbors ? weights[l] : ::std::nan("");
#endif
```

psychocoderHPC (changes requested):

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
psychocoderHPC: Please use the static assert to check the ranges. `std::nan()` is not constexpr

include/picongpu/param/fieldSolver.param
```diff
@@ -60,12 +60,16 @@ namespace fields
...
      *
      * Field Solver Selection:
-     *  - Yee< CurrentInterpolation > : standard Yee solver
+     *  - Yee< CurrentInterpolation > : standard 2nd order Yee solver
      *  - YeePML< CurrentInterpolation >: standard Yee solver using Perfectly Matched Layer Absorbing Boundary Conditions (PML)
      *  - Lehe< CurrentInterpolation >: Num. Cherenkov free field solver in a chosen direction
      *  - LehePML< CurrentInterpolation >: Num. Cherenkov free field solver in a chosen direction
      *                                     using Perfectly Matched Layer Absorbing Boundary Conditions (PML)
      *  - DirSplitting< CurrentInterpolation >: Sentoku's Directional Splitting Method
+     *  - ArbitraryOrderFDTD< CurrentInterpolation, 8 >: 16th order solver using 8 neighbors for the calculation of spatial derivatives.
```
sbastrakov, line 69: Do you think it's more easy for a user to operate with order or number of neighbors? I would think the order, and then we could internally derive the number of neighbors from it.

sbastrakov: An unrelated note: this order is just for spatial derivatives. So this solver is FDTD(2, 16) from the literature. Perhaps we could clarify it more in the comment.

steindev: I totally agree to your second comment.

Regarding the first, I chose number of neighbors, since users could be tempted to choose uneven numbers for the order which is not valid as the order is always twice the number of neighbors, i.e. 2n. Therefore an extra check with a corresponding error message would be necessary, when taking the order as input, which I wanted to avoid due to laziness :grin:

sbastrakov: I see your motivation. But feel here the order is a user-side parameter, while the number of neighbors needed is a purely computational one, and PIConGPU already automatically takes care of it via margins, so that technically a user is not required to know how order corresponds to the number of neighbors.

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
steindev: In order to calculate the weights at compile time we agreed to use `pow` from radiation which is `constexpr`.

sbastrakov: It is possible that we introduce a new generic `pow` and use it from here and from radiation?

include/picongpu/param/fieldSolver.param
steindev: On the contrary, I am convinced the user should know this connection. It is essential for a users to know the underlying principle of the used methods. This is also the reason why I added this larger amount of documentation.
From a user perspective, I would argue that it does not matter whether I give order or number of neighbors.

As a side effect, the user may be able to e.g. relate the compile error `Entry function '...' uses too much shared data` when using too many neighbors to its cause.

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
steindev: I would say it is generally a good idea to move the math functions out of the radiation plugin @PrometheusPi. Actually, I could also move the factorial function in my `struct Weights` to some other place where math functions are collected.

sbastrakov:
Note: when you rebase on the current `dev`, the output should be proper openPMD

PrometheusPi:
@steindev Your code still contains end of line white spaces and spaces before pre-compiler commands.

steindev:
> @steindev Your code still contains end of line white spaces and spaces before pre-compiler commands.

What I marked as resolved is only resolved in my source, I did not push yet....

steindev:
> Note: when you rebase on the current `dev`, the output of field solver properties should be proper openPMD now.

:+1: tested and works

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp
```diff
@@ -0,0 +1,262 @@
...
+     * @tparam T_direction direction to take derivative in, 0 = x, 1 = y, 2 = z
+     */
+     template<
+        uint32_t T_neighbors,
+        uint32_t T_direction
+     >
+     using ForwardDerivativeFunctor = detail::GeneralAofdtdDerivative< T_neighbors - 1, T_neighbors, T_neighbors, T_direction >;
+
+
+     /** Functor for backward derivative along the given direction used in ArbitraryOrderFDTD solver
```
psychocoderHPC, line 168: please remove this copy past comment block and use doxygen groups

```
/** nice comment for both functions
 * 
 * foooooooooooooo
 * @{
 */
struct Foo;
struct Bar;
//!@}
```

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp
```diff
@@ -0,0 +1,262 @@
...
+        maxwellSolver::aoFDTD::ForwardDerivativeFunctor<
+            T_neighbors,
+            T_direction
+        >
+    >
+    {
+    };
+
+
+    /** DerivativeFunctor type trait specialization for Backward derivative in
```
psychocoderHPC, line 233: Please use doxygen groups https://www.doxygen.nl/manual/grouping.html
You can also have a bried description per struct.

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
```diff
@@ -0,0 +1,85 @@
...
+template< uint32_t T_neighbors >
+struct AOFDTDWeights
+{
+    HDINLINE constexpr AOFDTDWeights()
+    {
+        namespace powSpace = ::picongpu::plugins::radiation::util;
+        // Set initial value
+        weights[0] = 4.0_X * T_neighbors *
+            powSpace::pow( (
+                factorial(2*T_neighbors)
```
psychocoderHPC, line 51: please use space around binary operators, e.g. `*`
Please check the full file.

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
```diff
@@ -0,0 +1,85 @@
...
+                / float_X( powSpace::pow(2.0_X, 2*T_neighbors) * powSpace::pow( factorial(T_neighbors), 2) )
+            ), 2);
+
+        // Compute all other values
+        for ( uint32_t l = 1; l < T_neighbors; ++l )
```
psychocoderHPC, line 56: should be `uint32_t l = 1u`

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
```diff
@@ -0,0 +1,85 @@
...
+    {
+        PMACC_ASSERT_MSG(l < T_neighbors, "NUMBER_OF_COEFFICIENTS_IS_LIMITED_BY_NUMBER_OF_NEIGHBORS");
+        return weights[l];
+    }
+
+private:
+
+    HDINLINE constexpr uint32_t factorial(uint32_t const n) const
+    {
+        return n <= 1 ? 1 : (n * factorial(n - 1));
```
psychocoderHPC, line 77: should be `return n <= 1u ? 1u : (n * factorial(n - 1u));`

psychocoderHPC (changes requested):
:heart:  Could you please rebase.

steindev:
I included @psychocoderHPC's latest requests and fixed some spelling mistakes in the docs. Ready for final review (or merge ;-D).

steindev:
Latest test results:

# Steady current flowing through a square wire implemented by fieldBackground J
For the magnetic fields (i.e. the only ones that are created), PIConGPU produces identical results to my sandbox implementation of AOFDTD. Here 8th order.

image

image


# Current oscillating with time in a square wire
Basically the same as above, just with a sinusoidal implementation of the current in time.
Again identical results. Not only in the B_x and B_y fields, but also in the E_z-Field.
These are all the fields I can simulate in my 2D python sandbox.

image

Only one thing left:
**Merge, Merge, Merge !!!**

steindev:
**ping** @sbastrakov. PR is ready for final review or merge.

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
```diff
@@ -0,0 +1,85 @@
...
+ *                     2 * T_neighbors
+ */
+template< uint32_t T_neighbors >
+struct AOFDTDWeights
+{
+    HDINLINE constexpr AOFDTDWeights()
+    {
+        namespace powSpace = ::picongpu::plugins::radiation::util;
+        // Set initial value
+        weights[0] = 4.0_X * T_neighbors *
```
PrometheusPi, line 49: shouldn' the weight be `0<w<1`. If I check this equation I get:
```
( 4 * N * ((2N)!/(2^(2N) (N!)^2 ))^2 ) 

N = 1 --> 1
N = 2 --> 9/8
N = 3 --> 75/64
```

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
```diff
@@ -0,0 +1,85 @@
...
+            powSpace::pow( (
+                factorial(2 * T_neighbors)
+                / float_X( powSpace::pow(2.0_X, 2 * T_neighbors) * powSpace::pow( factorial(T_neighbors), 2) )
+            ), 2);
+
+        // Compute all other values
+        for ( uint32_t l = 1u; l < T_neighbors; ++l )
+        {
+            weights[l] = -1.0_X
```
PrometheusPi, line 58: for `T_neighbors = 2` this would lead to:
```
w[0] = 9/8
w[1] = -9/8* 1/27 = -1  / (8*9) = -1/72
```

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
```diff
@@ -0,0 +1,85 @@
...
+    {
+        PMACC_ASSERT_MSG(l < T_neighbors, "NUMBER_OF_COEFFICIENTS_IS_LIMITED_BY_NUMBER_OF_NEIGHBORS");
+        return weights[l];
+    }
+
+private:
+
+    HDINLINE constexpr uint32_t factorial(uint32_t const n) const
+    {
+        return n <= 1u ? 1u : (n * factorial(n - 1u));
```
PrometheusPi, line 77: This recursive call might be quite expensive. @psychocoderHPC and @sbastrakov how do you think this behaves on the device side?

PrometheusPi (changes requested):
Great job @steindev. The excellent documentation impressed me very much :+1: 🚀. I have some minor comments but nothing serious. The most important issue is my (probably) misconception of the weighting.

docs/source/models/AOFDTD.rst
```diff
@@ -0,0 +1,373 @@
...
+
+when replacing with the lowest order central differences.
+Note, with this leapfrog discretization or staggering, derivatives of field quantities are calculated at positions
+between positions where the field quantities are known.
+
+The above discretization uses one neighbor to each side from the point where the derivative is calculated yielding a
+second order accurate approximation of the derivative.
+Using more neighbors for the approximation of the spatial derivative is possible in PIConGPU and reduces the
+discretization error.
+The error scales with twice the number of neighbors :math:`M` used to approximate the derivative.
```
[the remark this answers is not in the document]
steindev: Well, it tells you that it pays off twice to increase the number of neighbors. The order scales as 2nd, 4th, 6th, 8th, ... , so only even numbers. This sentence basically makes the connection between the number of neighbors and the order of the solver. 
Also if you look at the following definitions of the weights, these are easier to write and understand when the number of neighbors is known.

docs/source/models/AOFDTD.rst
```diff
@@ -0,0 +1,373 @@
...
+Another conclusion from this figure is, that a short-pulse laser with a large bandwidth will suffer from severe
+dispersion if the sampling is bad.
+In the extreme case where a wavelength is not even sampled twice on the mesh, its field is exponentially damped
+[Taflove2005]_.
+
+Given that most simulations employ short-pulse lasers propagating along the :math:`y`-axis and featuring a large bandwidth,
+the resolution of the laser wavelength should be a lot better than in the example, e.g. :math:`N_\lambda=24`, to reduce
+errors due to numerical dispersion.
+
+Note, the reduced phase speed of light can further cause the emission of numerical Cherenkov radiation by fast charged
```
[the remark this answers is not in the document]
steindev: Absolutely, see e.g. wikipedia
```
Cherenkov radiation is electromagnetic radiation emitted when a charged particle (such as an electron) passes through a dielectric medium at a speed greater than the phase velocity (speed of propagation of a wave in a medium) of light in that medium.
```

I am a bit undecided between `phase velocity` and `speed of light`... Any suggestions which to use preferably?

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.hpp
```diff
@@ -0,0 +1,57 @@
...
+        ::picongpu::fields::maxwellSolver::ArbitraryOrderFDTD<
+            T_CurrentInterpolation,
+            T_neighbors
+        >
+    >
+    {
+        static StringProperty get()
+        {
+            pmacc::traits::StringProperty propList( "name", "other" );
+            propList["param"] = std::string("Arbitrary order FDTD, order ") + std::to_string(T_neighbors);
```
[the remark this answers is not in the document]
steindev: No, in this case the standard requires to use the fieldSolverParam keyword. In contrast to the pusher where this param keyword does not exist.

docs/source/models/AOFDTD.rst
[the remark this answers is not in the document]
sbastrakov: I am for velocity. We can give a more casual description in parenthesis, like `(speed of numerical electromagnetic wave propagation)`

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
sbastrakov: I think these should not really matter, the idea is that this is done once per kernel, not per cell.

docs/source/models/AOFDTD.rst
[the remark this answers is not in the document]
PrometheusPi: So does the error scale as `rel_error^(2N)`or as `rel_error/(2N)`?

docs/source/models/AOFDTD.rst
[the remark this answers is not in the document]
PrometheusPi: @steindev Thanks the clarification :+1:  ( I had a misconception of Cherenkov radiation 😱) 
I agree with @sbastrakov on the naming.

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp
```diff
@@ -0,0 +1,234 @@
...
+                    ( data( upperIndex ) - data( lowerIndex ) );
+            }
+
+            return finiteDifference / cellSize[ T_direction ];
+        }
+    };
+} // namespace detail
+
+
+    /**@{*/
```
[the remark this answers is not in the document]
steindev: Group comment, suggested by psychocoderHPC since the documentation of the following two definitions is almost equal. Allows two have one doxygen comment string for multiple following definitions. See (https://github.com/ComputationalRadiationPhysics/picongpu/pull/3338#discussion_r496463295)

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.hpp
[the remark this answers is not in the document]
PrometheusPi: @steindev You are right - you are setting `fieldSolverParameters` while leaving `fieldSolver` as `"other"` :+1:

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
PrometheusPi: Thanks for clarifying this @sbastrakov

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
steindev: According to Ghrist2000 and Vincenti2016 it is correct that the weight of the nearest neighbor is larger than on for solvers of higher order.

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
steindev: ghrist_b0(m=2, l=0.5) =  1.125 ,  9/8 =  1.125
ghrist_b0(m=2, l=1.5) = -0.042, -1/72 = -0.0139

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.def
```diff
@@ -0,0 +1,96 @@
...
+{
+namespace aoFDTD
+{
+
+    /** Curl to be applied for the E-field
+     *
+     * @tparam T_neighbors Number of neighbors used to calculate
+     *                      the derivative from finite differences.
+     *                      Order of derivative approximation is
+     *                      2 * T_neighbors
```
sbastrakov, line 44: Misaligned by 1

sbastrakov: Also, maybe add that  the same number of neighbors is used along all directions?

sbastrakov: Same two comments for CurlB

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.def
```diff
@@ -0,0 +1,96 @@
...
+     *             PhD thesis (2000)
+     *             Department of Applied Mathematics, University of Colarado
+     *
+     *             H Vincenti et al
+     *             doi:10.1016/j.cpc.2015.11.009
+     *
+     * @tparam T_neighbors Number of neighbors used to calculate
+     *                      the derivative from finite differences.
+     *                      Order of derivative approximation is
+     *                      2 * T_neighbors
```
sbastrakov, line 78: Also misaligned by 1

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.def
```diff
@@ -0,0 +1,96 @@
...
+     */
+    template<
+        typename T_CurrentInterpolation = currentInterpolation::None,
+        uint32_t T_neighbors = 4
```
sbastrakov, line 82: I do not like the default here, as the name suggests it's arbitrary, so a user has to choose

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
steindev: It is evaluated at compile time. There is no evaluation of weights during runtime.

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp
```diff
@@ -0,0 +1,234 @@
...
+            using IndexType = pmacc::DataSpace< simDim >;
+
+            // Define indice vectors for data access
+            auto lowerIndex = IndexType{}; // Vector initialized with zeros
+            auto upperIndex = IndexType{};
+
+            // lowerIndex: 0 if ( Forward ) else -1
+            lowerIndex[ T_direction ] = static_cast<int32_t>(T_upperNeighbors) - static_cast<int32_t>(T_neighbors);
+            // upperIndex: 1 if ( Forward ) else 0
+            upperIndex[ T_direction ] = static_cast<int32_t>(T_neighbors) - static_cast<int32_t>(T_lowerNeighbors);
```
sbastrakov, line 113: Spacing is weird on these lines: no space after `<` and `(`

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+            AOFDTDWeights< T_neighbors > const weights{};
+
+            // shortest distance finite difference as initial value
+            auto finiteDifference = weights[0] *
```
sbastrakov, line 118: `[ 0 ]`

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp
```diff
@@ -0,0 +1,234 @@
...
+                ( data( upperIndex ) - data( lowerIndex ) );
+
+            // Compute next finite differences according to order
+            for ( uint32_t l = 1; l < T_neighbors; ++l )
```
sbastrakov, line 122: `1u`, no space after `for`

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp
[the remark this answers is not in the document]
sbastrakov: This is for a group

docs/source/models/AOFDTD.rst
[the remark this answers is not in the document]
steindev: The error should scale as `(Dx/L)^(2M)`, where `Dx` is the mesh spacing and `L` the total length of the simulation volume edge collinear to the `x` axis.

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp
```diff
@@ -0,0 +1,234 @@
...
+     *                     Order of derivative approximation is
+     *                     2 * T_neighbors
+     *
+     * @tparam T_direction direction to take derivative in, 0 = x, 1 = y, 2 = z
+     */
+     template<
+        uint32_t T_neighbors,
+        uint32_t T_direction
+     >
+     using ForwardDerivativeFunctor = detail::GeneralAofdtdDerivative< T_neighbors - 1, T_neighbors, T_neighbors, T_direction >;
```
sbastrakov, line 168: These template parameters are better formatted on separate lines, as there are a few, and the line is long.

sbastrakov (changes requested):
Very nice work! Love the documentation extensions as well!!

Just a couple small notes. I would also prefer to regroup commits into one for all code and one for all documentation

include/picongpu/param/fieldSolver.param
```diff
@@ -57,11 +57,18 @@ namespace fields
...
      * Field Solver Selection:
-     *  - Yee< CurrentInterpolation > : standard Yee solver
-     *  - YeePML< CurrentInterpolation >: standard Yee solver using Perfectly Matched Layer Absorbing Boundary Conditions (PML)
+     *  - Yee< CurrentInterpolation > : Standard Yee solver approximating derivatives with respect to time and space
+     *                                  by second order finite differences.
+     *  - YeePML< CurrentInterpolation >: Standard Yee solver using Perfectly Matched Layer Absorbing Boundary Conditions (PML)
      *  - Lehe< CurrentInterpolation >: Num. Cherenkov free field solver in a chosen direction
      *  - LehePML< CurrentInterpolation >: Num. Cherenkov free field solver in a chosen direction
      *                                     using Perfectly Matched Layer Absorbing Boundary Conditions (PML)
+     *  - DirSplitting< CurrentInterpolation >: Sentoku's Directional Splitting Method
```
sbastrakov, line 66: I thought this solver was removed already in dev?

include/picongpu/param/fieldSolver.param
```diff
@@ -57,11 +57,18 @@ namespace fields
...
+     *  - ArbitraryOrderFDTD< CurrentInterpolation, 4 >: Solver using 4 neighbors to each direction for the approximation of
+     *                                                   *spatial* derivatives by finite differences.
+     *                                                   The number of neighbors can be changed from 4 to any positive, integer number.
+     *                                                   The order of the solver will be twice the number of neighbors in each direction.
+     *                                                   Yee's method is a special case of this using one neighbor to each direction.
      *  - None< CurrentInterpolation >: disable the vacuum update of E and B
      */
     using Solver = maxwellSolver::Yee< CurrentInterpolation >;
```
sbastrakov, line 74: I think it would be best to also enable the new solver in at least one setup for CI so that it is compile-time tested.

docs/source/models/AOFDTD.rst
```diff
@@ -0,0 +1,373 @@
...
+For the discretization of Maxwell's equations on a mesh, PIConGPU assumes a
+charge free and magnetic-divergence-free space **initially**, i.e.
+
+.. math::
+
+   \nabla \cdot \vec E &= 0
+
+   \nabla \cdot \vec B &= 0
+
+Therefore, only the equations
```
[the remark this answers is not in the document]
steindev: Indeed, satisfying these initially somehow is enough at beginning.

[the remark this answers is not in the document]
steindev: But then again I want to talk explicitly about PIConGPU. I will extend with one more sentence.

docs/source/models/AOFDTD.rst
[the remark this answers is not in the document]
PrometheusPi: I agree with @sbastrakov. I was confused with scaling (.../2N) vs. order (...^2N). In the latter the factor two certainly makes a difference. @steindev Thanks for clarification - could you add this to the text?

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp
PrometheusPi: Thanks - not intuitive to me, but definitely correct

sbastrakov:
OH, one more thing @steindev . Please add a note, or should it even be a compile-time error, that PIConGPU laser is implemented for Yee solver and would not work / would not work the same for AOFDTD with more than 1 neighbor. Would warn against or prevent unexpected results when changing, as you have experienced.

docs/source/models/AOFDTD.rst
```diff
@@ -0,0 +1,373 @@
...
+.. math::
+
+   \frac{1}{c^2}\frac{\partial}{\partial t}\vec E &= \nabla \times \vec B - \mu_0 \vec J
+
+   \frac{\partial}{\partial t}\vec B &= - \nabla \times \vec E
+
+need to be solved in PIConGPU, since correctly solving Gauss's law
+:math:`\nabla \cdot \vec{E} = \frac{1}{\varepsilon_0}\sum_s \rho_s` is automatically taken care of by using
+Esirkepov's current deposition method [Esirkepov2001]_ (or variants thereof) which solve the discretized continuity
+equation exactly.
```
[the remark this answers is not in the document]
steindev: I think it is a bit the other way round. Calculating the electric field's temporal derivative by the current calculated from the continuity equation always ensures satisfaction of Gauss' laws in a non discretized world.
Calculating the current from the discretized continuity equation by Esirkepov's scheme, using his definitions of forward and backward discretized differential operators and the generating functions, retains this property for the discretized Maxwell equations.

docs/source/models/AOFDTD.rst
```diff
@@ -0,0 +1,373 @@
...
+
+
+Discretization on a staggered mesh
+----------------------------------
+In the Finite-Difference Time-Domain Method, above Maxwell's equations are solved by replacing the partial space and
```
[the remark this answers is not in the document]
steindev: I would like to leave as is, as this is the actual explanation what "discretizing" means. But I will change "solve" to "discretized" to make this clear.

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.def
steindev: Agreed, as there is already a usable value suggested in the `fieldSolver.param` this seems superfluous here.

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp
sbastrakov: Yes, everywheere like `static_cast< int32_t >( T_neighbors )`, this is the style that the rest of your code follows, and rightfully so

steindev:
> OH, one more thing @steindev . Please add a note, or should it even be a compile-time error, that PIConGPU laser is implemented for Yee solver and would not work / would not work the same for AOFDTD with more than 1 neighbor. Would warn against or prevent unexpected results when changing, as you have experienced.

I added a `.. note::` to the documentation of the `*.param`s.
As the effect is only a slightly higher amplitude and energy, due to the increased weight of the nearest neighbor compared to Yee, I would not take drastic steps as throwing a compile time error.
A user could easily correct by choosing a lower amplitude in the `laser.param`. 
I would guess that usually laser amplitudes will be less accurate known from experiment than the error following the initilization.

include/picongpu/param/fieldSolver.param
steindev: Ok. How to do this? Just changing the solver in one of the examples or in a `cmakeFlags`?

sbastrakov: Yes. The idea is that it is part of some .param + cmakeFlags configuration in at least either an example or test. For which example it fits, I think you know better. In case none, can add to test [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/share/picongpu/tests/compileFieldSolver/cmakeFlags)

steindev:
> > OH, one more thing @steindev . Please add a note, or should it even be a compile-time error, that PIConGPU laser is implemented for Yee solver and would not work / would not work the same for AOFDTD with more than 1 neighbor. Would warn against or prevent unexpected results when changing, as you have experienced.
> 
> I added a `.. note::` to the documentation of the `*.param`s.
> As the effect is only a slightly higher amplitude and energy, due to the increased weight of the nearest neighbor compared to Yee, I would not take drastic steps as throwing a compile time error.
> A user could easily correct by choosing a lower amplitude in the `laser.param`.
> I would guess that usually laser amplitudes will be less accurate known from experiment than the error following the initilization.

By now, I actually added an `* \attention` to the doxygen documentation of the `fieldSolver.param`. This way it is documented in the file and automatically in the readthedocs documentation of the `fieldSolver.param`. I think this is the best solution.

steindev:
@sbastrakov, @PrometheusPi, @psychocoderHPC: I just recognised, that the `fieldSolver.param` files in the `share/picongpu/examples/KelvinHelmholtz` and `share/picongpu/examples/LaserWakeField` examples are identical to the standard file in `include/picongpu/param`.

I propose to remove these files, in order to keep the files in an example at a minimum. i.e. just to keep these which are really different from the standard.

I would do it with this PR but in an extra commit.

sbastrakov:
@steindev nice observation! I would prefer in a separate PR tho

steindev:
> @steindev nice observation! I would prefer in a separate PR tho

Ok, I opened a new PR. See #3374.

steindev:
@sbastrakov, @PrometheusPi @psychocoderHPC: I just pushed a new update. It includes all your remarks from the several reviews and structures all changes into three separate commits (1. implementation, 2. documentation, 3. adaption of examples).

Ready to merge?

sbastrakov (approved):
Thanks for the updates! Fine with me

psychocoderHPC (approved):

share/picongpu/examples/FieldAbsorberTest/cmakeFlags
```diff
@@ -30,62 +30,66 @@
...
 
 # The following tests are for absorber performance and demonstration of
 # reasonable parameters, commented out to make compile-time tests faster
 
 # Exponential damping in 3D, default strength
-#flags[4]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=Yee;-DPARAM_ABSORBER_SIZE=12'"
-#flags[5]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=Yee;-DPARAM_ABSORBER_SIZE=16'"
-#flags[6]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=Yee;-DPARAM_ABSORBER_SIZE=24'"
-#flags[7]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=Yee;-DPARAM_ABSORBER_SIZE=32'"
+#flags[6]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=Yee;-DPARAM_ABSORBER_SIZE=12'"
```
PrometheusPi, line 47: In these commented test the `< CurrentInterpolation >` is missing:
```suggestion
#flags[6]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=Yee< CurrentInterpolation >;-DPARAM_ABSORBER_SIZE=12'"
```

PrometheusPi (changes requested):

share/picongpu/examples/FieldAbsorberTest/cmakeFlags
steindev: :facepalm:

PrometheusPi (dismissed):

sbastrakov (dismissed):

sbastrakov:
Hm, weird. The last commit is shown to have passed [here](https://gitlab.com/hzdr/crp/picongpu/-/commit/0389a595c165d8a0935c59b01d04e16ec510c2f9) but somehow the status got stuck. @psychocoderHPC do you know how to propagate it here? Or maybe just force-merge.

PrometheusPi:
@sbastrakov this seems to effect **all** pull request - same with #3355 - thus I would vote to not force merge if not absolutely necessary.

psychocoderHPC:
Our HZDR CI has a bug, if you assign someone to the PR the parser is crashing. I removed the assigned person, please push again to this PR.

PrometheusPi:
@psychocoderHPC Thanks for fixing this - so no pull request assignments for now. (Sorry for assigning - I did not of this CI bug.)

steindev:
> Our HZDR CI has a bug, if you assign someone to the PR the parser is crashing. I removed the assigned person, please push again to this PR.

I pushed without changing something (just `git commit --amend; git push -f`). CI ran and is happy. :godmode:

Just a reviewer missing who (re)approves and merges :smile: :rocket:

ping @PrometheusPi or @sbastrakov

PrometheusPi (approved):

PR #3348, opened 2020-09-14
Add generic utilities to get absorber thickness

sbastrakov:
These are to be used in the upcoming incident field implementation.

I did not change the usage in domain adjuster and exponential absorber so far, only in PML.

### DIFF
--- a/include/picongpu/fields/absorber/Absorber.hpp
+++ b/include/picongpu/fields/absorber/Absorber.hpp
@@ -167,6 +167,10 @@ namespace detail
 
     /** Number of absorber cells along each boundary
      *
+     * Stores the global absorber thickness in case the absorbing boundary
+     * conditions are used along each boundary. Note that in case of periodic
+     * boundaries the corresponding values will be ignored.
+     *
      * Is uniform for both PML and exponential damping absorbers.
      * First index: 0 = x, 1 = y, 2 = z.
      * Second index: 0 = negative (min coordinate), 1 = positive (max coordinate).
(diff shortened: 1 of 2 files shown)


include/picongpu/fields/absorber/Absorber.hpp
```diff
@@ -178,6 +182,116 @@ namespace detail
...
+     * This function takes into account which boundaries are periodic and
+     * absorbing.
+     */
+    inline Thickness getGlobalThickness()
+    {
+        Thickness thickness;
+        for( uint32_t axis = 0u; axis < 3u; axis++ )
+            for( uint32_t direction = 0u; direction < 2u; direction++ )
+                thickness( axis, direction ) = numCells[ axis ][ direction ];
+        const DataSpace< DIM3 > isPeriodicBoundary =
```
psychocoderHPC, line 248: You assign a N-DIM value which can be 2 to a 3 dimensional value. This should fail.

include/picongpu/fields/absorber/Absorber.hpp
```diff
@@ -178,6 +182,116 @@ namespace detail
...
+            Environment<simDim>::get().EnvironmentController().getCommunicator().getPeriodic();
+        for( uint32_t axis = 0u; axis < 3u; axis++ )
```
psychocoderHPC, line 250: You should go only until `simDim` all other axis should be zero.

include/picongpu/fields/absorber/Absorber.hpp
```diff
@@ -178,6 +182,116 @@ namespace detail
         { Absorber::zNegativeNumCells, Absorber::zPositiveNumCells }
     };
 
+    //! Thickness of the absorbing layer
```
psychocoderHPC, line 185: Could you please add a description that this object should not be stored over multiple timesteps. Since communication maks are used the cached values will be only valid within one simulation step.

psychocoderHPC (changes requested):

include/picongpu/fields/absorber/Absorber.hpp
sbastrakov: Discussed offline, the value on the right side is also always 3d. In case of a 2d simulation, the z axis is always set to periodic (which makes sense).

include/picongpu/fields/absorber/Absorber.hpp
sbastrakov: This is also correct, as for 2d `isPeriodicBoundary[2]` will be 1 and we should actually go into the `if` statement body

include/picongpu/fields/absorber/Absorber.hpp
sbastrakov: This concerns not the object itself, but whether `getLocalThickness()` results can be reused on another time step (which they can't generally). I will add an explanatory note there.

sbastrakov:
Fixed and rebased the third comment. The first two I think were fine already, answered there.

psychocoderHPC (approved):

PR #3352, opened 2020-09-16
Set default values for some PMacc game of life example arguments

sbastrakov:
While working on #3351 I've found out the default values for arguments are not set. Particularly for the number of steps and rule, which when not set lead to confusing behavior. This PR adds default values for these two.

### DIFF
--- a/share/pmacc/examples/gameOfLife2D/main.cpp
+++ b/share/pmacc/examples/gameOfLife2D/main.cpp
@@ -41,7 +41,7 @@ int main( int argc, char **argv )
 {
     typedef ::gol::Space Space;
 
-    std::vector<uint32_t> devices;  /* will be set by boost program argument option "-d 3 3 3" */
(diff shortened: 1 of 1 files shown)


share/pmacc/examples/gameOfLife2D/main.cpp
```diff
@@ -41,7 +41,7 @@ int main( int argc, char **argv )
 {
     typedef ::gol::Space Space;
 
-    std::vector<uint32_t> devices;  /* will be set by boost program argument option "-d 3 3 3" */
+    std::vector<uint32_t> devices;  /* will be set by boost program argument option "-d 3 3" */
```
sbastrakov, line 44: This is unrelated, but was noticed during this change. The example only supports 1d and 2d

psychocoderHPC (approved):

PrometheusPi (approved):

psychocoderHPC (approved):

PR #3354, opened 2020-09-17
radiation plugin: add new option

psychocoderHPC:
Add runtime option `numTmpResults` to increase independent work on a
device.
If set to one the code will behaves as before this PR.
The default will is two to utilize modern GPU devices with a typicaly configuration.

### DIFF
--- a/docs/TBG_macros.cfg
+++ b/docs/TBG_macros.cfg
@@ -124,7 +124,8 @@ TBG_stopWindow="--stopWindow 1337"
 #--<species>_radiation.end     Time step to stop calculating the radiation
 #--<species>_radiation.radPerGPU     If flag is set, each GPU stores its own spectra without summing the entire simulation area
 #--<species>_radiation.folderRadPerGPU     Folder where the GPU specific spectras are stored
-#--e_<species>_radiation.compression    If flag is set, the hdf5 output will be compressed.
+#--<species>_radiation.compression    If flag is set, the hdf5 output will be compressed.
+#--<species>_radiation.numJobs     Number of independent jobs used for the radiation calculation.
 TBG_radiation="--<species>_radiation.period 1 --<species>_radiation.dump 2 --<species>_radiation.totalRadiation \
                --<species>_radiation.lastRadiation --<species>_radiation.start 2800 --<species>_radiation.end 3000"
 
--- a/docs/source/usage/plugins/radiation.rst
+++ b/docs/source/usage/plugins/radiation.rst
@@ -287,6 +287,11 @@ Command line option                       Description
 ``--<species>_radiation.folderRadPerGPU`` Name of the folder, where the GPU specific spectra are stored.
                                           Default: ``radPerGPU``
 ``--<species>_radiation.compression``     If set, the hdf5 output is compressed.
+``--<species>_radiation.numJobs``         Number of independent jobs used for the radiation calculation.
+                                          This option is used to increase the utilization of the device by producing more independent work.
+                                          This option enables accumulation of data in parallel into multiple temporary arrays, thereby increasing the utilization of
+                                          the device by increasing the memory footprint
+                                          Default: ``2``
 ========================================= ==============================================================================================================================
 
 Memory Complexity
@@ -295,7 +300,8 @@ Memory Complexity
 Accelerator
 """""""""""
 
-each energy bin times each coordinate bin allocates one counter (``float_X``) permanently and on each accelerator.
+locally, ``numJobs`` times number of frequencies ``N_omega`` times number of directions ``N_theta`` is permanently allocated.
+Each result element (amplitude) is a double precision complex number.
 
 Host
 """"
--- a/include/picongpu/plugins/radiation/Radiation.hpp
+++ b/include/picongpu/plugins/radiation/Radiation.hpp
@@ -77,7 +77,6 @@ namespace idLabels
 }// end namespace idLabels
 
 
-
 ///////////////////////////////////////////////////////////////////////////////////////////////
 ///////////////////////////////  Radiation Plugin Class  ////////////////////////////////////
 ///////////////////////////////////////////////////////////////////////////////////////////////
@@ -97,8 +96,10 @@ private:
      * frequency. Layout of the radiation array is:
      * [omega_1(theta_1),omega_2(theta_1),...,omega_N-omega(theta_1),
      *   omega_1(theta_2),omega_2(theta_2),...,omega_N-omega(theta_N-theta)]
+     * The second dimension is used to store intermediate results if command
+     * line option numJobs is > 1.
      */
-    GridBuffer<Amplitude, DIM1> *radiation;
+    GridBuffer<Amplitude, 2> *radiation;
     radiation_frequencies::InitFreqFunctor freqInit;
     radiation_frequencies::FreqFunctor freqFkt;
 
@@ -119,6 +120,7 @@ private:
     bool radPerGPU;
     std::string folderRadPerGPU;
     DataSpace<simDim> lastGPUpos;
+    int numJobs;
 
     /**
      * Data structure for storage and summation of the intermediate values of
@@ -214,7 +216,8 @@ public:
             ((pluginPrefix + ".end").c_str(), po::value<uint32_t > (&radEnd)->default_value(0), "time index when radiation should end with calculation")
             ((pluginPrefix + ".radPerGPU").c_str(), po::bool_switch(&radPerGPU), "enable radiation output from each GPU individually")
             ((pluginPrefix + ".folderRadPerGPU").c_str(), po::value<std::string > (&folderRadPerGPU)->default_value("radPerGPU"), "folder in which the radiation of each GPU is written")
-            ((pluginPrefix + ".compression").c_str(), po::bool_switch(&compressionOn), "enable compression of hdf5 output");
+            ((pluginPrefix + ".compression").c_str(), po::bool_switch(&compressionOn), "enable compression of hdf5 output")
+            ((pluginPrefix + ".numJobs").c_str(), po::value<int > (&numJobs)->default_value(2), "Number of independent jobs used for the radiation calculation.");
     }
 
 
@@ -282,13 +285,22 @@ private:
     {
         if(!notifyPeriod.empty())
         {
+            if(numJobs <= 0)
+            {
+                std::cerr << "'numJobs' must be '>=1' value is adjusted from" << numJobs << " to '1'." << std::endl;
+                numJobs = 1;
+            }
             // allocate memory for all amplitudes for temporal data collection
             tmp_result = new Amplitude[elements_amplitude()];
 
             /*only rank 0 create a file*/
             isMaster = reduce.hasResult(mpi::reduceMethods::Reduce());
 
-            radiation = new GridBuffer<Amplitude, DIM1 > (DataSpace<DIM1 > (elements_amplitude())); //create one int on GPU and host
+            /* Buffer for GPU results.
+             * The second dimension is used to store intermediate results if command
+             * line option numJobs is > 1.
+             */
+            radiation = new GridBuffer<Amplitude, 2> (DataSpace<2>(elements_amplitude(), numJobs));
 
             freqInit.Init(frequencies_from_list::listLocation);
             freqFkt = freqInit.getFunctor();
@@ -387,6 +399,15 @@ private:
   {
     radiation->deviceToHost();
     __getTransactionEvent().waitForFinished();
+
+    auto dbox = radiation->getHostBuffer().getDataBox();
+    int numAmp = elements_amplitude();
+    // update the main result matrix (y index zero)
+    for( int resultIdx = 1; resultIdx < numJobs; ++resultIdx )
+        for( int ampIdx = 0; ampIdx < numAmp; ++ampIdx )
+        {
+            dbox(DataSpace< 2 >( ampIdx, 0 ) ) += dbox(DataSpace< 2 >( ampIdx, resultIdx ) );
+        }
   }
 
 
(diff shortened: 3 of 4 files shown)


include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -288,7 +291,7 @@ class Radiation : public ISimulationPlugin
             /*only rank 0 create a file*/
             isMaster = reduce.hasResult(mpi::reduceMethods::Reduce());
 
-            radiation = new GridBuffer<Amplitude, DIM1 > (DataSpace<DIM1 > (elements_amplitude())); //create one int on GPU and host
+            radiation = new GridBuffer<Amplitude, 2> (DataSpace<2>(elements_amplitude(), numTmpResults)); //create one int on GPU and host
```
psychocoderHPC, line 294: note: I explicitly uses `2` instead of `DIM2` to reflect that it is not representing the special dimension.

sbastrakov: Probably the comment in this line needs to be updated, since it's not one any more?

docs/source/usage/plugins/radiation.rst
```diff
@@ -287,6 +287,10 @@ Command line option                       Description
 ``--<species>_radiation.folderRadPerGPU`` Name of the folder, where the GPU specific spectra are stored.
                                           Default: ``radPerGPU``
 ``--<species>_radiation.compression``     If set, the hdf5 output is compressed.
+``--<species>_radiation.numTmpResults``   Number of temporary results per device.
+                                          This option is used to increase the utilization of the device by producing more independent work.
+                                          Increase results in more independent parallel work and a larger memory footprint.
```
sbastrakov, line 292: This sentence does not seem clear, also grammatically

docs/source/usage/plugins/radiation.rst
```diff
@@ -295,7 +299,8 @@ Memory Complexity
 Accelerator
 """""""""""
 
-each energy bin times each coordinate bin allocates one counter (``float_X``) permanently and on each accelerator.
+locally, ``numTmpResults`` matrices of the size number of frequencies times directions times is permanently allocated.
```
sbastrakov, line 302: I guess `times directions times` was meant to be `times number of directions`?

docs/TBG_macros.cfg
```diff
@@ -124,7 +124,8 @@ TBG_stopWindow="--stopWindow 1337"
 #--<species>_radiation.end     Time step to stop calculating the radiation
 #--<species>_radiation.radPerGPU     If flag is set, each GPU stores its own spectra without summing the entire simulation area
 #--<species>_radiation.folderRadPerGPU     Folder where the GPU specific spectras are stored
-#--e_<species>_radiation.compression    If flag is set, the hdf5 output will be compressed.
+#--<species>_radiation.compression    If flag is set, the hdf5 output will be compressed.
+#--<species>_radiation.numTmpResults Number of temporary results per device.
```
sbastrakov, line 128: I don't like `numTmpResults` as this feels very unclear for a user. I suggest using word `granularity`, as this more clearly represents the meaning, and also is used in the same meaning for general parallel processing.

docs/source/usage/plugins/radiation.rst
Anton-Le: I concur.
A better formulation would be:
"This option enables accumulation of data in parallel into multiple temporary arrays, thereby increasing the utilization of the device by increasing the memory footprint."

Regarding the memory footprint it may be meaningful to add: "The memory usage is increased by $N_{omega}N_{directions}$ for each temporary array."

docs/source/usage/plugins/radiation.rst
Anton-Le: Regarding the memory footprint:
"The memory usage is increased by $N_{omega}N_{directions}$ for each temporary array."
AFAIK we can render LaTeX in the documentation.

docs/TBG_macros.cfg
Anton-Le: I think `granularity` is confusing here. On first reading I would _not_ attribute the CS task granularity to it, when trying to set up a physics simulation.
Maybe something like `numParAccumulations` ?

docs/source/usage/plugins/radiation.rst
```diff
@@ -287,6 +287,10 @@ Command line option                       Description
 ``--<species>_radiation.folderRadPerGPU`` Name of the folder, where the GPU specific spectra are stored.
                                           Default: ``radPerGPU``
 ``--<species>_radiation.compression``     If set, the hdf5 output is compressed.
+``--<species>_radiation.numTmpResults``   Number of temporary results per device.
+                                          This option is used to increase the utilization of the device by producing more independent work.
+                                          Increase results in more independent parallel work and a larger memory footprint.
+                                          Default: ``2``
```
Anton-Le, line 293: As a matter of consistency I think the default should be 1.

Until we are reasonably sure that:

- This approach does not introduce some weird error.
- It's not detrimental to performance on e.g., K20 and K80 systems.
(admittedly test for both are ongoing).

psychocoderHPC: IMO K20 and k80 are not interesting anymore. The hardware is nearly a decade old. If you run on such systems you should adjust it by hand if required. 

`This approach does not introduce some weird error.` 

This is true for each code change. Therefore you are running currently some tests but if this PR is not introducing an general error during the reduce of the tmp arrays it is not critical.

docs/TBG_macros.cfg
psychocoderHPC: `numParAccumulations` must be `numParallelAccumulations` because `par` is the common used abbreviation  in PIConGPU for `particle`

psychocoderHPC: In general the arrays are `intermediate` results. The naming should more what it is instead where or how it is used.

psychocoderHPC: I changed it now to `numJobs` what is nearly similar to the option `j` in `make`

docs/source/usage/plugins/radiation.rst
PrometheusPi: I agree with @Anton-Le on this sentence.

docs/TBG_macros.cfg
PrometheusPi: I like @psychocoderHPC choice - alternatively how about `numAccumulationTasks`?

docs/source/usage/plugins/radiation.rst
PrometheusPi: I agree with @Anton-Le if the physical correctness has not yet been confirmed for the k20 and k80 platform. If this still produces the same result, I would be fine with a default of two.

psychocoderHPC: The changes are independent of the GPU architecture, so the tests can take place on any GPU hardware.
@Anton-Le could you please confirm the correctness of the results based on your tests.

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -97,8 +96,10 @@ class Radiation : public ISimulationPlugin
      * frequency. Layout of the radiation array is:
      * [omega_1(theta_1),omega_2(theta_1),...,omega_N-omega(theta_1),
      *   omega_1(theta_2),omega_2(theta_2),...,omega_N-omega(theta_N-theta)]
+     * The second dimension is used to store intermediate results if command
+     * line option numJobs is used.
```
PrometheusPi, line 100: optional 
```suggestion
     * line option numJobs > 1.
```

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -97,8 +96,10 @@ class Radiation : public ISimulationPlugin
...
      */
-    GridBuffer<Amplitude, DIM1> *radiation;
+    GridBuffer<Amplitude, 2> *radiation;
```
PrometheusPi, line 102: Why this hard coded 2?

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -119,6 +120,7 @@ class Radiation : public ISimulationPlugin
     bool radPerGPU;
     std::string folderRadPerGPU;
     DataSpace<simDim> lastGPUpos;
+    int numJobs = 2;
```
PrometheusPi, line 123: Why do you initialize this variable with `2` if you set it by default to two later one?

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -282,13 +285,22 @@ class Radiation : public ISimulationPlugin
...
             // allocate memory for all amplitudes for temporal data collection
             tmp_result = new Amplitude[elements_amplitude()];
 
             /*only rank 0 create a file*/
             isMaster = reduce.hasResult(mpi::reduceMethods::Reduce());
 
-            radiation = new GridBuffer<Amplitude, DIM1 > (DataSpace<DIM1 > (elements_amplitude())); //create one int on GPU and host
+            /* Buffer for GPU results.
+             * The second dimension is used to store intermediate results if command
+             * line option numJobs is used.
```
PrometheusPi, line 301:
```suggestion
             * line option numJobs > 1.
```

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -282,13 +285,22 @@ class Radiation : public ISimulationPlugin
...
+             */
+            radiation = new GridBuffer<Amplitude, 2> (DataSpace<2>(elements_amplitude(), numJobs));
```
PrometheusPi, line 303: Would it not be:
```suggestion
            radiation = new GridBuffer<Amplitude, numJobs> (DataSpace<numJobs>(elements_amplitude(), numJobs));
```
?

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -387,6 +399,15 @@ class Radiation : public ISimulationPlugin
...
     radiation->deviceToHost();
     __getTransactionEvent().waitForFinished();
+
+    auto dbox = radiation->getHostBuffer().getDataBox();
+    int numAmp = elements_amplitude();
+    // update the main result matrix (y index zero)
+    for( int resultIdx = 1; resultIdx < numJobs; ++resultIdx )
+        for( int ampIdx = 0; ampIdx < numAmp; ++ampIdx )
+        {
+            dbox(DataSpace<DIM2>( ampIdx, 0 ) ) += dbox(DataSpace< 2 >( ampIdx, resultIdx ) );
```
PrometheusPi, line 409: Double use of `DIM2` and `2` for `DataSpace` template argument

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -1188,8 +1209,8 @@ class Radiation : public ISimulationPlugin
       PMACC_KERNEL( KernelRadiationParticles<
           numWorkers
       >{} )(
-          gridDim_rad,
-          numWorkers
+          DataSpace< 2 >(gridDim_rad, numJobs),
+          DataSpace< 2 >(numWorkers,1)
```
PrometheusPi, line 1213: Here again `2`

PrometheusPi (changes requested):
There are some magic number used where I am not sure whether there should be `numJobs` instead.

include/picongpu/plugins/radiation/Radiation.hpp
psychocoderHPC: https://github.com/ComputationalRadiationPhysics/picongpu/pull/3354#discussion_r490109096

include/picongpu/plugins/radiation/Radiation.hpp
psychocoderHPC: Not possible because the`2` must be a compile time value and `numJobs` and this `2` is not the same.
`GridBuffer<Amplitude, 2>` is saying that we operate on a matrix of results, so the old one dimensional buffer (one row/array) is now a two dimensional buffer.

include/picongpu/plugins/radiation/Radiation.hpp
PrometheusPi: I think it would be better to set it via the boost option - there a documentation is also intrinsically given. But this is only my opinion.

include/picongpu/plugins/radiation/Radiation.hpp
psychocoderHPC: see #3354 (comment)
It is not a spatial dimension, to not mix it with the spatial  2D and 3D which is supported by picongpu I explicit set it to `2` instead of `DIM2`

include/picongpu/plugins/radiation/Radiation.hpp
PrometheusPi: Okay - if it tsays 2 dimensional even for `numJobs > 2`, I am fine with this. But this increases the memory footprint always and not only by default.

sbastrakov: It increase the footprint by `numJobs` times. So does not increase for `numJobs == 1`, as N x 1 matrix takes the same space as an array of size N.

psychocoderHPC:
@PrometheusPi I updated the PR with your suggestions

include/picongpu/plugins/radiation/Radiation.hpp
psychocoderHPC: fixed

docs/TBG_macros.cfg
psychocoderHPC: IMO `numAccumulationTasks` is to near to what we are doing. There is no need for the user (in the short name) to know how it is implemented.

include/picongpu/plugins/radiation/Radiation.hpp
PrometheusPi: Thanks for the clarification.

PrometheusPi (approved):

docs/source/usage/plugins/radiation.rst
Anton-Le: After the recent re-run of the reference simulation with the version from 4.08.2020 the maximum relative error in the  spectrum appears to be ~ 0.02 (for forward direction) and commonly ranges at ~ 0.004.
Hence I would consider this as correct.

Anton-Le (approved):

docs/source/usage/plugins/radiation.rst
PrometheusPi: @Anton-Le Thanks for testing the code.

PR #3371, opened 2020-09-29
Update "name" property of Higuera-Cary pusher

steindev:
The name update ensures conformity with the openPMD standard.
See openPMD/openPMD-standard#225.

### DIFF
--- a/include/picongpu/particles/pusher/particlePusherHigueraCary.hpp
+++ b/include/picongpu/particles/pusher/particlePusherHigueraCary.hpp
@@ -136,7 +136,7 @@ struct Push
 
     static pmacc::traits::StringProperty getStringProperties()
     {
-        pmacc::traits::StringProperty propList( "name", "HigueraCary" );
+        pmacc::traits::StringProperty propList( "name", "other:Higuera-Cary" );
         return propList;
     }
 };


sbastrakov (approved):

sbastrakov (changes requested):
Wait, but this seems not compliant actually. Let's wait for Axel's reply in that topic.

steindev:
> Wait, but this seems not compliant actually. Let's wait for Axel's reply in that topic.

Well, as it is his advice and he mentioned that they do it elsewhere too, I expect it to be an accepted policy. Although it is not written explicitly in the standard that it is allowed to supplement the allowed value by more text, it does not seem explicitly forbidden, too. The policy suggested is a very convenient way until the standard includes the `particlePushParameters` attribute or allows for `particlePush:Higuera-Cary`.

In any case, the way we do it now is wrong. The policy suggested *may* be wrong, too, but is at least closer to the standard. If it is wrong, there is currently no standard conform way to do it. An update of the standard itself may take months. So I don't see any other way to do it *now*. Do you?

sbastrakov:
I agree it's a reasonable policy and the new value feels better than the old. However, from how I understood the standard (and that's why I would like to wait for an "official" response) it is equally non-conforming as the old value. And the conformant way is to just put `other`. So I agree your fix improves the situation, but I think putting `other` would be even better.

PrometheusPi:
I agree with @sbastrakov - we should wait for the official decision.

steindev:
But writing just `other` is in my opinion the worst solution of all, as it does not document at all what pusher is used.

sbastrakov:
It would be openPMD-conformant (as I believe we want and maybe even claim to be). And we can always add another output attribute, not governed by the standard, to add additional info.

sbastrakov:
Btw can even be `particlePushParameters` that you suggested to be added to the standard. and is generally logical with the naming schema being used.

PrometheusPi:
@steindev I personally agree with you that a more detailed description would be better for documentation, but the standard is very clear on that:
```
particlePush

    type: (string)
    scope: required
    description: Particle-Pushing Algorithm
    allowed values:
        - Boris (J.P. Boris. Relativistic plasma simulation-optimization of a hybrid code. USA, 1970)
        - Vay (doi:10.1063/1.2837054)
        - free-streaming (constantly moving with initial momentum)
        - LLRK4 (reduced Laundau-Lifshitz pusher via RK4 and classical radiation reaction, doi:10.1016/j.cpc.2016.04.002)
        - none (static particles such as probes)
        - other
```
There are no other values allowed - thus not even an extension on the given name `other`. I am not aware of any post processing code that currently relies on that attribute, but if there is one, it might break when adding non-standar entries. Perhaps, we should not only push the standard to add the `Higuera-Cary` pusher, but to allow for a string after `other`.

PrometheusPi:
@steindev Sorry - that already what you suggested in openPMD/openPMD-standard#225.

steindev:
> And we can always add another output attribute, not governed by the standard, to add additional info
> Btw can even be `particlePushParameters` ...

Ahh, then let's do it this way. But somebody probably needs implement that it is written to the file.

PrometheusPi:
@ax3l Just stated that no tool etc. will break if we use `other: Higuera-Cary`. So let's implement this. I will do this now.

PrometheusPi (approved):

PrometheusPi:
@sbastrakov Are you still against it?

sbastrakov (approved):
I still find it illogical, as instead of adhering to the standard we rely on the workaround that all the tools would accept the non-standard version. But it is not a big deal

PR #3386, opened 2020-10-15
Add a new compile-time test suite to test all particle pushers

sbastrakov:
Some pushers were not enabled in any example, and so were not compile-time tested.

### DIFF
--- a/share/picongpu/tests/compileParticlePusher/cmakeFlags
+++ b/share/picongpu/tests/compileParticlePusher/cmakeFlags
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+#
+# Copyright 2013-2020 Axel Huebl, Rene Widera, Sergei Bastrakov
+#
+# This file is part of PIConGPU.
+#
+# PIConGPU is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# PIConGPU is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PIConGPU.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
+#
+# generic compile options
+#
+
+################################################################################
+# add presets here
+#   - default: index 0
+#   - start with zero index
+#   - increase by 1, no gaps
+
+# Pushers are generally independent from particle shapes, so do not attempt to
+# test all possible combinations, just all pushers except Boris (tested in examples)
+flags[0]="-DPARAM_OVERWRITES:LIST='-DPARAM_PARTICLEPUSHER=HigueraCary;-DPARAM_PARTICLESHAPE=NGP'"
+flags[1]="-DPARAM_OVERWRITES:LIST='-DPARAM_PARTICLEPUSHER=Vay;-DPARAM_PARTICLESHAPE=CIC'"
+flags[2]="-DPARAM_OVERWRITES:LIST='-DPARAM_PARTICLEPUSHER=ReducedLandauLifshitz;-DPARAM_PARTICLESHAPE=TSC'"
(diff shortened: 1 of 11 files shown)


share/picongpu/tests/compileParticlePusher/cmakeFlags
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+#
+# Copyright 2013-2020 Axel Huebl, Rene Widera
```
PrometheusPi, line 3: @sbastrakov your name is missing

share/picongpu/tests/compileParticlePusher/include/picongpu/param/density.param
```diff
@@ -0,0 +1,46 @@
+/* Copyright 2013-2020 Axel Huebl, Heiko Burau, Rene Widera, Felix Schmitt,
```
PrometheusPi, line 1: same here

share/picongpu/tests/compileParticlePusher/include/picongpu/param/fileOutput.param
```diff
@@ -0,0 +1,57 @@
+/* Copyright 2013-2020 Axel Huebl, Rene Widera, Felix Schmitt,
```
PrometheusPi, line 1: If you define all output lists to be empty, why not remove this param file and avoid calling output in the cfg - it is a compile test anyway.

share/picongpu/tests/compileParticlePusher/include/picongpu/param/isaac.param
```diff
@@ -0,0 +1,58 @@
+/* Copyright 2016-2020 Alexander Matthes
```
PrometheusPi, line 1: not needed

PrometheusPi (changes requested):
@sbastrakov There seem to be too man param files submitted.

share/picongpu/tests/compileParticlePusher/include/picongpu/param/fileOutput.param
sbastrakov: You are right, I copied the other compile-time test suite without thinking of it. Guess were redundant there as well, but at least I will not add those in my PR.

share/picongpu/tests/compileParticlePusher/include/picongpu/param/isaac.param
sbastrakov: Same happened here. Thanks for pointing out!

share/picongpu/tests/compileParticlePusher/include/picongpu/param/fileOutput.param
sbastrakov: Now I think i understood the logic of other compile-time tests. By making the lists of what to output explicitly empty we guarantee the output part is not being compiled, regardless of whether openPMD-API / ADIOS dependencies were found. While with the standard `fileOutput.param` would still compile this part and thus spend compile-time for it, although this is not what we are testing in this suite.

PrometheusPi: @sbastrakov Thanks for clarifying this. But this will result in a lot of code duplications. Would linking the duplicated files or a common base for all test resolve this issue?

sbastrakov: It is also true. I think the issue would become less severe while things are gradually moved out from `.param` files to `.cfg` files. Additionally, perhaps we could somehow explicitly specify build dependencies for tests? Then e.g. if I was positive there is no openPMD-API, ADIOS and ISAAC for sure, the corresponding files could have been safely deleted.

sbastrakov:
I do not understand why do some jobs fail. They output `All right!` and thumb up, and then just

 ```
Cleaning up file based variables
ERROR: Job failed: exit code 1
```
with no further details.

psychocoderHPC:
> I do not understand why do some jobs fail. They output `All right!` and thumb up, and then just
> 
> ```
> Cleaning up file based variables
> ERROR: Job failed: exit code 1
> ```
> 
> with no further details.

All right is coming from compiling. The fail is on runtime `picongpu --help` can not be executed

PrometheusPi (approved):

sbastrakov:
@PrometheusPi could you press merge please?

Issue #3382, opened 2020-10-13
openPMD adios2 output missing


PrometheusPi:
I just ran a default LWFA simulation on summit using ADIOS2 (adios2/2.5.0 module) and openPMD-api 0.12.0. PIConGPU compiled without problems and the interface for the openPMD plugin was created. However, no ~~`*.bp`~~ output files are created. The openPMD directory is completely empty. I ran both:
```
 --openPMD.period 5000:10000:2500 --openPMD.file simData --openPMD.ext bp       

 --openPMD.period 5000:10000:2500 --openPMD.file simData --openPMD.ext bp --openPMD.json '{ "adios2": { "engine": { "type": "file", "parameters": { "BufferGrowthFactor": "1.2", "InitialBufferSize": "2GB" } } } }' 
```

The openPMD directory is created and the simulation reaches 100%, but no output is created. 
stdout and stderr give no clue of an error. 

@ax3l @psychocoderHPC @franzpoeschel Any ideas what might cause this problem?

ax3l:
ADIOS2 does not create or need `.bp` meta files anymore. You can operate completely on the directory it creates, even without a call on `bpmeta`.

This is identical to ADIOS1 behavior with `nometa` - here you just had to run `bpmeta` before you could read the data again.

PrometheusPi:
@ax3l Sorry for the confusion: no data at all is created. (I will correct the post above.)

ax3l:
@PrometheusPi ok, I'll let @franzpoeschel investigate this, since it seems PIConGPU specific.

Also double-check that all backends you expect in openPMD-api, esp. the ADIOS2 backend, are available, please: https://openpmd-api.readthedocs.io/en/0.12.0-alpha/citation.html#dependent-software

franzpoeschel:
> using ADIOS2 (adios2/2.5.0 module) and openPMD-api 0.12.0

openPMD 0.12.0 requires ADIOS 2.6. Can you verify whether your openPMD build includes an ADIOS backend?

PrometheusPi:
Okay - adios 2 2.5.5 works with openPMD-api 0.11.1 - I installed it but now I get the following compile error:
```
/.../picongpu/include/pmacc/../picongpu/plugins/openPMD/openPMDWriter.def(113): error #135: namespace "openPMD" has no member "Access"
```

PrometheusPi:
[...]

PR #3395, opened 2020-10-19
Deactivate support for writing to ADIOS1 via the openPMD plugin

franzpoeschel:
See [here](https://github.com/ComputationalRadiationPhysics/picongpu/issues/3382#issuecomment-708935904) for a reasoning.

### DIFF
--- a/include/picongpu/plugins/openPMD/openPMDWriter.hpp
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.hpp
@@ -149,6 +149,7 @@ Please pick either of the following:
 * Use the openPMD plugin with another backend, such as ADIOS2.
   If the openPMD API has been compiled with support for ADIOS2, the openPMD API
   will automatically prefer using ADIOS2 over ADIOS1.
+  Make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                 )END");
             }
             if( at == ::openPMD::Access::CREATE )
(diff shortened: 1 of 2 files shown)


include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -140,6 +140,17 @@ namespace openPMD
             log< picLog::INPUT_OUTPUT >( "openPMD: open file: %1%" ) % fullName;
             openPMDSeries = std::unique_ptr<::openPMD::Series >(
                 new ::openPMD::Series( fullName, at, communicator, jsonConfig ) );
+            if( openPMDSeries->backend() == "MPI_ADIOS1" )
+            {
+                throw std::runtime_error(R"END(
```
sbastrakov, line 145: What does `R"END` mean?

franzpoeschel: It's C++11 syntax for multiline strings, see item 6 [here](https://en.cppreference.com/w/cpp/language/string_literal).

psychocoderHPC: @franzpoeschel  IS it possible to check in CMake that OpenPMD-api is at least build with ADIOS2 and/or HDF5?

franzpoeschel: Yep, that should be doable. Will get to it.

franzpoeschel: Recent commit does that. I left the runtime check in there.

sbastrakov (approved):

psychocoderHPC (approved):

sbastrakov:
@franzpoeschel we just discussed in the office that we should now disable reading ADIOS1 checkpoints as well, so to completely remove support of ADIOS1. However it seems from the code, that the exception thrown with this PRs change already affects reading, not just writing? (I don't have an old checkpoint to test it readily).

franzpoeschel:
Yep, that one should catch both reading and writing @sbastrakov. I think the fastest way to test would be by creating an ADIOS1 checkpoint. Quickly commenting out that check should do the trick.

PR #3398, opened 2020-10-20
Fix Ionization Current

jtrojok:
There was an issue with the ionization current where there was a division by zero. When the electric field was zero (such as before a laser pulse), the ionization current was still called with ionization energy being zero. I fixed with an if-else-statement as you can see in the IonizationCurrent.hpp. The ionization current is now working (tested).

### DIFF
--- a/include/picongpu/particles/ionization/byField/IonizationCurrent/IonizationCurrent.hpp
+++ b/include/picongpu/particles/ionization/byField/IonizationCurrent/IonizationCurrent.hpp
@@ -58,11 +58,15 @@ namespace ionization
         HDINLINE void
         operator()(IonizerReturn retValue, float_X const weighting, T_JBox jBoxPar, ValueType_E eField, T_Acc const & acc, floatD_X const pos)
         {
-            auto ionizationEnergy = weighting * retValue.ionizationEnergy * SI::ATOMIC_UNIT_ENERGY / UNIT_ENERGY; // convert to PIConGPU units
-            /* calculate ionization current at particle position */
-            float3_X jIonizationPar = JIonizationCalc{}(ionizationEnergy, eField);
-            /* assign ionization current to grid points */
-            JIonizationAssignment<T_Acc, T_DestSpecies, simDim>{}(acc, jIonizationPar, pos, jBoxPar);
+            /* If there is no ionization, the ionization energy is zero. In that case, there is no need for an ionization current. */
+            if(retValue.ionizationEnergy != 0.0_X)
+            {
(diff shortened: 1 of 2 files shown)


sbastrakov (commented):
Thanks for the fix.

Just a small question / note. From the changed piece of code it is not obvious (to non-experts) that zero is not a legal value. Perhaps this kind of a check can be moved to somewhere where it's more obvious? I am not sure if this is possible to implement, just to float an idea

jtrojok:
I could also implement the fix in the JIonizationCalc.hpp. It would then return 0 when the ionization energy is zero. But I'd rather let it be the way it is now, because it is way faster right now than it would be if I implemented it in the JIonizationCalc.hpp. Right now nothing happens if the ionization energy is zero, but if I changed it, the assignment would assign a null current. I'd just add an explanatory comment.

include/picongpu/particles/ionization/byField/IonizationCurrent/IonizationCurrent.hpp
```diff
@@ -58,11 +58,16 @@ namespace ionization
         HDINLINE void
         operator()(IonizerReturn retValue, float_X const weighting, T_JBox jBoxPar, ValueType_E eField, T_Acc const & acc, floatD_X const pos)
         {
-            auto ionizationEnergy = weighting * retValue.ionizationEnergy * SI::ATOMIC_UNIT_ENERGY / UNIT_ENERGY; // convert to PIConGPU units
-            /* calculate ionization current at particle position */
-            float3_X jIonizationPar = JIonizationCalc{}(ionizationEnergy, eField);
-            /* assign ionization current to grid points */
-            JIonizationAssignment<T_Acc, T_DestSpecies, simDim>{}(acc, jIonizationPar, pos, jBoxPar);
+            if(retValue.ionizationEnergy == 0)
```
psychocoderHPC, line 61: Could you please write `if(retValue.ionizationEnergy != 0.0_X)`. This will avoid the empty body.

psychocoderHPC (changes requested):

psychocoderHPC:
> Thanks for the fix.
> 
> Just a small question / note. From the changed piece of code it is not obvious (to non-experts) that zero is not a legal value (as this piece does not have division). Perhaps this kind of a check can be moved to somewhere where it's more obvious, that is where the division actually happens? I am not sure if this is possible to implement, just to float an idea. If this is too hard or causes code duplication or there is some other limitation, it is fine to leave this logic as is, just perhaps document it in the comment.

I agree with @sbastrakov. IMO this fix needs at least a comment above the if and the implementation f `JIonizationCalc` needs also a comment above the `operator()`

jtrojok:
Added two comments as requested

sbastrakov (approved):

PrometheusPi (approved):

PR #3399, opened 2020-10-20
Particle thinning plugin with reduction library prototype

KseniaBastrakova:
PIConGPU particle merger plugin, that uses reduction C++ library.
We define  traits for particles' values acess and run "in kernell" version of algorithm


KseniaBastrakova:
expected_actual

psychocoderHPC:
Please do jot forget the documentation (rst files) for the plugin and if this Plugin requires an optional dependency for PIConGPU add the dependency to the coresponding documentation too.

psychocoderHPC:
Let us talk next week about secure integration onto the PIConGPU event system to avoid data races.

psychocoderHPC:
@sbastrakov We should talk about this PR in person.

sbastrakov:
I think practically speaking we should close this PR. If we need such reduction in PIConGPU, we can just implement one of those algorithms on our side and not have that external dependency that no one supports anymore.

PR #3408, opened 2020-10-28
Use Iteration::close() in openPMD plugin

franzpoeschel:
When I wrote the openPMD plugin, this method was not yet available, so the only way to guarantee the release of filehandles was to intermittently close and re-open the `openPMD::Series`. No longer necessary.

### DIFF
--- a/include/picongpu/plugins/openPMD/openPMDWriter.hpp
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.hpp
@@ -1405,28 +1405,9 @@ Please pick either of the following:
             // avoid deadlock between not finished pmacc tasks and mpi calls in
             // openPMD
             __getTransactionEvent().waitForFinished();
-            if( threadParams->fileInfix.empty() )
-            {
-                // using group-based iteration layout (non-default)
-                // keep series open, just flush
-                log< picLog::INPUT_OUTPUT >(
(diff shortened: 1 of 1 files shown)


include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -1405,28 +1405,9 @@ Please pick either of the following:
...
-                /* using file-based iteration layout (default)
-                 * for now, close series between data dumps
-                 * Currently, this is the only way in the openPMD API
-                 * to guarantee that files are closed
-                 * @todo Change this as soon as this is properly implemented
-                 *       in the openPMD API
-                 */
-                log< picLog::INPUT_OUTPUT >( "openPMD: closing series: %1%" ) %
-                    threadParams->fileName;
-                mThreadParams.closeSeries();
```
sbastrakov, line 1428: Do we still need this method to exist at all? (Maybe we do, I'm just not an expert in this part of the code)

franzpoeschel: Yup, we're using it [for restarting](https://github.com/ComputationalRadiationPhysics/picongpu/blob/d3705bc35562cdc963e19cca4cb786f7820f9f90/include/picongpu/plugins/openPMD/openPMDWriter.hpp#L893).

sbastrakov (approved):

Issue #3409, opened 2020-10-28
ADIOS plugin deactivated in the example Hemera profiles


steindev:
I cannot use the ADIOS plugin for data output anymore. I believe this possibility was removed with #3395. If so, the documentation (https://picongpu.readthedocs.io/en/latest/usage/plugins.html) should be updated as well.

sbastrakov:
I think you are confusing two things @steindev . The native (old) ADIOS plugin still exists and uses ADIOS1. Then there is openPMD plugin which can use either HDF5 or ADIOS2. Before recently, it could have also used ADIOS1, and only that was disabled.

steindev:
> The native (old) ADIOS plugin still exists and uses ADIOS1.
But `picongpu --help` does not provide any `--adios` options anymore. But I just see that the `adios/1.13.xxx` module is not loaded anymore. It was thrown out of the `.profile` with  #3393. As the native ADIOS1 plugin is still available, this module should be part of the `.profile`. Blame @PrometheusPi.

sbastrakov:
`picongpu --help` only provides options that are valid for how that particular instance was built (which depends on the system and the used .profile). Maybe the adios1 module should not have been excluded from the example .profile (I don't know the details of that decision). But regardless those files are just examples, as a user you normally take the example and modify for your needs, which can be e.g. adding that module back.

steindev:
Ahh but loading both modules at the same time is not possible at hemera. They did not realize that `adios` and `adios2` should be two separate modules :disappointed:

steindev:
Offline discussion with @PrometheusPi: There is still an open support ticket with our IT. He will take care of it.

sbastrakov:
Solved with #3411, closing.

PR #3411, opened 2020-10-29
add both adios and adios2 module on hemera

PrometheusPi:
This pull request solves the issue #3409. The IT department separated the adios modules to an adios and and an adios2 base module. Both adios(1) and adios2 can thus be loaded in parallel.

### DIFF
--- a/etc/picongpu/hemera-hzdr/fwkt_v100_picongpu.profile.example
+++ b/etc/picongpu/hemera-hzdr/fwkt_v100_picongpu.profile.example
@@ -30,7 +30,8 @@ module load c-blosc/1.14.4
 module load hdf5-parallel/1.8.20-cuda102
 module load libsplash/1.7.0-cuda102
 module load python/3.6.5
-module load adios/2.6.0-cuda102
+module load adios/1.13.1-cuda102
+module load adios2/2.6.0-cuda102
 module load openpmd/0.12.0-cuda102
 
 module load libpng/1.6.35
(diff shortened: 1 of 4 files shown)


sbastrakov (approved):
Thanks!

This is not really related, but also concerns ADIOS on Hemera: do they still have only modules for CUDA?

steindev (approved):

pordyna:
```
Lmod has detected the following error:  Cannot load module "openpmd/0.12.0-cuda102" without
these module(s) loaded:
   adios/2.6.0-cuda102
```
It looks like the dependencies weren't updated after switching the module name to adios2. I have started a ticket with our IT Helpdesk.

steindev:
> ```
> Lmod has detected the following error:  Cannot load module "openpmd/0.12.0-cuda102" without
> these module(s) loaded:
>    adios/2.6.0-cuda102
> ```
> 
> It looks like the dependencies weren't updated after switching the module name to adios2. I have started a ticket with our IT Helpdesk.

Uhh I just tried to load the adios and adios2 module half an hour ago. But I didn't test the `openPMD` module... :roll_eyes:

pordyna:
So It should be fixed now. The CI failed but it looks like it was just some network error? Maybe we just need to restart it.

steindev:
Indeed, I can confirm that it works now as expected.

Regarding the CI, Richard needs to `git commit --amend; git push -f` in order to start the CI again with a new push (that actually changes nothing in the code base)

PrometheusPi:
@steindev I re-triggered the CI as you suggested.

steindev (approved):

Issue #3391, opened 2020-10-16
When generating particles, evaluate density at centers of cells ?


sbastrakov:
This is a side note found while looking at #3377. It is definitely not that reason of it, but can also cause some issues.

Currently inside `CreateDensity` we eventually go over each cell and estimate density at cell corners to compute the number of real particles in the corresponding cell (and then apply our logic to convert to number of macroparticles, but this is irrelevant for this topic). The fact that we estimate density at cell corners to compute the number of real particles seems to have the following two downsides:

- Implementors of density functors with cell-aligned sharp edges have to take care of rounding off (without necessarily knowing how to, unless they know exactly our scheme!). Which is not trivial, and if we estimated at centers more naive implementations would have worked as their authors expected. Of course, for smooth densities like Gaussian this is not a real issue.
- When computing number of real particles in a cell as a product of cell volume and density at a point. The exact number is integral of density over the cell. With the current way we approximate it by a side-rectangle rule, since the density is estimated at a corner. Estimating at center would also be a rectangle rule and same computational complexity, but would give a better central-rectangle rule.

sbastrakov:
IF we decide to change, however we would need to manually go over all density profiles and see if they implicitly assumed the currently used corner evaluation scheme.

steindev:
I agree with **sbastrakov** in both points and support changing the current method to evaluation of the density at the centers of cells.

bussmann:
> IF we decide to change, however we would need to manually go over all density profiles and see if they implicitly assumed the currently used corner evaluation scheme.

These assumptions must be made explicit!

pordyna:
Does `CreateDensity` use `picongpu::traits::FieldPosition`? I think we should avoid the situation when the density in the output is different form the one generated by `CreateDensity`. So we may need to consider changing `FieldPosition` to the center of the cell, for `FieldTmp`, as well. Though this would affect all scalar fields.

steindev:
There is a subtle difference in meaning here. During a simulation and when writing output, the position were density will be computed from the distribution of the macro particles will still be the cell corner. And this is the place were it needs to be in order to be self-consistent with our definition of the Yee cell, i.e.with the positions of the E- and B-field components.

[...]

PR #3414, opened 2020-11-02
Improve clarity of comments concerning density calculation and profile interface

sbastrakov:
This was tiggered by looking at this code while thinking of #3391. This PR does not solve it, merely improves documentation so that it's more clear what are the requirements to density profiles and so what things can and cannot be changed earily. I will provide a follow-up PR shortly to directly address #3391.


sbastrakov:
ping @psychocoderHPC

psychocoderHPC (approved):

PR #3415, opened 2020-11-02
Modify FreeFormulaImpl density profile to evalute a functor in centers of cells

sbastrakov:
Motivation is described in #3391, and user functors are already required to be evaluatable at any point. Thus, the change should merely improve accuracy and potentially avoid some rounding issues, without breaking anything.

This is technically independent from changes in #3414 .


sbastrakov:
ping @psychocoderHPC

psychocoderHPC (approved):

PR #3396, opened 2020-10-19
Improve output of kernel errors in PMacc with PMACC_BLOCKING_KERNEL=ON

sbastrakov:
Now in this mode it catches exceptions that cupla + alpaka can throw, prints kernel information and re-throws.

This change should hopefully help debugging crashes like #3381 by providing proper kernel name and file + line details.

Issue #3377, opened 2020-10-08
Benchmark simulations


HighIander:
### Synopsis:
We performed a set of "simple" ion acceleration test simulations and compared them against smilei. The results are sometimes in full agreement, sometimes they are off by more than  30%.

### Those are the details:
We simulated a transversely periodic box with a transversely plane laser. Target was pure H+electrons. Other parameters;
[coming soon!]
- cell size
- time step
- ionization
- particles per cell
- laser pulse duration
- traditional absorbing boundaries; laser is outside of absorbing cells

We varied a0 (5,10,20) and n (50,100,200). Our FOM is the evolution of the maximum H energy over time (cut-off edge, i.e. not the single one most energetic ion). 

### Results:
**1. Scanning a0:**
![benchmark2_a05_n200](https://user-images.githubusercontent.com/6409838/95456032-fec04c80-096e-11eb-8d36-cf9e6467a79b.png)
![benchmark2_a010_n200](https://user-images.githubusercontent.com/6409838/95456045-054ec400-096f-11eb-9fa1-2a8d3032f7a0.png)
![benchmark2_a20_n200](https://user-images.githubusercontent.com/6409838/95456060-08e24b00-096f-11eb-9d28-85e0e4f56c82.png)

**2. Scanning n0:**
![benchmark2_a010_n50](https://user-images.githubusercontent.com/6409838/95456145-2a433700-096f-11eb-9792-802ec9dba87f.png)
![benchmark2_a010_n100](https://user-images.githubusercontent.com/6409838/95456152-2d3e2780-096f-11eb-84e1-6897d042e544.png)
![benchmark2_a010_n200](https://user-images.githubusercontent.com/6409838/95456045-054ec400-096f-11eb-9fa1-2a8d3032f7a0.png)

**3. Comparison of the spectrum of a0=20,n=200:**
![picongpu9+smilei3_450_thin_quiet](https://user-images.githubusercontent.com/6409838/95456258-5eb6f300-096f-11eb-97f8-7728d0deb912.png)

PrometheusPi:
@HighIander Thanks for providing such a detailed benchmark :+1:. I assume with "quite" you mean quite/regular initialization. Is this correct? Do you observe a similar behavior when choosing random initialization? (A Difference would surprise me, but it would be interesting to see how the start position influence the result in both simulations - and thus might point us into the right direction.)

I do not understand your last plot - I only see red and green lines, no purple orange or blue.

[...]

PR #3416, opened 2020-11-04
binary collisions

pordyna:
This PR brings binary collisions to PIConGPU.  It is still work in progress. 
The collision algorithm described in [Perez 2012](doi.org/10.1063/1.4742167 ) and refined in [Higginson 2020](doi.org/10.1016/j.jcp.2020.109450) is now more or less implemented. It's based on @psychocoderHPC's collisions branch and partially on the realisation in [Smilei](https://github.com/SmileiPIC/Smilei). 

Up to now I have done one of the tests described in `smilei`'s [documentation](https://smileipic.github.io/Smilei/collisions.html) - thermalization. This worked very well. These are electron (blue) and ion (red) temperatures over time for different weightings ratios between both species: 
thermalisation_5to1_fixed 
thermalisation_1to5_fixed
thermalisation_1to1_fixed

Edit:
See the next post.
<s>Now I'm working on the beam relaxation tests, also taken from `smilei`'s benchmarks. For now the results are not correct. I'm currently trying to find out if it's the setup/analysis or if there's maybe still something wrong with the algorithm.</s>
 
More details soon :)

**Note:** For now, my test setups are in the examples dir but in the end we won't include all of them in the final PR. Same thing applies to those flow charts, `.puml` files, in `include/picongpu/particles/collision`.


pordyna:
@HighIander @bussmann @psychocoderHPC @steindev 
Another test from `smilei's benchmarks is ready! 
beam_relaxation_cmakeFlagsSetupNr0
beam_relaxation_cmakeFlagsSetupNr1
beam_relaxation_cmakeFlagsSetupNr2

Results from `smilei`  for comparison  https://smileipic.github.io/Smilei/collisions.html#beam1 .

bussmann:
Could you do a relative comparison to SMILEI?

psychocoderHPC:
> @HighIander @bussmann @psychocoderHPC @steindev
> Another test from `smilei's benchmarks is ready!

@pordyna Could you please short explain for a non-physicist what the curves showing and how I know that this is correct without comparing it directly to the SMILEI plots.

pordyna:
@psychocoderHPC @PrometheusPi  So a short explanation. The test case is an electron beam, electrons with a drift velocity in x direction (solid blue line)  and a very small temperature, colliding with an ion background. Ions are here just 10 times heavier than electrons so that everything happens a bit faster. As the electrons scatter on ions they lose their drift velocity, some of it is transformed into their transversal temperature (mean transversal speed grows, dashed blue line in these plots ) and some into the ions. With time ions get a small drift in the x direction (red line, mean velocity in x direction for ions).

The black lines show the theoretical values, the formulas are taken from the [NRL Plasma Formulary](https://www.nrl.navy.mil/ppd/sites/www.nrl.navy.mil.ppd/files/pdfs/NRL_Formulary_2019.pdf) (page 31). These assume a Maxwellian distribution for the ions but as the ions gain their drift their velocity distribution becomes less and less Maxwellian. So that's why the theoretical predictions are accurate only in the initial phase and deviate, at later times, from the simulated values.

pordyna:
@bussmann I could run `smilei` with the exact same number of cells, for matching statistics, and plot them in the same plot as PIConGPU. Would that be helpful?  Also, @paschk31 wants to do some variation of the FoilLTC example and compare it with an identical setup in `smilei`. i think, something like this https://github.com/ComputationalRadiationPhysics/picongpu/issues/3377 but with collisions on.

include/picongpu/particles/collision/IntraSpecies.hpp
```diff
@@ -0,0 +1,330 @@
...
+                    for( uint32_t i = 0; i < sizeAll; i += 2 )
+                    {
+                        uint32_t const sizeOdd = sizeAll / 2;
+                        uint32_t const iOdd = ( i / 2 ) % sizeOdd;
+                        uint32_t
+                            duplications =
+                            ( iOdd == 0 && sizeAll % 2 == 1 ) ? 2 : 1;
+                        duplications
+                            *=
+                            2u; // TODO move this somewhere else
```
psychocoderHPC, line 224: Please put this and the code above into a helper function/method

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,494 @@
...
+namespace particles
+{
+namespace collision
+{
+namespace binary
+{
+namespace acc
+{
+    using namespace pmacc;
+    struct DefaultAlg
```
psychocoderHPC, line 44: This functor and file should not be called `DefaultAlg`, please give it a name that reflects the method used.

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,494 @@
...
+            float_X gammaComs,
+            float3_X comsVelocity
+        )
+        {
+            float_X dot = pmacc::math::dot( comsVelocity, labMomentum );
+            float_X val = mass * gamma - dot / ( c*c );
+            return gammaComs * val;
+        }
+
+        static DINLINE float_X calcCosXi(
```
psychocoderHPC, line 130: please add a docstring what this function is doing. If the `s12` and `u` are limited to a defined range please add the valid range too.

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,494 @@
...
+            typename T_Par1
+        >
+        DINLINE void operator( )(
+            T_Context const & ctx,
+            T_Par0 & par0,
+            T_Par1 & par1
+        ) const
+        {
+            //feenableexcept(FE_INVALID | FE_OVERFLOW | FE_DIVBYZERO);
+            float_X const weight0 = par0[ weighting_ ]
```
psychocoderHPC, line 230: Maybe work with names like `normalizedWeighting` to avoid confusing

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,494 @@
...
+
+            //  f0 * f1 * f2^2
+            //  is equal  s12 * (n12/(n1*n2)) from [Perez 2012]
+            float_X comsMomentum0Abs2 = pmacc::math::abs2( comsMomentum0 );
+
+            if( comsMomentum0Abs2 == 0.0_X &&
+                ( comsMomentum0 != float3_X{ 0.0_X, 0.0_X, 0.0_X } )) return;
+
+            float_X s12Factor0 = ( DELTA_T * coulombLog * charge0 *
+                charge0 * charge1 * charge1 ) / ( 4.0_X * PI * EPS0 * EPS0
```
psychocoderHPC, line 327: please check if PI in PIConGPU is by default `float64`, if so please use `float_X(PI)`

psychocoderHPC (changes requested):

psychocoderHPC:
Note: we have now clang-format integrated into PIConGPU. After adding my review comments please rebase and format your PR.

include/picongpu/particles/collision/binary/DefaultAlg.hpp
sbastrakov: For PI we already have a way to get it at any type.

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,499 @@
...
+            float_X gamma0,
+            float_X gamma1,
+            float_X coeff0,
+            float_X coeff1,
+            float_X gammaComs
+        )
+        {
+            float_X val = ( mass0 * gamma0 + mass1 * gamma1 )
+                * comsMomentumMag0;
+            val = val / ( coeff0 * coeff1 * gammaComs );
```
psychocoderHPC, line 116: Please check if `( coeff0 * coeff1 * gammaComs )` can become zero and produce `NaN`

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,499 @@
...
+            float_X sinPhi, cosPhi;
+            pmacc::math::sincos( phi, sinPhi, cosPhi);
+            float_X sinXi = math::sqrt( 1 - cosXi * cosXi );
+
+            float3_X finalVec;
+            float_X const pAbs = math::sqrt( pmacc::math::abs2( p ) );
+            float_X const pPerp = math::sqrt( p.x( ) * p.x( ) + p.y( )
+                * p.y( ) );
+            // TODO chose a better limmit?
+            if( pPerp > std::max( std::numeric_limits< float_X >::epsilon, 1.0e-10_X ) *  pAbs )
```
psychocoderHPC, line 198: if `pAbs` == 0.0 than `pPerp` < epsilon will be allowed too and can result in `NaN` later.

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,499 @@
...
+            //feenableexcept(FE_INVALID | FE_OVERFLOW | FE_DIVBYZERO);
+            float_X const weight0 = par0[ weighting_ ]
+                / particles::TYPICAL_NUM_PARTICLES_PER_MACROPARTICLE;
+            float_X const weight1 = par1[ weighting_ ]
+                / particles::TYPICAL_NUM_PARTICLES_PER_MACROPARTICLE;
+
+            float3_X const labMomentum0 = par0[ momentum_ ] / weight0;
+            float3_X const labMomentum1 = par1[ momentum_ ]  / weight1;
+
+            float_X const mass0 = picongpu::traits::attribute::getMass(
```
psychocoderHPC, line 243: Note: We should check on the host site where the kernel is called that both species will not be `photon` like particles, because those have the mass zero.

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,499 @@
...
+                mass1,
+                gamma0,
+                gamma1,
+                coeff0,
+                coeff1,
+                gammaComs
+            );
+            // [Perez 2012] (21) ( without n1*n2/n12 )
+            float_X s12Max = math::pow(4 * PI / 3, 1./3.) * DELTA_T *
+                (mass0 + mass1) /  pmacc::math::max( mass0 * densitySqCbrt0,
```
psychocoderHPC, line 365: please check that this can not be zero to avoid `NaN`

psychocoderHPC:
@pordyna What is the state of this PR? Is something blocking us from bringing this into dev?

pordyna:
@psychocoderHPC I'm still fixing https://github.com/ComputationalRadiationPhysics/picongpu/issues/3461.

include/picongpu/particles/collision/InterCollision.hpp
```diff
@@ -0,0 +1,358 @@
...
+                        mapper,
+                        deviceHeap->getAllocatorHandle(),
+                        RNGFactory::createHandle(),
+                        CollisionFunctor(currentStep),
+                        coulombLog,
+                        Filter0(),
+                        Filter1());
+
+                    // Release particle data:
+                    // TODO: can I relase it like this? Or do I have to synchronize, make KERNEL call blocking?
```
psychocoderHPC, line 350: Yes this is correct, currently `release` is doing nothing but is already added to the code for future use.

include/picongpu/particles/collision/IntraCollision.hpp
```diff
@@ -0,0 +1,232 @@
...
+                            densityArray[linearIdx],
+                            densityArray[linearIdx],
+                            potentialPartners,
+                            coulombLog);
+                        for(uint32_t i = 0; i < sizeAll; i += 2)
+                        {
+                            uint32_t const sizeOdd = sizeAll / 2;
+                            uint32_t const iOdd = (i / 2) % sizeOdd;
+                            uint32_t duplications = (iOdd == 0 && sizeAll % 2 == 1) ? 2 : 1;
+                            duplications *= 2u; // TODO move this somewhere else
```
psychocoderHPC, line 159: Please move this and the code above into a helper function/method

include/picongpu/particles/collision/binary/DefaultAlg.def
```diff
@@ -0,0 +1,36 @@
...
+
+namespace picongpu
+{
+    namespace particles
+    {
+        namespace collision
+        {
+            namespace binary
+            {
+                struct DefaultAlg;
```
psychocoderHPC, line 31: Please use a name that reflects the method used

share/picongpu/examples/CollisionsThermalisation/README.rst
```diff
@@ -0,0 +1,2 @@
+Collisions: Testing example thermalization
```
psychocoderHPC, line 1: Please move the example into `share/picongpu/tests`

share/picongpu/examples/CollisionsBeamEquilibration/include/picongpu/param/precision.param
```diff
@@ -0,0 +1,56 @@
...
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file
+ *
+ * Define the precision of typically used floating point types in the
+ * simulation.
+ *
+ * PIConGPU normalizes input automatically, allowing to use single-precision by
```
psychocoderHPC, line 25: please add only param files that contain changes compared to the default param files.

include/picongpu/param/collision.param
```diff
@@ -0,0 +1,43 @@
+/* Copyright 2020 Rene Widera, Pawel Ordyna
```
psychocoderHPC, line 1: Please change everywhere `Copyright 2020` into `Copyright 2020 - 2021`

include/picongpu/particles/collision/Collider.def
```diff
@@ -0,0 +1,80 @@
+/* Copyright 2019 Rene Widera
```
psychocoderHPC, line 1: please change to `Copyright 2019 - 2021`

psychocoderHPC (changes requested):

share/picongpu/examples/CollisionsBeamEquilibration/README.rst
```diff
@@ -0,0 +1,2 @@
+Collisions: Testing example thermalization
+============================
```
psychocoderHPC, line 2: This line is to short, must be as long as the headline.

include/pmacc/algorithms/kernel/Histogram.hpp
```diff
@@ -0,0 +1,82 @@
...
+                uint32_t numParticlesInSupercell = superCell.getNumParticles();
+
+                onlyMaster([&](uint32_t const, uint32_t const) { firstFrame = pb.getFirstFrame(superCellIdx); });
+
+                /* loop over all particles in the frame */
+                ForEachIdx<ParticleDomCfg> forEachParticle(workerIdx);
+
+                forEachParticle([&](uint32_t const linearIdx, uint32_t const) { nppc[linearIdx] = 0u; });
+
+                __syncthreads();
```
psychocoderHPC, line 64: should be `cupla::__syncthreads()` else it is only running together with CUDA

include/pmacc/algorithms/kernel/Histogram.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+                for(uint32_t i = 0; i < numParticlesInSupercell; i += frameSize)
+                {
+                    forEachParticle([&](uint32_t const linearIdx, uint32_t const idx) {
+                        if(i + linearIdx < numParticlesInSupercell)
+                        {
+                            auto& particle = lastFrame[linearIdx];
```
psychocoderHPC, line 71: BUG: I would say this is a bug! it should not be valid to use a reference here.

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,652 @@
...
+                            float_COLL gammaComs,
+                            float_COLL factorA,
+                            float3_COLL comsVelocity)
+                        {
+                            // (13) in [Perez 2012]
+                            // TODO: try upgrading this to 64bit as well
+                            float_COLL dot = pmacc::math::dot(comsVelocity, comsMomentum);
+                            float_COLL factor = (factorA * dot + coeff * gammaComs);
+                            float3_COLL diff = factor * comsVelocity;
+                            PMACC_DEVICE_ASSERT(std::isfinite((comsMomentum + diff)[0]));
```
psychocoderHPC, line 134: This should not run on the device because `std::isfinite` is host only. For cuda there is a `isfinite` but this is not available in alpaka.

psychocoderHPC (changes requested):

include/picongpu/particles/collision/binary/DefaultAlg.hpp
pordyna: what about isnan and simiilar? std::isfinite was working fine but probably because it's using cuda implementation directly.  I guess the reason why it is not in alpaka is that it wouldn't work on amd?

include/pmacc/algorithms/kernel/Histogram.hpp
pordyna: @psychocoderHPC  Wait ! This shouldn't be a part of my PR? I didn't write it... Maybe I did sth weird with git...

sbastrakov: This is not actual runnable code here, it's just something inside a `struct`. This pattern of usage `auto & result = function();` is not necessarily a bug, but should be avoided as it causes confusion and does not give any benefit

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,652 @@
...
+                                    //                                    mass1: %e, " "gamma0: %e, gamma1: %e",
+                                    //                                    gammaComs,
+                                    //                                    comsVelocityAbs2,
+                                    //                                    labMomentum0,
+                                    //                                    labMomentum1,
+                                    //                                    mass0,
+                                    //                                    mass1,
+                                    //                                    gamma0,
+                                    //                                    gamma1);
+                                    a = 1 / a;
```
psychocoderHPC, line 286: should be `1.0_COLL`

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,652 @@
...
+                                    else
+                                        return 1.0_COLL;
+                                }
+                                else
+                                    return cosXi;
+                            }
+                            else
+                            {
+                                // + 1 in Perez 2012 but cos in [-1,1]. Smilei uses -1.
+                                return 2 * u - 1;
```
psychocoderHPC, line 307: should be `2.0_COLL` and `1.0_COLL`

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,652 @@
...
+                         * @param p momentum in the COM frame
+                         * @param cosXi cosine of the scattering angle
+                         * @param phi azimuthal scattering angle from [0, 2pi]
+                         */
+                        static DINLINE float3_COLL
+                        calcFinalComsMomentum(float3_COLL const p, float_COLL const cosXi, float_COLL const phi)
+                        {
+                            float_COLL sinPhi, cosPhi;
+                            pmacc::math::sincos(phi, sinPhi, cosPhi);
+                            float_COLL sinXi = math::sqrt(1 - cosXi * cosXi);
```
psychocoderHPC, line 322: should be `1.0_COLL`

include/pmacc/algorithms/kernel/Histogram.hpp
sbastrakov: Some small playground how it works or doesn't wrt particles and frames: https://godbolt.org/z/hjsMhj

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,652 @@
...
+                                mass0,
+                                mass1,
+                                gamma0,
+                                gamma1,
+                                coeff0,
+                                coeff1,
+                                gammaComs);
+                            // [Perez 2012] (21) ( without n1*n2/n12 )
+                            float_COLL s12Max
+                                = math::pow(4.0_COLL * pmacc::math::Pi<float_COLL>::value / 3._COLL, 1.0 / 3.0)
```
psychocoderHPC, line 488: should be `1.0_COLL` and `3.0_COLL`

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,652 @@
...
+                        DataSpace<simDim> const& offset,
+                        T_WorkerCfg const& workerCfg,
+                        float_X const& density0,
+                        float_X const& density1,
+                        uint32_t const& potentialPartners,
+                        float_X const& coulombLog) const
+                    {
+                        using namespace picongpu::particles::collision::precision;
+                        return acc::DefaultAlg(
+                            math::pow(precisionCast<float_COLL>(density0), 2. / 3.),
```
psychocoderHPC, line 637: should be `2.0_COLL / 3.0_COLL`, in the next line too

include/picongpu/particles/collision/detail/ListEntry.hpp
```diff
@@ -0,0 +1,292 @@
...
+                                        maxTries);
+#endif
+                                }
+                            }
+                            if(ptrToIndicies == nullptr)
+                            {
+#if BOOST_COMP_HIP
+                                // do nothing
+#elif BOOST_LANG_CUDA
+                                __trap();
```
psychocoderHPC, line 87: maybe use the new device assert

psychocoderHPC (changes requested):

include/picongpu/particles/collision/binary/DefaultAlg.hpp
psychocoderHPC: everything which is not `constexpr` e.g https://en.cppreference.com/w/cpp/numeric/math/isnan should not be available on GPU.

include/picongpu/particles/collision/binary/DefaultAlg.hpp
```diff
@@ -0,0 +1,652 @@
...
+                            // todo:: maybe this also with doubles?
+                            float3_COLL const comsVelocity
+                                = (labMomentum0 + labMomentum1) / (mass0 * gamma0 + mass1 * gamma1);
+                            float_COLL const comsVelocityAbs2 = pmacc::math::abs2(comsVelocity);
+                            float3_COLL comsMomentum0;
+                            float_COLL gammaComs, factorA, coeff0, coeff1;
+
+                            if(comsVelocityAbs2 != 0.0_COLL)
+                            {
+                                gammaComs = 1.0_COLL / math::sqrt(1.0_COLL - comsVelocityAbs2 / c / c);
```
sbastrakov, line 404: Discussed in VC regarding potential floating-point issues in this part of the code.

- Ideally, if possible it would be better to rewrite these equations to use momentums, not velocities, and use standard `picongpu::gamma()` to compute gamma from momentum as it should have no numerical issues
- As a quick potential fix to this expression (and similar one elsewhere if you have them) is to replace `1.0_COLL - comsVelocityAbs2 / c / c` with `(1.0_COLL - comsVelocityAbs2 / c) * (1.0_COLL + comsVelocityAbs2 / c)`. It should introduce no additional issues except a couple more operations, and be much better in case v is close to c. So that's a more safe way to write it essentially for free while you are debugging

include/picongpu/particles/collision/IntraCollision.hpp
```diff
@@ -0,0 +1,232 @@
...
+                        uint32_t potentialPartners = sizeAll - 1u + sizeAll % 2u;
+                        collisionFunctorCtx[idx] = collisionFunctor(
+                            acc,
+                            localSuperCellOffset,
+                            WorkerCfg<T_numWorkers>{workerIdx},
+                            densityArray[linearIdx],
+                            densityArray[linearIdx],
+                            potentialPartners,
+                            coulombLog);
+                        for(uint32_t i = 0; i < sizeAll; i += 2)
```
psychocoderHPC, line 154: BUG??? should be `for(uint32_t i = 0; i + 2 <= sizeAll; i += 2)` else the last particle pare could be outside of memory.
Same happened in InterCollision.

psychocoderHPC: ` auto parOdd = detail::getParticle(pb, firstFrame, listAll[iOdd * 2 + 1]);`
Will access particle `i+1`

psychocoderHPC: IN my original code this was: https://github.com/psychocoderHPC/picongpu/blob/48359030e9ea8cc34038cb9da3a3a9f25ba382da/include/picongpu/particles/collision/IntraSpecies.hpp#L359

Maybe your change work because of the next line 
```
uint32_t const sizeOdd = sizeAll / 2;
uint32_t const iOdd = (i / 2) % sizeOdd;
```
but from pure looking to the loop it is hard to see if it is safe.

include/picongpu/particles/collision/binary/DefaultAlg.hpp
pordyna: I reversed the the condition and the content of if and else. So for `pAbs = pPerp = 0` the result is (0,0,0).  An alternative would be to leave the momentum unchanged in that case.

include/picongpu/particles/collision/binary/DefaultAlg.hpp
pordyna: @psychocoderHPC how do I do that?

include/picongpu/particles/collision/detail/ListEntry.hpp
pordyna: but that would be triggered only when compiling without NDEBUG? That's not the usual case.  But I could just use `assert(0);` directly since it's supported by both HIP and CUDA.

include/picongpu/particles/collision/IntraCollision.hpp
pordyna: I simplified the loop significantly. Now it looks more like what you did.

include/picongpu/particles/collision/binary/DefaultAlg.hpp
pordyna: @sbastrakov  I don't really understand why it's better when I have to calculate `comsVelocityAbs` with `sqrt(comsVelocityAbs2)` anyway but I trust you with this.

pordyna:
@psychocoderHPC A quick question to your part of the code. Why do we need to use `ApplyGuard` in `Collider.hpp`?

psychocoderHPC:
> @psychocoderHPC A quick question to your part of the code. Why do we need to use `ApplyGuard` in `Collider.hpp`?

It has to do with the fact that you substitute the placeholder `_1` in a lambda which also contains the placeholder `_1` which will be internally substituted later.

```
X = f( guard(Foo(_1)), _1);
Y = apply(X,myvalue); // do not substitute _1 from Foo
// Y == f(Foo(_1), myvalue)
```

If you do not guard `Foo` you will end up with

```
X = f( Foo(_1), _1);
Y = apply(X,myvalue); 
// Y == f(Foo(myvalue), myvalue)
```

Something like this is the reason.

pordyna:
@sbastrakov @psychocoderHPC  Another question. I think the changes in `kernelEvents.hpp` are take from one of you, I think @sbastrakov if I remeber correctly. So this gives some better info on crush when `PMACC_BLOCKING_KERNEL` is on. Should it be really a part of this PR? It doesn't really have anything to do with collisions.

psychocoderHPC:
> @sbastrakov @psychocoderHPC Another question. I think the changes in `kernelEvents.hpp` are taken from one of you, I think @sbastrakov if I remeber correctly. So this gives some better info on crush when `PMACC_BLOCKING_KERNEL` is on. Should it be really a part of this PR? It doesn't really have anything to do with collisions.

NO, this should not part of this PR. If this is useful it should be provided via a separate PR.
Could you please squash your PR and remove unchanged files in the example e.g. precision.param

sbastrakov:
There is a merged PR on better error message output #3396. In case you used exactly the same stuff, it's enough to make it rebaseable without conflicts. In case you added something on top of that, please provide it as a separate PR as @psychocoderHPC suggested. If you are unsure, I can have a look at this part and see if it's same as currently in `dev` or different.

pordyna:
@sbastrakov it's a different file but I think it's doing the same thing. I will just roll back these changes.

include/picongpu/particles/collision/detail/ListEntry.hpp
psychocoderHPC: OK let us keep it for now until we have a better error notification mechanism in PIConGPU.

include/pmacc/functor/Interface.hpp
```diff
@@ -173,17 +173,18 @@ namespace pmacc
              * @return an instance of the user functor wrapped by the accelerator
              *         functor interface
              */
-            template<typename T_OffsetType, uint32_t T_numWorkers, typename T_Acc>
+            template<typename T_OffsetType, uint32_t T_numWorkers, typename T_Acc, typename... T_Args>
```
psychocoderHPC, line 176: Please add the new argument to the doxygen docu.

share/picongpu/tests/CollisionsBeamRelaxation/README.rst
```diff
@@ -0,0 +1,2 @@
+Collisions: Testing example beam relaxation
+==============================================
```
psychocoderHPC, line 2: The line should be as long as the headline.

Please add some documentation.

psychocoderHPC (changes requested):
Please run the following commands on your terminal:

```
git rebase -i HEAD~1
#- change the work `pick` to `r`
#- add to the commit message `Co-authored-by: René Widera <r.widera@hzdr.de>`
# store and force push again
```

Parts of this work were written by me, this will add me as co-author.

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
```diff
@@ -0,0 +1,513 @@
...
+                        static DINLINE float3_COLL comsToLab(
+                            float3_COLL comsMomentum,
+                            float_COLL mass,
+                            float_COLL coeff,
+                            float_COLL gammaComs,
+                            float_COLL factorA,
+                            float3_COLL comsVelocity)
+                        {
+                            // (13) in [Perez 2012]
+                            // TODO: try upgrading this to 64bit as well
```
PrometheusPi, line 125: Isn't it already 64 bit, since `float_COLL` is `float_64`, or do you mean that is should stay 64 bit even if `float_COLL` becomes `float_32`?

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
```diff
@@ -0,0 +1,513 @@
...
+                            float_COLL gamma0,
+                            float_COLL gamma1,
+                            float_COLL coeff0,
+                            float_COLL coeff1,
+                            float_COLL gammaComs)
+                        {
+                            float_COLL val = (mass0 * gamma0 + mass1 * gamma1) * comsMomentumMag0;
+                            val = val / (coeff0 * coeff1 * gammaComs); // TODO: null division?
+                            // TODO:
+                            // this is actually not the relative velocity! since we are missing (1 + v1v2/c^2).
```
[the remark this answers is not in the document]
PrometheusPi: This is a factor 2 in the worst case

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
```diff
@@ -0,0 +1,513 @@
...
+                            float_COLL dot = pmacc::math::dot(comsVelocity, labVelocity);
+
+                            float_COLL val = gammaComs * gamma - dot * (gammaComs * gamma / (c * c));
+                            val *= mass;
+                            return val;
+                        }
+
+                        /* Calculate the cosine of the scattering angle.
+                         *
+                         * The probabilty distribution  for the cosine depends on @f[ s_{12} @f]. The returned vale
```
PrometheusPi, line 186:
```suggestion
                         * The probability distribution for the cosine depends on @f[ s_{12} @f]. The returned vale
```

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
```diff
@@ -0,0 +1,513 @@
...
+                        static DINLINE float3_COLL
+                        calcFinalComsMomentum(float3_COLL const p, float_COLL const cosXi, float_COLL const phi)
+                        {
+                            float_COLL sinPhi, cosPhi;
+                            pmacc::math::sincos(phi, sinPhi, cosPhi);
+                            float_COLL sinXi = math::sqrt(1.0_COLL - cosXi * cosXi);
+
+                            // (12) in [Perez 2012]
+                            float3_COLL finalVec;
+                            // TODO: sqrt returns a pmacc complex number. cast to float?
```
PrometheusPi, line 257: Why does it return a complex number at all? 
Why is `pmacc::math::abs2(p)` not positive?

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
```diff
@@ -0,0 +1,513 @@
...
+                            float_COLL const pAbs = math::sqrt(pmacc::math::abs2(p));
+                            float_COLL const pPerp = math::sqrt(p.x() * p.x() + p.y() * p.y());
+                            // TODO chose a better limmit?
```
PrometheusPi, line 260:
```suggestion
                            // TODO chose a better limit?
```

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
```diff
@@ -0,0 +1,513 @@
...
+                                    particles::TYPICAL_NUM_PARTICLES_PER_MACROPARTICLE,
+                                    par0));
+                            float_COLL const charge1
+                                = precisionCast<float_COLL>(picongpu::traits::attribute::getCharge(
+                                    particles::TYPICAL_NUM_PARTICLES_PER_MACROPARTICLE,
+                                    par1));
+                            float_COLL const gamma0 = picongpu::gamma<float_COLL>(labMomentum0, mass0);
+                            float_COLL const gamma1 = picongpu::gamma<float_COLL>(labMomentum1, mass1);
+
+                            //                            PMACC_DEVICE_ASSERT_MSG(
```
PrometheusPi, line 320: can this be removed?

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
```diff
@@ -0,0 +1,513 @@
...
+                            //                                math::sqrt(pmacc::math::abs2(labMomentum0)) / gamma0 /
+                            //                                mass0 < c, "labVelocity0: %e",
+                            //                                sqrt(pmacc::math::abs2(labMomentum0)) / gamma0 / mass0);
+                            //                            PMACC_DEVICE_ASSERT_MSG(
+                            //                                math::sqrt(pmacc::math::abs2(labMomentum1)) / gamma1 /
+                            //                                mass1 < c, "labVelocity1: %e",
+                            //                                sqrt(pmacc::math::abs2(labMomentum1)) / gamma1 / mass1);
+                            // [Perez 2012] (1)
+                            // todo:: maybe this also with doubles?
```
PrometheusPi, line 329: what do you mean with doubles? Again a safeguard against `float_COLL` being `float_32`?

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
```diff
@@ -0,0 +1,513 @@
...
+                            float3_COLL const comsVelocity
+                                = (labMomentum0 + labMomentum1) / (mass0 * gamma0 + mass1 * gamma1);
+                            float_COLL const comsVelocityAbs2 = pmacc::math::abs2(comsVelocity);
+                            float3_COLL comsMomentum0;
+                            float_COLL gammaComs, factorA, coeff0, coeff1;
+
+                            if(comsVelocityAbs2 != 0.0_COLL)
+                            {
+                                // written as (1-v)(1+v) rather than (1-v^2) for better performance when v close to c
```
PrometheusPi, line 338: Is this a todo?

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
```diff
@@ -0,0 +1,513 @@
...
+                                float_COLL const comsVelocityAbs = math::sqrt(comsVelocityAbs2);
+                                gammaComs = 1.0_COLL
+                                    / math::sqrt((1.0_COLL - comsVelocityAbs / c) * (1.0_COLL + comsVelocityAbs / c));
+                                // used later for comsToLab:
+                                factorA = (gammaComs - 1.0_COLL) / comsVelocityAbs2;
+
+                                // Stared gamma times mass, from [Perez 2012].
+                                coeff0 = coeff(labMomentum0, mass0, gamma0, gammaComs, comsVelocity);
+                                // PMACC_DEVICE_ASSERT(coeff0 > 0.0_COLL);
```
PrometheusPi, line 347: Do you need this debug code here?

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
```diff
@@ -0,0 +1,513 @@
...
+                                // gamma^* . mass
+                                // if(coeff0 == 0.0_COLL) return;
+                                coeff1 = coeff(labMomentum1, mass1, gamma1, gammaComs, comsVelocity);
+                                // if(coeff1 == 0.0_COLL) return;
```
PrometheusPi, line 351: same here

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
```diff
@@ -0,0 +1,513 @@
...
+                                // PMACC_DEVICE_ASSERT(coeff1 > 0.0_COLL);
+                                // (2) in [Perez 2012]
+                                comsMomentum0
+                                    = labToComs(labMomentum0, mass0, gamma0, gammaComs, factorA, comsVelocity);
+                            }
+                            else
+                            {
+                                // From smilei implementation, for v_coms^2 = 0
```
PrometheusPi, line 359: Do we keep this reference?

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
```diff
@@ -0,0 +1,513 @@
...
+                                = gammaComs * math::sqrt(comsMomentum0Abs2) / (mass0 * gamma0 + mass1 * gamma1);
+                            float_COLL const s12Factor2 = coeff0 * coeff1 * c * c / comsMomentum0Abs2 + 1.0_COLL;
+                            // Statistical part from [Higginson 2020],
+                            // corresponds to n1*n2/n12 in [Perez 2012]:
+                            float_COLL const s12Factor3 = potentialPartners
+                                * pmacc::math::max(normalizedWeight0, normalizedWeight1) * WEIGHT_NORM_COLL
+                                / static_cast<float_COLL>(duplicationCorrection) / CELL_VOLUME_COLL;
+                            float_COLL s12n = s12Factor0 * s12Factor1 * s12Factor2 * s12Factor2 * s12Factor3;
+
+                            // Low Temeprature correction:
```
PrometheusPi, line 392:
```suggestion
                            // low temperature correction:
```

PrometheusPi (dismissed):
@pordyna Excellent job! I found a few typos and there are some questions. 

If you use "Commit suggestion" please deactivate auto check via adding the label "CI:no-compile" temporally.

pordyna:
@psychocoderHPC I just pushed a new commit with the documentation. Also how can a png image contain EOL white spaces? :sweat_smile: That's why the CI is failing now...

include/picongpu/particles/collision/IntraCollision.hpp
```diff
@@ -0,0 +1,253 @@
...
+             * @tparam T_Params A struct defining `coulombLog` for the collisions.
+             * @tparam T_FilterPair A pair of particle filters, each for each species
+             *     in the colliding pair.
+             * @tparam T_Species0 Colliding species.
+             * @tparam T_Species1 2nd colliding species.
+             */
+            template<typename T_CollisionFunctor, typename T_Params, typename T_FilterPair, typename T_Species>
+            struct DoIntraCollision;
+
+            // A single template specialization. This ensures that the code want compile if the FilterPair contains two
```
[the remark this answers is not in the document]
pordyna: ah, it's won't :D

include/picongpu/particles/collision/WithPeer.hpp
```diff
@@ -0,0 +1,108 @@
...
+                };
+            } // namespace detail
+
+            /* Runs the binary collision algorithm for a pair of colliding species.
+             *
+             * These struct choses the InterCollision algorithm if the colliding
+             * species are two different species and the IntraSCollision algorithm if
+             * they are identical.
+             *
+             * @tparam T_CollisionFunctor A binary particle functor defining a single
```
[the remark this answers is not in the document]
pordyna: What I meant, was that the functor only defines a collisions between two macro particles but not the whole algorithm ( so to say, it doesn't change the matching process). I  rewrote it to:
> A binary particle functor defining a collision between two macro particles.

include/picongpu/particles/collision/detail/ListEntry.hpp
```diff
@@ -0,0 +1,292 @@
...
+                    template<typename T_Acc, typename T_RngHandle>
+                    DINLINE void shuffle(T_Acc const& acc, T_RngHandle& rngHandle)
+                    {
+                        using UniformUint32_t = pmacc::random::distributions::Uniform<uint32_t>;
+                        auto rng = rngHandle.template applyDistribution<UniformUint32_t>();
+                        // shuffle the particle lookup table
+                        for(uint32_t i = size; i > 1; --i)
+                        {
+                            /* modulo is not perfect but okish,
+                             * because of the loop head mod zero is not possible
```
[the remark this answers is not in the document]
pordyna: It's  actually @psychocoderHPC's comment. Maybe he can say sth more.

include/picongpu/particles/collision/detail/ListEntry.hpp
```diff
@@ -0,0 +1,292 @@
...
+                     */
+                    DINLINE void swap(uint32_t& v0, uint32_t& v1)
+                    {
+                        uint32_t tmp = v0;
+                        v0 = v1;
+                        v1 = tmp;
+                    }
+                };
+
+                // TODO: simplify (maybe crate a class and inject all the required stuff as private references?), Check
```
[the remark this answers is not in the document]
pordyna: Yes, I started this a separate branch but didn't had time to finish it.

include/pmacc/meta/conversion/ApplyGuard.hpp
```diff
@@ -0,0 +1,60 @@
+/* Copyright 2019-2021 Rene Widera
```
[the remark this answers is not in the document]
pordyna: I took it over from @psychocoderHPC's branch and I didn't change anything here.

share/picongpu/tests/CollisionsThermalisation/include/picongpu/param/memory.param
```diff
@@ -0,0 +1,117 @@
+/* Copyright 2013-2021 Axel Huebl, Rene Widera, Benjamin Worpitz, Pawel Ordyna
```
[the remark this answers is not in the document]
pordyna: `SuperCellSize`

share/picongpu/tests/CollisionsThermalisation/include/picongpu/param/particle.param
```diff
@@ -0,0 +1,128 @@
...
+
+namespace picongpu
+{
+    namespace particles
+    {
+        /** a particle with a weighting below MIN_WEIGHTING will not
+         *      be created / will be deleted
+         *
+         *  unit: none */
+        constexpr float_X MIN_WEIGHTING = 0.01;
```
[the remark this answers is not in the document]
pordyna: Well I think the usual scenario for this model is that the weighting is >1. @HighIander and @n01r have some experience with running collisions in other PIC codes. They may now how many macro particles per cell one needs given a certain density. But I'm pretty sure in the end each macro particle will represent more than one real particle.

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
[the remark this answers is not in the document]
pordyna: So this whole low temperature correction is a little weird. We actually never check if we are really non-relativistic  when we use it. It is just assumed that all relativistic cases will end up in a smaller `s12` than `s12Max`. The relative velocity is calculated for every collisions but in theory it should be used only in a non-relativistc case as it is only used when calculating `s12Max`. I was confused by the missing factor and that's why I commented it here. 

Anyway that is how other codes do it. But I think we should think through the low temp limit  at some point and maybe check if the gammas are really that close to 1 etc. . Also `smilei` has a optional histogram for `s12` so one can verify that it is in an o.k. range. Sth like that could also be useful.

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
pordyna: I don't think that this is really important. It's that the return type for sqrt is comples but the result is never complex. I removed the comment.

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
pordyna: I moved this one line down. Not really, it's already like that.

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
pordyna: We could. I actually never verified it, just copied it from `smilei`. It's not in the referenced paper so that way we know where it comes from.

share/picongpu/tests/CollisionsBeamRelaxation/include/picongpu/param/density.param
```diff
@@ -0,0 +1,59 @@
...
+         *
+         * This is often taken as reference maximum density in normalized profiles.
+         * Individual particle species can define a `densityRatio` flag relative
+         * to this value.
+         *
+         * unit: ELEMENTS/m^3
+         */
+        // reference frequency like in smilei simulations
+        constexpr float_64 REFERENCE_OMEGA_SI = 2.0 * PI * 3.0e8 * 1.0e6;
+        // reference density from smilei simulations
```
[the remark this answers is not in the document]
pordyna: Well I scaled all parameters with `REFERENCE_OMEGA_SI` since that's how `smilei` is introducing absolute scales into similations. I wanted to make it easy to compare `smilei` and `PIConGPU` inputs because the setup is taken over from `smilei`'s benchmarks.

share/picongpu/tests/CollisionsBeamRelaxation/include/picongpu/param/memory.param
```diff
@@ -0,0 +1,117 @@
+/* Copyright 2013-2021 Axel Huebl, Rene Widera, Benjamin Worpitz, Pawel Ordyna
```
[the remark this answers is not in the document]
pordyna: SuperCellSize

share/picongpu/tests/CollisionsThermalisation/cmakeFlags
```diff
@@ -0,0 +1,50 @@
...
+################################################################################
+# add presets here
+#   - default: index 0
+#   - start with zero index
+#   - increase by 1, no gaps
+
+flags[0]="-DPARAM_OVERWRITES:LIST='-DPARAM_START_IONS=Quiet5000ppc;-DPARAM_START_ELECTRONS=Quiet5000ppc'"
+flags[1]="-DPARAM_OVERWRITES:LIST='-DPARAM_START_IONS=Quiet1000ppc;-DPARAM_START_ELECTRONS=Quiet5000ppc'"
+flags[2]="-DPARAM_OVERWRITES:LIST='-DPARAM_START_IONS=Quiet5000ppc;-DPARAM_START_ELECTRONS=Quiet1000ppc'"
+#flags[3]="-DPARAM_OVERWRITES:LIST='-DPARAM_ENABLE_PUSHER=1;-DPARAM_ENABLE_CURRENT=1'"
```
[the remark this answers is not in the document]
pordyna: no

share/picongpu/tests/CollisionsThermalisation/include/picongpu/param/collision.param
```diff
@@ -0,0 +1,55 @@
...
+
+namespace picongpu
+{
+    namespace particles
+    {
+        namespace collision
+        {
+            namespace precision
+            {
+                using float_COLL = float_64;
```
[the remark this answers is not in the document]
pordyna: I'm not sure what do you mean? This should be a compile time option.

share/picongpu/tests/CollisionsThermalisation/include/picongpu/param/grid.param
```diff
@@ -0,0 +1,106 @@
...
+#include "picongpu/param/density.param"
+
+namespace picongpu
+{
+    namespace SI
+    {
+        /** Duration of one timestep
+         *
+         *  unit: seconds */
+        constexpr float_64 DELTA_T_SI = (1.0 / REFERENCE_OMEGA_SI) * 0.2 * 2.0 * PI;
```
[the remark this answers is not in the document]
pordyna: https://github.com/SmileiPIC/Smilei/blob/master/benchmarks/collisions/thermalisation_ei1.py
I thought that way it's easier to compare. `smilei` has a totally different approach to units so I guess it looks a bit weird in picongpu.

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
sbastrakov: I think it's to avoid potential accuracy issues, not for better performance.

docs/source/models/binary_collisions.rst
```diff
@@ -0,0 +1,641 @@
+.. _model-binaryCollisions:
+
+Binary collisions
+=================
+
+.. sectionauthor:: Pawel Ordyna
+
+.. warning::
+    This is an experimental feature.
+    Fell free to use it but be aware that there is an unsolved bug causing some simulations to hang up.
```
sbastrakov, line 10: Typo: should be `feel`

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
```diff
@@ -0,0 +1,496 @@
...
+                            // [Perez 2012] (21) ( without n1*n2/n12 )
+                            float_COLL s12Max = math::pow(
+                                                    4.0_COLL * pmacc::math::Pi<float_COLL>::value / 3._COLL,
+                                                    1.0_COLL / 3.0_COLL)
+                                * DELTA_T_COLL * (mass0 + mass1)
+                                / pmacc::math::max(mass0 * densitySqCbrt0, mass1 * densitySqCbrt1)
+                                * relativeComsVelocity;
+                            s12Max *= s12Factor3;
+                            float_COLL s12;
+                            s12 = pmacc::math::min(s12n, s12Max);
```
sbastrakov, line 396: I guess this can be a single line?

include/picongpu/particles/collision/detail/CollisionContext.hpp
```diff
@@ -0,0 +1,53 @@
...
+                {
+                    T_Acc const* m_acc;
+                    mutable T_RngHandle* m_hRng;
+
+                    DINLINE CollisionContext(T_Acc const& acc, T_RngHandle& hRng) : m_acc(&acc), m_hRng(&hRng)
+                    {
+                    }
+                };
+
+                // Why can't we just use the constructor?
```
sbastrakov, line 43: I guess this comment is not final?

include/picongpu/particles/collision/detail/ListEntry.hpp
```diff
@@ -0,0 +1,292 @@
+/* Copyright 2014-2021 Rene Widera, Pawel Ordyna
```
sbastrakov, line 1: Is 2014 right? (it may be, i just don't know so wondering)

include/picongpu/particles/collision/detail/ListEntry.hpp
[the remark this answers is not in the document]
sbastrakov: I think I also asked about it a while ago offline, thinking it's a bad way to shuffle. But I think i was just wrong and this is fine.

include/picongpu/particles/collision/detail/ListEntry.hpp
```diff
@@ -0,0 +1,292 @@
...
+                    {
+                        using UniformUint32_t = pmacc::random::distributions::Uniform<uint32_t>;
+                        auto rng = rngHandle.template applyDistribution<UniformUint32_t>();
+                        // shuffle the particle lookup table
+                        for(uint32_t i = size; i > 1; --i)
+                        {
+                            /* modulo is not perfect but okish,
+                             * because of the loop head mod zero is not possible
+                             */
+                            int p = rng(acc) % i;
```
sbastrakov, line 133: Maybe `p` wants to be a fixed-width type, like `i` is

include/picongpu/particles/collision/detail/ListEntry.hpp
```diff
@@ -0,0 +1,292 @@
...
+                     *
+                     * @param v0 index of the 1st entry to swap
+                     * @param v0 index of the 2nd entry to swap
+                     */
+                    DINLINE void swap(uint32_t& v0, uint32_t& v1)
+                    {
+                        uint32_t tmp = v0;
+                        v0 = v1;
+                        v1 = tmp;
+                    }
```
sbastrakov, line 151: I think ideally this function should probably not be here, but as a some generic pmacc utility function like `std::swap<T>`. But it's okay as is now.

include/picongpu/unitless/collision.unitless
```diff
@@ -0,0 +1,47 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/unitless/precision.unitless"
+#include "picongpu/param/collision.param"
+namespace picongpu
```
sbastrakov, line 24: Probably an empty line would help readability

include/picongpu/particles/collision/WithPeer.hpp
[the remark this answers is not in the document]
PrometheusPi: Perfect! Thanks.

include/picongpu/particles/collision/detail/ListEntry.hpp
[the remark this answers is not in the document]
PrometheusPi: The first line helped me to at least guess the random method, but I was totally confused by the second line. @psychocoderHPC what did you want to say here?

share/picongpu/tests/CollisionsBeamRelaxation/include/picongpu/param/density.param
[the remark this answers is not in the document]
PrometheusPi: Okay - I understand that. @ComputationalRadiationPhysics/picongpu-maintainers @bussmann @ax3l what do you think, shall we reference another PIC code?

share/picongpu/tests/CollisionsBeamRelaxation/include/picongpu/param/memory.param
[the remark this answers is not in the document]
PrometheusPi: Thanks for the clarification, didn't see that.

share/picongpu/tests/CollisionsThermalisation/include/picongpu/param/collision.param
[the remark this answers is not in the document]
PrometheusPi: Yes, I get that. But why do you define it here and in param?

share/picongpu/tests/CollisionsThermalisation/include/picongpu/param/grid.param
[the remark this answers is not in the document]
PrometheusPi: Okay I get it. Could you perhaps make it:
```suggestion
        constexpr float_64 REFERENCE_TIME_SI = 2.0 * PI / REFERENCE_OMEGA_SI;
        constexpr float_64 DELTA_T_SI = 0.2 * REFERENCE_TIME_SI;
```
This would make clear how the numbers are related to each other without the need of extra comments.

share/picongpu/tests/CollisionsBeamRelaxation/include/picongpu/param/grid.param
```diff
@@ -0,0 +1,109 @@
...
+    {
+        /** Duration of one timestep
+         *
+         *  unit: seconds */
+        constexpr float_64 DELTA_T_SI = (1.0 / REFERENCE_OMEGA_SI) * PARAM_DELTA_T * 2.0 * PI;
+
+        /** equals X
+         *
+         *  unit: meter */
+        constexpr float_64 CELL_WIDTH_SI = (SPEED_OF_LIGHT_SI / REFERENCE_OMEGA_SI) * 2.0 * 2.0 * PI;
```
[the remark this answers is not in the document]
PrometheusPi: Okay, then I would propose the same as above:
```suggestion
        constexpr float_64 REFERENCE_LENGTH_SI = SPEED_OF_LIGHT_SI * REFERENCE_TIME_SI;
        constexpr float_64 CELL_WIDTH_SI = 2.0 * REFERENCE_LENGTH_SI;
```

share/picongpu/tests/CollisionsThermalisation/include/picongpu/param/memory.param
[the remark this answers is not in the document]
PrometheusPi: Thanks.

share/picongpu/tests/CollisionsThermalisation/include/picongpu/param/particle.param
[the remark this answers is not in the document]
PrometheusPi: Okay - thanks for the input. @n01r any comments on that?

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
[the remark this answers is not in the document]
PrometheusPi: Okay. Interesting aspect. Would be cool if you could investigate this further later. For this pull request, this can be omitted.

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
PrometheusPi: Could it be that you are using [this `Sqrt`](https://github.com/ComputationalRadiationPhysics/picongpu/blob/a2c46f7e5b5dd9e1103505fc3f4d00ce65f17386/include/pmacc/math/complex/Complex.tpp#L148-L167) not [this `Sqrt`](https://github.com/ComputationalRadiationPhysics/picongpu/blob/454a28efe6a6eb0fa1e0e97031ef2569a1ad313b/include/pmacc/algorithms/math/defines/sqrt.hpp#L39-L42) with [this implementation](https://github.com/ComputationalRadiationPhysics/picongpu/blob/454a28efe6a6eb0fa1e0e97031ef2569a1ad313b/include/pmacc/algorithms/math/doubleMath/sqrt.tpp#L37-L45). @psychocoderHPC and @sbastrakov why is the code here pointing to the complex square root?

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
PrometheusPi: okay

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
PrometheusPi: Okay - let's see what the others think (in above question regarding smilei).

include/picongpu/particles/collision/detail/ListEntry.hpp
pordyna: @psychocoderHPC should it be 2019?

share/picongpu/tests/CollisionsThermalisation/include/picongpu/param/particle.param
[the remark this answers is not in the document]
n01r: I don't have much of a rule for how many particles per density except that it should be at the very least one per Debye sphere. In general, there are also rules of thumb for how many particles in different dimensions one should take.
In 2D, particle weightings below 1 are okay due to the third dimension being of arbitrary height but I agree with @PrometheusPi that in 3D the weighting should be `>1`.

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
n01r: You could still ask them quickly in their user chat if you want to know where it comes from.
If you know it, it can be explained with a physics reason (preferred).
If not, you could reference it more like: ``// According to Smilei version <nr.>, <file>, <line>``

share/picongpu/tests/CollisionsThermalisation/include/picongpu/param/collision.param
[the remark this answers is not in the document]
pordyna: @PrometheusPi but that's collision.param? I'm confused.

include/picongpu/particles/collision/detail/ListEntry.hpp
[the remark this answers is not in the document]
pordyna: @PrometheusPi @sbastrakov I guess it only means that we don't divide by zero in `p=rng(acc) % i;` because of the `i>1` condition.

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
pordyna: @PrometheusPi @n01r 
This is just the case when COMS frame is the same as the lab frame. Normally `factorA = (gammaComs - 1.0_COLL) / comsVelocityAbs2`. The `comsVelocity --> 0` limit for `factorA` is 1/(2c^2) which is 0.5 for c=1. 
I changed line 348 to `factorA = 1.0_COLL/ (2.0_COLL * c * c);` and removed `smilei` from the comment.

include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
pordyna: @PrometheusPi  ccording to my IDE I'm using [this](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/thirdParty/cupla/include/cupla/device/math/Root.hpp). That looks  right.

pordyna:
I removed cmakeFlags from the file names because it looks like we still need another fix like #3555 but for TABs.

pordyna:
@psychocoderHPC It looks like the CI failed due to an CI-related error. Please, could you have a look at it? Maybe just restart the two failed jobs.

psychocoderHPC:
> @psychocoderHPC It looks like the CI failed due to an CI-related error. Please, could you have a look at it? Maybe just restart the two failed jobs.

I restarted the failing tests, thanks for the notification.

psychocoderHPC:
> @psychocoderHPC It looks like the CI failed due to an CI-related error. Please, could you have a look at it? Maybe just restart the two failed jobs.

I restarted the failing tests, thanks for the notification.

include/picongpu/particles/collision/detail/ListEntry.hpp
psychocoderHPC: I think I copy-pasted the header from somewhere, yes should be `2019 - 2021`

pordyna:
> > @psychocoderHPC It looks like the CI failed due to an CI-related error. Please, could you have a look at it? Maybe just restart the two failed jobs.
> 
> I restarted the failing tests, thanks for the notification.

@psychocoderHPC Unfortunately `g++-10_serial_boost1.72.0_examples.SingleParticleTest` failed again.

psychocoderHPC:
> > > @psychocoderHPC It looks like the CI failed due to an CI-related error. Please, could you have a look at it? Maybe just restart the two failed jobs.
> > 
> > 
> > I restarted the failing tests, thanks for the notification.
> 
> @psychocoderHPC Unfortunately `g++-10_serial_boost1.72.0_examples.SingleParticleTest` failed again.

np, restarted the job. Maybe before two jobs failed and I restarted only one. I think I missed the second.

pordyna:
@PrometheusPi @sbastrakov @psychocoderHPC I'm done with implementing your suggestions and the CI tests are ok now as well. So you can have a look at the newer commits whenever you've got the time. When everything is ok I will squash the commits and we can merge.

include/picongpu/particles/collision/detail/ListEntry.hpp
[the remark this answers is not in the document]
psychocoderHPC: `i > 1` is required because we access `ptrToIndicies[i - 1]`, without the guard it will be an out of memory access.
`rng(acc) % i` modulo should be not used if you like to keep the normal distribution of your random numbers. This should not be critical at this point.

include/picongpu/particles/collision/detail/ListEntry.hpp
psychocoderHPC: `std::swap` should not work on GPU it is not a constexpr function. This will be changed with C++20.

psychocoderHPC (approved):

psychocoderHPC:
@ComputationalRadiationPhysics/picongpu-maintainers Someone else should review this PR, I wrote parts of the code, therefore I am out^^

psychocoderHPC (changes requested):
@pordyna you removed again my co-authorship: https://github.com/ComputationalRadiationPhysics/picongpu/pull/3416#pullrequestreview-607542275

Please read me

psychocoderHPC (dismissed):
I am now a co-author, all fine^^

sbastrakov:
We need @PrometheusPi to approve his requested changes before merging anyways.

psychocoderHPC:
@pordyna With #3551 there are some tiny changes in the param files therefore the CI fails https://gitlab.com/hzdr/crp/picongpu/-/jobs/1112779559#L2055

should be a tiny change, I would say removing the include is all we need.

psychocoderHPC:
@pordyna There is still an issue in one file: https://gitlab.com/hzdr/crp/picongpu/-/jobs/1119909655#L2061

psychocoderHPC:
https://gitlab.com/hzdr/crp/picongpu/-/jobs/1121372429#L2056
The filed solver none has no template arguments https://github.com/ComputationalRadiationPhysics/picongpu/blob/16061e52bed810d6d61f1231865aaf9fe4e5e21c/include/picongpu/fields/MaxwellSolver/None/None.def#L31

psychocoderHPC:
@pordyna If the CI is passing could you please squash all except the first commit into one commit?

psychocoderHPC (approved):

psychocoderHPC:
@ComputationalRadiationPhysics/picongpu-maintainers This PR is ready to be merged but requires a review from someone else than me. I am a co-author of this PR.

sbastrakov (approved):

sbastrakov:
It also requires that @PrometheusPi lifts his "changes requested", or informs it's okay to dismiss it.

share/picongpu/tests/CollisionsBeamRelaxation/include/picongpu/param/density.param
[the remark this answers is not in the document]
psychocoderHPC: Why not, this test case based on the documentation in Smilei, therefore it schould be referenced.

psychocoderHPC (approved):

wmles:
Sorry to be so late (it's merged already) but I noticed some kind of a bug
It works fine if I don't touch the precision of float_64, but if I set it to _32 or _X the compiler breaks after half a minute with:
`error: no suitable user-defined conversion from "pmacc::math::Vector<picongpu::precision32Bit::float_X, 3, pmacc::math::StandardAccessor, pmacc::math::StandardNavigator, pmacc::math::detail::Vector_components>" to "const picongpu::particles::collision::precision::float3_COLL" exists`
[compile.log](https://github.com/ComputationalRadiationPhysics/picongpu/files/6295444/compile.log)

psychocoderHPC:
> Sorry to be so late (it's merged already) but I noticed some kind of a bug
> It works fine if I don't touch the precision of float_64, but if I set it to _32 or _X the compiler breaks after half a minute with:
> `error: no suitable user-defined conversion from "pmacc::math::Vector<picongpu::precision32Bit::float_X, 3, pmacc::math::StandardAccessor, pmacc::math::StandardNavigator, pmacc::math::detail::Vector_components>" to "const picongpu::particles::collision::precision::float3_COLL" exists`
> [compile.log](https://github.com/ComputationalRadiationPhysics/picongpu/files/6295444/compile.log)

Could you please open an issue else we can not track this problem.

psychocoderHPC:
> Sorry to be so late (it's merged already) but I noticed some kind of a bug
> It works fine if I don't touch the precision of float_64, but if I set it to _32 or _X the compiler breaks after half a minute with:
> `error: no suitable user-defined conversion from "pmacc::math::Vector<picongpu::precision32Bit::float_X, 3, pmacc::math::StandardAccessor, pmacc::math::StandardNavigator, pmacc::math::detail::Vector_components>" to "const picongpu::particles::collision::precision::float3_COLL" exists`
> [compile.log](https://github.com/ComputationalRadiationPhysics/picongpu/files/6295444/compile.log)

Please also describe what you changed. `It works fine if I don't touch the precision of float_64, but if ...` for me it is not clear where you changed it. Do you changed `presision.param?`

wmles:
Sorry! I thought it is clear for Pawel - but yes, I'll explain it :)

PrometheusPi:
@psychocoderHPC Sorry for not lifting my review earlier. Thanks for removing it.

PR #3417, opened 2020-11-05
Add PML support to arbitrary-order FDTD field solver

steindev:
Finally, the addition of PML support to the AOFDTD solver. 
In short, the `ArbitraryOrderFDTDPML` solver performs as can be expected from the performance of the standard `YeePML` solver.

# Test case
I tested with the following case described in Taflove, *computational electrodynamics*, 2005, sec. 7.11.1: Current Source Radiating in an Unbound Two-Dimensional Region.
A `Jy` current source of one cell width is located in the center of a 64x64 volume surrounded by a 10 cell thick layer of (C)PML (the standard PML configuration in PIConGPU).
The source radiates a short pulse of a transverse electric wave polarized along z.
The electric `Ey` field is recorded at two positions `A` and `B` in the volume, see picture below, for 1000 time steps.

The same simulation is performed in 1120x1120 volume (with periodic boundaries) and provides a reference `Ey_ref` for the field in the small volume of the simulation with PMLs.

In order to estimate the performance of the PMLs, a relative field error is estimated by comparing the `Ey` field in the small simulation with PMLs to the `Ey_ref` field in the reference simulation at the positions `A` and `B`, respectively,
```
  rel. error = abs( Ey - Ey_ref ) / max_t( abs(Ey_ref) )
```
where `max_t( abs(Ey_ref) )` is the maximum of the absolute value of the `Ey_ref` field over time at the positions `A` and `B`, respectively.

In order to verify the implementation of the PMLs in the `ArbitraryOrderFDTDPML` solver, the comparison described above is carried out for both, the `ArbitraryOrderFDTDPML` with 4 neighbors and `YeePML` solver.

In conclusion, the `ArbitraryOrderFDTDPML` performs as good as  `YeePML`.
The lower absolute value in the recorded `Ey` field is due to the small source thickness.
As the current source's thickness of one cell is smaller than the width of the arbitrary-order solver's stencil, source terms go into the field calculation with a lower weighting in the arbitrary-order solver than in the `YeePML` solver. 
The latter uses only one neighbor with weighting 1 while the weighting in the arbitrary-order solver depends on the number of neighbors but is always smaller than one for more than one neighbor.

image

I also tested the solver in 3D by extending the domain in z over 64 cells with periodic boundary conditions at the boundary planes normal to the z axis.
The results are the same.


sbastrakov:
Not sure, but perhaps the labels for relative errors for points A and B are should be the opposite? Point B should always have a larger error in this test.

share/picongpu/examples/FieldAbsorberTest/cmakeFlags
```diff
@@ -38,58 +38,60 @@ flags[3]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=ArbitraryOrderFDTD< 4 MY_
 # Test that PML compiles in 3D and 2D
 flags[4]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=YeePML< CurrentInterpolation >;-DPARAM_PML_SIZE=8;-DPARAM_PML_KAPPA_MAX=1.0;-DPARAM_PML_ALPHA_MAX=0.2'"
 flags[5]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=YeePML< CurrentInterpolation >;-DPARAM_PML_SIZE=8;-DPARAM_PML_KAPPA_MAX=1.0;-DPARAM_PML_ALPHA_MAX=0.2;-DPARAM_DIMENSION=DIM2'"
-
+# Test that arbitrary-order solver with PML absorbing boundary conditions compiles in 3D and 2D
+flags[6]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=ArbitraryOrderFDTDPML< 4 MY_AOFDTD_COMMA CurrentInterpolation >;-DPARAM_PML_SIZE=8;-DPARAM_PML_KAPPA_MAX=1.0;-DPARAM_PML_ALPHA_MAX=0.2'"
+flags[7]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=ArbitraryOrderFDTDPML< 4 MY_AOFDTD_COMMA CurrentInterpolation >;-DPARAM_PML_SIZE=8;-DPARAM_PML_KAPPA_MAX=1.0;-DPARAM_PML_ALPHA_MAX=0.2;-DPARAM_DIMENSION=DIM2'"
```
sbastrakov, line 44: I am not sure if a custom macro `MY_AOFDTD_COMMA` for comma is good. What do you think @psychocoderHPC ?

sbastrakov:
@steindev CI complains about white space at end of line.

steindev:
> @steindev CI complains about white space at end of line.

Yes, thanks for the note. I will wait for more reviewer comments so I can fix these in one run.

share/picongpu/examples/FieldAbsorberTest/cmakeFlags
psychocoderHPC: There is also a boost PP comma available.

PrometheusPi (approved):
looks good to me - there is still an ongoing discussion regarding comma separation

share/picongpu/examples/FieldAbsorberTest/cmakeFlags
steindev: Refactored to use the boost comma.

steindev:
Addressed reviewer comments and fixed white space.

psychocoderHPC (approved):

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/fieldSolver.param
```diff
@@ -37,6 +37,8 @@
 #include "picongpu/fields/MaxwellSolver/Solvers.def"
 #include "picongpu/fields/currentInterpolation/CurrentInterpolation.def"
 
+#include <boost/preprocessor/punctuation/comma.hpp>
+
```
sbastrakov, line 41: Sorry I didn't realize this before, when this solution was suggested. This adds a dependency on Boost.propocessor library, that we do not currently have. I do not think it's worth it honestly. @psychocoderHPC what are your thoughts? If we decide to add it, should update cmakelists.txt and documentation as well. I think a more practical solution would be to just make another macro that decides on Yee / AOFDTD and express a combination via 2 separate macros.

sbastrakov (changes requested):
There is one thing that concerns me, pointed at the code line.

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/fieldSolver.param
psychocoderHPC: PMacc already depends on boost PP, I am fine with both solutions.

PrometheusPi (approved):

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/fieldSolver.param
PrometheusPi: @sbastrakov How about you? Do you still require changes?

sbastrakov: Ah, of course. I only looked at PIConGPU's cmakelists which does not. Okay, then I'm lifting my concern.

sbastrakov (approved):

Issue #3390, opened 2020-10-15
Should we rename namespace picongpu::openPMD  ?


sbastrakov:
It causes to use openPDD API as `::openPMD::` and for those who are unaware this may cause trouble e.g. when copying from openPMD API examples

sbastrakov:
To clarify it is also (in my opinion) difficult to see at the code that uses both `::openPMD::` and `openPMD`, often in the same expression.

So I would prefer if `picongpu::openPMD` would become smth like `picongpu::openPMDHelper`.

ax3l:
Potentially.
Also I recommend [doing this](https://openpmd-api.readthedocs.io/en/0.12.0-alpha/usage/firstwrite.html):
```C++
namespace io = openPMD;
```
which makes it unique, too ;D

sbastrakov:
I think while there are other IO libraries in use that would be a weird alias to use

sbastrakov:
I mean, weird ourside of the openPMD "land", as one reading PIConGPU code would need to know io means oprnPMD. And having at the same time picongpu::openPMD does not become less confusing then imo

psychocoderHPC:
I for changing `picongpu::openPMD` to `picongpu::openPMDHelper`. 
`namespace io = openPMD;` is also possible but I think we do not have so many openPMD interactions in the code so being a little bit more verbose and writeing `openPMD::*` is better.

franzpoeschel:
Axel's suggestion would mean to rename the original `openPMD` namespace (i.e. the one by the openPMD API) as `io`? I agree that this might be confusing, I would leave that namespace untouched. What about renaming `picongpu::openPMD` as `picongpu::io` instead? That would make sense imo since the openPMD plugin is in fact PIConGPU's IO plugin.
I don't really like `openPMDHelper`, things like `*helper` or `*detail` always sound too much like it's some little namespace that you're not supposed to be touching, while in reality it's the complete plugin…

ax3l:
> Axel's suggestion would mean to rename the original openPMD namespace (i.e. the one by the openPMD API) as io?

No, this is just an example of how `import numpy as np` looks in C++ inside *user* code.

I agree with René's suggestion. I just mentioned this because namespace aliases are pretty and too seldomly used :)

franzpoeschel:
[...]

PR #3419, opened 2020-11-05
Remove unused ParticlePatches class

sbastrakov:
This class is never used. The openPMD plugin uses patches from openPMD API.

(Unless I've missed something, but then the CI should catch it).


steindev:
So, I guess, you checked that these files are nowhere included?

sbastrakov:
Well, it's a little bit more complicated in this case. As the `.cpp` file was automatically part of the `picongpu-hostonly` library. But it seems to me that except compiling them, there is no actual use.

steindev:
But I see it is used in struct WriteSpecies::operator(), line 531 of `include/picongpu/plugins/openpmd/WriteSpecies.hpp`. What about this?

sbastrakov:
No, that one is from openPMD API. That is confusing, as pointed out in issue #3390

steindev:
Then it is more a question whether this should be kept as a backup? I don't think so, as we have a versioning. And it will probably outdate at some point since we are not using it anyways.

Any objections @franzpoeschel, @psychocoderHPC?

PrometheusPi (approved):

psychocoderHPC:
> Then it is more a question whether this should be kept as a backup? I don't think so, as we have a versioning. And it will probably outdate at some point since we are not using it anyways.
> 
> Any objections @franzpoeschel, @psychocoderHPC?

Maybe it is a leftover of our HDF5 pluign. But as you wrote we have versioning, so there is no need to keep this unused code.

psychocoderHPC (approved):

PR #3427, opened 2020-11-11
Use openPMD::getVersion() function

franzpoeschel:
The `openPMD API` has a `std::string getVersion()` since `0.12.0`, so we don't need to construct the version string manually any longer.

### DIFF
--- a/include/picongpu/versionFormat.cpp
+++ b/include/picongpu/versionFormat.cpp
@@ -142,21 +142,10 @@ namespace picongpu
         adios << versionNotFound;
 #endif
 
-        std::stringstream openPMD;
 #if( ENABLE_OPENPMD == 1 )
-        openPMD << OPENPMDAPI_VERSION_MAJOR
-                << "."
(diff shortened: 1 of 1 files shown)


sbastrakov (approved):

PrometheusPi (commented):
looks good to me - still the compile suite throws an error

PrometheusPi (approved):
sorry - wanted to press "approve"

PrometheusPi (changes requested):
The old openPMD stringstream interface is still used in https://github.com/franzpoeschel/picongpu/blob/e43f9fa9fd90830ffff62bfbfdfd0e95f2eb112b/include/picongpu/versionFormat.cpp#L196

franzpoeschel:
Argh, I built the wrong branch in my own test. Lemme fix..

PrometheusPi (approved):
@franzpoeschel Thanks for quickly fixing it.

psychocoderHPC (approved):

PR #3430, opened 2020-11-13
Cleanse field solver compile test from species dependency

steindev:
Remove `species.param`, `speciesDefinition.param`, and `speciesInitialization.param` as these are not necessary for the compile test.
Furthermore, the current interpolation test is moved to the current solver compile test as the field solver test should be independent of the current calculation.
The `fieldSolver.param` is also updated in the field solver compile tests and it is added to the current solver compile test allowing to include the current interpolation compile test.


PrometheusPi (approved):

share/picongpu/tests/compileCurrentSolver/include/picongpu/param/fieldSolver.param
```diff
@@ -0,0 +1,86 @@
...
+     *   - None:
+     *     - default for staggered grids/Yee-scheme
+     *     - updates E
+     *   - Binomial: 2nd order Binomial filter
+     *     - smooths the current before assignment in staggered grid
+     *     - updates E & breaks local charge conservation slightly
+     */
+#ifndef PARAM_CURRENTINTERPOLATION
+#   define PARAM_CURRENTINTERPOLATION None
+#endif
```
sbastrakov, line 62: Can't this be removed now?

psychocoderHPC (approved):

sbastrakov (approved):

PR #3440, opened 2020-11-23
introduce clang format

psychocoderHPC:
- clang format file based on https://github.com/ComputationalRadiationPhysics/contributing
  - remove clang format 12 formating commands
  - add include folder ordering based on the usage of PIConGPU/PMac
- update documentation
- format all PIConGPU/PMacc files

# Review
Please check this PR commit by commit. 
There is a commit that is coping the clang format file from our contribution repository.
A second commit is currently disabling that clang format changes the include order. 
This will break the code and will be activated in a separate PR later.
The last commit is updating the documentation.
Do *NOT* review the formating! The formating commit is committed as `Tools` user.

PR #3445, opened 2020-11-24
fix documentation for `git clang-format`

psychocoderHPC:
Clang format was not interpreting all file extensions used as C++ files.
This PR updates the clang format section and add the required commands
to set file extension for `git clang-format`

@BeyondEspresso This solves the issue you had with the code formatting.

### DIFF
--- a/docs/COMMIT.md
+++ b/docs/COMMIT.md
@@ -11,8 +11,8 @@ We agree on the following simple rules to make our lives easier :)
 Format Code
 -----------
 
-- Install clang-format-11
-- To format all file in your working copy, you can run this command in bash from the root folder of PIConGPU:
+- Install *ClangFormat 11*
+- To format all files in your working copy, you can run this command in bash from the root folder of PIConGPU:
   ```bash
   find include/ share/picongpu/ share/pmacc -iname "*.def" \
   -o -iname "*.h" -o -iname "*.cpp" -o -iname "*.cu" \
@@ -21,8 +21,15 @@ Format Code
   | xargs clang-format-11 -i
   ```
  
-To format only lines you added with `git add` you can call `git clang-format-11` before you create a commit.
-Take care that un-staged changes will not be formatted.
+Instead of using the bash command above you can use *Git* together with *ClangFormat* to format your patched code only. 
+Before applying this command, you must extend your local git configuration **once** with all file endings used in *PIConGPU*:
+
+```
+git config --local clangFormat.extensions def,h,cpp,cu,hpp,tpp,kernel,loader,param,unitless
+```
+
+For only formatting lines you added using `git add`, call `git clang-format-11` before you create a commit.
+Please be aware that un-staged changes will not be formatted.
 
 Commit Messages
 ---------------
--- a/docs/source/dev/styleguide.rst
+++ b/docs/source/dev/styleguide.rst
@@ -31,15 +31,24 @@ To format all files in your working copy, you can run this command in bash from 
      -o -iname "*.loader" -o -iname "*.param" -o -iname "*.unitless" \
      | xargs clang-format-11 -i
 
-Format Git Staged Changes
-^^^^^^^^^^^^^^^^^^^^^^^^^
+Format Changes Using Git
+^^^^^^^^^^^^^^^^^^^^^^^^
 
-To format all changed added with `git add` you can run `git clang-format-11`.
+Instead of using the bash command above you can use *Git* together with *ClangFormat* to format your patched code only.
+Before applying this command, you must extend your local git configuration **once** with all file endings used in *PIConGPU*:
 
-Format Commits
-^^^^^^^^^^^^^^
+.. code-block:: bash
 
-To format the last commit you can run `git clang-format-11 HEAD~1`.
+   git config --local clangFormat.extensions def,h,cpp,cu,hpp,tpp,kernel,loader,param,unitless
+
+.. warning::
+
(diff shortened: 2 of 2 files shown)


docs/source/dev/styleguide.rst
```diff
@@ -30,15 +30,19 @@ find include/ share/picongpu/ share/pmacc -iname "*.def" \
 | xargs clang-format-11 -i
 ```
 
-Format Git Staged Changes
-^^^^^^^^^^^^^^^^^^^^^^^^^
+Format Changes With Git
```
BeyondEspresso, line 33: With --> Using

docs/source/dev/styleguide.rst
```diff
@@ -30,15 +30,19 @@ find include/ share/picongpu/ share/pmacc -iname "*.def" \
...
+^^^^^^^^^^^^^^^^^^^^^^^
 
-To format all changed added with `git add` you can run `git clang-format-11`.
+Instead of using the bash command above you can use *Git* together with *clang-format* to format your patched code, only.
+You must extent your local git configuration **once** with all file endings used in PIConGPU:
```
BeyondEspresso, line 37: Instead of using the bash command above you can use *git* together with clang-format to format your patched code only. Before applying this command, you must extend your local git configuration **once** with all file endings used in PIConGPU:

docs/source/dev/styleguide.rst
```diff
@@ -30,15 +30,19 @@ find include/ share/picongpu/ share/pmacc -iname "*.def" \
...
-Format Commits
-^^^^^^^^^^^^^^
+.. code-block:: bash
 
-To format the last commit you can run `git clang-format-11 HEAD~1`.
+   git config --local clangFormat.extensions def,h,cpp,cu,hpp,tpp,kernel,loader,param,unitless
+
+
+To format all changed added with `git add` you can execute `git clang-format-11`.
+The last commit can be formatted by executing `git clang-format-11 HEAD~1`.
```
BeyondEspresso, line 45: Formatting all changes of the previous commit can be achieved by executing the command `git clang-format-11 HEAD~1`.

BeyondEspresso (changes requested):

psychocoderHPC:
PR is updated by suggestions from @BeyondEspresso and https://github.com/ComputationalRadiationPhysics/picongpu/pull/3440#discussion_r529864361

docs/COMMIT.md
```diff
@@ -11,7 +11,7 @@ We agree on the following simple rules to make our lives easier :)
 Format Code
 -----------
 
-- Install clang-format-11
+- Install *ClangFormat 11*
 - To format all file in your working copy, you can run this command in bash from the root folder of PIConGPU:
```
sbastrakov, line 15: Should be `files`, not `file`

docs/COMMIT.md
```diff
@@ -21,8 +21,15 @@ Format Code
...
-To format only lines you added with `git add` you can call `git clang-format-11` before you create a commit.
-Take care that un-staged changes will not be formatted.
+Instead of using the bash command above you can use *Git* together with *ClangFormat* to format your patched code only. 
+Before applying this command, you must extend your local git configuration **once** with all file endings used in *PIConGPU*:
+
+```
+git config --local clangFormat.extensions def,h,cpp,cu,hpp,tpp,kernel,loader,param,unitless
+```
+
+For only formatting lines you added using `git add`, call `git clang-format-11` before you create a commit.
```
sbastrakov, line 31: I think this is not correct. Perhaps should be `git clang-format` ? At least only this way works on the dev system.

sbastrakov (changes requested):

docs/COMMIT.md
BeyondEspresso: I kindly disagree, `git clang-format-11` worked for me on my own system.

sbastrakov: Okay, so then it's system-specific.

psychocoderHPC: git clang-format-11 is using definitiv clangFormat version11. if you only use 'git clang-format' it can be any version depending on your system or what you defined in the git config.

sbastrakov: Yet on our dev system there is clang format 11 installed, and only `clang-format` works (and is version 11), but not `clang-format-11`. I am not sure, maybe this is some issue on the system.

docs/COMMIT.md
psychocoderHPC: fixed

docs/COMMIT.md
psychocoderHPC: It depends on how the module is provided. On our dev system, Clang11 is compiled via spack. I would like to keep the name `clang-format-11` to be clear which version is used. If an OS does not have `clang-format-11` and `clang-format` is available I think the developer will manage it. 
I do not like to describe edge cases for each OS/distribution.

psychocoderHPC: If you like I can add a warning box in the `rst` file, I would not add it to the `md` file.

```
The binary for `ClangFormat` is on some operating system called `clang-format`. If so please check that `clang-format --version` return version `11.X.X`.
```

sbastrakov: Okay, that seems good to me

psychocoderHPC: I added the warning to the RST file

sbastrakov (approved):

psychocoderHPC:
ready to be merged

Issue #3421, opened 2020-11-06
We need to add a check for Debye length resolution


sbastrakov:
Triggered by #3377.

For now I'm thinking of having it around the same place as we check resolution of the plasma frequency. So only at start of simulation and output it in the same fashion, so that it's a diagnostic warning, not a hard error.

To implement it, I plan to estimate the standard deviations of particle momentums per cell or supercell, and get an estimate of temperature out of it.

@psychocoderHPC suggested we could also consider doing it for the "new" area when a moving window slides.

cbontoiu:
Hello,
Another check that should be there (maybe it is but I am not aware) is a clear message to the user that there is no laser beam in use. That is there might be a laser defined but there is no ```using Selected = ... < ... > ;``` at the end of the ```laser.param``` file. I spent a few days wondering why none of my laser modifications gave any impact onto the target when in fact no laser was used at all. Of course PIConGPU can run models without laser beams, but I think there should be a warning when the simulation starts, because missing ```using Selected = ... < ... > ;``` doesn't end in a compilation error.
Thank you.

sbastrakov:
Sorry @cbontoiu I do not fully understand the issue with no `Selected`. Could you post an example of a `.param` file that illustrated it?

cbontoiu:
> 
> 
> Sorry @cbontoiu I do not fully understand the issue with no `Selected`. Could you post an example of a `.param` file that illustrated it?

The zip folder attached to my issue #3412, now closed contains a ```laser.param``` file with this situation. You can compare that file with other laser files and note that the last statement ```using Selected = ... < ... > ;```, which means my laser is defined but not used.

sbastrakov:
I can see it, however having `using Selected` is a compile-time requirement. And a setup with `laser.param` not having it does not compile. In case a setup does not define its own `laser.param` (does not have it in `include/picongpu/param` of the setup directory), a default version of this file will be used. Which has no laser by setting `using Selected = None<>;`. However, this type name must be defined, otherwise there is no way the code compiles.

steindev:
Closed with PR #3446

PR #3446, opened 2020-11-25
Add a check for Debye length resolution

sbastrakov:
The check is performed for the only electon-like species (if detected) at start of simulation. In case the check fails, there is a warning printed, no hard error.

Because of PIConGPU intricacies, I came to conclusion that it's not possible to simplistically check applications of `AddTemperature`, and instead the numerical temperature is evaluated based on initial momentums of electrons. The assumption is that there is a constant or relatively slowly varying component of momentums and a random temperature-caused part. The check evaluates the standard deviation inside a supercell to characterize that temperature part and computes the corresponding Debye length. In case the assumption does not hold, the check is meaningless I guess. 

With a moving window enabled, the check is done for all processes, including "out-of-window" ones. Currently it is not done when a window is moving, although I assume it is possible to also call it then with just a little development work. I did check that the internal estimate for temperature is quite accurate to what we set in `AddTemperature`.

Solves #3421.

TODO:

- [x] check that there is no mix of real and macro-particles in density calculation
- [x] change hard-coded "e" to a more sophisticated procedure, discussed offline
- [x] add docs on Debye length resolution and also explain how this check should be interpreted there
- [x] add a threshold on number of macroelectron per cell to do the check, use 10 there
- [x] add output of this threshold so that a user understands what's done
- [x] extend output with % of valid supercells that have < 1 cell per Debye length

Consider for further development, not in this PR:

- addintg confidence intervals and expected number of cells to fail the check due to estimate of temperature
- output of "heat map" of estimated Debye lengths, or a python tool to make it out of a checkpoint
- for this and other runtime checks (like plasma wavelength; not like CFL) consider having a command-line option to make them hard errors


include/picongpu/particles/debyeLength/Check.hpp
```diff
@@ -0,0 +1,180 @@
...
+             * This function must be called from all MPI ranks.
+             *
+             * @param cellDescription mapping for kernels
+             */
+            HINLINE void check(MappingDesc const cellDescription)
+            {
+                bool isPrinting = (Environment<simDim>::get().GridController().getGlobalRank() == 0);
+                using ElectronSpecies = pmacc::particles::meta::FindByNameOrType_t<
+                    VectorAllSpecies,
+                    PMACC_CSTRING("e"),
```
psychocoderHPC, line 150: We need to implement another way to find all-electron species because there is no guarantee that this is the name for electrons.
You suggested already to check the charge of each species. The fix should be one helper function only, take care you can have more than one electron species.

psychocoderHPC: A possible way is to have an entry function for the check and that you call this entry function for each species.
Add a guard that only electrons can pass.

```C++
template<typename T_Species)
void check(T_Species sp)
{
   if(chargeOf<T_Species>!= ELECTRON_CHARGE)
       return;
   //...
}
```

include/picongpu/particles/debyeLength/Check.kernel
```diff
@@ -0,0 +1,227 @@
...
+                        uint32_t const workerIdx = cupla::threadIdx(acc).x;
+                        while(frame.isValid())
+                        {
+                            // parallel loop over all particles in the frame
+                            ForEachIdx<ParticleDomCfg>{workerIdx}([&](uint32_t const linearIdx, uint32_t const) {
+                                if(linearIdx < particlesInFrame)
+                                {
+                                    auto const particle = frame[linearIdx];
+                                    // the evaluation is based on single- (not macro-) particle momentums
+                                    auto const singleParticleMomentum = precisionCast<float_64>(particle[momentum_])
```
psychocoderHPC, line 123: Why do you cast first to 64 bit instead of deciding first and then cast to 64bit?

include/picongpu/particles/debyeLength/Check.hpp
```diff
@@ -0,0 +1,180 @@
...
+                    auto status = detail::checkDebyeLength<ElectronSpecies>(cellDescription);
+                    if(isPrinting)
+                        switch(status)
+                        {
+                        case detail::Status::Passed:
+                            log<picLog::PHYSICS>("Estimated Debye length for species \"e\" is resolved by the grid");
+                            break;
+                        case detail::Status::Failed:
+                            log<picLog::PHYSICS>(
+                                "Warning: estimated Debye length for species \"e\" is not resolved by the grid\n"
```
psychocoderHPC, line 164: Could we please output the calculated temperature and the required? In case the user is getting a `warning` he can see how much he/she is off.

include/picongpu/particles/debyeLength/Check.hpp
```diff
@@ -0,0 +1,180 @@
...
+                    pmacc::errorHandlerPolicies::ReturnType<void>>;
+                bool isElectronsFound = !std::is_same<ElectronSpecies, void>::value;
+                if(isElectronsFound)
+                {
+                    auto status = detail::checkDebyeLength<ElectronSpecies>(cellDescription);
+                    if(isPrinting)
+                        switch(status)
+                        {
+                        case detail::Status::Passed:
+                            log<picLog::PHYSICS>("Estimated Debye length for species \"e\" is resolved by the grid");
```
psychocoderHPC, line 160: Could we please always output the calculated values and the required? IN case the user is using a too high temperature he is able to see it by this output.

psychocoderHPC (changes requested):

include/picongpu/particles/debyeLength/Check.hpp
sbastrakov: The temperature does not have to be constant. So it's not clear which value to output.

sbastrakov: The goal of this check how I see it is to say to a user "perhaps you forgot to set the temperature or set it incorrectly". I don't see an easy way to point directly at what's wrong. We can (and should regardless of this PR) however extend the documentation on which things should be resolved.

include/picongpu/particles/debyeLength/Check.hpp
sbastrakov: I do not see how to output it clearly. Please see my answer to your previous question.

include/picongpu/particles/debyeLength/Check.kernel
sbastrakov: So that all operations are in 64 bit. Although in this case it should not really matter.

include/picongpu/particles/debyeLength/Check.hpp
sbastrakov: Yes, I thought about it. But then I realized that if there are multiple electron species, there is no easy way to do this check at all. So I decided to only implement the understandable and common case of one electron species.

psychocoderHPC: Simply call this check for all species and do a return if the species is not an electron species.

psychocoderHPC: You can call a functor for each species with `meta::ForEach` e.g. https://github.com/ComputationalRadiationPhysics/picongpu/blob/84e03980f2a56c7aea24d88bc3be9eb43f1a3197/include/picongpu/plugins/openPMD/openPMDWriter.hpp#L266

psychocoderHPC:
I stopped the CI because some tests already failed: https://gitlab.com/hzdr/crp/picongpu/-/jobs/872094506#L953

include/picongpu/particles/debyeLength/Check.hpp
psychocoderHPC: offline discussed: we will check how many electrons like species with a current solver are used. In case only one electron species is used the test will be performed, else there will be a warning that this test is not executed.

include/picongpu/particles/debyeLength/Check.kernel
```diff
@@ -0,0 +1,227 @@
...
+                        using Frame = typename T_ElectronBox::FrameType;
+                        auto const mass = static_cast<double>(picongpu::traits::frame::getMass<Frame>());
+                        auto const kT = momentumVariance / mass;
+                        auto const electronCharge = picongpu::traits::frame::getCharge<Frame>();
+                        auto const debyeLength = sqrt(EPS0 * kT / (electronDensity * electronCharge * electronCharge));
+
+                        auto maxCellSize = cellSize[0];
+                        for(uint32_t d = 1; d < simDim; d++)
+                            maxCellSize = math::max(maxCellSize, cellSize[d]);
+                        auto const minCellsPerDebyeLength = 2.0;
```
steindev, line 210: By the way, Hockney says `minCellsPerDebyeLength = 1.0`, while Birdsall provides results of numerical experiments for `minCellsPerDebyeLength = 3.0` :man_shrugging:

include/picongpu/particles/debyeLength/Check.hpp
```diff
@@ -0,0 +1,180 @@
...
+                            return Status::Failed;
+                    }
+                    else
+                        return Status::Skipped;
+                }
+
+            } // namespace detail
+
+            /** Check Debye length resolution
+             *
```
steindev, line 136: Please add a line or two describing what the actual check/limit is. With a reference, please (e.g. Hockney, p. 32; Birdsall p. 389).

steindev (changes requested):
:clap: Very useful addition. Like it. Nevertheless, we should change the calculation of temperature from the momenta according to the comment and add a describing comment to the function.

sbastrakov:
Thanks @steindev for your comments, I really like your suggestions! There is one thing I am not sure about, probably easier to discuss offline.

sbastrakov:
I pushed improvements to how the check is done, and results are output. The code is basically finished, according to my current understanding. There are just some explanatory comments missing. And I plan to add some info readthedocs as part of this PR.

Now the output looks like
```
PIConGPUVerbose PHYSICS(1) | Resolving Debye length for species "e"?
PIConGPUVerbose PHYSICS(1) | Estimated temperature 24.9561 KeV and 2.42166 cells per Debye length
   Estimates are based on variance in initial momentums and assume a single electron species
```
That's in case "e" is the only species with electron-like mass and charge, and with current deposition. If there are 0 or >= 2 of such species, the check is skipped with the corresponding message. This is done because the check itself uses a simple single-electron-species way to compute Debye length.

psychocoderHPC:
What is the current state of this PR?

sbastrakov:
@psychocoderHPC I still haven't written the readthedocs.

PrometheusPi:
What is the current state of this PR?

sbastrakov:
Thanks for pinging, it's laying in the almost-finished state for way too long. I will finish it and mark ready for review by Wednesday.

sbastrakov:
I've completed the current version which is ready for review.

Now the output looks as follows, in this example for 25 KeV initial temperature
`Estimated temperature 24.9561 KeV and Debye length 4.29117e-07 m. The grid has 2.42166 cells per Debye length`
So as suggested by reviewers, it merely outputs the relevant numbers to the user, and does not judge if it's right or wrong. The number of cells per Debye length is computed using the largest of dx, dy, dz, as generally all 3 have to be resolved.

Regarding the documentation, after looking I didn't find a good place to put it. I feel we generally need a whole section explaining this and other things to resolve, probably I'm not the best person to write such a section. I have added reminders about Debye length resolution to `grid.param` (as it also talks about other things to resolve) and doxygen of `Temperature` manipulator, both displayed in the readthedocs.

include/picongpu/particles/debyeLength/Estimate.kernel
```diff
@@ -0,0 +1,228 @@
...
+                    pmacc::DataSpace<simDim> const supercellIdx,
+                    T_EstimateBox estimateBox) const
+                {
+                    using namespace pmacc::mappings::threads;
+                    constexpr uint32_t frameSize = pmacc::math::CT::volume<SuperCellSize>::type::value;
+                    constexpr uint32_t numWorkers = T_numWorkers;
+                    using ParticleDomCfg = IdxConfig<frameSize, numWorkers>;
+                    /* Estimate variance in momentum of macroparticles in the supercell.
+                     * Each thread processes its particles, then the master thread performs reduction.
+                     * Sums are weighted and operate with single- (not macro-) particle momentums.
```
PrometheusPi, line 110:
```suggestion
                     * Sums are weighted and operate with single- (not macro-) particle momenta.
```
I like momentums much better since it indicated the plural clearly, but sadly its momenta. 😢

include/picongpu/particles/debyeLength/Estimate.kernel
```diff
@@ -0,0 +1,228 @@
...
+
+                    /* For the Maxwell-Boltzmann distribution,
+                     * variance = mass * k * T, both for physical and macroparticle.
+                     * We assume this distribution and estimate k * T out of variance.
+                     * Applying the ideal gas law would result in the same estimate.
+                     */
+                    using Frame = typename T_ElectronBox::FrameType;
+                    auto const mass = static_cast<double>(picongpu::traits::frame::getMass<Frame>());
+                    auto const kT = (momentumVariance[0] + momentumVariance[1] + momentumVariance[2]) / (3.0 * mass);
+                    auto const temperatureKeV = kT * UNIT_ENERGY / UNITCONV_keV_to_Joule;
```
PrometheusPi, line 208: Why do you perform the SI conversion on the device?

include/picongpu/particles/debyeLength/Estimate.kernel
```diff
@@ -0,0 +1,228 @@
...
+                    cupla::atomicAdd(acc, &(estimateBox(0).sumWeighting), sumWeighting, ::alpaka::hierarchy::Blocks{});
+                    cupla::atomicAdd(
+                        acc,
+                        &(estimateBox(0).sumTemperatureKeV),
+                        sumWeighting * temperatureKeV,
+                        ::alpaka::hierarchy::Blocks{});
+                    cupla::atomicAdd(
+                        acc,
+                        &(estimateBox(0).sumDebyeLength),
+                        sumWeighting * debyeLength,
```
PrometheusPi, line 221: You are computing the average Debye length, which is great.
But assuming we have a inhomogeneous plasma, would computing the minimal Debye length not reveal issues with resolutions much clearer, or are there numerical issues for the per-supercell evaluation?

PrometheusPi (changes requested):
Great job, just some typos, some minor question and a general question regarding mean vs. minimal Debye length.

include/picongpu/particles/debyeLength/Check.hpp
```diff
@@ -0,0 +1,198 @@
...
+                        else
+                            log<picLog::PHYSICS>("Check skipped due to no particles");
+                    }
+                };
+
+            } // namespace detail
+
+            /** Check Debye length resolution
+             *
+             * The check is currently done only for species called "e" and is supposed
```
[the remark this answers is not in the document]
sbastrakov: You are right, that comnent is about my orininal implementation and should be updated

include/picongpu/particles/debyeLength/Estimate.kernel
sbastrakov: No apparent reason. I think it makes no significant difference in this case, whether to do it inside or aftet the kernel

include/picongpu/particles/debyeLength/Estimate.kernel
sbastrakov: That's a good point, i should have at least added to tge output it's average. With min I was a bit confused how to do it properly. I think just taking min is susceptible to one unfortunately sampled supercell raising a false alarm. So it has to be some more sophisticated check i guess

PrometheusPi: You could preform the check for resolving the Debye length already on every super cell and the just sum the Boolean result (fulfilled = 0, not fulfilled = 1). If the sum exceeds a "magic" threshold you could throw a warning.

sbastrakov: I agree that would be a more thorough check.

I actually tried to do almost what you are suggesting with the original version of this PR. I think it gets out of hand quite quickly, as then one needs to count "valid" supercells with at least 2 macroparticles and what ratio of those satisfy the check. And we need to somewhat arbitrarily decide what passes or not, see review of Klaus above, it probably should also depend on the particle shape. And it gets hardet to interpret.

I think i can add back this kind of check in addition to the average length output now.

PrometheusPi: I think adding this would be very helpful.  But you are right, with only 2 ppc this might get out of hand due to numerical issues quickly.

include/picongpu/particles/debyeLength/Estimate.kernel
PrometheusPi: Okay.

include/picongpu/particles/debyeLength/Estimate.kernel
sbastrakov: Yes, so I will try, but just to give my opinion since I've already considered it in the past. There are at least two fhings that could be problematic and that pushed me to do the implementation based on averages.

First, I assume that the initial momentum is a sum of a non-random part (like an initial drift) and a random part fully caused by the initial temperature. Since we need to fit distribution parameters (that are directly connected to temperature) of that random part based on the given sample, the sample needs to have a sufficient size. And so the larger volume considered, the better for this part.

However, while inside the simulation I also don't know how to split between the random and non-random parts. So I just assume the non-random part is slowly varying inside each sample volume size, and then I just use sample variance and assume it's fully caused by temperature. So that sample volume should be not too small and not too large, hence a supercell seems reasonable, also convenient computationally.

The averagizing done currently really helps to smooth out everything. I am thinking of adding another output in additing to that, saying something like "X% of supercells have < 1 cell per Debye length, Y% have < 2 cells, Z% have < 3 cells"

steindev: Actually we are only interested in the Debye Length at highest density. If the Debye length is resolved in regions of highest density, it will be resolved everywhere. I think, this is the origin of @PrometheusPi comment and is the reason why a global average is only of limited use.

Is it possible to calculate the average Debye length only for the highest density, or within a ten percent variation of this? The error on the Debye length will then become 5%, which is negligible for resolution considerations. But the value of the Debye length we obtain is the one we need to judge on resolution.

psychocoderHPC: I suggest discussing this tomorrow in the developer meeting and summarise it here. I will reserve 10min for this topic.

sbastrakov: Agree with @psychocoderHPC . However as a question, isn't it possible that initial temperature is not constant in the sumulation domain? If so, it seems only density is not sufficient to determine the "interesting" region for the check.

sbastrakov: I think what you suggest @steindev is not easy to do, but better discuss tomorrow indeed.

steindev: Actually, we could use the ´BASE_DENSITY´ for this estimate. This would be easy, wouldn't it? And as the typical workflow is to set `BASE_DENSITY` to the highest expected density, this would be sufficient. We could even compare the density you calculate to the `BASE_DENSITY` and, first, give a warning if a measured density value is larger than the `BASE_DENSITY`, second, calculate the Debye length only if the measured density is within a 10% variation (or so) around the `BASE_DENSITY` and calculate the average Debye length only from this reduced set of Debye lengths.

sbastrakov: Not sure if I follow. `BASE_DENSITY` is merely a unit, we do not require actual density to be close or not more than `BASE_DENSITY`. And a user doesn't have to initialize their macroparticles with our standard generators anyways. That's actually the whole issue with estimating Debye length: since we allow anything, there is nothing to rely upon before all the init pipeline is done and we have a set of macroparticles.

psychocoderHPC: `BASE_DENSITY` can not be used, as @sbastrakov said this is a unit. It is allowed to initialize a multiple of this value.

sbastrakov: Updated PR description with the points from VC discussion

sbastrakov:
I've pushed an update as discussed on the VC and implemented earlier review suggestions. Now the output looks like

```
Resolving Debye length for species "e"?
PIConGPUVerbose PHYSICS(1) | Estimate used momentum variance in 8784 supercells with at least 10 macroparticles each
PIConGPUVerbose PHYSICS(1) | 5502 (62.6366 percent) supercells had local Debye length estimate not resolved by a single cell
PIConGPUVerbose PHYSICS(1) | Estimated weighted average temperature 3.99277 KeV and corresponding Debye length 1.70519e-07 m.
   The grid has 0.962298 cells per average Debye length

```

include/picongpu/particles/debyeLength/Check.hpp
psychocoderHPC: @sbastrakov I think you missed this comment in the past.

psychocoderHPC (dismissed):

include/picongpu/particles/debyeLength/Check.hpp
psychocoderHPC: please check this comment: https://github.com/ComputationalRadiationPhysics/picongpu/pull/3446/files#r537676156

sbastrakov: done

sbastrakov:
@PrometheusPi @steindev ready for re-review.

psychocoderHPC:
@steindev @PrometheusPi Ping!

psychocoderHPC:
@PrometheusPi @steindev Could you give a time frame for the review.

If not I would like to merge this feature into PIConGPU. 
This PR is not critical for the code and "only" throws a warning and is therefore not breaking existing code.
If we postpone the merge too long it could cost us more developer power to rebase the code multiple times compared to adjust parts in a separate PR later.

psychocoderHPC:
@sbastrakov Could you update the PR description to reflect the latest implementation. One part that has changed is `The check is performed for species "e"  ...`

sbastrakov:
Done

steindev (approved):
This will do. Great! I also like the change to `grid.param`. I am wondering why this is not applied to all examples?

psychocoderHPC (approved):

psychocoderHPC:
> This will do. Great! I also like the change to `grid.param`. I am wondering why this is not applied to all examples?

I will apply this change to all `grid.param` files in a separate PR.

PR #3448, opened 2020-11-26
CI: perform branch check in HZDR CI

psychocoderHPC:
This PR is opened against the master branch and *should* not be merged.
CI should fail in the first test because we do not allow PR's from forks against the master branch.


psychocoderHPC:
This PR was a test for one of our CI tests only.

PR #3450, opened 2020-11-26
CI: auto adjust parallelism

psychocoderHPC:
For a while, there are environment variables available that give information about the available memory and cores in the CI runner.
This PR calculates the parallelism used to compile tests based on this information.

### DIFF
--- a/share/ci/run_picongpu_tests.sh
+++ b/share/ci/run_picongpu_tests.sh
@@ -40,20 +40,31 @@ cd buildCI
 export picongpu_DIR=$CI_PROJECT_DIR
 export PATH=$picongpu_DIR/bin:$PATH
 
-PIC_PARALLEL_BUILDS=$(nproc)
-# limit to $CI_MAX_PARALLELISM parallel builds to avoid out of memory errors
+# adjust number of parallel builds to avoid out of memory errors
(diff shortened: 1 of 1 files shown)


share/ci/run_picongpu_tests.sh
```diff
@@ -40,20 +40,32 @@ cd buildCI
 export picongpu_DIR=$CI_PROJECT_DIR
 export PATH=$picongpu_DIR/bin:$PATH
 
-PIC_PARALLEL_BUILDS=$(nproc)
-# limit to $CI_MAX_PARALLELISM parallel builds to avoid out of memory errors
+# includes 2GiB for the operating system
+PIC_BUILD_REQUIRED_MEM_BYTES=$((9*1024*1024*1024))
```
psychocoderHPC, line 44: I tested this on fwk394 and added a safety margin.

sbastrakov (dismissed):

psychocoderHPC:
The amount of memory required for one build is now configured in the CI web interface and can be adjusted at any time. I set it now to 3GiB per build instance. The reason is that it was before also running with 4 parallel processes on a 12GiB instance.
If there is any issue in the future we can adjust the value,

PrometheusPi (approved):
looks good to me

sbastrakov (approved):

PR #3452, opened 2020-11-26
CI: skip compile

psychocoderHPC:
CI is checking the pull request labels and is skipping compile/runtime tests if the label `CI:no-compile` is pinned.

Currently, our CI is only performing actions when the code is updated by a push/force push.
Therefore the label should be set before the PR is opened or a force push to update the PR is required after the label is set.
By adding the label and later removing it is easy to pass the CI without any compile tests. This can be used to sneak broken code into the mainline. Labels can only be set by maintainers and I assume they will not abuse the flag to bypass the CI.

The flag should be used for documentation updates where compile tests are useless.

**Note:** My original idea was to parse the PR description for actions e.g. `CI:no-compile, CI:reduced-tests, CI:test:examples/FoilLCT`.
Since everyone can change the PR description and we do not perform CI actions in this case I will not introduce this feature until gitlab can work with github pull request without the current CI-bot workaround we are using at HZDR.

### DIFF
--- a/share/ci/generate_reduced_matrix.sh
+++ b/share/ci/generate_reduced_matrix.sh
@@ -17,6 +17,15 @@ echo "  - local: '/share/ci/compiler_nvcc_cuda.yml'"
 echo "  - local: '/share/ci/compiler_clang_cuda.yml'"
 echo ""
 
+# handle CI actions
+has_label=$($CI_PROJECT_DIR/share/ci/pr_has_label.sh "CI:no-compile" && echo "0" || echo "1")
+if [ "$has_label" == "0" ] ; then
+  echo "skip-compile:"
+  echo "  script:"
+  echo "    - echo \"CI action - 'CI:no-compile' -> skip compile/runtime tests\""
+  exit 0
+fi
+
 folders=()
 for CASE in ${PIC_INPUTS}; do
   if [ "$CASE" == "examples" ] || [  "$CASE" == "tests"  ] ; then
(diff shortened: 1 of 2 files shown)


PrometheusPi:
Regarding the planed workflow:
>skipping compile/runtime tests if the label CI:no-compile is pinned

Labels can currently only be set by maintainers - thus either: 
 - we allow all users/developers to set flags
 - or we accept that pushes from developers will trigger the CI's compile test if we not set the labels timely. 

@ComputationalRadiationPhysics/picongpu-maintainers what do you think?

psychocoderHPC:
> Regarding the planed workflow:
> 
> > skipping compile/runtime tests if the label CI:no-compile is pinned
> 
> Labels can currently only be set by maintainers - thus either:
> 
> * we allow all users/developers to set flags
> * or we accept that pushes from developers will trigger the CI's compile test if we not set the labels timely.
> 
> @ComputationalRadiationPhysics/picongpu-maintainers what do you think?

It is not possible that users can set labels and they should not do it to avoid confuaing labeling.
The current choosen why would in the worst case change nothing compared to the current checkes. In the best way it will save few hours CI time. Since currently mostly maintainer push the documentation I do not see any negative effect from this restriction. 
As soon as github PRs can be integrated natively in the gitlab CI we can check if we are able to relax this feature. Currently we are limited. The right way would be to write a bot but I know that this will not happen and is worth the development time.

psychocoderHPC:
This PR can be merged after the CI passed. I tested in a previous CI run that this flag is working. The current CI urn will cover the full compile.

As I wrote above the flag can currently only set by maintainers.

sbastrakov (approved):

PR #3454, opened 2020-11-27
update to latest cupla 0.3.0-dev

psychocoderHPC:
- update cupla to 0.3.0-dev -> https://github.com/alpaka-group/cupla/tree/f60a0ac72cb175d4a8a67301882f7a6b1de4c3c3

```
GIT_AUTHOR_NAME="Third Party" GIT_AUTHOR_EMAIL="picongpu@hzdr.de" git subtree pull --prefix thirdParty/cupla git@github.com:alpaka-group/cupla.git dev --squash
```
- pull the alpaka 0.6.0rc3 via cupla.
- update mallocMC to 2.6.0crp-dev https://github.com/alpaka-group/mallocMC/tree/d7f061a964ad4499f047995763f64d95b7b16055

```
GIT_AUTHOR_NAME="Third Party" GIT_AUTHOR_EMAIL="picongpu@hzdr.de" \
  git subtree pull --prefix thirdParty/mallocMC \
  git@github.com:ComputationalRadiationPhysics/mallocMC.git dev --squash
rm -r thirdParty/mallocMC/alpaka
GIT_AUTHOR_NAME="Third Party" GIT_AUTHOR_EMAIL="picongpu@hzdr.de" git commit --amend -a
```
- apply interface changes for alpaka 0.6.0rc3

# Review:

Only the last commit has hand made changes.

PR #3458, opened 2020-12-02
Update supported CUDA versions

psychocoderHPC:
With #3454 we are now supporting CUDA 9.2 - 11.1.


sbastrakov (approved):
This PR is fine. We should also make the automatic tests run on newly supported CUDA, I guess by modifying containers and in a separate PR.

PR #3459, opened 2020-12-06
[JUWELS] Update modules in PIConGPU profiles.

Anton-Le:
The modules used in `etc/picongpu/juwels-jsc/gpus_picongpu.profile.example` have been updated to versions
available on the JUWELS cluster and booster as of 04.12.2020 (after the Booster + Cluster merge).

Modules used in `etc/picongpu/juwels-jsc/batch_picongpu.profile.example` are essentially identical to
those used in `etc/picongpu/juwels-jsc/gpus_picongpu.profile.example` except for the use of IntelMPI, Intel compiler and Boost 1.73.0.

**Important Note**: 
1. For the `gpu` partition the profile should be sufficient to compile and run PIConGPU, since differences to the `booster` nodes are marginal and the configuration works for the latter.
2. Compilation of ADIOS fails with the Intel compiler (`batch` profile).
3. For the `batch` partition (CPUs) no attempt has been made to run PIConGPU, since we have no access to that particular set of nodes.

## Building dependencies:

In the following the sample process of compiling the necessary dependencies is shown as it has been tested on JUWELS. The process differs minimally from the compilation commands delineated in the [dependencies documentation](https://picongpu.readthedocs.io/en/0.5.0/install/dependencies.html).

Assumptions:
	1. The appropriate `.profile` has successfully loaded.
	2. Sources are stored in a directory `DIR` under `src`, i.e., `DIR/src/adios-1.13.1` and build directories are located in build, i.e., `DIR/build/adios`.

### libSplash:
In the build directory for libSplash:

`cmake -DCMAKE_INSTALL_PREFIX=$LIBSPLASH_ROOT -DSplash_USE_MPI=ON -DSplash_USE_PARALLEL=ON ../../src/libSplash/`
`make && make install`

### c-blosc:
In the build directory for C-BLOSC:

`cmake -DCMAKE_INSTALL_PREFIX=$BLOSC_ROOT -DPREFER_EXTERNAL_ZLIB=ON ../../src/c-blosc/`
`make && make install`

### adios:
In the build riectory for ADIOS execute:

`export MPI_ROOT=$EBROOTPSMPI`

`CFLAGS="-fPIC" ../../src/adios-1.13.1/configure --enable-static --enable-shared --prefix=$ADIOS_ROOT --with-mpi=$MPI_ROOT --with-blosc=$BLOSC_ROOT`

This will pick up the loaded (by hidden module) ZLIB.
`make && make install`


etc/picongpu/juwels-jsc/gpus_picongpu.profile.example
```diff
@@ -11,29 +11,38 @@ export MY_NAME="$(whoami) <$MY_MAIL>"
...
-module load CMake/3.14.0
-module load MVAPICH2/2.3.3-GDR
-module load Python/3.6.8
-
-module load Boost/1.69.0-Python-3.6.8
-module load HDF5/1.10.5
+module load GCC/9.3.0
+module load CUDA/11.0
+module load CMake/3.18.0
+module load ParaStationMPI/5.4.7-1
```
psychocoderHPC, line 39: Is this MPI GPU aware?

psychocoderHPC (dismissed):

psychocoderHPC:
Thanks for updating the profiles. The only open question is if the choosen MPI is GPU aware or not.

psychocoderHPC:
@Anton-Le Could you please remove the end of line white spaces. https://gitlab.com/hzdr/crp/picongpu/-/jobs/893216230#L1722

PrometheusPi (approved):

PrometheusPi:
@Anton-Le Code update looks good - but, as @psychocoderHPC also mentioned, CI fails due to EOL white space - please fix.

etc/picongpu/juwels-jsc/gpus_picongpu.profile.example
Anton-Le: Yes, but by default the `mpi-settings/default` module will be loaded, which works fine but could be replaced by `mpi-settings/CUDA`.
I refrained from adding it, since I am not sure whether this module works OK on the _cluster_ (it does on the booster, though).

psychocoderHPC: Please add this option. Working with a non cuda aware MPI together with GPUs requires additional MPI flags to avoid the MPI is caching freed pointers.

Anton-Le:
I have removed the EOL WS and, for good measure, added the `mpi-settings/CUDA` module (albeit as a comment).

psychocoderHPC:
https://github.com/ComputationalRadiationPhysics/picongpu/pull/3459#discussion_r538103071 should be used.

etc/picongpu/juwels-jsc/gpus_picongpu.profile.example
Anton-Le: Alright. I added it as default.

AndiH: `mpi-settings/CUDA` should also be taken on the Cluster. It sets environment variables to tune PSMPI and especially UCX for GPU-awareness, for example `UCX_TLS=[…,]cuda[,…]`. The environment variables might not be identical on Cluster and Booster; but that's exactly the reason why the module exist – so we can care about this and not the user.

psychocoderHPC (approved):

sbastrakov (approved):

PrometheusPi (approved):

Issue #3357, opened 2020-09-23
Removing of libsplash dependency


psychocoderHPC:
I will remove the dependency `libsplash` latest on 31.10.2020.

The following plugins should be updated before:
- [x] PhaseSpace - will be updated by @franzpoeschel 
- [x] ParticleCalorimeter @franzpoeschel 
- [x] Radiation - should be updated by @PrometheusPi 
- [x] HDF5Writer - will be removed, please use new plugin `openPMDWriter`
- [x] density profile see: #3515 @sbastrakov 
- [x] plugin MacroParticleCounter @franzpoeschel 

@PrometheusPi @HighIander  Who can do this, who knows the plugin best?

PrometheusPi:
@psychocoderHPC I will do the radiation plugin - ParticleCalorimeter is probably best handled by @n01r - but he has no time.

sbastrakov:
We discussed offline with @psychocoderHPC , I could take on some work on this, at start of October. Do not have a particular preference.

sbastrakov:
Good note on updating the docs when all this is implemented: https://github.com/ComputationalRadiationPhysics/picongpu/pull/3361#issuecomment-708969488

steindev:
libsplash does not compile anymore against a recent version of the HDF5 library. We really should get this done. Volunteers?

franzpoeschel:
If there are still unassigned plugins to convert, I think I can take one. Which one would be the best?

steindev:
> If there are still unassigned plugins to convert, I think I can take one. Which one would be the best?

Wonderful! How about ParticleCalorimeter? Thank you so much!

steindev:
UPDATE: Just chatted with @PrometheusPi. He is working on the Radiation Plugin.

sbastrakov:
So with these updates only density profile is not assigned? If so, i will take it

franzpoeschel:
> How about ParticleCalorimeter

Ok, sounds good. I'll have a look.

sbastrakov:
There was already a PR submitted #3560 .

sbastrakov:
(Nevermind my deleted message, I can't read)

PrometheusPi:
@psychocoderHPC what will happen to `PICToSplash` after libSplash has been removed. Will it be still available and usable, as e.g. [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3560/files#r600551307)?

[...]

PR #3468, opened 2020-12-09
Use the openPMD API for the PhaseSpace plugin (formerly libsplash)

franzpoeschel:
See issue #3357.

TODO:

- [x] Remove libsplash output. Currently, both versions of output are produced. I'll keep libsplash around for a while, this makes for easier comparison.
- [x] File naming? The current plugin encodes the species name in the file name and writes separate files per species (e.g. `PhaseSpace_e_all_ypy_100.h5`). Since that information is classically encoded in the dataset path in openPMD, I think we could also write everything into one file (e.g. `PhaseSpace_400.h5 `). Would require some small refactoring to keep the `Series` handle open between different calls to the PhaseSpace plugin, much like the openPMD plugin already does. (Which is also necessary to support group-based iteration layout, e.g. for streaming).
    EDIT: As discussed below, we'll stick with the old file naming.
- [x] Attributes. The PhaseSpace plugin writes some attributes, and for some of them I had to take a guess concerning their meaning. 
    Some, I just copied verbatim (sim_unit, p_unit, p_min, p_max, dr, dV). 
    Others, I took a guess on what they mean within openPMD:
    * movingWindowOffset → becomes gridGlobalOffset. Not too sure about this one, since the gridGlobalOffset is in terms of physical units, while the movingWindowOffset is solely in terms of indices. Alternatively, use a custom attribute.
    * movingWindowSize: dropped, since it's the same as the dataset size in dimension x
    * dr_unit → becomes gridUnitSI. I don't know whether the scaling is the same in all dimensions.
    * dt → becomes dt (iteration attribute)
    * dt_unit → becomes timeUnitSI

    Some support by an actual user of the PhaseSpace plugin would be appreciated.

    Also, some openPMD attributes are still left as default, need to fill in those, too. 

    I'll need to figure out how to express datasets with different physical units and scalings per dimension.

    Old:
      hdf5
    New:
    ```
  float     /data/600/dt                                 attr   = 1
  string    /data/600/meshes/e_all_ypy/axisLabels        attr   = {"x"}
  float     /data/600/meshes/e_all_ypy/dV                attr   = 19.2232
  string    /data/600/meshes/e_all_ypy/dataOrder         attr   = "C"
  float     /data/600/meshes/e_all_ypy/dr                attr   = 1.06309
  string    /data/600/meshes/e_all_ypy/geometry          attr   = "cartesian"
  double    /data/600/meshes/e_all_ypy/gridGlobalOffset  attr   = {0, 0}
  double    /data/600/meshes/e_all_ypy/gridSpacing       attr   = 1
  double    /data/600/meshes/e_all_ypy/gridUnitSI        attr   = 4.16712e-08
  float     /data/600/meshes/e_all_ypy/p_max             attr   = 0.00014378
  float     /data/600/meshes/e_all_ypy/p_min             attr   = -0.00014378
  double    /data/600/meshes/e_all_ypy/p_unit            attr   = 1.89938e-18
  double    /data/600/meshes/e_all_ypy/position          attr   = 0
  double    /data/600/meshes/e_all_ypy/sim_unit          attr   = 1.53994e+07
  float     /data/600/meshes/e_all_ypy/timeOffset        attr   = 0
  double    /data/600/meshes/e_all_ypy/unitDimension     attr   = {0, 0, 0, 0, 0, 0, 0}
  double    /data/600/meshes/e_all_ypy/unitSI            attr   = 1
  double    /data/600/time                               attr   = 0
  double    /data/600/timeUnitSI                         attr   = 1.39e-16
    ```

- [x] Update the Python analysis scripts. EDIT: They seem to be fundamentally working, but I'll need to re-apply any changes there that are done to the attribute layout. Also some todos and further testing. Also, read from backends other than HDF5.
- [x] Documentation.
- [x] Cleanup

Related: https://github.com/openPMD/openPMD-standard/pull/193. Necessary for expressing this data layout properly in openPMD, merged already in the standard, not yet available in the API.


psychocoderHPC:
For the fule nameingn Can we close the file each time and open the file if required already existing. As you wrote you can call this pr for each species and for different directions. Since each species is an independent plugin instance it is not possible in PIC to figure out which instance mist close the file.

If this would be to complicated we should keep the file naming even if they have redundante information which are already in the file.

franzpoeschel:
> If this would be to complicated we should keep the file naming even if they have redundante information which are already in the file.

Then that's probably what we are going to have to do. Not all our backends support a ReadWrite mode, so this would not work in general.

psychocoderHPC:
@PrometheusPi @steindev @wmles @Anton-Le Could you please help and encrypt the meanings of the `sim_unit, p_unit, p_min, p_max, dr, dV`

I would say `dim_unit` is to convert the value from the matrix into SI and should be applied to each array cell.
`p_min` minimum energy and `p_max` the maximum energy (Y axis of the array)

steindev:
> @PrometheusPi @steindev @wmles @Anton-Le Could you please help and encrypt the meanings of the `sim_unit, p_unit, p_min, p_max, dr, dV`

Phase space information is particle information which combines one component of a particle's momentum vector `p = (p_x, p_y, p_z)` with one component of its position vector `r = (r_x, r_y, r_z)` at a particular time.

The phase space is the aggregation of this information for all particles for the chosen momentum and position component.

But the phase space is not a pure scatter plot of (r_i, p_j) for all particles.
Rather it is a histogram, meaning that the individual particle information is binned.
In conclusion, a phase space provides the number of particles in an interval of a momentum and position space in form of a charge density on a grid whose axes are defined by the chosen spatial and momentum directions.

`p_min` and `p_max` are the lower and upper bound of the momentum axis, respectively.
These are configured in a simulation's `*.cfg` by setting the `minEnergy` and `maxEnergy` options of the histogram plugin, respectively.

`dr` is the spacing of the spatial axis in PIConGPU units, i.e. the cell size along the chosen spatial coordinate. Can differ depending on the chosen component.

`dV` is the volume of a phase space cell, i.e. `dV = dp * dr`, where `dp` would be the grid spacing along the momentum axis.

`p_unit` is the factor to convert momentum values from PIConGPU scaling to SI.

`dr_unit` is the factor to convert spatial coordinates from PIConGPU scaling to SI. Should be the same for all dimensions.

`sim_unit` (I have to guess here) is the factor to convert the charge density value of the phase space from PIConGPU scaling to SI (remember the value provided is charge/length/momentum which needs to be converted from PIConGPU to SI).

franzpoeschel:
Thanks for the detailed description!

> `p_min` and `p_max` are the lower and upper bound of the momentum axis, respectively.
> These are configured in a simulation's `*.cfg` by setting the `minEnergy` and `maxEnergy` options of the histogram plugin, respectively.

In combination with [the openPMD standard files](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md), this would then mean that we would set: `gridGlobalOfset = (<movingWindowOffset>, p_min)`, where `<movingWindowOffset>` is given not as an index, but as the point in simulation coordinates. `p_max` could theoretically be restored by calculating it from the dataset dimensions. Writing custom attributes `p_min` and `p_max` additionally to `gridGlobalOffset` seems good though.
> 
> `dr` is the spacing of the spatial axis in PIConGPU units, i.e. the cell size along the chosen spatial coordinate. Can differ depending on the chosen component.
> 
> `dV` is the volume of a phase space cell, i.e. `dV = dp * dr`, where `dp` would be the grid spacing along the momentum axis.

Hence, `gridSpacing = (dr, dp)`.
> 
> `p_unit` is the factor to convert momentum values from PIConGPU scaling to SI.
> 
> `dr_unit` is the factor to convert spatial coordinates from PIConGPU scaling to SI. Should be the same for all dimensions.
 
I was not talking about the simulation dimensions, but the dimensions of the dataset (i.e. the momentum and the position dimension). `p_unit` and `dr_unit` both do roughly what openPMD calls `gridUnitSI`, but openPMD currently only allows one global value for all dataset dimensions, so we'll have to stick with custom attributes for now and see whether we should extend the standard for this use case. Similar issue with `unitDimension`.

> `sim_unit` (I have to guess here) is the factor to convert the charge density value of the phase space from PIConGPU scaling to SI (remember the value provided is charge/length/momentum which needs to be converted from PIConGPU to SI).

From the implementation, that looks like you're right:
```cpp
const float_64 unit = UNIT_CHARGE / UNIT_VOLUME;
```

I should additionally use custom attributes for `windowMoving(Offset|Extent)` probably.

steindev:
> > `p_min` and `p_max` are the lower and upper bound of the momentum axis, respectively.
> > These are configured in a simulation's `*.cfg` by setting the `minEnergy` and `maxEnergy` options of the histogram plugin, respectively.
> 
> In combination with [the openPMD standard files](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md), this would then mean that we would set: `gridGlobalOfset = (<movingWindowOffset>, p_min)`, where `<movingWindowOffset>` is given not as an index, but as the point in simulation coordinates.

`p_min` will usually be negative, while `p_max` will be positive. I would not offset the momentum values, just put them in the histogram as they are. 

>`p_max` could theoretically be restored by calculating it from the dataset dimensions. Writing custom attributes `p_min` and `p_max` additionally to `gridGlobalOffset` seems good though.

This requires to know `dp`, but currently this is not in the output. But will be with your addition of the value to `gridSpacing` :+1:
However, these values are in the output just for convenience anyways.
They could be retrieved from the data itself.
But the python script for plotting uses these, so leaving them in seems to be the better option to me.

> > `dr` is the spacing of the spatial axis in PIConGPU units, i.e. the cell size along the chosen spatial coordinate. Can differ depending on the chosen component.
> > `dV` is the volume of a phase space cell, i.e. `dV = dp * dr`, where `dp` would be the grid spacing along the momentum axis.
> 
> Hence, `gridSpacing = (dr, dp)`.
> 
> > `p_unit` is the factor to convert momentum values from PIConGPU scaling to SI.
> > `dr_unit` is the factor to convert spatial coordinates from PIConGPU scaling to SI. Should be the same for all dimensions.
> 
> I was not talking about the simulation dimensions, but the dimensions of the dataset (i.e. the momentum and the position dimension). `p_unit` and `dr_unit` both do roughly what openPMD calls `gridUnitSI`, but openPMD currently only allows one global value for all dataset dimensions, so we'll have to stick with custom attributes for now and see whether we should extend the standard for this use case. Similar issue with `unitDimension`.

The upcoming 2.0.0. standard allows `gridUnitSI` to be an array, just as grid spacing. Something similar applies to `gridUnitDimension`. I suggest to directly implement what will be conform with the new standard.

> > `sim_unit` (I have to guess here) is the factor to convert the charge density value of the phase space from PIConGPU scaling to SI (remember the value provided is charge/length/momentum which needs to be converted from PIConGPU to SI).
> 
> From the implementation, that looks like you're right:
> 
> ```c++
> const float_64 unit = UNIT_CHARGE / UNIT_VOLUME;
> ```
If `UNIT_VOLUME` is the phase space volume.
But usage of `sim_unit` in the python plotting script suggest that my guess is correct, too.
 
> I should additionally use custom attributes for `windowMoving(Offset|Extent)` probably.

I would love to have the original window size in the data.
As the particle distribution develops over time and can spatially spread out, the varying spatial axis scale is irritating when plotting with the current version where the axis extent is defined by the lowest and largest particle position.

The Offset is already in `gridGlobalOffset`, there is no need to put it again, is there?

franzpoeschel:
> > > `p_min` and `p_max` are the lower and upper bound of the momentum axis, respectively.
> > > These are configured in a simulation's `*.cfg` by setting the `minEnergy` and `maxEnergy` options of the histogram plugin, respectively.
> > 
> > 
> > In combination with [the openPMD standard files](https://github.com/openPMD/openPMD-standard/blob/latest/STANDARD.md), this would then mean that we would set: `gridGlobalOfset = (<movingWindowOffset>, p_min)`, where `<movingWindowOffset>` is given not as an index, but as the point in simulation coordinates.
> 
> `p_min` will usually be negative, while `p_max` will be positive. I would not offset the momentum values, just put them in the histogram as they are.

The `gridGlobalOffset` just states where the origin of the dataset is within the physical domain, so I wouldn't say this would be offsetting the momentum values, but rather just giving the right info for that attribute.
> 
> > `p_max` could theoretically be restored by calculating it from the dataset dimensions. Writing custom attributes `p_min` and `p_max` additionally to `gridGlobalOffset` seems good though.
> 
> This requires to know `dp`, but currently this is not in the output. But will be with your addition of the value to `gridSpacing` +1
> However, these values are in the output just for convenience anyways.
> They could be retrieved from the data itself.
> But the python script for plotting uses these, so leaving them in seems to be the better option to me.

Yep, I'll leave them in then. Although I need to touch those Python scripts anyway ;)
> 
> > > `dr` is the spacing of the spatial axis in PIConGPU units, i.e. the cell size along the chosen spatial coordinate. Can differ depending on the chosen component.
> > > `dV` is the volume of a phase space cell, i.e. `dV = dp * dr`, where `dp` would be the grid spacing along the momentum axis.
> > 
> > 
> > Hence, `gridSpacing = (dr, dp)`.
> > > `p_unit` is the factor to convert momentum values from PIConGPU scaling to SI.
> > > `dr_unit` is the factor to convert spatial coordinates from PIConGPU scaling to SI. Should be the same for all dimensions.
> > 
> > 
> > I was not talking about the simulation dimensions, but the dimensions of the dataset (i.e. the momentum and the position dimension). `p_unit` and `dr_unit` both do roughly what openPMD calls `gridUnitSI`, but openPMD currently only allows one global value for all dataset dimensions, so we'll have to stick with custom attributes for now and see whether we should extend the standard for this use case. Similar issue with `unitDimension`.
> 
> The upcoming 2.0.0. standard allows `gridUnitSI` to be an array, just as grid spacing. Something similar applies to `gridUnitDimension`. I suggest to directly implement what will be conform with the new standard.

Ah, that's good to know. I will need to check whether the openPMD API will let me implement it in that way ahead of time. If it doesn't (which I'm afraid it will), some custom solution will have to do in the meantime.
> 
> > > `sim_unit` (I have to guess here) is the factor to convert the charge density value of the phase space from PIConGPU scaling to SI (remember the value provided is charge/length/momentum which needs to be converted from PIConGPU to SI).
> > 
> > 
> > From the implementation, that looks like you're right:
> > ```c++
> > const float_64 unit = UNIT_CHARGE / UNIT_VOLUME;
> > ```
> 
> If `UNIT_VOLUME` is the phase space volume.
> But usage of `sim_unit` in the python plotting script suggest that my guess is correct, too.
> 
> > I should additionally use custom attributes for `windowMoving(Offset|Extent)` probably.
> 
> I would love to have the original window size in the data.

Can you specify what you mean by this? Make a global (non-window) dataset, but only write to those positions that are covered by the window?
> As the particle distribution develops over time and can spatially spread out, the varying spatial axis scale is irritating when plotting with the current version where the axis extent is defined by the lowest and largest particle position.
> 
> The Offset is already in `gridGlobalOffset`, there is no need to put it again, is there?

The  `gridGlobalOffset` is within the physical domain, while the movingWindowOffset are just indexes.

steindev:
> > I would love to have the original window size in the data.
> 
> Can you specify what you mean by this? Make a global (non-window) dataset, but only write to those positions that are covered by the window? dcrt
> 
> > As the particle distribution develops over time and can spatially spread out, the varying spatial axis scale is irritating when plotting with the current version where the axis extent is defined by the lowest and largest particle position.

Just the extent/size of the spatial grid. In order to set the extent of the axis of the spatial component to the original simulation window extent when plotting.

franzpoeschel:
Ah, alright. Yeah that should be no problem.

franzpoeschel:
> Just the extent/size of the spatial grid. In order to set the extent of the axis of the spatial component to the original simulation window extent when plotting.

To get back to this again, since I don't have such in-depth knowledge on the internals of PIC:
Currently, the extent of the dataset is computed via:
```
            const SubGrid<simDim>& subGrid = Environment<simDim>::get().SubGrid();
            const std::uint64_t rGlobalSize = subGrid.getGlobalDomain().size[axis_element.space];
            splash::Dimensions globalPhaseSpace_size(hBuffer.size().x(), rGlobalSize, 1);
```
If I'm interpreting this correctly, there is no moving window involved in that?
If that's insufficient, I guess it would be the dimensions given via command line parameter `-g` that you'd like to see?

franzpoeschel:
> Ah, that's good to know. I will need to check whether the openPMD API will let me implement it in that way ahead of time. If it doesn't (which I'm afraid it will), some custom solution will have to do in the meantime.

Update: It will not allow me to do this:
```
RuntimeError: Unexpected Attribute datatype for 'gridUnitSI'
```

steindev:
> If that's insufficient, I guess it would be the dimensions given via command line parameter `-g` that you'd like to see?
Yes, together with the `gridGlobalOffset` this would be the information one usually requires when plotting.
> EDIT: But wouldn't those really just be the other entries in `subGrid.getGlobalDomain().size`, and they are irrelevant to the PhaseSpace?

Now I am not sure what you mean. As far as I remember, information on the simulation grid size has never been part of the information given in phase space output. One could just infer the extent of the phase space's spatial axis from the data, and this extent is given by the positions of the particles. But these will usually not be at the simulation volume boundary. Therefore there was no chance to infer the simulation grid size (along the chosen spatial axis) from the phase space output. Your questions sounds like the simulation grid size along the chosen spatial axis will be in the phase space output anyways when using openPMD-api. If this is the case, this would be all I asked for.

franzpoeschel:
> > If that's insufficient, I guess it would be the dimensions given via command line parameter `-g` that you'd like to see?
> 
> Yes, together with the `gridGlobalOffset` this would be the information one usually requires when plotting.
> 
> > EDIT: But wouldn't those really just be the other entries in `subGrid.getGlobalDomain().size`, and they are irrelevant to the PhaseSpace?
> 
> Now I am not sure what you mean. As far as I remember, information on the simulation grid size has never been part of the information given in phase space output. One could just infer the extent of the phase space's spatial axis from the data, and this extent is given by the positions of the particles. But these will usually not be at the simulation volume boundary. Therefore there was no chance to infer the simulation grid size (along the chosen spatial axis) from the phase space output. Your questions sounds like the simulation grid size along the chosen spatial axis will be in the phase space output anyways when using openPMD-api. If this is the case, this would be all I asked for.

I have now added four new attributes: globalDomainOffset, globalDomainSize, totalDomainOffset and totalDomainSize, as found in [`SubGrid.hpp`](https://github.com/ComputationalRadiationPhysics/picongpu/blob/454a28efe6a6eb0fa1e0e97031ef2569a1ad313b/include/pmacc/mappings/simulation/SubGrid.hpp#L118). Hopefully this will be sufficient for your purposes?

A current comparison of the attributes written by the old PhaseSpace plugin:
Bildschirmfoto von 2021-01-04 14-51-45

…and the new one:
Bildschirmfoto von 2021-01-04 14-51-34

As stated in the [updated documentation](https://github.com/ComputationalRadiationPhysics/picongpu/blob/47052246b4dccab1bd1a38fab04385ac78731582/docs/source/usage/plugins/phaseSpace.rst) on the PhaseSpace plugin, the attributes `unitSI` and `unitDimension` are to be ignored.

steindev:
Excellent! I don't understand the difference between the pair `globalDomainSize`, `globalDomainOffset` and `totalDomainSize`, `totalDomainOffset`, though. It is a bit of a pity, that we cannot give the axis labels for the `globalDomainSize`, since the `axisLabels` key clearly must define the axis labels of the actual phase space.
This would help to select the item or the correct axis from the `globalDomainSize` (and the others) arrays.
Or could we add an out-of-standard key `globalDomainAxisLabels`? :sweat_smile: Sorry for asking for even more now. You do a great job!

Btw. I would do a test simulation with the new plugin this week. Besides the source code from this PR, do I need anything else to run this updated plugin on hemera at HZDR?

franzpoeschel:
> I don't understand the difference between the pair `globalDomainSize`, `globalDomainOffset` and `totalDomainSize`, `totalDomainOffset`, though.

To be fair, I don't know what the difference is either. Hence why I simply forwarded the terms that PIConGPU uses there. If someone with more knowledge on the internals of PIConGPU could quickly check (@psychocoderHPC ?), that would be great.

> Or could we add an out-of-standard key globalDomainAxisLabels?

More custom attributes can always be added, sure ;) 

> Btw. I would do a test simulation with the new plugin this week. Besides the source code from this PR, do I need anything else to run this updated plugin on hemera at HZDR?

Great! You will need openPMD API 0.12.0 or newer, and either ADIOS2 (2.6.0 or newer) or Parallel HDF5. HDF5 is the default output, so I suggest you use that (otherwise you will have to use `--e_phaseSpace.ext bp` and use the `file_ext` parameter in the Python scripts).

sbastrakov:
@steindev @franzpoeschel PIConGPU domain notation is explained [on this Wiki page](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-domain-definitions).

franzpoeschel:
Thanks for clarifying! I think, I'll leave both naming levels in the output then and link the Wiki page in the documentation?

psychocoderHPC:
@franzpoeschel Do you currently wait for more input from our side? If not I would suggest to clean the pr to bring it into a finale state.

franzpoeschel:
Yep, I'd need some feedback from testers @psychocoderHPC.

@steindev Did you do a test run yet?

psychocoderHPC:
> Yep, I'd need some feedback from testers @psychocoderHPC.
> 
> @steindev Did you do a test run yet?

@wmles @Anton-Le @PrometheusPi Please help with testing. If there is no feedback until 03.02.2021 we will merge this pr with the risk of breaking the plugin.

franzpoeschel:
[This branch](https://github.com/franzpoeschel/picongpu/tree/topic-phaseSpace-via-openPMD-removesplash) removes the Splash output.

psychocoderHPC:
> [This branch](https://github.com/franzpoeschel/picongpu/tree/topic-phaseSpace-via-openPMD-removesplash) removes the Splash output.

Thanks, we need to push the changes into this PR.

psychocoderHPC:
Could you please rebase. There is one conflict and squash the commits.

franzpoeschel:
Done

franzpoeschel:
The failing tests report `There has been a structural integrity problem detected, please contact system administrator`. I would assume, this has nothing to do with this PR?

psychocoderHPC:
> The failing tests report `There has been a structural integrity problem detected, please contact system administrator`. I would assume, this has nothing to do with this PR?

I restarted the broken tests

steindev:
@franzpoeschel Did you do a runtime test with new output method on `hemera`? I setup a real world test and get errors. See 
[stderr.txt](https://github.com/ComputationalRadiationPhysics/picongpu/files/5894129/stderr.txt)

Seems to be related to HDF5, so I am not sure if this is a hemera or filesystem issue.

franzpoeschel:
I tested it on FWK394, I'll run a test on Hemera to see whether I get the same error. @steindev

franzpoeschel:
It's running cleanly for me on Hemera/K80. Since this error throws from HDF5, can you maybe try setting `--e_phaseSpace.ext bp` to check whether this issue is contained to HDF5? @steindev

psychocoderHPC:
> @franzpoeschel Did you do a runtime test with new output method on `hemera`? I setup a real world test and get errors. See
> [stderr.txt](https://github.com/ComputationalRadiationPhysics/picongpu/files/5894129/stderr.txt)
> 
> Seems to be related to HDF5, so I am not sure if this is a hemera, filesystem or implementation issue.

Do you repeatet the run? It looks like a file system issue.

steindev:
> @wmles @Anton-Le @PrometheusPi Please help with testing. If there is no feedback until 03.02.2021 we will merge this pr with the risk of breaking the plugin.

Please do not merge, yet. I will continue with the test setup and the test itself next week.
Unfortunately I am struggling with hemera k20.

franzpoeschel:
Still the same file-accessing issue? If yes, does the old version of the plugin work?

steindev:
So, using ADIOS helped. As I did not know that the file type can be changed for every phase space (and not only for every species), I had a mixture of `h5` and `bp` phase spaces before which let the simulation run at least for a while. I will check the phase space data now.

steindev:
One thing I just noted (because I recompiled my example): picongpu does not compile without libsplash with a rather cryptic error: `PhaseSpace is not a template` in `/plugins/PluginController.hpp`. This dependency should be removed with this PR (at least for phase, if poosible).

franzpoeschel:
> One thing I just noted (because I recompiled my example): picongpu does not compile without libsplash with a rather cryptic error: `PhaseSpace is not a template` in `/plugins/PluginController.hpp`. This dependency should be removed with this PR (at least for the phase plugin, if possible).

Good catch, I forgot updating the condition on the `#include`s. Hope the recent commit should fix this.

psychocoderHPC:
@franzpoeschel before we merge this PR we need to wait until #3516 is in the dev branch and rebase this PR against dev. Currently, the CI is not activating OpenPMD.

sbastrakov:
@franzpoeschel #3516  is now merged, so no issues there.

franzpoeschel:
Ok, I rebased it

steindev:
Coming back to the error where h5 files are not written with the new plugin. (The one I received earlier.) Do you have a way to check whether this is an issue with the openPMD plugin, @franzpoeschel? The error only occurs when writing phase space output with the new plugin to `*.h5` files. It does not occur when writing to `*.bp` files it does **not** occur with the current *dev* and the old (libsplash based) plugin. To test this, I took the same setup and `*.cfg` file and compiled once against the current dev branch and once against your branch. The test ran on 2 k80 nodes at *hemera*, the partition you tried earlier yourself.
You can find and (hopefully) access the setup in `/bigdata/hplsim/scratch/steinigk/projects/2021-01_phase-space-plugin-on-openPMD-PR/01_current-filamentation-instability`

The `stderr` output is attached.
[stderr.txt](https://github.com/ComputationalRadiationPhysics/picongpu/files/5961314/stderr.txt)

Is there anything else I could try to pinpoint the problem?
@PrometheusPi are you currently writing `*.h5` files with the openPMD plugin on hemera or did you switch to ADIOS too?

franzpoeschel:
I'll try to see whether I can reproduce the error with this

steindev:
I am afraid to say, but there are massive differences between the new and the old plugin. I suspected this earlier, and it is the reason why I rerun everything (see comment above) and all of this testing takes so long. Running the exact same simulation, the old and new plugin produce different results in y-px phase space created from all particles (I did not check others). See the comment above for the location of the simulation setup. The setup simulates a relativistic laser pulse impinging on an 8µm thick, thrice pre-ionized, cold Aluminum foil which starts at 12µm in the simulation domain. Ionization modules are **not** activated.

image
`Total number of particles in phase space w/ openPMD PS plugin  = 2657180450816.000000`

image
`Total number of particles in phase space w/ libsplash PS plugin = 6771643514880.000000`

image
`Difference in total number of particles in phase space between plugins = -4114463064064`

Looking at the images, the differences concentrate at the beginning of the foil located around 12µm in the domain and the bulk of the foil (=electrons with no transverse momentum located in the region from 15-20µm).

Generally, I expect differences between the simulations, as the foil is initially cold and an instability is driven.
But the differences are too large to be attributed to this.
Especially the missing bulk electrons suggest an error in the new version of the phase space plugin.
Btw, I checked the total number of macro particles, too. It is equal in both simulations.

Is there anything else I can do to help finding the error?

franzpoeschel:
I can reproduce the file accessing issues with the Hemera openPMD module, a self-built version runs fine. I will try to locate the issue more precisely.

Thanks for the measurements, I will have a look at them.

franzpoeschel:
Alright let's talk about debugging for two days just to find out I forgot to pass the MPI communicator

steindev:
@franzpoeschel I think you need to trigger the CI again by `git commit --amend; git push -f` as the CI failed to download picongpu from github before actually starting compiling.

franzpoeschel:
Thanks for the hint, trying that again

psychocoderHPC (approved):

psychocoderHPC:
@steindev feel free to merge this PR after your final review

steindev:
Finally, tests finished. In order to see differences, I used that laser on Aluminum foil simulation described earlier but with 400keV initial electron temperature, made a checkpoint in the middle of the simulation and created phase spaces with the old plugin and the new plugin (ADIOS2 as well as HDF5) one step after restarting from the checkpoint.
Here are the results:
* total electron phase space with the old libsplash plugin
image

* difference of the total electron phase written the new openPMD PS plugin w/ h5 output to the old libsplash plugin
image
Difference in total number of particles in phase space between plugins = 0
Relative difference in total number of particles in phase space between plugins = 0.00e+00

* difference of the total electron phase written the new openPMD PS plugin w/ bp output to the old libsplash plugin
image
Difference in total number of particles in phase space between plugins = 0
Relative difference in total number of particles in phase space between plugins = 0.00e+00

* difference of the total electron phase written the new openPMD PS plugin w/ bp output to the new openPMD PS plugin w/ h5 output
image
Difference in total number of particles in phase space between plugins = 0
Relative difference in total number of particles in phase space between plugins = 0.00e+00

In essence, the difference is on the single particle level between both, new and old plugin as well as new and new plugin.
This leads me to the conclusion, that everything is fine now :smiley:

PR #3471, opened 2020-12-11
disable async device streams for CPU backends

psychocoderHPC:
When enabling asynchronous CPU backends in alpaka/cupla PIConGPU will block one core per MPI rank to poll for the task (alpaka kernel,memcopy) state. This will block one core and also an asynchronous will oversubscribe the CPU which results mostly in bad performance.

By switching to synchronous task execution, by setting the cmake flag `-DCUPLA_STREAM_ASYNC_ENABLE` we can get up to 30% more performance (example KHI).
MPI calls will still be asynchronous executed.

Since the year 2020 is ending and was definitely not the best year, compared to the years before, I would give this improvement to our CPU user-community. :christmas_tree:

### DIFF
--- a/bin/pic-configure
+++ b/bin/pic-configure
@@ -64,22 +64,22 @@ get_backend_flags()
             result+=" -DALPAKA_CUDA_ARCH=\"${backend_cfg[1]}\""
         fi
     elif [ "${backend_cfg[0]}" == "omp2b" ] ; then
-        result+=" -DALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE=ON"
+        result+=" -DALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE=ON -DCUPLA_STREAM_ASYNC_ENABLE=OFF"
         if [ $num_options -eq 2 ] ; then
             result+=" -DPMACC_CPU_ARCH=\"${backend_cfg[1]}\""
         fi
     elif [ "${backend_cfg[0]}" == "serial" ] ; then
-        result+=" -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE=ON"
+        result+=" -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE=ON -DCUPLA_STREAM_ASYNC_ENABLE=OFF"
         if [ $num_options -eq 2 ] ; then
             result+=" -DPMACC_CPU_ARCH=\"${backend_cfg[1]}\""
         fi
     elif [ "${backend_cfg[0]}" == "tbb" ] ; then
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
btw: we are doing that already for the SPEC Benchmark branch.

psychocoderHPC:
NVIDIA V100 KHI:

```
# dev branch
./bin/picongpu -d 1 1 1  -g  128 128 12 --periodic 1 1 1 -s 100
PIConGPUVerbose PHYSICS(1) | species e: omega_p * dt <= 0.1 ? 0.0319333
PIConGPUVerbose PHYSICS(1) | species i: omega_p * dt <= 0.1 ? 0.000745229
PIConGPUVerbose PHYSICS(1) | macro particles per device: 39321600
PIConGPUVerbose PHYSICS(1) | typical macro particle weighting: 326.577
PIConGPUVerbose PHYSICS(1) | UNIT_SPEED 2.99792e+08
PIConGPUVerbose PHYSICS(1) | UNIT_TIME 1.79e-16
PIConGPUVerbose PHYSICS(1) | UNIT_LENGTH 5.36628e-08
PIConGPUVerbose PHYSICS(1) | UNIT_MASS 2.97492e-28
PIConGPUVerbose PHYSICS(1) | UNIT_CHARGE 5.23234e-17
PIConGPUVerbose PHYSICS(1) | UNIT_EFIELD 9.5224e+12
PIConGPUVerbose PHYSICS(1) | UNIT_BFIELD 31763.3
PIConGPUVerbose PHYSICS(1) | UNIT_ENERGY 2.67372e-11
...
calculation  simulation time: 54sec 249msec = 54 sec
```

```
# This PR
./bin/picongpu -d 1 1 1  -g  128 128 12 --periodic 1 1 1 -s 100

...
calculation  simulation time: 44sec 919msec = 44 sec
...

sbastrakov (commented):
Do we want it for serial as well (like you did with this PR)? Since there we don't oversubscribe in that case anyhow. I don't think it ultimately matters much since this is a debug-only configuration

psychocoderHPC:
> Do we want it for serial as well (like you did with this PR)? Since there we don't oversubscribe in that case anyhow. I don't think it ultimately matters much since this is a debug-only configuration

Yes, we should enable it for serial too. On CPU it will only block one core and poll without any positive effect. 
We disabled even for serial accelerator the [usage of multiple streams on the CPU](https://github.com/ComputationalRadiationPhysics/picongpu/blob/fc613ae0a96a5ba83ecd9db16c2a0ef500c04392/include/picongpu/simulation/control/MySimulation.hpp#L443-L446)

sbastrakov (approved):

PR #3475, opened 2020-12-16
Fix the scheme of updating convolutional B in PML by half time step

### LINKED ISSUE #3418: PML seems to underperform in terms of absorbing strength
As shown by plots in #3417 if compared to results from the Taflove book. This should be orthogonal to changes in that PR tho, as every PML seem to (under)perform the same.

I am looking at it now.

sbastrakov:
The previously used scheme mistakenly assumed that updating by dt/2 twice is equivalent to updating by dt.
It resulted in worse than expected absorption.
The new scheme properly performs update of B for convolutional components.
The changes are fully contained inside the YeePML field solver and so the rest of the code is not affected.

Added comments of why do we have to update B by dt/2 this way.

Fixes #3418. Attaching a plot with new absorption after the fix for the test problem discussed there and in #3417: Taflove 3rd ed., section 7.11.1. edit: CPML uses parameter alpha = 0.2 as in the Taflove book for this test problem. By UPML, same as in Taflove, I mean CPML with alpha = 0. Theoretically it is supposed to be worse or about same as CPML. It would be more accurate to actually call it "CPML with properties of UPML". edit2: comparison of current dev vs. this branch is a later message.
pml__fix

### DIFF
--- a/include/picongpu/fields/MaxwellSolver/YeePML/YeePML.hpp
+++ b/include/picongpu/fields/MaxwellSolver/YeePML/YeePML.hpp
@@ -88,33 +88,38 @@ namespace picongpu
                             return *(fieldB.get());
                         }
 
-                        /** Propagate B values in the given area by half a time step
+                        /** Propagate B values in the given area by the first half of a time step
                          *
-                         * @tparam T_Area area to apply updates to, the curl must be
-                         * applicable to all points; normally CORE, BORDER, or CORE + BORDER
+                         * This operation propagates grid values of field B by dt/2 and prepares the internal state of
+                         * convolutional components so that calling updateBSecondHalf() afterwards competes the update.
+                         *
+                         * @tparam T_Area area to apply updates to, the curl must be applicable to all points;
+                         * normally CORE, BORDER, or CORE + BORDER
                          *
                          * @param currentStep index of the current time iteration
                          */
                         template<uint32_t T_Area>
-                        void updateBHalf(uint32_t const currentStep)
+                        void updateBFirstHalf(uint32_t const currentStep)
                         {
-                            constexpr auto numWorkers = getNumWorkers();
-                            using Kernel = yeePML::KernelUpdateBHalf<numWorkers, BlockDescription<CurlE>>;
-                            AreaMapper<T_Area> mapper{cellDescription};
-                            /* Note: here it is possible to first check if PML is enabled
-                             * in the local domain at all, and otherwise optimize by calling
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
Could you add a short note to the PR description of what UPML and CPML is? Also which is the curve from the current dev and which is the fixed curve (no need to replot the image).

sbastrakov:
No, they are both for the changed version. These are different sets of PML parameters.

include/picongpu/fields/MaxwellSolver/YeePML/YeePML.hpp
```diff
@@ -362,11 +399,21 @@ namespace picongpu
...
+                     * updates by dt/2 being equal to one update by dt, this is not true for the convolutional field
+                     * updates in PML. Thus, for PML we have to distinguish between the updates by dt/2 by introducing
+                     * first and second halves of the update. This distinction only concerns the convolutional field B
+                     * data used inside the PML, and not the full fields used by the rest of the code. In the very
+                     * first time step of a simulation we start with the second half right away, but this is no
+                     * problem, since the only meaningful initial conditions in the PML area are zero for the
+                     * to-be-absorbed components.
                      */
-                    solver.template updateBHalf<CORE + BORDER>(currentStep);
+                    solver.template updateBSecondHalf<CORE + BORDER>(currentStep);
```
psychocoderHPC, line 416: Why do you not call `updateBHalf`, since this function is part of YeePML there should be no need to hide that there is a bool parameter as second argument.

sbastrakov: I thought to do it, but decided to express more meaning of this parameter via additional function names. Since this is what it means: are we doing the first or second half

sbastrakov: So basically first and second half are terms on the level of the caller here. While the updating of convolutional components is internal terms that I did not want to expose here.

psychocoderHPC (approved):

sbastrakov:
Comparison of the current `dev` version (called `before fix`) to this branch (`after fix`), one plot for UPML and another for CPML. On this plots it is clearly seen that the current `dev` version has a way too large of an original reflection, which is now fixed. Kind of accidentally later on the error for the `dev` branch becomes smaller, but I think this is just due to the fact its unphysical behavior influenced the simulation. The results produced by this PR branch agree well with the Taflove book results.
UPML
CPML

psychocoderHPC (approved):

PR #3487, opened 2021-01-07
Update the FieldAbsorberTest example to match the Taflove book

sbastrakov:
The example previously used the wrong current component.
It also had some smaller discrepancies.

Some changes were found and fixed by @steindev during his work on #3417, so he is a co-author of these changes.


psychocoderHPC (approved):

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/fieldBackground.param
```diff
@@ -84,6 +86,39 @@ namespace picongpu
         /* Add this additional field? */
         static constexpr bool activated = true;
 
+        /* This setup is based on [Taflove, Hagness], section 7.11.1.
+         * The difference is we consider both 2D and 3D cases, and grid size may be increased due to our absorber being
+         * part of the simulation area, not located outside of it as in the book.
+         *
+         * Example of a rectangular conductor with a steady current.
+         *
+         * The conductor is oriented along the an axis.
```
PrometheusPi, line 95: What do you mean by `an` axis?

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/fieldBackground.param
```diff
@@ -84,6 +86,39 @@ namespace picongpu
...
+         * It shall have a diameter of 32 cells.
```
PrometheusPi, line 96: In my opinion, "diameter" is not the right word for a rectangularly shaped wire. How about edge length?

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/fieldBackground.param
```diff
@@ -84,6 +86,39 @@ namespace picongpu
...
+         * the wire
+         *     B = B_0 * (a / s),
+         * where
+         *     B_0 = mu_0 * j_0 * a / 2
+         * is the magnetic field on the surface of the wire.
+         *
+         * The temporal increase of the current I, and therefore its density too, can be regulated similar to an RL
+         * circuit with a coil of inductance L:
+         *     I(t) = I_0 * [1 - e^(- R * t / L)]
+         * https://www.leifiphysik.de/elektrizitaetslehre/elektromagnetische-induktion/ausblick/theorie-zum-einschalten-von-rl-kreisen
```
PrometheusPi, line 110: Is there a similar reference in English?

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/memory.param
```diff
@@ -0,0 +1,101 @@
+/* Copyright 2013-2021 Axel Huebl, Rene Widera, Benjamin Worpitz
```
PrometheusPi, line 1: @sbastrakov Did you apply any changes to this file or is this the default file? If yes you could add yourself as an author - if not, please remove the file.

PrometheusPi (changes requested):
Just some minor questions regarding documentation

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/fieldBackground.param
sbastrakov: Good point, is a result of not fully done text editing

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/fieldBackground.param
sbastrakov: I actually don't know, @steindev do you have an idea?

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/memory.param
sbastrakov: Yes, I had to add this file as I want supercell size to be multiple of 4 but not necessarily 8. Can add myself, but I thought it's not a substantial change

steindev: You should, so the others now whom to ask if there are any questions regarding file changes for this specific setup.

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/fieldBackground.param
steindev: Starting from line 96, the whole documentation is actually outdated and not referring to the Taflove example. This text is left over from a different example I once made and from which I derived for the Taflove example :sweat_smile:
Starting from line 95 to 110, the comment should read
```
The conductor is oriented along the y-axis.
Its edge length can be adjusted by the variable halfWidth in order to apply the test with meaningful results to higher-order solvers, too.

In accordance with Taflove, the current in the wire ramps up over time according to a differentiated Gaussian.
This defines the current density amplitude, too.
Therefore, the total current through the wire scales with the wire's halfWidth.
```
or similar.

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/fieldBackground.param
```diff
@@ -84,6 +86,39 @@ namespace picongpu
...
+         */
+
+        //! Conductor is oriented along y-axis with the following coordinates and size
+        static constexpr int32_t positionX = 30; // unit: cells
```
steindev, line 114: The wire is not centred in the simulation domain. Did you this on purpose?

steindev (changes requested):

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/fieldBackground.param
sbastrakov: Good point. I got mixed between 8-thick PML as default like it used to be and 10 as one of the book setups. I will switch to 10 everywhere.

sbastrakov:
Thanks @PrometheusPi @steindev , I have addressed your comments.

PrometheusPi (dismissed):

PrometheusPi:
@steindev Were all your remarks resolved? If yes, please merge.

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/fieldBackground.param
steindev: For the 8 GPU case, the conductor is still *not* in the center. However, I am not sure if this is of interest for the purpose of the 8 GPU test. What is your intention?

share/picongpu/examples/FieldAbsorberTest/include/picongpu/param/memory.param
```diff
@@ -0,0 +1,101 @@
...
+    constexpr size_t reservedGpuMemorySize = 350 * 1024 * 1024;
+
+    /* short namespace*/
+    namespace mCT = pmacc::math::CT;
+    /** size of a superCell
+     *
+     * volume of a superCell must be <= 1024.
+     * This setup may use grid size that is multiple of 4 but not 8.
+     */
+    using SuperCellSize = typename mCT::shrinkTo<mCT::Int<4, 4, 4>, simDim>::type;
```
steindev, line 50: Question: Can this become a problem for the arbitrary-order FDTD solver using more than 4 neighbours? Should this become at least the number of neighbours the solver uses? (We do not use more than 4 in this particular test, I am just wondering)

psychocoderHPC: This should be fine if the guard size is set to 2 or more: https://github.com/ComputationalRadiationPhysics/picongpu/blob/b17941d2b90afb83b17284cae1b70e62d33e088e/include/picongpu/param/memory.param#L70

sbastrakov: @steindev it is only centered for the 1.cfg configuration, as there is no simple way to have it centered regardless of grid size. I wrote in the .cfg file comments

steindev: But in principle the default `memory.param` should determine the `SuperCellSize` or `GuardSize` on the basis of the number of neighbours used by the solver, right?

steindev: > @steindev it is only centered for the 1.cfg configuration, as there is no simple way to have it centered regardless of grid size. I wrote it in the .cfg file comments

So the intent of the 8 GPU case really is to test whether PML runs on multiple MPI ranks, not whether it produces unexpected results when running on multiple ranks? However, I don't see a way to have a physics test on multiple MPI ranks without a second setup, either. :thinking:

Ohh maybe I do, Set the SupercellSize to `Int<2,2,2>` (if possible) and GuardSize to `Int<2,2,2>` as well. Then you can simulate the 60x60x60 volume with one or eight GPUs and keep the Arbitrary-oder solver working as well, right?

sbastrakov: Good idea @steindev , fixed. Also changed 8.cfg to 4.cfg since the setup also allows 2D case and so to not touch z supercells or MPI ranks

steindev (approved):

PR #3488, opened 2021-01-07
add device side assert macro

pordyna:
Since cuda supports assertion checks ((https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#assertion)), for compute compatibility 2.x and higher, I thought it would be nice to have something similar to `PMACC_ASSERT` that works on device side. So this PR would introduce `PMACC_DEVICE_ASSERT` that basicaly just renames the standard `assert` macro from the `assert.h`. I just added a little compatibility check for older cuda devices. According to the [HIP documentation](https://rocmdocs.amd.com/en/latest/Programming_Guides/HIP-GUIDE.html#assert) asserts are supported in the latest release but maybe we need to check the version  for HIP as well, no idea :sweat_smile: . I tried this out with a dummy assert in `ParticlesInit.kernel` with the `defq_picongpu.profile` and `k80_picongpu.profile` and it aborts and prints the assert message as expected. Maybe someone can check if it works on AMD as well.  What do you think? :smile: @ComputationalRadiationPhysics/picongpu-maintainers 

P.S. For testing I was compiling it without the `-NDEBUG`  flag (but still release target) and with `PMACC_BLOCKING_KERNEL=ON` .

### DIFF
--- a/include/pmacc/assert.hpp
+++ b/include/pmacc/assert.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2016-2021 Rene Widera
+/* Copyright 2016-2021 Rene Widera, Pawel Ordyna
  *
  * This file is part of PMacc.
  *
@@ -24,6 +24,8 @@
 
 #include "pmacc/debug/abortWithError.hpp"
 
+#include <cassert>
+
 #ifdef NDEBUG
 // debug mode is disabled
 
@@ -31,23 +33,30 @@
 #    define PMACC_ASSERT(expr) ((void) 0)
 
 /* `(void)0` force a semicolon after the macro function */
+#    define PMACC_DEVICE_ASSERT(expr) ((void) 0)
+
+/* `(void)0` force a semicolon after the macro function */
 #    define PMACC_ASSERT_MSG(expr, msg) ((void) 0)
+
+// debug mode is disabled
+/* `(void)0` force a semicolon after the macro function */
+#    define PMACC_DEVICE_ASSERT_MSG(expr, ...) ((void) 0)
 
 #else
 
 // debug mode is enabled
 
-/** assert check
+/** assert check (host side only)
  *
- * if `NDEBUG` is not defined: macro expands to (void)0
+ * if `NDEBUG` is defined: macro expands to (void)0
  *
  * @param expr expression to be evaluated
  */
 #    define PMACC_ASSERT(expr) (!!(expr)) ? ((void) 0) : pmacc::abortWithError(#    expr, __FILE__, __LINE__)
 
-/** assert check with message
+/** assert check with message (host side only)
  *
- * if `NDEBUG` is not defined: macro expands to (void)0
+ * if `NDEBUG` is defined: macro expands to (void)0
  *
  * @param expr expression to be evaluated
  * @param msg output message (of type `std::string`) which is printed if the
@@ -55,4 +64,25 @@
  */
 #    define PMACC_ASSERT_MSG(expr, msg) (!!(expr)) ? ((void) 0) : pmacc::abortWithError(#    expr, __FILE__, __LINE__, msg)
 
+/** assert check for kernels (device side)
+ *
+ * if `NDEBUG` is defined: macro expands to (void)0
+ * @param expr expression to be evaluated
+ */
+#    define PMACC_DEVICE_ASSERT(expr) assert(expr)
+
+/** assert check with message (device side)
+ *
+ * if `NDEBUG` is defined: macro expands to (void)0
+ *
+ * Beside the usual assert message an additional message is printed to stdout with `printf`.
+ * Pass your `printf` arguments after the evaluated expression, for example to print some local variables:
+ * @code{.cpp}
+ * PMACC_DEVICE_ASSERT_MSG((x > 0), "x was %e, a was %e", x, a);
+ * @endcode
+ *
+ * @param expr expression to be evaluated
+ * @param ... parameters passed to printf
+ */
+#    define PMACC_DEVICE_ASSERT_MSG(expr, ...) (!!(expr)) ? ((void) 0) : (printf(__VA_ARGS__), assert(expr))
 #endif


include/pmacc/assert.hpp
```diff
@@ -55,4 +59,16 @@
  */
 #    define PMACC_ASSERT_MSG(expr, msg) (!!(expr)) ? ((void) 0) : pmacc::abortWithError(#    expr, __FILE__, __LINE__, msg)
 
+/** assert check
+ *
+ * if `NDEBUG` is not defined: macro expands to (void)0
```
psychocoderHPC, line 64: wrong: if `NDEBUG` is defined the macro expands to (void)0. // see line 28

include/pmacc/assert.hpp
```diff
@@ -55,4 +59,16 @@
...
+ * in kernel asserts are supported only for cuda device compatibility >= 2.
+ * @param expr expression to be evaluated
+ */
+#if((BOOST_LANG_CUDA && PMACC_CUDA_ARCH < 200 ))
+#    define PMACC_DEVICE_ASSERT(expr) ((void) 0)
+#else
+// HIP is supposed to support asserts
```
psychocoderHPC, line 71: This comment can be removed. Yes HIP is supporting assert on device side, with the side effect that you require more registers but you guarded the assert that we do not need more registers if we compile in release mode.

psychocoderHPC (changes requested):

include/pmacc/assert.hpp
```diff
@@ -23,13 +23,17 @@
 #pragma once
 
 #include "pmacc/debug/abortWithError.hpp"
+#include <assert.h>
```
psychocoderHPC, line 26: should be `#include <cassert>`

psychocoderHPC (changes requested):

include/pmacc/assert.hpp
```diff
@@ -55,4 +59,16 @@
  */
 #    define PMACC_ASSERT_MSG(expr, msg) (!!(expr)) ? ((void) 0) : pmacc::abortWithError(#    expr, __FILE__, __LINE__, msg)
 
+/** assert check
```
sbastrakov, line 62: Maybe it's useful to extend a comment saying this one is for in-kernel use, and for the previously existing assert add that it's for out-of-kernel use? Otherwise it maybe unclear if someone just glanceds through this file

include/pmacc/assert.hpp
pordyna: that's true, will change it for the old macros as well.

include/pmacc/assert.hpp
```diff
@@ -24,35 +24,51 @@
...
 #    define PMACC_ASSERT_MSG(expr, msg) (!!(expr)) ? ((void) 0) : pmacc::abortWithError(#    expr, __FILE__, __LINE__, msg)
 
+/** assert check for kernels (device side)
+ *
+ * if `NDEBUG` is defined: macro expands to (void)0
+ * in kernel asserts are supported only for cuda device compatibility >= 2.
+ * @param expr expression to be evaluated
+ */
+#    if((BOOST_LANG_CUDA && PMACC_CUDA_ARCH < 200))
+#        define PMACC_DEVICE_ASSERT(expr) ((void) 0)
```
psychocoderHPC, line 70: You can remove rhe version for architectures smaller than 2.0  PIConGPU is not supporting anything older rhan architecture 2.0.
Please also adjust the comment

psychocoderHPC (changes requested):
Only one change than we can merge this PR

pordyna:
quick question, if we don't need to check cuda version, do we still need this?  `assert` from `assert.h` should already expand to `(void) 0` with `NDEBUG` defined. So this is now really just renaming.

sbastrakov:
@pordyna yes, so for this kind of use just `assert` is fine. The reason to have an own macro is maybe to enable/disable it separately of release/debug mode.

pordyna:
@sbastrakov @psychocoderHPC  Alright, so shouold I just close this PR? Maybe we just need a nice short explanation in the docs instead.

psychocoderHPC (changes requested):

include/pmacc/assert.hpp
```diff
@@ -24,35 +24,46 @@
...
  *            expression is evaluated to false
  */
 #    define PMACC_ASSERT_MSG(expr, msg) (!!(expr)) ? ((void) 0) : pmacc::abortWithError(#    expr, __FILE__, __LINE__, msg)
 
+/** assert check for kernels (device side)
+ *
+ * if `NDEBUG` is defined: macro expands to (void)0
+ * @param expr expression to be evaluated
+ */
+#    define PMACC_DEVICE_ASSERT(expr) assert(expr);
```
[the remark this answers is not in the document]
psychocoderHPC: You can always disable it if `NDEBUG` is defined, no need to check for HIP too.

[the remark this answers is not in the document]
pordyna: https://en.cppreference.com/w/cpp/error/assert:

> 
>  The definition of the macro assert depends on another macro, NDEBUG, which is not defined by the standard library.
> If NDEBUG is defined as a macro name at the point in the source code where <cassert> is included, then assert does nothing. 

And I found this here https://code.woboq.org/userspace/glibc/assert/assert.h.html :
```
#ifdef        NDEBUG

# define assert(expr)                (__ASSERT_VOID_CAST (0))
```

So I think we are save.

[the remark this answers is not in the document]
psychocoderHPC: No you are looking into the glibc implementation. For CUDA and/or HIP the implementation can be different. 
The link I posted it is also saying you should not keep asserts in your code.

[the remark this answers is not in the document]
psychocoderHPC: @pordyna ping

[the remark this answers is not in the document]
pordyna: @psychocoderHPC Oh, sorry. I don't know where to check it for  HIP.   CUDA docs say: 
> Assertions are for debugging purposes. They can affect performance and it is therefore recommended to disable them in production code. They can be disabled at compile time by defining the NDEBUG preprocessor macro before including assert.h

As you say, the HIP documentation says that one should comment it out. So maybe we should keep this new macro.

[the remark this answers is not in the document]
psychocoderHPC: This is the implementation in HIP 3.8

```C++
// TODO-HCC add a dummy implementation of assert, need to replace with a proper kernel exit call.
#if defined(__HCC__) && __HIP_DEVICE_COMPILE__ == 1
#undef assert
#define assert(COND)                                                                               \
    {                                                                                              \
        if (!(COND)) {                                                                             \
            abort();                                                                               \
        }                                                                                          \
    }
#endif
```

Please guard the implementation only for HIP or for all accelerators.

pordyna:
@sbastrakov @psychocoderHPC  So I left the guard for all platforms. Also, since both HIP and CUDA support `printf`, I added something like  `PMACC_ASSERT_MSG`  for device side as well. Please have a look :)

include/pmacc/assert.hpp
```diff
@@ -24,35 +24,65 @@
...
 
+/* `(void)0` force a semicolon after the macro function */
+#    define PMACC_DEVICE_ASSERT(expr) ((void) 0)
+
 /* `(void)0` force a semicolon after the macro function */
 #    define PMACC_ASSERT_MSG(expr, msg) ((void) 0)
 
+// debug mode is disabled
+/* `(void)0` force a semicolon after the macro function */
+#    define PMACC_DEVICE_ASSERT_MSG (expr, ...)((void) 0)
```
psychocoderHPC, line 43: I am wondering that the CI passed because I would say there is a space missing `(expr, ...)((void) 0)` should be `(expr, ...) ((void) 0)`. Do you use `clang-format-11`?

psychocoderHPC: after `PMACC_DEVICE_ASSERT_MSG` there is a space to much

psychocoderHPC (changes requested):

include/pmacc/assert.hpp
psychocoderHPC: For an unknown reason, this part is not wrong formatting. Never the less could you please apply the changes by hand to be consistent with the code around.

psychocoderHPC (dismissed):

pordyna:
should I squish the commits?

psychocoderHPC:
> should I squish the commits?

Yes would be nice. IMO no need to have more than on commit for this small change.

psychocoderHPC (approved):

PR #3490, opened 2021-01-13
HIP: fix memory allocation

psychocoderHPC:
A nonexisting define was used to select if a HIP GPU is used.

The result of this BUG is that on AMD devices only half GPU main memory was used which results in crashes if the problem size of the simulation is too large.

### DIFF
--- a/include/pmacc/nvidia/memory/MemoryInfo.hpp
+++ b/include/pmacc/nvidia/memory/MemoryInfo.hpp
@@ -75,7 +75,7 @@ namespace pmacc
                 /** Returns true if the memory pool is shared by host and device */
                 bool isSharedMemoryPool()
                 {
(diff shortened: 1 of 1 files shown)


sbastrakov (approved):

psychocoderHPC:
Note: using only the half memory main memory gives us a speedup with HIP. I am currently checking if this is an effect coming from the hash function in mallocMC to find free memory slots inside of the heap or if this is a TLB issue.

bussmann:
Interesting. I wanted to ask how this affects our M60 / M100 performance numbers we reported earlier.

psychocoderHPC:
> Interesting. I wanted to ask how this affects our M60 / M100 performance numbers we reported earlier.

No public comment possible.

psychocoderHPC:
@sbastrakov ping

PR #3504, opened 2021-01-27
openPMD support for `plot_chargeConservation.py`

psychocoderHPC:
- add new option `-t` to set the time step which should be loaded
- support openPMD file pattern for series
- fix option names two dash `--` are for long options
- use openPMD-api instead of native hdf5

CI test for compiling are disabled because python is not checked.

### DIFF
--- a/src/tools/bin/plot_chargeConservation.py
+++ b/src/tools/bin/plot_chargeConservation.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # Copyright 2015-2021 Richard Pausch
 #
@@ -20,14 +20,13 @@
 #
 
 import argparse
-import os
 import numpy as np
-import h5py
 import matplotlib.pyplot as plt
+import openpmd_api as io
 
 __doc__ = '''
 This program reads electric field and charge density data
-from hdf5 files created by PIConGPU and checks charge conservation
+from openPMD files created by PIConGPU and checks charge conservation
 for the Yee scheme.
 
 Three slice plots show the error in $div(E) - rho/epsilon_0$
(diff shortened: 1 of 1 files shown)


src/tools/bin/plot_chargeConservation.py
```diff
@@ -67,40 +67,46 @@ def plotError(h5file, slice_pos=[0.5, 0.5, 0.5]):
...
+    CELL_DEPTH = f.get_attribute("cell_depth")
+
+    eField = f.meshes["E"]["x"]
 
     # load electric field
-    Ex = np.array(f["/data/{}/fields/E/x".format(timestep)])
-    Ey = np.array(f["/data/{}/fields/E/y".format(timestep)])
-    Ez = np.array(f["/data/{}/fields/E/z".format(timestep)])
+    Ex = f.meshes["E"]["x"][:]
+    Ey = f.meshes["E"]["x"][:]
```
psychocoderHPC, line 89: must be `y` and next line `z`

src/tools/bin/plot_chargeConservation.py
```diff
@@ -52,55 +51,60 @@ def set_colorbar(cb):
         t.set_fontsize(16)
 
 
-def plotError(h5file, slice_pos=[0.5, 0.5, 0.5]):
+def plotError(file_pattern, slice_pos=[0.5, 0.5, 0.5], timestep=-1):
     """
-    read field data from hdf5 files
+    read field data from a openPMD file
```
sbastrakov, line 56: `an`, not `a`

src/tools/bin/plot_chargeConservation.py
```diff
@@ -52,55 +51,60 @@ def set_colorbar(cb):
...
     compute div(E) - rho/epsilon_0
     plot slices through simulation volume
 
     Parameters:
-    h5file: file name
-        file name to hdf5 data set from PIConGPU
+    file_pattern: file name
+         openPMD file series pattern e.g. simData_%%T.bp
```
sbastrakov, line 63: Add new parameter `timestep` here?

src/tools/bin/plot_chargeConservation.py
```diff
@@ -52,55 +51,60 @@ def set_colorbar(cb):
...
+    series.flush()
 
     # load and add charge density
     charge = np.zeros_like(Ex)
     norm = 0.0
-    for field_name in f["/data/{}/fields/".format(timestep)].keys():
-        if field_name[-14:] == "_chargeDensity":
+
+    for fieldName in f.meshes:
+        if fieldName[-14:] == "_chargeDensity":
```
sbastrakov, line 96: I guess this 14 is length of `_chargeDensity`, maybe better to express like that? So that if the name changes that 14 is not left out

sbastrakov (commented):
Some small comments above. Also please add yourself to the copyright note

src/tools/bin/plot_chargeConservation.py
psychocoderHPC: added

sbastrakov (dismissed):
You still haven't added yourself to both author lists (lines 1 and 35). Otherwise looks good to me

src/tools/bin/plot_chargeConservation.py
```diff
@@ -52,55 +51,65 @@ def set_colorbar(cb):
...
+    # load file
+    series = io.Series(file_pattern, io.Access.read_only)
 
     # read time step (python 2 and 3 save)
-    timestep = -1
-    for i in f['/data'].keys():
-        timestep = i
+    if timestep == -1:
+        for i in series.iterations:
+            timestep = i
```
PrometheusPi, line 78: This loop approach is a python 2.x (outdated) version to get the last iteration of `series.iterations` to get the last iteration. 
Please use:
```python
*_, timestep = series.iterations
```
instead. 

If you want to keep the python 2.x compatibility, I would prefer the slightly cleaner approach:
```python
for timestep in series.iterations:
    continue
```

PrometheusPi (changes requested):
Great work. Only a minor code improvement.

src/tools/bin/plot_chargeConservation.py
psychocoderHPC: will change to the first suggestion, python 2 is dead so there is no need to be compatible.

psychocoderHPC: I applied the change

PrometheusPi (approved):

PR #3508, opened 2021-01-28
fix access to temporary variable

psychocoderHPC:
The returned object from `getGlobalDomain()` and `getLocalDomain()` is accessed inside of PIConGPU over a references which is pointing on a temporary object.

It looks like it is a copy past issue we have since years inside our codebase :-(


sbastrakov (approved):
I think this was fine, not a bug. Binding l-value references to temporary objects prolongs lifetime of those until a reference goes out of scope. I think the fixed version is more clear. But it is equally correct as the old one, and does not constitute a bug fix.

psychocoderHPC:
> I think this was fine, not a bug. Binding l-value references to temporary objects prolongs lifetime of those until a reference goes out

https://blog.galowicz.de/2016/03/23/const_reference_to_temporary_object/ exactly describing what you'd said.
Thanks for pointing me to this behavior, I will remove the bug flag.

PR #3513, opened 2021-02-03
openPMD plugin: Extended JSON patterns

franzpoeschel:
This PR extends the `--openPMD.json` parameter of the openPMD plugin to allow for dataset-specific configuration as well as for loading the JSON configuration from a file.

Loading from a file is done by prepending the argument to `--openPMD.json` with an '@', e.g. `--openPMD.json @myconfig.json`.

Previously, a JSON configuration might look like the following:
```json
{
  "adios2": {
    "engine": {
      "usesteps": true,
      "parameters": {
        "InitialBufferSize": "2Gb",
        "Profile": "On"
      }
    },
    "dataset": {
      "operators": [
      ]
    }
  }
}
```

With this PR, the JSON string may now **additionally** be a list of patterns:
**EDIT: This is outdated, see [this comment](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3513#issuecomment-788820242) for the final style.**
```json
[
  {
    "pattern": [
      "particles/.*/position/.*",
      "particles/.*/positionOffset/.*"
    ],
    "cfg": {
      "adios2": {
        "dataset": {
          "operators": [
            {
              "type": "blosc",
              "parameters": {
                "clevel": "1",
                "doshuffle": "BLOSC_BITSHUFFLE"
              }
            }
          ]
        }
      }
    }
  },
  {
    "pattern": "",
    "cfg": {
      "adios2": {
        "engine": {
          "usesteps": true,
          "parameters": {
            "InitialBufferSize": "2Gb",
            "Profile": "On"
          }
        }
      }
    }
  }
]
```
In here, each single pattern is a JSON object of the following form:
```
{
  "pattern": <pattern>,
  "cfg": <cfg>
}
```
where `<pattern>` is either a string or a list of strings specifying a (currently egrep) regex to match datasets and `<cfg>` is the actual JSON-configuration that will be passed to the dataset.

The global configuration can be set by specifying a pattern that matches the empty string. 

The patterns are matched in given order, picking the first matching pattern. Hence, the catch-all pattern `".*"` should come at the end if it is used.

TODO:
- [x] Testing
- [x] cleanup, documentation
- [x] Use something different than `egrep`? Maybe bash-like globbing?
- [x] Verify that licensing is alright and put credits, this PR adds Nils Lohmann's [JSON library](https://github.com/nlohmann/json) to our internally shipped libraries. Also, I copy-pasted some code from the openPMD API that I wrote over there.
- [x] The JSON library gives a number of warnings, figure out how to quiet them.
    ~~I tried importing the nlohmann_json include directories as system directories, but CMake does not seem to use `-isystem` for `cupla_add_executable/nvcc` targets at all. I think the only way to silence these warnings might be by making the `Json.hpp` file a library of its own..~~
    I've added a `Json.cpp` file now for that purpose, so `nvcc` doesn't ever see the JSON library.
- [x] Compatibility with openPMD 0.12.*
- [x] Update docu
- [x] Dependency graph

Pinging @psychocoderHPC


psychocoderHPC:
> The patterns are matched in given order, picking the first matching pattern. Hence, the catch-all pattern ".*" should come at the end if it is used.

Is it possible to give an empty string to openPMD-api if no pattern matches? I know I suggested that a generic cfg should be in the file but to avoid mistakes we should by default fallback to something valid without a cfg incases were no rule match.

franzpoeschel:
> > The patterns are matched in given order, picking the first matching pattern. Hence, the catch-all pattern ".*" should come at the end if it is used.
> 
> Is it possible to give an empty string to openPMD-api if no pattern matches? I know I suggested that a generic cfg should be in the file but to avoid mistakes we should by default fallback to something valid without a cfg incases were no rule match.

That [already happens](https://github.com/ComputationalRadiationPhysics/picongpu/blob/7c3cd6cef8e580a94a60d0812059b05fd6b849de/include/picongpu/plugins/openPMD/Json.hpp#L195).

psychocoderHPC:
> Use something different than egrep? Maybe bash-like globbing?

IMO egrep is fine and can easily be tested on the command line.

franzpoeschel:
> > Use something different than egrep? Maybe bash-like globbing?
> 
> IMO egrep is fine and can easily be tested on the command line.

Fine with me. We should just be aware, that the catch-all pattern will then be `".*"` and not `"*"`.

psychocoderHPC:
please do not forget to squash the commits at the end to a reasonable number.

psychocoderHPC:
@franzpoeschel I am thinking about changing the format a little bit. Currently, we mixing elements that can be used only once e.g. engine and elements you can give to each dataset. Therefore it is possible to write config files that make no sense because the engine can not be selected per dataset.

I added an attribute to the operator. This would mean you the plugin should
 - search for the backend
 - search for operators with the best fitting pattern -> pass json to openPMD-api

New suggestion:

```json
[
   {
      "cfg":{
         "adios2":{
            "dataset":{
               "engine":{
                  "usesteps":true,
                  "parameters":{
                     "InitialBufferSize":"2Gb",
                     "Profile":"On"
                  }
               },
               "operators":[
                  {
                     "pattern":[
                        "particles/.*/position/.*",
                        "particles/.*/positionOffset/.*"
                     ],
                     "type":"blosc",
                     "parameters":{
                        "clevel":"1",
                        "doshuffle":"BLOSC_BITSHUFFLE"
                     }
                  },
                  {
                     "pattern":[
                        "fields/.*"
                     ],
                     "type":"zstd"
                  }
               ]
            }
         }
      }
   }
]
```

@franzpoeschel what do you think?

franzpoeschel:
I think, both are somewhat valid. The new suggestion has the advantage to repeat information at a more fine-grained level and the disadvantage that it makes this kind of dataset-specific configuration backend-specific (we will soon implement JSON options for HDF5 too).
If we go with the new suggestion, PIConGPU is no longer really the right place to implement this, but it should then be implemented directly in openPMD, instead of beginning to implement backend-specific code in PIConGPU.

With the layout that is currently implemented, I imagine it would usually be used like this:
1. First pattern: `{"pattern": "", <config>}`. This pattern matches only for the global configuration and specifies engine configuration and default settings.
2. Other patterns: Dataset-specific override patterns. If a user feels funny, he can put an engine config in here, but it will be ignored.

No matter how we do it, this is also somewhat a prototype for something that we will later likely implement directly in the openPMD API, so I will discuss this with Axel tomorrow.

What about generalizing both approaches? Make it possible to put something like `["EXPAND", <pattern_1>, <pattern_2>]` at any level in the JSON config. 
Disadvantage: Maybe too powerful
Advantage: Same advantage as your new suggestion, but without being backend-specific.

psychocoderHPC:
> I think, both are somewhat valid. The new suggestion has the advantage to repeat information at a more fine-grained level and the disadvantage that it makes this kind of dataset-specific configuration backend-specific (we will soon implement JSON options for HDF5 too). This concerns usage as well as implementation.
> If we go with the new suggestion, PIConGPU is no longer really the right place to implement this, but it should then be implemented directly in openPMD, instead of beginning to implement backend-specific code in PIConGPU.
> 
> With the layout that is currently implemented, I imagine it would usually be used like this:
> 
> 1. First pattern: `{"pattern": "", <config>}`. This pattern matches only for the global configuration and specifies engine configuration and default settings.
> 2. Other patterns: Dataset-specific override patterns. If a user feels funny, he can put an engine config in here, but it will be ignored.
> 
> No matter how we do it, this is also somewhat a prototype for something that we will later likely implement directly in the openPMD API, so I will discuss this with Axel tomorrow.
> 
> What about generalizing both approaches? Make it possible to put something like `["EXPAND", <pattern_1>, <pattern_2>]` at any level in the JSON config.
> Disadvantage: Maybe too powerful
> Advantage: Same advantage as your new suggestion, but without being backend-specific.

Thanks for the feedback. The expanded idea sounds somehow more complicated but we should talk about it in a VC that I do not misinterpret it.

psychocoderHPC:
@franzpoeschel and I discussed this PR again in a VC and came to the concussion to change the format based on his idea:
`dataset` will become an array and will have entries with `{pattern : "", cfg : {}}` Attributes from `engine` and `the selected dataset configuration will be copied together and passed to openPMD-api.

```
[
   {
      "adios2":{
         "engine":{
            "usesteps":true,
            "parameters":{
               "InitialBufferSize":"2Gb",
               "Profile":"On"
            }
         },
         "dataset":[
            {
               "pattern":[
                  "particles/.*/position/.*",
                  "particles/.*/positionOffset/.*"
               ],
               "cfg":{
                  "operators":[
                     {
                        "type":"blosc",
                        "parameters":{
                           "clevel":"1",
                           "doshuffle":"BLOSC_BITSHUFFLE"
                        }
                     }
                  ]
               }
            },
            {
               "pattern":[
                  ""
               ],
               "cfg":{
                  "operators":[
                     {
                        "type":"zstd"
                     }
                  ]
               }
            }
         ]
      }
   }
]
```

psychocoderHPC:
@franzpoeschel After thinking again and again about it. I realized that other backends do not have the keyword `engine` and `dataset`. This is something that we defined as backend-specific key-words. HDF5 had maybe something like `filter` which is maybe only configurable once for the file. 
We choose in the past json as a way to transport information to the backend which we do not put into concepts each backend should implement. If we now define `dataset` as cfg key we bring in concepts and enforce that all backends should understand these concepts.

> First pattern: {"pattern": "", <config>}. This pattern matches only for the global configuration and specifies engine configuration and default settings.

This will avoid that we need to define a general format of the son config the backend hierarchy in the json file.

ax3l:
Also discussed in VC, great idea.
We discussed that using the `dataset` key with pattern selections (better name: `select`) across all backends is a good way to go.

Also, at the dataset level an additional json option with the value inside the `dataset`-`cfg` key would take priority.

franzpoeschel:
The most recent commit is able to read the following config:
```json
{
  "adios2": {
    "unused": "parameter",
    "engine": {
      "usesteps": true,
      "rabimmel": "rambamel",
      "parameters": {
        "InitialBufferSize": "2Gb",
        "Profile": "On"
      }
    },
    "dataset": [
      {
        "pattern": "",
        "cfg": {
          "operators": [
            {
              "type": "blosc",
              "parameters": {
                "clevel": "1",
                "doshuffle": "BLOSC_BITSHUFFLE"
              }
            }
          ]
        }
      },
      {
        "pattern": [
          ".*positionOffset.*",
          ".*particlePatches.*"
        ],
        "cfg": {
          "operators": []
        }
      }
    ]
  }
}
```

Things are still a bit inefficient at the moment since there are unnecessary JSON <-> string conversions since I want to hide the JSON library from nvcc. Fixing that is up next.

psychocoderHPC:
NICE!

Is the first dataset not always used? Because the pattern is empty which is equal to use it for each dataset?

franzpoeschel:
> NICE!
> 
> Is the first dataset cfg not always used? Because the pattern is empty which is equal to use it for each dataset?

I've clarified things just now. The empty string is used for the global configuration. What you refer to is the catch-all pattern `".*"`.

psychocoderHPC:
Do you like to rename `pattern` into `selection` as suggested by @ax3l ?

franzpoeschel:
Yep, will do

ax3l:
I just realized I asked the same question on chat: are we positive on the `""` (empty selection) that this should refer to "all selection unless specified below"? Wouldn't the catch-all `".*"` fit more there?

Just wondering how we make it least-confusing which pattern applies with priority (on the Series level).

franzpoeschel:
You can definitely use `".*"`, but should be aware that this matches everything, meaning:
* This configuration will be chosen for the Series as well as for all datasets
* This configuration should be put at the end of the list, since anything that comes after will never be chosen

Counter-proposal: Use `"/"` for the Series, indicating the root-level.

franzpoeschel:
Discussed in VC: Having a separate value `""` to match a Series against is redundant and the difference between `"select": ".*"` and `"select": ""` is hard to explain and mostly irrelevant to the user. A default configuration for all datasets should be passed in a catch-all pattern `".*"`.
Downside is that this makes the use of a catch-all pattern necessary which we were able to avoid until now. Putting a `".*"` pattern in the wrong position is a thing that users might accidentally do. But top-down pattern-matching with a catch-all pattern at the end is a common enough pattern for this to be acceptable imo. We could guard against `".*"` occurring in any position but the last, throwing an error or a warning if such a case is detected. (Admittedly, the formal language of all possible words can also be expressed in other ways, but I expect that users who know how to use regexes would also know how to do pattern matching correctly.)

psychocoderHPC:
@franzpoeschel Is `select` an optional attribute? If so this would be a catch-all pattern by default.

franzpoeschel:
@psychocoderHPC No, it is mandatory at the moment. So you suggest setting a default configuration by omitting the `select` key?

psychocoderHPC:
> @psychocoderHPC No, it is mandatory at the moment. So you suggest setting a default configuration by omitting the `select` key?

My suggestion is that `selection` is optional and will always be used if **NO** other entry with a `selection` fits. But this could collide if you have a selection.*` in your configuration.

franzpoeschel:
In principle, I still like the idea of having a clear syntax for defaults and not having to rely on regexes for that, since that allows me to give some special treatment to default settings while at the same time avoiding the pitfalls of a catch-all pattern. Imo, the previous solution was confusing because it defined a special magic pattern for this purpose. But if we communicate clearly "specifying no `select` pattern defines the default behavior", that should no longer be an issue.

Also, we would then **not** advertise the use of something like `".*"`, but advertise René's solution as our preferred way to specify default behavior.

@ax3l Would this be an acceptable compromise?

franzpoeschel:
Recent commit reads a config like this:
```json
{
  "adios2": {
    "engine": {
      "usesteps": true,
      "parameters": {
        "InitialBufferSize": "2Gb",
        "Profile": "On"
      }
    },
    "dataset": [
      {
        "cfg": {
          "operators": [
            {
              "type": "blosc",
              "parameters": {
                "clevel": "1",
                "doshuffle": "BLOSC_BITSHUFFLE"
              }
            }
          ]
        }
      },
      {
        "select": [
          ".*positionOffset.*",
          ".*particlePatches.*"
        ],
        "cfg": {
          "operators": []
        }
      }
    ]
  }
}
```
The first pattern omits the `select` key and is chosen as the default. The config is identical in meaning as the one in [this comment](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3513#issuecomment-780661194) (that one is now outdated).

psychocoderHPC:
> The first pattern omits the `select` key and is chosen as the default. The config is identical in meaning as the one in [this comment](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3513#issuecomment-780661194) (that one is now outdated).

IMO this is the cleanest way. Because it is not possible to define the global config with a pattern because no dataset is involved.
I assume later per dataset we write with openPMD-api it is selecting the config where the selection fits and in case there is no selection fitting the global cfg will be used.

franzpoeschel:
> I assume later per dataset we write with openPMD-api it is selecting the config where the selection fits and in case there is no selection fitting the global cfg will be used.

At this point, that's an implementation detail and should not matter to users.
But what happens is:
1. The `Series` is constructed with the default configuration.
2. If a pattern matches on a dataset, it will be chosen, overriding the global configuration.
3. If no pattern matches on a dataset, it will get an empty configuration, telling openPMD to fall back to the default.

ax3l:
Looks good to me! I think we arrived at a clean solution, with `select` being optional and it missing indicating the default :)

psychocoderHPC:
@franzpoeschel and me discussed yesterday that the documentation should be updated to reflect the latest changes, then the PR is ready to merge.
@franzpoeschel do not forget to squasch the changes.

include/picongpu/plugins/openPMD/Json.cpp
```diff
@@ -153,38 +153,52 @@ namespace
         return res;
     }
 
-    void readPattern(std::vector<Pattern>& patterns, nlohmann::json const& object)
+    KindOfConfig readPattern(
+        std::vector<Pattern>& patterns,
```
sbastrakov, line 157: I think `#include <vector>` is missing.

franzpoeschel: That one is in `JSON_private.hpp` (as is the declaration of `readPattern`. Should I add it again?

include/picongpu/plugins/openPMD/Json.cpp
```diff
@@ -153,38 +153,52 @@ namespace
...
-            }
-            else
-            {
-                throw std::runtime_error(errorMsg);
+                else if(pattern.is_array())
+                {
+                    patterns.reserve(pattern.size());
+                    for(size_t i = 0; i < pattern.size(); ++i)
+                    {
+                        patterns.emplace_back(pattern[i].get<std::string>(), cfgShared);
```
sbastrakov: Why is `cfgShared` moved in the first branch, but not in the second?

franzpoeschel: In the first branch, it is only emplaced once. In the second branch, it is emplaced n times, so the shared pointer needs to be copied (which is unfortunately implicit in C++..). This part is the reason why I used shared pointers, to avoid copying the JSON configs too often.

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -143,7 +143,7 @@ namespace picongpu
                 // openPMD
                 __getTransactionEvent().waitForFinished();
                 openPMDSeries = std::unique_ptr<::openPMD::Series>(
-                    new ::openPMD::Series(fullName, at, communicator, jsonMatcher->get("")));
+                    new ::openPMD::Series(fullName, at, communicator, jsonMatcher->getDefault()));
```
sbastrakov: Not a big deal, but perhaps `std::make_unique` would be slightly bettter than `std::unique_ptr<...>(new ...)`

franzpoeschel: That is still a leftover from C++11 times, but I can take the chance to update that now.

include/picongpu/plugins/openPMD/Json.cpp
sbastrakov: Ah, I didn't get this immediately. I think it's fine then.

franzpoeschel:
I have taken the chance to update the dependency graph, but I don't actually think that nlohmann_json should be in there. It is not a dependency that users should have to care about since it is shipped internally.

psychocoderHPC:
> I have taken the chance to update the dependency graph, but I don't actually think that nlohmann_json should be in there. It is not a dependency that users should have to care about since it is shipped internally.

For compatibility reasons, there should be a switch to use the internal version of JSON or an external version.

You can find an example https://github.com/ComputationalRadiationPhysics/picongpu/blob/fa864412e591870d04641d48d5ff58289112f35c/thirdParty/cupla/cuplaConfig.cmake#L112-L122
You find the same functionality in openPMD-api too (the improved way)

include/picongpu/CMakeLists.txt
```diff
@@ -277,6 +277,15 @@ if(openPMD_FOUND)
     if(openPMD_HAVE_ADIOS2 OR openPMD_HAVE_HDF5)
         message(STATUS "Found openPMD: ${openPMD_DIR}")
         add_definitions(-DENABLE_OPENPMD=1)
+
+        # non of these should appear in cmake-gui, so make them internal
+        set(JSON_BuildTests OFF CACHE INTERNAL "")
+        set(JSON_MultipleHeaders OFF CACHE INTERNAL "")
+        set(JSON_ImplicitConversions OFF CACHE INTERNAL "")
+        set(JSON_Install OFF CACHE INTERNAL "")  # only used PRIVATE
+        add_subdirectory(
```
psychocoderHPC, line 286: Could you please add a witch to use an external version wi `find_package` or the internal version and then with `add_subdirectory`

example:
https://github.com/ComputationalRadiationPhysics/picongpu/blob/fa864412e591870d04641d48d5ff58289112f35c/thirdParty/cupla/cuplaConfig.cmake#L112-L122

franzpoeschel: Sure, I can add that

franzpoeschel: Done. I suppose, the CI containers should be updated with a build run that pulls the most recent nlohmann_json?

psychocoderHPC: Possible but not required. This can be shipped in a follow up PR.

INSTALL.rst
```diff
@@ -343,6 +343,10 @@ openPMD API
 
   - ``export CMAKE_PREFIX_PATH="$HOME/lib/openPMD-api:$CMAKE_PREFIX_PATH"``
   - ``export LD_LIBRARY_PATH="$HOME/lib/openPMD-api/lib:$LD_LIBRARY_PATH"``
+- If PIConGPU is built with openPMD output enabled, the JSON library
```
psychocoderHPC, line 346: The indention is wrong.

psychocoderHPC (changes requested):
Could you please squash all your commits except the last (lohmann integration).

For the last commit please run

```
git reset --soft HEAD~1
GIT_AUTHOR_NAME="Third Party" GIT_AUTHOR_EMAIL="picongpu@hzdr.de"  git commit -a
```
This will commit the third-party library as a generic user. This keeps the statistics of the contributors clean.

INSTALL.rst
franzpoeschel: Is it? This is supposed to be a top-level bullet-point, not part of the two above.

psychocoderHPC: ahh ok in that case it's fine.

psychocoderHPC (approved):

psychocoderHPC:
@franzpoeschel Thanks for this beautiful :sunflower:  PR.

PR #3517, opened 2021-02-09
fix segfault

psychocoderHPC:
Pointer in the thermal test is not initialized but deleted in the destructor, this could result in segfaults, e.g. when calling `picongpu --help`

### DIFF
--- a/share/picongpu/examples/ThermalTest/include/picongpu/ThermalTestSimulation.hpp
+++ b/share/picongpu/examples/ThermalTest/include/picongpu/ThermalTestSimulation.hpp
@@ -34,9 +34,6 @@
 #include <pmacc/nvidia/memory/MemoryInfo.hpp>
 #include <pmacc/mappings/kernel/MappingDescription.hpp>
 #include "picongpu/ArgsParser.hpp"
-
-#include <cassert>
-
 #include "picongpu/plugins/PluginController.hpp"
 
 #include <pmacc/cuSTL/container/DeviceBuffer.hpp>
@@ -54,6 +51,9 @@
 #include <pmacc/nvidia/functors/Add.hpp>
 #include <pmacc/cuSTL/algorithm/functor/GetComponent.hpp>
 #include <pmacc/cuSTL/algorithm/functor/Add.hpp>
+
+#include <cassert>
+#include <memory>
 
 namespace picongpu
 {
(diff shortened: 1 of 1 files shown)


share/picongpu/examples/ThermalTest/include/picongpu/ThermalTestSimulation.hpp
```diff
@@ -197,7 +197,7 @@ namespace picongpu
         //   you may like to let the plasma develope/thermalize a little bit
         static constexpr uint32_t firstTimestep = 1024;
 
-        container::HostBuffer<float, 2>* eField_zt[2];
+        container::HostBuffer<float, 2>* eField_zt[2] = {nullptr, nullptr};
```
ax3l, line 200: I think a better solution would be to default-init the `container::HostBuffer` pointer to `nullptr`

ax3l: oh wait, forget what I said. this is a raw pointer to a container here

ax3l, line 200: You could also use a `unique_ptr` here to modernize and tackle the problem.

psychocoderHPC: Yes, you are right, I was lazy because nobody is using the test anyway^^ I already thought about removing it. 
IMO it makes no sense to keep a dead test nobody is ever running. The test will not be lost because we have it since the beginning in the git history.

I will talk tomorrow with the other dev's about removing it.

psychocoderHPC: updated to `unqiue_ptr` but will still discuss this test in the dev meeting.

share/picongpu/examples/ThermalTest/include/picongpu/ThermalTestSimulation.hpp
```diff
@@ -98,8 +98,6 @@ namespace picongpu
 
         virtual ~ThermalTestSimulation()
         {
-            __delete(eField_zt[0]);
-            __delete(eField_zt[1]);
         }
```
sbastrakov, line 101: I think it can be `default`ed now

psychocoderHPC: updated

sbastrakov (approved):

PR #3519, opened 2021-02-10
Template the Amplitude class of the Radiation plug-in.

Anton-Le:
This PR is supposed to be a precursor for an implementation of a mixed-precision computation of emitted radiation.

The changes introduce a template parameter for the Amplitude class allowing one to switch from 64 bit to 32 bit precision and homogenize the usage of the basic data type within the class. Additionally a `precisionCast` of the amplitude is introduced 

Running the `Bunch` example with a single particle shows no difference in the computed amplitudes of radiation, i.e., vanishing absolute error (tested for 128 observation directions, 1024 frequencies and 200 time steps).


include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -24,33 +24,29 @@
...
-#include "picongpu/traits/PICToSplash.hpp"
 #include "picongpu/particles/traits/SpeciesEligibleForSolver.hpp"
-
-#include "picongpu/plugins/radiation/Radiation.kernel"
-#include "picongpu/plugins/radiation/ExecuteParticleFilter.hpp"
 #include "picongpu/plugins/ISimulationPlugin.hpp"
 #include "picongpu/plugins/common/stringHelpers.hpp"
+#include "picongpu/plugins/radiation/ExecuteParticleFilter.hpp"
+#include "picongpu/plugins/radiation/Radiation.kernel"
+#include "picongpu/simulation_defines.hpp"
```
sbastrakov, line 32: This `"picongpu/simulation_defines.hpp"` file is an exception and when included (rightfully from this file), has to be included first.

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -96,7 +92,7 @@ namespace picongpu
                  * The second dimension is used to store intermediate results if command
                  * line option numJobs is > 1.
                  */
-                GridBuffer<Amplitude, 2>* radiation;
+                GridBuffer<Amplitude<>, 2>* radiation;
```
sbastrakov, line 95: Perhaps there could be a `using` alias for `Amplitude<>`, so that it can be changed in one place if needed? Also, then the using code looks a little cleaner as it is decouple from knowledge that `Amplitude` is a template instantiated with detault arguments. But it's a matter of taste, not important.

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -317,7 +313,9 @@ namespace picongpu
                             numJobs = 1;
                         }
                         // allocate memory for all amplitudes for temporal data collection
-                        tmp_result = new Amplitude[elements_amplitude()];
+
+                        // ACCUMULATOR! Should be in double precision
+                        tmp_result = new Amplitude<>[elements_amplitude()];
```
sbastrakov, line 318: Not related to this PR, but perhaps as a separate refactoring these raw pointers want to become `unique_ptr`s ?

include/picongpu/plugins/radiation/Radiation.kernel
```diff
@@ -21,39 +21,32 @@
...
-#include "picongpu/plugins/radiation/particle.hpp"
 #include "picongpu/plugins/radiation/amplitude.hpp"
 #include "picongpu/plugins/radiation/calc_amplitude.hpp"
+#include "picongpu/plugins/radiation/check_consistency.hpp"
+#include "picongpu/plugins/radiation/nyquist_low_pass.hpp"
+#include "picongpu/plugins/radiation/particle.hpp"
+#include "picongpu/plugins/radiation/radFormFactor.hpp"
 #include "picongpu/plugins/radiation/windowFunctions.hpp"
-#include "picongpu/plugins/radiation/GetRadiationMask.hpp"
+#include "picongpu/simulation_defines.hpp"
```
sbastrakov, line 33: Same here with include order.

include/picongpu/plugins/radiation/amplitude.hpp
```diff
@@ -33,14 +34,16 @@ namespace picongpu
         {
             /** class to store 3 complex numbers for the radiated amplitude
              */
+            template<typename base_dtype = picongpu::float_64>
```
sbastrakov, line 37: We normally call template argument types following `TType` notation

sbastrakov (changes requested):
Nice. Some minor suggestions below.

In addition, I am not sure if this PR should also add a way to externally set accuracy? Right now it's only adding an internal option to change, but seems to use the default template parameter always.

include/picongpu/plugins/radiation/Radiation.hpp
sbastrakov: Nevermids, those are arrays. Maybe `std::vector`s then?

include/picongpu/plugins/radiation/amplitude.hpp
psychocoderHPC: @sbastrakov is right: https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md#16-template-declarations-and-specializations

include/picongpu/plugins/radiation/Radiation.hpp
Anton-Le: I completely agree. In a first attempt I messed up and tried to define a `using Amplitude = Amplitude<>` on a namespace level. As of 539449d  the `using` statement is in the functions and works properly.

include/picongpu/plugins/radiation/amplitude.hpp
Anton-Le: Renamed `base_dtype` to `T_Float` in 539449d .

include/picongpu/plugins/radiation/Radiation.hpp
Anton-Le: I reverted to the same order as `dev` in e99de6a

include/picongpu/plugins/radiation/Radiation.kernel
Anton-Le: Same as in the first case, e99de6a

include/picongpu/plugins/radiation/Radiation.hpp
Anton-Le: I think this should be possible, since the data in `Radiation.hpp` resides on the host.

psychocoderHPC:
@Anton-Le please format the code that the CI tests can run

include/picongpu/plugins/radiation/Radiation.hpp
Anton-Le: `clang-format, using the formatting codification at https://github.com/ComputationalRadiationPhysics/contributing/tree/master/formatting-tools appears to set `simulation_defines.hpp` not as a first file, so that is a necessary manual correction?

Anton-Le:
Upon suggestion replaced the bare pointers for `timeSumArray, tmp_result, detectorFrequencies, detectorPositions` by an `std::vector`. Thi required a change to `writeHDF5file`, `sumAmplitudesOverTime`. Since `writeFile` is used for other data, too it has not been modified, but the vectors simply pass a pointer to the `.data()` to the function.

include/picongpu/plugins/radiation/Radiation.hpp
Anton-Le: See below.

include/picongpu/plugins/radiation/Radiation.hpp
psychocoderHPC: PIConGPU is shipping its own clang-format file https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/.clang-format
There is no manual change for the code style required except using "east coast" `const`.

Please use this file, currently, header reordering is disabled in this file because it will break everything. 
This will be fixed hopefully in the future.

include/picongpu/plugins/radiation/amplitude.hpp
```diff
@@ -92,16 +93,28 @@ namespace picongpu
                 {
                 }
 
+                /** constructor
+                 *
+                 *  @param x
```
psychocoderHPC, line 98: please remove `param ...` or add some text.

psychocoderHPC (changes requested):

include/picongpu/plugins/radiation/Radiation.kernel
```diff
@@ -363,6 +355,7 @@ namespace picongpu
                                 /* storage for amplitude (complex 3D vector)
                                  * it  is initialized with zeros (  0 +  i 0 )
                                  */
+                                // ACCUMULATOR! Ensure double precision
```
psychocoderHPC, line 358: please move this comment into the comment above

include/picongpu/plugins/radiation/Radiation.kernel
```diff
@@ -391,6 +384,7 @@ namespace picongpu
                                             * precisionCast<float_64>(myRadFormFactor(radWeighting_s[j], omega, look));
 
                                         // complex amplitude for j-th particle
+                                        // LOCAL can be single precision
```
psychocoderHPC, line 387: could you please use a multi line comment

include/picongpu/plugins/radiation/amplitude.hpp
```diff
@@ -33,14 +33,16 @@ namespace picongpu
         {
             /** class to store 3 complex numbers for the radiated amplitude
              */
+            template<typename T_Float = picongpu::float_64>
             class Amplitude
             {
             public:
-                using complex_64 = pmacc::math::Complex<picongpu::float_64>;
-
+                // For the intermediate amplitude values we may use single precision
```
psychocoderHPC, line 40: please use a multi line comment `/* */`

include/picongpu/plugins/radiation/amplitude.hpp
```diff
@@ -163,13 +190,54 @@ namespace pmacc
...
+namespace pmacc
+{
+    namespace algorithms
+    {
+        namespace precisionCast
+        {
+            /* We want to be able to cast a low
+             * precision amplitude to a high precision one
+             */
+
```
psychocoderHPC, line 214: please remove the empty line

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -317,7 +310,9 @@ namespace picongpu
                             numJobs = 1;
                         }
                         // allocate memory for all amplitudes for temporal data collection
-                        tmp_result = new Amplitude[elements_amplitude()];
+
+                        // ACCUMULATOR! Should be in double precision
```
psychocoderHPC, line 314: please join with the comment above

psychocoderHPC (changes requested):
only tiny changes required, all other looks good

include/picongpu/plugins/radiation/Radiation.kernel
Anton-Le: Done in 1007070

include/picongpu/plugins/radiation/amplitude.hpp
Anton-Le: I added a description of the parameters as well as of the constructor instead.

psychocoderHPC (approved):

sbastrakov (approved):

PR #3525, opened 2021-02-11
guard ASSERT macros

psychocoderHPC:
When using pmacc assert macros in host/device methods compile can fail.
This PR is adding guards to avoid macro evaluation of device assert macros on host and vice versa.

This will avoid that https://github.com/ComputationalRadiationPhysics/picongpu/blob/be3986413fcdbae7cbff8c6558604c013526e9ac/include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp#L66-L70 will break compiling with hipcc.

### DIFF
--- a/include/pmacc/assert.hpp
+++ b/include/pmacc/assert.hpp
@@ -26,25 +26,16 @@
 
 #include <cassert>
 
-#ifdef NDEBUG
-// debug mode is disabled
+// disabled for no-debug mode or for the device compile path
+#if defined(NDEBUG) || (CUPLA_DEVICE_COMPILE == 1)
 
 /* `(void)0` force a semicolon after the macro function */
 #    define PMACC_ASSERT(expr) ((void) 0)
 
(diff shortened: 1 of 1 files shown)


sbastrakov:
But shouldn't the piece of code you inserted have been using `PMACC_DEVICE_ASSERT` instead to begin with?

psychocoderHPC:
> This code is host device and mostly for device actually, so using host-only assert seems not correct to me.


`PMACC_DEVICE_ASSERT` is not throwing any message where the normal assert is. Maybe it is possible to use the normal ASSERT also on the device and switch for device use automatically to the reduced functionality instead of handling it manually.
Please let us refactor it in an own PR because I like to fix it first to be able to bring HIP into the CI.

sbastrakov (approved):

psychocoderHPC:
I opened an issue for the refactoring: #3526

Issue #3521, opened 2021-02-10
superPusher feature required


BeyondEspresso:
After *offline discussions* with @psychocoderHPC  and @sbastrakov , we decided that we need an extension of the background field implementation, which we dubbed **superPusher**.

In the background field feature of picongpu, the background field (E,B,J) is added to the respective field before the particle-pusher is called and substracted afterwards. This saves memory, but has disadvantages with respect to numerical stability of long simulations, as well as the computation time, because the background fields are computed twice. Since current GPU architectures now have much more memory available per GPU, aforementioned issues can be avoided by storing the background field separately in GPU memory (i.e. compute bgr. field only once per time step and without addition-substraction-cycles). For the field solver only the standard picongpu fields should be made visible, while to all other kernels requiring the fields (especially particle pusher and the picongpu plugins), the sum of standard and background fields should be passed.

psychocoderHPC:
Let's focus on the first step on E and B field, the field for the current is already handled differently than these two.

sbastrakov:
Yes, current is not affected by this really, as it always starts from 0 each time iteration.

sbastrakov:
The PR is merged, closing.

PR #3527, opened 2021-02-11
Add option to duplicate field storage for FieldBackground aka superPusher

sbastrakov:
This is a, so far draft, implementation of another mode of field background operation #3521. The switch between old and new mode is done as a command-line flag `--fieldBackground.duplicateFields`, by default not set which corresponds to old behaviour (since it does not increase memory footprint then).

Implementation is more or less complete, but not properly tested yet. I will proceed with that tomorrow. I can do a technical setup and test that it didn't break the old mode, and the new one works the same, also that it operates fine with checkpoints.

For a real, physically meaningful test, however we do not have any standard example relying on field background. Perhaps to measure performance change and maybe effects on accuracy (although that should be a really long simulation than) you could share one of your setups @BeyondEspresso ?

### DIFF
--- a/include/picongpu/simulation/control/Simulation.hpp
+++ b/include/picongpu/simulation/control/Simulation.hpp
@@ -491,8 +491,6 @@ namespace picongpu
 
                     initialiserController->restart((uint32_t) this->restartStep, this->restartDirectory);
                     step = this->restartStep;
-
-                    fieldBackground.fillSimulation(step);
                 }
                 else
                 {
--- a/include/picongpu/simulation/control/Simulation.hpp
+++ b/include/picongpu/simulation/control/Simulation.hpp
@@ -82,9 +82,6 @@
 
 #include <pmacc/nvidia/reduce/Reduce.hpp>
 #include <pmacc/memory/boxes/DataBoxDim1Access.hpp>
-#include <pmacc/nvidia/functors/Add.hpp>
-#include <pmacc/nvidia/functors/Sub.hpp>
-
 #include <pmacc/meta/conversion/SeqToMap.hpp>
 #include <pmacc/meta/conversion/TypeToPointerPair.hpp>
 
@@ -134,6 +131,7 @@ namespace picongpu
         {
             SimulationHelper<simDim>::pluginRegisterHelp(desc);
             currentInterpolationAndAdditionToEMF.registerHelp(desc);
+            fieldBackground.registerHelp(desc);
             desc.add_options()(
                 "versionOnce",
                 po::value<bool>(&showVersionOnce)->zero_tokens(),
@@ -331,6 +329,10 @@ namespace picongpu
             // create field solver
             this->myFieldSolver = new fields::Solver(*cellDescription);
 
+            // initialize field background stage,
+            // this may include allocation of additional fields so has to be done before particles
+            fieldBackground.init(*cellDescription);
+
             // Initialize random number generator and synchrotron functions, if there are synchrotron or bremsstrahlung
             // Photons
             using AllSynchrotronPhotonsSpecies =
@@ -490,27 +492,7 @@ namespace picongpu
                     initialiserController->restart((uint32_t) this->restartStep, this->restartDirectory);
                     step = this->restartStep;
 
-                    /** restore background fields in GUARD
-                     *
-                     * loads the outer GUARDS of the global domain for absorbing/open boundary condtions
-                     *
-                     * @todo as soon as we add GUARD fields to the checkpoint data, e.g. for PML boundary
-                     *       conditions, this section needs to be removed
-                     */
-                    cellwiseOperation::CellwiseOperation<GUARD> guardBGField(*cellDescription);
-                    namespace nvfct = pmacc::nvidia::functors;
-                    guardBGField(
-                        fieldE,
-                        nvfct::Add(),
-                        FieldBackgroundE(fieldE->getUnit()),
-                        step,
-                        FieldBackgroundE::InfluenceParticlePusher);
-                    guardBGField(
-                        fieldB,
-                        nvfct::Add(),
-                        FieldBackgroundB(fieldB->getUnit()),
-                        step,
-                        FieldBackgroundB::InfluenceParticlePusher);
+                    fieldBackground.fillSimulation(step);
                 }
                 else
                 {
@@ -554,7 +536,7 @@ namespace picongpu
 #endif
             EventTask commEvent;
             ParticlePush{}(currentStep, commEvent);
-            FieldBackground{*cellDescription}(currentStep, nvidia::functors::Sub());
+            fieldBackground.subtract(currentStep);
             myFieldSolver->update_beforeCurrent(currentStep);
             __setTransactionEvent(commEvent);
             CurrentBackground{*cellDescription}(currentStep);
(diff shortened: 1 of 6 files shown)


include/pmacc/memory/buffers/DeviceBufferIntern.hpp
```diff
@@ -311,4 +314,25 @@ namespace pmacc
...
+    /** Factory for a new heap-allocated DeviceBufferIntern buffer object that is a deep copy of the given device
+     * buffer
+     *
+     * @tparam TYPE value type
+     * @tparam DIM index dimensionality
+     *
+     * @param source source device buffer
+     */
+    template<class TYPE, unsigned DIM>
+    HINLINE std::unique_ptr<DeviceBufferIntern<TYPE, DIM>> makeDeepCopy(DeviceBuffer<TYPE, DIM>& source)
```
sbastrakov, line 326: Note: this is logically const, but some methods used internally are mistakenly not const. So I left as a non-const.

psychocoderHPC:
> For a real, physically meaningful test, however we do not have any standard example relying on field background. Perhaps to measure performance change and maybe effects on accuracy (although that should be a really long simulation than) you could share one of your setups @BeyondEspresso ?

I suggest using our current Caar FOM branch to compare the performance.

steindev:
> > For a real, physically meaningful test, however we do not have any standard example relying on field background. Perhaps to measure performance change and maybe effects on accuracy (although that should be a really long simulation than) you could share one of your setups @BeyondEspresso ?
> 
> I suggest using our current Caar FOM branch to compare the performance.

We can do this as a performance test.
For a physics test you can use the SingleParticleTest.
But set the the background electric field component to zero and use the Higuera-Cary pusher, than the electron should turn circles in the magnetic field forever.

Second, please remove the old implementation of the background field.
We know it is unstable and must not be used.
I do not see a point in keeping it, even if it uses less memory.

sbastrakov:
I don't think the old implementation should be removed for two reasons.

- New one consumes more memory, twice as much for E and B if both backgrounds are used So if someone doesn't care about performance or numerical noise, they will be forced to still pay the memory price
- New one had to make an additional assumption that E and B do not change between calls to background, otherwise the copy we saved would have become outdated with no way to know or fix it. This is true currently, but what if there is some other addition to the PIC loop that violates it. Would make this new things mutually exclusive with background if only the new background is kept.

psychocoderHPC:
> New one had to make an additional assumption that E and B do not change between calls to background, otherwise the copy we saved would have become outdated with no way to know or fix it. 

Could you please add to the corresponding place a function that we can trigger a refreshing of the fields. I assume both field background fields are registered ad the data connector so we need only an update function.
The reason is that we are working on ionization back reaction. It would be quite easy to update the stored field (without background field) and then call update afterward to apply the field background again.

sbastrakov:
@psychocoderHPC thanks for your suggestion. I think easier to discuss via mattermost or a VC

psychocoderHPC:
@sbastrakov and I discussed the implementation. The assumption that the fields will not change before we add the current to the EField is fine for now. Our ionization current is not changing the E-field/B-field directly.

include/picongpu/simulation/stage/FieldBackground.hpp
```diff
@@ -38,55 +45,255 @@ namespace picongpu
...
+                        if(!isEnabled)
+                            return;
+                        DataConnector& dc = Environment<>::get().DataConnector();
+                        auto& field = *dc.get<Field>(Field::getName(), true);
+                        // Always add, conditionally make a copy first
+                        if(duplicateField)
+                        {
+                            auto& gridBuffer = field.getGridBuffer();
+                            duplicateBuffer->copyFrom(gridBuffer.getDeviceBuffer());
+                            __getTransactionEvent().waitForFinished();
```
psychocoderHPC, line 104: There is no need to wait for the event system.

include/picongpu/simulation/stage/FieldBackground.hpp
```diff
@@ -38,55 +45,255 @@ namespace picongpu
...
+                        if(!isEnabled)
+                            return;
+                        DataConnector& dc = Environment<>::get().DataConnector();
+                        auto& field = *dc.get<Field>(Field::getName(), true);
+                        // Either restore from the pre-made copy or subtract
+                        if(duplicateField)
+                        {
+                            auto& gridBuffer = field.getGridBuffer();
+                            gridBuffer.getDeviceBuffer().copyFrom(*duplicateBuffer);
+                            __getTransactionEvent().waitForFinished();
```
psychocoderHPC, line 128: There is no need to wait for the event system.

psychocoderHPC (changes requested):

include/picongpu/simulation/control/Simulation.hpp
```diff
@@ -486,27 +488,7 @@ namespace picongpu
...
-                        FieldBackgroundE(fieldE->getUnit()),
-                        step,
-                        FieldBackgroundE::InfluenceParticlePusher);
-                    guardBGField(
-                        fieldB,
-                        nvfct::Add(),
-                        FieldBackgroundB(fieldB->getUnit()),
-                        step,
-                        FieldBackgroundB::InfluenceParticlePusher);
+                    fieldBackground.fillSimulation(step);
```
psychocoderHPC, line 491: I tried to verify that all field including guard/core/border is always in the current time step.
I would say that we break the restart of a simulation with field background enabled.

The plugin output including checkpoint will always contain the fields + background. After a restart we need too remove the field background from the checkpointed data.

@sbastrakov could we discuss it in a VC, the problem that we have the background field included in the checkpoints too will introduce numerical artefacts we tried to avoid with the new method. We should brain storm how we taggle this issue.

psychocoderHPC (changes requested):

include/picongpu/simulation/control/Simulation.hpp
sbastrakov: As discussed in a VC, how checkpoints interact with `Simulation` and data in general may require some reworking to make it more simple. Currently checkpoints incorporated some changes and workarounds, which results in some not very natural things, like field background being added in the moving window check, also with a workaround for when we have just loaded a checkpoint. We discuss how to implement this PR within the existing workflow of checkpointing. And that workflow may be improved later in a separate PR.

sbastrakov: The operation indeed did nothing, as we thought during the VC as it's always overridden by the following field exchange. I have removed it, it also allowed to simplify some core, as it no longer needs to parametrize on area.

With respect to operating right after loading from a checkpoint, this is also fixed now.

include/picongpu/simulation/stage/FieldBackground.hpp
sbastrakov: Fixed

include/picongpu/simulation/stage/FieldBackground.hpp
sbastrakov: Fixed

include/picongpu/simulation/stage/FieldBackground.hpp
```diff
@@ -38,55 +45,214 @@ namespace picongpu
...
+                     * @param step index of time iteration
+                     */
+                    void add(uint32_t const step)
+                    {
+                        if(!isEnabled)
+                            return;
+                        DataConnector& dc = Environment<>::get().DataConnector();
+                        auto& field = *dc.get<Field>(Field::getName(), true);
+                        // Always add to the field, conditionally make a copy of the old values first
+                        if(duplicateField)
```
psychocoderHPC, line 98: please rename `useDublicate`

sbastrakov: Fixed

sbastrakov:
There was a merge conflict due to another PR modifying same lines, resolved it.

psychocoderHPC (approved):

sbastrakov:
@steindev @BeyondEspresso just FYI so now the superPush feature is in `dev`.

PR #3518, opened 2021-02-09
example: fix FieldAbsorberTest

psychocoderHPC:
Using spaces in definitions for the command line or CMake could result in parsing issues.

before this PR `-DPARAM_FIELDSOLVER=ArbitraryOrderFDTDPML< 4 BOOST_PP_COMMA() CurrentInterpolation >` was given to the command line, in some cases the compiler interpreted `4`, `BOOST_PP_COMMA()`, `CurrentInterpolation` and `>` as files because the define is only parsed until the first space.

This issue was the first time seen together with clang++-11 provided by HIP 3.18.

@steindev Do you remember why we used `BOOST_PP_COMMA()` instead of a normal comma? 
Could it be that we used spaces around the comma which produced the issues?

PR #3528, opened 2021-02-12
fix typo in FieldAbsorberTest

psychocoderHPC:
Unixaxial > Uniaxial

Reported by @steindev https://github.com/ComputationalRadiationPhysics/picongpu/pull/3518#discussion_r575060730


steindev (approved):

sbastrakov:
Thanks for fixing. It's actually even more convoluted (pun intended): it's convolutional PML with properties of uniaxial. But the Taflove book sometimes calls it just uniaxial after explaining that, so I decided to also follow this naming.

PR #3535, opened 2021-02-24
Fix some outdated docs regarding output and checkpoint backends

sbastrakov:
This was noticed during a mattermost discussion on checkpoints.

### DIFF
--- a/docs/source/usage/plugins/checkpoint.rst
+++ b/docs/source/usage/plugins/checkpoint.rst
@@ -62,7 +62,6 @@ PIConGPU command line option                  Description
 Depending on the available external dependencies (see above), the options for the ``<IO-backend>`` are:
 
 * :ref:`openPMD <usage-plugins-openPMD>`
-* :ref:`hdf5 <usage-plugins-HDF5>`
 * :ref:`adios <usage-plugins-ADIOS>` (keep in mind the :ref:`note on meta-files <usage-plugins-ADIOS-meta>` for restarts)
 
 Interacting Manually with Checkpoint Data
--- a/docs/source/usage/plugins/checkpoint.rst
+++ b/docs/source/usage/plugins/checkpoint.rst
@@ -21,12 +21,12 @@ What is the format of the created files?
 
 We write our fields and particles in an open markup called :ref:`openPMD <pp-openPMD>`.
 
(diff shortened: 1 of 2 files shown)


psychocoderHPC (dismissed):

sbastrakov:
Sorry, found out a missing `not`, fixed and rebased.

docs/source/usage/plugins/checkpoint.rst
```diff
@@ -62,7 +62,6 @@ PIConGPU command line option                  Description
 Depending on the available external dependencies (see above), the options for the ``<IO-backend>`` are:
```
steindev, line 62: Would you be so kind and remove the double back ticks in line 54 as well?

steindev (changes requested):
Nice catch! Few more spots to be corrected though.

And since you are editing the `TBG_macros.cfg` section of the documentation anyways: I suggest to add a comment to this cfg file, saying that this is really just a collection of snippets and not a working cfg file.
There is the
```bash
"$TBG_cfgPath"/submitAction.sh
```
missing at the end.
I actually stumbled upon this a few days ago, but @psychocoderHPC said it is just supposed to be a collection of snippets.
Although I think we might want to turn it into a full cfg by adding this line, since the file has the `.cfg` ending anyways and thus should be fully functional from my point of view.

sbastrakov:
Thanks @steindev , I've pushed another commit implementing your suggestion and one more thing I've noticed along the way.

psychocoderHPC:
> Nice catch! Few more spots to be corrected though.
> 
> And since you are editing the `TBG_macros.cfg` section of the documentation anyways: I suggest to add a comment to this cfg file, saying that this is really just a collection of snippets and not a working cfg file.
> There is the
> 
> ```shell
> "$TBG_cfgPath"/submitAction.sh
> ```
> 
> missing at the end.
> I actually stumbled upon this a few days ago, but @psychocoderHPC said it is just supposed to be a collection of snippets.
> Although I think we might want to turn it into a full cfg by adding this line, since the file has the `.cfg` ending anyways and thus should be fully functional from my point of view.

We can add it but IMO the file `TBG_macros.cfg` should be deleted and in the docu for each plugin there should a complete example with all options for each plugin that can be copy-pasted by the user.
The reason is that this will provide one point for the user to find all necessary information.

I am fine if you merge this PR first.

steindev:
@sbastrakov added the note to the `cfg`. Which, I think, is sufficient. I still like the idea to have this one place as an overview and example and get additional information via `picongpu --help`.

steindev (approved):

Issue #1100, opened 2015-09-14
cmakeFlags: Document CI Usage Only or Remember


ax3l:
Currently, as noticed by a power user in #1098, `cmakeFlags` are a hidden selection of NxN options for examples that can be played-through in the CI.

We should either clearly document that users shall not use them or alternatively, making things way more complex, the `submitAction.sh` should copy not only the `cmakeFlags` to the run folder (for later `createParameterSets`/`configure` calls) but also the last used `-t <number>` test number that was used with `configure` (if applicable).

I would prefer documenting that `cmakeFlags` are only used by the CI since this does not overcomplicate things.

The work-flow for users is:
- create a fresh case from the example file
- reduce the cmakeFlags to a specific single line if you want a test != 0

n01r:
We can keep it that way. I do not object your reasoning. The flags are just a powerful way for power users to do extensive testing but in the end the simple way is better.

ax3l:
I think the following: copying the `cmakeFlags` does not harm in `submitAction.sh` and I would add a PR for that.

Probably we should not "auto-remember" what the last `-t`. It complicates things and both solutions can be argued to be unexpected ("why does it remember -t if I do not select it?" and "why doesn't it select the same case again?")

n01r:
We can add a `LastConfiguration.txt` that just prints the complete `cmake` command. :stuck_out_tongue:    
> I think the following: copying the cmakeFlags does not harm in submitAction.sh and I would add a PR for that.

Copying the `cmakeFlags` would be fine when still the user directive is to not use them.
Of course while doing the tests I would remember which test case I used because I name my folders that way and not much time passes until I maybe need them again, if I want to correct anything.
> Probably we should not "auto-remember" what the last -t. It complicates things and both solutions can be

Yes, we should not auto-remember. You are only going to use the `cmakeFlags`-feature if you are a developer and not without thinking about it, anyway. The last configuration IS preserved anyway because  certain `PARAM_...`s are overwritten anyway. But when one configures it again without specifying a test case `configure -t 0` is executed effectively and based on which example is used, the previous settings are lost.
But as I said: only developers care about this and they know because they aim for a specific test case flag, anyway IF they choose to compile again.

ax3l:
need for in binary-documentation added to #867 

I think we should avoid writing an additional file, because using `-c` will also overwrite some of the options.

[...]

PR #3537, opened 2021-03-01
Save, for reference, the used cmakeFlags setup.

pordyna:
This introduces a new file `cmakeFlagsSetup` generated by `pic-configure` if there's `cmakeFlags` in the input directory. It stores the setup number used from `cmakeFlags`. `cmakeFlagsSetup` is copied to `<output directory>/input` for reference.

### DIFF
--- a/bin/pic-configure
+++ b/bin/pic-configure
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Copyright 2013-2021 Axel Huebl, Rene Widera
+# Copyright 2013-2021 Axel Huebl, Rene Widera, Pawel Ordyna
 #
 # This file is part of PIConGPU.
 #
@@ -165,6 +165,14 @@ if [ -f "$extension_param/cmakeFlags" ] ; then
         echo "       Is the file executable? (chmod u+x cmakeFlags)" >&2
         exit 2
     fi
+    # save the cmakeFlags setup number for future reference
+    if [ -f "$extension_param/cmakeFlagsSetup" ] ; then
+      rm $extension_param/cmakeFlagsSetup
+    fi
+    touch $extension_param/cmakeFlagsSetup
+    echo "Last configured cmakeFlags setup was the setup No.: $cmakeFlagsNr" > $extension_param/cmakeFlagsSetup
+    echo $cmake_flags >> $extension_param/cmakeFlagsSetup
+
 fi
 
 # legacy check: we removed simulation_defines/ after PIConGPU 0.3.X
--- a/etc/picongpu/submitAction.sh
+++ b/etc/picongpu/submitAction.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Copyright 2013-2021 Axel Huebl, Rene Widera, Felix Schmitt
+# Copyright 2013-2021 Axel Huebl, Rene Widera, Felix Schmitt, Pawel Ordyna
 #
 # This file is part of PIConGPU.
 #
@@ -45,6 +45,10 @@ fi
 if [ -f $TBG_projectPath/cmakeFlags ]
 then
   cp -a $TBG_projectPath/cmakeFlags input
+fi
+if [ -f $TBG_projectPath/cmakeFlagsSetup ]
+then
+  cp -a $TBG_projectPath/cmakeFlagsSetup input
 fi
 cp -a $TBG_cfgPath/openib.conf tbg
 cp -a $TBG_cfgPath/cuda.filter tbg


bin/pic-configure
```diff
@@ -165,6 +165,12 @@ if [ -f "$extension_param/cmakeFlags" ] ; then
         echo "       Is the file executable? (chmod u+x cmakeFlags)" >&2
         exit 2
     fi
+    # save the cmakeFlags setup number for future reference
+    if [ -f "$extension_param/cmakeFlagsSetup" ] ; then
+      rm $extension_param/cmakeFlagsSetup
+    fi
+    touch $extension_param/cmakeFlagsSetup
+    echo "Last configured cmakeFlags setup was: $cmakeFlagsNr" > $extension_param/cmakeFlagsSetup
```
psychocoderHPC, line 173: I would add additionally the content of `cmake_flags` too (on a new line), that `cmakeFlagsSetup` becomes more human-readable.

psychocoderHPC (changes requested):

psychocoderHPC:
IMO in case, there is the file `cmakeFlagsSetup` available and the user is not explicitly specialized `cmakeFlagsNr` the setup of `cmakeFlagsSetup` should be used.

Keep in mind we can call `pic-build` which is calling `pic-configure` within a freshly created setup with `pic-create` and we can call `pic-build` within an existing run where all data is located in `input/`. 

I suggest using the config from `cmakeFlagsSetup` if available except the user used `-t` for `pic-configure` then this should have the higher priority.

steindev:
@pordyna Please have a look at ComputationalRadiationPhysics/picongpu#1100. I think your PR will close this issue.

pordyna:
the discussion in  #1100 is basically saying that we shouldn't have this PR at all. Do we see it differently today? @ComputationalRadiationPhysics/picongpu-maintainers

sbastrakov:
I think this functionality is currently used not just for CI, as was sugested in that issue. So I see no problem with this PR.

n01r:
We definitely need something like this. A simulation run should have unambiguous information about which parameters were used.
In the mid-term future, all relevant parameters to repeat a simulation with the exact same setup should maybe even be copied to our raw particle- and field data.

bin/pic-configure
psychocoderHPC: Please add the content of the cmake_flags too, the number itself is not a good description.

pordyna:
@psychocoderHPC I added the `cmakeFlags` content to `cmakeFlagsSetup`.

About your comment on running with the saved setup as default, we overwrite with the -t option before checking if the directory with `cmakeFlagsSetup` exists.  I'm not sure what to do about it. Maybe someone with more bash experience can change this.

bin/pic-configure
```diff
@@ -165,6 +165,15 @@ if [ -f "$extension_param/cmakeFlags" ] ; then
...
+    echo -e "This was the content of cmakeFlags:\n"
+    cat  $extension_param/cmakeFlags >> $extension_param/cmakeFlagsSetup
```
psychocoderHPC, line 175: Do not copy the file in, copy the setup string only. The file is already available 

should be
```
echo $cmake_flags >> $extension_param/cmakeFlagsSetup
```

psychocoderHPC (changes requested):

psychocoderHPC:
> @psychocoderHPC I added the `cmakeFlags` content to `cmakeFlagsSetup`.
> 
> About your comment on running with the saved setup as default, we overwrite with the -t option before checking if the directory with `cmakeFlagsSetup` exists. I'm not sure what to do about it. Maybe someone with more bash experience can change this.

OK we will skip it. 
Please check my last comment, we need in the new file only the selected string not the full file.

bin/pic-configure
```diff
@@ -165,6 +165,14 @@ if [ -f "$extension_param/cmakeFlags" ] ; then
         echo "       Is the file executable? (chmod u+x cmakeFlags)" >&2
         exit 2
     fi
+    # save the cmakeFlags setup number for future reference
+    if [ -f "$extension_param/cmakeFlagsSetup" ] ; then
+      rm $extension_param/cmakeFlagsSetup
```
steindev, line 170: Just to be sure: The file is only deleted, if the option `-t` is used, isn't it?

pordyna: No, when `-t` is not used it would make the same thing for the default setup=0.

bin/pic-configure
```diff
@@ -165,6 +165,14 @@ if [ -f "$extension_param/cmakeFlags" ] ; then
...
+    fi
+    touch $extension_param/cmakeFlagsSetup
+    echo -e "Last configured cmakeFlags setup was the setup No.: $cmakeFlagsNr" > $extension_param/cmakeFlagsSetup
+    grep "^flags\[$cmakeFlagsNr\]" $extension_param/cmakeFlags >> $extension_param/cmakeFlagsSetup
```
psychocoderHPC, line 174: Please use the line I suggested [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3537#discussion_r595154549)

The file `cmakeFlags` is a bash script, there is no need that someone is adding an array you ca find with grep. A generator is also possible or `flags+="new entry"`

The variable `cmake_flags` contains already the content of the selected setup.

psychocoderHPC (changes requested):

psychocoderHPC (approved):

PR #3539, opened 2021-03-05
expose compute current worker multiplier 

psychocoderHPC:
Allow the user to tune the number of worker/threads uses to compute the current.
The worker multiplier is now part of the strategy used to configure the compute current task and kernel behavior.

The default for CUDA devices is changed from `2` to `1` with this PR.
This will result in better performance for all devices with compute architecture 6.0+. 
On Kepler the performance will be slightly reduced.

note: It is not possible, without increasing the compile-time, to switch automatically based on the GPU used between the old and the new default number for the multiplier.

- add new compute current strategies
- update doxygen documentation

This PR will **not** break existing param files. The implementation is backward compatible.

### DIFF
--- a/include/picongpu/fields/currentDeposition/Strategy.def
+++ b/include/picongpu/fields/currentDeposition/Strategy.def
@@ -30,6 +30,25 @@ namespace picongpu
     {
         namespace strategy
         {
+            namespace detail
+            {
+                /** Validate and adjust worker multiplier
+                 *
+                 * @param multiplicator Number used as multiplier to oversubscribe the number of threads for the
+                 * compute current task/kernel.
+                 * @return valid multiplier
+                 */
+                constexpr int validateAndAdjustWorkerMultiplier(int const multiplicator)
+                {
+#if BOOST_COMP_HIP && PIC_COMPUTE_CURRENT_THREAD_LIMITER
+                    // HIP-clang creates wrong results if more threads than particles in a frame will be used
+                    return 1;
+#else
+                    return multiplicator >= 1 ? multiplicator : 1;
+#endif
+                }
+            } // namespace detail
+
             /** Work on strided supercell domains with local caching strategy
              *
              * The current for each particle will be reduced with atomic operations into a supercell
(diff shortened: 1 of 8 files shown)


psychocoderHPC:
If you still run on old Kepler GPUs e.g. K20/K80 use `currentSolver::strategy::CachedSupercells<2>` in `species.param` as the second template parameter for your choosen current solver.

include/picongpu/fields/currentDeposition/Strategy.def
```diff
@@ -40,37 +40,85 @@ namespace picongpu
...
+                static constexpr bool useBlockCache = true;
+                static constexpr bool stridedMapping = true;
+                using BlockReductionOp = nvidia::functors::Atomic<::alpaka::AtomicAdd, ::alpaka::hierarchy::Threads>;
+                using GridReductionOp = nvidia::functors::Add;
+#if BOOST_COMP_HIP && PIC_COMPUTE_CURRENT_THREAD_LIMITER
+                // HIP-clang creates wrong results if more threads than particles in a frame will be used
+                static constexpr int workerMultiplier = 1;
+#else
+                static constexpr int workerMultiplier = T_workerMultiplier >= 1 ? T_workerMultiplier : 1;
+#endif
```
sbastrakov, line 70: Not a big deal. This piece seems to be present in several places, with these `ifdef` guards. Perhaps there can be a separate constexpr function, or a struct with internal int that expresses this logic, and so in this place it's just `static constexpr int workerMultiplier = newFunction<T_workerMultiplier>();` ?

sbastrakov (approved):
Nice! I have only one optional suggestion above.

include/picongpu/fields/currentDeposition/Strategy.def
psychocoderHPC: good idea, I will create a constexpr function for this HIP restriction + range correction.

sbastrakov (approved):

PrometheusPi (approved):

sbastrakov (approved):

PR #3549, opened 2021-03-11
Add installation instructions for ADIOS2 and HDF5 backends of openPMD API

### LINKED ISSUE #3533: Add install instructions for ADIOS2 to docs
Regarding installation of ADIOS2 I found out that we actually forgot to update our instructions page which now only has ADIOS1. Working on fixing it.

_Originally posted by @sbastrakov in https://github.com/ComputationalRadiationPhysics/picongpu/issues/3496#issuecomment-782146131_

franzpoeschel:
Fix #3533

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -330,6 +330,13 @@ openPMD API
 """""""""""
 - 0.12.0+ (bare minimum) / 0.13.0+ (for streaming IO)
 - *Spack*: ``spack install openpmd-api``
+- For usage in PIConGPU, the openPMD API must have been built either with support for ADIOS2 or HDF5.
+  If building the openPMD API from source (see below for that), those dependencies must be built and installed first.
+
+  - For ADIOS2, CMake build instructions can be found in the `official documentation <https://adios2.readthedocs.io/en/latest/setting_up/setting_up.html>`_.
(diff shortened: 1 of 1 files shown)


INSTALL.rst
```diff
@@ -330,6 +330,13 @@ openPMD API
 """""""""""
 - 0.12.0+ (bare minimum) / 0.13.0+ (for streaming IO)
 - *Spack*: ``spack install openpmd-api``
+- For usage in PIConGPU, the openPMD API must have been built either with support for ADIOS2 or HDF5.
```
sbastrakov, line 333: I think this may be mistakenly understood that only one of those is supported. I think it can be rewritten as 
`For usage in PIConGPU, the openPMD API must have been built with support for ADIOS2 or HDF5 (or both)`

INSTALL.rst
```diff
@@ -330,6 +330,13 @@ openPMD API
...
+  If building the openPMD API from source (see below for that), those dependencies must be built and installed first.
```
sbastrakov, line 334: I would slightly adjust it as
`When building the openPMD API from source (described below), these dependencies must be built and installed first.`

sbastrakov (commented):
Thanks! Just two minor text suggestions.

sbastrakov (approved):

PR #3551, opened 2021-03-11
Switch internal handling of current deposition from compile-time to run-time

sbastrakov:
As we discussed on the VC, I split the transition into run-time current interpolation parameters into two PRs.

This PR only does the internal stuff and does *not* change the user interface of PIConGPU.
The only difference is that custom current deposition functors declared in `.param` files are no longer allowed (but I am not aware of such a use case at all).
It made internal changes necessary for a full transition of this parameter towards run-time, to be done in a following PR.

To avoid confusion, I renamed names and template parameter names of old instances of current interpolation into functors (which they were already and remain that way). Also due to a name conflict with .param file naming the new class is called `CurrentInterpolationInfo`, to be renamed into `CurrentInterpolation` in the follow-up PR.

### DIFF
--- a/include/picongpu/fields/FieldJ.tpp
+++ b/include/picongpu/fields/FieldJ.tpp
@@ -1,5 +1,5 @@
 /* Copyright 2013-2021 Axel Huebl, Heiko Burau, Rene Widera, Felix Schmitt,
- *                     Richard Pausch, Benjamin Worpitz
+ *                     Richard Pausch, Benjamin Worpitz, Sergei Bastrakov
  *
  * This file is part of PIConGPU.
  *
@@ -23,6 +23,7 @@
 #include "picongpu/simulation_defines.hpp"
 #include "picongpu/fields/FieldJ.hpp"
 #include "picongpu/fields/FieldJ.kernel"
+#include "picongpu/fields/currentInterpolation/CurrentInterpolation.hpp"
 #include "picongpu/fields/currentDeposition/Deposit.hpp"
 #include "picongpu/particles/traits/GetCurrentSolver.hpp"
 #include "picongpu/traits/GetMargin.hpp"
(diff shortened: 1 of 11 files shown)


include/picongpu/fields/currentInterpolation/CurrentInterpolation.hpp
```diff
@@ -17,6 +17,77 @@
...
+            CurrentInterpolationInfo& operator=(CurrentInterpolationInfo const&) = delete;
+
+        private:
+            CurrentInterpolationInfo() = default;
+            ~CurrentInterpolationInfo() = default;
+        };
+
+    } // namespace currentInterpolation
+
+} // namespace picongpu
```
sbastrakov, line 93: Will fix

include/picongpu/fields/FieldJ.tpp
```diff
@@ -250,8 +244,8 @@ namespace picongpu
         deposit.template execute<T_area, numWorkers>(cellDescription, depositionKernel, solver, jBox, pBox);
     }
 
-    template<uint32_t T_area, class T_CurrentInterpolation>
-    void FieldJ::addCurrentToEMF(T_CurrentInterpolation& myCurrentInterpolation)
+    template<uint32_t T_area, class T_CurrentInterpolationFunctor>
+    void FieldJ::addCurrentToEMF(T_CurrentInterpolationFunctor myCurrentInterpolationFunctor)
```
PrometheusPi, line 248: Why did you remove the by reference `&`?

PrometheusPi (approved):
Just a minor question.

include/picongpu/fields/FieldJ.tpp
sbastrakov: It was a bit unnatural, but not incorrect originally. This parameter is a functor that's not only not modified internally, but is stateless generally. And with the changed version the call changed and it didn't compile with this reference to non-const. So changed to passing by value, whicj was imo better for the original version as well.

PrometheusPi: Thanks for the explanation.

psychocoderHPC (approved):

PR #3552, opened 2021-03-12
Move current interpolation from compile-time parameter to a command-line one

sbastrakov:
Add a new command-line option to set current interpolation.
Due to #3551, internal processing was already runtime.
Move current interpolation-related code to modern namespace and directory structure.

Current interpolation is no longer a template parameter of field solvers.
Adjust all `.param` files and `cmakeFlags` accordingly.

**Note to users**. This PR is a breaking change to interface of `fieldSolver.param` (and perhaps `cmakeFlags` if you used these options). In case you have a custom option there, your setup needs to be (slightly) updated to compile after this PR is merged. Now field solvers are losing their `T_CurrentInterpolation` template parameter. It was typically set in `fieldSolver.param` by first defining a type alias `CurrentInterpolation` based on cmakeFlags, and then submitting it to the field solver. After the change, most solvers (except arbitrary-order) would normally use all template parameters by default. Still `<>` are needed because that's C++ requirement, so now second-order `Yee` solver is `Yee<>`, etc. "None" solver has no template parameters left, so it's just `None`. I've added a comment to mind these brackets.


sbastrakov:
I am confused why the format check failed, on the dev system applying `clang-format` to my code changes nothing.

psychocoderHPC:
> I am confused why the format check failed, on the dev system applying `clang-format` to my code changes nothing.

Do you tried the for loop from the documentation or do you use git clang-format?

sbastrakov:
@psychocoderHPC thanks for the hint. You are right, my changed `.def` files were not formatted. When i just did `git add` and git `clang-format`. However, I am in turn confused why this hapens, my `git config --local clangFormat.extensions` is done according to the docs and so has `def` there.

psychocoderHPC (approved):

sbastrakov:
I know, and it seemed to not work with my modified `.def` file. Okay, maybe I just did something stupid there. I will see how it works next time

PR #3558, opened 2021-03-16
Fix and modernize the generation scheme for MPI communication tags

### LINKED ISSUE #982: MPI tag needs range check
Tags used for MPI communication needs a check that they were not out of range.

`MPI_TAG_UB` defines the upper limit for tags.

see discussion in #958

sbastrakov:
The old scheme was causing some tags to be outside of the valid range #982.
Apply fixes and modernization of the tag generation scheme.
Document the requirements of the new scheme.

The remaining pieces of `picongpu::CommunicationTag` are no longer used for communication.
But they are left to keep the field solver traits, to be fixed in a following PR.

Resolves #982.
@jprotze thanks for reporting the issue. Hopefully this PR solves it, but I haven't tested it fully yet, will continue tomorrow.

### DIFF
--- a/include/picongpu/simulation_types.hpp
+++ b/include/picongpu/simulation_types.hpp
@@ -34,16 +34,14 @@
 
 namespace picongpu
 {
-    //! define all elements which can send and resive
-
+    /** Legacy communication tags
+     *
+     * Now only used for some field traits, to be removed soon
+     */
     enum CommunicationTag
     {
-        NO_COMMUNICATION = 0u,
         FIELD_B = 1u,
-        FIELD_E = 2u,
-        FIELD_J = 3u,
-        FIELD_JRECV = 4u,
-        SPECIES_FIRSTTAG = 42u
+        FIELD_E = 2u
     };
 
 
--- a/include/pmacc/particles/memory/buffers/ParticlesBuffer.hpp
+++ b/include/pmacc/particles/memory/buffers/ParticlesBuffer.hpp
@@ -47,6 +47,7 @@
 #include <boost/mpl/pair.hpp>
 #include "pmacc/particles/ParticleDescription.hpp"
 #include "pmacc/particles/memory/dataTypes/ListPointer.hpp"
+#include "pmacc/traits/GetUniqueTypeId.hpp"
 
 #include <memory>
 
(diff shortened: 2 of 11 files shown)


sbastrakov:
My Hemera test job is still in the queue. Small runs on the dev system worked fine.

sbastrakov:
The test passed, ready for review @psychocoderHPC !

include/picongpu/simulation_types.hpp
```diff
@@ -34,16 +34,14 @@
 
 namespace picongpu
 {
-    //! define all elements which can send and resive
-
+    /** Legacy communication tags
+     *
+     * Now only used for some field traits, to be removed soon
+     */
     enum CommunicationTag
```
psychocoderHPC, line 41: Can we rename this type to something else, it is now only used to get the margins. 
As your new doxygen says it is no MPI communication tag anymore: https://github.com/ComputationalRadiationPhysics/picongpu/pull/3558/files#diff-ef03a551bc4457d540a9f7b8ee58246c94e3014958928ac33da6fd353a3c0d1dR82

It is something like an alias for the buffer/field. Maybe we should name it `fieldAlias`?

include/pmacc/particles/memory/buffers/ParticlesBuffer.hpp
```diff
@@ -196,12 +197,16 @@ namespace pmacc
...
-                true,
-                false);
+            /* Generate a new tag from this type
+             *
+             * The tag is the same each time this method is called (per instantiation of this template).
+             * Here it is fine, as there is the only instance object, and
+             * exchangeMemoryIndexer->addExchangeBuffer() requires the same tag on each call
+             */
+            using This = ParticlesBuffer<T_ParticleDescription, SuperCellSize_, T_DeviceHeap, DIM>;
+            auto const newTag = traits::GetUniqueTypeId<This, uint32_t>::uid();
```
psychocoderHPC, line 207: It should be possible to write `auto const newTag = traits::GetUniqueTypeId<ParticlesBuffer, uint32_t>::uid();`

include/pmacc/traits/GetUniqueTypeId.hpp
```diff
@@ -73,6 +73,8 @@ namespace pmacc
         /** Get next available type id
          *
          * Warning: is not thread-safe.
+         * Warning: when using it to generate matching tags from multiple MPI ranks, ensure the same call order.
```
psychocoderHPC, line 76: could you please use: `@warning` for this and the line above.

psychocoderHPC (changes requested):

include/picongpu/simulation_types.hpp
sbastrakov: I will remove it in the next PR after this is merged

include/pmacc/particles/memory/buffers/ParticlesBuffer.hpp
sbastrakov: Ah, yes, I am always forgetting it

include/picongpu/simulation_types.hpp
psychocoderHPC: I am fine to move it into a follow up PR

sbastrakov:
Thanks @psychocoderHPC . I've pushed a fix to the two latter points as `--amend`. The first one will be fixed in a follow-up PR by removing this `enum` alltogether - I didn't want to make this unrelated change as part of this PR.

include/pmacc/traits/GetUniqueTypeId.hpp
```diff
@@ -72,7 +72,9 @@ namespace pmacc
 
         /** Get next available type id
          *
-         * Warning: is not thread-safe.
+         * \warning is not thread-safe.
```
psychocoderHPC, line 75: Note please use next time `@warning` see https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md#5-doxygen-comments

psychocoderHPC (approved):

PR #3559, opened 2021-03-17
ParticleCalorimeter in openPMD

franzpoeschel:
#3357

TODO
- [x] Rewrite normal calorimeter output in openPMD
- [ ] Set some of the attributes correctly. For some of them I'll need the help of a plugin user, I marked them with `@todo` in the code.
- [ ] Remove the libsplash output, currently both are written.
- [ ] Checkpointing. How do I trigger a calorimeter checkpoint? I don't see a command line option for that.
- [ ] More testing.
- [ ] The Bremsstrahlung example has a Python analysis script. I suppose that that one should be updated too.


franzpoeschel:
Wrong base

PR #3560, opened 2021-03-17
ParticleCalorimeter in openPMD

franzpoeschel:
#3357

TODO
- [x] Rewrite normal calorimeter output in openPMD
- [x] Verify that attributes are set correctly.
- [x] Remove the libsplash output, currently both are written.
- [x] Checkpointing. How do I trigger a calorimeter checkpoint? I don't see a command line option for that.
    Update: I figured it out, apparently by activating checkpoints and the calorimeter at the same time.
- [x] More testing.
- [x] The Bremsstrahlung example has a Python analysis script. I suppose that that one should be updated too.
    `from particle_calorimeter import ParticleCalorimeter` Where does this come from? There's nothing I could find in `lib/python/picongpu`.
    **Update: I'll treat it as dead code and remove those scripts. Feel free to object.**
- [x] Is it right that this thing is non-parallel? I don't see any MPI in the old writing procedures (except for the checkpoints).
- [x] Merge #3570 first, I've rebased this PR on that one so I don't need to rewrite the attribute writing procedures twice.


include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -181,12 +194,30 @@ namespace picongpu
 
             hdf5DataFile.open(filename.str().c_str(), fAttr);
 
+            filename = std::stringstream{};
+            filename << checkpointDirectory << "/" << this->foldername << "/" << filenamePrefix << "_%T."
+                     << filenameExtension;
+            ::openPMD::Series series(filename.str(), ::openPMD::Access::CREATE);
+
             typename PICToSplash<float_X>::type SplashTypeX;
```
PrometheusPi, line 202: just a question: should/will this be replaces in the future or in other words is PICToSplash maintained in the future?

franzpoeschel: This line will go once I remove the libsplash output in this PR, the openPMD replacement is `::openPMD::determineDatatype<float_X>()`.

psychocoderHPC: This will be removed too, I would say we have a coresponding helper for OpenPMD https://github.com/ComputationalRadiationPhysics/picongpu/blob/454a28efe6a6eb0fa1e0e97031ef2569a1ad313b/include/picongpu/traits/PICToOpenPMD.hpp

include/picongpu/plugins/PluginController.hpp
```diff
@@ -205,10 +205,7 @@ namespace picongpu
             plugins::transitionRadiation::TransitionRadiation<bmpl::_1>
 #if(ENABLE_OPENPMD == 1)
             ,
-            plugins::xrayScattering::XrayScattering<bmpl::_1>
-#endif
-#if(ENABLE_HDF5 == 1)
-            ,
+            plugins::xrayScattering::XrayScattering<bmpl::_1>,
             plugins::radiation::Radiation<bmpl::_1>,
```
psychocoderHPC, line 209: This should only be activated if HDF5_ENABLED is enabled.

psychocoderHPC (changes requested):

include/picongpu/plugins/PluginController.hpp
sbastrakov: That is the `Radiation` plugin.

wmles:
> @franzpoeschel If you think it is ready please remove it from `Draft`
> 
> @n01r @wmles It would be nice if you could have a look at this PR and test it.

the branch did not compile with a freshly created paramSet from FoilLCT. Did I mess it up? I believe not. `/home/goethe93/simulations/testcalo/picongpu/include/pmacc/../picongpu/plugins/PluginController.hpp(209): error: namespace "picongpu::plugins::radiation" has no member "Radiation"`

But I created a test case, which I hope should compile just as well as FoilLCT. So as soon as it does I'm ready to test the feature quantitatively (i.e. bin particle output by hand and compare with the calorimeter output. Is this what you meant, Rene?

n01r:
@steindev also wanted to look into it, yesterday, and I gave him some scripts for analyzing the calorimeter. @wmles please interface with him about this, too :)

franzpoeschel:
My bad, I messed up some ifdefs

franzpoeschel:
Argh, there were some other issues, now it should be working.
@psychocoderHPC For some reason, `cmake -DPIC_USE_Splash=OFF` does not deactivate Splash for me?

psychocoderHPC:
> Argh, there were some other issues, now it should be working.
> @psychocoderHPC For some reason, `cmake -DPIC_USE_Splash=OFF` does not deactivate Splash for me?

Ohh I introduced a bug [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dbaddae80b9d781caf8815d04df782849c7e0444/include/picongpu/CMakeLists.txt#L400) therefore disabling is not working if adios1 is available. I will fix it soon.

Never the less the PR should also work if libspash is enabled.

psychocoderHPC:
> Argh, there were some other issues, now it should be working.
> @psychocoderHPC For some reason, `cmake -DPIC_USE_Splash=OFF` does not deactivate Splash for me?

Should be fixed with #3614

franzpoeschel:
> Never the less the PR should also work if libspash is enabled.

Yup, but I used this option to test whether both modes work and apparently I just ran splash-enabled builds twice.

psychocoderHPC:
@franzpoeschel please run https://picongpu.readthedocs.io/en/latest/dev/styleguide.html#format-all-files to format the code correctly.

franzpoeschel:
I'm not entirely sure where the CI compilation errors stem from, but it looks to me a bit like a bug with the include ordering. I tried restoring the old order now, maybe that fixes it.

psychocoderHPC:
@franzpoeschel Could we finish this PR next week?

franzpoeschel:
Next week after tuesday is good @psychocoderHPC

wmles:
I did some run time tests (not with the branch status that Franz force-pushed an hour ago, but the one directly before).
And can confirm, that it compiles, runs, accepts all parameters, and produces output.

A binning of the particle data by hand did not look like the calorimeter plugin output, but that is probably a mistake of mine that I couldn't resolve yet... So let's first look at the distribution in timestep 0 of a target initialized with an isotropic initial temperature of 50keV.

The dependencies on the pitch and yaw look like I would have expected for an isotropic distribution of particles. The dependence on the energy looks maxwellian for the electrons, but I don't understand the protons. I would have expected, that the energy units are always keV - for the initial-energy-functor for protons as well as electrons, and also for the calorimeter bins.
But it seems, the majority of protons are above the upper bound of 100keV?
The energy bins for electrons are logarithmic, those for protons linear, by the way.

e_calo_energy
e_calo_pitch
e_calo_yaw
H_calo_energy
H_calo_pitch
H_calo_yaw

some questions:
 - do you understand the difference between electrons and protons in the energy distribution? And why is the peak of the electron distribution so far from 50 keV?
 - am I supposed to find the meaning of the axes (which one is pitch, yaw, energy) in the metadata of the h5-file? I didn't. Neither in the docs (I found the docs for the old calorimeter plugin only, though).
 - also there is no clear description in the docs what pitch and yaw mean. In particular the order of the two movements is important. Apparently, it's in yaw direction first, then pitch - that fits to the fact, that the histogram is uniformly distributed in yaw coordinate.
 - also, the (old^^) docs say, that the deposited quantity is total energy per bin. That would mean that what I plotted (the simple sum of the data in the bins) is not supposed to be a maxwellian [in the usual form ](https://en.wikipedia.org/wiki/Maxwell%E2%80%93Boltzmann_distribution). I would have to divide the bindata by the energy in the bin - which gives a totally wrong curve. So I suspect, that the deposited quantitiy is particle number after all?

psychocoderHPC:
@n01r could you also have a look too what @wmles wrote. You used the plugin in the past and I think most questions can not be answered by @franzpoeschel because they are plugin related, where @franzpoeschel only changed the data output format without knowledge about plugin insides.

wmles:
> @n01r could you also have a look too what @wmles wrote. You used the plugin in the past and I think most questions can not be answered by @franzpoeschel because they are plugin related, where @franzpoeschel only changed the data output format without knowledge about plugin insides.

Okay, so technically to close this PR we should rather compare the new plugin with the old, and not with what I expect^^. 
The main implementation-related question is then: it's probably supposed to be in the hdf5-metadata, which axes means which physical quantity? I have not found it.

franzpoeschel:
Thanks for testing this out! I will go through your listing and see what parts of this I can address already.

> The main implementation-related question is then: it's probably supposed to be in the hdf5-metadata, which axes means which physical quantity? I have not found it.

The attributes are currently written very generically and need to be adapted to this plugin still, so some further information (such as correct axis labels) is yet to come. I will keep your findings in mind when adapting the metadata.
(Side note: This is going to be a new addition brought by this PR, the old plugin was very frugal with writing metadata, see screenshot below)

Old attributes left, new ones right. Not all of the new ones are correct yet.:
Bildschirmfoto vom 2021-06-02 15-28-21

n01r:
All in all @wmles is right that the comparison should be between the old and the new parameter set, as long as no other changes were made to the example, but no matter how physical the results are.
I used the `Bremsstrahlung` example for my checks back then and compared for instance the `energyHistogram` diagnostic against the calorimeter, taking the sum over the angular directions.

But to address the observations which I also find strange:

> do you understand the difference between electrons and protons in the energy distribution? And why is the peak of the electron distribution so far from 50 keV?

@wmles is the pusher activated in your example? Also, could it be that the functor that adds temperature, actually (weirdly) adds energy separately for each momentum direction?
On the other hand: the calorimeter, like its name suggests, actually collects **energy per bin** in **Joule** (if you multiply by the `UNIT_SI`). You still have my notebooks, with my analysis examples, right? To get a number of particles for each bin you need to divide the quantity inside the bin by its energy value in *Joule*! And if you want to compare to the `energyHistogram` you would still have to equalize the bin widths.

The first part of the ion energy spectrum looks almost Boltzmann-like. Perhaps it is. Definitely increase the maximum energy bin and check again. Either the overflow is is weird or the temperature functor for ions doesn't work correctly.

> also there is no clear description in the docs what pitch and yaw mean. In particular the order of the two movements is important. Apparently, it's in yaw direction first, then pitch - that fits to the fact, that the histogram is uniformly distributed in yaw coordinate.

The picture in our docs is still a bit wrong, as in: the coordinate system there is not right-handed. That needs to be fixed but it could be an extra pull-request.

> also, the (old^^) docs say, that the deposited quantity is total energy per bin. That would mean that what I plotted (the simple sum of the data in the bins) is not supposed to be a maxwellian in the usual form . I would have to divide the bindata by the energy in the bin - which gives a totally wrong curve. So I suspect, that the deposited quantitiy is particle number after all?

It definitely was total energy per bin when I last used it for production - as I said, I compared to `energyHistogram` and had to calculate back to particle number and then it matched exactly.

franzpoeschel:
I've updated attributes to conform with what the particle calorimeter does, and I've also updated the documentation for it. The gridGlobalOffset as well as the gridSpacing are now defined by `[energy bin, pitch, yaw]` (as is usual in PIC, read in reverse for column-based row orientation).
I agree with the slight confusion on units, it seems that the attributes are in keV while the data is in Joule. But that's just the way the old plugin did it..?

n01r:
> But that's just the way the old plugin did it..?

Yea, we might want to put that up for debate. But that's nothing that you need to change now.

franzpoeschel:
> I'm not entirely sure where the CI compilation errors stem from, but it looks to me a bit like a bug with the include ordering. I tried restoring the old order now, maybe that fixes it.

The error seems to have been outside this plugin and was apparently fixed on the dev in the meantime. I've removed the workaround and the CI now runs fine.

franzpoeschel:
Ready for review from my side

franzpoeschel:
Two notes:

1. The comments by @wmles address the general way that the calorimeter plugin works.
Since this PR only reimplements existing functionality without changing anything, fixing the stuff that he noticed is a task for a future PR.
2. We might want to set a custom `geometry` for the meshes defined by this plugin, since the 2/3-dimensional datasets produced by it are not to be interpreted as points in a plain 2/3-dimensional cartesian space, but additional information on binning and logarithmic scales should be respected. This PR does not do it yet since the openPMD API cannot read from such files yet, see [this PR](https://github.com/openPMD/openPMD-api/pull/1011) for a fix to that issue.

psychocoderHPC:
@reviewer Are thery any additiinal points to this PR?

psychocoderHPC (approved):

include/picongpu/plugins/common/openPMDWriteMeta.hpp
```diff
@@ -102,6 +102,13 @@ namespace picongpu
         {
             void operator()(
                 ::openPMD::Series& series,
+                /*
+                 * Sic! Callers must supply the iteration even though this method
+                 * would be able to retrieve it from the series by using currentStep.
+                 * But we don't know if callers are using the Streaming API or not,
+                 * so let's not fell that decision for them.
```
n01r, line 109: ~~fell~~ --> *make* that decision

n01r (approved):
Looking good :+1:

A little German false friend in the inline docs but nothing else apart from that.

wmles:
Here are some additions to my comments on the test simulations.
 - the documentation is better then in the old plugin now, also the metadata of the file. 
 - also I had found errors in my analysis of the distribution w.r.t. energy presented above, so some very strange things got less strange. 
 - I did not run the same setup with the old and new version of the code, unfortunately. This was not quite straightforward, since the example built for the new dev is not compatible to my older branch anymore, and changed too rapidly. Probably I should make it work?!
 - so what you see below, is the comparison of different methods to access the data in a sim with the new version of the plugin.
 
 - I had made a mistake computing the density from my logarithmically spaced calorimeter output. So the electron energy distribution looks better:
Bildschirmfoto von 2021-06-23 19-31-45
This is what one would expect from a thermal distribution. For the protons it looks similar, too. 
 - but let's compare the particle count calculated from the calorimeter to those calculated from particle output. It's similar, but not identical:
Bildschirmfoto von 2021-06-23 19-28-29
 - and both do not have the maximum where I would expect it for a temperature initialized to 50keV. But this is another/minor issue. 
 
 -> I do not know what I might have done wrong :( do you have ideas?

n01r:
@wmles the `Bremsstrahlung` example has the calorimeter activated by standard.
Why not take this from the current `dev`, compile it, run it, then check out this PR branch and repeat the same thing?
Both on the same hardware without changing anything else.
Then you could do a direct comparison.

psychocoderHPC:
I will merge this PR now as it is if you find any bugs in the new IO format please open an issue.

PR #3570, opened 2021-03-29
Reimplement MacroParticleCounter in openPMD

franzpoeschel:
#3357 

TODO
- [x] Remove old libsplash output
- [x] Move some common functionality with the openPMD plugin to a common directory
- [x] Attributes: Grid attributes done for now (need to be checked for correctness), Series attributes need doing still
- [ ] Testing by actual users of the plugin


psychocoderHPC:
I suggest handling the point `Move some common functionality with the openPMD plugin to a common directory` in a separate PR.

Grid spacing for openPMD should be `SuperCellSize::toRT() * cellSize` (both objects are pmacc vectors).

franzpoeschel:
> I suggest handling the point `Move some common functionality with the openPMD plugin to a common directory` in a separate PR.

I have already included the file `openPMDVersion.def` from the openPMD plugin and was planning to centralize the routine for defining the global attributes (such as `/software` and `softwareVersion`) which I would otherwise have to copy. So, things that are useful for this PR already mostly, other things can go separately?
> 
> Grid spacing for openPMD should be `SuperCellSize::toRT() * cellSize` (both objects are pmacc vectors).
> The value for each cell is a counter, so the scaling factor to SI is `1.0`.

Thanks for the hints!

franzpoeschel:
> Grid spacing for openPMD should be `SuperCellSize::toRT() * cellSize` (both objects are pmacc vectors).

That part fails with:
```
/home/franzpoeschel/git-repos/picongpu/include/pmacc/../picongpu/plugins/makroParticleCounter/PerSuperCell.hpp(295): error #349: no operator "*" matches these operands                                                                                                                  
            operand types are: pmacc::math::Vector<int, 3, pmacc::math::StandardAccessor, pmacc::math::StandardNavigator, pmacc::math::detail::Vector_components> * picongpu::pmacc_static_const_storage121::cellSize_t
```
I can't figure out what type `cellSize_t` is supposed to be, `grep -R` doesn't give me a single match in the whole source tree. Do you know how to fix this? @psychocoderHPC

franzpoeschel:
Dataset attributes look like this now:
Bildschirmfoto von 2021-03-30 19-18-12

psychocoderHPC:
> cellSize

Your workaround looks good, I am not fully sure why the short way is not working.
Could you please write 

```
auto gridSpacing = // SuperCellSize::toRT() * cellSize;
                []() {
                    auto superCellSize = SuperCellSize::toRT();
                    using float_type = decltype(superCellSize[0] * cellSize.x());
                    std::vector<float_type> spacing;
                        spacing.push_front(superCellSize[d] * cellSize[d]);
                    return spacing;
                }();
            mesh.setGridSpacing(gridSpacing);
```

to cover 2D simulations too.

psychocoderHPC:
> Dataset attributes look like this now:
> Bildschirmfoto von 2021-03-30 19-18-12

All looks good we need to take care that all vectors with 3 components must have two in the case where picongpu is running as 2D simulation. You can switch to 2D by manipulating [dimension.param](https://github.com/ComputationalRadiationPhysics/picongpu/blob/454a28efe6a6eb0fa1e0e97031ef2569a1ad313b/share/picongpu/examples/KelvinHelmholtz/include/picongpu/param/dimension.param#L23)

include/picongpu/plugins/makroParticleCounter/PerSuperCell.hpp
```diff
@@ -218,17 +232,34 @@ namespace picongpu
             /*############ dump data #############################################*/
             const SubGrid<simDim>& subGrid = Environment<simDim>::get().SubGrid();
 
+            DataSpace<simDim> localDomainSize(subGrid.getLocalDomain().size / SuperCellSize::toRT());
+            DataSpace<simDim> localDomainOffset(subGrid.getLocalDomain().offset / SuperCellSize::toRT());
+            DataSpace<simDim> globalDomainSize(subGrid.getGlobalDomain().size / SuperCellSize::toRT());
+
+            ::openPMD::Extent openPmdGlobalDomainExtent(3);
```
psychocoderHPC, line 239: Is it correct to hardcode it to `3`, IMO it should be `simDim`

include/picongpu/plugins/makroParticleCounter/PerSuperCell.hpp
```diff
@@ -257,21 +288,81 @@ namespace picongpu
...
+                const uint32_t numSlides = MovingWindow::getInstance().getSlideCounter(currentStep);
+                globalSlideOffset.y() += numSlides * localDomain.size.y();
+                for(uint32_t d = 0; d < simDim; ++d)
+                {
+                    gridGlobalOffset.at(simDim - 1 - d) = float_64(cellSize[d])
+                        * float_64(movingWindow.globalDimensions.offset[d] + globalSlideOffset[d]);
+                }
+            }
+            dataset.setUnitSI(1.0);
+            dataset.setPosition(std::vector<float>{0.5, 0.5, 0.5});
```
psychocoderHPC, line 325: This should handle `simDim==2` too

psychocoderHPC (changes requested):

franzpoeschel:
> All looks good we need to take care that all vectors with 3 components must have two in the case where picongpu is running as 2D simulation. You can switch to 2D by manipulating [dimension.param](https://github.com/ComputationalRadiationPhysics/picongpu/blob/454a28efe6a6eb0fa1e0e97031ef2569a1ad313b/share/picongpu/examples/KelvinHelmholtz/include/picongpu/param/dimension.param#L23)

Ah, good catch, I'll look over things again and check 2D compatibility.

sbastrakov:
> Could you please write ...
> to cover 2D simulations too.

Doesn't it also revert the order? Original code was normal x-y-z and the proposed change is z-y-x

franzpoeschel:
> > Could you please write ...
> > to cover 2D simulations too.
> 
> Doesn't it also revert the order? Original code was normal x-y-z and the proposed change is z-y-x

openPMD and libsplash handle that differently
I need to revert the order everywhere in order to keep the same output.

franzpoeschel:
If someone wants to test the openPMD output, that part is now finished (feature complete, yet maybe not bug-free).
Still need to remove the libsplash output.

include/picongpu/plugins/common/openPMDAttributes.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+
+        /******************
+         * IMPLEMENTATION *
+         ******************/
+
+        std::vector<std::string> SetMeshAttributes::initAxisLabels()
+        {
+            using vec_t = std::vector<std::string>;
+            return simDim == DIM2 ? vec_t{"y", "x"} : vec_t{"z", "y", "x"};
```
psychocoderHPC, line 80: Note you can simplify the code and remove the declaration above `return simDim == DIM2 ? {"y", "x"} : {"z", "y", "x"};`

psychocoderHPC (changes requested):
IMO you can remove the old HDF5 output.

include/picongpu/plugins/common/openPMDAttributes.hpp
franzpoeschel: That's what I had tried first, but apparently C++ cannot do type inference in this statement:
```
/home/franzpoeschel/git-repos/picongpu/include/pmacc/../picongpu/plugins/common/openPMDAttributes.hpp(79): error #29: expected an expression

/home/franzpoeschel/git-repos/picongpu/include/pmacc/../picongpu/plugins/common/openPMDAttributes.hpp(79): error #53: expected a ":"
```

sbastrakov: It cannot because initializer lists are only for objects. So you could have (but of course should not have) expressed it as
```cpp
return simDim == DIM2 ? std::initializer_list<std::string>({"y", "x"}) : std::initializer_list<std::string>({"z", "y", "x"});
```

include/picongpu/plugins/common/openPMDAttributes.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+            SetMeshAttributes const& operator()(::openPMD::MeshRecordComponent& dataset) const;
+        };
+
+
+        /******************
+         * IMPLEMENTATION *
+         ******************/
+
+        std::vector<std::string> SetMeshAttributes::initAxisLabels()
```
sbastrakov, line 77: Please don't forget `inline`. One day we will go to multiple translation units.

franzpoeschel:
> IMO you can remove the old HDF5 output.

Done

include/picongpu/plugins/common/openPMDAttributes.hpp
franzpoeschel: Done

psychocoderHPC (approved):

PR #3572, opened 2021-03-30
remove usage of `atomicAddNoRet` for HIP 4.1

psychocoderHPC:
`atomicAddNoRet` is deprecated with HIP 4.1.
see: https://github.com/ROCm-Developer-Tools/HIP/commit/019c556c7d5e90c248767b8b4fa897ca08f85616

This PR keeps the usage of `atomicAddNoRet` for older HIP versions. The reason is that I would keep the class `AtomicOpNoRet` for now to test some self-implemented atomics without changing too much code. 
I assume that we will remove `AtomicOpNoRet` in the future.

### DIFF
--- a/include/pmacc/nvidia/atomic.hpp
+++ b/include/pmacc/nvidia/atomic.hpp
@@ -69,7 +69,7 @@ namespace pmacc
                 }
             };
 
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
The performance difference of `atomicAdd` and `atomicAddNoRet` is less than 2% when I test it with the SPEC benchmark.

sbastrakov (approved):

Issue #3554, opened 2021-03-16
Filtered particleToGrid fields


pordyna:
As far as I know, there is currently no option to write out derived fields, like particle density, of a filtered species. This would be useful when one doesn't want to output particle data, to avoid large output files, but needs e.g. separate densities for particles in different particle energy ranges. (That's what I need it for ;) )

I think this functionality could be quickly added and I'm willing to do this. I have got two Ideas how.

 One would be to create a new class that takes it's base class (one of the derived attributes in `picongpu/particles/particleToGrid/derivedAttributes`) and a particle Filter as a template parameter. This class would overwrite the base class operator and return 0.0 (no particle contribution) if the filter fails and call the original operator and return the normal particle contribution if the particle passes the filter. It could be just used instead of the usual `DerivedAttribute`.
 
 Another idea would be to extend all classes in `picongpu/particles/particleToGrid/derivedAttributes` with a filter template parameter defaulting to `all` and sth like `using Filter = T_Filter;` and constructing and using the filter just before [this](https://github.com/ComputationalRadiationPhysics/picongpu/blob/c6743ab1a83ded791f3e38c353df48cc907eabf6/include/picongpu/particles/particleToGrid/ComputeGridValuePerFrame.hpp#L71) line. 
 
 The first one is a bit simpler. In the second one `ComputeGridValuePerFrame` is not completely run if we return after the filter fails. That should be faster, at least if there is no divergence on a warp level, which should often be the case if the filter outcome correlates with position. 
 
 What do you think? Do we want this and what would be the best approach? @ComputationalRadiationPhysics/picongpu-maintainers @psychocoderHPC @sbastrakov @n01r

psychocoderHPC:
Yes, it would be very useful to have a filter for derived fields too, I thought we have it already but looks like we do not have it.
As you wrote it should be implemented with the filters we already have and inside of the kernel used like [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/454a28efe6a6eb0fa1e0e97031ef2569a1ad313b/include/pmacc/particles/operations/ConcatListOfFrames.hpp#L131). Branch differences count be avoided and should not be very critical.

PR #3574, opened 2021-03-30
introduce filtered particleToGrid algorithm

pordyna:
This PR introduces derived fields of filtered species. See #3554.  `CreateEligible_t` used in `fileOutput.param` has now an additional template argument that defaults to `filter::All`. The filter is optional and by default `All` in all other relevant structs and methods (like `FieldTmp::computeValue`) so it shouldn't brake any code that is not using filters.  

I did a quick test with the Kelvin-Helmholtz example and it compiles and works as excepted with the `openPMD` plugin  when I use one of the filters from `particleFilter.param` in `fileOutput.param`.  Maybe someone can verify that it works with `isaac` too?  I've never used it before.

Also I did move  `SpeciesEligibleForSolver` in the `RelativeGlobalDomainPosition` to the `.def` file so I don't have to include the `.hpp` file too early an brake the compilation.

@ComputationalRadiationPhysics/picongpu-maintainers

### DIFF
--- a/include/picongpu/fields/FieldTmp.hpp
+++ b/include/picongpu/fields/FieldTmp.hpp
@@ -1,5 +1,5 @@
 /* Copyright 2013-2021 Axel Huebl, Rene Widera, Richard Pausch,
- *                     Benjamin Worpitz
+ *                     Benjamin Worpitz, Pawel Ordyna
  *
  * This file is part of PIConGPU.
  *
@@ -22,6 +22,7 @@
 
 #include "picongpu/simulation_defines.hpp"
 #include "picongpu/fields/Fields.def"
+#include "picongpu/particles/filter/filter.def"
 
 #include <pmacc/fields/SimulationFieldHelper.hpp>
 #include <pmacc/dataManagement/ISimulationData.hpp>
@@ -152,11 +153,13 @@ namespace picongpu
          *
          * @tparam T_area area to compute currents in
          * @tparam T_Species particle species type
+         * @tparam Filter particle filter used to filter contributing particles
+         *         (default is all particles contribute)
          *
          * @param species particle species
          * @param currentStep index of time iteration
          */
-        template<uint32_t AREA, class FrameSolver, class ParticlesClass>
+        template<uint32_t AREA, class FrameSolver, typename Filter = particles::filter::All, class ParticlesClass>
         HINLINE void computeValue(ParticlesClass& parClass, uint32_t currentStep);
 
         /** Bash particles in a direction.
--- a/include/picongpu/fields/Fields.def
+++ b/include/picongpu/fields/Fields.def
@@ -1,4 +1,4 @@
-/* Copyright 2013-2021 Rene Widera, Axel Huebl
+/* Copyright 2013-2021 Rene Widera, Axel Huebl, Pawel Ordyna
  *
  * This file is part of PIConGPU.
  *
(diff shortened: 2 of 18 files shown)


include/picongpu/fields/FieldTmp.hpp
```diff
@@ -156,7 +157,7 @@ namespace picongpu
          * @param species particle species
          * @param currentStep index of time iteration
          */
-        template<uint32_t AREA, class FrameSolver, class ParticlesClass>
+        template<uint32_t AREA, class FrameSolver, typename Filter = particles::filter::All, class ParticlesClass>
```
sbastrakov, line 160: Please add this parameter (and its type) to doxygen comment above.

include/picongpu/fields/Fields.def
```diff
@@ -43,6 +47,8 @@ namespace picongpu
             std::stringstream str;
             str << T_Species::FrameType::getName();
             str << "_";
+            str << T_Filter::getName();
+            str << "_";
```
sbastrakov, line 51: Is it in the middle for a reason? I have no idea what's better, I guess what's consistent with other namings

include/picongpu/particles/filter/RelativeGlobalDomainPosition.def
```diff
@@ -19,6 +19,9 @@
 
 
 #pragma once
+#include "picongpu/particles/traits/SpeciesEligibleForSolver.hpp"
+
+#include <pmacc/traits/HasIdentifiers.hpp>
```
sbastrakov, line 24: What is the reason for moving the trait imlementation to `.def` file?

Including `.hpp` files into `.def` files is potentially dangerous as some `.hpp` files include whole simulation and that may cause wrong inclusion order. So without a good reason we try to avoid doing it.

pordyna: We already do this in files like `Density.def` in `derivedAttributes`. I  got some compile errors when I tried to include the `filter.hpp`in `FilteredDerivedAttribute.hpp`. Some things where not defined yet.   Btw. no when I look at this  `FilteredDerivedAttribute.hpp` could probably also be `.def`.

include/picongpu/fields/Fields.def
pordyna: I thought that this would be better since the filter applies to the particles and not the field

psychocoderHPC:
Some CI tests are failing: e.g. https://gitlab.com/hzdr/crp/picongpu/-/jobs/1140559439#L1280

pordyna:
> Some CI tests are failing: e.g. https://gitlab.com/hzdr/crp/picongpu/-/jobs/1140559439#L1280

Already on it! This has something to do with that how this PR, for some reason, confuses the compiler so that it can't find traits. Basically you can't use `traits::..` while in `particles` namespace  to access sth in `picongpu::traits` anymore. `traits::` will point to `picongpu::particles::traits` instead.

pordyna:
@psychocoderHPC @sbastrakov the CI has passed now :)

include/picongpu/particles/filter/RelativeGlobalDomainPosition.def
sbastrakov: @psychocoderHPC could you comment on that? I personally try to keep `.def` contents (including what they include) at bare minimum, but in this case I don't think it's wrong to do.

psychocoderHPC: Moving the traits into the `def` file should be OK. As @pordyna wrote it could be a requirement to have it in the def file because this file is using the traits. 
Traits should not pull critical code e.g. solver implementations into the `def` file.

psychocoderHPC (approved):

PR #3577, opened 2021-04-07
Fix Particles<>::exchangeMemorySize

jkelling:
This PR fixes the scaling of the memory size allocated for particle exchange with rank-local  grid size.

The idea in the original code is to keep the size of of the particle exchange buffers proportional to the volume of a surface slices of the local domain with fixed thickness (times `userScalingFactor`). This is implemented correctly for domain faces in 3D, which scale proportional to the size of the in-plane domain sizes. The sizes for edges an corners are not computed correctly making the memory for corner, and to a lesser degree edge, exchanges explode for large local domains.

The correct generalization of this idea is to scale buffer sizes proportional to the domain extents *normal* to the exchange direction. While for a face exchange, the in-plane extents are normal to the exchange direction, there is no extent normal to a corner, so these buffers stay at constant size. This also simplifies the code greatly.

In this PR, `userScalingFactor` is applied to the buffer size parallel to the exchange direction, thereby influencing the thickness of the considered surface slices.

@psychocoderHPC can you please have a look?

### DIFF
--- a/include/picongpu/particles/Particles.tpp
+++ b/include/picongpu/particles/Particles.tpp
@@ -77,8 +77,15 @@ namespace picongpu
         template<typename T_ExchangeMemCfg, typename T_Sfinae = void>
         struct DirScalingFactor
         {
-            //! @return factor to scale the amount of memory for each direction
-            static floatD_64 get()
+            //! @return orthogonal contribution factor to scale the amount of memory for each direction
+            static floatD_64 getOrtho()
+            {
+                return floatD_64::create(1.0);
+            }
+
+            //! @return parallel contribution factor to scale the amount of memory for each direction
+            //! (userScalingFactor)
+            static floatD_64 getPara()
             {
                 return floatD_64::create(1.0);
             }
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
I will have a look into this pr next week after my vacation 

Note: for some reason the ci is failing

jkelling:
> Note: for some reason the ci is failing

This is because I missed the SFINAE handling the case that there is no reference size defined in the example, will fix.

include/picongpu/particles/Particles.tpp
```diff
@@ -94,39 +100,38 @@ namespace picongpu
...
-                    auto scaleDirection = std::ceil(
-                        float_64(localDomSize[dir1]) / float_64(baseLocalCells[dir1]) * float_64(localDomSize[dir2])
-                        / float_64(baseLocalCells[dir2]));
-                    float_64 scalingFactor = scaleDirection * userScalingFactor[d];
-                    // do not scale down
-                    scale[d] = std::max(scalingFactor, 1.0);
-                }
+                    scale[d] = userScalingFactor[d];
 
                 return scale;
```
jkelling, line 128: `pmacc::math::Vector` desperately needs a `Vector(const std::array<T_Type, T_dim>& )`, or does a utility for this purpose exist somewhere?

sbastrakov: Then probably not from an `std::array`, but a template constructor from any indexable type? Would also cover std::vector, etc.

jkelling: `std::vector` does not have a compile-time size, that c'tor would need to work differently. A c'tor from `std::array` could also be `constexpr`, btw.

psychocoderHPC (approved):

psychocoderHPC:
Thanks for fixing this.

@jkelling  I assume you tested this PR with the SPEC benchmark and different volumes to avoid that we resend particles because of a too-small particle exchange buffer.

jkelling:
> @jkelling I assume you tested this PR with the SPEC benchmark and different volumes to avoid that we resend particles because of a too-small particle exchange buffer.

Yes, I did test a range, did not go very high up though. Logically it should work. I will learn if there a re re-sends at very large sizes later. If so, this would be more a matter of adjusting the benchmark parameters though.

PR #3580, opened 2021-04-08
Fix openPMD plugin for non-GPU builds

### LINKED ISSUE #3576: openPMD plugin: Crash with default parameters (dataPreparationStrategy = doubleBuffer) when compiling a hostonly Alpaka backend
I'm not sure if this should be considered a bug, but I ran a little simulation previously with `ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE=ON`. Leaving the default parameters in the openPMD backend yields the error:
```
Unhandled exception of type 'St13runtime_error' with message 'DataConnector: Invalid dataset ID (MallocMCBuffer)', terminating
```
When specifying `--openPMD.dataPreparationStrategy mappedMemory`, the plugin runs fine, so the error is likely to stem from [this method](https://github.com/ComputationalRadiationPhysics/picongpu/blob/a36a5c5424a12ec321b3b8dbb3b77440304cb961/include/picongpu/plugins/openPMD/WriteSpecies.hpp#L128). Precise parameters found [here](https://github.com/openPMD/openPMD-example-datasets/pull/14#issuecomment-814692389).

Two questions:
1. Should the MallocMC thing work when picking a non-CUDA backend?
2. If the answer to (1) is "no", i.e. the double buffer strategy is a CUDA-specific thing, should we automatically use the `mappedMemory` strategy if other Alpaka backends are enabled?

franzpoeschel:
GPU-only code was executed also when building non-GPU Alpaka backends.

Fix #3576

### DIFF
--- a/include/picongpu/plugins/openPMD/WriteSpecies.hpp
+++ b/include/picongpu/plugins/openPMD/WriteSpecies.hpp
@@ -130,8 +130,6 @@ namespace picongpu
                 log<picLog::INPUT_OUTPUT>("openPMD:   (begin) copy particle host (with hierarchy) to "
                                           "host (without hierarchy): %1%")
                     % name;
-                auto mallocMCBuffer
-                    = rp.dc.template get<MallocMCBuffer<DeviceHeap>>(MallocMCBuffer<DeviceHeap>::getName(), true);
 
                 int globalParticleOffset = 0;
                 AreaMapping<CORE + BORDER, MappingDesc> mapper(*(rp.params.cellDescription));
@@ -139,7 +137,10 @@ namespace picongpu
                 pmacc::particles::operations::ConcatListOfFrames<simDim> concatListOfFrames(mapper.getGridDim());
 
 #if(PMACC_CUDA_ENABLED == 1 || ALPAKA_ACC_GPU_HIP_ENABLED == 1)
+                auto mallocMCBuffer
+                    = rp.dc.template get<MallocMCBuffer<DeviceHeap>>(MallocMCBuffer<DeviceHeap>::getName(), true);
                 auto particlesBox = rp.speciesTmp->getHostParticlesBox(mallocMCBuffer->getOffset());
+                rp.dc.releaseData(MallocMCBuffer<DeviceHeap>::getName());
 #else
                 /* This separate code path is only a workaround until
                  * MallocMCBuffer is alpaka compatible.
@@ -167,7 +168,6 @@ namespace picongpu
                     mapper,
                     rp.particleFilter);
 
-                rp.dc.releaseData(MallocMCBuffer<DeviceHeap>::getName());
 
                 /* this costs a little bit of time but writing to external is
                  * slower in general */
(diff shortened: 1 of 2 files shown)


include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -808,6 +808,7 @@ Please pick either of the following:
                     }
                 }
 
+#if(PMACC_CUDA_ENABLED == 1 || ALPAKA_ACC_GPU_HIP_ENABLED == 1)
```
franzpoeschel, line 811: I am not sure if skipping this on non-GPU builds is safe to do. Can someone verify?

sbastrakov: Right now `dc.releaseData()` does nothing, it's literally empty implementation. I think this call can be just removed as well.

franzpoeschel: Ok, I've removed those lines in the latest commit

psychocoderHPC: keeping it inside of the guard is correct. s @sbastrakov said it is currently an empty implementation but should keep in the code to simplify later refactoring where this function gets some functionality.

psychocoderHPC: Ohh Sry I have not read the full diff. 
The implementation is fine: For non-GPU devices the code must be skipped because we do not have a mallocMC buffer.

include/picongpu/plugins/openPMD/WriteSpecies.hpp
```diff
@@ -167,7 +168,6 @@ namespace picongpu
                     mapper,
                     rp.particleFilter);
 
-                rp.dc.releaseData(MallocMCBuffer<DeviceHeap>::getName());
```
psychocoderHPC, line 170: Could you keep this line and guard it for non-GPU devices `#if(PMACC_CUDA_ENABLED == 1 || ALPAKA_ACC_GPU_HIP_ENABLED == 1)`

psychocoderHPC (changes requested):

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -824,9 +825,8 @@ Please pick either of the following:
                     meta::ForEach<FileCheckpointParticles, CopySpeciesToHost<bmpl::_1>> copySpeciesToHost;
                     copySpeciesToHost();
                     lastSpeciesSyncStep = currentStep;
-
-                    dc.releaseData(MallocMCBuffer<DeviceHeap>::getName());
```
psychocoderHPC, line 828: It is better to keep this call and guard it. This will simplify refactoring where this function becomes important.

psychocoderHPC (approved):
After thinking few more mins about the `releaseDate` I came to the point that we should remove it everywhere in the code. Currently, it is not clear how this function should be used when the data is used within an asynchronous running kernel.

PR #3597, opened 2021-04-20
SPEC bechmark: add new configurations

psychocoderHPC:
Add two additional configurations to compile for
  - 32bit and TSC particle shape
  - 64bit and TSC particle shape

The particle exchange buffer memory is increased from 6MiB to 8MiB to avoid that we need to send particles within one simulation step twice.

The default case will not compile for 64bit because of the limited shared memory.
At all the selection should help in the CAAR project to validate the performance of 32bit and 64bit runs.

### DIFF
--- a/share/picongpu/benchmarks/SPEC/include/picongpu/param/memory.param
+++ b/share/picongpu/benchmarks/SPEC/include/picongpu/param/memory.param
@@ -79,7 +79,7 @@ namespace picongpu
         // memory used for a direction
         static constexpr uint32_t BYTES_EXCHANGE_X = 1 * 1024 * 1024; // 4 MiB
         static constexpr uint32_t BYTES_EXCHANGE_Y = 1 * 1024 * 1024; // 1 MiB
-        static constexpr uint32_t BYTES_EXCHANGE_Z = 6 * 1024 * 1024; // 6 MiB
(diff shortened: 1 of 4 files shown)


share/picongpu/benchmarks/SPEC/include/picongpu/param/memory.param
```diff
@@ -79,7 +79,7 @@ namespace picongpu
         // memory used for a direction
         static constexpr uint32_t BYTES_EXCHANGE_X = 1 * 1024 * 1024; // 4 MiB
         static constexpr uint32_t BYTES_EXCHANGE_Y = 1 * 1024 * 1024; // 1 MiB
-        static constexpr uint32_t BYTES_EXCHANGE_Z = 6 * 1024 * 1024; // 6 MiB
+        static constexpr uint32_t BYTES_EXCHANGE_Z = 8 * 1024 * 1024; // 6 MiB
```
sbastrakov, line 82: Comment in this line was not updated

sbastrakov (changes requested):
Just a small fix needed.

sbastrakov (approved):

psychocoderHPC:
CI passed

PrometheusPi:
@Anton-Le are you including this adjustment in your Juwels Booster 64bit runs?

PR #3602, opened 2021-04-22
Update AOFDTD manual section

steindev:
With values of the additional factor appearing in the stability condition
for M=1,..,8 neighbors.


sbastrakov (approved):
Nice!

Not sure if worth, but maybe add a direct instructions for a superficial reader? Like that for AOFDTD the max allowed `dt` is slightly smaller, so adjust by the factor given in the table.

sbastrakov (approved):

PR #3607, opened 2021-04-23
Add iteration start pipeline

sbastrakov:
Allow specifying a sequence of functors to be called at start of each time step. So this is similar to `InitPipeline`, but called repeatedly and not once. It is set in a new input file `iterationStart.param`. The pipeline is empty by default. The way of setting and requirements for functors are exactly the same as in `InitPipeline`, updated comments there as well. (Also note that `InitPipeline` could actually be whatever functors and doesn't have to deal with particles, even though the namespace and comments suggest it)

Note that if we didn't have awkward include order stuff, the whole PIC loop could have been put there externally in theory. Actually, almost all simulation stages already have the same interface as functors in the pipeline. So in far (well, or depends on prioritization) future maybe we could even do that.

Add a new compile test to test and demonstrate a potential use, but with no immediate physically meaningful simulation, hence it's a compile test and not an example.

### DIFF
--- a/include/picongpu/particles/InitFunctors.hpp
+++ b/include/picongpu/particles/InitFunctors.hpp
@@ -48,10 +48,8 @@ namespace picongpu
     {
         /** call a functor
          *
-         * @tparam T_Functor unary lambda functor
-         *                   operator() must take two params
-         *                      - first: storage tuple
-         *                      - second: current time step
+         * @tparam T_Functor unary lambda functor, must be default-constructible and
+         *         operator() must take the current time step as the only parameter
          */
         template<typename T_Functor = bmpl::_1>
         struct CallFunctor
--- a/share/picongpu/tests/compileIterationStart/include/picongpu/param/iterationStart.param
+++ b/share/picongpu/tests/compileIterationStart/include/picongpu/param/iterationStart.param
@@ -1,0 +1,81 @@
+/* Copyright 2013-2021 Rene Widera, Richard Pausch, Sergei Bastrakov
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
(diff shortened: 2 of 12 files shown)


include/picongpu/particles/InitFunctors.hpp
```diff
@@ -48,10 +48,8 @@ namespace picongpu
...
          *
-         * @tparam T_Functor unary lambda functor
-         *                   operator() must take two params
-         *                      - first: storage tuple
-         *                      - second: current time step
+         * @tparam T_Functor unary lambda functor, must be default-constructible and
+         *         operator() must take the current time step as the only parameter
          */
         template<typename T_Functor = bmpl::_1>
         struct CallFunctor
```
sbastrakov, line 55: I think `thisCallFunctor` should not be in `namespace particles`, as it has nothing to do with particles, and used to call general functors. But i'm not sure where to put it, so left here.

psychocoderHPC: Yes, let us move `CallFunctor` out of particles after this PR is merged.

share/picongpu/tests/compileIterationStart/include/picongpu/param/iterationStart.param
```diff
@@ -0,0 +1,81 @@
...
+                particleSrc[picongpu::momentum_],
+                picongpu::traits::attribute::getMass(particleSrc[picongpu::weighting_], particleSrc));
+            constexpr float_X thresholdGamma = 3.0_X;
+            if(gamma >= thresholdGamma)
+            {
+                pmacc::particles::operations::assign(particleDest, particleSrc);
+                particleSrc[picongpu::multiMask_] = 0;
+            }
+            else
+                particleDest[picongpu::multiMask_] = 0;
```
psychocoderHPC, line 58: Note to me: We need the possibility to create particles only if needed. The current implementation requires the same amount of memory for the destination species as the source species is consuming. The reason is that `ManipulateDerive` which is used to execute this functor is a simple copy where we invalidate a particle if the condition is not fulfilled.

psychocoderHPC (approved):

include/picongpu/particles/InitFunctors.hpp
sbastrakov: This change is implemented in #3608.

PR #3609, opened 2021-05-03
Use span-based storeChunk API in openPMD plugin (more memory efficient in ADIOS2)

franzpoeschel:
For measurements see [this comment](https://github.com/openPMD/openPMD-api/pull/901#issuecomment-784193402) on the same PR in the openPMD API.

Rationale: The openPMD plugin writes its field data to [an intermediate buffer](https://github.com/ComputationalRadiationPhysics/picongpu/blob/6aa3290a771388303e8a545582a85842729c458d/include/picongpu/plugins/openPMD/openPMDWriter.def#L101) before storing. The same holds for [particle data](https://github.com/ComputationalRadiationPhysics/picongpu/blob/6aa3290a771388303e8a545582a85842729c458d/include/picongpu/plugins/openPMD/writer/ParticleAttribute.hpp#L90). 
On the other hand, ADIOS2 internally serializes all of the given data into one large buffer before actually performing IO.
This means that the openPMD plugin + ADIOS2 redundantly stages all data twice before writing. The recently-added span-based storeChunk API in openPMD allows users to write directly into the ADIOS2 serialization buffer, while automatically handling a fallback-solution for the other backends. This saves needless allocations and memcopies in ADIOS2 and retains the same performance in HDF5.

The last commit adds compatibility with previous openPMD versions.

TODO
- [ ] Testing

### DIFF
--- a/include/picongpu/plugins/openPMD/openPMDWriter.hpp
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.hpp
@@ -1030,7 +1030,8 @@ Please pick either of the following:
 
                     // ask openPMD to create a buffer for us
                     // in some backends (ADIOS2), this allows avoiding memcopies
-                    auto span = mrc.storeChunk<float_X>(
+                    auto span = storeChunkSpan<float_X>(
+                        mrc,
                         asStandardVector(fieldsOffsetDims),
                         asStandardVector(fieldsSizeDims),
                         [&fieldBuffer](size_t size) {
--- a/include/picongpu/plugins/openPMD/openPMDWriter.hpp
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.hpp
@@ -776,9 +776,9 @@ Please pick either of the following:
 
             void initWrite()
             {
-                // may be zero
-                auto size = mThreadParams.window.localDimensions.size.productOfComponents();
-                mThreadParams.fieldBuffer.resize(size);
+                // fieldBuffer will only be resized if needed
+                // in some openPMD backends, it's more efficient to let the backend handle buffer creation
+                // (span-based RecordComponent::storeChunk() API)
             }
 
             /**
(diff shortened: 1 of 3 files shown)


franzpoeschel:
Alright, I ran two quick benchmarks, both using this PR:
* lefthand side: recent openPMD API dev
* righthand side: openPMD API 0.12.0 (I needed to used file-based iteration layout since that one doesn't use ADIOS2 steps yet, hence the gaps)

The important difference is the little spikes that you see from the intermediate buffers being used in the openPMD plugin on the righthand side. For recent openPMD API versions, this PR has been able to remove those spikes, reducing memory consumption by some 1.7GB.
(Note that I used `--openPMD.dataPreparationStrategy mappedMemory` to further reduce the memory usage of the openPMD plugin and make it more obvious in the plot where memory was saved)

Bildschirmfoto vom 2021-05-05 11-18-41
Bildschirmfoto vom 2021-05-05 11-31-26

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -998,9 +997,51 @@ Please pick either of the following:
                     fieldsOffsetDims[0] = globalOffsetFile;
                 }
 
+                auto const componentSize = field_no_guard.productOfComponents();
+
                 /* write the actual field data */
                 for(uint32_t d = 0; d < nComponents; d++)
                 {
+                    ::openPMD::MeshRecordComponent mrc
```
psychocoderHPC, line 1005: Why is this code moved into the loop? Was it a bug and before the meta data was added to the wrong object?

psychocoderHPC (changes requested):

include/picongpu/plugins/openPMD/openPMDWriter.hpp
franzpoeschel: The diff looks a bit misleading on this one. The code is still at the same level, but now it's in the beginning of the iteration. The braces at 1071/2 close a contained loop.

psychocoderHPC (approved):

PR #3610, opened 2021-05-04
Updated defq profile for hemera/defq

BrianMarre:
updated the current profile for hemera/defq to use the new non cuda modules to allow running picongpu on the cpu only nodes


sbastrakov (approved):
So as I understand the admins now added modules for output that do not require CUDA?

BrianMarre:
Yes ;), It took about 1 week and 2 workdays form initial request to finish and about 3 workdays of actual work.
Note: we still lack an adios1 non cuda module but that should not be that large of a problem.

PR #3612, opened 2021-05-05
Add option for hemera low (gpu) partitions

PrometheusPi:
This pull request adds the option to use the respective low partition instead of the regular (gpu) partition for `k20`, `k80`, `gpu` and `fwkt_v100`. 

**check:**
- [x] verify that `--account default` works for `gpu` partition
- [x] test all partitions

### DIFF
--- a/etc/picongpu/hemera-hzdr/gpu.tpl
+++ b/etc/picongpu/hemera-hzdr/gpu.tpl
@@ -22,6 +22,7 @@
 # PIConGPU batch script for hemera's SLURM batch system
 
 #SBATCH --partition=!TBG_queue
+#SBATCH --account=!TBG_account
 #SBATCH --time=!TBG_wallTime
 # Sets batch job's name
 #SBATCH --job-name=!TBG_jobName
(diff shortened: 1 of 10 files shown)


PrometheusPi:
All jobs get submitted to the correct queue. But I think using the `gpu` with the `default` accounting results in a lower priority than `gpu_low` with `low` accounting. I will contact IT to understand that.

etc/picongpu/hemera-hzdr/gpu.tpl
```diff
@@ -41,7 +42,8 @@
 
 
 ## calculations will be performed by tbg ##
-.TBG_queue="gpu"
+.TBG_queue=${TBG_partition:-"gpu"}
+.TBG_account=`if [ $TBG_partition == "gpu_low" ] ; then echo "low"; else echo "default"; fi`
```
PrometheusPi, line 46: Might be problematic.

PrometheusPi (commented):
Please do not merge (I can not "Request changes" with my own pull request to block merging.)

sbastrakov:
@PrometheusPi you can convert it to Draft to prevent merging.

etc/picongpu/hemera-hzdr/gpu.tpl
PrometheusPi: This is not problematic. The apperent higher priority originated from the fact that `gpu_low` encompass more nodes than `gpu` and thus could schedule my `low` job faster than my `default` one.

PrometheusPi:
Ready for review.

PrometheusPi:
@psychocoderHPC surprisingly the `low` partition might be suited for development if you use the P100 GPUs. The `low`options grants access to 3 mostly not used P100 nodes that might allow faster scheduling despite lower priority.

sbastrakov (approved):

PR #3616, opened 2021-05-17
refactor lockstep programming

psychocoderHPC:
- update and extent documentation
- all functors for lockstep programming support now three parameter interfaces
- introduce the namespace `pmacc::lockstep`
- rename `CtxArray` int `Variable`
  - forbid to access elements via an index
- add new class  `lockstep::Idx` for a single index within a lockstep domain
- introduce factories `lockstep::makeVar<>`, `lockstep::makeForEach<>`
- refactor how a context variable can be direct initialized with a functor

# Motivation

It is possible to use the context variable wrongly by using an arbitrary integer index. By adjusting the interface the wrong usage can be avoided.
Before this PR we passed always two parameter to the lambda function but mostly used only one and seldomly required no parameter. The new changes reflect the most used interfaces, this was already discussed internally a long time ago.

# Review

The first commit contains the required changes to follow the new interfaces. 
The second commit contains the updated documentation and the changes of the lockstep core classes.
The execution of the for each loop has not be changed except that the domain with one element will be executed a conditional if statement. A refactoring of the for each loop will be part of a follow up PR.

# Runtime differences

There are no runtime differences introduced by the refactored code.
The kernel register usage is mostly equal and for `KernelComputeCurrent`, `KernelSplitIntoListOfFrames`, `KernelComputeSupercell` and `KernelReduce` reduced.
The usage of context variables for `KernelShiftParticles` has changed a little bit to avoid 16bit reads to data.
I tried to find out why the compiler is transforming obvious 32bit reads into two 16bit reads but was not able to identify the problem but by transforming the code I was able to avoid the behavior.


psychocoderHPC:
I set this PR to draft mode to simplify the usage of lockstep programming much more.

psychocoderHPC:
I like to change to code that the usage looks like this:

```C++
    uint32_t const workerIdx = cupla::threadIdx( acc ).x;
    constexpr uint32_t numWorkers = 1u;
    constexpr uint32_t domainSize = 256u;
    auto frameDom = DomainConfig<domainSize, numWorkers>{workerIdx};
    
    frameDom.exec(
        [&](DomainIdx const domIdx)
        {
            // independent work
        });

    auto tmpVar = frameDom.var<int32_t>(
        [&]( DomainIdx const domIdx ) -> int32_t
        {
            return domIdx.lIdx();
        });

    frameDom.exec(
        [&]( DomainIdx const domIdx )
        {
            printf( "virtual worker linear idx: %u == %u\n", vIdx[ domIdx ], domIdx.lIdx() );
        });
```

@sbastrakov I would like to discuss this topic with you in a VC.

psychocoderHPC:
@sbastrakov This PR can be reviewed

psychocoderHPC:
The final syntax for the new interface will be:

```C++
    uint32_t const workerIdx = cupla::threadIdx( acc ).x;
    constexpr uint32_t numWorkers = 42u;
    constexpr uint32_t domainSize = 256u;
    auto forEach =lockstep::makeForEach<domainSize, numWorkers>(workerIdx);
    
    forEach(
        [](lockstep::Idx const idx)
        {
            // independent work
        });

    auto tmpVar = forEach(
        [](lockstep::Idx const idx) -> int32_t
        {
            return idx;
        });

    forEach(
        [&](lockstep::Idx const idx)
        {
            printf( "virtual worker linear idx: %u == %u\n", tmpVar[ idx ], static_cast<uint32_t>(idx) );
        });
```

CHANGELOG.md
```diff
@@ -846,7 +846,7 @@ and Sebastian Hahn for contributing to this release!
  - documentation:
    - new subtitle #2734
    - Lockstep Programming Model #2026 #2064
-   - `IdxConfig` append documentation #2022
+   - `DomainCfg` append documentation #2022
```
psychocoderHPC, line 849: must be referted!

docs/source/prgpatterns/lockstep.rst
```diff
@@ -19,21 +19,118 @@ Code which is implemented by the *lockstep programming model* is free of any dep
...
+    // is equal to
+
+    uint32_t const workerIdx = cupla::threadIdx(acc).x;
+    constexpr uint32_t frameSize = 256;
+    constexpr uint32_t numWorkers = 42;
+    auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize, numWorkers>(workerIdx);
+    // variable will be uninitialized
+    auto vIdx = lockstep::makeVar<int32_t>(forEachParticleSlotInFrame);
+    forEachParticleSlotInFrame(
+        [&](lockstep::Idx const idx)
```
sbastrakov, line 83: Could `lockstep::Idx` be `auto` instead? Asking both as a technical possibility, and if that is a good recommended way

docs/source/prgpatterns/lockstep.rst
```diff
@@ -19,21 +19,118 @@ Code which is implemented by the *lockstep programming model* is free of any dep
...
+.. doxygenstruct:: pmacc::lockstep::ForEach
    :project: PIConGPU
 
 Common Patterns
 ---------------
 
+Create a Context Variable
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* ... and initialize with the index of the virtual worker
```
sbastrakov, line 59: I think what is a `context variable` needs to be explained first.

docs/source/prgpatterns/lockstep.rst
```diff
@@ -45,16 +142,19 @@ Collective Loop
...
     {
-        uint32_t const workerIdx = cupla::threadIdx( acc ).x;
-        using ParticleDomCfg = IdxConfig<
-            frameSize,
-            numWorker
-        >;
-        ForEachIdx< ParticleDomCfg > forEachParticle( workerIdx );
-        forEachParticle(
-           [&]( uint32_t const linearIdx, uint32_t const idx )
+        uint32_t const workerIdx = cupla::threadIdx(acc).x;
```
sbastrakov, line 145: Could you add a comment that here we assume 1d indexing of threads inside block?

docs/source/prgpatterns/lockstep.rst
```diff
@@ -45,16 +142,19 @@ Collective Loop
...
+        constexpr uint32_t frameSize = 256;
+        constexpr uint32_t numWorkers = 42;
+        auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize, numWorkers>(workerIdx);
+        forEachParticleSlotInFrame(
+           [&](lockstep::Idx const idx)
+           {
+               // independent work, idx can be used to access a context variable
+           }
+        forEachParticleSlotInFrame(
+           [&](uint32_t const linearIdx)
```
sbastrakov, line 155: Perhaps I did not express my idea clearly during the VC. Do we really need to make a distinction between the two cases? I feel it confuses much more than helps, and simple `auto` and unified example would be easier to use for vast majority of cases

include/pmacc/lockstep/Config.hpp
```diff
@@ -0,0 +1,53 @@
...
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with PMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "pmacc/types.hpp"
```
sbastrakov, line 24: `#include <cstdint>` is missing

include/pmacc/lockstep/ForEach.hpp
```diff
@@ -0,0 +1,226 @@
...
+         * Algorithm to execute a subsequent lockstep for each index of the configured domain.
+         * @attention There is no implicit synchronization between workers before or after the execution of ForEach
+         * performed.
+         *
+         * @tparam T_Config Configuration for the domain and execution strategy.
+         *                  T_Config must provide: domainSize, numCollIter, numWorkers, and simdSize at compile time.
+         */
+        template<typename T_Config>
+        struct ForEach
+            : T_Config
```
sbastrakov, line 45: I would prefer not template-parametrizing for any `T_Config` but having `Config<T_a, T_b, T_c>` since this is not supposed to work with any other class and would make it easier to follow. But no big deal.

include/pmacc/lockstep/ForEach.hpp
```diff
@@ -0,0 +1,226 @@
...
+            using BaseConfig::numCollIter;
+            using BaseConfig::numWorkers;
+            using BaseConfig::simdSize;
+
+            /** constructor
+             *
+             * @param workerIdx index of the worker: range [0;workerSize)
+             */
+            HDINLINE
+            ForEach(uint32_t const workerIdx) : Worker<numWorkers>(std::move(workerIdx))
```
sbastrakov, line 60: I think `std::move` is doing nothing here? `workerIdx` is already a temporary

include/pmacc/lockstep/ForEach.hpp
```diff
@@ -0,0 +1,226 @@
...
+             * void operator()( lockstep::Idx< T_domainSize > const idx );
+             * // or
+             * void operator()( uint32_t const linearIdx );
+             * @endcode
+             */
+            template<
+                typename T_Functor,
+                std::enable_if_t<
+                    std::is_void<decltype(std::declval<T_Functor>()(std::declval<Idx const>()))>::value
+                        && domainSize != 1,
```
sbastrakov, line 89: Why is `domainSize == 1` specially forbidden?

include/pmacc/lockstep/ForEach.hpp
```diff
@@ -0,0 +1,226 @@
...
+        template<uint32_t T_domainSize, uint32_t T_numWorkers, uint32_t T_simdSize = 1u>
+        HDINLINE auto makeForEach(uint32_t const workerIdx)
+        {
+            return ForEach<Config<T_domainSize, T_numWorkers, T_simdSize>>(workerIdx);
+        }
+
+        HDINLINE auto makeMaster(uint32_t const workerIdx)
+        {
+            return Master(workerIdx);
+        }
```
sbastrakov, line 223: These are part of the public interface and should be doxygen-commented

include/pmacc/lockstep/Idx.hpp
```diff
@@ -0,0 +1,75 @@
...
+        struct Idx
+        {
+            /** Constructor
+             *
+             * @param domElemIndex linear index within the domain
+             * @param workerElemIndex virtual workers linear index of the work item
+             */
+            HDINLINE Idx(uint32_t const domElemIndex, uint32_t const workerElemIndex)
+                : domElemIdx(std::move(domElemIndex))
+                , workerElemIdx(std::move(workerElemIndex))
```
sbastrakov, line 44: As asked above for the similar case, is this `std::move` doing anything here?

sbastrakov (commented):
Very nice. I wrote some small points above. As individual messages, as my internet kept disconnected and I couldn't collect it all in one review.

docs/source/prgpatterns/lockstep.rst
psychocoderHPC: Yes could be auto too but I prefer to be explicit to see the difference between `uint32_t const` which can be used too.
`lockstep::Idx` is also better for search an replace^^

docs/source/prgpatterns/lockstep.rst
```diff
@@ -19,21 +19,118 @@ Code which is implemented by the *lockstep programming model* is free of any dep
...
+    );
+
+* To default initialize a context variable you can pass the arguments directly during the creation.
+
+.. code-block:: cpp
+
+    uint32_t const workerIdx = cupla::threadIdx(acc).x;
+    constexpr uint32_t frameSize = 256;
+    constexpr uint32_t numWorkers = 42;
+    auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize, numWorkers>(workerIdx, 42);
```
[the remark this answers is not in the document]
psychocoderHPC: Before we had no example to show how you can directly initialize a value, so this is a new example.

docs/source/prgpatterns/lockstep.rst
psychocoderHPC: You can use `auto` but if you like to enforce a data type you can do it.
This gives the user the possibility to avoid implicit casts in the lambda function.

include/pmacc/lockstep/ForEach.hpp
psychocoderHPC: For some reason, explicitly using `std::move` here is reducing the register footprint.

include/pmacc/lockstep/Idx.hpp
psychocoderHPC: Yes it is explicit enforcing a move and decrease the register footprint on NVIDIA V00

psychocoderHPC:
I hate compiler, g++6 and g++7 with CUDA 9.2 and CUDA 10 have problems to validate `std::enable_if` correctly :-(

https://gitlab.com/hzdr/crp/picongpu/-/jobs/1293799251#L729
https://gitlab.com/hzdr/crp/picongpu/-/jobs/1293799249#L719

and clang++ and CUDA 9.2 have issues with function qualifier :-(
https://gitlab.com/hzdr/crp/picongpu/-/jobs/1293799215#L669
A lambda is called from a device function, the compiler assumes that the lambda ist a host function :-(

psychocoderHPC:
Interesting we ship `alpaka::core::declval()` with alpaka.

sbastrakov:
I actually had a suspicion you might have misused `enable_if` the same way I did when doing alpak openMP schedule stuff. But I was not 100%, can look again if you like me to.

psychocoderHPC:
> I actually had a suspicion you might have misused `enable_if` the same way I did when doing alpak openMP schedule stuff. But I was not 100% sure, can look again if you like me to.

Yes please have a look to ot. I can write it also baded on return type validation but found the current implementation better because it is readable.
All compiler exept two old combinations works fine.
Looking to the error message also showed that definive different signatures are enabled which can not be because the condition is definitiv different.

psychocoderHPC:
This sounds like my issue and what you explained me last week, if I remember correctly.

https://stackoverflow.com/q/49876358

psychocoderHPC:
@sbastrakov I think I found a solution, not very nice but my tests showed that this works for old compiler. I hope it will not break new compiler. The last commit contains the workaround. The workaround should be documented but I will wait if this solves all issues.

psychocoderHPC:
> @sbastrakov I think I found a solution, not very nice but my tests showed that this works for old compiler. I hope it will not break new compiler. The last commit contains the workaround. The workaround should be documented but I will wait if this solves all issues.

g++ is now working but clang 8,9,10 + cuda 9.2 and 10.0 is failing now :-(

psychocoderHPC:
@sbastrakov I updated the documentation and addressed your point.
After a long night and testing the full morning the CI should now pass.

include/pmacc/lockstep/ForEach.hpp
```diff
@@ -0,0 +1,287 @@
...
+             * @code
+             * void operator()( lockstep::Idx< T_domainSize > const idx );
+             * // or
+             * void operator()( uint32_t const linearIdx );
+             * @endcode
+             */
+            template<
+                typename T_Functor,
+                // check if functor is invocable
+                typename = Void<decltype(alpaka::core::declval<T_Functor>()(alpaka::core::declval<Idx const>()))>,
```
sbastrakov, line 124: I think it would be a little easier to read if `decltype(alpaka::core::declval<T_Functor>()(alpaka::core::declval<Idx const>()))` or maybe even better a `Void<>` of that became its own named alias like
```cpp
template<typename T_Functor>
using IsCallableWithIndex = Void<decltype(alpaka::core::declval<T_Functor>()(alpaka::core::declval<Idx const>()))>;
```

But in terms of code lines that would only increase it, so do as you like.

sbastrakov (approved):
Nice. I wrote one more idea, but that is very optional to do.

include/pmacc/lockstep/ForEach.hpp
psychocoderHPC: Let us merge this PR as it is because I know  this is working on different compiler.
I will refactor the foreach in a follow up anyway and will test your suggestion too.

CHanging void is not necessary because it will be removed with alapaks `Void` as soon as we have alpaka 0.6.0 or 0.6.1 in PIConGPU.

psychocoderHPC: I marked it unresolved to find it faster, else I need to check all resolved comment boxes

include/picongpu/particles/creation/creation.kernel
```diff
@@ -153,26 +147,26 @@ namespace picongpu
...
+                        particleCreatorCtx[idx].init(acc, blockCell, idx, localCellIndex);
                     });
 
                     /* Declare counter in shared memory that will later tell the current fill level or
                      * occupation of the newly created target frames.
                      */
                     PMACC_SMEM(acc, newFrameFillLvl, int);
 
-                    ForEachIdx<IdxConfig<2, numWorkers>> onlyMasters{workerIdx};
+                    auto onlyMasters = lockstep::makeMaster(workerIdx);
```
psychocoderHPC, line 158: BUG!! This is not a single master we use two master worker to maintain out frame double buffer

docs/source/prgpatterns/lockstep.rst
```diff
@@ -19,21 +19,124 @@ Code which is implemented by the *lockstep programming model* is free of any dep
...
 
+Functors passed into lockstep routines can have three different parameter signatures.
+
+* No parameter, if the work is not requiring the linear index within a domain: ``[&](){ }``
+
+
+* An unsigned 32bit integral parameter if the work depends on indices within the domain ``range [0,domain size)``: ``[&](uint32_t const linearIdx){}``
+
+
+* ``lockstep::Idx`` as parameter. lockstep::Idx is holing the linear index within the domain and meta information to access a context variables: ``[&](pmacc::mappings::threads::lockstep::Idx const idx){}``
```
ax3l, line 30: typo: holding

include/picongpu/fields/FieldJ.kernel
```diff
@@ -276,19 +266,18 @@ namespace picongpu
...
-                        //   Change of the dE = - j / EPS0 * dt
-                        //                        j = current density (= current per area)
-                        //                          = fieldJ
-                        currentInterpolationFunctor(fieldE.shift(cell), fieldB.shift(cell), cachedJ.shift(cellIdx));
-                    });
+                lockstep::makeForEach<cellsPerSuperCell, numWorkers>(workerIdx)([&](uint32_t const linearIdx) {
+                    /* cell index within the superCell */
+                    DataSpace<simDim> const cellIdx
+                        = DataSpaceOperations<simDim>::template map<SuperCellSize>(linearIdx);
+                    DataSpace<simDim> const cell(blockCell + cellIdx);
```
ax3l, line 273: Looking at this repeating pattern, I think there should be a high-level helper `ForEachCell` that does the same logic but already performs the index shifts for passed fields.

PR #3625, opened 2021-05-25
[WIP] integrate parataxis in PIConGPU

pordyna:
I will update this PR with any major progress in my implementation of  [`parataxis`](https://github.com/ComputationalRadiationPhysics/parataxis) in `picongpu`. 
### parataxis
So `parataxis` is a Monte Carlo X-ray scattering code that simulates a  SAXS (small angle scattering) diagnostic. It uses photon like macro particles to represent possible photon paths in the simulation volume. These photons can be scattered (their propagation direction can be changed) at each time step with a given probability that depends on local plasma parameters. When a photon leaves the simulation it is further propagated towards the detector where it contributes to the detector signal. One can think of the photons as of representations of Huygens waves which complex amplitudes are then summed at the detector forming an interference pattern. 

### motivation
It can calculate multiple scattering which is not possible with the `xrayScattering` plugin. It could be combined with synchrotron radiation and Bremstrahlung.

### todo
There are 3 main implementation steps: 
- [x] detector part
- [ ]  scattering part 
- [ ]  laser part (spawn photons at simulation border)

### detector
For now, I have implemented the detector part in the `photonDetector` plugin. It is responsible for storing the detector data, particle ray tracing (from GUARD to a detector cell),  detector signal accumulation, and output writing. 

#### Plugin options
* runtime options:
   * __detector placement__ It can be placed behind one of the simulation box sides. It is always placed parallel to the simulation box and so that the center of the detector is in front of the simulation box center. 
   *  __detectorSizeX__ number of detector cells along the detector x coordinate
   *  __detectorSizeY__ number of detector cells along the detector y coordinate
   * parameters customizing the `openPMD` output
* compile time options:
   * __distance__ distance between the detector and the simulation
   * __cellSizeX, cellSizeY__ size of a single detector cell
   *  __AccumulationPolicy__ defines how the particles contribute to the detector signal. At the moment there are two policies. `CountParticles` simply sums weighting of particles landing in a given cell. `AddWaveParticles` adds amplitudes of wave like particles. 
 
#### main parts of the code
*  `photonDetector.hpp` main plugin file
*  `PhotonDetectorImpl` stores detector data, defines ray-tracing, creates a particle functor for particle detection (ray tracing + accumulation)
*  `photonDetector/accumulation/` contains accumulation policies
*  `PhotonDetectorWriter` handles output writing

#### photon types
The `AddWaveParticles` can be used with two types of photon like macro particles. The difference is in how the phase of a photon is calculated.  The particle phase can be obtained from:
1. A constant starting phase that can be different for each particle, and a value that is common for all particles in the species for a given simulation time step. This second part is calculated from the simulation time and the species `wavelength` flag. 
2. A particle attribute `phase` that needs to be updated by a pusher (based on the photon momentum attribute).
                     
* Approach 1 is used when the particle doesn't have the `phase` attribute. In that case the species must have the `wavelength` flag and the `startPhase` attribute.
* Approach 2 is used when the particle has the `phase` attribute.
                     

#### tests
There is a small test case that checks the ray tracing for different detector placements. The test example has two species. Photons with a fixed wavelength, set with the `wavelength` flag, and photons with the `phase` attribute and the `PhotonPhase` pusher.
Each species has just one particle in the simulation. The particles are initiated at one of the simulations boundaries and are send towards a 3x3 cell detector. There is set of runtime tests which verify that the correct detector cell is hit for different initial photon directions and detector placements. Output writing for both file and group based iteration encodings with the `json`, `h5`, and `BP4` (`adios`) `openPMD` backends is also checked. 

These tests are defined in a python script using the `pytest` module and can be run with the `pytest`executable from the command line. There is also an `sbatch` script in `share/tpicongpu/PhotonDetector/etc/picongpu/hemera-hzdr` for running these tests on the hemera gpu partition.

I plan on testing the phase calculation and the amplitude accumulation once I have all 3 parts implemented. I will probably just use the double slit example from `parataxis`.

#### some details
* The phase on the detector is calculated relative to a phase of a particle that leaves the simulation box at the center of the simulation box side and hits the particular detector cell. So that there is a constant phase factor missing from the output ( different for each detector cell). parataxis is writing the intensity in the output so it wasn't a problem. I'm writing directly the complex values so I should multiply the result with these constant factors before writing. (Not yet implemented)
* I think I should rename this plugin to `planarDetector` since it could be useful for other particles as well.
* I reused some code(`ProbingBeam`) from the `xrayScattering` plugin and moved it outside the `xrayScattering` namespace.


pordyna:
@ComputationalRadiationPhysics/picongpu-maintainers I have more of less finished the first part of my `parataxis` implementation in `picongpu`. Maybe you want to already have a look at my code while I'm working the next implementation step.

PrometheusPi:
@pordyna Could you link/reference the current branch/pull request now associated with this closed branch?

pordyna:
current WIP branch https://github.com/pordyna/picongpu/tree/topic-RadiationTransport

PR #3626, opened 2021-05-26
fix broken OpenPMD IO plugin

### LINKED ISSUE #3624: openPMD output fails on AMD MI100
I try to get PIConGPU running on [spock](https://docs.olcf.ornl.gov/systems/spock_quick_start_guide.html) cluster at OLCF

For now I had to compile blosc (1.21.0), ADIOS2 (2.7.1) and openPMD (0.13.4) myself.
When running the *LaserWakefield* example I receive the following error during write out of the first simulation data
```
Unhandled exception of type 'St13runtime_error' with message 'expression ((uint64_cu) globalParticleOffset == rp.globalNumParticles) failed in file (/ccs/home/ksteinig/src/picongpu/include/pmacc/../picongpu/plugins/openPMD/WriteSpecies.hpp:173) : ', terminating
```

The exact same setup runs on summit. Do you have an idea what to do?

**cc'ing** @sbastrakov, @psychocoderHPC, @franzpoeschel 

PS: I did not compile `zlib`, as it seems not required. (I would have expected a compile error if it would be required). However the manual lists it as required, so I am not 100% sure.

**EDIT**: the full error log is attach now.

[stderr.txt](https://github.com/ComputationalRadiationPhysics/picongpu/files/6538301/stderr.txt)

psychocoderHPC:
fix #3624

The number of local particles prepared was incorrect and additionally the assert to validate that the correct number of particles is processed was wrong too.
Use `PMACC_VERIFY` instead of `PMACC_ASSERT` to verify the number of processed particles even if we not compile with debug enabled.

### DIFF
--- a/include/picongpu/plugins/openPMD/WriteSpecies.hpp
+++ b/include/picongpu/plugins/openPMD/WriteSpecies.hpp
@@ -85,7 +85,7 @@ namespace picongpu
                 , filter(c_filter)
                 , particleFilter(c_particleFilter)
                 , particleOffset(c_particleOffset)
-                , myNumParticles(c_globalNumParticles)
+                , myNumParticles(c_myNumParticles)
                 , globalNumParticles(c_globalNumParticles)
             {
             }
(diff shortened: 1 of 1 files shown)


steindev (approved):
Tested on spock, where the original issue appeared. Works with this fix!

psychocoderHPC:
@steindev The PR has passed the CI

PR #3630, opened 2021-05-27
refactor `lockstep::ForEach`

psychocoderHPC:
Split the forEach compile-time loop into peel and reminder part.
This is slightly better for CPU compiler and is more human-readable compared to the original implementation.
Add two helper types to derive if a functor is callable with or without Idx as an argument. 
This was suggested by @sbastrakov in #3616.

### DIFF
--- a/include/pmacc/lockstep/ForEach.hpp
+++ b/include/pmacc/lockstep/ForEach.hpp
@@ -128,14 +128,16 @@ namespace pmacc
              */
             template<
                 typename T_Functor,
-                // check if functor is invokable
+                // check if functor is invocable
                 typename = IsCallableWithIndex<T_Functor>,
                 std::
                     enable_if_t<std::is_void<InvokeResult_t<T_Functor, Idx const>>::value && domainSize != 1, int> = 0>
             HDINLINE void operator()(T_Functor&& functor) const
             {
-                static constexpr uint32_t numFullLoops = domainSize / (simdSize * numWorkers);
-                for(uint32_t i = 0u; numFullLoops != 0u && i < numFullLoops; ++i)
+                // number of iterations each worker can safely execute without boundary checks
+                constexpr uint32_t peeledIterations = domainSize / (simdSize * numWorkers);
+                // peeledIterations != 0u avoid compiler warnings if peeledIterations == 0
+                for(uint32_t i = 0u; peeledIterations != 0u && i < peeledIterations; ++i)
                 {
                     uint32_t const beginWorker = i * simdSize;
                     uint32_t const beginIdx = beginWorker * numWorkers + simdSize * this->getWorkerIdx();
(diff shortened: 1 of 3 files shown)


include/pmacc/lockstep/ForEach.hpp
```diff
@@ -149,8 +162,8 @@ namespace pmacc
              */
             template<
                 typename T_Functor,
-                // check if functor is invocable
-                typename = Void<decltype(alpaka::core::declval<T_Functor>()())>,
+                // check if functor is invokable
```
sbastrakov, line 165: Original spelling with `c` was correct. You also changed it in another place

include/pmacc/lockstep/ForEach.hpp
```diff
@@ -120,26 +128,31 @@ namespace pmacc
...
-                typename = Void<decltype(alpaka::core::declval<T_Functor>()(alpaka::core::declval<Idx const>()))>,
+                // check if functor is invokable
+                typename = IsCallableWithIndex<T_Functor>,
                 std::
                     enable_if_t<std::is_void<InvokeResult_t<T_Functor, Idx const>>::value && domainSize != 1, int> = 0>
             HDINLINE void operator()(T_Functor&& functor) const
             {
-                for(uint32_t i = 0u; i < numCollIter; ++i)
+                static constexpr uint32_t numFullLoops = domainSize / (simdSize * numWorkers);
+                for(uint32_t i = 0u; numFullLoops != 0u && i < numFullLoops; ++i)
```
sbastrakov, line 138: I have a few questions/suggestions here:

- Maybe add a comment above that we split loop into peeled (or we can call it full like currently) and remainder, and explain what peeled means?
- Isn't `numFullLoops != 0u` redundant? If it is 0, then `i < numFullLoops` would be false initially.
- The name `numFullLoops` is misleading, I would consider renaming to `numFullIterations` or `numFullLoopIterations`, or with `Peeled` instead of `Full`

include/pmacc/lockstep/ForEach.hpp
```diff
@@ -120,26 +128,31 @@ namespace pmacc
...
                     uint32_t const beginWorker = i * simdSize;
                     uint32_t const beginIdx = beginWorker * numWorkers + simdSize * this->getWorkerIdx();
-                    if(outerLoopCondition || !innerLoopCondition || beginIdx < domainSize)
+                    for(uint32_t s = 0u; s < simdSize; ++s)
+                        functor(Idx(beginIdx + s, beginWorker + s));
+                }
+
+                // process left over indices if the domainSize is not a multiple of 'simdSie * numWorkers'
+                static constexpr bool hasRemainder = (domainSize % (simdSize * numWorkers)) != 0u;
+                static constexpr uint32_t leftOverIndices = domainSize - (numFullLoops * numWorkers * simdSize);
```
sbastrakov, line 148: Why are these variables static?

include/pmacc/lockstep/ForEach.hpp
```diff
@@ -120,26 +128,31 @@ namespace pmacc
...
+                for(uint32_t s = 0u; hasRemainder && s < simdSize; ++s)
+                    if(hasRemainder && this->getWorkerIdx() * simdSize + s < leftOverIndices)
```
sbastrakov, line 150: If we are here, `hasRemainder` is true. Seems no need to check in the `if` statement.

sbastrakov (changes requested):
Nice, just some small comments.

include/pmacc/lockstep/ForEach.hpp
psychocoderHPC: They should not be static. It is a copy passed bug because I had the definition during my development inside the class.

include/pmacc/lockstep/ForEach.hpp
psychocoderHPC: If this is executed we do not know if we need to calculate `simdSize` steps or less.
If e.g. simdSize is `2` it could be the first index is within the domain the second is not.
`hasRemainder` is required to hide comparison warning because all values are known at compile time. I should add a comment above.

sbastrakov: From  the readability point of view, I think the best would be to just compute a proper ending index and have a simple loop from 0 to that. I assume you didn't go for it, since it's additional registers?

psychocoderHPC: Unfortionally this loop condition is only allowed to have compile-time values therefore we have the runtime condition as `if` condition in the body.
This allows the compiler to perform range checks and map access at compile time.

include/pmacc/lockstep/ForEach.hpp
psychocoderHPC: > Isn't numFullLoops != 0u redundant? If it is 0, then i < numFullLoops would be false initially.

`numFullLoops` is unsigned and therefore the compiler will throw warnings if `numFullLoops == 0`.
I added a comment.

psychocoderHPC:
I updated the commit

sbastrakov (approved):

PR #3633, opened 2021-05-31
signal handling

### LINKED ISSUE #19: Signal Support
Add Posix signal support as described in our [doc/SIGNALS.md](https://github.com/ComputationalRadiationPhysics/picongpu/blob/0.1.0/doc/SIGNALS.md)

We will probably need a more general **Simulation Control** class to get the communication between the _signal handler_ (thread?), _plugins_ (signal reduce/sync) and the _simulation itself_ clean.

psychocoderHPC:
fix #19

- add new checkpoint parameter `--checkpoint.tryRestart`
- handle incoming signals
- `TimeSlice` extent constructor
- update documentation
- hemera-hzdr: 
  - map Slurm signals to PIConGPU needs
  - send signal 240 seconds before we run into the walltime limit (trigger checkpoint)

The signal mapping in this implementation is different from the suggestion in the issue. 
The reason is that [mpiexec from OpenMPI](https://www.open-mpi.org/doc/v3.0/man1/mpiexec.1.php#toc15) is only propagating a limited number of signals.
Using SIGUSR1 and SIGUSR2 will guarantee the best compatibility with other MPI flavors and different batch systems.

# Review

Best review the last three commits separate.

# Todo

- [x] merge after #3632


include/pmacc/pluginSystem/TimeSlice.hpp
```diff
@@ -74,6 +74,22 @@ namespace pmacc
...
+
+            /** Constructor
+             *
+             * Select any time step in the range [start,end].
+             *
+             * @param start first time step taken into account
+             * @param end last time step taken into account
+             * @param period select any period(st) time step between start and end
+             */
+            TimeSlice(uint32_t start, uint32_t end = uint32_t(-1), uint32_t period = 1)
```
sbastrakov, line 86: If `start = 0` by default, the default constructor above seems not required.

etc/picongpu/hemera-hzdr/defq.tpl
```diff
@@ -97,5 +105,5 @@ export OMPI_MCA_io=^ompio
 
 if [ $? -eq 0 ] ; then
   # Run PIConGPU
-  mpiexec --bind-to none !TBG_dstPath/tbg/cpuNumaStarter.sh !TBG_dstPath/input/bin/picongpu !TBG_author !TBG_programParams
+  source !TBG_dstPath/tbg/handleSlurmSignals.sh mpiexec --bind-to none !TBG_dstPath/tbg/cpuNumaStarter.sh !TBG_dstPath/input/bin/picongpu !TBG_author !TBG_programParams
```
sbastrakov, line 108: Is this a single line on purpose?

sbastrakov (dismissed):

etc/picongpu/hemera-hzdr/defq.tpl
psychocoderHPC: I can break with \ at the line end. Will add it tomorrow

sbastrakov: Maybe it's not required. I was mostly wondering if that was an accidental missing end-of-line. It this is intended, could stay as is

psychocoderHPC: Wait let me test if I can call it `$(!TBG_dstPath/tbg/handleSlurmSignals.sh mpiexec ...)`
It is required to call it in the same context as the batch script. I tested something else before and accidentally thought I need to use `source`.

include/pmacc/pluginSystem/TimeSlice.hpp
psychocoderHPC: updated

etc/picongpu/hemera-hzdr/defq.tpl
psychocoderHPC: updated

psychocoderHPC:
CI passed, ready to rumble

sbastrakov (approved):

PR #3635, opened 2021-06-02
Move absorber selection to runtime and unify PML and non-PML solvers

sbastrakov:
This PR changes a long-standing design decision to have PML as part of field solver. Which resulted in having two version of each field solver. Now a user just selects a solver in the .param file, and absorber in a .cfg file. Absorber parameters are still in their respective .param files and unchanged, those will be moved to .cfg files in a follow-up PR.

- [x] Rebase after #3629 is merged


include/picongpu/simulation/stage/FieldAbsorber.hpp
```diff
@@ -0,0 +1,80 @@
...
+                        Absorber::kind() = Absorber::Kind::Exponential;
+                    else if(kindName == "pml")
+                        Absorber::kind() = Absorber::Kind::Pml;
+                    else
+                        throw std::runtime_error("Unsupported field absorber type");
+                }
+
+            private:
+                //! Name set by program option
+                std::string kindName = "exponential";
```
psychocoderHPC, line 77: should we change the default to `pml` this is also a TODO from our developer meeting few months ago?

include/picongpu/simulation/stage/FieldAbsorber.hpp
```diff
@@ -0,0 +1,80 @@
...
+                /** Register program options for field absorber
+                 *
+                 * @param desc program options following boost::program_options::options_description
+                 */
+                void registerHelp(po::options_description& desc)
+                {
+                    desc.add_options()(
+                        "fieldAbsorber",
+                        po::value<std::string>(&kindName),
+                        (std::string("Field absorber kind [exponential, pml] default: " + kindName).c_str()));
```
psychocoderHPC, line 53: the outer parenthesis  can be removed `((...).c_str())` to `(...).c_str()`

include/picongpu/fields/absorber/exponential/Exponential.hpp
```diff
@@ -0,0 +1,134 @@
...
+                 * Sets instance and implements absorption.
+                 */
+                class Exponential : public Absorber
+                {
+                public:
+                    //! Create exponential damping absorber instance
+                    Exponential()
+                    {
+                        // Copy thickness from grid.param
+                        for(uint32_t axis = 0u; axis < 3u; axis++)
```
psychocoderHPC, line 57: Is this iterating over 3 dimensions for simplification or is this required? Else I would suggest to use `axis < simDim` and update the definition too.

Could you use `axis < DIM3`. We use the define `DIM2, DIM3` where we depend on the on how many dimensions PIConGPU is supporting in the current compiled version.
This would make finding such code parts which depend on the simulation dimensionality easier if we like to refactor picongpu.

docs/source/usage/workflows/boundaryConditions.rst
```diff
@@ -32,14 +32,13 @@ It is recommended to avoid, as possible, having particles in the field absorber
 Ideally, only particles leaving the simulation area are present there, on their way to be absorbed.
 Note that particle absorption happens at the external surface of the field absorber layer, matching the global simulation area border.
 
-The field absorber mechanism and user-controlled parameters depend on the field solver enabled.
+The field absorber mechanism and user-controlled parameters depend on the field absorber kind enabled.
 
-Field solvers ``Yee<>``, ``Lehe<>``, and ``ArbitraryOrderFDTD<>`` use the exponential damping absorber.
-Its parameters are controlled by :ref:`grid.param <usage-params-core>`.
+It is controlled by command-line option ``fieldAbsorber.kind``.
```
psychocoderHPC, line 37: maybe be `--fieldAbsorber.kind` even if you already mentioned that this is a command line parameter but in all other documentation we explicitly used `--`.

Could we add a sentence that the absorber is in all directions used were `--periodic` is set to zero.

include/picongpu/plugins/adios/ADIOSWriter.hpp
```diff
@@ -331,6 +331,9 @@ namespace picongpu
 #ifndef __CUDA_ARCH__
                     DataConnector& dc = Environment<simDim>::get().DataConnector();
 
+                    // Skip PML fields when PML is not enabled
+                    if(!dc.hasId(T_Field::getName()))
```
psychocoderHPC, line 335: This looks dangerous, it looks like all fields which are not registered in the data connector will be skipped.
Do we have maybe a better way to explicitly only skip PML fields and not all unknown?

Maybe:

```
if(T_Field::getName() == "pmlFields" && !dc.hasId(T_Field::getName())
```

include/picongpu/plugins/adios/ADIOSWriter.hpp
```diff
@@ -679,6 +682,9 @@ namespace picongpu
                     if(!traits::IsFieldDomainBound<T>::value)
                     {
                         DataConnector& dc = Environment<>::get().DataConnector();
+                        // Skip PML fields when PML is not enabled
+                        if(!dc.hasId(T::getName()))
```
psychocoderHPC, line 686: same as above sounds to me that we silently ignore all unknown fields.

include/picongpu/plugins/adios/restart/RestartFieldLoader.hpp
```diff
@@ -233,6 +233,10 @@ namespace picongpu
                 DataConnector& dc = Environment<>::get().DataConnector();
                 ThreadParams* tp = params;
 
+                // Skip PML fields when PML is not enabled
+                if(!dc.hasId(T_Field::getName()))
```
psychocoderHPC, line 237: Same as above we hide at this line all unknown fields.

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -449,6 +449,9 @@ Please pick either of the following:
 #ifndef __CUDA_ARCH__
                     DataConnector& dc = Environment<simDim>::get().DataConnector();
 
+                    // Skip PML fields when PML is not enabled
+                    if(!dc.hasId(T_Field::getName()))
```
psychocoderHPC, line 453: Same as above we hide at this line all unknown fields.

include/picongpu/plugins/openPMD/restart/RestartFieldLoader.hpp
```diff
@@ -208,6 +208,10 @@ namespace picongpu
                 DataConnector& dc = Environment<>::get().DataConnector();
                 ThreadParams* tp = params;
 
+                // Skip PML fields when PML is not enabled
+                if(!dc.hasId(T_Field::getName()))
```
psychocoderHPC, line 212: Same as above we hide at this line all unknown fields.

psychocoderHPC (changes requested):

include/picongpu/simulation/stage/FieldAbsorber.hpp
sbastrakov: Maybe. I think that should be a separate PR. Now my idea was that setups with `Yee` solver were supposed to use exponential absorber and so they still do.

include/picongpu/fields/absorber/exponential/Exponential.hpp
sbastrakov: `grid.param` requires setting all 3 as of now. I agree it makes more sense to have only till `simDim`.

include/picongpu/plugins/adios/ADIOSWriter.hpp
sbastrakov: Yes, so here we have an issue that some part, the typelist of fields in this case, is compile-time, and another part is run-time. I would not like to have any PML information here. Could make a trait, is it good with you?

docs/source/usage/workflows/boundaryConditions.rst
sbastrakov: Adding `--`: okay, will do.
Regarding `--periodic` this is already written at start of this file. Do you think it needs further clarification here? I don't

include/picongpu/fields/MaxwellSolver/Yee/Yee.hpp
```diff
@@ -45,114 +44,121 @@ namespace picongpu
     {
         namespace maxwellSolver
         {
-            template<class CurlE, class CurlB>
+            // TODO move this comment
```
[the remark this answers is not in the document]
sbastrakov: Ah yes. Was leftover by mistake.

docs/source/usage/workflows/boundaryConditions.rst
psychocoderHPC: >  Do you think it needs further clarification here?

No, I missed the part at the top of the file. All is fine. +1

include/picongpu/plugins/adios/ADIOSWriter.hpp
psychocoderHPC: Yes whatever fits best, as long as it avoids skipping IO if an unknown field is added somewhere in param or unitless.
This would avoid that we search in the future why our configured data is not written.

include/picongpu/fields/MaxwellSolver/Yee/Yee.hpp
```diff
@@ -163,6 +169,142 @@ namespace picongpu
...
+                        (mapper.getGridDim(),
+                         numWorkers)(mapper, updateFunctor, fieldE->getDeviceDataBox(), fieldB->getDeviceDataBox());
+                    }
+                    else
+                        PMACC_KERNEL(Kernel{})
+                    (mapper.getGridDim(), numWorkers)(
+                        mapper,
+                        yee::UpdateBHalfFunctor<CurlE>{},
+                        fieldE->getDeviceDataBox(),
+                        fieldB->getDeviceDataBox());
```
sbastrakov, line 250: @psychocoderHPC @j-stephan here is another example of our clang-formatting not managing the (admittedly, ugly) code. The resulting formatting is really illogical.

include/picongpu/plugins/adios/ADIOSWriter.hpp
sbastrakov: @psychocoderHPC as an alternative. I could make it so that the convolutional fields are always created and regiatered in the data connector. PML implementation will create them of the actual size, exponential implementation with size 0 (if that is allowed everywhere) or 1. Then in checkpointing and output there is no need to know about them or treat them separately. So that is better in terms of interdependencies in the code. On the other hand, it could make the output files more confusing. What do you think?

include/picongpu/fields/MaxwellSolver/Yee/Yee.hpp
psychocoderHPC: That's right, this looks strange.

include/picongpu/plugins/adios/ADIOSWriter.hpp
psychocoderHPC: I think if we can avoid creating PML helper field and skip dumping this would be better for IO scaling. Even writing out a dataset with zero entries can result in MPI calls (depending of the IO backend) and meta data is written.

psychocoderHPC: > Now we also have another place for separate treatment: field reset after the window moves. That part also now checks if the PML fields are registered in the data connector. Kinda like output implementation as in PR currently does.

I would keep checking. Maybe (not in this PR) we should think about refactoring our DataConnector a little bit and additional information if data should be part of a checkpoint, by adding a flag. We already talked about making checkpointing more explicit maybe this is falling into the same topic.

psychocoderHPC: Maybe you can make a doxygen comment out of the comment above and add `//!@todo`

include/picongpu/fields/MaxwellSolver/Yee/Yee.hpp
j-stephan: Does this change if you put braces around the statement? I can imagine that clang-format gets confused because of the multiline statement. IIRC some compilers even issue warnings if you do this.

include/picongpu/plugins/adios/ADIOSWriter.hpp
psychocoderHPC: Could you please add the filter trait or via name to finalize this PR.

sbastrakov: Done

include/picongpu/fields/MaxwellSolver/Yee/Yee.hpp
sbastrakov: Braces didn't change anything. However, `clang-format` didn't mind me somewhat fixing it manually (so both how it done it originally and how it is now pass).

psychocoderHPC (approved):

include/picongpu/plugins/adios/ADIOSWriter.hpp
psychocoderHPC: @sbastrakov Do you test your PR with a simulation where some GPUs do not have the PML field? The problem is that IO is a collective operation and all ranks must participate.
IMO if some ranks do not have this additional field than not all ranks will do the IO. 
It depends on the implementation. Do you register `pmlFields` with a size zero for ranks which will not be calculating a simulation boundery?

sbastrakov: When PML is enabled, all GPUs have it. And all of them have PML fields of the max size that a domain can have. It was like that even before this PR. The reason is that "central" GPUs can become "border" ones due to moving window, and we need to preallocate memory at simulation start. So I do not expect this PR to change anything there. But will recheck

psychocoderHPC: Thx for the clarification.

PR #3637, opened 2021-06-04
Update spock-ornl profile

steindev:
The ROCm module has been fixed.
Now we can use it and do not need to use the system installation
anymore.

(Tested with LWFA example which had problems with the old module)


PrometheusPi (approved):
LGTM

PR #3638, opened 2021-06-05
fix observation direction precision to float_X

PrometheusPi:
In case of ultra high gamma particles, the precision of the observation direction needs to be increased since the radiation cone decreases with ~1/gamma. `float_32` are not precise enough for particles above gamma>800. 
This fix removed the hard coded `float_32` implementation of the observation direction calculation and sets it to `float_X`. Thus it can be adjusted by setting PIConGPU's precision. This will coast a bit of performance, but can currently not be handled otherwise.

### DIFF
--- a/include/picongpu/param/radiationObserver.param
+++ b/include/picongpu/param/radiationObserver.param
@@ -75,6 +75,9 @@ namespace picongpu
                  */
                 HDINLINE vector_64 observation_direction(const int observation_id_extern)
                 {
+                    /* float type used in radiation direction calculations */
+                    using float_obs = picongpu::float_X;
+
                     /* generate two indices from single block index */
                     /** split distance of given index
                      * pseudo-code:
@@ -117,12 +120,12 @@ namespace picongpu
                     const picongpu::float_64 phi(my_index_phi * delta_angle_phi - angle_phi_start);
 
                     /* helper functions for efficient trigonometric calculations */
-                    picongpu::float_X sinPhi;
-                    picongpu::float_X cosPhi;
-                    picongpu::float_X sinTheta;
-                    picongpu::float_X cosTheta;
-                    pmacc::math::sincos(precisionCast<picongpu::float_X>(phi), sinPhi, cosPhi);
-                    pmacc::math::sincos(precisionCast<picongpu::float_X>(theta), sinTheta, cosTheta);
+                    float_obs sinPhi;
+                    float_obs cosPhi;
+                    float_obs sinTheta;
+                    float_obs cosTheta;
+                    pmacc::math::sincos(precisionCast<float_obs>(phi), sinPhi, cosPhi);
+                    pmacc::math::sincos(precisionCast<float_obs>(theta), sinTheta, cosTheta);
                     /** compute observation unit vector */
                     return vector_64(sinTheta * cosPhi, cosTheta, sinTheta * sinPhi);
                 }
--- a/include/picongpu/param/radiationObserver.param
+++ b/include/picongpu/param/radiationObserver.param
@@ -117,12 +117,12 @@ namespace picongpu
                     const picongpu::float_64 phi(my_index_phi * delta_angle_phi - angle_phi_start);
 
                     /* helper functions for efficient trigonometric calculations */
-                    picongpu::float_32 sinPhi;
-                    picongpu::float_32 cosPhi;
-                    picongpu::float_32 sinTheta;
-                    picongpu::float_32 cosTheta;
-                    pmacc::math::sincos(precisionCast<picongpu::float_32>(phi), sinPhi, cosPhi);
-                    pmacc::math::sincos(precisionCast<picongpu::float_32>(theta), sinTheta, cosTheta);
+                    picongpu::float_X sinPhi;
+                    picongpu::float_X cosPhi;
+                    picongpu::float_X sinTheta;
+                    picongpu::float_X cosTheta;
+                    pmacc::math::sincos(precisionCast<picongpu::float_X>(phi), sinPhi, cosPhi);
+                    pmacc::math::sincos(precisionCast<picongpu::float_X>(theta), sinTheta, cosTheta);
                     /** compute observation unit vector */
                     return vector_64(sinTheta * cosPhi, cosTheta, sinTheta * sinPhi);
                 }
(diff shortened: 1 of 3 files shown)


PrometheusPi:
@ComputationalRadiationPhysics/picongpu-maintainers could you please vote whether you consider this a bug :+1: or not :-1:. so that we can add the label accordingly.

PrometheusPi:
A followup update of the radiation plugin's documentation will follow that explicitly states the current limits of the radiation calculations.

PrometheusPi:
To compare the previous version (left) to the fixed version of this pull request (right), please compare the spectra of the non-linear Thomson scattering below. With too low precision, the angular distribution (banana shape) is not continuous anymore. With this fix, the expected angular shape is reinstalled. 
grafik

include/picongpu/param/radiationObserver.param
```diff
@@ -117,12 +117,12 @@ namespace picongpu
                     const picongpu::float_64 phi(my_index_phi * delta_angle_phi - angle_phi_start);
 
                     /* helper functions for efficient trigonometric calculations */
-                    picongpu::float_32 sinPhi;
-                    picongpu::float_32 cosPhi;
-                    picongpu::float_32 sinTheta;
-                    picongpu::float_32 cosTheta;
-                    pmacc::math::sincos(precisionCast<picongpu::float_32>(phi), sinPhi, cosPhi);
-                    pmacc::math::sincos(precisionCast<picongpu::float_32>(theta), sinTheta, cosTheta);
+                    picongpu::float_X sinPhi;
```
PrometheusPi, line 120: I thin this should be 
```suggestion
                    float_X sinPhi;
```
I will fix this asap.

psychocoderHPC:
Your fix allows to workaround the precision issue by setting PIConGPU's precision to 64bit. Is it not better to allow adjusting the type in the `radiation.param`. This will still allow that PIConGPU runs with 32bit precision and allow setting radiation to 64bit.

To resolve features in a plugin there should be no need to change the simulation precision.

include/picongpu/param/radiationObserver.param
psychocoderHPC: Please introduce your own type within `radiation.param` which can be adjusted by the user.

psychocoderHPC (changes requested):

steindev:
> @ComputationalRadiationPhysics/picongpu-maintainers could you please vote whether you consider this a bug +1 or not -1 so that we can add the label accordingly.

I added the refactoring label. As it was working as expected up until now in all simulation setups, I would not consider it a bug.

psychocoderHPC:
> > @ComputationalRadiationPhysics/picongpu-maintainers could you please vote whether you consider this a bug +1 or not -1 so that we can add the label accordingly.
> 
> I added the refactoring label. As it was working as expected up until now in all simulation setups, I would not consider it a bug.

I marked it as bug. The setup is maybe strange but it is resulting in wrong results so the user should be aware but this PR.
In the changelog we have a section where all bugs are listed, this is normally something the user is looking in. 
Only the user knows if it is an issue for his simulation or not.

PrometheusPi:
@psychocoderHPC The `ci/gitlab/gitlab.com` is waiting for 10 days now. (it should fail.) I think there is something wrong with the CI.

psychocoderHPC:
I removed the assigned developer, could you push please again. 
There is a bug in the CI that as soon as you assign someone the CI is throwing a parsing error.
This is not a big deal because there is no need that someone to be assigned for an issue, you need only someone who reviews your changes.

psychocoderHPC:
@PrometheusPi PLease rebase this issue and push it again

PrometheusPi:
@psychocoderHPC I pushed the branch again. 
Next I will implement your suggestions.

PrometheusPi:
@psychocoderHPC I added the change you suggested.

include/picongpu/param/radiationObserver.param
```diff
@@ -35,6 +35,9 @@ namespace picongpu
         {
             namespace radiation_observer
             {
+                /* float type used in radiation direction calculations */
+                typedef picongpu::float_X floatRadObserver;
```
psychocoderHPC, line 39: please use `using floatRadObserver = picongpu::float_X`

We do not use `typedef` anymore.

share/picongpu/examples/Bunch/include/picongpu/param/radiationObserver.param
```diff
@@ -28,6 +28,9 @@ namespace picongpu
         {
             namespace radiation_observer
             {
+                /* float type used in radiation direction calculations */
+                typedef picongpu::float_X floatRadObserver;
```
psychocoderHPC, line 32: please use `using`

share/picongpu/examples/KelvinHelmholtz/include/picongpu/param/radiationObserver.param
```diff
@@ -28,6 +28,9 @@ namespace picongpu
         {
             namespace radiation_observer
             {
+                /* float type used in radiation direction calculations */
+                typedef picongpu::float_X floatRadObserver;
```
psychocoderHPC, line 32: please use `using`

psychocoderHPC (changes requested):

include/picongpu/param/radiationObserver.param
psychocoderHPC: Can we shorten this name? The declaration is already in the namespace `radiation_observer` therefore radiation and observer should not be part of the name, or shortened.

I suggest `float_ob`, `float_obs` or `float_dir`

PrometheusPi: Will do this.

sbastrakov: I am not sure why is this type needed as public API at all? It seems just a details of that function below.

PrometheusPi: It is not needed as public API, but should be user defined - what would you recommend instead @sbastrakov ?

sbastrakov: I think this type can be moved inside the function. With a comment to a user that it can be changed just here to control precision vs. calculation time

PrometheusPi:
@psychocoderHPC and @sbastrakov I implemented your requests.

psychocoderHPC (dismissed):

psychocoderHPC:
Please re-run clang-format. All changes look good to me.

PrometheusPi:
@psychocoderHPC all clang-format issues have been resolved.

sbastrakov (approved):

Issue #3581, opened 2021-04-10
Upgrade run time check of CFL condition for AOFDTD solver


steindev:
There is an updated stability condition for the arbitrary-order finite-difference time-domain method. See manual (https://picongpu.readthedocs.io/en/latest/models/AOFDTD.html).

bussmann:
Great work. The colours in the velocity error vs courant factor fig seem a bit off, though. The graph looks brownish-red while the legend for 16th looks bright red.

PR #3641, opened 2021-06-11
Fix CFL conditions

sbastrakov:
For a very long time, CFL checks were wrong for each solver but the standard `Yee`. This PR fixes it for all other solvers:

- `None`
- `Lehe` (thanks @psychocoderHPC for pointing it out recently)
- arbitrary order #3581

I also added a new functor to check those, to provide easier specialization, and fixed a typo in Lewy.
The basic check is what we rely on in current deposition: a particle can't move by more than one cell size in one time step. For `None` and `Lehe` that is enough, Yee and arbitrary-order FDTD pose more strict conditions in their specializations.

Note: it is probably possible to make the arbitrary-order check compile-time as well. However, the `Weights` class there claims to be constexpr, but making a `constexpr` object of it doesn't work. @steindev could you comment on that?

### DIFF
--- a/include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.hpp
+++ b/include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.hpp
@@ -23,11 +23,48 @@
 
 #include "picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.def"
 #include "picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp"
+#include "picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp"
+#include "picongpu/fields/MaxwellSolver/CFLChecker.hpp"
 #include "picongpu/fields/differentiation/Curl.hpp"
 
 #include <cstdint>
+#include <stdexcept>
 #include <string>
 
+namespace picongpu
+{
+    namespace fields
+    {
+        namespace maxwellSolver
+        {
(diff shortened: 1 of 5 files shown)


PrometheusPi (approved):
Looks good to me. @sbastrakov you mentioned in the PR's description, that there is a difference between Yee and Lehe, but from the code, I do not see it. (Did I misunderstand your comment?)

sbastrakov:
@PrometheusPi now the `CFLChecker` specialization in `Yee.hpp` applies only for the classical Yee solver (`Yee<>`(. This is a change from the how it used to be inside all `Yee<CurlE, CurlB>`. For the `Lehe` solver the template is not partially specialized, so the general implementation. According to the paper, c * dt < dx is sufficient there, which is exactly what the default one checks (we need it for current deposition as well).

PrometheusPi:
Thanks for the extra information @sbastrakov.

psychocoderHPC (approved):

sbastrakov:
@PrometheusPi @psychocoderHPC can this be merged?

sbastrakov:
TODO for me as discussed on VC. Check the implemented Lehe CFL, since I may have misinterpreted that piece of the paper.

steindev:
> Note: it is probably possible to make the arbitrary-order check compile-time as well. However, the Weights class there claims to be `constexpr`, but making a `constexpr` object of it doesn't work. @steindev could you comment on that?

Hmm I remember that I wanted to have it this way. At least to have the weights calculated at compile time. But I also remember that I had problems doing it the way I originally wanted. Need to have closer if it is still necessary?

sbastrakov:
@steindev I can do it myself, that's just C++ after all. I was just not sure if the class was intended to be `constexpr` but somehow isn't (so seems the case from your reply), or the `constexpr` should be just removed.

PR #3642, opened 2021-06-16
fix signal handling

psychocoderHPC:
With #3633 we introduced signal handling.
@Anton-Le observed deadlocks when using signaling. 
The reason for the deadlocks was that PMacc used MPI collective blocking operations to handle signals, together with blocking operations in PIConGPU plugins or other simulation phases this can result in deadlocks.

This PR provided two interface extensions for PMacc to be able to handle signals correctly.
  - add method `pmacc::Manager::waitFor(T_Functor&&)` to wait for a user given functor.
  - extend interface with `pmacc::CommunucatorMPI::getMPISignalComm()` 
 
Additionally, the signal processing in `pmacc::SimulationHelper` is updated to avoid that the event system is blocked and signals are allowed to be handled even if MPI ranks of the simulation are in different time steps.

### DIFF
--- a/include/pmacc/eventSystem/Manager.tpp
+++ b/include/pmacc/eventSystem/Manager.tpp
@@ -163,6 +163,15 @@ namespace pmacc
         }
     }
 
+    template<typename T_Functor>
+    void Manager::waitFor(T_Functor&& func)
+    {
+        while(!func())
+        {
+            this->execute();
+        }
+    }
+
     inline void Manager::waitForAllTasks()
     {
         while(tasks.size() != 0 || passiveTasks.size() != 0)
--- a/include/pmacc/simulationControl/SimulationHelper.hpp
+++ b/include/pmacc/simulationControl/SimulationHelper.hpp
@@ -138,10 +138,13 @@ namespace pmacc
          */
         virtual void dumpOneStep(uint32_t currentStep)
         {
-            checkSignals(currentStep);
-
             /* trigger notification */
             Environment<DIM>::get().PluginConnector().notifyPlugins(currentStep);
+
+            /* Handle signals after we executed the plugins but before checkpointing, this will result into lower
+             * response latency if we have long running plugins
+             */
+            checkSignals(currentStep);
 
             /* trigger checkpoint notification */
             if(!checkpointPeriod.empty() && pluginSystem::containsStep(seqCheckpointPeriod, currentStep))
@@ -220,10 +223,10 @@ namespace pmacc
             if(useMpiDirect)
                 Environment<>::get().enableMpiDirect();
 
-            init();
-
             // Install a signal handler
             signal::activate();
+
+            init();
 
             // translate checkpointPeriod string into checkpoint intervals
             seqCheckpointPeriod = pluginSystem::toTimeSlice(checkpointPeriod);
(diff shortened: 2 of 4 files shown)


include/pmacc/simulationControl/SimulationHelper.hpp
```diff
@@ -401,41 +404,97 @@ namespace pmacc
...
+                    Environment<DIM>::get().GridController().getCommunicator().getMPISignalComm(),
+                    &signalMPI));
 
                 if(signal::createCheckpoint())
                 {
                     if(output)
-                        std::cout << "SIGNAL: active checkpointing for step " << maxTimestep << std::endl;
-                    // add a new checkpoint
-                    seqCheckpointPeriod.push_back(pluginSystem::TimeSlice(maxTimestep, maxTimestep));
+                        std::cout << "SIGNAL: Received at step " << currentStep << ". Schedule checkpointing. "
```
sbastrakov, line 456: Does it intentionally print to `cout` and not to the `log`?

psychocoderHPC: Yes, all other output in this class is printing to `std::cout` too. For pmacc we have no useful default log level, this is something I would like to change with a logger refactoring.

include/pmacc/eventSystem/Manager.tpp
```diff
@@ -163,6 +163,15 @@ namespace pmacc
...
+    template<typename T_Functor>
+    void Manager::waitFor(T_Functor&& func)
+    {
+        while(!func())
+        {
+            this->execute();
+        }
+    }
+
     inline void Manager::waitForAllTasks()
```
sbastrakov, line 175: This is not really part of this PR, just a note from my previous attempt to make multiple object files. Such methods, implemented outside of the class, should be made inline at declaration (so inside the class), but not here at definition.

include/pmacc/simulationControl/SimulationHelper.hpp
```diff
@@ -401,41 +404,97 @@ namespace pmacc
...
                     if(output)
-                        std::cout << "SIGNAL: stop simulation at step " << maxTimestep << std::endl;
-                    Environment<>::get().SimulationDescription().setRunSteps(maxTimestep);
+                        std::cout << "SIGNAL: Received at step " << currentStep << ". Schedule shutdown." << std::endl;
+                    signalStopSimulation = true;
+                }
+            }
+            /* We will never handle a signal at step zero.
+             * If we received a signal handleSignalAtStep will be set to currentStep + 1 (see above)
+             */
```
sbastrakov, line 469: I am confused, doesn't this comment contradict the first comment of this function? I mean the handling on step 0 part

include/pmacc/eventSystem/Manager.tpp
psychocoderHPC: Ahh yes I can make it `inline`

include/pmacc/simulationControl/SimulationHelper.hpp
psychocoderHPC: I will rephrase it and will separate `handleSignals` from the step zero. We have two phases here. One is seeing that a signal arrives and start the process to find the largest simulation step between MPI ranks and the second phase where we can translate the signal one simulation step after the signal arrived into an explicit action.

psychocoderHPC:
@Anton-Le reported offline that he was able to write checkpoints triggered by signals on JUWELS where it was not working before. So it looks like this PR is solving the deadlocks we observed with the original implementation.

Anton-Le:
I have tested the changes on JUWELS Booster with a simulation case using 128 nodes (512 GPUs) with the radiation plug-in enabled.
Formerly, when the simulation reached the radiation computation stage (with a per-step time of ca. 8 seconds) neither `scancel --signal=USR1 <job id>` nor `scancel --signal=USR1 --ctld <job id>` would result in a checkpoint and **both** cases would result in a hanging job.
This appears to be fixed now since with the proposed changes I can trigger a checkpoint using either of the two commands (but not `scancel --signal=USR1 --batch <job id>`).

_Note_: The tests were performed with the proposed `handleSlurmSignals.sh` wrapper script.

Anton-Le (approved):
As stated in a previous comment, the changes appear to fix the issue for the configuration for which deadlocks could be reliably reporduced: 128 nodes with radiation computation.

sbastrakov (approved):
There were some points above that are not addressed yet. But they are really minor. So I think we can merge the PR if @psychocoderHPC doesn't plan to fix those soon.

Anton-Le:
I would very much welcome the inclusion of these changes since it will make the code more forgiving to runtime estimation errors.

psychocoderHPC:
@sbastrakov I will address your points when I am back t work.

sbastrakov:
No problem, those were purely cosmetic anyways

PR #3643, opened 2021-06-16
Move the basic implementation of FDTD field solver to the new eponymous template

sbastrakov:
Now `Yee` solver refers to only the classic Yee solver. The general solver is `FDTD<CurlE, CurlB>`. It helps to avoid confusion when distuiguishing Yee as a particular second-order scheme, and FDTD as a generala method. Updated comments to reflect it.

**Note to PIConGPU users**:
In the fieldSolver.param, the Yee solver became just `Yee`, not `Yee<>`. It is updated in the `fieldSolver.param` files, but you may have your own ones inside your setups. Also beware when using it via a macro like `PARAM_FIELDSOLVER`.

**Note for review**: github seems to not undestand that contents of the new `FDTD` directory are basically moved from the old `Yee` plus slight renaming. The new Yee directory is just an alias of FDTD with the given curl types, same as Lehe and other solvers.


psychocoderHPC:
Can we remove `<>` from Lehe solver too?

sbastrakov:
@psychocoderHPC as is we can't because it has a default template parameter for the Cherenkov-free direction (Y by default). But I think it makes sense to have more aliases like `LeheX`, `LeheY`, `LeheZ` without <>. What do you think?

sbastrakov:
I am not sure why the HIP tests are failing. There is a warning in [this output](https://gitlab.com/hzdr/crp/picongpu/-/jobs/1352233379) that probably requires a fix. Not sure if that causes it. I can have a look myself, but maybe you know it already @psychocoderHPC ?

psychocoderHPC:
> @psychocoderHPC as is we can't because it has a default template parameter for the Cherenkov-free direction (Y by default). But I think it makes sense to have more aliases like `LeheX`, `LeheY`, `LeheZ` without <>. What do you think?

Yes IMO this makes sense to have aliases `LeheX,...`

psychocoderHPC:
> I am not sure why the HIP tests are failing. There is a warning in [this output](https://gitlab.com/hzdr/crp/picongpu/-/jobs/1352233379) that probably requires a fix. Not sure if that causes it. I can have a look myself, but maybe you know it already @psychocoderHPC ?

I will restart the broken CI jobs. For unknown reasons, HIP PMacc tests failing from time to time. After a restart of the CI node it always works again. I was not able to reproduce the issue on any other system, IMO this is an issue with the CI node.

sbastrakov:
From the warning it seems the build is indeed incorrect, as `TEST_DIM` is not matching. Could it be that we don't clean the old build artifacts?

psychocoderHPC:
Please rebase against the dev branch.

sbastrakov:
I think I already did after merging yesterday. But just to be sure, done again

sbastrakov:
So i was wrong and rebased helped indeed.

psychocoderHPC (approved):

PR #3645, opened 2021-06-17
Move all field absorber compile-time parameters to fieldAbsorber.param

sbastrakov:
This includes the old `pml.param`, and setting for exponential absorber in the `grid.param`. Unify the setting for absorber thickness. Modify all setups accordingly. It also allowed some simplification of the absorber implementation internals.

We discussed it on VC yesterday. Afterwards I figured it's better to first unify .param files, and then change the defaults.

**Note to users**. If you used non-default absorber parameters, you will need to update the setup to work with `dev`. However, afaik it was almost never non-default, in this case it will be fine with the default `fieldAbsorber.param` as that copies the old defaults.

- [x] I do not think it is interdependent with #3643, but just to make sure will wait for that one first.


psychocoderHPC:
You can rebase this PR.

sbastrakov:
@psychocoderHPC done and CI passed

psychocoderHPC (approved):

PR #3646, opened 2021-06-17
Add fieldAbsorber runtime parameter to TBG_macros.cfg

steindev:
Follow-up to #3635 were we forgot to add it.


sbastrakov (approved):
Ah, thanks for fixing! I forgot it indeed.

PR #3648, opened 2021-06-18
refactor field update functor

psychocoderHPC:
This PR refactors the field update, this removes the requirements that the update kernel required stencil information on the host side we only need within the kernel.
The new update kernel could be removed by a simple `ForEach` kernel with support of cached fields/Boxes but since we do not have it in pmacc or PIConGPU the explicit implementation is required.

- Make the stencil functor for field update more descriptive.
  - The functor is providing the required stencil width information.
- Remove the duplicated implementation of `KernelUpdateE` and `KernelUpdateB`

- [x] rebase against #3644  (therefore I stopped the CI by hand)
- [x] (Sergei) I think it will cause merge conflicts with #3643 (not hard to fix, just won't auto-merge)

### DIFF
--- a/include/picongpu/fields/MaxwellSolver/FDTD/FDTD.kernel
+++ b/include/picongpu/fields/MaxwellSolver/FDTD/FDTD.kernel
@@ -22,6 +22,7 @@
 #include "picongpu/simulation_defines.hpp"
 
 #include <pmacc/algorithms/math/floatMath/floatingPoint.tpp>
+#include <pmacc/dimensions/SuperCellDescription.hpp>
 #include <pmacc/lockstep.hpp>
 #include <pmacc/mappings/threads/ThreadCollective.hpp>
 #include <pmacc/math/operation.hpp>
@@ -47,6 +48,11 @@ namespace picongpu
                 class UpdateEFunctor
                 {
                 public:
+                    //! Stencil requirements for lower margins
+                    using LowerMargin = typename traits::GetLowerMargin<T_CurlB>::type;
+                    //! Stencil requirements for upper margins
+                    using UpperMargin = typename traits::GetUpperMargin<T_CurlB>::type;
+
                     /** Update electric field at the given position
                      *
                      * @tparam T_LocalBBox local magnetic field box type
@@ -72,6 +78,7 @@ namespace picongpu
                     }
 
                 private:
+                    // keep curl as member to support stateful types
                     T_CurlB const curl = T_CurlB{};
                 };
 
@@ -84,6 +91,11 @@ namespace picongpu
                 class UpdateBHalfFunctor
                 {
                 public:
+                    //! Stencil requirements for lower margins
+                    using LowerMargin = typename traits::GetLowerMargin<T_CurlE>::type;
+                    //! Stencil requirements for upper margins
+                    using UpperMargin = typename traits::GetUpperMargin<T_CurlE>::type;
+
                     /** Update magnetic field at the given position
                      *
                      * @tparam T_LocalEBox local electric field box type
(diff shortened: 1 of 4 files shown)


include/picongpu/fields/MaxwellSolver/Yee/Yee.kernel
```diff
@@ -111,127 +122,75 @@ namespace picongpu
...
                     /** Update the electric field using the given functor
                      *
                      * @tparam T_Acc alpaka accelerator type
                      * @tparam T_Mapping mapper functor type
-                     * @tparam T_UpdateFunctor update functor type
-                     * @tparam T_BBox pmacc::DataBox, magnetic field box type
-                     * @tparam T_EBox pmacc::DataBox, electric field box type
+                     * @tparam T_StencilFunctor stencil functor type to update a cell
+                     * @tparam T_SrcBox pmacc::DataBox, source field box type
+                     * @tparam T_DestBox pmacc::DataBox, electric field box type
```
psychocoderHPC, line 138: should be `destination field box`

include/picongpu/fields/MaxwellSolver/Yee/Yee.kernel
```diff
@@ -111,127 +122,75 @@ namespace picongpu
...
-                        auto const workerIdx = cupla::threadIdx(acc).x;
                         pmacc::math::operation::Assign assign;
-                        auto fieldEBlock = fieldE.shift(blockBeginIdx);
-                        ThreadCollective<T_BlockDescription, numWorkers> collectiveCacheE(workerIdx);
-                        auto cachedE = CachedBox::create<0u, typename T_EBox::ValueType>(acc, T_BlockDescription());
-                        collectiveCacheE(acc, assign, cachedE, fieldEBlock);
+                        auto srcFieldBlock = srcField.shift(superCellIdx);
+                        ThreadCollective<StencilCfg, numWorkers> collectiveCacheB(workerIdx);
+                        auto cachedSrcField = CachedBox::create<0u, typename T_SrcBox::ValueType>(acc, StencilCfg{});
+                        collectiveCacheB(acc, assign, cachedSrcField, srcFieldBlock);
```
sbastrakov, line 180: `collectiveCacheB`'s name should be generalized

include/picongpu/fields/MaxwellSolver/Yee/Yee.kernel
```diff
@@ -47,6 +48,11 @@ namespace picongpu
                 class UpdateEFunctor
                 {
                 public:
+                    //! Stencil requirements for lower margins
+                    using LowerMargin = typename traits::GetLowerMargin<T_CurlB>::type;
+                    //! Stencil requirements for upper margins
+                    using UpperMargin = typename traits::GetUpperMargin<T_CurlB>::type;
+
```
sbastrakov, line 55: I think it is worth to add a base class for stencils that takes `Curl` as template, and defines those margins. Besides removing the duplication of those, it would allow clarifying stencil requirements in the docs for the update field kernel.

sbastrakov (commented):
Very nice! I was thinking of something similar. Some small points above

include/picongpu/fields/MaxwellSolver/Yee/Yee.kernel
```diff
@@ -111,127 +122,75 @@ namespace picongpu
...
+                        T_StencilFunctor stencilFunctor,
+                        T_SrcBox const srcField,
+                        T_DestBox destField) const
                     {
-                        /* Each block processes grid values in a supercell,
+                        /* Each block processes all cells of a supercell,
                          * the index includes guards, same as all indices in this kernel
                          */
-                        auto const blockBeginIdx = mapper.getSuperCellIndex(DataSpace<simDim>(cupla::blockIdx(acc)))
+                        auto const superCellIdx = mapper.getSuperCellIndex(DataSpace<simDim>(cupla::blockIdx(acc)))
```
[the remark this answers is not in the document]
psychocoderHPC: But it is the index of the supercell, therefore I renamed it. Block index is wrong because it is `block index  + guard supercells`. This means the alpaka block index is already mapped into the super cell index domain.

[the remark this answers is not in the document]
sbastrakov: It is not. Index of the supercell is the first multiplier. It is then multiplied by the supercell size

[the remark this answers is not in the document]
sbastrakov: I think you introduced this bug in the PR, see how it was defined before

include/picongpu/fields/MaxwellSolver/Yee/Yee.kernel
psychocoderHPC: Each class functor has an implementation and if the stencil would use the curl plus something else the margin would be `max(curl, otherStancilOperator)`. You are right currently all implementations looks equal but only because the functor implementation is using one stancil. I do not think it helps to implement somewhere a class which collects upper and lower margin. The current definition of the margins via in-class definition is only one way, I could also specialize the trade for margins, in this case lower and upper margin would be independent too.

sbastrakov: Not sure if I agree. With how this part now functions, it can only solve equations in form Field += Curl(another field). So curl is the only parameter here kinda. If we wanted to support another structure of equations to solve, that would require modifications anyways.

My motivation to have a base class is as follows:

- More importanly, having a base class with a telling name allows using this name in the doxygen comments. And it makes it searchable in the code base.
- Less importantly, to remove duplication

include/picongpu/fields/absorber/pml/Pml.kernel
```diff
@@ -237,6 +237,11 @@ namespace picongpu
                 class UpdateEFunctor
                 {
                 public:
+                    //! Stencil requirements for lower margins
+                    using LowerMargin = typename traits::GetLowerMargin<T_CurlB>::type;
```
psychocoderHPC, line 241: Could it be that using the margins of `T_CurlB` is larger than we need for the PMLs?
Maybe the requirements for upper or lower margins can be reduced?
This question is coming up because you question if we can provide a unified class which provides upper and lower margin within one unified class based on curl.

include/picongpu/fields/MaxwellSolver/Yee/Yee.kernel
sbastrakov: That said, chances are this is not most actively changed part of the code, after the ongoing modernization is finished. So i don't insist on that.

psychocoderHPC: Where would you place this unified class and how would you name it. Yee functors and PML functors would use this class, for me, it sounds like we mix something. 
Could you maybe open a PR against my PR, maybe I understand something wrong and it gets clear when I see your intention in code.

include/picongpu/fields/absorber/pml/Pml.kernel
sbastrakov: I partly clarified my point in another thread at the same time. Briefly: with the current implementation it is not possible. And even if PML did same as exponential daming, so point-wise uptates, the functor also handles the non-absorbing area. So even then the requirements would remain the same.

include/picongpu/fields/MaxwellSolver/Yee/Yee.kernel
sbastrakov: E.g. `StelcilFunctor`. Note that now (as in your PR currently) it's not easily possible to find implementations of this functor by looking on the update field kernel. None of the functors have "stencil" at their name, nor their main doxygen string (only inside the class).

sbastrakov: Yes, perhaps I did not express my idea well. Will do, or discuss in mattermost for more simple communication

psychocoderHPC:
rebase and on the way to pass the CI

sbastrakov:
@psychocoderHPC CI failed. I guess for some random reason. I am a bit confused why I can't restart it myself. Also, we would need some other reviewer as I am a co-author of your second commit.

psychocoderHPC:
> @psychocoderHPC CI failed. I guess for some random reason. I am a bit confused why I can't restart it myself. Also, we would need some other reviewer as I am a co-author of your second commit.

IMO it is fine that you are the reviewer. You showed your review comment as pull request that I can understand your few point better.

sbastrakov (dismissed):

psychocoderHPC:
For unknown reason clang 11.1 + CUDA 11.2 is failing (internal compiler error). 
In general, it is an unsupported combination because clang 11.1 is not supporting CUDA 11.2 but it worked before. If I can not solve the issue I will disable the combination in our CI.

psychocoderHPC:
Solved all CI issues, CI should now pass all tests

sbastrakov:
One CI job still fails.

To reiterate our discussion from yesterday, if that is what causing the issue. Curls have an internal state, it is initialized in its default (and only) constructor and does not change later. Curls are supposed to be constructible and useable on both host and device. Making a new curl each time should be the same as reusing the curl object results-wise, performance-wise it is same or better to reuse (how `dev` branch does it).

psychocoderHPC:
> One CI job still fails.
> 
> To reiterate our discussion from yesterday, if that is what causing the issue. Curls have an internal state, it is initialized in its default (and only) constructor and does not change later. Curls are supposed to be constructible and useable on both host and device. Making a new curl each time should be the same as reusing the curl object results-wise, performance-wise it is same or better to reuse (how `dev` branch does it).

I keep the way where curl can be stateful, for an unknown reason the IC fails with the same error as before. 
My test from yesterday passed all, wondering why it is now failing, I will review my changes.

include/picongpu/fields/MaxwellSolver/FDTD/FDTD.kernel
```diff
@@ -68,20 +71,19 @@ namespace picongpu
                     {
                         constexpr auto c2 = SPEED_OF_LIGHT * SPEED_OF_LIGHT;
                         constexpr auto dt = DELTA_T;
+                        T_CurlB const curl = T_CurlB{};
```
psychocoderHPC, line 74: ohh this should not be in, maybe my rebase went wrong

psychocoderHPC:
:fireworks: CI passed

sbastrakov (approved):

PR #3649, opened 2021-06-18
pmacc::ConstVector: Add omp declare target around Name

jkelling:
Because an instance of this consexpr variabe is required  at runtime, it needs to be explicitly mapped to be compliant OpenMP.

This is a fix for the OpenMP target support already present in `ConstVector`. It also includes the same mapping for OpenACC, which may be requires by the OpenACC standart at the moment.


psychocoderHPC:
I stopped the CI for this job because #3650 must be merged first and then we need to rebase all open PRs.

psychocoderHPC:
could you please rebase against the dev branch again, we fixed some CI issues.

psychocoderHPC (approved):

PR #3655, opened 2021-06-23
Replace FromHDF5Impl density with FromOpenPMDImpl

### LINKED ISSUE #3515: libSplash still used in HDF5 density profile
While investigating #3514, I have seen hat the (seldomly used) hdf5 density profile method still uses libSplash instead of openPMD-api. 

This should be changed before the next release. @psychocoderHPC Or is this method already outdated (and shall be removed)?

sbastrakov:
It removes this dependency on `libSplash` (and so raw HDF5).

While working on that, I also found the general idea of this density a bit weird. It supports custom iterations for no apparent reason (which I kept, but i think customization on iteration and dataset name is unnecessary). Then it also tries to make use of the moving window, which in my opinion is not logical. If we are inside an iteration, I find it much more logical to just work in the global coordinates. It also makes the code of this part much easier. This is what I implemented in the reworked implementation. So it now operates kinda like `FromPNG` in @psychocoderHPC 's branch. I also removed the default value as it also feels not logical to me. Please tell me if there is something I missed with this decision. Of course, that is not tied with libSplash, so this change can be rolled back.

cc @n01r , see the changes inside `density.param` for requirements to the input file.

Resolves #3515 .


include/picongpu/particles/densityProfiles/profiles.def
```diff
@@ -21,7 +21,7 @@
 
 #include "picongpu/particles/densityProfiles/EveryNthCellImpl.def"
 #include "picongpu/particles/densityProfiles/FreeFormulaImpl.def"
-#include "picongpu/particles/densityProfiles/FromHDF5Impl.def"
+#include "picongpu/particles/densityProfiles/FromOpenPMDImpl.def"
```
sbastrakov, line 24: @psychocoderHPC how we do it now, the declaration in the `.def` file is always available, and the definition is only included when `ENABLE_OPENPMD` is true. Do we want to keep it this way, or to also conditionally enable the declaration? I am for the latter, as now it may be confusing for a user that the template parameters are not right, and that's why the implementation is not found (happened to me).

psychocoderHPC:
@sbastrakov If you say it works like my FromPNG does it mean it is only using a slice of data and repeats it in z-direction? Or is it supporting 3 dimentional volumes like the HDF5 density before?

sbastrakov:
@psychocoderHPC sorry for confusion. No, it is full 3d. What i meant is that it ignores moving window and so operates as if the density in the file is indexed in the global coordinates, not total. The original HDF5 implementation made a transformation, so it did consider the moving window. However, I do not understand the use of it, since we have a fixed iteration and so a user can always provide the right data

include/picongpu/particles/densityProfiles/profiles.def
psychocoderHPC: def file should be not avaiable or empty (contains guards) if the dependency is missing.
Currently we guard the include but mybe it is better to have guards within the files.

psychocoderHPC:
> @psychocoderHPC sorry for confusion. No, it is full 3d. What i meant is that it ignores moving window and so operates as if the density in the file is indexed in the global coordinates, not total. The original HDF5 implementation made a transformation, so it did consider the moving window. However, I do not understand the use of it, since we have a fixed iteration and so a user can always provide the right data

But do you support OpenPMD data larger then the global domain. So that we can use it for moving window?

include/picongpu/particles/densityProfiles/profiles.def
sbastrakov: Okay. So I will add guards inside the `.def` file so that the declaration is only available when the dependency is there. I think this way is slightly better as the code including the file doesn't have to know about its internal dependencies.

sbastrakov:
@psychocoderHPC yes, we can do that. However I do not see any purpose in it. When would that be useful? By design, this all works only for the given time iteration (set in `density.param`). It's not like a user can define density along the moving window and get different data on different calls. This thing seems to be designed to only be called once.

psychocoderHPC:
It is for the user much easier to provide one file with the complete density profile. The user do not know when the simulation is sliding, this depends on the simulation size and we do not give the user any formular to calculate slide points.
The profile dataset should be as large to contain the profile and the the functor should slice out whats reqired for the current time step.
If all data must be provided within the cell domain it is fine but the profile should jot be split in multiple files.

psychocoderHPC:
You can remove the time step, this was a requirement from the splash data format. It make no sense to have this but it was mendentory in splash.

sbastrakov:
I think time iteration is required as openPMD format only operates with time steps afaik. To clarify, it is for the file, not for when the density is calculated.

sbastrakov:
So wrt moving window do you want the following? A user file provides always enough data. Data is indexed in total domain. When PIConGPU instantiates that density functor, the functor loads the part of the file according to how the domain is currently.

psychocoderHPC:
> So wrt moving window do you want the following? A user file provides always enough data. Data is indexed in total domain. When PIConGPU instantiates that density functor, the functor loads the part of the file according to how the domain is currently.

Yes one file contains the full density. The density donain is allowed to be larger or smaller then the simulation domain.
In case it is larger only the visible part is used if it is smaller than all available data is used all other is set to zero.

n01r:
But then it would also be really useful to give an iteration either in the spatial or the temporal domain at which to load in a density.
If you consider the following case: Doing an LWFA simulation in a long channel with a "magic" (underresolved, but anyway) foil somewhere along the propagation path.
One would want to specify the gas density as a mere density profile that gets procedurally generated and the foil could be inserted with a file that is only defined where the foil density is. otherwise, if you have a millimeter long channel you could end up with terabytes of input. The only thing missing then is a switch that says if you want only the file to provide the density in this area or if you want additional particle density from the functor.

Of course, this feature could be a separate PR and it would be fine for now to reproduce the existing functionality in openPMD, with the improvements that you already made.

sbastrakov:
@n01r I feel this can be just solved with a single additional spatial offset? Combination with a custom functor can already be some by having two separate `CreateDensity`

n01r:
> I feel this can be just solved with a single additional spatial offset for the file?

Definitely! I think that would increase the usability and save memory

sbastrakov:
I implemented the total coordinates thing that @psychocoderHPC suggested, and added the offset as @n01r suggested.

include/picongpu/param/density.param
```diff
@@ -223,24 +223,32 @@ namespace picongpu
...
-            /* simulation step*/
-            (PMACC_C_VALUE(uint32_t, iteration, 0))(
-                PMACC_C_VALUE(float_X, defaultDensity, 0.0))); /* struct FromHDF5Param */
+            //! Iteration inside the file (only file, not related to the current simulation time iteration)
+            (PMACC_C_VALUE(uint32_t, iteration, 0))
 
-        /* definition of cloud profile */
-        using FromHDF5 = FromHDF5Impl<FromHDF5Param>;
+            //! Offset in the file in cells: each value is density at (total cell index - offset)
+            (PMACC_C_VECTOR_DIM(int, simDim, offset, 0, 0, 0))); /* struct FromOpenPMDParam */
```
sbastrakov, line 248: I am not sure how to write that in a way that works for 2d and 3d

include/picongpu/particles/densityProfiles/FromOpenPMDImpl.hpp
```diff
@@ -0,0 +1,171 @@
...
+                {
+                    pmacc::DataSpace<simDim> idx;
+                    auto tmpLinearIdx = linearIdx;
+                    for(int32_t d = simDim - 1; d >= 0; d--)
+                    {
+                        idx[d] = tmpLinearIdx % chunkExtent[d];
+                        tmpLinearIdx /= chunkExtent[d];
+                    }
+                    hostDataBox(idx) = rawData[linearIdx];
+                }
```
sbastrakov, line 152: @psychocoderHPC do you know a better way to write it? Since we no longer assume that the chunk is same size as the buffer, it seems I can't make good use of D1 access

include/picongpu/param/density.param
psychocoderHPC: This is automatically working for 2D and 3D because it is using `simDim`. In the code this objects looks like a `Vector<int,simDim>`.

include/picongpu/particles/densityProfiles/FromOpenPMDImpl.hpp
psychocoderHPC: It would be possible to use a 1D DataBox with the size of the loaded data if the loaded data is smaller than the local volume or with the size of the local volume if the OpenPMD data is larger.

psychocoderHPC: This should be what you need:

```
/* get the databox of the host buffer */
auto dataBox = fieldBuffer.getHostBuffer().getDataBox();
/* get a 1D access object to the databox */
using D1Box = DataBoxDim1Access<typename FieldTmp::DataBoxType>;
DataSpace<simDim> guards = fieldBuffer.getGridLayout().getGuard();


DataSpace<simDim> accessSpace;
for(int32_t d = 0; d < simDim; ++d)
{
    // chunkExtent <= localDomain.size is fulfilled by the data loading section
    accessSpace[d] = chunkExtent[d];
}

D1Box d1RAccess(dataBox.shift(guards), accessSpace);

/* copy from temporary buffer to fieldTmp host buffer */
for(uint32_t linearIdx = 0u; linearIdx < numElements; linearIdx++)
{
    d1RAccess[linearIdx].x() = rawData[linearIdx];
}
```

include/picongpu/particles/densityProfiles/FromOpenPMDImpl.hpp
```diff
@@ -0,0 +1,171 @@
...
+                auto const datasetExtent = dataset.getExtent();
+
+                // Offset of the local domain in file coordinates: global coordinates, no guards, no moving window
+                auto const& localDomain = Environment<simDim>::get().SubGrid().getLocalDomain();
+                bool readFromFile = true;
+                auto chunkOffset = ::openPMD::Offset(simDim, 0);
+                auto chunkExtent = ::openPMD::Extent(simDim, 0);
+                for(uint32_t d = 0; d < simDim; ++d)
+                {
+                    chunkOffset[d] = totalLocalDomainOffset[d] - ParamClass::offset[d];
```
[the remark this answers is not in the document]
psychocoderHPC: `chunkOffset` is used in line `129` and is maybe very large.

include/picongpu/param/density.param
```diff
@@ -223,24 +223,35 @@ namespace picongpu
...
-            /* simulation step*/
-            (PMACC_C_VALUE(uint32_t, iteration, 0))(
-                PMACC_C_VALUE(float_X, defaultDensity, 0.0))); /* struct FromHDF5Param */
+            //! Iteration inside the file (only file, not related to the current simulation time iteration)
+            (PMACC_C_VALUE(uint32_t, iteration, 0))
 
-        /* definition of cloud profile */
-        using FromHDF5 = FromHDF5Impl<FromHDF5Param>;
+            //! Offset in the file in cells: each value is density at (total cell index - offset)
+            (PMACC_C_VECTOR_DIM(int, simDim, offset, 0, 0, 0))
```
psychocoderHPC, line 248: Is this `offset` shifting the density defined in the file to the `right` or skipping data within the file?

psychocoderHPC (changes requested):

include/picongpu/param/density.param
psychocoderHPC: IMO if I use a negative offset my data will be loaded late in the simulation so I will add zeros in front of what I have defined in the file and if the offset is positive it is skipping/ignoring the first cells in the file.

Is this correct?

include/picongpu/particles/densityProfiles/FromOpenPMDImpl.hpp
psychocoderHPC: It could be that in the line `D1Box d1RAccess(dataBox.shift(guards), accessSpace);` `ParamClass::offset` must be taken into account if the data from the file is describing a density shifted into positive direction. 

Maybe it must be `D1Box d1RAccess(dataBox.shift(guards + ParamClass::offset), accessSpace);` 
I am not sure about the definition of `ParamClass::offset` therefore I am confused if we need to use `ParamClass::offset` if it is positive or egate.

include/picongpu/param/density.param
sbastrakov: It just shifts the indexing. So before I added this variable, the values in the file were total cell index. And now they are (total cell index - offset). You are right that the padding should be working for this case as well, I didn't think of that.

include/picongpu/particles/densityProfiles/FromOpenPMDImpl.hpp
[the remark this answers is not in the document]
sbastrakov: Thanks, I missed that indeed

include/picongpu/param/density.param
sbastrakov: Fixed now

sbastrakov:
Ready for review.

include/picongpu/param/density.param
sbastrakov: Yes, but how there are three values `0`.

include/picongpu/particles/densityProfiles/FromOpenPMDImpl.hpp
psychocoderHPC: @sbastrakov  could you address this too. IMO it would be easier to read if it is a simple loop without index transformations

include/picongpu/particles/densityProfiles/FromOpenPMDImpl.hpp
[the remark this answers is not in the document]
psychocoderHPC: I still do not understand whats happen if `ParamClass::offset[d]` is e.g. `100` and `totalLocalDomainOffset[d]` zero.
IN this case, you would read in line 129 from `[-100;-100+size)`

include/picongpu/param/density.param
psychocoderHPC: Are positive and negative shifts allowed? IMO the offset is not correctly handled in the implementation.

include/picongpu/param/density.param
psychocoderHPC: IN case `simDim` is `2` the last zero is ignored.

psychocoderHPC:
IMO there is still a bug in the code: https://github.com/ComputationalRadiationPhysics/picongpu/pull/3655#discussion_r662387090

include/picongpu/particles/densityProfiles/FromOpenPMDImpl.hpp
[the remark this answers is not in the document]
sbastrakov: It was like that in my original code. I have fixed it now.

include/picongpu/particles/densityProfiles/FromOpenPMDImpl.hpp
sbastrakov: I do not think your suggestion works. The buffers are of different sizes

include/picongpu/particles/densityProfiles/FromOpenPMDImpl.hpp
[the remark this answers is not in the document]
psychocoderHPC: Could you please point to the line where `chunkOffset` is avoided to be negative. I do not see in the code where this is solved.

include/picongpu/particles/densityProfiles/FromOpenPMDImpl.hpp
psychocoderHPC: It is no problem if the buffer has different extents bit I can add this place of code also after your PR.

include/picongpu/param/density.param
psychocoderHPC: @sbastrakov You missed this comment!

sbastrakov: Pushed the updated version that I forgot to push earlier.

psychocoderHPC (approved):

PR #3660, opened 2021-06-25
Ensure variances are >= 0

jkelling:
In `Estimate.kernel`, if the temperature drops to zero (particles are initialized with zero momentum variance), floating point math makes the variances become small, but sometimes negative inaccurate output.

This patch adds a `max` in `Estimate.kernel` to keep the variances from becoming negative.

```diff
< PIConGPUVerbose PHYSICS(1) | 0 (0 %) supercells had local Debye length estimate not resolved by a single cell
< PIConGPUVerbose PHYSICS(1) | Estimated weighted average temperature -7.58771e-11 KeV and corresponding Debye length -nan m.
<    The grid has -nan cells per average Debye length
---
> PIConGPUVerbose PHYSICS(1) | 1536 (100 %) supercells had local Debye length estimate not resolved by a single cell
> PIConGPUVerbose PHYSICS(1) | Estimated weighted average temperature 0 keV and corresponding Debye length 0 m.
>    The grid has 0 cells per average Debye length
```

Are there other places where variance are calculated which may deserve the same treatment?

### DIFF
--- a/include/picongpu/particles/debyeLength/Check.hpp
+++ b/include/picongpu/particles/debyeLength/Check.hpp
@@ -154,7 +154,7 @@ namespace picongpu
                                 "%1% (%2% %3%) supercells had local Debye length estimate not resolved"
                                 " by a single cell")
                                 % estimate.numFailingSupercells % ratioFailingSupercells % "%";
-                            log<picLog::PHYSICS>("Estimated weighted average temperature %1% KeV and corresponding "
+                            log<picLog::PHYSICS>("Estimated weighted average temperature %1% keV and corresponding "
                                                  "Debye length %2% m.\n"
                                                  "   The grid has %3% cells per average Debye length")
                                 % temperatureKeV % debyeLengthSI % cellsPerDebyeLength;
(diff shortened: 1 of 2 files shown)


sbastrakov (changes requested):
Indeed, that is a thing I didn't think of when writing this code. Thanks for fixing! There is a formatting issue, please fix following [this](https://picongpu.readthedocs.io/en/latest/dev/styleguide.html?highlight=xargs#format-all-files).

I am not sure we calculate variances anywhere else

jkelling:
> Indeed, that is a thing I didn't think of when writing this code. Thanks for fixing! There is a formatting issue, please fix following [this](https://picongpu.readthedocs.io/en/latest/dev/styleguide.html?highlight=xargs#format-all-files).

Fixed.

jkelling:
I added another commit fixing the spelling of the prefix "kilo" (`K` -> `k`) in output and a comment. This does only occur in the `debyeLength` area, so I would like to bunch in with this PR and no open a new one.

sbastrakov (approved):
I think both `keV` and `KeV` are used, but don't mind your change.

psychocoderHPC (approved):

PR #3661, opened 2021-06-28
Fix particle creation: increase primary thread number

psychocoderHPC:
With #3616 we broke the ionization kernel.

### DIFF
--- a/include/picongpu/particles/creation/creation.kernel
+++ b/include/picongpu/particles/creation/creation.kernel
@@ -155,7 +155,8 @@ namespace picongpu
                      */
                     PMACC_SMEM(acc, newFrameFillLvl, int);
 
-                    auto onlyMasters = lockstep::makeMaster(workerIdx);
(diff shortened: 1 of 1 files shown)


n01r:
This would happen during Bremsstrahlung as well because both methods use the `creation.kernel`, right?
We should adjust the title of the PR to *Fix <exact problem> during particle creation*

psychocoderHPC:
> This would happen during Bremsstrahlung as well because both methods use the `creation.kernel`, right?
> We should adjust the title of the PR to _Fix during particle creation_

Yes in this case bremsstrahlung is broken too. 
The problem is that two workers preparing the destination frame. The old code used a naming we typically use to say that only one worker is executing code. During the refactoring in #3616 I replaced the code executed with two workers by one worker. 
This PR is changing it back to two workers.

n01r (approved):

sbastrakov (approved):

PR #3665, opened 2021-06-30
Fix kernelName in kernelEvents.hpp

pordyna:
We were basically doing `typeid(typeid(kernelFucntor.name()).name()`. That's one `typeid().name()` too much. This is why `DEBUG_EVENTS` was printing `TaskKernel NSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE` for all slow kernel events.


sbastrakov (commented):
Nice catch!

Please fix the formatting so that the CI accepts it

sbastrakov (approved):

PR #3672, opened 2021-07-08
Make PML default field absorber

sbastrakov:
To enable the smooth transition, this PR also adds a new `None` absorber type and implementation that does nothing. It is not chooseable by a user directly, but is set when all boundaries are periodic. This is done for the following reasons, all valid only for the all-periodic case:

- compatibility with older checkpoints (otherwise PML would have tried to load its fields, although zero-sized, and that would not work)
- efficiency reasons: as @psychocoderHPC told me earlier, even writing a zero-sized field may take time
- also to lower confusion if someone looks in checkpoint and sees the PML fields

With this change, the setups with absorbing boundaries, default absorber, and particles present in the absorbing area may behave somewhat differently in this area besides just absorption. The reason is, the currents created in the area will behave differently than with the exponential absorber. I think neither the "new" nor the "old" behavior is more correct, since there ideally should not be currents in the absorbing area. If we observe this for some examples, I think we may switch back to exponential for those examples - now it is just a single line in `.cfg` files.

Edit: also the default absorber thickness in `fieldAbsorber.param` is changed to match the recommended range for PML.

### DIFF
--- a/include/picongpu/param/fieldAbsorber.param
+++ b/include/picongpu/param/fieldAbsorber.param
@@ -20,7 +20,7 @@
 /** @file
  *
  * Configure the field absorber parameters.
- * Field absorber type is set by command-line option --fieldAbsorber.kind.
+ * Field absorber type is set by command-line option --fieldAbsorber.
  */
 
 #pragma once
@@ -35,7 +35,7 @@ namespace picongpu
         namespace absorber
         {
             // This variable is not required, defined for convenience
-            constexpr uint32_t THICKNESS = 32;
+            constexpr uint32_t THICKNESS = 12;
 
             /** Thickness of the absorbing layer, in number of cells
              *
--- a/share/picongpu/examples/FoilLCT/include/picongpu/param/fieldAbsorber.param
+++ b/share/picongpu/examples/FoilLCT/include/picongpu/param/fieldAbsorber.param
@@ -20,7 +20,7 @@
 /** @file
  *
  * Configure the field absorber parameters.
- * Field absorber type is set by command-line option --fieldAbsorber.kind.
+ * Field absorber type is set by command-line option --fieldAbsorber.
  */
 
 #pragma once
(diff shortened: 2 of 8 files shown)


include/picongpu/param/fieldAbsorber.param
```diff
@@ -35,7 +35,7 @@ namespace picongpu
         namespace absorber
         {
             // This variable is not required, defined for convenience
-            constexpr uint32_t THICKNESS = 32;
+            constexpr uint32_t THICKNESS = 12;
```
psychocoderHPC, line 38: Why do you change the default absorber thickness?

share/picongpu/examples/FoilLCT/include/picongpu/param/fieldAbsorber.param
```diff
@@ -35,7 +35,7 @@ namespace picongpu
         namespace absorber
         {
             // This variable is not required, defined for convenience
-            constexpr uint32_t THICKNESS = 64;
+            constexpr uint32_t THICKNESS = 16;
```
psychocoderHPC, line 38: Why do you changed the default thickness too?

psychocoderHPC (changes requested):

include/picongpu/param/fieldAbsorber.param
sbastrakov: So that it corresponds to the recommended value for default absorber. For exponential we recommend 32, for PML we recommend 8-12

share/picongpu/examples/FoilLCT/include/picongpu/param/fieldAbsorber.param
sbastrakov: For the same reason. 64 indicates the example wanted a very strong absorption, so I put 16 which is very strong for PML

psychocoderHPC (approved):

psychocoderHPC:
I added the label `input/output` because this change is silently changing all examples. This is not a problem but we need to highlight this in the changelog oo.

PrometheusPi (approved):
Great generalization. Looks good to me.

PrometheusPi:
Should the PR title mention the `None` absorber added to mention this addition in a future change log?

sbastrakov:
@PrometheusPi i think this is not necessary, as this is only internal feature and introduced for technical reasons. A user does not have a "none" kind of absorber

PrometheusPi:
Okay - thanks for the clarification @sbastrakov - I will merge it now.

PR #3674, opened 2021-07-09
openPMD plugin: do not use JSON configurations when restarting

franzpoeschel:
It is currently impossible to specify different plugin parameters for checkpointing and restarting. ADIOS2 reading routines do apparently read the `InitialBufferSize` parameter even when opened in read-only mode. See screenshot below: Memory profile when reading 16 iterations, one time not specifying the `InitialBufferSize`, one time specifying it.
Bildschirmfoto vom 2021-07-09 11-11-05

Since there is currently no useful configuration to use for reading routines when restarting, just don't use JSON configurations at all when reading.


franzpoeschel:
wrong base again..

sbastrakov:
@franzpoeschel FYI the author of a PR can change the target branch after the PR is created, via editing.

PR #3675, opened 2021-07-09
 openPMD plugin: do not use JSON configurations when restarting #3674 

franzpoeschel:
It is currently impossible to specify different plugin parameters for checkpointing and restarting. ADIOS2 reading routines do apparently read the `InitialBufferSize` parameter even when opened in read-only mode. See screenshot below: Memory profile when reading 16 iterations, one time not specifying the `InitialBufferSize`, one time specifying it.
Bildschirmfoto vom 2021-07-09 11-11-05

Since there is currently no useful configuration to use for reading routines when restarting, just don't use JSON configurations at all when reading.

### DIFF
--- a/include/picongpu/plugins/openPMD/openPMDWriter.hpp
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.hpp
@@ -143,8 +143,15 @@ namespace picongpu
                 // avoid deadlock between not finished pmacc tasks and mpi calls in
                 // openPMD
                 __getTransactionEvent().waitForFinished();
-                openPMDSeries
-                    = std::make_unique<::openPMD::Series>(fullName, at, communicator, jsonMatcher->getDefault());
+                openPMDSeries = std::make_unique<::openPMD::Series>(
+                    fullName,
+                    at,
+                    communicator,
+                    /*
+                     * The openPMD plugin only supports configuring writing routines via JSON.
+                     * Reading routines get an empty JSON set.
+                     */
+                    at == ::openPMD::Access::READ_ONLY ? "{}" : jsonMatcher->getDefault());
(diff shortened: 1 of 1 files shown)


PrometheusPi:
@franzpoeschel Thank you for providing this fix that quickly :+1:

sbastrakov:
I am not sure if the "bug" label applies, feel free to change if necessary.

psychocoderHPC (approved):

PrometheusPi (approved):

PrometheusPi:
I compiled the pull request with a previously failing setup on taurus ml and will report back if it solves the issue.

franzpoeschel:
> I am not sure if the "bug" label applies, feel free to change if necessary.

Maybe technically not PIConGPU's fault, but the current behavior leads to crashes where one would not expect any. Calling that a bug is justified in my opinion.

sbastrakov (approved):

PrometheusPi:
I tried running a restart on taurus ml, but now got the following error:
```
Unhandled exception of type 'St13runtime_error' with message 'Chunk does not reside inside dataset (Dimension on index 2 - DS: 481689600 - Chunk: 482651136)', terminating
```

psychocoderHPC:
I will merge this PR, the issue from @PrometheusPi is with high probability triggered by problems.
@PrometheusPi is trying to restart a simulation run that was written with an older dev branch short before we havely refactored the absorbers.

PrometheusPi:
Yes, @psychocoderHPC is right, the error message above originated from a changed absorber default.

franzpoeschel:
Note: ADIOS2 has now read parameters, e.g. for setting the number of threads in decompression. So, we might want to add some functionality into PIConGPU to use this now again.

PR #3679, opened 2021-07-09
[WIP] Prototype proper support for multiple translation units

sbastrakov:
## TL;DR

This PR explores the changes to be made to transition to multiple translation units, without touching most of the code. It is done partially hacky and not intended to be eventually merged as is, but rather to serve as a prototype and as some concrete approach to discuss. If we decide to go this route, the changes can be implemented gradually and without breaking things along the way. I believe most of those make sense even in our current state of things with the single object file, to make the code a little cleaner.

This PR explores what changes need to be done in order to enable proper support for multiple translation units, namely:

- What exactly stops it in the current `dev` from the technical side
- What changes are needed to enable it, from the technical side
- How to make it consistent, so not depending on the only working include order and things like that
- How can we reasonably utilize multiple translation units, given most code is templated

## Current State 

As everyone is aware, PIConGPU had notoriously long compile time relative to its pretty compact code base. This is a combination of two factors

- Most of code is written as class and function templates.
Templates take much longer to compile than normal code. Also, templates normally have to be implemented in header files.
- Almost all code ends up in the same translation unit.
That limits parallelism during compilation, as normally parallel builds can compile different translation units by separate threads. Maybe memory footprint in our gigantic tranlation unit is unreasonably high as well, but I am not that sure about it.

These two things are somewhat connected as class and function templates normally have to be implemented in header files. However, generally it is still possible to have templates in header files, and only include such a header file from the only translation unit needing them. I feel in quite a lot of cases we overuse templates and that is a separate problem. However, this PR only deals with the single-translation-unit issue, without touching most of the code.

The current version compiles the LWFA example on CPU and GPU on the dev machine. However, there are warning about some `inline` function mishaps that need to be addressed ~~and the issue with units needs to be resolved (see the end of the post)~~ (edit: the issue with units is solved).

## What forces single-translation-unit scheme now

I believe there are 3 main issues:

1. Occasinal issues with multiple definitions and incorrect `inline`, which are hidden now due to being in a single translation unit.
PIConGPU used to have lots of those issues, but they have been mostly solved by now. So while such issues still exist, I do not consider them breaking, actually they appear quite straightforward to deal on the case-by-case basis when discovered. So this one appears to be not a serious obstacle.

2. Chaotic interdependencies between files, including circular dependencies
Basically, almost every file includes `simulation_defines.hpp`, and is also eventually included inside there. The most troublesome wrt this PR are `.param` files including (directly or transitively) not just `.def` files but also `.hpp` ones. It is somewhat of a problem regardless, as now one has to know which `.hpp` files are safe and which aren't. Again, it works now due to everything being included on top of `main.cpp` and only instantiated from there. However, this is a big issue wrt going to multiple translation units, as then many changes cause a "chain reaction" that breaks things down.

3. ["Special" `.tpp` files](https://github.com/ComputationalRadiationPhysics/picongpu/blob/5178dcdb08a5e094fb960e66e116322b883a1d8d/include/picongpu/simulation/control/Simulation.hpp#L736-L742) that have to be included in the end.
I believe this is a way of coping with the previous issue. The problem is, again, upon moving some parts to another translation unit the hidden issues start appear. So most of those would need to be moved alltogether. And this PR is a prototype of how I think we can approach it.

## Approach in this PR

Basically, those ["special" `.tpp` files](https://github.com/ComputationalRadiationPhysics/picongpu/blob/5178dcdb08a5e094fb960e66e116322b883a1d8d/include/picongpu/simulation/control/Simulation.hpp#L736-L742) would always be causing trouble. So they have to be transformed into independent `.cpp` files (and so translation units) first, and mostly alltogether. In the current state I've done it for everything ~~but `FieldTmp`. which should be no different from `FieldJ` in principle.~~ (edit: `FieldTmp` is now also in its own translation unit)

`Bremsstrahlung` and `SynchrotronPhotons` became integrated into respective simulation stages. I believe this is beneficial design-wise anyways. Note that the internals of those implementations are heavy on templates and remained in the same `.hpp` files, those were not changed. However, now they are only included from the new `.cpp` files for those simulation stages, and no longer part of `Simulation.hpp`, and so (not of) `simulation_defines.hpp`. Generally, we can do this pImpl-style implementation in most simulation stages and some plugins, the change there is kinda technical and relatively not dangerous to make.

The field types (`FieldE, B, J, Tmp`) as is now combine the data container logic and a little bit of calculation. The latter is causing a lot of dependencies, as it pulls things like particle-to-field classes and transitively more or less whole PIConGPU. I moved away that computational part (which I believe is a good change by its own), and only left the basic data container part. That allowed to remove dependency from `simulation_defines.hpp`. The only things needed are basically `simDim`, `float_X` and `MappingDesc`. I made a new `simulation_defines_lite.hpp` for only those things. This effectively breaks the circular dependency between the Field classes and everything else, which exists in most of the codebase. I admit, it is workaroundish and not nice, and also fragile. However, our current schema with those .tpp-after-Simulation is equally fragile in my opinion. Maybe there can be some better solution to it. ~~Unit values are also coupled to the whole code as of now, so far I didn't solve it. But at the worst case they can be separated into a trait or something.~~ Edit: units are now treated properly, it turned out not a big issue


PrometheusPi:
Great work @sbastrakov! 
For this prototype, could you give a number of independent translation units created and a time of compilation if used with a high number of parallel threads?

sbastrakov:
@PrometheusPi for the current state I've only added 4 translation units for fields (E, B, J, Tmp) and 2 for the Synchrotron photons and Bremsstrahlung stages. Those are listed in the modified CMakeLists.txt. I've only made those as they are kinda problematic currently and have to be solved together in order to properly enable multiple translation units.

After that is enabled, I believe each simulation stage and non-multi (as multi- are templates) plugin can quite easily become a translation unit. When doing it, we would isolate some parts of the code to be included into only one translation unit. Like how in this PR Bremsstrahlung stuff should be only in that .cpp and not everywhere.

I think only splitting the code like that, and trimming what goes to simulation_defines.hpp (and thus every translation unit) would help speeding up compilation.

But it still doesn't hurt to measure, I will do it after all issues and warnings are resolved. For the reasons described above i do not expect an immediate improvement, but that it enables further splitting and that should eventually help.

PrometheusPi:
No worries/hurry @sbastrakov, I know this is a prototype - I am just interested in these numbers because I believe you are tackling/solving a problem every users has and would like to get a feeling for the improvement to be expected. 
Thanks for the further information :+1:

sbastrakov:
I have no time to finish this one, so will close the PR.

If anyone wants to pick this effort up, I think it makes more sense to start a new branch as this one diverged too long ago. However the description in the PR header is still relevant. Feel free to ping me if more details or discussions are needed. Since that time a few other things were independently fixed, so it would be perhaps easier now, e.g. pmacc now [split some code to more object files](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4260). However while I think the task is now "just technical", it still appears to require a focused and moderate-scale effort, likely not smth to be done in 1-2 days.

PR #3680, opened 2021-07-09
Update summit profile

steindev:
Use latest openpmd-api/0.13.2 and adios2/2.7.1 module.
To do so, first, gcc module has to be downgraded due to module incompatibilities of gcc/9.3.0 with other required modules.
Second, adios1 can not be used anymore as it is incompatible with e4s module (which is on the other hand required for adios2 and openpmd-api).
But adios1 is deprecated anyway.

LaserWakefield example compiles and runs with these changes on summit.

### DIFF
--- a/etc/picongpu/summit-ornl/gpu_picongpu.profile.example
+++ b/etc/picongpu/summit-ornl/gpu_picongpu.profile.example
@@ -18,23 +18,22 @@ export proj=<yourProject>
 #export EDITOR="nano"
 
 # basic environment ###########################################################
-module load gcc/8.1.1
+module load gcc/6.4.0  e4s/21.02  spectrum-mpi/10.3.1.2-20200121
 
 export CC=$(which gcc)
 export CXX=$(which g++)
 
 # required tools and libs
-module load git
-module load cmake/3.18.2
-module load cuda/10.1.243
+module load git/2.29.0
+module load cmake/3.20.2
+module load cuda/10.2.89
 module load boost/1.66.0
 
 # plugins (optional) ##########################################################
-module load ums
-module load ums-aph114
-module load hdf5/1.10.4
-module load adios/1.13.1-py2 c-blosc zfp/0.5.5 sz lz4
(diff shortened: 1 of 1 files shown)


etc/picongpu/summit-ornl/gpu_picongpu.profile.example
```diff
@@ -18,23 +18,22 @@ export proj=<yourProject>
 #export EDITOR="nano"
 
 # basic environment ###########################################################
-module load gcc/8.1.1
+module load gcc/6.4.0  e4s/21.02  spectrum-mpi/10.3.1.2-20200121
```
sbastrakov, line 21: Is there no newer `gcc` compatible with all dependencies we need? I do not know of any particular issues with 6.4.0, just it is 4 years old

PrometheusPi (approved):
Looks good to me, but I agree with @sbastrakov: `gcc` is quite old - what's the issue with it on summit?

etc/picongpu/summit-ornl/gpu_picongpu.profile.example
steindev: Yes, this is exactly the point. The other available option would be gcc/9.3.0. But there is no boost with this gcc available that fits our requirements. I could ask for it though. Will do. But for now we can merge this, can't we?

PrometheusPi: Okay

steindev: On the other hand. I just saw that our readthedocs recommends
```
CUDA 10.1/10.2: Use gcc 5.5 - 8
```
Not sure where this is coming from, but with respect to this I would say that everything is fine the way it is as the only alternative to gcc 6 is gcc 9.

PrometheusPi: Could you ask for a gcc 8?

PR #3682, opened 2021-07-12
change how QuietImpl positions reduced number of particles

pordyna:
This changes how the `QuietImpl` `startPosition` functor positions particles when the number of created macro particles has to be reduced to satisfy `weighting >= MIN_WEGHTING` . The new code matches the previous calculation in `numMacroParticles`.

Now the number of particle rows / layers is reduced along the direction  with the highest amount of rows / layers until the condition is fulfilled. This is how the `numMacroParticles` was calculated before but this adjusted particle spacing was ignored in the operator body.

### DIFF
--- a/include/picongpu/particles/startPosition/QuietImpl.hpp
+++ b/include/picongpu/particles/startPosition/QuietImpl.hpp
@@ -115,9 +115,9 @@ namespace picongpu
                     }
 
                 private:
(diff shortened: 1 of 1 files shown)


sbastrakov:
I've put the "bug" label as the old implementation appears inaccurate in this situation. However, it is more of inaccuracy than a bug

sbastrakov (approved):

psychocoderHPC (approved):

PR #3684, opened 2021-07-13
Extend the comments to FromOpenPMD density parameters

sbastrakov:
Extend the comments to `FromOpenPMD` density profile. Those address the questions @n01r had when trying it out.

### DIFF
--- a/include/picongpu/param/density.param
+++ b/include/picongpu/param/density.param
@@ -1,5 +1,5 @@
 /* Copyright 2013-2021 Axel Huebl, Heiko Burau, Rene Widera, Felix Schmitt,
- *                     Richard Pausch
+ *                     Richard Pausch, Sergei Bastrakov
  *
  * This file is part of PIConGPU.
  *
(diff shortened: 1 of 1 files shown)


include/picongpu/param/density.param
```diff
@@ -1,5 +1,5 @@
 /* Copyright 2013-2021 Axel Huebl, Heiko Burau, Rene Widera, Felix Schmitt,
- *                     Richard Pausch
+ *                     Richard Pausch, Sergei Bastrakov
```
sbastrakov, line 2: I forgot it with my previous PR changing from HDF5 to from openPMD

n01r:
I'll keep trying it out and maybe we can add one or two more things, then :)

include/picongpu/param/density.param
```diff
@@ -235,11 +235,22 @@ namespace picongpu
...
 
-            //! Name of the dataset inside the file
-            (PMACC_C_STRING(datasetName, "density"))
+            /** Path to the openPMD input file
+             *
+             * It is recommended to use a full path to make it independent of how PIConGPU is launched.
+             *
+             * Relative paths require consistency to the current directory when PIConGPU is started.
+             * With tbg and the standard .tpl files, relative to the resulting simOutput directory.
+             */
```
sbastrakov, line 245: Will extend on the file-based iteration format

PR #3687, opened 2021-07-14
use IUnary with filters in collisions and FieldTmp

pordyna:
Without this only particle filters with a default constructor would work. With `IUnary` filters with a `Filter(uint32_t currentStep)` constructor should work as well.


sbastrakov (approved):

psychocoderHPC (approved):

psychocoderHPC:
Thanks for spotting the issue and for the fix. I will not mark it as bug because we had not seen issues with this behavior.

PR #3691, opened 2021-07-16
remove ADIOS1 IO plugin

psychocoderHPC:
As discussed in the PIConGPU developer meeting this PR is removing the native ADIOS1 IO plugin.

- update documentation
- remove all code required to handle ADIOS1 IO
- CI: remove ADIOS1

# Review

I suggest reviewing this PR commit-wise.


docs/TBG_macros.cfg
```diff
@@ -266,25 +244,20 @@ TBG_openPMD="--openPMD.period 100   \
 # Create a checkpoint that is restartable every --checkpoint.period steps
 #   http://git.io/PToFYg
 TBG_checkpoint="--checkpoint.period 1000"
-# Select the backend for the checkpoint, available are openPMD and adios
-#    --checkpoint.backend openPMD
-#                         adios
-# Available backend options are exactly as in --openPMD.* and --adios.* and can be set
+# Select the backend for the checkpoint, available are openPMD
+#    --checkpoint.backend openPMa
```
sbastrakov, line 248: `openPMa` is a typo.

Also, why do we have the backend option at all now? The only value is openPMD. If it's left for compatibility or some reason like that, we should document it. But i would prefer to just remove the option alltogether, we can bring it back if needed in the future

docs/TBG_macros.cfg
```diff
@@ -266,25 +244,20 @@ TBG_openPMD="--openPMD.period 100   \
...
+# Available backend options are exactly as in --openPMD.* and can be set
 # via:
 #   --checkpoint.<IO-backend>.* <value>
 # e.g.:
 #   --checkpoint.adios.compression zlib
```
sbastrakov, line 253: Forgot to change here?

docs/source/usage/plugins/checkpoint.rst
```diff
@@ -21,18 +21,17 @@ What is the format of the created files?
 
 We write our fields and particles in an open markup called :ref:`openPMD <pp-openPMD>`.
 
-For further details, see the according sections in :ref:`the openPMD API <usage-plugins-openPMD>` and :ref:`ADIOS <usage-plugins-ADIOS>`.
+For further details, see the according section in :ref:`the openPMD API <usage-plugins-openPMD>`.
```
sbastrakov, line 24: Should be "For further details, see the :ref:`the openPMD API <usage-plugins-openPMD>` section".

docs/source/usage/plugins/checkpoint.rst
```diff
@@ -21,18 +21,17 @@ What is the format of the created files?
...
 
 External Dependencies
 ^^^^^^^^^^^^^^^^^^^^^
 
-The plugin is available as soon as the :ref:`openPMD API or ADIOS libraries <install-dependencies>` are compiled in.
+The plugin is available as soon as the :ref:`openPMD API library <install-dependencies>` are compiled in.
```
sbastrakov, line 29: In "are compiled in" should be `is`.

sbastrakov (commented):
Some comments above.

I think we should remove the `checkpoint.backend` and rename `checkpoint.backend.X` options to `checkpoint.X`. This can be done in a follow-up PR as well. Your PR is self-consistent

docs/TBG_macros.cfg
psychocoderHPC: This option is optional and is by default `openPMD`. I like to keep this option to have a simpler life if we would introduce something else e.g. sionlib, a vendor library, .... Especially for checkpointing there is no strong need to use `openPMD`.

psychocoderHPC:
> Some comments above.
> 
> I think we should remove the `checkpoint.backend` and rename `checkpoint.backend.X` options to `checkpoint.X`. This can be done in a follow-up PR as well. Your PR is self-consistent

Could you bring up this point in the next developer meeting?

docs/TBG_macros.cfg
ax3l: That's correct. Yet the value in the example has a typo

docs/source/usage/plugins/checkpoint.rst
```diff
@@ -21,18 +21,17 @@ What is the format of the created files?
...
 
 .cfg file
 ^^^^^^^^^
 
 You can use ``--checkpoint.period`` to specify the output period of the created checkpoints.
-Note that this plugin will only be available if the openPMD API, libSplash (HDF5) or ADIOS is found during compile configuration.
```
PrometheusPi, line 35: Why is this note totally gone? Should it not state:

> Note that this plugin will only be available if the openPMD API is found during compile configuration.

PrometheusPi (changes requested):
Looks very good, most issues were already mentioned by @sbastrakov

docs/source/usage/plugins/checkpoint.rst
psychocoderHPC: we have a similar comment two lines above.

psychocoderHPC:
updated

sbastrakov (approved):

psychocoderHPC:
@PrometheusPi can you re-review this PR to remove your blocking review.

docs/source/usage/plugins/openPMD.rst
```diff
@@ -43,7 +43,7 @@ The openPMD API will parse the file name to decide the chosen backend and iterat
 In order to set defaults for these value, two further options control the filename:
 
 * ``--openPMD.ext`` sets the filename extension.
-  Possible extensions include ``bp`` for the ADIOS backends (default), ``h5`` for HDF5 and ``sst`` for Streaming via ADIOS2/SST.
+  Possible extensions include ``bp`` for the ADIOS2 backends (default), ``h5`` for HDF5 and ``sst`` for Streaming via ADIOS2/SST.
```
PrometheusPi, line 46: Is it not possible to write ADIOS(1) output anymore? I thought openPMD-api supports both ADIOS and ADIOS2.

PrometheusPi (approved):

docs/source/usage/plugins/openPMD.rst
sbastrakov: I also thought openPMD API would still use ADIOS1 if that's how it was compiled. Or if both were found, i think there is an environment variable that controlls which one to use for `.bp`

psychocoderHPC: Yes it is still possible. I changed it because we should not recommend using ADIOS1 anymore.
In openPMD ADIOS1 and ADIOS2 are mutually exclusive as I know e.g. if you set in PIConGPU the file extension to `bp` it will use ADIOS2 if possible.

PR #3696, opened 2021-07-19
Add simplified rad form factor

PrometheusPi:
This pull request adds another form factor which is emulates the `radFormFactor_Gauss_spherical` but avoids calling the `math::exp` function. 

The simplification does not perfectly match the original function but is good enough for any practical application. 
For reference, the adjustment looks as follows:
grafik

### DIFF
--- a/docs/source/usage/plugins/radiation.rst
+++ b/docs/source/usage/plugins/radiation.rst
@@ -30,7 +30,7 @@ Variable                       Meaning
 ============================== ================================================================================
 
 Currently this allows to predict the emitted radiation from plasma if it can be described by classical means.
-Not considered are emissions from ionization, Compton scattering or any bremsstrahlung that originate from scattering on scales smaller than the PIC cell size. 
+Not considered are emissions from ionization, Compton scattering or any bremsstrahlung that originate from scattering on scales smaller than the PIC cell size.
 
 External Dependencies
 ^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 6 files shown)


include/picongpu/plugins/radiation/radFormFactor.hpp
```diff
@@ -142,15 +142,41 @@ namespace picongpu
...
+            namespace radFormFactor_Gauss_spherical_simple
+            {
+                struct radFormFactor
+                {
+                    /** Form Factor for point-symmetric Gauss-shaped charge distribution of N discrete electrons:
+                     * \f[ <rho(r)> = N*q_e* 1/sqrt(2*pi*sigma^2) * exp(-0.5 * r^2/sigma^2) \f]
+                     * with sigma = 0.5*c/delta_t (0.5 because sigma is defined around center)
+                     * simplified to avoid calling too often special functions to:
+                     * sqrt( N + (N^2 - N) * F(omega) )
+                     * F(omega) = ( exp(-0.5 * (x/2.0 * delta_t)^2) )^2 ~ 1/(0.03*(delta_t)^2 * x^2 + 1)^12
```
BeyondEspresso, line 160: I would write ``~ 1/(0.025*(delta_t)^2 * x^2 + 1)^12`` as in the implementation further below. Based on which criterion did you choose the factor 0.025? Please document this because otherwise someone will search for meaning later, where there might be none.

BeyondEspresso (changes requested):

include/picongpu/plugins/radiation/radFormFactor.hpp
Anton-Le: Documentation could be improved, although it is "fairly obvious" that the simplification is basically just a series expansion of `exp` (to 4th order).

Anton-Le (dismissed):
The changes look fine to me.

They would eliminate the exponential. Square root is, in principle, also a non-elementary function and would be evaluated using a dedicated unit.

**But** in case of `fast-math` (which is used per default for the GPU, if memory serves) the operation should be replaced by a faster, less precise counterpart.

I originally expanded the expression w.r.t. the exponential argument AND `N`, making it into a pure polynomial.

PrometheusPi:
With your input @BeyondEspresso I had a deeper look into the underlying optimization. 

If we vary the scalar factor alpha and the exponent beta, we get an optimum (for the lowest relative error) as follows:
grafik

Since the previously used expoNent 12 results in the same number of multiplications as an exponent of 16, I decided to shift the paramEter and use:
```
alpha = 0.0172169
beta = 16
```
according to an optimization. 

This results in the following approximation and (in red) relative error:
grafik

include/picongpu/plugins/radiation/radFormFactor.hpp
BeyondEspresso: Well, it is not obvious whether you optimize for the minimum relative error integrated over the full spectrum for a given exponent (as Richard did) or if you do something different. You could integrate the absolute errors instead, try to minimize the maximum relative or absolute deviation or let's say compare the integrals of the form factor functions. All these metrics can be justified for all practical purposes considered here, but it is useful to document the metric, so that one does not have to do any forensic code reproduction later.

PrometheusPi: Since including graphs etc is not easily possible in the code and adding it in the radiation plugin documentation would be an overkill for one specific form factor, I referenced the PR id for those whowants to read about the details of the optimization.

docs/source/usage/plugins/radiation.rst
```diff
@@ -409,8 +410,8 @@ Dataset  Description                                           Dimensions
 
    Please be aware, that despite the fact, that the SI-unit of each amplitude entry is :math:`\mathrm{[\sqrt{Js}]}`, the stored ``unitSI`` attribute returns :math:`\mathrm{[Js]}`.
    This inconsistency will be fixed in the future.
-   Until this inconstincy is resolved, please multiply the datasets with the square root of the ``unitSI`` attribute to convert the amplitudes to SI units. 
-   
+   Until this inconstincy is resolved, please multiply the datasets with the square root of the ``unitSI`` attribute to convert the amplitudes to SI units.
```
sbastrakov, line 413: Typo: `inconstincy`

include/picongpu/plugins/radiation/radFormFactor.hpp
```diff
@@ -142,15 +142,43 @@ namespace picongpu
...
+                     * @param observer_unit_vec = observation direction
+                     * @return the Form Factor: \f$ \sqrt( | \mathcal{F} |^2 ) \f$
+                     */
+                    HDINLINE float_X
+                    operator()(const float_X N, const float_X omega, const vector_X observer_unit_vec) const
+                    {
+                        /* currently a fixed sigma of DELTA_T * c is used to describe the distribution - might become a
+                         * parameter */
+                        // optimized paramter for exponent beta=16 (see picongpu PR #3696 for details):
+                        const float_X alpha = 0.0172169f;
```
sbastrakov, line 173: Can be `constexpr` ?

sbastrakov: Also for `float_X` the suffix is `_X`, not `f` (which is for `float`).

psychocoderHPC: Yes as @sbastrakov said please use `0.0172169_X` to be able to switch PIConGPU's precision to 64bit without float upscale.

sbastrakov (approved):

PrometheusPi:
@sbastrakov found a better optimization that does not rely on simplifications. This should be implemented instead of this simplification. Thus I will close this pull request.

PR #3697, opened 2021-07-19
Ascent (ORNL) setup

PrometheusPi:
This pull request adds the setup for the Ascent (test cluster) at ORNL.

### DIFF
--- a/docs/source/install/profile.rst
+++ b/docs/source/install/profile.rst
@@ -279,3 +279,18 @@ Queue: gpu (2 x NVIDIA Tesla k40m GPUs)
 
 .. literalinclude:: profiles/aris-grnet/gpu_picongpu.profile.example
    :language: bash
+
+Ascent (ORNL)
+-------------
+
+**System overview and user guide:** `link <https://docs.olcf.ornl.gov/systems/ascent_user_guide.html#system-overview/>`_
+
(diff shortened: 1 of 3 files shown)


PrometheusPi:
Should I update the profile to point to our shared libraries? (Which would however only be readable to our project.)

sbastrakov:
I think that would work only during the hackathon? Then probably not worth, and the installation and settings you made at the shared directory were already helpful and sufficient. Having Ascent profile in general still has value, in case some user will have access to it in the future.

PrometheusPi:
@sbastrakov Okay, then it can be reviewed and merged as is. Changed the status from *draft* to *open*.

sbastrakov (approved):

sbastrakov (changes requested):
Please also add it to this doc page: https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/docs/source/install/profile.rst

PrometheusPi:
@sbastrakov I will add a documentation to the `profile.rst` file - thanks for pointing this out :+1: 
Do you know why we keep a copy (not link as far as I can tell) of the profiles in `picongpu/docs/source/install/profiles`? 
Should a add a copy/link as well?

sbastrakov (approved):

PR #3699, opened 2021-07-21
Fix FromOpenPMD density and make it account for axisLabels attribute

### LINKED ISSUE #3685: Change the indexing order in FromOpenPMDImpl density?
When solving #3515 and moving to openPMD, I made the indexing in file to match the PIConGPU indexing. So to use x-y-z order, as I considered natural. However as pointed out by @n01r , PIConGPU output is in the z-y-x order. And since this density often wants to use the output of another PIConGPU simulation (or a postprocessing of it) as input, it is probably better to also take input as z-y-x. Or support both?

sbastrakov:
The original implementation was technically correct, as it worked as documented, so I did not add the bug label. But it was not easy to use in the common case of using output of the previous simulation as input. Now all Cartesian axis orders should be supported properly. Also extend the comments to the parameters.

Edit: the preceeding paragraph concerns only the axis part.
There was also a bug with reading incorrect parts of the file when having multiple MPI ranks. This is unrelated to axes, but also fixed in this PR.

This PR includes the comment improvements from #3684, I will close that one.

Resolves #3685.

### DIFF
--- a/include/picongpu/particles/densityProfiles/FromOpenPMDImpl.hpp
+++ b/include/picongpu/particles/densityProfiles/FromOpenPMDImpl.hpp
@@ -32,10 +32,12 @@
 #    include <pmacc/static_assert.hpp>
 
 #    include <algorithm>
+#    include <array>
 #    include <cstdint>
 #    include <functional>
 #    include <memory>
 #    include <numeric>
+#    include <stdexcept>
 
 #    include <openPMD/openPMD.hpp>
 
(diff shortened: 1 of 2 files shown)


include/picongpu/particles/densityProfiles/FromOpenPMDImpl.hpp
```diff
@@ -166,6 +162,103 @@ namespace picongpu
                 __getTransactionEvent().waitForFinished();
             }
 
+            /** Helper class to convert indices between x-y-z coordinates and a system defined by openPMD API mesh
+             *
+             * The latter is defined by a combination of axisLabels and dataOrder of the given mesh object.
+             * Thus, the transformation is bound to the mesh object given in the constructor, not to openPMD API in
+             * general. However, for brevity in this class it is called "openPMD coordinates"
+             */
+            class IndexConverter
```
psychocoderHPC, line 171: Can we move this class somewhere into `plugin/misc` or somewhere else in PIConGPU that we can use the class everywhere?

sbastrakov: Yes, I can do that. Perhaps then it would need some slight generalization in functionality. I didn't really consider it and just programmed what was needed here (and checked there was nothing really reusable as-is). I think this can maybe become a follow-up PR.

sbastrakov:
@n01r @wmles so hopefully this PR fixes the issues with the from-openPMD density. There was a bug in my original implementation besides only working for the x-y-z order. This PR fixes both. Attaching sample writing files that I used to test the axis orders, it's a simplified version of the file Marco gave me before. Both files are now equivalent in terms of produced output. Added `.txt` at the end as github doesn't allow attaching `.py`.
[write_xyz_density_file.py.txt](https://github.com/ComputationalRadiationPhysics/picongpu/files/6901009/write_xyz_density_file.py.txt)
[write_zyx_density_file.py.txt](https://github.com/ComputationalRadiationPhysics/picongpu/files/6901010/write_zyx_density_file.py.txt)

psychocoderHPC (approved):

PR #3702, opened 2021-07-28
Radiation: Coalescence Write to Global

ax3l:
Fix a warning seen with Nsight Compute today. Since the `jobIdx` for splitting the `numJobs` is the slowest varying index, the order should be inverted.

Follow-up to #3354

Co-authored-by: @PrometheusPi & @sbastrakov

### DIFF
--- a/include/picongpu/plugins/radiation/Radiation.hpp
+++ b/include/picongpu/plugins/radiation/Radiation.hpp
@@ -421,7 +421,7 @@ namespace picongpu
                     for(int resultIdx = 1; resultIdx < numJobs; ++resultIdx)
                         for(int ampIdx = 0; ampIdx < numAmp; ++ampIdx)
                         {
-                            dbox(DataSpace<2>(ampIdx, 0)) += dbox(DataSpace<2>(ampIdx, resultIdx));
+                            dbox(DataSpace<2>(0, ampIdx)) += dbox(DataSpace<2>(resultIdx, ampIdx));
                         }
                 }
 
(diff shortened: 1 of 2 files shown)


ax3l:
Let's check this one works tomorrow.

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -1206,7 +1206,7 @@ namespace picongpu
 
                     // PIC-like kernel call of the radiation kernel
                     PMACC_KERNEL(KernelRadiationParticles<numWorkers>{})
-                    (DataSpace<2>(gridDim_rad, numJobs), DataSpace<2>(numWorkers, 1))(
+                    (DataSpace<2>(numJobs, gridDim_rad), DataSpace<2>(1, numWorkers))(
```
ax3l, line 1209: not entirely sure about this line

psychocoderHPC:
This change will not remove the issue with the memory access, the problem is the aplitude with 6 doubles, the indexing is already fine to provide coalesed access to memory if the data layout fits.

sbastrakov:
I think the indexing was not fine wrt coalescing: `jobIdx` should be slow index there, and it appeared to be the fast one. The 6-double thing you mentioned stays anyways.

psychocoderHPC:
> jobIdx

No the dev is fine. `jobIdx` is the fast index. We use PMacc boxes and access the box with `DataSpace(x,y)` where y is the slow moving index.

ax3l:
Thanks @psychocoderHPC.

Yep, you are right. Since our DataSpace constructor arguments are Fortran-ish :D we already have the right index order in `radiation(DataSpace<2>(jobIdx, theta_idx * radiation_frequencies::N_omega + o))`

Refs.:
- Constructor index order: https://github.com/ComputationalRadiationPhysics/picongpu/blob/454a28efe6a6eb0fa1e0e97031ef2569a1ad313b/include/pmacc/dimensions/DataSpace.hpp#L113
- Box access order: https://github.com/ComputationalRadiationPhysics/picongpu/blob/454a28efe6a6eb0fa1e0e97031ef2569a1ad313b/include/pmacc/memory/boxes/DataBox.hpp#L117

So yet, the warnings that we see are due to the elements (struct of 6 floats) not the index order.

sbastrakov:
Not sure I follow.
`radiation(DataSpace<2>(jobIdx, theta_idx * radiation_frequencies::N_omega + o))` - this is after your PR, not dev.

PR #3694, opened 2021-07-19
Cori: Add DGX (A100) Profile

ax3l:
Add documentation to compile & run on Cori's A100 GPUs.

Preparation: load profile once and build (HDF5,) ADIOS2, openPMD-api & PNGwriter:
```bash
. ~/a100_picongpu.profile
export SW_DIR=/global/cfs/cdirs/ntrain/dgx  # only Axel can write here, use $HOME/sw in doubt

# potentially good idea: get a node via:
getNode 8

mkdir src
cd src

curl -Lo hdf5-1.10.7.tar.gz https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.10/hdf5-1.10.7/src/hdf5-1.10.7.tar.gz
tar -xzf hdf5-1.10.7.tar.gz
cd hdf5-1.10.7
./configure --enable-parallel --enable-shared --prefix $SW_DIR/openmpi/hdf5-1.10.7 CC=mpicc CXX=mpic++
make -j 16
make install
cd ..

git clone https://github.com/ornladios/ADIOS2.git adios2
cd adios2
git checkout v2.7.1
cmake -S . -B build -DADIOS2_USE_Fortran=OFF -DCMAKE_INSTALL_PREFIX=$SW_DIR/openmpi/adios-2.7.1
cmake --build build -j 16
cmake --build build --target install
cd ..

git clone https://github.com/openPMD/openPMD-api.git
cd openPMD-api
git checkout 0.13.4
cmake -S . -B build -DopenPMD_USE_PYTHON=OFF -DCMAKE_INSTALL_PREFIX=$SW_DIR/openmpi/openPMD-api-0.13.4
cmake --build build -j 16
cmake --build build --target install
cd ..

git clone https://github.com/pngwriter/pngwriter.git
cd pngwriter
git checkout dev
cmake -S . -B build -DCMAKE_INSTALL_PREFIX=$SW_DIR/pngwriter-0.7.0-25-g0c30be5
cmake --build build -j 16
cmake --build build --target install
cd ..
```

PR #3708, opened 2021-07-30
Cori DGX: Module purge

ax3l:
Missing in #3694 and recommended by docs.
(Not part of the Cori Cray system software env.)

### DIFF
--- a/etc/picongpu/cori-nersc/a100_picongpu.profile.example
+++ b/etc/picongpu/cori-nersc/a100_picongpu.profile.example
@@ -21,6 +21,7 @@ export RESERVATION=""
 
 # General modules #############################################################
 #
+module purge
 module load dgx
 module swap PrgEnv-intel PrgEnv-gnu
 module load cuda openmpi


etc/picongpu/cori-nersc/a100_picongpu.profile.example
```diff
@@ -21,6 +21,7 @@ export RESERVATION=""
 
 # General modules #############################################################
 #
+module purge
 module load dgx
 module swap PrgEnv-intel PrgEnv-gnu
 module load cuda openmpi
```
ax3l, line 27: There is more to it, but have to hop in a meeting.

Trying
```suggestion
module purge
module load dgx
module load gcc
module load cuda
module load openmpi
```
but I did something else to get it working last time..

psychocoderHPC (approved):

etc/picongpu/cori-nersc/a100_picongpu.profile.example
psychocoderHPC: Should I wait, for the integration of this suggestion, before I merge this PR?

PrometheusPi (approved):
The change looks good - but I have no Cori account, thus I can not test this.

steindev:
@ax3l @psychocoderHPC Close or merge?

psychocoderHPC:
I will merge this PR now. In the worst case, we need to provide a fix as soon as someone has access to the system.

PR #3683, opened 2021-07-12
add radiation cases to SPEC benchmark

PrometheusPi:
This pull request adds radiation cases for the SPEC benchmarks.

PR #3711, opened 2021-08-04
Optimize radiation plugin math

sbastrakov:
This PR is based on the work we did last week during the NERSC hackathon. It is basically a nice way to implement [these changes](https://github.com/sbastrakov/picongpu/tree/topic-hackathonRadiationOptimization).

These changes should not be reducing accuracy. The results are not bitwise-identical due to reordering, but should not be consistently less accurate. Moving to all-float calculation would be of course faster, but that may be a significant loss of accuracy that should be studied before getting applied.

From what we observed at the hackathon, on the benchmark setup #3683 the commits 1 and 2 together (well, almost all must be 2) gave 1.11 x speedup of the radiation kernel. Note that these also make the previously proposed #3696 obsolete. Commit 3 gave further 1.19 x speedup.

Another thing we saw at the hackathon, not covered in this PR, is giving sufficient number of blocks to utilize the whole GPU at as much occupancy as possible, with properly choosing the `--numJobs` parameter.

@PrometheusPi checked the original hacky version and found it to work. This should produce the same results, but I didn't test it yet, as I do not have a ready setup. @PrometheusPi could you try it out?

**For reviewers**: I think it is easier to review commit-by-commit.

### DIFF
--- a/include/picongpu/plugins/radiation/Radiation.kernel
+++ b/include/picongpu/plugins/radiation/Radiation.kernel
@@ -1,5 +1,5 @@
 /* Copyright 2013-2021 Axel Huebl, Heiko Burau, Rene Widera, Richard Pausch,
- *                     Klaus Steiniger, Felix Schmitt, Benjamin Worpitz
+ *                     Klaus Steiniger, Felix Schmitt, Benjamin Worpitz, Sergei Bastrakov
  *
  * This file is part of PIConGPU.
  *
@@ -132,7 +132,7 @@ namespace picongpu
                     PMACC_SMEM(acc, real_amplitude_s, memory::Array<vector_64, blockSize>);
 
                     // retarded time
-                    PMACC_SMEM(acc, t_ret_s, memory::Array<picongpu::float_64, blockSize>);
+                    PMACC_SMEM(acc, t_ret_s, memory::Array<picongpu::float_X, blockSize>);
 
                     // storage for macro particle weighting needed if
                     // the coherent and incoherent radiation of a single
@@ -318,7 +318,7 @@ namespace picongpu
                                                 * particle_charge * picongpu::float_64(DELTA_T);
 
                                             // retarded time stored in shared memory
-                                            t_ret_s[saveParticleAt] = amplitude3.get_t_ret(look);
+                                            t_ret_s[saveParticleAt] = static_cast<float_X>(amplitude3.get_t_ret(look));
 
                                             lowpass_s[saveParticleAt] = NyquistLowPass(look, particle);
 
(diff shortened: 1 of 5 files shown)


PrometheusPi (dismissed):
Looks good to me - please do not merge till run-time tests have past.

PrometheusPi:
The single particle test (bunch example case 5) looks good: 

`dev` version:
grafik

`pr_3711` version:
grafik

absolute difference normalized to peak `dev` value: 
grafik

relative difference (caped to 1%  of peak `dev` value):
grafik

This looks excellent. 
I will post the many particle test asap.

PrometheusPi:
@sbastrakov The job ran into the interactive CI issue #3715 - could you please push again to trigger the CI?

sbastrakov:
@PrometheusPi thanks for checking. Could you also post results of the same comparison you did during the hackathon vs. my hacky branch? I would expect those two branches to provide exactly same result, this one is just with less hacking.

Yes, I will push, thanks for pinging about it

PrometheusPi:
I think this comparison will not reveal any information.
If I compare the V100 `dev` data from ascent to the k20 `dev`  from hemera, I already get differences in the 0.1% regime:
grafik

Comparing your current pull request with your ascent code leads to similar differences:
grafik

PrometheusPi:
The many particle test also looks great:

`dev` version:
grafik


`pr_3711` version:
grafik


absolute error makes no sense since a few pixels have order of magnitude more intensity. 

relative error needs to be caped at 1e-10 times the peak value:
grafik

PrometheusPi:
Caped to 1e-12 times the peal value, the errors are slightly higher, but that's in the "piratically no-intensity region":
grafik

PrometheusPi (dismissed):

sbastrakov:
In my original PR version I failed to update the use of radiation particle form factors in the transition radiation plugin. It turned out an easy fix, I squashed it together with those interface changes. I was not aware the form factors were also used there, in this case they should be moved from inside the radiation plugin. That would be in a separate PR, ofc.

Anton-Le:
As caught by the CI, but I will nonetheless state it here explicitly:

Compilation fails for the `TransitionRadiation.kernel` since `omega` used in the loop over all frequencies is not defined.

```
../picongpu/plugins/transitionRadiation/TransitionRadiation.kernel(295): error: identifier "omega" is undefined
```

sbastrakov:
Sorry it took long to make it pass CI. The issues were unrelated to the changes in the radiation plugin, but how those are used in the transition radiation one. So no re-check of accuracy with radiation is needed.

PrometheusPi (approved):

PR #3731, opened 2021-08-19
Fix warning: Attribute /date cannot be overwritten (openPMD plugin)

franzpoeschel:
Background: Using steps in openPMD/ADIOS2 with BP files is currently experimental and will stay so for a while (likely until we switch to an entirely new ADIOS2 schema). Until then, we should try to get PIConGPU to work in combination with steps without spamming warnings to stderr.

To fix the warning about attributes from previous steps not being modifiable, write the `date` attribute only once. (The other attributes are not affected since openPMD compares the attributes against those already specified, and does not write them anew if nothing changes)

I've already tested this and verified that the warnings do no longer occur.

This also fixes some low-key unspecified behavior: The `date` attribute is global in openPMD. If you specify it time after time again, there is no guarantee which instance you get. With this fix, you get the timestamp of the first iteration written with the plugin, i.e. simulation start time.

### DIFF
--- a/include/picongpu/plugins/common/openPMDWriteMeta.hpp
+++ b/include/picongpu/plugins/common/openPMDWriteMeta.hpp
@@ -139,8 +139,12 @@ namespace picongpu
                     softwareVersion << "-" << PICONGPU_VERSION_LABEL;
                 series.setSoftware(software, softwareVersion.str());
 
-                const std::string date = helper::getDateString("%F %T %z");
-                series.setDate(date);
+                // don't write this if a previous run already wrote it
+                if(!series.containsAttribute("date"))
+                {
+                    const std::string date = helper::getDateString("%F %T %z");
+                    series.setDate(date);
+                }
 
                 ::openPMD::Container<::openPMD::Mesh>& meshes = iteration.meshes;
 


sbastrakov (approved):

steindev:
@franzpoeschel But there is one warning left? It is just not repeated all the time? Because not providing a warning at all isn't a good solution, too?

franzpoeschel:
@steindev The warning comes from the openPMD-api because PIConGPU is trying to change the date of the dataset. With this PR, PIConGPU doesn't try to change the date any more, so there is no reason for warnings any longer.
The only reason for a warning that I can see is that setting the date to the simulation start is not wanted. If that is the case:
1. Then this PR should not be merged
2. The warnings in ADIOS2 should stay as overwriting attributes from a previous step is currently not possible
3. For other backends and iteration encodings, this should be tested in openPMD-api and overwriting across iterations should be explicitly supported there

Alternatively, set the date upon destruction of the openPMD plugin.

steindev:
@franzpoeschel In this case I think everything is alright. Setting the date of the simulation is useful.

steindev:
Will merge since @sbastrakov already approved.

PR #3747, opened 2021-09-01
Extend particles::Manipulate and the underlying implementation with area parameter

sbastrakov:
By default it is `CORE + BORDER`, which was hard-set before.
Thus, no existing code needs to be changed, but more flexibility is added.

I think this change will become useful for implemeting other types of particle boundary conditions, like reflective or thermal, that require per-particle processing in kernel. Then we can use this extended version with `T_area = GUARD`. Based on discussions with @steindev 

Edit: for now PIConGPU operates with compile-time areas and area mappings. In principle it would of course be nice to allow run-time ones as well at some point, as that would add much flexibility to e.g. define an area for absorber region, or maybe apply particle boundary conditions at adjustable points.

### DIFF
--- a/include/picongpu/particles/Manipulate.hpp
+++ b/include/picongpu/particles/Manipulate.hpp
@@ -31,6 +31,9 @@
 
 #include <boost/mpl/placeholders.hpp>
 
+#include <cstdint>
+#include <type_traits>
+
 
 namespace picongpu
 {
@@ -53,7 +56,7 @@ namespace picongpu
             };
         } // namespace detail
 
-        /** Run a user defined manipulation for each particle of a species
+        /** Run a user defined manipulation for each particle of a species in the given area
          *
          * Allows to manipulate attributes of existing particles in a species with
          * arbitrary unary functors ("manipulators").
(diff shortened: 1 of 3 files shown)


steindev (approved):

steindev:
As far as I remember, the `GUARD` size depends on the field solver. For arbitrary-order FDTD this can be more than one cell. Do you think that this an become an issue when manipulating particle momenta in the guard area?
(I think, if particles are not initialized in the guard area, than I do not see possible problems as they will not not penetrate deeper than one cell into the guard for all possible boundary conditions I can think of at the moment.)

sbastrakov:
`GUARD` and `BORDER` depend on basically all stencil-type things enabled in a simulation, both for fields and particles. So they are calculated to be enough (i.e. we never go out of bounds of the data in local domain) for all such operations possibly performed in the given simulation. All PIConGPU areas (in this sense of the word) are defined on supercell level, and so is the particle storage. So we would always have to take at least one supercell in the guard, and knowledge that particles are within one cell from boundary wouldn't really help. Except in very uncommon cases with multiple guard supercells, when running across whole `GUARD` would go through extra always-empty supercells, but even that is only small extra time, and no correctness issue.

psychocoderHPC (approved):

psychocoderHPC (approved):

PR #3750, opened 2021-09-01
Partially enable runtime kernel mapping

sbastrakov:
Traditionally PIConGPU kernels relied on compile-time defined areas `CORE`, `BORDER`, `GUARD` and combinations of those. Which is great for performance and not so great to express some operations. This PR makes a step towards allowing custom areas and mappings between alpaka thread blocks and supercells. It also adds documentation of existing use patterns, part of this will go to readthedocs as well in a follow up PR.

Currently the change affects some infrastructure to enable this in principle. The immediate effect is only on `pmacc::particles::algorithm::forEach` and its callees, mostly `picongpu::particles::Manipulate`. The default behavior is unchanged, including performance-wise. For now we dont have an existing use case for particles (but may need it for future boundary conditions). For fields there is a use case for absorbers, I will try it soon.

### DIFF
--- a/include/pmacc/mappings/kernel/AreaMapping.hpp
+++ b/include/pmacc/mappings/kernel/AreaMapping.hpp
@@ -26,6 +26,8 @@
 #include "pmacc/mappings/kernel/AreaMappingMethods.hpp"
 #include "pmacc/types.hpp"
 
+#include <cstdint>
+
 namespace pmacc
 {
     /** Mapping between block indices and supercells in the given area for alpaka kernels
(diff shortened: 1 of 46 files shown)


include/pmacc/mappings/kernel/AreaMapping.hpp
```diff
@@ -26,9 +26,34 @@
...
+     * A kernel must be launched with exactly getGridDim() blocks.
+     * Each block must process a single supercell with index getSuperCellIndex(blockIndex).
+     * Implementation is optimized for the standard areas in AreaType.
+     *
+     * In-block parallelism is independent of this mapping and is done by a kernel.
+     * Naturally, this parallelism should also define block size, again independent of this mapping.
+     *
+     * This pattern is used in most kernels in particle-mesh codes.
+     * Two most common parallel patterns are:
+     *    - for particle or particle-grid operations:
```
sbastrakov, line 48: I think this (or something similar) would be useful in readthedocs for developers. As it explains the parallelism pattern in most our kernels and how it is expressed in the code via a mapper.

sbastrakov:
@psychocoderHPC ready for review. Note: first commit is from #3747.

psychocoderHPC (approved):

PR #3758, opened 2021-09-02
Add checkpointing of internal RNG states

sbastrakov:
Previously they were re-initialized after loading thus potentially causing differences.
In case the states can't be loaded, the old scheme is used.
Thus, for old checkpoints there is no change in behavior.

Current status: compiles ~~but not checked yet~~ and works.

### DIFF
--- a/include/picongpu/plugins/openPMD/openPMDWriter.hpp
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.hpp
@@ -1,6 +1,6 @@
 /* Copyright 2014-2021 Axel Huebl, Felix Schmitt, Heiko Burau, Rene Widera,
  *                     Benjamin Worpitz, Alexander Grund, Franz Poeschel,
- *                     Pawel Ordyna
+ *                     Pawel Ordyna, Sergei Bastrakov
  *
  * This file is part of PIConGPU.
  *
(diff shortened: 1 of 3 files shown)


include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -1223,6 +1363,9 @@ Make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                 }
                 log<picLog::INPUT_OUTPUT>("openPMD: ( end ) writing particle species.");
 
+                log<picLog::INPUT_OUTPUT>("openPMD: ( begin ) writing RNG states.");
```
psychocoderHPC, line 1366: This code should be guarded by

```C++
if(mThreadParams.isCheckpoint == true)
{
// ...
}
```

RNG stated should not be stored into normal data dumps.

psychocoderHPC (changes requested):

include/picongpu/plugins/openPMD/openPMDWriter.hpp
sbastrakov: True. Why don't we do it for the id provider just below?

psychocoderHPC: It is a bug if we do not do it for the id provider too.

sbastrakov: Okay, fixed only in RNG. Id Provider will be fixed in another PR

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -591,6 +591,144 @@ Make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
...
+
+                const pmacc::Selection<simDim> localDomain = Environment<simDim>::get().SubGrid().getLocalDomain();
+                auto fieldsSizeDims = params->window.localDimensions.size;
+                auto fieldsOffsetDims = localDomain.offset;
+
+                if(fieldsSizeDims != rngProvider->getSize())
+                    throw std::runtime_error("openPMD: RNG state can't be loaded due to not matching size");
+
+                // Reinterpret state as chars, it must be bitwise-copyable for it
+                using ReinterpretedType = char;
```
franzpoeschel, line 689: Note that this kills portability, cross-platform as well as cross-application. The files written by the openPMD plugin are generally self-contained, but this field would be merely a raw byte-by-byte serialization. Is it an option to reinterpret_cast to an underlying machine type, e.g. by requiring that `RNGProvider::Buffer::ValueType` has an associated `UnderlyingType`?

sbastrakov: Cross-application is not needed here, it is only for checkpointing (and optional at that). Cross-platform and cross-application would be desired of course. The issue is I do not have a definition of this type, as it may come from a number of 3rd party libraries. The implementations I see have no fundamental underlying type, and are some structs with several members. So to me it appeared as the only solution.
I've actually [asked in the openPMD API repo about how to approach it](https://github.com/openPMD/openPMD-api/issues/1098).

franzpoeschel: I was expecting this answer, I just wanted to mention it. In the given case, I think it would be the easiest solution for now.

psychocoderHPC (approved):

psychocoderHPC (approved):

PR #3762, opened 2021-09-06
CI: compile on x86 runners 

psychocoderHPC:
Separate PMacc and PIConGPU jobs. PIConGPU is currently not executed on GPU therefore we can run the compile tests on CPU CI runner and only perform PMacc tests on the accelerator-specific runners. This is reducing the CI load and should result in result in faster CI response times.

To reduce the load PMacc is only testing each second boost version. The newest boost version is always included.

Update the CI base job descriptions to simplify switching to a new container tag or cuda/hip version.

The time to pass the CI is reduced from over 200 min to less than 100 min.

- [x] merge after #3761 (included in this PR)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,7 +2,11 @@
 stages:
   - validate
   - generate
-  - compile
+  - test_picongpu
+  - test_pmacc
+
+variables:
+  CONTAINER_TAG: "1.3"
 
 .base_generate-reduced-matrix:
   stage: generate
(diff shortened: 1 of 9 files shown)


.gitlab-ci.yml
```diff
@@ -56,16 +60,38 @@ pull-request-validation:
 #                    "examples" starts one gitlab job per directory in `examples/*`
 #                    "examples/" compile all directories in `examples/*` within one gitlab job
 #                    "examples/KelvinHelmholtz" compile all cases within one gitlab job
-generate-reduced-matrix:
+#                If path is 'pmacc' the folder will be ignored and tests for pmacc will be generated.
```
sbastrakov, line 63: Is this line misalighed?

share/ci/n_wise_generator.py
```diff
@@ -13,6 +13,9 @@
                     help='number of tuple elements')
 parser.add_argument('--compact', dest='compact', action="store_true",
                     help='print compact form of the test matrix')
+parser.add_argument('--limit_boost_versions', dest='limit_boost_versions',
+                    action="store_true",
+                    help='test only each second boost version')
```
sbastrakov, line 18: (not important) `every second`

sbastrakov (approved):
Some small notes above, but probably not important

SimeonEhrig:
Tobias added an new tag to the CPU runners: `cpuonly`
Please use it, to not block the GPU runners with CPU only jobs.

psychocoderHPC:
> Tobias added an new tag to the CPU runners: `cpuonly`
> Please use it, to not block the GPU runners with CPU only jobs.

Thx, I added the tag to this PR

sbastrakov (approved):

PR #3763, opened 2021-09-06
Enable support for custom particle boundaries

sbastrakov:
By default, same boundaries as set in `--periodic` are used. In fact, this is the only supported combination so far as other boundaries are not yet implemented. So there are no actual features added immediately by this PR.

However, it does some preparatory work and interfaces it to the command-line parameters:

- command-line options are added and propagated to species
- hook to put other boundary conditions is added with explanations for requirements
- the calling code to detect active boundaries and only process those is implemented for the general case

**For review**: might make sense to review commit-by-commit, however up to you

### DIFF
--- a/docs/TBG_macros.cfg
+++ b/docs/TBG_macros.cfg
@@ -91,6 +91,13 @@ TBG_gridDist="--gridDist '64{2}' '64,32{2},64'"
 TBG_periodic="--periodic 1 0 1"
 
 
+# Specifies boundary kinds for given species with a particle pusher, 'e' in the example.
+# The axis order matches --periodic.
+# Default: what is set by --periodic.
+# Currently only values that are same as default are supported, so this option can't change the behavior yet.
+TBG_boundaries="--e_boundary periodic absorbing periodic"
+
+
 # Set absorber type of absorbing boundaries.
 # Supported values: exponential, pml (default).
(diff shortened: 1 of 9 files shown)


docs/TBG_macros.cfg
```diff
@@ -91,6 +91,13 @@ TBG_gridDist="--gridDist '64{2}' '64,32{2},64'"
 TBG_periodic="--periodic 1 0 1"
 
 
+# Specifies boundary kinds for given species with a particle pusher, 'e' in the example.
+# The axis order matches --periodic.
+# Default: what is set by --periodic.
+# Currently only values that are same as default are supported, so this option can't change the behavior yet.
+TBG_boundaries="--e_boundary periodic absorbing periodic"
```
sbastrakov, line 98: Maybe `TBG_particleBoundaries` is better?

sbastrakov:
Btw @n01r with this PR in, it would be rather straightforward (well, by PIConGPU standards) to make alternative absorbing BCs at the center of PML or smth like that.

psychocoderHPC (approved):

docs/TBG_macros.cfg
sbastrakov: This is only a doc variable, I will change in a follow-up PR with other work on particle boundaries

PR #3764, opened 2021-09-07
Remove --openPMD.compression parameter

franzpoeschel:
This parameter was for legacy-style selection of compression in openPMD
which is implemented only in ADIOS1. The openPMD plugin does not support
ADIOS1, so this is inaccessible functionality.
The legacy-style compression support will likely be removed in openPMD,
compression should be configured in a backend-specific style via JSON.

### DIFF
--- a/docs/TBG_macros.cfg
+++ b/docs/TBG_macros.cfg
@@ -150,7 +150,6 @@ TBG_ranksPerDevice="--numRanksPerDevice 2"
 #--<species>_radiation.end     Time step to stop calculating the radiation
 #--<species>_radiation.radPerGPU     If flag is set, each GPU stores its own spectra without summing the entire simulation area
 #--<species>_radiation.folderRadPerGPU     Folder where the GPU specific spectras are stored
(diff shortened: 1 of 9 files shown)


sbastrakov (changes requested):
Thanks. Could you also remove it from the `XrayScattering` and `Radiation` plugins, including documentation for those. It seems you also left it in the `openPMD` plugin docs.

franzpoeschel:
Ah, good catch, yeah will do

sbastrakov:
Thanks! Don't forget the `TBG_macros.cfg`

franzpoeschel:
I've removed compression from code and documentation for openPMD and XRay scattering. There is none implemented or documented for Radiation?
Also, apparently we might want think about adding JSON reading to the other plugins if compression or other backend-specific configurations are wanted there.

sbastrakov:
You are right, for Radiation it was only in the docs that you are already removing in this PR

sbastrakov (approved):

PrometheusPi (approved):

psychocoderHPC (approved):

PR #3765, opened 2021-09-07
Clarify naming and comments for pmacc kernel mapping types

sbastrakov:
Fix mixing up of block and supercell indexing in the naming.
Move the `MapperConcept` and `MapperFactoryConcept` to a new file.
Link in the comments all mapping templates to the `MapperConcept` and the only (so far) existing factory to the factory concept.

### DIFF
--- a/include/pmacc/mappings/kernel/AreaMappingMethods.hpp
+++ b/include/pmacc/mappings/kernel/AreaMappingMethods.hpp
@@ -49,7 +49,7 @@ namespace pmacc
         }
 
         template<class Base>
-        HDINLINE static DataSpace<DIM> getBlockIndex(
+        HDINLINE static DataSpace<DIM> getSuperCellIndex(
             const Base&,
             const DataSpace<DIM>&,
             const DataSpace<DIM>& _blockIdx)
@@ -72,7 +72,7 @@ namespace pmacc
         }
 
         template<class Base>
-        HDINLINE static DataSpace<DIM> getBlockIndex(
+        HDINLINE static DataSpace<DIM> getSuperCellIndex(
             const Base& base,
             const DataSpace<DIM>& gBlocks,
             const DataSpace<DIM>& _blockIdx)
@@ -96,7 +96,7 @@ namespace pmacc
         }
 
         template<class Base>
-        HDINLINE static DataSpace<DIM> getBlockIndex(
+        HDINLINE static DataSpace<DIM> getSuperCellIndex(
             const Base& base,
             const DataSpace<DIM>& gBlocks,
             const DataSpace<DIM>& _blockIdx)
@@ -128,7 +128,7 @@ namespace pmacc
         }
 
         template<class Base>
-        HDINLINE static DataSpace<DIM2> getBlockIndex(
+        HDINLINE static DataSpace<DIM2> getSuperCellIndex(
             const Base& base,
             const DataSpace<DIM2>& gBlocks,
             const DataSpace<DIM2>& _blockIdx)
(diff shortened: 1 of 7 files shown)


include/pmacc/mappings/kernel/AreaMappingMethods.hpp
```diff
@@ -49,7 +49,7 @@ namespace pmacc
         }
 
         template<class Base>
-        HDINLINE static DataSpace<DIM> getBlockIndex(
+        HDINLINE static DataSpace<DIM> getSuperCellIndex(
```
sbastrakov, line 52: Note: this was named incorrectly. This function calculates supercell index using block index. So i renamed it. However it is only used internally in the respective mapper. Same for other similar functions changed in this PR

include/pmacc/mappings/kernel/BorderMapping.hpp
```diff
@@ -103,15 +108,14 @@ namespace pmacc
...
-         * Returns index of current logical block
+        /** Return index of a supercell to be processed by the given alpaka block
          *
-         * @param realSuperCellIdx current SuperCell index (block index)
-         * @return mapped SuperCell index
+         * @param blockIdx alpaka block index
+         * @return mapped SuperCell index including guards
          */
-        HDINLINE DimDataSpace getSuperCellIndex(const DimDataSpace& realSuperCellIdx) const
+        HDINLINE DimDataSpace getSuperCellIndex(const DimDataSpace& blockIdx) const
```
sbastrakov, line 116: This argument was also named incorrectly and misleadingly. The mapper is block -> supercell not vice versa as one might have understood in the old code

include/pmacc/mappings/kernel/MapperConcept.hpp
```diff
@@ -0,0 +1,114 @@
...
+     * @tparam T_dim dimensionality of area and block indices
+     * @tparam T_SupercellSize compile-time supercell size
+     */
+    template<
+        uint32_t T_areaType,
+        template<unsigned, typename>
+        typename T_MappingDescription,
+        unsigned T_dim,
+        typename T_SupercellSize>
+    class MapperConcept : public T_MappingDescription<T_dim, T_SupercellSize>
```
psychocoderHPC, line 64: If I see it correctly `MapperConcept` is not used anywhere. Is this intended?
I assumed that all mappers should inherit the concept class.

psychocoderHPC (changes requested):

include/pmacc/mappings/kernel/BorderMapping.hpp
psychocoderHPC: Yes your renaming at this point is correct.

include/pmacc/mappings/kernel/MapperConcept.hpp
sbastrakov: It is the same as field update functor concept we discussed some time ago. This class is only for documentation, so that if some user/developer sees template `T_Mapper` they know what interface to expect. Also to give such interface a name

psychocoderHPC: If all mapper inherits from this concept we can take care the developer is implementing all methods and that nobody is changing the interfaces and forget this concept class.
Another way is that you wrap the user mapper into this wrapper class to guarantee that all interfaces are implemented.

sbastrakov: Discussed offline. We now have a few of those "concept" types that are for documenting expected template interface. We should be making so that those are somehow used in the code: as wrappers, or maybe as alpaka's `ImplementsConcept`, or somehow else

psychocoderHPC (approved):

PR #3768, opened 2021-09-07
fix compile

psychocoderHPC:
In case OpenPMD is not available and PIConGPU is compiled for the CUDA backend the code is not compiling because of a syntax error.


sbastrakov (approved):

PrometheusPi (approved):

PrometheusPi:
Thanks for catching that bug @psychocoderHPC - sorry for introducing this bug.

PR #3770, opened 2021-09-08
Add a prototype and infrastructure for reflecting particle boundaries

sbastrakov:
This is not to be merged directly, but to serve as a basis for @lennertsprenger 's implementation.

PR #3772, opened 2021-09-08
Enable KernelShiftParticles to operate in guard area

sbastrakov:
As discussed [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3770#discussion_r704473988)

I didn't test it yet,  but maybe you can already have a look if that appears reasonable @psychocoderHPC ?

### DIFF
--- a/include/pmacc/particles/ParticlesBase.kernel
+++ b/include/pmacc/particles/ParticlesBase.kernel
@@ -1,4 +1,4 @@
-/* Copyright 2013-2021 Felix Schmitt, Heiko Burau, Rene Widera
+/* Copyright 2013-2021 Felix Schmitt, Heiko Burau, Rene Widera, Sergei Bastrakov
  *
  * This file is part of PMacc.
  *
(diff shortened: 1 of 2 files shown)


include/pmacc/particles/ParticlesBase.kernel
```diff
@@ -337,8 +344,8 @@ namespace pmacc
          *
          * @warning this kernel can only run with a double checker board
          */
-        template<typename T_ParBox, typename Mapping, typename T_Acc>
-        DINLINE void operator()(T_Acc const& acc, T_ParBox pb, Mapping mapper) const
+        template<typename T_ParBox, typename Mapping, typename T_Acc, typename T_Idx>
```
psychocoderHPC, line 347: Maybe we should call the type `T_SuperCellIDX`

sbastrakov: I don't mind it, but also think it's not better (nor worse). Do you want it changed?

psychocoderHPC: no not required, was only a suggestion.

psychocoderHPC (approved):

PR #3773, opened 2021-09-09
CI: use stages for downstream pipe

psychocoderHPC:
Group the downstream pipe into stages to reduce the load on the CI.
Additionally, this had the advantage that in case a job in the first stage is not passing the checks other stages will not be started, which keeps more CI resources free.

thx @SimeonEhrig for the hint that downstream pipes can have their own stages.

@SimeonEhrig said I should clean the python code, this will be a task for a follow-up PR.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,7 +15,7 @@ variables:
     - apt install -y python3-pip
     - pip3 install allpairspy
     - $CI_PROJECT_DIR/share/ci/git_merge.sh
-    - $CI_PROJECT_DIR/share/ci/generate_reduced_matrix.sh -n ${TEST_TUPLE_NUM_ELEM} > compile.yml
+    - $CI_PROJECT_DIR/share/ci/generate_reduced_matrix.sh -n ${TEST_TUPLE_NUM_ELEM} -j 15 > compile.yml
     - cat compile.yml
   artifacts:
     paths:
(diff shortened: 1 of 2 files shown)


sbastrakov (approved):

SimeonEhrig:
@psychocoderHPC I suggest, that you also implement a functionality to manual choose which jobs are executed or preferred. I think, this is really helpful, if a specific job crashes because of a bug. I already implemented a python function, which grep a list of job names from the commit message: https://gitlab.com/hzdr/crp/gitlab-ci-test/-/tree/jobControllCommitMessage/jobControllCommitMessage

You can integrate it in the job generator.

psychocoderHPC:
> @psychocoderHPC I suggest, that you also implement a functionality to manual choose which jobs are executed or preferred. I think, this is really helpful, if a specific job crashes because of a bug. I already implemented a python function, which grep a list of job names from the commit message: https://gitlab.com/hzdr/crp/gitlab-ci-test/-/tree/jobControllCommitMessage/jobControllCommitMessage
> 
> You can integrate it in the job generator.

Good idea but should be done in a follow up PR. IMO it is a nice feature but not strongly required.

It would also be possible to check older pipelines and find the broken jobs automatically.

psychocoderHPC:
@SimeonEhrig I opened an issue for your feature request https://github.com/ComputationalRadiationPhysics/picongpu/issues/3778

PR #3775, opened 2021-09-09
optimize kernel MoveAndMark

psychocoderHPC:
The pusher kernel contains a status flag per supercell which is a marker to know if at least one particle is leaving the supercell.
This status flag is located in shared memory and manipulated via an atomic function by each thread.
This PR aggregates the status per thread first into a register and uses atomic to synchronize the state between threads only if needed and not for each particle that leaves the cell.

### DIFF
--- a/include/picongpu/particles/Particles.kernel
+++ b/include/picongpu/particles/Particles.kernel
@@ -157,7 +157,7 @@ namespace picongpu
      *
      * Move frame-wise over a species and call a functor for each particle.
      * This kernel is optimized for the particle push step and handles the
-     * special flag `mustShift` of the supercell to optimize the kernel shift particles
+     * special flag `hasLeavingParticle` of the supercell to optimize the kernel shift particles
      * in pmacc.
      *
      * @tparam T_numWorkers number of workers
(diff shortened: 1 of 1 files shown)


include/picongpu/particles/Particles.kernel
```diff
@@ -211,15 +211,22 @@ namespace picongpu
...
 
-            PMACC_SMEM(acc, mustShift, int);
+            /* worker status flag to mark that a particle is leaving the supercell
+             * 1 if at least one particle is leaving the supercell else 0
+             */
+            int mustShift = 0;
+            /* shared memory status flag that a particle is leaving the supercell
+             * 1 if at least one particle is leaving the supercell else 0
+             */
+            PMACC_SMEM(acc, sharedMustShift, int);
```
sbastrakov, line 221: Not important, but why not a fixed-sized type here?

sbastrakov: I would prefer to tell those comments in names of the variables, as now they describe how they operate and not what they do.
E.g.
`mustShift` -> `mustShiftCurrentParticle`
`sharedMustShift` -> `mustShiftSupercell`

sbastrakov (commented):
I think this PR is fine. A suggestion for naming is above

include/picongpu/particles/Particles.kernel
psychocoderHPC: I will rename it next week.

psychocoderHPC: Maybe `mustShift` -> `hasLeavingParticle`

sbastrakov (approved):

PR #3776, opened 2021-09-09
Enable absorbing particle boundaries at custom positions

sbastrakov:
This adds a command-line-controlled offset inwards from the global boundaries. By default offset is 0 and this matches the old behavior both logically and performance-wise. Currently the implementation for offset > 0 is not very optimal performance-wise, but this should be a rather minor overhead. I will fix it once there are better abstractions in pmacc (soon).

Also now the particle BCs will be applied immediately after the initial pipeline is done. It means that one can set e.g. Homogeneous density and then have it cut according to the boundaries. The same insignificant efficiency shortcomings apply.

Update the boundary section in the docs with clarifications.
For now I will keep it as draft, but for me it appears to work. Feel free to test @n01r .

Note for review: may be easier done commit-by-commit.


sbastrakov:
Here are some pngs for the LWFA example, 1.cfg with absorbing boundaries on x, y and 50 offset. You can see the effect of the initial boundary application on density in the first .png. (Before i made the last commit in this PR, the density was full and only cut in the following images)


e_png_yx_0 5_000100
 
e_png_yx_0 5_000200

e_png_yx_0 5_000300

e_png_yx_0 5_000000

include/picongpu/particles/boundary/Absorbing.hpp
```diff
@@ -0,0 +1,145 @@
...
+            {
+                //! Functor to be applied to all particles in the active area
+                class AbsorbParticleIfOutside
+                {
+                public:
+                    //! Some name is required
+                    static constexpr char const* name = "absorbParticleIfOutside";
+
+                    //! Construct an instance, the parameters must be passed via staticParameters()
+                    AbsorbParticleIfOutside() : parameters(staticParameters())
```
sbastrakov, line 51: @psychocoderHPC this is quite awkward (also see that `staticParameters()` function). Do you know a better way? I understand the "global" data can be queried just inside this constructor via `Environment`. However, I also need to pass data that is only available in `ApplyImpl<Kind::Absorbing>::operator()` and nowhere else, and that pushed me to this ugly stuff.

sbastrakov:
Offline discussion with @psychocoderHPC . Doing it like currently in this PR would break plugins that rely on the `onParticleLeave()` hook for `offset > 0`. This is twofold:

- All those plugins are already not intended for this use case, so have to be adjusted anyways
- But currently the issue is, even if plugins are adjusted, the particles may already be deleted (or reflected, or modified otherwise) by that time by our boundary condition implementation.

We can probably modify the mechanism of how do plugins hook in the boundary treatment. Like if they provided a particle manupulator instead, we can call it for the relevant particles before the boundary condition modified them. E.g. for Calorimener that should work fine

sbastrakov:
@psychocoderHPC ready for review. The plugin hook we discussed yesterday is moved to correct position, and the ParticleCalorimeter plugin updated to work in this case.

include/picongpu/particles/boundary/Apply.hpp
```diff
@@ -80,38 +73,32 @@ namespace picongpu
...
-                        axis = 1;
-                    if(exchange >= BACK)
-                        axis = 2;
+                    /* Here is the only place in the computational loop where we can call hooks from plugins:
+                     *    - we know those particles crossed the active boundary
+                     *    - but we didn't yet apply the boundary conditions
+                     *      that would normally modify or delete the particles
+                     *      (this is done just afterwards in this function)
+                     */
+                    detail::callPluginHooks(species, exchange);
```
psychocoderHPC, line 90: offline discussed: Particles that leave the volume to the diagonal neighbor volume will be accounted twice. In the old implementation, each of the 26 exchanges was called independently.

psychocoderHPC (changes requested):

include/picongpu/particles/boundary/Apply.hpp
sbastrakov: In general you are right. I was making this change similarly to the new boundary conditions implementation, but it is not fully similar. However in this branch it's actually fine. Particles only leave when absorbing boundaries are used, and will be removed after crossing first such boundary and calling plugins once. With reflecting boundaries this will need to change, i got an idea how to do it

include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -643,6 +663,8 @@ namespace picongpu
                 particles->getDeviceParticlesBox(),
                 (MyCalorimeterFunctor) * this->calorimeterFunctor,
                 mapper,
```
sbastrakov, line 665: Discussed offline. The mapper should not be `GUARD`, but some form of runtime mapper for the boundary. I will wait for other open PRs to be merged to `dev`, then rebase, and make an appropriate mapper in this branch,

include/picongpu/particles/boundary/Absorbing.hpp
psychocoderHPC: unfortunately, I do not know a better way :-(

include/picongpu/particles/Particles.hpp
```diff
@@ -88,7 +91,7 @@ namespace picongpu
                       // fallback if the species has not defined the alias boundaryCondition
                       pmacc::HandleGuardRegion<
                           pmacc::particles::policies::ExchangeParticles,
-                          particles::boundary::CallPluginsAndDeleteParticles>>::type>,
+                          pmacc::particles::policies::DeleteParticles>>::type>,
```
psychocoderHPC, line 94: Discussed offline, this should become an empty functor because particle deleting will be handled by the particle boundary condition implementation.

include/picongpu/particles/Particles.hpp
```diff
@@ -107,7 +110,7 @@ namespace picongpu
                 // fallback if the species has not defined the alias boundaryCondition
                 pmacc::HandleGuardRegion<
                     pmacc::particles::policies::ExchangeParticles,
-                    particles::boundary::CallPluginsAndDeleteParticles>>::type>;
+                    pmacc::particles::policies::DeleteParticles>>::type>;
```
psychocoderHPC, line 113: same here, should be an empty functor.

psychocoderHPC (changes requested):
We need to perform a few runtime performance tests and check that we have no spilled registers due to the runtime mapper.

include/picongpu/particles/Particles.hpp
sbastrakov: Yes, I didn't have time to implement it yesterday

sbastrakov:
@psychocoderHPC on which part do you mean? The new versions are only used on the external boundaries. I mean e.g. the modified `pmacc::ParticlesBase::` methods are used with a custom mapper. For the other usage for an area exactly the same mapper type is still used internally. So i don't expect any change performance-wise. However it of course doesn't hurt to check.

include/picongpu/particles/Particles.hpp
sbastrakov: Fixed

include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
sbastrakov: Fixed now.

include/picongpu/particles/boundary/Apply.hpp
sbastrakov: Fixed now.

sbastrakov:
I believe those are (finally) all the changes required for this PR.

include/picongpu/particles/boundary/Absorbing.hpp
sbastrakov: I've now refactored it into a generic basic functor (last commit). So the ugly way is the same, but at least it's separated, documented and reusable.

psychocoderHPC:
I checked the register usage and runtime for this PR.
Register footprint is not changed.
There is no runtime variance visible.

psychocoderHPC (approved):

PR #3783, opened 2021-09-13
Add a new type of kernel mapping for a runtime-defined contiguous interval

sbastrakov:
This is the most basic runtime configuration, just an interval. But some useful stuff, like an active area for particle boundary conditions (it is already applied to one border which is a rectangle-type area), can already be expressed based on that. I will extend on a separate PR. And then I also plan to add a "frame" type area, like like GUARD but also runtime-defined.

### DIFF
--- a/include/pmacc/mappings/kernel/IntervalMapping.hpp
+++ b/include/pmacc/mappings/kernel/IntervalMapping.hpp
@@ -33,7 +33,7 @@ namespace pmacc
     /** Mapping from block indices to supercells in the given interval for alpaka kernels
      *
      * An interval is a T_dim-dimensional contiguous Cartesian range.
-     * Parameters of the inverval is defined at runtime, unlike most other mappers.
+     * Parameters of the inverval are defined at runtime, unlike most other mappers.
      *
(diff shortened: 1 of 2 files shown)


include/pmacc/mappings/kernel/IntervalMapping.hpp
```diff
@@ -0,0 +1,162 @@
...
+#include "pmacc/types.hpp"
+
+#include <cstdint>
+
+namespace pmacc
+{
+    /** Mapping from block indices to supercells in the given interval for alpaka kernels
+     *
+     * An interval is a T_dim-dimensional contiguous Cartesian range.
+     * Parameters of the inverval is defined at runtime, unlike most other mappers.
```
PrometheusPi, line 36:
```suggestion
     * Parameters of the inverval are defined at runtime, unlike most other mappers.
```

PrometheusPi (dismissed):

psychocoderHPC (approved):

sbastrakov:
As CI was stuck due to power outage, this PR could not pass in time, and I needed it urgently. So it is superceeded by #3776.

PR #3786, opened 2021-09-14
Make pmacc::math::Vector trivially default constructible

bernhardmgruber:
This PR makes `pmacc::math::Vector` trivially default constructible according to the language rules. It should not change any further semantic. This property is needed to use the type in LLAMA.


include/pmacc/math/vector/Vector.hpp
```diff
@@ -513,6 +503,8 @@ namespace pmacc
             }
         };
 
+        static_assert(std::is_trivially_default_constructible_v<Vector<float, 3>>, "");
```
bernhardmgruber, line 506: Assuming you are still on C++14. Otherwise we can remove the `, ""`.

sbastrakov (approved):

PrometheusPi (approved):

PrometheusPi:
Just  stupid question: Why could you remove the `Vector_components()` definition?

bernhardmgruber:
> Just stupid question: Why could you remove the `Vector_components()` definition?

Because `Vector_components` did not have any custom constructors with parameters, so the compiler can generate a default constructor itself.

include/pmacc/math/vector/Vector.hpp
bernhardmgruber: Ok, `std::is_trivially_default_constructible_v` is C++17. I have not been around C++14 for years. Will fix it.

psychocoderHPC: offline discussed: this check should be part of the vector class and should use the type and dimension from the vector.

bernhardmgruber: I removed the check now because it got too complicated. I cannot put it inside the class, because at the time of the check, the class is still incomplete.

bernhardmgruber:
So it seems I can leave the constructor of `Vector_components` removed, but I need to delegate to `Vector_components`'s implicitely generated constructor from `Vector`'s constructor. Since the constructor of `Vector_components` is defaulted, that should not perform value initialization.

psychocoderHPC:
@bernhardmgruber This PR is producing warnings:

```
/builds/hzdr/crp/picongpu/include/pmacc/../picongpu/param/ionizer.param(181): warning #811-D: const variable "picongpu::ionization::effectiveNuclearCharge::pmacc_static_const_storage27::pmacc_static_const_vector_device27::Hydrogen" requires an initializer -- class "pmacc::math::Vector<picongpu::float_X, 1, pmacc::math::StandardAccessor, pmacc::math::StandardNavigator, picongpu::ionization::effectiveNuclearCharge::pmacc_static_const_storage27::ConstArrayStorage>" has no user-provided default constructor
```

could be that https://github.com/ComputationalRadiationPhysics/picongpu/blob/4163c9f33838c4e4cf6d4d9f8c9c7cb8bd513659/include/pmacc/math/ConstVector.hpp#L99 needs a default constructor.

psychocoderHPC:
The error in one CI job is: 

```

/root/buildCI/params/Bunch/cmakePreset_0/include/picongpu/param/density.param(52): error #366-D: "picongpu::densityProfiles::pmacc_GaussianCloudParam68::GaussianCloudParam::GaussianCloudParam()" provides no initializer for:
            const member "picongpu::densityProfiles::pmacc_GaussianCloudParam68::GaussianCloudParam::center_SI"
            const member "picongpu::densityProfiles::pmacc_GaussianCloudParam68::GaussianCloudParam::sigma_SI"
```
https://gitlab.com/hzdr/crp/picongpu/-/jobs/1592254894

`GaussianCloudParam` can be provided by the user in the param files, if possible we like to avoid that the user must provide a constructor definition. Most users are physicists and it would not be obvious why they need to provide an constructor.

We can discuss this tomorrow in person.

sbastrakov:
@bernhardmgruber there seems to be more issues in the CI. I guess something similar to what is discussed above. We can look together if needed.

bernhardmgruber:
So I patched LLAMA in the meantime to provide a workaround and I do no longer need this PR. It is still technically wrong to reinterpret memory as a `Vector` without this PR, but since the ctor implementation is empty, it should not matter. I guess I will close this PR then, because it causes too much other troubles.

PR #3791, opened 2021-09-15
Add a documentation page for developers that explains how to extend PIConGPU

sbastrakov:
I think we are missing such documentation for a very long time. It is unfortunately quite hard to write, which is partially caused by PIConGPU being not very simple to extend. This is my effort to hopefully jumpstart this activity. If relatively recent contributors like @pordyna @BrianMarre can also have a look, it would be nice.

Note: this only extends readthedocs, you can see the result (once it builds) at the bottom, under "show all checks", where CI is.

### DIFF
--- a/docs/source/dev/extending.rst
+++ b/docs/source/dev/extending.rst
@@ -1,0 +1,103 @@
+.. _development-extending:
+
+Extending PIConGPU
+==================
+
+.. sectionauthor:: Sergei Bastrakov
+
+.. note::
+
+   A number of places in :ref:`.param files <usage-params>` allow providing user-defined functors.
+   The processing logic can largely be customized on that level, without modifying the source code.
+   Such an external way is recommended when applicable.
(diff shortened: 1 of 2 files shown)


sbastrakov:
Okay, my list formatting is off, will fix.

pordyna:
@sbastrakov I like it. This is a good start and we can always build up on this. One thing, it was not completely clear to me what is this common interface for stages and functors, maybe you could elaborate a bit more or include an example?

sbastrakov:
@pordyna good point. We actually do not have it well defined, only [this comment](https://github.com/ComputationalRadiationPhysics/picongpu/blob/d714a77ab4eae5766d4b5c8cc877bfcc164097e5/include/picongpu/param/speciesInitialization.param#L40-L41). I can make a stub "poor man's concept" class for it - a technique we already used in a number of places to document our template interfaces

psychocoderHPC (approved):

PR #3795, opened 2021-09-17
Remove leftover mentions of ADIOS1

### LINKED ISSUE #3751: Remove remaining support for ADIOS1
For now we are supposed to only support it for reading #3395, and that has been in this transition phase for 10 months. So before the upcoming release it seems reasonable to remove it at all.

sbastrakov:
We no longer support it, but some docs and code were still mentioning or specializing for it.
This commit removes all mentions, including `FindADIOS.cmake` which was no longer used.

Resolves #3751.


etc/picongpu/juwels-jsc/booster_picongpu.profile.example
```diff
@@ -63,7 +63,7 @@ export LD_LIBRARY_PATH=$PNGwriter_DIR/lib:$LD_LIBRARY_PATH
 export LD_LIBRARY_PATH=$ADIOS2_ROOT/lib64:$LD_LIBRARY_PATH
 export LD_LIBRARY_PATH=$OPENPMD_ROOT/lib64:$LD_LIBRARY_PATH
 
-export PATH=$ADIOS_ROOT/bin:$PATH
+export PATH=$ADIOS2_ROOT/bin:$PATH
```
sbastrakov, line 66: Note: this was probably a bug in the profile? It was alrady using ADIOS2, but this variable was still set for ADIOS1

include/picongpu/plugins/openPMD/Json_private.hpp
```diff
@@ -268,6 +268,6 @@ namespace picongpu
             std::string getDefault() const override;
         };
 
-        std::vector<std::string> const JsonMatcher::m_recognizedBackends = {"adios1", "adios2", "hdf5", "json"};
+        std::vector<std::string> const JsonMatcher::m_recognizedBackends = {"adios2", "hdf5", "json"};
```
PrometheusPi, line 271: If we do not provide adios2, openPMD-api could still use adios(1), correctly?
Why would you thus remove this `"adios1"` json tag?

PrometheusPi (dismissed):
Looks good to me - just a question @sbastrakov.

include/picongpu/plugins/openPMD/Json_private.hpp
sbastrakov: Yes, that is a valid question. I thought this piece is only for PIConGPU to parse arguments given as json, and so adios1 should be removed. However, i am not 100% sure. @franzpoeschel do you happen to know it?

franzpoeschel: Simple answer: I wrote this with the plan in mind that this could one day be mainlined into openPMD-api (listing the `json` backends doesn't make a lot of sense either)As . If a user defines a configuration for backends ADIOS1 or JSON, it will be handled by the regex matcher, but the resulting config will never be chosen.
We could alternatively just initialize this vector via `::openPMD::getVariants()`

franzpoeschel: > If we do not provide adios2, openPMD-api could still use adios(1), correctly?

No, ADIOS1 is deactivated in the openPMD plugin. If PIConGPU detects that the ADIOS1 backend has been selected, it throws an error.

PrometheusPi: Thanks for the last explanation @franzpoeschel  - that leads me to the following question ...

docs/source/usage/plugins/openPMD.rst
```diff
@@ -43,9 +43,9 @@ The openPMD API will parse the file name to decide the chosen backend and iterat
 In order to set defaults for these value, two further options control the filename:
 
 * ``--openPMD.ext`` sets the filename extension.
-  Possible extensions include ``bp`` for the ADIOS2 backends (default), ``h5`` for HDF5 and ``sst`` for Streaming via ADIOS2/SST.
-  If the openPMD API has been built with support for the ADIOS1 and ADIOS2 backends, ADIOS2 will take precedence over ADIOS1.
-  This behavior can be overridden by setting the environment variable ``OPENPMD_BP_BACKEND=ADIOS1``.
+  Possible extensions include ``bp`` for the ADIOS2 backend (default), ``h5`` for HDF5 and ``sst`` for Streaming via ADIOS2/SST.
+  If the openPMD API has been compiled with support for ADIOS2, the openPMD API will automatically prefer using ADIOS2 over ADIOS1.
```
PrometheusPi, line 47: If I understand @franzpoeschel [last comment](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3795#discussion_r710983399) corretly, using ADIOS(1) no option anymore.
Shouldn't than this sentence state:
```suggestion
  If the openPMD API has been compiled with support for ADIOS2, the openPMD API will automatically prefer using ADIOS2 over HDF5.
```

franzpoeschel: Those two sentences should go entirely, good catch. There is no preference of ADIOS2 over HDF5 or the other way around in openpMD-api, this was just there because the `.bp` extension could be used to select ADIOS2 as well as ADIOS1.

sbastrakov: I think the sensence is not about that. But it (plus the next one) should warn users who has both ADIOS1 and ADIOS2 in their openPMD, that in order to use it with PIConGPU they should take care of that enrironment variable

franzpoeschel: True, you can leave it in for that purpose. Then, I would formulate this a bit more explicitly to warn users against selecting ADIOS1 via that variable.

sbastrakov: Okay, i will reformulate

include/picongpu/plugins/openPMD/Json_private.hpp
sbastrakov: So is my change of this piece fine, or it should be `::openPMD::getVariants()`, or it doesn't matter for us?

docs/source/usage/plugins/openPMD.rst
franzpoeschel: [This error message](https://github.com/ComputationalRadiationPhysics/picongpu/blob/d7a4863e9a421fda8d50df31771c83d515b64f62/include/picongpu/plugins/openPMD/openPMDWriter.hpp#L152) may be a starting point

include/picongpu/plugins/openPMD/Json_private.hpp
franzpoeschel: I guess it doesn't really matter for PIConGPU

docs/source/usage/plugins/openPMD.rst
sbastrakov: Fixed

psychocoderHPC:
PLease do not remove the file `thirdParty/cmake-modules/FindADIOS.cmake` we need to update the repo https://github.com/ComputationalRadiationPhysics/cmake-modules and pull the third party file.

All other changes are fine.

sbastrakov:
@psychocoderHPC restored back FindADIOS.cmake

docs/source/install/libraryDependencies.dot
```diff
@@ -42,7 +36,6 @@ digraph PIConGPU {
 
     openpmd [label="openPMD API"];
     adios2 -> openpmd [style=dashed label="optional"];
-    adios -> openpmd [style=dashed label="optional"];
     hdf5 -> openpmd [style=dashed label="optional"];
     nlohmann_json [label="nlohmann_json\n(header-only)\n(internall shipped\nby default)"];
     nlohmann_json -> openpmd[style=dashed"]
```
franzpoeschel, line 41:
```suggestion
    nlohmann_json -> openpmd[style="dashed"]
```
Bug was previously there, but this thing does currently not build

docs/source/usage/plugins/openPMD.rst
```diff
@@ -43,9 +43,8 @@ The openPMD API will parse the file name to decide the chosen backend and iterat
 In order to set defaults for these value, two further options control the filename:
 
 * ``--openPMD.ext`` sets the filename extension.
-  Possible extensions include ``bp`` for the ADIOS2 backends (default), ``h5`` for HDF5 and ``sst`` for Streaming via ADIOS2/SST.
-  If the openPMD API has been built with support for the ADIOS1 and ADIOS2 backends, ADIOS2 will take precedence over ADIOS1.
-  This behavior can be overridden by setting the environment variable ``OPENPMD_BP_BACKEND=ADIOS1``.
+  Possible extensions include ``bp`` for the ADIOS2 backend (default), ``h5`` for HDF5 and ``sst`` for Streaming via ADIOS2/SST.
+  In case your openPMD API supports both ADIOS1 and ADIOS2, make sure that environment variable ``OPENPMD_BP_BACKEND`` is not set to ADIOS1.
```
franzpoeschel, line 47: Just a little note: We are soon going to support backend selection in the form of `{"backend": "adios1"}`. Since that's still WIP and might still change, we don't need to address it now, but that will be another way to select the ADIOS1 backend against PIConGPU's better judgement.

include/picongpu/plugins/Checkpoint.hpp
```diff
@@ -77,7 +77,7 @@ namespace picongpu
         {
             namespace po = boost::program_options;
             if(ioBackendsHelp.empty())
-                desc.add_options()("checkpoint", "plugin disabled [compiled without dependency HDF5 or Adios]");
+                desc.add_options()("checkpoint", "plugin disabled [compiled without dependency HDF5 or ADIOS2]");
```
franzpoeschel, line 80:
```suggestion
                desc.add_options()("checkpoint", "plugin disabled [compiled without dependency openPMD-api]");
```
I think that this message still comes from a time when we wrote to ADIOS1 and HDF5 directly. 
Do we still have ADIOS1? then `without dependency ADIOS1 or openPMD-api`

include/picongpu/plugins/PluginController.hpp
```diff
@@ -70,7 +70,7 @@
 #endif
 
 #if((ENABLE_OPENPMD == 1) && (openPMD_HAVE_HDF5 == 1))
-// needs to be openPMD-api only in the future to support adios
+// needs to be openPMD-api only in the future to support adios2
```
franzpoeschel, line 73: Isn't it openPMD-api only by now? I think this comment is outdated and should instead mention why it only uses the HDF5 backend of openPMD.

include/picongpu/plugins/Checkpoint.hpp
sbastrakov: No, ADIOS1 is effectively already removed. And this PR removes leftover mentions of it (that should have been removed before as well, but missed). So i think (the changed version of) this message is accurate now.

docs/source/usage/plugins/openPMD.rst
sbastrakov: Okay, i will keep the text as is for now.

include/picongpu/plugins/PluginController.hpp
sbastrakov: Correct. I think the message tried to express what you wrote, but did it not clearly. Rewrote

sbastrakov:
@franzpoeschel thanks , pushed an update.

franzpoeschel:
Looks good to me

psychocoderHPC (approved):

Issue #1893, opened 2017-02-22
laser correctionFactor needs more tests


psychocoderHPC:
The laser [correctionFactor](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1796/files) introduced in #1796 needs to be further tested:
- strongly focusing laser
- off-axis lineout (or 2D comparison)

sbastrakov:
Offline discussion with @psychocoderHPC @PrometheusPi . Now there is no up-to-date knowledge of why [such coefficients](https://github.com/ComputationalRadiationPhysics/picongpu/blob/33d8ffe595118d7499e1c8d01b85df369fa1b083/include/picongpu/fields/laserProfiles/PlaneWave.hpp#L114) are like that. The multiplication by 2 seems to follow from the 2-wave schema, but the rest is now unclear (probably was forgotten and seems to be not documented). It's not even clear if this scheme assumes `c * dt (almost) / dy close to 1` like for LWFA_type simulation or is it also fine for cubic cells as well. However, that we can easily check by compating the `initPlaneY == 0` and `!= 0` cases like what was done in #1796.

sbastrakov:
Okay, so this scheme seems to not depend on values for `dx`, `dz` in terms of the `y` direction propagation. So not specialized for LWFA-ish simulations.

PrometheusPi:
@sbastrakov Thanks for checking - that's great :+1:

sbastrakov:
The coefficient seems correct by value, but wrong in explanation. This should be fixed in the linked PR, closing the issue.

PR #3798, opened 2021-09-17
Refactor and document laser profiles

sbastrakov:
Refactors the common code in all the profiles, notably including our two-case source application.
The second commit adds a big chunk of description to how the laser actually works and what are the properties. There appears to be some confusion even from developers of those parts in PIConGPU, hopefully the new documentation resolves it. Also answers #1893 - basically @psychocoderHPC rediscovered a scheme almost no one else uses, but some ppl published (reference in the comment inside the code), and his fit coefficient matches theirs.

~~I think we should include this new doxygen somewhere as it basically explains how the laser works. But i'm not sure where.~~ found some place


sbastrakov:
Note to reviewers: I described everything in that big comment from my math / computer science point of view. If you see that it is weird from a physicist's point of view, feel free to introduce corrections.

psychocoderHPC (approved):

PR #3801, opened 2021-09-17
clang-tidy modernize

bernhardmgruber:
This PR runs clang-tidy with the modernize checks according to: https://github.com/ComputationalRadiationPhysics/contributing/blob/master/formatting-tools/.clang-tidy

For clang-tidy to work, PMacc needs to be included as non-SYSTEM headers.

I ran clang-tidy like this from the SPEC example .build folder:
```
cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON .
run-clang-tidy -fix -header-filter='^((?!/thirdParty/).)*$'
```
Then I revered changes to the cpp files of thirdParty tools.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,26 @@
+Checks: '
+modernize-raw-string-literal,
+modernize-replace-disallow-copy-and-assign-macro,
+modernize-return-braced-init-list,
+modernize-unary-static-assert,
+modernize-use-auto,
+modernize-use-bool-literals,
+modernize-use-default-member-init,
+modernize-use-equals-default,
+modernize-use-equals-delete,
+modernize-use-nullptr,
+modernize-use-override,
+modernize-use-transparent-functors,
+modernize-use-using'
+CheckOptions:
+  - { key: modernize-use-auto.MinTypeNameLength, value: 0}
+  - { key: modernize-use-auto.RemoveStars, value: 0}
+  - { key: modernize-use-bool-literals.IgnoreMacros, value: 0}
+  - { key: modernize-use-default-member-init.UseAssignment, value: 0}
+  - { key: modernize-use-default-member-init.IgnoreMacros, value: 0}
+  - { key: modernize-use-equals-default.IgnoreMacros, value: 0}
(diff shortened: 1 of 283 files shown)


psychocoderHPC:
You should run clang-format once.

include/pmacc/kernel/atomic.hpp
```diff
@@ -235,7 +235,8 @@ namespace pmacc
         DINLINE void atomicAllExch(const T_Acc& acc, T_Type* ptr, const T_Type value, const T_Hierarchy& hierarchy)
         {
             const auto mask = alpaka::warp::activemask(acc);
-            const auto leader = alpaka::ffs(acc, static_cast<std::make_signed_t<decltype(mask)>>(mask)) - 1;
+            [[maybe_unused]] const auto leader
+                = alpaka::ffs(acc, static_cast<std::make_signed_t<decltype(mask)>>(mask)) - 1;
```
psychocoderHPC, line 239: could you please use `alpaka::ignore_unused(leader);` to hide this warning. This will work for cpp14 too and not for cpp17+ only.

psychocoderHPC (changes requested):

include/pmacc/kernel/atomic.hpp
psychocoderHPC: Is it necessary to have this commit in this PR? This fix is mixing a feature with a warning fix. I would prefer to have a separate PR to be able to add correct lables.

bernhardmgruber: Done in #3803.

psychocoderHPC:
You missed formatting the code.^^

bernhardmgruber:
> You missed formatting the code.^^

```
PS C:\dev\picongpu> git clang-format upstream/dev
clang-format did not modify any files
```
No, i didn't :) I have clang-format 12 installed? Maybe that's the issue.

bernhardmgruber:
Reformatted with clang-format-11.

psychocoderHPC (dismissed):

psychocoderHPC (approved):

PR #3804, opened 2021-09-20
Add time-substepping versions of existing field solvers

sbastrakov:
In offline discussion at lunch on Friday we figured it may be sometimes useful to do sub-stepping inside the field solver. So that there is a general PIC dt (`DELTA_T` in code), but the field solver internally operates on a part of it. With the existing computational loop structure (E and B are at same time except between calls to the field solver, how current is added) it appears quite straightforward to add this solver version. Basically, only the parts called inside field solver - laser, incident field, and absorber - have to use the new `getTimeStep()` instead of `DELTA_T`. Note no changes to `Simulation::runOneStep()`.

As always, one has to be careful with dispersion, see [here](https://picongpu.readthedocs.io/en/latest/models/AOFDTD.html) and the comment for the new field solver. With substepping Yee solver we may move into bad dispersion zone (depending on other parameters). However, for high order FDTD we may be moving to a good zone instead according to the pictures @steindev put to the docs. Currently PIConGPU also prints phase velocity for the tield solver and its time step, so should be straightforward for a user to control.

While developing this feature, I noticed we enforce `c * dt < min(dx, dy, dz)` inside our default CFL check (which is superceded then for `Yee` and `AOFDTD` with their more demanding checks). In principle, this should be separated as this is required by our logic of particle movement and current deposition, not by field solvers themselves. This becomes relevant with substepping as then we should have both `c * dt_PIC < min(dx, dy, dz)` and field solver CFL for `dt_sub` with those two no longer being directly connected. In this PR this is followed, but probably not in most clear way. I will make a follow-up to make it more logical in the code and for users.

In order to make the implementation more natural, it also does a slight refactoring of the standard FDTD solvers.

I tried using the substepping solver for a few examples and it seems to behave reasonably. For LWFA as a target underdense case I compared the standard setup vs. same PIC `dt` and substepping vs. larger PIC `dt` and substepping so that field solver has original `dt` inside - the agreement appeared good.

As a small numerical comparison, here is standard KHI setup with `1.cfg`  and plots made with [`plotNumericalHeating`](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/src/tools/bin/plotNumericalHeating). There `dev` = standard Yee solver (as in the example), and `branch` is substepping with 2, 3, 4 times respectively. Main PIC `dt` is the same in all runs, so the substepping versions do better resolution inside the field solver. 
diff_substepping2
diff_substepping3
diff_substepping4

**Suggestion to reviewers:** go commit by commit.


include/picongpu/simulation/control/Simulation.hpp
```diff
@@ -674,6 +675,16 @@ namespace picongpu
...
     private:
+        //! Create and initialize field solver instance
+        void initFieldSolver()
+        {
+            myFieldSolver = new fields::Solver(*cellDescription);
+            // for substepping, hook the current addition stage
+            using IsSubsteppingFieldSolver = fields::maxwellSolver::traits::IsSubstepping<fields::Solver>;
+            pmacc::meta::invokeIf<IsSubsteppingFieldSolver::value>(
+                [this]() { myFieldSolver->setCurrentAddFunctor(currentInterpolationAndAdditionToEMF); });
+        }
```
sbastrakov, line 686: It causes warning as `setCurrentAddFunctor` is host-only.

`if constexpr` waiting room

psychocoderHPC: Maybe we need to add `PMACC_NO_NVCC_HDWARNING` over the operator implementation in `pmacc::meta::detail::IfElse<>`

sbastrakov: This part turned out not needed.

sbastrakov:
I went back to implement this, ETA end of this week.

Edit: as usual, took longer and new tasks appeared along the way

PrometheusPi:
@sbastrakov How is the status of the sub-stepping implementation?

sbastrakov:
@PrometheusPi unfortunately not ready yet. I believe the implementation works from my smaller runs, but have to do a proper test. It got pushed back many times due to solving other issues (such as mallocMC and incident field bugs)

sbastrakov:
@psychocoderHPC @steindev finally ready for review, I added some comparison plots and explanation to the first message.

include/picongpu/fields/MaxwellSolver/DispersionRelation.hpp
```diff
@@ -47,7 +49,7 @@ namespace picongpu
 
             protected:
                 //! Field solver time step as float_64
-                float_64 const timeStep = static_cast<float_64>(DELTA_T);
+                float_64 const timeStep = getTimeStep();
```
psychocoderHPC: why do you remove the explicit cast? The value type from `getTimeStep` is `float_X`

psychocoderHPC (dismissed):

include/picongpu/fields/MaxwellSolver/DispersionRelation.hpp
sbastrakov: Oh true - I was changing this part later and didn't notice lack of cast.

sbastrakov: Implemented in a new commit.

psychocoderHPC (approved):

psychocoderHPC (approved):

PR #3807, opened 2021-09-21
Avoid overaligning particle frame data

bernhardmgruber:
During my experiments with LLAMA I found that if the particle frame data is not overaligned with `PMACC_ALIGN` we have slight performance gains.

E.g. running the SPEC example on fwk394:
```
./picongpu -d 1 1 1 -g 256 256 256  -s 100 -p 5 --periodic 1 1 1 # dev
calculation  simulation time: 30sec 607msec = 30.607 sec

./picongpu -d 1 1 1 -g 256 256 256  -s 100 -p 5 --periodic 1 1 1 # this PR
calculation  simulation time: 29sec 512msec = 29.512 sec
```
That is a speedup of 1.037. I wonder if that effect is also visible in larger benchmarks. If it is, I would recommend merging this PR.

----------------------
update by @psychocoderHPC:
- [x] merge after 0.6.0 release branch is created

### DIFF
--- a/include/picongpu/particles/InitFunctors.hpp
+++ b/include/picongpu/particles/InitFunctors.hpp
@@ -28,7 +28,6 @@
 #include "picongpu/particles/manipulators/manipulators.def"
 
 #include <pmacc/Environment.hpp>
-#include <pmacc/math/MapTuple.hpp>
 #include <pmacc/meta/conversion/TypeToPointerPair.hpp>
 #include <pmacc/particles/meta/FindByNameOrType.hpp>
 #include <pmacc/traits/GetFlagType.hpp>
--- a/include/picongpu/particles/ParticlesFunctors.hpp
+++ b/include/picongpu/particles/ParticlesFunctors.hpp
@@ -28,7 +28,6 @@
 
 #include <pmacc/Environment.hpp>
 #include <pmacc/communication/AsyncCommunication.hpp>
-#include <pmacc/math/MapTuple.hpp>
 #include <pmacc/particles/meta/FindByNameOrType.hpp>
 #include <pmacc/traits/HasFlag.hpp>
 #if(PMACC_CUDA_ENABLED == 1)
--- a/include/picongpu/particles/access/Cell2Particle.tpp
+++ b/include/picongpu/particles/access/Cell2Particle.tpp
@@ -20,7 +20,6 @@
 #pragma once
 
 #include <pmacc/lockstep.hpp>
-#include <pmacc/math/MapTuple.hpp>
 #include <pmacc/math/Vector.hpp>
 #include <pmacc/memory/shared/Allocate.hpp>
 
--- a/include/pmacc/math/MapTuple.hpp
+++ b/include/pmacc/math/MapTuple.hpp
@@ -38,72 +38,24 @@ namespace pmacc
 
         /** wrap a datum
          *
-         * align the data structure with `PMACC_ALIGN`
-         *
          * @tparam T_Pair boost mpl pair< key, type of the value >
          */
         template<typename T_Pair>
-        struct AlignedData
-        {
-            using Key = typename T_Pair::first;
-            using ValueType = typename T_Pair::second;
-
-            PMACC_ALIGN(value, ValueType);
-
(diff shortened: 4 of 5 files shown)


sbastrakov:
Is overaligning caused by that `Align` policy (or idk what it is) in general, or the align amount used inside is too generous?

bernhardmgruber:
> Is overaligning caused by that `Align` policy (or idk what it is) in general, or the align amount used inside is too generous?

So `PMACC_ALIGN` aligns a type to the next power of two, clamped to [16;128]. So e.g. a `Vector<float, 3>[256]` (size 3072) is aligned to 128, which seems very excessive to me. That could cause a lot of padding in the particle frame.

sbastrakov:
But that whole logic appears weird. The alightment size for an array should only depend on array size for very small arrays. After a certain point (and rather quickly) it should be just a function of the element datatype.

bussmann:
In general, performance gains of a few percent should not be overrated.

psychocoderHPC:
Performance on AMD MI100

SPEC Benchmark
```
srun -n 1 ./bin/picongpu_noAlign -d 1 1 1 -g  128 128 128 -s 100 --periodic 1 1 1
calculation  simulation time:  6sec 759msec = 6.759 sec

srun -n 1 ./bin/picongpu_align -d 1 1 1 -g  128 128 128 -s 100 --periodic 1 1 1
calculation  simulation time:  6sec 850msec = 6.850 sec
```

```
srun -n 1 ./bin/picongpu_noAlign -d 1 1 1 -g  256 256 256 -s 100 --periodic 1 1 1
calculation  simulation time: 48sec 852msec = 48.852 sec

srun -n 1 ./bin/picongpu_align -d 1 1 1 -g  256 256 256 -s 100 --periodic 1 1 1
calculation  simulation time: 49sec 278msec = 49.278 sec
```

psychocoderHPC:
To allow this PR we need to align the data to the `sizeof` the largest component. The best would be ordering the data by size (largest first) and then align based on the component type.

bernhardmgruber:
SPEC on hemera V100:
dev | this PR
-- | --
24.498 | 24.141
24.495 | 24.119
24.659 | 24.231

Unfortunately, I cannot benchmark on k20/k80 because I don't have permissions to run there.

psychocoderHPC:
I am still unsure how we can guarantee the alignment of the particle attribute data if we do not use some kind of alignment policy.

Removing the alignment works fine because our particle frames have mostly a nice element count of 256.
I would feel better if we align our data at least to 16 byte, this helps hopefully on CPU the auto vectorization too.

bernhardmgruber:
> I am still unsure how we can guarantee the alignment of the particle attribute data if we do not use some kind of alignment policy.

There is the default alignment imposed by the language rules, which in my understanding is good enough for program correctness.

> Removing the alignment works fine because our particle frames have mostly a nice element count of 256. I would feel better if we align our data at least to 16 byte, this helps hopefully on CPU the auto vectorization too.

The tuple size in a frame is the number of cells in a supercell. Can this even not be a multiple of 32, since it is supposed to map nicely to CUDA warps. So you have at least an alignment of 32.

As for SIMD vectorization, unaligned loads have become so fast nowadays that I would not worry about the alignment anymore and rather strive for compact memory layouts to minimize the number of transferred bytes. But sure, eventually this needs to be benchmarked in the future when PIConGPU has a SIMD version.

psychocoderHPC:
> The tuple size in a frame is the number of cells in a supercell. Can this even not be a multiple of 32, since it is supposed to map nicely to CUDA warps. So you have at least an alignment of 32.

A supercell could theoretically be `3x3x3` cells but it is very unlikely.

bernhardmgruber:
SPEC on hemera k80:
Running ./picongpu -d 1 1 1 -g 256 256 256 -s 100 -p 5 --periodic 1 1 1

| dev | noalign |
| -- | -- |
| 206.490 | 206.264 |
| 206.959 | 206.126 |

psychocoderHPC (approved):

psychocoderHPC:
@bernhardmgruber thanks for the tests and the PR.

PR #3818, opened 2021-09-23
Guard definition of FromOpenPMD

bernhardmgruber:
I failed to compile PIConGPU without OpenPMD and I had to guard the offending line.

Do we have CI runners testing without OpenPMD?


sbastrakov (approved):
That is a bug i accidentally introduced when switching from hdf5 density to openPMD density

PR #3825, opened 2021-09-27
adjust physics log output

### LINKED ISSUE #3201: Log formatting.
Currently the output at runtime produces, for instance, the following:

    PIConGPUVerbose PHYSICS(1) | Courant c*dt <= 1.66765 ? 1
    PIConGPUVerbose PHYSICS(1) | species e: omega_p * dt <= 0.1 ? 3.56797e-08

The way the output is formatted I interpret it as a "question to the code", i.e., "Is c*dt <= <some quantity> ?" To which a natural answer would be "Yes/No" or "True/False" and not 1. 
Similarly for the second line.

Perhaps adding a `boolalpha` somewhere would be appropriate.

psychocoderHPC:
fix #3201

Simplify the log output for the field solver condition and omega_p.

@sbastrakov The field condition for our Lehe solver is currently `c * dt <= min(dx, dy, dz)` is this correct. 
I assume that we do not know the correct condition and therefore we fall back to the default.

### DIFF
--- a/include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.hpp
+++ b/include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.hpp
@@ -48,8 +48,11 @@ namespace picongpu
             template<uint32_t T_neighbors>
             struct CFLChecker<ArbitraryOrderFDTD<T_neighbors>>
             {
-                //! Check the CFL condition, throws when failed
-                void operator()() const
+                /** Check the CFL condition, throws when failed
+                 *
+                 * * @return value of 'X' to fulfill the condition 'c * dt <= X`
+                 */
+                float_X operator()() const
                 {
                     // The equations are given at https://picongpu.readthedocs.io/en/latest/models/AOFDTD.html
                     auto weights = aoFDTD::AOFDTDWeights<T_neighbors>{};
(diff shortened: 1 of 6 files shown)


sbastrakov:
For Lehe our CFL is correct

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.hpp
```diff
@@ -59,11 +62,17 @@ namespace picongpu
...
-                       > 1.0_X)
+                    auto const invCorrectedCell2Sum = INV_CELL2_SUM * additionalFactor * additionalFactor;
+                    auto const maxC_DT = 1.0_X / math::sqrt(invCorrectedCell2Sum);
+                    if(SPEED_OF_LIGHT * SPEED_OF_LIGHT * DELTA_T * DELTA_T * invCorrectedCell2Sum > 1.0_X)
+                    {
                         throw std::runtime_error(
-                            "Courant-Friedrichs-Lewy condition check failed, check your grid.param file");
+                            std::string("Courant-Friedrichs-Lewy condition check failed, check your grid.param file\n")
+                            + "Courant Friedrichs Lewy condition: c * dt <= " + std::to_string(maxC_DT)
+                            + " ? (c * dt = " + std::to_string(SPEED_OF_LIGHT * DELTA_T) + ")");
```
sbastrakov, line 72: I think this is incorrect. From this code it appears `maxC_DT` is an upper bound for `(c * dt)^2`, not `c * dt`

sbastrakov: If so, it is the same issue for all checkers, not just this one

sbastrakov (changes requested):
I think there is a confusion between c * dt, and a square of it, see above

include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.hpp
psychocoderHPC: No it is correct it should be `c * dt` this is equal to the output we already have.

sbastrakov: You are right, i got confused

sbastrakov (approved):

PR #3826, opened 2021-09-28
Update to clang-format-12

bernhardmgruber:
This PR updates and reformats the code with clang-format-12, as we have decided to do a long time ago: https://github.com/ComputationalRadiationPhysics/contributing/issues/49


bernhardmgruber:
Hmm. I don't get why the CI complains about wrong formatting. If I run the same command locally, I get no errors:
```
bgruber@dixxi3:/mnt/c/dev/picongpu$ find include/ share/picongpu/ share/pmacc -iname "*.def" \
  -o -iname "*.h" -o -iname "*.cpp" -o -iname "*.cu" \
  -o -iname "*.hpp" -o -iname "*.tpp" -o -iname "*.kernel" \
  -o -iname "*.loader" -o -iname "*.param" -o -iname "*.unitless" \
  | xargs clang-format-12 --dry-run --Werror
bgruber@dixxi3:/mnt/c/dev/picongpu$
```

psychocoderHPC:
https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/docs/source/dev/styleguide.rst#format-all-files and https://github.com/psychocoderHPC/picongpu/blob/dev/docs/COMMIT.md#format-code must be updated too.

bernhardmgruber:
> I tested `clang-format-12` on my machine and I get the same formating error the CI is reporting.

Now it gets weird:
```
bgruber@fwk394:~/picongpu$ spack load llvm@12.0.0%gcc@10.2.0
bgruber@fwk394:~/picongpu$ find include/ share/picongpu/ share/pmacc -iname "*.def"   -o -iname "*.h" -o -iname "*.cpp" -o -iname "*.cu"   -o -iname "*.hpp" -o -iname "*.tpp" -o -iname "*.kernel"   -o -iname "*.loader" -o -iname "*.param" -o -iname "*.unitless"   | xargs clang-format --dry-run --Werror
bgruber@fwk394:~/picongpu$
```
No errors on the dev system :S

psychocoderHPC:
Maybe the clang 12 installation on fwk394 is older and points to a pre release of clang 12.

bernhardmgruber:
I found it. I formatted with clang-format 12.0.0, but the CI uses clang-format 12.0.1. I installed LLVM 12.0.1 on the dev system and now I can reproduce the errors. But what should be do now? Mandate clang-format 12.0.1?

psychocoderHPC:
I do not know if the very small improvement in code style (east coast parentheses for lambda functions) is worth the issue developer will have because say use maybe a different `clang-format-12` version.

If we switch to clang format 12 we need to add the exact version into the documentation and add a clang format version check into the CI to avoid that the clang repository is silently updating to 12.0.X and all jobs failing and we waste developer time until we find out that clang 12 packages updated to a new minor version and break the formatation again.

bernhardmgruber:
Updated the documentation to specifically request version 12.0.1.

PR #3830, opened 2021-09-30
Remove boost::result_of

bernhardmgruber:
This PR removes all machinery related to `boost::result_of`, which in many cases can be replaced by C++11 deduction features.

### DIFF
--- a/include/pmacc/particles/memory/dataTypes/Particle.hpp
+++ b/include/pmacc/particles/memory/dataTypes/Particle.hpp
@@ -128,9 +128,7 @@ namespace pmacc
          * @return result of operator[] of the Frame
          */
         template<typename T_Key>
-        HDINLINE typename boost::result_of<
-            typename std::remove_reference_t<typename boost::result_of<FrameType(T_Key)>::type>(uint32_t)>::type
-        operator[](const T_Key key)
+        HDINLINE auto& operator[](const T_Key key)
         {
             PMACC_CASSERT_MSG_TYPE(key_not_available, T_Key, traits::HasIdentifier<Particle, T_Key>::type::value);
 
@@ -139,9 +137,7 @@ namespace pmacc
 
         /** const version of method operator(const T_Key) */
         template<typename T_Key>
-        HDINLINE typename boost::result_of<
-            typename std::remove_reference_t<typename boost::result_of<const FrameType(T_Key)>::type>(uint32_t)>::type
-        operator[](const T_Key key) const
+        HDINLINE const auto& operator[](const T_Key key) const
         {
             PMACC_CASSERT_MSG_TYPE(key_not_available, T_Key, traits::HasIdentifier<Particle, T_Key>::type::value);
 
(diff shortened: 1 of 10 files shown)


include/pmacc/particles/memory/dataTypes/Particle.hpp
```diff
@@ -139,9 +137,7 @@ namespace pmacc
 
         /** const version of method operator(const T_Key) */
         template<typename T_Key>
-        HDINLINE typename boost::result_of<
-            typename std::remove_reference_t<typename boost::result_of<const FrameType(T_Key)>::type>(uint32_t)>::type
-        operator[](const T_Key key) const
+        HDINLINE const auto& operator[](const T_Key key) const
```
sbastrakov, line 140: I'm not sure, but these operators look like forwarding, should those return `decltype(auto)` as well?

sbastrakov (approved):
Nice! One question above, but not essential

include/pmacc/particles/memory/dataTypes/Particle.hpp
bernhardmgruber: I decided to stick with explicitely stating that the return value is a const reference. The return value refers to 1 element of the array<T, 256> for `T_Key` held by the frame. I thought it gives a better idea of what this return type is. In fact, on my llama branch, I changed this into `decltype(auto)` because then this method can also return proxy references.

sbastrakov: Okay, thanks for explaining.

sbastrakov:
@bernhardmgruber now there is a conflict because of another your PR merged

include/pmacc/particles/memory/frames/Frame.hpp
```diff
@@ -125,15 +102,15 @@ namespace pmacc
          * @return result of operator[] of MapTupel
          */
         template<typename T_Key>
-        HDINLINE typename boost::result_of<ThisType(T_Key)>::type getIdentifier(const T_Key)
+        HDINLINE auto& getIdentifier(const T_Key)
         {
             using Key = typename GetKeyFromAlias<ValueTypeSeq, T_Key>::type;
```
psychocoderHPC, line 107: please add `errorHandlerPolicies::ThrowValueNotFound` to `GetKeyFromAlias`, The throw policy was handled in the removed `result struct`. The last policy is defining whats happens if the key is not available in the Frame.

include/pmacc/particles/memory/frames/Frame.hpp
```diff
@@ -125,15 +102,15 @@ namespace pmacc
...
             return BaseType::operator[](Key());
         }
 
         /** const version of method getIdentifier(const T_Key) */
         template<typename T_Key>
-        HDINLINE typename boost::result_of<const ThisType(T_Key)>::type getIdentifier(const T_Key) const
+        HDINLINE const auto& getIdentifier(const T_Key) const
         {
             using Key = typename GetKeyFromAlias<ValueTypeSeq, T_Key>::type;
```
psychocoderHPC, line 115: please add `errorHandlerPolicies::ThrowValueNotFound`

psychocoderHPC (changes requested):

psychocoderHPC:
Thanks for modernising the old C++98 parts in PMacc and PIConGPU :heart:

include/pmacc/particles/memory/frames/Frame.hpp
bernhardmgruber: done!

include/pmacc/particles/memory/frames/Frame.hpp
bernhardmgruber: done!

bernhardmgruber:
> Thanks for modernising the old C++98 parts in PMacc and PIConGPU ❤️

Well, I am touching it anyway during the LLAMA integration so I can just as well contribute it earlier :) And it makes the final LLAMA changeset much smaller, which is then easier to debug and review. You are welcome!

psychocoderHPC (approved):

PR #3832, opened 2021-09-30
Add a new particle manipulator combining total cell offset and rng

sbastrakov:
This functor is effectively a combination of two wrappers around `T_Functor`: `FreeTotalCellOffset` and `FreeRng`. Currently there is no easy way to achieve such a combination via sequantial application. So here is a special wrapper for it. We discussed with @psychocoderHPC a few times that ideally we should have some generic way for a functor to describe used resources (not in redgrapes sense, but logically it is similar) so that we don't have to specialize for each new case.

This functor is to enable work on thermal particle boundaries.

### DIFF
--- a/include/picongpu/particles/manipulators/unary/FreeTotalCellOffsetRng.hpp
+++ b/include/picongpu/particles/manipulators/unary/FreeTotalCellOffsetRng.hpp
@@ -1,0 +1,147 @@
+/* Copyright 2015-2021 Rene Widera, Alexander Grund, Axel Huebl, Sergei Bastrakov
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


include/picongpu/particles/manipulators/unary/FreeTotalCellOffsetRng.hpp
```diff
@@ -0,0 +1,147 @@
...
+                        T_Acc const& acc,
+                        DataSpace<simDim> const& localSupercellOffset,
+                        T_WorkerCfg const& workerCfg) const -> acc::FreeTotalCellOffsetRng<Functor, RngType>
+                    {
+                        auto& cellOffsetFunctor = *static_cast<CellOffsetFunctor const*>(this);
+                        RngType const rng
+                            = (*static_cast<RngGenerator const*>(this))(acc, localSupercellOffset, workerCfg);
+                        return acc::FreeTotalCellOffsetRng<Functor, RngType>(
+                            *static_cast<Functor const*>(this),
+                            cellOffsetFunctor(acc, localSupercellOffset, workerCfg),
```
psychocoderHPC, line 134: Note: If you like to avoid casting you could write: `CellOffsetFunctor::operator()(acc, localSupercellOffset, workerCfg)`

psychocoderHPC (approved):

include/picongpu/particles/manipulators/unary/FreeTotalCellOffsetRng.hpp
sbastrakov: True, I copied from another wrapper and didn't think about it.

PR #3834, opened 2021-09-30
Replace mpl by mp11 (part 1/2)

bernhardmgruber:
This PR replaced most uses of `boost::mpl` by `boost::mp11`. Only `boost::mpl::apply` and the corresponding placeholders have not been replaced yet, because there is no direct equivalent in mp11. The functionality of mp11 is directly pulled into the `pmacc` namespace, since all mp11 type functions are prefixed with `mp_`.

- [x] This PR requires at least Boost 1.66.0, because this is the first Boost version where Boost.Mp11 is available. Separate PR: #3939.
- [x] This PR also makes use of C++17 features, so merging #3949 is a prerequisit.


PrometheusPi:
Looks definitely much shorter.

psychocoderHPC (changes requested):
could you please rebase this PR and squash the changes, maybe add all .param changes into one commit, this would help to see what the user must change to be able to run old setups with the new required changes.

bernhardmgruber:
Questions to move on:
* How are users and developers expected to access mp11 functionality? Spelling `boost::mp11::mp_....` is quite long. With mpl, we had the namespace alias `bmpl`. So far I have used an alias `bmp` for `boost::mp11` in some places. But I think we could also just import mp11 directly into `pmacc` and `picongpu` since all metaprogramming functions are prefixed with `mp_`. The user could thus spell `mp_...` directly in any picongpu or pmacc namespace. What do you think?
* Can we move to at least Boost 1.66? Should we split this off and merge it sooner?
* There is no directy equivalent of MPL lambdas in mp11. That is, you cannot define a type function where several parameters are placeholders (e.g. `_1`) and then call something like `apply<TypeFunction, Args>::fn` which replaces the placeholders by actual template parameters. The workaround is to create alias templates and pass higher order type functions. In some cases this is more verbose, but faster to compile (no big impact though) and leads to less long error messages. We could also provide a similar facility on top of mp11. Currently, several places using `mpl::apply` with placeholders remain because they are not easy to transition. We should eventually move away from MPL completely (to not have 2 metaprogramming libraries). Can we merge this PR without finding a solution to MPL lambdas? Do we want to keep MPL for only its lambdas? Do we want a custom solution on top of mp11?
* I replaced `mpl::pair` by a plain `mp_list` in some cases. For others, I created `pmacc::meta::Pair`. We would not need it though and could just use `mp_list`. mp11 does not have a dedicated pair type and this is fine IMO.
* Should developers include mp11 headers directly, or should there be a central pmacc header pulling in mp11, that is supposed to be included by everyone else?
* The PR is huge, should/can we split it in parts?
* `pmacc::CT::Vector` could benefit from refactoring to use more `constexpr` instead of meta programming. The current version (with mp11) is not great, but this is not mp11's fault. Do anything here?
* This PR requires C++17 already because of `if constexpr` in some places. Has PIConGPU switched already?

psychocoderHPC:
> Questions to move on:
> 
> * How are users and developers expected to access mp11 functionality? Spelling `boost::mp11::mp_....` is quite long. With mpl, we had the namespace alias `bmpl`. So far I have used an alias `bmp` for `boost::mp11` in some places. But I think we could also just import mp11 directly into `pmacc` and `picongpu` since all metaprogramming functions are prefixed with `mp_`. The user could thus spell `mp_...` directly in any picongpu or pmacc namespace. What do you think?

Since pmacc is heavily depending on meta-programming I would be fine to pull mp11 into pmacc.

> * Can we move to at least Boost 1.66? Should we split this off and merge it sooner?

Yes would be possible but requires a separate PR. Why do you like to remove boost 1.65.1?

> * There is no directy equivalent of MPL lambdas in mp11. That is, you cannot define a type function where several parameters are placeholders (e.g. `_1`) and then call something like `apply<TypeFunction, Args>::fn` which replaces the placeholders by actual template parameters. The workaround is to create alias templates and pass higher order type functions. In some cases this is more verbose, but faster to compile (no big impact though) and leads to less long error messages. We could also provide a similar facility on top of mp11. Currently, several places using `mpl::apply` with placeholders remain because they are not easy to transition. We should eventually move away from MPL completely (to not have 2 metaprogramming libraries). Can we merge this PR without finding a solution to MPL lambdas? Do we want to keep MPL for only its lambdas? Do we want a custom solution on top of mp11?

We can not get rid of compile time lambdas. Most functionality depends on it. I am fine with going on with this PR without removing the mpl lambdas. 

> * I replaced `mpl::pair` by a plain `mp_list` in some cases. For others, I created `pmacc::meta::Pair`. We would not need it though and could just use `mp_list`. mp11 does not have a dedicated pair type and this is fine IMO.

I am fine with `pmacc::meta::Pair`. Could we maybe use `std::tuple`. The code is from c++98 times, nowadays usage of std::tuple should be ok, but maybe I miss something because I have not looked into the usage.

> * Should developers include mp11 headers directly, or should there be a central pmacc header pulling in mp11, that is supposed to be included by everyone else?

I think if pmacc is pulling this dependency it is fine.

> * The PR is huge, should/can we split it in parts?

If possible I would prefer small PR and changing the parts step by step but since mpl is everywhere used it could be easier to change it within one PR.

> * `pmacc::CT::Vector` could benefit from refactoring to use more `constexpr` instead of meta programming. The current version (with mp11) is not great, but this is not mp11's fault. Do anything here?

I am fine to use `constexpr` where it makes sense and is useful. As I said above most code is from C++98 times and three was no way's around template magic where we could use now more elegant ways and therefore reduce the compile time too.

> * This PR requires C++17 already because of `if constexpr` in some places. Has PIConGPU switched already?

No PIConGPU is still C++14. We can switch the dev branch any time to C+17. The last time it was blocked by an issue in the dependency openPMD-api. 
Switching to C++17 requires a separate PR.

bernhardmgruber:
> Since pmacc is heavily depending on meta-programming I would be fine to pull mp11 into pmacc.

Done. All functionality is available as `pmacc::mp_...`.

> Why do you like to remove boost 1.65.1?

Boost.Mp11 is available since Boost 1.66. PR here: #3939 

> We can not get rid of compile time lambdas. Most functionality depends on it. I am fine with going on with this PR without removing the mpl lambdas.

OK!

> I am fine with `pmacc::meta::Pair`. Could we maybe use `std::tuple`. The code is from c++98 times, nowadays usage of std::tuple should be ok, but maybe I miss something because I have not looked into the usage.

`pmacc::meta::Pair` is just a type list of fixed length 2. It's literally just:
```c++
    template<typename First, typename Second>
    struct Pair
    {
        using first = First;
        using second = Second;
    };
```

> I think if pmacc is pulling this dependency it is fine.

I created a new header `pmacc/meta/Mp11.hpp` that pulls in Boost.Mp11.

> We can switch the dev branch any time to C+17. The last time it was blocked by an issue in the dependency openPMD-api. Switching to C++17 requires a separate PR.

Can I take this as: Someone else is already working on that?

bernhardmgruber:
About MPL Lambdas: I had a closer look on how they work and how this is solved in Mp11. In Mp11, [`mp_bind` with placeholders](https://www.boost.org/doc/libs/develop/libs/mp11/doc/html/mp11.html#bind) is used, producing a quoted metafunction, instead of an MPL [placeholder expressions](https://www.boost.org/doc/libs/1_78_0/libs/mpl/doc/refmanual/placeholder-expression.html) and `mpl::apply`. In both cases the result is a type that can be put into type lists or passed as normal template argument.

Now, using `mp_bind` on the client side of APIs is probably less elegant. However, with some refactoring we can also eliminate placeholder expression entirely in many cases by passing type functions directly. For example most of `meta::ForEach` will be simplified from:
```c++
meta::ForEach<FileOutputFields, GetFields<boost::mpl::_1>> ForEachGetFields;
                        ForEachGetFields(threadParams);
```
to:
```c++
meta::ForEach<FileOutputFields, GetFields> ForEachGetFields;
                        ForEachGetFields(threadParams);
```

This is also reflected in Mp11, where many meta functions have a normal version accepting a meta function and a version accepting a quoted meta function (see `_q` suffixes).

What is still less elegant is transforming this from MPL:
```c++
using UnspecializedSpeciesPlugins = pmacc::mp_list<
    plugins::multi::Master<EnergyParticles<boost::mpl::_1>>,
    plugins::multi::Master<CalcEmittance<boost::mpl::_1>>,
    plugins::multi::Master<BinEnergyParticles<boost::mpl::_1>>,
    CountParticles<boost::mpl::_1>,
    PngPlugin<Visualisation<boost::mpl::_1, PngCreator>>,
    plugins::transitionRadiation::TransitionRadiation<boost::mpl::_1>
>;
```
With Mp11 bind:
```c++
using UnspecializedSpeciesPlugins = pmacc::mp_list<
    plugins::multi::Master<pmacc::mp_bind<EnergyParticles, pmacc::_1>>,
    plugins::multi::Master<pmacc::mp_bind<CalcEmittance, pmacc::_1>>,
    plugins::multi::Master<pmacc::mp_bind<BinEnergyParticles, pmacc::_1>>,
    pmacc::mp_bind<CountParticles, pmacc::_1>,
    PngPlugin<pmacc::mp_bind<Visualisation, pmacc::_1, PngCreator>>,
    pmacc::mp_bind<plugins::transitionRadiation::TransitionRadiation, pmacc::_1>
>;
```
If we want to support the original placeholder expressions (but using Mp11 placeholders), we could write our own meta function to evaluate it. The implementation is similar to `mp_bind` and probably in the order of 50-100 LOCs.

Ultimately, we are going to need to look at the various cases and apply local judgement. But I believe we can solve it and make it look nice :) In a separate PR of course.

bernhardmgruber:
We can implement the placeholder expressions ourselves and I am considering that for LLAMA as well now. The implementation is rather simple: https://github.com/alpaka-group/llama/pull/451/files

psychocoderHPC:
> We can implement the placeholder expressions ourselves and I am considering that for LLAMA as well now. The implementation is rather simple: https://github.com/alpaka-group/llama/pull/451/files

Ohh that's cool. I will check it next year^^

psychocoderHPC:
I have this PR on my radar and will try to rebase and test it together with #3983 soon.

include/pmacc/meta/ForEach.hpp
```diff
@@ -95,36 +53,27 @@ namespace pmacc::meta
...
-        template<typename... T_Types>
-        HDINLINE void operator()(T_Types&&... ts)
-        {
-            Functor()(std::forward<T_Types>(ts)...);
-            NextCall()(ts...);
+#ifdef __NVCC__
+#    pragma nv_diag_suppress 20011
+#    pragma nv_diag_suppress 20014
+#endif
+            mp_for_each<SolvedFunctors>([&](auto functor) { functor(std::forward<T_Types>(ts)...); });
```
psychocoderHPC: This line is showing warnings with nvcc 11.2

```
/home/rwidera/workspace/picongpu/include/picongpu/../pmacc/meta/ForEach.hpp(72): warning #20014-D: calling a __host__ function from a __host__ __device__ function is not allowed
```

include/picongpu/plugins/IsaacPlugin.hpp
```diff
@@ -56,8 +56,8 @@ namespace picongpu
         public:
             static const size_t featureDim = 3;
             static const ISAAC_IDX_TYPE guardSize = 0;
-            // static const bool hasGuard = bmpl::not_<boost::is_same<FieldType, FieldJ>>::value;
-            static const bool persistent = bmpl::not_<boost::is_same<FieldType, FieldJ>>::value;
+            // static const bool hasGuard = !std::is_same_v<FieldType, FieldJ>;
```
psychocoderHPC: note to @psychocoderHPC : This dead code should be removed

include/picongpu/particles/particleToGrid/CombinedDerive.def
```diff
@@ -142,7 +142,8 @@ namespace picongpu
                     typename particles::traits::SpeciesEligibleForSolver<T_Species, T_BaseDerivedAttribute>::type;
                 using EligibleForModifierAttribute =
                     typename particles::traits::SpeciesEligibleForSolver<T_Species, T_ModifyingDerivedAttribute>::type;
-                using type = typename bmpl::and_<EligibleForBaseAttribute, EligibleForModifierAttribute>;
+                using type
+                    = std::bool_constant<EligibleForBaseAttribute::value && EligibleForModifierAttribute::value>;
```
psychocoderHPC: this should be `pmacc::mp_and` same as in the other classes changed in the folder

include/picongpu/particles/particleToGrid/derivedAttributes/WeightedVelocity.def
```diff
@@ -120,7 +120,7 @@ namespace picongpu
                     typename pmacc::traits::HasIdentifiers<FrameType, RequiredIdentifiers>::type;
                 // needs mass
                 using SpeciesHasFlags = typename pmacc::traits::HasFlag<FrameType, massRatio<>>::type;
-                using type = typename bmpl::and_<SpeciesHasIdentifiers, SpeciesHasFlags>;
+                using type = std::bool_constant<SpeciesHasIdentifiers::value && SpeciesHasFlags::value>;
```
psychocoderHPC: This should be `pmacc::mp_and` same as other changes in this folder

include/picongpu/plugins/common/openPMDWriteMeta.hpp
```diff
@@ -44,24 +44,20 @@ namespace picongpu
...
-            /** write openPMD species meta data
+            /** write meta data for species
              *
-             * @tparam numSpecies count of defined species
+             * @param threadParams context of the openPMD plugin
+             * @param fullMeshesPath path to mesh entry
              */
-            template<uint32_t numSpecies = bmpl::size<VectorAllSpecies>::type::value>
-            struct OfAllSpecies
+            void ofAllSpecies(::openPMD::Series& series, uint32_t currentStep)
```
psychocoderHPC: The function must be `inline`, we aim to use multiple translation units in the future and try to be prepared for this moment and try to avoid ambigious linker issues.

include/pmacc/math/vector/compile-time/Size_t.hpp
```diff
@@ -51,7 +49,10 @@ namespace pmacc
                 size_t y = traits::limits::Max<size_t>::value,
                 size_t z = traits::limits::Max<size_t>::value>
             struct Size_t
-                : public CT::Vector<mpl::integral_c<size_t, x>, mpl::integral_c<size_t, y>, mpl::integral_c<size_t, z>>
+                : public CT::Vector<
```
psychocoderHPC: Could be equally changed as in `Int.hpp`

```
template<size_t... T_values>
using Size_t = CT::Vector<std::integral_constant<size_t, T_values>...>;
```

include/pmacc/math/vector/compile-time/UInt32.hpp
```diff
@@ -51,8 +49,10 @@ namespace pmacc
                 uint32_t y = traits::limits::Max<uint32_t>::value,
                 uint32_t z = traits::limits::Max<uint32_t>::value>
             struct UInt32
-                : public CT::
-                      Vector<mpl::integral_c<uint32_t, x>, mpl::integral_c<uint32_t, y>, mpl::integral_c<uint32_t, z>>
+                : public CT::Vector<
```
psychocoderHPC: Could be equally changed as in `Int.hpp`

```
template<uint32_t... T_values>
using UInt32_t = CT::Vector<std::integral_constant<uint32_t, T_values>...>;
```

include/pmacc/math/vector/compile-time/UInt64.hpp
```diff
@@ -51,8 +49,10 @@ namespace pmacc
                 uint64_t y = traits::limits::Max<uint64_t>::value,
                 uint64_t z = traits::limits::Max<uint64_t>::value>
             struct UInt64
-                : public CT::
-                      Vector<mpl::integral_c<uint64_t, x>, mpl::integral_c<uint64_t, y>, mpl::integral_c<uint64_t, z>>
+                : public CT::Vector<
```
psychocoderHPC: Could be equally changed as in `Int.hpp`

include/pmacc/math/vector/compile-time/Vector.hpp
```diff
@@ -182,11 +171,12 @@ namespace pmacc
                 typename T_TypeA1,
                 typename T_TypeB0,
                 typename T_TypeB1,
+                template<typename...>
                 typename T_BinaryOperator>
             struct applyOperator<CT::Vector<T_TypeA0, T_TypeA1>, CT::Vector<T_TypeB0, T_TypeB1>, T_BinaryOperator>
             {
-                using OpResult0 = typename mpl::apply<T_BinaryOperator, T_TypeA0, T_TypeB0>::type;
-                using OpResult1 = typename mpl::apply<T_BinaryOperator, T_TypeA1, T_TypeB1>::type;
+                using OpResult0 = T_BinaryOperator<T_TypeA0, T_TypeB0>;
```
psychocoderHPC, line 178: WARNING: This transition is changing maybe silently some definitions in PMAcc or PIConGPU.
`mpl::apply` is not assuming any order `X<boost::mpl_2,boost::mpl_1>` can be used which is not equal to `T_BinaryOperator<T_TypeA0, T_TypeB0>`
Could be that this is not important here but I need to check this part of code again.

include/pmacc/particles/memory/dataTypes/Particle.hpp
```diff
@@ -209,19 +203,17 @@ namespace pmacc
...
-                     * bmpl::contains has lower complexity than traits::HasIdentifier
+                     * mp_contains has lower complexity than traits::HasIdentifier
                      * and was used for this reason
                      */
-                    using CommonTypeSeq = typename bmpl::copy_if<
-                        DestTypeSeq,
-                        bmpl::contains<SrcTypeSeq, bmpl::_1>,
-                        bmpl::back_inserter<bmpl::vector0<>>>::type;
+#if BOOST_COMP_NVCC && BOOST_COMP_NVCC < BOOST_VERSION_NUMBER(11, 3, 0)
+                    using CommonTypeSeq = mp_copy_if_q<DestTypeSeq, mp_bind_front<mp_contains, SrcTypeSeq>>;
```
psychocoderHPC: Why is this workaround required? I am wondering because `mp_set_difference` is working without any issues.
We could rewrite `mp_set_intersection` as

```C++
using UniqueInDestTypeSeq = mp_set_difference<DestTypeSeq, SrcTypeSeq>;
using CommonTypeSeq = mp_set_difference<DestTypeSeq, UniqueInDestTypeSeq>;
```
This would remove the `#if`

psychocoderHPC (changes requested):
Minor changes are required and we need to solve the `warnings` created by `meta::ForEach`

include/picongpu/particles/particleToGrid/CombinedDerive.def
bernhardmgruber: I can do that, but my code has the same effect. Why do you prefer `mp_and` over using the built-in `&&`?

include/picongpu/plugins/common/openPMDWriteMeta.hpp
bernhardmgruber: Good catch!

include/pmacc/meta/ForEach.hpp
bernhardmgruber: Replaced by custom implementation.

include/pmacc/particles/memory/dataTypes/Particle.hpp
bernhardmgruber: I created the workaround because nvcc < 11.3 produced a compile error inside `mp_set_intersection`. I can try your workaround.

bernhardmgruber: Works.

include/pmacc/math/vector/compile-time/Vector.hpp
bernhardmgruber: That depends on whether you consider `pmacc::math::CT::applyOperator` a public API that you want to use with MPL metafunctions. I saw that it was only used in the local file, so I refactored all uses of it to use the corresponding mp11 meta functions.

Your call, I can revert the code here.

psychocoderHPC: I will keep it

include/picongpu/particles/particleToGrid/CombinedDerive.def
psychocoderHPC: We do not prefer it but most other files in the folder use `mp_and`, I like to have it consistent.

include/pmacc/math/vector/compile-time/Vector.hpp
bernhardmgruber: I interpret this as: You will keep my modification proposed in this PR. Right?

psychocoderHPC: yes, let us keep it as it is.

psychocoderHPC:
I will test this tomorrow on crusher and merge it if all works as expected.

psychocoderHPC (approved):

psychocoderHPC:
@bernhardmgruber Thanks for your hard work!

bernhardmgruber:
> @bernhardmgruber Thanks for your hard work!

Most work was done on a weekend during the lockdown more than a year ago and actually fun. But pushing it over the finish line was really tough now. Thank you for helping me get it done!

PR #3835, opened 2021-09-30
Fix cell range in calorimeter plugin processing leaving particles

sbastrakov:
By mistake it was using internal instead of external range. Also slightly change some variable names there for more clarity.

cc @pordyna


pordyna:
Thanks, I was confused by the variable names. I will include similar  changes in my detector plugin.

psychocoderHPC (approved):

PR #3839, opened 2021-10-01
Remove macro __deleteArray() 

sbastrakov:
Also replace its former users that were host-side C-style arrays in plugins with `std::vector `

### DIFF
--- a/include/picongpu/plugins/output/WriteSpeciesCommon.hpp
+++ b/include/picongpu/plugins/output/WriteSpeciesCommon.hpp
@@ -158,16 +158,13 @@ namespace picongpu
 #elif(ALPAKA_ACC_GPU_HIP_ENABLED == 1)
                 CUDA_CHECK((cuplaError_t) hipHostFree(ptr));
 #else
-                __deleteArray(ptr);
+                delete[] ptr;
 #endif
             }
         }
     };
 
-    /** free memory
-     *
-     * use `__deleteArray()` to free memory
-     */
+    //! Free memory
     template<typename T_Attribute>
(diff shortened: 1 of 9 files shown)


include/picongpu/plugins/output/WriteSpeciesCommon.hpp
```diff
@@ -178,11 +175,7 @@ namespace picongpu
             typedef typename pmacc::traits::Resolve<Attribute>::type::type type;
 
             type* ptr = value.getIdentifier(Attribute()).getPointer();
-            if(ptr != nullptr)
-            {
-                __deleteArray(ptr);
-                ptr = nullptr;
-            }
+            delete[] ptr;
```
psychocoderHPC, line 178: Note: `__deleteArray(ptr)` set the pointer to `nullptr`, this part of code is now deleted. It helps a lot if a pointer is set to nullptr. In case someone is using the pointer after it is deleted you will get an exception. 
In case the pointer is not set to nullptr it is very likely that you mostly not get any error.

psychocoderHPC (approved):

include/picongpu/plugins/output/WriteSpeciesCommon.hpp
psychocoderHPC: In general I feel better if we set the ptr to `nullptr`.

sbastrakov: I know and I checked when it was used. Turned out, was never required to set to `nullptr` as either the variable went out of stack immediately, or it was a member in a class destructor

PR #3840, opened 2021-10-01
fix -d in 1.cfg in SingleParticleExample

lennertsprenger:
I discovered a missing -d flag. So when specifying a number of devices for the different axis that can be used, which wasn't possible without -d


sbastrakov (approved):

psychocoderHPC (approved):

PrometheusPi (approved):
Good catch @lennertsprenger

PR #3842, opened 2021-10-04
Fix inaccuracies in incident field solver

sbastrakov:
As discovered by Max ~~(student of @steindev , doesn't have a github account)~~ @Enderpigmen there were some inaccuracies in the incident field solver implementation. It was still working, but produced more "noise" and inaccuracies than it should have. It was missed previously as the functors I used for testing were a bit unfortunately chosen and couldn't show it. The two commits are two unrelated fixes.

Draft while waiting for Max to check it on all directions, for now he did with Y (as we originally saw issues there). Thanks to him for running lots of simulations and demonstrating the problem!

### DIFF
--- a/include/picongpu/fields/incidentField/Solver.hpp
+++ b/include/picongpu/fields/incidentField/Solver.hpp
@@ -85,6 +85,9 @@ namespace picongpu
                     //! Time increment in the target field, in iterations
                     float_X timeIncrementIteration;
 
+                    //! Whether there is an active (non-none) min source for each axis
+                    pmacc::math::Vector<bool, simDim> hasMinSource;
+
                     //! Cell description for kernels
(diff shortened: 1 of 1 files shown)


sbastrakov:
Oh, now there is an account @Enderpigmen

PrometheusPi:
Great catch @Enderpigmen - thanks for such a quick fix @sbastrakov

Enderpigmen:
I checked the incident Field Method with the fix and pure Plane Waves and it works now from all directions ( when the source is either applied at XMin, XMax, ...).

Enderpigmen:
Here is the issue in the electrical field of a plane wave starting from YMin before the fix:

Incident Field before

and after the fix


Incident Field after

psychocoderHPC (approved):

Issue #3850, opened 2021-10-05
Extend RNG state storage to guards


sbastrakov:
Currently PIConGPU only stores the states for internal (CORE + BORDER) cells of each local domain. There was previously no use in storing it for the guard as well, as there are no particles in the guard. However there is a valid use case to implement thermal boundaries as pointed out by @lennertsprenger. In case the boundary offset is 0, the logic involving RNG states has to be applied to the particles that are currently (before the end of the boundary condition implementation) in the guard.

sbastrakov:
This is still not implemented as far as I can see. The issue is not super pressing, but stops thermal particle boundaries at global domain boundaries (they require at least 1 cell offset). Reassigning to @psychocoderHPC .

PR #3858, opened 2021-10-08
Add thermal boundary condition for particles

lennertsprenger:
My unfinished work since my time at the HZDR is over

In the thermal boundary condition the particles that crossed the simulation boundary will be thermaly radiated from the point where the particle would have hit the boundary, with the velocity from the Maxwell Boltzmann distribution. 

TODO:
- [x] add temperature parameter
- [x] check that the new position after the thermalization is not outside of the boundary of another axis
- [x] extend docs

For now positive boundary offset is required (and checked for) due to #3850.

cc @steindev @sbastrakov

### DIFF
--- a/include/picongpu/particles/boundary/Thermal.hpp
+++ b/include/picongpu/particles/boundary/Thermal.hpp
@@ -29,6 +29,8 @@
 #include "picongpu/particles/manipulators/unary/FreeTotalCellOffsetRng.hpp"
 #include "picongpu/traits/attribute/GetMass.hpp"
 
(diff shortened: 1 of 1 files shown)


sbastrakov:
@lennertsprenger you were right, it is actually required to check for all crossings in thermal implementation. In that regard, it is different from absorbing and reflecting, which were fully separable. I now fixed it, and also added some comments explaining the reasoning

sbastrakov:
@psychocoderHPC ready for re-review.

PR #3873, opened 2021-10-18
Spack load: no more "-r"

ax3l:
The `-r` argument was removed from `spack load` and is now implied.

X-ref: https://github.com/ComputationalRadiationPhysics/spack-repo/pull/14


psychocoderHPC:
I will merge this PR even if the CI failed. The lines changed do not effect the CI tests, there is something else wrong with our CI.

PR #3859, opened 2021-10-08
Unroll interpolation

bernhardmgruber:
This PR places a few ` #pragma unroll`s to request the compiler to unroll the 3 loops across for the trilinear interpolation and the 1 loop over the simulation dimensions. This change mostly affects the `MoveAndMark` kernel.

SPEC benchmark is run on NVIDIA GeForce RTX 2060.

Without unrolling:
```
	dev
		calculation  simulation time: 16sec 796msec = 16.796 sec
	llama (P: SoA, F: AoS)
		calculation  simulation time: 16sec 510msec = 16.510 sec
	llama (P: SoA, F: AoSSplit)
		calculation  simulation time: 26sec 350msec = 26.350 sec
```

With unrolling:
```
	dev
		calculation  simulation time: 16sec 464msec = 16.464 sec
	llama (P: SoA, F: AoS) #pragma unroll in AssignedTrilinearInterpolation.hpp:
		calculation  simulation time: 17sec 793msec = 17.793 sec
	llama (P: SoA, F: AoSSplit) #pragma unroll in AssignedTrilinearInterpolation.hpp
		calculation  simulation time: 29sec 148msec = 29.148 sec
	llama (P: SoA, F: AoS) #pragma unroll in AssignedTrilinearInterpolation.hpp and FieldToParticleInterpolation.hpp
		calculation  simulation time: 16sec 708msec = 16.708 sec
	llama (P: SoA, F: AoSSplit) #pragma unroll in AssignedTrilinearInterpolation.hpp and FieldToParticleInterpolation.hpp
		calculation  simulation time: 16sec 601msec = 16.601 sec
```

While the unrolling has little to no effect on the current `dev` branch (although CUDA now requires less registers and is a few ms faster). The effect is VERY noticable when more LLAMA magic is involved. See especially the `llama (P: SoA, F: AoSSplit)` run. The reason for the big improvement is that the cuSTL Cursors backed by LLAMA are more complex than the BufferCursor. This comlexity can mostly be folded away when enough code is inlined and unrolled together. Also note that applying the unrolling only to the `AssignedTrilinearInterpolation` has a worsening effect.

I am too hesitant yet to propose this change as it could make the register count explode if the trilinear interpolation accesses a larger range of values. However, with LLAMA such a change is absolutely needed.

- [x] do not merge before 0.6.0 branch is created
- [x] rebase against #3880

PR #3880, opened 2021-10-20
fix using not allowed particle shape

psychocoderHPC:
fix issue detected https://github.com/ComputationalRadiationPhysics/picongpu/pull/3859#issuecomment-947638386

- add check that particles use only shapes with assignment support >= 1.
- plugin XraxScattering: change particle shape type to `NGP`


sbastrakov:
As an extra note to this. `Counter` is not a real shape, but a workaround to be used in some kernels instead of the actual particles shape. It is only suitable for particular use patterns, e.g. its internal assignment function is not self-consistent and, again, can only provide reasonable values with a certain use pattern. Thus, `Counter` should never be enabled by a user as a general particle shape.

Issue #3889, opened 2021-10-26
EmZ implementation may be incorrect in 2d


sbastrakov:
The issue is with `Jz` and the case two (virtual) particles are used. In this case, both virtual particles get half of velocity in `z`. Which leads to correct total velocity, and so correct total amount of `Jz`. But distribution of resulting grid values of `Jz` along `x, y` appears generally incorrect.

The problem is, now equal amount of current in z is always assigned to both virtual particles, regardless of their movement to and from the relay point. While in reality this amount should depend on this movement. In particular, in the limit of relay point -> start or end position of the particle, the corresponding portion of velocity in `z` must go to 0, and now it always stays half.

To test it, I made a 2d simulation with a single particle, TSC formfactor, starting temperature and no field solver, and compared `EmZ` and `Esirkepov` implementations. Here is grid `Jz(x, y)` values around the particle position for the first time the two-particle scheme is used, and an equivalent condition in Esirkepov is met (the particle is the same in both cases). In this case, the relay point is in positive x, so the first row and both outer columns are all 0 as expected (all other values are also 0 as they are too far away). But the difference in values isn't expected:

Esirkepov (which I assume is correct):
```
0.000000 0.000000 0.000000 0.000000 0.000000 
0.000000 0.000210 0.001613 0.000340 0.000000 
0.000000 0.008910 0.070403 0.015177 0.000000 
0.000000 0.006506 0.052256 0.011419 0.000000 
0.000000 0.000008 0.000062 0.000014 0.000000 
```

Same values for EmZ (which I think is incorrect):
```
0.000000 0.000000 0.000000 0.000000 0.000000 
0.000000 0.000105 0.000806 0.000170 0.000000 
0.000000 0.007964 0.066160 0.014850 0.000000 
0.000000 0.006739 0.057079 0.013002 0.000000 
0.000000 0.000003 0.000031 0.000007 0.000000 
```

I think I understand what the issue is, but not sure how to easily fix it besides doing same as in Esirkepov implementation for this case.

cc @psychocoderHPC @steindev

sbastrakov:
Solved with #3893.

Issue #3234, opened 2020-04-22
Check numerics - KHI growth rates


PrometheusPi:
As suggested in the tests of #3227, we should test KHI growth rates in PIConGPU again. Currently it looks like the current `dev` produces slightly too low growth rates. 
This could be caused by either mismatches in the box sizes and the KHI wave length in the default setup, the use of noisy particle shapes, etc. 
I will have a look into that issue.

PrometheusPi:
I just tested the KHI in a large 3D box and got too small growth rates for the eKHI regime.

![grafik](https://user-images.githubusercontent.com/5121158/80115930-a1357080-8585-11ea-8e0f-9109c66fd023.png)
Dashed lines represent the theoretically predicted growth rate of the magnetic field energy. 
@ax3l Do you remember when you tested this the last time?

PrometheusPi:
@psychocoderHPC This test used the `TSC` particle shape - the next one will use `PCS` shape.

PrometheusPi:
@KseniaBastrakova In contrast to your analysis in #3227, I did not get any negative growth rates (so far). How did you compute the growth rate?

PrometheusPi:
@KseniaBastrakova Sorry my fault, I get the same initial negative growth rates for `B_x`. But otherwise the growth rate looks similarly bad.
![grafik](https://user-images.githubusercontent.com/5121158/80118282-7ac50480-8588-11ea-9c11-5da923ab295c.png)

PrometheusPi:
In contrast to [@KseniaBastrakova growth rate data](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3227#issuecomment-613722303), which at least are closer to theoretical predictions.

psychocoderHPC:
> @KseniaBastrakova Sorry my fault, I get the same initial negative growth rates for `B_x`. But otherwise the growth rate looks similarly bad.
> ![grafik](https://user-images.githubusercontent.com/5121158/80118282-7ac50480-8588-11ea-9c11-5da923ab295c.png)

Could you please upload both plots with higher resolution, I can not see any dashed lines in the plot.

psychocoderHPC:
[deleted my last comment] so ignore it in your mails.

PrometheusPi:
The 3D data for PCS particles shows a vary similar behavior. 
[...]

PR #3892, opened 2021-10-28
Fix incorrect assignment of Jz in 2d EmZ implementation

sbastrakov:
In case two virtual particles were used, the implementation was incorrect.
It resulted in proper total amount of Jz, but generally incorrect distribution in x, y.
See #3889 for more details.

This PR implements [the suggested fix](https://github.com/ComputationalRadiationPhysics/picongpu/issues/3889#issuecomment-952019984).
Now it produces same Jz values as `Esirkepov` implementation.

Solves #3889.

We have to port this fix to the release candidate branch. This error also seems to have been present in all PIConGPU versions on github.

It is hard to evaluate what was the effect of it. The bug affected only 2d case, and only particles with the two-virtual-particles schema, so the particles that changed the closest cell or half-cell (depending on even/odd shape support) in x or y due to movement during the last time step. For those particles, EmZ violated the continuity equation in the Jz term. Values of Jx, Jy were always calculated correctly (i.e. same as in Esirkepov). With enough particles and a stable regime it may have been not a big difference, as the total current was preserved and distributed among same grid nodes, but with somewhat off coefficients. However in an instability scenario it could have all be going wrong because of this effect. Need to recheck #3234 (in case that was 2d) and the ongoing hunt for instability.

### DIFF
--- a/include/picongpu/fields/currentDeposition/EmZ/EmZ.hpp
+++ b/include/picongpu/fields/currentDeposition/EmZ/EmZ.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2016-2021 Rene Widera
+/* Copyright 2016-2021 Rene Widera, Sergei Bastrakov
  *
  * This file is part of PIConGPU.
  *
@@ -25,6 +25,7 @@
 #include "picongpu/fields/currentDeposition/RelayPoint.hpp"
 
 #include <pmacc/cuSTL/cursor/Cursor.hpp>
+#include <pmacc/meta/InvokeIf.hpp>
 
 
 namespace picongpu
@@ -104,16 +105,10 @@ namespace picongpu
                     line.m_pos1[d] = calc_InCellPos(relayPoint[d], I[0][d]);
                 }
 
-                const bool twoParticlesNeeded = I[0] != I[1];
-
-                deposit(
-                    acc,
-                    dataBoxJ.shift(I[0]).toCursor(),
-                    line,
-                    chargeDensity,
-                    velocity.z() * (twoParticlesNeeded ? float_X(0.5) : float_X(1.0)));
+                deposit(acc, dataBoxJ.shift(I[0]).toCursor(), line, chargeDensity);
 
                 /* detect if there is a second virtual particle */
+                const bool twoParticlesNeeded = (I[0] != I[1]);
                 if(twoParticlesNeeded)
                 {
                     /* calculate positions for the second virtual particle */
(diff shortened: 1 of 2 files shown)


include/picongpu/fields/currentDeposition/EmZ/EmZ.hpp
```diff
@@ -123,8 +118,47 @@ namespace picongpu
...
+                        }
+                        /* Have to use DIM2, otherwise 3d case wouldn't compile due to
+                         * no computeCurrentZ() method.
+                         * In this case it is parsed even though the invokeIf condition is false and dataBoxJ is passed
+                         * as auto&&.
+                         */
+                        emz::DepositCurrent<
+                            typename T_Strategy::BlockReductionOp,
+                            typename T_ParticleShape::ChargeAssignment,
+                            begin - 1,
```
sbastrakov, line 154: Discussed offline. Here we can have a component-wise minimum between `I[0]` and `I[1]`, and calculate relative to that point. Then have the range from `begin` to `end + 1`

psychocoderHPC: @sbastrakov and I decided to merge this PR first and provide the optimization within a second PR.

include/picongpu/fields/currentDeposition/EmZ/EmZ.hpp
```diff
@@ -123,8 +118,47 @@ namespace picongpu
...
                     }
-                    deposit(acc, dataBoxJ.shift(I[1]).toCursor(), line, chargeDensity, velocity.z() * float_X(0.5));
+                    deposit(acc, dataBoxJ.shift(I[1]).toCursor(), line, chargeDensity);
                 }
+
+                /* 2d case requires special handling of Jz as explained in #3889.
+                 * Pass dataBoxJ as auto&& to defer evaluation for 3d case.
+                 */
+                pmacc::meta::invokeIf<simDim == 2>(
+                    [&](auto&& dataBoxJ)
```
psychocoderHPC, line 128:
```diff
diff --git a/include/picongpu/fields/currentDeposition/EmZ/EmZ.hpp b/include/picongpu/fields/currentDeposition/EmZ/EmZ.hpp
index 8e2fff17f..125490eaa 100644
--- a/include/picongpu/fields/currentDeposition/EmZ/EmZ.hpp
+++ b/include/picongpu/fields/currentDeposition/EmZ/EmZ.hpp
@@ -125,7 +125,7 @@ namespace picongpu
                  * Pass dataBoxJ as auto&& to defer evaluation for 3d case.
                  */
                 pmacc::meta::invokeIf<simDim == 2>(
-                    [&](auto&& dataBoxJ)
+                    [&, this](auto&& dataBoxJ)
                     {
                         /* For Jz we consider the whole movement on a step.
                          * This movement it is not necessarily on support.
@@ -140,8 +140,8 @@ namespace picongpu
                          */
                         for(uint32_t d = 0; d < simDim; ++d)
                         {
-                            line.m_pos0[d] = calc_InCellPos(posStart[d], I[1][d]);
-                            line.m_pos1[d] = calc_InCellPos(posEnd[d], I[1][d]);
+                            line.m_pos0[d] = this->calc_InCellPos(posStart[d], I[1][d]);
+                            line.m_pos1[d] = this->calc_InCellPos(posEnd[d], I[1][d]);
                         }
                         /* Have to use DIM2, otherwise 3d case wouldn't compile due to
                          * no computeCurrentZ() method.
```

Is required to pass the CI.

The current error is:

```
/picongpu/include/pmacc/../picongpu/simulation/stage/CurrentDeposition.hpp:79:52:   required from here
/picongpu/include/pmacc/../picongpu/fields/currentDeposition/EmZ/EmZ.hpp:143:60: error: 'calc_InCellPos' was not declared in this scope
                             line.m_pos0[d] = calc_InCellPos(posStart[d], I[1][d]);
```

psychocoderHPC (changes requested):

psychocoderHPC (approved):

include/picongpu/fields/currentDeposition/EmZ/EmZ.hpp
sbastrakov: Implemented in #3904.

PR #3898, opened 2021-11-04
"PIConGPU in 5 Minutes": Short Introduction for new Users

s9105947:
Dear Maintainer,

I'm currently working through the documentation of PIConGPU and getting familiar with the software.
I oftentimes found myself lacking context, and generally missing a good *overview* over the matters discussed in the individual sections.

(I'd propose creating a new manual section "Overview" outlining general architecture, capabilities, usage, an FAQ for further reading and some short guides -- a "REAMDE++" if you will. However writing and maintaining such a section would consume considerable ressources.)

As an intermediary solution I want to contribute a very brief guide which touches some of those aspects named "PIConGPU in 5 Minutes". It explains how to get a basic example scenario running from scratch. It assumes that the reader:

- uses an existing HPC cluster with an available profile
- has a basic understanding of how a cluster works & can perform simple tasks using it
- can use the shell on a basic level

> Note: I have a good understanding of a HPC cluster, can perform advanced tasks on it and can use the shell on a more than advanced level.
> Maybe this assumptions about the reader (or the level of detail in the explanations) have to be changed.

The following questions should be discussed before accepting this PR:
(Should I open seperate issues for those?)

## Where should this guide be placed?
Currently I moved it into the documentation where an "Overview" section would be located: It is the very first TOC entry.
**This breaks the current style that all top-level TOC entries are chapters.**

toc

I considered moving it into "Installation", but that seems wrong as it covers more than just installation.

## Where should this guide be referenced from?
This PR does not include links to this new guide.
Should those be added in the doc intro & readme?

## What should this guide reference?
Please propose links to specific ressources (other documentation sections, external guides etc.) that should be linked from inside the guide.
(e.g. git into, shell intro, 

## How to explain the the example scenario (`LaserWakefield`)?
Is it sufficient for a new user to read the [description provided with the scenario](share/picongpu/examples/LaserWakefield/README.rst) in `share/picongpu/examples/LaserWakefield/README.rst`?
Should that be rephrased with a different level of detail/focus?

## Is "Further Reading" sufficient?
I did not know what should be recommended as further reading apart from "Read the remaining manual".
A short table "if you want to do X look at section Y" would be great, but I don't have enough insights to competently write that.

## Are 5 Minutes enough?
5 min might be a very tight promise -- should that guide be renamed to "PIConGPU in [10,15] Minutes" in order to keep it's promise?

### DIFF
--- a/docs/source/in5min.rst
+++ b/docs/source/in5min.rst
@@ -1,0 +1,124 @@
+.. _in5min:
+
+PIConGPU in 5 Minutes
+=====================
+
+A guide to run, but not understand PIConGPU.
+It is aimed at existing and preconfigured high performance computing clusters (“supercomputers”, also: HPC systems).
+For more information on the individual steps please refer to the manual.
+
+This guide needs **shell access** (probably via ``ssh``) and **some auxilliary programs** (e.g. ``git``).
+Consider getting familiar with the shell and git.
+Please also read the tutorial for your local HPC cluster.
+
+We will use the following directories:
+
+- ``~/src/picongpu``: source files from github
+- ``~/picongpu.profile``: load the dependencies for your local environment
+- ``~/picongpu-projects``: scenarios to simulate
+- ``/scratch/mydir``: result data of the simulation runs
+
+Please replace them whenever appropriate.
+
+Get the source
+--------------
+
+Use git to obtain the source and use the current ``dev`` branch and put it into ``~/src/picongpu``::
+
+  mkdir -p ~/src
+  git clone -b dev https://github.com/ComputationalRadiationPhysics/picongpu ~/src/picongpu
+
+Setup
+-----
+
+You need :ref:`a lot of dependencies <install-dependencies>`.
+
+If you are on a known cluster you can take a shortcut:
+Select a profile file (``NAME.profile.example``) for your local HPC cluster from ``etc/picongpu`` and copy it to your home directory::
+
+  cp ~/src/picongpu/etc/picongpu/hemera-hzdr/k80_picongpu.profile.example ~/picongpu.profile
+
+This profile determines which part of the HPC cluster (“partition”, also: “queue”) – and thereby the type of GPUs – you will use.
+
+**Maybe you have to adjust this profile file.**
+Use your favorite editor.
+If unsure use nano: ``nano ~/picongpu.profile`` (Save with Ctrl+O, exit with Ctrl+X)
+Read the file (it’s short) and apply changes if appropriate.
+
+For this guide, add after the last line::
+
+  export SCRATCH=/scratch/mydir
+
+This is the location where runtime data and all results will be stored.
+Obviously change the path according to your local setup.
+Consult the documentation of your HPC cluster where to save your data.
+**On HPC clusters this is probably not your home directory.**
+
+Now activate your profile::
+
+  source ~/picongpu.profile
+
+You will have to repeat the last command every time you want to use PIConGPU.
+
+Create a Scenario
+-----------------
+
+Create a directory and copy an example::
+
+  mkdir -p ~/picongpu-projects/tinkering
+  pic-create $PIC_EXAMPLES/LaserWakefield ~/picongpu-projects/tinkering/try01
+  cd ~/picongpu-projects/tinkering/try01
+
+Usually you would now adjust the files in the newly created directory ``~/picongpu-projects/tinkering/try01`` – for this introduction we will use the parameters as provided.
+Also note how the variable ``$PIC_EXAMPLES`` has been provided becauses you executed ``source ~/picongpu.profile`` in the previous step.
+
+Compile and Run
+---------------
+
+**Now use a compute node.**
+Your profile probably provides a helper command for that::
+
+  getDevice
+
+(You can now run ``hostname`` to see which node you are using.)
+
+Now build (from inside your scenario directory, maybe use ``cd ~/picongpu-projects/tinkering/try01``)::
+
+  pic-build
+
+This will take a while, go grab a coffee.
+If this fails, read the manual or ask a colleague.
+
+After a successfull build, run (still on the compute node, from inside your scenario directory, maybe use ``cd ~/picongpu-projects/tinkering/try01``)::
+
+  tbg -s bash -c etc/picongpu/1.cfg -t etc/picongpu/bash/mpiexec.tpl $SCRATCH/tinkering/try01/run01
+
+- ``tbg``: tool provided by PIConGPU
+- ``bash``: the “submit system”, e.g. use ``sbatch`` for slurm
+- ``etc/picongpu/1.cfg``: runtime options (number of GPUs, etc.)
+- ``etc/picongpu/bash/mpiexec.tpl``: options for the chosen submit system
+- ``$SCRATCH/tinkering/try01/run01``: not-yet-existing destination for your result files
+
+E.g. for “Hypnos” the invocation could be (invoke from the login node)::
+
+  tbg -s qsub -c etc/picongpu/1.cfg -t etc/picongpu/hypnos-hzdr/k20.tpl $SCRATCH/tinkering/try01/run01
+
+Examine the results
+-------------------
+
+Results are located at ``$SCRATCH/tinkering/try01/run01``.
+
+To view pretty pictures from a linux workstation you can use the following process (execute on your workstation, **not the HPC cluster**):
+
+0. Create a “mount point” (empty directory): ``mkdir -p ~/mnt/scratch``
+1. Mount the data directory using sshfs:
+   ``sshfs -o default_permissions -o idmap=user -o uid=$(id -u) -o gid=$(id -g) HOST:DATADIR ~/mnt/scratch/``
+   Substitute HOST with the hostname (``ssh HOST`` should connect to the HPC cluster) and DATADIR with the full path to your data directory, e.g. ``/scratch/mydir``
+2. Browse the directory using a file browser/image viewer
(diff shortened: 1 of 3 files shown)


PrometheusPi:
Thanks for preparing this pull request @s9105947 :+1: 

Regarding your first question:
> Where should this guide be referenced from?

I would suggest to make it a separate chapter/section right on top. 

Could you please answer briefly:
What is the difference to the already existing [quick start](https://picongpu.readthedocs.io/en/latest/usage/basics.html#step-by-step)?

s9105947:
"PIConGPU in 5 Minutes" is not too different from the quick start (and in fact, heavily inspired from that), however:

1. It is among the very first documents a user reads.
2. It is built for grasping concepts, not understanding them. (The quick start explains more than the bare essentials.)
3. It explains the entire process (including context for unexperienced HPC cluster users), especially installation.

Generally it has a shallower level of detail.

This submission is an ordered version of my personal notes, and maybe better as an auxiliary document **not** included in the manual. (Maybe just as a personal github gist? -- Nobody would find that though.)

PrometheusPi:
@s9105947 Thanks for adding your motivation for providing this pull request. I totally agree that this is something we need. @ComputationalRadiationPhysics/picongpu-maintainers I hope you agree. (yes: :+1:, no: :-1:)

docs/source/in5min.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Please replace them whenever appropriate.
+
+Get the source
+--------------
+
+Use git to obtain the source and use the current ``dev`` branch and put it into ``~/src/picongpu``::
+
+  mkdir -p ~/src
+  git clone -b dev https://github.com/ComputationalRadiationPhysics/picongpu ~/src/picongpu
```
PrometheusPi, line 29: By default, the `master` branch will be checked out, but above you are speaking about the `dev` branch. Perhaps either stating to use the `master` branch or adding `git checkout dev # switching to the dev brach` would help here.

docs/source/in5min.rst
```diff
@@ -0,0 +1,124 @@
...
+-----
+
+You need :ref:`a lot of dependencies <install-dependencies>`.
+
+If you are on a known cluster you can take a shortcut:
+Select a profile file (``NAME.profile.example``) for your local HPC cluster from ``etc/picongpu`` and copy it to your home directory::
+
+  cp ~/src/picongpu/etc/picongpu/hemera-hzdr/k80_picongpu.profile.example ~/picongpu.profile
+
+This profile determines which part of the HPC cluster (“partition”, also: “queue”) – and thereby the type of GPUs – you will use.
```
PrometheusPi, line 41: While we mainly use GPUs, there also CPU partitions and setups we support via alpaka. Thus perhaps using *hardware (e.g. GPU and CPU)* instead of the specific *GPU* would be better here.

docs/source/in5min.rst
```diff
@@ -0,0 +1,124 @@
...
+We will use the following directories:
+
+- ``~/src/picongpu``: source files from github
+- ``~/picongpu.profile``: load the dependencies for your local environment
+- ``~/picongpu-projects``: scenarios to simulate
+- ``/scratch/mydir``: result data of the simulation runs
+
+Please replace them whenever appropriate.
+
+Get the source
```
PrometheusPi, line 23: Regarding [Capitalization](https://docs.microsoft.com/en-us/style-guide/capitalization): everywhere else you are using *title-style capitalization*, but here you are using *sentence-style capitalization*.  Could you please adjust this title.
This is only a very minor^10 style issue, but perhaps you can change this.

docs/source/in5min.rst
```diff
@@ -0,0 +1,124 @@
...
+-----------------
+
+Create a directory and copy an example::
+
+  mkdir -p ~/picongpu-projects/tinkering
+  pic-create $PIC_EXAMPLES/LaserWakefield ~/picongpu-projects/tinkering/try01
+  cd ~/picongpu-projects/tinkering/try01
+
+Usually you would now adjust the files in the newly created directory ``~/picongpu-projects/tinkering/try01`` – for this introduction we will use the parameters as provided.
+Also note how the variable ``$PIC_EXAMPLES`` has been provided becauses you executed ``source ~/picongpu.profile`` in the previous step.
```
PrometheusPi, line 73: Perhaps one could also state something similar for `pic-create`: that is only available because the systems `$PATH` included t after sourcing. (This is the most common error when people want to create their second example and wonder why `pic-create` does not exist.)

docs/source/in5min.rst
```diff
@@ -0,0 +1,124 @@
...
+- ``bash``: the “submit system”, e.g. use ``sbatch`` for slurm
+- ``etc/picongpu/1.cfg``: runtime options (number of GPUs, etc.)
+- ``etc/picongpu/bash/mpiexec.tpl``: options for the chosen submit system
+- ``$SCRATCH/tinkering/try01/run01``: not-yet-existing destination for your result files
+
+E.g. for “Hypnos” the invocation could be (invoke from the login node)::
+
+  tbg -s qsub -c etc/picongpu/1.cfg -t etc/picongpu/hypnos-hzdr/k20.tpl $SCRATCH/tinkering/try01/run01
+
+Examine the results
```
PrometheusPi, line 106: Please adjust style

PrometheusPi (changes requested):
This is a excellent/outstanding introduction - kudos 🎉 🚀 
I only have very minor remarks.

docs/source/in5min.rst
s9105947: I've gone for `-b dev` to check out `dev` in the `git clone` call, but this might be confusing for/overlooked by novices.
-> Will add explicit line w/ `git checkout`

docs/source/in5min.rst
sbastrakov: Or *device*

docs/source/in5min.rst
PrometheusPi: Ahh sorry 🤦 - didn't see the `-b dev` flag - it was now an unintentional DAU test from my side 😓 -  I think your solution to separate the steps is a very good way to go :+1:

ax3l:
Good idea to have a tutorial style chapter.

I would not put it upfront, but instead create a new section below the common install and usage part, reading "tutorials" and making them system specific. The reason behind this is that you assume a couple of specific prequisites for those walk-throughs to be successful, such as `/scatch` or `$SCRATCH` to exist or specific templates. The moment you assume a reader abstracts those aspects away you are writing again the general usage and not a tutorial.

So just a bit of re-ordering I would say and calling it "5 Minutes to PIConGPU on Hemera @HZDR" would be ideal.

ax3l:
For contexts of sections, feel free to also add more "See also" boxes that remind readers that jump into a section to have first read prior sections. For example here and the through-linked boxes on top: https://picongpu.readthedocs.io/en/0.5.0/install/profile.html

This is a common way to make sure people randomly crawling/searching through the full doc, instead of reading it linearly, don't get lost.

docs/source/in5min.rst
s9105947: adressed in 476398ca

docs/source/in5min.rst
s9105947: addressed in 7b1164f2

docs/source/in5min.rst
s9105947: addressed in a204c5003

I didn't use "device" because it is too much of a technical term (and even though in this context it typically refers to a device used for computation, it could also refer to a network card, storage device etc.) and I couldn't come up with a good, short explanation for it.

docs/source/in5min.rst
s9105947: addressed in d0cebfcb5

Note: I don't mention `$PATH` because I think how searching for binaries works is too much information if you're unfamiliar with Linux in general.

s9105947:
I was not familiar with the classification described in <https://documentation.divio.com/>, thanks for bringing that up.

I agree that this tutorial is more helpful if it is less abstract.
I suppose most of the users are using HPC clusters, so assuming the reader is using one makes sense.

However, are most novices using Hemera @ HZDR?
If the scripts are sufficiently portable (are they?) I would suggest the title "PIConGPU in [5,10,15] Minutes on an HPC cluster" (minor^11 sidenote: "a HPC cluster" or "an HPC cluster"? "...on HPC clusters"?) leaving simple replacements up to the user. Anything more complicated than replacing `hemera[_k80]` could justify a separate document.

s9105947:
Regarding the available documentation:

This should probably be mentioned in the readme, placing a line right below the title and all the badges:

[Showcase](#) | [Reference](#) | [Installation](#) | [Workshop](#) | [Tutorial](#) | [Contributing](#) | [Publication](#)

Because even if the badges are clickable, their destination is not always clear, and not all of the links above have corresponding badges.

But that should probably be a separate discussion/PR.

docs/source/in5min.rst
steindev: Well we usually say *compute device*, don't we?

s9105947: \>_< bc7b5f438

s9105947:
I've reworked the guide, explicitly using hemera as an example for HPC clusters, moving it into the tutorials section and adding references/notes where I deemed them fitting.

**This is the version that I'd like to submit.**
Before merging these items have to be done:

- [ ] proofread by an actual novice/not a maintainer (can you contact someone w/ access to hemera?)
- [ ] squash commits (I'll do that after the review)

Any other remarks (spelling, wording, grammar, links to add etc) are still welcome

PrometheusPi:
Regarding proofreading by a novice: I will ask Max or perhaps one of the MT guys.

PrometheusPi (approved):

Issue #3899, opened 2021-11-04
Weird simulation results due to checkpoints.


finnolec:
Hello, I have found some issues with the checkpoints with the current `dev` branch (54efd251bc0df62a8f7e84b98ca5fd7f8e0c7718).  For this run I used the `LaserWakefield` example, I increased all `static constexpr` in `DefaultExchangeMemCfg` in the `memory.param` by a factor of 8 (to clean up the `stderr`, the simulation output looks identical even if the memory.param is not changed). 
I ran the first simulation on the `k80` partition on hemera where I changed the following lines in `4.cfg`:
```shell
TBG_devices_y=8

TBG_steps="4096"

TBG_checkpoints="--checkpoint.period 2048:2048 --checkpoint.backend openPMD"
TBG_plugins="!TBG_pngYX                    \
             !TBG_e_histogram              \
             !TBG_e_PSypy                  \
             !TBG_e_macroCount             \
             !TBG_checkpoints"

```
With my second simulation I used the checkpoint at `2048` by changing the checkpoints command from above to
```shell
TBG_checkpoints="--checkpoint.restart --checkpoint.backend openPMD"
```
and I created the output files with `tbg` without the `-s` flag. Then I went to the directory of the second simulation, created a `simOutput` folder and created a symlink to the checkpoints from the first simulation `ln -s PATH/TO/CHECKPOINT/FOLDER .`

What happens now is, that the png from the first simulation from before the checkpoint (2000) looks like this:
![e_png_yx_0 5_002000](https://user-images.githubusercontent.com/45068748/140287693-9a62d99e-6b1a-440b-ade3-306458df16e6.png)
And after the checkpoint (2100) (in the first simulation) it looks like this:
![e_png_yx_0 5_002100](https://user-images.githubusercontent.com/45068748/140287766-2fa762bf-1f9b-4ca1-ac05-98c92e3d9c84.png)
But in the second simulation, with the checkpoint-restart, the same timestep (2100) looks like this:
![e_png_yx_0 5_002100](https://user-images.githubusercontent.com/45068748/140287925-51d84f89-bfe0-4b17-8fd0-4c96ff769d8a.png)
So there is something wrong in the restart... Since we can see the 7 transversal/vertical lines I would assume that it comes from me using 8 longitudinal gpus, but I didn't check that. And the bubbles look obviously very different too.

@steindev told me to tag @sbastrakov and @Anton-Le

sbastrakov:
[...]

PR #3902, opened 2021-11-05
Fix restarting with moving window

sbastrakov:
The error was reported in #3899.

It seems the wrong logic was introduced in #3507. The patch offset calculation was wrong. It double-counted the shift of the moving window. So after the window have slided, the domains were loading wrong patches, which could have led to all sorts of trouble: incorrect results, crashes, etc.

We investigated with @psychocoderHPC , found the issue and made the fix. The second commit also adds a check that this situation is reported more clearly (e.g. would have helped to find this one).

### DIFF
--- a/include/picongpu/plugins/openPMD/restart/LoadSpecies.hpp
+++ b/include/picongpu/plugins/openPMD/restart/LoadSpecies.hpp
@@ -40,6 +40,7 @@
 #include <boost/mpl/vector.hpp>
 
 #include <cassert>
+#include <stdexcept>
 
 #include <openPMD/openPMD.hpp>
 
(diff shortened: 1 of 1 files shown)


psychocoderHPC (approved):

PrometheusPi:
@sbastrakov Could it have caused errors on restart? (Because I had to use the `dev`branch for the ml partiton on taurus and restart always failed.) If yes, I will recheck the setup on taurus.

sbastrakov:
I believe so. In my simulations before the fix, besides the wrong looking pngs, it seemed some particles were in a weird state. So this may have resulted in a crash as well

PR #3903, opened 2021-11-05
fix performance issue with HIP 4.3+

psychocoderHPC:
When running PIConGPU with HIP 4.3+ and in single precision the
performance is by a factor 3 worse compared to HIP 4.2.

The reason are slow shared memory atomicAdd operations.

Solution: Add a float specialization for atomics between threads which
is emulating atomicAdd via atomicCAS (as in good old days^^)

Note: This fix should be ported to alpaka too to provide better performance to other projects too.

### DIFF
--- a/include/pmacc/kernel/atomic.hpp
+++ b/include/pmacc/kernel/atomic.hpp
@@ -67,11 +67,10 @@ namespace pmacc
                 }
             };
 
-#if(!defined(__CUDA__) && ALPAKA_ACC_GPU_HIP_ENABLED == 1 && (HIP_VERSION_MAJOR * 100 + HIP_VERSION_MINOR) < 401)
-            /** HIP backend specialization for atomic add
+#if(ALPAKA_ACC_GPU_HIP_ENABLED == 1 && (HIP_VERSION_MAJOR * 100 + HIP_VERSION_MINOR) < 401)
+            /** HIP backend specialization for atomic add float
              *
              * Uses the intrinsic atomicAddNoRet available for AMD gpus only.
-             * Not compatible with HIP-nvcc.
              */
             template<typename T_Hierarchy, typename... T_AccArgs>
             struct AtomicOpNoRet<::alpaka::AtomicAdd, alpaka::AccGpuHipRt<T_AccArgs...>, float, T_Hierarchy>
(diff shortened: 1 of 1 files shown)


include/pmacc/kernel/atomic.hpp
```diff
@@ -67,8 +67,8 @@ namespace pmacc
                 }
             };
 
-#if(!defined(__CUDA__) && ALPAKA_ACC_GPU_HIP_ENABLED == 1 && (HIP_VERSION_MAJOR * 100 + HIP_VERSION_MINOR) < 401)
-            /** HIP backend specialization for atomic add
+#if(ALPAKA_ACC_GPU_HIP_ENABLED == 1 && (HIP_VERSION_MAJOR * 100 + HIP_VERSION_MINOR) < 401)
```
sbastrakov, line 70: What should happen when `HIP_VERSION_MAJOR * 100 + HIP_VERSION_MINOR` equals 401 or 403 ?

include/pmacc/kernel/atomic.hpp
```diff
@@ -85,6 +85,45 @@ namespace pmacc
                     ::atomicAddNoRet(ptr, value);
                 }
             };
+#elif(ALPAKA_ACC_GPU_HIP_ENABLED == 1 && (HIP_VERSION_MAJOR * 100 + HIP_VERSION_MINOR) >= 403)
+            /** HIP backend specialization for atomic add float
+             *
+             * atomicAdd(float*,float) into shared memory is very slow for HIP 4.3+.
+             * HIP 4.3 introdiced scoped atomics, it looks like atomicAdd for float is not always executed in global
```
sbastrakov, line 92:
```suggestion
             * HIP 4.3 introduced scoped atomics, it looks like atomicAdd for float, even on shared memory data, is always actually executed in global
```

include/pmacc/kernel/atomic.hpp
```diff
@@ -85,6 +85,45 @@ namespace pmacc
...
+             * memory. This spezilization is only overwriting the atomicAdd implementation for float, for atomics
```
sbastrakov, line 93:
```suggestion
             * memory. This specialization is only overwriting the atomicAdd implementation for float, for atomics
```

include/pmacc/kernel/atomic.hpp
psychocoderHPC: if it is >= 403 it will us ethe new workaround. For version <401 (not 401) it is using the not specialized version, the alpaka implementation.

sbastrakov (commented):
Small suggestions above

include/pmacc/kernel/atomic.hpp
sbastrakov: Yes, i can see that. I am wondering what's happening if it is between those two checks, for 401 and 402. Or do they not exist?

sbastrakov: Discussed offline: for these versions nothing has to be done.

sbastrakov (approved):

PR #3906, opened 2021-11-09
Add optionDefined() to plugins::multi::Option struct

franzpoeschel:
I need this in the TOML PR. The logic was already being used inside `Option::get<>()`, this extracts it to another method and also uses that inside `get()`.

Also, make `getName()` a `const` method, I also need that for the TOML PR.

### DIFF
--- a/include/picongpu/plugins/multi/Option.hpp
+++ b/include/picongpu/plugins/multi/Option.hpp
@@ -77,7 +77,7 @@ namespace picongpu
                  *
                  * @return name
                  */
-                std::string getName()
+                std::string getName() const
                 {
                     return m_name;
                 }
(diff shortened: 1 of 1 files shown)


include/picongpu/plugins/multi/Option.hpp
```diff
@@ -157,7 +166,7 @@ namespace picongpu
                  */
                 T_ValueType get(uint32_t idx)
                 {
-                    if(StorageType::size() <= idx)
+                    if(not optionDefined(idx))
```
sbastrakov, line 169:
```suggestion
                    if(!optionDefined(idx))
```

sbastrakov (approved):

franzpoeschel:
Please wait before merging this, I think I'll need to add another change

franzpoeschel:
Ok, added it now

sbastrakov (approved):

psychocoderHPC (approved):

PR #3911, opened 2021-11-12
fix using host function in device code

psychocoderHPC:
`std::sinh` can not be used on device side.
The bug was showing up with HIP 4.5 but in general the code should be wrong in CUDA too.

This bug should be backported to 0.6.0.

### DIFF
--- a/include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
+++ b/include/picongpu/particles/collision/binary/RelativisticBinaryCollision.hpp
@@ -220,7 +220,9 @@ namespace picongpu
                                 {
                                     a = 3.0_COLL * math::exp(-1.0_COLL * s12);
                                 }
-                                float_COLL bracket = math::exp(-1.0_COLL * a) + 2.0_COLL * u * std::sinh(a);
+                                // on host std::sinh will be used and on device the HIP/CUDA equivalent
+                                using std::sinh;
+                                float_COLL bracket = math::exp(-1.0_COLL * a) + 2.0_COLL * u * sinh(a);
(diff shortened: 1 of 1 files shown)


pordyna (approved):

sbastrakov (commented):
I do not see that your code changes anything tho? The same `std::sinh` should be called

sbastrakov (approved):
Nevermind my earlier message, I misunderstood the fix.

Note that now `sinh` on CUDA/HIP is always for `double`, like the C-style (non-`std`) `sinh`. But we do not have an existing proper abstraction for it, so passable for now.

psychocoderHPC:
The problem with `sinh` was not only that we called a host function from a device function, an additional issue is that
HIP can not be used together with OpenMP. PIConGPU is normally [disabling using OpenMP with clang-cuda](https://github.com/ComputationalRadiationPhysics/picongpu/blob/48fc84b3d1396f65633533600c7e7afdef4c1d45/include/picongpu/CMakeLists.txt#L172-L183) but not for HIP. 
During the update to alpaka 0.8.0 I get the same issue with `isnan()`, `isfinite()`, `isinf()`. We discussed it in the alpaka dev meeting and it looks like all of these functions are in C99 macros, therefore `::isnan()` can be found.
It looks like clang is using C math headers as soon as the command line option `-fopenmp=libomp` is used.

```
picongpu/thirdParty/cupla/alpaka/include/alpaka/math/isfinite/IsfiniteUniformCudaHipBuiltIn.hpp:39:28: error: no member named 'isfinite' in the global namespace; did you mean simply 'isfinite'?
                    return ::isfinite(arg);
```

PR #3918, opened 2021-11-19
Summit templates: Add performance hints

franzpoeschel:
~~Copied~~ Inspired from https://github.com/ECP-WarpX/WarpX/pull/2495/files

TODO:
- [x] Test that things still work on Summit
- [x] Maybe add this to documentation somewhere


franzpoeschel:
I could verify that writing via HDF5 still works after this change, but it is necessary to disable chunking for HDF5 to work at all (regardless of performance hints), see #3919

steindev (approved):

PR #3924, opened 2021-11-26
PMacc: header include validation

psychocoderHPC:
- remove unused files 
- fix PMacc dependencies to PIConGPU simulation dimension and data types.  (will be provided as separate PR too)
- add a test case based on alpaka's header check implementation.

This check is not removing all cyclic dependencies. The test case `pmaccHeaderCheck` is defining `PMACC_NO_TPP_INCLUDE` to avoid including some `tpp` files in `Environment.tpp`. This workaround is required to be able to check each header file from PMacc. Even with this workaround, it is a big step to avoid including issues when using PMacc.

### DIFF
--- a/include/pmacc/eventSystem/tasks/TaskReceiveMPI.hpp
+++ b/include/pmacc/eventSystem/tasks/TaskReceiveMPI.hpp
@@ -26,7 +26,6 @@
 #include "pmacc/communication/ICommunicator.hpp"
 #include "pmacc/eventSystem/events/EventDataReceive.hpp"
 #include "pmacc/eventSystem/tasks/MPITask.hpp"
-#include "pmacc/memory/buffers/Buffer.hpp"
 
 #include <memory>
 
@@ -47,7 +46,7 @@ namespace pmacc
 
         void init() override
         {
-            Buffer<TYPE, DIM>* dst = exchange->getCommunicationBuffer();
+            auto* dst = exchange->getCommunicationBuffer();
 
             this->request = Environment<DIM>::get().EnvironmentController().getCommunicator().startReceive(
                 exchange->getExchangeType(),
--- a/include/pmacc/pluginSystem/containsStep.hpp
+++ b/include/pmacc/pluginSystem/containsStep.hpp
@@ -36,7 +36,7 @@ namespace pmacc
          * @param timeStep simulation time step to check
          * @return true if step is included in the interval list else false
          */
-        HINLINE bool containsStep(std::vector<pluginSystem::Slice> const& seqTimeSlices, uint32_t const timeStep)
+        inline bool containsStep(std::vector<pluginSystem::Slice> const& seqTimeSlices, uint32_t const timeStep)
         {
             for(auto const& timeSlice : seqTimeSlices)
             {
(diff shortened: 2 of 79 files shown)


psychocoderHPC:
After over one year I finished now the PR to increase the code quallity.

include/pmacc/pluginSystem/containsStep.hpp
```diff
@@ -36,7 +36,7 @@ namespace pmacc
          * @param timeStep simulation time step to check
          * @return true if step is included in the interval list else false
          */
-        HINLINE bool containsStep(std::vector<pluginSystem::Slice> const& seqTimeSlices, uint32_t const timeStep)
+        inline bool containsStep(std::vector<pluginSystem::Slice> const& seqTimeSlices, uint32_t const timeStep)
```
steindev, line 39: offline answered question: `HINLINE` -> `inline` avoids including `pmacc/types.pp` which pulls full alpaka.

steindev: And it does not matter here since it is CPU code anyways

include/pmacc/eventSystem/tasks/TaskReceiveMPI.hpp
```diff
@@ -47,7 +46,7 @@ namespace pmacc
 
         void init() override
         {
-            Buffer<TYPE, DIM>* dst = exchange->getCommunicationBuffer();
+            auto* dst = exchange->getCommunicationBuffer();
```
steindev, line 49: offline answered question: `Buffer<TYPE, DIM>*` -> `auto*` avoids otherwise required cyclic include

include/pmacc/traits/HasIdentifiers.hpp
```diff
@@ -46,10 +46,13 @@ namespace pmacc
         template<typename T_Object, typename T_SeqKeys>
         struct HasIdentifiers
         {
-            using SeqHasIdentifiers = typename bmpl::transform<T_SeqKeys, HasIdentifier<T_Object, bmpl::_1>>::type;
+            using SeqHasIdentifiers =
+                typename boost::mpl::transform<T_SeqKeys, HasIdentifier<T_Object, boost::mpl::_1>>::type;
```
steindev, line 50: offline answered question: `bmpl` -> `boost::mpl` avoids short notation which is defined somewhere.

steindev (approved):
I am fine with the PR.

My inner minimalist is wondering, whether there is an *automated* way to reduce the number of included files to those that are absolutely necessary.
Probably by generating some tree diagrams during compilation and manually comparing to what is currently included. Of course, this has no priority...

psychocoderHPC:
> I am fine with the PR.
> 
> My inner minimalist is wondering, whether there is an _automated_ way to reduce the number of included files to those that are absolutely necessary. Probably by generating some tree diagrams during compilation and manually comparing to what is currently included. Of course, this has no priority...

Automatically is possible by removing one include and compiling, the problem would be if we have `#ifdef` around your includes to include only in special situations or if dependencies are available. 
So it is not practical possible :-(

PR #3932, opened 2021-12-03
Set back Esirkepov as default current solver

psychocoderHPC:
Revert #3599 and switch back to Esirkepov as the default current solver.

@steindev and me working currently on the EmZ publication. There are some open questions, therefore we decided to switch back to the well-known and understood Esirkepov current deposition method until all open questions are answered.

Note: all our tests show that EmZ is charge conserving and not introducing any artificially heating.

This PR will be ported back to 0.6.0 before the release.


sbastrakov (approved):

psychocoderHPC:
can be merged

PR #3913, opened 2021-11-15
Version bump: openPMD-api from 0.12.0 to 0.14.3

franzpoeschel:
openPMD-api version 0.14.3 has recently been [released](https://github.com/openPMD/openPMD-api/commit/00532a7873f1ec3eff9aa4238f303597bd039b6b) and includes fixes that are necessary for usage with C++17. With older versions, there are ABI incompatibilities if building openPMD-api with C++14, but including it into a C++17 project.

Since PIConGPU is currently being released (still on C++14), this marks a good time for bumping the minimum required version once the release is out. This also removes a number of workarounds for newer features (namely: the streaming API, the span-based storeChunk() API, dataset-specific backend configuration) that were not available in openPMD-api 0.12.0.

TODO:
- [x] Maybe wait for PIConGPU release before merging this into dev
- [x] Update CI containers ([see also](https://gitlab.com/hzdr/crp/alpaka-group-container/-/merge_requests/30))
- [x] Check HDF5 still working in parallel
    Update: There seem to be issues with chunking. This can be deactivated, but should be clarified before merging this one.

PR #3933, opened 2021-12-06
Make the span-based storeChunk API opt-in in various openPMD-based IO routines

franzpoeschel:
This affects all plugins that use openPMD, and especially that use its span-based storeChunk() API.
While this helps us reduce memory usage, it is [incompatible with compression operators](https://github.com/ornladios/ADIOS2/issues/2965) since these take advantage of working with two buffers (compression is not in-place).
So we make this feature opt-in.

Note that with the upcoming BP5 engine, there will be no need for the span-based storeChunk API any longer, so this is the right call to make.

TODO
- [x] Document
- [x] Test
- [x] Is the env var name ok? openPMD_USE_STORECHUNK_SPAN

### DIFF
--- a/docs/source/usage/plugins/openPMD.rst
+++ b/docs/source/usage/plugins/openPMD.rst
@@ -140,6 +140,13 @@ PIConGPU command line option          description
 Backend-specific notes
 ^^^^^^^^^^^^^^^^^^^^^^
 
+ADIOS2
+======
+
+The memory usage of some engines in ADIOS2 can be reduced by specifying the environment variable ``openPMD_USE_STORECHUNK_SPAN=1``.
+This makes PIConGPU use the `span-based Put() API <https://adios2.readthedocs.io/en/latest/components/components.html#put-modes-and-memory-contracts>`_ of ADIOS2 which avoids buffer copies, but does not allow for compression.
+Do *not* use this optimization in combination with compression, otherwise the resulting datasets will not be usable.
+
 HDF5
 ====
 
--- a/include/picongpu/plugins/common/openPMDVersion.def
+++ b/include/picongpu/plugins/common/openPMDVersion.def
@@ -20,6 +20,9 @@
 
 #include "openPMD/openPMD.hpp"
 
+#include <cstdlib> // std::getenv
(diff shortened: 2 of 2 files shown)


psychocoderHPC:
@franzpoeschel How can I check if we use the span API in PIConGPU somewhere else? DO I need this fix in other plugins too?

psychocoderHPC:
`openPMD_USE_STORECHUNK_SPAN` IMO the name is ok.

franzpoeschel:
@psychocoderHPC You can see all places where it is used in [this commit](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3913/commits/554c38b618f15bdf3f602315f67e547cd4006935). Every usage of the span-based storeChunk API go through the wrapper in openPMDVersion.def, so this PR should cover every instance.

franzpoeschel:
I've added documentation for the openPMD plugin now. The Radiation plugin, while using the span-based storeChunk API, currently only supports HDF5 output and is unaffected, so I did not add documentation.

psychocoderHPC:
> The Radiation plugin, while using the span-based storeChunk API, currently only supports HDF5 output and is unaffected, so I did not add documentation.

Does this mean if we would you the radiation plugin with Adios it will fail?
I am asking because I do not know how I can see if the span-based API is used or not. I do not know the openPMD interfaces therefore it is for me not obvious to find affected code.

franzpoeschel:
> DO I understand your answers correctly and all plugins using OpenPMD in PIConGPU automatically get patched by this PR because they use a helper you fixed?

Yes, otherwise they wouldn't work with openPMD 0.12.0


> Does this mean if we would you the radiation plugin with Adios it will fail?

The other way around, there's no reason why the radiation plugin shouldn't work with Adios, but the interface doesn't give you the option
The radiation still already uses the span-based interface of openPMD, but openPMD then uses a fallback for HDF5

psychocoderHPC (approved):

franzpoeschel:
Note: openPMD-api 0.14.4 is out and fixes this, so we can automatically enable this again

PR #3937, opened 2021-12-08
Add missing communication in XrayScattering

pordyna:
I think we discussed when I was writing  the `XrayScattering` plugin that the `GUARD` to `BORDER` communication is not required but I think this was wrong. I finally understood what this does and I think this is missing. Without the communication the contributions from particles on the edges of the GPUs are not correctly added to the adjacent GPU borders. I was using `Counter` particle shape in my tests back then so this never caused any problems. Note, this is how it is done in other places in `PIConGPU`.


bussmann:
With particles you mean electrons? If you compute an average electron density from the macroparticle electrons the shape function plays a role. If you use the electrons as point scatterers, then it doesn't. But then, what does the electron weight even mean?

pordyna:
@bussmann Yes, I mean the electrons and I was computing the average electron density. But the field contributions from the adjacent GPUs were missing.

psychocoderHPC (approved):

PR #3940, opened 2021-12-09
Add --ntasks-per-node and -np to .tpl files on Hemera

### LINKED ISSUE #3926: Wrong configuration of Hemera .tpl files
Issue reported by @Anton-Le . The current .tpl files in `dev` do not work with multuple nodes, both when they are fully used and not fully. For fully used nodes reverting the changes in #3853 helps. Not fully utilized nodes do not work either way. I suspect something changed on Hemera again and we need to readjust the profiles

sbastrakov:
Fixes #3926. It reverts #3853 and also adds `-np` to `mpiexec`. The new combination seems to work on Hemera for all cases: 1 node (full and not full), several full nodes, several not full nodes.

I think the change should be backported to the release candidate branch as well.

### DIFF
--- a/etc/picongpu/hemera-hzdr/defq.tpl
+++ b/etc/picongpu/hemera-hzdr/defq.tpl
@@ -27,6 +27,7 @@
 #SBATCH --job-name=!TBG_jobName
 #SBATCH --nodes=!TBG_nodes
 #SBATCH --ntasks=!TBG_tasks
+#SBATCH --ntasks-per-node=!TBG_mpiTasksPerNode
 #SBATCH --cpus-per-task=!TBG_coresPerCPU
 #SBATCH --mem=!TBG_memPerNode
 # must be 1 else we can not use hyperthreading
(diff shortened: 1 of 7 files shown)


etc/picongpu/hemera-hzdr/defq.tpl
```diff
@@ -104,6 +105,6 @@ export OMPI_MCA_io=^ompio
 
 if [ $? -eq 0 ] ; then
   # Run PIConGPU
-  source !TBG_dstPath/tbg/handleSlurmSignals.sh mpiexec --bind-to none !TBG_dstPath/tbg/cpuNumaStarter.sh \
+  source !TBG_dstPath/tbg/handleSlurmSignals.sh mpiexec -np !TBG_tasks --bind-to none !TBG_dstPath/tbg/cpuNumaStarter.sh \
```
psychocoderHPC, line 108: In my opinion, this change points to the fact that mpiexec is not working correctly together with SLURM. The variables passed in the job script are not taken into account by MPI.
It looks like MPI is not compiled against SLURM.

sbastrakov: It works on N full nodes without it as well. But for e.g. 6 tasks and 4 tasks per node on v100 partition mpiexec by default incorrectly launches 8 processes. I think it supports your idea

psychocoderHPC (approved):

sbastrakov:
Can be merged

PR #3941, opened 2021-12-13
HIP: allow full architecture definition

psychocoderHPC:
The architecture string for AMD GPUs could contain `:`. The old way how we
separated the backend from the architecture details is therefore not
working.

```
pic-build -b "hip:gfx906:sramecc+:xnack-"
```

I marked this behavior as bug but it is somehow more a feature. There is no need to backport this PR to the current ongoing release process of 0.6.0.

### DIFF
--- a/bin/pic-configure
+++ b/bin/pic-configure
@@ -52,7 +52,9 @@ help()
 
 get_backend_flags()
 {
-    backend_cfg=(${1//:/ })
+    # Everything before the first : is the backend, the rest is the architecture definition which could contain : too.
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
I disabled compile-tests for this PR. This functionality can not be tested.
All PIConGPU tested passed already in this run https://gitlab.com/hzdr/crp/picongpu/-/pipelines/428194957 but runtime tests for HIP will fail anyway because #3942 is required to pass the CI.

sbastrakov (approved):

PR #3942, opened 2021-12-13
CI: change AMD CI node configuration

psychocoderHPC:
- The old AMD Fiji double slot GPU from the CI is removed therefore we must use device zero (default).
- The second commit is disabling ISAAC usage for the foilLCT test because the CI is very often running out of memory.
- fix that we build HIP tests for the wrong architecture, this was overseen in #3915

Backport to 0.6.0 required


sbastrakov (dismissed):

psychocoderHPC:
can be merged

steindev (approved):

PR #3943, opened 2021-12-13
add combined derived attributes

pordyna:
# Combined derived attributes

This PR introduces derived attributes which are a function of two primary derived attributes (such like we had until now). This is something I need for my online Faraday Rotation calculation but I found this a generally useful feature. 

What I needed was 1/gamma^2 * density, where gamma is the Lorentz factor for the average kinetic energy in the cell. Because of the square it wasn't possible to just add another usual derived field. This new combined field is introduced as `RelativisticDensity`. 

This PR also includes a template for an average value of any derived quantity `AverageAttribute` ( from: average value = total quantity density / number density) as an example of a more general use of this feature. 

I also changed how the `computeValue` algorithm is called in the `openPMD` and `isaac` plugins so that the same syntax can be used for computing a normal and combined derived attribute. 

I think in most cases writing both fields individually would be a simpler solution but when fields are written or streamed for a large number of iterations and output size starts to really matter this could be a great help (definitely would be in my Faraday Rotation simulations that I was running earlier this year).  And it can be used internally , wherever a more general solution is needed (like in my `Scatterer` stage implelemtation). 

I tested this for  `AverageAttribute<Energy>` with the `FoilLTC` example and could verify that it is same as writing `EnergyDensity` and `Denisty` individually and dividing the first by the second (and setting to 0 wherever Density is 0).

@ComputationalRadiationPhysics/picongpu-maintainers

### DIFF
--- a/include/picongpu/particles/particleToGrid/ComputeFieldValue.hpp
+++ b/include/picongpu/particles/particleToGrid/ComputeFieldValue.hpp
@@ -118,7 +118,7 @@ namespace picongpu
                     fieldTmp1.template computeValue<AREA, typename CombinedSolver::BaseAttributeSolver, T_Filter>(
                         *speciesTmp,
                         currentStep);
-                    /* Particles can contribute to cells in GUARD (due to their shape) this values need to be
+                    /* Particles can contribute to cells in GUARD (due to their shape) these values need to be
                      * added to the neighbouring GPU BORDERs.
                      * Start adding field contribution from the GUARD to the adjacent GPUs while the second attribute
                      * is computed. */
--- a/include/picongpu/particles/particleToGrid/ComputeFieldValue.hpp
+++ b/include/picongpu/particles/particleToGrid/ComputeFieldValue.hpp
@@ -1,0 +1,143 @@
+/* Copyright 2021 Pawel Ordyna
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/simulation_defines.hpp"
+
+#include "picongpu/fields/FieldTmp.hpp"
+#include "picongpu/particles/particleToGrid/CombinedDerive.def"
+#include "picongpu/particles/particleToGrid/ComputeGridValuePerFrame.def"
+
+#include <pmacc/Environment.hpp>
+
+
+namespace picongpu
+{
+    namespace particles
+    {
+        namespace particleToGrid
+        {
+            /** Compute a derived particle attribute
+             *
+             * @tparam AREA area to compute values in
+             * @tparam T_Solver derived attribute solver
+             * @tparam T_Species particle species to derive the attribute from
+             * @tparam T_Filter particle filter used to filter contributing particles
+             */
+            template<uint32_t AREA, typename T_Solver, typename T_Species, typename T_Filter>
+            struct ComputeFieldValue
+            {
+                /** Functor implementation
+                 *
+                 * @param fieldTmp tmp field for storing the result
+                 * @param currentStep current simulation step
+                 * @param extraSlotNr fieldTmp memory slot to use for processing additional tmp fields possibly
+                 * required for the computation. Can be reused after calling this functor.
+                 *
+                 * @returns pointer to a TaskEvent for handling a possibly unfinished asynchronous communication task
+                 * One should call
+                 * @code{c++}
+                 *  if (returnedPointer != nullPtr)
+                 *      __setTransactionEvent(*returnedPointer);
+                 *     @endcode
+                 * after calling this functor.
+                 * Moving this outside of this functor allows for doing it just before the field values are needed and
+                 * possibly taking an advantage of parallel execution alongside with other tasks. The nullPtr check is
+                 * needed since some specializations of this method may not need it and will return a nullptr instead.
+                 */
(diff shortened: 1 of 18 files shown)


psychocoderHPC:
I will stop the CI for your job, you need to rebase against #3942 or wait until #3942 is merged into the dev branch, else the CI will fail.

pordyna:
Ok, I will just wait for #3942.

psychocoderHPC:
Please rebase against the current dev branch to pull the latest fixes to be able to pass the CI.

pordyna:
done :)

include/picongpu/particles/particleToGrid/ComputeFieldValue.hpp
```diff
@@ -0,0 +1,143 @@
...
+                    auto speciesTmp = dc.get<T_Species>(T_Species::FrameType::getName(), true);
+                    // initialize both fields with zero
+                    fieldTmp1.getGridBuffer().getDeviceBuffer().setValue(FieldTmp::ValueType::create(0.0));
+                    fieldTmp2->getGridBuffer().getDeviceBuffer().setValue(FieldTmp::ValueType::create(0.0));
+
+                    // Derive the first attribute from the particle data.
+                    fieldTmp1.template computeValue<AREA, typename CombinedSolver::BaseAttributeSolver, T_Filter>(
+                        *speciesTmp,
+                        currentStep);
+                    /* Particles can contribute to cells in GUARD (due to their shape) this values need to be
```
PrometheusPi, line 121:
```suggestion
                    /* Particles can contribute to cells in GUARD (due to their shape) these values need to be
```

include/picongpu/particles/particleToGrid/combinedAttributes/AverageAttribute.hpp
```diff
@@ -0,0 +1,78 @@
...
+        namespace particleToGrid
+        {
+            namespace combinedAttributes
+            {
+                template<typename T_Acc>
+                HDINLINE void AverageDivideOperation::operator()(T_Acc const& acc, float1_X& dst, const float1_X& dens)
+                    const
+                {
+                    // avoid dividing by zero. Return zero if density is close to zero.
+                    if(dens[0] <= std::numeric_limits<float_X>::min())
```
PrometheusPi, line 42: How do you avoid close to infinite results caused by division of barely zero (e.g.  if you are in a vacuum boarder where macro-particles in neighboring cells just contribute very little due to their shape and thus you divide by a ver y small number?)

include/picongpu/plugins/IsaacPlugin.hpp
```diff
@@ -131,17 +131,21 @@ namespace picongpu
...
-
-                    fieldTmp->getGridBuffer().getDeviceBuffer().setValue(FieldTmp::ValueType(0.0));
-                    fieldTmp->template computeValue<CORE + BORDER, FrameSolver, ParticleFilter>(
-                        *particles,
-                        *currentStep);
-                    EventTask fieldTmpEvent = fieldTmp->asyncCommunication(__getTransactionEvent());
+                    constexpr uint32_t requiredExtraSlots
+                        = particles::particleToGrid::RequiredExtraSlots<FrameSolver>::type::value;
+                    PMACC_CASSERT_MSG(
+                        _please_allocate_at_least_one_or_two_when_using_combined_attributes_FieldTmp_in_memory_param,
```
PrometheusPi, line 137: Missing what you need to allocate (one or two of what?)

PrometheusPi (commented):
Overall looks really good to me - just a few remarks and questions

include/picongpu/particles/particleToGrid/combinedAttributes/AverageAttribute.hpp
pordyna: @PrometheusPi  Well, i hope this won't happen. At least, I didn't see it happen. Whenever density is very small the total quantity should also be very small (it scales with weighting as well) so they should cancel out. 

Btw. it probably should be:
```suggestion
                    if(dens * particles::TYPICAL_NUM_PARTICLES_PER_MACROPARTICLE * CELL_VOLUME 
                        <= std::numeric_limits<float_X>::min())
```

Maybe the limit should also be a bit higher? One could even go with `< 1.0` and set to zero for all cells with the number of real particles below 1 and just properly document this behavior. But I guess that would be some lost information.

include/picongpu/plugins/IsaacPlugin.hpp
pordyna:
```suggestion
                        _please_allocate_at_least_one_FieldTmp_slot_in_memory_param_or_two_when_using_combined_attributes,
```

PrometheusPi: I like that error message.

include/picongpu/particles/particleToGrid/combinedAttributes/AverageAttribute.hpp
PrometheusPi: I think you are right - it should cancel out as long as the power of numerator is not `< 1`. 

I like the `< 1.0` limit, but I agree with you that this might become some lost information and additionally might cause some strange errors for simulations with weightings below 1 (2D TNSA simulations). 

I would think keeping it for now and seeing, whether issues arise at vacuum boarders, might be a more practical approach. I assume you will use this methods extensively in the future for simulations where there is also vacuum involved, correct? Then you will  see whether issues arise and can fix it in a followup pull request.

PrometheusPi (dismissed):

pordyna:
@PrometheusPi Thanks for your review! I just added this suggestion https://github.com/ComputationalRadiationPhysics/picongpu/pull/3943#discussion_r772976773

PrometheusPi (approved):
Thanks for implementing the suggestions.

psychocoderHPC (approved):

PrometheusPi:
@psychocoderHPC Do we actually perform compile test with ISAAC in PIConGPU? 
When compiling ISAAC with the current dev I run into the following compile error:
```
In file included from /.../picongpu/include/picongpu/main.cpp:23:
In file included from /.../picongpu/include/pmacc/../picongpu/simulation_defines.hpp:55:
In file included from /.../picongpu/include/pmacc/../picongpu/unitless/starter.unitless:23:
In file included from /.../picongpu/include/pmacc/../picongpu/plugins/PluginController.hpp:69:
/.../picongpu/include/pmacc/../picongpu/plugins/IsaacPlugin.hpp:142:25: error: no member named 'ComputeFieldValue' in namespace 'picongpu::particles::particleToGrid'
                        ComputeFieldValue<CORE + BORDER, FrameSolver, ParticleType, ParticleFilter>()
```

PR #3945, opened 2021-12-15
Extended JSON configuration of openPMD plugin: Erase dataset options if no default is given

franzpoeschel:
I noticed this while testing that #3933 actually fixes the bugs fixed by it (it does). If using the extended PIConGPU-defined JSON schema for specifying dataset-specific compression and *not* specifying a default pattern, the pattern list under `json[<backend>]["dataset"]` will be forwarded as-is to the openPMD-api. This does not actually make the openPMD-api crash, but it's still a bug on PIConGPU's side.

Fix: If no pattern is found, delete that key.

```json
{
  "adios2": {
    "unused": "parameter",
    "engine": {
      "usesteps": true,
      "rabimmel": "rambamel",
      "parameters": {
        "InitialBufferSize": "2Gb",
        "Profile": "On"
      }
    },
    "dataset": [
      {"cfg": {}}, // <--- default pattern
      {
        "select": ".*/E/.*",
        "cfg": {
          "operators": [
            {
              "type": "bzip2"
            }
          ]
        }
      }
    ]
  }
}
```

### DIFF
--- a/include/picongpu/plugins/openPMD/Json.cpp
+++ b/include/picongpu/plugins/openPMD/Json.cpp
@@ -47,6 +47,7 @@ namespace picongpu
                 // simple layout: only one global JSON object was passed
                 // forward this one directly to openPMD
                 m_patterns.emplace_back("", std::make_shared<nlohmann::json>(config));
+                m_defaultConfig = config;
             }
             else if(config.is_array())
             {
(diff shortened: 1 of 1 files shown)


include/picongpu/plugins/openPMD/Json.cpp
```diff
@@ -47,6 +47,7 @@ namespace picongpu
                 // simple layout: only one global JSON object was passed
                 // forward this one directly to openPMD
                 m_patterns.emplace_back("", std::make_shared<nlohmann::json>(config));
+                m_defaultConfig = config;
```
franzpoeschel, line 50: This line was not needed before this fix, now it is.

psychocoderHPC:
I stopped the CI for your PR because we need #3942 in dev branch and after that you must rebase else not all tests will pass the CI.

psychocoderHPC:
Please rebase against the current dev branch to pull the latest fixes to be able to pass the CI.

franzpoeschel:
Done

psychocoderHPC (approved):

PR #3949, opened 2021-12-19
Switch to C++17

bernhardmgruber:
This PR switches picongpu to C++17 in CMake. It adds two more commits using C++17 features to see that the switch is propagated successfully to the compilers.


bernhardmgruber:
The move to C++17 will also entail to drop support for CUDA < 11.0 with nvcc (we can still have older CUDA versions with clang). I guess this needs changes in the CI job generator. I had a look into `n_wise_generator.py` but I would rather have @psychocoderHPC take a look and do it properly.

psychocoderHPC:
> The move to C++17 will also entail to drop support for CUDA < 11.0 with nvcc (we can still have older CUDA versions with clang). I guess this needs changes in the CI job generator. I had a look into `n_wise_generator.py` but I would rather have @psychocoderHPC take a look and do it properly.

Yes, for clang-cuda we keep the support for CUDA 10.X. We only drop combinations which not supporting C++17.
Therefore nvcc required CUDA 11.X.

Yes I will push a commit on top of this branch to remove some combinations.

psychocoderHPC:
It would be nice if we could use this PR to switch to C++17 first and than apply C++17 feature e.g. `_v` where possible, ... this will keep the PR shorter and we do not need to handle issues we could introduce from using new features. 
The reason is that even with C++17 it could be that some compilers do not support all features. Keep in mind that in alpaka we have a workaround for INTELs compiler for C++14 `std::decay_t` :-(

psychocoderHPC:
Looks like clang9 and clang10 is not supporting `std::is_constructable_v` if you like to create a `float` from an `int`.
For GCC this required gcc11+ too, I am not 100% if this is the failure in the CI but it looks like this.
https://gcc.godbolt.org/z/54encfb6o

psychocoderHPC:
I missed that the first error is something with `std::is_same_v`

```
In file included from /builds/hzdr/crp/picongpu/include/picongpu/main.cpp:23:
In file included from /builds/hzdr/crp/picongpu/include/pmacc/../picongpu/simulation_defines.hpp:27:
In file included from /builds/hzdr/crp/picongpu/include/pmacc/../picongpu/simulation_types.hpp:30:
/builds/hzdr/crp/picongpu/include/pmacc/../pmacc/meta/ForEach.hpp:47:74: error: no template named 'is_same_v' in namespace 'std'; did you mean 'is_same'?
            template<typename itBegin, typename itEnd, bool isEnd = std::is_same_v<itBegin, itEnd>>
                                                                    ~~~~~^~~~~~~~~
                                                                         is_same
```

bernhardmgruber:
That error definitely shows, that the compiler is not compiling in C++17 mode. If you look at the next line in the error, there is:
```
/builds/hzdr/crp/picongpu/include/pmacc/../pmacc/meta/ForEach.hpp:47:74: error: no template named 'is_same_v' in namespace 'std'; did you mean 'is_same'?
            template<typename itBegin, typename itEnd, bool isEnd = std::is_same_v<itBegin, itEnd>>
                                                                    ~~~~~^~~~~~~~~
                                                                         is_same
/usr/bin/../lib/gcc/x86_64-linux-gnu/7.5.0/../../../../include/c++/7.5.0/type_traits:1503:12: note: 'is_same' declared here
```
So you see that clang/nvcc use the gcc 7.5 standard library. gcc 7.5 does support `std::is_same_v`: https://godbolt.org/z/nGPYnrjYa. Also clang-9 does support it (with whatever standard library godbolt uses). And clang-9 for CUDA also supports it: https://godbolt.org/z/758nsfM4v. So I guess the C++17 flag is not correctly arriving at clang.

Can I somehow show the build flags that are used for compilation in the CI?

bernhardmgruber:
> It would be nice if we could use this PR to switch to C++17 first and than apply C++17 feature e.g. _v where possible, ... this will keep the PR shorter and we do not need to handle issues we could introduce from using new features.

I can strip the feature refactorings from the PR once it builds successfully.

psychocoderHPC:
> I can strip the feature refactorings from the PR once it builds successfully.

Not directly in the CI, you must build the example within the container. There is a way to enforce that that CMake is using `-stdc++17` instead of `-gnu++17`. I do not remember where I already fixed it. Maybe it was in alpaka.

psychocoderHPC:
I am currently loading the container to check the paramteres.

psychocoderHPC:
The problem is that `ALPAKA_CXX_STANDARD` is by default c++14. I am currently checking the fix in PMacc and will push fix soon.

psychocoderHPC:
I pushed a fix to this PR.

psychocoderHPC:
> > It would be nice if we could use this PR to switch to C++17 first and than apply C++17 feature e.g. _v where possible, ... this will keep the PR shorter and we do not need to handle issues we could introduce from using new features.
> 
> I can strip the feature refactorings from the PR once it builds successfully.

Yes this would be nice.

psychocoderHPC:
I pushed again because we use in PIConGPU a version where https://github.com/alpaka-group/alpaka/issues/1415 is not solved.

here my command line commands I used to build the broken case interactive
```
docker run -it registry.gitlab.com/hzdr/crp/alpaka-group-container/alpaka-ci-ubuntu18.04-cuda100-clangpic:2.0
git clone -b cpp17 https://github.com/bernhardmgruber/picongpu.gi
./picongpu/bin/pic-create ./picongpu/share/picongpu/examples/Bremsstrahlung/ brems
cd brems/
../picongpu/bin/pic-build --cmake " -DISAAC_MAX_FUNCTORS=1 -DCMAKE_BUILD_TYPE=Release -DCMAKE_CUDA_FLAGS=--no-cuda-version-check -DCMAKE_CXX_COMPILER=clang++-9 -DBOOST_ROOT=/opt/boost/1.66.0 -DCMAKE_CUDA_COMPILER=clang++-9 -DPIC_USE_openPMD=OFF -DPIC_USE_PNGwriter=OFF -DCMAKE_CXX_FLAGS=-D_FORTIFY_SOURCE=0"
```

psychocoderHPC (approved):

Issue #3954, opened 2022-01-04
Issue with Incident Field Laser example


finnolec:
Hello,

I tried to run the example for the incident field laser `IncidentField` with `1.cfg` with the current `dev` branch without any further changes on hemera and I noticed a difference in the png output. I would expect the pngs to look the same as in #3592 but my output looks like this:
![out_0](https://user-images.githubusercontent.com/45068748/148072413-e5c05f9f-f50c-47e3-b433-048534652408.gif)
The main difference is the bright spot in the bottom right. There is apparently some weird field artifact there... First I thought that this is just cut off in the original PR-gif because the original gif has the shorter duration, but if one looks closely at my new gif, one can see that the bright spot starts to appear already earlier, e.g. at t=500:
![e_png_yx_0 5_000500](https://user-images.githubusercontent.com/45068748/148072957-9d0f01b2-54b4-43b2-aea2-9cfa1f76c473.png)
and t=520:
![e_png_yx_0 5_000520](https://user-images.githubusercontent.com/45068748/148073039-75ce67d9-e57d-463e-8d5a-6769bb1c6b05.png)

Comparison: Just the last frame of the original PR gif:
![lastgifframepr](https://user-images.githubusercontent.com/45068748/148073582-19df7a86-8e13-415f-a445-016c0f01d598.png)

In a second simulation I played a bit around with the example incident field laser parameters as follows:
```cpp
static constexpr uint32_t DURATION_STEPS = 100;
/// These need to be changed together with grid size
static constexpr uint32_t CENTER_X_CELLS = 64;
static constexpr uint32_t WIDTH_X_CELLS = 64;
static constexpr uint32_t CENTER_Y_CELLS = 32;
static constexpr uint32_t WIDTH_Y_CELLS = 16;
```
(the simulation is 128x128x128) and it results into this gif:
![out_1](https://user-images.githubusercontent.com/45068748/148073810-890f0571-58f5-4e0e-9a5f-b306fd4dce54.gif)

Where I now get 2 bright spots instead of 1 on the right hand side.

This is probably interesting for @sbastrakov

finnolec:
Also interesting: If you look at the spot (zoom in a lot) where I found the bright spot in my simulation but in the original gif, 
![og_gif](https://user-images.githubusercontent.com/6825656/115287811-fc765080-a150-11eb-807f-d429bcf41679.gif)
[...]

PR #3955, opened 2022-01-05
Adjust laser in incident field example for a more realistic setup

sbastrakov:
The goal is to make it easier to build an own setup based on the example.

The laser parameters are now defined in SI, and only converted to cell values when necessary. Proper delay added to avoid a steep jump of amplitude from 0, which caused artifacts as it should have with TF/SF. Png output is reduced in time to stop when only noise and artifacts remain (since there is no convenient way to make a proper axis scaling of png in such a setup).

Note that it is **not** what caused #3954, however was found while investigating it.

### DIFF
--- a/share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
+++ b/share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
@@ -38,31 +38,37 @@ namespace picongpu
             {
             public:
                 // Constants for pulse generation
-                static constexpr float_64 TIME_PERIOD_SI = 0.038e-9; // 0.038 ns
+                static constexpr float_64 TIME_PERIOD_SI = 0.0211e-9; // 0.0211 ns
                 static constexpr float_64 WAVE_LENGTH_SI = TIME_PERIOD_SI * SI::SPEED_OF_LIGHT_SI;
                 static constexpr float_64 UNITCONV_A0_to_Amplitude_SI = -1.0 / WAVE_LENGTH_SI * SI::ELECTRON_MASS_SI
                     * SI::SPEED_OF_LIGHT_SI * SI::SPEED_OF_LIGHT_SI / SI::ELECTRON_CHARGE_SI;
                 static constexpr float_64 A0 = 8.0;
                 static constexpr float_64 AMPLITUDE_SI = A0 * UNITCONV_A0_to_Amplitude_SI;
-                static constexpr float_64 DURATION_SI = 0.6e-9; // 0.6 ns
+                static constexpr float_64 DURATION_SI = 0.026e-9; // 0.026 ns
 
                 /// These need to be changed together with grid size
-                static constexpr float_64 CENTER_X_SI = 80.0e-3; // 80 mm
-                static constexpr float_64 WIDTH_X_SI = 40.0e-3; // 40 mm
-                static constexpr float_64 CENTER_Y_SI = 40.0e-3; // 40 mm
-                static constexpr float_64 WIDTH_Y_SI = 20.0e-3; // 20 mm
+                static constexpr float_64 LASER_CENTER_X_SI = 80.0e-3; // 80 mm
+                static constexpr float_64 LASER_WIDTH_X_SI = 8.0e-3; // 8 mm
+                static constexpr float_64 LASER_CENTER_Y_SI = 40.0e-3; // 40 mm
+                static constexpr float_64 LASER_WIDTH_Y_SI = 8.0e-3; // 8 mm
+                static constexpr float_64 LASER_CENTER_Z_SI = 64.0e-3; // 64 mm
+                static constexpr float_64 LASER_WIDTH_Z_SI = 10.0e-3; // 10 mm
 
                 // Convert to grid units
                 static constexpr uint32_t DURATION_STEPS = DURATION_SI / SI::DELTA_T_SI;
-                static constexpr uint32_t CENTER_X_CELLS = CENTER_X_SI / SI::CELL_WIDTH_SI;
-                static constexpr uint32_t WIDTH_X_CELLS = WIDTH_X_SI / SI::CELL_WIDTH_SI;
-                static constexpr uint32_t CENTER_Y_CELLS = CENTER_Y_SI / SI::CELL_HEIGHT_SI;
-                static constexpr uint32_t WIDTH_Y_CELLS = WIDTH_Y_SI / SI::CELL_HEIGHT_SI;
+                static constexpr uint32_t LASER_CENTER_X_CELLS = LASER_CENTER_X_SI / SI::CELL_WIDTH_SI;
+                static constexpr uint32_t LASER_WIDTH_X_CELLS = LASER_WIDTH_X_SI / SI::CELL_WIDTH_SI;
+                static constexpr uint32_t LASER_CENTER_Y_CELLS = LASER_CENTER_Y_SI / SI::CELL_HEIGHT_SI;
+                static constexpr uint32_t LASER_WIDTH_Y_CELLS = LASER_WIDTH_Y_SI / SI::CELL_HEIGHT_SI;
+                static constexpr uint32_t LASER_CENTER_Z_CELLS = LASER_CENTER_Z_SI / SI::CELL_DEPTH_SI;
+                static constexpr uint32_t LASER_WIDTH_Z_CELLS = LASER_WIDTH_Z_SI / SI::CELL_DEPTH_SI;
+
+                // Delay at initialization to ensure a smooth start
+                static constexpr float_64 initializationDurationSI = 16.0 * DURATION_SI;
 
                 // Get value to calculate Ey / Bz for the XMin source
                 HDINLINE float_X getXMinSourceValueSI(const floatD_X& totalCellIdx, const float_X currentStep) const
                 {
-                    auto const initializationDurationSI = 16.0 * DURATION_SI;
                     auto const normalizedDuration
                         = (currentStep * SI::DELTA_T_SI - 0.5 * initializationDurationSI) / DURATION_SI;
                     auto const k = 2.0_X * M_PI / WAVE_LENGTH_SI;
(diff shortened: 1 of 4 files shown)


share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
```diff
@@ -50,25 +51,36 @@ namespace picongpu
...
+                // Get value to calculate Ey / Bz for the XMin source
+                HDINLINE float_X getXMinSourceValueSI(const floatD_X& totalCellIdx, const float_X currentStep) const
                 {
                     auto const durationStepsMiddle = DURATION_STEPS / 2;
                     auto const normalizedDuration = (currentStep - durationStepsMiddle) / durationStepsMiddle;
-                    auto const timeSI = currentStep * SI::DELTA_T_SI;
-                    const float_X sinArg = precisionCast<float_X>(timeSI / TIME_PERIOD_SI * 2.0 * PI);
-                    return AMPLITUDE_SI * math::exp(-normalizedDuration * normalizedDuration) * math::sin(sinArg);
+                    auto const k = 2.0_X * M_PI / WAVE_LENGTH_SI;
+                    auto const x = totalCellIdx.x() * SI::CELL_HEIGHT_SI;
```
finnolec: Shouldn't this here be `auto const x = totalCellIdx.x() * SI::CELL_WIDTH_SI;` to make it more flexible?

finnolec (commented):
If the goal is, to make this example more usable as a base for other, more realistic simulations, it might also make sense to have the `WIDTH`, `CENTER` and `DURATION` parameters in SI units instead of PIConGPU units. I think that people usually set up their simulations in SI units. Also setting up everything in SI units allows you to change the simulation size more easily in the `cfg` files.

share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
```diff
@@ -50,25 +51,36 @@ namespace picongpu
...
+                    auto const t = currentStep * SI::DELTA_T_SI;
+                    // note: one could also add x influence to normalizedDuration under exp
+                    auto const longitudinal = math::exp(-normalizedDuration * normalizedDuration)
```
steindev, line 78: Your laser definition seems odd. With the definition of `normalizedDuration` the longitudinal profile here always starts at e^{-1}, which is a very sudden jump from 0 and usually results in field solver issues with noise.

I would suggest to ask a for a `PULSE_DURATION_SI` and set an `initialization_duration = 16*PULSE_DURATION_SI`. 
Then time can be rescaled by `(current_step*DELTA_T_SI - initialization _duration_steps_half) / PULSE_DURATION_SI`.

share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
```diff
@@ -50,25 +51,36 @@ namespace picongpu
...
-                {
-                    auto const normalizedY = (totalCellIdx[1] - CENTER_Y_CELLS) / WIDTH_Y_CELLS;
-                    return math::exp(-normalizedY * normalizedY);
+                    auto const durationStepsMiddle = DURATION_STEPS / 2;
+                    auto const normalizedDuration = (currentStep - durationStepsMiddle) / durationStepsMiddle;
+                    auto const k = 2.0_X * M_PI / WAVE_LENGTH_SI;
+                    auto const y = totalCellIdx.y() * SI::CELL_HEIGHT_SI;
+                    auto const t = currentStep * SI::DELTA_T_SI;
+                    // note: one could also add y influence to normalizedDuration under exp
+                    auto const longitudinal = math::exp(-normalizedDuration * normalizedDuration)
```
steindev, line 97: Same as above.

share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
sbastrakov: Oh that was a really stupid setup then. Thanks for pointing out!

PrometheusPi:
@sbastrakov What is the status of this pull request?

share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
sbastrakov: Done, the artifacts indeed disappeared.

sbastrakov:
Suggestions implemented, ready for re-review.

share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
```diff
@@ -33,42 +33,64 @@ namespace picongpu
...
                 static constexpr float_64 AMPLITUDE_SI = A0 * UNITCONV_A0_to_Amplitude_SI;
-                static constexpr uint32_t DURATION_STEPS = 300;
+                static constexpr float_64 DURATION_SI = 0.6e-9; // 0.6 ns
+
                 /// These need to be changed together with grid size
-                static constexpr uint32_t CENTER_X_CELLS = 80;
-                static constexpr uint32_t WIDTH_X_CELLS = 40;
-                static constexpr uint32_t CENTER_Y_CELLS = 40;
-                static constexpr uint32_t WIDTH_Y_CELLS = 20;
+                static constexpr float_64 CENTER_X_SI = 80.0e-3; // 80 mm
```
steindev: With a grid of 128 x 128 x 128, as written in the `1.cfg` and `4.cfg`, neither `CENTER_X` nor `CENTER_Y`will be in the middle of the box. The middle of the box would be 64.0e-3 with a cell size of 1.e-3 as defined in `grid.param`.

steindev:
There is a comment in grid.param that this example is based on a Taflove example. This is probably outdated.

share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
```diff
@@ -33,42 +33,64 @@ namespace picongpu
...
+                static constexpr float_64 WIDTH_X_SI = 40.0e-3; // 40 mm
```
steindev: The `WIDTH_X` and `WIDTH_Y` values are quite large. With these values, there will be a substantial field in boundaries.
A good value for the width would be box_size / 16 = 128 * CELL_SIZE_SI / 16 = 8.e-3 in order to have no significant field in the boundary.

psychocoderHPC (dismissed):

share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
```diff
@@ -33,42 +33,64 @@ namespace picongpu
...
-                static constexpr float_64 UNITCONV_A0_to_Amplitude_SI = -2.0 * PI / WAVE_LENGTH_SI
-                    * SI::ELECTRON_MASS_SI * SI::SPEED_OF_LIGHT_SI * SI::SPEED_OF_LIGHT_SI / SI::ELECTRON_CHARGE_SI;
+                static constexpr float_64 TIME_PERIOD_SI = 0.038e-9; // 0.038 ns
+                static constexpr float_64 WAVE_LENGTH_SI = TIME_PERIOD_SI * SI::SPEED_OF_LIGHT_SI;
+                static constexpr float_64 UNITCONV_A0_to_Amplitude_SI = -1.0 / WAVE_LENGTH_SI * SI::ELECTRON_MASS_SI
+                    * SI::SPEED_OF_LIGHT_SI * SI::SPEED_OF_LIGHT_SI / SI::ELECTRON_CHARGE_SI;
                 static constexpr float_64 A0 = 8.0;
                 static constexpr float_64 AMPLITUDE_SI = A0 * UNITCONV_A0_to_Amplitude_SI;
-                static constexpr uint32_t DURATION_STEPS = 300;
+                static constexpr float_64 DURATION_SI = 0.6e-9; // 0.6 ns
```
steindev: For analysis of the laser pulse it would be helpful, when the full laser fits into the domain. In this case the duration should fulfill 16 * DURATION_SI * SPEED_OF_LIGHT_SI == 128 * dx, resulting DURATION_SI =  0.026e-9.
However, this is smaller than your TIME_PERIOD_SI.
In order to have a significant number of wavelengths in the laser, I would propose to change the time period to TIME_PERIOD_SI = 0.0211e-9, which results in a sampling of 11 time steps per laser period and 6 cells per wavelength.

share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
```diff
@@ -33,42 +33,64 @@ namespace picongpu
...
+                static constexpr uint32_t CENTER_Y_CELLS = CENTER_Y_SI / SI::CELL_HEIGHT_SI;
+                static constexpr uint32_t WIDTH_Y_CELLS = WIDTH_Y_SI / SI::CELL_HEIGHT_SI;
 
-                HDINLINE float_X getTransversalXComponentSI(const floatD_X& totalCellIdx) const
+                // Get value to calculate Ey / Bz for the XMin source
+                HDINLINE float_X getXMinSourceValueSI(const floatD_X& totalCellIdx, const float_X currentStep) const
                 {
-                    auto const normalizedX = (totalCellIdx[0] - CENTER_X_CELLS) / WIDTH_X_CELLS;
-                    return math::exp(-normalizedX * normalizedX);
+                    auto const initializationDurationSI = 16.0 * DURATION_SI;
```
steindev: As you define the variable `initializationDurationSI` the second time here, you could as well just define it as a member of the class. (Since the number 16 is used here, I think a central definition is more consistent)

steindev (commented):
Basically fine. Yet I would propose to change the number of the example a bit to foster analysis of the resulting pulse. The numbers adjust the pulse properties such that it completely fits into the domain.

steindev:
One more comment: In line 238 of `incidentField.param` the comment says 'plane wave', but it should be 'gaussian pulse'.

share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
sbastrakov: Those were center of the laser, not box. Do you normally call it with some other words?

steindev:
Ohhh, and even one more thing: With the proposed changes of numbers, the laser fills the complete domain after 221 time steps. The period of simulation output should respect this. That is, you could have output every 110 steps and stop the simulation after 660 steps when the full laser has penetrated the boundary on the opposite side of initialization.

share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
finnolec: Maybe it should be `LASER_CENTER_X_SI` to clarify that it's the center of the laser, not of the simulation box

share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
sbastrakov: Did you mean `TIME_PERIOD_SI = 0.00211e-9` ? So that it's 11 times smaler than `DURATION_SI`, and in what you wrote it's roughly same.

steindev: No. I mean it in the sense of TIME_PERIOD_SI = 11 * DELTA_T_SI. It is right that DURATIONS_SI ist then close to TIME_PERIOD_SI, but the laser will extend over several DURATION_SI, such that there are a few full wavelength in the laser field.

steindev (dismissed):

sbastrakov:
@steindev there was an issue with compiling of 2d case, as i've initially used `.z()` component without guarding - it's now fixed. Please re-review and merge if all fine. I've checked this PR works fine on top of current `dev` after #4038 was just merged.

steindev (approved):

Issue #3959, opened 2022-01-12
Erroneous checkpoint restart of radiation plug-in


Anton-Le:
I have recently observed that a restart from a checkpoint appears to zero-out the amplitudes accumulated in the radiation simulations (or the restore is incorrect for the plug-in).

Specifically:

Over the weekend I have run a simulation (electron bunch impacting a uniform background) using `59e9b53605f9a5c1bf271eeb055bc74370a99052` with radiation module enabled, with checkpoint restarts at steps 36k, 48k, 60k.
In the following figure you see, on the left, the squares of the amplitude vectors.
![CumulativeIntensity_detectorNr_45](https://user-images.githubusercontent.com/50339174/149118110-bb92232b-f778-4040-bb51-5296e0fb85f7.jpg)
At the given time steps one observe clear "cuts" in the spectrum, something I did _not_ expect since evaluation of older data obtained using the same evaluation routines did not display such cuts.

If one integrates over the frequency the following, sawtooth-like, curve is obtained:
![IntegratedIntensityVsTimeStep_detectorNr_45](https://user-images.githubusercontent.com/50339174/149119682-c9a68b03-ea6b-4d1d-8a4e-8ed40df54efe.png)

I'm guessing that the load of checkpointed radiation data is erroneous.

PrometheusPi:
@Anton-Le Which version of PIConGPU are you using and what did you adjust in the code? 
A restart bug could have been introduced with the switch to the openPMD-api briefly before the 0.6.0 release. 
Could you also please past your `stdout` and `stderr` into this issue. The radiation plugin is quite verbose regarding restarts and tells you what it found or did not find as a restart file. Could you also list all files in your `simOutput/checkpoints/`directory.

Anton-Le:
The PIC version is the commit noted in the opening post - the 0.6.0 release as it is in `master` (https://github.com/ComputationalRadiationPhysics/picongpu/commit/59e9b53605f9a5c1bf271eeb055bc74370a99052 )

No further changes to the code were made and, since I did not change the verbosity level of the radiation plug-in from its default settings there is nothing in `stdout`/`stderr` that looks out of the ordinary:

```
 new grid size (global|local|offset): {560,2304,552}|{280,144,276}|{0,1584,276}
PIConGPUVerbose PHYSICS(1) | used Random Number Generator: RNGProvider3XorMin seed: 42
[...]

PR #3566, opened 2021-03-24
Change rad plugin from libSplash to openPMD

PrometheusPi:
This pull request changes the use of libSplash in the radiation plugin from libSplash to openPMD-api as requested in #3357. 

Be aware that this is still a draft.

**ToDo:**
 - [x] merge #3688
 - [x] compare data attributes and layout in hdf5 files
 - [x] run time test on bunch example

PR #3961, opened 2022-01-14
fix rad restart

PrometheusPi:
This pull request fixes the issue #3959.

- [x] test that restarts work

The error was caused by the inconsistent naming of checkpoint file. In order to keep compatibility with the old python modules that were based on sequential libSplash output, files are still written with a `%T_0_0_0.h5` suffix. But the restart routine was looking for a more modern style `%T.h5`. The error was most likely introduced while refactoring the pull request #3566. 

Additionally, the radiation plugin defaults to not fail when not finding a restart file but to assume zeros. Since the warning is written to the PIConGPU IO log (by default not active)  instead to the radiation standard log (by default active) the warning was not prompted. 

Thus this pull request:
 - fixes file naming
 - switches log level 
 
 **For all those effected (e.g. @Anton-Le):**
 You can restart by changing the naming from `*%T_0_0_0.h5` to `*%T.h5`.
  
For data analysis, the loading of the restart files from PIConGPU is not needed. You just need to load them manually and add them accordingly. (Please be aware that any non-complex output like totalRad (or lastRad) are still wrong - but with using the openPMD output, there should be no problem if your simulation did not restart from the radiation checkpoint. - That's why PIConGPU does  not fail when it does not find a radiation restart file.)

### DIFF
--- a/include/picongpu/plugins/radiation/Radiation.hpp
+++ b/include/picongpu/plugins/radiation/Radiation.hpp
@@ -875,7 +875,7 @@ namespace picongpu
                 {
                     std::ostringstream filename;
                     /* add to standard file ending */
-                    filename << name << timeStep << ".h5";
+                    filename << name << timeStep << "_0_0_0.h5";
 
                     /* check if restart file exists */
                     if(!boost::filesystem::exists(filename.str()))
--- a/include/picongpu/plugins/radiation/Radiation.hpp
+++ b/include/picongpu/plugins/radiation/Radiation.hpp
@@ -880,7 +880,7 @@ namespace picongpu
                     /* check if restart file exists */
                     if(!boost::filesystem::exists(filename.str()))
                     {
-                        log<picLog::INPUT_OUTPUT>(
+                        log<radLog::SIMULATION_STATE>(
                             "Radiation (%1%): restart file not found (%2%) - start with zero values")
                             % speciesName % filename.str();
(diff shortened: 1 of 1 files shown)


Anton-Le:
Great.

I'll try it out as soon as the system permits!

sbastrakov (changes requested):
There is a problem with clang formatting, as pointed out by CI. Please format using [these instructions](https://picongpu.readthedocs.io/en/latest/dev/styleguide.html)

PrometheusPi:
@sbastrakov Thanks for pointing out the CI issue - I did see it since I was occupied with testing this PR on taurus. I will fix it asap.

PrometheusPi:
I tested the restart on taurus and it works. 
grafik

What I do not understand is, why the energy value between the previous test (on hemera) and this on (on taurus) differ that much. Previously it was 100fJ here it's 10^-30J. Perhaps I made a error in my previous analysis - which I can not validate as long as hemera is down. 

However, the restart works as expected. 

Also the log output:
```
PIConGPUVerboseRadiation SIMULATION_STATE(2) | Radiation (e): read radiation data from openPMD
```
 is in `stdout` now.

PrometheusPi:
Could be merged from my side.

sbastrakov (approved):

psychocoderHPC (approved):

PR #3964, opened 2022-01-19
Add crusher profile, update spock profile, update openPMD-api doc

steindev:
Adding a profile for crusher@OLCF.

While setting up PIConGPU for crusher@OLCF, I also saw that the modules on spock@OLCF changed and updated accordingly.
This is a separate commit.

Further, with changing to C++17 in dev we forgot to update our openPMD-api dependency section in the manual.
This is fixed in another separate commit, too.
However, it is part of this PR as the openPMD-api installation procedure in the docs is referred to in both profiles added/updated with this PR.

PIConGPU KHI example compiles and runs with openPMD output on crusher with this profile.
It compiles and runs without openPMD output on spock with this profile.
While compiling with openPMD output works on spock, running produces an `hipErrorOutOfMemory` when using ADIOS2 and HDF5 directly at start-up.
Will need to further investigate. (Since I am not allows to use more than 28GiB per MI250X GCD on crusher and an MI100 on spock has 32GiB, I know that the setup itself is not too large, so something else must be going on.)
However, this is independent of this PR in my opinion.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -259,7 +259,7 @@ c-blosc
 
 openPMD API
 """""""""""
-- 0.12.0+ (bare minimum) / 0.13.0+ (for streaming IO)
+- 0.14.3+
 - *Spack*: ``spack install openpmd-api``
 - For usage in PIConGPU, the openPMD API must have been built either with support for ADIOS2 or HDF5 (or both).
   When building the openPMD API from source (described below), these dependencies must be built and installed first.
@@ -281,7 +281,7 @@ openPMD API
 - environment:* (assumes install from source in ``$HOME/lib/openPMD-api``)
 
   - ``export CMAKE_PREFIX_PATH="$HOME/lib/openPMD-api:$CMAKE_PREFIX_PATH"``
-  - ``export LD_LIBRARY_PATH="$HOME/lib/openPMD-api/lib:$LD_LIBRARY_PATH"``
+  - ``export LD_LIBRARY_PATH="$HOME/lib/openPMD-api/lib64:$HOME/lib/openPMD-api/lib:$LD_LIBRARY_PATH"``
 - If PIConGPU is built with openPMD output enabled, the JSON library
   nlohmann_json will automatically be used, found in the ``thirdParty/``
   directory.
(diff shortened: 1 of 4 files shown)


sbastrakov:
Regarding openPMD API, does the old version not work with C++17?

steindev:
> Regarding openPMD API, does the old version not work with C++17?

No, I tried to use 0.13.4 as it is provided as a module, but received an error that at least 0.14.3 is required.
So I believe this is requested in CMake, but I did not look it up.

sbastrakov:
Ah true, so our cmake requests what it should, just docs were not updated. Thanks for fixing

PrometheusPi:
@steindev The CI failed - seems to be caused by the CI - I will restart the CI.

PrometheusPi (dismissed):
lgtm but I did not test it.

INSTALL.rst
```diff
@@ -281,7 +281,7 @@ openPMD API
 - environment:* (assumes install from source in ``$HOME/lib/openPMD-api``)
 
   - ``export CMAKE_PREFIX_PATH="$HOME/lib/openPMD-api:$CMAKE_PREFIX_PATH"``
-  - ``export LD_LIBRARY_PATH="$HOME/lib/openPMD-api/lib:$LD_LIBRARY_PATH"``
+  - ``export LD_LIBRARY_PATH="$HOME/lib/openPMD-api/lib64:$LD_LIBRARY_PATH"``
```
franzpoeschel: It is platform-specific whether the shared library is installed under `lib` or `lib64`. 
```suggestion
  - ``export LD_LIBRARY_PATH="$HOME/lib/openPMD-api/lib64:$HOME/lib/openPMD-api/lib:$LD_LIBRARY_PATH"``
```

franzpoeschel:
> Regarding openPMD API, does the old version not work with C++17?

We fixed an ABI incompatibility with openPMD-api 0.14.3 https://github.com/openPMD/openPMD-api/pull/1121

INSTALL.rst
steindev: Great suggestion! Will include.

PrometheusPi (approved):

PR #3967, opened 2022-01-25
Expand documentation on running interactively

steindev:
Adds a profile for system 'bash' and adds a paragraph in docs on how to run interactively using TBG.
Hopefully this helps new users to follow our introduction on their local workstation or laptop and decreases the effort to 'just  try' PIConGPU.

To be in line with this thought, the value of `PIC_BACKEND="omp2b:native"` in the profile is chosen on purpose to allow compilation and running on standard hardware.
I am not sure whether the 'native' in this definition is actually required or may not be required but rather a 'fail-safe' setting.
Please comment and I will change if necessary.

### DIFF
--- a/docs/source/install/profile.rst
+++ b/docs/source/install/profile.rst
@@ -7,16 +7,39 @@
 picongpu.profile
 ================
 
-.. sectionauthor:: Axel Huebl
+.. sectionauthor:: Axel Huebl, Klaus Steiniger
 
-Use a ``picongpu.profile`` file to set up your software environment without colliding with other software.
-Ideally, store that file directly in your ``$HOME/`` and source it after connecting to the machine:
+We recommend to use a ``picongpu.profile`` file, located directly in your ``$HOME/`` directory,
+to set up the environment within which PIConGPU will run by conviently performing
 
 .. code-block:: bash
 
    source $HOME/picongpu.profile
 
-We listed some example ``picongpu.profile`` files below which can be used to set up PIConGPU's dependencies on various HPC systems.
+on the command line after logging in to a system.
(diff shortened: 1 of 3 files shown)


docs/source/install/profile.rst
```diff
@@ -7,16 +7,39 @@
...
 
-We listed some example ``picongpu.profile`` files below which can be used to set up PIConGPU's dependencies on various HPC systems.
+on the command line after logging in to a system.
+PIConGPU is shipped with a number of ready-to-use profiles for different systems which are located in
+``etc/picongpu/<cluster>-<institute>/`` within PIConGPU's main folder.
+Have a look into this directory in order to see for which HPC systems profiles are already available.
+If you are working on one of these systems, just copy the respective ``*_picongpu.profile.example``
+from within this directory into your ``$HOME`` and make the necessary changes, such as e-mail address
+or PIConGPU source code location defined by ``$PICSRC``.
+If there is no profile for the HPC system you are wroking on available, feel free to create one and
```
PrometheusPi:
```suggestion
If no profile is available for the HPC system you are working on, feel free to create one and
```

PrometheusPi (changes requested):
just a minor typo that needs to be fixed

pordyna:
Maybe we should also mention that our `.profile` files usually have the `getNode` and `getDevice` functions. So that  on hemera, it  is possible to use the standard profiles to run interactively on any of the gpu queues with the combination of `getDevice` and `-s bash` in `tbg`.

steindev:
Thanks @PrometheusPi and @pordyna. I updated the text.

PrometheusPi (approved):

sbastrakov (approved):

PR #3927, opened 2021-11-29
PulseFromSpectrum laserProfile

nils-pr:
# PulseFromSpectrum laserProfile

New laser profile for PIConGPU.

In this laser profile the laser pulse is defined in the frequency domain, with:
- a Gaussian spectrum
- phase shift for GDD/TOD
- in addition the spectrum and phase is altered to get the spatial shape of a Gaussian beam

Time dependent E-Field is then calculated via Fourier transformation.
Thus a time dependent gaussian beam, with a temporal profile defined by spectrum and freely chosen GDD/TOD,
is described.

## Issue:
- [ ] Weird compile error:
https://github.com/nils-pr/picongpu/blob/da3aaf722762bbcba235ecd397ac657b1bda6cda/include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp#L92-L98

No problem with compiling when declaring a "new" `y_R`.
Same compile error with the latest PIConGPU version (error: `"...  ::Y:R"undefined in device code`),

Is this a series Issue? I couldn't figure it out why cpp does that.

PR #3968, opened 2022-01-26
Add new laser profile: PulseFromSpectrum by @nils-pr

steindev:
Initializes laser pulse with gaussian transverse envelope and longitudinal envelope calculated from complex electric field given in frequency space.

This PR addresses reviewer comments from #3927 which will be closed as the original author @nils-pr is currently unable to provide the requested changes.

From the original PR #3927:
```
New laser profile for PIConGPU.

In this laser profile the laser pulse is defined in the frequency domain, with:

    a Gaussian spectrum
    phase shift for GDD/TOD
    in addition the spectrum and phase is altered to get the spatial shape of a Gaussian beam

Time dependent E-Field is then calculated via Fourier transformation.
Thus a time dependent gaussian beam, with a temporal profile defined by spectrum and freely chosen GDD/TOD,
is described.
```

I added the `CI:no-compile` flag since this laser is not part of any setup we have in the compile suite.

### DIFF
--- a/include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp
+++ b/include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp
@@ -1,0 +1,330 @@
+/* Copyright 2013-2021 Axel Huebl, Heiko Burau, Anton Helm, Rene Widera,
+ *                     Richard Pausch, Alexander Debus
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/simulation_defines.hpp"
+
+#include "picongpu/fields/laserProfiles/BaseFunctor.hpp"
+
+#include <pmacc/dataManagement/DataConnector.hpp>
+#include <pmacc/mappings/simulation/SubGrid.hpp>
+
+
+namespace picongpu
+{
+    namespace fields
+    {
+        namespace laserProfiles
+        {
+            namespace pulseFromSpectrum
+            {
+                template<typename T_Params>
+                struct Unitless : public T_Params
+                {
+                    using Params = T_Params;
+
+                    static constexpr float_X WAVE_LENGTH
+                        = float_X(Params::WAVE_LENGTH_SI / UNIT_LENGTH); // unit: meter
+                    static constexpr float_X v0
+                        = (::picongpu::SI::SPEED_OF_LIGHT_SI / Params::WAVE_LENGTH_SI) * UNIT_TIME; // unit: seconds^-1
+                    static constexpr float_X PULSE_DURATION
+                        = float_X(Params::PULSE_DURATION_SI / UNIT_TIME); // unit: seconds (1 sigma)
+                    static constexpr float_X AMPLITUDE
+                        = float_X(Params::AMPLITUDE_SI / UNIT_EFIELD); // unit: volt / meter
+                    static constexpr float_X PULSE_INIT = float_X(Params::PULSE_INIT); // unit: none
+                    static constexpr float_X GDD
+                        = float_X(Params::GDD_SI / (UNIT_TIME * UNIT_TIME)); // unit: seconds^2
+                    static constexpr float_X TOD
+                        = float_X(Params::TOD_SI / (UNIT_TIME * UNIT_TIME * UNIT_TIME)); // unit: seconds^3
+                    static constexpr float_X W0 = float_X(Params::W0_SI / UNIT_LENGTH); // unit: meter
+                    static constexpr float_X FOCUS_POS = float_X(Params::FOCUS_POS_SI / UNIT_LENGTH); // unit: meter
+                    static constexpr float_X INIT_TIME = float_X(
+                        (Params::PULSE_INIT * Params::PULSE_DURATION_SI)
+                        / UNIT_TIME); // unit: seconds (full inizialisation duration)
+
+                    /* Rayleigh-Length in y-direction
+                     */
+                    static constexpr float_X Y_R = float_X(PI * W0 * W0 / WAVE_LENGTH); // unit: meter
+                };
+            } // namespace pulseFromSpectrum
+
+            namespace acc
+            {
+                template<typename T_Unitless>
+                struct PulseFromSpectrum
+                    : public T_Unitless
+                    , public acc::BaseFunctor<T_Unitless::initPlaneY>
+                {
+                    using Unitless = T_Unitless;
+                    using BaseFunctor = acc::BaseFunctor<T_Unitless::initPlaneY>;
+
+                    float_X m_currentStep;
+
+                    /** gauss spectrum
+                     * v0 is the central frequency
+                     * PULSE_DURATION is sigma of std. gauss for intensity (E^2) in time domain
+                     * @param v frequency
+                     */
+                    HDINLINE float_X gauss_spectrum(float_X const v) const
+                    {
+                        // Norm is choosen so that the maximum after fourier transformation is unity.
+                        float_X const norm = math::sqrt(float_X(PI)) * Unitless::PULSE_DURATION;
+                        float_X const exponent = 2._X * Unitless::PULSE_DURATION * float_X(PI) * (v - Unitless::v0);
+                        return norm * math::exp(-exponent * exponent);
+                    }
+
+                    /** transversal spectrum at the init plane
+                     * to implement a laser-pulse with a transverse profile, the spectrum needs to depend on the
+                     * distance to the beam axis: spectrum(v) --> spectrum(v, r) (the transversal profile is radial
+                     * symmetric)
+                     * @param v frequency
+                     * @param r2 distance to beam axis to the power of 2
+                     * @param y_to_focus y-coordinate relative to focus-position
+                     */
+                    HDINLINE float_X
+                    transversal_spectrum(float_X const v, float_X const r2, float_X const y_to_focus) const
+                    {
+                        float_X const y_rel = y_to_focus / Unitless::Y_R;
+                        float_X const waist = Unitless::W0 * math::sqrt(1._X + y_rel * y_rel);
+                        float_X const transversal_spectrum = gauss_spectrum(v) * math::exp(-r2 / (waist * waist));
+                        return transversal_spectrum;
+                    }
+
+                    /** Spectral phase phi of the pulse with complex electric field
+                     * E = E_amp * e^[-I phi].
+                     * Depends on frequency and radius to beam axis at the init plane.
+                     * The phase is altered for two purposes:
+                     * 1. to be able to choose freely the GDD/TOD of the laser-pulse
+                     * 2. to implement the transversal profile (thus the phase depends on the distance to the beam
+                     * axis)
+                     * @param v frequency
+                     * @param r2 distance to beam axis to the power of 2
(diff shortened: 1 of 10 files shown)


PrometheusPi:
@steindev Is this ready for review or still WIP?

steindev:
Ready for review!

steindev:
@PrometheusPi would you please review soon? @nils-pr wants to provide a circularly polarized version on top of this.

include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp
```diff
@@ -0,0 +1,335 @@
...
+                     */
+                    HDINLINE float_X gauss_spectrum(float_X v)
+                    {
+                        // Norm is choosen so that the maximum after fourier transformation is unity.
+                        float_X norm = math::sqrt(float_X(PI)) * Unitless::PULSE_DURATION;
+                        return norm
+                            * math::exp(
+                                   -1.0
+                                   * math::pow(
+                                       2._X * Unitless::PULSE_DURATION * float_X(PI) * (v - Unitless::v0),
```
sbastrakov: Why is this mixed-precision? Perhaps `-math::pow(...)` ?

include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp
```diff
@@ -0,0 +1,335 @@
...
+                                       2.0));
```
sbastrakov: If this is critical, could multiply by itself instead of `pow(x, 2)`, same also below

include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp
```diff
@@ -0,0 +1,335 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/simulation_defines.hpp"
+
```
sbastrakov, line 26: `#include "picongpu/fields/laserProfiles/BaseFunctor.hpp"` is missing

include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp
```diff
@@ -0,0 +1,335 @@
...
+                        float3_X const& elong,
+                        float_X const phase,
+                        uint32_t const currentStep)
+                        : BaseFunctor(dataBoxE, superCellToLocalOriginCellOffset, offsetToTotalDomain, elong)
+                        , m_phase(phase)
+                        , m_currentStep(currentStep)
+                    {
+                    }
+
+                    /** device side manipulation for init plane (transversal)
```
sbastrakov, line 199: Is this accurate? This operator appears to do the whole thing

include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp
```diff
@@ -0,0 +1,335 @@
...
+                        DataSpace<simDim> const localCell(
+                            cellIndexInSuperCell + this->m_superCellToLocalOriginCellOffset);
+
+                        // transform coordinate system to center of x-z plane of initialization
+                        constexpr uint8_t planeNormalDir = 1u;
+                        DataSpace<simDim> offsetToCenterOfPlane(this->m_offsetToTotalDomain);
+                        offsetToCenterOfPlane[planeNormalDir] = 0; // do not shift origin of plane normal
+                        floatD_X const pos
+                            = precisionCast<float_X>(localCell + offsetToCenterOfPlane) * cellSize.shrink<simDim>();
+                        // @todo add half-cells via traits::FieldPosition< Solver::NumicalCellType, FieldE >()
```
sbastrakov, line 218: I see this is copied from the existing gaussian profile. We probably can easily account for it in all relevant profiles in a separate PR

sbastrakov (changes requested):
Some small comments above. Also please add the new profile [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/docs/source/usage/param/laser/profiles.rst) so that it appears in readthedocs

PrometheusPi (changes requested):
Please adjust also all `laser.param` of all examples that use a laser pulse.

PrometheusPi:
Please also setup at least one compile test with this laser in the examples.

PrometheusPi:
@steindev What is the status oof this pull request?

include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp
steindev: Refactored

include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp
steindev: Refactored

include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp
steindev: I don't know. This is copy-paste from the gaussian profile. Any suggestion for improvement? (I actually don't even know what you mean by 'whole thing')

include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp
steindev: Hmm not sure how useful this would be, for the moment it is ok I think. I also think that we do not want to add a whole bunch of new laser profiles in the future, and if so, then preferably use incidentField.

steindev:
> Please also setup at least one compile test with this laser in the examples.

As we just found out, we don't have compile time test for lasers yet :sweat_smile:

steindev:
Addressed reviewer comments! Ready for re-review!

include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp
sbastrakov: Yes, it's just outdated in all profiles. The whole thing I meant the whole laser calculation and application

PrometheusPi:
Thanks @steindev 
I will have a look as soon as possible.

steindev:
@PrometheusPi @sbastrakov knock, knock.

sbastrakov (dismissed):

PrometheusPi:
@steindev I will have a look at this pull request today.

steindev:
**EDIT:** Forget about this message. See next message.

@PrometheusPi I encountered a couple of errors in the implementation and fixed these now (latest force-push). A comparison between a normal gaussian beam initialized via PulseFromSpectrum and one initialized via our standard GaussianBeam yields identical results now. Result is attached.
[laser-from-spectrum_validaton.xcf.zip](https://github.com/ComputationalRadiationPhysics/picongpu/files/8421870/laser-from-spectrum_validaton.xcf.zip)

Please re-review.

steindev:
Ohh :cursing_face:, in my test I ran the standard gauss laser again instead of PulseFromSpectrum. The latter still does not work. Will debug today :disappointed:

steindev:
Finally! The errors I introduced and errors in the original implementation are now fixed. See attached gimp file for comparison between our standard gauss laser in PIConGPU and PulseFromSpectrum without dispersion. Toggle visibility of the layers: What is in the top layer needs to be applied to the bottom layer(=gauss reference) in order to obtain the middle layer(=PulseFromSpectrum) => Excellent agreement!
[laser-from-spectrum_implementation-differences.xcf.zip](https://github.com/ComputationalRadiationPhysics/picongpu/files/8476680/laser-from-spectrum_implementation-differences.xcf.zip)

Thanks to @sbastrakov for the help!

@PrometheusPi please re-review and merge if there are no more comments.

sbastrakov:
Is the approach implemented in this PR somehow similar to what SMILEI does https://smileipic.github.io/Smilei/laser_offset.html#laser-propagation-preprocessing ?

steindev:
> Is the approach implemented in this PR somehow similar to what SMILEI does https://smileipic.github.io/Smilei/laser_offset.html#laser-propagation-preprocessing ?

Well, there is definitely a similarity in the sense that we have the spectrum over ω and calculate the field in time domain from this spectrum. However, there are two notable differences:
(1) We don't do the propagation from an arbitrary point to the initialization plane. Yet the "propagation" they do seems rather useless to me as they just shift the phase a little instead of performing a real propagation using the Kirchhoff-Fresnel diffraction integral. Their approach is only valid in the very near surrounding of the position where the field is known and I actually doubt that it results in a real difference on the final field in the simulation. But I might overlook something here. Employing the Kirchhoff-Fresnel diffraction integral would allow propagation by (almost) arbitrarily long distances. "Almost" refers to the fact that the propagation distance must be at least a few wavelengths.
(2) We already assume a spectrum which can be evaluated via a function call (either as a given analytic function such as a Gaussian or an interpolating polynomial of a more complex spectral shape), while they generate the spectrum by sampling and Fourier transforming a given temporal field evolution. In essence this is equal, as their spectrum obtained through Fourier transformation can be fitted by a function and this function is what we would use as spectrum. In principle, they save the user from performing this Fourier transform and fit by himself and their invest is additional memory to save the spectrum. Then again, we can measure the frequency spectrum directly in the lab (which is done to reconstruct the temporal laser pulse shape in a single shot) and fit it and put it in the simulation.

steindev:
ping @PrometheusPi. Merge! Merge! Merge! (It is tested and nothing to be afraid of!) :sunglasses:

sbastrakov:
@steindev thanks for explanation.  With my newly obtained knowledge of incident field at angle I may now also understand the SMILEI approach better than at time of sharing that link. Easier to discuss offline

psychocoderHPC:
@PrometheusPi please check if your points are addressed!

include/picongpu/param/laser.param
```diff
@@ -604,6 +606,120 @@ namespace picongpu
...
+                 *   - absorber cells negative Y < initPlaneY < cells in y direction of the top gpu
+                 */
+                static constexpr uint32_t initPlaneY = 0;
+
+                /** Available polarisation types
+                 */
+                enum PolarisationType
+                {
+                    LINEAR_X = 1u,
+                    LINEAR_Z = 2u,
```
PrometheusPi, line 716: why is there no circular polarized option?

include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp
```diff
@@ -0,0 +1,330 @@
...
+
+                    // gaussian beam waist in the nearfield: w_y(y=0) == W0
+                    float_64 const y_rel = focusPos / Unitless::Y_R;
+                    float_64 const waist = Unitless::W0 * math::sqrt(1.0 + y_rel * y_rel);
+
+                    float_64 envelope = float_64(Unitless::AMPLITUDE);
+                    if(simDim == DIM2)
+                        envelope *= math::sqrt(float_64(Unitless::W0) / waist);
+                    else if(simDim == DIM3)
+                        envelope *= float_64(Unitless::W0) / waist;
```
PrometheusPi, line 283: Do we do this for all lasers? 
What is the effect of this multiplication?

PrometheusPi (changes requested):
I ran the default LWFA setup with the following changes:
 - no particles
 - X polarized laser
 and compared it to the new spectral laser using the same parameters. Despite using the same a_0, w_0, tau, and pulse init - the laser pulse had a different position after sliding. Example plots for time step 4000 are given below. 
 
 Field energy looks fine - Gauss and (Gauss-like) spectral laser agree (after initialization).
 
 My guess is, that there is something different in how `PULSE_INIT` is defined or works. 
 @steindev If you say, there should be no perfect agreement between `GaussianBeam` and `PulseFromSpectrum` if GDD and TOD  are zero - I am fine with merging.

PrometheusPi:
`GaussianBeam`
spectral


`PulseFromSpectrum`
grafik

PrometheusPi:
I see that my resizing of the image caused the shift in init position - thus everything is fine!

PrometheusPi (approved):
No shift in init duration, energy etc. agrees.

include/picongpu/param/laser.param
steindev: @nils-pr wanted to add it once this PR is merged. But this is quite some time ago. Not sure if he still wants to do it.

include/picongpu/fields/laserProfiles/PulseFromSpectrum.hpp
steindev: Yes, it is required as the laser amplitude scales different with distance from the focus for two or three dimensional propagation.

Issue #3960, opened 2022-01-12
openPMD density input via plugin during runtime


n01r:
Currently, it is already possible to provide openPMD-flavored files for species `densityProfile` input within `density.param`.

It would improve usability and setup time majorly if specifying the path for the files was done via a plugin (via the `openPMD` plugin?) during runtime.

sbastrakov:
I suggest the following change. One still defines a type for density-from-file in `density.param`, so that it acts like all other densities interface-wise in functors like `CreateDensity`. But this struct only has one member - a string that is a prefix for command line parameters. And all other current members become runtime as `--[plugin].[prefix].[name]`, following our standard naming for command-line parameters. Note that we still need these prefixes as one could have multiple densities from a file.

PR #3972, opened 2022-02-02
Add a runtime option to set particle density file

sbastrakov:
Implements suggestion #3960. After starting the implementation i realized [my original idea](https://github.com/ComputationalRadiationPhysics/picongpu/issues/3960#issuecomment-1011018768) does not fit very well with the existing code.

So it is adjusted as follows. The `FromOpenPMDParam` struct in `density.param` is exactly the same as before. But now one could set `filename` to an empty string `""` which is treated as a special value. If such a density is applied for e.g. electrons, the path can be set as PIConGPU command-line option `--e_runtimeDensityFile`. Note that the density parameter struct itself does not know for which species it will be applied, but a user does know. The other values are still taken from the struct, however i imagine the same issue was with the file name.

I also added a log output at `PHYSICS` level that outputs that density for species `x` is loaded from file `y`, hopefully that also helps.

### DIFF
--- a/include/picongpu/simulation/stage/RuntimeDensityFile.hpp
+++ b/include/picongpu/simulation/stage/RuntimeDensityFile.hpp
@@ -1,0 +1,124 @@
+/* Copyright 2022 Sergei Bastrakov
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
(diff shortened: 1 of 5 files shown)


include/picongpu/simulation/stage/RuntimeDensityFile.hpp
```diff
@@ -0,0 +1,124 @@
...
+                     *
+                     * Done as operator() to simplify invoking with pmacc::meta::ForEach
+                     */
+                    void operator()(po::options_description& desc)
+                    {
+                        // Density from openPMD is conditionally enabled, so use same condition
+#if(ENABLE_OPENPMD == 1)
+                        auto* filename = &densityProfiles::RuntimeDensityFile<T_Species>::get();
+                        desc.add_options()(
+                            (prefix + "_runtimeDensityFile").c_str(),
```
psychocoderHPC, line 70: Is this option always visible even if the user is not using the density from OpenPMD functor?

I am asking because if this option is always visible it could be the thinks it is always possible to use this command line parameter and then wondering why no density is created.

sbastrakov: It is always visible. There is explanation in help and docs. I also made the option name longer e.g. not `--e_densityFile`

sbastrakov: I generally agree with the concern, but didn't see a reasonable way to do it otherwise. As usual, coupling between our compile- and runtime stuff is awkward

psychocoderHPC (approved):

PR #3973, opened 2022-02-04
crusher hipcc profile

psychocoderHPC:
- rename existing profile in `batch_craycc_picongpu.profile.example`
  - add warning which recomended to use hipcc instead of CRAY CC
- add hipcc profile `batch_hippcc.profile.example`

### DIFF
--- a/docs/source/install/profile.rst
+++ b/docs/source/install/profile.rst
@@ -40,6 +40,27 @@ and providing default parameters for :ref:`TBG <usage-tbg>`.
 .. literalinclude:: profiles/bash/bash_picongpu.profile.example
    :language: bash
 
+Crusher (ORNL)
+--------------
+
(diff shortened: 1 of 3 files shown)


sbastrakov (commented):
Neither of the Crusher profiles are now cited at readthedocs, could you add those to [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/docs/source/install/profile.rst)?

psychocoderHPC:
> Neither of the Crusher profiles are now cited at readthedocs, could you add those to [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/docs/source/install/profile.rst)?

Thanks, I added both profiles to the documentation

PrometheusPi (approved):
lgtm

PrometheusPi:
There was still an error in this pull request - please see @steindev fix in #3976 for detail.

PR #3720, opened 2021-08-12
Read IO plugin configuration (openPMD plugin) from reading applications (TOML)

franzpoeschel:
Close #3717, see there for a detailed description

TODO:

- [x] Add thirdparty library and integrate it with the build system
- [x] Implement new logic in openPMD plugin
- [x] Actually read config from TOML files (currently, a little example is hardcoded)
- [x] Check that `selfRegister==true` is handled correctly (checkpoint writing)
    keep the `.toml` option out of `--checkpoint.openPMD` somehow?
- [x] Discuss command line parameter schema (may drop the `--openPMD.period dynamic` thing, it's easy to implement for now) Update: `--openPMD.period` and `--openPMD.toml` are now mutually exclusive. If using TOML, then the parameter must be the only parameter.
- [x] Logging
- [x] parallel reading
- [x] Timeout for waiting on files?
- [x] Add documentation
- [x] Merge #3906 first
- [x] I'd appreciate if someone could test that your multiplugin workflows still work
- [x] rebase after #3919 to avoid documentation merge conflicts

PR #3978, opened 2022-02-09
Fix checkpointing regression introduced by TOML PR

### LINKED ISSUE #3977: Checkpoint plugin not processing options correctly?
As noted during the crusher hackathon, the error
```bash
Unhandled exception of type 'St13runtime_error' with message 'There is no default value defined for the option file and idx is out of range', terminating
```
is thrown when using the checkpoint plugin.

cc'ing @franzpoeschel 

Using picongpu-dev as of today and TBG config attached
[tbg_checkpoint.md](https://github.com/ComputationalRadiationPhysics/picongpu/files/8036197/tbg_checkpoint.md)

franzpoeschel:
...... nothing to be seen here ........

close #3977
Bug introduced by #3720

When running a checkpoint, the openPMD plugin must not attempt to read the cmd parameter `--checkpoint.openPMD.file` because it's not there. Instead, the `--checkpoint.file` parameter is used.

### DIFF
--- a/include/picongpu/plugins/openPMD/openPMDWriter.hpp
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.hpp
@@ -358,7 +358,11 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
             std::string const prefix = "openPMD";
         };
 
-        void ThreadParams::initFromConfig(Help& help, size_t id, std::string const& dir, std::string const& file)
+        void ThreadParams::initFromConfig(
+            Help& help,
+            size_t id,
+            std::string const& dir,
+            std::optional<std::string> file)
         {
             std::string strategyString;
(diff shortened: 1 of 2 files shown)


include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -373,23 +377,40 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                 }
             case ConfigurationVia::CommandLine:
```
franzpoeschel, line 378: Note that we don't need to fix the `ConfigurationVia::Toml` path since checkpoints cannot be configured via TOML.

steindev:
Tested writing checkpoints with this fix and it works now. Will merge.

steindev (approved):

PR #3982, opened 2022-02-10
add missing include in isaac

PrometheusPi:
This pull request fixes an error introduced with #3943.
The include of the new method was missing. 
Furthermore, according to @psychocoderHPC, the ordering of the includes as mentioned  in the comments before, is not needed anymore.


PrometheusPi:
Just a remark from @psychocoderHPC: This bug only occurs if openPMD-api is disabled. 
(This is why it wasn't caught by the CI.)

pordyna (approved):

sbastrakov (approved):

PR #3983, opened 2022-02-11
Replace mpl by mp11 (part 2/2)

bernhardmgruber:
This is a continuation of the epic saga started in #3834. It continues by refactoring the seqence meta algorithms and replaces `boost::mpl::apply`.

- [x] Merge after #3834
- [x] Merge after #4516

### DIFF
--- a/include/picongpu/particles/startPosition/generic/Free.hpp
+++ b/include/picongpu/particles/startPosition/generic/Free.hpp
@@ -79,10 +79,7 @@ namespace picongpu
                     using Functor = T_Functor;
 
                     template<typename T_SpeciesType>
-                    struct apply
-                    {
-                        using type = Free;
-                    };
+                    using fn = Free;
 
                     /** constructor
                      *
--- a/include/picongpu/particles/traits/GetMarginPusher.hpp
+++ b/include/picongpu/particles/traits/GetMarginPusher.hpp
@@ -36,29 +36,17 @@ namespace picongpu
          * @tparam T_GetLowerMargin lower margin for pusher getter type
          * @tparam T_GetUpperMargin upper margin for pusher getter type
          */
-        template<
-            typename T_Species,
-            typename T_GetLowerMargin = GetLowerMargin<GetPusher<boost::mpl::_1>>,
-            typename T_GetUpperMargin = GetUpperMargin<GetPusher<boost::mpl::_1>>>
+        template<typename T_Species, typename T_GetLowerMargin, typename T_GetUpperMargin>
         struct GetMarginPusher
         {
-            using AddLowerMargins
-                = pmacc::math::CT::add<GetLowerMargin<GetInterpolation<boost::mpl::_1>>, T_GetLowerMargin>;
-            using LowerMargin = typename boost::mpl::apply<AddLowerMargins, T_Species>::type;
+        private:
+            using Interpolation = typename GetInterpolation<T_Species>::type;
 
-            using AddUpperMargins
-                = pmacc::math::CT::add<GetUpperMargin<GetInterpolation<boost::mpl::_1>>, T_GetUpperMargin>;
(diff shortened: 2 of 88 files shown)


bernhardmgruber:
The only MPL thing missing here is `BOOST_MPL_ASSERT_MSG` in `static_assert.hpp`. @psychocoderHPC how would you replace that?

psychocoderHPC:
I have this PR on my radar and will try to rebase and test it together with https://github.com/ComputationalRadiationPhysics/picongpu/pull/3834 soon.

include/pmacc/meta/Apply.hpp
```diff
@@ -0,0 +1,82 @@
...
+        {
+            using type = mp_at_c<mp_list<Args...>, I>;
+        };
+
+        template<template<typename...> typename E, typename... Ts, typename... Args>
+        struct ReplacePlaceholdersImpl<E<Ts...>, Args...>
+        {
+            using type = E<typename ReplacePlaceholdersImpl<Ts, Args...>::type...>;
+            // nested ::type of E to mimic mpl::apply
+            //using type = typename E<typename ReplacePlaceholdersImpl<Ts, Args...>::type...>::type;
```
psychocoderHPC: @bernhardmgruber we need to activate this because I would say that we have many places where we rely on the recursive apply.

Never the less both versions currently not compile.

psychocoderHPC (changes requested):

include/picongpu/particles/startPosition/generic/Free.hpp
```diff
@@ -79,10 +79,7 @@ namespace picongpu
                     using Functor = T_Functor;
 
                     template<typename T_SpeciesType>
-                    struct apply
-                    {
-                        using type = Free;
-                    };
+                    using fn = Free;
```
psychocoderHPC, line 82: Why is this called `fn` and not `apply` as in mpl?

include/pmacc/functor/Filtered.hpp
```diff
@@ -95,28 +95,16 @@ namespace pmacc
...
-            template<typename... T_Params>
-            struct apply
-            {
-                using type = Filtered<
-                    T_FilterOperator,
-                    typename boost::mpl::apply<T_Filter, T_Params...>::type,
-                    typename boost::mpl::apply<T_Functor, T_Params...>::type>;
-            };
+            // template<typename... T_Params>
+            // using fn = Filtered<T_FilterOperator, Apply<T_Filter, T_Params...>, Apply<T_Functor, T_Params...>>;
```
psychocoderHPC, line 107: I do not think that it is correct that this is removed.

The T_Filter and T_Functor must be updated with new signatures.

include/picongpu/particles/traits/GetMarginPusher.hpp
```diff
@@ -36,29 +36,17 @@ namespace picongpu
          * @tparam T_GetLowerMargin lower margin for pusher getter type
          * @tparam T_GetUpperMargin upper margin for pusher getter type
          */
-        template<
-            typename T_Species,
-            typename T_GetLowerMargin = GetLowerMargin<GetPusher<boost::mpl::_1>>,
```
psychocoderHPC, line 41: This looks wrong, the margins of the pusher will not be taken into account anymore.

psychocoderHPC (changes requested):
found some issues.

include/picongpu/particles/startPosition/generic/Free.hpp
bernhardmgruber: Because `fn` is mp11's terminology. See: https://www.boost.org/doc/libs/master/libs/mp11/doc/html/mp11.html:

> A quoted metafunction is a class with a public metafunction member called fn, for example ...

chillenzer:
Is this still planned? @psychocoderHPC @bernhardmgruber

PR #3987, opened 2022-02-15
remove obsolete check in tpl

PrometheusPi:
@steindev and I wondered why restarting simulation on crusher never ran but always "crashed". The reason for this error originates from a obsolete check that should only run PIConGPU if `cuda_memtest` ran successfully. 
However, since on AMD GPUs, there is no `cuda_memtest`. The check thus validates the return value of the previous command: creating the symbolic link between `output` and `stderr`, which of course fails if the link already exists. Thus PIConGPU never ran when the link already existed. 
This pull request fixes this issue by removing the check completely.

### DIFF
--- a/etc/picongpu/crusher-ornl/batch.tpl
+++ b/etc/picongpu/crusher-ornl/batch.tpl
@@ -96,8 +96,7 @@ ln -s ../stdout output
 
 # cuda_memtest is available only on CUDA hardware
 
-if [ $? -eq 0 ] ; then
-  # Run PIConGPU
-  srun -K1 !TBG_dstPath/input/bin/picongpu !TBG_author !TBG_programParams
-fi
+
+# Run PIConGPU
+srun -K1 !TBG_dstPath/input/bin/picongpu !TBG_author !TBG_programParams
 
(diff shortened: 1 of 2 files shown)


PrometheusPi:
Added spock `*.tpl` as well since it had the same error.

sbastrakov (commented):
This change seems valid, but to have been done on top of another branch, not on clean `dev`. It is maybe fine if PRs are merged in the right order, but then please add a checkbox for dependency with another PR.

steindev:
As @sbastrakov pointed out, the first commit in this PR seems to be related to ISAAC, and not to this PR. I suggest to remove this first commit while squashing the others and updating the commit comment to be more verbose, like 'remove obsolete check for successful cuda memcheck on spock and crusher'.

steindev (changes requested):
As @sbastrakov pointed out, the first commit in this PR seems to be related to ISAAC, and not to this PR. I suggest to remove this first commit while squashing the others and updating the commit comment to be more verbose, like 'remove obsolete check for successful cuda memcheck on spock and crusher'.

PrometheusPi:
Sorry guys (@sbastrakov and @steindev) - will fix this now.

psychocoderHPC (dismissed):

steindev (approved):

PR #4008, opened 2022-03-03
openPMD plugin: Set default backend depending on available options

franzpoeschel:
If only HDF5 is available, we should automatically use that as a backend.

### DIFF
--- a/include/picongpu/plugins/openPMD/openPMDWriter.hpp
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.hpp
@@ -212,10 +212,24 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
             plugins::multi::Option<std::string> fileName = {"file", "openPMD file basename"};
 
             plugins::multi::Option<std::string> fileNameExtension
-                = {"ext",
-                   "openPMD filename extension (this controls the"
-                   "backend picked by the openPMD API)",
-                   "bp"};
+                = { "ext",
+                    "openPMD filename extension (this controls the"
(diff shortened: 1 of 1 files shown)


include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -496,7 +510,7 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
 
                 HDINLINE void operator()(ThreadParams* params)
                 {
-#ifndef __CUDA_ARCH__
+#    ifndef __CUDA_ARCH__
```
franzpoeschel: clang-format is being weird with those

sbastrakov (dismissed):

include/picongpu/plugins/openPMD/openPMDWriter.hpp
sbastrakov: There is a clang-format issue in CI (still?). Could it be that you are using a different version from our CI and so this part gets treated differently.

psychocoderHPC:
@franzpoeschel  Please run clang format to fix the formation issues.

include/picongpu/plugins/openPMD/openPMDWriter.hpp
franzpoeschel: clang-format on my machine was in a weird mood yesterday, now it's doing this correctly again

franzpoeschel:
@psychocoderHPC done
seems like the CI timeouted?

steindev (approved):

PR #4013, opened 2022-03-04
Fix fieldBackground constructors to be HINLINE

steindev:
fieldBackground constructors are instantiated on the host. Using `HDINLINE` produces compiler warnings.


sbastrakov (approved):
I think it's a bit unfortunate, as a user doesn't have to know how we use this class, and from the user side it's not clear constructor is host-only. But i guess it's lesser evil than having a warning on it.

PrometheusPi (approved):

PR #4021, opened 2022-03-22
Extend comments in density.param

sbastrakov:
As pointed out in #4020, documentation of density profiles in `density.param` is lacking. This PR adds the user-facing part as a "header" of the file that's hooked up to readthedocs.
Also added demonstration of every Nth cell density, as it was missing from the file.

Adding documentation to `.def` files and fixing naming of homogenous density will be done elsewhere.

Issue #4020, opened 2022-03-21
Homogenous Profile underdocumented


s9105947:
Dear Maintainers,

the `Homogenous` profile is never explained, [the respective section](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/include/picongpu/param/density.param#L90) is as follows:

```c++
/* definition of homogenous profile */
using Homogenous = HomogenousImpl;
```

This is not clear (at least to me).
I'd like to suggest a more detailed explanation.

*Note: The improved documentation could also explicitly mention that the [traditionally correct spelling](https://www.lexico.com/en/definition/homogeneous) "homogeneous" is never used within the source.*

sbastrakov:
Hello @s9105947 ,

This instance technically follows the same scheme as other profiles (like `using Gaussian = GaussianImpl<GaussianParam>;` just above), and unlike them `HomogenousImpl` is just a struct, not a template.

I agree with lack of comments in this part, it applies to all such density profiles, and those comments should be placed in respective `.def` files i think. E.g. all of them may prompt a user's question "so this is profile, but with which density and number of macroparticles per cell?", which is answered [in another place](https://picongpu.readthedocs.io/en/latest/usage/param/particles/init.html#createdensity). I will try to better link these pieces together.

Regarding the naming, i think this one is quite safe to rename, since it's not widely used anyways so is not likely to crash a lot of existing setups

sbastrakov:
Now some description is added at start of `density.param` which is also included [here](https://picongpu.readthedocs.io/en/latest/usage/param/core.html#density-param). Doxygen comments are also added to declarations of classes/templates like `HomogenousImpl`, how easily are they hooked up depends on one's development environment.

Regarding the renaming part, I see no problem with doing it and so can do it now, unless there are objections.

s9105947:
Thank you for adding the comments.

Is this explanation clear enough? Especially in terms of "what is the resulting density", i.e. "what is the density in SI units"?
(I assume it refers to setting BASE_DENSITY for the entire simulation volume.)

Renaming is probably overkill, because 1) that's a breaking change with not too much benefit (either spelling is complicated) and 2) the better alternative would be sth like "constant" or "uniform" -- which is an even stronger break.
(The spelling is now mentioned in this issue anyways, this should be enough documentation for such a minor detail.)

sbastrakov:
[...]

PR #4024, opened 2022-03-24
Add doxygen comments to declarations of density profile types

sbastrakov:
Those were missing before.

I decided to not copy-paste full documentation already present in `density.param`, as then we would have to support two things. Instead, I made a brief description in spirit of what is being added in #4020, and also "linked" the parameter requirements from `density.param` by names of those structs. Those are still not proper concepts, but at least now it should be more readable.

This is logically a continuation of PR #4021 to solve another point raised by #4020. Technically the two PRs are independent from one another.

### DIFF
--- a/include/picongpu/particles/densityProfiles/FreeFormulaImpl.def
+++ b/include/picongpu/particles/densityProfiles/FreeFormulaImpl.def
@@ -1,4 +1,4 @@
-/* Copyright 2014-2022 Rene Widera
+/* Copyright 2014-2022 Rene Widera, Sergei Bastrakov
  *
  * This file is part of PIConGPU.
  *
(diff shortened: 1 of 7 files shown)


psychocoderHPC (approved):

psychocoderHPC:
We could link the `struct GaussianImpl` in a rst file then the documentation will be visible in read the docs.

@sbastrakov Could you please push again, for some reasons read the docs and the CI was not able to test.

psychocoderHPC:
I was able to trigger the rebuild, therefore no force push is required.

psychocoderHPC:
#4026 must be merged to fix the documentation issue

psychocoderHPC:
I do not know why the docs are built now, they should still be broken. 
I will merge this PR as it is and the documentation build fix should be mergen in #4026.

sbastrakov:
> We could link the struct GaussianImpl in a rst file then the documentation will be visible in read the docs.

Sorry, I am not sure what do you mean. Now readthedocs just get what's written at start of `density.param`, and indeed do not hyperlink it to doxygen. But neither do other .param files [here](https://picongpu.readthedocs.io/en/latest/usage/param/core.html).

PR #4028, opened 2022-03-28
openPMD plugin: throw error when neither defining period or toml

franzpoeschel:
Example: A user calls `picongpu --openPMD.ext h5 --openPMD.json '{}'`

Current behavior: openPMD plugin silently does nothing because neither period nor toml are defined

With this PR: openPMD plugin throws an error

### DIFF
--- a/include/picongpu/plugins/openPMD/openPMDWriter.hpp
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.hpp
@@ -344,7 +344,27 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                     // If using periods in some instances and TOML sources in others, then the other parameter
                     // must be specified as empty.
                     // Not this method's task to check this though.
-                    return tomlSources.size() > notifyPeriod.size() ? tomlSources.size() : notifyPeriod.size();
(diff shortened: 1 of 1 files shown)


include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -344,7 +344,27 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
...
+                            source,
+                            fileName,
+                            fileNameExtension,
+                            fileNameInfix,
+                            jsonConfig,
+                            dataPreparationStrategy};
+                        for(auto const& option : theseMustBeEmpty)
+                        {
+                            if(option.size() > 0)
+                            {
```
sbastrakov, line 360: Feels like a manual `std::any_of`, but no big deal.

sbastrakov (approved):

include/picongpu/plugins/openPMD/openPMDWriter.hpp
franzpoeschel: The openPMD writer does not include `<algorithm>` and PIConGPU already compiles long enough without it, so I did it manually

franzpoeschel: Also, with `std::any` I don't get the name of the failing entry for the error message

PR #4029, opened 2022-03-28
Fix copying of cuSTL device buffers

sbastrakov:
The implementation was trying to directly write to device addresses from a host side. It resulted in a segfault with invalid permissions.

This behavior was introduced accidentally in #3862. That PR kept the wrong of the two pre-existing code paths. As a result, code was written as if in-kernel, while it is host-only. Restore the fitting code path.

I now re-reviewed whole #3862 , there seem to be no other wrong-branch mistakes there.

This error is part of the last release. Which means e.g. particle calorimeter doesn't work on GPU, and maybe some other plugins as well.

### DIFF
--- a/include/pmacc/cuSTL/container/copier/D2DCopier.hpp
+++ b/include/pmacc/cuSTL/container/copier/D2DCopier.hpp
@@ -37,27 +37,21 @@ namespace pmacc
         {
             static constexpr int dim = T_dim;
 
-            PMACC_NO_NVCC_HDWARNING /* Handled via CUDA_ARCH */
-                template<typename Type>
-                HINLINE static void copy(
-                    Type* dest,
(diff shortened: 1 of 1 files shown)


include/pmacc/cuSTL/container/copier/D2DCopier.hpp
```diff
@@ -37,27 +37,21 @@ namespace pmacc
         {
             static constexpr int dim = T_dim;
 
-            PMACC_NO_NVCC_HDWARNING /* Handled via CUDA_ARCH */
```
sbastrakov, line 40: This is unrelated to the bug, just was leftover.

psychocoderHPC:
well spotted, thanks for fixing it!

psychocoderHPC (approved):

Issue #3957, opened 2022-01-06
 Problem with getExternalCellsTotal


pordyna:
I reused the handling of the new particle boundary offset from `particleCalorimeter` and noticed the following:

In the `ParticleCalorimeter.kernel` 
https://github.com/ComputationalRadiationPhysics/picongpu/blob/0655b4426eac1122e23fb580315f867f7253c699/include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.kernel#L112-L113
the boundaries are used in such a way that the `beginCellIdx` is the first cell of the valid range and `endCellIdx` is the first cell beyond this valid range. This intention is confirmed by this comment
https://github.com/ComputationalRadiationPhysics/picongpu/blob/0655b4426eac1122e23fb580315f867f7253c699/include/picongpu/particles/boundary/Utility.hpp#L97 

But, `getExternalCellsTotal` returns other values. For example for the `BACK` (or z-positive) boundary in a 256x256x128 simulation:
```
PhotonDetector: [direction9] beginExternalCellsTotal = 0, 0, 128
PhotonDetector: [direction9] endExternalCellsTotal = 256, 256, 128
```
so there are no cells in the valid z range. I think, it should be `0, 0, 128` and `256, 256, 128 + guarding_cells`

And for the FRONT boundary (z -neagtive):
```
PhotonDetector: [direction18] beginExternalCellsTotal = 0, 0, 0
PhotonDetector: [direction18] endExternalCellsTotal = 256, 256, 0
```
Here it should be `0, 0, 0` and `256, 256, guarding_cells`

@sbastrakov since this is your code, could you maybe have a look at this? I think with this the `ParticleCalorimeter` is not processing any outgoing particles.

sbastrakov:
Thanks for reporting. This appears incorrect indeed, testing.

sbastrakov:
This function, as all others in that file, operate with domains as defined [here](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-domain-definitions) and no guards. It is a bit messy that we combine both with- and without-guard indices as same data types and names in other parts and it gets mixed. 

I think the correct result of this function for `z-max` boundary should be
```
PhotonDetector: [direction9] beginExternalCellsTotal = -guarding_cells_x, -guarding_cells_y, 128 - boundaryOffset_z
[...]

PR #4030, opened 2022-03-28
Fix getExternalCellsTotal() to properly account for guard cells

sbastrakov:
Previous implementation was attempting to do it in a wrong way. Now it is fully complimentary to `getInternalCellsTotal()`. Solves #3957.

This bug is present in the last release, it only affects particle calorimeter plugin and caused it to skip processing of leaving particles.

[Total coordinate system](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-domain-definitions) starts at what is cell 0 in user's coordinate system. So the left-side guard of the leftmost local domains goes to negative values in this system. (We normally shift it by guard value for in-kernel index calculations, but all boundary utilities operate with the system as described).

Thus. the expected behavior given by @pordyna in #3957 is also not correct. Here is an example of the function used to work and works now, and also result of (unchanged since it was correct) `getInternalCellsTotal()` for illustration. Note that these functions operate with "single-side" borders, so only consider what's internal and external wrt the boundary in question, regardless of other boundaries.

Grid size: 192 256 24
Boundary offset: 0 0 0

### Old (incorrect) behavior

X max border
`beginExternalCellsTotal = {192,0,0} endExternalCellsTotal = {192,256,24} beginInternalCellsTotal = {-8,-8,-4} endInternalCellsTotal = {192,264,28}`
X min border
`beginExternalCellsTotal = {0,0,0} endExternalCellsTotal = {0,256,24} beginInternalCellsTotal = {0,-8,-4} endInternalCellsTotal = {200,264,28}`

Here the external ranges happened to be empty, would be non-empty but still incorrect for non-zero boundary offset.
The union of external and internal cells doesn't cover the whole domain.

### New (correct) behavior

X max border
`beginExternalCellsTotal = {192,-8,-4} endExternalCellsTotal = {200,264,28} beginInternalCellsTotal = {-8,-8,-4} endInternalCellsTotal = {192,264,28}`
X min border
`beginExternalCellsTotal = {-8,-8,-4} endExternalCellsTotal = {0,264,28} beginInternalCellsTotal = {0,-8,-4} endInternalCellsTotal = {200,264,28}`

Now the guard is properly covered, and the union of external and internal cells covers the whole domain.

### DIFF
--- a/include/picongpu/particles/boundary/Utility.hpp
+++ b/include/picongpu/particles/boundary/Utility.hpp
@@ -105,14 +105,18 @@ namespace picongpu
                 pmacc::DataSpace<simDim>* end)
             {
                 auto axis = pmacc::boundary::getAxis(exchangeType);
-                auto offsetCells = static_cast<int>(getOffsetCells(species, exchangeType));
+                auto offsetCells = getOffsetCells(species, exchangeType);
                 SubGrid<simDim> const& subGrid = Environment<simDim>::get().SubGrid();
-                *begin = subGrid.getGlobalDomain().offset;
-                *end = (*begin) + subGrid.getGlobalDomain().size;
+                // For non-axis directions, we take all cells including the guards
+                auto const mappingDescription = species.getCellDescription();
+                auto const guardCells
+                    = mappingDescription.getGuardingSuperCells() * mappingDescription.getSuperCellSize();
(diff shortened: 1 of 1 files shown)


sbastrakov:

psychocoderHPC (approved):

PR #4034, opened 2022-03-30
fix soft restart

psychocoderHPC:
The sub-grid structure of PMacc is updated from external classes.
In the case of a soft restart, the global offset must be reset to be valid for the next simulation.

This bug was found with ISAAC where we mostly use the soft restart option. If the moving window is enabled and the simulation restarted all particles in the domain got deleted.

CC-ing: @PrometheusPi


sbastrakov (approved):

PrometheusPi:
@psychocoderHPC Do you know, when this bug was introduced?

PR #4037, opened 2022-04-01
fix openPMD outputs negative particle patch offset

psychocoderHPC:
The patch offset for particles in openPMD output is negative for the top GPU (y-direction) if sliding window is activated.
In openPMD data this offset is stored as unsigned type and therefore underflow happens and stored value is very large.
A tool which is using particle patches to find domains with particles within a special range will not work correctly and so won't find all particle patches.

### DIFF
--- a/include/picongpu/plugins/openPMD/WriteSpecies.hpp
+++ b/include/picongpu/plugins/openPMD/WriteSpecies.hpp
@@ -50,6 +50,8 @@
 #include <boost/mpl/size.hpp>
 #include <boost/mpl/vector.hpp>
 
+#include <algorithm>
+
 
 namespace picongpu
 {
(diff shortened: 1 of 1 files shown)


include/picongpu/plugins/openPMD/WriteSpecies.hpp
```diff
@@ -493,7 +493,10 @@ namespace picongpu
                         ds.options = params->jsonMatcher->get(basename + "/particlePatches/extent/" + name_lookup[d]);
                         extent_x.resetDataset(ds);
 
-                        offset_x.store<index_t>(mpiRank, particleOffset[d]);
+                        // particleOffset[d] is allowed to be negativ for the first GPU
```
sbastrakov:
```suggestion
                        // particleOffset[d] is allowed to be negative for the first GPU
```

include/picongpu/plugins/openPMD/WriteSpecies.hpp
```diff
@@ -493,7 +493,10 @@ namespace picongpu
...
+                        using OffsetType = decltype(particleOffset[d]);
+                        auto patchParticleOffset = std::max(static_cast<OffsetType>(0), particleOffset[d]);
```
sbastrakov:
```suggestion
                        auto const patchParticleOffset = std::max(static_cast<OffsetType>(0), particleOffset[d]);
```

sbastrakov: Forgot to `#include <algorithm>` ?

sbastrakov (commented):
I added small suggestions above

include/picongpu/plugins/openPMD/WriteSpecies.hpp
psychocoderHPC: Yes well catched

psychocoderHPC: added

include/picongpu/plugins/openPMD/WriteSpecies.hpp
psychocoderHPC: fixed

sbastrakov (approved):

PR #4038, opened 2022-04-01
Add basic support for laser profiles in incident field

sbastrakov:
It doesn't add any new profiles compared to the current state. But prepares to make this addition more simple. Also improves naming to be more consistent with existing laser and density profiles.

~~Adding current WIP. It is mostly ready, will continue next week.~~
Ready for review. After this PR is merged, I can start transferring the existing laser profiles one by one, with mostly reusing the old parameter structures.

### DIFF
--- a/include/picongpu/fields/incidentField/Solver.hpp
+++ b/include/picongpu/fields/incidentField/Solver.hpp
@@ -479,10 +479,10 @@ namespace picongpu
                         offsetMinBorder[axis] = absorberThickness(axis, 0) + GAP_FROM_ABSORBER[axis][0];
                         offsetMaxBorder[axis] = absorberThickness(axis, 1) + GAP_FROM_ABSORBER[axis][1];
                     }
-                    hasMinProfile[0] = !std::is_same_v<XMin, profiles::None>;
-                    hasMinProfile[1] = !std::is_same_v<YMin, profiles::None>;
-                    if(simDim == 3)
-                        hasMinProfile[2] = !std::is_same_v<ZMin, profiles::None>;
+                    hasMinProfile[0] = !std::is_same_v<XMinProfile, profiles::None>;
+                    hasMinProfile[1] = !std::is_same_v<YMinProfile, profiles::None>;
+                    if constexpr(simDim == 3)
+                        hasMinProfile[2] = !std::is_same_v<ZMinProfile, profiles::None>;
                 }
 
                 /** Apply contribution of the incident B field to the E field update by one time step
@@ -494,9 +494,9 @@ namespace picongpu
                  */
                 void updateE(float_X const sourceTimeIteration)
                 {
-                    updateE<0, XMin, XMax>(sourceTimeIteration);
-                    updateE<1, YMin, YMax>(sourceTimeIteration);
-                    updateE<2, ZMin, ZMax>(sourceTimeIteration);
+                    updateE<0, XMinProfile, XMaxProfile>(sourceTimeIteration);
+                    updateE<1, YMinProfile, YMaxProfile>(sourceTimeIteration);
+                    updateE<2, ZMinProfile, ZMaxProfile>(sourceTimeIteration);
                 }
 
                 /** Apply contribution of the incident E field to the B field update by half a time step
@@ -511,9 +511,9 @@ namespace picongpu
                  */
                 void updateBHalf(float_X const sourceTimeIteration)
                 {
-                    updateBHalf<0, XMin, XMax>(sourceTimeIteration);
-                    updateBHalf<1, YMin, YMax>(sourceTimeIteration);
-                    updateBHalf<2, ZMin, ZMax>(sourceTimeIteration);
+                    updateBHalf<0, XMinProfile, XMaxProfile>(sourceTimeIteration);
+                    updateBHalf<1, YMinProfile, YMaxProfile>(sourceTimeIteration);
+                    updateBHalf<2, ZMinProfile, ZMaxProfile>(sourceTimeIteration);
                 }
 
             private:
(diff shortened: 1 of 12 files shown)


include/picongpu/fields/incidentField/profiles/Free.def
```diff
@@ -0,0 +1,78 @@
...
+             * @tparam T_axis boundary axis, 0 = x, 1 = y, 2 = z
+             * @tparam T_direction direction, 1 = positive (from the min boundary inwards), -1 = negative (from the max
+             * boundary inwards)
+             *
+             * @{
+             */
+
+            /** Get functor for incident E values */
+            template<typename T_Profile, uint32_t T_axis, int32_t T_direction>
+            struct GetFunctorIncidentE;
```
sbastrakov: Not sure if we have a standard way of doing it. Basically, this is a trait that may be specialized by a user of the `Free` profile. So i put declaration of the trait to this `.def` file and its default implementation to `../Traits.hpp`

sbastrakov:
Ready for review.

steindev:
There is a comment in `grid.param` that this example is based on a Taflove example. This is probably outdated.

include/picongpu/fields/incidentField/Solver.hpp
```diff
@@ -125,7 +126,7 @@ namespace picongpu
                     auto const exPosition = traits::FieldPosition<cellType::Yee, FieldE>{}()[0];
                     PMACC_VERIFY_MSG(
                         exPosition[0] == 0.5_X,
```
psychocoderHPC, line 128: Offline discussed: We should check all components because PIConGPU is allowed to define arbitrary grid positions for the E field and B field even if it mostly makes no sense.
@sbastrakov suggested checking that the grid type is yee grid is much better.

include/picongpu/fields/incidentField/Solver.hpp
```diff
@@ -498,10 +479,10 @@ namespace picongpu
                         offsetMinBorder[axis] = absorberThickness(axis, 0) + GAP_FROM_ABSORBER[axis][0];
                         offsetMaxBorder[axis] = absorberThickness(axis, 1) + GAP_FROM_ABSORBER[axis][1];
                     }
-                    hasMinSource[0] = !std::is_same_v<XMin, None>;
-                    hasMinSource[1] = !std::is_same_v<YMin, None>;
+                    hasMinProfile[0] = !std::is_same_v<XMin, profiles::None>;
+                    hasMinProfile[1] = !std::is_same_v<YMin, profiles::None>;
                     if(simDim == 3)
```
psychocoderHPC: since we are in C++17 now you could write 
```C++
if constexpr(simDim == 3)
```

The advantage is that this would then also work if `hasMinProfile` is a `std::array<TYPE,simDIm>`.

include/picongpu/param/incidentField.param
```diff
@@ -95,23 +114,33 @@ namespace picongpu
...
 
-            /** Source of incident E and B fields.
+            /** Make a profile with the user-provided functors defined above.
              *
-             * Each source type combines functors for incident field E and B, which must be consistent to each other.
+             * The functors for incident field E and B must be consistent to each other.
+             * The profile can be used for XMin, XMax, etc.
              */
-            using MySource = Source<FunctorIncidentE, FunctorIncidentB>;
+            using MyProfile = profiles::Free<UserFunctorIncidentE, UserFunctorIncidentB>;
```
psychocoderHPC, line 122: Offline discussed: It is somehow unclear when we need to provide the trait `GetFunctorIncidentE` for a user incident field profile and when not.

psychocoderHPC (changes requested):
We need to discuss in which cases the user needs to provide the trails for the profile with some fresh eyes tomorrow.

include/picongpu/param/incidentField.param
sbastrakov: Yes, so I will rethink this part and try to simplify it if possible. From the discussion in the office I got the feedback that this trait is very confusing.

sbastrakov: Indeed, upon thinking more I came to a conclusion this trait is only needed internally, but not to be exposed to a user. And my hypothetical scenario which caused to expose it initially is actually better implemented on a user side without it. I will make a change in this PR (it's not much).

include/picongpu/fields/incidentField/Solver.hpp
sbastrakov: I tried now and recalled why i didn't go for checking the grid type. When initially developing incident field solver, at this place there was a problem with inclusion order / cyclic dependencies and I couldn't make it compile. I believe it's possible to work around it by adding more indirection layers, but i thought it's not worth it.

I tried again now and none of our usual ways of delaying evaluation help here. So i propose to keep this place as is, since we already are checking for FDTD type of field solver just above.

sbastrakov:
Ready for re-review @psychocoderHPC

include/picongpu/fields/incidentField/profiles/Free.def
sbastrakov: No longer relevant

include/picongpu/fields/incidentField/Solver.hpp
psychocoderHPC: ok so we will stay with this test, it's ok!

include/picongpu/fields/incidentField/profiles/Free.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+namespace picongpu
+{
+    namespace fields
+    {
+        namespace incidentField
+        {
+            namespace profiles
+            {
+                template<typename T_FunctorIncidentE, typename T_FunctorIncidentB>
```
psychocoderHPC, line 35: Should we by default set `T_FunctorIncidentB` to `ZeroFunctor`?

psychocoderHPC (approved):

include/picongpu/fields/incidentField/profiles/Free.hpp
sbastrakov: No, since zero-functor is not a good default value - it will cause two-sided generation that may be unexpected and even ruin the setup. So i think that is what a user has to set explicitly. However I have an idea for a follow-up PR to introduce functor that calculates B from E using a plane-wave approximation (we'll need this functor internally anyways, so why not also offer to a user) and so use this new `BFromE<T_FunctorIncidentE>` as a default second argument.

psychocoderHPC:
Thanks for this PR, the incidents field is very important and I see this as the future of the laser generation which will soon fully replace our current laser definition.

PR #4040, opened 2022-04-05
fix function `toTimeSlice`

psychocoderHPC:
The translation from strings into time slices processed empty strings e.g. `100,,200` or fully empty slice strings `` into
periods `0::1`.
So it could accedently registering plugins to dump each time step from the begin of the simulation up to the end.


sbastrakov:
@psychocoderHPC code formatting check failed

sbastrakov (approved):

Issue #4023, opened 2022-03-23
Time Checkpoints


BrianMarre:
It would be nice to be able to specify writing checkpoints by runtime, aka write a checkpoint after every x min of real time, as for example possible in smilie.

sbastrakov:
I think we can reasonably implement it.

sbastrakov:
Unrelated, but occured to me when reading this issue. Should we make a new label for "feature request"? As it's not exactly a question

psychocoderHPC:
> Unrelated, but occured to me when reading this issue. Should we make a new label for "feature request"? As it's not exactly a question

Yes please.

sbastrakov:
@psychocoderHPC you've explained to me how to add a label, but I must have forgotten or maybe don't have the right for it. I've tried to add a label now and didn't manage.

psychocoderHPC:
> It would be nice to be able to specify writing checkpoints by runtime, aka write a checkpoint after every x min of real time, as for example possible in smilie.

I have this feature mostly ready. As soon as I have some minutes I will finish it.

There are two ways to implement it

1. Suggested by @sbastrakov: Check in the timestep loop if a time threshold is passed and triggering checkpointing.
2. Using a sleeping thread which is notifying the simulation to perform a checkpoint.

I started with 2. to keep the time step loop small and not blow it up. The thread is using the signaling feature we to trigger checkpointing.

psychocoderHPC:
Feature is implemented in #4042

PR #4042, opened 2022-04-05
allow time dependent checkpointing

psychocoderHPC:
close: #4023

Add option `--checkpoint.timePeriod` for periodically checkpointing after a user selected time (in minutes).

- [x] merge after #4040 to avoid that each step is checkpointed of ``--checkpoint.period`` is not defined.

### DIFF
--- a/include/pmacc/simulationControl/SimulationHelper.hpp
+++ b/include/pmacc/simulationControl/SimulationHelper.hpp
@@ -36,11 +36,14 @@
 
 #include <boost/filesystem.hpp>
 
+#include <chrono>
+#include <condition_variable>
 #include <csignal>
 #include <fstream>
 #include <iomanip>
 #include <iostream>
 #include <string>
+#include <thread>
 #include <vector>
 
 
@@ -77,6 +80,12 @@ namespace pmacc
 
         ~SimulationHelper() override
         {
+            // notify all concurrent threads to exit
+            exitConcurrentThreads.notify_all();
+            // wait for time triggered checkpoint thread
+            if(checkpointTimeThread.joinable())
+                checkpointTimeThread.join();
+
             tSimulation.toggleEnd();
             if(output)
             {
@@ -149,7 +158,7 @@ namespace pmacc
         virtual void dumpOneStep(uint32_t currentStep)
         {
             /* trigger checkpoint notification */
-            if(!checkpointPeriod.empty() && pluginSystem::containsStep(seqCheckpointPeriod, currentStep))
+            if(pluginSystem::containsStep(seqCheckpointPeriod, currentStep))
             {
                 /* first synchronize: if something failed, we can spare the time
                  * for the checkpoint writing */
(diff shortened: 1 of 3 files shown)


psychocoderHPC:
It is important to implement this feature over the signaling interface. The reason is that the elapsed time per node can differ if the checkpointing is directly within the simulation loop based on elapsed time being triggered it could be that each instance/process is within a different time step. This would result in a deadlock. The signaling interface is already handling this behavior and will select a common time step for the next checkpointing.

include/pmacc/simulationControl/SimulationHelper.hpp
```diff
@@ -331,7 +356,9 @@ namespace pmacc
                 ("checkpoint.restart.step", po::value<int32_t>(&restartStep),
                  "Checkpoint step to restart from")
                 ("checkpoint.period", po::value<std::string>(&checkpointPeriod),
-                 "Period for checkpoint creation")
+                 "Period for checkpoint creation [interval(s) based on steps]")
+                ("checkpoint.timePeriod", po::value<std::uint64_t>(&checkpointTimePeriod)->default_value(0),
+                 "Time periodic checkpoint creation [period in minutes]")
```
sbastrakov: I think there should be no default value for this flag. Rather, flag itself should be optional, and I think it is already. So technically perhaps it means to remove default value from here, and add a check somewhere in the internal code?

(Also, imo 0 is not a good default value for periods in general as period == 0 is not defined, so 0 is rather a special abnormal value)

include/pmacc/simulationControl/SimulationHelper.hpp
```diff
@@ -379,12 +406,19 @@ namespace pmacc
...
 
-        /* period for checkpoint creation */
+        /* period for checkpoint creation [interval(s) based on steps]*/
         std::string checkpointPeriod;
 
         /* checkpoint intervals */
         SeqOfTimeSlices seqCheckpointPeriod;
 
+        /* period for checkpoint creation [period in seconds]*/
+        std::uint64_t checkpointTimePeriod;
```
sbastrakov: Seconds or minutes?

Also, perhaps this unit may become part of variable name here, and `checkpointPeriod` may become `checkpointPeriodSteps`

sbastrakov (commented):
I have two points concerning style above. Otherwise looks good

include/pmacc/simulationControl/SimulationHelper.hpp
psychocoderHPC: done

include/pmacc/simulationControl/SimulationHelper.hpp
psychocoderHPC: I added to the variable that zero is disabling time depended checkpointing but not to the doc string because I assume the user will not try zero at all.

sbastrakov (approved):

PR #4050, opened 2022-04-08
Add utilities for calculating incident B from E using slowly varying envelope approximation

sbastrakov:
We will need them for implementing laser profiles, and they also make sense for the `Free` profile as default mode.

PR #4053, opened 2022-04-11
Add plane wave incident field profile

sbastrakov:
This is the first non-trivial preset incident field profile, others will follow in a similar fashion.

The idea is to make the user side of incident field very close to existing lasers. So that a user has a very similar `struct PlaneWaveParams` in a `.param` file and then makes a profile out of that with e.g. `using XMin = profiles::PlaneWave<PlaneWaveParam>;`. Neither parameter structures (like `PlaneWaveParams`) nor the profiles themselves (like `profiles::PlaneWave<>`) know which side they are applied to. The binding happens only by a user putting those as aliases like `XMin`, and the other 5. So that in principle one could write smth like

```cpp
using XMin = profiles::PlaneWave<PlaneWaveParam>;
using XMax= profiles::PlaneWave<PlaneWaveParam>;
...
using ZMax = profiles::PlaneWave<PlaneWaveParam>;
```
and all those lasers would point inwards from the respective boundary.

Thus, the naming has to be generalized a little bit. Polarization is not hard-set to "linear X / linearZ" as in the existing laser. Instead, the laser define an internal coordinate system and polarizations are expressed using it. Focusing shifts will also be expressed as inward-shifts in this coordinate system (not applicable to plane wave). There is no `initPlaneY` member since the incident field generation positioning is controlled externally with `GAP_FROM_ABSORBER`. I also removed the time shift related to that `initPlaneY` as it doesn't make sense now; it could be compensated with other parameters when necessary. 

- [x] depends on #4050
- [x] check that all axes and directions work
- [x] add comparison plots of `YMin` incident field vs. laser 

Here is a comparison result. Only for the sake of this comparison I derived a formula of how to adjust the incident field functor to "represent" the `initPlaneY` in laser: in incident field, `PlaneWaveParam::RAMP_INIT` should be increased by `((laserInitPlaneY + 1.0) * SI::CELL_HEIGHT_SI / SI::SPEED_OF_LIGHT_SI) / (0.5 * PULSE_LENGTH_SI)`. Note that this is only to make it match the logic of the laser, is not relevant for "normal" incident field use. I will document this difference in approach in a separate PR.

Attaching comparison plots for laser with `initPlaneY = 32` vs. incident field with adjustment above, and other parameters as attached. Laser makes a two-sided pulse (as it should due to its implementation), so for the difference plot only use cells after y=37. The difference could probably have been made smaller if the `initPlaneY` is chosen so that analytical laser is 0 there, but I think the existing setup is sufficient proof the new profile works.
[planewave_test_problem.zip](https://github.com/ComputationalRadiationPhysics/picongpu/files/8474105/planewave_test_problem.zip)

laser_vs_incident_300
laser_vs_incident_500
laser_vs_incident_700

### DIFF
--- a/include/picongpu/fields/incidentField/profiles/BaseFunctorE.hpp
+++ b/include/picongpu/fields/incidentField/profiles/BaseFunctorE.hpp
@@ -1,0 +1,96 @@
+/* Copyright 2022 Sergei Bastrakov
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/simulation_defines.hpp"
+
+#include "picongpu/fields/incidentField/Functors.hpp"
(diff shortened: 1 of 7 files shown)


psychocoderHPC (approved):

sbastrakov:

PR #4054, opened 2022-04-13
refactor openPMD plugin

psychocoderHPC:
- Remove field sizes and offsets from the helper collection struct
`ThreadParams` and calculate these values at the place where those are
needed.
- Rename some variables pointing to a buffer, window or openPMD property/size.
buffer

This PR is the base to implement slicing for fields and particles. We already support sliding window, a slide will be an arbitrary window (contiguous area within the global domain)

### DIFF
--- a/include/picongpu/plugins/openPMD/openPMDWriter.hpp
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.hpp
@@ -537,7 +537,6 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                     if(traits::IsFieldOutputOptional<T_Field>::value && !dc.hasId(T_Field::getName()))
                         return;
                     auto field = dc.get<T_Field>(T_Field::getName());
-                    params->gridLayout = field->getGridLayout();
                     bool const isDomainBound = traits::IsFieldDomainBound<T_Field>::value;
 
                     const traits::FieldPosition<fields::CellType, T_Field> fieldPos;
@@ -559,7 +558,7 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                         params,
                         GetNComponents<ValueType>::value,
                         T_Field::getName(),
-                        field->getHostDataBox().getPointer(),
+                        *field,
                         getUnit(),
                         T_Field::getUnitDimension(),
                         std::move(inCellPosition),
@@ -660,14 +659,13 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                      * solver implementation */
                     const float_X timeOffset = 0.0;
 
-                    params->gridLayout = fieldTmp->getGridLayout();
                     bool const isDomainBound = traits::IsFieldDomainBound<FieldTmp>::value;
                     /*write data to openPMD Series*/
                     openPMDWriter::template writeField<ComponentType>(
                         params,
                         components,
                         getName(),
-                        fieldTmp->getHostDataBox().getPointer(),
+                        *fieldTmp,
                         getUnit(),
                         FieldTmp::getUnitDimension<Solver>(),
                         std::move(inCellPosition),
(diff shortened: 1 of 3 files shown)


include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -703,25 +701,27 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                 ::openPMD::MeshRecordComponent mrc = mesh[::openPMD::RecordComponent::SCALAR];
                 std::string datasetName = params->openPMDSeries->meshesPath() + name;
 
-                auto fieldsSizeDims = params->fieldsSizeDims;
-                auto fieldsGlobalSizeDims = params->fieldsGlobalSizeDims;
-                auto fieldsOffsetDims = params->fieldsOffsetDims;
+                // rng states are always of the domain size therefore querry sizes from pmacc
```
sbastrakov:
```suggestion
                // rng states are always of the domain size therefore query sizes from domain information
```

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -703,25 +701,27 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
...
+                const SubGrid<simDim>& subGrid = Environment<simDim>::get().SubGrid();
+                pmacc::math::UInt64<simDim> recordLocalSizeDims = subGrid.getLocalDomain().size;
+                pmacc::math::UInt64<simDim> recordOffsetDims = subGrid.getLocalDomain().offset;
+                pmacc::math::UInt64<simDim> recordGlobalSizeDims = subGrid.getGlobalDomain().size;
```
sbastrakov, line 708: What is the logic of changing the naming to `record`? I thought it's a field on our side, and becomes record in openPMD speech

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -1234,26 +1233,27 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
...
-                DataSpace<simDim> field_no_guard = params->window.localDimensions.size;
-                DataSpace<simDim> field_guard = field_layout.getGuard() + params->localWindowToDomainOffset;
+                DataSpace<simDim> localWindowSize = params->window.localDimensions.size;
+                DataSpace<simDim> bufferOffset = bufferGridLayout.getGuard() + params->localWindowToDomainOffset;
                 std::vector<char>& fieldBuffer = params->fieldBuffer;
 
-                auto fieldsSizeDims = params->fieldsSizeDims;
-                auto fieldsGlobalSizeDims = params->fieldsGlobalSizeDims;
-                auto fieldsOffsetDims = params->fieldsOffsetDims;
+                // rng states are always of the domain size therefore querry sizes from pmacc
```
sbastrakov: This part of the code is not about `rng states`

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -1287,13 +1287,13 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
...
-                    fieldsOffsetDims = pmacc::math::UInt64<simDim>::create(0);
-                    fieldsOffsetDims[0] = globalOffsetFile;
+                    recordGlobalSizeDims = pmacc::math::UInt64<simDim>::create(1);
+                    recordGlobalSizeDims[0] = globalSize;
+                    recordOffsetDims = pmacc::math::UInt64<simDim>::create(0);
+                    recordOffsetDims[0] = globalOffsetFile;
                 }
 
-                auto const componentSize = field_no_guard.productOfComponents();
+                auto const numDataPoints = localWindowSize.productOfComponents();
```
sbastrakov, line 1295: I think this may be a bug. Shouldn't it use `recordLocalSizeDims` instead (difference in case we went into `if(!isDomainBound)` )?

sbastrakov (commented):
Please see comments and questions above

include/picongpu/plugins/openPMD/openPMDWriter.hpp
psychocoderHPC: This variable is used for openPMD and therefore I  named these to point to the fact that we will use these for openPMD records.

include/picongpu/plugins/openPMD/openPMDWriter.hpp
psychocoderHPC: well catch! yes it is wrong

psychocoderHPC: I kept `recordLocalSizeDims` because `recordLocalSizeDims` is set to the same value. Never the less I write now inside of the `if(!isDomainBound)` -> `recordLocalSizeDims = precisionCast<uint64_t>(localWindowSize);` to be clear that both variables are equal, the only difference is the precision and that `recordLocalSizeDims` should be used to configure openPMD where variables named with `window` will be used to prepare local memory for dumping.

psychocoderHPC:
updated

include/picongpu/plugins/openPMD/openPMDWriter.hpp
sbastrakov: True, at first review I didn't realize you were changing both variables accordingly.

sbastrakov (approved):

Issue #4052, opened 2022-04-11
Consider changing parameters for incident field generation surface position


sbastrakov:
Discussed offline with @PrometheusPi , we may want to set it as fixed total position, and not as offset from field absorber, in order to better interact with moving window.

sbastrakov:
Discussed offline with @psychocoderHPC . It indeed makes sense to control the positioning as offsets from domain boundaries, not from absorber. Also for avoiding cases of a user changing absorber thickness but not the gap - the offset will avoid those potential mishaps, and we check that the overall configuration is valid anyways.

sbastrakov:
Implemented with #4056 and #4120.

PR #4056, opened 2022-04-13
Change way of controlling position of the Huygens surface

sbastrakov:
Now it is set by offset from global domain boundary. Updated comments and add a warning about setting it to absorber_size + const. Along the way also refactor validity checking code. Update and slightly extend the docs.

Implements part of #4052, but for now uses global and not total positining. For that issue it's still left to figure out how to properly couple it with moving window.


psychocoderHPC (dismissed):

sbastrakov:
@psychocoderHPC there was a minor bug with `constexpr` that passed through on the dev system and failed on CI. It's fixed, please re-approve

psychocoderHPC (approved):

PR #4057, opened 2022-04-13
Update alpakaCommon.cmake

frobnitzem:
Fallback to manually checking for hip 5 when hip 4 is not found.

update by @psychocoderHPC: 
- [ ] discuss with @ComputationalRadiationPhysics/picongpu-maintainers if we should merge it before the alpaka update (could make merging issue)


psychocoderHPC:
We do not update the modules in the third party folder by hand, we always pull updates from the corresponding repository.
We are on the way to releasing a new version of alpaka, the idea was to pull the latest alpaka where this fix is already in soon, unfortionally we found bug which shifts the alpaka release from this week into the next :-(
I will keep this PR open and will discuss with the other devs if we should make an exception and merge this.

sbastrakov:
@psychocoderHPC just a reminder about this

psychocoderHPC:
I will close this PR because #4081 will introduce alpaka 0.9.0 and therefore avoids manual changes in the subtree.

PR #4064, opened 2022-04-19
Add polynom incident field profile

sbastrakov:
Its parameters mirror ones of the `Polynom` laser profile when possible.

The logic is similar to  #4053 and all explanations there apply to this PR as well. Just here we are porting the `Polynom` laser and not the `PlaneWave` one.

I also made a similar comparison on a test problem that generates at `y_cell = 32`. In this case, it was not possible to perfectly translate the laser `initPlaneY` via only parameters. However, I just hard-coded it inside the incident field for the sake of testing (and removed after). Attaching the problem files and some plots. (also see [my message below](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4064#issuecomment-1105196202))
[polynom_test_problem.zip](https://github.com/ComputationalRadiationPhysics/picongpu/files/8513253/polynom_test_problem.zip)
laser_vs_incident_300
laser_vs_incident_500
laser_vs_incident_700

### DIFF
--- a/include/picongpu/fields/incidentField/profiles/Polynom.def
+++ b/include/picongpu/fields/incidentField/profiles/Polynom.def
@@ -1,0 +1,113 @@
+/* Copyright 2013-2022 Heiko Burau, Rene Widera, Richard Pausch, Axel Huebl, Sergei Bastrakov
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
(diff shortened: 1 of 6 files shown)


sbastrakov:
~~Oh I've noticed there is a difference from the laser in transversal component, that is not easily seen on the `x=x_center, y=y_center` plot as I made. However, it's present more visibly in other cuts. I think the laser actually did one thing wrongly there which I corrected in incident field profile without noticing the difference, will discuss which implementation is right here.~~

sbastrakov:
I did additional investigation. That simulation has a quite small spot size with 256^3 grid and `W0_X = 16.7 dx`,  `W0_Z = 66.7 dz`. At the center the match between laser and incident field is quite good, but the difference is much more visible father away, attaching the plots for `x, z` at quarter of grid instead of the center. The difference between the lasers goes to as high as 5%, an order of magnitude larger than in the center.

laser_vs_incident_300
laser_vs_incident_500
laser_vs_incident_700

After investigation and discussing with @PrometheusPi , probably the reason is the combination of small spot size and the way the original laser adds only to E, while incident field uses SVEA. Both are not precise, but probably incident field is less wrong here. When going to a larger spot size with 10x increased `W0` values the match at is much better at same `x, z` at quarter box size:

laser_vs_incident_300
laser_vs_incident_500
laser_vs_incident_700

Also during the investigation I realized my understanding of laser `initPlaneY` for non-focused laser was not complete. Probably we need to clarify it more in the docs. E.g. for this laser changing `initPlaneY` actually leads to changed laser, as it should, but that was not immediately clear to me. For some other non-focused lasers that also leads to effectively longer ramping which I am not sure is intended.

psychocoderHPC (approved):

PR #4068, opened 2022-04-20
Modify incident field functors to take time in constructor instead of operator()

sbastrakov:
Adjust all profiles and example accordingly. The motivation for the change is explained in #4065.
Fixes #4065.

- [x] Based on #4064.

PR #4076, opened 2022-04-25
Add GaussianBeam incident field profile 

sbastrakov:
- [x] Based on (today's version of) #4068

Its parameters mirror ones of the `GaussianBeam` laser profile when possible.

Unlike the previously added incident field profiles, this one is for a focused laser. So the internal translation from laser profile to incident field profile needed to be slightly different, and took some time to get right. For this reason (and because it's most frequently used type of laser), I also tested it more thouroughly.

First attaching results for the following test problem with circular polarization and focus position at 100th cell in y.
[gaussian_test_problem.zip](https://github.com/ComputationalRadiationPhysics/picongpu/files/8554177/gaussian_test_problem.zip)

Here is a slice at `x, z` at the center of the area:

laser_vs_incident_300
laser_vs_incident_500
laser_vs_incident_700

And here for `x, z`  at quater of the area:
laser_vs_incident_300
laser_vs_incident_500
laser_vs_incident_700

Then I also tried to put the focus position to be very near to the generation plane, at 32 cells. The results below are for this case and linear X polarization, also match well:

laser_vs_incident_500
laser_vs_incident_700

Thanks to @PrometheusPi and @steindev for discussions.

### DIFF
--- a/include/picongpu/fields/incidentField/profiles/GaussianBeam.def
+++ b/include/picongpu/fields/incidentField/profiles/GaussianBeam.def
@@ -1,0 +1,138 @@
+/* Copyright 2013-2022 Axel Huebl, Heiko Burau, Anton Helm, Rene Widera,
+ *                     Richard Pausch, Alexander Debus, Sergei Bastrakov
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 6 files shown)


psychocoderHPC (dismissed):

sbastrakov:
@psychocoderHPC I suggest force-merging, as conflict was only in comments.

sbastrakov:
Now also thinking of that, i think we really should not put those lists in source code files, but to documentation instead. It's really long, and pointless, and causes copy-paste, i'm really not a fan of that (but we do it quite consistently, so if changing lots of places would need to be addressed)

psychocoderHPC (approved):

Issue #4072, opened 2022-04-22
duplicate attribute name


s9105947:
Dear Maintainers,

during [the work on PICMI](https://github.com/s9105947/picongpu/tree/picmi-integration) the code generation was off and triggered a uncaught configuration error.

## Description
Define an particle species attribute **twice** (with the same name) for the same species:

```c++
// full code (wrongly) generated

                value_identifier(float_X, MassRatio_species_H, 1.6738233791328002e-27 / SI::BASE_MASS_SI);
                value_identifier(float_X, ChargeRatio_species_H, 1.6021766339999999e-19 / SI::BASE_CHARGE_SI);

        using ParticleFlags_species_H = MakeSeq_t<
                    massRatio<MassRatio_species_H>,
                    chargeRatio<ChargeRatio_species_H>,


                    atomicNumbers<ionization::atomicNumbers::Hydrogen_t>,
                    ionizationEnergies<ionization::energies::AU::Hydrogen_t>,

            particlePusher<UsedParticlePusher>,
            shape<UsedParticleShape>,
            interpolation<UsedField2Particle>,
            current<UsedParticleCurrentSolver>>;

        using ParticleAttributes_species_H = MakeSeq_t<
            momentum,
                position<position_pic>
                    ,
                weighting
                    ,
                momentum
            >;

        using species_H = Particles<
            PMACC_CSTRING("H"),
            ParticleFlags_species_H,
            ParticleAttributes_species_H>;



    using VectorAllSpecies = MakeSeq_t<
            species_H
        >;
```

Note that `momentum` is listed twice.

> Macroparticles of the species **must be created** in the init pipeline, otherwise the behavior will not occur.

## Expected Behavior
compilation error "duplicate names are forbidden"

## Actual Behavior
compilation works fine, but execution fails **after** initialization with:

```
initialization time: 10sec 393msec = 10.393 sec
double free or corruption (!prev)
```

According to debugging output, this occurs during writing in OpenPMD:

```
[...]
PMaccVerbose UNDEFINED_LVL(80) | create event
PMaccVerbose UNDEFINED_LVL(80) | create event
PMaccVerbose UNDEFINED_LVL(80) | create event
PMaccVerbose UNDEFINED_LVL(80) | create event
PMaccVerbose UNDEFINED_LVL(80) | create event
PMaccVerbose UNDEFINED_LVL(80) | create event
PMaccVerbose UNDEFINED_LVL(80) | create event
[...]

PR #4078, opened 2022-04-26
Check uniqueness of particle attributes

sbastrakov:
As pointed out in #4072 we never checked that typelists for particle flags or attributes are unique. And somewhat surprisingly PIConGPU was compiling in that case, but not running.

This PR adds a pmacc meta algorithm to test for uniqueness, and then applies it for the particle-related typelists in question. A build error message for repeated flag looks like that (and similarly for attribute):
`
/home/bastra54/src/picongpu/include/pmacc/../pmacc/particles/ParticleDescription.hpp(87): error #304: no instance of function template "mpl_::assertion_failed" matches the argument list
            argument types are: (mpl_::failed ************(pmacc::ParticleDescription<pmacc::meta::String<'e'>, picongpu::SuperCellSize, boost::mpl::v_item<picongpu::totalCellIdx, boost::mpl::v_item<picongpu::weighting, boost::mpl::v_item<picongpu::momentum, boost::mpl::v_item<picongpu::position<picongpu::position_pic, pmacc::pmacc_isAlias>, boost::mpl::vector0<mpl_::na>, 0>, 0>, 0>, 0>, picongpu::ParticleFlagsElectrons, pmacc::HandleGuardRegion<pmacc::particles::policies::ExchangeParticles, pmacc::particles::policies::DoNothing>, boost::mpl::vector0<mpl_::na>, boost::mpl::v_item<pmacc::NextFramePtr<mpl_::_1>, boost::mpl::v_item<pmacc::PreviousFramePtr<mpl_::_1>, boost::mpl::vector0<mpl_::na>, 0>, 0>>::_error_particles_must_not_have_duplicate_flags____check_your_speciesDefinition_param_file_________94::************)(pmacc::StaticAssertError))
`

### DIFF
--- a/include/pmacc/meta/conversion/Unique.hpp
+++ b/include/pmacc/meta/conversion/Unique.hpp
@@ -1,0 +1,63 @@
+/* Copyright 2022 Sergei Bastrakov
+ *
+ * This file is part of PMacc.
+ *
+ * PMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with PMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "pmacc/meta/conversion/ToSeq.hpp"
+
+#include <boost/mpl/fold.hpp>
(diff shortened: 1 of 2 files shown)


psychocoderHPC (dismissed):

sbastrakov:
Wow, checking if the given typelist only has unique elements turned out surprisingly tricky to implement in PIConGPU with `boost::mpl`. On a simple side, we can not just call `bmpl::unique<>` as it only removes duplicates located contiguously.

My initial implementation tried inserting all elements to a `bmpl::set<>` and that insertion in principle removes duplicates. That should work fine and it does on CPUs.

However, on GPUs at least with nvcc we are getting some very weird issues with some flags being skipped from that uniqueness test and so a valid setup was falsely considered to have repeating flags. In particular, aliases `ionizationEnergies<>` and `effectiveNuclearCharge<>` caused this issue when used with their normal template parameters. If for the sake of testing one switches to a buildin type instead (and does not otherwise use those attributes so that PIConGPU compiles), the check however passes. We looked at this with @psychocoderHPC but didn't manage to get to the exact reason, it probably has something to do with either GPU storage of pmacc vectors involved in the template parameters, or 2-pass compilation of nvcc.

So I scratched the original implementation of the uniquness test and wrote a new, more "stupid" but working one.

psychocoderHPC (approved):

PR #4096, opened 2022-05-04
OpenPMD plugin data range selection

psychocoderHPC:
Add a new parameter `--openPMD.range` to allow the selection of a
contiguous cell domain for the output.

- update plugin documentation
- add a new helper function to calculate a window based on the user-provided range per dimension

After this PR is merged we can remove the FieldSlice plugin and the corresponding issues.

### DIFF
--- a/docs/source/usage/plugins/openPMD.rst
+++ b/docs/source/usage/plugins/openPMD.rst
@@ -116,6 +116,9 @@ PIConGPU command line option          description
 ===================================== ====================================================================================================================================================
 ``--openPMD.period``                  Period after which simulation data should be stored on disk.
 ``--openPMD.source``                  Select data sources and filters to dump. Default is ``species_all,fields_all``, which dumps all fields and particle species.
+``--openPMD.range``                   Define a contiguous range of cells per dimension to dump. Each dimension is separated by a comma.
+                                      A range is defined as ``[BEGIN:END)`` where out of range indices will be clipped.
+                                      Default is ``:,:,:``, which dumps all cells.
 ``--openPMD.file``                    Relative or absolute openPMD file prefix for simulation data. If relative, files are stored under ``simOutput``.
 ``--openPMD.ext``                     openPMD filename extension (this controls thebackend picked by the openPMD API).
 ``--openPMD.infix``                   openPMD filename infix (use to pick file- or group-based layout in openPMD). Set to NULL to keep empty (e.g. to pick group-based iteration layout).
--- a/include/picongpu/plugins/openPMD/openPMDWriter.hpp
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.hpp
@@ -202,6 +202,8 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
             // defined later since we need openPMDWriter constructor
 
             plugins::multi::Option<std::string> notifyPeriod = {"period", "enable openPMD IO [for each n-th step]"};
+            plugins::multi::Option<std::string> range
+                = {"range", "define the output range in cells for each dimension e.g. 1:10,:,42:"};
 
             plugins::multi::Option<std::string> source = {"source", "data sources: ", "species_all, fields_all"};
 
@@ -294,6 +296,7 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                 std::string concatenatedSourceNames = plugins::misc::concatenateToString(allowedDataSources, ", ");
 
                 notifyPeriod.registerHelp(desc, masterPrefix + prefix);
+                range.registerHelp(desc, masterPrefix + prefix);
                 source.registerHelp(desc, masterPrefix + prefix, std::string("[") + concatenatedSourceNames + "]");
                 tomlSources.registerHelp(desc, masterPrefix + prefix);
                 fileName.registerHelp(desc, masterPrefix + prefix);
@@ -920,6 +923,7 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                 }
             }
 
+
             void notify(uint32_t currentStep) override
             {
                 // notify is only allowed if the plugin is not controlled by the
@@ -931,7 +935,17 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                 mThreadParams.initFromConfig(*m_help, m_id, outputDirectory);
 
                 /* window selection */
-                mThreadParams.window = MovingWindow::getInstance().getWindow(currentStep);
+                auto simulationOutputWindow = MovingWindow::getInstance().getWindow(currentStep);
+
+                // set default if the user is not providing the parameter.
+                std::string selectedRange = ":,:,:";
+                if(m_help->range.optionDefined(m_id) && !m_help->range.get(m_id).empty())
+                    selectedRange = m_help->range.get(m_id);
+
+                const SubGrid<simDim>& subGrid = Environment<simDim>::get().SubGrid();
+                mThreadParams.window
+                    = plugins::misc::intersectRangeWithWindow(subGrid, simulationOutputWindow, selectedRange);
+
                 mThreadParams.isCheckpoint = false;
                 dumpData(currentStep);
             }
@@ -1410,7 +1424,7 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                 /* y direction can be negative for first gpu */
                 const pmacc::Selection<simDim> localDomain = Environment<simDim>::get().SubGrid().getLocalDomain();
                 DataSpace<simDim> particleOffset(localDomain.offset);
-                particleOffset.y() -= threadParams->window.globalDimensions.offset.y();
+                particleOffset -= threadParams->window.globalDimensions.offset;
 
                 std::vector<std::string> vectorOfDataSourceNames;
                 if(m_help->selfRegister)
--- a/docs/TBG_macros.cfg
+++ b/docs/TBG_macros.cfg
@@ -243,10 +243,12 @@ TBG_countPerSuper="--<species>_macroParticlesPerSuperCell.period 100 --<species>
 
 # Dump simulation data (fields and particles) via the openPMD API.
 # Data is dumped every .period steps to the fileset .file.
+# To select only parts of data .range can be used.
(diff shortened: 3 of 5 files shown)


docs/source/usage/plugins/openPMD.rst
```diff
@@ -116,6 +116,9 @@ PIConGPU command line option          description
 ===================================== ====================================================================================================================================================
 ``--openPMD.period``                  Period after which simulation data should be stored on disk.
 ``--openPMD.source``                  Select data sources and filters to dump. Default is ``species_all,fields_all``, which dumps all fields and particle species.
+``--openPMD.range``                   Define a contiguous range of cells per dimension to dump. Each dimension is separated by a comma.
```
sbastrakov, line 119: Should we add it's in user coordinate system (so a total domain where 0 is start of `CORE + BORDER` on first domain), or is it obvious ?

include/picongpu/plugins/misc/rangeToWindow.hpp
```diff
@@ -0,0 +1,115 @@
+/* Copyright 2017-2022 Rene Widera
```
sbastrakov: Why does this file name start with a small `r` and not `R`?

docs/source/usage/plugins/openPMD.rst
sbastrakov: Also, is it total or global domain, and should we clarify it?

include/picongpu/plugins/misc/rangeToWindow.hpp
psychocoderHPC: The year is wrong but whats wrong with the name? it is *R*ene

psychocoderHPC: Because it contains only a function and no class. All files which define only functions starts with lower case letters.

sbastrakov: Ah, i forgot that was our convention.

docs/source/usage/plugins/openPMD.rst
psychocoderHPC: It is is in global domain coordinates. I assumed it is obvious because the user is giving the domain size via `-d X Y Z` but since you asking we should maybe add it to the docstring.

> Should we add it's in user coordinate system (so a total domain where 0 is start of CORE + BORDER on first domain), or is it obvious ?

The user does not know `CORE + BORDER` this is a developer term. Maybe  we should reverence to the command line parameter `-d`

sbastrakov: Upon further consideration, I think where 0 is positioned is probably clear, as we are consistent there with the user side. But total vs. global is not clear imo, as both options theoretically make sense

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -202,6 +202,8 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
             // defined later since we need openPMDWriter constructor
 
             plugins::multi::Option<std::string> notifyPeriod = {"period", "enable openPMD IO [for each n-th step]"};
+            plugins::multi::Option<std::string> range
+                = {"range", "define the output range in cells for each dimension e.g. 1:10,:,42:"};
```
sbastrakov, line 206: I think you forgot to add it to `TBG_macros.cfg` (which is imo quite a bad workflow that we have to add any option to multiple places, but at least that's the current workflow)

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -931,7 +935,16 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                 mThreadParams.initFromConfig(*m_help, m_id, outputDirectory);
 
                 /* window selection */
-                mThreadParams.window = MovingWindow::getInstance().getWindow(currentStep);
+                auto simulationOutputWindow = MovingWindow::getInstance().getWindow(currentStep);
+
+                // set default if the user is not providing the parameter.
+                std::string selectedRange = ":,:,:";
+                if(m_help->range.optionDefined(m_id) && !m_help->range.get(m_id).empty())
+                    selectedRange = m_help->range.get(m_id);
```
sbastrakov, line 943: Could this be done not here, in the computational function, but somewhere before that?

sbastrakov: (I mean, for clarity reasons, and so that this part is not directly dependent on command-line argument)

include/picongpu/plugins/misc/rangeToWindow.hpp
```diff
@@ -0,0 +1,115 @@
...
+             */
+            template<uint32_t T_dim>
+            inline Window rangeToWindow(
+                pmacc::SubGrid<T_dim> const& dataDomain,
+                Window const& inputWindow,
+                std::string const& selectedRange)
+            {
+                const SubGrid<simDim>& subGrid = Environment<simDim>::get().SubGrid();
+
+                auto parsedSlice = pmacc::pluginSystem::toTimeSlice(selectedRange);
```
sbastrakov: Should some file be included for this? Also i noticed no include to `simulation_defines.hpp` at start of this file, not sure if intentional

sbastrakov: I guess this `toTimeSlice` is getting renamed some time soon, as here it's used for just slice, unrelated to time?

include/picongpu/plugins/misc/rangeToWindow.hpp
```diff
@@ -0,0 +1,115 @@
...
+             * @param inputWindow Selected window within the data domain e.g. the selected moving window.
+             * @param selectedRange String with range description for each dimension. Dimensions are separated by
+             *                      comma. The range for a dimension is of the form BEGIN:END where BEGIN is included
+             *                      and END excluded from the domain.
+             *                      If the begin or end of the range is outside of the data domain the values will be
+             *                      cropped to fit the data domain.
+             * @return Window description with local and global information.
+             */
+            template<uint32_t T_dim>
+            inline Window rangeToWindow(
```
sbastrakov: I believe this function actually does intersection of `inputWindow` and `selectedRange`, and not merely conversion of range to window. If so, perhaps a more direct naming could help

include/picongpu/plugins/misc/rangeToWindow.hpp
```diff
@@ -0,0 +1,115 @@
...
+                // Parse the ranges only for dimensions the user is providing.
+                for(uint32_t d = 0; d < parsedSlice.size() && d < simDim; ++d)
+                {
+                    auto range = parsedSlice[d];
+                    uint32_t const windowOffset = inputWindow.globalDimensions.offset[d];
+                    uint32_t const windowSize = inputWindow.globalDimensions.size[d];
+                    uint32_t const windowEnd = windowOffset + windowSize;
+                    uint32_t const rangeBegin = range.values[0];
+                    // Crop the user given end range to the end of the input window.
+                    uint32_t const rangeEnd = std::min(range.values[1], windowEnd);
```
sbastrakov: Is this correct? Isn't here `windowEnd` as total coordinates and `range.values[1]` in global ones?

include/picongpu/plugins/misc/rangeToWindow.hpp
```diff
@@ -0,0 +1,115 @@
...
+                     */
+                    resultWindow.localDimensions.offset[d]
+                        = std::max(0, subGrid.getLocalDomain().offset[d] - static_cast<int>(newWindowBegin));
+
+
+                    /* If the resulting window is before or after the local data domain the window size will be local
+                     * zero because the local domain is outside the window else the local domain is part of the window.
+                     */
+                    if(newWindowEnd <= subGrid.getLocalDomain().offset[d]
+                       || subGrid.getLocalDomain().size[d] + subGrid.getLocalDomain().offset[d] < newWindowBegin)
```
sbastrakov: Can't we remove this `if`, always do second branch and then crop to 0 if calculated size becomes negative?

sbastrakov (commented):
That's an extremely useful addition!

I have a few questions, shown above. Perhaps easier to go over those in VC or tomorrow in the office

include/picongpu/plugins/misc/rangeToWindow.hpp
psychocoderHPC: Yes the file should be included and I forgot to mention that I will rename the files in a follow-up PR.

include/picongpu/plugins/misc/rangeToWindow.hpp
psychocoderHPC: `inputWindow.globalDimensions.offset` is the offset from an arbitrary window to the global domain. see https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-domain-definitions#current-naming-scheme
A window Is a view within the global domain.

include/picongpu/plugins/misc/rangeToWindow.hpp
psychocoderHPC: No this `if` is required because there is no need that the range to be spanned over all devices.
It is possible to define a range with a single cell somewhere in the middle. In this case there are GPUs before the window and after the window which are not participating in the data gathering process.

include/picongpu/plugins/misc/rangeToWindow.hpp
sbastrakov: Maybe we could call it `intersectRangeWithWindow` or `intersectRangeWindow` or something along these lines

include/picongpu/plugins/misc/rangeToWindow.hpp
sbastrakov: Discussed offline, all fine here, I didn't understand the code.

include/picongpu/plugins/openPMD/openPMDWriter.hpp
psychocoderHPC: I checked the code. I like to avoid that we have `selectedRange` as a global attribute for an IO instance.
In case we allow defining more than one range per IO instance in the future the current location of this code is fine.

sbastrakov (dismissed):

sbastrakov (approved):

PR #4099, opened 2022-05-05
Add Thermal electron benchmark

psychocoderHPC:
Add Thermal electron test used in the EZ paper.

Co-authored-by: Klaus Steiniger <k.steiniger@hzdr.de>

Issue #3596, opened 2021-04-20
remove example ThermalTest?


psychocoderHPC:
Evaluate if the example [ThermalTest](https://github.com/ComputationalRadiationPhysics/picongpu/tree/dev/share/picongpu/examples/ThermalTest) can be removed.
For years nobody executed this example, it is not clear if it is useful to keep it.

# todo if we keep it
-  port python scripts to python3
- move the example into `share/picongpu/tests`
- add an automatic script to perform the test which is returning true/false depending on the test result (required for the CI).

steindev:
Also see #7. It is on my list.

steindev:
ThermalTest is gone #4100. But I should add automated analysis scripts to the new Thermal Benchmark #4099 in order to make it a real benchmark.

PrometheusPi:
Since the old test is removed and the new test setup is merged and an automatic test is currently developed by @MSVoss, I will close this issue

PR #4100, opened 2022-05-05
remove example: ThermalTest

psychocoderHPC:
The thermal test is since years not maintained or used.
Additionally, we like to reduce the CUSTL dependencies. This test is based on CUSTL.
An additional benefit of removing this test is that our CI will have to check fewer examples^^.
If such a test is required there is no problem to introduce a new tested example/test.

- [x] please do not merge until @steindev @PrometheusPi are giving green light via a review


PrometheusPi:
@psychocoderHPC I agree with you that this example is outdated and currently does not serve us as test case anymore. 
On the other hand implementing a test equivalent to this example would be great. However, this requires new methods, like the ones you are currently implementing in `[topic-openPMDSlicing3](https://github.com/psychocoderHPC/picongpu/tree/topic-openPMDSlicing3)`. 

Thus: Why do you want an explicit green light from  @sbastrakov and me? (Did I miss something here?) 

What do you think of re-implementing this example as test later on (after removing this ThermalTest example)?

PrometheusPi:
@psychocoderHPC and @steindev Is #4099 already an equivalent test?

psychocoderHPC:
> @psychocoderHPC and @steindev Is #4099 already an equivalent test?

I do not think that this test is equivalent because what I know from the past was shown somehow which plasma frequencies can be resolved with PIC.

psychocoderHPC:
> Thus: Why do you want an explicit green light from @sbastrakov and me? (Did I miss something here?)

I would like to have a green light from two physicists to avoid deleting something useful and required.

My point of view: I see only code an example that was not executed for 10+ years.

PrometheusPi:
Okay - I think such a test is in general very useful - but you are right - if the test has not been executed for more than 10 years - it is better to remove it and build it from scratch as a real test. 
(Green light from my side.)

PrometheusPi (approved):

steindev:
> Okay - I think such a test is in general very useful - but you are right - if the test has not been executed for more than 10 years - it is better to remove it and build it from scratch as a real test. (Green light from my side.)

I agree. There is actually an open issue I have on my plate... #3596

steindev (approved):

PR #4101, opened 2022-05-06
Add missing include in Cupla Common.hpp

franzpoeschel:
PIConGPU dev branch does currently not compile for me due to missing include

update @psychocoderHPC:
- [ ] update cupla to pull the fix

### DIFF
--- a/thirdParty/cupla/include/cupla/device/math/Common.hpp
+++ b/thirdParty/cupla/include/cupla/device/math/Common.hpp
@@ -25,6 +25,7 @@
 #include "cupla/types.hpp"
 
 #include <alpaka/core/Concepts.hpp>
+#include <alpaka/math/MathStdLib.hpp>
 
 #include <type_traits>
 


sbastrakov:
@psychocoderHPC I guess it's related to recent switch to alpaka 0.9.0. Is this a correct fix (i mean, maybe this header is missing in some other place that is included to here)?

psychocoderHPC:
Thanks @franzpoeschel for opening this PR. Since this is a bug in the third party library we need to fix it in cupla directly.
We should include `alpaka.hpp` instead of `MathStdLib.hpp` to pull all meth function for all required accelerators.

psychocoderHPC:
I set this PR to draft mode and added a note that cupla must be updated. I will open a PR in cupla to fix the issue.

sbastrakov:
Fix is merged to cupla repo (not to PIConGPU repo yet)

psychocoderHPC:
I opened #4104 which should fix this bug.

PR #4102, opened 2022-05-06
Fix incident field math around corners

sbastrakov:
The existing implementation had two bugs that went unnoticed as they didn't appear for axis-aligned lasers, but are a problem for angled lasers. I encountered them when making my angled laser prototype, and after this fix the prototype is working fine.

Bug fixes are in separate commits, commit messages have some explanations. Also slightly update description of indexing used.


psychocoderHPC (approved):

sbastrakov:
Can be merged @psychocoderHPC

PR #4105, opened 2022-05-12
Generalize incident field laser profiles to all directions

sbastrakov:
Also all focus positions and polarizations.

Note for reviewers: I think it would be much easier to review commit-by-commit, even though there are quite a few commits. I spent some time organizing work into commits to have a logical flow of changes.

TODO:
- [x] recheck that everything still works, as the PR is quite old already


include/picongpu/param/incidentField.param
```diff
@@ -38,10 +38,10 @@
  * `OFFSET` that controls inward offset of the generating surface relative to global domain. For example:
  *
  * @code{.cpp}
- * using XMin = profiles::Free< UserFunctorIncidentE >;
+ * using XMin = profiles::Free< UserFunctorIncidentE,  UserFunctorIncidentB>;
```
psychocoderHPC: Two spaces after the comma.
Could you please fix the code style of the following lines too, there should be no space before and after `<,>`

share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
```diff
@@ -38,10 +38,10 @@
  * `OFFSET` that controls inward offset of the generating surface relative to global domain. For example:
  *
  * @code{.cpp}
- * using XMin = profiles::Free< UserFunctorIncidentE >;
+ * using XMin = profiles::Free< UserFunctorIncidentE,  UserFunctorIncidentB>;
```
psychocoderHPC: Code style update will be welcome.

include/picongpu/fields/incidentField/profiles/PlaneWave.hpp
```diff
@@ -42,122 +42,106 @@ namespace picongpu
...
-                            float_64 const runTime = DELTA_T * currentStep;
-                            float_64 envelope = float_64(Unitless::AMPLITUDE);
-                            float_64 const mue = 0.5 * Unitless::RAMP_INIT * Unitless::PULSE_LENGTH;
-                            float_64 const w = 2.0 * PI * Unitless::f;
-                            float_64 const tau = Unitless::PULSE_LENGTH * math::sqrt(2.0);
-                            float_64 const endUpramp = mue;
-                            float_64 const startDownramp = mue + Unitless::LASER_NOFOCUS_CONSTANT;
-                            float_64 integrationCorrectionFactor = 0.0;
-                            if(runTime > startDownramp)
+                            auto envelope = Unitless::AMPLITUDE;
```
psychocoderHPC, line 109: In the implementation before we used everywhere 64 bit floating-point numbers. I assume to guarantee stable numerics.
Is the precision change to 32 bit intended?

include/picongpu/fields/incidentField/profiles/Wavepacket.hpp
```diff
@@ -46,45 +44,41 @@ namespace picongpu
...
-                        static constexpr float_64 INIT_TIME = Params::PULSE_INIT * PULSE_LENGTH
-                            + LASER_NOFOCUS_CONSTANT; // unit: seconds (full initialization length)
-                        static constexpr float_64 endUpramp = -0.5_X * LASER_NOFOCUS_CONSTANT; // unit: seconds
-                        static constexpr float_64 startDownramp = 0.5_X * LASER_NOFOCUS_CONSTANT; // unit: seconds
-                        static constexpr float_64 f = SPEED_OF_LIGHT / WAVE_LENGTH;
-                        static constexpr float_64 w = 2.0 * PI * f;
+                            = static_cast<float_X>(Params::LASER_NOFOCUS_CONSTANT_SI / UNIT_TIME);
+
+                        // unit: UNIT_TIME
+                        static constexpr float_X INIT_TIME
```
psychocoderHPC, line 60: Is the precision change to 32 bit intended?

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
```diff
@@ -42,35 +43,30 @@ namespace picongpu
...
+                            = static_cast<float_X>(Params::LASER_NOFOCUS_CONSTANT_SI / UNIT_TIME);
 
-                        static constexpr float_64 TIME_PREPULSE
-                            = static_cast<float_64>(Params::TIME_PREPULSE_SI / UNIT_TIME);
-                        static constexpr float_64 TIME_PEAKPULSE
-                            = static_cast<float_64>(Params::TIME_PEAKPULSE_SI / UNIT_TIME);
-                        static constexpr float_64 TIME_1 = static_cast<float_64>(Params::TIME_POINT_1_SI / UNIT_TIME);
-                        static constexpr float_64 TIME_2 = static_cast<float_64>(Params::TIME_POINT_2_SI / UNIT_TIME);
-                        static constexpr float_64 TIME_3 = static_cast<float_64>(Params::TIME_POINT_3_SI / UNIT_TIME);
+                        static constexpr float_X TIME_PREPULSE
```
psychocoderHPC, line 58: precision change from 64 to 32 bit.

psychocoderHPC (changes requested):
After a long review, I found only a few precision changes and code style issues in a comment.

psychocoderHPC:
@sbastrakov Feel free to push the changes to a separate commit to avoid fighting to update single commits in the history.

include/picongpu/fields/incidentField/profiles/PlaneWave.hpp
sbastrakov: In previous implementation some operations were done on host (e.g. precalculating things for the given `currentStep`). Those were 64-bit. What is in-kernel was 32-bit. Now everything is in-kernel basically, so I converted all values to 32-bit.

sbastrakov: You have a point tho, that i converted everything more or less on auto pilot. I will go through these changes and try to see if they are potentially dangerous.

sbastrakov:
Just FYI we offline discussed with @steindev that he doesn't have time for review this week due to the retreat incoming, and so hopefully it will be done till end of May.

include/picongpu/fields/incidentField/profiles/PlaneWave.hpp
psychocoderHPC: I am fine if it is 32bit but rechecking if there is a potential issue would be fine.

psychocoderHPC:
If @steindev gives his OK I am also open to merging it before.

sbastrakov:
I think this one can actually benefit from a review from a physicist, at least in how the profiles are now formulated and if it (and the naming) makes sense from physical point of view.

include/picongpu/fields/incidentField/profiles/BaseParam.def
```diff
@@ -0,0 +1,149 @@
...
+                     * unit: none
+                     *
+                     * @{
+                     */
+                    static constexpr float_64 DIRECTION_X = 0.0;
+                    static constexpr float_64 DIRECTION_Y = 1.0;
+                    static constexpr float_64 DIRECTION_Z = 0.0;
+                    /** @} */
+
+                    /** Focus position in total cooridnate system
```
steindev: typo: coordinate

steindev: Also, with total coordinate system I assume you mean the system whose origin is at the same position where the origin is in our output files.

include/picongpu/fields/incidentField/Functors.hpp
```diff
@@ -86,6 +93,197 @@ namespace picongpu
...
+                                    = (static_cast<float_X>(totalDomainCells[axis] - OFFSET[axis][1]) - 0.75_X)
+                                    * cellSize[axis];
+                                firstIntersectionP
+                                    = std::min(firstIntersectionP, (maxPosition - focus[axis]) / direction[axis]);
+                            }
+                        }
+                        return focus + firstIntersectionP * direction;
+                    }
+
+                    /** Get unit axis vectors of internal cooridnate system
```
steindev: again typo

steindev (changes requested):
Only checked some of the files so far...need to proceed later.

include/picongpu/fields/incidentField/profiles/GaussianBeam.def
```diff
@@ -46,81 +48,32 @@ namespace picongpu
...
-                        enum PolarisationType
-                        {
-                            LINEAR_AXIS_1 = 1u,
-                            LINEAR_AXIS_2 = 2u,
-                            CIRCULAR = 4u,
-                        };
-                        /** Polarization selection
-                         */
-                        static constexpr PolarisationType Polarisation = CIRCULAR;
+                        /** @} */
```
steindev, line 76: A lot of the parameters, like wavelength, pulse duration and a0, vanished and went to into BaseParam. I am not sure this is a good idea, because a user will need to copy all the parameters from two different files into one struct in his example. This is quite a complex a task, as the user would, first, need to know in in which files to look, and second, need to know where these are located.

In general, we currently miss an example showing how to configure a laser from a predefined profile. which together with the above makes this task impossible for all who do not know how to do it beforehand.

However, it somehow does make sense to define a base class for the parameters in common to all profiles.
I do not know how to solve this issue at the moment.

sbastrakov: This is true, but I do not think it's reasonably possible to keep lasers generalized otherwise. The old lasers are all largely copy-paste of one another, and I really felt it when first making their incident field analogs with doing exactly the same transformation over and over. It also makes a really bad convention that e.g. amplitude is always called the same, but it's never formalized as each parameter class is technically independent.

Wrt navigating, this should be solvable with doxygen. An example would also help of course

sbastrakov: Regarding examples, the plan is to transform most existing examples to use the incident field laser after this PR is in.

include/picongpu/fields/incidentField/profiles/GaussianBeam.hpp
```diff
@@ -110,20 +115,17 @@ namespace picongpu
                      * We always take tilt value from the unitless params and apply the tile (which can be 0).
```
steindev: tile -> tilt?

include/picongpu/fields/incidentField/profiles/GaussianBeam.def
steindev: Well, I think for now (until we transform  the examples, but possibly also after this), can you add a comment to the section in `incidentField.param`, where the predefined profiles are mentioned, which explains how the pre-defined profiles are configured. That is, saying that profiles are configured via a param struct which can be constructed by taking the base struct + the laser specific struct, as well as mentioning the place where to find these. Also we should probably add a comment there that it would be fail-safe practice to add the laser source to all incidentField planes, except YMax in moving window simulations (while making sure that the laser does not propagate along -y for these kind of simulations).

Also Fabia (new student) could change the laser initialization method in at least one of the examples directly after this PR is merged to practice incidentField setup, before implementing new profiles.

steindev:
Finished reviewing. I did not check all files in detail, but the general procedure as well as the Gaussian pulse implementation and am in principle happy with the changes made. Please also answer the few comments above to which you did not yet react.

include/picongpu/fields/incidentField/profiles/GaussianBeam.hpp
```diff
@@ -187,113 +185,74 @@ namespace picongpu
...
+                    private:
+                        /** Get value for the given position
+                         *
+                         * @param totalCellIdx cell index in the total domain (including all moving window slides)
+                         * @param phaseShift additional phase shift to add on top of everything else,
+                         *                   in radian
+                         */
+                        HDINLINE float_X getValue(floatD_X const& totalCellIdx, float_X const phaseShift) const
+                        {
+                            // transform coordinate system to center of x-z plane of initialization
```
sbastrakov: Outdated comment, forgot to fix

include/picongpu/fields/incidentField/profiles/BaseParam.def
sbastrakov: Yes, `total` as in our [standard coordinates](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-domain-definitions)

include/picongpu/fields/incidentField/profiles/PlaneWave.hpp
sbastrakov: So i rechecked and in the original (non-incident field) laser profiles all those values were `float_X` except the double-precision `f`. Upon looking at the code, I also see no reason for `f` to require double precision.

Then for the incident field profiles as in the current dev I kinda automatically converted a lot of values to `float_64` as they were only used on the host side for the time-depending part, and only kept those with on-device use `float_X`. However, with this PR due to time depending on position, all the host-side stuff moved back on device and so these values are `float_X` again. This appears safe as, again, almost all were originally `float_X`, and for `f` it also seems fine.

include/picongpu/fields/incidentField/profiles/Wavepacket.hpp
sbastrakov: Same as above.

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
sbastrakov: This was actually `float_64` in the original laser profile. I checked how it's used, there is no requirement to have it as `float_64`, so changing to `float_X` appears fine.

include/picongpu/fields/incidentField/profiles/GaussianBeam.def
sbastrakov: Added a note about it to `incidentField.param`.

sbastrakov:
@psychocoderHPC your suggestions are addressed. With the 32-bit vs. 64-bit I checked and it appears safe to use 32-bit for all those cases, as it's all in internal units and also does not involve much operations.

sbastrakov:
Oh one more thing. When making the PR I accidentally used spelling `Polarization` and `PolarizationType`. In the codebase we use both spellings without any system, but in the existing laser parameter structs it is `Polarisation` and `PolarisationType`. I can change the spelling in this PR to match it.

Edit: I made this change, so we are back to `Polarisation`.

steindev (approved):

steindev (approved):

psychocoderHPC (approved):

PR #4106, opened 2022-05-13
KHI_growthRate Testsuite

MSVoss:
Integrate the test suite that compares the growth rates for the Kelvin Helmholtz instability setup

### DIFF
--- a/lib/python/test/KHI_growthRate/CalculateParameter.py
+++ b/lib/python/test/KHI_growthRate/CalculateParameter.py
@@ -1,5 +1,5 @@
 """
-This file is part of the PIConGPU.
+This file is part of PIConGPU.
 
 Copyright 2022 PIConGPU contributors
 Authors: Mika Soren Voss
@@ -7,6 +7,7 @@ License: GPLv3+
 """
 
 import numpy as np
+from scipy.constants import c, m_e, e, epsilon_0
 
 
 class CalculateParameter:
@@ -31,12 +32,6 @@ class CalculateParameter:
         this function calculate the growthrate of a given quantity f(t)
     """
 
-    # fixe parameter
-    __c = 299792458
-    __m_e = 9.1093837015e-31
-    __e = -1.602176634e-19
-    __e_0 = 8.8541878128e-12
-
     def __init__(self):
         pass
 
@@ -58,7 +53,7 @@ class CalculateParameter:
                  speed given by gamma
         """
 
-        return np.sqrt((1 - 1 / gamma**2) * CalculateParameter.__c ** 2)
+        return np.sqrt((1 - 1 / gamma**2) * c ** 2)
 
     @staticmethod
     def calculateTime(steps, deltaT: float, frequency):
@@ -100,7 +95,7 @@ class CalculateParameter:
 
         """
 
-        return v_0 / CalculateParameter.__c
+        return v_0 / c
 
     @staticmethod
     def theoryPlasmafrequence_MI(density):
@@ -124,8 +119,8 @@ class CalculateParameter:
 
         """
 
-        return np.sqrt((density * CalculateParameter.__e**2) / (
-               CalculateParameter.__e_0 * CalculateParameter.__m_e))
+        return np.sqrt((density * e**2) / (
+               epsilon_0 * m_e))
 
     @staticmethod
     def theoryPlasmafrequence_ESKHI(density, gamma):
@@ -153,9 +148,8 @@ class CalculateParameter:
 
         """
 
-        return np.sqrt((density * CalculateParameter.__e ** 2) /
-                       (CalculateParameter.__e_0 * gamma *
-                        CalculateParameter.__m_e))
+        return np.sqrt((density * e ** 2) /
+                       (epsilon_0 * gamma * m_e))
 
     @staticmethod
     def growthRate(f, time, interval=None):
--- a/lib/python/test/KHI_growthRate/CalculateParameter.py
+++ b/lib/python/test/KHI_growthRate/CalculateParameter.py
@@ -1,0 +1,205 @@
+"""
+This file is part of the PIConGPU.
+
+Copyright 2022 PIConGPU contributors
+Authors: Mika Soren Voss
+License: GPLv3+
+"""
+
+import numpy as np
+
+
+class CalculateParameter:
+    """
+    This class calculates essential quantities, such as the speed or the
+    growth rate, for the KHI
+
+    functions:
+    -------
+    calculateV_O(gamma):
+        calculate the speed v
+
+    calculateBeta(v_0)
+
+    theoryPlasmafrequence_MI(density):
+        calculate the plasmafrequency for the MI
+
+    theoryPlasmafrequence_ESKHI(density, gamma):
+        calculate the plasmafrequency for the ESKHI
+
+    growthRate(f, time, interval = None):
+        this function calculate the growthrate of a given quantity f(t)
+    """
+
+    # fixe parameter
+    __c = 299792458
+    __m_e = 9.1093837015e-31
+    __e = -1.602176634e-19
+    __e_0 = 8.8541878128e-12
+
+    def __init__(self):
+        pass
+
+    @staticmethod
+    def calculateV_O(gamma):
+        """
+        calculate the speed v and use
+
+            gamma  = 1 / (1 - beta**2)**0.5
+
+        input:
+        -------
+        gamma:   float or numpy.ndarray
+                 lorentz factor
+
+        return:
+        -------
+        v:       float or numpy.ndarray
+                 speed given by gamma
+        """
+
+        return np.sqrt((1 - 1 / gamma**2) * CalculateParameter.__c ** 2)
+
+    @staticmethod
+    def calculateTime(steps, deltaT: float, frequency):
+        """
+        calculate the time in [frequency ** -1]
+
+        input:
+        -------
+        steps:     list
+                   list of number of timesteps
+
+        deltaT:    float
+                   time step of the simulation
+
+        frequency: float
+
+        return:
+        -------
+        time:      list
+        """
+
+        return steps * deltaT * frequency
+
+    @staticmethod
+    def calculateBeta(v_0):
+        """
+        calculate beta with
+
+            beta = v_0 / c
+
+        input:
+        -------
+        v_0:    float or numpy.ndarray
+                speed
+
+        return:
+        -------
+        beta:    float or numpy.ndarray
+
+        """
+
+        return v_0 / CalculateParameter.__c
+
+    @staticmethod
+    def theoryPlasmafrequence_MI(density):
+        """
+        calculate the plasmafrequency for the MI using the formula
+
+            omega = sqrt(density * e ** 2 / ( eps_0 * m_e))
+
+            e...electron charge
+            eps_0...electric field constant
+            m_e... mass electron
+
+        input:
+        -------
+        density:  float
+
+        return:
+        -------
+        omega:    float
+                  plasmafrequency for the MI
+
+        """
+
+        return np.sqrt((density * CalculateParameter.__e**2) / (
+               CalculateParameter.__e_0 * CalculateParameter.__m_e))
+
+    @staticmethod
+    def theoryPlasmafrequence_ESKHI(density, gamma):
+        """
+        calculate the plasmafrequency for the ESKHI using the formula
+
+            omega = sqrt(density * e ** 2 / ( eps_0 * gamma * m_e))
+
+            e...electron charge
+            eps_0...electric field constant
+            gamma...lorentz factor
+            m_e... mass electron
+
+        input:
+        -------
+        density:  float
+
+        gamma:    float
+                  lorentz factor
+
+        return:
+        -------
+        omega:    float
+                  plasmafrequency for the ESKHI
+
+        """
+
+        return np.sqrt((density * CalculateParameter.__e ** 2) /
+                       (CalculateParameter.__e_0 * gamma *
+                        CalculateParameter.__m_e))
+
+    @staticmethod
+    def growthRate(f, time, interval=None):
+        """
+        This function calculate the growthrate of a given quantity f(t)
+
+        Gamma_f(t_k)= log(f(t_(k+1))/f(t_(k-1)))/t_(k+1)- t_(k-1)
+
+        input:
+        -------
+        f:         array
+                   quantity whose growth rate is to be determined
+
+        time:      array
+                   the array of the timesteps with the same size
+                   as the quantity f
+
+        interval:  [starttime, endtime], optional
+                   determines the time interval in which the growth rate
+                   is to be determined
+
+        return:
+        -------
+        growth rate: array if interval = None
+                     growth rate calculated using the above formula
+
+                     tuple of (growthrate, start, stop)
+                     growth rate calculated using the above formula and
+                     the start and end of the interval
+        """
+
+        if interval is None:
+            Gamma = 0.5 * np.log(f[3:]/f[1:-2])/(time[3:] - time[1:-2])
+            return Gamma
+
+        else:
+            # calculate lower limit
+            start = np.where(time < interval[0])[0][-1] + 1
+
+            # calculate upper limit
+            stop = np.where(time > interval[1])[0][0] - 1
+
+            Gamma = 0.5 * np.log(f[start + 1:stop + 1] / f[start - 1:stop - 1]
+                                 ) / (time[start + 1: stop + 1] -
+                                      time[start - 1: stop - 1])
+
+            return Gamma, start, stop
--- a/lib/python/test/KHI_growthRate/KHIData.py
+++ b/lib/python/test/KHI_growthRate/KHIData.py
@@ -1,5 +1,5 @@
 """
-This file is part of the PIConGPU.
+This file is part of PIConGPU.
 
 Copyright 2022 PIConGPU contributors
 Authors: Mika Soren Voss
@@ -39,18 +39,6 @@ class KHIData:
     setAllData(allParameter):
         sets the data for gamma, density and delta t
 
-    getLightSpeed() -> int:
-        return the speed of light [m/s]
-
-    getElectronMass() -> float:
-        return the electron mass [kg]
-
-    getElectronCharge() -> float:
-        return the electron charge [C]
-
-    getVacuumPermittivity() -> float:
-        return the permittivity of the vacuum [As/(Vm)]
-
     getFieldSum() -> dict:
         return the field data
 
@@ -70,12 +58,6 @@ class KHIData:
         set the growth rate of a specific field given by key
 
     """
-
-    # fixed parameters
-    __c = 299792458
-    __m_e = 9.1093837015e-31
-    __e = -1.602176634e-19
-    __e_0 = 8.8541878128e-12
 
     def __init__(self):
         """constructor"""
@@ -111,23 +93,6 @@ class KHIData:
         self.density = allParameter[1]
         self.deltaT = allParameter[2]
 
-    # for fixed paramters only getter
-    def getLightSpeed(self) -> int:
-
-        return self.__c
-
-    def getElectronMass(self) -> float:
-
-        return self.__m_e
-
-    def getElectronCharge(self) -> float:
-
-        return self.__e
-
-    def getVacuumPermittivity(self) -> float:
-
-        return self.__e_0
-
     def getFieldSum(self) -> dict:
         """
         return:
@@ -154,7 +119,7 @@ class KHIData:
             if an incorrect key was passed to the function
         """
 
-        if(key in self.__fieldsum.keys()):
+        if (key in self.__fieldsum.keys()):
             return self.__fieldsum.get(key)
         else:
             raise KeyError("Use a valid key(total, E_x, E_y, E_z,"
@@ -178,7 +143,7 @@ class KHIData:
         KeyError:
             if an incorrect key was passed to the function
         """
-        if(key in self.__fieldsum.keys()):
+        if (key in self.__fieldsum.keys()):
             self.__fieldsum[key] = value
         else:
             raise KeyError("Use a valid key(total, E_x, E_y, E_z,"
@@ -209,7 +174,7 @@ class KHIData:
             if an incorrect key was passed to the function
         """
 
-        if(key in self.__growthrate.keys()):
+        if (key in self.__growthrate.keys()):
             return self.__growthrate.get(key)
         else:
             raise KeyError("Use a valid key(total, E_x, E_y, E_z,"
@@ -234,7 +199,7 @@ class KHIData:
             if an incorrect key was passed to the function
         """
 
-        if(key in self.__growthrate.keys()):
+        if (key in self.__growthrate.keys()):
             self.__growthrate[key] = value
         else:
             raise KeyError("Use a valid key(total, E_x, E_y, E_z,"
--- a/lib/python/test/KHI_growthRate/KHIData.py
+++ b/lib/python/test/KHI_growthRate/KHIData.py
@@ -1,0 +1,241 @@
+"""
+This file is part of the PIConGPU.
+
+Copyright 2022 PIConGPU contributors
+Authors: Mika Soren Voss
+License: GPLv3+
+"""
+
+
+class KHIData:
+    """
+    This class holds any data required by the test-suite
+
+    attributes:
+    -------
+    gamma:  float
+
+    v_o:    float
+
+    density:float
+
+    w_pe:   float
+            plasma frequency
+
+    deltaT: float
+            time steps of the simulation
+
+    time:   list
+            list of the time of the simulation
+
+    field:  str
+            contains the name of the main field
+
+    regime: str
+            contains the predominate regime (ESKHI/MI)
+
+    functions:
+    -------
+    setAllData(allParameter):
+        sets the data for gamma, density and delta t
+
+    getLightSpeed() -> int:
+        return the speed of light [m/s]
+
+    getElectronMass() -> float:
+        return the electron mass [kg]
+
+    getElectronCharge() -> float:
+        return the electron charge [C]
+
+    getVacuumPermittivity() -> float:
+        return the permittivity of the vacuum [As/(Vm)]
+
+    getFieldSum() -> dict:
+        return the field data
+
+    getFieldSumByKey(key: str):
+        return the field data of a specific field given by key
+
+    setFieldSumByKey(key: str, value):
+        set the field data of a specific field given by key
+
+    getGrowthRate() -> dict:
+        return the growth rates for each field
+
+    getGrowthRateByKey(key: str):
+        return the growth rate of a specific field given by key
+
+    setGrowthRateByKey(key: str, value):
+        set the growth rate of a specific field given by key
+
+    """
+
+    # fixed parameters
+    __c = 299792458
+    __m_e = 9.1093837015e-31
+    __e = -1.602176634e-19
+    __e_0 = 8.8541878128e-12
+
+    def __init__(self):
+        """constructor"""
+
+        self.gamma = None
+        self.v_0 = None
+        self.density = None
+        self.deltaT = None
+        self.w_pe = None
+        self.time = None
+        self.field = "B_x"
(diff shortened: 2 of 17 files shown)


psychocoderHPC:
The python scripts are not pep8 formated: https://gitlab.com/hzdr/crp/picongpu/-/jobs/2453910558
All python scripts need to pass pep8 code style checks and flake8 checks.

test/KHI_growthRate/etc/picongpu/hemera-hzdr/k80_testKHI.tpl
```diff
@@ -0,0 +1,135 @@
...
+
+if [ $PIC_RET != 0 ]; then
+    echo "running PIConGPU failed with: " $PIC_RET
+    exit $PIC_RET
+fi
+
+# running the testsuite
+MAINTEST="$TBG_dstPath/input/lib/python/test/KHI_growthRate"
+cd $TBG_dstPath/simOutput/
+python $MAINTEST/MainTest.py $MAINTEST $TBG_dstPath/input/include/picongpu/param/ $TBG_dstPath/simOutput/
```
psychocoderHPC: We should not introduce a `tpl` file in examples because these file will be soon outdated.
Could you please provide a `test.sh` bash script (do not forget to set execution rights) in the root of the input set `KHI_growthRate` which returns the error code of the python script e.g.
```
#!/bin/bash
MAINTEST=input/lib/python/test/KHI_growthRate"
cd simOutput/
python $MAINTEST/MainTest.py $MAINTEST ../input/include/picongpu/param/ .
exit $?
```

I assume that the python script returns zero in case the test passed and non zero else.

I will add in a separate PR the short code snipped that copies `test.sh` when calling `tbg`

lib/python/test/KHI_growthRate/MainTest.py
```diff
@@ -0,0 +1,38 @@
...
+	plotter = ViewerKHI(paramDir, simDir)
+	plotter.plotGrowthRate(field, True, True, True)
+
+except:	
+	error0 = str(sys.exc_info()[0])
+	error1 = str(sys.exc_info()[1])
+	error2 = str(sys.exc_info()[2])	
+
+	#print error0 + error1 + error2
+	fobj_out = open(sys.argv[3]+ "/fehler.log","w")
```
psychocoderHPC: please use english terms `e.g. khi_error.log`

lib/python/test/KHI_growthRate/MainTest.py
```diff
@@ -0,0 +1,38 @@
+import sys
```
psychocoderHPC, line 9: Please add the license header to all of your python script https://github.com/ComputationalRadiationPhysics/picongpu/blob/d9ab3532a2b0d10e752b5b701ec14254f486d176/lib/python/picongpu/plugins/plot_mpl/base_visualizer.py#L1-L7

Do not forget to set the year to `2022` only and add yourself as the author.

lib/python/test/KHI_growthRate/MainTest.py
```diff
@@ -0,0 +1,38 @@
...
+from time import *
+
+lt = localtime()
+date = strftime("date: %d.%m.%Y", lt)
+timeOfDay = strftime("time: %H:%M:%S", lt)
+
+try:
+	sys.path.append(sys.argv[1])
```
psychocoderHPC: Why is this required? Does python has no way to get the file location of the executed file?

lib/python/test/KHI_growthRate/MainTest.py
```diff
@@ -0,0 +1,38 @@
...
+	from TestKHI import TestKHI
+	from ViewerKHI import ViewerKHI
+
+	field = ["total", "B_x", "B_y", "B_z", "E_x", "E_y", "E_z"]
+
+	paramDir = sys.argv[2]
```
psychocoderHPC: Where should param dir points too? From the `tpl` it looks like it is the folder of `MainTest.py` which is equal to the dirname of the first parameter.

lib/python/test/KHI_growthRate/MainTest.py
```diff
@@ -0,0 +1,38 @@
...
+	simDir = sys.argv[3]
```
psychocoderHPC: Which files must be located in `simDir`?
If you need the openPMD output maybe the prefix path + filename would be better. e.g. `simOutpur/opePMD/simData`.
The openPMD file names and folder can be chosen by the user, therefore we should allow to specify it.
IMO it would be useful to use the python argument parses instead of using unnamed fixed position parameters for the testing tool, see https://github.com/ComputationalRadiationPhysics/picongpu/blob/d9ab3532a2b0d10e752b5b701ec14254f486d176/src/tools/bin/plot_chargeConservation.py#L184

psychocoderHPC (changes requested):

test/KHI_growthRate/etc/picongpu/hemera-hzdr/k80_testKHI.tpl
```diff
@@ -0,0 +1,135 @@
...
+if [ -f !TBG_dstPath/input/bin/cuda_memtest ] && [ !TBG_numHostedGPUPerNode -eq !TBG_gpusPerNode ] ; then
+  # Run CUDA memtest to check GPU's health
+  mpiexec -np !TBG_tasks !TBG_dstPath/input/bin/cuda_memtest.sh
+else
+  echo "Note: GPU memory test was skipped as no binary 'cuda_memtest' available or compute node is not exclusively allocated. This does not affect PIConGPU, starting it now" >&2
+fi
+
+if [ $? -eq 0 ] ; then
+  # Run PIConGPU
+  mpiexec !TBG_dstPath/input/bin/picongpu !TBG_author !TBG_programParams
```
sbastrakov: Which `.cfg` files should be used to run it?

sbastrakov: I think some reasonable `.cfg` files should be included in the setup, like we do for existing examples.

lib/python/test/KHI_growthRate/ViewerKHI.py
```diff
@@ -0,0 +1,244 @@
+"""
```
sbastrakov, line 1: Please add `requirements.txt` for the added Python part.

lib/python/test/KHI_growthRate/MainTest.py
MSVoss: not really, paramDir points to the folder containing the .param files. sys.argv[0] is MainTest.py directly, [1] is the folder in which MainTest.py is located (I'm still checking your previous comment), [2] is as described above and [3] points to simOutput

test/KHI_growthRate/etc/picongpu/hemera-hzdr/k80_testKHI.tpl
MSVoss: I use 16.cfg and only switch TBG_steps to something smaller like 3000

PrometheusPi:
I had a brief discussion with @psychocoderHPC:
Please transfer the code from the `*.tpl` that runs the test to a test script `validate.sh` that returns `0` in case of passing and `1` in case of not passing the growth rate test. 
A second program called `ci.sh` should first run PIConGPU directly from the command line `picongpu -d 1 1 1 -g ... -s ... ...` as in your `submit.start` files and the call `validate.sh`. 
Both scripts should be included in a `bin` directory inside your test directory.
The reason for the separate `ci.sh` file is, that our GPU based test system does not have (yet) a scheduler and thus will run PIConGPU and later the test as in a regular script. 

We can discuss details in person after your lecture if you have time.

test/KHI_growthRate/etc/picongpu/hemera-hzdr/k80_testKHI.tpl
```diff
@@ -0,0 +1,135 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# PIConGPU is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PIConGPU.
```
psychocoderHPC: Please remove the tpl file from our PR. The file will be outdated soon. You can add a `README.md` into the example where you describe what the user must do to validate the results.

psychocoderHPC (changes requested):

psychocoderHPC:
@MSVoss Could you please squash your PRs into two single commits? One commit with the python stuff and one with the next ample.

sry for the late review I was always looking if the tpl files are already removed and thought the PR is not updated because the file was still in.

test/KHI_growthRate/bin/ci.sh
```diff
@@ -0,0 +1,92 @@
...
+        currentPath=$simDir/bin/
+    else
+        mkdir -p $simDir
+fi
+
+cd $currentPath/../
+pic-build -b cuda:70
+
+mkdir $simDir/simOutput
+if [ ! -d "$simDir/include/picongpu/param/" ] ; then
```
psychocoderHPC: Why is this required?
I do not understand why we need to copy the include files after PIConGPU is compiled in line 74.

psychocoderHPC (changes requested):

psychocoderHPC:
@MSVoss Sry for the late response, there is only one open question before we can merge this code.

test/KHI_growthRate/bin/ci.sh
MSVoss: Because I didn't want to pass another parameter to validate.sh. The copying can be removed if validate.sh gets another optional parameter giving it direction to .param files. With default simDir if not set. If this solution is better I'll change it.

test/KHI_growthRate/bin/ci.sh
```diff
@@ -0,0 +1,92 @@
...
+        echo "Destination path already in use, cannot create new folder" >&2
+        exit 1
+    fi
+fi
+
+if [ -z "$testPar" ] ; then
+        testPar=$HOME
+fi
+
+source etc/picongpu/bash-devServer-hzdr/gpu_picongpu.profile.example
```
steindev: I am not aware that the directory `bash-devServer-hzdr` exists

lib/python/test/KHI_growthRate/TestKHIManager.py
```diff
@@ -0,0 +1,440 @@
+"""
+This file is Part of the PIConGPU+
```
steindev: 2 spelling errors. Actually three. You say 'part of the PIConGPU' in every file, but it is supposed to be 'part of PIConGPU' without 'the'.

lib/python/test/KHI_growthRate/TestKHIManager.py
```diff
@@ -0,0 +1,440 @@
...
+    getTime():
+        returns time in [w_pe ** -1]
+
+    getFieldsSumByName(key):
+        returns field data for the passed fields
+
+    getGrowthRateByName(key):
+        returns growth rate for the passed fields
+
+    getAcceptanceLevel() -> float:
```
steindev, line 43: What is the meaning of `-> float` and why is it only written for some of the functions

lib/python/test/KHI_growthRate/TestKHIManager.py
```diff
@@ -0,0 +1,440 @@
...
+
+    getKHIData():
+        returns the object that contains the data of the KHI
+
+    getMaxGrowthRate_Sim(field) -> float:
+        return the max growth rate of the
+        transferred field from the simulation
+    """
+
+    __acceptanceLevel = 0.2
```
steindev, line 68: There needs to be a discussion somewhere, how this value comes about. (probably upload your thesis to zenodo and cite it here AND write a paragraph in the documentation?)

lib/python/test/KHI_growthRate/TestKHIManager.py
```diff
@@ -0,0 +1,440 @@
...
+        """
+
+        if field is None:
+            field = self.__khiData.field
+
+        if self.__khiData.regime == "ESKHI":
+            maxValue = theoGR.getTheoryESKHI_max(
+                                            self.__khiData.gamma)
+        else:
+            maxValue = theoGR.getTheoryMIGR_max(
```
steindev: Please add a comment which regime the else branch concerns

steindev: There is code duplication between this function and the next. Can't you call the next function in this function and then just to devision to obtain the percantage value?

lib/python/test/KHI_growthRate/ViewerKHI.py
```diff
@@ -0,0 +1,242 @@
...
+            plt.plot(gamma, growthRate, ".k", label="value from simulation")
+
+        if plotLog:
+            plt.loglog()
+
+        plt.xlabel(r"$\gamma$")
+        plt.ylabel(r"$\Gamma[\omega_{pe}]$")
+
+        plt.legend()
+        plt.show()
```
steindev: Is this script supposed to run interactively? I am wondering because the plots are not saved to disk but just shown.

test/KHI_growthRate/bin/ci.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
```
steindev, line 1: First, please add a license header here as well.

Second, as this is a, maybe THE,  central file, add a little description of its purpose.

test/KHI_growthRate/bin/validate.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
steindev, line 1: same as `ci.sh`

test/KHI_growthRate/bin/ci.sh
steindev: Or put this description as the first lines of `--help` (preferred)

lib/python/test/KHI_growthRate/ViewerKHI.py
```diff
@@ -0,0 +1,242 @@
...
+                    plt.plot(time, field_sum_ene[number],
+                             label="$" + lab + "$")
+
+        plt.legend()
+        plt.xlabel(r"$t[\omega_{pe}^ {-1}]$")
+        plt.ylabel(r"$\varepsilon_\mathrm{Fi}$")
+        plt.semilogy()
+        plt.grid()
+
+        plt.show()
```
steindev: same as above.

test/KHI_growthRate/include/picongpu/param/density.param
```diff
@@ -0,0 +1,46 @@
+/* Copyright 2013-2021 Axel Huebl, Heiko Burau, Rene Widera, Felix Schmitt,
```
steindev: This is an old license header (year 2021). You can also add your name, as you prepared the setup.

steindev: Same for all other files.

steindev (changes requested):
Excellent that this is going to mainline!

I have a general question:
The whole test suite setup is very KHI centric at the moment. I had expected that there would be general applicable parts and KHI specific parts,where the general part can be extended with more test cases and KHI is only the first of several test cases.
Is this separation into general part and setup specific part something you want to develop with the next test cases in order to study what this general part needs to look like?
Or do you think that each test case needs to provide its own infrastructure for e.g. start, success evaluation, and garbage collection?

Second major point:
I miss documentation in our readthedocs. It does not need to be a lot. Its main purpose is to tell users that we have a test suite.
Plus a top level view of the design (probably some graphic from your Bachelor thesis), a documentation of main design decisions (like the acceptance value), and the necessary references pointing to publications with more (or all) details.
I guess this will basically be an `executive summary` of your thesis, maybe one A4 page in total.

Disclaimer: You structured the source code very well and separated implementation details in distinct files. In the review I did not proceed until the lowest level of your data structures but dived only as deep as `TestKHIManager.py`.

lib/python/test/KHI_growthRate/TestKHIManager.py
MSVoss: As in other programming languages, the type of a return value can be specified in Python (here with -> float). However, this only makes sense where a fixed type exists and the function does not "swallow" what it gets.

test/KHI_growthRate/bin/ci.sh
MSVoss: This folder exists in the dev branch.

MSVoss: But there was still an error hidden in the line. (Correction comes with all others)

lib/python/test/KHI_growthRate/ViewerKHI.py
MSVoss: I had thought that these functions could possibly help one or the other physicist with the evaluation. Since these are only additional, I left them interactive for now. But if the possibility of storage should be given, I'll work this in. Just let me know.

MSVoss:
I don't think that every single test case needs its own structure. However, with each new test case, I work out a generalization of the test suite. In the new test case, which I'm working on now, some classes are renamed and generalized.

psychocoderHPC:
The CI is unhappy :-( there are some style issues in the py files.

test/KHI_growthRate/bin/ci.sh
psychocoderHPC: note: `#!/bin/bash` must be the first line and the license header must be after it.

lib/python/test/KHI_growthRate/TestKHIManager.py
MSVoss: I removed the function (it's still included in the source code in case I forgot to use the generalized one somewhere)

MSVoss:
I hope I've corrected everything except for the documentation

psychocoderHPC (approved):

lib/python/test/KHI_growthRate/CalculateParameter.py
```diff
@@ -0,0 +1,205 @@
...
+
+    growthRate(f, time, interval = None):
+        this function calculate the growthrate of a given quantity f(t)
+    """
+
+    # fixe parameter
+    __c = 299792458
+    __m_e = 9.1093837015e-31
+    __e = -1.602176634e-19
+    __e_0 = 8.8541878128e-12
```
PrometheusPi: Could be better taken from `scipy.constants` but fine for now

test/KHI_growthRate/bin/ci.sh
```diff
@@ -0,0 +1,98 @@
...
+        echo "Destination path already in use, cannot create new folder" >&2
+        exit 1
+    fi
+fi
+
+if [ -z "$testPar" ] ; then
+        testPar=$HOME
+fi
+
+source $testPar/picongpu/etc/picongpu/bash-devServer-hzdr/gpu_picongpu.profile.example
```
psychocoderHPC: please remove this line. Whoever is executing this script must set up their environment before.
The CI is not running on the local dev server.

psychocoderHPC: You can assume that `pic-create` is available.

test/KHI_growthRate/bin/ci.sh
```diff
@@ -0,0 +1,98 @@
...
+if [ "$create" == "true" ] ; then
+        mktemp -d $date_stamp
+        pic-create $testPar/picongpu/test/KHI_growthRate/ $simDir
+        currentPath=$simDir/bin/
+    else
+        mkdir -p $simDir
+fi
+
+cd $currentPath/../
+pic-build -b cuda:70
```
psychocoderHPC: please call `pic-build` only. Which backend we use and which architecture is setup by the user or the ci before

psychocoderHPC: We will execute this on CPU maybe too.

test/KHI_growthRate/bin/ci.sh
```diff
@@ -0,0 +1,98 @@
...
+
+mkdir $simDir/simOutput
+
+cd $simDir/simOutput
+mpiexec -n 1 $currentPath/picongpu -d 1 1 1 -g 192 512 12 --periodic 1 1 1 -s 3000 --fields_energy.period 10
+cd ..
+
+# test with python
+bash $currentPath/validate.sh -t $testPar -s $simDir
```
psychocoderHPC, line 91: The tool should return the error code of the validation

```bash
bash $currentPath/validate.sh -t $testPar -s $simDir
ret=$?

if [ "$create" == "true" ] ; then
        rm -r $simDir
fi

exit $ret
```

psychocoderHPC (changes requested):

lib/python/test/KHI_growthRate/KHIData.py
```diff
@@ -0,0 +1,241 @@
...
+    setGrowthRateByKey(key: str, value):
+        set the growth rate of a specific field given by key
+
+    """
+
+    # fixed parameters
+    __c = 299792458
+    __m_e = 9.1093837015e-31
+    __e = -1.602176634e-19
+    __e_0 = 8.8541878128e-12
```
PrometheusPi: as before, you could use scipy constants

lib/python/test/KHI_growthRate/ParamReader.py
```diff
@@ -0,0 +1,309 @@
...
+
+                    # calculate the parameter if multiplication is included
+                    elif "*" in line[line.find("=") + 1:line.find(";")]:
+                        s = line[line.find("=") + 1:line.find(";")].split("*")
+                        for entry in s:
+                            if parameter is None:
+                                parameter = float(entry)
+                            else:
+                                if entry in '  PARAM_FACTOR':
+                                    entry = 0.15789473684210525
```
PrometheusPi: Where does this magic number come from?

PrometheusPi: Please add a comment what this is, if you want to keep this number.

lib/python/test/KHI_growthRate/TestKHIManager.py
PrometheusPi: I agree - for now I would go for a comment.

lib/python/test/KHI_growthRate/TheoryGrowthRate.py
```diff
@@ -0,0 +1,114 @@
...
+        gamma:  float
+                lorentz factor
+
+        return:
+        -------
+        Lambda:  float
+                 maximum growth rate of the MI
+        """
+
+        c = 299792458
```
PrometheusPi: magic number - you could use scipy constants (or add a comment)

lib/python/test/KHI_growthRate/ViewerKHI.py
PrometheusPi: I think storing them would be great (but as long as the storing option for the last method exist, I am fine)

test/KHI_growthRate/include/picongpu/param/memory.param
```diff
@@ -0,0 +1,118 @@
+/* Copyright 2013-2022 Axel Huebl, Rene Widera, Benjamin Worpitz,
```
PrometheusPi: Is this file actually needed or can we use the default?

test/KHI_growthRate/include/picongpu/param/radiation.param
```diff
@@ -0,0 +1,195 @@
+/* Copyright 2013-2022 Rene Widera, Richard Pausch, Mika Soren Voss
```
PrometheusPi: This file should be removed since the test does not require a specific radiation plugin setup.

test/KHI_growthRate/include/picongpu/param/radiationObserver.param
```diff
@@ -0,0 +1,87 @@
+/* Copyright 2013-2022 Heiko Burau, Rene Widera, Richard Pausch,
```
PrometheusPi: same as with `radiation.param`

test/KHI_growthRate/include/picongpu/param/random.param
```diff
@@ -0,0 +1,60 @@
+/* Copyright 2014-2022 Axel Huebl, Rene Widera
```
PrometheusPi: Are here any changes compared to the default?

PrometheusPi (changes requested):
Just some minor remarks: 
 - ignore the scipy.constant comments if you want - but since scipy is already a requirement using it would not introduce an extra dependency
  - definitely remove all param files related to radiation
  - remove param files that do not differ from the base default `./include/picongpu/param/`

lib/python/test/KHI_growthRate/TestKHIManager.py
```diff
@@ -0,0 +1,420 @@
+"""
+This file is Part of PIConGPU.
```
steindev: 'Part' is still capital.

lib/python/test/KHI_growthRate/TestKHIManager.py
steindev: I do not really understand the comment. Is there a comma missing?

lib/python/test/KHI_growthRate/TestKHIManager.py
steindev: If it is not supposed to be used anymore, please remove it completely. We do not want to have dead code lying around. It will just be a source of confusion in the future.

lib/python/test/KHI_growthRate/ViewerKHI.py
steindev: How these can be visualized interactively is definitely something for the documentation.

steindev:
> I hope I've corrected everything except for the documentation

Documentation is left for a separate PR?

steindev (changes requested):
Almost done from my perspective. Added questions to old comments and two (or so) new ones.

lib/python/test/KHI_growthRate/TestKHIManager.py
MSVoss: Done

MSVoss:
> > I hope I've corrected everything except for the documentation
> 
> Documentation is left for a separate PR?

Yes

lib/python/test/KHI_growthRate/TestKHIManager.py
steindev: @MSVoss Don't forget to update this one as discussed last Thursday.

lib/python/test/KHI_growthRate/ViewerKHI.py
steindev: @MSVoss Plz add this to your notes of what needs to be in the documentation. Then you can mark this as resolved.

lib/python/test/KHI_growthRate/ViewerKHI.py
steindev: @MSVoss same as line 129 above. (By now lines 138 and 179)

lib/python/test/KHI_growthRate/ParamReader.py
steindev: @MSVoss @PrometheusPi Resolved?

steindev:
@MSVoss I checked all comments and marked those as resolved that I could evaluate. There are only a few missing that you need to check back. I believe we discussed last week something that has not been mentioned here. But I cannot remember what it was. If you do not remember as well or do not know what I mean than we will just merge.

lib/python/test/KHI_growthRate/TestKHIManager.py
steindev: :+1:

psychocoderHPC:
I will do the last review this week.

psychocoderHPC:
@MSVoss ~~could you please check the following formating/import issues~~

The formatting issue is in the third-party folder, please rebase your PR against the latest dev to solve the problems.

psychocoderHPC (approved):

psychocoderHPC:
@MSVoss thx for the updates, from my side all, looks fine!

@steindev @PrometheusPi could you please recheck this PR again until tomorrow so that we can merge it soon?

PrometheusPi:
@psychocoderHPC will do

PrometheusPi (approved):

steindev (approved):

PR #4107, opened 2022-05-16
use `if constexpr()` to check for simulation/object dimension

psychocoderHPC:
Use `if constexpr()` instead of runtime check for special code passed for 2 or 3-dimensional cases.

I have not touched the laser implementation even if there are cases that can use `if constexpr()` too. This avoids merging conflicts with the work @sbastrakov is currently doing.

### DIFF
--- a/include/picongpu/fields/absorber/pml/Field.tpp
+++ b/include/picongpu/fields/absorber/pml/Field.tpp
@@ -124,7 +124,7 @@ namespace picongpu
                     // Note: since this should compile for 2d, .z( ) can't be used
                     using detail::makeIdx;
                     int layerIdx = 0;
-                    if(simDim == 3)
+                    if constexpr(simDim == 3)
                     {
                         auto const negativeZLayer
                             = Layer{makeIdx(0, 0, 0), makeIdx(gridSize[0], gridSize[1], negativeSize[2])};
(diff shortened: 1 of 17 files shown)


sbastrakov (dismissed):
Not a shortcoming of this PR, but a suggestion for further improvement.

Since these places are already identified now (plus I have some in incident field lasers indeed), can we go one step further and wrap these checks into some helper variables or functions, so that we don't have to use the low level `simDim, DIM2, DIM3` ? E.g. to have `if constexpr picongpu::is3d` (or however we choose to call them) instead of current `if constexpr(simDim == DIM3)`

psychocoderHPC:
> Not a shortcoming of this PR, but a suggestion for further improvemen,
> 
> Since these places are already identified now (plus I have some in incident field lasers indeed), can we go one step further and wrap these checks into some helper variables or functions, so that we don't have to use the low level `simDim, DIM2, DIM3` ? E.g. to have `if constexpr picongpu::is3d` (or however we choose to call them) instead of current `if constexpr(simDim == DIM3)`

We can do it in a separate PR for `simDim`, good idea. Note, in PMacc this would not be possible because there is no global `simDim` available.

include/pmacc/mappings/kernel/MappingDescription.hpp
```diff
@@ -71,7 +71,7 @@ namespace pmacc
             {
                 /*minimal 3 blocks are needed if we have guarding blocks*/
                 int minBlock = std::min(gridSuperCells.x(), gridSuperCells.y());
-                if(DIM == DIM3)
+                if constexpr(DIM == DIM3)
                 {
                     minBlock = std::min(minBlock, gridSuperCells[2]);
```
sbastrakov: Now CI gives a warning/error `error: variable 'minBlock' set but not used [-Werror=unused-but-set-variable]` about line 73. Here we don't use variable `minBlock` at all, then this piece should be removed alltogether.

psychocoderHPC: I removed the change from this PR at all. The code part is "broken" and must be fixed. I will provide a fix in a seperate PR.

sbastrakov (approved):

PR #4108, opened 2022-05-16
Generalize incident field to support typelists of profiles

sbastrakov:
A user can now set either an individual profile or a typelist of profiles for each boundary. In case a typelist is used, the resulting field is a sum of effects of all profiles in the list. Update documentation to reflect the new interface.

Update incident field example to demonstrate the new typelist usage.

The change is logically independent from #4105. I believe it will cause a conflict technically, but that would be quite easy to resolve. This PR probably gets reviewed and merged first.

### DIFF
--- a/include/picongpu/param/incidentField.param
+++ b/include/picongpu/param/incidentField.param
@@ -19,7 +19,7 @@
 
 /** @file incidentField.param
  *
- * Configure incident field profile and gap between Huygence surface and absorber for each boundary.
+ * Configure incident field profile and offset of the Huygens surface for each boundary.
  *
  * Available profiles:
  *  - profiles::ExpRampWithPrepulse<> : exponential ramp with prepulse wavepacket with given parameters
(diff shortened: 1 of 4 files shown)


include/picongpu/param/incidentField.param
```diff
@@ -19,7 +19,7 @@
 
 /** @file incidentField.param
  *
- * Configure incident field profile and gap between Huygence surface and absorber for each boundary.
```
sbastrakov, line 22: This was either forgotten or missed during rebase when changing gap to offset a couple weeks ago.

sbastrakov:
@psychocoderHPC ready for review.

psychocoderHPC (approved):

share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
```diff
@@ -205,7 +206,7 @@ namespace picongpu
                  *                  field_internal = field_SI / unitField
                  */
                 HINLINE FunctorYMaxIncidentE(float_X const currentStep, float3_64 const unitField)
-                    : FunctorBase(currentStep)
+                    : FunctorBase(currentStep - static_cast<float_X>(T_delaySteps))
```
psychocoderHPC, line 209: Is `FunctorBase` able to handle negative times?

psychocoderHPC (approved):

share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param
sbastrakov: Yes

PR #4111, opened 2022-05-17
Add a new functor to sample macroparticles in a cell

sbastrakov:
It samples positions randomly uniformly inside a cell. The weights are sampled to somewhat mimic a true global sampling vs. cell-based sampling we do in other functors. That yields internal binomial distribution, hence the name as I didn't manage to come up with anything better.

I did some math derivation for that, as described in the `.def` file and was earlier posted in a mattermost channel.

Tbh I am not sure how it should perform. In theory it appears to be a more "statistically honest" sampling than what we normally do. However, perhaps what we normally do is actually better for our use case. This sampling may have more fidelity wrt mixing and natural irregularities. So perhaps it can e.g. more faithfully reproduce KHI growth rate. However, I did not check it yet, and honestly have no time for this side-task. Perhaps after #4106 is merged, we can more simply test it, or if @PrometheusPi has time for it.

- [x] Add to compile-time tests somewhere

### DIFF
--- a/include/picongpu/particles/InitFunctors.hpp
+++ b/include/picongpu/particles/InitFunctors.hpp
@@ -64,7 +64,7 @@ namespace picongpu
          *    - Evaluate the amount of real particles in the cell, Nr, using T_DensityFunctor.
          *    - If Nr > 0, decide how to represent it with macroparticles using T_PositionFunctor:
          *        - (For simplicity we describe how all currently used functors but RandomPositionAndWeightingImpl
-         *           operate, see below for customization)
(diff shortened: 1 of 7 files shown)


psychocoderHPC (approved):

sbastrakov:
@psychocoderHPC can be merged

PrometheusPi:
@sbastrakov Thanks for providing this new initialization method. I hope to find time to test it within the next to weeks. Perhaps @MSVoss has some to test this es well.

PR #4112, opened 2022-05-18
Remove SliceFieldPrinter plugin

### LINKED ISSUE #3965: CUDA dependency of SliceFieldPrinter
Hello,

I am currently working on a new plugin which is based on the `SliceFieldPrinter` plugin and I was wondering, why [the docs](https://picongpu.readthedocs.io/en/latest/usage/plugins/sliceFieldPrinter.html#known-limitations) say that the `SliceFieldPrinter` plugin is depending on CUDA. When @steindev and I looked through [the files](https://github.com/ComputationalRadiationPhysics/picongpu/blob/master/include/picongpu/plugins/SliceFieldPrinter.hpp) we couldn't really find anything CUDA-dependent. But when @steindev tried to compile and run a simulation on his laptop without CUDA installed but with the plugin manually enabled in the `PluginController.hpp`, PIConGPU could compile with the plugin, but it crashed without any output. So the questions are:

Where does the SliceFieldPrinter plugin depend on CUDA?
Would it be possible to bypass this dependency?

PS: Additionally we briefly looked at the `PositionsParticles` plugin and we are not sure, why it depends on CUDA either

sbastrakov:
The field slicing option is now part of general openPMD output since https://github.com/ComputationalRadiationPhysics/picongpu/pull/4096, so no need for a dedicated plugin.

Also fixes  #3965 and one fewer client of cuSTL in our code base.


psychocoderHPC (approved):

psychocoderHPC:
-640 lines of code :heart:

sbastrakov:
@psychocoderHPC and one step closer to no cuSTL 🙃

PR #4113, opened 2022-05-18
CUDA: fix false warning

psychocoderHPC:
follow up of #4107

```
include/pmacc/../pmacc/memory/buffers/DeviceBufferIntern.hpp(143): warning #940-D: missing return statement at end of non-void function "pmacc::DeviceBufferIntern<TYPE, DIM>::getPointer [with TYPE=picongpu::float3_X, DIM=3U]"
          detected during:
            instantiation of "TYPE *pmacc::DeviceBufferIntern<TYPE, DIM>::getPointer() [with TYPE=picongpu::float3_X, DIM=3U]"
```

The warning is a false warning because in any case there will be a
returned value. This PR rewrites the piece of code so that nvcc understands what's going on.


sbastrakov (approved):
Thanks for fixing! I can confirm those annoying walls of warnings disappear on dev system after this fix.

PR #4114, opened 2022-05-18
Switch to alpaka::Complex<>

sbastrakov:
Remove own `pmacc::math::Complex<>` and replace it with `alpaka::Complex<>`.

It only required a slight change due to `real()` vs. `get_real()`, `::type` vs. `::value_type` and similar.

- [x] Check that the `TransitionRadiation` example produces same results on `1.cfg`.


sbastrakov:

psychocoderHPC (approved):

PR #4122, opened 2022-05-20
Improve reporting of wrong MPI/devices configuration

### LINKED ISSUE #4121: Improve reporting of unmatching number of MPI ranks and devices requested
Currently it throws an exception with a not very clear wording `wrong parameters or wrong mpirun-call!`. This led to confusion e.g. [here](https://github.com/ComputationalRadiationPhysics/picongpu/issues/4109#issuecomment-1131627992). What happened was, the simulation was 2d but > 1 processes in z was requested in a `.cfg` file and the overall number of ranks was calculated as for 3d. So PIConGPU first silently cut the process configuration to `simDim`, so threw away z, and then the numbers didn't match and the error message was confusing.

I think these are two separate issues:

- unclear exception text
- silently cutting off `z` in the device configuration when it's not equal to 1

sbastrakov:
Fixes #4121, the motivation is described here.

For example, now if one asks for 2 MPI ranks and 1x1x2 process configuration in a 2d simulation, first a warning will be printed for a user to check their configuration: `Warning: 2 devices requested for z in a 2d simulation, this parameter will be reset to 1. Number of MPI ranks must be equal to the number of devices in x * y`. Then, since the internally used configuration will be `1x1x1` (we reset z to 1 and that is independent from this PR) the MPI configuration check will fail and throw with `Wrong configuration of processes or wrong MPI launch call: MPI_COMM_WORLD has 2 ranks, but process configuration 1x1x1 was requested`

### DIFF
--- a/include/picongpu/simulation/control/Simulation.hpp
+++ b/include/picongpu/simulation/control/Simulation.hpp
@@ -186,6 +186,11 @@ namespace picongpu
             else
                 while(devices.size() < 3)
                     devices.push_back(1);
+            // check for request of > 1 device in z for a 2d simulation, this is probably a user's mistake
+            if((simDim == 2) && (devices[2] > 1))
+                std::cerr
+                    << "Warning: " << devices[2] << " devices requested for z in a 2d simulation, this parameter "
+                    << "will be reset to 1. Number of MPI ranks must be equal to the number of devices in x * y\n";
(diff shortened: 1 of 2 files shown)


psychocoderHPC (approved):

sbastrakov:

PR #4124, opened 2022-05-25
Cached shapes

psychocoderHPC:
- Implement caching option for each shape which can be used as a particle shape (CIC-PCS)
- Add a wrapper to provide a unified interface to use cached and just in time evaluated shapes.
- Refactor particle assignment algorithm and use assignment shape wrapper.
- Refactor filed interpolation and use shape wrapper.

The new shapes will be used in a separate PR for the current deposition. The wrapper is not increasing the register footprint for the updated algorithms.
The performance will be slightly increased and the number of operations of the interpolation is strongly reduced.

### DIFF
--- a/include/picongpu/algorithms/AssignedTrilinearInterpolation.hpp
+++ b/include/picongpu/algorithms/AssignedTrilinearInterpolation.hpp
@@ -49,26 +49,36 @@ namespace picongpu
 {
     struct AssignedTrilinearInterpolation
     {
-        /** Does a 3D trilinear field-to-point interpolation for
+        /** Does a trilinear field-to-point interpolation for
          * arbitrary assignment function and arbitrary field_value types.
          *
-         * @tparam T_AssignmentFunction function for assignment
          * @tparam T_begin lower margin for interpolation
          * @tparam T_end upper margin for interpolation
+         * @tparam T_AssignmentFunction type of shape functors
          *
          * @param cursor cursor pointing to the field
-         * @param pos position of the interpolation point
+         * @param shapeFunctors Array with d shape functors, where d is the dimensionality of the field represented by
+         *                      cursor. The shape functor must have the interface to call
+         *                      `operator()(relative_grid_point)` and return the assignment value for the given grid
+         *                      point.
          * @return sum over: field_value * assignment
          *
          * interpolate on grid points in range [T_begin;T_end]
+         *
+         * @{
          */
-        template<typename T_AssignmentFunction, int T_begin, int T_end, typename T_Cursor>
-        HDINLINE static auto interpolate(const T_Cursor& cursor, const float3_X& pos) ->
-            typename ::pmacc::result_of::Functor<AssignedTrilinearInterpolation, T_Cursor>::type
+        template<int T_begin, int T_end, typename T_Cursor, typename T_AssignmentFunction>
+        HDINLINE static auto interpolate(
+            const T_Cursor& cursor,
+            const pmacc::memory::Array<T_AssignmentFunction, 3>& shapeFunctors)
         {
-            using type = typename ::pmacc::result_of::Functor<AssignedTrilinearInterpolation, T_Cursor>::type;
-
             [[maybe_unused]] constexpr auto iterations = T_end - T_begin + 1;
+
+            using type = decltype(*cursor(0, 0, 0) * shapeFunctors[0](0));
+
+            /* The implementation assumes that x is the fastest moving index to iterate over contiguous memory
+             * e.g. a row, to optimize memory fetch operations.
+             */
             auto result_z = type(0.0);
             PMACC_UNROLL(iterations)
             for(int z = T_begin; z <= T_end; ++z)
@@ -83,33 +93,39 @@ namespace picongpu
                         /* a form factor is the "amount of particle" that is affected by this cell
                          * so we have to sum over: cell_value * form_factor
                          */
-                        result_x += *cursor(x, y, z) * T_AssignmentFunction()(float_X(x) - pos.x());
+                        result_x += *cursor(x, y, z) * shapeFunctors[0](x);
 
-                    result_y += result_x * T_AssignmentFunction()(float_X(y) - pos.y());
+                    result_y += result_x * shapeFunctors[1](y);
                 }
-
-                result_z += result_y * T_AssignmentFunction()(float_X(z) - pos.z());
+                result_z += result_y * shapeFunctors[2](z);
             }
             return result_z;
         }
 
         /** Implementation for 2D position*/
-        template<class T_AssignmentFunction, int T_begin, int T_end, class T_Cursor>
-        HDINLINE static auto interpolate(T_Cursor const& cursor, float2_X const& pos) ->
-            typename ::pmacc::result_of::Functor<AssignedTrilinearInterpolation, T_Cursor>::type
+        template<int T_begin, int T_end, class T_Cursor, class T_AssignmentFunction>
+        HDINLINE static auto interpolate(
+            T_Cursor const& cursor,
+            const pmacc::memory::Array<T_AssignmentFunction, 2>& shapeFunctors)
         {
-            using type = typename ::pmacc::result_of::Functor<AssignedTrilinearInterpolation, T_Cursor>::type;
+            [[maybe_unused]] constexpr int iterations = T_end - T_begin + 1;
 
-            type result_y = type(0.0);
+            using type = decltype(*cursor(0, 0) * shapeFunctors[0](0));
+            /* The implementation assumes that x is the fastest moving index to iterate over contiguous memory
+             * e.g. a row, to optimize memory fetch operations.
+             */
+            auto result_y = type(0.0);
+            PMACC_UNROLL(iterations)
             for(int y = T_begin; y <= T_end; ++y)
             {
-                type result_x = type(0.0);
+                auto result_x = type(0.0);
+                PMACC_UNROLL(iterations)
                 for(int x = T_begin; x <= T_end; ++x)
                     // a form factor is the "amount of particle" that is affected by this cell
                     // so we have to sum over: cell_value * form_factor
-                    result_x += *cursor(x, y) * T_AssignmentFunction()(float_X(x) - pos.x());
+                    result_x += *cursor(x, y) * shapeFunctors[0](x);
 
-                result_y += result_x * T_AssignmentFunction()(float_X(y) - pos.y());
+                result_y += result_x * shapeFunctors[1](y);
             }
             return result_y;
         }
@@ -120,5 +136,7 @@ namespace picongpu
             return propList;
         }
     };
+
+    //! @}
 
 } // namespace picongpu
--- a/include/picongpu/algorithms/FieldToParticleInterpolation.hpp
+++ b/include/picongpu/algorithms/FieldToParticleInterpolation.hpp
@@ -23,6 +23,7 @@
 #include "picongpu/simulation_defines.hpp"
 
 #include "picongpu/algorithms/ShiftCoordinateSystem.hpp"
+#include "picongpu/particles/shapes.hpp"
 
 #include <pmacc/attribute/unroll.hpp>
 #include <pmacc/cuSTL/algorithm/functor/GetComponent.hpp>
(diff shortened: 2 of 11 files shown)


include/picongpu/algorithms/AssignedTrilinearInterpolation.hpp
```diff
@@ -52,64 +52,69 @@ namespace picongpu
...
+         *                          The shape functor must have the interface to call `operator()(relative_grid_point)`
+         *                          and return the assignment value for the given grid point.
          * @return sum over: field_value * assignment
          *
          * interpolate on grid points in range [T_begin;T_end]
          */
-        template<typename T_AssignmentFunction, int T_begin, int T_end, typename T_Cursor>
-        HDINLINE static auto interpolate(const T_Cursor& cursor, const float3_X& pos) ->
-            typename ::pmacc::result_of::Functor<AssignedTrilinearInterpolation, T_Cursor>::type
+        template<int T_begin, int T_end, typename T_Cursor, typename T_AssignmentFunctionArray>
```
sbastrakov: It seems like `T_AssignmentFunctionArray` is actually a single functor type, so should be named `T_AssignmentFunction` ?

include/picongpu/algorithms/AssignedTrilinearInterpolation.hpp
```diff
@@ -52,64 +52,69 @@ namespace picongpu
...
+        HDINLINE static auto interpolate(
+            const T_Cursor& cursor,
+            const pmacc::memory::Array<T_AssignmentFunctionArray, 3>& shapeFunctorArray)
```
sbastrakov: Maybe the variable could be just called `shapeFunctors` ? I don't think it's important that it's an array

include/picongpu/algorithms/AssignedTrilinearInterpolation.hpp
```diff
@@ -52,64 +52,69 @@ namespace picongpu
...
          * arbitrary assignment function and arbitrary field_value types.
          *
-         * @tparam T_AssignmentFunction function for assignment
          * @tparam T_begin lower margin for interpolation
          * @tparam T_end upper margin for interpolation
+         * @tparam T_AssignmentFunctionArray type of the array with shape functors
          *
          * @param cursor cursor pointing to the field
-         * @param pos position of the interpolation point
+         * @param shapeFunctorArray Array with d shape functors, where d is the number of simulation.
```
sbastrakov: Is `d` dimensionality? Then could `simDim` be used?

include/picongpu/algorithms/AssignedTrilinearInterpolation.hpp
```diff
@@ -52,64 +52,69 @@ namespace picongpu
...
             PMACC_UNROLL(iterations)
             for(int z = T_begin; z <= T_end; ++z)
             {
-                auto result_y = type(0.0);
+                auto result_y = float_X(0.0);
                 PMACC_UNROLL(iterations)
                 for(int y = T_begin; y <= T_end; ++y)
                 {
-                    auto result_x = type(0.0);
+                    auto result_x = float_X(0.0);
```
sbastrakov: Unrelated to this PR, but perhaps we should add some comment that here we process in the order of out memory layouts (any order is logically correct, but we do what's sensible for performance). So in this part we implicitly rely on knowldge of how buffers are organized

include/picongpu/algorithms/FieldToParticleInterpolation.hpp
```diff
@@ -63,6 +64,22 @@ namespace picongpu
         static constexpr int begin = -supp / 2 + (supp + 1) % 2;
         static constexpr int end = begin + supp - 1;
 
+        HDINLINE auto getShapeArray(float3_X const& pos) const
```
sbastrakov: Add doxygen comment here? At least explain that `pos` is in cell

include/picongpu/algorithms/FieldToParticleInterpolation.hpp
```diff
@@ -63,6 +64,22 @@ namespace picongpu
...
+            return result;
+        };
+
+        HDINLINE auto getShapeArray(float2_X const& pos) const
+        {
+            pmacc::memory::Array<shapes::Cached<AssignmentFunction>, 2> result;
+            result[0] = shapes::Cached<AssignmentFunction>(pos.x(), true);
+            result[1] = shapes::Cached<AssignmentFunction>(pos.y(), true);
+            return result;
+        };
```
sbastrakov: I think at least this part is easy to express generically for 2d and 3d and not have two copy-pasted functions?

include/picongpu/particles/shapes/NGP.hpp
```diff
@@ -76,7 +76,7 @@ namespace picongpu
                     /** form factor of this particle shape.
```
sbastrakov, line 124: Why isn't this shape extended with the new interface?
For `Counter` i understand it would never be used, as this is not a "real" macroparticle shape

include/picongpu/particles/shapes/PCS.hpp
```diff
@@ -86,63 +110,108 @@ namespace picongpu
...
+                        auto shapeValues = detail::PCS::shapeArray<support + 1>(x);
+
+                        // Update value so that a particle can be out of range without using lmem/local memory on GPUs
+                        // because of dynamic indexing into an array located in registers.
+                        shapeValues[5] = isOutOfRange ? shapeValues[4] : 0.0_X;
+                        shapeValues[4] = isOutOfRange ? shapeValues[3] : shapeValues[4];
+                        shapeValues[3] = isOutOfRange ? shapeValues[2] : shapeValues[3];
+                        shapeValues[2] = isOutOfRange ? shapeValues[1] : shapeValues[2];
+                        shapeValues[1] = isOutOfRange ? shapeValues[0] : shapeValues[1];
+                        shapeValues[0] = isOutOfRange ? 0.0_X : shapeValues[0];
```
sbastrakov, line 164: Could these be expressed as an `if` statement, which imo would be more readable then

include/picongpu/particles/shapes/ShapeSelector.hpp
```diff
@@ -0,0 +1,106 @@
...
+             * @return Assignment value based on the particle shape.
+             */
+            HDINLINE float_X operator()(int const gridPointOffset) const
+            {
+                return shapeArray[gridPointOffset - T_ParticleAssignFunctor::begin];
+            }
+
+            decltype(T_ParticleAssignFunctor().shapeArray(
+                alpaka::core::declval<float_X>(),
+                alpaka::core::declval<bool>())) shapeArray;
```
sbastrakov: Are `declval` needed here?
Also i would prefer to have the type as a `using` alias first, to simplify this declaration

include/picongpu/algorithms/AssignedTrilinearInterpolation.hpp
psychocoderHPC: should be the dimensionality of the field

include/picongpu/algorithms/FieldToParticleInterpolation.hpp
psychocoderHPC: IN general, it is depending on the field.

include/picongpu/particles/shapes/TSC.hpp
```diff
@@ -128,6 +182,17 @@ namespace picongpu
...
                         return result;
                     }
+
+                    /** Creates an array with assignment values.
+                     *
+                     * @param pos particle position relative to the assignment cell range [-0.5;0.5)
+                     * @param isOutOfRange input will be ignored because the particle shape is always on support.
+                     * @return Array with precomputed assignment values.
+                     */
+                    HDINLINE auto shapeArray(float_X const x, [[maybe_unused]] bool const isOutOfRange) const
```
sbastrakov, line 193: This is logically consistent only with `isOutOfRange = false`. So we should at least add this to the comment, and potneetially also as an assert

include/picongpu/particles/shapes/PCS.hpp
psychocoderHPC: This is the magic trick of the optimization. It is required that we use the ternary operator to allow the compiler to track all memory access and avoid that lmem will be used for the array.

include/picongpu/particles/shapes/TSC.hpp
psychocoderHPC: added a comment

include/picongpu/particles/shapes/NGP.hpp
psychocoderHPC: I thought it is a shape that is not allowed to be used as particle shape.

psychocoderHPC:
can be merged

include/picongpu/algorithms/AssignedTrilinearInterpolation.hpp
```diff
@@ -52,64 +52,75 @@ namespace picongpu
         /** Does a 3D trilinear field-to-point interpolation for
          * arbitrary assignment function and arbitrary field_value types.
          *
-         * @tparam T_AssignmentFunction function for assignment
          * @tparam T_begin lower margin for interpolation
          * @tparam T_end upper margin for interpolation
+         * @tparam T_AssignmentFunction type of the array with shape functors
```
sbastrakov:
```suggestion
         * @tparam T_AssignmentFunction type of shape functors
```

include/picongpu/algorithms/AssignedTrilinearInterpolation.hpp
```diff
@@ -52,64 +52,75 @@ namespace picongpu
...
             PMACC_UNROLL(iterations)
             for(int z = T_begin; z <= T_end; ++z)
             {
-                auto result_y = type(0.0);
+                auto result_y = float_X(0.0);
                 PMACC_UNROLL(iterations)
                 for(int y = T_begin; y <= T_end; ++y)
                 {
-                    auto result_x = type(0.0);
+                    auto result_x = float_X(0.0);
```
sbastrakov: I think it is not right that we hardcode `float_X` here, as in principle these utilities are supposed to work for any PIConGPU field, not just `E` or `B`, and that could have any value type. And i think this is the reason we do not assume `d == simDim` here but work for both `d = 2` and `d = 3`. So imo we should keep the `type` alias for element type as before, just it was also not correctly assigned previously. It should be either `using type = decltype(*cursor(0, 0, 0));` to only accound for field only, or `using type = decltype(*cursor(x, y, z) * shapeFunctors[0](0));` to account for both field and shape function

include/picongpu/algorithms/AssignedTrilinearInterpolation.hpp
```diff
@@ -52,64 +52,75 @@ namespace picongpu
...
             }
             return result_z;
         }
 
         /** Implementation for 2D position*/
-        template<class T_AssignmentFunction, int T_begin, int T_end, class T_Cursor>
-        HDINLINE static auto interpolate(T_Cursor const& cursor, float2_X const& pos) ->
-            typename ::pmacc::result_of::Functor<AssignedTrilinearInterpolation, T_Cursor>::type
+        template<int T_begin, int T_end, class T_Cursor, class T_AssignmentFunction>
+        HDINLINE static auto interpolate(
```
sbastrakov, line 107: I would prefer if this function template is doxygen-grouped with the previous 3d case, so that a general comment is shared and only slight commends for 2d/3d are particular for these functions

include/picongpu/algorithms/FieldToParticleInterpolationNative.hpp
```diff
@@ -51,6 +52,23 @@ namespace picongpu
         using LowerMargin = typename pmacc::math::CT::make_Int<simDim, lowerMargin>::type;
         using UpperMargin = typename pmacc::math::CT::make_Int<simDim, upperMargin>::type;
 
+        HDINLINE auto getShapeArray(float3_X const& pos) const
```
sbastrakov, line 62: Please add doxygen as in another file

sbastrakov: And for the next function

psychocoderHPC:
@sbastrakov added your suggestions. thx for the review!

sbastrakov (approved):

PR #4128, opened 2022-05-25
Remove zlib from required dependencies in the docs

### LINKED ISSUE #4125: Remove zlib dependency from docs
We no longer actually depend on that for some time, so need to remove it from the docs. [Was just removed from spack](https://github.com/ComputationalRadiationPhysics/spack-repo/pull/19)

sbastrakov:
It is no longer required directly by PIConGPU and so the docs were outdated. Also update all the profiles to no longer use it.

Fixes #4125


Anton-Le:
Correct me if I'm wrong, but to my knowledge ADIOS2 (and HDF5?) require zlib as a dependency. So while it's not a direct dependency it is one that would be required for the library stack and the appropriate modules should not be removed from the profiles.

sbastrakov:
I think ADIOS2 doesn't require it.
However, regardless of that even ADIOS2 and HDF5 are *not* our dependencies. openPMD API is. So i think we should only mark that and refer to its installation settings, otherwise our instructions would constantly get outdated anyways with changes in openPMD API. @psychocoderHPC what do you think?

psychocoderHPC:
I agree with @sbastrakov zlib is a requirement of HDF5 and can be a requirement of ADIOS2 but both are indirect dependencies coming from openPMD-api, therefore, there is no need for us to document these dependencies.

sbastrakov:
But continuing this logic, we should then also remove `HDF5` and `ADIOS2` as they are dependencies of `openPMD API` which is our direct dependency?

docs/source/install/profile.rst
```diff
@@ -156,7 +156,7 @@ Piz Daint (CSCS)
 
 **Production directory:** ``$SCRATCH`` (`link <https://user.cscs.ch/storage/file_systems/>`_).
 
-For this profile to work, you need to download the :ref:`PIConGPU source code <install-dependencies-picongpu>` and install :ref:`boost, zlib, libpng, c-blosc, PNGwriter and ADIOS <install-dependencies>` manually.
+For this profile to work, you need to download the :ref:`PIConGPU source code <install-dependencies-picongpu>` and install :ref:`boost, libpng, c-blosc, PNGwriter and ADIOS <install-dependencies>` manually.
```
psychocoderHPC: we should remove `c-blosc` here and instead of `ADIOS` write `ADIOS2`

etc/picongpu/aris-grnet/gpu_picongpu.profile.example
```diff
@@ -32,7 +32,6 @@ module load picongpu
 #module load hdf5/1.8.17/gnu
 # Other Software ##############################################################
 #
-# module load zlib/1.2.8
```
psychocoderHPC: could you please keep in all profiles `zlib` on most systems the module system is not correctly loading the dependencies of HDF5, therefore we should keep loading zlib in all profiles.

psychocoderHPC (changes requested):
Could you please keep in all profiles zlib. It is a dependency of HDF5 and mostly not loaded automatically when HDF5 is loaded.

psychocoderHPC:
> But continuing this logic, we should then also remove `HDF5` and `ADIOS2` as they are dependencies of `openPMD API` which is our direct dependency?

I agree we should remove it because OpenPMD-api is documenting the dependencies well. We should check where the dependency tree png in the documentation is used and keep the full dependency graph or remove it, or update it and put PIConGPU onto the top.

Anton-Le:
The documentation for the installation of HDF5 and ADIOS2 should then be moved from the [PIConGPU docs](https://picongpu.readthedocs.io/en/latest/install/dependencies.html#hdf5)  to the [openPMD docs](https://openpmd-api.readthedocs.io/en/latest/dev/dependencies.html#development-dependencies), as there the dependencies are just listed, but without the helpful hints regarding compilation.

sbastrakov:
@Anton-Le I agree, and we even partially do it already in [this section](https://picongpu.readthedocs.io/en/latest/install/dependencies.html#openpmd-api). But we still have standalone sections for HDF5 and ADIOS2 anyways, which then should be removed as well.

sbastrakov:
@psychocoderHPC ready for re-review

psychocoderHPC (approved):

sbastrakov:
I will follow it up with removing ADIOS2 and HDF5 as direct dependencies (from the docs, as we no longer directly depend on them actually), as discussed above

Issue #4123, opened 2022-05-20
Do we need own Bessel function implementation?


sbastrakov:
We now have own implementation for real and complex numbers. With the complex implementation there are at least two technical issues:

- [the .tpp file is never included](https://github.com/ComputationalRadiationPhysics/picongpu/commit/80152dd47c6dcc4bee3f20b2aa8ab1807362630e#r73941544), so the PIConGPU build won't see it
- [this function](https://github.com/ComputationalRadiationPhysics/picongpu/blob/7bb619f3f21f965363bd4121b6d67f2a033eaa8f/include/pmacc/math/complex/Bessel.hpp#L75-L79) wrongly does overloading (thinking it was specialization perhaps?). Instead, the function should be removed alltogether and only the underlying structure template `J0<>` specialized for complex. Same for `j1()`.

These issues are currently "avoided" by never using own implementations, neither real nor complex. We do actually calculate Bessel functions in the synchrotron radiation stage. However, there we do it on host using `Boost.Math` implementation and not own one.

So we need to decide if we need to keep own implementation and if so for what reason. If kept, we should fix the Complex version and also try to use them from somewhere, to avoid the current situation of errorneous code present in the repo and this getting unnoticed as it's never used.

cc @BeyondEspresso @steindev @PrometheusPi

BeyondEspresso:
We need to keep the implementation, because the boost-library does only work on host, but not on device. The complex-valued bessel-functions are highly interesting for various laser implementations, which are not in the main code just yet. These complex-valued bessel functions are an infrastructure ( #2161), which I originally tested by writing a short PMACC-unit test for it.

BeyondEspresso:
The implementation is an external open source implementation (see code comments).

sbastrakov:
Okay, I take it as we need to keep it and fix it. I will provide a fix, since I believe i've already figured out how to do it.

sbastrakov:
The existing implementation is fixed with #4135 and we intend to keep it (and hopefully use it so that it doesn't become desynched again), closing the issue.

PR #4135, opened 2022-06-01
Fix implementation of complex Bessel functions

sbastrakov:
As pointed out in #4123 there were a few small issues with that, and turned out there were even a couple more. I think it happens simply because the code was never used in mainline PIConGPU and so it gradually got outdated and stopped compiling.

This PR fixes the current issues and now the funtions compile again, done in a commit-per-issue way. However, as we still never actually use them, they are again at risk of getting desynched.

Solves #4123

### DIFF
--- a/include/pmacc/math/complex/Bessel.hpp
+++ b/include/pmacc/math/complex/Bessel.hpp
@@ -59,7 +59,6 @@
 
 #pragma once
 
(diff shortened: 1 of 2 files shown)


psychocoderHPC (approved):

sbastrakov:

PR #4151, opened 2022-06-13
Bugfix JUWELS Booster profile

Anton-Le:
In my last update of the JUWELS Booster profile I neglected to update the versions of the dependencies (via path names).

Additionally added the default `Stages/2022` module for completenes' sake, as well as the _hidden_ `libpng/.1.6.37` that is not loaded by default but required for PNG writer.

**ToDo**

- [X] Update link to the installation script for the dependencies.


sbastrakov (approved):
Lgtm

Issue #4143, opened 2022-06-09
AverageAttribute assumes but does not check that the base attribute is weighted


sbastrakov:
In [its definition](https://github.com/ComputationalRadiationPhysics/picongpu/blob/74b6e552f7a436d517574b93c35c17830619994d/include/picongpu/particles/particleToGrid/combinedAttributes/AverageAttribute.def#L66-L68) it seems to always divide by density, which is basically a normalized sum of weightings. So then a nominator is assumed to be also a sum of weighted contributions of macroparticles. And most attributes are like that, so there it's no problem for most attributes.

But we have at least `MacroCounter` and `MomentumComponent<>` which are not weighted, so using them as arguments to average would be wrong. While this is unlikely to happen actually with these two, it's also possible an (advanced) user defines their custom attributes in `fileOutput.param` and tries using `Average` for them, probably to confusing results. For non-weighted i think we should just divide by a normalized `MacroCounter` rather then `Density` there, and that would be sufficient?

I am not sure what is the best way to check for it automatically, perhaps we could add openPMD-esque `macroWeighted` metadata to these types? I mean, not to the resulting fields themselves as they are, well, fields. But for these attributes so that averagizing operations could treat all attributes properly.

cc @pordyna if you have some comments or corrections on that. Btw I love this feature!

sbastrakov:
Discussed offline with @pordyna . Perhaps it's easiest to only allow averagizing over the macroparticle-weighted attributes (which are most) rather then define what average means for other ones. For that we still need some extra informatino of which attributes are and are not weighted, so probably a new trait is needed. I can implement it, and this is not most urgent task.

sbastrakov:
Fixed with #4154.

PR #4154, opened 2022-06-14
Check that averagized derived attributes are weighted

sbastrakov:
The logic of averagizing implementation relied on that, but it was not checked or documented. This PR adds a new trait for weighted attributes, specializes it for all existing attributes, and adds a check and documentation to `AverageAttribute`. With the trait, a user could also specialize it for their custom attributes if they need averagation for them. Having the trait `std::true_type` by default would have saved a few lines of code in this PR, but then averaziging would be on for user-provided attributes automatically, and this is unsafe (wrt getting weird or unexpected results).

Fixes https://github.com/ComputationalRadiationPhysics/picongpu/issues/4143.


psychocoderHPC (dismissed):

pordyna (dismissed):

sbastrakov:
@psychocoderHPC ready to be merged (review passed already, delay was due to the CI system)

psychocoderHPC (approved):

Issue #4148, opened 2022-06-10
Different default file extensions in plugins


sbastrakov:
E.g. the [openPMD plugin](https://github.com/ComputationalRadiationPhysics/picongpu/blob/f2ebdc20e51d93f2d75ad96c52304ff660f41112/include/picongpu/plugins/openPMD/openPMDWriter.hpp#L217-L224) uses ADIOS2 if it's available, and otherwise HDF5. While the [PhaseSpace plugin](https://github.com/ComputationalRadiationPhysics/picongpu/blob/f2ebdc20e51d93f2d75ad96c52304ff660f41112/include/picongpu/plugins/PhaseSpace/PhaseSpace.hpp#L100-L107) first tries HDF5 and only then ADIOS2.

Which means, by default different extensions will be used for openPMD API with both backends. In my opinion, we should have matching defaults, and this logic could be implemented once and shared between all such plugins. For exceptional cases like radiation only supporting HDF5 there can still remain the special handling.

sbastrakov:
Solved with #4155.

PR #4155, opened 2022-06-14
Unify default openPMD file extension in plugins

### LINKED ISSUE #3907: Phasespace plugin uses openPMD output
Hello,

I'm currently trying to get PIConGPU running on `taurus` and I ran into some trouble because my `LaserWakefield` examples with the default `cfg` always kept crashing with the error message 
```
Unhandled exception of type 'St13runtime_error' with message 'openPMD-api built without HDF5 support', terminating
Unhandled exception of type 'St13runtime_error' with message 'openPMD-api built without HDF5 support', terminating
Unhandled exception of type 'St13runtime_error' with message 'openPMD-api built without HDF5 support', terminating
Unhandled exception of type 'St13runtime_error' with message 'openPMD-api built without HDF5 support', terminating
[taurusi2053:28657:0:28657] Caught signal 11 (Segmentation fault: Sent by the kernel at address (nil))
```
**I have already found the cause for this crash:** The phasespace plugin uses openPMD HDF5 output as default output although I set the regular openPMD backend to adios2 in the cfg with `--openPMD.ext bp`. I just assumed that I had a corrupted openPMD installation (one has to manually install it) instead of checking for the dependencies from every single plugin. So in case you also get this error message, you should look for openPMD dependencies within activated plugins and change the backend respectively.

So the 3 reasons for this issue are:
* Make the problem google-able in case someone else has a similar crash on another system which doesn't have a prebuilt openPMD-api module
* Is it possible to have a clearer error message which shows the responsible plugin?
* Would it be smart to change the backend for openPMD output from **all** plugins who use openPMD **globally** for the simulation so that one doesn't have to choose it manually in case no hdf5 support is installed? Maybe all openPMD dependent plugins could check for the `--openPMD.ext <backend>` parameter and use this as their default if no plugin specific backend is chosen.

sbastrakov:
It affects which of ADIOS2 and HDF5 is choosen by default when both are available. We were not consistent there before, so with both backends on by default PIConGPU would write files in different formats, which may be confusing. Now I switch all to ADIOS2 when both available, like the openPMD and checkpoints plugin were already doing, but not calorimeter and phase space.

Fixes https://github.com/ComputationalRadiationPhysics/picongpu/issues/4148.

### DIFF
--- a/include/picongpu/plugins/PhaseSpace/PhaseSpace.hpp
+++ b/include/picongpu/plugins/PhaseSpace/PhaseSpace.hpp
@@ -24,6 +24,7 @@
 #include "picongpu/particles/traits/SpeciesEligibleForSolver.hpp"
 #include "picongpu/plugins/PhaseSpace/AxisDescription.hpp"
 #include "picongpu/plugins/PhaseSpace/PhaseSpaceFunctors.hpp"
+#include "picongpu/plugins/common/openPMDDefaultExtension.hpp"
 
 #include <pmacc/communication/manager_common.hpp>
 #include <pmacc/cuSTL/algorithm/kernel/Foreach.hpp>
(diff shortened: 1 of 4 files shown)


sbastrakov:
@franzpoeschel for the potential issue you mentioned during the meeting, I've now checked and it should be fine: phase space postprocessing is now general openPMD API (thanks to you actually), and calorimeter does not have a specific one. But thanks for bringing it up anyways, as i indeed didn't check for it originally.

psychocoderHPC (approved):

PR #4156, opened 2022-06-14
Clarify description of numerical dispersion for Yee solver

sbastrakov:
The previous version used $\xi_\mathrm{max}$ without properly defining
$\xi$, however that value is important for dispersion. Not so much for CFL that deals only with
$\xi_\mathrm{max}$. Also correct the misleading statement about 2d/3d relation, as become clear with having $\xi$ explicitly.

We may also want to clearly distinguish between theoretical $\omega$ and numerical
$\tilde \omega$ like we do for the wavevector. The finite-difference dispersion relations are for numerical $\tilde \omega$ but in their current form a reader may mistake them for theoretical due to our notation. However, i did not change it in this PR.

### DIFF
--- a/docs/source/models/AOFDTD.rst
+++ b/docs/source/models/AOFDTD.rst
@@ -239,16 +239,39 @@ The corresponding dispersion relation reads
 
 Obviously, this is a special case of the general dispersion relation, where :math:`M=1`.
 
-Solving for a wave's numerical frequency :math:`\omega` in dependence on its wave vector
+Solving for a wave's numerical frequency :math:`\omega` in dependence on its numerical wave vector
 :math:`\vec{\tilde k} = (\tilde k\cos\phi\sin\theta, \tilde k\sin\phi\sin\theta, \tilde k\cos\theta)` (spherical coordinates),
 
 .. math::
 
-   \omega = \frac{2}{\Delta t} \arcsin \xi\,\text{, where } \xi_\mathrm{max} = c\Delta t
-     \sqrt{ \frac{1}{\Delta x^2} + \frac{1}{\Delta y^2} + \frac{1}{\Delta z^2}}
+   \omega = \frac{2}{\Delta t} \arcsin \xi\,,
 
-reveals two important properties of the field solver.
-(The 2D version is obtained by letting :math:`\Delta z \rightarrow \infty`.)
+where
+
+.. math::
+
+   \xi = c\Delta t \sqrt{
+      \left[
+         \frac{1}{\Delta x} \sin\left(\frac{\tilde k_x \Delta x}{2}\right)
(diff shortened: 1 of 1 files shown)


docs/source/models/AOFDTD.rst
```diff
@@ -301,11 +324,37 @@ Solving the higher order dispersion relation for the angular frequency yields:
...
+         \frac{1}{\Delta x} \sin\left(\frac{\tilde k_x \Delta x}{2}\right)
+      \right]^2 +
+     \left[
+         \frac{1}{\Delta y} \sin\left(\frac{\tilde k_y \Delta y}{2}\right)
+     \right]^2 +
+     \left[
+         \frac{1}{\Delta z} \sin\left(\frac{\tilde k_z \Delta z}{2}\right)
+     \right]^2
+     }\,.
+
```
steindev, line 340: As far as I remember, the expression for $\xi$ is more complicated.

steindev:
```
.. math::
   \xi = c\Delta t \sqrt{ & 
       \left[\sum\limits_{l=1/2}^{M - 1/2} g_l^{2M} \frac{\sin(\tilde k_x l \Delta x)}{\Delta x} \right]^2
       &  + \left[\sum\limits_{l=1/2}^{M - 1/2} g_l^{2M} \frac{\sin(\tilde k_y l \Delta y)}{\Delta y} \right]^2
       & + \left[\sum\limits_{l=1/2}^{M - 1/2} g_l^{2M} \frac{\sin(\tilde k_z l \Delta z)}{\Delta z} \right]^2
   }\,.
```

steindev (changes requested):
In principle I like the clarification, but it is not correct yet.

sbastrakov:
Oh wtf happened to the PR description. It looked just fine yesterday and was not changed since, but renders incorrectly now.

Edit: i think today there is some rendering bug caused by having multiple inline TeX equations with `$` on same line. Now I manually distributed them to be on separate lines and it renders okay now. Though i am very sure it was fine as-was yesterday.

Edit2: the misrendering on github seems related to browser cache or smth, as first opening a page seems fine, but with refreshing it gets weird. In private mode it's fine all the time.

docs/source/models/AOFDTD.rst
sbastrakov: Thanks for correction, you are right. I will implement your formula, struggling with line-breaks and multipline as usual (as-is it does not work with rendering)

sbastrakov:
@steindev the formatting problem was with a multi-line `sqrt`. I now rewrote the formula to avoid it and it seems fine. Ready for re-review.

steindev (approved):

PR #4157, opened 2022-06-15
CI: Reduce parallelism and disable ISAAC for builds of SPEC

sbastrakov:
With the old configuration this job was consistently failing due to compiler running out of memory.


psychocoderHPC (dismissed):

psychocoderHPC:
I set this PR to draft. The PR should not be required anymore, the issue was a configurarion issue in the CI (virtual computeninstances). After discussing the issue with the admins the issue should be fixed.

sbastrakov:
The issue turned out to be caused by configuration of the CI machine. So the changes in this PR are not needed.

PR #4170, opened 2022-06-17
Use cached shapes for current deposition

psychocoderHPC:
- extent current deposition strategies with the possibility to choose the used assignment shape function (Caching or Jit)

This PR is reducing the compute intensity of the kernel ComputeCurrent strongly. This affects a slightly better performance (up to 10% for typical SPEC benchmark sizes)

This is a follow-up of #4124 where Chached shapes were introduced.


psychocoderHPC:
Note: The documentation string in strategies is saying that `i` is the slow-moving index, this is currently not correct for the 2D implementations of EZ and Esirkepov. I will update both implementations as soon as this PR is merged.

include/picongpu/fields/currentDeposition/EmZ/DepositCurrent.hpp
```diff
@@ -199,19 +220,27 @@ namespace picongpu
...
                         for(int i = T_begin; i < T_end; ++i)
                         {
-                            const float_X s0i = this->S0(line, i, 0);
-                            const float_X dsi = this->S1(line, i, 0) - s0i;
-                            float_X W = s0i * this->S0(line, j, 1) + 0.5_X * (dsi * s0j + s0i * dsj)
-                                + (1.0_X / 3.0_X) * dsi * dsj;
+                            const float_X s0i = shape_i.S0(i);
+                            const float_X dsi = shape_i.S1(i) - s0i;
+                            float_X W
+                                = s0i * shape_j.S0(j) + 0.5_X * (dsi * s0j + s0i * dsj) + (1.0_X / 3.0_X) * dsi * dsj;
```
sbastrakov: Why do we have `shape_j.S0(j)` and not use previously-calculated `s0j` here?

psychocoderHPC: Good catch, this should be `s0j`. This was already an extra function call in the old code. 
I will update this peace of code.

include/picongpu/fields/currentDeposition/EmZ/DepositCurrent.hpp
```diff
@@ -35,8 +35,8 @@ namespace picongpu
         {
             using namespace pmacc;
 
-            template<typename T_AtomicAddOp, typename ParticleAssign, int T_begin, int T_end>
-            struct DepositCurrent<T_AtomicAddOp, ParticleAssign, T_begin, T_end, DIM3> : public Base<ParticleAssign>
+            template<typename T_Strategy, typename ParticleAssign, int T_begin, int T_end>
+            struct DepositCurrent<T_Strategy, ParticleAssign, T_begin, T_end, DIM3>
```
sbastrakov: I believe you forgot to change it in `EmZ.def` where this template is declared.

There, also docstring for it is missing (unrelated to this change)

include/picongpu/fields/currentDeposition/Esirkepov/TrajectoryAssignmentShapeFunction.hpp
```diff
@@ -0,0 +1,110 @@
...
+         * assignment value for the given grid point.
+         * @tparam T_StartAssignmentFunction type of shape functor of for the particle at the start position
+         * @tparam T_EndAssignmentFunction type of shape functor of for the particle at the end position
+         *
+         * @param shapeStartParticle Shape functor for the start particle.
+         * @param shapeEndParticle Shape functor for the start particle.
+         * @return Instance of the functor required for the Esirkepov like current deposition.
+         */
+        template<typename T_StartAssignmentFunction, typename T_EndAssignmentFunction>
+        HDINLINE auto make_TrajectoryAssignmentShapeFunction(
```
sbastrakov: Why is it called unconventionally with `_` and not the normal camelCase?

include/picongpu/fields/currentDeposition/EmZ/DepositCurrent.hpp
```diff
@@ -35,8 +35,8 @@ namespace picongpu
         {
             using namespace pmacc;
 
-            template<typename T_AtomicAddOp, typename ParticleAssign, int T_begin, int T_end>
-            struct DepositCurrent<T_AtomicAddOp, ParticleAssign, T_begin, T_end, DIM3> : public Base<ParticleAssign>
+            template<typename T_Strategy, typename ParticleAssign, int T_begin, int T_end>
```
sbastrakov: Unrelated to this PR, possibly for a follow-up you are planning anyways: `ParticleAssign` should be `T_ParticleAssign`

include/picongpu/fields/currentDeposition/Esirkepov/TrajectoryAssignmentShapeFunction.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include "picongpu/fields/currentDeposition/Esirkepov/Line.hpp"
+
+#include <cstdint>
+
+
+namespace picongpu
+{
+    namespace currentSolver
+    {
+        /** Helper for Esirkepov and Esirkepov-like current deposition implementation
```
sbastrakov, line 31: I think this file should be located one level above, as it is used by both `Esirkepov` and `EmZ`. Its namespace also implies that

include/picongpu/fields/currentDeposition/EmZ/DepositCurrent.hpp
```diff
@@ -80,18 +80,31 @@ namespace picongpu
                 {
                     if(line.m_pos0[2] == line.m_pos1[2])
                         return;
+
+                    auto const shape_i = make_TrajectoryAssignmentShapeFunction(
```
sbastrakov: I think here `x/y/z` would be better name than `i/j/k` as this is actually a function that describes that shape. And later on you call it with `i/j/k` to get the actual value for those indices.

If you decide to change, the same applies to naming of types inside strategy.

Also, why not use our normal camelCase?

include/picongpu/fields/currentDeposition/EmZ/DepositCurrent.hpp
```diff
@@ -80,18 +80,31 @@ namespace picongpu
...
+                        typename T_Strategy::template Shape_i<ParticleAssign>(line.m_pos0[0], true),
+                        typename T_Strategy::template Shape_i<ParticleAssign>(line.m_pos1[0], true));
+
+                    auto const shape_j = make_TrajectoryAssignmentShapeFunction(
+                        typename T_Strategy::template Shape_j<ParticleAssign>(line.m_pos0[1], true),
+                        typename T_Strategy::template Shape_j<ParticleAssign>(line.m_pos1[1], true));
```
sbastrakov: This looks very hard to read, and also repetitive. I think it can be wrapped in some helper function to better convay what is happening and also avoid repeating an almost similar line 6 times. Or at least please use `{}` for initialization to distinguish from a function call `()`, as here it is really easy to get confused with that `::template`

include/picongpu/fields/currentDeposition/EmZ/DepositCurrent.hpp
```diff
@@ -199,19 +220,26 @@ namespace picongpu
...
+                        const float_X dsj = shape_j.S1(j) - s0j;
                         for(int i = T_begin; i < T_end; ++i)
                         {
-                            const float_X s0i = this->S0(line, i, 0);
-                            const float_X dsi = this->S1(line, i, 0) - s0i;
-                            float_X W = s0i * this->S0(line, j, 1) + 0.5_X * (dsi * s0j + s0i * dsj)
-                                + (1.0_X / 3.0_X) * dsi * dsj;
+                            const float_X s0i = shape_i.S0(i);
+                            const float_X dsi = shape_i.S1(i) - s0i;
+                            float_X W = s0i * s0j + 0.5_X * (dsi * s0j + s0i * dsj) + (1.0_X / 3.0_X) * dsi * dsj;
```
sbastrakov: Unrelated to this PR, but in the spirit of separating this implementation from shape calculation details, it should use `auto` for these variables.

include/picongpu/fields/currentDeposition/Esirkepov/BitPacking.hpp
```diff
@@ -0,0 +1,76 @@
...
+        struct Bitpacking
+        {
+            enum class Status
+            {
+                //! Marker that a particle trajectory is crossing the assignment cell border.
+                leaveCell = 1,
+                //! Marker that the start particle is located in the base assignment cell.
+                startParticleInAssignmentCell = 2,
+                //! Marker that the end particle is located in the base assignment cell.
+                endParticleInAssignmentCell = 4
```
sbastrakov: I think we normally capitalize those? (I have no preference, just to ensure uniform style)

include/picongpu/fields/currentDeposition/Esirkepov/BitPacking.hpp
```diff
@@ -0,0 +1,76 @@
...
+    {
+        /** Helper base class for Esirkepov to manage bit packing operations to encode the status of the start and end
+         * particle to optimize the algorithm.
+         * Bit-packing for the status is required to keep the register footprint low.
+         */
+        struct Bitpacking
+        {
+            enum class Status
+            {
+                //! Marker that a particle trajectory is crossing the assignment cell border.
```
sbastrakov: `Marker that` can be dropped in those comments

include/picongpu/fields/currentDeposition/Esirkepov/BitPacking.hpp
```diff
@@ -0,0 +1,76 @@
...
+                leaveCell = 1,
+                //! Marker that the start particle is located in the base assignment cell.
+                startParticleInAssignmentCell = 2,
+                //! Marker that the end particle is located in the base assignment cell.
+                endParticleInAssignmentCell = 4
+            };
+
+            DINLINE void setStatus(int& packedValue, Status const status, bool condition) const
```
sbastrakov: Doxygen comment missing

include/picongpu/fields/currentDeposition/Esirkepov/BitPacking.hpp
```diff
@@ -0,0 +1,76 @@
...
+                packedValue |= condition ? static_cast<int>(status) : 0;
+            }
+
+            /*! Query the status for a given type
+             *
+             * @param packedValue status bit mask
+             * @param status status type
+             * @return true if status is set else false
+             */
+            DINLINE bool getStatus(int packedValue, Status status) const
```
sbastrakov: I think all 3 functions should be `static` as the class is stateless?

I think `Bitpacking` would be better a namespace rather than a class, but that is a matter of taste and I agree we often use classes in that situations. But then with `static` methods.

include/picongpu/fields/currentDeposition/Esirkepov/BitPacking.hpp
```diff
@@ -0,0 +1,76 @@
...
+                //! Marker that the end particle is located in the base assignment cell.
+                endParticleInAssignmentCell = 4
+            };
+
+            DINLINE void setStatus(int& packedValue, Status const status, bool condition) const
+            {
+                packedValue |= condition ? static_cast<int>(status) : 0;
+            }
+
+            /*! Query the status for a given type
```
sbastrakov: I think this and function name do not reflect its purpose well. The function seems to return if the given status is set?

Same for `getStatusValue()`

include/picongpu/fields/currentDeposition/Esirkepov/Esirkepov.hpp
```diff
@@ -38,7 +39,7 @@ namespace picongpu
     namespace currentSolver
     {
         template<typename T_ParticleShape, typename T_Strategy>
-        struct Esirkepov<T_ParticleShape, T_Strategy, DIM3> : public Base<typename T_ParticleShape::ChargeAssignment>
+        struct Esirkepov<T_ParticleShape, T_Strategy, DIM3> : private Bitpacking
```
sbastrakov: I think regardless of class/namespace for `Bitpacking` decision above, it should not be privately inherited here as we do not want to express what private inheritance normally means, neither is it really needed here.

include/picongpu/fields/currentDeposition/EmZ/DepositCurrent.hpp
```diff
@@ -199,19 +220,26 @@ namespace picongpu
...
 
                             const float_X j_z = W * currentSurfaceDensityZ;
-                            auto const atomicOp = T_AtomicAddOp{};
+                            auto const atomicOp = typename T_Strategy::BlockReductionOp{};
```
sbastrakov, line 240: More or less same style notes as for this file also apply to `Esirkepov`

include/picongpu/fields/currentDeposition/Esirkepov/TrajectoryAssignmentShapeFunction.hpp
```diff
@@ -0,0 +1,110 @@
...
+            DINLINE float_X S0(int const gridPoint) const
+            {
+                return m_shapeStartParticle(gridPoint);
+            }
+
+            //! Calculate term in S1 for end position
+            DINLINE float_X S1(int const gridPoint) const
+            {
+                return m_shapeEndParticle(gridPoint);
+            }
```
sbastrakov, line 78: I think it's okay now, but could `m_shapeStartParticle` just have been called `S0`, and the other one `S1` ?

I guess you went for a more telling name for the member, but then why does the interface still have those Esirkepov-paper-spefic names `S0`, `S1`

include/pmacc/eventSystem/tasks/TaskReceiveMPI.hpp
```diff
@@ -47,32 +47,58 @@ namespace pmacc
...
-
-            this->request = Environment<DIM>::get().EnvironmentController().getCommunicator().startReceive(
-                exchange->getExchangeType(),
-                reinterpret_cast<char*>(dst->getPointer()),
-                dst->getDataSpace().productOfComponents() * sizeof(TYPE),
-                exchange->getCommunicationTag());
+            if(Environment<>::get().isMpiDirectEnabled())
+            {
+                /* Wait to be sure that all device work is finished before MPI is triggered.
+                 * MPI will not wait for work in our device streams
```
sbastrakov: Are these changes related to this PR?

sbastrakov (commented):
I have nothing against the substance of this PR, it's great.

However, the style I feel could use some improvements, as suggested in the review. This part was already very hard to understand and with the PR as-is becomes even worse. So i think it's reasonable to invest a little bit of time to improve readability so that there is bus factor > 1 here

include/picongpu/fields/currentDeposition/Esirkepov/TrajectoryAssignmentShapeFunction.hpp
psychocoderHPC: Offline discussed: This file is part of Esirkepov because I see EmZ as an implementation of Esirkepov where parts can be reused. Same as the old `class Base` before. I do not like to move this file one level up because VillaBune is not using this file.

include/picongpu/fields/currentDeposition/EmZ/DepositCurrent.hpp
psychocoderHPC: I will keep `i/j/k` because the loops do not reflect or naming of the directions `x/y/z`

psychocoderHPC: I will keep `Shape_i`, `Sape_j` with `-` for now because using camel case  `ShapeI`, ShapeJ` is depending on the font not readable.
Is the last letter a capital `I` or a small `l`

include/picongpu/fields/currentDeposition/EmZ/DepositCurrent.hpp
psychocoderHPC: `0.5_X` is enforcing float_X in many cases anyway so I would like to keep it explicit.

include/picongpu/fields/currentDeposition/Esirkepov/BitPacking.hpp
psychocoderHPC: I changed it now to `getValue`. I do not know a better name for it. It is used only in one place to avoid a cast from `int`->`bool`->`int`.

Will add a comment to the doc string

include/picongpu/fields/currentDeposition/Esirkepov/TrajectoryAssignmentShapeFunction.hpp
psychocoderHPC: `m_shapeStartParticle` is now private. I do not like to mix access to a member and method access, therefore I kept the methods `S1` and `S0`

include/pmacc/eventSystem/tasks/TaskReceiveMPI.hpp
psychocoderHPC: no, I removed these changes.

include/picongpu/fields/currentDeposition/Esirkepov/BitPacking.hpp
psychocoderHPC: I am now using a namespace

psychocoderHPC:
I updated the code, ready for review

include/picongpu/fields/currentDeposition/EmZ/DepositCurrent.hpp
sbastrakov: The change is now put to `EmZ.def`, so it's okay now.
But would still be nice to also add a doxygen comment there.

include/picongpu/fields/currentDeposition/Strategy.def
```diff
@@ -47,6 +49,18 @@ namespace picongpu
...
 #endif
                 }
+
+                struct ShapeStrategy
+                {
+                    //! ShapeI is used for the outer loop (slow moving index) of the current deposition
+                    template<typename T_AssignmentShape>
+                    using ShapeI = shapes::Jit<T_AssignmentShape>;
+                    template<typename T_AssignmentShape>
+                    using ShapeJ = shapes::Cached<T_AssignmentShape>;
```
sbastrakov: Doxygen comment missing

include/picongpu/fields/currentDeposition/Strategy.def
```diff
@@ -62,7 +76,7 @@ namespace picongpu
              *
              * @{
              */
-            struct StridedCachedSupercells
+            struct StridedCachedSupercells : detail::ShapeStrategy
```
sbastrakov, line 83: Question. Currently all these strategies use the same `detail::ShapeStrategy`. Is it planned that in principle that can be different and depending on the strategy type?

include/picongpu/fields/currentDeposition/EmZ/DepositCurrent.hpp
sbastrakov: Better to discuss offline, but to float an idea. As this part is really connected to the internal loop structure and performance optimization, perhaps it makes sense to reflect it in the name. E.g. `shapeOuterLoop`, then a reader would less likely be thinking in the usual `i = x, j = y, k = z` way. You already use a similar notation in doxygen comment for `ShapeStrategy`, so my suggestion is to more or less extend it for variable/type names.

include/picongpu/fields/currentDeposition/Esirkepov/bitPacking.hpp
```diff
@@ -0,0 +1,90 @@
...
+    {
+        /** Helper base class for Esirkepov to manage bit packing operations to encode the status of the start and end
+         * particle to optimize the algorithm.
+         * Bit-packing for the status is required to keep the register footprint low.
+         */
+        namespace bitpacking
+        {
+            enum class Status
+            {
+                /** Marke that a particle trajectory is crossing the assignment cell border.
```
sbastrakov: I guess you decided to keep `Marker that`, but it is fine. But now there is a typo `Marke` which was not in the earlier version of this PR

include/picongpu/fields/currentDeposition/Esirkepov/bitPacking.hpp
```diff
@@ -0,0 +1,90 @@
...
+                 *
+                 * For performance reason it is required that LEAVE_CELL is represented by the first bit. This allow
+                 * that we can use getValue() to querry the integer value. It is used to optimize the loop contition in
+                 * the Esirkepov implemenation.
+                 */
```
sbastrakov, line 44: I think this doxygen comment should be moved a bit up, to be in front of `enum class`

include/picongpu/fields/currentDeposition/Esirkepov/bitPacking.hpp
```diff
@@ -0,0 +1,90 @@
...
+                LEAVE_CELL = 1,
+                //! Marke that the start particle is located in the base assignment cell.
+                START_PARTICLE_IN_ASSIGNMENT_CELL = 2,
+                //! Marke that the end particle is located in the base assignment cell.
```
sbastrakov: Also typos `Marke`

include/picongpu/fields/currentDeposition/Esirkepov/bitPacking.hpp
```diff
@@ -0,0 +1,90 @@
...
+                return (packedValue & static_cast<int>(status));
+            }
+
+            /*! Query the status for a given type
+             *
+             * @param packedValue status bit mask
+             * @param status status type
+             * @return true if status is set else false
+             */
+            DINLINE bool get(int packedValue, Status status)
```
sbastrakov: Perhaps `check` or `test` would be a more fitting name, but can be kept as-is

include/picongpu/fields/currentDeposition/Esirkepov/BitPacking.hpp
sbastrakov: Oh i meant not all caps, but e.g. `LeaveCell`. But since we do not have a fixed style, i'm fine with anything.

include/picongpu/fields/currentDeposition/Esirkepov/Esirkepov.hpp
```diff
@@ -97,8 +94,16 @@ namespace picongpu
...
                     RelayPoint<isSupportEven>()(iStart, iEnd, line.m_pos0[d], line.m_pos1[d]);
                     gridShift[d] = iStart < iEnd ? iStart : iEnd; // integer min function
+                    bitpacking::set(
+                        status[d],
+                        bitpacking::Status::START_PARTICLE_IN_ASSIGNMENT_CELL,
+                        gridShift[d] == iStart);
+                    bitpacking::set(
+                        status[d],
+                        bitpacking::Status::END_PARTICLE_IN_ASSIGNMENT_CELL,
+                        gridShift[d] == iStart);
```
sbastrakov: Self-reminder to discuss offline

include/picongpu/fields/currentDeposition/Esirkepov/Esirkepov.hpp
```diff
@@ -114,23 +119,23 @@ namespace picongpu
...
                     cellSize.y());
-                cptCurrent1D(acc, leaveCell, cursorJ, line, cellSize.z());
+                cptCurrent1D(acc, status, cursorJ, line, cellSize.z());
             }
 
             /**
              * deposites current in z-direction
              *
-             * @param leaveCell vector with information if the particle is leaving the cell
+             * @param LEAVE_CELL vector with information if the particle is leaving the cell
```
sbastrakov: Should be `parStatus` instead of `LEAVE_CELL`

include/picongpu/fields/currentDeposition/Esirkepov/Esirkepov.hpp
```diff
@@ -114,23 +119,23 @@ namespace picongpu
...
+                    DataSpace<simDim>(status.z(), status.x(), status.y()),
                     twistVectorFieldAxes<pmacc::math::CT::Int<2, 0, 1>>(cursorJ),
                     rotateOrigin<2, 0, 1>(line),
                     cellSize.y());
-                cptCurrent1D(acc, leaveCell, cursorJ, line, cellSize.z());
+                cptCurrent1D(acc, status, cursorJ, line, cellSize.z());
             }
 
             /**
              * deposites current in z-direction
```
sbastrakov: I think it would be useful to extend this a little bit, that here x/y/z are not necessarily the same as in main PIConGPU cooridnate system, but it's after rotation

include/picongpu/fields/currentDeposition/Esirkepov/Esirkepov.hpp
```diff
@@ -167,15 +198,15 @@ namespace picongpu
                  *     ( this helps the compiler to mask threads without work )
                  */
                 for(int i = begin; i < end + 1; ++i)
-                    if(i < end + leaveCell[0])
+                    if(i < end + bitpacking::getValue(parStatus[0], bitpacking::Status::LEAVE_CELL))
```
sbastrakov: Does it make sense to save `bitpacking::getValue(parStatus[0], bitpacking::Status::LEAVE_CELL)` to a variable since it does not change? (then could also be used earlier to construct variables like  `shapeI`)

include/picongpu/fields/currentDeposition/Esirkepov/Esirkepov2D.hpp
```diff
@@ -90,7 +93,7 @@ namespace picongpu
                  * 0 == stay in cell
                  * 1 == leave cell
                  */
-                DataSpace<DIM2> leaveCell;
+                DataSpace<DIM2> status;
```
sbastrakov: Forgot to update the comment (was updated in 3d version)

include/picongpu/fields/currentDeposition/Esirkepov/bitPacking.hpp
```diff
@@ -0,0 +1,90 @@
...
+{
+    namespace currentSolver
+    {
+        /** Helper base class for Esirkepov to manage bit packing operations to encode the status of the start and end
+         * particle to optimize the algorithm.
+         * Bit-packing for the status is required to keep the register footprint low.
+         */
+        namespace bitpacking
+        {
+            enum class Status
```
sbastrakov, line 37: I didn't think of that before, but perhaps instead of `bitpacking::Status` it could simply be `enum class ParticleStatus` without a namespace. So then the usage is `currentSolver::ParticleStatus` which makes sense.

"bitpacking" is imo too much into details of implementation rather then what it represents

include/picongpu/fields/currentDeposition/Strategy.def
psychocoderHPC: offline discussed: We will keep it to simplify how these configurations get propagated to the kernel. Strategies for now more for the expert (mostly me^^)

include/picongpu/fields/currentDeposition/Esirkepov/bitPacking.hpp
psychocoderHPC: Should be `Mark`

include/picongpu/fields/currentDeposition/Esirkepov/Esirkepov.hpp
psychocoderHPC: condition should be `gridShift[d] == iEnd)`, I will double check this part

include/picongpu/fields/currentDeposition/Esirkepov/bitPacking.hpp
psychocoderHPC: I want to keep the namespace separate from other current deposition code parts.

sbastrakov (dismissed):
I have no further suggestions to this PR. Just CI failed for some legitimate reason now, and after this is fixed we can merge

sbastrakov (approved):

include/picongpu/fields/currentDeposition/Esirkepov/Esirkepov2D.hpp
```diff
@@ -149,17 +155,33 @@ namespace picongpu
...
                     = this->charge * (1.0_X / float_X(CELL_VOLUME * DELTA_T)) * cellEdgeLength;
 
                 for(int j = begin; j < end + 1; ++j)
-                    if(j < end + leaveCell[1])
+                    if(j < end + bitpacking::getValue(parStatus[1], bitpacking::Status::LEAVE_CELL))
                     {
-                        const float_X s0j = this->S0(line, j, 1);
-                        const float_X dsj = this->S1(line, j, 1) - s0j;
+                        const float_X s0j = shapeJ.S0(j);
+                        const float_X dsj = shapeJ.S0(j) - s0j;
```
psychocoderHPC, line 184: **BUG**: must be shapeJ.S1(j) - s0j;

PR #4173, opened 2022-06-21
add device side `ForEachParticle`

psychocoderHPC:
Add device-side `ForEachParticle` algorithm.

- add a policy to describe if particle frames should be forward or backward processed
- add factory functions to create the particle forEach executor.
- The particle for each is providing two modes
  - iterate over all particles in a supercell
  - iterate over all frames -> required mostly for pmacc internal algorithms to compress the linked frame data structures.

Depending on the number of worker and the frame size the forach algorithm will work with skip lists to utilize all threads.

### DIFF
--- a/include/picongpu/particles/debyeLength/Estimate.kernel
+++ b/include/picongpu/particles/debyeLength/Estimate.kernel
@@ -24,6 +24,7 @@
 #include "picongpu/traits/frame/GetMass.hpp"
 
 #include <pmacc/lockstep.hpp>
+#include <pmacc/particles/algorithm/ForEach.hpp>
 
 #include <cstdint>
 
@@ -108,7 +109,6 @@ namespace picongpu
                     pmacc::DataSpace<simDim> const supercellIdx,
                     T_EstimateBox estimateBox) const
                 {
-                    constexpr uint32_t frameSize = pmacc::math::CT::volume<SuperCellSize>::type::value;
                     constexpr uint32_t numWorkers = T_numWorkers;
                     /* Estimate variance in momentum of macroparticles in the supercell.
                      * Each thread processes its particles, then the master thread performs reduction.
@@ -118,28 +118,24 @@ namespace picongpu
                     auto sumWeighting = float_64{0.0};
                     auto sumMomentum = float3_64::create(0.0);
                     auto sumMomentumSquared = float3_64::create(0.0);
-                    auto frame = electronBox.getLastFrame(supercellIdx);
-                    auto particlesInFrame = electronBox.getSuperCell(supercellIdx).getSizeLastFrame();
+
                     uint32_t const workerIdx = cupla::threadIdx(acc).x;
-                    while(frame.isValid())
-                    {
-                        lockstep::makeForEach<frameSize, numWorkers>(workerIdx)(
-                            [&](uint32_t const linearIdx)
-                            {
-                                if(linearIdx < particlesInFrame)
-                                {
-                                    auto const particle = frame[linearIdx];
-                                    auto const weighting = float_64{particle[weighting_]};
-                                    sumWeighting += weighting;
-                                    auto const singleParticleMomentum
-                                        = precisionCast<float_64>(particle[momentum_]) / weighting;
-                                    sumMomentum += weighting * singleParticleMomentum;
-                                    sumMomentumSquared += weighting * singleParticleMomentum * singleParticleMomentum;
-                                }
-                            });
-                        frame = electronBox.getPreviousFrame(frame);
-                        particlesInFrame = frameSize;
-                    }
+                    auto forEachParticle = pmacc::particles::algorithm::acc::makeForEach<numWorkers>(
+                        workerIdx,
+                        electronBox,
+                        supercellIdx);
+
+                    forEachParticle(
+                        acc,
+                        [&sumWeighting, &sumMomentum, &sumMomentumSquared](auto const& /*accelerator*/, auto& particle)
+                        {
+                            auto const weighting = float_64{particle[weighting_]};
+                            sumWeighting += weighting;
+                            auto const singleParticleMomentum
+                                = precisionCast<float_64>(particle[momentum_]) / weighting;
+                            sumMomentum += weighting * singleParticleMomentum;
+                            sumMomentumSquared += weighting * singleParticleMomentum * singleParticleMomentum;
+                        });
 
                     // Reduce for supercell in shared memory
                     PMACC_SMEM(acc, supercellSumWeighting, float_64);
--- a/include/pmacc/particles/algorithm/ForEach.hpp
+++ b/include/pmacc/particles/algorithm/ForEach.hpp
@@ -22,157 +22,445 @@
 #pragma once
 
 #include "pmacc/Environment.hpp"
+#include "pmacc/attribute/Constexpr.hpp"
 #include "pmacc/lockstep.hpp"
+#include "pmacc/lockstep/Config.hpp"
+#include "pmacc/lockstep/ForEach.hpp"
 #include "pmacc/mappings/kernel/AreaMapping.hpp"
+#include "pmacc/particles/algorithm/detail/ForEach.hpp"
 #include "pmacc/particles/frame_types.hpp"
 
 #include <cstdint>
+#include <type_traits>
 #include <utility>
 
 
-namespace pmacc
+namespace pmacc::particles::algorithm
 {
-    namespace particles
+    namespace acc
     {
-        namespace algorithm
+        //! Policy to forward iterate over particles/frames of a supercell.
+        struct Forward
         {
-            namespace acc
+            template<typename T_ParBox>
+            static DINLINE auto getFirst(T_ParBox& pb, DataSpace<T_ParBox::Dim> const& superCellIdx)
             {
-                namespace detail
+                return pb.getFirstFrame(superCellIdx);
+            }
+
+            template<typename T_ParBox>
+            static DINLINE auto next(T_ParBox& pb, typename T_ParBox::FramePtr const& framePtr)
+            {
+                return pb.getNextFrame(framePtr);
+            }
+
+            template<typename T_ForEach>
+            static DINLINE auto createCtx(T_ForEach& forEachParticleInFrame)
+            {
+                return lockstep::makeVar<uint32_t>(forEachParticleInFrame, 0u);
+            }
+        };
+
+        //! Policy to reverse iterate over particles/frames of a supercell.
+        struct Reverse
+        {
+            template<typename T_ParBox>
+            static DINLINE auto getFirst(T_ParBox& pb, DataSpace<T_ParBox::Dim> const& superCellIdx)
+            {
+                return pb.getLastFrame(superCellIdx);
+            }
+
+            template<typename T_ParBox>
+            static DINLINE auto next(T_ParBox& pb, typename T_ParBox::FramePtr const& framePtr)
+            {
+                return pb.getPreviousFrame(framePtr);
+            }
+
+            template<typename T_ForEach>
+            static DINLINE auto createCtx(T_ForEach& forEachParticleInFrame)
+            {
+                return lockstep::makeVar<lcellId_t>(forEachParticleInFrame, lcellId_t(0u));
+            }
+        };
+
+        /** Execute a particle or frame functor for each particle/frame.
+         *
+         * @tparam T_AccessTag Type of the tag to mark if the particle or frame interface will be used.
+         *                     valid options: detail::CallParticleFunctor or detail::CallFrameFunctor
+         * @tparam T_Order Iteration order. valid options: Forward or Reverse
+         * @tparam T_ParBox Type of the particle box to traverse.
+         * @tparam T_numWorkers Number of lockstep workers.
+         */
+        template<typename T_AccessTag, typename T_Order, typename T_ParBox, uint32_t T_numWorkers>
+        class ForEachParticle
+        {
+        private:
+            static constexpr uint32_t dim = T_ParBox::Dim;
+
+            using SuperCellSize = typename T_ParBox::FrameType::SuperCellSize;
+            static constexpr uint32_t frameSize = pmacc::math::CT::volume<SuperCellSize>::type::value;
+            static constexpr uint32_t numWorkers = T_numWorkers;
+
+            /** Number of frames to skip.
+             *
+             * If we have more workers than particles in a frame we skip frames depending on the worker id.
+             */
+            static constexpr uint32_t frameDistance = (numWorkers + frameSize - 1u) / frameSize;
+
+            using ForEachParticleInFrame
+                = lockstep::ForEach<lockstep::Config<frameSize * frameDistance, T_numWorkers, 1>>;
+
+            DataSpace<dim> const m_superCellIdx;
+            ForEachParticleInFrame const forEachParticleInFrame;
+            T_ParBox m_particlesBox;
+
+        public:
+            /** Provide the lockstep foreach executor to execute a functor for each particle within a set of frames.
+             *
+             * Depending of the number of workers used it could be that more than one frame is executed by all workers
+             * in parallel. The frame functor interface is therefore a context with one frame per virtual worker.
+             * It is guaranteed that the executor fits to the frameCtx provided by the frame functor interface. @see
+             * detail::FrameFunctorInterface
+             *
+             * @return Lockstep foreach executor to operate on frames.
+             */
+            DINLINE auto lockstepForEach() const
+            {
+                return forEachParticleInFrame;
+            }
+
+            /** Construct algoritm to operate on particles in a superCell
+             *
+             * @param workerIdx workerIdx index of the worker: range [0;workerSize)
+             * @param particlesBox particles memory
+             *                     It is not allowed concurrently to add or remove particles during the
+             *                     execution of this algorithm.
+             * @param superCellIdx index of the superCell where particles should be processed
+             */
+            DINLINE ForEachParticle(
+                uint32_t const workerIdx,
+                T_ParBox const& particlesBox,
+                DataSpace<dim> const& superCellIdx)
+                : m_superCellIdx(superCellIdx)
+                , forEachParticleInFrame(workerIdx)
+                , m_particlesBox(particlesBox)
+            {
+            }
+
+            /** Execute unary functor for each particle.
+             *
+             * @attention There is no guarantee in which order particles will be executed.
+             *            If the particle functor is used the algorithm assumes that the frame structure is following
+             *            the rule that all frames but the last are fully filled and in the last frame the low indices
+             *            are contiguously filled.
+             *
+             * @tparam T_Acc alpaka accelerator type
+             * @tparam T_Functor unary particle functor or frame functor following the interface concept
+             *                           detail::FrameFunctorInterface or detail::ParticleFunctorInterface
+             * @param acc alpaka accelerator
+             * @param unaryFunctor Functor executed for each particle/frame.
+             *                     The caller must ensure that calling the functor in parallel with
+             *                     different workers is data race free.
+             *                     - particle functor: 'void operator()(T_Acc const &, ParticleType)'.
+             *                       It is not allowed to call a synchronization function within the functor.
+             *                     - frame functor: 'void operator()(T_Acc const &, FrameCtx)'.
+             *                       Calling synchronization within the functor is allowed.
+             */
+            template<typename T_Acc, typename T_Functor>
+            DINLINE void operator()(T_Acc const& acc, T_Functor&& unaryFunctor) const
+            {
+                if constexpr(numWorkers > frameSize)
                 {
-                    /** operate on particles of a species
-                     *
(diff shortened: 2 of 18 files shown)


include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+                return pb.getPreviousFrame(framePtr);
+            }
 
-                            // end kernel if we have no particles
-                            if(numPartcilesInSupercell == 0)
-                                return;
+            template<typename T_ForEach>
+            static DINLINE auto createCtx(T_ForEach& forEachParticleInFrame)
+            {
+                return lockstep::makeVar<lcellId_t>(forEachParticleInFrame, lcellId_t(0u));
```
sbastrakov, line 82: It is different from `Forward<>::createCtx()` on purpose? Here you use `lcellId_t`, and there `uint32_t`.

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+        template<typename T_AccessTag, typename T_Order, typename T_ParBox, uint32_t T_numWorkers>
+        struct ForEachParticle
+        {
+        private:
+            static constexpr uint32_t dim = T_ParBox::Dim;
 
-                            // offset of the superCell (in cells, without any guards) to the origin of the local domain
-                            DataSpace<dim> const localSuperCellOffset = superCellIdx - mapper.getGuardingSuperCells();
+            using SuperCellSize = typename T_ParBox::FrameType::SuperCellSize;
+            static constexpr uint32_t frameSize = pmacc::math::CT::volume<SuperCellSize>::type::value;
```
sbastrakov, line 101: I think this may be a good moment to separate frame size from supercell size, e.g. in a follow-up PR. Could start as simple as a function that always returns `pmacc::math::CT::volume<SuperCellSize>::type::value` that we use now (it's not just here, but basically everywhere in the code).

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+            static constexpr uint32_t numWorkers = T_numWorkers;
 
-                            auto accFunctor
-                                = functor(acc, localSuperCellOffset, lockstep::Worker<T_numWorkers>{workerIdx});
+            /** Number of frames to skip.
+             *
+             * If we have more worker than particles in a frame we skip frames depending on the worker id.
```
sbastrakov: Typo: should be `workers`

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+             */
+            static constexpr uint32_t numFramesToSkip = (numWorkers + frameSize - 1u) / frameSize;
```
sbastrakov: I think this is confusing - does this variable actually mean the number of frames to skip + 1?  E.g. for `numWorkers == frameSize` its value will be 1, while we should skip 0 frames in this case.

(Same for the twin of this variable inside the lambda)

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+        /** Execute a particle or frame functor for each particle/frame.
+         *
+         * @tparam T_AccessTag Type of the tag to mark if the particle or frame interface will be used.
+         *                     valid options: detail::CallParticleFunctor or detail::CallFrameFunctor
+         * @tparam T_Order Iteration order. valid options: Forward or Reverse
+         * @tparam T_ParBox Type of the particle box to traverse.
+         * @tparam T_numWorkers Number of lockstep workers.
+         */
+        template<typename T_AccessTag, typename T_Order, typename T_ParBox, uint32_t T_numWorkers>
+        struct ForEachParticle
```
sbastrakov: Matter of preference, but I would make this a class given it has a private state

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+            ForEachParticleInFrame const forEachParticleInFrame;
+            T_ParBox m_particlesBox;
 
-            /** Run a unary functor for each particle of a species in the given area
+        public:
+            /** Provide the lockstep foreach executor to execute a functor for each particle within a set of frames.
              *
-             * Has a version for a fixed area, and for a user-provided mapper factory.
-             * They differ only in how the area is defined.
+             * Depending of the number of worker used it could be that more than one frame executed by all workers in
```
sbastrakov: Typo: should be `workers`

sbastrakov: Another typo: instead of `frame executed` should be `frame is executed`

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+            }
+
+            /** Construct algoritm to operate on particles in a superCell
              *
-             * @{
+             * @param workerIdx workerIdx index of the worker: range [0;workerSize)
+             * @param particlesBox particles memory
+             *                     It is not allowed concurrently to add or remove particles during the
+             *                     execution of this algorithm.
+             * @param superCellIdx index of the superCell where particles should be processed
```
sbastrakov, line 138: Is it really a supercell index, or more generally an index inside the given particle box?

If the latter, I think it is better to rename in order to separate the basic for each logic from the normal PIConGPU way of organizing parallel particle processing

sbastrakov: Perhaps a related thought - do we need to operate as the whole box and supercell index here? It would appear more logical to have a single supercell already as our collection of particles to operate on. As then we would have a more crisp and standard parallel for-each logic here.

A user interface could still be a box + index if that is more convenient, a transition could also be made inside

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+            }
 
-            /** Version for a custom area mapper factory
+            /** Execute unary functor for each particle.
              *
-             * @tparam T_AreaMapperFactory factory type to construct an area mapper that defines the area to process,
-             *                             adheres to the AreaMapperFactory concept
+             * @attention There is no guarantee in which order particles will be executed.
+             *            If the particle functor is used the algorithm assumes that the frame structure is following
+             *            the rule tat all except the last frame are fully filled and in the last frame the low indices
```
sbastrakov, line 154: Typo: should be `that`

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+             *            within the frames will be contigious filled too.
```
sbastrakov: Would better be pharsed as
`the rule that all frames but the last are fully filled and in the last frame the low indices are contigiously filled.`

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+             * @tparam T_Acc alpaka accelerator type
+             * @tparam T_Functor unary particle functor or frame functor following the interface concept
+             *                           detail::FrameFunctorInterface or detail::ParticleFunctorInterface
+             * @param acc alpaka accelerator
+             * @param unaryFunctor Functor executed for each particle/frame.
+             *                     The caller must ensure that calling the functor in parallel with
+             *                     different workers is data race free.
+             *                     particle functor: 'void operator()(T_Acc const &, ParticleType)'.
+             *                       It is not allowed to call a synchronization function within the functor.
+             *                     frame functor: 'void operator()(T_Acc const &, FrameCtx)'.
```
sbastrakov: Did you mean `For frame ...` ?

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
-            HINLINE void forEach(T_Species&& species, T_Functor functor, T_AreaMapperFactory const& areaMapperFactory)
+            template<typename T_Acc, typename T_Functor>
+            DINLINE void operator()(T_Acc const& acc, T_Functor&& unaryFunctor) const
+            {
+                if constexpr(numWorkers > frameSize)
+                {
+                    PMACC_CASSERT_MSG(
+                        __number_of_workers_must_be_a_multiple_of_framesize,
+                        (numWorkers % frameSize) == 0);
+                }
```
sbastrakov: The error message does not match the condition: from the condition `numWorkers` can be either `<= frameSize` or a multiple of it

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+
+                auto const& superCell = m_particlesBox.getSuperCell(m_superCellIdx);
+                uint32_t const numPartcilesInSupercell = superCell.getNumParticles();
+
+                // Information used to know if a particle within a frame is active without checking the multiMask.
+                auto frameConditionData = T_Order::createCtx(forEachParticleInFrame);
+
+                // end kernel if we have no particles
+                if(numPartcilesInSupercell == 0)
+                    return;
```
sbastrakov: I think this check is better put right after we got `numPartcilesInSupercell`. Or should `frameConditionData` be created always - if so, please add a comment about it?

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+            {
+                if constexpr(numWorkers > frameSize)
+                {
+                    PMACC_CASSERT_MSG(
+                        __number_of_workers_must_be_a_multiple_of_framesize,
+                        (numWorkers % frameSize) == 0);
+                }
+
+                auto const& superCell = m_particlesBox.getSuperCell(m_superCellIdx);
+                uint32_t const numPartcilesInSupercell = superCell.getNumParticles();
```
sbastrakov: Typo: should be `numParticlesInSupercell`

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+                auto framePtrCtx = forEachParticleInFrame(
+                    [&](lockstep::Idx const idx)
+                    {
+                        auto frame = T_Order::getFirst(m_particlesBox, m_superCellIdx);
+
+                        if constexpr(std::is_same_v<Reverse, T_Order>)
+                        {
+                            if(frame.isValid())
+                                frameConditionData[idx]
+                                    = m_particlesBox.getSuperCell(m_superCellIdx).getSizeLastFrame();
```
sbastrakov: Can variable `superCell` be used instead?

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+                            {
+                                frame = T_Order::next(m_particlesBox, frame);
+                            }
+                            if constexpr(std::is_same_v<Reverse, T_Order>)
+                                frameConditionData[idx] = frameIdx != 0 ? frameSize : frameConditionData[idx];
+                        }
+                        return frame;
+                    });
+
+                bool isOneFrameValid = true;
```
sbastrakov: Why `one`? To me it seems more like a `isCurrentFrameValid`

include/picongpu/particles/debyeLength/Estimate.kernel
```diff
@@ -118,28 +118,24 @@ namespace picongpu
...
-                                    sumMomentumSquared += weighting * singleParticleMomentum * singleParticleMomentum;
-                                }
-                            });
-                        frame = electronBox.getPreviousFrame(frame);
-                        particlesInFrame = frameSize;
-                    }
+                    auto forEachParticle = pmacc::particles::algorithm::acc::makeForEach<numWorkers>(
+                        workerIdx,
+                        electronBox,
+                        supercellIdx);
```
sbastrakov, line 126: As a idea for a follow-up PR (this one is already very large).

With these changes, I think we can finally go away from operating directly with `numWorkers` and `workerIdx` in "simple" kernels that just need to go over the particles. With shared memory like for current deposition I understand those are still needed as this point doesn't apply.

So perhaps we can make a simplified way to create an in-kernel for each particle that uses default `numWorkers`. This way those kernels would not need this (potentially confusing) template parameter or directly use thread indices. Ultimately, then such kernels would become a more direct for-each invocation as they imo should.

include/pmacc/particles/algorithm/ForEach.hpp
psychocoderHPC: Yes depending on the iteration order the variable is used differently. `lcellId_t` is enough to store the number of particles in a frame.
For the other directing the variable is used to store number of particles in a supercell, there it mus the 32Bit.

include/pmacc/particles/algorithm/ForEach.hpp
psychocoderHPC: You are right the distance to the next frame. Let's check this tomorrow offline and find a better naming.

include/pmacc/particles/algorithm/ForEach.hpp
psychocoderHPC: It is supercell index we need to operate on.

>  do we need to operate as the whole box and supercell index here? 

I also thought first of it but because of how pmacc is structured internally working with a supercell will not be easily possible without other refactorings in pmacc. There is no link from a supercell to the particle box you need to iterate over frames, therefore I decided to work with the supercell index.

include/pmacc/particles/algorithm/ForEach.hpp
psychocoderHPC: All looks fine, in cases where we have more workers than particles in a frame the number of workers must be a multiple of the frame size.

include/pmacc/particles/algorithm/ForEach.hpp
psychocoderHPC: I can move `frameConditionData` down.

include/pmacc/particles/algorithm/ForEach.hpp
psychocoderHPC: A worker can have multiple frames depending on the number of virtual threads the worker is representing.
We need to know if one of these frames a worker is holding is valid.

include/picongpu/particles/debyeLength/Estimate.kernel
psychocoderHPC: This is only possible if forEachParticle is rewritten to work with runtime sizes only. The number of workers depends on the alpaka accelerator we are using. I would avoid using any `default numWorkers` because the default will never be good. 
A forEach which is fully runtime would IMO make more sense.

include/pmacc/particles/algorithm/ForEach.hpp
sbastrakov: So the condition is fine. But then the name does not exactly express the condition. We can discuss tomorrow

include/pmacc/particles/algorithm/ForEach.hpp
sbastrakov: So `createCtx` is not a collective operation?

include/pmacc/particles/algorithm/ForEach.hpp
psychocoderHPC: I will change it to `frameDistance`

include/pmacc/particles/algorithm/ForEach.hpp
psychocoderHPC: I will put a dash in front to separate that we handle once a particle functor and once the frame functor.

include/pmacc/particles/algorithm/ForEach.hpp
psychocoderHPC: should be `__number_of_workers_must_be_either_less_or_a_multiple_of_framesize`

include/picongpu/particles/debyeLength/Estimate.kernel
sbastrakov: What i meant is, most (not sure, maybe some, not most) PIConGPU particle-based kernels (current deposition is exception) use template parameter `uint32_t T_numWorkers` and on the host side we use `pmacc::traits::GetNumWorkers<...>` to provide a value for it. I was thinking to just simplify this common case by using this value by default.

include/pmacc/particles/algorithm/ForEach.hpp
psychocoderHPC: offline discussed @sbastrakov suggested: `hasUnprocessedFrames` or `hasMoreWork`

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+
+                // get starting frame
+                auto framePtrCtx = forEachParticleInFrame(
+                    [&](lockstep::Idx const idx)
+                    {
+                        auto frame = T_Order::getFirst(m_particlesBox, m_superCellIdx);
+
+                        if constexpr(std::is_same_v<Reverse, T_Order>)
+                        {
+                            if(frame.isValid())
```
sbastrakov, line 200: Why is this condition needed? If we had no particles we would have returned already

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+                    PMACC_CASSERT_MSG(
+                        __number_of_workers_must_be_a_multiple_of_framesize,
+                        (numWorkers % frameSize) == 0);
+                }
+
+                auto const& superCell = m_particlesBox.getSuperCell(m_superCellIdx);
+                uint32_t const numPartcilesInSupercell = superCell.getNumParticles();
+
+                // Information used to know if a particle within a frame is active without checking the multiMask.
+                auto frameConditionData = T_Order::createCtx(forEachParticleInFrame);
```
sbastrakov: I think this variable is very confusingly named. Need to discuss on what does it express

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+                                }
+                                else if constexpr(std::is_same_v<Forward, T_Order>)
+                                {
+                                    uint32_t const parIdxInSupercell = frameConditionData[linearIdx] + linearIdx;
+                                    bool const isParticle = parIdxInSupercell < numPartcilesInSupercell;
+                                    if(!isParticle)
+                                        particle.setHandleInvalid();
+                                }
+                                else
+                                    static_assert(!sizeof(T_Order), "Unsupported order policy");
```
sbastrakov: Discussed offline: could be kept as-is, but may be moved somewhere (e.g. checked even outside this function), as it is a precondition on the class template argument.

include/pmacc/particles/algorithm/ForEach.hpp
```diff
@@ -22,157 +22,447 @@
...
+                                    = m_particlesBox.getSuperCell(m_superCellIdx).getSizeLastFrame();
+                        }
+
+                        /* Same as numFramesToSkip, variable is required to workaround an nvcc compiler bug.
+                         * Without redefinition within the lambda the variable will be undefined.
+                         */
+                        constexpr uint32_t skipNFrames = (numWorkers + frameSize - 1u) / frameSize;
+                        if constexpr(skipNFrames > 1)
+                        {
+                            uint32_t const frameIdx = idx / frameSize;
```
sbastrakov: This  variable looks more like the number of skipped frames

include/pmacc/particles/algorithm/ForEach.hpp
psychocoderHPC: It is a collective operation but is not requiring any synchronization.

psychocoderHPC:
The CI should now hopefully pass all tests.

include/picongpu/particles/debyeLength/Estimate.kernel
sbastrakov: Discussed offline, as a follow-up to this work we can make some helpers to simplify writing and launching such kernels. That may also include ensuring that the `T_numWorkers` (or a new way to express it) matches the kernel block size - which is a requirement even now, but we do not check for it.

include/pmacc/particles/algorithm/ForEach.hpp
sbastrakov: @psychocoderHPC what is your opinion on this?

include/pmacc/particles/algorithm/ForEach.hpp
sbastrakov: I still strongly dislike the name, but can live with it. After offline discussion it seems the concept represented by this variable is also quite confusing but has to be so due to performance and register reasons (and this is definitely a performance-critical bit).

include/pmacc/particles/algorithm/ForEach.hpp
psychocoderHPC: Yes I forgot about it. It is not necessarily required.
If there are no must-have changes I will fix it in a separate PR.

sbastrakov (approved):

PR #4174, opened 2022-06-21
Calculate and output laser phase velocity

sbastrakov:
This PR extends PIConGPU simulation start log output with phase velocity of the laser enabled in `laser.param`. 
It looks like this:
```bash
...
PIConGPUVerbose PHYSICS(1) | Field solver condition: c * dt <= 1.4912 ? (c * dt = 1)
PIConGPUVerbose PHYSICS(1) | Laser numerical dispersion: v_phase = 0.996018 * c
PIConGPUVerbose PHYSICS(1) | Resolving plasma oscillations?
...
```
This part does not perform any check or print a warning when far away from `c`, as it is hard to define reasonable margins. Same as Debye length check, It is information for a user to use their own judgement. (This may be particularly relevant for long simulations, also when using high-order solver and/or time substepping coming in the future).

The calculation is done for the grid parameters and field solver used in a simulation, so matches the current simulation. All our field solvers are supported with their respective numerical dispersion relations used. For now the calculation and output are done only for laser, not for incident field or background field. However, all the internals are general enough to add at least incident field quickly in a follow-up PR (there are some mostly unrelated technical reasons I didn't do it as part of this PR).

Even though our standard lasers always propagate in along `y`, all internal implementation is for general propagation direction (so that we can add incident fields soon). In that general case, the numerical dispersion relation has no explicit form for solution. So the new code always solves it numerically. However, it should  normally be very precise as we know a very good (for a reasonably set up simulation) approximate solution corresponding to `phase_velocity = c`. When I checked against the theoretically known solution for Yee solver and axis-aligned and diagonal propagation directions, the numerical solution had 6 correct digits. This is also why I did not implement these cases as special to fall back. I also checked that the general behavior is reasonable e.g. for Yee diagonal propagation is closer to c than axis-aligned, for Lehe solver our calculated phase velocity is slightly above c in that direction and below it in (substantially) different directions, and that ~~for high-order AOFDTD it could become quite substantially larger than c as also theoretically expected. So basically it all seems to work fine. Edit: for the latter also see my next message~~ edit: I initially implemented wrong dispersion relation for AOFDTD, now fixed and phase velocity is only slightly above c

### DIFF
--- a/include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.hpp
+++ b/include/picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/ArbitraryOrderFDTD.hpp
@@ -26,6 +26,7 @@
 #include "picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Derivative.hpp"
 #include "picongpu/fields/MaxwellSolver/ArbitraryOrderFDTD/Weights.hpp"
 #include "picongpu/fields/MaxwellSolver/CFLChecker.hpp"
+#include "picongpu/fields/MaxwellSolver/DispersionRelation.hpp"
 #include "picongpu/fields/MaxwellSolver/LaserChecker.hpp"
 #include "picongpu/fields/differentiation/Curl.hpp"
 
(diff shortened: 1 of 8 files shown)


sbastrakov:
Discussed offline with @steindev but to also document here. We need to check against the old calculations of arbitrary-order FDTD phase velocity (used to produce plots for readthedocs), as I got an impression the results calculated by this PR may differ. The scenario was different though, so a more thorough check would help.

steindev:
Just noted that the dispersion relation we try to solve is not correct. The derivation for 2D is the following
AOFDTD-disp-relation_2D

We need to derive this for 3D and change the respective paragraph in the manual as well.

steindev:
The 2D version posted above is actually equivalent to the expected disp. relation with squares of sums. All fine, will merge.

steindev (approved):

PR #4176, opened 2022-06-22
Rework implementation of pmacc::Unique

sbastrakov:
The previous implementation used an unconventional meaning of unique competely removing all duplicates.
(So that elements present multiple times were not present in the resulting sequence.)
Switch to conventional meaning of unique, always keeping exactly one instance.

Note that it does not affect existing checks for uniqueness of flags and attributes.
For those checks, both old and new semantics of unique result yield the same outcome.

So it is not a bug fix per se, but switching to conventional semantics.


sbastrakov:

psychocoderHPC (approved):

PR #4186, opened 2022-06-28
Output numerical phase velocity for enabled incident field profiles

sbastrakov:
Similarly to how it has recently been added for laser. Each different enabled incident field will be a separale line in the output, the calculation will use its wavelength and propagation direction. That is exept `None` (no phase velocity) and `Free` (no way to automatically calculate phase velocity).

This PR also fixes a mistake in string names in some incident field profiles, that was just introduced in #4175 (with a plan to be used here).


sbastrakov:

psychocoderHPC (approved):

PR #4187, opened 2022-06-29
Update boost and openpmd version in defq hemera profile

pordyna:


sbastrakov (commented):
It seems Boost version is below the required one in all Hemera profiles, not just that one?

pordyna:
true, it looks like it was only updated for fwkt_v100

pordyna:
`openpmd` and `gcc` are also outdated, I will check what modules are available on hemera and updated this PR accordingly.

pordyna:
@sbastrakov @psychocoderHPC this is a working profile that satisfies the current requirements for `boost` and `openpmd` versions. We probably should ask for new modules so that we can switch to a newer compiler (the no cuda openpmd module required 7.3.0 gcc).  I tested it by running all compile tests. Also, the thermalization test starts and writes the first iteration via `openpmd`. 

We also need to update other profiles. For gpu queue, I already have a working one. For the k80 and k20 we don't have an openpmd module for the older cuda version.
I will open a github issue  for that, but for the IT ticket, it would be great if one of you could compose a requirements list for each queue, since maybe there is also sth else that needs updating.

psychocoderHPC (approved):

Issue #4142, opened 2022-06-09
What is the meaning of MomentumComponent derived attribute?


sbastrakov:
This is a derived field available in the default `fileOutput.param`, so in a way it's available (but not enabled) by default. However, its meaning and intended use case feels not clear to me. There is [this description](https://github.com/ComputationalRadiationPhysics/picongpu/blob/74b6e552f7a436d517574b93c35c17830619994d/include/picongpu/particles/particleToGrid/derivedAttributes/MomentumComponent.def#L41-L50) that matches the [operations performed for each macroparticle](https://github.com/ComputationalRadiationPhysics/picongpu/blob/74b6e552f7a436d517574b93c35c17830619994d/include/picongpu/particles/particleToGrid/derivedAttributes/MomentumComponent.hpp#L43-L58).

However, then when calculating a derived field each value will be the sum of those ratios for all contributing macroparticles also taking account their shape, and what is the meaning of the sum? For example, imagine we are calculating it for x component (so template parameter 0) and there are two macroparticles at same position with different weights, one going strictly in +x and the other strictly in -x. By how it's implemented, first one will give +1 and second -1, so that the total value will be 0. However, if the weights are different should it actually be zero, as the average momentum in x is not 0? So roughly speaking we are doing a term-wise ratio and then sum, while I would have expected we first sum and then do a ratio of the two sums.

Basically, my question comes from trying to apply it to calculate an average component of momentum (as was asked by a user), then trying to use `MomentumComponent` and realizing it doesn't fit, and then wondering what does it do at all. It could well be i'm missing the context of its application, then it may point out to lacking documentation there. If anyone has experience using this attribute, please share.

P.S. this attribute also doesn't correctly interact with averagizing, but i think it's a problem in the averagizing, i will make a separate issue #4143 for that.

cc @steindev

pordyna:
Oh, this is weird. I would also expect  calculating the total momentum in 3D and the total momentum in a certain direction and then take the ratio. But this is actually sth. that can be done in a single attribute only with a combined attribute, which is a new feature.

sbastrakov:
@pordyna do you mean to make a new "simple" attribute for a component of momentum that for each macroparticle will return `weight * px` and then use `AverageAttribute` of that to get average value of `px` as a field?

sbastrakov:
[...]

PR #4191, opened 2022-06-30
Remove derived attribute/field MomentumComponent

sbastrakov:
As discussed in #4142, its meaning is unclear. Output of single components of momentum is now possible via `Momentum<>`, `MomentumDensity<>`, and `Average<Momentum<>>`. So they effectively supercede the old `MomentumComponent<>`.

**Note to users**: this is potentially a breaking change for existing setups. Before this PR we used that derived field in default `fileOutput.param`. So if your setup has a custom `fileOutput.param` derived from there and you kept this part, it will no longer compile after this PR is merged. However, a fix is very simple: remove this part if unused (e.g. same as how this PR changed the default `fileOutput.param`), or switch to one of the options described above.


psychocoderHPC:
I assigned @steindev for this issue. We talked yesterday very short about this and before we remove it he should maybe talk to @sbastrakov again.

sbastrakov:
Sure, let us discuss next week in the office.

sbastrakov:
Discussed offline with @steindev . This PR is fine as is, since we already have the new `Momentum<>` attribute available to users, so that they can switch from `MomentumComponent<>` to that.

steindev (approved):

PR #4196, opened 2022-07-01
Allow componentwise calculation of incident field functors

sbastrakov:
This PR adds a new altenative interaface for incident field functors that would provide values for components separately. In case a functor implements the new interface, it would be preferred. All old functors still work, and most of them should not switch to the new interface. The only reason to switch is performance for very computationally intensive incident fields like TWTS. So most functions should not switch as they would gain practically nothing and it will just complicate things for no reason.

Add a hook for component-wise calculation of TWTSfast. This is an alternative to the previously existing interface. It will allow more efficient usage of TWTSfast inside Free incident field functors. (I will add an example in a follow-up PR).

### DIFF
--- a/include/picongpu/fields/incidentField/Solver.kernel
+++ b/include/picongpu/fields/incidentField/Solver.kernel
@@ -25,6 +25,7 @@
 #include "picongpu/fields/incidentField/DerivativeCoefficients.hpp"
 
 #include <cstdint>
+#include <type_traits>
 
 /** @note In this file we use camelCase "updatedField" in both code and comments to denote field E or B that is being
  * updated (i.e. corrected) in the kernel. The other of the two fields is called "incidentField". And for the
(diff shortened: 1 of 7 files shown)


include/picongpu/fields/incidentField/Solver.kernel
```diff
@@ -41,6 +42,54 @@ namespace picongpu
...
+                 *
+                 * Is void for those types, ill-formed otherwise.
+                 *
+                 * @tparam T_FunctorIncidentField incident field source functor type
+                 * @tparam T_component field component, 0 = x, 1 = y, 2 = z
+                 */
+                template<typename T_FunctorIncidentField, uint32_t T_component>
+                using HasGetComponent
+                    = std::void_t<decltype(alpaka::core::declval<T_FunctorIncidentField>()
+                                               .template getComponent<T_component>(floatD_X::create(0.0_X)))>;
```
sbastrakov, line 55: This gotta be the ugliest C++ line I wrote in a while. But before concepts I see no other way to express this interface requirement.

include/picongpu/fields/incidentField/Solver.kernel
```diff
@@ -159,8 +208,12 @@ namespace picongpu
                      */
                     float3_X coeff1, coeff2;
 
-                    //! Indices of the incidentField components for the two terms
-                    uint32_t incidentComponent1, incidentComponent2;
+                    /** Indices of the incidentField components for the two terms
+                     *
+                     * Note the intentional cross combination, this is to match the general vector update stated above.
+                     */
+                    static constexpr uint32_t incidentComponent1 = (T_axis + 2) % 3;
```
psychocoderHPC, line 215: Could it be that this is only a redefinition of `dir2`?

sbastrakov: Well technically yes. But I think it is more clear to have both variables, as depending on the situation one name makes more sense than the other.

sbastrakov: Upon further consideration, I decided to keep the `dir1, dir2` variables separate for the reasons stated above. However, I now calculate them based on `incidentComponent1, incidentComponent2` to avoid differences in how the calculation is done.

sbastrakov:
@psychocoderHPC ready for re-review, your suggestion is implemented.

psychocoderHPC (approved):

sbastrakov:
Discussed offline with @psychocoderHPC . It could be that it was actually sufficient to just change in-kernel variables `incidentComponent1, 2` to be `constexpr` and then perhaps a compiler already only calculated the component in question by knowing only that was used. As i was comparing this PR branch vs. `dev` at that time and saw improvement, but should have compared against that case. Now they did become `constexpr`, so I shall compare again.

PR #4216, opened 2022-07-14
activate cuda_memtest on crusher

psychocoderHPC:
Check health of GPUs with cuda_memtest before running PIConGPU.


sbastrakov (approved):

PrometheusPi:
Should we rename this to `gpu_memtest` on AMD systems?

PR #4223, opened 2022-07-15
clang: fix early assert evaluation

psychocoderHPC:
clang is evaluating static asserts inside of a class/struct even if the object will never be used. Therefore the evaluation must be deferred by using `sizeof(TEMPLATE_PARAM)` or moving the assert into a method of the class.

### DIFF
--- a/include/picongpu/fields/currentDeposition/EmZ/EmZ.hpp
+++ b/include/picongpu/fields/currentDeposition/EmZ/EmZ.hpp
@@ -45,7 +45,8 @@ namespace picongpu
 
             PMACC_CASSERT_MSG(
                 __EmZ_supercell_or_number_of_guard_supercells_is_too_small_for_stencil,
-                pmacc::math::CT::min<typename pmacc::math::CT::mul<SuperCellSize, GuardSize>::type>::type::value
+                sizeof(T_ParticleShape*) // defer assert evaluation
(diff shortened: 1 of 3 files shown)


include/picongpu/fields/currentDeposition/EmZ/EmZ.hpp
```diff
@@ -45,7 +45,8 @@ namespace picongpu
 
             PMACC_CASSERT_MSG(
                 __EmZ_supercell_or_number_of_guard_supercells_is_too_small_for_stencil,
-                pmacc::math::CT::min<typename pmacc::math::CT::mul<SuperCellSize, GuardSize>::type>::type::value
+                sizeof(T_ParticleShape) // defer assert evaluation
```
sbastrakov: Shouldn't we use `T_ParticleShape*` here to not depend on the type definition being available (and pointer only requires a declaration)

sbastrakov: @psychocoderHPC would you like to address this? It's not critical at all and I have no other comments, so can merge as is.

psychocoderHPC: I also thought about it, I will use the ptr.

sbastrakov (approved):

PR #4227, opened 2022-07-20
Fix particle assignment used in EmZ 2d for calculating Jz

### LINKED ISSUE #4201: Field energy NaN values for a 2D set-up.
While setting up a 2D simulation with a colleague we have encountered the following problem:

The 2D simulation produces `nan` field values for the field energy from the 1st step onward:
```
#step total[Joule] Bx[Joule] By[Joule] Bz[Joule] Ex[Joule] Ey[Joule] Ez[Joule] 
0 0.000000000000000e+00 0 0 0 0 0 0
5 nan nan nan nan nan nan nan
```
if compiled and run using cde7ccdfa156e24b26074904752274935d323d0a and the hemera V100 profile/TPL therein.

If compiled and run using 88fd0117d35d463a2302b8d2852587a48d22c8b5 the field energies appear correct:
```
#step total[Joule] Bx[Joule] By[Joule] Bz[Joule] Ex[Joule] Ey[Joule] Ez[Joule] 
0 0.000000000000000e+00 0 0 0 0 0 0
5 8.583515945519158e-13 3.58032e-13 5.18102e-15 1.05577e-14 3.98794e-14 3.97184e-14 4.04984e-13
```

The `nan` values do not result in a termination of the simulation. The problem appears in the field values and not only in the field energy computation.

The only error provided in `stderr` pertains to the  PSM2 component of OpenMPI (which, if memory serves, is the integration with SLURM and can be treated as a warning rather than an error):
```
A requested component was not found, or was unable to be opened.  This
means that this component is either not installed or is unable to be
used on your system (e.g., sometimes this means that shared libraries
that the component requires are unable to be found/loaded).  Note that
Open MPI stopped checking at the first component that it did not find.

Host:      gv016.cluster
Framework: mtl
Component: psm2
--------------------------------------------------------------------------
[gv001.cluster:152978] 15 more processes have sent help message help-mca-base.txt / find-available:not-valid
[gv001.cluster:152978] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

sbastrakov:
In that case we are generally not on support and have to use the respective assignment. The bug was introduced as a copy-paste mistake in #4170. Restore the assignment used before in this part.

Fixes #4201.

### DIFF
--- a/include/picongpu/fields/currentDeposition/EmZ/EmZ.hpp
+++ b/include/picongpu/fields/currentDeposition/EmZ/EmZ.hpp
@@ -129,7 +129,7 @@ namespace picongpu
                     [&, this](auto&& dataBoxJ)
                     {
                         /* For Jz we consider the whole movement on a step.
-                         * This movement is not necessarily on support.
+                         * Note that this movement is not necessarily on support.
                          * A naive implementation would be to extend the bounds in x, y by 1 in both sides, and use
                          * general assignment function. To optimize it, we redefine shiftEnd as component-wise minimum
                          * between old shiftEnd and shiftStart. We calculate everything relative to the new shiftEnd.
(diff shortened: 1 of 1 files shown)


sbastrakov:
@psychocoderHPC ready for review.

psychocoderHPC (approved):

PR #4229, opened 2022-07-21
documentation: mention `CMAKE_BUILD_TYPE`

psychocoderHPC:
Document that `CMAKE_BUILD_TYPE=Debug` can be used to activate all
asserts in the code.

@BrianMarre I added this only for you ^^

### DIFF
--- a/docs/source/dev/debugging.rst
+++ b/docs/source/dev/debugging.rst
@@ -17,6 +17,7 @@ The following build options can assist the investigation:
 * ``PMACC_VERBOSE=<N>`` sets log detail level for pmacc, highest level is 127.
 * ``PMACC_BLOCKING_KERNEL=ON`` makes each kernel invocation blocking, which helps to narrow a crash down to a particular kernel.
 * ``CUPLA_STREAM_ASYNC_ENABLE=OFF`` disables asynchronous streams, also helps to narrow a crash down to a particular place.
+* ``CMAKE_BUILD_TYPE=Debug`` compile in debug mode where all assertions are activated. Compiling in debug mode will slowdown your kernels!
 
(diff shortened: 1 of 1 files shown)


docs/source/dev/debugging.rst
```diff
@@ -17,6 +17,7 @@ The following build options can assist the investigation:
 * ``PMACC_VERBOSE=<N>`` sets log detail level for pmacc, highest level is 127.
 * ``PMACC_BLOCKING_KERNEL=ON`` makes each kernel invocation blocking, which helps to narrow a crash down to a particular kernel.
 * ``CUPLA_STREAM_ASYNC_ENABLE=OFF`` disables asynchronous streams, also helps to narrow a crash down to a particular place.
+* ``CMAKE_BUILD_TYPE=Debug`` compile in debug mode where all assertions are activated.
```
sbastrakov: I think we should also add a warning that on GPUs it should be avoided as possible, and only used as last resort ?

psychocoderHPC: I added a note that kernels would be slower. In general, it is useful to test new code in debug mode too.

sbastrakov (approved):

pordyna:
It is enough to remove `-NDEBUG` from compiler flags and add `PMACC_BLOCKING_KERNEL` and `alpaka_CUDA_SHOW_CODELINES` to enable all asserts and get enough information out of it. There is no need  for compiling without optimization.  But, I think this change to documentation suggests  that  `CMAKE_BUILD_TYPE=Debug` is the way to go for enabling assertions.     @psychocoderHPC, @BrianMarre what do you think?

PR #4230, opened 2022-07-21
crusher: add slurm fault tolerance

psychocoderHPC:
Use `cuda_memtest` to find broken nodes.
Automatically exclude broken nodes if necessary.

The environment variable `PIC_NODE_OVERSUBSCRIPTION_PT` can be used to
adjust the additional nodes based on the job requirements.

### DIFF
--- a/etc/picongpu/crusher-ornl/batch.tpl
+++ b/etc/picongpu/crusher-ornl/batch.tpl
@@ -26,8 +26,8 @@
 #SBATCH --time=!TBG_wallTime
 # Sets batch job's name
 #SBATCH --job-name=!TBG_jobName
-#SBATCH --nodes=!TBG_nodes
-#SBATCH --ntasks=!TBG_tasks
+#SBATCH --nodes=!TBG_nodes_adjusted
+#SBATCH --ntasks=!TBG_tasks_adjusted
 #SBATCH --mincpus=!TBG_mpiTasksPerNode
 #SBATCH --cpus-per-task=!TBG_coresPerGPU
 #SBATCH --mem-per-gpu=!TBG_memPerDevice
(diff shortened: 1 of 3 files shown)


etc/picongpu/crusher-ornl/batch.tpl
```diff
@@ -93,19 +101,63 @@ mkdir simOutput 2> /dev/null
...
+            echo "cuda_memtest: Number of broken nodes has not increased but for unknown reasons cuda_memtest reported errors." >&2
+            break
+        fi
+        # if cuda_memtest not passed and we have no broken nodes something else went wrong
+        if [ $node_check_err -ne 0 ] ; then
+            if [ $n_broken_nodes -eq 0 ] ; then
+                echo "cuda_memtest: unknown error" >&2
+                break
+            else
+                echo "cuda_memtest: "$n_broken_nodes" broken node(s) detected!. The test will be repeated with health nodes only." >&2
```
sbastrakov:
```suggestion
                echo "cuda_memtest: "$n_broken_nodes" broken node(s) detected!. The test will be repeated with healthy nodes only." >&2
```

etc/picongpu/crusher-ornl/batch.tpl
```diff
@@ -93,19 +101,63 @@ mkdir simOutput 2> /dev/null
...
+fi
+
+if [ $node_check_err -eq 0 ] || [ $run_cuda_memtest -eq 0 ] ; then
+    # Run PIConGPU
+    echo "Start PIConGPU."
+    test $n_broken_nodes -ne 0 && exclude_nodes="-x./bad_nodes.txt"
+    srun -n !TBG_tasks --nodes=!TBG_nodes $exclude_nodes -K1 !TBG_dstPath/input/bin/picongpu --mpiDirect !TBG_author !TBG_programParams
+else
+    echo "Job stopped because of previous issues."
+    echo "Job stopped because of previous issues." >&2
```
sbastrakov, line 161: Why is this line twice?

etc/picongpu/crusher-ornl/batch_craycc_picongpu.profile.example
```diff
@@ -28,6 +28,11 @@ export MY_NAME="$(whoami) <$MY_MAIL>"
 #   - project for allocation and shared directories
 export PROJID=<yourProject>
 
+# Job control
+# Allocate more nodes then required to execute the jobs to be able to handle broken jobs
```
sbastrakov:
```suggestion
# Allocate more nodes then required to execute the jobs to be able to handle broken nodes
```

etc/picongpu/crusher-ornl/batch.tpl
psychocoderHPC: We have a file for errors and normal output. Many messages about the status are in the stdout output.
TO avoid the user missing that PIConGPU is not running I push the message to both files.

sbastrakov (approved):

Issue #4208, opened 2022-07-13
Radiation plugin: high frequency radiation "trees"


PrometheusPi:
The radiation plugin seems to produce spectra with minima that split over time and create a tree like structure. The effect is observable for frequencies much higher than the laser or plasma frequency.
It is currently unclear whether this is a numerical or real physical effect. 

The "trees" occurs for e.g. betatron radiation during LWFA or hosing during PWFA.

Anton-Le:
Here's an example of such a spectrum radiated by the bunch:
![BunchSpectrumLineDoubling](https://user-images.githubusercontent.com/50339174/178716695-7604178d-1f4c-4532-a879-ef3b3447c732.png)

The bunch travelling at a relativistic gamma of 1900 impacts the background plasma at around 17000 time steps. Frequency ind rad/s (angular frequency).

Anton-Le:
I have run the same set-up for comparison purposes with the **background plasma present** (same as in the post above)
![Spectrum_detector22_withbg](https://user-images.githubusercontent.com/50339174/179086456-63f44877-be71-4b3a-b55d-ac07c2f12c10.png)

and with the **background plasma removed**:
![Spectrum_detector22_nobg](https://user-images.githubusercontent.com/50339174/179087207-09916a08-6796-4736-afa5-3e854c1aace6.png)

Both plots show the same detector at an angle of ~0.08° to the propagation direction of the bunch.
While the upper limites of the log-colorscale are different, the overall structure of the spectrum prevails.

Hence the "trees" are unlikely to be related to the interaction with the background plasma.

Anton-Le:
The following shows the spectrum for the same detector and the same simulation set-up (**no background plasma**) but with the default computational precision set to 64 bit in `precision.param`:
![Spectrum_detector22_nobg_64bit](https://user-images.githubusercontent.com/50339174/179394735-7b4e024c-b252-4bf5-8786-0d2d1203d9cd.png)


Since the pattern vanishes with a switch to higher precision I conclude that the 'tree' structure shown in the above posts is of numerical nature.

Additional information:
For the 64bit simulation the super-cell size had to be reduced to `4x8x4` from `8x8x4` otherwise the PIConGPU throws a compile-time error indicating that the shared memory is not sufficient for the selected combination of precision (64) and particle shape (PQS).

sbastrakov:
[...]

PR #4231, opened 2022-07-25
Fix radiation trees.

Anton-Le:
This commit fixes the source of a periodic doubling of minima/maxima when computing radiation above the critical betatron frequency of the bunch as described in #4208 

The source are accumulations of rounding errors in `particle_locationNow`. Switching the variable to a default 64 bit precision fixes the problem.

### DIFF
--- a/include/picongpu/plugins/radiation/Radiation.kernel
+++ b/include/picongpu/plugins/radiation/Radiation.kernel
@@ -266,14 +266,20 @@ namespace picongpu
                                                     + DataSpaceOperations<simDim>::template map<SuperCellSize>(
                                                         cellIdx));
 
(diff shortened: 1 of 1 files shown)


sbastrakov (commented):
Nice!

Two small suggestions:

- Please add a comment in front of `particle_locationNow` declaration that is must be double precision to avoid the trees and link to the issue (so that no one will be tempted to "optimize" it to `float_X` in the future).
- Your commits are better squashed, as the second commit undoes part of the first commit and the messages together are somewhat confusing - the git log now would say that phase is back to `float_X` but it's eventually unchanged from your PR as a whole

In principle, both are optional, but should also take just a couple mins.

sbastrakov (approved):

PR #4232, opened 2022-07-27
Clarify docs on setting same sources for incident field on each side

sbastrakov:
Clarifications are based on questions I got from @pordyna , thanks for the feedback!

I will make another PR to expand the incident field example and showcase/explain it there as well.

### DIFF
--- a/docs/source/usage/workflows/addLaser.rst
+++ b/docs/source/usage/workflows/addLaser.rst
@@ -24,7 +24,7 @@ A user defines positioning of this box in the total domain in :ref:`incidentFiel
 
 With properly chosen sources, the generated waves only propagate inwards the volume bounded by the generation surface and there is no considerable leakage or noise.
 The generating surface is otherwise transparent for all signals, such as reflections of lasers from a target.
-A typical setup includes field absorber located outside of the surface (directly or at a distance) for these waves.
+A typical setup includes a field absorber located outside of the surface (directly or at a distance).
 
 The surface must be offset inwards relative to each boundary by at least the field absorber thickness along the boundary so that the generating surface is located in the internal area.
 An exception to this requirement is made for simulations using the moving window.
@@ -35,7 +35,7 @@ For each of the generation planes ``XMin, XMax, YMin, YMax, ZMin, ZMax`` (the la
 In case a typelist is used, the result is a sum of all profiles in the list.
 
 In principle, the same sources should be applied at the whole generating surface, not just at planes where the lasers enter.
-Then the generated incident field will only exist in the internal volume, with application at the opposite side compensating and effectively removing it.
+Then, the generated incident field will only exist in the internal volume, with application at the opposite side compensating and effectively removing it.
 Note that this effect is not related to having a field absorber, but a property of the total field/scattered field formulation.
 In practice there may be some noise due to numerical dispersion or imprecise source formulation.
 In this case, a user may apply sources only at the "enter" parts of the generating surface but not on the opposite side (which will then be transparent), and employ a field absorber if needed.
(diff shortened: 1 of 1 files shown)


docs/source/usage/workflows/addLaser.rst
```diff
@@ -22,13 +22,24 @@ The source is applied at a generating surface that is a boundary of an axis-alig
...
+
 The surface must be offset inwards relative to each boundary by at least the field absorber thickness along the boundary so that the generating surface is located in the internal area.
 An exception to this requirement is made for simulations using the moving window.
 Then the surface positions along window movement direction can be located outside of the initially simulated volume.
 In this case, parts of the surface located outside of the currently simulated volume are treated as if they had zero incident field and it is user's responsibility to apply a source matching such a case.
 
 For each of the generation planes ``XMin, XMax, YMin, YMax, ZMin, ZMax`` (the latter two for 3d) a user sets an incident profile, or a typelist of such profiles, to be applied.
 In case a typelist is used, the result is a sum of all profiles in the list.
+
+In principle, the same sources should be applied at the whole generating surface, not just at planes where the lasers enter.
```
pordyna, line 37:
```suggestion
In principle, the same sources should be applied at the whole generating surface, not just at planes where the lasers enters.
```

docs/source/usage/workflows/addLaser.rst
```diff
@@ -22,13 +22,24 @@ The source is applied at a generating surface that is a boundary of an axis-alig
...
+Then the generated incident field will only exist in the internal volume, with application at the opposite side compensating and effectively removing it.
```
pordyna:
```suggestion
Then, the generated incident field will only exist in the internal volume, with application at the opposite side compensating and effectively removing it.
```

docs/source/usage/workflows/addLaser.rst
```diff
@@ -22,13 +22,24 @@ The source is applied at a generating surface that is a boundary of an axis-alig
 The implementation is based on the total field/scattered field formulation described in detail :ref:`here <model-TFSF>`.
 A user defines positioning of this box in the total domain in :ref:`incidentField.param <usage-params-core>`.
 
+With properly chosen sources, the generated waves only propagate inwards the volume bounded by the generation surface and there is no considerable leakage or noise.
+The generating surface is otherwise transparent for all signals, such as reflections of lasers from a target.
+A typical setup includes field absorber located outside of the surface (directly or at a distance) for these waves.
```
pordyna:
```suggestion
A typical setup includes field absorber located outside of the surface (directly or at a distance).
```
What waves?

pordyna (changes requested):
Except for these 3 minor comments, it looks good to me.

docs/source/usage/workflows/addLaser.rst
sbastrakov: Here i think it was correct, as it was verb for `lasers`?

docs/source/usage/workflows/addLaser.rst
sbastrakov: I meant all other signals as in previous sentence . But i think it's already clear and `these waves` can be just dropped?

pordyna: yes, I would just drop it. But you could also reformulate it somehow. And I think it should be "a field absorber".

docs/source/usage/workflows/addLaser.rst
pordyna: you are right

sbastrakov:
Thanks for the review @pordyna , your suggestions are implemented.

pordyna (approved):

psychocoderHPC (approved):

PR #4235, opened 2022-07-28
`pic-configure` validate preset selection

psychocoderHPC:
In cases, the user is using `-t` we are checking now if the used preset id is available.

This PR additionally fixes that the path to the file `cmakeFlags` was shown as a relative path instead of an absolut. In cases `pic-build` was used the path was relative to `.build`.

### DIFF
--- a/bin/pic-configure
+++ b/bin/pic-configure
@@ -24,6 +24,12 @@ this_dir=$(cd $(dirname $0) && pwd)
 # PIConGPU prefix path
 picongpu_prefix=$(cd $this_dir/.. && pwd)
 
+function absolute_path()
+{
+    cd $1
+    pwd
+}
+
 help()
 {
     echo "Configure PIConGPU with CMake"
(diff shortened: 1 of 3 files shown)


bin/pic-configure
```diff
@@ -157,18 +163,32 @@ while true ; do
...
+extension_param="$(absolute_path $*)"
 
 if [ ! -d "$extension_param" ] ; then
     echo "Path \"$extension_param\" does not exist." >&2
     exit 1
 fi
 # check for cmakeFlags file (interprete with sh)
 if [ -f "$extension_param/cmakeFlags" ] ; then
+    if [ $cmakeFlagsNr -eq -1 ] ; then
+        # if the users has not given this value use zero as default.
```
sbastrakov:
```suggestion
        # if the user has not given this value use zero as default.
```

sbastrakov (dismissed):
Besides the typo above, no further suggestions.

sbastrakov (dismissed):

sbastrakov (approved):

PR #4236, opened 2022-07-29
auto adjust particle exchange buffers for double precision

psychocoderHPC:
PIConGPU setups are typically designed for single precision execution.
The memory for particle exchanges defined in `memory.param` is therefore not correct if someone is switching to double precision.

To avoid that users must adjust their setup by hand this PR is automatically adjusting the required buffer sizes by multiplying the size by a factor of 2.

I know that it is not the best way to consume more memory than defined in the param files but this was the only way how to deal with the problem without updating each example and enforcing the example developer to run their initial simulation always in double precision to avoid that we have multiple send because of the buffer size limitation.

### DIFF
--- a/include/picongpu/param/memory.param
+++ b/include/picongpu/param/memory.param
@@ -71,14 +71,14 @@ namespace picongpu
 
     /** bytes reserved for species exchange buffer
      *
-     * This is the default configuration for species exchanges buffer sizes.
-     * The default exchange buffer sizes can be changed per species by adding
-     * the alias exchangeMemCfg with similar members like in DefaultExchangeMemCfg
-     * to its flag list.
+     * This is the default configuration for species exchanges buffer sizes when performing a simulation with 32bit
+     * precision (default for PIConGPU). For double precision the amount of memory used for exchanges will
+     * be automatically doubled. The default exchange buffer sizes can be changed per species by adding the alias
+     * exchangeMemCfg with similar members like in DefaultExchangeMemCfg to its flag list.
      */
     struct DefaultExchangeMemCfg
     {
-        // memory used for a direction
+        // Memory used for a direction for a simulation performed with 32bit precision.
         static constexpr uint32_t BYTES_EXCHANGE_X = 1 * 1024 * 1024; // 1 MiB
         static constexpr uint32_t BYTES_EXCHANGE_Y = 3 * 1024 * 1024; // 3 MiB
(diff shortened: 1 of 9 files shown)


include/picongpu/param/memory.param
```diff
@@ -71,14 +71,14 @@ namespace picongpu
 
     /** bytes reserved for species exchange buffer
      *
-     * This is the default configuration for species exchanges buffer sizes.
-     * The default exchange buffer sizes can be changed per species by adding
-     * the alias exchangeMemCfg with similar members like in DefaultExchangeMemCfg
-     * to its flag list.
+     * This is the default configuration for species exchanges buffer sizes when performing a simulation with 32bit
+     * precision (default for PIConGPU). For double precision the amount of memory used for exchanges will
+     * automatically doubled. The default exchange buffer sizes can be changed per species by adding the alias
```
sbastrakov:
```suggestion
     * be automatically doubled. The default exchange buffer sizes can be changed per species by adding the alias
```

include/picongpu/particles/Particles.tpp
```diff
@@ -184,8 +184,12 @@ namespace picongpu
         else
             exchangeBytes = ExchangeMemCfg::BYTES_CORNER;
 
+        /* PIConGPU default case is running with 32bit (4byte) precision. In case double precision is used the particle
+         * buffers must be twice as large to avoid that a communication must be split into multiple sends.
+         */
+        double precisionMultiplier = static_cast<double>(sizeof(float_X) / 4u);
```
sbastrakov:
```suggestion
        double precisionMultiplier = static_cast<double>(sizeof(float_X) / sizeof(float));
```

Does not change the logic, but is more description

include/picongpu/param/memory.param
sbastrakov: Same for other files.

include/picongpu/particles/Particles.tpp
sbastrakov: Or maybe even convert both sizeofs to `double` and then divide if we are very generic here

psychocoderHPC: I had this first too but switched and used `4u`. Maybe I will use what you suggested.

psychocoderHPC: > Or maybe even convert both sizeofs to double and then divide if we are very generic here

Will do it but only to be ready to use half-precision floats.

sbastrakov (approved):

Issue #1072, opened 2015-09-01
Additional fieldSolverNone necessary with EMF centered in cell?


n01r:
Apart from the fact that field solvers should probably also be `struct`s, just as pushers and ionizers are now:   
*Should there be two (more?) specializations of the `fieldSolverNone` for each `NumericalCellType`?*

Currently this one just has the Yee cell:
```
namespace fieldSolverNone
{
    typedef picongpu::noSolver::NoSolver FieldSolver;
    typedef yeeCell::YeeCell NumericalCellType;
}
```
It might be useful to have the electromagnetic fields centered in the cell when they are defined by the background field and not the usual laser definition.   
I'd propose something like:
```
namespace picongpu
{
namespace fields
{
namespace MaxwellSolver
{
template<T_NumericalCellType>
struct None
{
    typedef fields::numericalCellTypes::T_NumericalCellType NumericalCellType
}
} // namespace MaxwellSolver
} // namespace fields
} // namespace picongpu
```

ax3l:
let's keep thing simple.
it's very convenient that field solvers determine the cell type.

nevertheless, I know some field solvers that could be in principle be implemented for centered and staggered fields.

Using a templated parameter for those (and `fieldSolverNone`) might be a good idea, for the others we could leave the cellType selection in the unitless file.

introducing the celltype in all solvers as an argument might be more consistent, we should evaluate what causes more/less confusion.

ax3l:
for the meantime: simply add the `unitless` file to your case and change it there. that is exactly the use case for it (advanced option that should not (yet) be exposed to non power-uses).

ax3l:
also the documentation is not 100% correct
```
fieldSolverNone: disable the vacuum update of E and B
```

the field solver also determines how `J` influences the change of `E` (and `B`).
`fieldSolverNone` only disables

n01r:
Also:   
`src/picongpu/include/fields/MaxwellSolver/None/NoSolver.hpp`:

```
template<typename T_Dummy>
struct ConditionCheck<NoSolver, T_Dummy>
{
    /* Courant-Friedrichs-Levy-Condition for Yee Field Solver: */
    PMACC_CASSERT_MSG(Courant_Friedrichs_Levy_condition_failure____check_your_gridConfig_param_file,
        (SPEED_OF_LIGHT*SPEED_OF_LIGHT*DELTA_T*DELTA_T*INV_CELL2_SUM)<=1.0);
};
```

[...]

PR #4237, opened 2022-07-29
Better separate CFL and pusher/current deposition requirements

sbastrakov:
This PR does a few related things.

It separates CFL which is a property of field solver from PIConGPU condition that particle passes at most 1 cell per time step. Though somewhat similar, these conditions are not actually related. CFL is a mathematical property of field solver. The latter condition is only required due to our organization of particle pusher and current deposition, it is not a principle limitation of the method (but is crucial for PIConGPU). Also clarify output of these checks.

Fix edge case of checking that particles don't pass more than a cell in a time step: for 2d the previous check then required (among other things) `c * dt < dz`, while it is irrelevant in 2d. Instead, `c * dt < min(dx, dy)` should be, and now is, used in this case.

Fix message on failed CFL for arbitrary order solver with substepping. The check was correct, but output used full dt and not field solver `dt`.

### DIFF
--- a/include/picongpu/fields/FieldJ.tpp
+++ b/include/picongpu/fields/FieldJ.tpp
@@ -43,8 +43,10 @@
 
 #include <boost/mpl/accumulate.hpp>
 
+#include <algorithm>
 #include <cstdint>
 #include <iostream>
+#include <limits>
 #include <memory>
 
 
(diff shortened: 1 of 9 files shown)


sbastrakov:
Some of related things were discussed [here](https://github.com/ComputationalRadiationPhysics/picongpu/issues/1072#issuecomment-138856862)

psychocoderHPC (approved):

PR #4240, opened 2022-07-31
Fix of radiation window for moving simulation.

### LINKED ISSUE #4239: [Radiation plug-in] Weighting window results in 0 radiation for a moving window
In case of a simulation with a moving window the way the radiation window for the suppression of ringing artefacts is applied may result in radiation amplitudes being =0.

`winFkt` in `Radiation.kernel` gets the global particle position as a first argument and the simulation box size as the second. Internally positions are shifted by 1/2 the box size and the weighting is applied.
This works well in case of a stationary domain, but for a simulation with a moving window the global positions will ultimately be larger than the box length in y-direction and the window will return a hard `0` for amplitude values.

Anton-Le:
This introduces a shift of the radiation window `winFkt`
by `globalOffset * cellSize` to shift the radiation window
to the moving window.

This should fix #4239.

### DIFF
--- a/include/picongpu/plugins/radiation/Radiation.kernel
+++ b/include/picongpu/plugins/radiation/Radiation.kernel
@@ -344,7 +344,7 @@ namespace picongpu
                                                 for(uint32_t d = 0; d < simDim; ++d)
                                                 {
                                                     windowFactor *= winFkt(
-                                                        particle_locationNow[d],
+                                                        particle_locationNow[d] - globalOffset[d] * cellSize[d],
                                                         simBoxSize[d] * cellSize[d]);
                                                 }
 


sbastrakov:
I believe `global` in this comment is wrong
https://github.com/ComputationalRadiationPhysics/picongpu/blob/c624c662ea615ed927a9275c4ff3d4786d11dcba/include/picongpu/plugins/radiation/Radiation.kernel#L279-L282
Based on how it's calculated, `particle_locationNow` is a total position, not global.

`globalPos` is also called in a misleading way, it is an index of the current cell in total coordinates, not global.

sbastrakov:
However, this fix seems valid. `particle_locationNow[d]` is in total coordinates and `globalOffset[d] * cellSize[d]` is a shift of global origin from total origin, so the difference is particle position in the global domain - which seems what we need there.

Imprecise naming as I pointed above made it not very clear, so perhaps could also be improved as a second commit to this PR?

Anton-Le:
My semantic understanding of `particle_locationNow`, based on the way it's computed, was that it is the _global_ poosition in the entire simulation, i.e. in-cell + in domain offset + domain offset.

Changing nomenclature is, of course, possible. But maybe it's better done in a separate PR, to keep this one focused on the actual problem. I expect some changes to this PR, since in discussion with @PrometheusPi a **potential problem** came up:

The way the shift is calculated above it may lead to a "jumping" radiation window due to `globalOffset` sliding over GPUs. Unfortunately I have no set-up that could be used to emphasize and evaluate this potential problem. Skipping the `y`-axis for the window function was one suggestion, but it would needlessly impact non-movin-window simulations.

sbastrakov:
> that it is the global poosition in the entire simulation, i.e. in-cell + in domain offset + domain offset.

This way describes calculating both total and global position. The difference is whether the `domain offset` in question is local domain to global domain start or local domain to total domain start. In this case I think it's the latter, so it is a total position at the end.

sbastrakov:
Sure, the naming change is optional and your fix as-is seems fine. I would say the naming was part of the problem since the chosen naming hid the issue. But don't insist, and if there are plans for further change maybe it's better to do then.

psychocoderHPC:
@Anton-Le please rebase against the dev branch to solve the CI issue.

Anton-Le:
I intend to keep this in draft state for the moment to see whether the undulator set-up will suffice as a test.

psychocoderHPC:
@Anton-Le What  ist the status of this PR?

psychocoderHPC:
@Anton-Le Ping!

PrometheusPi:
@psychocoderHPC I will test this PR.

psychocoderHPC:
@PrometheusPi Is this PR working, if so rebase it.

psychocoderHPC:
A new year a new ping ^^.

PrometheusPi:
Thanks for the ping @psychocoderHPC - will have a look asap

chillenzer:
I presume that [this comment here](https://github.com/ComputationalRadiationPhysics/picongpu/issues/4239#issuecomment-1952843194) means that we can close this, @PrometheusPi?

psychocoderHPC:
closed, no activity

Issue #4244, opened 2022-08-01
Checks for current addition to E ignore current background


sbastrakov:
The check if current exists only considers species. However, [this part](https://github.com/ComputationalRadiationPhysics/picongpu/blob/67c50a27a7d417a92f4ae6a596f30fefc6805727/include/picongpu/simulation/stage/CurrentInterpolationAndAdditionToEMF.hpp#L92-L96) should also consider if current background is enabled.

Now with the current background enabled and no relevant species (not macroparticles, but species) PIConGPU would not actually add background current. This bugs seems to exist for about forever. However I doubt it was ever actually encountered as we generally always have species with currents (and if the rare situation and this bug happened the result would be obviously unexpected).

The existance of species should only be a condition for MPI communication. And the correct condition for adding currents should be existance of species or current background.

I will provide a fix shortly, now working on related things and that's how the problem was discovered.

sbastrakov:
Solved with #4247.

PR #4245, opened 2022-08-01
Make adding J term part of field solver

sbastrakov:
It should logically be part of field solver as the solver defines the way it should be done. This PR does not change the results, but prepares for future changes with PML and substepping. Extend comments to reflect the new logic.

- [x] Depends on #4243.

### DIFF
--- a/include/picongpu/fields/MaxwellSolver/Substepping/Substepping.hpp
+++ b/include/picongpu/fields/MaxwellSolver/Substepping/Substepping.hpp
@@ -21,6 +21,7 @@
 
 #include "picongpu/simulation_defines.hpp"
 
+#include "picongpu/fields/FieldJ.hpp"
 #include "picongpu/fields/MaxwellSolver/CFLChecker.hpp"
 #include "picongpu/fields/MaxwellSolver/FDTD/FDTD.hpp"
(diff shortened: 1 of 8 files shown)


include/picongpu/fields/MaxwellSolver/Substepping/Substepping.hpp
```diff
@@ -126,9 +129,31 @@ namespace picongpu
...
+                 * @tparam T_area area to operate on
+                 */
+                template<uint32_t T_area>
+                void addCurrent()
+                {
+                    // Since this is also called internally, check if currents exist in the simulation
+                    using namespace pmacc;
+                    using SpeciesWithCurrentSolver =
+                        typename pmacc::particles::traits::FilterByFlag<VectorAllSpecies, current<>>::type;
+                    constexpr auto numSpeciesWithCurrentSolver = bmpl::size<SpeciesWithCurrentSolver>::type::value;
```
psychocoderHPC, line 143: It was already wrong before but we need to check if a species has a current solver or if field background J is active.
Nothing for this PR, we will fix this in an additional PR.

sbastrakov: Yes, i just made an issue for it yesterday #4244

psychocoderHPC (approved):

PR #4247, opened 2022-08-02
Properly account for current background when deciding to skip adding J

### LINKED ISSUE #4244: Checks for current addition to E ignore current background

sbastrakov:
Fixes #4244 , details are described there.

### DIFF
--- a/include/picongpu/simulation/stage/CurrentInterpolationAndAdditionToEMF.hpp
+++ b/include/picongpu/simulation/stage/CurrentInterpolationAndAdditionToEMF.hpp
@@ -94,8 +94,8 @@ namespace picongpu
                     using SpeciesWithCurrentSolver =
                         typename pmacc::particles::traits::FilterByFlag<VectorAllSpecies, current<>>::type;
                     auto const numSpeciesWithCurrentSolver = bmpl::size<SpeciesWithCurrentSolver>::type::value;
-                    auto const existsCurrent = numSpeciesWithCurrentSolver > 0;
-                    if(existsCurrent)
+                    auto const existsParticleCurrent = numSpeciesWithCurrentSolver > 0;
+                    if(existsParticleCurrent)
                     {
                         DataConnector& dc = Environment<>::get().DataConnector();
                         auto& fieldJ = *dc.get<FieldJ>(FieldJ::getName(), true);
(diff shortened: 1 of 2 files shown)


include/picongpu/simulation/stage/CurrentInterpolationAndAdditionToEMF.hpp
```diff
@@ -127,6 +127,16 @@ namespace picongpu
                             fieldSolver.addCurrent<type::CORE + type::BORDER>();
                         }
                     }
+                    else
+                    {
+                        /* With no current from macroparticles, there is no need for communication.
```
psychocoderHPC, line 132: Let us check this tomorrow together. Field background is only created on `CORE+BORDER` therefore we need to read from the neighbors to have valid data in the guard for the binomial smoothing operation.
Nevertheless, we need to avoid scattering guard data twice into the border, we only need to read border data from neighbors into the guard.
I think 4 eyes should avoid introducing a new bug.

psychocoderHPC (changes requested):

include/picongpu/simulation/stage/CurrentInterpolationAndAdditionToEMF.hpp
sbastrakov: Oh true, let's discuss tomorrow.

sbastrakov: I think what you describe is another bug that is present regardless of thus PR

sbastrakov: Discussed offline, summarized in #4250 . Unrelated to this PR.

psychocoderHPC (approved):

PR #4252, opened 2022-08-03
Disallow binomial current interpolation together with current background

### LINKED ISSUE #4250: Binomial smoothing of J is incompatible with current background
As discussed [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4247#discussion_r935755682).

There are several aspects for it:

- Technical - in current `dev` (and for a very long time) it does not work technically, as after the exchange the `GUARD` values will only have particle-created currents from other domains, but no background. Which is fine with no smoothing, but with `Binomial` smoothing it will be wrong at near-guard points of `BORDER`. This technical issue is in principle possible to fix with reasonable effort, however it's not the only issue.
- Logically smoothing should only be applied to particle currents as a low-pass filter, not to the current background which should be always taken as-is. This is also possible to fix, but perhaps would take a little more time - the reason is that PIConGPU groups together smoothing of J and adding J to E for potential performance benefit. And logically these two should not be grouped, causing issues like this one. Fixing this part would probably involve untying smoothing and adding J to E.
- Binomial smoothing lacks compensator and so should not be used #3883.

A practical solution that we discussed with @psychocoderHPC is to disable binomial smoothing together with current background (not sure if anyone ever used this combination). After we fix the technical and logical issues pointed above it could be enabled again, though it is probably on lower priority side of things.

sbastrakov:
These two options are currently incompatible.
Fixes #4250.

### DIFF
--- a/include/picongpu/simulation/stage/CurrentInterpolationAndAdditionToEMF.hpp
+++ b/include/picongpu/simulation/stage/CurrentInterpolationAndAdditionToEMF.hpp
@@ -59,10 +59,12 @@ namespace picongpu
                  */
                 void registerHelp(po::options_description& desc)
                 {
+                    // Current backround and binomial interpolation are incompatible #4250
+                    auto const options = std::string{"[none"} + (hasCurrentBackground ? "" : ", binomial") + "]";
                     desc.add_options()(
                         "currentInterpolation",
                         po::value<std::string>(&kindName),
-                        std::string("Current interpolation kind [none, binomial] default: " + kindName).c_str());
+                        (std::string{"Current interpolation kind "} + options + " default: " + kindName).c_str());
                 }
 
                 /** Initialize the current interpolation stage
@@ -78,7 +80,13 @@ namespace picongpu
                     if(kindName == "none")
                         interpolation.kind = CurrentInterpolation::Kind::None;
                     else if(kindName == "binomial")
+                    {
+                        // Current backround and binomial interpolation are incompatible #4250
+                        if(hasCurrentBackground)
+                            throw std::runtime_error(
+                                "With current background enabled, only None current interpolation is allowed");
                         interpolation.kind = CurrentInterpolation::Kind::Binomial;
+                    }
                     else
                         throw std::runtime_error("Unsupported current interpolation type");
                 }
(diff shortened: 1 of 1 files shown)


include/picongpu/simulation/stage/CurrentInterpolationAndAdditionToEMF.hpp
```diff
@@ -132,7 +140,7 @@ namespace picongpu
                         /* With no current from macroparticles, there is no need for communication.
                          * However we may still have J from the background (if it is activated) in CORE and BORDER.
                          */
-                        if(FieldBackgroundJ::activated)
+                        if(hasCurrentBackground)
```
psychocoderHPC, line 143: this could become a `if constexpr(hasCurrentBackground)`

psychocoderHPC (changes requested):

include/picongpu/simulation/stage/CurrentInterpolationAndAdditionToEMF.hpp
```diff
@@ -78,7 +80,13 @@ namespace picongpu
                     if(kindName == "none")
                         interpolation.kind = CurrentInterpolation::Kind::None;
                     else if(kindName == "binomial")
+                    {
+                        // Current backround and binomial interpolation are incompatible #4250
+                        if(hasCurrentBackground)
```
psychocoderHPC, line 85: can be a `if constexp()` too

include/picongpu/simulation/stage/CurrentInterpolationAndAdditionToEMF.hpp
sbastrakov: It could, but I don't think it should. Whether `hasCurrentBackground` is `constexpr` or not is not related to this check - it is not required neither for conditional compilation nor for performance here. So in my opinion it only makes this part dependent on `hasCurrentBackground` being `constexpr` for no benefit.

psychocoderHPC (approved):

include/picongpu/simulation/stage/CurrentInterpolationAndAdditionToEMF.hpp
sbastrakov: Yes, and for the same reason I decided against it.

PR #4280, opened 2022-08-26
pmacc//atomic: Fix compile condition for kepler warp opt

jkelling:
When compiling with an offloading compiler (OpenMP5/OpenACC) targeting an NVIDA GPU, `__CUDA_ARCH__` gets defined and hence the kepler warp-optimization branch may be taken. There the function `getLaneId()` is invoked, which is only defined when compiler with a CUDA (or HIP) compiler. This change further restrict the kepler warp-optimization to only be active with CUDA (or HIP) compilers.

### DIFF
--- a/include/pmacc/kernel/atomic.hpp
+++ b/include/pmacc/kernel/atomic.hpp
@@ -100,7 +100,7 @@ namespace pmacc
                 }
             };
 
-#if CUPLA_DEVICE_COMPILE == 1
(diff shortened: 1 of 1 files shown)


sbastrakov (approved):

psychocoderHPC (approved):
It is also possible to update https://github.com/ComputationalRadiationPhysics/picongpu/blob/a2043582635b40292008ca8f05d0ca78b1d0eade/include/pmacc/kernel/warp.hpp#L26 and provide an implementation of `getLane()` if available, but @jkelling fix is fine too.

PR #4087, opened 2022-04-29
Picmi integration

s9105947:
Adds experimental PICMI integration.

Please refer to the (automatically built) docs for details.

Breaks: Moved existing python modules.

Python 3.9 is required (is checked on include)

PR #4282, opened 2022-08-30
Picmi-rebase

BrianMarre:
rebase to current dev of PICMI-integration

see https://github.com/ComputationalRadiationPhysics/picongpu/pull/4087 for original pull request


BrianMarre:
Okay I give up, @psychocoderHPC, validation fails as pyflake complains about in origin namespace imports.
These are  necessary for the test framework to be able to find the tests for pyPIConGPU, but in contrast to flake8, pyflake does not allow disregarding files or lines, see [here](https://stackoverflow.com/questions/5033727/how-do-i-get-pyflakes-to-ignore-a-statement), or at least I did not find a way that actually worked.
We can therefore either filter the pyflake errors or skip the test folder. Your choice.

[Offline Discussion]: we will skip the test folder

BrianMarre:
implemented discussed solution, 
`__init__.py` and `__main__.py` files in the folder test/ have been removed from the CI pyflakes call, all other files are still required to follow the pyflake coding standards.

Note: I have also changed the author Names Tröpgen and Marré to their ASCII equivalents Troepgen and Marre

psychocoderHPC (approved):

share/picongpu/pypicongpu/template/include/picongpu/param/memory.param
```diff
@@ -0,0 +1,117 @@
...
+ */
+
+#pragma once
+#include <pmacc/mappings/kernel/MappingDescription.hpp>
+#include <pmacc/math/Vector.hpp>
+
+#include <array>
+
+
+namespace picongpu
```
psychocoderHPC, line 36: @BrianMarre Why is this file the only file named `.param` and all other ends with `.param.mustache`

BrianMarre: because the render engine does not substitute anything in there yet, no {{ ... }} structure.

I think it exists only as an example of how to include additional param files by hand

PR #4285, opened 2022-09-02
fix png artefacts in particle density

psychocoderHPC:
The png plugin is producing artifacts in the output for the particle density.
The issue was introduced with #4173.
An early exit of the calculation within a supercell without particles skipped the mapping of the density value to the correct color.

example:
e_png_yx_0 5_000780


sbastrakov (changes requested):
Please fix the format as CI failed, otherwise looks good to me.

sbastrakov (approved):

PR #4287, opened 2022-09-05
OpenPMD plugin: add IO file version

psychocoderHPC:
Store a PIConGPU specific major and minor IO format version number to handle incompatibilities of new features or bug fixes.

Old files without the openPMD series attribute `picongpuIOVersionMajor` and `picongpuIOVersionMinor` can be loaded and will be handled as version `0.0`.

### DIFF
--- a/include/picongpu/plugins/common/openPMDVersion.def
+++ b/include/picongpu/plugins/common/openPMDVersion.def
@@ -1,4 +1,4 @@
-/* Copyright 2020-2022 Franz Poeschel
+/* Copyright 2020-2022 Franz Poeschel, Rene Widera
  *
  * This file is part of PIConGPU.
  *
@@ -29,6 +29,26 @@ namespace picongpu
 {
     namespace openPMD
     {
+        /** PIConGPU's IO major file version.
+         *
+         * The version can be used to handle incompatibilities between checkpoint files in case breaking changes
+         * within the openPMD checkpoint code are introduced.
+         * A change in the major version points to a new feature/fix that cannot be handled by an older PIConGPU IO
+         * implementations. Newer PIConGPU IO implementations can optionally support old major versions.
+         *
+         * @attention If the version is changed please update openPMDWriter::checkIOFileVersionCompatibility().
+         */
+        static constexpr int picongpuIOVersionMajor = 0;
+
+        /** PIConGPU's IO minor file version.
+         *
+         * A change in the minor version means that the new introduced feature/fix can be loaded by all PIConGPU IO
+         * implementations with the same major IO file version.
+         *
+         * @attention If the version is changed please update openPMDWriter::checkIOFileVersionCompatibility() if
+         * needed.
+         */
+        static constexpr int picongpuIOVersionMinor = 1;
         /*
          * Do some SFINAE tricks to detect whether the openPMD API has
          * dataset-specific configuration or not.
--- a/include/picongpu/plugins/openPMD/openPMDWriter.hpp
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.hpp
@@ -85,6 +85,7 @@
 #include <algorithm>
 #include <cstdint>
 #include <cstdlib> // getenv
+#include <exception>
 #include <iostream>
 #include <list>
 #include <sstream>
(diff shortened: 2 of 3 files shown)


include/picongpu/plugins/common/openPMDVersion.def
```diff
@@ -29,6 +29,26 @@ namespace picongpu
...
+
+        /** PIConGPU's IO minor file version.
+         *
+         * A change in the minor version means that the new introduced feature/fix can be loaded by all PIConGPU IO
+         * implementations with the same major IO file version.
+         *
+         * @attention If the version is changed please update openPMDWriter::checkIOFileVersionCompatibility() if
+         * needed.
+         */
+        static constexpr int picongpuIOVersionMinor = 0;
```
sbastrakov: Can we start with version 0.1 ? As 0.0 may be confused as not a valid version number (so a user may think the version information itself is incorrect).

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -972,6 +973,79 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                 dumpData(currentStep);
             }
 
+            /** Checks if an loaded openPMD series is supported for restarting.
```
sbastrakov:
```suggestion
            /** Checks if a loaded openPMD series is supported for restarting.
```

include/picongpu/plugins/common/openPMDVersion.def
```diff
@@ -29,6 +29,26 @@ namespace picongpu
 {
     namespace openPMD
     {
+        /** PIConGPU's IO major file version.
+         *
+         * The version can be used to handle incompatibilities between checkpoint files in case breaking changes
+         * within the openPMD checkpoint code are introduced.
+         * A change in the major version points to a new feature/fix which can not be handled by older PIConGPU IO
```
sbastrakov:
```suggestion
         * A change in the major version points to a new feature/fix that cannot be handled by an older PIConGPU IO
```

include/picongpu/plugins/common/openPMDVersion.def
```diff
@@ -29,6 +29,26 @@ namespace picongpu
...
+         * implementations. Newer PIConGPU IO implementations can optionally be support old major versions.
```
sbastrakov:
```suggestion
         * implementations. Newer PIConGPU IO implementations can optionally support old major versions.
```

include/picongpu/plugins/common/openPMDVersion.def
```diff
@@ -29,6 +29,26 @@ namespace picongpu
 {
     namespace openPMD
     {
+        /** PIConGPU's IO major file version.
```
sbastrakov, line 32: Add yourself to author list for this file?

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -972,6 +973,79 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
...
+             *
+             * The openPMD series attribute 'picongpuIOVersionMajor', 'picongpuIOVersionMinor', and the current
+             * PIConGPU IO version is compared to find incompatible versions. In case the loaded series is incompatible
+             * an runtime error is thrown.
+             *
+             * @attention Loading a series with older IO major file version as the current PIConGPU IO version is only
+             * possible if it is explicit allowed by this function e.g. in cases the IO plugin is guaranteeing the
```
sbastrakov:
```suggestion
             * possible if it is explicitly allowed by this function, e.g. in cases the IO plugin is guaranteeing the
```

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -972,6 +973,79 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
...
+                 * minor not zero will be file IO version 1.1.
+                 */
+                int ioVersionUsedFileMajor = 0;
+                int ioVersionUsedFileMinor = 0;
+
+                try
+                {
+                    ioVersionUsedFileMajor = series->getAttribute("picongpuIOVersionMajor").get<int>();
+                }
+                catch(std::runtime_error const&)
```
sbastrakov: Why don't we catch everything? The handler seems not specific to a `std::` exception type.

Same in the `catch` below.

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -972,6 +973,79 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
...
+                    }
+                }
+
+                std::string seriesIoVersion
+                    = std::to_string(ioVersionUsedFileMajor) + "." + std::to_string(ioVersionUsedFileMinor);
+
+                std::string picongpuIoVersion
+                    = std::to_string(picongpuIOVersionMajor) + "." + std::to_string(picongpuIOVersionMinor);
+
+                log<picLog::INPUT_OUTPUT>("openPMD: Restart file IO version is %1% and PIConGPU'S version is %2%")
```
sbastrakov:
```suggestion
                log<picLog::INPUT_OUTPUT>("openPMD: Restart file IO version is %1% and PIConGPU's version is %2%")
```

sbastrakov (changes requested):
I like the addition! Please see some minor comments above

include/picongpu/plugins/common/openPMDVersion.def
psychocoderHPC: Possible.

include/picongpu/plugins/openPMD/openPMDWriter.hpp
psychocoderHPC: I would like to catch `openPMD::no_such_attribute_error` but somehow this is not visible within PIConGPU. I would like to catch only one specific error in cases where the backend throws other error types too. Catching everything will maybe hide different errors from the backend.

include/picongpu/plugins/common/openPMDVersion.def
psychocoderHPC: done

psychocoderHPC:
> I like the addition! Please see some minor comments above

I applied all changes except the exception handling.

include/picongpu/plugins/openPMD/openPMDWriter.hpp
sbastrakov: Oh, so here you are using exceptions just to check if the attribute exists? I think that is confusing and perhaps not the best mechanism for this task. But if there is no other way, fine.

psychocoderHPC: I missed that there is a `openPMD::containsAttribute()` available. I will use this function for the check.

psychocoderHPC: I updated the code and check now for the attribute before loading.

sbastrakov (approved):

PR #4289, opened 2022-09-06
remove Cell2Particle

### LINKED ISSUE #4267: Missing Documentation of CELL2PARTICLE_OPERATOR
The [cell2Particle](https://github.com/ComputationalRadiationPhysics/picongpu/blob/59e9b53605f9a5c1bf271eeb055bc74370a99052/include/picongpu/particles/access/Cell2Particle.tpp#L40) operator defined is missing a in source documentation, also has no out of source documentation I am aware of.
That makes using it difficult.
@psychocoderHPC could you add at least a short one liner describing what it is intended for?

psychocoderHPC:
close #4267

Since there is now a particle foreach in PMacc availble there is no need anymore for the old macro based `Cell2Particle` algorithm.

This PR replaces all usage of `Cell2Particle` with PMacc's in kernel forEachParticle.


include/picongpu/plugins/PhaseSpace/PhaseSpaceFunctors.hpp
```diff
@@ -209,19 +203,23 @@ namespace picongpu
 
             FunctorParticle<r_dir, num_pbins, SuperCellSize> functorParticle;
 
-            particleAccess::Cell2Particle<SuperCellSize, numWorkers> forEachParticleInCell;
-            forEachParticleInCell(
-                acc,
-                /* mandatory params */
-                particlesBox,
+            auto superCellIdx = indexBlockOffset / static_cast<pmacc::math::Int<simDim>>(SuperCellSize::toRT());
```
sbastrakov, line 206: Is the guard size handled correctly here (i mean, incides with guard vs. without)? I was not able to quickly understand it.

psychocoderHPC: Yes the handling is correct and equal to the way it was handled before the refactoring: https://github.com/ComputationalRadiationPhysics/picongpu/pull/4289/files#diff-436b5c29b52b01c3a106e7c1e6c7b3b23befa57765b56b01a7be30e6e7509644L58

sbastrakov (approved):

Issue #4179, opened 2022-06-23
Treating currents created by particles in PML


sbastrakov:
Currently we do not have a good knowledge or experience of how to do it properly, and there are sometimes numerical artifacts arising, which are probably caused by it. On today's seminar @ax3l shared a recent paper on that matter https://arxiv.org/abs/2201.09084 which we should look into.

Edit: since recently they uploaded v2 on arxiv.

sbastrakov:
Now the suggestion from the paper, section II C is implemented in PIConGPU and is now in effect for PML + absorbing particle boundaries. Documentation is updated to reflect it.

In principle, the "main" scheme from section II A & B can also be implemented, assuming some constant velocity (or even calculating e.g. average velocity on-the-fly). As the paper shows, using `v = c` for all particles should also be reasonable. Unfortunately, I had no time to try it out. If someone wants to implement it, here are my thoughts:

- Adding current density to solver already [has an arbitrary coefficient](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/include/picongpu/fields/MaxwellSolver/AddCurrentDensity.hpp#L67).
- For this scheme, this coefficient should be generalized Into a functor that returns a coefficient based on grid node index. So that for current usage the result is constant like now, and for the new scheme it will depend on the position.
- Getting PML implementation/parameters can be done similar to [this piece](https://github.com/ComputationalRadiationPhysics/picongpu/blob/62952124c60208969932c158c2fd1e0362b69d7b/include/picongpu/particles/boundary/Absorbing.hpp#L186-L199).
- Integral (4) in the paper is quite similar to [this part](https://github.com/ComputationalRadiationPhysics/picongpu/blob/62952124c60208969932c158c2fd1e0362b69d7b/include/picongpu/fields/absorber/pml/Pml.kernel#L377-L386). Note that it is only to this part, so a bit easier, as the rest of that function handles cases like particle just entering or just leaving PML which does not need to happen for the new integral.

PR #4293, opened 2022-09-14
Improve treating currents in PML

sbastrakov:
Implements #4179.

As discussed before, opening a PR with the current implementation. It is in principle ready for review and shows improvement over what we have. However, there are still some unclear things e.g. on a test problem our results are worse than in the paper but also field sovler is different (they used what we don't have). Ideally I would like to spend a couple more days investigating, while the code as-is appears mergeable once the following TODOs are completed.

### Test problem

Follows section I of [the paper](https://arxiv.org/abs/2201.09084v2), setup is attached. I made the images same way as they did in Fig. 1 and tried to match the colorbar in front row of the attached images. The second row is same data but on a wider range, as the original image is oversaturated.

absorber_100
absorber_150
absorber_180
absorber_210

For us the new scheme gives some improvement, but not as much as in the paper. It is unclear what is the reason - they also use another field solver and current smoothing (I also enabled it, but we have no compensation).

### TODOs:
[AbsorberSingleParticleTest.zip](https://github.com/ComputationalRadiationPhysics/picongpu/files/9574645/AbsorberSingleParticleTest.zip)

- [x] Add test problem results (have it, but no more time today)
- [x] Clean up git history
- [x] Resolve formatting issue I got today

Additionally it makes sense to store the changed weighting or the weighting multiplier separately. This is so that density etc. is not affected by this procedure, but at the cost of storing extra data per particle.

### DIFF
--- a/include/picongpu/particles/boundary/Absorbing.hpp
+++ b/include/picongpu/particles/boundary/Absorbing.hpp
@@ -37,25 +37,52 @@ namespace picongpu
     {
         namespace boundary
         {
-            //! Functor to be applied to all particles in the active area
-            struct AbsorbParticleIfOutside : public functor::misc::Parametrized<Parameters>
+            namespace detail
             {
-                //! Name, required to be wrapped later
-                static constexpr char const* name = "absorbParticleIfOutside";
-
-                /** Process the current particle located in the given cell
-                 *
-                 * @param offsetToTotalOrigin offset of particle cell in the total domain
-                 * @param particle handle of particle to process (can be used to change attribute values)
-                 */
-                template<typename T_Particle>
-                HDINLINE void operator()(DataSpace<simDim> const& offsetToTotalOrigin, T_Particle& particle)
+                //! Functor to be applied to all particles in the active area
+                struct AbsorbParticleIfOutside : public functor::misc::Parametrized<Parameters>
                 {
-                    if((offsetToTotalOrigin[m_parameters.axis] < m_parameters.beginInternalCellsTotal)
-                       || (offsetToTotalOrigin[m_parameters.axis] >= m_parameters.endInternalCellsTotal))
-                        particle[multiMask_] = 0;
+                    //! Name, required to be wrapped later
+                    static constexpr char const* name = "absorbParticleIfOutside";
+
+                    /** Process the current particle located in the given cell
+                     *
+                     * @param offsetToTotalOrigin offset of particle cell in the total domain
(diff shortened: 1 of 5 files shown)


include/picongpu/particles/boundary/Absorbing.hpp
```diff
@@ -83,6 +89,157 @@ namespace picongpu
...
+                        auto const timeStepIntegral = fields::absorber::pml::getNormalizedSigmaIntegral(
+                            positionPrevious,
+                            positionCurrent,
+                            DELTA_T,
+                            m_parameters.localParameters,
+                            axis);
+
+                        // Adjust weighting according to [Lehe2022], cap at MIN_WEIGHTING
+                        auto const weightingMultiplier = math::exp(-timeStepIntegral);
+                        auto const dampedWeighting = float_64(particle[weighting_]) * weightingMultiplier;
```
psychocoderHPC: is 64bit required ?

sbastrakov: I was not sure here and it is not performance-critical

sbastrakov: I checked - it is fine as `float_X`, changed to that.

include/picongpu/particles/boundary/Absorbing.hpp
```diff
@@ -83,6 +90,167 @@ namespace picongpu
...
+                        // Adjust weighting according to [Lehe2022], cap at MIN_WEIGHTING
+                        auto const weightingMultiplier = math::exp(-timeStepIntegral);
+                        auto const dampedWeighting = particle[weighting_] * weightingMultiplier;
+                        auto const minWeighting = MIN_WEIGHTING; // to avoid taking address of constexpr in max()
+                        /* Update macroparticle weighting and values of all weighted attributes.
+                         * We currently have no generic and consistent way to do that #4299.
+                         * So at least update weighted momentums so that particles don't artificially accelerate.
+                         */
+                        particle[weighting_] = math::max(dampedWeighting, minWeighting);
+                        particle[momentum_] *= weightingMultiplier;
```
psychocoderHPC, line 167: This modification must take into account that we change the weighting only to a value larger than `minWeighting`.

psychocoderHPC: discussed offline @sbastrakov will provide a function in a separate PR to sett he weighing which is then fixing this problem too.

psychocoderHPC (approved):

PR #4295, opened 2022-09-16
Update binary collisions, dynamic Coulomb log

pordyna:
# Update to the binary collisions
This PR introduces a dynamic calculation of the Coulomb logarithm for binary collisions. It also includes some fixes for the existing implementation. These changes continue to follow the implementation from `smilei`. @ComputationalRadiationPhysics/picongpu-maintainers @psychocoderHPC @HighIander 

## Dynamic Coulomb logarithm
The newly introduced automatic Coulomb log calculation is based on the 2012 paper from Perez [link](http://aip.scitation.org/doi/10.1063/1.4742167). The calculation uses a Debye length that is calculated on every time step from all charged particles (the list of included species is set in `collision.param`) for all simulation cells. These values are made available to the colliders via a `FieldTmp`. The collision functor has now two implementations: one for a constant and one for a dynamic Coulomb logarithm. More details are in the updated documentation.

## Other updates
- A factor of two difference in the s-parameter between the inter- and intra-collisions was removed. Originally, this difference comes from the 2020 Higginson paper [link](https://doi.org/10.1016/j.jcp.2020.109450), but it seems to be a typo. This factor two is not present in smilei.
- The sampling of the cosine of the cumulative scattering angle was updated to follow the new fit recently introduced in `smilei`.
- A debug output was introduced. It can write average values (averaged over the simulation) for the coulomb log, the s-paramer (proportional to the collision frequency) per collider and species pair. It is also possible to write an average value of the precalculated Debye length. It would be probably useful to be able to write these value per cell, but, for now, the averaging over the simulation is good enough for the homogenous test cases.

## Tests
The existing collision tests setups were extended to include the new automatic coulomb logarithm. They were also run for a longer time and include now a direct comparison with `smilei` values and some statistical analysis. Also, the tests use now only field output, so they run faster. A python tool for generating the comparison plots was also included for both tests. 

The test results show in general quite good agreement with `smilei. ~but there still seems to be some difference in the dynamic debye length and the coulomb log, for example in this beam relaxation set-up~:
EDIT: Now the debye length calculation was fixed.
#### Old results
beam_relaxation_SetupNr0_dynamic
beam_relaxation_SetupNr0_dynamic_debug
#### After fix
BR_main_3
BR_debug_3

~or this thermalization test:~
#### old results
thermalisation_1to1_all_dynamic
thermalisation_1to1_all_dynamic_debug
#### after fix:
TH_main_3
TH_debug_3

All the other combinations of particle weighting ratios, and coulomb logarithms are included in the updated docs.


psychocoderHPC:
The CI is not happy with your formation.

include/picongpu/fields/FieldTmp.tpp
```diff
@@ -198,6 +198,26 @@ namespace picongpu
...
     }
 
+    template<uint32_t AREA, bool sumSqrt>
+    float_X FieldTmp::calculateSum()
+    {
+        constexpr uint32_t numWorkers
+            = pmacc::traits::GetNumWorkers<pmacc::math::CT::volume<SuperCellSize>::type::value>::value;
+        auto mapper = makeAreaMapper<AREA>(cellDescription);
+        FieldTmp::DataBoxType thisBox = this->fieldTmp->getDeviceBuffer().getDataBox();
+        using Kernel = SumValuesKernel<numWorkers, MappingDesc::SuperCellSize, sumSqrt>;
```
psychocoderHPC: discussed with @pordyna offline: There is no need to write an own reduce we have a global and local reduce in PMacc.
With using the already existing reduce changes in `ThreadCollective` and `FieldTmp` should be not nessesary.

psychocoderHPC (changes requested):

include/picongpu/particles/collision/Collider.hpp
```diff
@@ -37,21 +37,42 @@ namespace picongpu
...
+                {
+                    template<size_t... I>
+                    HINLINE void operator()(
+                        std::index_sequence<I...>,
+                        std::shared_ptr<DeviceHeap> const& deviceHeap,
+                        uint32_t currentStep)
+                    {
+                        (collision::WithPeer<
+                             ApplyGuard<typename T_Collider::Functor>,
+                             typename bmpl::at_c<T_SpeciesPairList, I>::type::first_type,
```
psychocoderHPC: in `Collider.def` `boost::mpl::pair` is replaced by `std::pair`. I am not sure if `boost::mpl::at_c` can/should still be used here.

include/picongpu/particles/collision/Collider.def
```diff
@@ -36,22 +36,22 @@ namespace picongpu
...
-            using FilterPair = boost::mpl::pair<T_Filter0, T_Filter1>;
+            using FilterPair = std::pair<T_Filter0, T_Filter1>;
 
             /* Sets the same particle filter for both species in a pair.
              *
              * @tparam T_Filter A common particle filter for both colliding species.
              */
             template<typename T_Filter>
-            using OneFilter = boost::mpl::pair<T_Filter, T_Filter>;
+            using OneFilter = std::pair<T_Filter, T_Filter>;
```
psychocoderHPC: Not sure if this change should be part of this PR. It looks like it should be an independent PR.

psychocoderHPC (changes requested):

pordyna:
@psychocoderHPC as discussed offline, we should still try to investigate the difference in Debye length calculation. In the thermalization test the contributions from electrons and ions are similar, while in the beam relaxation test the ion contribution gets bigger overtime while ions gain energy. So, I think that the problem could be in ion contribution, but I don't know why yet.

pordyna:
@psychocoderHPC: fixed the missing absolute value, fixed the CI errors, updated the documentation with the new plots

psychocoderHPC:
@pordyna Could it be that you missed updating the plots in the first post of this PR? We should update those too to reflect the current state of the PR.

include/picongpu/simulation/stage/Collision.hpp
```diff
@@ -47,19 +66,127 @@ namespace picongpu
...
+                        // Invert the calculated value and apply the cut-off to get the squared debye length
+                        screeningLengthSquared->modifyByField<CORE + BORDER, GetSquaredScreeningLength>(*maxDensity);
+                        // the FieldTmp slot used for storing  the max density is not longer needed beyond this point
+                        maxDensity.reset();
+
+                        // write Debye length averaged over the whole simulation volume to a text file for debugging
+                        if constexpr(particles::collision::debugScreeningLength)
+                        {
+                            // calculate the average of square roots
+                            float_X sumRank = screeningLengthSquared->calculateSum<CORE + BORDER, true>();
```
psychocoderHPC: Please use `GolbalRecude` instead of the hand crafted version:

```C++
algorithms::GlobalReduce reduce(1024);
using D1Box = DataBoxDim1Access<typename GridBuffer<float3_X, simDim>::DataBoxType>;
D1Box d1access(buffer->getDeviceBuffer().getDataBox(), buffer->getGridLayout().getDataSpace());
int elements = buffer->getGridLayout().getDataSpace().productOfComponents();
auto result = reduce(pmacc::math::operation::Add(), d1access, elements);
```

If you need on device a handwritten functor you must split the command in a [local reduce](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/include/pmacc/device/Reduce.hpp) and pass the result into an [MPI reduce](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/include/pmacc/mpi/MPIReduce.hpp) with the operation `Add`

include/picongpu/fields/FieldTmp.tpp
```diff
@@ -198,6 +198,26 @@ namespace picongpu
         PMACC_KERNEL(Kernel{})(mapper.getGridDim(), numWorkers)(mapper, thisBox, modifyingBox);
     }
 
+    template<uint32_t AREA, bool sumSqrt>
+    float_X FieldTmp::calculateSum()
```
psychocoderHPC: This can be removed by [local reduce](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/include/pmacc/device/Reduce.hpp)

include/pmacc/mappings/threads/ThreadCollective.hpp
```diff
@@ -30,6 +30,15 @@
 
 namespace pmacc
 {
+    struct DefaultApplyOffset
```
psychocoderHPC: Changes in this file are not required if pmacc reduction method will be used.
offline discussed that this was used as a workaround to write its own reduction method.

psychocoderHPC (changes requested):

pordyna:
@psychocoderHPC I updated the first post

pordyna:
@psychocoderHPC If the CI passes, this PR is ready for review. I will rerun all tests once we are done with the review / introducing new changes.

include/pmacc/math/operation/SqrtAdd.hpp
```diff
@@ -0,0 +1,49 @@
...
+    {
+        namespace operation
+        {
+            //! Sum square roots
+            struct SqrtAdd
+            {
+                template<typename Dst, typename Src>
+                HDINLINE void operator()(Dst& dst, const Src& src) const
+                {
+                    dst += math::sqrt(src);
```
psychocoderHPC: this shoudl be `cupla::math::sqrt`

Sry I missed yesterday that this function in pmacc and not picongpu

include/pmacc/math/operation/SqrtAdd.hpp
```diff
@@ -0,0 +1,49 @@
...
+                }
+
+                template<typename Dst, typename Src, typename T_Acc>
+                HDINLINE void operator()(const T_Acc&, Dst& dst, const Src& src) const
+                {
+                    dst += math::sqrt(acc, src);
```
psychocoderHPC: `cupla::math::sqrt(acc, src)`

psychocoderHPC (changes requested):

psychocoderHPC:
@pordyna  Why is in [TH_debug_7.png](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4295/files#diff-337b95266d681ca18db99e450987bcb2a2c32048787c4f626c7b9bfdf0f4907e) the right plot for PIConGPU and smile differently?

include/picongpu/param/collision.param
```diff
@@ -32,6 +32,17 @@ namespace picongpu
...
+            // needs to be enabled if a collider with dynamic coulomb log calculation is used
+            constexpr bool calculateScreeningLength = false;
+            // write average (over all cells) Debye length for all timesteps to a file
+            constexpr bool debugScreeningLength = false;
+
+            /** List of (filtered) species contributing to the screening (Debye) length calculation
+             *
+             * Elements must be either a species types or a misc::SpeciesFilter specialization.
+             */
+            using CollisionScreeningSpecies = VectorAllSpecies;
```
psychocoderHPC: The list should be by default empty to reduce the compile time. You can add an example into the docs how to enable

```
     * @code{.cpp}
     *   using CollisionScreeningSpecies = VectorAllSpecies;
     * @endcode
```

```C++
using CollisionScreeningSpecies = bmpl::vector0<>;
// maybe this is possible too I am not sure
// using CollisionScreeningSpecies = MakeSeq_t<>;
```

include/picongpu/simulation/stage/Collision.hpp
```diff
@@ -47,19 +67,135 @@ namespace picongpu
...
                  */
-                void operator()(uint32_t const step) const
+                void operator()(uint32_t const currentStep) const
                 {
-                    pmacc::meta::ForEach<
-                        particles::collision::CollisionPipeline,
-                        particles::collision::CallCollider<bmpl::_1>>{}(m_heap, step);
+                    // Calculate squared Debye length using the formula form [Perez 2012].
+                    // A species temperature is assumed to be (2/3)<E_kin>
+                    if constexpr(particles::collision::calculateScreeningLength)
```
psychocoderHPC: Could you please move the full body including the if condition into a member method `debug` to make this method more readable?
This would visible separate the debug part from the real code usage.

share/picongpu/tests/CollisionsBeamRelaxation/etc/picongpu/1.cfg
```diff
@@ -30,14 +30,14 @@
 ## Section: Required Variables ##
 #################################
 
-TBG_wallTime="2:00:00"
+TBG_wallTime="24:00:00"
```
psychocoderHPC, line 33: are these changes required for this PR?

psychocoderHPC: I am fine with the changes but wondering why these are required.

share/picongpu/tests/CollisionsBeamRelaxation/include/picongpu/param/memory.param
```diff
@@ -102,7 +102,7 @@ namespace picongpu
     };
 
     /** number of scalar fields that are reserved as temporary fields */
-    constexpr uint32_t fieldTmpNumSlots = 1;
+    constexpr uint32_t fieldTmpNumSlots = 3;
```
psychocoderHPC, line 105: Why do we need 3 slots?

If one of the collision parts requires 3 solts please do not forget to add a static assert to the code snipped to guard against user configuration mistakes.

```
PMACC_CASSERT_MSG(_please_allocate_at_least_three_FieldTmp_in_memory_param, fieldTmpNumSlots >= 3);
```

psychocoderHPC (changes requested):

share/picongpu/tests/CollisionsBeamRelaxation/etc/picongpu/1.cfg
pordyna: the previous 2h were not always enough and now the examples are also longer (more steps)

include/pmacc/math/operation/SqrtAdd.hpp
pordyna: This also won't always work. I switched now to `DataBoxUnaryTransform` for `sqrt` and removed `AddSqrt`. I hope this will work fine.

pordyna:
CI has passed, I will rerun the tests

pordyna:
@psychocoderHPC tests results didn't change

psychocoderHPC (approved):

PR #4296, opened 2022-09-16
Fix calls to makeForEach in particle merger plugins

sbastrakov:
The change introduced in #4289 resulted in compile errors at these places. This went unnoticed as we do not compile those plugins in the CI.

Bug reported by @PrometheusPi via the helpline channel.

I will add it to CI in a follow-up PR.

### DIFF
--- a/include/picongpu/plugins/particleMerging/ParticleMerger.kernel
+++ b/include/picongpu/plugins/particleMerging/ParticleMerger.kernel
@@ -116,10 +116,11 @@ namespace picongpu
                     const uint32_t workerIdx
                         = DataSpaceOperations<simDim>::template map<SuperCellSize>(cellIdx % SuperCellSize::toRT());
 
+                    auto const superCellIdx = pmacc::DataSpace<simDim>{cellIdx / SuperCellSize::toRT()};
                     auto forEachParticle = pmacc::particles::algorithm::acc::makeForEach<numWorkers>(
                         workerIdx,
                         particlesBox,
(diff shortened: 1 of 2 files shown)


include/picongpu/plugins/particleMerging/ParticleMerger.kernel
```diff
@@ -116,10 +116,11 @@ namespace picongpu
                     const uint32_t workerIdx
                         = DataSpaceOperations<simDim>::template map<SuperCellSize>(cellIdx % SuperCellSize::toRT());
 
+                    auto const superCellIdx = pmacc::DataSpace<simDim>{cellIdx / SuperCellSize::toRT()};
                     auto forEachParticle = pmacc::particles::algorithm::acc::makeForEach<numWorkers>(
                         workerIdx,
                         particlesBox,
-                        cellIdx / static_cast<pmacc::math::Int<simDim>>(SuperCellSize::toRT()));
+                        superCellIdx);
```
sbastrakov, line 123: I am not sure why the type cast of `SuperCellSize::toRT()` was there, but at least now it seems to work without the cast as well, so removed it.

PrometheusPi (approved):
I can confirm that @sbastrakov solution solves the compile error.

include/picongpu/plugins/particleMerging/ParticleMerger.kernel
psychocoderHPC: The cast was in because CUSTL assumes `math::Vector` but looks like it is working without an cast too.

psychocoderHPC (approved):

PR #4298, opened 2022-09-21
Fix handling of last updated cell in incident field

sbastrakov:
The code was mistakenly checking for a locally last instead of globally last cell. Thus, the results were wrong for multiple local domains along an axis, given the source was also not zero in those cells (which it often was, but not always).

Thanks to @finnolec for reporting the bug (via the Helpline channel).

### DIFF
--- a/include/picongpu/fields/incidentField/Solver.hpp
+++ b/include/picongpu/fields/incidentField/Solver.hpp
@@ -297,6 +297,13 @@ namespace picongpu
                     functor.inCellShift1 = incidentFieldBaseShift + incidentFieldPositions[functor.incidentComponent1];
                     functor.inCellShift2 = incidentFieldBaseShift + incidentFieldPositions[functor.incidentComponent2];
 
+                    // Set local domain information
+                    auto& gridController = pmacc::Environment<simDim>::get().GridController();
+                    auto const localDomainIdx = gridController.getPosition();
+                    auto const numLocalDomains = gridController.getGpuNodes();
+                    for(uint32_t d = 0; d < simDim; d++)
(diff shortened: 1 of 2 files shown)


finnolec:
I have rerun my simulations and it looks like it works now. 

## Before this PR
Previously I had a "broken" Gaussian shape here when using 2 GPUs in x-direction:
dev_Gausslaser, 1 GPU in x_7
(the laser is propagating into the screen in these colormeshes, I can explain in more detail if that's required)
dev_Gausslaser, 2 GPU in x_7
This here is the time evolution of a line-out from the colormesh:
lineout_dev_Gausslaser, 2 GPU in x

## With this PR

Now my Gaussian laser actually looks like a Gaussian laser:
dev_Gausslaser, 1 GPU in x_7_4298_x
dev_Gausslaser, 2 GPU in x_7_4298_x
lineout_dev_Gausslaser, 2 GPU in x_4298_x

I did **not** look at the changed code in this PR.

@sbastrakov and I discussed offline, that the maximum field amplitude was a bit off (~1%)  previously and I think this is still the case in my naive `np.max(sqrt(E^2))` approach, I'm not sure right now if this is actually an issue.

finnolec:
> @sbastrakov and I discussed offline, that the maximum field amplitude was a bit off (~1%) previously and I think this is still the case in my naive np.max(sqrt(E^2)) approach, I'm not sure right now if this is actually an issue.

We talked about it again and we were talking about different things. He had different field amplitudes with different GPU setups, my difference is the difference to the expected maximum of the envelope which should be a bit higher than my so measured maximum. So I don't think this is a problem

psychocoderHPC (approved):

PR #4300, opened 2022-09-22
Add a new particle attribute for weighting damping factor 

sbastrakov:
Follow-up to #4293. It adds a new optional particle attribute to more precisely apply that procedure. Namely, the damping itself is exactly the same. The "added value" in this PR is that in case this attribute is defined the density calculations and all other parts of the code will be not affected by damping. And if not, it will all work the same as in #4293 just with [this inaccuracy](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4293#discussion_r977681034) fixed and a more clear expalantion. I believe there should be no reason to enable the new attribute for most simulations. And an exception would be if lots of particles travel from PML to internal area.

- [x] Merge after #4293.

### DIFF
--- a/include/picongpu/fields/FieldJ.kernel
+++ b/include/picongpu/fields/FieldJ.kernel
@@ -26,6 +26,7 @@
 #include "picongpu/fields/FieldJ.hpp"
 #include "picongpu/fields/currentDeposition/Cache.hpp"
 #include "picongpu/fields/currentDeposition/Strategy.def"
+#include "picongpu/traits/attribute/DampedWeighting.hpp"
 
 #include <pmacc/dimensions/DataSpaceOperations.hpp>
 #include <pmacc/lockstep.hpp>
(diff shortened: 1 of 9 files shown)


include/picongpu/fields/FieldJ.kernel
```diff
@@ -131,12 +132,17 @@ namespace picongpu
             template<typename T_Particle, typename BoxJ, typename T_Acc>
             DINLINE void operator()(T_Acc const& acc, T_Particle& particle, BoxJ& jBox)
             {
-                const float_X weighting = particle[weighting_];
+                /* Use (potentially) damped weighting for charge calculation.
+                 * @see particles::boundary::DampWeightsInPml for details.
+                 */
+                const auto dampedWeighting = attribute::getDampedWeighting(particle);
```
psychocoderHPC, line 138: Maybe we should add an optimized function in a separate PR where we return the damped and nondamped at the same moment to avoid that we touch the global memory for the same value twice in case there is no damping factor and three times if there is a damping value.

psychocoderHPC (approved):

include/picongpu/fields/FieldJ.kernel
sbastrakov: I think it should not really matter given the overall number of operations in current deposition. However, in principle indeed such a function can be written to load both a damped and undamped weight and do only minimal required number of global memory reads.

PR #4279, opened 2022-08-24
Dispersive Laser

fabidie:
The new profile "DispersiveLaser" allows to simulate laser pulses with dispersive properties like AD, SD, GDD and TOD. One can set these parameters in DispersiveLaser.def. The incident E field is calculated via Fourier transformation and the B field by the already existing "approximateB" method in PiconGPU, using SVEA. 
There is another possibility implemented to calculate the incident B field, also via DFT, but it is still buggy and leads to wrong results, that is why it is not in use yet (Klaus said to me he will fix this).

Here some proof that this profile works (with approximateB als incident Field):

For the following simulation output I used lambda = 0.8µm, focus position in propagation direction at 2e-5m, w0 = 5µm / 1.17741 and tau = 10fs / 1.17741

- setting AD to -2.45e-16 rad*s, which should result in a pulse front tilt of around 30° in focus
AD_30_focus
AD_30_zR

All other dispersion parameters have been tested too and lead to satisfying results aswell.

- comparing GaussianBeam profile with DispersiveLaser profile whilest setting all dispersion parameters to zero leads to similar results:
vgl_DL_GB_all
vgl_DL_GB_slice

PR #4302, opened 2022-09-26
Add DispersiveLaser incident field profile

steindev:
This PR adds the possibility to feed Gaussian laser pulses with dispersion properties, i.e. angular, spatial, group-delay, and third-order dispersion, in a simulation.

Replaces #4279. This PR is a refactor of the original PR by @fabidie. It incorporates the changes as requested by the reviewers in the original PR. See the description of #4279 for details and tests.

@fabidie is the author of the commit added to PIConGPU such that her contribution is properly acknowledged.

Co-authored-by: Klaus Steiniger <k.steiniger@hzdr.de>

The refactor of the non-approximated B-Field will be provided in a separate PR.
That is, with the current implementation of the the DispersiveLaser, laser pulses with a duration of the order of a few femtoseconds or shorter are most probably not correctly modeled.

### DIFF
--- a/include/picongpu/fields/incidentField/profiles/DispersiveLaser.def
+++ b/include/picongpu/fields/incidentField/profiles/DispersiveLaser.def
@@ -1,0 +1,107 @@
+/* Copyright 2013-2022 Axel Huebl, Heiko Burau, Anton Helm, Rene Widera,
+ *                     Richard Pausch, Alexander Debus, Sergei Bastrakov,
+ *                     Fabia Dietrich, Klaus Steiniger
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/fields/incidentField/profiles/BaseParam.def"
+
+namespace picongpu
+{
+    namespace fields
+    {
+        namespace incidentField
+        {
+            namespace profiles
+            {
+                namespace defaults
+                {
+                    struct DispersiveLaserParam : public BaseParam
+                    {
+                        /** Beam waist: distance from the axis where the pulse intensity (E^2)
+                         *              decreases to its 1/e^2-th part,
+                         *              at the focus position of the laser
+                         * W0_SI = FWHM_of_Intensity / sqrt{ 2* ln(2) }
+                         *                             [   1.17741    ]
+                         *
+                         * unit: meter
+                         */
+                        static constexpr float_64 W0_SI = 5.0e-6 / 1.17741;
+
+                        /** The laser pulse will be initialized PULSE_INIT times of the PULSE_LENGTH
+                         * WATCH OUT! Dispersion parameters may lead to pulse lenghtening
+                         * Please ensure to choose a value high enough to cover the whole pulse,
+                         * otherwise the Fourier Transformation will misbehave.
+                         *
+                         *  unit: none
+                         */
+                        static constexpr float_64 PULSE_INIT = 20.0;
+
+                        // Dispersion Parameters
+
+                        /** SD: spatial dispersion in focus
+                         * = d x_0 / d Omega
+                         *
+                         * unit: m * s
+                         */
+                        static constexpr float_64 SD_SI = 0.0;
+
+                        /** AD: angular dispersion in focus
+                         * d theta / d Omega, e.g.
+                         * AD = tan(alpha_tilt) / Omega_0
+                         *
+                         * unit: rad * s
+                         */
+                        static constexpr float_64 AD_SI = 0.0;
+
+                        /** GDD: Dispersion of group velocity in focus
+                         *     = (d^2 phi / d Omega^2)
+                         * with phi being the spectral laser phase.
+                         * The pulse's electric field in frequency domain is therefore
+                         * ~ exp(-i * 1/2 * GDD*(Omega - Omega_0)^2)
+                         *
+                         * unit: s^2
+                         */
+                        static constexpr float_64 GDD_SI = 0.0;
+
+                        /** TOD: third order dispersion in focus
+                         *     = (d^3 phi / d Omega^3)
+                         * with phi being the spectral laser phase.
+                         * The pulse's electric field in frequency domain is therefore
+                         * ~ exp(-i * 1/6 * TOD*(Omega - Omega_0)^3)
+                         *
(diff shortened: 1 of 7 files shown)


PrometheusPi:
@steindev I now restarted the CI several times and it still crashes - I thing there is something off in your branch.

psychocoderHPC:
@steindev What is the difference between this laser and our incident field profile for the gaussian laser?

include/picongpu/fields/incidentField/profiles/DispersiveLaser.hpp
```diff
@@ -0,0 +1,546 @@
...
+                         * @param currentStep current time step index, note that it is fractional
+                         * @param unitField conversion factor from SI to internal units,
+                         *                  fieldE_internal = fieldE_SI / unitField
+                         */
+                        HINLINE DispersiveLaserFunctorIncidentE(float_X const currentStep, float3_64 const unitField)
+                            : Base(currentStep, unitField)
+                        {
+                        }
+
+                        /** Calculate incident field E value for the given position
```
finnolec, line 109: Would it be helpful to quote some kind of source material for this implementation as it is down in [other](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/include/picongpu/fields/incidentField/profiles/GaussianBeam.hpp#L177) types of incident field lasers?

include/picongpu/param/incidentField.param
```diff
@@ -22,6 +22,8 @@
  * Configure incident field profile and offset of the Huygens surface for each boundary.
  *
  * Available profiles:
+ *  - profiles::DispersiveLaser<>     : Gaussian pulse allowing to set first-, second-, and third-order dispersion
```
finnolec, line 25: This is a more general remark: Do we want to call the dispersive laser `DispersiveLaser`? The Gaussian laser is not called `GaussianLaser`, but `GaussianBeam`... similar with the other laser profiles. Maybe it will be counterintuitive if new users only find a dispersive laser, but no other lasers

finnolec:
> @steindev What is the difference between this laser and our incident field profile for the gaussian laser?

We had an offline discussion, both lasers are the same if you set the Laguerre modes in the Gaussian implementation and SD, AD, GDD, and TOD in the dispersive implementation to zero. If any of these parameters are not equal to zero, the laser can't be replaced by the other implementation, so both lasers should be kept.

include/picongpu/fields/incidentField/profiles/DispersiveLaser.def
```diff
@@ -0,0 +1,107 @@
...
+                         * W0_SI = FWHM_of_Intensity / sqrt{ 2* ln(2) }
+                         *                             [   1.17741    ]
+                         *
+                         * unit: meter
+                         */
+                        static constexpr float_64 W0_SI = 5.0e-6 / 1.17741;
+
+                        /** The laser pulse will be initialized PULSE_INIT times of the PULSE_LENGTH
+                         * WATCH OUT! Dispersion parameters may lead to pulse lenghtening
+                         * Please ensure to choose a value high enough to cover the whole pulse,
```
finnolec, line 50: Would it make sense to provide a small python script that tells you the minimal required `PULSE_INIT` for a given set of dispersion parameters and a given time step etc, so that the users don't need to figure it out themselves?

finnolec: Additionally, I think it would be useful if we have separate parameters for the size of the DFT and the initialization delay. If someone wants to use these lasers for e.g. optical probing at later timesteps in the simulation, it is a bad idea to be forced to increase the size of the DFT from the dispersive laser implementation... But this might be another issue/pull request, since it also affects other incident field laser implementations. The only way to implement a time-delay with the `GaussianBeam` laser is the respective `PULSE_INIT`, which is not really the most convenient variable for this usage.

psychocoderHPC:
@steindev Please notify me if all changes are in. I am not sure if you would like to rename `DispersiveLaser` to `DispersiveBeam` as suggested by @finnolec.

include/picongpu/fields/incidentField/profiles/DispersiveLaser.hpp
steindev: No documentation available. Publication pending. Will add this once it is available.

include/picongpu/param/incidentField.param
steindev: Well, your remark is absolutely correct. But from my point of view, both names are inaccurate. We simulate laser pulses, which is not reflected by either name. Would it be ok for you, if I rename all lasers pulses later, just before the new release? That is, 'GaussianBeam' -> 'GaussianPulse', 'DispersiveLaser' -> 'DispersiveGaussianPulse', etc.

However, this will break compatibility to old setups, and we should consider whether the renaming is worth breaking downward compatibility. Then again, with the change to incident field as the preferred laser initialization method and deprecation of the old method with the next release, we have such a break anyways, so it may be just the right time to do this renaming.

steindev: If you agree, I will open an issue and assign to myself.

include/picongpu/fields/incidentField/profiles/DispersiveLaser.def
steindev: Absolutely agree, we should introduce a variable 'PULSE_DELAY' or so in all laser fields, to be able to delay the laser pulse arrival in the simulation box with respect to simulation start.
But this should be part of a separate PR.

If you agree, I will open an issue and assign to myself.

include/picongpu/param/incidentField.param
finnolec: I'm not sure what the best solution for the naming convention is, but it sounds like a good idea to generally refactor the names with the next release. For this PR the name `DispersiveLaser` might just be fine then.

include/picongpu/fields/incidentField/profiles/DispersiveLaser.def
finnolec: This would be great 👍🏻

finnolec: It would be great to have a separate PR for the `PULSE_DELAY` feature. But I think it is still a problem, that `PULSE_INIT` is used for 2 different things in this implementation and it would be better to have a separate variable for the size of the DFT.

finnolec:
I ran some test simulations to see if the dispersive laser behaves weird under different propagation angles, but it looks fine in my simulations 👍🏻

include/picongpu/fields/incidentField/profiles/DispersiveLaser.def
steindev: Independent of the introduction of `PULSE_DELAY` in a separate PR, `PULSE_INIT` and the number of points in the DFT are inseparable. `PULSE_INIT` defines the size of a time window within which we need to know the field values at specified instances of time given by the time step `DELTA_T`. The size of the window and time step together fix the required total number of samples in time domain, which by definition of the Fourier transform is exactly the number of samples required in frequency domain.

steindev: @finnolec if my last comment is a sufficient response to your comment, please mark this conversation as resolved and notify psychocoderHPC.

steindev:
@psychocoderHPC All remaining points were discussed with @finnolec. And a few issues opened, since properly resolving some of the questions requires refactorings beyond the scope of this PR. If you are fine with the PR, you can merge.

finnolec (approved):

psychocoderHPC (approved):

PR #4307, opened 2022-09-30
fix/workaround ROCm 5.2.0 compiler bug

### LINKED ISSUE #4305: HIP/ROCm 5.2 is producing wrong results
When using ROCm 5.2 than the simulation is producing the wrong results.
After analyzing the root of the issue I found out that `ShiftParticles` is broken. (we lose particles over time)
The bug is gone if I revert #4173 for the file `ParticlesBase.kernel`.

I created a reproducer branch to analyze the reason for the bug. https://github.com/psychocoderHPC/picongpu/tree/topic-rocm5.2issue
If the last commit is removed all works fine with ROCm 5.2. The last commit is only reading the number of particles in a superCell but this read is not influencing the workflow of the kernel.
Currently, I would say there is a 95% chance that it is this is a compiler bug.

Tested with dev branch: 72fc81da05890d3c25f3ae15b95c7bd82c0aa006

psychocoderHPC:
fix #4305

Fix broken `KernelShiftParticles` most likely produced by a compiler bug.

The workaround is not showing adverse effects on the registered footprint on NVIDIA GPUs (tested sm_70).

### DIFF
--- a/include/picongpu/particles/debyeLength/Estimate.kernel
+++ b/include/picongpu/particles/debyeLength/Estimate.kernel
@@ -83,7 +83,7 @@ namespace picongpu
                     T_EstimateBox estimateBox) const
                 {
                     auto const supercellIdx = mapper.getSuperCellIndex(DataSpace<simDim>(cupla::blockIdx(acc)));
-                    auto const& supercell = electronBox.getSuperCell(supercellIdx);
+                    auto& supercell = electronBox.getSuperCell(supercellIdx);
                     if(supercell.getNumParticles() >= minMacroparticlesPerSupercell)
                         process(acc, electronBox, supercellIdx, estimateBox);
                 }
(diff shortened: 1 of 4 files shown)


test/KHI_growthRate/include/picongpu/param/grid.param
```diff
@@ -45,7 +45,6 @@ namespace picongpu
 {
     namespace SI
     {
-
```
PrometheusPi: Not related to this PR - but fine with me

PrometheusPi (dismissed):

test/KHI_growthRate/include/picongpu/param/grid.param
psychocoderHPC: Ohh, this should not be changed.

psychocoderHPC:
@PrometheusPi I removed the unnecessary change, the PR can be merged.

PrometheusPi (approved):

PR #4308, opened 2022-10-04
fix wrong memory requirment calculation in .tpl files

pordyna:
There was an error in most of our `.tpl` files that made picongpu jobs run almost always exclusive, even when not requiring the full node. Jobs using less than a full node of GPUs were still asking for full node memory.


psychocoderHPC:
Additional information: The tpl files emulate something like slots for GPUs on the different system because many systems not handling GPUs and have no strict checks that users use only the allocated GPUs.
We need to take care that we not query more memory per GPU than a GPUs virtually had. The memory per GPU slot is `HostMemory/NumHostedGPUs`.
In case we do not use the node exclusive and a second job is using the GPUs we should take care that there is enough memory for the free GPU slots available.

psychocoderHPC (approved):

PR #4313, opened 2022-10-07
refactor CountParticles and particle filters

psychocoderHPC:
- Change the interface of filters from using a frame and the index within the frame to accessing particles by using a particle instance instead.
- Use on-device particle forEach.

Note: The way how `workerIdx` was calculated was wrong but luckily ended up in the same value as the correct solution. 
```
DataSpace<dim> const threadIndex(cupla::threadIdx(acc));
auto const workerIdx
    = static_cast<uint32_t>(DataSpaceOperations<dim>::template map<SuperCellSize>(threadIndex));
```
must be

```
uint32_t const workerIdx = cupla::threadIdx(acc).x;
```

because the kernel was always started with one-dimensional block size.

### DIFF
--- a/include/pmacc/particles/operations/CountParticles.hpp
+++ b/include/pmacc/particles/operations/CountParticles.hpp
@@ -26,6 +26,7 @@
 #include "pmacc/mappings/kernel/AreaMapping.hpp"
 #include "pmacc/memory/buffers/GridBuffer.hpp"
 #include "pmacc/memory/shared/Allocate.hpp"
+#include "pmacc/particles/algorithm/ForEach.hpp"
 #include "pmacc/particles/memory/dataTypes/FramePointer.hpp"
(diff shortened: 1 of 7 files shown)


include/pmacc/particles/operations/CountParticles.hpp
```diff
@@ -67,75 +68,46 @@ namespace pmacc
...
-                cupla::__syncthreads(acc);
-
-                onlyMaster(
-                    [&]()
+            forEachParticle(
+                acc,
+                [&counter, &filter, &accParFilter](auto const& accelerator, auto& particle)
+                {
+                    bool const useParticle = filter(particle);
+                    if(useParticle)
```
steindev, line 103: I am surprised, how the change in the construction of `forEachParticle` in line 81, changes its functor call signature.
We can discuss this offline tomorrow.

steindev (approved):
Discussed question offline.

PR #4314, opened 2022-10-07
fix growth rate test case folder location

psychocoderHPC:
With #4106 the KHI growth rate test was introduced and placed in the wrong folder and therefore never checked in the CI.
This PR is correcting the location of the test.

CC-ing: @MSVoss


steindev (commented):
@MSVoss: Weren't there code lines in the test suite that relied on knowing the position of the setup? Like when it is created before compiling and execution?

steindev:
@MSVoss I will merge anyways now. If there are problems, please provide a PR with a fix.

steindev (approved):

PR #4318, opened 2022-10-10
fix `pic-compile`

psychocoderHPC:
With #4235 a bug was introduces that `pic-compiele` is always using test case zero from `cmakeFlags`.

One side effect is that the CI since #4235 constantly tests case zero and not all other variants.
All users who compile parameter scans with `pic-compile` are affected too.

Thx @steindev for reporting the error.


steindev (approved):

steindev:
Thx for fixing instantly!

PrometheusPi (approved):

PR #4321, opened 2022-10-14
refactor lockstep programming

psychocoderHPC:
Entirely rewrite the lockstep programming of PMacc.
The initial implementation requires that the user starts the kernel with the correct number of threads and that the kernel uses the valid number of workers (compile time) and passes the thread index correctly to the lockstep classes.
This was hard to understand and the possibility of mistakes was high.

The new implementation simplifies the usage and avoids typical errors by passing wrong indices to the lockstep classes.

# Review

I split the PR into a few commits to simplifying the review. The most critical part is PMacc cuSTL and kernel/plugins using cuSTL. The reason is that cuSTL is not maintained since years and mostly not able to handle heterogeneous kernel very well.
Many cuSTL plugins can only run on GPU :-( anyway.
NOTE: we are working on removing cuSTL but this will take some time.

- [x] contains #4337 to fix compile issues
- [x] contains #4338 to fix broken game of life

The rewrite is not increasing the register memory footprint!

- [x] update documentation
- [x] perform more runtime tests/validation
- [x] rebase against https://github.com/ComputationalRadiationPhysics/picongpu/pull/4338

### DIFF
--- a/docs/source/prgpatterns/lockstep.rst
+++ b/docs/source/prgpatterns/lockstep.rst
@@ -62,11 +62,9 @@ A context variable must be defined outside of ``ForEach`` and should be accessed
 
 .. code-block:: cpp
 
-    // assume one dimensional indexing of threads within a block
-    uint32_t const workerIdx = cupla::threadIdx(acc).x;
+    // variable 'worker' is provides by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
     constexpr uint32_t frameSize = 256;
-    constexpr uint32_t numWorkers = 42;
-    auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize, numWorkers>(workerIdx);
+    auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize>(worker);
     auto vIdx = forEachParticleSlotInFrame(
         [](lockstep::Idx const idx) -> int32_t
         {
@@ -77,10 +75,8 @@ A context variable must be defined outside of ``ForEach`` and should be accessed
     // is equal to
 
     // assume one dimensional indexing of threads within a block
-    uint32_t const workerIdx = cupla::threadIdx(acc).x;
     constexpr uint32_t frameSize = 256;
-    constexpr uint32_t numWorkers = 42;
-    auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize, numWorkers>(workerIdx);
+    auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize>(worker);
     // variable will be uninitialized
     auto vIdx = lockstep::makeVar<int32_t>(forEachParticleSlotInFrame);
     forEachParticleSlotInFrame(
@@ -94,11 +90,9 @@ A context variable must be defined outside of ``ForEach`` and should be accessed
 
 .. code-block:: cpp
 
-    // assume one dimensional indexing of threads within a block
-    uint32_t const workerIdx = cupla::threadIdx(acc).x;
+    // variable 'worker' is provides by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
     constexpr uint32_t frameSize = 256;
-    constexpr uint32_t numWorkers = 42;
-    auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize, numWorkers>(workerIdx);
+    auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize>(worker);
     auto var = lockstep::makeVar<int32_t>(forEachParticleSlotInFrame, 23);
 
 
@@ -107,11 +101,9 @@ A context variable must be defined outside of ``ForEach`` and should be accessed
 
 .. code-block:: cpp
 
-    // assume one dimensional indexing of threads within a block
-    uint32_t const workerIdx = cupla::threadIdx(acc).x;
+    // variable 'worker' is provides by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
     constexpr uint32_t frameSize = 256;
-    constexpr uint32_t numWorkers = 42;
-    auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize, numWorkers>(workerIdx);
+    auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize>(worker);
     auto vIdx = forEachParticleSlotInFrame(
         [](lockstep::Idx const idx) -> int32_t
         {
@@ -146,14 +138,13 @@ Collective Loop
 
 .. code-block:: bash
 
+    // variable 'worker' is provides by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
     // `frame` is a list which must be traversed collectively
     while( frame.isValid() )
     {
         // assume one dimensional indexing of threads within a block
-        uint32_t const workerIdx = cupla::threadIdx(acc).x;
         constexpr uint32_t frameSize = 256;
-        constexpr uint32_t numWorkers = 42;
-        auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize, numWorkers>(workerIdx);
+        auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize>(worker);
         forEachParticleSlotInFrame(
            [&](lockstep::Idx const idx)
            {
@@ -175,11 +166,9 @@ Non-Collective Loop
 
 .. code-block:: cpp
 
-    // assume one dimensional indexing of threads within a block
-    uint32_t const workerIdx = cupla::threadIdx(acc).x;
+    // variable 'worker' is provides by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
     constexpr uint32_t frameSize = 256;
-    constexpr uint32_t numWorkers = 42;
-    auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize, numWorkers>(workerIdx);
+    auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize>(worker);
     auto vWorkerIdx = lockstep::makeVar<int32_t>(forEachParticleSlotInFrame, 0);
     forEachParticleSlotInFrame(
         [&](auto const idx)
@@ -205,9 +194,8 @@ Using a Master Worker
         bool
     );
 
-    // assume one dimensional indexing of threads within a block
-    uint32_t const workerIdx = cupla::threadIdx(acc).x;
-    auto onlyMaster = lockstep::makeMaster(workerIdx);
+    // variable 'worker' is provides by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
+    auto onlyMaster = lockstep::makeMaster(worker);
 
     // manipulate shared memory
     onlyMaster(
@@ -220,4 +208,4 @@ Using a Master Worker
     /* important: synchronize now, in case upcoming operations (with
      * other workers) access that manipulated shared memory section
      */
-    cupla::__syncthreads(acc);
+    worker.sync();
--- a/docs/source/prgpatterns/lockstep.rst
+++ b/docs/source/prgpatterns/lockstep.rst
@@ -62,7 +62,7 @@ A context variable must be defined outside of ``ForEach`` and should be accessed
 
 .. code-block:: cpp
 
-    // variable 'worker' is provides by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
+    // variable 'worker' is provided by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
     constexpr uint32_t frameSize = 256;
     auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize>(worker);
     auto vIdx = forEachParticleSlotInFrame(
@@ -90,7 +90,7 @@ A context variable must be defined outside of ``ForEach`` and should be accessed
 
 .. code-block:: cpp
 
-    // variable 'worker' is provides by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
+    // variable 'worker' is provided by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
     constexpr uint32_t frameSize = 256;
     auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize>(worker);
     auto var = lockstep::makeVar<int32_t>(forEachParticleSlotInFrame, 23);
@@ -101,7 +101,7 @@ A context variable must be defined outside of ``ForEach`` and should be accessed
 
 .. code-block:: cpp
 
-    // variable 'worker' is provides by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
+    // variable 'worker' is provided by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
     constexpr uint32_t frameSize = 256;
     auto forEachParticleSlotInFrame = lockstep::makeForEach<frameSize>(worker);
     auto vIdx = forEachParticleSlotInFrame(
@@ -138,7 +138,7 @@ Collective Loop
 
 .. code-block:: bash
 
-    // variable 'worker' is provides by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
+    // variable 'worker' is provided by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
     // `frame` is a list which must be traversed collectively
     while( frame.isValid() )
     {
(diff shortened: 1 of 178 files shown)


psychocoderHPC:
Here is a simple overview of the most used transition changes in this PR.

All PMacc kernel now uses the context element instance `worker` instead of `acc` to encapsulate the information of a thread/worker and provide access to the alpaka accelerator via `worker.getAcc()`
An alpaka accelerator is only required to interact with cupla, alpaka, or mallocMC interfaces.

# A typical kernel call 

- before looked like
```C++
constexpr uint32_t numWorkers
                        = pmacc::traits::GetNumWorkers<pmacc::math::CT::volume<SuperCellSize>::type::value>::value;
// If `numWorkers` in the kernel signature and the start parameter for the number of threads differ the kernel will end in undefined behavior.
PMACC_KERNEL(KernelName<numWorkers>{})
                    (numBlocks, numWorkers)(/*kernel argumenets*/);
```

- is transformed after into
```C++
auto workerCfg = lockstep::makeWorkerCfg(SuperCellSize{});
// The kernel is not required to hand over low level information about the number of threads used in the kernel.
PMACC_LOCKSTEP_KERNEL(KernelName{}, workerCfg)(numBlocks)( /*kernel argumenets*/);
```

# A typical kernel 

- before
```C++
template<uint32_t T_numWorkers>		
struct KernelName
{
  template<typename T_Acc>
  DINLINE void operator()(T_Acc const& acc /* more arguments*/)
  {
    constexpr uint32_t numWorker = T_numWorkers;
    uint32_t const workerIdx = cupla::threadIdx(acc).x;
    // indices and number of workers must always pass correctly to the lockstep functions else the behaviour is undefined.
    auto forEachParticle = lockstep::makeForEach<frameSize, numWorker>(workerIdx);
    // ...
  }
};
```

- after
```C++	
struct KernelName
{
  template<typename T_Worker>
  DINLINE void operator()(T_Worker const& worker /* more arguments*/)
  {
    // no index mismatch possible anymore
    auto forEachParticle = lockstep::makeForEach<frameSize>(worker);
    // ...
  }
};
```

include/pmacc/lockstep/WorkerCfg.hpp
```diff
@@ -0,0 +1,117 @@
...
+     * @attention: The real number of workers used for the lockstep kernel depends on the alpaka backend and will
+     * be adjusted by this class via the trait pmacc::traits::GetNumWorkers.
+     */
+    template<uint32_t T_numWorkers>
+    struct WorkerCfg
+    {
+        friend struct detail::WorkerCfgAssume1DThreads;
+
+        //! number of workers
+        static constexpr uint32_t numWorkers = pmacc::traits::GetNumWorkers<T_numWorkers>::value;
```
jkelling: I think the definition of `GetNumWorkers` is not (yet) changed by this PR, hence this should always result in `numWorkers = T_numWorkers`, [see here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/59e9b53605f9a5c1bf271eeb055bc74370a99052/include/pmacc/traits/GetNumWorkers.hpp#L41-L44)

include/pmacc/lockstep/WorkerCfg.hpp
```diff
@@ -0,0 +1,117 @@
...
+         *
+         * @return index of the worker
+         */
+        template<typename T_Acc>
+        HDINLINE static auto getWorker(T_Acc const& acc)
+        {
+            auto const localThreadIdx = alpaka::getIdx<alpaka::Block, alpaka::Threads>(acc);
+            auto const blockExtent = alpaka::getWorkDiv<alpaka::Block, alpaka::Threads>(acc);
+            auto const linearThreadIdx = alpaka::mapIdx<1u>(localThreadIdx, blockExtent)[0];
+            return Worker<T_Acc, T_numWorkers>(acc, linearThreadIdx);
```
jkelling: There is no check, including in `Worker`, whether `T_numWorkers` actually matches `blockExtent.prod()`. Routine runtime check here would be feasible, but there should at least be an assertion for debug builds somewhere, maybe an `ALPAKA_ASSERT_OFFLOAD(...)`.

include/pmacc/lockstep/WorkerCfg.hpp
psychocoderHPC: No, depending on the used cupla/accelerator the trait is giving other values back. For OMP2block the trait will return`1`.
The reason is that some alpaka accelerator has a restriction for the block size.
Before this PR the user always called this trait. Now the call is moved into `WorkerCfg` because calling this trait is an essential part of lockstep programming.

include/pmacc/lockstep/WorkerCfg.hpp
psychocoderHPC: Yes I also thought about it. I will use your suggestion. The check should only be enabled if we compile in debug mode else the performance will be bad and `assert` on GPU is requiring registers what is bad too. For a debug compile this is fine.

include/pmacc/lockstep/WorkerCfg.hpp
jkelling: Are you sure about this? The trait I linked to has a default value for the template parameter `T_ACC` and no second template parameter is passed here. To get another specialization, you would need to pass an accelerator type, e.g. `GetNumWorkers<T_numWorkers, AccOmp2Blocks<Idx, Dim>>`.

psychocoderHPC: Yes I am sure.

https://github.com/ComputationalRadiationPhysics/picongpu/blob/59e9b53605f9a5c1bf271eeb055bc74370a99052/include/pmacc/traits/GetNumWorkers.hpp#L41-L52

The default will be used and then the compiler selects the specialized implementation. If this did not work you would not be able to run `serial` or `omp2b` as PIConGPU backend because both backends have the limitation that only one thread per block is allowed.

jkelling: Ahh, so this is because `cupla::AccThreadSeq` is an alias for the default accelerator configured via cmake. I am not yet familiar enough with cupla: to me `AccThreadSeq` read like `::alpaka::AccCpuSerial`, I thought it was a fixed alias to a single alpaka backend.

include/pmacc/lockstep/Worker.hpp
```diff
@@ -24,53 +24,82 @@
...
-             */
-            HDINLINE static constexpr uint32_t getNumWorkers()
-            {
-                return T_numWorkers;
-            }
-        };
+        /** synchronize all workers
+         *
+         * @attention It is not allowed to call this method in side of a if branch or a loop if it is not guaranteed
+         * that all worker executing the same code branch.
```
jkelling:
```suggestion
         * @attention It is not allowed to call this method inside of an if branch or a loop if it is not guaranteed
         * that all workers are executing the same code branch.
```

include/pmacc/lockstep/Worker.hpp
```diff
@@ -24,53 +24,82 @@
...
 
-    } // namespace lockstep
-} // namespace pmacc
+        /** get the number of workers
+         *
+         * @return number of workers
+         */
+        HDINLINE static constexpr uint32_t getNumWorkers()
+        {
+            return T_numSuggestedWorkers;
```
jkelling: Why does this not return `numWorkers`? The caller may know what they suggested already, but is `numWorkers` not the number actually used?

share/pmacc/examples/gameOfLife2D/CMakeLists.txt
```diff
@@ -135,7 +135,7 @@ cupla_add_executable(gameOfLife
 
 target_link_libraries(gameOfLife PRIVATE pmacc::pmacc)
 if(PNGwriter_FOUND)
-    target_link_libraries(PNGwriter::PNGwriter)
+    target_link_libraries(gameOfLife PRIVATE PNGwriter::PNGwriter)
```
jkelling: This looks unrelated to this PR... are including this and some other changes to `gameOfLive` because you touched it anyway for the `acc` -> `worker` change? I can see the old code in places cause warnings, too...

docs/source/prgpatterns/lockstep.rst
```diff
@@ -62,11 +62,9 @@ A context variable must be defined outside of ``ForEach`` and should be accessed
 
 .. code-block:: cpp
 
-    // assume one dimensional indexing of threads within a block
-    uint32_t const workerIdx = cupla::threadIdx(acc).x;
+    // variable 'worker' is provides by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
```
jkelling:
```suggestion
    // variable 'worker' is provided by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
```

docs/source/prgpatterns/lockstep.rst
```diff
@@ -94,11 +90,9 @@ A context variable must be defined outside of ``ForEach`` and should be accessed
...
```
jkelling:
```suggestion
    // variable 'worker' is provided by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
```

docs/source/prgpatterns/lockstep.rst
```diff
@@ -146,14 +138,13 @@ Collective Loop
 
 .. code-block:: bash
 
+    // variable 'worker' is provides by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
```
jkelling:
```suggestion
    // variable 'worker' is provided by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
```

docs/source/prgpatterns/lockstep.rst
```diff
@@ -175,11 +166,9 @@ Non-Collective Loop
 
 .. code-block:: cpp
 
-    // assume one dimensional indexing of threads within a block
-    uint32_t const workerIdx = cupla::threadIdx(acc).x;
+    // variable 'worker' is provides by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
```
jkelling:
```suggestion
    // variable 'worker' is provided by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
```

docs/source/prgpatterns/lockstep.rst
```diff
@@ -205,9 +194,8 @@ Using a Master Worker
         bool
     );
 
-    // assume one dimensional indexing of threads within a block
-    uint32_t const workerIdx = cupla::threadIdx(acc).x;
-    auto onlyMaster = lockstep::makeMaster(workerIdx);
+    // variable 'worker' is provides by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
```
jkelling:
```suggestion
    // variable 'worker' is provided by pmacc if the kernel launch macro `PMACC_LOCKSTEP_KERNEL()` is used.
```

include/picongpu/fields/currentDeposition/Cache.hpp
```diff
@@ -46,47 +46,38 @@ namespace picongpu
...
-                    T_Acc const& acc,
-                    T_FieldBox fieldBox,
-                    T_FieldCache const& cachedBox,
-                    uint32_t const workerIdx)
+                template<typename T_BlockDescription, typename T_Worker, typename T_FieldBox, typename T_FieldCache>
+                DINLINE static void flush(T_Worker const& worker, T_FieldBox fieldBox, T_FieldCache const& cachedBox)
                 {
                     typename T_Strategy::GridReductionOp const op;
-                    ThreadCollective<T_BlockDescription, T_numWorkers> collectiveAdd(workerIdx);
+                    auto collectiveAdd = makeThreadCollective<T_BlockDescription>();
```
jkelling: Why is the functor instance called `collectiveAdd`? The operation `op` is supplied by the template parameter `T_Strategy` and the use type name does not suggest anything beyond, that the operation is a reduction. The documentation of this class also does not specify, that the reduction mus be an addition. Is this a misnomer?

include/picongpu/particles/flylite/NonLTE.tpp
```diff
@@ -29,7 +29,7 @@
 
 #include <pmacc/Environment.hpp>
 #include <pmacc/dataManagement/ISimulationData.hpp>
-#include <pmacc/traits/GetNumWorkers.hpp>
+#include <pmacc/lockstep/lockstep.hpp>
```
jkelling: Is this include even required here? Considering there are no other changes in this file, it could that `GetNumWorkers.hpp` was also superfluous.

include/pmacc/lockstep/WorkerCfg.hpp
psychocoderHPC: I added the check.

include/pmacc/lockstep/Worker.hpp
psychocoderHPC: This is a BUG and you are right! This must be `numWorkers`

share/pmacc/examples/gameOfLife2D/CMakeLists.txt
psychocoderHPC: Yes you are right. The problem is that this code is not CI validated and I would avoid the large refactoring is blocking all other PR to long.

psychocoderHPC: let me check if it is easy to move the patch into a separate PR.

include/picongpu/fields/currentDeposition/Cache.hpp
psychocoderHPC: I assume it was in the very past always an `Add` and after applying strategies to this part of code not changed.

share/pmacc/examples/gameOfLife2D/CMakeLists.txt
psychocoderHPC: opened PR #4338

include/picongpu/particles/flylite/NonLTE.tpp
psychocoderHPC: I removed it, it was wrong before and auto substitution changed it.

include/picongpu/fields/currentDeposition/Cache.hpp
jkelling: Can we remove the `Add` the? Or maybe replace it be `Reduction`. This is unrelated to this PR. But due to the small size of this change, it would never change otherwise, so this seems to be a good opportunity to fix it.

pordyna (dismissed):
I looked over the  `PIConGPU collision: use new lockstep programming interfaces` commit, and it looks fine.

include/picongpu/fields/currentDeposition/Cache.hpp
psychocoderHPC: I will do it with the rebase against https://github.com/ComputationalRadiationPhysics/picongpu/pull/4338

PrometheusPi:
@psychocoderHPC https://github.com/ComputationalRadiationPhysics/picongpu/pull/4338 was merged yesterday, you can rebase.

psychocoderHPC:
PR is rebased and should pass the CI now

include/picongpu/fields/currentDeposition/Cache.hpp
jkelling: It looks like the `Add` is still here...

include/picongpu/particles/collision/detail/CollisionContext.hpp
```diff
@@ -29,21 +29,23 @@ namespace picongpu
         {
             namespace detail
             {
-                template<typename T_Acc, typename T_RngHandle>
+                template<typename T_Worker, typename T_RngHandle>
                 struct CollisionContext
                 {
-                    T_Acc const* m_acc;
+                    T_Worker const* m_worker;
```
jkelling, line 35: I would consider making this member constant, for the benefit of the compiler, so this looks more like a stored reference. ... unless you intend to assign to this type.

jkelling: .. on second though: It just be a reference then. Both variants disable assignment and move ... I doubt than either is relevant with this context-type.

Anyways: this is unrelated to the PR, hence optional.

include/picongpu/particles/functor/misc/Rng.hpp
```diff
@@ -51,7 +51,7 @@ namespace picongpu
                     using Distribution = T_Distribution;
                     using RNGFactory = pmacc::random::RNGProvider<simDim, random::Generator>;
                     using RngHandle = typename RNGFactory::Handle;
-                    using RandomGen = RngWrapper<cupla::Acc, typename RngHandle::GetRandomType<Distribution>::type>;
+
 
                     /** constructor
```
jkelling: Not, sure, but I think there should only be one blank line here.

jkelling (changes requested):

include/picongpu/particles/collision/detail/CollisionContext.hpp
psychocoderHPC: This I would keep for now as it is because this is tested. Changing this could break thinks and postpone this PR to much

include/picongpu/particles/functor/misc/Rng.hpp
psychocoderHPC: removed the double blank line

include/picongpu/fields/currentDeposition/Cache.hpp
psychocoderHPC: changed to `collectiveFlush`

jkelling (approved):

Issue #4311, opened 2022-10-06
Documentation example output for MemoryCalculator is broken


BrianMarre:
In our current dev the documentation example of using the MemoryCalculator is broken.

see the bottom of [readthedocs](https://picongpu.readthedocs.io/en/latest/usage/workflows/memoryPerDevice.html#usage-workflows-memoryperdevice)

I am not quite sure why because the memory calculator itself is not using typesafe declarations.

steindev:
@BrianMarre Pls try to fix.

SimeonEhrig:
I will fix it

PR #4327, opened 2022-10-18
Install dependencies for memoryPerDevice example

SimeonEhrig:
The example is executed during build time of sphinx doc. The example has indirect dependencies to external libraries, which needs to be installed.

Addresses issue: #4311

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,16 @@
+version: 2
+build:
+    os: ubuntu-20.04
+    tools:
+        python: "3.9"
+    apt_packages:
+      - graphviz
+    
+python:
+   install:
+   - requirements: docs/requirements.txt
+
+formats:
+    - htmlzip
+    - pdf
+    - epub
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,6 +1,0 @@
-requirements_file: docs/requirements.txt
-
-formats:
-    - htmlzip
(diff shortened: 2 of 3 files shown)


SimeonEhrig:
@BrianMarre Ready for review. 

Unfortunately, I had to change the whole build environment of the sphinx doc. Before the `.readthedocs.yml` was outdated and ignored. Now, I set the container image for the build environment and the python version to a specific version, instead using the defaults. Therefore, the rest of the documentation needs also a quick check, that no new build problems occur.

steindev (dismissed):
Actually, I could not find other places in the documentation where we render python program output.
 @BrianMarre, @PrometheusPi do you know other places? 

The one mentioned in the issue works now. 

I approve.

BrianMarre:
I don't know of any other place either.

One quick note, the issues encountered by @franzpoeschel in #4329 could be related?

SimeonEhrig:
No, I rapidly skip trough the doc and I fond a place, where a plugin is not working anymore :-(
https://picongpu--4327.org.readthedocs.build/en/4327/pypicongpu/autoapi/picongpu/pypicongpu/species/attribute/index.html#picongpu.pypicongpu.species.attribute.Position

SimeonEhrig:
@steindev The issue is fixed. I have to install `graphviz` that the images are rendered.

steindev (approved):

PR #4328, opened 2022-10-18
fix PMacc reduce

psychocoderHPC:
`cupla::__syncthreads()` was used within a lockstep lambda. This is not allowed because `cupla::__syncthreads()` is a collective operation.

- moved `cupla::__syncthreads()` behind the lockstep lambda functor.

For the specific backends we usually use, this should not have an effect, the problem would be the alpaka OMP2 blocks backend.

[update]: luckily it looks like no backend is affected by this bug see https://github.com/ComputationalRadiationPhysics/picongpu/pull/4328#issuecomment-1284989744

### DIFF
--- a/include/pmacc/device/reduce/Kernel.hpp
+++ b/include/pmacc/device/reduce/Kernel.hpp
@@ -204,9 +204,9 @@ namespace pmacc
                                     && !(linearIdx != 0u && linearIdx >= active_threads);
                                 if(isActiveCtx[idx])
                                     func(acc, sharedMem[linearIdx], sharedMem[linearIdx + chunk_count]);
-
-                                cupla::__syncthreads(acc);
                             });
+
+                        cupla::__syncthreads(acc);
                     }
                 }
             };


jkelling:
> the problem would be the alpaka OMP2 blocks backend.

Do you mean OMP2 threads? I can see this  bug having a performance impact there, as in CUDA/HIP and the like, if `T_blockSize  > T_WorkerCfg::numWorkers`. It should have no impact in OMP2 blocks, as `__synchthreads()` is a no-op there. Just checking, that I did not completely misunderstand this code... .

jkelling (approved):

psychocoderHPC:
Yes I mean `OMP2 threads` but after thinking again about it, it could be that we are lucky because as you said for `OMP2 blocks` `synchronize()` is a noop and for all other, we performed this kerne with as many threads per block as we have worker/threads.

SO this fix is more generic fix to support e.g. a block/lockstep domain of 256 elements and 13 worker/threads. This is generally possible with lockstep but not the common case.

PR #4333, opened 2022-10-27
[WIP] Shadowgraphy plugin

finnolec:
Hello, here is the shadowgraphy plugin from my master's thesis.

## What does it do
The plugin integrates the Poynting vectors from the electric and magnetic field in a spatially static slice in the simulation to generate a synthetic 2D shadowgram.

When a probe laser (implemented with e.g. the incident field method) propagates through the plasma structures, it contains information about these plasma structures through refraction or diffraction. The probe laser can then be propagated with this in-situ PIConGPU plugin through a 4f correlator with a Fourier optics approach on a screen. This results in a 2D image, a so called "shadowgram". Additionally, other optical effects of the plasma are contained in this calculation method, that naturally occur through PIConGPU, like relativistic refraction index, plasma self emission or ionization effects.

Compared to the post-processing implementation that it is based on ([E Siminos et al 2016 Plasma Phys. Control. Fusion 58 065004](https://iopscience.iop.org/article/10.1088/0741-3335/58/6/065004/meta)), the in-situ shadowgraphy plugin allows to calculate synthetic shadowgrams without having to store and load the field data. The laser is also extracted slice-wise over a set period of time, instead of a 3D Fourier transform of simulation box volume.

## TODO
- [ ] Test simulations (I reworked a lot of code since the master's thesis, I should probably test if the results still look fine)
- [ ] Example setup
- [ ] FFTW/DFT improvements (There might be some symmetries that we can use to make them faster)
- [ ] Remove custl
- [ ] Make it multi-pluginable
- [ ] Clean up commit history
- [x] openPMD output


PrometheusPi:
@finnolec There seems to be a formatting issue. You can fix it with:
```
sed -i 's/\(include.*\)"\(pmacc\/.*\)"/\1<\2>/' include/picongpu/plugins/shadowgraphy/ShadowgraphyHelper.hpp
```

psychocoderHPC:
@finnolec could you update this PR with the latest changes I did + your openPMD meta data

PrometheusPi:
@finnolec What is the status of your PR?

finnolec:
> @finnolec What is the status of your PR?

I am sometimes working on it. Basically @psychocoderHPC removed the custl parts in the code and there is the start of an openPMD implementation running. However, this is not my top priority, I can't give an ETA when it's mergeable. There is still e.g. documentation work that needs to be done.

share/picongpu/examples/Shadowgraphy/include/picongpu/param/incidentField.param
```diff
@@ -0,0 +1,239 @@
...
+            using MyYMaxSource = Source<FunctorYMaxIncidentE, FunctorYMaxIncidentB>;
+
+            /**@{*/
+            /** Incident field source types along each boundary, these 6 types (or aliases) are required.
+             *
+             * Each type has to be either Source<> or None.
+             *
+             * Here we generate a plane wave from the X min border.
+             */
+            using XMin = MyXMinSource;
```
PrometheusPi, line 218: why do you define E and B separately (and not use a default laser setup)?

finnolec:
Not necessary anymore with #4868

PR #4334, opened 2022-10-28
Unified testsuite

MSVoss:
This is my suggestion for the generalization of the test suite. As already stated, the current version of the KHI test is very geared towards this. In order to make the test suite usable for different test cases, specific components had to be replaced. Among them are the following aspects in particular:

- Labeling and description within the results (plot and .log file), including e.g. title, author, ... (see TestKHI.py file, in which these are statically defined)
- Function for calculating the theoretical values ​​(see TheoryGrowthrate.py)
- Calculation of the values ​​from the simulation data
- Limited use of other functions for the general (e.g. function for calculating the deviation)

In order to remedy these shortcomings, the following has been changed (here only the essential aspects):
- New structure of the test suite for better encapsulation and clarity
- Introduction of Data.py: This file serves as a parameter file and contains essential parameters of the test, such as author, title, theory calculation function,.. . This eliminates the static character of the old structure.

There are a few things to keep in mind:
- This is only the first version of the generalization, extensions will follow later
- Deterioration compared to the previous version: no case distinction, it follows that the test case of the KHI (i.e. MI and ESKHI) currently has to be treated separately
- Some functions have not yet been completed, such as the 2D plot

Still under development or included in the next "version":
- Simple case distinction: Using a parameter, the test suite should be able to distinguish individual different theoretical cases from each other (e.g. see KHI: use the Lorentz factor to distinguish whether ESKHI or MI is present)
- Plotting and evaluating 2D plots
- Evaluation of openPMD data

Currently integrated tests:
- MI
- ESKHI
Both applicable for 3D as well as for 2D (however, the corresponding .param files have to be adapted to the case). Please note the dependency on the Lorenz factor in 3D. The included ci.bash script uses the MI. For more information, refer to the test case documentation.

Resulting advantage for the user compared to the old version.
- The result is easily customizable
- Test developer can freely define functions of the theory
- ...

Only the most important aspects are mentioned in the list above. Further information can be found in the individual documentation (function documentation)
Documentation will be added later for more information


MSVoss:
Flake8 will be corrected shortly

steindev:
@MSVoss There are about 300 lines of complaints by `flake8`, this is the reason the pull request validation failed. See report.

EDIT: I did not see your second comment, sorry for bothering you again :see_no_evil: 

EDIT2: Would you please update the PR description with an explanation of 'generalization'? That is, what is the aim of the generalization and how do you achieve this? (Just a few lines)

lib/python/test/setups/ESKHI/Data.py
```diff
@@ -0,0 +1,86 @@
...
+information, for example, see the bachelor thesis submitted
+by Mika Soren Voß).
+
+For information:
+    according to the documentation, the case distinction will only
+    be incorporated in a later version, which means that two case
+    s of KHI currently have to be treated differently.
+"""
+import sys
+sys.path.insert(1, "../../")
```
PrometheusPi: why do you replace the 2nd entry of the path list?

MSVoss: It's almost random. I could also have used append and simply added it afterwards. But here I'm looking for another solution anyway because flake8 is not fulfilled here.

docs/source/testing/examples/MI.rst
```diff
@@ -0,0 +1,46 @@
...
+
+.. sectionauthor:: Mika Soren Voß
+
+.. seealso::
+
+   For a better understanding, please read the :ref:`Kelvin Helmholtz instability chapter <.../examples/KelvinHelmholtz/README>` in the examples setups.
+
+
+
+The mushroom instability is the transverse dynamics in the KHI and dominates the plasma instability at high Lorentz factors of the plasma streams.
```
PrometheusPi:
```suggestion
The mushroom instability is the transverse dynamics of the KHI and dominates the plasma instability at high Lorentz factors of the plasma streams.
```

docs/source/testing/examples/MI.rst
```diff
@@ -0,0 +1,46 @@
...
+For the sake of simplicity, we consider two equally fast plasma streams, which flow past each other in opposite directions. 
+From this assumption, the growth rate :math:`$\Gamma_{max}$` can be described as:
+
+.. math::
+   $\Gamma_{max} = \omega_p \frac{v_0}{c\gamma^{0.5}}$
+   
+With :math:`\omega_{pe}` being the plasma frequency, c the Speed of Light.
```
PrometheusPi:
```suggestion
with :math:`\omega_{pe}` being the plasma frequency, c the Speed of Light. 
```

docs/source/testing/examples/MI.rst
```diff
@@ -0,0 +1,46 @@
...
+.. math::
+   $\Gamma_{max} = \omega_p \frac{v_0}{c\gamma^{0.5}}$
+   
+With :math:`\omega_{pe}` being the plasma frequency, c the Speed of Light. 
+For further information on the derivation, please refer to the references linked in the Kelvin Helmhotz instability chapter mentioned above.
+
+.. image:: ../../../images/ESKHI_MI.png
+   :alt: Comparison between the growth rates of the ESKHI and MI
+   
+This results in the point of intersection in the Lorentz factor with :math:`$\gamma = \sqrt{\frac{9}{8]}$`.
```
PrometheusPi:
```suggestion
The longitudinal KHI instability (esKHI) and the traversal instability (MI) have different growth rates depending the the Lorentz factor of the plasma streams. This results in a point of intersection at a Lorentz factor of :math:`$\gamma = \sqrt{\frac{9}{8]}$`, where both instabilities compete and have the same growth rate.
```

docs/source/testing/examples/MI.rst
```diff
@@ -0,0 +1,46 @@
...
+This has the advantage of being able to use all Lorentz factors.
+The second option is to use a 3D simulation of the KHI, whereby a sufficiently large Lorentz factor must be used.
+
+If one examines the deviation of the growth rate from the simulation with that from the theory as a function of the cell width, one obtains:
+
+.. image:: ../../../images/vortice_scan.png
+   :alt: Comparison between the growth rates of the ESKHI and MI
+
+One of the resulting minima in the deviation (3d) is obtained with :math:`$CELL_WIDTH_SI = 8.0378e-8m$`.
+From further considerations, it follows that the 3D simulation is best suited as a test case, since there are low deviations for all Lorentz factors.
```
PrometheusPi: Here it is unclear to me what you mean with "From further considerations".

PrometheusPi (commented):
The MI text is great. I just added some minor suggestions.

docs/source/testing/examples/MI.rst
MSVoss: ok, than i will change it in something like: Further investigations showed that for different Lorentz factors in the 2D simulation of the MI, other Cell_WIDTH_SI values provide a minimum in the deviation. Therefore, the 3D simulation is suitable as a test case due to low deviations for all Lorentz factors.

lib/python/test/setups/ESKHI/Data.py
pordyna: @MSVoss I think sth like this could possibly also work and would maybe not trigger flake8. But, no idea if this is any better. 
```python
from importlib.machinery import SourceFileLoader
testsuite = SourceFileLoader("testsuite", "../../testsuite/__init__.py").load_module()
````

File: share/picongpu/tests/KHI_growthRate/include/picongpu/param/particle.param [hunk 6]
```diff
@@ -27,7 +27,7 @@
 #include <pmacc/math/operation.hpp>
 
 #ifndef PARAM_GAMMA
-#    define PARAM_GAMMA 1.021
+#    define PARAM_GAMMA 1.2
```
psychocoderHPC: please discuss with @PrometheusPi, this can be set via cmakeFlags provided in the PR #4356

psychocoderHPC (changes requested):

share/picongpu/tests/KHI_growthRate/include/picongpu/param/particle.param
MSVoss: Without a working PICMI this will be difficult to change. In the example, I could define the line directly and not introduce an additional variable.

BrianMarre: I think what @psychocoderHPC want to say is that PARAM_GAMMA need not be a compile time setting but rather only runtime.
And yes we have a working PICMI for the newest interface once #4364 is merged and we can talk about a PCIMI integration, should be straightforward.

psychocoderHPC: @MSVoss 

You can change setups by adding cases into the file `cmakeFlags` see the original KHI for an example https://github.com/ComputationalRadiationPhysics/picongpu/blob/debc2f5a3646498a99379a91a9bc8822de3c420a/share/picongpu/examples/KelvinHelmholtz/cmakeFlags#L32-L34

By default case zero is build. If you like to compile a different case you can run `pic-build -t 1` to build the case one.

share/picongpu/tests/KHI_growthRate/bin/validate.sh
```diff
@@ -2,67 +2,50 @@
...
-            ;;
-        -h|--help)
-            echo -e "$(help)"
-            shift
-            exit 0
-            ;;
-        --) shift; break;;
-    esac
-    shift
+while [[ $# > 0 ]] ; do
```
psychocoderHPC: please check that you not remopve getopt

share/picongpu/tests/KHI_growthRate/bin/ci.sh
```diff
@@ -6,91 +6,58 @@
...
-            ;;
-        -h|--help)
-            echo -e "$(help)"
-            shift
-            exit 0
-            ;;
-        --) shift; break;;
-    esac
-    shift
+create="true"
```
psychocoderHPC: Please check that you not revert the getopt usage and reintroduce broken path to tools.

psychocoderHPC (changes requested):

share/picongpu/tests/KHI_growthRate/bin/ci.sh
MSVoss: I hope it's fixed

PrometheusPi:
@MSVoss is this pull request ready for a review?

MSVoss:
I tested it again and couldn't generate an error with the test cases. So yes

PrometheusPi:
The CI crashes in the `field absorber test`. Since this was not changed in this pull request, this is either some strange problem with the CI. (Thus I will restart the CI jobs.) or the base you are using @MSVoss is too old and still contains a bug in the `field absorber test`. Then a `git pull --rebase mainline dev` would help.

PrometheusPi:
Same CI jobs crashed again. @MSVoss could you rebase and resubmit?

share/picongpu/tests/KHI_growthRate/bin/ci.sh
```diff
@@ -66,7 +66,8 @@ while true ; do
             exit 0
             ;;
         --) shift; break;;
-    esac
+
+esac
```
psychocoderHPC: please revert this change.

psychocoderHPC (changes requested):
minior change required but only code style

psychocoderHPC:
@MSVoss How can I select if I like to verify an ESKHI or MI?

MSVoss:
> @MSVoss How can I select if I like to verify an ESKHI or MI?

Currently this is only possible in validate.sh. As before, care must be taken to adjust .param files correctly. I'm currently working on the possibility of reading out cmakeflag data. The two cases can thus be distinguished without an extra setting. Only a case distinction in the test suite is then still missing. But this is also emerging.

docs/source/testing/testbuilding.rst
```diff
@@ -0,0 +1,137 @@
...
+    
+In order to be able to evaluate the test, the test suite needs to know what data it can use and where it gets it from.
+To ensure this, the paths to the individual folders can be specified here.
+If data from these files is required, these parameters must be set unless they are passed to main.py. 
+(See the :ref:`main.py documentation <usage>` for this).
+
+``resultDirection`` designates the folder in which the result of the test is to be saved.
+If an automatic test is to be carried out and you are interested in the results in detail, this value should be set.
+
+``dataDirection`` gives the path to the .txt files of the simulation (note in-situ simulation).
```
PrometheusPi: What do you mean by `note in-situ simulation`?

docs/source/testing/testbuilding.rst
```diff
@@ -0,0 +1,137 @@
...
+    # parameter information found in .param files
+    param_Parameter = ["gamma", "BASE_DENSITY_SI", "DELTA_T_SI"]
+    data_Parameter = ["Bz", "step"]
+    
+3. test condition
+"""""""""""""""""
+
+In order for the test suite to be able to decide whether a test passed or not, it needs an acceptance range.
+This is done in the form of the parameter ``acceptance``, which contains the maximum deviation.
+An acceptance range of 0.2 thus describes a maximum deviation of 20% in which the test is passed.
```
PrometheusPi:
```suggestion
An acceptance range of 0.2 thus describes a maximum deviation of 20% from an expected value in which the test is passed.
```

docs/source/testing/testbuilding.rst
```diff
@@ -0,0 +1,137 @@
...
+    # for more values see the documentation (e.g. 2D plot needs zaxis and yaxis)
+    
+
+
+In this section, all data required for the plot display are transmitted.
+Possible parameters are ``plot_title``, ``plot_xlabel``, ``plot_ylabel``, ``plot_type`` and ``plot_xaxis``.
+The first three define the annotation of the plot.
+
+``plot_type`` can assume the values 1D or 2D and describes the dimensionality of the plot.
+If no value is specified, 1D is used by default.
```
PrometheusPi, line 107: Do you mean x-y plots with 1D?

docs/source/testing/testbuilding.rst
```diff
@@ -0,0 +1,137 @@
...
+``plot_xaxis`` beschreibt die WErte der x-Achse(default: time).
```
PrometheusPi:
```suggestion
``plot_xaxis`` describes the values of the x-axis (default: time).
```

docs/source/testing/usage.rst
```diff
@@ -0,0 +1,25 @@
+.. _testing-usage:
+
+How to use the test suite
+=========================
+
+As already mentioned, the test suite can be used in two different ways.
+On the one hand, the automatic validation based on a given test case.
```
PrometheusPi: verb is missing (*is* based?)

docs/source/testing/usage.rst
```diff
@@ -0,0 +1,25 @@
...
+If a test is simulated manually, only the associated bashscript has to be executed.
+
+.. note::
+
+    When running the bashscript, a temporary folder is created, labeled with the current date. 
+    This folder contains all essential data of the test including the test result.
+    By default, this folder is automatically deleted after the test has been run and only the result is reported to the system as 0 or 1.
+    If you are interested in the data, the corresponding line in the bashscript must be commented out.
+
+Furthermore, the test suite should also be able to be used to evaluate existing simulations.
```
PrometheusPi: OK, but in context of line 7:
```suggestion
On the other hand, the test suite should also be able to be used to evaluate existing simulations.
```

lib/python/test/setups/ESKHI/Data.py
```diff
@@ -0,0 +1,91 @@
...
+information, for example, see the bachelor thesis submitted
+by Mika Soren Voss).
+
+For information:
+    according to the documentation, the case distinction will only
+    be incorporated in a later version, which means that two case
+    s of KHI currently have to be treated differently.
+"""
+
+# flake8: noqa
```
PrometheusPi: This deactivated the linter for the entire file. If I remember correctly only the delayed include in line 26 caused the flake8 issue. Thus I would favor just excluding line 26 via `... # noqa`

details can be found in [this discussion](https://stackoverflow.com/questions/45346575/what-does-noqa-mean-in-python-comments).

lib/python/test/setups/ESKHI/Data.py
```diff
@@ -0,0 +1,91 @@
...
+import os
+import sys
+
+sys.path.insert(1, os.path.dirname(os.path.abspath(__file__)) + "/../../")
+from testsuite import Math as tm
```
PrometheusPi: see above comment
```suggestion
from testsuite import Math as tm #noqa
```

lib/python/test/setups/ESKHI/Data.py
```diff
@@ -0,0 +1,91 @@
...
+
+# general information about the test
+title = "KHI Growthrate (2D ESKHI)"
+author = "Mika Soren Voss"
+
+
+# directory specified
+resultDirection = None
+paramDirection = "/bigdata/hplsim/scratch/voss29/myFirst_PIConGPU_runs/runs/KHI/002_KHI/input/include/picongpu/param"
+dataDirection = "/bigdata/hplsim/scratch/voss29/myFirst_PIConGPU_runs/runs/KHI/002_KHI/simOutput"
```
PrometheusPi: This are very specific directories
perhaps something like:
 ```suggestion
 pathToDirs = "/pleae/add/your/directory/"
paramDirection = pathToDirs + "/runs/KHI/002_KHI/input/include/picongpu/param"
dataDirection = pathToDirs + "/runs/KHI/002_KHI/simOutput"
```
would be better

File: lib/python/test/setups/ESKHI/Data.py [hunk 19]
```diff
@@ -0,0 +1,91 @@
...
+
+# parameter information found in .param files
+param_Parameter = ["gamma", "BASE_DENSITY_SI", "DELTA_T_SI"]
+data_Parameter = ["Bx", "step"]
+
+# acceptance in percentage
+acceptance = 0.2
```
PrometheusPi, line 42: The comment does not represent the code.
Later `acceptance` is multiplied with 100 to get a percent value. Here it is still a ratio (to 1). 
Please adjust the comment or the number.

lib/python/test/setups/ESKHI/Data.py
```diff
@@ -0,0 +1,91 @@
...
+plot_type = "1D"
+# if None or not defined the time will be used
+plot_xaxis = None
+# for more values see the documentation (e.g. 2D plot needs zaxis and yaxis)
+
+
+def theory(gamma, *args):
+    """
+    this function indicates how the theoretical values
+    can be calculated from the data. It must be filled out
```
PrometheusPi: What do you mean with "filled out"? Implemented?

lib/python/test/setups/ESKHI/Data.py
```diff
@@ -0,0 +1,91 @@
...
+    out : theoretical values!
+    """
+
+    return (1 / ((8**0.5) * gamma))
+
+
+def simData(Bx, *args):
+    """
+    this function indicates how the values from the simulation
+    can be calculated from the data. It must be filled out
```
PrometheusPi: same here

lib/python/test/setups/ESKHI/main.py
```diff
@@ -0,0 +1,70 @@
...
+Authors: Mika Soren Voss
+License: GPLv3+
+"""
+
+# flake8: noqa
+import argparse
+import os
+import sys
+
+sys.path.insert(1, os.path.dirname(os.path.abspath(__file__)) + "/../../")
```
PrometheusPi, line 14: please add a comment what you want to include with this relative path (in case locations change in the future)

lib/python/test/setups/ESKHI/main.py
```diff
@@ -0,0 +1,70 @@
...
+parser.add_argument("-p", help="Path of the folder in which the parameter"
+                    " files .params of the simulation were saved.",
+                    dest="param",
+                    type=str,
+                    const=None,
+                    nargs="?",
+                    default=None)
+
+parser.add_argument("-o", help="Path of the folder in which the files for"
+                    " openPMD evaluation can be found.",
```
PrometheusPi, line 49: Is this just the general interface or does the esKHI test also have a openPMD mode?

lib/python/test/setups/MI/Data.py
```diff
@@ -0,0 +1,94 @@
...
+resultDirection = None
+paramDirection = None
+dataDirection = None
+
+# parameter information found in .param files
+param_Parameter = ["gamma", "BASE_DENSITY_SI", "DELTA_T_SI"]
+data_Parameter = ["Bz", "step"]
+
+# acceptance in percentage
+acceptance = 0.2
```
PrometheusPi, line 42: same as before

lib/python/test/setups/MI/Data.py
```diff
@@ -0,0 +1,94 @@
...
+plot_type = None
+# if None or not defined the time will be used
+plot_xaxis = None
+# for more values see the documentation (e.g. 2D plot needs zaxis and yaxis)
+
+
+def theory(gamma, *args):
+    """
+    this function indicates how the theoretical values
+    can be calculated from the data. It must be filled out
```
PrometheusPi: same as above

lib/python/test/setups/MI/Data.py
```diff
@@ -0,0 +1,94 @@
...
+    # gamma is calculated automatically, does not have to be passed
+    v = ts.Math.physics.calculateV_O()
+
+    return (v / (c * np.sqrt(gamma)))
+
+
+def simData(Bz, *args):
+    """
+    this function indicates how the values from the simulation
+    can be calculated from the data. It must be filled out
```
PrometheusPi: same as above

lib/python/test/setups/MI/main.py
```diff
@@ -0,0 +1,70 @@
...
+Authors: Mika Soren Voss
+License: GPLv3+
+"""
+
+# flake8: noqa
+import argparse
+import os
+import sys
+
+sys.path.insert(1, os.path.dirname(os.path.abspath(__file__)) + "/../../")
```
PrometheusPi, line 14: same as above

lib/python/test/setups/MI/main.py
```diff
@@ -0,0 +1,70 @@
+"""
```
PrometheusPi, line 1: is there a difference between this main and the esKHI main?

lib/python/test/testsuite/Math/_manager.py
```diff
@@ -0,0 +1,26 @@
...
+
+import testsuite.Math.deviation as dv
+import testsuite._checkData as cD
+
+
+def _calculate(theory, simulation):
+
+    acceptance = cD.checkVariables(variable="acceptance")
+
+    # only 1D
```
PrometheusPi, line 20: why this comment?

lib/python/test/testsuite/Math/_searchData.py
```diff
@@ -0,0 +1,81 @@
...
+        try:
+            result = Reader.paramReader.getValue(parameter)
+        except Exception:
+            result = None
+
+        if result is None:
+            try:
+                result = Reader.jsonReader.getValue(parameter)
+            except Exception:
+                result = None
```
PrometheusPi, line 72: Does this mean that if a a param file is found, the search for json is skipped?
I this the intended behavior?

lib/python/test/testsuite/Math/_searchData.py
```diff
@@ -0,0 +1,81 @@
...
+
+        if result is None:
+            try:
+                result = Reader.dataReader.getValue(parameter)
+            except Exception:
+                result = None
```
PrometheusPi, line 78: This is the same as line 60-63, except that it is checked if `result is None`. But since this has been checked before and the `result` is either written or not, this will not be executed except when running again into the `except`.
Did you intend to check for openPMD files here?

lib/python/test/testsuite/Math/deviation.py
```diff
@@ -0,0 +1,208 @@
...
+                 both must have the same length
+
+    Return:
+    -------
+    out : The maximum in the deviation
+    """
+
+    difference = np.abs(theory - simulation)
+
+    return min(difference)
```
PrometheusPi: you return `min` but the function name and documentation says `max`

lib/python/test/testsuite/Math/deviation.py
```diff
@@ -0,0 +1,208 @@
...
+    simulation : float or array
+                 Values from the simulation
+                 Note: if the list and the values
+                 of the theory are also a list,
+                 both must have the same length
+
+    Return:
+    -------
+    out : Array or float
+          the difference between the two values
```
PrometheusPi:
```suggestion
          the absolute difference between the two values
```

lib/python/test/testsuite/Math/deviation.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+
+    difference = np.abs(theory - simulation)
+
+    return min(difference)
+
+
+def getDifference(theory, simulation):
+    """
+    calculates the difference between the theory value and
```
PrometheusPi:
```suggestion
    calculates the absolute difference between the theory value and
```

perhaps also indicated this is the method name

lib/python/test/testsuite/Math/deviation.py
```diff
@@ -0,0 +1,208 @@
...
+    the simulation value
+
+    Calculates:
+    --------
+    theory -simulation
```
PrometheusPi:
```suggestion
    abs( theory - simulation )
```

lib/python/test/testsuite/Math/deviation.py
```diff
@@ -0,0 +1,208 @@
...
+import numpy as np
+
+
+def getMaxDifference(theory, simulation):
+    """
+    calculates the maximum deviation between theory and simulation
+
+    Calculates:
+    --------
+    theory -simulation
```
PrometheusPi:
```suggestion
    max( abs( theory - simulation ) ) 
```
or: What do you mean with this comment?

lib/python/test/testsuite/Math/deviation.py
```diff
@@ -0,0 +1,208 @@
...
+          the difference between the two values
+    """
+
+    return np.abs(theory - simulation)
+
+
+def getDifferenceInPercentage(theory, simulation):
+    """
+    calculates the difference between the theory value and
+    the simulation value in percentage
```
PrometheusPi:
```suggestion
    the simulation's maximum value in percentage
```

lib/python/test/testsuite/Math/deviation.py
```diff
@@ -0,0 +1,208 @@
...
+
+    Calculates:
+    --------
+    (theory - simulation) / simulation * 100
```
PrometheusPi:
```suggestion
    (theory - max(simulation) ) / theory * 100
```
max missing and normalization to theory not to simulation in code but not here in comment

lib/python/test/testsuite/Math/deviation.py
```diff
@@ -0,0 +1,208 @@
...
+    Return:
+    -------
+    out : float or array
+    """
+
+    acceptance = cD.checkVariables(variable="acceptance",
+                                   parameter=acceptance)
+
+    return theory - (theory * acceptance), theory + (
+                       theory * acceptance)
```
PrometheusPi:
```suggestion
    return theory * (1  - acceptance), theory * (1 + acceptance)
```
Yours is correct but this might be shorter

lib/python/test/testsuite/Math/math.py
```diff
@@ -0,0 +1,66 @@
...
+__all__ = ["growthRate"]
+
+import numpy as np
+
+
+def growthRate(f, time, interval=None):
+    """
+    This function calculate the growthrate of a given quantity f(t)
+
+    Gamma_f(t_k)= log(f(t_(k+1))/f(t_(k-1)))/t_(k+1)- t_(k-1)
```
PrometheusPi:
```suggestion
    Gamma_f(t_k)= log(f(t_(k+1))/f(t_(k-1)))/(t_(k+1)- t_(k-1))
```

lib/python/test/testsuite/Math/math.py
```diff
@@ -0,0 +1,66 @@
...
+
+    input:
+    -------
+    f:         array
+                quantity whose growth rate is to be determined
```
PrometheusPi:
```suggestion
               quantity whose growth rate is to be determined
```

lib/python/test/testsuite/Math/math.py
```diff
@@ -0,0 +1,66 @@
...
+
+    time:      array
+               the array of the timesteps with the same size
+               as the quantity f
+
+    interval:  [starttime, endtime], optional
+                determines the time interval in which the growth rate
+                is to be determined
```
PrometheusPi:
```suggestion
               determines the time interval in which the growth rate
               is to be determined
```

lib/python/test/testsuite/Math/math.py
```diff
@@ -0,0 +1,66 @@
...
+
+    return:
+    -------
+    growth rate: array if interval = None
+                  growth rate calculated using the above formula
```
PrometheusPi:
```suggestion
                 growth rate calculated using the above formula
```

lib/python/test/testsuite/Math/math.py
```diff
@@ -0,0 +1,66 @@
...
+
+                tuple of (growthrate, start, stop)
+                growth rate calculated using the above formula and
+                the start and end of the interval
+    """
+
+    if interval is None:
+        Gamma = 0.5 * np.log(f[3:]/f[1:-2])/(time[3:] - time[1:-2])
```
PrometheusPi, line 53: here, you exclude `f[0]` (which is good for KHI. But please add a comment why you do that especially in the context of having a time selector available.

lib/python/test/testsuite/Math/physics.py
```diff
@@ -0,0 +1,219 @@
...
+
+This module provides frequently used functions and laws of physics.
+Please note that this version is not complete and will be
+expanded over time.
+
+Routines in this module:
+
+calculateV_O(gamma = None):
+    calculate the speed v and use
+    gamma  = 1 / (1 - beta**2)**0.5
```
PrometheusPi: why this comment?

PrometheusPi: why not:
beta = v/c = sqrt(1 - 1 / gamma**2)

lib/python/test/testsuite/Math/physics.py
```diff
@@ -0,0 +1,219 @@
...
+def calculateV_O(gamma=None):
+    """
+    calculate the speed v and use
+
+    gamma  = 1 / (1 - beta**2)**0.5
+
+    Input:
+    -------
+    gamma :   float or numpy.ndarray, optional
+              lorentz factor
```
PrometheusPi:
```suggestion
              Lorentz factor
```

lib/python/test/testsuite/Math/physics.py
```diff
@@ -0,0 +1,219 @@
...
+
+    Return:
+    -------
+    beta:    float or numpy.ndarray
+    """
+
+    if v_0 is None:
+        v_0 = calculateV_O()
+
+    return v_0 / c
```
PrometheusPi: wouldn't it bet better to implement beta and than derive v_0 from beta? 
In v_0 you need to multiply with c. And here you have to divide by c again.

lib/python/test/testsuite/Math/physics.py
```diff
@@ -0,0 +1,219 @@
...
+    with parameter relativistic = True:
+        omega = sqrt((density * e ** 2) /(eps_0 * gamma * m_e)
+
+    Input:
+    -------
+    density :      float, optional
+                   If None, density is searched for in the parameter files.
+
+    gamma :        float, optional
+                   lorentz factor
```
PrometheusPi:
```suggestion
                   Lorentz factor
```

PrometheusPi (changes requested):
I am through half of the files so far. Here are my comments. 
I will try to review the rest as soon as possible.

lib/python/test/setups/ESKHI/main.py
MSVoss: The main.py file is the same for all test cases, so this is a general interface. A simplification is planned.

lib/python/test/setups/MI/main.py
MSVoss: As mentioned above, there is no difference.

lib/python/test/testsuite/Math/_manager.py
MSVoss: Is just a reminder for me that the following functions are only designed for the 1D case.

lib/python/test/testsuite/Math/_searchData.py
MSVoss: The functionality is a bit more complex, it will probably be better if we discuss this in person.

docs/source/testing/testbuilding.rst
PrometheusPi: @MSVoss ping

lib/python/test/setups/ESKHI/Data.py
PrometheusPi: fixed by @MSVoss

lib/python/test/setups/ESKHI/main.py
PrometheusPi: Okay

lib/python/test/setups/ESKHI/main.py
PrometheusPi: Okay

lib/python/test/setups/MI/Data.py
PrometheusPi: fixed

lib/python/test/setups/MI/main.py
PrometheusPi: Okay

lib/python/test/testsuite/Math/deviation.py
PrometheusPi: fixed

lib/python/test/testsuite/Math/math.py
PrometheusPi: still open (or I did not find the comment)

lib/python/test/testsuite/Output/Log.py
```diff
@@ -0,0 +1,133 @@
...
+import sys
+from time import localtime, strftime
+import testsuite._checkData as cD
+
+
+def resultLog(theory: float,
+              value_sim: float,
+              acceptance: float,
+              perc_diff: int,
+              acc_range,
```
PrometheusPi: only minor comment: no type given

lib/python/test/testsuite/Output/Log.py
```diff
@@ -0,0 +1,133 @@
...
+              result: bool,
+              difference,
```
PrometheusPi: only minor comment: no type given

lib/python/test/testsuite/Output/Log.py
```diff
@@ -0,0 +1,133 @@
...
+              direction: str = None,
+              title: str = None):
+    """
+    Creates the file testresult.log, in which all essential
+    parameters of the test and the result of the test are
+    summarized.
+
+    Input:
+    -------
+    direction : str, optional
```
PrometheusPi, line 56: There are more input parameters 
```python
              theory: float,
              value_sim: float,
              acceptance: float,
              perc_diff: int,
              acc_range
```
is there a reason they are not mentioned in the doc string?

lib/python/test/testsuite/Output/Log.py
```diff
@@ -0,0 +1,133 @@
...
+
+def errorLog(direction: str = None):
+    """
+    Catches errors while executing the test-suite and saves
+    them in the error.log file.
+
+    Input:
+    -------
+    direction : str, optional
+                The directory in which the log file should be saved.
```
PrometheusPi: optional: 
```suggestion
                The directory in which the error log file should be saved.
```

lib/python/test/testsuite/Output/Viewer.py
```diff
@@ -0,0 +1,129 @@
...
+    simulation : array, tuple or value
+                describes the value from the simulation.
+                If tuple: (x value, y value), plot only that value
+                          as a point.
+                If float: The value from the simulation is plotted as a line.
+                If array: must have the same length as x_value.
+
+    savefig :   bool, optional
+
+    direction : str, optional
```
PrometheusPi, line 50: not all inputs have been explained in the doc string

lib/python/test/testsuite/Output/Viewer.py
```diff
@@ -0,0 +1,129 @@
...
+            offset = theory * acceptance
+            plt.axhline(theory + offset, color="g")
+            plt.axhline(theory - offset, color="g")
+            plt.fill_between(x_value,
+                             theory + offset,
+                             theory - offset,
+                             color='green',
+                             alpha=.5)
+
+    elif len(theory) == 2:
```
PrometheusPi, line 93: could you add comments in the code what the different case (single scalar, two scalars and other stuff) do?

lib/python/test/testsuite/Output/Viewer.py
```diff
@@ -0,0 +1,129 @@
...
+        plt.plot(simulation[0], simulation[1], label="simulation")
+    else:
+        plt.plot(x_value, simulation, label="simulation")
+
+    if plotLog == "y":
+        plt.yscale("log")
+    elif plotLog == "x":
+        plt.xscale("log")
+    elif plotLog == "xy":
+        plt.loglog()
```
PrometheusPi, line 125: future update (not required for this pull request):
replace this with an args forwarding to matplotlib

lib/python/test/testsuite/Output/Viewer.py
```diff
@@ -0,0 +1,129 @@
...
+    plt.title(title)
+
+    plt.xlabel(xlabel)
+    plt.ylabel(ylabel)
+    plt.grid()
+
+    plt.legend()
+
+    if savefig:
+        plt.savefig(direction + "/testresult_1D.png")
```
PrometheusPi, line 153: fine for now:
Hard-coded file name

lib/python/test/testsuite/Output/Viewer.py
```diff
@@ -0,0 +1,129 @@
...
+    else:
+        plt.show()
+
+
+def plot_2D():
```
PrometheusPi, line 158: implementation missing

share/picongpu/tests/KHI_growthRate/include/picongpu/param/particle.param
PrometheusPi: what is the status of this change @MSVoss ?

lib/python/test/testsuite/Output/_manager.py
```diff
@@ -0,0 +1,119 @@
...
+from . import Log
+from . import Viewer
+import Data
+import sys
+import testsuite._checkData as cD
+import testsuite.Math.physics as ph
+from inspect import getmembers, isfunction
+
+
+def __calculate(axis, parameter, *args):
```
PrometheusPi, line 24: Could you add a short doc string describing what `__calculate` calculates?

lib/python/test/testsuite/Output/_manager.py
```diff
@@ -0,0 +1,119 @@
...
+    except Exception:
+        error1 = str(sys.exc_info()[1])
+        for var in parameter:
+            if var in error1.split(":")[-1]:
+                args = args + (parameter[var],)
+                return __calculate(axis, parameter, *args)
+
+
+def getaxisvalues(parameter, axis: str = "plot_xaxis"):
+
```
PrometheusPi, line 46: same here (but here the comments already help a lot)

lib/python/test/testsuite/Output/_manager.py
```diff
@@ -0,0 +1,119 @@
...
+        # check if the value is a parameter
+        if values in parameter.keys():
+            values = parameter[values]
+
+        # else a default value should be defined
+        if values == "":
+            frequency = ph.plasmafrequence()
+            time = ph.calculateTimeFreq(frequency,
+                                        step_direction="fields_energy.dat")
```
PrometheusPi, line 69: KHI specific default is fine

lib/python/test/testsuite/Output/_manager.py
```diff
@@ -0,0 +1,119 @@
...
+
+def _output(direction,
+            theory,
+            simulation,
+            max_diff,
+            perc,
+            acc_range,
+            result,
+            parameter):
+
```
PrometheusPi, line 87: please add a short doc string (not necessarily all input parameters if it is only used internally)

lib/python/test/testsuite/Output/_manager.py
```diff
@@ -0,0 +1,119 @@
...
+            x_value = x_value[:len(simulation)]
+
+        Viewer.plot_1D(theory,
+                       simulation,
+                       x_value,
+                       plotLog=plot_log,
+                       acceptance=acceptance,
+                       direction=direction)
+    else:
+        Viewer.plot_2D()
```
PrometheusPi, line 115: I think this explains my previous comment in `lib/python/test/testsuite/Output/Viewer.py` regarding this method -
this is just an interface definition and so far not implemented, correct @MSVoss ?

lib/python/test/testsuite/Reader/_manager.py
```diff
@@ -0,0 +1,70 @@
...
+json = {}
+param = {}
+data = {}
+
+
+def mainsearch(dataDirection: str = None,
+               paramDirection: str = None,
+               jsonDirection: str = None):
+
+    global json, param, data
```
PrometheusPi: Okay for now, in general `global` data is not a good idea 
Why did you have to settle with `global` data here, @MSVoss ?

lib/python/test/testsuite/Reader/_manager.py
```diff
@@ -0,0 +1,70 @@
...
+
+        paramDirection = cD.checkDirection(variable="paramDirection",
+                                           direction=paramDirection)
+
+        params = cD.checkVariables(variable="param_Parameter")
+        for parameter in params:
+            param[parameter] = paramReader.getValue(parameter,
+                                                    direction=paramDirection)
+
+        # if (cD.checkExistVariables(variable="dataDirection") or
```
PrometheusPi: Commented out code: please remove

lib/python/test/testsuite/Reader/_manager.py
```diff
@@ -0,0 +1,70 @@
...
+            # dataDirection is not None):
+            # dataDirection = cD.checkDirection(variable="dataDirection",
+            # direction=dataDirection)
+            # cmakeflagReader.checkflagInDir(direction=dataDirection)
+
+    # read .dat values
```
PrometheusPi: optional:
```suggestion
    # read data values
```
since not all input needs to be stored in a `*.dat` file (if given an explicit file name (as suggested in the next file))

lib/python/test/testsuite/Reader/dataReader.py
```diff
@@ -0,0 +1,281 @@
...
+    """
+    returns all parameters stored in the passed file
+
+    Note: this function cannot read histogram data,
+          but will not throw any errors if it reads it anyway
+
+    Input:
+    -------
+    file : str
+           file whose parameters are to be output
```
PrometheusPi: typo:
```suggestion
           file whose parameters are to be outputed
```

lib/python/test/testsuite/Reader/dataReader.py
```diff
@@ -0,0 +1,281 @@
...
+    if (len(all_files) >= 2 and
+        p_type is None and
+            "step" not in parameter):
+        raise ValueError("The parameter could be found more than once."
+                         " Please use the parameter p_type for this")
+
+    if "step" == parameter and step_direction is None:
+        result = np.loadtxt(direction + all_files[0])[:, 0]
+    elif "step" == parameter:
+        result = np.loadtxt(direction + step_direction)[:, 0]
```
PrometheusPi, line 259: I am not sure whether this works for different species. Have you tested it?
What is the use case for this  `step_direction` approach currently? 
And why did you choose the name `step_direction`?

lib/python/test/testsuite/Reader/dataReader.py
```diff
@@ -0,0 +1,281 @@
...
+    elif len(all_files) == 1:
+        params = allParamsinFile(direction + all_files[0])
+        index = params.index(parameter)
+        result = np.loadtxt(direction + all_files[0])[:, index]
+    elif len(all_files) == 0:
+        raise ValueError("The given Parameter could not be found")
+    else:
+        particles = [all_files[i].split("_")[0] for i in range(2)]
+
+        if p_type not in ["e", "i"]:
```
PrometheusPi: since species can have any name, wouldn't it be best to check whether the given species name exists as file name instead of checking against a given list?

PrometheusPi: Perhaps just try reading the file with the given species extension and catch the error when the file does not exist.

lib/python/test/testsuite/Reader/jsonReader.py
```diff
@@ -0,0 +1,200 @@
...
+           'getValue']
+
+import os
+import json
+import testsuite._checkData as cD
+
+
+def checkJSONFilesInDir(direction: str = None) -> bool:
+    """
+    checks if there are .json files in the directory
```
PrometheusPi, line 37: Is there currently a workflow that reads json files and thus tests this implementation?

lib/python/test/testsuite/Reader/jsonReader.py
```diff
@@ -0,0 +1,200 @@
...
+
+    # fixed value, just search for .json
+    fileExt = r".json"
+
+    all_files = [_ for _ in os.listdir(direction) if _.endswith(fileExt)]
+
+    if all_files:
+        return True
+    else:
+        return False
```
PrometheusPi, line 64: This looks like it could be generalized in a future version because `*.dat` and `*.json` files are treated exactly the same (perhaps via inheritance)

lib/python/test/testsuite/Reader/jsonReader.py
```diff
@@ -0,0 +1,200 @@
...
+          List of all names of .json files
+    """
+
+    direction = cD.checkDirection(variable="jsonDirection",
+                                  direction=direction)
+
+    # fixed value, just search for .json
+    fileExt = r".json"
+
+    return [_ for _ in os.listdir(direction) if _.endswith(fileExt)]
```
PrometheusPi, line 90: same here - perhaps use inheritance in a future version (not this pull request)

lib/python/test/testsuite/Reader/paramReader.py
```diff
@@ -0,0 +1,362 @@
...
+           "getValue"]
+
+import os
+import testsuite._checkData as cD
+from . import jsonReader as js
+import warnings
+
+
+# private functions -> no documentation
+def __checkifDefination(line: str, parameter) -> bool:
```
PrometheusPi: **very** minor: Your camel case notation changes, later you use:
```suggestion
def __checkIfDefination(line: str, parameter) -> bool:
```

PrometheusPi: Otherwise cool function due to its simplicity

lib/python/test/testsuite/Reader/jsonReader.py
PrometheusPi: Yes - used for json from cmake prefixes - see next file

lib/python/test/testsuite/Reader/paramReader.py
```diff
@@ -0,0 +1,362 @@
...
+
+        if len(all_paramFiles) > 1:
+            warnings.warn("Multiple files could be found"
+                          " with an \"undefined block\" for"
+                          " the same parameter.")
+
+        parameter = None
+
+        for filename in all_paramFiles:
+            all_Lines = paramInLine(search_u,
```
PrometheusPi, line 72: This overwrites `all_Lines` for all results in `all_paramFiles`and thus will only contain the last results. Is this intended? And if yes, why? (instead of e.g.)
```python
all_Lines = paramInLine(search_u, all_paramFiles[-1], direction).values()) 
```

lib/python/test/testsuite/Reader/paramReader.py
```diff
@@ -0,0 +1,362 @@
...
+                                    filename,
+                                    direction).values()
+
+            for line in all_Lines:
+                if "define" + search_u in line and parameter is None:
+                    parameter = float(line[line.find(search_u) +
```
PrometheusPi, line 78: This too will only use the last result

lib/python/test/testsuite/Reader/paramReader.py
```diff
@@ -0,0 +1,362 @@
...
+
+    if not getParam(parameter, direction):
+        raise ValueError("No .param file containing the"
+                         " parameter.")
+
+    all_paramFiles = getParam(parameter, direction)
+    for filename in all_paramFiles:
+        all_Lines = paramInLine(parameter, filename, direction).values()
+        for line in paramInLine(parameter, filename, direction).values():
+            if __checkifDefination(line, parameter):
```
PrometheusPi: see above comment on camel case

lib/python/test/testsuite/Template/Data.py
```diff
@@ -0,0 +1,85 @@
+"""
+This file is part of the PIConGPU.
+
+Copyright 2022 PIConGPU contributors
+Authors: Mika Soren Voss
+License: GPLv3+
+
+/ToDo: Description of the test case
```
PrometheusPi, line 8: What is with this to do?

lib/python/test/testsuite/Template/Data.py
```diff
@@ -0,0 +1,85 @@
...
+"""
+
+# /ToDo:
```
PrometheusPi: What is with this commented code ?

PrometheusPi: Perhaps just state: "uncomment if needed" ?

lib/python/test/testsuite/_checkData.py
```diff
@@ -0,0 +1,222 @@
+"""
+This file is part of the PIConGPU.
+
+Copyright 2022 PIConGPU contributors
+Authors: Mika Soren Voss
+License: GPLv3+
+
+This module checks whether Data.py exists or not.
+If not, it sets the default Data.py as a file.
```
PrometheusPi: What do you mean with "sets as a file?"

lib/python/test/testsuite/_checkData.py
```diff
@@ -0,0 +1,222 @@
...
+                data-reading module must be used for this.
+
+    """
+
+    val_name = "Data." + variable
+
+    # Since the variable does not necessarily have to be set,
+    # this must be checked beforehand
+    if variable in dir(Data):
+        val = eval(val_name)
```
PrometheusPi, line 93: `eval` is kind of dangerous (but ok if you can trust user input). 
I assume you want a numeric evaluation here, ow about using [numexpr](https://github.com/pydata/numexpr)?

lib/python/test/testsuite/_checkData.py
```diff
@@ -0,0 +1,222 @@
...
+
+    elif val is not None:
+        direction = val
+
+    elif (val is None and direction is None and
+          not errorhandling):
+        raise ValueError("Both " + variable + " and direction are none."
+                         "You must set at least one value.")
+
+    elif (val is None and direction is None and
```
PrometheusPi, line 114: I think all cases have been caught. Could you still add a `else` that throws an error so that readers of the code do not have to think about whether all cases have been handled?

lib/python/test/testsuite/_checkData.py
```diff
@@ -0,0 +1,222 @@
...
+    out : bool
+          False if the variable does not exist or is None,
+          else True
+    """
+    val_name = "Data." + variable
+
+    if variable not in dir(Data):
+        return False
+    else:
+        val = eval(val_name)
```
PrometheusPi, line 153: same as above

lib/python/test/testsuite/_checkData.py
```diff
@@ -0,0 +1,222 @@
...
+    out : type of default or parameter
+
+    """
+
+    val_name = "Data." + variable
+
+    # Since the variable does not necessarily have to be set,
+    # this must be checked beforehand
+    if variable in dir(Data):
+        val = eval(val_name)
```
PrometheusPi, line 197: same as above

lib/python/test/testsuite/_manager.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def __calculate(parameter, theory: bool = True, *args):
+    try:
+        if theory:
+            return Data.theory(*args)
+        else:
+            return Data.simData(*args)
+    except Exception:
+        error1 = str(sys.exc_info()[1])
```
PrometheusPi: what does this brach do?

PrometheusPi (changes requested):
finished review
some questions 
no mayor design problems (except for a more abstract code base via inheritance - but that's something for a later pr)

PrometheusPi (changes requested):
finished review
some questions 
no mayor design problems (except for a more abstract code base via inheritance - but that's something for a later pr)

lib/python/test/testsuite/Reader/jsonReader.py
psychocoderHPC: @MSVoss What is this code doing, we do not write json files.

lib/python/test/testsuite/Template/Data.py
```diff
@@ -0,0 +1,85 @@
...
+# import sys
+# sys.path.insert(1, "../../")
+# import testsuite as ts
+
+# general information about the test
+title = "Title"
+author = "Author"
+
+# directory specified
+resultDirection = None
```
psychocoderHPC, line 21: If this file is not explicitly used or called then the file will be soon outdated. I call this dead code.
I understand that this should be a template but in reality, everyone will use a working existing test case copy this to start.

lib/python/test/testsuite/Output/_manager.py
MSVoss: Not yet implemented in this pull request. The function already exists locally and is waiting for the next pull request

PrometheusPi: That's fine - thanks for the comment @MSVoss

share/picongpu/tests/KHI_growthRate/include/picongpu/param/particle.param
MSVoss: For this we need a reader to read the cmake flag files, this is intended for the next pull request

MSVoss: offline discussion with Richard, will be added in the next Pull Request

docs/source/testing/testbuilding.rst
MSVoss: Yes, thats correct

lib/python/test/testsuite/Math/_searchData.py
MSVoss: Exceptionhandling will add

lib/python/test/testsuite/Template/Data.py
MSVoss: This is a template with all the parameters in it. The templates from MI or ESKHI do not use all parameters to keep the test case as clear as possible. Pure copying of these test cases does not provide all the possibilities of the test suite.

MSVoss: So I think this code is very useful, because this file should serve as the basis for further test cases. It is thus comparable to docs/tbg_macros.cfg. But main.py in testsuite/Template can be removed, because it is every time the same.

lib/python/test/testsuite/_manager.py
MSVoss: As discussed in Person, the entire _calculate function will be removed.  It should pass only needed parameters to the function

lib/python/test/testsuite/_checkData.py
MSVoss: I will change it in one of the next pull request. I will also take a closer look at numexpr until then.

lib/python/test/testsuite/Reader/paramReader.py
MSVoss: I made a mistake when I explained this to you last week. In fact, this function does not necessarily overwrite and cannot be changed for now.

PrometheusPi: Okay - thanks for the update. What does it do then?

MSVoss:
I hope I haven't overlooked anything. I made a note of the inheritance for the next pull request.

lib/python/test/testsuite/Reader/dataReader.py
PrometheusPi: @MSVoss I am not sure whether I remember the conclusion from our initial discussion about it correctly? Could you please describe (as a reply to this text) what this is doing?

lib/python/test/testsuite/Template/Data.py
PrometheusPi: Discussed *offline* needs to be described by the implementer of the test case

lib/python/test/testsuite/_checkData.py
PrometheusPi: Done

PrometheusPi (approved):
Some open questions, but this looks good to me

PrometheusPi:
@psychocoderHPC @pordyna You had some open issues and questions on this pull request. What's the status?

PrometheusPi:
only open questions by @psychocoderHPC 
https://github.com/ComputationalRadiationPhysics/picongpu/pull/4334#issuecomment-1377156529

pordyna (dismissed):

PrometheusPi (changes requested):
I am blocking merging for now because when executing the CI via
```bash
./picongpu/share/picongpu/tests/KHI_growthRate/bin/ci.sh picongpu/share/picongpu/tests/KHI_growthRate/ ./KHI
```
(on hemera v100) I get the error `42`. 
Thus, when adding this test, the CI will always fail. 

@MSVoss could you clarify what I am doing wrong or what the error `42` is causing.

PrometheusPi:
The python code `validate.sh` throws the following error:
```
<class 'ValueError'> This should never happen <traceback object at 0x2aab462ff708>
```

PrometheusPi:
@MSVoss the PIP check of our CI is still unhappy with:
```
./lib/python/test/testsuite/_manager.py:13:1: E302 expected 2 blank lines, found 1
```

PrometheusPi (dismissed):

share/picongpu/tests/KHI_growthRate/include/picongpu/param/particle.param
psychocoderHPC: please add to cmakeFlags 

```
flags[3]="-DPARAM_OVERWRITES:LIST='-DPARAM_GAMMA=1.2'
# Depending on the GPU vendor and version compile can fail because of limited amound of shared memory.
# Please enable the test by hand!
# flags[4]="-DPARAM_OVERWRITES:LIST='-DPARAM_PRECISION=precision64Bit;-DPARAM_PARTICLESHAPE=PQS'"
```

psychocoderHPC: offline discussed with @PrometheusPi: we should not change the default behavior of this test. We have users which using this test case and we should not silently invalidate the possibility that old and now runs can be compared.

lib/python/test/testsuite/Reader/paramReader.py
psychocoderHPC: @MSVoss and @PrometheusPi please discuss this comment

psychocoderHPC (dismissed):

lib/python/test/testsuite/Reader/paramReader.py
PrometheusPi: @MSVoss for documentation purposes, could you please write, what the `paramInLine(...)` here does?  (As answer in this pull request discussion/thread)

lib/python/test/testsuite/Reader/jsonReader.py
MSVoss: I used JSON files for my bachelor thesis. There I needed them for the automatic test over many test cases. That's why I took it on here.

lib/python/test/testsuite/Reader/paramReader.py
MSVoss: This is a nested for loop. With the first for loop, the program goes through all the files found. After that it reads out all lines where the parameter is mentioned (that's what paramInLine does). Then it searches for the line where the parameter is defined. e.g. define PARAM_GAMMA= 1.021.

lib/python/test/testsuite/Reader/paramReader.py
MSVoss: see the comment on the following for-loop

lib/python/test/testsuite/Reader/dataReader.py
MSVoss: There are different steps that PIConGPU uses to save the individual simulation data. However, PIConGPU always saves them under the name step (hence the name step_direction). However, the test suite needs the right step size for the correct evaluation, so it must know from which file, i.e. from which step_direction, it gets it.

lib/python/test/testsuite/Reader/paramReader.py
PrometheusPi: *offline* discussion: there will only be zero or one entry

PrometheusPi (approved):

PR #4348, opened 2022-11-10
use cuplas/alpakas mapped memory

psychocoderHPC:
Instead of using directly CUDAs/ROCms mapped memory API calls switch using cupla and alpaka to remove backend-specific code from PIConGPU.

### DIFF
--- a/include/picongpu/plugins/output/WriteSpeciesCommon.hpp
+++ b/include/picongpu/plugins/output/WriteSpeciesCommon.hpp
@@ -21,19 +21,10 @@
 
 #include "picongpu/simulation_defines.hpp"
 
-#include "picongpu/plugins/ISimulationPlugin.hpp"
-
 #include <pmacc/dataManagement/DataConnector.hpp>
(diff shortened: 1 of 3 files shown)


include/picongpu/plugins/output/WriteSpeciesCommon.hpp
```diff
@@ -21,19 +21,10 @@
 
 #include "picongpu/simulation_defines.hpp"
 
-#include "picongpu/plugins/ISimulationPlugin.hpp"
```
PrometheusPi, line 24: Was removing this hpp file intentional?

PrometheusPi (approved):
so much cleaner code - great work
just one question

include/picongpu/plugins/output/WriteSpeciesCommon.hpp
psychocoderHPC: Yes all the removed includes was not used in this file, I assume it was in the past copy-pasted.

PR #4354, opened 2022-11-23
refactor PMacc plugin slicing

psychocoderHPC:
- rename `TimeSlice` into `Slice`
- add method `toRangeSlice` with special handling for the case that only the begin of the range is given

This PR fixes a bug @PrometheusPi found that setting only the begin of a range for the OpenPMD plugin dumped all data, instead of a single slice.

### DIFF
--- a/include/pmacc/pluginSystem/toSlice.hpp
+++ b/include/pmacc/pluginSystem/toSlice.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2018-2022 Rene Widera
+/* Copyright 2022 Rene Widera
  *
  * This file is part of PMacc.
  *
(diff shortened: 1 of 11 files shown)


psychocoderHPC:
@PrometheusPi can be merged

include/pmacc/pluginSystem/toSlice.hpp
```diff
@@ -53,14 +53,17 @@ namespace pmacc
 
         /** create a TimeSlice out of an string
          *
-         * Parse a comma separated list of time slices and creates a vector of TimeSlices.
-         * TimeSlice Syntax:
+         * Parse a comma separated list of time slices and creates a vector of slices.
+         * time slice syntax:
          *   - `start:stop:period`
          *   - a number ``N is equal to `::N`
```
PrometheusPi, line 59: Is this still true? I thought a single Number now maps to `N:N+1:1`.

PrometheusPi: yes - this is the (old) time slice - thus correctly maps to `0:-1:N`

PrometheusPi (approved):
I am a bit confused with `Slice`, but it looks good to me.

PR #4356, opened 2022-11-24
fix and refactor KHI growth test

psychocoderHPC:
- fix python script error if the error log file does not exist (create the file if it does not exist)
- change tool parameters of validate.sh
- change tool parameters of ci.sh
- give ci.sh and validate.sh executable rights

### DIFF
--- a/lib/python/test/KHI_growthRate/DifferenceKHIGrowthRate.py
+++ b/lib/python/test/KHI_growthRate/DifferenceKHIGrowthRate.py
@@ -90,4 +90,4 @@ class DifferenceKHIGrowthRate:
         """
 
         simMax = max(simMax)
-        return (theoryMax - simMax) / simMax * 100
+        return (theoryMax - simMax) / theoryMax * 100
--- a/share/picongpu/tests/KHI_growthRate/include/picongpu/param/particleFilters.param
+++ b/share/picongpu/tests/KHI_growthRate/include/picongpu/param/particleFilters.param
@@ -1,0 +1,119 @@
+/* Copyright 2013-2022 Rene Widera
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file
+ *
+ * A common task in both modeling and in situ processing (output) is the
+ * selection of particles of a particle species by attributes. Users can
+ * define such selections as particle filters in this file.
+ *
+ * Particle filters are simple mappings assigning each particle of a species
+ * either `true` or `false` (ignore / filter out).
+ *
+ * All active filters need to be listed in `AllParticleFilters`. They are then
+ * combined with `VectorAllSpecies` at compile-time, e.g. for plugins.
+ */
+
+#pragma once
+
+#include "picongpu/particles/filter/filter.def"
+#include "picongpu/particles/traits/SpeciesEligibleForSolver.hpp"
+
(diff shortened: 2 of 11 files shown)


PrometheusPi:
@psychocoderHPC please rebase and push again.

lib/python/test/KHI_growthRate/DifferenceKHIGrowthRate.py
```diff
@@ -90,4 +90,4 @@ def getDifferenceInPercentage(theoryMax: float, simMax):
         """
 
         simMax = max(simMax)
-        return (theoryMax - simMax) / simMax * 100
+        return (theoryMax - simMax) / theoryMax * 100
```
PrometheusPi, line 93: as discussed offline this should be correct - @MSVoss any objections?

share/picongpu/tests/KHI_growthRate/README.rst
```diff
@@ -0,0 +1,31 @@
+KelvinHelmholtz: Kelvin-Helmholtz Instability
```
PrometheusPi: Not really needed (file duplication)

share/picongpu/tests/KHI_growthRate/include/picongpu/param/particleFilters.param
```diff
@@ -0,0 +1,119 @@
+/* Copyright 2013-2022 Rene Widera
```
PrometheusPi, line 1: not needed - can be removed - why do we actually have this in the KHI example?

share/picongpu/tests/KHI_growthRate/include/picongpu/param/png.param
```diff
@@ -0,0 +1,100 @@
+/* Copyright 2013-2022 Axel Huebl, Heiko Burau, Rene Widera, Richard Pausch
```
PrometheusPi, line 1: not needed for the CI - there is no png output

PrometheusPi (changes requested):

share/picongpu/tests/KHI_growthRate/include/picongpu/param/radiationObserver.param
```diff
@@ -0,0 +1,86 @@
+/* Copyright 2013-2022 Heiko Burau, Rene Widera, Richard Pausch
```
PrometheusPi: not needed for CI

share/picongpu/tests/KHI_growthRate/include/picongpu/param/speciesDefinition.param
```diff
@@ -0,0 +1,106 @@
+/* Copyright 2013-2022 Rene Widera, Benjamin Worpitz, Heiko Burau
```
PrometheusPi: not needed for CI - we can use default

PrometheusPi (changes requested):

BrianMarre:
Short question why are the KHI growth test located in `lib/python/test/` and not in `test` where they belong.

Does not necessarily need to be fixed in this pull request, but should be corrected eventually.

share/picongpu/tests/KHI_growthRate/include/picongpu/param/particleFilters.param
psychocoderHPC: this file contains the selection of the particle moving up and down.

share/picongpu/tests/KHI_growthRate/include/picongpu/param/png.param
psychocoderHPC: I would keep it in if we need to debug the code

psychocoderHPC:
> Short question why are the KHI growth test located in `lib/python/test/` and not in `test` where they belong.
> 
> Does not necessarily need to be fixed in this pull request, but should be corrected eventually.

I do not know. In general this code could be in the test case only because it is test case specific. This is something we can rework in a separate PR.

psychocoderHPC:
@PrometheusPi I updated the PR

MSVoss:
If the code should be in a different folder, we can also integrate it directly into the pull request Unified Testsuite. Since a revision is taking place there anyway.

BrianMarre:
> If the code should be in a different folder, we can also integrate it directly into the pull request Unified Testsuite. Since a revision is taking place there anyway.

Ideally it should be next to pypicongpu and picmi tests in `test/...`

psychocoderHPC:
> If the code should be in a different folder, we can also integrate it directly into the pull request Unified Testsuite. Since a revision is taking place there anyway.

Let us keep the code where it is until my and our PR are in and then let us move the code.
If we move and change the code in one PR the diff will unreadable.

PrometheusPi (approved):

PR #4357, opened 2022-11-24
fix python code for new flake8 version=3.0.0

PrometheusPi:
Since our CI does not specify we are running into a CI flake8 error since today. 
The existing code does not follow the rule that variables `I` and `l` should be avoided.
See [E741](https://www.flake8rules.com/rules/E741.html) for details. 

This PR fixes that.


PrometheusPi:

psychocoderHPC (approved):

PR #4359, opened 2022-11-25
update of dev-server profile

BrianMarre:
some house keeping for compatibility with the gpu profile

### DIFF
--- a/etc/picongpu/bash-devServer-hzdr/cpu_picongpu.profile.example
+++ b/etc/picongpu/bash-devServer-hzdr/cpu_picongpu.profile.example
@@ -30,10 +30,10 @@ spack load cmake@3.23.1 %gcc@10.2.0
 spack load openpmd-api@0.14.4 %gcc@10.2.0 \
     ^adios2@2.8.0 \
     ^hdf5@1.12.1 \
-    ^openmpi@4.1.3 +atomics \
+    ^openmpi@4.1.3 +atomics ~cuda\
     ^python@3.9.12 \
     ^py-numpy@1.22.3
-spack load boost@1.78.0 +math +filesystem +system %gcc@10.2.0 ^python@3.9.12
+spack load boost@1.78.0 +math +filesystem +system %gcc@10.2.0
 
 # PIConGPU output dependencies ################################################
 #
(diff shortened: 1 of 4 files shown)


etc/picongpu/bash-devServer-hzdr/cpu_picongpu.profile.example
```diff
@@ -30,10 +30,10 @@ spack load cmake@3.23.1 %gcc@10.2.0
 spack load openpmd-api@0.14.4 %gcc@10.2.0 \
     ^adios2@2.8.0 \
     ^hdf5@1.12.1 \
-    ^openmpi@4.1.3 +atomics \
+    ^openmpi@4.1.3 +atomics ~cuda\
```
psychocoderHPC, line 33: I think this line is wrong. Why should we use an MPI without CUDA support? IS is very easy making problems.

psychocoderHPC (changes requested):

etc/picongpu/bash-devServer-hzdr/cpu_picongpu.profile.example
BrianMarre: because this is cpu-only

BrianMarre: there is a separate profile for gpu which explicitly uses cuda

BrianMarre: https://github.com/ComputationalRadiationPhysics/picongpu/blob/a17e2aa829e3f36c211959bf94f5d7121a3fba65/etc/picongpu/bash-devServer-hzdr/gpu_picongpu.profile.example#L33

BrianMarre: also this is intended as a minimal profile, therefore if we do not need it, it is not there

psychocoderHPC (dismissed):

BrianMarre:
@psychocoderHPC Sorry, found 1 last typo

psychocoderHPC (approved):

PR #4366, opened 2022-11-28
Plugin emittance: remove unused cuSTL includes

psychocoderHPC:
The plugin is including cuSTL but is not depending on cuSTL. Remove useless includes.


psychocoderHPC:
The plugin is using cuSTL for the MPI reduction part :-( https://github.com/ComputationalRadiationPhysics/picongpu/blob/6a17cb7997d2b14b2b6f6d1234fc1d407ee4444a/include/picongpu/plugins/Emittance.hpp#L734

I will close this PR and refactor the PR if we keep this PR.

PR #4368, opened 2022-11-30
remove current deposition VillaBune and EsirkepovNative

psychocoderHPC:
**Motivation:** Both algorithms are from the beginning of PIConGPU and were used in the past to have a reference to which we can compare our results of the current deposition. Both algorithms are not used in production runs and due to the fact that we have no test cases for the physics errors will therefore never be detected. EsirkepovNative doe not have a compile check.
Based on this fact it makes no sense to keep the algorithm and increase the maintenance overhead for each algorithmic change in the current deposition kernel.

- Remove the unused and untested current deposition algorithms.
- remove VillaBune current deposition from our documentation
- update param files
- remove VillaBune compile test


README.md
```diff
@@ -34,9 +34,8 @@ Its features for the electro-magnetic PIC algorithm include:
   [*Yee's*](http://dx.doi.org/10.1109/TAP.1966.1138693) and
   [*Lehe's*](http://dx.doi.org/10.1103/PhysRevSTAB.16.021301) scheme
 - rigorously charge conserving current deposition schemes, such as
-  [*Villasenor-Buneman*](http://dx.doi.org/10.1016/0010-4655%2892%2990169-Y),
   [*Esirkepov*](http://dx.doi.org/10.1016/S0010-4655%2800%2900228-9)
```
PrometheusPi, line 37: Is this `EsirkepovNative` and should thus be removed?

share/picongpu/examples/atomicPhysics/include/picongpu/param/species.param
```diff
@@ -46,14 +46,18 @@ namespace picongpu
...
- * - currentSolver::EmZ<SHAPE>        : particle shapes - CIC, TSC, PCS, P4S (1st to 4th order)
+/*! select current solver method
+ * - currentSolver::Esirkepov< SHAPE, STRATEGY > : particle shapes - CIC, TSC, PQS, PCS (1st to 4th order)
+ * - currentSolver::EmZ< SHAPE, STRATEGY >       : particle shapes - CIC, TSC, PQS, PCS (1st to 4th order)
+ * - currentSolver::EZ< SHAPE, STRATEGY >        : same as EmZ (just an alias to match the currently used naming)
  *
- * For development purposes: ---------------------------------------------------
- * - currentSolver::EsirkepovNative<SHAPE> : generic version of currentSolverEsirkepov
- *   without optimization (~4x slower and needs more shared memory)
+ * STRATEGY (optional):
```
PrometheusPi, line 54: This `STRATEGY` option was added to this PR. Might be better to introduce this in a different PR.

PrometheusPi: Sorry - This is already introduced in all other examples but was missing the the `atomicPhysics` example 
cc'ing @BrianMarre

README.md
PrometheusPi: No, we still have the (optimized) Esirkepov - thus we should definitely keep the reference. 
`EsirkepovNative` is just an Esirkepov without optimization for developing purposes

PrometheusPi (approved):

share/picongpu/examples/atomicPhysics/include/picongpu/param/species.param
PrometheusPi: is fine according the @psychocoderHPC (discussed offline) - will be added with this PR

PR #4375, opened 2022-12-01
remove PIConGPU install description using spack

psychocoderHPC:
Our spack recipe is not maintained and no PIConGPU developer/maintainer is using spack to use PIConGPU. To avoid users getting frustrated by the broken spack recipe the documentation is currently removed until we support and test the spack workflow.
The documentation of how the dependencies of PIConGPU can be installed with spack is still available.

Latest PIConGPU version with documentation referencing the spack
workflow: https://github.com/ComputationalRadiationPhysics/picongpu/commit/5d241da2ab80a53d3c6e91206e022a22a0e1b54d

**note** Before we reintroduce a spack recipe we should modernize our CMake script and add an install action. This was something we emulated in the spack recipe and is hard to maintain.


PrometheusPi (dismissed):
removing looks good to me 
Since @BrianMarre uses spack, he should comment wether his workflow relies on this.

BrianMarre:
It does not, I have recreated our module system profiles for spack, so different workflow to what Rene is referencing

PrometheusPi (approved):

cbontoiu:
As one of the few users of PIConGPU on Spack I must say thanks for all your support, sometimes related to Spack!

psychocoderHPC:
@cbontoiu You can still manage all your required dependencies with spack. We only removed the suggestion to use the broken picongpu spack recipe until we cleaned up PIConGPU to write a clean recipe and test this in the CI.

PR #4377, opened 2022-12-02
plugin radiation: use PMaccs forEachFrame

psychocoderHPC:
- Remove hand crafted iteration method over particles by using PMacc's forEachFrame.
- Use lockstep for each to iterate over frequencies

- [x] rebase against https://github.com/ComputationalRadiationPhysics/picongpu/pull/4378 to kick out the broken compiler g++-7

# Review

Use https://github.com/ComputationalRadiationPhysics/picongpu/pull/4377/files?w=1 for the review to ignore white space changes, this will make the diff much smaller.

### DIFF
--- a/include/picongpu/plugins/radiation/Radiation.kernel
+++ b/include/picongpu/plugins/radiation/Radiation.kernel
@@ -42,6 +42,7 @@
 #include <pmacc/memory/shared/Allocate.hpp>
 #include <pmacc/mpi/MPIReduce.hpp>
 #include <pmacc/mpi/reduceMethods/Reduce.hpp>
+#include <pmacc/particles/algorithm/ForEach.hpp>
 
 #include <cstdlib>
 #include <fstream>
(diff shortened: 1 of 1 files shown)


include/picongpu/plugins/radiation/Radiation.kernel
```diff
@@ -167,53 +168,39 @@ namespace picongpu
...
+                                auto forEachParticleInFrame = forEachFrameInSupercell.lockstepForEach();
+
+                                forEachParticleInFrame(
+                                    [&](lockstep::Idx const linearIdx)
                                     {
-                                        auto par = frame[linearIdx];
+                                        /* modulo is required if the kernel is started with more workers than particle
+                                         * in a supercell
+                                         */
+                                        auto par = frameCtx[linearIdx][linearIdx % frameSize];
```
PrometheusPi, line 203: @psychocoderHPC what does `[linearIdx % frameSize]` do/access?

include/picongpu/plugins/radiation/Radiation.kernel
```diff
@@ -347,85 +335,67 @@ namespace picongpu
...
-                            // wait till all radiation contributions for this super cell are done
-                            worker.sync();
-
-                            /* First threads starts loading next frame of the super-cell:
-                             *
-                             * Info:
-                             *   The calculation starts with the last SuperCell (must not be full filled)
-                             *   all previous SuperCells are full with particles
-                             */
-                            particlesInFrame = frameSize;
```
PrometheusPi, line 416: Is this now part of the `forEachParticleInFrame` loop?

PrometheusPi (approved):
Just two questions for you @psychocoderHPC

include/picongpu/plugins/radiation/Radiation.kernel
psychocoderHPC: Yes this is all hidden in the for each.

include/picongpu/plugins/radiation/Radiation.kernel
psychocoderHPC: The for each particle functor has the possibility to use a skip list if you execute the algorithm with more threads than particles in a frame. In this case we select a particle in the frame. 
offline discussed: the radiation shared memory structure is not ready to use this feature but to avoid later problems because we access a particle in the frame outside of the range the modulo is already included.

PrometheusPi:
Tests looks good:
grafik

This is the difference between a run of the bunch setup using the dev and one using the introduced changes. 
The relative difference is <4% in the peaks, which is in agreement with the random particle generation. 
(The higher differences are coming from the coherent differences and are to be expected between runs - see https://doi.org/10.1016/j.nima.2013.10.073 for details.

PrometheusPi:
The only thing that bothers me is why the PR version has these strange peaks on axis (bright dots at 64 direction index in above plot)

grafik

@psychocoderHPC is 64 a magic number in the new parallelization scheme?

PrometheusPi:
These artifacts only exist for direction 64 and vanish for direction 63 and 65. Thus I think this is a **bug**.

psychocoderHPC:
no there is no magic number in the iteration schema.
I will review my changes tomorrow again.

PrometheusPi:
The issue has been fixed with #4406

PR #4378, opened 2022-12-05
switch to CI container version 3

psychocoderHPC:
- test g++ 9-11
- test clang 10-15
- test boost 1.74-1.78


include/pmacc/boost_workaround.hpp
```diff
@@ -29,7 +29,11 @@
 /* workaround for compile error with clang-cuda
  * boost/type_traits/is_base_and_derived.hpp:142:25: error: invalid application of 'sizeof' to an incomplete type
  * 'boost::in_place_factory_base' BOOST_STATIC_ASSERT(sizeof(B) != 0);
+ *
+ * https://github.com/boostorg/config/issues/406#issuecomment-928151025
```
PrometheusPi, line 33: Is this boost workaround intentionally added to the new CI?

PrometheusPi (dismissed):
looks good to me - just a minor question @psychocoderHPC

include/pmacc/boost_workaround.hpp
psychocoderHPC: yes this is required to pass compile with new compilers. A compiler update was required because the new container used do not support old compilers anymore.

PrometheusPi (approved):

PR #4387, opened 2022-12-08
Radiation plugin: Enable streaming

franzpoeschel:
The radiation plugin currently hardcodes the HDF5 backend of openPMD, and opens a new instance of openPMD for each output time step.

This PR:
1. keeps one instance of `openPMD::Series` alive, just like the openPMD plugin does
2. adds runtime options to control the filename extension and JSON/TOML configuration for openPMD

TODO
- [x] I haven't actually tested it yet, do so
- [x] The plugin still requires HDF5 to compile at all and hardcodes HDF5 as default. Maybe use a more flexible default, as in other openPMD-based plugins.

### DIFF
--- a/include/picongpu/plugins/radiation/Radiation.hpp
+++ b/include/picongpu/plugins/radiation/Radiation.hpp
@@ -20,14 +20,15 @@
 
 #pragma once
 
-#if(ENABLE_OPENPMD * openPMD_HAVE_HDF5 != 1)
-#    error The activated radiation plugin (radiation.param) requires openPMD-api with a HDF5 backend
+#if !ENABLE_OPENPMD
+#    error The activated radiation plugin (radiation.param) requires openPMD-api.
 #endif
 
 #include "picongpu/simulation_defines.hpp"
 
 #include "picongpu/particles/traits/SpeciesEligibleForSolver.hpp"
 #include "picongpu/plugins/ISimulationPlugin.hpp"
+#include "picongpu/plugins/common/openPMDDefaultExtension.hpp"
 #include "picongpu/plugins/common/openPMDVersion.def"
 #include "picongpu/plugins/common/stringHelpers.hpp"
 #include "picongpu/plugins/radiation/ExecuteParticleFilter.hpp"
@@ -142,7 +143,8 @@ namespace picongpu
                 static const int numberMeshRecords = 3;
 
                 std::optional<::openPMD::Series> m_series;
-                std::string openPMDSuffix = "_%T_0_0_0.h5";
+                std::string openPMDSuffix
+                    = "_%T_0_0_0." + openPMD::getDefaultExtension(openPMD::ExtensionPreference::HDF5);
                 std::string openPMDConfig = "{}";
 
             public:
@@ -223,7 +225,9 @@ namespace picongpu
                         po::value<int>(&numJobs)->default_value(2),
                         "Number of independent jobs used for the radiation calculation.")(
                         (pluginPrefix + ".openPMDSuffix").c_str(),
-                        po::value<std::string>(&openPMDSuffix)->default_value("_%T_0_0_0.h5"),
+                        po::value<std::string>(&openPMDSuffix)
+                            ->default_value(
+                                "_%T_0_0_0." + openPMD::getDefaultExtension(openPMD::ExtensionPreference::HDF5)),
                         "Suffix for openPMD filename extension and iteration expansion pattern.")(
                         (pluginPrefix + ".openPMDConfig").c_str(),
                         po::value<std::string>(&openPMDConfig)->default_value("{}"),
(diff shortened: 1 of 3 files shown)


franzpoeschel:
New command line parameters:
```
  --e_radiation.openPMDSuffix arg (=_%T_0_0_0.h5)
                                        Suffix for openPMD filename extension 
                                        and iteration expansion pattern.
  --e_radiation.openPMDConfig arg (={}) JSON/TOML configuration for 
                                        initializing openPMD.
```

When the openPMD-api does not have HDF5:
```
  --e_radiation.openPMDSuffix arg (=_%T_0_0_0.bp4)
                                        Suffix for openPMD filename extension 
                                        and iteration expansion pattern.
  --e_radiation.openPMDConfig arg (={}) JSON/TOML configuration for 
                                        initializing openPMD.
```

franzpoeschel:
Question: By default, the radiation plugin writes only one single output. Is that always the case? 
This PR keeps the openPMD::Series open across time steps, so it is possible to have one single stream for several output steps. But maybe this is not needed?

psychocoderHPC:
> Question: By default, the radiation plugin writes only one single output. Is that always the case? This PR keeps the openPMD::Series open across time steps, so it is possible to have one single stream for several output steps. But maybe this is not needed?

Do you like to write a time series into a single file? This is not implemented because the python tools that exist do not support it.
The already existing tools are the reason why no adios is currently supported, the tools still read native hdf5 only and not using openPMD-api.

IMO adios2 support and time series support is welcome!

franzpoeschel:
> Do you like to write a time series into a single file? This is not implemented because the python tools that exist do not support it.

This is why this PR leaves the defaults as they are. For a stream, I need to write all data into a single "file". I just don't know if it is necessary, because in my test, the Radiation plugin only created one single output anyway.

psychocoderHPC:
we should throw away the prefix `_0_0_0` this is a left over from old times when native HDF5 with libsplash was used. It is still used to not touch the tools but with the changes, we need to touch the tools anyway.

Tools are here: https://github.com/ComputationalRadiationPhysics/picongpu/blob/master/lib/python/picongpu/plugins/data/radiation.py

Maybe the tools update and time series support should be added in a seperate PR.

franzpoeschel:
Ok, then I would also favor to have two PRs:

1. This one, which does not change the default output of the Radiation plugin, just enables other backends and iteration encodings
2. A second one which makes the Python scripts use openPMD insteady of h5py

I will push a further commit on this PR to automatically add a '.' before the specified suffix, then it should be ready for review.

psychocoderHPC (approved):

psychocoderHPC:
From my side it looks good, I will wait with merging until @PrometheusPi looks over it.

franzpoeschel:
One thing that I just noticed @psychocoderHPC: Upon `pluginUnload()`, should I manually close the `m_series`, or will the destructor be called then anyway?

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -268,7 +284,7 @@ namespace picongpu
                     {
                         writeOpenPMDfile(
                             tmp_result,
-                            restartDirectory + "/" + speciesName + std::string("_radRestart_"));
+                            restartDirectory + "/" + speciesName + std::string("_radRestart"));
```
PrometheusPi, line 287: @franzpoeschel Why is this renaming necessary?

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -521,7 +537,7 @@ namespace picongpu
                     {
                         writeOpenPMDfile(
                             timeSumArray,
-                            std::string("radiationOpenPMD/") + speciesName + std::string("_radAmplitudes_"));
+                            std::string("radiationOpenPMD/") + speciesName + std::string("_radAmplitudes"));
```
PrometheusPi, line 540: same here

PrometheusPi (approved):
Looks good to me if others agree that the changes in `openPMDDefaultExtension.hpp` should be included with this PR. 

Just some minor questions for you @franzpoeschel.

include/picongpu/plugins/radiation/Radiation.hpp
franzpoeschel: Because it now depends on the `--radiation.openPMDSuffix` parameter if this underscore is needed: By default, this plugin puts out something like `e_radAmplitudes_700_0_0_0.h5`. With this update, something like `e_radAmplitudes.h5` becomes possible, too – without an underscore. 
So I removed the hardcoded underscore in these two places, and put it instead into the default parameter `  --e_radiation.openPMDSuffix arg (=_%T_0_0_0.bp4)` (compare the parameter `--openPMD.infix _%06T`).

franzpoeschel: Without this change, we would get output like `e_radAmplitudes_.h5`.

PrometheusPi: Okay - thanks for answering - will merge

PrometheusPi:
`pluginUnload()` should only be called at the end of the entire simulation - thus not manually closing `m_series` should not be a problem. @psychocoderHPC do you agree?

PR #4392, opened 2022-12-12
fix broken slice parsing

psychocoderHPC:
with #4386 a new bug was introduced. The support for empty slices was removed, therefore PIConGPU is now calling all plugins each timestep.
Additionally, a doc string is added so that empty slices are supported.


PrometheusPi (dismissed):

PrometheusPi (approved):

PrometheusPi:
@psychocoderHPC Can I merge?

PR #4396, opened 2022-12-14
Integer exponent power function

BrianMarre:
reimplements the radiation plugin utils integer exponent power(base, exp) implementation directly in pmacc.
- New implementation uses c++ constexpr loop evaluation.
- Switched all pow() calls with integer exponents to the new implementation.
- Removes the old utils implementation
- includes some basic compile time tests for the new implementation

### DIFF
--- a/include/picongpu/particles/atomicPhysics/AtomicRate.hpp
+++ b/include/picongpu/particles/atomicPhysics/AtomicRate.hpp
@@ -17,11 +17,13 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
+#pragma once
+
 #include "picongpu/simulation_defines.hpp"
 
 #include <pmacc/algorithms/math.hpp>
 
-#pragma once
+#include <cstdint>
 
 /** rate calculation from given atomic data, extracted from flylite, based on FLYCHK
  *
@@ -119,7 +121,8 @@ namespace picongpu
                     for(uint8_t i = 0u; i < T_numLevels; i++)
                     {
                         result *= binomialCoefficients(
-                            static_cast<uint8_t>(2u * math::pow(float(i + 1), 2.0f)),
+                            static_cast<uint8_t>(2u)
+                                * pmacc::math::cPow(i + static_cast<uint8_t>(1u), static_cast<uint8_t>(2u)),
                             levelVector[i]); // unitless
                     }
 
@@ -238,7 +241,10 @@ namespace picongpu
                     // physical constants
                     // (unitless * m)^2 / unitless = m^2
                     float_X c0_SI = float_X(
-                        8._X * math::pow(picongpu::PI * picongpu::SI::BOHR_RADIUS, 2.0_X)
+                        8._X
+                        * pmacc::math::cPow(
+                            picongpu::PI * picongpu::SI::BOHR_RADIUS,
+                            static_cast<uint8_t>(2u))
                         / math::sqrt(3._X)); // uint: m^2, SI
 
                     // scaling constants * E_Ry^2/deltaE_Trans^2 * f * deltaE_Trans/E_kin
--- a/include/picongpu/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
+++ b/include/picongpu/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2016-2022 Marco Garten, Axel Huebl
+/* Copyright 2016-2022 Marco Garten, Axel Huebl, Brian Marre
  *
  * This file is part of PIConGPU.
  *
@@ -34,6 +34,8 @@
 #include "picongpu/traits/attribute/GetChargeState.hpp"
 
 #include <pmacc/algorithms/math/floatMath/floatingPoint.tpp>
+
+#include <cstdint>
 
 namespace picongpu
 {
@@ -100,7 +102,7 @@ namespace picongpu
 
                     float_64 const B = -math::exp(
                         thomasFermi::TFB0 + thomasFermi::TFB1 * T_F
-                        + thomasFermi::TFB2 * math::pow(T_F, float_64(7.)));
+                        + thomasFermi::TFB2 * pmacc::math::cPow(T_F, static_cast<uint8_t>(7u)));
 
                     float_64 const C = thomasFermi::TFC1 * T_F + thomasFermi::TFC2;
 
--- a/include/picongpu/plugins/radiation/radFormFactor.hpp
+++ b/include/picongpu/plugins/radiation/radFormFactor.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2013-2022 Heiko Burau, Rene Widera, Richard Pausch, Sergei Bastrakov
+/* Copyright 2013-2022 Heiko Burau, Rene Widera, Richard Pausch, Sergei Bastrakov, Brian Marre
  *
  * This file is part of PIConGPU.
  *
@@ -22,6 +22,7 @@
 
 #include "picongpu/plugins/radiation/utilities.hpp"
 
+#include <cstdint>
 
 namespace picongpu
 {
@@ -112,7 +113,7 @@ namespace picongpu
                         for(uint32_t d = 0; d < DIM3; ++d)
                             sincValue *= pmacc::math::sinc(
                                 observerUnitVec[d] * cellSize[d] / (SPEED_OF_LIGHT * 2.0_X) * omega);
-                        return util::pow(sincValue, 2 * T_shapeOrder);
+                        return pmacc::math::cPow(sincValue, static_cast<uint32_t>(2u) * T_shapeOrder);
                     }
                 };
             } // namespace radFormFactor_baseShape_3D
--- a/include/pmacc/algorithms/math/PowerFunction.hpp
+++ b/include/pmacc/algorithms/math/PowerFunction.hpp
@@ -1,0 +1,73 @@
+/* Copyright 2022 Brian Marre, Rene Widera, Richard Pausch
+ *
+ * This file is part of PMacc.
+ *
+ * PMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with PMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+
+#include "pmacc/types.hpp"
+#ifndef NDEBUG
+#    include "pmacc/static_assert.hpp"
+#endif
+
+#include <cstdint>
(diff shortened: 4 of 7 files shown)


BrianMarre:
@PrometheusPi, @steindev take a look and check that I did not break anything.

include/picongpu/plugins/radiation/radFormFactor.hpp
```diff
@@ -112,7 +113,7 @@ namespace picongpu
                         for(uint32_t d = 0; d < DIM3; ++d)
                             sincValue *= pmacc::math::sinc(
                                 observerUnitVec[d] * cellSize[d] / (SPEED_OF_LIGHT * 2.0_X) * omega);
-                        return util::pow(sincValue, 2 * T_shapeOrder);
+                        return pmacc::math::pow<float_X>(sincValue, 2u * T_shapeOrder);
```
PrometheusPi: lgtm

include/pmacc/algorithms/math/PowerFunction.hpp
```diff
@@ -0,0 +1,64 @@
...
+    {
+        T_Type result = static_cast<T_Type>(1u);
+        for(T_Exp e = static_cast<T_Exp>(0u); e < exp; e++)
+            result = result * x; // for whatever reason "*=" does not work, do not ask me ...
+        return result;
+    }
+
+    namespace test
+    {
+#if unitTest == true
```
psychocoderHPC: To avoid that you change with a macro code please prefix the e.g.
```
#if PMACC_UNIT_TEST == 1
```

In general you can also enable the test always if someone is compiling in debug mode instead of using a macro definition which is not set in CMake somewhere

```
#ifndef NDEBUG
```

This will disable the tests for release mode

include/pmacc/algorithms/math/PowerFunction.hpp
```diff
@@ -0,0 +1,64 @@
...
+        PMACC_CASSERT_MSG(FAIL_unitTest_2_power_0, pow<uint32_t>(2u, 0u) == static_cast<uint32_t>(1u));
+        PMACC_CASSERT_MSG(FAIL_unitTest_2_power_1, pow<uint8_t, uint8_t>(2u, 1u) == static_cast<uint8_t>(2u));
+        PMACC_CASSERT_MSG(FAIL_unitTest_4_power_4, pow<uint32_t, uint8_t>(4u, 4u) == static_cast<uint32_t>(256u));
+        PMACC_CASSERT_MSG(FAIL_unitTest_2_power_2, pow<double, uint8_t>(2., 2u) == 4.);
```
PrometheusPi: with this case, the floating point comparison should be fine, but in general such comparisons should be handled differently.

include/pmacc/algorithms/math/PowerFunction.hpp
```diff
@@ -0,0 +1,64 @@
...
+     * @tparam T_Exp exponent data type, default uint32_t
+     *
+     * @param x base
+     * @param exp exponent
+     */
+    template<typename T_Type, typename T_Exp = uint32_t>
+    HDINLINE constexpr T_Type pow(T_Type const x, T_Exp const exp)
+    {
+        T_Type result = static_cast<T_Type>(1u);
+        for(T_Exp e = static_cast<T_Exp>(0u); e < exp; e++)
```
PrometheusPi, line 51: @psychocoderHPC the actual reason for using template meta programming was to unroll this for loop.
Can we make sure that this for loop is unrolled on any architecture with today's compilers.

include/pmacc/algorithms/math/PowerFunction.hpp
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * @tparam T_Type return and accumulation data type
+     * @tparam T_Exp exponent data type, default uint32_t
+     *
+     * @param x base
+     * @param exp exponent
+     */
+    template<typename T_Type, typename T_Exp = uint32_t>
+    HDINLINE constexpr T_Type pow(T_Type const x, T_Exp const exp)
+    {
```
psychocoderHPC, line 45: please add a test to avoid that someone is calling this method with floats

```
static_assert(std::is_integral_v<T_Exp>, "This pow() implementation supports only integral types for the exponent.");
```

PrometheusPi (dismissed):
looks good to me - but I would wait till @psychocoderHPC answered on the for loop

include/pmacc/algorithms/math/PowerFunction.hpp
```diff
@@ -0,0 +1,64 @@
...
+        T_Type result = static_cast<T_Type>(1u);
+        for(T_Exp e = static_cast<T_Exp>(0u); e < exp; e++)
+            result = result * x; // for whatever reason "*=" does not work, do not ask me ...
```
psychocoderHPC: What is the error message?

include/pmacc/algorithms/math/PowerFunction.hpp
```diff
@@ -0,0 +1,64 @@
...
+#endif
+
+#include <cstdint>
+
+namespace pmacc::math
+{
+    /** power function for integer exponents, constexpr
+     *
+     * @tparam T_Type return and accumulation data type
+     * @tparam T_Exp exponent data type, default uint32_t
```
psychocoderHPC: plese add `must be an integral type`

include/pmacc/algorithms/math/PowerFunction.hpp
psychocoderHPC: This requires `#include <type_traits>`

include/pmacc/algorithms/math/PowerFunction.hpp
psychocoderHPC: we can add `PMACC_UNROLL(rounds)`, this is no guarantee but helps. Nevertheless in many cases, it will be unrolled.

include/pmacc/algorithms/math/PowerFunction.hpp
BrianMarre: You are right, in general floating point equality comparisons are troublesome. 

In this case I chose pure powers of twos which should be precisely representable in float.
I therefore do not expect a close but not the same error.

psychocoderHPC (changes requested):

include/pmacc/algorithms/math/PowerFunction.hpp
BrianMarre: Cool, the `NDEBUG` version sounds like a much better version of what I was trying to do.

And yes are correct, I should have used prefixes, still learning

include/pmacc/algorithms/math/PowerFunction.hpp
BrianMarre: done

include/pmacc/algorithms/math/PowerFunction.hpp
BrianMarre: none, but the function always returns `2` in the tests for some reason, caught it because the tests kept failing and i tried a bunch of things.
Would guess that more generally it would always return x in general but did not test.

include/pmacc/algorithms/math/PowerFunction.hpp
BrianMarre: done

include/pmacc/algorithms/math/PowerFunction.hpp
BrianMarre: switched to NDEBUG, instead of using a custom define done

include/pmacc/algorithms/math/PowerFunction.hpp
BrianMarre: added a PMACC_UNROLL(100), should be plenty for our us

BrianMarre: offline discussion: not possible since we do not always know at compile time how many loop iterations we have and unroll only gives back correct results in some implementations if `rounds<=iterations`

BrianMarre: Nevertheless a modern C++14-Compiler should unroll and evaluate at compile time, we can check using register usage and field solver test for the AOFD-solver, since it uses it in a fieldSolver-hot loop.
@psychocoderHPC agreed to take a look.
Add. Offline discussion: Probably it is not that important that the power function is evaluated at compile time, but the AOFD-weight array

BrianMarre:
we can not use PMACC_UNROLL since we do not know at compile time for all calls how many iterations there are going to be

include/pmacc/algorithms/math/PowerFunction.hpp
BrianMarre: done, added the bad baehviour to the comment

include/picongpu/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
```diff
@@ -100,7 +102,7 @@ namespace picongpu
 
                     float_64 const B = -math::exp(
                         thomasFermi::TFB0 + thomasFermi::TFB1 * T_F
-                        + thomasFermi::TFB2 * math::pow(T_F, float_64(7.)));
+                        + thomasFermi::TFB2 * math::pow<float_64, uint8_t>(T_F, 7u));
```
psychocoderHPC: There is no need to set `<float_64, uint8_t>`

psychocoderHPC: please use `pmacc::math::pow` else pow of alpak is called because `math::` points in picongpu to `alpaka::math`

include/picongpu/particles/atomicPhysics/AtomicRate.hpp
```diff
@@ -238,7 +241,10 @@ namespace picongpu
                     // physical constants
                     // (unitless * m)^2 / unitless = m^2
                     float_X c0_SI = float_X(
-                        8._X * math::pow(picongpu::PI * picongpu::SI::BOHR_RADIUS, 2.0_X)
+                        8._X
+                        * math::pow<float_64, uint8_t>(
```
psychocoderHPC: no need to use `<type,type2>`

include/pmacc/algorithms/math/PowerFunction.hpp
```diff
@@ -0,0 +1,67 @@
...
+    /** power function for integer exponents, constexpr
+     *
+     * @tparam T_Type return and accumulation data type
+     * @tparam T_Exp exponent data type, must be an integral type, default uint32_t
+     *
+     * @param x base
+     * @param exp exponent
+     */
+    template<typename T_Type, typename T_Exp = uint32_t>
+    HDINLINE constexpr T_Type pow(T_Type const x, T_Exp const exp)
```
psychocoderHPC: could you please rename this method in `cPow` or something not `pow` I see in the diff already that it is very easy to point by accident to the pow function provided by alpaka.

psychocoderHPC (changes requested):

include/picongpu/particles/atomicPhysics/AtomicRate.hpp
BrianMarre: done

include/picongpu/particles/ionization/byCollision/ThomasFermi/AlgorithmThomasFermi.hpp
BrianMarre: done

include/pmacc/algorithms/math/PowerFunction.hpp
BrianMarre: done

PrometheusPi:
@BrianMarre Could you please squash your commits.

psychocoderHPC (dismissed):

psychocoderHPC:
I squashed the changes because I am not sure if @BrianMarre is available before next year.

PrometheusPi (approved):
test with field absorber example looks good

PR #4399, opened 2022-12-15
Radiation plugin: Workaround BP5 bug

franzpoeschel:
With the BP5 engine, the Radiation plugin currently creates unreadable files:

```
$ bpls radiationOpenPMD_/e_radAmplitudes.bp5
Failed to open with BPFile engine: [Thu Dec 15 10:31:43 2022] [ADIOS2 EXCEPTION] <Core> <IO> <DefineAttribute> : modifiable attribute /data/0/dt has been defined with type double. Type cannot be changed to float

Failed to open with HDF5 engine: [Thu Dec 15 10:31:43 2022] [ADIOS2 EXCEPTION] <Engine> <HDF5ReaderP> <HDF5ReaderP> : Invalid HDF5 file found


Error: Could not open this file with any ADIOS2 file reading engines
```

The Radiation plugin sets some attributes quite late. This is normally supported by the openPMD-api, but can become a bit tricky, when the attribute is a default attribute with a default value. The Radiation plugin uncovers a bug in (depending on your point of view) either openPMD-api or in ADIOS2 with this procedure in the BP5 engine. Realistically the best fix for this would be to set defaults in the openPMD-api as late as possible, and not as early as possible (see the corresponding item in https://github.com/openPMD/openPMD-api/issues/1332). For now, defining these attributes as early as possible in the Radiation plugin is a sensible workaround.

TODO:
- [x] Merge #4401 first

### DIFF
--- a/include/picongpu/plugins/radiation/Radiation.hpp
+++ b/include/picongpu/plugins/radiation/Radiation.hpp
@@ -774,6 +774,13 @@ namespace picongpu
                     ::openPMD::Series& openPMDdataFile = series.value();
                     ::openPMD::Iteration openPMDdataFileIteration = openPMDdataFile.writeIterations()[currentStep];
 
+                    /* begin required openPMD global attributes */
+                    openPMDdataFileIteration.setDt<float_X>(DELTA_T);
+                    const float_X time = float_X(currentStep) * DELTA_T;
+                    openPMDdataFileIteration.setTime(time);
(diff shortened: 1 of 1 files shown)


PrometheusPi:
@franzpoeschel great catch

PrometheusPi (approved):

psychocoderHPC (approved):

PR #4406, opened 2022-12-21
fix radition plugin

psychocoderHPC:
With #4377 a bug was introduced.
Particle slots in a frame were processed even if this slot is no exiting particle (multi mask is zero).

### DIFF
--- a/include/picongpu/plugins/radiation/Radiation.kernel
+++ b/include/picongpu/plugins/radiation/Radiation.kernel
@@ -201,6 +201,12 @@ namespace picongpu
                                          * in a supercell
                                          */
                                         auto par = frameCtx[linearIdx][linearIdx % frameSize];
+                                        /* Not all particle slots in a frame representing an exiting particle.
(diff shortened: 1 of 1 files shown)


PrometheusPi (approved):
How do you ensure that after the `return` in case of `par[multiMask_] == 1`, the tread does not start another task?

PrometheusPi:
Test runs have been submitted to HZDR/hemera. As soon as the results are there and are looking good, I will merge.

psychocoderHPC:
> How do you ensure that after the `return` in case of `par[multiMask_] == 1`, the tread does not start another task?

The return is for the lambda function passed to the lockstep forach algorithm is will not jump out of the kernel or the loop over the frames.
After the foreach which is going over all elements in a frame we have a synchronize to sync all worker and wait before going on with the frequency part of the raditation.

PrometheusPi:
With this fix, we now have a perfect agreement with the previous `dev`. 

grafik

PR #4411, opened 2023-01-06
fix CIC caching shape

psychocoderHPC:
Fix wrong precomputed values for the CIC shape.

The bug was introduced with #4124 and activated by #4170. Only the CIC chape was affected.


steindev:
We compared single particle PIConGPU simulation w/ analytical prediction. They agree now.

steindev (approved):

PR #4412, opened 2023-01-06
Environment variables for support of ADIOS2 SST engine on ORNL Frontier and elsewhere

franzpoeschel:
Two changes:

1. The MPI implementation of the ADIOS2 SST engine requires threaded MPI. Add an environment variable `PIC_USE_THREADED_MPI` to activate threaded MPI.
2. `MPI_Finalize` will hang after using `MPI_Open_Port()` on Frontier. Add an environment variable `PIC_WORKAROUND_CRAY_MPI_FINALIZE` to activate the workaround described at https://github.com/ornladios/ADIOS2/blob/master/docs/user_guide/source/advanced/ecp_hardware.rst.

To use SST on Frontier:
```bash
export PIC_USE_THREADED_MPI=MPI_THREAD_MULTIPLE
export PIC_WORKAROUND_CRAY_MPI_FINALIZE=1
```

**`PIC_WORKAROUND_CRAY_MPI_FINALIZE` is only required for ADIOS2 SST streaming!**

TODO:
- [x] Documentation
- [x] Testing

PR #4413, opened 2023-01-09
Add template for async jobs on Crusher

franzpoeschel:
This PR adds a template similar to the async Summit template [documented here](https://picongpu.readthedocs.io/en/0.6.0/usage/plugins/openPMD.html#experimental-asynchronous-writing) for Crusher.
It somewhat resembles the system setup that we are going to have for the Gordon-Bell thing.

For the kind of workflow that we are trying to aim for, we need to activate a number of workarounds/options, all of which are found in this template:

1. Set `-S 0 --exclusive` batch parameters to avoid hangup of asynchronous GPU sub-tasks
2. Activate threaded MPI to support SST on ECP hardware (for the Python side of things, mpi4py automatically activates threaded MPI, I believe there is not even a way to not activate it), requires #4412
3. Activate [this workaround](https://github.com/ornladios/ADIOS2/blob/master/docs/user_guide/source/advanced/ecp_hardware.rst), requires #4412
4. Set `MPICH_OFI_CXI_PID_BASE` to enable asynchronous MPI jobs at all
5. Use CPU masking

.cfg file for this is `share/picongpu/examples/KelvinHelmholtz/etc/picongpu/6_pipe.cfg`.

TODO:
- [x] Actually test this. The system is currently offline.
- [x] Add .cfg file that corresponds to one Crusher/Frontier node
- [x] Consider recent batch system upgrades → CPU masking

### DIFF
--- a/etc/picongpu/crusher-ornl/batch_pipe.tpl
+++ b/etc/picongpu/crusher-ornl/batch_pipe.tpl
@@ -1,0 +1,247 @@
+#!/usr/bin/env bash
+# Copyright 2013-2022 Axel Huebl, Richard Pausch, Rene Widera, Sergei Bastrakov, Klaus Steinger, Franz Poeschel
+#
+# This file is part of PIConGPU.
+#
+# PIConGPU is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# PIConGPU is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PIConGPU.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
+
+# PIConGPU batch script for crusher's SLURM batch system
+
+#SBATCH --account=!TBG_nameProject
+#SBATCH --time=!TBG_wallTime
+# Sets batch job's name
+#SBATCH --job-name=!TBG_jobName
+#SBATCH --nodes=!TBG_nodes_adjusted
+#SBATCH --gpu-bind=closest
+#SBATCH --chdir=!TBG_dstPath
+# The following two parameters are needed to avoid batch system hangups
+# when trying to run two sub-jobs asynchronously.
+#SBATCH -S 0
+#SBATCH --exclusive # This one might not be strictly necessarily, haven't tested
+
+###################
+# Optional params #
+###################
+##SBATCH -C nvme
+#SBATCH --mail-type=!TBG_mailSettings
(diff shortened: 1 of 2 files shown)


franzpoeschel:
The `.tpl` file is now working, though not in combination with `share/picongpu/examples/KelvinHelmholtz/etc/picongpu/6_pipe.cfg`, since that spreads single MPI ranks across NUMA nodes and the system does not like that. I'll add a cfg file for Crusher.

share/picongpu/examples/KelvinHelmholtz/etc/picongpu/8_pipe.cfg
```diff
@@ -0,0 +1,121 @@
...
+{ \
+  \"rank_table\": \"hostname\", \
+  \"adios2\": { \
+    \"engine\": { \
+      \"parameters\": { \
+        \"QueueLimit\": \"1\", \
+        \"InitialBufferSize\": \"4Gb\", \
+        \"Profile\": \"Off\", \
+        \"Threads\": \"!TBG_coresPerGPU\", \
+        \"MarshalMethod\": \"BP5\" \
```
franzpoeschel, line 78: This is an undocumented parameter of SST which switches on a serialization backend that does not require the `InitialBufferSize` parameter. It does not properly send `vector<string>`-type attributes, so use with care.

franzpoeschel:
I've now added documentation and tested this on Crusher with one and two nodes.
It's ready for review @psychocoderHPC.

etc/picongpu/crusher-ornl/batch_pipe.tpl
```diff
@@ -0,0 +1,247 @@
...
+    sleep 1
+
+    # Need threaded MPI for SST on ECP systems
+    export PIC_USE_THREADED_MPI=MPI_THREAD_MULTIPLE
+    # Enable workaround as described here:
+    # https://github.com/ornladios/ADIOS2/blob/master/docs/user_guide/source/advanced/ecp_hardware.rst
+    export PIC_WORKAROUND_CRAY_MPI_FINALIZE=1
+    export MPICH_OFI_CXI_PID_BASE=$((MPICH_OFI_CXI_PID_BASE+1))
+    # Corresponds to cores 1-8 (exclude first core) in each L3 cache group
+    mask=0xfe,0xfe00,0xfe0000,0xfe000000,0xfe00000000,0xfe0000000000,0xfe000000000000,0xfe00000000000000
```
psychocoderHPC, line 225: is this really required, this should be done by the batch system, or is this required because we use two times srun.

psychocoderHPC (changes requested):

etc/picongpu/crusher-ornl/batch_pipe.tpl
franzpoeschel: This is required because of the two sruns. If you don't specify this, the job will hang up.

psychocoderHPC:
@franzpoeschel could you please rebase and force push again, for an unknown reason the CI is not providing the testing status.

franzpoeschel:
Done.

psychocoderHPC (approved):

PR #4421, opened 2023-01-17
fix CMake optional dependency checking

psychocoderHPC:
In cases where CMake was called already with an optional dependency and then again where we use `PIC_USE_*=OFF` it could be that we try to link the optional dependency library's from the previous cached CMake run.

EDIT by @steindev: We force disable a dependency if the user does not want it. The solution is to set the variable `PIC-HAVE_<name>`  to `FALSE` if the user explicitly denied using a dependency. Setting the variable makes use of `PIC_SEARCH_<name>` which is set according to user's wish.


include/picongpu/CMakeLists.txt
```diff
@@ -89,7 +89,7 @@ function(PIC_dependency_set_status name)
     else()
         set(value FALSE)
     endif()
-    if(${value} AND NOT PIC_HAVE_${name}_LOCKED)
+    if(${value} AND NOT PIC_HAVE_${name}_LOCKED AND PIC_SEARCH_${name})
```
PrometheusPi, line 92: Is this evaluated from left to right or right o left because the logic results differ for 
T, T, F etc.
 - right to left = `(T and not (T and F)) = (T and not F) = (T and T) = T`
 - left to right = ` `((T and not T) and F) = (T and F) and F = F and F = F`
 Could adding parentheses solve this required knowledge about CMake execution order?

PrometheusPi (commented):
Furthermore, the PR adds `PIC_SEARCH_${name}` but the pull request description speaks about `PIC_ADD_*` - could you please elaborate on what this change does in detail or add a comment?

include/picongpu/CMakeLists.txt
psychocoderHPC: it is from left to right. But the result is equal from both sides because `NOT` is only bound to `PIC_HAVE_${name}_LOCKED`

psychocoderHPC: precedence of CMake operators in if conditions: https://cmake.org/cmake/help/latest/command/if.html#condition-syntax

steindev: Since `NOT` operator behavior is well defined, we would forego setting additional (unnecessary) parenthesis. We also don't set additional parenthesis in other places.

steindev (approved):

PR #4423, opened 2023-01-18
calorimeter: remove cuSTL

psychocoderHPC:
- remove cuSTL with PMacc native functions

**This commit is fixing a muti-GPU bug of the calorimeter.**

# Test

FoilLCT example, 1 GPU and 4 GPUs compare step 2048

```
./bin/picongpu_dev -d 1 1 1 -g 256 1280 -s 2100 --e_calorimeter.period 128 --e_calorimeter.file ecalorimeter --e_calorimeter.filter all --e_calorimeter.ext h5
```

calorimenter

The test shows that the dev branch has a multi-GPU bug because single and multi-GPU examples differ.

### DIFF
--- a/include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
+++ b/include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
@@ -30,16 +30,12 @@
 #include "picongpu/plugins/multi/multi.hpp"
 
 #include <pmacc/algorithms/math.hpp>
-#include <pmacc/cuSTL/algorithm/functor/Add.hpp>
-#include <pmacc/cuSTL/algorithm/host/Foreach.hpp>
-#include <pmacc/cuSTL/algorithm/kernel/Foreach.hpp>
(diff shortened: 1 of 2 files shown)


include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
```diff
@@ -237,32 +213,42 @@ namespace picongpu
...
+
+            if(!this->allGPU_reduce->hasResult(mpi::reduceMethods::Reduce()))
                 return;
 
             std::stringstream filename;
             filename << checkpointDirectory << "/" << this->foldername << "/" << filenamePrefix << "_%T."
                      << filenameExtension;
             ::openPMD::Series series(filename.str(), ::openPMD::Access::CREATE);
 
+            auto dataSize64Bit = precisionCast<size_t>(dataSize);
```
PrometheusPi, line 240: How do you ensure that `size_t` is 64 bit? (according to the variable name) 
I assume this is only true on 64bit systems (most systems today)

PrometheusPi (approved):
loos formally ok to me 
just a minor question

PR #4424, opened 2023-01-18
charge conservation plugin: remove cuSTL

psychocoderHPC:
- remove cuSTL with PMacc native functions

I validated the results with our default KHI example for single and multi GPU. The dev branch and this PR produce the same output.

- [x] rebase against #4425

PR #4425, opened 2023-01-19
extend global reduce interface

psychocoderHPC:
Allow specifying if MPI all-reduce or normal reduce (single MPI rank has final data) will be used.

### DIFF
--- a/include/pmacc/algorithms/GlobalReduce.hpp
+++ b/include/pmacc/algorithms/GlobalReduce.hpp
@@ -35,43 +35,88 @@ namespace pmacc
         class GlobalReduce
         {
         public:
+            /** Constructor
+             *
+             * @attetion Don't create a instance before you have set you cupla device!
+             *
+             * @param byte how many bytes in global gpu memory can be reserved for the reduction algorithm
+             * @param sharedMemByte limit the usage of shared memory per block on gpu
+             */
             GlobalReduce(const uint32_t byte, const uint32_t sharedMemByte = 4 * 1024) : reduce(byte, sharedMemByte)
             {
             }
 
             /* Activate participation for reduce algorithm.
-             * Must called from any mpi process. This function use global blocking mpi calls.
-             * Don't create a instance befor you have set you cuda device!
-             * @param isActive true if mpi rank should be part of reduce operation, else false
+             *
+             * @attention Must be called from any mpi process. This function uses global blocking mpi calls.
+             *
+             * @param isActive true if MPI rank should participate in the reduction, else false
              */
             void participate(bool isActive)
             {
+                isParticipating = isActive;
                 mpi_reduce.participate(isActive);
+            }
+
+            /* defines if the result of the MPI operation is valid
+             *
+             * @tparam MPIMethod type of the reduction method
+             * @param method used reduction method e.g.,
+             *                reduceMethods::AllReduce, reduceMethods::Reduce
+             * @return if result of operator() is valid*/
+            template<class MPIMethod>
+            bool hasResult(const MPIMethod& method = ::pmacc::mpi::reduceMethods::AllReduce())
+            {
+                return mpi_reduce.hasResult(method);
             }
 
             /* Reduce elements in global gpu memeory
              *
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
@PrometheusPi @steindev @jkelling can someone merge this PR so that I can rebase https://github.com/ComputationalRadiationPhysics/picongpu/pull/4424

include/pmacc/algorithms/GlobalReduce.hpp
```diff
@@ -35,43 +35,89 @@ namespace pmacc
...
+            }
+
+            /**
+             * @tparam MPIMethod type of the reduction method
+             * @param method used reduction method e.g.,
+             *               reduceMethods::AllReduce, reduceMethods::Reduce
+             *               reduceMethods::AllReduce, reduceMethods::Reduce
+             */
+            template<class Functor, typename Src, class MPIMethod>
+            typename traits::GetValueType<Src>::ValueType operator()(
```
psychocoderHPC, line 97: Note: It was required to have two signatures of this method instead of

```C++
template<class Functor, typename Src, class MPIMethod>
typename traits::GetValueType<Src>::ValueType operator()(Functor func,
              Src src,
              uint32_t n,
              MPIMethod const& method = reduceMethods::AllReduce)
```

else the function was not callable without the method.

include/pmacc/algorithms/GlobalReduce.hpp
```diff
@@ -35,43 +35,89 @@ namespace pmacc
...
             GlobalReduce(const uint32_t byte, const uint32_t sharedMemByte = 4 * 1024) : reduce(byte, sharedMemByte)
             {
             }
 
             /* Activate participation for reduce algorithm.
-             * Must called from any mpi process. This function use global blocking mpi calls.
-             * Don't create a instance befor you have set you cuda device!
-             * @param isActive true if mpi rank should be part of reduce operation, else false
+             *
+             * @attention Must called from any mpi process. This function use global blocking mpi calls.
```
PrometheusPi:
```suggestion
             * @attention Must be called from any mpi process. This function uses global blocking mpi calls.
```

include/pmacc/algorithms/GlobalReduce.hpp
```diff
@@ -35,43 +35,89 @@ namespace pmacc
...
             typename traits::GetValueType<Src>::ValueType operator()(Functor func, Src src, uint32_t n)
+            {
+                return (*this)(func, src, n, ::pmacc::mpi::reduceMethods::AllReduce());
+            }
+
+            /**
+             * @tparam MPIMethod type of the reduction method
+             * @param method used reduction method e.g.,
+             *               reduceMethods::AllReduce, reduceMethods::Reduce
+             *               reduceMethods::AllReduce, reduceMethods::Reduce
```
PrometheusPi: Why two times the same line?

include/pmacc/algorithms/GlobalReduce.hpp
PrometheusPi: could you please elaborate what you mean with this comment?

PrometheusPi (dismissed):
looks good to me - but this was just a formal review 
some minor typos that could be ignored if this is time critical

include/pmacc/algorithms/GlobalReduce.hpp
psychocoderHPC: In general, it would be possible to have only one implementation of the `operator()` but in this case, it is required to have two else I was not able to call the operator without the last parameter which should be replaced by the default in cases the user is not providing it. I pointed to this fact to avoid that the question is coming up during the review.

include/pmacc/algorithms/GlobalReduce.hpp
psychocoderHPC: no reason copy past bug

include/pmacc/algorithms/GlobalReduce.hpp
psychocoderHPC: fixed

PrometheusPi (approved):

PR #4426, opened 2023-01-20
plugin emittance: remove cuSTL

psychocoderHPC:
Replace cuSTL with PMacc natives.

# Tests

I compared the results of the first 800 steps for the Laser Wakefield example from this PR and the dev branch.
The results are in the first digits equal. The small difference I saw is coming from the random particle position in the used example.

### DIFF
--- a/include/picongpu/plugins/Emittance.hpp
+++ b/include/picongpu/plugins/Emittance.hpp
@@ -31,9 +31,6 @@
 #include "picongpu/plugins/multi/multi.hpp"
 #include "picongpu/simulation/control/MovingWindow.hpp"
 
-#include <pmacc/cuSTL/algorithm/functor/Add.hpp>
-#include <pmacc/cuSTL/algorithm/mpi/Gather.hpp>
-#include <pmacc/cuSTL/algorithm/mpi/Reduce.hpp>
 #include <pmacc/dataManagement/DataConnector.hpp>
 #include <pmacc/kernel/atomic.hpp>
 #include <pmacc/lockstep.hpp>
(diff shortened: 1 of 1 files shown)


include/picongpu/plugins/Emittance.hpp
```diff
@@ -334,43 +331,29 @@ namespace picongpu
...
-                    inPlaneGPU[r_element] = 0;
-                    if(inPlaneGPU == pmacc::math::Int<simDim>::create(0))
-                        isGroupRoot = true;
-                }
-                auto* createReduce = new algorithm::mpi::Reduce<simDim>(zoneTransversalPlane, isGroupRoot);
-                if(isInGroup)
-                {
-                    planeReduce = createReduce;
-                    isPlaneReduceRoot = isGroupRoot;
+                    this->isPlaneReduceRoot = mpiReduce->hasResult(::pmacc::mpi::reduceMethods::Reduce{});
```
PrometheusPi, line 343: why is this now a member of the class and not just a local variable?

include/picongpu/plugins/Emittance.hpp
```diff
@@ -334,43 +331,29 @@ namespace picongpu
...
-                    delete createReduce;
             }
 
-            /* Create communicator with ranks of each plane reduce root */
+            /* Create MPI communicator for openPMD IO with ranks of each plane reduce root */
             {
                 /* Array with root ranks of the planeReduce operations */
                 std::vector<int> planeReduceRootRanks(gc.getGlobalSize(), -1);
                 /* Am I one of the planeReduce root ranks? my global rank : -1 */
                 int myRootRank = gc.getGlobalRank() * isPlaneReduceRoot - (!isPlaneReduceRoot);
```
PrometheusPi, line 353: why is no `this->isPlaneReduceRoot` needed here?

PrometheusPi (approved):
looks formally good to me - just some questions

include/picongpu/plugins/Emittance.hpp
psychocoderHPC: It was a member before. You can explicitly say via `this` it is a member or simply use the name. ONly if you have a local variable defined in the function with the same name and you like to assign to the class member you are required to use `this`

PrometheusPi: Thanks for the answer

include/picongpu/plugins/Emittance.hpp
psychocoderHPC: In cases  where `isPlaneReduceRoot == false` we will have `X * 0 - 1 == -1` else `X * 1 - 0 == X`

PrometheusPi: answered by previous comment

PR #4432, opened 2023-01-25
refactor particle filters

psychocoderHPC:
- Add `isDeterministic()` to filters to expose if the filter result is always the same result independent of the input data order.
- update documentation for plugin openPMD
- update param files `particleFilters.param`

The openPMD plugin is multiple times executing the counting and serialization kernel with the same filter. If a filter is not deterministic the openPMD plugin will most likely run into segfault issues because the counting kernel result and the number of particles touched in the serialization kernel will not be equal. The result of the counting kernel is used to allocate an intermediate buffer for the serialized particle data, in cases where we use a non-deterministic filter we serialize too many or not enouph particles into this buffer.

### DIFF
--- a/include/picongpu/particles/filter/All.hpp
+++ b/include/picongpu/particles/filter/All.hpp
@@ -76,6 +76,13 @@ namespace picongpu
                 {
                     return std::string("all");
                 }
+
+                /** A filter is deterministic if the filter outcome is equal between evaluations. If so, set this
+                 * variable to true, otherwise to false.
+                 *
+                 * Example: A filter were results depend on a random number generator must return false.
+                 */
+                static constexpr bool isDeterministic = true;
             };
 
         } // namespace filter
--- a/include/picongpu/particles/filter/RelativeGlobalDomainPosition.hpp
+++ b/include/picongpu/particles/filter/RelativeGlobalDomainPosition.hpp
@@ -138,6 +138,13 @@ namespace picongpu
                     return T_Params::name;
                 }
 
+                /** A filter is deterministic if the filter outcome is equal between evaluations. If so, set this
+                 * variable to true, otherwise to false.
+                 *
+                 * Example: A filter were results depend on a random number generator must return false.
+                 */
+                static constexpr bool isDeterministic = true;
+
                 DataSpace<simDim> localDomainOffset;
                 DataSpace<simDim> globalDomainSize;
             };
--- a/include/picongpu/particles/filter/generic/Free.def
+++ b/include/picongpu/particles/filter/generic/Free.def
@@ -48,6 +48,8 @@ namespace picongpu
                  *           return result;
                  *       }
                  *       static constexpr char const * name = "eachParticleAboveMiddleOfTheCell";
+                 *
+                 *       static constexpr bool isDeterministic = true;
                  *   };
                  *
                  *   using EachParticleAboveMiddleOfTheCell = generic::Free<
--- a/include/picongpu/particles/filter/generic/FreeRng.def
+++ b/include/picongpu/particles/filter/generic/FreeRng.def
@@ -57,6 +57,8 @@ namespace picongpu
                  *           return result;
                  *       }
                  *       static constexpr char const * name = "eachSecondParticle";
+                 *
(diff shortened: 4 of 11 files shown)


include/picongpu/particles/filter/generic/Free.def
```diff
@@ -48,6 +48,8 @@ namespace picongpu
                  *           return result;
                  *       }
                  *       static constexpr char const * name = "eachParticleAboveMiddleOfTheCell";
+                 *
+                 *       static constexpr bool isDeterministic = true;
```
steindev, line 52: Missing comment.

include/picongpu/particles/filter/All.hpp
```diff
@@ -76,6 +76,12 @@ namespace picongpu
                 {
                     return std::string("all");
                 }
+
```
steindev, line 79: The construction of this sentence is not easy to read. maybe
```
A filter is deterministic if the filter outcome is equal between evaluations.
If so, set this variable to true, otherwise to false.

Example: A filter were results depend on a random number generator must return false.
```

include/picongpu/particles/filter/generic/FreeRng.def
```diff
@@ -57,6 +57,8 @@ namespace picongpu
                  *           return result;
                  *       }
                  *       static constexpr char const * name = "eachSecondParticle";
+                 *
+                 *       static constexpr bool isDeterministic = false;
```
steindev, line 61: Missing comment.

include/picongpu/particles/filter/generic/FreeTotalCellOffset.def
```diff
@@ -58,6 +58,8 @@ namespace picongpu
                  *           return result;
                  *       }
                  *       static constexpr char const * name = "eachParticleInXCell5";
+                 *
+                 *       static constexpr bool isDeterministic = true;
```
steindev, line 62: Missing comment.

include/picongpu/particles/filter/generic/FreeRng.def
steindev: wait this is within a comment. Hmmm,...

include/picongpu/particles/filter/generic/FreeTotalCellOffset.def
steindev: also within comment

steindev (changes requested):

include/picongpu/particles/filter/generic/Free.def
steindev: wait, this is within a comment. I don't know if this is helpful without any explanation.

psychocoderHPC: We skipped comments in comments. One problem is that these comments will be outdated soon and it will trigger warnings in the compiler.

psychocoderHPC:
I updated the comment, thx for the review!

share/picongpu/benchmarks/TWEAC-FOM/include/picongpu/param/particleFilters.param
```diff
@@ -49,6 +49,12 @@ namespace picongpu
             {
                 static constexpr char const* name = "forwardPinhole";
 
+                /** true if the filter is producing deterministic results independent else false.
```
steindev: Forgot to update the comment here?

include/picongpu/particles/filter/RelativeGlobalDomainPosition.hpp
```diff
@@ -138,6 +138,12 @@ namespace picongpu
                     return T_Params::name;
                 }
 
+                /** true if the filter is producing deterministic results independent else false.
```
steindev: Forgot to update the comment here?

include/picongpu/particles/filter/All.hpp
steindev: Forgot to update the comment here?

steindev (changes requested):

psychocoderHPC:
I fixed the parts where missed to update the comment.

steindev (approved):

PR #4436, opened 2023-01-26
run CI down-stream pipes parallel

psychocoderHPC:
Due to a bug in the gitlab CI it was in the past not possible to run two down-stream pipes in parallel. 
In this case, the status of the CI could be wrong.
This bug is now fixed in gitlab, therefore we can run now PIConGPU and PMacc tests in parallel.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,8 +2,7 @@
 stages:
   - validate
   - generate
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
To avoid blocking the CI to much I decided to merge this change together with #4437

PR #4437, opened 2023-01-26
optimize CI

psychocoderHPC:
Due to a bug in the gitlab CI it was in the past not possible to run two down-stream pipes in parallel. In this case, the status of the CI could be wrong. This bug is now fixed in gitlab, therefore we can run now PIConGPU and PMacc tests in parallel.

- run examples, tests, and benchmarks in parallel downstream pipes
- run PIConGPU and PMacc pipes in parallel
- shuffle the job list in the generator to avoid all jobs for one
  backends will always be in the first stage

### DIFF
--- a/share/ci/n_wise_generator.py
+++ b/share/ci/n_wise_generator.py
@@ -7,6 +7,7 @@ from allpairspy import AllPairs
 import argparse
 import sys
 import math
+import random
 
 parser = argparse.ArgumentParser(description='Generate tesing pairs')
 parser.add_argument('-n', dest='n_pairs', default=1, action="store",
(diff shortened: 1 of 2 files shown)


share/ci/n_wise_generator.py
```diff
@@ -248,6 +249,12 @@ def is_valid_combination(row):
                  filter_func=is_valid_combination, n=n_pairs)):
         job_list.append(value)
 
+# set seed to be deterministic in each CI run
+random.seed(42)
+# Shuffle the job list to avoid that a to many jobs testing the
+# same backend are run at the same time.
```
PrometheusPi:
```suggestion
# Shuffle the job list to avoid that too many jobs, testing the
# same backend, run at the same time.
```

PrometheusPi (changes requested):
some typos

share/ci/n_wise_generator.py
psychocoderHPC: fixed

PrometheusPi (approved):

Issue #4439, opened 2023-01-26
Implementation of struct filter::generic::FreeRng needed?


steindev:
While working with Rng based particle filters in `ManipulateDerive<...>` in `speciesInitialization.param` I noted, that the implementation of filter::generic::FreeRng is broken.

[The shorthand definition of RngType in line 89 of include/picongpu/particles/filter/generic/FreeRng.hpp](https://github.com/ComputationalRadiationPhysics/picongpu/blob/59e9b53605f9a5c1bf271eeb055bc74370a99052/include/picongpu/particles/filter/generic/FreeRng.hpp#L89) is failing.

However, it is not needed anyways, as `RngType` can be replaced with `auto` and `ALPAKA_DECAY_T(decltype(rng))` in subsequent lines, just as in `include/picongpu/particles/manipulators/generic/FreeRng.hpp`

The point of this issue is to discuss whether the implementation of `filter::generic::FreeRng` is needed at all, as it is identical to the implementation of `manipulators::generic::FreeRng`, except for the return type of `operator()` in `struct generic::acc::FreeRng : private T_Functor`.

psychocoderHPC:
As you mentioned the return type is different. Filters and Manipulator have therefore a different interface. PIConGPU is compile time validating if the interface is fitting https://github.com/ComputationalRadiationPhysics/picongpu/blob/59e9b53605f9a5c1bf271eeb055bc74370a99052/include/picongpu/particles/filter/IUnary.def#L40

Could you please post the compile time error!

psychocoderHPC:
error message is:

```
picongpu/include/picongpu/../picongpu/particles/filter/generic/FreeRng.hpp(89): error: class "picongpu::particles::functor::misc::Rng<picongpu::particles::atomicPhysics::DistributionFloat>" has no member "RandomGen"
          detected during:
            instantiation of class "picongpu::particles::filter::generic::FreeRng<T_Functor, T_Distribution> [with T_Functor=picongpu::particles::filter::FunctorEachSecondParticle, T_Distribution=pmacc::random::distributions::Uniform<picongpu::float_X, pmacc::random::methods::RngPlaceholder>]" 
/opt/spack-modules/linux-ubuntu20.04-skylake_avx512/gcc-10.2.0/boost-1.75.0-rhit3f5bvbmaffhbwi27c6vhvf5ip7iw/include/boost/mpl/aux_/preprocessed/gcc/apply_wrap.hpp(38): here
[...]

PR #4440, opened 2023-01-26
Remove explicit definition of RngType in filter::generic::FreeRng

steindev:
See introduction of issue #4439. This PR solves the described initial problem.

PR #4442, opened 2023-01-27
fix picongpu RNG filer

psychocoderHPC:
fix bug mentioned in #4439

The bug was introduced with #4321

thx @steindev for reporting the bug and pointing to the fix.


steindev:
See (https://github.com/ComputationalRadiationPhysics/picongpu/pull/4440)

PR #4443, opened 2023-01-27
cleanup PMacc's event system

psychocoderHPC:
The motivation behind this PR is to reduce the usage of macros and avoid problems with cyclic includes.
In the past, I refactored PMacc so that we can access all manager singletons in PMacc via the god object `Environment`.
I know now that this was not a good idea because it introduces a lot of issues with cyclic includes and avoids separating non-template PMacc interfaces in `hpp`-files and implementations.

- remove macros `__startTransaction(...), ...`
  - `eventSystem.hpp` is providing for the user all necessary entry functions for the event system
- separate more classes into hpp,cpp files remove `Manager()`, `TransactionManager()` singleton access from the god object `Environment`

This PR should not affect any user input of PIConGPU, only the developer must take care that you need to use the interfaces in `pmacc/eventSystem/eventSystem.hpp` to interact with the event system of PMacc.


include/pmacc/Environment.def
```diff
@@ -32,5 +33,100 @@ namespace pmacc
     {
         struct Environment;
 
+        /** collect state variables of the environment context
+         *
+         * This class handle the initialization and finalize of the
+         * MPI context and the selection of the GPU.
+         */
+        class EnvironmentContext
```
psychocoderHPC, line 41: This class is moved into the `def` file because it is required to take care that the transaction system is not started before we selected a device.
Moving this class here is no problem because it is not pulling files with include side effects.

include/pmacc/particles/tasks/TaskParticlesSend.hpp
```diff
@@ -22,7 +22,8 @@
 #pragma once
 
 #include "pmacc/Environment.hpp"
-#include "pmacc/eventSystem/EventSystem.hpp"
+#include "pmacc/eventSystem/Manager.hpp"
+#include "pmacc/eventSystem/tasks/MPITask.hpp"
```
PrometheusPi, line 26: @psychocoderHPC  What was added in the code that now needs `MPITask.hpp`

psychocoderHPC: before it was transitively included via `EventSystem.hpp`

include/pmacc/eventSystem/streams/EventStream.cpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2013-2022 Heiko Burau, Rene Widera, Alexander Grund
+/* Copyright 2013-2022 Felix Schmitt, Rene Widera, Benjamin Worpitz
```
PrometheusPi, line 1: What happened here with the authors?

include/pmacc/eventSystem/streams/EventStream.cpp
```diff
@@ -19,12 +19,32 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
-#pragma once
```
PrometheusPi, line 22: Why did you remove the `#pragma once`?

PrometheusPi (approved):

include/pmacc/eventSystem/streams/EventStream.cpp
psychocoderHPC: The files got removed but github tries to link files and is thinking we renamed the file.
So this change is OK.

include/pmacc/eventSystem/streams/EventStream.cpp
psychocoderHPC: In cpp files you can not have pragma once. This file is new but github is thinking the file is moved.

Issue #4446, opened 2023-01-27
Number of macro particles between source and derived species do not match


steindev:
In my current setup I have 3 electron and one ion species. The three electron species represent (1) a spatially homogenous background distribution, (2) an inhomogenous spatial distribution on top of the homogenous background, and (3) a subsample of the inhomogenous distribution used to calculate radiation. The ion species is supposed to ensure charge neutrality at the beginning of the simulation, but ion movement in general is important for the scenario.

With the following `InitPipeline` I would expect as many ion macro-particles as electron-like macro-particles
```cpp
using InitPipeline = bmpl::vector<
  /* current sheet ions */
CreateDensity<densityProfiles::SechSquare, startPosition::RandomPositionAndWeighting, Ions>
  /* current sheet electrons */
, Derive<Ions, CurrentSheetElectrons>
  /* Derive RadiationElectrons from CurrentSheetElectrons */
, ManipulateDerive<manipulators::DeleteCurrentSheetElectrons, CurrentSheetElectrons, RadiationElectrons, filter::RadiationElectronsSelection>
, particles::FillAllGaps<CurrentSheetElectrons>
  /* background electrons */
, CreateDensity<densityProfiles::Homogenous, startPosition::RandomPositionAndWeighting, BackgroundElectrons>
  /* background ions */
, Derive<BackgroundElectrons, Ions>
  /* Add temperature to particles */
, Manipulate<manipulators::AddTemperature, Ions>
, Manipulate<manipulators::AddTemperature, BackgroundElectrons>
, Manipulate<manipulators::AddTemperature, CurrentSheetElectrons>
, Manipulate<manipulators::AddTemperature, RadiationElectrons>
>;
```
But at simulation start there is a difference between `BackgroundElectrons+CurrentSheetElectrons+RadiationElectrons` and `Ions` of
`902430720+2421593+241305454-1144406016 = 1751751`, according to plugin `macroParticleCounter`. (Names and numbers correspond within this sentence.)

However, when commenting the movement of `CurrentSheetElectrons` to `RadiationElectrons` (including `FillAllGaps`), there is an equal amount of electrons (background+current sheet) and ions!

[...]

PR #4447, opened 2023-01-30
improve command line argument validation

psychocoderHPC:
- Validate that the number of values passed via `--gridDist, --grid, ...` is not more than allowed.
- Validate the correctness of particle boundary settings only once.
- Move the help for particle boundaries, pml, ... behind the basic parameters e.g. `--devices`, ...

This PR helps to avoid issues like https://github.com/ComputationalRadiationPhysics/picongpu/issues/4446

Error message example:
```
Unhandled exception of type 'St13runtime_error' with message 'expression (offsets().size() <= 3) 
failed in file (picongpu/include/picongpu/../picongpu/simulation/stage/ParticleBoundaries.hpp:108) : 
Invalid number of particle boundary offsets for species 'ec'.', terminating
```

### DIFF
--- a/include/picongpu/simulation/control/Simulation.hpp
+++ b/include/picongpu/simulation/control/Simulation.hpp
@@ -116,11 +116,7 @@ namespace picongpu
         void pluginRegisterHelp(po::options_description& desc) override
         {
             SimulationHelper<simDim>::pluginRegisterHelp(desc);
-            currentInterpolationAndAdditionToEMF.registerHelp(desc);
-            fieldAbsorber.registerHelp(desc);
-            fieldBackground.registerHelp(desc);
-            particleBoundaries.registerHelp(desc);
-            runtimeDensityFile.registerHelp(desc);
+
             // clang-format off
             desc.add_options()(
                 "versionOnce", po::value<bool>(&showVersionOnce)->zero_tokens(),
@@ -153,6 +149,12 @@ namespace picongpu
                 ("numRanksPerDevice,r", po::value<uint32_t>(&numRanksPerDevice)->default_value(1u),
                  "set the number of MPI ranks using a single device together");
             // clang-format on
+
+            currentInterpolationAndAdditionToEMF.registerHelp(desc);
+            fieldAbsorber.registerHelp(desc);
+            fieldBackground.registerHelp(desc);
+            particleBoundaries.registerHelp(desc);
+            runtimeDensityFile.registerHelp(desc);
         }
 
         std::string pluginGetName() const override
@@ -166,26 +168,28 @@ namespace picongpu
             while(periodic.size() < 3)
                 periodic.push_back(0);
 
+
+            PMACC_VERIFY_MSG(
+                devices.size() >= 2 && devices.size() <= 3,
+                "Invalid number of devices.\nuse [-d dx=1 dy=1 dz=1]");
+
             // check on correct number of devices. fill with default value 1 for missing dimensions
-            if(devices.size() > 3)
-            {
-                std::cerr << "Invalid number of devices.\nuse [-d dx=1 dy=1 dz=1]" << std::endl;
-            }
-            else
-                while(devices.size() < 3)
-                    devices.push_back(1);
+            while(devices.size() < 3)
+                devices.push_back(1);
+
             // check for request of > 1 device in z for a 2d simulation, this is probably a user's mistake
             if((simDim == 2) && (devices[2] > 1))
                 std::cerr
                     << "Warning: " << devices[2] << " devices requested for z in a 2d simulation, this parameter "
                     << "will be reset to 1. Number of MPI ranks must be equal to the number of devices in x * y\n";
 
+
+            PMACC_VERIFY_MSG(
+                gridSize.size() >= 2 && gridSize.size() <= 3,
+                "Invalid or missing grid size.\nuse -g width height [depth=1]");
+
             // check on correct grid size. fill with default grid size value 1 for missing 3. dimension
-            if(gridSize.size() < 2 || gridSize.size() > 3)
-            {
-                std::cerr << "Invalid or missing grid size.\nuse -g width height [depth=1]" << std::endl;
-            }
-            else if(gridSize.size() == 2)
+            if(gridSize.size() == 2)
                 gridSize.push_back(1);
 
             if(slidingWindow && devices[1] == 1)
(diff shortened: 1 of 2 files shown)


include/picongpu/simulation/control/Simulation.hpp
```diff
@@ -166,26 +168,28 @@ namespace picongpu
...
+
             // check on correct number of devices. fill with default value 1 for missing dimensions
-            if(devices.size() > 3)
-            {
-                std::cerr << "Invalid number of devices.\nuse [-d dx=1 dy=1 dz=1]" << std::endl;
-            }
-            else
-                while(devices.size() < 3)
-                    devices.push_back(1);
+            while(devices.size() < 3)
```
steindev, line 177: `while()` here but `if(` in line 192. Homogenize?

include/picongpu/simulation/control/Simulation.hpp
```diff
@@ -218,6 +222,8 @@ namespace picongpu
                 }
             }
 
+            PMACC_VERIFY_MSG(gridDistribution.size() <= 3, "Invalid number of grid distributions.");
```
steindev: The other asserts have a comment how to use the program option correctly, see e.g. line 187. You think it is not necessary here?

include/picongpu/simulation/stage/ParticleBoundaries.hpp
```diff
@@ -98,54 +98,77 @@ namespace picongpu
                      */
                     void operator()()
                     {
-                        for(uint32_t d = 0; d < simDim; d++)
+                        static bool validateUserInputs = true;
+                        if(validateUserInputs)
```
steindev, line 102: Why encapsulate this in an `if()`? The condition is always true anyways.

include/picongpu/simulation/stage/ParticleBoundaries.hpp
```diff
@@ -98,54 +98,77 @@ namespace picongpu
...
+                                {
+                                    if(T_Species::boundaryDescription()[d].kind == particles::boundary::Kind::Periodic)
+                                        throw std::runtime_error(
+                                            "Boundary kind " + errorString
+                                            + " is not compatible with --periodic value");
+                                    T_Species::boundaryDescription()[d].kind = particles::boundary::Kind::Reflecting;
+                                }
+                                if(kindName == "thermal")
+                                {
+                                    if(T_Species::boundaryDescription()[d].kind == particles::boundary::Kind::Periodic)
```
steindev, line 144: There is something wrong, either here in the condition, her in the error message, or in the check above.

include/picongpu/simulation/stage/ParticleBoundaries.hpp
```diff
@@ -98,54 +98,77 @@ namespace picongpu
...
+                                            + " is not compatible with --periodic value");
+                                }
+                                if(kindName == "absorbing")
+                                {
+                                    // For now it must match the default-set boundary kind
+                                    if(T_Species::boundaryDescription()[d].kind
+                                       != particles::boundary::Kind::Absorbing)
+                                        throw std::runtime_error(
+                                            "Boundary kind " + errorString
+                                            + " is not compatible with --periodic value");
```
steindev, line 165: Again, I believe either here or in the check above something is wrong. Probably the error message.
Isn't it necessary to set the `T_Species::boundaryDescription()[d].kind` here or in the check above as well?

steindev (commented):
I have some questions. It is hard to follow the logic in the checks performed in `ParticleBoundaries.hpp`

include/picongpu/simulation/stage/ParticleBoundaries.hpp
psychocoderHPC: there is no need to validate it more than once

include/picongpu/simulation/stage/ParticleBoundaries.hpp
psychocoderHPC: What should be wrong here? I have not changed anything in this PR.

include/picongpu/simulation/control/Simulation.hpp
psychocoderHPC: Could you explain what you see here as a problem? This line is only filling up the vector `devices` with `1`
Line 192 is extending the `gridSize` by `1` in case we run a 2D simulation.

include/picongpu/simulation/control/Simulation.hpp
psychocoderHPC: In that case, I would not try to guide the user because there are many ways to user grid distributions.

include/picongpu/simulation/stage/ParticleBoundaries.hpp
psychocoderHPC: The condition is true for the first evaluation and then set to false and is not evaluated a second time.

include/picongpu/simulation/stage/ParticleBoundaries.hpp
psychocoderHPC: What is wrong, if so we need to handle it in a separate PR, there are no changes in these lines only the indention is changed because of the additional if condition.

include/picongpu/simulation/control/Simulation.hpp
psychocoderHPC: changed to `"Too many grid distribution directions given.  A maximum of three directions are supported."`

include/picongpu/fields/incidentField/profiles/Free.def
```diff
@@ -30,7 +30,7 @@ namespace picongpu
         {
             namespace profiles
             {
-                /** Free profile tag using given functors
+                /** Free profile tag usingInvalid number of grid distributions. given functors
```
PrometheusPi: I do not understand that comment/description - but this is fine

PrometheusPi (dismissed):
@steindev you had a number of comments. What's your current status on this pull request?

include/picongpu/fields/incidentField/profiles/Free.def
psychocoderHPC: This should not be part of the PR, I think I was in the wrong window of my IDE and past it in.

include/picongpu/simulation/stage/ParticleBoundaries.hpp
PrometheusPi: If there is an error, we need a separate PR as suggested by @psychocoderHPC

PrometheusPi (approved):

PR #4448, opened 2023-01-30
remove buffer `*Intern`

psychocoderHPC:
The `*Intern` naming of buffers is a result of the very beginning of PIConGPU and was a try to separate the interfaces for buffers a user is allowed to use and what PMacc can be used internally. In old times a user never created a Host or Device buffer, instead, always the GridBuffer was used. Over time we saw that this idea to hide buffer creation was wrong but never found time to refactor this part of the code.

# Review

There is no new code in the files. I only merged the `*BufferIntern` interfaces into the `*Buffer` file.


steindev:
@pordyna Would you "like" to review the changes?

include/pmacc/memory/buffers/Exchange.hpp
```diff
@@ -41,19 +52,218 @@ namespace pmacc
...
+                if(mask.containsExchangeType(RIGHT))
+                {
+                    tmp_offset[0] = size[0] - border[0] - guardingCells[0];
+                    if(area == GUARD)
+                    {
+                        tmp_offset[0] += guardingCells[0];
+                    }
+                    /* std::cout<<"offset="<<tmp_offset[0]<<"border"<<border[0]<<std::endl;*/
+                }
+                else
```
PrometheusPi, line 173: Can it happen, that a DataSpace contains both left and right exchange types? 
If yes, this would not be covered by this if/else branch.

PrometheusPi (dismissed):
Looks okay to me. It would be great if you @pordyna could have a second look. This is not a part of PIConGPU I am very familiar with.

include/pmacc/memory/buffers/Exchange.hpp
psychocoderHPC: No, it can only be LEFT or RIGHT.

I will remove the std our in the  line above!

include/pmacc/test/memory/HostBuffer/copyFrom.hpp
```diff
@@ -52,29 +52,29 @@ namespace pmacc
                         {
                             ::pmacc::DataSpace<T_Dim::value> const dataSpace
                                 = ::pmacc::DataSpace<T_Dim::value>::create(nElementsPerDim[i]);
-                            ::pmacc::HostBuffer<Data, T_Dim::value>* hostBufferIntern
-                                = new ::pmacc::HostBufferIntern<Data, T_Dim::value>(dataSpace);
-                            ::pmacc::DeviceBuffer<Data, T_Dim::value>* deviceBufferIntern
-                                = new ::pmacc::DeviceBufferIntern<Data, T_Dim::value>(dataSpace);
+                            ::pmacc::HostBuffer<Data, T_Dim::value>* HostBuffer
```
pordyna: Why `HostBuffer` and not `hostBuffer` (the same for `Device`Buffer` and other files in tests)? Shouldn't it start with lowercase, since it is not a type name?

include/pmacc/test/memory/HostBuffer/setValue.hpp
```diff
@@ -51,12 +51,12 @@ namespace pmacc
                         {
                             ::pmacc::DataSpace<T_Dim::value> const dataSpace
                                 = ::pmacc::DataSpace<T_Dim::value>::create(nElementsPerDim[i]);
-                            ::pmacc::HostBufferIntern<Data, T_Dim::value> hostBufferIntern(dataSpace);
+                            ::pmacc::HostBuffer<Data, T_Dim::value> HostBuffer(dataSpace);
```
pordyna: Should probably be lowercase

include/pmacc/test/memory/HostBuffer/reset.hpp
```diff
@@ -51,13 +51,13 @@ namespace pmacc
                         {
                             ::pmacc::DataSpace<T_Dim::value> const dataSpace
                                 = ::pmacc::DataSpace<T_Dim::value>::create(nElementsPerDim[i]);
-                            ::pmacc::HostBufferIntern<Data, T_Dim::value> hostBufferIntern(dataSpace);
+                            ::pmacc::HostBuffer<Data, T_Dim::value> HostBuffer(dataSpace);
```
pordyna: Should probably be lowercase

include/pmacc/memory/buffers/Exchange.hpp
```diff
@@ -80,19 +289,32 @@ namespace pmacc
...
+         */
+        std::unique_ptr<HostBuffer<TYPE, DIM>> hostBuffer;
+
+        //! This buffer is a vector which is used as message buffer for faster memcopy
+        std::unique_ptr<DeviceBuffer<TYPE, DIM>> deviceDoubleBuffer;
+        std::unique_ptr<DeviceBuffer<TYPE, DIM>> deviceBuffer;
 
-        uint32_t exchange;
-        uint32_t communicationTag;
+        uint32_t exchange = std::numeric_limits<uint32_t>::max();
```
pordyna, line 316: What's the reason for initializing the two variables like that? These are set in both constructors anyway. Is there anything special about the max uint32 value?

include/pmacc/memory/buffers/HostBuffer.hpp
```diff
@@ -38,50 +41,135 @@ namespace pmacc
     class Buffer;
 
     /**
-     * Interface for a DIM-dimensional Buffer of type TYPE on the host
-     *
-     * @tparam TYPE datatype for buffer data
-     * @tparam DIM dimension of the buffer
+     * Internal implementation of the HostBuffer interface.
```
pordyna: You probably want the docstring from the old `HostBuffer` and not the internal implementation. Switch + with -.

pordyna (changes requested):
There are just a few small things I found. But in total, I compared the old intern classes with the new merged once with git diff, and it looks like you are really just removing the virtual interface in between.

include/pmacc/test/memory/HostBuffer/reset.hpp
psychocoderHPC: ohh yes this is wrong, I assume I used search and replace wrong during the change.

include/pmacc/memory/buffers/Exchange.hpp
psychocoderHPC: I added the default so that everyone can see that this will definitively be initialized with something valid.
Any time where I see a member without a default I am asking myself if the value is set in the constructor or not.
I always need to find all constructors to be sure that they will have a valid value.  By giving a default here I know now even without a lookup that it has a valid value even if it is not part of the constructor.

psychocoderHPC:
@PrometheusPi I changed the issues @pordyna found.

include/pmacc/test/memory/HostBuffer/setValue.hpp
PrometheusPi: @psychocoderHPC Why was this not caught by the CI?

PrometheusPi (approved):

PrometheusPi:
Thanks @pordyna for your detailed review

include/pmacc/test/memory/HostBuffer/setValue.hpp
psychocoderHPC: It is not a bug, the compiler knows that the right side must be a variable name and not a type.
Nevertheless it is not following our naming rules.

PR #4451, opened 2023-02-03
fix incident field calculation order

psychocoderHPC:
The incident field is updating the electric field for the CORE and BORDER in different orders.
By moving the electric field CORE update after the incident field update the calculation order for CORE and BORDER is equal. 
The original code should not be wrong because updating the electric field is using an addition and this operation is commutative.


steindev (approved):
I checked both of your arguments (inconsistency in usage and permutation invariance) in the code and agree.

PR #4459, opened 2023-02-14
switch PICMI to incidentField

BrianMarre:
changes PICMI-interface generation to incidentField.param
- switch from x-z-init plane to y_min huygens surface

also makes available a bunch of PICMI settings now representable with the new Lasers, namely:
- centroid specification, location of laser maximum at t=0
- polarization direction specification as vector
- propagation direction specification as vector

passed PICMI-interface and PyPIConGPU unit test, we will see if we are missing tests if people start using it ;)


lib/python/picongpu/picmi/gaussian_laser.py
```diff
@@ -42,31 +84,56 @@ def get_as_pypicongpu(self) -> laser.GaussianLaser:
...
+
+        # check that initialising from y_min-plane only is sensible
+        assert (self.scalarProduct(self.propagation_direction, \
+                [0., 1., 0.]) > 0.), \
+            "laser propagation parallel to the y-plane or pointing outside \
+            from the inside of the simulation box is not supported by this \
+            laser in PIConGPU"
+        assert ((self.propagation_direction[0]**2
+                + self.propagation_direction[1]**2
+                + self.propagation_direction[2]**2) - 1. < 1e-9), \
```
psychocoderHPC: Is there something like `epsilon` in python available that we do not need to use a magic number `1e-9` here?

lib/python/picongpu/picmi/gaussian_laser.py
```diff
@@ -42,31 +84,56 @@ def get_as_pypicongpu(self) -> laser.GaussianLaser:
...
-            "focal position x and z MUST be equal to centroid x and z"
-        assert [0, 1, 0] == self.propagation_direction, \
-            "only support propagation along Y axis"
-        assert 0 == self.centroid_position[1], "centroid MUST have y=0"
+        assert abs(self.scalarProduct(self.polarization_direction,
+                self.polarization_direction) - 1) < 1e-9, \
+            "the polarization direction vector must be normalized"
+
+        # check for excessive phase values to avoid numerical precision errors
+        assert abs(self.picongpu_phase) <= 2*3.1416, \
```
psychocoderHPC: Do we need a hard coded PI, I assume python has a variable for that.

lib/python/picongpu/picmi/gaussian_laser.py
```diff
@@ -17,22 +18,63 @@
...
+            picongpu_phase: float = 0.,
+
+            # make sure to always place Huygens-surface inside PML-boundaries,
+            # default is valid for standard PMLs
+
+            # @todo create check for insufficient dimension
+            # @todo create check in simulation for conflict between PMLs and
+            # Huygens-surfaces
+            picongpu_huygens_surface_positions: typing.List[
+                typing.List[int]] = [[16, -16], [16, -16], [16, -16]],
```
psychocoderHPC, line 55: Do we have in PICME the possibility to define how thick the PML area is? If so these values should be `pml_cells + N` where N>1 (normally we need to know the stencil with of the field solver.

psychocoderHPC (changes requested):
No hard requirements, but more questions pop up during a quick first review.

lib/python/picongpu/picmi/gaussian_laser.py
BrianMarre: No, the PIMCI interface itself only allows setting the following
```
   lower_boundary_conditions: vector of strings
        Conditions at lower boundaries, periodic, open, dirichlet, or neumann
    upper_boundary_conditions: vector of strings
        Conditions at upper boundaries, periodic, open, dirichlet, or neumann
```
and PyPIConGPU does not have a template for FieldAbsorber.param, therefore we rely on the default setting for them. That also means that PyPIConGPU, right now, has no clue what the actual pml settings are unfortunately.
And yes I want to change that, even left a todo in the code, but not now.

lib/python/picongpu/picmi/gaussian_laser.py
BrianMarre: Yes, but it requires another non negligibel dependency
```
>>> import sys
>>> sys.float_info.epsilon
2.220446049250313e-16
```
Also I am not sure whether we want to test against machine epsilon necessarily.

BrianMarre: I could parametrize it if you prefer?

lib/python/picongpu/picmi/gaussian_laser.py
BrianMarre: Done

PrometheusPi:
@BrianMarre What is the status of this pull request? 
If you are still actively working on it, could you set it to draft?

BrianMarre:
still working actively on it, whenever I can steal a hour from somewhere

lib/python/picongpu/picmi/gaussian_laser.py
BrianMarre: offline discussion result: separate checks to separate function for better readability

BrianMarre: done

share/picongpu/pypicongpu/template/include/picongpu/param/png.param.mustache
```diff
@@ -35,18 +37,21 @@ namespace picongpu
...
+        // 4: Thermal:  [outdated]
+        // 5: BlowOut:  typical fields, assuming that a LWFA in the blowout
+        //              regime causes a bubble with radius of approx. the laser's
+        //              beam waist (use for bubble fields)
+        // 6: Custom:   user-provided normalization factors via customNormalizationSI
+        // 7: Incident: typical fields calculated out of the incident field amplitude,
+        //              uses max amplitude from all enabled incident field profile types ignoring Free
 {{#laser}}
-#define EM_FIELD_SCALE_CHANNEL1 1
+#define EM_FIELD_SCALE_CHANNEL1 7
```
psychocoderHPC, line 54: I am not sure what happens if we define no laser with PICMI. In that case it could be that value `7` is not valid.
@BrianMarre can PICMI create a setup without a laser?

psychocoderHPC (approved):
I looked over the code but need to say that I do not understand most of the parts, I have the feeling that somehow we have too much python code for such a simple change.
I do not say that we can simplify it but somehow I have the feeling that maintaining the python code is very time intensive.

test/python/picongpu/compiling/runner.py
```diff
@@ -120,7 +120,7 @@ def test_build(self):
         # but: injecting a generated (but not built) dir
         # from another runner works
         other_runner = Runner(self.sim)
-        other_runner.generate()
+        other_runner.generate(printDirToConsole=True)
```
psychocoderHPC, line 123: Why is this changed? Could it be this is a left over from testing?

psychocoderHPC: I saw this in many other files too.

psychocoderHPC (changes requested):

share/picongpu/pypicongpu/template/include/picongpu/param/png.param.mustache
BrianMarre: yes it can, unless I made an error somewhere.

And regarding your implied question, in this case it works due to the following
```
{{#laser}} // if a laser exist write the following
#define EM_FIELD_SCALE_CHANNEL1 7
{{/laser}}
{{^laser}} // if no laser exists write the following
#define EM_FIELD_SCALE_CHANNEL1 -1
{{/laser}}
```

test/python/picongpu/compiling/runner.py
BrianMarre: That is a quality of live improvement for a maintainer using the compile tests.
If the developer made an error in the templates or schema only, the quick tests will pass, since they do not test the that the generated picInput is acutally valid for PIConGPU(compiles).
Only the compiling tests will fail, but they fail with a cryptic error message, along the lines of "internal process has failed", the debugSwitch makes it so that in addition to the error message the runner prints the temporary (and automatically generated) path it generates the picInput to before actually running `pic-build`.
On a fail you may then go to the temp directory and manually call `pic-build` yourself to see the actual PIConGPU compile error messages and figure out what went wrong.
Before you had no idea where to look without looking into the code nor any idea which of the several picInputs had actually generated the error.

BrianMarre:
breakdown of the 944 changed lines:
+15 lines to docs, describing better how to debug the compile test failures
+/- 268 lines in the tests, generalising existing tests and changing setup methods
+/- 219 lines in actual code changes
+/- 90 lines in changing the schema (description how the serialized description must look like)
+/-338 lines in changing the templates (param-files in which we paste values from serialized description):

     -102 removing the old laser.param template
     +207 adding the new incidentField.param
     +/- 29 in the png.param, mostly from copying over the comments explaining the different modes

 rest is some spelling corrections and the above mentioned quality of live improvement

psychocoderHPC (approved):

PR #4460, opened 2023-02-15
Topic lockstep for each refactoring

psychocoderHPC:
Extend the lockstep foreach algorithm interface with the possibility to forward lockstep context variables to the user functor.
A variable forwarded to the functor can be accessed without any index operations.

- refactor forEach
  - separate the functor execution in hierarchical helper to simplify the     user functor signature detection
- update documentation to reflect the latest changes
- use new interface in PMacc and PIConGPU

I am aware that the new way requires few more lines of code but it eliminates the requirement of fiddling around with indices to access the distributed context variable.
The old way is still supported because I am not fully sure if I should forbid it to favor the new interface.
I will do many more refactorings of the lockstep usage/interface in the near future where the benefits of the current refactoring will be shown step by step.

### DIFF
--- a/include/picongpu/algorithms/Gamma.hpp
+++ b/include/picongpu/algorithms/Gamma.hpp
@@ -21,8 +21,6 @@
 
 #include "picongpu/simulation_defines.hpp"
 
-#include "picongpu/algorithms/Gamma.def"
-
 
 namespace picongpu
 {
@@ -30,8 +28,6 @@ namespace picongpu
     template<typename T_MomType, typename T_MassType>
     HDINLINE T_PrecisionType Gamma<T_PrecisionType>::operator()(T_MomType const& mom, T_MassType const mass) const
     {
-        using namespace pmacc;
-
         valueType const fMom2 = pmacc::math::abs2(precisionCast<valueType>(mom));
         constexpr valueType c2 = SPEED_OF_LIGHT * SPEED_OF_LIGHT;
 
--- a/include/picongpu/particles/collision/InterCollision.hpp
+++ b/include/picongpu/particles/collision/InterCollision.hpp
@@ -100,7 +100,7 @@ namespace picongpu
                     typename T_Worker,
                     typename T_DeviceHeapHandle,
                     typename T_RngHandle,
-                    typename T_CollisionFunctor,
+                    typename T_SrcCollisionFunctor,
                     typename T_Filter0,
                     typename T_Filter1,
                     typename T_SumCoulombLogBox,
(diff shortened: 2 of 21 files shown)


include/picongpu/particles/collision/InterCollision.hpp
```diff
@@ -201,23 +201,15 @@ namespace picongpu
...
-                    auto collisionFunctorCtx = lockstep::makeVar<decltype(collisionFunctor(
-                        worker,
-                        alpaka::core::declval<DataSpace<simDim> const>(),
-                        alpaka::core::declval<float_X const>(),
-                        alpaka::core::declval<float_X const>(),
-                        alpaka::core::declval<uint32_t const>()))>(forEachFrameElem);
-
-                    forEachFrameElem(
-                        [&](lockstep::Idx const idx)
+                    auto collisionFunctorCtx = forEachFrameElem(
```
psychocoderHPC, line 204: @pordyna Maybe the change in these lines is of interest to you. There is no need anymore to know the signature of the functor before the creation.

include/picongpu/algorithms/Gamma.hpp
```diff
@@ -21,17 +21,13 @@
...
 #include "picongpu/simulation_defines.hpp"
 
-#include "picongpu/algorithms/Gamma.def"
-
 
 namespace picongpu
 {
     template<typename T_PrecisionType>
     template<typename T_MomType, typename T_MassType>
     HDINLINE T_PrecisionType Gamma<T_PrecisionType>::operator()(T_MomType const& mom, T_MassType const mass) const
```
PrometheusPi, line 29: If you removed the `#include "picongpu/algorithms/Gamma.def"`, where do we get the `Gamma` function from?

include/pmacc/lockstep/ForEach.hpp
```diff
@@ -34,6 +34,77 @@ namespace pmacc
...
+            {
+                template<typename T_Functor, typename... T_CtxVars>
+                HDINLINE auto operator()(T_Functor&& functor, Idx idx, T_CtxVars&&... ctxVars) const
+                    -> decltype(functor(idx, std::forward<T_CtxVars>(ctxVars)...))
+                {
+                    return functor(idx, std::forward<T_CtxVars>(ctxVars)...);
+                }
+
+                template<typename T_Functor, typename... T_CtxVars>
+                HDINLINE auto operator()(T_Functor&& functor, Idx idx, T_CtxVars&&... ctxVars) const
```
PrometheusPi, line 53: The interfaces of these `operator()` methods (line 46 and line 53)  are equal. How does PIConGPU/the compiler decide which to use?

PrometheusPi: Offline discussion with @psychocoderHPC: the return type is set via `decltype` and only one of both options can exist - thus the compiler can decide which method to use.

PrometheusPi (approved):
Formally, it looks good to me.

PR #4463, opened 2023-02-17
fix particle position rounding issue

### LINKED ISSUE #3910: Invalid range of particle position after grid shifting
https://github.com/ComputationalRadiationPhysics/picongpu/blob/7bb8e27cdee9f1edf4cdaf972c50150105b60276/include/picongpu/algorithms/ShiftCoordinateSystem.hpp#L59-L63

ShiftCoordinateSystem produces invalid ranges for the particle position if the particle is using an even shape.
If the field position is `0.5` and the particle position is `0.499999999` in line `63` we end up with `-0.00000001f - (-1.0f)` what results in `1.0f` as result. The position must be `[0,1)`.

IMO this is not a big deal as long as we do not use the shifted position to access memory. We use the shifted position there only for shape calculations.

```diff
diff --git a/include/picongpu/algorithms/ShiftCoordinateSystem.hpp b/include/picongpu/algorithms/ShiftCoordinateSystem.hpp
index 447d93f2c..50e9d4a8a 100644
--- a/include/picongpu/algorithms/ShiftCoordinateSystem.hpp
+++ b/include/picongpu/algorithms/ShiftCoordinateSystem.hpp
@@ -61,6 +61,22 @@ namespace picongpu
             intShift[component] = GetOffsetToStaticShapeSystem<isEven>()(v_pos);
             cursor = cursor(intShift);
             pos[component] = v_pos - ValueType(intShift[component]);
+#if 1
+            if(isEven)
+            {
+                if(pos[component] < 0.0_X || pos[component] >= 1.0_X)
+                {
+                    printf("even vpos[%f] pos[%f,%f,%f]\n", v_pos, pos.x(),pos.y(),pos.z());
+                }
+            }
+            else
+            {
+                if(pos[component] < -0.5_X || pos[component] >= 0.5_X)
+                {
+                    printf("even vpos[%f] pos[%f,%f,%f]\n", v_pos, pos.x(),pos.y(),pos.z());
+                }
+            }
+#endif
         }
     };

```

psychocoderHPC:
fix #3910

- Besides fixing #3910 the code for shift coordinate system is simplified. Complicated template constructs get removed by a simple for loop.
- The correction schema in MoveParticles has changed.

The fix is based on the fact that the precision is around 0.0 and is equally distributed, in contrast to the previously used range `[0;1)`.
The new method in `MoveParticles` uses a few more floating point operations and a conditional compare but avoids using floor.

# Tests
- [x] KHI growth rate test TSC
- [x] KHI growth rate test PCS
- [x] KHI growth rate test PQS
- [x] compared register footprint (number of used registers for sm_70 is constant)
- [x] manual compared the PTX code

### DIFF
--- a/include/picongpu/algorithms/FieldToParticleInterpolation.hpp
+++ b/include/picongpu/algorithms/FieldToParticleInterpolation.hpp
@@ -100,7 +100,6 @@ namespace picongpu
              * The following calls seperate the vector interpolation into
              * independent scalar interpolations.
              */
-            using Supports = typename pmacc::math::CT::make_Int<simDim, supp>::type;
             using ResultType = typename T_FieldDataBox::ValueType;
 
             ResultType result;
@@ -110,7 +109,7 @@ namespace picongpu
                 // work on a copy to shift the field for each loop round separate
                 auto shiftedField = field;
                 floatD_X particlePosShifted = particlePos;
-                ShiftCoordinateSystem<Supports>()(shiftedField, particlePosShifted, fieldPos[i]);
+                ShiftCoordinateSystem<supp>()(shiftedField, particlePosShifted, fieldPos[i]);
 
                 auto accessFunctor = [&](DataSpace<simDim> const& idx) constexpr
                 {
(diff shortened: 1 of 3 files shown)


PrometheusPi (approved):
lgtm - great comments

PR #4464, opened 2023-02-17
change class RelayPoint to function

psychocoderHPC:
With c++17 there is no need to write a class specialization depending on a bool template to implement two different behaviors. 
A simple `if constexpr` is doing the same and is more readable.


include/picongpu/fields/currentDeposition/relayPoint.hpp
```diff
@@ -0,0 +1,66 @@
...
+     *                 particle at position x_2
+     * @param x_1 begin position of the particle trajectory
+     * @param x_2 end position of the particle trajectory
+     * @return relay point for particle trajectory
+     */
+    template<bool isEvenSupport>
+    DINLINE float_X relayPoint(int& i_1, int& i_2, const float_X x_1, const float_X x_2)
+    {
+        using namespace pmacc;
+        if constexpr(isEvenSupport)
```
PrometheusPi, line 51: cool - thanks for the explanation in the PR description

PrometheusPi (approved):

PR #4473, opened 2023-02-17
refactor differentiation trait

psychocoderHPC:

### DIFF
--- a/include/picongpu/fields/differentiation/Derivative.hpp
+++ b/include/picongpu/fields/differentiation/Derivative.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2013-2022 Heiko Burau, Rene Widera, Axel Huebl, Sergei Bastrakov
+/* Copyright 2013-2022 Rene Widera, Sergei Bastrakov
  *
  * This file is part of PIConGPU.
  *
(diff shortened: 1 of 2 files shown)


include/picongpu/fields/differentiation/Derivative.hpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2013-2022 Heiko Burau, Rene Widera, Axel Huebl, Sergei Bastrakov
```
psychocoderHPC, line 1: I removed the names because both have definitive not contributed to this files and it is also no copy past of an older version. I assume Sergei copied an old file and remove the full content.

include/picongpu/fields/differentiation/Traits.hpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2013-2022 Heiko Burau, Rene Widera, Axel Huebl, Sergei Bastrakov
```
psychocoderHPC, line 1: Same here. May name can be removed too because I have not contributed to the file.

include/picongpu/fields/differentiation/Derivative.hpp
```diff
@@ -30,51 +30,48 @@
...
+     * @tparam T_direction direction to take derivative in, 0 = x, 1 = y, 2 = z
+     */
+    template<typename T_Derivative, uint32_t T_direction>
+    HDINLINE auto makeDerivativeFunctor()
+    {
+        constexpr bool isSpatialDimension = T_direction < simDim;
+        if constexpr(isSpatialDimension)
+            return DerivativeFunctor<T_Derivative, T_direction>{};
+        else
+            return DerivativeFunctor<Zero, T_direction>{};
```
PrometheusPi, line 75: as comment: this functionality came originally from `include/picongpu/fields/differentiation/Traits.hpp`

PrometheusPi (approved):

PR #4474, opened 2023-02-17
refactor GetChargeState and GetCharge

psychocoderHPC:


BrianMarre (approved):
I would prefer using  `T_IonSpecies` instead of ´T_Particle´ since both functors are only valid for ions but the examples are short enough that it does not matter much

psychocoderHPC:
@BrianMarre I agree, I like to keep the scope of the PR to the constexpr change and tried to rename stuff, only if needed.
Never the less you could push the change as soon as this PR is in.

PrometheusPi (approved):

Issue #4472, opened 2023-02-17
calc_gamma_inv_square not used in radiation plugin


PrometheusPi:
A detailed analysis of the radiation plugin today showed that `calc_gamma_inv_square` is not used.
It should thus be removed.

PrometheusPi:
Same is true for `calc_cos_theta`.

psychocoderHPC:
Yes, let us remove the unsued code.

PrometheusPi:
These methods are used in the “special case” of `1/gamma`being very small, which we needed to add due to floating point imprecision. They are not used by the main code, but only in the case of ultra relativistic particles. Thus, we need to keep these methods.

PR #4484, opened 2023-02-28
remove unused methods in radiation plugin

PrometheusPi:
This pull request removes some not used methods of the radiation plugin to clean up the code base. 
It solves issue #4472.


PrometheusPi:
These methods are used in the “special case” of `1/gamma`being very small, which we needed to add due to floating point imprecision. They are not used by the main code, but only in the case of ultra relativistic particles. Thus, we need to keep these methods.

Issue #4487, opened 2023-03-01
Incident field laser 2D plane wave broken


psychocoderHPC:
It looks like the incident field laser profile at least in 2D is broken if plane wave is used.
When using png's with auto-scale we can see artifacts in the Huygens surface on the borders.
It could be that #4457 is not solving the problem for 2D, or we have not seen the artifacts in #4457 because for the tests we normalized the png's to the incident laser amplitude.

Setup used for the test: #4488

- [ ] check if 3D is affected too

# foil LCT with incident field step 100

![e_png_yx_0 5_000100](https://user-images.githubusercontent.com/4037125/222072113-44e5a510-a617-4d7c-a227-5de58db3e159.png)

# step 1100
![e_png_yx_0 5_001100](https://user-images.githubusercontent.com/4037125/222072148-58d077a0-9525-47b7-81ea-062e461cb341.png)

psychocoderHPC:
From a quick test with the bunch example, it looks like it is a 2D bug only. Most likely equal to what I fixed in https://github.com/ComputationalRadiationPhysics/picongpu/pull/4457 there are in 2D some components that will not be written to the fields on the outer borders.

psychocoderHPC:
The root of the evil is that for `ExpRampWithPrepulse` the trait `MakePeriodicTransversalHuygensSurfaceContiguous` is not defined because I was not aware that we have two types of plane wave lasers profiles.
The second problem s that somehow always `ExpRampWithPrepulse` instead of my plane wave laser is used even if I use the cmake flag case `-t0`,

psychocoderHPC:
Correction `ExpRampWithPrepulse` has a transversal envelope therefore `MakePeriodicTransversalHuygensSurfaceContiguous` should not be defined. The artifacts shown in the image of this image can not be avoided.
The user is still free to define the trait in their param file to reduce the artifact. Kepp in mind that the field in the image above are auto scaled.

psychocoderHPC:
I will close this issue because all is fine, the problem is described in the post before.

PR #4488, opened 2023-03-01
example foilLCT: use incident filed laser

psychocoderHPC:
- remove laser.param
- configure incident field laser equal to the old laser in laser.param

e_png_yx_0 5_001100

**Note:** ~~it looks like there is still a bug in the incident field laser profiles #4487~~ https://github.com/ComputationalRadiationPhysics/picongpu/issues/4487#issuecomment-1450290035


psychocoderHPC:
This PR is defining `MakePeriodicTransversalHuygensSurfaceContiguous` for the laser `ExpRampWithPrepulseBeam` in the param file. With this definition the artefacts will be reduced. Keep in mind the image is scaling the colors to the maximum of the current step.
e_png_yx_0 5_001100

PrometheusPi:
@psychocoderHPC can this be reviewed? @BrianMarre mentioned that this might not yet be ready.

psychocoderHPC:
It is ready. The tiny artifact in row zero and bottom on the left side can not be avoided because of the problem that the ExpRampWithPrepulseBeam is not constant in transversal direction.

PrometheusPi (approved):

PR #4496, opened 2023-03-02
Plugin: CountMacroParticlesPerSupercell and Chargeconservation

psychocoderHPC:
Both plugins can since one of the last refactorings run on CPU and GPU therefore there is no need for guards to avoid the usage of backends other than CUDA.

### DIFF
--- a/include/picongpu/plugins/makroParticleCounter/PerSuperCell.hpp
+++ b/include/picongpu/plugins/makroParticleCounter/PerSuperCell.hpp
@@ -106,7 +106,6 @@ namespace picongpu
         typedef MappingDesc::SuperCellSize SuperCellSize;
         typedef GridBuffer<size_t, simDim> GridBufferType;
 
-        MappingDesc* cellDescription;
         std::string notifyPeriod;
         std::string m_filenameExtension = openPMD::getDefaultExtension();
         std::string m_filenameInfix = "_%06T";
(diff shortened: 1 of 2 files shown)


include/picongpu/plugins/makroParticleCounter/PerSuperCell.hpp
```diff
@@ -106,21 +106,21 @@ namespace picongpu
...
-        MappingDesc* cellDescription;
         std::string notifyPeriod;
         std::string m_filenameExtension = openPMD::getDefaultExtension();
         std::string m_filenameInfix = "_%06T";
 
         std::string pluginName;
         std::string pluginPrefix;
         std::string foldername;
+        MappingDesc* cellDescription;
+
```
PrometheusPi, line 117: why did you change the order @psychocoderHPC

include/picongpu/plugins/makroParticleCounter/PerSuperCell.hpp
```diff
@@ -106,21 +106,21 @@ namespace picongpu
...
         mpi::MPIReduce reduce;
 
         std::unique_ptr<GridBufferType> localResult;
 
         // @todo upon switching to C++17, use std::option instead
         std::unique_ptr<::openPMD::Series> m_Series;
-        // set attributes for datacollector files
```
PrometheusPi, line 123: Why did you remove this comment @psychocoderHPC?

PrometheusPi (approved):
Looks good to me, just a few questions regarding the changes in `PerSuperCell.hpp`

include/picongpu/plugins/makroParticleCounter/PerSuperCell.hpp
PrometheusPi: *offline discussion with @psychocoderHPC:*
The original order caused warnings except for CUDA backends. Since now non-CUDA backends are allowed, changing the order is needed to pass the CI

include/picongpu/plugins/makroParticleCounter/PerSuperCell.hpp
PrometheusPi: *offline discussion with @psychocoderHPC:*
This comment was a long-time-ago left-over. Should have been removed before.

PR #4499, opened 2023-03-03
refactor `math::Vector`, use unsigned dimension

### LINKED ISSUE #395: Vector use signed T_dim template parameter
The class [Vector](https://github.com/ComputationalRadiationPhysics/picongpu/blob/366f29e669893e1b6ffa338224c8142a50223b51/src/libPMacc/include/math/vector/Vector.hpp#L71) uses a `signed int` as signature for T_dim. It should be `uint32_t`!

psychocoderHPC:
fix #395

Dimensions in PMacc/PIConGPU are always unsigned, only math::Vector, math::CT::Vector and math::ConstVector broke this.

- use `uint32_t` for the number of dimensions

- [x] rebase against #4501

### DIFF
--- a/include/pmacc/dimensions/DataSpace.hpp
+++ b/include/pmacc/dimensions/DataSpace.hpp
@@ -28,19 +28,19 @@
 namespace pmacc
 {
     /**
-     * A T_Dim-dimensional data space.
+     * A T_dim-dimensional data space.
      *
-     * DataSpace describes a T_Dim-dimensional data space with a specific size for each dimension.
+     * DataSpace describes a T_dim-dimensional data space with a specific size for each dimension.
      * It only describes the space and does not hold any actual data.
      *
-     * @tparam T_Dim dimension (1-3) of the dataspace
+     * @tparam T_dim dimension (1-3) of the dataspace
      */
-    template<unsigned T_Dim>
-    class DataSpace : public math::Vector<int, T_Dim>
+    template<unsigned T_dim>
+    class DataSpace : public math::Vector<int, T_dim>
     {
     public:
-        static constexpr int Dim = T_Dim;
-        using BaseType = math::Vector<int, T_Dim>;
+        static constexpr uint32_t Dim = T_dim;
+        using BaseType = math::Vector<int, T_dim>;
 
         /**
          * default constructor.
@@ -48,7 +48,7 @@ namespace pmacc
          */
         HDINLINE DataSpace()
         {
-            for(uint32_t i = 0; i < T_Dim; ++i)
+            for(uint32_t i = 0; i < T_dim; ++i)
             {
                 (*this)[i] = 0;
             }
@@ -64,7 +64,7 @@ namespace pmacc
          */
         HDINLINE explicit DataSpace(cupla::dim3 value)
         {
-            for(uint32_t i = 0; i < T_Dim; ++i)
+            for(uint32_t i = 0; i < T_dim; ++i)
             {
                 (*this)[i] = *(&(value.x) + i);
             }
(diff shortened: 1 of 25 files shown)


psychocoderHPC:
- [x] This PR pointed me to an out-of-memory access in the pml implementation. The second commit should fix the issue and must be pushed as separate PR if the CI passed.

psychocoderHPC:
@steindev @PrometheusPi PR is now ready and all dependencies are merged already

include/pmacc/math/ConstVector.hpp
```diff
@@ -87,15 +87,15 @@
         } /* namespace pmacc_static_const_vector_host + id  */                                                        \
         /* select host or device namespace depending on __CUDA_ARCH__ compiler flag*/                                 \
         PMACC_USING_STATIC_CONST_VECTOR_NAMESPACE(id);                                                                \
-        template<typename T_Type, int T_Dim>                                                                          \
+        template<typename T_Type, uint32_t T_Dim>                                                                     \
```
steindev: I am not exactly sure why you started to rename `T_Dim` -> `T_dim` at some places, but I guess it is supposed to be consistent?
You missed to rename here.

(Although I would have expected that `T_Dim` actually complies w/ our standard naming scheme and not `T_dim`)

include/pmacc/math/ConstVector.hpp
```diff
@@ -87,15 +87,15 @@
...
         struct ConstArrayStorage                                                                                      \
         {                                                                                                             \
             PMACC_CASSERT_MSG(                                                                                        \
                 __PMACC_CONST_VECTOR_dimension_needs_to_be_less_than_or_equal_to_the_number_of_arguments__,           \
                 Dim <= count);                                                                                        \
             static constexpr bool isConst = true;                                                                     \
             typedef T_Type type;                                                                                      \
-            static constexpr int dim = T_Dim;                                                                         \
+            static constexpr uint32_t dim = T_Dim;                                                                    \
```
steindev: `T_Dim` -> `T_dim`

include/pmacc/dimensions/DataSpace.hpp
```diff
@@ -39,7 +39,7 @@ namespace pmacc
     class DataSpace : public math::Vector<int, T_Dim>
```
steindev: `T_Dim` -> `T_dim`

include/pmacc/dimensions/DataSpace.hpp
```diff
@@ -39,7 +39,7 @@ namespace pmacc
...
     {
     public:
-        static constexpr int Dim = T_Dim;
+        static constexpr uint32_t Dim = T_Dim;
```
steindev: `T_Dim` -> `T_dim`

include/pmacc/dimensions/DataSpace.hpp
```diff
@@ -39,7 +39,7 @@ namespace pmacc
...
         using BaseType = math::Vector<int, T_Dim>;
```
steindev: `T_Dim` -> `T_dim`

include/pmacc/math/ConstVector.hpp
psychocoderHPC: If the dimension would be a type then the naming `T_Dim` would be fine. In this case, it is a value and should be `T_dim`. I will change it!

include/pmacc/dimensions/DataSpace.hpp
psychocoderHPC: I will change `T_Dim` but will keep `Dim` even if this is also against our naming rules but changing `Dim` will affect more code which is currently not touched and I like to keep this PR as small as possible.

psychocoderHPC:
ready to be merged

steindev (approved):

PR #4502, opened 2023-03-06
remove deprecated laser profiles

psychocoderHPC:
Remove deprecated laser profiles in favor of incident field lasers.
With incident field laser profiles we have we have an equivalent for each old laser profile.

**Note:** This step is required to reduce code duplications and code maintenance overhead.


docs/source/usage/plugins/png.rst
```diff
@@ -100,14 +100,12 @@ Since an adequate color scaling is essential, there several option the user can
...
-   // 1: Laser: typical fields calculated out of the laser amplitude
+   // 1: Laser: [outdated]
    // 2: Drift: [outdated]
    // 3: PlWave: typical fields calculated out of the plasma freq.,
    // assuming the wave moves approx. with c
    // 4: Thermal: [outdated]
-   // 5: BlowOut: typical fields, assuming that a LWFA in the blowout
-   // regime causes a bubble with radius of approx. the laser's
-   // beam waist (use for bubble fields)
+   // 5: BlowOut: [outdated]
```
PrometheusPi, line 108: why is blowout also outdated? 
Couldn't we derive its assumed value from incident field ?

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
```diff
@@ -188,10 +188,6 @@ namespace picongpu
                     private:
                         HDINLINE float_X getEnvelope(float_X const runTime) const
                         {
-                            /* workaround for clang 5 linker issues
-                             * `undefined reference to
-                             * `picongpu::fields::laserProfiles::ExpRampWithPrepulseParam::INT_RATIO_POINT_1'`
-                             */
```
PrometheusPi, line 194: only the comment was removed but not the actual workaround - @psychocoderHPC could you comment on that?

share/picongpu/examples/WeibelTransverse/include/picongpu/param/png.param
```diff
@@ -37,7 +37,7 @@ namespace picongpu
     {
         // normalize EM fields to typical laser or plasma quantities
         //-1: Auto:     enable adaptive scaling for each output
-        // 1: Laser:    typical fields calculated out of the laser amplitude
+        // 1: Laser:    [outdated]
         // 2: Drift:    [outdated]
         // 3: PlWave:   typical fields calculated out of the plasma freq.,
         //              assuming the wave moves approx. with c
```
PrometheusPi, line 43: why is blowout below not set to `[outdated]`?

PrometheusPi (changes requested):
- one change is still required (the missing blowout comment)
- there are a few questions
- why are no `laser.param` files in examples removed?

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
psychocoderHPC: The implementation is OK, we are required in someplace to store the constexpr in an intermediate variable before. I assume that clan5 in the past had a name lookup issue and used always the wrong variable.

docs/source/usage/plugins/png.rst
psychocoderHPC: Blow out depends on the laser profiles too, there is no incident field equivalent available. 
IMO this is no problem, no example was used it so the code was not tested anyway and png's are only helpers and real visualisation should be done via other tools.

share/picongpu/examples/WeibelTransverse/include/picongpu/param/png.param
psychocoderHPC: It is missing here and therefore not found by search and replace

psychocoderHPC: will do

psychocoderHPC: done

psychocoderHPC:
@PrometheusPi ready to be merged

docs/source/usage/plugins/png.rst
PrometheusPi: Okay - if no one used it, that should be fine. Otherwise it just depends on a_0, which we set in both the old laser and the new incident field.

PrometheusPi (approved):
Just again my previous question: why was no `laser.param` removed?

psychocoderHPC:
Ohh yes you are right the main laser.param must be removed too. 

Am 7. März 2023 14:45:27 MEZ schrieb Richard Pausch ***@***.***>:

include/picongpu/param/laser.param
```diff
@@ -1,728 +0,0 @@
-/* Copyright 2013-2022 Axel Huebl, Anton Helm, Rene Widera, Richard Pausch,
- *                     Alexander Debus, Nils Prinz, Klaus Steiniger
- *
- * This file is part of PIConGPU.
- *
```
psychocoderHPC, line 5: @PrometheusPi The removed laser.param is already in the PR, I thought I forgot it but it is here.
All other laser.param were removed before in the rPRs where we switched to incident field lasers.

PR #4504, opened 2023-03-06
remove `boost::filesystem usage`

psychocoderHPC:
Switch from `boost::filesystem` to `std::filesystem` and reduc ethe boost dependencies.

The second commit is dropping clang tests on ubuntu 18.04 do to libstdc++ incompatibilities between OS shipped g++7 and clang if we compile for C++17.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -71,8 +71,8 @@ MPI 2.3+
 
 Boost
 """""
-- 1.74.0+ (``program_options``, ``filesystem``, ``system``, ``serialization`` and header-only libs)
-- *Debian/Ubuntu:* ``sudo apt-get install libboost-program-options-dev libboost-filesystem-dev libboost-system-dev libboost-serialization-dev``
+- 1.74.0+ (``program_options``, ``atomic`` and header-only libs)
+- *Debian/Ubuntu:* ``sudo apt-get install libboost-program-options-dev libboost-atomic-dev``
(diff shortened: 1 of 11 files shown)


steindev:
@psychocoderHPC Please note, not only the CI failed, but there are also conflicts to the dev branch. (Maybe because I just merged the "remove boost::math" PR?)

psychocoderHPC:
> @psychocoderHPC Please note, not only the CI failed, but there are also conflicts to the dev branch. (Maybe because I just merged the "remove boost::math" PR?)

There are incompatibilities between gnu headers which are too old and C++17. I tried to solve it by using `libc++` but this creates linker issues because of ABI incompatibilities. 
Maybe I will drop ubuntu 18.04 in the CI, it's old anyway and at least with the CI container 3.1 we need to drop it.

psychocoderHPC:
@steindev PR can be merged, I will not provide the second commit as a separate PR. This will only postpone this PR with very little advantage that we have a PR. We will switch to new CI containers soon and then ubuntu 18.04 will be fully dropped anyways. Is it ok for you?

steindev (approved):

steindev:
FYI, I checked the appearances of `boost::filesystem` w/
```bash
$ grep -rnw include/ -e ".*boost::file.*"
$ #and
$ grep -rnw include/ -e ".*bfs::.*"
```

PR #4394, opened 2022-12-14
use incident field laser for LWFA example

psychocoderHPC:
- change number of cells in cfg's for X and Z direction to 192 that the laser is always centered
- add incident field laser with same properties than the old laser
  - note: the laser is shifted by 16 cells because the of the Huygens surface placement in `incidentFiled.parm`
  - Incident field laser have a absolut focus in all direction, therefore the laser is not centered if the number of cells for the setup is changed.
  - fix incident field base class initialization order

**Note**  The second commit contains a fix for the base class of incident fields, it is currently not possible to separate this commit into its own PR because of #4390.

Issue #4410, opened 2023-01-06
Possible output data bug


kimjohn1:
e_all_chargeDensity data provided by the PIConGPU dev branch running the Laser Wakefield example simulation and using the openPMD plugin on 8 December 2022  produced the image below

![e_all_chargeDensity scalar field with 85b04ff76 branch](https://user-images.githubusercontent.com/61957753/210902925-09a8675a-33e7-4a14-8c69-d13ee7f839ad.png)

 
e_all_chargeDensity data provided by the PIConGPU dev branch running the Laser Wakefield example simulation and using the openPMD plugin on 31 December 2022 produced the image below

![e_all_chargeDensity with dev branch](https://user-images.githubusercontent.com/61957753/210903152-28d2b6ba-3632-4f4e-8719-9161f6a0ef92.png)

The first image has been the generated result observed during testing with the PIConGPU dev branch running the Laser Wakefield example simulation frequently since June of 2022.  Testing since June has been conducted using a custom config file that sends data via openPMD SST to the visualization software.  There has been a lot of testing done during June to December, though an installation of PIConGPU that downloaded the latest dev branch was only occasionally needed.  Testing on 8 December included rebuilding the software stack on one of the lab computers, so we know that PIConGPU was installed that day.  Installation is done using a script, and the script indicates switching to dev branch, so there is pretty high confidence that the PIConGPU dev branch was installed and then used from 8 December to 31 December.  The test on 31 December included re-installing the software stack, and was to verify that PIConGPU issue 4382 was fixed.  PIConGPU dev branch was installed that day, and successfully ran so issue 4382 was verified fixed, but the output from the simulation produced the second image above.

There have been no changes to the openPMD Reader that is installed in the visualization software application since late June.

Through a bit of trial and error I have found that PIConGPU branch 85b04ff76, runs and produces the correct visual, although it precedes the issue 4382 fix, so I am using the workaround in the config file as suggested in the issue discussion.  Images that are attached to this post were produced today (5 Jan) using the current dev branch and the 85b04ff76 branch.  For now, I am able to continue with my project development, but I am exclusively using the 85b04ff76 branch (and issue 4382 workaround).

Kim

[Configs.tar.gz](https://github.com/ComputationalRadiationPhysics/picongpu/files/10356481/Configs.tar.gz)

[Images.tar.gz](https://github.com/ComputationalRadiationPhysics/picongpu/files/10356328/Images.tar.gz)

[...]

PR #4507, opened 2023-03-09
incident laser focus origin

### LINKED ISSUE #4414: unexpected change in LWA 1.cfg and 1_isaac.cfg
During investigation of issue #4410 I noted that the TBG_gridSize specified in 1.cfg and 1_isaac.cfg for the LaserWakefield example simulation is 192 1024 192 in the dev branch and 192 1024 12 in the Master branch.

That seemed like a pretty hefty change in the dev branch for a 1-GPU processed simulation.  Is that a typo or an intentional change?

Kim

psychocoderHPC:
With #4394 where we switched from the old laser profiles to incident field lasers it was required to define the focus position absolute to the origin of the total domain. Because of this absolute focus, it was not possible to in/decrease the simulation volume without adjusting the focus position if the laser should be centered in the domain. The drawback was that the simulation domain was adjustable at runtime and the focus change required recompiling PIConGPU.
For some workflows e.g. visualization the change was a disadvantage.
This PR is solving the issue by introducing a way to bring back the old behavior if needed.

- Add the optional variable `FOCUS_ORIGIN_*` to define the origin of the focus.
- Laser Wakefield and foil LCT example set focus origin of transversal laser direction to the center of the direction.
- update documentation strings

Note: This PR is backward compatible with the latest dev branch param files because the variables `FOCUS_ORIGIN_*` are optional.

### DIFF
--- a/include/picongpu/fields/incidentField/Functors.hpp
+++ b/include/picongpu/fields/incidentField/Functors.hpp
@@ -29,6 +29,7 @@
 #include <limits>
 #include <stdexcept>
 #include <string>
+#include <type_traits>
 
 
 namespace picongpu
@@ -112,6 +113,49 @@ namespace picongpu
             };
             namespace detail
             {
+                /** SFINAE dedection if the user parameter define the variable FOCUS_ORIGIN_*
+                 *
+                 * This allows that focus origin can be an optional variable a user must only define if needed.
+                 * The default if it is not defined is Origin::Zero
+                 * @{
+                 */
+                template<typename T, typename = void>
+                struct GetOriginX
+                {
+                    static constexpr Origin value = Origin::Zero;
+                };
+
+                template<typename T>
+                struct GetOriginX<T, decltype((void) T::FOCUS_ORIGIN_X, void())>
+                {
+                    static constexpr Origin value = T::FOCUS_ORIGIN_X;
+                };
+
+                template<typename T, typename = void>
+                struct GetOriginY
+                {
+                    static constexpr Origin value = Origin::Zero;
+                };
+
+                template<typename T>
+                struct GetOriginY<T, decltype((void) T::FOCUS_ORIGIN_Y, void())>
+                {
+                    static constexpr Origin value = T::FOCUS_ORIGIN_Y;
+                };
+
+                template<typename T, typename = void>
+                struct GetOriginZ
+                {
+                    static constexpr Origin value = Origin::Zero;
+                };
+
+                template<typename T>
+                struct GetOriginZ<T, decltype((void) T::FOCUS_ORIGIN_Z, void())>
+                {
+                    static constexpr Origin value = T::FOCUS_ORIGIN_Z;
+                };
+                /**@}*/
+
                 /** Base class for calculating incident E functors
                  *
                  * Defines internal coordinate system tied to laser focus position and direction.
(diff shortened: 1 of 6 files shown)


psychocoderHPC:
@kimjohn1 it is maybe interesting to you that I implemented a way to bring back the old laser behavior and the LWFA example can no be used with 12 cells in Z direction without adjusting the param files.

psychocoderHPC:
> @kimjohn1 it is maybe interesting to you that I implemented a way to bring back the old laser behavior and the LWFA example can no be used with 12 cells in Z direction without adjusting the param files.

@kimjohn1 SRy I notified you to early, I realized that the laser position of the planes also avoid setting the domain size to 12 cells for LWFA like in the past. I need to create a second PR to provide a workaround for this behavior to bring back the old look and feel but we are one step further.


Currently 

```C++
            namespace profiles
            {
                /* This setup is using by default periodic boundaries in X direction.
                 * Defining this trait will stretch the Huygens surface to create lesser artefacts in the first cell on
                 * the outer boundaries. Note: For ExpRampWithPrepulseBeam the fields in the transversal direction are
                 * not constant therefore  the artefacts can not be avoided.
                 */
                template<>
                struct MakePeriodicTransversalHuygensSurfaceContiguous<profiles::GaussianBeam<GaussianBeamParam>>
                {
                    static constexpr bool value = true;
                };
            } // namespace profiles
```

Must be added [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/2b82ffcf5121a53ca33ddb82fd5a03ef5ee05069/share/picongpu/examples/LaserWakefield/include/picongpu/param/incidentField.param#L235) to allow setting the Z-direction to 12 cells.

kimjohn1:
@psychocoderHPC 
I attempted your recommended workaround from issue #4410 where I set 192 1024 12 for GridSize and set a new z axis position in the incidentField.param file, line 164 to 1.0632e-06, but noticed that I get a warning during simulation setup that says
> PIConGPU Verbose PHYSICS (1) | Warning: volume bounded by the Huygens surface is zero, no incident field will be generated

That was enough to tell me that I wasn't going to get a visualization.  I found that if I set the GridSize to 192 1024 48, and the z axis position value to 4.253e-6, that works and produces a visualization.

I also continued the investigation into why my system was crashing with the unchanged GridSize of 192 1024 192 and unchanged incidentField.param file, and I found a weakness in the visualization side of the software.  I improved that code, and the system runs the LaserWakefield example just fine. So, the experience with a new version of LaserWakefield was quite beneficial because it revealed this previously unnoticed software problem.

I appreciate and support your effort to be able to run the LWFA simulation with a GridSize of 192 1024 12.  The smaller grid size allows the simulation to run a little faster, producing a slightly lower fidelity visual, and thus demonstrates the 'facts of life' with running Particle In Cell simulations.

I will definitely give your new implementation a try when you get it worked out.
Kim

PrometheusPi (approved):
looks good to me
what is with the other `incidentField.param` in `share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param`?

psychocoderHPC:
> looks good to me what is with the other `incidentField.param` in `share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param`?

The file you pointed to does is not using any laser profile. In general the new parameter is optional and there is no requirement to add it.

PR #4515, opened 2023-03-23
refactor `pic-create`

psychocoderHPC:
Update our tooling to reflect common workflows. 
The python tools will be used mostly from PIConGPU source code and not the input set because we set the python path to PIConGPU source lib folder.
With a new environment variable `PIC_SYSTEM_TEMPLATE_PATH` we can control that only the required profiles and templates for tbg will be compiled by `pic-create`. If the environment variable does not exist we will fall back to the old behavior and copy all system templates/profiles into the input set.

- do not copy picongpu/lib into the input set
- add `PIC_SYSTEM_TEMPLATE_PATH` to control which tbg  templates/profiles will be copied
- update system profiles to expose which system template folder should be used

### DIFF
--- a/bin/pic-create
+++ b/bin/pic-create
@@ -23,11 +23,22 @@ this_dir=$(cd $(dirname $0) && pwd)
 
 # PIConGPU prefix path
 picongpu_prefix=$(cd $this_dir/.. && pwd)
-# files always cloned from defaults, can be overwritten by input set
-default_folder_to_copy="etc/picongpu lib"
-# if we clone an input set we copy from it
+# All files in the folder (no recusion or subfolders) will be copied from PIConGPU source to the input set.
+default_folder_to_copy="etc/picongpu"
+
+# Folder which get recusivly copied from the source input set into the destination input set
 folder_to_clone="etc/picongpu bin include/picongpu lib"
 files_to_copy="cmakeFlags executeOnClone"
+
+if [ -n "$PIC_SYSTEM_TEMPLATE_PATH" ] ; then
+    default_folder_to_copy="$default_folder_to_copy $PIC_SYSTEM_TEMPLATE_PATH"
+else
+    # Fallback to be backward compatible to the old behaviour that all content of picongpu/etc is copied into the
+    # destination input set.
+    # Fallback is only aktive if the environment variable 'PIC_SYSTEM_TEMPLATE_PATH' not exists.
+    # All subfolder will be copied.
+    full_folder_to_copy="etc/picongpu"
+fi
 
 help()
 {
@@ -130,7 +141,7 @@ fi
 
 echo "Cloning from $src_path to $dst_path"
 
-#first copy default parameter
+# first copy from PIConGPU source into the destination input set
 for d in $default_folder_to_copy
 do
     dir_dst="$dst_path/$d"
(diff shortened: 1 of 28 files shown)


psychocoderHPC:
@ComputationalRadiationPhysics/picongpu-developers After this PR is merged you should add the environment variable `PIC_SYSTEM_TEMPLATE_PATH` to your profiles to avoid that `pic-create` and each `tbg` call is coping to many files to each input set.
**Note**: If you clone an already existing input set with `pic-create` the number of files will not be reduced. This change is only effecting input set newly cloned from the source code.

bin/pic-create
```diff
@@ -154,14 +181,11 @@ if [ "$src_path" != "$picongpu_prefix" ] ; then
     done
 fi
 
-#copy files
+# copy files from root of the source input set to the destination input set
 for d in $files_to_copy
 do
     file_dst="$dst_path/$d"
     file_src="$src_path/$d"
-    if [ ! -d "$dir_dst" ] ; then
```
psychocoderHPC, line 162: This is a copy past from the loop before and is wrong and therefore removed.

etc/picongpu/bash-devServer-hzdr/cpu_picongpu.profile.example
```diff
@@ -57,7 +61,7 @@ export PYTHONPATH=$PICSRC/lib/python:$PYTHONPATH
 
 # "tbg" default options #######################################################
 export TBG_SUBMIT="bash"
-export TBG_TPLFILE="etc/picongpu/bash/mpiexec.tpl"
+export TBG_TPLFILE="etc/picongpu/bash-devServer-hzdr/mpiexec.tpl"
```
psychocoderHPC, line 64: fixed that the wrong template was used.

bin/pic-create
```diff
@@ -23,12 +23,23 @@ this_dir=$(cd $(dirname $0) && pwd)
...
 files_to_copy="cmakeFlags executeOnClone"
 
+if [ -n "$PIC_SYSTEM_TEMPLATE_PATH" ] ; then
+    default_folder_to_copy="$default_folder_to_copy $PIC_SYSTEM_TEMPLATE_PATH"
+else
+    # Fallback to be backward compatible to the old behaviour that all content of picongpu/etc is copied into the
+    # destination input set.
+    # Fallback is only aktive if the environment variable 'PIC_SYSTEM_TEMPLATE_PATH' not exists.
+    # All subfolder will be copied.
+    full_folder_to_copy="etc/picongpu"
```
PrometheusPi, line 40: This is already set in line 27 - why do you repeat this here?

bin/pic-create
```diff
@@ -130,18 +141,34 @@ fi
...
     fi
-    rsync --inplace -r -q -avc --exclude=".*" $dir_src/ $dir_dst
+    rsync --inplace -q -avc --exclude=".*" --exclude="*/" $dir_src/ $dir_dst
+done
+
+# copy subfolder of PIConGPU source etc/picongpu into the destination input set
+for copy_folder in $full_folder_to_copy
+do
+    dir_to_copy="$picongpu_prefix/$copy_folder"
+    for d in $(ls -w 1 -d "$dir_to_copy/"*/)
```
PrometheusPi, line 159: for looping `-w 1` is not needed afaIk

PrometheusPi (changes requested):
One relevant question and just one minor (optional) change

bin/pic-create
psychocoderHPC: We handle `full_folder_to_copy` later in a separate loop to copy all subfolders.

PrometheusPi: You are right - misread the two variable names

bin/pic-create
PrometheusPi: since this should work equally well - I will merge

PrometheusPi (approved):

PR #4516, opened 2023-03-23
Replace mpl by mp11 (part 1.5/2)

bernhardmgruber:
This PR refactores the MPL sequence algorithms to use mp11.

Depends on:

- [x] #3834


psychocoderHPC:
I was forced to push the rebased version to your branch to get a better diff, I know that only the lat commit contains the newest changes but would like to be sure not to miss something.

psychocoderHPC (approved):

PR #4517, opened 2023-03-23
CI: fix compile issue if `cpuonly` job is executed on a GPU runner

psychocoderHPC:
In cases where the compile only job job is executed on an GPU runner but with different kind of accelerators we need to reset the variables `CI_GPUS` and `CI_GPU_ARCH` to avoid compiling for the wrong architecture and accelerator.

A list of environment variables available in the gitlab Ci: https://docs.gitlab.com/ee/ci/variables/predefined_variables.html

### DIFF
--- a/share/ci/run_picongpu_tests.sh
+++ b/share/ci/run_picongpu_tests.sh
@@ -10,6 +10,13 @@ if [[ ! -v PIC_BUILD_TYPE ]] ; then
     PIC_BUILD_TYPE=Release ;
 fi
 
+if [[ "$CI_RUNNER_TAGS" =~ .*cpuonly.* ]] ; then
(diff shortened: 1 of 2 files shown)


PrometheusPi (approved):

PrometheusPi:
@psychocoderHPC neither `CI_GPUS` nor `CI_GPU_ARCH` exists in the reference you link you posted

psychocoderHPC:
> @psychocoderHPC neither `CI_GPUS` nor `CI_GPU_ARCH` exists in the reference you link you posted

These variables are HZDR runner-specific variables. IN the link you ca find the definition of `CI_RUNNER_TAGS` to handle the special case that we run cpuonly jobs on an GPU CI node.

PR #4519, opened 2023-03-27
fix stdlibc++ filesystem compatibility

psychocoderHPC:
On a system with an old stdlibc++ version <= 7 `std::filesystem` is
experimental. Handle this case in PMacc and use the correct includes and
namespace if needed.

- change `#include <filesystem>` to `#inlcude <pmacc/filesystem.hpp>`
- CMake provides target `pmacc::filesystem`
- CI: Re-add the tests to check the workaround for c++17 filesystem includes

The filesystem issue was introduced with #4504 and affects crusher@ORNL.


PrometheusPi (approved):

franzpoeschel:
I verified that with these changes, PIConGPU successfully compiles and links again on Crusher.

PR #4520, opened 2023-03-27
Added script to edit copied PIConGPU checkpoints.

nicowrobel:
The python script can be used to copy checkpoints and add particles to the copy. 
The added Jupyter Notebook shows, how this could be used to add a driver into an empty checkpoint.
A lot of code from the [pipe script](https://github.com/openPMD/openPMD-api/blob/dev/src/binding/python/openpmd_api/pipe/__main__.py) from the openPMD-api was used, therfore Franz Poeschel is credited.
Tests for the current dev and, if needed, hdf5 checkpoints are pending.

Supported data for particle species inputs are arrays with the position, momentum and weighting of the particles.
probeE, probeB and id data is supported for the new particles.
Currently only overwriting checkpoint 0 is supported and only if there aren't already particles of the given species.

### DIFF
--- a/lib/python/picongpu/extra/input/createBunch_example.ipynb
+++ b/lib/python/picongpu/extra/input/createBunch_example.ipynb
@@ -13,10 +13,12 @@
    "source": [
     "## Intro\n",
     "\n",
-    "This notebook allows you to define a PWFA driver bunch using numpy.\n",
+    "This notebook allows you to define a simple PWFA driver bunch using numpy.\n",
     "Each step will be explained on the way.\n",
-    "In the end, you will add your particles (of the driver bunch) to an empty **openPMD-api** checkpoint in **HDF5** format.\n",
-    "Since rewriting the entire code would have taken too long (before you start Nico), you have to stick to **HDF5** for now. "
+    "In the end, the particles (of the driver bunch) will be added to an empty **openPMD-api** checkpoint in **bp** format.\n",
+    "This serves as an example on how to use the checkpoint edit tool.\n",
+    "\n",
+    "**!!! The copying of checkpoints is not optimised and takes up a lot of memory (Roughly the size of the empty checkpoint itself). At least 8 CPUs with 20 GB memory each should be used to run this notebook !!!**"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


lib/python/picongpu/extra/input/createBunch_example.ipynb
```diff
@@ -0,0 +1,294 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Intro\n",
+    "\n",
+    "This notebook allows you to define a PWFA driver bunch using numpy.\n",
+    "Each step will be explained on the way.\n",
+    "In the end, you will add your particles (of the driver bunch) to an empty **openPMD-api** checkpoint in **HDF5** format.\n",
+    "Since rewriting the entire code would have taken too long (before you start Nico), you have to stick to **HDF5** for now. "
```
PrometheusPi: "before you start Nico" 😉

lib/python/picongpu/extra/input/createBunch_example.ipynb
```diff
@@ -0,0 +1,294 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "## convert data and write to checkpoint"
+   ]
+  },
```
PrometheusPi: This looks like a heading to a removed piece of code, is this right?

lib/python/picongpu/extra/input/createBunch_example.ipynb
```diff
@@ -0,0 +1,294 @@
...
+    "mom = vec3D(px/mean_weight, py/mean_weight, pz/mean_weight)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# assigne to (existing) checkpoint file (needs to be hdf5 file)\n",
```
PrometheusPi: "needs to be hdf5 file" - no longer true

PrometheusPi (dismissed):
looks good to me - just some minor questions/comments, perhaps we can clarify this before merging

PrometheusPi (approved):

PR #4531, opened 2023-04-17
ORNL frontier template and profile

psychocoderHPC:
- add slurm configuration section
- add required modules

# Review
Please check the second commit to see the difference between frontier and crusher.

### DIFF
--- a/etc/picongpu/frontier-ornl/batch.tpl
+++ b/etc/picongpu/frontier-ornl/batch.tpl
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Copyright 2013-2022 Axel Huebl, Richard Pausch, Rene Widera, Sergei Bastrakov, Klaus Steinger
+# Copyright 2013-2023 Axel Huebl, Richard Pausch, Rene Widera, Sergei Bastrakov, Klaus Steinger
 #
 # This file is part of PIConGPU.
 #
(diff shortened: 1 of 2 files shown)


finnolec (commented):
When I only use this profile and a self-built pngwriter package, I get the following warning:
```
CMake Warning at CMakeLists.txt:390 (find_package):
  Found package configuration file:

    $PIC_LIBS/pngwriter-0.7.0/lib/cmake/PNGwriter/PNGwriterConfig.cmake

  but it set PNGwriter_FOUND to FALSE so package "PNGwriter" is considered to
  be NOT FOUND.  Reason given by package:

  PNGwriter could not be found because dependency Freetype could not be
  found.
```
But you can add `module load freetype/2.11.1` somewhere, which would fix the problem :) Otherwise the profile seems to work

psychocoderHPC:
> When I only use this profile and a self-built pngwriter package, I get the following warning:
> 
> ```
> CMake Warning at CMakeLists.txt:390 (find_package):
>   Found package configuration file:
> 
>     $PIC_LIBS/pngwriter-0.7.0/lib/cmake/PNGwriter/PNGwriterConfig.cmake
> 
>   but it set PNGwriter_FOUND to FALSE so package "PNGwriter" is considered to
>   be NOT FOUND.  Reason given by package:
> 
>   PNGwriter could not be found because dependency Freetype could not be
>   found.
> ```
> 
> But you can add `module load freetype/2.11.1` somewhere, which would fix the problem :) Otherwise the profile seems to work

The error shows up because you compiled with freetype and I have not compiled pngwriter ^^. You can disable freetype during the compilation of pngwriter. Never the less I will add the module to the profile,

psychocoderHPC:
`module load freetype/2.11.1` is added

steindev (approved):

Issue #4533, opened 2023-04-18
Dispersive laser z-polarization (in 2D) missing 


PrometheusPi:
The recently added dispersive laser implementation https://github.com/ComputationalRadiationPhysics/picongpu/pull/4302 is defect. 

Setting the dispersice laser to 
- $\tau_\mathrm{FWHM} = 30 \mathrm{fs}$, 
- $w_0 = 2.5 \mathrm{\mu m}$, 
- $\lambda_0 = 800 \, \mathrm{nm}$, 
- $a_0 = 50$ and 
- setting `PULSE_INIT = 10.0` 
- with polarization in `z` direction
- and keeping all AD, GDD and TOD values zero

in a 2D simulation should resemble the same as a Gaussian pulse implementation using the same parameters. 

However, the dispersive laser is much wider. 

Dispersive laser:
![grafik](https://user-images.githubusercontent.com/5121158/232759937-e8ee0fe6-c0fd-4200-b6ac-99bf963ab801.png)

Gauss laser:
![grafik](https://user-images.githubusercontent.com/5121158/232759785-cf042ad0-bc65-435b-b0b2-984b7e9eb403.png)

cc'ing @steindev and @fabidie

PrometheusPi:
Furthermore, the dispersive laser requires 45 minutes of run time for 40k steps, while the Gaussian pulse only requires 2 minutes. 
There seems to be a **significant slowdown** during initialization!

Approximately 147ms are needed per time step if the dispersive laser is initialized with some intensity vs. 1ms for the Gaussian pulse.

PrometheusPi:
Additionally, the field energy in the simulation explodes with the dispersive laser (but this might be either a boundary issue due to the large/infinite spot size - e.g. #4532) or an implementation error: 
![grafik](https://user-images.githubusercontent.com/5121158/232763248-f879ca98-f9c1-4928-b808-73bea04b8c1a.png)

fabidie:
I will try to take a look at this the next week. 
Regarding the init time: in contrary to the Gaussian Beam (where the incident field is calcualted analytically) the Dispersive Laser uses a DFT, which is why it takes longer. Nevertheless, the time difference seems to be pretty huge - perhaps there is a way to accelerate the implemented DFT.

PrometheusPi:
Thanks @fabidie For now, the slow time is not an issue for me. I am more concerned with the spot size issue.

PrometheusPi:
[...]

PR #4538, opened 2023-04-19
Fix dispersive laser z polarization

PrometheusPi:
This pull request fixes the bug discovered in #4533 

x-pol case: 
grafik

z-pol old:
grafik

z-pol fixed:
grafik


PrometheusPi:
wrong target branch

PR #4539, opened 2023-04-19
fix dispersive laser z-polarization in 2D

PrometheusPi:
This pull request fixes the bug #4533 

`dev` x - polarization $E_x$: 
grafik

`pull request` z - polarization (fixed) $E_z$: 
grafik

`dev` z - polarization (wrong) $E_z$: 
grafik

There seems to be still no curvature in the "fixed" laser - and thus no correct focusing.


PrometheusPi:
sorry - not from my branch

PR #4456, opened 2023-02-08
Radiation plugin: Save amplitudes additionally per rank

franzpoeschel:
Unlike the aggregated amplitudes, this uses std::complex to store the output, making things a bit inconsistent. The goal should be to use std::complex for the aggregated output too, but Richard says that some postprocessing still relies on the current format. (In the same instant, we should remove the useless third dimension there)

Output now looks like (with two ranks):
```
$ bpls radiationOpenPMD/e_radAmplitudes200.bp/   
  double          /data/200/DetectorMesh/Amplitude/x_Im           {128, 1024, 1}
  double          /data/200/DetectorMesh/Amplitude/x_Re           {128, 1024, 1}
  double          /data/200/DetectorMesh/Amplitude/y_Im           {128, 1024, 1}
  double          /data/200/DetectorMesh/Amplitude/y_Re           {128, 1024, 1}
  double          /data/200/DetectorMesh/Amplitude/z_Im           {128, 1024, 1}
  double          /data/200/DetectorMesh/Amplitude/z_Re           {128, 1024, 1}
  double          /data/200/DetectorMesh/DetectorDirection/x      {128, 1, 1}
  double          /data/200/DetectorMesh/DetectorDirection/y      {128, 1, 1}
  double          /data/200/DetectorMesh/DetectorDirection/z      {128, 1, 1}
  double          /data/200/DetectorMesh/DetectorFrequency/omega  {1, 1024, 1}
  double complex  /data/200/DetectorMesh/amplitude_distributed/x  {2, 128, 1024}
  double complex  /data/200/DetectorMesh/amplitude_distributed/y  {2, 128, 1024}
  double complex  /data/200/DetectorMesh/amplitude_distributed/z  {2, 128, 1024}
```

TODO
- [x] Testing, documentation
- [x] mesh naming?
- [x] Command line option to switch this on off

PR #4546, opened 2023-04-27
Documentation: link Romio HDF5 chunking issue

franzpoeschel:
In the openPMD plugin, we specify `{"hdf5":{"dataset":{"chunking":"none"}}}` by default. This seems to also affect #4456.
It looks like this is caused by https://github.com/open-mpi/ompi/issues/7795 (likely to be fixed with Openmpi 5). Put this in the documentation, so we find it again.

### DIFF
--- a/include/picongpu/plugins/openPMD/Json.cpp
+++ b/include/picongpu/plugins/openPMD/Json.cpp
@@ -296,7 +296,11 @@ The key 'select' must point to either a single string or an array of strings.)EN
 
     void addDefaults(nlohmann::json& config)
     {
-        // disable HDF5 chunking as it can conflict with MPI-IO backends
+        /*
+         * Disable HDF5 chunking as it can conflict with MPI-IO backends.
+         * This is very likely the same issue as
+         * https://github.com/open-mpi/ompi/issues/7795.
+         */
         {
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
Should we remove `export OMPI_MCA_io=^ompio` from our templates? We added it in the past because in HDF5 `ompio` was broken.

franzpoeschel:
Considering that it's ROMIO which is now broken, we should maybe switch sides, yeah.
Are you aware if the issues with OMPIO have been fixed in more recent versions?

psychocoderHPC:
> Are you aware if the issues with OMPIO have been fixed in more recent versions?

No, I do not know if this was fixed.

psychocoderHPC (approved):

psychocoderHPC:
OK found that @ax3l documented where the bug was fixed https://github.com/open-mpi/ompi/issues/6285#issuecomment-474392527
IMO we should remove the workaround in our templates.

psychocoderHPC:
I remove `ompio` in all our templates #4547

franzpoeschel:
Nice!
I think that this probably explains the various HDF5 issues that we have been recently experiencing.

PR #4549, opened 2023-05-03
update openPMD-api module to 14.4

PrometheusPi:
This pull request updates the suggested openPMD-api module on ORNL summit to version 14.4 since the previous module was outdated and not supported by PIConGPU anymore.


steindev:
Does compiling and output actually work? We usually had problems w/ incompatibilities between openPMD-api and available adios2 and hdf5 libs.

Unfortunately, the dependencies between modules are not named in their descriptions @OLCF.

PrometheusPi:
I have tested compiling, but not running. Will do this now.

steindev (approved):

steindev:
According to @PrometheusPi, runtime test succeeded as well. Will merge.

PR #4550, opened 2023-05-04
add picongpu call validation to tbg

pordyna:
`PIConGPU` can be run with the `--validate`to validate run time options instead of running the simulation. Until now, there was no easy way to use this feature together with `tbg` and `.cfg` files, and so almost no one was using it.

This feature adds the validation to `tbg`. With this commit, the `.tpl` files for `bash` and `hemera` will generate a `submit.start` that can be called with an optional  `--validate`  flag. In that case, `picongpu --validate` is run with the same parameters as the actual simulation, but without the `mpiexec` / `srun` call around it. `tbg` won't submit the job if the validation fails. 
 
With this, we should be able to avoid situations where we wait for days for a job to start just to find out that there was a typo in the `.cfg` file.

If you like this implementation, we can copy and paste it into the other templates.

### DIFF
--- a/bin/tbg
+++ b/bin/tbg
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Copyright 2013-2022 Axel Huebl, Rene Widera, Richard Pausch
+# Copyright 2013-2023 Axel Huebl, Rene Widera, Richard Pausch, Pawel Ordyna
 #
 # This file is part of PIConGPU.
 #
@@ -497,8 +497,13 @@ cp -a "$TBG_cfgFile" tbg/submit.cfg
 #warn, if there are still unresolved !TBG_ variables left
 check_final tbg/submit.start "$tooltpl_file_data"
 
-if [ -n "$submit_command" ] ; then
-    $submit_command tbg/submit.start
-else
-    echo "nothing to submit (-s option set?)"
+# verify picongpu call
+bash tbg/submit.start --validate
+
+if [ $? -eq 0 ] ; then
+  if [ -n "$submit_command" ] ; then
+      $submit_command tbg/submit.start
+  else
+      echo "nothing to submit (-s option set?)"
+  fi
 fi
--- a/etc/picongpu/bash-devServer-hzdr/mpiexec.tpl
+++ b/etc/picongpu/bash-devServer-hzdr/mpiexec.tpl
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Copyright 2013-2022 Axel Huebl, Anton Helm, Rene Widera
+# Copyright 2013-2023 Axel Huebl, Anton Helm, Rene Widera, Pawel Ordyna
 #
 # This file is part of PIConGPU.
 #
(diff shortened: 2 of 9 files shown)


etc/picongpu/bash-devServer-hzdr/mpiexec.tpl
```diff
@@ -44,13 +74,19 @@ umask 0027
 mkdir simOutput 2> /dev/null
 cd simOutput
 
-# test if cuda_memtest binary is available
-if [ -f !TBG_dstPath/input/bin/cuda_memtest ] ; then
-  mpiexec -am !TBG_dstPath/tbg/openib.conf --mca mpi_leave_pinned 0 -npernode !TBG_gpusPerNode -n !TBG_tasks !TBG_dstPath/input/bin/cuda_memtest.sh
+if [[ $VALIDATE_MODE == true ]]; then
+  echo "Validating PIConGPU call..."
+  !TBG_dstPath/input/bin/picongpu !TBG_author !TBG_programParams --validate | tee output
```
PrometheusPi, line 79: Why do we need to validate in the submit script (generated by this `*.tpl`) if `tbg` already checks?

PrometheusPi: Sorry - my fault - `tbg` just calls these lines

PrometheusPi (approved):
extremely useful - thanks for writing this

bin/tbg
```diff
@@ -497,8 +497,13 @@ cp -a "$TBG_cfgFile" tbg/submit.cfg
 #warn, if there are still unresolved !TBG_ variables left
 check_final tbg/submit.start "$tooltpl_file_data"
 
-if [ -n "$submit_command" ] ; then
-    $submit_command tbg/submit.start
-else
-    echo "nothing to submit (-s option set?)"
+# verify picongpu call
+bash tbg/submit.start --validate
```
PrometheusPi, line 501: According to @psychocoderHPC this is not allowed since `tbg` should be independent of PIConGPU.

PrometheusPi (changes requested):

bin/tbg
ax3l: As long as this is done conditionally, any code could support `--validate`.

I would suggest to add a tbg optional CLI argument for this.

etc/picongpu/bash-devServer-hzdr/mpiexec.tpl
```diff
@@ -18,6 +18,36 @@
...
+  echo "--validate      - validate picongpu call instead of running the simulation"
+  echo "--h | --help    - print this help message"
+}
+
+VALIDATE_MODE=false
+for arg in "$@"; do
+  case $arg in
+  --validate)
+    VALIDATE_MODE=true
+    shift # Remove --skip-verification from `$@`
```
ax3l, line 36: Outdated naming of the flag in the comment (`--skip-verification`)?

etc/picongpu/bash/mpiexec.tpl
```diff
@@ -18,6 +18,36 @@
...
+  echo "--validate      - validate picongpu call instead of running the simulation"
+  echo "--h | --help    - print this help message"
+}
+
+VALIDATE_MODE=false
+for arg in "$@"; do
+  case $arg in
+  --validate)
+    VALIDATE_MODE=true
+    shift # Remove --skip-verification from `$@`
```
ax3l, line 36: As above

etc/picongpu/bash/mpirun.tpl
```diff
@@ -18,6 +18,36 @@
 # If not, see <http://www.gnu.org/licenses/>.
 #
 
+help()
+{
+  echo "PIConGPU submit script generated with tbg"
+  echo ""
+  echo "usage: $0 [--verify]"
+  echo ""
+  echo "--validate      - validate picongpu call instead of running the simulation"
```
ax3l, line 27: Looks like there are three different namings here:
- validate
- verify
- skip-verification

pordyna:
From offline discussion with @psychocoderHPC: We will move the call from the `.tpl` files to `submitAction.sh`.

etc/picongpu/bash-devServer-hzdr/mpiexec.tpl
pordyna: ah yes, copy paste mistakes...

bin/tbg
pordyna: I think the problem is that we can't really check if the program that is called by `tbg` supports `--validate`, if not we would just run the program without an allocation. In that case, it would be the users' responsibility to disable the flag. I think the idea to put it in `submitAction.sh` is really good, that makes it independent of `tbg` and we can maybe control it through a PIConGPU environment variable.

PrometheusPi:
@pordyna Does this mean you will close this pull request and resubmit another one, or will you continue to adjust this pull request?

pordyna:
closing it for now

Issue #4228, opened 2022-07-20
[Discussion]Switch boundElectrons particle attribute from float_X to uint32_t


BrianMarre:
Macro particles in PIC are phase space samples following physical trajectories.
As such fractional ion charges are not useful, since physical ions always have integer charges and therefore follow integer charge trajectories, not fractional charge trajectories.

sbastrakov:
Could you comment on the reasoning given for `float_X` choice: https://github.com/ComputationalRadiationPhysics/picongpu/blob/fa0747168cb9e5b09ef83dd065fa09c143d9dc0e/include/picongpu/param/speciesAttributes.param#L104-L114

BrianMarre:
As far as I understand them they are essentially,
- we might use fractional charges in the future
- boundElectrons is required for ionisation
- all ionisation algorithms need to know the atomic number so, atomicNumber is always present if boundElectrons is present and vice versa, also negation

I would argue that from a physics point of view fractional charges are of little use, since they lead to fundamental errors in the PIC-prediction, and we can save memory and some casts in atomic Physics if we use uint instead.

We should actually be able to use uint8_t, since boundElectrons<=Z<=98(Califonium), everything larger is that short lived that creating a target is next to impossible.

PrometheusPi:
As far as I understand, the fractional charge states are already possible due to the reasons, that in principle shielding can (and should) be considered in ionization processes. This is only possible with fractional charge numbers.
However, this makes no sense with respect to the particle pusher. 
Shielding is currently dealt with by Z_eff values. Thus I think converting to integers should be fine.

sbastrakov:
Thanks for clarifications @BrianMarre @PrometheusPi .

chillenzer:
Is this closed?

BrianMarre:
Not yet, we currently are still using float_X for boundElectrons and while the consensus is that we should switch, this will only be done after FLYonPIC_v2 goes to mainline.

TLDR: On my ToDo List

chillenzer:
Thanks!

TLDR: Thanks! ;D

PR #4551, opened 2023-05-04
FLYonPIC v2.0

BrianMarre:
Second version of FLYonPIC
now includes:
- electronic collisional de-/excitation
- spontaneous deexcitation
- electronic collisional ionization
- autonomous ionization
- Stewart-Pyatt ionization potential depression(IPD)
- pressure ionization based on the IPD implementation

as well as a reworked atomic rate solver with adaptive time stepping

still missing:
- [x] FLYonPIC input specification
- [x] compile test
- [x] user facing description

requires
- [x] PR #4918 to be merged first
- [x] PR #4944 to be merged first


BrianMarre:
@PrometheusPi and @steindev for now no review required unless you really want to, a detailed examination together with @psychocoderHPC is already planned

include/picongpu/particles/atomicPhysics2/kernel/SpawnIonizationMacroElectrons.kernel
```diff
@@ -0,0 +1,366 @@
...
+
+            pmacc::DataSpace<simDim> const superCellIdx(
+                areaMapping.getSuperCellIndex(DataSpace<simDim>(cupla::blockIdx(worker.getAcc()))));
+            constexpr uint32_t frameSize = pmacc::math::CT::volume<SuperCellSize>::type::value;
+
+            auto forEachFrameSlot = pmacc::lockstep::makeForEach<frameSize, T_Worker>(worker);
+            auto forEachFrameMaster = pmacc::lockstep::makeForEach<2u, T_Worker>(worker);
+            auto onlyMaster = pmacc::lockstep::makeMaster(worker);
+
+            IonFramePtr ionFrame = ionBox.getLastFrame(superCellIdx);
```
psychocoderHPC: You must store this in shared memory if the master only is updating it to the next frame. Do not forget a sync after loading the last frame.
You can keep this in register if all worker update `ionFrame`, so changing

``` C++
onlyMaster(
                    [&ionFrame, &ionBox]()
                    {
                        ionFrame = ionBox.getPreviousFrame(ionFrame);
                        // no need to set "totalNumberMacroElectronsToSpawn = 0",
                        //  already guaranteed by break condition
                    });
```

to 

```C++
                        ionFrame = ionBox.getPreviousFrame(ionFrame);

```

is doing the same.

psychocoderHPC (changes requested):

include/picongpu/particles/atomicPhysics2/kernel/SpawnIonizationMacroElectrons.kernel
BrianMarre: I see where I went wrong, thanks that would have taken me ages

include/picongpu/particles/atomicPhysics2/rateCalculation/BoundBoundTransitionRates.hpp
```diff
@@ -122,8 +121,8 @@ namespace picongpu::particles::atomicPhysics2::rateCalculation
                 g = A * math::log(U) + B + C / (U + a) + D / ((U + a) * (U + a)); // unitless
```
psychocoderHPC: If you move `math::log(U)` before the loop the code will be faster `auto const logU = math::log(U);`

psychocoderHPC: **Attention** possible division by zero in case where `(U + a) == 0` This should be handled!

psychocoderHPC: **Attention2** `math::log(U)` can be NAN depending on the alpaka backend in cases where `U` is negative.
So we need to take care that `U` is guaranteed positive!

include/picongpu/param/physicalConstants.param
```diff
@@ -46,18 +46,22 @@ namespace picongpu
 
         // Electron properties
         /** unit: kg */
-        constexpr float_64 ELECTRON_MASS_SI = 9.109382e-31;
+        constexpr float_64 ELECTRON_MASS_SI = 9.1093837015e-31;
```
psychocoderHPC: feel free to push the precision change in a separate PR before you atomic PR.

include/picongpu/particles/atomicPhysics2/initElectrons/InitElectronAs.hpp
```diff
@@ -0,0 +1,266 @@
...
+            //      choose direction
+            float_X const u = rngGenerator();
+            float_X const v = rngGenerator();
+
+            float_X const cosTheta = 1._X - 2._X * v;
+            float_X const sinTheta = math::sqrt(v * (2._X - v));
+            float_X const phi = 2._X * static_cast<float_X>(picongpu::PI) * u;
+
+            float3_64 const directionVector
+                = float3_64(sinTheta * math::cos(phi), sinTheta * math::sin(phi), cosTheta);
```
psychocoderHPC: `pmacc::math::sincos()` can be used here

include/picongpu/particles/atomicPhysics2/initElectrons/InitElectronAs.hpp
```diff
@@ -0,0 +1,266 @@
...
+            // UNIT_MASS^2 * UNIT_LENGTH^2 / UNIT_TIME^2, not weighted
+            float_64 const momentumSquaredIon_LabSystem
+                = static_cast<float_64>(pmacc::math::l2norm2(ion[momentum_]) / ion[weighting_]);
+
+            // square of original beta of the ion in Lab-System
+            // beta^2 = 1/(1 + (m^2*c^2)/p^2)
+            //  unitless + (UNIT_MASS^2 * UNIT_SPEED^2)(not weighted)
+            //  /( UNIT_MASS^2 * UNIT_SPEED^2)(not weighted) = unitless
+            float_64 const normSquaredBetaIon_LabSystem
+                = 1._X / (1 + mSquaredCSquaredIon / momentumSquaredIon_LabSystem);
```
psychocoderHPC: is this intended that `1 + ...` is used and `1` is integer?Our should it be `1.0_X`?

psychocoderHPC: Attention ` 1._X` can be 32bit and the result is 64bit maybe `1.0` should be used everywhere.

include/picongpu/particles/atomicPhysics2/initElectrons/InitElectronAs.hpp
```diff
@@ -0,0 +1,266 @@
...
+                // magnitude
+                math::sqrt(normSquaredBetaIon_LabSystem)
+                // direction
+                * static_cast<float3_64>(ion[momentum_] / pmacc::math::l2norm(ion[momentum_]));
+
+            // unitless
+            auto beta = MatrixVector(betaIonSystem);
+
+            float_64 const gammaIonSystem
+                = math::sqrt(1. / (1. - pmacc::math::cPow(normSquaredBetaIon_LabSystem, 2u)));
```
psychocoderHPC: **attention* possible division by zero if  `normSquaredBetaIon_LabSystem == 1.0`

include/picongpu/particles/atomicPhysics2/initElectrons/InitElectronAs.hpp
```diff
@@ -0,0 +1,266 @@
...
+            //  /( UNIT_MASS^2 * UNIT_SPEED^2)(not weighted) = unitless
+            float_64 const normSquaredBetaIon_LabSystem
+                = 1._X / (1 + mSquaredCSquaredIon / momentumSquaredIon_LabSystem);
+
+            // unitless
+            float3_64 const betaIonSystem =
+                // magnitude
+                math::sqrt(normSquaredBetaIon_LabSystem)
+                // direction
+                * static_cast<float3_64>(ion[momentum_] / pmacc::math::l2norm(ion[momentum_]));
```
psychocoderHPC: **BUG??** division by zero in cases where the ion is not moving!!

include/picongpu/particles/atomicPhysics2/initElectrons/InitElectronAs.hpp
```diff
@@ -0,0 +1,266 @@
...
+        {
+// general contributions
+#pragma unroll
+            for(uint32_t i = 0u; i < 3u; i++)
+            {
+#pragma unroll
+                for(uint32_t j = static_cast<uint32_t>(0u); j < static_cast<uint32_t>(3u); j++)
+                {
+                    lorentzMatrix.element(i, j) = (gammaIonSystem - 1.) * beta.element(i, static_cast<uint32_t>(0u))
+                        * beta.element(j, static_cast<uint32_t>(0u)) / normBetaSquared;
```
psychocoderHPC: division by zero in cases where `normBetaSquared`

psychocoderHPC (changes requested):
I checked mostly for division by zero and found some critical parts, please check where it can happen and if function parameter must be `>0` to avoid a decision by zero please add it to the oxygen of the function that we do not forget about it.

include/picongpu/particles/atomicPhysics2/rateCalculation/BoundBoundTransitionRates.hpp
BrianMarre: good catch, U is always >0 due to being the central value of a histogram bin of a histogram over the range [0, max], nevertheless will document the assumption

BrianMarre: fixed

include/picongpu/param/physicalConstants.param
BrianMarre: will do!

include/picongpu/particles/atomicPhysics2/initElectrons/InitElectronAs.hpp
BrianMarre: done!

include/picongpu/particles/atomicPhysics2/initElectrons/InitElectronAs.hpp
BrianMarre: You are right I intended 64 bit every where, oops

include/picongpu/particles/atomicPhysics2/initElectrons/InitElectronAs.hpp
BrianMarre: fixed

BrianMarre: from a maths and physics point of view an ion beta will always be less than zero, i.e. will never move with light speed.
For very highly relativistic ions floating point rounding errors may create beta values >= 1, but this are highly unlikelyin our simulations, since even a 1GeV hydrogen ion only reaches a beta of 0.7 .
Will create a comment but no check.

include/picongpu/particles/atomicPhysics2/initElectrons/InitElectronAs.hpp
BrianMarre: good catch, fixed!

include/picongpu/particles/atomicPhysics2/initElectrons/InitElectronAs.hpp
BrianMarre: fixed

include/picongpu/particles/atomicPhysics2/atomicData/BoundFreeTransitionData.hpp
```diff
@@ -17,7 +17,7 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
-#pragma once
+#pragma onc
```
psychocoderHPC: @BrianMarre typo!

BrianMarre: thanks, good catch!

include/picongpu/particles/atomicPhysics2/initElectrons/Inelastic2BodyCollisionFromCoMoving.hpp
```diff
@@ -0,0 +1,278 @@
...
+                + (beta.sMul(gammaIonSystem * gammaStarIon_IonSystem * mcIon));
+
+            //          electron
+            // UNIT_MASS * UNIT_LENGTH / UNIT_TIME, weighted
+            momentumStarElectron_LabSystem = momentumStarElectron_LabSystem
+                + (beta.sMul(gammaIonSystem * gammaStarElectron_IonSystem * mcElectron));
+
+            // set to particle
+#pragma unroll
+            for(uint32_t i = 0u; i < picongpu::simDim; i++)
```
psychocoderHPC: momentum has always 3 components! even in a 2D siulation

include/picongpu/particles/atomicPhysics2/initElectrons/Inelastic2BodyCollisionFromCoMoving.hpp
```diff
@@ -0,0 +1,278 @@
...
+                ion[momentum_][i]
+                    = static_cast<float_X>(momentumStarIon_LabSystem.element(i, static_cast<uint32_t>(0u)));
+
+                // set to particle
+#pragma unroll
+            for(uint32_t i = 0u; i < picongpu::simDim; i++)
```
psychocoderHPC: should run over three components too

include/picongpu/particles/atomicPhysics2/initElectrons/Inelastic2BodyCollisionFromCoMoving.hpp
BrianMarre: fixed

include/picongpu/particles/atomicPhysics2/initElectrons/Inelastic2BodyCollisionFromCoMoving.hpp
```diff
@@ -0,0 +1,278 @@
...
+
+            float_64 const gammaIonSystem = math::sqrt(momentumSquaredIon_LabSystem / mSquaredCSquaredIon + 1.);
+
+            // lower 3x3 block of Lorentz transformation matrix for transformation from
+            // Ion-System to Lab-System
+            Matrix_3x3 lorentzMatrix;
+            fillLorentzMatrix(beta, gammaIonSystem, normSquaredBetaIon_LabSystem, lorentzMatrix);
+
+            //      space components of Lorentz boost
+            MatrixVector momentumStarElectron_LabSystem;
```
psychocoderHPC: This instance is not initialized because internal MAtrix is using an pmacc::Array which is not initializing its data

BrianMarre: yes, which is why I am initializing it in the line below

include/picongpu/particles/atomicPhysics2/initElectrons/Inelastic2BodyCollisionFromCoMoving.hpp
BrianMarre: fixed

include/picongpu/particles/atomicPhysics2/interfaceDefintion.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <pmacc/math/Vector.hpp>
+
+namespace picongpu::particles
+{
+    namespace compileTime = pmacc::math::CT;
+
+    // in-cell position to initialize macro particle to
+    CONST_VECTOR(
+        /*datatype*/ float_X,
+        /*dim*/ 3,
```
psychocoderHPC: should be `simDim` or is atomic physics assuming always 3 components?

include/picongpu/particles/atomicPhysics2/interfaceDefintion.hpp
```diff
@@ -0,0 +1,36 @@
...
+        /*name*/ InCellOffset,
+        /*x*/ 0.0,
+        /*y*/ 0.0,
+        /*z*/ 0.0);
+
+    // cell in superCell to initialize macro particle to
+    CONST_VECTOR(uint32_t, 3, CellIdx, 0u, 0u, 0u);
```
psychocoderHPC: should be `simDim` or is atomic physics assuming always 3 components?

BrianMarre: fixed

include/picongpu/particles/atomicPhysics2/interfaceDefintion.hpp
BrianMarre: that is part of the rejected `one particle per superCell position functor` that will not go into the mainline anyway. but yes that should have probably accounted for simDim 2.

include/picongpu/particles/atomicPhysics2/interfaceDefintion.hpp
BrianMarre: will not go to mainline as it us part of a rejected pull request

include/picongpu/particles/atomicPhysics2/initElectrons/Inelastic2BodyCollisionFromCoMoving.hpp
```diff
@@ -120,60 +181,44 @@ namespace picongpu::particles::atomicPhysics2::initElectrons
                             "is unphysical for deltaEnergy < 0, deltaEnergy = %.8f !\n",
                             deltaEnergy);
 
-                electron[momentum_] = float3_X::create(0._X);
-                ion[momentum_] = float3_X::create(0._X);
+                electron[momentum_] = ion[momentum_];
```
psychocoderHPC: @BrianMarre  you need to take the weighting into account because momentum in PIConGPU is weighted.

BrianMarre: electron and ion weight are equal, due to the electron being created from the ion through ionization.

BrianMarre: offline discussion; but psychocoderHPC nevertheless correct, while weights are equal, mass is not. I have to adjust the momentum for the different mass

BrianMarre:
@ikbuibui @chillenzer it is now at a state where a review from you would be helpful

chillenzer:
Wow! So, how many months do we get for 20k+ lines of code?

BrianMarre:
@psychocoderHPC do you have any idea why the CI fails at the KHI test/example? I am scratching my head on that one.

BrianMarre:
@psychocoderHPC please restart the KHI CI job, failed again with message "KILLED"

BrianMarre:
@psychocoderHPC @chillenzer @ikbuibui @PrometheusPi @steindev ready for review

BrianMarre:
@psychocoderHPC @PrometheusPi CI fails in the KHI_growth rate test with,
``` bash
[ 78%] Built target picongpu-hostonly
[ 80%] Building CUDA object CMakeFiles/picongpu.dir/main.cpp.o
Killed
make[2]: *** [CMakeFiles/picongpu.dir/build.make:76: CMakeFiles/picongpu.dir/main.cpp.o] Error 137
make[1]: *** [CMakeFiles/Makefile2:243: CMakeFiles/picongpu.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```
I am not sure how to avoid this, I think psychocoderHPC suggested that this may be due to the CI running out of memory, but I am unsure why this would be specific to the FLYonPIC PR, other PRs CI tests pass fine and atomicPhysics should not be active at all in the KHI_growth test.

In summary my guess is that I am doing something wrong but I am unsure what exactly. Any help would be appreciated.

PrometheusPi:
If the single KHI_growthrate test job fails again, we have an issue with 64bit precision for that specific setup. 
This would require a detailed study, since this would then reveal unexpected side effects.

chillenzer:
Hi, here's a first batch of comments. These are more or less general comments, more detailed code comments as well as comments concerning the rest of the files will follow. In general, I will tag the comments, etc. with one of the following pre-fixes:

REQUEST: Necessary change (or at least necessary discussion).

REFACTOR: Nice-to-have but given the circumstances, we should use the strategy from [here](https://github.com/ComputationalRadiationPhysics/picongpu/issues/4912) and create an issue with such comments to address them later.

CONSCIOUS DECISION: A decision that has a trade-off. By commenting on this, I want to provoke a conscious decision that this is the best trade-off. You can resolve immediately (or with a short reasoning) if you think you thought this through.

QUESTION: More for my understanding than to entail code changes.

Please excuse that these are abbreviated to the edge of being impolite. They are copied more or less directly from my even shorter notes.

### Comments:

- CONSCIOUS DECISION: `atomicPhysics_Debug.param`
    - It surprised me that they exist and are different in the `tests` folder:
        - Some of the settings are backend-specific. Does this render the tests backend-specific?
        - Wouldn't we want to test the production version and not some debug version that might behave differently? (If only by different scheduling in multi-threading.)
        - The tests are compilation-only, so why activate debug there (assuming it's runtime debug)?
- REQUEST: Tests are missing `README`s.
- REFACTOR: An example would be nice:
	- Or are the tests also valid physical cases?
	- If so, having an example with more context and explanation would probably still be beneficial.
- CONSCIOUS DECISION: `atomicPhysics.param` is identical between default and `tests/compileAtomicPhysics`. Should we remove the latter?
- REQEUST: Is the FLYonPIC step diagram in the docs up to date? (At some point you showed me a version of this that you said was outdated, so just checking.)
	- Where does the arrow from "binElectrons" lead to?
	- What does this `KERNEL (/transitionType)` mean floating in the middle of nowhere?
	- What does the `true` node mean?
- REQUEST: Why are there any changes in `include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp`?
- CONSCIOUS DECISION: Why remove DOI in `docs/source/usage/workflows/ionization.rst`?
- REQUEST: Unnecessary and questionable changes in `docs/source/models/collisional_ionization.rst`.

PrometheusPi (changes requested):
The CI test (`g++-11_cuda11.7_boost1.78.0_tests.KHI_growthRate`, KHI growth rate test cmake preset 1) fails consistently. There seems to be a consistent bug with a side dependency we do not yet understand. 
Until this side effect is not understood and solved, this pull request should be blocked.

BrianMarre:
@chillenzer
REQUEST FLYonPIC diagram:
- FLYonPIC diagram is updated
- good catch, I missed the extra floating text and too short life line for the electron histogram, fixed
- the true node indicates a `while(true)` , did not find a better way to draw it with PlantUML

REQUEST changes in ``docs/source/models/collisional_ionization.rst`` and ``include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp``:

- I am not sure what exactly you mean, according to Github the PR does not touch ``docs/source/models/collisional_ionization.rst`` or ``include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp``?

REFACTOR add Examples:
Well, the fundamental problem is that the atomic data input files I am using are not public, therefore we are unable to provide a setup that works at runtime.

docs/source/models/atomic_physics.rst
```diff
@@ -0,0 +1,230 @@
...
+
+FLYonPIC is based on averaged states and models the evolution of the atomic state distribution time and space resolved,
+
+- depending on the local plasma conditions
+- without assuming a temperature or equilibrium
+- self consistent with the PIC simulation.
+
+This allows a more accurate prediction of ionization dynamics and the prediction of the atomic state distributions in non-thermal, transient plasmas, for example high-intensity short-pulse laser plasma interactions.
+
+FLYonPIC is base on the [FlyCHK]_ atomic model and currently implements the following processes:
```
chillenzer:
```suggestion
FLYonPIC is based on the [FlyCHK]_ atomic model and currently implements the following processes:
```

docs/source/models/atomic_physics.rst
```diff
@@ -0,0 +1,230 @@
...
+
+Model Overview
+--------------
+
+FLYonPIC is based on tracking the atomic state of ions as a particle attribute of ion macro particles, with each "macro-ion" representing a phase space sample of exactly one atomic state.
+
+.. note::
+   FLYonPIC tracks only FlyCHK super configurations.
+
+   i.e. FLYonPIC distinguishes only between states differing in the number of electron contained in atomic shells
```
chillenzer:
```suggestion
   FLYonPIC tracks only FlyCHK super configurations, i.e., FLYonPIC distinguishes only between states differing in the number of electrons contained in atomic shells.
```

docs/source/models/atomic_physics.rst
```diff
@@ -0,0 +1,230 @@
...
+
+   Example: (1s^2)(2s^1 2p^2) and (1s^2)(2p^3) would be considered the same state (2,3,0, ...) by FLYonPIC.
+
+The atomic state dynamics are modelled as some average rate of change to another atomic state, dependent current state, transition and local plasma conditions.
+
+Over the set of all transitions, atomic state dynamics may be represented as a matrix of rates describing the change of the entire atomic state population with time.
+
+This rate equation is solved in FLYonPIC by a monte-carlo solver for every macro-ion individually and time dependent.
```
chillenzer:
```suggestion
This rate equation is solved in FLYonPIC by a Monte Carlo solver for every macro-ion individually and time dependent.
```

include/picongpu/particles/atomicPhysics/electronDistribution/LocalHistogramField.hpp
```diff
@@ -0,0 +1,130 @@
...
+     *
+     * @attention histograms are uninitialized upon creation of the field, use .getDeviceBuffer().setValue() to init
+     */
+    template<typename T_Histogram, typename T_MappingDescription>
+    struct LocalHistogramField : SuperCellField<T_Histogram, T_MappingDescription, false /*no guards*/>
+    {
+        using Histogram = T_Histogram;
+
+    private:
+        /// @todo should these be private?
```
chillenzer: QUESTION: What's the status of this?

include/picongpu/param/speciesAttributes.param
```diff
@@ -114,17 +115,201 @@ namespace picongpu
...
      * - float_X instead of integer types are reasonable because effective charge
-     *   numbers are possible
+     *   numbers are possible, NOTE: but not useful, Brian, 2022
      * - required for ion species if ionization is enabled
      * - setting it requires atomicNumbers to also be set
+     * @attention number bound electrons should always match atomicState
+     *  charge state!
      *
      * @todo connect default to proton number
+     * @todo change to uint8_t or uint32_t
```
chillenzer, line 125: QUESTION: What's the status of this?

include/picongpu/param/speciesAttributes.param
```diff
@@ -114,17 +115,201 @@ namespace picongpu
...
+     *
+     * The atomic data type specifies data types used in storage and active atomicPhysics paths
+     */
+    alias(atomicDataType);
+
+    /** path to file containing charge state input data for atomicPhysics for an ion species
+     *
+     * file gives parameter of atomic physics charge states for ion species
+     * - must follow PIConGPU specific formatting
+     *      @todo write specification, Brian Marre, 2023
```
chillenzer: QUESTION: What's the status of this?

include/picongpu/param/speciesAttributes.param
```diff
@@ -114,17 +115,201 @@ namespace picongpu
...
+     * - files are not shipped with PIConGPU, see documentation for creation and sources
+     *      @todo create documentation, Brian Marre, 2022
```
chillenzer: QUESTION: What's the status of this?

include/picongpu/param/speciesAttributes.param
```diff
@@ -114,17 +115,201 @@ namespace picongpu
...
+     * - all paths are either relative to the simOutput folder, or absolute
+     */
+    alias(chargeStatesFileName);
+
+    /** path to file containing atomic state input data for atomicPhysics for an ion species
+     *
+     * file gives parameter of atomic physics atomic states for ion species
+     * - must follow PIConGPU specific formatting
+     *      @todo write specification, Brian Marre, 2023
```
chillenzer: QUESTION: What's the status of this?

include/picongpu/param/speciesAttributes.param
```diff
@@ -114,17 +115,201 @@ namespace picongpu
...
+     * - files are not shipped with PIConGPU, see documentation for creation and sources
+     *      @todo create documentation, Brian Marre, 2022
```
chillenzer: QUESTION: What's the status of this?

include/picongpu/param/speciesAttributes.param
```diff
@@ -114,17 +115,201 @@ namespace picongpu
...
+     *       if at least one upward bound free transition exists.
+     *          i.e. the final state reachable by a defined upward bound-free transition from the current state
+     *           with energy closest to the initial state.
+     *      3.) disable pressure ionization for this state.
+     *
+     * further notes:
+     * - setting an atomic states the pressure ionization state equal to the state disables pressure ionization
+     *      for this states
+     * - file must follow PIConGPU specific formatting
+     *      @todo write specification, Brian Marre, 2023
```
chillenzer: Okay... this applies just for all todos: QUESTION: What's the status of this?

include/picongpu/particles/atomicPhysics/kernel/BinElectrons.kernel
```diff
@@ -0,0 +1,104 @@
...
+    template<typename T_Histogram>
+    struct BinElectronsKernel
+    {
+        /** call operator
+         *
+         * called by BinElectrons atomic physics sub-stage
+         *
+         * @param worker object containing the device and block
+         *  information, passed by PMACC_KERNEL call
+         * @param areMapping mapping of blockIndex to block superCell index
```
chillenzer:
```suggestion
         * @param areaMapping mapping of blockIndex to block superCell index
```

include/picongpu/simulation/stage/AtomicPhysics.hpp
```diff
@@ -0,0 +1,604 @@
...
+                // FoundUnboundIonSuperCellField
+                auto& localFoundUnboundIonField = *dc.get<S_FoundUnboundField>("LocalFoundUnboundIonField");
+                DataSpace<picongpu::simDim> const fieldGridLayoutFoundUnbound
+                    = localFoundUnboundIonField.getGridLayout().sizeWithoutGuardND();
+
+                // ElectronHistogramOverSubscribedSuperCellField
+                auto& localElectronHistogramOverSubscribedField
+                    = *dc.get<S_OverSubscribedField>("LocalElectronHistogramOverSubscribedField");
+                DataSpace<picongpu::simDim> const fieldGridLayoutOverSubscription
+                    = localElectronHistogramOverSubscribedField.getGridLayout().sizeWithoutGuardND();
```
chillenzer: REFACTOR: For these variables it would have been nice to have a
```C++
struct LocalField
    field;
    layout;
    
    LocalField(&dc) : field{*dc.get...}, layout{field.getGridLayout()...}
};
```
to have these 14 lines down to three.

REQUEST: Also: Remove comments. They are redundant with the (quite good!) naming.

include/picongpu/simulation/stage/AtomicPhysics.hpp
```diff
@@ -0,0 +1,604 @@
...
+
+                setTimeRemaining(); // = (Delta t)_PIC
+                // fix atomic state and charge state inconsistency
+                using ForEachIonSpeciesFixAtomicState = pmacc::meta::
+                    ForEach<AtomicPhysicsIonSpecies, particles::atomicPhysics::stage::FixAtomicState<boost::mpl::_1>>;
+                ForEachIonSpeciesFixAtomicState{}(mappingDesc);
+
+                uint16_t counterSubStep = 0u;
+                // atomicPhysics sub-stepping loop, ends when timeRemaining<=0._X
+                while(true)
```
chillenzer: REFACTOR: I would have preferred an actual condition or even better a for-loop here. Reason:
 1. It is hard to comprehend, check and verify the control flow with all those hidden `break`
 statements scattered around.
 2. In my head the following three have different semantics:

    - while(true):
      Infinite event loop that's supposed to run forever or until terminated by an external event.
    - while(condition):
      An iterative algorithm that does not progress but has a boolean success criterion, e.g.,
      hit-or-miss algorithms that return on the first hit.
    - for(...)
      An iterative algorithm that progresses in some meaningful way (not necessarily by `++i`)

 Admittedly, the border between while(condition) and for(...) is relatively vague and there is room
 for interpretation and personal preference there. But while(true) clearly communicates something
 different.

include/picongpu/simulation/stage/AtomicPhysics.hpp
```diff
@@ -0,0 +1,604 @@
...
+                {
+                    // debug only
```
chillenzer: REQUIRED: This does not look like "debug only". Remove comment or statement.

include/picongpu/traits/attribute/GetMass.hpp
```diff
@@ -34,7 +34,8 @@ namespace picongpu
              *
              * @param weighting weighting of the particle
              * @param particle a reference to a particle
-             * @return mass of the makro particle
+             *
+             * @return mass of the macro particle, UNIT_MASS
```
chillenzer, line 38: QUESTION: Why that addition? Is this pertinent?

include/pmacc/mappings/kernel/MappingDescription.hpp
```diff
@@ -87,6 +87,11 @@ namespace pmacc
             return this->gridSuperCells;
         }
 
+        HDINLINE DataSpace<DIM> getGridSuperCellsWithoutGuards() const
+        {
+            return this->gridSuperCells - 2 * this->guardingSuperCells;
```
chillenzer, line 92: QUESTION: Does the PMAcc model guarantee that the guards are symmetric?

share/picongpu/tests/compileAtomicPhysics/include/picongpu/param/atomicPhysics.param
```diff
@@ -0,0 +1,101 @@
...
+        // eV
+        static constexpr float_X maxEnergy = float_X(1e7);
+    };
+    using ElectronHistogram = particles::atomicPhysics::electronDistribution::LogSpaceHistogram<
+        100u, // T_numberBins
+        MaxEnergyParam>;
+    //}
+
+    // atomicPhysics rate solver settings
+    //{
```
chillenzer: REFACTOR: Do I understand this correctly that the curly braces are kind of creating an implicit subsection here? If so, why not use a struct and make it explicit? If not, this is likely deada code and should be removed.

share/picongpu/tests/compileAtomicPhysics/include/picongpu/param/atomicPhysics.param
```diff
@@ -0,0 +1,101 @@
...
+        float_X, // T_Value
+        uint32_t, // T_CollectionIndex
+        ConfigNumber_Argon, // T_ConfigNumber, data type used for storage
+        float_64, // T_Multiplicities, data type used for storage
+        true, // T_electronicExcitation
+        true, // T_electronicDeexcitation
+        true, // T_spontaneousDeexcitation
+        true, // T_electronicIonization
+        true, // T_autonomousIonization
+        false>; // T_fieldIonization
```
chillenzer, line 102: REFACTOR: Could these be enums such that we say `ElectronicExcitation::On` and so on?

share/picongpu/tests/compileAtomicPhysics/include/picongpu/param/speciesInitialization.param
```diff
@@ -0,0 +1,53 @@
...
+// init of particles with
+namespace picongpu
+{
+    namespace particles
+    {
+        /** InitPipeline define in which order species are initialized
+         *
+         * the functors are called in order (from first to last functor)
+         */
+        using InitPipeline = boost::mp11::mp_list<
```
chillenzer: QUESTION: Why not use a `pmacc::mp_list<>` here as per default? It's currently the same but it might not be in the future!

share/picongpu/tests/compileSparser/include/picongpu/param/speciesInitialization.param
```diff
@@ -0,0 +1,53 @@
...
+// init of particles with
+namespace picongpu
+{
+    namespace particles
+    {
+        /** InitPipeline define in which order species are initialized
+         *
+         * the functors are called in order (from first to last functor)
+         */
+        using InitPipeline = boost::mp11::mp_list<
```
chillenzer: same here

chillenzer (changes requested):
First batch of comments. From what I've just typed here, I'd say pretty good work so far. Mostly typos and stuff... But there's more to come! (... probably, well, I haven't checked yet. That's the whole point, you see.)

chillenzer:
>  I am not sure what exactly you mean, according to Github the PR does not touch `docs/source/models/collisional_ionization.rst` or `include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp`?

Okay. Then, ignore it. I pulled one specific commit last week Thursday night which I also found is outdated by now.

> REFACTOR add Examples: Well, the fundamental problem is that the atomic data input files I am using are not public, therefore we are unable to provide a setup that works at runtime.

Okay. (Well, not okay philosophically and concerning open science, etc., but that's probably not something we want to solve before merging this.)

include/picongpu/param/speciesAttributes.param
BrianMarre: this has been discussed in the #4228, last agreement was that yes we want to change boundElectrons to `uint`, but this should be done in a separate PR

include/picongpu/particles/atomicPhysics/electronDistribution/LocalHistogramField.hpp
BrianMarre: I would keep them private, but psychocoder's opinion differed in the past. therefore marked for discussion

include/pmacc/mappings/kernel/MappingDescription.hpp
BrianMarre: I actually not sure, @psychocoderHPC care to chime in?

share/picongpu/tests/compileAtomicPhysics/include/picongpu/param/atomicPhysics.param
BrianMarre: they are supposed to indicate that the comment describes several lines, start and end indicated by the braces.
I copied this style from psychocoder but would be open to change it

include/picongpu/param/speciesAttributes.param
BrianMarre: done, will remove

include/picongpu/param/speciesAttributes.param
BrianMarre: done, I removed the todo

include/picongpu/param/speciesAttributes.param
BrianMarre: done removed the todo

include/picongpu/param/speciesAttributes.param
BrianMarre: done, removed the todo

include/picongpu/param/speciesAttributes.param
BrianMarre: done, removed the todo

include/picongpu/param/speciesAttributes.param
chillenzer: Okay. Did you go through `grep -i todo -r .` and check the status of everything?

include/picongpu/simulation/stage/AtomicPhysics.hpp
BrianMarre: A for-loop is not possible, since we do not know in advance how many sub step we will do.

Fundamentally the entire loop body needs to be executed at least once therfore the best fit would be a do-while loop, but I am unsure how often people actually see do-whiles.

Open to changing it to do-while loops.

include/picongpu/traits/attribute/GetMass.hpp
BrianMarre: I changed the GetMass documentation to fit to the documentation style used for atomicPhyisc returns as it is used repeatedly in the atomcPhysics,  and while I was at it I fixed the typo.

Could be moved to separate PR, or removed, I am open to both.

share/picongpu/tests/compileAtomicPhysics/include/picongpu/param/atomicPhysics.param
BrianMarre: yes they could, but that entail changing a lot of kernels since they are used to toggle code branches on and off.

Would want to push that to a future refactor.

share/picongpu/tests/compileSparser/include/picongpu/param/speciesInitialization.param
BrianMarre: fixed

share/picongpu/tests/compileAtomicPhysics/include/picongpu/param/speciesInitialization.param
BrianMarre: the input is based on an older version, did not notice that we changed this bit, good catch fixed

include/picongpu/simulation/stage/AtomicPhysics.hpp
BrianMarre: done

include/picongpu/simulation/stage/AtomicPhysics.hpp
BrianMarre: comments removed

include/picongpu/traits/attribute/GetMass.hpp
BrianMarre: In general the problem I ran into is that atomicPhysics needs a different normalization than the PIC-loop, since
atomicPhysics deals with fundamentally much smaller quantities and therefore I had to sometime use different units than the main code to avoid numerical errors.

Consequently I had to document what the unit the return values of some of my functions were and to make less errors in the internal unit conversions I created documentation to keep track of units.

psychocoderHPC:
The failing CI test is the KHi with double precision. It is running into the time limit of 3h.

psychocoderHPC:
@BrianMarre please rebase against the dev branch. We removed ISAAC support, maybe this helps to pass the CI. This will reduce compile overhead, if not there is an issue with 64bit PIConGPU together with atomic physics.

include/pmacc/mappings/kernel/MappingDescription.hpp
psychocoderHPC: Guarding supercells can be different per dimension.

Why is this change here in this PR? looks like it is already merged.

docs/source/models/atomic_physics.rst
BrianMarre: fixed

docs/source/models/atomic_physics.rst
BrianMarre: fixed

docs/source/models/atomic_physics.rst
BrianMarre: fixed

include/picongpu/particles/atomicPhysics/kernel/BinElectrons.kernel
BrianMarre: fixed

include/picongpu/particles/atomicPhysics/electronDistribution/LocalHistogramField.hpp
BrianMarre: removed comment, will keep the string private to prevent name changes after creation

share/picongpu/tests/compileAtomicPhysics/include/picongpu/param/atomicPhysics.param
BrianMarre: followed your suggestion, and created a param struct

include/pmacc/mappings/kernel/MappingDescription.hpp
BrianMarre: @psychocoderHPC But are they the same for the two different directions of one dimension? 

I would assume yes, since MappingDescription stores only one value per dimension.

Note: This method is required by in the FLYonPIC SuperCell fields which are newly added with this PR, therefore this change also belongs to this PR

include/picongpu/simulation/stage/AtomicPhysics.hpp
BrianMarre: refactored to use do-while loops

chillenzer: Offline discussion yielded some good results but just for the record: A `for` loop is exactly the same as a `while` loop except for the fact that you have the option to give it a bit more structure by using the initialisation and increment field.

```C++
for(<initialisation>, <condition>, <increment>) {
    ...
}
```
is identical to
```C++
<initialisation>
while(<condition>) {
    ...
    <increment>
}
```
(except for scoping).

And the other way round
```C++
while(<condition>) {
    ...
}
```
is the same as
```C++
for (; <condition>; ) {
    ...
}
```

The choice is really about what you want to communicate.

include/pmacc/mappings/kernel/MappingDescription.hpp
psychocoderHPC: Yes guard size is per dimension and equal for the negative and positive direction.

include/picongpu/traits/attribute/GetMass.hpp
psychocoderHPC: @chillenzer not strongly necessary but okisch.

include/picongpu/simulation/stage/AtomicPhysics.hpp
```diff
@@ -350,63 +352,61 @@ namespace picongpu::simulation::stage
...
-                        // remove overSubscription loop, ends when overSubscription ended by rejecting enough
-                        // transitions
-                        while(true)
+                        if(histogramOverSubscribed)
                         {
-                            // debug only
-                            if constexpr(picongpu::atomicPhysics::debug::kernel::rollForOverSubscription::
-                                             PRINT_DEBUG_TO_CONSOLE)
+                            // remove overSubscription
+                            bool noHistogramOverSubscribed = false;
```
psychocoderHPC: no double negation please!

psychocoderHPC (changes requested):

include/picongpu/simulation/stage/AtomicPhysics.hpp
```diff
@@ -468,16 +466,13 @@ namespace picongpu::simulation::stage
...
-                    // timeRemaining <= 0? in all local superCells?
-                    if(deviceLocalReduce(
-                           pmacc::math::operation::Max(),
-                           linearizedTimeRemainingBox,
-                           fieldGridLayoutTimeRemaining.productOfComponents())
-                       <= 0._X)
-                    {
-                        break;
-                    }
+                    timeRemainingLessThanZero = deviceLocalReduce(
```
psychocoderHPC: no need to reflect the condition in the name. The name should be `stopExecuion`, `isValid...`

psychocoderHPC (changes requested):

include/picongpu/simulation/stage/AtomicPhysics.hpp
```diff
@@ -302,8 +303,9 @@ namespace picongpu::simulation::stage
                         particles::atomicPhysics::stage::CalculateStepLength<boost::mpl::_1>>;
                     ForEachIonSpeciesCalculateStepLength{}(mappingDesc);
 
-                    // reject overSubscription loop, ends when no histogram bin oversubscribed
-                    while(true)
+                    // choose transition loop
+                    bool histogramOverSubscribed = true;
```
psychocoderHPC: later you have a variable named `noHistogramOverSubscribed`. The namings are confusing, they sounds euqal but most likely they are not.

include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/kernel/CalculateIPDInput.kernel
```diff
@@ -166,4 +172,35 @@ namespace picongpu::particles::atomicPhysics::ionizationPotentialDepression::ker
                 });
         }
     };
+
+    template<>
+    struct CalculateIPDInputKernel<0u>
```
psychocoderHPC: please remove, it looks like it is not required

include/picongpu/param/speciesAttributes.param
```diff
@@ -147,146 +147,12 @@ namespace picongpu
...
-     * - atomicStateCollectionIndex
-     * - processClass
-     * - transitionIndex
-     * - binIndex
-     * - accepted
-     */
-    alias(isAtomicPhysicsIon);
+    //! configuration of species for atomicPhysics algorithm, see picongpu/particles/atomicPhysics/ParticleType.hpp for
+    //! details
+    alias(atomicPhysics)
```
psychocoderHPC: The CI says that `atomicPhysics` is colliding with the namespace `atomicPhysics`. Unfortunally you need to change one.

psychocoderHPC: maybe change all namespaces to `atomicPhy`? This will be a simple search and replace over the code (2min work).

psychocoderHPC: wait, maybe you missed the semicolon only should be `alias(atomicPhysics);`

include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/kernel/CalculateIPDInput.kernel
BrianMarre: done

include/picongpu/param/speciesAttributes.param
BrianMarre: well it was both, fixed

include/picongpu/simulation/stage/AtomicPhysics.hpp
BrianMarre: outdated, entire section has been refactored.

include/picongpu/simulation/stage/AtomicPhysics.hpp
BrianMarre: outdated, entire section has been refactored.

include/picongpu/simulation/stage/AtomicPhysics.hpp
BrianMarre: outdated, entire section has been refactored.

BrianMarre:
I checked that 94f1aab still fulfills the atomicPhysicsCI test

psychocoderHPC (approved):

psychocoderHPC:
@BrianMarre please add (seperate PR) the svg or source data of your big plot you added as png. The png can not be editied.

PR #4560, opened 2023-05-09
Workaround nvcc 12.1 compile error

bernhardmgruber:


include/pmacc/particles/ParticlesBase.hpp
```diff
@@ -45,9 +45,10 @@ namespace pmacc
     class ParticlesBase : public SimulationFieldHelper<T_MappingDesc>
     {
         using ParticleDescription = T_ParticleDescription;
+        
+    public:
         using MappingDesc = T_MappingDesc;
 
-    public:
```
bernhardmgruber: This is a bit unfortunate here. The base class `SimulationFieldHelper<T_MappingDesc>` already has a public `MappingDesc`. But because we declare the same member again in the derived class `ParticlesBase` but private, it is not accessible from outside.

psychocoderHPC (approved):

PR #4562, opened 2023-05-10
Update sphinx and its dependencies

pordyna:
# Update sphinx 
## Changes
* Switched sphinx version from 2.0.1 to 5.2.0
* Updated other dependencies so that they work with the new `sphinx`
* switched from `recommonmark` to `myst-parser`  for parsing markdown files since `recommonmark` was discontinued. `myst-parser` should be a drop-in replacement. 

## Justification
We were pinning a very old version of `sphinx` and `pip` wasn't able to install the dependencies for me anymore. The `sphinx-autoapi` requires newer versions of `sphinx`, most likely the very old version that works with 2.0.1 is not available anymore. 

## Validation
I build the readthedocs page and the doxygen page, both look good to me, but I wasn't able to look at all pages ;). There are even positive changes since code and doxygen generated headers support now line braking (our "dependencies" and e.g. "important pmacc classes pages" look better on smaller screens / in smaller windows).

Question to @psychocoderHPC, does the CI include these changes automatically, or do we need to force the software update somehow?

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,16 +1,19 @@
-sphinx_rtd_theme>=0.3.1
-recommonmark
-sphinx==2.0.1
-breathe>=4.12.0,<4.15.0
+# restrict up to the next major as suggested in the rtd documentation
+sphinx_rtd_theme<=2.0.0
+# recommonmark was discontinued myst-parser is the new replacement
+myst-parser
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
@pordyna The documentation is built via reading the docs. If I remember correctly ~~say~~ *they* have their own environment.

psychocoderHPC (approved):

PR #4563, opened 2023-05-10
OneSuperCellPosition particle initialization T_PositionFunctor

BrianMarre:
new T_PositionFunctor `OneSuperCellPosition` for the speciesInitialization pipeline,
- spawns all macro particles in at one position in one specified cell of each superCell

useful for super-cell based algorithm degbugging, like FLYonPIC

### DIFF
--- a/include/picongpu/particles/ParticlesInit.kernel
+++ b/include/picongpu/particles/ParticlesInit.kernel
@@ -152,7 +152,7 @@ namespace picongpu
 
                     if(realParticlesPerCell > 0.0_X)
                         numParsPerCell
-                            = posFunctor.template numberOfMacroParticles<ParticleType>(realParticlesPerCell);
+                            = posFunctor.template numberOfMacroParticles<ParticleType>(realParticlesPerCell, cellIdx);
 
                     if(numParsPerCell > 0)
                         kernel::atomicAllExch(
--- a/include/picongpu/particles/startPosition/generic/FreeRng.hpp
+++ b/include/picongpu/particles/startPosition/generic/FreeRng.hpp
@@ -68,9 +68,10 @@ namespace picongpu
                         }
 
                         template<typename T_Particle>
-                        HDINLINE uint32_t numberOfMacroParticles(float_X const realParticlesPerCell)
+                        HDINLINE uint32_t
+                        numberOfMacroParticles(float_X const realParticlesPerCell, DataSpace<simDim> const cellIdx)
                         {
-                            return Functor::template numberOfMacroParticles<T_Particle>(realParticlesPerCell);
+                            return Functor::template numberOfMacroParticles<T_Particle>(realParticlesPerCell, cellIdx);
                         }
 
                     private:
--- a/include/picongpu/param/particle.param
+++ b/include/picongpu/param/particle.param
@@ -1,5 +1,5 @@
-/* Copyright 2013-2022 Axel Huebl, Rene Widera, Benjamin Worpitz,
- *                     Richard Pausch, Klaus Steiniger
+/* Copyright 2013-2023 Axel Huebl, Rene Widera, Benjamin Worpitz,
+ *                     Richard Pausch, Klaus Steiniger, Brian Marre
  *
  * This file is part of PIConGPU.
  *
@@ -18,7 +18,7 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
-/** @file
+/** @file config file for particle manipulators
  *
  * Configurations for particle manipulators. Set up and declare functors that
  * can be used in speciesInitialization.param for particle species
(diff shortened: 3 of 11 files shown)


pordyna:
More general comment:
@BrianMarre I would suggest using a dedicated density profile instead, you can calculate the in supercell position of a cell from the cell index and the supercell size in the functor and return 0 for all other cells in a supercell. This way you don't have to change the position functor interface!

BrianMarre:
I see the reasoning but I would prefer to keep the placement of particles separate from the density profile, similar to the OnePosition functor

psychocoderHPC:
I am not sure if we can introduce this PR into the mainline.
PIConGPU is separating density per cell from particle placement within the cell, this PR is removing the strict separation between both concepts.
In general, I do not understand why someone would like to have one particle per supercell but if it is possible via a density profile.
I will talk to @BrianMarre offline about the motivation maybe I miss something.

include/picongpu/particles/ParticlesInit.kernel
```diff
@@ -152,7 +152,7 @@ namespace picongpu
 
                     if(realParticlesPerCell > 0.0_X)
                         numParsPerCell
-                            = posFunctor.template numberOfMacroParticles<ParticleType>(realParticlesPerCell);
+                            = posFunctor.template numberOfMacroParticles<ParticleType>(realParticlesPerCell, cellIdx);
```
psychocoderHPC, line 155: line 255

```
particle[localCellIdx_] = idx;
```
is avoiding that you can spawn particles outside of the cell the algorithm is selecting.

include/picongpu/particles/startPosition/generic/FreeRng.hpp
```diff
@@ -68,9 +68,10 @@ namespace picongpu
                         }
 
                         template<typename T_Particle>
-                        HDINLINE uint32_t numberOfMacroParticles(float_X const realParticlesPerCell)
+                        HDINLINE uint32_t
+                        numberOfMacroParticles(float_X const realParticlesPerCell, DataSpace<simDim> const cellIdx)
                         {
-                            return Functor::template numberOfMacroParticles<T_Particle>(realParticlesPerCell);
+                            return Functor::template numberOfMacroParticles<T_Particle>(realParticlesPerCell, cellIdx);
```
psychocoderHPC, line 74: This interface change would require an update of all free functors in our examples.
Nevertheless, it introduces the mixing of the separation of density profile and in-cell distribution.

In general, such a change if needed requires a deep review of `ParticlesInit.kernel` to avoid site effects due to the mixed concepts.

psychocoderHPC (changes requested):

BrianMarre:
offline discussion:

- Intent/Motivation clarifications:
   1. **Why do I need one particle per superCell?**
   FLYonPIC works on a per superCell modell, which means that it is helpful for
   debugging of FLYonPIC to have only one particle per superCell.
   Context: FLYonPIC needs large number of ion macro particles for statistics per FLYonPIC cell, to make that feasible I hijacked the superCell structure and do everything on a per superCell level, --> more macro particle samples without exploding the total number of particles.
   2. **Why am I not just handcrafting a density profile?**
   I want to keep the density profile separate from the in superCell macro particle sampling. If I use a handcrafted density profile, this will break this concept as well.
- Implementation discussion:
  **user vs. developer configurable**
  The implementation of the interface of the Free and FreeRng Functors should
  not pass the cellIdx on.
  Reason, the current way we pass the number of physical particles does not
  account for user configure able per superCell spawning of macro particles. My implementation just discards density outside of the spawn cell of each superCell, it therefore artificially reduces the total spawned physical particle density, without the standard user suspecting anything based on setup.
  I therefore agree with @psychocoderHPC 's reasoning that my current implementation has a side effects since it does not account for the missing density in the non spawn cells, this has to be explicitly accounted for by the density profile and therefore this interface should not be user facing to avoid unnoticed/unwanted side effects in user setups.

Summary: We should discuss in the next Dev-Meeting in detail what we can and want to do.

pordyna:
@BrianMarre @psychocoderHPC My suggestion would be to use a density profile. We could have a wrapper class that sets density to 0 for all but one cell in a supercell, and returns cells_in_a_super_cell times the density for the one remaining. So that the overall density remains unchanged. 

I don't think that this breaks our separation of density profiles and in cell positioning, this would be just a very fancy density profile. With a wrapper class, you wouldn't have to rewrite any functors and could use all your already defined profiles.

PrometheusPi:
I agree with @pordyna and would favor a super-cell agnostic approach. 
Is there a reason other than ease-of-use that prevents using the init pipline or density options to get the desired particle distribution?

BrianMarre:
separation of placement and density concepts, the same reason why we do have a start position and a density functor in stead of a unified sampling functor

PrometheusPi:
But with that argument, we could scale this up and decide to have GPU wide placement or simulation box size placement. Where does one draw the line?

BrianMarre:
where the largest physics algorithm element ends, which used to be the cell since the density profile is evaluated only at cell level, but with FLYonPIC we now actually evaluate the density on a superCell level.

BrianMarre:
Result of offline discussion of PIConGPU developers:

This pull request will not be merged.

Reasons:
- superCells are a domain division concept only and we should strive keep them separate from physics concepts such as particle densities. As such passing the in superCell index to a start position functor is not a good idea.
- the PR changes the interface of the free and freeRNG generic particle manipulators breaking the user interface
- there are less invasive approaches to get the same result, for example a sparser or a wrapper of the density profile
- the feature is only useful for debugging, therefore not justifying the considerable work to refactor the interface of initial particle spawning.

PR #4566, opened 2023-05-11
fix crash on AMD (frontier)

### LINKED ISSUE #4561: LWFA example crashes with PQS particle shape and Higuera Cary pusher on Frontier
Hello,
I'm running into errors on Frontier with the `LaserWakefield` example if the Higuera Cary pusher and the PQS particle shape are activated. The error message leads to a simulation crash after the first timestep (0% is included in stdout in a 100 timestep -p 1 simulation) and looks as follows:
```
Memory access fault by GPU node-11 (Agent handle: 0x917be0) on address 0x7ece8ca0d000. Reason: Unknown.
```
The crash is reproducible. 
If only one of them is changed to another pusher or shape, the simulation runs. This here is an overview of the different pusher/shape combinations that I tried out:

|  | Boris | Higuera Cary | Vay |
| ---- | ---- | ---- | ---- |
| TSC | runs | runs | ? |
| PQS | runs | crashes | runs |
| PCS | ? | runs | ? |

I'm using the current dev branch 366092b751561ab71ebd8d7a2abc909d11f629aa with the Frontier profile from `etc/picongpu/frontier-ornl`. Nothing else about the param-files from the `LaserWakefield`-example was changed. For the job submission I use the `8.cfg`-file with 100 timesteps, `-p 1` and **deactivated plugins**. 

I found this crash by trying to reproduce a different error message from a different simulation with the `LaserWakefield` example. This different simulation setup crashes after the second timestep (1% is included in stdout in a 100 timestep -p 1 simulation) with
```
:0:rocdevice.cpp            :2673: 590522892746 us: 63581: [tid:0x7fffd5a9f700] Device::callbackQueue aborting with error : HSA_STATUS_ERROR_MEMORY_APERTURE_VIOLATION: The agent attempted to access memory beyond the largest legal address. code: 0x29
```
which is also reproducible. This simulation is a bit different, I could not figure a way out yet, how to generate this exact error message with the `LaserWakefield` example. However, I believe that both error messages are connected. The original simulation also runs with any different pusher-shape combination that I tried.

@PrometheusPi suggested to add `constexpr size_t reservedGpuMemorySize = 16llu * 1024 * 1024 * 1024;` to `memory.param`, but this has no effect on the crash.

@steindev @psychocoderHPC

psychocoderHPC:
fix #4561

Most likly the compiler is removing the floating point rounding guard in the pusher and therefore the particle position is after the push not in the valid range `[0;1)`.
Validating the range via a a printf if showing no issue, therefore the possibility of an compiler bug is very high.

The current workaround add `volatile` to break the compiler optimization.

### DIFF
--- a/include/picongpu/particles/MoveParticle.hpp
+++ b/include/picongpu/particles/MoveParticle.hpp
@@ -58,7 +58,16 @@ namespace picongpu
              * direction the particle is leaving the cell.
              * The floating point precision is equal for -0.5 and 0.5.
              */
-            floatD_X pos = newPos - 0.5_X;
+#if(BOOST_COMP_HIP)
+            // workaround for a broken HIP optimization
+            // https://github.com/ComputationalRadiationPhysics/picongpu/issues/4561
+            volatile
+#else
+            constexpr
+#endif
+                auto shift
+                = 0.5_X;
+            floatD_X pos = newPos - shift;
 
             DataSpace<simDim> dir;
 
@@ -74,7 +83,7 @@ namespace picongpu
                     moveDir = 1.0_X;
 
                 pos[i] -= moveDir;
-                particle[position_][i] = pos[i] + 0.5_X;
+                particle[position_][i] = pos[i] + shift;
                 dir[i] = precisionCast<int>(moveDir);
             }
 


finnolec (approved):
I reran both the `LaserWakefield` simulation and the original simulation setup that are mentioned in #4561 and both simulations run with this pullrequest included.

PrometheusPi (approved):

Issue #4529, opened 2023-04-12
Plugins: ModuleNotFoundError for typeguard


cbontoiu:
Using a recent dev version of PIConGPU I tried 

```
from picongpu.plugins.data import EnergyHistogramData
from picongpu.plugins.plot_mpl import EnergyHistogramMPL
```
and obtained this errors

![as](https://user-images.githubusercontent.com/7108691/231580845-715ee0fe-a7fc-4d99-bd72-d0d8da56e027.png)

These two imports work when using the plugins found in ```/lib/python/picongpu``` of older versions.

BrianMarre:
your python package install lacks the typeguard library, call ˋpip install -r requirements.txtˋ in your picongpu repository

PrometheusPi:
Or alternatively, change the `sys.path` manually.

PrometheusPi:
@cbontoiu Did @BrianMarre solution fixed your problem. 
We restructured our python library a while ago, thus the python path needs to be set differently now.

cbontoiu:
@PrometheusPi I am dealing with installation issues at the moment but I should give you feedback soon. Thanks.

cbontoiu:
Two things:
**1.** The imports work now, but something else fails. For this code
```
energy = []; 
M = 0;
for itNo in range(len(ITER)):    
    if M >= len(energy):
        counts, bins, _, _ = eh_data.get(species = 'e', species_filter = 'all', iteration = ITER[itNo]); 
        energy.append(counts);
``` 
I get this error:
```
File ~/LASER3DEFFDENSCNT/visual/picongpu/plugins/data/base_reader.py:116, in DataReader.get(self, *args, **kwargs)
    112         iteration = self.find_time.get_iteration(
    113             time, method='closest')
    114     # print("got 'time'=", time, ", converted to iter", iteration)
--> 116 return self._get_for_iteration(iteration, *args, **kwargs)

File ~/LASER3DEFFDENSCNT/visual/picongpu/plugins/data/energy_histogram.py:133, in EnergyHistogramData._get_for_iteration(self, iteration, species, species_filter, include_overflow, **kwargs)
    103 """
    104 Get a histogram for a given iteration.
    105 
   (...)
    130     the timestep between consecutive iterations
    131 """
    132 if iteration is not None:
--> 133     if not isinstance(iteration, collections.Iterable):
    134         iteration = np.array([iteration])
    136 data_file_path = self.get_data_path(species, species_filter)

[...]

PR #4567, opened 2023-05-12
Fix errors in python plugin analysis scripts

finnolec:
This pull request first removes the SliceFieldReader from the python analysis script, the plugin was removed in #4112, this was probably just forgotten. 

Additionally `collections.Iterable` moved to `collections.abc.Iterable` in python 3.3. The second commit should fix this for all the plugin analysis scripts that require `Iterable`, as suggested by @cbontoiu. 

I tested it with python 3.9 (where the old `collections.Iterable` still works) and with python 3.10 (where the previous approach fails) and it works in both cases.

This should solve #4529

### DIFF
--- a/lib/python/picongpu/extra/input/parameters.py
+++ b/lib/python/picongpu/extra/input/parameters.py
@@ -9,6 +9,11 @@ import collections
 import pint
 ureg = pint.UnitRegistry()
 
+try:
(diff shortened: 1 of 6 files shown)


PrometheusPi (approved):
looks good to me

PR #4575, opened 2023-05-22
add Tests for PyPIConGPU

SimeonEhrig:
PICMI and PyPIConGPU CI self tests

# workflow

The CI allows to test PICMI and PyPIConGPU with different Python and dependency versions. If a new CI run is triggered, the following steps are executed for the tests.

1. Run job generator
  - Generates a full test matrix of python and dependencies versions (in the generator script, it is defined if a dependency is tested only with the latest supported version or different versions).
  - Generates for each combination a job where variables defines, which python version should be tested and which dependency should be fixed to a specific version.
 2. Prepare the test
  - Setup a python environment.
  - Takes the `requirments.txt` of PyPIConGPU and generates a new `requirments.txt` with fix dependency versions to a specific depending on the job variables.
 3. Run the quick test

# Following up work

In the following up PR, PyPIConGPU compile test will be implemented. They need more preparation for the environment and a test run takes more time. Therefore special handling is required.

### DIFF
--- a/share/ci/pypicongpu_generator.py
+++ b/share/ci/pypicongpu_generator.py
@@ -1,0 +1,322 @@
+import requests
+import sys
+import yaml
+import re
+from typing import List, Dict, Callable
+import pkg_resources
+
+"""
+This file is part of the PIConGPU.
+Copyright 2023 PIConGPU contributors
+Authors: Simeon Ehrig
+License: GPLv3+
+"""
+
+"""@file Generate different CI test jobs for different Python version and
+         depending on a requirement.txt.
+
+Prints yaml code for a GitLab CI child pipeline to stdout. The test parameters
+are split in two kinds of inputs. The Python versions to test are defined in
+the script, also the names of dependencies to test and it's test strategy. The
+version range of the dependencies to test are defined in a requirements.txt.
+The path of the requirements.txt is set via first application argument.
+
+First, the script reads the requirements.txt. If a dependency is marked to test
+in the script, it calculates the test versions. Therefore it downloads all
+available versions from pypi.org for each package. Afterwards it filters the
+versions via a filter strategy. For example, take all release versions or take
+each latest major version. Than the script removes all versions, which are not
+supported, defined in the requirements.txt. The result a complete list of all
+Python and dependencies versions to test.
+
+In the second part, the script creates the full combination matrix for all test
+versions and creates a CI job for each combination. Each job is printed on
+stdout.
+
+The number of combinations depends on:
+- number of supported Python version
+- number of dependencies to test
+- the test strategy of each dependencies to test
+- versions restrictions in the requirements.txt
+- releases of the dependencies
+
+@param First application argument: Path to the requirements.txt
+"""
+
+
+# caches the parsed dependencies of the requirements.txt here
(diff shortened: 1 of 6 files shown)


share/ci/pypicongpu_generator.py
```diff
@@ -0,0 +1,319 @@
...
+This file is part of the PIConGPU.
+Copyright 2023 PIConGPU contributors
+Authors: Simeon Ehrig
+License: GPLv3+
+"""
+
+"""@file Generate different CI test jobs for different Python version and
+         depending on a requirement.txt.
+
+Prints yaml code for a GitLab CI child pipeline to stout. The test parameter
```
BrianMarre: stout --> stdout

BrianMarre: parameter --> parameters

share/ci/pypicongpu_generator.py
```diff
@@ -0,0 +1,319 @@
...
+
+    output = ""
+    if color == "Red":
+        output += "\033[0;31m"
+    elif color == "Green":
+        output += "\033[0;32m"
+    elif color == "Yellow":
+        output += "\033[1;33m"
+
+    return output + text + "\033[0m"
```
BrianMarre: this should probably become an imported library function in a separate file, code duplication is bad

share/ci/pypicongpu_generator.py
```diff
@@ -0,0 +1,319 @@
...
+    """Returns all release versions of a package registered on pypi.org
+
+    Parameters
+    ----------
+        package_name (str): Name of the searched package.
+
+    Returns
+    -------
+        List[str]: List of release versions.
+    """
```
BrianMarre: is there a reason why you are sometimes using `@param` and some times not?

BrianMarre (approved):
nothing mayor, just a few grammar corrections and one possible code quality improvement

BrianMarre:
@psychocoderHPC ready for your review

psychocoderHPC (dismissed):

psychocoderHPC:
I would wait with merging for the fixes of the typos.

share/ci/pypicongpu_generator.py
SimeonEhrig: For normal software, I would agree. But that are standalone scripts. This scripts should be easily moveable, which is not possible if a dependency between two files exist. In our case we would already have the "complicated" situation, that `pypicongpu_generator.py` and `requirements_txt_modifier.py` shares the same code, but they are not located in the same folder.

To solve the problem correctly, we would need an installation script, which solves the problem. IMO, this is much more work, than keep the copied code persistent.

share/ci/pypicongpu_generator.py
SimeonEhrig: I copied the doc string style of the functions from other python scripts. Therefore it is inconsistent. I can fix it.

share/ci/pypicongpu_generator.py
SimeonEhrig: @BrianMarre and I agreed in a personal discussion that my approach is fine for this use case.

psychocoderHPC (approved):

SimeonEhrig:
@psychocoderHPC Ready to merge

PR #4582, opened 2023-06-01
fix 2D current deposition performance

psychocoderHPC:
EZ and Esirkepov current deposition used the cached/jit shape type of the outer loop for the inner loop.
This results in lower performance because the Cached shape which is the default for the outer loop requires more calculations and results in more thread branch differences.

This PR uses the correct shape type for the inner (fast indexing) loop.

### DIFF
--- a/include/picongpu/fields/currentDeposition/EmZ/DepositCurrent.hpp
+++ b/include/picongpu/fields/currentDeposition/EmZ/DepositCurrent.hpp
@@ -173,8 +173,8 @@ namespace picongpu
                         return;
 
                     auto const shapeI = makeTrajectoryAssignmentShapeFunction(
-                        typename T_Strategy::template ShapeOuterLoop<T_ParticleAssign>{line.m_pos0[0], true},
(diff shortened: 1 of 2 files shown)


PrometheusPi:
Code change looks good to me - I will test whether this fixes my 2D foil sims.

PrometheusPi:
@psychocoderHPC I am still confused, does this mean, the result of old and new code will be the same and this is just an optimization?

psychocoderHPC:
> @psychocoderHPC I am still confused, does this mean, the result of old and new code will be the same and this is just an optimization?

Yes the result should be equal but the performance of the new code should be better.

PrometheusPi (dismissed):
locks good to me.
A first test showed (as expected) no major difference in run time. 
I see slight differences in energy - but these are minimal and might be due to a different order of addition. 

I will not merge this pull request for now due to the conflict it causes with #4583

PrometheusPi:
As mentioned *offline* their is a merge conflict due to #4583. 
Please rebase @psychocoderHPC.

psychocoderHPC:
@PrometheusPi I rebased the PR

PrometheusPi (approved):

PR #4583, opened 2023-06-01
Esirekpov 2D use bitmask helper

psychocoderHPC:
With #4170 we missed to use the bitmask helper class instead we do inplace bit magic whihc leads to possible inconsitencies in cases the bitmask will be changed.
This PR using the bitmask helper clas everywhere now.


PrometheusPi (approved):
locks good to me - same as 3D case

PR #4584, opened 2023-06-01
fix Esirkepov current deposition 2D

### LINKED ISSUE #4581: Problem in Esirkepov CurrentSolver
The total energy is diverging for the (almost) vanilla FoilLCT-like example, the issue seems to be in the particleCurrentSolver, while Esirkepov shows diverging energy after a while, EmZ stays bound: 
image

psychocoderHPC:
fix #4581

With #4170 [broken line](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4170/files#diff-525c7dc809f62cebe12cfce4484e98b121bacfceb5c9179f3a41336450d402edL162) we introduced a bug in the 2D implementation of the 2D Esirkepov implementation.
The symptom of this bug is that global energy increases. We observed this issue with high-density target simulation, never the less even in low-densed physics the bug is there.

CC-ing: @HighIander

### DIFF
--- a/include/picongpu/fields/currentDeposition/Esirkepov/Esirkepov2D.hpp
+++ b/include/picongpu/fields/currentDeposition/Esirkepov/Esirkepov2D.hpp
@@ -181,7 +181,7 @@ namespace picongpu
                     if(j < end + bitpacking::getValue(parStatus[1], bitpacking::Status::LEAVE_CELL))
                     {
                         const float_X s0j = shapeJ.S0(j);
-                        const float_X dsj = shapeJ.S0(j) - s0j;
+                        const float_X dsj = shapeJ.S1(j) - s0j;
 
                         float_X tmp = -currentSurfaceDensity * (s0j + 0.5_X * dsj);
 


psychocoderHPC:
From our internal chart system, I copied @HighIander's bug report

> Regarding the problem of diverging energy for the (almost) vanilla FoilLCT-like example case in PIConGPU, the issue seems to be in the particleCurrentSolver, while Esirkepov shows diverging energy after a while, EmZ stays bound: 

image

> Interestingly, when using too few particles/resolution that introduces (nonphysical) noise, both solvers work, which might have hidden the issue earlier when there were maybe rounding errors etc. elsewhere in the code.

image


[note from @psychocoderHPC] The test cases are 2D simulations.

psychocoderHPC:
@steindev @PrometheusPi please merge this PR it is a bug fix and important e.g for @HighIander

PrometheusPi:
I just tested this pull request together with #4583. Looks good. (The setup is periodic, thus the laser will nit vanish.)

old (wrong / total energy increase):
grafik

new (no energy increase):
grafik

PrometheusPi (approved):

PR #4585, opened 2023-06-01
Range specification from --openPMD.toml

franzpoeschel:
The parameter `--openPMD.range` was the only way to specify ranges so far.
Since the parameter specification is a bit all over the place right now in the openPMD plugin, this PR tries to take the chance for cleaning things up a bit. Parameter specification is now mainly moved into the `openPMDWriter::Help` class.

TODO

- [x] Cleanup
- [x] Testing

### DIFF
--- a/include/picongpu/plugins/openPMD/Parameters.hpp
+++ b/include/picongpu/plugins/openPMD/Parameters.hpp
@@ -5,7 +5,7 @@
 
 namespace picongpu::openPMD
 {
-    struct PluginOptions
+    struct PluginParameters
     {
         std::string fileName = "simData";
         std::string fileInfix = "_%06T";
--- a/include/picongpu/plugins/openPMD/Parameters.hpp
+++ b/include/picongpu/plugins/openPMD/Parameters.hpp
@@ -16,9 +16,17 @@ namespace picongpu::openPMD
         std::string jsonRestartParams = "{}";
 
         // for using this with std::tie()
-        operator std::tuple<std::string&, std::string&, std::string&, std::string&, std::string&, std::string&, std::string&>()
+        operator std::
+            tuple<std::string&, std::string&, std::string&, std::string&, std::string&, std::string&, std::string&>()
         {
-            return std::tuple<std::string&, std::string&, std::string&, std::string&, std::string&, std::string&, std::string&>{
+            return std::tuple<
+                std::string&,
+                std::string&,
+                std::string&,
+                std::string&,
+                std::string&,
+                std::string&,
+                std::string&>{
                 fileName,
                 fileInfix,
                 fileExtension,
--- a/include/picongpu/plugins/openPMD/Parameters.hpp
+++ b/include/picongpu/plugins/openPMD/Parameters.hpp
@@ -13,17 +13,19 @@ namespace picongpu::openPMD
         std::string dataPreparationStrategy = "doubleBuffer";
         std::string jsonConfig = "{}";
         std::string range = ":,:,:";
+        std::string jsonRestartParams = "{}";
 
         // for using this with std::tie()
-        operator std::tuple<std::string&, std::string&, std::string&, std::string&, std::string&, std::string&>()
+        operator std::tuple<std::string&, std::string&, std::string&, std::string&, std::string&, std::string&, std::string&>()
         {
-            return std::tuple<std::string&, std::string&, std::string&, std::string&, std::string&, std::string&>{
+            return std::tuple<std::string&, std::string&, std::string&, std::string&, std::string&, std::string&, std::string&>{
                 fileName,
                 fileInfix,
                 fileExtension,
                 dataPreparationStrategy,
                 jsonConfig,
-                range};
+                range,
+                jsonRestartParams};
         }
     };
 } // namespace picongpu::openPMD
--- a/include/picongpu/plugins/openPMD/Parameters.hpp
+++ b/include/picongpu/plugins/openPMD/Parameters.hpp
@@ -1,0 +1,29 @@
+#pragma once
+
+#include <string>
+#include <tuple>
+
+namespace picongpu::openPMD
+{
+    struct PluginOptions
+    {
+        std::string fileName = "simData";
+        std::string fileInfix = "_%06T";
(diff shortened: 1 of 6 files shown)


include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -386,6 +484,42 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
...
+                    return tomlDataSources.value().openPMDPluginOptions;
+                }
+                else
+                {
+                    PluginOptions res;
+                    for(auto const& param : parameters)
+                    {
+                        auto const& cmd_param = param.commandLineOption;
+                        auto const& target = param.targetInConfigObject;
+                        if(!target.has_value() || !cmd_param->optionDefined(id))
```
franzpoeschel, line 547: Please review if the `!cmd_param->optionDefined(id)` condition is correct, especially when a default value is specified and not overwritten. This was so far only checked for the range option.

include/picongpu/plugins/openPMD/Parameters.hpp
```diff
@@ -0,0 +1,31 @@
...
+                fileName,
+                fileInfix,
+                fileExtension,
+                dataPreparationStrategy,
+                jsonConfig,
+                range,
+                jsonRestartParams};
+        }
+    };
+} // namespace picongpu::openPMD
```
psychocoderHPC: missing EOF

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -343,15 +447,9 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                     auto res = tomlSources.size() > notifyPeriod.size() ? tomlSources.size() : notifyPeriod.size();
                     if(res == 0)
                     {
-                        std::vector<plugins::multi::Option<std::string>> theseMustBeEmpty{
```
psychocoderHPC, line 346: I do not get the logic of how you decide in the refactoring which parameters must be empty. Could you describe how it is now working?

psychocoderHPC (changes requested):

psychocoderHPC:
Please run clang-format so that the CI can check the changes.

include/picongpu/plugins/openPMD/openPMDWriter.def
```diff
@@ -97,8 +91,6 @@ namespace picongpu
             std::unique_ptr<AbstractJsonMatcher> jsonMatcher;
             std::string jsonRestartParams;
```
psychocoderHPC: IMO this is not required anymore an could lead to the wrong usage of this variable instead of the value from the ThreadParameters.

include/picongpu/plugins/openPMD/openPMDWriter.hpp
franzpoeschel: All sub-parameters of the openPMD plugins (i.e. those that are not `--openPMD.period` and `--openPMD.toml`) are now listed in `std::vector<Parameter> openPMDWriter::Help::parameters`. All of those are checked for being empty.

include/picongpu/plugins/openPMD/openPMDWriter.def
franzpoeschel: Agreed, I will push a commit that merges `PluginParameters` and `ThreadParams`

include/picongpu/plugins/openPMD/Parameters.hpp
```diff
@@ -7,33 +7,12 @@ namespace picongpu::openPMD
 {
     struct PluginParameters
     {
-        std::string fileName = "simData";
+        std::string fileName = "simData"; /* Name of the openPMDSeries, excluding the extension */
```
psychocoderHPC, line 10: note: You should best comment above the line. The reason is that clang format could move the comment into the next line if the line is too long, maybe not today but at least if someone is changing e.g. the nested namespaces, ....
If the comment is in the next line you do not know anymore that it corresponds to the line above.

No need to change it in this PR, it is more of a general comment.

psychocoderHPC (approved):

include/picongpu/plugins/openPMD/Parameters.hpp
franzpoeschel: I normally do that, I just copied those comments here over from the deleted lines in `ThreadParams` and they were like that

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -244,6 +248,111 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
...
+                std::optional<std::string PluginParameters::*> targetInConfigObject;
+                // See description of enum class ApplyParameter
+                ApplyParameter applyParameter;
+                // Optionally the additionalDescription parameter of
+                // commandLineParameter->registerHelp()
+                std::optional<std::function<std::string()>> additionalDescription;
+
+                Parameter(
+                    plugins::multi::Option<std::string>* commandLineParameter_in,
+                    std::optional<std::string> tomlParameter_in,
```
psychocoderHPC, line 290: IMO there is no need for this parameter because the value can be derived with `commandLineParameter_in->getName()` in line 295 and would avoid that we use different names in toml and on the command line.

psychocoderHPC (changes requested):

include/picongpu/plugins/openPMD/openPMDWriter.hpp
franzpoeschel: The parameter names are not exactly the same. I used `backend_config` instead of `.json` in the TOML configuration.
`--openPMD.json` is still somewhat descriptive, but `json = "..."` in a TOML file was a bit weird. 
Also for some reason TOML uses `data_preparation_strategy`, but the command line parameter is `--openPMD.dataPreparationStrategy`.
We could try unifying this, but there would be API breaking.

psychocoderHPC (approved):

psychocoderHPC (approved):

PR #4586, opened 2023-06-01
Use SFINAE for determining if storeChunkRaw should be used

franzpoeschel:
The Hemera installation of openPMD-api uses the dev branch, so the version macro does not work correctly.
--> We need to be a bit more sneaky


psychocoderHPC (approved):
Thanks for the quick fix :heart:

PR #4134, opened 2022-05-31
[FLYonPIC] atomic physics for PIConGPU

BrianMarre:
First experimental implementation of atomic physics for PIConGPU.

This implementation is not ready for use,
- not performant, several known bottlenecks
- only one ion species may have atomic Physics
- ionization from exited states still missing
- atomic input data is loaded every timeStep
- nearly no documentation
- lacking configurability for several algorithms
- ...

PR #4592, opened 2023-06-09
pull charge state particle manipulator into code

BrianMarre:
Pulls the setChargeState functors from the `particle.param` into the code itself.

Up to now all PIConGPU setups with ionizers had their only private copy of a set charge state particle manipulator. This creates code duplication and maintenance issues as changes did not propagate to user setups making any changes error prone. This is especially problematic for FLYonPIC(atomicPhysics) which requires the charge state to be consistent with the atomic state and therefore required changes to the initialization of charge states.

Old setups are still valid and compatible, but I strongly encourage all users to switch to the new in code implementation for all future setups, this increases resilience with additional checks and makes the setups guaranteed compatible
 with FLYonPIC.

new manipulator is used as follows:
```C++
[particle.param]
namespace manipulators
{
     using SetChargeState = unary::ChargeState<[ChargeState you want]>;
} // namespace manipulators
```

to update an old setup replace your set charge state implementation in `particle.param`(usually), looking similar to this
```C++
[particle.param]
namespace picongpu::manipulators
{
    /** Define initial ion charge */
    struct SetIonTo16Ionized
    {
        /** set ion bound electrons to yield targetChargeState.
         */
        template<typename T_Particle>
        DINLINE void operator()(T_Particle& particle)
        {
            constexpr float_X targetChargeState = 16;

            // number of Electrons for neutral atom
            constexpr float_X numberBoundElectrons
                = picongpu::traits::GetAtomicNumbers<T_Particle>::type::numberOfProtons;

            PMACC_ASSERT_MSG(
                numberBoundElectrons >= targetChargeState,
                "too few protons for target charge state");

            picongpu::particles::atomicPhysics::SetToAtomicGroundStateForChargeState{}(
                particle,
                numberBoundElectrons - targetChargeState);
        }
    };

    /** Definition of manipulator that sets properties of an ion macro-particle
     * according properties defined in struct SetIonTo16Ionized.
     */
    using SetIonization = generic::Free<SetIonTo16Ionized>;
} // namespace picongpu::particles::manipulators
```
or in older setups, like this
```C++
[particle.param]
namespace picongpu::particles::manipulators
{
    /** Define initial charge state of ion macro-particel species. */
    struct SetIonToNeutral
    {
        /** Set number of bound electrons to number of protons
         * to create neutral atoms at initialization.
         */
        template<typename T_Particle>
        DINLINE void operator()(T_Particle& particle)
        {
            using Particle = T_Particle;

            // number of bound electrons at initialization state of the neutral atom
            float_X const protonNumber = picongpu::traits::GetAtomicNumbers<T_Particle>::type::numberOfProtons;

            particle[boundElectrons_] = protonNumber;
        }
    };
    /** Definition of manipulator that sets boundElectrons attribute an ion macro-particle
     * according properties defined in struct SetIonToNeutral.
     */
    using SetBoundElectrons = generic::Free<SetIonToNeutral>;
} // namespace picongpu::particles::manipulators
```
with this
```C++
namespace picongpu::particles::manipulators
{
     using SetIonization = unary::ChargeState<16u>;
} // namespace picongpu::particles::manipulators
```
or for the second example this
```C++
namespace picongpu::particles::manipulators
{
    using SetBoundElectrons = unary::ChargeState<0u>;
} // namespace picongpu::particles::manipulators
```

### DIFF
--- a/include/picongpu/particles/atomicPhysics/SetToAtomicGroundStateForChargeState.hpp
+++ b/include/picongpu/particles/atomicPhysics/SetToAtomicGroundStateForChargeState.hpp
@@ -29,66 +29,56 @@ namespace picongpu
         {
             struct SetToAtomicGroundStateForChargeState
             {
-                template<typename T_Particle>
-                DINLINE void operator()(T_Particle& particle, float_X numberBoundElectrons)
+                template<typename T_Ion>
+                DINLINE void operator()(T_Ion& ion, uint8_t numberBoundElectrons)
                 {
                     /** set a given ion to its ground state for a given number of bound
                      *  electrons
                      *
-                     *  Functor, sets a given particle's boundElectrons attribute,
-                     *  and, if present, atomicState attribute, to the correct value/s
-                     *  for its ground state.
+                     * Functor, sets, if present, the atomicConfigNumber(i.e. atomicState) attribute,
+                     *   to the ground state.
                      *
                      *  Example:
                      *    Ar(Z=18), boundElectrons=3 --> super configuration(2,1,0,0,...)
                      *
-                     *  BEWARE: Different Usage from previous SetIonization(),
-                     *      specify NUMBER of bound electrons NOT charge state.
-                     *  BEWARE: Uses simple fill from bottom shell wise for ground state
-                     *      determination, not actually entire truth.
-                     *  @todo : implement full madelung occupation schema, and exceptions,
-                     *      like Cu.
+                     * @attention uses simple fill from bottom shell wise for ground state
+                     *      determination, may different from input data set ground state.
+                     *  @todo : implement searching for ground state in atomic input data
                      */
 
                     PMACC_DEVICE_ASSERT_MSG(numberBoundElectrons >= 0, "Number of bound electrons must be >= 0");
 
-                    using Particle = T_Particle;
-
-                    particle[boundElectrons_] = numberBoundElectrons;
-
                     if constexpr(pmacc::traits::HasFlag<
-                                     typename T_Particle::FrameType,
+                                     typename T_Ion::FrameType,
                                      atomicPhysicsSolver<>>::type::value) // -> atomicState exists
                     {
                         // get current Configuration number object
-                        auto configNumber = particle[atomicConfigNumber_];
+                        auto configNumber = ion[atomicConfigNumber_];
 
                         // create blank occupation number vector
                         auto occupationNumberVector
                             = pmacc::math::Vector<uint8_t, configNumber.numberLevels>::create(0);
                         // could actually be reduced to uint8_t since Z<=98(Californium) for our purposes
 
-                        uint8_t numberElectronsRemaining = static_cast<uint8_t>(numberBoundElectrons);
-
                         // fill from bottom up until no electrons remaining -> ground state init
                         /// @todo : implement Mandelung Schema and exceptions, Brian Marre, 2022
                         for(uint8_t level = 1u; level <= configNumber.numberLevels; level++)
                         {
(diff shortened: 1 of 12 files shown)


BrianMarre:
@psychocoderHPC CI crashed :(

include/picongpu/particles/atomicPhysics/SetToAtomicGroundStateForChargeState.hpp
```diff
@@ -29,13 +29,13 @@ namespace picongpu
         {
             struct SetToAtomicGroundStateForChargeState
             {
-                template<typename T_Particle>
-                DINLINE void operator()(T_Particle& particle, float_X numberBoundElectrons)
+                template<typename T_Ion>
+                DINLINE void operator()(T_Ion& ion, float_X const numberBoundElectrons)
```
psychocoderHPC: should be to make the CI happy

```C++
DINLINE void operator()(T_Ion& ion, [[maybe_unused]] float_X const numberBoundElectrons) 
```

include/picongpu/particles/atomicPhysics/SetToAtomicGroundStateForChargeState.hpp
```diff
@@ -52,16 +52,12 @@ namespace picongpu
 
                     PMACC_DEVICE_ASSERT_MSG(numberBoundElectrons >= 0, "Number of bound electrons must be >= 0");
 
-                    using Particle = T_Particle;
-
-                    particle[boundElectrons_] = numberBoundElectrons;
```
psychocoderHPC, line 57: Is there no need to set `particle[boundElectrons_]` anymore?

psychocoderHPC (changes requested):

include/picongpu/particles/atomicPhysics/SetToAtomicGroundStateForChargeState.hpp
BrianMarre: done

include/picongpu/particles/atomicPhysics/SetToAtomicGroundStateForChargeState.hpp
BrianMarre: is done in the ChargeState manipulator itself, line 57 ChargeState.hpp, to keep atomicPhysics and charge state separate.

psychocoderHPC (approved):

include/picongpu/particles/atomicPhysics/SetToAtomicGroundStateForChargeState.hpp
psychocoderHPC: @BrianMarre Since @pordyna reported that this PR is the reason for a deadlock in the simulation this question is showing up again.
In an old PR bound electrons were manipulated in the ionization implementation. Later it was removed by calling `SetToAtomicGroundStateForChargeState` 

https://github.com/ComputationalRadiationPhysics/picongpu/pull/4134/files#diff-f536e1a1139f8a19bc17a44e02c62eb2a728cb75b9373b23bcf061c0986966a3R322


Now you wrote that bound electrons will be set in `ChargeState.hpp` but the problem is that the ionization implementations does not call this functor. Therefore IMO `boundElectrons_` will never be updated.

psychocoderHPC: I opened issue #4611 to document the debugging

BrianMarre: You are correct, I missed that side effect, sorry :(

PR #4600, opened 2023-06-13
PyPIConGPU compile test

SimeonEhrig:
Enable PyPIConGPU compile tests in the CI.

We test the compiling tests only with one configuration. This should be enough, because param files should be independent of the enabled backends.

Before this PR can be merged, the following PR can be merged:
- [x] #4575 
- [x] #4601

PR #4418, opened 2023-01-16
Refactor doc-strings in particle.param

steindev:
Rephrase for consistency between manipulators and
expand some of the comments for clarification of
the purpose of a specific manipulator.

PR #4601, opened 2023-06-13
fix pypicongpu error after PR #4418

BrianMarre:
PR #4418 broke PyPIConGPU, by replacing the hard coded typical ppc in particle.param with a dynamic ppc take from a simple density operation.

This change broke the particle.param template, `/share/picongpu/pypicongpu/template/include/picongpu/param/particle.param.mustache`, causing an incomplete name to be generated as the referenced quantity was only defined inside the context of a simple density operation but was used outside, leading to a mustache template generation miss and consequently the template place holder being replaced with an emtpy string, resulting in a not defined c++ name.

This PR fixes this bug by, adding an automatic fallback calculation of the typical_ppc from the set of all density generating operations and a PIConGPU specific parameter in the PICMI simulation object for specifying the typical_ppc of a setup, allowing a user to override the default behavior.

This bug was noticed by @SimeonEhrig while writing the PR #4600.

### DIFF
--- a/lib/python/picongpu/picmi/simulation.py
+++ b/lib/python/picongpu/picmi/simulation.py
@@ -105,6 +105,7 @@ class Simulation(picmistandard.PICMI_Simulation):
     def __init__(self,
                  picongpu_template_dir: typing.Optional[
                      typing.Union[str, pathlib.Path]] = None,
+                 picongpu_typical_ppc: typing.Optional[int] = None,
                  **kw):
         # delegate actual work to parent
         super().__init__(**kw)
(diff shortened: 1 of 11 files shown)


lib/python/picongpu/picmi/simulation.py
```diff
@@ -481,6 +484,17 @@ def get_as_pypicongpu(self) -> simulation.Simulation:
         self.__resolve_electrons()
 
         s.init_manager = self.__get_init_manager()
+
+        init_mng = s.init_manager
```
steindev: Is it used?

BrianMarre: left over from debugging, removed, thanks!

steindev (approved):

BrianMarre:
@steindev merge ready

PR #4606, opened 2023-06-19
momentum density output

### LINKED ISSUE #3996: Temperature map
Hi
I have a question about temperature profile analysis 

Is there is the way to create temperature map for system?
So that I can analyse how laser heats plasma
___

psychocoderHPC:
fix #3996

Currently, it is not possible to use `deriveField::derivedAttributes::MomentumDensity<...>` twice or more in an output because the name in the output is not unique and therefore overwritten and only the last configuration from `fileOutput.param` is visible.

This PR is 
- giving each component of `MomentumDensity` a unique name
- the naming is changed to `momentumDensity/<x,y,z>`

### DIFF
--- a/include/picongpu/particles/particleToGrid/derivedAttributes/MomentumDensity.def
+++ b/include/picongpu/particles/particleToGrid/derivedAttributes/MomentumDensity.def
@@ -22,6 +22,7 @@
 #include "picongpu/simulation_defines.hpp"
 
 #include "picongpu/particles/traits/SpeciesEligibleForSolver.hpp"
+#include "picongpu/plugins/misc/ComponentNames.hpp"
 #include "picongpu/traits/SIBaseUnits.hpp"
 
 #include <pmacc/static_assert.hpp>
@@ -75,7 +76,8 @@ namespace picongpu
                     //! Get text name
                     HINLINE static std::string getName()
                     {
-                        return "particleMomentumDensity";
+                        auto const componentNames = plugins::misc::getComponentNames(3);
+                        return "momentumDensity/" + componentNames[T_direction];
                     }
 
                     /** Calculate value of the derived attribute per particle


include/picongpu/particles/particleToGrid/derivedAttributes/MomentumDensity.def
```diff
@@ -75,7 +77,8 @@ namespace picongpu
                     //! Get text name
                     HINLINE static std::string getName()
                     {
-                        return "particleMomentumDensity";
+                        auto const componentNames = plugins::misc::getComponentNames(3);
+                        return "momentumDensity/" + componentNames[T_direction];
```
psychocoderHPC, line 80: The naming is derived from `MidCurrentDensityComponent`

include/picongpu/particles/particleToGrid/derivedAttributes/MomentumDensity.def
```diff
@@ -22,11 +22,13 @@
 #include "picongpu/simulation_defines.hpp"
 
 #include "picongpu/particles/traits/SpeciesEligibleForSolver.hpp"
+#include "picongpu/plugins/misc/ComponentNames.hpp"
 #include "picongpu/traits/SIBaseUnits.hpp"
 
 #include <pmacc/static_assert.hpp>
 #include <pmacc/traits/HasIdentifiers.hpp>
 
+#include <array>
```
jkelling:
```suggestion
```
This include should not be required by the new code. Is this a fix for a co-located inaccuracy?

include/picongpu/particles/particleToGrid/derivedAttributes/MomentumDensity.def
jkelling, line 80: If I am not mistaken, the copy of the string from the vector in the return cannot be elided because the value is accessed by (const) reference from an object, but we can avoid copying the payload by moving.
```suggestion
                        auto componentNames = plugins::misc::getComponentNames(3);
                        return "momentumDensity/" + std::move(componentNames[T_direction]);
```

jkelling (dismissed):

steindev:
@psychocoderHPC Did you test the code? If it works @jkelling's concern would be ruled out and we can merge.

psychocoderHPC:
> @psychocoderHPC Did you test the code? If it works @jkelling's concern would be ruled out and we can merge.

I tested only for compiling. The changes are equal to other older functors.

include/picongpu/particles/particleToGrid/derivedAttributes/MomentumDensity.def
psychocoderHPC: fixed

include/picongpu/particles/particleToGrid/derivedAttributes/MomentumDensity.def
psychocoderHPC: There is no reference in the code because we take the name by copy. `auto const componentNames = ...`

jkelling: `componentNames[T_direction]` is a const reference to an element of the vector `componentNames`.

jkelling:
> @psychocoderHPC Did you test the code? If it works @jkelling's concern would be ruled out and we can merge.

My comment is not about the code not working, only about avoiding an unnecessary copy of the buffer of an `std::string`.

include/picongpu/particles/particleToGrid/derivedAttributes/MomentumDensity.def
jkelling: The suggested change is intended to make the compiler use an rvalue overload of `operator+` ([here, overload (11)](https://en.cppreference.com/w/cpp/string/basic_string/operator%2B)).

steindev:
> > @psychocoderHPC Did you test the code? If it works @jkelling's concern would be ruled out and we can merge.
> 
> My comment is not about the code not working, only about avoiding an unnecessary copy of the buffer of an `std::string`.

Mhh, I am undecided. I think, if we apply this change, then we should do it everywhere. And this is worth a separate PR.

psychocoderHPC:
> > > @psychocoderHPC Did you test the code? If it works @jkelling's concern would be ruled out and we can merge.
> > 
> > 
> > My comment is not about the code not working, only about avoiding an unnecessary copy of the buffer of an `std::string`.
> 
> Mhh, I am undecided. I think, if we apply this change, then we should do it everywhere. And this is worth a separate PR.

Yes let us change this in a separate PR because we have a few places where this should be changed so we could group the changes.

PrometheusPi (approved):

PrometheusPi:
I would follow @steindev suggestion and avoid copies in a later PR in order fix this issue quickly. 
@jkelling are you fine with that?

PR #4607, opened 2023-06-19
change output naming in `WeightedVelocity`

psychocoderHPC:
Follow the naming of the plugin `MidCurrentDensityComponent`. Change naming from `particleVelocity` to `weightedVelocity/<x,y,z>`.

### DIFF
--- a/include/picongpu/particles/particleToGrid/derivedAttributes/WeightedVelocity.def
+++ b/include/picongpu/particles/particleToGrid/derivedAttributes/WeightedVelocity.def
@@ -22,6 +22,7 @@
 #include "picongpu/simulation_defines.hpp"
 
 #include "picongpu/particles/traits/SpeciesEligibleForSolver.hpp"
+#include "picongpu/plugins/misc/ComponentNames.hpp"
 #include "picongpu/traits/SIBaseUnits.hpp"
 
 #include <pmacc/static_assert.hpp>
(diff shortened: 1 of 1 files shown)


include/picongpu/particles/particleToGrid/derivedAttributes/WeightedVelocity.def
```diff
@@ -76,12 +77,8 @@ namespace picongpu
...
                     HINLINE static std::string getName()
                     {
-                        if constexpr(T_direction == 0)
-                            return "particleVelocityX";
-                        else if constexpr(T_direction == 1)
-                            return "particleVelocityY";
-                        else
-                            return "particleVelocityZ";
+                        auto const componentNames = plugins::misc::getComponentNames(3);
+                        return "weightedVelocity/" + componentNames[T_direction];
```
jkelling, line 81: Try to avoid copy of temporary string?
```suggestion
                        auto componentNames = plugins::misc::getComponentNames(3);
                        return "weightedVelocity/" + std::move(componentNames[T_direction]);
```

psychocoderHPC: IMO without move it is easier to read. I checked it also in compiler explorer and the assembler without move looks better

https://godbolt.org/z/xYaavqeKs

PrometheusPi (approved):
as discussed in #4606 the string copy should be solved in a separate PR 
Thus fine with me as is

PR #4608, opened 2023-06-19
derived output particle momentum

psychocoderHPC:
Rename output name from `particleMomentum` to
`particleMomentum/<x,y,z>`.
Similar to #4606 it is not possible to use this functor with different component twice or more in the output.

### DIFF
--- a/include/picongpu/particles/particleToGrid/derivedAttributes/Momentum.def
+++ b/include/picongpu/particles/particleToGrid/derivedAttributes/Momentum.def
@@ -22,6 +22,7 @@
 #include "picongpu/simulation_defines.hpp"
 
 #include "picongpu/particles/traits/SpeciesEligibleForSolver.hpp"
+#include "picongpu/plugins/misc/ComponentNames.hpp"
 #include "picongpu/traits/SIBaseUnits.hpp"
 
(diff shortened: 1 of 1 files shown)


include/picongpu/particles/particleToGrid/derivedAttributes/Momentum.def
```diff
@@ -75,7 +76,8 @@ namespace picongpu
                     //! Get text name
                     HINLINE static std::string getName()
                     {
-                        return "particleMomentum";
+                        auto const componentNames = plugins::misc::getComponentNames(3);
+                        return "particleMomentum/" + componentNames[T_direction];
```
jkelling, line 80: Maybe we can avoid copying that temporary string.
```suggestion
                        auto componentNames = plugins::misc::getComponentNames(3);
                        return "particleMomentum/" + std::move(componentNames[T_direction]);
```

psychocoderHPC: IMO without move it is easier to read. I checked it also in compiler explorer and the assembler without move looks better

https://godbolt.org/z/xYaavqeKs

PrometheusPi (approved):
see #4606 for discussion on string copy

PR #4614, opened 2023-06-26
Apply recent workarounds that are needed for Streaming on Frontier to .tpl file

franzpoeschel:
This is turning into a catalogue by now

* srun networking parameters
* hints how to compile ADIOS2
* check that at least two nodes are used

### DIFF
--- a/etc/picongpu/crusher-ornl/batch_pipe.tpl
+++ b/etc/picongpu/crusher-ornl/batch_pipe.tpl
@@ -32,6 +32,7 @@
 # when trying to run two sub-jobs asynchronously.
 #SBATCH -S 0
 #SBATCH --exclusive # This one might not be strictly necessarily, haven't tested
+#SBATCH --network=single_node_vni,job_vni
 
 ###################
 # Optional params #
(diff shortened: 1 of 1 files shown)


etc/picongpu/crusher-ornl/batch_pipe.tpl
```diff
@@ -95,6 +96,29 @@ export OMP_NUM_THREADS=!TBG_coresPerGPU
...
+if (( !TBG_nodes < 2 )); then
+    cat << EOF >&2
+Frontier does not reliably initialize its networking capabilities for
+single-node jobs. For a reliable streaming implementation, please adjust the
+.cfg file such that the job uses at least two nodes.
+EOF
+    exit 1
+fi
+
+cat << EOF >&2
```
steindev, line 108: I think this second comment is something we usually put in the system's profile. Would you please do that? Together with a hint that ADIOS2 needs to be self-compiled in order to use SST?

franzpoeschel: Should we really bother every user on Frontier with that note? Putting it in here means that only those that want to use streaming will see it (and also has the advantage that every workaround for streaming on Frontier stays in one file).

psychocoderHPC (approved):

etc/picongpu/crusher-ornl/batch_pipe.tpl
psychocoderHPC: I will merge it as it is, I think only a handful of users will use streaming.

PR #4618, opened 2023-07-03
Fix units of phase space plugin and in documentation

steindev:
When the phase space plugin started to use openPMD output in #3468, the setting of meta data was incorrect. Also in the documentation. This is corrected in this PR.

It also corrects a confusing comment on units in a python example script on phase space data handling/visualization.

### DIFF
--- a/docs/source/usage/plugins/phaseSpace.rst
+++ b/docs/source/usage/plugins/phaseSpace.rst
@@ -63,7 +63,7 @@ The output writes a number of non-standard custom openPMD attributes:
 
 * ``p_min`` and ``p_max``: The lower and upper bounds for the momentum axis, respectively.
 * ``dr``: The spacing of the spatial axis in PIConGPU units.
-* ``dV``: The volume of a phase space cell. Relates to ``dr`` via ``dV = dp * dr`` where ``dp`` would be the grid spacing along the momentum axis.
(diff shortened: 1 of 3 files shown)


PrometheusPi:
*offline discussion with @steindev:* 
The convention we use here in PIConGPU to store charge (real space volume) density in the bins is highly unintuitive. So far this seems to not affect any publications. 

This pull request will fix the wrong metadata, but will not fix the underlying issue of the unintuitive implementation.

franzpoeschel (approved):
Looks good to me

BrianMarre (approved):
looks good for now, though we should rework the wording of the documentation some time in the future.

Currently some important information is mentioned out of context.

PrometheusPi (approved):
We urgently add a second pull request to make the implementation less unintuitive.

PrometheusPi:
@steindev Is there already a followup pull request?

steindev:
No. Actually I wasn't aware that I should provide it. What is your opinion on a more intuitive definition of phase space? I guess the distribution function written in the Vlasov equ. is a common quantity, but 6 dimensional. So for the 2D phases spaces we create, the number of particles in the bin divided by the bin volume dx*dp_x would be similar. But is this quantity intuitive and useful?

PrometheusPi:
Yes, this normalization would be more common. 
Perhaps the work of @ikbuibui will solve this problem, and we can remove the entire phase space plugin.

PR #4623, opened 2023-07-05
add lumi template and setup profile

PrometheusPi:
This pull request adds the `*.tpl` for the MI250x queue on [LUMI](https://www.lumi-supercomputer.eu/user-support), called `standard-G` and provides a minimal profile. 
Both template and profile have been tested.

### DIFF
--- a/etc/picongpu/lumi-eurohpc/lumi-G_hipcc_picongpu.profile.example
+++ b/etc/picongpu/lumi-eurohpc/lumi-G_hipcc_picongpu.profile.example
@@ -1,0 +1,140 @@
+# Name and Path of this Script ############################### (DO NOT change!)
+export PIC_PROFILE=$(cd $(dirname $BASH_SOURCE) && pwd)"/"$(basename $BASH_SOURCE)
+
+# User Information ################################# (edit the following lines)
+#   - automatically add your name and contact to output file meta data
+#   - send me a mail on batch system jobs: NONE, BEGIN, END, FAIL, REQUEUE, ALL,
+#     TIME_LIMIT, TIME_LIMIT_90, TIME_LIMIT_80 and/or TIME_LIMIT_50
+export MY_MAILNOTIFY="NONE"
+export MY_MAIL="someone@example.com"
+export MY_NAME="$(whoami) <$MY_MAIL>"
+
+
+# Project Information ######################################## (edit this line)
+#   - project for allocation and shared directories
+#   - use lumi-allocations to see all your projects
+#export PROJID=<your_project>
+
+# Job control
+# Allocate more nodes then required to execute the jobs to be able to handle broken nodes
+# Oversubscribe the nodes allocated by N per thousand required nodes.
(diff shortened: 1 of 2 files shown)


etc/picongpu/lumi-eurohpc/lumi-G_hipcc_picongpu.profile.example
```diff
@@ -0,0 +1,147 @@
+printf "@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@\n" >&2
+printf "@ Set the initial buffer size in your ADIOS2              @\n" >&2
+printf "@ configuration of your job's *.cfg file to 28GiB,        @\n" >&2
```
psychocoderHPC: This message can be removed because we workaround the HIP issue that the reported free memory is wrong.

psychocoderHPC (changes requested):

etc/picongpu/lumi-eurohpc/lumi-G_hipcc_picongpu.profile.example
PrometheusPi: @psychocoderHPC is fixed now

etc/picongpu/lumi-eurohpc/standard-g.tpl
```diff
@@ -0,0 +1,167 @@
...
+export OMP_NUM_THREADS=!TBG_coresPerGPU
+
+# required GPUs per node for the current job
+.TBG_devicesPerNode=$(if [ $TBG_tasks -gt $TBG_numHostedDevicesPerNode ] ; then echo $TBG_numHostedDevicesPerNode; else echo $TBG_tasks; fi)
+
+# We only start 1 MPI task per device
+.TBG_mpiTasksPerNode="$(( TBG_devicesPerNode * 1 ))"
+
+# use ceil to caculate nodes
+.TBG_nodes="$((( TBG_tasks + TBG_devicesPerNode - 1 ) / TBG_devicesPerNode))"
```
PrometheusPi, line 78: This line creates with 16 x 16 x 16 GCDs not 512 but 513 nodes

PrometheusPi: But if I test it on the shell, I get (as expected):
```bash 
echo $((( 4096 + 8 - 1 ) / 8))
512
```

PrometheusPi: Okay - that's default tolerance adjustment - do we need that?

psychocoderHPC: Yes please keep it in the code it is very useful on larger systems. If `PIC_NODE_OVERSUBSCRIPTION_PT` is set to zero the adjustment is disabled.

psychocoderHPC (approved):

PR #4626, opened 2023-07-07
mpiInfo set RPath

psychocoderHPC:
With #4312 we forgot to set the rPATH for the helper tool mpiInfo. On systems with spack where no `LD_LIBRARY_PATH` the binary is missing the MPI library.

By adding rPATH to mpiInfo this bug is solved.


BrianMarre (commented):
Looks good to my untrained eye

steindev (approved):

PR #4632, opened 2023-07-12
collision: reduce memory fragmentation

psychocoderHPC:
Reduce the memory heap fragmentation by allocating always a multiple of the typical number of particles per cell for the particle cell list on GPUs.
This change will waste a little bit of memory but on the other side avoid the mallocMC heap fragmentation becoming too large which can result in out-of-memory errors.

**ATTENTION** in existing `collision.param` files `constexpr uint32_t cellListChunkSize = particles::TYPICAL_PARTICLES_PER_CELL;` must be added by hand to be compatible with this PR.

### DIFF
--- a/include/picongpu/particles/collision/detail/ListEntry.hpp
+++ b/include/picongpu/particles/collision/detail/ListEntry.hpp
@@ -64,8 +64,18 @@ namespace picongpu
                             for(int numTries = 0; numTries < maxTries; ++numTries)
                             {
 #if(BOOST_LANG_CUDA || BOOST_COMP_HIP) // Allocate memory on a GPU device
-                                ptrToIndicies
-                                    = (uint32_t*) deviceHeapHandle.malloc(worker.getAcc(), sizeof(uint32_t) * numPar);
+                                static_assert(
+                                    cellListChunkSize != 0u,
+                                    "collision::cellListChunkSize must be non zero!");
+                                /* Round-up the number of slots up to an mutiple of the typical amount of particles of
+                                 * a cell. This will waist memory but is avoiding that the mallocMC heap is fragmented
+                                 * which often results into out of memory crashes even if the amount of particles in
+                                 * the simulation is small.
+                                 */
+                                uint32_t const allocateNumParticles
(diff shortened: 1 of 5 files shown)


include/picongpu/particles/collision/detail/ListEntry.hpp
```diff
@@ -64,8 +64,18 @@ namespace picongpu
...
 #if(BOOST_LANG_CUDA || BOOST_COMP_HIP) // Allocate memory on a GPU device
-                                ptrToIndicies
-                                    = (uint32_t*) deviceHeapHandle.malloc(worker.getAcc(), sizeof(uint32_t) * numPar);
+                                constexpr uint32_t typicalParticlesPerSupercell
+                                    = particles::TYPICAL_PARTICLES_PER_CELL;
+                                /* Round-up the number of slots up to an mutiple of the typical amount of particles of
+                                 * a cell. This will waist memory but is avoiding that the mallocMC heap is fragmented
+                                 * which often results into out of memory crashes even if the amount of particles in
+                                 * the simulation is small.
+                                 */
```
pordyna, line 74: Maybe add a static assert that `typicalParticlesPerSupercell` is not 0, it shouldn't be, but it is sth that is usually set by the user in a param file.

pordyna:
Maybe consider adding sth like 
``` 
constexpr uint32_t LIST_CHUNK_SIZE = particles::TYPICAL_PARTICLES_PER_CELL;
```
in `collsion.param` and using it instead, so that it is configurable.

include/picongpu/particles/collision/detail/ListEntry.hpp
psychocoderHPC: done

psychocoderHPC:
> Maybe consider adding sth like
> 
> ```
> constexpr uint32_t LIST_CHUNK_SIZE = particles::TYPICAL_PARTICLES_PER_CELL;
> ```
> 
> in `collsion.param` and using it instead, so that it is configurable.

Done

pordyna (approved):

PrometheusPi (approved):

Issue #4635, opened 2023-07-18
openPMD-api does not set openPMDextension correctly


PrometheusPi:
As reported by @berceanu in #4633, we currently do not set the `openPMDextension` attribute in our main openPMD-api plugin output to `'ED-PIC'`. Thus, the weighting scaling introduced by the ED-PIC part of the standard is ignored by tools like the openPMD-viewer. This needs to be fixed (@psychocoderHPC before the next release).

psychocoderHPC:
should be fixed with #4636

PR #4636, opened 2023-07-19
plugin openPMD: activate support for ED-PIC extension

### LINKED ISSUE #4633: Energy histogram with PIConGPU 0.6
Hi, I am using the following code to plot the energy histogram for all the electrons of a PIConGPU 0.6 simulation:

```python
from openpmd_viewer import OpenPMDTimeSeries
import numpy as np

ts = OpenPMDTimeSeries("h5/gas_126175.h5", check_all_files=False, backend="h5py")

ux, uy, uz, w = ts.get_particle(
    ["ux", "uy", "uz", "w"], iteration=126175, species="e_all"
)

gamma = np.sqrt(1 + ux**2 + uy**2 + uz**2)
energy = (gamma - 1) * 0.511
print(energy)

Edata = energy
E_hist, nbins = np.histogram(Edata, bins=2048, weights=w)
bincenter = (nbins[:-1] + nbins[1:]) * 0.5

# Create a new figure and a set of subplots
fig, ax = plt.subplots()

# Set the x coordinates of the line plot. Use the midpoint of each bin as x-coordinates.

# Plot the histogram using a line plot
ax.plot(bincenter, E_hist)

# Set the title and labels
ax.set_title("Energy Distribution")
ax.set_xlabel("Energy (MeV)")
ax.set_ylabel("Number of 'real' electrons")

# Display the plot
plt.show()
```

which results in this figure:

Figure_1

I also tried with the `"openpmd-api"` backend, and the results are identical. Furthermore, I tried reading the momentum data directly from the `.h5` file using `openpmd-api` directly, without `openpmd-viewer` - same results.

The energy histogram plugin of PIConGPU gives the following plot:

e_energyhistogram_all_126000

Notice how the energy scale is vastly different from the one above. 
The thing is that this plotting script works just fine for PIConGPU 0.5, giving sensible values for the energy axis.

Did something change from 0.5 to 0.6 or am I missing something here?

Thanks! 

Here is the `.h5` file in question: [gas_126175.h5](https://transfer.sequanium.de/5sBkUGNPiF/gas_126175.h5) [279 MB].

psychocoderHPC:
fix #4633 #4635

PIConGPU missed to set that openPMD files using the ED_PIC extension.

### DIFF
--- a/include/picongpu/plugins/common/openPMDWriteMeta.hpp
+++ b/include/picongpu/plugins/common/openPMDWriteMeta.hpp
@@ -105,6 +105,10 @@ namespace picongpu
                  * optional metadata:
                  */
 
+                // PIConGPU is writing particles and fields based on the openPMD standard 'ED-PIC' extension.
+                constexpr uint32_t openPMDExtensionMask = 1u;
+                series.setOpenPMDextension(openPMDExtensionMask);
+
                 /*   recommended */
                 const std::string author = Environment<>::get().SimulationDescription().getAuthor();
                 if(author.length() > 0)


include/picongpu/plugins/common/openPMDWriteMeta.hpp
```diff
@@ -105,6 +105,10 @@ namespace picongpu
                  * optional metadata:
                  */
 
+                //  PIConGPU is writing particles based on the openPMD standard 'ED-PIC' extension.
```
steindev: Also the field data is in line with the ED-PIC extension of openPMD. I guess just the comment is wrong.

steindev (approved):

PR #4638, opened 2023-07-20
create E and B field from moving particles

psychocoderHPC:
Allow creating the E and B field based on initialized moving particles current.

- [ ] Currently, a draft to check if we need to add ISimulationData to more field solver implementations.

### DIFF
--- a/include/picongpu/particles/InitFunctors.hpp
+++ b/include/picongpu/particles/InitFunctors.hpp
@@ -23,6 +23,7 @@
 
 #include "picongpu/fields/Fields.def"
 #include "picongpu/particles/Manipulate.hpp"
+#include "picongpu/particles/StaticField.hpp"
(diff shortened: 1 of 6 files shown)


psychocoderHPC:
@PrometheusPi We should discuss how we should go on with this PR.

PrometheusPi:
@psychocoderHPC If I load my particles from a checkpoint, I cannot apply the new method, right?

psychocoderHPC:
I will close this PR because this PR is only producing the current from a warm plasma and propagating the field's N steps.
The concepts will be used in another upcoming feature by one of our students.

psychocoderHPC:
I reopened this PR because I need to port `ExecuteIfTimeStepEq` to our code base, these are useful objects for our particle init pipe and the per timestep pipe.

psychocoderHPC:
time dependent functor execution is moved out to PR #4934

PR #4644, opened 2023-08-04
fix openPMD plugin output

### LINKED ISSUE #4641: openPMD-api writes wrong blocks in meshes - indexing bug
As discussed in https://github.com/openPMD/openPMD-api/issues/1485 PIConGPU seems to write incorrect blocks in openPMD causing shift in data and undefined regions. 

Please see the linked issue for details.

psychocoderHPC:
fix #4641

bug introduced with #4585

In the openPMD plugin we use a param class `ThreadParams` where we collect many information for the dumping process to avoid forwarding all in via function parameters. The object is very dirty and we can not track life-times and different kind of data are stored together. The bug was introduced because the timestep which was not default initilized on all systems was required in one of the dumping function `intersectRangeWithWindow()`  to
calculate the intersection of the range parameter in the openPMD plugin and the moving window of the simulation.

**ATTENTION** The restart is affected by this bug too! IMO it is not guaranteed that the restarted timestep is what you selected or all could simply crash at runtime.

This PR refactore the interfaces and is passing now the timestep as function parameter.

### DIFF
--- a/include/picongpu/plugins/openPMD/openPMDWriter.hpp
+++ b/include/picongpu/plugins/openPMD/openPMDWriter.hpp
@@ -563,6 +563,7 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
         void ThreadParams::initFromConfig(
             Help& help,
             size_t id,
+            uint32_t const currentStep,
             std::string const& dir,
             std::optional<std::string> file)
         {
@@ -681,7 +682,7 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                     return createUnit(unit, T_Field::numComponents);
                 }
 
-                HDINLINE void operator()(ThreadParams* params)
+                HDINLINE void operator()(ThreadParams* params, uint32_t const currentStep)
                 {
 #ifndef __CUDA_ARCH__
                     DataConnector& dc = Environment<simDim>::get().DataConnector();
@@ -710,6 +711,7 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
 
                     openPMDWriter::writeField<ComponentType>(
                         params,
+                        currentStep,
                         GetNComponents<ValueType>::value,
                         T_Field::getName(),
                         *field,
@@ -740,9 +742,9 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                  * call virtual functions.
                  */
                 PMACC_NO_NVCC_HDWARNING
-                HDINLINE void operator()(ThreadParams* tparam)
+                HDINLINE void operator()(ThreadParams* tparam, uint32_t const currentStep)
                 {
-                    this->operator_impl(tparam);
+                    this->operator_impl(tparam, currentStep);
                 }
 
             private:
@@ -774,7 +776,7 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                     return FieldTmpOperation<Solver, Species, Filter>::getName();
                 }
 
-                HINLINE void operator_impl(ThreadParams* params)
+                HINLINE void operator_impl(ThreadParams* params, uint32_t const currentStep)
                 {
                     DataConnector& dc = Environment<>::get().DataConnector();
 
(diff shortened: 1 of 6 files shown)


include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -1297,7 +1305,6 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                 // local offset + extent
                 const pmacc::Selection<simDim> localDomain = Environment<simDim>::get().SubGrid().getLocalDomain();
                 mThreadParams.cellDescription = m_cellDescription;
-                mThreadParams.currentStep = currentStep;
```
psychocoderHPC, line 1300: This is the root of the bug, in `notify()` where we call this method `dumpData()' we call `intersectRangeWithWindow()` where the timestep is required. So we pointed always to the time-step of the last dump performed within this plugin instance.

psychocoderHPC:
I tested the PR on hemera (HZDR) and the issues are gone!

psychocoderHPC:
Since many of the maintainers have next week vacations I will merge this PR so that our IO in the dev branch is fixed.
@PrometheusPi Even the PR is already merged please give the PR a review.

PrometheusPi:
@psychocoderHPC I submitted simulations with your fix. Results will be available by tomorrow. Then I can confirm that the fix works by a run time test.

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -1455,6 +1464,10 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                 pmacc::math::UInt64<simDim> recordOffsetDims = params->window.localDimensions.offset;
                 pmacc::math::UInt64<simDim> recordGlobalSizeDims = params->window.globalDimensions.size;
 
+                std::cout << "record: flobalOffset" << recordOffsetDims.toString()
```
PrometheusPi, line 1467: typo: 
`flobalOffset` should be `globalOffset`

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -1455,6 +1464,10 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
...
+                          << " glbalsize=" << recordGlobalSizeDims.toString()
```
PrometheusPi, line 1468: typo: 
`glbalsize` should be `globalsize`

PrometheusPi (commented):
I see what you have changed and that looks valid. I do not understand, how this fixes the bug, but this is something we can discuss in person later. Whether the fix works will be validated by the run time test currently executed. 

There were some typos. I will provide a fix.

PrometheusPi:
The block distribution looks correct now. 
Only the last block in iteration zero has the same distribution as the first (C: 0 index) block:

```
  float     /data/0/fields/e_all_chargeDensity                        {1, 1488, 768}
        step 0:
          block   0: [0:0,    0:  47,   0:127]
          block   1: [0:0,   48:  95,   0:127]
          block   2: [0:0,   96: 143,   0:127]
          block   3: [0:0,  144: 191,   0:127]
          block   4: [0:0,  192: 239,   0:127]
          block   5: [0:0,  240: 287,   0:127]
          block   6: [0:0,  288: 335,   0:127]
          block   7: [0:0,  336: 383,   0:127]
          block   8: [0:0,  384: 431,   0:127]
          block   9: [0:0,  432: 479,   0:127]
          block  10: [0:0,  480: 527,   0:127]
          block  11: [0:0,  528: 575,   0:127]
          block  12: [0:0,  576: 623,   0:127]
          block  13: [0:0,  624: 671,   0:127]
          block  14: [0:0,  672: 719,   0:127]
          block  15: [0:0,  720: 767,   0:127]
          block  16: [0:0,  768: 815,   0:127]
          block  17: [0:0,  816: 863,   0:127]
          block  18: [0:0,  864: 911,   0:127]
          block  19: [0:0,  912: 959,   0:127]
          block  20: [0:0,  960:1007,   0:127]
          block  21: [0:0, 1008:1055,   0:127]
          block  22: [0:0, 1056:1103,   0:127]
          block  23: [0:0, 1104:1151,   0:127]
          block  24: [0:0, 1152:1199,   0:127]
          block  25: [0:0, 1200:1247,   0:127]
          block  26: [0:0, 1248:1295,   0:127]
          block  27: [0:0, 1296:1343,   0:127]
          block  28: [0:0, 1344:1391,   0:127]
          block  29: [0:0, 1392:1439,   0:127]
          block  30: [0:0, 1440:1487,   0:127]
          block  31: [0:0,    0:  47, 128:255]
```

PrometheusPi:
The strange issue with block 31 originates from me, clipping the output. There are 186 blocks in total. All lock good.

PR #4647, opened 2023-08-11
simplified directory specification, cmakeFlagReader

MSVoss:
Simplifies entering the simulation directory. All you have to do is specify the main path, all other paths are determined by the test suite itself. (Furthermore, the paths can be specified separately if the paths differ).

Furthermore, the cmake flag reader was added. This means that changed parameters can now also be read from the cmakFlags file. This is the basis for case differentiation by the test suite.

### DIFF
--- a/lib/python/test/setups/MI/main.py
+++ b/lib/python/test/setups/MI/main.py
@@ -21,6 +21,14 @@ parser = argparse.ArgumentParser(description="Starts the test suite."
                                  " passed. Please refer to the Data.py"
                                  " documentation for more information.")
 
+parser.add_argument("-d", help="Main path of the simulation, if this"
+                    " is specified, no other one must be specified.",
+                    dest="direction",
+                    type=str,
+                    const=None,
+                    nargs="?",
+                    default=None)
+
 parser.add_argument("-r", help="Path of the folder where the results"
                     " of the test-suite should be saved",
                     dest="result",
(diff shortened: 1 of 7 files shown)


lib/python/test/setups/MI/main.py
```diff
@@ -21,6 +21,14 @@
                                  " passed. Please refer to the Data.py"
                                  " documentation for more information.")
 
+parser.add_argument("-d", help="Main path of the simulation, if this"
+                    " is specified, no other one must be specified.",
```
PrometheusPi, line 25: very useful comment - thank you

lib/python/test/testsuite/Reader/readFiles.py
```diff
@@ -62,6 +62,12 @@ def __init__(self, fileExtension: str,
             raise TypeError("You must set at least one of the"
                             " values(direction or directiontype)")
 
+        self._fileExtension = fileExtension
+        self._directiontype = directiontype
+
+        if directiontype is None:
+            directiontype = "undefined"
```
PrometheusPi, line 66: shouldn't this be:
```suggestion
            self._directiontype = "undefined"
```

lib/python/test/testsuite/_manager.py
```diff
@@ -9,19 +9,87 @@
...
+    """
+    Main function of the test-suite, starts and runs the test-suite.
+
+    Input:
+    -------
+    direction:  str, optional
+                Main path of the simulation
+                All other paths are determined from this.
+                All others only have to be defined if they
+                deviate from the standard path or Direction is None.
```
PrometheusPi:
```suggestion
                deviate from the standard path or direction is None.
```

PrometheusPi (changes requested):
Looks (nearly) good to me - I am just confused about a single assignment. Is this correct?

lib/python/test/testsuite/Reader/readFiles.py
MSVoss: No, not in this particular case. A different name for directiontype is chosen here, since checkData.checkDirection cannot handle None.

PrometheusPi: Okay

PrometheusPi (approved):

PrometheusPi:
Thanks for your clarification @MSVoss. 
There is currently a CI issue with respect to building our read the docs preventing merging. 
@psychocoderHPC Do you know what causes this?

PrometheusPi:
I ran a run time test and it performed fine. I will merge this PR, despite `docs/readthedocs.org:picongpu` failing, since `docs/readthedocs.org:picongpu` has noting to do with this PR and is a separate issue #4654

PR #4649, opened 2023-08-16
TransitionRadiation plugin openPMD output

finnolec:
This PR adds openPMD output to the Transition Radiation plugin. Additionally, as a Quality of Life feature, it made the virtual foil position a command line parameter and it changed the plugin to a multi-plugin.

During the refactoring I noticed, that the current value storage implementation is quite messy with the internal 3D to 1D mapping, but this is a task for a future PR. Also this plugin would profit, as soon as the openPMD standard allows axes with different unit systems. The plugin with the openPMD output is probably less flexible in defining the observers now, this might need to be changed in the future as well.

@BeyondEspresso @psychocoderHPC

### DIFF
--- a/docs/source/usage/plugins/transitionRadiation.rst
+++ b/docs/source/usage/plugins/transitionRadiation.rst
@@ -209,7 +209,7 @@ Command line option                        Description
 ``--<species>_transRad.period``            Gives the number of time steps between which the radiation should be calculated.
 ``--<species>_transRad.foilPositionY``     Absolute position in SI units to put a virtual foil for calculating the transition radiation. See above for more information. Disabled = 0.0. Default: 0.0
 ``--<species>_transRad.file``              File name to stodre transition radiation in. Default: transRad
-``--<species>_transRad.ext``               openPMD filename extension. Default: ext
+``--<species>_transRad.ext``               Backend for openPMD output. Default: default that is used internally
 ``--<species>_transRad.datOutput``         Optional text file output in numpy-readable format. Enabled = 1. Default: 0
 ========================================== ==============================================================================================================================
 
--- a/docs/source/usage/plugins/transitionRadiation.rst
+++ b/docs/source/usage/plugins/transitionRadiation.rst
@@ -53,7 +53,7 @@ The transition radiation can only be calculated for electrons at the moment.
 External Dependencies
 ^^^^^^^^^^^^^^^^^^^^^
 
-There are no external dependencies.
+The plugin is available as soon as the :ref:`openPMD API <install-dependencies>` is compiled in.
 
 .param files
 ^^^^^^^^^^^^
@@ -135,15 +135,7 @@ Foil Position
 If one wants to virtually propagate the electron bunch to a foil in a further distance to get a rough estimate of the effect of the divergence on the electron bunch, one can include a foil position.
 A foil position which is unequal to zero, adds the electrons momentum vectors onto the electron until they reach the given y-coordinate.
 To contain the longitudinal information of the bunch, the simulation window is actually virtually moved to the foil position and not each single electron.
-
-.. code:: cpp
-
-    namespace SI
-    {
-        // y position of the foil to calculate transition radiation at
-        // leave at 0 for no virtual particle propagation
-        constexpr float_64 foilPosition = 0.0;
-    }
+This is a run time parameter which can be set with ``--<species>_transRad.foilPositionY``.
 
 .. note::
 
@@ -215,6 +207,10 @@ For a specific (charged) species ``<species>`` e.g. ``e``, the radiation can be 
 Command line option                        Description
 ========================================== ==============================================================================================================================
 ``--<species>_transRad.period``            Gives the number of time steps between which the radiation should be calculated.
+``--<species>_transRad.foilPositionY``     Absolute position in SI units to put a virtual foil for calculating the transition radiation. See above for more information. Disabled = 0.0. Default: 0.0
+``--<species>_transRad.file``              File name to stodre transition radiation in. Default: transRad
+``--<species>_transRad.ext``               openPMD filename extension. Default: ext
+``--<species>_transRad.datOutput``         Optional text file output in numpy-readable format. Enabled = 1. Default: 0
 ========================================== ==============================================================================================================================
 
 Memory Complexity
--- a/include/picongpu/plugins/transitionRadiation/TransitionRadiation.kernel
+++ b/include/picongpu/plugins/transitionRadiation/TransitionRadiation.kernel
@@ -65,7 +65,7 @@ namespace picongpu
                  * @param mapper MappingDesction object
                  * @param freqFkt frequency functor
                  * @param simBoxSize size of simulation box
-                 * @param foilPosition position of the virtual foil to calculate the TR for in PIC units
+                 * @param foilPositionY position of the virtual foil to calculate the TR for in PIC units
                  */
                 template<
                     typename T_ParBox,
@@ -84,7 +84,7 @@ namespace picongpu
                     T_Mapping mapper,
                     transitionRadiation::frequencies::FreqFunctor freqFkt,
                     DataSpace<simDim> simBoxSize,
-                    float_X foilPosition) const
+                    float_X foilPositionY) const
                 {
                     using complex_X = alpaka::Complex<float_X>;
                     using complex_64 = alpaka::Complex<float_64>;
@@ -138,7 +138,7 @@ namespace picongpu
                     int const numSuperCells = superCellsCount.productOfComponents();
 
                     // propagation distance for the particle bunch
-                    float_X const propagationDistance = foilPosition - globalOffset[1];
+                    float_X const propagationDistance = foilPositionY - globalOffset[1];
 
 
                     /* go over all super cells on GPU
@@ -264,7 +264,7 @@ namespace picongpu
 
                                                 // only propagate particles if it is set up in
                                                 // transitionRadiation.param
-                                                if(foilPosition != 0.0)
+                                                if(foilPositionY != 0.0)
                                                     particle.propagate(propagationDistance);
 
                                                 // create calculator for TR calculations
--- a/include/picongpu/plugins/transitionRadiation/TransitionRadiation.kernel
+++ b/include/picongpu/plugins/transitionRadiation/TransitionRadiation.kernel
@@ -65,6 +65,7 @@ namespace picongpu
                  * @param mapper MappingDesction object
                  * @param freqFkt frequency functor
                  * @param simBoxSize size of simulation box
+                 * @param foilPosition position of the virtual foil to calculate the TR for in PIC units
                  */
                 template<
                     typename T_ParBox,
@@ -82,7 +83,8 @@ namespace picongpu
                     DataSpace<simDim> globalOffset,
                     T_Mapping mapper,
                     transitionRadiation::frequencies::FreqFunctor freqFkt,
-                    DataSpace<simDim> simBoxSize) const
+                    DataSpace<simDim> simBoxSize,
+                    float_X foilPosition) const
                 {
                     using complex_X = alpaka::Complex<float_X>;
                     using complex_64 = alpaka::Complex<float_64>;
@@ -136,7 +138,7 @@ namespace picongpu
                     int const numSuperCells = superCellsCount.productOfComponents();
 
                     // propagation distance for the particle bunch
-                    float_X const propagationDistance = parameters::foilPosition - globalOffset[1];
+                    float_X const propagationDistance = foilPosition - globalOffset[1];
 
 
                     /* go over all super cells on GPU
@@ -262,7 +264,7 @@ namespace picongpu
 
                                                 // only propagate particles if it is set up in
                                                 // transitionRadiation.param
-                                                if(parameters::foilPosition != 0.0)
+                                                if(foilPosition != 0.0)
                                                     particle.propagate(propagationDistance);
 
                                                 // create calculator for TR calculations
--- a/include/picongpu/plugins/transitionRadiation/TransitionRadiation.x.cpp
+++ b/include/picongpu/plugins/transitionRadiation/TransitionRadiation.x.cpp
@@ -115,6 +115,7 @@ namespace picongpu
                 std::string pluginPrefix;
                 std::string folderTransRad;
                 std::string filenamePrefix;
+                std::string fileExtension;
 
                 bool isMaster = false;
                 uint32_t currentStep = 0;
@@ -173,7 +174,11 @@ namespace picongpu
                     desc.add_options()(
                         (pluginPrefix + ".period").c_str(),
                         po::value<std::string>(&notifyPeriod),
-                        "enable plugin [for each n-th step]");
+                        "enable plugin [for each n-th step]")(
+                        (pluginPrefix + ".ext").c_str(),
+                        po::value<std::string>(&fileExtension)->default_value("bp"),
+                        "openpmd file extension"
+                        );
                 }
 
                 /** Implementation of base class function.
@@ -338,6 +343,7 @@ namespace picongpu
                         writeFile(
                             theTransRad.data(),
                             folderTransRad + "/" + filenamePrefix + "_" + o_step.str() + ".dat");
+                        writeOpenPMDFile(currentStep);
                     }
                 }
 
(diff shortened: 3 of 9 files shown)


share/picongpu/examples/TransitionRadiation/etc/picongpu/16.cfg
```diff
@@ -52,8 +52,10 @@ TBG_pngYX="--e_png.period 10 \
            --e_png.slicePoint 0.5 \
            --e_png.folder pngElectronsYX"
 
-# transition radiation plugi
-TBG_e_transRad="--e_transRad.period 10"
+# transition radiation plugin
+TBG_e_transRad="--e_transRad.period 10 \
+                --e_transRad.file transRad \
+                --e_transRad.foilPosition 0.0"
```
PrometheusPi: Is a foil at 0 useful?

share/picongpu/examples/TransitionRadiation/etc/picongpu/1.cfg
```diff
@@ -52,8 +52,10 @@ TBG_pngYX="--e_png.period 10 \
            --e_png.slicePoint 0.5 \
            --e_png.folder pngElectronsYX"
 
-# transition radiation plugi
-TBG_e_transRad="--e_transRad.period 10"
+# transition radiation plugin
+TBG_e_transRad="--e_transRad.period 10 \
+                --e_transRad.file transRad \
+                --e_transRad.foilPosition 0.0"
```
PrometheusPi: is a foil at 0 useful?

include/picongpu/plugins/transitionRadiation/TransitionRadiation.kernel
```diff
@@ -138,7 +140,7 @@ namespace picongpu
                     int const numSuperCells = superCellsCount.productOfComponents();
 
                     // propagation distance for the particle bunch
-                    float_X const propagationDistance = parameters::foilPosition - globalOffset[1];
+                    float_X const propagationDistance = foilPosition - globalOffset[1];
```
PrometheusPi: Just for my understanding, why do you subtract  `globalOffset[1]`?

share/picongpu/examples/TransitionRadiation/etc/picongpu/1.cfg
PrometheusPi: Doesn't that mean `0==disabled`?

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -449,6 +528,204 @@ namespace picongpu
...
+                }
+
+
+                void restart(uint32_t restartStep, std::string const& restartDirectory) override
+                {
+                }
+
+                void checkpoint(uint32_t currentStep, std::string const& checkpointDirectory) override
+                {
+                }
```
PrometheusPi: Just a reminder for me: this plugin does not require a restart routine.

PrometheusPi (dismissed):
looks good to me. Just a question regarding the default foil position in the cfg files.

share/picongpu/examples/TransitionRadiation/etc/picongpu/1.cfg
finnolec: No, a foil at 0 is physically useless, so I have this is the "the feature is disabled"-value. The point of this virtual foil propagation is, that one can naively "simulate" the vacuum-propagation of an accelerated electron bunch from an LWFA after the plasma stage, without having to run the simulation millimeters to centimeters that would mimic the experimental setup. The evolution of the bunch during the plasma stage is already handled much better by the PIC loop itself, if you want to know how the CTR spectra look before the current bunch position, you just call the plugin earlier. 
I was struggling a bit with what the clearest way is to have this optional parameter. Maybe I should add a new parameter called `--e_transRad.foilEnabled 0` to separate the activation and the parameter of this feature.

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
finnolec: Should I add this as a comment?

include/picongpu/plugins/transitionRadiation/TransitionRadiation.kernel
finnolec: This took me some minutes to figure it out myself. I think, the idea is, that if we are in the simulation at `y=4mm`, but the foil is set to `foilPosition=10mm`, we only propagate the electrons 6mm instead of 10mm from their current position to the foil position. For this, it is using the simulation box position `globalOffest[1]` instead of the exact position of the single particles to not destroy the longitudinal/temporal coherence effects between the particles themselves.

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -77,12 +83,102 @@ namespace picongpu
...
+                           "openPMD filename extension. This controls the"
+                           "backend picked by the openPMD API. Available extensions: ["
+                               + openPMD::printAvailableExtensions() + "]",
+                           openPMD::getDefaultExtension().c_str()};
+                    plugins::multi::Option<bool> optionTextOutput
+                        = {"datOutput",
+                           "optional output: transition radiation as text file in readable format for the "
+                           "in picongpu provided python analysis script, 1==enabled",
+                           0};
+                    plugins::multi::Option<float_X> foilPositionSI
```
psychocoderHPC: The naming should reflect that the position is for the Y-direction.

psychocoderHPC (dismissed):

franzpoeschel:
I can have a look at this after my holiday (Sep 4th)

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -449,6 +528,204 @@ namespace picongpu
                     }
                 }
 
+                void writeOpenPMDFile(uint32_t currentStep)
+                {
+                    auto buffer = std::make_unique<HostBuffer<float_X, DIM3>>(DataSpace<DIM3>(
+                        parameters::nTheta,
+                        parameters::nPhi,
+                        transitionRadiation::frequencies::nOmega));
```
franzpoeschel: Is there a specific reason why this would be a unique pointer?

I see two possible ways to improve the memory footprint here:

1. Use the Span API. This works only if use of the HostBuffer is not strictly necessary, because the Span API lets openPMD allocate a buffer. (`auto span =  transitionRadiation.storeChunk(offset, extent); span.currentBuffer()[index] = "<your data here>";`)
2. Use the unique-ptr based storeChunk overload, e.g. instead of passing a shared pointer of float to `storeChunk()`, pass a unique pointer. The IO backend can then potentially use the knowledge that it is the unique owner for optimizations (especially: the BP5 engine of ADIOS2 can avoid one data copy this way)

Do tell me if any of these options might work, then I can help you prepare such a workflow.

If not, then please consider using `storeChunkRaw()` from `include/picongpu/plugins/common/openPMDVersion.def` instead of creating a non-owning shared pointer.

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -449,6 +528,204 @@ namespace picongpu
...
+                    ::openPMD::Offset offset = {0, 0, 0};
+                    ::openPMD::Datatype datatype = ::openPMD::determineDatatype<float_X>();
+                    ::openPMD::Dataset dataset{datatype, extent};
+
+                    auto mesh = series.iterations[currentStep].meshes["transitionradiation"];
+
+                    mesh.setAxisLabels(std::vector<std::string>{"omega index", "phi index", "theta index"});
+                    mesh.setDataOrder(::openPMD::Mesh::DataOrder::C);
+                    mesh.setGridUnitSI(1);
+                    mesh.setGridSpacing(std::vector<double>{1, 1, 1});
```
franzpoeschel: I take it that this is the use case you mentioned a while ago where the mesh is not purely spatial? Otherwise, this specifies that one cell is a 1m cube.

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -449,6 +528,204 @@ namespace picongpu
...
+                    auto transitionRadiation = mesh[::openPMD::RecordComponent::SCALAR];
+                    transitionRadiation.resetDataset(dataset);
+
+                    transitionRadiation.setUnitSI(
+                        SI::ELECTRON_CHARGE_SI * SI::ELECTRON_CHARGE_SI
+                        * (1.0 / (4 * PI * SI::EPS0_SI * PI * PI * SI::SPEED_OF_LIGHT_SI)));
+
+                    auto sharedDataPtr = std::shared_ptr<float_X>{buffer->getPointer(), [](auto const*) {}};
+
+                    transitionRadiation.storeChunk(sharedDataPtr, offset, extent);
```
franzpoeschel: See comment above, I recommend changing this either to (1) ideally a Span-based storeChunk(), (2) as a slightly weaker optimization a unique-ptr based storeChunk(), (3) or if those don't work a raw-pointer based storeChunk().

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -449,6 +528,204 @@ namespace picongpu
...
+
+                    ::openPMD::Extent extent
+                        = {static_cast<unsigned long int>(transitionRadiation::frequencies::nOmega),
+                           static_cast<unsigned long int>(parameters::nPhi),
+                           static_cast<unsigned long int>(parameters::nTheta)};
+                    ::openPMD::Offset offset = {0, 0, 0};
+                    ::openPMD::Datatype datatype = ::openPMD::determineDatatype<float_X>();
+                    ::openPMD::Dataset dataset{datatype, extent};
+
+                    auto mesh = series.iterations[currentStep].meshes["transitionradiation"];
```
franzpoeschel:
```suggestion
                    auto iteration = series.writeIterations()[currentStep];
                    auto mesh = iteration.meshes["transitionradiation"];
```

This API has better support for ADIOS2 steps.

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
finnolec: Yes, this grid has different units depending on the axis and I tried to use the radiation-plugin approach by having a separate grid for every axis, which determines the unit.

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
PrometheusPi: Not necessarily – as @psychocoderHPC code is our documentation, thus no code means no functionality – I am fine with that

docs/source/usage/plugins/transitionRadiation.rst
```diff
@@ -215,6 +207,10 @@ For a specific (charged) species ``<species>`` e.g. ``e``, the radiation can be
 Command line option                        Description
 ========================================== ==============================================================================================================================
 ``--<species>_transRad.period``            Gives the number of time steps between which the radiation should be calculated.
+``--<species>_transRad.foilPositionY``     Absolute position in SI units to put a virtual foil for calculating the transition radiation. See above for more information. Disabled = 0.0. Default: 0.0
+``--<species>_transRad.file``              File name to stodre transition radiation in. Default: transRad
+``--<species>_transRad.ext``               openPMD filename extension. Default: ext
```
PrometheusPi:
```suggestion
``--<species>_transRad.ext``               openPMD filename extension. Default: bp
```
or just write that it use the default used intrenally.

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -77,12 +83,102 @@ namespace picongpu
...
+                        = {"period", "enable plugin [for each n-th step]"};
+                    plugins::multi::Option<std::string> optionFileName
+                        = {"file", "file name to store transition radiation in: ", "transRad"};
+                    plugins::multi::Option<std::string> optionFileExtention
+                        = {"ext",
+                           "openPMD filename extension. This controls the"
+                           "backend picked by the openPMD API. Available extensions: ["
+                               + openPMD::printAvailableExtensions() + "]",
+                           openPMD::getDefaultExtension().c_str()};
+                    plugins::multi::Option<bool> optionTextOutput
```
PrometheusPi: only suggestion: how about `po::bool_switch` instead?

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -219,43 +303,36 @@ namespace picongpu
                  * transition radiation calculation and create a folder for transition
                  * radiation storage.
                  */
-                void pluginLoad() override
+                void init()
```
PrometheusPi: why is this now an `init()` and no longer `pluginLoad()`?

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -449,6 +528,177 @@ namespace picongpu
...
+                    meshOmega.setGridSpacing(std::vector<double>{1, 1, 1});
+                    meshOmega.setGeometry(::openPMD::Mesh::Geometry::cartesian); // set be default
+
+                    meshOmega.setUnitDimension(
+                        std::map<::openPMD::UnitDimension, double>{{::openPMD::UnitDimension::T, -1.0}});
+
+                    auto omega = meshOmega[::openPMD::RecordComponent::SCALAR];
+                    omega.resetDataset(datasetOmega);
+
+                    omega.setUnitSI(1.0);
```
PrometheusPi: shouldn't that be `1/UNIT_TIME`?

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
```diff
@@ -449,6 +528,177 @@ namespace picongpu
...
+                    meshPhi.setAxisLabels(std::vector<std::string>{"", "phi", ""});
+                    meshPhi.setDataOrder(::openPMD::Mesh::DataOrder::C);
+                    meshPhi.setGridUnitSI(1);
+                    meshPhi.setGridSpacing(std::vector<double>{1, 1, 1});
+                    meshPhi.setGeometry(::openPMD::Mesh::Geometry::cartesian); // set be default
+
+                    auto phi = meshPhi[::openPMD::RecordComponent::SCALAR];
+                    phi.resetDataset(datasetPhi);
+
+                    phi.setUnitSI(1.0);
```
PrometheusPi: is `phi.setUnitDimension(...)` be default all zeros?

PrometheusPi (dismissed):
looks good to me - just a few questions - and I will do a run time test before merging

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
franzpoeschel: Discussed today: We will leave this for now as I plan to implement https://github.com/openPMD/openPMD-standard/pull/193 soon which will help specifying this more precisely.

PrometheusPi:
@finnolec what is the status of this pull request? 
may I test it or will there still be major changes and I should wait?

finnolec:
I'm currently busy with different things, I will look into this PR in February again. I guess, you can wait until then with testing.

psychocoderHPC:
@finnolec  please finish this PR or can we close it?

docs/source/usage/plugins/transitionRadiation.rst
finnolec: done

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
finnolec: It was previously `pluginLoad()` because this is an inherited and implemented function in `ILightweightPlugin`. By switching to `plugins::multi::IInstance`, this function does not exist anymore. I prefer `init` slightly over `pluginLoad` and it's probably clearer, that this is now not function from the parent class

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
finnolec: What would be the advantage?

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
finnolec: The frequencies given by `transitionRadiation::frequencies::FreqFunctor` are not normalized, but in SI. So it should just be 1.0

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
finnolec: Offline discussion: It doesn't work for multi plugins

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
finnolec: I thought so, but I've clarified it now in the code :)

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
finnolec: done

finnolec:
@PrometheusPi I implemented your suggestions, do you want to have another look?

chillenzer:
Who would be capable of answering your last question? Maybe @PrometheusPi?

steindev:
ping @PrometheusPi

psychocoderHPC:
Will this PR ever be finished? If there is no update within the next two weeks I will close this PR it is open since 2023.

PrometheusPi:
I will have a look today.

PrometheusPi:
There were some issues with the CI and restart (and resubmit) was needed - I will have a look as soon as the CI gives a green light.

include/picongpu/plugins/transitionRadiation/TransitionRadiation.x.cpp
```diff
@@ -105,31 +204,43 @@ namespace picongpu
...
+
             public:
                 //! Constructor
-                TransitionRadiation()
-                    : pluginName("TransitionRadiation: calculate transition radiation of species")
-                    , speciesName(T_ParticlesType::FrameType::getName())
-                    , pluginPrefix(speciesName + std::string("_transRad"))
-                    , folderTransRad("transRad")
-                    , filenamePrefix(pluginPrefix)
+                TransitionRadiation(
```
PrometheusPi, line 229: For some reason, one of the CI compilers (*g++-10_serial_boost1.78.0_examples.TransitionRadiation*) does not recognize the arguments on the next line and assumes that there are 0 instead of 3 arguments)

```
/builds/hzdr/crp/picongpu/include/picongpu/plugins/transitionRadiation/TransitionRadiation.x.cpp:229:17: note: candidate: 'picongpu::plugins::transitionRadiation::TransitionRadiation<ParticlesType>::TransitionRadiation(std::shared_ptr<picongpu::plugins::multi::IHelp>&, size_t, picongpu::MappingDesc*) [with ParticlesType = picongpu::Particles<pmacc::meta::String<'e'>, boost::mp11::mp_list<picongpu::particlePusher<picongpu::particles::pusher::Boris, pmacc::pmacc_isAlias>, picongpu::shape<picongpu::particles::shapes::TSC, pmacc::pmacc_isAlias>, picongpu::interpolation<picongpu::FieldToParticleInterpolation<picongpu::particles::shapes::TSC, picongpu::AssignedTrilinearInterpolation>, pmacc::pmacc_isAlias>, picongpu::massRatio<picongpu::MassRatioElectrons, pmacc::pmacc_isAlias>, picongpu::chargeRatio<picongpu::ChargeRatioElectrons, pmacc::pmacc_isAlias> >, boost::mp11::mp_list<picongpu::position<picongpu::position_pic, pmacc::pmacc_isAlias>, picongpu::momentum, picongpu::weighting, picongpu::transitionRadiationMask> >; size_t = long unsigned int; picongpu::MappingDesc = pmacc::MappingDescription<3, pmacc::math::CT::Vector<std::integral_constant<int, 8>, std::integral_constant<int, 8>, std::integral_constant<int, 4> > >]'
  229 |                 TransitionRadiation(
      |                 ^~~~~~~~~~~~~~~~~~~
/builds/hzdr/crp/picongpu/include/picongpu/plugins/transitionRadiation/TransitionRadiation.x.cpp:229:17: note:   candidate expects 3 arguments, 0 provided
make[2]: *** [CMakeFiles/picongpu.dir/build.make:440: CMakeFiles/picongpu.dir/plugins/transitionRadiation/TransitionRadiation.x.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:207: CMakeFiles/picongpu.dir/all] Error 2
```

Retriggering the CI causes the same error.

PrometheusPi (changes requested):
There is either an issue with the CI or with the code.

include/picongpu/plugins/transitionRadiation/TransitionRadiation.x.cpp
psychocoderHPC: `PIC_REGISTER_SPECIES_PLUGIN(picongpu::plugins::multi::Master<picongpu::BinEnergyParticles<boost::mpl::_1>>);`

is an example how to use the macro at the end of this file. Each plugin must be constructable by `plugin()`

finnolec: That was the error!

include/picongpu/plugins/transitionRadiation/TransitionRadiation.hpp
PrometheusPi: Thanks for the clarification @finnolec.

PrometheusPi (dismissed):

PrometheusPi:
@psychocoderHPC Your review is still negative. Should I overroule? 
@finnolec Your branch still has a merge conflict with our current `dev` and can thus not be merged - please fix.

PrometheusPi:
Thanks @finnolec for fixing the merge conflict.

PrometheusPi (approved):

PR #4651, opened 2023-08-18
Add an optional time delay for incident field pulses

finnolec:
Add a time delay for ~~Gaussian~~ incident field pulses. This is extremely useful for simulation setups with multiple laser pulses, e.g. optical probe pulses. The parameter should be optional with the SFINAE detection, as many users will probably not need it and I don't want to break existing simulation setups.

@PrometheusPi @steindev @psychocoderHPC

### DIFF
--- a/include/picongpu/fields/incidentField/Functors.hpp
+++ b/include/picongpu/fields/incidentField/Functors.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2020-2023 Sergei Bastrakov
+/* Copyright 2020-2023 Sergei Bastrakov, Finn-Ole Carstens
  *
  * This file is part of PIConGPU.
  *
@@ -469,7 +469,8 @@ namespace picongpu
                     template<typename T_SeparableFunctor>
                     HDINLINE float3_X operator()(T_SeparableFunctor const& functor, floatD_X const& totalCellIdx) const
                     {
-                        auto const time = functor.getCurrentTime(totalCellIdx);
+                        auto const time
+                            = functor.getCurrentTime(totalCellIdx) - T_SeparableFunctor::Unitless::TIME_DELAY;
                         // Cut off when the laser has not entered at this point yet to avoid confusion.
                         if(time < 0.0_X)
                             return float3_X::create(0.0_X);
--- a/include/picongpu/fields/incidentField/Functors.hpp
+++ b/include/picongpu/fields/incidentField/Functors.hpp
@@ -113,7 +113,7 @@ namespace picongpu
             };
             namespace detail
             {
-                /** SFINAE dedection if the user parameter define the variable FOCUS_ORIGIN_*
+                /** SFINAE deduction if the user parameter define the variable FOCUS_ORIGIN_*
                  *
                  * This allows that focus origin can be an optional variable a user must only define if needed.
                  * The default if it is not defined is Origin::Zero
--- a/include/picongpu/fields/incidentField/Functors.hpp
+++ b/include/picongpu/fields/incidentField/Functors.hpp
@@ -201,9 +201,9 @@ namespace picongpu
 
                     /** Get current time to calculate field at the given point
                      *
-                     * It accounts for both current PIC iteration and location relative to origin.
-                     * Note that the result may be negative as well, and clients may want to set
-                     * field = 0 when the returned value is negative.
+                     * It accounts for both current PIC iteration, location relative to origin and the given time
+                     * delay. Note that the result may be negative as well, and clients may want to set field = 0 when
+                     * the returned value is negative.
                      *
                      * @param totalCellIdx cell index in the total domain
                      *
@@ -215,7 +215,7 @@ namespace picongpu
                     {
                         auto const shiftFromOrigin = totalCellIdx * cellSize - origin;
                         auto const distance = pmacc::math::dot(shiftFromOrigin, getDirection());
-                        auto const timeDelay = distance / phaseVelocity;
+                        auto const timeDelay = distance / phaseVelocity + Unitless::TIME_DELAY;
                         return currentTimeOrigin - timeDelay;
                     }
 
@@ -469,8 +469,7 @@ namespace picongpu
                     template<typename T_SeparableFunctor>
                     HDINLINE float3_X operator()(T_SeparableFunctor const& functor, floatD_X const& totalCellIdx) const
                     {
-                        auto const time
-                            = functor.getCurrentTime(totalCellIdx) - T_SeparableFunctor::Unitless::TIME_DELAY;
+                        auto const time = functor.getCurrentTime(totalCellIdx);
                         // Cut off when the laser has not entered at this point yet to avoid confusion.
                         if(time < 0.0_X)
                             return float3_X::create(0.0_X);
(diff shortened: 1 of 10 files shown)


finnolec:
As requested by @BrianMarre this PR added the time delay for all laser pulses.

include/picongpu/fields/incidentField/profiles/BaseParam.def
```diff
@@ -149,6 +149,17 @@ namespace picongpu
                     static constexpr Origin FOCUS_ORIGIN_Z = Origin::Zero;
                     /** @} */
 
+                    /** Time delay
+                     * For e.g. probe pulses, that enter the simulation volume later in time.
+                     * The laser will be initiation will start after the time delay.
```
BrianMarre: change to ```The laser initiation will be started after the time delay```

BrianMarre (approved):
just a short grammar error in the documentation

finnolec:
I quickly want to mention, that I only tested it for the Gaussian laser. However, I assume that it works the same for the other implementations

include/picongpu/fields/incidentField/profiles/BaseParam.def
PrometheusPi: I agree

include/picongpu/fields/incidentField/profiles/BaseParam.hpp
```diff
@@ -96,7 +96,35 @@ namespace picongpu
                             return 0.0;
                         }
 
+                        /** SFINAE detection if the user parameter define the variable TIME_DELAY_SI
```
PrometheusPi: What is: `SFINAE` ?

include/picongpu/fields/incidentField/profiles/BaseParam.hpp
```diff
@@ -96,7 +96,35 @@ namespace picongpu
...
+
                     public:
+                        /** Time delay
+                         *
+                         * unit: UNIT_TIME
+                         */
+                        static constexpr float_X TIME_DELAY
+                            = static_cast<float_X>(GetTimeDelay<Params>::value / UNIT_TIME);
+                        PMACC_CASSERT_MSG(
+                            _error_laser_time_delay_must_be_positive____check_your_incidentField_param_file,
```
PrometheusPi, line 125: Why does it have to be positive? 
Wouldn't a negative delay not just start the init before the simulation (which is not possible) and thus skip till start. 
Might be numerically questionably but valid option for delay scans to include negative values as well. 

This is just a question out of curiosity and does nor prevent merging.

PrometheusPi (dismissed):
looks good to me

include/picongpu/fields/incidentField/profiles/BaseParam.hpp
finnolec: SFINAE stands for [Substitution failure is not an error](https://en.wikipedia.org/wiki/Substitution_failure_is_not_an_error) which I use here to overload the template parameters of this function, which then makes the time delay optional by choosing the best fitting function to call. However, I copied this comment from `Functors.hpp` where this approach is also used to make the `Origin` optional. When copying, I tried to fix [the typo in "dedection"](https://github.com/ComputationalRadiationPhysics/picongpu/blob/7337a134fe7c2176f1e804a5385a2d3c467b8092/include/picongpu/fields/incidentField/Functors.hpp#L116) in the original code to "detection"... but looking at it now: Should it be "deduction"? @psychocoderHPC

include/picongpu/fields/incidentField/profiles/BaseParam.hpp
finnolec: The incident field pulses are only initialized on the Huygens-Surfaces, so a negative `TIME_DELAY` could lead to a cut-off laser pulse, because it is (unlike with the fieldBackground approach) not initialized in the simulation volume itself. I don't really see, why you physically would want to have that.

include/picongpu/fields/incidentField/Functors.hpp
```diff
@@ -469,7 +469,8 @@ namespace picongpu
                     template<typename T_SeparableFunctor>
                     HDINLINE float3_X operator()(T_SeparableFunctor const& functor, floatD_X const& totalCellIdx) const
                     {
-                        auto const time = functor.getCurrentTime(totalCellIdx);
+                        auto const time
+                            = functor.getCurrentTime(totalCellIdx) - T_SeparableFunctor::Unitless::TIME_DELAY;
```
psychocoderHPC: IMO this change should be done in the implementation of `getCurrentTime()` directly.
The reason for this is that now all places where `getCurrentTime()` is used we substitute the delay and this is an easy source to introduce errors by forgetting places.

Please update the implementation of `getCurrentTime()` and update the doxygen string with a hint that the time can be negative.

psychocoderHPC (changes requested):

include/picongpu/fields/incidentField/Functors.hpp
finnolec: The time delay is moved into `getCurrentTime()` now

psychocoderHPC (dismissed):

include/picongpu/fields/incidentField/profiles/BaseParam.hpp
psychocoderHPC: Yes should be deduction!

finnolec: Fixed this

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
PrometheusPi, line 1: here seem to be no changes except for the name in the license header
@finnolec should this be excluded from this PR?

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
PrometheusPi, line 1: same here

PrometheusPi (changes requested):
I think some changes requested by @psychocoderHPC resulted in touched files with no code or comment changes - but only license header changes. Please remove those from this PR.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
finnolec: In the meanwhile I added some code to this file in #4684, is it okay if I keep my name in `DispersivePulse.hpp` and then I just remove it from `GaussianPulse.hpp`?

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
finnolec: I will remove this in a second. I did some changes to this file when I started working on this PR, but these all got removed after I switched to implementing the delay directly though `Functors.hpp`

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
PrometheusPi: Yes

finnolec: And I did not add my name there because I wanted to avoid merge-conflicts with this PR

PrometheusPi (approved):

PR #4655, opened 2023-08-29
Deletion of deprecated code in the old KHI test

MSVoss:
from @PrometheusPi: This PR  deletes deprecated code in the old KHI test


PrometheusPi (approved):

PrometheusPi:
Thanks @MSVoss for deleting the old code in a separate pull request.

psychocoderHPC:
PLease add a description to the PR why the example will be deleted.

psychocoderHPC:
~~The KHI growth rate test is not shipping any cfg file, at least the file for single GPU should be shipped before KHI is removed.~~

@PrometheusPi ~~The growth rate test does not provide the cmakeFlags to enable radiation. There are two possibilities, the file `radiation.param` in the growth rate test will be deleted or one configuration in cmakeFlags should be added.~~

[update] due to the headline I misinterpreted what the PR is doing.

psychocoderHPC:
Ohh this PR is not removing the KHI test as in the headline mentioned. The dead verification code for the KHI growth rate test will be deleted. SO the headline need some tweaking and a description as mentioned before.

PR #4658, opened 2023-08-30
update LUMI standard-G tpl

PrometheusPi:
After the recent maintenance, using `--mpiDirect` on LUMI broke. 
The support informed us, that the pinning of GPUs to MPI tasks needs to be adjusted manually (now) since it is no longer handled by SLURM automatically. They provided the following example: https://lumi-supercomputer.github.io/LUMI-training-materials/User-Updates/Update-202308/lumig-lownoise/#mpi-based-job-example

This pull request now adds these changes to PIConGPU. It has been run time tested.

### DIFF
--- a/etc/picongpu/lumi-eurohpc/standard-g.tpl
+++ b/etc/picongpu/lumi-eurohpc/standard-g.tpl
@@ -28,10 +28,8 @@
 #SBATCH --job-name=!TBG_jobName
 #SBATCH --nodes=!TBG_nodes_adjusted
 #SBATCH --ntasks=!TBG_tasks_adjusted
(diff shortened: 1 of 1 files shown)


psychocoderHPC (approved):

etc/picongpu/lumi-eurohpc/standard-g.tpl
```diff
@@ -118,6 +113,26 @@ else
     run_cuda_memtest=0
 fi
 
+
+cat << EOF > select_gpu
+#!/bin/bash
+
+env
```
PrometheusPi, line 120: I introduced this line. It is too much output in most cases. I will open up a followup pull request to remove it.

PR #4557, opened 2023-05-08
Add CurrentDeposition test

MRZirper:
add single particle Current Deposition test from EZ-paper to CI

PR #4661, opened 2023-08-31
add CKC Solver

lennertsprenger:
This PR adds the ~Cowan-Cole-Kärkkäinen (CCK)~ Cole-Kärkkäinen-Cowan (CKC) Maxwell-solver, which was part of my bachelor thesis. For now only a draft because some further stability tests are needed (and documentation)

@steindev

EDIT (by @steindev): See further below for results of performed tests.

### DIFF
--- a/include/picongpu/fields/MaxwellSolver/CKC/CKC.hpp
+++ b/include/picongpu/fields/MaxwellSolver/CKC/CKC.hpp
@@ -1,0 +1,90 @@
+/* Copyright 2013-2023 Axel Huebl, Heiko Burau, Rene Widera, Remi Lehe,
+ *                     Sergei Bastrakov, Lennert Sprenger
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/simulation_defines.hpp"
+
+#include "picongpu/fields/MaxwellSolver/CFLChecker.hpp"
+#include "picongpu/fields/MaxwellSolver/CKC/CKC.def"
+#include "picongpu/fields/MaxwellSolver/CKC/Derivative.hpp"
+#include "picongpu/fields/MaxwellSolver/DispersionRelation.hpp"
+#include "picongpu/fields/MaxwellSolver/FDTD/FDTD.hpp"
+
+#include <pmacc/algorithms/math/defines/pi.hpp>
+#include <pmacc/traits/GetStringProperties.hpp>
+
+
+namespace picongpu
+{
+    namespace fields
+    {
+        namespace maxwellSolver
+        {
+            /** Specialization of the CFL condition checker for CKC solver
+             *
+             * @tparam T_Defer technical parameter to defer evaluation
+             */
+            template<typename T_Defer>
+            struct CFLChecker<CKC, T_Defer>
+            {
+                /** Check the CFL condition according to the paper, doesn't compile when failed
+                 *
+                 * @return upper bound on `c * dt` due to chosen cell size according to CFL condition
+                 */
+                float_X operator()() const
+                {
+                    // cellSize is not constexpr currently, so make an own constexpr array
+                    constexpr float_X step[3] = {CELL_WIDTH, CELL_HEIGHT, CELL_DEPTH};
+                    constexpr float_X cdt = SPEED_OF_LIGHT * getTimeStep(); // c * dt
+
+                    constexpr float_64 delta = std::min({step[0], step[1], step[2]});
+
+                    // Dependence on T_Defer is required, otherwise this check would have been enforced for each setup
+                    PMACC_CASSERT_MSG(
+                        Courant_Friedrichs_Lewy_condition_failure____check_your_grid_param_file,
+                        (cdt <= delta) && sizeof(T_Defer*) != 0);
+
+                    return delta;
+                }
+            };
+
+        } // namespace maxwellSolver
+    } // namespace fields
+} // namespace picongpu
+
+namespace pmacc
+{
+    namespace traits
+    {
+        template<>
+        struct StringProperties<::picongpu::fields::maxwellSolver::CKC>
+        {
+            static StringProperty get()
+            {
+                auto propList = ::picongpu::fields::maxwellSolver::CKC::getStringProperties();
+                // overwrite the name of the solver (inherit all other properties)
+                propList["name"].value = "CK";
+                return propList;
+            }
+        };
+
+    } // namespace traits
+} // namespace pmacc
--- a/include/picongpu/fields/MaxwellSolver/CKC/Derivative.hpp
+++ b/include/picongpu/fields/MaxwellSolver/CKC/Derivative.hpp
@@ -1,0 +1,157 @@
+/* Copyright 2013-2023 Axel Huebl, Heiko Burau, Rene Widera, Remi Lehe,
+ *                     Sergei Bastrakov, Lennert Sprenger
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/fields/MaxwellSolver/CKC/Derivative.def"
+#include "picongpu/fields/differentiation/Derivative.hpp"
+#include "picongpu/fields/differentiation/ForwardDerivative.hpp"
+#include "picongpu/fields/differentiation/Traits.hpp"
+#include "picongpu/traits/GetMargin.hpp"
+
+#include <pmacc/algorithms/math/defines/pi.hpp>
+#include <pmacc/dimensions/DataSpace.hpp>
+#include <pmacc/math/Vector.hpp>
+#include <pmacc/meta/accessors/Identity.hpp>
+#include <pmacc/types.hpp>
+
+#include <algorithm>
+
+
+namespace picongpu
+{
+    namespace fields
+    {
+        namespace maxwellSolver
+        {
+            namespace ckc
+            {
+                /** Functor for derivative used in the CKC solver
+                 *
+                 * Implements the modified derivative from
+                 *     Phys. Rev. ST Accel. Beams 16, 041303 (2013)
+                 *   https://dx.doi.org/10.1103/PhysRevSTAB.16.041303
+                 *
+                 * This derivative can only be applied for the E field.
+                 *
+                 * @tparam T_direction direction to take derivative in, 0 = x, 1 = y, 2 = z
+                 */
+                template<uint32_t T_direction>
+                struct DerivativeFunctor
+                {
+                private:
+                    //! Internally used derivative functor
+                    using InternalDerivativeFunctor
+                        = differentiation::DerivativeFunctor<differentiation::Forward, T_direction>;
+
+                public:
+                    /** Lower margin: we move by 1 along each direction and
+                     *  apply InternalDerivativeFunctor, add those up
+                     */
+                    using LowerMargin = typename pmacc::math::CT::add<
+                        typename pmacc::math::CT::make_Int<simDim, 1>::type,
+                        typename GetLowerMargin<InternalDerivativeFunctor>::type>::type;
+
+                    /** Upper margin: we move by 1 along each direction and
+                     *  effectively apply InternalDerivativeFunctor (for T_direction not
+                     *  literally, but structurally), add those up
+                     */
+                    using UpperMargin = typename pmacc::math::CT::add<
+                        typename pmacc::math::CT::make_Int<simDim, 1>::type,
+                        typename GetUpperMargin<InternalDerivativeFunctor>::type>::type;
+
+                    /** Return derivative value at the given point
+                     *
+                     * @tparam T_DataBox data box type with field data
+                     * @param data position in the data box to compute derivative at
+                     */
+                    template<typename T_DataBox>
+                    HDINLINE typename T_DataBox::ValueType operator()(T_DataBox const& data) const
+                    {
+                        constexpr uint32_t dir0 = T_direction;
+                        constexpr uint32_t dir1 = (dir0 + 1) % 3;
+                        constexpr uint32_t dir2 = (dir0 + 2) % 3;
+
+                        constexpr float_64 step[3] = {CELL_WIDTH, CELL_HEIGHT, CELL_DEPTH};
+                        constexpr float_64 step2[3] = {step[0] * step[0], step[1] * step[1], step[2] * step[2]};
+
+                        // delta: smallest step size
+                        constexpr float_64 delta = std::min({CELL_WIDTH, CELL_HEIGHT, CELL_DEPTH});
+
+                        constexpr float_64 delta2 = delta * delta;
+
+                        // r_i = delta^2 / (deltax_i)^2
+                        constexpr float_64 r[3] = {delta2 / step2[0], delta2 / step2[1], delta2 / step2[2]};
+
+                        constexpr float_64 r012 = r[0] * r[1] * r[2];
+                        constexpr float_64 rprodsum = r[0] * r[1] + r[1] * r[2] + r[2] * r[0];
+
+                        // equation (21)
+                        constexpr float_64 betaFactor = 0.125 * (1.0 - r012 / rprodsum);
+                        constexpr float_X betaDir1 = static_cast<float_X>(r[dir1] * betaFactor);
+                        constexpr float_X betaDir2 = static_cast<float_X>(r[dir2] * betaFactor);
+
+                        // equation (22)
+                        constexpr float_X gammaDir12 = static_cast<float_X>(
+                            r[dir1] * r[dir2] * (0.0625 - 0.125 * (r[dir1] * r[dir2] / rprodsum)));
+
+                        // from the condition alpha + 2 * betaDir1 + 2 * betaDir2 + 4 * gammaDir12 = 1
+                        constexpr float_X alpha
+                            = static_cast<float_X>(1.0 - 2.0 * betaDir1 - 2.0 * betaDir2 - 4.0 * gammaDir12);
+
+                        using Index = pmacc::DataSpace<simDim>;
+                        auto const upperNeighborDir1 = pmacc::math::basisVector<Index, dir1>();
+                        auto const upperNeighborDir2 = pmacc::math::basisVector<Index, dir2>();
+
+                        InternalDerivativeFunctor forwardDerivative
+                            = differentiation::makeDerivativeFunctor<differentiation::Forward, T_direction>();
+
+                        // smoothing operator applied after the derivative
+                        return alpha * forwardDerivative(data)
+                            + betaDir1 * forwardDerivative(data.shift(upperNeighborDir1))
+                            + betaDir1 * forwardDerivative(data.shift(-upperNeighborDir1))
+
+                            + betaDir2 * forwardDerivative(data.shift(upperNeighborDir2))
+                            + betaDir2 * forwardDerivative(data.shift(-upperNeighborDir2))
+
+                            + gammaDir12 * forwardDerivative(data.shift(upperNeighborDir1 + upperNeighborDir2))
+                            + gammaDir12 * forwardDerivative(data.shift(upperNeighborDir1 - upperNeighborDir2))
+                            + gammaDir12 * forwardDerivative(data.shift(-upperNeighborDir1 + upperNeighborDir2))
+                            + gammaDir12 * forwardDerivative(data.shift(-upperNeighborDir1 - upperNeighborDir2));
+                    }
+                };
+
+            } // namespace ckc
+        } // namespace maxwellSolver
+
+        namespace differentiation
+        {
+            namespace traits
+            {
+                template<uint32_t T_direction>
+                struct DerivativeFunctor<maxwellSolver::ckc::Derivative, T_direction>
+                    : pmacc::meta::accessors::Identity<maxwellSolver::ckc::DerivativeFunctor<T_direction>>
+                {
+                };
+
+            } // namespace traits
+        } // namespace differentiation
+    } // namespace fields
+} // namespace picongpu
--- a/include/picongpu/param/fieldSolver.param
+++ b/include/picongpu/param/fieldSolver.param
@@ -1,4 +1,5 @@
-/* Copyright 2013-2023 Axel Huebl, Heiko Burau, Rene Widera, Sergei Bastrakov, Klaus Steiniger
+/* Copyright 2013-2023 Axel Huebl, Heiko Burau, Rene Widera, Sergei Bastrakov,
+ *                     Klaus Steiniger, Lennert Sprenger
  *
  * This file is part of PIConGPU.
  *
@@ -41,8 +42,10 @@ namespace picongpu
         /** FieldSolver
          *
          * Field Solver Selection (note <> for some solvers), all in namespace maxwellSolver:
-         *  - Yee : Standard Yee solver approximating derivatives with respect to time and
+         *  - Yee: Standard Yee solver approximating derivatives with respect to time and
          * space by second order finite differences.
+         *  - CKC: Cole-Karkkainen-Cowan Solver, Dispersion free solver in the direction of the smallest
+         *  grid size.
          *  - Lehe<>: Num. Cherenkov free field solver in a chosen direction
          *  - ArbitraryOrderFDTD<4>: Solver using 4 neighbors to each direction to approximate
          * *spatial* derivatives by finite differences. The number of neighbors can be changed from 4 to any positive,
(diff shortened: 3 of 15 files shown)


include/picongpu/fields/MaxwellSolver/CCK/Derivative.def
```diff
@@ -0,0 +1,49 @@
...
+    namespace fields
+    {
+        namespace maxwellSolver
+        {
+            namespace cck
+            {
+                /** Tag for derivative used in the CCK solver
+                 *
+                 * Implements the modified derivative from
+                 *     Phys. Rev. ST Accel. Beams 16, 041303 (2013)
```
steindev: Add DOI pls.

include/picongpu/fields/MaxwellSolver/CCK/Derivative.hpp
```diff
@@ -0,0 +1,159 @@
...
+    namespace fields
+    {
+        namespace maxwellSolver
+        {
+            namespace cck
+            {
+                /** Functor for derivative used in the CCK solver
+                 *
+                 * Implements the modified derivative from
+                 *     Phys. Rev. ST Accel. Beams 16, 041303 (2013)
```
steindev: Add DOI

include/picongpu/param/fieldSolver.param
```diff
@@ -43,6 +43,8 @@ namespace picongpu
          * Field Solver Selection (note <> for some solvers), all in namespace maxwellSolver:
          *  - Yee : Standard Yee solver approximating derivatives with respect to time and
          * space by second order finite differences.
+         *  - CCK<>: Cowan-Cole-Karkkainen Solver, Dispersion free solver in the direction of the smallest
```
steindev: Correct spelling is 'Kärkkäinen'. But I don't know, whether umlauts are allowed by our coding guidelines.

include/picongpu/fields/MaxwellSolver/CCK/CCK.hpp
```diff
@@ -0,0 +1,98 @@
...
+                 *
+                 * @return value of 'X' to fulfill the condition 'c * dt <= X`
+                 */
+                float_X operator()() const
+                {
+                    // cellSize is not constexpr currently, so make an own constexpr array
+                    constexpr float_X step[3] = {CELL_WIDTH, CELL_HEIGHT, CELL_DEPTH};
+                    constexpr auto stepFreeDirection = step[T_cherenkovFreeDir];
+
+                    // Dependance on T_Defer is required, otherwise this check would have been enforced for each setup
```
steindev: Dependance -> Dependence

include/picongpu/param/fieldSolver.param
lennertsprenger: I wasn't sure about that either, so i left it out. But it can be added if its allowed

psychocoderHPC:
@lennertsprenger Please run clang-format-12 and push you changes again: https://picongpu.readthedocs.io/en/latest/dev/docs/COMMIT.html#format-code

include/picongpu/fields/MaxwellSolver/CCK/CCK.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+                    // Dependence on T_Defer is required, otherwise this check would have been enforced for each setup
+                    constexpr auto dt = getTimeStep();
+                    PMACC_CASSERT_MSG(
+                        Courant_Friedrichs_Lewy_condition_failure____check_your_grid_param_file,
+                        (SPEED_OF_LIGHT * dt) <= stepFreeDirection && sizeof(T_Defer*) != 0);
+
+                    // the cherenkov free direction has to be the smallest (or equal) step size
+                    PMACC_CASSERT_MSG(
+                        CFL_cherenkov_free_direction_smaller_grid_step_or_equal_to_all_others____check_your_grid_param_file,
```
steindev: Do you also need to do something with the `T_Defer` template parameter here?

psychocoderHPC: No there is no need because `T_cherenkovFreeDir` is a template parameter that is deferring the evaluation of the assert already.
The reason why `T_Defer` is required in the assert above is that the compiler is able to validate the assert even if the class will never be called. This can be avoided by making the evaluation depending on a template parameter.

include/picongpu/fields/MaxwellSolver/CCK/Derivative.hpp
```diff
@@ -0,0 +1,162 @@
...
+                        constexpr uint32_t dir0 = T_direction;
+                        constexpr uint32_t dir1 = (dir0 + 1) % 3;
+                        constexpr uint32_t dir2 = (dir0 + 2) % 3;
+
+                        constexpr float_64 step[3] = {CELL_WIDTH, CELL_HEIGHT, CELL_DEPTH};
+                        constexpr float_64 step2[3]
+                            = {CELL_WIDTH * CELL_WIDTH, CELL_HEIGHT * CELL_HEIGHT, CELL_DEPTH * CELL_DEPTH};
+
+                        // delta: smallest step size
+                        constexpr float_64 delta = step[freeDir];
```
steindev: I think you can make use of an `if constexpr` here fto deduce `delta` from the cell sizes and remove the `T_cherenkovFreeDirection` template argument.

include/picongpu/fields/MaxwellSolver/CCK/Derivative.hpp
```diff
@@ -0,0 +1,162 @@
...
+                        constexpr float_64 delta2 = delta * delta;
+
+                        // r_i = delta^2 / (deltax_i)^2
+                        constexpr float_64 r[3] = {delta2 / step2[0], delta2 / step2[1], delta2 / step2[2]};
+
+                        constexpr float_64 r012 = r[0] * r[1] * r[2];
+                        constexpr float_64 rprodsum = r[0] * r[1] + r[1] * r[2] + r[2] * r[0];
+
+                        // equation (21)
+                        constexpr float_X betaDir1 = static_cast<float_X>(r[dir1] * 0.125 * (1 - r012 / rprodsum));
```
steindev: Since you use the term in parenthesis twice, you can put it in a dedicated variable. Together with 0.125 factor.

include/picongpu/fields/MaxwellSolver/CCK/CCK.def
```diff
@@ -0,0 +1,47 @@
...
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/fields/MaxwellSolver/CCK/Derivative.def"
+#include "picongpu/fields/MaxwellSolver/FDTD/FDTD.def"
+#include "picongpu/fields/differentiation/Curl.def"
+
+//#include <pmacc/math/Vector.hpp>
```
steindev: remove

include/picongpu/fields/MaxwellSolver/CCK/CCK.def
steindev, line 1: Please add a little description that saying that this implements the CCK solver and reference the original paper. You can also point out that it uses a customized `Derivative` definition. Name the CFL condition for this solver.

include/picongpu/fields/MaxwellSolver/CCK/CCK.hpp
```diff
@@ -0,0 +1,92 @@
...
+                    // cellSize is not constexpr currently, so make an own constexpr array
+                    constexpr float_X step[3] = {CELL_WIDTH, CELL_HEIGHT, CELL_DEPTH};
+                    constexpr float_X cdt = SPEED_OF_LIGHT * getTimeStep(); // c * dt
+
+                    // Dependence on T_Defer is required, otherwise this check would have been enforced for each setup
+                    PMACC_CASSERT_MSG(
+                        Courant_Friedrichs_Lewy_condition_failure____check_your_grid_param_file,
+                        (cdt <= step[0]) && (cdt <= step[1]) && (cdt <= step[2]) && sizeof(T_Defer*) != 0);
+
+                    return step[0]; // TODO: return correct value for X
```
steindev: Process ToDo

include/picongpu/fields/MaxwellSolver/CCK/CCK.hpp
```diff
@@ -0,0 +1,92 @@
...
+    namespace traits
+    {
+        template<>
+        struct StringProperties<::picongpu::fields::maxwellSolver::CCK>
+        {
+            static StringProperty get()
+            {
+                auto propList = ::picongpu::fields::maxwellSolver::CCK::getStringProperties();
+                // overwrite the name of the solver (inherit all other properties)
+                propList["name"].value = "CCK";
```
steindev: Please use the name 'CK', which is defined by the openPMD standard (https://github.com/openPMD/openPMD-standard/blob/latest/EXT_ED-PIC.md#additional-attributes-for-the-group-meshespath)

include/picongpu/fields/MaxwellSolver/CCK/Derivative.hpp
```diff
@@ -0,0 +1,164 @@
...
+                    {
+                        constexpr uint32_t dir0 = T_direction;
+                        constexpr uint32_t dir1 = (dir0 + 1) % 3;
+                        constexpr uint32_t dir2 = (dir0 + 2) % 3;
+
+                        constexpr float_64 step[3] = {CELL_WIDTH, CELL_HEIGHT, CELL_DEPTH};
+                        constexpr float_64 step2[3] = {step[0] * step[0], step[1] * step[1], step[2] * step[2]};
+
+                        // delta: smallest step size
+                        constexpr float_64 delta = (step[0] < step[1]) ? ((step[0] < step[2]) ? step[0] : step[2])
```
steindev: please use `std::min()`

include/picongpu/param/fieldSolver.param
```diff
@@ -43,6 +43,8 @@ namespace picongpu
...
+         *  grid size (has to be the y-direction, CELL_HEIGHT<=CELL WIDTH && CELL_HEIGHT<=CELL_DEPTH)
```
steindev: ToDo: remove limitation

include/picongpu/param/fieldSolver.param
steindev, line 1: Please add the CCK solver to **all** `fieldSolver.param`s

steindev (changes requested):

steindev:
@lennertsprenger Please also add the (figures of the) test results to the PR description.

PrometheusPi:
@lennertsprenger what is the status of this PR?

lennertsprenger:
## How the CCK solver was tested:

As a difference to the standard Yee solver the CCK solver has no numerical dispersion along the axis with the smallest grid size, when using `dt = delta/c`, where delta is the smallest grid size in the simulation. The numerical dispersion results in light waves propagating with a velocity error based on the wavelength compared to the speed of light in a vacuum

The numerical dispersion is a numerical effect arising from the field Solver on the simulation grid. More information on the numerical dispersion of the solver can be found here: [numerical dispersion Yee (readthedocs)](https://picongpu.readthedocs.io/en/latest/models/AOFDTD.html#dispersion-relation-of-light-waves-on-a-mesh) and [numerical dispersion CCK (Phys. Rev. ST Accel. Beams 16, 041303 (2013))](https://dx.doi.org/10.1103/PhysRevSTAB.16.041303)

The first test case is a short current pulse in the center of the simulation domain. The results are shown side-by-side, left: Yee, right: CCK. The simulations were identical with a cubic grid except for the fieldSolver and the time step. The time step for the CCK Solver was chosen as $\Delta t_{CCK} = 0.9999 \cdot \Delta y / c$ and $\Delta t_{Yee} = 0.5 \Delta t_{CCK}$. The optimal time step for the Yee solver would have been $1/\sqrt{3}\Delta t_{CCK}$ but 0.5 was chosen for a better comparison between the two solvers. The axis show the cells in x and y direction.
yee_cck6

***

Another test case was laser propagation, where a Gaussian Pulse propagated with the Yee and the CCK solver. The laser used was simulated as a 800 nm pulse with a spatial resolution of 10 cells per wavelength and time step choosen the same way as the test above.
The following pictures show the electric field (x-polarized laser) along the propagation axis through the center of the pulse in the moving window frame of reference (light should be standing still) at different timesteps showing the same physical time.

LaserYee
Looking at the laser we see the laser being slower than the moving window. And also broadening of the envelope due to the different wavelength that have a different numerical phase velocity. This shows the effect of the numerical phase velocity along on of the simulation axis for the Yee solver.

LaserCCK
The laser pulse propagate with the CCK solver stays roughly at the same position in the moving window as expected.

***

The tests have shown the better numerical dispersion along the axis with the smallest grid size. While the first test example also shows the not optimal dispersion along the 45° propagation axis with both solvers.
The effect of numerical dispersion could be further reduced by choosing a finer spatial sampling of the pulse, because the resolution of cells per wavelength determines the error in the numerical phase velocity.

lennertsprenger:
I have now added the description and results of the tests I have done with the solver and the PR should now be ready for review.
@steindev

include/picongpu/param/fieldSolver.param
steindev: ping @psychocoderHPC: umlauts, or using 'ae', or just 'a'?

include/picongpu/fields/MaxwellSolver/CCK/Derivative.hpp
```diff
@@ -0,0 +1,161 @@
...
+                        constexpr float_64 delta2 = delta * delta;
+
+                        // r_i = delta^2 / (deltax_i)^2
+                        constexpr float_64 r[3] = {delta2 / step2[0], delta2 / step2[1], delta2 / step2[2]};
+
+                        constexpr float_64 r012 = r[0] * r[1] * r[2];
+                        constexpr float_64 rprodsum = r[0] * r[1] + r[1] * r[2] + r[2] * r[0];
+
+                        // equation (21)
+                        constexpr float_64 betaFactor = 0.125 * (1 - r012 / rprodsum);
```
steindev: Please use `1.0` instead of `1`

steindev (changes requested):
Great! Just one minor comment left!

include/picongpu/fields/MaxwellSolver/CCK/Derivative.def
```diff
@@ -0,0 +1,47 @@
...
+                 *   https://dx.doi.org/10.1103/PhysRevSTAB.16.041303
+                 *
+                 * @tparam T_cherenkovFreeDirection direction to remove numerical Cherenkov
```
steindev: This comment can be removed, no?

include/picongpu/fields/MaxwellSolver/CCK/Derivative.hpp
psychocoderHPC: Background. If you write `1` it is an integer, so we should avoid casts, in this case `1` and `1.0` are equal because `1.0` can be represented in floating point but always better to use the correct type.

include/picongpu/fields/MaxwellSolver/CCK/Derivative.def
lennertsprenger: removed the outdated line

include/picongpu/param/fieldSolver.param
lennertsprenger: ping @psychocoderHPC

psychocoderHPC:
@steindev The first task for you in 2024.

ax3l:
Nice work!

Just FYI, this solver type is usually abbreviated CKC :)

steindev:
> Nice work!
> 
> Just FYI, this solver type is usually abbreviated CKC :)

Yes, that seems to be the case. So we should rename it in PIConGPU, @lennertsprenger

steindev:
Apart from this last change I am happy with the PR. :partying_face:
So you can directly rebase and squash befor your next commit.

include/picongpu/param/fieldSolver.param
steindev: Only plain ASCII, no umlauts. So I guess we should use standard ä -> ae replacement.

steindev: Ohh I just see, that openPMD standard used the ä -> a replacement. Well, than lets just follow their scheme in order to have consistent naming. Since you used that form already, I will mark this thread as resolved.

lennertsprenger:
changed the name and squashed. @steindev

steindev (dismissed):

include/picongpu/fields/MaxwellSolver/CKC/CKC.hpp
```diff
@@ -0,0 +1,90 @@
...
+            /** Specialization of the CFL condition checker for CKC solver
+             *
+             * @tparam T_Defer technical parameter to defer evaluation
+             */
+            template<typename T_Defer>
+            struct CFLChecker<CKC, T_Defer>
+            {
+                /** Check the CFL condition according to the paper, doesn't compile when failed
+                 *
+                 * @return value of 'X' to fulfill the condition 'c * dt <= X'
```
chillenzer: It was hard to understand for me what `X` is. Only after reading the code, I understand what the return value is. I would rephrase it into something like this
```suggestion
                 * @return shortest edge length of the cell (necessarily fulfilling `>= c * dt` because CFL is fulfilled if this compiled at all)
```

include/picongpu/fields/MaxwellSolver/CKC/CKC.hpp
```diff
@@ -0,0 +1,90 @@
...
+    namespace traits
+    {
+        template<>
+        struct StringProperties<::picongpu::fields::maxwellSolver::CKC>
+        {
+            static StringProperty get()
+            {
+                auto propList = ::picongpu::fields::maxwellSolver::CKC::getStringProperties();
+                // overwrite the name of the solver (inherit all other properties)
+                propList["name"].value = "CK";
```
chillenzer, line 84: This is probably supposed to say `"CKC"` (missing last `C`)?

include/picongpu/fields/MaxwellSolver/CKC/Derivative.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+                    /** Upper margin: we move by 1 along each direction and
+                     *  effectively apply InternalDerivativeFunctor (for T_direction not
+                     *  literally, but structurally), add those up
+                     */
+                    using UpperMargin = typename pmacc::math::CT::add<
+                        typename pmacc::math::CT::make_Int<simDim, 1>::type,
+                        typename GetUpperMargin<InternalDerivativeFunctor>::type>::type;
+
+                    HDINLINE DerivativeFunctor()
```
chillenzer: Why does this have a user-defined ctor at all? Is this about the `HDINLINE` annotation?

include/picongpu/fields/MaxwellSolver/CKC/Derivative.hpp
```diff
@@ -0,0 +1,161 @@
...
+                    /** Return derivative value at the given point
+                     *
+                     * @tparam T_DataBox data box type with field data
+                     * @param data position in the data box to compute derivative at
+                     */
+                    template<typename T_DataBox>
+                    HDINLINE typename T_DataBox::ValueType operator()(T_DataBox const& data) const
+                    {
+                        constexpr uint32_t dir0 = T_direction;
+                        constexpr uint32_t dir1 = (dir0 + 1) % 3;
```
chillenzer, line 89: Sorry for my ignorance but does the hardcoded 3 in here harmonise with having a CT-defined `simDim`?

include/picongpu/fields/MaxwellSolver/CKC/CKC.hpp
```diff
@@ -0,0 +1,90 @@
...
+            struct CFLChecker<CKC, T_Defer>
+            {
+                /** Check the CFL condition according to the paper, doesn't compile when failed
+                 *
+                 * @return value of 'X' to fulfill the condition 'c * dt <= X'
+                 */
+                float_X operator()() const
+                {
+                    // cellSize is not constexpr currently, so make an own constexpr array
+                    constexpr float_X step[3] = {CELL_WIDTH, CELL_HEIGHT, CELL_DEPTH};
```
chillenzer, line 55: Why is this called `step`?

include/picongpu/fields/MaxwellSolver/CKC/CKC.hpp
```diff
@@ -0,0 +1,90 @@
...
+            template<typename T_Defer>
+            struct CFLChecker<CKC, T_Defer>
+            {
+                /** Check the CFL condition according to the paper, doesn't compile when failed
+                 *
+                 * @return value of 'X' to fulfill the condition 'c * dt <= X'
+                 */
+                float_X operator()() const
+                {
+                    // cellSize is not constexpr currently, so make an own constexpr array
```
chillenzer, line 54: For my information, @psychocoderHPC: Why's that? I've traced it back to these two lines: [CPU](https://github.com/ComputationalRadiationPhysics/picongpu/blob/5456d097efc3c6e60bb395e11ff452a3b280963e/include/pmacc/math/ConstVector.hpp#L109) and [GPU](https://github.com/ComputationalRadiationPhysics/picongpu/blob/5456d097efc3c6e60bb395e11ff452a3b280963e/include/pmacc/math/ConstVector.hpp#L93). The CPU version is obviously `constexpr`. My initial guess was that `constexpr` might not be allowed in CUDA code but I've found recent examples of that. Is it support for old CUDA versions or an oversight from my side?

chillenzer (changes requested):
Hi,

Thanks for your contribution. Looks like a nice and clean implementation. Thanks in particular for taking all the time to update all those lists in the examples. As you can see from #4813, I'm not happy with the necessity of that but as long as it is like that, thanks!

I've got a few comments below. Often rather for my understanding than for actual change. My main question would be: How did you produce these super cool plots above? Would it be possible to automate their generation as a test case and put that into the project, too? Even better, we could make that a fully automated end-to-end test by checking the (approximate) constancy of the wave packages. Would that be still in scope?

include/picongpu/fields/MaxwellSolver/CKC/CKC.hpp
BrianMarre: statement is no longer true, but used to be before we switched to C++17 allowing constexpr floats.
see [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/5456d097efc3c6e60bb395e11ff452a3b280963e/include/picongpu/unitless/grid.unitless#L30-L32)

should be changed

include/picongpu/fields/MaxwellSolver/CKC/CKC.hpp
lennertsprenger: It was called `X` since it was phrased like this in other solvers (e.g.[Yee Solver](https://github.com/ComputationalRadiationPhysics/picongpu/blob/5456d097efc3c6e60bb395e11ff452a3b280963e/include/picongpu/fields/MaxwellSolver/Yee/Yee.hpp#L48)) because i oriented myself on that. In general `X` is just a value dependent on the solver and the grid. It is just in this special implementation that it is the shortest edge length.
So this could also be a new general discussion on how to phrase it in the corresponding CFL conditions for the different solvers.

include/picongpu/fields/MaxwellSolver/CKC/CKC.hpp
lennertsprenger: It is supposed to say only `"CK"` as it is defined in the [openPMD-standard](https://github.com/openPMD/openPMD-standard/blob/latest/EXT_ED-PIC.md#additional-attributes-for-the-group-meshespath) and already discussed here https://github.com/ComputationalRadiationPhysics/picongpu/pull/4661#discussion_r1365402074

include/picongpu/fields/MaxwellSolver/CKC/CKC.hpp
lennertsprenger: It was called `step` because it shows how much you step in a direction when you go one cell and because it is oriented on the [Lehe solver](https://github.com/ComputationalRadiationPhysics/picongpu/blob/5456d097efc3c6e60bb395e11ff452a3b280963e/include/picongpu/fields/MaxwellSolver/Lehe/Derivative.hpp) implementation

lennertsprenger:
Hi @chillenzer,
the plots are generated with python notebooks from the simulation output. If you are interested in more detail about the simulations i can explain them more.
Theoretically it would be possible to automate the tests.

include/picongpu/fields/MaxwellSolver/CKC/CKC.hpp
chillenzer: Okay, I give you that. I still think that this is badly phrased (also in the Yee case) but you're off the hook. I've created #4815 to track this. Please come back to this and adjust if my suggestion gets merged before your PR. Otherwise, you don't have to act on this.

include/picongpu/fields/MaxwellSolver/CKC/CKC.hpp
chillenzer: Thanks for the explanation!

include/picongpu/fields/MaxwellSolver/CKC/CKC.hpp
chillenzer: Thanks! I still don't think that this is a very intuitive name but if that's the jargon of this domain, that's fine with me.

include/picongpu/fields/MaxwellSolver/CKC/CKC.hpp
chillenzer: Thanks! Opened #4816 to track this.

chillenzer:
Thanks @lennertsprenger! I would strongly advise to add them to the repo in an automated and reproducible form (including a simulation run to produce them, basically the equivalent of what these two PRs do: #4801 and #4557) and would request this as a further change to the PR. I would however give @steindev or @psychocoderHPC the chance to set other priorities or split this into multiple PRs.

include/picongpu/fields/MaxwellSolver/CKC/Derivative.hpp
lennertsprenger: I think this can be removed

lennertsprenger:
I'd have to adjust and check the scripts to make it an automated test. And I think that should be a new PR to add them

include/picongpu/fields/MaxwellSolver/CKC/Derivative.hpp
chillenzer: Then, please do so.

chillenzer:
Personally, I would strongly advise against splitting functional code from its tests but given the offline discussed time constraints, I can see that such a refactoring might be out of scope in this particular case. Please still make sure that one of us has access to the scripts (no matter what state they are in -- of course, the cleaner the better).

So, once you've removed the empty constructor, you've got my approval.

include/picongpu/fields/MaxwellSolver/CKC/Derivative.hpp
lennertsprenger: This it at this point at least correct for 2d simulations. It should also work in 2d in general. I have done a 2d simulation and it looks correct.

lennertsprenger:
Should now be ready to merge.
Removed the empty constructor and also redone the wording of the CFL checker since #4824 has been merged.

@steindev @chillenzer

chillenzer:
Sounds good. Fine with me once the CI is through.

chillenzer (approved):

steindev (approved):

steindev:
Will merge, since there are no more reviews expected.

PR #4670, opened 2023-09-07
add support for CMake option `alpaka_DISABLE_VENDOR_RNG`

psychocoderHPC:
Add support for alpaka to disable the vendor random number generators. For PIConGPU the user should change the default `XorMin` RNG in `tandom.param` to `AlpakaRAnd` if `alpaka_DISABLE_VENDOR_RNG` is set to `ON`.

#4672 is required that this feature is working but there is no need that #4672 is merged before this PR.

### DIFF
--- a/include/pmacc/random/distributions/normal/Normal_double.hpp
+++ b/include/pmacc/random/distributions/normal/Normal_double.hpp
@@ -21,15 +21,16 @@
 
 #pragma once
 
-#include "pmacc/algorithms/math.hpp"
-#include "pmacc/random/distributions/Normal.hpp"
-#include "pmacc/random/distributions/Uniform.hpp"
-#include "pmacc/random/distributions/misc/MullerBox.hpp"
-#include "pmacc/random/methods/MRG32k3aMin.hpp"
-#include "pmacc/random/methods/XorMin.hpp"
-#include "pmacc/types.hpp"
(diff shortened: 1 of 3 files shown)


include/pmacc/random/distributions/normal/Normal_double.hpp
```diff
@@ -56,8 +57,10 @@ namespace pmacc
...
                 {
                 };
-#endif
+#    endif
             } // namespace detail
         } // namespace distributions
     } // namespace random
 } // namespace pmacc
+
+#endif
```
PrometheusPi: please add newline

include/pmacc/random/distributions/normal/Normal_float.hpp
```diff
@@ -56,8 +58,10 @@ namespace pmacc
...
                 {
                 };
-#endif
+#    endif
             } // namespace detail
         } // namespace distributions
     } // namespace random
 } // namespace pmacc
+
+#endif
```
PrometheusPi: same here

PrometheusPi (changes requested):
minor style issues

include/pmacc/random/distributions/normal/Normal_float.hpp
psychocoderHPC: done

PrometheusPi (approved):

PR #4674, opened 2023-09-08
Non-Gaussian temporal envelope for Gaussian beam

pordyna:
## Current status
The current implementation of the `GaussianBeam` profile uses a Gaussian temporal envelope. On the other hand, the `ExpRampWithPrepulse`, that can be used for a more elaborate contrast curves, is also a Gaussian in the transversal plane, but ignores wavefront curvature, and so it is implicitly setting the focus of the Gaussian beam on the initialization plane. At the moment, it is not possible to have a focusing Gaussian beam that includes a contrast curve. 

The major difference between the two implementations is that `GaussianBeam` does not use the cell position for the temporal envelope, but rather the crossing point of the wave front surface (that goes through the cell) and the beam axis.

## Proposed changes
Generalize `GaussianBeam` temporal envelope definition and make it possible to use the `ExpRampWithPrepulse` envelope in `GaussianBeam`.  

## TODO:
- [x] Separate envelope calculation in `ExpRampWithPrepulse` from the rest of the functor.
- [x] Refractor the Gaussian pulse out of  `GaussianBeam`  and include it as a new template argument. 
- [x] Make an example using the `ExpRampWithPrepulse` envelope instead of the Gaussian in  `GaussianBeam`.
- [x] Tests:
  - [x]  verify `ExpRampWithPrepulse`
  - [x]  verify `GaussianBeam` with a Gaussian Profile
  - [x] verify `GaussianBeam` with a `ExpRampWithPrepulse`
- [ ] documentation
  - [ ] basic usage in `incidentField.param`
  - [ ] docstrings
- [x] PICMI integration (doesn't break PICMI, adding the new profile could be a separate PR?)

* (probably not in this PR) remove `ExpRampWithPrepulse` profile, just keep the envelope.

@steindev @wmles What do you think, is this going to work, or am I missing sth?

### DIFF
--- a/include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
+++ b/include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
@@ -191,7 +191,7 @@ namespace picongpu::fields::incidentField
                 {
                     auto const runTimeShifted = time + Unitless::time_start_init;
                     auto const phase = Unitless::w * runTimeShifted + Unitless::LASER_PHASE + phaseShift;
-                    return math::sin(phase) * Unitless::AMPLITUDE * Envelope::getEnvelope(runTimeShifted);
+                    return math::cos(phase) * Unitless::AMPLITUDE * Envelope::getEnvelope(runTimeShifted);
                 }
             };
         } // namespace detail
--- a/include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
+++ b/include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
@@ -44,6 +44,7 @@ namespace picongpu::fields::incidentField
             }
         };
 
+
         namespace detail
         {
             /** Unitless exponential ramp with prepulse parameters (envelope part)
@@ -113,8 +114,9 @@ namespace picongpu::fields::incidentField
                     "pulse "
                     "would give more than half of the pulse amplitude. This is not a Gaussian pulse at all "
                     "anymore - probably some of the parameters are different from what you think!?");
-                static constexpr float_X INIT_TIME
-                    = static_cast<float_X>((TIME_PEAKPULSE + Params::RAMP_INIT * Base::PULSE_DURATION) / UNIT_TIME);
+                //                static constexpr float_X INIT_TIME
+                //                    = static_cast<float_X>((TIME_PEAKPULSE + Params::RAMP_INIT *
+                //                    Base::PULSE_DURATION) / UNIT_TIME);
 
                 /* a symmetric pulse will be initialized at generation plane for
                  * a time of RAMP_INIT * PULSE_DURATION + LASER_NOFOCUS_CONSTANT = INIT_TIME.
@@ -136,97 +138,6 @@ namespace picongpu::fields::incidentField
             {
                 //! User SI parameters
                 using Params = T_Params;
-            };
-
-            template<typename T_Params>
-            struct ExpRampWithPrepulseLongitudinal : public ExpRampWithPrepulsesLongitudinalUnitless<T_Params>
-            {
-            public:
-                using Unitless = ExpRampWithPrepulsesLongitudinalUnitless<T_Params>;
-                HDINLINE static float_X getEnvelope(float_X const runTime)
-                {
-                    constexpr auto int_ratio_prepule = Unitless::INT_RATIO_PREPULSE;
-                    constexpr auto int_ratio_point_1 = Unitless::INT_RATIO_POINT_1;
-                    constexpr auto int_ratio_point_2 = Unitless::INT_RATIO_POINT_2;
-                    constexpr auto int_ratio_point_3 = Unitless::INT_RATIO_POINT_3;
-                    auto const AMP_PREPULSE = math::sqrt(int_ratio_prepule);
-                    auto const AMP_1 = math::sqrt(int_ratio_point_1);
-                    auto const AMP_2 = math::sqrt(int_ratio_point_2);
-                    auto const AMP_3 = math::sqrt(int_ratio_point_3);
-
-                    auto env = 0.0_X;
-                    bool const before_preupramp = runTime < Unitless::time_start_init;
-                    bool const before_start = runTime < Unitless::TIME_1;
-                    bool const before_peakpulse = runTime < Unitless::endUpramp;
-                    bool const during_first_exp = (Unitless::TIME_1 < runTime) && (runTime < Unitless::TIME_2);
-                    bool const after_peakpulse = Unitless::startDownramp <= runTime;
-
-                    if(before_preupramp)
-                        env = 0._X;
-                    else if(before_start)
-                    {
-                        env = AMP_1 * gauss(runTime - Unitless::TIME_1);
-                    }
-                    else if(before_peakpulse)
-                    {
-                        float_X const ramp_when_peakpulse
-                            = extrapolateExpo(Unitless::TIME_2, AMP_2, Unitless::TIME_3, AMP_3, Unitless::endUpramp);
-
-                        // This check exists in original laser, but can't print from device
-                        // if(ramp_when_peakpulse > 0.5_X)
-                        //{
-                        //    log<picLog::PHYSICS>(
-                        //        "Attention, the intensities of the laser upramp are very large! "
-                        //        "The extrapolation of the last exponential to the time of "
-                        //        "the peakpulse gives more than half of the amplitude of "
-                        //        "the peak Gaussian. This is not a Gaussian at all anymore, "
-                        //        "and physically very unplausible, check the params for misunderstandings!");
-                        //}
-
-                        env += (1._X - ramp_when_peakpulse) * gauss(runTime - Unitless::endUpramp);
-                        env += AMP_PREPULSE * gauss(runTime - Unitless::TIME_PREPULSE);
-                        if(during_first_exp)
-                            env += extrapolateExpo(Unitless::TIME_1, AMP_1, Unitless::TIME_2, AMP_2, runTime);
-                        else
-                            env += extrapolateExpo(Unitless::TIME_2, AMP_2, Unitless::TIME_3, AMP_3, runTime);
-                    }
-                    else if(!after_peakpulse)
-                        env = 1.0_X;
-                    else // after startDownramp
-                        env = gauss(runTime - Unitless::startDownramp);
-                    return env;
-                }
-
-                HINLINE static std::string getName()
-                {
-                    return "ExpRampWithPrepulse";
-                }
-
-            private:
-                /** takes time t relative to the center of the Gaussian and returns value
-                 * between 0 and 1, i.e. as multiple of the max value.
-                 * use as: amp_t = amp_0 * gauss( t - t_0 )
-                 */
-                HDINLINE static float_X gauss(float_X const t)
-                {
-                    auto const exponent = t / Unitless::PULSE_DURATION;
-                    return math::exp(-0.25_X * exponent * exponent);
-                }
-
-                /** get value of exponential curve through two points at given t
-                 * t/t1/t2 given as float_X, since the envelope doesn't need the accuracy
-                 */
-                HDINLINE static float_X extrapolateExpo(
-                    float_X const t1,
-                    float_X const a1,
-                    float_X const t2,
-                    float_X const a2,
-                    float_X const t)
-                {
-                    auto const log1 = (t2 - t) * math::log(a1);
-                    auto const log2 = (t - t1) * math::log(a2);
-                    return math::exp((log1 + log2) / (t2 - t1));
-                }
             };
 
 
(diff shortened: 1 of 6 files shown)


pordyna:
If we go through with this, we should rename `GaussianPulse` back to `GaussianBeam`.  Until now, it was both, with this PR it will be a `GaussianBeam` with a configurable pulse shape.

pordyna:
The number of changed lines is so high because I used the opportunity to switch to concatenated namespaces.

pordyna:
Note: All of the `time_start_init`, `TIME_INIT`, `RAMP_INIT`, `PULSE_INIT`s are a bit of chaos, need to find a way to use it consistently.  Maybe the longitudinal envelope class should define `INIT_TIME?` Right now `INIT_TIME` is not used in ExpRamp, but time_start_init is,. The two have different meaning.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -34,328 +34,345 @@
...
+            };
+
+            /** Helper type to check if T_Params has members TILT_AXIS_1_SI and TILT_AXIS_2_SI.
+             *
+             * Is void for those types, ill-formed otherwise.
+             *
+             * @tparam T_Params user (SI) parameters
+             */
+            template<typename T_Params>
+            using HasTilt = std::void_t<decltype(T_Params::TILT_AXIS_1_SI + T_Params::TILT_AXIS_2_SI)>;
```
PrometheusPi, line 65: What does this do and what happens if `TILT_AXIS_1_SI` ≠ 0 and `TILT_AXIS_2_SI` ≠ 0 but the sum =0? 

Above you define the variable without `_SI` here you use it with `_SI`. Where do you define the later one?

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -34,328 +34,345 @@
...
+                    auto const phase
+                        = Unitless::w * (time - mue - focusPos / SPEED_OF_LIGHT) + Unitless::LASER_PHASE + phaseShift;
+
+                    // Apply tilt if needed
+                    if constexpr(Unitless::TILT_AXIS_1 || Unitless::TILT_AXIS_2)
+                    {
+                        auto const tiltTimeShift = phase / Unitless::w + focusPos / SPEED_OF_LIGHT;
+                        auto const tiltPositionShift = SPEED_OF_LIGHT * tiltTimeShift
+                            / pmacc::math::dot(this->getDirection(), float3_X{cellSize});
+                        auto const tilt1 = Unitless::TILT_AXIS_1;
```
PrometheusPi, line 242: why do you do this extra declaration of a variable? Better readability?

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -34,328 +34,345 @@
...
+                        pos[1] += math::tan(tilt1) * tiltPositionShift;
+                        auto const tilt2 = Unitless::TILT_AXIS_2;
```
PrometheusPi, line 244: same here

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -34,328 +34,345 @@
...
+        template<typename T_Params, typename T_LongitudinalEnvelope>
+        struct GaussianPulse
+        {
+            using LongitudinalEnvelope = T_LongitudinalEnvelope;
+            //! Get text name of the incident field profile
+            HINLINE static std::string getName()
+            {
+                // This template is used for both Gaussian and PulseFrontTilt, distinguish based on tilt value
+                using TiltParam = detail::TiltParam<T_Params>;
+                bool isTilted = (std::abs(TiltParam::TILT_AXIS_1) + std::abs(TiltParam::TILT_AXIS_2) > 0);
```
PrometheusPi, line 355: you could use your smarter method from above 
```suggestion
                bool isTilted = TiltParam::TILT_AXIS_1 || TiltParam::TILT_AXIS_2;

```

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
```diff
@@ -30,276 +30,290 @@
...
+                static constexpr float_X TIME_3 = static_cast<float_X>(Params::TIME_POINT_3_SI / UNIT_TIME);
+                static constexpr float_X endUpramp = TIME_PEAKPULSE - 0.5_X * LASER_NOFOCUS_CONSTANT;
+                static constexpr float_X startDownramp = TIME_PEAKPULSE + 0.5_X * LASER_NOFOCUS_CONSTANT;
+
+
+                // compile-time checks for physical sanity:
+                static_assert(
+                    (TIME_1 < TIME_2) && (TIME_2 < TIME_3) && (TIME_3 < endUpramp),
+                    "The times in the parameters TIME_POINT_1/2/3 and the beginning of the plateau (which is "
+                    "at "
```
PrometheusPi, line 79: strange line break

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
```diff
@@ -30,276 +30,290 @@
...
+                    "TIME_PEAKPULSE - 0.5*RAMP_INIT*PULSE_DURATION) should be in ascending order");
+
+                // some prerequisites for check of intensities (approximate check, because I can't use exp and
+                // log)
+                static constexpr float_X ratio_dt
+                    = (endUpramp - TIME_3) / (TIME_3 - TIME_2); // ratio of time intervals
+                static constexpr float_X ri1
+                    = Params::INT_RATIO_POINT_3 / Params::INT_RATIO_POINT_2; // first intensity ratio
```
PrometheusPi, line 87: Could you please add a compile time assert to check if `INT_RATIO_POINT_2` ≠ 0 (or even >0)?

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
```diff
@@ -30,276 +30,290 @@
...
+                static constexpr float_X ri2
+                    = 0.2_X / Params::INT_RATIO_POINT_3; // second intensity ratio (0.2 is an arbitrary upper
```
PrometheusPi, line 89: same here

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
```diff
@@ -30,276 +30,290 @@
...
+            HDINLINE static float_X getEnvelope(float_X const runTime)
+            {
+                constexpr auto int_ratio_prepule = Unitless::INT_RATIO_PREPULSE;
+                constexpr auto int_ratio_point_1 = Unitless::INT_RATIO_POINT_1;
+                constexpr auto int_ratio_point_2 = Unitless::INT_RATIO_POINT_2;
+                constexpr auto int_ratio_point_3 = Unitless::INT_RATIO_POINT_3;
+                auto const AMP_PREPULSE = math::sqrt(int_ratio_prepule);
+                auto const AMP_1 = math::sqrt(int_ratio_point_1);
+                auto const AMP_2 = math::sqrt(int_ratio_point_2);
+                auto const AMP_3 = math::sqrt(int_ratio_point_3);
```
PrometheusPi, line 216: should `int_ratio_point_...` be checked for >= 0? 
I do not think this is necessary because then the `AMP_...` just becomes `nan`.

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
```diff
@@ -30,276 +30,290 @@
...
+            using Unitless = detail::ExpRampWithPrepulsesLongitudinalUnitless<T_Params>;
+
+            static constexpr float_X TIME_SHIFT = Unitless::time_start_init;
+
+            HDINLINE static float_X getEnvelope(float_X const runTime)
+            {
+                constexpr auto int_ratio_prepule = Unitless::INT_RATIO_PREPULSE;
+                constexpr auto int_ratio_point_1 = Unitless::INT_RATIO_POINT_1;
+                constexpr auto int_ratio_point_2 = Unitless::INT_RATIO_POINT_2;
+                constexpr auto int_ratio_point_3 = Unitless::INT_RATIO_POINT_3;
```
PrometheusPi, line 212: same question as before: why this renaming of variables?

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
```diff
@@ -30,276 +30,290 @@
...
-                        typename GetFunctorIncidentE<profiles::ExpRampWithPrepulse<T_Params>>::type>;
-                };
-            } // namespace detail
-        } // namespace incidentField
-    } // namespace fields
-} // namespace picongpu
+                    float_X const ramp_when_peakpulse
+                        = extrapolateExpo(Unitless::TIME_2, AMP_2, Unitless::TIME_3, AMP_3, Unitless::endUpramp);
+
+                    // This check exists in original laser, but can't print from device
```
PrometheusPi, line 236: what should we do with this commented-out code?

PrometheusPi (changes requested):
This is a first look at the PR (as requested by @psychocoderHPC) and not a thorough review. 
I just had a few remarks and many questions.

pordyna:
@PrometheusPi So we should of course look at that, but all of these comments are about the parts of code that I was only moving around.

PrometheusPi:
@pordyna I thought so.  Thanks for the clarification. Then, I would say that this should not be changed with your pull request but in a followup pull request.

pordyna:
## Testing
@steindev @PrometheusPi @psychocoderHPC @lennertsprenger @BrianMarre  could you take a look at this? I think we can say, the new changes are as correct as PIConGPU was before^^, but I have definitely seen some weird stuff here, so the question is, is this expected to happen or not.  
### Automatic tests
So I tried comparing the laser profiles with python generated values. I also tried comparing output from different versions. In the result, this seems to be really hardy, since I always get a few cells with very large relative error (>1) that are also at points where the expected value is not 0 (where absolute error comparison would be more suitable).  After briefly discussing this with @BrianMarre we came to the conclusion that this is most likely due to some solver artifacts. This is a problem even when comparing two PIC outputs, as soon as the generated values are not exactly the same. So, e.g., I factored out the max amplitude out of the `ExpRampWithPrepulse` calculation and already the comparison with the previous `ExpRampWithPrepulse` is showing those kinds of artifacts.  That means I won't be able to introduce a CI test for laser profiles with that PR, as those artifacts would need to be addressed somehow.  


### Verifying this PR
I did the following comparisons:
* `GaussianPulse` as it was before (Gaussian beam and Gaussian pulse) between dev and my branch. Perfect match when time for the envelope is calculated exactly as before. With the current simplification in time calculation: https://github.com/ComputationalRadiationPhysics/picongpu/blob/f6753273597ccd6e5f829433fb96b19a871b6bb0/include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp#L266 the numbers are not exact, but very similar excluding the artifacts. I'have seen in a debugger that the time calculation varies sometime by a value comparable with the machine epsilon. 
* `ExpRampWithPrepulse` as it was before, my branch with a branch that only contains the factoring out of the amplitude, resulting in a perfect match. When comparing with `dev` the resulting relative error includes the artifacts, but it is still in a very good agreement (compared before switching to cosine). This:  https://github.com/ComputationalRadiationPhysics/picongpu/blob/f6753273597ccd6e5f829433fb96b19a871b6bb0/include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp#L194 used to be a sine, which is unusual and different from `GaussianPulse` implementation.

* `ExpRampWithPrepulse`  as it was compared with a Gaussian beam with the same temporal profile and focus at the initalization plane. The results look very comparable after changing sine to cos in `ExpRampWithPrepulse`. 

* Finally, a setup with   `ExpRampWithPrepulse`  profile and a focus in the simulation box.
One can see the focus at 8um as expected from my setup:
Figure 30
The amplitude at the focus, shifted in time by (distance from the generating surface and focus /c), matches well a curve generated with a python model of the envelope
Figure 51
The field doesn't properly drop at the end, this seems to have something to do with reflections from the boundary, generally the field looks weird at the end:

Figure 56
Figure 57
Figure 58
somehow the beam seams to be stuck at the front, idk. But the same thing happens for a standard `GaussianPulse` with mainline dev.


Figure 27
Figure 3(1)
Figure 28(1)


 So I guess this is probably caused, by some setup related numerics, boundary effects etc.?  Should we be worried about it??

### Example figures showing the artifacts in rel err.
This is comparison for `GaussianPulse` mainline dev vs my branch (so with and without the new time calculation).
Fields for an exemplary time step:
Figure 3(2)
Relative error for Ex at this time. There are just a few single cells with very large errors (close to 1)
Figure 1(3) Some of the structure could be explained with places where Ex gets closer to zero, large relative error in that case is normal. 


## testing setup:
https://github.com/pordyna/picongpu-setups/tree/d6daabbdace408c39c92dd844d43ef37d6ddcb9d/setups/testLaserProfiles is what I used. Some parts of it were made for comparing with python and I didn't continue with that. Maybe someone can take over at some point. Here are notebooks that I used to generate images like the ones I pasted above:
[laser_verify.ipynb.txt](https://github.com/ComputationalRadiationPhysics/picongpu/files/12729309/laser_verify.ipynb.txt)
[laser_example.ipynb.txt](https://github.com/ComputationalRadiationPhysics/picongpu/files/12729310/laser_example.ipynb.txt)

pordyna:
PICMI should not be broken, as the interface for old profiles didn't change. We should add the new profile, but doesn't have to now.

pordyna:
What is left is fixing the CI (probably just formatting), and updating / adding in code documentation.

pordyna:
How relevant is this for the envelope used with the Gaussian Beam? I guess, still relevant? @wmles 
> * Be careful - problematic for few cycle pulses. Thought the rest is cloned
    * from laserWavepacket, the correctionFactor is not included (this made a
    * correction to the laser phase, which is necessary for very short pulses,
    * since otherwise a test particle is, after the laser pulse has passed, not
    * returned to immobility, as it should). Since the analytical solution is
    * only implemented for the Gaussian regime, and we have mostly exponential
    * regimes here, it was not retained here.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -263,10 +263,8 @@ namespace picongpu::fields::incidentField
                                 - pmacc::math::Pi<float_X>::doubleValue / Unitless::WAVE_LENGTH * r
                                 + (2._X * float_X(m) + 1._X) * xi + phase + typename Unitless::LAGUERREPHASES_t{}[m]);
                     }
+                    auto const shiftedTime = time - r / SPEED_OF_LIGHT;
```
PrometheusPi, line 274: this changes the physics
Could you please elaborate why this is still true?

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
```diff
@@ -191,7 +191,7 @@ namespace picongpu::fields::incidentField
                 {
                     auto const runTimeShifted = time + Unitless::time_start_init;
                     auto const phase = Unitless::w * runTimeShifted + Unitless::LASER_PHASE + phaseShift;
-                    return math::sin(phase) * Unitless::AMPLITUDE * Envelope::getEnvelope(runTimeShifted);
+                    return math::cos(phase) * Unitless::AMPLITUDE * Envelope::getEnvelope(runTimeShifted);
```
PrometheusPi, line 194: why this change to `cos`?

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
pordyna: Well, the time now runs in positive direction, this doesn't change anything for the symmetric Gaussian. And phase offset does not shift the envelope anymore, but I don't think this was ever desired behavior?

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
pordyna: We can change it back, I guess, since the functor is going to be depreciated with this PR. But cos is the usual definition I guess? It is also cos in `GaussianPulse` and the two implementations were shifted in phase because fo that.

pordyna:
@psychocoderHPC the CI fail is weird, it is some `pmacc` memory test that is failing. Can you have a look?

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
PrometheusPi: Code that existed before

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
PrometheusPi: Code that existed before

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
PrometheusPi: Code that existed before

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
PrometheusPi: Code that existed before

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
PrometheusPi: Code that existed before

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
PrometheusPi: Code that existed before

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
PrometheusPi: old code - not fixed here

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
PrometheusPi: since it is deprecated soon - no need to change

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.hpp
PrometheusPi: old code - will not be fixed here

PrometheusPi (dismissed):

PrometheusPi:
@pordyna what do you need to change on this pull request?

pordyna:
@PrometheusPi I was just a bit surprised that it is still a WIP. I will quickly look if there is enough documentation. We don't really have any laser profiles documentation on readthedocs, so in code documentation should be enough?

pordyna:
should be good

PrometheusPi:
I agree, as we had no documentation before, in code documentation is fine (with me).

PrometheusPi (approved):

PR #4680, opened 2023-09-12
change radiation plugin python module to openPMD-api

PrometheusPi:
This pull request changes the python module for reading radiation data from h5py to openPMD-api 
This follows the changes in #3566. 

- [x] fix pep8
- [x] update documentation

### DIFF
--- a/lib/python/picongpu/extra/plugins/data/radiation.py
+++ b/lib/python/picongpu/extra/plugins/data/radiation.py
@@ -49,16 +49,17 @@ class RadiationData:
         # Amplitude
         detectorAmplitude = self.iteration.meshes["Amplitude"]
         # A_x
-        self.Ax_Re = detectorAmplitude["x_Re"].load_chunk()
-        self.Ax_Im = detectorAmplitude["x_Im"].load_chunk()
+        self.Ax_Re = detectorAmplitude["x_Re"][:, :, :]
+        self.Ax_Im = detectorAmplitude["x_Im"][:, :, :]
         # A_y
-        self.Ay_Re = detectorAmplitude["y_Re"].load_chunk()
-        self.Ay_Im = detectorAmplitude["y_Im"].load_chunk()
+        self.Ay_Re = detectorAmplitude["y_Re"][:, :, :]
+        self.Ay_Im = detectorAmplitude["y_Im"][:, :, :]
         # A_z
-        self.Az_Re = detectorAmplitude["z_Re"].load_chunk()
-        self.Az_Im = detectorAmplitude["z_Im"].load_chunk()
+        self.Az_Re = detectorAmplitude["z_Re"][:, :, :]
+        self.Az_Im = detectorAmplitude["z_Im"][:, :, :]
 
         # conversion factor for spectra from PIC units to SI units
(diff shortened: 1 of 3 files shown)


PrometheusPi:
I will fix the pep8 issues tomorrow.
Furthermore, the documentation needs to be updated.

lib/python/picongpu/extra/plugins/data/radiation.py
```diff
@@ -17,69 +17,65 @@
...
         # A_y
-        self.h5_Ay_Re = detectorAmplitude["y_Re"]
-        self.h5_Ay_Im = detectorAmplitude["y_Im"]
+        self.Ay_Re = detectorAmplitude["y_Re"].load_chunk()
+        self.Ay_Im = detectorAmplitude["y_Im"].load_chunk()
         # A_z
-        self.h5_Az_Re = detectorAmplitude["z_Re"]
-        self.h5_Az_Im = detectorAmplitude["z_Im"]
+        self.Az_Re = detectorAmplitude["z_Re"].load_chunk()
+        self.Az_Im = detectorAmplitude["z_Im"].load_chunk()
```
franzpoeschel: You could consider using the Python `__getitem__` syntax here `detectorAmplitude["z_Re"][:, :, :]`. This would make the dimensionality more obvious from just looking at the code (and additionally, where the `__getitem__()` syntax is available, the `load_chunk()` syntax is technically deprecated.. ;) )

lib/python/picongpu/extra/plugins/data/radiation.py
```diff
@@ -17,69 +17,65 @@
...
-        """Returns simulation timestep of the hdf5 data."""
-        # this is a workaround till openPMD is implemented
-        str_timestep = self.filename.split("_")[-4]
-        if str_timestep.isdigit():
-            return int(str_timestep)
-        else:
-            raise Exception("Could not extract timestep from " +
-                            "filename (\"{}\") - ".format(self.filename) +
-                            "Extracted: {}".format(str_timestep))
+        self.convert_to_SI = detectorAmplitude["x_Re"].unit_SI
```
franzpoeschel, line 64:
```suggestion
        # The value for unit_SI is consistent across all Amplitude datasets
        self.convert_to_SI = detectorAmplitude["x_Re"].unit_SI
```

I don't think that we need to check this property here, but let's maybe document that we make this assumption

franzpoeschel (commented):
I have two small comments below. We should also maybe be careful that the distributed amplitude is written with a different layout (`std::complex` instead of distinct datasets for real and imaginary parts) which might make future changes necessary.

lib/python/picongpu/extra/plugins/data/radiation.py
PrometheusPi: Thanks @franzpoeschel, I changed it

lib/python/picongpu/extra/plugins/data/radiation.py
PrometheusPi: Good suggestion.I added your text

psychocoderHPC (approved):

PR #4681, opened 2023-09-13
refactor particle init kernel

psychocoderHPC:
Motivation: The original implementation assumes that the number of slots in a frame and the number of cells within a supercell are equal. All particles from cell N were always created in the frame slot N too. In case there is a density gradient within the supercell frames are not filled fully which increases the memory footprint until the kernel fill gaps are called.

This PR decuples the number of cells in a supercell and the species frame size. Each thread searches a free slot within the frame and creates the particle in the slot, there is no implicit relation between the slot id in a frame and the cell where the particle is produced in. If the particle creation kernel is called multiple times for the same species the last frame will be filled first if possible.

### DIFF
--- a/include/picongpu/particles/ParticlesInit.kernel
+++ b/include/picongpu/particles/ParticlesInit.kernel
@@ -100,7 +100,7 @@ namespace picongpu
             T_ParBox pb,
             T_Mapping mapper) const
         {
-            constexpr uint32_t frameSize = pmacc::math::CT::volume<SuperCellSize>::type::value;
+            constexpr int frameSize = pmacc::math::CT::volume<SuperCellSize>::type::value;
             PMACC_CONSTEXPR_CAPTURE uint32_t cellsPerSupercell = pmacc::math::CT::volume<SuperCellSize>::type::value;
 
             using FramePtr = typename T_ParBox::FramePtr;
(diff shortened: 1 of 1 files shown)


ikbuibui (approved):
As discussed in the office, if we ensure that framesize >= supercell size, this looks good to me

include/picongpu/particles/ParticlesInit.kernel
```diff
@@ -173,71 +177,126 @@ namespace picongpu
...
+            worker.sync();
+
             // distribute the particles within the cell
             do
             {
-                // wait that master updates the current used frame
-                worker.sync();
-
-                onlyMaster([&]() { finished = 1; });
+                /* Old value of the destination counter is stored to check later to check if a worker
```
PrometheusPi: Typo: double "to check" or do you really mean this?

PrometheusPi (changes requested):
Just a typo. I am not the right person to check code functionality. Great, you checked that @ikbuibui. :+1:

include/picongpu/particles/ParticlesInit.kernel
psychocoderHPC: is a typo I will fix it

PrometheusPi (approved):

PR #4682, opened 2023-09-15
Make fieldAbsorberTest a test for the CI

### LINKED ISSUE #4610: Taflove PML test (=fieldAbsorberTest) fails?
It seems that the PML boundaries do not absorb incoming waves properly.
The test for PML boundaries described in "computational electrodynamics, Tavlove, 2005", p.315, shows a relative error below far below 10^-2. However, when performing the test simulation "~/src/picongpu/share/picongpu/examples/FieldAbsorberTest", the relative error appears to be multiple orders of magnitude above what we expect:
image

Commit Hash:
commit 536e1b72139d526de8a0b6da34abe9cbb2dedc29

steindev:
Closes #4610

Automatization of the fieldAbsorber test in the common way of providing a `ci.sh` and a `validate.sh` to automatically build, run, and evaluate the test.
The test is also moved to share/picongpu/tests and the setup is changed as well. Since we test in 3D, the source, a current carrying wire, is extended from -z boundary to +z boundary instead of being a point source.

It runs on hemera. In order to run the simulation on other systems, the picongpu call with `mpiexec` needs to be changed. I do not know how this could be solved in a generally applicable fashion.


PrometheusPi:
@steindev is this now reviewable?

steindev:
Yes, please review!

share/picongpu/tests/FieldAbsorber/README.rst
```diff
@@ -0,0 +1,49 @@
...
+
+.. sectionauthor:: Klaus Steiniger, Sergey Ermakov
+.. moduleauthor:: Sergei Bastrakov, Klaus Steiniger
+
+Example of a rectangular conductor with a steady current.
+
+This setup is based on [Taflove2005]_, section 7.11.1.
+The difference is we consider both 2D and 3D cases, and grid size may be increased due to our absorber being part of the simulation area, not located outside of it as in the book.
+
+The conductor is located within the center of the x-y-plane, oriented along the z-axis and infinitely extended along this axis, but current flows along y.
```
PrometheusPi, line 14: Question: How can I have a wire in z that transports current in y?

share/picongpu/tests/FieldAbsorber/README.rst
```diff
@@ -0,0 +1,49 @@
...
+The conductor properties can be adjusted in `include/picongpu/param/fieldBackground.param` within the `class FieldBackgroundJ`.
+Specifically, the variable `halfWidth` adjusts its edge length and can be used to apply the test with meaningful results to higher-order solvers, too.
+The current in the wire ramps up over time according to a differentiated Gaussian.
+This defines the current density amplitude, too.
+Therefore, the total current through the wire scales with the wire's halfWidth.
+
+The simulation is performed twice.
+First the actual PML test is performed with a small simulation volume, here 60x60x60 cells.
+The simulation runs for 1000 steps which is much longer than the steady state response.
+Second, a reference simulation is performed with a large simulation volume of 1060x1060x1060 cells.
```
PrometheusPi: `1060` is not dividable by `8`, thus this does not fit into standard supercells and will be auto-adjusted.

PrometheusPi: This is also not in agreement with the actual call. There `660` cells are used.

share/picongpu/tests/FieldAbsorber/README.rst
```diff
@@ -0,0 +1,49 @@
...
+PML boundaries are applied at the boundaries perpendicular to the x and y axes.
```
PrometheusPi:
```suggestion
PML boundaries are applied at the boundaries with normals in the x and y axes.
```

share/picongpu/tests/FieldAbsorber/README.rst
```diff
@@ -0,0 +1,49 @@
...
+Periodic boundaries are applied at the boundary perpendicular to the z axis.
```
PrometheusPi:
```suggestion
Periodic boundaries are applied at the boundary with a normal in the z axis.
```

share/picongpu/tests/FieldAbsorber/README.rst
```diff
@@ -0,0 +1,49 @@
...
+Since the reference grid is sufficiently large to avoid reflections from the boundary at the points of measurement, the PML quality is defined as the relative error
+
+.. math::
+
+  \mathrm{Rel. error}\rvert^{n}_{i,j} =
+    \frac{\left| E\rvert^{n}_{i,j} - E_\mathrm{ref}\rvert^{n}_{i,j}  \right|}{\left| E_\mathrm{ref,max}\rvert_{i,j}  \right|}
+
+with :math:`n` being the time step, :math:`(i,j)` the distance in cells from the conductor, :math:`E_\mathrm{ref,max}\rvert_{i,j}` the maximum amplitude of the reference field at :math:`(i,j)` over the total simulation duration.
+
+According to [Taflove2005]_, the PML quality should be of the order of :math:`10^{-5}` for a 10 cell CPML.
```
PrometheusPi, line 40: What is a **C**PML?

share/picongpu/tests/FieldAbsorber/README.rst
PrometheusPi: while discussing offline, you @steindev mentioned, that you adjusted the supercell size. However, github shows me that no cfg (except for background field) file has been altered.

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/requirements.txt
```diff
@@ -0,0 +1,5 @@
+numpy
+matplotlib
+scipy
+adios2
```
PrometheusPi: do you need this explicitly or only via openPMD-api? 
If the latter, please remove

share/picongpu/tests/FieldAbsorber/bin/ci.sh
```diff
@@ -0,0 +1,132 @@
+#!/bin/bash
+#
+# This file is part of PIConGPU.
+# Copyright 2023 PIConGPU contributors
+# Authors: Mika Soren Voss, Hannes Wolf, Klaus Steiniger
+# License: GPLv3+
+#
+
+set -o pipefail
```
PrometheusPi, line 9: the entire script does not use any pipes - this is not needed

share/picongpu/tests/FieldAbsorber/bin/ci.sh
```diff
@@ -0,0 +1,132 @@
...
+#################################
+## validate simulation results ##
+#################################
+if [ -d $dataPath ] ; then
+    echo "Validate!"
+    ./bin/validate.sh -d "$dataPath"
+fi
+ret=$?
+
+## clean up
```
PrometheusPi: please remove commented out code

share/picongpu/tests/FieldAbsorber/bin/validate.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
+#
+# This file is part of PIConGPU.
+# Copyright 2023 PIConGPU contributors
+# Authors: Mika Soren Voss, Rene Widera, Hannes Wolf, Klaus Steiniger
+# License: GPLv3+
+#
+
+set -o pipefail
```
PrometheusPi, line 9: no pipe is used in the script

share/picongpu/tests/FieldAbsorber/cmakeFlags
```diff
@@ -29,63 +29,67 @@
...
+# Compile test for standard Yee solver in 3D and setup for runtime field absorber test
 flags[0]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=Yee;-DPARAM_PML_SIZE=10;-DPARAM_PML_KAPPA_MAX=1.0;-DPARAM_PML_ALPHA_MAX=0.2'"
-flags[1]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=Yee;-DPARAM_PML_SIZE=10;-DPARAM_PML_KAPPA_MAX=1.0;-DPARAM_PML_ALPHA_MAX=0.2;-DPARAM_DIMENSION=DIM2'"
-# Test that arbitrary-order solver compiles in 3D and 2D
-flags[2]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=ArbitraryOrderFDTD<4>;-DPARAM_CFLFACTOR=1.29;-DPARAM_ABSORBER_SIZE=10;-DPARAM_PML_KAPPA_MAX=1.0;-DPARAM_PML_ALPHA_MAX=0.2'"
-flags[3]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=ArbitraryOrderFDTD<4>;-DPARAM_CFLFACTOR=1.29;-DPARAM_ABSORBER_SIZE=10;-DPARAM_PML_KAPPA_MAX=1.0;-DPARAM_PML_ALPHA_MAX=0.2;-DPARAM_DIMENSION=DIM2'"
+# Setup for reference simulation of runtime field absorber test
+flags[1]="-DPARAM_OVERWRITES:LIST='-DPARAM_FIELDSOLVER=Yee;-DPARAM_PML_SIZE=10;-DPARAM_PML_KAPPA_MAX=1.0;-DPARAM_PML_ALPHA_MAX=0.2;-DPARAM_GRID_SIZE=660'"
+
+# Compile test for standard Yee solver compiles in 2D
```
PrometheusPi, line 37: The following 31 setups are not used and do not follow the small/large style of case 0 and 1.
Should we still keep them?

share/picongpu/tests/FieldAbsorber/etc/picongpu/1.cfg
PrometheusPi, line 1: This is not needed by the ci - all command line calls are hard coded in the `ci.sh`
Could be removed.

share/picongpu/tests/FieldAbsorber/etc/picongpu/4.cfg
```diff
@@ -39,10 +39,12 @@ TBG_devices_z=1
 
 # When changing the number of cells consider changing positionX, positionY in FieldBackgroundJ::operator().
 # To match the setup from the Taflove-Hagness book, the size should be 40 + PML size min border + PML size max border.
-TBG_numCells=60
+TBG_numCells=1060
```
PrometheusPi: number does not match setup in ci but the documentation 
please fix

share/picongpu/tests/FieldAbsorber/etc/picongpu/4.cfg
PrometheusPi, line 1: as above - this file is not needed

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
```diff
@@ -0,0 +1,176 @@
...
+        Two-Dimensional Region' of the Taflove book, point A corresponds
+        to an offset only along x, while point B corresponds to an offset
+        along x and y.
+    """
+    # Define the offset of the measurement point from the center of the
+    # conductor in cells.
+    # The conductor is located in the center of simulation domain.
+    # The offset is choosen such that the measurement point is located 2 cells
+    # in front of the PML in the small volume.
+    offset_cells = int(18)
```
PrometheusPi, line 50: why do you explicitly cast and int to an int?

share/picongpu/tests/FieldAbsorber/bin/ci.sh
```diff
@@ -0,0 +1,132 @@
...
+    if [ $i -eq 0 ]; then
+      # make sure that the grid size along z matches
+      # the grid size along z of the reference simulation below
+      mpiexec -n 1 ../bin/picongpu -d 1 1 1 -g 60 60 660 --periodic 0 0 1 -s 600 \
+        --fields_energy.period 10 \
+        --openPMD.period 100 --openPMD.ext bp --openPMD.file simData_test
+
+    elif [ $i -eq 1 ]; then
+      # match the grid size with the grid size defined in `flags[1]` of cmakeFlags
+      mpiexec -n 1 ../bin/picongpu -d 1 1 1 -g 660 660 660 --periodic 0 0 1 -s 600 \
```
psychocoderHPC, line 101: This does not fits on our CI nodes. :-( It will consume too much memory.

share/picongpu/tests/FieldAbsorber/bin/ci.sh
```diff
@@ -0,0 +1,132 @@
...
+  pic-build -t $i
+  ret_build=$?
+
+  if [ $ret_build -eq 0 ] ; then
+    cd $simPath
+    echo "Run setup ${i}!"
+    if [ $i -eq 0 ]; then
+      # make sure that the grid size along z matches
+      # the grid size along z of the reference simulation below
+      mpiexec -n 1 ../bin/picongpu -d 1 1 1 -g 60 60 660 --periodic 0 0 1 -s 600 \
```
psychocoderHPC, line 95: This should be `60 60 60` and not `60 60 660

psychocoderHPC (dismissed):

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
```diff
@@ -0,0 +1,176 @@
...
+    x_offset = -offset_cells
+    y_offset = -offset_cells
+    z_offset = int(0)
```
PrometheusPi, line 53: how about checking various z positions via:
```suggestion
    z_offset = int(np.random.choice(np.arange(Nz//2)))
```
?

psychocoderHPC:
@steindev Please add the file `speciesDefinitions.param` to remove all species.

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
```diff
@@ -0,0 +1,176 @@
...
+    # of openPMD
+    Ex = np.zeros(Nt, dtype=np.float32)
+    Ey = np.zeros(Nt, dtype=np.float32)
+    Ez = np.zeros(Nt, dtype=np.float32)
+
+    Ex_ref = np.zeros(Nt, dtype=np.float32)
+    Ey_ref = np.zeros(Nt, dtype=np.float32)
+    Ez_ref = np.zeros(Nt, dtype=np.float32)
+
+    i = 0
```
PrometheusPi: please use `enuenumerate`  instead of a manual index `i`

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
```diff
@@ -0,0 +1,176 @@
...
+
+        Efield["x"].load_chunk(
+            Ex[i:i+1], offset=(z_meas, y_meas, x_meas), extent=(1, 1, 1)
+        )
+        Efield["y"].load_chunk(
+            Ey[i:i+1], offset=[z_meas, y_meas, x_meas], extent=[1, 1, 1]
+        )
+        Efield["z"].load_chunk(
+            Ez[i:i+1], offset=[z_meas, y_meas, x_meas], extent=[1, 1, 1]
+        )
```
PrometheusPi, line 100: would testing the B fields not also be useful?

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
```diff
@@ -0,0 +1,176 @@
...
+
+        series_ref.iterations[it].close()
+
+        Ex_ref[i] *= unit_fieldE
+        Ey_ref[i] *= unit_fieldE
+        Ez_ref[i] *= unit_fieldE
+
+        i += 1
+
+    del i
```
PrometheusPi: not needed if you use `enumerate`

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
```diff
@@ -0,0 +1,176 @@
...
+
+    """ Calculate the quantity defining the
+        quality/performance of the absorber.
+    """
+    component, component_ref = (
+        np.sqrt(Ex**2 + Ey**2 + Ez**2),
+        np.sqrt(Ex_ref**2 + Ey_ref**2 + Ez_ref**2)
+    )
+
+    quality = np.abs(component - component_ref)/np.abs(component_ref).max()
```
PrometheusPi, line 145: would a component-wise test not also be useful? If components are messed up the amplitude might still appear correct in some cases?

PrometheusPi (changes requested):

share/picongpu/tests/FieldAbsorber/bin/ci.sh
PrometheusPi: since it is periodic in z, using 660 in that direction is not needed

psychocoderHPC: I do not think that 660 cells are required for 600 steps if you only validate the results in the small spot from the tiny simulation.
IMO the size if a small sim should be `56 56 56` to be a multiple of the default supercell size `8 8 8` the simulation will be auto-scaled and the created field will not start from the middle.
The large simulation can be 360 cell because CFL says that the wave travels in 600 steps 360 cells.

psychocoderHPC: Offline discussed with the 360 cells I missed that the wave is starting i the middle of the simulation :-(

psychocoderHPC: @steindev ~~There is no `memory.param` therefore the supercell size is 8x8x4 by default. This should be taken into account for the small simulation.~~

[update] My fault, I am blind the file is there and the super cell size is adjusted.

share/picongpu/tests/FieldAbsorber/README.rst
steindev: Well, the orientation of the current density vector just depends on which faces of the square shaped wire you apply the potential difference ;-)

Honestly, this is just the Taflove setup. I do not know exactly how this choice came about. I wonder, whether there is a correlation between the fact that there is no radiation along the x axis in this case, and the choice of measurement points.

share/picongpu/tests/FieldAbsorber/README.rst
steindev: Because `memory.param` has not been changed. But it defines `using SuperCellSize = typename mCT::shrinkTo<mCT::Int<2, 2, 4>, simDim>::type;`

share/picongpu/tests/FieldAbsorber/README.rst
steindev: Done

share/picongpu/tests/FieldAbsorber/README.rst
steindev: Done

share/picongpu/tests/FieldAbsorber/README.rst
steindev: Convolutional PML. There is also Uniaxial PML (UPML), all defined in Taflove.

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/requirements.txt
steindev: Done

share/picongpu/tests/FieldAbsorber/bin/ci.sh
steindev: As far as I know, this line makes the script stop whenever an instruction fails. This is not limited to pipes. And I think I added it later during development, because I noted some early errors only later during execution.

steindev: (for example, when `cd`ing in a directory that does not exist. The script continues running when pipefail is not set)

share/picongpu/tests/FieldAbsorber/bin/ci.sh
steindev: Done

share/picongpu/tests/FieldAbsorber/bin/validate.sh
steindev: Same answer as above.

share/picongpu/tests/FieldAbsorber/cmakeFlags
steindev: This is a point we need to discuss in the dev meeting for two reasons.
1. Up until now, the fieldAbsorberTest is our compile test for all field solvers. If we do not use these in any other compile test, not all of these are compile tested.
2. Since all of these setups demonstrate reasonable values for the (C/U)PML configuration, I would argue to leave them in. But comment all except the first two which are the ones actually required for the runtime test.

share/picongpu/tests/FieldAbsorber/etc/picongpu/1.cfg
steindev: Also a discussion for the dev meeting. In principle you are right. But the command line call in `ci.sh` is not portable. Specifically, you need to call `srun` instead of `mpiexec` on most systems. In order to make this test easy executable on other systems, the `*.cfg` files are provided for convenience.

share/picongpu/tests/FieldAbsorber/etc/picongpu/4.cfg
steindev: Done

share/picongpu/tests/FieldAbsorber/etc/picongpu/4.cfg
steindev: See answer above.

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
steindev: Mainly habit. In the end, to be explicit. (I have made bad experiences when I tried to use numbers as an array index, of which I thought they where `int`s, but in reality they where not. For this reason, I always explicitly define something that is supposed be an array index as an int)

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
steindev: Since I only load a single slice (in order to keep the memory footprint low and reduce execution time), this is not possible. Also when going to close to the boundary, reflections from the +/- z-boundaries will interfere at the measurement point.

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
steindev: Well, Taflove didn't do it, so we have no numbers to compare to :sweat_smile:
But I guess. as long as E is fine, B must be fine as well. Otherwise we would see it through their relationship via Maxwell's equations.

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
steindev: Done

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
steindev: In this example, due to the current mainly pointing along y, we effectively test only the y component anyways. The other field components are small compared to it.

To really test all components and boundaries, we would need 6 pairs of small/large simulations where either the current orientation or the wire orientation changes from one to the other.

psychocoderHPC:
@steindev can you please rebase against the dev branch again and add in memory.param the single line with the farmesize, see https://github.com/ComputationalRadiationPhysics/picongpu/pull/4685

share/picongpu/tests/FieldAbsorber/bin/ci.sh
steindev: No, z is the "symmetry axis" and supposed to be equally long in the small and large simuolation

steindev:
> @steindev Please add the file `speciesDefinitions.param` to remove all species.

Done

steindev:
> @steindev can you please rebase against the dev branch again and add in memory.param the single line with the farmesize, see #4685

Updated `memory.param` with all additions from latest dev.

share/picongpu/tests/FieldAbsorber/README.rst
PrometheusPi: Okay

share/picongpu/tests/FieldAbsorber/README.rst
PrometheusPi: Okay

share/picongpu/tests/FieldAbsorber/bin/ci.sh
PrometheusPi: No, this only applies to pipes - would be nice to have something like that for fails as well, but then catching non-zero returns would not work anymore. 
Not that important since one might use pipes in the future. This line does not hurt.

share/picongpu/tests/FieldAbsorber/cmakeFlags
PrometheusPi: Okay - valid point - let's discuss this today in the` dev meeting

share/picongpu/tests/FieldAbsorber/etc/picongpu/1.cfg
PrometheusPi: Okay - then I would suggest using `tbg` and the `1.cfg` instead in the ci script
Let's discuss this today in the dev meeting.

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
PrometheusPi: okay

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
PrometheusPi: Thanks for clarifying.

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
PrometheusPi: Okay

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
PrometheusPi: Fine with me

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
PrometheusPi: Should be discussed regarding python-style in today's dev meeting.

share/picongpu/tests/FieldAbsorber/cmakeFlags
PrometheusPi: Result from dev meeting: keep it because we also use the tests for compile tests

share/picongpu/tests/FieldAbsorber/etc/picongpu/1.cfg
PrometheusPi: Result from the dev meeting: For now, keep `*.cfg` files. If we find out later that `tbg` works on the CI nodes, we will use `1.cfg`.

share/picongpu/tests/FieldAbsorber/etc/picongpu/4.cfg
PrometheusPi: dev meeting result: keep it for now

share/picongpu/tests/FieldAbsorber/lib/python/test/FieldAbsorber/validate.py
PrometheusPi: result dev meeting: keep it for now - use pydantic later

PrometheusPi (dismissed):

PrometheusPi:
@psychocoderHPC are you fine with merging?

PrometheusPi:
@psychocoderHPC issues are all solved - I will dismiss his review and merge

PrometheusPi:
@steindev your `memory.param` conflicts with the `dev` - please rebase and fix.

PrometheusPi (approved):

PrometheusPi:
@psychocoderHPC There is something broken with the CI. 
The `ci/gitlab/gitlab.com/pmacc-compile-reduced-matrix` is finished (if one looks at the status of all jobs) but the status is still "yellow".

SimeonEhrig:
Looks like reporting back the current status of the pipeline to GitHub failed. I restarted a single job. Maybe it sends again the status.

SimeonEhrig:
@PrometheusPi Now, everything is green.

PrometheusPi:
Thanks for the solution and the reminder @SimeonEhrig

PR #4687, opened 2023-09-21
remove pulling namespace `pmacc::math::CT` into `pmacc`

psychocoderHPC:
Avoid that namespace `pmacc::math::CT` is pulled into the global `pmacc` namespace.


PrometheusPi (approved):

ikbuibui (approved):
I don't see any other locations where the ``ct::`` namespace is used. So, this looks okay.

PR #4688, opened 2023-09-22
refactor memory boxes

psychocoderHPC:
Rewrite `PicthedBox` and `SharedBox` to avoid explicit specializations for each dimension.
`PitchedBox` still used a specialization for the one-dimensional memory to reduce the memory footprint.
The address calculation is simplified by using a `n - 1 `-dimensional pitch vector.

This PR is reducing the register footprint for many kernels running on NVIDIA GPUs. The constant memory footprint is slightly increased.
Short tests and a profile of the KHI example do not show performance regressions.

### DIFF
--- a/include/picongpu/fields/incidentField/Solver.kernel
+++ b/include/picongpu/fields/incidentField/Solver.kernel
@@ -95,16 +95,11 @@ namespace picongpu
                  * Performs update by adding terms with the incidentField.
                  * The positions, indices and coefficients for these terms are determined by members.
                  *
-                 * @tparam T_UpdatedFieldBox updatedField box type
                  * @tparam T_CurlIncidentField curl(incidentField) functor type
                  * @tparam T_FunctorIncidentField incidentField source functor type
                  * @tparam T_axis boundary axis, 0 = x, 1 = y, 2 = z
                  */
-                template<
-                    typename T_UpdatedFieldBox,
-                    typename T_CurlIncidentField,
-                    typename T_FunctorIncidentField,
-                    uint32_t T_axis>
+                template<typename T_CurlIncidentField, typename T_FunctorIncidentField, uint32_t T_axis>
                 struct UpdateFunctor
                 {
                     /** Create an update functor instance on the host side for the given time step
@@ -144,7 +139,9 @@ namespace picongpu
                      * @param isLastUpdatedCell whether the cell is the last updated globally along each direction,
                      *                          note it is always 3d, for 2d the last element must be false
                      */
+                    template<typename T_DestFieldDataBox>
                     HDINLINE void operator()(
+                        T_DestFieldDataBox destField,
                         pmacc::DataSpace<simDim> const& beginGridIdx,
                         pmacc::DataSpace<simDim> const& updatedGridIdx,
                         pmacc::math::Vector<bool, 3u> const& isLastUpdatedCell) const
@@ -158,15 +155,12 @@ namespace picongpu
                         auto const incidentFieldShift1 = incidentFieldShiftBase + inCellShift1;
                         auto const incidentFieldShift2 = incidentFieldShiftBase + inCellShift2;
                         auto const updatedFieldShift = (updatedGridIdx - huygensSurfaceIdx)[T_axis];
-                        updatedField(updatedGridIdx) += getUpdatedFieldCorrection(
+                        destField(updatedGridIdx) += getUpdatedFieldCorrection(
                             updatedFieldShift,
                             incidentFieldShift1,
                             incidentFieldShift2,
                             isLastUpdatedCell);
                     }
-
-                    //! UpdatedField box
-                    T_UpdatedFieldBox updatedField;
 
                     //! Derivative functor along T_axis type
                     using DerivativeFunctor
@@ -436,16 +430,19 @@ namespace picongpu
                      *
                      * @tparam T_Worker lockstep worker type
                      * @tparam T_UpdateFunctor update functor type
+                     * @tparam T_DestFieldDataBox type of the destination field box
                      *
                      * @param acc alpaka accelerator
                      * @param functor update functor
+                     * @param destField destination field
                      * @param beginGridIdx begin active grid index, in the local domain with guards
                      * @param endGridIdx end active grid index, in the local domain with guards
                      */
-                    template<typename T_Worker, typename T_UpdateFunctor>
+                    template<typename T_Worker, typename T_UpdateFunctor, typename T_DestFieldDataBox>
                     HDINLINE void operator()(
                         T_Worker const& worker,
-                        T_UpdateFunctor const functor,
+                        T_UpdateFunctor const& functor,
+                        T_DestFieldDataBox destField,
                         DataSpace<simDim> beginGridIdx,
                         DataSpace<simDim> endGridIdx) const
                     {
@@ -476,7 +473,7 @@ namespace picongpu
                                         = functor.isLastLocalDomain[d] && (updatedGridIdx[d] == endGridIdx[d] - 1);
                                 }
                                 if(isInside)
-                                    functor(beginGridIdx, updatedGridIdx, isLastUpdatedCell);
+                                    functor(destField, beginGridIdx, updatedGridIdx, isLastUpdatedCell);
                             });
                     }
                 };
--- a/include/pmacc/memory/boxes/DataBox.hpp
+++ b/include/pmacc/memory/boxes/DataBox.hpp
@@ -27,48 +27,31 @@
 
 namespace pmacc
 {
-    namespace detail
-    {
-        template<typename DataBox>
-        HDINLINE decltype(auto) access(const DataBox& db, DataSpace<1> const& idx = {})
-        {
-            return db[idx.x()];
-        }
-
-        template<typename DataBox>
-        HDINLINE decltype(auto) access(const DataBox& db, DataSpace<2> const& idx = {})
-        {
-            return db[idx.y()][idx.x()];
-        }
-
-        template<typename DataBox>
-        HDINLINE decltype(auto) access(const DataBox& db, DataSpace<3> const& idx = {})
-        {
-            return db[idx.z()][idx.y()][idx.x()];
-        }
-    } // namespace detail
-
     template<typename Base>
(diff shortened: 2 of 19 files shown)


PrometheusPi:
There are various errors causing the CI to fail: from not defined `+=` operations to non-existing functions. 
I am not sure if this might not just be a CI problem.

psychocoderHPC:
> There are various errors causing the CI to fail: from not defined `+=` operations to non-existing functions. I am not sure if this might not just be a CI problem.

Hope I fixed these issues now. The new implementation is providing only const reference in case the databox is constant. Therefore a few places crashing now where we said the box is const but later changed the memory the box is pointing to.

In general, this is a C++ issue. If an instance is const it means in C++ that the object is not mutable but there is no way to express that memory an object is pointing to is read-only (not mutable). Very often implementation goes the way that they say if an instance is const then the memory the instance is pointing too is const to.

include/picongpu/fields/incidentField/Solver.kernel
```diff
@@ -158,16 +155,13 @@ namespace picongpu
                         auto const incidentFieldShift1 = incidentFieldShiftBase + inCellShift1;
                         auto const incidentFieldShift2 = incidentFieldShiftBase + inCellShift2;
                         auto const updatedFieldShift = (updatedGridIdx - huygensSurfaceIdx)[T_axis];
-                        updatedField(updatedGridIdx) += getUpdatedFieldCorrection(
+                        destField(updatedGridIdx) += getUpdatedFieldCorrection(
```
psychocoderHPC, line 158: The refactoring that the field is passed into the function instead of storing as a member is required because for unknown reasons the field box was calling the const version of the memory access even if the object was not defined const.

psychocoderHPC: My fault, the problem was that the data box was stored as a member of the functor. I called the const version of the functor and therefore all members are seen as const too.

PrometheusPi:
@psychocoderHPC your pull request still does not pass the CI

include/pmacc/memory/boxes/DataBox.hpp
```diff
@@ -27,48 +27,31 @@
...
         }
 
-        HDINLINE DataBox(DataBox const&) = default;
+        DataBox(DataBox const&) = default;
 
         HDINLINE decltype(auto) operator()(DataSpace<Base::Dim> const& idx = {}) const
         {
-            ///@todo(bgruber): inline and replace this by if constexpr in C++17
-            return detail::access(*this, idx);
+            return Base::operator[](idx);
```
PrometheusPi, line 43: does it return a const value? 
If yes, how does it differ from line 48?

include/pmacc/memory/boxes/SharedBox.hpp
```diff
@@ -111,12 +117,33 @@ namespace pmacc
...
+         */
+        HDINLINE T_Type const* ptr(DataSpace<T_dim> const& idx) const
+        {
+            // offset in elements
+            int offset = idx.x();
+
+            if constexpr(Dim >= DIM2)
+            {
+                using Pitch = typename detail::CalcPitch<Extent, Extent::dim>::type;
+                for(uint32_t d = 1u; d < T_dim; ++d)
```
PrometheusPi, line 140: This only iterates over `T_dim - 1` values. Is this intended?

include/pmacc/memory/boxes/PitchedBox.hpp
```diff
@@ -23,154 +23,195 @@
...
+         * @param pointer pointer to the memory
+         * @param extent extent of the memory in elements
+         * @param pitch line pitch in bytes
+         */
+        template<typename T_MemoryIdxType>
+        HDINLINE PitchedBox(T_Type* pointer, math::Vector<T_MemoryIdxType, T_dim> const& extent, size_t const pitch)
+            : m_ptr(pointer)
+        {
+            m_pitch.x() = pitch;
+            for(uint32_t d = 1; d < T_dim - 1; ++d)
```
PrometheusPi, line 77: same as below:
This iterates only over `T_dim - 2` entries - is this intended?

include/pmacc/memory/boxes/PitchedBox.hpp
```diff
@@ -23,154 +23,195 @@
...
+        HDINLINE T_Type const* ptr(math::Vector<T_MemoryIdxType, T_dim> const& idx) const
         {
-            return ReducedType((TYPE*) ((char*) this->fixedPointer + idx * pitch));
+            /** offset in bytes
+             *
+             * We calculate the complete offset in bytes even if it would be possible to change the x-dimension with
+             * the native types pointer, this is reducing the register footprint.
+             */
+            size_t offset = sizeof(T_Type) * idx.x();
+            for(uint32_t d = 1u; d < T_dim; ++d)
```
PrometheusPi, line 120: same as below/above:
This only iterates over `T_dim -1` values. Is this intended?

PrometheusPi (dismissed):
Just a few question. 
It would be great if @ikbuibui could review this as well. You are more familiar with these data structures.

include/pmacc/memory/boxes/PitchedBox.hpp
psychocoderHPC: Yes the pitch extents (in bytes) is one dimension less compared to the index dimension.

include/pmacc/memory/boxes/SharedBox.hpp
psychocoderHPC: Yes is intended, pitch has one dimension less compared to the index.

include/pmacc/memory/boxes/PitchedBox.hpp
psychocoderHPC: Yes it is intended, the pitch dimension is 1 dimension less compared to the extent dimension. x-dimension is skipped because these value is set before the loop.

include/pmacc/memory/boxes/DataBox.hpp
psychocoderHPC: The implementation does not differ but once the function signature is `const` and once not. In the const method `Base` is `const` and therefor ethe const implementation there is called. `decltype(auto)` is deriving the return type of the call and provide a const reverence in case we call the const method else a normal reference.

include/pmacc/memory/boxes/DataBoxDim1Access.hpp
```diff
@@ -48,16 +48,26 @@ namespace pmacc
         {
         }
 
-        HDINLINE DataBoxDim1Access(DataBoxDim1Access const&) = default;
+        DataBoxDim1Access(DataBoxDim1Access const&) = default;
 
-        HDINLINE RefValueType operator()(DataSpace<DIM1> const& idx = {}) const
```
ikbuibui, line 53: ```RefValueType``` on line 39 is most probably unused

include/pmacc/memory/boxes/DataBoxDim1Access.hpp
```diff
@@ -48,16 +48,26 @@ namespace pmacc
         {
         }
 
-        HDINLINE DataBoxDim1Access(DataBoxDim1Access const&) = default;
+        DataBoxDim1Access(DataBoxDim1Access const&) = default;
```
ikbuibui, line 50: Just a doubt. I see in some other places in the code that copy constructors are marked as HDINLINE. I'm not sure I understand the implications of this removal.

include/pmacc/particles/memory/dataTypes/SuperCell.hpp
```diff
@@ -46,12 +46,12 @@ namespace pmacc
             return lastFramePtr;
         }
 
-        HDINLINE T_FrameType const* FirstFramePtr() const
+        HDINLINE T_FrameType* FirstFramePtr() const
```
ikbuibui, line 49: When asking for the the ```FirstFramePtr``` from a ```const``` supercell, does it makes sense to return a ```T_FrameType * const``` instead of ```T_FrameType *``` or ```T_FrameType const *```? As currently implemented, the const versions of ```FirstFramePtr()```  and ```LastFramePtr()```  is exactly the same as the non const versions.

include/pmacc/memory/boxes/PitchedBox.hpp
```diff
@@ -23,154 +23,195 @@
...
         }
 
-        HDINLINE PitchedBox(PitchedBox const&) = default;
+        HDINLINE T_Type const& operator[](DataSpace<DIM1> const& idx) const
+        {
+            return *(m_ptr + idx.x());
+        }
 
-        HDINLINE ReducedType operator[](const int idx) const
+        HDINLINE T_Type& operator[](DataSpace<DIM1> const& idx)
```
ikbuibui: Are ```const_cast```s required here and on line 197?

ikbuibui: Also a general styling comment. There is mixed use of T_Type and ValueType in the code, which are defined to be equal.

include/pmacc/memory/boxes/DataBoxDim1Access.hpp
psychocoderHPC: for default constructors, the host/device attribute is not required. It is a left over from old times.

include/pmacc/memory/boxes/PitchedBox.hpp
psychocoderHPC: `const_cast` should not be required.
I fixed the second point.

PrometheusPi:
@ikbuibui Are you fine with @psychocoderHPC changes/explanations?

ikbuibui:
> @ikbuibui Are you fine with @psychocoderHPC changes/explanations?

Yes, everything is fine by me. The CI however has other opinions.

include/pmacc/particles/memory/dataTypes/SuperCell.hpp
psychocoderHPC: `T_FrameType * const` will result in qualifier ignored warnings. 
I will keep it for now and we will refactor this part again because I am very unhappy with the current implementations of or containers at all.

PrometheusPi (approved):

PR #4689, opened 2023-09-22
rename unit current to unit current density

PrometheusPi:
After a lengthy discussion today, @ikbuibui, @psychocoderHPC, @BrianMarre and @PrometheusPi found out that the naming of a unit conversion factor in the current density kernel was wrong and resulted in a lot of confusion. 

This pull request fixes the naming.

### DIFF
--- a/include/picongpu/fields/FieldJ.tpp
+++ b/include/picongpu/fields/FieldJ.tpp
@@ -189,8 +189,8 @@ namespace picongpu
     HDINLINE
     FieldJ::UnitValueType FieldJ::getUnit()
     {
-        const float_64 UNIT_CURRENT_DENSITY = UNIT_CHARGE / UNIT_TIME / (UNIT_LENGTH * UNIT_LENGTH);
-        return UnitValueType(UNIT_CURRENT_DENSITY, UNIT_CURRENT_DENSITY, UNIT_CURRENT_DENSITY);
+        const float_64 unitCurrentDensity = UNIT_CHARGE / UNIT_TIME / (UNIT_LENGTH * UNIT_LENGTH);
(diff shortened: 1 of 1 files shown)


PrometheusPi:
Details on current density: https://en.wikipedia.org/wiki/Current_density

ikbuibui (approved):

include/picongpu/fields/FieldJ.tpp
```diff
@@ -189,8 +189,8 @@ namespace picongpu
     HDINLINE
     FieldJ::UnitValueType FieldJ::getUnit()
     {
-        const float_64 UNIT_CURRENT = UNIT_CHARGE / UNIT_TIME / (UNIT_LENGTH * UNIT_LENGTH);
-        return UnitValueType(UNIT_CURRENT, UNIT_CURRENT, UNIT_CURRENT);
+        const float_64 UNIT_CURRENT_DENSITY = UNIT_CHARGE / UNIT_TIME / (UNIT_LENGTH * UNIT_LENGTH);
```
psychocoderHPC: Another reason why we where so confused is that the naming schema is not fitting to our style guide and we thought this is a unit because it is written all caps.
Please change the variable name to `unitCurrentDensity` because this is a simple intermediate variable.

psychocoderHPC (changes requested):

include/picongpu/fields/FieldJ.tpp
PrometheusPi: @psychocoderHPC fixed

psychocoderHPC (approved):

steindev (approved):

PR #4691, opened 2023-09-25
fix comment on hdf5 vs adios and update examples

PrometheusPi:
#4680 missed removing a comment on hdf5 vs adios in the docs.
This pull request fixes that.
Furthermore, it now updates the example python code to adios only.

### DIFF
--- a/docs/source/usage/plugins/radiation.rst
+++ b/docs/source/usage/plugins/radiation.rst
@@ -459,7 +459,7 @@ Please be aware that all datasets in the openPMD output are given in the PIConGP
 .. code:: python
 
    import openpmd_api as io
-   series = io.Series("e_radAmplitudes_%T.pb", io.Access_Type.read_only)
+   series = io.Series("e_radAmplitudes_%T.bp", io.Access_Type.read_only)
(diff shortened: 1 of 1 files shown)


psychocoderHPC (dismissed):

docs/source/usage/plugins/radiation.rst
```diff
@@ -461,7 +459,7 @@ Please be aware that all datasets in the openPMD output are given in the PIConGP
 .. code:: python
 
    import openpmd_api as io
-   series = io.Series("e_radAmplitudes_%T_0_0_0.h5", io.Access_Type.read_only)
+   series = io.Series("e_radAmplitudes_%T.pb", io.Access_Type.read_only)
```
steindev: The file ending of adios files is `*.bp`

steindev (dismissed):
Just one typo.

docs/source/usage/plugins/radiation.rst
PrometheusPi: good catch - will fix

psychocoderHPC (approved):

PR #4693, opened 2023-09-25
fix interactive call in 5min tutorial

PrometheusPi:
@max-lehman14 found out that after #4515 we do not copy the `bash` templates when calling `pic-create`. This results in a problem, when following the 5 minutes on hemera tutorial, where the `mpiexec.tpl` is used to run interactively.   

This pull request provided a workaround for the tutorial.

### DIFF
--- a/docs/source/tutorials/hemeraIn5min.rst
+++ b/docs/source/tutorials/hemeraIn5min.rst
@@ -146,12 +146,13 @@ If this fails, read the manual or ask a colleague.
 
 After a successfull build, run (still on the compute node, still inside your scenario directory)::
 
-  tbg -s bash -c etc/picongpu/1.cfg -t etc/picongpu/bash/mpiexec.tpl $SCRATCH/tinkering/try01/run01
+  tbg -s bash -t $PICSRC/etc/picongpu/bash/mpiexec.tpl -c /etc/picongpu/1.cfg $SCRATCH/tinkering/try01/run01
 
 - :command:`tbg`: tool provided by PIConGPU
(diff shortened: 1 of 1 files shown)


BrianMarre (approved):
Looks, good.

By the way while you are at it, could you also fix the references to `/bigdata/hplsim/extern` and replace them with `/bigdata/hplsim/external`

The folder extern is owned by Hannes and other users are unable to create new folders inside it.

**Edit:** At least it used to be, but is no longer, I guess it got deleted and a student recreated it accidentally while following the tutorial.

BrianMarre (approved):

steindev (approved):

docs/source/tutorials/hemeraIn5min.rst
```diff
@@ -146,12 +146,13 @@ If this fails, read the manual or ask a colleague.
 
 After a successfull build, run (still on the compute node, still inside your scenario directory)::
 
-  tbg -s bash -c etc/picongpu/1.cfg -t etc/picongpu/bash/mpiexec.tpl $SCRATCH/tinkering/try01/run01
+  tbg -s bash -t $PICSRC/etc/picongpu/bash/mpiexec.tpl -c /etc/picongpu/1.cfg $SCRATCH/tinkering/try01/run01
```
steindev, line 149: @PrometheusPi The initial `/` of the `-c` option is not supposed to be there!

psychocoderHPC: @steindev please provide directly a fix that this not get lost

PR #4695, opened 2023-09-27
simplify output structure colorimeter plugin 

PrometheusPi:
The openPMD output of the calorimeter plugin had an unusual file name and sub-directory structure.

Currently, it looks like:
```
e_calorimeter/
└── all
    ├── e_calorimeter_bp_all_0.bp
    ├── e_calorimeter_bp_all_100.bp
```

 - with a subdirectory for every filter **and** the filter name in the file name
 - and with the file format at the end and in the middle of the file name
 
 This pull request changes this to a more common form with no duplications:
 ``` 
 e_calorimeter/
├── e_calorimeter_all_0.bp
├── e_calorimeter_all_100.bp
```
- no subdirectory for filters - filter names are only in the file name
- file format only at the end of the file

### DIFF
--- a/include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
+++ b/include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp
@@ -1,4 +1,5 @@
-/* Copyright 2016-2023 Heiko Burau, Rene Widera, Sergei Bastrakov
+/* Copyright 2016-2023 Heiko Burau, Rene Widera, Sergei Bastrakov,
+ * Richard Pausch
  *
  * This file is part of PIConGPU.
  *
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by PrometheusPi** [comment]
The pull request has been run time tested.

File: include/picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp [hunk 1]
```diff
@@ -527,9 +528,8 @@ namespace picongpu
             , m_cellDescription(cellDescription)
             , leftParticlesDatasetName("calorimeterLeftParticles")
         {
-            foldername = m_help->getOptionPrefix() + "/" + m_help->filter.get(m_id);
-            filenamePrefix
-                = m_help->getOptionPrefix() + "_" + m_help->fileName.get(m_id) + "_" + m_help->filter.get(m_id);
+            foldername = m_help->getOptionPrefix() + "/";
+            filenamePrefix = m_help->getOptionPrefix() + "_" + m_help->filter.get(m_id);
```
franzpoeschel, line 532: I was under the impression that we have kept the name so far for compatibility with the naming that the Python post-processing expects. Is that correct? If yes, do we need to update the naming there?

PrometheusPi: So far, there was no python post-processing. I am currently in the process of adding it #4697 
Thus changing it should not break anything.

steindev (approved):

PR #4696, opened 2023-09-28
add missing euler number in ADK formula

BrianMarre:
In our implementation of the ADK rate formula the sub term D was missing an euler number compared to our documentation and the reference Publication we were citing.

Since the reference publication does not make it clear whether it meant elementary charge or euler Number, I asked @n01r for clarification which version is correct and he pointed me to a [publication](https://link.aps.org/doi/10.1103/PhysRevA.59.569) confirming the euler number interpretation.

**Note:** the error resulted in an ADK rate which is `1/(<eulerNumber>^(2*n_eff))` of the correct ADK rate.
With `n_eff = (Z+1)/(sqrt(2*E_ionization/E_Hartree)` Z ... current charge state of the ion

### DIFF
--- a/docs/source/models/field_ionization.rst
+++ b/docs/source/models/field_ionization.rst
@@ -127,6 +127,7 @@ Leaving out the pre-factor distinguishes ``ADKCircPol`` from ``ADKLinPol``.
     .. attention::
 
         Be aware that :math:`Z` denotes the **residual ion charge** and not the proton number of the nucleus!
+        Be aware that :math:`e` in :math:`D` denotes Euler's number, not the elementary charge
 
 In the following comparison one can see the ``ADKLinPol`` ionization rates for the transition from Carbon II to III (meaning 1+ to 2+).
 For a reference the rates for Hydrogen as well as the barrier suppression field strengths :math:`F_\mathrm{BSI}` have been plotted.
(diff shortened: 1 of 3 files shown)


BrianMarre:
@PrometheusPi @psychocoderHPC CI failed due to an internal error, please restart

PrometheusPi (dismissed):
Great catch! Mssing eEuler number from Stirling's approximation.

docs/source/models/field_ionization.rst
```diff
@@ -127,6 +127,7 @@ Leaving out the pre-factor distinguishes ``ADKCircPol`` from ``ADKLinPol``.
     .. attention::
 
         Be aware that :math:`Z` denotes the **residual ion charge** and not the proton number of the nucleus!
+        Be aware that :math:`e` in :math:`D` denotes the euler number, not the elementary charge
```
PrometheusPi:
```suggestion
        Be aware that :math:`e` in :math:`D` denotes the Euler's number, not the elementary charge
```
see: https://en.wikipedia.org/wiki/E_(mathematical_constant)

PrometheusPi (changes requested):
just a small typo

docs/source/models/field_ionization.rst
BrianMarre: Done

PrometheusPi (approved):

PR #4703, opened 2023-10-10
fix derived field dataset naming

### LINKED ISSUE #4698: Error reading openPMD data after #4606
I have installed `picongpu@530a46eb9b4f5b9d4239f2e7f0ea5af2ea24321b` version of `20.09.2023`
But I did it using spack, so I have `adios2@2.8.0` and `openpmd-api@0.14.5`

After commits https://github.com/ComputationalRadiationPhysics/picongpu/pull/4606 https://github.com/ComputationalRadiationPhysics/picongpu/pull/4607 https://github.com/ComputationalRadiationPhysics/picongpu/pull/4608 I can't read openPMD data using `openPMD-reader`
It returns 
```
[ADIOS2] Warning: Attribute with name /data/0/fields/e_all_momentumDensity/geometry has no type in backend.
[ADIOS2] Warning: Attribute with name /data/0/fields/e_all_momentumDensity/geometry has no type in backend.
[~ADIOS2IOHandler] An error occurred: [ADIOS2] Requested attribute (/data/0/fields/e_all_momentumDensity/geometry) not found in backend.
```

I see that it can be related to renaming, `particleMomentumDensity` -> `MomentumDensity`
But I just don't know what package need to be updated to include this infopmation 
Is it latest `openpmd-api` or `adios2` or else.
Here is an example of data I can't read [simData_200000.bp.zip](https://github.com/ComputationalRadiationPhysics/picongpu/files/12765829/simData_200000.bp.zip)
And here is old working example [simData_200000.bp_OLD_WORKING.zip](https://github.com/ComputationalRadiationPhysics/picongpu/files/12765929/simData_200000.bp_OLD_WORKING.zip)

Comes from https://github.com/ComputationalRadiationPhysics/picongpu/issues/3996

psychocoderHPC:
fix #4698

With #4606 #4607 and #4680 we allow that more than one direction of derived fields can be printed. W used `/` to separate the component name from the data/mesh name. This is not valid openPMD or can at least confuse users because `/` is the separator used for openPMD ADIOS data.

This PR is changing the separation to `-` (minus) to not collide with the `_` (underscore) used for the data set name.
Derived fields will be independent data sets even if both show the same property but a different component e.g. x,y,z

**This PR must be backported to the release candidate 0.7.0!**

### DIFF
--- a/include/picongpu/particles/particleToGrid/derivedAttributes/MidCurrentDensityComponent.def
+++ b/include/picongpu/particles/particleToGrid/derivedAttributes/MidCurrentDensityComponent.def
@@ -83,7 +83,7 @@ namespace picongpu
                     HINLINE static std::string getName()
                     {
                         auto const componentNames = plugins::misc::getComponentNames(3);
-                        return "midCurrentDensity/" + componentNames[T_direction];
+                        return "midCurrentDensity-" + componentNames[T_direction];
                     }
 
                     /** Calculate a new attribute per particle
--- a/include/picongpu/particles/particleToGrid/derivedAttributes/Momentum.def
+++ b/include/picongpu/particles/particleToGrid/derivedAttributes/Momentum.def
@@ -77,7 +77,7 @@ namespace picongpu
                     HINLINE static std::string getName()
                     {
                         auto const componentNames = plugins::misc::getComponentNames(3);
-                        return "particleMomentum/" + componentNames[T_direction];
+                        return "particleMomentum-" + componentNames[T_direction];
                     }
 
                     /** Calculate value of the derived attribute per particle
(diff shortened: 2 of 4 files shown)


psychocoderHPC:
@franzpoeschel After this change the openPMD standard should be fulfilled again because each dataset is a scalar mesh.

franzpoeschel:
I'll have time to look at this in closer detail in the second half of the week. As a short comment for now: This PR is a workaround, and applying a workaround is not necessary in this situation since the intended proper representation as `midCurrentDensity/x, midCurrentDensity/y, midCurrentDensity/z` **is** possible in openPMD, it just needs to be done correctly in PIConGPU.

Please don't merge yet before we can discuss this in further detail.

psychocoderHPC:
fixed with #4704

PR #4715, opened 2023-10-23
Vector/DataSpace N-dimensional constructor

psychocoderHPC:
Provide a constructor to create an N-dimensional Vector/DataSpace with N values.

This PR does not change the register footprint on NVIDIA GPUs.
The constructor supports implicit casts equal to the hard-coded implementation before.

### DIFF
--- a/include/pmacc/math/vector/Vector.hpp
+++ b/include/pmacc/math/vector/Vector.hpp
@@ -108,28 +108,18 @@ namespace pmacc
             {
             }
 
-            HDINLINE
-            constexpr Vector(const type x)
+            /** Constructor for N-dimensional vector
+             *
+             * @attention This constructor allows implicit casts.
+             *
+             * @param args value of each dimension, x,y,z,...
+             */
+            template<typename... T_Args, typename = std::enable_if_t<(std::is_convertible_v<T_Args, T_Type> && ...)>>
+            HDINLINE constexpr Vector(T_Args... args)
             {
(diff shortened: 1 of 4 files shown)


include/pmacc/math/vector/Vector.hpp
```diff
@@ -108,28 +108,18 @@ namespace pmacc
...
+            template<typename... T_Args, typename = std::enable_if_t<(std::is_convertible_v<T_Args, T_Type> && ...)>>
+            HDINLINE constexpr Vector(T_Args... args)
             {
-                PMACC_CASSERT_MSG(math_Vector__constructor_is_only_allowed_for_DIM3, dim == 3u);
-                (*this)[0] = x;
-                (*this)[1] = y;
-                (*this)[2] = z;
+                static_assert(sizeof...(T_Args) == dim, "Number of arguments must be equal to the vector dimension.");
+                int i = 0;
+                ([&] { (*this)[i++] = args; }(), ...);
```
PrometheusPi, line 122: just out of interest: How do you unrole `args` here?

PrometheusPi (approved):

include/pmacc/math/vector/Vector.hpp
psychocoderHPC: This is a C++17 fold expression https://en.cppreference.com/w/cpp/language/fold
It is unrolling the expression for each argument in `args`

PR #4553, opened 2023-05-04
devServer install and profile for a100/a30

BrianMarre:
new profile and spack install script for nvidi a100/a30 gpus

PR #4183, opened 2022-06-24
gpu profile for dev-server

MSVoss:

PR #4086, opened 2022-04-29
hzdr dev server profile

BrianMarre:
upgrades to python 3.9.12 for testing of pypicongpu,

includes spack install commands for future reference

PR #4716, opened 2023-10-23
change order of PATH and fix wrong dev server path

PrometheusPi:
This pull request 
 - changes the order of the `$PATH` in our example profiles since in the previous order any previously loaded profile will define the PIConGPU code version instead of the latest loaded version. 
  - while doing this manually (I should have automated it) I found a bug in the `$PATH` includes for all dev-server example profiles. These were introduced in #4086, #4183, and #4553 and this pull request fixes them too.
  
  @psychocoderHPC I think that this is not a critical bug that needs back-porting to 0.7.0 - do you agree?

### DIFF
--- a/etc/picongpu/aris-grnet/gpu_picongpu.profile.example
+++ b/etc/picongpu/aris-grnet/gpu_picongpu.profile.example
@@ -54,7 +54,6 @@ export PIC_BACKEND="cuda:35"
 # relative to the PIConGPU source code of the tool bin/pic-create.
 export PIC_SYSTEM_TEMPLATE_PATH=${PIC_SYSTEM_TEMPLATE_PATH:-"etc/picongpu/aris-grnet"}
 
-export PATH=$PICSRC:$PATH
 export PATH=$PICSRC/bin:$PATH
 export PATH=$PICSRC/src/tools/bin:$PATH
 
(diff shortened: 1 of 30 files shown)


psychocoderHPC:
0.7.0 has been out since last week. The next release will hopefully be faster out than the other.

etc/picongpu/aris-grnet/gpu_picongpu.profile.example
```diff
@@ -54,9 +54,9 @@ export PIC_BACKEND="cuda:35"
 # relative to the PIConGPU source code of the tool bin/pic-create.
 export PIC_SYSTEM_TEMPLATE_PATH=${PIC_SYSTEM_TEMPLATE_PATH:-"etc/picongpu/aris-grnet"}
 
-export PATH=$PATH:$PICSRC
-export PATH=$PATH:$PICSRC/bin
-export PATH=$PATH:$PICSRC/src/tools/bin
+export PATH=$PICSRC:$PATH
```
psychocoderHPC: Could you please remove this line from all files, we do not have binaries in the main directory. I do not know why we have this in at all.

psychocoderHPC (changes requested):

BrianMarre (approved):

PrometheusPi:
@psychocoderHPC I removed all occurrences of `PATH=$PICSRC:$PATH` via:
```bash
for f in `grep -rl PATH . `; do sed -i -e '/export PATH=\$PICSRC:\$PATH/d' $f; done
``` 
and set the author to "Tools".

psychocoderHPC (approved):

PR #4721, opened 2023-10-24
remove unsed files, functions and classes

psychocoderHPC:
Some files were forgotten in the past when we already cleaned the code others have been unsed for a very long time and can be removed.


psychocoderHPC:
removed unused stuff only, therefore I will merge it without a review.

PR #4722, opened 2023-10-24
remove `atomicAddNoRet()`

psychocoderHPC:
Remove an old optimized atomicAdd interface.
HIP <5.0 is not supported therefore the atomic specialization is not required anymore.


psychocoderHPC:
emoves a code path for an old HIP version we do not support anymore. Therefore I will merge it.

Issue #4513, opened 2023-03-22
Flexible ParticleBinning Plugin


tmiethlinger:
Hello,
I wanted to share my view about a diagnostic in Smilei that I particularly like, and wanted to initiate a discussion regarding the feasibility and meaningfulness of such a plugin in PIConGPU.

This refers to the [ParticleBinning diagnostic](https://smileipic.github.io/Smilei/Use/namelist.html#particlebinning-diagnostics). One advantage that I see is it's flexibility, but I may get ahead of myself.

In short, having a (macro)particle with index $i$ and with attributes weights $w_i$, charge $q_i$, mass $m_i$, velocity $v_i$, momentum $p_i=m_i \gamma_i v_i$, etc., allows us make integrals/projections of, e.g., the following:
$$\int{A(\Gamma) f(\Gamma) \ d\Gamma} \approx \sum_{i=1}^N{a_i w_i}, \text{ or } \int{A(\Gamma) f(\Gamma) \ d\Gamma}/\int{f(\Gamma) \ d\Gamma} \approx = \sum_{i=1}^N{a_i w_i}/\sum_{i=1}^N{w_i}$$
corresponding to **extensive** *and* **intensive** quantities (in a thermodynamic sense). Here, $\Gamma$ refers to a point in phase-space, and $a_i$, related to an observable $A$, can be any quantity that can be expressed via their particle attributes (see above).

Smilei's *ParticleBinning diagnostic* however is not only very flexible with regards to the desired quantity $a_i$ / observable $A$, but also with regards to their axis of integration. This last point would probably be a unfeasible problem for GPU-based codes: integrating over real space is simply not the same as integrating over momentum space, due to the way particles are localized in memory.

Nevertheless, maybe something similar could be considered and adopted in the code.

PrometheusPi:
@tmiethlinger thanks for opening this issue. This indeed sounds interesting and we discussed this in the bi-weekly `dev` meeting.

chillenzer:
Isn't this closed by #4724? At least the text there suggests it.

PrometheusPi:
@chillenzer yes it is. Thanks for pointing this out.

ikbuibui:
@tmiethlinger I was curious if you also had anything else in mind with this issue that is currently missing in the Binning plugin. In such a case I will add that to the binningPlugin roadmap and treat it as a separate issue.

PR #4724, opened 2023-10-26
A flexible particle binning plugin for particles properties. 

ikbuibui:
Users can 
- Define their own axes
 - Define their own quantity which is binned
 - Choose which species which are used for the binning
 - Choose how frequently they want the binning to be executed
 - Choose if the binning should be time averaging or normalized by bin volume
 - Write custom output to file, for example other quantites related to the simulation which the user is interested in
 - Execute multiple binnings at the same time

Intends to solve #4513


include/picongpu/param/binningSetup.param
```diff
@@ -0,0 +1,198 @@
...
+                    speciesTuple,
+                    getCounts,
+                    "1",
+                    2000,
+                    false,
+                    false);
+            }
+        }
+    } // namespace plugins::binning
+} // namespace picongpu
```
psychocoderHPC: missing EOF

include/picongpu/plugins/binning/utility.hpp
```diff
@@ -0,0 +1,79 @@
...
+    {
+        namespace detail
+        {
+            template<typename TFunc, typename... TArgs, std::size_t... Is>
+            HDINLINE constexpr auto apply_impl(
+                TFunc&& f,
+                pmacc::memory::tuple::Tuple<TArgs...>&& t,
+                std::index_sequence<Is...>)
+            {
+                // TODO give Is as a param to f, so compiler has knowledge
```
psychocoderHPC: please use `@todo`, docygen will create  a todo list with all todo's in the cde

include/picongpu/plugins/binning/utility.hpp
```diff
@@ -0,0 +1,79 @@
...
+        template<typename... Args>
+        auto createTuple(Args const&... args)
+        {
+            return std::make_tuple(args...);
+            // return pmacc::memory::tuple::Tuple<Args...>(args...);
+            // return pmacc::memory::tuple::make_tuple(args...);
+        }
+
+    } // namespace plugins::binning
+} // namespace picongpu
```
psychocoderHPC: missing EOF

include/picongpu/plugins/binning/Axis.hpp
```diff
@@ -0,0 +1,317 @@
...
+        namespace axis
+        {
+            /**
+             * @brief Holds the range in axis space over which the binning will be done
+             */
+            template<typename T_Data>
+            class Range
+            {
+            public:
+                T_Data min; ///< Minimum of binning range
```
psychocoderHPC: Please move comments always above the variable and use `/**` to start the doxygen comment.
Could you please add to the comment if the value is in PIC or SI units.

include/picongpu/plugins/binning/Axis.hpp
```diff
@@ -0,0 +1,317 @@
...
+            };
+
+            /**
+             * @brief Holds the axis range and information on how this range is split into bins
+             */
+            template<typename T_Data>
+            class AxisSplitting
+            {
+            public:
+                Range<T_Data> m_range; ///< Range object
```
psychocoderHPC: Same as above, please sue doxygen and add a comment about the unit.

include/picongpu/plugins/binning/Binner.hpp
```diff
@@ -0,0 +1,319 @@
...
+         * Factor with picongpu units
+         * User needs to deal with the units seperately
+         */
+        // TODO make this more generic? databox operations with another databox?
+        // maybe store the normalization in a buffer rather than computing it at every output timestep (memory vs ops)
+        // this style vs having the members as params of operator()
+        template<uint32_t blockSize, uint32_t nAxes>
+        struct BinNormalizationKernel
+        {
+            using result_type = void;
```
psychocoderHPC: should be named `ResultType` or `ValueType` to follow our other naming schemas.

include/picongpu/plugins/binning/Binner.hpp
```diff
@@ -0,0 +1,319 @@
...
+            Binner(TBinningData bd, MappingDesc* cellDesc) : binningData{bd}, cellDescription{cellDesc}
+            {
+                Environment<>::get().PluginConnector().setNotificationPeriod(this, binningData.notifyPeriod);
+
+                /**
+                 * Allocate and manage global histogram memory here, to facilitate time averaging
+                 */
+
+                this->histBuffer = std::make_unique<HostDeviceBuffer<TDepositedQuantity, 1>>(
+                    binningData.axisExtentsND.productOfComponents()); // TODO for auto n_bins. allocate full size
```
psychocoderHPC: please move the comment into the comment block above.

include/picongpu/plugins/binning/Binner.hpp
```diff
@@ -0,0 +1,319 @@
...
+                    currentStep,
+                    binningData.axisExtentsND);
+
+                PMACC_LOCKSTEP_KERNEL(functorBlock, workerCfg)
+                (mapper.getGridDim())(mapper);
+            }
+        };
+
+    } // namespace plugins::binning
+} // namespace picongpu
```
psychocoderHPC: missing EOF

include/picongpu/plugins/binning/BinningCreator.hpp
```diff
@@ -0,0 +1,89 @@
...
+                    normalizeByBinVolume,
+                    writeOpenPMDFunctor);
+                auto binner = std::make_unique<Binner<BinningData<TAxisTuple, TSpeciesTuple, TDepositionData>>>(
+                    bd,
+                    cellDescription);
+                binnerVector.emplace_back(std::move(binner));
+            }
+        };
+    } // namespace plugins::binning
+} // namespace picongpu
```
psychocoderHPC: missing EOF

include/picongpu/plugins/binning/BinningData.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+            static constexpr uint32_t getNAxes()
+            {
+                return std::tuple_size_v<T_AxisTuple>;
+                // return utility::tuple::size(axisTuple);
+                // return utility::tuple::size(declval<T_AxisTuple>());
+            }
+        };
+    }; // namespace plugins::binning
+} // namespace picongpu
```
psychocoderHPC: missing EOF

psychocoderHPC (changes requested):

include/picongpu/plugins/binning/Axis.hpp
```diff
@@ -0,0 +1,317 @@
...
+                // also to be used in search based binning
+                std::vector<double> binEdges;
+                struct LinearAxisKernel
+                {
+                    T_AttrFunctor
+                        getAttributeValue; ///< Function to place particle on axis, returns same type as min and max
+                    T min, max; /**< Min and max values in the range of the binning. Values outside this range are
+                                placed in overflow bins */
+                    uint32_t nBins; ///< Number of bins in range
+                    using ScalingType = std::
```
psychocoderHPC, line 106: please add a comment that we use a type that avoids precision loss.

include/picongpu/plugins/binning/DomainInfo.hpp
```diff
@@ -0,0 +1,54 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
```
franzpoeschel, line 20: I'm having some trouble compiling this plugin at the moment since some `.hpp` files (this one, but also others like `UnitConversion.hpp`) have no `#include` directives, so only a very specific include order will work.
Can you give a short guide on how to compile and run a working instance of this plugin?

psychocoderHPC: could you please post the compiler error log that we can trace from where the error starts.

include/picongpu/plugins/binning/Binner.hpp
ikbuibui: Moved

include/picongpu/plugins/binning/utility.hpp
ikbuibui: Switched

include/picongpu/plugins/binning/Axis.hpp
ikbuibui: Changed comment style and added info that it is in SI units

include/picongpu/plugins/binning/Axis.hpp
ikbuibui: added

include/picongpu/plugins/binning/Binner.hpp
ikbuibui: switched to ResultType

include/picongpu/param/binningSetup.param
ikbuibui: Added

include/picongpu/plugins/binning/DomainInfo.hpp
franzpoeschel: It's not only one file that is affected by this:

```
[ 97%] Building CUDA object CMakeFiles/picongpu.dir/versionFormat.cpp.o                                                                                                                                                                                                                                                    
/home/franzpoeschel/git-repos/picongpu/include/picongpu/../picongpu/plugins/binning/UnitConversion.hpp(72): error: the global scope has no "openPMD"                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                           
/home/franzpoeschel/git-repos/picongpu/include/picongpu/../picongpu/plugins/binning/UnitConversion.hpp(72): error: too few arguments for class template "std::map"                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                           
/home/franzpoeschel/git-repos/picongpu/include/picongpu/../picongpu/plugins/binning/UnitConversion.hpp(72): error: expected an identifier                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                           
/home/franzpoeschel/git-repos/picongpu/include/picongpu/../picongpu/plugins/binning/UnitConversion.hpp(74): error: the global scope has no "openPMD"                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                           
/home/franzpoeschel/git-repos/picongpu/include/picongpu/../picongpu/plugins/binning/UnitConversion.hpp(74): error: expected a ";"                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                           
/home/franzpoeschel/git-repos/picongpu/include/picongpu/../picongpu/plugins/binning/UnitConversion.hpp(76): error: name followed by "::" must be a class or namespace name                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                           
/home/franzpoeschel/git-repos/picongpu/include/picongpu/../picongpu/plugins/binning/UnitConversion.hpp(76): error: name followed by "::" must be a class or namespace name                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                           
/home/franzpoeschel/git-repos/picongpu/include/picongpu/../picongpu/plugins/binning/UnitConversion.hpp(76): error: name followed by "::" must be a class or namespace name                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                           
/home/franzpoeschel/git-repos/picongpu/include/picongpu/../picongpu/plugins/binning/UnitConversion.hpp(76): error: name followed by "::" must be a class or namespace name                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                           
/home/franzpoeschel/git-repos/picongpu/include/picongpu/../picongpu/plugins/binning/UnitConversion.hpp(76): error: name followed by "::" must be a class or namespace name                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                           
/home/franzpoeschel/git-repos/picongpu/include/picongpu/../picongpu/plugins/binning/UnitConversion.hpp(76): error: name followed by "::" must be a class or namespace name                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                           
/home/franzpoeschel/git-repos/picongpu/include/picongpu/../picongpu/plugins/binning/UnitConversion.hpp(76): error: name followed by "::" must be a class or namespace name  
```

When I fix the includes here, the next include error is in `DomainInfo.hpp`.

Error triggered by adding `extensionParam.loader` with below content to LaserWakefield example and trying to compile it:

```cpp
#pragma once
#include "picongpu/param/binningSetup.param"
```

ikbuibui: The param file actually doesn't need to be added in extensionParam.loader, as the include is internally dealt with in the plugin.

psychocoderHPC: This plugin is different compared to our old plugins. We kept the configuration in the param folder but this file is not allowed to be included from `param.loader` or the extension.
The plugin will include the param files. This will also solve typical cyclic includes we have in PIConGPU.

ikbuibui: Also, could you pull again before trying again :smile:

franzpoeschel: I pulled again. How do I run this then? There is nothing in `picongpu --help` except for the single line `Binning Dispatcher: A flexible binning plugin!:`

franzpoeschel: Or are this the `--e_energyHistogram.*` parameters?

ikbuibui: So there is no runtime configuration, and to run the plugin, the user config goes into `picongpu/param/binningSetup.param` . There is an existing example in that file which you can simply call in the `getBinning()` function as `binningExample(binningCreator)`

franzpoeschel: Ah. I also had a simulation where the `if constexpr` part in the example apparently evaluated to `false`, so nothing happened. I have it running now.

include/picongpu/param/binningSetup.param
```diff
@@ -0,0 +1,216 @@
...
+                 * doing
+                 * This is called before any other mesh data is written to openPMD
+                 * Make sure to capture by copy only, as the objects in this file are not kept alive
+                 */
+                auto writeOpenPMD =
+                    [ax_py](::openPMD::Series& series, ::openPMD::Iteration& iteration, ::openPMD::Mesh& mesh) -> void
+                {
+                    mesh.setAttribute("p_nbins", ax_py.getNBins());
+                    // mesh.setAttribute("p_min", ax_py.getMin()); // This line throws an error in particle
+                    // calorimeter mesh.setAttribute("p_max", ax_py.getMax()); // This line throws an error in
```
franzpoeschel: Can you try splitting this on multiple lines? I'd like to see if the error comes from `getMin()` or from `setAttribute()`. Given that it's an error somewhere inside CUDA, it's most likely the `getMin()` and `getMax()` thing. I can imagine that copying `ax_py` into the closure might at this point not be enough to keep the underlying data alive/valid.

include/picongpu/plugins/binning/DomainInfo.hpp
ikbuibui: I realized that it was always evaluating to false because i was using `mpl::true_` instead of `mp11::mp_true` :sweat_smile:

include/picongpu/param/binningSetup.param
ikbuibui: The copy into the closure works, and writes output as expected. 
I tried recreating the error, and though I was struggling with it for a long time, I can no longer trigger it. 
I first thought maybe the error went away when I renamed the `binningSetup` from `.hpp` to `.param`, but that's not the case. So, I will for now ignore this until it breaks in the future. I also cleaned up some includes and namespaces, so that might also have helped at some point.  
Sorry about the hunt that went nowhere, but thanks for being the first person to pull and compile the plugin :smile:, I've updated the documentation with the things learnt from your first use.

include/picongpu/plugins/binning/WriteHist.hpp
```diff
@@ -0,0 +1,171 @@
...
+                {
+                    histExtent.emplace_back(static_cast<size_t>(bufferExtents[i]));
+                    histOffset.emplace_back(static_cast<size_t>(0));
+                }
+
+                record.setUnitSI(get_conversion_factor(outputUnits));
+
+                record.resetDataset({::openPMD::determineDatatype<Type>(), histExtent});
+                std::shared_ptr<Type> data(hReducedBuffer.getBasePointer(), [](auto const&) {});
+                record.storeChunk<Type>(data, histOffset, histExtent);
```
franzpoeschel, line 167:
```suggestion
#if OPENPMDAPI_VERSION_GE(0, 15, 0)
                ::openPMD::UniquePtrWithLambda<Type> data(
                    hReducedBuffer.getBasePointer(),
                    [bfr = std::move(hReducedBuffer)](auto const*) {});
#else
                std::shared_ptr<Type> data(hReducedBuffer.getBasePointer(), [](auto const&) {});
#endif
                record.storeChunk<Type>(std::move(data), histOffset, histExtent);
```

This can help save memory in the BP5 engine when using a somewhat recent installation of openPMD-api. 
I'll need to check the exact version again and the `hReducedBuffer` would then need to be passed into this whole function by move instead of by reference.

If I don't forget it, I'll prepare a commit for this next week.

franzpoeschel: Hm, I get errors from this. Need to look more closely next week.

psychocoderHPC: Thanks for looking into it. We will keep this PR the code as it is and your optimization should go into the code a separate optimization.
I assume your optimization should be applied to other parts of PIConGPU too.

psychocoderHPC: IMO `std::move(hReducedBuffer)` is not allowed because `hReducedBuffer` is managed by PIConGPU and the plugin is reusing the buffer. If you move the object into the lambda it would mean you are not allowed to touch the buffer anymore.

franzpoeschel: > I assume your optimization should be applied to other parts of PIConGPU too.

No, most other parts of PIConGPU already use `storeChunkSpan()` from `include/picongpu/plugins/common/openPMDVersion.def` which is a better optimization than this one, but would require heavy restructuring here.

franzpoeschel: > IMO `std::move(hReducedBuffer)` is not allowed because `hReducedBuffer` is managed by PIConGPU and the plugin is reusing the buffer. If you move the object into the lambda it would mean you are not allowed to touch the buffer anymore.

Is that so? The way I see it, `Binner::notify()` in `include/picongpu/plugins/binning/Binner.hpp` allocates the buffer and the last use of it is in the call to `histWriter()`. Moving the buffer is the entire point of this optimization, so if we can't do that, then we can leave things as they are.

franzpoeschel: > then we can leave things as they are.

.. we should at least use `storeChunkRaw()` instead of a fake shared pointer though

include/picongpu/plugins/binning/WriteHist.hpp
```diff
@@ -0,0 +1,171 @@
...
+                if(std::any_of(extension.begin(), extension.end(), [](char const c) { return c == '.'; }))
+                {
+                    filename << binningData.binnerOutputName << extension;
+                }
+                else
+                {
+                    filename << binningData.binnerOutputName << '.' << extension;
+                }
+
+                auto series = ::openPMD::Series(dir + '/' + filename.str(), ::openPMD::Access::CREATE);
```
franzpoeschel, line 76: Not necessary to do in this PR, but if at some point we want more than file-based encoding (one file per iteration), we should consider leaving the `Series` open between calls to this plugin.

include/picongpu/plugins/binning/WriteHist.hpp
franzpoeschel: If I read `include/pmacc/memory/buffers/HostBuffer.hpp` correctly, then `HostBuffer` does not implement move semantics? There is nothing that ever modifies `pointer` and `ownPointer`, so after a move you get a double free.
(If we don't want to implement those operations, it would probably be good to explicitly `delete` them, to avoid others falling into the same problem)

franzpoeschel: [This commit](https://github.com/ComputationalRadiationPhysics/picongpu/commit/fe11baebaa840d7dd1a280e10b93f3c8389843ed) has the full suggestion for adding this optimization. Alternatively to the workaround there (putting the HostBuffer into a unique_ptr), one could also add move semantics to HostBuffer.
I've tested it, this one runs fine now.

include/picongpu/param/binningSetup.param
```diff
@@ -0,0 +1,211 @@
...
+
+        inline void getBinning(BinningCreator& binningCreator)
+        {
+            /// USER CODE GOES BELOW
+
+            /// USER CODE ENDS HERE
+        }
+
+        template<typename T_Identifier>
+        struct NameOrTypeExists
```
psychocoderHPC: Offline discussed we shoudl use `pmacc::particles::meta::FindByNameOrType_t` with the policy `ReturnType` or `ReturnValue`

include/picongpu/param/binningSetup.param
```diff
@@ -0,0 +1,211 @@
...
+                          T_Identifier,
+                          T_Value> || std::is_same_v<pmacc::traits::GetCTName_t<T_Value>, T_Identifier>;
+            };
+
+            using FilteredSeq = mp_copy_if<VectorAllSpecies, HasTypeOrName>;
+
+            static constexpr bool value = !std::is_same_v<boost::mp11::mp_true, mp_empty<FilteredSeq>>;
+        };
+
+        inline void binningExample(BinningCreator& binningCreator)
```
psychocoderHPC, line 42: please add a comment that this is the check that the species exists. Best use

```
constexpr speciesExists = ....;
if constexpr(speciesExists) ...
```

include/picongpu/param/binningSetup.param
ikbuibui: Changed.

include/picongpu/param/binningSetup.param
ikbuibui: Comment added

psychocoderHPC (approved):

PR #4729, opened 2023-11-03
Fix openPMD plugin parameter default values

franzpoeschel:
I noticed that the `picongpu --help` page claimed that the default value for `--openPMD.ext` is `bp4`, but files are actually created as `simData_000000.bp`. This is because the default values are specified at two places:

1. In the `Option` members of `openPMDWriter::Help`. These values are shown on the command line.
2. In the data members of `PluginParameters`. These are the values that actually count.

This PR removes all default value specification from `PluginParameters` and initializes them from the `Option` data members.
This also serves to unify the default values of command line and TOML parameters.

TODO:

- [x] Update documentation
- [x] The command line now shows that `openPMD.file` has a default value. Maybe make this a non-required parameter.

### DIFF
--- a/include/picongpu/plugins/openPMD/Parameters.hpp
+++ b/include/picongpu/plugins/openPMD/Parameters.hpp
@@ -8,7 +8,7 @@ namespace picongpu::openPMD
     /*
      * No default values here since those are registered in the
      * plugins::multi::Option data members of openPMDWriter::Help.
-     * The openPMD plugin will automatically those default values here.
+     * The openPMD plugin will automatically use those default values here.
      * Ref.: openPMDWriter::Help::pluginParameters() (when using cmd line parameters)
      *       openPMDWriter::openPMDWriter()          (when using TOML configuration)
      */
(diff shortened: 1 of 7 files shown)


include/picongpu/plugins/openPMD/Parameters.hpp
```diff
@@ -5,14 +5,21 @@
 
 namespace picongpu::openPMD
 {
+    /*
+     * No default values here since those are registered in the
+     * plugins::multi::Option data members of openPMDWriter::Help.
+     * The openPMD plugin will automatically those default values here.
```
PrometheusPi:
```suggestion
     * The openPMD plugin will automatically choose those default values here.
```

PrometheusPi (changes requested):
looks good to me - just a small typo (missing word)

include/picongpu/plugins/openPMD/Parameters.hpp
franzpoeschel: Thanks for spotting. Below is the wording that I was going for, there is not much of a choice going on in this logic.
```suggestion
     * The openPMD plugin will automatically use those default values here.
```

PrometheusPi (approved):

PR #4731, opened 2023-11-06
Get supercell index relative to user defined origin

### LINKED ISSUE #4683: Rework `pmacc::getSuperCellIndex()` to get Idx wrt origin of choice
Defined in in `AreaMapping.hpp`, `pmacc::getSuperCellIndex()` currently returns the supercell index with respect to the guard origin. 
This makes sense, but it might be easy to overlook this detail and assume the origin is defined with respect to the mapper for example.
It is probably helpful to have the functionality to let the user decide which origin should be used for the supercell index (GUARD/BORDER/CORE), thus also making this choice explicit.

ikbuibui:
Supercell index relative to user defined origin
Add a template parameter to `getSupercellIndex` which defines the origin with respect to which to return the supercell index (Core/Border/Guard)
Defaults to Guard to maintain current functionality
Closes #4683

### DIFF
--- a/include/pmacc/mappings/kernel/AreaMapping.hpp
+++ b/include/pmacc/mappings/kernel/AreaMapping.hpp
@@ -75,12 +75,24 @@ namespace pmacc
 
         /** Return index of a supercell to be processed by the given alpaka block
          *
+         * @tparam T_origin Which origin (CORE/BORDER/GUARD) to return supercell index relative to (default: GUARD)
          * @param blockIdx alpaka block index
          * @return mapped SuperCell index including guards
          */
+        template<uint32_t T_origin = GUARD>
         HDINLINE DataSpace<DIM> getSuperCellIndex(const DataSpace<DIM>& blockIdx) const
         {
-            return AreaMappingMethods<areaType, DIM>::getSuperCellIndex(*this, this->getGridSuperCells(), blockIdx);
+            auto result
+                = AreaMappingMethods<areaType, DIM>::getSuperCellIndex(*this, this->getGridSuperCells(), blockIdx);
+            if constexpr(T_origin == CORE)
+            {
+                result = result - 2 * this->getGuardingSuperCells();
+            }
+            if constexpr(T_origin == BORDER)
+            {
+                result = result - this->getGuardingSuperCells();
+            }
+            return result;
         }
     };
 
--- a/include/pmacc/mappings/kernel/AreaMappingMethods.hpp
+++ b/include/pmacc/mappings/kernel/AreaMappingMethods.hpp
@@ -30,10 +30,10 @@ namespace pmacc
      * Helper class for AreaMapping.
      * Provides methods called by AreaMapping using template specialization.
      *
-     * @tparam areaType the area to map to
+     * @tparam T_area the area to map to
      * @tparam DIM dimension of the mapping
      */
-    template<uint32_t areaType, unsigned DIM>
+    template<uint32_t T_area, unsigned DIM>
     class AreaMappingMethods;
 
     // CORE + BORDER + GUARD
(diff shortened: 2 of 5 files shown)


include/pmacc/mappings/kernel/AreaMapping.hpp
```diff
@@ -75,12 +75,17 @@ namespace pmacc
 
         /** Return index of a supercell to be processed by the given alpaka block
          *
+         * @tparam RelativeToArea Which origin (CORE/BORDER/GUARD) to return supercell index relative to
          * @param blockIdx alpaka block index
          * @return mapped SuperCell index including guards
          */
+        template<uint32_t RelativeToArea = GUARD>
```
psychocoderHPC: Could you rename thie to `T_Origin` (and in the comment). That will fit the naming we use in the parts of the code too.

include/pmacc/mappings/kernel/AreaMappingMethods.hpp
```diff
@@ -30,12 +30,44 @@ namespace pmacc
      * Helper class for AreaMapping.
      * Provides methods called by AreaMapping using template specialization.
      *
-     * @tparam areaType the area to map to
+     * @tparam AreaType the area to map to
      * @tparam DIM dimension of the mapping
+     * @tparam Origin Return supercellIdx wrt specified origin
      */
-    template<uint32_t areaType, unsigned DIM>
+    template<unsigned AreaType, unsigned DIM, uint32_t Origin = GUARD>
```
psychocoderHPC: `AreaType` the naming was before confusing and we should change it. It is not a type.

The naming rules are `uint32_t T_area` and `T_origin` if it is not a type.
Template types `typename T_CameCase` are always start with capital letter.

see https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md#16-template-declarations-and-specializations

psychocoderHPC (changes requested):
Can you add the functionallity to the other mappers too?

include/pmacc/mappings/kernel/AreaMapping.hpp
```diff
@@ -75,12 +75,17 @@ namespace pmacc
...
         HDINLINE DataSpace<DIM> getSuperCellIndex(const DataSpace<DIM>& blockIdx) const
         {
-            return AreaMappingMethods<areaType, DIM>::getSuperCellIndex(*this, this->getGridSuperCells(), blockIdx);
+            return AreaMappingMethods<areaType, DIM, RelativeToArea>::getSuperCellIndex(
```
psychocoderHPC: IMO it is not necessary to propagate the origin to the implementation!

include/pmacc/mappings/kernel/AreaMapping.hpp
ikbuibui: Renamed to T_origin

include/pmacc/mappings/kernel/AreaMapping.hpp
ikbuibui: Agreed. Moved the origin shift to this file.

include/pmacc/mappings/kernel/AreaMappingMethods.hpp
ikbuibui: Renamed to T_area

ikbuibui:
> Can you add the functionallity to the other mappers too?

Functionality now exists in 
- `AreaMapping`
- `BorderMapping`
- `ExchangeMapping`
- `StrideMapping`

It doesn't exist in 
- `IntervalMapping`
- `StrideIntervalMapping`

As discussed in office, this disparity is because the two interval mappers work with linear intervals and the concept doesn't translate to them.

psychocoderHPC (approved):

PR #4685, opened 2023-09-18
refactor to make superCellSize and frameSize independant

psychocoderHPC:
- param-files: add `numFrameSlots` to set the number of particle slots per frame
- Refactor the PIConGPU extents for the random number factory used in PIConGPU.
  - each number of RNG states for each supercell is now dimensionless

# Update old setups:
  - add to `memory.param` after the definition of `SuperCellSize`

```C++
    /** number of slots for particles within a frame */
    static constexpr uint32_t numFrameSlots = pmacc::math::CT::volume<SuperCellSize>::type::value;
```

# Refactorings required if you develop extensions for PIConGPU

There are now `numFrameSlots` RNG states per supercell available. Instead of mapping the worker ID to a cell within the supercell you should calculate the RNG state cell this way:

```C++
// offset of the superCell (in cells, without any guards) to the
// origin of the local domain
DataSpace<simDim> const localSuperCellOffset = superCellIdx - mapper.getGuardingSuperCells();
auto rngOffset = DataSpace<simDim>::create(0);
rngOffset.x() = worker.getWorkerIdx();
auto numRNGsPerSuperCell = DataSpace<simDim>::create(1);
numRNGsPerSuperCell.x() = numFrameSlots;
rngHandle.init(localSuperCellOffset * numRNGsPerSuperCell + rngOffset);
```

PR #4736, opened 2023-11-08
fix RNG init in ionization

psychocoderHPC:
With the refactoring of the frame size in #4685 the particle creation kernel used for ionization was not updated.

This PR is fixing the wrong index access to RNG states.

When ionization was enabled the code crashed with illegal memory access.

The bug was first observed by @pordyna.

### DIFF
--- a/include/picongpu/particles/creation/creation.kernel
+++ b/include/picongpu/particles/creation/creation.kernel
@@ -78,10 +78,6 @@ namespace picongpu
 
                     DataSpace<simDim> const supercellCellOffset = superCellIdx * SuperCellSize::toRT();
 
-                    // relative offset to the origin of the local domain (without any guarding cells)
-                    pmacc::math::Int<simDim> const simDomainSupercellCellOffset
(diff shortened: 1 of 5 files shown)


pordyna:
@psychocoderHPC FoilLCT serial fails with  `picongpu/particles/creation/creation.kernel:82:52: error: variable 'simDomainSupercellCellOffset' set but not used` :(

psychocoderHPC:
> @psychocoderHPC FoilLCT serial fails with `picongpu/particles/creation/creation.kernel:82:52: error: variable 'simDomainSupercellCellOffset' set but not used` :(

Fixed, the variable is not used anymore because the init is now getting the supercell index instead of the cell index.

PrometheusPi:
@psychocoderHPC - there are conflicts that need to be resolved.

psychocoderHPC:
> @psychocoderHPC - there are conflicts that need to be resolved.

I rebased the PR

PrometheusPi (approved):

PR #4746, opened 2023-11-20
remove accessor from `math::Vector<>`

psychocoderHPC:
Our implementation of `math::Vector` supports only accessors that provide references, other is not supported by design. To support an accessor that is not returning a reference we need to take care in all implementations if we only read the value or we like to mutate the vector.

Since a generic accessor was not supported and we do not use this feature we will remove the accessor to simplify the implementation.


PrometheusPi (dismissed):

include/pmacc/meta/conversion/SeqToMap.hpp
```diff
@@ -30,10 +30,8 @@ namespace pmacc
     /** convert a list to a map
      *
      * @tparam T_List an mp_list.
-     * @tparam T_UnaryOperator unary operator to translate type from the sequence
+     * @tparam T_MakePairUnaryOperator unary operator to translate type from the sequence
```
PrometheusPi, line 33: This is not necessary for this pull request, but this change is also too small for a separate pull request. Thus, this is fine with me.

PrometheusPi (approved):

include/pmacc/meta/conversion/SeqToMap.hpp
psychocoderHPC: Yes searched for accessor and deleted the line bellow and catched the wrong name.

PR #4752, opened 2023-11-24
add method to radiation python module to load distributed amplitudes

PrometheusPi:
This pull request adds a new method to the radiation plugin python module to load all distributed radiation amplitudes if available.

### DIFF
--- a/docs/source/usage/plugins/radiation.rst
+++ b/docs/source/usage/plugins/radiation.rst
@@ -532,16 +532,17 @@ This currently assumes adios output but any openPMD output, such as hdf5, will w
 There are various methods besides ``get_Spectra()`` that are provided by the python module.
 If a method exists for ``_x`` (or ``_X``) it also exists for ``_y`` and ``_z`` (``_Y`` and ``_Z``) accordingly.
 
-============================ ==============================================================================================================
-Method                       Description
-============================ ==============================================================================================================
-``.get_omega()``             get frequency :math:`\omega` of virtual detector bin in units of :math:`\mathrm{[1/s]}`
-``.get_vector_n()``          get observation direction :math:`\vec{n}`
-``.get_Spectra()``           get spectrum :math:`\mathrm{d}^2 I / \mathrm{d} \omega \mathrm{d} \Omega` in units of :math:`\mathrm{[Js]}`
-``.get_Polarization_X()``    get spectrum but only for polarization in x-direction
-``.get_Amplitude_x()``       get x-component of complex amplitude (unit: :math:`\mathrm{[\sqrt{Js}]}`)
-``.get_timestep()``          the iteration (timestep) at which the data was produced (unit: PIC-cycles)
-============================ ==============================================================================================================
+================================= ==============================================================================================================
+Method                            Description
+================================= ==============================================================================================================
+``.get_omega()``                  get frequency :math:`\omega` of virtual detector bin in units of :math:`\mathrm{[1/s]}`
+``.get_vector_n()``               get observation direction :math:`\vec{n}`
+``.get_Spectra()``                get spectrum :math:`\mathrm{d}^2 I / \mathrm{d} \omega \mathrm{d} \Omega` in units of :math:`\mathrm{[Js]}`
+``.get_Polarization_X()``         get spectrum but only for polarization in x-direction
+``.get_Amplitude_x()``            get x-component of complex amplitude (unit: :math:`\mathrm{[\sqrt{Js}]}`)
+``.get_timestep()``               the iteration (timestep) at which the data was produced (unit: PIC-cycles)
+``.get_distributedAmplitude()``   get 4D complex array of distributed amplitudes (shape: (N_gpus, N_observer, N_frequency, N_polarization))
+================================= ==============================================================================================================
 
 .. note::
 
--- a/lib/python/picongpu/extra/plugins/data/radiation.py
+++ b/lib/python/picongpu/extra/plugins/data/radiation.py
@@ -18,9 +18,13 @@
 #
 import numpy as np
 import openpmd_api as opmd
+import warnings
 
 
 class RadiationData:
+    # number of vector components - DO NOT CHANGE
+    N_vec_components = 3
+
     def __init__(self, filename, timestep):
         """
         Open references to an openPMD-api series or file to access radiation
@@ -118,7 +122,7 @@ class RadiationData:
 
         self.rad_series.flush()
 
-        n_vec = np.empty((len(n_x), 3))
+        n_vec = np.empty((len(n_x), RadiationData.N_vec_components))
         n_vec[:, 0] = n_x * n_x_unitSI
         n_vec[:, 1] = n_y * n_y_unitSI
         n_vec[:, 2] = n_z * n_z_unit_SI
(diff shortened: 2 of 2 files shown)


PrometheusPi:
Code has been tested.

lib/python/picongpu/extra/plugins/data/radiation.py
```diff
@@ -125,3 +126,38 @@ def get_vector_n(self):
...
+        shapeDistAmp.append(3)
+
+        # prepare data array
+        distAmp = np.empty(shape=shapeDistAmp, dtype=np.complex128)
+
+        # load data
+        for i, direction in enumerate(["x", "y", "z"]):
+            tmp = h_distAmp[direction][:, :, :]
+            self.rad_series.flush()
+            distAmp[:, :, :, i] = tmp
```
franzpoeschel, line 165: `distAmp[:, :, :, i]` is a non-contiguous memory slice, right? If not, you could load directly into that buffer, but otherwise the current implementation with an intermediate load buffer is fine.

PrometheusPi: Good point. As a physicist I would prefer keeping the polarization as my last component - but this is simply due to common convention. I agree with you that directly writing into a chunk of continuous memory would be more elegant and also easier to read. 
I will change the memory layout according to your suggestion.

franzpoeschel: I think that it's okay to have a different memory layout than on the disk. I just wanted to confirm that I interpreted this correctly; probably a short comment is all that's needed here.

PrometheusPi: Okay - then I will just a comment. Thanks for the feeback @franzpoeschel.

docs/source/usage/plugins/radiation.rst
```diff
@@ -521,16 +521,17 @@ This currently assumes adios output but any openPMD output, such as hdf5, will w
...
+================================= ==============================================================================================================
+Method                            Description
+================================= ==============================================================================================================
+``.get_omega()``                  get frequency :math:`\omega` of virtual detector bin in units of :math:`\mathrm{[1/s]}`
+``.get_vector_n()``               get observation direction :math:`\vec{n}`
+``.get_Spectra()``                get spectrum :math:`\mathrm{d}^2 I / \mathrm{d} \omega \mathrm{d} \Omega` in units of :math:`\mathrm{[Js]}`
+``.get_Polarization_X()``         get spectrum but only for polarization in x-direction
+``.get_Amplitude_x()``            get x-component of complex amplitude (unit: :math:`\mathrm{[\sqrt{Js}]}`)
+``.get_timestep()``               the iteration (timestep) at which the data was produced (unit: PIC-cycles)
+``.get_distributedAmplitude()``   get 4D complex array of distributed amplitudes shape: (N_gpus, N_observer, N_frequency, N_polarization)
```
steindev: 'shape' is supposed to be within the parenthesis?

lib/python/picongpu/extra/plugins/data/radiation.py
```diff
@@ -125,3 +126,40 @@ def get_vector_n(self):
...
+            warnings.warn("Warning: no distributed amplitude available.")
+            return None
+
+        # get shape of distributed amplitude
+        shapeDistAmp = h_distAmp["x"].shape
+        # add 3 vector components
+        shapeDistAmp.append(3)
+
+        # prepare data array
+        distAmp = np.empty(shape=shapeDistAmp, dtype=np.complex128)
```
steindev: Do you really want to create an empty array here, which has garbage values inside? Wouldn't zeroes be better to spot errors ? (Some day, maybe, the size of the y and z components do not equal the size of the x component. Although the assignment later should then throw an error)

I don't have an opinion on this, actually.

steindev:
ping @PrometheusPi

psychocoderHPC:
@PrometheusPi please rebase against dev branch because pre commits has changed and check now python code too

chillenzer:
Steering things up...

psychocoderHPC:
@PrometheusPi is this PR still of interest?

PrometheusPi:
Yes, I will update it now

PrometheusPi:
I will not be able to work on this PR until #5105 is merged.

lib/python/picongpu/extra/plugins/data/radiation.py
PrometheusPi: Since there is no way of ever seeing the empty array since it is filled in the code lines below, I favor having only the empty allocation and no extra setting to zero. However, the array won't be very big, so this optimization is only marginal.

docs/source/usage/plugins/radiation.rst
PrometheusPi: Yes

PrometheusPi:
@steindev @psychocoderHPC @ikbuibui ready for review

lib/python/picongpu/extra/plugins/data/radiation.py
chillenzer: You could, of course, use fancy numpy methods to do so, e.g., create your array such that you can load contiguously and afterwards do a `np.moveaxis()` or `.transpose()` or the like.

lib/python/picongpu/extra/plugins/data/radiation.py
chillenzer: I strongly agree with @steindev: Either you say that this is a premature optimisation (and we all know about how it relates to the tree structure of all evil) and use `np.zeros` or you say that it's necessary, then please make sure that what you say is true by extracting the allocation and initialisation into a separate function. That way you can make it part of the contract of the function that the returned array is always correctly initialised.

lib/python/picongpu/extra/plugins/data/radiation.py
```diff
@@ -127,3 +128,40 @@ def get_vector_n(self):
...
+        try:
+            h_distAmp = self.iteration.meshes["Amplitude_distributed"]
+        except IndexError:
+            warnings.warn("Warning: no distributed amplitude available.")
+            return None
+
+        # get shape of distributed amplitude
+        shapeDistAmp = h_distAmp["x"].shape
+        # add 3 vector components
+        shapeDistAmp.append(3)
```
chillenzer: What's the magic number `3`? Maybe a named constant would be nice.

PrometheusPi: 3 is for the 3 spacial dimensions 
Since there is no `const uint` in python, I personally favor using literals instead. 
What would you prefer here to make clear that is a *never changing* value?

PrometheusPi: @chillenzer I used a class variable to make the global scope of that 3 clear. Is this in agreement with what you wanted?

lib/python/picongpu/extra/plugins/data/radiation.py
PrometheusPi: @steindev and I had just a brief *offline* discussion. Initializing with NaNs will be better because it clearly indicates something going wrong.

steindev:
I am fine with the changes. Would you please squash (and rebase, if necessary)? I guess @chillenzer's comment is resolved as well. Then I will merge.

PrometheusPi:
@steindev squashed to 2 commites

steindev (approved):

PR #4747, opened 2023-11-21
Add openPMD-viewer test

max-lehman14:
A simple example is used and the input parameters are compared to the readout of the openPMD-viewer. This test involves the electromagnetic fields, electron and ion charge density, particle positions and momenta.

PR #4759, opened 2023-12-01
Switch default openPMD ADIOS2 file engine from BP4 to BP5

franzpoeschel:
While BP5 is not supposed to replace BP4 in ADIOS2, it is the better default engine since it requires no configuration to be performant. BP4 required careful specification of `InitialBufferSize` (which most users did probably not bother with, or were not even aware of it). Additionally, it has support for the `PerformDataWrite()` API call, which PIConGPU uses to flush data to disk at custom times.

Since this is a common header, this affects the default settings of the following plugins:

* PhaseSpace plugin
* MacroParticleCounter
* openPMD I/O plugin
* Radiation plugin
* Binning plugin

(Some of these set HDF5 by default, so they are only affected if openPMD-api is built without HDF5)

The BP4 engine can still be used. Since it requires specific configuration anyway, specifying the engine type is the least of troubles.

TODO:

- [x] Since ADIOS2 <= 2.9.1 had a bug that resulted in unreadable data created from PIConGPU when using Blosc2, activate this only when ADIOS2 is newer than that. (ADIOS2 v2.8 technically did not suffer from this since it still used Blosc1. But I never tested if that actually works with PIConGPU and I won't bother with it either.) 
    https://github.com/ornladios/ADIOS2/issues/3504

### DIFF
--- a/include/picongpu/CMakeLists.txt
+++ b/include/picongpu/CMakeLists.txt
@@ -535,6 +535,20 @@ if(PIC_HAVE_openPMD)
         SYSTEM PRIVATE
         $<TARGET_PROPERTY:toml11::toml11,INTERFACE_INCLUDE_DIRECTORIES>)
     target_link_libraries(picongpu-hostonly PRIVATE toml11::toml11)
+
+    # This is a workaround for a missing API call in the openPMD-api,
+    # which currently does not re-export backend version information.
+    # https://github.com/openPMD/openPMD-api/issues/1563
+    if(openPMD_HAVE_ADIOS2)
+        foreach(target picongpu picongpu-hostonly)
+            target_compile_definitions(
+                picongpu
+                PRIVATE
+                ADIOS2_VERSION_MAJOR=${ADIOS2_VERSION_MAJOR}
+                ADIOS2_VERSION_MINOR=${ADIOS2_VERSION_MINOR}
+                ADIOS2_VERSION_PATCH=${ADIOS2_VERSION_PATCH})
+        endforeach()
+    endif()
 endif()
 
 ## annotate with RPATH's
--- a/include/picongpu/CMakeLists.txt
+++ b/include/picongpu/CMakeLists.txt
@@ -542,7 +542,7 @@ if(PIC_HAVE_openPMD)
     if(openPMD_HAVE_ADIOS2)
         foreach(target picongpu picongpu-hostonly)
             target_compile_definitions(
-                picongpu
+                ${target}
                 PRIVATE
                 ADIOS2_VERSION_MAJOR=${ADIOS2_VERSION_MAJOR}
                 ADIOS2_VERSION_MINOR=${ADIOS2_VERSION_MINOR}
--- a/include/picongpu/plugins/common/openPMDDefaultExtension.hpp
+++ b/include/picongpu/plugins/common/openPMDDefaultExtension.hpp
@@ -46,6 +46,21 @@ namespace picongpu
             auto getADIOSExtension = []()
             {
                 auto availableExtensions = ::openPMD::getFileExtensions();
+                /*
+                 * Using macros for detecting the ADIOS2 version is not very nice since the openPMD-api can be
+                 * recompiled against different backends and backend versions without affecting the ABI. Upgrading
+                 * ADIOS2 will require recompiling openPMD-api, but only relinking PIConGPU. The below #ifdef logic
+                 * will then not be updated.
+                 *
+                 * A functionality for doing a runtime query here is unfortunately missing in the openPMD-api at the
+                 * moment.
+                 * https://github.com/openPMD/openPMD-api/issues/1563
+                 *
+                 * The reason for requiring ADIOS2 >= v2.9.2 before enabling BP5 is this bug, fixed in version v2.9.2:
+                 * https://github.com/ornladios/ADIOS2/issues/3504
+                 */
+#if openPMD_HAVE_ADIOS2
+#    if ADIOS2_VERSION_MAJOR * 10000 + ADIOS2_VERSION_MINOR * 100 + ADIOS2_VERSION_PATCH >= 20902
                 if(std::find(availableExtensions.begin(), availableExtensions.end(), "bp5")
                    != availableExtensions.end())
                 {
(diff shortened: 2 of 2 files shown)


steindev (dismissed):

steindev:
@franzpoeschel would you please execute the KHI runtime test? It is in `share/picongpu/tests/KHI_growthRate`. To do so, obtain a node on hemera, clone the setup, and execute `bin/ci.sh` in it.

franzpoeschel:
Does that test even create openPMD output? I let it run now (not on Hemera, that's down until friday) and it only wrote `fields_energy.dat`, `testresult.log` and `testresult_1D.png`.

PrometheusPi:
@franzpoeschel you are right. The KHI test does not create openPMD data. It uses text based output from the energy plugin. Thus @steindev suggestsions does not work. Perhaps you can use the FieldAbsorber test instead (authored by @steindev). It uses openPMD.

include/picongpu/plugins/common/openPMDDefaultExtension.hpp
```diff
@@ -46,19 +46,43 @@ namespace picongpu
...
-                    {
-                        return "bp4";
-                    }
-                }
                 /*
-                 * BP4 engine is always available in all supported ADIOS2 versions,
-                 * but the bp4 extensions is new in openPMD, so we might need to
-                 * fallback to "bp".
+                 * Using macros for detecting the ADIOS2 version is not very nice since the openPMD-api can be
+                 * recompiled against different backends and backend versions without affecting the ABI. Upgrading
```
PrometheusPi, line 51:
```suggestion
                 * recompiled against different backends and backend versions without affecting the API. Upgrading
```

include/picongpu/plugins/common/openPMDDefaultExtension.hpp
```diff
@@ -46,19 +46,43 @@ namespace picongpu
...
+                 * The reason for requiring ADIOS2 >= v2.9.2 before enabling BP5 is this bug, fixed in version v2.9.2:
+                 * https://github.com/ornladios/ADIOS2/issues/3504
                  */
-                return "bp";
+#if openPMD_HAVE_ADIOS2
+#    if ADIOS2_VERSION_MAJOR * 10000 + ADIOS2_VERSION_MINOR * 100 + ADIOS2_VERSION_PATCH >= 20902
+                if(std::find(availableExtensions.begin(), availableExtensions.end(), "bp5")
+                   != availableExtensions.end())
+                {
+                    // Engine available in ADIOS2 >= v2.8
```
PrometheusPi, line 67: If it is available for ADIOS2 >= v2.8, why do we check for >= 2.9.2 ?

PrometheusPi (changes requested):
minor typo and a question
(and possible tests to be run, if hemera becomes available again)

include/picongpu/plugins/common/openPMDDefaultExtension.hpp
franzpoeschel: The reason is in the comment above:
```cpp
                /*
                 * The reason for requiring ADIOS2 >= v2.9.2 before enabling BP5 is this bug, fixed in version v2.9.2:
                 * https://github.com/ornladios/ADIOS2/issues/3504
                 */
```

franzpoeschel:
> @franzpoeschel you are right. The KHI test does not create openPMD data. It uses text based output from the energy plugin. Thus @steindev suggestsions does not work. Perhaps you can use the FieldAbsorber test instead (authored by @steindev). It uses openPMD.

That test however overrides the default file extension to `--openPMD.ext bp`. Should I run it without that switch and see what happens?

include/picongpu/plugins/common/openPMDDefaultExtension.hpp
PrometheusPi: Okay - thanks for the explanation

include/picongpu/plugins/common/openPMDDefaultExtension.hpp
franzpoeschel: This is no typo, I'm referring to the application binary interface of openPMD-api. The API ideally does not change across different versions, but recompiling openPMD with different backends will not even change the ABI, meaning that you only have to re-link, but not recompile PIConGPU.
Under these circumstances it's not really ideal to rely on compile-time information about the backends like it's done here. But it's only about the default values, so it's mostly fine still.

PrometheusPi: Thanks for the clarification :+1: 
Just for others who are not aware of this abbreviation (like me):
*ABI = Application Binary Interface*

PrometheusPi (dismissed):

PrometheusPi:
@franzpoeschel How are the tests looking?

PrometheusPi:
@max-lehman14 This change will affect your pull request #4747 - we need to handle different ADIOS postfixes.

franzpoeschel:
> @franzpoeschel How are the tests looking?

The FieldAbsorber test explicitly sets `--openPMD.ext bp`, so it is unaffected. Setting any other extension here (such as `bp5`) leads to a failing test since the Python side tries reading `simData_%T.bp` and does not find those.
This is the correct behavior IMO, the test should be explicit about the filename and not rely on (changing) defaults, so it can rely on the name later on.

franzpoeschel:
> @max-lehman14 This change will affect your pull request #4747 - we need to handle different ADIOS postfixes.

I see `--openPMD.ext h5` in that PR, so it's probably not affected?

PrometheusPi:
@franzpoeschel there will be both ADIOS and HDF5 tested soon.

include/picongpu/CMakeLists.txt
```diff
@@ -535,6 +535,20 @@ if(PIC_HAVE_openPMD)
...
         $<TARGET_PROPERTY:toml11::toml11,INTERFACE_INCLUDE_DIRECTORIES>)
     target_link_libraries(picongpu-hostonly PRIVATE toml11::toml11)
+
+    # This is a workaround for a missing API call in the openPMD-api,
+    # which currently does not re-export backend version information.
+    # https://github.com/openPMD/openPMD-api/issues/1563
+    if(openPMD_HAVE_ADIOS2)
+        foreach(target picongpu picongpu-hostonly)
+            target_compile_definitions(
+                picongpu
```
franzpoeschel:
```suggestion
                target
```

PrometheusPi (approved):

PR #4763, opened 2023-12-06
Fix Notify period not being set in Binning Plugin

ikbuibui:
Fix notify not being set correctly, also swtich binners to IPlugins from INotify

### DIFF
--- a/include/picongpu/plugins/binning/Binner.hpp
+++ b/include/picongpu/plugins/binning/Binner.hpp
@@ -24,6 +24,8 @@
 #include "picongpu/plugins/binning/WriteHist.hpp"
 #include "picongpu/plugins/binning/utility.hpp"
 
+#include <pmacc/mpi/MPIReduce.hpp>
+
 #include <cstdint>
 #include <memory>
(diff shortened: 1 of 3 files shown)


psychocoderHPC (approved):

include/picongpu/plugins/binning/Binner.hpp
```diff
@@ -24,6 +24,8 @@
 #include "picongpu/plugins/binning/WriteHist.hpp"
 #include "picongpu/plugins/binning/utility.hpp"
 
+#include <pmacc/mpi/MPIReduce.hpp>
```
PrometheusPi, line 27: why do you need to linclude `MPIReduce.hpp`?

PrometheusPi (approved):

include/picongpu/plugins/binning/Binner.hpp
ikbuibui: I was missing this include in the file previously, while still using a reduce operation. I was sneaking in this small unrelated change.

PR #4740, opened 2023-11-15
Suggestions for the Binning plugin

franzpoeschel:
1. commit: Use optimized `storeChunk()` calls (unique_ptr or raw pointer) instead of fake shared pointer
2. commit: Add logic for configuring optional (compile-time) parameters and use this for making the openPMD extension and JSON config configurable 
3. commit: Keep `Series` open across calls to the plugin in order to enable iteration encodings that require this (group-based, variable-based, streaming)
4. commit: filename padding

PR #4764, opened 2023-12-07
Don't pass HostBuffer to lockstep kernel in Binning Plugin

ikbuibui:
Do time averaging, normalization on device, and do the MPI reduce afterwards.
Earlier the MPI reduction was done first, and then a host buffer was passed to the lockstep kernels doing the time averaging and normalization. This was not crashing, however it is dangerous and is probably slower than running the kernels on device first and then doing the reduction. 
Also we now actually use the jsonCfg (introduced in  #4740) for openPMD instead of silently ignoring it.

### DIFF
--- a/include/picongpu/plugins/binning/Binner.hpp
+++ b/include/picongpu/plugins/binning/Binner.hpp
@@ -145,7 +145,7 @@ namespace picongpu
             std::unique_ptr<HostDeviceBuffer<TDepositedQuantity, 1>> histBuffer;
             uint32_t accumulateCounter = 0;
             mpi::MPIReduce reduce{};
-            bool isMaster = false;
+            bool isMain = false;
             WriteHist histWriter;
             std::optional<::openPMD::Series> m_series;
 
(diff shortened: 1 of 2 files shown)


psychocoderHPC (approved):

PrometheusPi (approved):
From my limited understanding, this looks good. 
In your PR description, you mentioned not only time averaging but also normalization. Where does this happen in this change and how do you do it.

ikbuibui:
> From my limited understanding, this looks good. In your PR description, you mentioned not only time averaging but also normalization. Where does this happen in this change and how do you do it.

This PR changes the order of where the normalization and the time averaging are called in the code to before all MPI ranks reduce their local binning data to the main rank to write out. This code for normalization is in lines 228-253.
The actual logic of the normalization is in lines 83-124 and what it does is to go over every bin in parallel and for each bin it calculates the bin volume and does $binnedQuantityValue / binVolume$ . For each bin the bin volume is calculated in a for loop going over all axes, asking for the width of the bin along that axis and then taking the product of all of these bin widths. 
Lines 249-252 deal with the changes in the units which comes with such a normalization.

PrometheusPi:
Thanks @ikbuibui that makes it clear to me. 
I thought of a counting normalization (which is not easily parallelized) but a bin with normalization is perfect for parallelization and then this can be done also by all MPI ranks in parallel and independent.

PR #4766, opened 2023-12-11
Implement checkpoint and restart capability to the binning plugin

ikbuibui:
Implement checkpoint and restart capability to the binning plugin. 
Checkpointing and restarting is called by PIConGPU for the binningDispatcher, which handles checkpoint and restart for all individual "binners".
Currently the checkpoint information is stored in the default checkpoint directory, in a sub-folder called binningOpenPMD, and the file name and file type for the checkpoints aren't seperately configurable. File name and file type are the same as the ones created during a normal execution data dump.
Also added some previously missing includes

### DIFF
--- a/include/picongpu/plugins/binning/Binner.hpp
+++ b/include/picongpu/plugins/binning/Binner.hpp
@@ -196,7 +196,7 @@ namespace picongpu
                 ++accumulateCounter;
                 if(accumulateCounter >= binningData.dumpPeriod)
                 {
-                    auto bufferExtent = this->histBuffer->getDeviceBuffer().getDataSpace();
+                    auto bufferExtent = this->histBuffer->getHostBuffer().getDataSpace();
 
                     // Do time Averaging
                     if(binningData.dumpPeriod > 1 && binningData.timeAveraging)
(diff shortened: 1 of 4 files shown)


include/picongpu/plugins/binning/Binner.hpp
```diff
@@ -300,10 +300,103 @@ namespace picongpu
 
             void checkpoint(uint32_t currentStep, const std::string restartDirectory) override
             {
+                /**
+                 * State to hold, accumulateCounter and hReducedBuffer
+                 */
+
+                // do the mpi reduce (can be avoided if the notify did a data dump and histBuffer is empty)
+                this->histBuffer->deviceToHost();
+                auto bufferExtent = this->histBuffer->getDeviceBuffer().getDataSpace();
```
psychocoderHPC: can be `getHostBuffer()`

include/picongpu/plugins/binning/BinningCreator.hpp
```diff
@@ -46,8 +46,8 @@ namespace picongpu
 
             /**
              * Creates a binner from user input and adds it to the vector of all binners
-             * @param binnerOutputName filename for openPMD output. It must be unique or will cause overwrites during
-             * data dumps
+             * @param binnerOutputName filename for openPMD output. It musit be unique or will cause overwrites during
```
psychocoderHPC: `musit`-> must`

psychocoderHPC (changes requested):
only small changes required

include/picongpu/plugins/binning/Binner.hpp
ikbuibui: Yes, you are right. Also changed to getHostBuffer() in line 199.

include/picongpu/plugins/binning/BinningCreator.hpp
ikbuibui: fixed

psychocoderHPC (approved):

PR #4767, opened 2023-12-14
Add PMacc example - Solving the 2D HeatEquation

ikbuibui:
Added a more physically motivated example for PMacc. This is a simple solver for the 2D heat equation. 
The attempt was to make it more accessible than the game of life (GoL) example, so it skips over some of the command line configuration options included in the GoL example too. 
A README is included which explains how to build and run this example.
It aims to introduce some of the commonly used classes of PMacc in a logically linear fashion in the main file.
In many places it is heavily influenced by the GoL example and also depends on GatherSlice from GoL.

### DIFF
--- a/include/pmacc/mpi/MPIReduce.hpp
+++ b/include/pmacc/mpi/MPIReduce.hpp
@@ -67,7 +67,7 @@ namespace pmacc
 
             /** defines if the result of the MPI operation is valid
              *
-             * The reduction method reduceMethods::Reduce is used.
+             * The reduction method reduceMethods::AllReduce is used.
              *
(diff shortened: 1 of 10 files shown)


psychocoderHPC:
The code looks good. We need to add the example to the CI to be sure all is compiling well.
For PMacc examples we need to do it by hand: https://github.com/ComputationalRadiationPhysics/picongpu/blob/267ec7c471015e3e688cffbaf7c54566a464d9b4/share/ci/run_pmacc_tests.sh#L107-L115

psychocoderHPC (approved):

PR #4745, opened 2023-11-20
fix collision

psychocoderHPC:
With #4685 we missed to rewrite collision to support a frame size independent of the number of cells in the supercell.

This fix rewrites the required code to support any frame size and optimizes the particle access via the particle index.
This PR is optimizing the performance by a factor of ~20.

@pordyna Could you please validate the results? The README is pointing to SMILEI results which are not available, therefore a validation from my side is not possible.

A minimal fix was not possible therefore a larger refactoring was required.

# Review:
Check the code changes without white space changes! https://github.com/ComputationalRadiationPhysics/picongpu/pull/4745/files?w=1

PR #4772, opened 2023-12-20
fix collision (again)

psychocoderHPC:
bug was introduced with #4745

The frame index which is used to calculate particle index within the supercell is not allowed to be updated within the frame lamda, else an accelerator with less threads per block than slots within a frame will produce errors e.g. illegal memory access.


pordyna:
Nice catch. I don't know yet if it fixes my problem, but this was definitely a bug. Tbh it should not solve it, why would I have less threads on a gpu.

pordyna (approved):

PR #4779, opened 2024-01-09
Remove hemera k80 and k20 partition

PrometheusPi:
This pull request removes the k80 and k20 partition files and documentation for the HZDR hemera cluster. 
Both queues are no longer available. 

- [x] check if k20/k80 will be finally removed (ticket opened)


PrometheusPi:
The cluster admins confirmed that k20 and k80 will not return.

psychocoderHPC (approved):

PR #4780, opened 2024-01-12
Correct spack options for boost

chillenzer:
Just a tiny correction of the `spack` commandline to run (neither `program_options` nor `atomic`) are default options for `spack`'s `boost` installation). Checked on my local system.

PR #4781, opened 2024-01-12
Typos in comments

chillenzer:
While reading through some random files, I've found some random typos.

~~fb04f5516a7bcaa4615860ccc93d52d5a736b097 is from #4780 already.~~

I'm not 100% sure if the change in [InitialiserController.hpp](https://github.com/ComputationalRadiationPhysics/picongpu/compare/dev...chillenzer:topic-typosAndStuff?expand=1#diff-4b75d783c44cdb3d2a0316b1e1e387b289fb6d62e09834d62ba680f162206be9) is the one that was intended.

Also, I've found it surprising that `moveParticle` does not handle movement over more than one cell. (I think) I've found out why, so I've clarified that in the docstring. Feel free to enlighten me if I misunderstood.

### DIFF
--- a/include/picongpu/particles/MoveParticle.hpp
+++ b/include/picongpu/particles/MoveParticle.hpp
@@ -34,12 +34,14 @@ namespace picongpu
          * in case the new position is outside of the cell we adjust
          * the cell index and the multiMask of the particle.
          *
-         * we assume that CFL condition is fulfilled
+         * as everywhere in PIConGPU, we assume that the particle moves at most
+         * one cell per step. This has to be ensured by the caller and
+         * violations are NOT checked for or caught during execution.
          *
          * @tparam T_Particle particle type
          * @param particle particle handle
-         * @param newPos new position relative to the current cell, in units of cells
-         * newPos must be [-1.0, 2.0) (following CFL)
+         * @param newPos new position relative to the current cell's origin, in
+         * units of cells, so all entries of newPos must be [-1.0, 2.0) (see above)
          * @return whether the particle has left the original cell
          */
         template<typename T_Particle>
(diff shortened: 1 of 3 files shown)


psychocoderHPC:
Please rebase the PR we removed some classes and therefore the CI is failing.

chillenzer:
Rebased onto `dev`, so first comment does no longer apply.

include/picongpu/particles/MoveParticle.hpp
```diff
@@ -34,10 +34,12 @@ namespace picongpu
          * in case the new position is outside of the cell we adjust
          * the cell index and the multiMask of the particle.
          *
+         * we assume that CFL condition is fulfilled
```
psychocoderHPC: Not sure if CFL (Courant–Friedrichs–Lewy condition) is correct here @PrometheusPi @steindev could you please have a look.

psychocoderHPC (changes requested):
I am not sure if one point is correct so I pinged Klaus and Richard.

include/picongpu/particles/MoveParticle.hpp
BrianMarre: well its is CFL adjacent, but not strictly CFL.

We guarantee that a particle may never move more than one cell length separately from the CFL condition as described here.

https://picongpu.readthedocs.io/en/latest/models/AOFDTD.html#substepping

But I am not exactly sure how that will interact with the differrent CFL condiions of AOFDTD.

chillenzer: Well, I'm happy to revise this into something more explicit like "We guarantee that a particle may never move more than one cell length". Just think that there should be something in there stating that (if not checking that in code) -- at best with a hint about the physical/numerical explanation.

psychocoderHPC: IMO we can safely write it based on the PMacc restriction that a particle is not allowed to move more than one cell.
This restriction is not physically motivated it is simply an implementation restriction.
Additionally, there are restrictions from PIConGPU. The current deposition schema also restricts the largest move to one cel and field solvers have higher restrictions depending on the solver type.

psychocoderHPC (approved):

PR #4792, opened 2024-01-18
Pre-commit

### LINKED ISSUE #4787: Remove `correctBranchPR`
[This](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/test/correctBranchPR) seems to be outdated because there is no `master` branch (anymore). Shall I remove it?

chillenzer:
Introducing [`pre-commit`](https://pre-commit.com/): After installing (see the [docs](https://github.com/chillenzer/picongpu/blob/topic-precommit/docs/COMMIT.md#pre-commit)) it runs checks before each commit. Installing this is optional and running the checks can be prevented on a case-by-case basis via `git commit --no-verify [...]`.

I've currently implemented the following checks (which -- if possible -- immediately apply the necessary changes as unstaged changes):
- `clang-format`: obvious
- `no-commit-to-branch`: Forbids committing directly to `dev` (enforcing opening PRs)
- `check-merge-conflict`: Checks if there are any merge conflict markers in the diffs
- `trailing-whitespace`: Remove trailing whitespaces
- `end-of-file-fixer`: Files end on precisely one `\n`
- `flake8`: obvious
- `black`: obvious

In the process, I've also removed some of the obsolete tests and ran the hooks on all files as `Third Party`. I've opened #4793 to track further improvements around this subject.

Closes #4787


chillenzer:
PS: Also closes #4787

steindev (dismissed):

.gitlab-ci.yml
```diff
@@ -42,7 +42,7 @@ pull-request-validation:
     - source $CI_PROJECT_DIR/share/ci/git_merge.sh
     - pip3 install -U flake8 pyflakes
     # Test Python Files for PEP8 conformance
-    - flake8 --exclude=thirdParty .
+    - flake8 --exclude=thirdParty --extend-ignore E203 --max-line-length "120" .
```
psychocoderHPC, line 47: what is `E203` could you add a comment above

.gitlab-ci.yml
```diff
@@ -42,7 +42,7 @@ pull-request-validation:
     - source $CI_PROJECT_DIR/share/ci/git_merge.sh
     - pip3 install -U flake8 pyflakes
```
psychocoderHPC, line 43: Does it make sense to remove the explicit tests here and use `git-pre-commit`. This should prevent our tests and the rules in pre-commit will be de-sync at some point.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+    args: [-b, dev]
+  - id: check-merge-conflict
+  - id: trailing-whitespace
+    exclude_types: [markdown, rst]
+  - id: end-of-file-fixer
+- repo: https://github.com/PyCQA/flake8
+  rev: 7.0.0
+  hooks:
+  - id: flake8
+    args: [--extend-ignore, E203, --max-line-length, "120"]
```
psychocoderHPC, line 24: Please add a comment what `E203` is.

psychocoderHPC (changes requested):
IMO we should use `pre-commit` in the CI if there is an mode to test only without applying the changes.

.gitlab-ci.yml
psychocoderHPC: We can only use pre-commit if we can test only without changing the code.

psychocoderHPC: I think the feature will not come https://github.com/pre-commit/pre-commit/issues/2144 but we can run `pre-commit` and then check if we have unstaged files.

chillenzer:
> IMO we should use `pre-commit` in the CI if there is an mode to test only without applying the changes.

That is the next step, I'd say. For the moment, I would still take the CI as the ground truth and have `pre-commit` as a recommendation on how to get close to abiding by these rules. Having it like this will showcase if there are any inconsistencies or other drawbacks with using `pre-commit`. Working through #4787, we could then make the full transition in the next iteration after we have a bit of experience with it in the team (and worked out a strategy how precisely we want to integrate it).

.gitlab-ci.yml
chillenzer: But the discussion suggests it's pretty straightforward as we are already doing short-circuiting:
1. Install `pre-commit`
2. Run `pre-commit`
3. If `EXIT_CODE != 0`: fail pipeline, else we can be sure that running `pre-commit` didn't actually change anything and we can continue as usual.

(There could theoretically be some really badly designed hooks that apply changes without reporting a failure to `pre-commit` but that sounds like sufficiently bad design to reject them on these grounds alone.)

chillenzer: Following my other comment, I think we should move this discussion into #4787.

psychocoderHPC (approved):

steindev (approved):

PR #4794, opened 2024-01-19
Update fwkt_v100_picongpu.profile.example to new modules (CUDA 12.1)

steindev:
Tested by compiling and running the KHI example.


PrometheusPi (approved):
@steindev why were these changes necessary?

steindev:
> @steindev why were these changes necessary? 

Now I can run PyTorch and PIConGPU in the same environment.

PR #4798, opened 2024-01-26
update system profiles

psychocoderHPC:
- update and tested hemera and HZDR dev system profiles
- update in all other profiles with openpmd-api modules the version to 0.15.2

On systems where I have no access, I set the version explicitly to openpmd-api/0.15.2. If the user is loading it and it is not available there will be a clear error message.

I opened a ticket to update openpmd-api on ORNL frontier.

### DIFF
--- a/etc/picongpu/bash-devServer-hzdr/gpu_v100_install.sh
+++ b/etc/picongpu/bash-devServer-hzdr/gpu_v100_install.sh
@@ -1,10 +1,10 @@
-spack install --reuse cmake@3.24.2 %gcc@10.2.0
-spack load cmake@3.24.2 %gcc@10.2.0
+spack install --reuse cmake@3.26.3 %gcc@12.2.0
+spack load cmake@3.26.3 %gcc@12.2.0
 
 echo "openpmd-api:"
-spack install --reuse openpmd-api@0.14.5 %gcc@10.2.0 \
-    ^adios2@2.8.3 +cuda cuda_arch=70\
-    ^cmake@3.24.2 \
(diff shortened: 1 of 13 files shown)


psychocoderHPC:
OpenPMD with blosc will be added in a separate PR as soon as the IT compiled it for GPU and CPU

etc/picongpu/bash-devServer-hzdr/gpu_v100_install.sh
```diff
@@ -1,10 +1,10 @@
-spack install --reuse cmake@3.24.2 %gcc@10.2.0
-spack load cmake@3.24.2 %gcc@10.2.0
+spack install --reuse cmake@3.26.3 %gcc@12.2.0
```
steindev, line 1: You did not change gcc to version 12.1. in this profile?

etc/picongpu/bash-devServer-hzdr/gpu_v100_picongpu.profile.example
steindev, line 1: Same here, you did not change gcc to version 12.1?

etc/picongpu/bash-devServer-hzdr/gpu_v100_install.sh
psychocoderHPC: fixed it

etc/picongpu/bash-devServer-hzdr/gpu_v100_picongpu.profile.example
psychocoderHPC: It is a different system, fwk394 is using 12.2.0

steindev (approved):

PR #4750, opened 2023-11-21
Topic remove preprocessor struct generator

psychocoderHPC:
Motivation: The generator is hard to maintain and the usage is not easy
either. With modern C++ and constexpr there is no need for the preprocessor
magic anymore.

- remove the struct generator
- remove the usage of the preprocessor struct generator to create parameters for pre-defined density profiles.

- [x] rebase against #4748

## Note

This PR will close #4748 automatically after it is merged. #4748 produces CI issues those will be gone because this PR is removing the struct generator, therefore this PR must be merged before  #4748.

PR #4801, opened 2024-01-30
add Pusher Test

MRZirper:
- [x] rebase against dev branch after #4820 is merged

single Particle with constant speed in a homogenous magnetic field. Test the radius change, phase drift and its scaling (just for Boris Pusher yet)

### DIFF
--- a/share/picongpu/tests/Pusher/README.rst
+++ b/share/picongpu/tests/Pusher/README.rst
@@ -1,0 +1,23 @@
+Pusher: Testing its functionality in PIConGPU
+=============================================
+
+An electron with a velocity of 0.5c moves perpendicular to a homogeneous magnetic field (in the z-direction).
+The electron should, therefore, move in a perfect circle since there is no acceleration in the direction of movement.
+The test compares the radii of a particle simulated by PIConGPU. 
+The radius is calculated by the position of the particle and its momentum (r =  p / (q * B) ).
+If the relative change between two periods is greater than epsilon_position = 5e-5 or epsilon_momentum = 1e-5 respectively, the test fails.
+The calculation of the radius with the position of the particle has a higher error than the calculation with the momentum.
+Therefore the test compares the change of the radius against two different epsilons.
+Also, the absolute phase change during one turn is regarded. 
+It should be smaller than delta = 0.16 rad.
+
+the acceptance thresholds for position-, momentum- and phase-error were determined as follows:
+
+epsilon: The greatest change in the radius measured between 2 turns (2000 Turns simulated) is approximately 4e-6 for the calculation with the momentum.
+         For the calculation with the position of the particle, this value is approximately 2.5e-5.
+         Those values doubled yield the two epsilons as an approximation 
+         of the maximal uncertainty PIConGPU should have. Therefore, the test uses this value to check the radius change.
+
+delta: The measured phase difference between 2 revolutions is approximately 0.08rad. This doubled is the maximal error to be accepted for the phase difference between 2 turns.
+
+PIConGPU run September 2023; 50 steps per turn.
--- a/share/picongpu/tests/Pusher/bin/ci.sh
+++ b/share/picongpu/tests/Pusher/bin/ci.sh
@@ -1,0 +1,115 @@
+#!/bin/bash
+#
+# This file is part of PIConGPU.
+# Copyright 2023 PIConGPU contributors
+# Authors: Mika Soren Voss, Hannes Wolf, Klaus Steiniger
+# License: GPLv3+
+#
+
+function absolute_path()
+{
+    builtin cd -- $1 && pwd
+}
+
+help()
+{
+  echo "Execute a single particle simulation in a homogenious magnetic field to validate the pusher."
+  echo "Simulation results are validated against by checking the change of radius during the simulation and the quantitative phase shift"
+  echo ""
+  echo "Usage: ci.sh"
+  echo ""
+  echo "Options"
+  echo "-h | --help                   - show help"
+  echo "-t | --flag                   - flag number of cmakeFlags"
+  echo ""
+}
+
+
+#####################
+## option handling ##
+#####################
+# options may be followed by
+# - one colon to indicate they have a required argument
+OPTS=`getopt -ot:h help flag: -- "$@"`
+if [ $? != 0 ] ; then
+    # something went wrong, getopt will put out an error message for us
+    exit 1
+fi
+
+eval set -- "$OPTS"
+
+# standard value of -t|--flag:
+declare -i flag=0
+
+# parser
+while true ; do
+    case "$1" in
+        -t|--flag)
+            declare -i flag=$2
+            shift
+            ;;
+        -h|--help)
+            echo -e "$(help)"
+            shift
+            exit 0
+            ;;
+        --) shift; break;;
+    esac
+    shift
+done
+
+
+############################
+## build and run picongpu ##
+############################
+if [ -d "./include" ] ; then
+  pic-build -t $flag
+  ret_build=$?
+
+else
+  echo "Execute ci.sh from the directory where the simulation include dir is located!"
+  exit 1
+fi
+
+if [ $ret_build -eq 0 ] ; then
+  ## create simulation data directory
+  date_stamp=$(date +"%F-%h-%M-%S")
+  simPath="./simOutput_$date_stamp"
+
+  if [ -d "$simPath" ] ; then
+      echo "Destination path already in use, cannot create new folder" >&2
+      exit 1
+  fi
+
+  mkdir -p $simPath
+
+  # use absolut path
+  simPath=$(absolute_path $simPath)
+
+  cd $simPath
+  echo "Run!"
+  mpiexec -n 1 ../bin/picongpu -d 1 1 1 -g 64 64 32 -s 100 --fieldBackground.duplicateFields --periodic 1 1 1 --openPMD.period 1 --openPMD.file simData --openPMD.ext bp
+  cd ..
+
+  dataPath=$(absolute_path "$simPath/openPMD")
+else
+  echo "could not build sumulation"
+  exit 2
+
+fi
+
+
+#################################
+## validate simulation results ##
+#################################
+if [ -d $dataPath ] ; then
+    echo "Validate!"
+    ./bin/validate.sh -d "$dataPath/simData_%T.bp"
+else
+      echo "dataPath does not exist"
+      exit 3
+fi
+
+ret=$?
+
+exit $ret
--- a/share/picongpu/tests/Pusher/etc/picongpu/1.cfg
+++ b/share/picongpu/tests/Pusher/etc/picongpu/1.cfg
@@ -1,0 +1,73 @@
+# Copyright 2013-2023 Heiko Burau, Rene Widera, Felix Schmitt, Axel Huebl,
+#                     Franz Poeschel
+#
+# This file is part of PIConGPU.
+#
+# PIConGPU is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# PIConGPU is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PIConGPU.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+
+##
+## This configuration file is used by PIConGPU's TBG tool to create a
+## batch script for PIConGPU runs. For a detailed description of PIConGPU
+## configuration files including all available variables, see
+##
+##                      docs/TBG_macros.cfg
+##
+
+
+#################################
+## Section: Required Variables ##
+#################################
+
+TBG_wallTime="1:00:00"
+
+TBG_devices_x=1
+TBG_devices_y=1
+TBG_devices_z=1
+
+TBG_gridSize="64 64 32"
+TBG_steps="100"
+
+TBG_periodic="--periodic 1 1 1"
+
+
+#################################
+## Section: Optional Variables ##
+#################################
+
+TBG_openPMD="--openPMD.period 1 --openPMD.file simData --openPMD.ext bp"
+
+
+#################################
+## Section: Program Parameters ##
+#################################
+
+# Duplicate E and B field storage inside field background to improve performance at cost of additional memory
+TBG_fieldBackground="--fieldBackground.duplicateFields"
+
+TBG_deviceDist="!TBG_devices_x !TBG_devices_y !TBG_devices_z"
+
+TBG_programParams="-d !TBG_deviceDist         \
+                   -g !TBG_gridSize           \
+                   -s !TBG_steps              \
+           !TBG_fieldBackground         \
+                   !TBG_periodic              \
+                   !TBG_plugins               \
+                   --versionOnce"
+
+# TOTAL number of devices
+TBG_tasks="$(( TBG_devices_x * TBG_devices_y * TBG_devices_z ))"
+
+"$TBG_cfgPath"/submitAction.sh
--- a/share/picongpu/tests/Pusher/include/picongpu/param/particle.param
+++ b/share/picongpu/tests/Pusher/include/picongpu/param/particle.param
@@ -1,0 +1,104 @@
+/* Copyright 2013-2023 Axel Huebl, Rene Widera, Benjamin Worpitz,
+ *                     Richard Pausch, Klaus Steiniger
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/** @file
+ *
+ * Configurations for particle manipulators. Set up and declare functors that
+ * can be used in speciesInitialization.param for particle species
+ * initialization and manipulation, such as temperature distributions, drifts,
+ * pre-ionization and in-cell position.
+ */
+
+#pragma once
+
+#include "picongpu/particles/manipulators/manipulators.def"
+#include "picongpu/particles/startPosition/functors.def"
+
+#include <pmacc/math/operation.hpp>
+
+#include <limits>
+
+
+namespace picongpu
+{
+    namespace particles
+    {
+        /** a particle with a weighting below MIN_WEIGHTING will not
+         *      be created / will be deleted
+         *
+         *  unit: none */
+        constexpr float_X MIN_WEIGHTING = std::numeric_limits<float_X>::min();
+
+        namespace manipulators
+        {
+            /** Define Lorentz factor of initial particle drift. */
+            struct DriftParam
+            {
+                static constexpr float_64 gamma = 1.1547; // beta: 0.5
+                /** Define initial particle drift direction vector.
+                 *  This vector is used in struct DriftParam.
+                 */
+                static constexpr auto driftDirection = float3_X(1.0, 0.0, 0.0);
+            };
+            /** Definition of manipulator that assigns a drift in X
+             *  using parameters from struct DriftParam.
+             */
+            using AssignXDrift = unary::Drift<DriftParam, pmacc::math::operation::Assign>;
+
+        } // namespace manipulators
+
+        namespace startPosition
+        {
+            /** Define initial in-cell particle position used as parameter in OnePosition functor.
+             *
+             * Here, macro-particles sit directly in lower corner of the cell.
+             *
+             * This defines the type InCellOffset_t
+             * where every instance of this type has the preset values defined here.
+             */
+            struct OnePositionParameter
+            {
+                /** Maximum number of macro-particles per cell during density profile evaluation.
+                 *
+                 * Determines the weighting of a macro particle as well as the number of
+                 * macro-particles which sample the evolution of the particle distribution
+                 * function in phase space.
+                 *
+                 * unit: none
+                 */
+                static constexpr uint32_t numParticlesPerCell = 1u;
+
+                static constexpr auto inCellOffset = float3_X(0., 0., 0.);
+            };
+            /** Definition of OnePosition start position functor that
+             * places macro-particles at the initial in-cell position defined above.
+             */
+            using OnePosition = OnePositionImpl<OnePositionParameter>;
+
+        } // namespace startPosition
+
+        /** Approximate number of maximum macro-particles per cell.
+         *
+         * Used internally for unit normalization.
+         */
+        constexpr uint32_t TYPICAL_PARTICLES_PER_CELL = startPosition::OnePositionParameter::numParticlesPerCell;
+
+    } // namespace particles
+} // namespace picongpu
--- a/share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
+++ b/share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
@@ -1,0 +1,388 @@
+"""
+This file is part of the PIConGPU.
+
+Copyright 2023 PIConGPU contributors
+Authors: Hannes Wolf
+License: GPLv3+
+
+Test script to determine if a pusher works correctly. This is done by the
+comparison of the radii of a particle at constant speed in a homogeneous
+magnetic field. If the relative change between two periods is greater than
+epsilon = 1e-5, the test fails. Also, the absolute phase change during one turn
+is regarded. It should be smaller than 0.2 rad.
+"""
+
+import openpmd_api as io
+import numpy as np
+import sys
+
+
+def get_params(path):
+    """Creates a series of the data found at path.
+    Reads the simulation parameters and saves them in the list params.
+
+    Parameters:
+    path (string): Path to the data to be compared against the Python
+                   reference.
+                   Path should have the form 'data_%T' to provide access to
+                   all iterations.
+
+    Returns:
+    series: Created series
+    order (int): Order of the assignment function
+    parameter_list: List containing parameters:
+        charge (float): Charge of the particle
+        cell_depth/width/height (float): Cell dimensions (respectively)
+        dt (float): Duration of a timestep
+        unit_time (float): Conversion factor of arbitrary PIConGPU time units
+                           to SI
+        unit_charge (float): Conversion factor of arbitrary PIConGPU charge
+                             units to SI
+        unit_length (float): Conversion factor of arbitrary PIConGPU length
+                             units to SI
+
+    """
+    # load series
+    series = io.Series(path, io.Access.read_only)
+    indices = np.array(series.iterations)
+    i = series.iterations[indices[0]]
+
+    if len(series.iterations) == 1:
+        raise ValueError("There is just 1 iteration in the series" "make sure, there are at least two")
+
+    elif len(np.array(i.particles)) != 1:
+        raise ValueError("There is not only 1 particle in the series" "make sure, there is only one")
+
+    else:
+        # read parameters of the simulation
+        params = {
+            "cell_depth": i.get_attribute("cell_depth"),
+            "cell_height": i.get_attribute("cell_height"),
+            "cell_width": i.get_attribute("cell_width"),
+            "dt": i.get_attribute("dt"),
+            "unit_time": i.get_attribute("unit_time"),
+            "unit_length": i.get_attribute("unit_length"),
+            "unit_charge": i.get_attribute("unit_charge"),
+            "unit_bfield": i.get_attribute("unit_bfield"),
+            "unit_speed": i.get_attribute("unit_speed"),
+            "unit_mass": i.get_attribute("unit_mass"),
+        }
+
+        return series, params
+
+
+def read_series(series):
+    """Reads the position and momentum of the particle in the x-y-plane
+    provided by the simulation as well as the charge and mass of the particle.
+
+    Parameters:
+    series: Simulation data
+
+    Returns:
+    x/y_poss (np.array): Relative in-Cell positions of the particle in the
+                         series
+    x/y_offSet (np.array): All knot-positions of the particle in the series
+    period (int): openPMD period of the simulation
+    charge (float): Charge of particle
+    mass (float): Mass of particle
+    x/y_momentum (float): All momenta of the particle in the series
+    R_c (float): Radius of gyration of the particle at timestep 0
+
+    """
+    num_iter = len(series.iterations)
+
+    # create arrays for saving the data
+
+    # in-cell position
+    x_poss = np.zeros(num_iter)
+    y_poss = np.zeros(num_iter)
+
+    # cells
+    x_offSet = np.zeros(num_iter)
+    y_offSet = np.zeros(num_iter)
+
+    # momentum
+    x_momentum = np.zeros(num_iter)
+    y_momentum = np.zeros(num_iter)
+
+    i = series.iterations[0]
+    # charge
+    charge = i.particles["e"]["charge"][io.Mesh_Record_Component.SCALAR]
+    # mass
+    mass = i.particles["e"]["mass"][io.Mesh_Record_Component.SCALAR]
+
+    # calculating of the openPMD-period for later use
+    indices = np.array(series.iterations)
+    period = indices[1] - indices[0]
+
+    for iteration in indices:
+        i = series.iterations[iteration]  # current position
+
+        # node coordinates
+        pos2_x = i.particles["e"]["position"]["x"].load_chunk()
+        pos2_y = i.particles["e"]["position"]["y"].load_chunk()
+
+        # InCell coordinates
+        pos_off2_x = i.particles["e"]["positionOffset"]["x"].load_chunk()
+        pos_off2_y = i.particles["e"]["positionOffset"]["y"].load_chunk()
+
+        p_x = i.particles["e"]["momentum"]["x"].load_chunk()
+        p_y = i.particles["e"]["momentum"]["y"].load_chunk()
+
+        series.flush()
+
+        array_index = int(iteration / period)
+        # save the read data in arrays for later usage
+        x_poss[array_index] = pos2_x
+        y_poss[array_index] = pos2_y
+
+        x_offSet[array_index] = pos_off2_x
+        y_offSet[array_index] = pos_off2_y
+
+        x_momentum[array_index] = p_x
+        y_momentum[array_index] = p_y
+
+    return (x_poss, y_poss, x_offSet, y_offSet, x_momentum, y_momentum, charge, mass, num_iter, period)
+
+
+def compare_radius(
+    x_poss, y_poss, x_offSet, y_offSet, x_momentum, y_momentum, charge, mass, B, params, num_iter, period
+):
+    """Tests if the change in radius from one revolution to the other is
+    greater than epsilon = 1e-5.
+    The radii are calculated by the positions of the particle in the x-y-plane
+    in one test and by the momenta (and q*B) in the x-y-plane in the other one.
+
+    Parameters:
+    x/y_momentum(array): Series of the momenta of the particle in x/y-direction
+    charge (float): Charge of the particle
+    mass( float): Mass of the particle
+    B (float): Magnitude of the magnetic field B in the test
+    params (dict): Simulation parameters
+    num_iter (int): Number of iterations in the series
+    period (int): openPMD period of the simulation
+
+    Returns:
+    compare_result (int):
+        == 0 if the radius of gyration does not vary much (<epsilon)
+        != 0 if the change in radius of gyration large (>epsilon)
+
+    """
+    # the  change in of the radius, measured between 2 turns
+    # is ca. 2e-5 (calculated with position)
+    # Doubling this value yields epsilon as an approximation of the maximal
+    # uncertanty PIConGPU should have when the test uses the position.
+    # Therefore the test uses this value to test the radius change against.
+    # PIConGPU run September 2023
+    epsilon_position = 5e-5
+
+    # calculating gamma for the calculation of the periodic time
+    abs_momentum = np.sqrt(x_momentum[0] ** 2 + y_momentum[0] ** 2)
+    gamma = 1 / np.sqrt(1 - abs_momentum**2 / (abs_momentum**2 + 1))
+    # periodic time / timestep
+    steps_per_revolution = (2 * np.pi * gamma * mass[0] / (abs(charge[0]) * B / params["unit_bfield"])) / params["dt"]
+
+    revolutions_in_series = int(num_iter / steps_per_revolution)
+
+    compare_result_momentum = 0
+    compare_result_positions = 0
+
+    # compare with position
+    radius = np.sqrt((x_poss + x_offSet) ** 2 + (y_poss + y_offSet) ** 2)
+
+    # relative comparison
+    for k in range(revolutions_in_series):
+        index_of_revolution = int(steps_per_revolution / period) * k
+        compare_radii = abs(radius[index_of_revolution] - radius[index_of_revolution + 1]) / radius[index_of_revolution]
+
+        if compare_radii > epsilon_position:
+            print("pusher is not valid (position)")
+            print("please check the simulation")
+            compare_result_positions += 42
+
+        else:
+            print("pusher is valid (position)")
+
+    # the greatest change in of the radius, measured between 2 turns
+    # (2000 Turns simulated) is ca. 4e-6 (calculated with momentum)
+    # Doubling this value yields epsilon as an approximation of the maximal
+    # uncertanty PIConGPU should have when the test uses the momentum.
+    # Therefore the test uses this value to test the radius change against.
+    # PIConGPU run September 2023
+    epsilon_momentum = 1e-5
+
+    # comparision with momentum
+    radius = (
+        np.sqrt(x_momentum**2 + y_momentum**2)
+        * params["unit_mass"]
+        * params["unit_speed"]
+        / (abs(charge[0]) * params["unit_charge"] * B)
+        / params["cell_depth"]
+        / params["unit_length"]
+    )
+    # relative comparison
+    for k in range(revolutions_in_series):
+        index_of_revolution = int(steps_per_revolution / period) * k
+        compare_radii = abs(radius[index_of_revolution] - radius[index_of_revolution + 1]) / radius[index_of_revolution]
+
+        if compare_radii > epsilon_momentum:
+            print("pusher is not valid (position)")
+            print("please check the simulation")
+            compare_result_momentum += 46
+
+        else:
+            print("pusher is valid (momentum)")
+
+    return compare_result_momentum + compare_result_positions
+
+
+def compare_phases(
+    x_poss, y_poss, x_offSet, y_offSet, x_momentum, y_momentum, charge, mass, B, params, num_iter, period
+):
+    """Tests if the phase difference from one revolution to the other is
+    greater than delta = 0.25.
+    The radii are calculated by the momenta (and q*B) in the x-y-plane.
+
+    Parameters:
+    x/y_momentum(array): Series of the momenta of the particle in x/y-direction
+    charge (float): Charge of the particle
+    mass (float): Mass of the particle
+    B (float): Magnitude of the magnetic field B in the test
+    params (dict): Simulation parameters
+    num_iter (int): Number of iterations in the series
+    period (int): openPMD period of the simulation
+
+    Returns:
+    0: if the phase difference is small (<delta)
+    1: if the phase difference large (>delta)
+
+    """
+    # The measured phase difference between two revolutions is approximately
+    # 0.08rad. Doubling this value represents the maximal error to be accepted
+    # for the phase difference in the PIConGPU.
+    # PIConGPU run of September 2023.
+    delta = 0.16
+
+    # calculating gamma for the calculation of the periodic time
+    # (gamma factor for motion in the x-y plane)
+    abs_momentum = np.sqrt(x_momentum[0] ** 2 + y_momentum[0] ** 2)
+    gamma = 1 / np.sqrt(1 - abs_momentum**2 / (abs_momentum**2 + 1))
+    # periodic time / timestep
+    steps_per_revolution = (2 * np.pi * gamma * mass[0] / (abs(charge[0]) * B / params["unit_bfield"])) / params["dt"]
+
+    revolutions_in_series = int(num_iter / steps_per_revolution)
+
+    x = x_poss + x_offSet  # real x coordinates
+    R_c = np.mean(
+        (
+            np.sqrt(x_momentum**2 + y_momentum**2)
+            * params["unit_mass"]
+            * params["unit_speed"]
+            / (abs(charge[0]) * params["unit_charge"] * B)
+            / params["cell_depth"]
+            / params["unit_length"]
+        )
+    )
+
+    theta = np.arccos(x / R_c)  # calculate the phase
+
+    for k in range(revolutions_in_series):
+        index_of_revolution = int(steps_per_revolution / period) * k
+        # compare the phase after 1 periodic time
+        compare_phases = abs(
+            theta[index_of_revolution] - theta[index_of_revolution + int(steps_per_revolution / period)]
+        )
+        # compare with theoretical, i.e. stationary phase
+        compare_ana_phases = abs(theta[index_of_revolution] - np.pi / 2 + 2 * np.pi / 100)
+
+        compare_result_phases = 0
+
+        if compare_phases > delta:
+            print("pusher is not valid (phases)")
+            print("please check the simulation")
+            compare_result_phases += 1
+
+        if compare_ana_phases > (k + 1) * delta:
+            print("pusher is not valid (phases_analytical)")
+            print("please check the simulation")
+            compare_result_phases += 2
+
+        else:
+            print("pusher is valid (phases)")
+    return compare_result_phases
+
+
+def correct_starting_values_for_technical_details(x_poss, y_poss, x_offSet, y_offSet, R_c):
+    """The particle is initialized at the arbitary position (5, 32) in the x-y-plane (some arbitrary z which is irrelevant for our computation...).
+    So we have to shift the coordinates in a form, that the initialization point is exactly on the circle with the radius of the particle trajectory with its center at the origin.
+    This shifting is done by the transformation (subtraction of the initialization coordinates and radius).
+    But we also need to compensate for a half-step, which is automatically done by PIConGPU to improve the accuracy of the Boris Pusher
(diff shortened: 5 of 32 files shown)


share/picongpu/tests/Pusher/README.rst
```diff
@@ -0,0 +1,17 @@
...
+The electron should, therefore, move in a perfect circle since there is no acceleration in the direction of movement.
+The test compares the radii of a particle simulated by PIConGPU. If the relative change between two periods is greater than 
+epsilon = 1e-5, the test fails. Also, the absolute phase change during one turn is regarded. It should be smaller than delta = 0.25 rad.
+
+The thresholds epsilon and delta were determined as follows:
+
+epsilon: The greatest change in the radius measured between 2 turns (2000 Turns simulated) is approximately 4e-6. This doubled yields epsilon as an approximation 
+-------- of the maximal uncertainty PIConGPU should have. Therefore, the test uses this value to check the radius change.
+
+delta: The measured phase difference between 2 revolutions is approximately 0.125rad. This doubled is the maximal error to be accepted for the phase difference between 2 turns.
```
chillenzer: 0.125 rad doesn't seem especially small. Is it something to worry about?

share/picongpu/tests/Pusher/bin/ci.sh
```diff
@@ -0,0 +1,121 @@
...
+  echo ""
+  echo "Usage: ci.sh"
+  echo ""
+  echo "Options"
+  echo "-t | --flag                   - flag number of cmakeFlags"
+  echo "-h | --help                   - show help"
+  echo ""
+}
+
+## not used at the moment
```
chillenzer: If so, remove.

share/picongpu/tests/Pusher/bin/ci.sh
```diff
@@ -0,0 +1,121 @@
...
+
+help()
+{
+  echo "Execute a single particle simulation in a homogenious magnetic field to validate the pusher."
+  echo "Simulation results are validated against by checking the change of radius during the simulation and the quantitative phase shift"
+  echo ""
+  echo "Usage: ci.sh"
+  echo ""
+  echo "Options"
+  echo "-t | --flag                   - flag number of cmakeFlags"
```
chillenzer: Alphabetical order

share/picongpu/tests/Pusher/bin/ci.sh
```diff
@@ -0,0 +1,121 @@
...
+  simPath="./simOutput_$date_stamp"
+
+  if [ -d "$simPath" ] ; then
+      echo "Destination path already in use, cannot create new folder" >&2
+      exit 1
+  fi
+
+  mkdir -p $simPath
+
+  # use absolut path's
```
chillenzer: Remove `'s`

share/picongpu/tests/Pusher/bin/ci.sh
```diff
@@ -0,0 +1,121 @@
...
+  simPath=$(absolute_path $simPath)
+
+  cd $simPath
+  echo "Run!"
+  mpiexec -n 1 ../bin/picongpu -d 1 1 1 -g 64 64 32 -s 100 --fieldBackground.duplicateFields --periodic 1 1 1 --openPMD.period 1 --openPMD.file simData --openPMD.ext bp
+  cd ..
+
+  dataPath=$(absolute_path "$simPath/openPMD")
+else
```
chillenzer, line 95: I would move this branch up, so that the control flow and the actual logic are less interleaved. Concretely
```
if [ $ret_build -ne 0 ]; then
    echo "could not build simulation"
    exit 2
fi

< proceed as before >

File: share/picongpu/tests/Pusher/etc/picongpu/1.cfg [hunk 6]
```diff
@@ -0,0 +1,76 @@
...
+TBG_steps="100"
+
+TBG_periodic="--periodic 1 1 1"
+
+
+#################################
+## Section: Optional Variables ##
+#################################
+
+# write position to stdout (messy):
```
chillenzer: What's this?

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,387 @@
...
+License: GPLv3+
+
+Test script to determine if a pusher works correctly. This is done by the
+comparison of the radii of a particle at constant speed in a homogeneous
+magnetic field. If the relative change between two periods is greater than
+epsilon = 1e-5, the test fails. Also, the absolute phase change during one turn
+is regarded. It should be smaller than 0.2 rad.
+"""
+
+import openpmd_api as io
```
chillenzer, line 15: I see that this seems to be common in the repository but it's not ideal because `io` is [a python standard library module](https://docs.python.org/3/library/io.html) which might mislead people more familiar with the Python ecosystem than with PIConGPUs conventions. I've created #4802. You don't need to act on this unless my proposed changes get merged before your PR.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,387 @@
...
+    series = io.Series(path, io.Access.read_only)
+    indices = np.array(series.iterations)
+    i = series.iterations[indices[0]]
+
+    if len(series.iterations) == 1:
+        raise ValueError("There is just 1 iteration in the series"
+                         "make sure, there are at least two")
+
+    elif len(np.array(i.particles)) != 1:
+        raise ValueError("There is just 1 iteration in the series"
```
chillenzer: Copy-pasted wrong error message.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,387 @@
...
+        unit_charge (float): Conversion factor of arbitrary PIConGPU charge
+                             units to SI
+        unit_length (float): Conversion factor of arbitrary PIConGPU length
+                             units to SI
+
+    """
+    # load series
+    series = io.Series(path, io.Access.read_only)
+    indices = np.array(series.iterations)
+    i = series.iterations[indices[0]]
```
chillenzer, line 48: Choose a more descriptive name for `i`. Here and in the following.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,387 @@
...
+        cell_width = i.get_attribute("cell_width")
+        dt = i.get_attribute("dt")
+        unit_time = i.get_attribute("unit_time")
+        unit_length = i.get_attribute("unit_length")
+        unit_charge = i.get_attribute("unit_charge")
+        unit_bfield = i.get_attribute("unit_bfield")
+        unit_speed = i.get_attribute("unit_speed")
+        unit_mass = i.get_attribute("unit_mass")
+
+        return series, [
```
chillenzer: Use a self-describing data structure for the parameters instead of a list like, e.g., a `dict` or a `namedtuple` or the like. Having stuff like `params[7]` later on is absolutely cryptic.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,387 @@
...
+    """
+    # read parameters of series
+    series, params = get_params(dataPath)
+
+    B = 50
+    # extract all relevant data
+    (x_poss, y_poss, x_offSet, y_offSet, x_momentum, y_momentum, charge,
+     mass, num_iter, periode) = read_series(series)
+
+    # all radii in the series
```
chillenzer, line 355: It is likely my ignorance but it seems like `sqrt(x_momentum**2 + y_momentum**2)...` is the length of the momentum vector and not a radius.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,387 @@
...
+    radius = (np.sqrt(x_momentum**2 + y_momentum**2) * params[7] * params[8] /
+              (abs(charge[0])*params[5]*B) / params[0] / params[6])
+    R_c = radius[0]  # original radius
+
+    # transformation for the correct calculation of the radii by position
```
chillenzer, line 366: Just for my information: Could you elaborate on why these following 4 statements are necessary please?

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,387 @@
...
+    x/y_momentum(array): Series of the momenta of the particle in x/y-direction
+    charge (float): Charge of the particle
+    mass (float): Mass of the particle
+    B (float): Magnitude of the magnetic field B in the test
+    params (list): Simulation parameters
+    num_iter (int): Number of iterations in the series
+    period (int): openPMD period of the simulation
+
+    Returns:
+    0: if the radius of gyration does not vary much (<epsilon)
```
chillenzer: Might be personal style but I would prefer to return `bool` here instead of magic numbers. This would read particularly well with a function name like `is_momentum_deviation_acceptable -> True/False`.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,387 @@
...
+        y_offSet[array_index] = pos_off2_y
+
+        x_momentum[array_index] = p_x
+        y_momentum[array_index] = p_y
+
+    return (x_poss, y_poss, x_offSet, y_offSet, x_momentum, y_momentum,
+            charge, mass, num_iter, periode)
+
+
+def compare_momentum(x_momentum, y_momentum,
```
chillenzer: The `compare_momentum` and `compare_positions` functions are exactly duplicated (up to the normalisation of the vector lengths). Merge them, e.g., taking the length as single argument instead of doing the computation inside of the function.

chillenzer: PS: There are a few more issues with these functions (like copy-paste oversights etc.) that I have not pointed out because they are likely to be revised. If not, we should come back to this.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,387 @@
...
+                                                 x_momentum, y_momentum,
+                                                 charge, mass, B, params,
+                                                 num_iter, periode)
+    compare_result_phases = compare_phases(x_poss, y_poss, x_offSet, y_offSet,
+                                           x_momentum, y_momentum,
+                                           charge, mass, B, params,
+                                           num_iter, periode)
+
+    # yield both tests/comparisions a positive result?
+    compare_result = (compare_result_momentum + compare_result_positions +
```
chillenzer: In conjuction with the above suggestions about function naming and returning bool, the following few lines should read like
```python3
return (
    has_acceptable_deviation_in(momentum_radius)
    and has_acceptable_deviation_in(position_radius)
    and has_acceptable_phase_deviation(phase)
)
```
(It looked like the phase comparison could be sufficiently different to justify being an own function.)

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,387 @@
...
+                      compare_result_phases)
+
+    return compare_result
+
+
+if __name__ == "__main__":
+    try:
```
chillenzer: This looks very complicated. Why not?
```python3
if len(sys.argv) != 2:
    raise ...
```

share/picongpu/tests/Pusher/lib/python/test/Pusher/requirements.txt
```diff
@@ -0,0 +1,2 @@
+numpy
```
chillenzer: Should we specify versions here? Would be annoying to have a breaking test because of some unnoticed version bump that's not backwards compatible.

chillenzer (changes requested):
Hi, thanks for your contribution. Sounds like solid, physically reasonable tests. Good and clear explanations in the `README`s. Well done!

Don't let the number of comments scare you, I've also commented questions just for my information. After all, the review is as much for my information as for yours.

I'm not yet in the position to give a deep inside perspective but I've got a few thoughts about the colour of the [bikeshed](https://en.wikipedia.org/wiki/Law_of_triviality), if I may. As the two tests are very similar in their general structure, I only commented on the first yet. Please apply to both tests as appropriate.

MRZirper:
> Hi, thanks for your contribution. Sounds like solid, physically reasonable tests. Good and clear explanations in the `README`s. Well done!
> 
> Don't let the number of comments scare you, I've also commented questions just for my information. After all, the review is as much for my information as for yours.
> 
> I'm not yet in the position to give a deep inside perspective but I've got a few thoughts about the colour of the [bikeshed](https://en.wikipedia.org/wiki/Law_of_triviality), if I may. As the two tests are very similar in their general structure, I only commented on the first yet. Please apply to both tests as appropriate.

I am bussy with some exams right now. This is why I won't work on this until march. I appreciate all of your comments and I am eager to discuss them with you in march. until then!

chillenzer:
Great! Looking forward to that! =)

chillenzer:
PS: Good luck with the exams!

share/picongpu/tests/Pusher/README.rst
MRZirper: Yeah, I messed some things up there. The theoretical phase difference between a timestep would be around 0.125 rad, i.e. 2pi/50 for this test case. The deviation of the phase difference is tested here. I did a test run and evaluated the results for this derivation. It was mostly about 1e-3 rad, but there were a few peaks with ca. 0.017rad, i.e. around 14%.
I propose to accept deviations of 20% which refers to 0.025 rad (at the end I misplaced the decimal point). In the end, I messed up a few things here. I am sorry and I will correct them.

MRZirper: I messed some things up here. To be clear: 0.125 rad, i.e. 2*pi/50 is the value of the phaseshift after one timestep (no numerical effects considered). Due to an index error, I measured and used this value for the comparison. 
I corrected my mistake and measured again. So the approximate value of the phase shift due to numerical effects (caused by the Boris pusher in this case) is 0.083 rad. Going through the same argumentation as in the `README` we accept only derivations smaller than 0.16rad per revolution of the particle in the magnetic field.

I corrected It in the Code and the `README`. Sorry that I  have not questioned this large value.

share/picongpu/tests/Pusher/bin/ci.sh
MRZirper: done

share/picongpu/tests/Pusher/bin/ci.sh
MRZirper: done

share/picongpu/tests/Pusher/bin/ci.sh
MRZirper: done

share/picongpu/tests/Pusher/bin/ci.sh
MRZirper: I don't know what you mean... We cannot check whether the build worked or not if we did not do the build before. For me, everything is in the logic order.

share/picongpu/tests/Pusher/etc/picongpu/1.cfg
MRZirper: I don't know. I just copied an example to proceed with the test from there and adjusted the thinks I needed. I am not an expert of the PIConGPU file system. Should I remove it?

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
MRZirper: There is still no change in the documentation. So do I stick with the `io` until this happens or should I simply adapt the `opmd` as you discussed in the issue?

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
MRZirper: Thanks. I already corrected it in the `PusherScaling` test.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
MRZirper: The same thing @steindev mentioned in [#4802](https://github.com/ComputationalRadiationPhysics/picongpu/issues/4802#issuecomment-1921494128): it was introduced like this in the [openPMD documentation ](https://openpmd-api.readthedocs.io/en/latest/usage/firstread.html). Should I change the name despite that fact or should I just stick with that convention?

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
MRZirper: Thanks for the advise. I learned something new :)

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
MRZirper: The particle moves with constant speed in the homogeneous magnetic field. Therefore we can calculate the radius out of the absolute value of the particle velocity/momentum. The conversion of the length of the momentum vector to the radius is via the factors in the `...`

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
MRZirper: Sure! The particle is initialized at the position (5, 32) in the x-y-plane (some arbitrary z which is irrelevant for our computation...). So we have to shift the coordinates in a form, that the initialization point is exactly on the circle with the radius of the particle trajectory with its center at the origin. This shifting is done by the transformation (subtraction of the initialization coordinates and radius).
But we also need to compensate for a half-step, which is automatically done by PIConGPU to improve the accuracy of the Boris Pusher (see: B. Ripperda et al 2018 ApJS 235 21 https://iopscience.iop.org/article/10.3847/1538-4365/aab114 ; 10.3847/1538-4365/aab114). This half-step is the first data written by PIConGPU and is corrected by the trigonometry stuff in the transformation. I hope this helps.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
MRZirper: As I understand it: The CI works with 0 as true and every other integer as false. Therefore I use these return values. Also, we can distinguish the different Errors by their non-zero number. (I used different numbers down in the code)

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
MRZirper: I merged them

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
MRZirper: This should be relevant depending on how we decide the case above.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
MRZirper: This is right. I implemented it.

share/picongpu/tests/Pusher/lib/python/test/Pusher/requirements.txt
MRZirper: I am using version 1.24.2. I add it to the `requirments.txt`

MRZirper:
I tested the test again with the old data and noticed that I got a relative change of the radius per revolution of about 2e-5 so 10 times as much as noted before. I don't know where this is coming from... might be that I messed up things a bit. I would test tomorrow on hemera and then decide to adjust/higher epsilon.

MRZirper:
Running the test again yielded the same result. I think I calculated the maximal change in radius with the momentum back then yielding a more accurate result than the hole transformation thing and the calculating the radius with the position of the particle...
So I "fixed" the epsilon value and used 2 different ones for both kinds of calculating the radius.

share/picongpu/tests/Pusher/README.rst
```diff
@@ -0,0 +1,17 @@
...
+
+An electron with a velocity of 0.5c moves perpendicular to a homogeneous magnetic field (in the z-direction).
+The electron should, therefore, move in a perfect circle since there is no acceleration in the direction of movement.
+The test compares the radii of a particle simulated by PIConGPU. If the relative change between two periods is greater than 
+epsilon = 1e-5, the test fails. Also, the absolute phase change during one turn is regarded. It should be smaller than delta = 0.25 rad.
+
+The thresholds epsilon and delta were determined as follows:
+
+epsilon: The greatest change in the radius measured between 2 turns (2000 Turns simulated) is approximately 4e-6. This doubled yields epsilon as an approximation 
+-------- of the maximal uncertainty PIConGPU should have. Therefore, the test uses this value to check the radius change.
```
chillenzer: Why are there some random hyphens in here?

share/picongpu/tests/Pusher/README.rst
```diff
@@ -0,0 +1,17 @@
...
+
+delta: The measured phase difference between 2 revolutions is approximately 0.08rad. This doubled is the maximal error to be accepted for the phase difference between 2 turns. 
+------
```
chillenzer: ... and here?

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,364 @@
...
+        != 0 if the change in radius of gyration large (>epsilon)
+
+    """
+    # the  change in of the radius, measured between 2 turns
+    # is ca. 2e-5 (calculated with position)
+    # Doubling this value yields epsilon as an approximation of the maximal
+    # uncertanty PIConGPU should have when the test uses the position.
+    # Therefore the test uses this value to test the radius change against.
+    # PIConGPU run September 2023
+    epsilon_position = 5e-5
```
chillenzer, line 177: Is this the same epsilon that is mentioned in the README? If so, this has a different numerical value. I would propose to use `.rst`'s `..literalinclude::` functionality to avoid such mistakes.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,364 @@
...
+        else:
+            print("pusher is valid (position)")
+
+    # the greatest change in of the radius, measured between 2 turns
+    # (2000 Turns simulated) is ca. 4e-6 (calculated with momentum)
+    # Doubling this value yields epsilon as an approximation of the maximal
+    # uncertanty PIConGPU should have when the test uses the momentum.
+    # Therefore the test uses this value to test the radius change against.
+    # PIConGPU run September 2023
+    epsilon_momentum = 1e-5
```
chillenzer, line 212: Okay. Now I'm very confused. So, there are two epsilons with different numerical values and one of them is not mentioned at all in the README?

share/picongpu/tests/Pusher/bin/ci.sh
chillenzer: I meant to put the above code suggestion at line 74 instead of having everything inside the `if` block but it's not a strong opinion here.

share/picongpu/tests/Pusher/etc/picongpu/1.cfg
chillenzer: Yes, I'd say so. (Of course, this applies to both commented out lines here.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
chillenzer: Okay. OpenPMD clearly has a naming problem in the docs but that's not for us to fix right now. -.- I would still prefer a more descriptive name but you've got a good argument on your side, so keep it if you like.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
chillenzer: Thanks! That is definitely a non-trivial reason that I don't feel ashamed for not figuring out for myself. Should this be documented somehow? I could think of extracting this into a function with a nice docstring containing your explanation, e.g.
```python3
x_poss, y_poss, x_offSet, y_offSet = correct_starting_values_for_technical_details(R_c)
```

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,364 @@
...
+    # yield both tests/comparisions a positive result?
+    compare_result = compare_result_radius + compare_result_phases
+
+    return compare_result
+
+
+if __name__ == "__main__":
+    if len(sys.argv) != 2:
+        raise SystemExit(f"Usage: {sys.argv[0]} <path_to_simulation_data>")
+    main(sys.argv[1])
```
chillenzer: CAUTION: Currently the test can never fail!
```suggestion
    sys.exit(main(sys.argv[1]))
```

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,286 @@
...
+The standard deviation should be < 0.05.
+
+"""
+
+import openpmd_api as io
+import numpy as np
+import sys
+
+
+class compare:
```
chillenzer: Conventionally, class names should be PascalCase and nouns.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,286 @@
...
+        i = self.series.iterations[0]
+        # charge
+        self.charge = (i.particles["e"]["charge"]
+                       [io.Mesh_Record_Component.SCALAR])
+        # mass
+        self.mass = i.particles["e"]["mass"][io.Mesh_Record_Component.SCALAR]
+
+        # calculating the openPMD-periode for later use
+        indices = np.array(self.series.iterations)
+        periode = indices[1] - indices[0]  # openPMD-periode
```
chillenzer:
```suggestion
        period = indices[1] - indices[0]  # openPMD period
```

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,286 @@
...
+
+        for iteration in indices:
```
chillenzer, line 102: Doesn't something like this work as well:
```suggestion
        for i in self.series.iterations:
```

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,286 @@
...
+        y_momentum = np.zeros(num_iter)
+
+        i = self.series.iterations[0]
+        # charge
+        self.charge = (i.particles["e"]["charge"]
+                       [io.Mesh_Record_Component.SCALAR])
+        # mass
+        self.mass = i.particles["e"]["mass"][io.Mesh_Record_Component.SCALAR]
+
+        # calculating the openPMD-periode for later use
```
chillenzer:
```suggestion
        # calculating the openPMD period for later use
```

chillenzer (changes requested):
Hi, thanks for your explanations and updates. I understand much better now what's going on. There are still a few rough edges that I have pointed out below and one significant bug: This test can never fail! See below for details!

Also importantly: Don't forget to rebase against the latest `dev` branch. After doing so, you will find the CI failing. Please look at [this piece of documentation](https://picongpu.readthedocs.io/en/latest/dev/docs/COMMIT.html) to get `pre-commit` set up on your machine, run `pre-commit run --all-files --hook-stage manual` and commit the suggested changes.

share/picongpu/tests/Pusher/README.rst
MRZirper: They are underlining epsilon. You see this as long as your window is big enough. I will remove them for the case of a small window size like here.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
MRZirper: As I mentioned above I had to adjust the test because the calculation of the raduis with the position of the particle is not as precice as the calculation with the momentum. Therefore we have to regard two different thresholds for the test to call it a failure. I adjusted the readme.

share/picongpu/tests/Pusher/bin/ci.sh
MRZirper: Ah ok.. I think it is more clear to assign the data path after checking if the simulation was successfully built. So I would leave it like this.

share/picongpu/tests/Pusher/etc/picongpu/1.cfg
MRZirper: done

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
MRZirper: I used 
`else:
    main(sys.argv[1])`

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
MRZirper: Comparison

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
MRZirper: done everywhere in the file

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
MRZirper: Sadly this is not possible. For this case `i` would be an integer not an iteration...

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
chillenzer: No, that's not what we're fixing here: Your bash script roughly runs
```bash
python3 ./MainTest.py
exit "$?"
```
but `$?` always holds `0` (i.e. `EXIT_SUCCESS`) because no matter what the return value of `main()` is the _`python3` executable ran just fine_. You have to explicitly instruct Python in your script to exit with a non-zero exit code, hence
```python3
 sys.exit(main(sys.argv[1]))
```

MRZirper:
@chillenzer

share/picongpu/tests/Pusher/README.rst
chillenzer: Still, doesn't sound super tiny but that's not your fault. Resolved.

share/picongpu/tests/Pusher/bin/ci.sh
chillenzer: It's fine.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
chillenzer: It's fine. We'll handle that separately.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
chillenzer: Okay. I see. Maybe we can still find a better name that makes sure the reader can clearly distinguish this form the directly measured radius? On the spot, I came up with
* `charateristic_radius`
* `momentum_derived_radius`
* `momentum_equivalent_radius`

but none of them sounds particularly great. It's fine if you have as little success with this as me, but please give it a try.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
chillenzer: Resolved.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
chillenzer: Resolved.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
chillenzer: This is still open.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+
+Copyright 2023 PIConGPU contributors
+Authors: Hannes Wolf
+License: GPLv3+
+
+Test script to determine if a Boris pusher works correctly. This is done by
+calculating the phase shift caused by the pusher.
+
+The phase shift is proportional to (dt)**2. The script checks whether the
+calculated exponent of (dt)**x is in the interval [1.9, 2.1] or not.
```
chillenzer, line 12: Such numbers are likely to become outdated quickly. Maybe rather refer to the corresponding variable to look up the concrete value.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+The standard deviation should be < 0.05.
```
chillenzer: Same here.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+        """Creates a series of the data found at path.
+        Reads the simulation parameters and saves them in the list params.
+
+        Parameters:
+        path (string): Path to the data to be compared against the Python
+                       reference.
+                       Path should have the form 'data_%T' to provide access to
+                       all iterations.
+
+        Returns:
```
chillenzer: This method does not return anything but just stores the corresponding values. Also, there `parameter_list` you refer to has another name and is a dictionary!

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+        series: Created series
+        order (int): Order of the assignment function
```
chillenzer: This is never mentioned in the function body.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+class Comparison:
+    def __init__(self, path, B):
+        self.path = path
+        self.B = B
+
+    def get_params(self):
+        """Creates a series of the data found at path.
+        Reads the simulation parameters and saves them in the list params.
+
+        Parameters:
```
chillenzer, line 26: This function does not take parameters.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+                                 units to SI
+
+        """
+        # load series
+        series = io.Series(self.path, io.Access.read_only)
+        indices = np.array(series.iterations)
+        i = series.iterations[indices[0]]
+
+        if len(series.iterations) == 1:
+            raise ValueError("There is just 1 iteration in the series" "make sure, there are at least two")
```
chillenzer: Merge the strings.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+
+        elif len(np.array(i.particles)) != 1:
+            raise ValueError("There is not only 1 particle in the series" "make sure, there is only least one")
```
chillenzer: Merge the strings.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+            }
+            self.series = series
+            self.params = params
+
+    def read_series(self):
+        """Reads the position and momentum of the particle in the x-y-plane
+        provided by the simulation, as well as the charge and mass of the
+        particle.
+
+        Parameters:
```
chillenzer: Same here.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+        series: simulation data
+
+        Returns: (self.variable)
```
chillenzer: Same here.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
chillenzer: Okay.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+            / self.params["unit_length"]
+        )
+
+        R_c = radius[0]  # original radius
+
+        # half_step
+        steps_per_turn = 0.825e-12 / (self.params["dt"] * self.params["unit_time"])
+        halfStepPhase = np.pi / steps_per_turn
+
+        # coordinate transformation to reposition the origin at the center of
```
chillenzer, line 145: Whenever you've got a comment explaining what the next few lines do, extract those lines into a function and give them a name and a docstring.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+        # the circle
+        x_poss = x_poss + (R_c * np.sin(halfStepPhase))
+        y_poss = y_poss - R_c + int(R_c) + (R_c - R_c * np.cos(halfStepPhase))
+
+        x_offSet = x_offSet - 32
```
chillenzer: This magic number needs a name.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+        y_offSet = y_offSet - 5 - int(R_c)
```
chillenzer: Same here.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+        phase_class.get_params()
+
+        # extract all relevant data
+        phase_class.read_series()
+
+        phaseDifferences[j - 1], timesteps[j - 1] = phase_class.get_total_phasediff()
+
+    quotients = np.zeros(len(phaseDifferences) - 1)
+    for j in range(len(phaseDifferences) - 1):
+        quotients[j] = phaseDifferences[j + 1] / phaseDifferences[j]
```
chillenzer:
```suggestion
        quotients = phaseDifferences[1:] / phaseDifferences[:-1]
```
to replace these 3 lines.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+
+    b = np.mean(quotients)  # base b
+    expo = np.log(b) / np.log(2)  # exponent x of b = 2**x
+    sigma_b = (max(quotients) - min(quotients)) / 2  # std of b
```
chillenzer: That is not a standard deviation. Please replace all references to "standard deviation" with something factually correct (maybe just `spread` or `width` or something in that direction).

chillenzer: Alternatively: Why don't you just use `np.std`?

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+    sigma_x = 1 / (b * np.log(2)) * sigma_b
```
chillenzer, line 245: How does this formula come about and how is this related to the standard deviation of `x`?

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
```diff
@@ -0,0 +1,285 @@
...
+        y = self.y_poss + self.y_offSet  # y coordinates
+
+        res_cos = x / np.sqrt(x**2 + y**2)
+        theta = np.arccos(res_cos)
+
+        timestep = self.params["dt"] * self.params["unit_time"]
+
+        total_phasediff = 0
+        for counter in range(0, len(theta) - 1, 1):
+            total_phasediff += abs(theta[counter + 1] - theta[counter])
```
chillenzer:
```suggestion
            total_phasediff = np.abs(np.diff(theta))
```
to replace these three lines.

chillenzer (changes requested):
Thanks for the ping. Just went through this another time focusing on the second test that I didn't check before. In general, quite clear and readable code. Well done!

Still, there are a few comments below. Please again apply to both tests as appropriate.

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
BrianMarre: in this case use an [enumeration](https://docs.python.org/3/library/enum.html), this will allow you to write meaningful names in the code instead of just integers and for the CI yo can covnert them to an integer if required.

share/picongpu/tests/Pusher/README.rst
```diff
@@ -0,0 +1,23 @@
...
+It should be smaller than delta = 0.25 rad.
+
+The thresholds epsilon and delta were determined as follows:
+
+epsilon: The greatest change in the radius measured between 2 turns (2000 Turns simulated) is approximately 4e-6 for the calculation with the momentum.
+         For the calculation with the position of the particle, this value is approximately 2.5e-5.
+         Those values doubled yield the two epsilons as an approximation 
+         of the maximal uncertainty PIConGPU should have. Therefore, the test uses this value to check the radius change.
+
+delta: The measured phase difference between 2 revolutions is approximately 0.08rad. This doubled is the maximal error to be accepted for the phase difference between 2 turns.
```
BrianMarre: you first state in line 12 that $\Delta \varphi \leq 0.25\ \mathrm{rad}$, later in 21 $\Delta \varphi \leq 2 \cdot  0.08\ \mathrm{rad}$, choose one please.

share/picongpu/tests/Pusher/README.rst
```diff
@@ -0,0 +1,23 @@
...
+The electron should, therefore, move in a perfect circle since there is no acceleration in the direction of movement.
+The test compares the radii of a particle simulated by PIConGPU. 
+The radius is calculated by the position of the particle and its momentum (r =  p / (q * B) ).
+If the relative change between two periods is greater than epsilon_position = 5e-5 or epsilon_momentum = 1e-5 respectively, the test fails.
+The calculation of the radius with the position of the particle has a higher error than the calculation with the momentum.
+Therefore the test compares the change of the radius against two different epsilons.
+Also, the absolute phase change during one turn is regarded. 
+It should be smaller than delta = 0.25 rad.
+
+The thresholds epsilon and delta were determined as follows:
```
BrianMarre: change to
```
the acceptance thresholds for position-, momentum- and phase-error were determined as follows:
```

share/picongpu/tests/Pusher/include/picongpu/param/particle.param
```diff
@@ -0,0 +1,112 @@
...
+                 * Determines the weighting of a macro particle as well as the number of
+                 * macro-particles which sample the evolution of the particle distribution
+                 * function in phase space.
+                 *
+                 * unit: none
+                 */
+                static constexpr uint32_t numParticlesPerCell = 1u;
+
+                const InCellOffset_t inCellOffset;
+            };
```
BrianMarre: the use of CONST_VECTOR should be removed here,

please specify the in cell position as a constexpr float3_X vector as introduced introduced in PR #4750

see [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/d18b6822e3074e55b3eab6d7c4c9024564efea33/include/picongpu/param/particle.param#L59-L66) for an example of how to do that, compared to the [old state](https://github.com/ComputationalRadiationPhysics/picongpu/blame/0e1ca25ddd607cc9548eee71de900bc2fb36128e/include/picongpu/param/particle.param#L59-L73)

share/picongpu/tests/Pusher/lib/python/test/Pusher/requirements.txt
```diff
@@ -0,0 +1,2 @@
+numpy version 1.24.2
```
BrianMarre: please change this line to
```python
numpy >= 1.24.2
```
otherwise you might create conflicting requirements later on

share/picongpu/tests/Pusher/lib/python/test/Pusher/requirements.txt
```diff
@@ -0,0 +1,2 @@
...
+openpmd_api
```
BrianMarre: same here, change to

```python
openpmd_api >= <the version you tested with>
```

share/picongpu/tests/PusherScaling/README.rst
```diff
@@ -0,0 +1,20 @@
...
+The test checks this factor by calculating the quotient of the phase lags between two timestep values. It then calculates the exponent x
+of d(phi) = 2^x (which should be 2) and the standard deviation of x (which should be zero). The exponent should not differ by more than
+epsilon = 0.1, and the standard deviation should not exceed delta = 0.05 from the ideal values.
+
+The thresholds epsilon and delta were determined as follows:
+---------------------------------------------------------------
+epsilon = 0.1 (for the exponent of d(phi) = 2^x, which should be 2)
+delta = 0.05 (for the standard deviation of x)
+
+They double the deviation found in recent PIConGPU runs (2023-09-07).
```
BrianMarre, line 18: "These limits are"

share/picongpu/tests/PusherScaling/include/picongpu/param/density.param
```diff
@@ -0,0 +1,81 @@
...
+             * @return float_X density [normalized to 1.0]
+             */
+            HDINLINE float_X operator()(const floatD_64& position_SI, const float3_64& cellSize_SI)
+            {
+                const pmacc::math::UInt64<simDim> cell_id(position_SI / cellSize_SI.shrink<simDim>());
+
+                // add particle in cell in at [ 32 5 16 ]
+                // X=32: middle of X plane (gyro-motion in X-Y)
+                // Y=5:  do not start fully at border, e.g., if someone wants to increase E, and so mass over time
+                // Z=16: middle of box in Z, move slowly in positive Z as E-field drift
```
BrianMarre: change to: ... in positive Z **due to** E-field drift

share/picongpu/tests/PusherScaling/include/picongpu/param/particle.param
```diff
@@ -0,0 +1,112 @@
...
+                 * Determines the weighting of a macro particle as well as the number of
+                 * macro-particles which sample the evolution of the particle distribution
+                 * function in phase space.
+                 *
+                 * unit: none
+                 */
+                static constexpr uint32_t numParticlesPerCell = 1u;
+
+                const InCellOffset_t inCellOffset;
+            };
```
BrianMarre: same as before please switch to the style used in the above drift velocity definition.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/requirements.txt
```diff
@@ -0,0 +1,2 @@
+numpy version 1.24.2
+openpmd_api
```
BrianMarre: same as before,

please specify as at least this tested version for future compatibility

BrianMarre (changes requested):
noticed a few things while  catching up on my notifications

share/picongpu/tests/Pusher/lib/python/test/Pusher/requirements.txt
chillenzer: Not 100% sure that I agree. That is typically a good idea for production code but in the tests? It's an isolated environment that doesn't interact with anything downstream and if it needs updating, it's probably better to isolate the effects of a changing dependency from an actual test failure.

BrianMarre: If our test keep the python dependencies separate for every runtime test, yes you are right.

But I would still prefer,
```python
numpy  == 1.24.2
```
;)

share/picongpu/tests/Pusher/include/picongpu/param/particle.param
MRZirper: done

share/picongpu/tests/PusherScaling/include/picongpu/param/particle.param
MRZirper: done

share/picongpu/tests/Pusher/lib/python/test/Pusher/MainTest.py
MRZirper: I implemented the function. Thanks for the advise!

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
MRZirper: done as in the other test.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
MRZirper: done

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
MRZirper: I removed it as in all other methods.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
MRZirper: done for the two magic numbers in both tests

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
MRZirper: I learned in my statistic class in the first and second semester, that the common standart deviation formula `std = sqrt(mean(x)), where x = abs(a - a.mean())**2` (which is also used by `numpy.std`) s a good approximation of the (gaussian) std. for a large series of data (ca. N > 10). If the data base is smaller, a good estimation of the std. is to use the `(max -min)/2` approach as you commented above. 
For the calculation there are just 4 datapoints so I chose this estimation. 

But I am open to use width to describe this value instead of std. I would keep sigma tho.

share/picongpu/tests/PusherScaling/lib/python/test/Pusher/MainTest.py
MRZirper: [Propagation of uncertanty](https://en.wikipedia.org/wiki/Propagation_of_uncertainty) is used here. The formula is derived from the formula of the exponent by applying a Taylor series including the uncertanty.

chillenzer (approved):

steindev:
@MRZirper please squash, rebase, and push again. Then I will merge.

BrianMarre:
@steindev CI failed due to anaconda returning `439: too many request`. Please restart the CI job

steindev (approved):

PR #4804, opened 2024-02-05
Make cPow's exponent unsigned

### LINKED ISSUE #4795: Negative powers not supported`
I've just seen that `pmacc::math::cPow` is wrong if the exponent is negative but this is neither stated nor enforced. One should at the very least add a note to the docstring or even better bake that into concept/SFINAE for template parameters.

chillenzer:
Closes #4795 by enforcing an unsigned integer type via the static assert. This should potentially rather be a `concept` or SFINAE but I kept as much of the infrastructure as possible. I've also `grep`ed for `cPow` in the root folder and checked that there were no faulty calls. Only the heat equation example was a bit sloppy about the type (albeit not violating positivity), so I made it an unsigned literal there.

### DIFF
--- a/include/pmacc/algorithms/math/PowerFunction.hpp
+++ b/include/pmacc/algorithms/math/PowerFunction.hpp
@@ -35,7 +35,7 @@ namespace pmacc::math
     /** power function for integer exponents, constexpr
      *
      * @tparam T_Type return and accumulation data type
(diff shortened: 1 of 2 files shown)


BrianMarre (approved):
Looks good, good catch!

psychocoderHPC (approved):

PR #4806, opened 2024-02-06
fix frontier profile

psychocoderHPC:
- use `clang++` instead of `hipcc`
- set environment variable `HIP_FLAGS` to support MPI
- remove outdated module openpmd-api (at least 0.15.X` is required)

# Review

The first commit renames the profile to reflect that we are not using `hipcc` anymore.
This will show a better diff in the second commit.


steindev:
Please don't merge. I have a few comments. Will discuss these with @psychocoderHPC tomorrow.

psychocoderHPC:
@steindev Please merge since both of us have vacations and the fix should be available for others.

steindev:
Will merge the working solution and refactor according to above comment after vacation.

steindev (approved):

PR #4810, opened 2024-02-08
fix collision particle invalid memory

psychocoderHPC:
Revert partly changes from #4745 to avoid invalid memory access. Switch back to traversing the frame list for each particle access.

BUg reported by @pordyna


pordyna (approved):
Tested with the crashing setup. Did run 200 steps 672 times without a crash. The setup would previously crash within the first few reruns. 

Still, this PR does not solve the issue with the very slow linear search and the resulting O(2) scaling.

chillenzer:
This can be closed, right?

psychocoderHPC:
This PR is closed, the bug was fixed in mallocMC, it was a data race because of missing threadfences to synchronize data changes.

PR #4811, opened 2024-02-08
refactor ListEntry used in collision

psychocoderHPC:
Refactor changes introduced in #4745 to avoid void using void pointer as a place holder for the later used frame list.

### DIFF
--- a/include/picongpu/particles/collision/detail/ListEntry.hpp
+++ b/include/picongpu/particles/collision/detail/ListEntry.hpp
@@ -32,52 +32,12 @@ namespace picongpu::particles::collision
 {
     namespace detail
     {
-        /** Access all particles in a cell.
-         *
-         * An accessor provides access to all particles within a cell (NOT supercell).
-         *
-         * @tparam T_FramePtrType frame pointer type
-         */
-        template<typename T_FramePtrType>
-        struct ParticleAccessor
-        {
-            T_FramePtrType* m_framePtrList = nullptr;
-            uint32_t* m_parIdxList = nullptr;
-            uint32_t m_numPars = 0u;
-
-            static constexpr uint32_t frameSize = T_FramePtrType::type::frameSize;
-
-            DINLINE ParticleAccessor(uint32_t* parIdxList, uint32_t const numParticles, T_FramePtrType* framePtrList)
-                : m_framePtrList(framePtrList)
-                , m_parIdxList(parIdxList)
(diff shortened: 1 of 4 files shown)


pordyna:
Still fails with:
```
Unhandled exception of type 'St13runtime_error' with message '/global/u2/p/pordyna/picongpu-setups/picongpu/thirdParty/cupla/alpaka/include/alpaka/event/EventUniformCudaHipRt.hpp(143) 'ret = TApi::eventQuery(event.getNativeHandle())' returned error  : 'cudaErrorIllegalAddress': 'an illegal memory access was encountered'!', terminating
  what():  /global/u2/p/pordyna/picongpu-setups/picongpu/thirdParty/cupla/alpaka/include/alpaka/event/EventUniformCudaHipRt.hpp(190) 'TApi::eventSynchronize(event.getNativeHandle())' returned error  : 'cudaErrorIllegalAddress': 'an illegal memory access was encountered'!

```

psychocoderHPC:
We will merge this PR before going back to the slow particle access in collision #4810.

We need to test in a physical setup if #4828 + #4825 is solving the Perlmutter issues @pordyna is observing.

include/picongpu/particles/collision/detail/ListEntry.hpp
```diff
@@ -87,24 +47,58 @@ namespace picongpu::particles::collision
...
+                }
+
+                /** get particle
+                 *
+                 * @param idx index of the particle, range [0;size())
+                 */
+                DINLINE auto operator[](uint32_t idx) const
+                {
+                    const uint32_t inSuperCellIdx = m_parIdxList[idx];
+                    return m_framePtrList[inSuperCellIdx / frameSize][inSuperCellIdx % frameSize];
```
steindev, line 95: Is it guaranteed that the division `inSuperCellIdx / frameSize` does not yield a floating point number? (A comment may state that)

pordyna: With both integers? I don't think so

psychocoderHPC: Both types a integral types, they can not end in floating point numbers, that are defined by the language.
A comment is not useful because it will increase the code size without extra new knowledge.

pordyna (approved):

ikbuibui (approved):
Looks okay

PR #4812, opened 2024-02-08
Add feature `--dump-metadata`

chillenzer:
- [x] rebase against dev branch after #4820 is merged

Dear interested parties,

This documentation draft specifies the interface that I propose to implement for dumping metadata. It intends to clearly state the usage, the scope and the limitations. It also sneaks in a bit of developer documentation through the `Customisation` section.

Your job now is (in that order):
* Read the proposal!
* Clearly state if anything remains unclear.
* Agree with the high-level design decisions or clearly state what you would do differently.
* Answer the following questions.
* Feel free to add a review of grammar, phrasing, etc. afterwards.

MISSING FEEDBACK ON MONDAY MORNING WILL BE CONSIDERED EQUIVALENT TO THE STATEMENT: "Dear @chillenzer, your proposal is mind-blowingly great and absolutely perfect. I can't possibly think of anything to do better. Please go ahead and implement exactly as stated. Lovingly, your all-time admirer!"

There are a number of design decisions and defaults that I feel others might have a better understanding of:
* Propose a default filename.
* One could activate this feature by default. Shall we?
* Naming is important but hard! Please speak up if you can think of better naming or if you find anything inconsistent (within this proposal or in relation to the rest of the codebase).
* As alluded to above, the `Customisation` section goes well into developer knowledge. We might want to split this off and put it somewhere more pertinent. This would also open the door for a more natural ordering of the information and inclusion of more fundamentals. Thoughts?

State of this proposal:
* This proposal does not contain functional code. It is only meant as a specification of what to implement and will survive as part of the documentation once we have agreed on everything.
* This file does most likely not render correctly, yet. I'm not an expert in restructured text, so I postponed the task of beautifying this. It should be sufficiently understandable from the source file and I'll try to render it next.
* The code examples lack two features that the final version is supposed to exhibit:
    - They are supposed to be drawn in automatically from tested code, so they never become outdated.
    - They are supposed to be less abstract, i.e., `MyClass` should become, e.g., a concrete laser to showcase this later. This will only be possible after we have an implementation and content from the physicists.
* I've used British English in the documentation. For consistency, one might want to change that.

Road map:
- [x] Everyone: Agree on the specification.
- [x] @chillenzer: Implement infrastructure including an example and merge basic infrastructure.
- [ ] Physicists: Fill in content, i.e., default metadata specifications wherever needed, in separate PRs.

Now, it's @JessicaTiebel's task to ping everyone who is considered an "interested party". I'll go ahead and relieve her from the burden of pinging @psychocoderHPC.


BrianMarre:
ping @BrianMarre

chillenzer:
Seems to render okay, only lacking a lot of features.

But good enough for reading:
https://picongpu--4812.org.readthedocs.build/en/4812/usage/metadata.html#dumping-metadata

BrianMarre:
Two points for discussion:
1.) what an experimentalist and a theoretician search by are substantially different,

     Experimentalist: "I am looking for simulations of formvar foils with a thickness of 500nm."
     Theoretician: "What simulations have we done for with an laser pedestal an a density of 50n_c?" 
     <own Simulations>: "What laser models did I use in the simulation campaign for LT54?"

Therefore I would asume we would need for the stated goal a MultiLevel-MultiEntry description to fulfill all searchability cases, Something along the line of

     1. Level/EntryPoint[Intent] Experiment description(foil(thickness, material, layers, surface, ...)/gasJet(gas, nozzle, density/shock model, ...), geometry experiment(laser angle, ..., )), laserDescription,  ...)
     2. Level/entryPoint Simulation[Setup] (InitDescription, Distributions, Species, Solvers, Box, Grids, output plugins, ...)

I get from your proposal that want to avoid the associated complexity, therefore IMO only the theoretician simulation database stuff will be possible.

2.) For format we could use something along the following line
PICMI_structure_DetailMiddle

BrianMarre:
> Seems to render okay, only lacking a lot of features.
> 
> But good enough for reading: https://picongpu--4812.org.readthedocs.build/en/4812/usage/metadata.html#dumping-metadata

You can get a rendered version directly via view file ;)

BrianMarre:
Further things:
- **name suggestions:** picongpu.meta, picongpuMetaDescription.json
- **software engineering:** I like the proposed architecture! :)

JessicaTiebel:
As requested, i ping people: @ktippey-hzdr @PrometheusPi @steindev

JessicaTiebel:
Question: It is unclear to me what is meant by  "partially customisable". At first I got the impression, that the user is intended to specify every metadata he/she wants to be written. But there is the intention of having some predefined data basically as a default or minimal output, right?

chillenzer:
Hey, thanks for your initial contributions. I'll already go ahead and answer as much as I can:

> Therefore I would asume we would need for the stated goal a MultiLevel-MultiEntry description to fulfill all searchability cases, Something along the line of
> 
> ```
>  1. Level/EntryPoint[Intent] Experiment description(foil(thickness, material, layers, surface, ...)/gasJet(gas, nozzle, density/shock model, ...), geometry experiment(laser angle, ..., )), laserDescription,  ...)
>  2. Level/entryPoint Simulation[Setup] (InitDescription, Distributions, Species, Solvers, Box, Grids, output plugins, ...)
> ```
> 
> I get from your proposal that want to avoid the associated complexity, therefore IMO only the theoretician simulation database stuff will be possible.
>
> 2.) For format we could use something along the following line...

That sounds like good ideas but is not what I consider the scope of this PR. If I understand you correctly, this is what I would consider "content". The `json` format in use is capable of mapping to such a structure and the necessity to do so has to be assessed by domain experts (as you might be one) when filling in the default information to dump. The latter (see Roadmap) is supposed to be separate PRs.
 
>     * **name suggestions:** picongpu.meta, picongpuMetaDescription.json

Sounds good, we'll collect some more before we make a final decision. Where would you put the file then?
 
>     * **software engineering:** I like the proposed architecture! :)

Thanks! =)


> Question: It is unclear to me what is meant by "partially customisable". At first I got the impression, that the user is intended to specify every metadata he/she wants to be written. But there is the intention of having some predefined data basically as a default or minimal output, right?

Yes, the idea is that this is a feature that works "out-of-the-box" in the sense that (see Roadmap) once the infrastructure is implemented, domain experts come along and fill in default values for what's necessary to know. Despite being the longest section of the draft, customisation should typically not be necessary afterwards, i.e. with the first two lines of the documentation you should basically be able to get an output similar to your `meta_data_example.yaml`.

That's why I consider it kind-of developer documentation because it will mostly be interesting in cases when you develop a new feature information of which you want to include into the metadata dump. But there might be rare cases when just a regular user has a particular interest in this one property that is not in the default dump and so they can "partially customise" (in theory even fully if they bring a lot of time and nerve) their personal metadata output.

Is there any point where this could be formulated more explicitly? I would hope that the currently empty example in the format section would make this less ambiguous.

chillenzer:
### Update

The feature is in a functioning state. In particular, you should be able to `pic-create` and `pic-build` the `LaserWakefield` and run
```bash
./bin/picongpu -d 1 1 1 -g 24 24 24 --dump-metadata --no-start-simulation
```
(from the `input` directory).

The feature is not yet ready to merge. There is a lot of cleaning up left. In particular, suggestions on an appropriate file structure would be welcome. For development, I've simply put it into the main `include` path but likely you've got a better suggestion for how to structure this in the spirit of the project. The other thing to note is that the documentation is outdated because of some minor issues in confronting it with reality.

There are two moderate design flaws in the compile-time realm for which I couldn't find a satisfying solution within the C++ language:
- We currently need to assemble the list of compile-time metadata in one global place. This is a tiny bit against the design goal of every class handling their own metadata but I couldn't come up with a better way. Feel free to approach me, if you want to know what I tried so far.
- We're using `static` member functions for the default metadata of CT classes. These are bound to types and use the (static) data of that type. They are only inherited in a weak sense, i.e., they can be called from a derived type but will use the data from the base type to which they are bound. This means that one cannot -- as I had hoped for -- build up a nice successive set of metadata by inheriting a implementation and only adding one or two specialties of the derived class (this only works at RT). I've implemented a solution that smells a tiny bit like CRTP and works just fine but is probably not a nice interface for a C++ layperson to add their own metadata. (If that person would have solved their problem via inheritance in the first place is a different question.)

chillenzer:
Dear interested parties,

This feature is ready for review! Actually, it's been since Monday or Tuesday or so (up to a handful of doxygen comments and a very painful rebase) but the stupid CI just didn't want to run. Now, that it has finally run I've fixed one (hopefully) last oversight of mine and pushed the rebased version but you can go ahead and review already.

Current status: The code provides the cmdline option to dump metadata but the generated file is very much a stub containing just enough information to make a useful example how to extract RT and CT information. It will be your task to fill the output with content in separate PRs. In particular, @BrianMarre, you might want to come back to this and add similar content to the `.mustache` templates for PyPIConGPU after the other PRs are through.

Some hints for the review:
- I have tried to provide decent documentation. You might want to start with that, so you can get your head around what I was trying to achieve. I would recommend using the rendered version *(by readthedocs)* that can be accessed from the CI block.
- There are four different kinds of changes:
  * Main functionality (`traits/GetMetadata.hpp` and `MetadataAggregator.hpp`)
  * Implementation of the book-keeping in `simulation/control/SimulationStarter.hpp` and `param/metadata.param`
  * Unit tests (in `share/picongpu/unit/metadata/`) and one end-to-end test in `share/picongpu/tests/metadataFromLaserWakefield`
  * Peripheral changes that provide examples and implement some stubs, so all the examples, etc., compile
- Looking at the tests is a great way to get familiar with how it's supposed to work.
- Feel free to suggest better defaults for the filename and/or active/inactive by default (a third option would be: inactive by default but activated in all `.tpl` files such that most users would have it activated).

From above, I ping again:
@JessicaTiebel, @psychocoderHPC, @BrianMarre, @ktippey-hzdr, @PrometheusPi, @steindev

share/picongpu/tests/metadataFromLaserWakefield/picongpu-metadata.json.reference
```diff
@@ -0,0 +1,22 @@
+{
+  "incidentField": {
+    "XMax": {},
+    "XMin": {},
+    "YMax": {},
+    "YMin": {
+      "polarisation": {
```
psychocoderHPC: Offline discussed: We can have many lasers per direction therefore it is important that each laser is getting a name

docs/source/usage/metadata.rst
```diff
@@ -0,0 +1,215 @@
...
+
+You might want to dump metadata without actually running a simulation. In this case, you can add the
+`--no-start-simulation` flag which will make the code skip the actual simulation. If your intention is instead to also
+run the simulation afterwards, just leave it out and the program will proceed as normal.
+
+The dumping happens after all initialisation work is done immediately before the simulation starts (or is skipped). This implies that
+
+ * No dynamic information about the simulation can be included (e.g. information about the state at time step 10).
+ * The dumped information will represent the actual parameters used which might be different from the parameters given
+   in the input files, e.g., due to automatic adjustment of the grid size (see :ref:`autoAdjustGrid<usage-cfg>`).
```
psychocoderHPC: The grid size adjustment should be in your meta data because you create the metadata after the initialization where the adjustment already happened.

IMO this line can be removed.

share/picongpu/tests/metadataFromLaserWakefield/picongpu-metadata.json.reference
chillenzer: More precisely: Having a name is content which is later added by the domain experts. The important thing to make this correct as an infrastructure is that each entry "XMin", ... maps to a list of objects (which might or might not have a name then).

include/picongpu/MetadataAggregator.hpp
```diff
@@ -0,0 +1,145 @@
...
+#include <ostream>
+#include <string>
+
+#include <nlohmann/json.hpp>
+#include <picongpu/traits/GetMetadata.hpp>
+
+
+namespace picongpu
+{
+    using boost::program_options::options_description;
```
psychocoderHPC: Can we move this out of the namespace else we will produce very easy side effects?
Pulling stuff into the project namespace is not a good practice. 
We do this only for PMacc because we have a very strong relation between these projects.

include/picongpu/MetadataAggregator.hpp
```diff
@@ -0,0 +1,145 @@
...
+#pragma once
+
+#include <pmacc/boost_workaround.hpp>
+
+#include <pmacc/pluginSystem/IPlugin.hpp>
+
+#include <boost/program_options/options_description.hpp>
+#include <boost/program_options/value_semantic.hpp>
+
+#include <filesystem>
```
psychocoderHPC: please include `<pmacc/filesystem>` we still support compiler where filesystem is under the experimental namespace

include/picongpu/simulation_defines.hpp
```diff
@@ -53,6 +53,10 @@ namespace picongpu
 // ##### load unitless
 #include <picongpu/_defaultUnitless.loader>
 #include <picongpu/extensionUnitless.loader>
+
+// is added last, so that any possible metadata from the param files can be added:
+#include "picongpu/param/metadata.param"
```
psychocoderHPC: typically we include params in `_defaultParam.loader` except for a few params where we need to solve cyclic dependencies.

share/picongpu/unit/CMakeLists.txt
```diff
@@ -76,6 +103,12 @@ include_directories(BEFORE ./include)
 # CTest
 enable_testing()
 
+# We need to build and link this to check CLI aspects.
+add_library(argparser
+  "${CMAKE_CURRENT_SOURCE_DIR}/../../../include/picongpu/ArgsParser.cpp"
+  "${CMAKE_CURRENT_SOURCE_DIR}/../../../include/picongpu/versionFormat.cpp"
+)
+
```
psychocoderHPC, line 111: `target_link_libraries(argparser ${HOST_LIBS})` is missing

share/picongpu/unit/metadata/metadataCLI.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+#include <boost/program_options/options_description.hpp>
+
+#include <algorithm>
+#include <filesystem>
+#include <string>
+
+#include <catch2/catch_test_macros.hpp>
+
+using boost::program_options::options_description;
```
psychocoderHPC, line 33: in the test cases this is fine because this will not propagate into picongpu, it stays into the cpp file

share/picongpu/unit/metadata/metadataDescription.cpp
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    SECTION("customisation can extract private information as a friend")
+    {
+        int privateInfo = 42;
+        SomethingWithoutUsefulMetadata obj{privateInfo};
+        auto expected = json::object();
+        expected["privateInfo"] = privateInfo;
+        expected["customisedInfo"] = "Some other customised string.";
+        REQUIRE(obj.metadata() != expected); // make sure we test something non-trivial here
```
psychocoderHPC: please comment above the line. Adding comments behind the line ends up sometime ins strange unreadable comments if the line or indention is changing.

Sometimes this happen
```C++
The line become longer than in the initial PR; REQUIRE(obj.metadata() != expected); 
// make sure we test something non-trivial here
```

psychocoderHPC (changes requested):

pordyna:
I absolutely love this PR!

pordyna:
@BrianMarre 
> Further things:
> 
>     * **name suggestions:** picongpu.meta, picongpuMetaDescription.json

I think we shouldn't touch the .json extension as long as we write in the json format.

docs/source/usage/metadata.rst
BrianMarre: I disagree, I consider it a reminder for a reminder for a deviation from anticipated behaviour of some one who does not know that PIConGPU overwrite some user input if necessary.

BrianMarre:
current piconpgu-metadata.json is fine from my side, actually prefer it over my suggestions

docs/source/usage/metadata.rst
chillenzer: Forgot to add: In offline discussion, we agreed that the line is technically correct but apparently has a potential to be misunderstood, so I'll look for a rephrasing.

chillenzer: Next push provides a new suggestion. See if it's to your liking.

include/picongpu/simulation_defines.hpp
chillenzer: Offline discussion: It is yet unclear if there is a need to have this configurable. As long as we don't know, we will spare us the maintenance overhead and make this a regular header.

psychocoderHPC (approved):

PR #4819, opened 2024-02-19
remove mentioning of PoGit

PrometheusPi:
This pull request removes the mentioning of PoGit as suggested in ##2918


chillenzer (approved):
It is really hard to find anything that could possibly go wrong with this. Reassuringly
```bash
grep -ril pogit
```
in the root directory returns only this file, so I guess you're good to go!

psychocoderHPC (approved):

PR #4821, opened 2024-02-20
remove deactivation of ISAAC with HIP

### LINKED ISSUE #4714: Does ISAAC work with HIP?
In our cmake files we still state that ISAAC does not work with HIP:
https://github.com/ComputationalRadiationPhysics/picongpu/pull/4712#discussion_r1366671254

Probably, this is not the case anymore. This needs to be tested, and then the comment needs to be fixed.

PrometheusPi:
This pull request closes #4714 

ISAAC works with HIP as tested on Frontier.

### DIFF
--- a/share/ci/compiler_hipcc.yml
+++ b/share/ci/compiler_hipcc.yml
@@ -6,8 +6,7 @@
   image: registry.hzdr.de/crp/alpaka-group-container/alpaka-ci-${CI_CONTAINER_NAME}-rocm${HIP_CONTAINER_VERSION}-pic:${CONTAINER_TAG}
   variables:
     GIT_SUBMODULE_STRATEGY: normal
-    # ISAAC is not working with HIP
-    DISABLE_ISAAC: "yes"
+    DISABLE_ISAAC: "no"
     # CI_GPU_ARCH architecture of the hosted GPU is provided by the CI
     GPU_TARGETS: ${CI_GPU_ARCH}
   script:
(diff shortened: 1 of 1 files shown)


PrometheusPi:
@psychocoderHPC This pull request now activated ISAAC - let's see if the problems still exist

share/ci/compiler_hipcc.yml
```diff
@@ -6,8 +6,7 @@
   image: registry.hzdr.de/crp/alpaka-group-container/alpaka-ci-${CI_CONTAINER_NAME}-rocm${HIP_CONTAINER_VERSION}-pic:${CONTAINER_TAG}
   variables:
     GIT_SUBMODULE_STRATEGY: normal
-    # ISAAC is not working with HIP
-    DISABLE_ISAAC: "yes"
+    DISABLE_ISAAC: "no"
```
psychocoderHPC: Could you please remove this line fully? We have places in the scripts where we check only if the variable is set. IN this case we do not use ISAAC.
IMO ISAAC is still disabled and therefore all passed the CI.

psychocoderHPC (changes requested):

share/ci/compiler_hipcc.yml
PrometheusPi: @psychocoderHPC done

psychocoderHPC (approved):

Issue #4818, opened 2024-02-19
Doc: Coding Style


chillenzer:
Hi, just noticed that the [Commit rules](https://picongpu.readthedocs.io/en/latest/dev/docs/COMMIT.html) page talks about coding style guidelines but actually links to [the contributing repo](https://github.com/ComputationalRadiationPhysics/contributing) and not the [Coding Guidelines](https://picongpu.readthedocs.io/en/latest/dev/styleguide.html) page of the docs. Also, the latter seems to be outdated (at the very least concerning the `clang-format` version requirements). We should clean this up.

psychocoderHPC:
We should change the link and add a sentence that code style changes will be voted in the separate repository.

PR #4823, opened 2024-02-20
adjust link and explanation code styleguide

PrometheusPi:
This pull request solves #4818 

It adds the link to the style guide and explains that the repo linked before is the discussion on the code style.

- [ ] check if links in `*.md` work as in `*.rst`

### DIFF
--- a/docs/COMMIT.md
+++ b/docs/COMMIT.md
@@ -6,7 +6,7 @@ We agree on the following simple rules to make our lives easier :)
 - Stick to the **style** below for **commit messages**
 - **Commit compiling patches** for the `dev` branch,
   you can be less strict for (unshared) *topic branches*
-- Follow the code style and formatting which is democratically evolved in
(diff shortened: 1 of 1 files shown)


chillenzer:
Thanks for taking care of this! To my own surprise the `[name](link)` syntax seems to work. But the `:ref:...` syntax failed.

chillenzer:
Content-wise this seems okay. It's a very minimal solution and does not fully resolve #4818. That was more targeted towards a general overhaul of also `docs/source/dev/styleguide.rst` because that's outdated, too. But don't let the perfect be the enemy of the good, so please go ahead with this one but leave the issue open.

PrometheusPi:
The linking works differently in `*.md` then in `*.rst` - this the link is not shown properly. 
**Please do not merge.**

PrometheusPi:
Linking runs into the following issue #4831

chillenzer:
Which is probably precisely the reason why this was not done properly last time. XD Feel free to close #4831 first!

PR #4825, opened 2024-02-22
fix `atomicAllInc()`

psychocoderHPC:
The implementation of `atomicAllInc()` is broken for Volta and newer CUDA GPUs. CUDA warps are not implicit synchroniced and using warp functions can have side effects if we assume that asking for the active threads mask `activemask()` twice will result into the same result.

This bug exists in since 7 years or more in PIConGPU. I only observed problems if I compile in debug mode but in general the release mode should be affected too.

Possible effects of the bug:
- particles silently get deleted and other particles get cloned
- invalid memory access
- maybe it is also possible that we create particles out of the thin air (this is not fully clear)

### DIFF
--- a/include/pmacc/kernel/atomic.hpp
+++ b/include/pmacc/kernel/atomic.hpp
@@ -78,6 +78,7 @@ namespace pmacc
              * http://devblogs.nvidia.com/parallelforall/cuda-pro-tip-optimized-filtering-warp-aggregated-atomics/
              * (author: Andrew Adinetz, date: October 1th, 2014)
              *
+             * For modern architectures see https://developer.nvidia.com/blog/using-cuda-warp-level-primitives/ *
              */
             template<typename T_Type>
             struct AtomicAllIncKepler<T_Type, true>
@@ -85,6 +86,10 @@ namespace pmacc
                 template<typename T_Acc, typename T_Hierarchy>
                 HDINLINE T_Type operator()(const T_Acc& acc, T_Type* ptr, const T_Hierarchy& hierarchy)
                 {
+                    /* @attention mask must be used in any warp operation which supports a mask.
+                     * On CUDA calling activemask again could result int different results because warps are not
+                     * implicitly synchronized. This is different to HIP and old CUDA GPUs before Volta.
+                     */
                     const auto mask = alpaka::warp::activemask(acc);
                     const auto leader = alpaka::ffs(acc, static_cast<std::make_signed_t<decltype(mask)>>(mask)) - 1;
 
@@ -97,7 +102,7 @@ namespace pmacc
                             ptr,
                             static_cast<T_Type>(alpaka::popcount(acc, mask)),
                             hierarchy);
-                    result = warpBroadcast(result, leader);
+                    result = warpBroadcast(mask, result, leader);
                     /* Add offset per thread */
                     return result
                         + static_cast<T_Type>(
--- a/include/pmacc/kernel/warp.hpp
+++ b/include/pmacc/kernel/warp.hpp
@@ -53,66 +53,87 @@ namespace pmacc
 
 
 #    if(__CUDA_ARCH__ >= 300 || BOOST_COMP_HIP)
-        /** broadcast data within a warp
+
+        /** broadcast data within a warp without using shared memory
          *
-         * required PTX ISA >=3.0
-         *
+         * @param mask mask to select threads participating in the operation
          * @param data value to broadcast
          * @param srcLaneId lane id of the source thread
          * @return value send by the source thread
-         *
-         * \{
          */
-        //! broadcast a 32bit integer
-        DINLINE int32_t warpBroadcast(const int32_t data, const int32_t srcLaneId)
+        template<typename T_MaskType, typename T_DataType>
+        DINLINE T_DataType warpBroadcast(T_MaskType const mask, T_DataType const data, int const srcLaneId);
+
+        /** check if T is equal to at least one type in the tuple list
+         *
+         * @param tuple of types, each type is checked against T
+         * @return true if T is equal to at least one type from the tuple list, else false
+         */
+        template<typename T, typename... T_TypeList>
+        inline constexpr bool matchAnyType(std::tuple<T_TypeList...>)
         {
-#        if(__CUDACC_VER_MAJOR__ >= 9)
-            return __shfl_sync(__activemask(), data, srcLaneId);
-#        else
-            return __shfl(data, srcLaneId);
+            return std::disjunction_v<std::is_same<T, T_TypeList>...>;
+        }
+
+
+        /** fallback for not natively supported types
+         *
+         * emulates the broadcast by repeating 32bit warp broadcasts
+         */
+        template<typename T_MaskType, typename T_DataType, typename T_Sfinae = void>
+        struct WarpBroadcast
+        {
+            DINLINE T_DataType operator()(T_MaskType const mask, T_DataType data, int const srcLaneId) const
+            {
+                static_assert(sizeof(T_DataType) >= sizeof(int) && (sizeof(T_DataType) % sizeof(int)) == 0);
+                int* const pData = reinterpret_cast<int*>(&data);
+                for(int i = 0; i < sizeof(T_DataType) / sizeof(int); ++i)
+                    pData[i] = warpBroadcast(mask, pData[i], srcLaneId);
+
+                return data;
+            }
+        };
+
+        /** List of types natively supported by CUDAs and HIPs shfl function interface */
+        using VendorSupportedTypesForShfl = std::tuple<
+            int,
+            unsigned,
+            float
+        /* clang as cuda compiler does not support 64 bit warp shfl.
+         * For AMD GPUs 64bit shufl will be provided by HIP.
+         */
+#        if(!BOOST_COMP_CLANG_CUDA)
+            ,
+            double,
+            unsigned long,
+            unsigned long long,
+            long,
+            long long
 #        endif
-        }
+            >;
 
-        //! Broadcast a 64bit integer by using 2 32bit broadcasts
-        DINLINE int64_cu warpBroadcast(int64_cu data, const int32_t srcLaneId)
+        /** CUDA and HIP native supported data types */
+        template<typename T_MaskType, typename T_DataType>
+        struct WarpBroadcast<
+            T_MaskType,
+            T_DataType,
+            std::enable_if_t<matchAnyType<T_DataType>(VendorSupportedTypesForShfl{})>>
         {
-            int32_t* const pData = reinterpret_cast<int32_t*>(&data);
-            pData[0] = warpBroadcast(pData[0], srcLaneId);
-            pData[1] = warpBroadcast(pData[1], srcLaneId);
-            return data;
-        }
-
-        //! Broadcast a 32bit unsigned int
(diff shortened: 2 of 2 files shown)


include/pmacc/kernel/atomic.hpp
```diff
@@ -78,13 +78,18 @@ namespace pmacc
...
              *
+             * For modern architectures see https://developer.nvidia.com/blog/using-cuda-warp-level-primitives/ *
              */
             template<typename T_Type>
             struct AtomicAllIncKepler<T_Type, true>
             {
                 template<typename T_Acc, typename T_Hierarchy>
                 HDINLINE T_Type operator()(const T_Acc& acc, T_Type* ptr, const T_Hierarchy& hierarchy)
                 {
+                    /* @attention mask must be used in any warp operation whcih supports a mask.
```
chillenzer:
```suggestion
                    /* @attention mask must be used in any warp operation which supports a mask.
```

include/pmacc/kernel/atomic.hpp
```diff
@@ -78,13 +78,18 @@ namespace pmacc
...
+                     * On CUDA calling activemask again could result into different results because warps are not
```
chillenzer:
```suggestion
                     * On CUDA calling activemask again could result in different results because warps are not
```

include/pmacc/kernel/atomic.hpp
```diff
@@ -78,13 +78,18 @@ namespace pmacc
...
+                     * implicit synchronized. This is different to HIP and old CUDA GPUs before Volat.
```
chillenzer:
```suggestion
                     * implicitly synchronized. This is different to HIP and old CUDA GPUs before Volta.
```

include/pmacc/kernel/warp.hpp
```diff
@@ -53,66 +53,70 @@ namespace pmacc
...
 
-        //! Broadcast a 32bit unsigned int
-        DINLINE uint32_t warpBroadcast(const uint32_t data, const int32_t srcLaneId)
-        {
-            return static_cast<uint32_t>(warpBroadcast(static_cast<int32_t>(data), srcLaneId));
-        }
 
-        //! Broadcast a 64bit unsigned int
-        DINLINE uint64_cu warpBroadcast(const uint64_cu data, const int32_t srcLaneId)
+        /** fallback for not nativly supported types
```
chillenzer:
```suggestion
        /** fallback for not natively supported types
```

include/pmacc/kernel/warp.hpp
```diff
@@ -53,66 +53,70 @@ namespace pmacc
...
+         *
+         * emulates the brotcast by repeating 32bit warp broatcasts
+         */
+        template<typename T_MaskType, typename T_DataType, typename T_Sfinae = void>
+        struct WarpBroatcast
```
chillenzer:
```suggestion
        struct WarpBroadcast
```

include/pmacc/kernel/warp.hpp
```diff
@@ -53,66 +53,70 @@ namespace pmacc
...
+        };
+
+        /** CUDA and HIP native supported data types */
+        template<typename T_MaskType, typename T_DataType>
+        struct WarpBroatcast<
+            T_MaskType,
+            T_DataType,
+            std::enable_if_t<matchAnyType<
+                T_DataType,
+                int,
```
chillenzer: Could we give this list a name somehow? Haven't looked up the details but it would be much nicer to have something like
```c++
using NativelySupportedTypes = ???;
// ...
MatchAnyType<T_DataType, NativelySupportedTypes...>;
```

include/pmacc/kernel/warp.hpp
```diff
@@ -53,66 +53,70 @@ namespace pmacc
...
-        DINLINE uint32_t warpBroadcast(const uint32_t data, const int32_t srcLaneId)
-        {
-            return static_cast<uint32_t>(warpBroadcast(static_cast<int32_t>(data), srcLaneId));
-        }
 
-        //! Broadcast a 64bit unsigned int
-        DINLINE uint64_cu warpBroadcast(const uint64_cu data, const int32_t srcLaneId)
+        /** fallback for not nativly supported types
+         *
+         * emulates the brotcast by repeating 32bit warp broatcasts
```
chillenzer:
```suggestion
         * emulates the broadcast by repeating 32bit warp broadcasts
```

include/pmacc/kernel/warp.hpp
```diff
@@ -53,66 +53,70 @@ namespace pmacc
...
 
-        //! Broadcast a 32bit float
-        DINLINE float warpBroadcast(const float data, const int32_t srcLaneId)
+                return data;
+            }
+        };
+
+        /** CUDA and HIP native supported data types */
+        template<typename T_MaskType, typename T_DataType>
+        struct WarpBroatcast<
```
chillenzer:
```suggestion
        struct WarpBroadcast<
```

include/pmacc/kernel/warp.hpp
```diff
@@ -53,66 +53,70 @@ namespace pmacc
...
-        //! Broadcast a 64bit float by using 2 32bit broadcasts
-        DINLINE double warpBroadcast(double data, const int32_t srcLaneId)
+        template<typename T_MaskType, typename T_DataType>
+        DINLINE T_DataType warpBroadcast(T_MaskType const mask, T_DataType const data, int const srcLaneId)
         {
-            float* const pData = reinterpret_cast<float*>(&data);
-            pData[0] = warpBroadcast(pData[0], srcLaneId);
-            pData[1] = warpBroadcast(pData[1], srcLaneId);
-            return data;
+            return WarpBroatcast<T_MaskType, T_DataType>{}(mask, data, srcLaneId);
```
chillenzer:
```suggestion
            return WarpBroadcast<T_MaskType, T_DataType>{}(mask, data, srcLaneId);
```

chillenzer: Can't we rely on type deduction here?

include/pmacc/kernel/warp.hpp
```diff
@@ -53,66 +53,70 @@ namespace pmacc
...
+         */
+        template<typename T_MaskType, typename T_DataType, typename T_Sfinae = void>
+        struct WarpBroatcast
         {
-            return static_cast<uint64_cu>(warpBroadcast(static_cast<int64_cu>(data), srcLaneId));
-        }
+            DINLINE T_DataType operator()(T_MaskType const mask, T_DataType data, int const srcLaneId) const
+            {
+                static_assert(sizeof(T_DataType) >= sizeof(int) && (sizeof(T_DataType) % sizeof(int)) == 0);
+                int* const pData = reinterpret_cast<int*>(&data);
```
chillenzer, line 89: The previous implementation was more precise about the type here: `sizeof(int)` is not necessarily 32bit. Should we use `int32_t` here as before?

include/pmacc/kernel/warp.hpp
```diff
@@ -53,66 +53,70 @@ namespace pmacc
...
-        /** broadcast data within a warp
-         *
-         * required PTX ISA >=3.0
+
+        /** broadcast data within a warp without using shared memory
          *
+         * @param mask mask to select threads participating in the operation
          * @param data value to broadcast
          * @param srcLaneId lane id of the source thread
-         * @return value send by the source thread
```
chillenzer, line 62: Why is this deleted?

include/pmacc/kernel/warp.hpp
```diff
@@ -53,66 +53,70 @@ namespace pmacc
 
 
 #    if(__CUDA_ARCH__ >= 300 || BOOST_COMP_HIP)
-        /** broadcast data within a warp
-         *
-         * required PTX ISA >=3.0
```
chillenzer, line 58: Was this requirement dropped?

chillenzer (changes requested):
Hi, I could have waited for the CI but somehow I didn't. Looks good so far. Very nice how you abstracted away all the different special cases into one elegant template.

My comments are mostly typos (and I think GH has a button for immediately accepting the suggestions in simple cases) but I think I've found a mismatch of signatures that we should check. Might have overlooked something though.

include/pmacc/kernel/warp.hpp
psychocoderHPC: This would be nice but would make the generator expression more complicated because I need the variable template signature. ONly a workaround with `std::tuple` and unfolding the tuple with `std::Get<>` would come into my mind, but IMO the expression will be much more complicated to read.

include/pmacc/kernel/warp.hpp
psychocoderHPC: I changed it back to `int` because the native interfaces requires `int` and in principle `int32_t` could be anything.
So the new implementation will guarantee that all works. The old most likely worked but  if we are pedantic than there is no guarantee.

include/pmacc/kernel/warp.hpp
psychocoderHPC: All supported CUDA architectures are 3.5 anyway and even 3.5 is not supported by CUDA 12.X anymore.
So no need to mention it because it is supported for all GPUs (NVIDIA and AMD)

include/pmacc/kernel/warp.hpp
psychocoderHPC: No this is not working because I do not forwarded the arguments to the constructor so the type can not be deducted.

psychocoderHPC:
I applied @chillenzer comments. I hit a new record in typos per line of code^^

include/pmacc/kernel/warp.hpp
chillenzer: It isn't too ugly:
```c++
#include <iostream>
#include <type_traits>

template <typename...> struct TypeList {};

using NativeTypes = TypeList<int, float>;

template <typename T, typename... T_TypeList,
          template <typename...> typename T_Pack>
inline constexpr bool matchAnyType(T_Pack<T_TypeList...>) {
  return (std::is_same_v<T, T_TypeList> || ...);
}

template <typename T, typename = void> struct F {
  F() { std::cout << "Called fallback!\n"; }
};

template <typename T>
struct F<T, std::enable_if_t<matchAnyType<T>(NativeTypes{})>> {
  F() { std::cout << "Called implementation!\n"; }
};

int main() {
  std::cout << matchAnyType<int>(NativeTypes{}) << "\n";
  std::cout << matchAnyType<float>(NativeTypes{}) << "\n";
  std::cout << matchAnyType<char>(NativeTypes{}) << "\n";

  F<int>();
  F<float>();
  F<char>();
}
```

include/pmacc/kernel/warp.hpp
chillenzer: Okay. Admittedly, it is kind of the reason why we have these wrapper functions after all...^^

include/pmacc/kernel/warp.hpp
psychocoderHPC: I will try https://godbolt.org/z/4nzG1eree based on your solution

chillenzer: Nice one! Didn't know about `std::disjunction`!

psychocoderHPC: applied the changes to the latest version of this PR

chillenzer (approved):

steindev (approved):
Relying on @chillenzer's review :sweat_smile:

PR #4834, opened 2024-02-28
fix growth rate and charge conservation test tools

psychocoderHPC:
- fix using depricated `np.float` and `np.int`
- change `import importlib` to `import importlib.util`


ikbuibui (approved):
Changes don't affect behaviour. Approved.

PR #4838, opened 2024-03-04
replace cupla fully by alpaka

psychocoderHPC:
- update clang-format and remove cupla include rule
- update CI checks for cupla includes
- update documentation (dependency image)
- update code comments
- CMake: use alpaka subtree directly and not the version alpaka version
  shipped with cupla
- update the license section for alpaka
- remove cupla subtree
- add alpaka subtree with version 1.1.0

```
GIT_AUTHOR_NAME="Third Party" GIT_AUTHOR_EMAIL="picongpu@hzdr.de" git subtree add --prefix thirdParty/alpaka git@github.com:alpaka-group/alpaka.git 1.1.0 --squash
```

# review

Please review the first two commits and the last. Keep in mind I can not rebase this PR because it contains a git subtree update.


ikbuibui (approved):
Changes look fine. I did not check if there are any places cupla is still refereed to though. We can deal with those as they are spotted in the future

psychocoderHPC:
> Changes look fine. I did not check if there are any places cupla is still refereed to though. We can deal with those as they are spotted in the future

I checked if cupla is used anywhere. All mentions should be replaced.

PR #4843, opened 2024-03-08
changed timeOffset variable type to float_32

FilipO28555:
While running PIConGPU in 64 bit precision and writing particle data. PIConGPU threw folowing runtime error:
image

While reading data using openPMD api there was a following error:
```
ValueError: [Thu Mar 07 18:27:54 2024] [ADIOS2 EXCEPTION] <Core> <IO> <DefineAttribute> : modifiable attribute /data/0/particles/e/position/timeOffset has been defined with type float. Type cannot be changed to double
```

`bpls` also threw an error:
```
Failed to open with BPFile engine: [Fri Mar 08 13:24:13 2024] [ADIOS2 EXCEPTION] <Core> <IO> <DefineAttribute> : modifiable attribute /data/0/particles/e/position/timeOffset has been defined with type float. Type cannot be changed to double
```

### DIFF
--- a/include/picongpu/plugins/openPMD/writer/ParticleAttribute.hpp
+++ b/include/picongpu/plugins/openPMD/writer/ParticleAttribute.hpp
@@ -147,7 +147,7 @@ namespace picongpu
                 /* @todo check if always correct at this point,
                  * depends on attribute and MW-solver/pusher implementation
                  */
-                float_X const timeOffset = 0.0;
+                float_32 const timeOffset = 0.0;
                 record.setAttribute("timeOffset", timeOffset);
 
                 log<picLog::INPUT_OUTPUT>("openPMD:  ( end ) write species attribute: %1%") % Identifier::getName();


PrometheusPi (approved):
This change fixes the issue. 

@franzpoeschel since this is coming from ADIOS: should we apply this fix here or wait for a fix in ADIOS?

include/picongpu/plugins/openPMD/writer/ParticleAttribute.hpp
```diff
@@ -147,7 +147,7 @@ namespace picongpu
                 /* @todo check if always correct at this point,
                  * depends on attribute and MW-solver/pusher implementation
                  */
-                float_X const timeOffset = 0.0;
+                float_32 const timeOffset = 0.0;
                 record.setAttribute("timeOffset", timeOffset);
```
psychocoderHPC, line 151: offline discussed with @PrometheusPi please try to use 

```C++
record.setTimeOffset<float_X>(0.0);
```

For time offset there is a specific interface in OpenPMD maybe it is not allowed to set this with `setAttribute`

@franzpoeschel do you see why maybe a hard coded 32bit value could be nessesary?

psychocoderHPC:
Which OpenPMD version do you use and which Adios?

franzpoeschel:
The problem is on openPMD's side here. One of our upcoming projects is to entirely redo the logic for specifying scientific default values, since they are currently done  too simply. Looking into openPMD's source code:

```cpp
Record::Record()
{
    setTimeOffset(0.f);
}
```
This leads to the following behavior inside PIConGPU:
```cpp
auto pos_e = iteration.particles["e"]["position"];
// timeOffset is implicitly initialized by openPMD to 0.f

// write data to pos_e
pos_e.flush();

// now write metadata
pos_e.setTimeOffset<double>(10.);
// attribute was already created as a float, the backend will complain
```

The workaround (until the point where we properly fix this behavior and introduce a proper system for specifying default values in a less hardcoded way) would be to write such metadata as early as possible, before the first flush inside PIConGPU. The values can be changed within one iteration as long as the types don't change.

I can try doing that.

franzpoeschel:
Thank you for bringing this up @FilipO28555. 
This issue can be fixed without changing the output datatype, so I opened an alternative PR at #4844.

As written above, the long-term solution must be approached inside openPMD.

psychocoderHPC:
solved by https://github.com/ComputationalRadiationPhysics/picongpu/pull/4844

PR #4846, opened 2024-03-12
updated gpu.profile modules (adapted to fwkt_v100.profile)

MRZirper:
The modules of the `gpu_picongpu.profile.example` were outdated. A successful PIConGPU build was not possible so I updated them and adapted the structure and versions of the `fwkt_v100_picongpu.profile.example`.

### DIFF
--- a/etc/picongpu/hemera-hzdr/gpu_picongpu.profile.example
+++ b/etc/picongpu/hemera-hzdr/gpu_picongpu.profile.example
@@ -43,7 +43,7 @@ module load pngwriter/0.7.0
 
 export PICSRC=$HOME/src/picongpu
 export PIC_EXAMPLES=$PICSRC/share/picongpu/examples
-export PIC_BACKEND="cuda:70"
+export PIC_BACKEND="cuda:60"
 
 # Path to the required templates of the system,
 # relative to the PIConGPU source code of the tool bin/pic-create.
--- a/etc/picongpu/hemera-hzdr/gpu_picongpu.profile.example
+++ b/etc/picongpu/hemera-hzdr/gpu_picongpu.profile.example
@@ -17,24 +17,24 @@ export MY_NAME="$(whoami) <$MY_MAIL>"
 #
 module purge
 module load git
-module load gcc/11.2.0
-module load cmake/3.26.1
-module load cuda/11.5
-module load openmpi/4.1.1-cuda115
-module load boost/1.78.0
+module load gcc/12.2.0
+module load cuda/12.1
+module load libfabric/1.17.0
+module load ucx/1.14.0-gdr
+module load openmpi/4.1.5-cuda121-gdr
+module load python/3.10.4
+module load boost/1.82.0
 
 # Other Software ##############################################################
 #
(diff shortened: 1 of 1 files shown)


BrianMarre:
@MRZirper the change of cuda backend from 70 to 60 is not a good idea. the gpu queue contains some nodes with p100 gpus and these only support cuda 6.0 (corresponding to `PIC_BACKEND="cuda:60"`)

In contrast to the v100 gpus which support cuda 7.0

MRZirper:
Thanks, @BrianMarre I will change it. The Rest is ok like this? I am not an expert with all this stuff and just copy pasted ^^`

BrianMarre:
The rest looks fine, Just compile and run the LaserWakefield example on hemera using the new profile to be sure ;)

MRZirper:
Worked fine :)

pordyna:
This is actually still outdated. We should switch both fwkt_v100 and the gpu queue to load those modules:
```
Currently Loaded Modules:
 1) gcc/12.2.0                  5) libfabric/1.17.0   9) boost/1.82.0     13) pngwriter/0.7.0
 2) cuda/12.1                   6) cmake/3.26.1      10) zlib/1.2.11      14) hdf5-parallel/1.12.0-omp415-cuda121
 3) openmpi/4.1.5-cuda121-gdr   7) git/2.37.1        11) c-blosc2/2.31.1  15) adios2/2.9.2-cuda121-blosc2
 4) ucx/1.14.0-gdr              8) python/3.10.4     12) libpng/1.6.39    16) openpmd/0.15.2-cuda121-blosc2
```

This has been installed recently by the IT and includes a newer version of `blosc.` I don't think the old `blosc` will work with the new adios anyway. (This is only important for compression)

psychocoderHPC:
> This is actually still outdated. We should switch both fwkt_v100 and the gpu queue to load those modules:
> 
> ```
> Currently Loaded Modules:
>  1) gcc/12.2.0                  5) libfabric/1.17.0   9) boost/1.82.0     13) pngwriter/0.7.0
>  2) cuda/12.1                   6) cmake/3.26.1      10) zlib/1.2.11      14) hdf5-parallel/1.12.0-omp415-cuda121
>  3) openmpi/4.1.5-cuda121-gdr   7) git/2.37.1        11) c-blosc2/2.31.1  15) adios2/2.9.2-cuda121-blosc2
>  4) ucx/1.14.0-gdr              8) python/3.10.4     12) libpng/1.6.39    16) openpmd/0.15.2-cuda121-blosc2
> ```
> 
> This has been installed recently by the IT and includes a newer version of `blosc.` I don't think the old `blosc` will work with the new adios anyway. (This is only important for compression)

@pordyna is right the openPMD-api module loaded in this PR will not provide blosc support. Blosc1 is not compatible to the latest adios. The IT was not aware of it and provided the new packages.

Nevertheless, I will merge this PR. The correct modules for blosc2 can be applied in a separate PR. This requires additional testing because maybe the openmpi or ucx module is not compatible to this queue.

psychocoderHPC (approved):

PR #4849, opened 2024-03-19
change nameing of methods in Buffer

psychocoderHPC:
- change the naming for getting the size of an buffer to be near to the C++ STL.
- for N-dimensional values use the postfix `ND`

### DIFF
--- a/include/pmacc/memory/buffers/DeviceBuffer.hpp
+++ b/include/pmacc/memory/buffers/DeviceBuffer.hpp
@@ -71,17 +71,17 @@ namespace pmacc
 
         BufferType1D as1DBuffer()
         {
-            auto currentSize = this->getCurrentSize();
+            auto numElements = this->size();
             eventSystem::startOperation(ITask::TASK_DEVICE);
             return BufferType1D(
                 alpaka::getPtrNative(*view),
                 alpaka::getDev(*devBuffer),
-                MemSpace<DIM1>(currentSize).toAlpakaMemVec());
+                MemSpace<DIM1>(numElements).toAlpakaMemVec());
         }
 
         BufferType1D as1DBufferNElem(size_t const numElements)
         {
-            PMACC_ASSERT(numElements < this->getCurrentSize());
+            PMACC_ASSERT(numElements < this->size());
             eventSystem::startOperation(ITask::TASK_DEVICE);
             return BufferType1D(
                 alpaka::getPtrNative(*view),
@@ -122,7 +122,7 @@ namespace pmacc
             {
                 createSizeOnDeviceBuffers();
             }
-            this->setCurrentSize(size.productOfComponents());
+            this->setSize(size.productOfComponents());
             this->isMemoryContiguous = true;
             reset(false);
         }
@@ -154,7 +154,7 @@ namespace pmacc
             {
                 createSizeOnDeviceBuffers();
             }
-            this->setCurrentSize(size.productOfComponents());
+            this->setSize(size.productOfComponents());
             this->isMemoryContiguous = T_dim == DIM1;
             reset(true);
         }
@@ -166,7 +166,7 @@ namespace pmacc
 
         void reset(bool preserveData = true) override
         {
-            this->setCurrentSize(Buffer<T_Type, T_dim>::capacityND().productOfComponents());
+            this->setSize(Buffer<T_Type, T_dim>::capacityND().productOfComponents());
 
             eventSystem::startOperation(ITask::TASK_DEVICE);
             if(!preserveData)
@@ -205,7 +205,7 @@ namespace pmacc
          *
          * @return device side current size buffer
          */
-        CurrentSizeBufferDevice getCurrentSizeOnDeviceBuffer()
+        CurrentSizeBufferDevice sizeOnDeviceBuffer()
         {
             eventSystem::startOperation(ITask::TASK_DEVICE);
             if(!hasCurrentSizeOnDevice())
@@ -219,13 +219,13 @@ namespace pmacc
          *
          * @return host side current size buffer
          */
-        typename Buffer<T_Type, T_dim>::CurrentSizeBufferHost getCurrentSizeHostSideBuffer()
+        typename Buffer<T_Type, T_dim>::CurrentSizeBufferHost sizeHostSideBuffer()
         {
             eventSystem::startOperation(ITask::TASK_HOST);
             return this->currentSizeBufferHost;
         }
 
-        size_t getCurrentSize() override
+        size_t size() override
         {
             if(hasCurrentSizeOnDevice())
             {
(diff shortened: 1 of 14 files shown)


include/pmacc/memory/buffers/DeviceBuffer.hpp
```diff
@@ -71,7 +71,7 @@ namespace pmacc
 
         BufferType1D as1DBuffer()
         {
-            auto currentSize = this->getCurrentSize();
+            auto currentSize = this->size();
```
ikbuibui:
```suggestion
            auto numElements = this->size();
```
For consistency

include/pmacc/particles/memory/boxes/ExchangePushDataBox.hpp
```diff
@@ -42,12 +42,12 @@ namespace pmacc
 
         HDINLINE ExchangePushDataBox(
             VALUE* data,
-            TYPE* currentSizePointer,
+            TYPE* ptrToSize,
```
ikbuibui: Is it required or a good idea to call pointers to objects as ptrToObj? is it not evident from the type

ikbuibui (changes requested):

include/pmacc/particles/memory/boxes/ExchangePushDataBox.hpp
psychocoderHPC: good question, I do not know.

psychocoderHPC: I changed it to `particleCount`

include/pmacc/particles/memory/boxes/ExchangePushDataBox.hpp
```diff
@@ -42,12 +42,12 @@ namespace pmacc
...
         HDINLINE ExchangePushDataBox(
             VALUE* data,
-            TYPE* currentSizePointer,
+            TYPE* particleCount,
             TYPE maxSize,
             PushDataBox<TYPE, PushType> virtualMemory)
             : DataBox<PitchedBox<VALUE, DIM1>>(PitchedBox<VALUE, DIM1>(data))
             , virtualMemory(virtualMemory)
-            , currentSizePointer(currentSizePointer)
+            , parCount(particleCount)
```
chillenzer: Good renaming. I can see how you tried to use a different name here but why not simply call the internal variable `particleCount`, too? Would be much nicer to read later on and doesn't suffer from translation problems like me thinking of "parameter count".

chillenzer (approved):
Hi, sounds good to me. I like getting closer to STL.

My only pertinent question is: Why did you stop half way on the way to STL? E.g. `setSize` in STL jargon would be `resize`, wouldn't it?

Furthermore, I've got a few questions from looking at the code that are not immediately related to your changes:
* There are two functions `sizeDeviceSideBuffer` and `sizeOnDeviceBuffer`. From what I see, one is a checked version the other an unchecked much like the difference between `.at()` and `operator[]` for `std::vector`. First: Is that necessary and the distinction used somewhere (my simple `grep` says no)? Second: If so, couldn't we find a more expressive name?
* The comment above `sizeND` in `Buffer.hpp` l. 128 seems incorrect or at least easily misunderstood. Shall we change this?

include/pmacc/particles/memory/boxes/ExchangePushDataBox.hpp
psychocoderHPC: This would be throw warning that we have shadowing the internal variable. This is also a common source of mistakes.
Typical we use `m_*` for internal variables to avoid it. And use a nice name for the user interface.

chillenzer: Yes, I'm aware of this but given that we are dealing with an empty constructor this trade-off might be acceptable.

psychocoderHPC: I will change it to `m_particleCount` and also the next line where we have the shadowing issue.

ikbuibui (dismissed):

include/pmacc/particles/memory/boxes/ExchangePushDataBox.hpp
```diff
@@ -40,15 +40,22 @@ namespace pmacc
     public:
         using PushType = ExchangeMemoryIndex<TYPE, DIM>;
 
+        /**
+         *
+         * @param data particle data storage
+         * @param particleCount current fill level of data
+         * @param maxSize max capacity of data
+         * @param indexTabel index tabel which holds meta data to link particles in data to supercells
```
chillenzer:
```suggestion
         * @param indexTable index table which holds meta data to link particles in data to supercells
```
Of course, change everywhere.

include/pmacc/particles/memory/boxes/ExchangePushDataBox.hpp
```diff
@@ -74,21 +81,20 @@ namespace pmacc
             DataSpace<DIM> const& superCell,
             T_Hierarchy const& hierarchy)
         {
-            TYPE oldSize
-                = alpaka::atomicAdd(worker.getAcc(), currentSizePointer, count, hierarchy); // get count VALUEs
+            TYPE oldSize = alpaka::atomicAdd(worker.getAcc(), m_particleCount, count, hierarchy); // get count VALUEs
```
chillenzer:
```suggestion
            // get count VALUEs
            TYPE oldSize = alpaka::atomicAdd(worker.getAcc(), m_particleCount, count, hierarchy);
```
You know... once upon a time a wise old person taught me an important lesson...

include/pmacc/particles/memory/boxes/ExchangePushDataBox.hpp
```diff
@@ -74,21 +81,20 @@ namespace pmacc
...
 
-            if(oldSize + count > maxSize)
+            if(oldSize + count > m_maxSize)
             {
-                alpaka::atomicExch(worker.getAcc(), currentSizePointer, maxSize, hierarchy); // reset size to maxsize
-                if(oldSize >= maxSize)
+                alpaka::atomicExch(worker.getAcc(), m_particleCount, m_maxSize, hierarchy); // reset size to maxsize
```
chillenzer: Same here.

chillenzer (changes requested):

include/pmacc/particles/memory/boxes/ExchangePushDataBox.hpp
psychocoderHPC: `... old ...`? :older_man:

ikbuibui (dismissed):

chillenzer (approved):
Sounds good. I think we can ignore the failing CI (at least in pypicongpu) due to the TypeGuard error.

ikbuibui:
@psychocoderHPC could you please deal with the merge conflicts

ikbuibui (approved):

PR #4852, opened 2024-03-25
Fix default value for `--no-start-simulation`

chillenzer:
Hi,
there was a huge bug in #4812! We have a wrong default value for `--no-start-simulation`. The fix is easy and obvious.

But a more pressing matter: How is it possible that our CI can pass if there is not a single simulation ever running? Feel free to discuss!

### DIFF
--- a/include/picongpu/simulation/control/Simulation.hpp
+++ b/include/picongpu/simulation/control/Simulation.hpp
@@ -122,7 +122,7 @@ namespace picongpu
             desc.add_options()(
                 "versionOnce", po::value<bool>(&showVersionOnce)->zero_tokens(),
                 "print version information once and start")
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
I will merge this PR before the CI passed. The bug should be removed from the dev branch as soon as possible.

ikbuibui:
I didn't look too closely but since `skipSimulation` is initialized to false maybe this wasn't actually triggered?

psychocoderHPC:
> But a more pressing matter: How is it possible that our CI can pass if there is not a single simulation ever running? Feel free to discuss!

We do not run runtime tests in our CI. The problem is that all runners are shared shared.
To run runtime tests someone need to check all `ci.sh` and set `-r` cmd parameter that we do not use the full GPU memory and someone needs to integrate everything into our CI run scripts.

chillenzer:
Maybe we should at least run one single time just so we can catch such huge things as "It doesn't run at all under any circumstances."?

PR #4855, opened 2024-03-25
fix CI pypicongpu tests

### LINKED ISSUE #4853: pypicongpu tests broken
Our pypicongpu tests are broken and blocking now other PRs. https://gitlab.com/hzdr/crp/picongpu/-/jobs/6467970644

@BrianMarre COuld you please have a look into pypicongpu tests.

psychocoderHPC:
fix #4853 

Change supported typeguard versions to workaround the current CI issue.

Use typeguard `typeguard >= 4.1.3, < 4.1.5`

Equally to #4853 but try to change the required typeguard version until someone knows how to fix the error.

### DIFF
--- a/lib/python/picongpu/requirements.txt
+++ b/lib/python/picongpu/requirements.txt
@@ -1,4 +1,4 @@
-typeguard >= 4.1.3
+typeguard >= 4.1.3, <= 4.1.5
 sympy >= 1.9
 chevron >= 0.13.1
 jsonschema == 4.17.3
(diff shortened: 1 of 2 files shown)


BrianMarre:
@psychocoderHPC the CI fails since no typeguard version remains after the requiremtns and testing filters have been applied. In order for this change to also work you need to also update this line [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/68507d98500a2f3a8e3b2312503550b60ae61314/share/ci/pypicongpu_generator.py#L293) to read `get_all_pypi_versions` instead of `get_all_major_pypi_versions`

Currently we get only the latest version for each supported mayor version from pypi and then filter them using the requirements.txt settings, with your new setting, after both filters have been applied no valid version remains.

ikbuibui (dismissed):
Typeguard 4.2.0 fails in CI

psychocoderHPC:
I will merge the change with my admin power than other PRs can be rebased.

@BrianMarre will look into support for new typeguard versions.

PR #4863, opened 2024-04-04
Implement Log Axes for the binning plugin

ikbuibui:
Added ability to define log axes in the binning plugin
Refactored how the normalize by bin volume works so as to not pass an unnecessary potentially heavy object around


include/picongpu/plugins/binning/axis/LogAxis.hpp
```diff
@@ -0,0 +1,287 @@
...
+
+                        static_assert(
+                            std::is_same<decltype(val), decltype(logMin)>::value,
+                            "The return type of the axisAttributeFunctor should be the same as the type of Axis "
+                            "min/max ");
+
+                        uint32_t binIdx = 0;
+                        bool enableBinning = overflowEnabled; // @todo check if disableBinning is better
+
+                        if(0. < val)
```
psychocoderHPC: please cast `0.` to the value type to avoid casting to double, IMO you can use the template parameter `T`

include/picongpu/plugins/binning/axis/LogAxis.hpp
```diff
@@ -0,0 +1,287 @@
...
+                        {
+                            auto logVal = math::log2(val);
```
psychocoderHPC, line 136: for values < 1 `logVal` will be negative if I remember correctly you said that you must handle it in your implementation?

include/picongpu/plugins/binning/axis/LogAxis.hpp
```diff
@@ -0,0 +1,287 @@
...
+                        edge = next_edge;
+                    }
+                }
+
+                void initLAK()
+                {
+                    // do conversion to PIC units here, if not auto
+                    // auto picRange = toPICUnits(axSplit.range, units);
+                    auto min = toPICUnits(axisSplit.m_range.min, units);
+                    PMACC_ASSERT(0. < min);
```
psychocoderHPC: Maybe `PMACC_VERIFY` is better because than it will always be checked.

psychocoderHPC (changes requested):

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: @BrianMarre You were handling values from 0-1 as a special case, but I am not sure about the reason for this. I think we discussed that you deal with 0-1 separately to avoid a case where most log bins are in the 0-1 value range . From my point of view, we should do what the user asks, and if they want log bins with a minimum value which is less that 1 we should let them. But I am not sure what would be expected/wanted behavior. Also pinging @steindev @PrometheusPi for input

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: done

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: done

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: So this special handling will be moved to a different log-lin-log axis which has a log scales on either side of zero except for a user definable linear region around zero

include/picongpu/plugins/binning/axis/LinearAxis.hpp
```diff
@@ -0,0 +1,256 @@
...
+            public:
+                using T = T_Attribute;
+                using ScalingType = std::
+                    conditional_t<std::is_integral_v<T>, std::conditional_t<sizeof(T) == 4, float_X, double>, T>;
+
+                AxisSplitting<T_Attribute> axisSplit;
+                /** Axis name, written out to OpenPMD */
+                std::string label;
+                /** Units(Dimensionality) of the axis */
+                std::array<double, 7> units;
```
chillenzer: Give a name to the magic number `7`.

include/picongpu/plugins/binning/axis/LinearAxis.hpp
```diff
@@ -0,0 +1,256 @@
...
+             * Bins are closed open [) intervals [min, min + size), [min + size, min + 2*size) ,..., [max-size, max).
+             * Allocates 2 extra bins, for under and overflow. These are bin index 0 and (n_bins+2)-1
+             */
+            template<typename T_Attribute, typename T_AttrFunctor>
+            class LinearAxis
+            {
+            public:
+                using T = T_Attribute;
+                using ScalingType = std::
+                    conditional_t<std::is_integral_v<T>, std::conditional_t<sizeof(T) == 4, float_X, double>, T>;
```
chillenzer: What does this line tell me? If `T` is integral, I want something floating point. But if `T` is not 32-bit, say `char`, I definitely want the built-in double and not our custom define `float_64`? I think this needs an explaining comment at the very least.

include/picongpu/plugins/binning/axis/LinearAxis.hpp
```diff
@@ -0,0 +1,256 @@
...
+             * Linear axis with contiguous fixed sized bins.
+             * Axis splitting is defined with min, max and n_bins. Bin size = (max-min)/n_bins.
+             * Bins are closed open [) intervals [min, min + size), [min + size, min + 2*size) ,..., [max-size, max).
+             * Allocates 2 extra bins, for under and overflow. These are bin index 0 and (n_bins+2)-1
+             */
+            template<typename T_Attribute, typename T_AttrFunctor>
+            class LinearAxis
+            {
+            public:
+                using T = T_Attribute;
```
chillenzer: I don't quite see why you would want an even shorter name for this, but if you do, I would strongly advise to stick with one name (and never use `T_Attribute`). Unless, of course, they happen to be semantically distinct and only coincidentally equal in this implementation. If so, a name more expressive than `T` would be nice.

include/picongpu/plugins/binning/axis/LinearAxis.hpp
```diff
@@ -0,0 +1,256 @@
...
+                using ScalingType = std::
+                    conditional_t<std::is_integral_v<T>, std::conditional_t<sizeof(T) == 4, float_X, double>, T>;
+
+                AxisSplitting<T_Attribute> axisSplit;
+                /** Axis name, written out to OpenPMD */
+                std::string label;
+                /** Units(Dimensionality) of the axis */
+                std::array<double, 7> units;
+                /**
+                 * @TODO store edges? Copmute once at the beginning and store for later to print at every iteration,
```
chillenzer:
```suggestion
                 * @TODO store edges? Compute once at the beginning and store for later to print at every iteration,
```

include/picongpu/plugins/binning/axis/LinearAxis.hpp
```diff
@@ -0,0 +1,256 @@
...
+                 * also to be used in search based binning
+                 */
+                std::vector<double> binEdges;
```
chillenzer: So, does this actually now store edges? If not, remove!

include/picongpu/plugins/binning/axis/LinearAxis.hpp
```diff
@@ -0,0 +1,256 @@
...
+                struct LinearAxisKernel
+                {
+                    /** Function to place particle on axis, returns same type as min and max */
```
chillenzer, line 69: If there's such a restriction, should we enforce that at compile-time? Also, is that really the restriction or is `is_convertible` enough?

include/picongpu/plugins/binning/axis/LinearAxis.hpp
```diff
@@ -0,0 +1,256 @@
...
+                    constexpr LinearAxisKernel(T_AttrFunctor attrFunc, bool enableOverflowBins)
+                        : getAttributeValue{attrFunc}
+                        , overflowEnabled{enableOverflowBins}
+                    {
+                    }
+
+                    /**
+                     * @param n_bins User requested n_bins, actual nBins is n_bins + 2 for overflow and underflow bins
+                     */
+                    void initAxisSplit(T minR, T maxR, uint32_t n_bins, ScalingType scalingR)
```
chillenzer: Why is this not RAII?

include/picongpu/plugins/binning/axis/LinearAxis.hpp
chillenzer: Okay. I see that there's a `static_assert` later on. Might actually produce an easier to read error message than some SFINAE failure but communicates less about this contract to the user. Not sure what's better.

include/picongpu/plugins/binning/axis/LinearAxis.hpp
```diff
@@ -0,0 +1,256 @@
...
+                        validIdx = validIdx && enableBinning;
+                        return binIdx;
+                    }
+                };
+
+                LinearAxisKernel lAK;
+
+                struct BinWidthKernel
+                {
+                    ScalingType scaling;
```
chillenzer, line 140: Initialise.

include/picongpu/plugins/binning/axis/LinearAxis.hpp
```diff
@@ -0,0 +1,256 @@
...
+                    uint32_t nBins;
```
chillenzer, line 141: Initialise.

include/picongpu/plugins/binning/axis/LinearAxis.hpp
```diff
@@ -0,0 +1,256 @@
...
+
+                    BinWidthKernel()
```
chillenzer:
```suggestion
                    BinWidthKernel() = default;
```

But do we really need an empty constructor? `BinWidthKernel` is not in a usable state afterwards due to un-initialised memory. But even if we fix that, does it make sense, to have `scaling` and `nBins` on some (likely mismatching with other parts) default value?

include/picongpu/plugins/binning/axis/LinearAxis.hpp
```diff
@@ -0,0 +1,256 @@
...
+                    T_AttrFunctor attrFunctor,
+                    std::string label,
+                    std::array<double, 7> units) // add type T to the default label string
+                    : axisSplit{axSplit}
+                    , label{label}
+                    , units{units}
+                    , lAK{attrFunctor, axisSplit.enableOverflowBins}
+                {
+                    initLAK();
+                    bWK = BinWidthKernel(lAK);
```
chillenzer: After `LinearAxisKernel` is RAII, move this into the member initialiser list.

include/picongpu/plugins/binning/axis/LinearAxis.hpp
```diff
@@ -0,0 +1,256 @@
...
+                }
+
+                constexpr uint32_t getNBins() const
+                {
+                    return lAK.nBins;
+                }
+
+                double getUnitConversion() const
+                {
+                    return get_conversion_factor(units);
```
chillenzer: This function name does not abide by our `camelCase` naming convention.

include/picongpu/plugins/binning/axis/LogAxis.hpp
```diff
@@ -0,0 +1,287 @@
...
+namespace picongpu
+{
+    namespace plugins::binning
+    {
+        namespace axis
+        {
+            /**
+             * Log axis with logarithmically sized bins.
+             * Axis splitting is defined with min, max and n_bins.
+             * Beware of using a log axis with an integral range/axis splitting
```
chillenzer: If it's that bad just disable it completely?

include/picongpu/plugins/binning/axis/LogAxis.hpp
```diff
@@ -0,0 +1,287 @@
...
+             */
+            template<typename T_Attribute, typename T_AttrFunctor>
+            class LogAxis
+            {
+            public:
+                using T = T_Attribute;
+
+                AxisSplitting<T_Attribute> axisSplit;
+                /** Axis name, written out to OpenPMD */
+                const std::string label;
```
chillenzer: [Corresponding core guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-constref). If you think, you've got a reason to violate this, please add an explanatory comment.

include/picongpu/plugins/binning/axis/LogAxis.hpp
```diff
@@ -0,0 +1,287 @@
...
+                    T_AttrFunctor getAttributeValue;
+                    /**
+                     * logMin and logMax values in the range of the binning. Values outside this range are
+                     * placed in overflow bins
+                     */
+                    T logMin, logMax;
+                    /** Number of bins in range */
+                    uint32_t nBins;
+                    /** Using type depending on whether T is integer or floating point type to avoid precision loss */
+                    using ScalingType = std::
```
chillenzer: Why was that in the encapsulating class for `LinearAxis` and is in the kernel in `LogAxis`? Make consistent.

include/picongpu/plugins/binning/axis/LogAxis.hpp
```diff
@@ -0,0 +1,287 @@
...
+                        conditional_t<std::is_integral_v<T>, std::conditional_t<sizeof(T) == 4, float_X, double>, T>;
```
chillenzer: If this is a general thing, I'd suggest to abstract that away into a (meta-)function with an appropriate name.

include/picongpu/plugins/binning/axis/LogAxis.hpp
```diff
@@ -0,0 +1,287 @@
...
+                    {
+                        logMin = logMinR;
+                        logMax = logMaxR;
+                        nBins = n_bins;
+                        scaling = scalingR;
+                    }
+
+
+                    template<typename T_Worker, typename T_Particle>
+                    ALPAKA_FN_ACC uint32_t getBinIdx(
```
chillenzer, line 117: Is this function completely copy-paste? Could we remove this code duplication?

include/picongpu/plugins/binning/axis/LogAxis.hpp
```diff
@@ -0,0 +1,287 @@
...
+                const std::string label;
+                /** Units(Dimensionality) of the axis */
+                const std::array<double, 7> units;
+                std::vector<T> binWidths;
+
+                /**
+                 * @TODO store edges? Copmute once at the beginning and store for later to print at every
+                 * iteration, also to be used in search based binning
+                 */
+                struct LogAxisKernel
```
chillenzer, line 75: Is this struct different from `LinearAxisKernel` in any way?

include/picongpu/plugins/binning/axis/LogAxis.hpp
```diff
@@ -0,0 +1,287 @@
...
+                    std::string label,
+                    std::array<double, 7> units) // add type T to the default label string
+                    : axisSplit{axSplit}
+                    , label{label}
+                    , units{units}
+                    , lAK{attrFunctor, axisSplit.enableOverflowBins}
+                {
+                    initLAK();
+
+                    binWidths.reserve(axisSplit.nBins);
```
chillenzer, line 197: Extract the rest into a function -> "no raw loops".

include/picongpu/plugins/binning/axis/LogAxis.hpp
```diff
@@ -0,0 +1,287 @@
...
+
+
+                LogAxisKernel getAxisKernel() const
+                {
+                    return lAK;
+                }
+
+                BinWidthKernel getBinWidthKernel()
+                {
+                    // reset whenever binWidths changez
```
chillenzer: Does this happen and who's supposed to do that? This variable is only available inside the scope of this function and I don't see a statement that does so.

chillenzer:
```suggestion
                    // reset whenever binWidths changes
```

include/picongpu/plugins/binning/axis/LogAxis.hpp
```diff
@@ -0,0 +1,287 @@
...
+
+            /**
+             * @details Creates a log Axis
+             * @tparam T_Attribute Type of the deposition functor (This is also the type of min, max and return type of
+             * the attrFunctor and if these types dont match this will throw an error)
+             * @param axisSplitting
+             * @param functorDescription
+             */
+            template<typename T_Attribute, typename T_FunctorDescription>
+            HINLINE auto createLog(AxisSplitting<T_Attribute> axSplit, T_FunctorDescription functorDesc)
```
chillenzer, line 278: This could be misleading, a "log" without context could be interpreted with the meaning of "logging". Maybe we should add "Axis" to both factory functions.

chillenzer (changes requested):
Hi, got a few mostly stylistic things but I'll come over and discuss

chillenzer:
From offline discussion: We should add a few words about the design philosophy somewhere (probably dev docs).

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: Almost but not quite, this version calculates log of the value on line 119

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: The thought process behind the naming was that this function will be called with the leading namespace, like ``axis::createLog(...)``, so it would be reasonably clear. 
But I am open to your suggestion if you still feel it is best to append Axis to the funcion names

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: Currently only a placeholder. This will only happen once runtime axis resizing is implemented, then this set flag will be moved out and be invalidated when the axis is resized.

psychocoderHPC:
@ikbuibui could you please finish this PR

include/picongpu/plugins/binning/axis/LinearAxis.hpp
ikbuibui: Added a comment

include/picongpu/plugins/binning/axis/LinearAxis.hpp
ikbuibui: An axis is basically a range over a domain of a certain type. This type seems quite natural to be defined as "the type" for an axis, hence the use of ``T``. The idea is to make this type available as ``Axis::T``. Maybe I should rename it to ``type``. I am chagning some occourances of ``T_Attribute`` to ``T`` to stick to one of them.

include/picongpu/plugins/binning/axis/LinearAxis.hpp
ikbuibui: Added, calling it num_units

include/picongpu/plugins/binning/axis/LinearAxis.hpp
ikbuibui: Done

include/picongpu/plugins/binning/axis/LinearAxis.hpp
ikbuibui: Fixed

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: Done

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: I have moved it out in the LogAxis to maintain consistency as requested.

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: Yes, the getBinIDx method is different

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: If this comes up again, I will consider it.

include/picongpu/plugins/binning/axis/LinearAxis.hpp
ikbuibui: This was actually a bug, fixed. The edges are not currently stored. This can change in the future if needed.

include/picongpu/plugins/binning/axis/LinearAxis.hpp
ikbuibui: Silent conversions may lead to loss of precision, so currently enforcing that they really are the same.

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: I'm not sure about this, emplacing bin widths into a vector doesn't seem to me to be a reusable function here. Bin width calculation should only really be done in the constructor and really should never be called again. It isnt some algorithm to be abstracted out. If you have some good reason for why this should be its own function I'd be happy to change it

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: Removed the const. I appreciate the link

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: I improved the usability of integral log axes, and removed the warning

ikbuibui: Added a few changes to make the LogAxis more usable with integers. Changed the type of ``logMin`` and ``logMax`` to ``ScalingType``. Bin widths for integral log axes will still be integral, and the ceiling of the actual point of the logarithmic division will be used as the end of the bin

include/picongpu/plugins/binning/axis/LinearAxis.hpp
ikbuibui: As discussed offline, some design decisions were made to keeping in mind the feature to set the min and max automatically. Since this feature doesnt really exist yet it doesnt make sense to keep workarounds in the code currently. So this is now RAII.

include/picongpu/plugins/binning/axis/LinearAxis.hpp
ikbuibui: Done

include/picongpu/plugins/binning/axis/LinearAxis.hpp
ikbuibui: Removed the empty constructor

include/picongpu/plugins/binning/axis/LinearAxis.hpp
ikbuibui: Done

psychocoderHPC:
> From offline discussion: We should add a few words about the design philosophy somewhere (probably dev docs).

Do you mean for the binning plugin or the log axes implementation?

include/picongpu/plugins/binning/axis/LinearAxis.hpp
psychocoderHPC: Yes if you like to expose the name public you should use `type` or `*Type`.
The second is often used in PMacc and `type` is typically used only for traits.

include/picongpu/plugins/binning/axis/LinearAxis.hpp
psychocoderHPC: > `To avoid loss of precision, ..`

Float can only store 23bit without permission loss so we need to change it to `sizeof(T) <= 2` or the comment is wrong.

The comment is talking about `<4 byte` but the code is testing for exact 4 byte only.
The `ScalingType` derivation should be corrected.

psychocoderHPC (changes requested):

include/picongpu/plugins/binning/axis/LogAxis.hpp
chillenzer: Conceptual perspective: It's not about reusability but [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle) and [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_Concerns). The single responsibility of this function is "composition" and not the separate concern of how the bin widths are initialised in detail. Also, you gain one more name and names are important means of communication.

Practical perspective: I've seen quite a few 1000-line constructors (or even worse `init()` methods) that started out like this and ended up as a huge mess of interdependent `for` loops that broke every other aspect of the code as soon as you wanted to change a single thing. If the function ends up looking like this:
```C++
LogAxis(...) {
    initLAK();
    initBinWidths();
    initThisOtherThingThatWasAddedLater();
    andOfCourseModifyBinWidthsDueToThisNewThing();
    nowEnsureConsistency();
    makeSomeLastTweaks();
}
```
you can tell quite clearly what is going on without ever looking at any of the function definitions. If I would have given you the same code as a large bowl of `for`-loop spaghetti that reuse variables and names and nest concerns, that code would probably be impossible to understand (let alone change) later on.

include/picongpu/plugins/binning/axis/LogAxis.hpp
chillenzer: Okay. I see. That is sufficient I'd say. Let's hope that nobody does `using namespace ...::axis;`.

chillenzer:
> > From offline discussion: We should add a few words about the design philosophy somewhere (probably dev docs).
> 
> Do you mean for the binning plugin or the log axes implementation?

You are right, it's rather a general thing about the binning plugin. Some design choices and naming were obvious and reasonable only after I had talked to @ikbuibui about this, so I thought that could be documented to spare us the trouble next time. But not necessarily something that's related to this PR.

include/picongpu/plugins/binning/axis/LogAxis.hpp
chillenzer: See comments in `getBinIdx`.

include/picongpu/plugins/binning/axis/LogAxis.hpp
chillenzer: Okay. So, I would suggest to split this function up:
```C++
auto computeValue(...) {
            return math::log2(getAttributeValue(domainInfo, worker, particle));
}

auto computeBinIdx(const auto val) {
           // all the actual binning logic...
}
           
auto updateValidIdx(const auto idx, bool& validIdx) {
           // This formula might be off by one: You're the expert.
           validIdx = validIdx && (overflowEnabled || idx < nBins);
}
          
auto getBinIdx(...) {
           auto val = computeValue(...);
           auto idx = computeBinIdx(val);
           updateValidIdx(idx, validIdx);
           return idx;
}
```
Then, the `LinearAxisKernel` and `LogAxisKernel` are identical up to `computeValue`. So, I would extract that into one struct and use some form of polymorphism (inheritance or something static) to adjust `computeValue` accordingly.

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: Done

include/picongpu/plugins/binning/axis/LinearAxis.hpp
ikbuibui: Using Type

include/picongpu/plugins/binning/axis/LinearAxis.hpp
ikbuibui: Fixed comment and code. Now we check if the integer point type is <= 4 bytes and if so use a float else a double. This is a sensible value, but not explicitly to ensure loss of precision.

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: Yes, I agree some polymorphism can be used in this case as the LogAxis is doing the linear binning of log values. I would move this refactoring to a future PR when I implement a different (maybe a search based) axis and have a better idea of the more general components of these structs.

psychocoderHPC (approved):

include/picongpu/plugins/binning/axis/LogAxis.hpp
chillenzer: Okay.

include/picongpu/plugins/binning/axis/LinearAxis.hpp
chillenzer: Okay.

chillenzer (approved):

psychocoderHPC (approved):

PR #4871, opened 2024-04-17
PMacc: rename Stream into Queue

psychocoderHPC:
Our naming still based on CUDA namings, this PR is updating the naming to fit with the alpaka object naming.


psychocoderHPC:
replaced by  #4873 to avoid conflicts with other similar renamings

PR #4873, opened 2024-04-17
update deprecated naming from old CUDA times 

psychocoderHPC:
- rename `Stream` into `Queue`
- rename `AccStream` to `ComputeDeviceQueue`
- rename `AlpakaEventType` to `ComputeDeviceEvent`
- rename `CudaEventHandle` to `ComputeEventHandle`
- rename `CudaEvent` to `ComputeEvent`

This PR is not changing the logic it is only renaming things.


ikbuibui (approved):
Looks good, I think the names fit the concepts well

PR #4874, opened 2024-04-19
openPMD: add parameter `particleIOChunkSize`

psychocoderHPC:
# PIConGPU 
Allow defining a chunk size which is used to write particles with multiple write calls instead of a single.
The benefits are that the host memory footprint to write particles with the Adios2 bp5 engine is drastically reduced.
For systems like ORNL Frontier where the amount of GPU memory is equal to the host memory, we can now fill the full GPU memory with particles without running out of memory during the IO.

By default `particleIOChunkSize` is set to 10 GiB which is a reasonable value that our data chunks are not too small but typically fit for a system with a low amount of host memory.

Field IO is not affected by this feature.

# PMacc add `RangeMapper`
 - Add a mapper wrapper to map any PMacc mapper to a one dimensional index
    domain. The mapper supports addressing a sub index range of the wrapped
    mapping domain.

- [x] ~~run test on frontier~~ Currently it is not possible to get a job on Frontier. Never the less restarting is not possible because we have the same memory issue when restarting a simulation. This will be addressed in a separate PR.
- [x] hdf5/bp4 and bp5 test on our dev server (restart from a chunked checkpoint)

### DIFF
--- a/docs/source/usage/plugins/openPMD.rst
+++ b/docs/source/usage/plugins/openPMD.rst
@@ -134,6 +134,12 @@ PIConGPU command line option          description
 ``--checkpoint.openPMD.jsonRestart``  Set backend-specific parameters for openPMD backends in JSON format for restarting from a checkpoint.
 ``--openPMD.dataPreparationStrategy`` Strategy for preparation of particle data ('doubleBuffer' or 'mappedMemory'). Aliases 'adios' and 'hdf5' may be used respectively.
 ``--openPMD.toml``                    Alternatively configure the openPMD plugin via a TOML file (see below).
+``--openPMD.particleIOChunkSize``     Particle data will be written in chunks of the given size. unit: MiB
+                                      The memory footprint on the host side is reduced compared to writing all particles with one write call.
+                                      Currently the memory footprint is only reduced bp5 is used, HDF5 and bp4 does not support partial data dump.
+                                      If your system host memory is small compared to the accelerator memory you should use ``--openPMD.dataPreparationStrategy mappedMemory`` to have
+                                      a smaller host memory footprint.
+                                      The memory footprint required to dump fields will not be affected by this parameter.
 ===================================== ====================================================================================================================================================
 
 .. note::
@@ -254,6 +260,7 @@ As soon as the openPMD plugin is compiled in, one extra ``mallocMC`` heap for th
 During I/O, particle attributes are allocated one after another.
 Using ``--openPMD.dataPreparationStrategy doubleBuffer`` (default) will require at least 2x the GPU memory on the host side.
 For a smaller host side memory footprint (<< GPU main memory) pick ``--openPMD.dataPreparationStrategy mappedMemory``.
+By using additional to the mapped memory strategy ``--openPMD.particleIOChunkSize N`` you can reduce the memory footprint for particle IO to less than ``2*N`` MiB.
 
 Additional Tools
 ^^^^^^^^^^^^^^^^
--- a/include/picongpu/plugins/openPMD/WriteSpecies.hpp
+++ b/include/picongpu/plugins/openPMD/WriteSpecies.hpp
@@ -28,6 +28,7 @@
 #include "picongpu/plugins/openPMD/openPMDDimension.hpp"
 #include "picongpu/plugins/openPMD/openPMDWriter.def"
 #include "picongpu/plugins/openPMD/writer/ParticleAttribute.hpp"
+#include "picongpu/plugins/openPMD/writer/misc.hpp"
 #include "picongpu/plugins/output/ConstSpeciesAttributes.hpp"
 #include "picongpu/plugins/output/WriteSpeciesCommon.hpp"
 
@@ -35,6 +36,7 @@
 #include <pmacc/dataManagement/DataConnector.hpp>
 #include <pmacc/eventSystem/events/kernelEvents.hpp>
 #include <pmacc/mappings/kernel/AreaMapping.hpp>
+#include <pmacc/mappings/kernel/RangeMapping.hpp>
 #include <pmacc/meta/conversion/MakeSeq.hpp>
 #include <pmacc/meta/conversion/RemoveFromSeq.hpp>
 #include <pmacc/particles/ParticleDescription.hpp>
@@ -64,7 +66,8 @@ namespace picongpu
             Filter& filter;
             ParticleFilter& particleFilter;
             T_ParticleOffset& particleToTotalDomainOffset;
-            uint64_t myNumParticles, globalNumParticles;
+            uint64_t globalNumParticles;
+
             StrategyRunParameters(
                 pmacc::DataConnector& c_dc,
                 ThreadParams& c_params,
@@ -72,7 +75,6 @@ namespace picongpu
                 Filter& c_filter,
                 ParticleFilter& c_particleFilter,
                 T_ParticleOffset& c_particleToTotalDomainOffset,
-                uint64_t c_myNumParticles,
                 uint64_t c_globalNumParticles)
                 : dc(c_dc)
                 , params(c_params)
@@ -80,7 +82,6 @@ namespace picongpu
                 , filter(c_filter)
                 , particleFilter(c_particleFilter)
                 , particleToTotalDomainOffset(c_particleToTotalDomainOffset)
-                , myNumParticles(c_myNumParticles)
                 , globalNumParticles(c_globalNumParticles)
             {
             }
@@ -97,7 +98,14 @@ namespace picongpu
 
             virtual FrameType malloc(std::string name, uint64_cu const myNumParticles) = 0;
 
-            virtual void prepare(uint32_t const currentStep, std::string name, RunParameters) = 0;
+            virtual void prepare(
+                uint32_t const currentStep,
+                std::string name,
+                RunParameters,
+                uint64_t myNumParticles,
+                uint64_t begin,
+                uint64_t end)
+                = 0;
 
             virtual ~Strategy() = default;
         };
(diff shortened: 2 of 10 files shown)


psychocoderHPC:
@franzpoeschel could you please check if I implemented the json/toml CLI argument correctly.

include/picongpu/plugins/openPMD/WriteSpecies.hpp
```diff
@@ -398,39 +421,85 @@ namespace picongpu
...
+                            hostFrame,
+                            particleSpecies,
+                            basename,
+                            chunk.numberOfParticles,
+                            globalNumParticles,
+                            myParticleOffset + particleOffset);
+
+                        particleOffset += chunk.numberOfParticles;
+                        ++dumpIteration;
+                        log<picLog::INPUT_OUTPUT>("openPMD: (end) write particle records for %1%, dumping round %2%")
```
psychocoderHPC, line 498: Comment must be moved two lines up to have the correct iteration in the log message.

include/pmacc/mappings/kernel/RangeMapping.hpp
```diff
@@ -0,0 +1,166 @@
...
+#include "pmacc/dimensions/DataSpace.hpp"
+#include "pmacc/types.hpp"
+
+#include <cstdint>
+
+namespace pmacc
+{
+    /** Mapping from block indices to supercells in the given area for alpaka kernels
+     *
+     * Adheres to the MapperConcept.
```
psychocoderHPC: ????

include/pmacc/mappings/kernel/RangeMapping.hpp
```diff
@@ -0,0 +1,166 @@
...
+            return beginIdx;
+        }
+
+        /** Supercell index of the element behind the last valid element */
+        HDINLINE uint32_t end() const
+        {
+            return endIdx;
+        }
+
+        HDINLINE uint32_t last() const
```
psychocoderHPC, line 104: comments are missing for this and other mehods

include/pmacc/particles/memory/dataTypes/Particle.hpp
```diff
@@ -63,6 +76,19 @@ namespace pmacc
         using SuperCellSize = typename FrameType::SuperCellSize;
         using MethodsList = typename FrameType::MethodsList;
 
+        /** Size of an particle if all data members started without padding in a contigious data chunk.
```
psychocoderHPC: The sum of the bytes of all value identifiers of a particle.

include/picongpu/plugins/openPMD/openPMDWriter.hpp
```diff
@@ -339,7 +344,8 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
                 {&jsonRestartConfig,
                  std::nullopt,
                  &PluginParameters::jsonRestartParams,
-                 ApplyParameter::OnlyInCheckpoint}};
+                 ApplyParameter::OnlyInCheckpoint},
+                {&particleIOChunkSize, "particleIOChunkSize", &PluginParameters::particleIOChunkSizeString}};
```
franzpoeschel, line 348: The only reason why this is fixed to `std::string` parameters is because other types were not needed so far. We can also extend this to allow other types somehow.
However, I'm not sure if this should be done for memory sizes, because in future we might want to allow things like `--openPMD.particleIOChunkSize 10Gb`.

psychocoderHPC: We can stay with string, it's not a big issue and as you suggested it provides the possibility to allow postfix for the unit later on.

docs/source/usage/plugins/openPMD.rst
```diff
@@ -134,6 +134,12 @@ PIConGPU command line option          description
 ``--checkpoint.openPMD.jsonRestart``  Set backend-specific parameters for openPMD backends in JSON format for restarting from a checkpoint.
 ``--openPMD.dataPreparationStrategy`` Strategy for preparation of particle data ('doubleBuffer' or 'mappedMemory'). Aliases 'adios' and 'hdf5' may be used respectively.
 ``--openPMD.toml``                    Alternatively configure the openPMD plugin via a TOML file (see below).
+``--openPMD.particleIOChunkSize``     Particle data will be written in chunks of the given size. unit: MiB
+                                      The memory footprint on the host side is reduced compared to writing all particles with one write call.
+                                      Currently the memory footprint is only reduced bp5 is used, HDF5 and bp4 does not support partial data dump.
```
chillenzer, line 139:
```suggestion
                                      Currently the memory footprint is only reduced if bp5 is used, HDF5 and bp4 does not support partial data dump.
```

docs/source/usage/plugins/openPMD.rst
```diff
@@ -254,6 +260,7 @@ As soon as the openPMD plugin is compiled in, one extra ``mallocMC`` heap for th
 During I/O, particle attributes are allocated one after another.
 Using ``--openPMD.dataPreparationStrategy doubleBuffer`` (default) will require at least 2x the GPU memory on the host side.
 For a smaller host side memory footprint (<< GPU main memory) pick ``--openPMD.dataPreparationStrategy mappedMemory``.
+By using additional to the mapped memory strategy ``--openPMD.particleIOChunkSize N`` you can reduce the memory footprint for particle IO to less than ``2*N`` MiB.
```
chillenzer, line 263:
```suggestion
In addition to using the mapped-memory strategy, you can reduce the memory footprint for particle IO to less than ``2*N`` MiB ``--openPMD.particleIOChunkSize N``.
```

franzpoeschel:
Running some benchmarks with Heaptrack:

`picongpu -d 1 1 1 -g 192 2048 192 -s 300 --openPMD.period 100 --openPMD.ext bp5 --openPMD.particleIOChunkSize 1024`
Bildschirmfoto vom 2024-04-22 14-37-12

`picongpu -d 1 1 1 -g 192 2048 192 -s 300 --openPMD.period 100 --openPMD.ext bp5 --openPMD.particleIOChunkSize 1024 --openPMD.dataPreparationStrategy mappedMemory`
Bildschirmfoto vom 2024-04-22 14-57-14

`picongpu -d 1 1 1 -g 192 2048 192 -s 300 --openPMD.period 100 --openPMD.ext bp5`
Bildschirmfoto vom 2024-04-22 15-04-01

`picongpu -d 1 1 1 -g 192 2048 192 -s 300 --openPMD.period 100 --openPMD.ext bp5 --openPMD.dataPreparationStrategy mappedMemory`
Bildschirmfoto vom 2024-04-22 15-27-29


**Before this PR:**

`picongpu -d 1 1 1 -g 192 2048 192 -s 300 --openPMD.period 100 --openPMD.ext bp5`
Bildschirmfoto vom 2024-04-22 15-16-25


`picongpu -d 1 1 1 -g 192 2048 192 -s 300 --openPMD.period 100 --openPMD.ext bp5 --openPMD.dataPreparationStrategy mappedMemory`
Bildschirmfoto vom 2024-04-22 15-20-10

psychocoderHPC:
@franzpoeschel Thanks for running the benchmarks. Could you rerun once 
`picongpu -d 1 1 1 -g 192 2048 192 -s 300 --openPMD.period 100 --openPMD.ext bp5 --openPMD.particleIOChunkSize 128 --openPMD.dataPreparationStrategy mappedMemory`

This should show a clear difference in the memory footprint but will run slower. In the current plot the difference between before and after is hard to see expect some dips during the write because the example is not writing to much data.

franzpoeschel:
I was wondering where the 2GB come from but that's the BufferChunkSize of BP5, Heaptrack records the virtual memory. Let me run again with a changed size. We could theoretically change the default size here, but I'm not sure we should since the overhead is only virtual.

Bildschirmfoto vom 2024-04-23 10-22-41

franzpoeschel:
Bildschirmfoto vom 2024-04-23 10-28-54
This is with `picongpu -d 1 1 1 -g 192 2048 192 -s 300 --openPMD.period 100 --openPMD.ext bp5 --openPMD.particleIOChunkSize 128 --openPMD.dataPreparationStrategy mappedMemory --openPMD.json '{"adios2":{"engine":{"parameters":{"BufferChunkSize": "200Mb"}}}}'`

psychocoderHPC:
explanation: The 300MiB spike at the beginning is coming from the fields. Currently we only chunk particle IO.

Chunking of field IO is much easier and can be implemented later if needed.

include/picongpu/plugins/openPMD/WriteSpecies.hpp
```diff
@@ -97,7 +98,14 @@ namespace picongpu
...
             virtual FrameType malloc(std::string name, uint64_cu const myNumParticles) = 0;
 
-            virtual void prepare(uint32_t const currentStep, std::string name, RunParameters) = 0;
+            virtual void prepare(
+                uint32_t const currentStep,
+                std::string name,
+                RunParameters,
+                uint64_t myNumParticles,
+                uint64_t begin,
+                uint64_t end)
```
chillenzer, line 107: That is kind of a verbose interface and keeping all three in sync and free of typos will be a hassle. Compared to the C++ evolution of `for`-loop -> iterators and algorithms -> ranges, that's more or less on the `for`-loop level. Not sure if a full abstraction type similar to a range is needed but couldn't one of the three be deduced from the other two?

franzpoeschel:
Since the particles spread over different components (x, y, z), the BufferChunkSize could be turned even smaller probably, but I think the profiles show that this PR does what it intends to. Otherwise, I don't see anything that is obviously wrong.

franzpoeschel (approved):

psychocoderHPC:
@ComputationalRadiationPhysics/picongpu-maintainers PR can be merged

docs/source/usage/plugins/openPMD.rst
PrometheusPi, line 139: This is not a sentence. Is an if missing? 
```suggestion
                                      Currently the memory footprint is only reduced if bp5 is used, HDF5 and bp4 does not support partial data dump.
```

PrometheusPi (changes requested):
just a minor typo

docs/source/usage/plugins/openPMD.rst
PrometheusPi: Discussed *offline* with @psychocoderHPC: the typo will be fixed in a followup pll request.

PrometheusPi (approved):

chillenzer (commented):
Sorry, had a half-finished review open. Not yet through but maybe you want to consider this already.

include/picongpu/plugins/openPMD/WriteSpecies.hpp
psychocoderHPC: not fully sure what you mean, we should discuss it tomorrow offline. Never the less I will change it and forward `ChunkDescription` which contains all three values in a struct anyway.

PR #4880, opened 2024-04-24
fix lumi profile

SimeonEhrig:
KelvinHelmholtz example compiles, but I don't know if it is running because I have no idea to launch the example.

### DIFF
--- a/etc/picongpu/lumi-eurohpc/lumi-G_hipcc_picongpu.profile.example
+++ b/etc/picongpu/lumi-eurohpc/lumi-G_hipcc_picongpu.profile.example
@@ -48,13 +48,17 @@ module load buildtools/23.03
 
 export MPICH_GPU_SUPPORT_ENABLED=1
 
-export CXX=hipcc
+export PATH=$ROCM_PATH/llvm/bin:$PATH
+export CC=clang
+export CXX=clang++
+
 export CXXFLAGS="$CXXFLAGS -I${MPICH_DIR}/include"
+export HIPFLAGS=$CXXFLAGS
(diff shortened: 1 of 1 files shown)


BrianMarre:
you should be able to run after compiling with the following:

```bash
tbg -s sbatch -c etc/picongpu/1.cfg -t etc/picongpu/lumi-eurohpc/standard-g.tpl ~/test_khi
```

BrianMarre:
our tpl file for lumi also contains a comment that says it is intended for frontier, we might want to change that ...

SimeonEhrig:
Okay, I found the typo in the `tgb` command. Now, I'm confirmed, that something is wrong with the MPI module:

```bash
tbg/submit.start: line 46: mpiexec: command not found
```

PrometheusPi:
@SimeonEhrig strange - they must have changed something since last summer - thanks for fixing

SimeonEhrig:
@PrometheusPi I think so.

image

etc/picongpu/lumi-eurohpc/lumi-G_hipcc_picongpu.profile.example
```diff
@@ -48,13 +48,17 @@ module load buildtools/23.03
...
+export CC=clang
+export CXX=clang++
+
 export CXXFLAGS="$CXXFLAGS -I${MPICH_DIR}/include"
+export HIPFLAGS=$CXXFLAGS
 export LDFLAGS="$LDFLAGS -L${MPICH_DIR}/lib -lmpi ${PE_MPICH_GTL_DIR_amd_gfx90a} ${PE_MPICH_GTL_LIBS_amd_gfx90a}"
 
 # Other Software ##############################################################
 #
-libpng/1.6.38-cpeCray-23.03
```
PrometheusPi, line 57: sorry - I never use png output and thus did not see this error

PrometheusPi (approved):

psychocoderHPC (approved):

PR #4882, opened 2024-04-25
optimize OpenPMD IO

psychocoderHPC:
To avoid that particle data flushed to disk is to small to get a good IO performance we do not flush anymore each component of an particle atribute. Instead we flush always a full particle IO chunk which is by default 10 GiB data.

- [x] test restart on our dev server

### DIFF
--- a/include/picongpu/plugins/openPMD/writer/ParticleAttribute.hpp
+++ b/include/picongpu/plugins/openPMD/writer/ParticleAttribute.hpp
@@ -49,6 +49,8 @@ namespace picongpu
         {
             /** write attribute to openPMD series
              *
+             * @attention This is a MPI_Collective operation and all MPI ranks must participate.
+             *
              * @param params wrapped params
              * @param elements elements of this attribute
              */
@@ -98,8 +100,6 @@ namespace picongpu
 
                 log<picLog::INPUT_OUTPUT>("openPMD:  (begin) write species attribute: %1%") % Identifier::getName();
 
-                std::shared_ptr<ComponentType> storeBfr;
-
                 for(uint32_t d = 0; d < components; d++)
                 {
                     ::openPMD::RecordComponent recordComponent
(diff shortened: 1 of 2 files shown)


include/picongpu/plugins/openPMD/writer/ParticleAttribute.hpp
```diff
@@ -115,39 +117,23 @@ namespace picongpu
...
                     }
 
                     ValueType* dataPtr = frame.getIdentifier(Identifier()).getPointer(); // can be moved up?
                     // ask openPMD to create a buffer for us
-                    // in some backends (ADIOS2), this allows avoiding memcopies
-                    auto span = recordComponent
-                                    .storeChunk<ComponentType>(
-                                        ::openPMD::Offset{globalOffset},
-                                        ::openPMD::Extent{elements},
-                                        [&storeBfr](size_t size)
```
franzpoeschel, line 129: The storeBfr is defined above and can now be deleted too. Will look in further detail later or tomorrow, including tests.

franzpoeschel:
Running `picongpu -d 1 1 1 -g 192 2048 192 -s 300 --openPMD.period 100 --openPMD.ext bp5 --openPMD.particleIOChunkSize 128 --openPMD.dataPreparationStrategy mappedMemory --openPMD.json '{"adios2":{"engine":{"parameters":{"BufferChunkSize": "5Mb"}}}}'`:

**Before this PR**
Bildschirmfoto vom 2024-04-25 15-42-26

**With this PR**
Bildschirmfoto vom 2024-04-25 15-42-50

This shows that with this PR, the memory usage is very close to the specified buffer chunk size, while IO was too fine-grained before.

I noticed also that the particle attributes get initialized multiple times which might cause problems under certain conditions, currently looking into that. But that's probably unrelated to this PR anyway.

franzpoeschel:
> I noticed also that the particle attributes get initialized multiple times 

Ah, it's good that you do it that way at the moment, there is currently a bug on our dev that would lead to skipped flushes otherwise, so I need to fix that first.

include/picongpu/plugins/openPMD/writer/ParticleAttribute.hpp
psychocoderHPC: thx, done

include/picongpu/plugins/openPMD/writer/ParticleAttribute.hpp
```diff
@@ -48,6 +48,8 @@ namespace picongpu
         struct ParticleAttribute
         {
             /** write attribute to openPMD series
+             *
+             * @attention This is a MPI_Collective operation and all MPI ranks must participate.
```
franzpoeschel, line 52: The only collective part is the creation of the record which I will factor out in a follow-up PR. Fine to leave this comment in for now, but it will be very temporary only.

franzpoeschel (approved):

franzpoeschel:
The follow-up is https://github.com/ComputationalRadiationPhysics/picongpu/pull/4887, based on this PR. Feel free to modify as you please, I'm on vacation next week.

ikbuibui (approved):

PR #4883, opened 2024-04-25
ADIOS2 v2.10: BP5 is there, but the macro is missing

franzpoeschel:
Fixed in openPMD-api by https://github.com/openPMD/openPMD-api/pull/1614. Versions that don't have this fix will get confused and hand down their confusion to PIConGPU.


psychocoderHPC:
please rebase, our CI was broken because of pip issues

psychocoderHPC (approved):
all fine needs only a rebase and CI needs passing.

PR #4885, opened 2024-04-26
optimize particle restart

psychocoderHPC:
Currently we need to hold all particles for the device in CPU memory to restart from a checkpoint. On systems with large on device memory but low amount of CPU memory e.g. ORNL Frontier we can not restart large simulations.
The OpenPMD plugin has since years chunked particle restart but applied this chunked load only to the frame de-serialization kernel and not to the file load.

This PR is loading particles in chunks/batches with openPMD-api, upload these to the device and is repeating this until all particles are loaded.
This will reduce the CPU memory footprint to `2 * numParticlePerChunk * sizeof(Particle)`.

- [x] tested on our dev server (2 MPI ranks)

### DIFF
--- a/include/picongpu/plugins/openPMD/restart/LoadSpecies.hpp
+++ b/include/picongpu/plugins/openPMD/restart/LoadSpecies.hpp
@@ -136,28 +136,49 @@ namespace picongpu
                 BufferType buffers;
                 FrameType mappedFrame;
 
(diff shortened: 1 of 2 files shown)


ikbuibui (approved):

ikbuibui:
Looks good

PR #4886, opened 2024-04-26
fix CI

psychocoderHPC:
The CI is failing with

```
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
...
```

when installing packages with pip on new ubuntu versions. I workaround is to pass the parameter `--break-system-packages` to pip. A more clean way would be to use python environments, which is not part of this PR.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -11,6 +11,8 @@ variables:
 
 .base_generate-reduced-matrix:
   stage: generate
+  variables:
+    PIP_BREAK_SYSTEM_PACKAGES: 1
   script:
     - apt update
     - apt install -y python3-pip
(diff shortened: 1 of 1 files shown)


chillenzer:
Have you tried if `--user` suffices to fix this? The line that throws [here](https://gitlab.com/hzdr/crp/picongpu/-/jobs/6723465947) is one without `--user/-U`.

psychocoderHPC:
> Have you tried if `--user` suffices to fix this? The line that throws [here](https://gitlab.com/hzdr/crp/picongpu/-/jobs/6723465947) is one without `--user/-U`.

No, the solution I found was via setting an environment var or using the parameter to disable this warning.

psychocoderHPC:
I will merge this PR to before the CI passed to allow rebasing all blocked PR's

PR #4887, opened 2024-04-26
Sliced writing fixes

franzpoeschel:
- [x] Merge #4882 first

The current implementation of sliced writing re-initializes the components over and over again for each component. For the attributes, that's just duplicate work, but additionally the array sizes get "changed" over and over again (to the same size as before). This might lead to trouble in backends that have restricted support for extensible arrays.

HOWEVER, this PR runs into a bug on the current dev branch of the openPMD-api, probably introduced by https://github.com/openPMD/openPMD-api/pull/1598 and fixed by https://github.com/openPMD/openPMD-api/pull/1615. Will need to do more testing to confirm this. The latest release seems unaffected, but please don't merge just yet.

Apart from this, to workaround another parallel flushing bug that René discovered, I should probably work sth in to make the iteration artificially dirty anyway.

- [x] see above, find a way to make this all work nicely

### DIFF
--- a/include/picongpu/plugins/openPMD/WriteSpecies.hpp
+++ b/include/picongpu/plugins/openPMD/WriteSpecies.hpp
@@ -480,6 +480,9 @@ namespace picongpu
                             globalNumParticles,
                             myParticleOffset + particleOffset);
 
+                        log<picLog::INPUT_OUTPUT>("openPMD: flush particle records for %1%, dumping round %2%")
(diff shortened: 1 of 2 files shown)


franzpoeschel:
I now added a workaround to forcibly mark the iteration dirty. This ensures that things are flushed when they should.

psychocoderHPC (dismissed):

psychocoderHPC:
I rebased this PR

psychocoderHPC (approved):

PR #4889, opened 2024-04-26
fix Gaussian and Dispersive pulse

### LINKED ISSUE #4884: Negative focus for Gaussian and Dispersive pulse does not work
When using a Gaussian laser and putting the focus outside of the simulation box behind the propagation direction, one would expect an evenly diverging beam inside the simulation. I have tested this for a Gaussian pulse propagating in positive z direction and a negative focus. Here are my param files: [params.zip](https://github.com/ComputationalRadiationPhysics/picongpu/files/15127021/params.zip). This is called with 2x2x2 v100s on a 512x512x4096 grid with 10k timesteps. In the simulation we then see a focusing beam as if the focus was positive. You can see that here:
 
defocus

@steindev @pordyna

psychocoderHPC:
fix #4884

Setting a negative focus was not possible.
This PR projects the PIConGPU coordinates into the laser coordinate system before calculating the distance in laser propagation direction from the laser plane to the current point(cell) we evaluate the pulse.

@finnolec Please check if your issue is solved by this PR.

### DIFF
--- a/include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
+++ b/include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
@@ -154,10 +154,20 @@ namespace picongpu
                             // transform to 3d internal coordinate system
                             float3_X pos = this->getInternalCoordinates(totalCellIdx);
 
-                            // calculate focus position relative to the current point in the propagation direction
-                            auto const focusRelativeToOrigin = this->focus - this->origin;
-                            // current distance to focus position
-                            float_X const focusPos = math::sqrt(pmacc::math::l2norm2(focusRelativeToOrigin)) - pos[0];
+                            /* Calculate focus position relative to the current point in the propagation direction.
+                             * Coordinate system is PIConGPUs and not the laser internal coordinate system where X is
+                             * the propagation direction.
+                             */
+                            float3_X const focusRelativeToOrigin = this->focus - this->origin;
+
+                            /* Relative focus distance from the laser origin to the focus, transformed into the laser
+                             * coordination system where X is the propagation direction. */
+                            float_X const distanceFocusRelativeToOrigin
+                                = pmacc::math::dot(focusRelativeToOrigin, this->getAxis0());
+
+                            // Distance from the current cell to the focus in laser propagation direction.
+                            float_X const focusPos = distanceFocusRelativeToOrigin - pos[0];
+
                             // beam waist at the generation plane so that at focus we will get W0
                             float_X const waist = Unitless::W0
                                 * math::sqrt(1.0_X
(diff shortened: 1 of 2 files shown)


include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -225,10 +225,14 @@ namespace picongpu::fields::incidentField
...
-                    // calculate focus position relative to the current point in the propagation direction
+                    /* Calculate focus position relative to the current point in the propagation direction.
+                     * Coordinate system is PIConGPUs and not the laser internal coordinate system where X is the
+                     * propagation direction.
+                     */
                     auto const focusRelativeToOrigin = this->focus - this->origin;
 
-                    float_X const focusPos = math::sqrt(pmacc::math::l2norm2(focusRelativeToOrigin)) - pos[0];
+                    // Distance from the current cell to the focus in laser propagation direction.
+                    float_X const focusPos = focusRelativeToOrigin.sumOfComponents() - pos[0];
```
PrometheusPi: I am not sure whether this still works if less than 2 of the 3 components are zero.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -211,9 +216,14 @@ namespace picongpu
...
-                            // calculate focus position relative to the current point in the propagation direction
+                            /* Calculate focus position relative to the current point in the propagation direction.
+                             * Coordinate system is PIConGPUs and not the laser internal coordinate system where X is
+                             * the propagation direction.
+                             */
                             auto const focusRelativeToOrigin = this->focus - this->origin;
-                            float_X const focusPos = math::sqrt(pmacc::math::l2norm2(focusRelativeToOrigin)) - pos[0];
+
+                            // Distance from the current cell to the focus in laser propagation direction.
+                            float_X const focusPos = focusRelativeToOrigin.sumOfComponents() - pos[0];
```
PrometheusPi: I am not sure whether this still works if less than 2 of the 3 components are zero.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -154,10 +154,15 @@ namespace picongpu
...
+                            /* Calculate focus position relative to the current point in the propagation direction.
+                             * Coordinate system is PIConGPUs and not the laser internal coordinate system where X is
+                             * the propagation direction.
+                             */
                             auto const focusRelativeToOrigin = this->focus - this->origin;
-                            // current distance to focus position
-                            float_X const focusPos = math::sqrt(pmacc::math::l2norm2(focusRelativeToOrigin)) - pos[0];
+
+                            // Distance from the current cell to the focus in laser propagation direction.
+                            float_X const focusPos = focusRelativeToOrigin.sumOfComponents() - pos[0];
```
PrometheusPi: I am not sure whether this still works if less than 2 of the 3 components are zero.

PrometheusPi (changes requested):
I am very certain, it will fail under the mentioned condition. I will perform test to validate or disprove my claim.

psychocoderHPC:
~~This should not happen because only the orthogonal to the laser plane component should be non zero~~

[Update] my tests was wrong because I added a laser which propagates in Y direction only and not diagonal.

psychocoderHPC:
@PrometheusPi could you please have a look agin into it, IMO it should now be correct.

PrometheusPi:
I checked the physics on a non-axis-parallel laser with negative focus position. The pull request fixes the issue. 

In below plots, the focus should be outside the box in, half a box left and up. Thus, curvature should be away from the ficus point. 

The `dev`branch focus to the center:
grafik

The *pull request* fixes that:
grafik

PrometheusPi (approved):

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -225,10 +225,19 @@ namespace picongpu::fields::incidentField
...
+                    /* Calculate focus position relative to the current point in the propagation direction.
+                     * Coordinate system is PIConGPUs and not the laser internal coordinate system where X is the
+                     * propagation direction.
+                     */
+                    float3_X const focusRelativeToOrigin = this->focus - this->origin;
+
+                    /* Relative focus distance from the laser origin to the focus, transformed into the laser
+                     * coordination system where X is the propagation direction. */
+                    float_X const distanceFocusRelativeToOrigin
+                        = pmacc::math::dot(focusRelativeToOrigin, this->getAxis0());
```
PrometheusPi, line 237: @psychocoderHPC `this->getAxis0()` is a unit vector, right?

PR #4892, opened 2024-04-30
openPMD extension: fix broken C++ syntax

psychocoderHPC:
Bug introduced with #4883.
The bug is not detected by the CI because the code path depends on the ADIOS2 version.


PrometheusPi (approved):
tested - works

Issue #4893, opened 2024-04-30
Shadowgraphy plugin causes compile issue


PrometheusPi:
When compiling the current `dev`, `LaserWakefield` example on hemera, using the up-to-date V100 profile, we run into the following compile error:
```
/bigdata/hplsim/development/username/pr_4889/picongpu/include/picongpu/../picongpu/plugins/shadowgraphy/Shadowgraphy.hpp(254): error: expected a declaration
                  .
                  ^

/bigdata/hplsim/development/username/pr_4889/picongpu/include/picongpu/../picongpu/plugins/shadowgraphy/Shadowgraphy.hpp(285): warning #12-D: parsing restarts here after previous syntax error
                      int localStep = (currentStep - startTime) / params::tRes;
                                                                              ^

Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"

/bigdata/hplsim/development/username/pr_4889/picongpu/include/picongpu/../picongpu/plugins/shadowgraphy/Shadowgraphy.hpp(288): error: expected a declaration
                      if(!dumpFinalData)
                      ^

/bigdata/hplsim/development/username/pr_4889/picongpu/include/picongpu/../picongpu/plugins/shadowgraphy/Shadowgraphy.hpp(312): warning #12-D: parsing restarts here after previous syntax error
                                  sliceBufferB);
                                               ^

/bigdata/hplsim/development/username/pr_4889/picongpu/include/picongpu/../picongpu/plugins/shadowgraphy/Shadowgraphy.hpp(287): error: identifier "localStep" is undefined
                      bool const dumpFinalData = localStep == (adjustedDuration / params::tRes);
                                                 ^

/bigdata/hplsim/development/username/pr_4889/picongpu/include/picongpu/../picongpu/plugins/shadowgraphy/Shadowgraphy.hpp(70): error: object of abstract class type "picongpu::plugins::shadowgraphy::Shadowgraphy" is not allowed:
            pure virtual function "pmacc::INotify::notify" has no overrider
                          return std::shared_ptr<IInstance>(new Shadowgraphy(help, id, cellDescription));
                                                                ^

/bigdata/hplsim/development/username/pr_4889/picongpu/include/picongpu/../picongpu/plugins/shadowgraphy/Shadowgraphy.hpp(315): error: expected a ";"
                          if(gather->isMaster())
                          ^

[...]

PR #4868, opened 2024-04-09
Add shadowgraphy plugin

finnolec:
Hello, here is the shadowgraphy plugin from my master's thesis.

## What does it do

The plugin integrates the Poynting vectors from the electric and magnetic field in a spatially static slice in the simulation to generate a synthetic 2D shadowgram.

When a probe laser (implemented with e.g. the incident field method) propagates through the plasma structures, it contains information about these plasma structures through refraction or diffraction. The probe laser can then be propagated with this in-situ PIConGPU plugin through a 4f correlator with a Fourier optics approach on a screen. This results in a 2D image, a so called "shadowgram". Additionally, other optical effects of the plasma are contained in this calculation method, that naturally occur through PIConGPU, like relativistic refraction index, plasma self emission or ionization effects.

Compared to the post-processing implementation that it is based on ([E Siminos et al 2016 Plasma Phys. Control. Fusion 58 065004](https://iopscience.iop.org/article/10.1088/0741-3335/58/6/065004/meta)), the in-situ shadowgraphy plugin allows to calculate synthetic shadowgrams without having to store and load the field data. The laser is also extracted slice-wise over a set period of time, instead of a 3D Fourier transform of simulation box volume.

## Tests

I have included a test script for the plugin that tests the gathering and Fourier transforms of the plugin. This can be executed by starting `bin/ci.sh`. After the test the Gaussian laser looks physical with the expected properties from the set parameters in `incidentField.param`. I have tested the propagator while cleaning the code and the code propagates the fields as expected. This test was done by comparing the waist of a propagated Gaussian beam with the beam waist given by the rayleigh length. I could add this as an additional test in a further pull request.

### FFT things

Now that there is a running test, I might optimize the FFTs in the future. The addition would be a task for a new pull request.

### Checkpoints

Currently checkpointing is not supported. It would generally be a useful feature, since the plugin is doing a time integration. I will not implement it in the near future though because the plugin seems fast enough that it's not required yet.

PR #4894, opened 2024-04-30
fix Shadowgraphy compile

PrometheusPi:
This solves #4893

There was an extra dot in #4868 which caused compile issues and which this pull request removes. 

@finnolec With what version did you run your tests?


psychocoderHPC (approved):

finnolec:
For history reasons: I ran my tests with 62a492f47426cdace5aaca459e986a36b1a97d9a, but not with the following commit (b91b14f048d1cf8435e88a772571f7247358bc55) which then broke the plugin.

PR #4898, opened 2024-05-02
fix include guard Shadowgraphy plugin

PrometheusPi:
fix https://github.com/ComputationalRadiationPhysics/picongpu/issues/4897

This pull request fixes #4897 the missing openPMD-api include guard for the shadowgraphy plugin.

PR #4899, opened 2024-05-02
fix missing include guard for shadowgraphy plugin

psychocoderHPC:
fix #4898


psychocoderHPC:
closed and solved already with #4898

#4898 is not using nested guards, same as on other places in the file

PR #4903, opened 2024-05-08
CI: test latest ISAAC

psychocoderHPC:
- install glm 1.0.1 for ISAAC
- test latest ISAAC dev branch

### DIFF
--- a/share/ci/install/isaac.sh
+++ b/share/ci/install/isaac.sh
@@ -6,11 +6,11 @@ set -o pipefail
 if [ -z "$DISABLE_ISAAC" ] ; then
     cd $CI_PROJECT_DIR
 
-    export GLM_ROOT=/opt/glm/0.9.9.9-dev
+    export GLM_VERSION=1.0.1
+    export GLM_ROOT=/opt/glm/$GLM_VERSION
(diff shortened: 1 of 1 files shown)


share/ci/install/isaac.sh
```diff
@@ -19,7 +19,7 @@ if [ -z "$DISABLE_ISAAC" ] ; then
     cd $CI_PROJECT_DIR
     git clone https://github.com/ComputationalRadiationPhysics/isaac.git
     cd isaac
-    git checkout 195420ca1c43f6148c7f6c3a10ad9cad32e04d6a
+    git checkout 6bf8e189142da7ab16f8367316fb21dcaa7e3b78
```
PrometheusPi, line 22: @psychocoderHPC why do you specify a specific branch (the current head?)

PrometheusPi (approved):
Just a question

share/ci/install/isaac.sh
PrometheusPi: *offline discussion with @psychocoderHPC*: This avoids that the PIConGPU CI breaks because we accidentally introduce an error in ISAAC.

PR #4904, opened 2024-05-14
add warning for AOFD solver

PrometheusPi:
This pull request adds a warning in the user documentation of the AOFDTD solver to avoid the common mistake of setting the time step close to the CFL condition and thus ignoring the issues airing with that with the numerical dispersion relation.  

There is **noting wrong** with the documentation, this should just avoid a **very common** pitfall.


chillenzer (approved):
Great idea!

psychocoderHPC:
@steindev ping

steindev (approved):

PR #4905, opened 2024-05-15
fix slow tbg substitution

psychocoderHPC:
tbg is very slow on systems with many environment variables e.g. on JUWELS, the reasons are NxN substitutions.

This PR is using a more clever way to resolve tbg variables. Bash substring manipulations were used whcih solves ugly escaping issue we had in the past.

The tbg execution time goes down on Juwels from >5 minutes to 26 seconds.

fix #4409

A side effect of this PR is that #4409 got solved.

PR #4906, opened 2024-05-16
modernize `tbg`

psychocoderHPC:
- rename some variables, no need to start variables with `tooltpl` anymore since we removed a lot of our ugly `sed` magic in #4905
- use `$(command)` indtead of `'comamnd'`

### DIFF
--- a/bin/tbg
+++ b/bin/tbg
@@ -37,7 +37,7 @@ function resolve_vars
 
     while read -r data_set
     do
-        unresolved_vars=`echo "$data_set" | grep -o "\![[:alpha:]][[:alnum:]_]*"`
+        unresolved_vars=$(echo "$data_set" | grep -o "\![[:alpha:]][[:alnum:]_]*")
         unresolved_vars_old="dummy data for the first round"
         # loop until all variables are resolved, if nothing is changing because of a cyclic dependency than stop
         while [ -n  "$unresolved_vars" ] && [ "$unresolved_vars" != "$unresolved_vars_old" ]
@@ -52,7 +52,7 @@ function resolve_vars
                 fi
             done
             unresolved_vars_old="$unresolved_vars"
-            unresolved_vars=`echo "$data_set" | grep -o "\![[:alpha:]][[:alnum:]_]*"`
+            unresolved_vars=$(echo "$data_set" | grep -o "\![[:alpha:]][[:alnum:]_]*")
         done
         if [ -n  "$unresolved_vars" ] && [ "$unresolved_vars" == "$unresolved_vars_old" ] ; then
             echo "Possible cyclic dependency detected or unknown variables used!" >&2
@@ -64,7 +64,7 @@ function resolve_vars
         fi
 
         export "$data_set"
-    done < <(echo "$replace_input" | grep -v tooltpl | grep "^[[:alpha:]][[:alnum:]_]*=.*" )
+    done < <(echo "$replace_input" | grep "^[[:alpha:]][[:alnum:]_]*=.*" )
 }
 
 #######################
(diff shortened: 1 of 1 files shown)


PrometheusPi (dismissed):
Looks good to me - will test asap

psychocoderHPC:
This will give a small overview about both ways to execute an command https://www.redhat.com/sysadmin/backtick-operator-vs-parens

PrometheusPi:
If I run my test from this morning, I get a `grep` warning:
```
grep: character class syntax is [[:space:]], not [:space:]
```
and it takes longer, now 28s.

bin/tbg
```diff
@@ -208,15 +207,15 @@ function check_final
...
+    not_replaced=$(grep -o "\![[:alpha:]][[:alnum:]_]*" $final_file | sort | uniq)
+    not_replaced_cnt=$(echo $not_replaced | wc -w)
 
     if [ $not_replaced_cnt -gt 0 ] ; then
         echo "ERROR: $not_replaced_cnt variable(s) _not_ replaced from template (tpl):"
         echo $not_replaced
 
         #create an OR concated pattern
-        nrv_or=`echo $not_replaced | sed 's/[[:space:]]/|/g'`
+        nrv_or=$(echo $not_replaced | sed 's/[[:space:]]/|/g')
```
PrometheusPi, line 218: That's most likely the line causing the warning

psychocoderHPC: I fixed it. It was coming from `egrep` not sed. I do not fully understand why `[[:blank:]]` is throwing the warning.
I switched to the old-school way of defining a list of characters.

psychocoderHPC: https://github.com/ComputationalRadiationPhysics/picongpu/pull/4906/files#diff-7b05b982bf71e63ce4dd3f2fd29332ae3e3023f0357d503da6d0f011d936a155R393 is the problematic line

chillenzer:
I'm not familiar enough with the code of this to see if there are any bad things hidden under a thick layer of trivial changes. Two suggestions:
1. Splitting this into several commits might help reviewing this.
2. Most of the changes I've seen are not something anyone should be doing by hand. Should we adapt [shellcheck](https://github.com/koalaman/shellcheck). Not sure if this one also automatically applies fixes but probably something like that exists.

bin/tbg
psychocoderHPC: @PrometheusPi this is already solved in the PR

psychocoderHPC:
> 1. Splitting this into several commits might help reviewing this.

All changes are in one file splitting it now will only introduce errors.

> 2. Most of the changes I've seen are not something anyone should be doing by hand. Should we adapt [shellcheck](https://github.com/koalaman/shellcheck). Not sure if this one also automatically applies fixes but probably something like that exists.

Yes, we can do it independent of this PR.

bin/tbg
PrometheusPi: I checked that I used your (new?) version with `nrv_or=$(echo $not_replaced | sed 's/[[:space:]]/|/g')` but I still get the warning:
```
grep: character class syntax is [[:space:]], not [:space:]
```

PrometheusPi (changes requested):

bin/tbg
psychocoderHPC: offline checked: @PrometheusPi did a mistake and used the old version and not the latest PR.

PrometheusPi: Yes, I pulled something from GitHub, but not the newest version. Tested it now, and it does not show the warning anymore.

PrometheusPi (approved):

PR #4907, opened 2024-05-17
remove atomic physics

psychocoderHPC:
The atomic physics code in its current form is not to be used and will be replaced with a new version. To avoid that we bring it in under the current work title name `AtomicPhysics2` the current code and documentation are removed.
Remove atomic physics except for species attributes `atomicPhysicsSolver` and `atomicConfigNumber`.
Charge manipulation methods will be kept too.


BrianMarre (approved):
looks good to me

ikbuibui (approved):
I am slightly surprised that the interaction with the rest of the code was so little, only the `atomic levels` attribute and the atomic physics simulation stage. This looks good to me.

PR #4914, opened 2024-05-31
matrix multiplication for pmacc

BrianMarre:
adds a matrix class and common matrix operations for small matrices.

Is intended for making writing lorentz transformations easier.

### DIFF
--- a/include/pmacc/math/Matrix.hpp
+++ b/include/pmacc/math/Matrix.hpp
@@ -1,0 +1,200 @@
+/* Copyright 2023 Brian Marre, Rene Widera
+ *
+ * This file is part of PMacc.
+ *
+ * PMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with PMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+//! @file implementation of matrix and matrix multiplication
+
+#pragma once
+
+#include "pmacc/attribute/unroll.hpp"
+#include "pmacc/math/Vector.hpp"
+#include "pmacc/memory/Array.hpp"
+#include "pmacc/static_assert.hpp"
+
+#include <cstdint>
+
+namespace pmacc::math
+{
+    /** 2 d matrix matrix definition
+     *
+     * @tparam T_Type data type used for storage of elements
+     * @tparam T_Extent pmacc::math::CT::Vector, two dimensional vector describing extent of matrix in each dimensions
+     *
+     * @attention uses mathematical indexing
+     */
+    template<typename T_Type, typename T_Extent>
+    struct Matrix
+    {
+        PMACC_CASSERT_MSG(matrix_must_have_2_dimensional_extent, T_Extent::dim == 2);
+
+        using Extent = T_Extent;
+        using Type = T_Type;
+
+        static auto constexpr m = Extent::template at<0u>::type::value;
+        static auto constexpr n = Extent::template at<1u>::type::value;
+
+        using S_Elements = pmacc::memory::Array<T_Type, CT::volume<T_Extent>::type::value>;
+        using Idx = typename S_Elements::size_type;
+
+    private:
+        S_Elements elements;
+
+        /** get linear memory storage index from n-dimensional index
+         *
+         * indexation with the following scheme for a matrix with extent = < <N0>, <N1> >
+         *
+         * #(linear Index) | 0 | 1 | ... | <N0>-1 | <N0> | <N0> + 1 | ... | <N0>+<N0>-1 | 2 * <N0>| ...
+         * ----------------|---|---|-----|--------|------|----------|-----|-------------|---------|----
+         *                m| 0 | 1 | ... | <N0>-1 |    0 |        1 | ... |  (<N0> - 1) | 0, ...  | ...
+         *                n| 0 | 0 | ... |      0 |    1 |        1 | ... |           1 | 2, ...  | ...
+         *
+         * @attention no range checks outside debug, invalid input will lead to illegal memory access!
+         */
+        HDINLINE static Idx getLinearIndex(uint32_t const mIdx, uint32_t const nIdx)
+        {
+#ifndef NDEBUG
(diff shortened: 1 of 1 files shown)


include/pmacc/math/Matrix.hpp
```diff
@@ -0,0 +1,201 @@
...
+     */
+    template<typename T_Type, typename T_Extent>
+    struct Matrix
+    {
+        PMACC_CASSERT_MSG(matrix_must_have_2_dimensional_extent, T_Extent::dim == 2);
+
+        using Extent = T_Extent;
+        using Type = T_Type;
+
+        static auto constexpr m = Extent::template at<0u>::type::value;
```
psychocoderHPC, line 50: only a note: for CT::Vector you can use `::x::value` instead of `at`

include/pmacc/math/Matrix.hpp
```diff
@@ -0,0 +1,201 @@
...
+         * indexation with the following scheme for a matrix with extent = < <N0>, <N1> >
+         *
+         * #(linear Index) | 0 | 1 | ... | <N0>-1 | <N0> | <N0> + 1 | ... | <N0>+<N0>-1 | 2 * <N0>| ...
+         * ----------------|---|---|-----|--------|------|----------|-----|-------------|---------|----
+         *                m| 0 | 1 | ... | <N0>-1 |    0 |        1 | ... |  (<N0> - 1) | 0, ...  | ...
+         *                n| 0 | 0 | ... |      0 |    1 |        1 | ... |           1 | 2, ...  | ...
+         *
+         * @attention no range checks outside debug, invalid input will lead to illegal memory access!
+         */
+        HDINLINE static Idx getLinearIndex(uint32_t const& mIdx, uint32_t const& nIdx)
```
psychocoderHPC: should be `HDINLINE static Idx getLinearIndex(uint32_t const mIdx, uint32_t const nIdx)`

Types with size `<=64` bit should be passed by copy, better for the compiler optimizations.

psychocoderHPC (changes requested):

include/pmacc/math/Matrix.hpp
BrianMarre: Done!

include/pmacc/math/Matrix.hpp
```diff
@@ -0,0 +1,201 @@
...
+     *
+     * @tparam T_Type data type used for storage of elements
+     * @tparam T_Extent pmacc::math::CT::Vector, tow dimensional vector describing extent of matrix in each dimensions
+     *
+     * @attention uses mathematical indexing
+     */
+    template<typename T_Type, typename T_Extent>
+    struct Matrix
+    {
+        PMACC_CASSERT_MSG(matrix_must_have_2_dimensional_extent, T_Extent::dim == 2);
```
PrometheusPi, line 45: @psychocoderHPC would it be better to specialize the template for `T_Extent::dim == 2` instead of throwing an error or doesn't it really matter since the error is compile-time.

include/pmacc/math/Matrix.hpp
```diff
@@ -0,0 +1,201 @@
...
+         *
+         * @attention no range checks outside debug, invalid input will lead to illegal memory access!
+         */
+        HDINLINE static Idx getLinearIndex(uint32_t const mIdx, uint32_t const nIdx)
+        {
+#ifndef NDEBUG
+            // debug range check
+            if(mIdx >= m)
+            {
+                printf("PMACC_ERROR: invalid mIdx for matrix access!\n");
```
PrometheusPi, line 76: Could you use the log level IO from PIConGPU?

include/pmacc/math/Matrix.hpp
```diff
@@ -0,0 +1,201 @@
...
+                return static_cast<uint32_t>(0u);
+            }
+            if(nIdx >= n)
+            {
+                printf("PMACC_ERROR: invalid nIdx for matrix access!\n");
```
PrometheusPi, line 81: same here

include/pmacc/math/Matrix.hpp
```diff
@@ -0,0 +1,201 @@
...
+         * #(linear Index) | 0 | 1 | ... | <N0>-1 | <N0> | <N0> + 1 | ... | <N0>+<N0>-1 | 2 * <N0>| ...
+         * ----------------|---|---|-----|--------|------|----------|-----|-------------|---------|----
+         *                m| 0 | 1 | ... | <N0>-1 |    0 |        1 | ... |  (<N0> - 1) | 0, ...  | ...
+         *                n| 0 | 0 | ... |      0 |    1 |        1 | ... |           1 | 2, ...  | ...
+         *
+         * @attention no range checks outside debug, invalid input will lead to illegal memory access!
+         */
+        HDINLINE static Idx getLinearIndex(uint32_t const mIdx, uint32_t const nIdx)
+        {
+#ifndef NDEBUG
```
PrometheusPi, line 72: Could this be linked to a specific debug level?

include/pmacc/math/Matrix.hpp
```diff
@@ -0,0 +1,201 @@
...
+    public:
+        /** constructor
+         *
+         * initialize each member with the given value
+         *
+         * @param value value assigned to each matrix entry
+         */
+        HDINLINE explicit Matrix(T_Type const& value) : elements(value)
+        {
+            PMACC_CASSERT_MSG(not_a_matrix, T_Extent::dim == 2u);
```
PrometheusPi: this seems to be a duplication of line 45, @BrianMarre why ae both needed?

PrometheusPi (dismissed):
some minor issues

include/pmacc/math/Matrix.hpp
BrianMarre: doesn't matter in execution time, since the check for correct dimension is compile time only and I can't really specialize for a dimension since every compile time vector is it's own type.

include/pmacc/math/Matrix.hpp
BrianMarre: No, this code is supposed to be executable on device and host, and only printf ouptut is possible on gpu.

include/pmacc/math/Matrix.hpp
BrianMarre: see above

include/pmacc/math/Matrix.hpp
BrianMarre: Which one would you prefer?

include/pmacc/math/Matrix.hpp
BrianMarre: you are right, removed the constructor compile time assert, good catch

include/pmacc/math/Matrix.hpp
psychocoderHPC: InN this case an assert is fine because only 2D is supported. Specialization would make only sense if we need the object for Sfinae calls to test which signature is available.

include/pmacc/math/Matrix.hpp
psychocoderHPC: `#ifndef NDEBUG`

Is a general C++ define to know if someone is running debug code or a release version without security checks e.g. range checks.

psychocoderHPC (approved):

PR #4915, opened 2024-05-31
add logical-and and logial-or Operations

BrianMarre:
adds logical-and and logical-or for use in reduce operations to pmacc

### DIFF
--- a/include/pmacc/math/operation/And.hpp
+++ b/include/pmacc/math/operation/And.hpp
@@ -1,0 +1,44 @@
+/* Copyright 2023-2024 Brian Marre
+ *
+ * This file is part of PMacc.
+ *
+ * PMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with PMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "pmacc/types.hpp"
+
+#include <cstdint>
+
+namespace pmacc::math::operation
+{
+    //! logical and
+    struct And
+    {
+        HDINLINE void operator()(uint32_t& destination, uint32_t const& source) const
+        {
+            destination = static_cast<uint32_t>(static_cast<bool>(destination) && static_cast<bool>(source));
+        }
+
+        template<typename T_Worker>
+        HDINLINE void operator()(T_Worker const&, uint32_t& destination, uint32_t const& source) const
+        {
+            destination = static_cast<uint32_t>(static_cast<bool>(destination) && static_cast<bool>(source));
+        }
+    };
+} // namespace pmacc::math::operation
--- a/include/pmacc/math/operation/Or.hpp
+++ b/include/pmacc/math/operation/Or.hpp
@@ -1,0 +1,44 @@
+/* Copyright 2023-2024 Brian Marre
+ *
+ * This file is part of PMacc.
+ *
+ * PMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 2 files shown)


psychocoderHPC (dismissed):

include/pmacc/math/operation/And.hpp
```diff
@@ -0,0 +1,43 @@
+/* Copyright 2023 Brian Marre
```
chillenzer:
```suggestion
/* Copyright 2023-2024 Brian Marre
```

include/pmacc/math/operation/Or.hpp
```diff
@@ -0,0 +1,43 @@
+/* Copyright 2023 Brian Marre
```
chillenzer:
```suggestion
/* Copyright 2023-2024 Brian Marre
```

chillenzer (approved):

include/pmacc/math/operation/And.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include "pmacc/types.hpp"
+
+#include <cstdint>
+
+namespace pmacc::math::operation
+{
+    struct And
+    {
+        HDINLINE void operator()(uint32_t& destination, uint32_t const& source) const
```
PrometheusPi, line 33: more a question to @psychocoderHPC:
Years ago, I learned that passing by reference is only useful for larger data. I would consider `uint32_t`to be rather small and would always have chosen passing a copy. What is your opinion on that?

include/pmacc/math/operation/And.hpp
```diff
@@ -0,0 +1,43 @@
...
+        {
+            destination = destination && source;
+        }
+
+        template<typename T_Worker>
+        HDINLINE void operator()(T_Worker const&, uint32_t& destination, uint32_t const& source) const
+        {
+            destination = static_cast<uint32_t>(static_cast<bool>(destination) && static_cast<bool>(source));
```
PrometheusPi, line 41: @BrianMarre why do you cast here while not in the implementation above?

include/pmacc/math/operation/Or.hpp
```diff
@@ -0,0 +1,43 @@
...
+    {
+        HDINLINE void operator()(uint32_t& destination, uint32_t const& source) const
+        {
+            destination = destination || source;
+        }
+
+        template<typename T_Worker>
+        HDINLINE void operator()(T_Worker const&, uint32_t& destination, uint32_t const& source) const
+        {
+            destination = static_cast<uint32_t>(static_cast<bool>(destination) || static_cast<bool>(source));
```
PrometheusPi, line 41: same question here

PrometheusPi (dismissed):
Just a few questions

include/pmacc/math/operation/And.hpp
BrianMarre: Done

include/pmacc/math/operation/Or.hpp
BrianMarre: Done

include/pmacc/math/operation/And.hpp
BrianMarre: I followed the examples of the general add/sub/mul, they use a template type instead of hard coded bit width, maybe that explains the difference in interface?

include/pmacc/math/operation/And.hpp
BrianMarre: valid point I should probably cast in both

BrianMarre: Fixed

include/pmacc/math/operation/Or.hpp
BrianMarre: Fixed

include/pmacc/math/operation/And.hpp
psychocoderHPC: @PrometheusPi is right, the second parameter should not be a reference. But if other operators are written equally we should keep it like that and change it independently.

include/pmacc/math/operation/And.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#pragma once
+
+#include "pmacc/types.hpp"
+
+#include <cstdint>
+
+namespace pmacc::math::operation
+{
+    struct And
```
psychocoderHPC, line 31: Please add a comment that this is a logical `AND`

include/pmacc/math/operation/Or.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#pragma once
+
+#include "pmacc/types.hpp"
+
+#include <cstdint>
+
+namespace pmacc::math::operation
+{
+    struct Or
```
psychocoderHPC, line 31: Please add a comment that this is a logical `OR`

psychocoderHPC (changes requested):

include/pmacc/math/operation/Or.hpp
BrianMarre: Done

include/pmacc/math/operation/And.hpp
BrianMarre: Done

psychocoderHPC (approved):

include/pmacc/math/operation/And.hpp
PrometheusPi: Okay, the let's do this later*

*=never ;)

PrometheusPi (approved):

PR #4919, opened 2024-05-31
improve physicalConstants.param

BrianMarre:
updates physical constants values to recommended SI values

### DIFF
--- a/include/picongpu/param/physicalConstants.param
+++ b/include/picongpu/param/physicalConstants.param
@@ -30,8 +30,13 @@ namespace picongpu
         /** unit: m / s */
         constexpr float_64 SPEED_OF_LIGHT_SI = 2.99792458e8;
 
-        /** unit: N / A^2 */
-        constexpr float_64 MUE0_SI = PI * 4.e-7;
+        /** vacuum magnetic permability
+         *
+         * unit: N / A^2
+         *
+         * 2022 CODATA Value, https://physics.nist.gov/cgi-bin/cuu/Value?mu0
+         */
+        constexpr float_64 MUE0_SI = 1.25663706127e-6;
         /** unit: C / (V m) */
         constexpr float_64 EPS0_SI = 1.0 / MUE0_SI / SPEED_OF_LIGHT_SI / SPEED_OF_LIGHT_SI;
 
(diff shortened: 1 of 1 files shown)


include/picongpu/param/physicalConstants.param
```diff
@@ -46,18 +46,22 @@ namespace picongpu
 
         // Electron properties
         /** unit: kg */
-        constexpr float_64 ELECTRON_MASS_SI = 9.109382e-31;
+        constexpr float_64 ELECTRON_MASS_SI = 9.1093837015e-31;
```
chillenzer:
```suggestion
        constexpr float_64 ELECTRON_MASS_SI = 9.1093837139e-31;
```
from [wikipedia](https://en.wikipedia.org/wiki/Electron_mass).

chillenzer: Okay. Also found [this](https://en.wikipedia.org/wiki/Atomic_units) which has the value you used. One of them must be lying... -.-

include/picongpu/param/physicalConstants.param
```diff
@@ -46,18 +46,22 @@ namespace picongpu
...
         /** unit: C */
         constexpr float_64 ELECTRON_CHARGE_SI = -1.602176e-19;
 
-        /* atomic unit for energy:
-         * 1 Rydberg = 27.21 eV --> converted to Joule
+        /** atomic unit for energy:
+         * 2 Rydberg = 27.21 eV --> converted to Joule
          */
         constexpr float_64 ATOMIC_UNIT_ENERGY = 4.36e-18;
```
chillenzer, line 64: Why is this two digits of precision while everything else has so high precision?

include/picongpu/param/physicalConstants.param
```diff
@@ -46,18 +46,22 @@ namespace picongpu
...
 
+        /** Rydberg energy, see https://en.wikipedia.org/wiki/Rydberg_constant, unit: eV
+         */
+        constexpr float_X RYDBERG_ENERGY = 13.605693122994_X; // eV
```
chillenzer:
```suggestion
        constexpr float_X RYDBERG_ENERGY = 13.605693122990_X; // eV
```
from the source you give yourself.

include/picongpu/param/physicalConstants.param
```diff
@@ -46,18 +46,22 @@ namespace picongpu
...
+
         /** bohr radius, unit: m
          */
-        constexpr float_64 BOHR_RADIUS = 5.292e-7;
+        constexpr float_64 BOHR_RADIUS = 5.29177210903e-11;
```
chillenzer:
```suggestion
        constexpr float_64 BOHR_RADIUS = 5.29177210544e-11;
```
from [here](https://en.wikipedia.org/wiki/Bohr_radius#Definition_and_value)

chillenzer (changes requested):
I've found different values for various of those constants. You should probably cite a paper for each of them.

include/picongpu/param/physicalConstants.param
BrianMarre: This value is used by the existing ionization implementation by marco, I would guess it was deemed good enough by him.

I am hesitant to touch them since I haven't looked in detail into the normalization of of marcos implementation, if we want to change them i would propose to do that in a separate PR.

include/picongpu/param/physicalConstants.param
BrianMarre: Done

include/picongpu/param/physicalConstants.param
BrianMarre: Done

include/picongpu/param/physicalConstants.param
```diff
@@ -30,8 +30,13 @@ namespace picongpu
         /** unit: m / s */
         constexpr float_64 SPEED_OF_LIGHT_SI = 2.99792458e8;
 
-        /** unit: N / A^2 */
-        constexpr float_64 MUE0_SI = PI * 4.e-7;
```
PrometheusPi, line 34: Cool - I did not know that they changed the standard in 2019.

include/picongpu/param/physicalConstants.param
PrometheusPi: I agree with @BrianMarre

include/picongpu/param/physicalConstants.param
```diff
@@ -40,24 +45,35 @@ namespace picongpu
...
-         * 1 Rydberg = 27.21 eV --> converted to Joule
+         * 2 Rydberg = 27.21 eV --> converted to Joule
          */
         constexpr float_64 ATOMIC_UNIT_ENERGY = 4.36e-18;
 
+        /** Rydberg energy, unit: eV
+         *
+         * 2022 CODATA value, https://physics.nist.gov/cgi-bin/cuu/Value?rydhcev
+         */
+        constexpr float_X RYDBERG_ENERGY = 13.605693122990_X; // eV
```
PrometheusPi: Why is this stored as `float_X` and not as `float_64`?

PrometheusPi (dismissed):
Looks good to me, but please answer the open question @BrianMarre.

include/picongpu/param/physicalConstants.param
BrianMarre: good catch, fixed

PrometheusPi (approved):

PR #4920, opened 2024-05-31
fix RiscV CMakeSupport

psychocoderHPC:
When compiling on a native RiscV system the the compiler flag `-mtune` cannot be set to the same value than `-march`.
This PR add handling for this behaviour.


ikbuibui (approved):

include/pmacc/PMaccConfig.cmake
```diff
@@ -256,6 +258,8 @@ elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel")
 elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
     if("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "aarch64")
         set(PMACC_CPU_ARCH_TEMPLATE "-mcpu={}")
+    elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "riscv64")
+        set(PMACC_CPU_ARCH_TEMPLATE "-march={}")
```
PrometheusPi, line 262: Just for my curiosity: what's the difference between `-march={}` and `mcpu={}`?

PrometheusPi (approved):

PR #4922, opened 2024-06-04
test libc6

BrianMarre:
tries to trigger a suspected bug in libc6


psychocoderHPC:
issue tried to detect with this PR is solved in #4923

PR #4827, opened 2024-02-23
Addition of Synchrotron radiation

FilipO28555:
Adding a Synchrotron Radiation extension based on paper:
["Extended particle-in-cell schemes for physics in ultrastrong laser fields: Review and developments" by A. Gonoskov et.Al.](https://core.ac.uk/reader/70613657?utm_source=linkout)

- [x] Perform runtime test
- [x] Extract parameters to .param file
- [x] Add automatic test 
- [x] Add documentation

PR #4923, opened 2024-06-04
fix SynchrotronRadiation

psychocoderHPC:
**IMPORTANT** Due to the bug fixed in this PR the current dev is not usable.

Fix  that the current dev is always failing with
```
terminate called after throwing an instance of 'char const*'
```
due to wrong usage of `throw`.
The second issue is that the stage is always consuming memory and is performing initializing helper fields based on the default param input. This is one reason why the first issue was showing even if the synchrotron extension is not activated.

Bug was introduced with #4827

### DIFF
--- a/include/picongpu/simulation/stage/SynchrotronRadiation.hpp
+++ b/include/picongpu/simulation/stage/SynchrotronRadiation.hpp
@@ -63,15 +63,18 @@ namespace picongpu::simulation::stage
         {
             if(xLeft == xMiddle)
             {
-                throw "SynchrotronRadiation.hpp: integrateAsExponential(): xLeft and xMiddle cannot be the same.";
+                throw std::runtime_error(
+                    "SynchrotronRadiation.hpp: integrateAsExponential(): xLeft and xMiddle cannot be the same.");
             }
             if(yLeft <= 0 || yMiddle <= 0)
             {
-                throw "SynchrotronRadiation.hpp: integrateAsExponential(): yLeft and yMiddle must be positive.";
+                throw std::runtime_error(
+                    "SynchrotronRadiation.hpp: integrateAsExponential(): yLeft and yMiddle must be > 0.");
             }
             if(yLeft == yMiddle)
             {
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
@PrometheusPi @FilipO28555 please check the default parameters in our synchrotron.param file because if we use these will will get the error

```
[1,0]<stdout>:Caught exception when precomputing firstSynchrotronFunction at 1192. index: SynchrotronRadiation.hpp: integrateAsExponential(): yLeft and yMiddle must be positive.

```

PrometheusPi (approved):

PrometheusPi:
When running the Synchrotron test, I get: 
```
﻿﻿calculation  simulation time:  1min  4sec 925msec = 64.925 sec
full simulation time:  1min 22sec 135msec = 82.135 sec
Validate!
PIConGPU prefix = ...PIConGPU/_pr_4923/picongpu/
Load buildsystem/CompileSuite/color.sh from PIConGPU!
Execute validate.py with data from ...PIConGPU/_pr_4923/projects/Synchrotron/simOutput
Validating data in ...PIConGPU/_pr_4923/projects/Synchrotron/simOutput
/trinity/shared/pkg/devel/python/3.10.4/lib/python3.10/site-packages/scipy/integrate/_quadpack_py.py:463: ComplexWarning: Casting complex values to real discards the imaginary part
  return _quadpack._qagse(func,a,b,args,full_output,epsabs,epsrel,limit)
...PIConGPU/_pr_4923/projects/Synchrotron/./lib/validate.py:98: IntegrationWarning: The occurrence of roundoff error is detected, which prevents
  the requested tolerance from being achieved.  The error may be
  underestimated.
  analytical_integrated.append(quad(lambda x: analytical_Propability(x, gamma, Heff, dt), x0, x1)[0])
Poorness: 1428967203.6987355
[testSuite] Synchrotron radiation plugin has bad performance!
[testSuite] Check Synchrotron plugin implementation!

```

PrometheusPi:
The failing synchrotron run time test will be fixed after this pull request.

PrometheusPi:
Both the synchrotron test and the LWFA example run fine. (Only the check of the synchrotron test fails.)

Issue #4921, opened 2024-05-31
Enforce filling gaps


chillenzer:
In https://github.com/ComputationalRadiationPhysics/picongpu/pull/4916#discussion_r1622374787 I've stumbled across a commented warning to always fill the gaps after an operation. If necessary, we should have a mechanism to enforce this. Theoretically, this could be done via compile-time list operations and tagging of the corresponding manipulators. Practically, we will  probably wait until we have the init pipeline at runtime which will make this kind of operation easier to implement.

Issue #4912, opened 2024-05-29
Enhance Synchrotron extension 


PrometheusPi:
The pull request #4827 adds/added the synchrotron radiation module. 

There were some enhancements suggested by @chillenzer that did not go to the first pull request. 
This is a list of this suggested enhancements for a future implementation. 

- [ ] [open todo](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1619023968) to test speed of logarithmic interpolation 
- [ ] [open question](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609607742) to @psychocoderHPC 
- [ ] [fix python module version](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609721221) in `requirements.txt`
- [ ] [naming](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609723259) `SynchrotronIdea` might be too unspecific 
- [ ] [too long function](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609748347) in `AlgorithmsSynchrotron.hpp` 
- [x] use `pathlib` instead of `os.path` [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609863570) and [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609864862)
- [ ] [question on usage](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609879390) of `change_parameters_randomly.py`
- [x] use [`main` function](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609881468) instead of a script like code
- [x] [question on comment](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609882515) regarding output 
- [x] use of `np.asarray` [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609891973) and [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609899282)
- [x] [question on parameters](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609902008) in `validate.py`
[...]

PR #4926, opened 2024-06-05
Changes to Synchrotron Radiation from issue #4921. Part 1.

FilipO28555:
in validate():
- [x] subtract_functions() is now without a bug and unnecessary list truncation
- [x] Changed the number of bins to 200 for better accuracy

Todos from [issue #4912](https://github.com/ComputationalRadiationPhysics/picongpu/issues/4912):
- [x] use `pathlib` instead of `os.path` [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609863570) and [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609864862)
- [x] [question on usage](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609879390) of `change_parameters_randomly.py`
- [x] use [`main` function](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609881468) instead of a script like code
- [x] [question on comment](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609882515) regarding output 
- [x] use of `np.asarray` [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609891973) and [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609899282)
- [x] [question on parameters](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609902008) in `validate.py`
- [x] use seperate functions in python [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609908954) and [here](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609909367)
- [x] better way to handle [input arguments](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4827#discussion_r1609940439)

### DIFF
--- a/share/picongpu/tests/Synchrotron_plugin/lib/change_parameters_randomly.py
+++ b/share/picongpu/tests/Synchrotron_plugin/lib/change_parameters_randomly.py
@@ -1,19 +1,20 @@
-import os
 import numpy as np
 from synchrotron_lib import calculate_dt, predictNumPhotons
+from pathlib import Path
 # dt is in grid.param as DELTA_T_SI
 # Heff is in fieldBackground.param as field_Strength_SI
 # gamma is in particle.param as gamma
 # all are in ./include/picongpu/param
 
-paramsPath = os.path.join(os.path.dirname(os.path.realpath(__file__)), "../include/picongpu/param")
+paramsPath = Path(__file__).absolute().parent / "../include/picongpu/param"
 
 
 # read the files change the values and write them back
 def change_param_file(file_name, param_name, value):
-    with open(os.path.join(paramsPath, file_name), "r") as f:
+    filePath = paramsPath / file_name
+    with open(filePath, "r") as f:
         lines = f.readlines()
-    with open(os.path.join(paramsPath, file_name), "w") as f:
+    with open(filePath, "w") as f:
         for line in lines:
             if param_name + " = " in line:
                 changedLine = line.split("=")[0] + "= " + value + ";\n"
@@ -23,6 +24,7 @@ def change_param_file(file_name, param_name, value):
                 f.write(line)
 
 
+# change the three parameters
 def changeParams(dt, Heff, gamma):
     change_param_file("grid.param", "DELTA_T_SI", dt)
     change_param_file("fieldBackground.param", "field_Strength_SI", Heff)
@@ -31,25 +33,30 @@ def changeParams(dt, Heff, gamma):
 
 # the code checks if we have enough photons generated for the given parameters:
 # gamma, Heff can be any value, dt is calculated based on gamma and Heff
-yNum = 0
-while yNum < 5e6:
-    gamma = [10, 50, 100, 500, 1000]  # include all gamma values you want to check
-    Heff = [1e13, 1e14, 1e15, 1e16, 1e17, 1e18]  # similarly with Heff
+def main():
+    yNum = 0
+    while yNum < 5e6:
+        gamma = [10, 50, 100, 500, 1000]  # include all gamma values you want to check
+        Heff = [1e13, 1e14, 1e15, 1e16, 1e17, 1e18]  # similarly with Heff
 
-    gamma = gamma[np.random.randint(len(gamma))]
-    Heff = Heff[np.random.randint(len(Heff))]
-    dt = calculate_dt(gamma, Heff) * 0.95
-    dt = dt if dt < 1e-16 else 1e-16  # if dt is larger than the grid condition, set it to the grid condition
-    yNum = predictNumPhotons(gamma, Heff, dt, 4000, 5e5)
+        gamma = gamma[np.random.randint(len(gamma))]
+        Heff = Heff[np.random.randint(len(Heff))]
+        dt = calculate_dt(gamma, Heff) * 0.95
+        dt = dt if dt < 1e-16 else 1e-16  # if dt is larger than the grid condition, set it to the grid condition
+        yNum = predictNumPhotons(gamma, Heff, dt, 4000, 5e5)
 
-# make Heff, gamma and dt strings in scientific notation
-Heff = "{:.2e}".format(Heff)
-gamma = "{:.3e}".format(gamma)
-dt = "{:.2e}".format(dt)
-changeParams(dt, Heff, gamma)
-print("gamma: ", gamma, "Heff: ", Heff, "dt: ", dt)
-print(f"Predicted number of photons: {yNum}")
+    # make Heff, gamma and dt strings in scientific notation
+    Heff = "{:.2e}".format(Heff)
+    gamma = "{:.3e}".format(gamma)
+    dt = "{:.2e}".format(dt)
+    changeParams(dt, Heff, gamma)
+    print("gamma: ", gamma, "Heff: ", Heff, "dt: ", dt)
+    print(f"Predicted number of photons: {yNum}")
 
-# save parameters in a file in the output directory ./RUN{runNum}
-with open("./simOutput/params.txt", "w") as f:
-    f.write(f"gamma: {gamma}, Heff: {Heff}, dt: {dt}")
+    # save parameters in a file in the output directory ./simOutput
+    with open("./simOutput/params.txt", "w") as f:
+        f.write(f"gamma: {gamma}, Heff: {Heff}, dt: {dt}")
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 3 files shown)


share/picongpu/tests/Synchrotron_plugin/lib/change_parameters_randomly.py
```diff
@@ -23,6 +24,7 @@ def change_param_file(file_name, param_name, value):
                 f.write(line)
 
 
+# change the three parameters
```
chillenzer: Remove. That is redundant.

share/picongpu/tests/Synchrotron_plugin/lib/requirements.txt
```diff
@@ -1,4 +1,5 @@
 os
-numpy
-scipy
-openpmd_api
+numpy==1.26.4
+scipy==1.11.4
+pathlib
```
chillenzer: `pathlib` is part of the standard library, so we don't need to include it.

share/picongpu/tests/Synchrotron_plugin/lib/validate.py
```diff
@@ -7,15 +7,8 @@
...
 def subtract_functions(x0, y0, x1, y1):
-    minX = max(np.min(x0), np.min(x1))
-    maxX = min(np.max(x0), np.max(x1))
-    #
-    x0 = [x for x in x0 if x >= minX and x <= maxX]
-    x1 = [x for x in x1 if x >= minX and x <= maxX]
-    y0 = [y for x, y in zip(x0, y0) if x >= minX and x <= maxX]
-    y1 = [y for x, y in zip(x1, y1) if x >= minX and x <= maxX]
-
     y1 = np.interp(x0, x1, y1)  # interpolate y1 to the x0 values
```
chillenzer: Okay, so you are saying that you are fine with the default behaviour of `np.interp` here? I would have assumed that for histograms using `left=0, right=0` would have been the most faithful extrapolation.

share/picongpu/tests/Synchrotron_plugin/lib/validate.py
```diff
@@ -38,57 +31,84 @@ def read_setup(folderName):
...
     series.flush()
     p_y = np.float64(p_y)
     p_y *= p_y_SI / w
-    hist_data = np.abs(p_y)
+
+    # Calculate photon energy
+    hist_data = np.abs(p_y * const.c / const.elementary_charge)
+
     if len(hist_data) < 5e5:
         raise SystemExit(
```
chillenzer: Just noticed: Do we really want to raise `SystemExit` here? Raising something a little milder like `ValueError` or a custom `Exception` would have the same effect for running this script. But in case anyone ever tries to build something on top of this, a `ValueError` could be caught and handled appropriately.

share/picongpu/tests/Synchrotron_plugin/lib/validate.py
```diff
@@ -38,57 +31,84 @@ def read_setup(folderName):
...
+
+    # Reading electron weights
     h1 = e["weighting"][io.Mesh_Record_Component.SCALAR]
     e_w = h1[:]  # get the weighting
     e_w_SI = h1.unit_SI  # get the weighting unit
     series.flush()
     e_w *= e_w_SI  # convert to SI units
+
+    # Reading electron momentum
     h1 = e["momentum"]["y"]
```
chillenzer, line 76: Just to be extra sure: The `"y"` in this line has nothing to do with the `"y"` that approximates a $\gamma$ and stands for photons further up, right? It's probably the direction of the momentum?

share/picongpu/tests/Synchrotron_plugin/lib/validate.py
```diff
@@ -114,11 +134,12 @@ def main(dataPath):
     analytical_integrated = np.array(analytical_integrated)[mask]
 
     x, y = subtract_functions(delta, analytical_integrated, b, a)
-    poorness = np.sum(y) / len(y)  # average error
+    poorness = np.sum(y) / len(y)  # average relative error
```
chillenzer: Use `np.mean` (potentially with `axis=0`) here.

chillenzer (commented):
Hi, looks good. Thanks for going the extra mile to make this a little more sustainable code. In what sense are the questions answered? They are already ticked but that information has not propagated to me. So, it is likely that whatever method you used to answer them, does not help the uninvolved future reader of your code.

share/picongpu/tests/Synchrotron_plugin/lib/validate.py
FilipO28555: yes, you are right

share/picongpu/tests/Synchrotron_plugin/lib/change_parameters_randomly.py
FilipO28555: done

share/picongpu/tests/Synchrotron_plugin/lib/requirements.txt
FilipO28555: Done

FilipO28555: The same with os, right?

share/picongpu/tests/Synchrotron_plugin/lib/validate.py
FilipO28555: You are right. I did that. I also changed the name to be more descriptive

share/picongpu/tests/Synchrotron_plugin/lib/validate.py
FilipO28555: I changed it to ValueError

share/picongpu/tests/Synchrotron_plugin/lib/validate.py
FilipO28555: Yes

share/picongpu/tests/Synchrotron_plugin/lib/validate.py
FilipO28555: We left it because in the future we might do something like np.sum(y*np.sqrt(a)) / len(y)

share/picongpu/tests/Synchrotron_plugin/lib/requirements.txt
chillenzer: Yes

share/picongpu/tests/Synchrotron_plugin/lib/validate.py
chillenzer: That's still a mean isn't it? `np.mean(y*np.sqrt(a))`

chillenzer (approved):
Good work!

share/picongpu/tests/Synchrotron_plugin/lib/validate.py
PrometheusPi: I agree with @chillenzer

share/picongpu/tests/Synchrotron_plugin/lib/validate.py
```diff
@@ -113,17 +133,18 @@ def main(dataPath):
...
     retValue = int(0) if poorness <= poornessBound else int(1)
+    print(f"Test {'passed' if retValue == 0 else 'failed'}")
 
     sys.exit(retValue)
 
 
 if __name__ == "__main__":
     if len(sys.argv[1:]) != 1:
-        raise SystemExit(f"Usage: {sys.argv[0]} <path_to_simulation_data>")
+        raise ValueError(f"Usage: {sys.argv[0]} <path_to_simulation_data>")
```
PrometheusPi: Just a question: why this change?

PrometheusPi (changes requested):
Just some minor requests.

share/picongpu/tests/Synchrotron_plugin/lib/validate.py
chillenzer: I guess that was in response to [this comment](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4926#discussion_r1629120358) applying to a different exception in the same script. The change you are referring to will most likely not make any difference at all and one could argue that not doing it might convey the intent a tiny bit better. So, weakly agreeing with you but probably doesn't matter.

share/picongpu/tests/Synchrotron_plugin/lib/validate.py
PrometheusPi: and why is this now "relative" ?

FilipO28555: It was always relative.
Also changed to np.mean()

PrometheusPi (approved):

PR #4927, opened 2024-06-06
fix 2D dispersive pulse

psychocoderHPC:
Fix the B-field implementation of the dispersive pulse for 2D simulations.
An invalid vector component was accessed and the result is an invalid memory access.

### DIFF
--- a/include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
+++ b/include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
@@ -149,7 +149,9 @@ namespace picongpu
                          * @param totalCellIdx cell index in the total domain (including all moving window slides)
                          * @param Omega frequency for which the E-value is calculated
                          */
-                        HDINLINE float_X amp(floatD_X const& totalCellIdx, float_X const Omega) const
+                        template<uint32_t T_dim>
+                        HDINLINE float_X
+                        amp(math::Vector<float_X, T_dim> const& totalCellIdx, float_X const Omega) const
                         {
                             // transform to 3d internal coordinate system
                             float3_X pos = this->getInternalCoordinates(totalCellIdx);
@@ -215,8 +217,11 @@ namespace picongpu
                             return mag;
                         }
 
+                        template<uint32_t T_dim>
                         HDINLINE float_X
-                        phi(floatD_X const& totalCellIdx, float_X const Omega, float_X const phaseShift) const
+                        phi(math::Vector<float_X, T_dim> const& totalCellIdx,
+                            float_X const Omega,
+                            float_X const phaseShift) const
                         {
                             // transform to 3d internal coordinate system
                             float3_X pos = this->getInternalCoordinates(totalCellIdx);
@@ -276,7 +281,6 @@ namespace picongpu
                             return phase;
                         }
 
-                    private:
                         /** Get value of E field in time domain for the given position, using DFT
                          * Interpolation order of DFT given via timestep in grid.param and INIT_TIME
                          * Neglecting the constant part of DFT (k = 0) because there should be no constant field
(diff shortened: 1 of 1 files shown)


PrometheusPi (dismissed):

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -215,8 +215,11 @@ namespace picongpu
                             return mag;
                         }
 
+                        template<uint32_t T_dim>
                         HDINLINE float_X
-                        phi(floatD_X const& totalCellIdx, float_X const Omega, float_X const phaseShift) const
+                        phi(math::Vector<float_X, T_dim> const& totalCellIdx,
```
steindev, line 222: You can stick with the `floatD_X` type, since totalCellIdx is only used to obtain the `pos` which takes `float2_X` and `float3_X`.

(by the way, you also did not change the signature of the `amp()` in line 152.)

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -285,7 +287,9 @@ namespace picongpu
                          * @param phaseShift additional phase shift to add on top of everything else,
                          *                   in radian
                          */
-                        HDINLINE float_X getValueE(floatD_X const& totalCellIdx, float_X const phaseShift) const
+                        template<uint32_t T_dim>
+                        HDINLINE float_X
+                        getValueE(math::Vector<float_X, T_dim> const& totalCellIdx, float_X const phaseShift) const
```
steindev, line 294: Same comment as for `phi()` above.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -405,13 +409,13 @@ namespace picongpu
...
+                            float3_X iIdxForw = makeVector3(totalCellIdx);
                             iIdxForw[i] = totalCellIdx[i] + 0.5_X;
-                            float3_X iIdxBackw = float3_X(totalCellIdx[0], totalCellIdx[1], totalCellIdx[2]);
+                            float3_X iIdxBackw = makeVector3(totalCellIdx);
                             iIdxBackw[i] = totalCellIdx[i] - 0.5_X;
-                            float3_X jIdxForw = float3_X(totalCellIdx[0], totalCellIdx[1], totalCellIdx[2]);
+                            float3_X jIdxForw = makeVector3(totalCellIdx);
                             jIdxForw[j] = totalCellIdx[j] + 0.5_X;
-                            float3_X jIdxBackw = float3_X(totalCellIdx[0], totalCellIdx[1], totalCellIdx[2]);
+                            float3_X jIdxBackw = makeVector3(totalCellIdx);
```
steindev, line 420: Well this might be one of the problems this implementation has and why it did not work. Actually, I think the best solution would be to just delete `struct DispersivePulseFunctorIncidentB` completely.
It is not used anyway.
That is, there is no fix required, but a cleanup :sweat_smile:

PrometheusPi: @steindev What should I make out of your comments? They are not rejecting, but sound like you actually would like to requite changes. This pull request could be merged now if you do not object.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
psychocoderHPC: We call `phi` from B field calculation always with float3_X now. Therefore `floatD_X` is wrong but you are right we need to change `amp()` too.

psychocoderHPC: fixed

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
psychocoderHPC: I opened #4932 where the B-field implementation is deleted. Please merge this or the other PR.

steindev:
Replaced by #4932.

PR #4929, opened 2024-06-06
Improved checkpoint editing tool

nicowrobel:
Changes were made to the checkpoint editing script, added in my last [pull request](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4520). Mainly these should improve memory efficiency as the original script needed to hold the entire copied checkpoint in memory until flushing.

This is done by:
- flushing after each record set, so the memory consumption is roughly capped by the largest record set
- explicitly writing to a `bp5` file, as ADIOS2 else does not give the option to free up memory after flushing
- giving the option to not copy the RNG values in the `RNGProvider3XorMin` field as these take up roughly half of the checkpoint size and are not needed for a restart

Further changes include some restructuring of parts of the code, compatibility with future `openPMD-api` versions and smaller fixes.

Future pull requests might further improve speed and memory consumption by exploiting the block structure of records and parallelization with `mpi`.

Thanks to @PrometheusPi and @franzpoeschel for the help with this commit.


lib/python/picongpu/extra/input/bunchInit_openPMD_bp.py
```diff
@@ -422,7 +420,7 @@ def __init__(
         outfile,
         particles=[],
         inconfig="{}",
-        outconfig="{}",
+        outconfig="adios2.engine.parameters.BufferChunkSize = 2147381248",
```
PrometheusPi, line 439: Where does this number come from?
Please add a comment on the number.

lib/python/picongpu/extra/input/createBunch_example.ipynb
```diff
@@ -230,16 +228,19 @@
...
    "outputs": [],
    "source": [
-    "# assigne to (existing) checkpoint file\n",
+    "# assign to (existing) checkpoint file\n",
     "# replace with your own paths and species name (from speciesDefinition.param of the input simulation)\n",
+    "# copyRNG can be used to skip the copying of the RNG values\n",
     "checkPoint_b = addParticles2Checkpoint(\n",
-    "    \"/bigdata/hplsim/production/wrobel45/PWFA-bunch-script_test/001_empty_test/simOutput/checkpoints/checkpoint_%T.bp\",\n",
-    "    \"/bigdata/hplsim/production/wrobel45/PWFA-bunch-script_test/002_bunch_test/simOutput/checkpoints/checkpoint_%06T.bp\",\n",
+    "    \"/bigdata/hplsim/production/wrobel45/PWFA_script/script_empty_021/simOutput/checkpoints/checkpoint_%T.bp5\",\n",
```
PrometheusPi: Should we check in real paths? 
Could you clean them up to something not working but readable?

lib/python/picongpu/extra/input/createBunch_example.ipynb
```diff
@@ -200,7 +198,7 @@
     "\n",
     "# shift particles to new center position\n",
     "x_PIConGPU = x + center_pos_x\n",
-    "y_PIConGPU = y - np.mean(y) + center_pos_y\n",
+    "y_PIConGPU = y + center_pos_y\n",
```
PrometheusPi, line 201: Just a question:
Why this change?

PrometheusPi (changes requested):
Only minor requests and a question

lib/python/picongpu/extra/input/bunchInit_openPMD_bp.py
nicowrobel: @franzpoeschel added this parameter when overlooking the script.
In the [ADIOS2 Documentation](https://adios2.readthedocs.io/en/latest/engines/engines.html#bp5) it is stated that it can be used to increase write performance:

> BufferChunkSize: (for chunk buffer type) The size of each memory buffer chunk, default is 128MB but it is worth increasing up to 2147381248 (a bit less than 2GB) if possible for maximum write performance.

I didn't notice the change but I can add a comment and check if it makes a difference in performance.

lib/python/picongpu/extra/input/createBunch_example.ipynb
nicowrobel: It seemed redundant to me as the bunch is already distributed around 0 in the example script so I threw it away for simplicity.

PrometheusPi: Thanks for the clarification

lib/python/picongpu/extra/input/bunchInit_openPMD_bp.py
franzpoeschel: I think there's no need to check the performance as in this script it likely makes no difference. Given the data sizes that we're dealing with here, there's just no use in *not* setting this parameter to the maximum possible value, especially since on Linux systems there is basically no downside due to how those systems handle virtual memory. Linking the ADIOS2 documentation here should be enough.

nicowrobel: I can verify that it makes no difference to the performance but I will keep it and add a comment.

franzpoeschel: I mostly expect that it could make a difference if we switch to writing smaller block sizes

PrometheusPi (approved):

PR #4930, opened 2024-06-06
Increase performance of DispersivePulse

steindev:
Shorten initialization time by reducing the number of frequencies being used in the discrete Fourier transform of the pulse from frequency-space domain to time-space domain. Usually, not all frequencies in the range Omega=0 to pi/dt will contribute to the field since the laser's spectrum is centered narrowly around the central laser frequency. Thus, the DFT is limited to the frequencies within +/- 4*sigma around the central laser frequency.

### DIFF
--- a/include/picongpu/fields/incidentField/profiles/DispersivePulse.def
+++ b/include/picongpu/fields/incidentField/profiles/DispersivePulse.def
@@ -55,6 +55,16 @@ namespace picongpu
                          */
                         static constexpr float_64 PULSE_INIT = 20.0;
 
+                        /** Width of the spectral support for the discrete Fourier transform of the pulse's field
+                         * from frequency-space domain to time-space domain.
+                         * Defined as number of sigmas of the laser pulse spectrum.
+                         * NOTE: A too small value will artificially increase the pulse duration and
+                         * too high values will increase the the time per step during initialization of the pulse.
+                         *
+                         * unit: none
+                         */
+                        static constexpr float_X SPECTRAL_SUPPORT = 4._X;
+
                         // Dispersion Parameters
 
                         /** SD: spatial dispersion in focus
--- a/include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
+++ b/include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
@@ -57,6 +57,9 @@ namespace picongpu
                         //! Base unitless parameters
                         using Base = BaseParamUnitless<T_Params>;
 
+                        // unit: none
+                        using Params::SPECTRAL_SUPPORT;
+
                         // unit: UNIT_LENGTH
                         static constexpr float_X W0 = static_cast<float_X>(Params::W0_SI / UNIT_LENGTH);
 
@@ -295,25 +298,49 @@ namespace picongpu
                             else if(time > Unitless::INIT_TIME)
                                 return 0.0_X;
 
-                            // timestep also in UNIT_TIME
-                            float_X const dt = static_cast<float_X>(picongpu::SI::DELTA_T_SI / UNIT_TIME);
                             // interpolation order
-                            float_X N_raw = Unitless::INIT_TIME / dt;
-                            int n = static_cast<int>(N_raw * 0.5_X); // -0 instead of -1 for rounding up N_raw
+                            float_X N_raw = Unitless::INIT_TIME / DELTA_T;
+                            int const n = static_cast<int>(N_raw * 0.5_X); // -0 instead of -1 for rounding up N_raw
 
-                            float_X E_t = 0.0_X; // electric field in time-domain
-                            float_X Omk = 0.0_X; // stores angular frequency for DFT-loop; Omk= 2pi*k/T
+                            // frequency step for DFT
+                            float_X const dOmk = pmacc::math::Pi<float_X>::doubleValue / Unitless::INIT_TIME;
 
-                            for(int k = 1; k < n + 1; k++)
+                            // Since the (Gaussian) spectrum has only significant values near the central frequency,
+                            // the summation over all frequencies is reduced to frequencies within an interval
+                            // around the central frequency.
+
+                            // standard deviation of the Gaussian distributed spectrum
+                            float_X const sigma_Om = 1._X / (pmacc::math::sqrt(2._X) * Unitless::PULSE_DURATION);
+
+                            // index of the mean frequency of the Gaussian distributed spectrum
+                            // unit: [dOmk]
+                            int const center_k
+                                = static_cast<int>(SPEED_OF_LIGHT * Unitless::INIT_TIME / Unitless::Base::WAVE_LENGTH);
+
+                            // index of the lowest frequency in the Gaussian distributed spectrum which is used in the
+                            // DFT 4*sigma_Om distance from central frequency
+                            int const minOm_k
+                                = center_k - static_cast<int>(Unitless::SPECTRAL_SUPPORT * sigma_Om / dOmk);
+                            int const k_min = math::max(minOm_k, 1);
+
+                            // index of the highest frequency in the Gaussian distributed spectrum which is used in the
+                            // DFT
+                            int const maxOm_k = 2 * center_k - minOm_k;
+                            int const k_max = math::min(maxOm_k, n);
+
+                            // electric field in time-domain
+                            float_X E_t = 0.0_X;
+
+                            for(int k = k_min; k <= k_max; k++)
                             {
-                                Omk = static_cast<float_X>(k) * pmacc::math::Pi<float_X>::doubleValue
-                                    / Unitless::INIT_TIME;
+                                // stores angular frequency for DFT-loop
+                                float_X const Omk = static_cast<float_X>(k) * dOmk;
                                 float_X sinPhi, cosPhi;
                                 float_X sinOmkt, cosOmkt;
                                 pmacc::math::sincos(phi(totalCellIdx, Omk, phaseShift), sinPhi, cosPhi);
                                 pmacc::math::sincos(Omk * time, sinOmkt, cosOmkt);
-                                E_t += 2.0_X * amp(totalCellIdx, Omk) * cosPhi / dt * cosOmkt
-                                    + 2.0_X * amp(totalCellIdx, Omk) * sinPhi / dt * sinOmkt;
+                                E_t += 2.0_X * amp(totalCellIdx, Omk) * (cosPhi * cosOmkt + sinPhi * sinOmkt)
+                                    / DELTA_T;
                             }
 
                             E_t /= static_cast<float_X>(2 * n + 1); // Normalization from DFT


include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -299,15 +299,39 @@ namespace picongpu
...
+
+                            // frequency step for DFT
+                            float_X const dOmk = pmacc::math::Pi<float_X>::doubleValue / Unitless::INIT_TIME;
+
+                            // Since the (Gaussian) spectrum has only significant values near the central frequency,
+                            // the summation over all frequencies is reduced to frequencies within an interval
+                            // around the central frequency.
+
+                            // standard deviation of the Gaussian distributed spectrum
+                            float_X const sigma_Om = 1._X / (pmacc::math::sqrt(2._X) * Unitless::PULSE_DURATION);
```
finnolec, line 313: Where does this come from? Should this not be `2 * pi * 0.441 / Unitless::Unitless::PULSE_DURATION` for a bandwidth limited pulse?

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -299,15 +299,39 @@ namespace picongpu
...
+
+                            // index of the mean frequency of the Gaussian distributed spectrum
+                            // unit: [dOmk]
+                            int const center_k
+                                = static_cast<int>(SPEED_OF_LIGHT * Unitless::INIT_TIME / Unitless::Base::WAVE_LENGTH);
+
+                            // index of the lowest frequency in the Gaussian distributed spectrum which is used in the
+                            // DFT 4*sigma_Om distance from central frequency
+                            int const minOm_k = center_k - static_cast<int>(4._X * sigma_Om / dOmk);
```
finnolec: Do we want to hard-code the `4._X` here or should it not rather be a parameter in the `incidentField.param`?

psychocoderHPC: it should be part of the dispersive pulse

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
fabidie: It should be right, since it is exactly the standard deviation used a few lines  above (184) to define the longitudinal (gaussian) envelope of the pulse in the frequency domain.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -299,15 +299,39 @@ namespace picongpu
...
+                            int const k_min = minOm_k < 1 ? 1 : minOm_k;
+
+                            // index of the highest frequency in the Gaussian distributed spectrum which is used in the
+                            // DFT
+                            int const maxOm_k = 2 * center_k - minOm_k;
+                            int const k_max = maxOm_k <= n ? maxOm_k : n;
```
psychocoderHPC: should be `int const k_max = math::min(maxOm_k,n);`

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -299,15 +299,39 @@ namespace picongpu
...
+
+                            // index of the mean frequency of the Gaussian distributed spectrum
+                            // unit: [dOmk]
+                            int const center_k
+                                = static_cast<int>(SPEED_OF_LIGHT * Unitless::INIT_TIME / Unitless::Base::WAVE_LENGTH);
+
+                            // index of the lowest frequency in the Gaussian distributed spectrum which is used in the
+                            // DFT 4*sigma_Om distance from central frequency
+                            int const minOm_k = center_k - static_cast<int>(4._X * sigma_Om / dOmk);
+                            int const k_min = minOm_k < 1 ? 1 : minOm_k;
```
psychocoderHPC: should be 

```C++
int const k_min = math::min(n,math::max(1,minOm_k));
```

This avoids that min is larger max.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -299,15 +299,39 @@ namespace picongpu
...
+                            // the summation over all frequencies is reduced to frequencies within an interval
+                            // around the central frequency.
+
+                            // standard deviation of the Gaussian distributed spectrum
+                            float_X const sigma_Om = 1._X / (pmacc::math::sqrt(2._X) * Unitless::PULSE_DURATION);
+
+                            // index of the mean frequency of the Gaussian distributed spectrum
+                            // unit: [dOmk]
+                            int const center_k
+                                = static_cast<int>(SPEED_OF_LIGHT * Unitless::INIT_TIME / Unitless::Base::WAVE_LENGTH);
```
fabidie, line 318: Just a minor and maybe irrelevant detail, but for correct rounding one could add +0.5

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
finnolec: Yes but it could be part of the `DispersivePulseParam` in case users want to play around with this value. Since it might confuse people, it could still have a default value of 4.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -299,15 +299,39 @@ namespace picongpu
...
+                            // DFT
+                            int const maxOm_k = 2 * center_k - minOm_k;
+                            int const k_max = maxOm_k <= n ? maxOm_k : n;
 
                             float_X E_t = 0.0_X; // electric field in time-domain
-                            float_X Omk = 0.0_X; // stores angular frequency for DFT-loop; Omk= 2pi*k/T
+                            float_X Omk = 0.0_X; // stores angular frequency for DFT-loop; Omk = 2pi*k/T = dOmk * k
 
-                            for(int k = 1; k < n + 1; k++)
+                            for(int k = k_min; k < k_max + 1; k++)
```
fabidie: Did you check the output especially for high AD and/or TOD values? Since they scale with $(\Omega - \Omega_0)^3$, there is a high contribution away from `center_k` (definitely less than the exponential decay of the envelope, nevetheless I just want to be sure)

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
finnolec: Hm, it's some different definition of pulse duration here than in the shadowgraphy plugin or in my post-processing. I can't find this right now, but if this here is self-consistent and producing pulses with the correct duration it should be right

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -299,15 +299,39 @@ namespace picongpu
...
+                            int const k_min = minOm_k < 1 ? 1 : minOm_k;
+
+                            // index of the highest frequency in the Gaussian distributed spectrum which is used in the
+                            // DFT
+                            int const maxOm_k = 2 * center_k - minOm_k;
+                            int const k_max = maxOm_k <= n ? maxOm_k : n;
 
                             float_X E_t = 0.0_X; // electric field in time-domain
-                            float_X Omk = 0.0_X; // stores angular frequency for DFT-loop; Omk= 2pi*k/T
+                            float_X Omk = 0.0_X; // stores angular frequency for DFT-loop; Omk = 2pi*k/T = dOmk * k
```
psychocoderHPC: please move the definition into the loop because the value is not required outside.

psychocoderHPC (changes requested):

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
psychocoderHPC: could it be that it must be `<= k_max`?

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
psychocoderHPC: IMO `center_k` should be `float_X`, and later `min_k` should be a floor() if the floating point-min and `max_k` should be a round-up of the corresponding float value.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
psychocoderHPC: ohh I see you have `k < k_max + 1` what is equal to `k <= k_max`. I prefere the second version.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
steindev: Well, the exact number really doesn't matter. And `center_k` is used repeatedly and therefore should have the correct type already.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
steindev: @fabidie AD or TOD does not change the spectral width. Spectral width is only determined by the pulse duration and this is used to calculate the spectral width. Therefore all should be fine the way it is?

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
psychocoderHPC:
```C++
int const k_min = math::max(1,minOm_k);
```

Is enough because we sue this value in the for loop only.

steindev: Using only max is fine.

include/picongpu/fields/incidentField/profiles/DispersivePulse.def
```diff
@@ -55,6 +55,16 @@ namespace picongpu
...
 
+                        /** Width of the spectral support for the discrete Fourier transform of the pulse's field
+                         * from frequency-space domain to time-space domain.
+                         * Defined as number of sigmas of the laser pulse spectrum.
+                         * NOTE: A too small value will artificially increase the pulse duration and
+                         * too high values will increase the the time per step during initialization of the pulse.
+                         *
+                         * unit: none
+                         */
+                        static constexpr float_X SPECTRAL_SUPPORT = 4._X;
```
finnolec, line 66: I did originally think about doing this the SFINAE approach like in 

https://github.com/ComputationalRadiationPhysics/picongpu/blob/860f433600a956d8dd21236e53c550454d0199b9/include/picongpu/fields/incidentField/profiles/BaseParam.def#L163 

so that a user really doesn't have to touch it at all and it doesn't break any existing workflows.

finnolec: https://github.com/ComputationalRadiationPhysics/picongpu/blob/860f433600a956d8dd21236e53c550454d0199b9/include/picongpu/fields/incidentField/profiles/BaseParam.hpp#L99

psychocoderHPC: IMO the current user base is so small that we should enforce the value.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -57,6 +57,9 @@ namespace picongpu
                         //! Base unitless parameters
                         using Base = BaseParamUnitless<T_Params>;
 
+                        // unit: none
+                        using SPECTRAL_SUPPORT = Params::SPECTRAL_SUPPORT;
```
psychocoderHPC: should be

```C++
using Params::SPECTRAL_SUPPORT;
```

steindev:
@finnolec @psychocoderHPC If you approve and/or resolve the the conversations, I will squash and perform the final push.

finnolec (approved):

psychocoderHPC (approved):

psychocoderHPC:
@steindev Is this now mergeable?

steindev:
@psychocoderHPC Squash and rebase performed. PR Is mergeable now. (Once the CI is finished)

JessicaTiebel:
I tested and compared this improvement with a simulation before these changes with very satisfactory results.

The average time per step during initialisation was:
- before the improvement ~1600ms
- after the improvement ~90 ms (approx. as fast as after initialisation)

I used :
- SPECTRAL_SUPPORT=4
- Lehe solver

Thank's @steindev for this massiv improvement 🎉

PR #4931, opened 2024-06-07
constructor for `math::Vector` with generator

psychocoderHPC:
- add constructor for generator expressions
- add tiny unit test

```C++
constexpr auto vec = pmacc::math::Vector<int, 3>([](uint32_t const i) constexpr { return i; });
```

### DIFF
--- a/include/pmacc/math/vector/Vector.hpp
+++ b/include/pmacc/math/vector/Vector.hpp
@@ -123,6 +123,27 @@ namespace pmacc
 
             constexpr Vector() = default;
 
+            /** Initialize via a generator expression
+             *
+             * The generator must return the value for the corresponding index of the component which is passed to the
+             * generator.
+             */
+            template<
+                typename F,
+                std::enable_if_t<std::is_invocable_v<F, std::integral_constant<uint32_t, 0u>>, uint32_t> = 0u>
+            ALPAKA_FN_HOST_ACC constexpr explicit Vector(F&& generator)
+                : Vector(std::forward<F>(generator), std::make_integer_sequence<uint32_t, dim>{})
+            {
+            }
+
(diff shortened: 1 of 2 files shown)


include/pmacc/math/vector/Vector.hpp
```diff
@@ -123,6 +123,25 @@ namespace pmacc
...
             constexpr Vector() = default;
 
+            /** Initialize via a generator expression
+             *
+             * The generator must return the value for the corresponding index of the component which is passed to the
+             * generator.
+             */
+            template<typename F, std::enable_if_t<std::is_invocable_v<F, std::integral_constant<int, 0>>, int> = 0>
+            ALPAKA_FN_HOST_ACC constexpr explicit Vector(F&& generator)
+                : Vector(std::forward<F>(generator), std::make_integer_sequence<int, dim>{})
```
ikbuibui:
```suggestion
                : Vector(std::forward<F>(generator), std::make_index_sequence<dim>{})
```

ikbuibui (dismissed):
Looks fine, just pointing out that `std::make_index_sequence<N>`exists which exactly fits what we want

include/pmacc/math/vector/Vector.hpp
psychocoderHPC: will change it to using `uint32_t` because `size_t` range is to large and not required.

psychocoderHPC: @ikbuibui I updated this to `uint32_t`

ikbuibui (dismissed):

ikbuibui (approved):

psychocoderHPC:
@ikbuibui PR can be merged

PR #4933, opened 2024-06-10
refactor flexible binning

psychocoderHPC:
- simplify the internal functor object usage by auto detect type signature during the functor call.
- use right hand side `const`
- add missing includes
- remove usage of `SIMDIM` by `simDim`


ikbuibui:
Looks much better, the required data is now passed in the functor call instead of functor instantiation, thus types can be deduced. It also reduces the size of the functor structs.

ikbuibui (dismissed):

ikbuibui (approved):

PR #4934, opened 2024-06-10
time-step dependent init functor execution

psychocoderHPC:
Add wrapper functors which can wrap particle init functors and allow the execution depending on the simultation time step.

Functionality is extracted from PR #4638

### DIFF
--- a/include/picongpu/particles/InitFunctors.hpp
+++ b/include/picongpu/particles/InitFunctors.hpp
@@ -222,5 +222,59 @@ namespace picongpu
             }
         };
 
+        namespace detail
+        {
+            /** Time step conditional functor execution of init functors
+             * @tparam T_timeStep Timestep to compare
+             * @tparam T_Functor Functor which is executed any time the T_Comparator::operator()(current step,
+             *                   T_timeStep) is true.
+             * @tparam T_Comparator binary comparison operator
+             * */
+            template<uint32_t T_timeStep, typename T_Functor, typename T_Comparator>
+            struct ExecuteIfTimeStep
+            {
+                HINLINE void operator()(const uint32_t currentStep)
+                {
+                    if(T_Comparator{}(currentStep, T_timeStep))
+                        T_Functor{}(currentStep);
+                }
+            };
+        } // namespace detail
+
+        /** Time step conditional functor execution
+         *
+         * @tparam T_timeStep Timestep to compare
+         * @tparam T_Functor Functor which is executed any time the condition is true.
+         *
+         * @code{.cpp}
+         *    // add this for example to the species InitPipeline if you like to add a
+         *    // temperature only in the time-step 100
+         *    ExecuteIfTimeStepEq<100,Manipulate<manipulators::AddTemperature, PIC_Electrons>>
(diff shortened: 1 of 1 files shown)


include/picongpu/particles/InitFunctors.hpp
```diff
@@ -222,5 +222,69 @@ namespace picongpu
...
+        struct ExecuteIfTimeStepLe
+        {
+            HINLINE void operator()(const uint32_t currentStep)
+            {
+                if(currentStep <= T_time)
+                    T_Functor{}(currentStep);
+            }
+        };
+
+        /** @} */
```
ikbuibui:
```suggestion
        /** Time step conditional functor execution
         *
         * @tparam T_time Timestep to compare
         * @tparam T_Functor Functor which is executed any time the condition is true.
         *
         * @{
         */

        template<uint32_t T_time, typename T_Functor, typename Comparator>
        struct ExecuteIfTimeStep
        {
            HINLINE void operator()(const uint32_t currentStep)
            {
                if(Comparator{}(currentStep, T_time))
                    T_Functor{}(currentStep);
            }
        };
        
        // Equal
        template<uint32_t T_time, typename T_Functor>
        using ExecuteIfTimeStepEq = ExecuteIfTimeStep<T_time, T_Functor, std::equal_to<uint32_t>>;
        
        // Greater than
        template<uint32_t T_time, typename T_Functor>
        using ExecuteIfTimeStepGt = ExecuteIfTimeStep<T_time, T_Functor, std::greater<uint32_t>>;
        
        // Greater or Equal
        template<uint32_t T_time, typename T_Functor>
        using ExecuteIfTimeStepGe = ExecuteIfTimeStep<T_time, T_Functor, std::greater_equal<uint32_t>>;
        
        // Less than
        template<uint32_t T_time, typename T_Functor>
        using ExecuteIfTimeStepLt = ExecuteIfTimeStep<T_time, T_Functor, std::less<uint32_t>>;
        
        // Less or Equal
        template<uint32_t T_time, typename T_Functor>
        using ExecuteIfTimeStepLe = ExecuteIfTimeStep<T_time, T_Functor, std::less_equal<uint32_t>>;
        /** @} */
```

ikbuibui (commented):
I don't see how it is going to be used so just for my information is passing the time as a template parameter really necessary or can it be done with constexpr variables?

steindev:
I like @ikbuibui's suggestion.

include/picongpu/particles/InitFunctors.hpp
psychocoderHPC: I will look into your suggestion I think it is not possible because `std::less_equal` and other are binary methods where PIConGPUs interface requires the unary interface where we pass the timestep only.

psychocoderHPC: Ohh I missed the generic implementation. I will take you suggestion and add it

psychocoderHPC: @ikbuibui I pushed my changes

ikbuibui (approved):

PR #4935, opened 2024-06-11
Seperate PICoNGPU compile in independent compile units

psychocoderHPC:
Split PIConGPU code to compile parallel in independent compile units.

**Do not merge** this is a prototype to check if we observe issues in the CI.


psychocoderHPC:
@BrianMarre do you see whats wrong here? Why is pypicongpu crashing during the compile?

https://gitlab.com/hzdr/crp/picongpu/-/jobs/7408406476

psychocoderHPC:
> @BrianMarre do you see whats wrong here? Why is pypicongpu crashing during the compile?
> 
> https://gitlab.com/hzdr/crp/picongpu/-/jobs/7408406476

Maybe it took so long because we can not set `pic-build -j` for pyicongpu and it is running out of memory. For some reasons the job is also very slow and is going near to the 3h walltime.

psychocoderHPC:
this PR is replaced by many others where the ID generator is given to each particle creation kernel therefore we can close this PR

PR #4936, opened 2024-06-12
refactor `pic-build` and `pic-configure`

psychocoderHPC:
- expose `-j` parameter for paralell build to the user
- use cmake instead of `make` to call compile and install
- `pic-configure` add parameter `-G` for setting CMake build system

### DIFF
--- a/bin/pic-build
+++ b/bin/pic-build
@@ -40,11 +40,12 @@ help()
     echo ""
     echo "usage: pic-build [OPTIONS]"
     echo ""
+    echo "-j [N]                - allow N jobs at once; infinite jobs with no arg"
     # lend the rest of the options from pic-configure
     echo "$($this_dir/pic-configure --help | tail -n +12)"
 }
 
-cmd_line_args=("$@")
+cmd_line_args=()
 
 # show help
 while [[ $# -gt 0 ]] ; do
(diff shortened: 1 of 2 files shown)


bin/pic-build
```diff
@@ -40,11 +40,12 @@ help()
     echo ""
     echo "usage: pic-build [OPTIONS]"
     echo ""
+    echo "-j [N]                - allow N jobs at once; infinite jobs with no arg"
```
finnolec, line 43: Maybe we should call it `unlimited jobs` instead of `infinite jobs` like in the `make` documentation? The infinite looks very weird

finnolec: Nevermind, this here is 1 to 1 what `make --help` says, I looked at the man-pages.

bin/pic-configure
```diff
@@ -41,19 +41,20 @@ help()
...
+    echo "                        syntax: backend[:architecture]"
+    echo "                        supported backends: cuda, hip, omp2b, serial, tbb, threads"
+    echo "                        (e.g.: \"cuda:35;37;52;60\" or \"omp2b:native\" or \"omp2b\")"
+    echo "                        default: \"cuda\" if not set via environment variable PIC_BACKEND"
+    echo "                        note: architecture names are compiler dependent"
+    echo "-c | --cmake          - overwrite options for cmake"
+    echo "                        (e.g.: \"-DPIC_VERBOSE=21 -DCMAKE_BUILD_TYPE=Debug\")"
+    echo "-t <presetNumber>     - configure this preset from cmakeFlags"
+    echo "-f | --force          - clear the cmake file cache and scan for new param files"
+    echo "-G <cmakeBuildSystem> - path were picongpu shall be installed"
```
ikbuibui: Documentation is incorrect

ikbuibui (changes requested):

bin/pic-configure
psychocoderHPC: updated

ikbuibui (approved):

PR #4938, opened 2024-06-12
Add laser parameter output for --dump-metadata

max-lehman14:
This pull request adds laser parameter output in the `BaseParam.def` and `GaussianPulse.def` files, based on the new parameter output [feature](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4812) from @chillenzer .

In the `picongpu-metadata.json` file the standard laser parameters from `incidentField.param` are displayed with units, for example:

```
"W0": {
    "unit": "m",
    "value": 4.246609082647506e-06
}
```
or
```
"direction": {
    "unit": "none",
    "value": [
      0.0,
      1.0,
      0.0
    ]
},

```

The next things on my list are:

- [x] ~~adding README~~ [EDIT/REMOVED by @PrometheusPi]

- [x] invest deeper in the problem, that Laguerremodes and -phases in the incidentField.param file are not correctly displayed if there are several Laguerremodes (seems like its falling back to the default 0th Laguerre mode for a standard Gaussian)

### DIFF
--- a/include/picongpu/fields/incidentField/profiles/BaseParam.def
+++ b/include/picongpu/fields/incidentField/profiles/BaseParam.def
@@ -213,43 +213,45 @@ namespace picongpu
                         return origin == Origin::Zero ? "zero" : "center";
                     }
 
-                    template<typename My = BaseParam>
+                    template<typename T_Self = BaseParam>
                     static nlohmann::json metadata()
                     {
                         // We make this a function template in order to gain control over the type we are applying this
                         // to. A derived class of the BaseParam should similarly make their own `.metadata()` member a
-                        // `template<typename My =Derived>` and should use `BaseParam::metadata<My>()` as a starting
-                        // point for customisation.
+                        // `template<typename T_Self =Derived>` and should use `BaseParam::metadata<T_Self>()` as a
+                        // starting point for customisation.
 
                         auto result = nlohmann::json::object();
 
-                        result["wavelength"] = {{"value", My::WAVE_LENGTH_SI}, {"unit", "m"}};
-                        result["amplitude"] = {{"value", My::AMPLITUDE_SI}, {"unit", "V/m"}};
-                        result["pulse_duration"] = {{"value", My::PULSE_DURATION_SI}, {"unit", "s"}};
-                        result["laser_phase"] = {{"value", My::LASER_PHASE}, {"unit", "rad"}};
+                        result["wavelength"] = {{"value", T_Self::WAVE_LENGTH_SI}, {"unit", "m"}};
+                        result["amplitude"] = {{"value", T_Self::AMPLITUDE_SI}, {"unit", "V/m"}};
+                        result["pulse_duration"] = {{"value", T_Self::PULSE_DURATION_SI}, {"unit", "s"}};
+                        result["laser_phase"] = {{"value", T_Self::LASER_PHASE}, {"unit", "rad"}};
                         result["direction"]
-                            = {{"value", vector<float_64>{{My::DIRECTION_X, My::DIRECTION_Y, My::DIRECTION_Z}}},
+                            = {{"value",
+                                vector<float_64>{{T_Self::DIRECTION_X, T_Self::DIRECTION_Y, T_Self::DIRECTION_Z}}},
                                {"unit", "none"}};
                         result["focus_position"]
                             = {{"value",
                                 vector<float_64>{
-                                    {My::FOCUS_POSITION_X_SI, My::FOCUS_POSITION_Y_SI, My::FOCUS_POSITION_Z_SI}}},
+                                    {T_Self::FOCUS_POSITION_X_SI,
+                                     T_Self::FOCUS_POSITION_Y_SI,
+                                     T_Self::FOCUS_POSITION_Z_SI}}},
                                {"unit", "m"}};
                         result["focus_origin"]
                             = {{"type",
                                 vector<std::string>{
-
-                                    originToString(My::FOCUS_ORIGIN_X),
-                                    originToString(My::FOCUS_ORIGIN_Y),
-                                    originToString(My::FOCUS_ORIGIN_Z)}}};
-                        result["time_delay"] = {{"value", GetTimeDelay<My>::value}, {"unit", "s"}};
+                                    originToString(T_Self::FOCUS_ORIGIN_X),
+                                    originToString(T_Self::FOCUS_ORIGIN_Y),
+                                    originToString(T_Self::FOCUS_ORIGIN_Z)}}};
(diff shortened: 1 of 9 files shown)


max-lehman14:
The next things on my list are:

- [ ] adding README

- [ ] invest deeper in the problem, that Laguerremodes and -phases in the incidentField.param file are not correctly displayed if there are several Laguerremodes (seems like its falling back to the default 0th Laguerre mode for a standard Gaussian)

chillenzer:
Without having looked at the content I can already see that [this file](https://github.com/ComputationalRadiationPhysics/picongpu/blob/5f1523e8b19377af40751a909da836d7ae2d23e7/share/picongpu/tests/metadataFromLaserWakefield/picongpu-metadata.json.reference) didn't change, so almost certainly you didn't run the tests (as the failing CI is also telling you). Please update said file and check that the tests and examples all compile and pass.

chillenzer:
@PrometheusPi @max-lehman14 What's the status here?

PrometheusPi:
@chillenzer I will take over this PR

PrometheusPi:
This still does not output anything else than the default parameters. And there should be a bug in the Gaussian code that is actually never causing any compile time issues.

PrometheusPi:
@steindev @chillenzer @psychocoderHPC ready for review
I will not add a README since I do not see the benefit of that. If you see any, please let m eknow.

share/picongpu/examples/LaserWakefield/include/picongpu/param/incidentField.param
```diff
@@ -250,6 +242,14 @@ namespace picongpu
                 static constexpr auto laguerreModes = floatN_X<numModes + 1>(1.0);
                 static constexpr auto laguerrePhases = floatN_X<numModes + 1>(0.0);
                 /** @} */
+
+                template<typename My = GaussianPulseParam>
```
steindev: We typically use `T_` as default names for template arguments and not `My`. I would like to stick to that guideline.

steindev (commented):
Just one comment.

share/picongpu/examples/LaserWakefield/include/picongpu/param/incidentField.param
PrometheusPi: This was already part of the original implementation from @chillenzer. 
I agree with you. @psychocoderHPC should this change go in this PR as well?

PrometheusPi:
Interstingly, we get compile errors like:
```
crp/picongpu/include/picongpu/../picongpu/fields/incidentField/profiles/BaseParam.def:226:63: error: no member named 'TIME_DELAY_SI' in 'picongpu::fields::incidentField::profiles::MyGaussianParam'
  226 |                         result["time_delay"] = {{"value", My::TIME_DELAY_SI}, {"unit", "s"}};
  ```
  which I did not see when I tested the code on hemera.

File: share/picongpu/tests/metadataFromLaserWakefield/picongpu-metadata.json.reference [hunk 2]
```diff
@@ -11,13 +11,82 @@
...
             0.0,
             0.0
           ],
           "type": "circular"
+        },
+        "pulse_duration": {
+          "unit": "s",
+          "value": 5e-15
+        },
+        "time_delay": {
```
PrometheusPi, line 83: The time delay was actually outputted correctly for my LWFA test - strange

share/picongpu/examples/LaserWakefield/include/picongpu/param/incidentField.param
PrometheusPi: renamed `My` to `T_Self`

include/picongpu/fields/incidentField/profiles/BaseParam.def
```diff
@@ -223,7 +241,7 @@ namespace picongpu
                                     originToString(My::FOCUS_ORIGIN_X),
                                     originToString(My::FOCUS_ORIGIN_Y),
                                     originToString(My::FOCUS_ORIGIN_Z)}}};
-                        result["time_delay"] = {{"value", My::TIME_DELAY_SI}, {"unit", "s"}};
+                        result["time_delay"] = {{"value", GetTimeDelay<My>()}, {"unit", "s"}};
```
psychocoderHPC: MUST BE `GetTimeDelay<My>::value`

psychocoderHPC (dismissed):

include/picongpu/fields/incidentField/profiles/BaseParam.def
PrometheusPi: Thanks - I applied it

chillenzer:
You've done everything right so far but the CI failures are due to `PlaneWaveParam` in some examples not inheriting from `BaseParam`, see e.g. in [Bunch](https://github.com/ComputationalRadiationPhysics/picongpu/blob/7132138f71bc44a98bdde565ba5247339834e691/share/picongpu/examples/Bunch/include/picongpu/param/incidentField.param#L76) and analogously in [FoilLCT](https://github.com/ComputationalRadiationPhysics/picongpu/blob/7132138f71bc44a98bdde565ba5247339834e691/share/picongpu/examples/FoilLCT/include/picongpu/param/incidentField.param#L81). This needs to be added and in the best case, we go through the examples and tests by hand and check for such missing inheritance.

share/picongpu/examples/LaserWakefield/include/picongpu/param/incidentField.param
chillenzer: I'm aware of this and I didn't stick to that convention because I see it less as a template parameter and more as a hack/workaround for injecting the correct type.

I'm fine with changing it but if so, we should probably do it consistently everywhere and adopt that as the new convention.

PrometheusPi:
The last CI error comes from PyPIConGPU. The focus position doesn't seem to be defined. I will take a look at it, but will most likely need some guidance from @BrianMarre and @psychocoderHPC

BrianMarre:
As discussed offline, the error is caused by the `incidentField.param.mustache` template.

BrianMarre:
@PrometheusPi do we also need to add the inheritance to BaseParam [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/2c6fc62d26c4395369f7cfa694bc411fb92a558a/share/picongpu/examples/IncidentField/include/picongpu/param/incidentField.param#L77) and [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/2c6fc62d26c4395369f7cfa694bc411fb92a558a/share/picongpu/tests/compileLaser/include/picongpu/param/incidentField.param#L73)?

PrometheusPi:
@BrianMarre No, the first doesn't contain a Gaussian profile and the second uses the default one, that is correct.

ikbuibui (approved):

PR #4942, opened 2024-06-13
update profile and tpl for JURECA 

ikbuibui:
Update the profile and tpl for JURECA. Modified from the JUWELS booster profile which is a very similar system, also with A100s.


ikbuibui:
Tested working profile for the gpu nodes of JURECA

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
```diff
@@ -10,55 +10,65 @@ export MY_MAIL="someone@example.com"
...
-export CMAKE_PREFIX_PATH=$BOOST_ROOT:$CMAKE_PREFIX_PATH
-export LD_LIBRARY_PATH=$BOOST_ROOT/lib:$LD_LIBRARY_PATH
-
-HDF5_ROOT=$PARTITION_LIB/hdf5
-export PATH=$HDF5_ROOT/bin:$PATH
-export CMAKE_PREFIX_PATH=$HDF5_ROOT:$CMAKE_PREFIX_PATH
-export LD_LIBRARY_PATH=$HDF5_ROOT/lib:$LD_LIBRARY_PATH
+# needs to be compiled by the user
+# Check the install script at
+# https://gist.github.com/steindev/5a2ab0c4957a1e57ed671ac198b491c6
```
steindev: This is not up-to-date and should not be linked here. @PrometheusPi don't you have something recent?

etc/picongpu/jureca-jsc/gpus.tpl
```diff
@@ -27,21 +27,26 @@
 # Sets batch job's name
 #SBATCH --job-name=!TBG_jobName
 #SBATCH --nodes=!TBG_nodes
+#SBATCH --exclusive
```
steindev, line 30: Just to be sure: Did you test this profile on multiple nodes? I once made bad experience with `--exclusive` at ZIH. But I guess how this option is treated is system dependent.

PrometheusPi: I tested it on 4 nodes and it works fine.

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
```diff
@@ -10,55 +10,65 @@ export MY_MAIL="someone@example.com"
...
 module load CMake
-module load MVAPICH2/2.3-GDR
-module load Python/3.6.6
-
-# Other Software ##############################################################
+module load OpenMPI/4.1.5
+module load Python/3.11.3
+module load ADIOS2/2.9.2
+module load HDF5/1.14.2
+module load Boost/1.82.0
```
PrometheusPi, line 44: No FFTW module - but tha's fine with me.

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
PrometheusPi: No, I never wrote a script for it - too much testing in the end. 
I would favor keeping the link as can be seen as a reference.

PrometheusPi (approved):
Looks good to me. Tested compiling and running the KHI example on 4 nodes. All worked flawlessly.

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
PrometheusPi: Short *offline* discussion with @steindev. The gist is so outdated that we shoudl remove it.

PrometheusPi:
```suggestion
# Check the install description at
# https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/INSTALL.rst
```

PrometheusPi (changes requested):
@steindev wanted this change

PrometheusPi (approved):

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
ikbuibui: Since shadowgraphy is not used, not including FFTW should reduce compile time, hence the omission.

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
ikbuibui: I edited this script which I am now using to set up the CI environment. I can post that if it makes sense, but I am not sure where I should upload it. @PrometheusPi @steindev  Or we can change to point to install.rst as suggested

etc/picongpu/jureca-jsc/gpus.tpl
ikbuibui: Yes, I also tested it earlier and it was working fine, but I went through the documentation again and found [this](https://apps.fz-juelich.de/jsc/hps/jureca/known-issues.html#slurm-srun-options-exact-and-exclusive-change-default-pinning). Apparently as a side effect --exclusive pins to SMT cores instead of physical cores. Following the workaround listed, I will edit the script to add ``-m *:block``.

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
PrometheusPi: Thanks for the update.

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
PrometheusPi: If you upload the script to your gist, that would be perfect. Then you could discard my last commit (to your pull request), add the link to your gist and squash everything to a single commit.

steindev (approved):

PR #4945, opened 2024-06-17
fix custom input rendering

BrianMarre:
fixes an bug encountered when trying to render a PICMI-simulation after adding custom input to it.

To avoid a repeat also adds a test case that test that a simulation with custom input may be rendered.

### DIFF
--- a/lib/python/picongpu/pypicongpu/simulation.py
+++ b/lib/python/picongpu/pypicongpu/simulation.py
@@ -103,7 +103,7 @@ class Simulation(RenderedObject):
             "found custom user input with tags: "
             + str(serialized["customuserinput"]["tags"])
             + "\n"
-            + "\t WARNING: custom input is not checked, it is the users responsibility to check inputs and generated input.\n"
+            + "\t WARNING: custom input is not checked, it is the user's responsibility to check inputs and generated input.\n"
             + "\t WARNING: custom templates are required if using custom user input.\n"
         )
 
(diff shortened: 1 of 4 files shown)


PrometheusPi:
@psychocoderHPC There seems to be something wrong with the CI. It says 
> Currently unable to fetch data for this pipeline.

BrianMarre:
The CI got confused because I tried to install the json dependency which does not exist. Json is actually already included in the core python install.

BrianMarre:
@PrometheusPi ready for review

lib/python/picongpu/pypicongpu/simulation.py
```diff
@@ -98,11 +99,13 @@ def __render_custom_user_input_list(self) -> dict:
...
 
     def __foundCustomInput(self, serialized: dict):
-        print("NOTE: found custom user input with tags: " + str(serialized["customuserinput"]["tags"]))
-        print(
-            "\t WARNING: custom input is not checked, it is the users responsibility to check inputs and generated input."
+        logging.info(
+            "found custom user input with tags: "
+            + str(serialized["customuserinput"]["tags"])
+            + "\n"
+            + "\t WARNING: custom input is not checked, it is the users responsibility to check inputs and generated input.\n"
```
PrometheusPi: typo 
```suggestion
            + "\t WARNING: custom input is not checked, it is the user's responsibility to check inputs and generated input.\n"
```

PrometheusPi (dismissed):

lib/python/picongpu/pypicongpu/simulation.py
PrometheusPi: Looks good, could just fix the typo.

PrometheusPi (approved):

PR #4949, opened 2024-06-20
Updated snakemake example

JessicaTiebel:
fixed #4939
Simplyfied and updated snakemake example.
New features include:
- usage of config file
- automated generation of pic-build command
- no need to touch Snakefile anymore for simple applications
- example for LSF clusters
- simplified Snakefile

PR #4950, opened 2024-06-24
remove python from snakemake requirements.txt

PrometheusPi:
This pull request fixes a little bug I oversaw when merging #4949. There should  be no `python` in `requirements.txt` since it is not itself a python package.


JessicaTiebel (approved):

PrometheusPi:
I tested this using python 3.12.2 on hemera. With `python` it caused the error: 
>ERROR: Could not find a version that satisfies the requirement python (from versions: none)
>ERROR: No matching distribution found for python

without `python`, `pip install -m requirements.txt` worked fine.

psychocoderHPC (approved):

PR #4953, opened 2024-06-24
fix `pic-build -c`

psychocoderHPC:
With #4936 we broke the option `-c`.

The CI is disabled because we do not use `pic-build` in the CI.

### DIFF
--- a/bin/pic-build
+++ b/bin/pic-build
@@ -61,7 +61,7 @@ while [[ $# -gt 0 ]] ; do
             fi
             ;;
         *)
-            cmd_line_args+=($1)
+            cmd_line_args+=("$1")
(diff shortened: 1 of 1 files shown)


ikbuibui:
I am not good with bash but I was wondering if the ``""`` are also required around ``$2`` on lines 58 and 59

psychocoderHPC:
> I am not good with bash but I was wondering if the `""` are also required around `$2` on lines 58 and 59

No in line 58 we know that we need a single value only. In the fixed line we can have spaces in and we need to keep the parameter as a single value, so we need to group the value.

PrometheusPi (approved):

PR #4957, opened 2024-06-24
fix import statements for plugins

PrometheusPi:
Since quite a while (since PICMI), our plugin example codes snippets were outdated.  This pull request updates that.


BrianMarre (approved):
looks good

psychocoderHPC (approved):

PR #4960, opened 2024-06-26
fix C++20 compiler

psychocoderHPC:
- fix using reference to possible temporary
- `is_constructible_v` in C++20 is returning true if a member can be explicit initialized even if there is no constructor available.

### DIFF
--- a/include/picongpu/particles/functor/User.hpp
+++ b/include/picongpu/particles/functor/User.hpp
@@ -49,7 +49,9 @@ namespace picongpu
                 template<typename DeferFunctor = Functor>
                 HINLINE User(
                     uint32_t currentStep,
-                    std::enable_if_t<std::is_constructible_v<DeferFunctor, uint32_t>>* = 0)
+                    std::enable_if_t<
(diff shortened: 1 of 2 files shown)


include/picongpu/particles/functor/User.hpp
```diff
@@ -49,7 +49,9 @@ namespace picongpu
                 template<typename DeferFunctor = Functor>
                 HINLINE User(
                     uint32_t currentStep,
-                    std::enable_if_t<std::is_constructible_v<DeferFunctor, uint32_t>>* = 0)
+                    std::enable_if_t<
+                        !std::is_default_constructible_v<
+                            DeferFunctor> && std::is_constructible_v<DeferFunctor, uint32_t>>* = 0)
```
PrometheusPi, line 54: This line breaking looks very unintuitive since the AND is somewhere in the middle, but if the style check is fine with it, okay.

PrometheusPi (approved):
Looks good to me, I just don't like the line breaks

PR #4963, opened 2024-06-26
Optimize dispersive pulse

psychocoderHPC:
Avoid multiple call to the expensive `exp()` function.

### DIFF
--- a/include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
+++ b/include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
@@ -185,15 +185,16 @@ namespace picongpu
                                 + SPEED_OF_LIGHT * alpha * focusPos / (Unitless::W0 * Unitless::w);
 
                             // gaussian envelope in frequency domain
-                            float_X mag = math::exp(
-                                -(Omega - Unitless::w) * (Omega - Unitless::w) * Unitless::PULSE_DURATION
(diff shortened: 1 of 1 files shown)


include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -185,15 +185,16 @@ namespace picongpu
                                 + SPEED_OF_LIGHT * alpha * focusPos / (Unitless::W0 * Unitless::w);
 
                             // gaussian envelope in frequency domain
-                            float_X mag = math::exp(
-                                -(Omega - Unitless::w) * (Omega - Unitless::w) * Unitless::PULSE_DURATION
-                                * Unitless::PULSE_DURATION);
+                            float_X const freqExp = -(Omega - Unitless::w) * (Omega - Unitless::w)
+                                * Unitless::PULSE_DURATION * Unitless::PULSE_DURATION;
```
fabidie, line 189: One could rename it to envFreqExp, so that it would be in accordance with the following variable naming schemes.

fabidie (approved):

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
psychocoderHPC: done

steindev (approved):

PR #4964, opened 2024-06-26
change pulse laguerre mode definition

psychocoderHPC:
Use `floatN_X<>` to configure laguerre modes instead of `PMACC_CONST_VECTOR`.
This comes with the advantage that the modes are now part of the configuration struct and there is no need for a namespace we reference within the struct.

**ATTENTION** This PR makes older setups incompatible. Change the laguerre modes to solve issue.

```
static constexpr uint32_t numModes = 0;
static constexpr auto laguerreModes = floatN_X<numModes + 1>(1.0);
static constexpr auto laguerrePhases = floatN_X<numModes + 1>(0.0);
```

New shorthand notations to create `pmacc::math::Vector<floatX,dim>` are introduced.
  - `floatN_X<n>`
  - `floatN_64<n>`
  - `floatN_32<n>`

- [x] merge after #4965 (first commit of this PR can be ignored because it is from #4965)

### DIFF
--- a/include/picongpu/fields/incidentField/profiles/GaussianPulse.def
+++ b/include/picongpu/fields/incidentField/profiles/GaussianPulse.def
@@ -30,21 +30,6 @@ namespace picongpu::fields::incidentField::profiles
 {
     namespace defaults
     {
-        namespace gaussianPulse
-        {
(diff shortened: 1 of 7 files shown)


ikbuibui (approved):
Looks ok :+1:

PR #3246, opened 2020-05-19
remove old HZDR systems

psychocoderHPC:
- hypnos: Is not longer available, the compute resource are now in the
hemera cluster
  - remove also old paraview file for hypnos (you can find it still in old releases)
- hydra: Is still available but not used by PIConGPU users. Due to mixed compute nodes /Main memory settings) and a not maintained software stack it is not useful to keep the system template.

PR #4968, opened 2024-06-27
drop HIP 5.2/5.3 and CUDA 11.0 support

psychocoderHPC:
- HIP 5.2/5.3 does not support device relocatable linking with CMake
- CUDA 11.0 produces compile issues in an upcoming PR #4935 
  - see: kokkos/kokkos#3496  
  ```
  Error #3246: Internal Compiler Error (codegen): "there was an error in verifying the lgenfe output!"
  ```

This is a preparation PR for #4935 (multiple compile units)

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -134,7 +134,7 @@ Optional Libraries
 
 CUDA
 """"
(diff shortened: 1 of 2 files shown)


steindev (approved):
Also discussed offline. Is fine. Will merge with superpowers.

PR #4985, opened 2024-07-09
add configuration of ionizers to PICMI

mafshari64:
adds the option of configuring the ionization models from PICMI


BrianMarre:
@ikbuibui may be closed, replaced by PR #5007

PR #4987, opened 2024-07-10
add moving window configuration to the interface PICMI

BrianMarre:
adds the ability to configure the moving window setting of picongpu from PICMI


BrianMarre:
@ikbuibui ready for review

share/picongpu/pypicongpu/template/etc/picongpu/N.cfg.mustache
```diff
@@ -60,15 +61,13 @@ TBG_wallTime="1:00:00"
...
-{{/grid.boundary_condition}}
-
-
 #################################
 ## Section: Optional Variables ##
 #################################
 
+{{#grid.boundary_condition}}
+  TBG_periodic="--periodic {{{x}}} {{{y}}} {{{z}}}"
+{{/grid.boundary_condition}}
```
ikbuibui, line 70: why have these been moved under optional variables?

test/python/picongpu/compiling/simulation.py
```diff
@@ -56,6 +67,12 @@ def test_minimal(self):
         runner.build()
         runner.run()
 
+    def test_moving_window(self):
+        picmi_sim = self._set_up_sim(picongpu_moving_window_move_point=0.9)
+        runner = pypicongpu.Runner(picmi_sim)
+        runner.generate(printDirToConsole=True)
+        runner.build()
+
```
ikbuibui, line 79: Do we also need a runner.run() here?

share/picongpu/pypicongpu/template/etc/picongpu/N.cfg.mustache
BrianMarre: because they are considered optional by the PIConGPU parsing of runtime parameters.

PIConGPU uses the default  set in the .param files if periodic boundaries are not specified.

In contrast PIConGPU  will not start the simulation if grid size or device number is not specified.

I therefore opted to move them to the optional section while reworking the template, doesn't change anything function wise, just groups stuff correctly.

test/python/picongpu/compiling/simulation.py
BrianMarre: yes, you are right. Since Moving Window configuration is run time only we need to do a short run to test

BrianMarre: fixed

ikbuibui (dismissed):

test/python/picongpu/compiling/simulation.py
BrianMarre: But we can't since our CI jobs only ever have access to one device therefore the PyPIConGPU runner can't run a simulation with moving window, since we require at least 2 devices for a moving window.

We could use the `-r 2` option to run a moving window simulation but the PyPIConGPU runner can't do that currently.

Therefore would shift the run time tests to a future PR.

ikbuibui: Sounds okay. Could you please create an issue to track the addition of the runtime test and I will merge this PR as is.

BrianMarre: created the Issue #4995 to track the missing feature.

BrianMarre:
@ikbuibui Please restart the failed CI jobs, the seem to have failed due to reasons outside of the PR.

ikbuibui (approved):

PR #4628, opened 2023-07-07
fixing failing PyPIConGPU CI jobs

BrianMarre:
CI jobs are failing due to a breaking change/bug in the PyPIConGPU dependency jsonschema, in yesterday's new version 4.18.0.

I have fixed this issue for now by fixing the jsonschema version to the last release before the current version, 4.17.3.

fix #4627

PR #4990, opened 2024-07-10
refactor requirements.txt of PICMI and PyPIConGPU

BrianMarre:
this PR:
- creates separate `requirements.txt` for PICMI and PyPIConGPU
- reverts the earlier fix of the dependency `jsonschema` in PR #4628
- updates the `jsonschema` to the current version
- updates the pypicongpu ci to allow more than one `requirements.txt`
- updates the CI to use both the PICMI and the PyPIConGPU `requirements.txt`
- adds python 3.12 and pydantic to the CI

### DIFF
--- a/share/ci/pypicongpu_generator.py
+++ b/share/ci/pypicongpu_generator.py
@@ -4,6 +4,8 @@ import yaml
 import re
 from typing import List, Dict, Callable
 import pkg_resources
+import copy
+import typeguard
 
 """
 This file is part of PIConGPU.
@@ -18,19 +20,21 @@ License: GPLv3+
 Prints yaml code for a GitLab CI child pipeline to stdout. The test parameters
 are split in two kinds of inputs. The Python versions to test are defined in
 the script, also the names of dependencies to test and it's test strategy. The
-version range of the dependencies to test are defined in a requirements.txt.
-The path of the requirements.txt is set via first application argument.
+version range of the dependencies to test are defined in the passed
+requirements.txt files. The paths of the requirements.txt are set via the
+application arguments.
 
-First, the script reads the requirements.txt. If a dependency is marked to test
-in the script, it calculates the test versions. Therefore it downloads all
-available versions from pypi.org for each package. Afterwards it filters the
-versions via a filter strategy. For example, take all release versions or take
-each latest major version. Than the script removes all versions, which are not
-supported, defined in the requirements.txt. The result a complete list of all
-Python and dependencies versions to test.
+First, the script reads the requirements.txt files. If a dependency is marked
+as to be tested in the script, it calculates the test versions.
+For this it downloads all available versions from pypi.org for each package.
+Afterwards it filters the versions via a filter strategy. For example, take
+all release versions or take each latest major version. Than the script
+removes all versions, which are not supported, as defined in the combined
+requirements.txt.
+The result a complete list of all Python- and dependency- versions to test.
 
 In the second part, the script creates the full combination matrix for all test
-versions and creates a CI job for each combination. Each job is printed on
+versions and creates a CI job for each combination. Each job is printed to
 stdout.
 
 The number of combinations depends on:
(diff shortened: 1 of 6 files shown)


BrianMarre:
@ikbuibui CI failed due to docker container fetch time out, please restart the failed jobs

share/ci/pypicongpu_generator.py
```diff
@@ -127,6 +129,50 @@ def get_all_major_pypi_versions(package_name):
...
+                    else:
+                        # already exists
+                        total_requirements[requirement.project_name] = combine_requirements(
+                            requirement, total_requirements[requirement.project_name]
+                        )
+
+            except Exception:
+                # ignore all lines, which cannot be parsed
+                # e.g. `-r extra/requirements.txt`
+                pass
```
ikbuibui, line 183: Can these lines be safely ignored and the exception just passed?

share/ci/pypicongpu_generator.py
```diff
@@ -127,6 +129,50 @@ def get_all_major_pypi_versions(package_name):
     return [str(v) for v in version_map.values()]
 
 
+@typeguard.typechecked
+def combine_requirements(one: pkg_resources.Requirement, two: pkg_resources.Requirement) -> pkg_resources.Requirement:
+    if one.project_name != two.project_name:
+        print(f"requirements for {one.project_name} and {two.project_name} are incomparable!")
```
ikbuibui, line 138: Is this print sufficient or should an exception be raised

share/ci/pypicongpu_generator.py
```diff
@@ -127,6 +129,50 @@ def get_all_major_pypi_versions(package_name):
...
+
+    # accumulate all specs
+    specs: list = copy.deepcopy(one.specs)
+    specs.extend(two.specs)
+
+    last_tuple = specs.pop()
```
ikbuibui: We pop, but is it ever possible for the specs list to be empty?

ikbuibui (commented):
Just some questions mainly because I am not familiar with this stuff, otherwise looks okay

share/ci/pypicongpu_generator.py
BrianMarre: Yes
If all dependencies required are included in the passed file this never comes up.
If that is not the case, we will get an error in the CI later when we run the tests for code that uses a missing dependency, because it will not be installed in the CI container.

share/ci/pypicongpu_generator.py
BrianMarre: good catch the if is missing an `exit(1)` call afterwards, fixed

share/ci/pypicongpu_generator.py
BrianMarre: good catch, added a check for an empty lists.

BrianMarre:
@SimeonEhrig A review from you would also be useful, after all you are the original author

share/ci/pypicongpu_generator.py
```diff
@@ -127,6 +129,54 @@ def get_all_major_pypi_versions(package_name):
...
+def combine_requirements(one: pkg_resources.Requirement, two: pkg_resources.Requirement) -> pkg_resources.Requirement:
+    if one.project_name != two.project_name:
+        print(f"requirements for {one.project_name} and {two.project_name} are incomparable!")
+        exit(1)
+
+    # accumulate all specs
+    specs: list = copy.deepcopy(one.specs)
+    specs.extend(two.specs)
+
+    if len(specs) != 0:
```
SimeonEhrig, line 145: What is this if else statement doing?

share/ci/pypicongpu_generator.py
```diff
@@ -144,19 +194,13 @@ def get_supported_versions(package_name: str, versions: List[str]) -> List[str]:
...
-            try:
-                for req in parsed_req:
-                    req_versions[req.project_name] = req
-            except Exception:
-                pass
+        req_versions = read_in_requirement_files(requirement_file_names)
 
     if package_name not in req_versions:
-        print(cs(f"ERROR: {package_name} is not defined in {sys.argv[1]}", "Red"))
+        print(cs(f"ERROR: {package_name} is not defined in {sys.argv[1]} or {sys.argv[2]}", "Red"))
```
SimeonEhrig: If I'm not wrong, the modification allows to handle N `requirements.txt`. Therefore the error messages should be modified.

SimeonEhrig:
Please update the documentation. Otherwise, I cannot review in a year again :sweat_smile: 
https://github.com/ComputationalRadiationPhysics/picongpu/pull/4990/files#diff-6157b38912e3f305b07c85e2ed33e86d8b6e2c1165506f31722f1b29caafa868R23

share/ci/pypicongpu_generator.py
BrianMarre: added documentation, to make it clearer what happens

share/ci/pypicongpu_generator.py
BrianMarre: fixed

BrianMarre:
updated the documentation as by Simeons request

SimeonEhrig (approved):

ikbuibui (approved):

PR #4994, opened 2024-07-17
Fix plugin use with filters

ikbuibui:
Using filters with plugins was broken and only worked when the filter used by the user was the same as the first filter in the internal ``EligibleFilters`` type list (usually it is the filter ``all``).
In a plugin, when ``meta::ForEach<typename Help::EligibleFilters, plugins::misc::ExecuteIfNameIsEqual<boost::mpl::_1>>{}(m_help->filter.get(m_id), currentStep, bindKernel);`` is called, the filter name string returned from ``m_help->filter.get(m_id)`` is only passed on correctly to the first expansion from the ForEach List. All other expansions were receiving an empty string.
I am not sure why this happens, I suspect that for some reason the compiler is optimizing the string being passed by value into a move and isn't dealing with the ForEach correctly. As a fix the string is now passed as a const reference, which prevents moves. This fixes the problem in my tests. 
Bug found by @JessicaTiebel

### DIFF
--- a/docs/source/usage/workflows/particleFilters.rst
+++ b/docs/source/usage/workflows/particleFilters.rst
@@ -56,6 +56,10 @@ Let us select particles with momentum vector within a cone with an opening angle
           FunctorParticlesForwardPinhole
        >;
 
+.. note::
+
+    User defined filter functors must be wrapped to fit the general filter interface. This can be done using wrappers like ``generic::Free<T_UserFunctor>`` and ``generic::FreeTotalCellOffset<T_UserFunctor>``
+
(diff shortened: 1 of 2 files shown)


chillenzer (approved):
Looks okay to me but I don't have enough information to reproduce the bug, so that's a very weak statement here. We should wait for @JessicaTiebel's approval.

JessicaTiebel (approved):
I tested it with my setup and it works.
Thanks for the quick fix and also update of documentation ✨

ikbuibui:
closing because I cant undo the CI nocompile flag. Opening #4998

PR #5007, opened 2024-07-25
[PICMI] Add ionization model configuration support

BrianMarre:
adds the ability to configure ionization model interactions in the PIConGPU-PICMI interface.

replaces PR #4985

- [x] update documentation
- [x] add LWFA PICMI script as user example
- [x] add Test for Ionization configuration
- [x] verify correct rendering of LWFA with Richard

also fixes a bug in the initialization configuration of the `boundElectrons` attribute, noticed in the refactoring of the `SetBoundElectrons` Operation, renamed in this PR to `SetChargeState`.


docs/source/usage/picmi/intro.rst
```diff
@@ -113,13 +113,31 @@ Parameters/Methods prefixed with ``picongpu_`` are PIConGPU-exclusive.
 
 - **Simulation**
 
-  - ``__init__(..., picongpu_template_dir)``:
+  additional constructor/configuration options:
+  - ``picongpu_template_dir``:
```
chillenzer, line 134: Doesn't render correctly. This happens multiple times down the site. I won't tag all of them.

docs/source/usage/picmi/intro.rst
```diff
@@ -113,13 +113,31 @@ Parameters/Methods prefixed with ``picongpu_`` are PIConGPU-exclusive.
...
     Specify the template dir to use for code generation,
```
chillenzer:
```suggestion
    Specify the template directory to use for code generation,
```

docs/source/usage/picmi/intro.rst
```diff
@@ -113,13 +113,31 @@ Parameters/Methods prefixed with ``picongpu_`` are PIConGPU-exclusive.
...
     please refer to :ref:`the documentation on the matter for details <picmi-custom-generation>`
-  - ``__init__(...,  picongpu_typical_ppc)`` typical ppc to be used for normalization in PIConGPU
+  - ``picongpu_typical_ppc``:
+    typical ppc to be used for normalization in PIConGPU, if not set explicitly, PIConGPU will use the median ppc of all defined species
```
chillenzer:
```suggestion
    typical particles per cell (ppc) to be used for normalization in PIConGPU, if not set explicitly, PIConGPU will use the median ppc of all defined species
```

docs/source/usage/picmi/intro.rst
```diff
@@ -113,13 +113,31 @@ Parameters/Methods prefixed with ``picongpu_`` are PIConGPU-exclusive.
...
+  - ``picongpu_moving_window_move_point``:
+    portion of the simulation window a light ray reaches from the time of the start of the simulation until the simulation window begins to move.
```
chillenzer, line 140: This description reads somehow convoluted (although probably technically correct). Is there another piece of documentation about this feature in the pure PIConGPU docs that you could refer to? If not, we should create that.

docs/source/usage/picmi/intro.rst
```diff
@@ -113,13 +113,31 @@ Parameters/Methods prefixed with ``picongpu_`` are PIConGPU-exclusive.
...
+
+    .. warning::
+
+      If the moving window is active, one gpu in y direction is reserved for initializing new spaces, thereby reducing the simulation window size accordingly
```
chillenzer: Just for my information (FMI): Is it really one in total or one on each side of the window?

docs/source/usage/picmi/intro.rst
```diff
@@ -113,13 +113,31 @@ Parameters/Methods prefixed with ``picongpu_`` are PIConGPU-exclusive.
...
+
+  - ``picongpu_moving_window_stop_iteration``:
+    iteration at which to stop moving the simulation window
+  - ``picongpu_interaction``:
+    ``Interaction`` object specifying all interactions of the simulation, i.e. all ionization models and their configurations and so on.
+    This replaces the PICMI ``add_interaction`` method.
```
chillenzer, line 150: FMI: Why replace that? Is the PICMI original version so insufficient for our approach?

docs/source/usage/picmi/intro.rst
```diff
@@ -113,13 +113,31 @@ Parameters/Methods prefixed with ``picongpu_`` are PIConGPU-exclusive.
...
+  additional method arguments:
   - ``write_input_file(..., pypicongpu_simulation)``:
     use a :ref:`PyPIConGPU simulation<PyPIConGPU_Intro>` object instead of an PICMI- simulation object to generate a PIConGPU input.
-  - ``get_as_pypicongpu()``:
+
+  additional methods:
+    - ``get_as_pypicongpu()``:
     convert the PICMI simulation object to an equivalent :ref:`PyPIConGPU <PyPIConGPU_Intro>` simulation object.
   - ``picongpu_get_runner()``:
     Retrieve a :ref:`PyPIConGPU Runner <pypicongpu-running>` for running a PIConGPU simulation from Python, **not recommended**
```
chillenzer:
```suggestion
    Retrieve a :ref:`PyPIConGPU Runner <pypicongpu-running>` for running a PIConGPU simulation from Python, **not recommended (see <here>)**
```
where `<here>` should, of course, link to the actual recommendation.

docs/source/usage/picmi/intro.rst
chillenzer: Okay. Found it below. A link to that explanation might be nice. Or put the one unsupported method on top and link the other way round, so the reader is already aware.

docs/source/usage/picmi/intro.rst
```diff
@@ -150,17 +173,34 @@ Parameters/Methods prefixed with ``picongpu_`` are PIConGPU-exclusive.
...
+
+    e = picmi.Species(name="e", particle_type="electron")
+    nitrogen = picmi.Species(name="nitrogen", particle_type="N", charge_state=2)
+
+    ADK_ionization = ADK(ADK_variant = ADKVariant.LinearPolarization, ion_species = nitrogen, ionization_electron_species=e)
+    interaction = Interaction(ground_state_ionizaion_model_list=[ADK_Ionization])
+
+    sim = picmi.simulation(picongpu_interaction=interaction)
+    sim.add_species(e, ...)
+    sim.add_species(nitrogen, ...)
```
chillenzer: Don't place code into the documentation. Place it in an executable example script and pull it in from there via the corresponding `.rst` feature (see how it is solved in https://picongpu--5007.org.readthedocs.build/en/5007/usage/metadata.html).

docs/source/usage/picmi/intro.rst
```diff
@@ -150,17 +173,34 @@ Parameters/Methods prefixed with ``picongpu_`` are PIConGPU-exclusive.
...
+  .. code:: python
+
+    from picongpu import picmi
+    from picongpu.interaction.ionization.fieldionization import ADK, ADKVariant
+
+    e = picmi.Species(name="e", particle_type="electron")
+    nitrogen = picmi.Species(name="nitrogen", particle_type="N", charge_state=2)
+
+    ADK_ionization = ADK(ADK_variant = ADKVariant.LinearPolarization, ion_species = nitrogen, ionization_electron_species=e)
+    interaction = Interaction(ground_state_ionizaion_model_list=[ADK_Ionization])
```
chillenzer: `Interaction` is an undefined name in this example.

lib/python/picongpu/picmi/interaction/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .interactioninterface import InteractionInterface
+from .interaction import Interaction
+from . import ionization
+
+__all__ = ["InteractionInterface", "Interaction", "ionization"]
```
chillenzer: I think as long as you're not changing the default behaviour, adding this line is a maintenance burden. I'd suggest to add it later if ever needed. Applies to all (but I won't tag it everywhere).

lib/python/picongpu/picmi/interaction/ionization/__init__.py
```diff
@@ -0,0 +1,10 @@
+from .ionizationmodel import IonizationModel
+from .groundstateionizationmodel import GroundStateIonizationModel
+from . import fieldionization
+
+__all__ = [
+    "IonizationModel",
+    "GroundStateIonizationModel",
+    "fieldionization",
+    "electroniccollisionalequilibrium",
```
chillenzer, line 10: This name does not exist, does it?

lib/python/picongpu/picmi/interaction/ionization/electroniccollisonalequilibrium/thomasfermi.py
```diff
@@ -0,0 +1,20 @@
+"""
+This file is part of PIConGPU.
+Copyright 2024 PIConGPU contributors
+Authors: Brian Edward Marre
+License: GPLv3+
+"""
+
+from ..groundstateionizationmodel import GroundStateIonizationModel
+from ..... import pypicongpu
```
chillenzer: Wow! That is a lot of dots... XD

But on a more serious note: I see that a lot further below. That feels like really hard to read (in the sense of mentally resolving the reference) and is probably a great annoyance to get right. Couldn't we have `pypicongpu` as a separate module in our environment such that `import pypicongpu` suffices?

lib/python/picongpu/picmi/predefinedparticletypeproperties.py
```diff
@@ -0,0 +1,126 @@
...
+        charge=pdg.connect().get_particle_by_name("H").charge * consts.elementary_charge,
+    ),
+}
+
+non_element_particle_type_properties = {}
+non_element_particle_type_properties.update(_quarks)
+non_element_particle_type_properties.update(_leptons)
+non_element_particle_type_properties.update(_neutrinos)
+non_element_particle_type_properties.update(_nucleons)
+non_element_particle_type_properties.update(_gauge_bosons)
```
chillenzer:
```suggestion
non_element_particle_type_properties = {**_quarks, **_leptons, **_neutrinos, **_nucleons. **_gauge_bosons}
```

lib/python/picongpu/picmi/predefinedparticletypeproperties.py
chillenzer, line 1: Of course, I haven't checked all the numbers, etc.

chillenzer (commented):
Hi, just a first batch. Not really touched on the functional code yet.

lib/python/picongpu/picmi/simulation.py
```diff
@@ -28,10 +31,85 @@ class Simulation(picmistandard.PICMI_Simulation):
...
+    light
+
+    in multiples of the simulation window size
+
+    @attention if moving window is active, one gpu in y direction is reserved for initializing new spaces,
+        thereby reducing the simulation window size accordingrelative spot at which to start moving the simulation window
+    """
+
+    picongpu_moving_window_stop_iteration = pypicongpu.util.build_typesafe_property(typing.Optional[int])
+    """iteration, at which to stop moving the simulation window"""
```
chillenzer, line 72: Looking at this, this is mostly duplicated information with the list in the documentation. Can we come up with a way to de-duplicate? E.g., pulling in the docstrings from the code to auto-generate that list for the documentation or referring to an auto-generated API documentation for details and explanations?

lib/python/picongpu/picmi/simulation.py
```diff
@@ -28,10 +31,85 @@ class Simulation(picmistandard.PICMI_Simulation):
...
+        if picongpu_template_dir is not None:
+            self.picongpu_template_dir = str(picongpu_template_dir)
+        else:
+            self.picongpu_template_dir = picongpu_template_dir
+
+        self.picongpu_typical_ppc = picongpu_typical_ppc
+        self.picongpu_moving_window_move_point = picongpu_moving_window_move_point
+        self.picongpu_moving_window_stop_iteration = picongpu_moving_window_stop_iteration
+        self.picongpu_interaction = picongpu_interaction
+        self.picongpu_custom_user_input = None
```
chillenzer, line 96: Why is this not added in the `__init__`?

lib/python/picongpu/picmi/simulation.py
```diff
@@ -28,10 +31,85 @@ class Simulation(picmistandard.PICMI_Simulation):
...
+        picongpu_typical_ppc: typing.Optional[int] = None,
+        picongpu_moving_window_move_point: typing.Optional[float] = None,
+        picongpu_moving_window_stop_iteration: typing.Optional[int] = None,
+        picongpu_interaction: typing.Optional[Interaction] = None,
+        **keyword_arguments,
+    ):
+        if picongpu_template_dir is not None:
+            self.picongpu_template_dir = str(picongpu_template_dir)
+        else:
+            self.picongpu_template_dir = picongpu_template_dir
```
chillenzer, line 90: Why does this not handle a `Path` directly? Hazarding to state the obvious:`Path` objects are generally safer to use for paths. If you later need it as a string for plugging it into a template or something, I'd call `str()` on that as late as possible.

lib/python/picongpu/picmi/simulation.py
```diff
@@ -28,10 +31,85 @@ class Simulation(picmistandard.PICMI_Simulation):
...
+            self.__yee_compute_cfl_or_delta_t()
+
+        # checks on picongpu specific stuff
+        ## template_path is valid
+        if picongpu_template_dir == "":
+            raise ValueError("picongpu_template_dir MUST NOT be empty string")
+        if picongpu_template_dir is not None:
+            template_path = pathlib.Path(picongpu_template_dir)
+            if not template_path.is_dir():
+                raise ValueError("picongpu_template_dir must be existing directory")
```
chillenzer, line 113: This together with the above `if` conditions could be part of the setter. It's a property anyways, so why burden the reader with the details here when it could handle all that itself?

lib/python/picongpu/picmi/species.py
```diff
@@ -86,192 +135,247 @@ def __get_drift(self) -> typing.Optional[species.operation.momentum.Drift]:
...
+            everything else requires a call to the corresponding getter of this class
+        """
+
+        # error on unsupported options
+        pypicongpu.util.unsupported("method", self.method)
+        pypicongpu.util.unsupported("particle shape", self.particle_shape)
+        # @note placement params are respected in associated simulation object
+
+        self.check(interaction)
         self.__maybe_apply_particle_type()
```
chillenzer, line 251: This pattern seems to appear multiple times. Consider making it a decorator.

lib/python/picongpu/picmi/species.py
```diff
@@ -86,192 +135,247 @@ def __get_drift(self) -> typing.Optional[species.operation.momentum.Drift]:
...
-        """
-        returns true iff a species will have ionizers (algorithms)
-        """
-        return self.charge_state is not None
+        return s, pypicongpu_model_by_picmi_model
 
     def get_independent_operations(
-        self, pypicongpu_species: species.Species
-    ) -> typing.List[species.operation.Operation]:
+        self, pypicongpu_species: pypicongpu.species.Species, interaction: InteractionInterface | None
```
chillenzer: This method should not need to get `pypicongpu_species` from the outside. It should generate that itself. If you allow this, then you could someone could hand a `pypicongpu` ion to a `picmi` electron. I know that you handle that with an assertion on the name but if someone runs with `-O` or messed with the species name that's all weird.

chillenzer: Also, why does it take an interaction? Couldn't we just hand that to the constructor, so the PICMI species just knows that stuff and doesn't have to take it from the outside on every call?

lib/python/picongpu/picmi/simulation.py
```diff
@@ -299,13 +293,12 @@ def __get_init_manager(self) -> species.InitManager:
...
-        initmgr = species.InitManager()
+        initmgr = pypicongpu.species.InitManager()
 
-        # check preconditions
+        # check preconditions, @todo move to picmistandard, Brian Marre 2024
         assert len(self.species) == len(self.layouts)
 
-        # either no layout AND no profile, or both
-        # (also: no ratio without layout and profile)
+        # check either no layout AND no profile, or both and ratio only set if leyout and profile also set
```
chillenzer:
```suggestion
        # check either no layout AND no profile, or both and ratio only set if layout and profile also set
```
But I'd say that this is redundant. It is exactly what the code does (and even explicitly tells you in the assertion).

lib/python/picongpu/picmi/simulation.py
```diff
@@ -315,28 +308,31 @@ def __get_init_manager(self) -> species.InitManager:
                 None
             ), "species need BOTH layout AND initial distribution set (or neither)"
 
-            # ratio only set if
+            # ratio only set if, layout and profile are also set
```
chillenzer: Redundant.

lib/python/picongpu/picmi/simulation.py
```diff
@@ -315,28 +308,31 @@ def __get_init_manager(self) -> species.InitManager:
...
             if ratio is not None:
                 assert (
                     layout is not None and profile is not None
                 ), "layout and initial distribution must be set to use density scale"
 
         # get species list
```
chillenzer: Whenever you feel the need to have a section-header-like comment in your code, extract that into a function.

lib/python/picongpu/picmi/simulation.py
```diff
@@ -315,28 +308,31 @@ def __get_init_manager(self) -> species.InitManager:
...
-        ##
 
-        # note: cache to reuse *exactly the same* object in operations
+        ## @details cache to reuse *exactly the same* object in operations
```
chillenzer: Why's that important? And if so, why not give the variable a name stating what it does? The current name conveys only information about the implementation details, not the intention of the code.

lib/python/picongpu/picmi/simulation.py
```diff
@@ -315,28 +308,31 @@ def __get_init_manager(self) -> species.InitManager:
...
             initmgr.all_species.append(pypicongpu_species)
 
         # fill inter-species dependencies
-        ##
 
-        # ionization (PICMI species don't know which PyPIConGPU species they
-        # use as electrons)
-        self.__fill_ionization_electrons(pypicongpu_by_picmi_species)
+        # ionization electron species need to be set after species translation is complete since the PyPIConGPU electron
+        #   species is not known by the PICMI ion species
```
chillenzer:
```suggestion
        # species is not known by the PICMI ion species
```

lib/python/picongpu/picmi/simulation.py
```diff
@@ -456,25 +363,38 @@ def write_input_file(self, file_name: str, pypicongpu_simulation: simulation.Sim
...
-        self.__runner = runner.Runner(pypicongpu_simulation, self.picongpu_template_dir, setup_dir=file_name)
+        self.__runner = pypicongpu.runner.Runner(pypicongpu_simulation, self.picongpu_template_dir, setup_dir=file_name)
         self.__runner.generate()
 
+    def picongpu_add_custom_user_input(self, custom_user_input: pypicongpu.customuserinput.InterfaceCustomUserInput):
+        """add custom user input to previously stored input"""
+        if self.picongpu_custom_user_input is None:
+            self.picongpu_custom_user_input = [custom_user_input]
+        else:
+            self.picongpu_custom_user_input.append(custom_user_input)
```
chillenzer:
```suggestion
        self.picongpu_custom_user_input = (self.picongpu_custom_user_input or []) + [custom_user_input]
```

lib/python/picongpu/picmi/species.py
```diff
@@ -1,54 +1,103 @@
...
-    *usually* ionization is expected to be used on elements -- use this to
-    explicitly DISABLE ionization
+    mass/charge to use when passed a non-element particle_type
+
+    @attention ONLY set non-element particles here, all other are handled by element
     """
 
-    def __init__(self, picongpu_fully_ionized: typing.Optional[bool] = None, picongpu_ionization_electrons=None, **kw):
-        self.picongpu_fully_ionized = picongpu_fully_ionized
+    __non_element_particle_types: dict[str, _PropertyTuple] = __non_element_particle_type_properties.keys()
```
chillenzer: That doesn't sound about right. Maybe this instead?
```suggestion
    __non_element_particle_types: dict[str, _PropertyTuple] = __non_element_particle_type_properties
```

chillenzer (commented):
Okay. Calling it a day for today. Here's another batch.

General comment: Assertions are not a safe way to check integrity of the data in functional code because they can be deactivated. Such checks should throw other appropriate (potentially custom) forms of exceptions.

docs/source/usage/picmi/intro.rst
BrianMarre: missing empty line, fixed

docs/source/usage/picmi/intro.rst
BrianMarre: fixed

docs/source/usage/picmi/intro.rst
BrianMarre: fixed

docs/source/usage/picmi/intro.rst
BrianMarre: Not as far as I know in our out of code documentation, as far as I know the option is only documented in the ``--help`` option of picongpu and some example ``.cfg``'s.

I definitely agree that we should add documentation for this in a separate PR

docs/source/usage/picmi/intro.rst
BrianMarre: one row actually, fixed, thanks for asking :+1:

docs/source/usage/picmi/intro.rst
BrianMarre: followed your suggestion

docs/source/usage/picmi/intro.rst
BrianMarre: followed your suggestion

docs/source/usage/picmi/intro.rst
BrianMarre: fixed

lib/python/picongpu/picmi/interaction/ionization/__init__.py
BrianMarre: fixed

lib/python/picongpu/picmi/predefinedparticletypeproperties.py
BrianMarre: followed your suggestion

lib/python/picongpu/picmi/predefinedparticletypeproperties.py
BrianMarre: that is fine

lib/python/picongpu/picmi/simulation.py
BrianMarre: Probably, but that should be a separate PR in my opinion.

I was tempted to try [py2puml](https://pypi.org/project/py2puml/) but decided against it, PR was already large enough and I wanted to get the feature done.

TLDR: Future refactor/improvement/PR.

lib/python/picongpu/picmi/simulation.py
BrianMarre: because the interface is supposed to be newby friendly, so strings are the default option for now.

lib/python/picongpu/picmi/simulation.py
BrianMarre: because the setter is auto generated by the `build_type_safe_property` and should not be overwritten, it also follows the current code style of the picongpu PICMI interface and PICMI standard in general.

I admittedly don't particularly like the style and would prefer to change it, but following the expectation was probably better in this case.

lib/python/picongpu/picmi/species.py
BrianMarre: fixed

BrianMarre: they type annotation was wrong it should have been `list[str]`

lib/python/picongpu/picmi/simulation.py
BrianMarre: followed your suggestion

docs/source/usage/picmi/intro.rst
chillenzer: Offline discussion: Resolve after opening an issue.

lib/python/picongpu/picmi/interaction/ionization/electroniccollisonalequilibrium/thomasfermi.py
chillenzer: Offline discussion: Resolve after opening an issue to look into this further if we have resources to investigate.

lib/python/picongpu/picmi/simulation.py
chillenzer: Offline discussion: Don't open up a second way to do things.

lib/python/picongpu/picmi/species.py
chillenzer: Offline discussion: Maybe future refactoring if necessary.

mafshari64:
I reviewed the contents for PICMI collisional ionization.

mafshari64 (approved):

mafshari64 (approved):

mafshari64 (approved):

lib/python/picongpu/picmi/simulation.py
chillenzer: Offline discussion: The whole design is based on reference semantics and mutable state, so we must be concerned with instances.

mafshari64 (commented):
I reviewed the content

lib/python/picongpu/picmi/simulation.py
chillenzer: Offline discussion: This indicates a linebreak within one statement as opposed to not having the spaces meaning "new line == new statement".

mafshari64 (approved):

mafshari64 (approved):

mafshari64 (approved):

mafshari64 (commented):
I reviewed the content.

lib/python/picongpu/picmi/species.py
BrianMarre: Offline discussion: The whole design is based on reference semantics and mutable state, so we must be concerned with instances.

BrianMarre: Offline discussion: The whole design is based on reference semantics and mutable state, so we must be concerned with instances.

docs/source/usage/picmi/intro.rst
BrianMarre: issue opened

lib/python/picongpu/picmi/interaction/__init__.py
BrianMarre: will be fixed in a future PR

lib/python/picongpu/picmi/interaction/ionization/electroniccollisonalequilibrium/thomasfermi.py
BrianMarre: issue created #5012

docs/source/usage/picmi/intro.rst
BrianMarre: issue #5013

lib/python/picongpu/picmi/interaction/__init__.py
BrianMarre: issue #5014

lib/python/picongpu/picmi/simulation.py
BrianMarre: moved to separate function as requested.

BrianMarre: comment removed

lib/python/picongpu/picmi/simulation.py
BrianMarre: comment removed

lib/python/picongpu/picmi/simulation.py
BrianMarre: moved to separate function as suggested

docs/source/usage/picmi/intro.rst
```diff
@@ -48,16 +48,25 @@ After you have installed the dependencies you must include the PIConGPU PICMI im
...
 Example User Script for a warm plasma setup:
 --------------------------------------
-.. _example_PICMI_setup:
+.. _example_PICMI_setup_warm_plasma:
 
 .. literalinclude:: ../../../../share/picongpu/pypicongpu/examples/warm_plasma/main.py
    :language: python
 
-Creates a directory ``generated_input``, where you can run ``pic-build`` and subsequently ``tbg``.
+Creates a directory ``warm_plasma``, where you can run ``pic-build`` and subsequently ``tbg``.
```
chillenzer, line 60:
```suggestion
Creates a directory ``warm_plasma``, where you can run ``pic-build`` and subsequently ``tbg``. You can find this and more elaborate examples in `share/picongpu/pypicongpu/examples`.
```

docs/source/usage/picmi/intro.rst
```diff
@@ -48,16 +48,25 @@ After you have installed the dependencies you must include the PIConGPU PICMI im
...
+
+Example User Script for a laser wakefield setup:
+--------------------------------------
+.. _example_PICMI_setup_lwfa:
+
+.. literalinclude:: ../../../../share/picongpu/pypicongpu/examples/laser_wakefield/main.py
+   :language: python
+
+Creates a directory ``LWFA``, where you can run ``pic-build`` and subsequently ``tbg``.
```
chillenzer, line 70: This is quite a lot of code to scroll through for a documentation. If this level of detail is not examined and discussed in the main text, I think a reference would do. (Maybe add to the above suggestion "... more elaborate examples like an LWFA setup including ionisation...".

lib/python/picongpu/picmi/interaction/ionization/ionizationmodel.py
```diff
@@ -0,0 +1,59 @@
...
+
+from .... import pypicongpu
+
+import pydantic
+import typeguard
+import typing
+
+
+@typeguard.typechecked
+class IonizationModel(pydantic.BaseModel):
```
chillenzer, line 16: Classes inheriting from this don't implement all the methods specified by the interface.

lib/python/picongpu/picmi/interaction/ionization/ionizationmodel.py
```diff
@@ -0,0 +1,59 @@
...
+                    hash_value += hash(value)
+            except TypeError:
+                print(self)
+                print(type(self))
+                raise TypeError
+        return hash_value
+
+    def check(self):
+        # import here to avoid circular import
+        from ... import Species
```
chillenzer: Typically a symptom of a larger design issue but not for us to fix right now.

lib/python/picongpu/picmi/predefinedparticletypeproperties.py
```diff
@@ -0,0 +1,122 @@
...
+        charge=pdg.connect().get_particle_by_name("Z").charge * consts.elementary_charge,
+    ),
+    "higgs": _PropertyTuple(
+        mass=pdg.connect().get_particle_by_name("H").mass * 1e9 * consts.elementary_charge / consts.speed_of_light**2,
+        charge=pdg.connect().get_particle_by_name("H").charge * consts.elementary_charge,
+    ),
+}
+
+non_element_particle_type_properties = {}
+non_element_particle_type_properties.update(**_quarks, **_leptons, **_neutrinos, **_nucleons, **_gauge_bosons)
```
chillenzer:
```suggestion
non_element_particle_type_properties = {**_quarks, **_leptons, **_neutrinos, **_nucleons, **_gauge_bosons}
```

lib/python/picongpu/pypicongpu/species/constant/constant.py
```diff
@@ -45,8 +45,17 @@ class Constant(RenderedObject):
...
+        """two constants are equal if the have the same attributes values"""
+        for key in self.__dict__.keys():
+            if key not in other.__dict__:
+                return False
+
+        for key, value in self.__dict__.items():
+            if value != other.__dict__[key]:
+                return False
+
+        return True
```
chillenzer:
```suggestion
    def __eq__(self, other) -> bool:
        """two constants are equal if they have the same attributes values"""
        return set(self.__dict__.items()) == set(other.__dict__.items())
```

lib/python/picongpu/pypicongpu/species/initmanager.py
```diff
@@ -68,6 +68,10 @@ def __init__(self) -> None:
         self.all_operations = []
         self.__baked = False
 
+    def __hash__(self) -> int:
+        # every simulation may only ever have one InitManager
+        return hash(type(self))
+
```
chillenzer, line 74: Bold move! Let's hope that nobody ever uses this in any other context.

lib/python/picongpu/picmi/interaction/interaction.py
```diff
@@ -0,0 +1,147 @@
...
+                    if constant != constant_new:
+                        # same type of constant but conflicting values
+                        raise ValueError(f"Constants {constant} and {constant_new} conflict with each other")
+
+            if not exists_already:
+                new_constant_list.append(constant_new)
+            # ignore already existing constants
+
+        # update constant_list
+        existing_list.extend(new_constant_list)
```
chillenzer: Could this be sufficient?
```python
novel_elements = set(new_list).difference(existing_list)
if len(inconsistent_duplicates := set(map(type, novel_elements)).intersection(map(type, existing_list))) > 0:
    raise ValueError(f"The following constants are inconsistent: {inconsistent_duplicates}!")
existing_list.extend(novel_elements)
```
I think, in general, I would have reached for a `set` instead of a `list` as a data structure here.

lib/python/picongpu/pypicongpu/species/constant/groundstateionization.py
```diff
@@ -0,0 +1,93 @@
...
+    def get_constant_dependencies(self) -> list[type]:
+        """get all constants one of the ionization models in ionization_model_list depends on"""
+        total_constant_dependencies = []
+        for ionization_model in self.ionization_model_list:
+            constant_dependencies = ionization_model.get_constant_dependencies()
+            for constant in constant_dependencies:
+                if constant not in total_constant_dependencies:
+                    total_constant_dependencies.append(constant)
+
+        return total_constant_dependencies
```
chillenzer, line 84: These are all identical in structure. They could be one-liners like this
```python
from functools import reduce
return reduce(set.union, (model.get_constant_dependencies() for model in self.ionization_model_list), set())
```

lib/python/picongpu/pypicongpu/species/species.py
```diff
@@ -75,7 +95,7 @@ def check(self) -> None:
         # position
         if Position not in [type(a) for a in self.attributes]:
             raise ValueError("Each species must have the position attribute!")
-        # momentum
+        # momentum, @todo really necessary?, Brian Marre, 2024
```
chillenzer, line 98: What does the answer to this depend on? Would this be a requirement from PIConGPU main code? If so, @psychocoderHPC  should be able to answer. Or from PICMI?

share/picongpu/pypicongpu/examples/laser_wakefield/.gitignore
```diff
@@ -0,0 +1 @@
+LWFA
```
chillenzer, line 2: I presume...
```suggestion
# This folder is produced upon execution:
LWFA
```

share/picongpu/pypicongpu/examples/warm_plasma/.gitignore
chillenzer, line 1: As above.

test/python/picongpu/compiling/compileexamples.py
chillenzer, line 1: Clever! I like it!

test/python/picongpu/quick/pypicongpu/species/species.py
```diff
@@ -1,4 +1,4 @@
-"""
+"""sshfs fwk394:/home/marre55 ~/mnt/fwk394
```
chillenzer: ???

chillenzer (dismissed):
Looks good. I'd say nothing blocking it from merging but there are a few things that could be more elegant.

lib/python/picongpu/picmi/interaction/ionization/ionizationmodel.py
chillenzer: Offline discussion: There are a few layers of inheritance I overlooked.

`IonizationModel` -> `GroundStateIonizationModel` -> `FieldIonizationModel` -> actual implementations like `BSI`...

lib/python/picongpu/picmi/interaction/ionization/ionizationmodel.py
chillenzer: Offline discussion:
```suggestion
        # import here to avoid circular import that stems from projecting different species types from PIConGPU onto the same `Species` type in PICMI 
        from ... import Species
```

lib/python/picongpu/pypicongpu/species/species.py
BrianMarre: This is a requirement set by PyPIConGPU currently.
PIConGPU does not care and delegates responsibility to the user for choosing the correct attributes.

In theory it would be imaginable have species without a momentum, i.e. un-moving probe-particles, but that is not currently supported by PyPIConGPU because Hannes decided to make the momentum attribute mandatory as there is now way to specify a species as unmoving in the PICMI standard.

test/python/picongpu/compiling/compileexamples.py
BrianMarre: thanks!

docs/source/usage/picmi/intro.rst
BrianMarre: included

lib/python/picongpu/picmi/interaction/ionization/ionizationmodel.py
BrianMarre: good summary, updated comment as suggested

lib/python/picongpu/picmi/predefinedparticletypeproperties.py
BrianMarre: changes as suggested

lib/python/picongpu/pypicongpu/species/constant/constant.py
BrianMarre: changed as suggested

lib/python/picongpu/pypicongpu/species/initmanager.py
BrianMarre: Well it's documented ;)

lib/python/picongpu/picmi/interaction/interaction.py
BrianMarre: We want to explicitly allow a two instances of the type, so long as they do not differ in value wise.

This allows different ionization models to share constants and still be translate themselves independently, i.e. generate independent instances.

This is supposed to mimic the type definition of a single constant in PIConGPU where the users manages a global set of particle flags (called constants in PyPIConGPU) shared by all algorithms while making sure that every constant required is actually present, conflicting requirements correctly cause an exception and constants are shared as much as possible.

I generally agree that using a set would be better, but PyPIConGPU uses list for constant as we did not anticipate the same constant being needed more than once.

I would like to postpone this refactor to a later PR, to not create additional work for me right now.

BrianMarre: And no, your suggestion is not equivalent, since it does not allow different instances of the same type with the same attribute values. :grinning:

share/picongpu/pypicongpu/examples/laser_wakefield/.gitignore
BrianMarre: good suggestion, implemented

share/picongpu/pypicongpu/examples/warm_plasma/.gitignore
BrianMarre: done

lib/python/picongpu/pypicongpu/species/constant/groundstateionization.py
BrianMarre: would shift this refactor to a future PR.

lib/python/picongpu/picmi/interaction/interaction.py
chillenzer: Offline discussion: It does.

test/python/picongpu/quick/pypicongpu/species/species.py
BrianMarre: typo, removed

chillenzer:
Just for the record: @PrometheusPi takes over the merge after verifying that the rendered input files look correct.

docs/source/usage/picmi/intro.rst
BrianMarre: Will keep it for now, but will create an issue to improve this at a later point.

chillenzer (dismissed):

lib/python/picongpu/picmi/distribution/Distribution.py
```diff
@@ -32,6 +33,7 @@
 """
 
 
+@typeguard.typechecked
```
PrometheusPi, line 36: Was the check just missing before?

PrometheusPi (dismissed):
Currently, calling `tbg` as `tbg -s -t -c etc/picongpu/N.cfg run` fails. This needs further investigation. I will block this PR for now (just to avoid accidental merging.)

PrometheusPi:
The `tbg` issue is not caused by the pull request but a general bug in PIConGPU right now: #5026
Thus, I removed my "changes required".

lib/python/picongpu/picmi/distribution/Distribution.py
BrianMarre: yes, I thought that pydantic also automatically checked method signatures upon call of an method, but as it turns out it does not.
We therefore also need to annotate the class with `typeguard.typechecked`

PrometheusPi (changes requested):
I tested the LWFA PICMI setup. 
With electrons only, both the original `dev` and PICMI show the same physics. 

With the default PICMI = case 2 with CMake prefixes, which includes ions and electrons, the results look not totally but significantly different. 

I dought that this is a PICMI issue but an error on my side in understanding the setup details (since they differ in a few points regrading naming etc.) 

the electron acceleration in LWFA dev looks as 
grafik

While in PICMI, the electron acceleration undergoes oscillation:
grafik

Until I find the cause, I will block merging.

PrometheusPi:
I looked into the pngs at iteration 3520, and they differ (not only in color) but in where the elecrons ae created:
`dev`:
grafik

`picmi`:
grafik

PrometheusPi:
In contrast to a previous (now corrected) post, I do not belief that the error is with PIConGPU dev case 2, but with picmi. 

The PICMI setup states:
```python
ENABLE_IONS = True
ENABLE_IONIZATION = True
```

But, while there are ions included, according to the macro particle counter, electrons are initially present which should not be the case for the `ENABLE_IONIZATION = True` (as long as the laser in not there):

grafik

BrianMarre:
The example mistakenly always added electrons even if ionization was activated. Fixed

PrometheusPi:
@BrianMarre You last commit now causes an error when running the script: 
```
$ python lwfa_picmi_setup2.py
  File "/...//_pr_5007/lwfa_picmi_setup2.py", line 204, in <module>
    sim.write_input_file(OUTPUT_DIRECTORY_PATH)
  File "/...//_pr_5007/picongpu/lib/python/picongpu/picmi/simulation.py", line 394, in write_input_file
    pypicongpu_simulation = self.get_as_pypicongpu()
  File "/...//_pr_5007/picongpu/lib/python/picongpu/picmi/simulation.py", line 458, in get_as_pypicongpu
    s.init_manager = self.__get_init_manager()
  File "/...//_pr_5007/picongpu/lib/python/picongpu/picmi/simulation.py", line 361, in __get_init_manager
    self.__fill_in_ionization_electrons(pypicongpu_by_picmi_species, ionization_model_conversion_by_species)
  File "/...//_pr_5007/picongpu/lib/python/picongpu/picmi/simulation.py", line 339, in __fill_in_ionization_electrons
    self.picongpu_interaction.fill_in_ionization_electron_species(
  File "/...//_pr_5007/picongpu/lib/python/picongpu/picmi/interaction/interaction.py", line 126, in fill_in_ionization_electron_species
    pypicongpu_ionization_electron_species = pypicongpu_by_picmi_species[
KeyError: <picongpu.picmi.species.Species object at 0x2aaab29bbd90>
```

BrianMarre:
@PrometheusPi forgot to add a non placed electron species in the enable ionization case, fixed the example script and improved the error message printed to the user to make the error easier to understand.

lib/python/picongpu/picmi/interaction/interaction.py
```diff
@@ -123,9 +123,16 @@ def fill_in_ionization_electron_species(
...
-                        picmi_ionization_model.ionization_electron_species
-                    ]
+                    try:
+                        pypicongpu_ionization_electron_species = pypicongpu_by_picmi_species[
+                            picmi_ionization_model.ionization_electron_species
+                        ]
+                    except KeyError:
+                        raise ValueError(
+                            f"Ionization electron species of {picmi_ionization_model} not known to simulation."
+                            + f"Please add species {picmi_ionization_model.ionization_electron_species.name} to"
```
PrometheusPi, line 134: at leat 2 spaces missing 
Perhaps add a newline before "Please ... simulation setup."

PrometheusPi:
Your second change fixed the setup.

BrianMarre:
One CI job failed to acess the github repository, @PrometheusPi coulde you please restart it?

docs/source/usage/picmi/intro.rst
BrianMarre: see Issue #5022

docs/source/usage/picmi/intro.rst
BrianMarre: see Issue #5056

lib/python/picongpu/picmi/simulation.py
BrianMarre: see Issue #5057

lib/python/picongpu/picmi/interaction/interaction.py
```diff
@@ -0,0 +1,154 @@
...
+        new_list: list[pypicongpu.species.constant.Constant],
+    ) -> None:
+        """check if dicts may be merged without overwriting previously set values"""
+
+        new_constant_list = []
+
+        for constant_new in new_list:
+            exists_already = False
+            for constant in existing_list:
+                if type(constant) == type(constant_new):
```
PrometheusPi, line 51: Why do you just check for type comparison?

lib/python/picongpu/picmi/interaction/interaction.py
PrometheusPi: Since the offline discussion turned out to show that @chillenzer solution**is equvilant**: What is the status of this request @BrianMarre?
(I will make this as unresolved for now.)

lib/python/picongpu/picmi/interaction/interaction.py
```diff
@@ -0,0 +1,154 @@
...
+        pypicongpu_by_picmi_species: dict[picmistandard.PICMI_Species, pypicongpu.species.Species],
+        ionization_model_conversion_by_type_and_species: dict[
+            picmistandard.PICMI_Species,
+            None | dict[IonizationModel, pypicongpu.species.constant.ionizationmodel.IonizationModel],
+        ],
+    ) -> None:
+        """
+        add ionization models to pypicongpu species
+
+        in PICMI ioniaztion is defined as a list ionization models owned by the simulation, with each ionization model
```
PrometheusPi: What do you mean with "owned by the simulation"?

lib/python/picongpu/picmi/interaction/interaction.py
```diff
@@ -0,0 +1,154 @@
...
+        In contrast in PyPIConGPU each ion PyPIConGPU species owns a list of ionization models, each storing its
+        PyPIConGPU ionization electron species.
+
+        This creates the problem that upon translation of the PICMI species to an PyPIConGPU species the PyPIConGPU
+        ionization electron species might not exist yet.
+
+        Therefore we leave the ionization electron unspecified upon species creation and fill it in from the PICMI
+        simulation ionization model list later.
+
+        (An because python uses pointers, this will be applied to the existing species objects passed in
```
PrometheusPi: typo:
```suggestion
        (And because python uses pointers, this will be applied to the existing species objects passed in
```

lib/python/picongpu/picmi/interaction/ionization/fieldionization/ADK.py
```diff
@@ -0,0 +1,47 @@
...
+
+@typeguard.typechecked
+class ADKVariant(enum.Enum):
+    LinearPolarization = 0
+    CircularPolarization = 1
+
+
+@typeguard.typechecked
+class ADK(FieldIonization):
+    """Barrier Suppression Ioniztion model"""
```
PrometheusPi: typo:
```suggestion
    """Barrier Suppression Ionization model"""
```

lib/python/picongpu/picmi/interaction/ionization/fieldionization/ADK.py
```diff
@@ -0,0 +1,47 @@
...
+
+    MODEL_NAME: str = "ADK"
+
+    ADK_variant: ADKVariant
+    """extension to the BSI model"""
```
PrometheusPi: What do you want to say with this comment?

lib/python/picongpu/picmi/interaction/ionization/fieldionization/ADK.py
```diff
@@ -0,0 +1,47 @@
...
+    def get_as_pypicongpu(self) -> IonizationModel:
+        self.check()
+
+        if self.ADK_variant is ADKVariant.LinearPolarization:
+            return ADKLinearPolarization(ionization_current=None_())
+        if self.ADK_variant is ADKVariant.CircularPolarization:
+            return ADKCircularPolarization(ionization_current=None_())
+
+        # unknown/unsupported ADK variant
+        pypicongpu.util.unsupported(f"ADKVariant {self.ADK_variant}")
```
PrometheusPi: Is a raise or return missing here? 
```suggestion
        raise pypicongpu.util.unsupported(f"ADKVariant {self.ADK_variant}")
```

lib/python/picongpu/picmi/interaction/ionization/fieldionization/BSI.py
```diff
@@ -0,0 +1,48 @@
...
+
+import enum
+import typeguard
+
+
+@typeguard.typechecked
+class BSIExtension(enum.Enum):
+    StarkShift = 0
+    EffectiveZ = 1
+    # consider_excitation = 2
```
PrometheusPi: please do not use commented out code

lib/python/picongpu/picmi/interaction/ionization/fieldionization/BSI.py
```diff
@@ -0,0 +1,48 @@
...
+from .....pypicongpu.species.constant.ionizationcurrent import None_
+from .....pypicongpu.species.constant import ionizationmodel
+
+import enum
+import typeguard
+
+
+@typeguard.typechecked
+class BSIExtension(enum.Enum):
+    StarkShift = 0
```
PrometheusPi, line 20: Why do you use class variables (static member variables). I might have an instance with one version for one species (in PIConGPU) and another version for a different species. With class variables, you only allow for one.

lib/python/picongpu/picmi/interaction/ionization/fieldionization/BSI.py
```diff
@@ -0,0 +1,48 @@
...
+            return ionizationmodel.BSI(ionization_current=None_())
+
+        if len(self.BSI_extensions) > 1:
+            pypicongpu.util.unsupported("more than one BSI_extension, will use first entry only")
+
+        if self.BSI_extensions[0] is BSIExtension.StarkShift:
+            return ionizationmodel.BSIStarkShifted(ionization_current=None_())
+        if self.BSI_extensions[0] is BSIExtension.EffectiveZ:
+            return ionizationmodel.BSIEffectiveZ(ionization_current=None_())
+        raise ValueError(f"unknown BSI_extension {self.BSI_extensions[0]}")
```
PrometheusPi: differs from how you treat this kind of undefined behavior in ADK - please make this consistent

lib/python/picongpu/picmi/interaction/ionization/fieldionization/keldysh.py
```diff
@@ -0,0 +1,25 @@
...
+
+from .....pypicongpu.species.constant.ionizationcurrent import None_
+from .....pypicongpu.species.constant import ionizationmodel
+
+import typeguard
+
+
+@typeguard.typechecked
+class Keldysh(FieldIonization):
+    """Barrier Suppression Ioniztion model"""
```
PrometheusPi, line 18: typo

lib/python/picongpu/picmi/interaction/ionization/ionizationmodel.py
```diff
@@ -0,0 +1,59 @@
...
+        return hash_value
+
+    def check(self):
+        # import here to avoid circular import that stems from projecting different species types from PIConGPU onto the same `Species` type in PICMI
+        from ... import Species
+
+        assert isinstance(self.ion_species, Species), "ion_species must be an instance of the species object"
+        assert isinstance(
+            self.ionization_electron_species, Species
+        ), "ionization_electron_species must be an instance of species object"
```
PrometheusPi: to be consistent with the previous error:
```suggestion
        ), "ionization_electron_species must be an instance of the species object"
```

lib/python/picongpu/picmi/predefinedparticletypeproperties.py
```diff
@@ -0,0 +1,121 @@
...
+        mass=pdg.connect().get_particle_by_name("Z").mass * 1e9 * consts.elementary_charge / consts.speed_of_light**2,
+        charge=pdg.connect().get_particle_by_name("Z").charge * consts.elementary_charge,
+    ),
+    "higgs": _PropertyTuple(
+        mass=pdg.connect().get_particle_by_name("H").mass * 1e9 * consts.elementary_charge / consts.speed_of_light**2,
+        charge=pdg.connect().get_particle_by_name("H").charge * consts.elementary_charge,
+    ),
+}
+
+non_element_particle_type_properties = {**_quarks, **_leptons, **_neutrinos, **_nucleons, **_gauge_bosons}
```
PrometheusPi, line 125: I did not check for correctness since:
- most of the particles will not be used in PIConGPU within the next decade
- you are planning to read in data automatically (on a shorter time scale)

lib/python/picongpu/pypicongpu/runner.py
```diff
@@ -333,7 +333,12 @@ def __run(self):
         chdir(self.setup_dir)
         runArgs(
             "PIConGPU",
-            "tbg -s bash -c etc/picongpu/N.cfg -t " "etc/picongpu/bash/mpiexec.tpl".split(" ") + [self.run_dir],
+            (
+                ("tbg -s bash -c etc/picongpu/N.cfg -t " + environ["PIC_SYSTEM_TEMPLATE_PATH"] + "/mpiexec.tpl").split(
```
PrometheusPi: What does the split do here?
(You also do not need `" "` as an argument, since space is the default.)

lib/python/picongpu/pypicongpu/runner.py
PrometheusPi, line 1: This change has nothing to do with the ionization. 
I assume this is a test, right?

lib/python/picongpu/pypicongpu/rendering/renderedobject.py
PrometheusPi, line 1: What does this file changes with respect to ionization?

lib/python/picongpu/pypicongpu/simulation.py
```diff
@@ -99,7 +99,14 @@ def __found_custom_input(self, serialized: dict):
             + "\t WARNING: custom templates are required if using custom user input.\n"
         )
 
+    def check(self):
+        """check validity of self"""
+        if self.typical_ppc < 1:
+            raise ValueError("typical_ppc must be >= 1")
+
```
PrometheusPi: This is a check independent of ionization.

chillenzer:
@BrianMarre @PrometheusPi What's the status of this?

BrianMarre:
Richard is still in the process of doing an intial review, then I will split up the PR in several independent PRs. At least that is the current plan

chillenzer:
Thanks for the update. I'll keep quiet then.

BrianMarre:
Well ... since you ask, you could take over part of the review. ;)

psychocoderHPC:
I set this PR to draft that it not get merged see: https://github.com/ComputationalRadiationPhysics/picongpu/pull/5007#issuecomment-2341357981

BrianMarre:
rebased PR against current dev

lib/python/picongpu/pypicongpu/simulation.py
PrometheusPi: Please add it in a separate pull request.

lib/python/picongpu/picmi/simulation.py
```diff
@@ -100,56 +179,10 @@ def __yee_compute_cfl_or_delta_t(self) -> None:
...
-        self.__runner = None
-
-        self.__electron_species = None
-
     def __get_operations_simple_density(
         self,
-        pypicongpu_by_picmi_species: typing.Dict[picmistandard.PICMI_Species, species.Species],
-    ) -> typing.List[species.operation.SimpleDensity]:
+        pypicongpu_by_picmi_species: typing.Dict[Species, pypicongpu.species.Species],
+    ) -> typing.List[pypicongpu.species.operation.SimpleDensity]:
```
PrometheusPi, line 185: This change has nothing to do with adding ionization or an LWFA example, right?
Please add it in a separate PR.

PrometheusPi: Sorry - this is just the general renaming that was introduced with this PR as well. You can keep it in the ionization PR. (Just started to get into reviewing your PR again.)

lib/python/picongpu/picmi/simulation.py
```diff
@@ -484,22 +431,22 @@ def get_as_pypicongpu(self) -> simulation.Simulation:
         else:
             raise ValueError("runtime not specified (neither as step count nor max time)")
 
-        util.unsupported("verbose", self.verbose)
-        util.unsupported("particle shape", self.particle_shape, "linear")
-        util.unsupported("gamma boost", self.gamma_boost)
+        pypicongpu.util.unsupported("verbose", self.verbose)
+        pypicongpu.util.unsupported("particle shape", self.particle_shape, "linear")
+        pypicongpu.util.unsupported("gamma boost, use picongpu_moving_window_move_point instead", self.gamma_boost)
```
PrometheusPi: Extension of warning/error: not part of neither LWFA example nor ionization. Please move to separate PR.

share/picongpu/pypicongpu/examples/laser_wakefield/.gitignore
PrometheusPi, line 1: This should go to a LWFA example PR

share/ci/bash.profile
```diff
@@ -22,6 +22,9 @@ export CMAKE_PREFIX_PATH=$ADIOS2_ROOT:$CMAKE_PREFIX_PATH
 export PATH=$ADIOS2_ROOT/bin:$PATH
 export LD_LIBRARY_PATH=$ADIOS2_ROOT/lib:$LD_LIBRARY_PATH
 
+# set environment variable for path to tpls for PyPIConGPU runner
```
PrometheusPi, line 25: What has this change to do with LWFA example or ionization methods?

share/picongpu/pypicongpu/examples/laser_wakefield/main.py
PrometheusPi, line 1: This should go to a LWFA example PR

lib/python/picongpu/pypicongpu/species/util/element.py
```diff
@@ -7,65 +7,94 @@
...
+    def parse_openpmd_isotopes(openpmd_name: str) -> tuple[int | None, str]:
+        if openpmd_name == "":
+            raise ValueError(f"{openpmd_name} is not a valid openPMD particle type")
+        if openpmd_name[0] != "#" and re.match(r"[A-Z][a-z]?$|n$", openpmd_name):
+            return None, openpmd_name
+
+        m = re.match(r"#([1-9][0-9]*)([A-Z][a-z]?)$", openpmd_name)
+
+        if m is None:
+            raise ValueError(f"{openpmd_name} is not a valid openPMD particle type")
```
PrometheusPi, line 46: What are valid openPMD particle types?

share/picongpu/pypicongpu/examples/laser_wakefield/main.py
```diff
@@ -0,0 +1,208 @@
...
+import numpy as np
+
+"""
+@file PICMI user script reproducing the PIConGPU LWFA example
+
+This Python script is example PICMI user script reproducing the LaserWakefield example setup, based on 8.cfg.
+"""
+
+# generation modifiers
+ENABLE_IONS = True
```
PrometheusPi, line 19: If this suppose to reproduce the LWFA example case 0, please deactivate ions and ionization.

share/picongpu/pypicongpu/examples/warm_plasma/.gitignore
PrometheusPi, line 1: another example that should go to a separate PR.

share/picongpu/pypicongpu/examples/warm_plasma/main.py
PrometheusPi, line 1: seperate PR

test/python/picongpu/.gitignore
```diff
@@ -1,2 +1,4 @@
 # generated by coverage.py
 .coverage
+LWFA
+warm_plasma
```
PrometheusPi, line 4: why this gitignore here again, haven't you added them in the examples already?

test/python/picongpu/compiling/compileexamples.py
```diff
@@ -0,0 +1,45 @@
...
+
+        return sim
+
+    def build_simulation(self, sim):
+        """build the given instance of simulation"""
+        runner = pypicongpu.Runner(sim)
+        runner.generate(printDirToConsole=True)
+        runner.build()
+
+    def test_LWFA_example(self):
```
PrometheusPi, line 33: You could add this as part of the LWFA example PR.

test/python/picongpu/quick/pypicongpu/species/util/element.py
```diff
@@ -7,65 +7,90 @@
...
-    def test_periodic_table_names(self):
-        """names must follow the periodic table"""
-        element_re = re.compile("^[A-Z][a-z]?$")
-        for element in list(Element):
-            self.assertTrue(element_re.match(element.name))
+    def setUp(self):
+        # create test case data
+        self.test_element = ["H", "#2H", "Cu", "#12C", "C"]
+        self.name = ["H", "D", "Cu", "C", "C"]
+        self.picongpu_names = ["Hydrogen", "Deuterium", "Copper", "Carbon", "Carbon"]
```
PrometheusPi, line 21: Why are there 2 Carbons (same name) with slightly different mass?

PrometheusPi: Okay - #12C vs C

test/python/picongpu/quick/pypicongpu/species/util/element.py
PrometheusPi, line 1: This is a test for how to handle species with new ionization attributes - please keep in ionization PR

test/python/picongpu/quick/pypicongpu/species/constant/groundstateionization.py
```diff
@@ -0,0 +1,142 @@
+"""
+This file is part of PIConGPU.
+Copyright 2021-2023 PIConGPU contributors
```
PrometheusPi: minor: in most other places you updated the year

PrometheusPi (changes requested):
This pull request contains:
- minor code fixes to the existing code base
- introduction of an LWFA example (that is also used as a potential test case)
- change to a more ionization models
- test for species setup with new ionization attributes
- restructuring of how to access pypicongpu 
- and potentially more ...

While separating e.g. access to pypicongpu and ionization will be hard to separate, I would strongly suggest separating the above parts as good as possible that they can be:
 - reviewed separately 
 - are documented as different topics both in the git history and as GitHub PR

PrometheusPi:
FYI: do to the sheer size of this PR, I just took a brief look at the later files as I already saw that his needs to be splitt up anyway.

lib/python/picongpu/picmi/simulation.py
BrianMarre: opened issue #5137 to keep track of this.

lib/python/picongpu/picmi/predefinedparticletypeproperties.py
BrianMarre: I opened issue #5138 to track this.

lib/python/picongpu/picmi/interaction/interaction.py
BrianMarre: Offline discussion result was that we will keep the current implementation for readability reasons.

lib/python/picongpu/pypicongpu/species/constant/groundstateionization.py
BrianMarre: opened Issue #5139 to track this

lib/python/picongpu/picmi/interaction/interaction.py
BrianMarre: fixed by adding a linebreak after the first sentence and a space before to.

lib/python/picongpu/picmi/interaction/interaction.py
BrianMarre: I do an equality check in the if-body to check for conflicting member values.

The idea is to first check whether we already have an instance of the class to decide whether we need to append the instance to the accumulation list and secondarily check whether if we already have an instance of the class if their settings are the same and if they are not we throw an error.

lib/python/picongpu/picmi/interaction/interaction.py
BrianMarre: "Owned by A" as in is solely associated with object A, instead of two or more object holding reference to it.

Also the documentation string was somewhat outdated, since the list of ionization models is actually owned by the interaction object which is in turn a member of the simulation object.
Updated the comment to reflect the current state

lib/python/picongpu/picmi/interaction/interaction.py
BrianMarre: fixed

lib/python/picongpu/picmi/interaction/interaction.py
BrianMarre: Request was moved to the "will be fixed in future" category and I opened Issue #5139 to keep track of it

lib/python/picongpu/picmi/interaction/ionization/fieldionization/ADK.py
BrianMarre: fixed

lib/python/picongpu/picmi/interaction/ionization/fieldionization/ADK.py
BrianMarre: Good Catch, copy paste error, fixed by changing to """ADK model variant specification""" also fixed the class documentation comment, changed to """ADK Tunneling Ionization model""", was still refering to BSI.

lib/python/picongpu/picmi/interaction/ionization/fieldionization/ADK.py
BrianMarre: You are right, switched to raising a Value Error since there is now way to recover from an unknown ADK Variant.

lib/python/picongpu/picmi/interaction/ionization/fieldionization/BSI.py
BrianMarre: good catch, comment removed

lib/python/picongpu/picmi/interaction/ionization/fieldionization/BSI.py
BrianMarre: done ADK should also have thrown an error.

lib/python/picongpu/picmi/interaction/ionization/fieldionization/BSI.py
BrianMarre: This is intended.

I am using the enum to define the specified BSI Extensions in code.

lib/python/picongpu/picmi/interaction/ionization/fieldionization/keldysh.py
BrianMarre: fixed

lib/python/picongpu/picmi/interaction/ionization/ionizationmodel.py
BrianMarre: fixed

lib/python/picongpu/pypicongpu/runner.py
BrianMarre: runArgs needs the entire command as a list of strings and we trie to keep it readable by first creating the string as one would type it and then split it into a list as requited by runArgs.

BrianMarre: removed the specification of a space as separation character as suggested.

lib/python/picongpu/pypicongpu/runner.py
BrianMarre: I ran into the problem that I could not execute the tests locally since my tpl path was not equal to previously hard coded the CI template path.
I therefore made the switch to getting the `tpl` path form the bash environment, essentially a fix.

lib/python/picongpu/pypicongpu/rendering/renderedobject.py
BrianMarre: adds a hash function to allow storing a lists in and still use them as a dictionary key.

Specifically this was necessary since we use PyPIConGPU Species objects as keys in the Operation member attributes_by_species and with the addition of ionization I added to the species a GroundStateIonization object which stores a list of IonizationModels as a member.

Without the custom hash fucntion this would lead to a ValueError since lists are mutable and therefore non hashable

lib/python/picongpu/pypicongpu/simulation.py
BrianMarre: done

lib/python/picongpu/picmi/simulation.py
BrianMarre: It actually a little bit more than a renaming.

I changed from using the PICMI species class to mandating PIConGPU inheriting PICMI interface species objects since we rely on methods only present in the daughter class.

lib/python/picongpu/picmi/simulation.py
BrianMarre: done

lib/python/picongpu/pypicongpu/species/util/element.py
BrianMarre: as defined in [openPMD upcoming 2.0.0 particle type extension](https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/EXT_SpeciesType.md) under atoms and isotopes

share/picongpu/pypicongpu/examples/laser_wakefield/.gitignore
BrianMarre: Unfortunately the LWFA-example is the only test of the configuration of ionization.

Therefore removing it would remove all test coverage of the PICMI ionization configuration.

share/picongpu/pypicongpu/examples/laser_wakefield/main.py
BrianMarre: Unfortunately the LWFA-example is the only test of the configuration of ionization.

Therefore removing it would remove all test coverage of the PICMI ionization configuration something I am hesitant to do.

share/picongpu/pypicongpu/examples/laser_wakefield/.gitignore
BrianMarre: see below main.py

share/ci/bash.profile
BrianMarre: Only as in necessary to get the CI to pass.

The CI does not set a `PIC_SYSTEM_TEMPLATE_PATH` by default causing the PICMI run time tests to fail due to the switch to using the bash environment for `tpl` file resolution in the runner. It was therefore necessary to add setting it by hand to the CI-environment init script.

share/picongpu/pypicongpu/examples/laser_wakefield/main.py
BrianMarre: No the default is supposed to represent the case with ions.

Since we use the LWFA example to test the Ionizer configuration we may not disable ionization in the example, which in turn requires ions.

share/picongpu/pypicongpu/examples/warm_plasma/.gitignore
BrianMarre: example did exist previously but was slightly tweaked to fit structure expected by the example compile test.

share/picongpu/pypicongpu/examples/warm_plasma/main.py
BrianMarre: as above

test/python/picongpu/.gitignore
BrianMarre: this ensure that the directories generated upon executing the the example test by hand are ignored by git, In contrast the example git ignore are used to ignore the directories generated when executing the examples from a copy or directly in PIConGPU.
The setup folder will always be generated at the call location by PyPIConGPU therefore we need to provide two different git ignore.

BrianMarre: discussed offline, current solution is fine

test/python/picongpu/compiling/compileexamples.py
BrianMarre: Yes, but since it currently is the only test coverage I am hesitant to do that.

test/python/picongpu/quick/pypicongpu/species/util/element.py
BrianMarre: exactly, Carbon-12 vs natural isotope abundance Carbon

test/python/picongpu/quick/pypicongpu/species/util/element.py
BrianMarre: I concur ;)

BrianMarre:
Rebased to the current dev,

All reviewer comment changes are in last commit for easier review of last changes.

test/python/picongpu/quick/pypicongpu/species/constant/groundstateionization.py
BrianMarre: fixed

BrianMarre:
Switched from `pdg` to `particle` due to `pdg` using the `distutils` removed with python 3.12 causing  installs of `pdg` to fail

lib/python/picongpu/picmi/distribution/Distribution.py
```diff
@@ -9,6 +9,7 @@
 
 import typing
 import pydantic
+import typeguard
```
PrometheusPi, line 12: The typeguard check could be separated for those files that have not been added newly with thid PR.

PrometheusPi: Okay - these changes have mainly been introduced with the [`fixes` commit](https://github.com/ComputationalRadiationPhysics/picongpu/pull/5007/commits/dc564f171a6875081ff9e00f97267b8074dcdea2) and might be hard to separate. If @psychocoderHPC agrees in committing them together, keep it as one PR.

BrianMarre:
@psychocoderHPC please restart the Ci jobs, the failed due to an un specified timeout without an error log

PrometheusPi (approved):
No full check of functionality - I am fine the cleanup

Issue #4802, opened 2024-01-31
Use other alias for `openpmd_api`


chillenzer:
I've noticed in #4801 that it seems to be a convention to do
```python3
import openpmd_api as io
```
but I think this is unfortunate because `io` is [a python standard library module](https://docs.python.org/3/library/io.html) which might mislead people more familiar with the Python ecosystem than with PIConGPUs conventions. If nobody complains in the near future, I'll submit a PR changing this into something less ambiguous.

psychocoderHPC:
ping @PrometheusPi @steindev @ikbuibui @BrianMarre @pordyna What is you opinion?

steindev:
It is sensible to change this. But it is not a PIConGPU-only convention. The root cause of this unfortunate naming is openPMD's documentation. See (https://openpmd-api.readthedocs.io/en/0.15.2/usage/firstwrite.html#python).

chillenzer:
Well, I'll open an issue there, too.

pordyna:
I sometimes do `import openpmd_api as api` as it used to be earlier in the `openpmd` docs. Though, I don't think that this is a good option either. We should probably let the `openpmd` devs discuss it first.

chillenzer:
As you can see, I've opened the issue over at the openPMD repo. Let's wait for a few days and come back to this, then. Is there a "remind-me-later" button for GH issues?

BrianMarre:
@chillenzer Reminder ;)

I think
```python
import openpmd_api as opmd
```
works for our purposes.

chillenzer:
Yeah, I think that's the way to go. I'll do a final ping in the corresponding OpenPMD issue and set it on my to-do list for PIConGPU. Any final objections, @ComputationalRadiationPhysics/picongpu-maintainers ?

PrometheusPi:
What's the status of this change @chillenzer?

chillenzer:
If only, I'd know. =P

It's okay. I'll go ahead and fix it.

PR #5008, opened 2024-07-30
Add FromOpenPMDPulse profile

fabidie:
The FromOpenPMDPulse profile is an IncidentField method that reads the laser pulse from an INSIGHT measurement (at the focus position) into PIConGPU.

Since the INSIGHT measurement is done in the frequency domain, the E-field data first has to be transformed into the time domain. Afterwards, the field data chunk is saved to openPMD  and from there it will be read onto the incident field plane at the start of the simulation volume (via linear interpolation). 
Those "preparation steps" are done with a python script, which is provided under picongpu/lib/python/picongpu/extra/input. There, one can find also an example workflow which explains the preparation routine more in detail.

Currently, there are some restrictions in using this profile:
- propagation and polarization only work along the axes, **not** diagonally!
- the pulse will always be aligned centered at the incident plane
- one needs to increase the reserved GPU memory size in memory.param - since the field data chunk will be stored on each used device at timestep 0 (and not during init, before allocating the particle memory), the simulation could otherwise run into memory issues.
- the Insight measurement scales have to be mm and fs, otherwise the preparation routine has to be adapted

There have been various tests to validate this new laser profile:
#### 1. Using a gaussian pulse profile

A Gaussian pulse field chunk ($\lambda$ = 800nm, $\tau$ = 30fs, $w_0$ = 10 $\mu$ m, $z$ = $-z_R$ = -0.393mm) has been read from OpenPMD into the simulation volume and then propagated to the focus. First, the amplitude at the incident plane ("PIC") has been compared to the analytical definition of the Gaussian pulse at $z$ = $-z_R$ ("Original") for different time steps covering the pulse duration.
incidentPlaneIte600
The vertical line in the last subplot indicates the position of the incident plane at this moment. The amplitudes look similar in extent, but different in structure and amplitude, since the longitudinal sampling of both datasets is different. These small differences in space/time cause the visible differences, since the amplitude varies with $\lambda$ in the longitudinal direction.

Then, the amplitude at the focal plane ("PIC") has been compared to the analytical definition of the Gaussian pulse at the focus ("Original"), also for different time steps covering the pulse duration.
focusPlaneIte5666
(amplitude in the focal plane)
focusDiffIte5666
(transversely integrated amplitude, the black line represents the focus position)
xzPlaneIte5666
(amplitude in propagation direction, the blue lines represent the focus position)
Here, the lines again represent the focal plane position for this certain time step. Both pulses look similar in extent and amplitude.

Furthermore, the beam energy (density) was investigated. The original input beam had an energy of 4.5 J and the measured beam energy in the focus  (shown in the pictures above) was a bit below with 4.3 J. The result could probably be improved with a finer sampling. Also, the energy density in the focus (integrated along the propagation axis)  looks similar in extent and value (and is probably a better indicator than the beam amplitudes at incident and focal plane, since the effects of different spacing vanish due to the integration):
energyDensityIte5666
The energy was calculated according to $W = \int dV \frac{\epsilon_0}{2} (E^2 + c^2 B^2)$ and the energy density $\mathcal{W} = \int dz \frac{\epsilon_0}{2} (E^2 + c^2 B^2)$.

Finally, the simulation has been re-run with the GaussianPulse profile (and the same parameters) to compare it to the former results (they looked the same).

#### 2. Using a Gauss-Laguerre pulse
According to the definition of a Gauss-Laguerre pulse in the GaussianPulse profile, a Gauss-laguerre pulse field chunk has been written to OpenPMD and read into PIConGPU. The pulse parameters were the same as for the Gaussian pulse above, just $\tau$ was set shorter to $30/2\sqrt{2 ln 2}$ fs. As Laguerre coefficients, the example set in GaussianPulse.def was used. For two different positions, the simulation results ("PIC") were compared to the analytical solution ("Original"):
**a) at $z \approx -z_R$**
xyPlaneIte1285
(amplitude along the longitudinal direction)
energyDensityIte1285
(energy density, definition see above)
**b) at the focus position**
xyPlaneIte5597
(amplitude along the longitudinal direction,  the blue line is the focus position at this time step)
energyDensityIte5597
(energy density, definition see above)
focusDiffIte5597
(transversal integrated amplitude,  the black line is the focus position at this time step)
In the above mentioned simulations, the transversal simulation cell was square-shaped, as well as the transversal extent of the test pulse field chunk. To test also the asymmetrical case, the simulation cell in $x$ direction was halved in size. The results were similar: 
energyDensityIte5649
(energy density for an asymmetrical cell size; the asymmetry in amplitude shows that the energy is conserved better for a smaller cell size)
Then, the cell size was reset to original (transversal square-shaped), but the (original) field chunk was chosen to be asymmetrical:
originalChunkExtent
(original Gauss-Laguerre field at $z=-z_R$, before reading it into the simulation)
The resulting simulation did behave just the same as above, e.g. the pulse was still centered and not distorted after propagating it into the focus:
yxPlaneIte5597
yzPlaneIte5597
(amplitude in longitudinal direction, along both transversal directions; blue is the focus position)
energyDensityIte5597
( energy density)

#### 3. Using an Insight measurement

For that, the measured field data has been propagated out of the focus (-150, -300, -500 and -1280 (= $-z_R$ ) $\mu$ m, during the preparation routine), saved to OpenPMD and then read into the simulation. This has been done for different propagation and polarization directions, for different (transversal) simulation window extents, for different transversal grid spacing ($w_0/6$ and $w_0/10$) and for a different number of devices (1, 2, 4).
The field at the incident plane ("PIC") has been compared to the read-in ("Original") data:
incidentPlaneIte1800
(propagation in z direction, polarization in x direction, just as the original)
incidentPlaneNegPolIte1800
(propagation in z direction, polarization in -x direction; this automatically also swaps the y axis to keep the original pulse orientation)
xzPlaneNegPropIte1800
(propagation in -z direction, polarization in x direction; the time axis of the original data has been transformed to a spatial one by multiplication with $c$ to allow a better comparison)

For z = -300 $\mu$ m, the pulse has been propagated in PIConGPU into the focus and then compared there to the Insight measurement at the focus: 
focusDiffIte4400
(amplitude at the beam center in propagation direction)
focusPlaneIte4400
(amplitude in focal plane)
xzPlaneIte4500
(amplitude in longitudinal direction, blue is the focal plane; the time axis of the original data has been transformed to a spatial one by multiplication with $c$ to allow a better comparison)

Differences in the above pictures arise from:
- different spacing (so they are not completely at the same place in spacetime)
- using the $c t$ approximation for the longitudinal space axis
- numerical errors ($N_\lambda = 10$ as longitudinal grid spacing, could be done more fine)
- slightly different results when propagating the pulse with the angular spectrum method (as in the preparation routine) compared to a Maxwell solver (here, Yee has been used)

As another test, the beam energy has been calculated. The Original Insight data pulse had an energy of 4.5 J and the energy of the same pulse in PIConGPU was about 2 - 3% less (using the Yee solver), which could be improved by a finer sampling.


BrianMarre:
@fabidie Our CI is unhappy with the code formatting, you ned to run the code formatting tools to fix most and fix the remaining code style issues.

Please install the pre-commit code formater hooks as described [here](https://picongpu.readthedocs.io/en/latest/dev/docs/COMMIT.html) and run ``pre-commit run --all-files`` in your repository afterwards.

include/picongpu/fields/incidentField/Traits.hpp
```diff
@@ -171,6 +183,16 @@ namespace picongpu
                 static constexpr float_X value = 0.0_X;
             };
 
+#if(ENABLE_OPENPMD == 1) && (SIMDIM == DIM3)
+            //! Specialization for InsightPulse profile which has unknown amplitude
+            // kann man die Templates so leer lassen und es fkt trd?
```
steindev: Outdated question?

include/picongpu/fields/incidentField/profiles/InsightPulse.def
```diff
@@ -0,0 +1,122 @@
...
+                         * Note that only C dataOrder is supported.
+                         *
+                         * @warning it is only the dataset itself, a simple text name and not something like
+                         * "/[directories]/[iteration]/meshes/E".
+                         */
+                        static constexpr char const* datasetEName = "E";
+
+                        /** Datatype of the field record
+                         *
+                         * OpenPMD needs this information at compile time.
```
steindev: openPMD is written with lowercase 'o'

include/picongpu/fields/incidentField/profiles/InsightPulse.def
```diff
@@ -0,0 +1,122 @@
...
+                         * (Normalization is required on a user side as internally it is awkward to do with the
+                         * static-constexpr style of using parameters.)
+                         *
+                         * unit: none
+                         */
+                        static constexpr float_64 DIRECTION_X = 0.0;
+                        static constexpr float_64 DIRECTION_Y = 1.0;
+                        static constexpr float_64 DIRECTION_Z = 0.0;
+
+                        /** Unit E polarization direction in PIConGPU
```
steindev: So far, this seems to be the only time where polarisation is written with 'z'

include/picongpu/fields/incidentField/profiles/InsightPulse.def
```diff
@@ -0,0 +1,122 @@
...
+    {
+        namespace incidentField
+        {
+            namespace profiles
+            {
+                namespace defaults
+                {
+                    struct InsightPulseParam
+                    {
+                        /** Path to the openPMD input file
```
steindev: I think you should mention here, how this openPMD file can be obtained from the measurement file.
It might be best to describe this workflow in the manual, e.g. as a new section in `/usage/workflows/addLaser.html` and just refer to the documentation here.

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
```diff
@@ -0,0 +1,618 @@
...
+                            _error_propagationAxisOpenPMD_and_polarisationAxisOpenPMD_have_to_be_different_____check_your_parameters,
+                            (Params::polarisationAxisOpenPMD != Params::propagationAxisOpenPMD));
+                    };
+
+                    template<typename T_Params>
+                    struct InsightPulseFunctorIncidentE;
+
+                    /** Singleton to load field data from OpenPMD to device
+                     *
+                     * The complete dataset will be loaded (equally) to all GPUs, aswell as
```
steindev: -> 'as well as'

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
```diff
@@ -0,0 +1,618 @@
...
+                     * the necessary attributes (extent, cell size, offset to simulation window).
+                     *
+                     * Right now, the data will be loaded at timestep 0, which means that the user
+                     * has to **increase the reserved GPU memory** in memory.param, since otherwise
+                     * the simulation will run into memory issues.
```
steindev: This is an important prerequisite which should also be mentioned in the documentation and the example Param struct.

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
```diff
@@ -0,0 +1,618 @@
...
+                     * Future improvements of this implementation should consider loading the data
+                     * before the start of the simulation (before the particle memory gets allocated),
+                     * since this would solve this bug.
+                     *
+                     * @tparam T_Params user parameters, providing filename etc.
+                     */
```
steindev: Actually, you can also add a todo list here or at the beginning of the file with all the known limitations that should be tackled in refactorings.

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
```diff
@@ -0,0 +1,618 @@
...
+
+                            //! necessary attributes
+                            // Raw = not yet aligned
+                            ::openPMD::Extent const extentRaw = meshRecord.getExtent();
+                            auto const cellSizeRaw = mesh.gridSpacing<dataType>();
+
+                            bufferExtentOpenPMD = std::make_shared<pmacc::HostDeviceBuffer<float_X, 1u>>(3u);
+                            bufferCellSizeOpenPMD = std::make_shared<pmacc::HostDeviceBuffer<float_X, 1u>>(3u);
+                            bufferOffsetOpenPMD = std::make_shared<pmacc::HostDeviceBuffer<float_X, 1u>>(3u);
+
```
steindev: You assign to three different variables the same thing. Why do you have three variables at all?

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
```diff
@@ -0,0 +1,618 @@
...
+                                hostFieldDataBox(openPMDIdx)
+                                    = static_cast<float_X>(fieldData.get()[linearIdx] * meshRecord.unitSI())
+                                    / UNIT_EFIELD;
+                            }
+
+                            /* If the transversal simulation window is smaller than the transversal DataBox extent,
+                             * data at the chunk borders will be discarded. The maximum discarded value (relative to
+                             * the maximum amplitude) will be logged. A possible improvement of this code could include
+                             * storing just the necessary parts of the recorded field data.
+                             */
```
steindev: Do you decide only on the basis of cell numbers? But the cell sizes in the data file and PIConGPU can be different, can't they? Shouldn't you decide on the basis of real space sizes?

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
```diff
@@ -0,0 +1,618 @@
...
+                                                }
+                                            } // k
+                                        } // j
+                                    } // i
+                                    discard = true;
+                                }
+                            } // d
+
+                            // This message is shown for every used device, but 1x would be sufficient (aswell as the
+                            // loop before)
```
steindev: Regarding the message, you can put an if statement with the rank number which evaluates to true only for rank 0.

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
```diff
@@ -0,0 +1,618 @@
...
+
+                        //! loading data to device
+                        static OpenPMDdata& get()
+                        {
+                            static OpenPMDdata dataBuffers{};
+                            return dataBuffers;
+                        }
+
+                    private:
+                        OpenPMDdata()
```
steindev: Is it important to declare this constructor as `HINLINE`? @psychocoderHPC

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
```diff
@@ -0,0 +1,618 @@
...
+
+                            for(uint32_t d = 0u; d < simDim; d++)
+                            {
+                                // return zero if transversal simulation window or simulation timestep exceeds chunk
+                                // extent
+                                if(d != propAxisIdx
+                                       and (posPIC[d] < offsetOpenPMDdataBox(d) or posPIC[d] > offsetOpenPMDdataBox(d) + (extentOpenPMDdataBox(d) - 1.0_X) * cellSizeOpenPMDdataBox(d))
+                                   or (timeOriginPIC - Unitless::TIME_DELAY) < 0.0_X
+                                   or (timeOriginPIC - Unitless::TIME_DELAY) > (extentOpenPMDdataBox(d) - 1.0_X)
+                                           * cellSizeOpenPMDdataBox(d) / SPEED_OF_LIGHT)
```
steindev: I would be grateful if you could indicate with another pair of parenthesis with which block of `or` conditions the `and` condition is evaluated.

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
```diff
@@ -0,0 +1,618 @@
...
+                                    if(d == pmacc::math::abs(polAxisIdxPlusOne) - 1 and polAxisIdxPlusOne < 0
+                                       or d == pmacc::math::abs(transvAxisIdxPlusOne) - 1
+                                           and (transvAxisIdxPlusOne < 0 and rh == true or transvAxisIdxPlusOne > 0 and rh == false))
+                                        // invert direction
+                                        idxClosestRaw[d] = extentOpenPMDdataBox(d) - 1.0_X
+                                            - (posPIC[d] - offsetOpenPMDdataBox(d)) / cellSizeOpenPMDdataBox(d);
+                                    else // keep original direction
+                                        idxClosestRaw[d]
+                                            = (posPIC[d] - offsetOpenPMDdataBox(d)) / cellSizeOpenPMDdataBox(d);
+                                }
```
steindev: same as above.

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
```diff
@@ -0,0 +1,618 @@
...
+                            }
+
+                            // correct longitudinal index
```
steindev: What does correct mean here? In the sense of 'fix' or 'choose the right'

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
```diff
@@ -0,0 +1,618 @@
...
+                    private:
+                        /** Get value of E field for the given position
+                         * Linear interpolation of measured field data, which is aligned centered at the chosen
+                         * incident field plane. If the simulation window extent is greater than the field data chunk,
+                         * zero will be returned. In the other case, the field data chunk will be (transversally)
+                         * cropped.
+                         *
+                         * @param totalCellIdx cell index in the total domain (including all moving window slides)
+                         */
+                        HDINLINE float_X getValueE(floatD_X const& totalCellIdx) const
```
steindev: This routine is very long. Can you split it a little in subroutines, e.g. outsourcing the index finding and interpolation?

steindev (commented):
Good work so far! There are a couple of questions/comments/wishes from my side, of course :grin:

I did not review the python code so far. Will do this next week.

include/picongpu/fields/incidentField/Traits.hpp
fabidie: Oh yes, I checked every file for stupid comments except for Traits haha. I'll remove it

include/picongpu/fields/incidentField/profiles/InsightPulse.def
fabidie: I changed it, hopefully I caught every uppercase O 😅

include/picongpu/fields/incidentField/profiles/InsightPulse.def
fabidie: In BaseParam.hpp it is spelled "polarization" as well. Since I used some methods from there, I also copied the 'z'; but I will change it to 's' for consistency reasons.

include/picongpu/fields/incidentField/profiles/InsightPulse.def
fabidie: Okay, I'll do that👍🏼

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
fabidie: changed ✅

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
fabidie: I'm not so familiar with the concept of shared pointers in C++. If there is another way to solve this, please let me know - the fact is that I need 3 float-vectors of length 3 to describe the field data chunk completely (extent, cell size and the offset w.r.t. the simulation window). I calculate those values at the same timestep when the data is loadad  and since I need them outside the openPMD-singleton, I put them in a HostDeviceBuffer (just like the corresponding field data chunk).

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
fabidie: Yes, thats true, and thats why I converted the openPMD cell size into internal units (so they have the same length scale as `cellSize`)
https://github.com/fabidie/picongpu/blob/9acaccdb80cd952359076fff86189de4af313195/include/picongpu/fields/incidentField/profiles/InsightPulse.hpp#L263C33-L264C110
```C++
dataBoxCellSize(aligningAxisIndex[d])
     = static_cast<float_X>(cellSizeRaw[d] * mesh.gridUnitSI()) / UNIT_LENGTH;
```

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
fabidie: I meant it in the sense of 'fix' since the for-loop in the lines before just tackles the transversal index, but produces nonsense for the longitudinal one. But (as written two commetns above) I changed this line anyway so that I do not have to 'fix' anything at all.

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
fabidie: I have reworked this whole section (better structure and more comments) to make it better understandable. This also solves your next two comments.

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
fabidie: Done ✅

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
fabidie: Done ✅

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
fabidie: Thanks!

include/picongpu/fields/incidentField/profiles/InsightPulse.hpp
fabidie: Done ✅

docs/source/usage/workflows/addLaser.rst
```diff
@@ -46,14 +46,20 @@ The differences between matching incident field- and laser profiles are:
...
 Please refer to :ref:`the detailed description <model-TFSF>` for setting up ``Free`` profile, also for the case when only one of the external fields is known in explicit form.
-For a laser profile with non zero field amplitudes on the transversal borders of the profile e.g. defined by the profile ``Free`` without a transversal envelop the trait ``MakePeriodicTransversalHuygensSurfaceContiguous`` must be specialized and returning true to handle field periodic boundaries correctly.
+For a laser profile with non zero field amplitudes on the transversal borders of the profile e.g. defined by the profile ``Free`` without a transversal envelope the trait ``MakePeriodicTransversalHuygensSurfaceContiguous`` must be specialized and returning true to handle field periodic boundaries correctly.
+
+Another special case is the ``InsightPulse`` profile, which reads an experimental Laser pulse into the simulation, which was obtained from an Insight measurement.
```
steindev: How about: "which reads a laser pulse profile, defined in frequency-space domain, from an openPMD file into the simulation. Specifically this allows to read a pulse profile obtained from an (Insight) measurement"?

Coming to think of it: Maybe we should change the profile's name to reflect its universal applicability. I guess something along the lines of `PulseFromFile` (similar to the density profile) would fit?

docs/source/usage/workflows/addLaser.rst
```diff
@@ -46,14 +46,20 @@ The differences between matching incident field- and laser profiles are:
...
+It is less flexible in choosing propagation and polarisation direction, since those are only allowed along the cell edges, not diagonally.
```
steindev: -> "For now, it is ..."

include/picongpu/fields/incidentField/profiles/InsightPulse.def
```diff
@@ -0,0 +1,131 @@
...
+                    struct InsightPulseParam
+                    {
+                        /** Path to the openPMD input file containing the E-field chunk
+                         *
+                         * This file can be obtained from the Insight measurement data with a python script located
+                         * in lib/python/picongpu/extra/input/. This script provides methods to correct the measured
+                         * data, propagate it to the desired position and transform it into the time domain before
+                         * saving the resulting E-field chunk to openPMD.
+                         *
+                         * This field chunk wil be stored at timestep 0 of the simulation on every used device, which
```
steindev: -> "will"

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
+"""
+This file is part of PIConGPU.
+
+Copyright 2024 Fabia Dietrich
+"""
+
+import openpmd_api as io
```
steindev: I think we wanted to change `io` due to possible ambiguity. Please use `opmd` instead. See #4802

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+    w: waist size
+    """
+    x, y = xy
+    g = A * np.exp(-((x - x0) ** 2 / (w**2) + (y - y0) ** 2 / (w**2)))
+    return g.ravel()
+
+
+def supergauss2D(xy, A, x0, y0, w, n=4):
+    """
+    2D supergaussian distribuition
```
steindev: -> "distribution"

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+# please correct if other units than mm and fs are used!
+c = 2.99792e-4  # speed of light in mm/fs
+
+
+def gauss2D(xy, A, x0, y0, w):
+    """
+    2D gaussian distribution
+
+    Arguments:
+    xy: coordinates
```
steindev: Since you do ` x,y = xy` later, you should specify more specifically what kind of data structure (and its extents) is expected here.

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+    g = A * np.exp(-((x - x0) ** 2 / (w**2) + (y - y0) ** 2 / (w**2)))
+    return g.ravel()
+
+
+def supergauss2D(xy, A, x0, y0, w, n=4):
+    """
+    2D supergaussian distribuition
+
+    Arguments:
+    xy: coordinates
```
steindev: same as above

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+    w: waist size
+    n: superpower (default is 4)
+    """
+    assert n > 2, "n > 2 required"
+    x, y = xy
+    g = A * np.exp(-(((x - x0) ** 2 / (w**2) + (y - y0) ** 2 / (w**2)) ** (n - 2)))
+    return g.ravel()
+
+
+class preproutines:
```
steindev: Class names are typically written camelcase: "PrepRoutines"

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+import numpy as np
+import h5py
+from scipy.interpolate import RegularGridInterpolator
+from scipy.optimize import curve_fit
+
+# please correct if other units than mm and fs are used!
+c = 2.99792e-4  # speed of light in mm/fs
+
+
+def gauss2D(xy, A, x0, y0, w):
```
steindev: please do `A` -> `amp` to comply with code style (lower case for variables and meaningful name)

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+    x, y = xy
+    g = A * np.exp(-(((x - x0) ** 2 / (w**2) + (y - y0) ** 2 / (w**2)) ** (n - 2)))
+    return g.ravel()
+
+
+class preproutines:
+    """
+    class to manipulate and prepare Insight data for PIConGPU usage:
+        - phase corrections
+        - artefact corrections in the near field
```
steindev: -> "artifact"

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+    dx_NF, dy_NF: coordinate spacing
+    waist_NF: waist size in near field
+
+    Et: field in the time domain
+    t, dt: time axis and spacing
+    """
+
+    def __init__(self, path, name, foc):
+        """
+        Load Insight data (= far field), propagate to near field
```
steindev: please write "Insight data (in focus = far field)", so this is clearly defined once.

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+        and extract (gaussian) fit parameters.
+        The far field has to be measured in dependence of "x", "y" and "w".
```
steindev: -> "in dependence of the transverse coordinates "x", "y", and frequency "w"."

lib/python/picongpu/extra/input/preparingInsightData.py
steindev: You should also mention that the field is assumed to be the first (=0th) group in the h5 file

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+            (np.abs(self.Ew[:, :, self.wc_idx]) ** 2).ravel(),
+            p0=(1, 0, 0, 0.05),
+        )
+        self.xc = popt[1]
+        self.yc = popt[2]
+        self.waist = popt[3] * np.sqrt(2)
+        self.zR = self.waist**2 / (2 * c) * self.w[self.wc_idx]
+
+        # propagate to near field (right before the lens, i.e. d=0)
+        preproutines.FFtoNF(self)
```
steindev: should be `self.FFtoNF()`. Well, actually it should be `self.ff_to_nf()` in order to comply with our standard code style. i.e. function names are all lowercase with underscores.

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+            interp_FF = RegularGridInterpolator((a * scale[i], b * scale[i]), FF[:, :, i], method=method)
+            self.Ew[:, :, i] = interp_FF((Y, X))
+
+        self.Ew = self.Ew / np.abs(self.Ew).max()  # rescale amplitude to 1
+
+    def correctPhase(self, GVD=0, TOD=0):
+        """
+        Since there is no information about the global phase of every wavelength measured with Insight,
+        there has to be a phase correction.
+        For that, we assume perfect compression of the beam center in the near field, thus we substract
```
steindev: -> 'subtract'

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+        a global phase for every wavelength.
```
steindev: ", thus we subtract a frequency's phase value at the beam center from every phase value in the frequency's  transverse distribution in the NF." is more precise and correct, isn't it?

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+        Furthermore, there is the possibility to add dispersion parameters (GVD, TOD) to the data.
+
+        Arguments:
+        GVD : Group velocity disperion in fs**2/rad,  optional. The default is 0.
```
steindev: -> "dispersion"

(good to see that you do not copy-paste :wink:)

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+        """
+        # phase in near field beam center
+        phase_centr = np.unwrap(
+            np.angle(self.Ew_NF[np.abs(self.y_NF - self.yc_NF).argmin(), np.abs(self.x_NF - self.xc_NF).argmin(), :])
+        )
+        dispersion = -GVD / 2 * (self.w - self.w[self.wc_idx]) ** 2 - TOD / 6 * (self.w - self.w[self.wc_idx]) ** 3
+
+        self.Ew_NF = np.abs(self.Ew_NF) * np.exp(
+            1j * (np.angle(self.Ew_NF) - phase_centr + dispersion)
+        )  # correcting near field
```
steindev: please put the comment above the line. Same for the next line.

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+        """
+        # index borders of hole
+        x_ugly_idx = np.abs(self.x_NF - x_ugly).argmin()
+        y_ugly_idx = np.abs(self.y_NF - y_ugly).argmin()
+        xidx_low = x_ugly_idx - uglybins
+        xidx_upp = x_ugly_idx + uglybins
+        yidx_low = y_ugly_idx - uglybins
+        yidx_upp = y_ugly_idx + uglybins
+
+        x_surr = np.concatenate((self.x_NF[xidx_low - 3 : xidx_low], self.x_NF[x_ugly_idx + uglybins : xidx_upp + 3]))
```
steindev: This magic `3` is another safety margin? Would you add a comment explaining that number please? Also, the rule of thumb 'Whatever you use twice, automate!' applies here too...

steindev: And why do you apply it only along the x direction? (Comment in code please)

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+            axis=1,
+        )
+        interp = RegularGridInterpolator((self.y_NF[yidx_low:yidx_upp], x_surr, self.w), Ew_surr, method=method)
+        Y_hole, X_hole, W_hole = np.meshgrid(
+            self.y_NF[yidx_low:yidx_upp], self.x_NF[xidx_low:xidx_upp], self.w, indexing="ij"
+        )
+        self.Ew_NF[yidx_low:yidx_upp, xidx_low:xidx_upp, :] = interp((Y_hole, X_hole, W_hole))
+
+        # correct the far field
+        preproutines.NFtoFF(self)
```
steindev: -> `self.nf_to_ff()`

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+            np.abs(self.Ew_NF[:, :, self.wc_idx]).ravel(),
+            p0=(1, 0, 0, 10),
+        )
+        self.xc_NF = popt_NF[1]
+        self.yc_NF = popt_NF[2]
+        self.waist_NF = popt_NF[3]
+        print("new near field center coordinates: xc = %.2f mm, yc = %.2f mm" % (self.xc_NF, self.yc_NF))
+
+        # correct the far field
+        preproutines.NFtoFF(self)
```
steindev: -> `self.nf_to_ff()`

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+
+        print("corrected ugly spot in near field at x = %.2f mm, y = %.2f mm" % (x_ugly, y_ugly))
+
+    def shiftNFtoCenter(self):
+        """
+        If the near field is not centered around (0, 0), it can be done with this function.
+        It automatically also corrects the far field by propagating the centered near field back into the focus.
+        """
+        nx = int(self.xc_NF / self.dx_NF + 0.5)  # number of pixels to shift in x-direction
```
steindev: comment above the line please. Same for the next line.

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+        ny = int(self.yc_NF / self.dy_NF + 0.5)  # number of pixels to shift in y-direction
+
+        if nx == 0 and ny == 0:
+            print("already centered, no corrections necessary")
+            return 0
```
steindev: You should probably check whether `nx` and `ny` are within half the size of the data extent. Otherwise you just shift the pulse from one half of the screen to the other, don't you?

steindev (commented):
Further comment. Reviewed until `propagate()` method in `lib/python/picongpu/extra/input/preparingInsightData.py`.

docs/source/usage/workflows/addLaser.rst
fabidie: I changed it to `FromOpenPMDPulse`

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+        Returns:
+        propagated (complex) field data
+        """
+        assert self.isPhaseCorrected, "Oops, you forgot to correct the phase!"
+
+        # check that propagated beam will still fit into the window
+        w_exp = self.waist * np.sqrt(1 + (z / self.zR) ** 2)
+        assert w_exp < 0.2 * min(
+            self.dx * self.Ew.shape[0], self.dy * self.Ew.shape[1]
+        ), "Oops, you wanted to propagate too far!"
```
steindev: please add "The pulse will not fit in the transverse window."

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+
+        a = np.fft.fftfreq(self.x.size, d=self.dx)
+        b = np.fft.fftfreq(self.y.size, d=self.dy)
+        B, A, W = np.meshgrid(b, a, self.w)
+        cond = (2 * np.pi * c / W) ** 2 * (A**2 + B**2)
```
steindev: Would you please explain in a comment in the code what this mask is for? (or maybe reference to an equation in Godman if appropriate)

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+        idx_out = np.where(cond > 1)
+        F_Ew = np.fft.fft2(self.Ew, axes=(0, 1))
+        F_Ew[idx_in] *= np.exp(1j * W[idx_in] / c * np.sqrt(1 - cond[idx_in]) * z)
+        F_Ew[idx_out] *= np.exp(-W[idx_out] / c * np.sqrt(cond[idx_out] - 1) * z)
+        lin_phase = np.exp(-1j * W * z / c)  # cancel out the time shift
+
+        print("far field propagated to z = %.2f mm" % (z))
+        return np.fft.ifft2(F_Ew, axes=(0, 1)) * lin_phase
+
+    def toTimeDomain(self, Ew, lamb_supp=10):
```
steindev: please rename to `to_time_domain()`

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+
+        # the data has to be sorted according to fft requirements:  w = dw* (0, 1, ... n/2, -n/2, ... -1)
+        # so we have to extent (and interpolate) first the positive side of the spectrum
+        w_fft = np.arange(0, self.w[-1] + zeros * self.dw, self.dw)
+        Y_fft, X_fft, W_fft = np.meshgrid(self.y, self.x, w_fft, indexing="ij")
+        fft_interp = RegularGridInterpolator(
+            (self.y, self.x, self.w), Ew, bounds_error=False, fill_value=0, method="linear"
+        )
+        Ew_fft = fft_interp((Y_fft, X_fft, W_fft))
+        # set the negative frequency part to 0 (instead of the complex conjugate) and neglegt the imaginary part afterwards
```
steindev: -> "neglect"

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+
+        # transform to time domain
+        self.Et = np.fft.fftshift(np.fft.ifft(Ew_fft), axes=-1)
+        self.Et = self.Et / np.abs(self.Et).max()  # rescale amplitude to one
+        self.t = np.fft.fftshift(np.fft.fftfreq(w_fft.size, self.dw / (2 * np.pi)))  # fs
+        self.dt = np.diff(self.t)[0]
+
+        print("field data size: %.1f GB" % (np.prod(self.Et.shape) * 8 * 10**-9))  # assume datatype double
+
+    def saveToOpenPMD(self, outputpath, outputname, energy, pol="x", crop_x=0, crop_y=0, crop_t=0):
```
steindev: please rename to `save_to_openpmd()`

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+        """
+        Save the field data in time domain to an OpenPMD checkpoint. This output will be used for the InsightPulse profile.
```
steindev: -> "openPMD"
-> "FromOpenPMDPulse" profile

steindev (commented):
reviewed until `saveToOpenPMD`

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -0,0 +1,472 @@
...
+        The propagation axis ("z") will be transformed via z = c*t to a spatial one.
```
steindev: -> "The pulse's time evolution at a specific z position will be transformed to a spatial evolution along the propagation direction via z=c*t".

Well, actually this procedure is certainly wrong to do, as we often encounter pulses with length of the order of the Rayleigh length such that the true spatial evolution is affected by defocusing.

In general, it is possible to store several iterations in one openPMD file, which is what we should do here. But this might complicate reading the file and the initialization procedure in PIConGPU. I cannot estimate the effort to perform this refactoring and since you back transform from z to t when reading in PIConGPU, such that the whole workflow is consistent, I guess it is ok to leave it as is for the moment.

However, you should put a big fat Attention Note that this procedure is in principle not correct and needs to be refactored. Please put such a note at the beginning of this file and in the beginning of PIConGPU's FromOpenPMDPulse file.

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
steindev, line 1: Is there a specific reason to leave this as a jupyther notebook and not check it in as a python script? I ask because notebooks require more maintenance than scripts...

And a another point: Please add the copyright notice.

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "## Get the data\n",
+    "The far field data from an Insight measurement is stored in a h5 file and typically measured in dependence of two transversal coordinates (\"x\" and \"y\" in mm) and the frequency (\"w\" in rad/fs). If there are any deviations from this scheme, you will have to adjust the _preproutines_ source code. \n",
+    "\n",
+    "Then, the first steps are:\n",
+    "1. read the far field data and store it in numpy arrays\n",
+    "2. fit the far field data intensity with a 2D gaussian to extract the beam center and waist size\n",
+    "3. propagate to the near field\n",
+    "4. fit the near field data with a 2D supergaussian to extract the beam center and waist size\n",
+    "\n",
+    "For that, we provide path, filename and focal distance (same unit as the transversal scales) to the init funcion."
```
steindev: Change from 'we' to 'you need to' in order to be consistent with the beginning of the file.

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "2c7db135-e858-4152-9ec6-1d0358180f43",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# some nice colorful pictures of our data\n",
```
steindev: our -> your

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
steindev: typo "function"

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "ax1.imshow(\n",
+    "    np.sum(np.abs(insight.Ew), axis=-1),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    extent=(insight.x[0], insight.x[-1], insight.y[0], insight.y[-1]),\n",
+    ")\n",
+    "ax1.set_title(\"spectral integrated amplitude, far field\")\n",
+    "ax1.set_xlabel(\"x [mm]\")\n",
+    "ax1.set_ylabel(\"y [mm]\")\n",
+    "ax1.set_xlim(-0.2, 0.2)\n",
```
steindev: Better delete this manually set limits? They might have a surprising effect.

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "fig = plt.figure(figsize=(14, 8))\n",
+    "\n",
+    "ax1 = fig.add_subplot(231)\n",
+    "ax1.imshow(\n",
+    "    np.sum(np.abs(insight.Ew), axis=-1),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    extent=(insight.x[0], insight.x[-1], insight.y[0], insight.y[-1]),\n",
+    ")\n",
+    "ax1.set_title(\"spectral integrated amplitude, far field\")\n",
```
steindev: -> "spectrally"?

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "ax1.set_ylim(-0.2, 0.2)\n",
+    "\n",
+    "ax2 = fig.add_subplot(232)\n",
+    "ax2.imshow(\n",
+    "    np.sum(np.abs(insight.Ew_NF), axis=-1),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    extent=(insight.x_NF[0], insight.x_NF[-1], insight.y_NF[0], insight.y_NF[-1]),\n",
+    ")\n",
+    "ax2.set_title(\"spectral integrated amplitude, near field\")\n",
```
steindev: -> "spectrally"?

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "ax2.set_xlabel(\"x [mm]\")\n",
+    "ax2.set_ylabel(\"y [mm]\")\n",
+    "ax2.set_xlim(-100, 100)\n",
```
steindev: Delete limits?

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "ax3 = fig.add_subplot(233)\n",
+    "ax3.plot(\n",
+    "    insight.w,\n",
+    "    np.angle(\n",
+    "        insight.Ew_NF[np.abs(insight.y_NF - insight.yc_NF).argmin(), np.abs(insight.x_NF - insight.xc_NF).argmin(), :]\n",
+    "    ),\n",
+    ")\n",
+    "ax3.set_title(\"phase in near field beam center\")\n",
+    "ax3.set_xlabel(r\"$\\omega$ [rad/fs]\")\n",
+    "ax3.set_ylabel(\"rad/fs\")\n",
```
steindev: the phase unit is probably just 'rad', isn't it?

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "    np.sum(np.abs(insight.Ew), axis=0),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    aspect=\"auto\",\n",
+    "    extent=(insight.w[0], insight.w[-1], insight.x[0], insight.x[-1]),\n",
+    ")\n",
+    "ax4.set_title(r\"SD$_x$ in focus\")\n",
+    "ax4.set_ylabel(\"x [mm]\")\n",
+    "ax4.set_xlabel(r\"$\\omega$ [rad/fs]\")\n",
+    "ax4.set_ylim(-0.2, 0.2)\n",
```
steindev: Delete limits?

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "    np.sum(np.abs(insight.Ew), axis=1),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    aspect=\"auto\",\n",
+    "    extent=(insight.w[0], insight.w[-1], insight.x[0], insight.x[-1]),\n",
+    ")\n",
+    "ax5.set_title(r\"SD$_y$ in focus\")\n",
+    "ax5.set_xlabel(r\"$\\omega$ [rad/fs]\")\n",
+    "ax5.set_ylabel(\"y [mm]\")\n",
+    "ax5.set_ylim(-0.2, 0.2)\n",
```
steindev: Delete limits?

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "\n",
+    "ax6 = fig.add_subplot(236)\n",
+    "# sum just over the main beam spot to extract the spectrum\n",
+    "ax6.plot(\n",
+    "    insight.w,\n",
+    "    np.sum(\n",
+    "        np.sum(\n",
+    "            np.abs(\n",
+    "                insight.Ew[\n",
+    "                    np.abs(insight.y - insight.yc + 2 * insight.waist).argmin() : np.abs(\n",
```
steindev, line 139: I cannot follow this calculation, you need to explain it offline :sweat_smile:

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "18bee9ee-bd16-4e28-b9a7-ac5bc12e112e",
+   "metadata": {},
+   "source": [
+    "## Correct the data\n",
+    "### Adjust beam compression and add dispersion parameters\n",
+    "Before we go on with any calculations, we have to correct the phase. Insight reconstructs the amplitude of the far field beam aswell as the phase, up to a unknown global phase for every frequency.\n",
```
steindev: there is "we" again...

steindev: -> "as well as"

steindev: -> "an unknown"

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "11125a32-e618-40b4-aa4b-e1e51def5c3e",
+   "metadata": {},
+   "source": [
+    "### Correct ugly spots in the near field (optional)\n",
+    "Sometimes, the amplitude in the near field looks weird, showing some (unphysical) peaks or holes. These can cause artefacts in the far field and will thus be smoothened out."
```
steindev: -> "artifacts"

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "fig = plt.figure(figsize=(10, 4))\n",
+    "\n",
+    "ax1 = fig.add_subplot(121)\n",
+    "ax1.imshow(\n",
+    "    np.sum(np.abs(insight.Ew), axis=-1),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    extent=(insight.x[0], insight.x[-1], insight.y[0], insight.y[-1]),\n",
+    ")\n",
+    "ax1.set_title(\"spectral integrated amplitude, far field\")\n",
```
steindev: -> "spectrally"

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "ax1.set_ylim(-0.2, 0.2)\n",
+    "\n",
+    "ax2 = fig.add_subplot(122)\n",
+    "ax2.imshow(\n",
+    "    np.sum(np.abs(insight.Ew_NF), axis=-1),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    extent=(insight.x_NF[0], insight.x_NF[-1], insight.y_NF[0], insight.y_NF[-1]),\n",
+    ")\n",
+    "ax2.set_title(\"spectral integrated amplitude, near field\")\n",
```
steindev: -> "spectrally"

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "fig = plt.figure(figsize=(14, 4))\n",
+    "\n",
+    "ax1 = fig.add_subplot(131)\n",
+    "ax1.imshow(\n",
+    "    np.sum(np.abs(insight.Ew), axis=-1),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    extent=(insight.x[0], insight.x[-1], insight.y[0], insight.y[-1]),\n",
+    ")\n",
+    "ax1.set_title(\"spectral integrated amplitude, far field\")\n",
```
steindev: -> "spectrally"

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "ax1.set_ylim(-0.2, 0.2)\n",
+    "\n",
+    "ax2 = fig.add_subplot(132)\n",
+    "ax2.imshow(\n",
+    "    np.sum(np.abs(insight.Ew_NF), axis=-1),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    extent=(insight.x_NF[0], insight.x_NF[-1], insight.y_NF[0], insight.y_NF[-1]),\n",
+    ")\n",
+    "ax2.set_title(\"spectral integrated amplitude, near field\")\n",
```
steindev: -> "spectrally"

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "ax1.imshow(\n",
+    "    np.sum(np.abs(insight.Ew), axis=-1),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    extent=(insight.x[0], insight.x[-1], insight.y[0], insight.y[-1]),\n",
+    ")\n",
+    "ax1.set_title(\"spectral integrated amplitude, far field\")\n",
+    "ax1.set_xlabel(\"x [mm]\")\n",
+    "ax1.set_ylabel(\"y [mm]\")\n",
+    "ax1.set_xlim(-0.2, 0.2)\n",
```
steindev: delete limits?

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "ax2.imshow(\n",
+    "    np.sum(np.abs(insight.Ew_NF), axis=-1),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    extent=(insight.x_NF[0], insight.x_NF[-1], insight.y_NF[0], insight.y_NF[-1]),\n",
+    ")\n",
+    "ax2.set_title(\"spectral integrated amplitude, near field\")\n",
+    "ax2.set_xlabel(\"x [mm]\")\n",
+    "ax2.set_ylabel(\"y [mm]\")\n",
+    "ax2.set_xlim(-100, 100)\n",
```
steindev: delete limits?

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "ax1.imshow(\n",
+    "    np.sum(np.abs(insight.Ew), axis=-1),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    extent=(insight.x[0], insight.x[-1], insight.y[0], insight.y[-1]),\n",
+    ")\n",
+    "ax1.set_title(\"spectral integrated amplitude, far field\")\n",
+    "ax1.set_xlabel(\"x [mm]\")\n",
+    "ax1.set_ylabel(\"y [mm]\")\n",
+    "ax1.set_xlim(-0.2, 0.2)\n",
```
steindev: delete limits?

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "ax2.imshow(\n",
+    "    np.sum(np.abs(insight.Ew_NF), axis=-1),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    extent=(insight.x_NF[0], insight.x_NF[-1], insight.y_NF[0], insight.y_NF[-1]),\n",
+    ")\n",
+    "ax2.set_title(\"spectral integrated amplitude, near field\")\n",
+    "ax2.set_xlabel(\"x [mm]\")\n",
+    "ax2.set_ylabel(\"y [mm]\")\n",
+    "ax2.set_xlim(-100, 100)\n",
```
steindev: delete limits?

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "plt.imshow(\n",
+    "    np.sum(np.abs(Ew_prop), axis=-1),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    extent=(insight.x[0], insight.x[-1], insight.y[0], insight.y[-1]),\n",
+    ")\n",
+    "plt.title(\"spectral integrated amplitude, propagated to %.2f mm\" % (z))\n",
+    "plt.xlabel(\"x [mm]\")\n",
+    "plt.ylabel(\"y [mm]\")\n",
+    "# plt.xlim(-.2, .2)\n",
```
steindev: delete limits?

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "9988b6e5-f3e4-4395-a1ff-4159ede3b4cd",
+   "metadata": {},
+   "source": [
+    "## Propagate\n",
+    "Now the far field data is ready to be propagated. For that, the angular spectrum method is used. \n",
+    "Watch out not to propagate too far, since then the growing beam diameter could reach the transversal window borders and thus cause fourier transform artefacts."
```
steindev: -> "artifacts"

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "ax1 = fig.add_subplot(211)\n",
+    "ax1.imshow(\n",
+    "    np.sum(np.abs(insight.Et), axis=0),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    aspect=\"auto\",\n",
+    "    extent=(insight.t[0], insight.t[-1], insight.x[0], insight.x[-1]),\n",
+    ")\n",
+    "ax1.set_ylabel(\"x [mm]\")\n",
+    "ax1.set_xlim(-100, 100)\n",
```
steindev: delete limits?

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "ax2.imshow(\n",
+    "    np.sum(np.abs(insight.Et), axis=1),\n",
+    "    cmap=\"cubehelix\",\n",
+    "    origin=\"lower\",\n",
+    "    aspect=\"auto\",\n",
+    "    extent=(insight.t[0], insight.t[-1], insight.x[0], insight.x[-1]),\n",
+    ")\n",
+    "ax2.set_xlabel(\"t [fs]\")\n",
+    "ax2.set_ylabel(\"y [mm]\")\n",
+    "ax2.set_xlim(-100, 100)\n",
```
steindev: delete limits?

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "\n",
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8b5e94b1-d37e-4639-ad5c-8990206df22e",
+   "metadata": {},
+   "source": [
+    "## Save data to OpenPMD\n",
```
steindev: -> "openPMD"

lib/python/picongpu/extra/input/preparingInsightData.py
steindev, line 1: Small but valuable feature suggestion: Since you have the frequency domain field you, you can easily provide the field envelope in time domain (i.e. get rid of the field oscillations), which eases pulse analysis tremendously. You only need to calculate `self.Et.real**2+self.Et.imag**2`.

Or you can show this as an additional example plot in the jupyter notebook...

steindev (commented):
Now I am through all files once. I did not look yet at your recent changes after the first part of the review.

Overall, you did a very good job @fabidie!
...only small things I have seen and commented. And a few requests :sweat_smile:

lib/python/picongpu/extra/input/preparingInsightData.py
fabidie: That is true. This transformation step is nonsense from a physical poin of view, it is just done to fulfil the openPMD requirements to store a field chunk (in just one iteration) and undone stright away again by the FromOpenPMDPulse profile. Certainly it would be more correct to store each time slice in their respective iteration, but since there are around 1000 of them per chunk, it will not only complicate the source code but probably also increase the computation time (I guess, didn't try it out yet). 
Nevertheless, I added it to the refactoring list.

lib/python/picongpu/extra/input/preparingInsightData_example.ipynb
fabidie: Because it looks nice 😅 
With the plots and comment sections in between all the necessary preparaion steps I can easily explain what happens and why it has to happen. But I can put it in a python script anyway, if you think that is better.

lib/python/picongpu/extra/input/preparingInsightData.py
fabidie: I added this information to the notebook. 
In the corresponding field plots I already worked with the envelope instead of the (oscillating) field (since `self.Et` is already complex), and now I added another plot, showing both field and envelope to make it more clear. I also calculated an estimation for the pulse duration (measured in the beam center).

psychocoderHPC:
@fabidie I rebased our PR and pushed it again to fix CI issues

psychocoderHPC:
@fabidie please do not forget to rebase the PR to the current development branch

psychocoderHPC:
I have set this PR to draft. IT shows always green but is not ready to be merged. 

Never the less we added many changes to the dev branch therefore a rebase is required.

fabidie:
I closed this one because there happened too much the last 14 days and rebasing is a mess. There is a new one though.

PR #5009, opened 2024-07-30
fix tbg: resolving variables

psychocoderHPC:
If a tbg variable depends on another variable we like to insert vai `!varname` the process of evaluating the right hand side of a variable assignment was not working.

### DIFF
--- a/bin/tbg
+++ b/bin/tbg
@@ -47,7 +47,8 @@ function resolve_vars
                 var_name=${i/\!}
                 # check if variable is defined in the environment
                 if declare -p "$var_name" &>/dev/null ; then
-                    var_value=$(eval echo \$$var_name)
+                    # do not evaluate the variable here because we can still have unresolved tbg variables
+                    var_value=$(printf '%s\n' "${!var_name}")
                     data_set=${data_set/\!$var_name/$var_value}
                 fi
             done
@@ -63,7 +64,11 @@ function resolve_vars
             done
         fi
 
-        export "$data_set"
+        var=$(echo -e "$data_set" | cut -d"=" -f1)
+        var_value=$(echo -e "$data_set" | cut -d"=" -f2-)
+        # the right hand side is fully resolved so we can evaluate the right hand side
+        var_value=$(eval echo "$var_value")
+        export "$var=$var_value"
     done < <(echo "$replace_input" | grep "^[[:alpha:]][[:alnum:]_]*=.*" )
 }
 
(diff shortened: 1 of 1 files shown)


franzpoeschel (approved):
Thank you!
The job did not run yet, but I just tested this on the same template file and it looks like things are substituted correctly again.

bin/tbg
```diff
@@ -186,16 +191,17 @@ function run_cfg_and_set_solved_variables
...
     do
-        var_name=$(echo "$data_set" | cut -d"=" -f1)
+        var_name=$(echo -e "$data_set" | cut -d"=" -f1)
         var_value=$(eval echo \$$var_name)
         if [ -n  "$var_value" ] ; then
-            eval "$data_set"
+            export "$var_name=$var_value"
         fi
-    done < <(echo "$tbg_vars_overwritten" |  grep "^[[:blank:]]*[[:alpha:]][[:alnum:]_]*=.*")
+    done < <(echo "$all_tbg_vars" |  grep "^[[:blank:]]*[[:alpha:]][[:alnum:]_]*=.*")
```
PrometheusPi, line 204: @psychocoderHPC why this change from `$tbg_vars_overwritten` to `$all_tbg_vars`?

PrometheusPi (approved):
Delayed evaluation looks good to me. Thanks @franzpoeschel for testing.

bin/tbg
psychocoderHPC: With the new changes it is required to evaluate and register all variables again and not only the overwritten.

franzpoeschel:
Another issue seems to remain, the string escaping logic has apparently changed

```
TBG_PIC_config="\
{ \
  \"rank_table\": \"posix_hostname\", \
  \"iteration_encoding\": \"variable_based\", \
  \"adios2\": { \
    \"use_group_table\": true, \
    \"attribute_writing_ranks\": 0, \
    \"engine\": { \
      \"parameters\": { \
        \"QueueLimit\": \"1\", \
        \"DataTransport\": \"!TBG_DataTransport\", \
        \"Profile\": \"Off\", \
        \"Threads\": \"!TBG_coresPerGPU\", \
        \"MarshalMethod\": \"BP5\" \
      } \
    } \
  } \
} \
"
```

Lines like these occur in some .cfg files, and while they're not particularly pretty, they worked well so far for defining JSON configs. It seems that the new version of `tbg` now swallows the quotation marks, as seen in the generated `submit.start`:

```bash
      /mnt/bb/$USER/sync_bins/picongpu      \
        --author "fpoeschel <someone@example.com>"                         \
        -d 2 4 2 -g 192 1024 32 -s 400 --periodic 1 1 1 --openPMD.period 100 --openPMD.file simData --openPMD.infix NULL --openPMD.ext sst --openPMD.dataPreparationStrategy mappedMemory --openPMD.json { rank_table: posix_hostname, iteration_encoding: variable_based, adios2: { use_group_table: true, attribute_writing_ranks: 0, engine: { parameters: { QueueLimit: 1, DataTransport: ucx, Profile: Off, Threads: 7, MarshalMethod: BP5 } } } } --versionOnce                  \  
        > ../pic.out 2> ../pic.err          &
```

psychocoderHPC:
@franzpoeschel The support for escaped strings should be fixed with #5011

PR #5011, opened 2024-07-31
tbg: fix support for escaped strings

psychocoderHPC:
Fix support for escaped strings.

example of broken string with escaped quotes.

```
TBG_PIC_config="\
{ \
  \"rank_table\": \"posix_hostname\", \
  \"iteration_encoding\": \"variable_based\", \
} \
"
```

bug introduced with #5009

### DIFF
--- a/bin/tbg
+++ b/bin/tbg
@@ -66,9 +66,15 @@ function resolve_vars
 
         var=$(echo -e "$data_set" | cut -d"=" -f1)
         var_value=$(echo -e "$data_set" | cut -d"=" -f2-)
-        # the right hand side is fully resolved so we can evaluate the right hand side
+
         var_value=$(eval echo "$var_value")
-        export "$var=$var_value"
+        if echo "$data_set" | grep -q -e '^.*$(' -e  '^.*${' ; then
+          # the right hand side contains expressions in form of $( or ${ therefore we evaluate it explicitly
+          export "$var=$var_value"
+        else
+          # the right hand side is fully resolved so we can evaluate the right hand side
+          export "$data_set"
+        fi
     done < <(echo "$replace_input" | grep "^[[:alpha:]][[:alnum:]_]*=.*" )
 }
 


bin/tbg
```diff
@@ -66,9 +66,15 @@ function resolve_vars
 
         var=$(echo -e "$data_set" | cut -d"=" -f1)
         var_value=$(echo -e "$data_set" | cut -d"=" -f2-)
-        # the right hand side is fully resolved so we can evaluate the right hand side
+
         var_value=$(eval echo "$var_value")
-        export "$var=$var_value"
+        if echo "$data_set" | grep -q -e '^.*$(' -e  '^.*${' ; then
```
psychocoderHPC, line 71: This was the only solution I found to fix the issue. I check for expressions and only evaluate the right hand side if needed to avoid that escaped strings get resolved.

franzpoeschel:
Does this mean that e.g. [this config](https://github.com/ComputationalRadiationPhysics/picongpu/blob/fce5b8082228ee2d1509ab9bdf078de7c98f6db4/share/picongpu/examples/KelvinHelmholtz/etc/picongpu/8_pipe.cfg#L67) will no longer be possible?
This has two items in the JSON cfg that are only filled by the template later on.

psychocoderHPC:
> Does this mean that e.g. [this config](https://github.com/ComputationalRadiationPhysics/picongpu/blob/fce5b8082228ee2d1509ab9bdf078de7c98f6db4/share/picongpu/examples/KelvinHelmholtz/etc/picongpu/8_pipe.cfg#L67) will no longer be possible? This has two items in the JSON cfg that are only filled by the template later on.

I tried you cfg with the new fix and it is working. The point I changed are intermediet evaluations. It does NOT mean that you can not use TBG variables there

franzpoeschel:
Nice, I'll give it a test run.

franzpoeschel (approved):
Thank you! This now ran successfully again :)

ikbuibui (approved):

PR #5020, opened 2024-08-07
refactor id provider

psychocoderHPC:
- value_identifier: allows creating identifier which can only be default initiated on the compute device
- refactor IdProvider to avoid multiple definitions of the device global id counter variable
- CMake: enable link-time optimization for device code (required to avoid slowdown due to activated device relocatable code)

The change in `speciesAttributes.param` for the identifier `particleId` is required because we call this method from a pure device function. The compilers was able to resolve issues within a single compile unit but since IDProvider is now split into its own cpp file this change is required.

### DIFF
--- a/include/picongpu/param/speciesAttributes.param
+++ b/include/picongpu/param/speciesAttributes.param
@@ -55,8 +55,8 @@ namespace picongpu
      */
     alias(position);
 
-    //! unique identifier for a particle
-    value_identifier(uint64_t, particleId, IdProvider<simDim>::getNewId());
+    //! unique identifier for a particle, the attribute can only be initialized on the compute device
+    value_identifier_device(uint64_t, particleId, IdProvider<simDim>::getNewId());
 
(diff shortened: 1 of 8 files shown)


psychocoderHPC:
I analysed the clang as cuda compiler issues and it looks like enabling relocatable code is in many situations a problem for the compiler.
Therefore I need to rewrite our id generator handling and remove the usage of an global device variable. It looks like this will be the only way to go to multiple compile units :-(

chillenzer:
So, shall we close this in favour of #5030?

psychocoderHPC:
This PR is closed because clang as cuda compiler has many problems to compile and link device code if we switch PIConGPU to 64bit.
It reports usage of to much shared memory in kernel where we only have 3 integers, ...

https://github.com/ComputationalRadiationPhysics/picongpu/pull/5030 is providing a refactoring to remove the global device variable which is solving all issue and opens the way to use multiple compile units.

PR #5021, opened 2024-08-07
fix missing include and rename type in filename

psychocoderHPC:
- rename `ScreeningInvSqaured.def` to `ScreeningInvSquared.def`
- add includes to def files to ensure the definition of the functors are included before they are implemented

This bug is not critical because the order of includes got managed somewhere in the code but this bug showed up during new development for multi-compile unit support.


chillenzer (approved):
Looks okay to me.

PR #5030, opened 2024-08-15
refactor on device id generation

psychocoderHPC:
- remove global id
- add default id generator
- forward id-generator object to all places needed
- update filter/functor constructor interfaces

# review

The PR is split into different commits to simplify the review. Particle creation and the id generator goes hand in hand, therefore it was not possible to push changes in smaller pre PRs.


include/pmacc/particles/IdProvider.hpp
```diff
@@ -21,164 +21,147 @@
...
+        template<typename T_Worker>
+        HDINLINE uint64_t fetch(T_Worker const& worker)
         {
-            template<typename T_Acc>
-            DINLINE void operator()(const T_Acc&, uint64_cu id) const
-            {
-                nextId = id;
-            }
-        };
+            return alpaka::atomicCas(worker.getAcc(), nextId, uint64_t{0u}, uint64_t{0u});
```
ikbuibui, line 43: It seems to me this does nothing. Is it really being used as an atomic load?

include/picongpu/particles/densityProfiles/IProfile.hpp
```diff
@@ -40,14 +40,41 @@ namespace picongpu
...
+             *                      the constructor
+             * @param currentStep current simulation time step
+             *
+             * @{
              */
-            HINLINE IProfile(uint32_t const currentStep) : T_Profile(currentStep)
+            template<typename DeferFunctor = T_Profile>
+            HINLINE IProfile(
+                uint32_t currentStep,
+                IdGenerator&,
```
ikbuibui: Small comment, not blocking.
IdGenerator is passed sometimes by reference and sometimes by value. It only holds a pointer, so it isnt a big deal, but it can be made more uniform. Also the argument can be marked as `const`.
This happens in many places in this PR

include/pmacc/particles/IdProvider.hpp
psychocoderHPC: yes alpaka and cuda does not provide a atomic load, this is a simple workaround

include/picongpu/particles/densityProfiles/IProfile.hpp
psychocoderHPC: The methods are non const therefore it should take it as non const.

include/picongpu/plugins/misc/ExecuteIfNameIsEqual.hpp
```diff
@@ -41,11 +44,14 @@ namespace picongpu
                  * @param unaryFunctor any unary functor
                  */
                 template<typename T_Kernel, typename... T_Args>
-                void operator()(std::string const& filterName, uint32_t const currentStep, T_Kernel const unaryFunctor)
-                    const
+                void operator()(
+                    std::string const& filterName,
+                    uint32_t const currentStep,
+                    pmacc::IdGenerator const idGen,
```
ikbuibui: It is marked const here.

ikbuibui (dismissed):

ikbuibui (approved):

PR #5032, opened 2024-08-16
refactor `value_identifier`

psychocoderHPC:
- remove `value_identifier_constexpr`
- `value_identifier`'s `getValue()` is now constexpr
- for device functions `value_identifier_func` should be used, `getValue()` is not constexpr and only callable from device

# breaking input change

The definition of the attribute `particleId` has changed to 

```C++
value_identifier_func(uint64_t, particleId, [] ALPAKA_FN_ACC() { return IdProvider<simDim>::getNewId(); });
```
Normally the file `speciesAttributes.param` is not part of the user input and therefore old input setups should work out of the box without changes.

### DIFF
--- a/include/pmacc/identifier/value_identifier.hpp
+++ b/include/pmacc/identifier/value_identifier.hpp
@@ -31,7 +31,6 @@
 /** define a unique identifier with name, type and a default value
  * @param in_type type of the value
  * @param name name of identifier
- * @param in_value user defined value of in_type (can be a constructor of a class)
  *
  * The created identifier has the following options:
  *      getValue()        - return the user defined value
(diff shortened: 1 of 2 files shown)


ikbuibui (approved):

include/pmacc/identifier/value_identifier.hpp
```diff
@@ -47,15 +46,32 @@
  *      `length()` or `length_`
  * @{
  */
```
chillenzer, line 48: This docstring is disconnected from the continuation (?) below.

include/pmacc/identifier/value_identifier.hpp
```diff
@@ -47,15 +46,32 @@
...
+ */
+#define value_identifier_func(in_type, name, ...)                                                                     \
     identifier(                                                                                                       \
-        name, using type = in_type; HDINLINE static type getValue()                                                   \
-        { return in_default; } static std::string getName() { return std::string(#name); })
+        name, using type = in_type; DINLINE static type getValue()                                                    \
+        {                                                                                                             \
+            auto const func = __VA_ARGS__;                                                                            \
+            return func();                                                                                            \
+        } static std::string getName() { return std::string(#name); })
```
chillenzer, line 64: Just because it's a macro, do we have to throw away all formatting rules and not even have linebreaks around function definitions?

chillenzer (approved):
Nothing blocking but two minor things.

include/pmacc/identifier/value_identifier.hpp
psychocoderHPC: We choose whatever the auto formater is doing. IN this case the formating is ugly but we can not change it if we not disable the auto formating for this code.

chillenzer: Wow! That is a severe limitation of `clang-format`, if it can't handle this properly. (Although it's probably not quite trivial to do.) I've found [SkipMacroDefinitionBody](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#skipmacrodefinitionbody) but that's a bit too broad for my taste.

psychocoderHPC: Since we try to avoid large macros this is a rare case that we are affected by the problem.

include/pmacc/identifier/value_identifier.hpp
chillenzer: Offline discussion: This is the doxygen way to apply a preamble to multiple docstrings (see the opening of a grouping `@{`).

psychocoderHPC:
only PICMI test is failing, will be addressed in a seperate PR

PR #5033, opened 2024-08-16
refactor functor/filter interfaces

psychocoderHPC:
- remove the possibility to call each particle functor with an abitrary number of arguments
- optional allow `worker` as first argument for functors
- use perfectly forwarding of template packs
-
User input files should not be affected by the changes.

Note: in the past isit was required that particle functors uses a template parameter pack even if they implement a unary functor. I do not remember why.


include/pmacc/functor/Interface.hpp
```diff
@@ -164,11 +165,12 @@ namespace pmacc
             template<typename T_OffsetType, typename T_Worker, typename... T_Args>
             HDINLINE auto operator()(T_Worker const& worker, T_OffsetType const& domainOffset, T_Args... args) const
```
ikbuibui:
```suggestion
            HDINLINE auto operator()(T_Worker const& worker, T_OffsetType const& domainOffset, T_Args&&... args) const
```

include/picongpu/particles/manipulators/generic/Free.hpp
```diff
@@ -57,12 +57,24 @@ namespace picongpu
...
-                        HDINLINE void operator()(T_Worker const&, T_Args&&... args)
+                        HDINLINE auto operator()(T_Worker const& worker, T_Args&&... args)
+                            -> decltype(std::declval<Functor>()(worker, std::forward<T_Args>(args)...))
+                        {
+                            Functor::operator()(worker, std::forward<T_Args>(args)...);
+                        }
+
+                        template<typename T_Worker, typename... T_Args>
+                        HDINLINE auto operator()(T_Worker const&, T_Args&&... args)
+                            -> decltype(std::declval<Functor>()(std::forward<T_Args>(args)...))
```
ikbuibui, line 72: The return type is being used for overload resolution here. It is now necessary that the return type of the Functor ``()`` operator is void, else we will get compile errors.

ikbuibui (changes requested):

include/picongpu/particles/manipulators/generic/Free.hpp
psychocoderHPC: You are right but this is checked within `pmacc/functor/Interface`, if this is violated we should get a readable error message.

include/pmacc/functor/Interface.hpp
psychocoderHPC: fixed

ikbuibui (approved):

psychocoderHPC:
only PICMI test is failing, will be addressed in a seperate PR

PR #5035, opened 2024-08-16
simplify device compiler `cpp` file selection

psychocoderHPC:
Instead of maintaining a list of files which should compiled by the device compiler we select all files ending with `.x.cpp`.

This is required for the future support of multiple compile units. Maintaining the list by hand would be a nightmare for 20+ cpp files in different folders.


include/picongpu/CMakeLists.txt
```diff
@@ -586,7 +583,7 @@ if(PIC_HAVE_openPMD)
         picongpu-hostonly
         SYSTEM PRIVATE
         $<TARGET_PROPERTY:toml11::toml11,INTERFACE_INCLUDE_DIRECTORIES>)
-    target_link_libraries(picongpu-hostonly PRIVATE toml11::toml11)
+    target_link_libraries(picongpu-hostonly PUBLIC toml11::toml11)
```
ikbuibui: Why is this changed to ``PUBLIC``?

ikbuibui (commented):
Seems like a reasonable idea. x for xcelerator

include/picongpu/CMakeLists.txt
psychocoderHPC: oh this sneaked in from the PR where I cherry-picked it from. I will remove it.

psychocoderHPC:
> Seems like a reasonable idea. x for xcelerator

Yes and it keeps syntax highlighting in all IDE's without any configuration. This is much better than our `.kernel`, `.tpp` we are using very often.

chillenzer (approved):
Sounds okay.

ikbuibui (approved):

psychocoderHPC:
failing CI jobs will be fixed with #5044

PR #5037, opened 2024-08-19
add PluginRegistry

psychocoderHPC:
Add helper construct to register the automatically plugin at the start of the application.
Guard plugin header to fulfill compile dependencies.

This PR is a pre-preparation to move each plugin into an independent compile unit. The plugin `.hpp` file needs to be renamed into `x.cpp`. Additionally the `SimulationStarter` should be separated to avoid that each compile unit is comping an instance of this class, this would only waste compile time.

### DIFF
--- a/include/picongpu/plugins/PluginController.hpp
+++ b/include/picongpu/plugins/PluginController.hpp
@@ -24,177 +24,61 @@
 
 #include "picongpu/simulation_defines.hpp"
 
+#include "picongpu/particles/traits/SpeciesEligibleForSolver.hpp"
 #include "picongpu/plugins/BinEnergyParticles.hpp"
+#include "picongpu/plugins/ChargeConservation.hpp"
+#include "picongpu/plugins/Checkpoint.hpp"
 #include "picongpu/plugins/CountParticles.hpp"
 #include "picongpu/plugins/Emittance.hpp"
 #include "picongpu/plugins/EnergyFields.hpp"
 #include "picongpu/plugins/EnergyParticles.hpp"
+#include "picongpu/plugins/ILightweightPlugin.hpp"
+#include "picongpu/plugins/ISimulationPlugin.hpp"
+#include "picongpu/plugins/IsaacPlugin.hpp"
+#include "picongpu/plugins/PhaseSpace/PhaseSpace.hpp"
+#include "picongpu/plugins/PluginRegistry.hpp"
+#include "picongpu/plugins/PngPlugin.hpp"
 #include "picongpu/plugins/SumCurrents.hpp"
+#include "picongpu/plugins/binning/BinningDispatcher.hpp"
+#include "picongpu/plugins/makroParticleCounter/PerSuperCell.hpp"
 #include "picongpu/plugins/multi/Master.hpp"
-#include "picongpu/plugins/output/images/PngCreator.hpp"
-#include "picongpu/plugins/output/images/Visualisation.hpp"
+#include "picongpu/plugins/openPMD/openPMDWriter.hpp"
+#include "picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp"
+#include "picongpu/plugins/radiation/Radiation.hpp"
+#include "picongpu/plugins/radiation/VectorTypes.hpp"
+#include "picongpu/plugins/shadowgraphy/Shadowgraphy.hpp"
 #include "picongpu/plugins/transitionRadiation/TransitionRadiation.hpp"
 
 #include <pmacc/assert.hpp>
-/* That's an abstract plugin for image output with the possibility
- * to store the image as png file or send it via a sockets to a server.
- *
- * \todo rename PngPlugin to ImagePlugin or similar
- */
-#include "picongpu/plugins/PngPlugin.hpp"
-
-#if(SIMDIM == DIM3 && PIC_ENABLE_FFTW3 == 1 && ENABLE_OPENPMD == 1)
-#    include "picongpu/plugins/shadowgraphy/Shadowgraphy.hpp"
-#endif
-
-#if(ENABLE_OPENPMD == 1)
-#    include "picongpu/plugins/PhaseSpace/PhaseSpace.hpp"
-#    include "picongpu/plugins/binning/BinningDispatcher.hpp"
-#    include "picongpu/plugins/openPMD/openPMDWriter.hpp"
-#    include "picongpu/plugins/particleCalorimeter/ParticleCalorimeter.hpp"
-#endif
-
-#include "picongpu/plugins/ChargeConservation.hpp"
-#if(ENABLE_OPENPMD == 1)
-#    include "picongpu/plugins/makroParticleCounter/PerSuperCell.hpp"
-#endif
-
-#if(ENABLE_ISAAC == 1) && (SIMDIM == DIM3)
-#    include "picongpu/plugins/IsaacPlugin.hpp"
-#endif
-
-#if ENABLE_OPENPMD
-#    include "picongpu/plugins/radiation/Radiation.hpp"
-#    include "picongpu/plugins/radiation/VectorTypes.hpp"
-#endif
-
-#include "picongpu/particles/traits/SpeciesEligibleForSolver.hpp"
-#include "picongpu/plugins/Checkpoint.hpp"
-#include "picongpu/plugins/ILightweightPlugin.hpp"
-#include "picongpu/plugins/ISimulationPlugin.hpp"
-
 #include <pmacc/mappings/kernel/MappingDescription.hpp>
 #include <pmacc/meta/AllCombinations.hpp>
 
-#include <list>
 #include <memory>
+#include <vector>
 
 
 namespace picongpu
 {
-    using namespace pmacc;
-
     /**
      * Plugin management controller for user-level plugins.
      */
     class PluginController : public ILightweightPlugin
     {
     private:
-        std::list<std::shared_ptr<ISimulationPlugin>> plugins;
-
-        template<typename T_Type>
-        struct PushBack
-        {
-            template<typename T>
-            void operator()(T& list)
-            {
-                list.push_back(std::make_shared<T_Type>());
-            }
-        };
-
-        struct TupleSpeciesPlugin
-        {
-            enum Names
-            {
-                species = 0,
-                plugin = 1
-            };
-
-            /** apply the 1st vector component to the 2nd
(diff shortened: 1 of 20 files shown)


include/picongpu/plugins/PluginRegistry.hpp
```diff
@@ -0,0 +1,198 @@
+/* Copyright 2013-2024 Axel Huebl, Benjamin Schneider, Felix Schmitt,
+ *                     Heiko Burau, Rene Widera, Richard Pausch,
+ *                     Benjamin Worpitz, Erik Zenker, Finn-Ole Carstens,
+ *                     Franz Poeschel
```
chillenzer, line 4: You had a lot of contributors for this freshly created file. Are you sure this licence header is correct?

include/picongpu/plugins/PluginRegistry.hpp
```diff
@@ -0,0 +1,198 @@
...
+            meta::ForEach<MakeSeq_t<T_Plugin>, PushBack<boost::mpl::_1>> pushBack;
+            pushBack(PluginRegistry::get().pluginList);
+        }
+
+        static PluginRegistry& get()
+        {
+            static PluginRegistry instance = PluginRegistry{};
+            return instance;
+        }
+
```
chillenzer: Very weak non-blocking but as we are just setting this file up, maybe put this [closer to the top of the class because it's kind of like a constructor](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-order).

include/picongpu/plugins/PluginRegistry.hpp
```diff
@@ -0,0 +1,198 @@
...
+            void operator()(T& list)
+            {
+                list.push_back(std::make_shared<RegistryEntry<T_Type>>());
+            }
+        };
+    };
+
+} // namespace picongpu
+
+#define PIC_JOIN(x, y) x##y
```
chillenzer: Why don't you use `PMACC_JOIN` here?

include/picongpu/plugins/PluginRegistry.hpp
```diff
@@ -0,0 +1,198 @@
...
+
+/** Create a static objects which registers a plugin automatically at the start of the application to the registry.
```
chillenzer:
```suggestion
/** Create a static object which registers a plugin automatically at the start of the application to the registry.
```

include/picongpu/plugins/PluginRegistry.hpp
```diff
@@ -0,0 +1,198 @@
...
+ *
+ * @param counter compile time unique counter __COUNTER__
```
chillenzer, line 155: Just a question: From what I could find out, `__COUNTER__` is not standardised. I don't expect it to break as we are quite exhaustively testing compilers in the CI but just so someone mentioned that: Should we rely on it?

include/picongpu/plugins/PngPlugin.hpp
```diff
@@ -24,6 +24,9 @@
 
 #include "picongpu/particles/traits/SpeciesEligibleForSolver.hpp"
 #include "picongpu/plugins/ILightweightPlugin.hpp"
+#include "picongpu/plugins/PluginRegistry.hpp"
+#include "picongpu/plugins/output/images/PngCreator.hpp"
+#include "picongpu/plugins/output/images/Visualisation.hpp"
```
chillenzer, line 29: Why are these two lines added?

include/picongpu/plugins/radiation/Radiation.hpp
```diff
@@ -20,39 +20,41 @@
...
-#include <optional>
-#include <string>
-#include <vector>
-
-#include <openPMD/openPMD.hpp>
+#if(ENABLE_OPENPMD == 1)
+
+#    if !ENABLE_OPENPMD
+#        error The activated radiation plugin (radiation.param) requires openPMD-api.
+#    endif
```
chillenzer:
```suggestion
```
This looks kind of redundant given the line above.

include/picongpu/plugins/PluginRegistry.hpp
```diff
@@ -0,0 +1,198 @@
...
+         *
+         * @return list of factories where createPlugin() can be called on each entry
+         */
+        auto getPluginFactories() const
+        {
+            return pluginList;
+        }
+
+    private:
+        std::list<std::shared_ptr<IRegistryEntry>> pluginList;
```
chillenzer: Why is this a list and not a vector? And also I think I'd prefer `plugins` over `pluginList` as to [not suggest a type](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-name-type). Both are weak non-blocking.

include/picongpu/plugins/PluginController.hpp
```diff
@@ -82,119 +59,26 @@
...
         {
-            meta::ForEach<AllPlugins, PushBack<boost::mpl::_1>> pushBack;
-            pushBack(plugins);
+            // get all plugins from plugin registry
+            auto const pluginFactories = PluginRegistry::get().getPluginFactories();
+            plugins.reserve(pluginFactories.size());
+            for(auto const& pluginFactory : pluginFactories)
+            {
+                plugins.emplace_back(std::move(pluginFactory->createPlugin()));
+            }
```
chillenzer: Why are you insisting on ~~`.resize`~~ `.reserve` upfront and move semantics within? We're handling a very likely tiny amount of shared pointers (tiny compared to anything where ~~`.resize`~~ `.reserve` would make a difference). A significant simplification would be:
```suggestion
            // get all plugins from plugin registry
            auto const pluginFactories = PluginRegistry::get().getPluginFactories();
            std::transform(pluginFactories.cbegin(), pluginFactories.cend(), plugins.back_insert_iterator(), [](auto const& factory){return factory->createPlugin();});
```

chillenzer (dismissed):
Looks good. I've commented a few minor things but nothing blocking I'd say.

include/picongpu/plugins/PluginRegistry.hpp
psychocoderHPC: yes the copyright is correct. I copied parts of the code from PluginController and these names are from there.

include/picongpu/plugins/PluginRegistry.hpp
psychocoderHPC: done

include/picongpu/plugins/PluginRegistry.hpp
psychocoderHPC: done

include/picongpu/plugins/PluginRegistry.hpp
psychocoderHPC: `__COUNTER__` is used for shared memory too. We need to rely on it becasue I do not know any other possibility.
I have not seen any compiler where the precompiler `__COUNTER__` was not working.

include/picongpu/plugins/PngPlugin.hpp
psychocoderHPC: We need it at the end of the files to register the PngPlugin.  IN the  past there was also the possibilty to send data ofer linux sockets instead of creating PnG's, therefore PngCreator is a template parameter.

include/picongpu/plugins/radiation/Radiation.hpp
psychocoderHPC: fixed, it was in before too but not visible because the guard and check were in two different files.

include/picongpu/plugins/PluginRegistry.hpp
psychocoderHPC: It is a list because iteration performance is not important at this point. I can change the type, I copied most of the code from pluginCOntroller where it was a list too.

psychocoderHPC: renamed `pluginList` to `pluginFactories`

include/picongpu/plugins/PluginController.hpp
psychocoderHPC: `reserve()` is called to avoid resizing the vector with each insertion. 

I will use your transform sugestion

psychocoderHPC: I kept the for loop. Getting the back inserter is a very ugly syntax, over all the code becomes very ugly compared to a for loop.

include/picongpu/plugins/PluginRegistry.hpp
chillenzer: That's what I thought.

include/picongpu/plugins/PngPlugin.hpp
chillenzer: I see. Thanks for the explanation.

include/picongpu/plugins/PluginController.hpp
chillenzer: Okay, I agree that that's not the most beautiful code. But we could do a `.resize()` upfront and then use `plugins.begin()`. (This time I really mean `.resize()` and not `.reserve()`.) My point was that even if we have 100 plugins, the reallocations are not an issue because it's not in a hot loop.

include/picongpu/plugins/PluginController.hpp
```diff
@@ -24,177 +24,61 @@
...
-#include "picongpu/particles/traits/SpeciesEligibleForSolver.hpp"
-#include "picongpu/plugins/Checkpoint.hpp"
-#include "picongpu/plugins/ILightweightPlugin.hpp"
-#include "picongpu/plugins/ISimulationPlugin.hpp"
-
 #include <pmacc/mappings/kernel/MappingDescription.hpp>
 #include <pmacc/meta/AllCombinations.hpp>
 
-#include <list>
+#include <algorithm>
```
chillenzer:
```suggestion
#include <vector>
```
In your current version, `algorithm` isn't used. `vector` is.

chillenzer (dismissed):
Good improvements! I'm fine with the loop version instead of the transform if you don't like my new suggestion either.

include/picongpu/plugins/PluginController.hpp
psychocoderHPC: `resize` would add invalid objects into the container. `reserve()` is allocating enouph space for the coming plugins and always guarantee a clean valid state of the system.

psychocoderHPC:
@chillenzer ready to be merged

chillenzer (approved):

PR #5040, opened 2024-08-19
Fix remove anaconda from ci

chillenzer:


psychocoderHPC (approved):

share/ci/install/pypicongpu.sh
```diff
@@ -25,14 +25,17 @@ export PIC_EXAMPLES=$PICSRC/share/picongpu/examples
...
-wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh
-chmod u+x Miniconda3-latest-Linux-x86_64.sh
-./Miniconda3-latest-Linux-x86_64.sh -b -p /miniconda3
-export PATH=/miniconda3/bin:$PATH
+curl -Ls https://micro.mamba.pm/api/micromamba/linux-64/latest | tar -xvj bin/micromamba
+export MAMBA_ROOT_PREFIX=/tmp/mamba-forge/
+mkdir -p "${MAMBA_ROOT_PREFIX}"
+eval "$(./bin/micromamba shell hook -s posix)"
+export PATH=$(pwd -P)/bin:$PATH
+alias conda="micromamba"
```
SimeonEhrig, line 35: Why do you use an alias?
In the other PR, the `conda` is replaced by `micromamba`: https://github.com/ComputationalRadiationPhysics/picongpu/pull/5041/files
Also, I think the alias is confusing.

chillenzer: It's supposed to be a minimally invasive hot fix. If this passes the CI, we can make a `sed -i` but for now I wanted to have the least possible ways for this to fail.

SimeonEhrig: I did a quick check. `conda` only appears only one times else in the file share/picongpu/examples/LaserWakefield/lib/python/snakemake/requirements.txt as comment.

Therefore I recommend to change it now. Otherwise you have to open immediately the next PR. If will not fix this this week, you will forget it and other people are wondering about the stuff. And if somebody does not understand it, it will simply put code on it, the scripts become complicated and buggy. This happened in the alpaka CI and at the moment, I spent weeks to remove stuff like this.

chillenzer: As I said, I'll do it after the CI ran through. No reason to refactor this if there is a more fundamental problem with the approach.

chillenzer:
I'll re-open a clean version of this.

PR #5041, opened 2024-08-19
remove conda from our documentation and scripts

psychocoderHPC:
Conda is using as default software channel the anaconda reposetory. Software in this channel is not always free to use and can require a license. Therefore we do not point the user to this possible problem.

### DIFF
--- a/docs/source/postprocessing/python.rst
+++ b/docs/source/postprocessing/python.rst
@@ -11,11 +11,11 @@ If you are new to python, get your hands on the tutorials of the following impor
 - https://docs.python.org/3/tutorial/index.html
 
 An easy way to get a Python environment for analysis of PIConGPU data up and running is to download and install
-Miniconda 
+Micromamba
 
-https://docs.conda.io/en/latest/miniconda.html
+https://mamba.readthedocs.io/en/latest/installation/micromamba-installation.html
 
-and set up a conda environment with the help of this conda environment file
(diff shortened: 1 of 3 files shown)


docs/source/postprocessing/python.rst
```diff
@@ -11,11 +11,11 @@ If you are new to python, get your hands on the tutorials of the following impor
...
-Miniconda 
+Micromamba
 
-https://docs.conda.io/en/latest/miniconda.html
+https://mamba.readthedocs.io/en/latest/installation/micromamba-installation.html
 
-and set up a conda environment with the help of this conda environment file
+and set up a Micromamba environment with the help of this Micromamba environment file
 
 https://gist.github.com/steindev/d19263d41b0964bcecdfb1f47e18a86e
```
chillenzer, line 20: This file is just not outdated enough to complain. A few tiny things in that docstring don't quite fit our new narrative but that's okay I'd say.

chillenzer (approved):

docs/source/postprocessing/python.rst
psychocoderHPC: ~~I do not understand what you mean,~~ I change only the names and do not care about the content because I am not familiar with the python ecosystem.

psychocoderHPC: ahh I see, I informed @steindev already and he will update the file.

psychocoderHPC:
@chillenzer can be merged

PR #5042, opened 2024-08-19
CI: reduce runtime by disabling some example combinations

psychocoderHPC:
Disable combinations from `cmakeFlags` for the CI if the combination is only changing a parameter that has no affect on the compiled code. 
Combinations we test in other tests, examples, or benchmarks are disabled too.

# Motivation

Our CI is now running over 5h per commit test if there are no errors in. We reached again a point where it is much to long and it is blocking our development.

# Possible optimization in the future

We could think about running all tests once a week, this would require checking for another variable instead of `CI_PROJECT_DIR` to detect if we would like to run all or only partial tests.
Never the less weekly test could require refactorings in the CI.


psychocoderHPC:
I opened https://github.com/ComputationalRadiationPhysics/picongpu/issues/5043 to keep track of future compile test reductions

share/picongpu/examples/Bunch/cmakeFlags
```diff
@@ -32,15 +32,18 @@
...
-flags[8]="-DPARAM_OVERWRITES:LIST='-DPARAM_RADWINDOW=radWindowFunctionHamming;-DPARAM_RADFORMFACTOR=radFormFactor_TSC_3D'"
-flags[9]="-DPARAM_OVERWRITES:LIST='-DPARAM_RADWINDOW=radWindowFunctionTriplett;-DPARAM_RADFORMFACTOR=radFormFactor_PCS_3D'"
-flags[10]="-DPARAM_OVERWRITES:LIST='-DPARAM_RADWINDOW=radWindowFunctionGauss;-DPARAM_RADFORMFACTOR=radFormFactor_CIC_1Dy'"
+flags[3]="-DPARAM_OVERWRITES:LIST='-DPARAM_SINGLE_PARTICLE=true;-DPARAM_RADFORMFACTOR=radFormFactor_coherent'"
+flags[4]="-DPARAM_OVERWRITES:LIST='-DPARAM_FILTER_GAMMA=1;-DPARAM_RADFORMFACTOR=radFormFactor_incoherent'"
+flags[5]="-DPARAM_OVERWRITES:LIST='-DPARAM_RADWINDOW=radWindowFunctionTriangle;-DPARAM_RADFORMFACTOR=radFormFactor_CIC_3D'"
+flags[6]="-DPARAM_OVERWRITES:LIST='-DPARAM_RADWINDOW=radWindowFunctionHamming;-DPARAM_RADFORMFACTOR=radFormFactor_TSC_3D'"
+flags[7]="-DPARAM_OVERWRITES:LIST='-DPARAM_RADWINDOW=radWindowFunctionTriplett;-DPARAM_RADFORMFACTOR=radFormFactor_PCS_3D'"
+flags[8]="-DPARAM_OVERWRITES:LIST='-DPARAM_RADWINDOW=radWindowFunctionGauss;-DPARAM_RADFORMFACTOR=radFormFactor_CIC_1Dy'"
```
chillenzer, line 40: Just for my information: Why do we keep so many of these?

share/picongpu/tests/CollisionsThermalisation/cmakeFlags
```diff
@@ -29,22 +29,26 @@
...
+flags[1]="-DPARAM_OVERWRITES:LIST='-DPARAM_START_IONS=Corner5000ppc;-DPARAM_START_ELECTRONS=Corner5000ppc;-DPARAM_DYNAMIC_LOG_INTER=1;-DPARAM_DYNAMIC_LOG_INTRA=1'"
+flags[2]="-DPARAM_OVERWRITES:LIST='-DPARAM_START_IONS=Corner5000ppc;-DPARAM_START_ELECTRONS=Corner5000ppc;-DPARAM_DYNAMIC_LOG_INTER=1;-DPARAM_DYNAMIC_LOG_INTRA=0'"
 # test support for 32bit precision
-flags[9]="-DPARAM_OVERWRITES:LIST='-DPARAM_START_IONS=Corner5000ppc;-DPARAM_START_ELECTRONS=Corner1000ppc;-DPARAM_COLLISION_PRECISION=float_32;-DPARAM_PRECISION=precision32Bit'"
+flags[3]="-DPARAM_OVERWRITES:LIST='-DPARAM_START_IONS=Corner5000ppc;-DPARAM_START_ELECTRONS=Corner1000ppc;-DPARAM_COLLISION_PRECISION=float_32;-DPARAM_PRECISION=precision32Bit'"
 # test support for mixed precision
-flags[10]="-DPARAM_OVERWRITES:LIST='-DPARAM_START_IONS=Corner5000ppc;-DPARAM_START_ELECTRONS=Corner1000ppc;-DPARAM_PRECISION=precision32Bit'"
+flags[4]="-DPARAM_OVERWRITES:LIST='-DPARAM_START_IONS=Corner5000ppc;-DPARAM_START_ELECTRONS=Corner1000ppc;-DPARAM_PRECISION=precision32Bit'"
```
chillenzer, line 38: At first I thought the comments are mixed up but I'm not so sure. Maybe for clarity one could make the `PARAM_COLLISION_PRECISION` explicit in both cases.

chillenzer (approved):
Looks good although I'm not really qualified to approve this. Someone better versed in the code paths related to these flags should check that no important combination is missing.

share/picongpu/examples/Bunch/cmakeFlags
psychocoderHPC: They test/compile different window implementations and we like to avoid that one of it breaks. This is an easy target for unit tests.

share/picongpu/tests/CollisionsThermalisation/cmakeFlags
psychocoderHPC: Good point but I do not like to re-run the CI because of that again. 
We have this in other examples too. We have no rule or best practice for it. In general how we do al lof it is not good an should be changes with #5043

psychocoderHPC:
failing CI jobs will be fixed with #5044

psychocoderHPC:
The compile time is with this PR only reduced to 430 minutes :-(

PR #5046, opened 2024-08-23
update JURECA profile and tpl

ikbuibui:
SBATCH --distribution=*:block is no longer recommended
MPI-settings/CUDA needs to be loaded for CUDA aware MPI to work properly

### DIFF
--- a/etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
+++ b/etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
@@ -38,6 +38,7 @@ module load GCC/12.3.0
 module load CUDA/12
 module load CMake
 module load OpenMPI/4.1.5
+module load MPI-settings/CUDA
 module load Python/3.11.3
 module load ADIOS2/2.9.2
 module load HDF5/1.14.2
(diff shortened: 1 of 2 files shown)


etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
```diff
@@ -47,6 +48,8 @@ module load Boost/1.82.0
 module load SciPy-bundle/2023.07
 module load matplotlib/3.7.2
 module load h5py/3.9.0
+module load numba
```
chillenzer, line 51: What are we loading `numba` for?

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
```diff
@@ -47,6 +48,8 @@ module load Boost/1.82.0
...
+module load tqdm
```
chillenzer, line 52: And this?

chillenzer (approved):

etc/picongpu/jureca-jsc/gpus_picongpu.profile.example
ikbuibui: These are dependencies of @BrianMarre's atomic physics validation ~~scripts~~ EDIT: library. They are not actively used in the project currently, but were used in the past, and may be used again in the future.

BrianMarre: it's a library, but otherwise correct.

PR #5049, opened 2024-08-23
reduce cyclic include dependencies by moving code into own compile units

psychocoderHPC:
follow up of #5045 

Move all plugins, some stages and Fields (e.g. FieldTmp, FieldJ) into own compile units by using cpp's to compile non template dependent parts.

# Review:

Please visit the changes commit by commit. I have not changed the logic but sometimes small refactorings was required.


include/picongpu/fields/FieldJ.x.cpp
```diff
@@ -18,13 +18,13 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
-#pragma once
+
+#include "picongpu/fields/FieldJ.hpp"
```
ikbuibui, line 22: Included twice

psychocoderHPC: fixed

ikbuibui (approved):

PR #5051, opened 2024-08-26
raditation unit tests

psychocoderHPC:
- unit test: radiaton window and form factor tests: Add test to compare host and device resuts for both functor types.
- Reduce variation of radiation window and formfactor cmakeFlags to reduce the compile time. Both type of functors have there own unit tests.

### DIFF
--- a/include/picongpu/plugins/radiation/VectorTypes.hpp
+++ b/include/picongpu/plugins/radiation/VectorTypes.hpp
@@ -20,8 +20,7 @@
 #pragma once
 
 
-#include "picongpu/simulation_defines.hpp"
-
+#include "picongpu/defines.hpp"
 #include "picongpu/plugins/radiation/vector.hpp"
 
 
--- a/share/picongpu/unit/radiationFormFactor.cpp
+++ b/share/picongpu/unit/radiationFormFactor.cpp
@@ -1,0 +1,184 @@
+/* Copyright 2024 Rene Widera
+ *
+ * This file is part of PIConGPU.
+ *
+ * PIConGPU is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PIConGPU is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <pmacc/boost_workaround.hpp>
+
+#include <pmacc/test/PMaccFixture.hpp>
+
+// STL
+#include <pmacc/Environment.hpp>
+#include <pmacc/dimensions/DataSpace.hpp>
+#include <pmacc/lockstep.hpp>
+#include <pmacc/math/ConstVector.hpp>
+#include <pmacc/memory/buffers/DeviceBuffer.hpp>
+#include <pmacc/memory/buffers/HostBuffer.hpp>
+#include <pmacc/meta/conversion/MakeSeq.hpp>
+
+#include <cstdint>
+#include <iostream>
+#include <string>
+
+#include <catch2/catch_approx.hpp>
+#include <catch2/catch_test_macros.hpp>
+#include <picongpu/param/precision.param>
+
+/* Values must be defined because they are required by the radiation functions.
+ * Typically these values coming from simulation_defines but we try to avoid including these in unit tests.
+ */
+namespace picongpu
+{
+    constexpr picongpu::float_X SPEED_OF_LIGHT = 1.0;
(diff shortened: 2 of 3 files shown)


include/picongpu/plugins/radiation/VectorTypes.hpp
PrometheusPi, line 1: @psychocoderHPC What changed with this file and why did it still compile?

share/picongpu/unit/radiationWindowFunction.cpp
```diff
@@ -0,0 +1,163 @@
...
+                deviceLxBuffer.getDataBox(),
+                resultDevice.getDataBox());
+
+        resultHost.copyFrom(resultDevice);
+
+        auto res = resultHost.getDataBox();
+        for(uint32_t i = 0u; i < numValues; ++i)
+        {
+            auto pos = inCellPositionBuffer.getDataBox()[i];
+            auto lx = inLxBuffer.getDataBox()[i];
```
PrometheusPi, line 125: This will mostly create garbage since it will lie ~~outside~~ near one edge of a real simulation box size. 
The code change above should fix that and thus cover the full window range.

share/picongpu/unit/radiationWindowFunction.cpp
```diff
@@ -0,0 +1,163 @@
...
+            forEach(
+                [&](uint32_t const idx)
+                {
+                    auto valueIdx = blockIdx * elemPerBlock + idx;
+
+                    if(valueIdx < numValues)
+                    {
+                        auto windowFunc = T_WindowFunction{};
+
+                        result[valueIdx] = windowFunc(positions[valueIdx], lx[valueIdx]);
```
PrometheusPi:
```suggestion
                        result[valueIdx] = windowFunc(positions[valueIdx] * lx[valueIdx], lx[valueIdx]);
```

share/picongpu/unit/radiationWindowFunction.cpp
```diff
@@ -0,0 +1,163 @@
...
+ * Compares the on host and on device computed result.
+ *
+ * @tparam T_WindowFunction radiation window function
+ */
+template<typename T_WindowFunction = boost::mpl::_1>
+struct TestWindowFunction
+{
+    /** Validates the shape
+     *
+     * @param inCellPositionBuffer Buffer with positions, each value must be in rage [0.0;1.0).
```
PrometheusPi, line 74: should be global position in [0, Lx) - using [0, 1.0) * Lx is fine - see code suddgestion below.

share/picongpu/unit/radiationFormFactor.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+            forEach(
+                [&](uint32_t const idx)
+                {
+                    auto valueIdx = blockIdx * elemPerBlock + idx;
+
+                    if(valueIdx < numValues)
+                    {
+                        float_X const omg = omega[valueIdx];
+                        auto const windowFunc = T_FormFactorFunctions{omg, {1.0, 1.0, 1.0}};
```
PrometheusPi:
```suggestion
                        auto const formFactorFunc = T_FormFactorFunctions{omg, {1.0, 1.0, 1.0}};
```

PrometheusPi: Furthermore `{1.0, 1.0, 1.0}` is not a unit vector (of length 1) and thus will create nonsense. 
How about 
```c++
std::uniform_real_distribution<> phi_dist(0.0, 2.0*PI);
std::uniform_real_distribution<> theta_dist(0.0, 1.0*PI);
...
... {math::cos(phi) * math::sin(theta), math::sin/phi) * math::sin(theta), math::cos(theta)} ...
```

PrometheusPi (changes requested):
some minor issues regrading the completeness of testing.

include/picongpu/plugins/radiation/VectorTypes.hpp
psychocoderHPC:
```C++
```
to

```
#include "picongpu/defines.hpp"
```

It avoids including all the garbage of picongpu

share/picongpu/unit/radiationWindowFunction.cpp
psychocoderHPC: for me garbage is ok as long as it is equally wrong on CPU and GPU

share/picongpu/unit/radiationWindowFunction.cpp
psychocoderHPC: added

share/picongpu/unit/radiationFormFactor.cpp
psychocoderHPC: added

psychocoderHPC:
@PrometheusPi I pushed the changes to the last commit

share/picongpu/unit/radiationWindowFunction.cpp
psychocoderHPC: I used your suggestion

share/picongpu/unit/radiationWindowFunction.cpp
PrometheusPi: Okay - fine with me - might be that we just test creation of zeros and never hit the box (non zero values).

PrometheusPi (approved):

PR #5055, opened 2024-08-27
Fix polarization component in shadowgraphy plugin

finnolec:
Previously the B fields had 2 x components instead of an x and a y component...


BrianMarre:
@finnolec you're PR failed in the CI since you did not rebase it to the current dev and therefore still lack the fix to the PICMI CI jobs removing anaconda

finnolec:
@steindev please review :)

steindev (approved):

PR #5059, opened 2024-08-28
refactor `pmacc::math::Vector`

psychocoderHPC:
- refactor `math::Vector`
  - add `constexpr` to nearly all operations, remove HDINLINE if method is `constexpr`
  - default storage policy will be `std::array`
  - remove navigator policy
  - remove `TwistComponents.hpp` (is not used somewhere)
- `compile-time/Vector`
  - refactor runtime vector creation
-  add basic unit tests for vector

The navigator in math::Vector is removed because we do not use this policy anymore and if needed you can wrap the storage policy to have the same features. Simple code was more important for this refactoring.
The added tests have not the aim to be complete but will increase the coverage compared to the current status.


include/pmacc/math/vector/Vector.hpp
```diff
@@ -23,99 +23,47 @@
...
-                /*align full vector*/
-                PMACC_ALIGN(v[dim], type) = {};
-
-                HDINLINE
-                constexpr type& operator[](const uint32_t idx)
-                {
-                    return v[idx];
-                }
+            using BaseType = std::array<T_Type, T_dim>;
+            // using BaseType::array;
```
ikbuibui:
``` suggestion
```
Dead code

include/pmacc/math/vector/Vector.hpp
```diff
@@ -23,99 +23,47 @@
...
-                const type& operator[](const uint32_t idx) const
-                {
-                    return v[idx];
-                }
-            };
+            constexpr ArrayStorage() = default;
+            constexpr ArrayStorage(ArrayStorage const&) = default;
+            constexpr ArrayStorage(ArrayStorage&&) = default;
+            constexpr ArrayStorage& operator=(ArrayStorage const&) = default;
+            constexpr ArrayStorage& operator=(ArrayStorage&&) = default;
```
ikbuibui:
``` suggestion
```
These are all defaulted. This is correct, but they can simply be removed and the compiler generates the same stuff.

include/pmacc/math/vector/Vector.hpp
```diff
@@ -151,41 +99,44 @@ namespace pmacc
              * @param args value of each dimension, x,y,z,...
              */
             template<typename... T_Args, typename = std::enable_if_t<(std::is_convertible_v<T_Args, T_Type> && ...)>>
-            HDINLINE constexpr Vector(T_Args... args) : Storage(Navigator{}, args...)
+            constexpr Vector(T_Args... args)
+                : Vector([values = std::array<T_Type, sizeof...(args)>{static_cast<T_Type>(args)...}](
+                    uint32_t const i) constexpr { return values[i]; })
```
ikbuibui:
```suggestion
: Storage{static_cast<T_Type>(args)...}
```

include/pmacc/math/vector/Vector.hpp
```diff
@@ -201,138 +152,115 @@ namespace pmacc
...
-            {
-                return (*this)[0];
-            }
-
-            HDINLINE type& y()
-            {
-                PMACC_CASSERT_MSG(math_Vector__access_to_y_is_not_allowed_for_DIM_lesser_than_2, dim >= 2u);
-                return (*this)[1];
+                    result[i] = -(*this)[i];
+                return result;
```
ikbuibui: Can be written as `return Vector([=](uint32_t const i) constexpr { return -(*this)[i]; })`

ikbuibui (changes requested):

ikbuibui (approved):

PR #5060, opened 2024-08-29
fix possible deadlock

psychocoderHPC:
PMacc's event system must be empty before global blocking methods e.g. MPI collectives are called. 
OpenPMD's flush is possibly executing blocking collectives.


include/picongpu/plugins/openPMD/openPMDWriter.x.cpp
```diff
@@ -1658,6 +1660,7 @@ make sure that environment variable OPENPMD_BP_BACKEND is not set to ADIOS1.
 
                     if(numDataPoints == 0)
                     {
+                        // avoid deadlock between not finished pmacc tasks and mpi blocking collectives
```
steindev, line 1663: @psychocoderHPC here you just added the comment, but not the code `eventSystem::getTransactionEvent().waitForFinished();`. Is this correct?

psychocoderHPC: no this is a bug :-(

psychocoderHPC (commented):
fixed

include/picongpu/plugins/openPMD/openPMDWriter.x.cpp
psychocoderHPC: fixed

ikbuibui (approved):

PR #5066, opened 2024-09-02
refactor unit/si and pic unit system and param files

psychocoderHPC:
Introduce a class `Sim` currently located in `Simulation.unitless` which provides access to base PIConGPU si values, picongpu normalized values, and units.
This refactoring brings structure into the usage of units and input values and is the base to reduce cyclic dependencies.
In the far future we could hopefully give a instance of this class into each kernel and switch easily from compile time units to runtime units.

# Review

Please review each commit separate.

Helper tool used 

```bash
cd <picongpu code>
../rename.sh CELL_DEPTH CELL_DEPT "getCellSize().z()"
```

`rename.sh`
```
#!/bin/bash

# $1 short name CELL_WIDTH
# $2 reduced name CELL_WIDT
# $3 c++ code getCellSize().x()
cd picongpu

for i in include docs share test ; do

    echo run: $i
    find $i | xargs -n1 -P8 -I{} sed -i "s/$1_SI/$2_SI/g" {}
    find $i | xargs -n1 -P8 -I{} sed -i "s/$1/sim.pic.$3/g" {}
    find $i | xargs -n1 -P8 -I{} sed -i "s/SI::$2_SI/sim.si.$3/g" {}
    find $i | xargs -n1 -P8 -I{} sed -i "s/$2_SI/$1_SI/g" {}
    #find $i | xargs -n1 -P8 -I{} sed -i "s/$1_SI/sim.si.$3/g" {}
    git checkout HEAD CHANGELOG.md
done
```


share/picongpu/tests/CollisionsBeamRelaxation/cmakeFlags
```diff
@@ -29,15 +29,15 @@
 #   - start with zero index
 #   - increase by 1, no gaps
 
-flags[0]="-DPARAM_OVERWRITES:LIST='-DPARAM_DENSITY_RATIO_IONS=10.0;-DPARAM_DELTA_T=0.2;-DPARAM_CHARGE_IONS=IonCharge1;-DPARAM_DRIFT_ELECTRONS=AssignFastDrift'"
-flags[1]="-DPARAM_OVERWRITES:LIST='-DPARAM_DENSITY_RATIO_IONS=10.0;-DPARAM_DELTA_T=0.001;-DPARAM_CHARGE_IONS=IonCharge1;-DPARAM_DRIFT_ELECTRONS=AssignSlowDrift;-DPARAM_DYNAMIC_LOG=1'"
+flags[0]="-DPARAM_OVERWRITES:LIST='-DPARAM_DENSITY_RATIO_IONS=10.0;-DPARAM_sim.pic.getDt()=0.2;-DPARAM_CHARGE_IONS=IonCharge1;-DPARAM_DRIFT_ELECTRONS=AssignFastDrift'"
```
psychocoderHPC: these changes should be reverted

include/picongpu/particles/atomicPhysics/kernel/FillLocalRateCache_BoundBound.kernel
```diff
@@ -20,7 +20,7 @@
 #pragma once
 
 #include "picongpu/simulation_defines.hpp"
-// need grid.unitless for CELL_VOLUME
+// need grid.unitless for sim.pic.getCellSize().productOfComponents()
```
ikbuibui:
``` suggestion
// need simulation.unitless for sim.pic.getCellSize().productOfComponents()
```

include/picongpu/particles/collision/detail/cellDensity.hpp
```diff
@@ -50,7 +50,7 @@ namespace picongpu::particles::collision::detail
                     auto particle = parAccess[partIdx];
                     density += particle[weighting_];
                 }
-                densityArray[linearIdx] = density / CELL_VOLUME;
```
ikbuibui, line 53: I would suggest we define `sim.pic.getCellVolume()` etc

ikbuibui (commented):
There are some places still in the code comments that mention unit.param

psychocoderHPC:
> There are some places still in the code comments that mention unit.param

We need to do cleanups after all refactorings. The problem is that we can not keep the PR to long open because all other changes in the can break the PR what can become had to resolve where broken comments are no big deal.

Never the less good that you mention it, that we can keep track of it.

include/picongpu/particles/collision/detail/cellDensity.hpp
psychocoderHPC: I would like to avoid it because we can not provide in this object interfaces for all typical cases. If wee need it shorter we should think about renamings in vector or provide a function `volume(sim.pic.getCellSize)` within picongpu.

ikbuibui (approved):
This PR also has a fix which #5075 is also trying to solve. I will merge this to avoid rebasing and merge conflicts

include/picongpu/unitless/simulation.unitless
```diff
@@ -26,24 +26,189 @@
...
+            {
+                return -1.0 * si.getBaseCharge() * typicalNumParticlesPerMacroParticle();
+            }
+            constexpr float_64 energy() const
+            {
+                return mass() * length() * length() / (time() * time());
+            }
+            constexpr float_64 eField() const
+            {
+                return 1.0 / (time() * time() / mass() / length() * length());
```
psychocoderHPC, line 177: bug it must be `/ length() * charge());`

PR #2699, opened 2018-09-04
Release 0.4.0-rc3

ax3l:
Collection of all relevant PRs for the `release-0.4.0` branch in the next days/weeks.
Update to `0.4.0-rc3`.

## Bug Fixes

- [x] fix plotting tool for numerical heating #2697 
- [x] fix gameOfLife #2700 
- [x] Boost Template Aliases: Fix Workaround #2706
- [x] fix tbg error message #2712
- [x] fix compile with boost 1.68.0 #2713 
- [ ] [Alpaka 0.3.4](https://github.com/ComputationalRadiationPhysics/alpaka/projects/4)

## Misc

- [x] bump version to `0.4.0-rc3`
- [ ] update `CHANGELOG.md` with [rc2](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2686) and [rc3](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2699) changes
- [x] Boost: Deprecated Math Header #2707
- [x] accelerator `threads` #2701
- [ ] Taurus: SCS5 Update #2667
- [ ] Titan: Update profile to CUDA 9.1

PR #5068, opened 2024-09-03
PMacc: fix `constexpr math::Vector`

psychocoderHPC:
- fix CUDA 11.6 compile time evaluation issues by
  - unsing `{}` instead of `()` for type generation
  - constexpr lambda capture by reference instead of copy
- add `initializer_list` support and add coresponding test
- simplify the vector deduction guide

### DIFF
--- a/include/pmacc/math/vector/Vector.hpp
+++ b/include/pmacc/math/vector/Vector.hpp
@@ -97,6 +97,16 @@ namespace pmacc
              * @attention This constructor allows implicit casts.
              *
              * @param args value of each dimension, x,y,z,...
+             *
+             * A constexpr vector should be initialized with {} instead of () because at least
+             * CUDA 11.6 has problems in cases where a compile time evaluation is required.
+             * @code{.cpp}
+             *   constexpr auto vec1 = Vector{ 1 };
+             *   constexpr auto vec2 = Vector{ 1, 2 };
+             *   //or explicit
+             *   constexpr auto vec3 = Vector<int, 3u>{ 1, 2, 3 };
+             *   constexpr auto vec4 = Vector<int, 3u>{ {1, 2, 3} };
+             * @endcode
              */
             template<typename... T_Args, typename = std::enable_if_t<(std::is_convertible_v<T_Args, T_Type> && ...)>>
             constexpr Vector(T_Args... args) : Storage(static_cast<T_Type>(args)...)
@@ -109,7 +119,7 @@ namespace pmacc
              */
             template<typename T_OtherStorage>
             constexpr Vector(const Vector<T_Type, T_dim, T_OtherStorage>& other)
-                : Vector([=](uint32_t const i) constexpr { return other[i]; })
+                : Vector([&](uint32_t const i) constexpr { return other[i]; })
             {
             }
 
@@ -118,7 +128,7 @@ namespace pmacc
                 typename T_OtherStorage,
                 typename = std::enable_if_t<std::is_convertible_v<T_OtherType, T_Type>>>
             constexpr explicit Vector(const Vector<T_OtherType, dim, T_OtherStorage>& other)
-                : Vector([=](uint32_t const i) constexpr { return static_cast<T_Type>(other[i]); })
+                : Vector([&](uint32_t const i) constexpr { return static_cast<T_Type>(other[i]); })
             {
             }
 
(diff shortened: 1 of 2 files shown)


include/pmacc/math/vector/Vector.hpp
```diff
@@ -473,14 +492,9 @@ namespace pmacc
...
-        ALPAKA_FN_HOST_ACC Vector(T_Args... args)
-            ->Vector<
-                typename std::tuple_element<0, std::tuple<T_Args...>>::type,
-                uint32_t(sizeof...(T_Args)),
-                ArrayStorage<
-                    typename std::tuple_element<0, std::tuple<T_Args...>>::type,
-                    uint32_t(sizeof...(T_Args))>>;
+        template<typename T_1, typename... T_Args>
+        ALPAKA_FN_HOST_ACC Vector(T_1, T_Args...)
+            ->Vector<T_1, uint32_t(sizeof...(T_Args) + 1u), ArrayStorage<T_1, uint32_t(sizeof...(T_Args) + 1u)>>;
```
ikbuibui, line 488: Much nicer

include/pmacc/math/vector/Vector.hpp
```diff
@@ -97,19 +98,37 @@ namespace pmacc
...
              */
             template<typename... T_Args, typename = std::enable_if_t<(std::is_convertible_v<T_Args, T_Type> && ...)>>
             constexpr Vector(T_Args... args) : Storage(static_cast<T_Type>(args)...)
             {
             }
 
+            constexpr Vector(std::initializer_list<T_Type> initList)
+                : Vector([&](uint32_t const i) constexpr { return *(initList.begin() + i); })
+            {
+            }
```
ikbuibui: This would cause an out of memory access if the initList is smaller than the dim.
In general everywhere we assume that the generator can generate dim values, but it isnt a problem anywhere else.
One solution can be to ensure the initializer list has dim elements (`initList.size()` is constexpr).
In this case should we also allow construction from `initializer_list<T_Args>` if `std::is_convertible_v<T_Args, T_Type>`?

Another solution can be to simply remove this constructor. This would "just work" for the default `ArrayStorage` type. 
The potential problem with this is that the initalizer list construction would break if  `Vector` or `Storage` have a data member aside from the actual storage (idk if they might need one at some point because `Storage` wants to do some extra magic or something else), or if the underlying storage doesnt support aggregate initialization. (Both problems dont happen right now, but if they happen they will need workarounds) 
```suggestion
```
As an effect of removing the constructor we would allow construction of `Vector` even if the `initList` is
- smaller than `dim`, with this approach the rest of the vector will be default value initilaized for `T_Type`. 
- of a different type which is convertible to `T_Type`

ikbuibui (changes requested):

include/pmacc/math/vector/Vector.hpp
psychocoderHPC: Removing the constructor will result into the fact that you ca not use `Vectro<int,3>{{1,2,3}}` anymore.

To short list is `constexpr auto vec3 = Vector<int, 3>{{3, 2}};` will be `note #2699-D: attempt to access storage one position past the end of an array of 2 elements`.

To long list is currently supported because we never access invalid elements. If we would disallow it we could add a static_assert but I would like to move this to another PR. I am blocked agin by our CI :-(

> Another solution can be to simply remove this constructor ... (Both problems dont happen right now, but if they happen they will need workarounds)

This I would avoid because this is wasting a lot of time and e.g. CUDA 11.6 is very restrictive and avoids many implementations those normally should work.

ikbuibui: I didnt think the compiler would catch `note #2699-D`. Cool. A static assert in a seperate PR sounds fine

ikbuibui: > Removing the constructor will result into the fact that you ca not use `Vectro<int,3>{{1,2,3}}` anymore.

Is this now broken?
Only as a note, we should now be careful that the Vector and Storage classes don't define data members, else our aggregate initialization will fail.

psychocoderHPC: I removed it from this PR and provide the fix for CUDA 11.6 only. The initt list will come with a seperate PR

ikbuibui (approved):

PR #5070, opened 2024-09-03
reduce CI compile time for atomicPhysics compile test

psychocoderHPC:
- avoid compiling derived fields (added fileOutput.param)
- remove unused particle and density functors.

# Motivation

There is no need that our compile test contains functors which will be not used in the example or compiles more than required to test atomic physics compile.

### DIFF
--- a/share/picongpu/tests/compileAtomicPhysics/include/picongpu/param/particle.param
+++ b/share/picongpu/tests/compileAtomicPhysics/include/picongpu/param/particle.param
@@ -46,18 +46,6 @@ namespace picongpu::particles
              */
             static constexpr uint32_t numParticlesPerCell = 10u;
         };
-
-        struct OnePositionParameter
-        {
-            /** Count of particles per cell at initial state
-             *  unit: none
-             */
-            static constexpr uint32_t numParticlesPerCell = 1u;
-            // sit directly in lower corner of the cell
-            // each x, y, z in-cell position component in range [0.0, 1.0)
-            static constexpr auto inCellOffset = float3_X(0.0, 0.0, 0.0);
-        };
-        using OnePosition = OnePositionImpl<OnePositionParameter>;
         using Random10ppc = RandomImpl<RandomParameter10ppc>;
     } // namespace startPosition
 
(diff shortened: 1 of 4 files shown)


BrianMarre (approved):
Looks good!

share/picongpu/tests/compileAtomicPhysics/include/picongpu/param/particle.param
```diff
@@ -46,19 +46,6 @@ namespace picongpu::particles
...
-            /** Count of particles per cell at initial state
-             *  unit: none
-             */
-            static constexpr uint32_t numParticlesPerCell = 1u;
-            // sit directly in lower corner of the cell
-            // each x, y, z in-cell position component in range [0.0, 1.0)
-            static constexpr auto inCellOffset = float3_X(0.0, 0.0, 0.0);
-        };
-        using OnePosition = OnePositionImpl<OnePositionParameter>;
-        using Random10ppc = RandomImpl<RandomParameter10ppc>;
```
BrianMarre, line 61: Line 61 is still required

BrianMarre (changes requested):
@psychocoderHPC you also removed the definition of the 10ppc random particle functor used in the init pipeline

share/picongpu/tests/compileAtomicPhysics/include/picongpu/param/particle.param
PrometheusPi: @BrianMarre is right about that. The `Random10ppc` is used [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/2572a435a2b0a1e2121ee12cd885a9817ea0c571/share/picongpu/tests/compileAtomicPhysics/include/picongpu/param/speciesInitialization.param#L44). It is surprising for me that the CI still passes as this should cause a compile error. Do we execute this test? (EDIT: CI fails, didn't see that before)

PrometheusPi (changes requested):
One bug

share/picongpu/tests/compileAtomicPhysics/include/picongpu/param/particle.param
psychocoderHPC: re-added it

BrianMarre (approved):

PrometheusPi (approved):

PR #5072, opened 2024-09-04
reduce CI runtime

psychocoderHPC:
- Move single particle example to tests and rename it to `SingleParticle`.
- Reduce the number of variations for PIConGPU CI test to one for all exept the `compile` test where two combinations are compiled to guarantee that we cover 64bit.


ikbuibui (approved):
Changes look okay, but I do not know if the preprocessor macros set by cmake are used to conditionally include code in some places in these examples/tests thus disabling compilation of code.

psychocoderHPC:
> Changes look okay, but I do not know if the preprocessor macros set by cmake are used to conditionally include code in some places in these examples/tests thus disabling compilation of code.

Yes this is possible but as I know not happen, except for setting SIMDIM. This is nothing we can avoid but normally we do not use these defines to control the code workflow and defines to perfrom optional includes should only be used for optional PIConGPU dependencies.

PR #5078, opened 2024-09-09
avoid using SIMDIM for controlflow

### LINKED ISSUE #5062: SIMDIM not set error
In `dimsension.param` one has to use the pre-compiler definition `#define SIMDIM DIM2` since `SIMDIM` is also used outside the scope of `dimension.param` (in contrast to other pre-compiler setups). 
In case of setting the definition directly in C++ and not via pre-compiler:
 - CFL is assumed to be 1D (which we can not do)
 - and various code blocks will be used depending on their local 2D or 3D decision logic. 
 
 This 2-language-problem needs to be fixed. 

@psychocoderHPC will take care of this since he is currently in the process of refactoring this code anyway.

psychocoderHPC:
fix #5062

Avoid using the macro SIMDIM for program control-flow. SIMDIM should only be used for include guards.
Fix missing include in untested code IsaacPlugin.x.cpp.

### DIFF
--- a/include/picongpu/plugins/IsaacPlugin.x.cpp
+++ b/include/picongpu/plugins/IsaacPlugin.x.cpp
@@ -18,6 +18,8 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
+// required for SIMDIM definition
+#include "picongpu/defines.hpp"
 
 #if(ENABLE_ISAAC == 1) && (SIMDIM == DIM3)
 
--- a/include/picongpu/unitless/simulation.unitless
+++ b/include/picongpu/unitless/simulation.unitless
@@ -199,6 +199,9 @@ namespace picongpu
 
         static constexpr uint32_t dim()
         {
+            static_assert(
+                simDim == SIMDIM,
+                "Preprocessor define SIMDIM and simDim must be equal! see file dimension.param");
             return simDim;
         }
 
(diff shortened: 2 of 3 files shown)


include/picongpu/plugins/IsaacPlugin.x.cpp
```diff
@@ -18,7 +18,9 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "picongpu/defines.hpp"
 
+// required for SIMDIM definition
```
ikbuibui: Comment position is slightly odd

ikbuibui (dismissed):

include/picongpu/unitless/simulation.unitless
```diff
@@ -199,6 +199,9 @@ namespace picongpu
 
         static constexpr uint32_t dim()
         {
+            static_assert(
+                simDim == SIMDIM,
```
PrometheusPi, line 203: @psychocoderHPC I assume this is just a check for now until fully getting rid of `SIMDIM` in the code, right?

PrometheusPi (dismissed):

include/picongpu/plugins/IsaacPlugin.x.cpp
psychocoderHPC: ggrr yes :-(

psychocoderHPC: fixed

include/picongpu/unitless/simulation.unitless
psychocoderHPC: we can not get rid of the define SIMDIM because SIMDIM is required for include guards

chillenzer (approved):

PrometheusPi (approved):

PR #5079, opened 2024-09-09
change atomic physics particle attributes

psychocoderHPC:
Atomic physics particle attributes was set by default to a possible valid value but in case an algorithm forget to set the value correctly we got a false valid value.
By setting it to an invalid value such errors should become visible at leat by a crash of the simulation.
An additional advantage is that we can now give an electron some of the properties which allows to track from which ion state and which transition state the electron was created.

### DIFF
--- a/include/picongpu/param/speciesAttributes.param
+++ b/include/picongpu/param/speciesAttributes.param
@@ -156,16 +156,16 @@ namespace picongpu
     alias(atomicPhysicsParticle);
 
     //! particle attribute, collectionIndex of atomic state of ions
-    value_identifier(uint32_t, atomicStateCollectionIndex, 0u);
+    value_identifier(uint32_t, atomicStateCollectionIndex, std::numeric_limits<uint32_t>::max());
 
     //! particle attribute, process class of selected atomicPhysics transition
-    value_identifier(uint8_t, processClass, 0u);
+    value_identifier(uint8_t, processClass, std::numeric_limits<uint8_t>::max());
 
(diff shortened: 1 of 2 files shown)


BrianMarre:
also use the same value [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/b34250af68e9743e23e8440fe2087049065b678e/include/picongpu/particles/atomicPhysics/SetChargeState.hpp#L58)

BrianMarre (changes requested):
correct apart from wrong type used in type cast for `processClass`

include/picongpu/param/speciesAttributes.param
```diff
@@ -156,16 +156,16 @@ namespace picongpu
     alias(atomicPhysicsParticle);
 
     //! particle attribute, collectionIndex of atomic state of ions
-    value_identifier(uint32_t, atomicStateCollectionIndex, 0u);
+    value_identifier(uint32_t, atomicStateCollectionIndex, std::numeric_limits<uint32_t>::max());
 
     //! particle attribute, process class of selected atomicPhysics transition
-    value_identifier(uint8_t, processClass, 0u);
+    value_identifier(uint8_t, processClass, std::numeric_limits<uint32_t>::max());
```
BrianMarre: this needs to to cast to `uint8_t` not `uint32_t`

psychocoderHPC:
done

PrometheusPi (approved):

BrianMarre (approved):

PR #5080, opened 2024-09-09
refactor `value_identifier_func`

psychocoderHPC:
Refactor `value_identifier_func` and `SetAttributeToDefault` to allow deriving any kind of attribute from the source particle.
The refactoring required adding a default constructor for `Particle`.

### DIFF
--- a/include/pmacc/identifier/value_identifier.hpp
+++ b/include/pmacc/identifier/value_identifier.hpp
@@ -49,19 +49,22 @@
  * @{
  */
 
-/** @param ... must be a device function/lambda with arguments worker and  IdGenerator
+/** @param ... Must be a device function/lambda with arguments worker, IdGenerator and source particle.
+ *             @attention source particle can be invalid, isHandleValid() can be used to check the status.
+ *             To stay generic the lambda should be guarded against deriving from a particle which does not provide the
+ *             required attributes to initialize the value.
  *
  * @code{.cpp}
- * [] ALPAKA_FN_ACC(auto const& worker, IdGenerator& idGen) { return idGen.fetchInc(worker) };
+ * [] ALPAKA_FN_ACC(auto const& worker, IdGenerator& idGen, auto const& srcParticle) { return idGen.fetchInc(worker) };
  * @endcode
  */
 #define value_identifier_func(in_type, name, ...)                                                                     \
     identifier(                                                                                                       \
-        name, using type = in_type; template<typename T_Worker>                                                       \
-        DINLINE static type getValue(T_Worker const& worker, IdGenerator& idGen)                                      \
+        name, using type = in_type; template<typename T_Worker, typename T_SrcParticleType> DINLINE static type       \
+            getValue(T_Worker const& worker, IdGenerator& idGen, T_SrcParticleType const& srcParticle)                \
         {                                                                                                             \
             auto const func = __VA_ARGS__;                                                                            \
-            return func(worker, idGen);                                                                               \
+            return func(worker, idGen, srcParticle);                                                                  \
         } static std::string getName() { return std::string(#name); })
 
 /** getValue() is defined constexpr
@@ -74,7 +77,10 @@
  */
 #define value_identifier(in_type, name, ...)                                                                          \
     identifier(                                                                                                       \
-        name, using type = in_type; template<typename T_Worker> HDINLINE static constexpr type getValue(              \
+        name, using type = in_type; template<typename T_Worker, typename T_SrcParticleType>                           \
+        HDINLINE static constexpr type getValue(                                                                      \
             [[maybe_unused]] T_Worker const& acc,                                                                     \
-            [[maybe_unused]] IdGenerator& idGen) { return __VA_ARGS__; } HDINLINE static constexpr type getValue()    \
+            [[maybe_unused]] IdGenerator& idGen,                                                                      \
+            [[maybe_unused]] T_SrcParticleType const& srcParticle)                                                    \
+        { return __VA_ARGS__; } HDINLINE static constexpr type getValue()                                             \
         { return __VA_ARGS__; } static std::string getName() { return std::string(#name); })
--- a/include/pmacc/particles/memory/dataTypes/Particle.hpp
+++ b/include/pmacc/particles/memory/dataTypes/Particle.hpp
@@ -94,10 +94,10 @@ namespace pmacc
          * ATTENTION: The pointer must be the last member to avoid local memory usage
          *            https://github.com/ComputationalRadiationPhysics/picongpu/pull/762
          */
-        PMACC_ALIGN(frame, FrameType*);
+        PMACC_ALIGN(frame, FrameType*) = nullptr;
 
         /** index of particle inside the Frame*/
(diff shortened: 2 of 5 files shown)


include/pmacc/particles/operations/SetAttributeToDefault.hpp
```diff
@@ -38,15 +38,31 @@ namespace pmacc
...
+        template<typename T_Worker, typename T_DestParticleType>
+        HDINLINE void operator()(T_Worker const& worker, IdGenerator idGen, T_DestParticleType& destParticle)
+        {
+            using ResolvedAttr = typename pmacc::traits::Resolve<Attribute>::type;
+            /* set attribute to it's user defined default value */
+            destParticle[Attribute()] = ResolvedAttr::getValue(worker, idGen, T_DestParticleType{});
+        }
+
+        template<typename T_Worker, typename T_DestParticleType, typename T_SrcParticleType>
+        HDINLINE void operator()(
```
ikbuibui, line 56: I think separating this to another functor, something like, `SetAttributeFromParticle` is maybe better than having an overload of `SetAttributeToDefault` as I think the current name doesn't make sense.

ikbuibui (dismissed):

include/pmacc/identifier/value_identifier.hpp
```diff
@@ -49,19 +49,23 @@
  * @{
  */
 
-/** @param ... must be a device function/lambda with arguments worker and  IdGenerator
+/** @param ... Must be a device function/lambda with arguments worker, IdGenerator and source particle.
+ *             @attention source particle can be invalid, it is required to check the state via particles
+ *             isHandleValid() method.
+ *             To stay generic the lambda should be guarded against deriving from a particle which does not provide the
+ *             required attributes to initialize the value.
```
ikbuibui, line 55: Just for my understanding, do these lines mean the same as 
To stay generic the lambda does not check if the source particle provides the required attributes to initialize the value.

include/pmacc/identifier/value_identifier.hpp
```diff
@@ -49,19 +49,23 @@
  * @{
  */
 
-/** @param ... must be a device function/lambda with arguments worker and  IdGenerator
+/** @param ... Must be a device function/lambda with arguments worker, IdGenerator and source particle.
+ *             @attention source particle can be invalid, it is required to check the state via particles
+ *             isHandleValid() method.
```
ikbuibui: For my understanding, using `isHandleValid()` is not strictly required in the lambda correct? There may be attributes which dont need the particle to be valid.

include/pmacc/identifier/value_identifier.hpp
psychocoderHPC: No what I like to say is that the author of the lambda for the value_identiier is responsible to check and handle cases where the source particle is not providing the required attributes to derive the data.

include/pmacc/identifier/value_identifier.hpp
psychocoderHPC: Yes it's optional.

I will change it to `..., isHandleValid() can be used to check the status.`

include/pmacc/particles/operations/SetAttributeToDefault.hpp
psychocoderHPC: For short I thought about following your suggestion but I decided against it.

- Both methods are equivalent but for one no extern particle is required. Nevertheless the particle attributes were set to default.
- There is not need to encode the parameters into the name where `SetAttributeFromParticle` is exactly doing it.

I added a sentence to the doc string to point to value_identifier behaviors.

```
Check value_identifier and value_identifier_func for special behaviours on how defaults get derived.
```

psychocoderHPC:
@psychocoderHPC I pushed my changes

include/pmacc/particles/operations/SetAttributeToDefault.hpp
ikbuibui: Fair. I am okay with keeping it like this.
But maybe the best would be to rename the functor `SetAttribute` with these overloads. 
Attribute is set from a particle and if no source particle is provided, it uses the defaults.

psychocoderHPC: > Attribute is set from a particle and if no source particle is provided, it uses the defaults.

Ohh I think there is some misunderstanding about what this functor is doing. It will always set all attributes to the default.
A value_identifier also always gets a source particle that can be taken into account or not to derive the default value.
`value_identifier` has always a constant default value. `value_identifier_func` can derive the default from the id generator and or the source particle, where it could be that the source particle is invalid at the moment while `SetAttributeToDefault` is executed.
In any case what we set is the default.

include/pmacc/particles/operations/SetAttributeToDefault.hpp
```diff
@@ -38,15 +38,32 @@ namespace pmacc
...
+         *
+         * @{
          */
-        template<typename T_Worker, typename T_Particle>
-        HDINLINE void operator()(T_Worker const& worker, IdGenerator idGen, T_Particle& particle)
+        template<typename T_Worker, typename T_DestParticleType>
+        HDINLINE void operator()(T_Worker const& worker, IdGenerator idGen, T_DestParticleType& destParticle)
+        {
+            using ResolvedAttr = typename pmacc::traits::Resolve<Attribute>::type;
+            /* set attribute to it's user defined default value */
```
PrometheusPi: typo:
```suggestion
            /* set attribute to its user defined default value */
```

PrometheusPi (changes requested):
Sorry, minor typo

include/pmacc/particles/operations/SetAttributeToDefault.hpp
psychocoderHPC: fixed

psychocoderHPC:
@PrometheusPi CI finished

include/pmacc/particles/memory/dataTypes/Particle.hpp
```diff
@@ -94,10 +94,10 @@ namespace pmacc
          * ATTENTION: The pointer must be the last member to avoid local memory usage
          *            https://github.com/ComputationalRadiationPhysics/picongpu/pull/762
          */
```
chillenzer, line 96: As we are already touching this part of the code. Is this still a thing? If so, I think we're violating it here. If not, remove.

chillenzer (approved):
Just a minor thing.

include/pmacc/particles/memory/dataTypes/Particle.hpp
psychocoderHPC: It is independent of this PR the order was changed before.
Maybe this not not required anymore, I will open an issue that we can look into it.

psychocoderHPC: issue is opened: #5095

psychocoderHPC: @chillenzer btw good catch

PrometheusPi (approved):

PR #5077, opened 2024-09-06
CI:: fix download micro mamba

psychocoderHPC:
HZDR is blocking all `anaconda` server. If we download micromamba via the offical link `https://micro.mamba.pm/api/micromamba/linux-64/latest` the load balancer is sometimes pointing us to the anaconda mirrow which is not accesible from within HZDR.

The current workaround is to pull micromamba 1.5.9 directly from github.

PR #5082, opened 2024-09-09
Switch from mamba to uv

chillenzer:
#5077 seems to have been insufficient.

### DIFF
--- a/share/ci/install/pypicongpu.sh
+++ b/share/ci/install/pypicongpu.sh
@@ -27,20 +27,14 @@ cd $CI_PROJECT_DIR
 # use miniconda as python environment
 apt update && apt install -y curl
 cd /tmp/
-curl -Ls https://github.com/mamba-org/micromamba-releases/releases/download/1.5.9-0/micromamba-linux-64.tar.bz2 | tar -xvj bin/micromamba
-export MAMBA_ROOT_PREFIX=/tmp/mamba-forge/
-mkdir -p "${MAMBA_ROOT_PREFIX}"
-eval "$(./bin/micromamba shell hook -s posix)"
-export PATH=$(pwd -P)/bin:$PATH
-micromamba --version
-micromamba config append channels conda-forge
-micromamba config set channel_priority strict
+curl -LsSf https://astral.sh/uv/install.sh | sh
+source $HOME/.cargo/env
+uv venv --python="${PYTHON_VERSION}"
+source .venv/bin/activate
+python3 --version
 
 cd $CI_PROJECT_DIR
 # generates modified requirements.txt
-micromamba create -n pypicongpu python=${PYTHON_VERSION} --ssl-verify false
-micromamba activate pypicongpu
-python3 --version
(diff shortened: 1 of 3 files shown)


psychocoderHPC (dismissed):

BrianMarre:
PyPIConGPU test fails, might have something to do with this setting https://github.com/ComputationalRadiationPhysics/picongpu/blob/89b9640670127075f753c4bc63ec92aa99f5b1d6/.gitlab-ci.yml#L184

psychocoderHPC:
> PyPIConGPU test fails, might have something to do with this setting
> 
> https://github.com/ComputationalRadiationPhysics/picongpu/blob/89b9640670127075f753c4bc63ec92aa99f5b1d6/.gitlab-ci.yml#L184

@chillenzer I stopped the CI because this must be fixed

chillenzer:
Thanks for pointing this out. As far as I can see it, the syntax `.*` is not supported by `uv` for selecting a Python version. Pushed an update. Btw, why aren't we on `3.12`?

BrianMarre:
Because when we built the PyPIConGPU CI Python 3.12 wasn't out yet

psychocoderHPC:
I stopped the CI again because of errors https://gitlab.com/hzdr/crp/picongpu/-/jobs/7785698774

psychocoderHPC:
https://github.com/ComputationalRadiationPhysics/picongpu/blob/817b9ebbfec871bf76e8f374a7a974153aeb0663/share/ci/pypicongpu_generator.py#L327

is adding `.*` to the version number

chillenzer:
Nasty. I have now also run `grep -ri "python" share` and `grep -r '\.\* share` with no further results.

share/ci/install/pypicongpu.sh
```diff
@@ -27,20 +27,14 @@ cd $CI_PROJECT_DIR
 # use miniconda as python environment
```
PrometheusPi, line 27: comment is now outdated

PrometheusPi (changes requested):
minor comment correction

share/ci/install/pypicongpu.sh
chillenzer: Good catch. I'll remove it once the CI passed such that we can merge without another CI run.

psychocoderHPC:
It looks like it is not as simple as it sounds to switch the package manager for python.
Let's evaluate if we go on with the switch or stay with micromambda. For now, it looks like IT solved the issues by updating their black list.

chillenzer:
Apparently, IT has fixed the problem with `mamba` which makes this change unnecessary. Also, we have `snakemake` workflows which would make it particularly inconvenient to leave `mamba` completely. So, we'll close this PR for now.

PS: The failing CI probably just needs a renaming `pip3` -> `[uv ]pip` (not sure if we'd need/want `uv` in front there).

BrianMarre:
For future reference, Simeon recommended to switch to tox to simplify the PyPIConGPU generator, since tox apparently provides a native way to fix software versions for CI tests.

PR #5086, opened 2024-09-09
move stage particle boundaries to compile unit

psychocoderHPC:
- [ ] PR #5090 is combining this and other PRs to reduce the CI load

### DIFF
--- a/include/picongpu/simulation/stage/ParticleBoundaries.hpp
+++ b/include/picongpu/simulation/stage/ParticleBoundaries.hpp
@@ -19,20 +19,7 @@
 
 #pragma once
 
-#include "picongpu/simulation_defines.hpp"
-
-#include "picongpu/particles/Particles.hpp"
-#include "picongpu/particles/boundary/Kind.hpp"
-
-#include <pmacc/Environment.hpp>
-#include <pmacc/meta/ForEach.hpp>
-#include <pmacc/particles/traits/FilterByFlag.hpp>
-
 #include <boost/program_options.hpp>
-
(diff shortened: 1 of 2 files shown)


include/picongpu/simulation/stage/ParticleBoundaries.hpp
```diff
@@ -212,26 +40,13 @@ namespace picongpu
...
-                {
-                    processSpecies();
-                }
-
-            private:
-                //! Only allow customization for species with pusher
-                using SpeciesWithPusher =
-                    typename pmacc::particles::traits::FilterByFlag<VectorAllSpecies, particlePusher<>>::type;
-                //! Functor to process all affected species
-                meta::ForEach<SpeciesWithPusher, detail::ParticleBoundariesCommandLine<boost::mpl::_1>> processSpecies;
```
ikbuibui, line 234: Just for information, why was this moved into both the functions rather than being left here

psychocoderHPC: This avoids that this compile-time algorithm is executed even if it is not required.
The evaluation is now in the cpp because it is only there required.
This reduces the compile time.

ikbuibui (approved):

psychocoderHPC:
merged together with https://github.com/ComputationalRadiationPhysics/picongpu/pull/5090 due to rebases of https://github.com/ComputationalRadiationPhysics/picongpu/pull/5090 in between github was not able to match this PR and auto close it.

PR #5094, opened 2024-09-10
PMacc: refactor `FileSystem.hpp`

psychocoderHPC:
- remove dimension from the filesystem class
- remove Filesystem from pmacc::Environment<>

In the past, we thought that it was good to get access to all singletons from a single object, the Environment.
This introduces many issues, e.g. cyclic includes. Additionally, we templated all objects with the simulation dimension.

This PR removes the dimension from the Filesystem object and allows the creation of the filesystem singleton from everywhere.

### DIFF
--- a/include/pmacc/mappings/simulation/Filesystem.cpp
+++ b/include/pmacc/mappings/simulation/Filesystem.cpp
@@ -19,9 +19,6 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
-#if __cplusplus < 201703L
-#    error "C++17 is required"
-#endif
 
 #include "pmacc/mappings/simulation/Filesystem.hpp"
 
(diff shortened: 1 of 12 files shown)


include/pmacc/mappings/simulation/Filesystem.cpp
```diff
@@ -31,15 +28,13 @@
 
 namespace pmacc
 {
-    template<unsigned DIM>
-    void Filesystem<DIM>::createDirectory(const std::string dir) const
+    void Filesystem::createDirectory(const std::string dir) const
     {
         /* does not throw if the directory exists or has been created */
```
chillenzer: Would you mind adding this minor clarification while you're touching this file?
```suggestion
        /* using `create_directories` instead of `create_directory` because the former does not throw if the directory exists or has been created */
```

chillenzer (dismissed):
Sounds like a very good idea. Found an unrelated comment that could do with a minor clarification but nothing blocking (as long as CI is happy).

include/pmacc/mappings/simulation/Filesystem.cpp
psychocoderHPC: fixed, thx

ikbuibui (approved):

PR #5096, opened 2024-09-10
Adjust openPMD alias in python

### LINKED ISSUE #4802: Use other alias for `openpmd_api`

chillenzer:
Closes #4802 .

I used a simple `sed`:
```bash
sed 's/\<io\>/opmd/g' -i $(grep -rl 'import openpmd')
```

### DIFF
--- a/docs/source/usage/plugins/radiation.rst
+++ b/docs/source/usage/plugins/radiation.rst
@@ -469,8 +469,8 @@ Please be aware that all datasets in the openPMD output are given in the PIConGP
 
 .. code:: python
 
-   import openpmd_api as io
-   series = io.Series("e_radAmplitudes_%T.bp", io.Access_Type.read_only)
+   import openpmd_api as opmd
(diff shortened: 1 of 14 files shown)


BrianMarre:
You do know that the quick tests only cover the PIConGPU PICMI interface and PyPIConGPU? Which do not use openPMD...

chillenzer:
Mmmhhh... so it was a particularly easy task to not make them fail. I guess I could have still managed to do so... -.-

BrianMarre (approved):
Looks good!

psychocoderHPC (approved):

PR #5097, opened 2024-09-10
Add automatic adjustment docs

### LINKED ISSUE #4830: [Docs] Automatic adjustment of RT arguments
In trying to fix a link in the documentation for my #4812, I've realised that I couldn't find any mentioning of the automatic adjustment of arguments in the documentation. Should we add that?

chillenzer:
Closes #4830. It's not the most prominent place to put this information but it's better than nothing I guess.

### DIFF
--- a/docs/source/usage/tbg.rst
+++ b/docs/source/usage/tbg.rst
@@ -45,6 +45,25 @@ Feel free to copy & paste sections of the files below into your ``.cfg``, e.g. t
 .. literalinclude:: ../../TBG_macros.cfg
    :language: bash
 
+
+Automatic Grid Adjustment
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+It is important to note that PIConGPU has a feature to automatically adjust the grid to comply with PIConGPU's constraints.
+This feature is ACTIVE by default, meaning that unless explicit disabled PIConGPU does not guarantee that the grid you submitted is actually run.
+This can be deactivated by the commandline flag `--autoAdjustGrid 0`.
+If automatic adjustment kicks in, it prints a warning.
+If automatic adjustment is deactivated, the run will fail during the initialisation for incompatible parameters.
+
+The checks performed on the given grid parameters are the following:
+
+.. literalinclude:: ../../../include/picongpu/simulation/control/DomainAdjuster.hpp
+   :language: C++
+   :start-after: doc-include-start: automatic-grid-adjustment
+   :end-before: doc-include-end: automatic-grid-adjustment
+   :dedent:
+
+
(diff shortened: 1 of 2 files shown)


docs/source/usage/tbg.rst
```diff
@@ -45,6 +45,25 @@ Feel free to copy & paste sections of the files below into your ``.cfg``, e.g. t
 .. literalinclude:: ../../TBG_macros.cfg
    :language: bash
 
+
+Automatic Grid Adjustment
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+It is important to note that PIConGPU has a feature to automatically adjust the grid to comply with PIConGPU's constraints.
+This feature is ACTIVE by default, meaning that unless explicit disabled PIConGPU does not guarantee that the grid you submitted is actually run.
+This can be deactivated by the commandline flag `--autoAdjustGrid=0`.
```
psychocoderHPC: should be `--autoAdjustGrid 0`

include/picongpu/simulation/control/DomainAdjuster.hpp
```diff
@@ -81,9 +81,16 @@ namespace picongpu
 
             for(uint32_t d = 0; d < simDim; ++d)
             {
+                // doc-include-start: automatic-grid-adjustment
+                // The checks are performed for each dimension `d` individually.
+                // If not, the size in that dimension is rounded up to the nearest multiple:
                 multipleOfSuperCell(d);
+                // We need at least one CORE cell and one GUARD cell per MPI rank on either side. If not, this
```
psychocoderHPC: `We need at least one CORE and two GUARD supercell per MPI rank.`

The former text could be misinterpreted that two CORE supercells are required. I changed also cell to supercell.

psychocoderHPC (changes requested):

docs/source/usage/tbg.rst
chillenzer: Done

include/picongpu/simulation/control/DomainAdjuster.hpp
chillenzer: I hope I've found the best of both worlds. Done.

docs/source/usage/tbg.rst
chillenzer: PS: Both do exactly the same but if your suggestion is more in line with the style of the documentation I've changed it.

PrometheusPi (approved):
looks good to me 
@psychocoderHPC are you fine too?

psychocoderHPC (approved):

PR #5099, opened 2024-09-11
fix warning in PyPIConGPU version tests

SimeonEhrig:
- If the script `requirements_txt_modifier.py` should modify a dependency in a requirements.txt, which does not exist, it throws a warning because this is unexpected behavior. Because warnings are hard to detect in the CI, change it to an error.
- Implement a function to ignore specific environment variables which are set to change the version in a requirements.txt.

Example: https://gitlab.com/hzdr/crp/picongpu/-/jobs/7789974598

### DIFF
--- a/share/ci/install/pypicongpu.sh
+++ b/share/ci/install/pypicongpu.sh
@@ -43,8 +43,17 @@ micromamba activate pypicongpu
 python3 --version
 MODIFIED_REQUIREMENT_TXT_PICMI=$CI_PROJECT_DIR/lib/python/picongpu/picmi/modified_requirements.txt
 MODIFIED_REQUIREMENT_TXT_PYPICONGPU=$CI_PROJECT_DIR/lib/python/picongpu/pypicongpu/modified_requirements.txt
-python3 $CI_PROJECT_DIR/share/ci/install/requirements_txt_modifier.py $CI_PROJECT_DIR/lib/python/picongpu/picmi/requirements.txt $MODIFIED_REQUIREMENT_TXT_PICMI
-python3 $CI_PROJECT_DIR/share/ci/install/requirements_txt_modifier.py $CI_PROJECT_DIR/lib/python/picongpu/pypicongpu/requirements.txt $MODIFIED_REQUIREMENT_TXT_PYPICONGPU
+# The environment variables for changing the version of a dependency package are set globally.
+# Therefore, all calls of the requirements_txt_modfifier.py script are affected.
+# To avoid unexpected changes to dependencies, the list of ignored environment variables.
+python3 $CI_PROJECT_DIR/share/ci/install/requirements_txt_modifier.py \
+    -i $CI_PROJECT_DIR/lib/python/picongpu/picmi/requirements.txt \
+    -o $MODIFIED_REQUIREMENT_TXT_PICMI \
+    --ignore_env_args PYPIC_DEP_VERSION_referencing PYPIC_DEP_VERSION_jsonschema
+python3 $CI_PROJECT_DIR/share/ci/install/requirements_txt_modifier.py \
+    -i $CI_PROJECT_DIR/lib/python/picongpu/pypicongpu/requirements.txt \
+    -o $MODIFIED_REQUIREMENT_TXT_PYPICONGPU \
+    --ignore_env_args PYPIC_DEP_VERSION_pydantic PYPIC_DEP_VERSION_picmistandard
 
 echo "modified_requirements.txt: "
 cat $MODIFIED_REQUIREMENT_TXT_PICMI
--- a/share/ci/install/requirements_txt_modifier.py
+++ b/share/ci/install/requirements_txt_modifier.py
@@ -1,25 +1,21 @@
-import os
-import sys
-
 """
 This file is part of PIConGPU.
 Copyright 2023-2023 PIConGPU contributors
 Authors: Simeon Ehrig
 License: GPLv3+
-"""
 
-"""@file Fix package version in requirement.txt to a specific version.
+@file Fix package version in requirement.txt to a specific version.
 
 Reads an existing requirements.txt file, sets one or more of the packages to a
 fixed version and creates a new requirements.txt file from the modified
 hard set packages and fills up with the remaining packages.
 
-@param First application argument: Path to the original requirements.txt
-@param Second application argument: Path to the mew requirements.txt
-
-@attention Versions of the packages are set via environment variables. The
-variables need to have the shape of: PYPIC_DEP_VERSION_<package_name>=<version>
+Run `python requirements_txt_modifier.py --help` to check the usage.
 """
+
+import os
+import sys
+import argparse
 
 
 def cs(text: str, color: str):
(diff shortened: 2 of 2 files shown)


SimeonEhrig:
@BrianMarre please review

BrianMarre (approved):
Looks good!

psychocoderHPC:
@SimeonEhrig Does it mean the test you pointed to will now fail with this PR?

SimeonEhrig:
The PR also contains the fix for the current setup. But in future, if the setup will be changed, hopefully yes.

share/ci/install/requirements_txt_modifier.py
chillenzer, line 1: Please also adjust the docstring and add the further arguments [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/5d6a9a21cb7c59d4228e79e56c35f5ab144b1ecf/share/ci/install/requirements_txt_modifier.py#L18) (sorry, can't comment so far away from your changes).

share/ci/install/pypicongpu.sh
```diff
@@ -43,8 +43,8 @@ micromamba activate pypicongpu
 python3 --version
 MODIFIED_REQUIREMENT_TXT_PICMI=$CI_PROJECT_DIR/lib/python/picongpu/picmi/modified_requirements.txt
 MODIFIED_REQUIREMENT_TXT_PYPICONGPU=$CI_PROJECT_DIR/lib/python/picongpu/pypicongpu/modified_requirements.txt
-python3 $CI_PROJECT_DIR/share/ci/install/requirements_txt_modifier.py $CI_PROJECT_DIR/lib/python/picongpu/picmi/requirements.txt $MODIFIED_REQUIREMENT_TXT_PICMI
-python3 $CI_PROJECT_DIR/share/ci/install/requirements_txt_modifier.py $CI_PROJECT_DIR/lib/python/picongpu/pypicongpu/requirements.txt $MODIFIED_REQUIREMENT_TXT_PYPICONGPU
+python3 $CI_PROJECT_DIR/share/ci/install/requirements_txt_modifier.py PYPIC_DEP_VERSION_referencing PYPIC_DEP_VERSION_jsonschema $CI_PROJECT_DIR/lib/python/picongpu/picmi/requirements.txt $MODIFIED_REQUIREMENT_TXT_PICMI
```
chillenzer: Just for my information: Where do these environment variables come from?

share/ci/install/requirements_txt_modifier.py
```diff
@@ -53,21 +53,27 @@ def cs(text: str, color: str):
 
 if __name__ == "__main__":
     if len(sys.argv) < 3:
-        print("Set path to requirements.txt as first argument and output path as" " second argument.")
-        exit(1)
+        print("Set the path of the input requirements.txt as the second last argument "
+              "and the path of the output requirements.txt as the last argument.")
+        print("Set the name of the environment variables to ignore "
+              "before the path of the input requirements.txt")
+        sys.exit(1)
```
chillenzer: Why do you use `sys.exit(1)` here? Wouldn't an exception have the same effect but be more specific and flexible? (Same at the bottom, of course.)

share/ci/install/requirements_txt_modifier.py
```diff
@@ -53,21 +53,27 @@ def cs(text: str, color: str):
...
+                packages[envvar.split("_")[-1]] = os.environ[envvar]
 
     print("Try to set following package to a fix version")
     for pkg_name, pkg_version in packages.items():
         print(f"  {pkg_name} -> {pkg_version}")
 
-    with open(sys.argv[1], "r", encoding="utf-8") as input:
-        with open(sys.argv[2], "w", encoding="utf-8") as output:
+    with open(sys.argv[-2], "r", encoding="utf-8") as input:
+        with open(sys.argv[-1], "w", encoding="utf-8") as output:
```
chillenzer: This looks horribly dangerous. I'm aware that we're guarding against less than 3 arguments above but if the next person touching this file isn't aware of that, we might write the content of this very Python script into the output file. Maybe we could already extract the filenames in the else branch of the `if len(...` condition above, so there is at least proximity between those.

share/ci/install/requirements_txt_modifier.py
```diff
@@ -53,21 +53,27 @@ def cs(text: str, color: str):
 
 if __name__ == "__main__":
     if len(sys.argv) < 3:
-        print("Set path to requirements.txt as first argument and output path as" " second argument.")
-        exit(1)
+        print("Set the path of the input requirements.txt as the second last argument "
+              "and the path of the output requirements.txt as the last argument.")
+        print("Set the name of the environment variables to ignore "
+              "before the path of the input requirements.txt")
```
chillenzer: Wouldn't we want to print them coloured as below?

chillenzer (changes requested):
Looks good but there are one or two things to discuss.

share/ci/install/requirements_txt_modifier.py
SimeonEhrig: This is the help text, which you get, if you run the script without arguments. Normally it is white.

share/ci/install/requirements_txt_modifier.py
SimeonEhrig: No, you call a normal application without required arguments, it say you that the arguments are missing. But it will not print the call stack and other technical information.

share/ci/install/requirements_txt_modifier.py
SimeonEhrig: If I want make it save, I need to use `argparse`. I check the option.

SimeonEhrig:
I didn't fixed the comments. I pushed again to trigger the CI to see, if the script is called somewhere else. I will fix the issues later.

share/ci/install/requirements_txt_modifier.py
chillenzer: You are correct. But I think you commented on the wrong comment here, right?

share/ci/install/requirements_txt_modifier.py
chillenzer: Yes, that's okay.

share/ci/install/requirements_txt_modifier.py
chillenzer: As I said, an intermediate step towards safety could be to extract those in the `if` condition.

share/ci/install/requirements_txt_modifier.py
chillenzer: I think you wanted your other comment to appear here. I'll resolve this one.

share/ci/install/pypicongpu.sh
SimeonEhrig: The script `pypicongpu_generator.py` generates the jobs dynamically during the CI pipeline lifetime. It set also the environment variables.

SimeonEhrig:
@chillenzer Can you check it again. I think now it looks much better.

share/ci/install/pypicongpu.sh
```diff
@@ -43,8 +43,17 @@ micromamba activate pypicongpu
 python3 --version
 MODIFIED_REQUIREMENT_TXT_PICMI=$CI_PROJECT_DIR/lib/python/picongpu/picmi/modified_requirements.txt
 MODIFIED_REQUIREMENT_TXT_PYPICONGPU=$CI_PROJECT_DIR/lib/python/picongpu/pypicongpu/modified_requirements.txt
-python3 $CI_PROJECT_DIR/share/ci/install/requirements_txt_modifier.py $CI_PROJECT_DIR/lib/python/picongpu/picmi/requirements.txt $MODIFIED_REQUIREMENT_TXT_PICMI
-python3 $CI_PROJECT_DIR/share/ci/install/requirements_txt_modifier.py $CI_PROJECT_DIR/lib/python/picongpu/pypicongpu/requirements.txt $MODIFIED_REQUIREMENT_TXT_PYPICONGPU
+# The environment variables for changing the version of a dependency package are set globally.
+# Therefore, all calls of the requirements_txt_modfifier.py script are affected.
+# To avoid unexpected changes to dependencies, the list of ignored environment variables.
```
chillenzer, line 48: Very good! Thanks!

share/ci/install/requirements_txt_modifier.py
```diff
@@ -52,35 +53,63 @@ def cs(text: str, color: str):
...
+        type=str,
+        required=True,
+        help="Set the path of the output requirements.txt",
+    )
+    parser.add_argument(
+        "--ignore_env_args",
+        type=str,
+        nargs="*",
+        default=[],
+        help="Ignore this environment variables, which are set to modify the requirements.txt. The environment variables starts with `PYPIC_DEP_VERSION_`.",
```
chillenzer:
```suggestion
        help="Ignore these environment variables, which are set to modify the requirements.txt. The environment variables starts with `PYPIC_DEP_VERSION_`.",
```

share/ci/install/requirements_txt_modifier.py
chillenzer: I think you never responded to this one. It would be nice to add a line or two after line 19.

chillenzer (dismissed):
There are some minor things but I'd say nothing strictly blocking.

SimeonEhrig:
ping @chillenzer

chillenzer (approved):
Thought I already gave my approval but here it comes again.

PR #5101, opened 2024-09-11
remove cell type `centered`

psychocoderHPC:
- remove the trait getCellType
- remove cell type centered
- create file `ÝeeCell.hpp` and move Yee cell definition to this file
- refactor incident field traits to resolve include order issues

To simplify the code we removed the possibility to use different cell types in PIConGPU.
We kept it in the past but in the current code, we often used already very often the Yee cell and any algorithms e.g. current depeosion require the Yee cell.

Note: the refactoring in the incident field trait part was necessary because before it was not possible to resolve the cyclic include dependencies.

This change removes another source of cyclic includes.


ikbuibui (dismissed):

include/picongpu/fields/CellType.hpp
```diff
@@ -1,36 +0,0 @@
...
-#include "picongpu/fields/MaxwellSolver/Solvers.hpp"
-#include "picongpu/traits/GetCellType.hpp"
-
-
-namespace picongpu
-{
-    namespace fields
-    {
-        //! Alias for a cell type used by the field solver
-        using CellType = traits::GetCellType<Solver>::type;
```
chillenzer, line 33: Why didn't you do
```C++
using CellType = cellType::Yee;
```
here? If we're not planning to support multiple different cell types, the below syntax `fields::cellType::Yee` seems like an unnecessary indirection.

Thinking this to the end, we could even move the code of `cellType::Yee` into this file directly.

include/picongpu/traits/GetMargin.hpp
```diff
@@ -19,8 +19,6 @@
 
 #pragma once
 
-#include "picongpu/simulation_defines.hpp"
-
```
chillenzer, line 23: Wohooo! That's what I wanna see!

chillenzer (dismissed):
For the unrelated includes added, I have assumed that you know what you're doing. There's just one thing we should quickly discuss before merging.

psychocoderHPC:
Note: The additional includes coming from the fact that I removed a include that was in `CellType.hpp`. Because of this change on other places definitions are missing.

include/picongpu/fields/CellType.hpp
psychocoderHPC: I took your comment seriously and refactored it all again.

I added a file `YeeCell.hpp` with the cell description. Due to cyclic includes it was required to refactor the trait part in the incident field folder. A reason for the issues was that the trait file contained different traits and trait specialization for profiles. I separated those to remove that profiles included traits and traits includes profiles.

ikbuibui (approved):

PR #5104, opened 2024-09-11
Clarify use of snakemake params in scripts

### LINKED ISSUE #4958: Explanation snakemake params usage
The documentation of how to use snakemake contains the following statement:

>The parameters set with the params keyword, can be accessed in your python script via snakemake.params[i] or snakemake.params.name_of_param. Accordingly one can use snakemake.input or snakemake.output.

This is great to use in python, but intrinsically these parameters are of a snakemake specify type and need casting before usage. Perhaps that should be added to the documentation.

chillenzer:
I've taken the liberty to act on #4958 ~~and added a handy little idiom on top~~.

Closes #4958.

### DIFF
--- a/docs/source/usage/parameter_scans.rst
+++ b/docs/source/usage/parameter_scans.rst
@@ -157,7 +157,8 @@ Here is an example of what this might look like for a Python script called ``pos
             "post_processing.py"
 
 The parameters set with the ``params`` keyword, can be accessed in your python script via `snakemake.params[i]` or `snakemake.params.name_of_param`. 
-Accordingly one can use ``snakemake.input`` or ``snakemake.output``.
+Accordingly one can use ``snakemake.input`` or ``snakemake.output``. Please note that they are list/dictionary-like Snakemake-specific types.
+More details can be found in `Snakemake's documentation <https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#python>`_.
 
 To perform this evaluation on the cluster, add the required resource to the "config.yaml". For example, like this:
 


JessicaTiebel (approved):
The trick with falling back to commandline arguments seems really handy. I like it. 🙂

docs/source/usage/parameter_scans.rst
```diff
@@ -157,7 +157,29 @@ Here is an example of what this might look like for a Python script called ``pos
...
+            extraordinarily_cool_parameter = snakemake.params.fancy_name
+            # ...
+        except NameError:
+            # Apparently, snakemake didn't exist. Let's check the commandline.
+            input_filename = sys.argv[1]
+            extraordinarily_cool_parameter = sys.argv[2]
+            # or something more elaborate like argparse, etc.
+            # ...
+
+        main(input_filename, extraordinarily_cool_parameter)
```
PrometheusPi: I am a bit confused, what this code snipped should do-

chillenzer: This snippet is supposed to show a prototypical "if-main" section from some analysis script you might have written. It assumes that you have your script well structured into a `main(...)` that takes parameters from the outside (e.g. the commandline) and the very short section of free code that extracts the parameters and passes them to `main()`. I hoped for this to be straightforward to recognise.

Now, the snippet prepares the code to be also usable from within Snakemake. It attempts to extract the information from the `snakemake` object that gets injected by Snakemake upon running it, but falls back to using the commandline if that didn't work. Phrased differently: It detects, if we're running under Snakemake and uses a different mechanism for taking arguments than when you run it by hand. If you don't do this, these two methods of running the script would be exclusive (because running by hand you'd run into exactly the `NameError` we're catching and running from Snakemake you'd not be able to provide commandline arguments ­-- unless you're using the `shell` directive).

I personally find this very handy because it's oftentimes much easier to debug a script if you can run it by hand as opposed to firing up the whole machinery of Snakemake which might perform unrelated work and have non-deterministic aspects to it.

JessicaTiebel: I want to mention that I also stumbeled upon this section quite a bit and needed some minutes to figure out what it is supposed to do. Perhaps it would be nice to clarify this a bit more in the docu.

chillenzer: Offline discussion: This recommendation is worthwhile but its placement was ill-advised. I have split this change off and will open a separate PR.

chillenzer:
Sorry, messed up something during the rebase. I'll un-draft it in a minute.

JessicaTiebel (approved):

PrometheusPi (approved):

PR #5105, opened 2024-09-11
fix case insensitive bug

PrometheusPi:
With #5045, we introduced a naming change that causes a cyclic renaming between `absober.hpp` and `Absorber.hpp` on all  case-insesitive file sysetms (e.g. MacOS) when working with git, raping users in an "infinite loop of unstaged changes". 
This pull request should fix this issue.


psychocoderHPC (approved):

chillenzer (approved):
I think that "raping users" like you is intolerable behaviour and therefore I strongly support this PR.

PR #5107, opened 2024-09-11
CI: add PIConGPU header check

psychocoderHPC:
- check if all PIConGPU header files include their dependencies directly or transitively
- fix include issues
- add CI chain to validate PIConGPU headers

**Note:** Currently the test is only testing a few folders within PICoNGPU where it was easy to fix the include cycles.
More folders should be added in the future but this requires larger refactorings.


share/ci/generate_reduced_matrix.sh
```diff
@@ -66,6 +66,13 @@ elif [ "$PIC_INPUTS" == "pmacc_header" ] ; then
     # create header validation test for PMacc
     echo "pmacc_header" | tr " " "\n" | n_wise_generator.py $@ --limit_boost_version $ADDITIONAL_GENRATOR_FLAGS
   fi
+elif [ "$PIC_INPUTS" == "picongpu_header" ] ; then
+  if [ $commit_pmacc_only -ne 1 ] ; then
+    # create unit tests for PIConGPU
+    echo "picongpu_header" | tr " " "\n" | n_wise_generator.py $@ --limit_boost_version $ADDITIONAL_GENRATOR_FLAGS
```
chillenzer, line 74:
```suggestion
    echo "picongpu_header" | tr " " "\n" | n_wise_generator.py $@ --limit_boost_version $ADDITIONAL_GENERATOR_FLAGS
```
There's a typo in that variable's name (in general, not just this line). No need to fix immediately, but one could...

chillenzer (dismissed):
Nice one!

share/ci/generate_reduced_matrix.sh
psychocoderHPC: I will do this change in a separate PR because #5108 is fixing a bug and both PRs will collide if I rename the variable. After both are merged I can perform the renaming.

ikbuibui (dismissed):

test/picongpuHeaderCheck/CMakeLists.txt
```diff
@@ -0,0 +1,128 @@
...
+#
+# This file is part of PIConGPU.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+
+cmake_minimum_required(VERSION 3.18)
+project(PMaccHeaderTest)
```
ikbuibui, line 12: Is this project name correct? is it even important?

ikbuibui (approved):

PR #5131, opened 2024-09-19
refactor: `HBAR`, `RYDBERG_ENERGY` and `BOHR_RADIOUS`

psychocoderHPC:
Move variables to the corresponding function into the PIConGPU `sim` object. 
Provide methods to convert from joule to eV and back.

The param files changed but this should not affect the user because `physicalConstants.param` is typically not part of the user setup.


ikbuibui (approved):

BrianMarre (commented):
looks good

PR #5148, opened 2024-10-01
Add FromOpenPMDPulse profile

fabidie:
TL;DR: This is the same PR as before. I opened a new one because rebasing was a mess...

The FromOpenPMDPulse profile is an IncidentField method that reads a laser pulse in space-time domain from openPMD into PIConGPU. 

Furthermore, there are a python script provided at picongpu/lib/python/picongpu/extra/input to prepare an INSIGHT measurement as input for this laser profile. Such preparation steps are:
- restoring the missing phase relation between the measured frequencies
- propagation out of the focal plane (where the measurement took place)
- transformation from frequency to tome domain
- add group velocity / third order dispersion
- measure angular / spatial dispersion
- correct possible measurement artifacts
- save to openPMD
Further explanation and an example workflow can be found in the jupyter notebook at the same place.

Currently, there are some restrictions using the FromOpenPMDPulse profile:
- propagation and polarization only work along the axes, **not** diagonally!
- the pulse will always be aligned centered at the incident plane
- one needs to increase the reserved GPU memory size in memory.param - since the field data chunk will be stored on each used device at timestep 0 (and not during init, before allocating the particle memory), the simulation could otherwise run into memory issues.
- the Insight measurement scales have to be mm and fs, otherwise the preparation routine has to be adapted
- the time axis is read in inverted (which is wrong), but it only works this way (apparently cancelling out another error either in the preparation routine or in the incident field method). Until the bug is found, the wrong direction has to be kept.

There have been various tests to validate this new laser profile:
#### 1. Using a Gaussian pulse profile

A Gaussian pulse field chunk has been read from openPMD into the simulation. The focus was set to cell 26 (in propagation direction $y$, where $dy = \lambda/10$), where the field was read at every simulation timestep during the pulse duration. This was compared to a simulation with the already existing GaussianPulse profile with the same parameters. Since the electric field oscillates strongly in the longitudinal direction, a direct comparision is difficult (small shifts in time or of the incident field plane position can already cause quite some differences), both simulation outputs were transformed into the fourier domain:
spectral_abs_int
spectral_abs_int_log
spectral_angle_center
As one can see, it fits quite nice; differences start only to appear at small intensity values and are probably caused by numerical precision. With an inverse fourier transform, the temporal envelope was reconstructed: 
time_env_center
time_env_center_log
Again, the only differences appear at small intensity values; *but only* when the simulation timestep and the time sampling of the field chunk are adjusted to each other. Otherwise the temporal envelope can look wiggly, which leads to deviating results when it comes to interaction with plasma. So it is recommended to adjust those timesteps to each other.

#### 2. Using a Gauss-Laguerre pulse
According to the definition of a Gauss-Laguerre pulse in the GaussianPulse profile, a Gauss-laguerre pulse field chunk has been written to OpenPMD and read into PIConGPU. The pulse parameters were $\lambda$ = 800nm, $\tau = 30/2\sqrt{2 ln 2}$ fs, $w_0$ = 10 µm and $z$ = $-z_R$ = -0.393mm. As Laguerre coefficients, the example set in GaussianPulse.def was used. For two different positions, the simulation results ("PIC") were compared to the analytical solution ("Original"):
**a) at $z \approx -z_R$**
xyPlaneIte1285
(amplitude along the longitudinal direction)
energyDensityIte1285
(energy density, definition see below)
**b) at the focus position**
xyPlaneIte5597
(amplitude along the longitudinal direction,  the blue line is the focus position at this time step)
energyDensityIte5597
(energy density, definition see above)
focusDiffIte5597
(transversal integrated amplitude,  the black line is the focus position at this time step)
In the above mentioned simulations, the transversal simulation cell was square-shaped, as well as the transversal extent of the test pulse field chunk. To test also the asymmetrical case, the simulation cell in $x$ direction was halved in size. The results were similar: 
energyDensityIte5649
(energy density for an asymmetrical cell size; the asymmetry in amplitude shows that the energy is conserved better for a smaller cell size)
Then, the cell size was reset to original (transversal square-shaped), but the (original) field chunk was chosen to be asymmetrical:
originalChunkExtent
(original Gauss-Laguerre field at $z=-z_R$, before reading it into the simulation)
The resulting simulation did behave just the same as above, e.g. the pulse was still centered and not distorted after propagating it into the focus:
yxPlaneIte5597
yzPlaneIte5597
(amplitude in longitudinal direction, along both transversal directions; blue is the focus position)
energyDensityIte5597
( energy density)

#### 3. Using an Insight measurement

For that, the measured field data has been propagated out of the focus by one rayleigh length (-1,28 mm) during the preparation routine, saved to openPMD and then read into the simulation. The simulation ran until the pulse reached the focal plane again (in vacuum), where the simulation output was compared to the original INSIGHT measurement: 
VacuumValidation
The shape, amplitude and extent of both pulses fit well.
A cut through both beam centers reveals the longitudinal pulse envelope:
VacuumValidation2
Here, one can see slight deviations. They are mostly caused by field solver dispersion (Yee was used with $dy = \lambda/10$ over a propagation distance of 1.28 mm) and can be improved with a finer grid. Furthermore, there is a spatial axis compared to a temporal one ("ct"), but since the pulse is short, this approximation should be valid, since pulse divergence does not play a role at that small scales.
The energy density (longitudinally integrated, $\mathcal{W} = \int dz \frac{\epsilon_0}{2} (E^2 + c^2 B^2)$ ) shows that also the transverse pulse extent in the simulation fits welll to the expected one:
VacuumValidation_energyDensity
Only the x axis is switched because to keep the system handiness when choosing  a different propagation and polarization direction (y, -x)  than the original measurement (z, x). Also, the pulse energy is the same, only a bit smaller in PIConGO again due to field solver dispersion. When calculating the beam energy via $W = \int dV \frac{\epsilon_0}{2} (E^2 + c^2 B^2)$, the simulation output lies slightly below the input (7% less). The loss is in the range of field solver dispersion.

Furthermore, there were tests with different propagation and polarization directions, where the field at the incident plane ("PIC") has been compared to the input ("Original") data:
incidentPlaneIte1800
(propagation in z direction, polarization in x direction, just as the original)
incidentPlaneNegPolIte1800
(propagation in z direction, polarization in -x direction; this automatically also swaps the y axis to keep the original pulse orientation)
xzPlaneNegPropIte1800
(propagation in -z direction, polarization in x direction; the time axis of the original data has been transformed to a spatial one by multiplication with $c$ to allow a better comparison)

Tests for different (transversal) simulation window extents, different transverse grid spacing ($w_0/6$ and $w_0/10$) and for a different number of devices (1, 2, 4) also lead to the expected results.

### DIFF
--- a/docs/source/usage/workflows/addLaser.rst
+++ b/docs/source/usage/workflows/addLaser.rst
@@ -3,7 +3,7 @@
 Adding Laser
 ------------
 
-.. sectionauthor:: Sergei Bastrakov
+.. sectionauthor:: Sergei Bastrakov, Fabia Dietrich
 
 There are several alternative ways of adding an incoming laser (or any source of electromagnetic field) to a PIConGPU simulation:
 
@@ -12,6 +12,8 @@ There are several alternative ways of adding an incoming laser (or any source of
 
 These ways operate independently of one another, each has its features and limitations.
 Incident field- and laser profiles currently match one another.
+
+Please note the sign convention for plane waves propagating forward in time: :math:`e^{-i(kz + \omega t)}`
 
 Incident Field
 """"""""""""""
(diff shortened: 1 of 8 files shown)


BrianMarre:
@fabidie As a pointer for future commits, please use the CI-control options documented [here](https://picongpu.readthedocs.io/en/latest/dev/ci.html) to avoid running pmacc compile tests even though you're PR does not contain any changes of pmacc.
Not the end of the world if you miss to add the flag but try to use it in the future ;).

psychocoderHPC (approved):

PR #5153, opened 2024-10-02
Fix atomic physics example docs

ikbuibui:
`validateLocally` already includes a call to execute PIConGPU, fix to not run twice
Fixes the pic-build call in validate script


BrianMarre:
good catch @ikbuibui, please also fix the `pic-build` in `validateLocally`

chillenzer (approved):

PR #5147, opened 2024-09-30
PIConGPU: add fields and particles to header check

psychocoderHPC:
- add `fields` and `particle` header check
- fix missing includes and resolve cyclic includes

ci: picongpu

I tried to move as less as possible code but sometimes it was required to resolve cyclic includes.
I also removed two unused files. They were not included somewhere but broke the header test.

This PR is reducing the compile memory footprint to 78 GiB for KHI with nvcc 12.3

PR #5155, opened 2024-10-02
fix broken dev branch

psychocoderHPC:
- fix missing include
- fix unused variable errors due to constexpr if's

ci: full-compile


ikbuibui:
Could you maybe include this change https://github.com/ComputationalRadiationPhysics/picongpu/pull/5147#discussion_r1784134264  into the PR as well?

include/picongpu/particles/atomicPhysics/kernel/FillLocalRateCache_BoundBound.kernel
```diff
@@ -154,7 +154,8 @@ namespace picongpu::particles::atomicPhysics::kernel
 
             auto forEachAtomicStateAndBin
                 = pmacc::lockstep::makeForEach<T_numberAtomicStates * T_numberBins, T_Worker>(worker);
-            auto forEachAtomicState = pmacc::lockstep::makeForEach<T_numberAtomicStates, T_Worker>(worker);
+            [[maybe_unused]] auto forEachAtomicState
+                = pmacc::lockstep::makeForEach<T_numberAtomicStates, T_Worker>(worker);
```
chillenzer:
```suggestion
            auto forEachAtomicState
                = pmacc::lockstep::makeForEach<T_numberAtomicStates, T_Worker>(worker);
```
and move it down to where it's actually used. As far as I can see that's a single place. And maybe do the same for the `forEachAtomicStateAndBin` above.

chillenzer (dismissed):
Not blocking.

include/picongpu/particles/atomicPhysics/kernel/FillLocalRateCache_BoundBound.kernel
psychocoderHPC: it is used twice in two `if constexpr` branches but would like avoid creating the for each twice.

psychocoderHPC:
> Could you maybe include this change [#5147 (comment)](https://github.com/ComputationalRadiationPhysics/picongpu/pull/5147#discussion_r1784134264) into the PR as well?

No, I try to avoid mixing it. There is a high change that the header checks will fail if I remove it. Most likely there is a reason why the include was added there. If there is an include too much we should remove it in a seperat PR. It could also be that this is indirect fixing a missing include somewhere.

include/picongpu/particles/atomicPhysics/kernel/FillLocalRateCache_BoundBound.kernel
chillenzer: Doing "Click `...`" -> "View File" -> "Ctrl-F" -> "`forEachAtomicState`" -> "Enter" yields exactly four results for me: One definition and one usage per variable. Are you aware that these are two different variables?

psychocoderHPC: I missed it, the names are so equal that I oversaw it. 

fixed it

include/picongpu/particles/atomicPhysics/kernel/FillLocalRateCache_BoundBound.kernel
```diff
@@ -154,7 +154,6 @@ namespace picongpu::particles::atomicPhysics::kernel
 
             auto forEachAtomicStateAndBin
                 = pmacc::lockstep::makeForEach<T_numberAtomicStates * T_numberBins, T_Worker>(worker);
```
chillenzer, line 156: Sorry for being pedantic but the same applies to this one. It should also move into the `if constexpr`.

chillenzer (changes requested):

include/picongpu/particles/atomicPhysics/kernel/FillLocalRateCache_BoundBound.kernel
psychocoderHPC: Yes, you are right possible to change in a separate PR. I only react on broken tests in the CI I have not done a code refactoring.

chillenzer: Okay.

chillenzer (approved):

PR #5157, opened 2024-10-07
fix value identifier implementation

psychocoderHPC:
With #5130 we introduced two bugs.
- an alias is used to call methods of `value_identifier`
- `value_identifier` accessed without checking the attribute of the source particle to derive the value


psychocoderHPC:
I merge this PR with my super admin right to fix the current broken dev branch.

PR #5162, opened 2024-10-08
fix wrong unit of AtomicPhysics(FLYonPIC) ADK field ionization rate

BrianMarre:
Bug fix, fixes a wrong unit conversion in the ADK field ionization rate calculation in AtomicPhysics(FLYonPIC).

ci: picongpu


chillenzer (approved):
Cant' say much about what the correct units would be but from the technical side it's fine.

psychocoderHPC (approved):

PR #5164, opened 2024-10-08
fix maximum ADK rate estimation

BrianMarre:
Bug fix, AtomicPhysics(FLYonPIC) mistakenly assumed that the highes ADK rate is reached with for the highest EField.

- [x] requires PR #5163 to be merged first
- [x] requires PR #5162 to be merged first
- [x] requires PR #5159 to be merged first
- [x] needs to be rebased upon dev

ci: picongpu

PR #5165, opened 2024-10-09
Instant transition for states with high field ionization rate.

BrianMarre:
Field ionization rates in AtomicPhysics(FLYonPIC) do not reduce with consecutive sub-steps since we do not update the local field yet.
These high rates require a large number o atomicPhyiscs sub steps to resolve and therefore lead to a slow down of the simulation.
To avoid this this PR adds a new sub-stage checking for states with time step lengths above a user defined limit and instantly applies field ionization transitions to ions in a state found to have a field ionization loss rate above the limit 

- [x] requires PR #5164 to be merged first
- [ ] needs to be rebased to dev

This PR will break existing setups using a custom `atomicPhysics.param`-file. To fix a setup add the following lines inside the `RateSolverParam` -struct
```c++
        // maximum number of atomicPhysics sub-steps per PIC time step for fieldIonization
        static constexpr uint32_t maximumNumberSubStepsFieldIonization = 2000;
```

### DIFF
--- a/include/picongpu/param/atomicPhysics.param
+++ b/include/picongpu/param/atomicPhysics.param
@@ -55,6 +55,9 @@ namespace picongpu::atomicPhysics
         // which probability approximation to use for the acceptance step
         using ProbabilityApproximationFunctor
             = picongpu::particles::atomicPhysics::ExponentialApproximationProbability;
+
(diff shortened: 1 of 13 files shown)


BrianMarre:
replaced by PR #5218

PR #5166, opened 2024-10-10
Bugfix in TWEAC-FOM when switching to the incidentFields-method

BeyondEspresso:
The PR fixes a bug in the TWEAC-FOM benchmark for the incidentFields configuration. The bug was not caught before, because the recent changes to the `simulation_defines.hpp` include structure has caused an include to go missing in the benchmark. As a consequence, initial tests of the the TWEAC-FOM benchmark using the incidentFields configuration led to "empty" functors. The then resulting benchmark was faster than anticipated.

### DIFF
--- a/share/picongpu/benchmarks/TWEAC-FOM/include/picongpu/param/incidentField.param
+++ b/share/picongpu/benchmarks/TWEAC-FOM/include/picongpu/param/incidentField.param
@@ -196,9 +196,8 @@ namespace picongpu
                 template<uint32_t T_component>
                 HDINLINE float_X getComponent(floatD_X const& totalCellIdx) const
                 {
-                    float3_64 const invUnitField
-                        = float3_64(1.0 / m_unitField[0], 1.0 / m_unitField[1], 1.0 / m_unitField[2]);
-                    float3_X const amplitude = precisionCast<float_X>(Params::AMPLITUDE_SI * invUnitField);
+                    float_64 const invUnitField = 1.0 / m_unitField[T_component];
+                    float_X const amplitude = precisionCast<float_X>(Params::AMPLITUDE_SI * invUnitField);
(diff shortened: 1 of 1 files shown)


share/picongpu/benchmarks/TWEAC-FOM/include/picongpu/param/incidentField.param
```diff
@@ -56,6 +56,7 @@
 #pragma once
 
 #include "picongpu/fields/background/templates/twtstight/twtstight.hpp"
+#include "picongpu/fields/background/templates/twtstight/twtstight.tpp"
```
psychocoderHPC: This will be fixed in https://github.com/ComputationalRadiationPhysics/picongpu/pull/5155
If you include this in param you create cyclic includes. If this change would solve the problem we could simply include the top in the hpp. This is not the case here.

psychocoderHPC (dismissed):
after https://github.com/ComputationalRadiationPhysics/picongpu/pull/5155 is merge the include should be removed. The rest looks good.

psychocoderHPC:
@BeyondEspresso I stopped the CI because we have currently nearly no CI resources available. As soon as the PR is rebased we can run this PR in the CI.
I will notify you as soon as we merged #5155

chillenzer:
@BeyondEspresso @psychocoderHPC The time has come!

chillenzer (dismissed):

share/picongpu/benchmarks/TWEAC-FOM/include/picongpu/param/incidentField.param
BeyondEspresso: Done.

BeyondEspresso:
@psychocoderHPC @chillenzer I just removed the include and rebased the PR to latest dev.

chillenzer (approved):

PR #5167, opened 2024-10-11
PIConGPU: complete header check

psychocoderHPC:
- add all of PIConGPU to the header check
- fix missing includes and resolve cyclic includes

note: The header check is running without optional dependencies e.g openPMD, fftw3 or pngwriter. This will be addressed after the release.


include/picongpu/plugins/binning/UnitConversion.hpp
```diff
@@ -19,18 +19,25 @@
...
 
 namespace picongpu
 {
     namespace plugins::binning
     {
         constexpr unsigned numUnits = 7;
+
         // @todo add where this 7D format is from
-        std::array<double, numUnits> UnitDimensions{
+        static std::array<double, numUnits> UnitDimensions{
```
chillenzer, line 40: Why has this become static?

include/picongpu/plugins/binning/WriteHist.hpp
```diff
@@ -17,21 +17,19 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
-#if !ENABLE_OPENPMD
-#    error The activated Binning plugin requires openPMD-api.
-#endif
+#if(ENABLE_OPENPMD == 1)
```
chillenzer, line 21: Wouldn't this make the error message worse?

include/picongpu/plugins/openPMD/openPMDWriter.def
```diff
@@ -147,8 +160,6 @@ namespace picongpu
 #endif
             };
 
-            ::openPMD::Series& openSeries(::openPMD::Access at);
-
             void closeSeries();
```
chillenzer, line 163: You have probably shifted `openSeries` but not `closeSeries` because the header checks don't fail for the latter (as there is no reference to openPMD types in there). I'd still suggest that we also shift `closeSeries` because it doesn't make sense if we can't have an open series.

include/picongpu/plugins/radiation/param.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#pragma once
+
+#include "picongpu/defines.hpp"
+
+// clang-format off
+#    include "picongpu/param/radiation.param"
+#    include "picongpu/param/radiationObserver.param"
+#    include "picongpu/unitless/radiation.unitless"
+// clang-format on
```
chillenzer:
```suggestion
#include "picongpu/param/radiation.param"
#include "picongpu/param/radiationObserver.param"
#include "picongpu/unitless/radiation.unitless"
```
I'm aware that this a a copy-paste thing from another file but in doing so, this tiny formatting fix should be allowed.

include/picongpu/plugins/transitionRadiation/param.hpp
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+#pragma once
+
+#include "picongpu/defines.hpp"
+
+// clang-format off
+#include "picongpu/param/transitionRadiation.param"
+#    include "picongpu/unitless/transitionRadiation.unitless"
+// clang-format on
```
chillenzer: Same here.

test/picongpuHeaderCheck/CMakeLists.txt
```diff
@@ -76,18 +76,7 @@ add_subdirectory(${PIC_ROOT_DIR}/../../thirdParty/catch2 ${CMAKE_BINARY_DIR}/cat
...
-append_recursive_files("${PIC_SUFFIXED_INCLUDE_DIR}/traits" "hpp" "PIC_FILES_HEADER")
-append_recursive_files("${PIC_SUFFIXED_INCLUDE_DIR}/fields" "hpp" "PIC_FILES_HEADER")
-append_recursive_files("${PIC_SUFFIXED_INCLUDE_DIR}/particles" "hpp" "PIC_FILES_HEADER")
-
-# remove traits which where include dependencies can not be solved easily
-file(GLOB_RECURSE PIC_ATTRIBUTE_TRAITS "${PIC_SUFFIXED_INCLUDE_DIR}/traits/attribute/*.hpp")
-file(GLOB_RECURSE PIC_FRAME_TRAITS "${PIC_SUFFIXED_INCLUDE_DIR}/traits/frame/*.hpp")
-list(REMOVE_ITEM PIC_FILES_HEADER ${PIC_ATTRIBUTE_TRAITS})
-list(REMOVE_ITEM PIC_FILES_HEADER ${PIC_FRAME_TRAITS})
+append_recursive_files("${PIC_SUFFIXED_INCLUDE_DIR}/." "hpp" "PIC_FILES_HEADER")
```
chillenzer, line 80: Isn't it beautiful how it all falls in place?

chillenzer (dismissed):
Nothing blocking. Can be merged once the CI is happy.

include/picongpu/plugins/binning/UnitConversion.hpp
psychocoderHPC: double definition in multiple openmp units

include/picongpu/plugins/binning/WriteHist.hpp
psychocoderHPC: The behavior changed. Need now that each header can be included independent of which requirements it has. 
Therefore it is not possible anymore to show this error.

include/picongpu/plugins/openPMD/openPMDWriter.def
psychocoderHPC: You are right. I only worked on the needs to solve header issues and not touch stuff which is not required to be touched. The reason is that this easily starts a wave of changes.

include/picongpu/plugins/radiation/param.hpp
psychocoderHPC: fixed

chillenzer:
You made it pass! =)

chillenzer (approved):

PR #5174, opened 2024-10-17
Generalize twtstight for arbitrary beta0 and code refactoring

BeyondEspresso:
The symmetry scales dY and dT are used to maintain longterm numerical stability by exploiting a spatio-temporal symmetry in the twtstight laser pulse. There has been a bug in dY and dT. Since their equations were unnecessarily convoluted, I also refactored them in the fix. The equations now pass the dY-dT-symmetry test.


chillenzer:
Without having had much more than a quick glance on it, would it be an option to add that test you were talking about as a unit/end-to-end test?

BeyondEspresso:
Yes, that should be quite simple.

chillenzer:
Great, thanks!

BeyondEspresso:
@chillenzer @steindev The small PR has grown quite a bit. After applying the original updates I noticed that the twtstight-implementation does not work for `beta0 != 1.0`, which is not the normal usecase, but it is in the interface and actually nice to have. I updated the laser equations to now also include beta0. I tested the new implementation that compared to my model implementation correctly generates the old (beta0=1.0) and now also new results (beta0 != 1.0). While applying the changes, I seized the opportunity for some reductions in boiler plate code that still existed from previous, now outdated, twts implementations.

BeyondEspresso:
@chillenzer Please leave the draft-tag for now until Monday. It is late and it is a good idea to have some time pass,

BeyondEspresso:
@chillenzer I actually have further minor changes to make the PR more round. I will submit them soon.

BeyondEspresso:
@chillenzer @steindev  OK, I rebased my PR and completed my generalization and refactoring. Additionally, I successfully tested the code in an actual PIConGPU run. The draft-tag can be removed now.

include/picongpu/fields/background/templates/twtstight/BField.hpp
```diff
@@ -38,7 +38,7 @@ namespace picongpu
             class BField
             {
             public:
-                using float_T = float_64;
+                using float_T = float_X;
```
chillenzer: Was there a reason to have it 64bit before or was this just an oversight in the first place?

chillenzer: Okay. Got it from the comments below. Thanks for the clear explanation! Very good!

share/picongpu/benchmarks/TWEAC-FOM/include/picongpu/param/incidentField.param
```diff
@@ -262,11 +262,29 @@ namespace picongpu
...
+             *
+             * Except moving window simulations, the position must be inside the global domain.
+             * The distance between the Huygens surface and each global domain boundary must be at least
+             * absorber_thickness + (FDTD_spatial_order / 2 - 1). However beware of setting position = direction *
+             * (absorber_thickness + const), as then changing absorber parameters will affect laser positioning.
+             * When all used profiles are None, the check for POSITION validity is skipped.
+             *
+             * For moving window simulations, POSITION for the YMax side can be located outside the initially
+             * simulated volume. In this case, parts of the generation surface outside of the currently simulated
+             * volume is are treated as if they had zero incident field and it is user's responsibility to apply a
```
chillenzer:
```suggestion
             * volume are treated as if they had zero incident field and it is user's responsibility to apply a
```

include/picongpu/fields/background/templates/twtstight/BField.tpp
```diff
@@ -205,29 +205,13 @@ namespace picongpu
...
-
-                /* Angle between the laser pulse front and the y-axis. Not used, but remains in code for
-                 * documentation purposes.
-                 * float_T const eta = float_T(PI/2) - (phiReal - alphaTilt);
+                 * y-axis of the coordinate system in this function.
                  */
+                auto const phiT = float_T(math::abs(phi));
+                float_T sinPhi;
+                float_T cosPhi;
+                pmacc::math::sincos(phiT, sinPhi, cosPhi);
```
chillenzer: Sorry for my ignorance but could you please quickly explain if `math` and `pmacc::math` are different namespaces. Also, given that those are frequently used, a `using` directive might be beneficial for this function.

include/picongpu/fields/background/templates/twtstight/BField.tpp
```diff
@@ -205,29 +205,13 @@ namespace picongpu
                 auto const beta0 = float_T(beta_0);
```
chillenzer: Is there a particular benefit in using your own float here? I see it is currently `float_X` and was previously `float_64`. Is this a remainder that you want to keep?

include/picongpu/fields/background/templates/twtstight/BField.tpp
```diff
@@ -262,15 +242,12 @@ namespace picongpu
                 /* To avoid underflows in computation, fields are set to zero
                  * before and after the respective TWTS pulse envelope.
                  */
-                if(math::abs(y - z * math::tan(phiT / float_T(2.0)) - (cspeed * t)) > (numSigmas * tauG * cspeed))
+                if(math::abs(y - z * tanAlpha - (beta0 * cspeed * t)) > (numSigmas * tauG * cspeed))
                     return float_T(0.0);
 
                 /* Calculating shortcuts for speeding up field calculation */
```
chillenzer: I think you're perfectly fine to assume that the compiler would be capable of figuring this out for itself and will happily ignore if these are present or not, using its own heuristics instead. The only benefit of these could be simplifying formulae later on. You can decide for yourself if for any or all of these this purpose is fulfilled.

include/picongpu/fields/background/templates/twtstight/BField.tpp
chillenzer, line 1: I'm not quite happy with the three functions being mostly duplications of one another with a few switches of sin and cos. It is well annotated and I could follow along quite well but it would be a huge pain to change every definition three times (as I believe you have experienced during this refactoring). Could we make this more elegant? One option could be to extract some common methods for computing the basics at the top, potentially using structured bindings or a shallow struct for the return value. Alternatively, one could think about merging the functions computing and returning all three components (if they are not needed separately).

include/picongpu/fields/background/templates/twtstight/EField.tpp
chillenzer, line 1: Skimming through this, the same comments apply as for the B field. Even more so, I think we should extract common functions, so we don't have a 6-fold code duplication of round about 70 lines.

chillenzer (changes requested):
Very well annotated, thanks for the great guidance. Still there are a few things to discuss. Also, I can't say anything about the actual physics formula. Needs a check from @steindev .

include/picongpu/fields/background/templates/twtstight/BField.tpp
BeyondEspresso: Was this comment meant for a different line? Without this if-statement, the code produces NaNs due to underflow. We already were severely bitten by this. The function of this statement corresponds to the "laser window" of a standard Gaussian laser.

include/picongpu/fields/background/templates/twtstight/BField.hpp
BeyondEspresso: It is not finally decided, whether single precision is good enough for all instances and parameters of twtstight. Therefore, we need a switch to change the precision of twtstight. However, even in if float_T = float_X, there are a number of variables that need to stay at double-precision. Those variables are always declared as float_64.

share/picongpu/benchmarks/TWEAC-FOM/include/picongpu/param/incidentField.param
BeyondEspresso: I'll change this.

include/picongpu/fields/background/templates/twtstight/BField.tpp
BeyondEspresso: During the development of this code, namespaces have changed. I guess that this is a remainder that can be removed now. I'll check.

include/picongpu/fields/background/templates/twtstight/BField.tpp
BeyondEspresso: I would like to keep this, as I need to be able to switch between all float_64 computations of twtstight and the faster float_X /float_X hybrid implementation.

include/picongpu/fields/background/templates/twtstight/BField.tpp
BeyondEspresso: I fully agree, as I also spend some thought on it. However, I always postponed it until later. Let us discuss the most suitable approach offline (or in a VC call).

include/picongpu/fields/background/templates/twtstight/EField.tpp
BeyondEspresso: See above. Let us discuss this before I make changes.

include/picongpu/fields/background/templates/twtstight/BField.tpp
chillenzer: Sorry, always forget how badly GH renders comments on a single line. This was supposed to be a comment on "Calculating shortcuts for speeding up field calculation".

chillenzer:
@BrianMarre should probably follow the developments here after we've done a similar refactoring with his code.

chillenzer:
> Without having had much more than a quick glance on it, would it be an option to add that test you were talking about as a unit/end-to-end test?

Have you actually added said test yet or is this still on your todo list?

BeyondEspresso:
@chillenzer Yes, adding the test is still on my todo list.

include/picongpu/fields/background/templates/twtstight/BField.tpp
BeyondEspresso: @chillenzer @BrianMarre I just successfully compiled a small proof of concept for the structured bindings within twtstight using 3 variables on both NVIDIA and AMD GPUs.

BeyondEspresso:
@psychocoderHPC @chillenzer Ok, here is the refactored version. To reduce duplicate code I moved the implementation into the new template class TWTSTight and specialized for the respective E- and B-Field implementation. In the math, I also did a separation of complex and float-numbers via reordering and brackets according to @psychocoderHPC suggestions. Due to the new struct, I renamed twtstight.tpp to TWTSTight.tpp. In total this results in a code reduction by almost 600 LOCs.

BeyondEspresso:
@chillenzer Before you review the code, I need to review it myself, as there are changes that were not intended.

BeyondEspresso:
@chillenzer Should be OK now.  In addition to updating attributions and fixing typos, there was a missing `"`, which failed the CI.

chillenzer:
Nice! With this refactoring, the PR actually reduces the lines of code in the repo while still adding new functionality if I understood correctly.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
```diff
@@ -0,0 +1,366 @@
...
+                 */
+                auto const absPhi = float_T(math::abs(phi));
+
+                /* Propagation speed of overlap normalized to the speed of light [Default: beta0=1.0] */
+                auto const beta0 = float_T(beta_0);
+
+                float_T sinPhi;
+                float_T cosPhi;
+                math::sincos(absPhi, sinPhi, cosPhi);
+                float_T const tanAlpha = (float_T(1.0) - beta0 * cosPhi) / (beta0 * sinPhi);
```
chillenzer: This trigonometry part is independent of the lower physics-constants part. Split this function into two. Potentially, merge the trigonometry part with `defineTrigonometryShortcuts`.

If you prefer, you can still have a single entry point `defineBasicHelperVariables()` but that should consist of just three lines then: Call trigonometry part, call physics-constants part, return some form of merger of this result.

BeyondEspresso:
That is correct.

include/picongpu/fields/background/templates/twtstight/BField.tpp
```diff
@@ -41,273 +36,104 @@ namespace picongpu
...
-            float3_X BField::operator()(DataSpace<simDim> const& cellIdx, uint32_t const currentStep) const
-            {
-                traits::FieldPosition<fields::YeeCell, FieldB> const fieldPosB;
-                return getValue(precisionCast<float_X>(cellIdx), fieldPosB(), static_cast<float_X>(currentStep));
-            }
+                /* To avoid underflows in computation, fields are set to zero
+                 * before and after the respective TWTS pulse envelope.
+                 */
+                if(math::abs(y - z * tanAlpha - (beta0 * cspeed * t)) > (numSigmas * tauG * cspeed))
+                    return float_T(0.0);
```
chillenzer:
```suggestion
                if(outsideOfEnvelope(...))
```
Extract function.

include/picongpu/fields/background/templates/twtstight/BField.tpp
```diff
@@ -41,273 +36,104 @@ namespace picongpu
...
-                    // we should never be here
-                    return NAN;
-                }
-                if constexpr(simDim != DIM3)
-                    return (*this)(cellIdx, currentStep)[T_component];
+                /* A 180deg-rotation of the field vector around the y-axis
+                 * leads to a sign flip in the x- and z- components, respectively.
+                 * This is implemented by multiplying the result by "phiPositive".
+                 */
+                return phiPositive * result.real() / sim.unit.speed();
```
chillenzer: If you're only returning the real part here, what are we computing with complex numbers for? Couldn't you extract the real part right from the start to save computation? (Not checked, just a guess.)

include/picongpu/fields/background/templates/twtstight/BField.tpp
```diff
@@ -41,273 +36,104 @@ namespace picongpu
...
+                                     * (rhom2 + rhom2 * cosPhi_2 - x2 * sinPhi_2 + x * cosPhi * sinPhi_2 * Xm))
+                          + besselJ1const
+                              * (cosPolAngle * sinPhi
+                                     * (rhom2 - float_T(2.0) * x2 + I * Xm * rhom2 * (k * sinPhi)
+                                        + x * cosPhi * (float_T(-2.0) * Xm - I * rhom2 * (k * sinPhi)))
+                                 + sinPolAngle
+                                     * ((rhom2 - float_T(2.0) * x2) * sinPhi
+                                        + I * rhom2 * (-Xm + x * cosPhi) * (k * sinPhi_2) + x * sin2Phi * Xm)))
+                       * psi0)
+                    / (cspeed * besselI0const * rhom * rhom2);
```
chillenzer: Extract function. Maybe let that function even take a single tuple and extract inside. Not sure what feels more ergonomic to read.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
```diff
@@ -0,0 +1,366 @@
...
+                    / (math::sqrt(float_T(2.0))
+                       * math::exp(
+                           beta02 * omega0 * math::cPow(t - nu + xi, static_cast<uint32_t>(2u))
+                           / (beta02 * omega0 * tauG2 - complex_T(0, 2) * (beta02 * (nu - xi) * cotPhi * cotPhi)
+                              + complex_T(0, 2) * (beta0 * (float_T(2.0) * nu - xi) * cotPhi / sinPhi)
+                              - complex_T(0, 2) * (nu / sinPhi_2)))
+                       * math::sqrt(
+                           ((beta02 * omega0 * tauG2) / float_T(2.0) - I * (beta02 * (nu - xi) * cotPhi * cotPhi)
+                            + I * (beta0 * (float_T(2.0) * nu - xi) * cotPhi / sinPhi) - I * (nu / sinPhi_2))
+                           / omega0));
```
chillenzer: I would probably extract this into a separate function and call this separately for semantic reasons.

BeyondEspresso:
@chillenzer Now I also added the unit test to the PR.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
```diff
@@ -0,0 +1,366 @@
...
+#include <pmacc/types.hpp>
+
+namespace picongpu
+{
+    /* Load pre-defined background field */
+    namespace templates
+    {
+        /* Traveling-wave Thomson scattering laser pulse */
+        namespace twtstight
+        {
```
chillenzer:
```suggestion
namespace picongpu::templates::twtstight {
```

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
```diff
@@ -0,0 +1,366 @@
...
+                if constexpr(simDim == DIM2)
+                {
+                    for(uint32_t k = 0; k < detail::numComponents; ++k)
+                    {
+                        /* 2D (y,z) vectors are mapped on 3D (x,y,z) vectors. */
+                        for(uint32_t i = 0; i < simDim; ++i)
+                        {
+                            pos[k][i + 1] = fieldPositions_SI[k][i];
+                        }
+                    }
```
chillenzer: Didn't we want to do the `simDim-i` here?

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
```diff
@@ -0,0 +1,366 @@
...
+            template<typename T_Field>
+            HDINLINE float3_X
+            TWTSTight<T_Field>::operator()(DataSpace<simDim> const& cellIdx, uint32_t const currentStep) const
+            {
+                traits::FieldPosition<fields::YeeCell, T_Field> const fieldPos;
+                return getValue(precisionCast<float_X>(cellIdx), fieldPos(), static_cast<float_X>(currentStep));
+            }
+
+            template<typename T_Field>
+            HDINLINE float3_X TWTSTight<T_Field>::operator()(floatD_X const& cellIdx, float_X const currentStep) const
```
chillenzer: It's kind of surprising to me that also `currentStep` became float here. Is that intentional and if so, why?

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
```diff
@@ -0,0 +1,366 @@
...
+                using PosVecVec = pmacc::math::Vector<float3_64, detail::numComponents>;
+                PosVecVec pos(PosVecVec::create(float3_64::create(0.0)));
+
+                if constexpr(simDim == DIM3)
+                {
+                    for(uint32_t k = 0; k < detail::numComponents; ++k)
+                    {
+                        for(uint32_t i = 0; i < simDim; ++i)
+                        {
+                            pos[k][i] = fieldPositions_SI[k][i];
```
chillenzer: Actually: Why do we have to copy the `fieldPositions_SI` before handing them over? Is that only for having it in a 3D vector afterwards? Why doesn't `defineMinimalCoordinates` handle this case?

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
```diff
@@ -0,0 +1,366 @@
...
+                float_T,
+                float_T,
+                float_T,
+                float_T,
+                float_T,
+                float_T,
+                float_T,
+                float_T,
+                float_T>
+            TWTSTight<T_Field>::defineBasicHelperVariables() const
```
chillenzer: This function does not take any arguments and all it refers to are constant members of this class. Is there a reason why this is not called simply once in the constructor and initialises all the values as data members?

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
```diff
@@ -0,0 +1,366 @@
...
+                    pmacc::math::Vector<floatD_X, detail::numComponents> zeroShifts;
+                    for(uint32_t component = 0; component < detail::numComponents; ++component)
+                        zeroShifts[component] = floatD_X::create(0.0);
+                    pmacc::math::Vector<floatD_64, detail::numComponents> const fieldPositions_SI
+                        = detail::getFieldPositions_SI(cellIdx, halfSimSize, zeroShifts, unit_length, focus_y_SI, phi);
+                    // Explicitly use a 3D vector so that this function compiles for 2D as well
+                    auto const pos = float3_64{
+                        fieldPositions_SI[T_component][0],
+                        fieldPositions_SI[T_component][1],
+                        fieldPositions_SI[T_component][2]};
```
chillenzer: Same question: Why do we copy vectors around here?

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
```diff
@@ -0,0 +1,366 @@
...
+                    else
+                    {
+                        if constexpr(T_component == 1)
+                        {
+                            return static_cast<float_X>(calcTWTSFieldY(pos, time_SI));
+                        }
+                        if constexpr(T_component == 2)
+                        {
+                            return static_cast<float_X>(calcTWTSFieldZ(pos, time_SI));
+                        }
```
chillenzer: Couldn't we make `calcTWTSField<T_Component>` a template? This code would be much easier.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
```diff
@@ -0,0 +1,366 @@
...
+                    }
+
+                    // we should never be here
+                    return NAN;
+                }
+                if constexpr(simDim != DIM3)
+                    return (*this)(cellIdx, currentStep)[T_component];
```
chillenzer:
```suggestion
                else if constexpr(simDim == DIM2)
                    return (*this)(cellIdx, currentStep)[T_component];
                else
                    return floatD_X{NAN}; // or whatever the correct syntax for this is...
```
Should we be explicit here?

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
```diff
@@ -0,0 +1,366 @@
...
+            {
+                /* In order to calculate in single-precision and in order to account for errors in
+                 * the approximations far from the coordinate origin, we use the wavelength-periodicity and
+                 * the known propagation direction for realizing the laser pulse using relative coordinates
+                 * (i.e. from a finite coordinate range) only. All these quantities have to be calculated
+                 * in double precision.
+                 */
+                float_64 const deltaT = wavelength_SI / sim.si.getSpeedOfLight()
+                    / (1.0 - beta_0 * math::cos(precisionCast<float_64>(phi)));
+                float_64 const deltaY = beta_0 * sim.si.getSpeedOfLight() * deltaT;
```
chillenzer: These do not depend on the arguments and could be data members of the class.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
```diff
@@ -0,0 +1,366 @@
...
+            {
+                /* Calculating shortcuts for speeding up field calculation */
+                float_T const tanPhi = math::tan(absPhi);
+                float_T const cotPhi = float_T(1.0) / tanPhi;
+                float_T const sinPhi_2 = sinPhi * sinPhi;
+                float_T const sinPolAngle = math::sin(polAngle);
+                float_T const cosPolAngle = math::cos(polAngle);
+
+                return std::make_tuple(tanPhi, cotPhi, sinPhi_2, sinPolAngle, cosPolAngle);
+            }
```
chillenzer: Does not depend on non-const data. -> Data members.

include/picongpu/fields/background/templates/twtstight/TWTSTight.hpp
```diff
@@ -29,17 +64,28 @@
...
+             *  @tparam T_dim Specializes for the simulation dimension
+             */
+            template<typename T_Field>
+            struct TWTSTight
             {
             public:
-                using float_T = float_64;
-
                 /** Center of simulation volume in number of cells */
                 PMACC_ALIGN(halfSimSize, DataSpace<simDim>);
```
chillenzer: Is there a reason why this is non-const? Same for `phiPositive` and `tdelay`.

share/picongpu/unit/TWTSTight.cpp
```diff
@@ -0,0 +1,225 @@
...
+static bool isApproxEqual(T const& a, T const& b)
+{
+    return a == Catch::Approx(b).margin(std::numeric_limits<T>::epsilon());
+}
+
+template<typename T>
+static bool isApproxEqual(T const& a, T const& b, T const& epsilon)
+{
+    return a == Catch::Approx(b).margin(epsilon);
+}
```
chillenzer:
```suggestion
template<typename T>
static bool isApproxEqual(T const& a, T const& b, T const& epsilon=std::numeric_limits<T>::epsilon())
{
    return a == Catch::Approx(b).margin(epsilon);
}
```

share/picongpu/unit/TWTSTight.cpp
```diff
@@ -0,0 +1,225 @@
...
+        const float_64 Bz = float_64(-6.090365721598194e-10);
+        const float_T epsilon = float_T(5.0e-6);
+        const float_T epsilon2 = float_T(5.0e-15);
+        bool isCorrect = isApproxEqual(precisionCast<float_T>(Ex), res[0], epsilon)
+            && isApproxEqual(testEfield.calcTWTSFieldX(pos, time), res[0], epsilon2);
+        std::cerr << std::setprecision(std::numeric_limits<float_T>::digits10);
+        if(!isCorrect)
+            std::cerr << "Ex (analytical reference):   " << Ex << "  , EField.calcTWTSFieldX (Device):   " << res[0]
+                      << "  , EField.calcTWTSFieldX (Host):   " << testEfield.calcTWTSFieldX(pos, time) << std::endl;
+        REQUIRE(isCorrect);
```
chillenzer: A few things about this:
- I would prefer to use `CHECK` over `REQUIRE`. The latter will stop your test case which means you're holding back on important information like "Oh, the x-direction failed but the y-direction passes. That's a hint!". (`REQUIRE` is more intended for pre-conditions the failure of which invalidates the whole test case.)
- I would prefer to have the two conditions per dimension and field in separate `CHECK` blocks immediately. If you do that, Catch2 is already smart enough to show you the values before the check and the equality that failed. It might be that Catch2 is not smart enough to see through the encapsulation inside a function. If so, I'd prefer to have that check inline.
- Afterwards, you can safely remove your debug statements because that information should be provided by Catch2 automatically as said above.
- After the above are taken care of, we can make E and B vectors as use a `for`-loop here. Something along the lines of
```C++
auto hostE = testEfield.getTWTSField_Normalized(...);
for (int i...) {
    CHECK(isApproxEqual(precisionCast<float_T>(E[i]), res[i], epsilon));
    CHECK(isApproxEqual(hostE[i], res[i], epsilon2));
    // ... same for B
}
```

share/picongpu/unit/TWTSTight.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+        resultHost.copyFrom(resultDevice);
+
+        auto res = resultHost.getDataBox();
+        const float_64 Ex = float_64(0.18329124052693974);
+        const float_64 Ey = float_64(-0.009402050968104002);
+        const float_64 Ez = float_64(0.1054028749666347);
+        const float_64 Bx = float_64(3.5299706879027803e-10);
+        const float_64 By = float_64(5.334111474127282e-11);
+        const float_64 Bz = float_64(-6.090365721598194e-10);
```
chillenzer: Make these vectors.

chillenzer (changes requested):
Hi,

thanks for the good work! Thanks for adding the test. Also: Very good documentation. It was really pleasant to immediately know where to look for an answer. Finally, this is already way better readable and a reduction by 600 lines is impressive! Now that we've got the first layer of clutter off, we start to see the structure here.

I might have overlooked something, so please correct me if I'm wrong here. Also, I have only considered the Bfield version of the file because the Efield is more or less the same up to those nasty details you can get wrong particularly easily.

1. A large part of your helper variables doesn't seem to depend on the function arguments at all. Is there a reason to not initialise them in the constructor and use the as data members directly?
2. I would prefer to pass around large argument lists as tuples and only extract them where the individual names are actually needed, e.g.
```
auto basicHelpers = defineBasicHelperVariables();
// ...
auto commonHelpers = defineCommonHelperVariables(basicHelpers, ...);
// ...
auto [absPhi, sinPhi, cosPhi, ...] = basicHelpers;
// do actual computation where we need the names individually
```
3. Also, with such a huge number of return values it might be reasonable to think about actually defining a struct to carry them around and refer to them by name. Too easy to swap two names here I think.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
BeyondEspresso: We only need it for having 2D (y,z) vectors in a 3D (0,y,z) vector afterwards. I now simplified the DIM3 case without the copying over. Otherwise I think the place is correct here.  `defineMinimalCoordinates` is for exploiting the y-t-periodicity. Also I prefer to keep the interface of the field functions as clean 3D vectors (instead of floatD_64) for better separations of concerns.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
BeyondEspresso: Using the general statement `pos[k][i + 3- simdim] = ...` would make the code shorter. I guess the question is whether compilers will (correctly) optimize away the unecessary copying over for the 3D case.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
BeyondEspresso: It is the required interface for incidentFields, which was defined in a general way, such that it can be extended to a whole range of possible field solvers. Similar as fractional cells it demands access to sub-timesteps.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
BeyondEspresso: Because the compiler would complain in 2D cases if we directly put fieldPositions_SI into the calcTWTSField-functions. In such a 2D case, the compiler nevertheless tries to process the (simDim == DIM3) even when its not fulfilled, finds a type mismatch between floatD and float3 and aborts.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
BeyondEspresso: Just because of 10 lines of code of limited complexity? We at least add the 10 lines again as boiler plate. The complexity of this function originates mainly from the floatD-business.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
BeyondEspresso: It would be less general, slightly easier to understand, but would go with an additional else case. Although, we probably do not need to keep 1D or 4D code usage in mind, I will opt for the current shorter version.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
BeyondEspresso: Managing the correct order of TWTSTight members for calculations within initializer lists is error prone. It makes the number of data members cluttered and documentation less easy to follow. This could in principle be solved with nested structs, but then I cannot use the struct members in the same way as the members of TWTSTight, but have to type something like `structInstance.tanAlpha` which makes consecutive equations much more cluttered. This argument is the reason we are using constructive bindings instead to keep everything lightweight.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
BeyondEspresso: If we are outside the TWTSTight envelope `defineTrigonometryShortcuts` is never called, but I need `tanAlpha` to compute this envelope conditions directly afterwards. Defining a seperate function to encapsulate `sincos` does not help code readability.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
BeyondEspresso: Cluttering the class with const data members that are consecutively initialized in initializer lists is a nightmare because I need to keep track of the order of member declarations. Nested structs clutter the code when it is used. Is there a syntax-suger solution to this problem?

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
BeyondEspresso: How can I avoid the follow-up problems that I generate in doing this? The constructive binding already achieves the goal to reduce common boiler plate.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
BeyondEspresso: For semantic reasons I fully agree. The question is mainly: Does the compiler generate warnings if I do not use every variable of an unpacked tuple? If yes, then it is not worth un- and repackaging a tailored argument list for zeroOrder if it can just ride here for free.

include/picongpu/fields/background/templates/twtstight/BField.tpp
BeyondEspresso: No chance and I tried. Complex number computations make stuff short and fast. If we did this, this explodes and we are talking about several 1000 more lines. And the BesselJ-functions cannot be separated into real and imaginary parts.

include/picongpu/fields/background/templates/twtstight/BField.tpp
BeyondEspresso: I am not entirely convinced, I am still thinking about this.

include/picongpu/fields/background/templates/twtstight/BField.tpp
BeyondEspresso: Good idea!

include/picongpu/fields/background/templates/twtstight/BField.tpp
BeyondEspresso: Maybe it is just my taste. I'd argue that it already feels organic to read. Such function extraction would not be about code reuse, but just to get something out of the view.

include/picongpu/fields/background/templates/twtstight/BField.tpp
chillenzer: Ah, of course, I see.

include/picongpu/fields/background/templates/twtstight/BField.tpp
chillenzer: It would be about separation of concerns: The outer function `calcTWTSField`'s concern is providing the interface with convenient arguments by translating them into the variables used by the formula. The new function's concern would be to compute the value from precisely the variables it is formulated with. By separating these concerns the class takes full control over what interface it provides without mangling interface decisions with variable naming conventions. The final function could be as short as:
```C++
auto calcTWTSFieldX(auto pos, ...) {
    auto variables = translateVariables(pos, ...);
    return applyFormula(variables);
}
```
In this formulation, I can easily change the interface of `calcTWTSFieldX` by adjusting `translateVariables` without ever touching something as brittle as 14 lines of difficult math. To me, the latter feels like something you'd really never want to touch again if you got it right once.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
chillenzer: Thanks for the explanation.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
chillenzer: Initialiser lists are typically preferred but if it's better maintainable to simply call a function in the inside of the constructor I don't think it would hurt tremendously, so initialisation could happen inside the constructor body.

Also, structs can be unpacked in the same way as tuples see [here](https://en.cppreference.com/w/cpp/language/structured_binding). So, you could still unpack the names where you need them even if you have them in a struct. But I kind of see that it defeats the purpose a little.

share/picongpu/unit/TWTSTight.cpp
psychocoderHPC: In case we add the add `a == Catch::Approx(b).margin(epsilon)`  we can remove the `std::err`, I am for it.

btw This line is triggering a compiler issue we should check if using `CHECK()` is solving the issue.

psychocoderHPC (dismissed):

include/picongpu/fields/background/templates/twtstight/TWTSTight.hpp
BeyondEspresso: It is now const.

include/picongpu/fields/background/templates/twtstight/TWTSTight.tpp
BeyondEspresso: The compiler does not generate warning, that enabled simplifications.

BeyondEspresso:
@chillenzer @psychocoderHPC I adressed your comments. There were quite some changes in how the helper variables are handled. Now there are two arrays of variables part of the data members, which makes everything more concise. The test case compiles, runs and passes on Frontier.

BeyondEspresso:
@psychocoderHPC @chillenzer The core dump problem has not gone away by changing to `CHECK()`.
```
/builds/hzdr/crp/picongpu/share/picongpu/unit/TWTSTight.cpp(177): internal error: assertion failed at: "error.c", line 7967 in diagnostic_pragma
1 catastrophic error detected in the compilation of "/builds/hzdr/crp/picongpu/share/picongpu/unit/TWTSTight.cpp".
Compilation aborted.
Aborted (core dumped)
make[2]: *** [CMakeFiles/UnitTest-TWTSTight.cpp-2D.dir/build.make:76: CMakeFiles/UnitTest-TWTSTight.cpp-2D.dir/TWTSTight.cpp.o] Error 134
```
Other compilers compiled this without any issue.

share/picongpu/benchmarks/TWEAC-FOM/include/picongpu/param/incidentField.param
chillenzer: ~~Not yet done.~~ Sorry, GH only shows the pertinent commit when you click on the filename in here. Checking all commits, I see that it's done.

include/picongpu/fields/background/templates/twtstight/BField.tpp
chillenzer: This is still inconsistent but it's not blocking for me.

include/picongpu/fields/background/templates/twtstight/BField.tpp
```diff
@@ -26,174 +26,147 @@
...
+                             * ((rhom2 - float_T(2.0) * x2) * sinPhi + I * rhom2 * (-Xm + x * cosPhi) * (k * sinPhi_2)
+                                + x * sin2Phi * Xm)))
+               * psi0)
+            / (cspeed * besselI0const * rhom * rhom2);
+
+        /* A 180deg-rotation of the field vector around the y-axis
+         * leads to a sign flip in the x- and z- components, respectively.
+         * This is implemented by multiplying the result by "phiPositive".
+         */
+        return phiPositive * result.real() / sim.unit.speed();
```
chillenzer, line 80: Just to make sure: The return lines of these functions look slightly different each time. (With/without `/ sim.unit.speed()` and `phiPositive`.) Would you please check that that's on purpose and not by oversight?

include/picongpu/fields/background/templates/twtstight/TWTSTight.hpp
```diff
@@ -62,249 +62,219 @@
...
-            /** To avoid underflows in computation, numsigmas controls where a zero cutoff is made.
-             *  The fields thus are set to zero at a position (numSigmas * tauG * cspeed) ahead
-             *  and behind the respective TWTS pulse envelope.
-             *  Developer note: In case the float_T-type is set to float_64 instead of float_X,
-             *  numSigma can be increased to numSigmas = 10 without running into numerical issues.
-             */
-            constexpr uint32_t numSigmas = 6;
+    public:
+        //! Helper method to define basic TWTS variables
+        HDINLINE static std::array<float_T, 11u> defineBasicHelperVariables(
```
chillenzer, line 91: I think I would still have preferred a struct because we'd gain the advantage of having names for the variables, so it's slightly harder to mix them up. Decomposition into structured bindings would work exactly the same. But your solution achieves what we wanted to achieve, it's slightly shorter and the naming of variables doesn't really take effect because we're using structured bindings anyways, so no blocking objections.

chillenzer (approved):
Tremendous work! Thanks a bunch! This looks really great now, I'd say. One paranoid question below but nothing blocking anymore.

... well, except for that nasty compiler bug. Not sure what to do about that. I'll have a look at it.

include/picongpu/fields/background/templates/twtstight/BField.tpp
BeyondEspresso: I just re-checked. All x- and -z components have the `phiPositive` and all B-field components have the `/ sim.unit.speed()`. Everything is on purpose and fine.

chillenzer: Thanks!

chillenzer:
Okay. I found the offending piece of code using gcc 11.3.0 and CUDA 11.6.0. But I'm not sure if we can do anything about this: The specific combination of compiler and CUDA version fails at the following line:
```C++
#pragma nv_diagnostic push
```
which can be reproduced by putting this anywhere in the code. This line is inserted by Catch2 in `thirdParty/catch2/src/catch2/internal/catch_test_macro_impl.hpp` which defines `INTERNAL_CATCH_TEST` (which is more or less what `CHECK` and `REQUIRE` etc. compile to) through `CATCH_INTERNAL_START_WARNINGS_SUPPRESSION` (from `thirdParty/catch2/src/catch2/internal/catch_compiler_capabilities.hpp`).

Catch2 claims in its [changelog](https://github.com/catchorg/Catch2/blob/f5cee49c71e7c0e805695b03e8af533783789762/docs/release-notes.md?plain=1#L305) that they have fixed warning suppression for NVIDIA in 3.3.0 (and we're using 3.4.0) but apparently that doesn't apply to our compiler/CUDA combination. I tried updating to the latest Catch2 (3.7.1) but that doesn't change anything.

Things that we could do here:
- Disable this test for this specific combination.
- Hack/patch/customise our Catch2 to remove this line.
- Try to make Catch2 believe it's on a different compiler (the host compiler might be an option).

psychocoderHPC:
Thanks for investigating it. Can we please disable the test in the source code for this compiler. 
There are alpaka macros to control the version.

share/picongpu/unit/TWTSTight.cpp
```diff
@@ -0,0 +1,194 @@
...
+        auto res = resultHost.getDataBox();
+        auto hostEfield = float3_X(
+            testEfield.calcTWTSFieldX(pos, time),
+            testEfield.calcTWTSFieldY(pos, time),
+            testEfield.calcTWTSFieldZ(pos, time));
+        auto hostBfield = float3_X(
+            testBfield.calcTWTSFieldX(pos, time),
+            testBfield.calcTWTSFieldY(pos, time),
+            testBfield.calcTWTSFieldZ(pos, time));
+
```
chillenzer, line 167:
```suggestion

// This combination of compilers has a bug that is triggered by Catch2 internally suppressing warnings.
// See https://github.com/ComputationalRadiationPhysics/picongpu/pull/5174#issuecomment-2467890326
#if(__GNUC__ != 11 || __CUDACC_VER_MAJOR__ != 11)

```

share/picongpu/unit/TWTSTight.cpp
```diff
@@ -0,0 +1,194 @@
...
+        for(uint32_t i = 0; i < 3; i++)
+        {
+            CHECK(isApproxEqual(refEfieldT[i], res[i], epsilonAlgebra));
+            CHECK(isApproxEqual(hostEfield[i], res[i], epsilonHostDevice));
+        }
+        for(uint32_t i = 0; i < 3; i++)
+        {
+            CHECK(isApproxEqual(refBfieldT[i], res[i + 3], epsilonAlgebra));
+            CHECK(isApproxEqual(hostBfield[i], res[i + 3], epsilonHostDevice));
+        }
```
chillenzer, line 190:
```suggestion
        }
#endif
```

chillenzer (changes requested):

chillenzer:
@BeyondEspresso, I've suggested the corresponding patch. Feel free to adopt it. @psychocoderHPC, you should probably have a look if that macro's what you expected. Don't think alpaka has anything like that (except for CI specific macros).

BeyondEspresso:
@chillenzer Thanks for the investigation and suggested code change. Let us see whether the updated PR now passes all CI hoops.

share/picongpu/unit/TWTSTight.cpp
psychocoderHPC: In prinziple we have BOOST macros for version control e.g.

```
#if BOOST_COMP_NVCC && BOOST_COMP_NVCC < BOOST_VERSION_NUMBER(11, 3, 0)
```

but I am fine with this change from the PR.

chillenzer (approved):

PR #5176, opened 2024-10-18
shift AtomicConfigNumber to standard unit tests

BrianMarre:
moves the atomicConfigNumber unit test of FLYonPIC from it's custom location to the standard unit test folder of PIConGPU and makes it part of the the unit test suite of PIConGPU.

@attention This PR removes the `picongpu::configNumber::RUN_UNIT_TEST` option in the `atomicPhysics_Debug.param` file. Old setups setting with this option will still work, but any setting of the option will be ignored.

If you want to run this test, use the PIConGPU unit test framework instead.

ci: picongpu

### DIFF
--- a/include/picongpu/param/atomicPhysics_Debug.param
+++ b/include/picongpu/param/atomicPhysics_Debug.param
@@ -27,11 +27,6 @@
 
 namespace picongpu::atomicPhysics::debug
 {
-    namespace configNumber
-    {
-        constexpr bool RUN_UNIT_TESTS = false;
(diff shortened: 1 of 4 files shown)


psychocoderHPC (approved):

chillenzer (approved):
As planned, this is not yet properly integrated into Catch2 and I believe to leverage its full potential, we should do so at some point. I have opened https://github.com/ComputationalRadiationPhysics/picongpu/issues/5186 to track this.

PR #5180, opened 2024-10-19
Update reference for validation of Atomic Physics example

ikbuibui:
Update reference data and fix reference generation scripts for the atomic physics example 
Required for JUREAP CI


ikbuibui:
@BrianMarre Added commit so that we only validate if abundance is above 10^-5. Should we stick to the old reference data?

share/picongpu/examples/AtomicPhysics/validation/GenerateReference.py
```diff
@@ -16,8 +16,8 @@ def generateReference():
         "output_10",
     ]
 
-    results = np.empty((10, 869))
-    for i, output in enumerate(outputs):
+    results = np.empty((10, 26, 869))
```
PrometheusPi, line 19: I do not see where this increase in dimensions comes from. Was it wrong before?

PrometheusPi (approved):
Just a questions

share/picongpu/examples/AtomicPhysics/validation/GenerateReference.py
ikbuibui: Yes, it was a bug in the script before

PR #5185, opened 2024-10-22
bug fixes atomicPhysics(FLYonPIC) RateCalculationReference

BrianMarre:
fix typo in `lib/python/picongpu/extra/utils/__init__.py` and adds an `__init__.py` to FLYonPICRateCalculationReference for faster import and easier use.

ci: no-compile

### DIFF
--- a/lib/python/picongpu/extra/utils/FLYonPICRateCalculationReference/__init__.py
+++ b/lib/python/picongpu/extra/utils/FLYonPICRateCalculationReference/__init__.py
@@ -1,0 +1,5 @@
+from .BoundBoundTransitions import BoundBoundTransitions
+from .BoundFreeFieldTransitions import BoundFreeFieldTransitions
+from .BoundFreeCollisionalTransitions import BoundFreeCollisionalTransitions
+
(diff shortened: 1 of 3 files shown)


lib/python/picongpu/extra/utils/FLYonPICRateCalculationReference/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .BoundBoundTransitions import BoundBoundTransitions
+from .BoundFreeFieldTransitions import BoundFreeFieldTransitions
+
+__all__ = ["BoundBoundTransitions", "BoundFreeCollsionalTransitions", "BoundFreeFieldTransitions"]
```
chillenzer: There are three items in `__all__` but only two imports. That sounds suspicious!

BrianMarre:

chillenzer (changes requested):

lib/python/picongpu/extra/utils/FLYonPICRateCalculationReference/__init__.py
BrianMarre: good catch, fixed

chillenzer (approved):
Waiting for CI but happy to merge afterwards.

PR #5189, opened 2024-10-22
use lumi as default cluster directory 

PrometheusPi:
Another fix for LUMI setup: the default cluster directory was still frontier


PrometheusPi:
~~Do not merge - still wrong~~
Worked fine - but only when loaded for the first time.

steindev (approved):

Issue #5198, opened 2024-11-04
frontier setup outdated


PrometheusPi:
The recently updated frontier profile (#5110), is already outdated again. 
@BeyondEspresso do you have an updated version you can provide quickly? If not, I will work on setting one up.

BeyondEspresso:
@PrometheusPi I think I can provide an updated version.

BrianMarre:
@PrometheusPi is this issue fixed by the PR?

PR #5200, opened 2024-11-04
Update Frontier profile

BeyondEspresso:
As @PrometheusPi pointed out in #5198 the current Frontier profile is outdated. This PR provides the required update.


PrometheusPi (approved):
Compilation works on Frontier

PR #5203, opened 2024-11-06
bug fix atomicPhysics particle type

BrianMarre:
The function `has`, intended for checking whether the species of a passed particle handle is marked as given atomicPhysics ParticleType, mistakenly ignored the specified `ParticleTypeTag` and instead always checked for the Ion Tag.

This PR fixes that and refactors the function for better useability in the intended use case.

ci: picongpu

### DIFF
--- a/include/picongpu/particles/atomicPhysics/ParticleType.hpp
+++ b/include/picongpu/particles/atomicPhysics/ParticleType.hpp
@@ -119,11 +119,15 @@ namespace picongpu::particles::atomicPhysics
 
     namespace traits
     {
+        //! check if specified ParticleType has the specified ParticleTag
+        //!@{
+        //! not found default case
         template<typename T_ParticleType, typename T_ParticleTypeTag, typename = void>
         struct IsParticleType : std::false_type
         {
         };
 
+        //! found case
         template<typename T_ParticleType, typename T_ParticleTypeTag>
         struct IsParticleType<
             T_ParticleType,
@@ -132,8 +136,10 @@ namespace picongpu::particles::atomicPhysics
         {
         };
 
+        //! short hand
         template<typename T_ParticleType, typename T_ParticleTypeTag>
         using IsParticleType_t = typename IsParticleType<T_ParticleType, T_ParticleTypeTag>::type;
+        //@}
 
         template<typename T_FrameType>
         struct GetParticleType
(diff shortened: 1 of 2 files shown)


include/picongpu/particles/atomicPhysics/ParticleType.hpp
```diff
@@ -132,42 +136,58 @@ namespace picongpu::particles::atomicPhysics
...
+        //! short hand
         template<typename T_FrameType>
         using GetParticleType_t = typename GetParticleType<T_FrameType>::type;
 
-        template<typename T_Particle, typename T_ParticleTypeTag>
-        constexpr bool has([[maybe_unused]] T_ParticleTypeTag const& tag)
+        //! check whether particle belongs to a species with given ParticleTypeTag
+        //!@{
+        template<typename T_ParticleTypeTag, typename T_Particle>
+        bool hasParticleTypeTag(T_Particle const&)
```
psychocoderHPC: please add `HDINLNE`

BrianMarre: done

include/picongpu/particles/atomicPhysics/ParticleType.hpp
```diff
@@ -132,42 +136,55 @@ namespace picongpu::particles::atomicPhysics
...
+        //! short hand
         template<typename T_FrameType>
         using GetParticleType_t = typename GetParticleType<T_FrameType>::type;
 
-        template<typename T_Particle, typename T_ParticleTypeTag>
-        constexpr bool has([[maybe_unused]] T_ParticleTypeTag const& tag)
+        //! check whether particle belongs to a species with given ParticleTypeTag
+        //!@{
+        template<typename T_ParticleTypeTag, typename T_Particle>
+        HDINLINE bool hasParticleTypeTag(T_Particle const&)
```
chillenzer, line 157:
```suggestion
        HDINLINE constexpr bool hasParticleTypeTag(T_Particle const&)
```

chillenzer (approved):
You could change that tiny thing but I guess it doesn't really matter, so not blocking here.

include/picongpu/particles/atomicPhysics/ParticleType.hpp
BrianMarre: The problem is that in the intended usage, the particle handle passed in lockstep is not constexpr, therefore we may mark the function as constexpr but actually using it within a lockstep kernel leads to a compiler error.

I therefore deliberately did not mark it as constexpr to nudge a user to use the overloaded version if a constexpr call is intended.

PR #5204, opened 2024-11-07
bugFix calculate IPD input

BrianMarre:
Fixes two bug in the calculation of the debye length in atomicPhysics(FLYonPIC) for the StewartPyatt IPD Model.

I mistakenly used the cell volume instead of the SuperCellVolume, forgot a unit conversion factor for `k_B * T` an dmissed that the accumualtion quantities are scaled with a factor of 1/`sim.unit.typicalNumberOfParticlesPerMacropParticle()`.

PR #5205, opened 2024-11-08
BSI minimal implementation

BrianMarre:
Implements barrier suppression ionization(BSI) in AtomicPhysics(FLYonPIC) as an ionization potential depression(IPD) contribution in the `ApplyIPDIonization`-kernel.

**ATTENTION:** The BSI IPD contribution is only considered in IPDIonization.

Other ionizing transition only consider particle based IPD contributions, like Stewart-Pyatt, when determining the ionization energy.

This is due to Field-based IPD contributions requiring the electric field at the particle position, therefore necessitating access to every superCell cell's field value, which currently is not available in all ChooseTransitionKernels.

A future rework of the IPD implementation is planed that will fix this, but this will take some time.

- [x] requires PR #5204  to be merged first
- [x] needs to be rebased to the dev

### DIFF
--- a/include/picongpu/particles/creation/SpawnFromSourceSpecies.kernel
+++ b/include/picongpu/particles/creation/SpawnFromSourceSpecies.kernel
@@ -156,6 +156,13 @@ namespace picongpu::particles::creation
             // create shared memory variable for kernel state
             PMACC_SMEM(worker, sharedKernelState, typename Modules::KernelStateType);
 
+            // create cache
+            auto cache = Modules::template InitCacheFunctor<T_KernelConfigOptions...>::getCache(
+                worker,
+                superCellIndex,
+                std::forward<T_AdditionalData>(additonalData)...);
+            // thread sync not necessary since we sync all threads later anyway, after completing shared memory init
+
             auto numberProductParticlesCtxArr
                 = lockstep::makeVar<TypeNumber>(forEachFrameSlot, static_cast<TypeNumber>(0u));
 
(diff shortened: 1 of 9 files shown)


ikbuibui (dismissed):

ikbuibui (dismissed):

include/picongpu/particles/creation/SpawnFromSourceSpecies.kernel
```diff
@@ -156,6 +156,13 @@ namespace picongpu::particles::creation
             // create shared memory variable for kernel state
             PMACC_SMEM(worker, sharedKernelState, typename Modules::KernelStateType);
 
+            // create cache
+            auto cache = Modules::template InitCacheFunctor<T_KernelConfigOptions...>::getCache(
+                worker,
+                superCellIndex,
+                std::forward<T_AdditionalData>(additonalData)...);
+            // worker.sync() not necessary since shared memory init may be executed in parallel
```
psychocoderHPC: I do not understand why sync is not required since init is executed in parallel.

psychocoderHPC (changes requested):

include/picongpu/particles/creation/SpawnFromSourceSpecies.kernel
BrianMarre: offline discussion, I wanted to note that I skip the thread sync after the collective cache init, because we sync after completing the shared memory init anyway and the shared memory init does not depend on the cache init.

psychocoderHPC (approved):

PR #5206, opened 2024-11-08
Fix infinite ADK rates

BrianMarre:
fixes an infinite loop in the AtomicPhysics FLYonPIC stage due to a numerical underflow of the atomicPhysics timestep due to very high ADK rates of BSI unbound states still being present, by moving the Removal of IPD unbound states before the filling of the rate cache.

- [x] requires PR #5205 to be merged first
- [x] requires PR #5204 to be merged first
- [x] requries PR #5207 to be merged first
- [x] needs to be rebased to the dev

### DIFF
--- a/include/picongpu/simulation/stage/AtomicPhysics.x.cpp
+++ b/include/picongpu/simulation/stage/AtomicPhysics.x.cpp
@@ -464,10 +464,7 @@ namespace picongpu::simulation::stage
                 do
                 {
                     resetFoundUnboundIon();
-                    picongpu::atomicPhysics::IPDModel::
-                        template calculateIPDInput<T_numberAtomicPhysicsIonSpecies, IPDIonSpecies, IPDElectronSpecies>(
-                            mappingDesc,
-                            currentStep);
+                    calculateIPDInput(mappingDesc, currentStep);
                     picongpu::atomicPhysics::IPDModel::template applyIPDIonization<AtomicPhysicsIonSpecies>(
                         mappingDesc,
                         currentStep);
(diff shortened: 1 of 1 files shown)


include/picongpu/simulation/stage/AtomicPhysics.x.cpp
```diff
@@ -584,10 +581,12 @@ namespace picongpu::simulation::stage
                     recordChanges(mappingDesc);
                     updateElectrons(mappingDesc, currentStep);
                     updateElectricField(mappingDesc);
-                    doIPDIonization(mappingDesc, currentStep, deviceLocalReduce);
                     updateTimeRemaining(mappingDesc);
                     isSubSteppingComplete = isSubSteppingFinished(mappingDesc, deviceLocalReduce);
                 } // end atomicPhysics sub-stepping loop
+
+                // ensure no unbound states are visible to the rest of the loop
+                doIPDIonization(mappingDesc, currentStep, deviceLocalReduce);
```
psychocoderHPC, line 589: Does the mean we execute the IPD ionization twice per pic time step? Because we call this already in line 540.

psychocoderHPC (changes requested):

include/picongpu/simulation/stage/AtomicPhysics.x.cpp
BrianMarre: yes, that is intended and necessary.

The second call of `doIPDIonization` is necessary here since we need to make sure that we do not pass an unbound state with a too low charge state to the pusher and we need to call `doIPDIonization` again at the beginning of the next AtomicPhyiscs step since the we updated the field in between causing a possible change result of the call.

Also note calling `doIPDIonization` twice on the same particle set with an unchanged E-Field does not change particle set anymore after the first call.

psychocoderHPC (dismissed):

psychocoderHPC:
@BrianMarre can be rebased

psychocoderHPC (approved):

PR #5212, opened 2024-11-13
Refactor passed atomic data boxes check

BrianMarre:
extract the check of the passed bound-free atomic data boxes to avoid code duplication 

This is done in preparation for the implementation of the instant field ionization transitions.

- [ ] requires PR #5206 to be merged first
- [ ] needs to be rebased to dev

every but the last commit belongs to the base PRs


BrianMarre:
replaced by PR #5220 to reduce CI load

PR #5213, opened 2024-11-13
Refactor bound free field ionization rate calculation

BrianMarre:
refactor the bound free field calculation to allow changing the precision of the calculation.

This is done in preparation for the implementation of instant field ionization transitions.

- [ ] requires PR #5206 to be merged first
- [ ] requires PR #5212 to be merged first
- [ ] needs to be rebased to the dev

every but the last commit belongs to the base PRs


BrianMarre:
replaced by PR #5220 to reduce CI load

PR #5218, opened 2024-11-13
instant field ionization transitions

BrianMarre:
implementation of instant field ionization transitions.

- [x] requires PR #5206 to be merged first
- [x] requires PR #5220 to be merged first
- [x] requires PR #5227 to be merged first 
- [x] needs to be rebased to dev

Also fixes a bug in the `FillRateCache` atomicPhysics(FLYonPIC) sub-stage and removes the unused probability estimator implementations, FLYonPIC currently always uses linear probability estimation only.

Only commits after `refactor Record Changes kernel` are part of this PR, all other commits belong to a base PR.

**Attention!** This PR breaks setups using custom `atomicPhysics.param` files. To fix a setup add the following new member to the `RateSolverParam` struct.
```c++
/** maximum number of atomicPhysics sub-steps per PIC time step
 *
 * all rates above timeStepAlpha / maximumNumberSubStepsPerPICTimeStep will be solved in
 *  rate fix point(equilibirium) approximation only.
 *
 * @attention This limit is currently only enforced for field ionization transitions! If the number of
 * sub-steps required by collisional processes is higher than the limit set, AtomicPhysics will perform more
 * subs-steps than set here.
 */
static constexpr uint32_t maximumNumberSubStepsPerPICTimeStep = 200u;
```
**Attention!** This PR breaks setups using custom `atomicPhysics_Debug.param` files. To fix an setup add the following to the old file.
```c++
//! @attention performance relevant
namespace picongpu::atomicPhysics::debug::kernel::chooseTransition
{
    //! @attention performance relevant
    constexpr bool CHECK_FOR_INFINITE_FIELD_IONIZATION_RATES = false;
} // namespace picongpu::atomicPhysics::debug::kernel::chooseTransition
namespace picongpu::atomicPhysics::debug::kernel::chooseInstantTransition
{
    //! @attention performance relevant
    constexpr bool CHECK_FOR_INFINITE_FIELD_IONIZATION_RATES = false;
} // namespace picongpu::atomicPhysics::debug::kernel::chooseInstantTransition
```


BrianMarre:
@psychocoderHPC the failed CI-jobs got stuck somehow, please restart them

include/picongpu/particles/atomicPhysics/kernel/CheckForFieldEnergyOverSubscription.kernel
```diff
@@ -0,0 +1,122 @@
...
+            auto const superCellIdx = KernelIndexation::getSuperCellIndex(worker, areaMapping);
+            auto const superCellFieldIdx
+                = KernelIndexation::getSuperCellFieldIndexFromSuperCellIndex(areaMapping, superCellIdx);
+
+            auto const timeRemaining = timeRemainingBox(superCellFieldIdx);
+            if(timeRemaining <= 0._X)
+                return;
+
+            using FieldEnergyCache = typename T_FieldEnergyUseCacheBox::ValueType;
+            particles::atomicPhysics::localHelperFields::RejectionProbabilityCache_Cell<FieldEnergyCache::numberCells>&
```
psychocoderHPC: I wrote it already in an other PR. It looks very dangerous to use a reference here.
If it is a pmacc box please take the return value always by copy. There is only one case where a reference is ok and this is if you use native alpaka to get shared memory.

psychocoderHPC: Could be that this should be change in other code parts too.

include/picongpu/particles/atomicPhysics/stage/CheckForFieldEnergyOverSubscription.hpp
```diff
@@ -0,0 +1,103 @@
...
+            auto& rejectionProbabilityCacheField_Cell
+                = *dc.get<picongpu::particles::atomicPhysics::localHelperFields::RejectionProbabilityCacheField_Cell<
+                    picongpu::MappingDesc>>("RejectionProbabilityCacheField_Cell");
+
+            auto& eField = *dc.get<FieldE>(FieldE::getName());
+
+            using FieldEnergyUseCacheField
+                = picongpu::particles::atomicPhysics::localHelperFields::FieldEnergyUseCacheField<
+                    picongpu::MappingDesc>;
+            auto& fieldEnergyUseCacheField = *dc.get<FieldEnergyUseCacheField>("FieldEnergyUseCacheField");
```
psychocoderHPC: shoudl be `dc.get` and later use `fieldEnergyUseCacheField->` to access methods.

include/picongpu/particles/atomicPhysics/stage/ChooseInstantTransition.hpp
```diff
@@ -0,0 +1,122 @@
...
+                pmacc::AreaMapping<CORE + BORDER, MappingDesc> mapper(mappingDesc);
+                pmacc::DataConnector& dc = pmacc::Environment<>::get().DataConnector();
+
+                auto& timeRemainingField
+                    = *dc.get<particles::atomicPhysics::localHelperFields::TimeRemainingField<picongpu::MappingDesc>>(
+                        "TimeRemainingField");
+                auto& foundUnboundIonField = *dc.get<
+                    particles::atomicPhysics::localHelperFields::FoundUnboundIonField<picongpu::MappingDesc>>(
+                    "FoundUnboundIonField");
+                auto& eField = *dc.get<FieldE>(FieldE::getName());
```
psychocoderHPC: same here too, please avoid deeling with references.

psychocoderHPC (changes requested):

psychocoderHPC:
short rule: if you write code like `type& x = function();` this is in 99% of all cases wrong and requires removing the reference.
This is a typical case to produce segfaults.

include/picongpu/particles/atomicPhysics/kernel/CheckForFieldEnergyOverSubscription.kernel
BrianMarre: The use of a reference is correct and necessary here, since the dataBox will return a reference to global device memory and we don't want to store a copy of the superCell's rejection probability cache for every worker due to its significant size, ~ 512 * 32-bit.

include/picongpu/particles/atomicPhysics/stage/CheckForFieldEnergyOverSubscription.hpp
BrianMarre: fixed

include/picongpu/particles/atomicPhysics/stage/ChooseInstantTransition.hpp
BrianMarre: fixed

BrianMarre:
A general comment, most AtomicPhysics(FLYonPIC) sub stages still use the old `auto& <name> = *dc.get<type>(<ID>)` style. I updated all new and all changed sub-stages in this PR but we will have to update all other old stages separately.

psychocoderHPC (approved):

PR #5221, opened 2024-11-15
Fix energy calculation when applying a mask

fabidie:
Apparently I have been unable to follow basic calculation rules 🤨  (first squaring, then summing)

### DIFF
--- a/lib/python/picongpu/extra/input/preparingInsightData.py
+++ b/lib/python/picongpu/extra/input/preparingInsightData.py
@@ -411,8 +411,8 @@ class PrepRoutines:
             return None
 
         else:
-            # original field integral (needed for field ratio calculation)
-            int_orig = np.sum(np.abs(self.Ew))
+            # original field intensity integral (needed for energy ratio calculation)
+            int_orig = np.sum(np.abs(self.Ew) ** 2)
 
             # propagation FF to MF
             X, Y, W = np.meshgrid(self.x, self.y, self.w)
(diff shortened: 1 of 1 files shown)


lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -411,8 +411,8 @@ def aperture_in_mf(self, d, R, xc=0, yc=0):
             return None
 
         else:
-            # original field integral (needed for field ratio calculation)
-            int_orig = np.sum(np.abs(self.Ew))
+            # original field intensity integral (needed for energy ratio calculation)
+            int_orig = np.sum(np.abs(self.Ew) ** 2)
```
psychocoderHPC, line 415: note: `abs` is not required if you square the value

lib/python/picongpu/extra/input/preparingInsightData.py
```diff
@@ -439,8 +439,10 @@ def aperture_in_mf(self, d, R, xc=0, yc=0):
             self.is_masked = True
             print("Aperture of radius %.2f mm at distance %.2e mm has been applied." % (R, d))
 
-            # we need the ratio of the masked far field data to the original one to calculate the correct beam energy later
-            self.field_ratio = np.sum(np.abs(self.Ew)) / int_orig
+            # we need the integrated intensity ratio of the masked far field data to the original one to account in for
+            # the energy loss due to the masking process
+            self.field_ratio = np.sum(np.abs(self.Ew) ** 2) / int_orig
```
psychocoderHPC, line 444: same here

psychocoderHPC (approved):

lib/python/picongpu/extra/input/preparingInsightData.py
fabidie: Yes it is, since Ew is complex

chillenzer (approved):

PR #5223, opened 2024-11-18
Metadata optimization: Only write ADIOS2 attributes from rank 0 in openPMD-api >= 0.16

franzpoeschel:
We were thinking about using this as the default option generally in openPMD, but [our documentation](https://openpmd-api.readthedocs.io/en/0.15.2/details/mpi.html) specifies that attribute writing is collective only in HDF5, and attributes can be written independently in ADIOS2. So it would have broken the API contract to assume that attributes outside of rank 0 can be ignored.
In PIConGPU, however, they can.

### DIFF
--- a/include/picongpu/plugins/openPMD/Json.cpp
+++ b/include/picongpu/plugins/openPMD/Json.cpp
@@ -339,6 +339,37 @@ The key 'select' must point to either a single string or an array of strings.)EN
          * careful memory configuration is necessary anyway, so the
          * defaults don't matter.
          */
+#    if OPENPMDAPI_VERSION_GE(0, 16, 0)
+        /*
+         * In openPMD >= 0.16, additionally ignore all attribute metadata
+         * that comes from a rank other than rank 0.
+         * The openPMD plugin of PIConGPU writes attributes synchronously
+         * (since this is required in common configurations of HDF5).
+         * For ADIOS2 however, it's better to write attributes from single
+         * ranks only and avoid metadata duplication across ranks.
+         * This option tells the ADIOS2 backend of openPMD that it can
+         * safely ignore any attribute write if the current MPI rank is
+         * not 0.
+         */
+        std::string const& defaultValues = R"(
+{
+  "hdf5": {
+    "dataset": {
+      "chunks": "none"
+    }
+  },
+  "adios2": {
+    "attribute_writing_ranks": 0,
+    "engine": {
+      "preferred_flush_target": "buffer",
+      "parameters": {
+        "BufferChunkSize": 2147381248
+      }
+    }
+  }
+}
+        )";
+#    else
         std::string const& defaultValues = R"(
 {
   "hdf5": {
@@ -356,6 +387,7 @@ The key 'select' must point to either a single string or an array of strings.)EN
   }
 }
         )";
+#    endif
         std::stringstream json_to_string;
         json_to_string << config;
         auto merged = openPMD::json::merge(defaultValues, json_to_string.str());
--- a/include/picongpu/plugins/openPMD/Json.cpp
+++ b/include/picongpu/plugins/openPMD/Json.cpp
@@ -339,7 +339,6 @@ The key 'select' must point to either a single string or an array of strings.)EN
          * careful memory configuration is necessary anyway, so the
          * defaults don't matter.
          */
-#    if OPENPMDAPI_VERSION_GE(0, 16, 0)
         /*
          * In openPMD >= 0.16, additionally ignore all attribute metadata
          * that comes from a rank other than rank 0.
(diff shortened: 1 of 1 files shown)


psychocoderHPC (dismissed):

include/picongpu/plugins/openPMD/Json.cpp
```diff
@@ -339,6 +339,37 @@ The key 'select' must point to either a single string or an array of strings.)EN
...
+         */
+        std::string const& defaultValues = R"(
+{
+  "hdf5": {
+    "dataset": {
+      "chunks": "none"
+    }
+  },
+  "adios2": {
+    "attribute_writing_ranks": 0,
```
chillenzer: Why isn't this more targeted? Like this
```suggestion
#    if OPENPMDAPI_VERSION_GE(0, 16, 0)
    "attribute_writing_ranks": 0,
#    endif
```
(with appropriate string concatenation). The next person changing something in this configuration will have to do so in two places. I'm aware that these are close together and it's not a huge burden but still it's something that could get overlooked in a quick diff that doesn't show the other section. (In fact, I've only seen the duplication because I've looked at the full file to understand what's going on. The current diff doesn't even show the other version.)

chillenzer (changes requested):

include/picongpu/plugins/openPMD/Json.cpp
franzpoeschel: Tried that at first, it's far worse to read than just seeing directly the configuration listed for the supported versions:

```cpp
#    if OPENPMDAPI_VERSION_GE(0, 16, 0)
#        define PICONGPU_ATTRIBUTE_WRITING_RANKS                                                                      \
            "\n"                                                                                                      \
            R"(    "attribute_writing_ranks": 0,)"
#    else
#        define PICONGPU_ATTRIBUTE_WRITING_RANKS ""
#    endif

        std::string const& defaultValues = R"(
{
  "hdf5": {
    "dataset": {
      "chunks": "none"
    }
  },
  "adios2": {)" PICONGPU_ATTRIBUTE_WRITING_RANKS
                                           R"(
    "engine": {
      "preferred_flush_target": "buffer",
      "parameters": {
        "BufferChunkSize": 2147381248
      }
    }
  }
}
        )";
```

The only solution that I'd really consider an improvement in maintainability would be to define separately:

```cpp
        std::string const& baseConfiguration = R"(
{
  "hdf5": {
    "dataset": {
      "chunks": "none"
    }
  },
  "adios2": {
    "engine": {
      "preferred_flush_target": "buffer",
      "parameters": {
        "BufferChunkSize": 2147381248
      }
    }
  }
}
        )";
    std::string const& openPMD_0_16_additionalConfiguration = 
        R"({"adios2": {"engine":{"attribute_writing_ranks": 0}}})";
```

And then use `openPMD::json::merge()` instead of manual string concatenation, but that's an additional round of parsing, an additionaly call to an external library and an additional runtime error source, too.

For a long-term solution, I'd suggest that we add something like `{"dont_warn": ["preferred_flush_target"]}` to openPMD so that warnings about unknown keys can be suppressed, making this above distinction unnecessary:
```json
{
  "hdf5": {
    "dataset": {
      "chunks": "none"
    }
  },
  "adios2": {
    "engine": {
      "dont_warn": ["preferred_flush_target"],
      "preferred_flush_target": "buffer",
      "parameters": {
        "BufferChunkSize": 2147381248
      }
    }
  }
}
```

chillenzer:
A very online "offline" discussion: The current solution oscillates between string and json representation quite a bit. A cleaner solution that would also allow for a very readable solution of this particular problem would be to stay completely in json-world. @franzpoeschel will consider if changing this a beneficial move and act accordingly.

franzpoeschel:
With the recent commit:

Executing PIConGPU compiled against openPMD-api 0.16.0 
```bash
picongpu -g 32 32 32 -d 1 1 1 -s 100 --openPMD.period 100 --openPMD.json '{"adios2":{"attribute_writing_ranks": [0, 1], "engine": {"parameters":{"StatsLevel": 1}}}}'
```
Results in this configuration sent to the openPMD-api:
```json
{
  "adios2": {
    "attribute_writing_ranks": [
      0,
      1
    ],
    "engine": {
      "parameters": {
        "BufferChunkSize": 2147381248,
        "StatsLevel": 1
      },
      "preferred_flush_target": "buffer"
    }
  },
  "hdf5": {
    "dataset": {
      "chunks": "none"
    }
  }
}
```

With:
```bash
picongpu -g 32 32 32 -d 1 1 1 -s 100 --openPMD.period 100
```
Results in this configuration sent to the openPMD-api:
```json
{
  "adios2": {
    "attribute_writing_ranks": 0,
    "engine": {
      "parameters": {
        "BufferChunkSize": 2147381248
      },
      "preferred_flush_target": "buffer"
    }
  },
  "hdf5": {
    "dataset": {
      "chunks": "none"
    }
  }
}
```

Against openPMD-api 0.15.2:
With:
```bash
picongpu -g 32 32 32 -d 1 1 1 -s 100 --openPMD.period 100 --openPMD.json '{"adios2":{"engine": {"parameters":{"StatsLevel": 1}}}}'
```
Results in :
```json
{
  "adios2": {
    "engine": {
      "parameters": {
        "BufferChunkSize": 2147381248,
        "StatsLevel": 1
      },
      "preferred_flush_target": "buffer"
    }
  },
  "hdf5": {
    "dataset": {
      "chunks": "none"
    }
  }
}
```

With
```bash
picongpu -g 32 32 32 -d 1 1 1 -s 100 --openPMD.period 100
```
Results in:
```json
{
  "adios2": {
    "engine": {
      "parameters": {
        "BufferChunkSize": 2147381248
      },
      "preferred_flush_target": "buffer"
    }
  },
  "hdf5": {
    "dataset": {
      "chunks": "none"
    }
  }
}
```

chillenzer (approved):
Slightly unconventional to still go through the string. I would have expected to construct the json object directly but it's fine with me.

PR #5224, opened 2024-11-18
openPMD plugin: Add option for Append mode in writing

franzpoeschel:
https://openpmd-api.readthedocs.io/en/0.16.0/usage/workflow.html#access-modes
The Append mode is more secure as it protects better against accidentally overwriting data, e.g. when restarting from a checkpoint. I'm hesitant to fully replace the Create mode with Append mode, since (1) overwriting could be intended and (2) there were bugs in the past that made it impossible to create new files in Append mode with ADIOS2 on some platforms (I think mostly a Windows issue).

TODO:

- [x] What should we set as the default? (psychocoderHPC: we will not change the default do to the risk of breaking thinks because the meaning of `append` in openPM depends if you write file based or group based IO)

### DIFF
--- a/docs/source/usage/plugins/openPMD.rst
+++ b/docs/source/usage/plugins/openPMD.rst
@@ -140,6 +140,14 @@ PIConGPU command line option          description
                                       If your system host memory is small compared to the accelerator memory you should use ``--openPMD.dataPreparationStrategy mappedMemory`` to have
                                       a smaller host memory footprint.
                                       The memory footprint required to dump fields will not be affected by this parameter.
+``--openPMD.writeAccess``             openPMD Access mode for file writing. Default: ``create``. Selecting ``append`` can be useful for checkpoint-restart workflows for avoiding
+                                      truncation of data upon reopening the output Series. Note that the Append mode in openPMD-api is used for adding new Iterations to a data Series
+                                      that (possibly) already exists, but does not specify what happens when an Iteration is written for a second time.
+                                      This situation will typically occur in checkpoint-restart workflows where a handful of iterations is usually recomputed.
+                                      In this situation, the backends will pick a sensible, but unspecified implementation for proceeding. This may include:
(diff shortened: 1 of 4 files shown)


psychocoderHPC:
IMO for PIConGPU the default should be that we create a new file. If you like to append we should perform this explicit.

franzpoeschel:
Just to make this clear: The only difference between both modes is when a file already exists. The current setting will truncate the file, the Append setting will add new Iterations, while leaving the old data intact.
If no file exists yet, both modes will create a new file.

I have no strong preference for either option, so if you prefer sticking with the current setting, then I'll leave this as it is now.

psychocoderHPC:
IMO keeping the current behavior is better. If someone is restarting a simulation after a checkpoint lets say at step 1000 but the simulation before was running until 1500 and than crashed at 1700 before writing a new checkpoint at 2000.
If the output contains particles and we append data to the existing data at step 1500 we will have twice as many particles in the output. Thi will mean the output is broken because the data makes no sense.
IMO append should be controlled explicitly and only used where it is guaranteed to be save.

PrometheusPi:
@franzpoeschel and @psychocoderHPC what is the conclusion of your discussion and what is the status of this PR (is it reviewable)?

psychocoderHPC:
Let me move this into draft that it will not be merged. The feature to set the mode is very important for workflows we need to initialize bunches but append should not be the default.

psychocoderHPC (dismissed):

psychocoderHPC:
@franzpoeschel only to avoid mistakes can I merge this PR?

franzpoeschel:
I should still add documentation. Currently traveling, so I'll do it when I
find a free minute

René Widera ***@***.***> schrieb am Do., 28. Nov. 2024, 08:45:

franzpoeschel:
@psychocoderHPC can now be reviewed

psychocoderHPC (approved):

PR #5226, opened 2024-11-25
Fix gcc 9 unused parameter error

ikbuibui:
Fix for the failing full CI run

### DIFF
--- a/include/picongpu/particles/atomicPhysics/spawnFromSourceSpeciesModules/CacheEFieldForSuperCell.hpp
+++ b/include/picongpu/particles/atomicPhysics/spawnFromSourceSpeciesModules/CacheEFieldForSuperCell.hpp
@@ -48,13 +48,13 @@ namespace picongpu::particles::atomicPhysics::spawnFromSourceSpeciesModules
             typename... T_IPDInputBoxes>
         HDINLINE static auto getCache(
             T_Worker const& worker,
-            pmacc::DataSpace<picongpu::simDim> const superCellIndex,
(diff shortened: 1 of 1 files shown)


chillenzer (approved):

PrometheusPi (commented):
looks okay
but do we need to add `[[maybe_unused]]` to line 62 too?

ikbuibui:
`[[maybe_unused]]` suppresses warnings about variables that might not be used in the current scope. This old compiler sees that the `superCellIndex` and `eFieldBox` are not used in the `else` branch of the `if constexpr`, so it throws a warning/error, even though they are conditionally used in the `if` branch on line 62.
`[[maybe_unused]]` isnt required on line 62 because it isnt defining any parameters which are later not used. It is only returning the cache. It is the responsibility of the caller of this function to store this cache and then use it. If the caller assigns the returned cache to a variable and then doesn't use it this warning/error will be thrown at the caller side.
Just for information, kind of similarly if you want to make sure that the return value of a function is stored (and not ignored/discarded) by the caller, you can declare the function ` [[nodiscard]]`.

PrometheusPi:
@ikbuibui Thanks for the clarification. Thant makes sense, I was following the documentation example [here](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) and was a bit confused.

PrometheusPi (approved):

psychocoderHPC (approved):

psychocoderHPC:
Thanks @ikbuibui I planed to fix it too, good that I checked the PRs first

PR #5240, opened 2024-12-18
Update reference data for atomic physics vaidation

ikbuibui:
**UPDATE** : This PR updates the atomic physics reference data. After a discussion with @BrianMarre we think this is required because of bug fixes in  #5207,  #5206 and related PRs.  
I have tested the new reference data on HAL and it looks correct. Once this PR is in, this test will be run on JEDI, which will be additional confirmation.

I am leaving the initial PR description here, as the correctness and robustness of the validation script is still questionable. 
OLD PR discussion :
This PR makes the validation threshold less strict.
An offline discussion with @BrianMarre led to the conclusion that the validation script has subtle and difficult to diagnose problems (it is unclear if they are related to reference statistics, or to physics), which require a significant amount of work to analyze.
A crude short term workaround of something reasonable is increasing the threshold for now, with validation improvements as a TODO.


ikbuibui:
@PrometheusPi @steindev @psychocoderHPC @chillenzer Sorry to ping everyone, but can someone merge this. I need it for runs on JEDI

BrianMarre (approved):

PrometheusPi (approved):

PR #5242, opened 2024-12-19
update README.md

psychocoderHPC:
- add Thomas, Klaus and Richard to the Scientific Supervision section for supervising the students contributing to PIConGPU
- update the list of active developers/maintainers

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,15 +187,20 @@ Active Team
 ### Scientific Supervision
 
 - Dr. Michael Bussmann
+- Dr. Thomas Kluge
+- Dr. Richard Pausch
+- Dr. Klaus Steiniger
 
 ### Maintainers* and core developers
 
-- Dr. Sergei Bastrakov*
 - Finn-Ole Carstens
 - Dr. Alexander Debus
 - Dr. Marco Garten*
 - Dr. Axel Huebl*
+- Dr. Jeffrey Kelling*
+- Dr. Julian Lenz* 
 - Brian Edward Marre
+- Tapish Narwal*
 - Pawel Ordyna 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -206,11 +211,11 @@ Active Team
 
 The PIConGPU Team expresses its gratitude to:
 
-Florian Berninger, Heiko Burau, Fabia Dietrich, Robert Dietrich, Carlchristian Eckert,
+Dr. Sergei Bastrakov, Florian Berninger, Heiko Burau, Fabia Dietrich, Robert Dietrich, Carlchristian Eckert,
 Simeon Ehrig, Wen Fu, Ph.D., Alexander Grund, Sebastian Hahn, Anton Helm, Wolfgang Hoehnig,
```
ikbuibui:
```suggestion
 Wen Fu, Ph.D.,
```
The title should be fixed for consistency

steindev: Well, yes probably. You mean we should omit 'Ph.D.' and write 'Dr.' at the beginning of the name everywhere, right?

README.md
```diff
@@ -206,11 +211,11 @@ Active Team
 
 The PIConGPU Team expresses its gratitude to:
```
steindev: I think, 'Team' should not be written with a capital 'T'.

README.md
psychocoderHPC: I will keep Ph.D. as it was before but put it to the beginning but only for the names where it was already in but it was blindly copied and therefore add the seperator comma.

README.md
psychocoderHPC: done

ikbuibui (approved):

PR #5244, opened 2024-12-19
update memory calculator

BrianMarre:
This PR,
- refactors the memory calculator:
  - type validation using pydantic
  - function signature checks using typeguard
  - remove explicit passing of extents by dimension
  - changes particle attribute specification to be by list of names
  - updates the pml border default size to the current defaults
  - changes the pml border size specification to match the param files
- refactors the use example
  - bug-fix in the particle memory calculation, now uses sim dimension 2 instead of the old default value 3
  - update to the new interfaces
  - generalize the implementation
  - switches to per gpu output instead of per row output
- adds the memory calculation for atomic physics super cell fields and atomic physics particle attributes

### DIFF
--- a/docs/source/usage/workflows/memoryPerDevice.py
+++ b/docs/source/usage/workflows/memoryPerDevice.py
@@ -18,14 +18,41 @@ import numpy as np
 This file contains a usage example of the ``MemoryCalculator`` class
 for our :ref:`FoilLCT example <usage-examples-foilLCT>` and its ``4.cfg``.
 
-It is an estimate for how much memory is used per device if the whole
-target would be fully ionized but does not move much. Of course the real
-memory usage depends on the case and the dynamics inside the simulation.
+It calculates an estimate for how much memory is used per device if the whole
+target would be fully ionized but does not move much.
 
-We calculate the memory of just one device per row of GPUs in laser
-propagation direction. We hereby assume that particles are distributed
-equally in transverse direction, like it is set up in the FoilLCT example.
+The no-movement approximation may severely underestimate the memory requirements
+of individual GPUs, especially in setups with significant laser compression
+or implosion, due to macro-particle movement over the simulation run.
+The real-world memory usage always depends on the case and the dynamics inside
+the simulation and no-movement estimates should be taken with a grain of salt as
+an order-of-magnitude estimation.
+For a more accurate estimation use estimated compression multipliers.
 """
+
+
+def check_distributions(simulation_dimension, grid_distribution, gpu_particle_cell_distribution):
+    if len(grid_distribution) != simulation_dimension:
+        raise ValueError("grid_distribution must have one array for each simulation dimension")
+    if len(gpu_particle_cell_distribution) != simulation_dimension:
+        raise ValueError("gpu_particle_cell_distribution must have array for each simulation dimension")
+
+
+def check_extents(simulation_dimension, global_gpu_extent, gpu_cell_extent, gpu_particle_cell_extent):
+    if np.shape(gpu_cell_extent) != (simulation_dimension, *global_gpu_extent):
+        raise ValueError(
+            "grid_distribution and global_gpu_extent are not consistent.\n"
+            "\tCheck for each simulation dimension that grid_distribution has as many entries as gpu"
+            "rows are specified in global_gpu_extent."
+        )
+
+    if np.shape(gpu_particle_cell_extent) != (simulation_dimension, *global_gpu_extent):
+        raise ValueError(
+            "gpu_particle_cell_distribution and global_gpu_extent are not consistent.\n"
+            "\tCheck for each simulation dimension that gpu_particle_cell_distribution"
+            "has as many entries as gpu rows are specified in global_gpu_extent."
+        )
+
 
 cell_size = 0.8e-6 / 384.0  # 2.083e-9 m
 y0 = 0.5e-6  # position of foil front surface (m)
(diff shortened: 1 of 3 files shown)


docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -20,72 +21,131 @@
...
+
+#   for simplicity's sake we will distribute cells evenly, but this is not required
+if np.any(global_cell_extent % global_gpu_extent != 0):
+    raise ValueError("global cell extent must be divisble by the global gpu extent")
+global_domain_division = []
+for dim in range(simulation_dimension):
+    row_division = np.full(global_gpu_extent[dim], int(global_cell_extent[dim] / global_gpu_extent[dim]), dtype=np.int_)
+    global_domain_division.append(row_division)
+
+print(f"global domain division: {global_domain_division}")
```
pordyna: For me, it is not clear what domain division is supposed to be. I think you should name it differently.

docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -20,72 +21,131 @@
...
+
+# get cell extent of each GPU, <extent of the gpu domain>[simulation dimension, multi dimensional gpu index]
```
pordyna: Not sure I get this notation

docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -20,72 +21,131 @@
...
+                gpu_particle_cell_extent[dim][gpu_index] = gpu_cell_extent[dim][gpu_index]
+            else:
+                gpu_particle_cell_extent[dim][gpu_index] = gpu_cell_extent[dim][gpu_index] - (
+                    start_foil - start_gpu_domain
+                )
+
     else:
-        GPU_rows[ii] -= extra_cells
-        break
+        # fully in side target
```
pordyna:
```suggestion
        # fully inside target
```

lib/python/picongpu/extra/utils/memory_calculator.py
```diff
@@ -3,317 +3,399 @@
...
+            # bytes
+            return np.float32().itemsize
+        elif precision == 64:
+            # bytes
+            return np.float64().itemsize
         else:
+            raise ValueError("unsupported precision {precision}")
+
+    @staticmethod
+    def get_predefined_attribute_dict(simulation_dimension: int, precision: int) -> dict[str, int]:
```
pordyna, line 81: Why is it a static method? `simulation_dimension`, `precision` are also instance attributes. So as long as you put in the class, I would make just use these values.

lib/python/picongpu/extra/utils/memory_calculator.py
```diff
@@ -3,317 +3,399 @@
...
+            "processClass": 8,
+            "transitionIndex": 32,
+            "binIndex": 32,
+            "accepted": 1,
+            "atomicPhysicsIonParticleAttributes": value_size + 32 + 8 + 32 + 32 + 1,
+            "totalCellIdx": simulation_dimension * 32,
+        }
+
+    @typeguard.typechecked
+    def check_cell_extent(self, cell_extent: nptype.NDArray):
```
pordyna: Not part of the user interface. I think you should hide it by prepending `_` to the method name.

lib/python/picongpu/extra/utils/memory_calculator.py
```diff
@@ -3,317 +3,399 @@
...
+        if not (np.all(cell_extent > 0)):
+            raise ValueError("number cells must be > 0 in all dimensions")
+
+        if np.any(cell_extent % self.super_cell_size != 0):
+            raise ValueError(
+                "device number cells must be an integer multiple of the super_cell_size,"
+                " please set super_cell_size to a correct value"
+            )
+
+    def checkDimensionsOfArrays(self):
```
pordyna: Also, not part of the user interface. I think you should hide it by prepending `_` to the method name.

lib/python/picongpu/extra/utils/memory_calculator.py
```diff
@@ -3,317 +3,399 @@
...
+        """check all set array have expected dimension"""
+        if (self.pml_border_size).ndim != 2:
+            raise ValueError("pml_border_size must be 2D array")
+        if (self.super_cell_size).ndim != 1:
+            raise ValueError("super_cell_size must be 1D array")
+        if (self.guard_size).ndim != 1:
+            raise ValueError("guard_size must be 1D array")
+
+    def shrink_to_simulation_dimension(self):
```
pordyna: Also, not part of the user interface. I think you should hide it by prepending `_` to the method name.

lib/python/picongpu/extra/utils/memory_calculator.py
```diff
@@ -3,317 +3,399 @@
...
+                self.pml_border_size = self.pml_border_size[:2]
+
+            if (self.super_cell_size).shape[0] == 3:
+                self.super_cell_size = self.super_cell_size[:2]
+
+            if (self.guard_size).shape[0] == 3:
+                self.guard_size = self.guard_size[:2]
+
+    @typeguard.typechecked
+    def check(self):
```
pordyna: Also, not part of the user interface. I think you should hide it by prepending `_` to the method name.

lib/python/picongpu/extra/utils/memory_calculator.py
```diff
@@ -3,317 +3,399 @@
...
+        self.check_cell_extent(cell_extent)
 
-            # cells around core-border region due to particle shape
-            double_buffer_cells = (n_x + pso) * (n_y + pso) * (n_z + pso) - n_x * n_y * n_z
-        else:
-            raise ValueError("PIConGPU only runs in either 2D or 3D: ", sim_dim, " =/= {2, 3}")
+        # PML size cannot exceed the local grid size
+        pml_border_size = np.minimum(self.pml_border_size, cell_extent)
+
+        # one scalar each for temp fields, E_x, B_x, E_y, B_y, ...
```
pordyna:
```suggestion
        # one scalar for each temp field, E, B, and J field component
```

lib/python/picongpu/extra/utils/memory_calculator.py
```diff
@@ -3,317 +3,399 @@
...
+
+        return cell_memory + double_buffer_memory + pml_memory
+
+    @typeguard.typechecked
+    def memory_required_by_super_cell_fields(
+        self,
+        super_cell_extent: nptype.NDArray,
+        number_atomic_states_by_atomic_physics_ion_species: list[int],
+        number_electron_histogram_bins: int,
+        IPDactive: bool = True,
```
pordyna: `IPDactive` -> `ipd_active`

pordyna (changes requested):
Nice! Looks mostly good to me, though I made some mostly just pedantic comments.

docs/source/usage/workflows/memoryPerDevice.py
BrianMarre: after an offline discussion I followed Pawel's advice and renamed it to ``grid_distribution`` to match more closely the PIConGPU terminology.

docs/source/usage/workflows/memoryPerDevice.py
BrianMarre: rephrased the comment to make it clearer.

docs/source/usage/workflows/memoryPerDevice.py
BrianMarre: fixed

lib/python/picongpu/extra/utils/memory_calculator.py
BrianMarre: since these methods might be usable in other contexts I decided to make them independent of the current implementation.

lib/python/picongpu/extra/utils/memory_calculator.py
BrianMarre: fixed

lib/python/picongpu/extra/utils/memory_calculator.py
BrianMarre: fixed

lib/python/picongpu/extra/utils/memory_calculator.py
BrianMarre: fixed

lib/python/picongpu/extra/utils/memory_calculator.py
BrianMarre: fixed

lib/python/picongpu/extra/utils/memory_calculator.py
BrianMarre: fixed

lib/python/picongpu/extra/utils/memory_calculator.py
BrianMarre: fixed

lib/python/picongpu/extra/utils/memory_calculator.py
pordyna: @BrianMarre ok, but in that case I would just take it out of the class :)

docs/source/usage/workflows/memoryPerDevice.py
chillenzer, line 1: [This line](https://github.com/ComputationalRadiationPhysics/picongpu/blob/79ee6eff1ff0ec8240db75171251bc1a1fa2977d/docs/source/usage/workflows/memoryPerDevice.rst?plain=1#L26) needs adaption. See the wrong rendering of the example code [here](https://picongpu--5244.org.readthedocs.build/en/5244/usage/workflows/memoryPerDevice.html).

docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -20,72 +21,131 @@
 It is an estimate for how much memory is used per device if the whole
 target would be fully ionized but does not move much. Of course the real
 memory usage depends on the case and the dynamics inside the simulation.
+
 We calculate the memory of just one device per row of GPUs in laser
 propagation direction. We hereby assume that particles are distributed
 equally in transverse direction, like it is set up in the FoilLCT example.
```
chillenzer: While we're at it, this is a duplication of the text in [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/79ee6eff1ff0ec8240db75171251bc1a1fa2977d/docs/source/usage/workflows/memoryPerDevice.rst?plain=1#L13). I would suggest to to deduplicate by using another literal include in the rst version. Maybe something like "The following script's docstring explains its functionality: ... (<literal include here>)

docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -20,72 +21,131 @@
...
+grid_distribution = []
+for dim in range(simulation_dimension):
+    row_division = np.full(global_gpu_extent[dim], int(global_cell_extent[dim] / global_gpu_extent[dim]), dtype=np.int_)
+    grid_distribution.append(row_division)
+
+print(f"grid distribution: {grid_distribution}")
+
+# get cell extent of each GPU: list of np.array[np.int_], one per simulation dimension, with each array entry being the
+#   cell extent of the corresponding gpu in the simulation, indexation by [simulation_dimension, gpu_index[0], gpu_index[1], ...]
+gpu_cell_extent = np.meshgrid(*grid_distribution)
```
chillenzer:
```suggestion
gpu_cell_extent = (global_cell_extent / global_gpu_extent * np.ones([simulation_dimension, *global_gpu_extent])).astype(int).T
```
Equivalent oneliner (up to the fact that we're getting a numpy array, of course). I was tempted to use an explicit `np.floor` to hint on the fact that we're potentially cutting something off here. But the `ValueError` above takes care of this. We might want to print the `gpu_cell_extent` instead of the `grid_distribution` in this case.

docs/source/usage/workflows/memoryPerDevice.py
```diff
@@ -20,72 +21,131 @@
 It is an estimate for how much memory is used per device if the whole
 target would be fully ionized but does not move much. Of course the real
 memory usage depends on the case and the dynamics inside the simulation.
+
```
chillenzer: Maybe a more explicit warning would be nice here: Something along the lines of "... and there are scenarios (e.g. imploding setups) where the not-moving approximation severely underestimates memory requirements of individual GPUs. So take it with a grain of salt as an order-of-magnitude estimate."

lib/python/picongpu/extra/utils/memory_calculator.py
```diff
@@ -3,317 +3,399 @@
...
+            "transitionRadiationMask": 1,
+            "boundElectrons": value_size,
+            "atomicStateCollectionIndex": 32,
+            "processClass": 8,
+            "transitionIndex": 32,
+            "binIndex": 32,
+            "accepted": 1,
+            "atomicPhysicsIonParticleAttributes": value_size + 32 + 8 + 32 + 32 + 1,
+            "totalCellIdx": simulation_dimension * 32,
+        }
```
chillenzer: Offline discussion: Use bytes and bools are (undefined but typically) larger than 1 bit. Technically, a byte could have a different size than 8.

lib/python/picongpu/extra/utils/memory_calculator.py
```diff
@@ -3,317 +3,399 @@
...
         # number of additional PML field components: when enabled,
         # 2 additional scalar fields for each of Ex, Ey, Ez, Bx, By, Bz
-        num_pml_fields = 12
+        number_pml_fields = 12
 
-        req_mem = (
-            self.value_size * num_fields * local_cells
-            + double_buffer_mem
-            + self.value_size * num_pml_fields * local_pml_cells
+        number_local_cells = int(np.prod(cell_extent + self.super_cell_size * 2 * self.guard_size))
```
chillenzer, line 189: It's not quite obvious what `cell_extent` refers to. It looks like it means `extent_in_number_of_cells` while I first read it as `extent_of_a_cell`. Better name would be great.

lib/python/picongpu/extra/utils/memory_calculator.py
```diff
@@ -3,317 +3,399 @@
...
+
+        @detail species flag/constants do not occupy global device memory, since they are compiled in.
+
+        @param particle_filled_cells either number or extent of particle filled cell on device
+        @param species_attribute_list list of species attribute names of species see ``species.param``,
+            for example ["momentum", "position", "weighting"]
+        @param additional_attributes list of size of additional attributes, **in bits**, for example {"custom":32}
+        @param particles_per_cell number of particles of the species per cell
+
+        @return unit: bytes
```
chillenzer: Offline discussion: Remove default ppc and number of particles directly needs ppc=1

lib/python/picongpu/extra/utils/memory_calculator.py
```diff
@@ -3,317 +3,399 @@
...
+        Memory required for cell fields on a specific device(GPU/CPU/...)
 
-        # guard size in super cells in x, y, z
-        guard_size_supercells = np.array([1, 1, 1])
+        @attention In PIConGPU different devices may handle different number of cells. This naturally also changes the
+            memory required on each device. This function returns the memory required by cell fields on one single
+            device handling the specified cell extent, not the global memory requirement!
 
-        pso = particle_shape_order
+        @param cell_extent device cell extent
```
chillenzer, line 172: Just to distinguish from below: Be explicit that a volume (a number) is not sufficient, we need a vector here.

chillenzer (changes requested):
I have strong opinions about the qualities of my bike's accommodation but there's actually one thing about the counting in bits and bytes that needs reworking. Others are optional. Haven't checked if the details of the memory usage actually match up with what the code does.

chillenzer:
Forgot: Offline discussion in the meeting: We want to keep the main example minimal and concrete (and not have the "kind of general" computation for varying parameters) and might add the "how to compute the used values" in a separate section.

docs/source/usage/workflows/memoryPerDevice.py
BrianMarre: fixed, also left a not for future developers

BrianMarre: fixed

docs/source/usage/workflows/memoryPerDevice.py
BrianMarre: fixed

docs/source/usage/workflows/memoryPerDevice.py
BrianMarre: done.

docs/source/usage/workflows/memoryPerDevice.py
BrianMarre: as discussed offline, we decided to hard code values, for better readability.

lib/python/picongpu/extra/utils/memory_calculator.py
BrianMarre: done

lib/python/picongpu/extra/utils/memory_calculator.py
BrianMarre: done

lib/python/picongpu/extra/utils/memory_calculator.py
BrianMarre: done

BrianMarre:
> Forgot: Offline discussion in the meeting: We want to keep the main example minimal and concrete (and not have the "kind of general" computation for varying parameters) and might add the "how to compute the used values" in a separate section.

simplified the example, will leave the general description of how to compute the used values to a later refactor.

lib/python/picongpu/extra/utils/memory_calculator.py
BrianMarre: It is supposed the describe the extent of the spatial domain of a GPU in cells, I failed not come up with a better name so would keep it for now and let the next person figure out a better name.

chillenzer (approved):

PR #5247, opened 2025-01-08
Adapted Snakemake param file to match LWFA example parameters

JessicaTiebel:
I have adapted the example `param.csv` for the Snakemake workflow with the LaserWakeField example to match the names of the parameters. The naming was not consistent before and led to confusion.


chillenzer:
Offline discussion: @JessicaTiebel clarified that this is a bug fix. The naming not only led to confusion but wrong parameter names were injected into the generated param files such that the given values were not recognized. *Verification that it works now is still pending.* (Both runs should yield different results.)

chillenzer (approved):

Issue #4622, opened 2023-07-04
CI does not test gcc-12


BrianMarre:
Our documentation, list gcc-12 as a supported compiler but the CI does not test gcc-12
https://picongpu.readthedocs.io/en/latest/install/dependencies.html

SimeonEhrig:
@BrianMarre Done in #5255

PR #5246, opened 2025-01-08
switch to C++20

psychocoderHPC:
- update docuemntation
- update CI tests
  - remove HIP 5.X
  - remove CUDA 11.X
  - remove GCC9
- set in all CMake files the C++ standard to C++20
-  update to [mallocMC 2.7.0-dev](https://github.com/alpaka-group/mallocMC/tree/127b29fabbb01c9a055bb5d21213daf2db5e6843)
  - This update is required to remove a deprecated warning triggered when switching to C++20 because most volatile usages are deprecated with C++20.

  ```
  GIT_AUTHOR_NAME="Third Party" GIT_AUTHOR_EMAIL="picongpu@hzdr.de" \
    git subtree pull --prefix thirdParty/mallocMC \
    git@github.com:ComputationalRadiationPhysics/mallocMC.git dev --squash
  rm -r thirdParty/mallocMC/alpaka
  git merge --continue
  ```
- CI: **disable HIP** checks
  - the HIP container using a to old libstdc++ and base compiler which is not C++20 compatible
  - using clangs `libc++` is not possible because it is not shipped with the clang compiler provided by AMD's ROCM
  - the checks will be re-enabled after new base container are available

PR #5255, opened 2025-01-27
CI: update to container version 4.0

psychocoderHPC:
- switch to container version 4.0
- re-enable HIP tests disabled with #5246
- add HIP 6.2 and 6.3
- add CUDA 12.6
- add boost 1.86 and 1.87
- add clang 19
- add gcc 14
- adjust filter rules for new CUDA and HIP versions
- adjust filter rules to support ubuntu 22.04 and 24.04

used `ci: full-compile` to run the full test matrix


ikbuibui:
Looks okay. I am wondering if we should start testing more modern clang versions with cuda 
17 seems to fully support 12.1 https://github.com/llvm/llvm-project/blob/release/17.x/clang/docs/ReleaseNotes.rst
18 seems to fully support 12.3 https://github.com/llvm/llvm-project/blob/release/18.x/clang/docs/ReleaseNotes.rst

psychocoderHPC:
> Looks okay. I am wondering if we should start testing more modern clang versions with cuda 17 seems to fully support 12.1 https://github.com/llvm/llvm-project/blob/release/17.x/clang/docs/ReleaseNotes.rst 18 seems to fully support 12.3 https://github.com/llvm/llvm-project/blob/release/18.x/clang/docs/ReleaseNotes.rst


CLang is only supporting CUDA up to 12.3 and this is [already in the code](https://github.com/ComputationalRadiationPhysics/picongpu/pull/5255/files#diff-808864eba06dc2856bde25a5696c4d0e4193eccea9101f8c7ae2132d99b4a2bfR152)

Clang 17 showed the last time where I added clang 17 and 18 support issues and therefore I have not added clang 17 and CUDA tests

ikbuibui (dismissed):

share/ci/n_wise_generator.py
```diff
@@ -127,7 +128,7 @@ def is_valid_combination(row):
             return False
         else:
             # install/clang.sh is currently only providing apt sources up to clang 18
```
ikbuibui:
```suggestion
            # install/clang.sh is currently only providing apt sources up to clang 19
```

SimeonEhrig:
@psychocoderHPC You can also add GCC 12. There was a bug in the Ubuntu 20.04 ppa, so we could not install it. In Ubuntu 22.04 and 24.04 the GCC is available. This would fix issue #4622

share/ci/install/clang.sh
```diff
@@ -7,10 +7,8 @@ set -o pipefail
 
 cd $CI_PROJECT_DIR
 
-# provide sources for clang versions missing in the containers version 3.2
-echo -e "deb http://apt.llvm.org/focal/ llvm-toolchain-focal-16 main\ndeb-src http://apt.llvm.org/focal/ llvm-toolchain-focal-16 main" >> /etc/apt/sources.list.d/llvm.list
-echo -e "deb http://apt.llvm.org/focal/ llvm-toolchain-focal-17 main\ndeb-src http://apt.llvm.org/focal/ llvm-toolchain-focal-17 main" >> /etc/apt/sources.list.d/llvm.list
-echo -e "deb http://apt.llvm.org/focal/ llvm-toolchain-focal-18 main\ndeb-src http://apt.llvm.org/focal/ llvm-toolchain-focal-18 main" >> /etc/apt/sources.list.d/llvm.list
+# provide sources for clang versions missing in the containers version 4.0
+#echo -e "deb http://apt.llvm.org/noble/ llvm-toolchain-noble-17 main\ndeb-src http://apt.llvm.org/noble/ llvm-toolchain-noble-17 main" >> /etc/apt/sources.list.d/llvm.list
```
chillenzer: Why is this still in? It looks like that's dead code?

share/ci/n_wise_generator.py
```diff
@@ -100,10 +100,9 @@ def is_valid_combination(row):
         os_name = row[2][0] if n >= 3 else ""
         os_version = get_version(row[2]) if n >= 3 else 0
 
-        if is_cuda and os_name == "ubuntu":
-            # CI container version 3.1 do not support ubuntu 18.04 anymore
-            if os_version < 20.04:
-                return False
+        # old OS versions base compile mostly does not support C++20
```
chillenzer:
```suggestion
        # old OS versions' base compiler does not fully support C++20
```

share/ci/n_wise_generator.py
chillenzer: Better because lower maintenance overhead:
```suggestion
            # install/clang.sh is currently not providing apt sources later than this clang version
```
Proof: See your comment

share/ci/n_wise_generator.py
```diff
@@ -245,9 +272,11 @@ def is_valid_combination(row):
     "1.83.0",
     "1.84.0",
     "1.85.0",
+    "1.86.0",
+    "1.87.0",
```
chillenzer, line 276: For reducing CI load, we could probably strip one or two of these. Say, we take the oldest supported and then with a stride of 2 or 3 to the latest.

share/ci/n_wise_generator.py
```diff
@@ -245,9 +272,11 @@ def is_valid_combination(row):
...
 ]
 
-operating_system = [("ubuntu", 20.04)]
+operating_system = [("ubuntu", 22.04), ("ubuntu", 24.04)]
```
chillenzer, line 279: Why do we actually use both?

chillenzer (commented):
Hey, it has finally passed!

Mostly questions for my information. One or two non-blocking suggestions.

share/ci/n_wise_generator.py
SimeonEhrig: ROCm 6.0 until 6.2 only supports Ubuntu 22.04

share/ci/n_wise_generator.py
```diff
@@ -177,22 +184,39 @@ def is_valid_combination(row):
...
     ("clang++", 13),
     ("clang++", 14),
     ("clang++", 15),
     ("clang++", 16),
     ("clang++", 17),
     ("clang++", 18),
+    ("clang++", 19),
 ]
-gnu_compilers = [("g++", 10), ("g++", 11), ("g++", 13)]
+gnu_compilers = [("g++", 10), ("g++", 11), ("g++", 13), ("g++", 14)]
```
SimeonEhrig:
```suggestion
gnu_compilers = [("g++", 10), ("g++", 11), ("g++", 12), ("g++", 13), ("g++", 14)]
```

share/ci/n_wise_generator.py
psychocoderHPC: possible but will reduce coverage. there is a flag in this python code `--limit_boost_versions` to use only each second version

psychocoderHPC:
I applied the changes.

chillenzer:
No, you didn't. At least you didn't push as far as I can see.

ikbuibui (approved):

PR #5249, opened 2025-01-23
Add per species particle filtering

ikbuibui:
Adds ability to filter particles before binning. Filters can be defined per species 
Breaks existing input sets using the binning plugin 
Users now need to use createSpeciesTuple instead of create tuple

PR #5256, opened 2025-01-27
set C++ compiler in PyPIConGPU tests to gcc 11

### LINKED ISSUE #5250: Pypicongpu compile test compiles with GCC9
GCC 9 should be disabled since #5246 but while working on #5249 I ran into this error - https://gitlab.com/hzdr/crp/picongpu/-/jobs/8938122835
It looks like we still try compiling with GCC9. I had a quick look but it wasnt immediately obvious to me where this is coming from

SimeonEhrig:
fix #5250

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -197,7 +197,7 @@ pypicongpu-compiling-test:
     PIP_BREAK_SYSTEM_PACKAGES: 1
     CI_CONTAINER_NAME: 'ubuntu20.04'
     PYTHON_VERSION: '3.11.*'
-    CXX_VERSION: 'g++-11'
+    CXX_VERSION: 'g++-13'
     CMAKE_VERSION: '3.25'
     BOOST_VERSION: '1.75.0'
     PYTHON_COMPILING_TEST: 'ON'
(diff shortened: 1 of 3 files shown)


SimeonEhrig:
Looks like the change is working. Unfortunately the CI does not show the used gcc version, if everything works. But set the compiler with `export CXX=$CXX_VERSION`, it failed the first time because I used the wrong container image and the CI could not find gcc-13 and than I fixed it.

Waiting for review from @ikbuibui , if this PR fixes his problem.

ikbuibui:
Without checking against my PR it seems like it is still GCC 9 that is used in the CI. This is the pypicongpu-compiling-test from the CI of this PR - https://gitlab.com/hzdr/crp/picongpu/-/jobs/8967984161
But I think with #5255 this problem should be automatically be solved

SimeonEhrig:
> Without checking against my PR it seems like it is still GCC 9 that is used in the CI. This is the pypicongpu-compiling-test from the CI of this PR - https://gitlab.com/hzdr/crp/picongpu/-/jobs/8967984161 But I think with #5255 this problem should be automatically be solved

The host compiler (GCC 9) is only used to install the dependencies. For the actual Python tests, it uses the compiler which was specified. The PR #5255 does not solve the problem, because it does not use `export CXX=CXX_VERSION`.

ikbuibui (approved):
Tested that it successfully works in #5249

PR #5262, opened 2025-01-31
Update binning plugin documentation for filtered species and log axis

ikbuibui:
Update documentation with 
- log axis
- filtered species

Filtered species introduced in #5249 breaks existing param files as users are now required to use ``createSpeciesTuple`` instead of ``createTuple`` for species, filtered or otherwise

### DIFF
--- a/docs/source/usage/plugins/binningPlugin.rst
+++ b/docs/source/usage/plugins/binningPlugin.rst
@@ -78,7 +78,7 @@ If no units are given, the quantity is assumed to be dimensionless.
     momentumDimension[SIBaseUnits::mass] = 1.0;
     momentumDimension[SIBaseUnits::time] = -1.0;
 
-.. doxygenenum:: picongpu::traits::SIBaseUnits::SIBaseUnits_t
+.. doxygenenum:: picongpu::SIBaseUnits::SIBaseUnits_t
 
 
 Axis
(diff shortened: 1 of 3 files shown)


docs/source/usage/plugins/binningPlugin.rst
```diff
@@ -126,10 +137,24 @@ Species can be instances of a species type or a particle species name as a PMACC
     
     auto electronsObj = PMACC_CSTRING("e"){};
 
+Optionally, users can specify a filter to be used with the species. This is a predicate functor, i.e. it is a functor with a signature as described above and returns a boolean. If the filter returns true it means the particle is included in the binning.
```
PrometheusPi:
```suggestion
Optionally, users can specify a filter to be used with the species. This is a predicate functor, i.e. it is a functor with a signature as described above and which returns a boolean. If the filter returns true it means the particle is included in the binning.
```

PrometheusPi: I am not sure if this is needed here @ikbuibui Please correct me and ignore my suggestion or update the sentence as you like.

PrometheusPi (dismissed):
just a minor typo

docs/source/usage/plugins/binningPlugin.rst
ikbuibui: applied

PrometheusPi (dismissed):

psychocoderHPC (approved):

PR #5263, opened 2025-01-31
Fix: added default number of compile threads.

FilipO28555:
Added the default number of compile threads.
This number should be dependent on available RAM.

@psychocoderHPC am I right?

### DIFF
--- a/bin/pic-build
+++ b/bin/pic-build
@@ -47,6 +47,9 @@ help()
 
 cmd_line_args=()
 
+# default number of compile threads. 20 uses a little over 40GB of RAM compiling Bunch Example
+nCompileThreads=20
+
 # show help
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
`pic-build -j` behaves like `make -j` is using the maximum required parallelism. If you like to limit it you should call `pic-build -j 20`.
As you wrote the maximum number depends on the available memory and if you compile for CUDA or CPU.

The behavior of `-j` is described if you call `pic-build --help` or in the docs https://picongpu.readthedocs.io/en/latest/usage/basics.html#pic-build

I think we are not able to add any default because we do not know how much memory per compile unit we require.

FilipO28555:
OK, thanks, than I'm closing the pr

PR #5265, opened 2025-02-04
Add field binning to binning plugin

### LINKED ISSUE #4800: Inconsistent docstring for `addBinner`
Hi @ikbuibui, just read through your binning plugin and found that the docstring [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/59d2b696055b826cf281a0a83b34ac019d71f46c/include/picongpu/plugins/binning/BinningCreator.hpp#L54) appears to describe arguments that are not actually taken by the function.

ikbuibui:
- Add ability to bin fields
- Add a host side hook for the user to execute code on the host side, before the binning is executed for the particular notify step. For example to fill ``fieldTmp``. 
- Add ability to pass extra data to the user defined binning functors
- Add ability to optionally get particle position to sub-cell precision and also optionally in SI units instead of number of cells.
 
Refactored the ``Binner`` into a parent binner class and specialized child particle and field binners.
Separated the helper kernels from ``Binner``into a separate file

This PR closes #4800

**Important for users**, changes to ``DomainInfo``:

``getParticlePosition<DomainOrigin>`` now accepts two additional optional template parameters, ``PositionPrecision`` and ``PositionUnits`` as ``getParticlePosition<DomainOrigin, PositionPrecision, PositionUnits>``.  The former by default uses cell level precision, but if you want the exact particle position and not just the cell, you can use ``PositionPrecision::SubCell``. The latter by default uses cell units, but it is also possible to get positions in SI units by specifying ``PositionUnits::SI``
The field binning has its own ``DomainInfo`` object. It additionally holds the ``localCellIndex`` in the supercell and has a method to ``getCellIndex<DomainOrigin, CellUnits>`` to get the current cell index being binned relative to an origin (global, total, local).  

**Breaking changes** - Note: The following break existing binning param files. 

- The old ``addBinner`` method of ``binningCreator`` is renamed to ``addParticleBinner`` and ``binningCreator`` has a new method ``addFieldBinner``
- All functors now take arguments in the order ``worker, domainInfo, ...`` instead of ``domainInfo, worker, ...`` for consistency with the rest of the codebase.
- Quantity functors now get access to domain info as well, thus unifying the functor interface with the current one for axes as ``(auto const& worker, auto const& domainInfo, auto const& particle)`` `
- The ``writeOpenPMDFunctor`` has been moved from a optional argument to ``addBinner`` to a setter similar to other properties as ``setWriteOpenPMDFunctor``.
- The setter``setJsonCfg`` has been renamed to ``setOpenPMDJsonCfg``


include/picongpu/plugins/binning/BinningFunctors.hpp
```diff
@@ -59,21 +58,53 @@ namespace picongpu
                         // This assumes n_bins and getBinIdx exist
                         validIdx
                             = (
-                                   [&]
+                                   [&](auto const& tupleArg)
```
ikbuibui, line 61: Passing explicitly as argument as a workaround for some compiler issues

PR #5266, opened 2025-02-05
CI: fix filter rules

psychocoderHPC:
Only CUDA was tested in the CI because the new filter rules from #5255 used the OS version which was not available during the execution of the filter.

### DIFF
--- a/share/ci/n_wise_generator.py
+++ b/share/ci/n_wise_generator.py
@@ -80,7 +80,7 @@ def get_base_image(compiler, backend):
 def is_valid_combination(row):
     n = len(row)
 
-    if n >= 2:
+    if n >= 3:
         v_compiler = get_version(row[0])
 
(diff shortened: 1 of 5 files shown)


share/ci/n_wise_generator.py
```diff
@@ -186,7 +186,7 @@ def is_valid_combination(row):
 
         # clang as host compiler
         if is_clang:
-            if os_name == "ubuntu" and os_version == 22.04 and v_compiler <= 12:
+            if os_name == "ubuntu" and os_version == 22.04 and os_version <= 12:
```
ikbuibui:
```suggestion
            if os_name == "ubuntu" and os_version == 22.04 and v_compiler <= 12:
```

ikbuibui (changes requested):

ikbuibui (dismissed):

ikbuibui (approved):
As a note : Disables clang 14 for reasons documented in code

PR #5272, opened 2025-02-11
fix CUDA lambda error

psychocoderHPC:
fix compiler error seen with CUDA 12.1.
Remove usage of gerenic lambda by functor like kernel.

```
pmacc/particles/IdProvider.hpp(91): error: __host__ __device__ extended lambdas cannot be generic lambdas
              auto kernel = []
```

@BeyondEspresso  This should be the fix for your compile issue

### DIFF
--- a/include/pmacc/particles/IdProvider.hpp
+++ b/include/pmacc/particles/IdProvider.hpp
@@ -47,6 +47,18 @@ namespace pmacc
     };
 
 
+    struct StoreId
+    {
+        /** get unique id
+         *
+         * @nextId storage for a single valid id
+         */
+        DINLINE void operator()(auto const& worker, auto idGenerator, uint64_t* nextId) const
+        {
+            *nextId = idGenerator.fetchInc(worker);
+        }
+    };
+
     class IdProvider : public ISimulationData
     {
     public:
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
The error showed up in @BeyondEspresso unit tests with PiConGPU. The reason was that he had not activated alpaka CUDA only mode.
Never the less the fix makes the code base more stable and because it removes a optional CUDA feature.

ikbuibui (approved):

BeyondEspresso:
@psychocoderHPC With this fix, the error just shifts to further files:
```
include/pmacc/random/methods/MRG32k3aMin.hpp(61)
include/pmacc/random/methods/MRG32k3aMin.hpp(71)
include/pmacc/random/methods/XorMin.hpp(95)
share/picongpu/unit/radiationFormFactor.cpp(105)
picongpu/unit/radiationWindowFunction.cpp(93)
```
Instead update `share/picongpu/unit/README.rst` by replacing
```bash
cmake <path_to_picongpu_source>/share/picongpu/unit/ -Dalpaka_ACC_GPU_CUDA_ENABLE=ON
```
with
```bash
cmake <path_to_picongpu_source>/share/picongpu/unit/ -Dalpaka_ACC_GPU_<CUDA|HIP>_ENABLE=ON -Dalpaka_ACC_GPU_CUDA_ONLY_MODE=ON
```

ikbuibui:
Not merging yet since this same error is still seen in other parts of the code. Maybe we want to fix all instances, or leave them all

psychocoderHPC:
we should use https://github.com/ComputationalRadiationPhysics/picongpu/pull/5272#issuecomment-2651261101 suggestion and update the rest file only, else we need to remove all lambda functions

psychocoderHPC:
I will close this PR, as long as we have only one alpaka accelerator it is fine to use generic lambda functions defined on host side. 
If we start using host accelerator together with GPU accelerators we can not use anymore generic lambda functions except the lambda is only executable on the GPU device.

PR #5273, opened 2025-02-11
Allow double precision in TWTSTight unit test

BeyondEspresso:
This applies only if a developer tests the TWTSTight laser at double precision by setting `using float_T = float_64;` in `fields/background/templates/twtstight/TWTSTight.hpp`. Usually, only single precision is required.

### DIFF
--- a/share/picongpu/unit/TWTSTight.cpp
+++ b/share/picongpu/unit/TWTSTight.cpp
@@ -56,8 +56,15 @@ using namespace pmacc;
  * @return true if equal, else false
  */
 template<typename T>
-static bool isApproxEqual(T const& a, T const& b, T const& epsilon = std::numeric_limits<T>::epsilon())
+static bool isApproxEqual(T const& a, T const& b, T const& epsilon)
 {
+    return a == Catch::Approx(b).margin(epsilon);
+}
+
+template<typename T>
+static bool isApproxEqual(T const& a, T const& b)
+{
+    T const epsilon = std::numeric_limits<T>::epsilon() * max(abs(a), abs(b));
     return a == Catch::Approx(b).margin(epsilon);
 }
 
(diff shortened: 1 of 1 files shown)


share/picongpu/unit/TWTSTight.cpp
```diff
@@ -56,11 +56,18 @@ using namespace pmacc;
...
-static bool isApproxEqual(T const& a, T const& b, T const& epsilon = std::numeric_limits<T>::epsilon())
+static bool isApproxEqual(T const& a, T const& b, T const& epsilon)
 {
     return a == Catch::Approx(b).margin(epsilon);
 }
 
+template<typename T>
+static bool isApproxEqual(T const& a, T const& b)
+{
+    T const epsilon = std::numeric_limits<T>::epsilon() * max(abs(a), abs(b));
```
ikbuibui, line 67: This now defines a new default behavior if epsilon is not provided. Earlier it was using ``std::numeric_limits<T>::epsilon()`` now it is using a relative error, i.e. ``abs(val) * std::numeric_limits<T>::epsilon()``
Im not sure why this change was made since it seems like it isnt used, since in the code it looks like an epsilon is always provided.

share/picongpu/unit/TWTSTight.cpp
```diff
@@ -177,7 +184,7 @@ struct twtsTightNumberTest
          * PMacc with boost library calls that also work on device. */
         const float_T epsilonAlgebra = float_T(5.0e-6);
         /* Epsilon to compare host implementation to device implementation */
-        const float_T epsilonHostDevice = float_T(5.0e-15);
+        const float_T epsilonHostDevice = float_T(1.0e-7);
```
ikbuibui, line 187: Since, it was not mentioned in the PR, is this change really required? Was the old epsilon too strict?

BeyondEspresso: Yes, the old epsilon was too strict and worked only by accident. I identified the less strict epsilon by varying the parameters of the test case

share/picongpu/unit/TWTSTight.cpp
BeyondEspresso: `std::numeric_limits<T>::epsilon()` is only valid when comparing to values of about unity. The extension is a generalization of the old code. However, the calculation can only be done at run time now, so that I used overloading instead of a default value. Originally, I tried to obtain automatic epsilon values for `epsilonHostDevice` this way. However, those still turned out to be too strict, so that I kept the manual setting. In principle the 3 argument case can be safely deleted in this unit test case.

ikbuibui (approved):

PR #5274, opened 2025-02-11
Fix static_cast-bug in TWTSTight laser

BeyondEspresso:
Bad static_casts on float3_X led to wrong results. Now, precisionCasts are being used.


ikbuibui (approved):
Looks okay. In ``TWTSTight.hpp`` ``float3_T`` is already defined as a vector of 3 ``float_X``s ,so the precision casts are not strictly needed. But since it looks like ``using float_T = float_X;`` is a customization point, for example to allow using doubles internally but keeping the function output as ``float_X`` to keep the API stable, I will accept

PR #5275, opened 2025-02-12
Made bunch creation tool compatible with CTR plugin

nicowrobel:
The transition radiation plugin expects a `transitionRadiationMask` as an particle property in the speciesDefinition.param file. The bunch initialization tool now detects, if this property is needed and initializes it for the created bunch (all values set to FALSE).
Without this commit, the simulation would crash when the CTR plugin is enabled and it restarts from a checkpoint, where the  `transitionRadiationMask` is missing from the particles.

### DIFF
--- a/lib/python/picongpu/extra/input/bunchInit_openPMD_bp.py
+++ b/lib/python/picongpu/extra/input/bunchInit_openPMD_bp.py
@@ -189,6 +189,7 @@ class addParticles2Checkpoint:
         self.has_probeB = False
         self.has_momentumPrev1 = False
         self.has_id = False
+        self.has_transitionRadiationMask = False
 
         # extract data type for position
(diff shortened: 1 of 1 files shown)


chillenzer (commented):
Hi,
Thanks for your contribution. Looks fine to me but admittedly I'm looking at this script for the first time. Is this tested? And if so, how can I test this myself?

nicowrobel:
I tested it for my setup and it works, but I am also more or less the only user, so it's hard for me to test different setups.
But you can test it yourself by running the example notebook `createBunch_example.ipynb` in the same folder. One must provide a path to an existing empty (no particles) checkpoint and the tool will create a new one with particles inside. The code is working, if PIConGPU can run from this new checkpoint.

chillenzer (approved):

PR #5279, opened 2025-02-18
Update alpaka

chillenzer:
Updates to the latest `develop` of alpaka which includes one patch necessary for mallocMC. In the process, an interface change was accounted for and for some reason the `constexpr` functions of the PMAcc tuple had to become `HDINLINE`.

```bash
GIT_AUTHOR_NAME="Third Party" GIT_AUTHOR_EMAIL="picongpu@hzdr.de" git subtree pull --prefix thirdParty/alpaka git@github.com:alpaka-group/alpaka.git develop --squash
```


ikbuibui:
When you resubmit i would recommend setting full-compile

ikbuibui (commented):
Hopefully removing HDINLINE from the constructors, assignment operators should fix the errors popping up in the CI

chillenzer:
> Hopefully removing HDINLINE from the constructors, assignment operators should fix the errors popping up in the CI

Discussed this with @psychocoderHPC already. `HDINLINE` resolves to some `[[attribute]]` in some situations which is required to come first on the line. I've put it after the `constexpr`. Didn't fail with my test compiler on HAL but apparently some `clang` versions are not quite happy. Should be fine now.

ikbuibui:
> > Hopefully removing HDINLINE from the constructors, assignment operators should fix the errors popping up in the CI
> 
> Discussed this with @psychocoderHPC already. `HDINLINE` resolves to some `[[attribute]]` in some situations which is required to come first on the line. I've put it after the `constexpr`. Didn't fail with my test compiler on HAL but apparently some `clang` versions are not quite happy. Should be fine now.

Ah, good to know.

ikbuibui (approved):

PR #5286, opened 2025-02-19
guard against rate cache equal to zero

BrianMarre:
fixes a division by zero in the choose transition kernels if the rate cache entry for the state happens to be zero.

- [x] requires PR #5284 to be merged first
- [x] requires PR #5285 to be merged first
- [x] needs to be rebased to dev

### DIFF
--- a/include/picongpu/particles/atomicPhysics/kernel/ChooseTransition_Autonomous.kernel
+++ b/include/picongpu/particles/atomicPhysics/kernel/ChooseTransition_Autonomous.kernel
@@ -157,15 +157,25 @@ namespace picongpu::particles::atomicPhysics::kernel
                     // get random number
                     float_X const r = rngGeneratorFloat();
 
-                    // compare to cumulative sum of normalized transition rates to find choosen transition
+                    auto const stateTotalLossRate = rateCache.rate(
+                        u32(s_enums::ChooseTransitionGroup::autonomousDownward),
+                        atomicStateCollectionIndex);
 
+                    if(stateTotalLossRate == 0._X)
+                    {
+                        ion[processClass_] = u8(s_enums::ProcessClass::noChange);
+                        // no need to set ion[transitionIndex_] since already uniquely known by processClass = noChange
+                        // no-change transitions are not bin based therefore we don't set a bin, old values are ignored
+                        ion[accepted_] = true;
+                        return;
+                    }
+
+                    // compare to cumulative sum of normalized transition rates to find choosen transition
                     float_X cumSum = 0._X;
                     for(uint32_t transitionID = 0u; transitionID < numberTransitionsDown; ++transitionID)
                     {
-                        cumSum += transitionDataBox.rate(transitionID + startIndexTransitionBlock)
-                            / rateCache.rate(
-                                u32(s_enums::ChooseTransitionGroup::autonomousDownward),
-                                atomicStateCollectionIndex);
+                        cumSum
+                            += transitionDataBox.rate(transitionID + startIndexTransitionBlock) / stateTotalLossRate;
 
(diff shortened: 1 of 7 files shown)


ikbuibui (dismissed):
Looks okay. A possible suggestion is setting the ion properties for the case of ``rate = 0`` can be moved into a function similar to ``updateIon``. Maybe even an overload. Upto you.

include/picongpu/particles/atomicPhysics/kernel/ChooseTransition_Autonomous.kernel
```diff
@@ -157,15 +157,25 @@ namespace picongpu::particles::atomicPhysics::kernel
...
+                        atomicStateCollectionIndex);
+
+                    if(stateTotalLossRate == 0._X)
+                    {
+                        ion[processClass_] = u8(s_enums::ProcessClass::noChange);
+                        // no need to set ion[transitionIndex_] since already uniquely known by processClass = noChange
+                        // no-change transitions are not bin based therefore we don't set a bin, old values are ignored
+                        ion[accepted_] = true;
+                        return;
+                    }
```
chillenzer: Should this be extracted into helper method?

include/picongpu/simulation/stage/AtomicPhysics.x.cpp
```diff
@@ -554,7 +554,7 @@ namespace picongpu::simulation::stage
 
                 // pressure ionization loop, ends when no ion in unbound state anymore
                 bool foundUnbound = true;
-                do
+                while(foundUnbound)
```
chillenzer: This is a change in functionality. Why don't I need to do the rest anymore?

chillenzer (commented):
I go with @ikbuibui and not yet approve until that question is answered.

include/picongpu/particles/atomicPhysics/kernel/ChooseTransition_Autonomous.kernel
BrianMarre: probably yes, I will refactor it after i return from vacation.

include/picongpu/simulation/stage/AtomicPhysics.x.cpp
psychocoderHPC: `do`-`while` should be preferred above `while` because it says that the body is performed at least once independent of the initial condition.
In principle the code change is not change anything but the old code was clearer compared to the current change.

include/picongpu/particles/atomicPhysics/kernel/ChooseTransition_Autonomous.kernel
BrianMarre: done

ikbuibui (dismissed):

chillenzer (dismissed):

include/picongpu/particles/atomicPhysics/kernel/UpdateIon.hpp
```diff
@@ -0,0 +1,64 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/defines.hpp"
+#include "picongpu/particles/atomicPhysics/ConvertEnum.hpp"
+
```
ikbuibui, line 25:
```suggestion
#include "picongpu/particles/atomicPhysics/enums/ProcessClass.hpp"
```

BrianMarre: fixed

ikbuibui (approved):

Issue #5103, opened 2024-09-11
Conventions around `snakemake`


chillenzer:
In closing [this PR](https://github.com/ComputationalRadiationPhysics/picongpu/pull/5082#issuecomment-2343515776), I had a look at the Snakemake workflows [here](https://github.com/ComputationalRadiationPhysics/picongpu/tree/64671e69a64071ac5f93b5cbf10ab81d69a79d9f/share/picongpu/examples/LaserWakefield/lib/python/snakemake) and there are a few things I stumbled across:

1. There is a [best practices page](https://snakemake.readthedocs.io/en/stable/snakefiles/best_practices.html) in the Snakemake docs suggesting amongst other things to use
    - their linter `snakemake --lint`
    - their formatter [`snakefmt`](https://github.com/snakemake/snakefmt)
    - and a [standardised directory structure](https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html#distribution-and-reproducibility).

2. Snakemake provides its own [environment management/deployment](https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html#integrated-package-management) as well as [batch system abstraction](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#mpi-support) (as a plugin). This is functionality that significantly overlaps with `tbg` and both of these are delicate issues in the fragile and diverse HPC ecosystems we run in.

In the current stage, our usage does not conform to any of these standards. If our strategic decisions imply to professionalise the use of Snakemake, I think there would be value in moving more into that direction. However, my personal experience is that these things are awfully fiddly to setup on a cluster and their adoption would create tensions with `tbg`. (There are, of course, some low-hanging fruits like the adoption of the directory structure and formatter, e.g., in `pre-commit`.)

This issue is supposed to spark a discussion and trigger a conscious decision about such things. It is by no means intended as a suggestion to implement the above ideas. Involved in this discussion should probably be at least @psychocoderHPC, @PrometheusPi, @JessicaTiebel.

JessicaTiebel:
Thanks for looking into the Snakemake workflow and giving input.

For 1.:
To be honest I wasn't aware of these things. Having a code quality checker, a formatter and standardised directory structure sounds like a very good idea and on the first glance also doesn't seem to be super much work. Thanks for bringing this up.

For 2.:
[...]

PR #5288, opened 2025-02-20
Harden cut command in Snakefiles

chillenzer:
Found this bug with Arthur. Hemera just hit the next digit in job ids and these `cut` commands were not up to that challenge. The one for LSF is an educated guess from what the internet says the output should look like. I think I've never run on an LSF cluster.

### DIFF
--- a/share/picongpu/examples/LaserWakefield/lib/python/snakemake/Snakefile_LSF
+++ b/share/picongpu/examples/LaserWakefield/lib/python/snakemake/Snakefile_LSF
@@ -77,13 +77,13 @@ rule simulate:
             bsub {input}/tbg/submit.start > simulated/job_id_{params.name}.txt
         fi
 
-        job_id=$(cut -c 6-12 simulated/job_id_{params.name}.txt)
+        job_id=$(cut -d'<' -f2 simulated/job_id_{params.name}.txt | cut -d'>' -f1)
         status=$(bjobs -noheader -o stat:4 $job_id)
 
         # if job isn't running or pending -> restart job
         if ! [[ "$status" == "PEND" || "$status" == "RUN " ]]; then
             bsub {input}/tbg/submit.start >  simulated/job_id_{params.name}.txt
-            job_id=$(cut -c 6-12 simulated/job_id_{params.name}.txt)
+            job_id=$(cut -d'<' -f2 simulated/job_id_{params.name}.txt | cut -d'>' -f1)
             status=$(bjobs -noheader -o stat:4 $job_id)
         fi
         # wait till job is finished
--- a/share/picongpu/examples/LaserWakefield/lib/python/snakemake/Snakefile
+++ b/share/picongpu/examples/LaserWakefield/lib/python/snakemake/Snakefile
@@ -70,13 +70,13 @@ rule simulate:
             sbatch {input}/tbg/submit.start > simulated/job_id_{params.name}.txt
         fi
 
-        job_id=$(cut -c 21-27 simulated/job_id_{params.name}.txt)
+        job_id=$(cut -d' ' -f4 simulated/job_id_{params.name}.txt)
         status=$(squeue --jobs $job_id -o "%2t" | tail -n 1 )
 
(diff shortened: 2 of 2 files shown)


ikbuibui (commented):
Moving from characters to fields looks okay, but without looking at the input file I cannot say anything about correctness. If this has been tested as working, its okay for me

share/picongpu/examples/LaserWakefield/lib/python/snakemake/Snakefile_LSF
```diff
@@ -77,13 +77,13 @@ rule simulate:
             bsub {input}/tbg/submit.start > simulated/job_id_{params.name}.txt
         fi
 
-        job_id=$(cut -c 6-12 simulated/job_id_{params.name}.txt)
+        job_id=$(cut -d'<' -f2 simulated/job_id_{params.name}.txt | cut -d'>' -f1)
```
JessicaTiebel, line 80: Arthur told me, that the `cut -d' ' -f4` command worked on hemera. Why did you choose `cut -d'<' -f2` here? What is the difference?

JessicaTiebel (commented):
Thanks for fixing this bug, during my absence! Arthur already told me that it works and it looks good to me. For the LSF case I just don't understand why you used the `cut` command with other settings. You probably have a reason, if you let me know I will approve this PR. 😃

share/picongpu/examples/LaserWakefield/lib/python/snakemake/Snakefile_LSF
chillenzer: The `-d` sets the delimiter (to whitespace resp. `<`/`>` in this case) and the `-f` selects the "field", i.e., the colum with respect to this delimiter (indexing starts with 1). So, `cut -d' ' -f4` means: Split the string with respect to whitespaces and select the fourth field resulting from that.

The LSF version: My best guess was that the output from `bsub` is as shown in [this stackoverflow answer](https://stackoverflow.com/questions/44716779/lsf-job-submission-stdout-stderr-redirection/44741527#44741527). The two `cut`s split the string into 
```
Full: "Job <930> is submitted to default queue <normal>."
        1  ^          2                        ^   3
First cut: "XXX> ..."
           1   ^   2
Second cut: "XXX"
```
Could have been a regex but this was faster.

chillenzer:
The slurm version has been tested on hemera. LSF is untested.

share/picongpu/examples/LaserWakefield/lib/python/snakemake/Snakefile_LSF
JessicaTiebel: Thank you! This seems very reasonable.

JessicaTiebel:
I wrote the LSF version for the Summit cluster. As far as I know, Summit is now decommissioned and I don't know if any other relevant cluster is using LSF. So testing is difficult.

JessicaTiebel (approved):

share/picongpu/examples/LaserWakefield/lib/python/snakemake/Snakefile_LSF
psychocoderHPC: why we not use a more robust search method via sed and search simply the first number in the string

```
sed -n 's/.*<\([0-9]*\)>.*/\1/p'
```

chillenzer: Offline discussion: This is almost precisely as robust as the one above. There are a number of formatting changes that would make either version fail. Robustness would only be achieved by using correct cluster execution from Snakemake instead of hand-rolling a fragile solution here (and even then you might have a bad experience in some scenarios). This is a sufficient, slightly more robust hot fix and re-running the CI is probably not worth it.

chillenzer: See #5103

psychocoderHPC (approved):

PR #5291, opened 2025-02-21
Updates Frontier profile after software update

BeyondEspresso:
This PR updates the Frontier profiles after OLCF's recent updates of the software environment.


chillenzer (approved):
Can't really say much. But it looks like reasonable changes and I trust that this is tested.

PrometheusPi:
We will test the new changes today with the hackathon.

PR #5278, opened 2025-02-14
implementing phase space plugin in picmi

mafshari64:

PR #5292, opened 2025-02-21
Add picmi custom templates

chillenzer:
Based on work of @BrianMarre and @mafshari64: This PR adds the custom templates that are necessary for the custom user input (already present in the PICMI laser wakefield example) to actually render to something.

More detail: The [laser_wakefield example](https://github.com/ComputationalRadiationPhysics/picongpu/blob/28825ebdf93fb187ecabe173323848291a6cb54e/share/picongpu/pypicongpu/examples/laser_wakefield/main.py#L131) already contains custom user input on the front-end level. But current mainline does not include the custom templates such that this additionally provided information is actually used for anything. This PR adds custom template files that allow this information to be used and as such actually adding value to the corresponding code block in the `main.py` file. These files existed in an outdated form [here](https://github.com/mafshari64/Custom_template_PICMI_LWFA/tree/main) already and I've updated them to compile with the latest dev.

Also: There's one autoformatting commit and one sentence added to the docs. If you only want to see the changes, restrict your diff to the first two commits.

@ikbuibui: Please check if the C++ side of things looks up-to-date. In particular, I've replaced a few instances of `SI::...` with the new `sim.si...`. Maybe you find more (or mistakes in the those).

@BrianMarre: Please check for adequate use of PICMI-related things. I've also added a tiny piece of functionality to CustomUserInput, viz. that it converts bools to their lower-case string representation. This is better aligned with C++.

@mafshari64: Please check that I didn't do any content-wise alterations during my refactoring.

### DIFF
--- a/docs/source/usage/picmi/custom_template.rst
+++ b/docs/source/usage/picmi/custom_template.rst
@@ -20,6 +20,8 @@ Fundamentally there are three different approaches to custom template, differing
 - variable custom templates utilising custom user input
 
 To get started take a look at the examples below, have a look at :ref:`step-by-step guide <step_by_step>`, or at the detailed :ref:`rendering example <pypicongpu-translation-example-boundingbox>` in the developer documentation.
+Alternatively, you can jump right in and have a look at the `laser_wakefield example <https://github.com/ComputationalRadiationPhysics/picongpu/tree/dev/share/picongpu/pypicongpu/examples/laser_wakefield>`
+utilising custom user input and custom templates.
 
 Step-by-Step Guide
 ------------------
--- a/lib/python/picongpu/pypicongpu/customuserinput.py
+++ b/lib/python/picongpu/pypicongpu/customuserinput.py
@@ -59,6 +59,19 @@ class InterfaceCustomUserInput(RenderedObject, pydantic.BaseModel):
         raise NotImplementedError("Abstract Interface only!")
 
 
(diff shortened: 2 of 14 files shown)


mafshari64:
Hi, I took a look. For now, it is fine, I hope we soon change the main.py when we add plugins to the PICMI standard.

mafshari64 (approved):
Hi, from my side it is fine for now.

chillenzer:
Konrad found that this is not yet compiling. (Strangely, because my checks worked. But looks like a real mistake on my side.) Undrafting when it's working.

chillenzer:
Works now.

chillenzer:
PS: I committed with `ci: no-compile` because I don't think that anything related to this is actually checked in the CI currently.

BrianMarre:
The CI fails because there somehow already exists a setup folder in the [test compile of the lwfa example](https://github.com/ComputationalRadiationPhysics/picongpu/blob/c7163a7c113a3116daed66e00c530f075fff7143/test/python/picongpu/compiling/compileexamples.py#L21C28-L21C39), we probably need to explicitly remove the folder before we start the test.

chillenzer:
This is already outdated again after having merged #5278 which is probably symptomatic for #5293. Setting this to draft until we've fixed this.

chillenzer:
The basis has changed dramatically such that is better done from scratch again.

PR #5294, opened 2025-02-21
Add openPMD to PICMI

chillenzer:
@BrianMarre This is a first sketch of how openPMD output could look like. Submitting this draft early, so you can have a say before I sink too much time into this. It's not actually in PICMI yet but mostly just the boilerplate of pypicongpu. Planned next:

- [ ] Go through the parameter list and check if any of the types would be worth an own class. In particular, I was thinking about using `pathlib.Path` instead of strings but that doesn't seem to be standard in the codebase yet. Also some enums could help for `ext` and others that come from a closed set.
- [ ] Of course, remove the hardcoded `openPMD` in `Simulation`.
- [ ] Propagate this interface into PICMI somehow.

The last bullet is definitely the interesting one. There's [diagnostics](https://picmi.readthedocs.io/en/latest/standard/diagnostics.html) defined already but I have not looked at them in enough detail to figure out if they align with our perspective on this.

### DIFF
--- a/lib/python/picongpu/pypicongpu/output/openPMD.py
+++ b/lib/python/picongpu/pypicongpu/output/openPMD.py
@@ -5,6 +5,7 @@ Authors: Julian Lenz
 License: GPLv3+
 """
 
+from typing import Optional
 from .. import util
 from ..rendering import RenderedObject
 
@@ -16,17 +17,44 @@ class OpenPMD(RenderedObject):
     """
     Class to configure openPMD output.
 
-    This class requires a period (in time steps) and file name.
+    This class requires a period (in time steps) and will enable OpenPMD output.
     """
 
     period = util.build_typesafe_property(int)
     """period to print data at"""
 
-    file_name = util.build_typesafe_property(str)
-    """file name for openPMD output"""
+    source = util.build_typesafe_property(str)
+    """data sources and filters to dump"""
 
-    file_extension = util.build_typesafe_property(str)
-    """file extension for openPMD output (e.g., 'bp', 'h5')"""
+    range_ = util.build_typesafe_property(str)
+    """contiguous range of cells per dimension to dump"""
+
+    file_name = util.build_typesafe_property(str)
+    """file name for OpenPMD output"""
+
+    ext = util.build_typesafe_property(str)
+    """openPMD filename extension"""
+
+    infix = util.build_typesafe_property(Optional[str])
+    """openPMD filename infix"""
+
+    json_ = util.build_typesafe_property(Optional[str])
+    """backend-specific parameters for openPMD backends in JSON format"""
+
+    jsonRestart = util.build_typesafe_property(Optional[str])
+    """backend-specific parameters for openPMD backends in JSON format for restarting from a checkpoint"""
+
+    dataPreparationStrategy = util.build_typesafe_property(Optional[str])
+    """strategy for preparation of particle data"""
+
+    toml = util.build_typesafe_property(Optional[str])
+    """configure the openPMD plugin via a TOML file"""
+
+    particleIOChunkSize = util.build_typesafe_property(Optional[int])
+    """particle data will be written in chunks of the given size"""
+
+    writeAccess = util.build_typesafe_property(Optional[str])
+    """openPMD Access mode for file writing"""
 
     def check(self) -> None:
         """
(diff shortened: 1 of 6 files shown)


lib/python/picongpu/pypicongpu/output/openPMD.py
```diff
@@ -0,0 +1,96 @@
...
+    Class to configure openPMD output.
+
+    This class requires a period (in time steps) and will enable OpenPMD output.
+    """
+
+    period = util.build_typesafe_property(int)
+    """period to print data at"""
+
+    source = util.build_typesafe_property(str)
+    """data sources and filters to dump"""
```
BrianMarre, line 27: I am not really happy with the representation of data sources as strings, we should use objects to define the sources, and then generate a matching `fileOutput.param` and cfg block based on them,

... but I would be willing to do that in a later PR.

lib/python/picongpu/pypicongpu/output/openPMD.py
```diff
@@ -0,0 +1,96 @@
...
+@typeguard.typechecked
+class OpenPMD(RenderedObject):
+    """
+    Class to configure openPMD output.
+
+    This class requires a period (in time steps) and will enable OpenPMD output.
+    """
+
+    period = util.build_typesafe_property(int)
+    """period to print data at"""
```
BrianMarre, line 24: As far as I know we support the general [PIConGPU period syntax](https://picongpu.readthedocs.io/en/latest/usage/plugins.html#period-syntax), for the openPMD plugin. This should be reflected in the PyPIConGPU description, probably object oriented?

lib/python/picongpu/pypicongpu/output/openPMD.py
```diff
@@ -0,0 +1,96 @@
...
+    """file name for OpenPMD output"""
+
+    ext = util.build_typesafe_property(str)
+    """openPMD filename extension"""
+
+    infix = util.build_typesafe_property(Optional[str])
+    """openPMD filename infix"""
+
+    json_ = util.build_typesafe_property(Optional[str])
+    """backend-specific parameters for openPMD backends in JSON format"""
```
BrianMarre, line 42: As far as I know json and toml both describe the same thing, and should not be used at the same time, we should reflect that by having just one configuration entry with different config classes types depending on the description used.

BrianMarre (changes requested):
a good first start, but there are a few general problems I see so far.
- we currently demand a configuration of every plugin, but usually a user want to configure just the plugins he actually uses, we probably should convert the output specification to a list of plugins, with each each plugin having it's own object which is appended to the list. For rendering something like that see [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/share/picongpu/pypicongpu/template/include/picongpu/param/density.param.mustache)
-  please include test for the new functionality, see [here](https://github.com/ComputationalRadiationPhysics/picongpu/tree/dev/test/python/picongpu/quick/pypicongpu/output)
- for PICMI I would start with just defining a picongpu specific user friendly interface for PyPIConGPU, we will figure out the PICMI standard later.

lib/python/picongpu/pypicongpu/output/openPMD.py
chillenzer: Yes, I was shortly thinking about actually using slice objects for this. But compared to how nice slicing is within `[...]`, creating a bare slice object is pretty ugly. One could hack together a factory that uses `__get_item__` to create nice slice-like objects for us. But no matter how elegant, I'm not sure that would be intuitive.

lib/python/picongpu/pypicongpu/output/openPMD.py
chillenzer: Well caught!

lib/python/picongpu/pypicongpu/output/openPMD.py
BrianMarre: I would suggest that we use a basic object with attribute start, stop, and period for now. We can make it prettier later.
Also be aware that it is possible to combine several `start:stop:period` blocks as a comma separated list in the same period argument in PIConGPU, my be use a list of Period specifier objects?.

lib/python/picongpu/pypicongpu/output/openPMD.py
chillenzer: _Offline discussion_: The mutual exclusion is implemented. The unification has more benefits closer to the user interface.

chillenzer:
Superseded by #5353

PR #5297, opened 2025-02-28
Update docs binary_collisions.rst

FilipO28555:
One needs those includes.

@pordyna


pordyna:
I think I meant it more like an example on how to setup this part of the file, the `default.param` has the includes. But sure we can include it also here.

pordyna (approved):

psychocoderHPC (approved):

PR #5298, opened 2025-03-03
Enable 2D simulations with the FromOpenPMDPulse profile

fabidie:
In 2D simulations, the central slice of the input field parallel to propagation and polarisation direction will be extracted and serve as incident field input. The whole 3D input field will still be loaded to the GPUs. Until I implement field-slice-loading, I recommend to trim the extent of the omitted transverse axis of the input field to $\geqslant 2$, centered around the central slice, to reduce GPU memory occupancy.

### DIFF
--- a/include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.def
+++ b/include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.def
@@ -1,4 +1,4 @@
-/* Copyright 2024-2024 Fabia Dietrich
+/* Copyright 2024-2025 Fabia Dietrich
  *
  * This file is part of PIConGPU.
  *
@@ -17,7 +17,7 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
-#if(ENABLE_OPENPMD == 1) && (SIMDIM == DIM3)
+#if(ENABLE_OPENPMD == 1)
 
 #    pragma once
 
@@ -42,14 +42,21 @@ namespace picongpu
                          * saving the resulting E-field chunk to openPMD.
                          *
                          * This field chunk will be stored at timestep 0 of the simulation on every used device, which
-                         * is why one has to increase the reserved GPU memory size in memory.param. Otherwise, the
-                         * simulation could run into memory issues.
+                         * is why one has to increase the reserved GPU memory size in memory.param by about the size of
+                         * the openPMD file. Otherwise, the simulation will run into memory issues.
+                         *
+                         * For 2D simulations, the input still has to be 3D. Then, only the central slice parallel to
+                         * propagation and polarisation direction of the input field will be used in the simulation.
+                         * To reduce GPU memory occupancy, it is recommended to trim the extent of the omitted
+                         * transverse axis to a minimum (but still including the slice that shall feed the simulation).
+                         * In case of working with INSIGHT data, this can be achieved during the data pre-procession
+                         * with 'preparingInsightData.py' with the parameters 'crop_x' or 'crop_y'.
                          *
                          * It is recommended to use a full path to make it independent of how PIConGPU is launched.
                          * Relative paths require consistency to the current directory when PIConGPU is started
                          * with tbg and the standard .tpl files, relative to the resulting simOutput directory.
                          */
-                        static constexpr char const* filename = "/path/to/file%T.h5";
+                        static constexpr char const* filename = "/path/to/file%T.bp5";
 
                         /** Iteration inside the file (where the whole field is stored in, not related to the current
                          * simulation time iteration)
--- a/include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
+++ b/include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
@@ -17,7 +17,7 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
-#if(ENABLE_OPENPMD == 1) && (SIMDIM == DIM3)
+#if(ENABLE_OPENPMD == 1)
 
 #    pragma once
 
@@ -54,7 +54,8 @@
  * - get rid of the 'wrong' transformation from time to space (z = c*t) of the
  *   longitudinal axis by using several iterations inside the openPMD file
  *   instead of just one
- * - allow 2D simulations
+ * - in 2D simulations, load only the necessary slice into the simulation, not
+ *   the whole 3D chunk, and provide the index of the used slice as user parameter
  */
 
 namespace picongpu
@@ -126,6 +127,14 @@ namespace picongpu
                         PMACC_CASSERT_MSG(
                             _error_laser_polarisation_direction_vector_must_be_orthogonal_to_propagation_direction____check_your_incidentField_param_file,
                             (dotPropagationPolarisation > -0.0001) && (dotPropagationPolarisation < 0.0001));
+
+                        // Check that z is not used as proopagation or polarisation direction in 2D simulations
+                        PMACC_CASSERT_MSG(
+                            _error_laser_propagation_direction_cannot_be_z_in_2D_simulations____check_your_incidentField_param_file,
+                            ((simDim == DIM2) && (DIR_Z > -0.0001) && (DIR_Z < 0.0001)) || (simDim == DIM3));
+                        PMACC_CASSERT_MSG(
+                            _error_laser_polarisation_direction_cannot_be_z_in_2D_simulations____check_your_incidentField_param_file,
+                            ((simDim == DIM2) && (POL_DIR_Z > -0.0001) && (POL_DIR_Z < 0.0001) || (simDim == DIM3)));
 
                         /** Time delay
                          *
@@ -507,11 +516,15 @@ namespace picongpu
                          */
                         HDINLINE float_X getValueE(floatD_X const& totalCellIdx) const
                         {
-                            auto const posPIC = totalCellIdx * sim.pic.getCellSize(); // position in simulation volume
+                            // position in simulation volume
+                            floatD_X const posPIC = totalCellIdx * sim.pic.getCellSize().shrink<simDim>();
 
                             DataSpace<3u> internalAxisIndex = getInternalAxisIndex();
 
-                            // check whether we are outside the field chunk extent
+                            /* Check whether we are outside the field chunk extent.
+                             * Omits d = 2 in the 2D case. There, internalAxisIndex[2] = 2 holds automatically, hence
+                             * d = 2 implies the second transverse direction, which is irrelevant in 2D simulations.
+                             */
                             for(uint32_t d = 0u; d < simDim; d++)
                             {
                                 // return zero if transversal simulation window exceeds chunk extent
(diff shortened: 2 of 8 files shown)


include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.def
```diff
@@ -41,6 +41,9 @@ namespace picongpu
                          * data, propagate it to the desired position and transform it into the time domain before
                          * saving the resulting E-field chunk to openPMD.
                          *
+                         * For 2D simulations, the input still has to be 3D. Then, only the central slice parallel to
+                         * propagation and polarisation direction of the input field will be used in the simulation.
+                         *
                          * This field chunk will be stored at timestep 0 of the simulation on every used device, which
                          * is why one has to increase the reserved GPU memory size in memory.param. Otherwise, the
                          * simulation could run into memory issues.
```
chillenzer: _I know it's not technically something you've touched for this PR but while you're at it:_
By how much? Should we add an estimate? (Likely something like "... increase the reserved GPU memory size by about the size of the openPMD file..."?)

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.def
```diff
@@ -49,7 +52,7 @@ namespace picongpu
                          * Relative paths require consistency to the current directory when PIConGPU is started
                          * with tbg and the standard .tpl files, relative to the resulting simOutput directory.
                          */
-                        static constexpr char const* filename = "/path/to/file%T.h5";
+                        static constexpr char const* filename = "/path/to/file%T.bp5";
```
chillenzer, line 59: This doesn't sound like a location that is consistently valid in arbitrary environments. Yet, it is not in a `.param` file, so not something a user should change or even touch at all. Not sure if I'm misunderstanding something here but did you mean to put this into a `.param` file?

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
```diff
@@ -127,6 +128,14 @@ namespace picongpu
...
                             (dotPropagationPolarisation > -0.0001) && (dotPropagationPolarisation < 0.0001));
 
+                        // Check that z is not used as proopagation or polarisation direction in 2D simulations
+                        PMACC_CASSERT_MSG(
+                            _error_laser_propagation_direction_cannot_be_z_in_2D_simulations____check_your_incidentField_param_file,
+                            (DIR_Z > -0.0001) && (DIR_Z < 0.0001));
+                        PMACC_CASSERT_MSG(
+                            _error_laser_polarisation_direction_cannot_be_z_in_2D_simulations____check_your_incidentField_param_file,
+                            (POL_DIR_Z > -0.0001) && (POL_DIR_Z < 0.0001));
+
```
chillenzer: These seem to be specific to 2D but I can't seem to figure out how they are guarded against being triggered in 3D. Depending on the situation, please do one of the following:
1. Explain to me how they are guarded against firing in a 3D scenario.
2. Explain to me why they are also valid in a 3D scenario.
3. Guard them against getting triggered in a 3D scenario.

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
```diff
@@ -507,11 +516,17 @@ namespace picongpu
                          */
                         HDINLINE float_X getValueE(floatD_X const& totalCellIdx) const
                         {
-                            auto const posPIC = totalCellIdx * sim.pic.getCellSize(); // position in simulation volume
+                            // position in simulation volume
+                            floatD_X posPIC;
+                            for(uint32_t d = 0u; d < simDim; d++)
+                                posPIC[d] = totalCellIdx[d] * sim.pic.getCellSize()[d];
```
chillenzer: Just for my understanding: Did this need to change because one of `totalCellIdx` or `sim.pic.getCellSize()` is always 3D and we are explicitly initialising a 2D vector here?

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
```diff
@@ -544,9 +559,11 @@ namespace picongpu
                             float_X const transvAxisDirection
                                 = pmacc::math::cross(getDirection(), getPolarisationVector()).sumOfComponents();
 
-                            // find the position in the field data chunk corresponding to totalCellIdx and
-                            // timeOriginPIC
-                            float3_X idxClosestRaw; // raw = not yet rounded to integers
+                            /* Find the position index in the field data chunk corresponding to totalCellIdx and
+                             * timeOriginPIC. Raw = not yet rounded to integers.
+                             * Is 2D in 2D simulations and refers then to the position on the 2D central slice.
```
chillenzer:
```suggestion
                             * Is 2D in 2D simulations and refers to the position on the 2D central slice in this case.
```

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
```diff
@@ -594,52 +611,76 @@ namespace picongpu
...
+                            // the other 7 (in 3D) / 3 (in 2D) nearest neighbour indices still have to be found
+                            DataSpace<3u> idxShift; // shift from idxClosest to idxNext
                             DataSpace<3u> idxNext; // nearest neighbour index
                             float3_X weight;
-                            for(uint32_t d = 0u; d < 3u; d++)
+                            for(uint32_t d = 0u; d < simDim; d++)
                             {
-                                if(idxClosest[d] == 0) // to avoid border problems
-                                    idxShift[d] = 1;
+                                if(idxClosest[d] == 0u) // to avoid border problems
```
chillenzer: In order to avoid that this comment ever shifts to the next line and becomes detached from what it points to:
```suggestion
                                // to avoid border problems
                                if(idxClosest[d] == 0u)
```

Also, just for my information: What is the edge case you're guarding against here? It looks like the `else if` branch could never be taken (assuming `pos` to be positive), so this explicit branch doesn't seem to be necessary. And if there is a problem: Why don't we have to guard the "upper" edge case as well?

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
```diff
@@ -594,52 +611,76 @@ namespace picongpu
...
-                                }
-                                else
+                                if constexpr(simDim == DIM3)
                                 {
-                                    idxNext[d + 1] = idxClosest[d + 1] + idxShift[d + 1];
-                                    ones[d + 1] = 0.0_X;
+                                    if(d == 2)
+                                    {
+                                        idxNext[0] = idxClosest[0] + idxShift[0];
+                                        ones[0] = 0.0_X;
```
chillenzer: This sounds like the kind of naming that is bound to make the next debugging person quit their job: _"The bug only ever happens on this line that could never be encountered because `ones[0]` will certainly always be one. What else could it possibly be?"_ Writing such a line should be the moment anyone sits back and rethinks their naming scheme...

chillenzer (changes requested):
Looks good! Most of the things are for my information or stylistic.

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.def
fabidie: Good idea. I added it.

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.def
fabidie: This is the location to and the name of the openPMD file (so h5 or bp5. I changed it to bp5 for consistency reasons with the jupyter notebook for preparing INSIGHT measurements). In my simulations, I usually put this file in the directory from where I start my simulations (but this can be changed depending on the user preferences). If you want, I can specify that so that it becomes more clear.

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
fabidie: It's 3 😅

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
fabidie: Yes, exactly

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
fabidie: The upper edge case is included in the `else if` branch. And both the `else if` and the `else` brach are the standard scenario used in 99.9% of the cases. `pos` is $\geqslant 0$ as you said, but it is `float_X` and `pos[d] - static_cast<float_X>(idxClosest[d])` can be both greater or smaller than 0, depending on the position of `pos` with respect to the grid points of `dataBox`. I hope I expressed myself somehow understandable. If not, please ask again!

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
fabidie: You're right. I renamed it to `invertWeight`.

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.def
BrianMarre: Incident field profiles classes are supposed to be defined by the user in his `incidentField.param`-file. The profile definition here is just an interface definition, see documentation [here](https://picongpu.readthedocs.io/en/latest/usage/param/core.html#incidentfield-param).

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
BrianMarre: in picongpu the cellSize is always 3D

psychocoderHPC: this can be written as ... too

```C++
floatD_X posPIC = totalCellIdx * sim.pic.getCellSize().shrink<simDim>();
```

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
```diff
@@ -594,52 +611,79 @@ namespace picongpu
...
+                                    idxShift[d] = -1u;
+                                // inside dataBox
                                 else
-                                    idxShift[d] = 1;
+                                    idxShift[d] = 1u;
 
                                 weight[d] = pmacc::math::abs(static_cast<float_X>(idxClosest[d]) - pos[d]);
                             }
 
+                            if(simDim == DIM2)
```
psychocoderHPC: can be 

```C++
if constexpr(simDim == DIM3)
```

psychocoderHPC (dismissed):
tiny change requested

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
chillenzer: > this can be written as ... too
> 
> ```c++
> floatD_X posPIC = totalCellIdx * sim.pic.getCellSize().shrink<simDim>();
> ```

I prefer this version.

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.hpp
chillenzer: Well done! Coming up with a decent name myself was the part I was struggling with. Criticizing is so much easier than being constructive. ;D

include/picongpu/fields/incidentField/profiles/FromOpenPMDPulse.def
chillenzer: Okay, so what's the workflow then? Something like this?
```C++
struct MyPulse : FromOpenPMDPulse {
    // ...
    static constexpr char const* filename = "/my/custom/path/to/file%T.bp5";
    // ...
};

using XMin = MyPulse;
```

BrianMarre: actually currently the usual workflow is to not inherit from the interface definition at all, just define all attributes and pass an independent type that happens to have all required members as defined by the interface. But your approach is the intention.

P.S.: c++ concepts should allow us to enforce inheritance from the interface and we should probably switch to them.

chillenzer (approved):

PR #5304, opened 2025-03-10
Fix installation of mallocMC and alpaka

### LINKED ISSUE #5303: alpaka and mallocMC files added during install
We might have introduced a minor bug with #5279 and #5280.
When installing PIConGPU, there is now an addition `lib64` directory that contains `cmake/alpaka/` and `cmake/mallocMC-3.0.0/` directories and files that we should not include here.

chillenzer:
Closes #5303


PrometheusPi (approved):

PrometheusPi:
Tested it and it works :+1: as expected. Thanks @chillenzer for the fix.

PR #5305, opened 2025-03-10
Update TWTSTight-interface

BeyondEspresso:
For historical reasons, i.e. originally to reduce register usage on device, the parameters (`phi`, `beta_0` and `polAngle`) of TWTS were defined as `float_X`. While this is fine for calculating TWTSTight at single precision, this choice limits precision when TWTSTight is used at double precision (currently non-standard usage for development and tests only). With this PR that replaces the `float_X` with the corresponding `float_64` types, the number of significant digits at double precision that agree with the Mathematica reference implementation increases from >=5 digits to >=13 digits.

### DIFF
--- a/include/picongpu/fields/background/templates/twtstight/GetInitialTimeDelay_SI.tpp
+++ b/include/picongpu/fields/background/templates/twtstight/GetInitialTimeDelay_SI.tpp
@@ -1,4 +1,4 @@
-/* Copyright 2014-2024 Alexander Debus
+/* Copyright 2014-2025 Alexander Debus
  *
  * This file is part of PIConGPU.
  *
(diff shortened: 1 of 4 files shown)


chillenzer (dismissed):
Looks fine. Does this have performance implications?

BeyondEspresso:
I don't expect performance implications, but my test case is in the cluster queue. This run should also test, whether the increased precision has physics implications.

PrometheusPi:
@BeyondEspresso The CI fails with a compile error:
```
picongpu/unit/TWTSTight.cpp:62:39: error: use of undeclared identifier 'max'
    T const epsilonScaled = epsilon * max(abs(a), abs(b));
```

chillenzer (approved):

PR #5307, opened 2025-03-11
pyproject.toml

pordyna:
Adds `pyproject.toml` into `lib/python`. With that, you can install the `picongpu` python package into any python environment with `pip install -e <picongpu_src>/lib/python`. You can also install it not in edit mode, but I think it makes more sense for picongpu to include the `-e`.
With this PR there is no need for editing the `PYTHONPATH`, which is quite ugly and defies the whole point of having python envs. 

The dependencies list is taken from all the `requirements.txt` files. At some point (with this PR or after?) we should remove the requirements.txt and only use the `pyproject.toml` specification. 

I also add here a default python `.gitignore`, since pip will create some hidden files in `/lib/pytho` when installing with `-e`, that need to be excluded.  

Everything in `lib/python/synchrotronRadiationExtension` and in `lib/python/test` is ignored. 
## TODO:
- [x] remove requirements.txt (?)
- [x] edit all mentions of `PYTHONPATH` and  `requirements.txt` in the docs.
- [x] remove `PYTHONPATH` from all `*.profile.example`
- [x] probably switch to the new install method in tests / CI, otherwise it will look for missing  `requirements.txt` ?

@psychocoderHPC, should we switch completely at once and remove `requirements.txt`, edit all documentation mentions? 
@chillenzer @BrianMarre @SimeonEhrig can you comment on the possibly needed CI / tests changes?


SimeonEhrig:
# Current state

For the CI, we need to change the way how dependency versions to be tested are configured. At the moment, we have tow custom script, which does the job:
1.  `share/ci/pypicongpu_generator.py`: Reads a given `requirements.txt`, determines all dependency versions to be tested (read the supported version range and check on pypy.org which version exists in this range) and generated for each version combination a separated CI job. Each job has different environment variables, which defines which dependency versions need to be set.
2. `share/ci/pypicongpu.sh`: Is executed in each generated job. It takes the original `requirements.txt`, replace the version ranges for the dependency to be tested by a specific version given by the environment variables and install the modified `requirements.txt`.

# Possible Solution

## Modify the python

Instead reading and modifying `requirements.txt`, we can read and modify the `pyproject.toml`. Changing the read and write functions is not the big deal, but I'm not sure, if we get new problems, because the separated `requirements.txt` are composed in a single `pyproject.toml` now.

## Using an existing test framework

[tox](https://tox.wiki/en/latest/) and [nox](https://nox.thea.codes/en/stable/) should be able to test different dependency versions by a given configuration. Unfortunately I have no experience with it. After a short review, I'm not sure if this solution works with our CI workflow of generating a CI job for each test combination. But if it works, we save the maintenance of the custom scripts. Maybe somebody has experience with it at the center.

chillenzer:
I think this is to be seen in a larger context. We will probably move more towards using the Python interface in general. There's also been discussions about properly modularising and individually packaging the different components. It might even be a way to provide an "installable" version in the sense of making the PICMI interface available as a package and coordinating the whole build and run process via Python. So, this is an interesting attempt and definitely in the right direction. The question is if we should approach this incrementally with this draft being a first step or as a general overhaul completely reorganising our Python approach.

Concretely:
1. I like your approach of how you gathered the dependencies including where they come from. We'd likely want to remove those additional comments at some point but it's certainly nice to have this information at hand for now.
2. I think we could add some more metadata to the `pyproject.toml` but that doesn't need be in this PR.
3. I don't like overblown `.gitignore`s. My philosophy about `.gitignore` is that it should reflect the general things to ignore not individual choices. Example: I'm in favour of ignoring `__pycache__` because there is most likely a workflow in which those pop up sooner or later. I'm also in favour of ignoring some of the temporaries for packaging. I'm sternly against ignoring any editor or tool specific stuff. That's what global `.gitignore`s are for. We'll be busy updating that list whenever the next student arrives with their new and shiny text editor not yet in the list. So my approach would be: If you follow the simplest, most standardised workflow to package and `import` that code, what are the files that get created. Anything else is a liability for the project, I'd say.

chillenzer:
Concerning the possible CI solutions: I think that `nox` looks like a good fit. In fact, for Python-internal dependencies I think spawning a separate runner for each combination is a bit over the top anyways. It will probably significantly reduce the overhead on the runners to have a single job with `nox` handling the dependencies internally.

SimeonEhrig:
> Concerning the possible CI solutions: I think that `nox` looks like a good fit. In fact, for Python-internal dependencies I think spawning a separate runner for each combination is a bit over the top anyways. It will probably significantly reduce the overhead on the runners to have a single job with `nox` handling the dependencies internally.

In an offline discussion with @pordyna, we agreed that he is checking if `nox/tox` is able to manage dependencies like we need it. Most of the examples only shows how to test different python versions, which is the easiest part. If `nox/tox` saves work, we will use it and fuse all jobs in a single CI job. Otherwise I will update the scripts, that they modify the `pyproject.toml` instead the `requirements.txt`.

chillenzer:
Loading from `pyproject.toml`: https://nox.thea.codes/en/stable/tutorial.html#loading-dependencies-from-pyproject-toml-or-scripts
Parametrization of dependencies: https://nox.thea.codes/en/stable/tutorial.html#parametrization

pordyna:
@SimeonEhrig @chillenzer  @psychocoderHPC Since the discussion about dependencies testing is still ongoing, I would suggest adding a very minimal `pyproject.toml` for now, that does not list dependencies, and keeping the  requirements files. So that user installation goes like:
```
pip install -r requirements.txt
pip install -e .
```
We can still do our tests just like before, we could consider also already switching to  `pip install -e .` in the CI. In the meantime, we can figure out an elegant solution with sth like `nox`.

SimeonEhrig:
@pordyna I will present the problem on the RSE meeting on Thursday. Let's see, if somebody can help me.

pordyna:
just a note so I don't forget it. @SimeonEhrig we should probably add https://pyproject-fmt.readthedocs.io/en/latest/ with our pre-commit once we switch

pordyna:
Btw. the old requirements.txt were compiled to one list with https://gist.github.com/pordyna/ff27ce949a8e944fe30252f47b5b831f

pordyna:
Note to myself:
https://github.com/ComputationalRadiationPhysics/picongpu/pull/5378/commits/93e1fa419b8488edb0a85bbc353c73e7c8c8e646
in #5378

pordyna:
Ready for final review @psychocoderHPC @chillenzer @SimeonEhrig, does not include the changes from https://github.com/ComputationalRadiationPhysics/picongpu/pull/5378/commits/93e1fa419b8488edb0a85bbc353c73e7c8c8e646 since it is not merged yet.

chillenzer:
Great! Thanks! I'd suggest that we wait for #5378 because (a) that's higher priority and (b) potential fixes due to the required rebase and conflict resolution would add further non-pertinent baggage to that PR while such changes would match the purpose of this PR pretty well.

PrometheusPi:
@chillenzer #5378 was merged 2 months ago. Can we proceed with the review @pordyna?

pordyna:
@PrometheusPi @SimeonEhrig  @chillenzer, I finished rebasing. Ready for review.

docs/source/usage/python_utils.rst
```diff
@@ -3,6 +3,13 @@
 Python Utilities
 ================
 
+PIConGPU python package includes plugin output processing tools, tools listed below and our :ref:`PICMI <PICMI>` and PyPIConGPU interface.
+It can be installed into your python environment with:
```
chillenzer: Should we stress here again that `-e` is necessary?

lib/python/.gitignore
chillenzer, line 1: That's too bloated. I'd use the following approach:
- Run the supported way(s) of installing it and check what files that generates.
- Compile exactly these files into this `.gitignore`.
- All other user-specific ignores should be put into a (user-)global `.gitignore`s elsewhere.

lib/python/pyproject.toml
```diff
@@ -0,0 +1,50 @@
+# This file is part of PIConGPU.
+# Copyright 2025 PIConGPU contributors
+# Authors: Pawel Ordyna
+# License: GPLv3+
+
+[build-system]
+requires = ["setuptools>=61.0", "wheel"]
+build-backend = "setuptools.build_meta"
+[tool.setuptools]
```
chillenzer:
```suggestion
build-backend = "setuptools.build_meta"

[tool.setuptools]
```

lib/python/pyproject.toml
```diff
@@ -0,0 +1,50 @@
...
+packages=["picongpu"]
+
+[project]
+name = "picongpu"
+# should be always identical with PIConGPU version
+version = "0.9.0.dev"
```
chillenzer, line 16: How do we ensure this?

share/ci/install/pypicongpu.sh
```diff
@@ -41,28 +41,26 @@ cd $CI_PROJECT_DIR
...
-python3 $CI_PROJECT_DIR/share/ci/install/requirements_txt_modifier.py \
-    -i $CI_PROJECT_DIR/lib/python/picongpu/pypicongpu/requirements.txt \
-    -o $MODIFIED_REQUIREMENT_TXT_PYPICONGPU \
-    --ignore_env_args PYPIC_DEP_VERSION_pydantic PYPIC_DEP_VERSION_picmistandard
-
-echo "modified_requirements.txt: "
-cat $MODIFIED_REQUIREMENT_TXT_PICMI
-cat $MODIFIED_REQUIREMENT_TXT_PYPICONGPU
+# install requirements of pyproject_toml_modifier.txt
+pip3 install -r ${CI_PROJECT_DIR}/share/ci/install/pyproject_toml_modifier_requirments.txt
```
chillenzer:
```suggestion
pip3 install -r ${CI_PROJECT_DIR}/share/ci/install/pyproject_toml_modifier_requirements.txt
```
(Of course, also rename the file itself.)

share/ci/install/pypicongpu.sh
```diff
@@ -41,28 +41,26 @@ cd $CI_PROJECT_DIR
...
+# uninstall requirements of pyproject_toml_modifier.txt
+pip3 uninstall -y -r ${CI_PROJECT_DIR}/share/ci/install/pyproject_toml_modifier_requirments.txt
+
+echo "modified pyproject.toml: "
+cat $PYPROJECT_TOML_PATH
 echo ""
 
+# install pypicongpu dependencies
+cd ${CI_PROJECT_DIR}/lib/python/
+pip3 install -e .
```
chillenzer:
```suggestion
pip3 install -e ${CI_PROJECT_DIR}/lib/python/
```

chillenzer (changes requested):
Not yet tested if it actually works but some comments just from looking through it.

chillenzer:
Worked like a charm locally. Cool!

SimeonEhrig:
@pordyna Can you please do a `git commit --amend` and for push again to trigger the CI. The bot is working again.

lib/python/pyproject.toml
pordyna: @psychocoderHPC can we maybe just add it to our release checklist? I believe we had sth like that. Otherwise, we could also try to automate is somehow.

@chillenzer we could also use https://setuptools-scm.readthedocs.io/en/latest/ to read the version from git tags. Not sure what is the behavior if not on a tag? Maybe it would include the hash? Who knows, but this could be actually quite nice! 

Alternatively, sth like this would be possible. The `__version__` could be dynamically defined in the main `__init__.py`. There we could in theory read it from some picongpu file. 
```toml
# ...
[project]
name = "my_package"
dynamic = ["version", "readme"]
# ...
[tool.setuptools.dynamic]
version = {attr = "my_package.__version__"}
```

docs/source/usage/python_utils.rst
pordyna: It wouldn't hurt us, I think :D.  Better safe than sorry.

lib/python/.gitignore
pordyna: done

share/ci/install/pypicongpu.sh
pordyna: done

lib/python/.gitignore
pordyna: Done, can you check if it works for you?

docs/source/usage/python_utils.rst
pordyna: Done

lib/python/pyproject.toml
SimeonEhrig: In bashi, I write the version number in a central file and use it everywhere. In the `pyproject.toml` is described how to this.
https://github.com/alpaka-group/bashi/

In the case of C++, an `#include` or `#embed` should work, to embed the content of the file in a header.

pordyna: So we could do sth like:
```toml
[build-system]
requires = ["setuptools>=80.0", "wheel", "setuptools-scm[simple]>=8"]
build-backend = "setuptools.build_meta"

[tool.setuptools]
packages=["picongpu"]

[tool.setuptools_scm]
root = "../../"
version_scheme = "no-guess-dev"
```
The thing is, all of this does not really make much sense when installing in edit mode (`-e`). So I guess we should fix our picmi paths first.

SimeonEhrig: I have no idea, what this configuration is doing.

pordyna:
@chillenzer I'm now  done with your comments.

lib/python/pyproject.toml
pordyna: @SimeonEhrig sorry somehow your previous answer was not here, when I was writing it. SCM reads the version from the latest git tag, as long as they follow versioning conventions.  The `no-guess-dev`  just configures how it should deal with commits not exactly on a tag.  root is here the relative path to the git root. 

I kinda like this, but still if we install in edit mode, versions are not so important since the code can be changed without updating the version information after installation. 

If we dynamically read version from a central file it changes the __version__ when someone checkouts a new version, but not the package info in pip, this will stay as at the installation moment.  

Tbh no idea what to do with this :D

lib/python/.gitignore
chillenzer: Worked!

chillenzer (approved):

lib/python/pyproject.toml
chillenzer: This is a difficult thing that I have discussed with many people already. No satisfactory answer has ever reached me.

PR #5309, opened 2025-03-11
Fix bugs in and refactor toSlice.hpp

chillenzer:
This PR fixes a bug in `to{Range,Time}Slice`. Despite claims to the contrary in the documentation and their respective docstrings they didn't handle `-1` input but threw an exception. In the process of fixing this, I added unit tests and refactored them by extracting a common function.

### DIFF
--- a/share/picongpu/unit/toSlice.cpp
+++ b/share/picongpu/unit/toSlice.cpp
@@ -1,0 +1,163 @@
+/* Copyright 2025 Julian Lenz
+ *
+ * This file is part of PMacc.
+ *
+ * PMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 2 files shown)


share/picongpu/unit/toSlice.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    SECTION("accept : and , syntax.")
+    {
+        std::vector<std::pair<std::string, std::vector<Slice>>> testCases = {
+            {"", {}},
+            {"1:5", {Slice{1, 5, 1}}},
+            {"1:5:2", {Slice{1, 5, 2}}},
+            {"1:-1", {Slice{1, static_cast<uint32_t>(-1), 1}}},
+            {":", {Slice{0, static_cast<uint32_t>(-1), 1}}},
+            {"::", {Slice{0, static_cast<uint32_t>(-1), 1}}},
```
psychocoderHPC: we should add `::2`, `2::`, `:5:`

psychocoderHPC (changes requested):
could you please add the folowing tests too

psychocoderHPC (approved):

PR #5310, opened 2025-03-11
Add period syntax to PICMI

chillenzer:
This PR adds an interface to the period syntax for plugins to PICMI. In the process of doing so, I've found a bug that is fixed separately in #5309. This PR also restructures our PICMI implementation by adding a separate folder for `diagnostics` as discussed in #5316.

After a number of discussions, I decided to let the semantics of slicing/indexing slightly deviate from Python's interpretation by making it INCLUSIVE on the upper end as well. That is in line with what warpx is doing and feels natural for time intervals where you would expect to cover the whole range from start to end.

Hints for review:
- The main feature is implemented and documented in `lib/python/picongpu/picmi/diagnostics/timestepspec.py`. It forwards to the rather minimalistic PyPIConGPU pendant.
- The interface has an extensive test suite that should clarify the semantics of slicing, etc. Feel free to (a) consult it in case of uncertainties and (b) suggest further edge cases if you find some.
- Another big source of changes is the re-organisation of the directory structure. That's not very important.
- Finally the feature unit conversions are lazily evaluated at `.get_as_pypicongpu`, so that function needs the time step size and number of steps, so there are some interface changes in that regard propagating through plugins into `simulation.py`.
- I've seen that sometimes the auto-formatter kicked in. If that's too annoying I can try to remove those changes.

### DIFF
--- a/lib/python/picongpu/picmi/diagnostics/timestepspec.py
+++ b/lib/python/picongpu/picmi/diagnostics/timestepspec.py
@@ -5,13 +5,31 @@ Authors: Julian Lenz
 License: GPLv3+
 """
 
-from enum import Enum
+from enum import Enum, EnumMeta
 from math import ceil
 
 from ...pypicongpu.output import TimeStepSpec as PyPIConGPUTimeStepSpec
 
 
-class TimeStepUnits(Enum):
+class CustomStrEnumMeta(EnumMeta):
+    """
+    This class provides some functionality of 3.12 StrEnum,
+    namely its __contains__() method.
+
+    You can safely remove this and inherit directly from StrEnum
+    once we switched to 3.12.
+    """
+
+    def __contains__(cls, val):
+        try:
+            cls(val)
+        except ValueError:
+            return False
+        else:
+            return True
+
+
+class TimeStepUnits(Enum, metaclass=CustomStrEnumMeta):
     """
     Units allowed in TimeStepSpec.
     """
--- a/lib/python/picongpu/picmi/diagnostics/timestepspec.py
+++ b/lib/python/picongpu/picmi/diagnostics/timestepspec.py
@@ -1,0 +1,180 @@
+"""
+This file is part of PIConGPU.
+Copyright 2025 PIConGPU contributors
+Authors: Julian Lenz
+License: GPLv3+
+"""
+
+from enum import StrEnum, auto
+from math import ceil
+
+from ...pypicongpu.output import TimeStepSpec as PyPIConGPUTimeStepSpec
+
+
+class TimeStepUnits(StrEnum):
+    """
+    Units allowed in TimeStepSpec.
+    """
+
+    STEPS = auto()
+    SECONDS = auto()
+
+    @classmethod
+    def _missing_(cls, value):
+        value = str(value).lower()
+        for member in cls:
+            if member.value == value:
+                return member
+        raise ValueError(f"Unknown time step unit. You gave {value}.")
+
+
+# The following might look slightly convoluted but is preferred over the more obvious use of `__class_getitem__`.
+# This is because the latter is supposed to return a GenericAlias which is not what we want.
+# That would be like list[int].
+# We are more like an `Enum` where indexing into the class means something different.
+# See [here](https://docs.python.org/3/reference/datamodel.html#object.__getitem__) and
+# [here](https://docs.python.org/3/reference/datamodel.html#classgetitem-versus-getitem).
+
+
+class _TimeStepSpecMeta(type):
+    """
+    Custom metaclass providing the [] operator on its children.
+    """
+
+    # Provide this to have a nice syntax picmi.diagnostics.TimeStepSpec[10:200:5, 3, 7, 11, 17]
+    def __getitem__(cls, args):
+        if not isinstance(args, tuple):
+            args = (args,)
+        return cls(*args)
+
+
+class TimeStepSpec(metaclass=_TimeStepSpecMeta):
+    """
+    A class to specify time steps for simulation output.
+
+    This class allows for flexible specification of time steps using slices
+    or individual indices. Its custom metaclass provides a [] operator on the class itself
+    for slicing and the () operator for choosing the unit such that the most convenient
+    way to use it is as follows:
+
+        ts = TimeStepSpec[:12:2, 7]("steps") + TimeStepSpec[1.e-15:5.e-15:2.e-16]("seconds")
+
+    In this example, `ts` specifies:
+    - every other time step for the first 12 time steps inclusively (`0, 2, 4, 6, 8, 10, 12`)
+    - AND the 7th time step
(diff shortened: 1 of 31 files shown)


lib/python/picongpu/pypicongpu/output/timestepspec.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def _serialize(spec):
+    if isinstance(spec, slice):
+        return {
+            "start": spec.start if spec.start is not None else 0,
+            "stop": spec.stop if spec.stop is not None else -1,
+            "step": spec.step if spec.step is not None else 1,
+        }
+    raise ValueError(f"Unknown serialization for {spec=} as a time step specifier (--period argument).")
```
BrianMarre, line 21: Wouldn't passing a list[int] cause a ValueError here? Or is  list[int] a slice object for some reason? If yes please add a comment explaining that.

But according to the implementation of TimeStepSpec that should be a valid input value

share/picongpu/pypicongpu/examples/laser_wakefield/main.py
```diff
@@ -107,14 +120,19 @@
...
 
 diagnostics_list = []
 diagnostics_list.append(
-    picmi.PhaseSpace(
+    picmi.diagnostics.PhaseSpace(
         species=electrons,
-        period=100,
+        # Resulting values for period:
+        # 17, 34, 50, 51, 57, 64, 68, 71, 85, 100, 102, ...
+        period=picmi.diagnostics.TimeStepSpec[::100, 50:72:7, 17],
```
BrianMarre: the syntax for the entry 17 is in my opinion misleading, I would assume that that indicated that we will write an output at the step 17, not every 17 steps as the implementation suggests.

Also the step 0 is missing from the list.

ikbuibui: 17 really means every 17 steps and not just only timestep number 17

lib/python/picongpu/picmi/diagnostics/timestepspec.py
```diff
@@ -0,0 +1,33 @@
...
+# The following might look slightly convoluted but is preferred over the more obvious use of `__class_getitem__`.
+# This is because the latter is supposed to return a GenericAlias which is not what we want.
+# That would be like list[int].
+# We are more like an `Enum` where indexing into the class means something different.
+# See [here](https://docs.python.org/3/reference/datamodel.html#object.__getitem__) and
+# [here](https://docs.python.org/3/reference/datamodel.html#classgetitem-versus-getitem).
+
+
+class _TimeStepSpecMeta(type):
+    # Provide this to have a nice syntax picmi.diagnostics.TimeStepSpec[10:200:5, [3,7,11,17]]
```
BrianMarre: This comment is misleading since list[int] is not supported, rather only slices or integers

lib/python/picongpu/pypicongpu/output/timestepspec.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class TimeStepSpec(RenderedObject):
+    specs = build_typesafe_property(list[slice | list[int]])
```
BrianMarre: allowing a list of integers looks to be remainder of a previous implementation, the current one actually only seems to work with list of slices and I consider that a good thing.

share/picongpu/pypicongpu/examples/laser_wakefield/main.py
chillenzer: @ikbuibui is correct but we agreed to change this behaviour. I'll make this mean just time step 17.

lib/python/picongpu/pypicongpu/output/timestepspec.py
chillenzer: You are correct. Thanks for catching that.

lib/python/picongpu/picmi/diagnostics/timestepspec.py
```diff
@@ -0,0 +1,180 @@
...
+            )
+        self.unit_system = unit_system
+        if unit_system == "seconds":
+            self.specs_in_seconds = self.specs
+            self.specs = tuple()
+        return self
+
+    def __add__(self, other):
+        if not (isinstance(other, TimeStepSpec)):
+            raise TypeError("unsupported operand type(s) for +: TimeStepSpec and {type(other)}")
```
ikbuibui:
```suggestion
            raise TypeError(f"unsupported operand type(s) for +: TimeStepSpec and {type(other)}")
```

lib/python/picongpu/picmi/diagnostics/timestepspec.py
```diff
@@ -0,0 +1,198 @@
...
+        # The following guards against setting units on the result of the addition.
+        # Otherwise one could specify time steps in "steps" unit, add that to
+        # another TimeStepSpec and reset the units.
+        ts.unit_system = "mixed"
+        return ts
+
+    def _transform_to_steps(self, specs_in_seconds, time_step_size):
+        return tuple(
+            slice(
+                int(spec.start / time_step_size if spec.start is not None else 0),
```
ikbuibui, line 165: There should probably be a check somewhere to ensure ``time_step_size > 0``. I say probably, since a subset of all possible simulations could possibly be time reversal invariant. 
It should definitely not be zero though, and this should be checked.

lib/python/picongpu/pypicongpu/output/timestepspec.py
chillenzer: Resolving on behalf of @BrianMarre

share/picongpu/pypicongpu/examples/laser_wakefield/main.py
chillenzer: Resolving on behalf of @BrianMarre

lib/python/picongpu/picmi/diagnostics/timestepspec.py
chillenzer: Resolving on behalf of @BrianMarre

lib/python/picongpu/pypicongpu/output/timestepspec.py
chillenzer: Resolving on behalf of @BrianMarre

ikbuibui (dismissed):
Small comments, nothing blocking.

lib/python/picongpu/picmi/diagnostics/timestepspec.py
chillenzer: Thanks for catching this.

lib/python/picongpu/picmi/diagnostics/timestepspec.py
chillenzer: I've added a check to `_transform_to_steps` because that is one place where `time_step_size == 0` would actually become a problem. But I think it's actually the responsibility of `picmi.Simulation` to check arguments for consistency. I'll add that to my todo list.

ikbuibui (approved):

PR #5311, opened 2025-03-12
fix picmi time steps calc

pordyna:
Number of time steps needs to be an integer, I think `ceil` is better than `floor`, but we could also do it the other way around.


BrianMarre (approved):
Good catch, just shows that we should start enforcing types in PICMI.

chillenzer (approved):

PR #5312, opened 2025-03-12
Add ability to pass extra data to kernels and execute a host side hook to binning plugin

ikbuibui:
Changes
- Add a host side hook for the user to execute code on the host side, before the binning is executed for the particular notify step. 
- Add ability to pass extra data to the user defined binning functors
- The default binningSetup.param is now empty and users are pointed to the documentation which provides a code example and points them to the LWFA and AtomicPhysics binning setups. The case which was compiled in the default param file is now moved to tests/compile2  

Abstractions and refactorings done for implementing field binning in an upcoming PR. Note: I will make documenation changes in the PR where I actually add field binning
- Refactored the ``Binner`` into a parent binner class and specialized child particle binner.
- Separated the helper kernels from ``Binner``into a separate file
- ``DomainInfo`` is refactored into a base and child class
- ``BinningData`` is refactored into a base and child class 

**Breaking changes** - Note: The following break existing binning param files. 

- The ``addBinner`` method of ``binningCreator`` is renamed to ``addParticleBinner`` 
- All functors now take arguments in the order ``worker, domainInfo, ...`` instead of ``domainInfo, worker, ...`` for consistency with the rest of the codebase.
- Quantity functors now get access to domain info as well, thus unifying the functor interface with the current one for axes as ``(auto const& worker, auto const& domainInfo, auto const& particle)`` `
- The ``writeOpenPMDFunctor`` has been moved from a optional argument to ``addBinner`` to a setter similar to other properties as ``setWriteOpenPMDFunctor``.
- The setter``setJsonCfg`` has been renamed to ``setOpenPMDJsonCfg``


include/picongpu/param/binningSetup.param
```diff
@@ -166,13 +168,14 @@ namespace picongpu
...
 
-                binningCreator
-                    .addBinner("chargeDensityBinning", axisTuple, speciesTuple, chargeDepositionData, writeOpenPMD)
+                binningCreator.addParticleBinner("chargeDensityBinning", axisTuple, speciesTuple, chargeDepositionData)
                     .setNotifyPeriod("1:500")
-                    .setJsonCfg(R"({"hdf5":{"dataset":{"chunks":"auto"}}})")
-                    .setOpenPMDExtension("h5");
+                    .setOpenPMDJsonCfg(R"({"hdf5":{"dataset":{"chunks":"auto"}}})")
+                    .setOpenPMDExtension("h5")
+                    .setOpenPMDWriteFunctor(writeOpenPMD);
```
chillenzer: This interface looks verbose and inflexible. Why not?
```suggestion
                    .setOpenPMDWriter(writeOpenPMD, WriteOpenPMDParams{"h5", ...});
```

This has the potential to be more concise and it allows changing what parameters the openPMD plugin takes without touching the interface of the binner.

include/picongpu/plugins/binning/BinningCreator.hpp
```diff
@@ -52,33 +51,44 @@ namespace picongpu
...
                 std::string const& binnerOutputName,
                 TAxisTuple const& axisTupleObject,
                 TSpeciesTuple const& speciesTupleObject,
                 TDepositionData const& depositionData,
-                std::function<
-                    void(::openPMD::Series& series, ::openPMD::Iteration& iteration, ::openPMD::Mesh& mesh)> const&
-                    writeOpenPMDFunctor
-                = [](::openPMD::Series& series, ::openPMD::Iteration& iteration, ::openPMD::Mesh& mesh) {})
-                -> BinningData<TAxisTuple, TSpeciesTuple, TDepositionData>&
+                T_Extras const& extraData = std::tuple<>{})
```
chillenzer:
```suggestion
                T_Extras const& extraData = {})
```
DRY, because if you ever forget to apply a change to both instances, weird things will happen:
```C++

#include <iostream>
#include <string>
#include <type_traits>

// clang-format off
/*
 * Does not compile with
 *
 * $ g++ default_args_vs_template_args.cpp && ./a.out
 * default_args_vs_template_args.cpp: In function 'auto f(T) [with T= int]': default_args_vs_template_args.cpp:5:47: error: could not convert
 * 'std::__cxx11::basic_string<char>()' from 'std::string' {aka
 * 'std::__cxx11::basic_string<char>'} to 'int'
 *    5 | template <typename T = int> auto f(T x = std::string{}) { return x; }
 */

// template <typename T = int> auto f(T x = std::string{}) { return x; }

// clang-format on

// Compiles, of course:
template <typename T = int> auto g(T x = {}) { return x; }

// ALSO COMPILES!
template <typename T = int> auto h(T x = 1.5) { return x; }

int main(int argc, char *argv[]) {
  // static_assert(std::is_same_v<decltype(f()), int>);
  static_assert(std::is_same_v<decltype(g()), int>);
  static_assert(std::is_same_v<decltype(h()), int>);

  std::cout << g() << std::endl;
  std::cout << h() << std::endl;
  return 0;
}
```
which prints
```
0
1
```

include/picongpu/plugins/binning/BinningCreator.hpp
```diff
@@ -52,33 +51,44 @@ namespace picongpu
...
+                -> ParticleBinningData<TAxisTuple, TSpeciesTuple, TDepositionData, T_Extras>&
             {
-                auto bd = BinningData<TAxisTuple, TSpeciesTuple, TDepositionData>(
+                auto bd = ParticleBinningData<TAxisTuple, TSpeciesTuple, TDepositionData, T_Extras>(
```
chillenzer:
```suggestion
                auto bd = ParticleBinningData(
```
Use template argument deduction because, you know... DRY. I've seen somewhere that bad things could happen in this case.

include/picongpu/plugins/binning/BinningData.hpp
```diff
@@ -49,48 +49,44 @@ namespace picongpu
...
-            T_SpeciesTuple speciesTuple;
             T_DepositionData depositionData;
-            std::function<void(::openPMD::Series& series, ::openPMD::Iteration& iteration, ::openPMD::Mesh& mesh)>
-                writeOpenPMDFunctor;
-            DataSpace<std::tuple_size_v<T_AxisTuple>> axisExtentsND;
+            T_Extras extraData;
+            pmacc::DataSpace<std::tuple_size_v<T_AxisTuple>> axisExtentsND;
 
             /* Optional parameters not initialized by constructor.
              * Use the return value of addBinner() to modify them if needed. */
```
chillenzer:
```suggestion
             * Use the return value of addParticleBinner() to modify them if needed. */
```

include/picongpu/plugins/binning/BinningData.hpp
```diff
@@ -49,48 +49,44 @@ namespace picongpu
...
             std::string notifyPeriod = "1";
             uint32_t dumpPeriod = 0u;
-            uint32_t particleRegion{ParticleRegion::Bounded};
 
             std::string openPMDInfix = "_%06T.";
             std::string openPMDExtension = openPMD::getDefaultExtension();
+            std::function<void(::openPMD::Series& series, ::openPMD::Iteration& iteration, ::openPMD::Mesh& mesh)>
+                writeOpenPMDFunctor = std::function<
+                    void(::openPMD::Series& series, ::openPMD::Iteration& iteration, ::openPMD::Mesh& mesh)>();
+            std::string openPMDJsonCfg = "{}";
```
chillenzer, line 79: As mentioned already with respect to the interface: I'd rather store a `WriteOpenPMDParams` instance.

include/picongpu/plugins/binning/BinningData.hpp
```diff
@@ -108,53 +104,105 @@ namespace picongpu
             }
 
             /** @brief Time average the accumulated data when doing the dump. Defaults to true. */
-            BinningData& setTimeAveraging(bool timeAv)
+            Child& setTimeAveraging(bool timeAv)
             {
                 this->timeAveraging = timeAv;
-                return *this;
+                return *static_cast<Child*>(this);
```
chillenzer: It is a beautiful paradigm but it's kind of annoying to read, isn't it? We should keep in mind that this will be much simpler and much more expressive in C++23: https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern#Deducing_this

include/picongpu/plugins/binning/DomainInfo.hpp
```diff
@@ -20,21 +20,29 @@
...
 
 #include <cstdint>
 
 namespace picongpu
 {
     namespace plugins::binning
     {
+        enum class BinningType
+        {
+            Field,
```
chillenzer, line 37: Technically speaking, this doesn't exist in your world, yet. But it's fine. My crystal ball tells me that it will soon be correct.

include/picongpu/plugins/binning/axis/LogAxis.hpp
```diff
@@ -113,13 +112,10 @@ namespace picongpu
...
-                    template<typename T_Worker, typename T_Particle>
-                    ALPAKA_FN_ACC std::pair<bool, uint32_t> getBinIdx(
-                        const DomainInfo& domainInfo,
-                        const T_Worker& worker,
-                        const T_Particle& particle) const
+                    template<typename... Args>
+                    ALPAKA_FN_ACC std::pair<bool, uint32_t> getBinIdx(Args const&... args) const
                     {
-                        auto val = getAttributeValue(domainInfo, worker, particle);
+                        auto val = getAttributeValue(args...);
```
chillenzer, line 118: Why this change? Does the field binning get other arguments?

include/picongpu/plugins/binning/BinningFunctors.hpp
```diff
@@ -59,21 +58,53 @@ namespace picongpu
...
                                && ...);
                     },
                     axes);
 
                 if(validIdx)
                 {
                     auto const idxOneD = pmacc::math::linearize(extents, binsDataspace);
-                    DepositionType depositVal = quantityFunctor(worker, particle);
-                    alpaka::atomicAdd(worker.getAcc(), &(histBox[idxOneD]), depositVal, ::alpaka::hierarchy::Blocks{});
+                    if constexpr(pmacc::memory::tuple::tuple_size_v < decltype(userFunctorData) >> 0)
```
chillenzer: That's what's meant, right?
```suggestion
                    if constexpr(pmacc::memory::tuple::tuple_size_v<decltype(userFunctorData)> > 0)
```

include/picongpu/plugins/binning/BinningFunctors.hpp
```diff
@@ -59,21 +58,53 @@ namespace picongpu
...
+                    {
+                        DepositionType depositVal = binning::apply(
+                            [&](auto const&... userFunctorData)
+                            { return quantityFunctor(worker, domainInfo, particle, userFunctorData...); },
```
chillenzer: I'm not quite sure if I should love or hate that you're using the same name `userFunctorData` shadowing the outer scope's variable here. Please make this a conscious decision.

include/picongpu/plugins/binning/BinningFunctors.hpp
```diff
@@ -59,21 +58,53 @@ namespace picongpu
...
+                            userFunctorData);
+                        alpaka::atomicAdd(
+                            worker.getAcc(),
+                            &(histBox[idxOneD]),
+                            depositVal,
+                            ::alpaka::hierarchy::Blocks{});
+                    }
+                    else
+                    {
+                        DepositionType depositVal = quantityFunctor(worker, domainInfo, particle);
```
chillenzer: Not sure about `binning::apply` but `std::apply` is perfectly fine with handling empty tuples, so this is identical to the `binning::apply` line above. Suggestion: Remove the `if` branching and discard the complete `else` branch.

include/picongpu/plugins/binning/BinningFunctors.hpp
```diff
@@ -59,21 +58,53 @@ namespace picongpu
...
+
+                                           binsDataspace[pmacc::memory::tuple::get<0>(tupleArg)] = binIdx;
+                                           return isValid;
+                                       }
+                                       else{
+                                           auto [isValid, binIdx] = pmacc::memory::tuple::get<1>(tupleArg).getBinIdx(worker, domainInfo, particle);
+                                           binsDataspace[pmacc::memory::tuple::get<0>(tupleArg)] = binIdx;
+                                           return isValid;
+                                       }
+                                   }(tupleArgs)
```
chillenzer: See below: Only keep `if` branch.

chillenzer (changes requested):
Looks good. It was very readable now that it's split. I've found a few cases of DRY that I'd appreciate a fix for but other than that, you're good to go.

include/picongpu/plugins/binning/BinningData.hpp
ikbuibui: Definitely

include/picongpu/plugins/binning/axis/LogAxis.hpp
ikbuibui: Not only the field binning, this also now gets ``extraData`` from the tuple optionally passed in by the user

include/picongpu/plugins/binning/BinningFunctors.hpp
ikbuibui: Yes, but the auto formatting seems broken. Maybe we need to update the clang-format version we use

include/picongpu/plugins/binning/BinningData.hpp
ikbuibui: Not changed because of the future PRs, as this section of the code is the same for both ``addParticleBinner`` and ``addFieldBinner``, so i collectively referred to them as ``addBinner``.

include/picongpu/param/binningSetup.param
ikbuibui: I will consider this in a follow up PR. Thanks for bringing it up

include/picongpu/plugins/binning/BinningData.hpp
chillenzer:
```suggestion
             * Use the return value of add...Binner() to modify them if needed. */
```

include/picongpu/plugins/binning/BinningData.hpp
chillenzer: We could get around this by defining a conversion operator.
```C++

#include <iostream>
template <typename TChild> struct Parent {

  operator TChild &() { return *static_cast<TChild *>(this); }

  TChild &get_child() { return *this; }
};

struct Child : Parent<Child> {};

int main(int argc, char *argv[]) {

  Child c;

  if (&c == &c.get_child())
    std::cout << "Identical!" << std::endl;
  else
    std::cout << "Bad!" << std::endl;

  return 0;
}
```

ikbuibui: as discussed offline, implemented this as a function instead of a conversion operator to avoid implicit conversions.

chillenzer: Offline discussion: We'll make a compromise and use an explicit method for the case.

include/picongpu/plugins/binning/BinningCreator.hpp
```diff
@@ -78,19 +79,15 @@ namespace picongpu
                 TAxisTuple const& axisTupleObject,
                 TSpeciesTuple const& speciesTupleObject,
                 TDepositionData const& depositionData,
-                T_Extras const& extraData = std::tuple<>{})
-                -> ParticleBinningData<TAxisTuple, TSpeciesTuple, TDepositionData, T_Extras>&
+                T_Extras const& extraData = {}) -> auto&
```
chillenzer: This looks kind of weird. I know that trailing return types are the cool new kid around the block (or at least were when I was still young XD) but I think just adding an ampersand to the starting `auto` would suffice. Not blocking though.

include/picongpu/plugins/binning/BinningCreator.hpp
```diff
@@ -78,19 +79,15 @@ namespace picongpu
...
                     binnerOutputName,
                     axisTupleObject,
                     speciesTupleObject,
                     depositionData,
                     extraData);
-                auto binner = std::make_unique<
-                    ParticleBinner<ParticleBinningData<TAxisTuple, TSpeciesTuple, TDepositionData, T_Extras>>>(
-                    bd,
-                    cellDescription);
+                auto binner = binning::make_unique<ParticleBinner>(bd, cellDescription);
```
chillenzer, line 89: We are soooooo cool! XD

chillenzer (dismissed):
One tiny non-blocking thing. If you fix it before the CI's done, fine. If not, I'll merge anyways.

include/picongpu/plugins/binning/BinningCreator.hpp
ikbuibui: no you <3

chillenzer: :blush:

include/picongpu/plugins/binning/BinningCreator.hpp
ikbuibui: Changed

chillenzer (dismissed):
How dare you make the CI fail? I felt so gracious to give you my approval again and now that! Anyhow... here it is.

chillenzer (approved):
You did it!

PR #5313, opened 2025-03-13
Topic macro particle count plugin for picmi

mafshari64:
- [x] Rebase after #5314 has been merged.

PR #5316, opened 2025-03-13
add auto output to picmi

pordyna:
## Auto output
`PyPIConGPU` comes with the 'auto' output configuration, but there was no counterpart in PICMI. This adds the missing part on picmi side.

Now it is possible to enable the auto configuration with sth like  `sim.add_diagnostic(picmi.output.Auto(period=100))`

### Open questions
I noticed we have a weird way of importing submodules in picmi, and putting everything on the top module level. Is this somehow given by the picmi standard? Here I did it differently, but I'm willing to change it to stay consistent. If we want to stay within what PICMI  standard implementation is doing, we should probably follow their naming scheme and call this class `AutoDiagnostic` and append the `Diagnostic` to all new incoming output definitions. `picmi.Auto` is not descriptive enough. @chillenzer @BrianMarre @psychocoderHPC  what is your opinion on that?

### DIFF
--- a/lib/python/picongpu/picmi/__init__.py
+++ b/lib/python/picongpu/picmi/__init__.py
@@ -11,6 +11,7 @@ from . import constants
 from .phase_space import PhaseSpace
 from .energy_histogram import EnergyHistogram
 from .macro_particle_count import MacroParticleCount
+from . import output
 
 from .distribution import FoilDistribution, UniformDistribution, GaussianDistribution
 from .interaction import Interaction
@@ -44,6 +45,7 @@ __all__ = [
     "PhaseSpace",
     "EnergyHistogram",
     "MacroParticleCount",
+    "output",
 ]
 
 
--- a/lib/python/picongpu/picmi/output/__init__.py
+++ b/lib/python/picongpu/picmi/output/__init__.py
@@ -1,0 +1,3 @@
+from .auto import Auto
+
+__all__ = ["Auto"]
--- a/lib/python/picongpu/picmi/output/auto.py
+++ b/lib/python/picongpu/picmi/output/auto.py
@@ -1,0 +1,47 @@
+"""
+This file is part of PIConGPU.
+Copyright 2025 PIConGPU contributors
+Authors: Pawel Ordyna
+License: GPLv3+
+"""
+
+from ...pypicongpu.output.auto import Auto as PyPIConGPUAuto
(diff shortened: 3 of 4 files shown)


BrianMarre:
This was a decision by Hannes the original developer, to keep names short for users.

I personally prefer names keeping imported stuff which is not exposed to the user in the original name spaces and have used non top level imports in stuff I implemented, see for example [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/lib/python/picongpu/picmi/gaussian_laser.py), but did not want to update unchanged files unnecessarily.

I would therefore prefer new files to **not** import non user exposed stuff to top level.

pordyna:
@BrianMarre so what I meant is `picmi.ouptut.Auto` vs `picmi.Auto` vs `picmi.AutoDiagnostic` on the user side.

BrianMarre:
Current PICMI standard style is `picmi.AutoDiagnostic`, I would actually prefer `picmi.output.Auto`.

For me a diagnostic indicates a debugging or performance tracking tool, not something that writes out data from a simulation but I think they named as a reference to a diagnostic of a computer experiment.

chillenzer:
~~PyPIConGPU already got the submodules, our PICMI implementation currently doesn't. I'd be willing to change this but there are at least two further files that need moving and an [open PR](https://github.com/ComputationalRadiationPhysics/picongpu/pull/5313) with the old convention. I'd say we keep it as is for now and change the interface in one dedicated PR.~~

Edit: This change of convention has already started in #5310. That one will need a rebase but I only wanted to merge it after the other work on plugins has been merged, so I don't hinder other people's progress and incur a lot of rebasing effort for others.

chillenzer:
Just to be explicit: I'm happy to go through with this. But I'd suggest the following changes:
1. Rebase against the latest dev.
2. Add this to at least one of the examples. (We're currently mostly pushing the `laser_wakefield` one, so that's the most up-to-date.)
3. Optional: Restructure to put it at the top-level. (Keeping it as is would make the interface in the latest `dev` slightly inconsistent but we're planning to transition to the structure proposed in the current revision of the PR in the near future. So it would be double work to go back and forth.)

Wanna push this, @pordyna?

pordyna:
@chillenzer I added it to warm plasma since it had not output at all.

chillenzer (approved):

chillenzer:
The PICMI interface is slightly inconsistent now but #5310 will fix this again.

PR #5317, opened 2025-03-14
Supercharge getParticlePosition in binning plugin

ikbuibui:
Following input from @BrianMarre and @PrometheusPi, made ``getParticlePosition`` in the binning plugin more flexible and hopefully easier to use.

``getParticlePosition<DomainOrigin>`` now accepts two additional optional template parameters, ``PositionPrecision`` and ``PositionUnits`` as ``getParticlePosition<DomainOrigin, PositionPrecision, PositionUnits>``. 

``PositionPrecision`` by default uses cell level precision, thus returning the cell index, but if you want the exact particle position within the cell as well and not just the cell index, you can use ``PositionPrecision::SubCell``. 
``PositionUnits`` by default uses cell units, but it is also possible to get positions in SI units by specifying PositionUnits::SI

The default behaviour of this function is maintained, and this is NOT a breaking change.

### DIFF
--- a/include/picongpu/plugins/binning/DomainInfo.hpp
+++ b/include/picongpu/plugins/binning/DomainInfo.hpp
@@ -76,7 +76,7 @@ namespace picongpu
 
         enum class PositionPrecision
         {
-            // Returns the cell index in which the particle
+            // Returns the cell index of the particle
             Cell,
             // Returns the position of the particle as the cell index + the position of the particle inside the cell
             // [0,1) This value is a floating point number of cells
--- a/include/picongpu/plugins/binning/DomainInfo.hpp
+++ b/include/picongpu/plugins/binning/DomainInfo.hpp
@@ -22,6 +22,7 @@
 #include "picongpu/defines.hpp"
 
 #include <pmacc/dimensions/DataSpace.hpp>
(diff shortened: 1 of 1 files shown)


include/picongpu/plugins/binning/DomainInfo.hpp
```diff
@@ -73,12 +74,68 @@ namespace picongpu
             LOCAL
         };
 
-        template<DomainOrigin T_Origin, typename T_Particle>
-        ALPAKA_FN_ACC auto getParticlePosition(DomainInfo domainInfo, T_Particle particle) -> pmacc::DataSpace<simDim>
+        enum class PositionPrecision
+        {
+            // Returns the cell index in which the particle
```
psychocoderHPC: Your AI stopped to generate the sentence ^^

psychocoderHPC (changes requested):
small doxygen issue^^

include/picongpu/plugins/binning/DomainInfo.hpp
ikbuibui: Done

psychocoderHPC (approved):

PR #5318, opened 2025-03-18
Refactor laser implementation

### LINKED ISSUE #4718: incident laser profiles need improvements
During the review of #4674 I stumbled upon a few non-ideal code blocks. @pordyna pointed out correctly that these were in the code before his pull request, and thus should not be addressed with the changes he introduced. 

This issue is a reminder that in #4674 is a list of stuff that needs improvement in the incident laser profiles.

### LINKED ISSUE #5269: ToDo in laser refactoring
Since the implementation of the lasers is in parts at the least confusing and sometimes inconsistent, it is going to be refactored. The following action items are foreseen:
- [x] check if all separable lasers have a correct phase calculation
- [x] refactor `BaseFunctorE`:
    * ~move `getCurrentTime()` into `struct BaseSeparableFunctorE`~ 👈 don't do it, according to the variables used, this is a base class functionality
    * rename this 👆 function to `getTminusXoverC()` in order to reflect its purpose
- [x] remove the laser cutoff, as it is wrong. Laser phases and envelopes should be computed for all times at all positions on the incident field plane.
```c++
// Cut off when the laser has not entered at this point yet to avoid confusion.
if(time < 0.0_X)
    return float3_X::create(0.0_X);
```
- [ ] ~Remove Wavepacket laser, as it is almost identical to PlaneWave? Except for the *transverse* Gaussian Envelope. Is it possible to at least reuse the plane wave laser implementation? Similar to the pulse-front tilt for the Gaussian? But why not use the gaussian instead of wavepacket for lasers with transverse envelope?~
- [x] Remove the pulse-front tilt feature in the standard Gaussian laser? It is not physical anyways, since we do not know how pulse-front tilts propagate for Laguerre-Gaussian modes, and for standard Gaussian pulses there is the DispersivePulse implementing the feature correctly.
- [x] check if lasers define variables in their `Params` struct again although these are already defined in their `BaseParam` struct (happens at least with `LASER_PHASE` in wavepacket laser)
- [x] since all lasers have define a focus (via `BaseParam`), the laser internal coordinate system should be relative to the focus position and not the `origin` (the former being the typical way of describing laser fields analytically).
  `getInternalCoordinates()` is only used in `DispersivePulse` and `GaussianPulse` anyways.
- [x] reimplement Gaussian laser in order to have correct phases and envelope computation
    * allow for different width along the transverse axis (cf. polynom laser)
- [x] refactor `DispersivePulse`:
    - [x] check the time delay computation in the `DispersivePulse` for correctness wrt to `BaseFunctorE` internal coordinates and time
    - [x] check correctness of the sign of the phase computation. **Is fine, but focusPos has the wrong sign, which is properly accounted for in the DispersivePulse implementation** 
    - [x] move computation of internal coordinates from `amp()` and `phi()` to `getValueE()`
        * when doing so, change `focusPos` to `focusDis` and take the proper sign (i.e. switch)
    - [x] make dispersion axis parallel to polarization
    - [x] Update comments according to other branch
    - [x] shorten computation of magnitude in `amp()`
        ```cpp
        // Normalization to Amplitude
        mag *= math::sqrt(pmacc::math::Pi<float_X>::doubleValue * 0.5_X) * 2.0_X
            * Unitless::PULSE_DURATION * Unitless::AMPLITUDE;
        
        // Dividing amplitude by 2 to compensate doubled spectral field strength
        // resulting from E(-Omega) = E*(Omega), which has to be fulfilled for E(t) to be real
        mag *= 0.5_X;
        ```
- [x] rename `Unitless::w` -> `Unitless::OMEGA0` (since there is confusion in the Gaussian profile with the width)
- [x] does accessing `Unitless::LASER_PHASE` really work if it is defined in a mother::mother::mother::mother class?
- [x] fix grammar errors:
    * **PlaneWave.def**:
        * l. 59: fulfulled
        * l. 42: `PULSE_INIT` -> `RAMP_INIT` in documentation OR rename parameter. `PULSE_INIT` is used in all other profiles.
    * **PlaneWave.hpp**:
        * l. 124: `Unitless::RAMP_INIT`  does not exist (but `Unitless::INIT_TIME`)
        * l. 146: altough announced, there is no description of formula -> ask @PrometheusPi
    * **BaseParam.hpp**:
        * l.150 coordinate
    * **Functors.hpp**
        * l.142 coordinate
        * l. 308 comma missing after 'However'
    * **Wavepacket.def**:
        * l.70: `LASER_PHASE` is defined here and in `BaseParam`, too!
    * **Traits.hpp**:
        * l. 27 non-zero
        * l. 24: Huyhens -> Huygens
        * l.39: short-hand

steindev:
This refactoring straightens the implementation of laser profiles and removes confusing terms being added and later subtracted.

Major changes are:
* redefinition of the internal coordinate system: The focus becomes the actual 'origin' of  the coordinate system, while the `origin` variable refers to the position on the Huygens surface where the field needs to be evaluated in order to feed it into the simulation
  => This allows to implement expressions for laser fields as they are printed in publications and text books w/o additional phase and envelope terms 
* the GaussianPulse can have different waist sizes along the transverse directions now
* the pulse-front tilt feature of the GaussianPulse is removed, because it is unknown how pulse-front tilted Laguerre-Gaussian pulses propagate (our old implementation fed some field in the simulation which propagated physically toward the focus, but this (most probably) did not correspond to the expectation for the in-focus field)
  => For pulse-front tilted standard Gaussian pulses the DispersivePulse can and should be used
* dispersions in the DispersivePulse are applied along the polarization axis, while the other transverse axis is dispersion-free

Draft status of this PR will be removed when the following action items are performed
* `GaussianPulse`:
  - [x] Compile test simulation w/ and w/o Laguerre modes and for various propagation directions and polarizations
  - [x] Compare results of test sim to old implementation
  - [x] resolve possible issues
* `DispersivePulse`:
  - [x] Compile test simulation for various propagation directions and dispersion values
  - [x] Compare results of test sim to old implementation
  - [x] resolve possible issues

Close #5269


steindev:
@psychocoderHPC @PrometheusPi @chillenzer @ikbuibui PR is ready for review now.

It is intentionally split in three commits each of which refactors a specific part of the old implementation.

Test results have been discussed offline already and are documented in #5269.
Main takeaway from the tests: Results are equal between old and refactored implementation in various cases. Additionally, stray fields are reduced in the new implementation.

steindev:
ping @PrometheusPi @finnolec @psychocoderHPC @ikbuibui @chillenzer. No review within 3 weeks is actually quite bad... :disappointed:

include/picongpu/fields/incidentField/Functors.hpp
```diff
@@ -448,10 +457,7 @@ namespace picongpu
                     template<typename T_SeparableFunctor>
                     HDINLINE float3_X operator()(T_SeparableFunctor const& functor, floatD_X const& totalCellIdx) const
                     {
-                        auto const time = functor.getCurrentTime(totalCellIdx);
-                        // Cut off when the laser has not entered at this point yet to avoid confusion.
-                        if(time < 0.0_X)
-                            return float3_X::create(0.0_X);
```
chillenzer, line 454: Why's this removed?

include/picongpu/fields/incidentField/profiles/BaseParam.hpp
```diff
@@ -147,7 +147,7 @@ namespace picongpu
                             _error_laser_polarization_direction_vector_must_be_orthogonal_to_propagation_direction____check_your_incidentField_param_file,
                             (dotPropagationPolarization > -0.0001) && (dotPropagationPolarization < 0.0001));
 
-                        /** Focus position in total cooridnate system
```
chillenzer, line 150: That seems to be a common typo?

include/picongpu/fields/incidentField/profiles/DispersivePulse.def
```diff
@@ -23,96 +23,110 @@
...
-                         * ~ exp(-i * 1/6 * TOD*(Omega - Omega_0)^3)
-                         *
-                         * unit: s^3
-                         */
-                        static constexpr float_64 TOD_SI = 0.0;
-                    };
-                } // namespace defaults
+                baseMetadata["Dispersive Gaussian parameters"] = dispersiveMetadata;
+                return baseMetadata;
+            }
```
chillenzer, line 120: You did it! Metadata!

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.def
```diff
@@ -57,8 +60,6 @@ namespace picongpu
                         static constexpr float_64 TIME_POINT_3_SI = -100.0e-15;
                         /** @} */
 
-                        static constexpr float_64 AMPLITUDE_SI = 1.0e6;
-
```
chillenzer, line 61: Where did this go?

include/picongpu/fields/incidentField/profiles/Wavepacket.hpp
```diff
@@ -135,17 +135,18 @@ namespace picongpu
...
                             {
                                 // upramp = start
                                 auto const exponent
                                     = ((runTime - Unitless::endUpramp) / Unitless::PULSE_DURATION / math::sqrt(2.0_X));
                                 envelope *= math::exp(-0.5_X * exponent * exponent);
-                                correctionFactor = (runTime - Unitless::endUpramp) / (tau * tau * Unitless::w);
+                                correctionFactor = (runTime - Unitless::endUpramp) / (tau * tau * Unitless::OMEGA0);
                             }
-                            auto const phase = Unitless::w * runTime + Unitless::LASER_PHASE + phaseShift;
+                            auto const phase = Unitless::OMEGA0 * runTime + Unitless::LASER_PHASE + phaseShift;
```
chillenzer, line 149: Didn't we just remove `LASER_PHASE` in `Wavepacket.def`?

chillenzer (commented):
You are right, here's my first batch of review. I've yet circumvented the two main files but you might want to start with what I've got here. It's mostly questions for getting an overview.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -31,367 +31,378 @@
...
+                                    + Unitless::TOD / 6.0_X * (Omega - Unitless::OMEGA0) * (Omega - Unitless::OMEGA0)
+                                          * (Omega - Unitless::OMEGA0)
+                                    + phaseShift + Unitless::LASER_PHASE;
+
+                    // Initial frequency dependent complex phase
+                    float_X alpha = expandedWaveVectorX(Omega);
+
+                    // Center of a frequency's spatial distribution
+                    float_X center = Unitless::SD * (Omega - Unitless::OMEGA0)
+                                     - sim.pic.getSpeedOfLight() * alpha * pos[0] / (Unitless::W0 * Unitless::OMEGA0);
```
chillenzer: Could be extracted into an own function.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -31,367 +31,378 @@
...
+
+                    // inverse radius of curvature of the pulse's wavefronts
+                    auto const R_inv
+                        = pos[0] / (Unitless::rayleighLength * Unitless::rayleighLength + pos[0] * pos[0]);
+
+                    /* Parabolic phase curvature out of focus
+                     * Apply shift of frequency's distribution center only along polarization direction */
+                    phase += ((pos[1] - center) * (pos[1] - center) + pos[2] * pos[2]) * Omega * 0.5_X * R_inv
```
chillenzer, line 236: Why's there a `pos[2]` term here?

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -31,367 +31,378 @@
...
+                     *
+                     * The window of laser initialization is [-0.5*INIT_TIME, 0.5*INIT_TIME]*c along the central beam
+                     * axis around the laser pulse maximum. Note, the position of the pulse maximum is influenced by
+                     * TIME_DELAY.
+                     *
+                     * In case of oblique propagation of the pulse with respect to the Huygens surface, points on the
+                     * Huygens surface can be within the window or outside the window depending on their transverse
+                     * offset from the pulse's origin on the surface.
+                     */
+                    auto const shiftedTime = this->getTminusXoverC(totalCellIdx); // >0: inside, <0: outside
```
chillenzer, line 281: What weird compiler bug made you use `this->` here?

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -136,29 +127,37 @@ namespace picongpu::fields::incidentField
...
                 }
 
+                //!
+                HINLINE static float_X getLaguerreNorm()
+                {
+                    float_X value = 0.0_X;
+                    for(uint32_t m = 0; m < Unitless::laguerreModes.size(); ++m)
+                        value += Unitless::laguerreModes[m];
+
+                    return value;
```
chillenzer, line 146: Should be
```suggestion
                    using namespace std;
                    return reduce(cbegin(Unitless::laguerreModes), cend(Unitless::laguerreModes), std::Plus<float_X>{});
```
but @psychocoderHPC explained that `pmacc::Vector` doesn't have an iterator interface.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -204,96 +207,123 @@ namespace picongpu::fields::incidentField
...
-                              1.0_X + (focusPos / Unitless::rayleighLength) * (focusPos / Unitless::rayleighLength));
+                    /* Shifting pulse for TIME_SHIFT, TIME_DELAY, and travel time from the focus to
+                     * the origin in order to start with the front of the laser pulse at the origin. */
+                    constexpr auto mue = LongitudinalEnvelope::TIME_SHIFT + Unitless::TIME_DELAY;
+                    float_X const timeDelay = mue - (distanceOriginRelativeToFocus / sim.pic.getSpeedOfLight());
 
+                    // transform current point of evaluation to laser internal coordinates
+                    // The laser internal coordinate system's origin
+                    // is in the focus and the laser propagates along the x-axis.
+                    float3_X internalPosition = this->getInternalCoordinates(totalCellIdx);
```
chillenzer, line 232: This is the same as in `DispersivePulse`. Would it make sense to extract this transformation into a separate function?

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -204,96 +207,123 @@ namespace picongpu::fields::incidentField
...
+                              / (internalPosition[0] * internalPosition[0]
+                                 + Unitless::rayleighLength_AXIS_2 * Unitless::rayleighLength_AXIS_2);
+
+                        gouy += .5_X * math::atan(normalizedDistanceFromFocus_AXIS_2);
+                        evaluationTime -= internalPosition[2] * internalPosition[2] * inverseR_AXIS_2
+                                          / (2._X * sim.pic.getSpeedOfLight());
+                        amplitudeExponent -= internalPosition[2] * internalPosition[2] / (waist_AXIS_2 * waist_AXIS_2);
+                        amplitudeAbs /= math::sqrt(
+                            math::sqrt(
+                                1._X + normalizedDistanceFromFocus_AXIS_2 * normalizedDistanceFromFocus_AXIS_2));
```
chillenzer, line 285: Extract this into function, so you can reduce duplication with the `AXIS_1` part.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -204,96 +207,123 @@ namespace picongpu::fields::incidentField
...
+                    // distance from focus in units of the Rayleigh length
+                    float_X const normalizedDistanceFromFocus_AXIS_1
+                        = internalPosition[0] / Unitless::rayleighLength_AXIS_1;
 
-                    // Apply tilt if needed
-                    if constexpr(Unitless::TILT_AXIS_1 || Unitless::TILT_AXIS_2)
+                    // beam waist at the generation plane so that at focus we will get W0
+                    float_X const waist_AXIS_1
+                        = Unitless::W0_AXIS_1
+                          * math::sqrt(1._X + normalizedDistanceFromFocus_AXIS_1 * normalizedDistanceFromFocus_AXIS_1);
```
chillenzer, line 241: Could be a separate function.

chillenzer (dismissed):
Can't say much about the physical correctness but I've checked with the formulae from the given reference and it kinda makes sense to me. Some questions for my understanding and a few minor refactoring ideas are in the comments but they are mostly up to personal style I'd guess, so not blocking.

Wouldn't wanna merge myself in order to have a physicist have a look at this as well, so would you, @PrometheusPi, do the honours please?

chillenzer:
PS: Generally very nice documentation. With the given references and the additional hints in the comments it was quite straightforward to follow the formulae (mostly).

PrometheusPi:
@steindev Will you first implement @chillenzer's suggestions are do you want to have more reviews at the current state?

psychocoderHPC:
@PrometheusPi please already start the review. @chillenzer mostly ask for renamings.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
psychocoderHPC: If `getTminusXoverC()` is defined in the base class it is sometimes required but I think only if the method is templated.

chillenzer:
Yes, definitely! Please review. My comments were at most low-priority colouring ideas concerning our new bike shed. Did you know that research has shown that bikes in a green shed are 12.657498% less likely to get stolen by people wearing a blue or yellow shirt? (If the shirt is blue and yellow, for example in stripes, the results are inconclusive and further research is needed.)

chillenzer:
Any movement here?

PrometheusPi:
@chillenzer soon (tomorrow)

include/picongpu/fields/incidentField/Functors.hpp
steindev: Since the definition of time has changed during the refactoring the field actually needs to be evaluated for time<0 in order to produce the correct field before the main pulse.

include/picongpu/fields/incidentField/profiles/BaseParam.hpp
steindev: That is why it is fixed.

include/picongpu/fields/incidentField/profiles/DispersivePulse.def
steindev: :sunglasses:

include/picongpu/fields/incidentField/profiles/ExpRampWithPrepulse.def
steindev: It is already defined in the `BaseParam` (or so). Its occurence here is a shadowing of the previous definition in a mother param struct.

include/picongpu/fields/incidentField/profiles/Wavepacket.hpp
steindev: `LASER_PHASE` is a constant offset, that is also defined in a mother param struct and has been shadowed with a second definition in `Wavepacket.def`.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
steindev: Yes, could be. But I use it only once and its length is comparable to the other definitions. i.e. not as long as `expandedWaveVectorX()`.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
steindev: In order to evaluate the field along the third dimension, i.e. along z, too.

include/picongpu/fields/incidentField/Functors.hpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2020-2024 Sergei Bastrakov, Finn-Ole Carstens
+/* Copyright 2020-2024 Sergei Bastrakov, Finn-Ole Carstens, Klaus Steiniger
```
PrometheusPi, line 1: Year will be updated next year - fine with me to keep it

include/picongpu/fields/incidentField/profiles/PlaneWave.hpp
```diff
@@ -66,10 +67,6 @@ namespace picongpu
                         // unit: sim.unit.time()
                         static constexpr float_X LASER_NOFOCUS_CONSTANT
                             = static_cast<float_X>(Params::LASER_NOFOCUS_CONSTANT_SI / sim.unit.time());
-                        // unit: sim.unit.time()
-                        static constexpr float_X INIT_TIME = static_cast<float_X>(
-                            (Params::RAMP_INIT * Params::PULSE_DURATION_SI + Params::LASER_NOFOCUS_CONSTANT_SI)
-                            / sim.unit.time());
```
PrometheusPi, line 72: Just for my understanding: where does this get defined now?

include/picongpu/fields/incidentField/profiles/PlaneWave.def
```diff
@@ -38,8 +38,9 @@ namespace picongpu
                          *  unit: seconds */
                         static constexpr float_64 LASER_NOFOCUS_CONSTANT_SI = 13.34e-15;
 
-                        /** The laser pulse will be initialized after TIME_DELAY_SI half of PULSE_INIT times of the
-                         * PULSE_DURATION before and after the plateau unit: none */
+                        /** The laser pulse will be initialized after TIME_DELAY_SI half of RAMP_INIT times of the
```
PrometheusPi: I know I introduced this, but the factor half of ramp-init is confusing since the variable name already states init as the up-ramp in my opinion..

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
steindev: I think it has been like this in the original version and I just continued to use it. In case I drop it, the error message is:
```bash
/home/.../picongpu/fields/incidentField/profiles/DispersivePulse.hpp(283): error: identifier "getTminusXoverC" is undefined
                      auto const shiftedTime = getTminusXoverC(totalCellIdx);
```
with `GNU 12.2.0` and `NVIDIA 12.1.105`

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
steindev: It is only the first three lines that are the same. And two of these three varaibles that are defined in these lines are reused later on. Actually, I see no advantage in outsourcing this.

include/picongpu/fields/incidentField/profiles/DispersivePulse.def
```diff
@@ -23,96 +23,110 @@
...
+
+            /** Width of the spectral support for the discrete Fourier transform of the pulse's field
+             * from frequency-space domain to time-space domain.
+             * Defined as number of sigmas of the laser pulse spectrum.
+             * NOTE: A too small value will artificially increase the pulse duration and
+             * too high values will increase the time per step during initialization of the pulse.
+             *
+             * unit: none
+             */
+            static constexpr float_X SPECTRAL_SUPPORT = 6._X;
```
PrometheusPi, line 60: @steindev Why did the spectral support increase from 4 to 6?

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
steindev: I do not know how. These are three variables reused multiple times to compute additional factors which are added to four different variables. And whether this addition takes place or not depends on `simDIM`.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
steindev: Could be, but used only once...

include/picongpu/fields/incidentField/profiles/DispersivePulse.def
```diff
@@ -23,96 +23,110 @@
...
+
+            /** The laser pulse will be initialized PULSE_INIT times of the PULSE_DURATION after
+             * TIME_DELAY_SI.
+             * WATCH OUT! Dispersion parameters may lead to pulse lenghtening
+             * Please ensure to choose a value high enough to cover the whole pulse,
+             * otherwise the Fourier Transformation will misbehave.
+             *
+             *  unit: none
+             */
+            static constexpr float_64 PULSE_INIT = 17.0;
```
PrometheusPi, line 50: Why did this change from 20 to 17?

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -170,6 +169,10 @@ namespace picongpu::fields::incidentField
                  * pos[0] is the propagation direction.
                  *
                  * References:
+                 * R. Pausch et al. (2022), Modeling Beyond Gaussian Laser Pulses in Particle-in-Cell Simulations - The
```
PrometheusPi, line 172: I wouldn't trust this author.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -204,96 +207,123 @@ namespace picongpu::fields::incidentField
...
+                          * math::sqrt(1._X + normalizedDistanceFromFocus_AXIS_1 * normalizedDistanceFromFocus_AXIS_1);
+
+                    // inverse radius of curvature
+                    float_X const inverseR_AXIS_1
+                        = internalPosition[0]
+                          / (internalPosition[0] * internalPosition[0]
+                             + Unitless::rayleighLength_AXIS_1 * Unitless::rayleighLength_AXIS_1);
+
+                    // Gouy phase
+                    float_X gouy = .5_X * math::atan(normalizedDistanceFromFocus_AXIS_1);
```
PrometheusPi, line 250: Why is there a factor 0.5 here? 
This differs from [this definition](https://en.wikipedia.org/wiki/Gaussian_beam#Gouy_phase).

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -204,96 +207,123 @@ namespace picongpu::fields::incidentField
...
+                                                   / (2._X * sim.pic.getSpeedOfLight());
+
+                    auto amplitudeExponent
+                        = -internalPosition[1] * internalPosition[1] / (waist_AXIS_1 * waist_AXIS_1);
+
+                    /* Compute amplitude and its reduction outside the focus due to defocussing. */
+                    auto amplitudeAbs
+                        = Unitless::AMPLITUDE
+                          / math::sqrt(
+                              math::sqrt(
```
PrometheusPi, line 262: This looks terribly unprecise.
If we have 6 digits of precision, `sqrt(sqrt(...))` will reduce precision to 1-2 digits.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -204,96 +207,123 @@ namespace picongpu::fields::incidentField
...
+                        float_X const waist_AXIS_2
+                            = Unitless::W0_AXIS_2
+                              * math::sqrt(
+                                  1._X + normalizedDistanceFromFocus_AXIS_2 * normalizedDistanceFromFocus_AXIS_2);
+                        float_X const inverseR_AXIS_2
+                            = internalPosition[0]
+                              / (internalPosition[0] * internalPosition[0]
+                                 + Unitless::rayleighLength_AXIS_2 * Unitless::rayleighLength_AXIS_2);
+
+                        gouy += .5_X * math::atan(normalizedDistanceFromFocus_AXIS_2);
```
PrometheusPi, line 279: Okay - here another factor 1/2 is added to lead to the standard definition.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -204,96 +207,123 @@ namespace picongpu::fields::incidentField
...
+                        evaluationTime -= internalPosition[2] * internalPosition[2] * inverseR_AXIS_2
+                                          / (2._X * sim.pic.getSpeedOfLight());
+                        amplitudeExponent -= internalPosition[2] * internalPosition[2] / (waist_AXIS_2 * waist_AXIS_2);
+                        amplitudeAbs /= math::sqrt(
+                            math::sqrt(
```
PrometheusPi, line 284: double sqrt

steindev:
@chillenzer Thanks for your review :pray:
Your questions are answered. I have had issues with one of your requests. Let's talk offline about this.

Also, it seems I do not see any need for changes from your comments, sorry :sweat_smile:
But if a certain change is important to you, I will do it. Please tell me.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
PrometheusPi: This comes from the (potentially) two different $w_0$

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -31,367 +31,378 @@
...
             {
-                namespace detail
+                //! User SI parameters
+                using Params = T_Params;
+
+                //! Base unitless parameters
+                using Base = BaseParamUnitless<T_Params>;
+
+                // unit: none
+                using Params::SPECTRAL_SUPPORT;
```
PrometheusPi, line 54: just for my curiosity: is `using` here loading a variable?

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -31,367 +31,378 @@
...
-                    template<typename T_Params>
-                    struct DispersivePulseFunctorIncidentE
-                        : public DispersivePulseUnitless<T_Params>
-                        , public incidentField::detail::BaseFunctorE<T_Params>
+                }
+
+                /** Calculate incident field E value for the given position
+                 * The implementation is based on a definition of the electric field E given in frequency-space domain,
+                 * which is printed in eq. (6) of ref. [1]. Compared to the reference, higher order derivatives of the
+                 * frequency dependent propagation angle are set to zero, i.e. theta'' = theta''' = 0.
```
PrometheusPi, line 110: What are the `'` standing for here?

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -31,367 +31,378 @@
...
+                 */
+                HDINLINE float_X expandedWaveVectorX(float_X const Omega) const
                 {
-                    //! Get text name of the incident field profile
-                    HINLINE static std::string getName()
+                    return Unitless::W0 / sim.pic.getSpeedOfLight()
+                           * (Unitless::OMEGA0 * Unitless::AD * (Omega - Unitless::OMEGA0)
+                              + Unitless::AD * (Omega - Unitless::OMEGA0) * (Omega - Unitless::OMEGA0)
+                              - Unitless::OMEGA0 / 6.0_X * Unitless::AD * Unitless::AD * Unitless::AD
+                                    * (Omega - Unitless::OMEGA0) * (Omega - Unitless::OMEGA0)
```
PrometheusPi: for this equation, I would suggest storing `(Omega - Unitless::OMEGA0)` in a const variable before and use it here. This avoids 6-1=5 subtractions.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -31,367 +31,378 @@
...
+
+                    // Initial frequency dependent complex phase
+                    float_X alpha = expandedWaveVectorX(Omega);
+
+                    // Center of a frequency's spatial distribution
+                    float_X center = Unitless::SD * (Omega - Unitless::OMEGA0)
+                                     - sim.pic.getSpeedOfLight() * alpha * pos[0] / (Unitless::W0 * Unitless::OMEGA0);
+
+                    // gaussian envelope in frequency domain
+                    float_X const envFreqExp = -(Omega - Unitless::OMEGA0) * (Omega - Unitless::OMEGA0)
```
PrometheusPi: here as well

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -31,367 +31,378 @@
...
+                    float_X const waist
+                        = Unitless::W0
+                          * math::sqrt(
+                              1.0_X + (pos[0] / Unitless::rayleighLength) * (pos[0] / Unitless::rayleighLength));
+
+                    // Initial frequency dependent complex phase
+                    float_X alpha = expandedWaveVectorX(Omega);
+
+                    // Center of a frequency's spatial distribution
+                    float_X center = Unitless::SD * (Omega - Unitless::OMEGA0)
```
PrometheusPi: here as well

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
```diff
@@ -31,367 +31,378 @@
...
+                HDINLINE float_X phi(float3_X const& pos, float_X const Omega, float_X const phaseShift) const
                 {
-                    using type = profiles::detail::DispersivePulseFunctorIncidentE<T_Params>;
-                };
+                    /* 'initial value' of phase due to distance of position where E-value is computed from focus
+                     * and in-focus dispersion values */
+                    float_X phase = Omega * pos[0] / sim.pic.getSpeedOfLight()
+                                    + 0.5_X * Unitless::GDD * (Omega - Unitless::OMEGA0) * (Omega - Unitless::OMEGA0)
+                                    + Unitless::TOD / 6.0_X * (Omega - Unitless::OMEGA0) * (Omega - Unitless::OMEGA0)
+                                          * (Omega - Unitless::OMEGA0)
```
PrometheusPi: here as well

PrometheusPi (changes requested):
only minor things and questions

include/picongpu/fields/incidentField/profiles/PlaneWave.hpp
steindev: Nowhere, because it has not been used. But `RAMP_INIT` is used in the code further below again.

include/picongpu/fields/incidentField/profiles/PlaneWave.def
steindev: :thinking: I see what you mean...will change it

include/picongpu/fields/incidentField/profiles/DispersivePulse.def
steindev: Results of numerical experiments. See #5269.

include/picongpu/fields/incidentField/profiles/DispersivePulse.def
steindev: Results of numerical experiments. See #5269.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
steindev: :rofl: :rofl: :rofl:

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
steindev: Exactly. It is 0.5 per transverse axis, i.e. per dimension.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
steindev: Well, one could argue it is only in the amplitude, not the phase. But do you have a better solution? Use `pow()` instead?

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
steindev: Suggestion?

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
steindev: Yes, it is to make it known in this derived class. But I cannot exactly recall anymore why it is necessary. I guess, because the mother class actually is a template argument.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
steindev: Read `'` as 'prime' and it means derivative with respect to frequency. Standard notation. Obviously... :unamused: :wink:

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
steindev: Also did it in the other functions.

steindev:
Thanks for your review @PrometheusPi :pray:
I made changes to the code accordingly and directly squashed.

include/picongpu/fields/incidentField/Functors.hpp
chillenzer: Thanks!

include/picongpu/fields/incidentField/profiles/Wavepacket.hpp
chillenzer: Okay, thanks!

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
chillenzer: Okay, that's the part that's not in the paper, right? Will be more clear once you added the 3D formulae to the docs.

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
chillenzer: :shrug:

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
chillenzer:
```C++
auto my_cool_calculation() {
    // ...
    return std::make_tuple(whatever, you, want);
}
auto [whatever, you, want] = my_cool_calculation();
```

BTW, reuse is typically not my main reason to extract functions. It is rather that it increases the coherence of abstraction within a function. Reading an algorithm described in, say, 3 to 5 function calls will give you a very good idea what's going on and will allow you to efficiently narrow your focus to the part of an algorithm you want to investigate. It is usually a sign that you have found a consistent level of abstraction and it allows a function to conveniently [fit into your working memory](https://en.wikipedia.org/wiki/The_Magical_Number_Seven,_Plus_or_Minus_Two).

But I typically only comment on this in PRs if I see also a benefit in code reuse because the above more nuanced reasoning is typically more opinionated and could be considered more in the realm of stylistic discussions. Still resolving this comment. Feel free to un-resolve, should you change your mind.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
chillenzer: See above.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
chillenzer: I agree. Do we find a better reference? I think my daughter recently scribbled something on a piece of paper that vaguely resembled some mathematical symbols. Shall I figure out how to cite that?

include/picongpu/fields/incidentField/profiles/DispersivePulse.hpp
chillenzer: Hey, 1970 just called: They are super-excited about their latest idea of [Common Subexpression Elimination](https://en.wikipedia.org/wiki/Common_subexpression_elimination). They say that maybe in a distant future compilers might be capable of applying that to real-world code. I wonder how long we're gonna have to wait for this.

steindev:
@PrometheusPi: The latest changes are force-pushed again and run through the CI now.

Second, here is the comparison between 32bit and 64bit regarding the `sqrt(sqrt(...))` discussion.

**32bit:**
GaussianPulse-refactor_32bit_normal-prop_2050

**64bit:**
GaussianPulse-refactor_64bit_normal-prop_2050

(no error in labels, really)

PrometheusPi:
@steindev Thanks for providing the comparison between 32bit and 64bit. This look excellent. 
Could you please also plot $|E_{32bit} - E_{64bit}| / E_{max}$? 
Do you know, what causes the spike at x=[700, 800] and y~30 ?

steindev:
@PrometheusPi

1) The difference plot
32bit-vs64bit


2) The spike: It lies directly on the incident field plane and does not propagate inside.

PrometheusPi:
Thanks @steindev Looks good.

PrometheusPi (approved):

PR #5320, opened 2025-03-19
Fix a bug in getParticlePosition where position is returned in PIC units instead of the expected SI units

ikbuibui:
@pordyna noticed a bug where when position is requested in SI units, it is actually returned in PIC units.
The bug has been fixed and an option to return the position in PIC units has been added


chillenzer (dismissed):
The `typename std::conditional_t<...` section is rather poetic.

ikbuibui:
Yes, I could have left it as auto but I thought maybe this is important documentation if the user wants to know the type

chillenzer (approved):

PR #5322, opened 2025-03-20
Binning Plugin: Fix output logic

franzpoeschel:
The binning plugin currently fails when trying to run a default LaserWakefield simulation. Two reasons:

1. The openPMD-api needs to know upon first flush that the Mesh is scalar, currently the definition for that comes too late
2. The functor must not be called when it is default-constructed

### DIFF
--- a/include/picongpu/plugins/binning/WriteHist.hpp
+++ b/include/picongpu/plugins/binning/WriteHist.hpp
@@ -131,9 +131,13 @@ namespace picongpu
                  * Write the histogram
                  */
                 ::openPMD::Mesh mesh = iteration.meshes["Binning"];
+                ::openPMD::MeshRecordComponent record = mesh[::openPMD::RecordComponent::SCALAR];
 
                 // Call the user defined OpenPMD
-                binningData.writeOpenPMDFunctor(series, iteration, mesh);
+                if(binningData.writeOpenPMDFunctor)
+                {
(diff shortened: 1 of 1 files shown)


pordyna:
Does it mean I need this PR if I have a non-empty  `writeOpenPMD`?  @franzpoeschel

franzpoeschel:
> Does it mean I need this PR if I have a non-empty `writeOpenPMD`? @franzpoeschel

If you don't run into errors, then probably not. I was not able to run a non-modified LaserWakefield simulation from dev.
In doubt, add `mesh[RecordComponent::SCALAR];` as the first line of your functor (so openPMD knows that this is scalar) and you should be fine.

ikbuibui (approved):
Thanks for the fixes, was this triggered by a recent change in openPMD?
I was able to reproduce the error.

franzpoeschel:
> Thanks for the fixes, was this triggered by a recent change in openPMD? I was able to reproduce the error.

I tried this against our dev branch and our latest release 0.16.1 from January, both ran into that error. Also the second error (checking if the functor is there) is unrelated to openPMD. 
Was the Binning plugin recently enabled by default in LaserWakefield setups? It might just be that we never checked if the Binning plugin works with a no-op user functor.

ikbuibui:
Yes, we merged a change ( #5312) this morning to the LWFA setup which causes this change in behaviour. Thanks for the catch

PR #5323, opened 2025-03-20
Checkpoint-restart: Fix reading of RNG states

franzpoeschel:
With the current release of the openPMD-api, this apparently works anyway, but with the dev version (precisely: after merging https://github.com/openPMD/openPMD-api/pull/1592) this throws an error.
With current versions of openPMD, random-access reading works via `series.iterations`, in future, there will be one unified API call `series.snapshots()` that works for writing and reading, streaming and random-access.

I don't know if this solves Alex's simulation freezes on Frontier, but it should fix the RNG state issue he saw.


ikbuibui (approved):

ikbuibui:
@psychocoderHPC @BeyondEspresso

PR #5324, opened 2025-03-21
Remove unused return statement that breaks compile

pordyna:
@ikbuibui, this return statement is never executed, but since it is not guarded by an `if  constexpr` block the compiler will try to ensure that the return type is correct. This will fail if either precision or unit is different from cell. Basically, `rellative_cellpos` is a vector of `int`, but for all other configurations the return type is a vector of `float_X`.


ikbuibui:
Thanks for the catch, I forgot to remove this return statement after having the correct return in the else block

ikbuibui (approved):

PR #5319, opened 2025-03-19
Add field binning to the binning plugin

ikbuibui:
Add the ability to bin fields in the binning plugin
Note : The ``domainInfo`` object for field binning is different from particle binning, as fields have no parallel for sub cell particle position,

PR #5325, opened 2025-03-21
update particle binner interface in the docs

pordyna:
Update documentation with changes intorduced by #5312


ikbuibui:
Thanks for fixes, #5312 was a part of a series of PRs (split from one large PR) and the changes you made are already in #5319, which should be merged soon @chillenzer

PR #5328, opened 2025-03-25
add warning message for force constant temperature

BrianMarre:
Adds a warning to console if the force constant electron temperature debug option is enabled.

This is intended to ensure that the user does not unwittingly leaves it active in production simulations and to make debugging a setup with atomicPhysics easier.


chillenzer (changes requested):
Where does this actually happen? Where's the functional code for this? I would prefer to have the warning close to that.

BrianMarre:
the functional code is [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/c19bf445efcc78d99b59f37b20a6f3cd43de91d1/include/picongpu/simulation/stage/AtomicPhysics.x.cpp#L296-L306) and uses [this](https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/include/picongpu/particles/atomicPhysics/SetTemperature.hpp).

Unfortunately I can't just add the warning there because it will then be printed for every mpi-thread and every atomic physics sub-step instead of just once per mpi-thread at simulation start.

chillenzer (approved):

PR #5329, opened 2025-03-25
Use lazy parsing for checkpoint-restart in the openPMD plugin

franzpoeschel:
Currently, the openPMD plugin opens the entire checkpoint series for restarting, which is not necessary since only one Iteration will be used for restarting. This enables lazy parsing by default.

Additionally, use `Attributable::seriesFlush()` instead of direct `Series::flush()`. When called from within an Iteration, openPMD-api >= 0.16.1 will guarantee that this Iteration is opened, no matter if modified or not.

~~This PR is also an attempt to workaround an IO issue that was observed in a restart workflow. Since the actual cause for the observed error is not yet known, please do not merge yet.~~ The cause for the error is now known and it has nothing to do with PIConGPU. Still, the first commit makes restarting more efficient, and the second commit makes it somewhat more resilient against errors, so we can still go ahead with this change.

### DIFF
--- a/include/picongpu/plugins/openPMD/NDScalars.hpp
+++ b/include/picongpu/plugins/openPMD/NDScalars.hpp
@@ -164,7 +164,7 @@ namespace picongpu
                 auto datasetName = baseName + "/" + group + "/" + dataset;
                 ::openPMD::Series& series = *params.openPMDSeries;
                 ::openPMD::MeshRecordComponent mrc
-                    = series.iterations[currentStep].meshes[baseName + "_" + group][dataset];
+                    = series.iterations[currentStep].open().meshes[baseName + "_" + group][dataset];
                 auto ndim = mrc.getDimensionality();
                 if(ndim != simDim)
                 {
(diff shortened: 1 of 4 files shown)


chillenzer:
You forgot to format your code. In the best case: Install and run pre-commit.

chillenzer (approved):
I don't quite understand the impact of these changes but they look consistent. As long as the CI passes, I'll trust your expertise.

franzpoeschel:
> You forgot to format your code. In the best case: Install and run pre-commit.

Fixed now. That best case requires fixing this error which PIConGPU's pre-commit config keeps giving me and I did not bother doing that so far

```
> git commit include/
An error has occurred: InvalidManifestError: 
==> File /home/franz/.cache/pre-commit/repoltqh6695/.pre-commit-hooks.yaml
==> At Hook(id='enforce-ascii')
==> At key: language
=====> Expected one of conda, coursier, dart, docker, docker_image, dotnet, fail, golang, haskell, lua, node, perl, pygrep, python, r, ruby, rust, script, swift, system but got: 'python_venv'
Check the log at /home/franz/.cache/pre-commit/pre-commit.log
```

Is that a known issue?

chillenzer:
> Is that a known issue?

Yes, that's known (to me at least) and the fix is straightforward (see https://github.com/ComputationalRadiationPhysics/picongpu/issues/5331). Just haven't got around to actually implement it yet.

franzpoeschel:
Ah alright, then I'll just use an older version of pre-commit until that is fixed

PR #5332, opened 2025-03-26
Add json formatter and replace ASCII enforcement in pre-commit

### LINKED ISSUE #5331: Fix enforce ASCII in pre-commit
The latest major version bump in `pre-commit` has introduced some breaking changes and the `enforce-ascii` hook that we're running is not yet taking those into account. The error message is as follows:

```
> git commit include/
An error has occurred: InvalidManifestError: 
==> File /home/franz/.cache/pre-commit/repoltqh6695/.pre-commit-hooks.yaml
==> At Hook(id='enforce-ascii')
==> At key: language
=====> Expected one of conda, coursier, dart, docker, docker_image, dotnet, fail, golang, haskell, lua, node, perl, pygrep, python, r, ruby, rust, script, swift, system but got: 'python_venv'
Check the log at /home/franz/.cache/pre-commit/pre-commit.log
```

We are not yet seeing this in CI because we're still running some 3.x version of `pre-commit` there.

We should probably switch to another hook because `enforce-ascii` has seen no updates in 5 years. Maybe https://github.com/jumanjihouse/pre-commit-hooks?tab=readme-ov-file#require-ascii would be an option.

_Originally posted by @franzpoeschel in https://github.com/ComputationalRadiationPhysics/picongpu/issues/5329#issuecomment-2753805963_

chillenzer:
Adds a json formatter because handling all those pypicongpu schemas became really annoying. Also tackles the outdated `enforce-ascii` thing.

For review: The first and last commit are third-party autoformatting commits that can be ignored during review.

Closes #5331

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,72 +1,72 @@
-minimum_pre_commit_version: 3.2.0  # necessitated by Lucas-C's hooks
+minimum_pre_commit_version: 3.2.0 # necessitated by Lucas-C's hooks
 default_install_hook_types: [pre-commit, pre-push]
-exclude: 'thirdParty'
+exclude: "thirdParty"
 repos:
-- repo: https://github.com/pre-commit/mirrors-clang-format
-  rev: v12.0.1
-  hooks:
-  - id: clang-format
-    types_or: [file]
-    files: \.(cpp|def|h|cu|tpp|kernel|loader|unitless|hpp|param)$
-- repo: https://github.com/pre-commit/pre-commit-hooks
-  rev: v4.5.0
-  hooks:
-  - id: no-commit-to-branch
-    args: [-b, dev]
-  - id: check-merge-conflict
-  - id: trailing-whitespace
-    exclude_types: [markdown, rst]
-  - id: end-of-file-fixer
-  - id: check-toml
-  - id: check-yaml
-  - id: mixed-line-ending
-  - id: check-executables-have-shebangs
-  - id: check-shebang-scripts-are-executable
-    exclude: \.tpl$
-- repo: https://github.com/Lucas-C/pre-commit-hooks
-  rev: v1.5.4
-  hooks:
-    - id: forbid-tabs
-      types_or: [file]
-      exclude_types: [rst, tex, tsv]
-    - id: remove-tabs
-      types_or: [file]
-      exclude_types: [rst, tex, tsv]
-    - id: forbid-crlf
-    - id: remove-crlf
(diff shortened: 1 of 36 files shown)


.pre-commit-config.yaml
```diff
@@ -70,3 +67,9 @@ repos:
       - id: ruff-format
         types_or: [python, pyi, jupyter]
         args: ["--line-length", "120"]
+  - repo: https://github.com/jumanjihouse/pre-commit-hooks
+    rev: 3.0.0
+    hooks:
+      - id: require-ascii
+        exclude_types: [rst]
```
ikbuibui, line 75: Offline discussion: This requirement only applied to text files, so we don't need to exclude all the file extensions we were excluding previously

ikbuibui (dismissed):

.pre-commit-config.yaml
chillenzer: See https://github.com/jumanjihouse/pre-commit-hooks/blob/38980559e3a605691d6579f96222c30778e5a69e/.pre-commit-hooks.yaml#L82

.zenodo.json
```diff
@@ -1,288 +1,288 @@
...
+      "name": "Garten, Marco",
+      "orcid": "0000-0001-6994-2475",
+      "type": "Other"
     },
     {
-      "affiliation":"HZDR, TU-Dresden",
-      "name":"Göthel, Ilja",
-      "type":"Other"
+      "affiliation": "HZDR, TU-Dresden",
+      "name": "G\u00f6thel, Ilja",
```
psychocoderHPC: This looks not correct!

.zenodo.json
```diff
@@ -1,288 +1,288 @@
...
-      "name":"Tiebel, Jessica"
+      "affiliation": "HZDR, TU Dresden",
+      "name": "Tiebel, Jessica"
     },
     {
-      "affiliation":"HZDR",
-      "name":"Widera, René",
-      "orcid":"0000-0003-1642-0459"
+      "affiliation": "HZDR",
+      "name": "Widera, Ren\u00e9",
```
psychocoderHPC: same here not sure what zenodoo is making out of this. I suggest excluding this wile from pre commit

psychocoderHPC (changes requested):

.zenodo.json
chillenzer: Well caught! I'll check that.

ikbuibui (approved):

psychocoderHPC (approved):

Issue #3464, opened 2020-12-08
Cannot merge old feature because of code reformattings


psychocoderHPC:
If you developed a feature before we switched to `clang-format` and you have issues with rebasing and merging the following easy :smiley: steps will solve your pain.

```bash
# Note: code below assumes clang-format-12 from llvm 12.0.1
git checkout dev
git fetch mainline   # assuming you have PIConGPU repo stored as remote named `mainline`
git rebase mainline/dev
cp .clang-format .. # save format file

git checkout <feature branch>  # <> is only used to show the name length
cp ../.clang-format .
find include/ share/picongpu/ share/pmacc -iname "*.def" \
  -o -iname "*.h" -o -iname "*.cpp" -o -iname "*.cu" \
  -o -iname "*.hpp" -o -iname "*.tpp" -o -iname "*.kernel" \
  -o -iname "*.loader" -o -iname "*.param" -o -iname "*.unitless" \
  | xargs clang-format-12 -i
git commit -a -m "clang-format"
rm .clang-format

git checkout <hash where feature branch split of>  # e.g. 00eb7516968fbc982e020b8e6895d772357b644b
git checkout -b topic-base
cp ../.clang-format .
find include/ share/picongpu/ share/pmacc -iname "*.def" \
  -o -iname "*.h" -o -iname "*.cpp" -o -iname "*.cu" \
  -o -iname "*.hpp" -o -iname "*.tpp" -o -iname "*.kernel" \
  -o -iname "*.loader" -o -iname "*.param" -o -iname "*.unitless" \
  | xargs clang-format-12 -i
git commit -a -m "clang-format"
rm .clang-format

git checkout <feature branch>
git diff topic-base..HEAD > ../feature.diff

git checkout dev
git checkout -b topic-newFeatureWithFormating
git apply ../feature.diff

# you have now all changes unstaged and format available
git status
# open a new PR or overwrite your old origin branch (please validate the new branch first)
```

@ComputationalRadiationPhysics/picongpu-developers

PrometheusPi:
@psychocoderHPC Thank you for posting this detailed description.

sbastrakov:
Note: for the simple case of adding new stuff and (almost) not modifying existing code lines, one can get away with just `git reset` -> `git add` -> `git clang-format` with maybe a couple of manual fixes. One can however use their judgement of what is and isn't a simple case.

BrianMarre:
[...]

PR #5334, opened 2025-03-28
Update clang format

### LINKED ISSUE #5315: Update clang-format version
We should definitely update the clang-format version. It's ancient. Also, it might resolve this weird formatting:

_Originally posted by @ikbuibui in https://github.com/ComputationalRadiationPhysics/picongpu/pull/5312#discussion_r1993193714_

chillenzer:
This commit updates `clang-format` to the latest version. This comes with a quite a few formatting changes (second commit).

As I know that everybody will whine and complain about this, I've also added a section in the docs to explain how you can easily rebase onto a change-of-formatting commit. It is a more modern version of #3464.

Closes #5315


ikbuibui (changes requested):
Needs an update of the clang-format file as well, to use new formatting arguments. This should be derived from the code style repo

chillenzer:
@franzpoeschel Specific task: Please focus on the documentation I've written. Do you think that's a good approach? It's a bit more manual than yours but more transparent and requires significantly less infrastructure.

docs/COMMIT.md
```diff
@@ -42,6 +42,73 @@ details. Some hints:
...
+git restore --source REBASE_HEAD .
+# If you decided against creating that file, paste the lines from above:
+bash apply-change-of-convention.sh
+git add -u
+# make sure you're doing the right thing:
+# git diff --staged
+git rebase --continue
+```
+
+Unfortunately, I haven't found a convenient way to automate this within an (interactive) rebase.
```
franzpoeschel, line 109: Your documentation reads very similar to the approach that is taken by [this script](https://github.com/franzpoeschel/reformat-rebase), so that would be the convenient automated way to do this. Should we link it in the docs?

Application would then be via

```bash
reformat-rebase.py --reformat bash ./apply-change-of-convention.sh --old-base ... --new-branch change-formatting
```

franzpoeschel:
> @franzpoeschel Specific task: Please focus on the documentation I've written. Do you think that's a good approach? It's a bit more manual than yours but more transparent and requires significantly less infrastructure.

Your documentation uses the same overall strategy as my script, so we seem to both have independently from each other concluded that this is the way to go, which seems good.
Your "implementation" of this strategy is a bit different: You use an interactive rebase, which you note yourself means that "No patch will apply cleanly". My script does things a bit more directly:

> For each reapplied commit `(c)`, the script will:
>
>   * Reapply the commit by calling `git checkout (c) -- ..`
>   * Reformat the commit by using the supplied script.
>   * Commit the changes by using the same commit message.

.. which has the downside that the commit messages need to be reapplied manually (not really a problem in a script).

Otherwise, I like to think that the documentation I wrote for https://github.com/franzpoeschel/reformat-rebase is quite helpful, so you also might want to check that out and maybe link the project or sth.

docs/COMMIT.md
chillenzer: I like your idea but I would suggest that we do so in a separate PR because I didn't have time to try it out and I don't want to put something into the docs that's not thoroughly tested.

ikbuibui (approved):
Looks good. I will merge and place the responsibility of bringing in `reformat-base` to a separate PR

PR #5335, opened 2025-03-28
fix position offsets not being calculated correctly in Binning Plugin

ikbuibui:
Bug found by Konrad Schlicke, particle and cell positions aren't calculated correctly when requested relative to `TOTAL` origin.


psychocoderHPC (approved):

ikbuibui:
@BrianMarre I think you probably encounter this bug

pordyna:
Oh unlucky, but good to know

PR #5337, opened 2025-03-31
Set the mesh unit dimension correctly if normalization by volume is enabled in the Binning plugin

ikbuibui:
Set the correct `unitDimension` in case normalization by bin volume is enabled. 
Note : `UnitSI` was set correctly even before this fix and has had no change
If normalization by bin volume is enabled in the binning plugin, this changes the units of the deposited quantity from `UNITS` to `UNITS/NORMALIZATION_UNITS`.  For example if we are binning energy over 2d space, the units change from `J` to `J/m^2`

### DIFF
--- a/include/picongpu/plugins/binning/WriteHist.hpp
+++ b/include/picongpu/plugins/binning/WriteHist.hpp
@@ -165,9 +165,7 @@ namespace picongpu
                 mesh.setGridSpacing(gridSpacingVector);
                 mesh.setGridGlobalOffset(gridOffsetVector);
 
-                {
-                    mesh.setUnitDimension(makeOpenPMDUnitMap(binningData.depositionData.units)); // charge density
-                }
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
Since we have a 2D3V simulation code and always have a 3D cell definition I would say 

> For example if we are binning energy over 2d space, the units change from J to J/m^2

This should be `J/m^3` too for the 3D case. The advantage would be that you can compare 2D and 3D simulations.
But I could be wrong in case the user is selecting the special dimensions.

pordyna:
@psychocoderHPC I think it should always just normalize over the user provided bins, regardless of the simulation dimension.

pordyna (approved):
Just keep in mind, we still have a problem that the overflow bins have a different unit

ikbuibui:
Closing this since normalization is now being removed with #5342

PR #5343, opened 2025-04-02
Remove circular imports in generic contexts

chillenzer:
This refactoring introduces a protocol for PyPIConGPU classes to self-register as part of a group like `Plugin`, `DensityProfile`, ...

The previous mechanism was built on circularly including their own subclasses inside of the body of their function and maintaining a manual dictionary of those mapping them to some given names. This mechanism was implemented in multiple places with slightly different naming. This PR
- unifies the naming
- replaces the manual generation of the dictionary by a self-registration mechanism
- removes the distinction between `get_rendering_context()` and various slightly different versions of `get_generic_*_context()`

The new protocol consists of two parts:
1. New "groups" (like `Plugin` or `DensityProfile`) are created by inheriting from `SelfRegisteringRenderedObject`.
2. Subclasses of those (like `PhaseSpace` or `EnergyHistogram`) define the variable `_name` to something reasonable. At instantiation of their class (not an instance/object of that class but the class itself) `__init_subclass__()` adds those names to a list.

The manual maintenance of a list of plugins, density profiles, etc. is not completely removed. The schemas of each of their subclasses are still maintained separately and the overarching schemas for `Plugin`, `DensityProfile`, ... still have to list all allowed specialisations but if we ever want to change this, we can do so in a later PR.

The PR is not yet squashed and rebased, but please start reviewing it. I'll do so once in the end if you wish.

### DIFF
--- a/lib/python/picongpu/pypicongpu/rendering/renderedobject.py
+++ b/lib/python/picongpu/pypicongpu/rendering/renderedobject.py
@@ -1,7 +1,7 @@
 """
 This file is part of PIConGPU.
 Copyright 2021-2024 PIConGPU contributors
-Authors: Hannes Troepgen, Brian Edward Marre
+Authors: Hannes Troepgen, Brian Edward Marre, Julian Lenz
 License: GPLv3+
 """
 
@@ -207,7 +207,7 @@ class RenderedObject:
         return serialized
 
     @staticmethod
-    def check_context_for_type(type_to_check: type, context: dict | None) -> None:
+    def check_context_for_type(type_to_check: type, context: dict | None):
         """
         check if the given context is valid for the given type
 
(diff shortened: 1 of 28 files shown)


lib/python/picongpu/pypicongpu/rendering/renderedobject.py
```diff
@@ -222,3 +222,99 @@ def check_context_for_type(type_to_check: type, context: dict | None) -> None:
...
+    _name = _dummy_name
+
+    @classmethod
+    def _register(cls):
+        if cls._name not in cls._names:
+            cls._names.append(cls._name)
+
+    def __init_subclass__(cls):
+        super().__init_subclass__()
+        if cls._name != cls._dummy_name:
```
ikbuibui, line 257: Should we fail more loudly here?

chillenzer: There's no "failure" here. The `if` condition allows for subclasses to decide for themselves if they want to register or not. If a subclass sets `_name`, it registers. Otherwise it doesn't (because `_name == _dummy_name`). The only thing we might want to check is the other way round: We might want to guard against registering the same name twice. This would be a likely scenario in cases where you inherit from a class that is already registered. But that doesn't currently happen. More generally: For now I've kept it simple in that I only register names. If need arises, we should go back and register classes instead. That would also make it very hard to accidentally register a class twice.

ikbuibui:
Looks okay. A non blocking suggestion is to add documentation highlighting the importance of not forgetting to set the name for a subclass. You can squash commits, and  i will merge it.

lib/python/picongpu/pypicongpu/rendering/renderedobject.py
```diff
@@ -222,3 +222,102 @@ def check_context_for_type(type_to_check: type, context: dict | None) -> None:
...
+
+class SelfRegistering:
+    """
+    A parent class the children of which will register their names in a list.
+
+    Any subclass that wants to register has to overwrite _name.
+    Its name will appear in _names afterwards.
+
+    Only leafs of the inheritance tree should have a _name.
+    Everything else is undefined behaviour.
```
chillenzer, line 236: The protocol is already described in this docstring. I've added the last paragraph about leafs during the squash.

ikbuibui (approved):

PR #5346, opened 2025-04-09
Refactor moving window to avoid redundant calculations

ikbuibui:
`getCurrentSlideInfo` was recomputing values which are constant for a simulation run every time was called.

### DIFF
--- a/include/picongpu/simulation/control/MovingWindow.hpp
+++ b/include/picongpu/simulation/control/MovingWindow.hpp
@@ -39,112 +39,134 @@ namespace picongpu
 
         MovingWindow(MovingWindow& cc) = delete;
 
+        struct ComputedConstants
+        {
+            uint32_t gpuNumberOfCellsInMoveDirection;
+            float_64 cellSizeInMoveDirection;
+            float_64 deltaWayPerStep;
+            uint32_t virtualParticleInitialStartCell;
+            uint32_t firstSlideStep;
+            int32_t firstMoveStep;
+        };
+
+        ComputedConstants calculateConstants() const
+        {
+            ComputedConstants constants;
+            SubGrid<simDim> const& subGrid = Environment<simDim>::get().SubGrid();
+
+            /* speed of the moving window */
+            auto const windowMovingSpeed = static_cast<float_64>(sim.pic.getSpeedOfLight());
+
+            /* the moving window is smaller than the global domain by exactly one
+             * GPU (local domain size)
+             */
+            uint32_t const globalWindowSizeInMoveDirection
+                = subGrid.getGlobalDomain().size[moveDirection] - subGrid.getLocalDomain().size[moveDirection];
+
+            constants.gpuNumberOfCellsInMoveDirection = subGrid.getLocalDomain().size[moveDirection];
+
+            /* unit PIConGPU length */
+            constants.cellSizeInMoveDirection = static_cast<float_64>(sim.pic.getCellSize()[moveDirection]);
+
+            constants.deltaWayPerStep = windowMovingSpeed * static_cast<float_64>(sim.pic.getDt());
+
+            /* How many cells the virtual particle with speed of light is pushed forward
+             * at the begin of the simulation.
+             * The number of cells is round up thus we avoid window moves and slides
+             * depends on half cells.
+             */
+            constants.virtualParticleInitialStartCell = math::ceil(
+                static_cast<float_64>(globalWindowSizeInMoveDirection) * (static_cast<float_64>(1.0) - movePoint));
+
+            /* Is the time step when the virtual particle **passed** the GPU next to the last
+             * in the current to the next step
+             */
+            constants.firstSlideStep
+                = static_cast<uint32_t>(math::ceil(
+                      static_cast<float_64>(
+                          subGrid.getGlobalDomain().size[moveDirection] - constants.virtualParticleInitialStartCell)
+                      * constants.cellSizeInMoveDirection / constants.deltaWayPerStep))
+                  - 1;
+
+            /* way which the virtual particle must move before the window begins
+             * to move the first time [in pic length] */
+            auto const wayToFirstMove
+                = static_cast<float_64>(globalWindowSizeInMoveDirection - constants.virtualParticleInitialStartCell)
+                  * constants.cellSizeInMoveDirection;
+            /* Is the time step when the virtual particle **passed** the moving window
+             * in the current to the next step
+             * Signed type of firstMoveStep to allow for edge case movePoint = 0.0
+             * for a moving window right from the start of the simulation.
+             */
+            constants.firstMoveStep = static_cast<int32_t>(math::ceil(wayToFirstMove / constants.deltaWayPerStep)) - 1;
+            return constants;
+        }
+
+        std::optional<ComputedConstants> computedConstants;
+
+        ComputedConstants const& getOrComputeConstants()
+        {
+            if(!computedConstants)
+            {
+                computedConstants = calculateConstants();
+            }
+            return *computedConstants;
+        }
+
+        // Has a side effect. May increment the slide counter
+        // It is assumed that this is only called after movePoint and endSlideOnStep are set
         void getCurrentSlideInfo(uint32_t currentStep, bool* doSlide, float_64* offsetFirstGPU)
         {
             if(doSlide)
                 *doSlide = false;
-
+            // Default offset before movement starts
             if(offsetFirstGPU)
-                *offsetFirstGPU = 0.0;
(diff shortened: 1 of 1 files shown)


include/picongpu/simulation/control/MovingWindow.hpp
```diff
@@ -39,112 +39,134 @@ namespace picongpu
...
-                *doSlide = false;
+            uint32_t gpuNumberOfCellsInMoveDirection;
+            float_64 cellSizeInMoveDirection;
+            float_64 deltaWayPerStep;
+            uint32_t virtualParticleInitialStartCell;
+            uint32_t firstSlideStep;
+            int32_t firstMoveStep;
+        };
+
+        ComputedConstants calculateConstants() const
```
chillenzer: I think this should be the constructor of `ComputedConstants`.

include/picongpu/simulation/control/MovingWindow.hpp
```diff
@@ -39,112 +39,134 @@ namespace picongpu
...
+             * Signed type of firstMoveStep to allow for edge case movePoint = 0.0
+             * for a moving window right from the start of the simulation.
+             */
+            constants.firstMoveStep = static_cast<int32_t>(math::ceil(wayToFirstMove / constants.deltaWayPerStep)) - 1;
+            return constants;
+        }
 
-                /* unit PIConGPU length */
-                auto const cellSizeInMoveDirection = float_64(sim.pic.getCellSize()[moveDirection]);
+        std::optional<ComputedConstants> computedConstants;
```
chillenzer: This is a pattern I feel to see more often recently. This is avoiding RAII in some way. Please write a comment to explain why this must be filled lazily.

include/picongpu/simulation/control/MovingWindow.hpp
```diff
@@ -39,112 +39,134 @@ namespace picongpu
...
 
-                float_64 const deltaWayPerStep = (windowMovingSpeed * float_64(sim.pic.getDt()));
+        ComputedConstants const& getOrComputeConstants()
+        {
+            if(!computedConstants)
+            {
+                computedConstants = calculateConstants();
+            }
+            return *computedConstants;
+        }
```
chillenzer: I would have loved to see this replaced by `.or_else()` but that's C++23. -.-

include/picongpu/simulation/control/MovingWindow.hpp
```diff
@@ -261,9 +291,10 @@ namespace picongpu
              * Later changes of step will not influence if the sliding window is activated.
              */
             if(firstCall && endSlidingOnStep != 0u)
+            {
```
chillenzer, line 294: Wait... was this a bug before? Now both bools are touched only inside the loop, before it was only one of them.

include/picongpu/simulation/control/MovingWindow.hpp
```diff
@@ -382,15 +413,15 @@ namespace picongpu
                 /* the moving window is smaller than the global domain by exactly one
                  * GPU (local domain size) in moving (y) direction
                  */
-                window.globalDimensions.size.y() -= subGrid.getLocalDomain().size.y();
+                window.globalDimensions.size[moveDirection] -= subGrid.getLocalDomain().size[moveDirection];
```
chillenzer, line 416: Why is there a change here? Was this a bug before or is `moveDirection` hard-coded to `y`?

chillenzer (changes requested):

include/picongpu/simulation/control/MovingWindow.hpp
ikbuibui: Yes, that would be quite appropriate

include/picongpu/simulation/control/MovingWindow.hpp
ikbuibui: Added. It is because these constants are not required if moving window is not enabled

include/picongpu/simulation/control/MovingWindow.hpp
ikbuibui: Good point. Done. The reason I missed it was that I initially didnt think of using an optional, and when I thought that leaving the struct uninitialized isnt great (even though that is also how this class deals with movePoint), I forgot to move it to the constructor.

include/picongpu/simulation/control/MovingWindow.hpp
ikbuibui: I wouldn't really call it a bug. Earlier `firstCall` was being set to `false` every time the function was called. This wasn't really required since it is sufficient to set it to `false` once, namely after the first call.

include/picongpu/simulation/control/MovingWindow.hpp
ikbuibui: `moveDirection` is hard-coded to y as a static constexpr. I just made the use more consistent as in someplaces `.y()` was used directly and other `moveDirection` was used.

include/picongpu/simulation/control/MovingWindow.hpp
```diff
@@ -39,112 +39,130 @@ namespace picongpu
...
                     uint32_t const virtualParticleWayPassedInCells
-                        = uint32_t(math::floor(virtualParticleWayPassed / cellSizeInMoveDirection));
+                        = uint32_t(math::floor(virtualParticleWayPassed / constants.cellSizeInMoveDirection));
                     uint32_t const virtualParticlePositionInCells
-                        = virtualParticleWayPassedInCells + virtualParticleInitialStartCell;
+                        = virtualParticleWayPassedInCells + constants.virtualParticleInitialStartCell;
 
                     /* calculate the position of the virtual particle after the current step is calculated */
-                    float_64 const nextVirtualParticleWayPassed = deltaWayPerStep * float_64(currentStep + 1);
+                    float_64 const nextVirtualParticleWayPassed = virtualParticleWayPassed + constants.deltaWayPerStep;
```
ikbuibui: This shouldn't increase floating point inaccuracy compared to the old method IMO, but it would be nice to get an explicit approval from someone else. @psychocoderHPC @chillenzer

chillenzer: Well, it's one further fp operation, so it's certainly not helping precision but I'd let that slip. @psychocoderHPC can decide.

chillenzer (dismissed):
For me this looks okay but @psychocoderHPC explicitly said that he also wants to check this code, I'll leave it to him to merge.

PrometheusPi (dismissed):

PrometheusPi:
@psychocoderHPC we are waiting for your final approval.

include/picongpu/simulation/control/MovingWindow.hpp
psychocoderHPC: You currently solution should have a higher numerical error because you add a small floating point number to a large value. The original solution has only one multiplication and therefore a  smaller error.

I think the original is the way to go.
```
float_64 const nextVirtualParticleWayPassed = constants.deltaWayPerStep * (static_cast<float_64>(currentStep+1));
```

psychocoderHPC (changes requested):

include/picongpu/simulation/control/MovingWindow.hpp
```diff
@@ -39,112 +39,130 @@ namespace picongpu
...
+            // Default offset before movement starts
+            if(offsetFirstGPU)
+                *offsetFirstGPU = 0; // Assume 0 offset if window hasn't started moving relative to particle frame
+
+            if(slidingWindowEnabled)
+            {
+                auto const& constants = getOrComputeConstants();
+
+                // Clamp currentStep if sliding has ended
+                if(currentStep >= endSlidingOnStep)
```
psychocoderHPC, line 127: we found a bug which was already i before and should be fixed after this PR is merged. IN case `endSlideingOnStep` is exactly the slide step we will slide in each following step.

fix

```C++
// Clamp currentStep if sliding has ended
uint32_t realTime=currentStep;
if(currentStep > endSlidingOnStep)
    currentStep = endSlidingOnStep;

//...
if(endOfInitialGlobalDomain && virtualParticlePassesGPUBorder && realTime <= endSlidingOnStep)
{
    incrementSlideCounter(currentStep);
    if(doSlide)
        *doSlide = true;
}
```

include/picongpu/simulation/control/MovingWindow.hpp
ikbuibui: reverted

include/picongpu/simulation/control/MovingWindow.hpp
ikbuibui: Will be fixed with #5358

psychocoderHPC (approved):

PR #5347, opened 2025-04-11
Binning plugin axis rework

ikbuibui:
Refactored both Linear and Log axes and added tests.
Detailed changes
- Bug fixes: 
    - The linear axis for integers had an integer division error due to a misplaced cast. This could lead to incorrect bin edges in the openPMD output for some axis splittings. This is now fixed.
   _Note_: In general the behaviour of the old linear axis in integral mode was susceptible to floating point errors in certain cases, so it has been reworked to use only integer operations. 
  **Warning**: The log axis suffers from floating point precision errors, so for certain combinations of min, max, nBins and values to be binned the bin index calculated might be off by one . If using an integral log axis, be careful of what you are doing.
  - The log axis bin edges were being converted from PIC units to SI units, but actually they were already in SI units, leading to an extra conversion factor being multiplied.  
- Feature addition:
The log axis can now work for negative ranges as well. This is possible because technically, the axis only defines geometrically spaced bins and doesn't face problems with taking a log of a negative number.
_Note_: Of course ranges including zero are still invalid inputs.
- Tests added:
Range min/max validity
Bin edge correctness, including unit conversion
Correct bin index calculation while binning (unit conversion not tested here)

### DIFF
--- a/docs/source/usage/plugins/binningPlugin.rst
+++ b/docs/source/usage/plugins/binningPlugin.rst
@@ -155,6 +155,10 @@ Currently implemented axis types
 ..            axis::createLog(cellY_splitting, cellPositionYDescription);
 
 
+.. attention::
+The log axis suffers from floating point precision errors, so for certain combinations of min, max, nBins and values to be binned the bin index calculated might be off by one. If using an integral log axis, be careful of what you are doing.
+
+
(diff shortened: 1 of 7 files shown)


chillenzer:
I already love this PR because it has tests, so I can actually see what it's supposed to do! :tada:

ikbuibui:
@pordyna This PR will fix the "super-SI" unit log axis edges bug you faced

psychocoderHPC (approved):

PR #5352, opened 2025-04-22
Checkpoint plugin picmi

mafshari64:
@chillenzer,
Please take a look at this PR. I created relevant files, compiled pic-build, and ran successfully via tbg.

### DIFF
--- a/lib/python/picongpu/picmi/diagnostics/checkpoint.py
+++ b/lib/python/picongpu/picmi/diagnostics/checkpoint.py
@@ -1,0 +1,128 @@
+"""
+This file is part of PIConGPU.
+Copyright 2021-2025 PIConGPU contributors
+Authors: Masoud Afshari
+License: GPLv3+
+"""
+
+from ...pypicongpu.output.checkpoint import Checkpoint as PyPIConGPUCheckpoint
+from .timestepspec import TimeStepSpec
+
+import typeguard
+from typing import Optional, Dict
+
+
+@typeguard.typechecked
+class Checkpoint:
+    """
+    Specifies the parameters for creating checkpoints in PIConGPU simulations.
+
+    This plugin saves simulation state snapshots at specified intervals,
+    allowing for simulation restarts or analysis.
+
+    Attention: ** At least one of period or timePeriod must be provided.**
+
+    Parameters
+    ----------
+    period: TimeStepSpec, optional
+        Specify on which time steps to create checkpoints.
+        Unit: steps (simulation time steps). Required if timePeriod is not provided.
+
+    timePeriod: float, optional
+        Specify the interval in minutes for creating checkpoints.
+        Unit: minutes. Required if period is not provided.
+
+    directory: str, optional
+        Directory inside simOutput for writing checkpoints (default: "checkpoints").
+
+    file: str, optional
+        Relative or absolute fileset prefix for checkpoint files.
+
+    restart: bool, optional
+        If True, restart simulation from the latest checkpoint.
+
+    tryRestart: bool, optional
+        If True, restart from the latest checkpoint if available, else start from scratch.
+
+    restartStep: int, optional
+        Specific checkpoint step to restart from.
+
+    restartDirectory: str, optional
+        Directory inside simOutput containing checkpoints for restart (default: "checkpoints").
+
+    restartFile: str, optional
+        Relative or absolute fileset prefix for reading checkpoints.
+
+    restartChunkSize: int, optional
+        Number of particles processed in one kernel call during restart.
+
+    restartLoop: int, optional
+        Number of times to restart the simulation after it finishes.
+
+    openPMD: Dict, optional
+        Dictionary of openPMD-specific settings (e.g., ext, json, infix).
+    """
+
+    def check(self):
+        if self.period is None and self.timePeriod is None:
+            raise ValueError("At least one of period or timePeriod must be provided")
+        if self.timePeriod is not None and self.timePeriod < 0:
+            raise ValueError("timePeriod must be non-negative")
+        if self.restartStep is not None and self.restartStep < 0:
+            raise ValueError("restartStep must be non-negative")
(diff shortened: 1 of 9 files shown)


share/picongpu/pypicongpu/examples/laser_wakefield/main.py
```diff
@@ -189,6 +189,16 @@
...
     ),
+    picmi.diagnostics.Checkpoint(
+        period=picmi.diagnostics.TimeStepSpec[::100],
+        #     timePeriod=10.0,          # Checkpoints every 10 minutes (optional if period is provided)
+        #     directory="checkpoints",  # Optional: default directory inside simOutput
+        #     file="my_checkpoint",     # Optional: fileset prefix
+        #     openPMD={
+        #        "ext": "h5",  # Optional: file extension for openPMD
+        #        "json": "{}"  # Optional: JSON configuration for openPMD
+        #    }
```
chillenzer: Why there a lot of commented out code here? Remove or clarify its purpose.

lib/python/picongpu/picmi/diagnostics/checkpoint.py
```diff
@@ -0,0 +1,129 @@
...
+        Relative or absolute fileset prefix for reading checkpoints.
+
+    restartChunkSize: int, optional
+        Number of particles processed in one kernel call during restart.
+
+    restartLoop: int, optional
+        Number of times to restart the simulation after it finishes.
+
+    openPMD: Dict, optional
+        Dictionary of openPMD-specific settings (e.g., ext, json, infix).
```
chillenzer, line 63: Is the content of this supposed to be more or less the same as the input to the openPMD plugin? If so, we should probably come back to this once we have an openPMD plugin and simply hand an instance of that.

lib/python/picongpu/picmi/diagnostics/checkpoint.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Parameters
+    ----------
+    period: TimeStepSpec, optional
+        Specify on which time steps to create checkpoints.
+        Unit: steps (simulation time steps). Required if timePeriod is not provided.
+
+    timePeriod: float, optional
+        Specify the interval in minutes for creating checkpoints.
+        Unit: minutes. Required if period is not provided.
```
chillenzer: Are you sure that this is allowed to be a floating point number? The docs read more like that should be an int.

lib/python/picongpu/pypicongpu/output/checkpoint.py
```diff
@@ -0,0 +1,67 @@
...
+            "restart": self.restart,
+            "tryRestart": self.tryRestart,
+            "restartStep": self.restartStep,
+            "restartDirectory": self.restartDirectory,
+            "restartFile": self.restartFile,
+            "restartChunkSize": self.restartChunkSize,
+            "restartLoop": self.restartLoop,
+            "openPMD": self.openPMD,
+        }
+        # print("Checkpoint serialized:", serialized)
```
chillenzer: Remove

chillenzer (changes requested):
Looks good. Some tiny clean-up and waiting for the CI.

lib/python/picongpu/picmi/diagnostics/checkpoint.py
mafshari64: I think float is a better choice. Using floats allows for more flexible checkpointing intervals, especially in short simulations or testing, e.g.,  every 0.5 minutes (30 seconds) is valid and useful.

share/picongpu/pypicongpu/examples/laser_wakefield/main.py
mafshari64: I thought maybe in the future a user wants to customize the files and folders and this clarifies better how to do that. I can remove.

lib/python/picongpu/pypicongpu/output/checkpoint.py
mafshari64: I done.

lib/python/picongpu/picmi/diagnostics/checkpoint.py
chillenzer: That is a valid opinion but does it work? Have you ran a simulation that way?

mafshari64: do you mean it is better to use an integer?

share/picongpu/pypicongpu/examples/laser_wakefield/main.py
chillenzer: That is a good idea but commented out code is not a good place to do so. This will easily go stale without anyone noticing and the next user actually trying this will be misled and despair. If you want to showcase what it can do, I'd suggest you uncomment all that code. If it's actually running in the examples, chances are we easily catch if some changes invalid the given examples. If you also want to show the minimal configuration, you can configure checkpoints on the other example as well.

lib/python/picongpu/picmi/diagnostics/checkpoint.py
chillenzer: It's not about better/worse. It's about: Does PIConGPU handle an input of `--timePeriod 0.7586947` or `--timePeriod 1.e-2` correctly? Have you checked this by running the code or can you point me to the corresponding line of C++ that parses this kind of input?

mafshari64: i found   ` ("checkpoint.timePeriod", po::value<std::uint64_t>(&checkpointPeriodMinutes), ` at [include/pmacc/simulationControl/SimulationHelper.cpp](https://github.com/ComputationalRadiationPhysics/picongpu/blob/6b85d1da58db1cd40aa057ee01ce4d90a5d5e926/include/pmacc/simulationControl/SimulationHelper.cpp#L318)`.

so I changed timePeriod to int.

chillenzer (approved):
Great! Thanks!

Gonna merge tomorrow morning when the CI has passed.

PR #5353, opened 2025-04-28
openpmd.py plugin for PICMI

mafshari64:
@chillenzer 
@BrianMarre


lib/python/picongpu/picmi/diagnostics/openpmd.py
```diff
@@ -0,0 +1,154 @@
...
+        if self.period is None:
+            raise ValueError("period is mandatory")
+        if self.ext is not None and self.ext not in ["bp", "h5", "sst"]:
+            raise ValueError("ext must be one of 'bp', 'h5', 'sst'")
+        if self.data_preparation_strategy is not None and self.data_preparation_strategy not in [
+            "doubleBuffer",
+            "adios",
+            "mappedMemory",
+            "hdf5",
+        ]:
```
BrianMarre: we might want to use Literals and enforce them via the type system in the constructor

lib/python/picongpu/picmi/diagnostics/openpmd.py
```diff
@@ -0,0 +1,154 @@
...
+        if not isinstance(self.source, Source):
+            raise ValueError("source must be a Source object")
+
+    def __init__(
+        self,
+        period: TimeStepSpec,
+        source: Optional[Source] = None,
+        range: Optional[str] = None,
+        file: Optional[str] = None,
+        ext: Optional[str] = None,
```
BrianMarre: the ext parameter should defined as a literal not a str

lib/python/picongpu/picmi/diagnostics/openpmd.py
```diff
@@ -0,0 +1,154 @@
...
+        infix: Optional[str] = None,
+        json: Optional[Union[str, Dict]] = None,
+        json_restart: Optional[Union[str, Dict]] = None,
+        data_preparation_strategy: Optional[str] = None,
+        toml: Optional[str] = None,
+        particle_io_chunk_size: Optional[int] = None,
+        file_access: Optional[str] = None,
```
BrianMarre: the default value slisted in the documentation do not match the actual constructor defaults

lib/python/picongpu/picmi/diagnostics/openpmd.py
```diff
@@ -0,0 +1,154 @@
...
+from .openpmd_sources.source import Source
+
+import typeguard
+from typing import Optional, Dict, Union
+
+
+@typeguard.typechecked
+class OpenPMD:
+    """
+    Specifies the parameters for the openPMD plugin in PIConGPU simulations.
```
BrianMarre: the PICMI implementation should not reference PIConGPU

lib/python/picongpu/picmi/diagnostics/openpmd.py
```diff
@@ -0,0 +1,154 @@
...
+
+    Parameters
+    ----------
+    period: TimeStepSpec
+        Specify on which time steps to output data.
+        Unit: steps (simulation time steps). Required.
+
+    source: Source, optional
+        Data sources and filters to dump (e.g., Source([ChargeDensity(filter="filterX"), "species_all"])).
+        Default: Source(["species_all", "fields_all"]).
```
BrianMarre: I am not a fan of allowing "str" specification of sources, that jsut adds another way to get the same results and I would prefer just one good way to specify stuff.

lib/python/picongpu/picmi/diagnostics/openpmd.py
```diff
@@ -0,0 +1,154 @@
...
+
+    range: str, optional
+        Contiguous range of cells per dimension to dump (e.g., ":,:,:").
+        Format: comma-separated ranges like "begin:end,begin:end,begin:end".
+        Default: ":,:,:"
```
BrianMarre: using string specifiers is fine for now, but we want to change to something similar to the TimeSpecStep in the future

lib/python/picongpu/picmi/diagnostics/openpmd.py
```diff
@@ -0,0 +1,154 @@
...
+    Attention: **period is mandatory.**
+
+    Parameters
+    ----------
+    period: TimeStepSpec
+        Specify on which time steps to output data.
+        Unit: steps (simulation time steps). Required.
+
+    source: Source, optional
+        Data sources and filters to dump (e.g., Source([ChargeDensity(filter="filterX"), "species_all"])).
```
BrianMarre: I am not sure why we need that accumulation object here, from my point of view we should pass a list of objects directly here.

lib/python/picongpu/picmi/diagnostics/openpmd_sources/charge_density.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@typeguard.typechecked
+class ChargeDensity:
+    """
+    Represents the ChargeDensity data source for the openPMD plugin in PIConGPU.
+
+    This class defines the charge density field, derived from particle species at runtime,
+    which can be output using the openPMD plugin. An optional filter can be applied to select
+    which particles contribute to the charge density.
```
BrianMarre: same problem as before, we should not refer to PIConGPU specific things on the PICMI side

lib/python/picongpu/picmi/diagnostics/openpmd_sources/source.py
```diff
@@ -0,0 +1,86 @@
...
+    This class aggregates individual data sources (e.g., ChargeDensity) or predefined
+    keywords (species_all, fields_all) to define the --openPMD.source parameter:
+    https://picongpu.readthedocs.io/en/latest/usage/plugins/openPMD.html
+
+    Parameters
+    ----------
+    sources: List[Union[str, <DataSourceClasses>]]
+        List of data sources, either as strings (e.g., "species_all", "fields_all") or
+        as data source objects (e.g., ChargeDensity, Density).
+    """
```
BrianMarre: @mafshari64 could you please comment on why you consider the aggregator functionality class necessary? I currently don' see why it is needed, it looks like the entire functionality could be moved to the openPMD plugin.
Instead I would use this class as an interface for all source implementations

lib/python/picongpu/picmi/diagnostics/openpmd_sources/bound_electron_density.py
```diff
@@ -0,0 +1,74 @@
...
+Authors: Masoud Afshari
+License: GPLv3+
+"""
+
+import typeguard
+from typing import Optional
+
+
+@typeguard.typechecked
+class BoundElectronDensity:
```
BrianMarre: this class should inherit from a common interface of all Sources for the openPMD plugin

lib/python/picongpu/picmi/diagnostics/openpmd_sources/bound_electron_density.py
```diff
@@ -0,0 +1,74 @@
...
+
+        Returns
+        -------
+        str
+            The dataset name with optional filter (e.g., "bound_electron_density" or
+            "bound_electron_density:filterX").
+        """
+        if self.filter:
+            return f"bound_electron_density:{self.filter}"
+        return "bound_electron_density"
```
BrianMarre: the conversion to strings should not happen on the PICMI side, that string is already PIConGPU specific

lib/python/picongpu/picmi/diagnostics/openpmd_sources/charge_density.py
```diff
@@ -0,0 +1,73 @@
...
+        Return the source string for use in --openPMD.source.
+
+        Returns
+        -------
+        str
+            The dataset name with optional filter (e.g., "charge_density" or "charge_density:filterX").
+        """
+        if self.filter:
+            return f"charge_density:{self.filter}"
+        return "charge_density"
```
BrianMarre: same problem here, the source string should happen at the earliest in PyPIConGPU, as the generated string is PIConGPU specific.
also the generated string is wrong, it should be `source_all`

lib/python/picongpu/pypicongpu/output/openpmd.py
```diff
@@ -0,0 +1,138 @@
...
+                Energy,
+                EnergyDensity,
+                EnergyDensityCutoff,
+                LarmorPower,
+                MacroCounter,
+                MidCurrentDensityComponent,
+                Momentum,
+                MomentumDensity,
+                WeightedVelocity,
+            ]
```
BrianMarre: I am not a fan of the many repeats of the all the different source,please switch to using interfaces and testing with isinstance for the interface

BrianMarre (changes requested):

BrianMarre:
fundamentally and in summary, the current PR lacks an common interface for all the current sources of the openPMD plugin and this interface is required.

lib/python/picongpu/picmi/diagnostics/openpmd.py
mafshari64: follwoing [here](https://picongpu.readthedocs.io/en/latest/usage/plugins/openPMD.html) we have

`--openPMD.source | Select data sources and filters to dump. Default is species_all,fields_all,
`
for this reason  self.source = source if source is not None else Source(["species_all", "fields_all"]). is it still wrong?

BrianMarre: that's not what I meant, for example
the default for `ext` is `bp`  following the documentation, but the constructor default argument is `None` it should be `bp` instead.
Please match the default arguments in the documentation to the defaults specified by the constructor.

And more general, most of the settings are currently PIConGPU specific. That is fine for now but we have to figure out a way to abstract that in the future

lib/python/picongpu/picmi/diagnostics/openpmd_sources/charge_density.py
```diff
@@ -0,0 +1,59 @@
...
+    filter: str, optional
+        Name of a filter to select particles contributing to the charge density.
+        Default: None (all valid particles).
+    """
+
+    filter = util.build_typesafe_property(typing.Optional[str])
+
+    def __init__(self, filter: typing.Optional[str] = None):
+        self.filter = filter
+        self.check()
```
BrianMarre: default filter should be `all`, as this is the PIConGPU standard. And since we will stick to str for now, I would like to avoid using Optional here as it only complicates the implementation.

In addition, this class is missing the option to configure the species contributing to this source, please add a list of species that should contribute to this source.

lib/python/picongpu/picmi/diagnostics/openpmd_sources/bound_electron_density.py
```diff
@@ -0,0 +1,59 @@
...
+    filter: str, optional
+        Name of a filter to select particles contributing to the bound electron density.
+        Default: None (all valid particles).
+    """
+
+    filter = util.build_typesafe_property(typing.Optional[str])
+
+    def __init__(self, filter: typing.Optional[str] = None):
+        self.filter = filter
+        self.check()
```
BrianMarre: same problem as charge density source, the user must be able to specify the species that will contribute to the derived field source

BrianMarre (changes requested):
general the current version is still missing the configuration of species contributing to the derived field sources.

lib/python/picongpu/picmi/diagnostics/openpmd.py
```diff
@@ -0,0 +1,151 @@
...
+    particle_io_chunk_size: int, optional
+        Size of particle data chunks for writing (in MiB).
+        Reduces host memory footprint for certain backends.
+        Default: None (backend-dependent).
+
+    file_access: Literal["create", "append"], optional
+        File access mode for writing.
+        Options: "create" (new files), "append" (for checkpoint-restart workflows).
+        Default: "create" (new files).
+    """
```
BrianMarre:
```suggestion
    """
    openPMD diagnostic output

    This diagnostic writes simulation data (base fields , derived fields and/or particles) to disk using the openPMD
    standard, with configurable periods, data sources and backend settings.

    @param period specification of the the time steps for data output, outputs will always be written at the end of a PIC time step.
    @param source list of data source objects to include in the dump (e.g., [ChargeDensity(filter="all")]),
       Setting this to None will cause an empty dump
    @param range contiguous range of cells to dump the base- and derived field for
       specification as comma-separated "begin:end" range for each dimension, i.e. "begin:end,begin:end,begin:end"
       Notes: Values will be clipped to the simulation box. Begin and/or end may be omitted to indicate the limit of the simulation box in the dimension. The Default value ":,::,:" indicates that fields should be dumped for all cells.
    @param file relative or absolute file path prefix for openPMD output files. Relative path are interpreted as relative to the simulation output directory, the default value None indicates the backend's default.
    @param ext file extension controlling the openPMD backend, options are "bp" (default backend ADIOS2), "bp4" (bp4 backend ADIOS2), "bp5" (bp5 backend ADIOS2), "h5" (HDF5), "sst" (ADIOS2/SST for streaming).
    @param infix filename infix for the iteration layout (e.g., "_%06T"), use "NULL" for the group-based layout, ext="sst" requires infix="NULL".
    @param json openPMD backend configuration as a JSON string, dictionary, or filename (filename must be prepended with "@").
    @param json_restart: backend-specific parameters for restarting, as a JSON string, dictionary, or filename (filenames must be prepended with "@").
   @param data_preparation_strategy: strategy for particle data preparation, options: "doubleBuffer" or "adios" (ADIOS2-based), "mappedMemory" or "hdf5" (HDF5-based), the default value None indicates the backend default
   @param toml path to a TOML file for openPMD configuration. Replaces the JSON or keyword configuration.
   @param particle_io_chunk_size size of particle data chunks used in writing (in MiB), reduces host memory footprint for certain backends, default  "None"  indicates the backend default.
    @param file_access file access mode for writing, options: "create" (new files), "append" (for checkpoint-restart workflows).
    """
```

BrianMarre: please use the doxygen annotations for the documentation.

BrianMarre (changes requested):
looks generally good, but documentation strings should use doxygen annotations

lib/python/picongpu/picmi/diagnostics/openpmd.py
```diff
@@ -36,31 +36,32 @@ class OpenPMD:
     range: str, optional
         Contiguous range of cells per dimension to dump (e.g., ":,:,:").
         Format: comma-separated ranges like "begin:end,begin:end,begin:end".
-        Default: None (all cells).
+        Default: ":,::,:" (all cells).
```
PrometheusPi: I think this is a typo, right? 
```suggestion
        Default: ":,:,:" (all cells).
```

lib/python/picongpu/picmi/diagnostics/openpmd.py
```diff
@@ -36,31 +36,32 @@ class OpenPMD:
...
     file: str, optional
         Relative or absolute file prefix for openPMD output files.
+        If relative, files are stored under the simulation output directory.
         Default: None (backend-dependent default).
 
     ext: Literal["bp", "h5", "sst"], optional
         File extension controlling the openPMD backend.
         Options: "bp" (ADIOS2), "h5" (HDF5), "sst" (ADIOS2/SST for streaming).
-        Default: None (backend-dependent).
+        Default: "bp" (ADIOS2 backend).
```
PrometheusPi: This is not ensured, as we might compile with hdf5 only

PrometheusPi: Do you really want to hard code your selection below here?

lib/python/picongpu/picmi/diagnostics/openpmd_sources/auto.py
```diff
@@ -0,0 +1,59 @@
+"""
+This file is part of PIConGPU.
+Copyright 2021-2024 PIConGPU contributors
```
PrometheusPi:
```suggestion
Copyright 2021-2025 PIConGPU contributors
```

lib/python/picongpu/picmi/diagnostics/openpmd_sources/auto.py
```diff
@@ -0,0 +1,59 @@
...
+import typing
+
+
+@typeguard.typechecked
+class Auto(SourceBase):
+    """
+    Represents a default data source for openPMD output in particle-in-cell simulations.
+
+    This class specifies a backend-specific default source for dumping simulation data
+    using the openPMD standard. For example, in some backends, it may dump all particle
```
PrometheusPi: Why only in some backends. As I understand it, backends are e.g. ADIOS or HDF5, but that doesn't determine what you write out.

lib/python/picongpu/picmi/diagnostics/openpmd_sources/auto.py
```diff
@@ -0,0 +1,59 @@
...
+    species and fields.
+
+    Parameters
+    ----------
+    filter: str, optional
+        Name of a filter to select data contributing to the source.
+        Default: None (backend-dependent).
```
PrometheusPi: Why is the filter backend dependent. Our filters are defined in PIConGPU.

lib/python/picongpu/picmi/diagnostics/openpmd_sources/bound_electron_density.py
```diff
@@ -1,59 +1,90 @@
 """
 This file is part of PIConGPU.
-Copyright 2021-2025 PIConGPU contributors
+Copyright 2021-2024 PIConGPU contributors
```
PrometheusPi:
```suggestion
Copyright 2021-2025 PIConGPU contributors
```

PrometheusPi: Why does it start in 2021? Was it copied from some file?

lib/python/picongpu/picmi/diagnostics/openpmd_sources/auto.py
PrometheusPi: Why does it start in 2021? Was it copied from somewhere?

lib/python/picongpu/picmi/diagnostics/openpmd_sources/charge_density.py
```diff
@@ -1,59 +1,90 @@
 """
 This file is part of PIConGPU.
-Copyright 2021-2025 PIConGPU contributors
+Copyright 2021-2024 PIConGPU contributors
```
PrometheusPi: same as above

lib/python/picongpu/pypicongpu/output/openpmd_source/bound_electron_density.py
```diff
@@ -0,0 +1,34 @@
+"""
+This file is part of PIConGPU.
+Copyright 2021-2024 PIConGPU contributors
```
PrometheusPi: same here

PrometheusPi (changes requested):
just a first review - no test yet

lib/python/picongpu/picmi/diagnostics/openpmd.py
mafshari64: I just mentioned `bp `as the default as it is in [documentation ](https://picongpu.readthedocs.io/en/latest/usage/plugins/openPMD.html) not hardcoded. users can specify what they want.

lib/python/picongpu/picmi/diagnostics/openpmd_sources/auto.py
mafshari64: sorry I forgot to change it. I will do

lib/python/picongpu/picmi/diagnostics/openpmd.py
mafshari64: thanks for review, 
a short question: in other picmi/diagnositcs files I used standard Python docstring format eg [lib/python/picongpu/picmi/diagnostics/macro_particle_count.py](https://github.com/mafshari64/picongpu/blob/topic-openpmd/lib/python/picongpu/picmi/diagnostics/macro_particle_count.py)  and not oxygen annotations. just for this file 
`lib/python/picongpu/picmi/diagnostics/openpmd.py` for documentation or do for all?

lib/python/picongpu/picmi/diagnostics/__init__.py
```diff
@@ -1,17 +1,12 @@
-"""
-This file is part of PIConGPU.
-Copyright 2024 PIConGPU contributors
-Authors: Julian Lenz, Masoud Afshari
-License: GPLv3+
-"""
-
```
chillenzer: Why's this removed?

lib/python/picongpu/picmi/diagnostics/__init__.py
```diff
@@ -1,17 +1,12 @@
...
 from .auto import Auto
 from .phase_space import PhaseSpace
 from .energy_histogram import EnergyHistogram
 from .macro_particle_count import MacroParticleCount
 from .png import Png
 from .timestepspec import TimeStepSpec
 from .checkpoint import Checkpoint
+from .openpmd import OpenPMD
+from .openpmd_sources.source_base import SourceBase
```
chillenzer, line 17: I don't think that this should be here. We don't want to expose this to the user.

lib/python/picongpu/picmi/diagnostics/openpmd.py
```diff
@@ -0,0 +1,108 @@
...
+
+    This diagnostic writes simulation data (base fields, derived fields and/or particles) to disk using the openPMD
+    standard, with configurable periods, data sources and backend settings.
+
+    @param period specification of the time steps for data output, outputs will always be written at the end of a PIC time step.
+    @param source list of data source objects to include in the dump (e.g., [ChargeDensity(filter="all")]),
+        Setting this to None will cause an empty dump
+    @param range contiguous range of cells to dump the base- and derived field for
+        specification as comma-separated "begin:end" range for each dimension, i.e. "begin:end,begin:end,begin:end"
+        Notes: Values will be clipped to the simulation box. Begin and/or end may be omitted to indicate the limit of the simulation box in the dimension. The Default value ":,::,:" indicates that fields should be dumped for all cells.
```
chillenzer: This should become a class similar to `TimeStepSpec`. But be aware that there are subtle differences in semantics that would need to be represented in that new class.

lib/python/picongpu/picmi/diagnostics/openpmd.py
```diff
@@ -0,0 +1,108 @@
...
+    @param particle_io_chunk_size size of particle data chunks used in writing (in MiB), reduces host memory footprint for certain backends, default "None" indicates the PIC code default.
+    @param file_writing file writing mode for writing, options: "create" (new files), "append" (for checkpoint-restart workflows).
+    """
+
+    def check(self):
+        """
+        Validate the provided parameters.
+        """
+        if self.period is None:
+            raise ValueError("period is mandatory")
```
chillenzer: This can't happen.

PrometheusPi:
@mafshari64 There seems to be conflicts with the current PR. See info.

chillenzer (changes requested):
Offline discussion: We need further testing before approval.

lib/python/picongpu/picmi/diagnostics/openpmd.py
mafshari64: Do you mean should I delete it?

lib/python/picongpu/picmi/diagnostics/__init__.py
mafshari64: so delete it?

lib/python/picongpu/picmi/diagnostics/openpmd.py
mafshari64: I retained for defensive programming probably in future other external users might subclass or modify behavior in unexpected ways. but if you want I can remove it.

mafshari64:
Hi Julian,

@chillenzer, I have finally solved the issues and finalized this branch.  In a separate PR I will add all CI tests for picim and pypicongpu. agree?

PrometheusPi:
@mafshari64 Please rebase your PR to the current `dev`. The branch has conflicts.

PrometheusPi:
@mafshari64 I do not see a fix of this PR yet, do you have problems resolving the conflict? This feature would be great to have in `dev`.

chillenzer:
Superseded by #5461 which includes these commits.

PR #5356, opened 2025-04-28
openPMD backend: Fix warning when using openPMD-api without ADIOS2

franzpoeschel:
It seems that no one really uses the openPMD-api without ADIOS2, but when you do, there is currently a warning printed from `include/picongpu/plugins/common/openPMDDefaultExtension.hpp`. This fixes that by rearranging one ifdef pragma a bit.


PrometheusPi:
There were some network issues that cause the CI to fail - I restarted the CI jobs that failed.

PrometheusPi (approved):

PR #5357, opened 2025-04-30
Remove self referential include

ikbuibui:


include/picongpu/fields/incidentField/profiles/BaseParam.def
```diff
@@ -19,6 +19,8 @@
 
 #pragma once
 
+#include "picongpu/defines.hpp"
```
chillenzer, line 22: Why's this?

chillenzer (approved):

include/picongpu/fields/incidentField/profiles/BaseParam.def
ikbuibui: The file was missing an include for using float_X. Sorry for sneaking it in

PR #5360, opened 2025-05-06
Add reference to paper from Eitan Y. Levine

finnolec:


PrometheusPi (dismissed):
Looks good. You could also reference them in the descriptive list above, but not needed.

finnolec:
> Looks good. You could also reference them in the descriptive list above, but not needed.

Done

PrometheusPi (approved):

PR #5368, opened 2025-05-19
control simulation lifetime

psychocoderHPC:
Objects stored as member in the SimulationStarter could life until the end of the main function. This ca create issues with the call to pmacc `finalize()` where the event system is stopped and the accelerator context is destroyed.

### DIFF
--- a/include/picongpu/main.x.cpp
+++ b/include/picongpu/main.x.cpp
@@ -44,25 +44,30 @@ int runSimulation(int argc, char** argv)
 {
     using namespace picongpu;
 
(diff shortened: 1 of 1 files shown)


ikbuibui:
Change looks good. In general, is there no better place to place the finalize? Maybe in the destructor of the base of SimulationStarter (ISimulationStarter)?

chillenzer (approved):

chillenzer:
This fixes an immediate problem. We can pick up the suggestion by @ikbuibui at another point.

PR #5370, opened 2025-05-19
Bug fix: Vacuum cells were multiplied by cell size twice in Gaussian profile

chillenzer:
The multiplication already happens a few lines above, so the removed second multiplication is wrong. In effect, the number of vacuum cells up front was too small.


ikbuibui (dismissed):
Good catch. A minor non blocking comment is that I am not a fan of the way the check is done by first converting to position units. Even though the post PR version is correct, if something is defined logically in cells, like `vacuumCellsY` in this case, it would be good to apply the condition also with a comparison done in cells directly, so something like `totalCellOffset.y() < ParamClass::vacuumCellsY`.

chillenzer:
That sounds reasonable. I'm gonna implement that.

ikbuibui (approved):

PR #5371, opened 2025-05-19
Add GriddedLayout to PICMI

chillenzer:
Adds a `GriddedLayout` to PICMI which translates into the `OnePosition` layout inside of PIConGPU. This is a partial implementation, only containing features necessary for reliably testing particle distributions.

### DIFF
--- a/lib/python/picongpu/picmi/layout.py
+++ b/lib/python/picongpu/picmi/layout.py
@@ -8,6 +8,8 @@ License: GPLv3+
 import picmistandard
 import typeguard
 
+from ..pypicongpu.species.operation.layout import Random, OnePosition
+
 
 @typeguard.typechecked
 class PseudoRandomLayout(picmistandard.PICMI_PseudoRandomLayout):
@@ -24,3 +26,41 @@ class PseudoRandomLayout(picmistandard.PICMI_PseudoRandomLayout):
         assert self.n_macroparticles_per_cell > 0, "at least one particle per cell required"
 
         # Note: Call PICMI check interface once available upstream
+
+    def get_as_pypicongpu(self):
+        return Random()
+
+
+@typeguard.typechecked
+class GriddedLayout(picmistandard.PICMI_GriddedLayout):
+    # note: is translated from outside, does not do any checks itself
+
+    def __init__(self, *args, **kwargs):
+        # The standard seems to have a typo here:
+        # PseudoRandomLayout allows for n_macroparticles_per_cell (with an s)
+        # while this one only has n_macroparticle_per_cell.
+        # We fix this here:
+        if "n_macroparticles_per_cell" in kwargs:
+            if len(args) == 0:
+                args = (kwargs.pop("n_macroparticles_per_cell"),)
+            else:
+                raise ValueError("You provided n_macroparticles_per_cell and an unnamed first arg.")
+        super().__init__(*args, **kwargs)
+        # The standard apparently has a typo in its interface.
+        self.n_macroparticles_per_cell = self.n_macroparticle_per_cell
+        if self.grid is not None:
+            raise NotImplementedError("Non-default grid is not implemented.")
+
+    def check(self):
+        """
+        check validity of self
+
+        if ok pass silently, raise on error
+        """
+        assert self.n_macroparticles_per_cell is not None, "macroparticles per cell must be given"
+        assert self.n_macroparticles_per_cell > 0, "at least one particle per cell required"
+
+        # Note: Call PICMI check interface once available upstream
+
+    def get_as_pypicongpu(self):
+        return OnePosition()
--- a/share/picongpu/pypicongpu/schema/species/operation/layout/layout.Layout.json
+++ b/share/picongpu/pypicongpu/schema/species/operation/layout/layout.Layout.json
@@ -1,0 +1,40 @@
+{
+  "$id": "https://registry.hzdr.de/crp/picongpu/schema/picongpu.pypicongpu.species.operation.layout.layout.Layout",
+  "description": "Any layout. Consists of (1) the type of layout used and (2) the actual data for that layout.",
+  "type": "object",
+  "required": [
+    "typeID",
+    "data"
+  ],
+  "unevaluatedProperties": false,
+  "properties": {
+    "typeID": {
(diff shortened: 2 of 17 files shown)


share/picongpu/pypicongpu/schema/species/operation/layout/layout.Layout.json
```diff
@@ -0,0 +1,40 @@
...
+  "required": [
+    "typeID",
+    "data"
+  ],
+  "unevaluatedProperties": false,
+  "properties": {
+    "typeID": {
+      "description": "Enum-equivalent of selected type. Note that only one entry should be marked as true, and all others as false, but that is not enforced by the schema.",
+      "type": "object",
+      "required": [
```
mafshari64, line 14: I think here for a single species in a single operation, you must choose one layout: random (particles placed randomly in a cell) or one_position (particles at a fixed position). Using both simultaneously for the same species/operation I think it is invalid, as they define conflicting placement strategies.
Different species or operations in the same simulation can use different layouts (e.g., Species A uses random, Species B uses one_position), but each layout instance is singular.

```
"typeID": {
  "oneOf": [
    {"properties": {"random": {"const": true}, "one_position": {"const": false}}, "required": ["random", "one_position"]},
    {"properties": {"random": {"const": false}, "one_position": {"const": true}}, "required": ["random", "one_position"]}
  ]
}
```

lib/python/picongpu/picmi/layout.py
```diff
@@ -24,3 +26,41 @@ def check(self):
...
+        if self.grid is not None:
+            raise NotImplementedError("Non-default grid is not implemented.")
+
+    def check(self):
+        """
+        check validity of self
+
+        if ok pass silently, raise on error
+        """
+        assert self.n_macroparticles_per_cell is not None, "macroparticles per cell must be given"
```
pordyna, line 60: What did you tell me about using `assert`s @chillenzer?

share/picongpu/pypicongpu/schema/species/operation/layout/layout.Layout.json
pordyna: I think @chillenzer commented on this in the description. Hmm, this is probably sth we should check with the schema, but the` oneOf` solution is quite repetitive. Do what you think is the best @chillenzer, maybe we can think about sth better later.

test/python/picongpu/quick/picmi/layout.py
```diff
@@ -14,20 +14,36 @@ class TestPicmiPseudoRandomLayout(unittest.TestCase):
...
-        """pseudo random layout is not translated directly itself"""
-        layout = picmi.PseudoRandomLayout(n_macroparticles_per_cell=7)
-        with self.assertRaises(AttributeError):
-            layout.get_as_pypicongpu()
-
     def test_invalid(self):
         """erros for invalid params entered"""
         with self.assertRaisesRegex(Exception, ".*per.*"):
             layout = picmi.PseudoRandomLayout(n_macroparticles=700)
+            layout.ppc = 1
```
pordyna, line 24: why do you need to set ppc to 1 here? What should throw the Exception, the constructor or the check method?

test/python/picongpu/quick/picmi/layout.py
```diff
@@ -14,20 +14,36 @@ class TestPicmiPseudoRandomLayout(unittest.TestCase):
...
+        layout = picmi.GriddedLayout(n_macroparticle_per_cell=7)
+        layout.ppc = 1
+        layout.check()
+
+    def test_invalid(self):
+        """erros for invalid params entered"""
+        with self.assertRaisesRegex(Exception, ".*per.*"):
+            layout = picmi.GriddedLayout(n_macroparticles=700)
+            layout.ppc = 1
+            layout.check()
```
pordyna, line 45: Same as above, why are we overriding the value with 1?

pordyna (changes requested):
Heey looks good in general, but I'm a bit confused about the picmi part of the tests

lib/python/picongpu/picmi/layout.py
pordyna: btw, this is non-blocking :D

share/picongpu/pypicongpu/schema/species/operation/layout/layout.Layout.json
chillenzer: I'm not an expert on schemas but given how we currently generate these, I think it would need a targeted manipulation rather than a mere oversight to arrive at something passing the schemas but violating our assumptions. My feeling is that there is an underlying problem about our general approach in pypicongpu of which this is a symptom. Finally, this is one of a few instances of this pattern (so much so that we just recently introduced the infrastructure to abstract that away), so this would definitely go into a separate PR.

test/python/picongpu/quick/picmi/layout.py
chillenzer: ~~Indeed this might not be needed. It's the wrong argument `n_macroparticles` as opposed to `n_macroparticles_per_cell` or something like that, that's triggering here (and already in the first line). I might have felt to put it there as to make sure that we're not triggering the "wrong" exception in a sense (although that can't happen here) but in either case, I certainly didn't spend much time thinking about this test.~~

Sorry, just realised that it's the `check` on the line below that's actually triggering. In this case, the `ppc=1` is needed to have a unique reason for failure.

test/python/picongpu/quick/picmi/layout.py
chillenzer: See above. ~~But in this case I'd have to look up if `check` is firing or if it's already in the constructor. If it's the `check`, then we should keep the `ppc=1` line for having a unique reason for failure here.~~

test/python/picongpu/quick/picmi/layout.py
pordyna: oh ok so it only checks the kwarg value not the member?

pordyna (approved):
Ok looks good, you can maybe add a comment to the test explaining why you set ppc there, and maybe move the `with ...` to only go around the check call. But I guess that is very optional.

psychocoderHPC (approved):

PR #5372, opened 2025-05-20
Update misc picmi stuff

chillenzer:
Minor updates/modernisations/... around PICMI and PyPIConGPU that I made in recent times.

- Allow using numpy arrays in grid input
- Make base density configurable (this adds infrastructure for having a heuristic but the heuristic is not yet implemented after there were concerns about the resource requirements of such)
- Update some license headers that I forgot to touch in previous PRs
- Removing the restriction to 4 cores in the build (`-j 4` -> `-j`)

### DIFF
--- a/share/picongpu/pypicongpu/schema/simulation.Simulation.json
+++ b/share/picongpu/pypicongpu/schema/simulation.Simulation.json
@@ -9,7 +9,7 @@
     },
     "base_density": {
       "type": "number",
-      "description": "value to normalise densities"
+      "description": "PIConGPU unit system reference number density, used to normalize macro particle momenta and mass"
     },
     "time_steps": {
       "type": "integer",
--- a/share/picongpu/pypicongpu/schema/simulation.Simulation.json
+++ b/share/picongpu/pypicongpu/schema/simulation.Simulation.json
@@ -7,9 +7,13 @@
       "type": "number",
       "description": "width of a single timestamp in s"
     },
+    "base_density": {
+      "type": "number",
+      "description": "value to normalise densities"
+    },
     "time_steps": {
       "type": "integer",
-      "minimum": 1,
+      "minimum": 0,
       "description": "total number of time steps to run for"
     },
     "typical_ppc": {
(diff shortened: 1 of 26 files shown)


test/python/picongpu/quick/picmi/simulation.py
```diff
@@ -181,7 +181,7 @@ def test_wrong_explicitly_set_typical_ppc(self):
         grid = get_grid(1, 1, 1, 64)
         solver = picmi.ElectromagneticSolver(method="Yee", grid=grid)
 
-        wrongValues = [0, -1, -15]
```
BrianMarre: Why did you remove the 0 as a wrong value for typicalPPC?

TypicalPPC should never be 0 since this will lead to our momentum and mass unit to be `NaN` due to a division by zero in the calculation of the typicalNumberParticlesPerMacroParticle, `typicalNumberParticlesPerMacroParticle = typicalDensity * cellVolume / typicalPPC`, [see here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/f8f705d9603e00257e41347e38f1fe8783a44b8c/include/picongpu/unitless/simulation.unitless#L479-L483)

share/picongpu/pypicongpu/schema/simulation.Simulation.json
```diff
@@ -7,9 +7,13 @@
       "type": "number",
       "description": "width of a single timestamp in s"
     },
+    "base_density": {
+      "type": "number",
+      "description": "value to normalise densities"
```
BrianMarre:
```suggestion
      "description": "PIConGPU unit system reference number density, used to normalize macro particle momenta and mass"
```

BrianMarre (changes requested):
one possible reversion in the tests, please fix or explain, otherwise looks good

test/python/picongpu/quick/picmi/simulation.py
chillenzer: Offline discussion: The case of 0 is now interpreted as requesting a sensible default. This was a byproduct of some internal refactoring and actually considered nice because users wouldn't have to do case distinctions for cases where the typical ppc is computed from a formula that might round down to (or otherwise arrive at) 0. But it's a weak preference on my side, so no hard feelings to revert this.

chillenzer: Done

BrianMarre (approved):

psychocoderHPC (approved):

PR #5373, opened 2025-05-20
Fix picmi png plugin

chillenzer:
There was a bug in our PNG template: For any configuration, there must be content in that file. That content is either provided by the PNG configuration if given and otherwise there needs to be some default. So, the correct condition is
```
if any( (plugin is png) for plugin in output):
    ... use configuration from png plugin...
else:
    ... use default...
```
(We are currently not handling the case where there's more than one PNG plugin given in the PICMI configuration. I suspect that that would be problematic as well. That would be adding a check for that somewhere in Python world but as we don't really want users to use this plugin anyways, it's fine with me if it has some rough edges.)


pordyna (approved):

pordyna:
@psychocoderHPC @chillenzer looks good

psychocoderHPC (approved):

PR #5376, opened 2025-05-22
Topic cylyndrical density profile

pordyna:
Cylindrical density distribution with optional exponential preplasma and arbitrary orientation and position in space. Used for cryogenic jet and wire simulations. PR Includes the PIConGPU + PICMI implementation. As agreed before, this should be the last predefined density distribution that will introduce codes on both sides. We should switch to PICMI only afterwards.  This was already used in some simulations, and it generates the correct distribution. The distribution  already accounts for mass conservation when including preplasma following the reduced radius formula in [Long et al.]( https://doi.org/10.1038/s42005-023-01473-w). This has been documented hopefully in all the docstrings and in the readthedocs.
Happy reviewing @chillenzer! :) 

I added some AI generated test based on the Foil distribution, they even managed to find one or two bugs! I'm not sure if the added mention in docs renders correctly, so we should take a look at the CI docs build.

### DIFF
--- a/docs/source/usage/picmi/intro.rst
+++ b/docs/source/usage/picmi/intro.rst
@@ -198,6 +198,37 @@ Parameters/Methods prefixed with ``picongpu_`` are PIConGPU-exclusive.
 
     If neither is set a warning is printed prompting for either of the options above.
 
+  **Density distribution**
+  In addition to the standard `picmi` distribution options we also support the following:
+  - ``CylyndricalDistribution``:
+    .. code=block::
+      picmi.CylyndricalDistribution(
+        *,
+        rms_velocity: Tuple[float, float, float] = (0, 0, 0),
+        directed_velocity: Tuple[float, float, float] = (0, 0, 0),
+        fill_in: bool = True,
+        density: float,
+        center_position: tuple[float, float, float],
+        radius: float,
+        cylinder_axis: tuple[float, float, float],
+        exponential_pre_plasma_length: float | None,
+        exponential_pre_plasma_cutoff: float | None,
+        lower_bound: Union[Tuple[float, float, float], Tuple[NoneType, NoneType, NoneType]] = (None, None, None),
+        upper_bound: Union[Tuple[float, float, float], Tuple[NoneType, NoneType, NoneType]] = (None, None, None),
+      ) -> None
+      Docstring:     
+      Describes a cylyndrical density distribution of particles with gaussian up-ramp
+      with a constant density region in between. It can have an arbitrary orientation
+      and position in space.
+      Will create the following profile:
+        n = density if r < reduced_radius
+        n is 0 or follows the exponential ramp if r > reduced_radius
+        n is 0 if r > reduced_radius + prePlasmaCutoff
+        the reduced_radius is equal = @f[\sqrt{R^2 -L^2} -L @f] 
+        with R - cylinder_radius and L - prePlasmaLength (scale length of the ramp)
+        the reduced radius ensures mass conservation
+
+    
 Ionization:
 ^^^^^^^^^^^
 The PIConGPU PICMI interface currently supports the configuration of ionization only through a picongpu specific PICMI extension, not the in the PICMI standard defined interface, due to the lack of standardization of ionization algorithm names in the PICMI standard.
--- a/docs/source/usage/picmi/intro.rst
+++ b/docs/source/usage/picmi/intro.rst
@@ -198,36 +198,6 @@ Parameters/Methods prefixed with ``picongpu_`` are PIConGPU-exclusive.
 
     If neither is set a warning is printed prompting for either of the options above.
 
-  **Density distribution**
-  In addition to the standard `picmi` distribution options we also support the following:
-  - ``CylyndricalDistribution``:
-    .. code=block::
-      picmi.CylyndricalDistribution(
-        *,
-        rms_velocity: Tuple[float, float, float] = (0, 0, 0),
-        directed_velocity: Tuple[float, float, float] = (0, 0, 0),
-        fill_in: bool = True,
-        density: float,
-        center_position: tuple[float, float, float],
-        radius: float,
-        cylinder_axis: tuple[float, float, float],
-        exponential_pre_plasma_length: float | None,
-        exponential_pre_plasma_cutoff: float | None,
-        lower_bound: Union[Tuple[float, float, float], Tuple[NoneType, NoneType, NoneType]] = (None, None, None),
-        upper_bound: Union[Tuple[float, float, float], Tuple[NoneType, NoneType, NoneType]] = (None, None, None),
-      ) -> None
-      Docstring:     
-      Describes a cylyndrical density distribution of particles with gaussian up-ramp
-      with a constant density region in between. It can have an arbitrary orientation
-      and position in space.
-      Will create the following profile:
-        n = density if r < reduced_radius
-        n is 0 or follows the exponential ramp if r > reduced_radius
-        n is 0 if r > reduced_radius + prePlasmaCutoff
-        the reduced_radius is equal = @f[\sqrt{R^2 -L^2} -L @f] 
-        with R - cylinder_radius and L - prePlasmaLength (scale length of the ramp)
-        the reduced radius ensures mass conservation
-
     
 Ionization:
 ^^^^^^^^^^^
--- a/include/picongpu/param/density.param
+++ b/include/picongpu/param/density.param
@@ -238,7 +238,7 @@ namespace picongpu
         /* definition of sphere profile with flanks */
         using SphereFlanks = SphereFlanksImpl<SphereFlanksParam>;
 
-        struct CylyndricalParam
+        struct CylindricalParam
         {
             /** Position of the center of the cylinder
              * x, y, z coordinates
@@ -274,8 +274,8 @@ namespace picongpu
             static constexpr float_64 prePlasmaCutoff_SI = 7.0e-8;
         };
 
-        /* definition of cylyndrical profile */
-        using Cylyndrical = CylyndricalImpl<CylyndricalParam>;
+        /* definition of cylindrical profile */
+        using Cylindrical = CylindricalImpl<CylindricalParam>;
 
         /* definition of checkerboard density profile with particles every fourth cell in all directions */
         using EveryFourthCell = EveryNthCellImpl<mCT::UInt32<4, 4, 4>>;
--- a/include/picongpu/param/density.param
+++ b/include/picongpu/param/density.param
@@ -238,6 +238,45 @@ namespace picongpu
         /* definition of sphere profile with flanks */
         using SphereFlanks = SphereFlanksImpl<SphereFlanksParam>;
 
+        struct CylyndricalParam
+        {
+            /** Position of the center of the cylinder
+             * x, y, z coordinates
+             * unit: meter
+             */
+            static constexpr float3_64 centerPosition_SI{8.0 - 6, 8.0 - 6, 1.0e-5};
+
+            /** Cylinder axis
+             * Vector pointing in the direction of the cylinder axis
+             * does not need to be normalized
+             * x, y, z coordinates
+             * unit: none
+             */
+            static constexpr float3_X jetAxis{0.0_X, 0.0_X, 1.0_X};
+
+            /** Radius of the cylinder
+             * unit: meter
+             */
+            static constexpr float_64 radius_SI = 2.2e-6;
+
+            //! factor to multiply BASE_DENSITY_SI by to get density
+            static constexpr float_X densityFactor = static_cast<float_X>(1.0_X);
+
+            // Preplasma:
+            //! Scale length of the exponential preplasma ramp
+            static constexpr float_64 prePlasmaLength_SI = 1.0e-8;
+
+            /** Prelasma ramp length
+             * Distance from the surface of the cylinder (reduced radius) to the point where the preplasma
+             * will be cut of and set to zero.
+             * unit: meter
+             */
+            static constexpr float_64 prePlasmaCutoff_SI = 7.0e-8;
+        };
+
+        /* definition of cylyndrical profile */
+        using Cylyndrical = CylyndricalImpl<CylyndricalParam>;
+
         /* definition of checkerboard density profile with particles every fourth cell in all directions */
         using EveryFourthCell = EveryNthCellImpl<mCT::UInt32<4, 4, 4>>;
 
@@ -293,6 +332,7 @@ namespace picongpu
             //! Default value to be used for cells with no corresponding file value
             static constexpr float_X defaultDensity = 0.0_X;
         };
+
 
 #if (ENABLE_OPENPMD == 1)
         //! Definition of density-from-file profile
--- a/include/picongpu/param/density.param
+++ b/include/picongpu/param/density.param
@@ -244,7 +244,7 @@ namespace picongpu
              * x, y, z coordinates
              * unit: meter
              */
-            static constexpr float3_64 centerPosition_SI{8.0 - 6, 8.0 - 6, 1.0e-5};
+            static constexpr float3_64 centerPosition_SI{8.0e-6, 8.0e-6, 1.0e-5};
 
             /** Cylinder axis
              * Vector pointing in the direction of the cylinder axis
--- a/include/picongpu/particles/densityProfiles/CylindricalImpl.def
+++ b/include/picongpu/particles/densityProfiles/CylindricalImpl.def
@@ -33,7 +33,7 @@ namespace picongpu
          * the reduced radius ensures mass conservation
          *
          * @tparam T_ParamClass structure with parameters,
-         *  requirements are documented by CylyniderParam in density.param
+         *  requirements are documented by CyliniderParam in density.param
          */
         template<typename T_ParamClass>
         struct CylindricalImpl;
--- a/include/picongpu/particles/densityProfiles/CylindricalImpl.hpp
+++ b/include/picongpu/particles/densityProfiles/CylindricalImpl.hpp
@@ -34,10 +34,10 @@ namespace picongpu
             template<typename T_SpeciesType>
             struct apply
             {
-                using type = CylyndricalImpl<ParamClass>;
+                using type = CylindricalImpl<ParamClass>;
             };
 
-            HINLINE CylyndricalImpl(uint32_t currentStep)
+            HINLINE CylindricalImpl(uint32_t currentStep)
             {
             }
 
--- a/include/picongpu/particles/densityProfiles/CylindricalImpl.hpp
+++ b/include/picongpu/particles/densityProfiles/CylindricalImpl.hpp
@@ -20,7 +20,6 @@
 #pragma once
 
 #include "picongpu/defines.hpp"
-#include "picongpu/simulation/control/MovingWindow.hpp"
 
 namespace picongpu
 {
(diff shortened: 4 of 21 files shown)


share/picongpu/pypicongpu/template/include/picongpu/param/density.param.mustache
```diff
@@ -146,6 +146,39 @@ namespace picongpu
...
+           {
+              static constexpr float3_64  centerPosition_SI{
+                  {{{profile.data.center_position_si.0.component}}},
+                  {{{profile.data.center_position_si.1.component}}},
+                  {{{profile.data.center_position_si.2.component}}}
+              };
+              static constexpr float3_X jetAxis{
+                  {{{profile.data.cylinder_axis.0.component}}}_X,
+                  {{{profile.data.cylinder_axis.1.component}}}_X,
+                  {{{profile.data.cylinder_axis.2.component}}}_X
```
BrianMarre: Just to mention it, there is a mustache syntax for list with variable length, see for example [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/7ebeee98def059f32968c1075617c7620aed19c3/share/picongpu/pypicongpu/template/include/picongpu/param/speciesDefinition.param.mustache#L67-L74), which avoids having to write every component by hand.

docs/source/usage/picmi/intro.rst
```diff
@@ -198,6 +198,37 @@ Parameters/Methods prefixed with ``picongpu_`` are PIConGPU-exclusive.
...
+      Describes a cylyndrical density distribution of particles with gaussian up-ramp
+      with a constant density region in between. It can have an arbitrary orientation
+      and position in space.
+      Will create the following profile:
+        n = density if r < reduced_radius
+        n is 0 or follows the exponential ramp if r > reduced_radius
+        n is 0 if r > reduced_radius + prePlasmaCutoff
+        the reduced_radius is equal = @f[\sqrt{R^2 -L^2} -L @f] 
+        with R - cylinder_radius and L - prePlasmaLength (scale length of the ramp)
+        the reduced radius ensures mass conservation
```
BrianMarre: please use a literal include here, while it might not read as nicely it avoids the documentation going out of snyc with the implementation

BrianMarre:
general comment, it's cylindrical not cylyndrical, see [here](https://www.merriam-webster.com/dictionary/cylindrical) ;)

chillenzer:
Please fix the CI. The remaining errors are now real errors.

docs/source/usage/picmi/intro.rst
chillenzer, line 1: To be honest, I don't think it makes much sense to record all our deviations from the PICMI standard in the current phase. These would be massive and ever-growing currently. We'll have to rework our whole documentation once we switch to (our interpretation of) PICMI as the main interface and at that point we'll figure out what to do about the docs. I'll not stop you from adding stuff here but an incomplete list might give the false impression that there's little deviation.

Also, it doesn't render correctly.

include/picongpu/param/density.param
```diff
@@ -238,6 +238,45 @@ namespace picongpu
         /* definition of sphere profile with flanks */
         using SphereFlanks = SphereFlanksImpl<SphereFlanksParam>;
 
+        struct CylyndricalParam
+        {
+            /** Position of the center of the cylinder
+             * x, y, z coordinates
+             * unit: meter
+             */
+            static constexpr float3_64 centerPosition_SI{8.0 - 6, 8.0 - 6, 1.0e-5};
```
chillenzer: Isn't that like infinitely far away relative to our typical setups? It kinda looks like you might have forgotten an `e` here.

include/picongpu/param/density.param
```diff
@@ -238,6 +238,45 @@ namespace picongpu
...
+             */
+            static constexpr float3_X jetAxis{0.0_X, 0.0_X, 1.0_X};
+
+            /** Radius of the cylinder
+             * unit: meter
+             */
+            static constexpr float_64 radius_SI = 2.2e-6;
+
+            //! factor to multiply BASE_DENSITY_SI by to get density
+            static constexpr float_X densityFactor = static_cast<float_X>(1.0_X);
```
chillenzer, line 263: I see how this is (a) the general convention everywhere in this file, (b) probably convenient for some use cases and (c) probably numerically more stable. But a density in SI units would have felt more natural to me. Not actionable.

include/picongpu/particles/densityProfiles/CylyndricalImpl.def
```diff
@@ -0,0 +1,41 @@
+/* Copyright 2014-2024 Rene Widera, Sergei Bastrakov
```
chillenzer:
```suggestion
/* Copyright 2014-2025 Rene Widera, Sergei Bastrakov, Pawel Ordyna
```

chillenzer: Funnily I've put you into an unsolvable conundrum here: By applying my change, I'd kinda be author, too, so you'd have to add me to the file as well. But then again, doing so wouldn't be applying my suggestion which means putting me there would be wrong again. Let's see what you'll do about this!

chillenzer: Also, do this for all license headers. (There's more.)

include/picongpu/particles/densityProfiles/CylyndricalImpl.hpp
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PIConGPU.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "picongpu/defines.hpp"
+#include "picongpu/simulation/control/MovingWindow.hpp"
```
chillenzer: Why's that in here?

lib/python/picongpu/picmi/distribution/CylyndricalDistribution.py
```diff
@@ -0,0 +1,100 @@
...
+    #   all attributes as keyword arguments
+
+    # @note user may add additional attributes by hand, these will be available but not type verified
+
+    def get_as_pypicongpu(self) -> species.operation.densityprofile.Cylinder:
+        util.unsupported("fill in not active", self.fill_in, True)
+
+        # @todo support bounds, Brian Marre, 2024
+        util.unsupported("lower bound", self.lower_bound, (None, None, None))
+        util.unsupported("upper bound", self.upper_bound, (None, None, None))
```
chillenzer: Why's that?

lib/python/picongpu/picmi/distribution/CylyndricalDistribution.py
```diff
@@ -0,0 +1,100 @@
...
+
+        profile = species.operation.densityprofile.Cylinder()
+
+        if self.density <= 0.0:
+            raise ValueError("density must be > 0")
+        if self.radius <= 0.0:
+            raise ValueError("radius must be > 0")
```
chillenzer: Why do you exclude 0 here?

lib/python/picongpu/pypicongpu/species/operation/densityprofile/cylinder.py
```diff
@@ -0,0 +1,59 @@
...
+        if self.radius_si <= 0.0:
+            raise ValueError("radius must be > 0")
+        self.pre_plasma_ramp.check()
+
+    def _get_serialized(self) -> dict:
+        self.check()
+
+        return {
+            "density_si": self.density_si,
+            "center_position_si": [{"component": x} for x in self.center_position_si],
```
chillenzer, line 70: Why not?
```suggestion
            "center_position_si": {key: val for key, val in zip("xyz",self.center_position_si)},
```

lib/python/picongpu/pypicongpu/species/operation/densityprofile/cylinder.py
```diff
@@ -0,0 +1,59 @@
...
+            "radius_si": self.radius_si,
+            "cylinder_axis": [{"component": x} for x in self.cylinder_axis],
```
chillenzer, line 72: Same

share/picongpu/pypicongpu/template/include/picongpu/param/density.param.mustache
chillenzer, line 1: Why's there lots of whitespaces added in this file?

share/picongpu/pypicongpu/template/include/picongpu/param/density.param.mustache
chillenzer: I think it's fine like that. If the number is three (and that's fixed), it will not exactly foster readability to dive deeply into mustache syntax. And with this formatting, typos would be quite easily visible. A good editor will also let you edit these collectively.

include/picongpu/particles/densityProfiles/CylyndricalImpl.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+namespace picongpu
+{
+    namespace densityProfiles
+    {
+        template<typename T_ParamClass>
+        struct CylyndricalImpl : public T_ParamClass
```
BrianMarre, line 30: typo in class and file name

include/picongpu/particles/densityProfiles/CylyndricalImpl.def
```diff
@@ -0,0 +1,41 @@
...
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+namespace picongpu
+{
+    namespace densityProfiles
+    {
+        /** Cylyndrical density profile, optionally with a gaussian preplasma ramp
```
BrianMarre:
```suggestion
        /** Cylindrical density profile, optionally with a gaussian preplasma ramp
```

include/picongpu/particles/densityProfiles/CylyndricalImpl.def
```diff
@@ -0,0 +1,41 @@
...
+         * n is 0 if r > reduced_radius + prePlasmaCutoff
+         * the reduced_radius is equal = @f[\sqrt{R^2 -L^2} -L @f]
+         * with R - cylinder_radius and L - prePlasmaLength (scale length of the ramp)
+         * the reduced radius ensures mass conservation
+         *
+         * @tparam T_ParamClass structure with parameters,
+         *  requirements are documented by CylyniderParam in density.param
+         */
+        template<typename T_ParamClass>
+        struct CylyndricalImpl;
```
BrianMarre:
```suggestion
        struct CylindricalImpl;
```

include/picongpu/particles/densityProfiles/CylyndricalImpl.hpp
BrianMarre:
```suggestion
        struct CylindricalImpl : public T_ParamClass
```

BrianMarre: and also remember to fix the file name to match

lib/python/picongpu/picmi/distribution/CylyndricalDistribution.py
```diff
@@ -0,0 +1,100 @@
...
+from ...pypicongpu import util
+
+from .Distribution import Distribution
+
+import typeguard
+import typing
+
+
+@typeguard.typechecked
+class CylyndricalDistribution(Distribution):
```
BrianMarre:
```suggestion
class CylindricalDistribution(Distribution):
```

test/python/picongpu/quick/picmi/distribution.py
```diff
@@ -540,3 +540,117 @@ def test_drift(self):
...
+    ):
+        return picmi.CylyndricalDistribution(
+            density=density,
+            center_position=center_position,
+            radius=radius,
+            cylinder_axis=cylinder_axis,
+            exponential_pre_plasma_length=exponential_pre_plasma_length,
+            exponential_pre_plasma_cutoff=exponential_pre_plasma_cutoff,
+        )
+
```
chillenzer: I don't see a point in inheriting from `HelperTestPicmiBoundaries` if all its tests are skipped. The necessity to hide the constructor call in a separate method obfuscates what's happening in the tests I believe. But this is not blocking. I'll tackle that in a later refactoring.

include/picongpu/particles/densityProfiles/CylyndricalImpl.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+                float3_X globalCellPos;
+                if constexpr(simDim == 2)
+                {
+                    globalCellPos = float3_X{globalCellPosSimDim[0], globalCellPosSimDim[1], 0.0_X};
+                }
+                else
+                {
+                    globalCellPos = globalCellPosSimDim;
+                }
```
chillenzer: Not all compilers are happy with this. Please fix.

chillenzer (changes requested):

include/picongpu/param/density.param
pordyna: oh yes :D

lib/python/picongpu/picmi/distribution/CylyndricalDistribution.py
pordyna: We don't support it in any Distribution as far as I know. This, I think, comes from `warpx`, where they create particles everywhere within the bounds. In picongpu this is not needed as we have the minimal weighting and wherever the density is zero, no particles are created.

lib/python/picongpu/picmi/distribution/CylyndricalDistribution.py
pordyna: Why would you like to have a cylinder with no radius?

chillenzer: As a limiting and/or reference case:
```python
for r in np.linspace(0,10):
    density = CylindricalDistribution(r)
    [...]
    sim.step(1000)
```
is clean and straightforward while
```
for r in np.linspace(1, 10):
    density = CylindricalDistribution(r)
    [...]
    sim.step(1000)
# do something else where we don't add a density
[...]
sim.step(1000)
```
requires an extra step and is way less uniform. In particular, you would always be in doubt if your limiting case is, in fact, identical to your other cases save that one value you actually want to have changed.

lib/python/picongpu/pypicongpu/species/operation/densityprofile/cylinder.py
pordyna: Ah, I think this is something that we need to do for variable length vectors. Or did this change? But yeah here it could be simplified.

lib/python/picongpu/picmi/distribution/CylyndricalDistribution.py
chillenzer: Then the question is: Why do you even have these arguments at all? Your density doesn't appear in the standard anyway, so compatibility is not an issue. You don't inherit from anything that has these arguments. Just remove it and feel the relief of just writing the code that actually runs.

share/picongpu/pypicongpu/template/include/picongpu/param/density.param.mustache
pordyna: ha! no idea :D will fix it

lib/python/picongpu/picmi/distribution/CylyndricalDistribution.py
pordyna: This is a great idea!

BrianMarre:
@pordyna heads up, the file names are still `CylyndircalDistribution.py`, `CylyndricalImpl.hpp` and `CylyndricalImpl.def`

lib/python/picongpu/pypicongpu/species/operation/densityprofile/cylinder.py
chillenzer: Ah, I see. That might be. First of all, we currently don't support 2D (but that's hopefully not true in the long run). Also it depends on your rendering. Probably, it wouldn't hurt to render an unused or defaulted value for `z` in the 2D case (it might even be necessary),  so you could go with a fixed-size dictionary from here on. (Actually, the `zip` would terminate after the shortest iterator is exhausted, so you would have to potentially fill up `self.center_position_si` with a default value.) You could also hack together a schema that optionally has the third component and check its existence before using it during rendering. 

But all in all that's not a blocker. Focus your energy on the more important aspects here.

pordyna:
There are still a few things I need to fix now, I may have also just broken some imports.

docs/source/usage/picmi/intro.rst
pordyna: removed it

include/picongpu/particles/densityProfiles/CylyndricalImpl.hpp
pordyna: copy paste, do I still need pragma once and defines.hpp?

chillenzer: Thanks. Haven't been through your code again but I'd strongly presume you do.

test/python/picongpu/quick/picmi/distribution.py
pordyna: ok

share/picongpu/pypicongpu/template/include/picongpu/param/density.param.mustache
pordyna: There were weird 2 spaces intendeds there in the mainline file, I think this confused my IDE. I went for consistent 4 spaces now.

pordyna:
"Could not resolve host gitlab.com", can someone restart the one failed CI compile job?
@psychocoderHPC @chillenzer

share/picongpu/pypicongpu/template/include/picongpu/param/density.param.mustache
pordyna: @chillenzer  take a look if it is ok now, I tried to make it consistent

pordyna:
@chillenzer still missing license headers, sorry. But the rest should be ok

include/picongpu/particles/densityProfiles/CylindricalImpl.def
```diff
@@ -33,9 +33,9 @@ namespace picongpu
          * the reduced radius ensures mass conservation
          *
          * @tparam T_ParamClass structure with parameters,
-         *  requirements are documented by CylyniderParam in density.param
+         *  requirements are documented by CyliniderParam in density.param
```
chillenzer, line 36: There's still a typo here!

include/picongpu/particles/densityProfiles/CylindricalImpl.hpp
```diff
@@ -51,15 +50,17 @@ namespace picongpu
...
-                }
+                    if constexpr(simDim == 2)
+                    {
+                        return float3_X{globalCellPosSimDim[0], globalCellPosSimDim[1], 0.0_X};
+                    }
+                    else
+                    {
+                        return globalCellPosSimDim;
+                    }
+                }();
```
chillenzer, line 63: Cool, that worked! Good idea!

share/picongpu/pypicongpu/template/include/picongpu/param/density.param.mustache
chillenzer: There's still something whitespace-y going on here but whatever... the formatting wasn't great before, so it's fine.

chillenzer (approved):
Okay, I'll merge as is. There's some tiny typos in the comments and the license headers, so I could say something utopian like "Please come back to this later" but well... it's merged.

PR #5377, opened 2025-05-22
Fix some typos in paths in perlmutter profile

pordyna:


pordyna:
The fail reason looks CI related, can we restart it @psychocoderHPC?

PrometheusPi (approved):

PR #5349, opened 2025-04-16
Free formula density

chillenzer:
DRAFT: This compiled just fine but I haven't got around to extensive testing and exploration yet.

PR #5378, opened 2025-05-23
Free formula density for PICMI

chillenzer:
No longer a draft PR after
- [x] #5373
- [x] #5371
- [x] Rebase onto latest dev.

But still taking over from #5349.

----

### Main feature:
This PR adds support for providing arbitrary particle densities as Python functions in our implementation of the PICMI interface. For now, the signature of a density function is that it takes `(x, y, z)` spatial coordinates in SI units and returns a float (kind of, see below) representing the density in SI units. Due to implementation details the given function must be valid if given sympy symbols as arguments which basically means that you have to use sympy functions to express the density. This is not a significant restriction because you can basically use everything you can express in a mathematical formula. For any expressions not involving `x`, `y` or `z`, you can even use normal control flow. A variety of densities is implemented in `test/python/picongpu/end-to-end/free_formula_density.py`. Feel free to look around there. For more information documentation currently exists in the docstring of `AnalyticDistribution`.

### Additional features:
- Every PICMI distribution currently implemented has acquired a call operator. You can use it evaluate your density at an arbitrary position and the arguments follow numpy broadcasting rules (so, you can provide numpy arrays to efficiently evaluate a whole grid).
- Our only implemented grid acquired a `get_cell_size` function.
- The Gaussian distribution's `vacuum_(cells_)front` is now a float value in SI units in the frontend. This is still cast to an integer during conversion into PIConGPU code but having it in SI is more consistent as an interface.

### Testing:
- There are some superficial unit tests added.
- Some unit tests needed changes due to (internal) interface changes.
- I added the first "end-to-end" test. It sets up a simulation with a lot of particle species using different densities, writes a checkpoint and compares the densities to the functional form implemented in Python. Furthermore, I implemented all the predefined densities as free formulae again and compared the particle content of the predefined and free-formula species. 

### Caveats and implementation surprises:
There's a weird inconsistency in how densities are evaluated in PIConGPU. Predefined densities get a cell index and typically choose to evaluate their function at the origin of the cell. Free-formula functors get a position in SI units which corresponds to the centre of the cell. This entails some nasty surprises: Most importantly, the call operators of the predefined densities must be aware of the cell size in order to accurately reproduce the same density. I have added an ugly workaround that provides this knowledge on a best effort basis and otherwise raises an exception informing the user about a solution. The mid-term plan is to change all predefined densities' implementations to rely on free-formula functors in the backend such that this problem should disappear in the future. That future change might, however, entail shifts in particle densities by half a cell size. Also, all the free-formula densities implemented for testing have those shifts. Be careful when you copy-paste stuff!

### For review:
The actual contribution of functional code is relatively small: The main feature lives in `picmi/distribution/AnalyticDistribution.py`, `pypicongpu/species/operation/densityprofile/free_formula.py` and `pypicongpu/rendering/pmaccprinter.py` (with the usual schema and template adjustments/additions). The minor features are where you'd expect them. Testing is a large part because I implemented quite a number of comparisons and also had to set up the infrastructure. To get a feeling for the feature, it's probably best to start with `end-to-end/free_formula_density.py`. Afterwards, I'd work my way through the call stack of the `AnalyticDistribution.py` (so, `free_formula.py`, `pmaccprinter.py`, and so on...) and spread from there.


lib/python/picongpu/pypicongpu/rendering/pmaccprinter.py
```diff
@@ -0,0 +1,35 @@
...
+
+from sympy import S
+from sympy.printing.cxx import cxx_code_printers
+
+PMACC_MATH_FUNCTIONS = {
+    "Abs": "abs",
+}
+
+
+class PMAccPrinter(cxx_code_printers["c++17"]):
```
BrianMarre, line 16: PIConGPU already uses c++20, since #5246, is there a specific reason why we use the c++17 version?

BrianMarre:
very cool!
Just one minor thing ... The analytic distribution tests [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/f105f05acbb7f7840ba0f19ff54f8e0cb500ceeb/test/python/picongpu/quick/picmi/distribution.py#L117-L196) seem to have slipped through your net. ;)

lib/python/picongpu/pypicongpu/rendering/pmaccprinter.py
chillenzer: Because there's no C++20 printer in sympy. Not an issue because we're only printing some math functions that are not expected to experience dramatic changes.

chillenzer:
> very cool! Just one minor thing ... The analytic distribution tests [here](https://github.com/ComputationalRadiationPhysics/picongpu/blob/f105f05acbb7f7840ba0f19ff54f8e0cb500ceeb/test/python/picongpu/quick/picmi/distribution.py#L117-L196) seem to have slipped through your net. ;)

Thanks for catching! I'll take care of them.

lib/python/picongpu/picmi/distribution/AnalyticDistribution.py
```diff
@@ -38,36 +38,157 @@
...
+      Either you can define multiple sympy.Piecewise with
+      (0.0, True) as the last condition which means 0 everywhere else.
+      (Make sure it's the last!)
+      Summing those up, works just as you'd expect.
+      Alternatively, you can define only the (expression, condition) tuples
+      and assemble them in a sympy.Piecewise in one go.
+      That's the way chosen in the end-to-end tests.
+
+    Parameters:
+        density_expression (function):
```
pordyna: I think a "function" like object type is usually called `Callable` .

lib/python/picongpu/picmi/distribution/AnalyticDistribution.py
```diff
@@ -38,36 +38,157 @@
...
+            A Python function that takes x, y, z coordinates (in SI units)
+            and returns the density (in SI units) at that point.
+        directed_velocity (3-tuple of float):
+            A collective velocity for the particle distribution.
+            (currently untested)
```
pordyna: Please mention also here that the callable needs to use the `sympy` expressions.

lib/python/picongpu/picmi/distribution/AnalyticDistribution.py
```diff
@@ -38,36 +38,157 @@
...
+                # multiple partial Piecewise instances.
+                # Without simplification, sympy tries to translate them individually.
+                # and fails to do so even if they supplement each other
+                # into a function defined everywhere.
+                sympy.simplify(self.density_expression(x, y, z))
+                # density_expression might be independent of any or all of the three variables.
+                # (This might even happen due to the simplification.)
+                # In order to be sure to arrive at a function of these three variables,
+                # we add this trivial additional term.
+                + (0 * x * y * z)
```
pordyna: why are we doing it again here? What is wrong with `density_expression=self.density_expression?`

lib/python/picongpu/picmi/distribution/GaussianDistribution.py
```diff
@@ -78,7 +79,22 @@ def get_as_pypicongpu(self) -> species.operation.densityprofile.DensityProfile:
         gaussian_profile.gas_sigma_rear = self.sigma_rear
         gaussian_profile.gas_factor = self.factor
         gaussian_profile.gas_power = self.power
-        gaussian_profile.vacuum_cells_front = self.vacuum_cells_front
+        gaussian_profile.vacuum_cells_front = int(self.vacuum_front * grid.get_cell_size()[1])
```
pordyna: you sure this should be `*` and not `/`?

lib/python/picongpu/pypicongpu/rendering/pmaccprinter.py
```diff
@@ -0,0 +1,35 @@
...
+    # Originally, the C++ printers use `_ns = "std::"`.
+    _ns = "pmacc::math::"
+    _kf = cxx_code_printers["c++17"]._kf | PMACC_MATH_FUNCTIONS
+    # The original math_macros contained macros like M_PI from math.h
+    # We want to use the pmacc versions, so we remove this.
+    math_macros = None
+
+    def __init__(self, settings=None):
+        super().__init__(
+            (settings or {})
```
pordyna, line 26: Nice magic, didn't know this returns `{}` and not `False`

test/python/picongpu/end-to-end/free_formula_density.py
```diff
@@ -0,0 +1,254 @@
...
+            picongpu_fixed_charge=True,
+        )
+        sim.add_species(species_hydrogen, random_layout)
+
+
+def setup_sim():
+    grid = picmi.Cartesian3DGrid(
+        number_of_cells=[64, 64, 64],
+        lower_bound=[0, 0, 0],
+        upper_bound=[64.0, 64.0, 64.0],
```
pordyna: I would suggest non-equal number of cells and non-equal cell sizes per dimension for testing, otherwise things my work by accident.

test/python/picongpu/end-to-end/free_formula_density.py
```diff
@@ -0,0 +1,254 @@
...
+                        df.positionOffset_x.to_numpy() + 0.5,
+                        df.positionOffset_y.to_numpy() + 0.5,
+                        df.positionOffset_z.to_numpy() + 0.5,
+                    ),
+                    index=df.set_index(["positionOffset_x", "positionOffset_y", "positionOffset_z"], drop=True).index,
+                ).astype(float),
+                include_groups=False,
+            )
+        )
+        self.assertTrue(np.allclose(densities.found, densities.expected))
```
pordyna: be carefull, this, in contrary to `np.testing.assert_allclose`, has a different default value for absolute tolerance. Here it shouldn't matter, but if the reference values are expected to go below 1, one should maybe use `atol=0.0`

test/python/picongpu/quick/picmi/distribution/analytic_distribution.py
```diff
@@ -0,0 +1,80 @@
...
+from picongpu.picmi import AnalyticDistribution
+import typeguard
+
+VALID_CALLS = [
+    ((1, 2, 3), 6),
+    (([1, 2], 2, 3), [6, 12]),
+    ((1, 2, [3, 4]), [6, 8]),
+    (([1, 2], [3, 4], [5, 6]), [15, 48]),
+]
+
```
pordyna: Tbh, I don't understand what some of them do. Are these valid inputs for x, y, z?

pordyna (commented):
Looks good, just some minor comments. No idea how the `PMaccPrinter` test work, but I believe you :D

lib/python/picongpu/pypicongpu/rendering/pmaccprinter.py
chillenzer: Resolving on behalf of @BrianMarre

lib/python/picongpu/picmi/distribution/GaussianDistribution.py
chillenzer: Good catch! My test case has cell size equal to 1.

lib/python/picongpu/pypicongpu/rendering/pmaccprinter.py
chillenzer: I use this all the time. My favourite debugging for functional style, lambdas and generator expressions and stuff is changing `lambda x: <actual expression to compute>` to `lambda x: breakpoint() or <actual expression to compute>` (and same within generator expressions and the like).

test/python/picongpu/end-to-end/free_formula_density.py
chillenzer: As seen above... of course, that's good practice. I was just lazy. That's what code reviews are for, right? :blush:

test/python/picongpu/quick/picmi/distribution/analytic_distribution.py
chillenzer: Yes, that's basically testing numpy broadcasting rules. I'll make this explicit:
```suggestion
# allow numpy broadcasting (see https://numpy.org/doc/stable/user/basics.broadcasting.html)
# some examples to check:
VALID_CALLS = [
    # scalar arguments produce scalar results
    ((1, 2, 3), 6),
    # broadcasting in the first argument, function is evaluated for (1,2,3) and (2,2,3)
    (([1, 2], 2, 3), [6, 12]),
    # broadcasting in the last argument, (1,2,3) and (1,2,4)
    ((1, 2, [3, 4]), [6, 8]),
    # broadcasting in all arguments, shapes must match, scalar arguments are (1,3,5) and (2,4,6)
    (([1, 2], [3, 4], [5, 6]), [15, 48]),
]
```

test/python/picongpu/end-to-end/free_formula_density.py
chillenzer: Wow! Didn't know that. I was unsure about how `unittest` would handle assertion errors because it uses its own assert methods, so I opted for this version. But I just checked and it handles assertion errors just as expected, so I'll convert to `np.testing.assert_allclose` because it is a better expression of what's going on here.

lib/python/picongpu/picmi/distribution/AnalyticDistribution.py
chillenzer: Thanks for catching. That must have been a rebasing mistake.

lib/python/picongpu/picmi/distribution/AnalyticDistribution.py
chillenzer: Done

lib/python/picongpu/picmi/distribution/AnalyticDistribution.py
chillenzer: Done

lib/python/picongpu/picmi/distribution/AnalyticDistribution.py
chillenzer: Done

lib/python/picongpu/picmi/distribution/GaussianDistribution.py
chillenzer: Done

test/python/picongpu/end-to-end/free_formula_density.py
chillenzer: Done (This was such a pain!)

test/python/picongpu/end-to-end/free_formula_density.py
chillenzer: Done

test/python/picongpu/quick/picmi/distribution/analytic_distribution.py
chillenzer: Done

chillenzer:
Sorry, the tests add two further dependencies which I've put into the picmi `requirements.txt` for now. That definitely shouldn't be the final solution for this. In the long term, our Python ecosystem should follow the standard Python package structure and the each `pyproject.toml` should clearly separate dev-dependencies and user dependencies, etc. #5307 is a first step and could already include such aspects. I'd argue for not blocking this PR due to the dependency handling in this late stage.

pordyna:
@chillenzer Sth I observed when setting up my laser profile with sympy: 

The c++ printer enforces that a Piecewise has always a default value, but sometimes simplify can optimize it out. Do we account for this somehow here? Since we are always simplifying the expression

chillenzer:
I had to surrender: The e2e tests have uncovered a number of problems with the runner already and while the root cause of why it's not running in CI is still unknown. Adding them and picking up all these other little bugs along the way will easily add up to a separate PR. (One which will hopefully be pursued in times of more stable CI.) For now, as with most other tests, runtime testing remains a manual task.

chillenzer:
> @chillenzer Sth I observed when setting up my laser profile with sympy:
> 
> The c++ printer enforces that a Piecewise has always a default value, but sometimes simplify can optimize it out. Do we account for this somehow here? Since we are always simplifying the expression

If this would be the case, it would clearly be a bug in sympy. But my experience is that handling Piecewise objects correctly is a little bit fiddly because any oversight (in terms of undefined regions) quickly spreads. We should investigate this further but we have agreed offline that this is non-blocking.

test/python/picongpu/compiling/distribution.py
```diff
@@ -93,7 +93,7 @@ def test_gaussian(self):
             sigma_rear=5.0e-6,
             power=4.0,
             factor=-1.0,
-            vacuum_cells_front=50,
+            vacuum_front=50 * 9.07264e-5,
```
ikbuibui: Missing a division by  numCellsY

share/picongpu/pypicongpu/template/include/picongpu/param/density.param.mustache
```diff
@@ -176,6 +176,29 @@ namespace picongpu
...
+                /** This formula uses SI quantities only
+                 *
+                 * The profile will be multiplied by BASE_DENSITY.
+                 *
+                 * @param position_SI total offset including all slides [in meter]
+                 * @param cellSize_SI cell sizes [in meter]
+                 *
+                 * @return float_X density [normalized to 1.0]
+                 */
+                HDINLINE float_X operator()(const floatD_64& position_SI, const float3_64& cellSize_SI)
```
ikbuibui, line 192: Is position SI really float_D? you bind it to 3 variables below

lib/python/picongpu/picmi/distribution/FoilDistribution.py
```diff
@@ -40,7 +41,7 @@ class FoilDistribution(picmistandard.PICMI_FoilDistribution):
     def picongpu_get_rms_velocity_si(self) -> typing.Tuple[float, float, float]:
         return tuple(self.rms_velocity)
 
-    def get_as_pypicongpu(self) -> species.operation.densityprofile.DensityProfile:
+    def get_as_pypicongpu(self, grid) -> species.operation.densityprofile.DensityProfile:
```
ikbuibui, line 44: If i am not mistaken, the distribution already holds the upper and lower bounds? If that is true this seems a little risky since the grid input as a parameter might have different bounds to the one set earlier.

test/python/picongpu/quick/pypicongpu/rendering/pmaccprinter.py
```diff
@@ -0,0 +1,103 @@
...
+    ),
+    "log8": (sympy.log(x, 8), "pmacc::math::log(x)/pmacc::math::log(8)"),
+    # for these two there's technically specialised functions in PMAcc
+    # but it's kind of a pain to catch these expressions
+    # before they got through the printer
+    # because sympy treats them as log(x)/log(2), ...
+    # so it's very hard to actually identify such expressions.
+    "log2": (sympy.log(x, 2), "pmacc::math::log(x)/pmacc::math::log(2)"),
+    "log10": (sympy.log(x, 10), "pmacc::math::log(x)/pmacc::math::log(10)"),
+    # PMAcc functionality that's not supported/expressible with sympy:
```
ikbuibui, line 66: Also not expressible are `sincos`, `sinc`, `dot`, `cross`, `norm`s, and bessel functions

lib/python/picongpu/picmi/grid.py
```diff
@@ -83,9 +89,9 @@ def get_as_pypicongpu(self):
             raise ValueError("Z: boundary condition not supported")
 
         g = grid.Grid3D()
-        g.cell_size_x_si = (self.upper_bound[0] - self.lower_bound[0]) / self.number_of_cells[0]
-        g.cell_size_y_si = (self.upper_bound[1] - self.lower_bound[1]) / self.number_of_cells[1]
-        g.cell_size_z_si = (self.upper_bound[2] - self.lower_bound[2]) / self.number_of_cells[2]
+        g.cell_size_x_si = self.get_cell_size()[0]
+        g.cell_size_y_si = self.get_cell_size()[1]
+        g.cell_size_z_si = self.get_cell_size()[2]
```
ikbuibui:
```suggestion
        g.cell_size_x_si, g.cell_size_y_si, g.cell_size_z_si = self.get_cell_size()
```
Not really important but this seems nicer to me.

ikbuibui (commented):
Looks good. Test seems okay. Just a few comments, nothing really major.
As you already noted in your comment, `compare_particles_per_setup` is doing more than twice the number of comparisons. Just for information, my AI recommends using `combinations` from the `itertools` package to solve your problem. 
I'm not a big fan of setting the cell size to one and then keeping `vacuum_front` unchanged. I seems easy to some day change the cell size and forget changing the `vacuum_front`

test/python/picongpu/end-to-end/free_formula_density.py
```diff
@@ -0,0 +1,356 @@
...
+        center_rear,
+        sigma_front,
+        sigma_rear,
+        power,
+        factor,
+    ):
+        # apparently, our SI position is the centre of the cell
+        y += -0.5 * CELL_SIZE[1]
+        # The last term undoes the shift to the cell origin.
+        vacuum_y = int(vacuum_front / CELL_SIZE[1]) * CELL_SIZE[1] - 0.5 * CELL_SIZE[1]
```
chillenzer: This function is using `CELL_SIZE[1]` where it should be using `cell_size_y` as pointed out by @ikbuibui.

test/python/picongpu/end-to-end/free_formula_density.py
```diff
@@ -0,0 +1,356 @@
...
+            "center_rear": 54,
+            "sigma_front": 10,
+            "sigma_rear": 3,
+            "power": 2.0,
+            "factor": -2.0,
+        }
+        self.distributions = {
+            "predefined": picmi.GaussianDistribution(**self.parameters),
+            "free_form": picmi.AnalyticDistribution(
+                lambda x, y, z: self.free_form(y, cell_size_y=1.0, **self.parameters)
```
chillenzer: Relatedly:
```suggestion
                lambda x, y, z: self.free_form(y, cell_size_y=CELL_SIZE[1], **self.parameters)
```

share/picongpu/pypicongpu/template/include/picongpu/param/density.param.mustache
chillenzer: Offline: Our PICMI implementation currently only supports 3D. So, the question is not if but only what kind of error you'll get. I'll leave it as is.

lib/python/picongpu/picmi/distribution/FoilDistribution.py
chillenzer: Offline: This is a not quite satisfying intermediate state on the road to a purely SI-based interface.

ikbuibui (approved):

Issue #5380, opened 2025-05-30
Missing particle filter in openPMDWriter


StevE-Ong:
I am compiling LaserWakeField examples in `picongpu-0.8.0` with `particleFilters.param`. I can compile the setup successfully with the compile log in `stdout`. However, it shows the `highGamma` and `all` filters in openPMD.source, while the `each2nd` is missing:

>  --openPMD.source arg                data sources: [species_all, fields_all, e_all, e_highGamma, E, B, e_all_chargeDensity, e_all_energyDensity, e_all, e_highGamma, E, B, e_all_chargeDensity,e_all_energyDensity] | default: species_all, fields_all

In addition, the printout is duplicated (two times E,B,....)

This does not happen in calorimeter:
> --e_calorimeter.filter arg            particle filter: [all, each2nd, highGamma]
> --e_phaseSpace.filter arg             particle filter: [all, each2nd, highGamma]

I wonder what could go wrong with the setup.

[LaserWakeField_0.8.0.zip](https://github.com/user-attachments/files/20518377/LaserWakeField_0.8.0.zip)

The filter is copied from [here](https://picongpu.readthedocs.io/en/latest/usage/param/particles/init.html#usage-params-core-particles-filters).

psychocoderHPC:
@StevE-Ong Thanks for opening this issue.

First, that filters are added twice to the source property for the OpenPMD plugin is a bug and will be fixed in #5381.

That the filter `each2nd` is not showing up in the source list of the  openPMD plugin is the correrct behavior.
In the documentation for the pplugin https://picongpu.readthedocs.io/en/latest/usage/plugins/openPMD.html#usage-plugins-openpmd it is sying

```
--openPMD.source

Select data sources and filters to dump. 
Default is species_all,fields_all, which dumps all fields and particle species. 
Only deterministic filters will be shown.
```

The filter is a non-deterministic filter, this means if you executed the filter twice it will select different particles.
For openPMD IO it is required that we can executed the particle multiple times because the IO has stages where we first count how many particles we would like to dump and than copy only partilces selected by the filter. This is not working for non-deterministic filters.

The calorimenter plugin is operating in a single path only does not have this restriction and therefore the filter is available there.

PR #5381, opened 2025-06-03
plugin: fix openPMD source list

psychocoderHPC:
All filters are added twice to the openPMD supported source list.

reported in #5380


chillenzer:
Where's the other location where they are added?

psychocoderHPC:
It is a little bit above where we generate the help: https://github.com/ComputationalRadiationPhysics/picongpu/blob/5b15062abd93546982e401a715f49b0f5caad527/include/picongpu/plugins/openPMD/openPMDWriter.x.cpp#L333-L338

chillenzer (approved):

PR #5384, opened 2025-06-04
Documentation fixes for perlmutter-nersc

AnonNick:
Modified perlmutter-nersc documentation to address the following:

- Changes required in the `gpu.profile` (project number and PICONGPU repository location)
- First time setup not including sourcing `gpu.profile` as `dependencies_autoinstall.sh` handles it
- Informing the user that sourcing `gpu.profile` is required for subsequent use 
- Adding code snippets to make it easier for the user

### DIFF
--- a/etc/picongpu/perlmutter-nersc/README.md
+++ b/etc/picongpu/perlmutter-nersc/README.md
@@ -27,7 +27,7 @@ First-Time Setup:
 
 Subsequent Use:
 
--  Simply source the `gpu.profile` each time before using PICONGPU:
+-  Simply source the `gpu.profile` each time before using PIConGPU:
    ```bash
    source gpu.profile
    ```
--- a/etc/picongpu/perlmutter-nersc/README.md
+++ b/etc/picongpu/perlmutter-nersc/README.md
@@ -19,7 +19,7 @@ First-Time Setup:
    ```bash
    export PIC_PROFILE=/path/to/gpu.profile
    ```
-2. Install missing dependecies from source with `dependencies_autoinstall.sh` 
+2. Install missing dependencies from source with `dependencies_autoinstall.sh` 
    ```bash
    source dependencies_autoinstall.sh
    ```
--- a/etc/picongpu/perlmutter-nersc/README.md
+++ b/etc/picongpu/perlmutter-nersc/README.md
@@ -3,7 +3,35 @@
 PIConGPU can be compiled on a login node, but remember to limit the number of used CPU cores. `pic-build -j 16` works fine; reduce the number if the compiler gets killed.
 
 ## Installing Dependencies
-Missing dependencies can be installed from source with `dependencies_autoinstall.sh`. Change the project number in the `gpu.profile`, source it, and execute the install script on a login node.
+
+Before you begin, update the following in the `gpu.profile` file:
+- Change the project number (`export proj` variable)
+  ```bash
+  export proj="m0000"
+  ```
+- Change path to PICONGPU repository (`PICSRC` variable)
+  ```bash
+  export PICSRC=$HOME/src/picongpu
+  ```
+
+First-Time Setup:
+1. Set the `PIC_PROFILE` environment variable to the path of your `gpu.profile` file:  
+   ```bash
+   export PIC_PROFILE=/path/to/gpu.profile
+   ```
+2. Install missing dependecies from source with `dependencies_autoinstall.sh` 
+   ```bash
+   source dependencies_autoinstall.sh
+   ```
+
+
+Subsequent Use:
+
+-  Simply source the `gpu.profile` each time before using PICONGPU:
+   ```bash
+   source gpu.profile
+   ```
+
 
 ## Preemptive Jobs
 When using preemptive queues, you should uncomment the `sleep 120` command at the end of the tpl file.
--- a/etc/picongpu/perlmutter-nersc/README.md
+++ b/etc/picongpu/perlmutter-nersc/README.md
@@ -9,7 +9,7 @@ Before you begin, update the following in the `gpu.profile` file:
   ```bash
   export proj="m0000"
   ```
-- Change path to PICONGPU repository (`PICSRC` variable)
+- Change path to PIConGPU repository (`PICSRC` variable)
   ```bash
   export PICSRC=$HOME/src/picongpu
   ```
--- a/etc/picongpu/perlmutter-nersc/README.md
+++ b/etc/picongpu/perlmutter-nersc/README.md
@@ -19,7 +19,7 @@ First-Time Setup:
    ```bash
    export PIC_PROFILE=/path/to/gpu.profile
    ```
-2. Install missing dependencies from source with `dependencies_autoinstall.sh` 
+2. Install missing dependencies from source with `dependencies_autoinstall.sh`, this can be done on a login node: 
    ```bash
    source dependencies_autoinstall.sh
    ```


PrometheusPi (approved):

etc/picongpu/perlmutter-nersc/README.md
```diff
@@ -3,7 +3,35 @@
...
+   ```
+2. Install missing dependecies from source with `dependencies_autoinstall.sh` 
+   ```bash
+   source dependencies_autoinstall.sh
+   ```
+
+
+Subsequent Use:
+
+-  Simply source the `gpu.profile` each time before using PICONGPU:
```
PrometheusPi: minor typo 
```suggestion
-  Simply source the `gpu.profile` each time before using PIConGPU:
```

etc/picongpu/perlmutter-nersc/README.md
```diff
@@ -3,7 +3,35 @@
...
+  ```bash
+  export PICSRC=$HOME/src/picongpu
+  ```
+
+First-Time Setup:
+1. Set the `PIC_PROFILE` environment variable to the path of your `gpu.profile` file:  
+   ```bash
+   export PIC_PROFILE=/path/to/gpu.profile
+   ```
+2. Install missing dependecies from source with `dependencies_autoinstall.sh`
```
PrometheusPi: typo:
```suggestion
2. Install missing dependencies from source with `dependencies_autoinstall.sh` 
```

etc/picongpu/perlmutter-nersc/README.md
```diff
@@ -3,7 +3,35 @@
...
 
 ## Installing Dependencies
-Missing dependencies can be installed from source with `dependencies_autoinstall.sh`. Change the project number in the `gpu.profile`, source it, and execute the install script on a login node.
+
+Before you begin, update the following in the `gpu.profile` file:
+- Change the project number (`export proj` variable)
+  ```bash
+  export proj="m0000"
+  ```
+- Change path to PICONGPU repository (`PICSRC` variable)
```
PrometheusPi: minor typo:
```suggestion
- Change path to PIConGPU repository (`PICSRC` variable)
```

PrometheusPi (changes requested):
Thanks for providing this detailed description. 
I just found some minor typos.

etc/picongpu/perlmutter-nersc/README.md
```diff
@@ -3,7 +3,35 @@
...
+- Change path to PIConGPU repository (`PICSRC` variable)
+  ```bash
+  export PICSRC=$HOME/src/picongpu
+  ```
+
+First-Time Setup:
+1. Set the `PIC_PROFILE` environment variable to the path of your `gpu.profile` file:  
+   ```bash
+   export PIC_PROFILE=/path/to/gpu.profile
+   ```
```
pordyna, line 21: `# Name and Path of this Script ############################### (DO NOT change!)`
This should not be edited, the expression in the profile sets this correctly to the sourced file path. Never had a problem with this.

etc/picongpu/perlmutter-nersc/README.md
```diff
@@ -3,7 +3,35 @@
...
+2. Install missing dependencies from source with `dependencies_autoinstall.sh` 
+   ```bash
+   source dependencies_autoinstall.sh
+   ```
```
pordyna: Please add back the information that this can be done on a login node.

pordyna (changes requested):
Hi! Thanks a lot for the contribution! I just have two comments. The PIC_PROFILE thing is kind of important, idk why you needed to do this. But if this is somehow not working for you, we can figure it out separately.

etc/picongpu/perlmutter-nersc/README.md
PrometheusPi: @AnonNick I had a brief discussion with @pordyna: he has a good point here:
 The variable `PIC_PROFILE` is usually set by the file itself, thus you can store it anywhere and the variable will always have the correct path. Thus, please leave it as is. 
 To avoid the double sourcing you mentioned yesterday, I would suggest to running 
 ```bash
 export PIC_PROFILE=/path/to/gpu.profile
 ```
 on the terminal / in the shell before executing the installation script. Could you add this instead to the `README.md`?

AnonNick: The reason I'd added the setting of PIC_PROFILE was because `dependencies_autoinstall.sh` has the following lines:
```bash
# get PIConGPU profile
if [ ! -f "$PIC_PROFILE" ]; then
    printf "Source a profile!\n"
    exit 1
else
    source "$PIC_PROFILE"
fi
```

When I followed the old steps required for the initial setup:
1. I sourced `gpu.profile`, which set PIC_PROFILE
2. I then ran `dependencies_autoinstall.sh` which sourced the `gpu.profile` again

The steps were causing `gpu.profile` to be sourced twice, that's why as an initial step I'd asked the user to set PIC_PROFILE so as to source the `gpu.profile` only once.

If I follow the steps of setting PIC_PROFILE and then running `dependencies_autoinstall.sh`, `gpu.profile` is sourced within `dependencies_autoinstall.sh` and PIC_PROFILE is updated. User will not need to change/update PIC_PROFILE after that point.

Let me know If it doesn't suite the requirements of PIConGPU, I'll remove the PIC_PROFILE step and replace it with the following:
First-Time Setup:
1. Source the `gpu.profile`.
   ```bash
   source gpu.profile
   ```
2. Install missing dependecies from source with `dependencies_autoinstall.sh`. (This can be done on a login node.) 
   ```bash
   source dependencies_autoinstall.sh
   ```

AnonNick: @PrometheusPi I seemed to have missed your message in the thread.

In my update I've listed the following for initial setup:
First-Time Setup:
1. Set the `PIC_PROFILE` environment variable to the path of your `gpu.profile` file:  
   ```bash
   export PIC_PROFILE=/path/to/gpu.profile
   ```
2. Install missing dependencies from source with `dependencies_autoinstall.sh`, this can be done on a login node: 
   ```bash
   source dependencies_autoinstall.sh
   ```
   
Should I add a note saying that PIC_PROFILE is being used by `dependencies_autoinstall.sh` to make it clearer to the users that they only need to do it when they want to run `dependencies_autoinstall.sh`?

PrometheusPi: @AnonNick Yes:
> If I follow the steps of setting PIC_PROFILE and then running dependencies_autoinstall.sh, gpu.profile is sourced within dependencies_autoinstall.sh and PIC_PROFILE is updated. User will not need to change/update PIC_PROFILE after that point.

sounds good. Do you agree @pordyna?
Your suggestion looks good.

pordyna: Sure, let's do this like this

chillenzer: Just for my information because I haven't been on that system yet: Is there a significant cost in sourcing the profile more than once?

PrometheusPi: @chillenzer Not really if compared to the build script. It's just confusing if you get all the verbose output of loading, unloading and loading again.

PrometheusPi (approved):

etc/picongpu/perlmutter-nersc/README.md
chillenzer: Thanks!

PR #5385, opened 2025-06-04
mark field ionization as implemented in FLYonPIC

BrianMarre:
moves field ionization in FLYonPIC documentation from planned to implemented

EDIT: naming of processes is based on https://www.nist.gov/pml/quantum-measurement/atomic-spectroscopy/electron-beam-ion-trap-ebit/electron-ion-collisions

### DIFF
--- a/docs/source/models/atomic_physics.rst
+++ b/docs/source/models/atomic_physics.rst
@@ -22,17 +22,17 @@ This allows a more accurate prediction of ionization dynamics and the prediction
 
 FLYonPIC is based on the [FlyCHK]_ atomic model and currently implements the following processes:
 
-- electronic collisional de-/excitation
+- electron-ion collisional de-/excitation
 - spontaneous deexcitation
-- electronic ionization (with ionization potential depression)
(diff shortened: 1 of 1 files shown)


docs/source/models/atomic_physics.rst
```diff
@@ -24,15 +24,15 @@ FLYonPIC is based on the [FlyCHK]_ atomic model and currently implements the fol
 
 - electronic collisional de-/excitation
 - spontaneous deexcitation
-- electronic ionization (with ionization potential depression)
+- electronic ionization (with Stewart-Pyatt ionization potential depression)
```
PrometheusPi: This seems like a typo (same as two lines above) 
Please provide a reference here to defend your naming scheme

PrometheusPi (changes requested):

PrometheusPi (approved):

docs/source/models/atomic_physics.rst
BrianMarre: You are right, naming is probably confusing for users

PR #5392, opened 2025-06-18
Fix integer casting in TimeStepSpec

chillenzer:
Bug found by @pordyna. The previous code passed my tests because I used easy to read numbers which didn't trigger the case that @pordyna found. I added a quick regression test which should catch the most outrageous of my mistakes. I'll update the tests with numbers better probing our typical regime in the mid-term.

### DIFF
--- a/lib/python/picongpu/picmi/diagnostics/timestepspec.py
+++ b/lib/python/picongpu/picmi/diagnostics/timestepspec.py
@@ -124,7 +124,7 @@ class TimeStepSpec(metaclass=_TimeStepSpecMeta):
             # In PIConGPU, a single integer would be interpreted as
             # slice(None, None, value) but this is unnatural for the
             # Python [] operator.
-            spec if isinstance(spec, slice) else slice(int(spec), int(spec), 1)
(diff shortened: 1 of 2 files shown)


lib/python/picongpu/picmi/diagnostics/timestepspec.py
```diff
@@ -124,7 +124,7 @@ def __init__(self, *args, specs_in_seconds=tuple()):
             # In PIConGPU, a single integer would be interpreted as
             # slice(None, None, value) but this is unnatural for the
             # Python [] operator.
-            spec if isinstance(spec, slice) else slice(int(spec), int(spec), 1)
+            spec if isinstance(spec, slice) else slice(spec, spec, None)
```
ikbuibui, line 127: Is it correct that `slice(x, y, 1)` equivalent to `slice(x, y, None)`

chillenzer: For integers (unit="steps"): Yes. For floats (unit="seconds"): No, that was (one of) the bug(s).

ikbuibui (approved):

PR #5389, opened 2025-06-17
Fix isaac

chillenzer:
There's not really something wrong on the PIConGPU side of things. Only a missing header. Once the corresponding ISAAC PR (https://github.com/ComputationalRadiationPhysics/isaac/pull/184) is through, ISAAC should be usable. Important learning from my experiments: I needs sufficient GPU memory, so increasing `reservedGpuMemorySize` in `memory.param` is necessary. On our dev-Server, this led to a clear out-of-memory error message but in larger runs by @BeyondEspresso the simulation seems to have run but no data was transferred.

PR #5395, opened 2025-06-23
fix  broken test case

psychocoderHPC:
Since the reactivation of the ISAAC plugin #5389 the `compileLaser` test case is not compiling because of a missing include.

PR is tested with `ci: full-compile`

### DIFF
--- a/include/picongpu/plugins/IsaacPlugin.x.cpp
+++ b/include/picongpu/plugins/IsaacPlugin.x.cpp
@@ -21,6 +21,8 @@
 // required for SIMDIM definition
 #include "picongpu/defines.hpp"
 
+#include <concepts>
+
 #if (ENABLE_ISAAC == 1) && (SIMDIM == DIM3)
 
 #    include "picongpu/param/isaac.param"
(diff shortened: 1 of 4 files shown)


steindev (dismissed):

include/picongpu/plugins/IsaacPlugin.x.cpp
```diff
@@ -862,17 +862,20 @@ namespace alpaka
 {
 
     template<>
+    requires(!mp_empty<typename picongpu::isaacP::IsaacPlugin::SourceList>::value)
     struct IsKernelArgumentTriviallyCopyable<typename picongpu::isaacP::IsaacPlugin::SourceList> : std::true_type
```
ikbuibui: I think the problem is that template specializations cant have requires clauses

psychocoderHPC: I pushed a new version. IMO this should fix it.

PrometheusPi (dismissed):

chillenzer (approved):

PR #5402, opened 2025-06-30
Field binning fixes to pass fields and setters to configure binning

ikbuibui:
- Added a type called ``FieldInfo`` which is how users are now supposed to pass fields to the binning plugin. 
This was necessary because users cant directly access the field data box from the binning setup. This is because binning setup is called on plugin load, and this happens before the fields are registered with the data connector.
The ``FieldInfo`` object is only passed to the binning functors AFTER being transformed to the corresponding field databox.
Similarly, for passing field information when doing particle binning, passing a ``FieldInfo`` object as extra data will provide the user with access to the field databox.
- Fixed setters of field binning data configuration not working (for example dump period)
- Made the integer linear axis binning more robust against overflows by using 64 bit types internally (thanks @pordyna for testing an axis with almost a million bins)
- Add a domain origin `LOCAL_WITH_GUARDS` to be used with `DomainInfo`, mainly with the view of making local field boxes correctly usable more easily. Added an example to the documentation which shows how to use field binning correctly
- Fix `getParticlePosition` missing a cast, fixes cases where integral cell sizes are used

### DIFF
--- a/include/picongpu/plugins/binning/utility.hpp
+++ b/include/picongpu/plugins/binning/utility.hpp
@@ -107,22 +107,5 @@ namespace picongpu
             return std::unique_ptr<std::remove_cvref_t<decltype(*ptr)>>(ptr);
         }
 
-        template<typename, template<typename...> typename>
-        struct is_specialization_of : std::false_type
-        {
-        };
-
-        template<template<typename...> typename Template, typename... Args>
-        struct is_specialization_of<Template<Args...>, Template> : std::true_type
-        {
-        };
-
-        // TODO reverse this order
-        template<typename T, template<typename...> typename Template>
-        inline constexpr bool is_specialization_of_v = is_specialization_of<T, Template>::value;
-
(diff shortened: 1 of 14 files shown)


include/picongpu/plugins/binning/utility.hpp
```diff
@@ -107,6 +107,22 @@ namespace picongpu
             return std::unique_ptr<std::remove_cvref_t<decltype(*ptr)>>(ptr);
         }
 
+        template<typename, template<typename...> typename>
+        struct is_specialization_of : std::false_type
```
psychocoderHPC: Discussed offline that this will go into pmacc and should follow our naming rules.

psychocoderHPC (dismissed):

include/picongpu/plugins/binning/utility.hpp
ikbuibui: Fixed

psychocoderHPC (approved):

PR #5403, opened 2025-07-01
fix folder creation

### LINKED ISSUE #5401: png plugin - error during file/directory creation if run on multiple nodes
The png plugin encounters a write error when run on multiple nodes when creating the directory for the png files. 
The png plugin outputs the following error to `stderr`:
```
PNGwriter::close - ERROR **: Error creating file (fopen() returned NULL pointer).
PNGwriter::close - ERROR **: No such file or directory
```
but PIConGPU ignores this **silently**.

This issue is not limited to one cluster. It could be reproduced on 
 - hemera/HZDR
 - capella/ZIH/TU Dresden

While on hemera, no png directory was created at all (but other directories were), only one of two png directories was created on capella. 

To test this, one can use the default LWFA example and the `32.cfg`.

psychocoderHPC:
fix #5401

The method `createDirectory()` and `createDirectoryWithPermissions()` was not clearly mention that only one MPI rank is allowed to call this method. The original implementation guarded the permission setting method by using the global rank zero only but in cases where the IO rank was not rank zero the permission was not set and since the refactoring #5094 also the output folder was not created.

EDIT/ADDITION by @PrometheusPi:
Originally, the directory creation was done by all MPI ranks and access-right changes were done by a selected group of ranks. The right-changes were already broken, and not always were access-rights set correctly. The change #5094 moved the directory-creation into the same condition as the (broken) right-changes. Thus, directories were not always created. This pull request fixed **both** directory-creation and access-right-changes.

### DIFF
--- a/include/picongpu/plugins/PhaseSpace/PhaseSpace.x.cpp
+++ b/include/picongpu/plugins/PhaseSpace/PhaseSpace.x.cpp
@@ -318,9 +318,6 @@ namespace picongpu
 
             if(activatePlugin)
             {
-                /** create dir */
-                pmacc::Filesystem::get().createDirectoryWithPermissions("phaseSpace");
-
                 uint32_t const r_element = axis_element.space;
 
                 /* CORE + BORDER elements for spatial bins */
@@ -381,6 +378,12 @@ namespace picongpu
                     std::vector<int> ranks(
                         std::lower_bound(planeReduceRootRanks.begin(), planeReduceRootRanks.end(), 0),
                         planeReduceRootRanks.end());
+
+                    if(static_cast<int>(gc.getGlobalRank()) == ranks.front())
+                    {
+                        /** only one rank is creating the output directory */
+                        pmacc::Filesystem::get().createDirectoryWithPermissions("phaseSpace");
+                    }
 
                     MPI_CHECK(MPI_Comm_group(MPI_COMM_WORLD, &world_group));
                     MPI_CHECK(MPI_Group_incl(world_group, ranks.size(), ranks.data(), &new_group));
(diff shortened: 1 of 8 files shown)


PrometheusPi:
@psychocoderHPC The CI fails currently.

ikbuibui (approved):

PrometheusPi:
I tested your PR and it works @psychocoderHPC. 
But why is your fix significantly more complex than my [minimal fix](https://github.com/ComputationalRadiationPhysics/picongpu/issues/5401#issuecomment-3019541842)?

PrometheusPi (approved):
Tested for png plugin only.

psychocoderHPC:
> I tested your PR and it works @psychocoderHPC. But why is your fix significantly more complex than my [minimal fix](https://github.com/ComputationalRadiationPhysics/picongpu/issues/5401#issuecomment-3019541842)?

offline discussed: There was an additional bug in since very long time that the rights are not set in cases the MPI rank zero does not called this method. This PR is fixing both because they fall into the same categorie.

chillenzer (dismissed):
Two things about this:
1. `createImage` gets a docstring that only one MPI rank is allowed to call this method but there's no functional changes related to the use of this method. Wasn't that the original issue? I would have expected that calls to that method would become guarded in a similar manner. (Just checking!)
2. I'm not quite at ease with your design. The next user of this method might very well overlook that warning again and create the next race condition. Also, if `grep <method name> -r` is the best way to check if a single change was consistent, that's not a sustainable design. How about an interface like this?
```C++
void createDirectoryFromRank(std::filesystem::path dir, int rank=0);
```
wherein the method itself handles MPI stuff. Would cover all the changes you've made and nobody can get around thinking about the MPI rank that's performing the operation. Still allows an arbitrary subset of ranks to create directories (simply specifying a rank that depends on my own MPI rank).
 
Easy to use, hard to misuse!

chillenzer:
Okay.

PR #5407, opened 2025-07-07
Add profile for Meluxina at Luxprovide

steindev:
Tested and runs.


ikbuibui (approved):

chillenzer (approved):
Can't really say much about this. Do we have a second person in the team to confirm that it works?

ikbuibui:
I have access as well and in principle I can test to confirm, but I don't think this is required. Klaus already tested this, and that plus a second (or third) pair of eyes for a sanity check of the actual profile, is a sufficient review for a profile IMO.

PR #5409, opened 2025-07-14
Add particle binning to picmi

chillenzer:
## Particle Binning for PICMI

This PR adds initial support for particle binning to PICMI. This support is not yet complete, see list below for what's missing. The interface is to be considered experimental and subject to change (because it is yet unclear if the interface will have to distinguish between particle and field binning explicitly). Because of that there's no documentation at the moment. Power-users like @pordyna will figure it out and others can look at the description in this PR. Experienced users should check the generated C++ code if things don't do what's expected.

### Status of this PR and your mission

Please review, test and start using this PR, so we can get decent feedback about this. ~~It's not completely polished yet as it would benefit from some rebasing and working on this has uncovered some bugs that probably have to be fixed in C++ world. I'll leave this as green, so people will actually care to look at it. But I expect no review to pass in every detail yet.~~

### UI Description

The UI can be seen in the end-to-end test. Users can add an arbitrary number of `picmi.diagnostics.binning.Binning` objects to `sim.diagnostics` each of which contains four necessary pieces of information:
```python
            Binning(
                name="particle_density",
                deposition_functor=BinningFunctor(
                    name="particle_density",
                    functor=lambda particle: particle.get("weighting"),
                    # Can be arithmetic python types or a string which would be inserted literally:
                    return_type=float,
                ),
                axes=[BinningAxis(...), ...],
                # could be a list:
                species=species_hydrogen,
                # and more with defaults...
                period=TimeStepSpec[:],
                # ...most of the other things that the `Binner` has as setters...
            )
```
This will produce the desired binning output in the folder `${run_dir}/simOutput/binningOpenPMD/${Binning.name}_${infix}.${extension}`. Some details:
- The `BinningFunctor` takes a function with the signature `def functor(particle)`. The particle passed to the function exposes a `.get` method that can take any particle attribute name as string (anything that we'd get via `particle[attribute_]` in C++) as well as `position`, `timestep` and the attributes implemented as traits like `picongpu::traits::attribute::getMass`. For `position` it can take additional keyword arguments `origin`, `precision` and `unit` (but see below because that's not yet fully functional). The return value of `particle.get` is either a `sympy.Symbol` or a tuple of `sympy.Symbol` (depending on whether a scalar or a vector was asked for). We might find a more appropriate collection later on. The obvious candidates `sympy.Vector` and/or `sympy.Point` unfortunately have their own quirks and bugs and also a relatively bulky interface, so I'm not yet sure about what's the best interface here.
- The `BinningAxis` class bundles a `BinningFunctor` and a `BinSpec(kind="linear"/"log", start, stop, nsteps)`.
- The further configuration options include all the simple things like `timeaveraging`, openPMD configuration (as a dictionary that's converted to a json string for code generation) and so on. It does not contain the tricky stuff like host-side hooks and openPMD functors. These will likely never get exposed but will only be used internally as means to provide higher level features in the frontend.

### Implementation details

The translation of functors from the UI to the generated code is roughly organised as follows:
- On the PICMI level, the interface is provided. Upon translation into the PyPIConGPU layer via `.get_as_pypicongpu`, the `BinningFunctor` calls the given functor with an opaque `BinningParticle` object. While it returns the corresponding symbols from the `.get` method, it keeps track of the requested properties and handed out symbols. The sympy expression as well as a dictionary mapping symbols to attributes is passed to PyPIConGPU.
- On the PyPIConGPU level, the `functor_expression` is handled as in the free-formula density case via the `PMAccPrinter`. This produces an expression containing the variables as they were handed out as `sympy.Symbol`s above. In order to declare and define these variables, PyPIConGPU's `BinningFunctor` generates a custom preamble to preceed the generated `functor_expression` in the generated code. It roughly looks like this (wherein, of course, only statements for actually used variables are generated)
```C++
auto const weighting = particle[weighting_];
auto const [px, py, pz] = particle[momentum_];
// but also things like...
auto const [x_cell_cell,y_cell_cell,z_cell_cell] = getParticlePosition<DomainOrigin::TOTAL, PositionPrecision::Cell, PositionUnits::Cell>(domainInfo, particle);
```
- The template organises each `Binner` into its own function and calls all those functions in `getBinning` at the end.

The other data are straightforwardly passed through the layers and not much magic is happening. I have also restrained from putting in strong guardrails at the moment because a number of aspects are planned as (or at least conceivable as) extension points or even explicitly allow for arbitrary strings to be passed through.

### Testing
Tests for this feature are built on the infrastructure set up for free-formula densities and are expected to suffer from the same problems in CI, so there's no point in activating them at the moment. This makes it even more important that the reviewers please run the tests! I will come up with a better separation of concerns in the process of bringing the tests into CI.

The main test again computes the particle density and compares to the other methods that already existed. I'm planning to add some further tests for different functionalities but success in this direction was limited by the backend bugs described below. As submitted, the tests currently do not succeed but fail at compile time. Deactivating the (as of now unused-in-the-tests) position binnings makes them pass.

### Missing features
In general, this PR does not yet attempt to add field binning. Furthermore, not all details from the particle binning are implemented as of now. This is the list of missing feature I'm aware of. Please ping me and add to this list if there's anything I've overlooked.
- Add units to binning.
- Species filtering.
~~- Add supercell as precision and potentially local cell index as origin, etc.~~
- Accumulate operation is not configurable.

~~Backend Bugs~~

~~Flexible binning is a relatively young feature by itself, so there's still some bugs in the backend that have been triggered by the systematic testing for this PR:~~

~~- Some initial tests ran into segfaults with more than 5 axes. @ikbuibui is supposed to investigate this further.~~
~~- The position functors which are currently added but not yet tested lead to compile-time errors about inconvertible return types. @ikbuibui is supposed to investigate this further.~~

### DIFF
--- a/lib/python/picongpu/picmi/diagnostics/binning.py
+++ b/lib/python/picongpu/picmi/diagnostics/binning.py
@@ -1,0 +1,191 @@
+"""
+This file is part of PIConGPU.
+Copyright 2025 PIConGPU contributors
+Authors: Julian Lenz
+License: GPLv3+
+"""
+
+import sympy
+from ...pypicongpu.output.binning import (
+    Binning as PyPIConGPUBinning,
+    BinningFunctor as PyPIConGPUBinningFunctor,
+    BinningAxis as PyPIConGPUBinningAxis,
+    BinSpec as PyPIConGPUBinSpec,
+)
+from ...pypicongpu.species.species import Species as PyPIConGPUSpecies
+
+from ..species import Species as PICMISpecies
+from typing import Callable, Any, Optional
+from .timestepspec import TimeStepSpec
+import typeguard
+
+_COORDINATE_SYSTEM = {
+    (
+        origin.lower(),
+        precision.lower(),
+        unit.lower(),
+    ): tuple(sympy.Symbol(f"{c}_{precision.lower()}_{unit.lower()}") for c in coords)
+    for (origin, coords) in (
+        ("TOTAL", ("xt", "yt", "zt")),
+        ("GLOBAL", ("xg", "yg", "zg")),
+        ("LOCAL", ("xl", "yl", "zl")),
+        ("MOVING_WINDOW", ("xmw", "ymw", "zmw")),
+        ("LOCAL_WITH_GUARDS", ("xlg", "ylg", "zlg")),
+    )
+    for precision in ("CELL", "SUB_CELL")
+    for unit in ("CELL", "PIC", "SI")
+}
+
+
+class BinningParticle:
+    def __init__(self):
+        self.used_attributes = {}
+
+    def get_attribute_map(self):
+        return self.used_attributes
+
+    def get(self, attribute, **kwargs):
+        if attribute == "position":
+            origin = kwargs.get("origin", "total")
+            precision = kwargs.get("precision", "cell")
+            unit = kwargs.get("unit", "cell")
+            symbols = _COORDINATE_SYSTEM[(origin, precision, unit)]
+            self.used_attributes |= {symbols: ("position", origin, precision, unit)}
+
+        elif attribute == "momentum":
+            symbols = sympy.symbols("px,py,pz")
+            self.used_attributes |= {symbols: "momentum"}
+
+        elif attribute in ["gamma", "kinetic energy", "velocity"]:
+            # This relies on python dictionaries having a stable ordering.
+            # We first add mass and momentum
+            # and later use their symbols inside of the same preamble.
+            self.get("mass")
+            self.get("momentum")
+            if attribute == "gamma":
+                symbols = sympy.Symbol("gamma")
+            if attribute == "kinetic energy":
+                symbols = sympy.Symbol("Ekin")
+            if attribute == "velocity":
+                symbols = sympy.symbols("vx,vy,vz")
+            self.used_attributes |= {symbols: attribute}
+
+        else:
+            symbols = sympy.Symbol(attribute)
+            self.used_attributes |= {symbols: attribute}
+
+        return symbols
+
+
+@typeguard.typechecked
+class BinningFunctor:
+    def check(self):
+        pass
+
+    def __init__(
+        self,
+        name: str,
+        functor: Callable[[BinningParticle], Any],
+        return_type: type | str,
+    ):
+        self.name = name
+        self.functor = functor
+        self.return_type = return_type
+
+    def get_as_pypicongpu(self) -> PyPIConGPUBinningFunctor:
+        self.check()
+        particle = BinningParticle()
+        functor_expression = self.functor(particle)
+        return PyPIConGPUBinningFunctor(
+            name=self.name,
+            functor_expression=functor_expression,
+            attribute_mapping=particle.get_attribute_map(),
+            return_type=self.return_type,
+        )
+
+
+@typeguard.typechecked
+class BinSpec:
+    def __init__(self, kind, start, stop, nsteps):
+        self.kind = kind
+        self.start = start
+        self.stop = stop
+        self.nsteps = nsteps
+
+    def get_as_pypicongpu(self):
+        return PyPIConGPUBinSpec(self.kind.lower().capitalize(), self.start, self.stop, self.nsteps)
+
+
+@typeguard.typechecked
+class BinningAxis:
+    def __init__(
+        self,
+        functor: BinningFunctor,
+        bin_spec: BinSpec,
+        name: str | None = None,
+        use_overflow_bins: bool = True,
+    ):
+        self.functor = functor
+        self.bin_spec = bin_spec
+        self.name = name or functor.name
+        self.use_overflow_bins = use_overflow_bins
+
+    def get_as_pypicongpu(self) -> PyPIConGPUBinningAxis:
+        return PyPIConGPUBinningAxis(
+            name=self.name,
+            functor=self.functor.get_as_pypicongpu(),
+            bin_spec=self.bin_spec.get_as_pypicongpu(),
+            use_overflow_bins=self.use_overflow_bins,
+        )
+
+
+@typeguard.typechecked
+class Binning:
+    def __init__(
+        self,
+        name: str,
+        deposition_functor: BinningFunctor,
+        axes: list[BinningAxis],
+        species: PICMISpecies | list[PICMISpecies],
+        period: Optional[TimeStepSpec] = None,
+        openPMD: Optional[dict] = None,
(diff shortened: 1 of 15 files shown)


ikbuibui:
I looked into the issues you faced. 
1. I tried your setup and indeed ran into a segfault if I enable that axis. This happens not because of a limit on supported axes but because we try to allocate a 64 GiB buffer on host and device and we run out of memory. I tried decreasing the size of the binning axes and was able to test a binning of 8 axes. 
2. The errors regarding inconvertible return types occur because the "backend" currently doesn't support integral cell sizes. I have added the fix for this to #5402. At the heart of this lies a deeper issue regarding lack of conversions in arithmetic operators of `pmacc::Vector`, which we dont do to avoid silent-ish upcasting of types inadvertently by the user.

ikbuibui:
It would make sense to add access to ``picongpu/algorithms/``, ``picongpu/traits/attribute`` and some of the symbols available from ``simulation.unitless`` like `getEps0` and `getSpeedOfLight` in the binning functors as these might be useful for the physicists

chillenzer:
This PR is ready for review. Notable changes since last push:
- Included a fix to `ValidPositionRequest` that adds `LOCAL_WITH_GUARDS`.
- Completion of origin, precision and unit support for domain info requests.
- Added `.get` access to pertinent functionality from `picongpu/algorithms`
- Refactored the end-to-end tests for better readability and structure. (They are still not included in CI and still need a general overhaul but it's a step forward I hope.)
- Cleaned up the history into a digestible number of commits.
- Potentially, there are futher minor things.

lib/python/picongpu/picmi/diagnostics/binning.py
```diff
@@ -0,0 +1,191 @@
...
+
+        elif attribute in ["gamma", "kinetic energy", "velocity"]:
+            # This relies on python dictionaries having a stable ordering.
+            # We first add mass and momentum
+            # and later use their symbols inside of the same preamble.
+            self.get("mass")
+            self.get("momentum")
+            if attribute == "gamma":
+                symbols = sympy.Symbol("gamma")
+            if attribute == "kinetic energy":
```
ikbuibui, line 67:
```suggestion
            elif attribute == "kinetic energy":
```
Minor thing, but logically `elif` seems more appropriate. Same thing also happens below with velocity.

lib/python/picongpu/pypicongpu/output/binning.py
```diff
@@ -0,0 +1,177 @@
...
+        precision.lower(),
+        unit.lower(),
+    ): f"getParticlePosition<DomainOrigin::{origin}, PositionPrecision::{precision}, PositionUnits::{unit}>(domainInfo, particle)"
+    for origin in ("TOTAL", "GLOBAL", "LOCAL", "MOVING_WINDOW", "LOCAL_WITH_GUARDS")
+    for precision in ("CELL", "SUB_CELL")
+    for unit in ("CELL", "PIC", "SI")
+} | {
+    "mass": "picongpu::traits::attribute::getMass(particle[weighting_], particle)",
+    # CAUTION: The names in the gamma formula are currently hardcoded.
+    # We'll certainly trip over this, should we ever dare to change the internal names.
```
ikbuibui, line 39: 🙃

ikbuibui (commented):
Aside from the fact that I need to create a directory called `scratch` in my home to hold the `end-to-end` test output, things look okay.  Tests pass on my machine.
The refactoring of the distributions and the related test is sneaked in but its fine.

ikbuibui (approved):

lib/python/picongpu/picmi/diagnostics/binning.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+@typeguard.typechecked
+class Binning:
+    def __init__(
+        self,
+        name: str,
+        deposition_functor: BinningFunctor,
+        axes: list[BinningAxis],
+        species: PICMISpecies | list[PICMISpecies],
```
BrianMarre, line 149: I would simplify the interface here and always require a list of species

lib/python/picongpu/picmi/diagnostics/binning.py
```diff
@@ -0,0 +1,191 @@
...
+        period: Optional[TimeStepSpec] = None,
+        openPMD: Optional[dict] = None,
+        openPMDExt: Optional[str] = None,
+        openPMDInfix: Optional[str] = None,
+        timeAveraging: bool = True,
```
BrianMarre, line 154: do we really want to default to time averaging?

lib/python/picongpu/picmi/diagnostics/binning.py
```diff
@@ -0,0 +1,191 @@
...
+        dumpPeriod: int = 1,
```
BrianMarre, line 155: probably a dumb question what happens in the binning plugin if dumpPeriod < period ?

ikbuibui: Dump period is defined as number of notifies after which a file is written out, so it is orthogonal to the period. A period of 5 and a dump period of 2 would mean a file is written every 10 steps, with data accumulated over 2 notifies.

lib/python/picongpu/picmi/diagnostics/binning.py
ikbuibui: Its the current behaviour in C++. Im not sure either if it is a sensible default. Time averaging applies when dump period is set to greater than 1. Then it averages the values in the binning by the dump period. The original reasoning behind making it enabled by default was so that values in bins can be understood and compared independent of the sump period.

lib/python/picongpu/picmi/diagnostics/binning.py
chillenzer: Nope, I don't want that. That's precisely the kind of thing that leads to users coming over and complaining why they have to cast every numpy array to a list. The intent of the user is obvious when they give a single species, so we should use the one obvious interpretation and go with it.

lib/python/picongpu/picmi/diagnostics/binning.py
BrianMarre: Okay, we should keep the C++ and python defaults consistent, although I would propose to change both defaults in a follow up PR

lib/python/picongpu/picmi/diagnostics/binning.py
BrianMarre: Thanks for the clarification, this should be documented here to ensure that the user knows that dumpPeriod is in notfies, not simulation-steps as I assumed

PR #5410, opened 2025-07-14
By using zsh one would get the following error:

grussdorian:
When using zsh the following error occurred:

```
dirname: missing operand
Try 'dirname --help' for more information.
basename: missing operand
Try 'basename --help' for more information
```

The error occurs because the script uses Bash-specific syntax ($BASH_SOURCE) which is not available in zsh

`export PIC_PROFILE=$(cd $(dirname ${BASH_SOURCE:-$0}) && pwd)/$(basename ${BASH_SOURCE:-$0})`

$0 in Zsh contains the path to the current script, similar to $BASH_SOURCE in Bash.
dirname $0 extracts the directory path of the script.
basename $0 extracts the script's filename.
$(cd $(dirname $0) && pwd) changes to the script's directory and outputs the full path, ensuring it's absolute.

This checks for $BASH_SOURCE (for Bash) and falls back to $0 (for Zsh)

### DIFF
--- a/etc/picongpu/hemera-hzdr/fwkt_v100_picongpu.profile.example
+++ b/etc/picongpu/hemera-hzdr/fwkt_v100_picongpu.profile.example
@@ -1,6 +1,5 @@
 # Name and Path of this Script ############################### (DO NOT change!)
-export PIC_PROFILE=$(cd $(dirname $BASH_SOURCE) && pwd)"/"$(basename $BASH_SOURCE)
-
+export PIC_PROFILE=$(cd $(dirname ${BASH_SOURCE:-$0}) && pwd)/$(basename ${BASH_SOURCE:-$0}) # for compatibility with both zsh and bash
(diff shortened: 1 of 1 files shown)


ikbuibui:
Thanks for the PR. I know in our group atleast @chillenzer is using zsh and has his own versions of these files. 
@chillenzer @psychocoderHPC What is your opinion on support other shells?
If we accept this PR to support alternative shells, this change must also be made in all the many other profiles we have. Another option is to add documentation for known modifications required for particular system setups and not actually change the source, but this is also on best effort basis, based on what software is being used by this community.

chillenzer (approved):
This works fine for bash as well as zsh, so there's no downside in having it, I'd presume.

chillenzer:
Just for context: https://stackoverflow.com/questions/35006457/choosing-between-0-and-bash-source/35006505#35006505

PR #5416, opened 2025-07-25
Improve pmacc dataspace 

ikbuibui:
Uses `index_sequence` based parameter pack expansions instead of for loops in `DataSpace`. Im not sure about this change, because although it helps prevent first constructing and then assigning in many cases, it is harder to read. Also most likely the compiler is quite smart and sees through the for loops, but I haven't looked into the assembly to compare.
 
Also removes user defined defaulted copy constructor for `Selection` and `DataSpace`, as this causes clang to throw a warning 
```
Definition of implicit copy assignment operator for 'Selection<3>' is deprecated because it has a user-declared copy constructor (clang -Wdeprecated-copy)
```

### DIFF
--- a/include/pmacc/dimensions/DataSpace.hpp
+++ b/include/pmacc/dimensions/DataSpace.hpp
@@ -65,13 +65,10 @@ namespace pmacc
         {
         }
 
-        constexpr DataSpace(DataSpace const&) = default;
-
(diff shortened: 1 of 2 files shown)


ikbuibui:
Currently this causes a segfault inside the compiler with gcc11 and cuda 12.1

chillenzer:
What's the status of this?

PR #5289, opened 2025-02-20
Flexible accumulation for binning

ikbuibui:
Add the ability to specify the accumulation operation for the binning plugin

PR #5417, opened 2025-07-25
Fix alternate reduction operations in the binning plugin

ikbuibui:
The binning plugin now correctly reduces the histograms. Earlier the MPI reduction between GPUs was always done by addition and the neutral element for all reduction operations was assumed to be zero.
This was broken since the feature release in #5289.
This is fixed now and binning uses the correct MPI op and neutral element.

For users, this means that they must use `pmacc::math::operation`s now as the accumulate op, instead of alpaka atomics.
Documentation has been updated with valid ops.

### DIFF
--- a/include/pmacc/math/operation/Or.hpp
+++ b/include/pmacc/math/operation/Or.hpp
@@ -21,6 +21,7 @@
 
 #pragma once
 
+#include "pmacc/mpi/GetMPI_Op.hpp"
 #include "pmacc/types.hpp"
 
 #include <cstdint>
@@ -42,3 +43,12 @@ namespace pmacc::math::operation
         }
     };
 } // namespace pmacc::math::operation
+
+namespace pmacc::mpi
+{
+    template<>
+    HINLINE MPI_Op getMPI_Op<pmacc::math::operation::Or>()
+    {
+        return MPI_LOR;
+    }
+} // namespace pmacc::mpi
--- a/include/pmacc/math/operation/Sub.hpp
+++ b/include/pmacc/math/operation/Sub.hpp
@@ -21,6 +21,7 @@
 
 #pragma once
 
+#include "pmacc/math/operation/traits.hpp"
 #include "pmacc/types.hpp"
 
 namespace pmacc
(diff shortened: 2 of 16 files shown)


PrometheusPi (approved):
looks good - have you ran tests with e.g. max or min operators?

ikbuibui:
Yes, it was tested with max, although quite trivially. I tried a particle counter and checked that every location with non zero number of particle returns the value of one, since instead of  adding, we now just keep looking at the max of 1 and 1

include/pmacc/math/operation/And.hpp
```diff
@@ -42,3 +44,12 @@ namespace pmacc::math::operation
...
 } // namespace pmacc::math::operation
+
+namespace pmacc::mpi
+{
+    template<>
+    HINLINE MPI_Op getMPI_Op<pmacc::math::operation::And>()
+    {
+        return MPI_LAND;
+    }
+} // namespace pmacc::mpi
```
chillenzer, line 55: Have I overlooked something or does this lack a neutral element?

include/pmacc/math/operation/Or.hpp
```diff
@@ -42,3 +43,12 @@ namespace pmacc::math::operation
...
 } // namespace pmacc::math::operation
+
+namespace pmacc::mpi
+{
+    template<>
+    HINLINE MPI_Op getMPI_Op<pmacc::math::operation::Or>()
+    {
+        return MPI_LOR;
+    }
+} // namespace pmacc::mpi
```
chillenzer, line 54: Same here.

include/pmacc/math/operation/Sub.hpp
```diff
@@ -43,6 +44,16 @@ namespace pmacc
...
+
+            namespace traits
+            {
+                template<>
+                struct AlpakaAtomicOp<Sub>
+                {
+                    using type = alpaka::AtomicSub;
+                };
+            } // namespace traits
+
```
chillenzer, line 56: And here?

chillenzer (commented):
Sorry for being late but I had made a note to check on the neutral elements explicitly.

include/pmacc/math/operation/And.hpp
ikbuibui: In principle a neutral element does exist for this (namwly true). I did not bother with this and the others, most notably multiplication, because they lack support for reduction as the corresponding alpaka atomics dont exist. These can be added in the future if we decide to somehow support reduction for these ops

PR #5419, opened 2025-08-07
Fix free-formula expression handling and call operator

### LINKED ISSUE #5411: To specific Error in AnalyticDistribution lamdify try block, and potential optimization with numba
@chillenzer, [This](https://github.com/ComputationalRadiationPhysics/picongpu/blob/565d6615a43a729cd5a52b399c0ca731bd17e38a/lib/python/picongpu/picmi/distribution/AnalyticDistribution.py#L165C16-L165C26) can also fail with a `TypeError` too, for example:
`TypeError: invalid entry 1 in condlist: should be boolean ndarray`
In my particular case your vectorize also doesn't solve the problem. This is because you are vectorizing the sympy expression itself and not a non numpy "lambdification", which will only work for some expressions.  So I think maybe we should also change the alternative method to:

```
vectorize(sympy.lambdify(sympy.symbols("x,y,z"), self.density_expression, "math"))(x,y,z)
```
, which worked for me now. 

actually sth that works also great and maybe even faster than `sympy.lambdify` with `numpy` support is:
```
numba.vectorize(sympy.lambdify((x,y,z), dens_profile.density_expression, ["math"]))
```
That is also creating a `numpy` compatible function. Or even faster with a parallel execution: 
```
f = numba.njit(sympy.lambdify((x,y,z), dens_profile.density_expression, ["math"]))
@numba.vectorize([numba.float64(numba.float64, numba.float64, numba.float64)], target='parallel')
def fn(x,y,z):
    return  f(x,y,z)
```
Though, one needs to specify the signature here, but multiple signatures are also possible. So we should cover most cases (and all reasonable cases xD) with a `float64` and a `float32` signature. 

You could even go crazy and use `numba`'s `cuda` target :D

So maybe we could consider automatically switching to `numba` if it is installed. Making it a hard dependency is probably not a good idea, since `numba` tends to have some strong restrictions on `numpy` and `python` versions.

chillenzer:
Closes #5411 

This fix removes the `sympy.simplify` in handling density expressions and broadens the caught exception in the call operator of `AnalyticDistribution`. The hope is to make the use of the latter more robust.

### DIFF
--- a/lib/python/picongpu/picmi/distribution/AnalyticDistribution.py
+++ b/lib/python/picongpu/picmi/distribution/AnalyticDistribution.py
@@ -120,26 +120,13 @@ class AnalyticDistribution:
         self.rms_velocity = (0.0, 0.0, 0.0)
         self.directed_velocity = tuple(float(v) for v in directed_velocity)
         x, y, z = sympy.symbols("x,y,z")
-        self.density_expression = (
-            # We add the simplify because of the following:
-            # Translating to C++ requires ALL cases of a Piecewise to be defined
-            # such that there's a fallback for if-conditions.
-            # The user might have written their formula piecing together
-            # multiple partial Piecewise instances.
-            # Without simplification, sympy tries to translate them individually.
-            # and fails to do so even if they supplement each other
-            # into a function defined everywhere.
-            sympy.simplify(self.density_function(x, y, z))
-            # density_expression might be independent of any or all of the three variables.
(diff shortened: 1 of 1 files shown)


pordyna (approved):

pordyna:
@psychocoderHPC looks good, can be merged

psychocoderHPC (approved):

PR #5421, opened 2025-08-08
Add lasers to picmi

chillenzer:
This PR adds the PlaneWaveLaser, DispersivePulse and FromOpenPMDPulse to the PICMI interface and sets up the infrastructure for further lasers.

### DIFF
--- a/lib/python/picongpu/picmi/dispersive_pulse_laser.py
+++ b/lib/python/picongpu/picmi/dispersive_pulse_laser.py
@@ -1,0 +1,217 @@
+"""
+This file is part of PIConGPU.
+Copyright 2021-2024 PIConGPU contributors
+Authors: Julian Lenz
+License: GPLv3+
+"""
+
+from ..pypicongpu import laser
+from . import constants
+
+import math
+
+import typeguard
+import typing
+
+
+@typeguard.typechecked
+class DispersivePulseLaser:
+    """
+    Specifies a dispersive Gaussian pulse with dispersion parameters
+
+    Parameters
+    ----------
+    wavelength: float
+        Laser wavelength [m], defined as :math:`\\lambda_0` in the above formula
+
+    duration: float
+        Duration of the Gaussian pulse [s], defined as :math:`\\tau` in the above formula
+
+    propagation_direction: unit vector of length 3 of floats
+        Direction of propagation [1]
+
+    polarization_direction: unit vector of length 3 of floats
+        Direction of polarization [1]
+
+    focal_position: vector of length 3 of floats
+        Position of the laser focus [m]
+
+    centroid_position: vector of length 3 of floats
+        Position of the laser centroid at time 0 [m]
+
+    a0: float
+        Normalized vector potential at focus
+        Specify either a0 or E0 (E0 takes precedence).
+
+    E0: float
+        Maximum amplitude of the laser field [V/m]
+        Specify either a0 or E0 (E0 takes precedence).
+
+    phi0: float
+        Carrier envelope phase (CEP) [rad]
+
+    spectral_support: float
+        Width of the spectral support for the discrete Fourier transform [none]
+
+    sd_si: float
+        Spatial dispersion in focus [m*s]
+
+    ad_si: float
+        Angular dispersion in focus [rad*s]
+
+    gdd_si: float
+        Group velocity dispersion in focus [s^2]
+
+    tod_si: float
+        Third order dispersion in focus [s^3]
+    """
+
+    def __init__(
+        self,
+        waist,
+        wavelength,
+        duration,
+        propagation_direction,
+        polarization_direction,
+        focal_position,
+        centroid_position,
+        a0=None,
+        E0=None,
+        phi0=None,
+        picongpu_phase: float = 0.0,
+        picongpu_polarization_type=(laser.DispersivePulseLaser.PolarizationType.LINEAR),
+        picongpu_spectral_support: float = 6.0,
+        picongpu_sd_si: float = 0.0,
+        picongpu_ad_si: float = 0.0,
+        picongpu_gdd_si: float = 0.0,
+        picongpu_tod_si: float = 0.0,
+        # make sure to always place Huygens-surface inside PML-boundaries,
+        # default is valid for standard PMLs
+        # @todo create check for insufficient dimension
+        # @todo create check in simulation for conflict between PMLs and
+        # Huygens-surfaces
+        picongpu_huygens_surface_positions: typing.List[typing.List[int]] = [
+            [16, -16],
+            [16, -16],
+            [16, -16],
+        ],
+        **kw,
+    ):
+        assert E0 is not None or a0 is not None, "One of E0 or a0 must be speficied"
+
+        k0 = 2.0 * math.pi / wavelength
+        if E0 is None:
+            E0 = a0 * constants.m_e * constants.c**2 * k0 / constants.q_e
+        if a0 is None:
+            a0 = E0 / (constants.m_e * constants.c**2 * k0 / constants.q_e)
+
+        self.waist = waist
+        self.wavelength = wavelength
+        self.k0 = k0
+        self.duration = duration
+        self.focal_position = focal_position
+        self.centroid_position = centroid_position
+        self.propagation_direction = propagation_direction
+        self.polarization_direction = polarization_direction
+        self.a0 = a0
+        self.E0 = E0
+        self.phi0 = phi0
+
+        self.picongpu_phase = picongpu_phase
+        self.picongpu_polarization_type = picongpu_polarization_type
+        self.picongpu_spectral_support = picongpu_spectral_support
+        self.picongpu_sd_si = picongpu_sd_si
+        self.picongpu_ad_si = picongpu_ad_si
+        self.picongpu_gdd_si = picongpu_gdd_si
+        self.picongpu_tod_si = picongpu_tod_si
+        self.picongpu_huygens_surface_positions = picongpu_huygens_surface_positions
+
+    """PICMI object for Dispersive Pulse Laser"""
+
+    def scalarProduct(self, a: typing.List[float], b: typing.List[float]) -> float:
+        assert len(a) == len(b), "the scalar product is only defined for two \
+            vector of equal dimension"
+
+        result = 0.0
+        for i in range(len(a)):
+            result += a[i] * b[i]
+
+        return result
+
+    @staticmethod
+    def testRelativeError(trueValue, testValue, relativeErrorLimit):
+        return abs((testValue - trueValue) / trueValue) < relativeErrorLimit
+
+    def get_as_pypicongpu(self) -> laser.DispersivePulseLaser:
+        assert self.testRelativeError(
+            1,
+            self.scalarProduct(self.polarization_direction, self.polarization_direction),
+            1e-9,
+        ), "the polarization direction vector must be normalized"
+
+        # check for excessive phase values to avoid numerical precision errors
+        assert abs(self.picongpu_phase) <= 2 * math.pi, "abs(phase) must be < 2*pi"
+
+        # check that initialising from y_min-plane only is sensible
+        assert (
+            self.scalarProduct(self.propagation_direction, [0.0, 1.0, 0.0]) > 0.0
+        ), "laser propagation parallel to the y-plane or pointing outside \
+            from the inside of the simulation box is not supported by this \
+            laser in PIConGPU"
+        assert self.testRelativeError(
+            1,
+            (
+                self.propagation_direction[0] ** 2
+                + self.propagation_direction[1] ** 2
+                + self.propagation_direction[2] ** 2
+            ),
+            1e-9,
+        ), "propagation vector must be normalized"
+
+        # check centroid outside box
+        assert self.centroid_position[1] <= 0, "the laser maximum must be \
+            outside of the \
+            simulation box, otherwise it is impossible to correctly initialize\
+            it using a huygens surface in the box, centroid_y <= 0"
+        # @todo implement check that laser field strength sufficiently small
+        # at simulation box boundary
+
+        # check polarization vector normalization
+
+        assert self.testRelativeError(
+            1,
+            (
+                self.propagation_direction[0] ** 2
+                + self.propagation_direction[1] ** 2
+                + self.propagation_direction[2] ** 2
+            ),
+            1e-9,
+        ), "polarization vector must be normalized"
+
+        pypicongpu_laser = laser.DispersivePulseLaser()
+        pypicongpu_laser.wavelength = self.wavelength
+        pypicongpu_laser.waist = self.waist
+        pypicongpu_laser.duration = self.duration
+        pypicongpu_laser.focus_pos = self.focal_position
+        pypicongpu_laser.phase = self.picongpu_phase
+        pypicongpu_laser.E0 = self.E0
+
+        pypicongpu_laser.pulse_init = max(
+            -2 * self.centroid_position[1] / (self.propagation_direction[1] * constants.c) / self.duration,
+            15,
+        )
+        # unit: duration
+
+        pypicongpu_laser.polarization_type = self.picongpu_polarization_type
+        pypicongpu_laser.polarization_direction = self.polarization_direction
+        pypicongpu_laser.propagation_direction = self.propagation_direction
+
+        pypicongpu_laser.spectral_support = self.picongpu_spectral_support
+        pypicongpu_laser.sd_si = self.picongpu_sd_si
+        pypicongpu_laser.ad_si = self.picongpu_ad_si
+        pypicongpu_laser.gdd_si = self.picongpu_gdd_si
+        pypicongpu_laser.tod_si = self.picongpu_tod_si
+
+        pypicongpu_laser.huygens_surface_positions = self.picongpu_huygens_surface_positions
+
+        return pypicongpu_laser
--- a/lib/python/picongpu/picmi/lasers/base_laser.py
+++ b/lib/python/picongpu/picmi/lasers/base_laser.py
@@ -1,0 +1,84 @@
+"""
+This file is part of PIConGPU.
+Copyright 2025 PIConGPU contributors
+Authors: Julian Lenz
+License: GPLv3+
+"""
+
+
+import math
+import typing
+
+import numpy as np
+
+from .. import constants
+
+
+class BaseLaser:
+    """
+    Base class for all PICMI laser implementations to reduce code duplication
+    """
+
+    def _propagation_connects_centroid_and_focus(self):
+        # check that propagation_direction is parallel to the difference of focal_position and centroid_position
+        diff_vec = self.difference(self.focal_position, self.centroid_position)
+        cross_vec = self.crossProduct(diff_vec, self.propagation_direction)
+        length_of_cross_product = self.scalarProduct(cross_vec, cross_vec)
+        return length_of_cross_product < 1.0e-5
+
+    def _compute_E0_and_a0(self, k0, E0, a0):
+        if E0 is not None or a0 is not None:
+            raise ValueError(f"One of E0 or a0 must be speficied. You gave {E0=} and {a0=}.")
+        if E0 is not None and a0 is not None:
+            raise ValueError("At least one of E0 or a0 must be specified.")
+
+        if E0 is None:
+            E0 = a0 * constants.m_e * constants.c**2 * k0 / constants.q_e
+        if a0 is None:
+            a0 = E0 / (constants.m_e * constants.c**2 * k0 / constants.q_e)
+        return a0, E0
+
+    def scalarProduct(self, a: typing.List[float], b: typing.List[float]) -> float:
+        return np.dot(a, b).tolist()
+
+    def crossProduct(self, a: typing.List[float], b: typing.List[float]) -> typing.List[float]:
+        return np.linalg.cross(a, b).tolist()
+
+    def difference(self, a: typing.List[float], b: typing.List[float]) -> typing.List[float]:
+        return (np.asarray(a) - b).tolist()
+
+    @staticmethod
+    def testRelativeError(trueValue, testValue, relativeErrorLimit):
+        return abs((testValue - trueValue) / trueValue) < relativeErrorLimit
+
+    def _validate_common_properties(self):
+        """Common validation logic for all lasers"""
+        if not np.allclose(n := np.linalg.norm(self.polarization_direction), 1):
+            raise ValueError(
+                "The polarization direction vector must be normalized. "
+                f"You gave {self.polarization_direction=} with norm {n}."
+            )
+
+        if not np.allclose(n := np.linalg.norm(self.propagation_direction), 1):
+            raise ValueError(
+                "The propagation direction vector must be normalized. "
+                f"You gave {self.propagation_direction=} with norm {n}."
+            )
+
+        if abs(self.phi0) > 2 * math.pi:
+            raise ValueError(f"abs(phase) must be < 2*pi. You gave{self.phi0=}.")
+
+        if self.scalarProduct(self.propagation_direction, [0.0, 1.0, 0.0]) <= 0.0:
+            raise ValueError(
+                "Laser propagation parallel to the y-plane or pointing outside "
+                "from the inside of the simulation box is not supported by this "
+                f"laser in PIConGPU. You gave {self.propagation_direction=}."
+            )
+
+        if self.centroid_position[1] > 0:
+            raise ValueError(
+                "The laser maximum must be outside of the "
+                "simulation box, otherwise it is impossible to correctly initialize"
+                "it using a huygens surface in the box, centroid_y <= 0. "
+                f"You gave {self.centroid_position=}."
+            )
--- a/lib/python/picongpu/picmi/lasers/base_laser.py
+++ b/lib/python/picongpu/picmi/lasers/base_laser.py
@@ -5,13 +5,23 @@ Authors: Julian Lenz
 License: GPLv3+
 """
 
-
-import math
-import typing
+import logging
 
 import numpy as np
 
 from .. import constants
+
+
+def scalarProduct(a: list[float], b: list[float]) -> float:
+    return np.dot(a, b).tolist()
+
+
+def crossProduct(a: list[float], b: list[float]) -> list[float]:
+    return np.cross(a, b).tolist()
+
+
+def difference(a: list[float], b: list[float]) -> list[float]:
+    return (np.asarray(a) - b).tolist()
 
 
 class BaseLaser:
@@ -21,9 +31,9 @@ class BaseLaser:
 
     def _propagation_connects_centroid_and_focus(self):
         # check that propagation_direction is parallel to the difference of focal_position and centroid_position
-        diff_vec = self.difference(self.focal_position, self.centroid_position)
-        cross_vec = self.crossProduct(diff_vec, self.propagation_direction)
-        length_of_cross_product = self.scalarProduct(cross_vec, cross_vec)
+        diff_vec = difference(self.focal_position, self.centroid_position)
+        cross_vec = crossProduct(diff_vec, self.propagation_direction)
+        length_of_cross_product = scalarProduct(cross_vec, cross_vec)
         return length_of_cross_product < 1.0e-5
 
     def _compute_E0_and_a0(self, k0, E0, a0):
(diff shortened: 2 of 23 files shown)


share/picongpu/pypicongpu/template/include/picongpu/param/incidentField.param.mustache
```diff
@@ -195,4 +216,4 @@ namespace picongpu::fields::incidentField
     using ZMin = profiles::None;
     using ZMax = profiles::None;
     /**@}*/
-} // namespace picongpu::fields::incidentField
+}
```
PrometheusPi: Why was the closing comment removed?

share/picongpu/pypicongpu/schema/laser.PlaneWaveLaser.json
```diff
@@ -0,0 +1,141 @@
...
+        "required": [
+          "component"
+        ],
+        "unevaluatedProperties": false,
+        "properties": {
+          "component": {
+            "type": "number"
+          }
+        }
+      }
```
PrometheusPi, line 32: there should be no focus position in the plane wave

lib/python/picongpu/pypicongpu/simulation.py
```diff
@@ -45,7 +45,7 @@ class Simulation(RenderedObject):
     grid = util.build_typesafe_property(typing.Union[Grid3D])
     """Used grid Object"""
 
-    laser = util.build_typesafe_property(typing.Optional[GaussianLaser])
+    laser = util.build_typesafe_property(typing.Union[GaussianLaser, PlaneWaveLaser, None])
```
PrometheusPi: What is `typing.Union` now doing differently?

lib/python/picongpu/picmi/plane_wave_laser.py
```diff
@@ -0,0 +1,184 @@
...
+        Duration of the Gaussian pulse [s], defined as :math:`\\tau` in the above formula
+
+    propagation_direction: unit vector of length 3 of floats
+        Direction of propagation [1]
+
+    polarization_direction: unit vector of length 3 of floats
+        Direction of polarization [1]
+
+    focal_position: vector of length 3 of floats
+        Position of the laser focus [m]
```
PrometheusPi: not needed

lib/python/picongpu/picmi/plane_wave_laser.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def scalarProduct(self, a: typing.List[float], b: typing.List[float]) -> float:
+        assert len(a) == len(b), "the scalar product is only defined for two \
+            vector of equal dimension"
+
+        result = 0.0
+        for i in range(len(a)):
+            result += a[i] * b[i]
+
+        return result
```
PrometheusPi: This looks like something that should be inherited

lib/python/picongpu/picmi/plane_wave_laser.py
```diff
@@ -0,0 +1,184 @@
...
+
+    @staticmethod
+    def testRelativeError(trueValue, testValue, relativeErrorLimit):
+        return abs((testValue - trueValue) / trueValue) < relativeErrorLimit
```
PrometheusPi: same here

lib/python/picongpu/pypicongpu/laser.py
```diff
@@ -1,27 +1,33 @@
 """
 This file is part of PIConGPU.
 Copyright 2021-2024 PIConGPU contributors
-Authors: Hannes Troepgen, Brian Edward Marre, Alexander Debus
+Authors: Hannes Troepgen, Brian Edward Marre, Alexander Debus, Julian Lenz
 License: GPLv3+
 """
 
+from .rendering import SelfRegisteringRenderedObject
```
PrometheusPi, line 8: Why is there this change in import order and in what aspect does `SelfRegisteringRenderedObject` differ from `RenderedObject`?

lib/python/picongpu/pypicongpu/laser.py
```diff
@@ -102,3 +108,80 @@ def _get_serialized(self) -> dict:
...
+    Holds Parameters to specify a plane wave laser
+    """
+
+    _name = "planewave"
+
+    class PolarizationType(enum.Enum):
+        """represents a polarization of a laser (for PIConGPU)"""
+
+        LINEAR = 1
+        CIRCULAR = 2
```
PrometheusPi: Why are these class-level variables and not instance-level variables? 
Ffor these two variables, I get it, since they are global, but especially later (wavelength etc.) wouldn't make instance-level make more sense?

lib/python/picongpu/pypicongpu/laser.py
```diff
@@ -102,3 +108,80 @@ def _get_serialized(self) -> dict:
...
+    """propagation direction(normalized vector)"""
+    polarization_type = util.build_typesafe_property(PolarizationType)
+    """laser polarization"""
+    polarization_direction = util.build_typesafe_property(typing.List[float])
+    """direction of polarization(normalized vector)"""
+    laser_nofocus_constant_si = util.build_typesafe_property(float)
+    """constant for plane wave laser without focus (unitless)"""
+    huygens_surface_positions = util.build_typesafe_property(typing.List[typing.List[int]])
+    """Position in cells of the Huygens surface relative to start/
+       edge(negative numbers) of the total domain"""
```
PrometheusPi: here I  do not understand the class-level definition

lib/python/picongpu/picmi/plane_wave_laser.py
```diff
@@ -0,0 +1,184 @@
...
+        pypicongpu_laser.wavelength = self.wavelength
+        pypicongpu_laser.duration = self.duration
+        pypicongpu_laser.focus_pos = [0.0, 0.0, 0.0]
+        pypicongpu_laser.phase = self.picongpu_phase
+        pypicongpu_laser.E0 = self.E0
+
+        pypicongpu_laser.pulse_init = max(
+            -2 * self.centroid_position[1] / (self.propagation_direction[1] * constants.c) / self.duration,
+            15,
+        )
```
PrometheusPi: This looks strange. It defines ramp_init in `[incidentField.param.mustache](https://github.com/ComputationalRadiationPhysics/picongpu/pull/5421/files#diff-b3d5882b2f21fec96f70e826331edb86f88257edaca0e6e22bfbf14621231db8)`. 
But it is pretty much uneffected by centroid and falls back to PIConGPUs default `RAMP_INIT` up until a magic threshold.

PrometheusPi (changes requested):

lib/python/picongpu/picmi/dispersive_pulse_laser.py
```diff
@@ -0,0 +1,217 @@
...
+        waist,
+        wavelength,
+        duration,
+        propagation_direction,
+        polarization_direction,
+        focal_position,
+        centroid_position,
+        a0=None,
+        E0=None,
+        phi0=None,
```
pordyna: why do we have phi0 and picongpu_phase? Is there difference?

lib/python/picongpu/picmi/dispersive_pulse_laser.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def scalarProduct(self, a: typing.List[float], b: typing.List[float]) -> float:
+        assert len(a) == len(b), "the scalar product is only defined for two \
+            vector of equal dimension"
+
+        result = 0.0
+        for i in range(len(a)):
+            result += a[i] * b[i]
+
+        return result
```
pordyna: Why is this a class method? Maybe put it in a utility function? Or just use `list(np.dot(a, b))`.

lib/python/picongpu/picmi/dispersive_pulse_laser.py
```diff
@@ -0,0 +1,217 @@
...
+
+    @staticmethod
+    def testRelativeError(trueValue, testValue, relativeErrorLimit):
+        return abs((testValue - trueValue) / trueValue) < relativeErrorLimit
```
pordyna: again, looks like it shoudn't be part of this class^^

lib/python/picongpu/picmi/dispersive_pulse_laser.py
```diff
@@ -0,0 +1,217 @@
...
+            1e-9,
+        ), "propagation vector must be normalized"
+
+        # check centroid outside box
+        assert self.centroid_position[1] <= 0, "the laser maximum must be \
+            outside of the \
+            simulation box, otherwise it is impossible to correctly initialize\
+            it using a huygens surface in the box, centroid_y <= 0"
+        # @todo implement check that laser field strength sufficiently small
+        # at simulation box boundary
```
pordyna: @steindev @PrometheusPi @BrianMarre I feel like this check is quite arbitrary? Maybe someone for some weird reason would like to only use the pulse down ramp. We always cut some preceding intensity anyway, so why do we check that the center is in the box?

pordyna: Ok I see this is so that you can define pulse init from it. I guess for centroid in the box this still works with an arbitrary pulse init and a delay.

pordyna: Another thing, the centroid could still be outside x or z (even with the check above), but you only check the y component.

lib/python/picongpu/picmi/plane_wave_laser.py
```diff
@@ -0,0 +1,184 @@
...
+        self,
+        wavelength,
+        duration,
+        propagation_direction,
+        polarization_direction,
+        centroid_position,
+        a0=None,
+        E0=None,
+        phi0=None,
+        picongpu_phase: float = 0.0,
```
pordyna: same as above

lib/python/picongpu/pypicongpu/laser.py
pordyna: @PrometheusPi I think here this is  just a python implementation of an enum class. And below these are actually member variables, pypicongpu uses currently pydantic models as base class. These are basically python dataclasess with type checks and this is the syntax for defining members.

lib/python/picongpu/pypicongpu/laser.py
pordyna: @PrometheusPi see comment above

lib/python/picongpu/pypicongpu/laser.py
```diff
@@ -102,3 +108,172 @@ def _get_serialized(self) -> dict:
...
+                "row_y": {
+                    "negative": self.huygens_surface_positions[1][0],
+                    "positive": self.huygens_surface_positions[1][1],
+                },
+                "row_z": {
+                    "negative": self.huygens_surface_positions[2][0],
+                    "positive": self.huygens_surface_positions[2][1],
+                },
+            },
+        }
```
pordyna: I guess this applies to everywhere we define the surfaces. What do we do if we have more than one laser? Huygens surface position are the same for all lasers in PIConGPU.

lib/python/picongpu/pypicongpu/laser.py
```diff
@@ -102,3 +108,172 @@ def _get_serialized(self) -> dict:
...
+        LINEAR = 1
+        CIRCULAR = 2
+
+        def get_cpp_str(self) -> str:
+            """retrieve name as used in c++ param files"""
+            cpp_by_ptype = {
+                DispersivePulseLaser.PolarizationType.LINEAR: "Linear",
+                DispersivePulseLaser.PolarizationType.CIRCULAR: "Circular",
+            }
+            return cpp_by_ptype[self]
```
pordyna: Probably should be external and not part of the class, since it is used in more implementations

pordyna (commented):
Had only time to look over the Python files so far.

lib/python/picongpu/pypicongpu/simulation.py
chillenzer: `typing.Union` means that any of the given types is allowed. In that sense, `typing.Optional[...] == typing.Union[..., None]` but apparently they don't fancy nesting.

lib/python/picongpu/picmi/plane_wave_laser.py
chillenzer: See #5422

lib/python/picongpu/picmi/dispersive_pulse_laser.py
BrianMarre: Good catch, I wasn't sure what `phi0` was supposed to encode. I therefore erred on the side of caution and introduced the laser phase as a new parameter, but you are right phi0 is supposed to be the laser phase according to the PICMI documentation.

lib/python/picongpu/picmi/dispersive_pulse_laser.py
BrianMarre: because when I wrote that we still wanted to only use python inbuilt libraries, therefore no numpy, and it was only used in this implementation so I stuck it there.

lib/python/picongpu/picmi/dispersive_pulse_laser.py
BrianMarre: It is supposed to ensure that the laser actually becomes what the user set.
PICMI is supposed to be descriptive. The simulation should contain what your description set, not what a particular init method happens to produce. Consequently we disallow stuff that will not initialise as described by the laser.
If the user wants just the downramp in his simualtion than he needs to set that explicitly.

lib/python/picongpu/pypicongpu/laser.py
BrianMarre: @pordyna is right, this is a class local definition of an enum

BrianMarre: but since we now support more than one laser the definition should be moved outside of the laser definition.

lib/python/picongpu/pypicongpu/laser.py
BrianMarre: This the equivalent of a attribute definition in python. We define a set of attributes each instance of the class will have.
This is necessary to ensure that each instance will have the attribute, that the attribute will only accept values of the defined type and  that accessing that attribute prior to it being set will throw an error.
This is the old homemade style without using pydantic. With pydantic this gets much more readable.

lib/python/picongpu/pypicongpu/laser.py
BrianMarre: currently our PICMI implementation is still limited to one or no laser, so a future problem @chillenzer

lib/python/picongpu/pypicongpu/simulation.py
BrianMarre: we could also use the more modern style `Gaussian Laser | PlaneWaveLaser | None`

share/picongpu/pypicongpu/schema/laser.PlaneWaveLaser.json
chillenzer: That is following PIConGPU's header structure. `PlaneWaveParams` inherit from `BaseParams` which contain a focus position even though it is unused by plane wave lasers. But that value is internally set to 0 and not propagated to the public interface.

share/picongpu/pypicongpu/template/include/picongpu/param/incidentField.param.mustache
chillenzer: Re-instantiated.

lib/python/picongpu/picmi/plane_wave_laser.py
chillenzer: Done.

lib/python/picongpu/picmi/plane_wave_laser.py
chillenzer: Done.

lib/python/picongpu/pypicongpu/laser.py
chillenzer: I've ran `ruff`s import ordering tool to bring this into a canonical order. In fact, there's a linter rule about this and we should activate it at some point.

lib/python/picongpu/pypicongpu/laser.py
chillenzer: Done.

lib/python/picongpu/picmi/dispersive_pulse_laser.py
chillenzer: Done.

lib/python/picongpu/picmi/dispersive_pulse_laser.py
chillenzer: Done.

lib/python/picongpu/picmi/dispersive_pulse_laser.py
chillenzer: Done.

lib/python/picongpu/picmi/dispersive_pulse_laser.py
chillenzer: Beyond the scope of this PR

lib/python/picongpu/picmi/plane_wave_laser.py
chillenzer: Done.

lib/python/picongpu/pypicongpu/laser.py
chillenzer: Not yet applicable.

lib/python/picongpu/pypicongpu/laser.py
chillenzer: Done.

PrometheusPi:
@chillenzer The CI is failing with e.g.
> module 'numpy.linalg' has no attribute 'cross'

chillenzer:
Turns out that it's a numpy v2 feature to (also) have it in `np.linalg.` but my quick search only turned up that one.
https://numpy.org/doc/2.3/reference/generated/numpy.linalg.cross.html
https://numpy.org/doc/1.26/reference/generated/numpy.cross.html#numpy.cross

lib/python/picongpu/picmi/lasers/base_laser.py
```diff
@@ -0,0 +1,83 @@
...
+    def scalarProduct(self, a: typing.List[float], b: typing.List[float]) -> float:
+        return np.dot(a, b).tolist()
+
+    def crossProduct(self, a: typing.List[float], b: typing.List[float]) -> typing.List[float]:
+        return np.cross(a, b).tolist()
+
+    def difference(self, a: typing.List[float], b: typing.List[float]) -> typing.List[float]:
+        return (np.asarray(a) - b).tolist()
+
+    @staticmethod
```
PrometheusPi: Why is this a static method and e.g. the dot or cross product not?

lib/python/picongpu/picmi/lasers/base_laser.py
```diff
@@ -0,0 +1,83 @@
...
+                f"You gave {self.polarization_direction=} with norm {n}."
+            )
+
+        if not np.allclose(n := np.linalg.norm(self.propagation_direction), 1):
+            raise ValueError(
+                "The propagation direction vector must be normalized. "
+                f"You gave {self.propagation_direction=} with norm {n}."
+            )
+
+        if abs(self.phi0) > 2 * math.pi:
```
PrometheusPi: Why can the phase be within -2 pi and +2 pi? That covers a range of 4 pi.

lib/python/picongpu/picmi/lasers/base_laser.py
```diff
@@ -0,0 +1,83 @@
...
+            raise ValueError(f"abs(phase) must be < 2*pi. You gave{self.phi0=}.")
+
+        if self.scalarProduct(self.propagation_direction, [0.0, 1.0, 0.0]) <= 0.0:
+            raise ValueError(
+                "Laser propagation parallel to the y-plane or pointing outside "
+                "from the inside of the simulation box is not supported by this "
+                f"laser in PIConGPU. You gave {self.propagation_direction=}."
```
PrometheusPi:
```suggestion
                f"laser in PICMI. You gave {self.propagation_direction=}."
```

PIConGPU support such lasers.

lib/python/picongpu/picmi/lasers/dispersive_pulse_laser.py
```diff
@@ -0,0 +1,158 @@
...
+        pypicongpu_laser.wavelength = self.wavelength
+        pypicongpu_laser.waist = self.waist
+        pypicongpu_laser.duration = self.duration
+        pypicongpu_laser.focus_pos = self.focal_position
+        pypicongpu_laser.phase = self.phi0
+        pypicongpu_laser.E0 = self.E0
+
+        pypicongpu_laser.pulse_init = max(
+            -2 * self.centroid_position[1] / (self.propagation_direction[1] * constants.c) / self.duration,
+            15,
```
PrometheusPi: Please get rid of these ugly hard coded defaults of INIT = 15. 
These have already been removed by a previous PR.

lib/python/picongpu/picmi/lasers/plane_wave_laser.py
```diff
@@ -0,0 +1,120 @@
...
+        pypicongpu_laser = laser.PlaneWaveLaser()
+        pypicongpu_laser.wavelength = self.wavelength
+        pypicongpu_laser.duration = self.duration
+        pypicongpu_laser.focus_pos = [0.0, 0.0, 0.0]
+        pypicongpu_laser.phase = self.phi0
+        pypicongpu_laser.E0 = self.E0
+
+        pypicongpu_laser.pulse_init = max(
+            -2 * self.centroid_position[1] / (self.propagation_direction[1] * constants.c) / self.duration,
+            15,
```
PrometheusPi: again introduction of hard coded init=15, please remove

share/picongpu/pypicongpu/examples/laser_wakefield/main.py
```diff
@@ -69,9 +69,9 @@
         -0.5 * pulse_init * laser_duration * c,
         float(numberCells[2] * cellSize[2] / 2.0),
     ],
-    picongpu_polarization_type=pypicongpu.laser.GaussianLaser.PolarizationType.LINEAR,
+    picongpu_polarization_type=picmi.lasers.PolarizationType.LINEAR,
     a0=8.0,
-    picongpu_phase=0.0,
+    phi0=0.0,
```
PrometheusPi, line 79: should we set the (default) value manually or rely on the default being set internally?

share/picongpu/pypicongpu/schema/laser.DispersivePulseLaser.json
```diff
@@ -0,0 +1,169 @@
...
+      "description": "Waist of the Gaussian pulse at focus [m]",
+      "exclusiveMinimum": 0
+    },
+    "pulse_duration_si": {
+      "type": "number",
+      "description": "sigma of std. gauss for intensity (E^2) [s]",
+      "exclusiveMinimum": 0
+    },
+    "focus_pos_si": {
+      "description": "position of focus in total coordinate system, [m]",
```
PrometheusPi, line 22: total or global?

share/picongpu/pypicongpu/schema/laser.DispersivePulseLaser.json
```diff
@@ -0,0 +1,169 @@
...
+        "properties": {
+          "component": {
+            "type": "number"
+          }
+        }
+      }
+    },
+    "phase": {
+      "type": "number",
+      "minimum": -3.1416,
```
PrometheusPi: the actual minimum in code is - 2 pi thus -6.28

share/picongpu/pypicongpu/schema/laser.PlaneWaveLaser.json
```diff
@@ -0,0 +1,141 @@
...
+      "description": "wavelength of laser [m]",
+      "exclusiveMinimum": 0
+    },
+    "pulse_duration_si": {
+      "type": "number",
+      "description": "sigma of std. gauss for intensity (E^2) [s]",
+      "exclusiveMinimum": 0
+    },
+    "focus_pos_si": {
+      "description": "position of focus in total coordinate system, [m]",
```
PrometheusPi, line 17: see above: total or global

test/python/picongpu/quick/picmi/gaussian_laser.py
```diff
@@ -27,7 +26,7 @@ def test_basic(self):
             E0=5,
             picongpu_laguerre_modes=[2.0, 3.0],
             picongpu_laguerre_phases=[4.0, 5.0],
-            picongpu_phase=-2,
+            phi0=-2,
```
PrometheusPi, line 29: If we fix the phi range to 0 - 2pi this will be out of range

PrometheusPi (changes requested):
There are some reintroductions of magic numbers.

lib/python/picongpu/picmi/lasers/base_laser.py
chillenzer: Why do we have restrictions here at all?

share/picongpu/pypicongpu/examples/laser_wakefield/main.py
chillenzer: I don't quite see what you mean by this. But whatever you think could be the alternative, I think the current version expresses the intent quite clearly in the interface: Without even looking at the code, the caller will immediately know that not setting this will result in phi0 being 0.0 which is precisely what we want.

share/picongpu/pypicongpu/schema/laser.DispersivePulseLaser.json
chillenzer: This should definitely be total in the public interface because all other coordinate systems are algorithmic details and it translates to `FOCUS_POSITION_X_SI`, ... in the `BaseParams`. You're more of an expert than me in what that means.

lib/python/picongpu/picmi/lasers/base_laser.py
BrianMarre: Because large phase values create floating point inaccuracies in the `sin` and `cos` evaluation in the PIConGPU laser implementation.
Therefore the PIConGPU param files recommend keeping the phase value small. For why we use the 4 pi range, my guess would be because this allows a `%2pi` on the user side and Hannes did not consider this in detail. 
Strictly speaking we could limit the range further to [-pi, pi]. This would cover all possible unique phase values bijectively and force the user to pick a small value that avoid numerical issues.

share/picongpu/pypicongpu/examples/laser_wakefield/main.py
BrianMarre: the idea was to show the user what options has has without needing to look through the code, therefore it is set to the default value explicitly.

share/picongpu/pypicongpu/schema/laser.DispersivePulseLaser.json
BrianMarre: Good Catch! we should be consistent to either -pi or -2pi

lib/python/picongpu/picmi/lasers/base_laser.py
chillenzer: I'm aware of these arguments and still against restricting the interface:
- It is unlikely that the caller actually gives a very large argument.
- Arguments in the vicinity of the first period are kind of useful, e.g., for scanning in a more natural range. It can be very annoying and potentially ambiguous to unwrap after the fact.
- The caller is free to perform an appropriate modding operation themselves, if they think it's necessary. As any of the choices is arbitrary, it would be kind of arbitrary which one to implement/restrict to.

I would find it acceptable to internally (on the python level) insert some modding operation if there's a large majority in favour of it but I'm not a big fan.

PrometheusPi: I am fine with @chillenzer solution of removing the check

share/picongpu/pypicongpu/schema/laser.DispersivePulseLaser.json
PrometheusPi: Will not be limited anymore - thus no need for this - resolved

test/python/picongpu/quick/picmi/gaussian_laser.py
PrometheusPi: should be no issue with removing the ranges

share/picongpu/pypicongpu/schema/laser.DispersivePulseLaser.json
PrometheusPi: According to https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-domain-definitions, you are right. Since there has never been a difference between total and global (for us/for me), I didn't even know we had a total coordinate system 😄

share/picongpu/pypicongpu/schema/laser.PlaneWaveLaser.json
PrometheusPi: solved

lib/python/picongpu/picmi/lasers/dispersive_pulse_laser.py
PrometheusPi: I like your `_compute_pulse_init()` solution @chillenzer

lib/python/picongpu/picmi/lasers/plane_wave_laser.py
PrometheusPi: fixed via `_compute_pulse_init()`

lib/python/picongpu/picmi/lasers/base_laser.py
PrometheusPi: This is nitpicky anyway - can be ignored for now.

PrometheusPi (approved):

PrometheusPi:
@chillenzer did you perform run time tests?

chillenzer:
Only in the sense that everything seems to compile just fine. I only checked by changing the laser in the LWFA example. I plan to write proper tests similar to the free-formula density at the point when I implement the free-formula laser. Each laser will get a call operator and a tiny test case can check that the corresponding boundaries host the correct incident fields. This will give us an automated test (although running the end-to-end tests in CI is a different beast).

PrometheusPi:
This sounds fine enough for me. Physics comparisons are a bit more involved but as long as all newly introduced lasers actually compile, I am fine.

PrometheusPi:
@chillenzer is squashing part of your PR a good idea?

chillenzer:
Squashed as requested. We're one final CI run away from a merge, @PrometheusPi !

PR #5422, opened 2025-08-08
fix picmi laser centroid to laser init calculation

PrometheusPi:
The current implementation of the Gauss laser used a "cheat" `max()`-evaluation that resulted in a fallback to a `PULSE_INIT` of 15 in most cases and `PULSE_INIT` could never be set lower than 15. 
This PR removes that fallback and hands over the accountability to the user. It just throws a warning if we initialize for less than 3 half-pulse-durations.

### DIFF
--- a/share/picongpu/pypicongpu/examples/laser_wakefield/main.py
+++ b/share/picongpu/pypicongpu/examples/laser_wakefield/main.py
@@ -1,7 +1,7 @@
 """
 This file is part of PIConGPU.
 Copyright 2024 PIConGPU contributors
-Authors: Masoud Afshari, Brian Edward Marre
+Authors: Masoud Afshari, Brian Edward Marre, Richard Pausch
 License: GPLv3+
 """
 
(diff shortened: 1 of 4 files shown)


ikbuibui (changes requested):
The test fails currently because the centroid `y` of the Gaussian laser in `test/python/picongpu/compiling/species.py` is `0`. It should be updated to have a valid `pulse_init` now that the fallback value `15` has been removed.

PrometheusPi:
Thanks @ikbuibui for pointing this out. I updated the definition of the test.

chillenzer (dismissed):
I think `pulse_init` is not quite a self-explanatory name but in the example files, the formula what it actually does is written down, so that's fine.

ikbuibui (approved):

share/picongpu/pypicongpu/examples/laser_wakefield/main.py
```diff
@@ -64,7 +66,7 @@
     ],
     centroid_position=[
         float(numberCells[0] * cellSize[0] / 2.0),
-        0.0,
+        -0.5 * pulse_init * laser_duration,
```
PrometheusPi, line 69: here a `* c` is missing, as it should be a distance not a time - I will open a PR to fix it

PR #5423, opened 2025-08-08
Add range checks for parameters in atomicPhysics/IPD

steindev:
Avoids division by zero in computations.

### DIFF
--- a/include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/StewartPyattIPD.hpp
+++ b/include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/StewartPyattIPD.hpp
@@ -227,7 +227,9 @@ namespace picongpu::particles::atomicPhysics::ionizationPotentialDepression
 
             // (eV * sim.unit.length()) / (eV * sim.unit.length()), not weighted
             // unitless, not weighted
-            float_X const K = constFactor / (temperatureTimesk_Boltzman * debyeLength);
+            float_X const K = (debyeLength <= 0._X || temperatureTimesk_Boltzman <= 0._X)
+                                  ? 0._X
+                                  : constFactor / (temperatureTimesk_Boltzman * debyeLength);
 
             // unitless, not weighted
             float_X const zStar = zStarBox(superCellFieldIdx);
--- a/include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/StewartPyattIPD.hpp
+++ b/include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/StewartPyattIPD.hpp
@@ -227,7 +227,7 @@ namespace picongpu::particles::atomicPhysics::ionizationPotentialDepression
 
             // (eV * sim.unit.length()) / (eV * sim.unit.length()), not weighted
             // unitless, not weighted
-            float_X const K = (debyeLength <= 0._X || temperatureTimesk_Boltzman <= 0._X)
+            float_X const K = (pmacc::math::isApproxZero(temperatureTimesk_Boltzman * debyeLength))
                                   ? 0._X
                                   : constFactor / (temperatureTimesk_Boltzman * debyeLength);
 
--- a/include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/kernel/CalculateIPDInput.kernel
+++ b/include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/kernel/CalculateIPDInput.kernel
@@ -30,6 +30,7 @@
 #include <pmacc/particles/algorithm/ForEach.hpp>
 
 #include <cstdint>
+#include <limits>
 
 namespace picongpu::particles::atomicPhysics::ionizationPotentialDepression::kernel
 {
@@ -135,7 +136,10 @@ namespace picongpu::particles::atomicPhysics::ionizationPotentialDepression::ker
                 {
                     /* unit: (unitless * weight / typicalNumParticlesPerMacroParticle)
                      *  / (unitless * weight / typicalNumParticlesPerMacroParticle) = unitless */
-                    zStarBox(superCellFieldIdx) = sumChargeNumberSquared / sumChargeNumber;
+                    zStarBox(superCellFieldIdx)
+                        = (math::abs(sumChargeNumber) < std::numeric_limits<float_X>::epsilon())
+                              ? 0._X
+                              : sumChargeNumberSquared / sumChargeNumber;
 
                     // eV / unit_energy
                     constexpr float_X eV = sim.pic.get_eV();
@@ -144,7 +148,8 @@ namespace picongpu::particles::atomicPhysics::ionizationPotentialDepression::ker
                      *  / (weight / typicalNumParticlesPerMacroParticle) = unit_energy
                      *
                      * k_Boltzman * Temperature, energy equivalent to the local temperature */
-                    float_X const temperatureEnergy_PIC = sumTemperatureTerm / sumWeightAllNormalized;
+                    float_X const temperatureEnergy_PIC
+                        = (sumWeightAllNormalized > 0._X) ? sumTemperatureTerm / sumWeightAllNormalized : 0._X;
 
                     // (eV / unit_energy) * unit_energy = eV
                     temperatureEnergyBox(superCellFieldIdx) = eV * temperatureEnergy_PIC;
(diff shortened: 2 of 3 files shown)


BrianMarre (commented):
Looks good to me but I wrote part of the fix, so I declare myself biased.

ikbuibui (dismissed):

include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/kernel/CalculateIPDInput.kernel
```diff
@@ -135,7 +135,8 @@ namespace picongpu::particles::atomicPhysics::ionizationPotentialDepression::ker
                 {
                     /* unit: (unitless * weight / typicalNumParticlesPerMacroParticle)
                      *  / (unitless * weight / typicalNumParticlesPerMacroParticle) = unitless */
-                    zStarBox(superCellFieldIdx) = sumChargeNumberSquared / sumChargeNumber;
+                    zStarBox(superCellFieldIdx)
+                        = (sumChargeNumber != 0) ? sumChargeNumberSquared / sumChargeNumber : 0._X;
```
chillenzer: Floating point comparison with 0

include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/kernel/CalculateIPDInput.kernel
```diff
@@ -169,9 +171,12 @@ namespace picongpu::particles::atomicPhysics::ionizationPotentialDepression::ker
                      *
                      * unit:  sqrt(unit_length^2 / unit_energy / typicalNumParticlesPerMacroParticle
                      *  * unit_energy / (1/typicalNumParticlesPerMacroParticle)) = unit_length */
-                    debyeLengthBox(superCellFieldIdx) = math::sqrt(
-                        constFactorDebyeLength * temperatureEnergy_PIC
-                        / (sumChargeNumberSquared + sumWeightElectronNormalized));
+                    debyeLengthBox(superCellFieldIdx)
+                        = (sumChargeNumberSquared != 0._X || sumWeightElectronNormalized > 0._X)
```
chillenzer: Floating point comparison with 0

chillenzer (changes requested):

include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/kernel/CalculateIPDInput.kernel
steindev: So what do you suggest @chillenzer?

chillenzer: Typically, `abs(sumChargeNumber) < epsilon` with some small `epsilon`.

steindev:
@chillenzer: Comparison is now against `std::numeric_limits::epsilon()` as is used also in other parts of PIConGPU. Compilation is tested. Please re-review. Thanks!

ikbuibui:
Just for reference, these are the values we are comparing against when we compare to epsilon. These are relatively large compared to the values which can be represented in float/double. Care should be taken to make sure these values are appropriate
```
float epsilon:       1.192093e-07
double epsilon:      2.220446e-16
```

chillenzer (dismissed):

BrianMarre:
question, why are the formulas in StewartPyattIPD.hpp still comparing to 0?

include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/StewartPyattIPD.hpp
```diff
@@ -227,7 +227,9 @@ namespace picongpu::particles::atomicPhysics::ionizationPotentialDepression
 
             // (eV * sim.unit.length()) / (eV * sim.unit.length()), not weighted
             // unitless, not weighted
-            float_X const K = constFactor / (temperatureTimesk_Boltzman * debyeLength);
+            float_X const K = (debyeLength <= 0._X || temperatureTimesk_Boltzman <= 0._X)
```
PrometheusPi: Is this enough, or do you need an epsilon check as below? 
```suggestion
            float_X const K = ( (debyeLength * temperatureTimesk_Boltzman) < std::numeric_limits<float_X>::epsilon() )
```

include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/kernel/CalculateIPDInput.kernel
```diff
@@ -135,7 +136,10 @@ namespace picongpu::particles::atomicPhysics::ionizationPotentialDepression::ker
                 {
                     /* unit: (unitless * weight / typicalNumParticlesPerMacroParticle)
                      *  / (unitless * weight / typicalNumParticlesPerMacroParticle) = unitless */
-                    zStarBox(superCellFieldIdx) = sumChargeNumberSquared / sumChargeNumber;
+                    zStarBox(superCellFieldIdx)
+                        = (math::abs(sumChargeNumber) < std::numeric_limits<float_X>::epsilon())
+                              ? sumChargeNumberSquared / sumChargeNumber
+                              : 0._X;
```
PrometheusPi: shouldn't that be interchanged?

PrometheusPi: if `math::abs(sumChargeNumber)` you want to have zero, right? 
That's not the case right now

include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/kernel/CalculateIPDInput.kernel
```diff
@@ -169,9 +174,13 @@ namespace picongpu::particles::atomicPhysics::ionizationPotentialDepression::ker
                      *
                      * unit:  sqrt(unit_length^2 / unit_energy / typicalNumParticlesPerMacroParticle
                      *  * unit_energy / (1/typicalNumParticlesPerMacroParticle)) = unit_length */
-                    debyeLengthBox(superCellFieldIdx) = math::sqrt(
-                        constFactorDebyeLength * temperatureEnergy_PIC
-                        / (sumChargeNumberSquared + sumWeightElectronNormalized));
+                    debyeLengthBox(superCellFieldIdx)
+                        = (math::abs(sumChargeNumberSquared) < std::numeric_limits<float_X>::epsilon()
+                           || sumWeightElectronNormalized > 0._X)
```
PrometheusPi: here is the same issue

PrometheusPi (dismissed):
logic errors

include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/StewartPyattIPD.hpp
chillenzer: Depends.

The equality sign in there has virtually no impact in case the quantities are actually the result of floating point operations because hitting an exact 0 is unlikely. This could even mean that an analytically exact 0 acquires a tiny, positive contribution that would lift it over that limit. So, the original version more or less means `<` numerically (unless those values are set 0.0 explicitly because 0.0 is representable as a floating point number). If numerically computed 0.0 are a likely case, an epsilon check might be necessary.

Your version is inequivalent to the one above: If both quantities are large and negative, the original version will pass while your version will fail. Not sure if that's the logic error you wanted to point out.

include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/kernel/CalculateIPDInput.kernel
chillenzer: Yes, certainly!

include/picongpu/particles/atomicPhysics/ionizationPotentialDepression/kernel/CalculateIPDInput.kernel
chillenzer: That looks strange anyways... what is this supposed to do precisely?

ikbuibui:
I added some helpers to PMacc to compare floating point numbers

chillenzer (dismissed):

chillenzer:
One could have even abstracted away a function `divide_or(numerator, denominator, alt_value) = abs(denominator) > 0? numerator/denominator : alt_value` but it's fine with me. Those checks are probably a common pattern, one could scan the codebase and replace further usage.

PrometheusPi:
@steindev what is the status of this PR? Can we review it again?

steindev:
@PrometheusPi @chillenzer PR is squashed and rebased. All reviewer comments are in. Please re-review.

Thanks to @ikbuibui and @BrianMarre for contributing! :pray:

psychocoderHPC (approved):

PR #5424, opened 2025-08-08
fix bug in PICMI ionization configuration for some elements

BrianMarre:
I mistakenly assumed that the ``periodictable`` library returns a list of all charge states of an element when calling, ``Element.ions``.
It actually returns typically encountered chemical ion charge states. For example,

```python
periodictable.Be.ions = (1, 2)
periodictable.Ne = ()
periodictable.Cu = (-2, 1, 2, 3, 4)
```

Which throws an error for noble gases and returns  a wrong value for the charge of most species.
This error was obscured by a the ``test_charge`` test comparing two really small number for almost equal, causing the error to go unnoticed.

### DIFF
--- a/test/python/picongpu/quick/pypicongpu/species/util/element.py
+++ b/test/python/picongpu/quick/pypicongpu/species/util/element.py
@@ -16,24 +16,19 @@ from picongpu.pypicongpu.rendering import RenderedObject
 class TestElement(unittest.TestCase):
     def setUp(self):
         # create test case data
-        self.test_element = ["H", "#2H", "Cu", "#12C", "C"]
-        self.name = ["H", "D", "Cu", "C", "C"]
-        self.picongpu_names = ["Hydrogen", "Deuterium", "Copper", "Carbon", "Carbon"]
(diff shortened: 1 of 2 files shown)


test/python/picongpu/quick/pypicongpu/species/util/element.py
```diff
@@ -16,24 +16,19 @@
 class TestElement(unittest.TestCase):
     def setUp(self):
         # create test case data
-        self.test_element = ["H", "#2H", "Cu", "#12C", "C"]
-        self.name = ["H", "D", "Cu", "C", "C"]
-        self.picongpu_names = ["Hydrogen", "Deuterium", "Copper", "Carbon", "Carbon"]
```
PrometheusPi, line 21: This adds a feature (new element) to a bug fix

BrianMarre: Actually, no this just fixes the test supposed to find the bug to ensure that the error will not creep through again.

PrometheusPi: Thanks for clarifying

PrometheusPi (approved):

PR #5425, opened 2025-08-08
add test to picmi Gauss laser

PrometheusPi:
The added test checks that `propagation_direction` connects `centroid_position` and `focus_position`.


lib/python/picongpu/picmi/gaussian_laser.py
```diff
@@ -29,6 +29,17 @@ def scalarProduct(self, a: typing.List[float], b: typing.List[float]) -> float:
...
+        assert len(a) == len(b) == 3, "the cross product is only defined for 3d vectors"
+        return [a[1] * b[2] - a[2] * b[1], a[2] * b[0] - a[0] * b[2], a[0] * b[1] - a[1] * b[0]]
+
+    def difference(self, a: typing.List[float], b: typing.List[float]) -> typing.List[float]:
+        assert len(a) == len(b), "the difference between two vectors is only defined if they have the same length"
+        result = []
+        for i in range(len(a)):
+            result.append(a[i] - b[i])
+        return result
+
```
BrianMarre, line 42: implementing these here is okay for now, but since we decided to make scipy a dependency I would argue for just adding numpy as a dependency and use its cross product implementation.
And I know I implemented the scalar product above these lines, but at that time we still wanted to keep to native python libraries and I therefore avoided using numpy.

BrianMarre (changes requested):
CI is unhappy, probably the test failing is outdated.

chillenzer (approved):

PR #5426, opened 2025-08-08
add Lehe solver to picmi

PrometheusPi:
This PR adds the Lehe solver to PICMI. Since the cfl condition is the same as for Yee, it reuses the cfl check.


lib/python/picongpu/picmi/solver.py
```diff
@@ -22,6 +22,7 @@ class ElectromagneticSolver(picmistandard.PICMI_ElectromagneticSolver):
     def get_as_pypicongpu(self) -> solver.Solver:
         solver_by_method = {
             "Yee": solver.YeeSolver(),
+            "Lehe": solver.LeheSolver(),
         }
```
BrianMarre: that is an old style, @chillenzer built a self registration mechanism for the density profiles to avoid explicit book keeping such as this.

I am open to keep it as is in this PR for better reviewability, but you then should implement it in a follow up PR, soon.

*EDIT: Chillenzer pointed out the mechanism I meant is actually not powerful enough, we need to both get a list of names and get the PyPIConGPU type to translate to, I forgot the latter part. Therefore a future refactor is required to avoid the explicit book keeping.*

See here https://github.com/ComputationalRadiationPhysics/picongpu/blob/3984864197804347f7661f9ab1fbcae045bacdb7/lib/python/picongpu/pypicongpu/rendering/renderedobject.py#L261-L323

lib/python/picongpu/pypicongpu/solver.py
```diff
@@ -33,3 +33,18 @@ def _get_serialized(self) -> dict:
...
+    Lehe solver as defined by PIConGPU
+
+    note: has no parameters
+    """
+
+    def _get_serialized(self) -> dict:
+        # @todo + "<>" needs to be fixed later
+        return {
+            "name": "Lehe" + "<>",
+        }
```
BrianMarre: Okay, I know PICMI does not follow that convention always, but we should strive to always have only one class per file.
Makes getting an overview much easier

BrianMarre (changes requested):
Looks good, but lacks tests that the Lehe solver is actually useable. Please provide at least the basic test that the Lehe solver may be used in configuring a simulation.

lib/python/picongpu/picmi/solver.py
PrometheusPi: Okay - I would go for let's keep it for consistency reasons and as soon as this PR is merged I will update to the new style.
@BrianMarre Could you provide an example of this new style of self registration mechanism?

lib/python/picongpu/pypicongpu/solver.py
PrometheusPi: Fine with that

lib/python/picongpu/picmi/simulation.py
```diff
@@ -107,7 +107,11 @@ def __init__(
 
         # additional PICMI stuff checks, @todo move to picmistandard, Brian Marre, 2024
         ## throw if both cfl & delta_t are set
-        if self.solver is not None and "Yee" == self.solver.method and isinstance(self.solver.grid, Cartesian3DGrid):
+        if (
+            self.solver is not None
+            and self.solver.method in ["Yee", "Lehe"]
```
chillenzer, line 112: Introduce a named constant for this
```suggestion
            and self.solver.method in SOLVERS_WITH_YEELIKE_CFL_CONDITION
```
and also use it further below for the same list.

lib/python/picongpu/picmi/solver.py
```diff
@@ -30,8 +31,8 @@ def get_as_pypicongpu(self) -> solver.Solver:
         # todo: stencil order, cfl
         util.unsupported("stencil order", self.stencil_order)
         util.unsupported("field smoother", self.field_smoother)
-        if self.method != "Yee":
-            # for yee the cfl will be respected -- this behavior is coordinated
+        if self.method != "Yee" and self.method != "Lehe":
```
chillenzer, line 34: Another place to use your freshly named constant.

lib/python/picongpu/picmi/solver.py
chillenzer: That's not quite correct. I was targetting the other end of the PyPIConGPU layer (generating the rendering context). A `SelfRegisteringRenderedObject` does register its own name in [this list](https://github.com/ComputationalRadiationPhysics/picongpu/blob/9c4eab95563ff8daba9083c6a177a3660c89c4da/lib/python/picongpu/pypicongpu/rendering/renderedobject.py#L241) and one could potentially extend this to become a dictionary that contains the corresponding type. $^1$ In that case, one could theoretically use it for your purpose of instantiating the correct class from its self-registered name but that's beyond the scope of the current `SelfRegisteringRenderedObject`. It would also couple our low-level implementation detail of what name we use in the rendering context to the public interface of PICMI.

$^1$ I haven't checked this actually. There's a lot of details about instantiation order and stuff like that, so I wouldn't make any promises without a detailed study.

chillenzer (dismissed):
It's actually fine with me. If you want to touch anything about this, I'll leave this open for a bit but there's no urgently needed changes.

lib/python/picongpu/picmi/solver.py
BrianMarre: You first define a `BaseClass` that inherits from `SelfRegisteringRenderedObject` instead of `RenderedObject`.
This then defines a category and every `SubClass` that inherits from that `BaseClass` will be automatically added to that category. Each `Subclass` needs to define a name as an attribute called `_name` and the list of sub classes in the category will then be available as attribute `_names` in `BaseClass` and `SubClass`

```python
class Solver(SelfRegisteringRenderedObject):
  pass

class Yee(Solver):
  _name = "yee"

class Lehe(Solver):
  _name = "lehe"

Lehe()._names == ["lehe", "yee"]
```

chillenzer: Let's talk this over in the office tomorrow.

BrianMarre:
@chillenzer you are right, I forgot that we are using it for the translation here, well then we keep the manual book keeping for now.

PrometheusPi:
Since @BrianMarre wants some more features and I want to split "add Lehe feature" and "clean up file structure". I put this PR as draft.

PrometheusPi:
I will push two pull requests:
 - one with the field solver file splitting (refactoring)
 - one with adding the Lehe solver (addition)
and thus close this PR

PR #5429, opened 2025-08-12
separate picmi default and yee solver into two files

PrometheusPi:
This pull request is part of the split approach after closing #5426.
It separates the field solvers into different files (which will be later more useful when adding more field solvers such as Lehe)


BrianMarre (approved):
Should work

chillenzer (approved):

PR #5432, opened 2025-08-13
fix pypicongpu translation of ADK linear polarization

PrometheusPi:
Currently, the picmi ADK ionization model for linearly polarized lasers is translated wrongly by pypicongpu resulting in a compile error when building PIConGPU. 
The C++ name should change  `ADKLinearPol` -> `ADKLinPol`.
This also points to the fact that this is not tested.


BrianMarre (approved):
Good Catch

ikbuibui (approved):

PR #5433, opened 2025-08-13
change LWFA example laser to linear polarized

PrometheusPi:
This pull request just changes:
 - the laser polarization from circular to linear
 - the ADK model from circular to linear
 
 since the more common use case is a linear-polarized laser. 
 This should help avoid wrong setups based on this example.


ikbuibui (dismissed):

BrianMarre:
@PrometheusPi please also update the PICMI example to ensure it matches the LWFA example

chillenzer (changes requested):
Just to make it explcit. I think it would be a good idea to adjust the PICMI example as well.

PrometheusPi:
I updated the picmi lwfa example

chillenzer (approved):

PrometheusPi:
@ikbuibui I rebased against the current `dev` - hopefully the CI passes now

PR #5434, opened 2025-08-13
Cleanly define the co-moving origin and add moving window tests

ikbuibui:
- Redefined the co-moving origin to make it more simple and clear
- Added moving window tests

While doing this, I ran into a problem with including the param file in the dimensional tests. The following changes were also made
- fix the cmake to properly include the dimensional test param files
- use simDim instead of TEST_DIM in all dimensional tests
- use target_include_directories instead of include_directories


include/picongpu/simulation/control/MovingWindow.hpp
```diff
@@ -275,12 +277,26 @@ namespace picongpu
...
+            endSlidingOnStep = 0u;
+            computedConstants.reset();
+            firstCallEndSlideOnStep = true;
         }
 
         /**
          * Set step where the simulation stops the moving window
          *
          * @param step 0 means no sliding window, else sliding is enabled until step is reached.
+         * Negative numbers are the way to use this if if don't want the sliding to end (practically) -1 sets it to the
```
chillenzer:
```suggestion
         * Negative numbers are the way to use this if don't want the sliding to end (practically) -1 sets it to the
```

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
chillenzer, line 1: :tada: Tests!

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
```diff
@@ -0,0 +1,209 @@
...
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/generators/catch_generators.hpp>
+#include <picongpu/defines.hpp>
+#include <picongpu/simulation/control/MovingWindow.hpp>
+
+using namespace picongpu;
+
+//! Helper to setup the PMacc environment
+using TestFixture = pmacc::test::PMaccFixture<simDim>;
+static TestFixture pmaccFixture;
```
chillenzer: Not just this?
```suggestion
static pmacc::test::PMaccFixture<simDim> pmaccFixture;
```

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
```diff
@@ -0,0 +1,209 @@
...
+    MovingWindowTestFixture() : window(MovingWindow::getInstance())
+    {
+        window.resetForTesting();
+    }
+
+    ~MovingWindowTestFixture()
+    {
+        window.resetForTesting();
+    }
+};
```
chillenzer, line 50: That is a clear symptom of badly testable design: There's nothing about a moving window that needs to be a singleton. We should split the implementation of the moving window into a class providing the functionality that you can easily just instantiate and a form of storage that can happily still be a singleton. Might be beyond the scope of this PR though, I'm aware of that.

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+TEST_CASE("unit::MovingWindow_origin", "[movingWindow test]")
+{
+    MovingWindowTestFixture fixture;
+    auto& movingWindow = fixture.window;
+
+    pmaccFixture.initGrids(
+        pmacc::DataSpace<3>{500, 1000, 500}.shrink<simDim>(),
+        pmacc::DataSpace<3>{100, 100, 100}.shrink<simDim>(),
+        pmacc::DataSpace<3>{0, 0, 0}.shrink<simDim>());
```
chillenzer: Could we use all different numbers here? It's a bit of mental exercise, I'm aware, but it
1) exhibits bugs that might be hidden due to accidental equality
2) makes it easier to identify the problem in case there's a failure.

Of course, argument (1) works in reverse as well, so the best would be to use a Catch2 generator expression to generate tests with equal and different arguments and make the whole independent of the precise values.

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    auto const globalWindowSizeInMoveDirection = 900;
+
+    auto const c = static_cast<float_64>(sim.pic.getSpeedOfLight());
+    auto const dt = static_cast<float_64>(sim.pic.getDt());
+    // should both be 1 for correctness of the tests
+    assert(c == 1. && dt == 1.);
```
chillenzer:
```suggestion
    REQUIRE(c == 1.);
    REQURE(dt == 1.);
```

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    auto const cellSize = sim.pic.getCellSize();
+
+    constexpr auto moveDirection = MovingWindow::moveDirection;
+    auto const cellSizeInMoveDirection = static_cast<float_64>(cellSize[moveDirection]);
+    auto const cellsPerStep = 1. / cellSizeInMoveDirection;
```
chillenzer: I would probably factor out this whole initialisation into an own function. The reader quickly skimming this to find out about a particular behaviour doesn't need to bother how precisely `cellsPerStep` was computed. In fact, they don't even need to see `c` and `dt` at all.
```suggestion
    auto [globalWindowSizeInMoveDirection, ...] = init();
```

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    SECTION("Window disabled")
+    {
+        movingWindow.setEndSlideOnStep(0);
+        REQUIRE_FALSE(movingWindow.isEnabled());
+
+        auto origin = movingWindow.getMovingWindowOriginPositionCells(100);
```
chillenzer: Don't use magic numbers...
```suggestion
        auto origin = movingWindow.getMovingWindowOriginPositionCells(arbitrary_step);
```

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+        // Calculate firstMoveStep again for the given movePoint.
+        auto const virtualParticleInitialStartCell = static_cast<uint32_t>(
+            std::ceil(static_cast<float_64>(globalWindowSizeInMoveDirection) * (1.0 - movePoint)));
+
+        auto const wayToFirstMove
+            = static_cast<float_64>(globalWindowSizeInMoveDirection - virtualParticleInitialStartCell)
+              * cellSizeInMoveDirection;
+
+        auto const firstMoveStep = static_cast<int32_t>(std::ceil(wayToFirstMove / 1.)) - 1;
```
chillenzer: Did you know that you can nest `SECTION`s? Catch2 will generate a tree of sections and subsections and every leave of that tree will define a separate run in which all other branches are skipped. That would significantly reduce code duplication here. Roughly:

```C++
SECTION("Window is enabled") {
    float_64 const movePoint = 0.5;
    // ...
    
    SECTION("but not yet moving.") {
        // ...
        auto const origin = movingWindow.getMovingWindowOriginPositionCells(firstMoveStep - 1);
        // ...
    }
    
    SECTION("and moving.") {
        // ...
        auto const origin = movingWindow.getMovingWindowOriginPositionCells(firstMoveStep + 1);
        // ...
    }
}
```
will generate (functionally) the same code that you wrote with duplication.

Btw. do you have a test precisely at the `firstMoveStep`? What weird edge cases can we run into there?

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
chillenzer, line 1: On a more serious note, I think these tests are very valuable. Although from what I hear, the moving window is a particularly fiddly thing and I'd strongly expect that these fill up quickly with regression tests and edge cases. But nevertheless: Great to have a start! Thanks!

share/picongpu/unit/dimensional_tests/TWTSTight.cpp
```diff
@@ -44,13 +44,13 @@
...
-//! Helper to setup the PMacc environment
-using TestFixture = pmacc::test::PMaccFixture<TEST_DIM>;
-static TestFixture fixture;
-
 using namespace picongpu;
 using namespace pmacc;
 
+//! Helper to setup the PMacc environment
+using TestFixture = pmacc::test::PMaccFixture<simDim>;
+static TestFixture fixture;
```
chillenzer: Same as before, I presume we don't actually use that typedef anywhere else? But, that's cosmetics of course. Only act on these if you happen to find yourself pushing changes anyways.

include/picongpu/simulation/control/MovingWindow.hpp
```diff
@@ -381,6 +396,11 @@ namespace picongpu
 
         /**
          * Get the moving window origin position in cell, not discretized to the cell grid
+         * This position is the total origin (0,0,0) until we get to the first move step
+         * On the first move step the position is equal to the moving window move location plus the distance moved with
+         * c in one timestep. This ensures that this comoving particle is always inside the region of the moving window
+         * After this, the particle continues moving with c, until we reach the step where the window stops
+         * moving, Then, the particle stops where ever it is (maybe in the middle of a cell)
```
chillenzer, line 403: The function you're commenting on here explicitly has "Cells" in the name (`getMovingWindowOriginPositionCells`). I would probably be gullible enough to be tricked into believing that this would return a vector of integers. Should we adjust this? One could even consider generalising this function to take the same (compile-time) arguments as `DomainInfo` does (`CELL`/`SUB_CELL` precision and maybe even the others). Of course, we should only do that if we see a clear usecase for this.

include/picongpu/simulation/control/MovingWindow.hpp
```diff
@@ -394,23 +414,24 @@ namespace picongpu
...
-            {
-                auto cellsPerStep = constants.deltaWayPerStep / constants.cellSizeInMoveDirection;
+            auto cellsPerStep = constants.deltaWayPerStep / constants.cellSizeInMoveDirection;
 
-                SubGrid<simDim> const& subGrid = Environment<simDim>::get().SubGrid();
-                float_64 const virtualParticleCellsPassed = cellsPerStep * static_cast<float_64>(currentStep);
-                float_64 const moveLocationInCells
-                    = constants.moveRatio
-                      * (subGrid.getGlobalDomain().size[moveDirection] - constants.gpuNumberOfCellsInMoveDirection);
+            auto firstStep = constants.firstMoveStep > 0 ? constants.firstMoveStep : 0;
```
chillenzer, line 428: Why would this ever be necessary? Can this be negative?

chillenzer (changes requested):
It's not exactly blocking. Just forcing some discussion on this before you get the approval and merge.

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
ikbuibui: I have combined the sections as suggested.
I have only sort of tested precisely at `firstMoveStep`. I have the test where the window starts moving at step zero, which is also the `firstMoveStep`. 
I had to be careful how to define the co-moving particle to deal with some edge cases. 
By being careful I mean that I needed to realize that the moving window can possibly already move on `firstMoveStep`, depending on the cell size and the move point. 
This is why the co-moving particle has an offset of `c*dt` to make sure for all cases that the co-moving particle is inside the moving window, since if the first move happens at first move step the moving window location would jump by one cell and this might leave the co-moving particle behind. Now the co-moving particle is always inside the first cell of the moving window, with the effect that particle positions relative to this origin might be negative..
I will add a test for this actually, making sure the particle is always in the first cell.

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
ikbuibui: Done

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
ikbuibui: Done.

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
ikbuibui: Done,

share/picongpu/unit/dimensional_tests/TWTSTight.cpp
ikbuibui: Done

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
ikbuibui: I see your point but am not particularly enthusiastic about this, since I think if someone wants to start reasoning about the logic of the tests they might care if these values are calculated correctly. Also a long structured binding isn't exactly pretty either. I do agree `c` and `dt` can be completely hidden, but I am for keeping it as it is to not add indirection. 
So unless you really want me to change this, I am going to leave it

include/picongpu/simulation/control/MovingWindow.hpp
ikbuibui: Yes, it is equal to `-1` if `movePoint` is set to `0`.

ikbuibui (commented):
Pushed a new commit with more tests, and applied most reviewer comments

include/picongpu/simulation/control/MovingWindow.hpp
ikbuibui: Fixed

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
ikbuibui: I have changed the number to not be identical anymore, however I decided not to put in the effort to test the case where they are equal

PrometheusPi:
Started from a discussion weeks ago with @ikbuibui, I set up a test of the co-moving position using the binning plugin.
Below you see the result for the current dev (minor jumps) vs. this PR, no jumps. For reference, the y-cell-size is 0.04µm, which is also the binning width of the binning plugin.

grafik

include/picongpu/simulation/control/MovingWindow.hpp
chillenzer: Offline discussion: The name is correct in that it refers to the units returned. The number will be a (floating point, so potentially fractional) multiple of the cell size.

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
```diff
@@ -0,0 +1,247 @@
...
+        // Manually calculate when the window should start moving to verify pre-movement phase.
+        auto const virtualParticleInitialStartCell = static_cast<uint32_t>(
+            std::ceil(static_cast<float_64>(globalWindowSizeInMoveDirection) * (1.0 - movePoint)));
+
+        // Calculate firstMoveStep again for the given movePoint.
+        auto const wayToFirstMove
+            = static_cast<float_64>(globalWindowSizeInMoveDirection - virtualParticleInitialStartCell)
+              * cellSizeInMoveDirection;
+
+        auto const firstMoveStep = static_cast<int32_t>(std::ceil(wayToFirstMove / 1.)) - 1;
```
chillenzer, line 106: Now I'm curious what horrible debugging resulted in `std::ceil(wayToFirstMove / 1.)`? Or could this possibly be a unnamed magic number?

chillenzer (approved):

share/picongpu/unit/dimensional_tests/MovingWindow.cpp
ikbuibui: Oh actually this is only to find the integral time step where the window starts moving. It is just `ceil(time_taken) - 1`. It isn't `floor(time_taken)` because of implementation details of the moving window which instantly starts moving.
Maybe the `1.` is the source of confusion, which is `c*dt`

chillenzer: Which is a magic constant!

chillenzer: Why not `c*dt`?

BrianMarre: Strictly speaking PIConGPU does not guarantee `c*dt == 1` that is just a convention in our typical param files

ikbuibui: Yes, which is why the tests `REQUIRE` that both are actually one. I did start with `c*dt` everywhere, but IMO it started to become hard to parse through, so I opted for this.

chillenzer: Nope, I still think that a named constant is the way to go. Just `c * dt` is not a good abstraction. I'd rather call it `characteristic_length` or `shift_length` or something... (those are admittedly not great but you get the idea).

ikbuibui: Fair, I'll refactor the tests soon

PR #5435, opened 2025-08-14
remove exponential ramp laser from LWFA example

PrometheusPi:
This pull request removes the exponential ramp laser from the LWFA example because:
 - it is not relevant for LWFA
 - and @pordyna also stated that one should use a different approach anyway
  
If one wants to keep this laser as a reference in an example, I could add it to FoilLCT.


chillenzer:
@pordyna will add to this PR his new Gauss with temporal envelope implementation. DO NOT MERGE BEFORE!

PrometheusPi:
Discussion with @pordyna:
He will replace the exponential ramp laser with his improved version:
 - this will be a follow-up PR
 - thus we do not need to remove the exponential ramp laser implantation here

pordyna:
This PR is obsolete, this commit is now included in #5437 @PrometheusPi can be closed

PrometheusPi:
(We seem to have a CI problem right now)

PR #5437, opened 2025-08-14
Free temporal envelope for the GaussianPulse profile, replaces ExpRampWithPrepulse

pordyna:
* Adds a user free functor for defining GaussianPulse temporal envelope (the transversal envelope remains Gaussian). 
* Also, removes ExpRampWithPrepulse from PIConGPU that could be used as an envelope for the GaussianPulse already before. Such complicated profiles can be now defined with the free functor. 
* Adds GaussianPulse laser profile to FoilLCT example together with a free functor example (that just reproduces the standard GaussianPulse). 

I tested this feature already in production, it works as expected. I believe @paschk31 was also already using it. 
@steindev @PrometheusPi @psychocoderHPC This is ready for review. The draft state is just so that I can remove the last commit before merging (it is there just so that we can test the examples with CI without running all possible tests).


PrometheusPi:
Something failed in the CI but there was no trace, thus I restarted the failed CI jobs.

chillenzer:
@pordyna What's the status of this? Any movement?

PrometheusPi:
@pordyna Is this reviewable and the draft mode a left-over, or should we wait for more changes?

pordyna:
@PrometheusPi you can have a look already, but I still need to move the extra test out of the CI (was just temporary). Also, I will provide (not in picongpu code but will link it here) an example on how to reproduce the `ExpRampWithPrepulse`, and test it against the old implementation.  Should be done with it today.

pordyna:
@finnolec  Here is a gist on how to create the old `ExpRampWithPrepulse` temporal envelope using the Free Functor and `sympy`. https://gist.github.com/pordyna/3425d2a053bb052833717a8719ac78f3

chillenzer:
I love how you immediately ripped that feature from deep inside the guts of PyPIConGPU out of context built something useful out of it!

pordyna:
Comparing Electric field values between the simulation, that uses a `sympy` generated input, and the `sympy` expression evaluated in python. Orange: `sympy` E-field envelope (corrected for propagation distance), Blue: $$\mathrm{abs}(E_x)$$ at the focus position in the middle of the simulation box. All in units of a0. The input profile is the same as in the gist mentioned in my last comment. Fits quite well, I would say.

compare_sympy

pordyna:
@chillenzer @psychocoderHPC ready for review

PrometheusPi (dismissed):
looks good to me - I would like to test the regular Gauss case before merging

include/picongpu/fields/incidentField/profiles/GaussianPulse.def
```diff
@@ -119,6 +119,27 @@ namespace picongpu::fields::incidentField::profiles
     template<typename T_Param>
     struct GaussianPulseEnvelope;
 
+    /** User temporal envelope Functor
+     *
+     * This functor describes the E field evelope (sqrt(I))
```
steindev: typo 'envelope'.
Please also add 'temporal' again, just to make this clear.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
```diff
@@ -387,6 +387,22 @@ namespace picongpu::fields::incidentField
             }
         };
 
+        template<typename T_Func>
+        struct FreeEnvelope
+        {
+            static constexpr float_X TIME_SHIFT = 0.0_X;
```
steindev, line 393: What is this doing here? And is it shadowing the `float_64 TIME_DELAY` from the `BaseParam`?

share/picongpu/examples/FoilLCT/include/picongpu/param/incidentField.param
```diff
@@ -214,90 +213,69 @@ namespace picongpu
 
             using PlaneWaveBeam = profiles::PlaneWave<PlaneWaveParams>;
 
-            //! Laser Profile is taking plane wave param configuration and is overwriting only a few settings
-            struct ExpRampWithPrepulseParam : public PlaneWaveParams
+            //! Extends plane wave setup with a gassian pulse paramters
```
steindev: Typo 'Gaussian' & 'parameters'.
Why is the word 'parameter' written here?

share/picongpu/examples/FoilLCT/include/picongpu/param/incidentField.param
```diff
@@ -214,90 +213,69 @@ namespace picongpu
...
+            struct GaussianPulseParam : public PlaneWaveParams
```
steindev, line 217: Why does this Gaussian param struct inherit from a plane wave param and does not use a normal Gaussian param struct?

share/picongpu/examples/FoilLCT/include/picongpu/param/incidentField.param
```diff
@@ -214,90 +213,69 @@ namespace picongpu
...
+            using GaussianBeam = profiles::GaussianPulse<GaussianPulseParam>;
 
-            namespace profiles
+            /** Example of a free temporal envelope functor
+             *
+             * Describes the E field evelope (sqrt(I))
+             * 1.0 corresponds to the a_0 set in the gaussian param struct
+             * This functor reproduces the standard Gaussian envelope
+             */
+            struct FreeEnvelopeExample
```
steindev: The typical convention for the name would be `FreeEnvelopeFunctor`, wouldn't it?

steindev (commented):
So far so good. Only minor comments. but it is a pity, that the Exponential Ramp Profile or `GaussianBeamWithContrastCurve` is gone as an example from the source code.

Would you add your gist, which showcases the construction of a complex profile, to readthedocs? I think this would aid the loss :-)

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
pordyna: Oh, this is part of the interface for the temporal envelopes. I think, it was there because `ExpRampWithPrepulse` defined one. I guess now we could get rid of it. But this is also not exposed to user. So, this is a coordinate transform between pic and some internal profile time.

For now, it is required here:
https://github.com/ComputationalRadiationPhysics/picongpu/blob/0d5821fe732c12942cc249476955a4f9425821f1/include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp#L226

share/picongpu/examples/FoilLCT/include/picongpu/param/incidentField.param
pordyna: `PlaneWaveParam` is just `BaseParam` in that context. Just reusing the parameters.

share/picongpu/examples/FoilLCT/include/picongpu/param/incidentField.param
pordyna: >  Why is the word 'parameter' written here?

Since it is just the param class. What would be your suggestion?

Fixed the typos

share/picongpu/examples/FoilLCT/include/picongpu/param/incidentField.param
pordyna: corrected

pordyna:
@steindev I addressed your comments.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
steindev: In the case of not exposed and not used (since always zero)we should get rid of it.

share/picongpu/examples/FoilLCT/include/picongpu/param/incidentField.param
steindev: Maybe: Extend PlaneWave setup to Gaussian pulse

include/picongpu/fields/incidentField/profiles/GaussianPulse.def
steindev: Maybe like this: 'This functor describes the longitudinal E-field envelope (sqrt(I))'

steindev:
@pordyna I responded to your two comments. Also, it seems you forgot to address the first of the initial comments.

Please squash and rebase before the next push. Then we can directly merge. Thanks!

include/picongpu/fields/incidentField/profiles/GaussianPulse.def
pordyna: lol I'm so sure I commented on this and changed it? This is some git magic haha. I will fix it again.

include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp
pordyna: Offline discussion with @steindev: It is used here https://github.com/ComputationalRadiationPhysics/picongpu/blob/0d5821fe732c12942cc249476955a4f9425821f1/include/picongpu/fields/incidentField/profiles/GaussianPulse.hpp#L375 and it affects not only the envelope but also the phase. We decided that we should keep it.

pordyna:
Squashed commits and rebased on dev. Also addressed remaining @steindev comments (only affects docstrings). Ready to merge from my side. @PrometheusPi do you still want to run your test?

PrometheusPi:
I submitted the test today to hemera. Will check them as soon as they finished.

PrometheusPi:
There is no deviation in field energy components between the current dev and this PR. 
Can be merged from my side.

chillenzer (approved):

PR #5441, opened 2025-08-15
add walltime option to picmi / pypicongpu

### LINKED ISSUE #5438: pypicongpu does not support walltime settings
Currently, the walltime is fixed to 1 hour when using PICMI/pypicongpu. 
https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/share/picongpu/pypicongpu/template/etc/picongpu/N.cfg.mustache#L50

PrometheusPi:
This pull request closes #5438  by adding the option to set the simulation walltime via picmi and pypicongpu.

### DIFF
--- a/lib/python/picongpu/pypicongpu/simulation.py
+++ b/lib/python/picongpu/pypicongpu/simulation.py
@@ -73,7 +73,7 @@ class Simulation(RenderedObject):
     moving_window = util.build_typesafe_property(typing.Optional[MovingWindow])
     """used moving Window, set to None to disable"""
 
-    walltime = util.build_typesafe_property(typing.Optional[datetime.timedelta])
+    walltime = util.build_typesafe_property(typing.Optional[Walltime])
     """time limit of the simulation run"""
 
     plugins = util.build_typesafe_property(typing.Optional[list[Plugin] | typing.Literal["auto"]])
@@ -144,7 +144,7 @@ class Simulation(RenderedObject):
             serialized["moving_window"] = None
 
         if self.walltime is not None:
-            serialized["walltime"] = Walltime(walltime=self.walltime).get_rendering_context()
+            serialized["walltime"] = self.walltime.get_rendering_context()
         else:
             serialized["walltime"] = Walltime(walltime=datetime.timedelta(hours=1)).get_rendering_context()
 
--- a/lib/python/picongpu/pypicongpu/simulation.py
+++ b/lib/python/picongpu/pypicongpu/simulation.py
@@ -73,7 +73,7 @@ class Simulation(RenderedObject):
     moving_window = util.build_typesafe_property(typing.Optional[MovingWindow])
     """used moving Window, set to None to disable"""
 
-    walltime = util.build_typesafe_property(typing.Optional[Walltime])
+    walltime = util.build_typesafe_property(typing.Optional[datetime.timedelta])
     """time limit of the simulation run"""
 
     plugins = util.build_typesafe_property(typing.Optional[list[Plugin] | typing.Literal["auto"]])
@@ -144,7 +144,7 @@ class Simulation(RenderedObject):
             serialized["moving_window"] = None
 
         if self.walltime is not None:
-            serialized["walltime"] = self.walltime.get_rendering_context()
+            serialized["walltime"] = Walltime(walltime=self.walltime).get_rendering_context()
         else:
             serialized["walltime"] = Walltime(walltime=datetime.timedelta(hours=1)).get_rendering_context()
 
--- a/lib/python/picongpu/pypicongpu/simulation.py
+++ b/lib/python/picongpu/pypicongpu/simulation.py
@@ -16,10 +16,12 @@ from .rendering import RenderedObject
 from .customuserinput import InterfaceCustomUserInput
 from .output.plugin import Plugin
 from .output.timestepspec import TimeStepSpec
+from .walltime import Walltime
 
 import typing
 import typeguard
 import logging
+import datetime
 
 
 @typeguard.typechecked
@@ -70,6 +72,9 @@ class Simulation(RenderedObject):
 
     moving_window = util.build_typesafe_property(typing.Optional[MovingWindow])
     """used moving Window, set to None to disable"""
+
+    walltime = util.build_typesafe_property(typing.Optional[Walltime])
+    """time limit of the simulation run"""
 
     plugins = util.build_typesafe_property(typing.Optional[list[Plugin] | typing.Literal["auto"]])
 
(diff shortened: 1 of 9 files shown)


PrometheusPi:
I currently do not like this ugly dict in dict solution (see last commit) @BrianMarre and @chillenzer are there better solutions to just return a string and not a dict in `walltime.Walltime`? (I tried just a string and ran into schema issues)

PrometheusPi:
Currently, there are small commits that allows to follow the development. Thius should be squashed before going mainline.

BrianMarre:
Regarding the dict in dict issue raised by PrometheusPi, I would keep this structure, just rename it to `SimulationResources`.
In the long term I want to move everything related to simulation resources into one object. That is:
- walltime (new feature of this PR)
- number of devices per dimension (currently stored in the grid object)
- cell distribution on the GPUs (currently stored in the grid object)

This has the advantage of explicitly decoupling the physics description of the simulation (not cluster/hardware dependent) from the resource description(cluster/hard ware dependent).

EDIT: I was mistaken we already support configuring the grid distribution from PICMI and it currently is a member of the grid definition

PrometheusPi:
Since I have to fix some errors in this PR, I will do the renaming with it if this is fine with everybody.

lib/python/picongpu/pypicongpu/walltime.py
```diff
@@ -0,0 +1,30 @@
...
+        if self.walltime.total_seconds() <= 0.0:
+            raise ValueError("walltime must be > 0.")
+
+    def _get_serialized(self) -> dict:
+        return {
+            "walltime": "{:d}:{:02d}:{:02d}".format(
+                int(self.walltime.total_seconds() // 3600),  # hours
+                int(self.walltime.total_seconds() % 3600 // 60),  # minutes
+                int(self.walltime.total_seconds() % 3600),  # seconds
+            )
```
chillenzer: How to do this according to [Python docs](https://docs.python.org/3/library/datetime.html#datetime.timedelta.total_seconds):
```suggestion
            "walltime": "{:d}:{:02d}:{:02d}".format(
                (r := divmod(self.walltime, HOUR))[0],  # hours
                (r := divmod(r[1], MINUTE))[0],  # minutes
                (r[1] // SECOND),  # seconds
            )
```
with global constants
```python
HOUR = timedelta(hours=1)
MINUTE = timedelta(minutes=1)
SECOND = timedelta(seconds=1)
```
PS: Your seconds calculation has a bug, so even if you don't fancy that other method, please fix:
```suggestion
            "walltime": "{:d}:{:02d}:{:02d}".format(
                int(self.walltime.total_seconds() // 3600),  # hours
                int(self.walltime.total_seconds() % 3600 // 60),  # minutes
                int(self.walltime.total_seconds() % 60),  # seconds
            )
```

chillenzer (changes requested):

lib/python/picongpu/pypicongpu/walltime.py
BrianMarre: Good Catch!

share/picongpu/pypicongpu/template/etc/picongpu/N.cfg.mustache
```diff
@@ -47,7 +47,9 @@ set -u
 ## Section: Required Variables ##
 #################################
 
-TBG_wallTime="1:00:00"
+{{#walltime}}
+TBG_wallTime="{{{walltime}}}"
+{{/walltime}}
```
chillenzer, line 52: What about the case when this check fails?

chillenzer (changes requested):

share/picongpu/pypicongpu/template/etc/picongpu/N.cfg.mustache
PrometheusPi: What do you mean with that?

chillenzer: `{{#walltime}}...{{/walltime}}` means "render this block if `walltime` evaluates to true" (for whatever precise meaning that has in mustache). If you use that, there must be some thought about what's the other case. Is it fine to just have nothing rendered here if `walltime` does not evaluate to true (say, it's `None`)? Skimming through your code it looks like that's not a case easily produce-able by proper use. But if that's not allowed simply not having a check here would be best. Otherwise, I'd suggest to put a default here via `{{^walltime}}TBG_walltime=1:00:00{{/walltime}}` for example (meaning "render this block if `walltime` evaluates to false).

chillenzer: PS: Just for completion, in the context of array data `{{#walltime}}` can also mean "for each element in `walltime`, of which the `if... then...` semantics is a reasonable generalisation for non-array-like content.

PrometheusPi: Default value is a good idea - will do

BrianMarre: Well no, by design we don't want logic in our templates, they are supposed to be pure search and replace placeholders.
Instead defaults are supposed to set on the PICMI level, on PyPIConGPU level all settings need to set explicitly and there setting enforced.
In this particular case, if no walltime is set in the simulation tbg will fall back to the scheduler default for the queue set in the use profile, therefore no default is required.

lib/python/picongpu/pypicongpu/simulation.py
```diff
@@ -71,6 +73,9 @@ class Simulation(RenderedObject):
     moving_window = util.build_typesafe_property(typing.Optional[MovingWindow])
     """used moving Window, set to None to disable"""
 
+    walltime = util.build_typesafe_property(typing.Optional[datetime.timedelta])
+    """time limit of the simulation run"""
```
BrianMarre:
```suggestion
    walltime = util.build_typesafe_property(typing.Optional[Walltime])
    """time limit of the simulation run"""
```

lib/python/picongpu/pypicongpu/simulation.py
```diff
@@ -138,6 +143,11 @@ def _get_serialized(self) -> dict:
         else:
             serialized["moving_window"] = None
 
+        if self.walltime is not None:
+            serialized["walltime"] = Walltime(walltime=self.walltime).get_rendering_context()
```
BrianMarre:
```suggestion
            serialized["walltime"] =self.walltime.get_rendering_context()
```

lib/python/picongpu/pypicongpu/simulation.py
BrianMarre: since we have a dedicated, `Walltime` class we should also use it

PrometheusPi:
@chillenzer @BrianMarre ready for another review - CI passes

chillenzer (approved):

Issue #5443, opened 2025-08-18
hemera defq simulation crashes with 'sched_setaffinity: Invalid argument'


finnolec:
I have pulled #5442 onto the current dev version of picongpu. 

```
4f7ad07d1 - PrometheusPi, 2025-08-18 10:42:36 +0200 : fix defq.tpl

af2cd1387 - René Widera, 2025-08-17 08:47:11 +0200 : Merge pull request #5440 from PrometheusPi/fix_exitCode_pic-create

add missing exit codes pic-create
```
and tried to compile and run the LWFA example
```bash
source ~/src/picongpu/etc/picongpu/hemera-hzdr/defq_picongpu.profile.example
pic-create LaserWakefield lwfa
getDevice
cd lwfa
source ~/src/picongpu/etc/picongpu/hemera-hzdr/defq_picongpu.profile.example
pic-build -j 16
tbg -s -t -c etc/picongpu/1.cfg /path/to/output
``` 
, and it crashes with following output (the PR is already fixing the tpl issues though and should be merged):

```
> cat stdout
Running program...
PID = 26136

------------------------------
- - - - - job epilog - - - - -
------------------------------
Job ID: 14413826
was running on nodes: csk067
by user: carste06
in partition: defq
using account: default
number of CPUs used: 40
number of nodes requested: 1
------------------------------
walltime reqd: 01:00:00
walltime used: 00:00:11
------------------------------
Mon Aug 18 14:00:25 CEST 2025
------------------------------
```
```
> cat stderr
Unloading module pngwriter/0.7.0
Unloading module libpng/1.6.39
Unloading module cmake/3.26.1
Unloading module openpmd/0.15.2-csk
Unloading module adios2/2.9.2-csk
Unloading module c-blosc/1.21.4
Unloading module hdf5-parallel/1.14.0-omp415
Unloading module zlib/1.2.11
Unloading module boost/1.82.0
Unloading module python/3.10.4
Unloading module openmpi/4.1.5-ucx
Unloading module libfabric/1.17.0-co79-csk
Unloading module ucx/1.14.0
Unloading module gcc/12.2.0
Unloading module git/2.37.1
Loading module git/2.37.1
Loading module gcc/12.2.0
Loading module ucx/1.14.0
Loading module libfabric/1.17.0-co79-csk
Loading module openmpi/4.1.5-ucx
Loading module python/3.10.4
Loading module boost/1.82.0
Loading module zlib/1.2.11
[...]

PR #5444, opened 2025-08-19
allow restart/remove outdated check in tpl w/o cuda_memtest

PrometheusPi:
When investigating #5443, I ran into the issue that rerunning / restarting of a PIConGPU simulation on hemera's `defq` never ran PIConGPU a second time. This bug was caused by an outdated check for `cuda_memtest`, that prevented `picongou` to run, if the GPUs showed any errors. However, on CPUs, we do not perform this test, but the guard was kept, thus checking if the linking to `stdout` was successful, which it never is an a second run, because the link already exists on the second attempt.  

I will check, whether other `*.tpls` are affected by this as well.

### DIFF
--- a/etc/picongpu/davinci-rice/picongpu.tpl
+++ b/etc/picongpu/davinci-rice/picongpu.tpl
@@ -45,7 +45,7 @@
 # settings that can be controlled by environment variables before submit
 .TBG_mailSettings=${MY_MAILNOTIFY:-"n"}
 .TBG_mailAddress=${MY_MAIL:-"someone@example.com"}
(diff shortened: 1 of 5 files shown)


PrometheusPi:
As suspected, more `*.tpl`s were affected. 
I also found a simply code-breaking typo that I would like to include here (as separate commit).

chillenzer (approved):
Looks okay to me.

chillenzer:
Please, re-push so the CI can run.

PrometheusPi:
@chillenzer PR is re-pushed

PR #5447, opened 2025-08-21
Binning plugin dump improvements

### LINKED ISSUE #5449: Restarting simulation requires binning checkpoints although not actually required
Hello, I have created a checkpoint with a simulation where I did not use the binning plugin while running the simulation. Now I activated the binning plugin where I do not need any time-integration or similar, so the binning plugin should not require a checkpoint to restart but according to the stderr it does:

```
The following modules were not unloaded:
  (Use "module --force purge" to unload all):

  1) Stages/2025
Unhandled exception of type 'N7openPMD5error9ReadErrorE' with message 'Read Error in frontend
Object type:	File
Error type:	Inaccessible
Further description:	Supplied directory is not valid: checkpoints/binningOpenPMD/', terminating
Fatal error in internal_Barrier: Other MPI error, error stack:
```

Quick fix is to comment out this part in the `BinningDispatcher.x.cpp`, but I guess this is no long-term solution :D 
```cpp
void restart(uint32_t restartStep, std::string const restartDirectory) override
{
    /* restart from a checkpoint here
     * will be called only once per simulation and before notify() */
    for(auto&& binner : binnerVector)
    {
        //binner->restart(restartStep, restartDirectory);
    }
}
```

ikbuibui:
- **Remove time averaging**. **WARNING FOR USERS** - Default behavior has changed and setting removed. If you now need time averaging you can do so by dividing the binned values by the openPMD attribute `reduceCounter`.  
- Add openPMD attribute `reduceCounter`  to all output files
- Dump partial buffers at the end of run. Partial means that dump period was not reached before the simulation ended.
- If restarting from a checkpoint which was created without the binning plugin enabled, we now warn the user and then start a new binning. Closes #5449


include/picongpu/plugins/binning/binners/Binner.hpp
```diff
@@ -252,27 +253,38 @@ namespace picongpu
...
+                        // @todo divide histBuffer by gc.getGlobalSize and call from all ranks
+
+                        // transfer restored data to device so that it is not overwritten
+                        this->histBuffer->hostToDevice();
+                    }
+                    catch(...)
+                    {
+                        std::cout << "Warning loading binning data from checkpoint. Will start a new binning, with "
+                                     "reduce counter starting from 0 at the first notify after restart. "
+                                  << std::endl;
```
chillenzer: I would have to look up what it's called exactly but there is warning functionality in PMAcc. We should use that here.

include/picongpu/plugins/binning/binners/Binner.hpp
```diff
@@ -81,7 +82,7 @@ namespace picongpu
                     binningData.axisExtentsND.productOfComponents());
                 this->histBuffer->getDeviceBuffer().setValue(
                     pmacc::math::operation::traits::
-                        NeutralElement<typename TBinningData::AccumulationOp, TDepositedQuantity>::value);
+                        NeutralElement<typename TBinningData::ReductionOp, TDepositedQuantity>::value);
```
chillenzer: Maybe you want to have `using ReductionOp = TBinningData::ReductionOp;`? That's used quite a bit in this file.

include/picongpu/plugins/binning/binners/Binner.hpp
```diff
@@ -282,6 +294,64 @@ namespace picongpu
...
+                    auto bufferExtent = this->histBuffer->getHostBuffer().capacityND();
+
+                    // Do time Averaging
+                    if(binningData.timeAveraging)
+                    {
+                        TDepositedQuantity factor = 1.0 / static_cast<double>(reduceCounter);
+
+                        constexpr uint32_t blockSize = 256u;
+                        // @todo is + blocksize - 1/ blocksize a better ceil for ints
+                        auto gridSize = (bufferExtent[0] + blockSize - 1) / blockSize;
```
chillenzer: There's a PMAcc function for doing this. `divceil` or something.

include/picongpu/plugins/binning/binners/Binner.hpp
```diff
@@ -282,6 +294,64 @@ namespace picongpu
...
+                                std::string("end_of_run_") + binningData.binnerOutputName,
+                                binningData.openPMDInfix,
+                                binningData.openPMDExtension,
+                                binningData.openPMDJsonCfg},
+                            std::move(hReducedBuffer),
+                            binningData,
+                            Environment<>::get().SimulationDescription().getRunSteps() - 1,
+                            reduceCounter);
+                    }
+                }
```
chillenzer: Extract method, so that you can combine this with the time averaging and control flow from the regular case.

chillenzer (changes requested):
Sorry, I might be a bit late with this discussion but does time averaging make sense for anything other than addition? (My answer's: Surely not.) I'd probably prefer to express this by introducing an operation "mean" next to "add", "min", "max", etc. That is a much clearer expression of intent and reduces the risk of the user doing something unintended.

PS: This could, of course, happen in the front-end alone, so you can keep most of the code that you've got already.

ikbuibui:
I really like your suggestion to use a ``mean`` operation for time averaging. However, I am thinking of removing the time averaging functionality altogether. Let's discuss this in person tomorrow. If we keep it in, I will refactor things to define a ``mean`` op.

include/picongpu/plugins/binning/binners/Binner.hpp
ikbuibui: Oh yeah, there is ``alpaka::core::divCeil`` which I will use

include/picongpu/plugins/binning/binners/Binner.hpp
ikbuibui: Switching to use `log<picLog::SIMULATION_STATE>`

include/picongpu/plugins/binning/binners/Binner.hpp
ikbuibui: Removed time averaging and moved the reduction to a function

include/picongpu/plugins/binning/binners/Binner.hpp
ikbuibui: I removed time averaging, so this is not relevant anymore

include/picongpu/plugins/binning/binners/Binner.hpp
ikbuibui: The problem here is that our default log level is `PHYSICS=1` and thus for most setups using this logging will not be seen by the user. There is a log level `CRITICAL=4` but this is also disabled by default. 
I think we should rename log level `0` to `ALWAYS` or `MANDATORY` from `NOTHING` and use this for cases where logging should always be done.

chillenzer (changes requested):
Silently includes renaming loglevel NOTHING to MANDATORY.

chillenzer:
While we're at it, should we find a better name for `dump_period`? It is rather an `num_reduction_steps` or something like that...

ikbuibui:
@chillenzer I rebased to remove the `NOTHING` to `MANDATORY` renaming for logging. I am going to stick to using `std::cout` for now until we decide on a way forward for such logging situations #5460. I am not renaming `dump_period` currently as it is a user facing API change, and I don't see it as a something really required. If needed we can do it in another PR.

chillenzer (approved):

PR #5448, opened 2025-08-21
Update moving window tests with variable for distance traveled by light

ikbuibui:
Use variable for `c * dt` instead of using the value `1`.


chillenzer (approved):
:heart:

PR #5450, opened 2025-08-22
Add a table in the binning documentation describing config options 

ikbuibui:
Adds a table to the binning documentation giving an overview of the configuration options available, and their default values.

### DIFF
--- a/docs/source/usage/plugins/binningPlugin.rst
+++ b/docs/source/usage/plugins/binningPlugin.rst
@@ -318,6 +318,42 @@ OpenPMD JSON Configuration
 --------------------------
 Users can set a json configuration string for the OpenPMD output format using the ``setOpenPMDJsonCfg`` setter method.
 
+Configuration Options Summary
+-----------------------------
+The following table summarizes the configuration options available for the binner object returned by `addParticleBinner()` or `addFieldBinner()`. These methods allow for fine-tuning the behavior of the binning plugin.
+
+.. list-table:: Binning Configuration Options
+   :widths: 30 50 20
+   :header-rows: 1
+
+   * - Option
+     - Description
+     - Default Value
+   * - ``setNotifyPeriod``
+     - Sets the periodicity of the output. Follows the period syntax defined :ref:`here <usage/plugins:period syntax>`.
+     - ``"1"`` (every time step)
+   * - ``setDumpPeriod``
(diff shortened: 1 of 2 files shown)


PrometheusPi:
@ikbuibui since the PR is still in draft mode: should we already start reviewing?

ikbuibui:
@PrometheusPi Yes, you can start reviewing. Only the last two commits belong to this PR.  It is rebased on top of #5447 because that PR introduces a renaming from accumulate to reduce, which I want to build upon.

docs/source/usage/plugins/binningPlugin.rst
```diff
@@ -10,6 +10,7 @@ Users can
 	- Choose which species are used for particle binning
 	- Choose which fields are used for field binning
 	- Choose how frequently they want the binning to be executed
+	- Choose how many times the binned quantity should be combined over time before dumping
```
chillenzer, line 13:
```suggestion
	- Choose how many times the binned quantity should be recorded before dumping
```

chillenzer (approved):
I think it's fine like this. But as you mentioned it: If you want to have the table further at the top, I'd suggest to add a quickstart section with a minimal example and all configuration option tables, etc.

```C++
// `myDepositionFunctor` is analogous
auto myAxisFunctor = [](auto const& worker, auto const& domainInfo, auto const& particle) {
        return doCalculation(particle);
    };

// define more axes as needed
auto myAxis = axis::createLinear(
        axis::AxisSplitting(axis::Range(min, max), numStep),
        createFunctorDescription(myAxisFunctor, "myDescription", myUnits)
    );
    
binningCreator.addParticleBinner(
        "myName",
        createAxisTuple(myAxis),
        createSpeciesTuple(mySpecies),
        createFunctorDescription(myDepositionFunctor, "myDepositionName", myDepositionUnits)
   ).setNotifyPeriod("0:100:10");
```

For quick reference the following options are available...

chillenzer:
Offline: We agreed that the quickstart section would be a good idea. But having this table in is valuable also in the current position, so we merge and add the quickstart section in another PR.

docs/source/usage/plugins/binningPlugin.rst
ikbuibui: Concise. I like it.

PR #5451, opened 2025-08-22
activate warning log output for all picmi examples

PrometheusPi:
When using PICMI we now throw a bunch of warnings for bad setups. But these warnings are never printed. 
This PR adds a setter for the logging level to both PICMI examples so that the user can set the value manually.

### DIFF
--- a/share/picongpu/pypicongpu/examples/laser_wakefield/main.py
+++ b/share/picongpu/pypicongpu/examples/laser_wakefield/main.py
@@ -9,8 +9,13 @@ from picongpu import picmi
 from picongpu import pypicongpu
 import numpy as np
 from scipy.constants import c
+import logging
 
 from picongpu.pypicongpu.output.png import EMFieldScaleEnum, ColorScaleEnum
+
+# set log level:
+# options (in ascending order) are: DEBUG, INFO, WARNING, ERROR, CRITICAL
+logging.basicConfig(level=logging.WARNING)
 
 """
 @file PICMI user script reproducing the PIConGPU LWFA example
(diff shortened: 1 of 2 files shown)


BrianMarre:
short question what is the `#simulation.py#` file for?

chillenzer (changes requested):
The `#simulation.py#` (as @BrianMarre already pointed out) and the `walltime.Walltime.json` are not pertinent. Please remove. Other than that, it's fine but I would have preferred a method that's general and configurable for all of PICMI like querying an environment variable inside of the package's code and having a default for that.

PrometheusPi:
@BrianMarre and @chillenzer I removed the two files I accidentally staged when committing
@chillenzer Where would you read out such an environment variable so that in any use case of PICMI/pypicongy it is set correctly?

chillenzer:
I was vague about this on purpose because this is a non-trivial question indeed. From the top of my head, I'd probably add a file `picmi_logging.py` roughly like this
```python
# untested, interpret as pseudo-code
import logging
import os

logging.basicConfig(level=os.environ.get('PICMI_LOGLEVEL', 'WARN'))
```
and replace every usage of `import logging` by `from picmi_logging import logging`. Not sure how to handle the separation of PICMI/PyPIConGPU here. Maybe a refinement could be `os.environ.get('PICMI_LOGLEVEL', os.environ.get("PYPICONGPU_LOGLEVEL', 'WARN'))` for PICMI and in reverse order for PyPIConGPU, so that each one can be set separately but respects the other if it's not set explicitly.

PrometheusPi:
@chillenzer Since this is, as you mentioned, "a non-trivial question", I would suggest discussing how we should handle this in the next `dev` meeting and for now use the explicit setting on the setup files. Do you agree?

chillenzer (approved):

PR #5452, opened 2025-08-25
add another vim/emacs backup file type to gitignore

PrometheusPi:
Since I now accidentally added such backup files twice, here is the updated gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # tmp files
 *~
+\#*#
 
 # netbeans project files
 /nbproject/


chillenzer (changes requested):
I would argue against adding this type of user-specifics to the project. You are totally free to use your user-global `~/.gitignore` to ignore stuff concerning your personal editor choice. That's nothing to do with PIConGPU.

ikbuibui:
I am guilty of this in the past myself, but I think @chillenzer has a fair point. Another alternative if you want to keep project specific ignores is to add them to `.git/info/exclude`

PrometheusPi:
@chillenzer fine with me. I will write my own `~/.gitignore` and close this PR.

PR #5454, opened 2025-08-26
Add tests for SelfRegisteringRenderedObject

chillenzer:
As a special gift to @BrianMarre, here's the requested tests.

### DIFF
--- a/lib/python/picongpu/pypicongpu/rendering/renderedobject.py
+++ b/lib/python/picongpu/pypicongpu/rendering/renderedobject.py
@@ -211,7 +211,7 @@ class RenderedObject:
         """
         check if the given context is valid for the given type
 
-        Raises on error, passes silently if okay.
+        Raises on error, returns given context if okay.
 
         :raise ValidationError: on schema violation
         :raise RuntimeError: on schema not found
--- a/lib/python/picongpu/pypicongpu/rendering/renderedobject.py
+++ b/lib/python/picongpu/pypicongpu/rendering/renderedobject.py
@@ -213,6 +213,8 @@ class RenderedObject:
 
         Raises on error, returns given context if okay.
 
+        :return: given rendering context
+
         :raise ValidationError: on schema violation
         :raise RuntimeError: on schema not found
         """
--- a/lib/python/picongpu/pypicongpu/rendering/renderedobject.py
+++ b/lib/python/picongpu/pypicongpu/rendering/renderedobject.py
@@ -251,6 +251,10 @@ class SelfRegistering:
     def _register(cls):
         if cls._name not in cls._names:
             cls._names.append(cls._name)
+        else:
+            raise TypeError(
+                f"Attempt to register {cls=} with name {cls._name=} failed because that was registered before."
+            )
 
     def __init_subclass__(cls):
         super().__init_subclass__()
--- a/test/python/picongpu/quick/pypicongpu/rendering/renderedobject.py
+++ b/test/python/picongpu/quick/pypicongpu/rendering/renderedobject.py
@@ -326,19 +326,19 @@ class TestSelfRegisteringRenderedObject(unittest.TestCase):
         RenderedObject.check_context_for_type = lambda _, c: c
 
     def test_names_list_is_empty(self):
-        self.assertSequenceEqual(self.Base().get_rendering_context()["typeID"].keys(), [])
+        self.assertSetEqual(set(self.Base().get_rendering_context()["typeID"].keys()), set())
 
     def test_subclass_without_name_is_not_registered(self):
         class UnregisteredSubclass(self.Base):
             pass
 
-        self.assertSequenceEqual(list(self.Base().get_rendering_context()["typeID"].keys()), [])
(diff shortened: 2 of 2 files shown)


test/python/picongpu/quick/pypicongpu/rendering/renderedobject.py
```diff
@@ -312,3 +313,128 @@ def _get_serialized(self):
...
+
+class TestSelfRegisteringRenderedObject(unittest.TestCase):
+    def setUp(self):
+        class Base(SelfRegisteringRenderedObject):
+            def _get_serialized(self):
+                return {}
+
+        self.Base = Base
+        # We don't want to check against a schema for now:
+        RenderedObject.check_context_for_type = lambda _, c: c
```
BrianMarre, line 326:
```suggestion
        # replace the method checking against a schema with a no-op, necessary since no schema exists for the test classes
        RenderedObject.check_context_for_type = lambda _, c: c
```

BrianMarre: @chilenzer Please note: The documentation of `RenderedObject.check_context_for_type(type_to_check, context)` is wrong. You changed `check_context_for_type()` to return the rendering context but the documentation still says it only passes silently and is missing the documentation of the return.
Your implementation breaks expectation a bit, check methods are supposed to only check and are consistently implemented in PICMI/PyPIConGPU in this way.
I see why you built this feature in and am willing to keep it, but the documentation should match behaviour

test/python/picongpu/quick/pypicongpu/rendering/renderedobject.py
```diff
@@ -312,3 +313,128 @@ def _get_serialized(self):
...
+
+    def test_names_list_is_empty(self):
+        self.assertSequenceEqual(self.Base().get_rendering_context()["typeID"].keys(), [])
```
BrianMarre: Question: why do we not check the `_names` member directly here? Is it because you consider `_names` a non public interface that might change?

test/python/picongpu/quick/pypicongpu/rendering/renderedobject.py
```diff
@@ -312,3 +313,128 @@ def _get_serialized(self):
...
+        class RegisteredSubclass1(self.Base):
+            _name = "arbitrary_name1"
+
+        class RegisteredSubclass2(self.Base):
+            _name = "arbitrary_name2"
+
+        self.assertSequenceEqual(
+            list(self.Base().get_rendering_context()["typeID"].keys()),
+            [RegisteredSubclass1._name, RegisteredSubclass2._name],
+        )
```
BrianMarre: Question: Do we care that registering keeps the order of registration or is that an implementation detail? If the latter, we should use a ordering agnostic comparison here, for example by calling `sort()` on both list.

BrianMarre (changes requested):
We are missing a test for only one typeID in `get_rendering_context()` is true, and that we will get an error if two leafs with the same name register in the same base class.

In addition the test of the `SelfRegistering` functionality seem to be mixed into the `SelfRegisteringRenderedObject` test.
Would it make sense to separate the tests of the self registering functionality and the rendering via typeID dicts in separate test classes, one for `SelfRegistering` and one for `SelfRegisteringRenderedObject`

test/python/picongpu/quick/pypicongpu/rendering/renderedobject.py
chillenzer: Offline discussion: The latter, `_names` is not public interface.

chillenzer:
Offline discussion:
- Checking "exists not" statements by example is logically impossible. If someone can find a way to produce a case with more than one `True` in the rendering context by valid use of the public interface, we should add a check internally and a test that this check is triggered.
- Two leaves with the same name is currently undefined behaviour. We should define this to be erroneous, add corresponding functionality and a test checking this.
- The `SelfRegistering` class is an implementation detail. We don't test implementation details.

test/python/picongpu/quick/pypicongpu/rendering/renderedobject.py
chillenzer: You are completely correct. The order is an implementation detail that the tests are not supposed to define. Thanks for catching this!

lib/python/picongpu/pypicongpu/rendering/renderedobject.py
```diff
@@ -211,7 +211,7 @@ def check_context_for_type(type_to_check: type, context: dict | None):
         """
         check if the given context is valid for the given type
 
-        Raises on error, passes silently if okay.
+        Raises on error, returns given context if okay.
```
BrianMarre, line 214:
```suggestion
        Raises on error, returns given context if okay.

        :return: given rendering context
```

BrianMarre (changes requested):
Otherwise looks good

lib/python/picongpu/pypicongpu/rendering/renderedobject.py
chillenzer: Merci!

BrianMarre (approved):

ikbuibui (approved):

PR #5455, opened 2025-08-27
rename log level 0 to mandatory

ikbuibui:
Rename the log level which is always active to `MANDATORY` instead of `NOTHING`. This level hasn't been used anywhere in the code.

### DIFF
--- a/include/picongpu/plugins/radiation/debug/PIConGPUVerboseLogRadiation.hpp
+++ b/include/picongpu/plugins/radiation/debug/PIConGPUVerboseLogRadiation.hpp
@@ -36,12 +36,12 @@ namespace picongpu
             DEFINE_VERBOSE_CLASS(PIConGPUVerboseRadiation)
             (
                 /* define log levels for later use
-                 * e.g. log<pmaccLogLvl::NOTHING>("TEXT");*/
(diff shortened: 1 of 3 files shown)


include/picongpu/plugins/radiation/debug/PIConGPUVerboseLogRadiation.hpp
```diff
@@ -36,12 +36,12 @@ namespace picongpu
             DEFINE_VERBOSE_CLASS(PIConGPUVerboseRadiation)
             (
                 /* define log levels for later use
-                 * e.g. log<pmaccLogLvl::NOTHING>("TEXT");*/
-                DEFINE_LOGLVL(0, NOTHING); DEFINE_LOGLVL(1, PHYSICS); DEFINE_LOGLVL(2, SIMULATION_STATE);
```
chillenzer, line 40: Did you actually test that this works? It might be me but to me it looks very much like this would be a bitmask and `0` in a bitmask doesn't sound like it should work.

chillenzer (changes requested):

include/picongpu/plugins/radiation/debug/PIConGPUVerboseLogRadiation.hpp
ikbuibui: You are right.

ikbuibui: It collides with the physics level

PR #5456, opened 2025-08-28
Refactor passing parameters from PICMI to PyPIConGPU

chillenzer:
This is a major refactoring to simplify passing through identical parameters from PICMI to PyPIConGPU. With the currently style of development, we often run into the case where many classes on the two levels are basically identical in their data content. Writing out the conversion by hand is tedious and error-prone.

I propose to add a helper decorator `converts_to` that transforms a class as follows:
```python
class A:
    attribute=0
    # ... many more attributes
    
@converts_to(A)
class B:
    attribute=42
    # ... many more attributes
    
# results in:
class B:
    attribute=42
    # ... many more attributes
    
    def get_as_pypicongpu(self, *args, **kwargs):
        a = A()
        a.attribute = self.attribute
        # ... many more attributes
        return a
```
There's a number of customisation options available like renamings, custom conversion functions, etc. as well as a few convenient conventions that are documented in the docstring and the tests. The `default_converts_to` adds some more tailored defaults for our codebase.

The main contribution is `picmi/copy_attributes.py` which has a corresponding `quick/picmi/copy_attributes.py` test file. Carefully read the tests to understand the interface and functionality.

Other than that, most changes are concerned with applying the new functionality throughout some parts of the codebase that lend themselves particularly well to it. So, if you want to see it in action, have a look at the laser or plugin implementations.

It should be noted that over time our PICMI interface will probably develop into a more independent ecosystem such that this kind of technology might be needed less (or at least might be contained to a subset of functionality as present today) in the distant future. Also, it can -- admittedly -- make some edge cases harder to debug.

I have applied some renamings on the PyPIConGPU side for convenience and also let `ruff` reorder imports into canonical order while I was at it.

### DIFF
--- a/lib/python/picongpu/picmi/copy_attributes.py
+++ b/lib/python/picongpu/picmi/copy_attributes.py
@@ -1,0 +1,121 @@
+"""
+This file is part of PIConGPU.
+Copyright 2025 PIConGPU contributors
+Authors: Julian Lenz
+License: GPLv3+
+"""
+
+import inspect
+from typing import Callable
+
+
+def _sanitize_conversions(conversions, from_instance, to_instance):
+    conversions = conversions or {}
+
+    for new_key, old in conversions.items():
+        if not hasattr(to_instance, new_key):
+            failed_conversion = {new_key: old}
+            message = (
+                f"Conversion failed. '{new_key}' is not a member of the receiver {to_instance}. "
+                f"You gave {failed_conversion}."
+            )
+            raise ValueError(message)
+        if isinstance(old, str) and not hasattr(from_instance, old):
+            failed_conversion = {new_key: old}
+            message = (
+                f"Conversion failed. '{old}' is not a member of the provider {from_instance}. "
+                f"You gave {failed_conversion}."
+            )
+            raise ValueError(message)
+
+    return {name: _value_generator(old) if isinstance(old, str) else old for name, old in conversions.items()}
+
+
+def _value_generator(name):
+    return lambda self: getattr(self, name)
+
+
+def copy_attributes(
+    from_instance, to, conversions: None | dict[str, str | Callable] = None, remove_prefix: str = "", ignore=tuple()
+):
+    """
+    Copy attributes from one object to another.
+
+    This function copies attributes from the `from_instance` to `to` if `to` is an class instance.
+    If `to` is a class, an instance will be created via `to()`, filled and returned.
+
+    This function only copies attributes under the following circumstances:
+        - The attribute exists in `to`.
+        - The attribute is not private in `from_instance`.
+
+    Optionally, `conversions` allows to define custom mappings between attributes.
+    Keys are strings and interpreted as the name of the attribute to be set on `to`.
+    For the values, two semantics are supported:
+        - If a `value` is a string, the `(key, value)` pair is interpreted as renaming, i.e.
+          `to.key = from_instance.value`.
+        - Otherwise, `value` must be a Callable that takes `from_instance` as first and only argument
+          and returns the value that `key` is supposed to have in `to`, i.e.
+          `to.key = value(from_instance)`.
+    """
+    if isinstance(to, type):
+        try:
+            return copy_attributes(
+                from_instance, to(), conversions=conversions, remove_prefix=remove_prefix, ignore=ignore
+            )
+        except TypeError as e:
+            message = (
+                "Instantiation failed. The receiving class must be default constructible. "
+                f"You gave {to} which expects {len(inspect.signature(to.__init__).parameters) - 1} argument in its constructor. "
+                "You can work with an instance instead of a class in this case."
+            )
+            raise ValueError(message) from e
+
+    assignments = {
+        to_name: _value_generator(from_name)
+        for from_name, _ in inspect.getmembers(from_instance)
+        if from_name not in ignore
+        and not from_name.startswith("_")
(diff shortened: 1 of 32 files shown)


lib/python/picongpu/picmi/diagnostics/binning.py
```diff
@@ -5,19 +5,26 @@
...
+)
+from ...pypicongpu.output.binning import (
     BinningAxis as PyPIConGPUBinningAxis,
+)
+from ...pypicongpu.output.binning import (
+    BinningFunctor as PyPIConGPUBinningFunctor,
+)
+from ...pypicongpu.output.binning import (
     BinSpec as PyPIConGPUBinSpec,
 )
```
ikbuibui: This looks weird

lib/python/picongpu/picmi/distribution/GaussianDistribution.py
```diff
@@ -15,6 +16,12 @@
 import numpy as np
 
 
+@converts_to(
+    Gaussian,
+    conversions={"vacuum_cells_front": lambda self, _: int(self.vacuum_front / self.cell_size[1])},
+    preamble=lambda self, grid: setattr(self, "cell_size", grid.get_cell_size()) or self.check(),
```
ikbuibui, line 22: A sign of my python inexperience, but I had a hard time figuring it out that check will actually be executed. Is this a normal way to do this?

lib/python/picongpu/picmi/copy_attributes.py
```diff
@@ -0,0 +1,170 @@
...
+
+    return decorator
+
+
+def default_converts_to(to_class, conversions=None, preamble=None, remove_prefix="", ignore=tuple()):
+    return converts_to(
+        to_class,
+        conversions=conversions,
+        preamble=preamble
+        or (lambda self, *args, **kwargs: self.check(*args, **kwargs) if has_attribute(self, "check") else None),
```
ikbuibui, line 164: The current design choice is to run `check`, if it exists, if no preamble is provided. If it is provided, the user is responsible to run `check` if they want it. 
This is fine, but I was wondering if we should always check, separate it from the preamble, and make it opt out.

chillenzer:
One minor thing that I'm still not sure about is the following: `converts_to` is mostly a tool agnostic of being used in `pypicongpu` except for the fact that the name of the method it adds is hardcoded. Originally, I didn't plan to introduce `default_converts_to`, so that was a reasonable compromise. But now that we have `default_converts_to`, we could make `converts_to` fully agnostic of the context it's designed in and hide the name of the function in `default_converts_to`. On the other hand, there's no need to do so and it's unlikely to be a customisation point in the medium-term future, so it might rather become a maintenance burden.

lib/python/picongpu/picmi/copy_attributes.py
chillenzer: Mmmmhhhh... I'm not so much in favour of this whole `check` business anyways. I'd prefer to have the constructor handle argument normalisation and let it raise an exception if it fails to do so. In that sense, I have no inclination to lean into that further by defining an (implicit) protocol around this.

ikbuibui:
>  But now that we have `default_converts_to`, we could make `converts_to` fully agnostic of the context it's designed in and hide the name of the function in `default_converts_to`. On the other hand, there's no need to do so and it's unlikely to be a customisation point in the medium-term future, so it might rather become a maintenance burden.

I would agree that since it is unlikely to become a customization point, we can keep it as it is. Even if you need to change this in the future, as it stands its mostly tests which would need to be refactored.

lib/python/picongpu/picmi/copy_attributes.py
chillenzer: Detecting and running `check` is mostly a concession to the current reality of the code and an attempt to separate concerns. Otherwise I'd have restructured the code even more.

ikbuibui:
Just to say it, I really like the flexibility of the decorator. I almost enjoyed looking at it.

lib/python/picongpu/picmi/distribution/GaussianDistribution.py
chillenzer: That is admittedly kind of a hack: I wanted a lambda that does two things and still returns `None`. So, the first expression returns `None` which evaluates to `False` in the the `or` concatenation, so the lazy evaluation of the second expression will always happen, again returning `None` which is the return value of the function. I agreed that it's not quite the obvious way to write this but it's very concise.

lib/python/picongpu/picmi/diagnostics/binning.py
chillenzer: Sorry, not sure why it did this. I'll revert if there are other changes to make.

ikbuibui (dismissed):

chillenzer:
Resolved conflicts and removed weird import formatting. Ready for merge once the CI has passed. @ikbuibui

lib/python/picongpu/picmi/copy_attributes.py
BrianMarre: the reason why our PICMI implementation has `check()` methods is that otherwise you get 300+ line constructors, which carries with its own baggage.
There are ways to avoid a lot of the checks using pydantic, but that requires a mayor refactor of the code base, something I am planning to do some time in the future, but not right now.

ikbuibui (approved):

PR #5458, opened 2025-08-29
fix max memory per device for capella/zih

PrometheusPi:
Our current `getDevice` on ZIH's capella requested too much memory and triggered the error:
> srun: lua: Non-exclusive jobs on Alpha Centauri are allowed to request at most 193250 MB memory per requested GPU to allow a fair sharing of resources within a single node. 
> srun: error: cli_filter plugin terminated with error

This PR fixes this. 

Additionally, this PR removes ZIH's decommissioned partitions/cluster and renames the directory to `zih-tud` instead of being cluster specific (and outdated) as before.


PrometheusPi:
I set this PR to draft mode: There seem to be more problems with how we allocate memory on Capella. I will try to fix them and will add these fixes here. So please do not yet merge.

PrometheusPi:
FYI: ZIH support just answered by ticket: 
They are aware of the problem of wrongly configured memory requests. They will now forward this issue to experts.

PrometheusPi:
I received the magic number how much memory has to be subtracted from the available memory for the OS so that the jobs actually start but also do not crash due to out-of-memory issues. Ready for review.

ikbuibui (approved):

chillenzer:
Anything blocking this?

PR #5362, opened 2025-05-06
add capella cluster at ZIH / TU Dresden setup

PrometheusPi:
This adds the example profile and the `*.tpl` file for the [Capella cluster](https://compendium.hpc.tu-dresden.de/jobs_and_resources/capella/) at ZIH / TU Dresden. 

There seems to be no clear documentation on how much memory one can request per node. Thus, I had to test it out. 

@BeyondEspresso Could you please test compiling with the pre-compiled libraries.

PR #5463, opened 2025-08-31
Remove profiles of decommisioned HPC systems

steindev:


ikbuibui (approved):

PrometheusPi:
Removing this is a severe issue - as of #5362, the current profiles reside in there as well and have undergone fixes as of 3 days ago that are currently pending #5458

PR #5470, opened 2025-09-03
Add configuration to enable/disable background field for plugins and dumps

ikbuibui:
Add state to `FieldBackground`.
Add config options to `fieldBackground.param` to enable it for plugins and dumps

Feature requested by @PrometheusPi

### DIFF
--- a/include/picongpu/simulation/stage/FieldBackground.hpp
+++ b/include/picongpu/simulation/stage/FieldBackground.hpp
@@ -165,7 +165,7 @@ namespace picongpu
             } // namespace detail
 
             //! Functor for the stage of the PIC loop applying field background
-            class FieldBackground
+            class FieldBackground : public ISimulationData
             {
             public:
                 /** Register program options for field background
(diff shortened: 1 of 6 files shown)


ikbuibui:
Tested and ready for review

ikbuibui:
I will open another PR for the refactoring on the initialization, since it is independent from this PR and is holding this one back.

include/picongpu/simulation/stage/FieldBackground.hpp
```diff
@@ -227,7 +326,6 @@ namespace picongpu
                     applyB->subtract(step);
                 }
 
-            private:
```
chillenzer, line 230: Is removing this a debugging artifact?

chillenzer (approved):
Looks fine. You removed one `private` and it doesn't look like it would be on purpose.

include/picongpu/simulation/stage/FieldBackground.hpp
ikbuibui: Not removed but moved above to the new line number 296. `add` and `subtract` are now private, since `enable` and `disable` should be used.

PR #5478, opened 2025-09-17
Remove unused simulation init and start abstractions 

ikbuibui:
Simulation starter was needlessly configurable. Simplified the design. Removed the `ISimulationStarter` interface.

Removed an unused workflow for the initialization of the `DataConnector` on simulation `init` and `slide` . Removed `IInitPlugin` and and `AbstractInitializer` interfaces and their corresponding empty concrete classes `InitPluginNone` and `SimStartInitializer`

You can see [here](https://github.com/ComputationalRadiationPhysics/picongpu/commit/553583b832b20ee3fa7f74ab64bfa5cf479a5b55#diff-616b3f819afbcab83fd416f5cf55ed16b3af03ab9a54279e9b9ae86821ed098eR149) the most likely reason the event system wait was required, is that it was before a `delete` to `SimStartInitializer` . Similarly just below in slide. So, I think removing the waits should be safe.
_You may need to open the link multiple times to actually get to the code I am referring to since the linked commit added ~83k lines and GitHub struggles a bit with this._


include/pmacc/dataManagement/DataConnector.cpp
```diff
@@ -26,11 +26,11 @@
 #include "pmacc/dimensions/Definition.hpp"
 
 #include <algorithm>
+#include <list>
```
chillenzer, line 29: You're sure you need this?

include/pmacc/dataManagement/DataConnector.cpp
```diff
@@ -47,16 +47,8 @@ namespace pmacc
...
     {
-        currentStep = initialiser.setup();
-
-        for(auto& data : datasets)
-        {
-            initialiser.init(*data, currentStep);
-        }
-
-        initialiser.teardown();
     }
```
chillenzer: Why do we still keep this?

chillenzer (dismissed):

include/pmacc/dataManagement/DataConnector.cpp
psychocoderHPC: If we do nothing here, we should best remove it completely.

psychocoderHPC (dismissed):

include/pmacc/dataManagement/DataConnector.cpp
ikbuibui: It exists in the `hpp` so its not necessary. Im not sure what the best practice is regarding having this in the cpp as well. I saw vector was included and unused, so I replaced it with what was actually being used.

include/pmacc/dataManagement/DataConnector.cpp
ikbuibui: Removed.

chillenzer (approved):

PR #5490, opened 2025-09-26
add class to evalute c++ density functions from picmi meta-data output

PrometheusPi:
This PR adds a helper class that converts the c++ code string from PICMI free density formula to a sympy expression and enables evaluating it.

### DIFF
--- a/lib/python/picongpu/extra/utils/__init__.py
+++ b/lib/python/picongpu/extra/utils/__init__.py
@@ -2,6 +2,6 @@ from .find_time import FindTime
 from .memory_calculator import MemoryCalculator
 from .field_ionization import FieldIonization
 from . import FLYonPICRateCalculationReference
-from picmi_density_reader import cpp_fct_reader
+from .picmi_density_reader import CppFctReader
 
-__all__ = ["FindTime", "MemoryCalculator", "FieldIonization", "FLYonPICRateCalculationReference", "cpp_fct_reader"]
+__all__ = ["FindTime", "MemoryCalculator", "FieldIonization", "FLYonPICRateCalculationReference", "CppFctReader"]
--- a/lib/python/picongpu/extra/utils/__init__.py
+++ b/lib/python/picongpu/extra/utils/__init__.py
@@ -2,5 +2,6 @@ from .find_time import FindTime
 from .memory_calculator import MemoryCalculator
 from .field_ionization import FieldIonization
 from . import FLYonPICRateCalculationReference
+from picmi_density_reader import cpp_fct_reader
 
-__all__ = ["FindTime", "MemoryCalculator", "FieldIonization", "FLYonPICRateCalculationReference"]
+__all__ = ["FindTime", "MemoryCalculator", "FieldIonization", "FLYonPICRateCalculationReference", "cpp_fct_reader"]
--- a/lib/python/picongpu/extra/utils/picmi_density_reader.py
+++ b/lib/python/picongpu/extra/utils/picmi_density_reader.py
@@ -122,11 +122,11 @@ class CppFctReader:
 if __name__ == "__main__":
     # load pypicongpu.json, convert json to dict and extract equation for density
     # a pypicongpu.json is created for every PICMI call
-    file = open("pypicongpu.json")
-    sim_dict = json.load(file)
-    density_fct_str = sim_dict["species_initmanager"]["operations"]["simple_density"][0]["profile"]["data"][
-        "function_body"
-    ]
+    with open("pypicongpu.json") as file:
+        sim_dict = json.load(file)
+        density_fct_str = sim_dict["species_initmanager"]["operations"]["simple_density"][0]["profile"]["data"][
+            "function_body"
+        ]
 
     # create cpp_fct_reader class for later evaluation
     reader = CppFctReader(density_fct_str)
--- a/lib/python/picongpu/extra/utils/picmi_density_reader.py
+++ b/lib/python/picongpu/extra/utils/picmi_density_reader.py
@@ -4,9 +4,13 @@ import sympy
 import json
 
 
-class cpp_fct_reader:
+class CppFctReader:
     """class that alllows to evalute the PIConGPU free density
     from a c++ code string
+
+    BE AWARE: the code assumes that:
+      1.) a free formular density profile was used in PICMI
+      2.) if sympy.Piecewise was used, that the c++ code first case, will not branch
     """
 
     def __init__(self, s, debug=False):
@@ -55,7 +59,7 @@ class cpp_fct_reader:
             return res
 
     def eval_cases(self, s, x, symbol):
-        """handle c++ cases o form cond ? case1 ? case 2
+        """handle c++ cases of form condition ? case1 : case 2
         s ... string code
         x ... float value to evalute
         symbol ... symbol to replace in string s
@@ -117,6 +121,7 @@ class cpp_fct_reader:
 
 if __name__ == "__main__":
     # load pypicongpu.json, convert json to dict and extract equation for density
+    # a pypicongpu.json is created for every PICMI call
     file = open("pypicongpu.json")
     sim_dict = json.load(file)
     density_fct_str = sim_dict["species_initmanager"]["operations"]["simple_density"][0]["profile"]["data"][
@@ -124,7 +129,7 @@ if __name__ == "__main__":
     ]
 
     # create cpp_fct_reader class for later evaluation
-    reader = cpp_fct_reader(density_fct_str)
+    reader = CppFctReader(density_fct_str)
 
     # define positions where to evaluate the density
     x_array = np.linspace(0.0, 5.0e-3, 1000)
@@ -137,6 +142,6 @@ if __name__ == "__main__":
     # plot density
     plt.plot(x_array, n_array)
     plt.xlabel(r"$y \, \mathrm{[m]}$")
-    plt.xlabel(r"$n \, \mathrm{[m^-3]}$")
+    plt.ylabel(r"$n \, \mathrm{[m^-3]}$")
     plt.yscale("log")
     plt.show()
--- a/lib/python/picongpu/extra/utils/picmi_density_reader.py
+++ b/lib/python/picongpu/extra/utils/picmi_density_reader.py
@@ -1,0 +1,142 @@
+import numpy as np
+import matplotlib.pyplot as plt
+import sympy
+import json
+
+
+class cpp_fct_reader:
+    """class that alllows to evalute the PIConGPU free density
+    from a c++ code string
+    """
+
+    def __init__(self, s, debug=False):
+        """constructor
+        s ... string from pypicongpu.json file
+        debug ... bool for debug output
+        """
+        self.debug = debug
+        self.cpp_code = s.replace("\n", " ")
+        self.cpp_code = self.initial_clean(self.cpp_code)
+
+    def evaluate(self, x, symbol="y"):
+        """evalute expression
+        x ... float: position [m] where to evaluate the density
+        symbol ... string: symbol to replace (default: 'y')
+        """
+        return self.inner_evaluate(self.cpp_code, x, symbol=symbol) + 0.0  # add float for cast
+
+    def test_for_cases(self, s):
+        """internal method checking for c++ cases
+        s ... string to check
+        """
+        index_first_q = s.find("?")
+        index_first_c = s.find(":")
+        if index_first_q == -1 and index_first_c == -1:
+            return False
+        else:
+            return True
+
+    def inner_evaluate(self, s, x, symbol):
+        """evalute string - this is a recursevly called internal method
+        s ... string code
+        x ... float value to evalute density at
+        symbol ... string symbol to replace
+        """
+        if self.test_for_cases(s):
+            if self.debug:
+                print("CASES")
+            return self.eval_cases(s, x, symbol)
+
+        else:
+            s_sym = sympy.parsing.sympy_parser.parse_expr(s)
+            res = s_sym.subs(symbol, x)
+            if self.debug:
+                print("eval:", res)
+            return res
+
+    def eval_cases(self, s, x, symbol):
+        """handle c++ cases o form cond ? case1 ? case 2
+        s ... string code
+        x ... float value to evalute
+        symbol ... symbol to replace in string s
+        """
+        if self.debug:
+            print("-->", s)
+        s = self.clean_substring(s)
+        if self.debug:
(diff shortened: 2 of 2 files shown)


PrometheusPi:
@chillenzer I would be happy to implement any idea to shorten this parser.

lib/python/picongpu/extra/utils/__init__.py
```diff
@@ -2,5 +2,6 @@
 from .memory_calculator import MemoryCalculator
 from .field_ionization import FieldIonization
 from . import FLYonPICRateCalculationReference
+from picmi_density_reader import cpp_fct_reader
 
-__all__ = ["FindTime", "MemoryCalculator", "FieldIonization", "FLYonPICRateCalculationReference"]
+__all__ = ["FindTime", "MemoryCalculator", "FieldIonization", "FLYonPICRateCalculationReference", "cpp_fct_reader"]
```
ikbuibui:
```suggestion
__all__ = ["FindTime", "MemoryCalculator", "FieldIonization", "FLYonPICRateCalculationReference", "CppFctReader"]
```
Doesn't follow naming scheme.

lib/python/picongpu/extra/utils/picmi_density_reader.py
```diff
@@ -0,0 +1,142 @@
...
+            s = s[:-1]
+
+        if s[0] == "(" and s[-1] == ")":
+            s = s[1:-1]
+        return s
+
+
+if __name__ == "__main__":
+    # load pypicongpu.json, convert json to dict and extract equation for density
+    file = open("pypicongpu.json")
```
ikbuibui: missing corresponding file close.

lib/python/picongpu/extra/utils/picmi_density_reader.py
```diff
@@ -0,0 +1,142 @@
...
+    n_array = np.zeros_like(x_array)
+
+    # evalute density
+    for i, x in enumerate(x_array):
+        n_array[i] = reader.evaluate(x)
+
+    # plot density
+    plt.plot(x_array, n_array)
+    plt.xlabel(r"$y \, \mathrm{[m]}$")
+    plt.xlabel(r"$n \, \mathrm{[m^-3]}$")
```
ikbuibui: I assume this is y label
```suggestion
    plt.ylabel(r"$n \, \mathrm{[m^-3]}$")
```

lib/python/picongpu/extra/utils/picmi_density_reader.py
```diff
@@ -0,0 +1,142 @@
...
+
+        else:
+            s_sym = sympy.parsing.sympy_parser.parse_expr(s)
+            res = s_sym.subs(symbol, x)
+            if self.debug:
+                print("eval:", res)
+            return res
+
+    def eval_cases(self, s, x, symbol):
+        """handle c++ cases o form cond ? case1 ? case 2
```
ikbuibui:
```suggestion
        """handle c++ cases of form cond ? case1 : case 2
```

lib/python/picongpu/extra/utils/picmi_density_reader.py
```diff
@@ -0,0 +1,142 @@
...
+        x ... float value to evalute
+        symbol ... symbol to replace in string s
+        """
+        if self.debug:
+            print("-->", s)
+        s = self.clean_substring(s)
+        if self.debug:
+            print("==>", s)
+        index_first_q = s.find("?")
+        index_first_c = s.find(":")  # this assumes that there are no cases in the first branch
```
ikbuibui, line 73: This comment is important because it notes that this will not work with nested ternary ops (in the position of case 1). I would move this limitation to the function description.

ikbuibui (changes requested):
My main comment is that I think the ternary op handling and parsing can be done in `init`, maybe use `sympy.Piecewise` for the ternary op, and then an `evaluate` will become quite straightforward.
You will need to provide the symbol e.g. `y` to `init` though.  
This can be useful depending on how this is used.
_Offline discussion: This feature is a stopgap till we start storing `sympy` expressions instead of recreating them from `C++`. This code is not intended to be kept in long term._
Other small comments below

lib/python/picongpu/extra/utils/__init__.py
```diff
@@ -2,5 +2,6 @@
 from .memory_calculator import MemoryCalculator
 from .field_ionization import FieldIonization
 from . import FLYonPICRateCalculationReference
+from picmi_density_reader import cpp_fct_reader
```
ikbuibui:
```suggestion
from .picmi_density_reader import cpp_fct_reader
```

lib/python/picongpu/extra/utils/__init__.py
ikbuibui: Personally I also prefer `Fn` for functions instead of `Fct`

lib/python/picongpu/extra/utils/picmi_density_reader.py
PrometheusPi: I added a comment how to generate this file

lib/python/picongpu/extra/utils/picmi_density_reader.py
PrometheusPi: great catch - fixed

lib/python/picongpu/extra/utils/picmi_density_reader.py
PrometheusPi: fixed

lib/python/picongpu/extra/utils/picmi_density_reader.py
PrometheusPi: added to the class description

lib/python/picongpu/extra/utils/__init__.py
PrometheusPi: fixed

lib/python/picongpu/extra/utils/picmi_density_reader.py
ikbuibui: I was wondering if you need a call to file close at the end of main to prevent resource leaks.
https://stackoverflow.com/questions/7395542/is-explicitly-closing-files-important

lib/python/picongpu/extra/utils/picmi_density_reader.py
```diff
@@ -0,0 +1,147 @@
...
+        if s[0] == "(" and s[-1] == ")":
+            s = s[1:-1]
+        return s
+
+
+if __name__ == "__main__":
+    # load pypicongpu.json, convert json to dict and extract equation for density
+    # a pypicongpu.json is created for every PICMI call
+    file = open("pypicongpu.json")
+    sim_dict = json.load(file)
```
chillenzer:
```suggestion
    with open("pypicongpu.json") as file:
        sim_dict = json.load(file)
```

lib/python/picongpu/extra/utils/picmi_density_reader.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def clean_substring(self, s):
+        """clean any substrings from cases from paraneties and whitespace
+        s ... string
+        """
+        while s[0].isspace():
+            s = s[1:]
+
+        while s[-1].isspace():
+            s = s[:-1]
```
chillenzer:
```suggestion
        s = s.strip()
```

lib/python/picongpu/extra/utils/picmi_density_reader.py
```diff
@@ -0,0 +1,147 @@
...
+
+        if s[0] == "(" and s[-1] == ")":
+            s = s[1:-1]
```
chillenzer: Is this needed only a single time? Otherwise we'd need at least a while loop here.
```suggestion
        while s.startswith('('):
            if not s.endswith(')'):
                raise ValueError(f"Missing closing parenthesis in {s=}.")
            s = s[1:-1]
```

lib/python/picongpu/extra/utils/picmi_density_reader.py
```diff
@@ -0,0 +1,147 @@
...
+    def initial_clean(self, s):
+        """initially clean string from c++/picongpu methods
+        making it readably for sympy
+        s ... string
+        """
+        if self.debug:
+            print("clean before:", s)
+        s = s.replace("^", "**")
+        s = s.replace("pmacc::math::exp", "exp")
+        s = s.replace("pmacc::math::pow", "pow")
```
chillenzer: This won't be sufficient. There's a long list of `pmacc::math` functions that could be generated. It might be more reliable to do `s.replace('pmacc::math::', 'std')` (or just remove the prefix).

lib/python/picongpu/extra/utils/picmi_density_reader.py
```diff
@@ -0,0 +1,147 @@
...
+            print("-->", s)
+        s = self.clean_substring(s)
+        if self.debug:
+            print("==>", s)
+        index_first_q = s.find("?")
+        index_first_c = s.find(":")  # this assumes that there are no cases in the first branch
+
+        condition = s[0:index_first_q]
+        case1 = s[index_first_q + 1 : index_first_c]
+        case2 = s[index_first_c + 1 :]
```
chillenzer, line 77: Algorithm for detecting matching parenthesis (adapted for this problem):
```python
from itertools import groupby
from sympy import Piecewise


def nesting_level(c):
    if c == "?":
        return 1
    if c == ":":
        return -1
    return 0


class Counter:
    counter = 0

    def __add__(self, addition):
        self.counter += addition
        return self.counter


def parse_branchless(string):
    # ... do parsing with guaranteed no ternary operators...
    return string


def parse_with_keys(keys_and_expressions):
    if len(keys_and_expressions) == 1:
        return parse_branchless(keys_and_expressions[0][1].strip("?:() "))

    condition = parse_with_keys(keys_and_expressions[:1])
    # poor man's index finding because list.index doesn't support key functions
    else_index = (
        next(
            filter(
                lambda x: x[1][0] == keys_and_expressions[0][0],
                enumerate(keys_and_expressions[1:]),
            )
        )[0]
        + 1
    )
    then_expression = parse_with_keys(keys_and_expressions[1:else_index])
    else_expression = parse_with_keys(keys_and_expressions[else_index:])
    # should be
    # return Piecewise([(then_expression, condition), (else_expression, True)])
    # but I didn't implement the parse function
    return [(then_expression, condition), (else_expression, True)]


def parse(string):
    counter = Counter()
    return parse_with_keys(
        list(
            map(
                lambda x: (x[0], "".join(x[1])),
                groupby(s, lambda c: counter + nesting_level(c)),
            )
        )
    )


if __name__ == "__main__":
    # a little demonstration
    s = "c>1 ? (c<3 ? 5: 6) : (c<-5 ? 42 : 2)"
    print(parse(s))
```

chillenzer (changes requested):
Offline discussion: The idiomatic approach would probably be to define the corresponding `transformations` and add them to a call list `sympy_parser.parse_expr(s, transformations=transformations)`. But it is unclear whether we want to build all that infrastructure for some transitional code that will be removed in due time. (Saying this is a sure way for a piece of code to become critical infrastructure relied upon by generations and generations of future users (see https://xkcd.com/2730/).)

lib/python/picongpu/extra/utils/picmi_density_reader.py
PrometheusPi: Valid point
will fix

lib/python/picongpu/extra/utils/picmi_density_reader.py
PrometheusPi: Done (already suggested by @ikbuibui)

lib/python/picongpu/extra/utils/picmi_density_reader.py
PrometheusPi: Thanks - I forgot about that method :-)

lib/python/picongpu/extra/utils/picmi_density_reader.py
PrometheusPi: I like your code

lib/python/picongpu/extra/utils/picmi_density_reader.py
PrometheusPi: That's true - I will `s.replace('pmacc::math::', ' ')`

lib/python/picongpu/extra/utils/picmi_density_reader.py
PrometheusPi: I will add this tomorrow

PrometheusPi: I might need some more time to do so

lib/python/picongpu/extra/utils/picmi_density_reader.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+if __name__ == "__main__":
+    # load pypicongpu.json, convert json to dict and extract equation for density
+    # a pypicongpu.json is created for every PICMI call
+    with open("pypicongpu.json") as file:
+        sim_dict = json.load(file)
+        density_fct_str = sim_dict["species_initmanager"]["operations"]["simple_density"][0]["profile"]["data"][
+            "function_body"
+        ]
```
chillenzer, line 126: Heads-up: Nitpicking coming your way!

I'd probably even unindent that last line. That way the file will only be kept open for as long as it is really necessary. In general, that's a good thing to aim for in order to avoid various risks of file corruption, blocking other code's access, etc. but also to more clearly communicate the dependencies in your code. (`json.load(...)`'s return value might still hold a reference to the file and its `[]` operators could potentially rely on that file still being open.)

In practice, the dictionary look-up afterwards is certainly completely irrelevant in that respect but one might choose following such style rules in the harmless and simple cases, so they come naturally in the dangerous and difficult-to-decipher ones as well.

```suggestion
    with open("pypicongpu.json") as file:
        sim_dict = json.load(file)
    density_fct_str = sim_dict["species_initmanager"]["operations"]["simple_density"][0]["profile"]["data"][
            "function_body"
        ]
```

chillenzer:
Is this something that's actively worked on? Otherwise we should probably set it to draft to not distract reviewers from important PRs.

chillenzer:
Converted to draft as to not distract reviewers from active PRs.

PrometheusPi:
Thanks @chillenzer for converting this to a draft. I should have done this from the start 😄

chillenzer:
Is this still relevant and/or a desired feature?

PrometheusPi:
No, with the new PICMI sympy implementation, this is not needed anymore. I will close this PR.

PR #5492, opened 2025-09-29
Revert try restart state bug introduced in #5470

### LINKED ISSUE #5487: `tryRestart` fails hard
While investigating #5480, looking at the restart code I noticed this issue. The documentation for the `checkpoint.tryRestart` says "`Try to restart if a checkpoint is available else start the simulation from scratch`". However it seems to me, from my quick tests and going over the logic in the code, like if a checkpoint is not found `tryRestart` just fails and the simulation terminates. It does not start a new simulation.

ikbuibui:
Partially fixes #5487. I say partially, because the code is back to the pre #5470 working state, but still if you specify try restart with a specific step, and that step doesn't exist then the simulation hard fails. Will fix in a follow up PR. This PR only reverts a bug which I introduced below..
Revert change - https://github.com/ComputationalRadiationPhysics/picongpu/commit/efc82d49dcf42e59fb808f3143761f1ec0bbbdf1#r166655540 . Thanks to @psychocoderHPC for catching this

### DIFF
--- a/include/picongpu/simulation/control/Simulation.hpp
+++ b/include/picongpu/simulation/control/Simulation.hpp
@@ -480,7 +480,10 @@ namespace picongpu
                         else
                             this->restartStep = checkpoints.back();
                     }
+                }
 
+                if(this->restartRequested)
+                {
                     initialiserController->restart((uint32_t) this->restartStep, this->restartDirectory);
                     step = this->restartStep;
                 }


psychocoderHPC (approved):

psychocoderHPC:
@ikbuibui, without testing the code, IMO this code snippet should fix the other issue that was available before

```C++
std::vector<uint32_t> checkpoints = readCheckpointMasterFile();
if(this->restartStep < 0)
{
    
    if(checkpoints.empty())
    {
        if(this->tryRestart == false)
        {
            throw std::runtime_error(
                "Restart failed. You must provide the "
                "'--checkpoint.restart.step' argument. See picongpu --help.");
        }
        else
        {
            // no checkpoint found: start simulation from scratch
            this->restartRequested = false;
        }
    }
    else
        this->restartStep = checkpoints.back();
}
else
{
    auto checkPointIter = std::find(checkpoints.begin(),checkpoints.end(),this->restartStep);
    if(checkPointIter == checkpoints.end())
        this->restartRequested = false;
    else
        this->restartStep = checkpoints.back();
}
```

ikbuibui:
@psychocoderHPC I have something very similar here https://github.com/ComputationalRadiationPhysics/picongpu/pull/5493/files#r2391948507

PR #5493, opened 2025-09-29
Disable checkpointing properly if openPMD is missing

### LINKED ISSUE #5480: checkpoint flags still exist without openPMD
When compiling PIConGPU without openPMD, the help still shows options for checkpoints and thus might hide the fact that checkpointing will not work.

ikbuibui:
- [x]  #5492 needs to be merged before this.

Moves checkpointing simulation control in its own class
Checkpoint class now has a compile time toggle to disable checkpoint/restart functionality (fixes #5480)
Simplified restart state management and changed from using bools to an enum tracking state

### DIFF
--- a/include/picongpu/simulation/control/Simulation.hpp
+++ b/include/picongpu/simulation/control/Simulation.hpp
@@ -39,6 +39,7 @@
 #include "picongpu/random/seed/ISeed.hpp"
 #include "picongpu/simulation/control/DomainAdjuster.hpp"
 #include "picongpu/simulation/control/MovingWindow.hpp"
+#include "picongpu/simulation/control/checkpointingState.hpp"
 #include "picongpu/simulation/stage/AtomicPhysics.hpp"
 #include "picongpu/simulation/stage/Collision.hpp"
 #include "picongpu/simulation/stage/CurrentBackground.hpp"
@@ -69,6 +70,7 @@
 #include <pmacc/particles/traits/FilterByIdentifier.hpp>
 #include <pmacc/random/RNGProvider.hpp>
 #include <pmacc/random/methods/methods.hpp>
+#include <pmacc/simulationControl/Checkpointing.hpp>
 #include <pmacc/simulationControl/SimulationHelper.hpp>
 #include <pmacc/types.hpp>
 #include <pmacc/verify.hpp>
@@ -91,6 +93,8 @@ namespace picongpu
 {
     using namespace pmacc;
 
+    using SimHelper = SimulationHelper<simDim, simulationControl::Checkpointing<checkpointingEnabled>>;
+
     /**
      * Global simulation controller class.
      *
@@ -99,7 +103,7 @@ namespace picongpu
      *
      * @tparam DIM the dimension (2-3) for the simulation
      */
-    class Simulation : public SimulationHelper<simDim>
+    class Simulation : public SimHelper
     {
     public:
         /**
@@ -109,7 +113,7 @@ namespace picongpu
 
         void pluginRegisterHelp(po::options_description& desc) override
         {
-            SimulationHelper<simDim>::pluginRegisterHelp(desc);
+            SimHelper::pluginRegisterHelp(desc);
 
             // clang-format off
             desc.add_options()(
@@ -159,7 +163,7 @@ namespace picongpu
         void startSimulation() override
         {
             if(!skipSimulation)
-                SimulationHelper<simDim>::startSimulation();
+                SimHelper::startSimulation();
         }
 
         nlohmann::json metadata() const
@@ -283,7 +287,7 @@ namespace picongpu
             log<picLog::DOMAINS>("rank %1%; localsize %2%; localoffset %3%;") % myGPUpos.toString()
                 % gridSizeLocal.toString() % gridOffset.toString();
 
-            SimulationHelper<simDim>::pluginLoad();
+            SimHelper::pluginLoad();
 
             GridLayout<simDim> layout(gridSizeLocal, GuardSize::toRT() * SuperCellSize::toRT());
             cellDescription = std::make_unique<MappingDesc>(layout.sizeND(), DataSpace<simDim>(GuardSize::toRT()));
@@ -304,7 +308,7 @@ namespace picongpu
         {
             DataConnector& dc = Environment<>::get().DataConnector();
 
-            SimulationHelper<simDim>::pluginUnload();
+            SimHelper::pluginUnload();
 
             /** unshare all registered ISimulationData sets
              *
@@ -456,36 +460,10 @@ namespace picongpu
             if(initialiserController)
             {
                 initialiserController->printInformation();
-                if(this->restartRequested)
+                if(this->checkpointing.checkRestart(step))
                 {
-                    /* we do not require '--checkpoint.restart.step' if a master checkpoint file is found */
-                    if(this->restartStep < 0)
-                    {
-                        std::vector<uint32_t> checkpoints = readCheckpointMasterFile();
-
-                        if(checkpoints.empty())
-                        {
-                            if(this->tryRestart == false)
-                            {
-                                throw std::runtime_error(
-                                    "Restart failed. You must provide the "
-                                    "'--checkpoint.restart.step' argument. See picongpu --help.");
-                            }
-                            else
-                            {
-                                // no checkpoint found: start simulation from scratch
-                                this->restartRequested = false;
-                            }
-                        }
-                        else
-                            this->restartStep = checkpoints.back();
-                    }
-                }
-
-                if(this->restartRequested)
-                {
-                    initialiserController->restart((uint32_t) this->restartStep, this->restartDirectory);
-                    step = this->restartStep;
+                    step = static_cast<uint32_t>(this->checkpointing.getRestartStep());
+                    initialiserController->restart(step, this->checkpointing.getRestartDir());
                 }
                 else
                 {
@@ -547,13 +525,13 @@ namespace picongpu
         void dumpOneStep(uint32_t currentStep) override
         {
             fieldBackground->toDumpState(currentStep);
-            SimulationHelper<simDim>::dumpOneStep(currentStep);
+            SimHelper::dumpOneStep(currentStep);
         }
 
         void notifyPlugins(uint32_t currentStep) override
         {
             fieldBackground->toPluginState(currentStep);
-            SimulationHelper<simDim>::notifyPlugins(currentStep);
+            SimHelper::notifyPlugins(currentStep);
         }
 
         void movingWindowCheck(uint32_t currentStep) override
--- a/include/pmacc/simulationControl/Checkpointing.hpp
+++ b/include/pmacc/simulationControl/Checkpointing.hpp
@@ -1,0 +1,440 @@
+/* Copyright 2025 Tapish Narwal
+ *
+ * This file is part of PMacc.
+ *
+ * PMacc is free software: you can redistribute it and/or modify
+ * it under the terms of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * PMacc is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with PMacc.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#pragma once
+
+#include "pmacc/Environment.hpp"
+#include "pmacc/filesystem.hpp"
+#include "pmacc/mappings/simulation/Filesystem.hpp"
+#include "pmacc/mappings/simulation/GridController.hpp"
+#include "pmacc/pluginSystem/Slice.hpp"
+#include "pmacc/pluginSystem/containsStep.hpp"
+#include "pmacc/pluginSystem/toSlice.hpp"
+#include "pmacc/simulationControl/signal.hpp"
+#include "pmacc/types.hpp"
+
+#include <charconv>
+#include <condition_variable>
+#include <cstdint>
+#include <fstream>
+#include <string>
+#include <string_view>
+#include <unordered_map>
+#include <vector>
+
+namespace pmacc::simulationControl
+{
+    // State of restart.
+    // DISABLED means no restart is attempted.
+    // TRY means a restart is attempted, but the simulation will continue even if restart fails.
+    // FORCE means a restart is attempted, and the simulation will fail if restart fails.
+    // SUCCESS means a restart was successful.
+    // FAILED means a restart failed.
+    enum class RestartState : uint8_t
+    {
+        DISABLED,
+        TRY,
+        FORCE,
+        SUCCESS,
+        FAILED
+    };
+
+    /**
+     * @brief Stream insertion operator for RestartState, converting enum to string.
+     * @param out The output stream.
+     * @param state The RestartState value to write.
+     * @return The output stream.
+     */
+    inline std::ostream& operator<<(std::ostream& out, RestartState const& state)
+    {
+        static std::unordered_map<RestartState, std::string_view> const stateToString
+            = {{RestartState::DISABLED, "disabled"},
+               {RestartState::TRY, "try"},
+               {RestartState::FORCE, "force"},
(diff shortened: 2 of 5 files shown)


include/pmacc/simulationControl/Checkpointing.hpp
```diff
@@ -0,0 +1,374 @@
...
+                                "Restart failed. No checkpoints found and no '--checkpoint.restart.step' provided.");
+                        }
+                        restartState = RestartState::FAILED;
+                        return false;
+                    }
+                    restartStep = checkpoints.back();
+                }
+
+                // Checkpoints are expected to be sorted chronologically.
+                bool const stepFound = std::binary_search(checkpoints.cbegin(), checkpoints.cend(), restartStep);
```
ikbuibui: I have this logic here

include/picongpu/simulation/control/Simulation.hpp
```diff
@@ -69,6 +70,7 @@
 #include <pmacc/particles/traits/FilterByIdentifier.hpp>
 #include <pmacc/random/RNGProvider.hpp>
 #include <pmacc/random/methods/methods.hpp>
+#include <pmacc/simulationControl/Checkpointing.hpp>
```
chillenzer, line 73: I would have expected `Checkpoint.hpp` but I presume you are distinguishing it from a single checkpoint (like referencing iteration 4892)?

include/pmacc/simulationControl/Checkpointing.hpp
```diff
@@ -0,0 +1,440 @@
...
+
+        if(auto it = stateToString.find(state); it != stateToString.end())
+        {
+            out << it->second;
+        }
+        else
+        {
+            // Handle unknown enum values gracefully.
+            out << "UNKNOWN_RESTART_STATE(" << static_cast<int>(state) << ")";
+        }
```
chillenzer: Let's hope, we've never gotta do this!

include/pmacc/simulationControl/Checkpointing.hpp
```diff
@@ -0,0 +1,440 @@
...
+     * @return The input stream.
+     */
+    inline std::istream& operator>>(std::istream& in, RestartState& state)
+    {
+        static std::unordered_map<std::string, RestartState> const stringToState
+            = {{"disabled", RestartState::DISABLED},
+               {"try", RestartState::TRY},
+               {"force", RestartState::FORCE},
+               {"success", RestartState::SUCCESS},
+               {"failed", RestartState::FAILED}};
```
chillenzer: Should we extract this? Having to update this in multiple places seems bound to fail. On the other hand, we'll hopefully not really have to update this ever.

include/pmacc/simulationControl/Checkpointing.hpp
```diff
@@ -0,0 +1,440 @@
...
+        in >> token;
+
+        if(auto it = stringToState.find(token); it != stringToState.end())
+        {
+            state = it->second;
+        }
+        else
+        {
+            // If the token is not a valid state, set the stream's failbit.
+            // boost::program_options will catch this and report an error to the user.
```
chillenzer: Wow! Dark magic. Why not just throw yourself here?

include/pmacc/simulationControl/Checkpointing.hpp
```diff
@@ -0,0 +1,440 @@
...
+        bool doSoftRestart()
+        {
+            static uint32_t nthSoftRestart = 0;
+            if(nthSoftRestart <= softRestarts)
+            {
+                nthSoftRestart++;
+                return true;
+            }
+            return false;
+        }
```
chillenzer: Not guarded by `checkpointingEnabled`.

include/pmacc/simulationControl/Checkpointing.hpp
```diff
@@ -0,0 +1,440 @@
...
+     * (checkpointing) and to resume a simulation from a saved state (restarting).
+     * The checkpointing can be triggered based on simulation steps or wall-clock time.
+     * The entire functionality can be enabled or disabled at compile time via the
+     * `checkpointingEnabled` template parameter.
+     *
+     * @tparam checkpointingEnabled A boolean to enable/disable checkpointing features.
+     */
+    template<bool checkpointingEnabled>
+    struct Checkpointing
+    {
```
chillenzer: I strongly dislike the approach of guarding each individual function body with an `if constexpr` because you have to repeat this over and over again. I would have expected to just provide a specialisation
```C++
template<>
struct Checkpointing<false> {
    void registerHelp(...) {}
    ...
};
```
Is there a strong reason for your design?

include/pmacc/simulationControl/Checkpointing.hpp
```diff
@@ -0,0 +1,440 @@
...
+                    ("checkpoint.period", po::value<std::string>(&checkpointPeriod),
+                     "Period for checkpoint creation [interval(s) based on steps]")
+                    ("checkpoint.timePeriod", po::value<std::uint64_t>(&checkpointPeriodMinutes),
+                     "Time periodic checkpoint creation [period in minutes]")
+                    ("checkpoint.directory", po::value<std::string>(&checkpointDirectory)->default_value(checkpointDirectory),
+                     "Directory for checkpoints");
+                // clang-format on
+                // translate checkpointPeriod string into checkpoint intervals
+                seqCheckpointPeriod = pluginSystem::toTimeSlice(checkpointPeriod);
+            }
```
chillenzer: Diff'ed against the removed parts from `SimulationHelper` and looks correct.

include/pmacc/simulationControl/SimulationHelper.cpp
```diff
@@ -27,52 +27,33 @@
...
-        {
-            // notify all concurrent threads to exit
-            std::unique_lock<std::mutex> lk(this->concurrentThreadMutex);
-            exitConcurrentThreads.notify_all();
-        }
-        // wait for time triggered checkpoint thread
-        if(checkpointTimeThread.joinable())
-            checkpointTimeThread.join();
-
+        checkpointing.endTimeBasedCheckpointing();
```
chillenzer: Wow, camel case is not great here. I had to dig into the code to realise that it's not an apocalytic "end-time-based checkpointing" but it's just "ending the time-based checkpointing". Could we rename to "finish"/"cancel"/...?

(I would have been curious what an apocalyptic "end-time-based checkpointing" could look like, though, if you ever want to implement one...)

chillenzer (changes requested):
Partial review.

include/picongpu/simulation/control/Simulation.hpp
ikbuibui: Yeah, this was the main cause of my unease.

include/pmacc/simulationControl/Checkpointing.hpp
chillenzer: Please rename this function. Starting with `do...` suggests heavily that it actually `do`es something. In the `while` context it's used in it doesn't read too bad but out of context it's weird. You might want to try something along the lines of `moreSoftRestartAttemptsAllowed`.

include/pmacc/simulationControl/Checkpointing.hpp
```diff
@@ -0,0 +1,440 @@
...
+                            while(exitConcurrentThreads.wait_until(
+                                      lk,
+                                      std::chrono::system_clock::now() + std::chrono::minutes(checkpointPeriodMinutes))
+                                  == std::cv_status::timeout)
+                            {
+                                signal::detail::setCreateCheckpoint(1);
+                            }
+                        });
+            }
+        }
```
chillenzer: Same here: `do...` suggests that it does something while it actually rather `start...`s or `activate...`s.

chillenzer (changes requested):
Looks okay to me otherwise. Please do the marked renamings.

PrometheusPi:
@ikbuibui what is the status of this PR?

ikbuibui:
I still need to address the requested changes, will happen sometime this week.

include/pmacc/simulationControl/SimulationHelper.cpp
ikbuibui: renamed to `finishTimeBasedCheckpointing`

include/pmacc/simulationControl/Checkpointing.hpp
ikbuibui: Used your suggestion

include/pmacc/simulationControl/Checkpointing.hpp
ikbuibui: We allowed restart attempts, even without checkpoints (from the start of the simulation). This only maintains old functionality.

include/pmacc/simulationControl/Checkpointing.hpp
ikbuibui: I prefer to leave the responsibility of IO to boost po

include/pmacc/simulationControl/Checkpointing.hpp
ikbuibui: Unified

include/pmacc/simulationControl/Checkpointing.hpp
ikbuibui: renamed to `canSoftRestart`. But as I type it now i think this name has the `empty` problem. It is not clear if the caller asking or telling

include/pmacc/simulationControl/Checkpointing.hpp
ikbuibui: renamed to `startTimeBasedCheckpointing` to complement the `finishTimeBasedCheckpointing`

include/pmacc/simulationControl/Checkpointing.hpp
ikbuibui: I moved it to this class just to keep the user facing `checkpoint.restart.loop` input

ikbuibui: now changed to the more verbose  `tryConsumeRestartAttempt`

include/pmacc/simulationControl/Checkpointing.hpp
chillenzer: Is this an idiomatic and documented way to use boost po? If so, could you please point me to a reference? Haven't found it anywhere in [the docs](https://www.boost.org/doc/libs/latest/doc/html/program_options.html).

include/pmacc/simulationControl/Checkpointing.hpp
chillenzer: Looks good. Two tiny things:
- Shall we declare the default class template without a definition and define only the specialisations?
```C++
template<bool checkpointingEnabled>
struct Checkpointing;

template<>
structu Checkpointing<true> {
...
```
- Boolean interfaces are an anti-pattern (see e.g. [this explanation and example](https://docs.astral.sh/ruff/rules/boolean-type-hint-positional-argument/)). I'd prefer to have this as an enum.
```C++
enum class CheckpointingAvailability {
...

template<CheckpointingAvailability checkpointingEnabled>
struct Checkpointing {
...
```

Just to be explicit: The second without the first would not be a good idea because the second opens the possibility of more modes.

chillenzer: PS: Not blocking.

include/pmacc/simulationControl/Checkpointing.hpp
chillenzer: I don't like that you still have a verb in the name that could be read as an imperative because it still conveys the meaning of doing something. Options would be any of the following:
- third person forms of a verb, particularly `is...`, `has...` but also `allows...`
- any participle `...allowed`, `...left`, etc.
- any attribute `...available`

Suggestions:
```C++
while(checkpointing.softRestartAttemptsAreLeft())
while(checkpointing.hasSoftRestartAttemptsLeft())
while(checkpointing.allowsSoftRestartAttempts())
```

include/pmacc/simulationControl/Checkpointing.hpp
ikbuibui: Using a custom validator is most likely the idiomatic way to do it (discussed in the SO question below), but I chose the (IMO) simpler option.
I took over the solution from [here](https://stackoverflow.com/questions/27526817/input-stream-operator-lookup-order-with-boost-program-options)

ikbuibui: I can add a validate as well if you think it helps

include/pmacc/simulationControl/Checkpointing.hpp
ikbuibui: Changed

include/pmacc/simulationControl/Checkpointing.hpp
ikbuibui: using `hasSoftRestartAttemptsLeft`

include/pmacc/simulationControl/Checkpointing.hpp
```diff
@@ -0,0 +1,508 @@
...
+                            "Restart failed. No checkpoints found and no '--checkpoint.restart.step' provided.");
+                    }
+                    restartState = RestartState::FAILED;
+                    return false;
+                }
+                restartStep = checkpoints.back();
+            }
+
+            // Checkpoints are expected to be sorted chronologically.
+            bool const stepFound = std::binary_search(checkpoints.cbegin(), checkpoints.cend(), restartStep);
```
chillenzer: This assertion is wrong, I believe: You've got the `addCheckpoint` which will push a random number to the end of the vector. Even if that's not an immediate bug currently due to the precise path through the code that's taken to arrive here, I still think that the performance gain at this point is not worth risking this to become a problem.

include/pmacc/simulationControl/Checkpointing.hpp
```diff
@@ -0,0 +1,508 @@
...
+                std::unique_lock<std::mutex> lk(this->concurrentThreadMutex);
+                exitConcurrentThreads.notify_all();
+            }
+            // wait for time triggered checkpoint thread
+            if(checkpointTimeThread.joinable())
+                checkpointTimeThread.join();
+        }
+
+        /* Returns whether a restart needs to be performed */
+        bool checkRestart(uint32_t& step)
```
chillenzer, line 237: If you have that comment right above, why not?
```suggestion
        bool restartNeedsToBePerformed(uint32_t& step)
```

chillenzer (changes requested):

include/pmacc/simulationControl/Checkpointing.hpp
ikbuibui: `restartNeedsToBePerformed` also sounds like an instruction to do something

chillenzer: I see... I was thinking about
```C++
if(checkpointing.restartNeedsToBePerformed)
```
which reads quite clearly but that might be just to my ears. This is not blocking. We can keep it as is.

include/pmacc/simulationControl/Checkpointing.hpp
ikbuibui: This is a good point, I now replaced this with a more resilient reversed find, since I think the most recent matching checkpoint should be preferred over older ones (in the event such a case occurs)

include/pmacc/simulationControl/Checkpointing.hpp
chillenzer: Okay. Not 100% happy with the code duplication that happened due to separating the template specialisations but I'll let that slip.

include/pmacc/simulationControl/Checkpointing.hpp
chillenzer: If that's not an officially supported protocol, I'd refrain from using this obscure way to signal a failure. If it's not official interface, boost po would be allowed to change this behaviour without further notice, e.g., just swallow the failbit and keep going. I can see the advantages of the "initialisation = validation" approach that Pydantic is taking as well but I wouldn't do so at the expense of relying on implementation details of a code we don't control. 

My suggestion (admittedly I haven't checked the `validate` protocol, so please fill in the blanks):
```C++
std::optional<RestartState> stringToRestartState(std::string const given) {
    // ...
    return it == ...end()? std::nullopt : it->first;
}

inline auto validate(std::istream& in, RestartState& state) {
    // ...
    auto state = stringToRestartState(token);
    if (not state.has_value()) {
        // ...
    }
    return ...
}

inline std::istream& operator>>(std::istream& in, RestartState& state) {
    auto state = stringToRestartState(token);
    // ...
}
```
(Also I'm presuming that this is not at all performance critical or anything, so not taking care of how many copies we make here. etc.)

ikbuibui: I was able to use `validate` and get away without defining the stream operators at all as there is an overload of `implicit_value` which takes the string representation.

chillenzer (approved):
Thanks for bearing with me for so long. We finally made it!

Issue #5500, opened 2025-10-10
Transition PyPIConGPU to Pydantic


chillenzer:
# Transition PyPIConGPU to Pydantic

## Introduction

This issue is supposed to track the status of transitioning PyPIConGPU to Pydantic. It will clearly layout what I consider a viable path and provide a space to discuss this. (At least) @PrometheusPi, @BrianMarre, @pordyna, @ikbuibui and @mafshari64 might want to closely monitor its progress.

I have already implemented a prototype that allows to formulate PyPIConGPU classes as `pydantic.BaseModel`s, rely on `BaseModel.model_dump` for serialisation and remove their corresponding schemata. Applying this to 2 small/medium classes already has a net -200 lines of code in the `git diff` against `mainline/dev` and that's including probably all the infrastructural changes that it will take for most of the roll-out.

Probably (or at least hopefully), the proof-of-concept PRs implementing the logic but not rolling out the changes to the whole codebase will enter `mainline/dev` relatively soon. Afterwards, new developments can already make use of them. The roll-out PRs will probably involve a lot of (slightly-above-)monkey work and might uncover shortcomings of the original implementations. It might be beneficial if the main PICMI developers contribute to them to become familiar with the new infrastructure. (Although all-in-all it should become simpler and more obvious, so hopefully more to un-learn than to learn.)

The rest of this issue will discuss motivation, scope, logical steps and plan of attack.

## Motivation

The current status relies on schemata and code-logic on various levels to validate (often redundantly) the input. That is extremely verbose but more importantly there's no official ground truth for validating against. Also, the hand-rolled code doing that is not quite user-friendly with its extremely rigid approach to type-safety.

Pydantic already offers well-tested validation and serialisation and as well as all reasonable type-conversion and interpretation of input out of the box. Formulating `pydantic.BaseModel`s will fix the schemata and easily take care of sophisticated validation in one place while being much less verbose with a more friendly UI.

In a long-term scenario in which PICMI becomes the official UI and the C++ UI is no longer officially supported, we could even move validation from within the C++ code into the classes potentially removing runtime checks and/or reducing compilation times.

## Scope and constraints

- This is an internal refactoring in a production codebase, so we will attempt to keep `mainline/dev` functional at all times.

- PICMI UI shouldn't change. This is an internal refactoring!

[...]

PR #5501, opened 2025-10-10
[Transition PyPIConGPU to pydantic] Proof-of-concept I: Remove `_get_serialized`

chillenzer:
This is the first of a series of PRs relating to #5500.

This introduces 3 new functionalities:
- `copy_attributes` will first attempt to use a constructor to instantiate a class with all the necessary information. This is necessary because `pydantic.BaseModel`s want to be fully and validly instantiated.
- `get_rendering_context` will fall back to `model_dump` if it doesn't find `_get_serialized`.
- A workaround for `SelfRegistering` to work with `pydantic.BaseModel`s. They sure love dark Python voodoo as much as we do. -.-

Afterwards it includes two examples:
- `Grid3D`: Straightforward class but includes some custom transformations and renamings. Medium-term strategy might involve adjusting (names in) the templates to reduce the necessity for such conversion logic but with the laid-out constraints in #5500 this workaround is the best we can get.
- `Auto` plugin: This one was a little trickier. It includes a non-trivially typed member (`TimeStepSpec`) which had to be converted to Pydantic as well and also relies on the self-registering feature.

Some tests were adjusted to reflect the new interface and more forgiving interface.

Each commit consists of one of the mentioned above and you might want to review each commit individually.

### DIFF
--- a/lib/python/picongpu/picmi/copy_attributes.py
+++ b/lib/python/picongpu/picmi/copy_attributes.py
@@ -8,8 +8,13 @@ License: GPLv3+
 import inspect
 from typing import Callable
 
+from pydantic import BaseModel, ValidationError
+
 
 def has_attribute(instance, name):
+    if isinstance(instance, type) and issubclass(instance, BaseModel):
+        return name in instance.model_fields or name in map(lambda x: x.alias, instance.model_fields.values())
+
     # It should be this:
     #
     #     return hasattr(instance, name)
@@ -58,8 +63,8 @@ def copy_attributes(
     """
     Copy attributes from one object to another.
 
-    This function copies attributes from the `from_instance` to `to` if `to` is an class instance.
-    If `to` is a class, an instance will be created via `to()`, filled and returned.
+    This function copies attributes from the `from_instance` to `to` if `to` is a class instance.
+    If `to` is a class, it will try to construct and return an instance filled with values from `from_instance`.
 
     This function only copies attributes under the following circumstances:
         - The attribute exists in `to`.
@@ -73,26 +78,15 @@ def copy_attributes(
         - Otherwise, `value` must be a Callable that takes `from_instance` as first and only argument
           and returns the value that `key` is supposed to have in `to`, i.e.
           `to.key = value(from_instance)`.
-    """
-    if isinstance(to, type):
-        try:
-            to_instance = to()
-        except TypeError as e:
-            message = (
-                "Instantiation failed. The receiving class must be default constructible. "
-                f"You gave {to} which expects {len(inspect.signature(to.__init__).parameters) - 1} argument in its constructor. "
-                "You can work with an instance instead of a class in this case."
-            )
-            raise ValueError(message) from e
-        return copy_attributes(
-            from_instance,
-            to_instance,
-            conversions=conversions,
-            remove_prefix=remove_prefix,
-            ignore=ignore,
-            default_converter=default_converter,
-        )
 
+    Further useful features:
+        - `remove_prefix` allows to remove a prefix from `from_instance` member names
+          before looking them up and inserting them into `to`.
+        - `ignore` allows to ignore some attributes in `from_instance` and not copy them.
+          A custom conversion takes precedence and overrides this behaviour.
+        - `default_converter` is applied to all values retrieved from `from_instance`
+          before they are put into `to`.
+    """
     assignments = {
         to_name: _value_generator(from_name)
         for from_name, _ in inspect.getmembers(from_instance)
@@ -101,9 +95,41 @@ def copy_attributes(
         and has_attribute(to, to_name := from_name.removeprefix(remove_prefix))
     } | _sanitize_conversions(conversions, from_instance, to)
 
-    for key, value_generator in assignments.items():
-        setattr(to, key, default_converter(value_generator(from_instance)))
-    return to
+    # This is a two-pass process because after generating the defaults
+    # we had to apply the custom conversions on top.
+    assignments = {
+        key: default_converter(value_generator(from_instance)) for key, value_generator in assignments.items()
+    }
+
+    if isinstance(to, type):
+        try:
+            # First case: `to` is a class and can be constructed with a fully-qualified constructor call (pydantic.BaseModel).
+            return to(**assignments)
+        except Exception:
+            try:
+                # Second case: `to` is a default-constructible class to which we can copy attributes afterwards.
+                to_instance = to()
+            except (ValidationError, TypeError) as e:
+                message = (
+                    "Instantiation failed. The receiving class must be default constructible. "
+                    f"You gave {to} which expects {len(inspect.signature(to.__init__).parameters) - 1} argument in its constructor. "
+                    "You can work with an instance instead of a class in this case."
+                )
+                raise ValueError(message) from e
+            # We've got an instance now, proceed via the path for instances.
+            return copy_attributes(
+                from_instance,
+                to_instance,
+                conversions=conversions,
+                remove_prefix=remove_prefix,
+                ignore=ignore,
+                default_converter=default_converter,
+            )
+    else:
+        # Third case: We've been given an instance directly. Copy over attributes.
+        for key, value in assignments.items():
+            setattr(to, key, value)
+        return to
 
 
 def converts_to(
--- a/lib/python/picongpu/picmi/copy_attributes.py
+++ b/lib/python/picongpu/picmi/copy_attributes.py
@@ -105,7 +105,7 @@ def copy_attributes(
         try:
             # First case: `to` is a class and can be constructed with a fully-qualified constructor call (pydantic.BaseModel).
             return to(**assignments)
-        except Exception:
+        except TypeError:
             try:
                 # Second case: `to` is a default-constructible class to which we can copy attributes afterwards.
                 to_instance = to()
--- a/lib/python/picongpu/pypicongpu/grid.py
+++ b/lib/python/picongpu/pypicongpu/grid.py
@@ -5,14 +5,15 @@ Authors: Hannes Troepgen, Brian Edward Marre, Richard Pausch, Julian Lenz
 License: GPLv3+
 """
 
-from . import util
-import typeguard
-import typing
 import enum
+from typing import Annotated
+
+from pydantic import BaseModel, Field, PlainSerializer, model_validator
+from typing_extensions import Self
+
 from .rendering import RenderedObject
 
 
-@typeguard.typechecked
 class BoundaryCondition(enum.Enum):
     """
     Boundary Condition of PIConGPU
@@ -38,8 +39,27 @@ class BoundaryCondition(enum.Enum):
         return literal_by_boundarycondition[self]
 
 
-@typeguard.typechecked
-class Grid3D(RenderedObject):
+def serialise_vec(value) -> dict:
+    return dict(zip("xyz", value))
+
+
+Vec3_float = Annotated[tuple[float, float, float], PlainSerializer(serialise_vec)]
+Vec3_int = Annotated[tuple[int, int, int], PlainSerializer(serialise_vec)]
+
+
+def serialise_grid_dist(value):
+    return (
+        value
+        if value is None
+        else {
+            "x": [{"device_cells": x} for x in value[0]],
+            "y": [{"device_cells": x} for x in value[1]],
+            "z": [{"device_cells": x} for x in value[2]],
+        }
+    )
+
+
+class Grid3D(BaseModel, RenderedObject):
     """
     PIConGPU 3 dimensional (cartesian) grid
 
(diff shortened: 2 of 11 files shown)


lib/python/picongpu/picmi/copy_attributes.py
```diff
@@ -101,9 +95,41 @@ def copy_attributes(
...
+    # we had to apply the custom conversions on top.
+    assignments = {
+        key: default_converter(value_generator(from_instance)) for key, value_generator in assignments.items()
+    }
+
+    if isinstance(to, type):
+        try:
+            # First case: `to` is a class and can be constructed with a fully-qualified constructor call (pydantic.BaseModel).
+            return to(**assignments)
+        except Exception:
```
pordyna: We could maybe also at least check if to is a subclass of `pydantic.BaseModel`, so that we are sure it is nothing with sth like `def __init__(**kwargs): pass` :D

lib/python/picongpu/pypicongpu/grid.py
```diff
@@ -38,8 +39,27 @@ def get_cfg_str(self) -> str:
         return literal_by_boundarycondition[self]
 
 
-@typeguard.typechecked
-class Grid3D(RenderedObject):
+def serialise_vec(value) -> dict:
+    return dict(zip("xyz", value))
```
pordyna, line 43:
```suggestion
    return dict(zip("xyz", value, strict=True))
```
Otherwise, iterable longer or shorter than 3 will not fail. Requires python 3.10 or higher.

lib/python/picongpu/pypicongpu/grid.py
```diff
@@ -38,8 +39,27 @@ def get_cfg_str(self) -> str:
...
+
+
+Vec3_float = Annotated[tuple[float, float, float], PlainSerializer(serialise_vec)]
+Vec3_int = Annotated[tuple[int, int, int], PlainSerializer(serialise_vec)]
```
pordyna, line 47: What does Pydantic does if input is a float with a non zero integer part? This should work for `[1.0, 2, 3.0]` but not for `[1.4 ,2, 3.2]`.

lib/python/picongpu/pypicongpu/grid.py
pordyna: Or is the wrong length of the iterator already caught automatically by pydantic with  `tuple[int,int,int]`?

lib/python/picongpu/pypicongpu/grid.py
```diff
@@ -48,47 +68,35 @@ class Grid3D(RenderedObject):
...
 
-    boundary_condition = util.build_typesafe_property(tuple[BoundaryCondition, BoundaryCondition, BoundaryCondition])
+    boundary_condition: Annotated[
+        tuple[BoundaryCondition, BoundaryCondition, BoundaryCondition],
+        PlainSerializer(lambda x: serialise_vec(map(BoundaryCondition.get_cfg_str, x)), return_type=dict),
+    ]
     """behavior towards particles crossing each boundary"""
 
-    n_gpus = util.build_typesafe_property(typing.Tuple[int, int, int])
+    gpu_cnt: Vec3_int = Field((1, 1, 1), alias="n_gpus")
```
pordyna, line 83: Does it mean it only works for 1 GPU?  why not `tuple[int,int,int]` ?

lib/python/picongpu/pypicongpu/grid.py
```diff
@@ -48,47 +68,35 @@ class Grid3D(RenderedObject):
...
     """number of GPUs in x y and z direction as 3-integer tuple"""
 
-    grid_dist = util.build_typesafe_property(typing.Tuple[list[int], list[int], list[int]] | None)
+    grid_dist: Annotated[tuple[list[int], list[int], list[int]] | None, PlainSerializer(serialise_grid_dist)] = None
     """distribution of grid cells to GPUs for each axis"""
 
-    super_cell_size = util.build_typesafe_property(typing.Tuple[int, int, int])
+    super_cell_size: Vec3_int
```
pordyna, line 89: We should probably add a Validator that checks that values are non-negative or even positive.

lib/python/picongpu/pypicongpu/grid.py
```diff
@@ -48,47 +68,35 @@ class Grid3D(RenderedObject):
...
     """size of super cell in x y and z direction as 3-integer tuple in cells"""
 
-    def _get_serialized(self) -> dict:
+    @model_validator(mode="after")
+    def check(self) -> Self:
         """serialized representation provided for RenderedObject"""
         assert all(x > 0 for x in self.cell_cnt), "cell_cnt must be greater than 0"
-        assert all(x > 0 for x in self.n_gpus), "all n_gpus entries must be greater than 0"
```
pordyna, line 72: Looks like you could get rid of these checks by defining an annotated type  `Vect3_uint` with a `>0` Validator.

test/python/picongpu/quick/pypicongpu/grid.py
```diff
@@ -8,96 +8,74 @@
...
         """test raising errors if types are wrong"""
-        g = self.g
-        with self.assertRaises(typeguard.TypeCheckError):
-            g.cell_size_si = ("54.3", 2.3, 4.5)
-        with self.assertRaises(typeguard.TypeCheckError):
-            g.cell_size_si = (1.2, "2", 4.5)
-        with self.assertRaises(typeguard.TypeCheckError):
-            g.cell_size_si = (1.2, 2.3, "126")
-        with self.assertRaises(typeguard.TypeCheckError):
-            g.cell_cnt = (11.1, 7, 8)
```
pordyna, line 48: sth like this should still fail I think? Or what do you think?

lib/python/picongpu/pypicongpu/output/timestepspec.py
```diff
@@ -5,28 +5,33 @@
...
-    if isinstance(spec, slice):
-        return {
-            "start": spec.start if spec.start is not None else 0,
-            "stop": spec.stop if spec.stop is not None else -1,
-            "step": spec.step if spec.step is not None else 1,
-        }
-    raise ValueError(f"Unknown serialization for {spec=} as a time step specifier (--period argument).")
+class Spec(BaseModel):
+    start: Annotated[int | None, PlainSerializer(lambda x: x or 0)]
+    stop: Annotated[int | None, PlainSerializer(lambda x: x or -1)]
```
pordyna: With stop=0 (which is weird, but I think should be valid?) this returns -1
Also with step=0 it returns 1. I feel like these cases should fail and not silently change.  Maybe do not rely on None being falsly but explicitly check for `is None`. Or, maybe we do not allow these to be zero somewhere else? Anyway, I like `is None` more :)

lib/python/picongpu/pypicongpu/output/timestepspec.py
```diff
@@ -5,28 +5,33 @@
...
+    specs: list[Spec]
+
+    def __init__(self, *args, **kwargs):
+        # allow to give specs as positional argument
+        if len(args) > 0 and "specs" not in kwargs:
+            kwargs |= {"specs": args[0]}
+        super(TimeStepSpec, self).__init__(*args[1:], **kwargs)
+
+    @field_validator("specs", mode="before")
+    @classmethod
```
pordyna, line 32: Since you are not using `cls`, why not @staticmethod?

pordyna (changes requested):
Hey, I looked over it. These review is mostly me not being sure what `pydantic` already does and what one has to tell it to do :D

lib/python/picongpu/picmi/copy_attributes.py
chillenzer: I don't think we should. There's no implication in either direction:

I feel that it's very reasonable to assume that `MyClass(attribute1=1, attribute2=2)` constructs an instance of `MyClass` that follows the semantics of holding/containing/... `attribute1=1` and `attribute2=2`. It might well be that its implementation does not actually store them as member variables which is totally fine with me. But whatever it does internally, if it doesn't represent something like "filled with `attribute1=1` and `attribute2=2`" after construction, the author of that class is clearly misusing the constructor.

The other way round, you can provide custom `__init__` functions to `pydantic.BaseModel` children, so they could equally well mess with whatever's going on there.

There is actually one thing fishy about what you highlighted, though: The `except Exception` is clearly too general. In #5502, I actually had to change that to the more appropriate `TypeError` that's thrown if the constructor doesn't accept any or all of the arguments passed. I could rebase that change to this PR already.

lib/python/picongpu/pypicongpu/grid.py
chillenzer: That's likely true. Also, I explicitly wrote it like this (quite a while ago, this is just moved from further below) because it would already apply to the (not-yet-supported) 2D case. Not a strong argument for or against it though, so I'd be willing to change.

lib/python/picongpu/pypicongpu/grid.py
chillenzer: Let's try and find out:
```
$ python
Python 3.13.1 | packaged by conda-forge | (main, Jan  8 2025, 09:16:48) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
warning: can't use pyrepl: setupterm: could not find terminfo database; TERM=foot-extra
>>> from pydantic import BaseModel
>>> class A(BaseModel):
...     a: int
... 
>>> A(a=1)
A(a=1)
>>> A(a=1.0)
A(a=1)
>>> A(a=1.5)
  File "<stdin>", line 1, in <module>
    A(a=1.5)
    ~^^^^^^^
  File "/home/lenz/opt/pckgs/miniforge/envs/picongpu/lib/python3.13/site-packages/pydantic/main.py", line 214, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for A
a
  Input should be a valid integer, got a number with a fractional part [type=int_from_float, input_value=1.5, input_type=float]
    For further information visit https://errors.pydantic.dev/2.10/v/int_from_float
```

lib/python/picongpu/pypicongpu/grid.py
chillenzer: That IS `tuple[int, int, int]` (resolved from `Vec3_int`). The `(1, 1, 1)` is just the default value.

lib/python/picongpu/pypicongpu/grid.py
chillenzer: That's happening in the next PR. Currently, it's still checked in the corresponding schema (which still exists). In the roll-out of actually removing the schemata, we'll have a lot to do to transfer the information from the schemata into Pydantic.

lib/python/picongpu/pypicongpu/grid.py
chillenzer: See above.

test/python/picongpu/quick/pypicongpu/grid.py
chillenzer: Actually, that's true. I'll leave them in.

lib/python/picongpu/pypicongpu/output/timestepspec.py
chillenzer: You are correct. It was infuriatingly verbose in the version before and I remembered I couldn't get it shorter back then. I already wondered why I managed to do so this time. Thanks for catching it!

lib/python/picongpu/pypicongpu/output/timestepspec.py
chillenzer: That's Pydantic convention. I presume it's that way because you are allowed to access the "schema" (i.e. the class definition) during the validation already, so you can draw in extra information, but I haven't checked.

lib/python/picongpu/pypicongpu/grid.py
pordyna: oh that makes sense

lib/python/picongpu/pypicongpu/grid.py
pordyna: looks good

lib/python/picongpu/pypicongpu/grid.py
pordyna: ah i forgot this is also catched by schema, we have sooo many redundant checks.

lib/python/picongpu/pypicongpu/output/timestepspec.py
pordyna: alright, good to know

lib/python/picongpu/picmi/copy_attributes.py
pordyna: Alright, I see your point.  I agree that the Exception is a bit too broad :D Rebase sounds good.

pordyna:
@chillenzer, Just the broad Exception and the serializer in Spec are left. The rest looks good to me.

chillenzer:
Thanks! I'm on it!

test/python/picongpu/quick/pypicongpu/grid.py
chillenzer: Funnily enough, upon closer inspection the first of the highlighted test cases is actually obsolete now because `cell_size_si` can be any float and can be converted from string. However, the second test case had a sibling below which should both still apply but I got a type annotation wrong in `grid.py`, so they didn't fire. All fixed now!

lib/python/picongpu/picmi/copy_attributes.py
chillenzer: Done

lib/python/picongpu/pypicongpu/output/timestepspec.py
chillenzer: Done

chillenzer:
All requested changes are integrated. I've taken the liberty to squash them into the last commit under the grand theme of "fixing things".

pordyna (approved):

PrometheusPi (approved):
not fully review, trusting on @pordyna

PrometheusPi:
I think there is something off with the CI. It's still not finished.

PrometheusPi:
I restarted the CI.

PR #5502, opened 2025-10-10
[Transition PyPIConGPU to pydantic] Proof-of-concept II: Remove schemata

chillenzer:
- [x] Rebase after merging #5501. Until the rebase, only the last three commits are new content.

This is the second of a series of PRs relating to #5500.

This introduces one new functionality:
The schema validation falls back to requesting a schema directly from the class (which is hopefully a `pydantic.BaseModel` in that case) if it can't find a pertinent one in the schema registry. This unfortunately only works for the immediate class at hand and does not apply to following references in other schemata. The latter would involve some bookkeeping to map the referenced URI to the corresponding Python class. Playing around with that a little yielded no really satisfying solutions.

Given that
- all of this is transitional technology that's supposed to be removed after the transition is completed
- and Pydantic already handles the validation for us and is supposed to become the ground truth eventually

I haven't put to much effort into making referencing work. Instead, we can put wildcards where there are references currently.

I've applied this to two examples from #5501: `Grid3D` and `TimeStepSpec`, but not `Auto`. Removing the reference in `plugin.Plugin.json` would allow for plugins to contain arbitrary content and effectively circumvent all schema validation which is not what we want. Removing the plugin schemata should consider all plugins in one go instead.


chillenzer:
Ready for review but the CI is not up to the task yet.
Still, feel free to have a look @pordyna and @PrometheusPi!

lib/python/picongpu/pypicongpu/grid.py
```diff
@@ -59,6 +59,21 @@ def serialise_grid_dist(value):
     )
 
 
+def all_gt(iterable, m):
+    if all(correct := [x > m for x in iterable]):
+        return iterable
+    else:
+        message = f"{iterable=} contains values <= {m=} while all should be greater than m. Valid are the following: {correct=}."
```
pordyna, line 66: So what does the `=` do in the f-string?

pordyna: I like how you do not say what is wrong, but what is correct :D Positive feedback is so important

pordyna (approved):
Looks good to me! Out of curiosity, how does such a `pydantic` generated schema look like?

lib/python/picongpu/pypicongpu/grid.py
chillenzer:
```python
name = "asdf"
print(f"{name=} == name={name}")
# prints:
# name='asdf' == name='asdf'
```

pordyna:
@PrometheusPi Do you still want to take a look? Otherwise, you can just merge it.

lib/python/picongpu/pypicongpu/grid.py
chillenzer: https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals

pordyna: What? Why didn't I know about it!? Love this

chillenzer:
> Looks good to me! Out of curiosity, how does such a `pydantic` generated schema look like?

Well, let's try:
```python
>>> from pydantic import BaseModel
>>> class A(BaseModel):
...     a: str='asdf'      
...     b: int=42
... 
>>> A.model_json_schema()
{'properties': {'a': {'default': 'asdf', 'title': 'A', 'type': 'string'}, 'b': {'default': 42, 'title': 'B', 'type': 'integer'}}, 'title': 'A', 'type': 'object'}
```

PR #5504, opened 2025-10-10
 add CI test for auto plugin

mafshari64:
ci: no-compile

### DIFF
--- a/lib/python/picongpu/picmi/diagnostics/auto.py
+++ b/lib/python/picongpu/picmi/diagnostics/auto.py
@@ -1,11 +1,13 @@
 """
 This file is part of PIConGPU.
 Copyright 2025 PIConGPU contributors
-Authors: Pawel Ordyna
+Authors: Pawel Ordyna, Masoud Afshari
 License: GPLv3+
 """
 
 import typeguard
+from typing import Union
+
 
 from ...pypicongpu.output.auto import Auto as PyPIConGPUAuto
 from ..copy_attributes import default_converts_to
@@ -20,13 +22,22 @@ class Auto:
 
     Parameters
     ----------
-    period: int
+    period: int or TimeStepSpec
         Number of simulation steps between consecutive outputs.
         Unit: steps (simulation time steps).
     """
 
-    period: TimeStepSpec
-    """Number of simulation steps between consecutive outputs. Unit: steps (simulation time steps)."""
+    def __init__(self, period: Union[int, TimeStepSpec]) -> None:
+        if not isinstance(period, (int, TimeStepSpec)):
+            raise TypeError("period must be an integer or TimeStepSpec")
+        if isinstance(period, int):
+            if period < 0:
+                raise ValueError("period must be non-negative")
+            self.period = TimeStepSpec[::period]("steps") if period > 0 else TimeStepSpec([])("steps")
+        else:
+            self.period = period
 
-    def __init__(self, period: TimeStepSpec) -> None:
-        self.period = period
+    def check(self, *args, **kwargs):
+        """Validate that period is a valid TimeStepSpec."""
+        if not isinstance(self.period, TimeStepSpec):
+            raise TypeError("period must be a TimeStepSpec")
--- a/test/python/picongpu/quick/picmi/diagnostics/auto.py
+++ b/test/python/picongpu/quick/picmi/diagnostics/auto.py
@@ -1,0 +1,64 @@
+"""
+This file is part of PIConGPU.
+Copyright 2025 PIConGPU contributors
+Authors: Masoud Afshari
+License: GPLv3+
+"""
+
+import unittest
+import typeguard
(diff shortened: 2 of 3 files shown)


test/python/picongpu/quick/picmi/diagnostics/auto.py
```diff
@@ -0,0 +1,64 @@
...
+
+                # Validate rendered specs
+                serialized = pypicongpu_auto.get_rendering_context()
+                self.assertTrue(serialized["typeID"]["auto"])
+                self.assertEqual(serialized["data"]["period"]["specs"], expected_specs)
+
+    def test_invalid_period_type(self):
+        """Test invalid input types."""
+        with self.assertRaises(typeguard.TypeCheckError):
+            Auto(period="invalid")
```
PrometheusPi, line 48: What's the difference between this line and line 57+58? 
And why are both tests needed?

mafshari64: The key difference is when and how the invalid value is introduced.

test_invalid_period_type is done During def __init__.	Auto is decorated with @typeguard.typechecked, typeguard automatically checks the type of auto period.  so the error   happens during object creation.

test_check_invalid_period	is During check().  first we create  a valid object: Auto(period=10). then, we manually break it by assigning an invalid value to auto.period.. Typeguard does not automatically recheck assignments.	auto.check(), then raises a TypeError because period is no longer a TimeStepSpec.

lib/python/picongpu/picmi/diagnostics/auto.py
```diff
@@ -20,13 +22,22 @@ class Auto:
...
-    """Number of simulation steps between consecutive outputs. Unit: steps (simulation time steps)."""
+    def __init__(self, period: Union[int, TimeStepSpec]) -> None:
+        if not isinstance(period, (int, TimeStepSpec)):
+            raise TypeError("period must be an integer or TimeStepSpec")
+        if isinstance(period, int):
+            if period < 0:
+                raise ValueError("period must be non-negative")
+            self.period = TimeStepSpec[::period]("steps") if period > 0 else TimeStepSpec([])("steps")
+        else:
+            self.period = period
```
chillenzer, line 38: This is not the `auto` plugin's responsitbility. If we want to have such a check, we should have it in `TimeStepSpec`. But keep in mind that this kind of constraint is supposed to be codified as part of a pydantic validation in the future.

test/python/picongpu/quick/picmi/diagnostics/auto.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def test_negative_period(self):
+        """Test negative integer period raises error."""
+        with self.assertRaisesRegex(ValueError, "period must be non-negative"):
+            Auto(period=-5)
```
chillenzer, line 53: See above.

lib/python/picongpu/picmi/diagnostics/auto.py
```diff
@@ -20,13 +22,22 @@ class Auto:
...
 
-    def __init__(self, period: TimeStepSpec) -> None:
-        self.period = period
+    def check(self, *args, **kwargs):
+        """Validate that period is a valid TimeStepSpec."""
+        if not isinstance(self.period, TimeStepSpec):
+            raise TypeError("period must be a TimeStepSpec")
```
chillenzer, line 43: This duplicates the check in the `__init__` method. I'd rather do something like
```python
def __init__(self, ...):
    # ...
    self.check()
```

lib/python/picongpu/picmi/diagnostics/auto.py
chillenzer, line 38: I'd rather hand anything given here over to `TimeStepSpec` and make that interpret the request. This ensures that there is a consistent conversion logic, error handling, etc. across all using classes of `TimeStepSpec`.

chillenzer (changes requested):
Thanks for your effort. Unit testing really important to maintain a high quality code base, so this is a valuable attempt.

In the details, the PR attempts to hand responsibilities to `Auto` that it really shouldn't have. It should delegate everything related to `TimeStepSpec` to `TimeStepSpec` because, you know, that's what it is actually related to. As it afterwards doesn't have any code logic in it, I don't see what could possibly be tested about it.

However, the best implementation would be have `Auto` as a shallow convenience interface like this
```python
class Auto(Png):
    def __init__(self, period):
        super().__init__(
            period=period,
            # put some default arguments here
        )
```
This would get rid of the special treatment of `Auto`, remove a lot of code in the process and clearly communicate to the user what `Auto` actually does.

PrometheusPi:
@mafshari64 What is the status of this PR?

mafshari64:
Currently, I am working on web design and hope to finish it in 1-2 weeks. Then fully focus on these PRs.

chillenzer:
Thanks for the update. Converting to draft for the moment.

PR #5505, opened 2025-10-14
Add ParticleDump and FieldDump (openPMD plugin) to PICMI

chillenzer:
This PR adds rudimentary support for output via the openPMD plugin.

Available capabilities:
- `ParticleDump(species)` with native species
- `FieldDump(fieldname)` with native fields
- `OpenPMDConfig` except for `range` (and `backend_config` untested)

Missing capabilities:
- `DerivedFieldDump` from particles
- `RangeSpec` for modelling ranges
- `Filter`
- Potentially convenience interfaces like `species_all`, `fields_all`.
- Due to an oversight the current field `J` cannot yet be dumped. This will be added in a future PR.

Some explanations:
- In discussion with @franzpoeschel we decided that the best implementation is to generate a `.toml` file for each instance of the openPMD plugin that PIConGPU is supposed to use. This is straightforward to achieve but is not quite aligned with the pypicongpu approach to rendering:
    - In pypicongpu, the runner is responsible for handling the creation of a hardcoded (!) set of files in a specific location. The individual components are only supposed to provide some content to fill in.
    - With the current implementation, the individual component is allowed to create an arbitrary set of files. I have hacked in a way to spread the information about the location where to put them. It's not particularly nice but the alternative ain't pretty either. Open for discussions here.
- In above approach, most information circumvents the `pypicongpu.json` by being directly located inside of the `.toml` files. This means a two-step approach would be necessary to get the full information from the `pypicongpu.json`: 1. Parse `pypicongpu.json` and 2. extract the corresponding `.toml` file's name and parse that as well. If parsing `pypicongpu.json` is relevant to your workflow, there are two options:
    - My current sketches for database integration, etc., have a reporting capability built-in at a later stage in which the plugins explicitly report which output is to be expected where. As I provide tooling to easily extract that information, no parsing is necessary at all and we are free to adjust the internal structure.
    - We could add redundant information to the `pypicongpu.json` file that's not necessary for rendering the templates but just meant to inform the user after the fact. I typically don't fancy this kind of solution. 
- The user interface in PIConGPU and PICMI differs: PIConGPU has a technical perspective in which the user explicitly instantiates plugins inside of PIConGPU while PICMI has a more semantic perspective in which the user describes what physical output they want to see. Translation from one to the other is done by `picmi/simulation.py` which gathers all diagnostics getting dumped into the same file(s) into one instance of the openPMD plugin.
- I have written another end-to-end test for testing this feature. As they are not yet run in CI, please be so kind to run them upon review. I'll eventually take the time to figure out how to fix running them in CI, I promise!

CAUTION: The handling of `range`s is still inconsistent (because it's not yet implemented). Should you manage to somehow pass a `range` through the different layers (not sure if anything's hindering you), an `OpenPMDConfig` instance that's identical up to `range` would be treated as different. This implies that there would be two `OpenPMDWriter` instances in PIConGPU writing to the same file. I'll take care of this once `range` is properly fleshed out.

### DIFF
--- a/test/python/picongpu/end-to-end/compare_particles.py
+++ b/test/python/picongpu/end-to-end/compare_particles.py
@@ -5,9 +5,25 @@ Authors: Julian Lenz
 License: GPLv3+
 """
 
+from pathlib import Path
 import numpy as np
 import openpmd_api as opmd
 import pandas as pd
+from picongpu.picmi.diagnostics import ParticleDump
+
+
+def load_diagnostic_result(diagnostic, result_path):
+    path = Path(str(diagnostic.result_path(result_path)).replace("%06T", 6 * "0"))
+    if isinstance(diagnostic, ParticleDump):
+        return read_particles(path).loc(axis=0)[*diagnostic.species.name.split("_", maxsplit=1)]
+    return read_fields(path, [diagnostic.fieldname])[diagnostic.fieldname]
+
+
+def read_fields(series_name, names=("E", "B")):
+    series = opmd.Series(str(series_name), opmd.Access.read_only)
+    tmp = {name: [series.iterations[0].meshes[name][c].load_chunk() for c in "xyz"] for name in names}
+    series.flush()
+    return {key: np.array(value) for key, value in tmp.items()}
 
 
 def read_particles(series_name):
--- a/lib/python/picongpu/picmi/diagnostics/__init__.py
+++ b/lib/python/picongpu/picmi/diagnostics/__init__.py
@@ -13,13 +13,20 @@ from .macro_particle_count import MacroParticleCount
 from .png import Png
 from .timestepspec import TimeStepSpec
 from .checkpoint import Checkpoint
+from .particle_dump import ParticleDump
+from .field_dump import FieldDump
+from .backend_config import BackendConfig, OpenPMDConfig
 
 __all__ = [
     "Auto",
+    "BackendConfig",
+    "OpenPMDConfig",
     "Binning",
     "PhaseSpace",
     "EnergyHistogram",
     "MacroParticleCount",
+    "ParticleDump",
+    "FieldDump",
     "Png",
     "TimeStepSpec",
     "Checkpoint",
--- a/lib/python/picongpu/picmi/diagnostics/backend_config.py
+++ b/lib/python/picongpu/picmi/diagnostics/backend_config.py
@@ -1,0 +1,18 @@
+"""
+This file is part of PIConGPU.
+Copyright 2025 PIConGPU contributors
+Authors: Julian Lenz
+License: GPLv3+
+"""
+
+from picongpu.pypicongpu.output.openpmd_plugin import OpenPMDConfig as PyPIConGPUOpenPMDConfig
+
+
+class BackendConfig:
+    def result_path(self, prefix_path):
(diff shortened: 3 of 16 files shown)


chillenzer:
Sorry, forgot to add `tomli_w` to the `pyproject.toml`. I'll do so tomorrow.

PrometheusPi (changes requested):
I am currently testing this PR (since last week). I now got it to create PIConGPU code. 

An example would be very helpful. 
And: How do you set the output period? Right now it just puts into the `*.toml` file:
```
[sink.dummy_application_name.period]
"0:-1:1" = [
    "E",
    "B",
    "J",
]
```

PrometheusPi:
When using the field output `J`, one can compile PIConGPU successful but when executing PIConGPU runs into the run-time error:
> Unhandled exception of type 'St13runtime_error' with message '[openPMD plugin]: unknown data source 'J'', terminating

This is caused by the setup falling back to the default `fileOutput.param`, in `$PICSRC/include/picongpu/param/fileOutput.param` where the native fields are defined as:
```c++
using NativeFileOutputFields = MakeSeq_t<FieldE, FieldB>;
```
and thus `J` is unknown. 

Adding only fields that were requested in `NativeFileOutputFields` and adding `FieldJ` if needed would solve this problem.

chillenzer:
Oh, I see. Sorry for the oversight. The whole feature is completely designed already, including a corresponding `fileOutput.param.mustache`. Once the preparatory #5505, #5513, #5514 are merged, I'll submit the full free-formula derived fields one. Do you insist on having `J` available in this PR already or do we push for the full feature and just acknowledge for now that this PR only supports `E` and `B`?

chillenzer:
> And: How do you set the output period?

Just for completeness (as already communicated elsewhere): There's a `period=TimeStepSpec[...]` kwarg support just as with every other plugin.

PrometheusPi:
As I would prefer going quickly and if this is already in the pipeline, merging now is fine with me. 
Just allow me a quick look at the rest of the code, then I will merge.

PrometheusPi:
Just for documentation:
With or without timestep setiings, there are many warnings of the form:
> WARNING:root:schema does not explicitly forbid unevaluated properties:
> picongpu.pypicongpu.output.timestepspec.TimeStepSpec

when executing the PICMI python file to generate the PIConGPU setup code.

chillenzer:
Yes, sorry. This is related to #5500 and resolved in #5517. It's a temporary transitional annoyance.

test/python/picongpu/end-to-end/compare_particles.py
```diff
@@ -5,9 +5,25 @@
...
+
+
+def load_diagnostic_result(diagnostic, result_path):
+    path = Path(str(diagnostic.result_path(result_path)).replace("%06T", 6 * "0"))
+    if isinstance(diagnostic, ParticleDump):
+        return read_particles(path).loc(axis=0)[*diagnostic.species.name.split("_", maxsplit=1)]
+    return read_fields(path, [diagnostic.fieldname])[diagnostic.fieldname]
+
+
+def read_fields(series_name, names=("E", "B")):
```
PrometheusPi, line 22: Ah, the test passes because there is no "J"

PrometheusPi (approved):

PR #5506, opened 2025-10-15
add Cowan paper for CKC solver

PrometheusPi:
Adds missing reference to CKC solver

### DIFF
--- a/docs/source/usage/reference.rst
+++ b/docs/source/usage/reference.rst
@@ -1,4 +1,4 @@
-.. _usage-reference:
+.. _usaAAAge-reference:
 
 Reference
 =========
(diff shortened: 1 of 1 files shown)


docs/source/usage/reference.rst
```diff
@@ -1,4 +1,4 @@
-.. _usage-reference:
+.. _usaAAAge-reference:
```
chillenzer: That is an interesting change! What does the author want to say to us here? Is it supposed to be triple-A rating for the code in general or does it want to subtly draw attention to the USA?

chillenzer (changes requested):

docs/source/usage/reference.rst
PrometheusPi: Thanks for catching this.

chillenzer (approved):

PR #5508, opened 2025-10-16
openPMD plugin: Restart from checkpoint with non-matching domain decomposition

franzpoeschel:
Factored out of #5405 as a standalone feature.

Different domain decomposition means a different distribution of parallel processes over the simulation domain. The number of parallel processes may vary in between runs.

* Fields: Mostly support this already, just do random IO accesses into the sub-grid of interest. The PML fields do not support this, skip them (for now?).
* Particles: So far, the restart logic expected that there would be exactly one precisely matching particle patch. The new logic now accepts any number of particle patches and filters the patches if they only overlap with the local domain.

TODO:

- [x] Auto-detect if PML load operations should be skipped
- [ ] Maybe someone has a good test case for this

To reviewers, please also check the comments below

### DIFF
--- a/include/picongpu/plugins/openPMD/NDScalars.hpp
+++ b/include/picongpu/plugins/openPMD/NDScalars.hpp
@@ -174,11 +174,17 @@ namespace picongpu
                 DataSpace<simDim> gridPos = Environment<simDim>::get().GridController().getPosition();
                 ::openPMD::Offset start;
                 ::openPMD::Extent count;
+                ::openPMD::Extent extent = mrc.getExtent();
                 start.reserve(ndim);
                 count.reserve(ndim);
                 for(int d = 0; d < ndim; ++d)
                 {
-                    start.push_back(gridPos.revert()[d]);
+                    /*
+                     * When restarting with more parallel processes than the checkpoint had originally been written
+                     * with, we must take care not to index past the dataset boundaries. Just loop around to the start
+                     * in that case. Not the finest way, but it does the job for now..
+                     */
+                    start.push_back(gridPos.revert()[d] % extent[d]);
                     count.push_back(1);
                 }
 
--- a/include/picongpu/plugins/openPMD/restart/LoadSpecies.hpp
+++ b/include/picongpu/plugins/openPMD/restart/LoadSpecies.hpp
@@ -96,7 +96,8 @@ namespace picongpu
                 auto positionOffset = loadedData.getIdentifier(totalCellIdx()).getPointer();
 
                 // grid-strided loop over the chunked data
-                for(int dataBlock = worker.blockDomIdx(); dataBlock < numDataBlocks; dataBlock += worker.gridDomSize())
+                for(uint32_t dataBlock = worker.blockDomIdx(); dataBlock < numDataBlocks;
+                    dataBlock += worker.gridDomSize())
                 {
                     auto dataBlockOffset = dataBlock * blockDomSize;
                     auto forEach = pmacc::lockstep::makeForEach(worker);
--- a/include/picongpu/plugins/openPMD/restart/LoadSpecies.hpp
+++ b/include/picongpu/plugins/openPMD/restart/LoadSpecies.hpp
@@ -355,8 +355,9 @@ namespace picongpu
                                 numParticlesCurrentBatch,
                                 filterRemove);
 
-                            std::cout << "Filtered " << remapCurrent << " out of " << numParticlesCurrentBatch
-                                      << " particles" << std::endl;
+                            log<picLog::INPUT_OUTPUT>(
+                                "openPMD: Keeping %1% of the current batch's %2% particles after filtering.")
+                                % remapCurrent % numParticlesCurrentBatch;
 
                             pmacc::particles::operations::splitIntoListOfFrames(
                                 *speciesTmp,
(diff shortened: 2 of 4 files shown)


include/picongpu/plugins/openPMD/NDScalars.hpp
```diff
@@ -174,11 +174,17 @@ namespace picongpu
...
                 count.reserve(ndim);
                 for(int d = 0; d < ndim; ++d)
                 {
-                    start.push_back(gridPos.revert()[d]);
+                    /*
+                     * When restarting with more parallel processes than the checkpoint had originally been written
+                     * with, we must take care not to index past the dataset boundaries. Just loop around to the start
+                     * in that case. Not the finest way, but it does the job for now..
+                     */
+                    start.push_back(gridPos.revert()[d] % extent[d]);
```
franzpoeschel, line 187: Is this really the best solution? This loads `nextId` and `startId`. If the GPU grid does not match the old one, this logic currently just uses modulus to loop around and hand out some random ID. No idea if that has downsides.

include/picongpu/plugins/openPMD/restart/LoadSpecies.hpp
```diff
@@ -46,6 +48,83 @@ namespace picongpu
...
+                            auto idx = dataBlockOffset + inBlockIdx;
+                            if(idx < size)
+                            {
+                                auto& positionVec = position_[idx];
+                                auto& positionOffsetVec = positionOffset[idx];
+                                char filterCurrent = filterKeep;
+                                for(size_t d = 0; d < simDim; ++d)
+                                {
+                                    auto positionInD = positionVec[d] + positionOffsetVec[d];
+                                    if(positionInD < patchTotalOffset[d] || positionInD >= patchUpperCorner[d])
```
franzpoeschel, line 116: This line decides if a particle from a partially overlapping domain should be considered. I treat the lower boundary as inclusive and the upper boundary as exclusive, otherwise I had single particles loaded into both processes.

include/picongpu/plugins/openPMD/restart/LoadSpecies.hpp
```diff
@@ -70,6 +149,283 @@ namespace picongpu
...
+                     */
+                    DataSpace<simDim> localToTotalDomainOffset(localDomain.offset + globalDomain.offset);
+
+                    /* params->localWindowToDomainOffset is in PIConGPU for a restart zero but to stay generic we take
+                     * the variable into account.
+                     */
+                    DataSpace<simDim> const patchTotalOffset
+                        = localToTotalDomainOffset + threadParams->localWindowToDomainOffset;
+                    DataSpace<simDim> const patchExtent = threadParams->window.localDimensions.size;
+                    DataSpace<simDim> const patchUpperCorner = patchTotalOffset + patchExtent;
```
franzpoeschel, line 285: Yeah someone please check if all these position computations match up. They seem to do in my tests.

include/picongpu/plugins/openPMD/restart/LoadSpecies.hpp
```diff
@@ -70,6 +149,283 @@ namespace picongpu
...
+                            PMACC_LOCKSTEP_KERNEL(KernelFilterParticles{})
+                                .config<DIM1>(pmacc::math::Vector{numParticlesCurrentBatch})(
+                                    mappedFrame,
+                                    numParticlesCurrentBatch,
+                                    patchTotalOffset,
+                                    patchUpperCorner,
+                                    filterKeep,
+                                    filterRemove,
+                                    alpaka::getPtrNative(filter));
+                            eventSystem::getTransactionEvent().waitForFinished();
```
franzpoeschel, line 331: Someone please verify the kernel and its call. I suspect that the call can be made more efficient.

include/picongpu/plugins/openPMD/restart/LoadSpecies.hpp
```diff
@@ -235,18 +555,48 @@ namespace picongpu
...
-                    if(patchTotalOffset == offsets[i] && patchExtent == extents[i])
-                        return i;
+                    // std::cout << "Comp.: " << patchTotalOffset << " - " << (patchTotalOffset + patchExtent)
+                    //           << "\tAGAINST " << offsets[i] << " - " << (offsets[i] + extents[i])
+                    //           << "\toffsets: " << (patchTotalOffset <= offsets[i])
+                    //           << ", extents: " << ((offsets[i] + extents[i]) <= (patchTotalOffset +
+                    //           patchExtent)) <<
+                    //           '\n';
+                    if((patchTotalOffset <= offsets[i]) == true_
+                       && ((offsets[i] + extents[i]) <= (patchTotalOffset + patchExtent)) == true_)
```
franzpoeschel, line 519: Again, someone please verify the positioning logic in here. Note that both ends are inclusive here, since ideally (i.e. in the normal case) we compare a particle patch from disk against itself, i.e. the local window which is the same.

franzpoeschel (commented):
Some lines to pay special attention to for code reviewers

include/picongpu/plugins/openPMD/restart/RestartFieldLoader.hpp
```diff
@@ -219,6 +219,13 @@ namespace picongpu
 
                 /* load from openPMD */
                 bool const isDomainBound = traits::IsFieldDomainBound<T_Field>::value;
+
+                // Skip PML fields for load balancing purposes
+                if(!isDomainBound)
+                {
+                    return;
+                }
```
franzpoeschel: TODO

PrometheusPi:
Regarding the open point "maybe someone has a good test case for this":
I guess the ion simulations from @pordyna or @paschk31 might qualify for this.

psychocoderHPC (approved):

psychocoderHPC:
Let me merge this, there is no need to delay this nice PR. I will soon add the correct particle id counter fix but this is also something which should not block the PR.

PR #5511, opened 2025-10-20
Add unit dimension vector to binning diagnostic in picmi

PrometheusPi:
This pull request adds a rudimentary unit system to the PICMI / pypicongpu binning diagnostic. 
Right now, it still uses a list of seven ~~deadly sins~~ floats. This should probably be changed before merging to a class that can allow setting units without knowledge about [the order of the unit vector (L, M, T, I, ?,? ?)](https://github.com/ComputationalRadiationPhysics/picongpu/blob/ea3ffaaf766f1bfa8a1dcb823c3eb2ff8e6b4326/include/picongpu/plugins/binning/UnitConversion.hpp#L40-L48)

So far, it also comes without test, as the test infrastructure will change with #5500

### DIFF
--- a/lib/python/picongpu/picmi/diagnostics/unit.py
+++ b/lib/python/picongpu/picmi/diagnostics/unit.py
@@ -1,0 +1,82 @@
+import numpy as np
+
+class Unit:
+    """
+    class to describe units
+    """
+    
+    # number of unit dimension
+    N_unit_dim = 7
+    
+    """
+    The unit index map associates names with array indices.
+    The name definition can be found at https://en.wikipedia.org/wiki/SI_base_unit.
+    The array index order follows that of PIConGPU:
+    include/picongpu/plugins/binning/UnitConversion.hpp, lines 40–48.
+    """    
+    _unit_index_map = {
+        "length": 0, "L": 0,
+        "mass": 1, "M": 1,
+        "time": 2, "T": 2,
+        "electric current": 3, "I": 3,
+        "thermodynamic temperature": 4, "Θ": 4,
+        "amount of substance": 5, "N": 5,
+        "luminous intensity": 6, "J": 6
+    }    
+    
+    def __init__(self, name = None):
+        """set unit vector either empty or by name"""
+        self.unit_vector = np.zeros((self.N_unit_dim))
+
+        if name != None:
+            index = self._unit_index_map.get(name)
+            if index is not None:
+                self.unit_vector[index] = 1.0
+            else:
+                raise ValueError(f"Unknown unit name: {name}")
+
+    def __getitem__(self, name):
+        """access component by name"""
+        index = self._unit_index_map.get(name)
+        if index is None:
+            raise KeyError(f"Unknown unit name: {name}")
+        return self.unit_vector[index]
+
+    def __iter__(self):
+        """return iterator of unit based on PIConGPU order: LMTIΘNJ"""
+        return iter(self.unit_vector)
+
+    def __str__(self):
+        """return string representation that only outputs relevant units"""
+        output = ""
+        # invert from (name to index) to (index to (short) name)
+        inverted = {v: k for k, v in self._unit_index_map.items()}
+
+        for i in range(self.N_unit_dim):
+            if self.unit_vector[i] != 0.0:
+                output += f"{inverted[i]}^{self.unit_vector[i]} "
+        return output
+    
+    def __pow__(self, exponent):
+        """rase unit to a power"""
+        result = Unit()
+        result.unit_vector = self.unit_vector * exponent
+        return result
+    
+    def __mul__(self, factor):
+        """multiply units with each other"""
+        result = Unit()        
+        result.unit_vector = self.unit_vector + factor.unit_vector
+        return result
+        
+    def __truediv__(self, divisor):
+        """divide one unit by another"""
+        result = Unit()
+        result.unit_vector = self.unit_vector - divisor.unit_vector
+        return result 
+
+# predefined units 
+T = Unit("T")
+M = Unit("M")
+L = Unit("L")
+I = Unit("I")
--- a/lib/python/picongpu/pypicongpu/output/binning.py
+++ b/lib/python/picongpu/pypicongpu/output/binning.py
@@ -74,6 +74,10 @@ def translate_to_cpp_type(return_type):
 
 
 def translate_to_cpp_unitDimVec(unit_dimension):
+    if unit_dimension == None:
+        # catch picmi default input
+        unit_dimension = [0., 0., 0., 0., 0., 0., 0.]
+
     if (isinstance(unit_dimension, list) and
         len(unit_dimension) == 7 and
         all(isinstance(item, float) for item in unit_dimension)):
--- a/lib/python/picongpu/pypicongpu/output/binning.py
+++ b/lib/python/picongpu/pypicongpu/output/binning.py
@@ -15,6 +15,7 @@ from ..rendering.pmaccprinter import PMAccPrinter
 from ..species import Species
 from .. import util
 from .plugin import Plugin
+from ...picmi.diagnostics.unit import Unit
 
 import typeguard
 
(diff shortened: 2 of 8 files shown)


chillenzer (changes requested):
Haven't looked at the implementation but for the interface I'd suggest to do it like this
```python
# file: .../picmi/diagnostics/unit.py

class Unit:
    exponents = np.zeros(7)
    
    def __pow__(self, exponent):
        self.exponents *= exponent
        return self
        
    def __mul__(self, other: Unit):
        if not isinstance(other, Unit):
            raise ValueError(...)
        return Unit(self.exponents + other.exponents)
    
    def __truediv__(self, other):
        ...
        
L = Unit(np.eye(7)[0,:])
T = ...
M = ...
# ...
```
That would allow stuff like
```python
from picongpu.picmi.diagnostics.unit import L, T, M
BinningFunctor(
    ...,
    unit_dimension=M * L**2 / T**2
)
```

PrometheusPi:
@chillenzer I like your Idea of having a Unit class which overwrites `__pow__`

PrometheusPi:
@chillenzer I added a unit class and added a binning example for the eSpec as in the normal LWFA example

chillenzer:
`pre-commit`'s unhappy. Please fix, @PrometheusPi !

ikbuibui (dismissed):
Looks ok once the CI is happy

chillenzer:
There will be quite a bit of movement in this part of the code. We should probably push this one if we want to have it. Is there anything missing @PrometheusPi? Shall I take this one over?

lib/python/picongpu/pypicongpu/output/binning.py
```diff
@@ -73,19 +74,36 @@ def translate_to_cpp_type(return_type):
...
+    if unit_dimension == None:
+        # catch picmi default input
+        unit_dimension = Unit()
+
+    if isinstance(unit_dimension, type(Unit())):
+        result = "std::array<double, 7>{ "
+        for num in unit_dimension:
+            result += f"{num}, "
+        result += "}"
+        return result
```
chillenzer, line 87:
```suggestion
        return f"std::array<double, 7>{{{','.join(*unit_dimension)}}}"
```

lib/python/picongpu/pypicongpu/output/binning.py
```diff
@@ -73,19 +74,36 @@ def translate_to_cpp_type(return_type):
     raise ValueError(f"Cannot translate {return_type=} to a C++ type.")
 
 
+def translate_to_cpp_unitDimVec(unit_dimension):
+    if unit_dimension == None:
```
chillenzer, line 78:
```suggestion
    if unit_dimension is None:
```
See e.g. https://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none

lib/python/picongpu/pypicongpu/output/binning.py
```diff
@@ -73,19 +74,36 @@ def translate_to_cpp_type(return_type):
...
+        unit_dimension = Unit()
+
+    if isinstance(unit_dimension, type(Unit())):
+        result = "std::array<double, 7>{ "
+        for num in unit_dimension:
+            result += f"{num}, "
+        result += "}"
+        return result
+    else:
+        raise ValueError(f"The input {unit_dimension} is not a unit class.")
```
chillenzer, line 89: Generally, I would have probably written this function as one-liner
```python
return f"std::array<double, 7>{{{','.join(*Unit(unit_dimension))}}}"
```
and would have implemented
```python
def __init__(self, other: None | Unit | ...):
```
on `class Unit` for handling the conversion/error-checking/...

lib/python/picongpu/picmi/diagnostics/unit.py
```diff
@@ -0,0 +1,82 @@
+import numpy as np
+
+class Unit:
```
chillenzer, line 3: Should we call this `UnitDimension`? Currently, the usage of "unit" vs. "unit dimension" is inconsistent across the PR (the interfaces use `unit_dimension`). I think "unit dimension" is more precise and leaves room to implement (or better: adopt) a proper unit system down the line.

lib/python/picongpu/picmi/diagnostics/unit.py
```diff
@@ -0,0 +1,82 @@
...
+        "thermodynamic temperature": 4, "Θ": 4,
+        "amount of substance": 5, "N": 5,
+        "luminous intensity": 6, "J": 6
+    }    
+    
+    def __init__(self, name = None):
+        """set unit vector either empty or by name"""
+        self.unit_vector = np.zeros((self.N_unit_dim))
+
+        if name != None:
```
chillenzer, line 31:
```suggestion
        if name is not None:
```

lib/python/picongpu/picmi/diagnostics/unit.py
```diff
@@ -0,0 +1,82 @@
...
+                self.unit_vector[index] = 1.0
+            else:
+                raise ValueError(f"Unknown unit name: {name}")
+
+    def __getitem__(self, name):
+        """access component by name"""
+        index = self._unit_index_map.get(name)
+        if index is None:
+            raise KeyError(f"Unknown unit name: {name}")
+        return self.unit_vector[index]
```
chillenzer, line 43: Should we also allow indexing by integer?

lib/python/picongpu/picmi/diagnostics/unit.py
```diff
@@ -0,0 +1,82 @@
...
+        "length": 0, "L": 0,
+        "mass": 1, "M": 1,
+        "time": 2, "T": 2,
+        "electric current": 3, "I": 3,
+        "thermodynamic temperature": 4, "Θ": 4,
+        "amount of substance": 5, "N": 5,
+        "luminous intensity": 6, "J": 6
+    }    
+    
+    def __init__(self, name = None):
```
chillenzer, line 27: A convenient interface could also be `**kwargs` allowing
```python
Unit(L=2, M=1, T=-2)
```
Also, see my other comment about handling `None` and `Unit` as input.

lib/python/picongpu/picmi/diagnostics/unit.py
```diff
@@ -0,0 +1,82 @@
...
+        return iter(self.unit_vector)
+
+    def __str__(self):
+        """return string representation that only outputs relevant units"""
+        output = ""
+        # invert from (name to index) to (index to (short) name)
+        inverted = {v: k for k, v in self._unit_index_map.items()}
+
+        for i in range(self.N_unit_dim):
+            if self.unit_vector[i] != 0.0:
```
chillenzer, line 56: Should we care about floating point inaccuracies here? The current value could be the result of a computation.

lib/python/picongpu/picmi/diagnostics/unit.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __iter__(self):
+        """return iterator of unit based on PIConGPU order: LMTIΘNJ"""
+        return iter(self.unit_vector)
+
+    def __str__(self):
+        """return string representation that only outputs relevant units"""
+        output = ""
+        # invert from (name to index) to (index to (short) name)
+        inverted = {v: k for k, v in self._unit_index_map.items()}
```
chillenzer, line 53: The mapping `_unit_index_map` is not invertible because it is not injective. As dictionaries are order-preserving in Python, the resulting dictionary here will depend on the order in which it is initialised up above. That could be quite surprising. But then again, a code for which a change in here would be breaking probably does some horrible log parsing that we might find to deserve the occasionally breaking.

lib/python/picongpu/picmi/diagnostics/unit.py
```diff
@@ -0,0 +1,82 @@
...
+
+        for i in range(self.N_unit_dim):
+            if self.unit_vector[i] != 0.0:
+                output += f"{inverted[i]}^{self.unit_vector[i]} "
+        return output
```
chillenzer, line 58:
```suggestion
    def __str__(self):
        """return string representation that only outputs relevant units"""
        return " ".join(f"{name}^{val} for name, index in self._unit_index_map.items() if len(name) == 1 and (val:=self.unit_vector[index]) != 0.0)
```
```

File: lib/python/picongpu/picmi/diagnostics/unit.py [hunk 11]
```diff
@@ -0,0 +1,82 @@
...
+
+    def __str__(self):
+        """return string representation that only outputs relevant units"""
+        output = ""
+        # invert from (name to index) to (index to (short) name)
+        inverted = {v: k for k, v in self._unit_index_map.items()}
+
+        for i in range(self.N_unit_dim):
+            if self.unit_vector[i] != 0.0:
+                output += f"{inverted[i]}^{self.unit_vector[i]} "
```
chillenzer, line 57: Introduces a trailing whitespace

lib/python/picongpu/picmi/diagnostics/unit.py
```diff
@@ -0,0 +1,82 @@
...
+        """multiply units with each other"""
+        result = Unit()        
+        result.unit_vector = self.unit_vector + factor.unit_vector
+        return result
+        
+    def __truediv__(self, divisor):
+        """divide one unit by another"""
+        result = Unit()
+        result.unit_vector = self.unit_vector - divisor.unit_vector
+        return result
```
chillenzer, line 76: It might be advantageous to implement as
```suggestion
    def __truediv__(self, divisor):
        """divide one unit by another"""
        return self * Unit(-divisor.unit_vector)
```
(with a corresponding `__init__`) such that any changes that we might want to do to `__mul__` in the future immediately apply to `__truediv__`.

chillenzer (changes requested):
Don't let the number of comments scare you away. It's mostly nitpicking. Only really blocking thing is the naming inconsistency "unit dimension" vs. "unit". Everything else you can ignore.

chillenzer:
PS: Of course run `pre-commit` while you're at it.

```bash
pre-commit install
pre-commit run --all-files
```

chillenzer:
Closing in favour of #5531. Sorry, I had to rebase, etc., so continuing this one was infeasible.

PR #5513, opened 2025-10-23
Add rudimentary shape support to picmi

chillenzer:
This is a very simple fall-out implementation from the work on supporting derived-field dumps via openPMD. I needed the `Counter` shape for simplifying testing, so I implemented this. Very basic implementation, just an enum. The schema is not very strict.

### DIFF
--- a/lib/python/picongpu/picmi/species.py
+++ b/lib/python/picongpu/picmi/species.py
@@ -5,6 +5,7 @@ Authors: Hannes Troepgen, Brian Edward Marre
 License: GPLv3+
 """
 
+from picongpu.pypicongpu.species.species import Shape
 from .predefinedparticletypeproperties import PredefinedParticleTypeProperties
 from .interaction import Interaction
 
@@ -239,7 +240,6 @@ class Species(picmistandard.PICMI_Species):
 
         # error on unsupported options
         pypicongpu.util.unsupported("method", self.method)
-        pypicongpu.util.unsupported("particle shape", self.particle_shape)
         # @note placement params are respected in associated simulation object
 
         self.check(interaction)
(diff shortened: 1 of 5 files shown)


lib/python/picongpu/pypicongpu/species/species.py
```diff
@@ -191,9 +211,15 @@ def _get_serialized(self) -> dict:
             else:
                 constants_context[constant_name] = None
 
+        try:
+            shape = self.shape
+        except AttributeError:
+            shape = Shape["TSC"]
+
```
ikbuibui, line 218: Why don't we simply default initialize ``self.shape`` to `TSC`?

lib/python/picongpu/pypicongpu/species/species.py
```diff
@@ -5,14 +5,32 @@
...
+    DensityRatio,
+    ElementProperties,
+    GroundStateIonization,
+    Mass,
+)
+
+
+class Shape(Enum):
+    CIC = "CIC"
+    COUNTER = "Counter"
```
ikbuibui, line 29:
```suggestion
    COUNTER = "COUNTER"
```
Maybe an upper case value is easier to deal with?

ikbuibui (approved):
Small questions for my understanding, nothing blocking

lib/python/picongpu/picmi/species.py
```diff
@@ -284,6 +284,8 @@ def get_as_pypicongpu(
         else:
             pypicongpu_model_by_picmi_model = None
 
+        s.shape = Shape[(self.particle_shape or "TSC").upper()]
```
PrometheusPi, line 287: Why do you fallback on "TSC" and how why is this not a boolean evaluation? (I am just confused what magic you are using here - I tested it and it worked.)

PrometheusPi (approved):

PR #5514, opened 2025-10-23
Minor overhaul binning in picmi

chillenzer:
Second fall-out PR from working on derived fields. This is a minor overhaul of the `Binning` implementation in PICMI. The main reason is to extract the `ParticleFunctor` into a separate file (and have a more general name) to then use it for the derived fields. I've taken the liberty to include a bit of maintenance like ordering imports, modernising the codebase, stuff like that.

As a tiny feature, the `Binning` can now report its `result_path`. It doesn't yet use the sneaked in `OpenPMDConfig` object in the front- or backend. This will be a separate PR (or added to this one if you aren't fast enough to review this here.) This should be handy for workflows in general but was mostly needed to automate the tests. Also, the `ParticleFunctor` got a `__call__` operator which currently just returns the sympy expression. It will get a second functionality with derived fields.

### DIFF
--- a/lib/python/picongpu/picmi/diagnostics/__init__.py
+++ b/lib/python/picongpu/picmi/diagnostics/__init__.py
@@ -6,7 +6,7 @@ License: GPLv3+
 """
 
 from .auto import Auto
-from .binning import Binning
+from .binning import Binning, BinningAxis, BinSpec
 from .phase_space import PhaseSpace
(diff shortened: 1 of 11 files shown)


chillenzer:
@PrometheusPi @pordyna @BrianMarre Rebased and resolved conflicts. Ready for review again.

PrometheusPi (approved):
looks good, I will run a quick test

PR #5516, opened 2025-10-24
add exclusive flag

PrometheusPi:
It seems that PIConGPU crashes due to an out-of-memory error on host side when we run without the SLURM `--exclusive` flag. This PR adds this flag. 

- [ ] perform further tests - this solution seems to be too simple 

**Do not merge yet.**


PrometheusPi:
By now, I am pretty certain that the performance issues we observed on Frontier were not related to this flag. I will close this PR.

